PR #182, opened 2020-09-22
fix/make-filter-work

decanus:
Turns the filter protocol into one that works. What we need to do for this is handle both message types in the Handle function. This means that a filter protocol can then open a new conn for every filter push.


waku/node/v2/waku_types.nim
```diff
@@ -61,12 +61,20 @@ type
   MessagePush* = object
     messages*: seq[WakuMessage]
 
+  FilterRPC* = object
```
oskarth, line 64: This is not in spec is it? If you want to change the spec I'd create a spec PR to begin with

oskarth: Considering things still don't work and this improves things and that spec is in raw I'll approve PR, but we should get into habit of having at least a spec issue before we make spec changes

oskarth (approved):

waku/node/v2/waku_types.nim
decanus: @oskarth yeah will get the spec done. My main priority was getting this merged so we can actually use filter protocol

PR #184, opened 2020-09-24
sync/filter-spec-alpha6

decanus:
Final fix for filter. Syncs it with alpha6 so we do not forget. We are generating a random requestId now.

### DIFF
--- a/waku/node/v2/waku_types.nim
+++ b/waku/node/v2/waku_types.nim
@@ -5,6 +5,7 @@
 import
   std/tables,
   chronos,
+  random,
   libp2p/[switch, peerinfo, multiaddress, crypto/crypto],
   libp2p/protobuf/minprotobuf,
   libp2p/protocols/protocol,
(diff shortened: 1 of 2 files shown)


oskarth (approved):

waku/node/v2/waku_types.nim
```diff
@@ -5,6 +5,7 @@
 import
   std/tables,
   chronos,
+  random,
```
kdeme, line 8: We should not be using stdlib random for this. Its random implementation is considered insecure. 
Can use libp2p crypto or eth keys rng for this (which is actually using the same one, the one of bearssl).

e.g. https://github.com/status-im/nim-waku/blob/a8dbf8a7b6fbf6a8dcdea0d8308236dcdcb6515e/waku/node/v2/start_network2.nim#L32

oskarth: https://github.com/status-im/nim-waku/issues/187

PR #186, opened 2020-09-24
feature/rpc-query

### LINKED ISSUE #178: End to end integration of store protocol
# Problem

There's no way to actually use store protocol right now.

Additionally: I have no confidence the store protocol does what it is supposed to do.

# Acceptance criteria

I can run two nodes, one store client and one store server. This should, as per clarified spec https://github.com/vacp2p/specs/issues/187 get historical messages for some node.

By historical messages we mean some messages there were previously published over relay protocol, for example with `rpc_publish` script.

Minimal documentation should also be provided.

# Proposed solution

This shouldn't be a complex task, but something that can be done in a day or two assuming previous work has been done. If it were me, I'd base it on examples script and use Nim Node API.

1. Start a node with store protocol enabled. This can be indicated for example by a command line flag, or simple enabled by default (with a boolean variable set to true for now).
2. Start another node and publish a few messages over relay.
3. Ensure messages are stored in-memory in the store server node, e.g. by looking at logs.
4. Start a store client node in a similar fashion to examples. This can have a hardcoded store server node and topic, and uses the Nim node `query` API.
5. Observe messages being received.
6. Write documentation markdown file describing how to replicate this.

- The topics and store server node may be hardcoded
- This may use Nim Node API or RPC (as per RPC scripts)
- It can be done as a separate binary (eg examples) or as flags to wakunode

# Notes

There's some initial work here https://github.com/status-im/nim-waku/pull/174 by @decanus but it doesn't target a specific issue.

There's also this https://github.com/status-im/nim-waku/issues/168 which @kdeme helpfully created but I think we can be more specific about what is accepted here. This issue here is about end to end integration.

decanus:
closes #178

This works nicely with the nangang test net. Simply set the `storenode` and then execute `rpc_query`. Nangang docs are updated.

### DIFF
--- a/docs/tutorial/nangang.md
+++ b/docs/tutorial/nangang.md
@@ -20,7 +20,7 @@ Run two nodes and connect them:
 ./build/wakunode2 --ports-shift:0
 
 # Run another node with staticnode argument
-./build/wakunode2 --ports-shift:1 --staticnode:/ip4/0.0.0.0/tcp/60000/p2p/16Uiu2HAmF4tuht6fmna6uDqoSMgFqhUrdaVR6VQRyGr6sCpfS2jp --storenode:/ip4/0.0.0.0/tcp/60000/p2p/16Uiu2HAmF4tuht6fmna6uDqoSMgFqhUrdaVR6VQRyGr6sCpfS2jp
+./build/wakunode2 --ports-shift:1 --staticnode:/ip4/0.0.0.0/tcp/60000/p2p/16Uiu2HAmF4tuht6fmna6uDqoSMgFqhUrdaVR6VQRyGr6sCpfS2jp
 ```
 
 You should see your nodes connecting.
@@ -31,7 +31,6 @@ Do basic RPC calls:
 ./build/rpc_subscribe 8545
 ./build/rpc_subscribe 8546
 ./build/rpc_publish 8545 # enter your message in STDIN
-./build/rpc_query 8546 # enter your topic default is "foobar"
 ```
 
 You should see other node receive something.
--- a/docs/tutorial/nangang.md
+++ b/docs/tutorial/nangang.md
@@ -20,7 +20,7 @@ Run two nodes and connect them:
 ./build/wakunode2 --ports-shift:0
 
 # Run another node with staticnode argument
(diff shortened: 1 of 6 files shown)


decanus:
Sequence of commands:
image

Output in node:
image

oskarth (approved):
Dope, so if you have a node that is offline it will get all the in memory payloads on that topic once it goes online and does rpc_query?

docs/tutorial/nangang.md
```diff
@@ -20,7 +20,7 @@ Run two nodes and connect them:
 ./build/wakunode2 --ports-shift:0
 
 # Run another node with staticnode argument
-./build/wakunode2 --ports-shift:1 --staticnode:/ip4/0.0.0.0/tcp/60000/p2p/16Uiu2HAmF4tuht6fmna6uDqoSMgFqhUrdaVR6VQRyGr6sCpfS2jp
+./build/wakunode2 --ports-shift:1 --staticnode:/ip4/0.0.0.0/tcp/60000/p2p/16Uiu2HAmF4tuht6fmna6uDqoSMgFqhUrdaVR6VQRyGr6sCpfS2jp --storenode:/ip4/0.0.0.0/tcp/60000/p2p/16Uiu2HAmF4tuht6fmna6uDqoSMgFqhUrdaVR6VQRyGr6sCpfS2jp
```
oskarth, line 23: Could we put this in a separate self-contained docs markdown file? E.g. how to use store client in general.

Nangang is already done, so store wouldn't be part of the scope retroactively. For next testnet, we'll make sure the cluster node works as a store node and it'd make sense to have it in docs for that.

docs/tutorial/nangang.md
```diff
@@ -31,6 +31,7 @@ Do basic RPC calls:
 ./build/rpc_subscribe 8545
 ./build/rpc_subscribe 8546
 ./build/rpc_publish 8545 # enter your message in STDIN
+./build/rpc_query 8546 # enter your topic default is "foobar"
```
oskarth, line 34: this is content topic right?

decanus: Yeah

decanus:
@oskarth yup exactly. Right now it doesn't do anything with it but it is to prove that query works end to end.

PR #190, opened 2020-09-28
feature/use-crypto-rng

### LINKED ISSUE #187: Use proper random function for request id
# Problem

We should not be using stdlib random for this. Its random implementation is considered insecure.

We should not be using stdlib random for this. Its random implementation is considered insecure. 
Can use libp2p crypto or eth keys rng for this (which is actually using the same one, the one of bearssl).

e.g. https://github.com/status-im/nim-waku/blob/a8dbf8a7b6fbf6a8dcdea0d8308236dcdcb6515e/waku/node/v2/start_network2.nim#L32

_Originally posted by @kdeme in https://github.com/status-im/nim-waku/pull/184#discussion_r495041835_

## Acceptance criteria

Use a better random function in `waku_type`

## Notes

Note `rng` usage in different parts of the code.

decanus:
closes #187


waku/node/v2/waku_types.nim
```diff
@@ -132,5 +132,9 @@ proc notify*(filters: Filters, msg: WakuMessage) =
         filter.handler(msg.payload)
 
 proc generateRequestId*(): string =
-  for _ in .. 10:
-    add(result, char(rand(int('A') .. int('z'))))
\ No newline at end of file
+  let 
+    rng = crypto.newRng()
```
kdeme, line 136: Should not be doing this for every `generateRequestId`. Doing it some where in an init, store this ref in an object and pass it as parameter in this call.

waku/node/v2/waku_types.nim
```diff
@@ -131,6 +134,9 @@ proc notify*(filters: Filters, msg: WakuMessage) =
       if msg.contentTopic in filter.contentFilter.topics:
         filter.handler(msg.payload)
 
-proc generateRequestId*(): string =
-  for _ in .. 10:
-    add(result, char(rand(int('A') .. int('z'))))
\ No newline at end of file
+proc generateRequestId*(rng: ref BrHmacDrbgContext): string =
```
kdeme, line 137: Something I should probably not raise here but in the vacp2p specs, but does it make sense to have this as string and not just seq[byte]?
And if we do want string, shouldn't it be hex encoded?

decanus: could be hex encoded yeah. Doesn't really matter just needs to be something random. Should I hex encode it? @kdeme

kdeme: I think that makes more sense yes, here is an example: https://github.com/status-im/nim-eth/blob/master/eth/p2p/rlpx_protocols/whisper/whisper_types.nim#L161

kdeme: Because we pass a fixed array now, we can remove the addr/unsafeaddr  and length parts. Pushed a commit for this.

kdeme (approved):

PR #194, opened 2020-09-30
fix/makefile

### LINKED ISSUE #193: Makefile naming makeover
# Problem

Currently the Makefile targets are a bit inconsistent which makes it hard to remember what to type.

https://github.com/status-im/nim-waku/blob/master/Makefile#L71-L105

# Proposed solution

1) Remove `waku` prefix from targets
2) Consistently append name all targets with either 1 or 2

# Acceptance criteria

Makefile done over and clarified, including the Nimbus target file.

--

cc @kdeme @decanus for thoughts

decanus:
closes #193

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,9 +22,9 @@ DOCKER_IMAGE_NIM_PARAMS ?= -d:chronicles_colors:none -d:insecure
 	deps \
 	update \
 	sim \
-	node \
-	node2 \
-	example \
+	wakunode1 \
+	wakunode2 \
+	example1 \
 	test \
 	clean \
 	libbacktrace
@@ -45,7 +45,7 @@ GIT_SUBMODULE_UPDATE := git submodule update --init --recursive
 else # "variables.mk" was included. Business as usual until the end of this file.
 
 # default target, because it's the first one that doesn't start with '.'
-all: | node sim example node2 sim2 example2
+all: | wakunode1 sim example wakunode2 sim2 example2
 
 # must be included after the default target
 -include $(BUILD_SYSTEM_DIR)/makefiles/targets.mk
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -68,27 +68,27 @@ update: | update-common
 		$(MAKE) waku.nims $(HANDLE_OUTPUT)
 
 # a phony target, because teaching `make` how to do conditional recompilation of Nim projects is too complicated
-wakunode: | build deps
+node: | build deps
```
kdeme, line 71: wakunode (and version 2) is perhaps the only one I would keep as is, as the application has the same name, but no big preference really.

decanus: I would be more happy with changing the wakunode application name to node. Seems more logical, if I am in the waku repository I of course expect the node to be waku.

oskarth: It is a good point though, because wakunode is expected to be put in your PATH. E.g. see Dockerfile. You don't want to confuse wakunode with NodeJS ;)

All the rest are only used in the repository so can be more local, afaik.

Makefile
```diff
@@ -21,10 +21,10 @@ DOCKER_IMAGE_NIM_PARAMS ?= -d:chronicles_colors:none -d:insecure
...
 	deps \
 	update \
-	wakusim \
-	wakunode \
-	wakunode2 \
-	wakuexample \
+	sim \
+	node \
+	node2 \
+	example \
```
oskarth, line 27: I'd call example example1 here to make it clear.

This means test refers to both, test1 to v1 and test2 to v2

Makefile
oskarth: I'd also call this wakunode1, wakunode should be symlinked to the latest stable release (e.g)

oskarth (changes requested):

Makefile
```diff
@@ -21,10 +21,10 @@ DOCKER_IMAGE_NIM_PARAMS ?= -d:chronicles_colors:none -d:insecure
 	all \
 	deps \
 	update \
-	wakusim \
-	wakunode \
+	sim \
```
oskarth, line 24:
```suggestion
	sim1 \
```

oskarth: Plus other references

Makefile
```diff
@@ -21,10 +21,10 @@ DOCKER_IMAGE_NIM_PARAMS ?= -d:chronicles_colors:none -d:insecure
...
+	wakunode1 \
 	wakunode2 \
-	wakuexample \
+	example1 \
```
oskarth, line 27: I'd add example2 here too

oskarth (commented):
Almost there

oskarth (approved):

PR #202, opened 2020-10-05
Node Info API call and JSON RPC

oskarth:
Closes https://github.com/status-im/nim-waku/issues/201

- Adds a Info API call and JSON RPC call
- Fixes some unrelated errors in query rpc and makefile
- Update node api docs

This allows one to get the listen string by JSON RPC, which is needed for cluster health check.

(Further enhancements can be done to display more diagnostic information)

## Example

```
> ./build/rpc_info 
DBG 2020-10-05 13:18:07.528+08:00 Message sent to RPC server                 topics="JSONRPC-HTTP-CLIENT" tid=22956 file=httpclient.nim:188 address=127.0.0.1:8545 msg_len=57
Waku info res: /ip4/0.0.0.0/tcp/60000/p2p/16Uiu2HAmUehj5XstFP41SEmtmbxjkYjogtL3G14Rn4Tsj1ya2nQh
> 
```

### DIFF
--- a/waku/node/v2/rpc/rpc_info.nim
+++ b/waku/node/v2/rpc/rpc_info.nim
@@ -12,19 +12,8 @@ template sourceDir: string = currentSourcePath.rsplit(DirSep, 1)[0]
 const sigWakuPath = sourceDir / "wakucallsigs.nim"
 createRpcSigs(RpcHttpClient, sigWakuPath)
 
-if paramCount() < 1:
-  echo "Please provide rpcPort as argument."
-  quit(1)
-
-let rpcPort = Port(parseInt(paramStr(1)))
-
-echo "Please enter your topic:"
-let raw_input = readLine(stdin)
-let input = fmt"{raw_input}"
-echo "Input is:", input
-
 var node = newRpcHttpClient()
-waitfor node.connect("localhost", rpcPort)
+waitfor node.connect("localhost", Port(8545))
 
-var res = waitfor node.wakuQuery(@[input])
-echo "Waku query response: ", res
+var res = waitfor node.wakuInfo()
+echo "Waku info res: ", res
--- a/Makefile
+++ b/Makefile
@@ -105,10 +105,6 @@ test2:
 	echo -e $(BUILD_MSG) "build/$@" && \
 		$(ENV_SCRIPT) nim test2 $(NIM_PARAMS) waku.nims
 
-example2:
-	echo -e $(BUILD_MSG) "build/$@" && \
-		$(ENV_SCRIPT) nim example2 $(NIM_PARAMS) waku.nims
-
 chat2:
 	echo -e $(BUILD_MSG) "build/$@" && \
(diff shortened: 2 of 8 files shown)


oskarth:
@jakubgs is this good enough for your purposes?

jakubgs:
Not sure how I'm supposed to build this:
```
 > git submodule update --init --recursive

 > make scripts2 V=6                      
Git submodules not found. Running 'git submodule update --init --recursive'.

Git submodules not found. Running 'git submodule update --init --recursive'.

 > ls -l build
total 0

```

oskarth:
Did you do `make update`? You shouldn't need the V=6

jakubgs:
Yes I have, they appear to do nothing:
```
 > make
Git submodules not found. Running 'git submodule update --init --recursive'.

make: *** No targets.  Stop.

 > make update
Git submodules not found. Running 'git submodule update --init --recursive'.

make: Nothing to be done for 'update'.
```

jakubgs:
I purged contents of `vendor` and then did `git checkout .` and `git submodule update --init --recursive` and now it builds.
Weird but okay...

decanus (approved):

jakubgs (changes requested):
If we want to use this on the fleet then we'll need to include this tool in the Docker image.
Probably another make target after this one?
https://github.com/status-im/nim-waku/blob/065016393a3c27720f543b00517d405fed94568a/Dockerfile#L20-L21
Though not sure how that will work with `MAKE_TARGET`, since you have no `scripts`, just `scipts2`.

jakubgs:
One option that I considered once would be to just drop `MAKE_TARGET` parameter from the `Dockerfile` and just include both `v1` and `v2` binaries and scripts in one image to avoid messy logic.

oskarth:
You don't need to use the script tool to do JSON RPC. It is just a helper. It should work the same way as it did for v1 I believe.

In any case I think we can do more tooling stuff in a separate PR if needed, this is more about adding support to the Node API.

jakubgs (approved):
Okay, if we don't need the script than I'm cool with it.

waku/node/v2/rpc/rpc_info.nim
```diff
@@ -0,0 +1,19 @@
...
+from strutils import rsplit
+template sourceDir: string = currentSourcePath.rsplit(DirSep, 1)[0]
+
+const sigWakuPath = sourceDir / "wakucallsigs.nim"
+createRpcSigs(RpcHttpClient, sigWakuPath)
+
+var node = newRpcHttpClient()
+waitfor node.connect("localhost", Port(8545))
+
+var res = waitfor node.wakuInfo()
```
kdeme, line 18: Would be better to change this into a test.

oskarth: Yeah agree, was on my TODO but decided to skip it for now (a) to get it merged (b) until we know what we want with JSON RPC (c) until there is more meat in the WakuInfo call

PR #209, opened 2020-10-05
feature/start-protocol-command

### LINKED ISSUE #207: Start protocol command line arguments
# Problem

Currently when starting a node, `relay`, `store`, and `filter` protocols are started automatically.

# Acceptance Criteria

There should be command line flags that allow a user to signal which protocols should be started.

decanus:
closes #207


decanus:
this is sadly still buggy, in the tests. Need to figure out why

decanus:
@oskarth @kdeme this works, but it seems a bit ugly. Any suggestions on cleanup?

waku/node/v2/config.nim
```diff
@@ -73,6 +73,16 @@ type
       defaultValue: ""
       name: "storenode" }: string
 
+    store* {.
```
oskarth, line 76: Do we want to have relay here too?

waku/node/v2/wakunode2.nim
```diff
@@ -132,6 +119,21 @@ proc start*(node: WakuNode) {.async.} =
   ## XXX: this should be /ip4..., / stripped?
   info "Listening on", full = listenStr
 
+proc startFilter*(node: WakuNode) =
```
oskarth, line 122: Maybe log here?

waku/node/v2/wakunode2.nim
```diff
@@ -272,6 +277,12 @@ when isMainModule:
 
   waitFor node.start()
 
+  if conf.store:
+    startStore(node)
```
oskarth, line 281: Is it starting it vs initing or mounting? Seems like latter to me

oskarth (approved):
In general lgtm, left some comments and suggestions

oskarth:
> suggestions on clean up

I'd look at how NBC does it probably

waku/node/v2/config.nim
decanus: @oskarth probably

waku/node/v2/wakunode2.nim
```diff
@@ -292,6 +299,15 @@ when isMainModule:
...
   waitFor node.start()
 
+  if conf.store:
+    mountStore(node)
+  
+  if conf.filter:
+    mountFilter(node)
+
+  if conf.filter:
+    await mountRelay(node)
```
kdeme, line 309: You'll have to do waitFor here.

waku/node/v2/config.nim
```diff
@@ -73,6 +73,21 @@ type
...
+      name: "store" }: bool
+
+    filter* {.
+      desc: "Flag whether to start filter protocol",
+      defaultValue: false
+      name: "filter" }: bool
+    
+    relay* {.
+      desc: "Flag whether to start relay protocol",
+      defaultValue: false
```
kdeme, line 88: I guess this can be true? Or one of the protocols at least? Else nothing starts default?

kdeme (commented):
I kinda had in mind not exposing this relay, filter and store (perhaps not the latter) directly to the user.
Sounds awfully difficult and I think the WakuNode abstraction could take care of this?

decanus:
> I kinda had in mind not exposing this relay, filter and store (perhaps not the latter) directly to the user.
> Sounds awfully difficult and I think the WakuNode abstraction could take care of this?

@kdeme not sure what you are saying here.

kdeme:
What I meant is that the application should probably have cli options in the trend of: `--full-node:true` or `--content-topics-only:blabla` (or `--light-node:true` if you want), `--historical-node:false`, etc. And then this could be mapped to the right protocols, which would be the role of the WakuNode abstraction imo. 
But ok, we haven't thought well about this yet, so I guess it is fine for testing for now.

decanus:
@kdeme what do you think of the new flags?

waku/node/v2/config.nim
```diff
@@ -73,6 +73,21 @@ type
       defaultValue: ""
       name: "storenode" }: string
 
+    historicalnode* {.
```
oskarth, line 76: Sorry but these three ones seems worse to me, the previous one was simplistic but clear. Full node eg seems like a misnormer.

I'd make sure we know what we are doing and what we want before we increase the sophistication. KISS. Make sure we understand the problem well enough before writing a solution for it. See https://github.com/vacp2p/specs/issues/167 for context

oskarth:
> What I meant is that the application should probably have cli options in the trend of: --full-node:true or --content-topics-only:blabla (or --light-node:true if you want), --historical-node:false, etc. And then this could be mapped to the right protocols, which would be the role of the WakuNode abstraction imo.
But ok, we haven't thought well about this yet, so I guess it is fine for testing for now.

Full node seems off to me. What does that mean for protocols? Not obvious. Content topics - boolean or list? For client or for receiver? Light node in what context? Historical node as server? On what topic?

What behaviour should be done for each of these aside from mounting the protocol?

All of this should be discussed at the problem formulation state, not in a PR imo.

kdeme:
I was merely giving these as example to show that the current solution is not fine for the end user, not to actually copy over and use as is. Hence the "in the trend of". 
And I also stated that I'm fine with merging this as is, as we haven't resolved that issue.

waku/node/v2/wakunode2.nim
```diff
@@ -125,20 +128,21 @@ proc start*(node: WakuNode) {.async.} =
   node.switch.mount(node.wakuFilter)
   node.subscriptions.subscribe(WakuFilterCodec, node.wakuFilter.subscription())
 
+proc mountStore*(node: WakuNode) =
+  info "mounting store"
+  node.wakuStore = WakuStore.init(node.switch, node.rng)
+  node.switch.mount(node.wakuStore)
+  node.subscriptions.subscribe(WakuStoreCodec, node.wakuStore.subscription())
+
+proc mountRelay*(node: WakuNode) {.async, gcsafe.} =
```
kdeme, line 137: In mountRelay nothing gets mounted, as this happens in `init` already. 
Why this difference with the other protocols exactly? Could it be made consistent? Or would there be a downside to that?

Or could all get mounted in `init`, and then just run subscriptions/subscribe per protocol to activate it?
What would that do on a libp2p level? How would the connections look at after that `init` state then?

decanus:
ok so let me revert to the previous.

oskarth:
What's the state of this PR?

decanus:
@oskarth theres some weird issues with the PR, I am working on fixing them.

decanus:
@oskarth fixed. ptal

decanus:
@oskarth should we modify the docker file to start the clusternode as a `wakunode2` running all 3 protocols?

oskarth:
> @oskarth should we modify the docker file to start the clusternode as a wakunode2 running all 3 protocols?

This is the current behaviour right? And yeah we should

oskarth:
Why is this PR touching nim-libp2p etc? Is it missing a rebase? I pushed a recent bump of submodules so I don't think we need one in this one.

waku/node/v2/wakunode2.nim
```diff
@@ -168,8 +124,11 @@ proc subscribe*(node: WakuNode, request: FilterRequest, handler: ContentFilterHa
   ## Status: Implemented.
   info "subscribe content", filter=request
 
-  # @TODO: ERROR HANDLING
-  let id = await node.wakuFilter.subscribe(request)
+  # @TODO: GENERATE
```
oskarth, line 127: Why is this a TODO? It looks like it is doing it

waku/node/v2/wakunode2.nim
```diff
@@ -223,6 +182,56 @@ proc info*(node: WakuNode): WakuInfo =
...
+  node.switch.mount(node.wakuFilter)
+  node.subscriptions.subscribe(WakuFilterCodec, node.wakuFilter.subscription())
+
+proc mountStore*(node: WakuNode) =
+  info "mounting store"
+  node.wakuStore = WakuStore.init(node.switch, node.rng)
+  node.switch.mount(node.wakuStore)
+  node.subscriptions.subscribe(WakuStoreCodec, node.wakuStore.subscription())
+
+proc mountRelay*(node: WakuNode, topics: seq[string] = newSeq[string]()) {.async, gcsafe.} =
```
oskarth, line 202: Out of curiosity, does it work when not mounting relay?

waku/node/v2/wakunode2.nim
```diff
@@ -223,6 +182,56 @@ proc info*(node: WakuNode): WakuInfo =
...
+  await sleepAsync(5.seconds)
+
+  info "mounting relay"
+  proc relayHandler(topic: string, data: seq[byte]) {.async, gcsafe.} =
+    let msg = WakuMessage.init(data)
+    if msg.isOk():
+      node.filters.notify(msg.value(), "")
+      await node.subscriptions.notify(topic, msg.value())
+
+  await node.wakuRelay.subscribe("waku", relayHandler)
```
oskarth, line 224: How does this interact with the other PR? Are they supposed to be merged in a specific order? I see later that we have mountRelay with topics as options, so why not use that argument?

waku/node/v2/wakunode2.nim
```diff
@@ -300,10 +309,19 @@ when isMainModule:
...
 
   waitFor node.start()
 
+  if conf.store:
+    mountStore(node)
+  
+  if conf.filter:
+    mountFilter(node)
+
+  if conf.filter:
```
oskarth, line 322:
```suggestion
  if conf.relay:
```

Typo

oskarth (changes requested):

waku/node/v2/wakunode2.nim
decanus: It does not seem to work right now, I can investigate further why not.

waku/node/v2/wakunode2.nim
decanus: It doesn't really matter how it is merged as long as the five second sleep is there before subscribing from my understanding.

waku/node/v2/wakunode2.nim
decanus: @oskarth so I started 3 nodes:
 A) One with Relay, Filter and Store
 B) One with Relay
 C) One with Store

I sent a message from B, it was received by A, I then queried from C and received the expected message. So I would say it does work.

oskarth (changes requested):
Previous feedback not addressed: vendor removal, dockerfile, (possibly more?).

Please comment on previous feedback with ack or so so it is clear if you have resolved it.

waku/node/v2/config.nim
oskarth: This is still false.

decanus:
Alright @oskarth, docker has been addressed as have gitsubmodules and default option.

oskarth (approved):

decanus:
@kdeme ptal

waku/node/v2/wakunode2.nim
```diff
@@ -223,6 +181,56 @@ proc info*(node: WakuNode): WakuInfo =
...
+    #msgIdProvider = msgIdProvider,
+    triggerSelf = true,
+    sign = false,
+    verifySignature = false
+  )
+
+  node.wakuRelay = wakuRelay
+  node.switch.mount(wakuRelay)
+
+  await sleepAsync(5.seconds)
```
oskarth, line 214: This was added here

PR #210, opened 2020-10-06
Begin of waku v2 rpc tests

kdeme:


kdeme:
Started a basic rpc test as mentioned here: https://github.com/status-im/nim-waku/pull/202#discussion_r500035124

Merged this already as it is better than no test.

PR #211, opened 2020-10-06
Cleanup warnings in v1 and split test target in Makefile

kdeme:
Some cleanup in v1. And split the test in Makefile so you can run v1 tests separately.


Makefile
```diff
@@ -93,22 +96,29 @@ sim2: | build deps wakunode2
...
 	echo -e $(BUILD_MSG) "build/$@" && \
-		$(ENV_SCRIPT) nim test2 $(NIM_PARAMS) waku.nims
+		$(ENV_SCRIPT) nim scripts2 $(NIM_PARAMS) waku.nims
 
-chat2:
+chat2: | build deps
 	echo -e $(BUILD_MSG) "build/$@" && \
 		$(ENV_SCRIPT) nim chat2 $(NIM_PARAMS) waku.nims
 
+protocol2: | build deps
```
kdeme, line 115: Does anybody still care about this target? Else I remove it...

oskarth: can remove

oskarth (approved):

PR #212, opened 2020-10-07
Is it because of an update?

kdeme:
Don't merge, testing if CI is so slow because this gets updated or not.


kdeme:
Update on the v0 was showing a lot of errors and timeouts. Not sure if it actually finished it successfully as it seems to redo this again the commit after.
Tried v2 and it is faster and without errors, so lets go with this for now.

I think we should probably set this up ourselves though not to be depended on this.

PR #224, opened 2020-10-13
Pagination: Pagination feature/datatypes

staheri14:
This PR addresses [this issue](https://github.com/vacp2p/specs/issues/214).

### DIFF
--- a/waku/node/v2/waku_types.nim
+++ b/waku/node/v2/waku_types.nim
@@ -41,6 +41,11 @@ type
     msg*: WakuMessage
     index*: Index
 
+  # PagingInfo holds the information needed for the pagination
+  PagingInfo* = object
+    pageSize*: int
+    cursor*: Index
+    direction*: bool
 
   HistoryQuery* = object
     topics*: seq[string]
--- a/waku/node/v2/waku_types.nim
+++ b/waku/node/v2/waku_types.nim
@@ -34,6 +34,7 @@ type
 
 
   Index* = object
+    ## This type contains the  description of an index used in the pagination of waku messages
     digest*: string
     timestamp*: string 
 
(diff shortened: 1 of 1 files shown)


waku/node/v2/waku_types.nim
```diff
@@ -32,11 +32,29 @@ type
...
+
+  PagingInfo* = object
+    ## This type holds the information needed for the pagination 
+    pageSize*: int
+    cursor*: Index
+    direction*: bool
+
   HistoryQuery* = object
     topics*: seq[string]
+    pagingInfo*: PagingInfo
```
oskarth, line 53: Might want to have this be optional for implementations that don't want to care about this.

oskarth (approved):
Seems reasonable to me! Would be good with a corresponding protobuf definition in the specs to make sure the semantics are the same.

Then the protobuf encode/decode has to be done, but can be done separately if you prefer :)

waku/node/v2/waku_types.nim
staheri14: Sure, for a HistoryQuery with an unspecified PagingInfo, no pagination will take place.

decanus (approved):
@staheri14 lgtm, next time I would recommend doing logic and data types in the same PR though. Makes it less confusing.

oskarth:
Huh I was confused why this PR didn't show up in master and why I got a notification re new commit to branch. Turns out it was targeting `feature/pagination` branch, not master.

@staheri14 Sorry I should've been more clear about this, next time please target the PR to `origin/master` :) We generally practice trunk based development (https://trunkbaseddevelopment.com/)

oskarth:
@staheri14 Would you be able to re-open this PR and change the target branch to be master? I can't reopen this PR but only create a new one in my name.

Once that's done, we can merge it, rebase the other PR on top of master and change that target as well.

Alternatively, if you want to change the target of the other PR and incorporate them into the same PR.

PR #225, opened 2020-10-14
Pagination: Pagination feature/compute index

staheri14:
This PR addresses [this issue](https://github.com/vacp2p/specs/issues/215).

### DIFF
--- a/tests/v2/test_waku_pagination.nim
+++ b/tests/v2/test_waku_pagination.nim
@@ -1,28 +1,25 @@
 import
-    std/unittest, times,
-    ../../waku/node/v2/waku_types,
-    ../test_helpers,
-    nimcrypto/sha2
+    std/unittest,
+    ../../waku/node/v2/waku_types
 
-procSuite "pagination":
+suite "pagination":
+  test "computeIndex: empty contentTopic test":
+      let
+          wm = WakuMessage(payload: @[byte 1, 2, 3])
+          index = wm.computeIndex()
+      check:
+          # the fields of the index should be non-empty
+          len(index.digest.data) != 0
+          index.receivedTime != 0
 
-    test "computeIndex: empty contentTopic test":
-        let
-            wm = WakuMessage(payload: @[byte 1, 2, 3])
-            index = wm.computeIndex()
-        check:
-            # the fields of the index should be non-empty
-            index.digest.len != 0
-            index.receivedTime != 0
+  test "computeIndex: identical WakuMessages test":
+      let
+          wm = WakuMessage(payload: @[byte 1, 2, 3], contentTopic: "topic2")
+          index1 = wm.computeIndex()
+          wm2 = WakuMessage(payload: @[byte 1, 2, 3], contentTopic: "topic2")
+          index2 = wm2.computeIndex()
 
-    test "computeIndex: identical WakuMessages test":
-        let
-            wm = WakuMessage(payload: @[byte 1, 2, 3], contentTopic: "topic2")
-            index1 = wm.computeIndex()
-            wm2 = WakuMessage(payload: @[byte 1, 2, 3], contentTopic: "topic2")
-            index2 = wm2.computeIndex()
-
-        check:
-            # the digests of two identical WakuMessages must be the same
-            index1.digest == index2.digest
+      check:
+          # the digests of two identical WakuMessages must be the same
+          index1.digest == index2.digest
 
--- a/tests/v2/test_waku_pagination.nim
+++ b/tests/v2/test_waku_pagination.nim
@@ -6,22 +6,31 @@ import
 
 procSuite "pagination":
 
-    test "computeIndex empty contentTopic test":
-        let 
+    test "computeIndex: empty contentTopic test":
+        let
             wm = WakuMessage(payload: @[byte 1, 2, 3])
-            index=wm.computeIndex()
+            index = wm.computeIndex()
         check:
             # the fields of the index should be non-empty
             index.digest.len != 0
-            index.receivedTime!=0 
-        
-    test "computeIndex identical WakuMessages test":
-        let 
+            index.receivedTime != 0
+
+    test "computeIndex: identical WakuMessages test":
+        let
             wm = WakuMessage(payload: @[byte 1, 2, 3], contentTopic: "topic2")
-            index1=wm.computeIndex()
+            index1 = wm.computeIndex()
             wm2 = WakuMessage(payload: @[byte 1, 2, 3], contentTopic: "topic2")
-            index2=wm2.computeIndex()
-            
+            index2 = wm2.computeIndex()
+
         check:
             # the digests of two identical WakuMessages must be the same
-            index1.digest==index2.digest
+            index1.digest == index2.digest
+
+    test "computeIndex: empty WakuMessages test":
+        let
+            wm = WakuMessage(payload: @[byte 1, 2, 3], contentTopic: "topic2")
+            index1 = wm.computeIndex()
+
+        check:
+            # the digests of two identical WakuMessages must be the same
+            index1.digest == index2.digest
--- a/tests/v2/test_waku_pagination.nim
+++ b/tests/v2/test_waku_pagination.nim
@@ -1,8 +1,9 @@
 import
   std/unittest,
-  ../../waku/node/v2/waku_types
+  ../../waku/node/v2/waku_types,
+  ../test_helpers
 
-suite "pagination":
+procSuite "pagination":
   test "computeIndex: empty contentTopic test":
       let
           wm = WakuMessage(payload: @[byte 1, 2, 3])
@@ -10,7 +11,8 @@ suite "pagination":
       check:
           # the fields of the index should be non-empty
           len(index.digest.data) != 0
-          index.receivedTime != 0
+          len(index.digest.data) == 32 # sha2 output length in bytes
+          index.receivedTime != 0 # the timestamp should be a non-zero value
 
   test "computeIndex: identical WakuMessages test":
       let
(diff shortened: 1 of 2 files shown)


waku/node/v2/waku_types.nim
```diff
@@ -154,3 +174,12 @@ proc generateRequestId*(rng: ref BrHmacDrbgContext): string =
   var bytes: array[10, byte]
   brHmacDrbgGenerate(rng[], bytes)
   toHex(bytes)
+
+proc computeIndex* (w: WakuMessage): Index=
```
oskarth, line 178:
```suggestion
proc computeIndex* (w: WakuMessage): Index =
```

waku/node/v2/waku_types.nim
```diff
@@ -154,3 +174,12 @@ proc generateRequestId*(rng: ref BrHmacDrbgContext): string =
...
+  ## Takes a WakuMessage and returns its index
+  let payloadStr=w.payload.join("") #convers payload to string
```
oskarth, line 180:
```suggestion
  let payloadStr=w.payload.join("") # converts payload to string
```

waku/node/v2/waku_types.nim
```diff
@@ -154,3 +174,12 @@ proc generateRequestId*(rng: ref BrHmacDrbgContext): string =
...
+  var data: string
+  data.add(w.contentTopic)
+  data.add(payloadStr)
+  result.digest= $secureHash(data) #computes sha1 of data
+  result.timeStamp= $getTime().utc #gets the timestamp in utc with seconds precision
```
oskarth, line 185:
```suggestion
proc computeIndex*(w: WakuMessage): Index =
  ## Takes a WakuMessage and returns its index
  let payloadStr = w.payload.join("") # converts payload to string
  var data: string
  data.add(w.contentTopic)
  data.add(payloadStr)
  result.digest = $secureHash(data) # computes sha1 of data
  result.timeStamp = $getTime().utc # gets the timestamp in utc with seconds precision
```

oskarth: you can also use `&` for string concatenation, so it could probably be written something like (untested, assuming $ is implemented on payload):

let data = msg.contentTopic & $msg.payload

oskarth: Few more comments:

- I'd make it clear that the timestamp is for receiver node, so `receivedTime` or something like this.

- I believe contentTopic is optional, so might want to make sure this is handled.

- Perhaps a test here would be useful to make sure it works as intended?

oskarth (commented):
See above

waku/node/v2/waku_types.nim
```diff
@@ -32,11 +34,29 @@ type
 
   QueryHandlerFunc* = proc(response: HistoryResponse) {.gcsafe, closure.}
 
+
+  Index* = object
+    ## This type contains the  description of an index used in the pagination of waku messages
+    digest*: string
+    timestamp*: string
```
kdeme, line 41: I think string is not needed here and `uint32` could be used as this is probably going to be a unix timestamp?

waku/node/v2/waku_types.nim
kdeme, line 185: What about `epochTime`, wouldn't that be easier to work with?

staheri14: I can also use `epochTime`, it gives a `float` value, however, the documentation of `epochTime` proc recommends  `getTime` instead of `epochTime`:
> "``getTime`` should generally be preferred over this proc."
I am not sure why.

But storage-wise, float is better than string.

waku/node/v2/waku_types.nim
staheri14: I will change the timestamp to unix timestamp (which is of type `float`)

waku/node/v2/waku_types.nim
```diff
@@ -154,3 +174,12 @@ proc generateRequestId*(rng: ref BrHmacDrbgContext): string =
...
   brHmacDrbgGenerate(rng[], bytes)
   toHex(bytes)
+
+proc computeIndex* (w: WakuMessage): Index=
+  ## Takes a WakuMessage and returns its index
+  let payloadStr=w.payload.join("") #convers payload to string
+  var data: string
+  data.add(w.contentTopic)
+  data.add(payloadStr)
+  result.digest= $secureHash(data) #computes sha1 of data
```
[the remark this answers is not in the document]
staheri14: Thanks! I will use nimcrypto instead, could you also please give me a resource for the insecurity of stdlib hash implementation, I would like to read more about it.

waku/node/v2/waku_types.nim
staheri14: > Perhaps a test here would be useful to make sure it works as intended?

I was going to implement the tests in the next issue, but per your comment, I'll include it in this PR.

staheri14: > I'd make it clear that the timestamp is for receiver node, so receivedTime or something like this.
> 
> I believe contentTopic is optional, so might want to make sure this is handled.

Thanks, addressed!

oskarth:
See https://github.com/status-im/nim-waku/pull/224#issuecomment-708884595 re branches etc. This PR also needs to be rebased on master I believe.

@staheri14 ping here / re-request reviews once it is ready!

waku/node/v2/waku_types.nim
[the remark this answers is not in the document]
kdeme: I'm honestly not really sure if it is insecure, or how insecure it is. We use nimcrypto hashes in our projects and that code has been audited for our nimbus-eth2 project.  So I put a lot more thrust in that codebase. I don't think the stdlib code has ever been audited, but I might be mistaken.

Additionally, sha1 should not be used any more for secure hashing in new projects. Stdlib doesn't have any other hashing implementations (aside from MD5, heh lets not go there). 
Also, a smaller item, the APIs from the stdlib are different for the different hash implementations, which is just weird.

waku/node/v2/waku_types.nim
```diff
@@ -32,11 +36,29 @@ type
 
   QueryHandlerFunc* = proc(response: HistoryResponse) {.gcsafe, closure.}
 
+
+  Index* = object
+    ## This type contains the  description of an index used in the pagination of waku messages
+    digest*: string
```
kdeme, line 42: I would hold this value here as `MDigest[256]`, and send it over the wire as byte array.

waku/node/v2/waku_types.nim
```diff
@@ -154,3 +176,14 @@ proc generateRequestId*(rng: ref BrHmacDrbgContext): string =
...
   toHex(bytes)
+
+proc computeIndex*(msg: WakuMessage): Index =
+  ## Takes a WakuMessage and returns its index
+  var data: string
+  if msg.contentTopic.len != 0: # checks for non-empty contentTopic
+    data = msg.contentTopic
+  let payloadStr = msg.payload.join("") # converts payload to string
+  data = data & payloadStr
+  result.digest = $sha256.digest(data) # computes the hash
```
kdeme, line 187: When stored as MDigest, you can just use something like:
```nim 
var ctx: keccak256
ctx.init()
ctx.update(msg.contentTopic.toBytes()) # Use the one from nim-stew byteutils
ctx.update(msg.payload)
let digest = ctx.finish()
```

waku/node/v2/waku_types.nim
staheri14: Sure, done

staheri14:
> See [#224 (comment)](https://github.com/status-im/nim-waku/pull/224#issuecomment-708884595) re branches etc. This PR also needs to be rebased on master I believe.
> 
> @staheri14 ping here / re-request reviews once it is ready!

@oskarth, I rebased the current PR to master, it is ready for merge with master.

oskarth (approved):
LGTM!

Once @decanus approves, feel free to merged to master at your own leisure @staheri14. The only thing to make sure it is rebased on latest master and that CI (at least Ubuntu) passes.

If you are unable to merge for some reason (access control or w/e), please ping @decanus or me and we can merge it.

tests/v2/test_waku_pagination.nim
```diff
@@ -0,0 +1,28 @@
+import
+    std/unittest, times,
+    ../../waku/node/v2/waku_types,
+    ../test_helpers,
+    nimcrypto/sha2
+
+procSuite "pagination":
```
decanus, line 7: @staheri14 we use 2 spaces for indentation.

decanus (changes requested):

tests/v2/test_waku_pagination.nim
```diff
@@ -0,0 +1,25 @@
...
+      let
+          wm = WakuMessage(payload: @[byte 1, 2, 3], contentTopic: "topic2")
+          index1 = wm.computeIndex()
+          wm2 = WakuMessage(payload: @[byte 1, 2, 3], contentTopic: "topic2")
+          index2 = wm2.computeIndex()
+
+      check:
+          # the digests of two identical WakuMessages must be the same
+          index1.digest == index2.digest
+
```
decanus, line 25: @staheri14 this all should be 2 space indentation rather than 4. The mix won't work.

decanus (changes requested):

tests/v2/test_waku_pagination.nim
staheri14: Thanks! Addressed the indentation issues.

waku/node/v2/waku_types.nim
```diff
@@ -3,14 +3,17 @@
 ## TODO Move more common data types here
 
 import
-  std/tables,
+  std/[tables, sha1],
+  times,
```
kdeme, line 7:
```suggestion
  std/[tables, times],
```
Not necessary and more of a nitpick but something we've recently started doing in the Nimbus team is put `std/` in front of all stdlib imports, that way it is immediately visible which imports are from the stdlib. 
Also, I guess sha1 can be dropped here?

oskarth:
I think you missed this review yesterday @decanus? Looks like it was just formatting that was addressed, going ahead and merging.

PR #226, opened 2020-10-14
Fix timing issue with subscribe in chat2

### LINKED ISSUE #215: Cluster node not relaying messages (deployed recently?)
# Problem

After https://github.com/status-im/nim-waku/issues/198 I can't use the cluster node in e.g. `chat2`.

Looking at [Kibana](https://kibana.status.im/app/kibana#/discover?_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:%272020-09-09T20:21:49.910Z%27,to:now))&_a=(columns:!(message,severity_name),filters:!((%27$state%27:(store:appState),meta:(alias:!n,disabled:!f,index:d6db7610-60fd-11e9-98fa-2f101d13f938,key:program.keyword,negate:!f,params:(query:docker%2Fnim-waku-node),type:phrase),query:(match_phrase:(program.keyword:docker%2Fnim-waku-node))),(%27$state%27:(store:appState),meta:(alias:!n,disabled:!f,index:d6db7610-60fd-11e9-98fa-2f101d13f938,key:fleet.keyword,negate:!f,params:(query:wakuv2.test),type:phrase),query:(match_phrase:(fleet.keyword:wakuv2.test)))),index:d6db7610-60fd-11e9-98fa-2f101d13f938,interval:auto,query:(language:kuery,query:Listening),sort:!())) it seems like the last deploy was a week ago.

It is unclear to me exactly how https://github.com/status-im/nim-waku/issues/198 was tested end to end with chat2.

## Details

Looking at [recent logs](https://kibana.status.im/app/kibana#/discover?_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:'2020-10-12T22:23:04.892Z',to:now))&_a=(columns:!(message,severity_name,_type),filters:!(('$state':(store:appState),meta:(alias:!n,disabled:!f,index:d6db7610-60fd-11e9-98fa-2f101d13f938,key:program.keyword,negate:!f,params:(query:docker%2Fnim-waku-node),type:phrase),query:(match_phrase:(program.keyword:docker%2Fnim-waku-node))),('$state':(store:appState),meta:(alias:!n,disabled:!f,index:d6db7610-60fd-11e9-98fa-2f101d13f938,key:fleet.keyword,negate:!f,params:(query:wakuv2.test),type:phrase),query:(match_phrase:(fleet.keyword:wakuv2.test)))),index:d6db7610-60fd-11e9-98fa-2f101d13f938,interval:auto,query:(language:kuery,query:waku),sort:!())), I see a few suspicious things that may or may not be related:

`DBG 2020-10-13 03:10:26.006+00:00 Error generated                            tid=1 file=router.nim:115 error="{"code":-32601,"id":1,"message":"Method not found","data":"admin_nodeInfo is not a registered RPC method."}" id=1`

Though no ERR or WAR AFAICT.

To be investigated further.

oskarth:
Fixes https://github.com/status-im/nim-waku/issues/221
Fixes #215 

The problem is the timing of subscribe and connection of nodes. In theory it shouldn't matter I suppose, but in practice it does. Additionally, there seems to be a need for a sleep before subscribe being called. Otherwise, a GRAFT message isn't sent. This means the number of peers in the relevant mesh is zero. For two nodes this isn't visible because of `floodPublish` setting.

It is unclear when or how this was introduced, but it didn't seem to impact wakunode2 because we call `rpc_subscribe` after the node "has settled".

This also means the default subscribe mechanism appears to be faulty in here https://github.com/status-im/nim-waku/commit/03895ffaf559c9c3f8f3cc360dd9af746a3d9ed8 - issue needed to fix this.

It'd be useful to fix FloodSub https://github.com/status-im/nim-waku/issues/161 since this would make it easier to see the exact problem.

It'd also be useful to cut out a testing issue and pr https://github.com/status-im/nim-waku/issues/158#issuecomment-707500402 to ensure we capture this again

Finally, it'd be great if this timing issue wasn't a concern and one could call subscribe at any time. I've not looked into the rebalance mesh semantics enough to know if this is a bug or something that could easily be addressed. Perhaps @dryajov or @sinkingsugar knows the answer here?

## Changelog

- Support static node in chat2
- Ensure nodes are connected before subscribing
- Sleep to allow settling down before subscribing
- Default chat2 log to DEBUG

# Testing done

Spin up three chat2 nodes, A, B, C with B and C specifying A as staticnode. Notice how subscribe action takes a few second and the resulting grafting etc. Send a message from node B, and notice how A picks it up and forwards it to C as expected. Also notice that sendPeer len for A is 1 as opposed to 0, which it was before.

For wakunode2, the default behaviour was not changed. Using `rpc_subscribe` after the node has settled still works.

### DIFF
--- a/examples/v2/chat2.nim
+++ b/examples/v2/chat2.nim
@@ -66,6 +66,11 @@ proc dialPeer(c: Chat, address: string) {.async.} =
   discard await c.node.switch.dial(remotePeer, WakuRelayCodec)
   c.connected = true
 
+proc connectToNodes(c: Chat, nodes: openArray[string]) =
+  echo "Connecting to nodes"
+  for nodeId in nodes:
+    discard dialPeer(c, nodeId)
+
 proc publish(c: Chat, line: string) =
   let payload = cast[seq[byte]](line)
   let message = WakuMessage(payload: payload, contentTopic: DefaultContentTopic)
@@ -158,6 +163,11 @@ proc processInput(rfd: AsyncFD, rng: ref BrHmacDrbgContext) {.async.} =
   # waitFor vs await
   await node.start()
 
+  var chat = Chat(node: node, transp: transp, subscribed: true, connected: false, started: true)
+
+  if conf.staticnodes.len > 0:
+    connectToNodes(chat, conf.staticnodes)
+
   let peerInfo = node.peerInfo
   let listenStr = $peerInfo.addrs[0] & "/p2p/" & $peerInfo.peerId
   echo &"Listening on\n {listenStr}"
@@ -171,9 +181,10 @@ proc processInput(rfd: AsyncFD, rng: ref BrHmacDrbgContext) {.async.} =
     let payload = cast[string](message.payload)
     echo &"{payload}"
     info "Hit subscribe handler", topic=topic, payload=payload, contentTopic=message.contentTopic
-  await node.subscribe(topic, handler)
 
-  var chat = Chat(node: node, transp: transp, subscribed: true, connected: false, started: true)
+  # XXX Timing issue with subscribe, need to wait a bit to ensure GRAFT message is sent
+  await sleepAsync(5.seconds)
+  await node.subscribe(topic, handler)
 
   await chat.readWriteLoop()
   runForever()
--- a/waku/node/v2/wakunode2.nim
+++ b/waku/node/v2/wakunode2.nim
@@ -97,6 +97,8 @@ proc init*(T: type WakuNode, nodeKey: crypto.PrivateKey,
     filters: initTable[string, Filter]()
   )
 
+  # TODO This is _not_ safe. Subscribe should happen in start and after things settled down.
+  # Otherwise GRAFT message isn't sent to a node.
   for topic in topics:
     proc handler(topic: string, data: seq[byte]) {.async, gcsafe.} =
       debug "Hit handler", topic=topic, data=data
--- a/waku.nimble
+++ b/waku.nimble
@@ -80,4 +80,5 @@ task chat2, "Build example Waku v2 chat usage":
   let name = "chat2"
   # NOTE For debugging, set debug level. For chat usage we want minimal log
   # output to STDOUT. Can be fixed by redirecting logs to file (e.g.)
-  buildBinary name, "examples/v2/", "-d:chronicles_log_level=WARN"
+  #buildBinary name, "examples/v2/", "-d:chronicles_log_level=WARN"
+  buildBinary name, "examples/v2/", "-d:chronicles_log_level=DEBUG"


waku/node/v2/wakunode2.nim
```diff
@@ -97,6 +97,8 @@ proc init*(T: type WakuNode, nodeKey: crypto.PrivateKey,
     filters: initTable[string, Filter]()
   )
 
+  # TODO This is _not_ safe. Subscribe should happen in start and after things settled down.
```
kdeme, line 100: hmm, this doesn't seem to be doing anything? It just subscribes a handler that does nothing to WakuRelay? Can be removed I guess.
And then similar code but with an actual handler and on "waku" topic is subscribed in start, which will actually do the work it seems.

examples/v2/chat2.nim
```diff
@@ -171,9 +181,10 @@ proc processInput(rfd: AsyncFD, rng: ref BrHmacDrbgContext) {.async.} =
     let payload = cast[string](message.payload)
     echo &"{payload}"
     info "Hit subscribe handler", topic=topic, payload=payload, contentTopic=message.contentTopic
-  await node.subscribe(topic, handler)
 
-  var chat = Chat(node: node, transp: transp, subscribed: true, connected: false, started: true)
+  # XXX Timing issue with subscribe, need to wait a bit to ensure GRAFT message is sent
```
kdeme, line 185: Strange. Are you sure that is what is going wrong here? The subscribe is basically also done already in the `start()` higher up. By the way, can you see the confusion here by allowing both type of subscribes in the public api ?

waku/node/v2/wakunode2.nim
oskarth: Right it doesn't show up in the diff, the next section is:

```
    # XXX: Is using discard here fine? Not sure if we want init to be async?
    # Can also move this to the start proc, possibly wiser?
    discard result.subscribe(topic, handler)

```

examples/v2/chat2.nim
oskarth: Without this the scenario as tested doesn't work, with it it works. The problem is subscribe being done too early.

What is the confusion you see with subscribe in public API? This is a floodsub vs gossipsub thing, not anything to do with filter subscriptions.

examples/v2/chat2.nim
```diff
@@ -66,6 +66,11 @@ proc dialPeer(c: Chat, address: string) {.async.} =
   discard await c.node.switch.dial(remotePeer, WakuRelayCodec)
   c.connected = true
 
+proc connectToNodes(c: Chat, nodes: openArray[string]) =
```
decanus, line 69: @oskarth can we somehow make the one in node accessible. So we have some code reuse. One of the issues I had with debugging was that half the functions we had there were copied into here.

examples/v2/chat2.nim
```diff
@@ -171,9 +181,10 @@ proc processInput(rfd: AsyncFD, rng: ref BrHmacDrbgContext) {.async.} =
     let payload = cast[string](message.payload)
     echo &"{payload}"
     info "Hit subscribe handler", topic=topic, payload=payload, contentTopic=message.contentTopic
-  await node.subscribe(topic, handler)
 
-  var chat = Chat(node: node, transp: transp, subscribed: true, connected: false, started: true)
+  # XXX Timing issue with subscribe, need to wait a bit to ensure GRAFT message is sent
+  await sleepAsync(5.seconds)
```
decanus, line 186: @dryajov is there no more efficient way to get info from libp2p when everything was successful?

decanus (changes requested):

examples/v2/chat2.nim
oskarth: This is out of scope of this PR. This fixes a blocking issue. See https://github.com/status-im/nim-waku/issues/227

examples/v2/chat2.nim
oskarth: Not in this PR.

decanus (approved):

examples/v2/chat2.nim
oskarth: This is also application code (example chat), not library code. The goal is to support the critical path of Dingpu

Issue #221, opened 2020-10-13
Messages not being forwarded properly with a third node


oskarth:
# Problem

Sometimes recently, the chat2 example stopped working for multiple nodes.

Note that this is both before and after latest nim-libp2p bump (see open PR).

Topology:
- A, B, C nodes
- B and C connects to C

Scenario:
- C publishes message, A receives it, but not B
- B publishes message, A receives it, but not A
- A publishes message, B and C receives it

## Details

Using trace:

```
INF 2020-10-13 13:16:08.282+08:00 Hit subscribe handler                      tid=10767 topic=waku payload=blol contentTopic=dingpu                
TRC 2020-10-13 13:16:08.282+08:00 broadcasting messages to peers             topics="pubsub" tid=10767 peers=0 msg="(subscriptions: @[], messages: @[(fromPeer: \"16*hmkWy3\", data: \"0a04626c6f6c
...64696e677075\", seqno: \"0000000000000001\", topicIDs: \"@[\\\"waku\\\"]\", signature: \"\", key: \"\")], control: (ihave: @[], iwant: @[], graft: @[], prune: @[]))"
TRC 2020-10-13 13:16:08.282+08:00 forwared message to peers                  topics="gossipsub" tid=10767 peers=0 msgId=000000000000000116Uiu2HAkv8dhuLiN1FXML96iDVEUZVsauJJviZny1o2FjvhmkWy3 peer=
16*hmkWy3
```

Notice the `peers=0`. If A is sending a message directly, it says `peers=2`. So for some reason it doesn't know to send messages to B and C.

oskarth:
I looked a bit but didn't go too deep, nor tried bisecting etc

decanus:
I am currently looking at libp2p as to where this issue could arise, I haven't found anything productive yet. Will keep digging.

PR #228, opened 2020-10-14
fixes/wait-for-start

### LINKED ISSUE #227: Ensure subscribe happens after node has started and settled down
# Problem

See https://github.com/status-im/nim-waku/pull/226 for context.

Subscribe to default topics should happen after node has started and settled down. https://github.com/status-im/nim-waku/commit/03895ffaf559c9c3f8f3cc360dd9af746a3d9ed8

For now, doing this with a sleep 2-5s seems OK.

## Acceptance criteria

Scenario described in https://github.com/status-im/nim-waku/pull/226 and https://github.com/status-im/nim-waku/issues/221 works for wakunode2 _without_ calling rpc_subscribe.

Basically what https://github.com/status-im/nim-waku/pull/152 should've fixed (unclear if regression or untested, perhaps due to use of floodsub by default back then).

## Future enhancements

Instead of sleeping either:

- Make it wait for the right event (meshIsReady or whatever), or
- Ensure it can be done in any order (what vs how)

Neither is needed for now, but would be a more correct way of doing it.

decanus:
closes #227


tests/v2/test_wakunode.nim
```diff
@@ -187,3 +187,51 @@ procSuite "WakuNode":
       (await completionFut.withTimeout(5.seconds)) == true
     await node1.stop()
     await node2.stop()
+
+  asyncTest "Messages are correctly relayed":
```
oskarth, line 191: Just to make sure, does this test fail with the previous behaviour?

waku/node/v2/wakunode2.nim
```diff
@@ -97,17 +97,7 @@ proc init*(T: type WakuNode, nodeKey: crypto.PrivateKey,
...
-  for topic in topics:
-    proc handler(topic: string, data: seq[byte]) {.async, gcsafe.} =
-      debug "Hit handler", topic=topic, data=data
-
-    # XXX: Is using discard here fine? Not sure if we want init to be async?
-    # Can also move this to the start proc, possibly wiser?
-    discard result.subscribe(topic, handler)
-
-proc start*(node: WakuNode) {.async.} =
+proc start*(node: WakuNode, topics = newSeq[string]()) {.async.} =
```
oskarth, line 100: Hm so we are changing the signature of `start`. It has a default argument which is good. API docs still have to be updated.

Is this the best way to do this in general? It might be, just not sure. What do you think @kdeme? Perhaps we could do it like all other "started node actions" and do it right after start in main?

waku/node/v2/wakunode2.nim
```diff
@@ -300,9 +295,9 @@ when isMainModule:
       Port(uint16(conf.tcpPort) + conf.portsShift),
       Port(uint16(conf.udpPort) + conf.portsShift))
     node = WakuNode.init(conf.nodeKey, conf.libp2pAddress,
-      Port(uint16(conf.tcpPort) + conf.portsShift), extIp, extTcpPort, conf.topics.split(" "))
+      Port(uint16(conf.tcpPort) + conf.portsShift), extIp, extTcpPort)
 
-  waitFor node.start()
+  waitFor node.start(conf.topics.split(" "))
```
oskarth, line 300: E.g. instead do it here and do it after static node etc has been connected to.

oskarth: Speaking of, I suppose this how the code actually executes. First the connect to nodes is called, then async we are waiting until we subscribe. It'd be easier to read and be more separated in terms of concerns if it was there

waku/node/v2/wakunode2.nim
```diff
@@ -324,4 +319,8 @@ when isMainModule:
       startMetricsServer(conf.metricsServerAddress,
         Port(conf.metricsServerPort + conf.portsShift))
 
+    # XXX: Is using discard here fine? Not sure if we want init to be async?
+    # Can also move this to the start proc, possibly wiser?
```
oskarth, line 323: Huh? Why are you copy pasting a comment that is irrelevant? Very misleading.

waku/node/v2/wakunode2.nim
oskarth: The downside of doing it in main I suppose is that some applications will use it as a library and then it wouldn't be called automatically. Honestly I wonder if that isn't fine. The reason we are doing this in first place is to get nodes in cluster to subscribe to some topics by default so machine can do proper relaying.

@kdeme wdyt?

waku/node/v2/wakunode2.nim
```diff
@@ -324,4 +319,8 @@ when isMainModule:
...
+    await result.subscribe(topic, handler)
```
oskarth, line 324: This call in general confuses me, I don' know why we are calling subscribe on result and we are also calling relay subscribe above.

Have you ran this and looked at logs and made sure subscribe is called accurately? Are both needed? I can't see the handler here either.

oskarth (changes requested):

tests/v2/test_wakunode.nim
decanus: apparently it does not. I need to investigate why.

waku/node/v2/wakunode2.nim
decanus: @oskarth so the reason why I did it here is that we get this functionality elsewhere too, eg chat. Otherwise we need to reimplement the wait there too if it is something thats part of the main module.

decanus:
@oskarth tested with chat, seems to work fine so ready to merge.

waku/node/v2/wakunode2.nim
kdeme: I think it makes sense to do this here.
But it depends on how much flexibility we want/need. Do we really need to be able to subscribe to different PubSub topics at the same time? Meaning, having different "messaging networks" where we publish and subscribe to?

(I'm not really talking about partitioning the topics, this could be arranged "underneath" I believe)

waku/node/v2/wakunode2.nim
kdeme: Well, if we do want to have this done here, some helper call should be created. Realize also that a node will not necessarily start `WakuRelay` (but instead e.g. only WakuFilter), so some extra logic will eventually still be needed.

kdeme (approved):

tests/v2/test_wakunode.nim
oskarth: Did you figure this out? To make sure test works properly, it should definitely fail before this chunk was introduced.

(It might be that this requires setting up three full nodes for some reason, in which case the test case might be more involved and could be cut up into a separate issue and pr, per https://github.com/status-im/nim-waku/issues/158#issuecomment-707500402

waku/node/v2/wakunode2.nim
oskarth: >  Do we really need to be able to subscribe to different PubSub topics at the same time with different handlers?

Yes, I believe so. Basic example: node that supports Status app using a Status topic, and another WalletConnect that uses another topic (completely separate from Status Network).

waku/node/v2/wakunode2.nim
oskarth: @decanus please leave a comment when resolving, otherwise it isn't clear to me how something was resolved.

waku/node/v2/wakunode2.nim
```diff
@@ -129,13 +119,7 @@ proc start*(node: WakuNode) {.async.} =
...
   node.subscriptions.subscribe(WakuFilterCodec, node.wakuFilter.subscription())
 
-  proc relayHandler(topic: string, data: seq[byte]) {.async, gcsafe.} =
-    let msg = WakuMessage.init(data)
-    if msg.isOk():
-      node.filters.notify(msg.value(), "")
-      await node.subscriptions.notify(topic, msg.value())
-
-  await node.wakuRelay.subscribe("waku", relayHandler)
+  await sleepAsync(5.seconds)
```
oskarth, line 122: Comment on why we are sleeping and under what conditions it can be removed. No one knows what this is for otherwise.

waku/node/v2/wakunode2.nim
oskarth: Agree re helper function. And yes, the FilterRequest would be an enhancement to do as well.

The reason for this default change in the first place is to allow for "non human operated" nodes like cluster node. Without it it doesn't work. I think there's diminishing returns for providing defaults when there's an API that should be used in standard operation.

oskarth (commented):
See above, still some unresolved issues and comments

waku/node/v2/wakunode2.nim
oskarth: That said, I can see reasoning above and think we can probably do it here. In the case of multiple networks this can be probably be done with API and/or handler-of-handler and/or extended later as we see specific needs.

API docs still needs updating.

docs/api/v2/node.md
```diff
@@ -22,7 +22,7 @@ proc init*(T: type WakuNode, nodeKey: crypto.PrivateKey,
   ##
   ## Status: Implemented.
 
-proc start*(node: WakuNode) {.async.} =
+proc start*(node: WakuNode, topics = newSeq[string]()) {.async.} =
   ## Starts a created Waku Node.
```
oskarth, line 26: Explain what topics is.

docs/api/v2/node.md
```diff
@@ -22,7 +22,7 @@ proc init*(T: type WakuNode, nodeKey: crypto.PrivateKey,
   ##
   ## Status: Implemented.
 
-proc start*(node: WakuNode) {.async.} =
+proc start*(node: WakuNode, topics = newSeq[string]()) {.async.} =
```
oskarth, line 25: Explain what topics is below.

oskarth: As in, explain the semantics.

As a user, I want to be able to read the signature and docstring and know what it is doing without knowing anything else.

tests/v2/test_wakunode.nim
decanus: @oskarth I still haven't fully figured it out.  We are starting 3 nodes.

oskarth: I mean that this test should fail for the current code in master, and succeed with this code. Otherwise it doesn't test what we want.

(Out of process nodes don't seem necessary to me to reproduce behaviour, but they might be, in which case that could be a separate issue).

oskarth:
@decanus please make sure you respond to the outstanding items before asking for re-reviews, otherwise it is a bit of a waste of time

waku/node/v2/wakunode2.nim
oskarth: does the absence of a response here means you fixed it?

decanus: @oskarth this isn't really relevant anymore. now that we mountRelay the topic subscriptions are there.

tests/v2/test_wakunode.nim
decanus: @oskarth I think they are, when testing it with this code in master it seems to still pass. So I assume its an in / out of process thing, I wouldn't know what else the issue could be on why this test works in master.

oskarth: Ok, can we break that out into an issue as suggested here https://github.com/status-im/nim-waku/issues/158#issuecomment-707500402? That'd be the obvious way to improve testing, as that's a bug that bit us.

oskarth: When you ran the test on master, did you run it before this commit? https://github.com/status-im/nim-waku/commit/b2edfe8dda3e28d0e45f7ae7118c7e2def9c41af

oskarth (commented):
I don't understand how this PR addresses the problem it claims to do: https://github.com/status-im/nim-waku/issues/227

It isn't obvious to me how the acceptance criteria has been met.

Additionally, the test generally seems fine, but if it doesn't actually capture the problem we saw last week it isn't clear to me that it is the test that would be most useful?

waku/node/v2/wakunode2.nim
```diff
@@ -99,7 +99,6 @@ proc start*(node: WakuNode) {.async.} =
   let listenStr = $peerInfo.addrs[0] & "/p2p/" & $peerInfo.peerId
   ## XXX: this should be /ip4..., / stripped?
   info "Listening on", full = listenStr
-  await sleepAsync(5.seconds)
```
oskarth: Why was this removed?

oskarth:
Looking at this PR, a previous PR, and a DM from you this is what I think you are saying is that this PR already addresses https://github.com/status-im/nim-waku/pull/209/files#r508976416

First, if that's what you are claiming it should be obvious in the original PR description. Right now it requires _a lot_ of context to have an idea. And that is assuming my interpretation is even correct. It takes a lot of time and effort to figure this out when you should explain what you are doing in a PR and why.

Second, if the above is correct, can you confirm that the acceptance criteria in https://github.com/status-im/nim-waku/issues/227 "Scenario described in #226 and #221 works for wakunode2 without calling rpc_subscribe." wasn't met _before_ PR 209 was merged and that it was met _after_ it was merged (current master).

Third, ditto for the test. To make sure the test actually tests what we want (https://github.com/status-im/nim-waku/issues/158#issuecomment-707500402), it needs to be failing _before_ the fix was in place and then work after. So, based on above reasoning, before this https://github.com/status-im/nim-waku/pull/209/files#r508976416 it should fail and after it should work.

Assuming you can confirm that 2 and 3 are tested and accurate, we can merge this. Assuming 2 isn't accurate, this PR should address that. Assuming 2 is accurate and 3 isn't, then either the test is wrong in some way, or the testing involves more work. If that work seems like too much, then at a minimum it should be written up as an issue and added to the board (then we can merge this as partial testing).

tests/v2/test_wakunode.nim
decanus: @oskarth yes I did

decanus:
@oskarth so to address the mentioned issues:
  - I have added a description to the PR to be a little more encompassing of what the pull-request does.
  - Testing manually is a bit harder to do from pre and post that actual commit because we were testing with chat and that was resolved before that commit by adding a timeout in the chat app.
 - Weirdly enough the test works both before and after. I think this may be an issue with it being in process and would need to be resolved with a more complex test.

decanus:
Also I have found a little bit of an issue, being that the sleep is in the wrong place. WE should move this from mount relay to after when peers are dialed.

decanus:
@oskarth ok so the reason why the test worked when I was testing it in master is because of the sleep which was there, see (https://github.com/status-im/nim-waku/blob/6138ef59f56ca784e3f1c39f95305b42f9e872ab/tests/v2/test_wakunode.nim#L229) The real fix to the problem was sleeping after connecting to nodes, not after subscribing to the relay or anything like that. What I have done now is to do that in the `connectToNodes` function additionally I have stopped using `switch.dial` and used the wrapped function instead. I tested this end-to-end with the chat2 and it seems to work. I am happy with this code now if this explanation makes sense to you.

waku/node/wakubridge.nim
```diff
@@ -85,7 +85,7 @@ proc startWakuV2(config: WakuNodeConf): Future[WakuNode] {.async.} =
     waitFor mountRelay(node, config.topics.split(" "))
 
   if config.staticnodesv2.len > 0:
-    connectToNodes(node, config.staticnodesv2)
+    await connectToNodes(node, config.staticnodesv2)
```
oskarth, line 88: Why is this await vs waitFor in wakunode2?

waku/node/v2/wakunode2.nim
```diff
@@ -265,11 +263,24 @@ proc setFilterPeer*(n: WakuNode, address: string) =
...
+proc connectToNodes*(n: WakuNode, nodes: seq[string]) {.async.} =
   for nodeId in nodes:
     info "connectToNodes", node = nodeId
     # XXX: This seems...brittle
-    discard dialPeer(n, nodeId)
+    await dialPeer(n, nodeId)
+    
+  # This sleep is done to ensure we only subscribe when the protocol has been mounted.
+  # Meaning graft messages were received.
+  await sleepAsync(5.seconds)
```
oskarth, line 274: My guess would be we can lower this to 2s but we can also leave it as is for now

waku/node/v2/wakunode2.nim
```diff
@@ -265,11 +263,24 @@ proc setFilterPeer*(n: WakuNode, address: string) =
 
   n.wakuFilter.setPeer(remotePeer)
 
-proc connectToNodes*(n: WakuNode, nodes: openArray[string]) =
+proc connectToNodes*(n: WakuNode, nodes: seq[string]) {.async.} =
```
oskarth, line 266: What's the difference between when these two functions are used?

oskarth: It looks like one is calling on relay protocol specifically, maybe worth capturing that in the function name. Also not clear to me why there is a difference with PeerInfo and string here

oskarth (commented):
See above

oskarth:
> I have added a description to the PR to be a little more encompassing of what the pull-request does.

Thanks. (I suppose now that description is out of date as well, since this PR is changing behaviour of connection to nodes.)

> Testing manually is a bit harder to do from pre and post that actual commit because we were testing with chat and that was resolved before that commit by adding a timeout in the chat app.

There are two ways to test this:
1) Use wakunode2 with default topics and only publish without calling rpc subscribe script
2) Modify chat2 to remove that quickfix

> Weirdly enough the test works both before and after. I think this may be an issue with it being in process and would need to be resolved with a more complex test.

Would be good with issue for this as described above. (Though I believe this is outdated by comment below)

oskarth:
> The real fix to the problem was sleeping after connecting to nodes, not after subscribing to the relay or anything like that. What I have done now is to do that in the connectToNodes function additionally I have stopped using switch.dial and used the wrapped function instead. I tested this end-to-end with the chat2 and it seems to work. I am happy with this code now if this explanation makes sense to you.

Gotcha. Seems reasonable. So with this change, can you confirm test didn't pass before but now passes?

Once that's done and some of the comments above are resolved we are good to merge

oskarth (approved):
I'll go ahead and approve this though as it has been going through many rounds already. OK to merge assuming we confirm test didn't work before but now works.

Some of the comments would be nice to fix, but not a dealbreaker I suppose

decanus:
@oskarth can confirm the test did not pass before.

waku/node/v2/wakunode2.nim
decanus: Both are calling dial with the relay protocol, I made a second function called connectToNodes for when we use it from tests or when we already have PeerInfo vs just the stringified version of the peer address. Otherwise we would need to stringify all before passing it or parsing all before passing.

decanus:
@oskarth so the manual testing turned out to be relatively simple when the fix was actually correct.

waku/node/v2/wakunode2.nim
kdeme: I guess I don't understand the problem well enough, but it is strange that this sleep is required here (well, it shouldn't be required at all, but ok).
The comment doesn't make it any more clear to me, this part of the code is dialing peers. How is it related exactly to the subscribing?

kdeme: Ok I read this now:

> The real fix to the problem was sleeping after connecting to nodes, not after subscribing to the relay or anything like that. What I have done now is to do that in the connectToNodes function additionally I have stopped using switch.dial and used the wrapped function instead. I tested this end-to-end with the chat2 and it seems to work. I am happy with this code now if this explanation makes sense to you.

Still not clear to me what the originally problem is/was.
If we would like to merge this as is, can you update this sleep with a proper `TODO` comment explaining the problem and the current workaround by sleeping.

PR #230, opened 2020-10-15
enhancement/integrate-store-chat

decanus:


decanus:
@oskarth made it optional

examples/v2/chat2.nim
```diff
@@ -55,21 +55,23 @@ proc initAddress(T: type MultiAddress, str: string): T =
...
-# NOTE Dialing on WakuRelay specifically
-proc dialPeer(c: Chat, address: string) {.async.} =
+proc parsePeer(address: string): PeerInfo = 
   let multiAddr = MultiAddress.initAddress(address)
   let parts = address.split("/")
-  let remotePeer = PeerInfo.init(parts[^1], [multiAddr])
+  result = PeerInfo.init(parts[^1], [multiAddr])
 
-  echo &"dialing peer: {multiAddr}"
+# NOTE Dialing on WakuRelay specifically
```
oskarth, line 63: How come this was changed from string to peer?

examples/v2/chat2.nim
```diff
@@ -172,10 +176,24 @@ proc processInput(rfd: AsyncFD, rng: ref BrHmacDrbgContext) {.async.} =
   let listenStr = $peerInfo.addrs[0] & "/p2p/" & $peerInfo.peerId
   echo &"Listening on\n {listenStr}"
 
+  if conf.storenode != ""
```
oskarth, line 179: This is correct behavior, just like it is (or should be) in wakunode2 and https://github.com/status-im/nim-waku/blob/34a8af4cc11387078b046c37a984ea03bab90b56/examples/v2/chat2.nim#L172

docs/tutorial/dingpu.md
```diff
@@ -2,11 +2,13 @@
 
 ## Basic chat usage
 
+> The chat app requires that the remote peer connected to is supports the WakuStore protocol. For Dingpu this is already the case.
```
oskarth, line 5:
```suggestion
> If historical messaging is desired, the chat app requires that the remote peer specified in `storenode` option supports the WakuStore protocol. For the current cluster node deployed as part of Dingpu this is already the case.
```

oskarth (approved):

oskarth:
Approving assuming it has been tested for common scenarios.

decanus:
there seems to be an issue with the relay that was reintroduced which does not seem right to me as that fix was not really touched in this PR.

oskarth:
Hm weird, did you verify that it works with same logic in master? Perhaps it is indeterministic :<

decanus:
@oskarth ok works now with the staticnode argument and store.

PR #235, opened 2020-10-19
fix/content-type-int

### LINKED ISSUE #234: Waku v2 contentTopic type
Implement the spec contentType change: https://github.com/vacp2p/specs/pull/217

decanus:
closes #234

### DIFF
--- a/examples/v2/chat2.nim
+++ b/examples/v2/chat2.nim
@@ -33,7 +33,7 @@ const Help = """
 const DefaultTopic = "waku"
 
 const Dingpu = "dingpu".toBytes
-const DefaultContentTopic = ContentTopic(uint32.fromBytes([Dingpu[0], Dingpu[1], Dingpu[2], Dingpu[3]]))
+const DefaultContentTopic = ContentTopic(uint32.fromBytes(Dingpu))
 
 # XXX Connected is a bit annoying, because incoming connections don't trigger state change
 # Could poll connection pool or something here, I suppose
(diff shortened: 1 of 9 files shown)


examples/v2/chat2.nim
```diff
@@ -31,7 +31,7 @@ const Help = """
 """
 
 const DefaultTopic = "waku"
-const DefaultContentTopic = "dingpu"
+const DefaultContentTopic = ContentTopic(1)
```
oskarth, line 34: Can we make the constructor us "dingpu" in byte format? Instead of using 1 everywhere.

decanus: @oskarth yeah, was going to do that. Was just testing to see if everything works.

decanus:
Tested with Chat2

examples/v2/chat2.nim
```diff
@@ -31,7 +31,9 @@ const Help = """
 """
 
 const DefaultTopic = "waku"
-const DefaultContentTopic = "dingpu"
+
+const Dingpu = "dingpu".toBytes
+const DefaultContentTopic = ContentTopic(uint32.fromBytes([Dingpu[0], Dingpu[1], Dingpu[2], Dingpu[3]]))
```
oskarth, line 36: This seems weird and manual to me. Why not do something like `cast[seq[byte](...)`?

decanus: This currently seems like the most decent way, the fromBytes function needs a fixed sized array that is the length of the int.

kdeme: You should be able to pass just `Dingpu`, as long as it contains 4 bytes that should work. Or you can take the 0..4 slice to be explicit I guess.

oskarth (approved):
lgtm

jm-clius (approved):
Will also merge these into my Unsubscribe-from-contentTopic changes.

PR #238, opened 2020-10-20
Wakuv1v2 process

kdeme:
First steps of having a bridge, application with the two nodes, resolves https://github.com/status-im/nim-waku/issues/188.

There are two remaining issues:
- Multiple ports to map has some problems: https://github.com/status-im/nim-waku/issues/239
- Waku rpc conflicts: https://github.com/status-im/nim-waku/issues/240

To be solved in other PR's and out of the scope here.

Code has been tested with another regular wakunode2 for v2 and with the Status test fleet for v1.


waku/node/config.nim
```diff
@@ -0,0 +1,170 @@
+import
```
oskarth, line 1: Is this a bridge specific config? If so we might want to make it explicit (e.g. bridge folder or bridge in name). If it is a general config, how does this relate to v1/config and v2/config?

oskarth (commented):
Oh nice!

Could we also add a basic instruction of how to run this in the Dingpu docs? Copy paste style.

waku/node/config.nim
kdeme: Yes it is specific for the bridge application, so it's similar as options for v1 and v2 but I only kept the options that make sense for a bridge (e.g. no light node settings for v1). Thought about putting it all in a subfolder but there are already so many folders :). I think I'll rename it.

kdeme:
Add some very brief documentation to dingpu docs.

oskarth (approved):

decanus (approved):

Issue #85, opened 2020-07-28
Waku v2 Node API: Unsubscribe from content filter


oskarth:
# Problem

Currently there is no way to unsubscribe from a content filter.

This may require changes to the filter spec.

# Acceptance criteria

According to docs https://github.com/status-im/nim-waku/blob/master/docs/api/v2/node.md

No longer receive messages on content topic.

# Related
 - https://github.com/vacp2p/specs/issues/206

oskarth:
@decanus is this possible with the current filter protocol or would this require a spec change? Not obvious to me how this would be done.

decanus:
Not possible right now, would require a spec change.

oskarth:
@jm-clius can this be closed?

PR #241, opened 2020-10-21
Feature/unsubscribe from topic

### LINKED ISSUE #84: Waku v2 Node API: Unsubscribe from topic
# Problem

Unable to unsubscribe from a pubsub topic.

# Notes

I believe this is possible with the current relay spec, due to PubSub behaviour. Should be quite straightforward.

## Acceptance criteria

According to docs https://github.com/status-im/nim-waku/blob/master/docs/api/v2/node.md

Don't receive messages on pubsub topic.

jm-clius:
closes #84 

This PR:
- adds "unsubscribeAll"-method to waku_relay to mirror PubSub functionality
- modifies waku_node2 API to mirror underlying PubSub functionality:
  - adds "unsubscribeAll"-method for a PubSub topic
  - modifies "unsubscribe" signature to include handler as parameter
- updates API documentation to reflect above changes

### DIFF
--- a/docs/api/v2/node.md
+++ b/docs/api/v2/node.md
@@ -40,11 +40,15 @@ proc subscribe*(node: WakuNode, request: FilterRequest, handler: ContentFilterHa
   ##
   ## Status: Implemented.
 
-proc unsubscribe*(w: WakuNode, topic: Topic) =
-  ## Unsubscribe from a topic.
+proc unsubscribe*(node: WakuNode, topic: Topic, handler: TopicHandler) {.async.} =
+  ## Unsubscribes a handler from a PubSub topic.
   ##
-  ## Status: Not yet implemented.
-  ## TODO Implement.
+  ## Status: Implemented.
+
+proc unsubscribeAll*(node: WakuNode, topic: Topic) {.async.} =
+  ## Unsubscribes all handlers registered on a specific PubSub topic.
+  ##
+  ## Status: Implemented.
 
 proc unsubscribe*(w: WakuNode, contentFilter: ContentFilter) =
   ## Unsubscribe from a content filter.
--- a/waku/protocol/v2/waku_relay.nim
+++ b/waku/protocol/v2/waku_relay.nim
@@ -74,6 +74,14 @@ method unsubscribe*(w: WakuRelay,
   else:
     await procCall FloodSub(w).unsubscribe(topics)
 
+method unsubscribeAll*(w: WakuRelay,
+                       pubSubTopic: string) {.async.} =
+  debug "unsubscribeAll"
+  if w.gossipEnabled:
+    await procCall GossipSub(w).unsubscribeAll(pubSubTopic)
+  else:
+    await procCall FloodSub(w).unsubscribeAll(pubSubTopic)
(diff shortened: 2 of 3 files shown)


decanus (changes requested):
Where is the part for the filters?

jm-clius:
> Where is the part for the filters?

Hi @decanus, this PR only addresses issue #84. The filters are part of #85 which I'm working on in parallel. Unless I'm missing something?

decanus:
AH ok

decanus (changes requested):

waku/protocol/v2/waku_relay.nim
```diff
@@ -74,6 +74,14 @@ method unsubscribe*(w: WakuRelay,
   else:
     await procCall FloodSub(w).unsubscribe(topics)
 
+method unsubscribeAll*(w: WakuRelay,
+                    pubSubTopic: string) {.async.} =
```
oskarth, line 78: Formatting, either indent to same column or bring up on same row

docs/api/v2/node.md
```diff
@@ -40,11 +40,15 @@ proc subscribe*(node: WakuNode, request: FilterRequest, handler: ContentFilterHa
   ##
   ## Status: Implemented.
 
-proc unsubscribe*(w: WakuNode, topic: Topic) =
-  ## Unsubscribe from a topic.
+proc unsubscribe*(w: WakuNode, topic: Topic, handler: TopicHandler) {.async.} =
```
oskarth, line 43: We started calling WakuNode node instead of w to be more explicit and avoid ambiguity with other things that could be called w. It hasn't been done uniform for all calls yet, but would be good to fix now at least for the unsubscribe calls.

oskarth (approved):
LGTM! Thanks for the PR description and updating of docs. Just some small nitpicks.

As for unsubscribe vs unsubscribeAll, I'm OK with both. There are arguments for both having it being transparent and using same function name with overloading. Assuming we go down this route and use the transparent API approach, we should probably be consistent and also do the same for (un)subscribing to content filters, calling them something like (un)subscribeContent or so?

kdeme (approved):
I don't have any strong opinions on the API naming.
But considering libp2p goes for unsubscribeAll, it might be better to remain consistent with that.

oskarth:
@decanus I think you forgot to review this one again? Both Kim and I are fine to merge this. Do you still have a strong opinion on changing it?

decanus (approved):

docs/api/v2/node.md
jm-clius: Fixed in 4979718

waku/protocol/v2/waku_relay.nim
jm-clius: Fixed in 4979718

oskarth (approved):

oskarth:
>  The base branch requires all commits to be signed. Learn more about signing commits. 

@jm-clius I think you need to setup commit signing https://docs.github.com/en/free-pro-team@latest/github/authenticating-to-github/signing-commits

PR #242, opened 2020-10-22
Cleanup the wakunode2 config

kdeme:
Not much to see here, just removal of some unused config code.
Wanted to add some better input parsing on staticnodes, etc. also, but that will give too much conflict with https://github.com/status-im/nim-waku/pull/228 , I'll do that later.

### DIFF
--- a/docs/tutorial/dingpu.md
+++ b/docs/tutorial/dingpu.md
@@ -32,10 +32,10 @@ To just run a node and not interact on the chat it is enough to run `wakunode2`:
 ./build/wakunode2 --staticnode:<multiaddr>
 ```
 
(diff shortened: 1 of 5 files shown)


oskarth:
One thing I had as a TODO to write up is the difference between `staticnode` and `ports-shift`. The inconsistency is annoying IMO. Perhaps we could make it consistent, or make sure both forms are allowed? Might be a good time to do it in this PR

oskarth (approved):

kdeme:
I noticed your comment after I merged. Yeah, probably should be consistent there. I'll make that change in a later PR.

PR #244, opened 2020-10-26
feature/add-filter

decanus:
Adds the filter protocol to the chat2 example. This was tested locally. If a filternode is set, the relay does not subscribe to any topics.

### DIFF
--- a/examples/v2/chat2.nim
+++ b/examples/v2/chat2.nim
@@ -182,6 +182,18 @@ proc processInput(rfd: AsyncFD, rng: ref BrHmacDrbgContext) {.async.} =
 
     await node.query(HistoryQuery(topics: @[DefaultContentTopic]), storeHandler)
 
+  if conf.filternode != "":
+    node.mountFilter()
+
+    node.wakuFilter.setPeer(parsePeer(conf.filternode))
+
+    proc filterHandler(msg: WakuMessage) {.gcsafe.} =
+      let payload = cast[string](msg.payload)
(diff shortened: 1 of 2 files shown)


examples/v2/chat2.nim
```diff
@@ -158,7 +158,11 @@ proc processInput(rfd: AsyncFD, rng: ref BrHmacDrbgContext) {.async.} =
 
   # waitFor vs await
   await node.start()
-  await node.mountRelay(conf.topics.split(" "))
+
+  if conf.filternode != "":
```
oskarth, line 162: Maybe log which one is being used to make it explicit?

oskarth (approved):
Seems reasonable, does it work? And the node doesn't subscribe to other messages?

jm-clius (approved):

oskarth:
@decanus how come this is in draft?

decanus:
@oskarth forgot to move to ready cause I was still testing.

examples/v2/chat2.nim
kdeme, line 162: But wouldn't it require a node to also be able to run Relay and Filter, so that it can actual push messages from the Relay network to a Filter only node?

decanus: We are starting a relay, we just aren't subscribing, so we can push but aren't receiving

oskarth: you like reviewing PRs after they are merged @kdeme! :D

PR #247, opened 2020-10-28
feature/persistence

### LINKED ISSUE #245: Production-ready: Persistence of store messages
# Problem

As a node runner, I want to be able to restart a node and still keep state of historical messages.

As a user, I want historical messages to work better for HA nodes that restart.

# Notes

Requires basic integration with sqlite, a la what Desktop has done.

Out of scope:
- Efficient indices and optimizations

Suggested approach:
- Either do a basic append-only log with JSON/BLOB which reads into memory on startup and then appends each new state change, or
- Do more traditional interactive INSERT/LOOKUP for each item etc (probably a bit more work)

cc @decanus I believe this is needed at a basic level for getting nim-waku production ready for Status.

# Further issues

Look into how the existing mailserver production setup works, and make sure this can be re-used in some form.

decanus:
closes #245 

This adds SQLite, the area to store the db can be determined with the `dbpath` param.


tests/v2/test_wakunode.nim
```diff
@@ -139,9 +139,9 @@ procSuite "WakuNode":
     var completionFut = newFuture[bool]()
 
     await node1.start()
-    node1.mountStore()
+    node1.mountStore("")
```
oskarth, line 142: Is there a reason this can't be default argument? Passing empty string kind of ugly

waku/node/config_bridge.nim
```diff
@@ -137,6 +137,11 @@ type
       defaultValue: ""
       name: "filternode" }: string
 
+    dbpath* {.
```
oskarth, line 140: Might want to make it explicit this is DB for Waku v2

waku/node/v2/wakunode2.nim
```diff
@@ -206,9 +207,10 @@ proc mountFilter*(node: WakuNode) =
   node.switch.mount(node.wakuFilter)
   node.subscriptions.subscribe(WakuFilterCodec, node.wakuFilter.subscription())
 
-proc mountStore*(node: WakuNode) =
+proc mountStore*(node: WakuNode, path: string) =
```
oskarth, line 210: default arg empty string

waku/node/v2/wakunode2.nim
```diff
@@ -206,9 +207,10 @@ proc mountFilter*(node: WakuNode) =
...
+  var db = db_sqlite.open(path, "", "", "")
```
oskarth, line 211: error handling if it can't find db

waku/protocol/v2/waku_store.nim
```diff
@@ -78,11 +80,18 @@ proc encode*(rpc: HistoryRPC): ProtoBuffer =
...
-  for msg in w.messages:
-    if msg.contentTopic in query.topics:
-      result.messages.insert(msg)
+
+  let topics = join(query.topics, ", ")
+  for row in w.db.rows(SqlQuery("SELECT contentTopic, payload FROM messages WHERE contentTopic IN (" & topics & ")")):
+    result.messages.add(WakuMessage(payload: cast[seq[byte]](row[1]), contentTopic: ContentTopic(row[0].parseInt)))
 
 method init*(ws: WakuStore) =
+  ws.db.exec(sql"""CREATE TABLE IF NOT EXISTS messages (
```
oskarth, line 89: I'd add a comment explaining semantics briefly. Aslo might want to break out sql statement into constant.

waku/protocol/v2/waku_store.nim
```diff
@@ -78,11 +80,18 @@ proc encode*(rpc: HistoryRPC): ProtoBuffer =
...
+                 topic TEXT NOT NULL,
+                 contentTopic   INTEGER,
```
oskarth, line 91: Why is contentTopic an integer?

waku/protocol/v2/waku_store.nim
oskarth, line 89: also error handling

waku/protocol/v2/waku_store.nim
```diff
@@ -115,7 +125,10 @@ proc subscription*(proto: WakuStore): MessageNotificationSubscription =
   ## the filter should be used by the component that receives
   ## new messages.
   proc handle(topic: string, msg: WakuMessage) {.async.} =
-    proto.messages.add(msg)
+    proto.db.exec(
```
oskarth, line 128: error handling

oskarth (commented):
See in-line comments

Could you update docs with how running with and without db works?

waku/node/v2/wakunode2.nim
decanus: Why would there be error handling on that? On first time it just creates it. Would one need to explicitly create the db beforehand?

waku/protocol/v2/waku_store.nim
decanus: Because we changed it to a uint32 in the protobuf.

waku/node/v2/wakunode2.nim
oskarth: I don't know the API, but I would be surprised if no errors can occur here for opening a file. Permissions, no file, folder, busy resource, etc.

Worst case it is an error path that doesnt get hit.

waku/protocol/v2/waku_store.nim
```diff
@@ -115,7 +138,13 @@ proc subscription*(proto: WakuStore): MessageNotificationSubscription =
...
   ## new messages.
   proc handle(topic: string, msg: WakuMessage) {.async.} =
-    proto.messages.add(msg)
+    try:
+      proto.db.exec(
+        sql"INSERT INTO messages (topic, contentTopic, payload) VALUES (?, ?, ?)",
+        topic, msg.contentTopic, cast[string](msg.payload)
+      )
+    except:
+      warn "dropped message due to sqlite error: " & getCurrentExceptionMsg()
```
decanus, line 147: @kdeme is there a nicer way to do that?

michaelsbradleyjr (commented):
I don't have any comments on the changes to the Nim sources *per se*. But two thoughts:

**1.** An alternative to the built-in `db_sqlite` module is [`tiny_sqlite`](https://github.com/GULPF/tiny_sqlite). The bullet points at the top of the README explain the differences. One of the differences might be of particular interest:

> represents database values with a typesafe case object called `DbValue` instead of treating every value as a string, which among other things means that SQLite `NULL` values can be properly supported.

In the *Usage* section of the README you can see the `NULL` handling and also that there isn't a need to e.g. cast from string to integer.

**2.** The [`nim-sqlcipher`](https://github.com/status-im/nim-sqlcipher) library that I've been working on for use in `nim-status`, originally spearheaded by @richard-ramos, is inspired by `tiny_sqlite` and it may be interesting to consider its integration in `nim-waku`. However, at the present time there are some problems building it when using shared libraries instead of static linking, there aren't any docs yet, and I'm not 100% happy with the Makefile+nimterop approach I've hacked together though I'm short on ideas for a different and better way to go about it. Long story short, it's not yet ready for general usage but it's getting there and I'd very much welcome input from others interested in using it.

waku/node/v2/waku_types.nim
```diff
@@ -70,7 +71,7 @@ type
     switch*: Switch
     rng*: ref BrHmacDrbgContext
     peers*: seq[HistoryPeer]
-    messages*: seq[WakuMessage]
```
oskarth, line 73: Why is messages removed here?

1) What if I don't want to use a DB
2) Wouldn't there be a "cache" anyway for easy access?

waku/node/v2/wakunode2.nim
```diff
@@ -342,7 +360,10 @@ when isMainModule:
   waitFor node.start()
 
   if conf.store:
-    mountStore(node)
+    let res = mountStore(node, conf.dbpath)
+    if res.isErr:
+      info "failed to mount store"
```
oskarth, line 365: error or warn

waku/node/v2/wakunode2.nim
```diff
@@ -206,12 +211,25 @@ proc mountFilter*(node: WakuNode) =
   node.switch.mount(node.wakuFilter)
   node.subscriptions.subscribe(WakuFilterCodec, node.wakuFilter.subscription())
 
-proc mountStore*(node: WakuNode) =
+proc mountStore*(node: WakuNode, path = ""): Result[void, WakuNodeError] =
+  var db: DbConn
+  try:
```
oskarth, line 216: Is this the semantics we want? What happens if I don't specify or want a db?

Alternatively, we could make this a WARN/ERR and keep operating. Not sure what would be best?

cc @kdeme

waku/protocol/v2/waku_store.nim
```diff
@@ -78,11 +91,17 @@ proc encode*(rpc: HistoryRPC): ProtoBuffer =
...
-  for msg in w.messages:
-    if msg.contentTopic in query.topics:
-      result.messages.insert(msg)
 
-method init*(ws: WakuStore) =
+  let topics = join(query.topics, ", ")
+  for row in w.db.rows(SqlQuery("SELECT contentTopic, payload FROM messages WHERE contentTopic IN (" & topics & ")")):
+    result.messages.add(WakuMessage(payload: cast[seq[byte]](row[1]), contentTopic: ContentTopic(row[0].parseInt)))
+
+method init*(ws: WakuStore): Result[void, StoreError] =
```
oskarth, line 99: I feel like this should be conditional on there being a DB backend. The default should be the same IMO with in-memory

waku/node/v2/wakunode2.nim
kdeme: It would probably be better if the database part and the protocol mounting part has a better separation.
And yes, default it could use a memory db perhaps, and you could provide it with a persistent db instead, so you'd need some abstraction.

waku/node/v2/wakunode2.nim
```diff
@@ -206,12 +211,25 @@ proc mountFilter*(node: WakuNode) =
...
+    db = db_sqlite.open(path, "", "", "")
+  except:
+    return err(FailedToOpenDatabase)
+  
   info "mounting store"
-  node.wakuStore = WakuStore.init(node.switch, node.rng)
+
+  let store = WakuStore.init(node.switch, node.rng, db)
```
kdeme, line 223: What if I'm a node that only wants to request data through the store protocol, but not actually store it?

waku/protocol/v2/waku_store.nim
kdeme: I don't know this db code so I don't know what errors can occur here still (and how fatal they are or might be). 
So it would be better to define properly which errors you are catching here, e.g.:

```Nim
except ThisTypeOfError as e:
warn "dropped message due to sqlite error", error = e.msg
```
But I guess you can opt for just `CatchableError` too (I prefer this typically less, as it usually means that nobody properly looked at what can go wrong, but it depends on the case). Definitely don't use only `except` without mentioning a type as this will catch all Exceptions including Defects.

kdeme (commented):
I think in general the store protocol and the actual storing of data (whether in memory or persistent) should be better separated.

arnetheduck:
what we use in nimbus is https://github.com/status-im/nim-eth/blob/master/eth/db/kvstore_sqlite3.nim, which in turn uses https://github.com/arnetheduck/nim-sqlite3-abi which automatically includes a working and correct version of sqlite3 as part of the build system, on all platforms (windows, linux etc) - we tend to avoid the `std` modules because they often leak memory and have other correctness issues, including crashing randomly when the wrong dll version is used etc. 

@michaelsbradleyjr it would probably make sense to somewhat unify the API's of sqlcipher and kvstore_sqlite3 / sqlite3_abi - they are the same mostly - when writing wrappers, it pays off to separate the ABI part (the "raw" C function definitions) from fluff on top that makes it pleasant to use from nim - in the case of sqlcipher/sqlite-abi, the fluff-on-top part would ideally be the same since sqlcipher exposes the same ABI

arnetheduck:
one thing about waku storage is that it can probably be expressed as a key-value store in general - it might make sense as well to use  the abstract kvstore interface in nim-eth if that's sufficient (then it's easy to switch to for example rocksdb if needs grow)

waku/node/v2/waku_types.nim
decanus: 1. If you don't you don't pass a path and sqlite is only used in memory.
2. Hmm not sure. Going off of status-go it didn't really do that either.

decanus:
@michaelsbradleyjr so I am going to opt for not using sqlcipher right now due to the fact that as you said its relatively unstable. I think we can move over in the future once @arnetheduck's comments have been taken into account.

waku/protocol/v2/waku_store.nim
decanus: i mean when no dbpath is passed the database is *in memory*

waku/node/v2/waku_types.nim
oskarth: What if I don't have SQLite and dont want it or can't use it for some reason? To me it should still work with in memory naive data structure.  No reason to remove this capability.

waku/node/v2/wakunode2.nim
oskarth: Exactly, it should be a choice which backend is used and this can be abstracted away.

This is also needed for tests and mocking.

arnetheduck: sqlite supports in-memory databases - it's an easy way to not introduce lots of unnecessary abstraction

waku/node/v2/waku_types.nim
arnetheduck: don't have sqlite? it's included in the binary, assuming you start using nim-sqlite3-abi instead of the std library sqlite module (which can be assumed to be broken)

waku/node/v2/wakunode2.nim
kdeme: In memory or not, doesn't matter. As a light node just requesting messages, I don't want this code to run.

oskarth:
@decanus What's the state of this?

decanus:
@oskarth I am having various issue which I am trying to solve with the new sqlite ABI recommended by @arnetheduck.

oskarth:
Can you explain in more detail what you are trying to do and what problems you are running into?

decanus:
@oskarth Yeah so the problem I am facing right now, I have no idea how to select multiple rows from the database. The reason is the KVStore example in nim-eth uses a callback but I don't think its possible to know when all rows are returned.

oskarth:
I don't understand, can't you just call it like the following test? https://github.com/status-im/nim-eth/blob/master/tests/db/test_kvstore_sqlite3.nim#L54-L65

decanus:
@oskarth ptal, managed to make things work over sqlite now. Next step is to reimplemnt @staheri14 work in sql as it seems more efficient to do it that way vs iterating through arrays with lots of code all the time.

decanus:
@kdeme not that this is done fully, but now the DB is optional, how do you feel about this solution?

waku/node/v2/message_store.nim
```diff
@@ -0,0 +1,211 @@
...
+  elif val is int32:
+    sqlite3_bind_int(s, n.cint, val)
+  elif val is uint32:
+    sqlite3_bind_int(s, int(n).cint, int(val).cint)
+  elif val is int64:
+    sqlite3_bind_int64(s, n.cint, val)
+  else:
+    {.fatal: "Please add support for the 'kek' type".}
+
+proc put*(db: MessageStore, message: WakuMessage): MessageStoreResult[void] =
```
staheri14, line 148: You may also consider creating timestamp for each message in the database

waku/node/v2/message_store.nim
```diff
@@ -0,0 +1,211 @@
...
+      "DESC"
+
+  var pageSize = paging.pageSize
+  if pageSize == 0:
+    pageSize = MaxPageSize
+
+  let query = """
+    SELECT contentTopic, payload FROM messages
+    WHERE contentTopic IN (""" & join(topics, ", ") & """)
+    ORDER BY id """ & direction & """ LIMIT """ & $pageSize
```
staheri14, line 186: Following my prior comment, then messages shall be ordered by timestamp, then by id

waku/node/v2/message_store.nim
```diff
@@ -0,0 +1,211 @@
...
+proc get*(db: MessageStore, topics: seq[ContentTopic], paging: PagingInfo, onData: DataProc): MessageStoreResult[bool] =
+  let direction =
+    case paging.direction
+    of FORWARD:
+      "ASC"
+    of BACKWARD:
+      "DESC"
+
+  var pageSize = paging.pageSize
+  if pageSize == 0:
```
staheri14, line 180: In the pagination specs, pageSize being zero means no pagination is needed, so no need to stick to the maxPageSize, all the messages can be returned

staheri14: Also, in case pageSize is larger than MaxPageSize, it should be adjusted to MaxPageSize

waku/node/v2/message_store.nim
staheri14: If direction== FORWARD, messages with ids larger than the `paging.cursor` must be returned, otherwise, smaller than the `paging.cursor`

waku/node/v2/message_store.nim
staheri14: Also,  the `pageSize` must be set to the `MaxPageSize`, in case that it is larger than the `MaxPageSize`

decanus: ok perfect, will make that change

waku/node/v2/message_store.nim
```diff
@@ -0,0 +1,211 @@
...
+  discard sqlite3_clear_bindings(s) # no errors possible
+
+  res
+
+proc close*(db: MessageStore) =
+  discard sqlite3_close(db.env)
+
+  db[] = MessageStore()[]
+
+proc get*(db: MessageStore, topics: seq[ContentTopic], paging: PagingInfo, onData: DataProc): MessageStoreResult[bool] =
```
staheri14, line 171: You will probably need to return a PaginInfo as part of this proc, and its elements must be set according to the pagination specs
its direction the same as the input paging
its pageSize is the length of the list of retrieved waku messages
its cursor is either the highest index (if direction is FORWARD) or the lowest index (if direction is BACKWARD) in the retrieved set

staheri14: I keep using index, it may be confusing, by index I mean the combination of timestamp and id. While comparing timestamp+id s, the timestamp is the most significant part

waku/node/v2/message_store.nim
staheri14: If I am not mistaken, 2 query messages with differing WHERE clauses (corresponding to 2 different directions) would be needed at this point

waku/node/v2/message_store.nim
oskarth, line 148: docstring

waku/node/v2/waku_types.nim
```diff
@@ -188,14 +196,11 @@ proc generateRequestId*(rng: ref BrHmacDrbgContext): string =
   brHmacDrbgGenerate(rng[], bytes)
   toHex(bytes)
 
-proc computeIndex*(msg: WakuMessage): Index =
-  ## Takes a WakuMessage and returns its Index 
+proc id*(msg: WakuMessage): MDigest[256] =
```
oskarth, line 199: Why this change?

waku/node/v2/message_store.nim
decanus: We are

waku/node/v2/message_store.nim
decanus: @staheri14 can you have another look

tests/v2/test_waku_pagination.nim
```diff
@@ -1,217 +0,0 @@
-{.used.}
```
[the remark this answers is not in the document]
decanus: @oskarth this was testing pagination implementations in nim which have been moved to SQL

waku/node/v2/message_store.nim
```diff
@@ -0,0 +1,216 @@
...
+    limit = "LIMIT " & $paging.pageSize
+  if paging.pageSize > MaxPageSize:
+    limit = "LIMIT " & $MaxPageSize
+
+  let direction =
+    case paging.direction
+    of FORWARD:
+      "timestamp > " & $paging.cursor.receivedTime & " AND id > ?"
+    of BACKWARD:
+      "timestamp < " & $paging.cursor.receivedTime & " AND id < ?"
```
staheri14, line 185: The combination of $paging.cursor.receivedTime + paging.cursor.digest must be compared but not each of them individually.  Maybe, its better to find the message corresponding to $paging.cursor.receivedTime and  paging.cursor.digest in the sorted database , and then apply the `limit` depending on the direction

staheri14: btw, if a message corresponding to  $paging.cursor.receivedTime + paging.cursor.digest cannot be found in the database, it is an invalid cursor, an no waku message shall be returned

staheri14: Also, if `pageSize` is not zero AND the cursor has the default value i.e., `paging.cursor.receivedTime=0` and `paging.cursor.digest `is an array of zero bytes, this corresponds to an initial paging request. If `direction` is `FORWARD`, slice from the beginning of the database, and if `BACKWARD` slice from the end of the database.

waku/protocol/v2/waku_store.nim
```diff
@@ -173,108 +177,23 @@ proc encode*(rpc: HistoryRPC): ProtoBuffer =
...
+  var messages = newSeq[WakuMessage]()
+  proc data(val: WakuMessage) =
+    messages.add(val)
 
-method init*(ws: WakuStore) =
+  let res = w.store.get(query.topics, query.pagingInfo, data)
+  if res.isErr:
+    return
+
+  result.messages = messages
```
staheri14, line 194: The `HistoryResponse` should also contain the `pagingInfo` field

waku/node/v2/message_store.nim
staheri14: > You will probably need to return a PaginInfo as part of this proc, and its elements must be set according to the pagination specs
> its `direction` the same as the input `paging`
> its `pageSize` is the length of the list of retrieved waku messages
> its `cursor` is either the highest index (if direction is FORWARD) or the lowest index (if direction is BACKWARD) of the retrieved set

To clarify my comment, the `cursor` can be set as follows:
if direction == FORWARD then `cursor.digest `and `cursor.receivedTime` must be set to the `id` and `timestamp` of the last message in the retrieved page 
if direction == BACKWARD then `cursor.digest `and `cursor.receivedTime` must be set to the `id` and `timestamp` of the first message in the retrieved page

waku/node/v2/message_store.nim
decanus: @staheri14 whats the compelling rationale for having implemented it like that in the first place?

staheri14: The value of `pageSize` is used as the indicator of whether the pagination should take place or not. We could alternatively define a new boolean field in the `HistoryQuery` to capture the same thing (true for pagination and false for no pagination), though we can infer the same thing by just looking at the `pageSize` as well.
In fact, if the querying node is not aware of pagination at all and leaves the `pagingInfo` part of the query (hence the `pageSize`) unspecified, everything will work correctly (the queried node can understand that no pagination is required). 
This is how I implemented it in the in-memory version, feel free to follow a different approach :)

staheri14: Also, for the initial pagination request, the querying node does not have a valid `cursor`, that is why the queried node  performs somewhat a default pagination. Regarding the default pagination policy i.e., slicing messages from the beginning or the end of the database, the rationale was that when the requested paging `direction` is `FORWARD`, the user is willing to read messages from the beginning to the end, and vice versa when the direction is `BACKWARD`.

decanus: I meant why are we doing $paging.cursor.receivedTime + digest?

staheri14: I see! the full discussion is in this issue https://github.com/vacp2p/specs/issues/195, but the overall idea is that, nodes are interested in browsing messages based on the time (usually the most recent ones), that is why messages have timestamp to allow pagination by time. The hash of messages i.e., digests are added to break the tie between two messages that may happen to arrive simultaneously. 
I also think that having hash in the cursor will also allow content-addressing of messages (for the future cases). The querying node can use the hash part of the cursor (obtained  from one node) to resume its history message request by contacting another node.

Also as a side note, the fields of cursor correspond to all the dimensions based on which the pagination is done.

By the way, as the in-memory version is going to be replaced by the database (hence reworking the pagination), we may also modify the pagination method and  use a sender-generated timestamp instead (mentioned in https://github.com/vacp2p/specs/issues/219). What do you think @oskarth

decanus: Well what this implementation now has is `time` and `cursor` so wouldn't sort by time and then by cursor eg ID yield the result we wanted?

staheri14: Consider this example of two waku messages, 
w1, timetamp=2, digest=1

staheri14: If I understood your implementation correctly, then the following made-up example would clarify my point (w1 and w2 are waku messages)
W1, timestamp=2, hash=10
W2, timestamp=3, hash=9

W1 precedes w2 in the sorted database

Consider a cursor with the following fields: timestamp=2 and hash=10
If you search for messages whose hash AND timestamp are larger than the cursor then you will miss W2 because cursor.hash is not larger than w2.hash

decanus: Ah I see what you are saying, so instead of it being a WHERE clause we should find all messages greater than timestamp ordered by the hash?

staheri14: I think the hash values should never be treated alone, hashes always make sense in conjunction with timestamp. Maybe it is better to create a temporary column `new_col` from the concatenation of` timestamp` and `hash`, then sort based on `new_col`, and then use the WHERE clause as `WHERE new_col>cursor.timestamp||cursor.digest`

staheri14: Regarding my prior comment, we should make sure that there is a record for the queried cursor (to validate the cursor).

decanus: i don't think this actually makes sense

staheri14: You don't really have to create another column, the concatenation should happen as part of the sql query that you make

decanus: I'm saying the way it all works seems unintuitive

oskarth:
@decanus This PR has been open for more than two weeks now, and it seems to be increasing in scope, complexity and uncertainty. Considering this is just one of several issues in https://github.com/status-im/nim-waku/issues/256 I would suggest cutting the scope and uncertainty of this PR, and make sure to isolate specific problem areas to follow up PRs.

If a minimal approach is taken, then it wouldn't need to replace existing logic, bleed into spec changes that are unrelated, etc. It'd just be an additional capability, i.e. minimal persistence for nodes who desire it. I don't see a good reason for not following this approach.

waku/node/v2/message_store.nim
staheri14: Sorry I missed your last comment, I would be glad to clarify, let me know which parts are unintuitive?

waku/node/v2/message_store.nim
```diff
@@ -0,0 +1,266 @@
...
+          "ORDER BY timestamp ASC, id ASC"
+        of BACKWARD:
+          "ORDER BY timestamp DESC, id DESC"
+
+    let where =
+      case paging.direction
+        of FORWARD:
+          "AND timestamp > ?"
+        of BACKWARD:
+          "AND timestamp < ?"
```
staheri14, line 234: Haven't tested the following code, you may also try this
`        of FORWARD:
          "AND CONCAT(timestamp,id) > ?"
        of BACKWARD:
          "AND CONCAT(timestamp,id) < ?"`

waku/node/v2/message_store.nim
```diff
@@ -0,0 +1,266 @@
...
+
+    var query = "SELECT timestamp, contentTopic, payload FROM messages WHERE contentTopic IN (" & join(topics, ", ") & ") " & where & " " & order
+
+    if pageSize > 0:
+      query &= " LIMIT " & $pageSize
+
+    s = prepare(db.env, query): discard
+    checkErr bindParam(s, 1, int64(paging.cursor.receivedTime))
```
staheri14, line 242: Combined with my prior edit (i.e., the concatenation of `id` and `timestamp` columns), you may also try this (fromBytes is from `stew/byteutils`), though haven't tested it
`checkErr bindParam(s, 1, $int64(paging.cursor.receivedTime) & string.fromBytes(paging.cursor.digest))`

waku/node/v2/message_store.nim
```diff
@@ -0,0 +1,266 @@
...
+        of BACKWARD:
+          "ORDER BY timestamp DESC, id DESC"
+
+    var query = "SELECT timestamp, contentTopic, payload FROM messages WHERE contentTopic IN (" & join(topics, ", ") & ") " & order
+
+    if pageSize > 0:
+      query &= " LIMIT " & $pageSize
+
+    s = prepare(db.env, query): discard
+  else:
```
staheri14, line 221: please also check whether the paging.cursor is valid, i.e., whether there is a record in the database with `timestamp == paging.cursor.receivedTime`  and `id == paging.cursor.digest`

waku/node/v2/message_store.nim
```diff
@@ -0,0 +1,266 @@
...
+      case paging.direction
+        of FORWARD:
+          "AND timestamp > ?"
+        of BACKWARD:
+          "AND timestamp < ?"
+
+    var query = "SELECT timestamp, contentTopic, payload FROM messages WHERE contentTopic IN (" & join(topics, ", ") & ") " & where & " " & order
+
+    if pageSize > 0:
+      query &= " LIMIT " & $pageSize
```
staheri14, line 239: if `pageSize == 0`, no limit is needed (`pageSize` being zero means no pagination)

waku/node/v2/message_store.nim
decanus: I'll try this

waku/node/v2/message_store.nim
decanus: will try thanks.

waku/node/v2/message_store.nim
decanus: yeah, thats why we are checking if its greater than 0, otherwise we do not add the limit.

tests/v2/test_waku_pagination.nim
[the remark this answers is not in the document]
oskarth: Does that mean we also ported all the relevant tests?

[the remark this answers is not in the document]
oskarth: Does that mean we also ported all the relevant tests?

waku/node/v2/message_store.nim
```diff
@@ -0,0 +1,211 @@
...
+      discard sqlite3_finalize(journalModePragma)
+      return err($sqlite3_errstr(x))
+
+    if (let x = sqlite3_column_text(journalModePragma, 0);
+        x != "memory" and x != "wal"):
+      discard sqlite3_finalize(journalModePragma)
+      return err("Invalid pragma result: " & $x)
+
+  # TODO: check current version and implement schema versioning
+  checkExec "PRAGMA user_version = 1;"
```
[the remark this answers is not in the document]
oskarth: Why don't we use upstream? This still isn't answered.

If we are going to copy paste hundreds of lines of code, we better:

1) have a great reason
2) make the separation between custom and 'upstream' code clear
- ideally with a separate module or so

waku/node/v2/message_store.nim
```diff
@@ -0,0 +1,266 @@
...
+
+  let journalModePragma = prepare("PRAGMA journal_mode = WAL;"): discard
+  checkWalPragmaResult(journalModePragma)
+  checkExec(journalModePragma)
+
+  ## Table is the SQL query for creating the messages Table.
+  ## It contains:
+  ##  - 4-Byte ContentTopic stored as an Integer
+  ##  - Payload stored as a blob
+  checkExec """
```
oskarth, line 120: This part looks custom. Anything else ?

waku/node/v2/message_store.nim
```diff
@@ -0,0 +1,266 @@
...
+    env: env.release
+  ))
+
+template prepare(env: Sqlite, q: string, cleanup: untyped): ptr sqlite3_stmt =
+  var s: ptr sqlite3_stmt
+  checkErr sqlite3_prepare_v2(env, q, q.len.cint, addr s, nil):
+    cleanup
+  s
+
+proc bindParam(s: RawStmtPtr, n: int, val: auto): cint =
```
oskarth, line 139: I have no idea what this is, and I'm not confident in our ability to address bugs if things break here. I'd much rather this be upstream, as right now it looks like magic that we don't really understand. (I'm sure it makes sense to people in the Nimbus team).

Ditto a bunch of the templates above.

waku/node/v2/message_store.nim
```diff
@@ -0,0 +1,266 @@
...
+    sqlite3_bind_int(s, n.cint, val)
+  elif val is uint32:
+    sqlite3_bind_int(s, int(n).cint, int(val).cint)
+  elif val is int64:
+    sqlite3_bind_int64(s, n.cint, val)
+  else:
+    {.fatal: "Please add support for the 'kek' type".}
+
+proc put*(db: MessageStore, message: WakuMessage): MessageStoreResult[void] =
+  ## Adds a message to the storage.
```
oskarth, line 155: Is this a PubSub Message or a WakuMessage?

waku/node/v2/waku_types.nim
```diff
@@ -11,10 +11,13 @@ import
   libp2p/switch,
   libp2p/stream/connection,
   libp2p/protocols/pubsub/[pubsub, gossipsub],
-  nimcrypto/sha2
+  nimcrypto/sha2,
+  sqlite3_abi
+
 # constants required for pagination -------------------------------------------
-const MaxPageSize* = 100 # Maximum number of waku messages in each page 
+const MaxPageSize* = uint64(100) # Maximum number of waku messages in each page
```
oskarth, line 18: What is this about? Seems like this belongs to whatever proc needs it to be uint64

waku/node/v2/waku_types.nim
```diff
@@ -75,12 +72,18 @@ type
   HistoryPeer* = object
     peerInfo*: PeerInfo
 
+  MessageStoreResult*[T] = Result[T, string]
+
+  Sqlite* = ptr sqlite3
+
+  MessageStore* = ref object of RootObj
+    env*: Sqlite
```
oskarth, line 80: Can we make this accommodate for how things are currently being done with WakuMessages? That will significantly reduce risk of regression where we don't try to do three things at once in a PR.

waku/node/v2/waku_types.nim
```diff
@@ -189,14 +192,11 @@ proc generateRequestId*(rng: ref BrHmacDrbgContext): string =
   brHmacDrbgGenerate(rng[], bytes)
   toHex(bytes)
 
-proc computeIndex*(msg: WakuMessage): Index =
-  ## Takes a WakuMessage and returns its Index 
+proc id*(msg: WakuMessage): MDigest[256] =
```
oskarth, line 195: Why is this signature being changed for a PR introducing persistence?

oskarth (changes requested):
In general, this PR is too big to review.

I've mentioned this a few times the last few weeks, and I'll say it again: I'd strongly prefer a more incremental and gradual approach. That is, leave old code as it is and give this as an option. Persistence is a capability that we are introducing, it doesn't mean throwing away existing working things because "we don't need it". Theoretically and over a longer time horizon this might be correct, but that's not how these things work out in practice in most cases. Once we know for certain that things are working as expected, we can remove old logic code.

As it stands, my confidence that this PR won't introduce regression is very low. Additionally, fixing it with a revert is non-trivial due to how big the surface area of the code is. If the surface area was a simple feature toggle, it'd be a one line change and no biggie. This is how easy it should be to revert code that does a lot of things at once and can be isolated: https://github.com/status-im/nimbus-eth2/commit/937decb0e99baee02bdf9445d069b16728c94743

Also see in-line comments, many the same as last review round.

waku/node/v2/message_store.nim
oskarth: The put and get functions as well, but that seems like a level above. Why not separate upper and lower level? That way all this stuff becomes a black box, and we can treat it as a library, which it is (or should be).

oskarth:
Practically speaking, here's how to get out of this situation:

0. Stash current changes
1. Rebase on master and gracefully integrate parts of this PR in a piecemeal fashion
2. Introduce a flag/option for persistence
- this can be a simple variable for now with if statements
3. Ensure tests pass with flag off and on
4. Architecture the persistence layer to have a clear separation between black box library (what you copy pasted) and what is custom (put/get/sql query?).
- this means we can easily keep up to date with upstream, and not have to worry about the specifics if how SQL WAL ABI works or doesn't
5. Ensure persistence works by turning it off and on again, additionally by looking at DB manually/with tests (e.g.)
- doesn't require efficient indicies for now, unless these are already done or trivial (seems not - I'd leave this for 8, as per original problem scope).
6. Ensure tests pass with flag off and on
7. Issue new PR with this
8. Write up issue for any other changes you want to make, such as replacing Nim data structures with SQLite operations

- If there are regressions with persistence, we can simply turn the flag off.

- If there are no regression with persistence after some time, and issues in 8 makes sense to work on, then you can use your saved stash to address it then.

waku/node/v2/message_store.nim
```diff
@@ -0,0 +1,266 @@
...
+  ##
+  ## **Example:**
+  ##
+  ## .. code-block::
+  ##   let res = db.put(message)
+  ##   if res.isErr:
+  ##     echo "error"
+  let s = prepare(db.env, "INSERT INTO messages (id, timestamp, contentTopic, payload) VALUES (?, ?, ?, ?);"): discard
```
oskarth, line 163: These and other SQL statements seem like constant strings to me, this is the actual meat of semantics that one cares about. Though not sure how that plays together with the error handling below

PR #250, opened 2020-10-30
feature/waku-message-version

decanus:
This adds the version field as defined in the waku message spec.


kdeme (approved):
I've done the same change in https://github.com/status-im/nim-waku/pull/251/.

I'll go ahead and merge this.

PR #251, opened 2020-10-30
Add encoding/decoding of payload for Waku v2

kdeme:
Resolving https://github.com/status-im/nim-waku/issues/192. There are quite some changes/improvements that can be done still, see TODOs.

It will require changes in nim-eth, and I guess eventually, if we want to support different version (say v2) with the same (case) objects, will need to move/duplicate some code.


kdeme:
I did create https://github.com/vacp2p/specs/issues/237 regarding the version field being a string.

examples/v2/chat2.nim
```diff
@@ -68,9 +83,23 @@ proc connectToNodes(c: Chat, nodes: seq[string]) {.async.} =
...
 
 proc publish(c: Chat, line: string) =
-  let payload = cast[seq[byte]](line)
-  let message = WakuMessage(payload: payload, contentTopic: DefaultContentTopic)
-  c.node.publish(DefaultTopic, message)
+  when PayloadV1:
+    # Use Waku v1 payload encoding/encryption
+    let
+      payload = Payload(payload: line.toBytes(), symKey: some(DefaultSymKey))
+      version = "1"
```
oskarth, line 90: Version was changed to integer in other PR I believe

examples/v2/chat2.nim
```diff
@@ -49,6 +52,18 @@ type
   PrivateKey* = crypto.PrivateKey
   Topic* = waku_types.Topic
 
+
+# Similarly as Status public chats now.
```
oskarth, line 56: Does this mean that the symkey here is identical to one generated for "dingpu" chat in Waku v1 (generateFromPassword)? Maybe a good test, considering compatibility/bridging etc.

waku/node/v2/waku_types.nim
```diff
@@ -144,6 +144,8 @@ type
     listenStr*: string
     #multiaddrStrings*: seq[string]
 
+  WakuResult*[T] = Result[T, cstring]
```
oskarth, line 147: Can you explain why this is needed? I don't understand what the cstring is doing here.

oskarth (approved):
Generall lgtm, some minor comments and questions in-line

waku/node/v2/waku_types.nim
kdeme: Because I'm planning on only using string literals as error. Not adding any values that can change to the error string, so no need to use `string`.
I could have used `Result[T, cstring]` directly, but I created the type as I'm sure we can use it on other locations also.

examples/v2/chat2.nim
kdeme: Yup, aware of this :)

examples/v2/chat2.nim
kdeme: Not exactly, hence the 'similarly" ;-).

I'd need to do this call not on the topic but on the full "channel name".

waku/node/v2/waku_types.nim
decanus: @kdeme I usually like the pattern of using an enum. That means that we can compare the error to an enum for error handling and can also print its string value. What do you think?

tests/v2/test_waku_payload.nim
```diff
@@ -0,0 +1,111 @@
...
+    # Decoding
+    let msgDecoded = WakuMessage.init(pb.buffer)
+    check msgDecoded.isOk()
+
+    let
+      keyInfo = KeyInfo(kind:None)
+      decoded = decodePayload(msgDecoded.get(), keyInfo)
+
+    check:
+      decoded.isErr()
```
decanus, line 111: Add eol

waku/node/v2/waku_types.nim
decanus, line 147: I'd much prefer an enum

decanus (changes requested):

waku/node/v2/waku_types.nim
kdeme: It is not needed at the moment as no different action is taken based on the error outcome, see chat example.

If this turns out to be needed for the applications that use it, we can adjust (or have an additional one that holds an enum for specific API calls).

kdeme: Well sure, if there is a use case that such comparison is needed. If not, just keep the error string for logging, which is what I am doing now.

I'm fine with changing this later if there are such use cases.

jm-clius (approved):

decanus (approved):

PR #255, opened 2020-11-03
Feature/unsubscribe from content filter

### LINKED ISSUE #85: Waku v2 Node API: Unsubscribe from content filter

jm-clius:
closes #85 
This PR:
- brings API and RPC protobuf up to date with wakufilter 2.0.0-beta2 specification
- adds method to unsubscribe from content filter to wakunode2
- adds unsubscribe functionality to wakufilter. This includes a change to index subscribers on PeerID
- adds simple unit test to verify unsubscribe works


waku/node/v2/waku_types.nim
```diff
@@ -107,7 +108,7 @@ type
     rng*: ref BrHmacDrbgContext
     switch*: Switch
     peers*: seq[FilterPeer]
-    subscribers*: seq[Subscriber]
+    subscribers*: Table[PeerID, Subscriber]
```
decanus, line 111: This seems weird, `Subscriber` contains `PeerInfo` which has the `PeerID`. If you do this then a `Peer` can also not subscrib to multipl filters. I think this will cause various weird side effects.

decanus (changes requested):

waku/node/v2/waku_types.nim
jm-clius: Mmm. Yeah, you're right. Indexing on PeerID (which has a fast hash) was meant to simplify lookup when modifying the filter, but does limit the number of filters per subscriber to one. Rather than complicating this further, I'll rather revert to a Subscriber sequence and just filter on the PeerID already included in the Subscriber.

tests/v2/test_waku_filter.nim
```diff
@@ -63,3 +63,68 @@ procSuite "Waku Filter":
...
+    var responseCompletionFuture = newFuture[bool]()
+    proc handle(requestId: string, msg: MessagePush) {.gcsafe, closure.} =
+      check:
+        msg.messages.len() == 1
+        msg.messages[0] == post
+      responseCompletionFuture.complete(true)
+
+    let
+      proto = WakuFilter.init(dialSwitch, crypto.newRng(), handle)
+      rpc = FilterRequest(contentFilters: @[ContentFilter(topics: @[contentTopic])], topic: "topic", subscribe: true)
```
oskarth, line 90: I'd use default topic here as defined elswhere, mostly for clarity

waku/node/v2/wakunode2.nim
```diff
@@ -47,6 +47,30 @@ proc initAddress(T: type MultiAddress, str: string): T =
     raise newException(ValueError,
                        "Invalid bootstrap node multi-address")
 
+proc removeContentFilters(filters: var Filters, contentFilters: seq[ContentFilter]) {.gcsafe.} =
```
oskarth, line 50: Maybe some trace/debug logs here? To see updated content filter state

waku/protocol/v2/waku_filter.nim
```diff
@@ -157,3 +189,12 @@ proc subscribe*(wf: WakuFilter, request: FilterRequest): Future[string] {.async,
     let conn = await wf.switch.dial(peer.peerId, peer.addrs, WakuFilterCodec)
     await conn.writeLP(FilterRPC(requestId: id, request: request).encode().buffer)
   result = id
+
+proc unsubscribe*(wf: WakuFilter, request: FilterRequest) {.async, gcsafe.} =
+  # @TODO: NO REAL REASON TO GENERATE REQUEST ID FOR UNSUBSCRIBE OTHER THAN CREATING SANE-LOOKING RPC.
```
oskarth, line 194: Maybe we can cut this request id stuff out into a separate GH issue?

waku/node/v2/waku_types.nim
jm-clius: Reverted to seq implementation in 3de5025

waku/node/v2/wakunode2.nim
jm-clius: Added in cb146e3

waku/protocol/v2/waku_filter.nim
jm-clius: Opened #257 to keep track of outstanding work on `requestId`

oskarth (approved):
LGTM!

jm-clius:
@decanus if you're also happy with the changes (reverting to seq implementation of subscribers), I'll merge.

waku/node/v2/wakunode2.nim
```diff
@@ -47,6 +47,34 @@ proc initAddress(T: type MultiAddress, str: string): T =
...
+
+  var rIdToRemove: seq[string] = @[]
+  for rId, f in filters.mpairs:
+    # Iterate filter entries to remove matching content topics
+    for cf in f.contentFilters.mitems:
+      # Iterate content filters in filter entry
+      cf.topics.keepIf(proc (t: auto): bool = t notin unsubscribeTopics)
+    # make sure we delete the content filter
+    # if no more topics are left
+    f.contentFilters.keepIf(proc (cf: auto): bool = cf.topics.len > 0)
```
decanus, line 66: Not that it matters much, but I think here you could do something like:

```nim
f.contentFilters.keepItIf(it.cf.topics notin unsubscribeTopics)
```

not sure tho

jm-clius: I think we'll need separate topic iteration, as the `notin` operator for a sequence of items can only be performed one item at a time. But perhaps this will also only modify a content filter if _all_ its content topics are unsubscribed (as opposed to doing so on an individual basis)?

oskarth:
Ping @decanus

decanus (approved):
LGTM

PR #262, opened 2020-11-11
Accounting SWAP protocol basic data types and test

oskarth:
Partial progress towards https://github.com/status-im/nim-waku/issues/260

Ready to be reviewed and merged.

Making it more end to end and integrated will be done in subsequent PRs.

### DIFF
--- a/tests/v2/test_waku_swap.nim
+++ b/tests/v2/test_waku_swap.nim
@@ -1,0 +1,21 @@
+import
+  std/[unittest, options, tables, sets],
+  chronos, chronicles,
+  ../../waku/protocol/v2/waku_swap,
+  ../../waku/node/v2/waku_types,
+  ../test_helpers, ./utils
+
+procSuite "Waku SWAP Accounting":
+  test "Handshake Encode/Decode":
+    let
+      beneficiary = @[byte 0, 1, 2]
+      handshake = Handshake(beneficiary: beneficiary)
+      pb = handshake.encode()
+
+    let decodedHandshake = Handshake.init(pb.buffer)
+
+    check:
+      decodedHandshake.isErr == false
+      decodedHandshake.get().beneficiary == beneficiary
+
+    # TODO Cheque test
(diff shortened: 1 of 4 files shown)


waku/node/v2/waku_types.nim
```diff
@@ -146,8 +148,21 @@ type
...
+  Beneficiary* = seq[byte]
+
+  # TODO Consider adding payment threshhold and terms field
+  Handshake* = object
+    beneficiary*: Beneficiary
+
+  Cheque* = object
+    beneficiary*: Beneficiary
+    date*: uint32
+    amount*: uint32
```
staheri14, line 161: For the sake of authentication, a signature field can be added to the Cheque.

tests/v2/test_waku_swap.nim
```diff
@@ -0,0 +1,33 @@
...
+    check:
+      decodedHandshake.isErr == false
+      decodedHandshake.get().beneficiary == beneficiary
+
+  test "Cheque Encode/Decode":
+    let
+      amount = 1'u32
+      date = 9000'u32
+      beneficiary = @[byte 0, 1, 2]
+      cheque = Cheque(beneficiary: beneficiary, amount: amount, date: date)
```
decanus, line 26: This is probably an issue for the specs but I am not sure i like the naming of Cheque

waku/node/v2/waku_types.nim
oskarth: Yep, I'll do that in follow up PRs. Still thinking about how this should work exactly

tests/v2/test_waku_swap.nim
oskarth: Why? It describes what it does and it is what is used in the Swarm book. We could change it, but we'd have to have a good rationale and alternative then.

oskarth:
@jm-clius @decanus @staheri14 From my point of view this can be merged, let me know if there are any concerns. Changes are quite isolated so should have minimal impact on other parts of the code base. Further capabilities to be introduced in follow up PRs.

Issue #258, opened 2020-11-06
Waku v2 JSON-RPC REST API: Store protocol proof of concept


jm-clius:
# Problem
We need a basic proof of concept for the API approach decided on in #243. It should allow for historical queries via a JSON-RPC REST API. Pagination is excluded for now, so the aim is only to create an endpoint to return all history for a list of topics.
# Acceptance criteria
A unit testable and runnable JSON-RPC API for wakunode2. It must provide at least a `get_waku_v2_store_history` method that returns the history for a list of topics. Similar functionality (minus pagination) as described in [vacp2p/specs/#218](https://github.com/vacp2p/specs/pull/218).

PR #263, opened 2020-11-11
Waku v2 JSON-RPC REST API: Store protocol proof of concept

jm-clius:
Addresses #258 
Draft POC for a simple history query for a JSON-RPC based REST API.
There are many potentially controversial things to reach consensus on before specifying any other endpoints. Some things I can immediately think of:

1. _Creating an "api" subfolder and postfixing types with API_ - this is mostly because "RPC" is currently ambiguous and rpc_api is a bit unwieldy. Open for any other suggestions.
2. _Conversion between types for the RPC API and the internal Nim API_ - although many of the return and parameter types for the RPC API is likely to look very similar to the Nim API types, I believe they should be kept separate to show their conceptual difference (and different audience). This means some conversion is necessary.
3. _Using uint64 vs uint32_ - I can't get marshalling of uint32 to work in the nim-json-rpc library, though I'll discuss this with the Nimbus team. For POC purposes it was easier to just use uint64.
**UPDATE**: marshalling is failing because compiler tries to convert the uint32 to uint64 (which is immutable) and pass that as a var parameter. I'll fix by just adding explicit marshalling support for uint32 as per https://github.com/status-im/nim-json-rpc#fromjson
4. _Expected response for failed queries_ - as it stands, the POC implementation returns an empty response if the internal query fails. This is not where we want to end up (empty response merely implies no history, which is misleading), so I'm looking at different options, including just raising an exception. Any thoughts/ideas welcome.


waku/node/v2/api/waku_api_rpc.nim
```diff
@@ -0,0 +1,23 @@
+import
+  json_rpc/rpcserver,
+  eth/async_utils,
+  waku_api_types, waku_api_tools,
+  ../waku_types,../wakunode2
+
+proc setupWakuRPCAPI*(node: WakuNode, rpcsrv: RpcServer) =
+  
+  rpcsrv.rpc("get_waku_v2_history") do(query: HistoryQueryAPI) -> HistoryResponseAPI:
```
decanus, line 9:
```suggestion
  rpcsrv.rpc("get_waku_v2_history") do(query: HistoryQueryAPI) -> HistoryResponseAPI:
```

I really hate this name, will this be what the function is called or how exactly will this work?

tests/v2/test_rpc_api.nim
```diff
@@ -0,0 +1,88 @@
...
+      WakuMessage(payload: @[byte 3], contentTopic: ContentTopic(1)),
+      WakuMessage(payload: @[byte 4], contentTopic: ContentTopic(1)),
+      WakuMessage(payload: @[byte 5], contentTopic: ContentTopic(1)),
+      WakuMessage(payload: @[byte 6], contentTopic: ContentTopic(1)),
+      WakuMessage(payload: @[byte 7], contentTopic: ContentTopic(1)),
+      WakuMessage(payload: @[byte 8], contentTopic: ContentTopic(1)), 
+      WakuMessage(payload: @[byte 9], contentTopic: ContentTopic(2))]
+
+  for wakuMsg in msgList:
+    waitFor subscriptions.notify("foo", wakuMsg)
```
decanus, line 76: Id move this all into the test for now until other tests may need it.

kdeme (commented):
1. I think that api as folder name is perhaps a bit to general as a name, unless we also move the native api here then.
2. Yes, for Waku v1 there is also a mapping for several types (but not for all) but that was needed as the rpc calls didn't map directly on the native implementation: https://github.com/status-im/nim-waku/blob/master/waku/node/v1/rpc/rpc_types.nim. However, if it is not needed I and we can still choose it I would try to avoid it. Not fully sure here, what do you see as actual benefits with that approach?
3. Yes, as you already noticed, uint32 mashalling is simply not in the json-rpc module: https://github.com/status-im/nim-json-rpc/blob/master/json_rpc/jsonmarshal.nim#L87, so it needs to be added just as for other custom types. I think it can be added directly in the json-rpc module though.
4. The way nim-json-rpc works is that all exceptions end up in the json rpc response object its error field. So the way to do this is indeed to raise an exception with a specific message for your specific error.

waku/node/v2/api/waku_api_rpc.nim
oskarth: This is done to be explicit with mapping to REST API, see https://github.com/status-im/nim-waku/issues/243 This means it corresponds to a GET call with `/waku/v2/history` endpoint. How would you solve mapping problem without it?

My only comment here would be that it might make sense to call it `store` instead of `history` to make mapping more explicit. On the other hand, store is more of a verb, and history is the resource, so maybe it does make sense from a REST POV. @jm-clius do you have any thoughts on what the other endpoints would be called and whether Store API or History API makes more sense?

oskarth:
Thanks for the detailed PR description @jm-clius.

After thinking about the questions and reading the thread, I realize I have nothing to add to what @kdeme said. Agree with all his concerns, questions and comments :P

waku/node/v2/api/waku_api_rpc.nim
jm-clius: Mmm. After giving this some thought I'm backtracking on using `history`.  I used "get...history" because it felt more natural for a read action than "get...store", but this unnecessarily diverges from the underlying protocols. Endpoints for `relay` and `filter` will be called just that, so I'd rather maintain consistency.
The proposed scheme for naming read methods are `get_waku_v2_<API>_<resource>`. The most explicit mapping for this method would then be `get_waku_v2_store_query`.

jm-clius:
Thanks for the helpful comments!
@kdeme specific response:
1. I agree. Will come up with something more specific.
2. Well, the benefit would be that changes made to the "internal" Nim API does not necessarily affect the "external" REST API (only how the conversion is done). The external API could even be versioned differently and stability guaranteed via the conversion layer. If we don't foresee things diverging like this I'd agree it's an unnecessary layer of indirection.
3. Thanks! Will do.
4. Ah, that's good to know. Will implement in this way.

oskarth:
As for 2 above I don't have a strong opinion either way other than agreeing concerns raised on both sides.

jm-clius:
> As for 2 above I don't have a strong opinion either way other than agreeing concerns raised on both sides.
> 
> @jm-clius please mark as ready for review when you want us to look into the code per se as opposed to overall structure/direction

@oskarth Will do. For now it's still in draft waiting for https://github.com/status-im/nim-json-rpc/pull/88 to be merged, otherwise build will fail.
Other minor improvements added in the meantime:
1. Subfolder now called `rpc_api`. Could replace `rpc`-subfolder in future once more feature-complete
2. Adopted `get_waku_v2_<api>_<resource>` method naming pattern (i.e. `get_waku_v2_store_query` for POC)
3. Raise exception if no timely response received
4. As for RPC API vs Nim API types: decided to only use separate RPC API types and conversion where there is a clear divergence from the Nim API types. One such divergence is how optional parameters are treated. json-rpc library requires optional parameters to be explicitly marked as such (e.g. optional PagingInfo), whereas just leaving parameters empty on Nim API implies the same thing in many instances.

Direction is clear enough now to start drafting raw specs. Thanks!

tests/v2/test_rpc_api.nim
```diff
@@ -0,0 +1,91 @@
...
+      rng = crypto.newRng()
+      privkey = crypto.PrivateKey.random(Secp256k1, rng[]).tryGet()
+      bindIp = ValidIpAddress.init("0.0.0.0")
+      extIp = ValidIpAddress.init("127.0.0.1")
+      port = Port(9000)
+      node = WakuNode.init(privkey, bindIp, port, some(extIp), some(port))
+
+    waitFor node.start()
+
+    waitFor node.mountRelay(@["waku"])
```
oskarth, line 35: I think this should be the default waku topic, mostly for consistency

tests/v2/test_rpc_api.nim
```diff
@@ -0,0 +1,91 @@
...
+        WakuMessage(payload: @[byte 3], contentTopic: ContentTopic(1)),
+        WakuMessage(payload: @[byte 4], contentTopic: ContentTopic(1)),
+        WakuMessage(payload: @[byte 5], contentTopic: ContentTopic(1)),
+        WakuMessage(payload: @[byte 6], contentTopic: ContentTopic(1)),
+        WakuMessage(payload: @[byte 7], contentTopic: ContentTopic(1)),
+        WakuMessage(payload: @[byte 8], contentTopic: ContentTopic(1)), 
+        WakuMessage(payload: @[byte 9], contentTopic: ContentTopic(2))]
+
+    for wakuMsg in msgList:
+      waitFor subscriptions.notify("foo", wakuMsg)
```
oskarth, line 79: Why is this foo but above it is test? In general I find the subscription code a bit hard to reason about, perhaps due to how indirect it is

waku/v2/node/rpc_api/waku_api_types.nim
```diff
@@ -0,0 +1,12 @@
+import
+  ../../waku_types,
+  options
+
+type
+  HistoryQueryAPI* = object
```
oskarth, line 6: How does this differ from HistoryQuery? If this is about the JSONRPC specific types, I'd be explicit about that. HistoryQuery is used in Nim Node API, so seems like overloading. Same for filename.

What specific differences do we expect here? That would give a hint to the name. Alternatively, perhaps this can be namespaced? e.g. rpc_api.HistoryQuery. What issues did you run into with that?

Alternatively, maybe HistoryQueryJSONRPC? Though it isn't really JSON here...

waku/v2/node/rpc_api/waku_api_tools.nim
```diff
@@ -0,0 +1,21 @@
...
+## Since the Waku v2 JSON-RPC API has its own defined types,
+## we need to convert between these and the types for the Nim API
+proc convertToAPI*(response: HistoryResponse): HistoryResponseAPI =
+  if response.pagingInfo != PagingInfo():
+    # PagingInfo is not empty. Set pagingInfo Option on API response.
+    result = HistoryResponseAPI(messages: response.messages, pagingInfo: some(response.pagingInfo))
+  else:
+    result = HistoryResponseAPI(messages: response.messages, pagingInfo: none(PagingInfo))
+
+proc convertFromAPI*(query: HistoryQueryAPI): HistoryQuery =
```
oskarth, line 16: Ditto below re API ambiguity

oskarth (commented):
Generally and as a PoC it looks good, I think the main issue is one of naming. We need to give this API a good name, to make it separate from Nim Node API. My current thinking is that this should be:

- Nim Node API
- JSON RPC API

And later on the JSON RPC API will be mirrored/morphed into HTTP REST API.

For Nim Node API, naming convention is (1) HistoryQuery is native and not qualified, since it happens in Nim land (2) node.query if you use the Nim Node API. We need something similar for JSON RPC, and the best one I can think of is "JSONRPC" or "RPC". What do you think?

tests/v2/test_rpc_api.nim
jm-clius: Will do!

tests/v2/test_rpc_api.nim
jm-clius: Subscriptions are indexed against codec ("test" above). "foo" is the PubSub topic - since the test subscription is not initialised with any PubSub topics, it will be notified of all messages, regardless of topic. This is confusing to follow, so I'll try to improve by clearer naming.

waku/v2/node/rpc_api/waku_api_types.nim
jm-clius: The main difference is that the json-rpc library requires optional parameters/fields to be explicitly marked as such (specifically `PagingInfo` in this instance), whereas it can be implicitly so on the Nim API.

I've gone back and forth in my thinking about this, but I'm leaning more and more towards abandoning unnecessary parameter envelopes on the JSON-RPC API (this also goes for many functions on the [spec PR](https://github.com/vacp2p/specs/pull/252)). This would imply that, rather than taking a HistoryQuery `Object` parameter, `get_waku_v2_store_query` would simply take an array of `topics` and `Option[PagingInfo]` as two separate parameters. This will be clearer to document and hopefully avoid confusing overloading issues. Lmk what you think.

jm-clius:
> Generally and as a PoC it looks good, I think the main issue is one of naming. We need to give this API a good name, to make it separate from Nim Node API. My current thinking is that this should be:
> 
> * Nim Node API
> * JSON RPC API
> 
> And later on the JSON RPC API will be mirrored/morphed into HTTP REST API.
> 
> For Nim Node API, naming convention is currently (1) `HistoryQuery` is native and not qualified, since it happens in Nim land (2) `node.query` if you use the Nim Node API. We need something similar for JSON RPC, and the best one I can think of is "JSONRPC" or "RPC". What do you think?

As to your first point, I like "JSON RPC API" and "HTTP REST API". :)
As to naming convention: I prefer "JSONRPC" to be explicit, but am not sure in what context you suggest using this. Do you propose that this be used as qualifier for resource naming? In other words, something like `get_waku_v2_store_jsonrpc_query`?

tests/v2/test_rpc_api.nim
jm-clius: Addressed in 4047bed

tests/v2/test_rpc_api.nim
jm-clius: Addressed in 4047bed

waku/v2/node/rpc_api/waku_api_types.nim
oskarth: > optional parameters/fields to be explicitly marked as such (specifically PagingInfo in this instance), whereas it can be implicitly so on the Nim API.

Is there a reason not to do this explicitly in the Nim Node API?

> This would imply that, rather than taking a HistoryQuery Object parameter, get_waku_v2_store_query would simply take an array of topics and Option[PagingInfo] as two separate parameters. This will be clearer to document and hopefully avoid confusing overloading issues. Lmk what you think.

Seems reasonable, perhaps writing out a few example queries (eg 1) initial query 2) paging with cursor and custom page size) for: JSON RPC API (existing and proposed) and what this would be mapped to for REST HTTP would make it easier to make a decision? Can e.g. put it in a notes.status.im and share in #waku or as you prefer.

oskarth:
> As to naming convention: I prefer "JSONRPC" to be explicit, but am not sure in what context you suggest using this. Do you propose that this be used as qualifier for resource naming? In other words, something like get_waku_v2_store_jsonrpc_query?

Qualifier for data types and namespaces, what is currently `*API` right now (assuming these are distinct data types). Resource would be same, `get_waku_v2_store_query`, which would map to `GET waku/v2/store/query` (actually I'm not sure what GETting a query means, maybe this is POSTing? Getting would be for a specific resource under `store`, I think.

waku/v2/node/rpc_api/waku_api_tools.nim
```diff
@@ -0,0 +1,21 @@
...
+  options,
+  ../wakunode2
+
+## Conversion tools
+## Since the Waku v2 JSON-RPC API has its own defined types,
+## we need to convert between these and the types for the Nim API
+proc convertToAPI*(response: HistoryResponse): HistoryResponseAPI =
+  if response.pagingInfo != PagingInfo():
+    # PagingInfo is not empty. Set pagingInfo Option on API response.
+    result = HistoryResponseAPI(messages: response.messages, pagingInfo: some(response.pagingInfo))
```
arnetheduck, line 12: we avoid `result` as it turns off useful compiler warnings and is prone to bugs - `expression` return is preferred, followed by explicit `return` (expressions don't work in all contexts, such as `async`) - in this case, just remove `result =`

waku/v2/node/rpc_api/waku_api_rpc.nim
```diff
@@ -0,0 +1,29 @@
+import
+  json_rpc/rpcserver,
+  eth/async_utils,
+  waku_api_tools, waku_api_types,
```
arnetheduck, line 4: prefer explicit `./waku_api_tools` - nim path handling is fickle

waku/v2/node/rpc_api/waku_api_tools.nim
```diff
@@ -0,0 +1,21 @@
+import
+  waku_api_types,
+  options,
```
arnetheduck, line 3: `std/options` - std for anything from std lib

waku/v2/node/rpc_api/waku_api_types.nim
jm-clius: > Is there a reason not to do this explicitly in the Nim Node API?

It can be done, but I tried to avoid having concerns from the JSON-RPC API spill over to the Nim Node API, even if just "dictating" how options should be implemented. In this instance it is a tiny change, so probably easy enough to do. Hopefully though the whole type separation can be avoided if we use multi-parameter approach as below.

> Seems reasonable, perhaps writing out a few example queries

Good idea! After my discussion with WalletConnect and reading up more about REST good practices, there are a couple of changes I'd like to make to this POC and the [draft spec PR](https://github.com/vacp2p/specs/pull/252). It will be good to gain consensus about an example before continuing. I will write something up.

waku/v2/node/rpc_api/waku_api_tools.nim
jm-clius: Thanks. Addressed in 6bea1b3. General issue in #287.

waku/v2/node/rpc_api/waku_api_rpc.nim
jm-clius: Addressed in 6bea1b3.

waku/v2/node/rpc_api/waku_api_tools.nim
jm-clius: Addressed in 6bea1b3.

arnetheduck:
the naming convention currently in nimbus: https://github.com/status-im/nimbus-eth2/tree/master/beacon_chain/rpc

jm-clius:
I've jotted down an idea (with examples) for the Store API. https://notes.status.im/s/5lsytm9ac

jm-clius:
Most recent commit introduced changes to follow conventions discussed in [these examples](https://notes.status.im/s/5lsytm9ac) and in the [nimbus-eth2 API](https://github.com/status-im/nimbus-eth2/tree/master/beacon_chain/rpc). Most importantly:
- changed `method` name to reflect `get_waku_v2_<api>_<api_version>_<resource>` pattern
- changed folder structure and module names. Tried to do something roughly similar to the [nimbus-eth2 API](https://github.com/status-im/nimbus-eth2/tree/master/beacon_chain/rpc)
- use parameters as in [the examples](https://notes.status.im/s/5lsytm9ac)

oskarth (approved):
LGTM!

PR #264, opened 2020-11-13
Minimal accounting

oskarth:
Closes https://github.com/status-im/nim-waku/issues/260

Start of accounting PoC bookkeeping and keeping track of balance with peer. See  and https://github.com/vacp2p/specs/blob/master/specs/waku/v2/waku-swap-accounting.md

Ready to review. Changes under flag so no functional code changes. With flag on, it prints accounting stats to STDOUT:

```
NF 2020-11-16 11:43:02.860+08:00 mounting store                             topics="wakunode" tid=7524
hi
hi
INF 2020-11-16 11:43:04.105+08:00 Hit store handler                          tid=7524
DBG 2020-11-16 11:43:04.105+08:00 Using SWAPAccounting query                 topics="wakunode" tid=7524
INF 2020-11-16 11:43:05.363+08:00 Accounting for                             topics="wakuswap" tid=7524 peerId=16*3q4YjS n=2
```

I.e. it has received two WakuMessages from that peer.

Follow up issues to be addressed in future PRs:
- Better state management
- End to end

### DIFF
--- a/waku/protocol/v2/waku_store.nim
+++ b/waku/protocol/v2/waku_store.nim
@@ -353,9 +353,6 @@ proc query*(w: WakuStore, query: HistoryQuery, handler: QueryHandlerFunc) {.asyn
 
   handler(response.value.response)
 
-# TODO Move to better place
-type AccountUpdateFunc = proc(peerId: PeerId, amount: int) {.gcsafe.}
-
 # NOTE: Experimental, maybe incorporate as part of query call
 proc queryWithAccounting*(w: WakuStore, query: HistoryQuery, handler: QueryHandlerFunc,
                           accountFor: AccountUpdateFunc) {.async, gcsafe.} =
--- a/waku/protocol/v2/waku_store.nim
+++ b/waku/protocol/v2/waku_store.nim
@@ -382,8 +382,7 @@ proc queryWithAccounting*(w: WakuStore, query: HistoryQuery, handler: QueryHandl
   # NOTE Perform accounting operation
   #  if SWAPAccountingEnabled:
   let peerId = peer.peerInfo.peerId
-  # TODO Replace with WakuMessage's
-  let volume = 1
-  accountFor(peerId, volume)
+  let messages = response.value.response.messages
+  accountFor(peerId, messages.len)
 
   handler(response.value.response)
--- a/waku/protocol/v2/waku_store.nim
+++ b/waku/protocol/v2/waku_store.nim
@@ -352,3 +352,38 @@ proc query*(w: WakuStore, query: HistoryQuery, handler: QueryHandlerFunc) {.asyn
     return
 
   handler(response.value.response)
+
+# TODO Move to better place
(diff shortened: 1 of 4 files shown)


waku/protocol/v2/waku_store.nim
```diff
@@ -336,3 +336,38 @@ proc query*(w: WakuStore, query: HistoryQuery, handler: QueryHandlerFunc) {.asyn
     return
 
   handler(response.value.response)
+
+# TODO Move to better place
+type AccountUpdateFunc = proc(peerId: PeerId, amount: int) {.gcsafe.}
+
+# NOTE: Experimental, maybe incorporate as part of query call
+proc queryWithAccounting*(w: WakuStore, query: HistoryQuery, handler: QueryHandlerFunc,
```
oskarth, line 344: Reason for this change is:

1) Not disrupt public API (yet)
2) We need access to node, and peer. Callback only gives us messages, and inside function doesn't have a reference to node. With this handler we can callback to node object and then pass onto accounting module.

jm-clius: Makes sense.
Should we consider collating all handlers into a single `WakuStoreListener` object and make that a parameter for the API call so we can more easily extend in future? Handlers inside the listener are optional and only triggered if set.

oskarth: Yeah was thinking if an open options map or similar would make sense here, in general.

@decanus IIRC you were doing something similar somewhere but forgot where exactly?

decanus: Yeah, thats what the subscriptions are for. Or at least it sounds like it.

oskarth: So how would this work with subscribe? Do you get access to the same data with peer etc? Not obvious to me (from phone).

decanus: Oh actually, I don't think these parts work, let me think and I'll get back to you.

staheri14:
@oskarth what does "WakuMessage volume metric" mean?

waku/protocol/v2/waku_store.nim
oskarth: In any case, above is out of scope for initial PoC.

oskarth:
@staheri14 Good question! SWAP was designed for bandwidth incentives, similar to Bittorrent. See https://github.com/vacp2p/specs/blob/master/specs/waku/v2/waku-swap-accounting.md#accounting As a closely related metric in number of WakuMessages, which is the primary unit that is being exchanged in the service network that we are created. It is also conceptually simpler than thinking in terms of raw bandwidth. In some cases WakuMessages are quite big etc, so we might want to revisit it, but for a PoC I believe it makes sense.

jm-clius (approved):
As I've learned to say, LGTM! 👍

decanus (approved):

PR #268, opened 2020-11-13
feature/persistence-simple

### LINKED ISSUE #245: Production-ready: Persistence of store messages

decanus:
closes #245


waku/node/v2/message_store.nim
```diff
@@ -0,0 +1,222 @@
...
+  libp2p/protobuf/minprotobuf,
+  libp2p/stream/connection,
+  stew/results, metrics
+
+{.push raises: [Defect].}
+
+# The code in this file is an adaptation of the Sqlite KV Store found in nim-eth.
+# https://github.com/status-im/nim-eth/blob/master/eth/db/kvstore_sqlite3.nim
+#
+# Most of it is a direct copy, the only unique functions being `get` and `put`.
```
oskarth, line 17: It'd be nice if we could keep this the custom stuff in a separate file, but maybe this is tricky? Not sure

waku/node/v2/waku_types.nim
```diff
@@ -204,14 +212,22 @@ proc generateRequestId*(rng: ref BrHmacDrbgContext): string =
...
+  ## Generates an ID from the message
   var ctx: sha256
   ctx.init()
   ctx.update(msg.contentTopic.toBytes()) # converts the contentTopic to bytes
   ctx.update(msg.payload)
-  let digest = ctx.finish() # computes the hash
+  result = ctx.finish() # computes the hash
   ctx.clear()
 
+<<<<<<< HEAD
```
oskarth, line 224: rebase

waku/node/v2/waku_types.nim
```diff
@@ -204,14 +212,22 @@ proc generateRequestId*(rng: ref BrHmacDrbgContext): string =
   brHmacDrbgGenerate(rng[], bytes)
   toHex(bytes)
 
-proc computeIndex*(msg: WakuMessage): Index =
-  ## Takes a WakuMessage and returns its Index 
+proc id*(msg: WakuMessage): MDigest[256] =
```
oskarth, line 215: unclear what this is

waku/node/v2/message_store.nim
decanus: I don't think it makes sense because its all part of the db. might be tricky.

waku/node/v2/waku_types.nim
decanus: It's an id for the message, theres a comment explaining it.

oskarth: that's not what I mean, what I mean is why we are changing the signature

waku/node/v2/wakunode2.nim
```diff
@@ -243,9 +243,9 @@ proc mountFilter*(node: WakuNode) =
   node.switch.mount(node.wakuFilter)
   node.subscriptions.subscribe(WakuFilterCodec, node.wakuFilter.subscription())
 
-proc mountStore*(node: WakuNode) =
+proc mountStore*(node: WakuNode, store: MessageStore = nil) =
```
oskarth, line 246: I like that there is a default argument here

waku/node/v2/message_store.nim
arnetheduck: at some point, we might make a "nice" nim api for sqlite out of these various usages and then refactor - wouldn't bother until then, really

waku/node/v2/waku_types.nim
decanus: Oh, because I only need to gather the ID of the message at one point and not the complete index. So I seperated the 2.

waku/node/v2/waku_types.nim
```diff
@@ -204,14 +212,17 @@ proc generateRequestId*(rng: ref BrHmacDrbgContext): string =
...
   var ctx: sha256
   ctx.init()
   ctx.update(msg.contentTopic.toBytes()) # converts the contentTopic to bytes
   ctx.update(msg.payload)
-  let digest = ctx.finish() # computes the hash
+  result = ctx.finish() # computes the hash
   ctx.clear()
 
-  result.digest = digest
+proc computeIndex*(msg: WakuMessage): Index =
```
oskarth, line 224: Oh computeIndex is here, diff was confusing me

decanus: I moved it back, realized I can just use computeIndex.

waku/protocol/v2/waku_store.nim
```diff
@@ -292,10 +292,20 @@ method init*(ws: WakuStore) =
...
 
-proc init*(T: type WakuStore, switch: Switch, rng: ref BrHmacDrbgContext): T =
+  if ws.store.isNil:
+    return
+
+  proc onData(timestamp: uint64, msg: WakuMessage) =
+    ws.messages.add(IndexedWakuMessage(msg: msg, index: Index(digest: msg.id(), receivedTime: float(timestamp))))
+
+  # @TODO ERROR
+  discard ws.store.getAll(onData)
```
oskarth, line 302: Basic error handling and log

waku/protocol/v2/waku_store.nim
```diff
@@ -292,10 +292,20 @@ method init*(ws: WakuStore) =
   ws.handler = handle
   ws.codec = WakuStoreCodec
 
-proc init*(T: type WakuStore, switch: Switch, rng: ref BrHmacDrbgContext): T =
+  if ws.store.isNil:
```
oskarth, line 295: Is there a eason to do it this way and not be more explicit with `if ws.store`/flag then <store path>?

waku/protocol/v2/waku_store.nim
```diff
@@ -310,7 +320,10 @@ proc subscription*(proto: WakuStore): MessageNotificationSubscription =
   proc handle(topic: string, msg: WakuMessage) {.async.} =
     let index = msg.computeIndex()
     proto.messages.add(IndexedWakuMessage(msg: msg, index: index))
-    
+    if proto.store.isNil:
```
oskarth, line 323: Ditto above

waku/protocol/v2/waku_store.nim
```diff
@@ -310,7 +320,10 @@ proc subscription*(proto: WakuStore): MessageNotificationSubscription =
...
+      return
+  
+    discard proto.store.put(index, msg)
```
oskarth, line 326: Basic error handling

oskarth (commented):
Almost there, see inline comments.

How can one interact with the DB out of proc? Where is it saved? I expect something like `data` dir to be created and then `sqlite data/messages.db` to interact with it. is this accurate? Worth adding to doc perhaps, or as comment

waku/protocol/v2/waku_store.nim
decanus: I mean adding a flag is just another thing, we only initizalize the store right now if a dbpath is set. I think that is explicit.

waku/node/v2/message_store.nim
oskarth: aight let's leave it like this for now

decanus:
@oskarth a user passes the `--dbpath` that initializes the database, so whatever you select as that is the directory where the db is saved.

oskarth:
Does it work with and without store enabled? What is default behaviour now? Does chat2 work with one node stopping and restating now (history loaded)?

oskarth:
Do we want to make a default data dir and argument then perhaps? Or add to docs

decanus:
@oskarth yes it works with and without store. If store is enabled, it initialzies the DB with the path and tries to load all the messages from the db into memory.

waku/protocol/v2/waku_store.nim
oskarth: aight

oskarth (approved):
LGTM, if you have tested this I don't see this introducing any regressions

oskarth:
I'd add basic docs though

jm-clius (approved):

oskarth:
@decanus CI is failing

decanus:
Just tested locally by:

1 - Starting a node
2 - Connecting with chat
3 - Sending messages to node
4 - Restarting both node and chat
5 - Ensuring sent messages are returned by history

PR #276, opened 2020-11-17
update/nim-libp2p

decanus:


oskarth (changes requested):
Update all submodules at once, otherwise bound to run into issues.

Also, I believe @jm-clius might already be working on this rebase? Probably best to coordinate on effort here.

jm-clius:
@decanus @oskarth yes, did an update, but some recent changes to nim-confutils broke the compilation of some modules if we don't `import std/options` in each of them. Spoke to Zahary who agreed confutils should `export options` rather than requiring all clients to `import` that. Was just about to open a PR for that change, then do another all-submodules-update and open for review, though there may be some turnaround time.

oskarth:
@jm-clius I'd probably do the `import` for now to not be blocked by it, then both @decanus and your PoC PR can proceed. Later on we can clean up imports.

arnetheduck:
fyi just pushed a major fix to libp2p yesterday, make sure to get it as well

jm-clius:
Will do, @oskarth and @arnetheduck

jm-clius:
https://github.com/status-im/nim-waku/pull/279 should contain an update for all submodules

oskarth:
Out of date

PR #280, opened 2020-11-19
refactor/db-message-store-split

decanus:
This PR splits the MessageStore implementation from the general Sqlite stuff. This makes it so that implementing filter-persistence will be far easier once we decide to along with any other database related wrappers.

Currently thinking about upstreaming this into @arnetheduck's library


waku/v2/node/message_store.nim
```diff
@@ -10,8 +10,6 @@ import
   ../waku_types,
   ./sqlite
 
-{.push raises: [Defect].}
```
dryajov, line 12: You probably do want this, if you're getting errors around unhandled exceptions, it's probably because there are unhandled exceptions and they will cause issues in the long run, specially if using `Result`.

oskarth (changes requested):
The PR description doesn't explain what issue it is targeting, what problem it is solving or what the PR is doing.

decanus:
@arnetheduck I could move this into your nim-sqlite library if you think that would be useful. Its a very simple wrapper around sqlite. We can probably also add some other functions like `queryOne` etc. Right now we support `exec` and `query`

waku/v2/node/message_store.nim
decanus: @dryajov this was mainly on the callback function DataProc, it didn't seem to ever really call an exception. We do still have this in the database section however.

arnetheduck:
> I could move this into your nim-sqlite library if you think that would be useful. Its a very simple wrapper around sqlite. We can probably also add some other functions like queryOne etc. Right now we support exec and query

it's not a bad idea in general, though it would require that we refactor the library slightly so that it has a structure similar to https://github.com/status-im/nim-secp256k1 - ie a raw interface that exposes C, then a "high-level" wrapper on top of that

waku/v2/node/message_store.nim
arnetheduck: you need to annotate the `DataProc` alias specifically with `raises: [Defect]`  so that anyone that provides a callback is forced to handle their exceptions, else the compiler must make a conservative estimate

arnetheduck: see `kvstore.nim`

waku/v2/node/sqlite.nim
```diff
@@ -0,0 +1,207 @@
+import
```
oskarth, line 1: Does this belong in the node package? Think we might be putting a bit too much in there

oskarth (approved):
Generally seems reasonable to me, without looking too much into details

waku/v2/node/sqlite.nim
decanus: im happy to move it elsewhere.

jm-clius (approved):

PR #282, opened 2020-11-20
Refactor: waku_types separation of concern (trial with waku_swap)

oskarth:
Starts to address https://github.com/status-im/nim-waku/issues/281

Opening PR to start discussion and to get some input on what is idiomatic here. Once there is some form of agreement we can do it for other protocols etc too.

No functional changes.

# Notes

There are two commits with two approaches.

1) The first uses types in the file a la nim-libp2p.

2) The second uses a separate folder and name for types. This is the style used in nimbus-eth2.

I prefer the second approach, as it seems cleaner in terms of separation of concern, having an interface file at the bottom, etc. I also believe nimbus-eth2 is the more mature code base.

Things could probably be structured differently as well, open to ideas (as long as they fit the scope of PR and are simple).

As for names, I considered using `swap/swap_types.nim`. However, `swap` is also a proc in Nim, so it gets syntax highlighted weirdly in my editor and this might lead to odd behaviour (don't know how Nim details with this, but seems safer to avoid duplicate names). If it isn't a problem and this can be addressed with "symbol exclusion" (e.g. as exists in Clojure) I'd be fine with it.


kdeme:
I mostly disliked the part that some of the types were only used in 1 (protocol) module, yet you had to go look for them several folders up in the node folder.

I think, especially for "private" types, they can be just defined within the actual protocol module itself, or in a separate file in the same folder. Basically, your first and second option.

The second option, is a bit better I guess for public types that are used around (without requiring functionality of the actual module). And for types that don't have a clear module as "owner", e.g. `WakuMessage`, you could have a separate common_types or so file I guess, or keep them there where it makes most sense. I mean, no need to create this for only 1 type for example.

decanus:
@oskarth I definately prefer having certain types in their own files. For example the protocol specific types for store, swap and filter.

oskarth:
Alright, sounds this general approach seems reasonable to people (though it'd be better to indicate with actual review). Agree re no need to overdo it, and that more private types can be done in-line.

I'll go ahead and merge this once CI passes, as it is useful for waku_swap work. Then follow up PRs to come for other waku_types fixes/protocols, etc.

PR #292, opened 2020-11-23
Accounting credit receiving node

oskarth:
Closes https://github.com/status-im/nim-waku/issues/285

- Does accounting on receiving end
- Exposes swap protocol through store protocol
- Replaces accountFor with credit and debit function
- Simplify query sig and misc formatting


oskarth:
Ping @decanus @staheri14 @jm-clius

PR #293, opened 2020-11-23
fix/add-metrics

decanus:


oskarth (commented):
This is an internal thing right? Probably best to mark it as such. Might make sense to have it in docs folder?

oskarth:
The link also goes to v1, and if I select v2 I see https://grafana.infra.status.im/d/yy23i_zGk/nim-waku-v1?orgId=1&refresh=5m&var-host=.*&var-fleet=wakuv2.test&var-dc=ac-cn-hongkong-c&var-dc=do-ams3&var-dc=gc-us-central1-a which provides nonsense data (v1 envelopes). Where is the issue where we explained what metrics we want to expose? This doesn't seem to have been done properly, and is what I would expect when we talk about "metrics for production ready".

oskarth:
I see peer count but no idea if it is accurate, guessing not since there should be 3 peers in cluster so minimum of two peer count? Where is the issue this PR is addressing?

oskarth:
Re location, I created a folder in https://github.com/status-im/nim-waku/pull/308/files which can be used for `metrics.md` as an internal resource.

decanus:
@oskarth I linked the wrong dashboard, check out the current v2 dashboard. I will make an issue for other metrics.

oskarth (changes requested):
This is internal right? Also we have Waku v1 and v2 specific READMEs. For internal stuff, docs are here https://github.com/status-im/nim-waku/tree/master/docs/contributors as mentioned before.

oskarth:
As for the metrics themselves, https://grafana.infra.status.im/d/qrp_ZCTGz/nim-waku-v2?orgId=1&refresh=5m I see some Libp2p skeleton stuff but:

1) They are all zero, so seem to be broken
2) Still lacking an issue with clear problem description of what we need for metrics to be production-ready, as mentioned a few times in last 1-2 months.

decanus:
https://github.com/status-im/nim-waku/issues/324

oskarth:
Ping on above @decanus

oskarth:
Closing as outdated

PR #304, opened 2020-11-26
Accounting end to end

oskarth:
Addresses end to end in a minimal fashion https://github.com/status-im/nim-waku/issues/270

- Defines threshhold for payment
- Upon reaching this threshhold a cheque is sent
- The cheque updates accounting state for both peers
- Also defines and logs disconnect threshhold

Out of scope of PR (to come in future issues/prs):
- Handshakes
- Authentication
- More parameters
- Smart peer selection
- Disconnect policy

### DIFF
--- a/waku/v2/protocol/waku_swap/waku_swap.nim
+++ b/waku/v2/protocol/waku_swap/waku_swap.nim
@@ -81,10 +81,44 @@ proc init*(T: type Cheque, buffer: seq[byte]): ProtoResult[T] =
 
 # TODO Test for credit/debit operations in succession
 
+proc sendCheque*(ws: WakuSwap) {.async.} =
+  # TODO Better peer selection, for now using hardcoded peer
+  let peer = ws.peers[0]
+  let conn = await ws.switch.dial(peer.peerInfo.peerId, peer.peerInfo.addrs, WakuSwapCodec)
+
+  info "sendCheque"
+
+  # TODO Add beneficiary, etc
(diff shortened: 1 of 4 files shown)


waku/v2/protocol/waku_swap/waku_swap.nim
```diff
@@ -103,6 +145,15 @@ proc init*(wakuSwap: WakuSwap) =
       wakuSwap.accounting[peerId] = n
     info "Accounting state", accounting = wakuSwap.accounting[peerId]
 
+    # TODO Isolate to policy function
+    # TODO Tunable payment threshhold, hard code for PoC
+    let paymentThreshhold = 1
+    if wakuSwap.accounting[peerId] >= paymentThreshhold:
```
jm-clius, line 151: Ok, so for now the peer takes no action to "remind" the indebted node to send a cheque or agree on the outstanding balance. Do we foresee persistence issue in future where peers have differing views on outstanding balances after a restart, or will this be a non-issue when contracts are specified?

jm-clius (approved):
LGTM!

waku/v2/protocol/waku_swap/waku_swap.nim
oskarth: TBD, some persistence etc might be needed indeed. In general, it is useful if both peers agree but if one peer is misbehaving/dishonest that shouldn't break things, conceptually.

Issue #299, opened 2020-11-24
Waku v2 JSON-RPC REST API: Basic implementation


jm-clius:
### Description

Implement basic end-to-end functionality for the Waku v2 JSON-RPC REST API. This is a subtask of the wider API implementation described in #259. Parallel issue to track spec write-up in https://github.com/vacp2p/specs/issues/233

### Acceptance criteria

1. Basic Pub/Sub functionality specified and implemented:

- [x] Client can subscribe to PubSub topic(s)

- [x] Client can receive message notifications for a subscription by polling the server

- [x] Client can publish to PubSub topic(s)

2. Basic ability to filter on content specified and implemented:

- [x] Client can subscribe to content filter(s)

- [x] Client can receive message notifications for a content filter subscription in a similar manner to the polling described above.

3. Basic Debug API specified and implemented

- [x] Client can request and receive information about the WakuNode

PR #305, opened 2020-11-26
Some basic Debug and Relay API calls

jm-clius:
Incremental PR towards #299 

It adds:
- Basic Debug API
- Ability to subscribe and unsubscribe from PubSub topics

Still TODO:
- Publishing WakuMessages
- Polling subscribed PubSub topics for new messages
- Filter API

### DIFF
--- a/waku/v2/node/jsonrpc/jsonrpc_callsigs.nim
+++ b/waku/v2/node/jsonrpc/jsonrpc_callsigs.nim
@@ -1,1 +1,12 @@
+# Debug API
+
+proc get_waku_v2_debug_v1_info(): WakuInfo
+
+# Relay API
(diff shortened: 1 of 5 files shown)


waku/v2/node/jsonrpc/jsonrpc_callsigs.nim
```diff
@@ -1 +1,12 @@
+# Debug API
+
+proc get_waku_v2_debug_v1_info(): WakuInfo
```
oskarth, line 3: This PR doesn't remove old one, but once we do this will impact health checks and status for cluster, just FYI

oskarth (approved):
LGTM

PR #311, opened 2020-11-30
Feature/jsonrpc basic impl

jm-clius:
This is an incremental PR towards #299.

It adds a basic polling call to the API, whereby a client can receive messages on subscribed topics by polling an in-memory cache on the server. The cache clears after every call, so that only new messages are returned by each new call.

Still outstanding:
- ability to limit the number of cached topics and messages (NB to avoid memory leaks)
- ability to specify the number of messages to read
- priming the cache with topics that the node is already subscribed to

### DIFF
--- a/tests/v2/test_jsonrpc_waku.nim
+++ b/tests/v2/test_jsonrpc_waku.nim
@@ -106,6 +106,77 @@ procSuite "Waku v2 JSON-RPC API":
     server.stop()
     server.close()
     waitfor node.stop()
+  
+  asyncTest "relay_polling": 
+    let
+      nodeKey1 = crypto.PrivateKey.random(Secp256k1, rng[])[]
+      node1 = WakuNode.init(nodeKey1, bindIp, Port(60000))
+      nodeKey2 = crypto.PrivateKey.random(Secp256k1, rng[])[]
+      node2 = WakuNode.init(nodeKey2, bindIp, Port(60002))
+      nodeKey3 = crypto.PrivateKey.random(Secp256k1, rng[])[]
+      node3 = WakuNode.init(nodeKey3, bindIp, Port(60003), some(extIp), some(port))
+      pubSubTopic = "polling"
+      contentTopic = ContentTopic(1)
+      payload = @[byte 9]
+      message = WakuMessage(payload: payload, contentTopic: contentTopic)
+
+    await node1.start()
+    await node1.mountRelay(@[pubSubTopic])
+
+    await node2.start()
+    await node2.mountRelay(@[pubSubTopic])
+
+    await node3.start()
+    await node3.mountRelay(@[pubSubTopic])
+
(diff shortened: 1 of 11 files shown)


tests/v2/test_jsonrpc_waku.nim
```diff
@@ -106,6 +106,77 @@ procSuite "Waku v2 JSON-RPC API":
...
+      server = newRpcHttpServer([ta])
+    
+    # Let's connect to node 3 via the API
+    installRelayApiHandlers(node3, server)
+    server.start()
+
+    let client = newRpcHttpClient()
+    await client.connect("127.0.0.1", rpcPort)
+
+    echo "Now try to subscribe using API"
```
oskarth, line 148: Maybe DEBUG?

tests/v2/test_jsonrpc_waku.nim
```diff
@@ -106,6 +106,77 @@ procSuite "Waku v2 JSON-RPC API":
     server.stop()
     server.close()
     waitfor node.stop()
+  
+  asyncTest "relay_polling":
```
oskarth, line 110: Can we have a more descriptive name of what this is testing?

waku/v2/node/jsonrpc/relay_api.nim
```diff
@@ -1,12 +1,32 @@
...
   ../wakunode2,
   ./jsonrpc_types, ./jsonrpc_utils
 
+const futTimeout = 5.seconds
+
 proc installRelayApiHandlers*(node: WakuNode, rpcsrv: RpcServer) =
-  const futTimeout = 5.seconds
+  ## Create a per-topic message cache
+  var
+    topicCache = initTable[string, seq[WakuMessage]]()
```
oskarth, line 17: Hm, does this state belong in JSONRPC code? I suppose that's the only place it is currently used. Seems a bit odd to me though, with this local state? Not sure.

oskarth (approved):
Generally looks good, some comments and questions in line

tests/v2/test_jsonrpc_waku.nim
jm-clius: Ah, missed this. Thanks. Should certainly be removed.

tests/v2/test_jsonrpc_waku.nim
jm-clius: Will do!

waku/v2/node/jsonrpc/relay_api.nim
jm-clius: Good question. After considering a couple of options I settled on this being the _safest_ place for the state, where it is as short- or long-lived as the RPC handlers and will be garbage collected with everything else. Still, I will definitely consider a better (more global) place for this, though asynchronous read/write access to the table may become tricky.

Issue #146, opened 2020-09-10
Refactor: Waku v2: Better error handling


oskarth:
# Problem

In a lot of places we have `discard ...` statements for  things that can throw errors etc. We should catch these.

One of the reasons this wasn't done is due to lazyness/PoC Spike, but the other is that it wasn't clear to me what an idiomatic solution looks like. Based on https://github.com/status-im/nim-libp2p/pull/345 it seems it is slightly clearer how this should be dealt with.

## Acceptance criteria

Replace naive discard statements that can fail (e.g. async rpc/network calls etc) with await, try catch and log error etc. Let's follow the standards developing in nim-libp2p, Nimbus etc.

cc @kdeme

kdeme:
Yes, `discard` and `AsyncCheck` should not be used for launching an `async` call. Unless you are really sure that the call cannot raise exceptions (However, you can't enforce this with the compiler right now as `raises[...]` doesn't work with `async`). In the past `traceAsyncErrors` was often used, but this is likely not the best approach as it allows you to not "care" about the exceptions a call can have and also likely miss them as logging happens only on trace.
And nim-chronos now has `asyncSpawn` (see also the PR you link) which, if I understand correctly, requires you to make sure the call will not raise any Exceptions. And if it does, you get a Defect instead. So it is sort of enforcing you (eventually) to have no exceptions, being it at runtime ;-)

oskarth:
Better issues capture this

jm-clius:
This has been addressed in https://github.com/status-im/nim-waku/issues/639. Closing.

PR #313, opened 2020-12-01
Await WakuRelay publish on node

jm-clius:
Relates to #146 and #299 

Rather `await` underlying GossipSub `publish` than `discard` it. This allows API calls to also `await` a result and raise an exception on error.


oskarth (approved):
LGTM

What about adding a oneline section to CHANGELOG with ## Next header? For next release

PR #314, opened 2020-12-03
Added content topic polling on FilterAPI

jm-clius:
A further incremental PR towards #299

This adds a simple way to poll a content topic for new messages on the Filter API. Messages are cached in-memory and cleared after every call.

#### What is still outstanding for polling?
- **Must have**: a proper way to limit cache size to prevent memory leaks
- **Could have**: a limit on the number of messages returned per call
- **Could have**: moving the message cache from RPC server setup code to a better place. Perhaps the node itself?

#### What is still outstanding for the broader API?
- **Should have**: Admin API
- **Should have**: Private API

### DIFF
--- a/tests/v2/test_jsonrpc_waku.nim
+++ b/tests/v2/test_jsonrpc_waku.nim
@@ -1,5 +1,5 @@
 import
-  std/[unittest, options, sets, tables, os, strutils],
+  std/[unittest, options, sets, tables, os, strutils, sequtils],
   stew/shims/net as stewNet,
   json_rpc/[rpcserver, rpcclient],
   libp2p/standard_setup,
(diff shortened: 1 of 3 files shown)


tests/v2/test_jsonrpc_waku.nim
```diff
@@ -281,3 +282,64 @@ procSuite "Waku v2 JSON-RPC API":
...
+    await client.connect("127.0.0.1", rpcPort)
+
+    # First ensure subscription exists
+
+    let sub = await client.post_waku_v2_filter_v1_subscription(contentFilters = @[ContentFilter(topics: @[cTopic])], topic = some(defaultTopic))
+    check:
+      sub
+
+    # Now prime the node with some messages before tests
+    var
```
oskarth, line 312: If we do this in every test it make make sense to pull out into a util function, to make tests a tad less verbose. This is fine too though

oskarth (approved):

PR #317, opened 2020-12-04
Limit cache size on Filter API and Relay API

jm-clius:
This PR adds minor improvements for #299, specifically the Filter and Relay APIs:
- limit number of messages cacheable per topic/content topic
- modify a copy rather than the underlying data structure (this split should make it easier to eventually move the cache somewhere else, or even make it global)

### DIFF
--- a/tests/v2/test_jsonrpc_waku.nim
+++ b/tests/v2/test_jsonrpc_waku.nim
@@ -334,12 +334,28 @@ procSuite "Waku v2 JSON-RPC API":
       response.allIt(it.contentTopic == cTopic)
 
     # No new messages
-    
     response = await client.get_waku_v2_filter_v1_messages(cTopic)
 
     check:
       response.len() == 0
-      
+    
+    # Now ensure that no more than the preset max messages can be cached
+
+    let maxSize = filter_api.maxCache
+
+    for x in 1..(maxSize + 1):
+      # Try to cache 1 more than maximum allowed
+      filters.notify(WakuMessage(payload: @[byte x], contentTopic: cTopic), requestId)
+
+    response = await client.get_waku_v2_filter_v1_messages(cTopic)
+    check:
+      # Max messages has not been exceeded
+      response.len == maxSize
+      response.allIt(it.contentTopic == cTopic)
+      # Check that oldest item has been removed
+      response[0].payload == @[byte 2]
+      response[maxSize - 1].payload == @[byte (maxSize + 1)]
+
     server.stop()
     server.close()
     waitfor node.stop()
--- a/waku/v2/node/jsonrpc/filter_api.nim
+++ b/waku/v2/node/jsonrpc/filter_api.nim
@@ -7,7 +7,8 @@ import
   ../../waku_types,
   ../wakunode2
 
-const futTimeout = 5.seconds
+const futTimeout* = 5.seconds # Max time to wait for futures
+const maxCache* = 100 # Max number of messages cached per topic @TODO make this configurable
 
 proc installFilterApiHandlers*(node: WakuNode, rpcsrv: RpcServer) =
   ## Create a message cache indexed on content topic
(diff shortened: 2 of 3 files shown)


waku/v2/node/jsonrpc/filter_api.nim
```diff
@@ -16,10 +17,20 @@ proc installFilterApiHandlers*(node: WakuNode, rpcsrv: RpcServer) =
...
   
   proc filterHandler(msg: WakuMessage) {.gcsafe, closure.} =
-    debug "WakuMessage received", msg=msg
     # Add message to current cache
-    # @TODO limit max content topics and messages
-    messageCache.mgetOrPut(msg.contentTopic, @[]).add(msg)
+    trace "WakuMessage received", msg=msg
+    
+    # Make a copy of msgs for this topic to modify
+    var msgs = messageCache.getOrDefault(msg.contentTopic, @[])
```
staheri14, line 24: Why not changing `messageCache` directly without copying?

staheri14 (commented):
Looks good!

waku/v2/node/jsonrpc/filter_api.nim
```diff
@@ -16,10 +17,20 @@ proc installFilterApiHandlers*(node: WakuNode, rpcsrv: RpcServer) =
     messageCache = initTable[ContentTopic, seq[WakuMessage]]()
```
oskarth, line 17: Might want to pull out the type above. Also, here are some other LRU cache stuff that might be interesting https://github.com/status-im/nimbus-eth2/pull/1007/files and https://github.com/status-im/nim-libp2p/pull/360 in case you haven't seen them. I think the current implementation is fine for now, but just to be aware of

waku/v2/node/jsonrpc/filter_api.nim
```diff
@@ -16,10 +17,20 @@ proc installFilterApiHandlers*(node: WakuNode, rpcsrv: RpcServer) =
...
-    # @TODO limit max content topics and messages
-    messageCache.mgetOrPut(msg.contentTopic, @[]).add(msg)
+    trace "WakuMessage received", msg=msg
+    
+    # Make a copy of msgs for this topic to modify
+    var msgs = messageCache.getOrDefault(msg.contentTopic, @[])
+
+    if msgs.len >= maxCache:
+      # Message cache on this topic exceeds maximum. Delete oldest.
+      msgs.delete(0,0)
```
oskarth, line 28: This seems like it'd be hit quite frequently with a lot of messages coming in. How performant is this if you receive N messages a second and there's a lot of churn? Probably worth profiling first and seeing if it s a hotspot, but might be worth adding a comment here.

waku/v2/node/jsonrpc/filter_api.nim
jm-clius: This makes it a bit easier to avoid concurrent modification errors if `messageCache` can be written to/read from multiple threads (say a `handler` inserting new messages while the cache is being read). No real benefit as it stands, but in future I would like to move `messageCache` to a more sensible place where it could theoretically be accessible from other threads.

tests/v2/test_jsonrpc_waku.nim
```diff
@@ -334,12 +334,28 @@ procSuite "Waku v2 JSON-RPC API":
...
 
     check:
       response.len() == 0
-      
+    
+    # Now ensure that no more than the preset max messages can be cached
+
+    let maxSize = filter_api.maxCache
+
+    for x in 1..(maxSize + 1):
```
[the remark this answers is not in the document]
jm-clius: TBH it was purely to increase (slightly) the readability of the test purpose, i.e. "inserting one more than the allowed `maxSize`" vs "inserting up to the `maxSize` index"

waku/v2/node/jsonrpc/filter_api.nim
jm-clius: Thanks, will do and will take a look. At first glance seems very relevant to what I want to achieve.

Is pulling out the type purely to increase readability/maintainability?

waku/v2/node/jsonrpc/filter_api.nim
jm-clius: Will have to profile, but this is probably very inefficient if hit frequently. I will add the comment and consider ways of improving. As I envisioned it, the cache should eventually be dimensioned such that reaching the max size between polling calls would be an infrequent "error" condition. But there is of course no reason to believe clients would always implement it this way.

jm-clius: One way to address this would be to have a timed thread that sweeps the map every minute or so, but that would open the "concurrent modification" can of worms. For now I've added comment in da7df23 to profile performance.

waku/v2/node/jsonrpc/filter_api.nim
jm-clius: Done in da7df23

oskarth: Mostly intuition in terms of where there might be some overlap in the future, i.e. highly likely we'll want to have a cache in other places. If it is a separate type it might be easier to spot/reason about/abstract later (e.g. using efficient and tested LRU impl later on). Not a must though.

jm-clius: Makes sense. Thanks.

oskarth (approved):

PR #321, opened 2020-12-18
Feature/add private api

### LINKED ISSUE #300: Waku v2 JSON-RPC REST API: Private API
### Description

Implement Private API functionality for the Waku v2 JSON-RPC REST API. This is a subtask of the wider API implementation described in #259. Parallel issue to track spec write-up in vacp2p/specs#233

### Acceptance criteria

- [ ] API provides some functionality to generate and use `symkeys` as for Waku v1. v1-like `KeyStorage` is an option, but should preferably be avoided.

- [ ] Client should have option to use API with "version 1"-encoded WakuMessage payloads. Internally the `version` field should be set to 1 for each message. API should provide encode/decode ability.

jm-clius:
This PR closes #300

It introduces a Private API to allow symmetric and asymmetric encryption/decryption of message payloads and thus provide backwards compatibility with Waku v1.

#### What is still outstanding i.t.o. API work:
- this implementation splits `symmetric` from `asymmetric` endpoints in order to avoid ambiguity in client calls (from the client's perspective a `symkey` looks similar to a `privateKey` or `publicKey`). This is not yet reflected in the spec. Once this PR is approved I will update the spec document accordingly.
- none of the API handlers have yet been integrated with a `wakunode2`. I am busy working on a basic API tutorial and will add this, together with API integration, in a subsequent PR.
- the unit test file for the different API's have become large and unwieldy. This is because each API requires a subtly different test setup. The best solution is to just split into smaller test files for each API. Will do in subsequent PR.


oskarth (approved):

tests/v2/test_jsonrpc_waku.nim
```diff
@@ -2,15 +2,17 @@ import
...
   libp2p/protobuf/minprotobuf,
   libp2p/stream/[bufferstream, connection],
   libp2p/crypto/crypto,
   libp2p/protocols/pubsub/pubsub,
   libp2p/protocols/pubsub/rpc/message,
+  ../../waku/v1/node/rpc/hexstrings,
   ../../waku/v2/waku_types,
   ../../waku/v2/node/wakunode2,
-  ../../waku/v2/node/jsonrpc/[jsonrpc_types,store_api,relay_api,debug_api,filter_api,admin_api],
+  ../../waku/v2/node/jsonrpc/[jsonrpc_types,store_api,relay_api,debug_api,filter_api,admin_api,private_api],
```
decanus, line 15: At this point it may be nice to have this multi-line of just import everything in there, I think you can do that with a `*`

decanus (changes requested):

tests/v2/test_jsonrpc_waku.nim
oskarth: Or export from one jsonrpc package I suppose

jm-clius: After experimenting with imports/exports I changed this to a multiline import and had the API modules export their types (in 577b7a7). I'll also create a separate test module for each API, which should hopefully clean things up even further.

PR #325, opened 2020-12-21
enhancement/relay-beta-bump

decanus:
This bumps the relay, matching what is now in the spec with the `StrictNoSign`


oskarth (approved):
Was this done in spec?

decanus:
Yeah the spec was already bumped @oskarth

PR #326, opened 2020-12-21
generate/docs

decanus:
This should generate docs similar to these: https://decanus.github.io/minima/


oskarth (approved):
Nice let's see what it looks like

Issue #339, opened 2021-01-05
waku_info wrong values


decanus:
# Problem

It seems as though `waku_info` returns the wrong IP for a node. We are currently creating a node with the following paramters:

```console
cmd:~ % d inspect nim-waku-v2 | jq '.[].Config.Cmd'
[
  "--store=true",
  "--relay=true",
  "--filter=true",
  "--dbpath=/data",
  "--nodekey=d4735e3a265e16eee03f59718b9b5d03019c07d8b6c51f90da3a666eec13ab35",
  "--nat=extip:134.209.139.210",
  "--log-level=info",
  "--rpc-port=8545",
  "--rpc-address=0.0.0.0",
  "--tcp-port=30303",
  "--udp-port=30303",
  "--metrics-server=True",
  "--metrics-server-port=8008",
  "--metrics-server-address=0.0.0.0"
]
```

notice the **`--nat=extip:134.209.139.210`** however upon calling the `waku_info` we get:

```console
cmd:~ % curl -s -XPOST -H "Content-type: application/json" --data "{\"method\":\"waku_info\",\"params\":[],\"id\":1}" http://127.0.0.1:8545/ | jq .
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": "/ip4/0.0.0.0/tcp/30303/p2p/16Uiu2HAmPLe7Mzm8TsYUubgCAW1aJoeFScxrLj8ppHFivPo97bUZ",
  "error": null
}
```

This causes https://fleets.status.im to show incorrect data.

# Acceptance Criteria

 - `waku_info` returns correct information.

oskarth:
This seems to come from `info` method in `wakunode2`. Did you check if the IP is logged correctly in peerInfo? See this line `let listenStr = $peerInfo.addrs[0] & "/p2p/" & $peerInfo.peerId` - it might be that 0.0.0.0 is just the first address and not the one we want to print.

jm-clius:
#### Suggested easy solution
That is exactly right, @oskarth. The `info` method returns the only announced `peerInfo` address and that is the _host_ address. Setting the host address requires configuring the `--listen-address` option for the `wakunode2`. Since this is missing from the configuration, the address defaults to 0.0.0.0.

I think the easiest solution, for now, is to modify the config by adding `--listen-address=134.209.139.210` explicitly. I'm not sure where exactly this config lives (@jakubgs, perhaps you can direct me?) or if NAT should be configured for this host at all.

#### Next steps
[...]

PR #348, opened 2021-01-14
Prefer announced addrs in listenStr

jm-clius:
This PR relates to #339.

It changes `info` to prefer the latest announced address as returned `listenStr`.

### Outstanding items:
There is currently no way to manually specify an external port while performing [NAT setup](https://github.com/status-im/nim-waku/blob/master/waku/common/utils/nat.nim#L23-L24). The current implementation assumes the external port is the same as the bind port if only an external IP is configured. Unless this was by design, we should probably allow for manual `extPort` configuration that defaults to the bind port if no port mapping is configured.

### DIFF
--- a/waku/v2/node/wakunode2.nim
+++ b/waku/v2/node/wakunode2.nim
@@ -120,7 +120,9 @@ proc init*(T: type WakuNode, nodeKey: crypto.PrivateKey,
   info "Initializing networking", hostAddress,
                                   announcedAddresses
   # XXX: Add this when we create node or start it?
-  peerInfo.addrs.add(hostAddress)
+  peerInfo.addrs.add(hostAddress) # Index 0
+  for multiaddr in announcedAddresses:
+    peerInfo.addrs.add(multiaddr) # Announced addresses in index > 0
 
   var switch = newStandardSwitch(some(nodekey), hostAddress,
     transportFlags = {ServerFlags.ReuseAddr}, rng = rng)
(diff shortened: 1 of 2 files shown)


waku/v2/node/wakunode2.nim
```diff
@@ -120,7 +120,9 @@ proc init*(T: type WakuNode, nodeKey: crypto.PrivateKey,
   info "Initializing networking", hostAddress,
                                   announcedAddresses
   # XXX: Add this when we create node or start it?
-  peerInfo.addrs.add(hostAddress)
+  peerInfo.addrs.add(hostAddress) # Index 0
+  for multiaddr in announcedAddresses:
```
staheri14, line 124: Just curious to know where does the content of `announcedAddresses` change? (except its initialization in [line 117](https://github.com/status-im/nim-waku/blob/8b0b515297a9a1871232b04b9d868e5ccc642d82/waku/v2/node/wakunode2.nim#L117))

staheri14 (approved):

waku/v2/node/wakunode2.nim
jm-clius: Yeah, it never does change. :) In fact, up to now it was never used after initialisation, similar to the [eth2 implementation](https://github.com/status-im/nimbus-eth2/blob/7d97228208db770ae5f8b37fb5f47f9f75bb5ee4/beacon_chain/eth2_network.nim#L1541-L1551) that this is based on.

oskarth:
cc @kdeme as FYI since I believe you've been looking at some of this with Discovery v5 etc

kdeme:
Yes, currently port mapping would always be one to one. Either the user does it manually, or UPnP or NAT-PMP arranges it like that (in case of a gateway/router that supports and enables any of those).

I'm reworking this for eth2, and one of the questions is indeed if we should allow a custom mapping too. I'd guess yes? But it will require adjustment of the cli option too, as it is not so convenient to add in the already quite confusing `--nat:...` option from UX perspective. Don't want to confuse the user even more by adding more options that will perhaps be used in more rare set-ups.

This is also a related issue that I'll try to fix in the same PR: https://github.com/status-im/nim-waku/issues/239

PR #349, opened 2021-01-16
Rln relay/contract dependencies

staheri14:
To enable rln-relay tests by default, for that ganache-cli has to be installed and run prior to the tests.


kdeme:
So the docker step fails, perhaps you can try with this: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idcontainer . But it looks like on the windows version of github (windows-2019) it will still be a problem (see current failure also), so you might be better off with just installing `ganache-cli` directly.

The file to adjust would be https://github.com/status-im/nim-waku/blob/master/.github/workflows/test.yml

tests/all_tests_v2.nim
```diff
@@ -11,5 +11,5 @@ import
   ./v2/test_waku_swap,
   ./v2/test_message_store,
   ./v2/test_jsonrpc_waku,
-  ./v2/test_web3 # will remove it when rln-relay tests get added
-  # TODO ./v2/test_waku_rln_relay
+  ./v2/test_web3, # will remove it when rln-relay tests get added
```
oskarth, line 14: What is this comment referring to?

oskarth (approved):

jm-clius (approved):
LGTM. Glad there's a workaround for the macos timeout issue.

tests/all_tests_v2.nim
staheri14: test_web3 is a test for the nim-web3 module, at that time I added this test we did not have any rln-relay test (to use nim-web3) so I added this small test to check if nim-web3 works. I did not remove it in this PR intentionally because I was not sure whether, in the future, the rln-relay test will hit another error or not. So I am going to wait till the end of PoC and if everything went well, I will remove the web3 test.

Issue #329, opened 2020-12-22
nim-waku integration in nim-status: potential SQLite conflict


michaelsbradleyjr:
Sorry for the length, but I want to give plenty of context so there's less chance of misunderstanding.

To date, the work to integrate nim-waku in nim-status has not proceeded quickly, or very far (yet), because e.g. the nim-status team faced a number of challenges around the decision to integrate [SQLCipher](https://github.com/sqlcipher/sqlcipher), by way of [nim-sqlcipher](https://github.com/status-im/nim-sqlcipher). Also, we're still in the early phases of building a "background task runner" so that, like status-go (which nim-status is intended to replace eventually), work done by the library doesn't interfere with the thread/s of an application using it without some effort by that application's developer/s.

The work on nim-sqlcipher and its integration in nim-status has just wrapped up, but there *may* be an additional hurdle related to a *possible* conflict with nim-waku's usage of [nim-sqlite3-abi](https://github.com/arnetheduck/nim-sqlite3-abi).

I will note that in the past it was suggested that [attention be paid](https://github.com/status-im/nim-waku/pull/247#issuecomment-719461469) to nim-sqlcipher's API design in relation to nim-sqlite3-abi (if I understood correctly). One of the challenges for the nim-status team has been the team's dearth of experience with makefiles, compilers, linkers, and all the concepts and concerns that go along with those things. In addition to slowing us down, that probably led to our underappreciating the future interplay between sqlcipher and sqlite3-abi.

Over the past weekend we realized that SQLite symbols in `libstatus.dll|dylib|so` ([status-go](https://github.com/status-im/status-go) built/linked as a [shared library](https://github.com/status-im/status-go/blob/develop/Makefile#L138)) can conflict with those of `libsqlcipher` linked to nim-status. We were hitting a strange edge case involving Windows; after realizing the nature of the conflict, we confirmed that it can affect other platforms as well, depending on the linking order. I then realized we could face a similar problem once the work to integrate nim-waku into nim-status really takes off. I did some quick experiments and here's what I found:

If `import waku/v2/node/sqlite` is added to our [`db_smoke.nim`](https://github.com/status-im/nim-status/blob/master/test/nim/db_smoke.nim) test, then compilation fails with errors like this: https://gist.github.com/michaelsbradleyjr/37a95e11925cd2338d5efdfe284cb89a.

[...]

PR #350, opened 2021-01-18
Extract MessageStore interface (avoids SQLite conflicts)

### LINKED ISSUE #334: Data Loader Interface
# Problem

Currently we have a message store, this uses SQLite to store messages during a nodes lifetime and retrieve them into memory when the node is started. However, this can lead to conflicts with the new nim-status. See: https://github.com/status-im/nim-waku/issues/329.

# Solution

To solve this problem what we should do is abstract the MessageStore in such a way, that an application using nim-waku can provide its own MessageStore which is responsible for saving and retrieving messages. The interface can be relatively simple for now and only needs 2 functions:

```nim
proc put*(db: MessageStore, cursor: Index, message: WakuMessage): MessageStoreResult[void]
proc getAll*(db: MessageStore, onData: DataProc): MessageStoreResult[bool]
```

jm-clius:
This PR closes #334 and addresses #329.

It extracts a data loader interface for a `MessageStore`. This will allow `nim-waku` clients to provide their own `MessageStore` implementations.

As such:
- the `nim-waku` `sqlite` implementation for a `MessageStore` remains the same, but is renamed to `WakuMessageStore`.
- related modules moved to a `message_store` subfolder

(Cleaning up imports as I go along.)


staheri14 (approved):
looks good!

michaelsbradleyjr:
Looks good. I'm going to do some experiments just to make sure that with these changes the C symbols no longer collide, will report back shortly.

michaelsbradleyjr:
No collisions, nice work.

michaelsbradleyjr (approved):

Issue #342, opened 2021-01-07
Fix broken network simulation


oskarth:
# Problem

Currently the network simulation fails. E.g. (from Waku v2 README)

```
./build/start_network2 --topology:FullMesh --amount:6 --test-node-peers:2
# In another shell run
./build/quicksim2
```

## Notes

```
│INF 2021-01-07 11:46:52.560+08:00 Listening on                               topics="wakunode" tid=9820 file=wakunode2.nim:154 full=/ip4/0.0.0.0/tcp/600│
t.│03/p2p/16Uiu2HAmPDRAL8rfFdJtfLcBbzTGCjUnqUn9jeNH97jjh7YrJXM5                                                                                            │pt
Er│DBG 2021-01-07 11:46:52.560+08:00 initWakuRelay                              topics="wakurelay" tid=9820 file=waku_relay.nim:38                         │
] │DBG 2021-01-07 11:46:52.560+08:00 init                                       topics="wakurelay" tid=9820 file=waku_relay.nim:23                         │r]
[n│DBG 2021-01-07 11:46:52.560+08:00 init                                       topics="wakurelay" tid=9820 file=waku_relay.nim:23                         │
--│TRC 2021-01-07 11:46:52.560+08:00 registering protocols                      topics="multistream" tid=9820 file=multistream.nim:183 protos="@[\"/vac/wak│--
  │u/relay/2.0.0-beta2\", \"/vac/waku/relay/2.0.0-beta2\"]"                                                                                                │
02│INF 2021-01-07 11:46:52.560+08:00 mounting relay                             topics="wakunode" tid=9820 file=wakunode2.nim:312                          │43
t.│DBG 2021-01-07 11:46:52.560+08:00 subscribe                                  topics="wakurelay" tid=9820 file=waku_relay.nim:53 pubSubTopic=/waku/2/defa│pt
/h│ult-waku/proto                                                                                                                                          │
t.│TRC 2021-01-07 11:46:52.560+08:00 subscribing to topic                       topics="pubsub" tid=9820 file=pubsub.nim:265 name=/waku/2/default-waku/prot│pt
Er│o                                                                                                                                                       │
] │TRC 2021-01-07 11:46:52.560+08:00 rebalancing mesh                           topics="gossipsub" tid=9820 file=gossipsub.nim:405 mesh=0 gossipsub=0 topic│r]
[n│=/waku/2/default-waku/proto                                                                                                                             │
[...]

PR #351, opened 2021-01-20
Fix network simulation and remote peer ID CLI argument parsing

jm-clius:
This PR addresses #342

It contains:
1. Following similar restrictions in `nim-libp2p`, `wakunode2` now applies strict formatting to the `PeerInfo` multiaddresses obtained from parsing CLI arguments. Specifically, `PeerInfo` gets initialised with a `peerId` string and a `multiaddr` wire address. This fixes the `./build/start_network2` network simulation.
2. `./build/chat2` example fixed in a similar manner to (1).
3. Updated `quicksim2` from using the old RPC methods to the new JSON-RPC API.

(1) and (3) have the combined effect of fixing the [quick network simulation](https://github.com/status-im/nim-waku/tree/master/waku/v2#waku-quick-simulation)


examples/v2/chat2.nim
```diff
@@ -78,8 +78,27 @@ proc initAddress(T: type MultiAddress, str: string): T =
 
 proc parsePeer(address: string): PeerInfo =
```
staheri14, line 79: Does it have any unit test? if not, might be good to have it tested

staheri14 (approved):

examples/v2/chat2.nim
jm-clius: Agree! This is quite a central part of `wakunode2` connectivity. Will add.

jm-clius: Done in 105bbd4.

Also extracted this functionality to a utils module. Cleaning up imports as I go along.

tests/v2/test_wakunode.nim
```diff
@@ -251,3 +254,16 @@ procSuite "WakuNode":
     await node1.stop()
     await node2.stop()
     await node3.stop()
+
+  asyncTest "Peer info parses correctly":
+    ## This is such an important utility function for wakunode2
+    ## that it deserves its own test :)
```
staheri14, line 260: nice comment! :D

waku/v2/node/wakunode2.nim
```diff
@@ -14,6 +14,7 @@ import
   ../protocol/waku_store/waku_store,
   ../protocol/waku_swap/waku_swap,
   ../protocol/waku_filter/waku_filter,
+  ../utils/peers,
```
staheri14, line 17: a minor comment, do we follow any naming conventions to distinguish utils modules from others? e.g., peers_utils or something like this

staheri14 (approved):
Looks nice!

tests/v2/test_wakunode.nim
```diff
@@ -251,3 +254,16 @@ procSuite "WakuNode":
...
+    
+    let
+      addrStr = "/ip4/127.0.0.1/tcp/60002/p2p/16Uuu2HBmAcHvhLqQKwSSbX6BG5JLWUDRcaLVrehUVqpw7fz1hbYc"
```
staheri14, line 263: Just came to my mid, you may want to consider some corner cases, with wrong formats (in case such situation may happen), otherwise, it is fine

waku/v2/node/wakunode2.nim
jm-clius: As far as I can tell, the (current) convention is simply to group these under a `utils` subfolder. I think there's scope to improve how we handle utils in general, so watch this space. :)

tests/v2/test_wakunode.nim
jm-clius: Done :) 4b5094d

oskarth (approved):

Issue #324, opened 2020-12-21
Waku V2 Metrics


decanus:
# Problem

We currently have a metrics [dashboard for Waku V2](https://grafana.infra.status.im/d/qrp_ZCTGz/nim-waku-v2?orgId=1&refresh=5m), this was recently created as part of the scope of issue #256. However, it seems as though the metrics are not correct and there is information missing that could be deemed important.

# Acceptance Criteria

 - The metrics no longer show `0` for all libp2p related metrics, this is incorrect unless the cluster was not set up correctly.
 - We add relevant metrics that could aid in debugging, for example the amount of error messages and warnings present in a nodes logs.

oskarth:
What do we think are relevant metrics? Pubsub peers seems useful, e.g.

oskarth:
Ping on this @decanus

jm-clius:
The following observations:
1. Most metrics (except for `total_messages`, see below) work correctly if running a local `wakunode2`, Prometheus and Grafana server, which suggests that this is related to the cluster setup. This will likely be fixed by https://github.com/status-im/infra-nim-waku/pull/3 and https://github.com/status-im/nim-waku/pull/348.
2. However, the only metric defined for `wakunode2`, namely `total_messages`, is declared but never set/incremented in the code. I will fix this.
3. We should define at least some metrics related to `filter`, `store`, `swap` functionality, debugging etc.
4. The grafana dashboards under `/metrics/`, and documented [here](https://github.com/status-im/nim-waku/tree/master/waku/v2#using-metrics), are outdated and should be updated for Waku v2.

PR #356, opened 2021-01-28
Added waku metrics

jm-clius:
This PR addresses #324 

It adds a number of metrics for `wakunode2` and related protocols.

1. All waku-related metric names are prefixed `waku_` to make group queries easier (and sorting on Prometheus).
2. Message counters are labelled according to protocol ("relay", "filter" and "store"). For debugging, we could consider adding topic/content topic labels. If the number of topics grow excessively large, however, there will be a corresponding label explosion, so I've avoided this approach for now.
3. There is a single `error` counter per protocol, with different labels for different error types. It's easier to build dashboards/query metrics this way, though care must of course be taken when defining new error types to avoid a label explosion.
4. I've also added a number of gauges to keep track of peers/subscriptions/filters/etc for each `waku` protocol. Under the hood, `libp2p` already provides enough GossipSub/Pubsub metrics for visibility into `WakuRelay` (including PubSub peers, subscribed topics, etc.).

Updates to the grafana dashboards to follow.

### DIFF
--- a/waku/v2/node/wakunode2.nim
+++ b/waku/v2/node/wakunode2.nim
@@ -1,6 +1,6 @@
 import
   std/[options, tables, strutils, sequtils],
-  chronos, chronicles, stew/shims/net as stewNet,
+  chronos, chronicles, metrics, stew/shims/net as stewNet,
   # TODO: Why do we need eth keys?
   eth/keys,
   libp2p/multiaddress,
@@ -17,6 +17,10 @@ import
   ../utils/peers,
   ./message_store/message_store,
   ../utils/requests
+
+declarePublicCounter waku_node_messages, "number of messages received", ["type"]
+declarePublicGauge waku_node_filters, "number of content filter subscriptions"
+declarePublicGauge waku_node_errors, "number of wakunode errors", ["type"]
 
 logScope:
   topics = "wakunode"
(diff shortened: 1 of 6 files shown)


waku/v2/node/wakunode2.nim
```diff
@@ -433,8 +443,11 @@ when isMainModule:
       {.gcsafe.}:
         # TODO: libp2p_pubsub_peers is not public, so we need to make this either
         # public in libp2p or do our own peer counting after all.
-        let
-          totalMessages = total_messages.value
+        var
+          totalMessages = 0.float64
```
oskarth, line 447: How come total messages is a float?

oskarth (approved):
LGTM

waku/v2/protocol/waku_filter/waku_filter.nim
```diff
@@ -188,6 +195,7 @@ proc init*(T: type WakuFilter, switch: Switch, rng: ref BrHmacDrbgContext, handl
 # @TODO THIS SHOULD PROBABLY BE AN ADD FUNCTION AND APPEND THE PEER TO AN ARRAY
 proc setPeer*(wf: WakuFilter, peer: PeerInfo) =
   wf.peers.add(FilterPeer(peerInfo: peer))
+  waku_filter_peers.inc()
```
oskarth, line 198: What happens if a node disconnects? Does it get decremented somewhere?

waku/v2/node/wakunode2.nim
jm-clius: Metric values are stored as `float64` and `totalMessages` just sums over these `float64` values.

waku/v2/protocol/waku_filter/waku_filter.nim
jm-clius: Good point. Currently this gauge just tracks the number of filter `peers` in the `seq` and, since we don't remove any disconnected peers (yet), the metric is never decremented. It does make sense, however, to at least reset this gauge to `0` when the `wakunode` is initialised. Will add.

PR #357, opened 2021-01-29
Updated grafana dashboard to show Waku v2 metrics

jm-clius:
Relates to #324 

Updates the grafana dashboard to show Waku v2 metrics.

Since the JSON diff is difficult to interpret, an example of what the updated dashboard looks like below. Many metrics aren't showing as they were simply not logged by the simulation I was running, but it should illustrate the general idea.

image


staheri14 (approved):
Looks nice!

oskarth (approved):

PR #359, opened 2021-01-29
Rln relay/membership contract

staheri14:
Closes https://github.com/status-im/nim-waku/issues/353 and https://github.com/status-im/nim-waku/issues/323 
The summary of changes:
1- Enables passing input the a contract constructor, this is done as part of test_utils module, further comments and instructions on how to prepare constructor's input are provided
2- Adds two .txt files which include the byte-codes of Poseidon hasher and the membership contract (did not hardcode bytecodes in the test file because they were huge and would cause scrolling issues)
3- Updates the text_waku_rln_relay.nim by adding the new membership contract interface and testing the new methods. The updated test also can send inputs to the contract constructor.


Makefile
```diff
@@ -124,6 +125,9 @@ bridge: | build deps
 # Builds and run the test suite (Waku v1 + v2)
 test: | test1 test2
 
+installganache:
```
oskarth, line 128: As mentioned in call, feel free to make this optional at PoC stage with 1) a comment on problem 2) issue to fix it

tests/v2/test_waku_rln_relay.nim
```diff
@@ -2,99 +2,141 @@ import
...
+# the address of Ethereum client (ganache-cli for now)
+const ethClient = "ws://localhost:8541/"
+
+const poseidonHasherCode = readFile("tests/v2/poseidonHasher.txt")
```
oskarth, line 9: Can we add a link to how this contract code was found? Where is the canonical source of it? A few lines of comments here would be good

Makefile
oskarth: 3) Comment in README on Ganache pre-req and how to run this test

staheri14: The following PR https://github.com/status-im/nim-waku/pull/349 automates the process of ganache installation and launch, as soon as it gets merged into the master branch,  there will be no need to install ganache manually

oskarth (approved):
LGTM

Issue #358, opened 2021-01-29
PoC: Waku peer management using libp2p peer store


jm-clius:
### Problem

As a first step towards peer management, `nim-libp2p` recently added a [peer store](https://github.com/status-im/nim-libp2p/issues/504) to keep track of the addresses, keys and registered protocols of known peers in separate maps (termed `books`). Clients can instantiate and populate a store in a variety of methods, but the aim is to eventually have a single libp2p peer store instance with integration on ad-hoc dialing, peer discovery, peer identification, etc.

For `wakunode2`, we can create a rudimentary peer manager by abstracting peer dialing from the switch. In other words, the manager acts as intermediary between the `wakunode2` and the libp2p `switch`, adds dialed peers to a `peer store`, keeps track of their connected state and attempts a redial on disconnect. In its simplest form, this can serve as a PoC of peer management using a peer store.

### Acceptance criteria

- [x] Manages `wakunode2` peers passed as cli arguments.
- [x] Keeps track of dialed peers in a peer store.
- [x] Keeps track of the connected state of peers. UPDATE: ~Attempts reconnect for staticnodes (`WakuRelay`) on disconnect (perhaps adding a back-off table in case of failures)~. Reconnection is the type of advanced functionality that falls outside the scope of the PoC. This should rather be addressed in the [peer and connection management](https://github.com/status-im/nim-libp2p/issues/537) effort in `nim-libp2p`
- [x] Provides (`admin`) access to the internal peer list in the API. Extend JSON-RPC `Admin API` to do the same.

### Related issues

https://github.com/status-im/nim-waku/issues/265, https://github.com/status-im/nim-waku/issues/266

jm-clius:
Closing issue as PoC peer and connection management added for all Waku protocols. Extended [peer and connection management](https://github.com/status-im/nim-libp2p/issues/537) effort ongoing in `nim-libp2p`.

PR #364, opened 2021-02-03
Added basic peer manager

jm-clius:
This is an incremental PR towards #358

It introduces a basic peer manager for `wakunode2`. For now the manager:
- abstracts the dialing of `WakuRelay` peers
- keeps track of dialed peers in a peer store
- provides an API to retrieve combined info from both the `peerStore` and `switch` about peers and their connectedness

### Suggested next steps:

1. general improvements 
2. use peer manager to present information about `relay` peers on the JSON-RPC Admin API
3. keep track locally of waku-related metadata for peers (incl. their connection status and behaviour). For this waku must implement its own extended peer store to add homegrown metadata types. Currently the manager just uses the "standard" peer store and assumes all peers must remain connected, as in a peer pool.
4. extend manager to be used for other Waku v2 protocols: `filter`, `store` and `swap` (requires _**3.**_)

### Future work:

1. Strategy to attempt redialing disconnected peers
2. Peer scoring mechanism
3. Peer store persistence
4. Peer discovery


staheri14 (approved):
Nice and clear comments on the code!

waku/v2/node/peer_manager.nim
```diff
@@ -0,0 +1,67 @@
...
+# Dialer interface #
+####################
+
+proc dialPeer*(pm: PeerManager, peerInfo: PeerInfo, proto: string): Future[Connection] {.async.} =
+  # Dial a given peer and add it to the list of known peers
+  # @TODO check peer validity, duplicates and score before continuing. Limit number of peers to be managed.
+  
+  # First add dialed peer info to peer store...
+
+  debug "Adding dialed peer to manager", peerId = peerInfo.peerId, addr = peerInfo.addrs[0], proto = proto
```
oskarth, line 27: Might want to set a topic as log scope

waku/v2/node/peer_manager.nim
```diff
@@ -0,0 +1,67 @@
...
+
+proc new*(T: type PeerManager, switch: Switch): PeerManager =
+  T(switch: switch,
+    peerStore: PeerStore.new())
+
+####################
+# Dialer interface #
+####################
+
+proc dialPeer*(pm: PeerManager, peerInfo: PeerInfo, proto: string): Future[Connection] {.async.} =
```
oskarth, line 21: What happens if this call fails?

oskarth (approved):
In general looks good, some comments and suggestions

waku/v2/node/peer_manager.nim
jm-clius: Will do in subsequent PR.

waku/v2/node/peer_manager.nim
jm-clius: Currently still a very ungraceful unhandled `DialFailedError`. The plan is to timeout the `dial` future on the `switch`, log these failed/timed-out connections and keep track of such peers (e.g. changing their `connectedness` state to `CannotConnect` for a period of time). I'm also considering adding at least one redial attempt per failure, if there is a sensible way of doing this.

PR #368, opened 2021-02-05
Peer manager: improvements; waku_filter integration

jm-clius:

It adds the following to the Waku peer manager:
- duplicate handling (a requirement for `filter`, `store` and `swap` peers that are dialed on an ad-hoc basis)
- first step towards integration with `filter` protocol. Since the peer manager includes basic error handling for dialing failures, this has the side effect of improving error handling for the `filter` protocol.

#### Suggested next steps:
1. Peer selection per protocol in peer manager (eventually using scoring mechanisms to select most appropriate peer).
2. Integrate manager with other Waku v2 protocols: `store` and `swap`
3. Remove local `set` of `peers` for `filter`, `store` and `swap` protocols (i.e. use only `peer manager` for peer storing and selection)
4. Keep track of waku-related metadata for peers (incl. their connection status and behaviour) in an extended peer store.

### DIFF
--- a/tests/v2/test_waku_filter.nim
+++ b/tests/v2/test_waku_filter.nim
@@ -8,7 +8,8 @@ import
   libp2p/stream/[bufferstream, connection],
   libp2p/crypto/crypto,
   libp2p/multistream,
-  ../../waku/v2/protocol/[message_notifier],
+  ../../waku/v2/node/peer_manager,
+  ../../waku/v2/protocol/message_notifier,
   ../../waku/v2/protocol/waku_filter/waku_filter,
   ../test_helpers, ./utils
 
@@ -37,7 +38,7 @@ procSuite "Waku Filter":
       responseRequestIdFuture.complete(requestId)
 
     let
-      proto = WakuFilter.init(dialSwitch, crypto.newRng(), handle)
+      proto = WakuFilter.init(PeerManager.new(dialSwitch), crypto.newRng(), handle)
       rpc = FilterRequest(contentFilters: @[ContentFilter(topics: @[contentTopic])], topic: defaultTopic, subscribe: true)
 
     dialSwitch.mount(proto)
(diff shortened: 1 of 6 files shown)


waku/v2/protocol/waku_filter/waku_filter.nim
```diff
@@ -208,26 +208,47 @@ proc subscription*(proto: WakuFilter): MessageNotificationSubscription =
...
-          await conn.writeLP(push.encode().buffer)
+          
+          let connOpt = await proto.peerManager.dialPeer(subscriber.peer, WakuFilterCodec)
+
+          if connOpt.isSome:
+            await connOpt.get().writeLP(push.encode().buffer)
+          else:
+            # @TODO more sophisticated error handling here
+            error "failed to push messages to remote peer"
+            waku_filter_errors.inc(labelValues = ["dial_failure"])
```
staheri14, line 219: A suggestion, you may define failure messages like "dial_failure" as constant variables  and then keep using those variables in the code.

tests/v2/test_waku_filter.nim
```diff
@@ -128,3 +129,30 @@ procSuite "Waku Filter":
...
       (await responseCompletionFuture.withTimeout(5.seconds)) == false
+  
+  asyncTest "handle filter subscribe failures":
+    const defaultTopic = "/waku/2/default-waku/proto"
+
+    let
+      key = PrivateKey.random(ECDSA, rng[]).get()
+      peer = PeerInfo.init(key)
+      contentTopic = ContentTopic(1)
+      post = WakuMessage(payload: @[byte 1, 2, 3], contentTopic: contentTopic)
```
staheri14, line 140: I think `key`, `peer` and `post` are not used in the test, you mat remove them.

waku/v2/protocol/waku_filter/waku_filter_types.nim
```diff
@@ -45,7 +46,7 @@ type
 
   WakuFilter* = ref object of LPProtocol
     rng*: ref BrHmacDrbgContext
-    switch*: Switch
+    peerManager*: PeerManager
```
oskarth, line 49: It'd be good if we made this change consistently in other protocols, though I assume that's the plan. Also worth notifying Stimbus so they are aware of this change.

Oh, and CHANGELOG.md please!

oskarth (approved):

waku/v2/protocol/waku_filter/waku_filter_types.nim
jm-clius: Yep, that is indeed the plan. Will inform Stimbus as well.

Oh, and of course - CHANGELOG! 🤦 Done in 4e7f068

tests/v2/test_waku_filter.nim
jm-clius: Thanks, Sanaz. Done in 4e7f068

waku/v2/protocol/waku_filter/waku_filter.nim
jm-clius: Thanks, good idea. I generally hate magic variables and this way it's easier to keep track of them in one place and avoid errors. Done in 4e7f068

PR #373, opened 2021-02-10
Nim wrappers for Rln lib (Key and Params generation) 

staheri14:
This PR addresses https://github.com/status-im/nim-waku/issues/297
It contains the wrappers for rln lib as well as their unit tests. 
The code changes and tests are non effective, as they rely on the rlnlib.dylib which is currently not part of the nim-waku repo (to see the result of the unit tests,  the rlnlib  must be placed in the the following directory waku/v2/protocol/waku_rln_relay/ and then the test_rln_relay_wrappers.nim must be compiled and run). 
Will follow up this PR with another issue for the integration of rlnlib.dylib (the integration should be done through Makefile).


tests/v2/test_rln_relay_wrappers.nim
```diff
@@ -0,0 +1,49 @@
...
+    check:
+      # check whether the keys are generated successfully
+      done == true
+
+    if done:
+      var generatedKeys = cast[ptr array[64, byte]](keysBufferPtr.`ptr`)[]
+      check:
+        # the public and secret keys together are 64 bytes
+        generatedKeys.len == 64
+      echo "generated keys: ", generatedKeys
```
jm-clius, line 48: Should this be writing to `stdout` or rather a `debug` log?

jm-clius (approved):
LGTM

tests/v2/test_rln_relay_wrappers.nim
staheri14: Thanks! Changed it to `debug`.

PR #375, opened 2021-02-10
Removed local peer sets for filter, swap and store

jm-clius:

It removes the local `set` of `peers` for `filter`, `store` and `swap` protocols and uses the common `peer store` instead

#### Suggested next steps:
1. More sophisticated peer selection per protocol in peer manager (eventually using scoring mechanisms to select most appropriate peer).
2. Keep track of waku-related metadata for peers (incl. their connection status and behaviour) in an extended peer store.
3. Improved behaviour for disconnected peers (incl. attempted redials)
4. Peer discovery


waku/v2/protocol/waku_store/waku_store.nim
```diff
@@ -391,8 +391,14 @@ proc query*(w: WakuStore, query: HistoryQuery, handler: QueryHandlerFunc) {.asyn
   #  - latency?
   #  - default store peer?
 
-  let peer = w.peers[0]
-  let connOpt = await w.peerManager.dialPeer(peer.peerInfo, WakuStoreCodec)
+  let peerOpt = w.peerManager.selectPeer(WakuStoreCodec)
+
+  if peerOpt.isNone():
+    error "failed to connect to remote peer"
```
staheri14, line 397: Or maybe there is no peer to connect to?

staheri14 (approved):
Looks fine to me!

waku/v2/protocol/waku_store/waku_store.nim
jm-clius: Yes, thanks. That will be clearer.

PR #377, opened 2021-02-11
Feat/pm connection tracking

jm-clius:

It adds local connection tracking to the Peer Manager by extending the `libp2p` peer store. Peers' `connectedness` can now be any of the values `NotConnected`, `CannotConnect`, `CanConnect` and `Connected` (based on the [go-implementation](https://github.com/libp2p/go-libp2p-core/blob/a39b84ea2e340466d57fdb342c7d62f12957d972/network/network.go#L50-L62))

#### Suggested next steps:
1. More sophisticated peer selection per protocol in peer manager (eventually using scoring mechanisms to select most appropriate peer).
2. Improved behaviour for disconnected peers (incl. attempted redials)
3. Peer discovery


staheri14 (approved):
Looks good to me!

PR #385, opened 2021-02-19
Integrate registration into wakunode and Enable signed registration transactions

staheri14:
Addresses https://github.com/status-im/nim-waku/issues/345 and https://github.com/status-im/nim-waku/issues/384

Summary of changes:
- Adds `mountRlnRelay` proc to `wakunode2`
- Mounts the rln-relay in `wakunode2`
- Adds integration test for `mountRlnRelay`
- Renames the `RlnRelayPeer` data type to `WakuRlnRelay` for the sake of consistency with other similar structs (e.g., `WakuFilter`, etc.)
- Enables signed transactions for registration to the membership contract (due to this a new field of `ethAccountPrivateKey` is added into `WakuRlnRelay`)
- Addresses a bug for fetching rln repo, adds bash command to the Makefile to delete any existing `rln` directory before cloning the rln repo


waku/v2/node/wakunode2.nim
```diff
@@ -347,6 +369,12 @@ proc mountRelay*(node: WakuNode, topics: seq[string] = newSeq[string](), rlnRela
 
     node.subscribe(topic, handler)
 
+  if rlnRelayEnabled:
```
oskarth, line 372: Should wakuRelay be enabled in rlnRelay is?

oskarth (approved):

waku/v2/node/wakunode2.nim
staheri14: You mean the other way around? I think rln-relay is a special service for the relay protocol, that is why I have added it inside the relay, let me know whether it makes sense

staheri14: I am going to merge the PR for now, and will apply all the changes related to this comment in the subsequent PRs

oskarth: Typo, what I meant is: If rlnRelay protocol is active, should wakuRelay be active too? Or are they mutually exclusive (if-else)?

oskarth: My understanding of rlnRelay is that it wraps the wakuRelay protocol, so you wouldn't have both rlnRelay and vanilla wakuRelay, but I could be off.

staheri14: > Typo, what I meant is: If rlnRelay protocol is active, should wakuRelay be active too? Or are they mutually exclusive (if-else)?

I see your point, they are not mutually exclusive, indeed rlnRelay **CAN** be active conditioned that wakuRelay is also active i.e., rlnRelay ->wakuRelay but wakuRelay -/-> rlnRelay

staheri14: > My understanding of rlnRelay is that it wraps the wakuRelay protocol, so you wouldn't have both rlnRelay and vanilla wakuRelay, but I could be off.

I considered rln-relay as a separate module/service that can be activated on top of the relay protocol. However, in our current design, given that all the peers are treated as one group, they all have to either have rln-relay enabled or not (this is not an individual decision). But if in the future we consider let's say pubsub topic groups, then enabling rln-relay depends on the topic that a peer wants to contribute to and whether in that topic group spam protection is a concern or not

PR #400, opened 2021-03-01
ci: add rust and cargo to fix builds

jakubgs:
Fixes:
```
15:01:12  Building: Nim compiler
15:01:14  HEAD is now at a80f5d0 move member index to auth object
15:01:14  bash: cargo: command not found
```


jm-clius:
Thanks, @jakubgs

jm-clius (approved):

Issue #399, opened 2021-03-01
Dogfooding Waku v2 through mini chat app


oskarth:
# Problem

We want to dogfood Waku v2 to find any issues that might exist with it and to work out kinks. While there is ongoing work to do this with Stimbus/Desktop/App, this is taking a bit longer than expected and is quite complex. How can we dogfood Waku as early as possible?

# Suggested solutions

Luckily, we don't need the whole Status protocol or app to do basic dogfooding. We just need a basic chat interface (provided for by e.g. `chat2` protocol and a single "chat topic" to replace #waku conversations. We don't need more complex encryption features, and things like nicknames can simply be hardcoded.

# Acceptance criteria

`chat2` app good enough to replace daily use for #waku conversations. This means: can send messages and receive, and fetch messages received while gone. Also some basic instructions for how to start a node and join chat.

# Enhancements

Interop with JS client to guide Waku JS critical path.

Obviously Status desktop use but that's out of scope for this issue and repo ;)

oskarth:
Running `chat2` it seems to work but need second node to confirm. Probably need to clean up the logs to make UX a bit nicer (pipe to logfile and leave STDOUT for chat?).

oskarth:
Can confirm that `./build/chat2 --storenode:/ip4/134.209.139.210/tcp/30303/p2p/16Uiu2HAmPLe7Mzm8TsYUubgCAW1aJoeFScxrLj8ppHFivPo97bUZ --staticnode:/ip4/134.209.139.210/tcp/30303/p2p/16Uiu2HAmPLe7Mzm8TsYUubgCAW1aJoeFScxrLj8ppHFivPo97bUZ` works for three of us, including historical messages.

What do we need for this to replace current chat? Sketch:

- logs out of stdout
- commandline flag nick
- not hardcoded store/staticnodes (?)

- maybe use random of cluster node? (otherwise a bit odd we all connect to same node)
- thoughts on how the peer(s) selection should work by default here?

Discord<>Wakuv2 bridge would be nice too, e.g. using existing chat2 STDIN/OUT, or Nim JSON RPC API.

jm-clius:
Agreed, @oskarth. Off the top of my head I'd add:
1. some timestamps
2. better chat prompts

> * maybe use random of cluster node? (otherwise a bit odd we all connect to same node)
> * thoughts on how the peer(s) selection should work by default here?

[...]

PR #405, opened 2021-03-03
Further chat2 improvements

jm-clius:
This is a further incremental PR towards #399

## What's new?

This PR adds the following improvements/fixes to `chat2`:
1. Short date now also prepended to chat messages, in addition to UTC time and nickname.
2. Fixed `metrics error`.
3. Random `storenode` now auto-selected if `store` set to `true` and no `storenode` specified.
4. A basic tutorial.

In principle it should now be possible to run a `chat2` application that auto-selects both a `relay` and `store` peer from the test cluster by simply running the command
```
./build/chat2 --store:true
```
However, see existing issue re historical messages below ->

### Why are my historical messages absent or not ordered by time?

> TL;DR: Basically something strange is happening on the test cluster, likely to do with connectivity or a bug in the `store` protocol. Use `/ip4/134.209.139.210/tcp/30303/p2p/16Uiu2HAmPLe7Mzm8TsYUubgCAW1aJoeFScxrLj8ppHFivPo97bUZ` as both `staticnode` and `storenode` in the meantime.

The issues with historical messages and the test nodes acting up has proven more difficult to solve. Here's what I found so far:
1. If you connect to `/ip4/104.154.239.128/tcp/30303/p2p/16Uiu2HAmPLe7Mzm8TsYUubgCAW1aJoeFScxrLj8ppHFivPo97bUZ` as _staticnode_, historical messages won't load and you may not be able to relay messages to other peers, regardless of what `storenode` is configured. It works perfecly well as `storenode`, though, but only if `/ip4/134.209.139.210/tcp/30303/p2p/16Uiu2HAmPLe7Mzm8TsYUubgCAW1aJoeFScxrLj8ppHFivPo97bUZ` is configured as `staticnode`. :shrug:
2. If (1) is false, but you have configured any node other than `/ip4/134.209.139.210/tcp/30303/p2p/16Uiu2HAmPLe7Mzm8TsYUubgCAW1aJoeFScxrLj8ppHFivPo97bUZ` as `storenode`, historical messages will either appear in a strange order or not at all.

Given the above, it's basically a gamble if you're going to be able to download historical messages in a consistent manner if node selection is done at random. _For now_ I therefore recommend specifying _both_ the `staticnode` and `storenode` as `/ip4/134.209.139.210/tcp/30303/p2p/16Uiu2HAmPLe7Mzm8TsYUubgCAW1aJoeFScxrLj8ppHFivPo97bUZ` when starting your application, while I'm investigating the cluster issue and trying to find the bug(s) responsible.

### What's next?
1. Fixing the issue(s) with historical message retrieval.
2. Adding topic selection.

### DIFF
--- a/examples/v2/chat2.nim
+++ b/examples/v2/chat2.nim
@@ -25,11 +25,10 @@ import   ../../waku/v2/node/[config, wakunode2, waku_payload],
          ../../waku/common/utils/nat
 
 const Help = """
-  Commands: /[?|help|connect|disconnect|exit]
+  Commands: /[?|help|connect|nick]
   help: Prints this help
   connect: dials a remote peer
-  disconnect: ends current session
-  exit: closes the chat
+  nick: change nickname for current chat session
 """
 
 const
@@ -166,8 +165,8 @@ proc writeAndPrint(c: Chat) {.async.} =
     else:
       # XXX connected state problematic
       if c.started:
-        # Get message timestamp
-        let time = getTime().utc().format("'<'HH:mm'>'")
+        # Get message date and timestamp
+        let time = now().utc().format("'<'MMM' 'dd,' 'HH:mm'>'")
 
         c.publish(time & " " & c.nick & ": " & line)
         # TODO Connect to peer logic?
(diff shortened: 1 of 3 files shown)


examples/v2/chat2.nim
```diff
@@ -234,10 +233,21 @@ proc processInput(rfd: AsyncFD, rng: ref BrHmacDrbgContext) {.async.} =
...
 
-  if conf.storenode != "":
+  if (conf.storenode != "") or (conf.store == true):
     node.mountStore()
 
-    node.wakuStore.setPeer(parsePeerInfo(conf.storenode))
+    var storenode: string
+
+    if conf.storenode != "":
+      storenode = conf.storenode
```
staheri14, line 242: Not sure if there is a format checking in the `parsePeerInfo` proc,  but I'd check whether the passed address is valid and if not set the store node randomly

staheri14 (approved):
Nice!

oskarth (approved):
Good stuff!

PR #406, opened 2021-03-04
Fix Docker image - missing librln.so file

jm-clius:
@jakubgs now that the rln library is included as a nim-waku submodule, I can fix the Docker image and specifically the error: `Error loading shared library vendor/rln/target/debug/librln.so: No such file or directory`

Just copying the missing file to the image works for me locally. This may violate a dozen best practices, though, so lmk if there's a better way.

The aim is to have the Jenkins jobs `deploy-v2-test` and `deploy-v2-prod` point to the `nim-waku` `master` branch again.

Thanks!

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -42,6 +42,9 @@ RUN ln -s /usr/lib/libpcre.so /usr/lib/libpcre.so.3
 # Copy to separate location to accomodate different MAKE_TARGET values
 COPY --from=nim-build /app/build/$MAKE_TARGET /usr/local/bin/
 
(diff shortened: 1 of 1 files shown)


jakubgs (approved):
But I see this file is not in the repo:
https://github.com/kilic/rln/tree/a80f5d013eb092ff18bd1d946c57565e2cdc65da

So I assume it is created at build time. And I see nothing wrong with that.

jm-clius:
> So I assume it is created at build time. And I see nothing wrong with that.

It is. Thanks, @jakubgs

PR #407, opened 2021-03-05
Minor fixes for chat2 and store

jm-clius:
Adding bugfixes as I go along.

This PR:
1. Shows chat prompt in `chat2` application after receiving messages on subscribed topics.
2. Ensures historical messages from persistent storage are returned in time order.

### DIFF
--- a/examples/v2/chat2.nim
+++ b/examples/v2/chat2.nim
@@ -48,6 +48,7 @@ type Chat = ref object
     connected: bool         # if the node is connected to another peer
     started: bool           # if the node has started
     nick: string            # nickname for this chat session
+    prompt: bool            # chat prompt is showing
 
 type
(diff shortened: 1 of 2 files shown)


staheri14 (approved):
LGTM!

examples/v2/chat2.nim
```diff
@@ -288,6 +294,8 @@ proc processInput(rfd: AsyncFD, rng: ref BrHmacDrbgContext) {.async.} =
         if decodedPayload.isOK():
           let payload = string.fromBytes(decodedPayload.get().payload)
           echo &"{payload}"
+          chat.prompt = false
```
staheri14, line 297: The switching between true and false outside of the proc may get confusing in the future, not sure what is the exact logic, but is it possible to do it inside the `showChatPrompt`?

PR #409, opened 2021-03-08
Improved chat2 message encoding

jm-clius:
This is a further increment towards #399

It improves the encoding of `chat2` messages by serialising the chat text and metadata together using protocol buffers. For now the included fields for every `chat2` message are: `timestamp`, `nick` and `payload`.

One immediate advantage of this approach is that `chat2` peers can interpret these metadata fields, e.g. displaying message timestamps in local time.

### Outstanding issue:
In order to seamlessly use the `chat2` application for dogfooding, there should be [peering between cluster nodes](https://github.com/status-im/nim-waku/issues/408).


staheri14 (approved):
LGTM!

PR #414, opened 2021-03-11
 Filtering messages in the relay protocol using libp2p topic validators

staheri14:
Addresses https://github.com/vacp2p/specs/issues/255
This is just a test to see how we can filter messages in the relay protocol using libp2p topic validators. In subsequent PRs re rln-relay, the validators will implement the zkSNARKs proof verification logic.

### DIFF
--- a/tests/v2/test_wakunode.nim
+++ b/tests/v2/test_wakunode.nim
@@ -383,9 +383,9 @@ procSuite "WakuNode":
 
     check:
       (await completionFut.withTimeout(10.seconds)) == true
-      # check validators is called for message1
+      # check that validator is called for message1
       (await completionFutValidatorAcc.withTimeout(13.seconds)) == true
-      # check validators is called for message2
+      # check that validator is called for message2
       (await completionFutValidatorRej.withTimeout(13.seconds)) == true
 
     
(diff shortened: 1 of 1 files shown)


tests/v2/test_wakunode.nim
```diff
@@ -293,3 +295,103 @@ procSuite "WakuNode":
...
+        check:
+          topic == pubSubTopic
+          # check that only messages with contentTopic1 is relayed (but not contentTopic2)
+          val.contentTopic == contentTopic1
+      # relay handler is called
+      completionFut.complete(true)
+  
+
+    node3.subscribe(pubSubTopic, relayHandler)
+    await sleepAsync(4000.millis)
```
jm-clius, line 377: For interest's sake, have you found it necessary in experimental results to sleep 4s for the subscribe/publish to take effect? My default is usually 2000.millis, but I found that to be too short for certain operations, such as connecting to a new peer and unnecessary long for others. Seems like we may have to benchmark the time it takes to complete common API calls for our test framework.

jm-clius (approved):
LGTM!

tests/v2/test_wakunode.nim
staheri14: I increased the sleep time because I had an issue with the published messages, one out of two published messages wouldn't reach the relayHandler and I  wasn't sure whether  it  is because of the validator or the short sleep time. That was just for debugging. Now that I change it back to 2000.millis it seems fine, I am going to keep it as 2000.millis  to be consistent with the other tests :P 

Benchmarking the time is a great idea! that would be helpful, and specially for debugging the tests.

PR #423, opened 2021-03-16
Add hook on newly added envelope

### LINKED ISSUE #267: Add hook on newly added envelope
We need to be able to trigger a function call when a new envelope is being added to the envelope queue in Waku v1

This is useful e.g. for:
- when storing the envelopes for later retrieval -> Waku v1 mailservers
- when creating a bridge between waku v1 and waku v2.

jm-clius:
This PR closes #267

@kdeme this seemed like the best place to add a "new incoming envelope" hook in `waku_protocol`, while avoiding unnecessary duplicates. Lmk if this is not quite what you had in mind.


kdeme (approved):
lgtm

PR #424, opened 2021-03-16
Update submodules

### LINKED ISSUE #419: libp2p/identify message does not contain protocols field
There is a known [issue](https://github.com/vacp2p/waku-web-chat/blob/master/patches/waku-hack-identify-peer-proto-manual.patch) with the usage of the identify protocol.

My findings are slightly different that what @oskarth's patch insinuates.

The identify message is [defined in the specs](https://github.com/libp2p/specs/tree/master/identify#the-identify-message) as follow:

```
message Identify {
  optional string protocolVersion = 5;
  optional string agentVersion = 6;
  optional bytes publicKey = 1;
  repeated bytes listenAddrs = 2;
  optional bytes observedAddr = 4;
  repeated string protocols = 3;
}
```

It is defined in [js-libp2p](https://github.com/libp2p/js-libp2p/blob/master/src/identify/message.js) as follow:

```
message Identify {
  optional string protocolVersion = 5; // e.g. ipfs/1.0.0
  optional string agentVersion = 6; // e.g. go-ipfs/0.1.0
  optional bytes publicKey = 1;
  repeated bytes listenAddrs = 2;
  optional bytes observedAddr = 4;
  repeated string protocols = 3;
  optional bytes signedPeerRecord = 8;
}
```

When extracting the message sent by nim-waku, my findings are:

```
publicKey <Buffer 08 02 12 21 03 63 c0 e2 6a ca 55 44 e8 85 6d 29 67 88 5b 9c e6 5b f3 64 12 1f 7d 78 ec 6c 2c 42 6d 46 d2 94 ac>
listenAddrs [ <Buffer 04 7f 00 00 01 06 f2 7c> ]
protocols []
observedAddr <Buffer 04 7f 00 00 01 06 ab 79>
signedPeerRecord null
```

The `signedPeerRecord` is not present. IMHO, it is ok because this field does not exist in the libp2p specs and is marked as `optional` in js-libp2p.

IMO, the actual issue is that the `protocols` field is empty. Which means that js-waku does not seem to nim-waku as a peer under the waku protocol. This leads to message exchange issues (ie, this [test](https://github.com/status-im/js-waku/blob/548c50668ae211bafc7c4a9a33f734d6791a7e94/src/lib/node.spec.ts#L99) fails without a patch).

I would expect the protocol field to at least include `/vac/waku/relay/2.0.0-beta2`.

jm-clius:
- Switched `nim-libp2p` submodule (back) from branch `unstable` to `master`
- Expanded `.gitignore`
- Updated all submodules. Should fix and close #419

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,5 @@
 
 rln
 *.log
+package-lock.json
+node_modules/
--- a/.gitmodules
+++ b/.gitmodules
@@ -12,7 +12,7 @@
 	path = vendor/nim-libp2p
 	url = https://github.com/status-im/nim-libp2p.git
 	ignore = dirty
-	branch = unstable
+	branch = master
 [submodule "vendor/nim-stew"]
 	path = vendor/nim-stew
 	url = https://github.com/status-im/nim-stew.git
--- a/examples/v1/example.nim
+++ b/examples/v1/example.nim
@@ -33,7 +33,7 @@ let
 # Create Ethereum Node
 var node = newEthereumNode(config.nodekey, # Node identifier
   address, # Address reachable for incoming requests
-  1, # Network Id, only applicable for ETH protocol
+  NetworkId(1), # Network Id, only applicable for ETH protocol
   nil, # Database, not required for Waku
   clientId, # Client id string
   addAllCapabilities = false, # Disable default all RLPx capabilities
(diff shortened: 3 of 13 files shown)


staheri14 (approved):
LGTM!

Issue #205, opened 2020-10-05
Waku bridge implementation


kdeme:
Implement node which talks both Waku v1 and Waku v2 and transfers Waku messages from 1 network to the other.

Depends on:
- [x] https://github.com/status-im/nim-waku/issues/188
- [x] Having a WakuRelay API call subscribe on all contentTopics (and on one gossipsub topic)
- [x] Having a WakuRelay API call publish 
- [x] Might require changes on Waku v1 API to be able to access the envelope queue (or rather trigger on incoming new envelopes), as currently subscribes there are triggered after decryption: https://github.com/status-im/nim-waku/issues/267

The issue is only about moving messages across networks. Considering https://github.com/status-im/nim-waku/issues/192 is still open, the messages themselves will not be useful (not get decrypted).

This is also somewhat blocked by https://github.com/vacp2p/specs/issues/180 , for now, we can just skip this and use the first 4 bytes.

PR #425, opened 2021-03-17
Add message passing to waku bridge script

jm-clius:
This PR is a first increment towards #205

It scripts the running of both a Waku v1 and Waku v2 node in a single process and passes received data/payload from v1 to v2 and vice versa. A common RPC server exists between the v1 and v2 nodes, allowing for interaction using both v1 and v2 RPC commands.

This is not yet a standalone bridge node, neither does it check for duplicates or do any envelope verification. These will be addressed in subsequent PRs.

### DIFF
--- a/waku/common/wakubridge.nim
+++ b/waku/common/wakubridge.nim
@@ -1,6 +1,7 @@
 import
   std/[strutils, tables],
   chronos, confutils, chronicles, chronicles/topics_registry, metrics,
+  stew/endians2,
   stew/shims/net as stewNet, json_rpc/rpcserver,
   # Waku v1 imports
   eth/[keys, p2p], eth/common/utils,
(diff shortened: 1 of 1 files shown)


staheri14 (approved):

waku/common/wakubridge.nim
```diff
@@ -22,7 +23,24 @@ import
...
+  defaultTTL = 5'u32
+
+proc toWakuMessage(env: Envelope): WakuMessage =
+  # Translate a Waku v1 envelope to a Waku v2 message
+  WakuMessage(payload: env.data,
+              contentTopic: ContentTopic(uint32.fromBytes(env.topic)),
+              version: 1)
+
+proc toWakuV2(env: Envelope, nodev2: WakuNode) {.async.} =
+  await nodev2.publish(defaultBridgeTopic, env.toWakuMessage())
```
staheri14, line 38: Wondering why not using `defaultTopic` https://github.com/status-im/nim-waku/blob/0050f9da59a129bfab34c5eaaa03f864afd01034/waku/v2/node/wakunode2.nim#L35 of wakunode2 in here?

jm-clius: For now I'd like to keep the default topic for bridging messages separate from the one we use for everything else. This may certainly change once I implement a standalone WakuBridge.

PR #426, opened 2021-03-18
Add WakuBridge and test

jm-clius:
This PR is a further increment towards #205

### Description

Implementation is based on the [Waku bridge specification](https://specs.vac.dev/specs/waku/v2/waku-bridge).

It modifies the Waku bridge script that previously ran a Waku v1 and Waku v2 node in a single process with message bridging, and:
- Introduces the `WakuBridge` node containing both a Waku v1 and Waku v2 node under the hood:
	- The Waku v1 node has `Waku` protocol enabled by default
	- The Waku v2 node mounts `relay` by default and listens/publishes on a default PubSub topic specifically for bridging (currently: `"/waku/2/default-bridge/proto"`)
- The `WakuBridge` transfers envelopes from Waku v1 to Waku v2 and `WakuMessages` from Waku v2 to Waku v1:
	- envelope `topic` corresponds to `WakuMessage` `contentTopic`
	- envelope `data` corresponds to `WakuMessage` `payload`
	- `WakuMessages` have `version` set to `1`

### Next steps:
- More sophisticated tests
- Add bridge metrics
- Filter duplicates to ensure no back and forth "bouncing" of messages between v1 and v2

### DIFF
--- a/waku/common/wakubridge.nim
+++ b/waku/common/wakubridge.nim
@@ -23,63 +23,91 @@ import
   # Common cli config
   ./config_bridge
 
+logScope:
+  topics = "wakubridge"
+
+##################
+# Default values #
+##################
+
 const
+  defaultBridgeTopic* = "/waku/2/default-bridge/proto"
   clientIdV1 = "nim-waku v1 node"
-  defaultBridgeTopic = "/waku/2/default-bridge/proto"
   defaultTTL = 5'u32
 
-proc toWakuMessage(env: Envelope): WakuMessage =
+#########
+# Types #
+#########
+
+type
+  WakuBridge* = ref object of RootObj
+    nodev1*: EthereumNode
+    nodev2*: WakuNode
+
+###################
+# Helper funtions #
+###################
+
+func toWakuMessage(env: Envelope): WakuMessage =
   # Translate a Waku v1 envelope to a Waku v2 message
   WakuMessage(payload: env.data,
               contentTopic: ContentTopic(uint32.fromBytes(env.topic)),
               version: 1)
 
-proc toWakuV2(env: Envelope, nodev2: WakuNode) {.async.} =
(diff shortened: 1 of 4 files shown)


waku/common/wakubridge.nim
```diff
@@ -23,63 +23,91 @@ import
...
 
+logScope:
+  topics = "wakubridge"
+
+##################
+# Default values #
+##################
+
 const
+  defaultBridgeTopic* = "/waku/2/default-bridge/proto"
```
oskarth, line 34: Could you elaborate on rationale of this? I'm still a bit jetlagged so it might be obvious, I don't see a mention of a dedicated bridge topic in the spec though https://specs.vac.dev/specs/waku/v2/waku-bridge

oskarth (approved):

waku/common/wakubridge.nim
jm-clius: Well, since there is no Waku v1 equivalent to a PubSub topic we need something to default to when translating from v1 to v2. The question then is, when we do the inverse, if the bridge should translate _from_ all subscribed PubSub topics or whether we should also provide some default topic that is used for communication on both v1 and v2 networks. 
In both cases the usual `defaultTopic` would suffice, but I've decided to keep the bridged messages in a separate `defaultBridgeTopic` to avoid noise - at least while the `bridge` is still being developed and proper duplicate filtering is still lacking.

The spec does not mention PubSub topics, but the wording in #205 made me think that this may have been the original intention. Perhaps @kdeme can comment?

oskarth: > decided to keep the bridged messages in a separate defaultBridgeTopic to avoid noise - at least while the bridge is still being developed and proper duplicate filtering is still lacking.

makes sense

PR #427, opened 2021-03-18
Rln-relay zkp module Nim bindings

staheri14:
This is an incremental PR towards https://github.com/status-im/nim-waku/issues/298 and closes https://github.com/status-im/nim-waku/issues/433
Summary of changes:
- The Nim wrappers of the zkSNARKs component of rln library are added i.e., `generate_proof` and `verify`
- The Nim wrapper of `hash` is also added
- Unit tests of `hash`, `generate_proof` and `verify` are added
- The interface of  `createRLNInstance`  is updated to address https://github.com/status-im/nim-waku/issues/433.  Other parts of the code are also refactored based on this new interface.


oskarth:
My Nim-foo isn't strong enough to know if this is the best way to simplify the type argument, roping in @zah to see if he has any thoughts

oskarth (approved):
Seems OK to me but all the ptr and addr types makes me somewhat uneasy, not sure if this is idiomatic Nim. See if Zahary has some more constructive thoughs

waku/v2/protocol/waku_rln_relay/rln.nim
```diff
@@ -23,19 +23,40 @@ type Buffer* = object
   `ptr`*: ptr uint8
```
oskarth, line 23: What happens if you add .push raises: [Defect].} a la https://github.com/status-im/nim-waku/issues/289 here? Curious if you risk getting a bunch of NullPointerExceptions here

jm-clius (approved):

waku/v2/protocol/waku_rln_relay/rln.nim
staheri14: I added that to the rln module, it compiled just fine. Do you also want me to add it to all the other procs that involve rln lib procedures?

staheri14:
> Seems OK to me but all the ptr and addr types makes me somewhat uneasy, not sure if this is idiomatic Nim. See if Zahary has some more constructive thoughs

I am not expert in this as well, but as far as I know passing addresses while interfacing with other libs in other languages is typical, e.g., look at listing 8.3 of this https://livebook.manning.com/book/nim-in-action/chapter-8/96
Would be glad to know if there is any alternative ways. :)

waku/v2/protocol/waku_rln_relay/rln.nim
oskarth: oh nice, I think the general idea is to have it in each module/namespace, but this should be fine too

oskarth:
Looking at the PR I don't see what would make Windows build fail here, could be some random CI artifact?

Issue #432, opened 2021-03-23
Peer manager: add persistence


jm-clius:
### Problem

`wakunode2` uses a [rudimentary peer manager](https://github.com/status-im/nim-waku/issues/358) to keep track of peers registered on the node. Currently, however, managed peers are not persisted. This implies that a node loses all its `relay` connections and registered `filter`, `swap` and `store` peers after a restart. One solution would be to extend the [existing SQLite persistence](https://github.com/status-im/nim-waku/pull/247) to store and load peer information.

### Acceptance criteria

Managed peers are persisted across restarts.

cc @jakubgs

PR #435, opened 2021-03-24
Added persistent storage for peers

jm-clius:
This PR introduces persistent peer storage for Waku v2. It is an increment towards #432. 

### Background

Persistence was original deprioritised as a feature for the PoC Waku peer manager in #358, while waiting for a more complete implementation in https://github.com/status-im/nim-libp2p/issues/537.

However, in order to ensure stability in our testnets, and specifically the single network cluster deployments (`test` and `prod`), persisting connected peers across restarts has become more urgent. This is especially necessary for [Waku v2 dogfooding](#399) using the `chat2` app and `test` cluster.

### Next step:

- Integrate persistent storage with peer manager and test

Relates to #358 #399

### DIFF
--- a/waku/v2/node/storage/peer/waku_peer_storage.nim
+++ b/waku/v2/node/storage/peer/waku_peer_storage.nim
@@ -1,0 +1,136 @@
+import
+  std/sets, 
+  sqlite3_abi,
+  chronos, metrics,
+  libp2p/protobuf/minprotobuf,
+  stew/results,
+  ../sqlite,
+  ../../peer_manager
+
+export sqlite
+
+type
+  WakuPeerStorage* = ref object of RootObj
+    database*: SqliteDatabase
+  
+  WakuPeerStorageResult*[T] = Result[T, string]
+
+  DataProc* = proc(peerId: PeerID, storedInfo: StoredInfo,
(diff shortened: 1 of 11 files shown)


staheri14 (approved):
Nice!

waku/v2/node/storage/peer/waku_peer_storage.nim
```diff
@@ -0,0 +1,136 @@
...
+  chronos, metrics,
+  libp2p/protobuf/minprotobuf,
+  stew/results,
+  ../sqlite,
+  ../../peer_manager
+
+export sqlite
+
+type
+  WakuPeerStorage* = ref object of RootObj
```
oskarth, line 13: Is this a separate DB from the one used for store? How will it interact with the whole sqlite/encrypted mismatch we saw with Stimbus previously?

oskarth (approved):
LGTM! Also changelog :)

waku/v2/node/storage/peer/waku_peer_storage.nim
jm-clius: The idea is to use the same DB and just add a separate table. I foresee the same clash as previously, so will similarly extract an interface for Stimbus to provide their own (optional) PeerStorage.

PR #436, opened 2021-03-24
Bump chronos and fix exception tracking issues

kdeme:
Exception tracking support was added recently to chronos : https://github.com/status-im/nim-chronos/pull/166

This PR bumps to that chronos version and attempts to fix any exception effect issue in order to be compatible with that chronos version (bumping nim-eth and nim-libp2p along the way).

~~Work in progress: Need to fix a proc type still in nim-eth and relies on https://github.com/status-im/nim-chronos/pull/168 of which I'm not sure it is the right fix.~~

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -110,3 +110,13 @@
 	path = vendor/rln
 	url = https://github.com/kilic/rln
 	branch = full-node
+[submodule "vendor/nim-testutils"]
+	path = vendor/nim-testutils
+	url = https://github.com/status-im/nim-testutils.git
+	ignore = untracked
+	branch = master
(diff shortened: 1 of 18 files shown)


kdeme:
Also bumped all other submodules in the process, except for nim-metrics as that requires some changes to adjust the http server used, which is out of scope here and for another PR.

oskarth (approved):

arnetheduck:
@kdeme signed commits?

jm-clius (approved):

kdeme:
> @kdeme signed commits?

woops, resolved.

PR #437, opened 2021-03-25
Integrate persistent peer storage

jm-clius:
This PR closes #435 

The immediate goal is to improve [cluster](https://fleets.status.im/) stability in aid of [Waku v2 dogfooding](#399)

It:
- Integrates persistent peer storage in Waku v2.
- Extracts a PeerStorage interface for flexible implementation
- Allows node to attempt reconnection to `WakuRelay` peers after a restart.
- Adds various tests


tests/v2/test_peer_manager.nim
```diff
@@ -162,3 +163,53 @@ procSuite "Peer Manager":
...
+    node2.mountRelay()
+
+    discard await node1.peerManager.dialPeer(peerInfo2, WakuRelayCodec, 2.seconds)
+    check:
+      # Currently connected to node2
+      node1.peerManager.peers().len == 1
+      node1.peerManager.peers().anyIt(it.peerId == peerInfo2.peerId)
+      node1.peerManager.connectedness(peerInfo2.peerId) == Connected
+
+    # Simulate restart by initialising a new node using the same storage
```
staheri14, line 194: Here a restart means stopping the node and then starting it? curious to know was there any issue with restarting node1 (as opposed to spinning a new node with the same storage)?

staheri14 (approved):
Good!

waku/v2/node/peer_manager/peer_manager.nim
```diff
@@ -3,75 +3,122 @@
...
+    debug "Dialing remote peer failed", msg = e.msg
+    waku_peers_dials.inc(labelValues = ["failed"])
+    
+    pm.peerStore.connectionBook.set(peerId, CannotConnect)
+    if not pm.storage.isNil:
+      pm.storage.insertOrReplace(peerId, pm.peerStore.get(peerId), CannotConnect)
+    
+    return none(Connection)
+
+proc primeFromStorage(pm: PeerManager) =
```
oskarth, line 76: What does this name mean? I find the "prime" verb a bit confusing

oskarth (approved):

tests/v2/test_peer_manager.nim
jm-clius: No issue at all, but, since `node1` is already initialised in this scope, just stopping it won't destroy its peer manager and the persistent storage won't really be necessary. The easiest way to simulate a node being initialised "from scratch" is to just start a new one that loads from persistent storage.

waku/v2/node/peer_manager/peer_manager.nim
jm-clius: I meant it in the sense of "prepare for usage". I think `loadFromStorage` may be less ambiguous.

PR #444, opened 2021-03-30
Update kibana link to only filter on fleet

D4nte:
Also, return last 7 days of logs instead of
fixed date in the past.


oskarth (approved):
Thanks!

PR #446, opened 2021-03-31
Start relay protocol on wakunode2

### LINKED ISSUE #445: `wakurelay`, and underlying `gossipsub`, is never started
### Description
`WakuRelay` [defines](https://github.com/status-im/nim-waku/blob/master/waku/v2/protocol/waku_relay.nim#L76-L78) `start()` on the underlying `GossipSub` protocol, but this is never called for a `wakunode2`.

### Impact
The impact is that (a) the GossipSub `heartbeat` never runs, and (b) direct peers are not maintained. The latter is not yet used for `wakunode2`, but `heartbeat` is required to maintain the mesh, send `IHAVE` and `IWANT` messages, etc.

jm-clius:
This PR fixes #445

It ensures that `relay` protocol, and the underlying `gossipsub`, is started on a `wakunode2`.

This fix assumes the following:
1. A `node` can be started and `relay` protocol mounted in any order.
2. When `relay` is mounted, it should start if and only if the `node` was already started. Otherwise it should start on `node` start.

**Note**: I'm not sure if the original intention was to be somewhat more strict in terms of the ordering above, e.g. that protocols should always be mounted _before_ the node starts listening. In our current examples, tests and scripts, the `mount()`<->`start()` order seems to be interchangeable, which allows for mounting protocols on live nodes.

### DIFF
--- a/waku/v2/node/wakunode2.nim
+++ b/waku/v2/node/wakunode2.nim
@@ -162,6 +162,9 @@ proc start*(node: WakuNode) {.async.} =
   ## XXX: this should be /ip4..., / stripped?
   info "Listening on", full = listenStr
 
+  if not node.wakuRelay.isNil:
+    await node.wakuRelay.start()
+
 proc stop*(node: WakuNode) {.async.} =
   if not node.wakuRelay.isNil:
     await node.wakuRelay.stop()
@@ -416,6 +419,13 @@ proc mountRelay*(node: WakuNode, topics: seq[string] = newSeq[string](), rlnRela
     # TODO currently the message validator is set for the defaultTopic, this can be configurable to accept other pubsub topics as well 
     addRLNRelayValidator(node, defaultTopic)
     info "WakuRLNRelay is mounted successfully"
+  
+  if node.libp2pTransportLoops.len > 0:
+    # Node has already started. Start the WakuRelay protocol
+
+    waitFor node.wakuRelay.start()
+
+    info "relay mounted and started successfully"
 
 
 ## Helpers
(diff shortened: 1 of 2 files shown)


waku/v2/node/wakunode2.nim
```diff
@@ -416,6 +419,13 @@ proc mountRelay*(node: WakuNode, topics: seq[string] = newSeq[string](), rlnRela
     # TODO currently the message validator is set for the defaultTopic, this can be configurable to accept other pubsub topics as well 
     addRLNRelayValidator(node, defaultTopic)
     info "WakuRLNRelay is mounted successfully"
+  
+  if node.libp2pTransportLoops.len > 0:
```
oskarth, line 423: Hm, is this the idiomatic way of doing this? How does it work in nimbus-eth2?

oskarth (approved):

waku/v2/node/wakunode2.nim
jm-clius: nimbus-eth2 follows a strict order where the protocol is always mounted before starting the node. In fact, they mount GossipSub when [creating the eth2 node](https://github.com/status-im/nimbus-eth2/blob/stable/beacon_chain/networking/eth2_network.nim#L1555), implying that my assumptions in the description above don't hold for them. We may consider doing the same (perhaps only for `WakuRelay`?), but that would require a fairly significant change in approach.

Seeing if related futures are set to know if the protocol/node has started doesn't feel idiomatic, but is at least not without precedent - this is similar to how [GossipSub itself avoids a "double" start](https://github.com/status-im/nim-libp2p/blob/master/libp2p/protocols/pubsub/gossipsub.nim#L583-L585). A cleaner alternative, of course, is to just add `started` state.

jm-clius: In fact, I've changed this now to rather use an explicit `started` state variable (31700c8). Lmk if you have strong feelings about either the order in which we mount/start or what type of state we keep.

PR #450, opened 2021-04-01
Add timestamp field to the waku message

staheri14:
This PR is the first increment towards  https://github.com/vacp2p/research/issues/64 and closes https://github.com/status-im/nim-waku/issues/449


oskarth (commented):
@staheri14 This looks reasonable! Could we please add a corresponding spec PR? To make sure these stay in sync and we understand proposed semantics.

PR #451, opened 2021-04-01
Adds startTime and endTime to the HistoryQuery

staheri14:
This is an incremental PR towards vacp2p/research#64. 
Two new fields of `startTime` and `endTime` are added to the `HistoryQuery` to enable time-window queries. 

**Some notes:**
The initial thought was to add time-window queries and the relevant fields to the pagination logic i.e., add them to https://github.com/status-im/nim-waku/blob/683577f04552f4fd22f2b302584f23abd3f0e3d4/waku/v2/protocol/waku_store/waku_store_types.nim#L33 , but inspired by  the Twitter pagination technique https://developer.twitter.com/en/docs/twitter-api/pagination, I found it more meaningful to treat the query process separate from the pagination part.  The idea is that we query a history based on the combination of time-window and topics, then we segment the result into pages in case that the number of retrieved messages is more than the `MaxPageSize` https://github.com/status-im/nim-waku/blob/683577f04552f4fd22f2b302584f23abd3f0e3d4/waku/v2/protocol/waku_store/waku_store_types.nim#L17 

In the subsequent PRs, I will rework the history query process by handling time-based queries.


oskarth (approved):
Ditto re spec PR

PR #454, opened 2021-04-05
Enabling time window history query

staheri14:
An incremental PR towards https://github.com/vacp2p/research/issues/64 and https://github.com/status-im/nim-waku/issues/458.

As part of enabling a fault-tolerant store protocol, peers must be able to retrieve history messages based on a time range (during which they were absent). In this PR, the logic for handling temporal history queries is added to the store protocol. The unit tests are also included.

### DIFF
--- a/tests/v2/test_waku_store.nim
+++ b/tests/v2/test_waku_store.nim
@@ -330,11 +330,11 @@ procSuite "Waku Store":
       decodedEmptyPagingInfo.isErr == false
       decodedEmptyPagingInfo.value == emptyPagingInfo
   
-  test "HistoryQuery Protobuf encod/init test":
+  test "HistoryQuery Protobuf encode/init test":
     let
       index = computeIndex(WakuMessage(payload: @[byte 1], contentTopic: ContentTopic(1)))
       pagingInfo = PagingInfo(pageSize: 1, cursor: index, direction: PagingDirection.BACKWARD)
-      query=HistoryQuery(topics: @[ContentTopic(1)], pagingInfo: pagingInfo)
+      query=HistoryQuery(topics: @[ContentTopic(1)], pagingInfo: pagingInfo, startTime: float64(10), endTime: float64(11))
       pb = query.encode()
       decodedQuery = HistoryQuery.init(pb.buffer)
 
--- a/tests/v2/test_waku_store.nim
+++ b/tests/v2/test_waku_store.nim
@@ -411,19 +411,49 @@ procSuite "Waku Store":
 
     for wakuMsg in msgList:
       await subscriptions.notify("foo", wakuMsg)
-    var completionFut = newFuture[bool]()
+    
+    asyncTest "handle temporal history query with a valid time-window":
+      var completionFut = newFuture[bool]()
 
-    proc handler(response: HistoryResponse) {.gcsafe, closure.} =
-      echo response.messages[0]
-      echo response.messages[1]
+      proc handler(response: HistoryResponse) {.gcsafe, closure.} =
+        check:
+          response.messages.len() == 2
+          response.messages.anyIt(it.timestamp == float(3))
+          response.messages.anyIt(it.timestamp == float(5))
+        completionFut.complete(true)
+
+      let rpc = HistoryQuery(topics: @[ContentTopic(1)], startTime: float(2), endTime: float(5))
+      await proto.query(rpc, handler)
+
       check:
-        response.messages.len() == 2
-        response.messages.anyIt(it.timestamp == float(3))
-        response.messages.anyIt(it.timestamp == float(5))
-      completionFut.complete(true)
+        (await completionFut.withTimeout(5.seconds)) == true
 
-    let rpc = HistoryQuery(topics: @[ContentTopic(1)], startTime: float(2), endTime: float(5))
-    await proto.query(rpc, handler)
+    asyncTest "handle temporal history queries with zero-size window":
+      var completionFut = newFuture[bool]()
 
-    check:
-      (await completionFut.withTimeout(5.seconds)) == true
+      proc handler(response: HistoryResponse) {.gcsafe, closure.} =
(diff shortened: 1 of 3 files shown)


tests/v2/test_waku_store.nim
```diff
@@ -377,3 +377,84 @@ procSuite "Waku Store":
       decodedEmptyRes.isErr == false
       decodedEmptyRes.value == emptyRes
     
+  asyncTest "temporal history queries":
+    let
+      key = PrivateKey.random(ECDSA, rng[]).get()
+      peer = PeerInfo.init(key)
+    var
+      msgList = @[WakuMessage(payload: @[byte 0], contentTopic: ContentTopic(2)),
+        WakuMessage(payload: @[byte 1],contentTopic: ContentTopic(1), timestamp: float(1)),
```
oskarth, line 386: I thought this was a double?

oskarth:
I see the linked Vac research issue, is there also some corresponding nim-waku implementation issue?

tests/v2/test_waku_store.nim
oskarth, line 386: huh my comment disappeared, let's try again:

I thought this was a double? https://github.com/vacp2p/rfc/pull/324/files

oskarth (commented):
I see there's a Vac research issue, is there also some corresponding nim-waku implementation issue?

tests/v2/test_waku_store.nim
staheri14: Yeah, float (float64 which is the default in Nim) is equivalent to double in protobuf types https://developers.google.com/protocol-buffers/docs/proto3#scalar

staheri14: Yeah, float (float64 which is the default in Nim) is equivalent to double in protobuf types https://developers.google.com/protocol-buffers/docs/proto3#scalar

staheri14:
> I see the linked Vac research issue, is there also some corresponding nim-waku implementation issue?

Just created the issue https://github.com/status-im/nim-waku/issues/458

tests/v2/test_waku_store.nim
oskarth: oh I didn't realize, thanks!

oskarth (approved):

tests/v2/test_waku_store.nim
```diff
@@ -377,3 +377,84 @@ procSuite "Waku Store":
...
+
+    asyncTest "handle temporal history query with a zero-size time window":
+      var completionFut = newFuture[bool]()
+
+      proc handler(response: HistoryResponse) {.gcsafe, closure.} =
+        check:
+          response.messages.len() == 5
+        completionFut.complete(true)
+
+      let rpc = HistoryQuery(topics: @[ContentTopic(1)], startTime: float(2), endTime: float(2))
```
jm-clius, line 439: I may be missing something, but shouldn't the special case here be to only return messages with timestamp = startTime = endTime = float(2)? It seems a bit strange to me to return all historical messages if the window size is 0.

tests/v2/test_waku_store.nim
```diff
@@ -377,3 +377,84 @@ procSuite "Waku Store":
...
+    asyncTest "handle temporal history query with an invalid time window":
+      var completionFut = newFuture[bool]()
+
+      proc handler(response: HistoryResponse) {.gcsafe, closure.} =
+        check:
+          response.messages.len() == 5
+        completionFut.complete(true)
+
+      # time window is invalid since start time > end time
+      let rpc = HistoryQuery(topics: @[ContentTopic(1)], startTime: float(5), endTime: float(2))
```
jm-clius, line 454: My first thought would be that a query with an explicit but invalid time range should return zero messages rather than quietly ignoring the time arguments and returning everything?

waku/v2/protocol/waku_store/waku_store.nim
```diff
@@ -308,6 +308,12 @@ proc findMessages(w: WakuStore, query: HistoryQuery): HistoryResponse =
   result = HistoryResponse(messages: newSeq[WakuMessage]())
   # data holds IndexedWakuMessage whose topics match the query
   var data = w.messages.filterIt(it.msg.contentTopic in query.topics)  
+
+  # temporal filtering
+  if (query.endTime > query.startTime):
```
jm-clius, line 313: Since the endpoints are inclusive in the tests below, I think `(query.endTime >= query.startTime)` may be the more general condition (this relates to my comments on the test).

jm-clius (approved):
LGTM! I just added a couple of questions on the expected responses for invalid queries. :)

tests/v2/test_waku_store.nim
staheri14: I considered a zero-size time window as an invalid query (i.e., no temporal filter should take place). Somehow it did not make sense that someone sends a request whose response is already empty! Indeed, this is equivalent to leaving the time window unspecified wherein the window size is also zero and means that no time filter is requested.
However, in this case, I am assuming that the requester checks its window size prior to query which might not be true.
Conclusion: I am going to change this logic, and adopt your suggested approach, it would be safer. Thanks!

tests/v2/test_waku_store.nim
staheri14: Thanks @jm-clius! Oskar also had a similar comment, please see my answer https://github.com/status-im/nim-waku/pull/454#discussion_r608838867

waku/v2/protocol/waku_store/waku_store.nim
staheri14: Sure, makes sense!

staheri14:
> LGTM! I just added a couple of questions on the expected responses for invalid queries. :)

Thanks @jm-clius! please let me know if my recent commits addressed your questions :)

jm-clius (approved):
Thanks for clarifying!

PR #457, opened 2021-04-06
Updates the field number of proof and timestamp

staheri14:
Addresses https://github.com/vacp2p/rfc/pull/325#pullrequestreview-628565224
This is to change the field number of proof and timestamp fields in the waku message protobuf schema. 
- The field number of proof is changed from 4 to 21, this is because the proof is related to RLN which is yet an experimental feature. Thus, for now `proof` is an optional field. 
- Following this update, the timestamp field number is shifted to 4.

### DIFF
--- a/waku/v2/protocol/waku_message.nim
+++ b/waku/v2/protocol/waku_message.nim
@@ -33,6 +33,7 @@ proc init*(T: type WakuMessage, buffer: seq[byte]): ProtoResult[T] =
   discard ? pb.getField(2, msg.contentTopic)
   discard ? pb.getField(3, msg.version)
   discard ? pb.getField(4, msg.timestamp)
+# XXX Experimental, this is part of https://rfc.vac.dev/spec/17/ spec and not yet part of WakuMessage spec
   discard ? pb.getField(21, msg.proof)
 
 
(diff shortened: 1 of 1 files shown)


waku/v2/protocol/waku_message.nim
```diff
@@ -30,8 +30,9 @@ proc init*(T: type WakuMessage, buffer: seq[byte]): ProtoResult[T] =
   discard ? pb.getField(1, msg.payload)
   discard ? pb.getField(2, msg.contentTopic)
   discard ? pb.getField(3, msg.version)
-  discard ? pb.getField(4, msg.proof)
-  discard ? pb.getField(5, msg.timestamp)
+  discard ? pb.getField(4, msg.timestamp)
+  discard ? pb.getField(21, msg.proof)
```
oskarth, line 34: Probably worth adding a comment here such as

oskarth, line 34:
```suggestion
# XXX Experimental, this is part of https://rfc.vac.dev/spec/17/ spec and not yet part of WakuMessage spec
  discard ? pb.getField(21, msg.proof)
```

oskarth: Though this comment could possibly be part of the Nim data structure instead

oskarth (approved):

jm-clius (approved):

waku/v2/protocol/waku_message.nim
staheri14: Thanks, noted this comment under the waku message type, please see the last commit

Issue #447, opened 2021-03-31
Change `contentTopic` type back to `string`


jm-clius:
### Background

In Waku v2, `contentTopic` was changed from a `string` to an `int32` to enable easy mapping to Waku v1 topics (4-byte arrays). However, this has reduced the usability of the `contentTopic`-field in a number of ways:

1. Integers are less user-friendly/readable than strings, especially for API calls (e.g. publishing on "my/content/topic" using the JSON-RPC API is easier than publishing on some hash of that).
2. Namespacing will be easier implemented for strings (see https://github.com/vacp2p/specs/issues/227).
3. A 4-byte space is not wide enough for applications requiring a large amount of content topics (e.g. WalletConnect want to use 32-byte hex strings as content topics).

### Suggestion

Revert to a `string` representation for `contentTopic` as set out by @oskarth [here](https://github.com/vacp2p/specs/issues/227#issue-731115095). Individual applications of the Waku protocol can then design their contentTopics according to their own requirements. 

### What about bridging to V1?
Bridging to Waku v1 is one application of the Waku protocol, so shouldn't dictate the protocol design but rather how it's implemented. It has a requirement to have 4-byte topics, which can neatly fit into the above paradigm by [introducing namespacing](https://github.com/vacp2p/specs/issues/227).

cc @oskarth for comments

oskarth:
Agree with OP, main question for me is what was set out in https://github.com/vacp2p/rfc/issues/227#issuecomment-811625966  (1) Core's POV (2) How strict/freeform we want namespaces to be

jm-clius:
I commented on vacp2p/rfc#227, but in summary:
(1) I don't think Core has specific use cases in mind (yet), though I still think this will be a useful feature when designing new features. We can probably spec namespacing in a way that already suggests possible use cases.
(2) I think we should provide optional support for namespacing but make it mandatory for certain applications within our control, the most obvious one being the bridging-to-V1 use case.

oskarth:
See https://github.com/vacp2p/rfc/pull/335

Any chance we could do this soon @jm-clius? Might help with interop and WalletConnect use case

oskarth:
Is there something left on this issue @jm-clius?

jm-clius:
Merged and tested. Closing issue.

PR #463, opened 2021-04-07
Change contentTopic to string

jm-clius:
This PR addresses #447 

It changes the `WakuMessage` `contentTopic` back to `string`. Spec changes done in https://github.com/vacp2p/rfc/pull/337

#### Implementation notes:
 - The "default" content topic is now `"waku/2/default-content/proto"`. This is a rather arbitrary choice and may change once we spec [namespacing for content topics](https://github.com/vacp2p/rfc/issues/227)
 - For now, `WakuBridge` maps v2 content topics to v1 topics by selecting the first four bytes. This will change when we introduce [namespacing](https://github.com/vacp2p/rfc/issues/227)
 - Note that changing the `contentTopic` type as argument in JSON-RPC calls affects users of the API (cc @D4nte)
 - `WakuStore` [index computation for messages](https://github.com/status-im/nim-waku/blob/master/waku/v2/protocol/waku_store/waku_store.nim#L44) would have changed somewhat, as it is now using `string.toBytes()` rather than `uint32.toBytes()` in the algorithm. @staheri14, I assume this will be OK going forward, though existing stores (e.g. `test` fleet) may be affected?

#### Next steps:
- Specify namespacing for content topics: https://github.com/vacp2p/rfc/issues/227

### DIFF
--- a/examples/v2/basic2.nim
+++ b/examples/v2/basic2.nim
@@ -35,7 +35,7 @@ proc runBackground() {.async.} =
 
   # Publish to a topic
   let payload = cast[seq[byte]]("hello world")
-  let message = WakuMessage(payload: payload, contentTopic: ContentTopic(1))
+  let message = WakuMessage(payload: payload, contentTopic: ContentTopic("/waku/2/default-content/proto"))
   await node.publish(topic, message)
 
 # TODO Await with try/except here
--- a/examples/v2/chat2.nim
+++ b/examples/v2/chat2.nim
@@ -36,8 +36,7 @@ const
   PayloadV1* {.booldefine.} = false
   DefaultTopic = "/waku/2/default-waku/proto"
 
-  Dingpu = "dingpu".toBytes
-  DefaultContentTopic = ContentTopic(uint32.fromBytes(Dingpu))
+  DefaultContentTopic = ContentTopic("dingpu")
 
 # XXX Connected is a bit annoying, because incoming connections don't trigger state change
 # Could poll connection pool or something here, I suppose
--- a/tests/v2/test_jsonrpc_waku.nim
+++ b/tests/v2/test_jsonrpc_waku.nim
@@ -32,6 +32,7 @@ createRpcSigs(RpcHttpClient, sigPath)
 
 procSuite "Waku v2 JSON-RPC API":
   const defaultTopic = "/waku/2/default-waku/proto"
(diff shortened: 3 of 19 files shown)


staheri14:
> * `WakuStore` [index computation for messages](https://github.com/status-im/nim-waku/blob/master/waku/v2/protocol/waku_store/waku_store.nim#L44) would have changed somewhat, as it is now using `string.toBytes()` rather than `uint32.toBytes()` in the algorithm. @staheri14, I assume this will be OK going forward,

yes, as long as it is converted to bytes, it would be fine

staheri14 (commented):
It looks fine to me but the windows checks are failed

D4nte:
> WakuStore index computation for messages would have changed somewhat, as it is now using string.toBytes() rather than uint32.toBytes() in the algorithm. @staheri14, I assume this will be OK going forward, though existing stores (e.g. test fleet) may be affected?


Can we please clarify the encoding of string so we know what `toBytes` does?

oskarth (approved):

oskarth:
I wouldn't worry about Windows CI in this PR, can attack in isolation

staheri14:
> > WakuStore index computation for messages would have changed somewhat, as it is now using string.toBytes() rather than uint32.toBytes() in the algorithm. @staheri14, I assume this will be OK going forward, though existing stores (e.g. test fleet) may be affected?
> 
> Can we please clarify the encoding of string so we know what `toBytes` does?

@D4nte we use the toBytes proc as defined in https://github.com/status-im/nim-stew/blob/ee78822e057ac5f39804ecb6ac1096734be13ef8/stew/byteutils.nim#L142

D4nte:
> @D4nte we use the toBytes proc as defined in [status-im/nim-stew@`ee78822`/stew/byteutils.nim#L142](https://github.com/status-im/nim-stew/blob/ee78822e057ac5f39804ecb6ac1096734be13ef8/stew/byteutils.nim#L142)

Unfortunately this does not help me understand what encoding is used for nim strings. It's ok, I do not need this information anymore as topics are now string so the conversion string to bytes is not used anymore.

oskarth:
@D4nte Did you ask in Nimbus discord as suggested for more details on encoding?

D4nte:
> @D4nte Did you ask in Nimbus discord as suggested for more details on encoding?

So in short, there is no encoding, meaning the encoding is set by the input (e.g. terminal, browser, etc).

PR #464, opened 2021-04-07
fix: direction encoding

richard-ramos:
The way the direction is being encoded right should be roughly equivalent to this code. The direction is a Protobuffer, instead of being an enum/uint32 value:
```js
message PagingInfo {
  int64 pageSize = 1;
  Index cursor = 2;
  Direction direction = 3;
}

Direction {
  uint32 value = 1
}
```

To match the [spec description](https://rfc.vac.dev/spec/13/#protobuf), in this PR the `PagingInfo` message encoding/decoding is modified, so its 3rd field (`direction`) contains an uint32 value instead of a encoded message

### DIFF
--- a/tests/v2/test_waku_store.nim
+++ b/tests/v2/test_waku_store.nim
@@ -301,7 +301,7 @@ procSuite "Waku Store":
   test "PagingInfo Protobuf encod/init test":
     let
       index = computeIndex(WakuMessage(payload: @[byte 1], contentTopic: defaultContentTopic))
-      pagingInfo = PagingInfo(pageSize: 1, cursor: index, direction: PagingDirection.BACKWARD)
+      pagingInfo = PagingInfo(pageSize: 1, cursor: index, direction: PagingDirection.FORWARD)
       pb = pagingInfo.encode()
       decodedPagingInfo = PagingInfo.init(pb.buffer)
 
@@ -309,6 +309,7 @@ procSuite "Waku Store":
       # the fields of decodedPagingInfo must be the same as the original pagingInfo
       decodedPagingInfo.isErr == false
       decodedPagingInfo.value == pagingInfo
+      decodedPagingInfo.value.direction == pagingInfo.direction
     
     let
       emptyPagingInfo = PagingInfo()
--- a/tests/v2/test_waku_store.nim
+++ b/tests/v2/test_waku_store.nim
@@ -298,18 +298,6 @@ procSuite "Waku Store":
       decodedEmptyIndex.isErr == false
       decodedEmptyIndex.value == emptyIndex
 
-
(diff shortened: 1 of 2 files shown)


oskarth:
Thanks for the bug fix @richard-ramos! I'll defer to @staheri14 if she agrees with this fix as she worked on the pagination spec/impl

tests/v2/test_waku_store.nim
```diff
@@ -296,29 +296,18 @@ procSuite "Waku Store":
...
-    check:
-      # the decodedPagingDirection must be the same as the original pagingDirection
-      decodedPagingDirection.isErr == false
-      decodedPagingDirection.value == pagingDirection
-
   test "PagingInfo Protobuf encod/init test":
     let
       index = computeIndex(WakuMessage(payload: @[byte 1], contentTopic: ContentTopic(1)))
-      pagingInfo = PagingInfo(pageSize: 1, cursor: index, direction: PagingDirection.BACKWARD)
+      pagingInfo = PagingInfo(pageSize: 1, cursor: index, direction: PagingDirection.FORWARD)
```
staheri14, line 302: Wondering why this is changed to `FORWARD`, did you find any issue with the `BACKWARD` option?

staheri14 (approved):
Thanks for the PR! looks good to me. Just left a comment re one of the test scenarios.

staheri14:
@richard-ramos  Not sure, but do we have to update the specs? I think we should be able to use the direction enum that was initially defined in the protobuf.

tests/v2/test_waku_store.nim
richard-ramos: No issues at all. I changed because `PagingDirection.Forward` has a value of `1`, and I wanted to be sure that decoding the direction would return the same and not a default value of `0`. (I don't even know if it's a possible scenario)

richard-ramos:
> @richard-ramos Not sure, but do we have to update the specs? I think we should be able to use the direction enum that was initially defined in the protobuf.

It's not necessary. With this PR the code match will match the current the spec.

tests/v2/test_waku_store.nim
staheri14: Oh I see, makes sense!

staheri14:
> > @richard-ramos Not sure, but do we have to update the specs? I think we should be able to use the direction enum that was initially defined in the protobuf.
> 
> It's not necessary. With this PR the code match will match the current the spec.

Sounds good!

staheri14:
@richard-ramos No more comments from my side, PR looks good for merge! (just need to update it with the master branch)

oskarth:
@richard-ramos sorry about that, something up with Windows CI, let me override

PR #467, opened 2021-04-08
Fix Windows CI

### LINKED ISSUE #477: Windows CI failing after `contentTopic` type changed to string
Windows CI has been failing since https://github.com/status-im/nim-waku/pull/463

Failures are in two places:
1. `WakuStore` tests - this was due to a difference in speed to compute `epochTime()` between platforms which uncovered a bug in the sorting of historical messages. This has now been fixed in https://github.com/status-im/nim-waku/pull/467
2. `Filter API` tests - this is most likely also caused by an unexpected change in execution speed for certain functions, which has knock-on effects on concurrency. A well-placed `sleepAsync` may fix this test, but first the issue and place of failure must be understood.

jm-clius:
This PR fixes #477

Windows CI has been failing since `contentTopic` changed to `string`. There were two unrelated failures.

### 1. `Store` retrieval failure 
This failure was caused by two factors:
- **A difference in message sorting order in the `store` protocol between platforms**: This only affected the `store` unit tests. Sorting order is supposed to prioritise timestamps to achieve chronological ordering, followed by `digest`. Due to a difference in the result/speed of the `epochTime()` function between platforms/OSs, in the `windows-latest` virtual environment the timestamps were exactly equal for all messages. An artificial delay between message notifications in the `store` test solved this problem.
- **A bug in `store` protocol in determining the start/end index of stored messages**:  The start/end indexes were determined from the unsorted list (which maintained insertion order) rather than the sorted list (which maintains chronological order). On non-Windows systems these lists were identically ordered due to `1` above.

#### Why has this not affected us before?

Since `contentTopic` forms the most significant part of the `digest` for stored messages, this was affected by the type change to `string`. Previously we were lucky that the _sort-by-digest_ list (on Windows, due to timestamp resolution) equalled the _sort-by-timestamp_ list (on non-Windows) equalled the _unsorted-insertion-order_ list (for `start` or `end` index).

### 2. `Filter API` failure

The API cache (that allows for message polling) had a default size sufficiently large that the [max RPC response body length](https://github.com/status-im/nim-json-rpc/blob/master/json_rpc/clients/httpclient.nim#L24) could be exceeded in some responses. This likely only affected Windows due to differences in string encodings, though I haven't been able to pinpoint this yet. 

A more complete solution will allow for configurable cache sizes. Tracking issue created [here](https://github.com/status-im/nim-waku/issues/478)

### DIFF
--- a/tests/v2/test_waku_store.nim
+++ b/tests/v2/test_waku_store.nim
@@ -88,6 +88,7 @@ procSuite "Waku Store":
     listenSwitch.mount(proto)
 
     await subscriptions.notify("foo", msg)
+    await sleepAsync(1.millis)  # Sleep a millisecond to ensure messages are stored chronologically
     await subscriptions.notify("foo", msg2)
 
     var completionFut = newFuture[bool]()
@@ -162,6 +163,7 @@ procSuite "Waku Store":
 
     for wakuMsg in msgList:
       await subscriptions.notify("foo", wakuMsg)
+      await sleepAsync(1.millis)  # Sleep a millisecond to ensure messages are stored chronologically
 
     var completionFut = newFuture[bool]()
 
@@ -212,6 +214,7 @@ procSuite "Waku Store":
 
     for wakuMsg in msgList:
       await subscriptions.notify("foo", wakuMsg)
+      await sleepAsync(1.millis)  # Sleep a millisecond to ensure messages are stored chronologically
     var completionFut = newFuture[bool]()
 
     proc handler(response: HistoryResponse) {.gcsafe, closure.} =
@@ -262,6 +265,7 @@ procSuite "Waku Store":
 
     for wakuMsg in msgList:
       await subscriptions.notify("foo", wakuMsg)
+      await sleepAsync(1.millis)  # Sleep a millisecond to ensure messages are stored chronologically
     var completionFut = newFuture[bool]()
 
     proc handler(response: HistoryResponse) {.gcsafe, closure.} =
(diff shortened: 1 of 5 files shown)


oskarth:
Nice find!

oskarth:
In retrospect I should probably not have merged that PR before Windows CI was merged, alternatively disable require check earlier. I just get impatient with the slow Windows CI...

tests/v2/test_waku_store.nim
```diff
@@ -88,6 +88,7 @@ procSuite "Waku Store":
     listenSwitch.mount(proto)
 
     await subscriptions.notify("foo", msg)
+    await sleepAsync(1.millis)  # Sleep a millisecond to ensure messages are stored chronologically
```
oskarth, line 91: Hmm, not sure I follow here?

oskarth, line 91: Shouldn't this be on second resolution if we are using epoch?

jm-clius: The epoch time supports subsecond resolution (up to a microsecond). 1 millisecond is therefore plenty to force floating timestamps to increase.

jm-clius:
> Nice find!

Thanks. Elation was short-lived, though, as this is still failing, but now for a different reason in a different test. 🤦 Interesting what unintended side effects changing a single type can have.

staheri14 (approved):
Nice! 👏

oskarth (approved):
err this seems a bit like magic but if it works why not :D

PR #470, opened 2021-04-08
Fix the issue in the order of fields of HistoryQuery 

staheri14:
## Problem
As discussed in https://github.com/vacp2p/rfc/issues/326, the order of fields in the `HistoryQuery` protobuf does not match the specs. Specs dictates that fields number should start from `2` , while in the implementation, fields are numbered starting from `1`. This PR is to match the implementation with the specs.


oskarth:
@staheri14 please ask specific people for review as it is easy to not see notification otherwise

oskarth (approved):

jm-clius (approved):

Issue #473, opened 2021-04-09
Fix or docs for compile time log flags


D4nte:
EDIT: This is for compiler flag, see comments below

----

When trying to use the `--log-level` argument to set the log level, it has no effect, e.g.:

Debug logs show when log level is set to `info`.
```
▶ ./build/wakunode2 --log-level=info
DBG 2021-04-09 11:03:35.776+10:00 UPnP                                       topics="nat" tid=930894 file=nat.nim:67 msg="Internet Gateway Device not found. Giving up."
DBG 2021-04-09 11:03:35.787+10:00 NAT-PMP                                    topics="nat" tid=930894 file=nat.nim:91 msg="the gateway does not support nat-pmp"
INF 2021-04-09 11:03:35.791+10:00 Initializing networking                    topics="wakunode" tid=930894 file=wakunode2.nim:127 hostAddress=/ip4/0.0.0.0/tcp/60000 announcedAddresses=@[]
```

No trace logs are show hen log level is set to `trace`

```
▶ ./build/wakunode2 --log-level=trace
DBG 2021-04-09 11:05:26.001+10:00 UPnP                                       topics="nat" tid=931087 file=nat.nim:67 msg="Internet Gateway Device not found. Giving up."
DBG 2021-04-09 11:05:26.012+10:00 NAT-PMP                                    topics="nat" tid=931087 file=nat.nim:91 msg="the gateway does not support nat-pmp"
INF 2021-04-09 11:05:26.016+10:00 Initializing networking                    topics="wakunode" tid=931087 file=wakunode2.nim:127 hostAddress=/ip4/0.0.0.0/tcp/60000 announcedAddresses=@[]
DBG 2021-04-09 11:05:26.016+10:00 About to accept incoming connection        topics="libp2p switch" tid=931087 file=switch.nim:318
DBG 2021-04-09 11:05:26.016+10:00 Started libp2p node                        topics="libp2p switch" tid=931087 file=switch.nim:368 peer="(peerId: \"16Uiu2HAmMqF1RsxqHfbYsJ2Mk8UXWiVHVG9LA57beLhNWznFqhVm\", addrs: @[\"/ip4/0.0.0.0/tcp/60000\"], protocols: @[\"/ipfs/id/1.0.0\"], protoVersion: \"\", agentVersion: \"\")"
[...]

PR #462, opened 2021-04-07
Make wakunode2 and wakunode2 default to DEBUG log

oskarth:
Logs too noisy by default, full of gossipsub TRC making it hard to see what is going on on a local node

```
=behavior.nim:260 peers=0 mesh=0 gossipsub=0 topic=/waku/2/default-waku/proto
TRC 2021-04-07 17:59:47.919+08:00 grafting                                   tid=3849991 file=behavior.nim:284 grafting=0 mesh=0 gossipsub=0 topic=/waku/2/default-waku/proto
TRC 2021-04-07 17:59:47.919+08:00 mesh balanced                              tid=3849991 file=behavior.nim:443 mesh=0 gossipsub=0 topic=/waku/2/default-waku/proto
TRC 2021-04-07 17:59:47.919+08:00 getting gossip peers (iHave)               tid=3849991 file=behavior.nim:490
TRC 2021-04-07 17:59:48.920+08:00 running heartbeat                          tid=3849991 file=behavior.nim:602 instance=140081537917000
TRC 2021-04-07 17:59:48.920+08:00 updating scores                            tid=3849991 file=scoring.nim:96 peers=0
TRC 2021-04-07 17:59:48.920+08:00 updated scores                             tid=3849991 file=scoring.nim:267 peers=0
TRC 2021-04-07 17:59:48.920+08:00 rebalancing mesh                           tid=3849991 file=behavior.nim:251 mesh=0 gossipsub=0 topic=/waku/2/default-waku/proto
TRC 2021-04-07 17:59:48.920+08:00 replenishing mesh                          tid=3849991 file=behavior.nim:260 peers=0 mesh=0 gossipsub=0 topic=/waku/2/default-waku/proto
TRC 2021-04-07 17:59:48.920+08:00 grafting                                   tid=3849991 file=behavior.nim:284 grafting=0 mesh=0 gossipsub=0 topic=/waku/2/default-waku/proto
TRC 2021-04-07 17:59:48.920+08:00 mesh balanced                              tid=3849991 file=behavior.nim:443 mesh=0 gossipsub=0 topic=/waku/2/default-waku/proto
TRC 2021-04-07 17:59:48.920+08:00 getting gossip peers (iHave)               tid=3849991 file=behavior.nim:490
TRC 2021-04-07 17:59:49.921+08:00 running heartbeat                          tid=3849991 file=behavior.nim:602 instance=140081537917000
[...]

PR #474, opened 2021-04-09
Revert "Make wakunode2 and wakunode2 default to DEBUG log (#462)"

D4nte:
This reverts commit 139f1480 has it removes the ability to get any trace logs due to https://github.com/status-im/nim-waku/issues/473

As it introduced a feature regression, I would suggest to remove it until #473 is fixed.

### DIFF
--- a/waku.nimble
+++ b/waku.nimble
@@ -44,7 +44,7 @@ proc test(name: string, lang = "c") =
 
 ### Waku v1 tasks
 task wakunode1, "Build Waku v1 cli node":
-  buildBinary "wakunode1", "waku/v1/node/", "-d:chronicles_log_level=DEBUG"
+  buildBinary "wakunode1", "waku/v1/node/", "-d:chronicles_log_level=TRACE"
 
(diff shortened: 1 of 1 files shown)


oskarth:
Huh? Just do this in your local checkout if you need it...

D4nte:
> Huh? Just do this in your local checkout if you need it...

I also need it in my GitHub actions. I worked around it by setting my nim-waku submodule to this `revert-debug` branch.

oskarth:
Err...I don't know why your CI is relying on TRACE logs? That sounds like an anti pattern if I ever saw one.

DEBUG logs should be the default experience (later on maybe even just INFO/WARN). We only put it to TRACE to debug some issues in cluster temporarily, now with (fixed) gossipsub it is way too noisy for default. TRACE is for detailed debugging by devs, not something normal users should care about.

oskarth (changes requested):
This is not a regression, it was and should be default. The bug here is having a CI that relies on TRACE logs.

PR #483, opened 2021-04-12
Fix: `filter` protocol only works on default `topic`

### LINKED ISSUE #480: `Filter` protocol only works on default `topic`
`Filter` protocol currently fails for messages not published on the default topic.

This is because `notify` is only called on `wakunode` subscriptions for messages published on the default topic and is therefore likely to affect the `store` protocol in a similar manner.

### How to reproduce:

Consider the following steps:
1. Start a full node `A` with `filter` and `relay` protocols mounted
2. Start a light node `B`, with `filter` protocol mounted and `A` set as the `filternode`
3. Subscribe `A` to some pubsub `topic`, say `my_topic`.
4. Subscribe `B` to some contentTopic `ct1`. This creates a subscription entry for `ct1` against `B` in `A`, as expected.

If you now publish/receive a `relay` message on `A` with `contentTopic` set to `ct1` it will only be pushed to `B` if this message is published on the default topic (`/waku/2/default-waku/proto`). `my_topic`, and any other topic, will fail.

This bug was noticed by WalletConnect while trying to subscribe to content filters on their global topic. It does not show up in CI tests, as all filter tests either make use of the default topic or bypass the node's own notification process to force a `notify` on a subscription.

### Acceptance criteria:

1. Content subscription in the `filter` protocol should work, regardless of which pubsub `topic` a message is published on.
2. A content filter could _optionally_ contain a pubsub `topic` field, which should narrow matches to only messages published on that `topic`.
3. CI tests for `filter` should also cover cases where pubsub `topic` is not the default.

jm-clius:
This PR fixes #480 

It adds a default handler for all newly subscribed PubSub topics to ensure that the node's `subscriptions` and `filters` are notified of all new messages. This ensures that content filtering works for messages received on any subscribed topic, not just the `waku` default.

Content filters can still optionally specify a pubsub `topic`, which will narrow matches down to only messages published on that `topic`.

### DIFF
--- a/waku/v2/node/wakunode2.nim
+++ b/waku/v2/node/wakunode2.nim
@@ -176,16 +176,37 @@ proc stop*(node: WakuNode) {.async.} =
 
   node.started = false
 
+proc subscribe(node: WakuNode, topic: Topic, handler: Option[TopicHandler]) =
+  info "subscribe", topic=topic
+
+  proc defaultHandler(topic: string, data: seq[byte]) {.async, gcsafe.} =
+    # A default handler should be registered for all topics
+    trace "Hit default handler", topic=topic, data=data
+
+    let msg = WakuMessage.init(data)
+    if msg.isOk():
+      node.filters.notify(msg.value(), "")  # Trigger filter handlers on a light node
+      await node.subscriptions.notify(topic, msg.value()) # Trigger subscription handlers on a store/filter node
+      waku_node_messages.inc(labelValues = ["relay"])
+
+  let wakuRelay = node.wakuRelay
+
+  if topic notin PubSub(wakuRelay).topics:
+    # Add default handler only for new topics
+    debug "Registering default handler", topic=topic
+    wakuRelay.subscribe(topic, defaultHandler)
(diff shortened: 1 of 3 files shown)


staheri14 (approved):
Nice!

waku/v2/node/wakunode2.nim
```diff
@@ -176,16 +176,37 @@ proc stop*(node: WakuNode) {.async.} =
...
+  if topic notin PubSub(wakuRelay).topics:
+    # Add default handler only for new topics
+    debug "Registering default handler", topic=topic
+    wakuRelay.subscribe(topic, defaultHandler)
+
+  if handler.isSome:
+    debug "Registering handler", topic=topic
+    wakuRelay.subscribe(topic, handler.get())
+
 proc subscribe*(node: WakuNode, topic: Topic, handler: TopicHandler) =
```
oskarth, line 203: Is there a specific reason not to make the Option[TopicHandler] the default public signature?

oskarth (approved):

PR #488, opened 2021-04-13
Update store protocol interface: add history content filter

staheri14:
This PR is an increment towards https://github.com/status-im/nim-waku/issues/485
It modifies the structure of the `HistoryRequest`  by replacing the `topics` field  with a sequence of `HistoryContentFilter`s.
Each `ContentFilter` encapsulates a single content topic (later on, more fields can be embedded inside a `HistoryContentFilter`). 

- The protobuf's encode and init procedures are modified accordingly.
- A unit test is developed to check the correct retrieval of historical messages for queries with more than one `HistoryContentFilter`.
- Other parts of codebase which relied on `HistoryQuery` are refined.
- This change also affected the store json rpc api. The internals of "get_waku_v2_store_v1_messages" is changed to convert the set of input `topics` to their `HistoryContentFilter` equivalence. However, the proc input is left untouched. The reason for not yet changing the proc signature is explained in this issue https://github.com/status-im/nim-waku/issues/489. 

## Future steps
The specs ~~shall~~ mirror the change in the `HistoryRequest` protobuf message as indicated in this PR https://github.com/vacp2p/rfc/pull/344.

### DIFF
--- a/waku/v2/node/jsonrpc/store_api.nim
+++ b/waku/v2/node/jsonrpc/store_api.nim
@@ -17,7 +17,7 @@ proc installStoreApiHandlers*(node: WakuNode, rpcsrv: RpcServer) =
 
   ## Store API version 1 definitions
 
-  rpcsrv.rpc("get_waku_v2_store_v1_messages") do(contentFilters: seq[HistoryContentFilter], pagingOptions: Option[StorePagingOptions]) -> StoreResponse:
+  rpcsrv.rpc("get_waku_v2_store_v1_messages") do(topics: seq[ContentTopic], pagingOptions: Option[StorePagingOptions]) -> StoreResponse:
     ## Returns history for a list of content topics with optional paging
     debug "get_waku_v2_store_v1_messages"
 
@@ -27,6 +27,9 @@ proc installStoreApiHandlers*(node: WakuNode, rpcsrv: RpcServer) =
       debug "get_waku_v2_store_v1_messages response"
       responseFut.complete(response.toStoreResponse())
     
+    var contentFilters: seq[HistoryContentFilter] = @[]
+    for topic in topics:
+      contentFilters.add(HistoryContentFilter(topic: topic))
     let historyQuery = HistoryQuery(contentFilters: contentFilters,
                                     pagingInfo: if pagingOptions.isSome: pagingOptions.get.toPagingInfo() else: PagingInfo())
     
--- a/waku/v2/node/jsonrpc/store_api.nim
+++ b/waku/v2/node/jsonrpc/store_api.nim
@@ -17,7 +17,7 @@ proc installStoreApiHandlers*(node: WakuNode, rpcsrv: RpcServer) =
 
   ## Store API version 1 definitions
 
-  rpcsrv.rpc("get_waku_v2_store_v1_messages") do(topics: seq[ContentTopic], pagingOptions: Option[StorePagingOptions]) -> StoreResponse:
+  rpcsrv.rpc("get_waku_v2_store_v1_messages") do(contentFilters: seq[HistoryContentFilter], pagingOptions: Option[StorePagingOptions]) -> StoreResponse:
     ## Returns history for a list of content topics with optional paging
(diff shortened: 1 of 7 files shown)


oskarth (commented):
Is there a corresponding spec PR?

What's the rationale for having it be HistoryContentFilter as opposed to ContentFilter?

waku/v2/protocol/waku_store/waku_store_types.nim
```diff
@@ -17,6 +17,8 @@ export pagination
 const MaxPageSize* = uint64(100) # Maximum number of waku messages in each page
 
 type
+  HistoryContentFilter* = object
+    topic*: ContentTopic
```
jm-clius, line 21: I'd suggest we name this now `contentTopic` to avoid confusion as per [this discussion](https://github.com/vacp2p/rfc/issues/327)

waku/v2/protocol/waku_store/waku_store_types.nim
```diff
@@ -37,7 +39,7 @@ type
     direction*: PagingDirection
 
   HistoryQuery* = object
-    topics*: seq[ContentTopic]
+    contentFilters*: seq[HistoryContentFilter]
```
jm-clius, line 42: I understood a history query should also now optionally define a pubsub `topic`, or is that for a next step?

staheri14:
> What's the rationale for having it be HistoryContentFilter as opposed to ContentFilter?
@oskarth 
ContentFilter and HistoryContentFilter have different structures, the former embodies a sequence of content topics whereas the latter encapsulates a single content topic. To avoid any confusion, I haven't used the same identifier for both.

waku/v2/protocol/waku_store/waku_store_types.nim
staheri14: That is right, it is going to be added in the next step.

waku/v2/protocol/waku_store/waku_store_types.nim
staheri14: right! that makes sense! will do.

staheri14:
> Is there a corresponding spec PR?

@oskarth 
https://github.com/vacp2p/rfc/pull/344

waku/v2/node/jsonrpc/store_api.nim
```diff
@@ -27,7 +27,10 @@ proc installStoreApiHandlers*(node: WakuNode, rpcsrv: RpcServer) =
       debug "get_waku_v2_store_v1_messages response"
       responseFut.complete(response.toStoreResponse())
     
-    let historyQuery = HistoryQuery(topics: topics,
+    var contentFilters: seq[HistoryContentFilter] = @[]
+    for topic in topics:
```
oskarth, line 31: I'd probably either add some comment here or be more explicit with contentTopic here, otherwise we can easily introduce bugs in future with topic/contenttopic confusion

waku/v2/protocol/waku_store/waku_store.nim
```diff
@@ -110,16 +110,29 @@ proc init*(T: type PagingInfo, buffer: seq[byte]): ProtoResult[T] =
   pagingInfo.direction = PagingDirection(direction)
 
   ok(pagingInfo) 
-  
+
+proc init*(T: type HistoryContentFilter, buffer: seq[byte]): ProtoResult[T] =
+  let pb = initProtoBuffer(buffer)
+
+  var topic: ContentTopic
```
oskarth, line 117: Ditto here, I'd be more explicit with contentTopic or even "ct" would be better than "topic"

oskarth (approved):

jm-clius (approved):
LGTM!

waku/v2/node/jsonrpc/store_api.nim
staheri14: Sure, addressed!

waku/v2/protocol/waku_store/waku_store.nim
staheri14: Sure, addressed.

Issue #495, opened 2021-04-19
Statically-defined peers should be considered `directPeers`


jm-clius:
### Background

Currently, all statically-defined peers for a `wakunode2` are simply dialled and, in a `relay` context, will be considered candidates for the GossipSub `mesh` from where they can be pruned/grafted according to spec. However, we expect the node to unconditionally exchange messages with these peers, always maintain a connection and persist these peers after a restart. This can be done through an [explicit peering agreement](https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.1.md#explicit-peering-agreements), which requires static peers to be added to the `directPeers` list.

### Acceptance Criteria

- [ ] Statically-defined nodes undergo explicit peering agreements and persist smoothly across restarts.

Related issue: https://github.com/status-im/nim-waku/issues/487

jm-clius:
On closer inspection, an explicit peering agreement is not suitable for statically-defined `wakunode2` peers, for the following reasons:
1. `directPeer` setup has to be reciprocal, according to specification. It is an error to GRAFT an explicit peer, so the peers must be mutually marked as direct. For our use cases we often want to connect to a static peer that might not consider us as direct.
2. `directPeer` must be set up out-of-band, when the peer is started. We have several use cases where we want to add static peers to a node that has already started.

The idea to move to explicit peering agreements was triggered by peer persistence issues (#487). An alternative solution has been implemented in #497.

Issue #487, opened 2021-04-13
`test` cluster nodes do not form mesh on default topic


jm-clius:
The [`wakuv2.test` cluster](https://fleets.status.im/) nodes do not form a single mesh for the default topic after a redeployment and restart.

Previously this was caused by a lack of peer persistence across restarts. The logs now suggest that peer persistence works and that each node successfully reconnects to its peers after starting:

```
DBG 2021-04-13 06:15:36.554+00:00 Dial successful                            topics="libp2p switch" tid=1 file=switch.nim:154 conn=16U*3q4YjS:6075370718777d6b2166c245 peerInfo="(peerId: "16Uiu2HAmJb2e28qLXxT5kZxVUUoJt72EMzNGXB47Rxx5hw3q4YjS", addrs: @["/ip4/0.0.0.0/tcp/30303"], protocols: @["/ipfs/id/1.0.0", "/vac/waku/store/2.0.0-beta1", "/vac/waku/filter/2.0.0-beta1", "/vac/waku/relay/2.0.0-beta2"], protoVersion: "ipfs/0.1.0", agentVersion: "nim-libp2p/0.0.1")"
```

However, no messages are relayed between nodes on the default topic. This may be related to the timing of the peer reconnection and subscription to the default topic.

After about an hour the peers are unsubscribed and disconnected.

```
DBG 2021-04-13 07:01:13.843+00:00 Could not establish send connection        topics="libp2p pubsubpeer" tid=1 file=pubsubpeer.nim:199 msg="Couldn't get muxed stream"
DBG 2021-04-13 07:01:13.843+00:00 exiting pubsub read loop                   topics="libp2p pubsubpeer" tid=1 file=pubsubpeer.nim:154 conn=16U*3q4YjS:6075372d18777d6b2166c259 peer=16U*3q4YjS closed=true
DBG 2021-04-13 07:01:13.843+00:00 Timeout handler cancelled                  topics="libp2p connection" tid=1 file=connection.nim:121 s=16U*3q4YjS:6075372d18777d6b2166c257
DBG 2021-04-13 07:01:13.843+00:00 unsubscribing pubsub peer                  topics="libp2p pubsub" tid=1 file=pubsub.nim:117 peerId=16U*3q4YjS
```

cc @D4nte

jm-clius:
After some analysis, I've noticed the following libp2p behaviour whenever we deploy to the `test` cluster:

#### Summary

[...]

PR #497, opened 2021-04-20
Add persistent backoff for peers

jm-clius:
This PR addresses #487. It is an alternative approach to persisting statically-defined peers than #495.

- It adds a persistent backoff period for peers after disconnection based on the `pruneBackoff` and `BackoffSlackTime` defined for [GossipSub 1.1](https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.1.md#prune-backoff-and-peer-exchange).
- Peer persistence is now a separately configurable option. Since this is a specialised use case, it is disabled by default.

### DIFF
--- a/waku/v2/node/peer_manager/peer_manager.nim
+++ b/waku/v2/node/peer_manager/peer_manager.nim
@@ -1,7 +1,7 @@
 {.push raises: [Defect, Exception].}
 
 import
-  std/[options, sets, sequtils],
+  std/[options, sets, sequtils, times],
   chronos, chronicles, metrics,
   ./waku_peer_store,
   ../storage/peer/peer_storage
@@ -20,8 +20,8 @@ type
     peerStore*: WakuPeerStore
     storage: PeerStorage
 
-const
-  defaultDialTimeout = 1.minutes # @TODO should this be made configurable?
+let
+  defaultDialTimeout = chronos.minutes(1) # @TODO should this be made configurable?
 
 ####################
 # Helper functions #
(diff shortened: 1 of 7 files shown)


waku/v2/node/peer_manager/peer_manager.nim
```diff
@@ -195,13 +198,31 @@ proc selectPeer*(pm: PeerManager, proto: string): Option[PeerInfo] =
...
   
   for storedInfo in pm.peers(proto):
+    # Check if peer is reachable.
+    if pm.peerStore.connectionBook.get(storedInfo.peerId) == CannotConnect:
+      debug "Not reconnecting to unreachable peer", peerId=storedInfo.peerId
+      continue
+    
+    # Respect optional backoff period where applicable.
+    let
+      disconnectTime = Moment.init(pm.peerStore.disconnectBook.get(storedInfo.peerId), Second)  # Convert
```
staheri14, line 215: Curious why this conversion is needed?

staheri14 (approved):
Nice!

waku/v2/node/storage/peer/waku_peer_storage.nim
```diff
@@ -82,19 +84,20 @@ proc new*(T: type WakuPeerStorage, db: SqliteDatabase): PeerStorageResult[T] =
...
             storedInfo: StoredInfo,
-            connectedness: Connectedness): PeerStorageResult[void] =
+            connectedness: Connectedness,
+            disconnectTime: int64): PeerStorageResult[void] =
 
   ## Adds a peer to storage or replaces existing entry if it already exists
   let prepare = db.database.prepareStmt(
-    "REPLACE INTO Peers (peerId, storedInfo, connectedness) VALUES (?, ?, ?);",
-    (seq[byte], seq[byte], int32),
+    "REPLACE INTO Peer (peerId, storedInfo, connectedness, disconnectTime) VALUES (?, ?, ?, ?);",
```
oskarth, line 92: This is outside of scope of PR, but I wonder if we want to do REPLACE here? Might make sense to make this an append-log and then take last entry. The historical state could be useful to understand connect history, perhaps?

oskarth (approved):

waku/v2/node/peer_manager/peer_manager.nim
jm-clius: `chronos` library makes use of the concepts [`Moment` and `Duration`](https://github.com/status-im/nim-chronos/blob/master/chronos/timer.nim#L180-L182), which on their own has no direct meaning but is a fast and efficient way to compare different moments in time.  The `Duration` used to `sleepAsync`, for example, points to this library.
In contrast, the Nim `times` module defines absolute time and duration in terms of the epoch/system time (`getTime()`). This is a roundabout way of saying, that since I need to compare Moments across restarts I convert a saved epoch time from `std/times` to a Moment in `chronos` for ease of use. Hope this explanation makes sense. 😁

waku/v2/node/storage/peer/waku_peer_storage.nim
jm-clius: True. And then rather perform `UPDATE` during a single connection session, which will likely be more efficient.

waku/v2/node/peer_manager/peer_manager.nim
staheri14: Yeah, that makes sense! thanks!

PR #499, opened 2021-04-21
Updates store json rpc api

staheri14:
Closes https://github.com/status-im/nim-waku/issues/489
The corresponding specs PR https://github.com/vacp2p/rfc/pull/356

### DIFF
--- a/tests/v2/test_jsonrpc_waku.nim
+++ b/tests/v2/test_jsonrpc_waku.nim
@@ -21,7 +21,7 @@ import
                               private_api],
   ../../waku/v2/protocol/message_notifier,
   ../../waku/v2/protocol/waku_relay,
-  ../../waku/v2/protocol/waku_store/waku_store,
+  ../../waku/v2/protocol/waku_store/[waku_store, waku_store_types],
   ../../waku/v2/protocol/waku_swap/waku_swap,
   ../../waku/v2/protocol/waku_filter/waku_filter,
   ../test_helpers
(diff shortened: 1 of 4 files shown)


tests/v2/test_jsonrpc_waku.nim
```diff
@@ -259,7 +259,7 @@ procSuite "Waku v2 JSON-RPC API":
     let client = newRpcHttpClient()
     await client.connect("127.0.0.1", rpcPort)
 
-    let response = await client.get_waku_v2_store_v1_messages(@[defaultContentTopic], some(StorePagingOptions()))
+    let response = await client.get_waku_v2_store_v1_messages(@[HistoryContentFilter(contentTopic: defaultContentTopic)], some(StorePagingOptions()))
```
oskarth, line 262: What happens if we don't pass any content filter here?

jm-clius (approved):
LGTM!

tests/v2/test_jsonrpc_waku.nim
jm-clius: I think the underlying `store` still expects a `HistoryQuery` with populated contentFilters, but I also think changing to an `Option` will be an improvement.

staheri14: The store json rpc api was designed this way with `topics` filed which is now `contentFilters`  being mandatory. I also agree that making it optional makes sense and would also match the store protocol api.

staheri14: Opened an issue re making it optional: https://github.com/status-im/nim-waku/issues/505

oskarth (approved):

Issue #498, opened 2021-04-21
Issue: `WakuFilter` requires `WakuRelay`


jm-clius:
## Background
Currently the `WakuFilter` protocol requires `WakuRelay` to be mounted on a light node. This is because `WakuFilter` itself is not a PubSub protocol but in its current form requires PubSub functionality on at least three counts:
1. **In order to subscribe topic handlers:** this is necessary for `WakuFilter` to be notified of new messages. Currently topic handlers can only be registered via the `WakuRelay` protocol.
2. **In order to publish:** even if the above were not true, a node with only `WakuFilter` mounted would not be able to publish messages without having a PubSub-compatible protocol mounted. `WakuRelay` is currently the only such protocol implemented, though a [`LIGHTPUSH` protocol](https://github.com/vacp2p/rfc/issues/354) is under discussion.
3. **In order to dial the full node:** the full node, where `WakuRelay` is mounted by definition, will [attempt to subscribe](https://github.com/status-im/nim-libp2p/blob/master/libp2p/protocols/pubsub/pubsub.nim#L374-L379) the light node as a pubsub peer when dialed and negotiate a PubSub connection. However, if `WakuRelay` is not mounted on the light node the connection attempt will fail. This currently prohibits light nodes without `WakuRelay` mounted from dialing full nodes at all, including to subscribe to a content filter.

## Why do we not want to mount `WakuRelay` on a light node?
Once a light node with `WakuRelay` mounted has opened a connection to a full node (say, to subscribe a content filter), it will be considered a "normal" PubSub peer in the full node, undergo grafting and pruning and generally behave like a full node.

oskarth:
Possibly candidate for release scope

jm-clius:
The logs show the following is happening. Note that although `RELAY` and `FILTER` protocols are used as example, this seems to be true for any two nodes trying to open a stream for a non-`PubSub` protocol where `PubSub` is mounted on only one of the nodes.

@dryajov @sinkingsugar, it would be great to hear your thoughts! This may be expected behaviour, but that would mean that we're using the `switch` in a naïve way for our non-`PubSub` protocols.

Say we have two Waku nodes, **RF** and **F**. **RF** has both `RELAY` and `FILTER` protocols mounted (full node); **F** has only `FILTER` mounted (light node). `RELAY` closely mirrors `GossipSub 1.1`. `FILTER` is _not_ a `PubSub` protocol.
 
[...]

PR #501, opened 2021-04-22
Fix filter notifications

### LINKED ISSUE #500: Bug: Light `FILTER` nodes receive duplicate messages on filters
### Problem

Light nodes are expected to be notified only _once_ of each message matching their subscribed filters in the associated full node. However, if the light node subscribed using a `FilterRequest` that contains a `pubSubTopic` it is notified of each matching message twice. This is a bug.

### Acceptance Criteria
- [ ] Light nodes no longer receive duplicate messages

jm-clius:
This PR is a first increment towards fixing #498 and fixes #500

It does the following:
 - helps separate `FILTER` from `RELAY` by removing light node notification logic from default `RELAY` topic handlers. Light nodes are now only notified about new filter messages in the filter push message handler. Since `RELAY` is mounted on light nodes, this double `notify()` also caused duplicate messages on the light node (#500).
 - fixes full `FILTER` node logic that would previously only push messages to a light node if it had defined a `pubSubTopic` in the filter. `12/WAKU-FILTER`, however, specifies that a `FilterRequest` only optionally needs to specify a `pubsubTopic` - an empty `pubsubTopic` implies that matching should only be done on the `contentFilter`.
 - changes ambiguous `topic` variable names to `pubSubTopic` and `contentTopic` within `FILTER` context

### Further improvements:
- where fields are optional they should explicitly use `Option` from the `std/options` module. Currently we rely on empty initialisations to test whether optional fields are set or not. This approach is prone to errors.

### DIFF
--- a/examples/v2/chat2.nim
+++ b/examples/v2/chat2.nim
@@ -318,7 +318,7 @@ proc processInput(rfd: AsyncFD, rng: ref BrHmacDrbgContext) {.async.} =
       info "Hit filter handler"
 
     await node.subscribe(
-      FilterRequest(contentFilters: @[ContentFilter(topics: @[DefaultContentTopic])], topic: DefaultTopic, subscribe: true),
+      FilterRequest(contentFilters: @[ContentFilter(contentTopics: @[DefaultContentTopic])], pubSubTopic: DefaultTopic, subscribe: true),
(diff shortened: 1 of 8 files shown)


oskarth (approved):
LGTM

PR #507, opened 2021-04-23
Fix store protocol identifier

staheri14:
Closes https://github.com/status-im/nim-waku/issues/456
The store protocol identifier was behind the identifier indicated in the specs.

### DIFF
--- a/tests/v2/test_wakunode.nim
+++ b/tests/v2/test_wakunode.nim
@@ -568,67 +568,67 @@ procSuite "WakuNode":
     
     await allFutures([node1.stop(), node2.stop()])
 
-  asyncTest "Lightpush message return success":
-    let
-      nodeKey1 = crypto.PrivateKey.random(Secp256k1, rng[])[]
-      node1 = WakuNode.init(nodeKey1, ValidIpAddress.init("0.0.0.0"),
-        Port(60000))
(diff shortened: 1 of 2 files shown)


oskarth (approved):

oskarth:
Deployment here is a bit tricky, we are still in development so not a big deal, but if WalletConnect or someone is assuming beta1 and we update cluster to beta2 then that might break expectations.

The correct libp2p way is to advertise compatible versions, so we'd announce both beta2 and beta1, then each node would picker the higher one. We have not yet exposed this. See https://github.com/status-im/nim-waku/issues/165

D4nte (approved):

staheri14:
@oskarth I am getting the same exact windows error as this PR https://github.com/status-im/nim-waku/pull/506/checks#step:7:1332. Going to merge as it seems the error is not caused by the current PR.

FYI, I was trying to spot the cause of the error, and  have commented out "Lightpush message return success" test and the error disappeared. Here is the link to the CI checks for that commit https://github.com/status-im/nim-waku/pull/507/checks?check_run_id=2451519244

oskarth:
Oh weird that that test would trigger it and on Windows, nice spot! Sorry about that.

PR #508, opened 2021-04-24
A bug fix re pagination

staheri14:
There seems to be an issue regarding index comparison in the pagination process. The compareIndex proc marks identical indexes as unequal. This problem is reported by walletConnect, however, the comparison is indeed working correctly in our unit tests and  I could not still replicate the problem.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,3 +34,5 @@ rln
 package-lock.json
 node_modules/
 /.update.timestamp
+
+libwaku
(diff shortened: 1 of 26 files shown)


oskarth:
What's the state of this PR @staheri14?

staheri14:
@oskarth Still in progress, could not replicate the issue reported by walletconnect, though I can close the PR till the issue gets solved.

PR #510, opened 2021-04-26
Libwaku spike start

oskarth:
Just playing around with C interop over the weekend, might be nice to merge partial progress.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,3 +34,5 @@ rln
 package-lock.json
 node_modules/
 /.update.timestamp
+
+libwaku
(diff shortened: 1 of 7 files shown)


oskarth:
I was messing around with linking errors, static/dynamic lib and undefined errors, strong deja vu, when I realized things had moved a bit here: https://github.com/status-im/nimbus-eth1/tree/master/wrappers - next step is probably to incorporate all the snazzy Makefile linker magic that is done here, porting it to nim-waku, and hopefully it'll work then (TM).

jm-clius (approved):
Great to see this take shape!

PR #518, opened 2021-04-28
Allow relay=false for light filter nodes

jm-clius:
This PR addresses #498. It is closer to a workaround than a permanent fix.

It allows running `FILTER` protocol on a light node without enabling `RELAY` protocol. A known issue, however, requires `RELAY` to be mounted on the light node in order to dial the full node where `RELAY` _is_ mounted. See [this comment](https://github.com/status-im/nim-waku/issues/498#issuecomment-828543874) for details. We therefore still initialise and mount `RELAY`. `WakuRelay` is never started, though, and the node is never subscribed to any PubSub topics, so it does not participate in any full-message peerings. This way it behaves like a "true" light node.

### DIFF
--- a/waku/v2/node/wakunode2.nim
+++ b/waku/v2/node/wakunode2.nim
@@ -180,6 +180,11 @@ proc stop*(node: WakuNode) {.async.} =
   node.started = false
 
 proc subscribe(node: WakuNode, topic: Topic, handler: Option[TopicHandler]) =
+  if node.wakuRelay.isNil:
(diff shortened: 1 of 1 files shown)


staheri14 (approved):
Looks good to me! Just as a suggestion, a flag may be defined for the light mode of the relay protocol e.g., `--light`. As such, having the relay protocol (no matter in the light mode or full mode) mounted would be a prerequisite for the filter protocol.

oskarth (approved):

PR #519, opened 2021-04-28
Enables perssist-message flag in the store protocol for wakunode2

staheri14:
Addresses https://github.com/status-im/nim-waku/issues/494
It introduces `persist-messages` flag to activate the store protocol  in the light mode. It defaults to `false`.
This flag impacts wakunode2 and disables persisting messages in a database. 

However, as far as I know, in chat2  there is no option to persist messages. Please correct me if I am mistaken @jm-clius @oskarth .
In chat2, the store protocol is mounted in https://github.com/status-im/nim-waku/blob/7e9dac67c8927c953687ee6391c63286377bb9c9/examples/v2/chat2.nim#L280
which invokes this proc whose `store` field is Nil by default
https://github.com/status-im/nim-waku/blob/7e9dac67c8927c953687ee6391c63286377bb9c9/waku/v2/node/wakunode2.nim#L362

### DIFF
--- a/waku/v2/node/config.nim
+++ b/waku/v2/node/config.nim
@@ -60,7 +60,7 @@ type
 
     store* {.
       desc: "Enable store protocol: true|false",
-      defaultValue: false
+      defaultValue: true
       name: "store" }: bool
 
     filter* {.
--- a/waku/v2/node/config.nim
+++ b/waku/v2/node/config.nim
@@ -55,7 +55,7 @@ type
 
     persistmessages* {.
       desc: "Enable message persistence: true|false",
-      defaultValue: false
+      defaultValue: true
       name: "persist-messages" }: bool
 
     store* {.
--- a/waku/v2/node/config.nim
+++ b/waku/v2/node/config.nim
@@ -55,7 +55,7 @@ type
 
     persistmessages* {.
       desc: "Enable message persistence: true|false",
-      defaultValue: true
+      defaultValue: false
       name: "persist-messages" }: bool
 
     store* {.
--- a/waku/v2/node/config.nim
+++ b/waku/v2/node/config.nim
@@ -53,6 +53,11 @@ type
       defaultValue: ""
       name: "storenode" }: string
 
+    persistmessages* {.
+      desc: "Enable message persistence: true|false",
+      defaultValue: false
+      name: "persist-messages" }: bool
+
     store* {.
       desc: "Enable store protocol: true|false",
       defaultValue: false
--- a/docs/tutorial/chat2.md
+++ b/docs/tutorial/chat2.md
@@ -47,7 +47,7 @@ The `chat2` application can retrieve historical chat messages from a node suppor
 Alternatively, the `chat2` application will select a random `storenode` for you from the test fleet if `storenode` left unspecified.
 
 ```
-./build/chat2 --store:true
+./build/chat2
 ```
 
 > *NOTE: Currently (Mar 3, 2021) the only node in the test fleet that provides reliable store functionality is `/ip4/134.209.139.210/tcp/30303/p2p/16Uiu2HAmPLe7Mzm8TsYUubgCAW1aJoeFScxrLj8ppHFivPo97bUZ`. We're working on fixing this.*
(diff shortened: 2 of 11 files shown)


oskarth (approved):
Probably want to set `store` to true as well by default

jm-clius:
> in chat2 there is no option to persist messages
Correct. `chat2` should retrieve history from one of the cluster nodes

jm-clius (commented):
Though this allows disabling persisting messages to the database, a node with `store=true` will still locally "store" messages in-memory in the `messages` sequence. I understood from [the previous discussion](https://github.com/status-im/nim-waku/issues/494#issuecomment-825706738) that the flag would cover this case as well, or do we want to configure this separately?

oskarth:
Correct, that case should ideally be covered as well

staheri14:
> Though this allows disabling persisting messages to the database, a node with `store=true` will still locally "store" messages in-memory in the `messages` sequence. I understood from [the previous discussion](https://github.com/status-im/nim-waku/issues/494#issuecomment-825706738) that the flag would cover this case as well, or do we want to configure this separately?

Thanks @jm-clius and @oskarth ! 

Further changes on the PR to be reviewed:

1. The functionality of the light node is now limited to sending history query and getting response. 
The storage of waku messages is totally disabled. This is done by disabling the subscription part when the persist-messages flag is false. This means a light node will not receive any waku messages hence wont store any messages. (Alternatively, instead of disabling the subscription,  an empty handler can be set when light mode is active . However, I am not sure about any future usecase for such implementation. wdyt?)

2. The default of `persist-messages` is also set to `true`, this is because persistent storage of messages has been a default behaviour of the store protocol and is used in this way in all the waku v2 tutorials and instructions. Light nodes will have to set the `persist-messages` to false.

3- A similar  `persist-messages` flag is included in the config_bridge, to be able to configure the mode of store protocol while bridging between v1 and v2, please let me know whether the use of flag in this context is correct, or should the store protocol be always mounted in the full mode? @jm-clius

waku/v2/node/config.nim
```diff
@@ -55,12 +55,12 @@ type
 
     persistmessages* {.
       desc: "Enable message persistence: true|false",
-      defaultValue: false
+      defaultValue: true
```
oskarth, line 58: Do we want this to be true? To me the most common path is to be a light store client, but I guess one could argue both ways

oskarth (approved):
> The default of persist-messages is also set to true, this is because persistent storage of messages has been a default behaviour of the store protocol and is used in this way in all the waku v2 tutorials and instructions. Light nodes will have to set the persist-messages to false.

Intuitively I'd probably set this to false, especially since `store` was false by default before. For the tutorial, I'd just update the `store.md` one to replace that flag with `persist-mesages`. WDYT? It isn't a big deal though and we could also change this later if we want to.

waku/v2/node/config.nim
jm-clius: I'm leaning towards `false` as well. In my view, the most common "default" client is one that starts up, retrieves history as a "light" store node and participates in the network as a relay node. Thus, defaults of `store=true`, `relay=true`, `persist-messages=false`. Such clients will, of course, have to be configured with a `storenode` as well.

jm-clius (approved):
LGTM! There is just the conversation around `persist-messages` being `false` default. As for
> similar persist-messages flag is included in the config_bridge

<-- absolutely fine as is. I don't foresee a bridge ever doubling as a storenode (which is why `store` should remain `false` by default for the bridge), but it's good to keep the config options up to date

staheri14:
Thanks @oskarth and @jm-clius 
I updated the defaults to `store=true`, `persist-messages=false` both in wakunode2 (and chat2 which inherits the wakunode2 config) and in the bridge config. 
The instructions around the `persist-messages` is also included in the `store.md` file as well as `chat2.md`.

I would be thankful if you please have a quick look at these recent  updates and feel free to merge this PR if things look OK.

oskarth (approved):

jm-clius (approved):

PR #520, opened 2021-04-29
Rust is needed to compile the project

D4nte:


oskarth:
Ideally this would be optional and only apply when compiling the RLN module...

oskarth (approved):

Issue #504, opened 2021-04-22
Cluster nodes time out connections after a while


jm-clius:
## Problem

A well-balanced and live cluster `mesh` will seemingly disintegrate after a while if left alone. Even if all nodes started up without issue, reconnected to every other node in the cluster and managed to form a single, stable network, it seems that after a time period of inactivity (+-30 minutes) their mutual connections will time out and fail to be reestablished.

Log snippet:

```
DBG 2021-04-22 14:31:33.179+00:00 Timeout handler cancelled                  topics="libp2p connection" tid=1 file=connection.nim:121 s=16U*o97bUZ:608182e90676ed285743a3ee
DBG 2021-04-22 14:31:33.179+00:00 exiting pubsub read loop                   topics="libp2p pubsubpeer" tid=1 file=pubsubpeer.nim:154 conn=16U*o97bUZ:6081866d0676ed285743a425 peer=16U*o97bUZ closed=true
DBG 2021-04-22 14:31:33.179+00:00 exiting pubsub read loop                   topics="libp2p pubsubpeer" tid=1 file=pubsubpeer.nim:154 conn=16U*o97bUZ:6081866d0676ed285743a424 peer=16U*o97bUZ closed=true
DBG 2021-04-22 14:31:33.183+00:00 unsubscribing pubsub peer                  topics="libp2p pubsub" tid=1 file=pubsub.nim:117 peerId=16U*o97bUZ
DBG 2021-04-22 14:31:33.184+00:00 Could not establish send connection        topics="libp2p pubsubpeer" tid=1 file=pubsubpeer.nim:199 msg="Couldn't get muxed stream"
```

There are some gaps until we've tested with `TRACE` logging, but the above seems to suggest that a timeout was triggered and its handler then cancelled, likely due to a connection being closed from elsewhere.

## Acceptance criteria
- [ ] Connections do not time out and the cluster `mesh` can sustain itself indefinitely.

jm-clius:
Not much gained (yet) from trace logging except to confirm the sequence of events above and gain a better idea of timing. Connected, but idle, peers will time out their secure connection within 10 mins:

```
TRC 2021-04-23 11:25:44.640+02:00 Connection timed out                       topics="libp2p connection" tid=9928 file=connection.nim:111 s=16U*NBnuAz:60829040c2c3e939517d6f5d                                  
[...]

PR #523, opened 2021-04-29
Add optional keepAlive for relay peers

jm-clius:
This PR provides a workaround fix for #504 and addresses #487.

It adds an optional 5 min `keepAlive` for `WakuRelay` mesh peers. This is to work around the known `libp2p` behaviour where a `chronosstream` will time out after 10 minutes of inactivity.

The `keepAlive` itself sends a `GRAFT` message every 5 minutes to the mesh peers for an idle topic. Since the local node is in the remote peer's mesh already, the remote simply ignores the `GRAFT` but resets the activity timer thus keeping the connection alive. A longer-term solution would use a more appropriate keep-alive mechanism, e.g. [libp2p ping-protocol](https://docs.libp2p.io/concepts/protocols/#ping).

Since this functionality is only necessary where peers must maintain a stable connection while idle for long periods of time (e.g. the `test` cluster) it's disabled by default.

A possible future improvement is to only trigger the `keepAlive` for topics that are actually idle, i.e. tracking activity.

### DIFF
--- a/waku/v2/protocol/waku_relay.nim
+++ b/waku/v2/protocol/waku_relay.nim
@@ -4,6 +4,7 @@
 ## for spec.
 
 import
+  std/[tables, sequtils, sets],
   chronos, chronicles, metrics,
   libp2p/protocols/pubsub/[pubsub, gossipsub],
   libp2p/protocols/pubsub/rpc/messages,
@@ -12,10 +13,26 @@ import
 logScope:
     topics = "wakurelay"
 
-const WakuRelayCodec* = "/vac/waku/relay/2.0.0-beta2"
+const
+  WakuRelayCodec* = "/vac/waku/relay/2.0.0-beta2"
+  DefaultKeepAlive = 5.minutes # 50% of the default chronosstream timeout duration
 
 type
   WakuRelay* = ref object of GossipSub
+    keepAlive*: bool
+
+proc keepAlive*(w: WakuRelay) {.async.} =
+  while w.keepAlive:
+    # Keep all mesh peers alive when idle
+    trace "Running keepalive"
+
+    for topic in w.topics.keys:
+      trace "Keepalive on topic", topic=topic
+      let mpeers = w.mesh.getOrDefault(topic)
+
+      w.broadcast(toSeq(mpeers), RPCMsg(control: some(ControlMessage(graft: @[ControlGraft(topicID: topic)]))))
+    
+    await sleepAsync(DefaultKeepAlive)
 
 method init*(w: WakuRelay) =
   debug "init"
@@ -81,7 +98,14 @@ method unsubscribeAll*(w: WakuRelay,
 method start*(w: WakuRelay) {.async.} =
   debug "start"
   await procCall GossipSub(w).start()
+  
+  if w.keepAlive:
+    # Keep connection to mesh peers alive over periods of idleness
+    asyncSpawn keepAlive(w)
 
 method stop*(w: WakuRelay) {.async.} =
   debug "stop"
+
+  w.keepAlive = false
+
   await procCall GossipSub(w).stop()
--- a/waku/v2/node/config.nim
+++ b/waku/v2/node/config.nim
@@ -72,6 +72,11 @@ type
       desc: "Enable spam protection through rln-relay: true|false",
       defaultValue: false
       name: "rlnrelay" }: bool
+    
+    keepAlive* {.
+      desc: "Enable keep-alive for idle connections: true|false",
+      defaultValue: false
+      name: "keep-alive" }: bool
 
     swap* {.
       desc: "Enable swap protocol: true|false",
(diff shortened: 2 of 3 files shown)


waku/v2/protocol/waku_relay.nim
```diff
@@ -12,10 +13,26 @@ import
...
 type
   WakuRelay* = ref object of GossipSub
+    keepAlive*: bool
+
+proc keepAlive*(w: WakuRelay) {.async.} =
+  while w.keepAlive:
+    # Keep all mesh peers alive when idle
+    trace "Running keepalive"
+
+    for topic in w.topics.keys:
```
staheri14, line 29: Curious to know how the idle topics are identified in here?

staheri14 (approved):
Nice!

waku/v2/protocol/waku_relay.nim
```diff
@@ -12,10 +13,26 @@ import
...
+      trace "Keepalive on topic", topic=topic
+      let mpeers = w.mesh.getOrDefault(topic)
+
+      w.broadcast(toSeq(mpeers), RPCMsg(control: some(ControlMessage(graft: @[ControlGraft(topicID: topic)]))))
```
oskarth, line 33: Any side effects of sending graft to already connected peers that we should be careful with? Re peer scoring etc

oskarth (approved):
Thanks! Just a question on type of empty control message sent and possible side effects

waku/v2/protocol/waku_relay.nim
jm-clius: For now we keep all topics alive, whether idle or not. This is not ideal, so I might add some per-topic activity indicator. It could be as simple as an `activity` boolean that is set to `true` on any activity and resets to `false` whenever the `keepAlive` runs. The `keepAlive` will then only ping on topics with no activity since the last time the process kicked off.

waku/v2/protocol/waku_relay.nim
jm-clius: No side-effects that I could find in specs, the code or in the local tests I've run. The `GRAFT` is simply handled by the connected peer up to [this point](https://github.com/status-im/nim-libp2p/blob/master/libp2p/protocols/pubsub/gossipsub/behavior.nim#L142) with no penalty.
That said, I think `keep-alive` should only be enabled in rare situations (i.e. `test` cluster) and be deprecated once more permanent solutions are in place (e.g. peer discovery, `ping` protocol in `libp2p`, etc.). It's also a good idea to double-check with @sinkingsugar that a `graft` from a peer already in the mesh has no negative side-effects.

sinkingsugar: If you graft an explicit/direct peer it's a violation, why not just send a regular message instead? Or implement a ping protocol?

I would not mess with the gossip protocol itself personally. It's not the way it's meant to be used.

sinkingsugar:
This is not the way to use graft messages. Graft is a control message used by the protocol internally to communicate that in fact a peer was added to the mesh. Remote peers assume that state afterwards.
Moreover grafting sometimes is a violation.

I think a dedicated user app ping protocol should be implemented instead.

jm-clius:
> I think a dedicated user app ping protocol should be implemented instead.

Thanks, @sinkingsugar. Yeah, a ping protocol would be much better suited and less of a hack. Do you know if https://docs.libp2p.io/concepts/protocols/#ping is on the nim-libp2p roadmap? I'll add a feature request. In the meantime, of course, I'll look into creating a proper ping protocol on Waku level as well.

sinkingsugar:
> > I think a dedicated user app ping protocol should be implemented instead.
> 
> Thanks, @sinkingsugar. Yeah, a ping protocol would be much better suited and less of a hack. Do you know if https://docs.libp2p.io/concepts/protocols/#ping is on the nim-libp2p roadmap? I'll add a feature request. In the meantime, of course, I'll look into creating a proper ping protocol on Waku level as well.

Well, I don't know but if you implement it for Waku you might as well just implement the standard one in libp2p directly. Win-Win 😄

oskarth:
If we create a ping protocol it should indeed be the standard one in libp2p. I believe above was suggested as a short-term workaround for existing issues until it makes sense to do a proper ping protocol

jm-clius:
> > > I think a dedicated user app ping protocol should be implemented instead.
> > 
> > 
> > Thanks, @sinkingsugar. Yeah, a ping protocol would be much better suited and less of a hack. Do you know if https://docs.libp2p.io/concepts/protocols/#ping is on the nim-libp2p roadmap? I'll add a feature request. In the meantime, of course, I'll look into creating a proper ping protocol on Waku level as well.
> 
> Well, I don't know but if you implement it for Waku you might as well just implement the standard one in libp2p directly. Win-Win 😄 (cc @dryajov )

Indeed!

PR #525, opened 2021-04-30
Apply keep-alive for chat2

jm-clius:
Simple change to apply `keep-alive` config to `chat2` applications. This way `chat2` clients can be configured to remain connected if idle for a long time.


staheri14 (approved):
LGTM!

oskarth (approved):

PR #529, opened 2021-05-03
Fix: light-mode relay for all light protocols

### LINKED ISSUE #498: Issue: `WakuFilter` requires `WakuRelay`

jm-clius:
This PR closes #498. It follows the same approach as https://github.com/status-im/nim-waku/pull/518, but fixes the issue in a more general way to apply to any `light` protocol.

(Also sneaked in an unrelated CHANGELOG.md entry, which I forgot about earlier 😄.)

## Background recap

Currently `nim-libp2p` requires the same `PubSub` protocol used in full nodes (in our case, `RELAY`) to be mounted on light nodes for protocol negotiation to succeed. Detailed explanation [here](https://github.com/status-im/nim-waku/issues/498#issuecomment-828543874).
However, once a light node with `RELAY` mounted has connected to a full node (say, to `subscribe` a content filter or to perform a `lightpush`), it will be considered a "normal" PubSub peer in the full node, undergo grafting and pruning and generally behave like a full node.
This fix allows a light node to behave like a "true" light node by mounting a "light" mode of `RELAY` allowing light protocol negotation but no full-message peerings. `RELAY` is only mounted on the switch and never started. `node.wakuRelay` remains `nil` and the light node does not subscribe to any PubSub topics.

### DIFF
--- a/tests/v2/test_wakunode.nim
+++ b/tests/v2/test_wakunode.nim
@@ -211,7 +211,7 @@ procSuite "WakuNode":
     await node1.stop()
     await node2.stop()
   
-  asyncTest "Filter protocol works between light node and full node":
+  asyncTest "Filter protocol works on node without relay capability":
     let
       nodeKey1 = crypto.PrivateKey.random(Secp256k1, rng[])[]
       node1 = WakuNode.init(nodeKey1, ValidIpAddress.init("0.0.0.0"), Port(60000))
@@ -228,7 +228,7 @@ procSuite "WakuNode":
     node1.mountFilter()
 
     await node2.start()
-    node2.mountLightRelay() # Do not start WakuRelay or subscribe to any topics
+    node2.mountRelay(relayMessages=false) # Do not start WakuRelay or subscribe to any topics
     node2.mountFilter()
     node2.wakuFilter.setPeer(node1.peerInfo)
 
--- a/tests/v2/test_wakunode.nim
+++ b/tests/v2/test_wakunode.nim
@@ -210,6 +210,54 @@ procSuite "WakuNode":
 
     await node1.stop()
     await node2.stop()
+  
+  asyncTest "Filter protocol works between light node and full node":
+    let
+      nodeKey1 = crypto.PrivateKey.random(Secp256k1, rng[])[]
+      node1 = WakuNode.init(nodeKey1, ValidIpAddress.init("0.0.0.0"), Port(60000))
+      nodeKey2 = crypto.PrivateKey.random(Secp256k1, rng[])[]
+      node2 = WakuNode.init(nodeKey2, ValidIpAddress.init("0.0.0.0"), Port(60002))
+      defaultTopic = "/waku/2/default-waku/proto"
+      contentTopic = "defaultCT"
+      payload = @[byte 1]
+      message = WakuMessage(payload: payload, contentTopic: contentTopic)
+      filterRequest = FilterRequest(contentFilters: @[ContentFilter(contentTopics: @[contentTopic])], subscribe: true)
+
+    await node1.start()
+    node1.mountRelay()
+    node1.mountFilter()
+
+    await node2.start()
+    node2.mountLightRelay() # Do not start WakuRelay or subscribe to any topics
+    node2.mountFilter()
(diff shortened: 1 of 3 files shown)


staheri14 (approved):

tests/v2/test_wakunode.nim
```diff
@@ -210,6 +210,54 @@ procSuite "WakuNode":
 
     await node1.stop()
     await node2.stop()
+  
+  asyncTest "Filter protocol works between light node and full node":
```
oskarth, line 214: I'd be careful with using words such as "light" and "full", I know we often take this shortcut but being more precise here would be useful. I.e. we are really talking about adaptive nodes, that have a varying set of capabilities.

Maybe we can specify what we mean by light and full here?

Also see https://github.com/vacp2p/rfc/issues/87 for some previousy discussion

waku/v2/node/wakunode2.nim
```diff
@@ -359,6 +359,20 @@ proc info*(node: WakuNode): WakuInfo =
   let wakuInfo = WakuInfo(listenStr: listenStr)
   return wakuInfo
 
+proc mountLightRelay*(node: WakuNode) =
```
oskarth, line 362: Instead of using "light", what about something like SkeletonRelay? EmptyRelay? NonRelayingRelay? :D

Intuition: with topic sharding a node can subscribe to anything from 0 to 10 topics (say), what exactly is a light node in such a scenario? Add to that that some nodes might do storing, for a certain period of time, for some topics...etc. Desktop node is a good way to think about it, which would be neither "minimally light" nor "maximally full".

oskarth, line 362: Actually, can this just be mountRelay? Then have as an argument if we want to subscribe to default topic or not. Might be cleaner? So you have

mountRelay(sub=false)
mountRelay(sub=default)
mountRelay(sub=shard1, shard2, shard3...)
etc

oskarth (commented):
Some comments on getting naming and concepts right

tests/v2/test_wakunode.nim
jm-clius: Ah, thanks, that discussion is very enlightening. Conceptually much cleaner to think in terms of capabilities.

waku/v2/node/wakunode2.nim
jm-clius: Yes, this makes sense. One extra layer of complexity here is that we want to (are forced to) mount relay on the switch without actually having the _capability_ to relay messages, but this works well in this model as just another argument.

jm-clius:
Ok, in 71b4bfb I try avoiding overloaded concepts like "light" and "full" node. I now rather refer to the _capability_ to relay messages or not. Lmk if this clears things up. cc @oskarth

oskarth (approved):

PR #530, opened 2021-05-04
Storing waku message version inside the store db

staheri14:
Closes https://github.com/status-im/nim-waku/issues/482


oskarth (approved):

jm-clius (approved):
LGTM!

Just note that this breaks the schema again, but I'm assuming we can just follow a similar process as before when we added the `pubsubTopic` column? This time, however, the table name is not changing so clients will throw errors until they recreate their DBs.

richard-ramos (approved):
:100:

staheri14:
> LGTM!
> 
> Just note that this breaks the schema again, but I'm assuming we can just follow a similar process as before when we added the `pubsubTopic` column? This time, however, the table name is not changing so clients will throw errors until they recreate their DBs.

That is right! I am going to announce both in waku discord and walletConnect channel that the Message table must be deleted from the store.sqlite3 database (or the database file can be deleted instead).

PR #537, opened 2021-05-06
Add feature for running targeted tests

### LINKED ISSUE #535: Hassle In Running Targeted Tests
Its currently a hassle to run a specific test. This should be made as simple as running a make command.  A possible implementation should be to have a  generic command that can target a particular test on a given version.

Eg: `make runtest test=test_name v=2`

EbubeUd:
closes #535

Added a generic command to the makefile for running targeted tests. 
Tests can be run using `make runtest test=test_file_name v=2`

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "makefile.extensionOutputFolder": "./.vscode"
+}
--- a/Makefile
+++ b/Makefile
@@ -148,6 +148,16 @@ chat2bridge: | build deps
 # Builds and run the test suite (Waku v1 + v2)
 test: | test1 test2
 
+runtest: | build deps
+ifndef test
+	echo -e "test case was not set. Run make runtest version=<version_number> test=<target_test>"
+else ifndef v
+	echo -e "version number was not set. Run make runtest version=<version_number> test=<target_test>"
+else
+	echo -e $(BUILD_MSG) "build/$@" && \
+	$(ENV_SCRIPT) nim testScript $(NIM_PARAMS) waku.nims $(test) $(v)
+endif
+
 # symlink
 waku.nims:
 	ln -s waku.nimble $@
--- a/waku.nimble
+++ b/waku.nimble
@@ -25,20 +25,20 @@ requires "nim >= 1.2.0",
   "rln"
 
 ### Helper functions
-proc buildBinary(name: string, srcDir = "./", params = "", lang = "c") =
+proc buildBinary(name: string, srcDir = "./", params = "", lang = "c", paramOffset = 0) =
   if not dirExists "build":
     mkDir "build"
   # allow something like "nim nimbus --verbosity:0 --hints:off nimbus.nims"
   var extra_params = params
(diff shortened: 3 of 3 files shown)


staheri14 (approved):
Nice!

Are you going to include some instructions around `runtest` in here https://github.com/status-im/nim-waku/blob/master/README.md?

EbubeUd:
> Nice!
> 
> Are you going to include some instructions around `runtest` in here https://github.com/status-im/nim-waku/blob/master/README.md?

Sure I'll do that. Thanks for pointing out. :)

arnetheduck:
One easy way to run tests is `./env.sh nim c -r folder/testfile.nim` - if the module is using `unittest` you can also tack on the test name

oskarth (commented):
I don't quite understand the problem this is solving, what's the issue with the command @arnetheduck posted? We also have this as an example in the README https://github.com/status-im/nim-waku/blob/master/waku/v2/README.md#waku-v2-protocol-test-suite:

```
You can also run a specific test (and alter compile options as you want):

# Get a shell with the right environment variables set
./env.sh bash
# Run a specific test
nim c -r ./tests/v2/test_waku_filter.nim
```

EbubeUd:
> One easy way to run tests is `./env.sh nim c -r folder/testfile.nim` - if the module is using `unittest` you can also tack on the test name

Thanks a lot  for this. I tried running a test through the .nim file earlier but I think it needed some dependencies for it to run properly, so it got terminated. Then I checked the previous tests that were executed and saw that they were built first before execution. Probably, the essence of building first was to have all the dependencies Included. So I decided to follow a similar pattern. 
But I'm currently trying to see how the process can be simplified using the  command you sent.

jm-clius:
@EbubeUd my bad, I should have pointed you to the env script! See if this solves your problem(s).
Otherwise, is there anything else that your changes add other than being a shortcut? If not, perhaps we should close without merging. WDYT?

EbubeUd:
> I don't quite understand the problem this is solving, what's the issue with the command @arnetheduck posted? We also have this as an example in the README https://github.com/status-im/nim-waku/blob/master/waku/v2/README.md#waku-v2-protocol-test-suite:
> 
> ```
> You can also run a specific test (and alter compile options as you want):
> 
> # Get a shell with the right environment variables set
> ./env.sh bash
> # Run a specific test
> nim c -r ./tests/v2/test_waku_filter.nim
> ```
> 
> The pattern of "v=2" also doesn't seem very general. All we are doing here is running a test which is in some `foo/bar/path`, there's nothing magic about a version in path here. So if we want to do this the argument should probably be full path relative to tests folder.


> @EbubeUd my bad, I should have pointed you to the env script! See if this solves your problem(s).
> Otherwise, is there anything else that your changes add other than being a shortcut? If not, perhaps we should close without merging. WDYT?




Yea I agree, Its more of a shortcut. I believe this can be closed without merging. It seems the problem I have with running the env script currently is specific to my environment/Setup so I'll look into it properly.
Thanks for the assistance.

PR #539, opened 2021-05-07
Fault tolerant store protocol: resume message history

staheri14:
This is to close https://github.com/status-im/nim-waku/issues/512

- In this basic implementation,  when a restart occurs, the wakunode2 which has the store protocol mounted attempts to fetch the history of message since his last on-time. 
- The last on-time is considered to be the timestamp of the most recent message in the nodes message db.
- A historical time query for that time period is made and the fetched messages get stored in the db as well as in the memory. 
- An offset of 20 second is added to the time window due to the assumption that nodes may be out of sync by at most 20 seconds.

The applied changes are not yet integrated in wakunode2 and  just a placeholder to run the `resume` proc is included in the code (waiting for hackathon to finish to make the changes effective). 

## Further improvements
- Currently, we rely on the peer manager component to pick one of the already existing store enabled connections to query from. A better solution would be to identify peers which were online for that targeted time period and pass a list of such candidates  to the resume procedure to query from. Will be working on it in the next PR.
- Another aspect is to mention is that an offline node can only query for the messages that are published on the waku default pubsub topic, more specific, can only query for a known pubsub topic. Ideally one should be all the pubsub topics (or should not?) i.e., regardless of pubsub topic. However, this is not doable in the current design as the messages embedded in a history response have no field to indicate their pubsub topic  therefore the querying node has no way to associate the retrieved messaged with their pubsub topic.

### DIFF
--- a/tests/v2/test_waku_store.nim
+++ b/tests/v2/test_waku_store.nim
@@ -663,18 +663,18 @@ procSuite "Waku Store":
       check:
         (await completionFut.withTimeout(5.seconds)) == true
 
-    asyncTest "fin last seen ":
+    test "find last seen ":
       var
-        msgList = @[WakuMessage(payload: @[byte 0], contentTopic: ContentTopic("2")),
-          WakuMessage(payload: @[byte 1],contentTopic: ContentTopic("1"), timestamp: float(1)),
-          WakuMessage(payload: @[byte 2],contentTopic: ContentTopic("2"), timestamp: float(2)),
-          WakuMessage(payload: @[byte 3],contentTopic: ContentTopic("1"), timestamp: float(3)),
-          WakuMessage(payload: @[byte 4],contentTopic: ContentTopic("2"), timestamp: float(4)),
-          WakuMessage(payload: @[byte 5],contentTopic: ContentTopic("1"), timestamp: float(5)),
-          WakuMessage(payload: @[byte 6],contentTopic: ContentTopic("2"), timestamp: float(6)),
-          WakuMessage(payload: @[byte 7],contentTopic: ContentTopic("1"), timestamp: float(7)),
-          WakuMessage(payload: @[byte 8],contentTopic: ContentTopic("2"), timestamp: float(8)),
-          WakuMessage(payload: @[byte 9],contentTopic: ContentTopic("1"),timestamp: float(9))]      var completionFut = newFuture[bool]()
+        msgList = @[IndexedWakuMessage(msg: WakuMessage(payload: @[byte 0], contentTopic: ContentTopic("2"))),
+          IndexedWakuMessage(msg: WakuMessage(payload: @[byte 1],contentTopic: ContentTopic("1"), timestamp: float(1))),
+          IndexedWakuMessage(msg: WakuMessage(payload: @[byte 2],contentTopic: ContentTopic("2"), timestamp: float(2))),
+          IndexedWakuMessage(msg: WakuMessage(payload: @[byte 3],contentTopic: ContentTopic("1"), timestamp: float(3))),
+          IndexedWakuMessage(msg: WakuMessage(payload: @[byte 4],contentTopic: ContentTopic("2"), timestamp: float(4))),
+          IndexedWakuMessage(msg: WakuMessage(payload: @[byte 5],contentTopic: ContentTopic("1"), timestamp: float(9))),
+          IndexedWakuMessage(msg: WakuMessage(payload: @[byte 6],contentTopic: ContentTopic("2"), timestamp: float(6))),
+          IndexedWakuMessage(msg: WakuMessage(payload: @[byte 7],contentTopic: ContentTopic("1"), timestamp: float(7))),
+          IndexedWakuMessage(msg: WakuMessage(payload: @[byte 8],contentTopic: ContentTopic("2"), timestamp: float(8))),
+          IndexedWakuMessage(msg: WakuMessage(payload: @[byte 9],contentTopic: ContentTopic("1"),timestamp: float(5)))]     
 
       check:
-        findLastSeen(msgList) = float(9)
+        findLastSeen(msgList) == float(9)
--- a/tests/v2/test_waku_store.nim
+++ b/tests/v2/test_waku_store.nim
@@ -581,6 +581,7 @@ procSuite "Waku Store":
     var dialSwitch = newStandardSwitch()
     discard await dialSwitch.start()
 
+    # to be connected to
     var listenSwitch = newStandardSwitch(some(key))
     discard await listenSwitch.start()
 
@@ -647,24 +648,7 @@ procSuite "Waku Store":
       check:
         (await completionFut.withTimeout(5.seconds)) == true
 
-  
-    asyncTest "handle temporal history query with a zero-size time window":
-      # a zero-size window results in an empty list of history messages
-      var completionFut = newFuture[bool]()
-
-      proc handler(response: HistoryResponse) {.gcsafe, closure.} =
-        check:
-          # a zero-size window results in an empty list of history messages
-          response.messages.len() == 0
-        completionFut.complete(true)
-
-      let rpc = HistoryQuery(contentFilters: @[HistoryContentFilter(contentTopic: ContentTopic("1"))], startTime: float(2), endTime: float(2))
-      await proto.query(rpc, handler)
-
-      check:
-        (await completionFut.withTimeout(5.seconds)) == true
-
-    test "find last seen ":
+    test "find last seen message":
       var
         msgList = @[IndexedWakuMessage(msg: WakuMessage(payload: @[byte 0], contentTopic: ContentTopic("2"))),
           IndexedWakuMessage(msg: WakuMessage(payload: @[byte 1],contentTopic: ContentTopic("1"), timestamp: float(1))),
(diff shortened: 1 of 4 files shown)


waku/v2/protocol/waku_store/waku_store.nim
```diff
@@ -457,9 +457,42 @@ proc query*(w: WakuStore, query: HistoryQuery, handler: QueryHandlerFunc) {.asyn
...
 
   waku_store_messages.set(response.value.response.messages.len.int64, labelValues = ["retrieved"])
-
   handler(response.value.response)
 
+  
+proc findLastSeen*(list: seq[IndexedWakuMessage]): float = 
+  var lastSeenTime = float64(0)
+  for iwmsg in list.items : 
+    lastSeenTime = if iwmsg.msg.timestamp>lastSeenTime: iwmsg.msg.timestamp else: lastSeenTime
```
jm-clius, line 466: Not sure, but any way we could reconsider how we order messages in `ws.messages`?
We already compute an [epoch-based index for each message](https://github.com/status-im/nim-waku/blob/master/waku/v2/protocol/waku_store/waku_store.nim#L40). Elsewhere we [sort the messages](https://github.com/status-im/nim-waku/blob/master/waku/v2/protocol/waku_store/waku_store.nim#L256) in response to a `HistoryQuery` on index (which will have the effect of sorting them by time, due to the epoch indices). Here we cycle through all messages to find the youngest timestamp. All of these can be computationally quite expensive. Wouldn't it be easier to just maintain a time-based sorting order on `ws.messages`? In that case `findLastSeen` would just be the timestamp of the last message in the sequence.

tests/v2/test_waku_store.nim
```diff
@@ -595,7 +596,8 @@ procSuite "Waku Store":
     listenSwitch.mount(proto)
 
     for wakuMsg in msgList:
-      await subscriptions.notify("foo", wakuMsg)
+      # the pubsub topic should be DefaultTopic
```
oskarth, line 599: leftover comment?

waku/v2/protocol/waku_store/waku_store.nim
```diff
@@ -457,9 +457,42 @@ proc query*(w: WakuStore, query: HistoryQuery, handler: QueryHandlerFunc) {.asyn
...
   handler(response.value.response)
 
+  
+proc findLastSeen*(list: seq[IndexedWakuMessage]): float = 
+  var lastSeenTime = float64(0)
+  for iwmsg in list.items : 
+    lastSeenTime = if iwmsg.msg.timestamp>lastSeenTime: iwmsg.msg.timestamp else: lastSeenTime 
+  return lastSeenTime
+
+proc resume*(ws: WakuStore){.async, gcsafe.} =
```
oskarth, line 469:
```suggestion
proc resume*(ws: WakuStore) {.async, gcsafe.} =
```

oskarth: Can we add a more detailed docstring to this? Right now it isn't obvious what one would expect this call to do, or what has to be true for it to work properly, etc.

waku/v2/protocol/waku_store/waku_store.nim
oskarth, line 466: I'd also split this if-else on separate lines for clarity, right now it is a bit hard to read

waku/v2/protocol/waku_store/waku_store.nim
```diff
@@ -457,9 +457,42 @@ proc query*(w: WakuStore, query: HistoryQuery, handler: QueryHandlerFunc) {.asyn
...
+  ## fetch the message history of the DefaultTopic since the last seen message in the db
+  debug "resume"
+  var currentTime = epochTime()
```
oskarth, line 472: Assuming clock is accurate here, so maybe add that to log as well and make a note?

People are definitely going to be bitten by this, so at some point in future we might want to consider requiring NTP server or so to be running, or print some form of warning when starting up a node about clock

tests/v2/test_waku_store.nim
staheri14: It is indeed intentional, the node should query for messages published on DefaultTopic, and it should not change in the future updates of this unit test.

waku/v2/protocol/waku_store/waku_store.nim
staheri14: Clocks can be out of sync for at most 20 sec. 
In more details, for every pair of nodes in the system lets say P and Pi, the following relation holds between their clocks i.e., `ti` and `t` 
`ti = t + ci + fi(t)`

`ci` is  a constant value and represents the skew
`fi(t)` is a function of `t` cnd signifies the clock drift between P and Pi

We assume that `-10s<ci<+10s` and `-10s<fi(t)<+10s`
Considering these, `ti` and `t` can differ by at most `+-20s`

Following the preceding assumption, the offline time window is adjusted by a 20 sec offset (lines 475-478 copied below). Let me know whether this makes sense.  
```
# adjust the time window with an offset of 20 seconds
  let offset: float64 = 200000
  currentTime = currentTime + offset
  lastSeenTime = max(lastSeenTime - offset, 0)
```

waku/v2/protocol/waku_store/waku_store.nim
staheri14: For sure!

waku/v2/protocol/waku_store/waku_store.nim
staheri14: sure, done!

waku/v2/protocol/waku_store/waku_store.nim
staheri14: Thanks for the comment!

`findLastSeen` does the search based on sender generated timestamp and this is only needed once (at least for now it is a one time invocation), thus a linear search would be cheaper than sorting all the messages based on their sender generated `timestamp` (messages Index is based on received generated timestamp)

However, we indeed sort messages based on their Index (which encapsulated receiver generated timestamp but not sender generated) redundantly for every incoming history query. We can certainly optimize on that part and sort them once when messages are loaded from db. Will do it in the subsequent PRs.

waku/v2/protocol/waku_store/waku_store.nim
oskarth: To clarify, I understand why it is done, how it is done and that it is done. What I suggest is:

- Add a comment in the code describing this assumption / link to spec portion describing it
- Log this as part of debug statement with time so people can debug it with faulty clocks
- Note for future that we'll have to figure out a way to make this user friendly
--  potentially add to README? way to detect/sync NTP as part of setup script? etc (out of scope of PR but something to be mindful of)

oskarth: I see that a comment was added, thanks!

waku/v2/protocol/waku_store/waku_store.nim
```diff
@@ -457,9 +457,54 @@ proc query*(w: WakuStore, query: HistoryQuery, handler: QueryHandlerFunc) {.asyn
...
+  ## all the messages within that window is fetched and persisted in the store node's messages field and in message store db
+  ## a history query for the offline time window is made and sent over to one of the persisted store enabled peer 
+  ## the peer selection for the query is implicit and is handled as part of the query procedure through peer manager
+  ## the assumption is that the queried node has been online for that time window and hence the history will be fetched successfully
+  ## no query failure is supported for now
+  ## the history will be fetched successfully assuming that the store node 
+  ## TODO we need to develop a peer discovery method to obtain list of nodes that have been online for a specific time window
+  ## TODO such list then can be passed to the resume proc 
+  debug "resume"
+  var currentTime = epochTime()
```
oskarth, line 484:
```suggestion
  var currentTime = epochTime()
  debug "resume", currentEpochTime=currentTime
```

Something like this, because then a bug report with debug logs in the future can be used to debug potentially out of sync clocks (untested)

waku/v2/protocol/waku_store/waku_store.nim
oskarth: For the issue, I'd probably create a small issue in nim-waku to figure out better way to detect/deal with out of sync real time clocks. Then we can figure out when it makes sense to attack it (not prio now, more relevant when people start using this on wonky desktop/server setups without things like NTP).

oskarth (approved):
See comments above otherwise LGTM

waku/v2/protocol/waku_store/waku_store.nim
jm-clius: Ah, thanks for clarifying.

jm-clius (approved):
LGTM!

waku/v2/protocol/waku_store/waku_store.nim
staheri14: You are right and the ultimate goal is to use sender generated timestamp, however, that requires a separate PR as it would also affect the pagination logic. It is on the plan though!

waku/v2/protocol/waku_store/waku_store.nim
staheri14: Thanks!

waku/v2/protocol/waku_store/waku_store.nim
staheri14: Thanks for the clarification!
Created the issue https://github.com/status-im/nim-waku/issues/547

Issue #540, opened 2021-05-10
Waku config inconsistencies


jm-clius:
As Waku v2 has grown, some inconsistencies with config naming and conventions slipped in.
Notably:
- message persistence is enabled using `--persist-messages`, but peer persistence uses `--peerpersist` (former seems better convention).
- the _staticnodes_ (plural) config property, is singular in the parameter name: `--staticnode`

There are three V2 config files:
1. https://github.com/status-im/nim-waku/blob/master/waku/v2/node/config.nim
2. https://github.com/status-im/nim-waku/blob/master/waku/common/config_bridge.nim
3. https://github.com/status-im/nim-waku/blob/master/examples/v2/matterbridge/config_chat2bridge.nim

PR #543, opened 2021-05-10
Fix nim-waku config inconsistencies

jm-clius:
## Background

This PR partially addresses #540. It fixes some inconsistencies in the naming of Waku v2 config items. It is the first of three steps necessary to change CLI parameters without breaking existing deployments.

### Step 1 (this PR):

Add new, consistent config items in such a way that we still support the current items. In other words, keep the old CLI parameters intact. Refactor existing field names by postfixing them with `_depr`. This will mark them for removal in `Step 3`.

### Step 2:

Propagate changes to existing deployments:
- [ ] open PR adding support for the new config items on `prod` and `test` fleets
- [ ] inform external `nim-waku` clients, such as WalletConnect

### Step 3:

Remove deprecated config items from the codebase.

## Conventions followed

I've tried adhering to the following conventions:
1. config items with similar semantics should follow similar naming conventions (e.g. `persist-peers` should look similar to `persist-messages`)
2. where a config item is composed of more than one word, use camelCase for field names while hyphenating the CLI parameter
3. parameters that take a "plural" argument should be in plural form (e.g. prefer `staticnodes` to `staticnode`)
4. the field name and CLI parameter for each item should agree
5. use consistent namespacing where possible (e.g. change `log-metrics` to `metrics-logging` to be consistent with other `metrics` config items)

## Summary of changes
The following CLI parameters have changed. The old parameters will be deprecated and removed in `Step 3`:

### For `wakunode2`, `chat2` and `chat2bridge`
- ~`nodekey` changed to `node-key`~ UPDATE: this would make it inconsistent with Waku v1. Reverted in 455edea
- `dbpath` changed to `db-path`
- `peerpersist` changed to `persist-peers`
- `rlnrelay` changed to `rln-relay`
- ~`staticnode` changed to `staticnodes`~ UPDATE: reverted in f4dd214. Reasoning [here](https://github.com/status-im/nim-waku/pull/543#discussion_r630059461)
- `log-metrics` changed to `metrics-logging`

### For `wakubridge`
All items overlapping with the above list, with the following additions
- `fleetv1` changed to `fleet-v1`
- `staticnodev1` changed to `staticnode-v1`
- `nodekeyv1` changed to `node-key-v1`
- `staticnodev2` changed to `staticnode-v2`
- `nodekeyv2` changed to `node-key-v2`

## What has not changed?
The concepts `staticnodes`, `filternode` and `storenode` are such tight concepts that I did not consider them as composites of two words. In other words, the config parameters were **not** changed to `filter-node`, `store-node` or `static-nodes`.

### DIFF
--- a/examples/v2/matterbridge/config_chat2bridge.nim
+++ b/examples/v2/matterbridge/config_chat2bridge.nim
@@ -66,14 +66,25 @@ type
       name: "metrics-server-port" .}: uint16
 
     ### Waku v2 options
-    staticnodes* {.
+    # @TODO: deprecate this item. Name changed from `staticnode` -> `staticnodes`
+    staticnodes_depr* {.
       desc: "Multiaddr of peer to directly connect with. Argument may be repeated"
       name: "staticnode" }: seq[string]
+    
+    staticnodes* {.
+      desc: "Multiaddr of peer to directly connect with. Argument may be repeated"
+      name: "staticnodes" }: seq[string]
 
-    nodekey* {.
+    # @TODO: deprecate this item. Name changed from `nodekey` -> `node-key`
+    nodekey_depr* {.
       desc: "P2P node private key as hex"
       defaultValue: crypto.PrivateKey.random(Secp256k1, newRng()[]).tryGet()
       name: "nodekey" }: crypto.PrivateKey
+    
+    nodeKey* {.
+      desc: "P2P node private key as hex"
+      defaultValue: crypto.PrivateKey.random(Secp256k1, newRng()[]).tryGet()
+      name: "node-key" }: crypto.PrivateKey
 
     topics* {.
       desc: "Default topics to subscribe to (space separated list)"
--- a/examples/v2/matterbridge/config_chat2bridge.nim
+++ b/examples/v2/matterbridge/config_chat2bridge.nim
@@ -66,14 +66,10 @@ type
       name: "metrics-server-port" .}: uint16
 
     ### Waku v2 options
-    # @TODO: deprecate this item. Name changed from `staticnode` -> `staticnodes`
-    staticnodes_depr* {.
-      desc: "Multiaddr of peer to directly connect with. Argument may be repeated"
-      name: "staticnode" }: seq[string]
     
     staticnodes* {.
       desc: "Multiaddr of peer to directly connect with. Argument may be repeated"
-      name: "staticnodes" }: seq[string]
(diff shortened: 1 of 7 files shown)


examples/v2/matterbridge/config_chat2bridge.nim
```diff
@@ -66,14 +66,25 @@ type
...
     ### Waku v2 options
-    staticnodes* {.
+    # @TODO: deprecate this item. Name changed from `staticnode` -> `staticnodes`
+    staticnodes_depr* {.
       desc: "Multiaddr of peer to directly connect with. Argument may be repeated"
       name: "staticnode" }: seq[string]
+    
+    staticnodes* {.
+      desc: "Multiaddr of peer to directly connect with. Argument may be repeated"
+      name: "staticnodes" }: seq[string]
```
kdeme, line 76: I use this singular typically because of the way it works in confutils. At CLI it expects the usage as such `--staticnode:... --staticnode:...` for multiple staticnodes. 
The actual variable can be plural as it holds the full provided list in a `seq`. 
So to me it makes more sense like this from a UX PoV. It would be different when the CLI would expect a comma separated list. I've seen some users making that error in the past actually, trying to provide the comma separated list when it was called `--staticnodes`

waku/v2/node/config.nim
```diff
@@ -38,81 +51,112 @@ type
...
+      name: "rln-relay" }: bool
+    
+    # @TODO: deprecate this item. Name changed from `staticnode` -> `staticnodes`
+    staticnodes_depr* {. 
+      desc: "Peer multiaddr to directly connect with. Argument may be repeated."
+      name: "staticnode" }: seq[string]
+    
+    staticnodes* {.
+      desc: "Peer multiaddr to directly connect with. Argument may be repeated."
+      name: "staticnodes" }: seq[string]
```
kdeme, line 109: idem

examples/v2/matterbridge/config_chat2bridge.nim
jakubgs: This is a good point. As long as `--help` clearly indicates that it can be used multiple times.

It would also match how flags like these are named in Nimbus:
https://github.com/status-im/nimbus-eth2/blob/e1a8049e/beacon_chain/conf.nim#L109-L111
https://github.com/status-im/nimbus-eth2/blob/e1a8049e/beacon_chain/conf.nim#L314-L316

PR #548, opened 2021-05-14
Ft-store: Peer management

staheri14:
This PR closes https://github.com/status-im/nim-waku/issues/513 and closes https://github.com/status-im/nim-waku/issues/458

A list of candidate peers is passed to the resume proc from which the history of messages is going to be retrieved. Such candidates should be found through a discovery method (to be developed in the future).  Passing this list to the resume proc is optional, and if left empty then the underlying peer manager unit of the store protocol will pick one of the current connections randomly to fetch the history from.

### DIFF
--- a/waku/v2/protocol/waku_store/waku_store.nim
+++ b/waku/v2/protocol/waku_store/waku_store.nim
@@ -459,7 +459,43 @@ proc query*(w: WakuStore, query: HistoryQuery, handler: QueryHandlerFunc) {.asyn
   waku_store_messages.set(response.value.response.messages.len.int64, labelValues = ["retrieved"])
   handler(response.value.response)
 
+proc queryFrom*(w: WakuStore, query: HistoryQuery, handler: QueryHandlerFunc, peer: PeerInfo): Future[bool] {.async.} =
+  # sends the query to the given peer
+  # returns true if no error occurs, otherwise false
+  let connOpt = await w.peerManager.dialPeer(peer, WakuStoreCodec)
+  echo "here"
+
+  if connOpt.isNone():
+    # TODO more sophisticated error handling here
+    error "failed to connect to remote peer"
+    waku_store_errors.inc(labelValues = [dialFailure])
+    return false
+
+  await connOpt.get().writeLP(HistoryRPC(requestId: generateRequestId(w.rng),
+      query: query).encode().buffer)
+
+  var message = await connOpt.get().readLp(64*1024)
+  let response = HistoryRPC.init(message)
+
+
+  if response.isErr:
+    error "failed to decode response"
+    waku_store_errors.inc(labelValues = [decodeRpcFailure])
+    return false
+  waku_store_messages.set(response.value.response.messages.len.int64, labelValues = ["retrieved"])
+  handler(response.value.response)
+  return true
   
+
+proc queryLoop*(w: WakuStore, query: HistoryQuery, handler: QueryHandlerFunc, candidateList: seq[PeerInfo]): Future[bool]  {.async.}= 
+  ## loops through the candidateList sequentially and sends the query to them until one of them gets through without error. 
+  ## returns false if all the requests fail
+  for peer in candidateList.items: 
+    let success = await w.queryFrom(query, handler, peer)
+    if success:
+      return true
+  debug "no peer was available"
+  return false
 proc findLastSeen*(list: seq[IndexedWakuMessage]): float = 
   var lastSeenTime = float64(0)
   for iwmsg in list.items : 
@@ -467,7 +503,7 @@ proc findLastSeen*(list: seq[IndexedWakuMessage]): float =
       lastSeenTime = iwmsg.msg.timestamp 
   return lastSeenTime
 
-proc resume*(ws: WakuStore) {.async, gcsafe.} =
+proc resume*(ws: WakuStore, peerList: Option[seq[PeerInfo]]) {.async.} =
   ## resume proc retrieves the history of waku messages published on the default waku pubsub topic since the last time the waku store node has been online 
   ## messages are stored in the store node's messages field and in the message db
   ## the offline time window is measured as the difference between the current time and the timestamp of the most recent persisted waku message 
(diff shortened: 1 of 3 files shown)


waku/v2/protocol/waku_store/waku_store.nim
```diff
@@ -459,24 +459,58 @@ proc query*(w: WakuStore, query: HistoryQuery, handler: QueryHandlerFunc) {.asyn
   waku_store_messages.set(response.value.response.messages.len.int64, labelValues = ["retrieved"])
   handler(response.value.response)
 
+proc queryFrom*(w: WakuStore, query: HistoryQuery, handler: QueryHandlerFunc, peer: PeerInfo): Future[bool] {.async.} =
```
oskarth, line 462: If queryFrom and resume are public functions, should they be exposed in node API at some point?

waku/v2/protocol/waku_store/waku_store.nim
```diff
@@ -459,24 +459,58 @@ proc query*(w: WakuStore, query: HistoryQuery, handler: QueryHandlerFunc) {.asyn
...
+  # sends the query to the given peer
+  # returns true if no error occurs, otherwise false
+  let connOpt = await w.peerManager.dialPeer(peer, WakuStoreCodec)
+  echo "here"
```
oskarth, line 466: Use log and more descriptive statement here, or remove

waku/v2/protocol/waku_store/waku_store.nim
```diff
@@ -459,24 +459,58 @@ proc query*(w: WakuStore, query: HistoryQuery, handler: QueryHandlerFunc) {.asyn
...
+    waku_store_errors.inc(labelValues = [dialFailure])
+    return false
+
+  await connOpt.get().writeLP(HistoryRPC(requestId: generateRequestId(w.rng),
+      query: query).encode().buffer)
+
+  var message = await connOpt.get().readLp(64*1024)
+  let response = HistoryRPC.init(message)
+
+  if response.isErr:
```
oskarth, line 479: I wonder if we want to make this into a Result type, so we can propagate error to caller of function? Instead of bool.

waku/v2/protocol/waku_store/waku_store.nim
```diff
@@ -459,24 +459,58 @@ proc query*(w: WakuStore, query: HistoryQuery, handler: QueryHandlerFunc) {.asyn
...
   
+
+proc queryLoop(w: WakuStore, query: HistoryQuery, handler: QueryHandlerFunc, candidateList: seq[PeerInfo]): Future[bool]  {.async.}= 
+  ## loops through the candidateList in order and sends the query to each until one of the query gets resolved successfully without error
+  ## returns false if all the requests fail
+  for peer in candidateList.items: 
+    let success = await w.queryFrom(query, handler, peer)
+    if success: return true
+
+  debug "failed to resolve the query"
```
oskarth, line 496: Same here re Results type

waku/v2/protocol/waku_store/waku_store.nim
```diff
@@ -459,24 +459,58 @@ proc query*(w: WakuStore, query: HistoryQuery, handler: QueryHandlerFunc) {.asyn
...
+
 proc findLastSeen*(list: seq[IndexedWakuMessage]): float = 
   var lastSeenTime = float64(0)
   for iwmsg in list.items : 
     if iwmsg.msg.timestamp>lastSeenTime: 
       lastSeenTime = iwmsg.msg.timestamp 
   return lastSeenTime
 
-proc resume*(ws: WakuStore) {.async, gcsafe.} =
+proc resume*(ws: WakuStore, peerList: Option[seq[PeerInfo]]) {.async.} =
```
oskarth, line 506: Return type maybe? success or not, maybe with number of messages stored as some form of feedback?

jm-clius (approved):
LGTM as a general idea. I imagine this will eventually be integrated with the existing peer manager? Discovered `store` peers can be added there and candidates for a query retrievedas a list.

staheri14:
> LGTM as a general idea. I imagine this will eventually be integrated with the existing peer manager? Discovered `store` peers can be added there and candidates for a query retrievedas a list.
@jm-clius 
It depends! the candidate peers are those that have been available for a particular time window, and that window is going to be queried once. This means those candidates would be of no use for the future. As such, I think we may not want to persist them in the peer manager part. On the other hand, in case that we have strongly available candidates that are always online (not just for a specific time span), then keeping them in the peer manager part makes a lot of sense.

waku/v2/protocol/waku_store/waku_store.nim
staheri14: Sorry, that was not supposed to be here, forgot to remove it.

waku/v2/protocol/waku_store/waku_store.nim
staheri14: Done!

waku/v2/protocol/waku_store/waku_store.nim
staheri14: Done!

waku/v2/protocol/waku_store/waku_store.nim
staheri14: Can you elaborate on exposing this proc in the node API? Which api?

waku/v2/protocol/waku_store/waku_store.nim
staheri14: Done!

waku/v2/protocol/waku_store/waku_store.nim
oskarth: `wakunode` api https://github.com/status-im/nim-waku/blob/master/docs/api/v2/node.md

waku/v2/protocol/waku_store/waku_store.nim
```diff
@@ -459,28 +459,65 @@ proc query*(w: WakuStore, query: HistoryQuery, handler: QueryHandlerFunc) {.asyn
   waku_store_messages.set(response.value.response.messages.len.int64, labelValues = ["retrieved"])
   handler(response.value.response)
 
+proc queryFrom*(w: WakuStore, query: HistoryQuery, handler: QueryHandlerFunc, peer: PeerInfo): Future[Result[int64, string]] {.async.} =
```
oskarth, line 462: I'd consider pulling out the Result[int64, string] as a separate type, such as `QueryResult` or similar. Minor detail though!

oskarth (approved):

jm-clius:
> It depends! the candidate peers are those that have been available for a particular time window, and that window is going to be queried once.

Ah, yeah, good point!

waku/v2/protocol/waku_store/waku_store.nim
staheri14: Sure, addressed!

waku/v2/protocol/waku_store/waku_store.nim
staheri14: Thanks for the link, I also agree that both `queryFrom` and `resume` must be exposed in node api. Will do it in subsequent PRs.

staheri14: Corresponding issue https://github.com/status-im/nim-waku/issues/557

PR #556, opened 2021-05-19
Duplicate filtering for `bridge`

### LINKED ISSUE #554: Duplicate filtering for Waku  v1 <> v2 bridge
## Problem
During [`bridge` testing](https://github.com/status-im/nim-waku/issues/553) an issue was identified where bridging an envelope from Waku v1 to Waku v2 will trigger bridging that same message back to v1. This is because the default handler for the v2 topic will be triggered after receiving the bridged message, which in turn triggers the v2 -> v1 bridge.
Duplicate messages only occur in the direction v1 -> v2.

## Acceptance criteria
- [x] A message/envelope should be bridged only once. No duplicate messages in either direction.

jm-clius:
This PR closes #554

It adds duplicate filtering to the Waku v1 <> v2 bridge.

Duplicate messages occurred because the default relay handler, registered on the bridge for the default bridge topic, will be triggered after publishing the message bridged from v1. This in turn triggers the v2 -> v1 bridge with a duplicate.
This only happens in direction v1 -> v2.

PR includes general improvements to unit test.

### DIFF
--- a/waku/common/wakubridge.nim
+++ b/waku/common/wakubridge.nim
@@ -1,5 +1,5 @@
 import
-  std/tables,
+  std/[tables, hashes, sequtils],
   chronos, confutils, chronicles, chronicles/topics_registry, metrics,
   stew/[byteutils, objects],
   stew/shims/net as stewNet, json_rpc/rpcserver,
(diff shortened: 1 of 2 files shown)


waku/common/wakubridge.nim
```diff
@@ -36,29 +38,60 @@ type
...
 
 ###################
 # Helper funtions #
 ###################
 
+proc containsOrAdd(sequence: var seq[hashes.Hash], hash: hashes.Hash): bool =
+  if sequence.contains(hash):
+    return true 
+
+  if sequence.len >= DeduplQSize:
```
staheri14, line 51: Why only 20 messages in the queue?

staheri14 (approved):
LGTM!

oskarth (approved):

PR #563, opened 2021-05-21
Minor bridge improvements

jm-clius:
Minor improvements to `bridge` and `chat2bridge`:

1. Sets the `relay` property `triggerSelf` to `false`, to avoid the bridge triggering a callback for a message it has just bridged to Waku v2. Also see https://github.com/status-im/nim-waku/issues/562.
2. Lower logs to more realistic levels

### DIFF
--- a/examples/v2/matterbridge/chat2bridge.nim
+++ b/examples/v2/matterbridge/chat2bridge.nim
@@ -172,7 +172,9 @@ proc start*(cmb: Chat2MatterBridge) {.async.} =
   debug "Start listening on Waku v2"
   await cmb.nodev2.start()
   
-  cmb.nodev2.mountRelay() # Always mount relay for bridge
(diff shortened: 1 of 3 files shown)


jm-clius:
@oskarth, my assumption here is that `triggerSelf` on a `wakunode2` was originally set to `true` by default to allow the `subscription` model? In other words, it enables triggering `pubsubTopic` handlers even for messages that originated from that node. If a `bridge` node is not configured as a `store` or `filter` node (which I don't think it should be), it should have `triggerSelf` set to `false`.

staheri14 (approved):

EbubeUd (approved):

oskarth (approved):
I think triggerSelf preceded the subscription model, but I can't figure out what the initial rationale was. It is probably connected, and if this seems to work let's go for it :)

PR #564, opened 2021-05-21
Add resume to waku node api

staheri14:
This is to close https://github.com/status-im/nim-waku/issues/557
- The `resume` method is introduced to the `wakunode2` API to capture the node's offline period and fetch historical messages for that time.
- The `resume` proc is executed on `wakunode2` after each restart (given that the the store protocol is enabled in the full mode). 
- Also included some more logs  for the sake of debugging.

### DIFF
--- a/docs/api/v2/node.md
+++ b/docs/api/v2/node.md
@@ -13,6 +13,7 @@ the consumer wants to make. These methods are:
 5. **Publish** - to a topic, or a topic and a specific content filter.
 6. **Query** - for historical messages.
 7. **Info** - to get information about the node.
+8. **Resume** - to retrieve and persist the message history since the node's last online time.
 
 ```Nim
 proc init*(T: type WakuNode, nodeKey: crypto.PrivateKey,
--- a/docs/api/v2/node.md
+++ b/docs/api/v2/node.md
@@ -74,6 +74,16 @@ proc info*(node: WakuNode): WakuInfo =
   ##
   ## Status: Implemented.
   ##
+
+proc resume*(node: WakuNode, peerList: Option[seq[PeerInfo]]): WakuInfo =
+  ## Retrieves the history of waku messages published on the default waku pubsub topic since the last time the waku node has been online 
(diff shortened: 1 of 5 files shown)


docs/api/v2/node.md
```diff
@@ -74,6 +74,16 @@ proc info*(node: WakuNode): WakuInfo =
   ##
   ## Status: Implemented.
   ##
+
+proc resume*(node: WakuNode, peerList: Option[seq[PeerInfo]]) =
```
oskarth, line 78: I find the resume name not very descriptive. In the case of store protocol it might sense (kind of?) but at node level it doesn't.

Maybe `resumeFetchHistory`? or something that describes the intent better.

jm-clius (approved):
LGTM!

EbubeUd (approved):

docs/api/v2/node.md
```diff
@@ -74,6 +74,16 @@ proc info*(node: WakuNode): WakuInfo =
   ##
```
oskarth, line 74: Missing comment and brief description at top of file:


    Init - create a node.
    Start - start a created node.
    Subscribe - to a topic or a specific content filter.
    Unsubscribe - to a topic or a specific content filter.
    Publish - to a topic, or a topic and a specific content filter.
    Query - for historical messages.
    Info - to get information about the node.

docs/api/v2/node.md
oskarth: Maybe `resume` is fine actually, since all other 7 API calls are simple verbs https://github.com/status-im/nim-waku/blob/master/docs/api/v2/node.md don't know what others think?

docs/api/v2/node.md
staheri14: Sure, done!

staheri14: @oskarth Let me know if further updates are needed.

oskarth (approved):
LGTM thanks!

PR #566, opened 2021-05-24
Automatically add new issues to project board

oskarth:
Same as https://github.com/vacp2p/rfc/pull/372 but for nim-waku - thanks @D4nte for idea!

- [x] add secret GH_ACTION_PROJECT_MGMT

### DIFF
--- a/.github/workflows/add-issue-to-project.yml
+++ b/.github/workflows/add-issue-to-project.yml
@@ -1,0 +1,15 @@
+name: Add new issues Vac PM Board
+
+on:
+  issues:
+    types: [opened]
+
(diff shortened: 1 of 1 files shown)


D4nte (approved):
Is this going to work? As the vac pm board is in a different organization I am not convinced it'll work.

You'll probably have to trigger a repo dispatch from here and then update the action in one of the vacp2p repo to be triggered on repository_dispatch.

oskarth:
Um good point. " trigger a repo dispatch from here and then update the action i" um sounds too complicated :D I'll merge and see if GH can figure it out

PR #584, opened 2021-05-31
Fix: No keep-alive for backoff peers

jm-clius:
This is a stability fix for the keep-alive function.
It ensures that mesh peers being kept alive do not have any backoff period set against them.

This is to address a possible race condition between kicking off [`prune` procedures](https://github.com/status-im/nim-libp2p/blob/master/libp2p/protocols/pubsub/gossipsub/behavior.nim#L42) for a specific peer and [removing that peer](https://github.com/status-im/nim-libp2p/blob/master/libp2p/protocols/pubsub/gossipsub/behavior.nim#L181) from the `mesh`. During this period it has a simultaneous entry in the `backingOff` and `mesh` tables.

### DIFF
--- a/waku/v2/protocol/waku_relay.nim
+++ b/waku/v2/protocol/waku_relay.nim
@@ -28,9 +28,15 @@ proc keepAlive*(w: WakuRelay) {.async.} =
 
     for topic in w.topics.keys:
       trace "Keepalive on topic", topic=topic
-      let mpeers = w.mesh.getOrDefault(topic)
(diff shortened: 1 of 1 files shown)


staheri14 (approved):

oskarth (approved):

jm-clius:
Thanks for reviewing. This is an improvement, but does not yet solve root issue here: https://github.com/status-im/nim-waku/issues/586

PR #588, opened 2021-06-01
Fix keepalive for connected peers

### LINKED ISSUE #586: Bug: keep-alive failing when mesh is full
### Background

`wakunode2` requires a keep-alive mechanism on idle connections to stop them from timing out. This is especially true of peers, such as the fleet nodes, that must remain connected for long periods of time without actually relaying messages to each other.
The correct way to do this is using a ping mechanism built on the libp2p ping protocol. Since this is not yet supported in nim-libp2p (awaiting https://github.com/status-im/nim-libp2p/issues/567), an optional workaround was implemented for `wakunode2` that periodically sends GRAFT messages to peers already in the mesh.

### Problem

The expected behaviour is that mesh peers will ignore these GRAFT messages, but that the connection will remain active. However, if one of these peers' mesh is already full (reached `D_high` capacity), it will not simply ignore the GRAFT keep-alive but respond with a PRUNE message without itself removing that peer from its own mesh. You then have a situation where one peer considers the other to be part of its mesh but the other does not. This causes significant instability.

### Solution

The long-term solution is to implement ping protocol in libp2p (https://github.com/status-im/nim-libp2p/issues/567). In the short term, a workaround must be found, such as a temporary, initial implementation of `ping` or increasing the maximum mesh size.

jm-clius:
This PR fixes #586

## Description

It does so by introducing a simple (and optional) ping-like protocol for `nim-waku` that periodically sends an empty message to connected peers. This a temporary workaround until `libp2p` ping protocol has been implemented, as set out in https://github.com/status-im/nim-libp2p/issues/567. Ping protocol is currently prioritised on the nim-libp2p side, but will likely not be finished before the [planned testnet](https://github.com/status-im/nim-waku/issues/555) for `nim-waku`. Waku v2 dogfooding has also commenced and some `keepalive` mechanism is necessary to maintain stable fleet connections and bridge `toy-chat` messages to Discord.

## How is this different from [the `libp2p` ping protocol](https://docs.libp2p.io/concepts/protocols/#ping)?
It is simpler in at least two ways:
- the keepalive messages do not carry any payload (as opposed to 32-byte ping messages), and
- it does not provide or expect any response. It is therefore simply a best effort to ensure there is minimal activity on open connections.

## Limitations of this approach:
- it is temporary, though implemented in such a way that it can easily be replaced by ping protocol, once available.
- it is not interoperable with non-nim implemenations of Waku. Inter-client `keepalive` will likely only be available once ping protocol is implemented.

### DIFF
--- a/examples/v2/chat2.nim
+++ b/examples/v2/chat2.nim
@@ -248,9 +248,11 @@ proc processInput(rfd: AsyncFD, rng: ref BrHmacDrbgContext) {.async.} =
   await node.start()
 
   if conf.filternode != "":
-    node.mountRelay(conf.topics.split(" "), rlnRelayEnabled = conf.rlnRelay, keepAlive = conf.keepAlive)
+    node.mountRelay(conf.topics.split(" "), rlnRelayEnabled = conf.rlnRelay)
   else:
-    node.mountRelay(@[], rlnRelayEnabled = conf.rlnRelay, keepAlive = conf.keepAlive)
+    node.mountRelay(@[], rlnRelayEnabled = conf.rlnRelay)
+  
+  node.mountKeepalive()
   
   let nick = await readNick(transp)
   echo "Welcome, " & nick & "!"
@@ -376,6 +378,9 @@ proc processInput(rfd: AsyncFD, rng: ref BrHmacDrbgContext) {.async.} =
   node.subscribe(topic, handler)
 
   await chat.readWriteLoop()
+
+  if conf.keepAlive:
+    node.startKeepalive()
 
   runForever()
   #await allFuturesThrowing(libp2pFuts)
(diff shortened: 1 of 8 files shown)


staheri14 (approved):
Looks good!

oskarth (approved):
Hm, at what point does it make sense to just do the PING protocol? Any ETA on it? EDIT: I see issue was just assigned

D4nte:
Are you using a content topic or are messages fully empty?

jm-clius:
> Hm, at what point does it make sense to just do the PING protocol? Any ETA on it? EDIT: I see issue was just assigned

Yeah, discussed with nim-libp2p and it's now prioritised on their side, but I'd like us to be unblocked to continue with testnet/dogfooding. This is an intermediate fix necessary to get the fleets and bridge in a working condition again.

jm-clius:
> Are you using a content topic or are messages fully empty?

Messages are completely empty protobufs, yes. It has no relation to either the PubSub/Waku domains, so contains no reference to content topics or similar. It simply keeps the connection between two switches open.

PR #593, opened 2021-06-03
Filter and lightpush showcase

### LINKED ISSUE #570: Test node with only `filter` and `lightpush`
### Problem

One of the aims of the [upcoming Waku v2 testnet](https://github.com/status-im/nim-waku/issues/569) is to test a limited-resource node participating in the network using only `filter` and `lightpush` capabilities - in other words, a node with no `relay` capability. This issue keeps track of this goal.

### Linked issues

https://github.com/status-im/nim-waku/issues/555

jm-clius:
closes #570  and https://github.com/status-im/nim-waku/issues/516
A `wakunode2` can now fully participate in a network with only `lightpush` and` filter` protocols mounted (i.e. with no `relay` capabilities). 

The toy chat (`chat2`) has been modified to showcase this ability. Just set a `filternode` and `lightpushnode` when running `chat2` and specify `relay=false`.

```
 ./build/chat2 --filternode=/ip4/8.210.222.231/tcp/30303/p2p/16Uiu2HAm4v86W3bmT1BiH6oSPzcsSr24iDQpSN5Qa992BCjjwgrD --lightpushnode=/ip4/8.210.222.231/tcp/30303/p2p/16Uiu2HAm4v86W3bmT1BiH6oSPzcsSr24iDQpSN5Qa992BCjjwgrD --relay=false
```

### DIFF
--- a/examples/v2/chat2.nim
+++ b/examples/v2/chat2.nim
@@ -22,6 +22,7 @@ import   ../../waku/v2/node/[wakunode2, waku_payload],
          ../../waku/v2/protocol/waku_message,
          ../../waku/v2/protocol/waku_store/waku_store,
          ../../waku/v2/protocol/waku_filter/waku_filter,
+         ../../waku/v2/protocol/waku_lightpush/waku_lightpush,
          ../../waku/v2/utils/peers,
          ../../waku/common/utils/nat,
          ./config_chat2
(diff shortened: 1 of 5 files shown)


waku/v2/node/wakunode2.nim
```diff
@@ -747,8 +754,11 @@ when isMainModule:
     waitFor connectToNodes(node, conf.staticnodes)
 
   # NOTE Must be mounted after relay
-  if conf.lightpush:
+  if (conf.lightpushnode != "") or (conf.lightpush):
```
staheri14, line 757: Is it possible that `lightpushnode`   is passes but not `lightpush`? I think checking `lightpush` would be enough, also to disallow wrong set of inputs (i.e.., light push is not set but light push node is passed).

examples/v2/chat2.nim
```diff
@@ -115,6 +116,38 @@ proc showChatPrompt(c: Chat) =
...
+          contentTopic = msg.contentTopic
+      else:
+        debug "Invalid encoded WakuMessage payload",
+          error = decodedPayload.error
+  else:
+    # No payload encoding/encryption from Waku
+    let
+      pb = Chat2Message.init(msg.payload)
+      chatLine = if pb.isOk: pb[].toString()
+                else: string.fromBytes(msg.payload)
```
staheri14, line 144: wondering why the message is shown despite its conversion to a chat2message is failed?

staheri14 (approved):
Looks good! just minor comments below.

oskarth (approved):
Thanks!

waku/v2/node/wakunode2.nim
jm-clius: It is possible, yes. Absolutely agree with you that the two settings should probably be clarified. For now I've just kept this consistent with how `filter` is also done elsewhere (setting either `filternode`, `filter` or both).

In future, though, a "client" for either filter/lightpush may want to set only the server-side as node without providing server capabilities itself (i.e. not really mounting the protocol in the same way). Since we don't have this clear separation in config yet, I'm allowing either setting to be used.

examples/v2/chat2.nim
jm-clius: Initially, this was to be compatible with previous versions of `chat2` that simply passed around the message payload without encoding into `Chat2Message`. For now, it will at least allow those using the testnet content topic to pass around non-specified payloads if they so choose. Since the testnet is designed for experimental use, I thought this a better approach than to quietly drop the message. May help with debugging different clients! 😄

PR #594, opened 2021-06-03
Update submodules

jm-clius:
Updates submodules. Prerequisite for https://github.com/status-im/nim-waku/issues/571

This necessitated a rather large number of changes in `nim-waku`, including:
- better error-handling
- import refactoring

and many miscellaneous changes.

In order to make this process smoother in future, I propose that we update submodules at least once a month.

cc @kdeme for changes to Waku v1 code

### DIFF
--- a/tests/all_tests_v2.nim
+++ b/tests/all_tests_v2.nim
@@ -12,7 +12,7 @@ import
   ./v2/test_jsonrpc_waku,
   ./v2/test_peer_manager,
   ./v2/test_web3, # TODO  remove it when rln-relay tests get finalized
-  ./v2/test_waku_rln_relay,
+  # ./v2/test_waku_rln_relay,
   ./v2/test_waku_bridge,
   ./v2/test_peer_storage,
   ./v2/test_waku_keepalive
--- a/tests/all_tests_v2.nim
+++ b/tests/all_tests_v2.nim
@@ -12,10 +12,13 @@ import
   ./v2/test_jsonrpc_waku,
   ./v2/test_peer_manager,
   ./v2/test_web3, # TODO  remove it when rln-relay tests get finalized
-  # ./v2/test_waku_rln_relay,
   ./v2/test_waku_bridge,
   ./v2/test_peer_storage,
   ./v2/test_waku_keepalive
+
+when defined(rln):
+  import ./v2/test_waku_rln_relay
+
 
 # TODO Only enable this once swap module is integrated more nicely as a dependency, i.e. as submodule with CI etc
 # For PoC execute it manually and run separate module here: https://github.com/vacp2p/swap-contracts-module
--- a/waku/v1/protocol/waku_protocol.nim
+++ b/waku/v1/protocol/waku_protocol.nim
@@ -171,7 +171,7 @@ proc read*(rlp: var Rlp, T: typedesc[StatusOptions]): T =
     of bloomFilterKey:
       let bloom = rlp.read(seq[byte])
       if bloom.len != bloomSize:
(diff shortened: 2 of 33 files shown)


tests/all_tests_v2.nim
```diff
@@ -12,7 +12,7 @@ import
   ./v2/test_jsonrpc_waku,
   ./v2/test_peer_manager,
   ./v2/test_web3, # TODO  remove it when rln-relay tests get finalized
-  ./v2/test_waku_rln_relay,
+  # ./v2/test_waku_rln_relay,
```
jm-clius, line 15: @staheri14, you may find it interesting that updating submodules causes [CI to fail with a "connection refused" `TransportOsError`](https://github.com/status-im/nim-waku/pull/594/checks?check_run_id=2771905449#step:6:3003) on the local `EthClient` address opened in `test_waku_rln_relay`. This only happens during CI, not when testing on a local machine. I've disabled the RLN tests for now, though https://github.com/status-im/nim-waku/pull/602 should similarly work around the issue.

waku/v1/protocol/waku_protocol.nim
```diff
@@ -171,7 +171,7 @@ proc read*(rlp: var Rlp, T: typedesc[StatusOptions]): T =
     of bloomFilterKey:
       let bloom = rlp.read(seq[byte])
       if bloom.len != bloomSize:
-        raise newException(UselessPeerError, "Bloomfilter size mismatch")
+        raise newException(RlpError, "Bloomfilter size mismatch")
```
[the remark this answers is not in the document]
kdeme: Yeah that is fine. It should have been an `RlpError` to begin with as it is RLP decoding of `StatusOptions` (even though that size check is not technically an RlpError I guess).

Though perhaps you better use `RlpTypeMismatch` instead of the general `RlpError`.

kdeme (approved):
I've only looked well at the waku v1 parts. lgtm.

The waku v2 parts have some odd `raises: [Defect, Exception]` in there but I'm assuming that is still needed right now when using libp2p?

waku/v1/protocol/waku_protocol.nim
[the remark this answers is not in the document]
jm-clius: Thanks, @kdeme

jm-clius:
> The waku v2 parts have some odd `raises: [Defect, Exception]` in there but I'm assuming that is still needed right now when using libp2p?

It's needed for some Sqlite DB access functions, but there were a couple of odd ones that I've now removed. Thanks!

kdeme:
> > The waku v2 parts have some odd `raises: [Defect, Exception]` in there but I'm assuming that is still needed right now when using libp2p?
> 
> It's needed for some Sqlite DB access functions, but there were a couple of odd ones that I've now removed. Thanks!

Ideally you should never have raw `Exception` in the raises list. Only `Defect` and any type of Error (objects of `CatchableError`, imo ideally not `CatchableError` itself either, as it makes you possibly miss handling certain errors differently, and it is also simply not very clear for the API user what might go wrong, but there might be cases).

Of course, some modules might still raise raw `Exceptions`. If these can not easily be fixed and you know/think that no errors should be raised and it is some by-product, then I find in generally useful to catch the exception, re-raise it if it is a Defect, and assert it else. But I didn't look into the code to know if this applies here.

tests/all_tests_v2.nim
staheri14: Sorry, missed this comment of yours, I think now that the rln compilation is optional, this problem should not happen, right?

jm-clius: Right! I revered this change once I merged the rln optional stuff.

Issue #582, opened 2021-05-28
WakuMessage timestamp is not set


D4nte:
# Problem

WakuMessage has a `timestamp` field. This field is not set by the `chat2` app for outgoing messages.

Note: this field is an input parameter on the JSON RPC interface of wakunode2

# Notes

@jm-clius proposed that filling the timestamp is likely to be a waku node concern (and not application level), ie, the waku library should set the timestamp when sending a message, not the chat2 app.

If that's the direction that is taken, then I would also suggest to review the fact that the `timestamp` is an optional parameter on the JSON RPC interface of wakunode2. If it should not be an application concern then shouldn't wakunode2 the one to set the timestamp for outgoing messages?

oskarth:
fyi @staheri14 re store node moving to sender by default

staheri14:
@oskarth 
Currently, the waku store defaults to sender time
Are you suggesting removing the receiver timestamp from the waku messages?

oskarth:
The issue is about (sender) timestamp not being set in WakuMessage by default. It probably should be, given that store protocol is relying on it.

jm-clius:
Closing, as `chat2` now sets the timestamp.

PR #597, opened 2021-06-03
Persist waku message timestamp & convert receiver timestamp data type to float64

staheri14:
This PR closes https://github.com/status-im/nim-waku/issues/596
- I also sneaked in and fixed an issue regarding the data type of receiver timestamp. This issue is that the db column holding receiver timestamp is set to `INTEGER` type, while timestamp is a `float64` value. Due to this, we would lose data precision during the insertion to  and reading from the db. The type now is changed to `float64` (the db column holds it as a seq of bytes and proper conversion takes place while reading and writing).
- The db column names are updated to better reflect the origin of the timestamps (two types exist, one for the sender and one for the receiver)

### DIFF
--- a/waku/v2/node/storage/message/waku_message_store.nim
+++ b/waku/v2/node/storage/message/waku_message_store.nim
@@ -31,11 +31,12 @@ proc init*(T: type WakuMessageStore, db: SqliteDatabase): MessageStoreResult[T] 
   let prepare = db.prepareStmt("""
     CREATE TABLE IF NOT EXISTS """ & TABLE_TITLE & """ (
         id BLOB PRIMARY KEY,
-        timestamp INTEGER NOT NULL,
+        receiverTimestamp INTEGER NOT NULL,
         contentTopic BLOB NOT NULL,
         pubsubTopic BLOB NOT NULL,
         payload BLOB,
-        version INTEGER NOT NULL
+        version INTEGER NOT NULL,
+        senderTimestamp DOUBLE NOT NULL
     ) WITHOUT ROWID;
     """, NoParams, void)
 
@@ -59,15 +60,15 @@ method put*(db: WakuMessageStore, cursor: Index, message: WakuMessage, pubsubTop
   ##     echo "error"
   ## 
   let prepare = db.database.prepareStmt(
-    "INSERT INTO " & TABLE_TITLE & " (id, timestamp, contentTopic, payload, pubsubTopic, version) VALUES (?, ?, ?, ?, ?, ?);",
-    (seq[byte], int64, seq[byte], seq[byte], seq[byte], int64),
+    "INSERT INTO " & TABLE_TITLE & " (id, receiverTimestamp, contentTopic, payload, pubsubTopic, version, senderTimestamp) VALUES (?, ?, ?, ?, ?, ?);",
+    (seq[byte], int64, seq[byte], seq[byte], seq[byte], int64, float64),
     void
   )
 
   if prepare.isErr:
     return err("failed to prepare")
 
-  let res = prepare.value.exec((@(cursor.digest.data), int64(cursor.receivedTime), message.contentTopic.toBytes(), message.payload, pubsubTopic.toBytes(), int64(message.version)))
+  let res = prepare.value.exec((@(cursor.digest.data), int64(cursor.receivedTime), message.contentTopic.toBytes(), message.payload, pubsubTopic.toBytes(), int64(message.version), message.timestamp))
   if res.isErr:
     return err("failed")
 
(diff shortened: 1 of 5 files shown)


waku/v2/node/storage/message/waku_message_store.nim
```diff
@@ -23,6 +23,20 @@ type
   WakuMessageStore* = ref object of MessageStore
     database*: SqliteDatabase
 
+proc toBytes(x: float64): seq[byte] =
+  let xbytes =  cast[array[0..7, byte]](x)
+  return @xbytes
+
+proc fromBytes(T: type float64, bytes: seq[byte]): T =
```
oskarth, line 30: Hm, any reason we don't use nim-stew/byteutils for this? Seems error-prone

waku/v2/node/storage/message/waku_message_store.nim
```diff
@@ -31,11 +45,12 @@ proc init*(T: type WakuMessageStore, db: SqliteDatabase): MessageStoreResult[T]
   let prepare = db.prepareStmt("""
     CREATE TABLE IF NOT EXISTS """ & TABLE_TITLE & """ (
         id BLOB PRIMARY KEY,
-        timestamp INTEGER NOT NULL,
+        receiverTimestamp BLOB NOT NULL,
```
oskarth, line 48: Does this require db migrations? What does a graceful upgrade path look like here?

jm-clius:
Agree re migration - perhaps we should update the `user_version` and upgrade all earlier versions on startup?

For interest's sake, do we not support [`REAL` type](https://www.sqlite.org/datatype3.html)?

waku/v2/node/storage/message/waku_message_store.nim
staheri14: Haven't found any support for `float64` in nim-stew/byteutils

staheri14:
> Agree re migration - perhaps we should update the `user_version` and upgrade all earlier versions on startup?
> 
> For interest's sake, do we not support [`REAL` type](https://www.sqlite.org/datatype3.html)?

Good question,`REAL` is certainly supported by sqlite3 in general, but it is missing in the `sqlite.nim` module and inside the parameters binding proc i.e., https://github.com/status-im/nim-waku/blob/b615c6aa8bb90062ac157a49b77cd838e8d8a925/waku/v2/node/storage/sqlite.nim#L130 (perhaps it can be added, I will work on it in the subsequent PRs)
Conversion to the byte sequence is a workaround that works fine.

staheri14:
> Agree re migration - perhaps we should update the `user_version` and upgrade all earlier versions on startup

@jm-clius 
Sure, I am reading up about migration through `user-version` pragma. In the meantime, if you have seen or done similar migration, that would be great if you could please let me know or give me some pointers, Thanks!

waku/v2/node/storage/message/waku_message_store.nim
staheri14: A short-term solution is to delete the existing db given that no important data would be lost. In the meanwhile, I am working on a decent migration approach using `user-version` pragma.

oskarth:
Not exactly the same solution, but perhaps some inspiration re DB migrations in Nimbus:

 https://github.com/status-im/nimbus-eth2/blob/eb55fe241908b35db179aea04ae1db37a7d18b9e/beacon_chain/validators/slashing_protection_v2.nim#L628 and https://github.com/status-im/nimbus-eth2/blob/eb55fe241908b35db179aea04ae1db37a7d18b9e/tests/slashing_protection/test_migration.nim
 
 Appears they don't use sqlite built-in style, but instead did this to migrate from KV DB to sqlite.

EbubeUd (approved):

D4nte:
Is this fixing https://github.com/status-im/nim-waku/issues/582 too?

jm-clius:
If found [this](https://levlaz.org/sqlite-db-migrations-with-pragma-user_version/) blog post which I think does a fair explanation of using the `user_version` pragma when migrating. You'll have to bump `user_version` and then run the migration script for all previous `user_versions`. We'll likely build up a list of migration scripts this way (we can keep them in a separate folder/module) that will have to be run in consecutive order if someone were to upgrade from, say `user_version=1` to `user_version=10`.

staheri14:
> Is this fixing #582 too?

@D4nte 
No, it is only about storing message timestamps in the db but not generating them.

staheri14:
> If found [this](https://levlaz.org/sqlite-db-migrations-with-pragma-user_version/) blog post which I think does a fair explanation of using the `user_version` pragma when migrating. You'll have to bump `user_version` and then run the migration script for all previous `user_versions`. We'll likely build up a list of migration scripts this way (we can keep them in a separate folder/module) that will have to be run in consecutive order if someone were to upgrade from, say `user_version=1` to `user_version=10`.

Thanks @jm-clius, @oskarth for your direction. I am on it!

staheri14:
For this PR, I am assuming that it  should wait for the migration script to be finalized. Rather than that, it is ready to be merged. @jm-clius @oskarth

jm-clius:
Yes, let's wait for the migration script. Thanks!

waku/v2/node/storage/message/waku_message_store.nim
oskarth: Might want to create an upstream issue (and keep code local here for now)

oskarth:
Sounds good to me!

staheri14:
Please see this PR https://github.com/status-im/nim-waku/pull/607 for the db migration, it also covers all the changes in this PR with additional fixes

PR #599, opened 2021-06-04
Append instance and job to expr string in json file for Grafana Dashboard

### LINKED ISSUE #598: Local Grafana Dashboard does not display Metrics.
**Problem**

Currently, following the procedure on [using waku metrics](https://github.com/status-im/nim-waku/tree/master/waku/v2#using-metrics  ) sends data successfully to the prometheus server but fails to display the metrics on the Grafana Dashboard. However, this works well for Nimbus-eth.

**Possible Solution**
Just as @jm-clius suggested [here](https://github.com/status-im/nim-waku/pull/585#issuecomment-853755806), there was an issue with the expression In the json config file for the grafana dashboard.
Appending  `{instance=\"localhost:8008\",job=\"waku\"}` to the `expr` value, similar to what was done [here](https://github.com/status-im/nimbus-eth1/blob/ef7773daa6af7cf71916c31526b1257943498754/examples/Nimbus-Grafana-dashboard.json#L56) on nimbus-eth1 seems to resolve this issue.  

There could be a better alternative, but this seems to work well.

EbubeUd:
closes #598

### DIFF
--- a/metrics/waku-grafana-dashboard.json
+++ b/metrics/waku-grafana-dashboard.json
@@ -67,25 +67,25 @@
       "pluginVersion": "7.3.7",
       "targets": [
         {
-          "expr": "libp2p_pubsub_peers{node=\"0\"}",
+          "expr": "libp2p_pubsub_peers{instance=\"localhost:8008\",job=\"waku\"}",
           "interval": "",
           "legendFormat": "LibP2P PubSub Peers",
           "refId": "A"
         },
         {
-          "expr": "waku_filter_peers{node=\"0\"}",
+          "expr": "waku_filter_peers{instance=\"localhost:8008\",job=\"waku\"}",
           "interval": "",
           "legendFormat": "Filter Peers",
           "refId": "B"
         },
         {
-          "expr": "waku_store_peers{node=\"0\"}",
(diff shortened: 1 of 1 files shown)


jm-clius (changes requested):
The grafana dashboard that you edited is a template. This one is used, for example, to script new dashboards, here: https://github.com/status-im/nim-waku/blob/master/waku/v2/node/start_network2.nim#L103

You should configure it for your local instance locally.

EbubeUd:
> The grafana dashboard that you edited is a template. It should not be edited with local-specific configuration. This one is used, for example, to script new dashboards, here: https://github.com/status-im/nim-waku/blob/master/waku/v2/node/start_network2.nim#L103
> 
> You should configure it for your local instance locally.

Ohh alright I get it now. It'll be nice if we can also add this to the doc on setting up metrics, so subsequent users know they're meant to append `{instance="localhost:8008", job="waku"}` in the query field for each of the panels.

> https://github.com/status-im/nim-waku/blob/master/waku/v2/node/start_network2.nim#L103

EbubeUd:
I think this PR should be closed. We can make necessary updates to the doc [here](https://github.com/status-im/nim-waku/tree/master/waku/v2#using-metrics)   if need be.

jm-clius:
Closing pull request.

PR #600, opened 2021-06-04
Filter out duplicate messages in the resume procedure

staheri14:
Closes https://github.com/status-im/nim-waku/issues/592


oskarth (commented):
How does this relate to https://github.com/status-im/nim-waku/pull/597 in terms of order of merging? Is this also blocked by db migration script?

waku/v2/protocol/waku_store/waku_store.nim
```diff
@@ -526,17 +535,41 @@ proc resume*(ws: WakuStore, peerList: Option[seq[PeerInfo]] = none(seq[PeerInfo]
...
   debug "the  offline time window is", lastSeenTime=lastSeenTime, currentTime=currentTime
 
+  var dismissed: uint = 0
+  var added: uint = 0
+
   proc handler(response: HistoryResponse) {.gcsafe.} =
+    debug "resume handler is called"
+
+    # exclude index from the comparison criteria
+    let currentMsgSummary = ws.messages.map(proc(x: IndexedWakuMessage): WakuMessage = x.msg)
```
jm-clius, line 545: I'm wondering if there is a way to do this more effectively. I don't think this is the only place where we essentially create a copy of the existing history - perhaps we could instead maintain a single, ordered hash set of messages in the store? This may also speed up `isDuplicate` - finding a match should be more effective on a `HashSet` than on a `seq`

jm-clius:
As far as I can tell this is now outdated after https://github.com/status-im/nim-waku/pull/607/? I will review again once those changes have been merged.

waku/v2/protocol/waku_store/waku_store.nim
staheri14: Thanks for the comment! Agree, the hash set would work here and indeed removes the duplicates by default (no need to even check). Nevertheless, I expect that the hash set should be slower than the normal seq in terms of insertion (which is the most recurring operation in the store protocol) as it involves hashing. Also, wrt storage, I think hash sets consume more space than sequences (depending on the HashSet implementation).  Let me investigate it a bit further and make sure there would not be a huge performance issue with the hash set compared to seq, and then I will update it.

staheri14: Here is the link to a discussion in Nimbus help-desk about hashset vs seq: https://discord.com/channels/613988663034118151/616299964242460682/855126317498105887
Following this conversation, either we can do a benchmarking on HashSet and seq and pick the one that performs the best, or we can proceed with replacing seq with HashSet with no stress test. There is a third option of keeping it as seq. 
@oskarth @jm-clius Let me know your thoughts.

staheri14:
> How does this relate to #597 in terms of order of merging? Is this also blocked by db migration script?

Yes, it was blocked, but now it is ready to review.

waku/v2/protocol/waku_store/waku_store.nim
jm-clius: If I understand the thread of conversation, moving to `HashSet` will require at least one less copy of the historical messages? It also naturally deduplicates which will simplify what we have, so if it's a relatively easy change I'd go with "replacing seq with HashSet with no stress test". We still have plans to do proper benchmarking of all our protocols, so this could be a future task.
I'm also fine with the change to HashSet being deferred to a future PR.

jm-clius (approved):
LGTM.

waku/v2/protocol/waku_store/waku_store.nim
staheri14: > We still have plans to do proper benchmarking of all our protocols, so this could be a future task

Oh, great! Then I will replace it with HashSet, though I will do it in a separate PR, as it has a large surface of changes namely, the pagination, DB handlers, the history query internals, and many more.

oskarth (approved):
LGTM; needs rebase

PR #601, opened 2021-06-06
Add mode to waku swap

### LINKED ISSUE #587: Add Mode to Waku Swap Object
**Description:** 

Before the swap protocol is mounted, there needs to be an option to specify the mode it will be mounted with. 
The Swap Modes are well detailed here [#524](https://github.com/vacp2p/research/discussions/61).

The mode will determine:
  - What type of Swap Account Logs to be displayed
  - The test cases that will be checked. 
  - The functionality of the policy function (Send Cheque / Disconnect peer)

**Suggestion**
The mode should probably be in the form of an `enum` in the `WakuSwap` Object. The enum should contain: 
  - Soft
  - Mock 
  - Hard

The initial value for this enum will be set in the init function of WakuSwap and the tests will be modified to be dependent on the value of enum. The  full implementation of the policy function will also be made to be dependent on the mode.

EbubeUd:
closes #587

**Current Implementation:**
- Added `mode` to swap protocol
- Set `mode` to Soft on init.
- Execute SendCheque in Policy function if payment Threshold is reached
- Run test to check to balance after sendCheque function has been called.

### DIFF
--- a/tests/v2/test_waku_swap.nim
+++ b/tests/v2/test_waku_swap.nim
@@ -106,12 +106,15 @@ procSuite "Waku SWAP Accounting":
 
     var futures = [newFuture[bool](), newFuture[bool]()]
 
+    # Define the waku swap Config for this test
+    let swapConfig = SwapConfig(mode: SwapMode.Mock, paymentThreshold: 1, disconnectThreshold: -1)
+
     # Start nodes and mount protocols
     await node1.start()
-    node1.mountSwap()
+    node1.mountSwap(swapConfig)
     node1.mountStore(persistMessages = true)
     await node2.start()
-    node2.mountSwap()
+    node2.mountSwap(swapConfig)
     node2.mountStore(persistMessages = true)
 
     await node2.subscriptions.notify("/waku/2/default-waku/proto", message)
@@ -131,16 +134,11 @@ procSuite "Waku SWAP Accounting":
     await node1.query(HistoryQuery(contentFilters: @[HistoryContentFilter(contentTopic: contentTopic)]), handler1)
     await node1.query(HistoryQuery(contentFilters: @[HistoryContentFilter(contentTopic: contentTopic)]), handler2)
 
-    # Only check these conditions if the Swap was mounted In the Mock Mode
-    if node1.wakuSwap.mode == SwapMode.Mock and node2.wakuSwap.mode == SwapMode.Mock:
-      check:
-        (await allFutures(futures).withTimeout(5.seconds)) == true
-        # Accounting table updated with credit and debit, respectively
-        # After sending a cheque the balance is partially adjusted
-        node1.wakuSwap.accounting[node2.peerInfo.peerId] == 1
-        node2.wakuSwap.accounting[node1.peerInfo.peerId] == -1
-    else:
-      check:
-        true
+    check:
+      (await allFutures(futures).withTimeout(5.seconds)) == true
+      # Accounting table updated with credit and debit, respectively
+      # After sending a cheque the balance is partially adjusted
+      node1.wakuSwap.accounting[node2.peerInfo.peerId] == 1
+      node2.wakuSwap.accounting[node1.peerInfo.peerId] == -1
     await node1.stop()
     await node2.stop()
--- a/tests/v2/test_waku_swap.nim
+++ b/tests/v2/test_waku_swap.nim
@@ -91,8 +91,8 @@ procSuite "Waku SWAP Accounting":
     await node1.stop()
     await node2.stop()
 
-  #TODO Add cheque here
-  #This test will only Be checked if in Mock mode
+  # TODO Add cheque here
+  # This test will only Be checked if in Mock mode
   asyncTest "Update accounting state after sending cheque":
     let
       nodeKey1 = crypto.PrivateKey.random(Secp256k1, rng[])[]
--- a/tests/v2/test_waku_swap.nim
+++ b/tests/v2/test_waku_swap.nim
@@ -91,51 +91,56 @@ procSuite "Waku SWAP Accounting":
     await node1.stop()
     await node2.stop()
 
-  # TODO Add cheque here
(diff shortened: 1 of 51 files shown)


waku/v2/protocol/waku_swap/waku_swap_types.nim
```diff
@@ -6,6 +6,14 @@ import
   ../../node/peer_manager/peer_manager  
 
 type
+   #The Swap Mode determines the functionality available in the swap protocol.
```
jm-clius, line 9: Make sure your indentation is consistent here. 
Also, very pedantic from my side, but if there is a space between the `#` and the first word in that comment, it becomes easier to read, spell-check and add to docs. 😃

waku/v2/protocol/waku_swap/waku_swap_types.nim
```diff
@@ -6,6 +6,14 @@ import
...
+  #It determines the kind of logs to be diplayed as well as tests to be ran.
```
jm-clius, line 10:
```suggestion
  #It determines the kind of logs to be displayed as well as tests to run.
```

tests/v2/test_waku_swap.nim
```diff
@@ -91,51 +91,56 @@ procSuite "Waku SWAP Accounting":
...
+      futures[0].complete(true)
+    proc handler2(response: HistoryResponse) {.gcsafe, closure.} =
+      futures[1].complete(true)
+
+    # TODO Handshakes - for now we assume implicit, e2e still works for PoC
+    await node1.query(HistoryQuery(contentFilters: @[HistoryContentFilter(contentTopic: contentTopic)]), handler1)
+    await node1.query(HistoryQuery(contentFilters: @[HistoryContentFilter(contentTopic: contentTopic)]), handler2)
+
+    # Only check these conditions if the Swap was mounted In the Mock Mode
+    if node1.wakuSwap.mode == SwapMode.Mock and node2.wakuSwap.mode == SwapMode.Mock:
```
jm-clius, line 135: Rather than setting up the test only to discover at this point that `Mock` mode is not configured, you may want to add `swapMode` as an argument to the `mountSwap` function (with default value `Soft`). That way the test can be set up to use `Mock` mode.

jm-clius (changes requested):
Great! Just a couple of changes, mostly around making this just slightly more configurable while still defaulting to `Soft`-mode for now.

waku/v2/protocol/waku_swap/waku_swap_types.nim
EbubeUd: Thanks for the review. The spacing format is worth adhering to. ☀︎

waku/v2/node/wakunode2.nim
```diff
@@ -393,9 +393,9 @@ proc mountFilter*(node: WakuNode) =
 
 # NOTE: If using the swap protocol, it must be mounted before store. This is
 # because store is using a reference to the swap protocol.
-proc mountSwap*(node: WakuNode) =
+proc mountSwap*(node: WakuNode, swapConfig: SwapConfig = SwapConfig(mode: SwapMode.Soft, paymentThreshold: 100, disconnectThreshold: -100)) =
```
jm-clius, line 396: One thing you may still want to do is provide an `init*()`-function for `SwapConfig` where you set the default values  (somewhat like a default constructor). That way users of the type can choose to only set the config fields they're specifically interested in and leave the rest as default. See [this example](https://github.com/status-im/nim-libp2p/blob/master/libp2p/protocols/pubsub/gossipsub.nim#L41-L71) for `GossipSubParams` in libp2p.
This can be in a subsequent PR though.

jm-clius (approved):
LGTM! Well done. Hold off on merging though, until the release is done (currently blocked by https://github.com/status-im/nim-waku/pull/610)

waku/v2/node/wakunode2.nim
EbubeUd: Thanks for pointing out. I have made the update and learnt something new as well.

waku/v2/node/wakunode2.nim
```diff
@@ -393,9 +393,9 @@ proc mountFilter*(node: WakuNode) =
 
 # NOTE: If using the swap protocol, it must be mounted before store. This is
 # because store is using a reference to the swap protocol.
-proc mountSwap*(node: WakuNode) =
-  info "mounting swap"
-  node.wakuSwap = WakuSwap.init(node.peerManager, node.rng)
+proc mountSwap*(node: WakuNode, swapConfig: SwapConfig = SwapConfig()) =
```
jm-clius, line 396: Now that you have an `init` for `SwapConfig` you could do something like:

```suggestion
proc mountSwap*(node: WakuNode, swapConfig = SwapConfig.init()) =
```

EbubeUd: Thanks. Made the changes. 

But what is the difference between having `SwapConfig()` this way and specifying `init` in `SwapConfig.init()`?
Both seem to work.

jm-clius: An empty constructor, like `SwapConfig()` will initialise each field with the default value for that _type_ (e.g. `paymentThreshold`, `disconnectThreshold` will be `0`, as that is the default for `int`). `init()` will construct a `SwapConfig` object by calling the `init()` function you defined, in other words use the default values as specified by you.

EbubeUd: Thanks. This was helpful.

waku/v2/protocol/waku_swap/waku_swap_types.nim
```diff
@@ -6,6 +6,19 @@ import
   ../../node/peer_manager/peer_manager  
 
 type
+  # The Swap Mode determines the functionality available in the swap protocol.
```
oskarth, line 9: I'd make the docstring self contained and explain briefly what each mode implies here, as opposed to linking to a (possibly in the future outdated) discussion

oskarth (approved):
LGTM! Minor comment on docstring

PR #602, opened 2021-06-07
Make rln optional (dependencies and compilation)

staheri14:
# Problem
This PR isolates rln, disables it by default unless proper flags are set.
- rln-related codes and tests will not be compiled unless the `rln` flag is set (i.e., `nim c -d:rln`). 
- all of the rln-related build/dependency targets in the Make file are conditioned to the `RLN` variable being true. This variable affects two targets: namely, `rlnlib` that fetches and builds the rln dynamic library, and `installganache` that spins up ganache-cli.  When the `RLN` var also controls the rln compilation flag.
- Run `make <target> -RLN=true` to enable compilation of rln and to resolve its dependencies. 

It is also a  workaround for https://github.com/status-im/nim-waku/issues/546

### DIFF
--- a/tests/v2/test_waku_rln_relay.nim
+++ b/tests/v2/test_waku_rln_relay.nim
@@ -1,560 +1,559 @@
-{.used.}
+when defined(rln):
+  {.used.}
 
-import
-  std/options,
-  testutils/unittests, chronos, chronicles, stint, web3,
-  stew/byteutils, stew/shims/net as stewNet,
-  libp2p/crypto/crypto,
-  ../../waku/v2/protocol/waku_rln_relay/[rln, waku_rln_relay_utils],
-  ../../waku/v2/node/wakunode2,
-  ../test_helpers,
-  ./test_utils
-
-# the address of Ethereum client (ganache-cli for now)
-# TODO this address in hardcoded in the code, we may need to take it as input from the user
-const EthClient = "ws://localhost:8540/"
-
-# poseidonHasherCode holds the bytecode of Poseidon hasher solidity smart contract: 
-# https://github.com/kilic/rlnapp/blob/master/packages/contracts/contracts/crypto/PoseidonHasher.sol 
-# the solidity contract is compiled separately and the resultant bytecode is copied here
-const poseidonHasherCode = readFile("tests/v2/poseidonHasher.txt")
-# membershipContractCode contains the bytecode of the membership solidity smart contract:
-# https://github.com/kilic/rlnapp/blob/master/packages/contracts/contracts/RLN.sol
-# the solidity contract is compiled separately and the resultant bytecode is copied here
-const membershipContractCode = readFile("tests/v2/membershipContract.txt")
-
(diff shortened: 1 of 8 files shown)


EbubeUd (approved):

EbubeUd:
Tested with the M1 on this branch and it seems to work very well. Thanks.

tests/v2/test_waku_rln_relay.nim
```diff
@@ -1,560 +1,559 @@
...
-
-  return contractAddress
-
-procSuite "Waku rln relay":
-  asyncTest  "contract membership":
-    let contractAddress = await uploadContract(EthClient)
-    # connect to the eth client
-    let web3 = await newWeb3(EthClient)
-    debug "web3 connected to", EthClient
+when defined(rln):
```
jm-clius, line 1: Why not just disable this whole suite here: https://github.com/status-im/nim-waku/blob/master/tests/all_tests_v2.nim#L15 ?

waku/v2/protocol/waku_rln_relay/rln.nim
```diff
@@ -2,61 +2,62 @@
...
-                              ctx: ptr (ptr RLN[Bn256])): bool {.importc: "new_circuit_from_params".}
-
-proc hash*(ctx: ptr RLN[Bn256],
-           inputs_buffer: ptr Buffer,
-           input_len: uint,
-           output_buffer: ptr Buffer): bool {.importc: "hash".}
-{.pop.}
\ No newline at end of file
+#  the following module is only available and get compiled if the compiler flag rln is passed as -d:rln 
+when defined(rln):
```
jm-clius, line 6: Not sure if this will work, but instead of surrounding the code within a conditional block, why not rather just add the `when defined` wherever this module gets imported? If it never gets imported it should never be compiled, or is it more complicated than that?

tests/v2/test_waku_rln_relay.nim
staheri14: Sure, done!

waku/v2/protocol/waku_rln_relay/rln.nim
staheri14: This and also all the calls to the procedures defined in that module should be wrapped insider `when defined(rln)` otherwise the compiler complains about not being able to find the procedure's definition.

staheri14:
> Thanks! This will also help with some issues I've been having with RLN after updating the submodules :)
> 
> I've left some comments on rather excluding modules from being imported at all than surrounding a full module in a conditional `when defined`-block - not sure if this will work, though, so up to you.

Thanks for the comments! and glad that it is a helpful change!

staheri14:
> Tested with the M1 on this branch and it seems to work very well. Thanks.

@EbubeUd 
Great!

waku/v2/protocol/waku_rln_relay/rln.nim
staheri14: Done!

PR #607, opened 2021-06-10
Persisting Waku message timestamp & implementing DB migration & convert receiver timestamp data type to float64

staheri14:
This PR closes https://github.com/status-im/nim-waku/issues/596 and closes https://github.com/status-im/nim-waku/issues/603
This PR replaces this one https://github.com/status-im/nim-waku/pull/597.  It overwrites the prior PR to a large extent and also provides fixes on top of that, hence deserves a fresh review. 

It embodies the following major changes:
- Implements and integrates DB migration
- Persists waku messages timestamp inside the Message table
- Enables the storage and passing of float64 values as a parameter of SQLite commands
- fixes the issue regarding the type of the column storing receiver timestamp

**Summary of the edits**:
- Adds `senderTimestamp` to the `Message` table and also persists waku message timestamps in this column
- It also contains a fix for an issue regarding the data type of the receiver-side timestamp. The issue is that the DB column holding receiver timestamp is set to `INTEGER` type, while the timestamp is a `float64` value. This results in precision loss during insertion to/ reading from the DB. The type is now changed to `float64`. 
- I also managed to resolve the issue re storing `float64` values inside the DB, in the previous PR (https://github.com/status-im/nim-waku/pull/597) I had to save them in the binary format. The problem was that the `sqlite.nim` module had no support to bind `float64` values to the parameters of a SQL statement. After some investigation, I found the related nim binding in the `sqlite3_abi` module. This binding is now added inside `bindParam` of the `sqlite.nim` module where the application data type of `float64` can be attached to a statement parameter type of `double` ( a subtype of `REAL`).  This enabled the storage of timestamps inside the database as `REAL` instead of `BLOB`.
- The DB column names are updated to better reflect the origin of the timestamps (two types exist, one for the sender and one for the receiver)
- DB migration is implemented, it works using user_version pragma
  - DB migration utils and types are added.
  - DB migration proc is implemented and included under the `waku_message_store`. This proc handles DB migration scripts relevant to the message store. However, the implementation is general and can support potential peer store migrations as well.
   - Two utility procs for setting and getting `user_version` are introduced inside the `sqlite.nim` module.
   -  Appropriate SQL migration script is written for the `user_version = 2`.

**Future enhancements**:
-~Atomic execution of SQL migration scripts: currently, SQL migration scripts are executed one command at a time, if some failure happens in between, then there might be some commands left unexecuted and the migration would not go through as a whole. Atomix execution makes sure either all the commands of the same script are executed or none.~ Already addressed!
- Parsing and executing SQL migration scripts that contain comments


staheri14:
@richard-ramos I would be thankful to know your insights and comments on the db migration part!

waku/v2/node/storage/message/waku_message_store.nim
```diff
@@ -115,3 +130,60 @@ method getAll*(db: WakuMessageStore, onData: message_store.DataProc): MessageSto
...
+      # a script may contain multiple queries
+      let queries = script.splitScript()
+      # TODO queries of the same script should be executed in an atomic manner
+      for query in queries:
+        let res = db.query(query, handler)
+        if res.isErr:
+          debug "failed to run the query", query=query
+          return err("failed to run the script")
+        else:
+          debug "query is executed", query=query
```
richard-ramos, line 181: I wonder if this could be somehow improved by using SQL transactions to rollback data manipulation changes (inserts/updates), and then, after returning the error, attempt to execute the rollback script for the specific migration script that failed.

waku/v2/node/storage/message/waku_message_store.nim
```diff
@@ -115,3 +130,60 @@ method getAll*(db: WakuMessageStore, onData: message_store.DataProc): MessageSto
...
+    
+    
+    proc handler(s: ptr sqlite3_stmt) = 
+      discard
+    
+    # run the scripts
+    for script in scripts:
+      debug "script", script=script
+      # a script may contain multiple queries
+      let queries = script.splitScript()
```
richard-ramos, line 173: Is splitting the sql scripts strictly necessary? SQLite should be able to execute multiple scripts in a single execution

waku/v2/node/storage/sqlite.nim
```diff
@@ -208,3 +209,28 @@ proc close*(db: SqliteDatabase) =
...
+
+
+proc setUserVerion*(database: SqliteDatabase, version: int64): DatabaseResult[bool] = 
+  ## sets  the value of the user-version integer at offset 60 in the database header. 
+  ## some context borrowed from https://www.sqlite.org/pragma.html#pragma_user_version
+  ## The user-version is an integer that is available to applications to use however they want. 
+  ## SQLite makes no use of the user-version itself
+  proc handler(s: ptr sqlite3_stmt) = 
+    discard
+  let query = "PRAGMA user_version=" & $version & ";"
```
richard-ramos, line 230: TIL about `PRAGMA user_version`, very cool!

waku/v2/node/storage/message/waku_message_store.nim
```diff
@@ -33,11 +38,12 @@ proc init*(T: type WakuMessageStore, db: SqliteDatabase): MessageStoreResult[T]
   let prepare = db.prepareStmt("""
     CREATE TABLE IF NOT EXISTS """ & TABLE_TITLE & """ (
```
richard-ramos, line 39: Could this `CREATE TABLE` become part of the migration scripts? I guess it depends on how important it is being able to set the table name, right?

waku/v2/node/storage/sqlite.nim
```diff
@@ -208,3 +209,28 @@ proc close*(db: SqliteDatabase) =
...
+  var version: int64
+  proc handler(s: ptr sqlite3_stmt) = 
+    version = sqlite3_column_int64(s, 0)
+  let res = database.query("PRAGMA user_version;", handler)
+  if res.isErr:
+      return err("failed to get user_version")
+  ok(version)
+
+
+proc setUserVerion*(database: SqliteDatabase, version: int64): DatabaseResult[bool] =
```
richard-ramos, line 223:
```suggestion
proc setUserVersion*(database: SqliteDatabase, version: int64): DatabaseResult[bool] = 
```

waku/v2/node/storage/sqlite.nim
```diff
@@ -208,3 +209,28 @@ proc close*(db: SqliteDatabase) =
   discard sqlite3_close(db.env)
 
   db[] = SqliteDatabase()[]
+
+proc getUserVerion*(database: SqliteDatabase): DatabaseResult[int64] =
```
richard-ramos, line 213:
```suggestion
proc getUserVersion*(database: SqliteDatabase): DatabaseResult[int64] = 
```

waku/v2/node/storage/message/waku_message_store.nim
```diff
@@ -115,3 +130,60 @@ method getAll*(db: WakuMessageStore, onData: message_store.DataProc): MessageSto
 proc close*(db: WakuMessageStore) = 
   ## Closes the database.
   db.database.close()
+
+
+proc migrate*(db: SqliteDatabase, path: string = MIGRATION_PATH, tragetVersion: int64 = USER_VERSION): MessageStoreResult[bool] =
```
richard-ramos, line 135:
```suggestion
proc migrate*(db: SqliteDatabase, path: string = MIGRATION_PATH, targetVersion: int64 = USER_VERSION): MessageStoreResult[bool] = 
```

waku/v2/node/storage/message/waku_message_store.nim
staheri14: Thanks! I tried passing the whole script and it worked. I recall trying it before, but it did not get through for some reason (I think the script had some comments in it which weren't properly handled by SQlite).

waku/v2/node/storage/message/waku_message_store.nim
staheri14: Indeed it can be part of the migration scripts and I cannot think of any issue why it should not be. @oskarth @jm-clius wdyt? if there is no concern, then I can separate it as a migration script.

staheri14:
Thanks @richard-ramos for the review! very helpful and insightful comments!

waku/v2/node/storage/message/waku_message_store.nim
staheri14: Re the table name, you are right, however, it is currently hardcoded in the code and is not a user input. As such, effectively there is no difference in having it as part of the migration scripts or in the code.

tests/v2/test_message_store.nim
```diff
@@ -15,38 +16,100 @@ suite "Message Store":
...
+
+    let res = database.setUserVersion(5)
+    check res.isErr == false
 
+    let ver = database.getUserVersion()
+    check:
+      ver.isErr == false
+      ver.value == 5
+  test "migration":
+  # This is commented out but is useful for future debugging
```
oskarth, line 99: Leftover comment? Don't see anything commented out

waku/v2/node/storage/message/waku_message_store.nim
```diff
@@ -91,22 +97,31 @@ method getAll*(db: WakuMessageStore, onData: message_store.DataProc): MessageSto
   proc msg(s: ptr sqlite3_stmt) {.raises: [Defect, Exception].} =
     gotMessages = true
     let
-      timestamp = sqlite3_column_int64(s, 0)
+      receiverTimestamp = sqlite3_column_double(s, 0)
+
       topic = cast[ptr UncheckedArray[byte]](sqlite3_column_blob(s, 1))
       topicL = sqlite3_column_bytes(s,1)
```
oskarth, line 103: Is this topic length? I'd call it `topicLen` or something like this. Similar with `p` and `l` - not very descriptive, have a hard time following

waku/v2/node/storage/message/waku_message_store.nim
```diff
@@ -115,3 +130,55 @@ method getAll*(db: WakuMessageStore, onData: message_store.DataProc): MessageSto
 proc close*(db: WakuMessageStore) = 
   ## Closes the database.
   db.database.close()
+
+
+proc migrate*(db: SqliteDatabase, path: string = MIGRATION_PATH, targetVersion: int64 = USER_VERSION): MessageStoreResult[bool] = 
+  ## compares the user_version of the db with the targetVersion 
+  ## runs migration scripts if the user_version is outdated (does not support down migration)
```
oskarth, line 137: Should we add a basic `markdown` file in docs folder re how migrations work? Ie something like:

"Migrations should work out of the box, however if you want to be extra sure take a backup of SQL DB first since we currently don't support downgrades of DB"

oskarth: Then we can point people to this every release we do with db migrations

waku/v2/node/storage/migration/migrations_scripts/message/00002_addSenderTimeStamp.up.sql
```diff
@@ -0,0 +1,27 @@
+CREATE TABLE IF NOT EXISTS Message_backup (
```
oskarth, line 1: Do we expect there might be more backups in future? If so, we might want to number these like `Message_backup1` or so (user version in it?)

waku/v2/node/storage/message/waku_message_store.nim
oskarth: This doc can also mention to contributors to bump user_version etc

waku/v2/node/storage/sqlite.nim
```diff
@@ -110,8 +109,8 @@ proc init*(
       discard sqlite3_finalize(journalModePragma)
       return err("Invalid pragma result: " & $x)
 
-  # TODO: check current version and implement schema versioning
-  checkExec "PRAGMA user_version = 1;"
+  # checkExec "PRAGMA user_version = 1;"
```
oskarth, line 112: Comment on why commented out or remove

oskarth (approved):
LGTM, nice! Minor comments

waku/v2/node/storage/migration/migrations_scripts/message/00002_addSenderTimeStamp.up.sql
jm-clius, line 1: I think this script should only run if `Message` actually exists (i.e. add an `IF EXISTS Message` that determines if script is run or not). 
I might be missing something, but as it stands, migrations will always run if there is a `dbPath` set - whether `store` and `persistMessages` are enabled or not. I think migration scripts for specific tables should only run if those tables were created before.

waku/v2/node/wakunode2.nim
```diff
@@ -695,7 +695,15 @@ when isMainModule:
       waku_node_errors.inc(labelValues = ["init_db_failure"])
     else:
       sqliteDatabase = dbRes.value
-  
+    
+    # run migration 
+    info "running migration ... "
```
jm-clius, line 700: This will run migration scripts without considering what tables should be created according to the rest of the config (i.e. `persistMessages` or `persistPeers`). This is totally fine, as long as the scripts themselves only run if the tables they migrate already exists (see my comment elsewhere).

waku/v2/node/storage/message/waku_message_store.nim
jm-clius: It could be done, but I would prefer it if it wasn't part of the migration. Since the SQLite tables should only be created if the node is configured to do so (i.e. if `persistMessages` or `persistPeers` are enabled), the migration would have to be aware of these config items to decide whether or not the `CREATE TABLE` should be run. In my view the order in which things should happen:
1. Node starts
2. Node checks if it has an existing SQLite DB (`dbPath`)
3. If true, node migrates DB to desired version (with each script only running for tables that already exists).
4. Node creates new tables based on configuration.

This is more or less what is already being done, with only (3) requiring modification to the scripts.

waku/v2/node/storage/message/waku_message_store.nim
```diff
@@ -115,3 +130,55 @@ method getAll*(db: WakuMessageStore, onData: message_store.DataProc): MessageSto
 proc close*(db: WakuMessageStore) = 
   ## Closes the database.
   db.database.close()
+
+
+proc migrate*(db: SqliteDatabase, path: string = MIGRATION_PATH, targetVersion: int64 = USER_VERSION): MessageStoreResult[bool] =
```
jm-clius, line 135: I think `migrate` is agnostic of what table(s) it migrates? If so, this function can move out of `waku_message_store` to https://github.com/status-im/nim-waku/blob/master/waku/v2/node/storage/sqlite.nim. This will signal that it's multipurpose and related to the DB and not just a single table (excited that this can be used for `Peer`-table in future as well!).

jm-clius (approved):
LGTM. Thanks - this will be very useful!

I've left some comments on how migration is/isn't affected by configuration. Specifically, the migration scripts should only run for tables that were already created earlier. Otherwise, you may end up creating tables that your node is not configured to use - unless I've missed this check somewhere?

waku/v2/node/storage/message/waku_message_store.nim
staheri14: > Since the SQLite tables should only be created if the node is configured to do so (i.e. if persistMessages or persistPeers are enabled), the migration would have to be aware of these config items to decide whether or not the CREATE TABLE should be run.

In the current implementation, the migration is done only if the waku node is a full store node and is instructed to persist messages (if the configuration indicates so). It means the call to the migrate proc is already tied to the `persistMessages` (or `persistPeers`) flags. No migration will take place (hence no table will be created) if these flags are not true.

staheri14: >If DB exists, node migrates DB to desired version (with each script only running for tables that already exists)

I understand your point.  However, this approach of "running scripts only when the corresponding table exists", is subject to error e.g., when the table is supposed to exist but is externally deleted from the DB. 

Right now we can perfectly distinguish between peer store migration flow and message store migration part. For this sake, the `migrate` procedure takes a path to the directory containing the migration scripts and starts running them.  The migration scripts of the message store and the peer store are currently maintained in two separate folders exactly for this purpose. As such, the migration procedure would be like this
- Check `persistMessages`, if set then call `migrate(<path to the migration scripts of the message store>)`
- Check `persistPeers `, if set then call `migrate(<path to the migration scripts of the peer store>)`

Please let me know whether this addresses your concern.

tests/v2/test_message_store.nim
staheri14: right, thanks!

waku/v2/node/storage/message/waku_message_store.nim
staheri14: Sure, done!

waku/v2/node/storage/migration/migrations_scripts/message/00002_addSenderTimeStamp.up.sql
staheri14: This is only a temp table used to copy the content of the old Message table to the new one, this temp table gets deleted at the end of this script. Please see line 27
`DROP TABLE Message_backup;`

staheri14: Thanks for identifying the issue! you are right. I have changed the code structure now, the migration is only run when the `persistMessages` is enabled (there is no migration call for the peer store at the moment). 
I am working on updating the SQLite migration scripts in a way that they only get run when the corresponding table exists, however, SQLite is a bit tricky about conditional statement execution and it is more involved than `IF EXISTS Message` It does not allow conditioning the creation of one table to the existence of another table within the same script. Will address it in the following PRs.

waku/v2/node/storage/sqlite.nim
staheri14: Done!

waku/v2/node/storage/message/waku_message_store.nim
staheri14: Makes sense, done!

waku/v2/node/wakunode2.nim
staheri14: Right, I addressed this issue, now the migration is only run when the `persistMessages` is enabled.

waku/v2/node/storage/message/waku_message_store.nim
jm-clius: Yeah, thanks, this works for me.

waku/v2/node/storage/message/waku_message_store.nim
staheri14: Sure, good idea! I am working on a tutorial and will add it in the next PR.

waku/v2/node/storage/message/waku_message_store.nim
staheri14: Nice idea, will investigate it!

staheri14:
A couple of follow-up issues:
Rollback failed migration transactions: https://github.com/status-im/nim-waku/issues/630
DB Migration tutorial: https://github.com/status-im/nim-waku/issues/631

PR #609, opened 2021-06-10
Use chronosStrictException for Waku v1 code

kdeme:
And accompanying changes and clean-up

### DIFF
--- a/waku/common/wakubridge.nim
+++ b/waku/common/wakubridge.nim
@@ -63,7 +63,7 @@ func toWakuMessage(env: Envelope): WakuMessage =
               contentTopic: ContentTopic(string.fromBytes(env.topic)),
               version: 1)
 
-proc toWakuV2(bridge: WakuBridge, env: Envelope) {.async, raises: [Defect].} =
+proc toWakuV2(bridge: WakuBridge, env: Envelope) {.async.} =
   let msg = env.toWakuMessage()
 
   if bridge.seen.containsOrAdd(msg.encode().buffer.hash()):
--- a/waku/common/wakubridge.nim
+++ b/waku/common/wakubridge.nim
@@ -63,7 +63,7 @@ func toWakuMessage(env: Envelope): WakuMessage =
               contentTopic: ContentTopic(string.fromBytes(env.topic)),
               version: 1)
 
-proc toWakuV2(bridge: WakuBridge, env: Envelope) {.async.} =
+proc toWakuV2(bridge: WakuBridge, env: Envelope) {.async, raises: [Defect].} =
   let msg = env.toWakuMessage()
 
   if bridge.seen.containsOrAdd(msg.encode().buffer.hash()):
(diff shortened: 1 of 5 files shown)


waku/common/wakubridge.nim
```diff
@@ -63,7 +63,7 @@ func toWakuMessage(env: Envelope): WakuMessage =
               contentTopic: ContentTopic(string.fromBytes(env.topic)),
               version: 1)
 
-proc toWakuV2(bridge: WakuBridge, env: Envelope) {.async.} =
+proc toWakuV2(bridge: WakuBridge, env: Envelope) {.async, raises: [Defect].} =
```
arnetheduck, line 66: `async` should never have explicit raises (for now)

kdeme: Why is that?

I'm basically trying to make clear that this call should never raise as I want to use with `asyncSpawn`.

arnetheduck: that should get deduced automatically - ie chronos puts the transformed code in a try/except and the `proc` that gets the raises annotation you wrote is just a thin wrapper - it's sometimes useful for debugging, but later, the intention would be that such explicit annotations are actually checked by the transformation which may or may not work (we'll see)

arnetheduck: I've thought about a special `asyncTask` annotation to be used specifically with `asyncSpawn` which would ensure that _your_ code doesn't leak exceptions.

kdeme: But if I understand it correctly, currently `asyncSpawn` will raise `FutureDefect` when any exception would occur.

To avoid this I add the `raises: [Defect]` here. Will this not work because the `raises: [Defect]` will not yet work well with `async`? I'm still not fully understanding why you suggest to remove that, unless it just doesn't work.

I believe ideally there would be a compile error when the proc can raise and used with `asyncSpawn`, so that you don't need these explicit raises. An `asyncTask` could be used for making it "not raising", by try/excepting all and perhaps logging errors (although the name doesn't clarify this intent). 

I'm assuming here that the raising of `FutureDefect` on exceptions is there as it is currently the best we can do. But it can't be the end goal, right?

arnetheduck:
for `proc someAsyncProc() {.async.}` the transformed code looks like (there's more to it, in particular an iterator, but it doesn't matter for the sake of this explanation):

```
proc someAsyncProc(): Future[void] {.whatever_annotations_you_put_go_here.} =
  var res = future[void]
  proc generated() =
    try:
      your_code_goes_here
    except CatchableError as exc:
      res.fail(exc)
   bigGlobalScheduler.callSoon(generated)
   return res
```

basically, when you annotate a function with raises, the raises apply to the wrapper only - not to your code. `asyncSpawn` doesn't care about the wrapper, it cares about the `exc` that ends up in the future.

the underlying reason it happens this way is that instead of generating a _new_ function to use as a wrapper, chronos takes the function you write and strips out the body, which it places in an iterator, which together with the scheduler provides the "resumable" function machinery.

waku/common/wakubridge.nim
arnetheduck: for `proc someAsyncProc() {.async.}` the transformed code looks like (there's more to it, in particular an iterator, but it doesn't matter for the sake of this explanation):

```
proc someAsyncProc(): Future[void] {.whatever_annotations_you_put_go_here.} =
  var res = future[void]
  proc generated() =
    try:
      your_code_goes_here
    except CatchableError as exc:
      res.fail(exc)
   bigGlobalScheduler.callSoon(generated)
   return res
```

basically, when you annotate a function with raises, the raises apply to the wrapper only - not to your code. `asyncSpawn` doesn't care about the wrapper, it cares about the `exc` that ends up in the future.

the underlying reason it happens this way is that instead of generating a _new_ function to use as a wrapper, chronos takes the function you write and strips out the body, which it places in an iterator, which together with the scheduler provides the "resumable" function machinery.

kdeme: Right, that makes sense now, thanks.

PR #610, opened 2021-06-10
Fix: metrics-server use after submodule update

jm-clius:
Fix required by changes in `metrics` submodule.

This is necessary where compiling with `-d:insecure` Nim flag, such as during deployment to fleet nodes.

For this reason the `test` deployment for Waku v1 has been [failing](https://ci.status.im/job/nim-waku/job/deploy-v1-test/162/)

### DIFF
--- a/waku/common/wakubridge.nim
+++ b/waku/common/wakubridge.nim
@@ -1,6 +1,7 @@
 import
   std/[tables, hashes, sequtils],
-  chronos, confutils, chronicles, chronicles/topics_registry, metrics,
+  chronos, confutils, chronicles, chronicles/topics_registry, 
+  metrics, metrics/chronos_httpserver,
   stew/[byteutils, objects],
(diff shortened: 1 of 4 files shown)


waku/common/wakubridge.nim
```diff
@@ -290,6 +291,6 @@ when isMainModule:
         address = conf.metricsServerAddress
         port = conf.metricsServerPort + conf.portsShift
       info "Starting metrics HTTP server", address, port
-      metrics.startHttpServer($address, Port(port))
+      startMetricsHttpServer($address, Port(port))
```
kdeme, line 294: You can remove the `when defined(insecure):` compile time option everywhere now that we can use the chronos http server in metrics.

jm-clius: Right! Thanks.

EbubeUd (approved):

kdeme (approved):

PR #621, opened 2021-06-14
Add ping protocol support and use for keep-alive

### LINKED ISSUE #534: Mount `/ipfs/ping/1.0.0` protocol in wakunode2
To avoid some errors over websocket connections (see #123), js-waku needs to implement a keep alive.
The most appropriate protocol for this would be the ping one.

Could you mount it by default? I understand it is not currently the case.

jm-clius:
This PR closes #534 

It adds libp2p `ping` protocol support for Waku v2, mounts it on all nodes and uses `ping` as the underlying mechanism for keep-alive functionality.

In order to use the `unstable` branch of `nim-libp2p`, the `WakuPeerStore` had to be separated from the underlying libp2p `PeerStore`. See [this comment](https://github.com/status-im/nim-libp2p/commit/93156447ba99fa9f70a29c885c5a1509da3969b0#r52132832) for background. https://github.com/status-im/nim-waku/issues/622 keeps track of the effort to reintegrate the libp2p `PeerStore` into Waku v2.


staheri14 (approved):
LGTM!

oskarth (approved):
Should we mention this as a MAY in specs perhaps? For topics/connections that don't see a lot of traffic

jm-clius:
> Should we mention this as a MAY in specs perhaps? For topics/connections that don't see a lot of traffic

Good idea. See https://github.com/vacp2p/rfc/pull/400

PR #624, opened 2021-06-14
Gets rln lib path at the compile time

staheri14:
The path to the rln lib has been initially set relative to the nim-waku root directory. However, this path does not get recognized by the Nim compiler when nim-waku is imported as a submodule. This PR is to resolve this problem.
As the solution, the `currentSourcePath()` template is used to get the full file-system path of the `rln.nim` module at the compilation time, and then the `rln` lib path is set relative to that.

### DIFF
--- a/waku/v2/protocol/waku_rln_relay/rln.nim
+++ b/waku/v2/protocol/waku_rln_relay/rln.nim
@@ -1,10 +1,13 @@
 # this module contains the Nim wrappers for the rln library https://github.com/kilic/rln/blob/3bbec368a4adc68cd5f9bfae80b17e1bbb4ef373/src/ffi.rs
 
 import
-  os,
+  std/[os,strutils],
   waku_rln_relay_types
 
-const libPath = "vendor/rln/target/debug/"
+
+template sourceDir: string = currentSourcePath.rsplit(DirSep, 1)[0]
+
+const libPath = sourceDir / "../../../../vendor/rln/target/debug/"
 when defined(Windows):
   const libName* = libPath / "rln.dll"
 elif defined(Linux):


michaelsbradleyjr:
Commonly, the name of a shared library is specified without any directory prefixing it. A couple of examples:

https://github.com/GULPF/tiny_sqlite/blob/master/src/tiny_sqlite/sqlite_wrapper.nim#L1-L13

https://github.com/nim-lang/Nim/blob/8ccde68f132be4dba330eb6ec50f4679e564efac/lib/wrappers/pcre.nim#L313-L327

Both examples deal with some complexity (e.g. `(|.0)`) that you don't need to worry about re: rln.

An advantage to this approach is that it's less fragile, e.g. if the directory structure of nim-waku were to change then `sourceDir / "../../../../vendor/rln/target/debug/"` might become invalid. It also avoids hardcoding for a debug build of rln.

Also, it enforces less assumptions about how a user of nim-waku will build and supply the rln library, generally speaking.

But...

As a consequence of supplying only the library name, the C compiler/linker must be able to find the library in "the usual places", which differ by platform.

If it can't be found in the usual places, then it's up to the person invoking the Nim compiler to make it findable.

If you use macOS and `brew` to install open source libraries, you sometimes see this kind of thing after `brew install [bottle]` or in the output of `brew info [bottle]`.

```
$ brew info sqlite

...

sqlite is keg-only, which means it was not symlinked into /usr/local,
because macOS already provides this software and installing another version in
parallel can cause all kinds of trouble.

...

For compilers to find sqlite you may need to set:
  export LDFLAGS="-L/usr/local/opt/sqlite/lib"
  export CPPFLAGS="-I/usr/local/opt/sqlite/include"

...
```

So setting or adding to those environment variables is one way.

Another would be to add these options when invoking the Nim compiler (same example context of sqlite installed with `brew` on macOS):

```
--passC:"-I/usr/local/opt/sqlite/include" --passL:"-L/usr/local/opt/sqlite/lib"
```

There's another variation if you want to bypass the dynlib pragma in the Nim sources and link statically.

```
--dynlibOverride:sqlite3 --passC:"-I/usr/local/opt/sqlite/include" --passL:"-L/usr/local/opt/sqlite/lib/libsqlite3.a"
```

Note that in many cases the `--passC:"..."` isn't necessary, only `--passL:"..."`, but I included it for parity with `CPPFLAGS` (corresponds to `--passC`) and `LDFLAGS` (corresponds to `--passL`) in the example `brew info` output above. 

Since the build of rln is in nim-waku's `vendor/`, then it's in this kind of situation where env variables or Nim compiler options would be needed so clang/gcc will be able to find the compiled rln lib.

That's something that could be handled in nim-waku's Makefile or `waku.nimble`.

In the Makefile for nim-status, I wire things up for rln like this:

https://github.com/status-im/nim-status/blob/feat/examples/client/Makefile#L134-L168

https://github.com/status-im/nim-status/blob/feat/examples/client/Makefile#L349-L384

Currently, though, in nim-status we don't invoke the Nim compiler directly in the Makefile, and instead use a nimble task; so for that I set `RLN_LDFLAGS` and `RLN_STATIC` as environment variables when invoking nimble:

https://github.com/status-im/nim-status/blob/feat/examples/client/Makefile#L443-L444

And then in the nimble task I setup `----dynlibOverride` and/or `--passL` for the rln lib:

https://github.com/status-im/nim-status/blob/feat/examples/client/nim_status.nimble#L56-L62

https://github.com/status-im/nim-status/blob/feat/examples/client/nim_status.nimble#L69

It's not the prettiest setup, by any means, and one day I'm going to clean up some of the redundancy and/or learn better ways to go about these things, but it does work.

If you go the Makefile route, keep in mind that since rln is now an optional part of nim-waku you'd probably want a make variable such as `RLN_ENABLED ?= false`, which could then be toggled on the command-line with `make RLN_ENABLED=true ...`. And you'd need to implement some Makefile logic around whether it's true or false.

---

I'm not an expert, and everything I wrote above is based on things learned since I started working with Nim as part of the Status Desktop project. What I've figured out that "does work" is maybe not the best way to do things.

For that reason I'd like to loop in e.g. @arnetheduck and @zah because maybe they can offer some insight and advice for avoiding unnecessary complexity.

arnetheduck:
> Commonly, the name of a shared library is specified without any directory prefixing it.

not quite - when a shared library is built as part of the project (as in - it's not a a system library) it's common to specify a full path at compile time and add an `RPATH` entry for runtime using a relative `$ORIGIN` path - the smarter thing to do _usually_ however is to use a statically linked library to avoid the runtime mess - ie a shared library is usually only used when it's a "must" due to constraints. rust can create static libraries with the right incantation so I'd go with that unless there were specific reasons.

Examples of this kind of build would be https://github.com/status-im/nim-nat-traversal/ .

> desktop

I think desktop shot itself in the foot a bit by depending on a library with weird requirements and build constraints - `sqlcipher` and `openssl` are two libraries that have excessively complex builds, and this complexity spilled over to desktop - this is in large part the reason we don't use them.

arnetheduck:
> https://github.com/GULPF/tiny_sqlite/blob/master/src/tiny_sqlite/sqlite_wrapper.nim#L1-L13

we don't use this one either, for similar reasons - it's _much_ easier to create a reliable reproducible build if you avoid DLL:s completely.

arnetheduck:
> https://github.com/nim-lang/Nim/blob/8ccde68f132be4dba330eb6ec50f4679e564efac/lib/wrappers/pcre.nim#L313-L327

neither do we use this one :)

The way nim imports DLL:s in general, by loosely matching multiple dll names, is deeply problematic - upstream considers it a feature but like many things in nim, it's an 80% solution that works for the simple case but fails catastrophically for corner cases, creating very hard-to-debug issues on top of already complex problems.

michaelsbradleyjr:
@arnetheduck thanks for the feedback.

It actually makes me feel better about having completely stripped use of the dynlib pragma from the [copied `sqlite_wrapper.nim`](https://github.com/status-im/nim-sqlcipher/blob/master/sqlcipher/sqlite_wrapper.nim) that stimbus team embedded in nim-sqlcipher, as opposed to having changed it to use `libsqlcipher` instead of `libsqlite`.

So then it's a matter of supplying e.g. `--passL` to the Nim compiler? Or what do recommend as the best practice re: wiring up some defaults for linking when the compiled library isn't in a platform's usual place for `.dylib`, `.a`, etc.

michaelsbradleyjr:
Whoops, I scrolled down too fast and missed @arnetheduck's first comment, digesting it now.

michaelsbradleyjr:
> the smarter thing to do usually however is to use a statically linked library to avoid the runtime mess

I prefer (have learned to prefer) static linking whenever possible and in the dogfooding TUI client that's being built for nim-status I've gone to lengths to make sure that the build can be completely static on all platforms (or as static as possible), including statically linking rln.

> rust can create static libraries with the right incantation so I'd go with that unless there were specific reasons

Yes, and I created an [fork](https://github.com/michaelsbradleyjr/rln/tree/staticlib) of rln to make that possible for rln. in nim-status' `vendor/`, I'm pointing at a branch of nim-waku that uses my fork of rln since my PR to the upstream has gone ignored: https://github.com/kilic/rln/pull/1.

staheri14:
Thanks, @michaelsbradleyjr, and @arnetheduck for sharing your valuable insights and practical examples! 🙏

staheri14:
My current understanding is that it is better to link RLN lib statically, there is no reason for it to be linked dynamically. 
So, given it will be linked statically, do you see any issue  (rather than the one you mentioned in your comments about i.e., the future changes in the nim-waku directory) with prefixing the lib with its path, like what I did in this PR? @michaelsbradleyjr

michaelsbradleyjr:
> My current understanding is that it is better to link RLN lib statically, there is no reason for it to be linked dynamically.

That's a valid understanding, I think.

> So, given it will be linked statically, do you see any issue (rather than the one you mentioned in your comments about the future changes in the nim-waku directory) with prefixing the lib with its path, like what I did in this PR? @michaelsbradleyjr

So if you stick with combining:
```nim
const libPath = sourceDir / "../../../../vendor/rln/target/debug/"
const libName = ...

{.push dynlib: libName, raises: [Defect].}
```
There are a couple of less than ideal effects.
* It's hardwired for a debug build per `target/debug`.
* A project consuming nim-waku with rln enabled will generally need to pass <br/> `--dynlibOverride:vendor/rln/target/debug/librln` to the Nim compiler, whether shared or static linking is preferred, and regardless of debug or release build, because the path is only valid if you're compiling waku's Nim sources from the root of the nim-waku repo.

I think there are three better (mutually exclusive) options:

---
**Unprefixed `libName`**

```nim
when defined(Windows):
  const libName* = "rln.dll"
elif defined(Linux):
  const libName* = "librln.so"
elif defined(MacOsX):
  const libName* = "librln.dylib"

# all the following procedures are Nim wrappers for the functions defined in libName
{.push dynlib: libName, raises: [Defect].}
```

Upsides: no hardwired path or build type, consumer of nim-waku can build rln where/how they wish, maybe inside `nim-waku/vendor/rln` or maybe somewhere else that's more convenient.

Downsides: see https://github.com/status-im/nim-waku/pull/624#issuecomment-861762010.
> The way nim imports DLLs in general, by loosely matching multiple dll names, is deeply problematic - upstream considers it a feature but like many things in nim, it's an 80% solution that works for the simple case but fails catastrophically for corner cases, creating very hard-to-debug issues on top of already complex problems.

It's probably not going to be a problem for waku+rln, because it's a simple case, but those observations are worth keeping in mind.

Also, any project that wants to enable rln, including waku's own tests, examples, etc., would need to use e.g. <br/>`--passL:"-L/path/to/built/rln -lrln"` or `--passL:"/path/to/built/librln.a"` so that the Nim compiler can tell the C compiler where to find it.

And there are some other things you need to do at runtime if you don't use static linking; not difficult, can be explained later.

---
**Don't use `{.push dynlib: libName}`**

It's not strictly necessary.

Upsides: same as previous option.

Downsides: same as the part above that starts with *"Also, any project that wants to enable rln, including waku's own tests..."*.

---
**Setup `{.passL: ...}` in waku's Nim sources**

Example: https://github.com/status-im/nim-nat-traversal/blob/master/nat_traversal/miniupnpc.nim#L27.

Upsides: although the waku consumer needs to ensure rln is built before compiling the Nim code (which is also true for the previous two options), they don't need to inform the Nim compiler where it's located.

Downsides: rln would always be the one inside nim-waku's `vendor/rln`. In the waku Nim sources there needs to be a decision whether to hardwire the path to a `rln/target/release` build or a `rln/target/debug` build, or parameterizing somehow, e.g. on `when defined(release)`.

Also, it would make sense to go ahead and use static linking, but https://github.com/kilic/rln/pull/1 would need to be merged and that change added to rln's `full-node` branch.

arnetheduck:
> Also, it would make sense to go ahead and use static linking, but kilic/rln#1 would need to be merged and that change added to rln's full-node branch.

we tend to branch things under `status-im` in cases like this - ie once we need to patch something, it's often easier to keep a fork around anyway

arnetheduck:
>  needs to ensure rln is built before compiling the Nim code

[staticExec](https://nim-lang.org/docs/system.html#staticExec%2Cstring%2Cstring%2Cstring) can maybe be used constructively here

oskarth:
@staheri14 What's the state of this PR? If we aren't working on it, perhaps we can close it?

staheri14:
> @staheri14 What's the state of this PR? If we aren't working on it, perhaps we can close it?

@oskarth Unfortunately, some other high-priority tasks came up and I could not spend enough time to actively work on the PR. 
Thanks @michaelsbradleyjr and @arnetheduck for your insights and help!
In order to keep the vac board clean and update, going to close the PR but I will continue the discussion on the closed PR

PR #627, opened 2021-06-15
Add Issuer to Cheque Object 

### LINKED ISSUE #626: Refactor the Swap Cheque object
A follow up to #620 

To achieve Proper signature verification, the cheque object needs to be refactored a bit to contain the public key of the issuer. 
This will be used to verify the signature by the recipient of the cheque.

EbubeUd:
closes #626

`Issuer` has been added to the cheque object. The essence Of this is to enable the signatures to be verified by the recipient using the Issuer's public key.

### DIFF
--- a/waku/v2/protocol/waku_swap/waku_swap.nim
+++ b/waku/v2/protocol/waku_swap/waku_swap.nim
@@ -148,12 +148,13 @@ proc handleCheque*(ws: WakuSwap, cheque: Cheque) =
   let peerOpt = ws.peerManager.selectPeer(WakuSwapCodec)
   let peerId = peerOpt.get().peerId
 
-    # Get the original signer using web3. Check if web3.eth.personal.ecRecover(messageHash, signature); or an equivalent function has been implemented in nim-web3
+  # Get the original signer using web3. For notw, a static value is used.
+  # Check if web3.eth.personal.ecRecover(messageHash, signature); or an equivalent function has been implemented in nim-web3
   let signer = @[byte 0, 1, 2]
 
+  # Verify that the Issuer was the signer of the signature
   if signer != cheque.issuer:
     warn "Invalid cheque: The address of the issuer is different from the signer."
-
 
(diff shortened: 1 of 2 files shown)


jm-clius (commented):
I will review again once the `issuer` field is used.

oskarth (commented):
Same, minimal PRs are good but ideally they are minimal useful. So you might have a test case, or somewhere where issue field is used.

waku/v2/protocol/waku_swap/waku_swap.nim
```diff
@@ -146,6 +147,14 @@ proc handleCheque*(ws: WakuSwap, cheque: Cheque) =
   let peerOpt = ws.peerManager.selectPeer(WakuSwapCodec)
   let peerId = peerOpt.get().peerId
 
+  # Get the original signer using web3. For now, a static value (0x6C3d502f1a97d4470b881015b83D9Dd1062172e1) will be used.
+  # Check if web3.eth.personal.ecRecover(messageHash, signature); or an equivalent function has been implemented in nim-web3
```
oskarth, line 151: You need something like https://github.com/vacp2p/swap-contracts-module/blob/master/src/swap-helpers.js#L132 don't think it is nim-web3

oskarth (approved):
Seems OK as partial progress, I guess the next thing is to actual verify the signature?

jm-clius (approved):

waku/v2/protocol/waku_swap/waku_swap.nim
EbubeUd: Nice. I'll need to create a PR in the swap Contracts module for adding a `verifySignature` function similar to what was done [here](https://docs.ethers.io/v4/cookbook-signing.html#signing-a-string-message). Essence of this function is to return the public key of the signer. 
For now, I'll assume the returned address is `0x6C3d502f1a97d4470b881015b83D9Dd1062172e1`

PR #628, opened 2021-06-15
Update Swap Account Metrics

### LINKED ISSUE #625: Metric Labels for Swap Account Balance are at high risks of being overpopulated
Currently, the `peer` label in the swap account Metrics keeps track of each peer id with an account balance. 

**Problem**
The problem with this is that there is a very high chance of the peer ids being so numerous that they cause negative effects On the dashboard.

**Solution**
A possible solution to this to change the way the metrics are being accounted for. For now, we can have a counter indicating how many Peers have exceeded the thresholds. (Payment threshold and disconnect threshold).

EbubeUd:
closes #625 

Change `waku_peer_swap_account_balance` from guage to histogram.
Update peer counter variable name to `waku_swap_peers_count`

### DIFF
--- a/waku/v2/protocol/waku_swap/waku_swap.nim
+++ b/waku/v2/protocol/waku_swap/waku_swap.nim
@@ -36,9 +36,11 @@ import
 
 export waku_swap_types
 
-declarePublicGauge waku_swap_peers, "number of swap peers"
+const swapAccountBalanceBuckets = [-Inf, -200.0, -150.0, -100.0, -50.0, 0.0, 50.0, 100.0, 150.0, 200.0, Inf]
+
+declarePublicGauge waku_swap_peers_count, "number of swap peers"
 declarePublicGauge waku_swap_errors, "number of swap protocol errors", ["type"]
-declarePublicGauge waku_swap_account_state, "swap account state for each peer", ["peer"]
+declarePublicHistogram waku_peer_swap_account_balance, "Swap Account Balance for waku peers, aggregated into buckets based on threshold limits", buckets = swapAccountBalanceBuckets
 
 logScope:
   topics = "wakuswap"
@@ -194,7 +196,7 @@ proc handleCheque*(ws: WakuSwap, cheque: Cheque) =
 
 # Log Account Metrics
 proc logAccountMetrics*(ws: Wakuswap, peer: PeerId) {.async.}=
-  waku_swap_account_state.set(ws.accounting[peer].int64, labelValues = [$peer])
+  waku_peer_swap_account_balance.observe(ws.accounting[peer].int64)
 
 
 proc init*(wakuSwap: WakuSwap) =
@@ -270,6 +272,6 @@ proc init*(T: type WakuSwap, peerManager: PeerManager, rng: ref BrHmacDrbgContex
 
 proc setPeer*(ws: WakuSwap, peer: PeerInfo) =
   ws.peerManager.addPeer(peer, WakuSwapCodec)
-  waku_swap_peers.inc()
+  waku_swap_peers_count.inc()
 
 # TODO End to end communication


oskarth (approved):
Nice! Does it work?

jm-clius (approved):
Curious how this is presented on the Grafana dashboard - perhaps you can add screenshot?

EbubeUd:
image


Yea it does but the thing is... The metrics might be a Little difficult to keep track of. 
So from what I read up on histograms, each bucket gets incremented if the value observed  is less than or equal to the bucket value.

Here, I created 5 buckets. [-100, -50, 0, 50, 100] so that I can easily keep track of the number of peers with a balance swap under each range. Now I ran 6 nodes. 1 store node and 5 chat clients  and sent in a coupe of messages in each of them. But the result Is a bit off the target.

Here, it displays `6` in each bucket, but 0 in -`100` and -`50`. Reason being that none of the swap balance observed was less than `-50` or `-100`. But this Is going to be a bit confusing in the long run. In the sense that `50` and `100` will always display the total number Of peer balances observed. 

A solution will be to have a way of applying ranges to the buckets. Such that the buckets are only incremented if the observed balance is less than an upper limit and greater than a lower limit.

oskarth:
That view is a bit confusing indeed. It seems like something that can be addressed at view level (Grafana). Did you have a look at https://grafana.com/blog/2020/06/23/how-to-visualize-prometheus-histograms-in-grafana/? Also cc @jakubgs in case you have any specific recommendations here

EbubeUd:
> That view is a bit confusing indeed. It seems like something that can be addressed at view level (Grafana). Did you have a look at https://grafana.com/blog/2020/06/23/how-to-visualize-prometheus-histograms-in-grafana/? Also cc @jakubgs in case you have any specific recommendations here

Yea it can be resolved within grafana dashboard. I had a look at the link you sent and other sources as well. I'll be able to get It sorted out in Grafana Dashboard. Simply subtracting the value in the lower bucket from that of the higher bucket will give me the number of items in the higher bucket. Same continues for all buckets till I get to the least bucket.

jakubgs:
There's many ways to visualize a histogram, but I usually just use a normal graph, like this:

image

But there are other ways:
https://grafana.com/blog/2020/06/23/how-to-visualize-prometheus-histograms-in-grafana/

Also, would you not want more higher value buckets? Like -150, +150, -200, +200, or are those values unrealistic or irrelevant?

oskarth:
> Also, would you not want more higher value buckets? Like -150, +150, -200, +200, or are those values unrealistic or irrelevant?

Probably a good idea, the above was mostly to capture most immediate threshholds for policy, but more is useful too

jakubgs:
If you expect large values as well there's often a bucket called `Inf` or `+Inf` and `-Inf`, which represent values bigger or smaller than what can fit in all other buckets. Not sure how that's implemented with this library, but I've seen metrics like that often.

jakubgs:
Example:
```
admin@master-01.do-ams3.metrics.hq:~ % curl -s 'localhost:9092/metrics' | grep cortex_labels_per_sample_bucket
cortex_labels_per_sample_bucket{le="5"} 6.1372163e+07
cortex_labels_per_sample_bucket{le="10"} 7.004786566e+09
cortex_labels_per_sample_bucket{le="15"} 7.047679807e+09
cortex_labels_per_sample_bucket{le="20"} 7.077835361e+09
cortex_labels_per_sample_bucket{le="25"} 7.077835361e+09
cortex_labels_per_sample_bucket{le="+Inf"} 7.077835361e+09
```

EbubeUd:
Screenshot 2021-06-17 at 11 34 49

Looks a lot better now. 
The blue label shows the balances that are within the range of `-50 and 0 ` while the yellowish label shows the balances that are between `-100 and -50`. It'll be easier to understand the readings from the graph subsequently.

EbubeUd:
> Screenshot 2021-06-17 at 11 34 49
> 
> Looks a lot better now.
> The blue label shows the balances that are within the range of `-50 and 0 ` while the yellowish label shows the balances that are between `-100 and -50`. It'll be easier to understand the readings from the graph subsequently.

BTW, the gap between the left and right lines in the metrics was as a result of Closing the connection in all the nodes and restarting them.

oskarth:
Nice! Needs rebase and then I think we can merge it + update the dashboard in Grafana cluster. Can always adjust bucket sizes (now if you want or later would work too).

jakubgs:
@EbubeUd I think you should be using `git rebase master`, not creating merge commits to master:
image

EbubeUd:
> @EbubeUd I think you should be using `git rebase master`, not creating merge commits to master:
> 
> image
> 
> You should drop those merge commits and then rebase.

Ohh Alright. Thanks For pointing out

PR #633, opened 2021-06-17
Feat/bridge topic mapping

### LINKED ISSUE #576: Improvement: Waku v1 <> v2 bridge topic mapping
## Problem

The Waku v1 <> b2 bridge maps Waku v1 topics (4 byte arrays) to Waku v2 content topics (strings). It currently uses a simple scheme where any v2 content topic is converted to bytes and the first four bytes are considered the v1 topic. This scheme:
- may lead to content topic clashes,
- does not implement [namespacing](https://github.com/vacp2p/rfc/issues/227) or the [recommended use of topics](https://github.com/vacp2p/rfc/pull/383), and
- is not reversible (you cannot reliably convert a v1 topic back to the original content topic).

## Solution

Implement a bridge topic mapping according to the namespaced scheme:

```
/waku/v1/<4bytes-waku-v1-topic>/rlp
```

This creates a direct mapping between the v1 and v2 protocols.
For example:

```
/waku/v1/0x007f80ff/rlp
```

jm-clius:
This PR closes #576

It introduces a namespaced scheme to bridge topic mappings.
More specifically, Waku v1 topics (4-byte arrays) are mapped to Waku v2 content topics (strings) according to the scheme:

```
/waku/1/<4bytes-waku-v1-topic>/rlp
```

where `<4bytes-waku-v1-topic>` is the big-endian hex-representation of the Waku v1 topic

This is done according to the recommended namespaced topic structure set out in https://rfc.vac.dev/spec/23/.

Namespacing utils are also introduced for applications that choose to follow the recommendations in `23/WAKU2-TOPICS`. Note that the ambiguous use of `Topic` in the namespacing utils is to cover use cases for both pubsub topics and content topics. These essentially use the same recommended namespacing scheme, with Waku v2 defining itself as an "application" over PubSub and content topics defining applications over Waku v2.


waku/v2/utils/namespacing.nim
```diff
@@ -0,0 +1,54 @@
...
+                     topicName: parts[3],
+                     encoding: parts[4]))
+
+proc `$`*(namespacedTopic: NamespacedTopic): string =
+  ## Returns a string representation of a namespaced topic
+  ## in the format `/<application>/<version>/<topic-name>/<encoding>`
+  
+  var topicStr = newString(0)
+
+  topicStr.add("/")
```
staheri14, line 45: Just curious why not using `&` in here for string concatenation?

staheri14 (approved):
Nice set of unit tests!
LGTM!

waku/v2/utils/namespacing.nim
jm-clius: Good question! One reason is that using the `add`/append syntax is somewhat more readable for a very structured string, especially where the format may be modified in future.
Another is that `add` modifies the original string (which is the behaviour I want here), whereas `&` creates a new string composed of the two strings being concatenated.

PR #635, opened 2021-06-18
Fix issues with Peer reference in Cheque Functions

### LINKED ISSUE #634: Incorrect peer ID in Handle Cheque and Send Cheque function
Follow up to #620

Currently, the `sendCheque` function [here](https://github.com/status-im/nim-waku/blob/2f53c2c60bf7e2dae86203de712689f8349fd85f/waku/v2/protocol/waku_swap/waku_swap.nim#L102) sends out the cheque to the first connected peer . Which might not actually be the correct peer to receive the cheque. The `handleCheque` function, which gets executed on the part of the recipient, also assumes that the `peer` issuing the cheque is the first connected peer [here](https://github.com/status-im/nim-waku/blob/2f53c2c60bf7e2dae86203de712689f8349fd85f/waku/v2/protocol/waku_swap/waku_swap.nim#L148). This will be wrong in most cases. 

A solution to this is to pass in the `peer id` as a parameter to the `sendCheque` and Handle Cheque functions.

EbubeUd:
closes #634

- Added `PeerId` to SendCheque function (Indicates the Peer id of the recipient of the cheque)
- Added `PeerId` to the HandleCheque Function (Indicates the Peer Id of the cheque issuer)
- Added a function in PeerManager to return the peer info using the `PeerId`


waku/v2/node/peer_manager/peer_manager.nim
```diff
@@ -200,6 +200,16 @@ proc selectPeer*(pm: PeerManager, proto: string): Option[PeerInfo] =
   else:
     return none(PeerInfo)
 
+proc getPeerInfo*(pm: PeerManager, peerId: PeerId) : Option[PeerInfo] =
```
oskarth, line 203: @jm-clius since you designed the peer manager, is there some other preferred way of doing this?

oskarth (approved):
LGTM

waku/v2/node/peer_manager/peer_manager.nim
jm-clius: This `peerInfo` for the `peerId` is already available in the connection handler that triggers the `credit`/`debit` function against this peer in the first place (e.g. [here](https://github.com/status-im/nim-waku/blob/master/waku/v2/protocol/waku_store/waku_store.nim#L363)). In other words, the correct way to do this would be to change the `credit`, `debit`, `sendCheque` functions to use `peerInfo` rather than `peerId`. Unless there's some reason to use `peerId` that I'm missing?

waku/v2/protocol/waku_swap/waku_swap.nim
```diff
@@ -98,8 +98,8 @@ proc init*(T: type Cheque, buffer: seq[byte]): ProtoResult[T] =
 
 
 # TODO Assume we calculated cheque
-proc sendCheque*(ws: WakuSwap) {.async.} =
-  let peerOpt = ws.peerManager.selectPeer(WakuSwapCodec)
+proc sendCheque*(ws: WakuSwap, peerId : PeerId) {.async.} =
```
jm-clius, line 101: See my comment elsewhere on changing this to `peerInfo`. It would require changing the `credit` and `debit` procedures to take `peerInfo` argument as well. `peerInfo` should be available on the connection when calling `credit`/`debit` from `waku_store`. The `getPeerInfo()` procedure in the manager then becomes obsolete.

jm-clius (changes requested):

waku/v2/protocol/waku_swap/waku_swap.nim
EbubeUd: Right. I'll make the necessary changes now. Thanks

waku/v2/node/peer_manager/peer_manager.nim
oskarth: Right, correct

jm-clius (approved):
Nice!

PR #641, opened 2021-06-23
Add tutorial to setup websockify

D4nte:


oskarth (approved):
Thanks! cc @dryajov fyi for libp2p team

dryajov:
Hopefully, this should be short lived as we've implemented the websockets as per https://datatracker.ietf.org/doc/html/rfc6455. We'll work in integrating it with nim-libp2p next.

oskarth:
@dryajov is it accurate that with the native implementation no domain or certificate management is necessary? even for `wss` to browsers?

dryajov:
> @dryajov would the native implementation require any form of domain or certificate management for `wss` to browsers?

Certs are managed by chronos which uses bearssl for it's ssl implementation - bearssl embeds mozilla's certificate store from https://curl.haxx.se/ca/cacert.pem

For DNS, chronos uses `getaddrinfo` (which is blocking, so beware), which should do ipv4,6 and domain name resolution.

PR #643, opened 2021-06-23
Add feature for disconnecting a peer exceeding threshold

EbubeUd:
Before Messages are sent to a peer by the store node, a check is done to see if the peer's balance is below the disconnect threshold. 

closes #642

### DIFF
--- a/tests/v2/test_peer_manager.nim
+++ b/tests/v2/test_peer_manager.nim
@@ -2,8 +2,8 @@
 
 import
   std/[options, sets, tables, sequtils],
-  testutils/unittests, stew/shims/net as stewNet,
-  json_rpc/[rpcserver, rpcclient],
+  testutils/unittests, stew/shims/net as stewNet, 
+  json_rpc/[rpcserver, rpcclient], chronos, chronicles,
   eth/[keys, rlp], eth/common/eth_types,
   libp2p/[builders, switch, multiaddress],
   libp2p/protobuf/minprotobuf,
@@ -209,3 +209,98 @@ procSuite "Peer Manager":
       node3.peerManager.connectedness(peerInfo2.peerId) == Connected
     
     await allFutures([node1.stop(), node2.stop(), node3.stop()])
+
+
+  asyncTest "Peer dialing fails when peer is blacklisted":
+      let
+        nodeKey1 = crypto.PrivateKey.random(Secp256k1, rng[])[]
+        node1 = WakuNode.new(nodeKey1, ValidIpAddress.init("0.0.0.0"),
+          Port(60000))
+        nodeKey2 = crypto.PrivateKey.random(Secp256k1, rng[])[]
+        node2 = WakuNode.new(nodeKey2, ValidIpAddress.init("0.0.0.0"),
+          Port(60002))
+        peerInfo2 = node2.peerInfo
+      
+      await allFutures([node1.start(), node2.start()])
+
+      node1.mountRelay()
+      node2.mountRelay()
+      node2.peerManager.addToBlacklist(node1.peerInfo.peerId)
+
+      # Dial node2 from node1
+      let conn = (await node1.peerManager.dialPeer(peerInfo2, WakuRelayCodec)).get()
+
+      #check blacklist
+      check:
+        node1.peerInfo.peerId in node2.peerManager.blacklist == true
+
+      # Check connection
+      check:
+        conn.activity
+        conn.peerInfo.peerId == peerInfo2.peerId
+      
+      # Check that node2 is being managed in node1
+      check:
+        node1.peerManager.peers().anyIt(it.peerId == peerInfo2.peerId)
+
+      # Check connectedness
+      check:
+        node2.peerManager.connectedness(node1.peerInfo.peerId) == Connectedness.ShouldNotConnect
+      
+      await allFutures([node1.stop(), node2.stop()])
+
+
+
+  asyncTest "Test Adding Peer to Blacklist":
+      let
+        nodeKey1 = crypto.PrivateKey.random(Secp256k1, rng[])[]
+        node1 = WakuNode.new(nodeKey1, ValidIpAddress.init("0.0.0.0"),
+          Port(60000))
+        nodeKey2 = crypto.PrivateKey.random(Secp256k1, rng[])[]
+        node2 = WakuNode.new(nodeKey2, ValidIpAddress.init("0.0.0.0"),
+          Port(60002))
+        peerInfo2 = node2.peerInfo
+      
+      await allFutures([node1.start(), node2.start()])
+
+      node1.mountRelay()
+      node2.mountRelay()
+      node2.peerManager.addToBlacklist(node1.peerInfo.peerId)
+
+      # Dial node2 from node1
+      let conn = (await node1.peerManager.dialPeer(peerInfo2, WakuRelayCodec)).get()
+
+      #check blacklist
+      check:
+        node1.peerInfo.peerId in node2.peerManager.blacklist == true
+
+      await allFutures([node1.stop(), node2.stop()])
+
+
+  asyncTest "Test Removing Peer from Blacklist":
+      let
+        nodeKey1 = crypto.PrivateKey.random(Secp256k1, rng[])[]
+        node1 = WakuNode.new(nodeKey1, ValidIpAddress.init("0.0.0.0"),
+          Port(60000))
+        nodeKey2 = crypto.PrivateKey.random(Secp256k1, rng[])[]
+        node2 = WakuNode.new(nodeKey2, ValidIpAddress.init("0.0.0.0"),
+          Port(60002))
+        peerInfo2 = node2.peerInfo
+      
+      await allFutures([node1.start(), node2.start()])
+
+      node1.mountRelay()
+      node2.mountRelay()
+      node2.peerManager.addToBlacklist(node1.peerInfo.peerId)
+      node2.peerManager.removeFromBlacklist(node1.peerInfo.peerId)
+
+      # Dial node2 from node1
+      let conn = (await node1.peerManager.dialPeer(peerInfo2, WakuRelayCodec)).get()
+
+      #check blacklist
+      check:
+        node1.peerInfo.peerId in node2.peerManager.blacklist == false
+
+      await allFutures([node1.stop(), node2.stop()])
+
+   
--- a/tests/v2/test_peer_manager.nim
+++ b/tests/v2/test_peer_manager.nim
@@ -228,20 +228,15 @@ procSuite "Peer Manager":
       node2.peerManager.addToBlacklist(node1.peerInfo.peerId)
 
       # Dial node2 from node1
-      let conn = (await node1.peerManager.dialPeer(peerInfo2, WakuRelayCodec)).get()
+      let conn = (await node2.peerManager.dialPeer(node1.peerInfo, WakuRelayCodec))
 
       #check blacklist
       check:
         node1.peerInfo.peerId in node2.peerManager.blacklist == true
 
-      # Check connection
-      check:
-        conn.activity
-        conn.peerInfo.peerId == peerInfo2.peerId
-      
       # Check that node2 is being managed in node1
       check:
-        node1.peerManager.peers().anyIt(it.peerId == peerInfo2.peerId)
+        node2.peerManager.peers().anyIt(it.peerId == node1.peerInfo.peerId)
 
       # Check connectedness
       check:
--- a/tests/v2/test_peer_manager.nim
+++ b/tests/v2/test_peer_manager.nim
@@ -303,4 +303,31 @@ procSuite "Peer Manager":
 
       await allFutures([node1.stop(), node2.stop()])
 
-   
+  asyncTest "Test Disconnecting Peer":
+    let
+      nodeKey1 = crypto.PrivateKey.random(Secp256k1, rng[])[]
+      node1 = WakuNode.new(nodeKey1, ValidIpAddress.init("0.0.0.0"),
+        Port(60000))
+      nodeKey2 = crypto.PrivateKey.random(Secp256k1, rng[])[]
+      node2 = WakuNode.new(nodeKey2, ValidIpAddress.init("0.0.0.0"),
+        Port(60002))
+      peerInfo2 = node2.peerInfo
+    
+    await allFutures([node1.start(), node2.start()])
+
+    node1.mountRelay()
+    node2.mountRelay()
+
+    # Dial node2 from node1
+    let conn = (await node1.peerManager.dialPeer(peerInfo2, WakuRelayCodec)).get()
+
+    #Disconnect the Peer
+    await node1.peerManager.disconnectPeer(node2.peerInfo, WakuRelayCodec)
+
+    #check if peer is still managed in node1
+    check:
+      node1.peerManager.hasPeer(node2.peerInfo, WakuRelayCodec) == false
+
+    await allFutures([node1.stop(), node2.stop()])
+
+  
--- a/tests/v2/test_peer_manager.nim
+++ b/tests/v2/test_peer_manager.nim
@@ -230,15 +230,12 @@ procSuite "Peer Manager":
       # Dial node2 from node1
       let conn = (await node2.peerManager.dialPeer(node1.peerInfo, WakuRelayCodec))
 
-      #check blacklist
-      check:
-        node1.peerInfo.peerId in node2.peerManager.blacklist == true
-
+    
       # Check that node2 is being managed in node1
       check:
         node2.peerManager.peers().anyIt(it.peerId == node1.peerInfo.peerId)
 
-      # Check connectedness
+      # Check connectedness. Should be equal to ShouldNotConnect if the peer has been blacklisted
       check:
         node2.peerManager.connectedness(node1.peerInfo.peerId) == Connectedness.ShouldNotConnect
       
@@ -267,7 +264,8 @@ procSuite "Peer Manager":
 
       #check blacklist
       check:
-        node1.peerInfo.peerId in node2.peerManager.blacklist == true
+        #node1.peerInfo.peerId in node2.peerManager.blacklist == true
+        node2.peerManager.connectedness(node1.peerInfo.peerId) == Connectedness.ShouldNotConnect
 
       await allFutures([node1.stop(), node2.stop()])
 
--- a/tests/v2/test_peer_manager.nim
+++ b/tests/v2/test_peer_manager.nim
@@ -322,7 +322,7 @@ procSuite "Peer Manager":
     let conn = (await node1.peerManager.dialPeer(peerInfo2, WakuRelayCodec)).get()
 
     #Disconnect the Peer
-    await node1.peerManager.disconnectPeer(node2.peerInfo, WakuRelayCodec)
+    node1.peerManager.disconnectPeer(node2.peerInfo, WakuRelayCodec)
 
     #check if peer is still managed in node1
     check:
--- a/tests/v2/test_peer_manager.nim
+++ b/tests/v2/test_peer_manager.nim
@@ -258,10 +258,7 @@ procSuite "Peer Manager":
       node1.mountRelay()
       node2.mountRelay()
       node2.peerManager.addToBlacklist(node1.peerInfo.peerId)
-
-      # Dial node2 from node1
-      let conn = (await node1.peerManager.dialPeer(peerInfo2, WakuRelayCodec)).get()
-
+      
       #check blacklist
       check:
         #node1.peerInfo.peerId in node2.peerManager.blacklist == true
@@ -292,7 +289,8 @@ procSuite "Peer Manager":
 
       #check blacklist
       check:
-        node1.peerInfo.peerId in node2.peerManager.blacklist == false
+        #node1.peerInfo.peerId in node2.peerManager.blacklist == false
+        node2.peerManager.connectedness(node1.peerInfo.peerId) == Connectedness.Connected
 
       await allFutures([node1.stop(), node2.stop()])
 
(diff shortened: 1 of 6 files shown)


EbubeUd:
@jm-clius is there an existing function for disconnecting a peer?

jm-clius:
> @jm-clius @staheri14 is there an existing function for disconnecting a peer?

Yes, there is a `disconnect` function on the `switch` [here](https://github.com/status-im/nim-libp2p/blob/master/libp2p/switch.nim#L97)

Calling that function should automatically change its `Connectedness` state in the `PeerManager` to `Disconnected`. If you want to also remove the peer from the manager, you'll have to add a `disconnect` proc to the manager that disconnects from the switch and clears the entry from the manager.

oskarth:
How is this going @EbubeUd?

EbubeUd:
Going well, fixing up some resulting errors.

waku/v2/protocol/waku_swap/waku_swap.nim
```diff
@@ -276,4 +286,12 @@ proc setPeer*(ws: WakuSwap, peer: PeerInfo) =
   ws.peerManager.addPeer(peer, WakuSwapCodec)
   waku_swap_peers_count.inc()
 
+proc getPeerConnectionStatus*(ws: WakuSwap, peerId : PeerId) : PeerConnectionStatus = 
+  let balance = ws.accounting[peerId];
```
oskarth, line 290: What if peerId doesn't exist?

waku/v2/protocol/waku_store/waku_store.nim
```diff
@@ -614,6 +614,12 @@ proc queryWithAccounting*(ws: WakuStore, query: HistoryQuery, handler: QueryHand
     waku_store_errors.inc(labelValues = [dialFailure])
     return
 
+  #Check if the Peer has gone below the disconnect threshold
+  let peerConnectionStatus = ws.wakuSwap.getPeerConnectionStatus(peerOpt.get().peerId)
```
oskarth, line 618: is peerOpt.get().peerId always defined?

waku/v2/node/peer_manager/peer_manager.nim
```diff
@@ -246,3 +246,10 @@ proc dialPeer*(pm: PeerManager, peerInfo: PeerInfo, proto: string, dialTimeout =
     return none(Connection)
 
   return await pm.dialPeer(peerInfo.peerId, peerInfo.addrs, proto, dialTimeout)
+
+proc disconnect*(pm: PeerManager, peerInfo: PeerInfo) : Future[bool] {.async.} =
+  let peerId =  peerInfo.peerId
+  debug "Removing peer from manager", peerId = peerInfo.peerId, addr = peerInfo.addrs[0]
+  let resp =  pm.peerStore.addressBook.delete(peerId) and pm.peerStore.protoBook.delete(peerId) and pm.peerStore.keyBook.delete(peerId)
```
oskarth, line 253: is peerId always in table?

EbubeUd: Cant guarantee that. I have added a check. Thanks

waku/v2/protocol/waku_store/waku_store.nim
EbubeUd: Thanks. Though I Felt that will be covered by if `peerOpt.isNone()`

waku/v2/protocol/waku_swap/waku_swap.nim
EbubeUd: Thanks. This was actually the main cause of the problems I had earlier. Later figured it out with trace logs enabled.

oskarth:
I see CI is failing and the branch is out of date. Can we make it up to date and green first?

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ This release contains the following:
 
 #### General refactoring
 #### Docs
+- Adds the database migration tutorial.
```
oskarth, line 14: Seems like rebase went wrong? This isn't part of the PR

EbubeUd:
Is it normal for the CI to check multiple commits? I can see some commits fail on the Jenkins builds.

waku/v2/protocol/waku_swap/waku_swap.nim
```diff
@@ -276,4 +288,14 @@ proc setPeer*(ws: WakuSwap, peer: PeerInfo) =
   ws.peerManager.addPeer(peer, WakuSwapCodec)
   waku_swap_peers_count.inc()
 
+proc getPeerConnectionStatus*(ws: WakuSwap, peerId : PeerId) : PeerConnectionStatus = 
+  result.balance = 0;
```
oskarth, line 292:
```suggestion
  result.balance = 0
```

oskarth: we avoid using result returns https://status-im.github.io/nim-style-guide/language.result.html

tests/v2/test_waku_swap.nim
```diff
@@ -107,7 +107,7 @@ procSuite "Waku SWAP Accounting":
     var futures = [newFuture[bool](), newFuture[bool]()]
```
oskarth, line 107: Maybe worthwhile to add a test to make sure disconnect logic works?

waku/v2/node/peer_manager/peer_manager.nim
```diff
@@ -246,3 +246,13 @@ proc dialPeer*(pm: PeerManager, peerInfo: PeerInfo, proto: string, dialTimeout =
     return none(Connection)
 
   return await pm.dialPeer(peerInfo.peerId, peerInfo.addrs, proto, dialTimeout)
+
+proc disconnect*(pm: PeerManager, peerInfo: PeerInfo) : Future[bool] {.async.} =
+  try:
+    let peerId =  peerInfo.peerId
+    debug "Removing peer from manager", peerId = peerInfo.peerId, addr = peerInfo.addrs[0]
+    waitFor pm.switch.disconnect(peerId)
+    let resp =  pm.peerStore.addressBook.delete(peerId) and pm.peerStore.protoBook.delete(peerId) and pm.peerStore.keyBook.delete(peerId)
```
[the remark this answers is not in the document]
jm-clius: Do you specifically not want to delete the peer from the `connectionBook` and `disconnectBook`? If it should be deleted from the `peerStore` completely, it's probably worth adding an explicit `deletePeer` function to the `waku_peer_store` that removes it from all books.

waku/v2/node/peer_manager/peer_manager.nim
```diff
@@ -246,3 +246,13 @@ proc dialPeer*(pm: PeerManager, peerInfo: PeerInfo, proto: string, dialTimeout =
     return none(Connection)
 
   return await pm.dialPeer(peerInfo.peerId, peerInfo.addrs, proto, dialTimeout)
+
+proc disconnect*(pm: PeerManager, peerInfo: PeerInfo) : Future[bool] {.async.} =
+  try:
+    let peerId =  peerInfo.peerId
+    debug "Removing peer from manager", peerId = peerInfo.peerId, addr = peerInfo.addrs[0]
```
jm-clius, line 253: To completely remove the peer from the manager, it would have to be removed from both the in-memory `peerStore` and the persistent `peerStorage` SQLite DB.
That would mean, however, that you would not "remember" a peer's blacklisted status after a restart and would happily reconnect if prompted to. Perhaps it's better to persist the peer's blacklisted-status as an extension to its [`Connectedness` state in the store](https://github.com/status-im/nim-waku/blob/master/waku/v2/node/peer_manager/waku_peer_store.nim#L15)
As it stands though, the peer will simply be reloaded into the store after a restart.

oskarth:
What's happening with this PR?

oskarth:
@EbubeUd Please reply to comments/resolve once resolved; rebase to latest; and ask for re-review again once it is ready to be reviewed.

In general, it shouldn't take a week to resolve PR comments, more like within a day.

EbubeUd:
> @EbubeUd Please reply to comments/resolve once resolved; rebase to latest; and ask for re-review again once it is ready to be reviewed.
> 
> In general, it shouldn't take a week to resolve PR comments, more like within a day.

Sorry, I've been away for some time. Just got back.  
I've refactored the code cos I found out that a couple of things were not really done properly initially.

oskarth:
@EbubeUd could you please respond to individual comments and click resolved if they've been addressed? That way when reviewing we can cover any potential new points. (The branch is also out of date, but might be quite recent)

waku/v2/node/peer_manager/waku_peer_store.nim
```diff
@@ -57,3 +59,18 @@ proc peers*(peerStore: WakuPeerStore): seq[StoredInfo] =
                        toSeq(keys(peerStore.keyBook.book))).toHashSet()
 
   return allKeys.mapIt(peerStore.get(it))
+
+
+proc deletePeer*(peerStore: WakuPeerStore, peerId: PeerID) =
```
jm-clius, line 64: What happens to a deleted peer after a restart? Wouldn't it just be loaded again from persistent storage (where that is configured)? This may be the behaviour we want for now - i.e. we don't "delete" blacklisted peers, just keep track of their blacklisted status?

waku/v2/node/peer_manager/peer_manager.nim
```diff
@@ -150,6 +155,8 @@ proc connectedness*(pm: PeerManager, peerId: PeerId): Connectedness =
   if (storedInfo == StoredInfo()):
     # Peer is not managed, therefore not connected
     return NotConnected
+  elif peerId in pm.blacklist:
+    return ShouldNotConnect
```
jm-clius, line 159: `NotConnected` is a special case since it means the peer is not in the peer store. 
`ShouldNotConnect` should probably be updated in the `connectionBook` and read from there.

waku/v2/node/peer_manager/peer_manager.nim
```diff
@@ -228,6 +235,13 @@ proc reconnectPeers*(pm: PeerManager, proto: string, backoff: chronos.Duration =
     trace "Reconnecting to peer", peerId=storedInfo.peerId
     discard await pm.dialPeer(storedInfo.peerId, toSeq(storedInfo.addrs), proto)
 
+proc addToBlacklist*(pm: PeerManager, peer: PeerId) = 
+  warn "Adding peer to blacklist"
+  pm.blacklist.add(peer)
```
jm-clius, line 240: Perhaps this should disconnect the peer and change its `connectedness` state to `ShouldNotConnect`? That way you don't have to keep track of a separate blacklist.

jm-clius:
Thanks, Ebube! Agree with @oskarth about responding and marking previous conversations as resolved so we can keep track of changes.
Also, please check your git diff before committing and unstage any local-only changes - you have two test binaries and two submodules in this PR that should not be merged.

EbubeUd:
> Thanks, Ebube! Agree with @oskarth about responding and marking previous conversations as resolved so we can keep track of changes.
> Also, please check your git diff before committing and unstage any local-only changes - you have two test binaries and two submodules in this PR that should not be merged.

Thanks. Noticed that as well. I'll unstage those files.

waku/v2/node/peer_manager/peer_manager.nim
EbubeUd: Hmm. I get what you mean here.  Currently, the blacklisting is not persistent just stores in an array. On a case of restart, the blacklist data will be lost.

waku/v2/node/peer_manager/waku_peer_store.nim
EbubeUd: We just want to keep track of the blacklisted peer, so persisting the blacklist should solve this.

waku/v2/node/peer_manager/peer_manager.nim
oskarth: I think this could be OK in the interest of getting this PR merged, but let's make sure we capture this in an enhancement issue then.

tests/v2/test_peer_manager.nim
```diff
@@ -209,3 +209,125 @@ procSuite "Peer Manager":
...
+    #Disconnect the Peer
+    node1.peerManager.disconnectPeer(node2.peerInfo, WakuRelayCodec)
+
+    #check if peer is still managed in node1
+    check:
+      node1.peerManager.hasPeer(node2.peerInfo, WakuRelayCodec) == false
+
+    await allFutures([node1.stop(), node2.stop()])
+
+
```
oskarth, line 333: Binary file `test_waku_swap` should be unstaged

waku/v2/protocol/waku_swap/waku_swap_types.nim
```diff
@@ -21,6 +21,8 @@ type
     mode* : SwapMode
     paymentThreshold* : int
     disconnectThreshold* : int
+    stateUpdateOverRide*: bool
```
oskarth, line 24: What is stateUpdateOveride? Not obvious to me. Perhaps add a comment?

waku/v2/protocol/waku_swap/waku_swap_types.nim
```diff
@@ -21,6 +21,8 @@ type
...
+    closePeerConnection*: bool
```
oskarth, line 25: This I think I understand but I'd add a comment here anyway

tests/v2/test_waku_swap.nim
```diff
@@ -107,7 +107,7 @@ procSuite "Waku SWAP Accounting":
...
 
     # Define the waku swap Config for this test
-    let swapConfig = SwapConfig(mode: SwapMode.Mock, paymentThreshold: 1, disconnectThreshold: -1)
+    let swapConfig = SwapConfig(mode: SwapMode.Mock, paymentThreshold: 1, disconnectThreshold: -1, stateUpdateOverRide: true, closePeerConnection: false)
```
oskarth, line 110: Does this mean that we do not disconnect the peer here? What does state update override signify?

Assuming above is correct, the disconnectThreshold being set seems a bit contradictory, though I can see why. Maybe leave a comment here about how this should probably be made optional if we don't disconnect peer?

Finally, shouldn't `SwapMode.Mock` imply that (1) There is no threshold for disconnects (2) closePeerConnection is false? As opposed to setting all three things at once.

tests/v2/test_waku_swap.nim
```diff
@@ -142,3 +142,108 @@ procSuite "Waku SWAP Accounting":
...
+        Port(60001))
+      contentTopic = ContentTopic("/waku/2/default-content/proto")
+      message = WakuMessage(payload: "hello world".toBytes(), contentTopic: contentTopic)
+
+    var completionFut = newFuture[bool]()
+    var futures = [newFuture[bool](), newFuture[bool]()]
+
+
+    # Define the waku swap Config for this test
+    let swapConfig = SwapConfig(mode: SwapMode.Mock, paymentThreshold: 1, disconnectThreshold: -1, stateUpdateOverRide: true, closePeerConnection: false)
```
oskarth, line 164: Ditto above, I'm confused about this SwapConfig considering the test comment "Add peer to blacklist after disconnect Threshold is reached"

tests/v2/test_waku_swap.nim
```diff
@@ -142,3 +142,108 @@ procSuite "Waku SWAP Accounting":
...
+    await node1.query(HistoryQuery(contentFilters: @[HistoryContentFilter(contentTopic: contentTopic)]), handler1)
+    await node1.query(HistoryQuery(contentFilters: @[HistoryContentFilter(contentTopic: contentTopic)]), handler2)
+
+    check:
+      (await allFutures(futures).withTimeout(5.seconds)) == true
+      # Accounting table updated with credit and debit, respectively
+      # After sending a cheque the balance is partially adjusted
+      node1.peerInfo.peerId in node2.wakuSwap.peerManager.blacklist == false
+    await node1.stop()
+    await node2.stop()
```
oskarth, line 249: nim-eth vendor should probably not be staged here?

oskarth: `nim-web3` in vendor directory should not be edited. If we need to update it, it should be a separate PR to that repo. Assuming it is just debugging stuff, unstage.

waku/v2/node/peer_manager/peer_manager.nim
```diff
@@ -4,7 +4,8 @@ import
   std/[options, sets, sequtils, times],
   chronos, chronicles, metrics,
   ./waku_peer_store,
-  ../storage/peer/peer_storage
+  ../storage/peer/peer_storage,
+  ../../../../vendor/nim-libp2p/libp2p/errors
```
oskarth, line 8:
```suggestion
libp2p/libp2p/errors
```

This should work, I think

waku/v2/node/peer_manager/waku_peer_store.nim
```diff
@@ -16,7 +16,9 @@ type
     # CanConnect: was recently connected to peer and disconnected gracefully.
     CanConnect,
     # Connected: actively connected to peer.
-    Connected
+    Connected,
+    # ShouldNotConnect: connection was refused
+    ShouldNotConnect
```
oskarth, line 21: If we are adding states to peer manager this should be updated in the https://rfc.vac.dev/spec/27/ spec. Is there any issue/PR in rfc repo for this?

waku/v2/node/peer_manager/waku_peer_store.nim
oskarth: On this note, we aren't really deleting a peer, more like removing it from peers we try to connect to (e.g.)

waku/v2/protocol/waku_swap/waku_swap.nim
```diff
@@ -192,18 +192,19 @@ proc handleCheque*(ws: WakuSwap, cheque: Cheque, peerInfo : PeerInfo) {.raises:
...
-  # makes waku_swap test pass for now
-  # Consider desired logic here
-  var stateUpdateOverRide = true
-
   if res.isOk():
     info "Updating accounting state with redeemed cheque"
     ws.accounting[peerId] += int(cheque.amount)
+
+    #Remove the Peer from the blacklist 
+    info "Removing peer from blacklist"
```
oskarth, line 200: I'd add peerId here for debugging

waku/v2/protocol/waku_swap/waku_swap.nim
```diff
@@ -251,28 +252,38 @@ proc init*(wakuSwap: WakuSwap) =
       wakuSwap.accounting[peerId] = n
     info "Accounting state", accounting = wakuSwap.accounting[peerId]
     wakuSwap.applyPolicy(peerInfo)
-    
-  proc applyPolicy(peerInfo: PeerInfo)
-    {.gcsafe, closure, raises: [Defect, KeyError, Exception].} =
+
+  proc disconnectThresholdReached(ws: WakuSwap, peerInfo: PeerInfo) {.gcsafe, closure.} =
```
oskarth, line 256: This is a predicate function and should return a boolean. Ditto below.

waku/v2/protocol/waku_swap/waku_swap.nim
```diff
@@ -251,28 +252,38 @@ proc init*(wakuSwap: WakuSwap) =
...
     info "Accounting state", accounting = wakuSwap.accounting[peerId]
     wakuSwap.applyPolicy(peerInfo)
-    
-  proc applyPolicy(peerInfo: PeerInfo)
-    {.gcsafe, closure, raises: [Defect, KeyError, Exception].} =
+
+  proc disconnectThresholdReached(ws: WakuSwap, peerInfo: PeerInfo) {.gcsafe, closure.} = 
+    let peerId = peerInfo.peerId
+    warn "Disconnect threshhold has been reached: ", threshold=wakuSwap.config.disconnectThreshold, balance=wakuSwap.accounting[peerId]
+    ws.peerManager.addToBlacklist(peerId);
```
oskarth, line 259: Mixing logic of applying policy with what should be a simple predicate function

waku/v2/protocol/waku_swap/waku_swap.nim
```diff
@@ -251,28 +252,38 @@ proc init*(wakuSwap: WakuSwap) =
...
+    if wakuSwap.config.mode == Mock:
+      discard wakuSwap.sendCheque(peerInfo)
+
+
+  proc applyPolicy(peerInfo: PeerInfo) {.gcsafe, closure, raises: [Defect, KeyError, Exception].} =
 
     #Check if the Disconnect Threshold has been hit. Account Balance nears the disconnectThreshold after a Credit has been done
-    if wakuSwap.accounting[peerId] <= wakuSwap.config.disconnectThreshold:
-      warn "Disconnect threshhold has been reached: ", threshold=wakuSwap.config.disconnectThreshold, balance=wakuSwap.accounting[peerId]
+    if wakuSwap.accounting[peerInfo.peerId] <= wakuSwap.config.disconnectThreshold:
```
oskarth, line 275: The point of the predicate functions is to use them here...

waku/v2/protocol/waku_swap/waku_swap_types.nim
EbubeUd: Added a comment. It indicates whether or not the account state of the peer will proceed to be updated be updated if there was an issue redeeming a cheque.

tests/v2/test_waku_swap.nim
EbubeUd: The Mock Mode is actually meant to cover peer disconnection, unlike the Soft mode. But the reason why there was an extra boolean field indicating whether or not the disconnect should occur is because there is an uncaught error that always occurs after switch.disconnect is called. The error seems to occur in libp2p, and it causes the test cases dependent on the disconnect function to fail.

waku/v2/node/peer_manager/peer_manager.nim
EbubeUd: Okay, I'll create a new Issue for this.

oskarth:
@EbubeUd please re-request review once you feel like PR is ready

oskarth:
Closing this PR as it is no longer being worked on.

PR #644, opened 2021-06-23
ci: improve GitHub actions build times

jakubgs:
Splitting GitHub builds and tests into more stages for clarity.

Adding a single Jenkinsfile for PR builds that can run on all three platforms supported by our Jenkins CI setup.


oskarth (approved):
Looks good to me!

Not sure about Windows time but I guess best we can do for now, assuming we want Windows as part of CI builds and not just nightly?

@jm-clius any preferences here? E.g. re what we do in PR vs nightly, etc

jm-clius (approved):
LGTM! Thanks, Jakub.

jm-clius:
> Not sure about Windows time but I guess best we can do for now, assuming we want Windows as part of CI builds and not just nightly?
> 
> @jm-clius any preferences here? E.g. re what we do in PR vs nightly, etc

Windows CI has been a pain, but it has been useful in identifying issues that didn't show up on the other platforms (generally because of subtle timing differences). I think we should maintain per-PR builds to have clear traceability of the origin of such issues.

oskarth:
Fair point, let's do this

kdeme (approved):
Not for this PR, but if you want to cut down on CI time more you can also split up the testing of waku v1 and waku v2 code in different Github Action CIs and run just one or the other depending on what files are changed. You can use include paths and paths-ignore for that: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-ignoring-paths

See also the set-up in nimbus-eth1:
https://github.com/status-im/nimbus-eth1/blob/master/.github/workflows/ci.yml#L4
and 
https://github.com/status-im/nimbus-eth1/blob/master/.github/workflows/fluffy.yml#L4

The downside is, that when changing common code, or for example updating the vendor modules, both CIs will run creating an actual longer run (as Nim and such need to be build twice)

oskarth:
@kdeme How much on an impact would you estimate this would have on Windows, given current bottlenecks, nim-waku code base and what you have seen for Nimbus? Specifically in the case where only Waku v2 code is changed.

kdeme:
> @kdeme How much on an impact would you estimate this would have on Windows, given current bottlenecks, nim-waku code base and what you have seen for Nimbus? Specifically in the case where only Waku v2 code is changed.

Well, in the case that you change code for Waku v2 only, the CI Windows time would drop about 11 minutes. Estimated from the times I see that it takes in this PR to do the "Build V1 Binaries" step (8m 11s) and the "Run V1 Tests" step (3m 6s).

Well worth it compared to the current total time of 45m. Especially considering the relatively low amount of "set-up" time before these tests (Checkout, submodules, etc.).

jakubgs:
Right, and that's simple to check for `waku/v1` and `waku/v2` folders, but what about things like `vendor` and `waku/common`? Changes in those folder would probably have to trigger both v1 and v2 builds.

michaelsbradleyjr (approved):

kdeme:
> Right, and that's simple to check for `waku/v1` and `waku/v2` folders, but what about things like `vendor` and `waku/common`? Changes in those folder would probably have to trigger both v1 and v2 builds.

Yup, vendor and waku/common and some other files (For example perhaps the Makefile / nimble file?, perhaps others ) will need to trigger both v1 and v2 builds. In the nimbus-eth1 repo, a similar thing is done. You can play around with those included paths (or ignored paths) to make that work.

jakubgs:
Well, I'm going to merge this as is, and then we can try playing with some logic like that.

Once merged I will disable github actions for this repo.

PR #645, opened 2021-06-24
DNS discovery: start of Merkle tree impl

jm-clius:
Increment towards https://github.com/status-im/nim-waku/issues/616

I've started on a framework towards implementing a Merkle Tree consisting of DNS TXT records formatted according to [spec](https://rfc.vac.dev/spec/25/#entry-types).

This PR contains a number of entry parsers, validators and tests. Code is still isolated from the main codebase.

A remaining question is where the DNS discovery code should live. For now, since our focus is on getting this working for Waku v2, I've added to `nim-waku`. However, since any peer `multiaddr` can be encoded in the discovered list, this potentially has a wider application than merely for Waku v2. As far as I remember, `nim-libp2p` does not want discovery in their codebase, as it falls outside specified scope. Any thoughts?


staheri14 (approved):

oskarth:
> A remaining question is where the DNS discovery code should live. For now, since our focus is on getting this working for Waku v2, I've added to nim-waku. However, since any peer multiaddr can be encoded in the discovered list, this potentially has a wider application than merely for Waku v2. As far as I remember, nim-libp2p does not want discovery in their codebase, as it falls outside specified scope. Any thoughts?

Separate module that can be used by any nim-libp2p code base, including nim-waku? What's the rationale for not having discovery in nim-libp2p, is it about stability/security for Nimbus?

waku/v2/dnsdisc/tree.nim
```diff
@@ -0,0 +1,125 @@
+{.push raises: [Defect]}
```
oskarth, line 1: I wouldn't put this in v2 folder, seems orthogonal to v2

arnetheduck:
> Separate module that can be used by any nim-libp2p code base, including nim-waku? What's the rationale for not having discovery in nim-libp2p, is it about stability/security for Nimbus?

discovery can in general be used by any process, including those that _don't_ use libp2p - specially when it's structured in a way that doesn't actively reuse anything libp2p-related - as discussed in the other ticket, we will need enr-based dns discovery for sure in nimbus (because that's what ethereum uses) 

what libp2p potentially could have is an interface for discovery mechanisms such that it can request peers from a variety of sources - I'm a bit wary of that approach though - it works for 100%-libp2p-based applications but anything that relies on a mixed stack will have trouble fitting it in because the libp2p interface will inevitably be based on libp2p needs - in other words, it makes sense to develop each discovery mechanism to the maximum of its potential and functionality outside of libp2p, then plug it into the subset that is the libp2p interface if needed


cc @dryajov @Menduist

Menduist:
Yes, a dns discovery interface is on the roadmap https://github.com/status-im/nim-libp2p/issues/140

It would be totally optional and probably fairly simple, though I don't have a clear idea of it ATM

jm-clius:
Right, so this will work best as its own library. My proposal then is to move this to a separate repo called `nim-disc` or `nim-dns-disc`. I prefer the former as it may well grow to include more general discovery utilities, though ofc other discovery mechanisms already live elsewhere.

dryajov:
> > Separate module that can be used by any nim-libp2p code base, including nim-waku? What's the rationale for not having discovery in nim-libp2p, is it about stability/security for Nimbus?
> 
> discovery can in general be used by any process, including those that _don't_ use libp2p - specially when it's structured in a way that doesn't actively reuse anything libp2p-related - as discussed in the other ticket, we will need enr-based dns discovery for sure in nimbus (because that's what ethereum uses)
> 
> what libp2p potentially could have is an interface for discovery mechanisms such that it can request peers from a variety of sources - I'm a bit wary of that approach though - it works for 100%-libp2p-based applications but anything that relies on a mixed stack will have trouble fitting it in because the libp2p interface will inevitably be based on libp2p needs - in other words, it makes sense to develop each discovery mechanism to the maximum of its potential and functionality outside of libp2p, then plug it into the subset that is the libp2p interface if needed
> 
> cc @dryajov @Menduist

Yeah, discovery should have a standard interface so that it is easily plugable, but the libp2p core should be as agnostic to the underlying mechanism as possible so that applications can provide their own.

jm-clius:
Implementation moved to https://github.com/status-im/nim-dnsdisc

PR #648, opened 2021-06-24
Add db migration support for peer store

staheri14:
Closes https://github.com/status-im/nim-waku/issues/647
- It reorganizes the migration code and moves all the migration paths to the `migration_types` module.
- Adds the initial migration script for the first version of the peer store. 
- I also sneaked in and fixed a subtle issue in the resume proc which I came across while interacting with wakunode (it does not show up in the tests and I am still working on replicating the problem).  The issue is that there is an inconsistency between the total number of messages stored in the memory vs DB.  The DB insertion fails for some of the messages. To preserve consistency, I swapped the order of insertion so that a message would be inserted inside the `messages` field only if its insertion to the DB goes through. 

Following this PR, the migration process would be straightforward:
- Increase the `user_version`
- Place the migration script under the migrations/[message or peer] folder


jm-clius (approved):
LGTM. Thanks for adding support for peer store. 👍

PR #656, opened 2021-07-01
ci: build and test only what changed

jakubgs:
This is a pretty cute regex-based test, but should work.

Without any changes that would trigger builds the CI run looks like this:
image

But with only the v1 path triggered it looks like this:
image

For the diff to work this refspec needs to be added to the fetch: `+refs/heads/master:refs/remotes/@{remote}/master`

### DIFF
--- a/ci/Jenkinsfile.prs
+++ b/ci/Jenkinsfile.prs
@@ -41,7 +41,11 @@ pipeline {
   }
 
   stages {
-    stage('Deps') { steps {
+    stage('Deps') { steps { script {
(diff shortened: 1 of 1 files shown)


jm-clius (approved):
Looks reasonable. Thanks! Seems like `docs`-only changes won't trigger builds which is 👍

oskarth (approved):
Thanks! Hopefully it works too ;)

oskarth:
As in: I can't tell if it'll work or not, not that I doubt it would!

PR #657, opened 2021-07-02
History response pagination using sender-generated timestamp

staheri14:
Closes https://github.com/vacp2p/rfc/issues/219 and partial progress towards https://github.com/vacp2p/rfc/issues/406
Paging based on sender timestamp is more conventional, and in this PR the role of receiver timestamp in the pagination is entirely replaced by the sender timestamp.


staheri14:
@jakubgs the Jenkins log does not give much info to be able to track the failure, what are your thoughts? Thanks!

staheri14:
@jakubgs I restarted the runs and errors are gone!

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ This release contains the following:
 - Updates the `Message` table of the persistent message store:
   - Adds `senderTimestamp` column.
   - Renames the `timestamp` column to `receiverTimestamp` and changes its type to `REAL`.
+  - Deletes  `receiverTimestamp`.
```
oskarth, line 22: Do we not want to track this still? Seems a bit weird to me to remove this completely, especially as it is more "objective"

waku/v2/node/storage/message/waku_message_store.nim
```diff
@@ -64,15 +63,15 @@ method put*(db: WakuMessageStore, cursor: Index, message: WakuMessage, pubsubTop
   ##     echo "error"
   ## 
   let prepare = db.database.prepareStmt(
-    "INSERT INTO " & TABLE_TITLE & " (id, receiverTimestamp, contentTopic, payload, pubsubTopic, version, senderTimestamp) VALUES (?, ?, ?, ?, ?, ?, ?);",
-    (seq[byte], float64, seq[byte], seq[byte], seq[byte], int64, float64),
+    "INSERT INTO " & TABLE_TITLE & " (id, contentTopic, payload, pubsubTopic, version, senderTimestamp) VALUES (?, ?, ?, ?, ?, ?);",
```
oskarth, line 66: I'd keep both around, and then just have one that is defaulted to index (this can be sender if we think it is a more common use case)

tests/v2/test_wakunode.nim
```diff
@@ -821,7 +821,8 @@ procSuite "WakuNode":
 
     # count the total number of retrieved messages from the database
     var responseCount = 0
-    proc data(receiverTimestamp: float64, msg: WakuMessage, psTopic: string) =
+    proc data(msg: WakuMessage, psTopic: string) =
+      echo "here"
```
jm-clius, line 825: leftover `echo`? :)

waku/v2/node/storage/migration/migrations_scripts/message/00003_deleteReceiverTimeStamp.up.sql
```diff
@@ -0,0 +1,26 @@
...
+        id BLOB PRIMARY KEY,
+        contentTopic BLOB NOT NULL,
+        pubsubTopic BLOB NOT NULL,
+        payload BLOB,
+        version INTEGER NOT NULL,
+        senderTimestamp REAL NOT NULL
+    ) WITHOUT ROWID;
+
+
+INSERT INTO Message SELECT id, contentTopic, pubsubTopic, payload, version, senderTimestamp  FROM Message_backup;
```
jm-clius, line 24: Wondering for interest's sake what the benefit is of doing it this way (creating a new table based on the previous) vs just deleting the column (`ALTER TABLE...DROP COLUMN`)? This will still involve making a backup copy, of course.

staheri14: Good question, unfortunately, SQlite does not support the `ALTER TABLE` command

waku/v2/node/storage/message/waku_message_store.nim
staheri14: Sure, I can keep the receiver timestamp column inside the DB, however, we will have no code to generate the receiver timestamp of the incoming messages. The receiver timestamp used to be part of computing the index of the received waku message which is now replaced by the sender timestamp.
The reason I have removed the receiver timestamp entirely was to avoid subsequent confusion.
What would you see as the use-case for the receiver timestamp?

FYI, we had also a conversation during EU/NA call about the potential use-case of the receiver timestamp to spot messages with falsified timestamps, However, such timestamp validation can be done simply by comparing the receiver current time against the timestamp of the incoming message without having to store the receiver timestamp locally.

tests/v2/test_wakunode.nim
staheri14: Forgot to clean it up

waku/v2/protocol/waku_store/waku_store.nim
```diff
@@ -47,8 +47,8 @@ proc computeIndex*(msg: WakuMessage): Index =
   ctx.update(msg.payload)
   let digest = ctx.finish() # computes the hash
   ctx.clear()
-  result.digest = digest
-  result.receivedTime = epochTime() # gets the unix timestamp
```
staheri14, line 51: @oskarth related to my comment about the lack of receiver timestamp computation, please see this line. In here we generate the receiver-side timestamp as part of the Index struct, but in this PR, this is going to be replaced by the sender timestamp,

staheri14:
Updated the PR: @oskarth @jm-clius 
In the current update, the receiver timestamp remains in the `Index` struct and gets transferred as part of the pagination, however, it has no effect on the sorting of waku messages during the pagination. Sorting is only based on the message digest and sender timestamp.

tests/v2/test_waku_pagination.nim
```diff
@@ -44,9 +45,9 @@ procSuite "pagination":
     for x in data3.mitems: x = 3
       
     let
-      index1 = Index(receivedTime: 1, digest: MDigest[256](data: data1))
-      index2 = Index(receivedTime: 1, digest: MDigest[256](data: data2))
-      index3 = Index(receivedTime: 2, digest: MDigest[256](data: data3))
+      index1 = Index(receiverTime: 2, senderTime: 1, digest: MDigest[256](data: data1))
+      index2 = Index(receiverTime: 2, senderTime:1, digest: MDigest[256](data: data2))
+      index3 = Index(receiverTime: 1, senderTime:2, digest: MDigest[256](data: data3))
```
oskarth, line 50:
```suggestion
      index2 = Index(receiverTime: 2, senderTime: 1, digest: MDigest[256](data: data2))
      index3 = Index(receiverTime: 1, senderTime: 2, digest: MDigest[256](data: data3))
```

oskarth (approved):
LGTM

jm-clius (approved):
Nice!

PR #660, opened 2021-07-08
Adding error message to the History Responses

staheri14:
This closes https://github.com/vacp2p/rfc/issues/248 and is incremental progress towards https://github.com/vacp2p/rfc/issues/406.

I also utilized this PR to do some clean-up on the pagination logic as part of https://github.com/status-im/nim-waku/issues/484. Removed the private proc `paginateWithoutIndex` and shifted its code inside the `findMessage` proc for the sake of better code readability. Moreover,  the `paginateWithoutIndex` had simple logic i.e., peeling off the `Index` from `IndexedWakuMessages`, so having it as a separate proc would only complicate the code. Following this change, I have renamed another existing proc i.e., `paginateWithIndex` to `paginate`.

Also, replaced the use of `Result`  with the explicit `return` call (as suggested in track 1 checklist of the refactor week https://hackmd.io/1imOGULZRsed2HpgmzGleA).


waku/v2/protocol/waku_store/waku_store.nim
```diff
@@ -299,20 +305,13 @@ proc paginateWithIndex*(list: seq[IndexedWakuMessage], pinfo: PagingInfo): (seq[
       newCursor = msgList[s].index # the new cursor points to the begining of the page
 
   if (retrievedPageSize == 0):
-    return (@[], PagingInfo(pageSize: 0, cursor:pinfo.cursor, direction: pinfo.direction))
+    return (@[], PagingInfo(pageSize: 0, cursor:pinfo.cursor, direction: pinfo.direction), HistoryResponseError.OK)
```
oskarth, line 308: What is HistoryResponseError.OK? Seems like an oxymoron, isn't 0 pages retrieved, if correct, an OK response?

waku/v2/protocol/waku_store/waku_store_types.nim
```diff
@@ -52,9 +52,15 @@ type
     startTime*: float64 # used for time-window query
     endTime*: float64 # used for time-window query
 
+  HistoryResponseError* {.pure.} = enum
+    ## HistoryResponseError contains error message to inform  the querying node about the state of its request
+    OK = uint32(0)
```
oskarth, line 57: Why is this an error?

oskarth:
I might be missing some context, but to me:

> In the current implementation of pagination, a history query with an invalid cursor (as part of the paging info) gets responded by an empty set of waku messages. Likewise, if the topics field of the history query does not match any of the historical messages, the returned set of messages is empty. The preceding cases are indistinguishable in the current implementation of store protocol thus the querying node has no clue as to why no waku messages is retrieved. (From https://github.com/vacp2p/rfc/issues/248)

means we need give error for invalid cursors. If topic field doesn't match, then seems like an OK response and not an error? Perhaps informational, as in "no messages for topic". What do other standard APIs do in this scenario?

waku/v2/protocol/waku_store/waku_store.nim
```diff
@@ -245,7 +251,7 @@ proc findIndex*(msgList: seq[IndexedWakuMessage], index: Index): Option[int] =
       return some(i)
   return none(int)
 
-proc paginateWithIndex*(list: seq[IndexedWakuMessage], pinfo: PagingInfo): (seq[IndexedWakuMessage], PagingInfo) =
+proc paginate*(list: seq[IndexedWakuMessage], pinfo: PagingInfo): (seq[IndexedWakuMessage], PagingInfo, HistoryResponseError) =
```
jm-clius, line 254: Perhaps if this defaults to `OK` I imagine you wouldn't have to explicitly set `HistoryResponseError.OK` in so many places?

waku/v2/protocol/waku_store/waku_store_types.nim
jm-clius: I'd suggest making this `NONE` and the default. That way we're not mixing "Success" and "Error" codes (which should be avoided) and don't have to use it other than for error conditions.

waku/v2/protocol/waku_store/waku_store.nim
```diff
@@ -299,20 +305,13 @@ proc paginateWithIndex*(list: seq[IndexedWakuMessage], pinfo: PagingInfo): (seq[
...
-
-proc paginateWithoutIndex(list: seq[IndexedWakuMessage], pinfo: PagingInfo): (seq[WakuMessage], PagingInfo) =
-  ## takes list, and perfomrs paging based on pinfo 
-  ## returns the page i.e, a sequence of WakuMessage and the new paging info to be used for the next paging request  
-  var (indexedData, updatedPagingInfo) = paginateWithIndex(list,pinfo)
-  for indexedMsg in indexedData:
-    result[0].add(indexedMsg.msg)
-  result[1] = updatedPagingInfo
+    retMessages.add(msgList[i])
+  return (retMessages, PagingInfo(pageSize : retrievedPageSize, cursor : newCursor, direction : pinfo.direction), HistoryResponseError.OK)
```
jm-clius, line 314: Thanks for removing `result`! :)

waku/v2/protocol/waku_store/waku_store.nim
staheri14: The goal of the error message is to distinguish between two cases that both result in 0 message history. 1) when the cursor is invalid 2) the cursor is valid but there is no match for the submitted query
When an error message is OK, it means the second scenario happened. Having zero messages in the response is not by itself an error, but if it is because of an invalid cursor then it is an error (as the input is invalid).

waku/v2/protocol/waku_store/waku_store_types.nim
staheri14: Sure, that makes sense. I will change its name to none!

waku/v2/protocol/waku_store/waku_store.nim
staheri14: The default is OK, however, as part of the unit testing, I need to check that the different execution paths of `paginate` that yield OK behave as expected and do not produce an error.
Please let me know if there is a specific part of the code that the use of OK is unnecessary and can be avoided.

staheri14: BTW, as I am using `return` to terminate the proc, I cannot leave the third output value unspecified.

waku/v2/protocol/waku_store/waku_store.nim
staheri14: Following @jm-clius comment https://github.com/status-im/nim-waku/pull/660#discussion_r666993971, I have renamed OK to NONE to resolve the naming ambiguity.

waku/v2/protocol/waku_store/waku_store_types.nim
staheri14: OK means no error, and I agree that NONE makes more sense! @oskarth

staheri14:
> APIs

Our pagination method is almost identical to the one used in Slack https://api.slack.com/docs/pagination#cursors
The only error message associated with the pagination on the Slack is the same as ours i.e., `invalid_cursor`

I also checked Twitter pagination, it appears they do not consider any error message for invalid cursors/tokens https://developer.twitter.com/en/docs/twitter-api/pagination

staheri14:
@oskarth 
I initially did not think of the query mismatch as an error, I can still add further error messages related to other fields of queries 
1- the pubsub topic was not found
2- the time range was not found 
3- the content topic was not found

Let me know if further error messages is desired

waku/v2/protocol/waku_store/waku_store.nim
staheri14: Sure!:)

waku/v2/protocol/waku_store/waku_store_types.nim
```diff
@@ -52,9 +52,15 @@ type
...
 
+  HistoryResponseError* {.pure.} = enum
+    ## HistoryResponseError contains error message to inform  the querying node about the state of its request
+    NONE = uint32(0)
+    INVALID_CURSOR = uint32(1)
+
   HistoryResponse* = object
     messages*: seq[WakuMessage]
     pagingInfo*: PagingInfo # used for pagination
+    error*: HistoryResponseError
```
oskarth, line 63: Maybe I'm still missing something, but if `HistoryResponseError` is `NONE`, couldn't that intent be communicated with `error*: Option[HistoryResponsError]`?

oskarth: In protobuf terms, that field would be emitted, so the equivalent thing (`0`) would go over the wire, I believe.

oskarth:
See comment above re possibly using `Option` type, to me that seems cleaner but I might still be missing something here.

> Let me know if further error messages is desired

Generally an extensible way of dealing with this is to have an optional error string, that way things can be elaborated on without a protocol upgrade. This helps with logs and debugging, for example if further down the line you want errors like "permission denied" "rate limiting reached (300/500)". It can be complemented with an error code, like 404, 500 etc in HTTP, but for now I think knowing 1) It is an error 2) Invalid cursor is good enough!

oskarth:
As for what error are needed _now_, might make sense to get @richard-ramos @siphiuel POV in terms of what Core/Desktop needs

staheri14:
> See comment above re possibly using `Option` type, to me that seems cleaner but I might still be missing something here.
> 
> > Let me know if further error messages is desired
> 
> Generally an extensible way of dealing with this is to have an optional error string, that way things can be elaborated on without a protocol upgrade. This helps with logs and debugging, for example if further down the line you want errors like "permission denied" "rate limiting reached (300/500)" or the "foobar not found" you suggested. It can be complemented with an error code, like 404, 500 etc in HTTP, for more automatic error handling but for now I think knowing 1) It is an error 2) Invalid cursor is good enough!

Thanks for the comments @oskarth!

As in the case of the HTTP response status codes, there are plenty of options covering a successful response like [200 OK](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/200). 
The reason I used error codes instead of error strings was to have a well-defined and precise set of error codes with exact definitions that can be consistently perceived by the two ends of the communication, I found error strings error-prone and a bit ad-hoc in this sense.

waku/v2/protocol/waku_store/waku_store_types.nim
staheri14: >Maybe I'm still missing something, but if HistoryResponseError is NONE, couldn't that intent be communicated with error*: Option[HistoryResponsError]? Right now, the way I'm reading it, it seems like an ad-hoc implementation of Option type.

If we want the error codes to include the successful responses, then the `error` field should not be optional. Otherwise, this can be converted to an optional field. 

I suggest not to make the `error` field optional (although in protobuf fields are optional by default).  The reason is to make sure that the `error` is not omitted by mistake rather the response has indeed been concluded successfully.

Let me know whether this makes sense or shall I convert it to optional (or make it a simple string error). I'll follow your discretion.

oskarth (approved):
Let's get this merged, if we want to tweak it at a later point it shouldn't be an issue. Thanks for the patience!

jm-clius (approved):

staheri14:
> Let's get this merged, if we want to tweak it at a later point it shouldn't be an issue. Thanks for the patience!

Sure! thanks for the comments @oskarth , definitely please let me know if further updates would be required to make the error handling more effective and useful! :)

PR #662, opened 2021-07-12
Simplify wakunode2 subscriptions model

### LINKED ISSUE #574: Simplify `wakunode2` `subscriptions` model
## Problem

In order to notify mounted protocols of new messages, `wakunode2` makes use of `subscriptions` - an extendable map of protocol strings to handlers. This is used, for example, to notify the `store` of new messages or to check if a message match any registered `filters`.

This pattern works, but is unnecessarily confusing for two reasons:
1. It is difficult to read and follow due to the indirection it introduces. It should be possible to simply call `notify` on mounted protocols without having them register a handler in a separate map.
2. It overloads the idea of a `subscription`, which is also used to refer to topic subscriptions in the PubSub model.

## Acceptance criteria

- [x] something that simplifies the above, is clear to read and doesn't overload terminology

jm-clius:
This PR closes #574 

It forms part of Track 2 of the [refactor week](https://hackmd.io/beMbu_fdRymYvaXDANukIw)

Note that I have not yet refactored any of the modules according to the [Track 1 checklist](https://hackmd.io/1imOGULZRsed2HpgmzGleA), though that will be the next step.

## Summary of changes

The `MessageNotificationSubscription` workflow is removed. This was previously used to notify mounted protocols of new messages. 
Instead, each protocol that should handle new messages (currently only `filter` and `store`) now exposes an explicit `handleMessage` function on its API. This is called from the `wakunode2` for all such mounted protocols on every new message.

## What are the benefits?

1. This explicit approach is much easier to read and debug than the indirect `subscriptions` approach.
2. It removes the ambiguous use of the overloaded term `subscription`, which was used to refer to both topic subscriptions and message notification subscriptions.

@staheri14, note that this change required moving the `persistMessages` logic to the `waku_store` itself. I think it makes sense for this to be part of the `WakuStore` state, but LMK if you have any concerns.


oskarth (approved):
Perfect! This is exactly what I wanted way back when this feature was being implemented :)

PR #663, opened 2021-07-13
Move Whisper types and protocol from nim-eth to this repo

oskarth:
Addresses https://github.com/status-im/nim-waku/issues/654 as part of refactor week

### DIFF
--- a/waku/v1/protocol/waku_bridge.nim
+++ b/waku/v1/protocol/waku_bridge.nim
@@ -12,7 +12,8 @@
 
 import
   eth/p2p,
-  eth/p2p/rlpx_protocols/whisper_protocol,
+  #eth/p2p/rlpx_protocols/whisper_protocol,
+  ../../whisper/whisper_protocol,
   ./waku_protocol
 
 proc shareMessageQueue*(node: EthereumNode) =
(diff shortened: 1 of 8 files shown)


oskarth:
@zah since you wrote the p2p macro IIRC, perhaps you have some ideas for how to do explicit error handling in above?

oskarth:
After https://github.com/status-im/nim-eth/pull/374 has been merged the above should be easier, I'll look at it next week

jm-clius (approved):
LGTM

waku/v1/protocol/waku_bridge.nim
```diff
@@ -12,7 +12,8 @@
 
 import
   eth/p2p,
-  eth/p2p/rlpx_protocols/whisper_protocol,
+  #eth/p2p/rlpx_protocols/whisper_protocol,
```
kdeme, line 15: Left over comment

kdeme (commented):
I'm rather indifferent with keeping whisper_protocol in nim-eth or moving it here. But if it gets moved, shouldn't the tests for whisper protocol also be moved (which are rather similar, I know)?

oskarth:
> I'm rather indifferent with keeping whisper_protocol in nim-eth or moving it here. But if it gets moved, shouldn't the tests for whisper protocol also be moved (which are rather similar, I know)?

Right, fair point

PR #664, opened 2021-07-13
Refactor wakunode2.nim

jm-clius:
## Background

This PR focuses on refactoring `wakunode2.nim` according to the [Track 1 refactoring checklist](https://hackmd.io/1imOGULZRsed2HpgmzGleA?view)

As a result multiple referenced modules required similar refactoring and improvements, especially related to error handling.

The first step towards improved error handling is being explicit about what exceptions we raise, and to avoid generic types such as `Exception` or `CatchableError`. A next step is to determine which of these raised exceptions we should catch, though such changes fall outside the scope of this single PR.

## Note:
- Some underlying modules, especially utils, still raise `Exception` where they should more properly return a `Result`. This will be addressed in follow-up PRs, once those modules undergo Track 1 refactoring.
- `import`s are likely to be cleaned up further once referenced modules export their own referenced types.
- `waitFor` raises `CatchableError`. In future we should probably catch these and raise our own error.
- `metrics` raises `Exception` in some places. Could be part of our upstream effort to fix this/log issue?

### DIFF
--- a/waku/v2/node/peer_manager/peer_manager.nim
+++ b/waku/v2/node/peer_manager/peer_manager.nim
@@ -36,7 +36,7 @@ proc insertOrReplace(ps: PeerStorage,
                      peerId: PeerID,
                      storedInfo: StoredInfo,
                      connectedness: Connectedness,
-                     disconnectTime: int64 = 0) {.raises: [Defect, Exception]} =
+                     disconnectTime: int64 = 0) =
   # Insert peer entry into persistent storage, or replace existing entry with updated info
   let res = ps.put(peerId, storedInfo, connectedness, disconnectTime)
   if res.isErr:
@@ -77,7 +77,7 @@ proc dialPeer(pm: PeerManager, peerId: PeerID,
     
     return none(Connection)
 
-proc loadFromStorage(pm: PeerManager) {.raises: [Defect, Exception]} =
+proc loadFromStorage(pm: PeerManager) =
   # Load peers from storage, if available
   proc onData(peerId: PeerID, storedInfo: StoredInfo, connectedness: Connectedness, disconnectTime: int64) =
     if peerId == pm.switch.peerInfo.peerId:
@@ -112,7 +112,7 @@ proc onConnEvent(pm: PeerManager, peerId: PeerID, event: ConnEvent) {.async.} =
       pm.storage.insertOrReplace(peerId, pm.peerStore.get(peerId), CanConnect, getTime().toUnix)
     return
 
-proc new*(T: type PeerManager, switch: Switch, storage: PeerStorage = nil): PeerManager {.raises: [Defect, Exception]} =
+proc new*(T: type PeerManager, switch: Switch, storage: PeerStorage = nil): PeerManager =
   let pm = PeerManager(switch: switch,
                        peerStore: WakuPeerStore.new(),
                        storage: storage)
@@ -162,7 +162,7 @@ proc hasPeers*(pm: PeerManager, proto: string): bool =
   # Returns `true` if manager has any peers for the specified protocol
   pm.peers.anyIt(it.protos.contains(proto))
 
-proc addPeer*(pm: PeerManager, peerInfo: PeerInfo, proto: string) {.raises: [Defect, Exception]} =
+proc addPeer*(pm: PeerManager, peerInfo: PeerInfo, proto: string) =
   # Adds peer to manager for the specified protocol
 
   if peerInfo.peerId == pm.switch.peerInfo.peerId:
--- a/waku/v2/node/storage/peer/waku_peer_storage.nim
+++ b/waku/v2/node/storage/peer/waku_peer_storage.nim
@@ -1,3 +1,5 @@
+{.push raises: [Defect].}
+
 import
   std/sets, 
   sqlite3_abi,
@@ -36,7 +38,7 @@ proc init*(T: type StoredInfo, buffer: seq[byte]): ProtoResult[T] =
 
   ok(storedInfo)
 
-proc encode*(storedInfo: StoredInfo): ProtoBuffer =
+proc encode*(storedInfo: StoredInfo): PeerStorageResult[ProtoBuffer] =
   var pb = initProtoBuffer()
 
   pb.write(1, storedInfo.peerId)
@@ -47,9 +49,12 @@ proc encode*(storedInfo: StoredInfo): ProtoBuffer =
   for proto in storedInfo.protos.items:
     pb.write(3, proto)
   
-  pb.write(4, storedInfo.publicKey)
+  try:
+    pb.write(4, storedInfo.publicKey)
+  except ResultError[CryptoError] as e:
+    return err("Failed to encode public key")
 
-  return pb
+  ok(pb)
 
 ##########################
 # Storage implementation #
@@ -97,7 +102,12 @@ method put*(db: WakuPeerStorage,
   if prepare.isErr:
     return err("failed to prepare")
 
-  let res = prepare.value.exec((peerId.data, storedInfo.encode().buffer, int32(ord(connectedness)), disconnectTime))
+  let encoded = storedInfo.encode()
+
+  if encoded.isErr:
+    return err("failed to encode: " & encoded.error())
+
+  let res = prepare.value.exec((peerId.data, encoded.get().buffer, int32(ord(connectedness)), disconnectTime))
   if res.isErr:
     return err("failed")
 
@@ -107,7 +117,7 @@ method getAll*(db: WakuPeerStorage, onData: peer_storage.DataProc): PeerStorageR
   ## Retrieves all peers from storage
   var gotPeers = false
 
-  proc peer(s: ptr sqlite3_stmt) = 
+  proc peer(s: ptr sqlite3_stmt) {.raises: [Defect, LPError, ResultError[ProtoError]].} = 
     gotPeers = true
     let
       # Peer ID
@@ -125,8 +135,13 @@ method getAll*(db: WakuPeerStorage, onData: peer_storage.DataProc): PeerStorageR
 
     onData(peerId, storedInfo, connectedness, disconnectTime)
 
-  let res = db.database.query("SELECT peerId, storedInfo, connectedness, disconnectTime FROM Peer", peer)
-  if res.isErr:
+  var queryResult: DatabaseResult[bool]
+  try:
+    queryResult = db.database.query("SELECT peerId, storedInfo, connectedness, disconnectTime FROM Peer", peer)
+  except LPError, ResultError[ProtoError]:
+    return err("failed to extract peer from query result")
+  
+  if queryResult.isErr:
     return err("failed")
 
   ok gotPeers
--- a/waku/v2/node/wakunode2.nim
+++ b/waku/v2/node/wakunode2.nim
@@ -144,13 +144,15 @@ proc new*(T: type WakuNode, nodeKey: crypto.PrivateKey,
   #    msgIdProvider = msgIdProvider,
   #    triggerSelf = true, sign = false,
   #    verifySignature = false).PubSub
-  return WakuNode(
+  let wakuNode = WakuNode(
     peerManager: PeerManager.new(switch, peerStorage),
     switch: switch,
     rng: rng,
     peerInfo: peerInfo,
     filters: initTable[string, Filter]()
   )
+
+  return wakuNode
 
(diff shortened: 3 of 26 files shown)


waku/v2/node/peer_manager/peer_manager.nim
```diff
@@ -162,7 +162,7 @@ proc hasPeers*(pm: PeerManager, proto: string): bool =
   # Returns `true` if manager has any peers for the specified protocol
   pm.peers.anyIt(it.protos.contains(proto))
 
-proc addPeer*(pm: PeerManager, peerInfo: PeerInfo, proto: string) {.raises: [Defect, Exception]} =
```
staheri14, line 165: Could you please explain why the `raises` pragma is removed in here?

staheri14: I think that is because of the `{.push raises: [Defect].}` https://github.com/status-im/nim-waku/blob/37c5a6eb222921375a2f9bcedc9133b2c4b1b705/waku/v2/node/peer_manager/peer_manager.nim#L1` at the top of the file

waku/v2/node/storage/peer/waku_peer_storage.nim
```diff
@@ -47,9 +49,12 @@ proc encode*(storedInfo: StoredInfo): ProtoBuffer =
   for proto in storedInfo.protos.items:
     pb.write(3, proto)
   
-  pb.write(4, storedInfo.publicKey)
+  try:
+    pb.write(4, storedInfo.publicKey)
+  except ResultError[CryptoError] as e:
```
staheri14, line 54: I am asking this for my own understanding, I thought we should use `Result` when multiple failure paths exist, is it the case here? It seems there is only one failure path

waku/v2/node/storage/peer/waku_peer_storage.nim
```diff
@@ -107,7 +117,7 @@ method getAll*(db: WakuPeerStorage, onData: peer_storage.DataProc): PeerStorageR
   ## Retrieves all peers from storage
   var gotPeers = false
 
-  proc peer(s: ptr sqlite3_stmt) = 
+  proc peer(s: ptr sqlite3_stmt) {.raises: [Defect, LPError, ResultError[ProtoError]].} =
```
staheri14, line 120: Should we use `{.raises.}` for non public procs?

waku/v2/node/wakunode2.nim
```diff
@@ -406,14 +395,14 @@ proc mountFilter*(node: WakuNode) =
 
 # NOTE: If using the swap protocol, it must be mounted before store. This is
 # because store is using a reference to the swap protocol.
-proc mountSwap*(node: WakuNode, swapConfig: SwapConfig = SwapConfig.init()) =
+proc mountSwap*(node: WakuNode, swapConfig: SwapConfig = SwapConfig.init()) {.raises: [Defect, LPError].} =
```
staheri14, line 398: There is already a `{.push raises: [Defect].}` at the top, do we need another one in here?

waku/v2/node/wakunode2.nim
```diff
@@ -406,14 +395,14 @@ proc mountFilter*(node: WakuNode) =
...
   info "mounting swap", mode = $swapConfig.mode
   node.wakuSwap = WakuSwap.init(node.peerManager, node.rng, swapConfig)
   node.switch.mount(node.wakuSwap, protocolMatcher(WakuSwapCodec))
   # NYI - Do we need this?
   #node.subscriptions.subscribe(WakuSwapCodec, node.wakuSwap.subscription())
 
-proc mountStore*(node: WakuNode, store: MessageStore = nil, persistMessages: bool = false) =
+proc mountStore*(node: WakuNode, store: MessageStore = nil, persistMessages: bool = false) {.raises: [Defect, LPError].} =
```
staheri14, line 405: Ditto

waku/v2/node/wakunode2.nim
```diff
@@ -529,7 +522,7 @@ proc mountRelay*(node: WakuNode,
 
     info "relay mounted and started successfully"
 
-proc mountLightPush*(node: WakuNode) =
+proc mountLightPush*(node: WakuNode) {.raises: [Defect, LPError].} =
```
staheri14, line 525: Ditto

waku/v2/node/wakunode2.nim
```diff
@@ -541,10 +534,15 @@ proc mountLightPush*(node: WakuNode) =
   
   node.switch.mount(node.wakuLightPush, protocolMatcher(WakuLightPushCodec))
 
-proc mountLibp2pPing*(node: WakuNode) =
+proc mountLibp2pPing*(node: WakuNode) {.raises: [Defect, LPError].} =
```
staheri14, line 537: Ditto

waku/v2/protocol/waku_store/waku_store.nim
```diff
@@ -396,7 +396,7 @@ proc init*(ws: WakuStore) {.raises: [Defect, Exception]} =
 
 
 proc init*(T: type WakuStore, peerManager: PeerManager, rng: ref BrHmacDrbgContext,
-                   store: MessageStore = nil, wakuSwap: WakuSwap = nil, persistMessages = true): T {.raises: [Defect, Exception]} =
```
staheri14, line 399: Why is this removed?

staheri14 (commented):
Looks good to me! Just had a few questions :)

waku/v2/node/storage/peer/waku_peer_storage.nim
staheri14: I have a general question regarding the use of `Result` but related to the files in this PR, if you check line https://github.com/status-im/nim-waku/blob/37c5a6eb222921375a2f9bcedc9133b2c4b1b705/waku/v2/node/storage/peer/waku_peer_storage.nim#L23, the output is defined as `ProtoResult[T]` however, this proc never results in an error, why do we need the Result type? I am going over the waku protocol modules and was wondering whether we should remove such `Result` usage or not?

waku/v2/node/storage/peer/waku_peer_storage.nim
oskarth: I think it isn't a must but in general it is prudent as it helps us. Public functions more about providing clean interface to consumer with clear error paths. At least that's how I understand.

waku/v2/node/wakunode2.nim
```diff
@@ -155,7 +144,7 @@ proc init*(T: type WakuNode, nodeKey: crypto.PrivateKey,
   #    msgIdProvider = msgIdProvider,
   #    triggerSelf = true, sign = false,
   #    verifySignature = false).PubSub
-  result = WakuNode(
+  return WakuNode(
```
oskarth, line 147: Nitpick but I'd probably first init it then `return wakuNode`. Just because we are creating a new peer manager etc, less stuff going on in one expression/statement

waku/v2/node/wakunode2.nim
```diff
@@ -393,9 +382,9 @@ proc info*(node: WakuNode): WakuInfo =
   let wakuInfo = WakuInfo(listenStr: listenStr)
   return wakuInfo
 
-proc mountFilter*(node: WakuNode) =
+proc mountFilter*(node: WakuNode) {.raises: [Defect, KeyError, LPError]} =
   info "mounting filter"
-  proc filterHandler(requestId: string, msg: MessagePush) {.gcsafe.} =
+  proc filterHandler(requestId: string, msg: MessagePush) {.gcsafe, raises: [Defect, KeyError].} =
```
oskarth, line 387: Nitpick but if line is too long might want to pull down whole pragma on newline

waku/v2/node/wakunode2.nim
```diff
@@ -478,7 +470,8 @@ proc mountRelay*(node: WakuNode,
                  topics: seq[string] = newSeq[string](),
                  rlnRelayEnabled = false,
                  relayMessages = true,
-                 triggerSelf = true) {.gcsafe.} =
+                 triggerSelf = true)
+                 {.gcsafe, raises: [Defect, InitializationError, LPError, CatchableError].} = # @TODO: Better error handling: CatchableError is raised by `waitFor`
```
oskarth, line 474: This form is a bit hard long and hard to read, is it possible to format at different somehow?

oskarth (approved):
Thanks!

waku/v2/node/peer_manager/peer_manager.nim
jm-clius: True, but also because the underlying exception is now caught and handled elsewhere in the `peer_manager` and therefore not raised by `addPeer`.

waku/v2/node/storage/peer/waku_peer_storage.nim
jm-clius: Yeah, I don't have strong feelings about this, but I think `Result` is also prudent to use where we can imagine multiple failure paths in future. Encoding/decoding is a good example, because this can often fail in multiple ways. I'd rather elevate more return types to `Result` than remove `Result` for functions where we don't return `err` yet. There's also the consistency of other protobuf `init*` functions returning `ProtoResult`, and we probably want to maintain this pattern. Another reason to use `Result` above `Opt`, for example, is that you can be explicit about what caused the error.

waku/v2/node/storage/peer/waku_peer_storage.nim
jm-clius: In some cases, where non public procs are used by public procs, it becomes necessary to add explicit `{.raises.}`.
In this instance, for example, we can be explicit about what `peer()` raises and then catch and handle the exceptions [here](https://github.com/status-im/nim-waku/blob/37c5a6eb222921375a2f9bcedc9133b2c4b1b705/waku/v2/node/storage/peer/waku_peer_storage.nim#L139-L142). This leaves the public proc free of exceptions.

waku/v2/node/wakunode2.nim
jm-clius: The top-level gets _replaced_ by the function specific raises `{.raises: [Defect, LPError].}`. https://status-im.github.io/nim-style-guide/errors.exceptions.html. `{.raises: [Defect]}` is too strict for this function which raises an `LPError` too, hence it needs its own annotation.

waku/v2/node/wakunode2.nim
jm-clius: See above

waku/v2/protocol/waku_store/waku_store.nim
jm-clius: In general, raising the non-specific `Exception` is bad practice. Here it was done because the underlying modules weren't explicit about what they're raising. Now that the underlying modules have better error-handling, the top-level `{.push raises: [Defect].}` is enough for this function.

waku/v2/node/wakunode2.nim
jm-clius: I've moved the comment to above the pragma line and the pragma to a single indent inside the proc block [here](https://github.com/status-im/nim-waku/blob/47981cfdf527fed4728462794e5e41d381c2a2b9/waku/v2/node/wakunode2.nim#L478-L479). Let me know if this looks reasonable to you.

waku/v2/node/wakunode2.nim
jm-clius: Done in 47981cf

waku/v2/node/wakunode2.nim
jm-clius: Done. 47981cf

waku/v2/node/peer_manager/peer_manager.nim
staheri14: Thanks for the clarification!

waku/v2/node/storage/peer/waku_peer_storage.nim
staheri14: Thanks for the reply! 
What do you refer to by `Opt`, is it `Option`?

waku/v2/node/storage/peer/waku_peer_storage.nim
staheri14: I see, I could not find any exact item on the track1 checklist for this one, good to know! makes sense, I will consider it while refactoring other modules.

staheri14 (approved):
Thanks!

waku/v2/node/storage/peer/waku_peer_storage.nim
jm-clius: Indeed!

PR #665, opened 2021-07-13
Waku store: explicit return instead of result

staheri14:
It is an incremental PR towards the checklist of track1 of the refactor week https://hackmd.io/1imOGULZRsed2HpgmzGleA?view
All the `result`s are replaced with explicit `return`s.

In the follow-up PRs, will cover other protocols.


oskarth (approved):

jm-clius (approved):
Thanks!

PR #666, opened 2021-07-14
Add GH project board automation

oskarth:
Now we have a project board for nim-waku, no longer cross-org, so should be simpler. Based on js-waku workflow.

### DIFF
--- a/.github/workflows/add-action-project.yml
+++ b/.github/workflows/add-action-project.yml
@@ -1,0 +1,26 @@
+name: Add new issues to nim-waku project board
+
+on:
+  issues:
+    types: [opened]
+  pull_request:
+    types: [opened]
+
+jobs:
+  add-new-issue-to-new-column:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Add issue to column "New"
+        if: ${{ !!github.event.issue }}
+        uses: alex-page/github-project-automation-plus@v0.6.0
+        with:
+          project: nim-waku
+          column: New
+          repo-token: ${{ secrets.GH_ACTION_PROJECT_MGMT }}
+      - name: Add pull request to column "In Progress"
+        if: ${{ !!github.event.pull_request }}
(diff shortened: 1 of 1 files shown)


.github/workflows/add-action-project.yml
```diff
@@ -0,0 +1,26 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Add issue to column "New"
+        if: ${{ !!github.event.issue }}
+        uses: alex-page/github-project-automation-plus@v0.6.0
+        with:
+          project: nim-waku
+          column: New
+          repo-token: ${{ secrets.GH_ACTION_PROJECT_MGMT }}
+      - name: Add pull request to column "In Progress"
```
jm-clius, line 20: Not sure why "In Progress" is not working - perhaps difference in capitalisation?

jm-clius (approved):
LGTM, except for it not yet working :D. Think it's a minor change though.

D4nte:
> LGTM, except for it not yet working :D. Think it's a minor change though.

It needs to be merged to work

.github/workflows/add-action-project.yml
```diff
@@ -0,0 +1,26 @@
...
+        if: ${{ !!github.event.pull_request }}
+        uses: alex-page/github-project-automation-plus@v0.6.0
+        with:
+          project: nim-waku
+          column: In Progress
```
D4nte, line 25:
```suggestion
          column: In progress
```
I'd recommend to use exact same capitalization.

.github/workflows/add-action-project.yml
oskarth: Fixed, also changed around a bit

.github/workflows/add-action-project.yml
oskarth: Indeed, fixed and also changed a bit

PR #667, opened 2021-07-14
Refactor waku protocols and utils

staheri14:
This PR is progress towards 
- https://github.com/status-im/nim-waku/issues/639
-  https://github.com/status-im/nim-waku/issues/287.

I have reviewed all the modules under `waku/v2/protocol` and `waku/v2/utils` and made sure they comply with the following items of [nim-waku refactor week: Track 1 checklist](https://hackmd.io/1imOGULZRsed2HpgmzGleA?view).

- [x] Top-level {.push raises: [Defect].} (except for the Swap module, it throws some exceptions which I needed more time to understand their hierarchy and resolve them properly)
- [x] No result returns
- [x] logscope defined for each module that uses chronicles logging
- [x] new() instead of init() for constructors of ref object types

**Next**
Will focus on cleaning imports and better exception handling

PR #671, opened 2021-07-15
General refactoring: `nim-waku`

jm-clius:
This PR refactors `nim-waku` according to [Track 1 refactoring checklist](https://hackmd.io/1imOGULZRsed2HpgmzGleA?view)

It covers all modules not addressed in the existing/closed PRs: https://github.com/status-im/nim-waku/pull/664 #667.
Not everything on the Track 1 checklist is necessary for modules in `/test`-directory, so I've mostly focused on `import` cleanup.

Once all Track 1 refactoring PRs are merged, another pass of code maintenance (especially `import` cleanups) will be necessary to account for the effect of new `export`s and other improvements.

### DIFF
--- a/examples/v2/chat2.nim
+++ b/examples/v2/chat2.nim
@@ -4,6 +4,8 @@
 when not(compileOption("threads")):
   {.fatal: "Please, compile this program with the --threads:on option!".}
 
+{.push raises: [Defect].}
+
 import std/[tables, strformat, strutils, times, httpclient, json, sequtils, random, options]
 import confutils, chronicles, chronos, stew/shims/net as stewNet,
        eth/keys, bearssl, stew/[byteutils, endians2],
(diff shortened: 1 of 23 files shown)


examples/v2/chat2.nim
```diff
@@ -156,16 +154,23 @@ proc printReceivedMessage(c: Chat, msg: WakuMessage) {.raises: [Defect].} =
     trace "Printing message", topic=DefaultTopic, chatLine,
       contentTopic = msg.contentTopic
 
-proc selectRandomNode(fleetStr: string): string =
+proc selectRandomNode(fleetStr: string): Option[string] =
```
oskarth, line 157: If a select operation fails, should it not be a Result type that can error?

Alternatively, if we are getting it from a list and the list happens to be empty, then Option makes sense.

Here it seems like the empty case is a HTTP exception, which seems more like an error to me. Semantics though...

oskarth: With error the exception message can be propagated upward too, such as if it is a permission issue, server down, change of format for fleets json, etc.

examples/v2/chat2.nim
```diff
@@ -295,6 +300,7 @@ proc readInput(wfd: AsyncFD) {.thread.} =
     let line = stdin.readLine()
     discard waitFor transp.write(line & "\r\n")
 
+{.pop.} # @TODO confutils.nim(775, 17) Error: can raise an unlisted exception: ref IOError
```
oskarth, line 303: Oh til

oskarth (approved):

examples/v2/chat2.nim
jm-clius: Agree. `Result` seems more idiomatic here.

examples/v2/chat2.nim
jm-clius: Yeah, it's a life-saver, as it's a pain to get around the `confutils` issue otherwise (read: I don't know how :D)

PR #673, opened 2021-07-15
Cleaning imports of the waku store module

staheri14:
Addresses the following for the waku_store module. 
Partially addresses https://github.com/status-im/nim-waku/issues/639.

- [x] export all modules whose types are used in public functions/types
- [x] use std/ prefix for any import from stdlib
- [x] use explicit import paths (e.g. ./ for modules in same path)
- [X] judiciously, but enthusiastically, clean up unused/redundant imports (e.g. if underlying modules start using export there will be redundant imports).

May need to clean up further if circular imports happen.

**Next:**
Working on the other modules imports/exports.

### DIFF
--- a/waku/v2/protocol/waku_store/waku_store.nim
+++ b/waku/v2/protocol/waku_store/waku_store.nim
@@ -17,7 +17,7 @@ import
   libp2p/protobuf/minprotobuf,
   libp2p/stream/connection,
   metrics,
-  stew/[results, byteutils, endians2],
+  stew/[results, byteutils],
   # internal imports
   ../../node/storage/message/message_store,
   ../../node/peer_manager/peer_manager,
--- a/waku/v2/protocol/waku_store/waku_store.nim
+++ b/waku/v2/protocol/waku_store/waku_store.nim
@@ -4,21 +4,36 @@
 
 {.push raises: [Defect].}
 
+# Group by std, external then internal imports
 import
+  # std imports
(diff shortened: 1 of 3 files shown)


oskarth (approved):
Nice!

waku/v2/protocol/waku_store/waku_store.nim
```diff
@@ -4,21 +4,36 @@
...
 
-export waku_store_types
+# export all modules whose types are used in public functions/types
+export 
+  options,
+  chronos,
+  bearssl,
+  minprotobuf,
+  peer_manager,
+  waku_store_types
```
jm-clius, line 36: I think we can include `message_store` in the exports as well?

jm-clius (approved):
Thanks, @staheri14. You may get a merge conflict with this line in my changes: https://github.com/status-im/nim-waku/blob/6da8083e968604a3108b3fac219872a583100646/waku/v2/protocol/waku_store/waku_store.nim#L21
If you discard my change and just add `messsage_store` to your `export`s there should be no other issues.

waku/v2/protocol/waku_store/waku_store.nim
staheri14: Right, done!

staheri14: Sure, done!

PR #675, opened 2021-07-16
Refactor wakunode2 setup

### LINKED ISSUE #581: Improvement: `wakunode2` setup
### Problem

Currently the setup procedures for a `wakunode2` is a complex combination of reading configuration, initialising, mounting protocols, starting servers, and preparing to participate actively in the network by resuming history, starting to relay messages, subscribing to default and configured topics, etc.
Many of these procedures have a fragile dependency on other procedures being done first. For example, the relay protocol has to be mounted before a store node can attempt to resume history after setup. Currently, that would mean that a node starts relaying (and storing) messages _before_ resuming history, which may not be ideal. It should be possible to mount relay without actually starting to relay messages, but this is currently tied in with starting the switch, which _is_ required to resume history.
This intricate web of dependencies is prone to bugs and need to be untangled. See https://github.com/status-im/nim-waku/issues/577

### Suggested solution

Some requirements on ordering are unavoidable, such as `relay` having to be mounted before other protocols. It should be possible, though, to separate protocol mounting behaviour (passive i.t.o. network participation, e.g. initialise and mount protocol on switch) and protocol starting behaviour (often active, e.g. resume history, subscribe to default topics, etc.) in the setup procedures. The latter should be deferred until all initialisation and configuration has been done.

jm-clius:
This PR closes #581

It forms part of Track 2 of the [refactor week](https://hackmd.io/beMbu_fdRymYvaXDANukIw)

### Background

The setup procedure when running a `wakunode2` binary is currently extremely brittle and convoluted. It is a complex combination of reading configuration, initialising, mounting protocols, starting servers, and preparing to participate actively in the network by resuming history, starting to relay messages, subscribing to default and configured topics, etc. Regression bugs are often introduced whenever a step needs to be added to the setup procedures.

### Solution

This PR follows the following approach as an attempted solution:
- clearer separation of concerns, document in code
- as far as possible, separate node initialisation (passive i.t.o. network participation, e.g. initialise and mount protocol on switch) from node starting behaviour (often active, e.g. resume history, subscribe to default topics, etc.)
- defer node startup until all initialisation and configuration has been done.

As such node setup is divided into 6 phases:
1. Set up storage
2. Initialise node
3. Mount and initialise configured protocols
4. Setup graceful shutdown hooks
5. Start node and mounted protocols
6. Start monitoring tools and external interfaces

> NOTE: This PR is marked `Draft` as I want to do more comprehensive testing, but review comments will be appreciated.

### DIFF
--- a/waku/v2/node/waku_setup.nim
+++ b/waku/v2/node/waku_setup.nim
@@ -28,7 +28,7 @@ logScope:
   topics = "wakunode.setup"
 
 type
-  SetupResult*[T] = Result[WakuNode, string]
+  SetupResult*[T] = Result[T, string]
 
 ##########################
 # Setup helper functions #
--- a/waku/v2/node/waku_setup.nim
+++ b/waku/v2/node/waku_setup.nim
@@ -65,10 +65,13 @@ proc startRpc*(node: WakuNode, rpcIp: ValidIpAddress, rpcPort: Port, conf: WakuN
   rpcServer.start()
   info "RPC Server started", ta
 
-proc startMetricsServer*(serverIp: ValidIpAddress, serverPort: Port) {.raises: [Defect, Exception].} =
+proc startMetricsServer*(serverIp: ValidIpAddress, serverPort: Port) =
     info "Starting metrics HTTP server", serverIp, serverPort
     
-    startMetricsHttpServer($serverIp, serverPort)
+    try:
+      startMetricsHttpServer($serverIp, serverPort)
+    except Exception as e:
+      raiseAssert("Exception while starting metrics HTTP server: " & e.msg)
 
     info "Metrics HTTP server started", serverIp, serverPort
 
--- a/waku/v2/node/waku_setup.nim
+++ b/waku/v2/node/waku_setup.nim
@@ -1,0 +1,112 @@
+{.push raises: [Defect].}
+
+## Collection of utilities commonly used
+## during the setup phase of a Waku v2 node
+
+import
+  std/tables,
+  chronos,
+  chronicles,
+  json_rpc/rpcserver,
+  metrics,
+  metrics/chronos_httpserver,
+  stew/results,
+  stew/shims/net,
+  ./storage/sqlite,
+  ./storage/migration/migration_types,
+  ./jsonrpc/[admin_api,
+             debug_api,
+             filter_api,
+             relay_api,
+             store_api,
+             private_api,
+             debug_api],
+  ./config,
+  ./wakunode2
+
+logScope:
+  topics = "wakunode.setup"
+
+type
+  SetupResult*[T] = Result[WakuNode, string]
+
+##########################
+# Setup helper functions #
+##########################
+
+proc startRpc*(node: WakuNode, rpcIp: ValidIpAddress, rpcPort: Port, conf: WakuNodeConf)
+  {.raises: [Defect, RpcBindError, CatchableError].} =
+  # @TODO: API handlers still raise CatchableError
+
+  let
+    ta = initTAddress(rpcIp, rpcPort)
+    rpcServer = newRpcHttpServer([ta])
+  installDebugApiHandlers(node, rpcServer)
+
+  # Install enabled API handlers:
+  if conf.relay:
+    let topicCache = newTable[string, seq[WakuMessage]]()
+    installRelayApiHandlers(node, rpcServer, topicCache)
+    if conf.rpcPrivate:
+      # Private API access allows WakuRelay functionality that 
+      # is backwards compatible with Waku v1.
+      installPrivateApiHandlers(node, rpcServer, node.rng, topicCache)
+  
+  if conf.filter:
+    let messageCache = newTable[ContentTopic, seq[WakuMessage]]()
+    installFilterApiHandlers(node, rpcServer, messageCache)
+  
+  if conf.store:
+    installStoreApiHandlers(node, rpcServer)
+  
+  if conf.rpcAdmin:
+    installAdminApiHandlers(node, rpcServer)
+  
+  rpcServer.start()
+  info "RPC Server started", ta
+
+proc startMetricsServer*(serverIp: ValidIpAddress, serverPort: Port) {.raises: [Defect, Exception].} =
+    info "Starting metrics HTTP server", serverIp, serverPort
+    
+    startMetricsHttpServer($serverIp, serverPort)
+
+    info "Metrics HTTP server started", serverIp, serverPort
+
+proc startMetricsLog*() =
+  # https://github.com/nim-lang/Nim/issues/17369
+  var logMetrics: proc(udata: pointer) {.gcsafe, raises: [Defect].}
+  logMetrics = proc(udata: pointer) =
+    {.gcsafe.}:
+      # TODO: libp2p_pubsub_peers is not public, so we need to make this either
+      # public in libp2p or do our own peer counting after all.
+      var
+        totalMessages = 0.float64
+
+      for key in waku_node_messages.metrics.keys():
+        try:
+          totalMessages = totalMessages + waku_node_messages.value(key)
+        except KeyError:
+          discard
+
+    info "Node metrics", totalMessages
+    discard setTimer(Moment.fromNow(2.seconds), logMetrics)
+  discard setTimer(Moment.fromNow(2.seconds), logMetrics)
+  
+proc runMigrations*(sqliteDatabase: SqliteDatabase, conf: WakuNodeConf) =
+  # Run migration scripts on persistent storage
+
+  var migrationPath: string
+  if conf.persistPeers and conf.persistMessages:
+    migrationPath = migration_types.ALL_STORE_MIGRATION_PATH
+  elif conf.persistPeers:
+    migrationPath = migration_types.PEER_STORE_MIGRATION_PATH
+  elif conf.persistMessages:
+    migrationPath = migration_types.MESSAGE_STORE_MIGRATION_PATH
+
+  # run migration 
+  info "running migration ... "
+  let migrationResult = sqliteDatabase.migrate(migrationPath)
+  if migrationResult.isErr:
+    warn "migration failed"
+  else:
+    info "migration is done"
--- a/waku/v2/node/wakunode2.nim
+++ b/waku/v2/node/wakunode2.nim
@@ -453,9 +453,7 @@ proc startRelay*(node: WakuNode) {.async.} =
   info "starting relay"
   
   # Topic subscriptions
-  node.subscribe(defaultTopic, none(TopicHandler))
-
-  for topic in node.wakuRelay.confTopics:
+  for topic in node.wakuRelay.defaultTopics:
     node.subscribe(topic, none(TopicHandler))
 
   # Resume previous relay connections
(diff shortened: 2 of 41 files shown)


waku/v2/node/wakunode2.nim
```diff
@@ -476,6 +445,38 @@ when defined(rln):
...
+
+  for topic in node.wakuRelay.confTopics:
+    node.subscribe(topic, none(TopicHandler))
+
+  # Resume previous relay connections
+  if node.peerManager.hasPeers(WakuRelayCodec):
+    info "Found previous WakuRelay peers. Reconnecting."
+    # Reconnect to previous relay peers. This will respect a backoff period, if necessary
+    await node.peerManager.reconnectPeers(WakuRelayCodec,
+                                          node.wakuRelay.parameters.pruneBackoff + chronos.seconds(BackoffSlackTime))
```
oskarth, line 467: I'd put this into a variable above

waku/v2/protocol/waku_relay.nim
```diff
@@ -19,9 +19,12 @@ const
 
 type
   WakuRelay* = ref object of GossipSub
+    confTopics*: seq[string] # Configured PubSub topics
```
oskarth, line 22: Conf seems like a concept that exists before we get here, maybe something like activeTopics? defaultTopics? topics?

oskarth (commented):
I like the cleaner separation! I wonder if we can split things up even further, in terms of:

1) Moving util functions to separate module (1000 LOCs here)
2) Making each of the 6 "stages" be a logical separate piece of code, so basically
- each step is a function call (no need for comments
- at each step we either get a node/equivalent with more stuff done to it, or some form of error

oskarth:
Basically:

```
setupStorage(node, ...)
init(node, ...)
setupProtocols(node, ...)
setupHooks(node, ...)
setupExternal(node, ...)
```

Or something like this, with a Result type per each fn (where applicable).

In Clojure this type of pipeline would idiomatically look something like:

```
(-> ;; or err-> or eqv to create a result-error pipeline
     node
     (setupStorage db-conf)
     init!
     (setupProtocols conf)
     setupHooks
     (setupExternal conf stdout))
```

Probably the same isn't directly possible in Nim without needless complications, but the general gist. It makes intent of code, adding and changing stages, debugging, etc even easier.

waku/v2/protocol/waku_relay.nim
jm-clius: Yeah, perhaps `defaultTopics`! Can then add the `/default-waku/` to the same list and treat all of these equivalently.

jm-clius:
Thanks for the comments! Agree re taking this a step further i.t.o separation.

waku/v2/protocol/waku_relay.nim
jm-clius: Done in 7bc5d0d

jm-clius:
`wakunode2` setup is now divided into 6 phases, utility functions have been split off into a separate module and setup logging has ben clarified. I have tested configuring the binary in a local fleet-like setup.

Example of setup logging with the phased approach:
image

waku/v2/node/wakunode2.nim
```diff
@@ -495,44 +496,34 @@ proc mountRelay*(node: WakuNode,
...
     ## This is to allow protocol stream negotation with relay-capable nodes to succeed.
     ## Here we mount relay on the switch only, but do not proceed to subscribe to any pubsub
     ## topics. We also never start the relay protocol. node.wakuRelay remains nil.
-    ## @TODO: in future, this WakuRelay dependency will be removed completely
-    return
-
-  node.wakuRelay = wakuRelay
-
-  node.subscribe(defaultTopic, none(TopicHandler))
+    ## @TODO: in future, this WakuRelay dependency will be removed completely
```
oskarth, line 512: Not sure I understand this comment

waku/v2/node/wakunode2.nim
```diff
@@ -648,195 +639,252 @@ proc connectToNodes*(n: WakuNode, nodes: seq[PeerInfo]) {.async.} =
...
-    mountFilter(node)
+  let (pStorage, mStorage) = setupStorage(conf)
 
-    if conf.filternode != "":
-      setFilterPeer(node, conf.filternode)
+  debug "2/6 Initialising node"
 
-  if conf.rpc:
-    startRpc(node, conf.rpcAddress, Port(conf.rpcPort + conf.portsShift), conf)
+  node = initialiseNode(conf, pStorage).tryGet()
```
oskarth, line 869: I'd call this initNode, also initialise/initialize easy to get wrong (I usually prefer initialize, AmE)

waku/v2/node/wakunode2.nim
```diff
@@ -648,195 +639,252 @@ proc connectToNodes*(n: WakuNode, nodes: seq[PeerInfo]) {.async.} =
...
-    mountLightPush(node)
+  debug "1/6 Setting up storage"
 
-    if conf.lightpushnode != "":
-      setLightPushPeer(node, conf.lightpushnode)
-  
-  # Filter setup. NOTE Must be mounted after relay
-  if (conf.filternode != "") or (conf.filter):
-    mountFilter(node)
+  let (pStorage, mStorage) = setupStorage(conf)
```
oskarth, line 865: What happens if this fails? Would it make sense to have each stage return Result and exit with error if any fails?

waku/v2/node/wakunode2.nim
```diff
@@ -648,195 +639,252 @@ proc connectToNodes*(n: WakuNode, nodes: seq[PeerInfo]) {.async.} =
...
 
-  if conf.metricsServer:
-    startMetricsServer(conf.metricsServerAddress,
-      Port(conf.metricsServerPort + conf.portsShift))
+  setupProtocols(node, conf, mStorage)
+
+  debug "4/6 Starting node and mounted protocols"
   
-  # Setup graceful shutdown
+  startNode(node, conf)
```
oskarth, line 877: Ditto re failure path

waku/v2/node/wakunode2.nim
```diff
@@ -648,195 +639,252 @@ proc connectToNodes*(n: WakuNode, nodes: seq[PeerInfo]) {.async.} =
...
+
+  debug "5/6 Starting monitoring and external interfaces"
+
+  startExternal(node, conf)
```
oskarth, line 881: Ditto re failure path

waku/v2/node/wakunode2.nim
```diff
@@ -648,195 +639,252 @@ proc connectToNodes*(n: WakuNode, nodes: seq[PeerInfo]) {.async.} =
...
-        warn "failed to init WakuMessageStore", err = res.error
-        waku_node_errors.inc(labelValues = ["init_store_failure"])
-      else:
-        store = res.value
+      if conf.filternode != "":
+        setFilterPeer(node, conf.filternode)
 
-    mountStore(node, store, conf.persistMessages)
+  # 4/6 Start node and mounted protocols
+  proc startNode*(node: WakuNode, conf: WakuNodeConf) =
```
oskarth, line 817: Since these (and other functions) are external functions, might want to update Nim Node API https://github.com/status-im/nim-waku/blob/master/docs/api/v2/node.md

Also, CHANGELOG!

oskarth (commented):
Nice work, looking much cleaner! Some comments

waku/v2/node/wakunode2.nim
jm-clius: > Since these (and other functions) are external functions, might want to update Nim Node API

Ah, they shouldn't be public. Thanks! Had them in external module first (which caused a 100 issues) and forgot to remove the `*` after moving them back. Changed in 4cca119

> Also, CHANGELOG!

Done in 4cca119

waku/v2/node/wakunode2.nim
jm-clius: Yeah, I've changed it now that each setup phase returns a `SetupResult`. Most of these functions, however, do too many things to have clear failure paths (e.g. it is possible that `setupProtocols` may fail to mount swap, but mount relay without any problems, etc.). This can obviously still be improved in future, but for now I've tried to keep error handling simple-ish. Done in 4cca119.

waku/v2/node/wakunode2.nim
jm-clius: 😄 My inner South African schooling system (BrE) is somewhat pained, but I'm all for consistency! Changed to `init` in 4cca119 and _standardized_ (😰) spelling to AmE in comments.

waku/v2/node/wakunode2.nim
jm-clius: Old comment indicating that we have a dependency on `relay` being mounted for any protocol to work. Hopefully can remove this dependency in future with protocol (peer) management.

waku/v2/node/wakunode2.nim
jm-clius: 4cca119

waku/v2/node/wakunode2.nim
jm-clius: 4cca119

waku/v2/node/waku_setup.nim
```diff
@@ -0,0 +1,112 @@
...
+  if conf.rpcAdmin:
+    installAdminApiHandlers(node, rpcServer)
+  
+  rpcServer.start()
+  info "RPC Server started", ta
+
+proc startMetricsServer*(serverIp: ValidIpAddress, serverPort: Port) {.raises: [Defect, Exception].} =
+    info "Starting metrics HTTP server", serverIp, serverPort
+    
+    startMetricsHttpServer($serverIp, serverPort)
```
kdeme, line 71: It doesn't matter that much but I guess it is cleaner to try/except the `Exception` here and raiseAssert in case it is not a Defect

kdeme (approved):

waku/v2/node/waku_setup.nim
jm-clius: Thanks! Done in 80b71d1

waku/v2/node/wakunode2.nim
```diff
@@ -652,195 +643,283 @@ proc connectToNodes*(n: WakuNode, nodes: seq[PeerInfo]) {.async.} =
...
+  
+  logScope:
+    topics = "wakunode.setup"
+  
+  ###################
+  # Setup functions #
+  ###################
+
+  # 1/6 Setup storage
+  proc setupStorage(conf: WakuNodeConf):
```
staheri14, line 702: Wondering whether it can be moved to waku_setup?

waku/v2/node/waku_setup.nim
```diff
@@ -0,0 +1,115 @@
+{.push raises: [Defect].}
```
staheri14, line 1: I'd suggest using a different name like `waku_setup_utils` or `waku_setup_helpers`  to indicate this module contains the helpers, not the setup functions

staheri14 (approved):
Very Nice!

oskarth (approved):
Nice work!

waku/v2/node/wakunode2.nim
jm-clius: I experimented with moving some of the phases (and not just the utilities) to `waku_setup`, but found that it caused some issues, including confusing circular dependencies. `setupStorage` _is_ a phase that could be moved, but that would impede readability somewhat (as it will be the only setup phase not in `wakunode2.nim`). WDYT? Going to merge as-is for now, but it should be easy to move things around in future.

PR #676, opened 2021-07-16
Fixes the order of fields of History Response protobuf

staheri14:
This is to fix the mismatch between the implementation and the store specs regarding the order of fields of History Response protobuf. 

The store specs indicate that the first field of `HistoryResponse` is reserved for future use (as an identifier to match up queries with their corresponding requests). However, the nim-waku implementation does not comply with this specification.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ This release contains the following:
 - Support for stable version of `relay` protocol, with protocol ID `/vac/waku/relay/2.0.0`
 - Support for multiple protocol IDs - now matches any protocol that adds postfix to stable ID.
 - Updates the order of fields of `HistoryResponse` protobuf message. 
-  The filed numbers are shifted up by one to match up the [13/WAKU2-STORE](https://rfc.vac.dev/spec/13/) specs. 
+  The filed numbers of `HistoryResponse` are shifted up by one to match up the [13/WAKU2-STORE](https://rfc.vac.dev/spec/13/) specs. 
 
 #### General refactoring
 #### Docs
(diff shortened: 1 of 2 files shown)


oskarth (approved):
Thanks!

Worth adding to changelog too

jm-clius (commented):
Thanks! LGTM, except that I think this is a breaking change? How is this currently implemented in the other clients?

staheri14:
> Thanks! LGTM, except that I think this is a breaking change? How is this currently implemented in the other clients?

Not exactly a breaking change, because the specs were correct, it was only an issue on the nim-waku side which was not aligned with the specs.  If other clients are based on the specs, they won't have to undertake any updates, as the specs were correct. However, I am not sure how the interaction with the nim-waku was working previously!

jm-clius:
> However, I am not sure how the interaction with the nim-waku was working previously!

Yeah, this is what worries me - if the other clients were working, they might have implemented it (probably did) based on the `nim-waku` implementation and not the spec.
In this case it would be "breaking" inasmuch as other clients won't work with store protocol anymore. May need to coordinate the change with them, then. Let's discuss in PM. :)

staheri14:
> > However, I am not sure how the interaction with the nim-waku was working previously!
> 
> Yeah, this is what worries me - if the other clients were working, they might have implemented it (probably did) based on the `nim-waku` implementation and not the spec.
> In this case it would be "breaking" inasmuch as other clients won't work with store protocol anymore. May need to coordinate the change with them, then. Let's discuss in PM. :)

Right, I think that their implementation matches nim-waku not the specs, Lets discuss it in the call

staheri14:
> Thanks!
> 
> Worth adding to changelog too

Changelog has been updated!

staheri14:
> Thanks! LGTM, except that I think this is a breaking change? How is this currently implemented in the other clients?

Checked with @D4nte  and @richard-ramos, the go and js versions are also based on the nim implementation, not the specs. 
I am going to push the changes to the master and inform them to make the necessary changes.
There should be no concern with walletConnect as well, afaik they also interact with the test fleets which is based on the updated master. 
I think the only concern is re prod fleets that are not synced with the master, those are also going to have this change incorporated after v 5.0. 
@jm-clius  please let me know if there is anyone else that I should check with.

PR #678, opened 2021-07-19
Whisper tests

oskarth:
Follow up to https://github.com/status-im/nim-waku/pull/663 re comments

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -82,6 +82,11 @@ update: | update-common
 
 # a phony target, because teaching `make` how to do conditional recompilation of Nim projects is too complicated
 
+# Whisper tests
+testwhisper: | build deps
+	echo -e $(BUILD_MSG) "build/$@" && \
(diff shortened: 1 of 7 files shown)


kdeme (approved):

Makefile
```diff
@@ -82,6 +82,11 @@ update: | update-common
 
 # a phony target, because teaching `make` how to do conditional recompilation of Nim projects is too complicated
 
+# Whisper tests
+testwhisper: | build deps
```
jm-clius, line 86: I'm assuming this is purposefully left out of the `test` target?

jm-clius (approved):
LGTM

Makefile
oskarth: Yeah don't really see a need to have it there, will just prolong CI. It passes locally now and is basically identical to waku1, which is what we care about here I think

PR #681, opened 2021-07-20
Adds timestamp to waku relay message

staheri14:
This PR closes https://github.com/status-im/nim-waku/issues/674. 

For more context please see https://github.com/status-im/nim-waku/issues/674#issuecomment-882812404 and https://github.com/status-im/nim-waku/issues/674#issuecomment-882814937.

I also used this PR to tide up the imports in  quicksim2 module as part of refactoring week.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ This release contains the following:
 - Support for multiple protocol IDs - now matches any protocol that adds postfix to stable ID.
 - Updates the order of fields of `HistoryResponse` protobuf message. 
   The filed numbers of the `HistoryResponse` are shifted up by one to match up the [13/WAKU2-STORE](https://rfc.vac.dev/spec/13/) specs. 
-
+- Adds optional `timestamp` to `WakuRelayMessage`.
 #### General refactoring
 #### Docs
 - Adds the database migration tutorial.
--- a/docs/tutorial/jsonrpc-api.md
+++ b/docs/tutorial/jsonrpc-api.md
@@ -109,7 +109,8 @@ To publish a message to a pubsub topic, call the [`post_waku_v2_relay_v1_message
   "params": [
     "my_topic_1",
     {
-      "payload": "0x1a2b3c4d5e6f"
+      "payload": "0x1a2b3c4d5e6f",
+      "timestamp": 1626813243.916377
     }
   ]
 }
--- a/tests/v2/test_jsonrpc_waku.nim
+++ b/tests/v2/test_jsonrpc_waku.nim
@@ -1,7 +1,7 @@
 {.used.}
 
 import
-  std/[options, sets, tables, os, strutils, sequtils],
+  std/[options, sets, tables, os, strutils, sequtils, times],
   testutils/unittests, stew/shims/net as stewNet,
   json_rpc/[rpcserver, rpcclient],
   eth/[keys, rlp], eth/common/eth_types,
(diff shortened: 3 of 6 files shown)


oskarth (approved):

jm-clius (approved):
Thanks! Does this need to be added to https://rfc.vac.dev/spec/16/#wakurelaymessage?

staheri14:
> Thanks! Does this need to be added to https://rfc.vac.dev/spec/16/#wakurelaymessage?

Right! will do it :)

staheri14:
@jm-clius 
The specs PR https://github.com/vacp2p/rfc/pull/437

CHANGELOG.md
```diff
@@ -13,7 +13,7 @@ This release contains the following:
 - Support for multiple protocol IDs - now matches any protocol that adds postfix to stable ID.
 - Updates the order of fields of `HistoryResponse` protobuf message. 
   The filed numbers of the `HistoryResponse` are shifted up by one to match up the [13/WAKU2-STORE](https://rfc.vac.dev/spec/13/) specs. 
-
+- Adds optional `timestamp` to `WakuRelayMessage`.
```
arnetheduck, line 16: what can the sender do with an invalid timestamp (ie one in the future)?

In general, any functionality that relies on timestamps set by the server is prone to exploit (the classic example is a message that forever stays "latest")

staheri14: @arnetheduck 
I think you mean what can a **receiver** do? not a sender?

Currently, a receiver does not validate the timestamp. One way to do it is that the receiver compares the received message timestamp against its own local time (UNIX epoch) and discards messages with more than +-20 secs gap (or some other upper bound).
A similar technique is used in [Status payload specs](https://specs.status.im/spec/6#clock-vs-timestamp-and-message-ordering) as well. 
Please let me know your thoughts.

arnetheduck: My feeling in general would be that this might not be the best layer for a timestamp since it reveals a bit too much and is prone to abuse in a way that's difficult to reason about, a bit while at the same time being too weak to be a reliable mechanism for actual ordering of "important stuff" - ie when you're just relaying messages, you shouldn't have to care in which order they come - that only comes into play if you successfully decode the message and then presumably, the timestamp can sit inside the encoded message.

That said, I don't have the full background story - seeing the PR just struck me as something unusual and risky.

staheri14: Thanks for the comment @arnetheduck!
- Regarding the first part of your concern, i.e.,
> this might not be the best layer for a timestamp

I think the naming might have caused some confusion. Despite the message is called `WakuRelayMessage`, it corresponds to the [`WakuMessage`](https://rfc.vac.dev/spec/14/) and is different from the actual relay [`Message`](https://rfc.vac.dev/spec/11/#message-fields). It has been also noted in the [`WakuRelayMessage`](https://rfc.vac.dev/spec/16/#wakurelaymessage).

What it means is that the timestamp is not part of the relay `Message` fields but it is encoded in the `WakuMessage` which is carried as the `data` field of the relay `Message`. As such, a relay node typically should not attempt to look at the timestamp as it is an internal field of the `data`.
Nevertheless, I also want to clarify that In the current implementation, while `timestamp` is not part of the relay `Message`, but it is always accessible in plaintext even when the `WakuMessage` is encrypted (encryption only affects the `payload` but not other fields like `timestamp`). As such, a relay node can extract the `timestamp` from the `WakuMessage` if he wants. So there can be still concerns about this.
@oskarth we may want to consider other fields for the encryption as well.

- Now regarding the security issue that can cause by exposing the timestamp to the relay node i.e.,
> it reveals a bit too much and is prone to abuse in a way that's difficult to reason about
I hope you could give more details about the security issues, is it because messages timestamp can allow linking messages published by the same author?

- Also about the following comment
> a bit while at the same time being too weak to be a reliable mechanism for actual ordering of "important stuff" - ie when you're just relaying messages, you shouldn't have to care in which order they come -

`timestamp` is the unix epoch time (not a logical clock like Lamport) and  what I am suggesting is not that a relay node drops messages that are out of order, rather a relay node drops messages whose sender's timestamp has a huge gap with the relay node's current time. For example, if two messages m1 and m2 arrive (m1 comes before m2) and m1.timestamp > m2.timestamp and |m1.timestamp - relaynode.currenttime|<=20s and |m2.timestamp - relaynode.currenttime|<=20s, then the relay node relays them both, it does not care about them coming out of order.

One more thing is that the `timestamp` does not capture the "happen before" / "causality" relation thus does not capture messages dependency. We are still working on that.

Please let me know whether this makes sense and addresses your concern.

PR #682, opened 2021-07-21
Update submodules

jm-clius:
This PR updates `nim-waku` submodules.

This is both a [Track 3 task of refactor week](https://hackmd.io/beMbu_fdRymYvaXDANukIw) and a prerequisite for [upcoming release v0.5](https://github.com/status-im/nim-waku/issues/672).

In addition:
- added submodules `nim-websock` and `nim-zlib`: they are now required by some existing submodules.
- cleaned up overlooked modules according to the [Track 1 checklist](https://hackmd.io/1imOGULZRsed2HpgmzGleA) for refactoring

### DIFF
--- a/waku/v2/node/jsonrpc/debug_api.nim
+++ b/waku/v2/node/jsonrpc/debug_api.nim
@@ -1,3 +1,5 @@
+{.push raises: [Defect, CatchableError].}
+
 import
   chronicles,
   json_rpc/rpcserver,
--- a/.gitmodules
+++ b/.gitmodules
@@ -120,3 +120,13 @@
 	url = https://github.com/status-im/nim-unittest2.git
 	ignore = untracked
 	branch = master
+[submodule "vendor/nim-websock"]
+	path = vendor/nim-websock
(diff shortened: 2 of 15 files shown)


waku/v2/node/jsonrpc/debug_api.nim
```diff
@@ -1,3 +1,5 @@
+{.push raises: [Defect, CatchableError].}
```
staheri14, line 1: `CatchableError` means all types of exceptions are allowed, should not we limit them?

oskarth (approved):

waku/v2/node/jsonrpc/debug_api.nim
jm-clius: Indeed! However, the underlying JSON-RPC library can still raise `CatchableError` and there's no way to isolate this in code (likely because parts are macro-generated), As a result all RPC API modules has a top-level `{.push raises: [Defect, CatchableError].}` so we at least have visibility of this.

PR #687, opened 2021-07-26
Patch release v0.5.1 - fix multiple protocol IDs in persistent storage

jm-clius:
This PR is a patch for [release v0.5](https://github.com/status-im/nim-waku/releases/tag/v0.5)

It adds support for multiple protocol IDs when reconnecting to previous peers in the peer manager.

Although multiple protocol IDs is supported by the protocols as per https://github.com/status-im/nim-waku/pull/652, the peer manager would only reconnect to peers that previously matched the current mounted protocol ID.

The impact is that a fleet system, being upgraded to the stable version of the `relay` protocol, will not reconnect to previously connected peers if they were not previously registered with the stable protocol ID.

This is a blocker for installing the latest v0.5 release on the `prod` fleet, an outstanding task on https://github.com/status-im/nim-waku/issues/672

The issue affects only `nim-waku` fleet clients. It is completely backwards compatible with release v0.5.

### DIFF
--- a/waku/v2/node/peer_manager/peer_manager.nim
+++ b/waku/v2/node/peer_manager/peer_manager.nim
@@ -3,6 +3,7 @@
 import
   std/[options, sets, sequtils, times],
   chronos, chronicles, metrics,
+  libp2p/multistream,
   ./waku_peer_store,
   ../storage/peer/peer_storage
 
@@ -78,8 +79,11 @@ proc dialPeer(pm: PeerManager, peerId: PeerID,
     return none(Connection)
 
 proc loadFromStorage(pm: PeerManager) =
+  debug "loading peers from storage"
   # Load peers from storage, if available
   proc onData(peerId: PeerID, storedInfo: StoredInfo, connectedness: Connectedness, disconnectTime: int64) =
+    trace "loading peer", peerId=peerId, storedInfo=storedInfo, connectedness=connectedness
+
     if peerId == pm.switch.peerInfo.peerId:
       # Do not manage self
       return
@@ -94,6 +98,8 @@ proc loadFromStorage(pm: PeerManager) =
   if res.isErr:
     warn "failed to load peers from storage", err = res.error
     waku_peers_errors.inc(labelValues = ["storage_load_failure"])
+  else:
+    debug "successfully queried peer storage"
   
 ##################
 # Initialisation #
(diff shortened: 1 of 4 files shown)


waku/v2/node/peer_manager/peer_manager.nim
```diff
@@ -140,6 +151,10 @@ proc peers*(pm: PeerManager, proto: string): seq[StoredInfo] =
   # Return the known info for all peers registered on the specified protocol
   pm.peers.filterIt(it.protos.contains(proto))
 
+proc peers*(pm: PeerManager, protocolMatcher: Matcher): seq[StoredInfo] =
+  # Return the known info for all peers matching the provided protocolMatcher
+  pm.peers.filter(proc (storedInfo: StoredInfo): bool = storedInfo.protos.anyIt(protocolMatcher(it)))
```
staheri14, line 156: Probably explicit `return` would be better :)

waku/v2/node/peer_manager/peer_manager.nim
```diff
@@ -162,6 +177,10 @@ proc hasPeers*(pm: PeerManager, proto: string): bool =
   # Returns `true` if manager has any peers for the specified protocol
   pm.peers.anyIt(it.protos.contains(proto))
 
+proc hasPeers*(pm: PeerManager, protocolMatcher: Matcher): bool =
+  # Returns `true` if manager has any peers matching the protocolMatcher
+  pm.peers.any(proc (storedInfo: StoredInfo): bool = storedInfo.protos.anyIt(protocolMatcher(it)))
```
staheri14, line 182: Ditto

waku/v2/node/peer_manager/peer_manager.nim
staheri14: Also, a temporary variable to hold the return result would be beneficial for future debugging

waku/v2/node/wakunode2.nim
```diff
@@ -80,15 +80,16 @@ type
     rng*: ref BrHmacDrbgContext
     started*: bool # Indicates that node has started listening
 
-func protocolMatcher(codec: string): Matcher =
+proc protocolMatcher(codec: string): Matcher =
```
staheri14, line 83: Is this proc going to support more than one codec? If no, then I am wondering why not using simple string comparison?

staheri14 (approved):
LGTM!

oskarth (approved):

PR #694, opened 2021-08-11
Revert PRUNE backoff period to default

jm-clius:
Relates to https://github.com/vacp2p/rfc/pull/451

It reverts a `nim-waku`-specific change to GossipSub parameters that were made earlier.
Specifically, the PRUNE backoff period was set to 30 seconds (50% of the default 1 minute). Although this theoretically shortens the restart-and-reconnect time for clients implementing peer persistence, this is not a good enough reason to deviate from the recommended default and may result in inconsistencies between client implementations.

This PR ensures the `nim-waku` implementation follows the recommendations set out in `29/WAKU2-CONFIG`.

cc @richard-ramos @D4nte to confirm that prune backoff is set to `1 min` in `go` and `js` implementations respectively

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,26 @@
 # Changelog
 
+## Next version
+
+This release contains the following:
+
+### Features
+
+### Changes
+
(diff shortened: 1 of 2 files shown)


D4nte:
Yes, confirmed 1 min for js-waku

D4nte (approved):

richard-ramos:
1 min for go-waku as well

richard-ramos (approved):

oskarth (approved):

PR #722, opened 2021-09-21
Restructuring Rln-Relay mounting process

staheri14:
An increment towards https://github.com/status-im/nim-waku/issues/714

Summary of changes:
- Decouples the rlnrelay mounting process from the relay mount and start procedures, and makes rln the prerequisite of rlnrelay. This will simplify the future devs and debugging process. This update resulted in cleaner interfaces for both relay and rln and enhanced code readability.
- Adds a command-line option to get pubsub topic on which rln-relay will get mounted (instead of the prior version which was static).

### DIFF
--- a/waku/v2/node/config.nim
+++ b/waku/v2/node/config.nim
@@ -76,7 +76,7 @@ type
       name: "rln-relay" }: bool
     
     rlnRelayMemIndex* {.
-      desc: "(experimental) the index of node in the rln-relay group: a value between 0-49 inclusive",
+      desc: "(experimental) the index of node in the rln-relay group: a value between 0-99 inclusive",
       defaultValue: 0
       name: "rln-relay-membership-index" }: uint32
     
--- a/waku/v2/node/config.nim
+++ b/waku/v2/node/config.nim
@@ -78,12 +78,12 @@ type
     
     rlnRelayMemIndex* {.
       desc: "(experimental) the index of node in the rln-relay group: a value between 0-99 inclusive",
-      defaultValue: MembeshipIndex(0)
-      name: "rln-relay-membership-index" }: MembeshipIndex
+      defaultValue: MembershipIndex(0)
+      name: "rln-relay-membership-index" }: MembershipIndex
 
     rlnRelayPubsubTopic* {.
       desc: "the pubsub topic for which rln-relay gets enabled",
-      defaultValue: ""
+      defaultValue: "waku/2/rlnrelay/proto"
       name: "rln-relay-pubsub-topic" }: string
     staticnodes* {.
       desc: "Peer multiaddr to directly connect with. Argument may be repeated."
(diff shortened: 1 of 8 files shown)


waku/v2/node/config.nim
```diff
@@ -76,10 +77,14 @@ type
       name: "rln-relay" }: bool
     
     rlnRelayMemIndex* {.
-      desc: "(experimental) the index of node in the rln-relay group: a value between 0-49 inclusive",
-      defaultValue: 0
-      name: "rln-relay-membership-index" }: uint32
-    
+      desc: "(experimental) the index of node in the rln-relay group: a value between 0-99 inclusive",
+      defaultValue: MembeshipIndex(0)
+      name: "rln-relay-membership-index" }: MembeshipIndex
```
jm-clius, line 82: I missed this during the first reading, but perhaps now is the time to fix the `MembeshipIndex` type to `MembershipIndex` (missing `r`)? :)

waku/v2/node/config.nim
```diff
@@ -76,10 +77,14 @@ type
...
+
+    rlnRelayPubsubTopic* {.
+      desc: "the pubsub topic for which rln-relay gets enabled",
+      defaultValue: ""
```
jm-clius, line 86: Does it make sense to define a sensible default topic here, such as `waku/2/rlnrelay/proto`? Otherwise, with `rln` defined the node may consider an empty string as valid configuration.

jm-clius (approved):
LGTM. Really like how concerns are separated. Thanks!

waku/v2/node/wakunode2.nim
```diff
@@ -408,18 +408,41 @@ proc mountStore*(node: WakuNode, store: MessageStore = nil, persistMessages: boo
...
     # TODO return a bool value to indicate the success of the call
     # check whether inputs are provided
+
+    # relay protocol is the prerequisite of rln-relay
+    if node.wakuRelay.isNil:
+      error "Failed to mount WakuRLNRelay. Relay protocol is not mounted."
+      return
+    # check whether the pubsub topic is supported at the relay level
+    if pubsubTopic notin node.wakuRelay.defaultTopics:
+      error "Failed to mount WakuRLNRelay. The relay protocol does not support the configured pubsub topic.", pubsubTopic=pubsubTopic
```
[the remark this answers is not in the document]
staheri14: Indeed! thanks!

waku/v2/node/config.nim
staheri14: Good idea, although an empty topic would not get through (rln relay will not consider it valid) since it is checked against all the pubsub topics supported at the relay level (that there would be no support for an empty topic).

waku/v2/node/config.nim
staheri14: :D done in https://github.com/status-im/nim-waku/pull/722/commits/58589923801d2dbef6463367f76275718f8f5f64

waku/v2/node/config.nim
staheri14: done in https://github.com/status-im/nim-waku/pull/722/commits/58589923801d2dbef6463367f76275718f8f5f64

waku/v2/node/wakunode2.nim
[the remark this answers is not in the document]
staheri14: done in https://github.com/status-im/nim-waku/pull/722/commits/58589923801d2dbef6463367f76275718f8f5f64

PR #728, opened 2021-10-04
Remove `udpPort` CLI option from waku v1 and v2 nodes

PascalPrecht:
The UDP port is merely used to satisfy underlying `nim-eth` APIs
(namely `proc redirectPorts` which expects a UDP port).

Since there's no support for UDP transport in nim-waku, there's no
need to have a CLI option for that.

This commit removes the option from the CLI while keeping the value
inside application code in tact to not break calls to third-party
APIs.


waku/v1/node/wakunode1.nim
```diff
@@ -15,20 +15,23 @@ const clientId = "Nimbus waku node"
 
 proc run(config: WakuNodeConf, rng: ref BrHmacDrbgContext)
       {.raises: [Defect, ValueError, RpcBindError, CatchableError, Exception]} =
+  ## `udpPort` is only supplied to satisfy underlying APIs but is not
+  ## actually a supported transport.
+  let udpPort = conf.tcpPort
```
jm-clius, line 20:
```suggestion
  let udpPort = config.tcpPort
```

jm-clius (changes requested):
Looks reasonable to me. Thanks! Just the one change - `conf` -> `config` to have `wakunode1` compile.

waku/v1/node/wakunode1.nim
PascalPrecht: D'oh sorry. Missed that one.

waku/common/wakubridge.nim
```diff
@@ -265,24 +265,28 @@ when isMainModule:
   if conf.logLevel != LogLevel.NONE:
     setLogLevel(conf.logLevel)
 
+  ## `udpPort` is only supplied to satisfy underlying APIs but is not
+  ## actually a supported transport.
+  let udpPort = conf.tcpPort
```
jm-clius, line 270: The bridge has separate TCP configurations for v1 and v2. For now we can probably just read the devp2pTcpPort (v1), though it may be worth having a `devp2pUdpPort` and a `libp2pUdpPort`?

```suggestion
  let udpPort = conf.devp2pTcpPort
```

jm-clius (changes requested):
I think there are two more places where compilation will fail.
One is indicated below (`conf.tcpPort` does not exist for the bridge).
The other is hidden under the examples, where `basic2.nim` still attempts to read the `conf.udpPort`: https://github.com/status-im/nim-waku/blob/53854105cb6026bfee6acf0911d216fd2ca9e08a/examples/v2/basic2.nim#L21

waku/common/wakubridge.nim
```diff
@@ -265,24 +265,28 @@ when isMainModule:
   if conf.logLevel != LogLevel.NONE:
     setLogLevel(conf.logLevel)
 
+  ## `udpPort` is only supplied to satisfy underlying APIs but is not
+  ## actually a supported transport.
+  let udpPort = conf.devp2pTcpPort
```
staheri14, line 270: I am not very familiar with the context of this PR, but just got curious whether this assignment would cause any issue/conflicts in the lower level protocols? it seems like a port configured for a TCP connection is also used as a `udpPort`? Or maybe the naming is misleading! :)

PascalPrecht: @staheri14 I've tried to clarify that in the commit message + all the comments in the code where changes where made:

> The UDP port is merely used to satisfy underlying nim-eth APIs
(namely proc redirectPorts which expects a UDP port).

> Since there's no support for UDP transport in nim-waku, there's no
need to have a CLI option for that.

> This commit removes the option from the CLI while keeping the value
inside application code in tact to not break calls to third-party
APIs.

The UDP port isn't used anywhere but we need to have it unless we want to introduce optional parameters in libp2p APIs (which is a bigger effort). Hence, we've decided to just assign the TCP port for now, but it could be any value really.

Does this make sense? Happy to add more clarification to the commit if needed!

oskarth (commented):
Thanks for the PR!

Was there something specific problem that prompted this? Don't we want to keep the UDP port for Discovery v5?

waku/common/wakubridge.nim
PascalPrecht: To clarify further: 

The alternative would be to either hard code some dummy port in all the places where UDP port is used, or have a variable defined somewhere to avoid repetition.

The second option is more convenient, so we went with that use the already existing TCP port for that.

jm-clius:
Thanks for your help with this, @PascalPrecht. I'm gonna go ahead and merge this. 👍

PR #731, opened 2021-10-06
Make maximum connections configurable

jm-clius:
This PR partly addresses https://github.com/status-im/nim-waku/issues/721 and is a prerequisite vir the upcoming [`v0.6` release]()

It makes the max allowed libp2p connections per `wakunode2` configurable.
Once this change is released, it will be installed on the `wakuv2.prod` fleet, after which the `prod` fleet connection limit can be increased from the hardcoded `50`.

### DIFF
--- a/waku/v2/node/config.nim
+++ b/waku/v2/node/config.nim
@@ -46,6 +46,11 @@ type
       desc: "Specify method to use for determining public address. " &
             "Must be one of: any, none, upnp, pmp, extip:<IP>."
       defaultValue: "any" }: string
+
+    maxConnections* {.
+      desc: "Maximum allowed number of libp2p connections."
+      defaultValue: 50
+      name: "max-connections" }: uint16
     
     ## Persistence config
     
--- a/waku/v2/node/wakunode2.nim
+++ b/waku/v2/node/wakunode2.nim
@@ -128,7 +128,8 @@ template tcpEndPoint(address, port): auto =
 proc new*(T: type WakuNode, nodeKey: crypto.PrivateKey,
     bindIp: ValidIpAddress, bindPort: Port,
     extIp = none[ValidIpAddress](), extPort = none[Port](),
-    peerStorage: PeerStorage = nil): T 
+    peerStorage: PeerStorage = nil,
+    maxConnections = MaxConnections): T 
     {.raises: [Defect, LPError].} =
   ## Creates a Waku Node.
   ##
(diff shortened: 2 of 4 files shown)


waku/v2/node/wakunode2.nim
```diff
@@ -128,7 +128,8 @@ template tcpEndPoint(address, port): auto =
 proc new*(T: type WakuNode, nodeKey: crypto.PrivateKey,
     bindIp: ValidIpAddress, bindPort: Port,
     extIp = none[ValidIpAddress](), extPort = none[Port](),
-    peerStorage: PeerStorage = nil): T 
+    peerStorage: PeerStorage = nil,
+    maxConnections = MaxConnections): T
```
staheri14, line 132: Could not find the `MaxConnections` declaration? is it added in this PR?

waku/v2/node/config.nim
```diff
@@ -46,6 +46,11 @@ type
       desc: "Specify method to use for determining public address. " &
             "Must be one of: any, none, upnp, pmp, extip:<IP>."
       defaultValue: "any" }: string
+
+    maxConnections* {.
+      desc: "Maximum allowed number of libp2p connections."
+      defaultValue: 50
+      name: "max-connections" }: uint16
```
staheri14, line 53: Is there any specific reason for `uint16` type? I see a conversion to `int` in the `WakuNode.new(...` (`conf.maxConnections.int`) for the max connection, what do you think about defaulting it to `int`? unsigned values are prone to computational errors

waku/v2/node/wakunode2.nim
staheri14: Wondering where the `maxConnection` gets checked?

waku/v2/node/config.nim
jm-clius: The underlying `switch` builder requires an `int`, hence the conversion. In general, though, we want to avoid using `int`, as its size [varies according to platform](https://status-im.github.io/nim-style-guide/language.integers.html).
For now, `uint16` is what's used for some other config options (so I tried to stay consistent) and will ensure that people don't accidentally configure a silly max connection value, like a negative value or one that's in the millions (it ensures 0 < `maxConnections` < 65535). LMK if this reasoning makes sense!

waku/v2/node/wakunode2.nim
jm-clius: Hah. Good point.

It's an underlying `switch` configuration variable (we try and keep these hidden, unless we actually need to configure one). This one is specifically defined and [checked in the `libp2p` `connmanager`](https://github.com/status-im/nim-libp2p/blob/75bfc1b5f7679afc104bd1ceee1a0dc3bab7a316/libp2p/connmanager.nim#L97). `MaxConnections` is also [defined there](https://github.com/status-im/nim-libp2p/blob/75bfc1b5f7679afc104bd1ceee1a0dc3bab7a316/libp2p/connmanager.nim#L27). Since we import that module (via `builders`) we can use it here. I've tried to make it a little bit clearer now, by adding the imported module as namespacing.

waku/v2/node/wakunode2.nim
```diff
@@ -128,7 +128,8 @@ template tcpEndPoint(address, port): auto =
 proc new*(T: type WakuNode, nodeKey: crypto.PrivateKey,
     bindIp: ValidIpAddress, bindPort: Port,
     extIp = none[ValidIpAddress](), extPort = none[Port](),
-    peerStorage: PeerStorage = nil): T 
+    peerStorage: PeerStorage = nil,
+    maxConnections = builders.MaxConnections): T
```
staheri14, line 132: nice, thanks!

waku/v2/node/config.nim
staheri14: Ah, right! not `int` but some int with a specific size 👍 

Another possible problem with unsigned int is in subtraction operations where you never get a negative value :P I imagine since the max connection is checked at the p2p level (in which the maxConnection is an `int` type), there should be no issue. Thanks for the explanation! :)

oskarth (approved):

jm-clius:
> Did something happen to the CI?

@oskarth, yeah, being tracked here: https://github.com/status-im/nim-waku/issues/738 and the main reason why this has not been merged yet.

rshiv (commented):
[REDACTED]

PR #732, opened 2021-10-06
Waku v2 Websocket support

rshiv:
This PR introduces web socket transport to wakunode and subsequent unit tests. 
Next iteration will introduce secure websocket capablities to the same.

Closes issue https://github.com/status-im/nim-waku/issues/756 and increment towards issue https://github.com/status-im/nim-waku/issues/434

The PR is ready to be merged


waku/v2/node/wakunode2.nim
```diff
@@ -122,6 +124,9 @@ proc removeContentFilters(filters: var Filters, contentFilters: seq[ContentFilte
 template tcpEndPoint(address, port): auto =
   MultiAddress.init(address, tcpProtocol, port)
 
+template addWssFlag() =
+  MultiAddress.init(multiCodec("ws"))
```
jm-clius, line 128: Shouldn't this be "wss" (codec 0x01DE) rather than "ws" (codec 0x01DD). You should also be able to initialise the MultiAddress just with the string "wss". It has an `init` function that will parse the `multiCodec` for you.

waku/v2/node/wakunode2.nim
```diff
@@ -136,7 +141,10 @@ proc new*(T: type WakuNode, nodeKey: crypto.PrivateKey,
   ##
   let
     rng = crypto.newRng()
-    hostAddress = tcpEndPoint(bindIp, bindPort)
+    hostAddress = tcpEndPoint(bindIp, bindPort) 
+
+    hostAddressWithWss = hostAddress & addWssFlag.get()
```
jm-clius, line 146: I don't think this is used anywhere? It should be possible to build a `switch` with both TCP and wss endpoints (use `withAddresses()` rather than `withAddress()`). Also, your host would need to use a different TCP port for the wss endpoint.

waku/v2/node/wakunode2.nim
```diff
@@ -153,6 +161,17 @@ proc new*(T: type WakuNode, nodeKey: crypto.PrivateKey,
   for multiaddr in announcedAddresses:
     peerInfo.addrs.add(multiaddr) # Announced addresses in index > 0
   
+  # WSS switch builder code
+  let WssSwitch = SwitchBuilder
```
jm-clius, line 165:
```suggestion
  let wssSwitch = SwitchBuilder
```

Variable names should be camelCase with first letter lowercase.

waku/v2/node/wakunode2.nim
```diff
@@ -153,6 +161,17 @@ proc new*(T: type WakuNode, nodeKey: crypto.PrivateKey,
...
+    .new()
+    .withAddress(hostAddress)
+    .withRng(rng)
+    .withMplex()
+    .withTransport(proc (upgr: Upgrade): Transport = WsTransport.new(upgr))
+    .withNoise()
+    .build()
+
+
   var switch = newStandardSwitch(some(nodekey), hostAddress,
```
jm-clius, line 175: I understand that it's easier to test with a dedicated switch for `wss`, but I don't think we want two switches (a single `switch` supports multiple transports and multiple host endpoints already). What's more your `wssSwitch` does not have a private key, etc. My suggestion: `wakunode2` should stop relying on `newStandardSwitch()` and instead use its own constructor that is based on that function. The only difference to what was before, is that wss endpoint and transport information _can_ also be included, if enabled in config. This constructor should preferably live in another module, like `waku_setup`.

waku/v2/node/wakunode2.nim
D4nte: I'd expect the host to be able to run both TCP and WSS at the same time (just in case it was not clear).

rshiv:
> How is this going @rshiv? Would you mind updating PR description to explain current state and blocker? Feel free to include logs in here too with what is the current problem you are trying to figure out.

The integration is successful I will update the PR to reflect the changes for web socket support, and later add wss transport support as those would need tls certificate to work.

waku/v2/node/wakunode2.nim
rshiv: I was reading about the same, as per libp2p documentation this case of multiple transport is handled using circuit relay, however I am not sure, I am trying to get more information, and to check if there is some workaround.

waku/v2/node/wakunode2.nim
```diff
@@ -138,28 +181,38 @@ proc new*(T: type WakuNode, nodeKey: crypto.PrivateKey,
   let
     rng = crypto.newRng()
     hostAddress = tcpEndPoint(bindIp, bindPort)
+    wsHostAddress = tcpEndPoint(bindIp, wsbindPort) & addWsFlag
     announcedAddresses = if extIp.isNone() or extPort.isNone(): @[]
```
rshiv, line 185: The extIp is the tcp Ip, wakunode new function requires change for the new config to work.

waku/v2/node/config.nim
```diff
@@ -47,6 +52,10 @@ type
       defaultValue: 50
       name: "max-connections" }: uint16
     
+    transport* {.
+      desc: "Specify transport to use for wakunode " &
```
oskarth, line 56: This should allow for multiple transports. We don't want to add support for Websockets and remove support for TCP.

waku/v2/node/wakunode2.nim
```diff
@@ -138,28 +181,38 @@ proc new*(T: type WakuNode, nodeKey: crypto.PrivateKey,
...
   for multiaddr in announcedAddresses:
     peerInfo.addrs.add(multiaddr) # Announced addresses in index > 0
   
-  var switch = newStandardSwitch(
-    some(nodekey),
-    hostAddress,
-    transportFlags = {ServerFlags.ReuseAddr},
-    rng = rng,
-    maxConnections = maxConnections)
+  var switch = newWakuSwitch(some(nodekey), wsHostAddress,
```
oskarth, line 207: I think the existing formatting is fine and more clear, this change seems unrelated to PR

oskarth (commented):
I see this PR is in draft mode. It'd be good to communicate clearly what is working and what is still a WIP. For example, with excerpt of logs showing how it works and/or how it doesn't.

Also seems to miss tests. Useful tests would be to ensure ws switch works in isolation, and that it works when a node supports tcp and ws (for example).

oskarth:
>  later add wss transport support as those would need tls certificate to work.

We can do this in a separate PR once we basic capabilities done here

waku/v2/node/config.nim
```diff
@@ -31,7 +31,12 @@ type
       desc: "TCP listening port."
       defaultValue: 60000
       name: "tcp-port" }: Port
-
+    
+    wsPort* {.
+      desc: "Web Socket listening port."
```
D4nte, line 36:
```suggestion
      desc: "WebSocket listening port."
```

waku/v2/node/config.nim
```diff
@@ -31,7 +31,12 @@ type
...
+      defaultValue: 1400
```
D4nte, line 37: This is an odd value.

It's an http port. Commonly used value for http server run by a user (ie, port > 1024) are:
- 8080
- 8101
- 8000
- Something that looks like 80, the default http port.

waku/v2/node/wakunode2.nim
```diff
@@ -122,14 +123,56 @@ proc removeContentFilters(filters: var Filters, contentFilters: seq[ContentFilte
...
+      .withAddress(address)
+      .withRng(rng)
+      .withMaxConnections(maxConnections)
+      .withMaxIn(maxIn)
+      .withMaxOut(maxOut)
+      .withMaxConnsPerPeer(maxConnsPerPeer)
+      .withMplex(inTimeout, outTimeout)
+      .withNoise()
+
+    if transport == "ws" or transport == "wss":
```
D4nte, line 160: The addition of `wss` is not consistent across the PR. I'd suggest to leave it totally out for now and add it once you do a PR to add `wss`.

tests/v2/test_wakunode.nim
```diff
@@ -478,7 +478,7 @@ procSuite "WakuNode":
       # invalid IP address
       discard parseRemotePeerInfo("/ip4/127.0.0.0.1/tcp/60002/p2p/16Uuu2HBmAcHvhLqQKwSSbX6BG5JLWUDRcaLVrehUVqpw7fz1hbYc")
     
-    expect ValueError:
+    expect LPError:
```
oskarth, line 481: Can ValueError no longer happen or was this wrong?

tests/v2/test_wakunode.nim
```diff
@@ -1028,3 +1028,140 @@ procSuite "WakuNode":
...
+    await sleepAsync(2000.millis)
+
+
+    check:
+      (await completionFut.withTimeout(5.seconds)) == true
+    await node1.stop()
+    await node2.stop()
+
+
+asyncTest "Messages are relayed with one tcp and one multi transport (tcp and websocket)":
```
oskarth, line 1078:
```suggestion
asyncTest "Messages are relayed between nodes with multiple transports (TCP and Websockets)":
```

tests/v2/test_wakunode.nim
```diff
@@ -1028,3 +1028,140 @@ procSuite "WakuNode":
...
+    await node2.publish(pubSubTopic, message)
+    await sleepAsync(2000.millis)
+
+
+    check:
+      (await completionFut.withTimeout(5.seconds)) == true
+    await node1.stop()
+    await node2.stop()
+
+asyncTest "Messages relay fails with one tcp and only websocket transport":
```
oskarth, line 1122:
```suggestion
asyncTest "Messages relaying fails with non-overlapping transports (TCP or Websockets)":
```

waku/v2/node/wakunode2.nim
```diff
@@ -138,28 +144,43 @@ proc new*(T: type WakuNode, nodeKey: crypto.PrivateKey,
...
-    rng = rng,
-    maxConnections = maxConnections)
+  var switch = newWakuSwitch(some(nodekey),
+  hostAddress,
+  wsHostAddress, 
+  transportFlags = {ServerFlags.ReuseAddr},
+  rng = rng, 
+  maxConnections = maxConnections,
+  wsFlag = wsFlag)
+  # TODO Untested - verify behavior after switch interface change
```
oskarth, line 177: What is this comment referring to?

Can write `wsFlag` as `wsEnabled` to make intent a bit more clear

oskarth (commented):
Nice one! Definitely taking shaping :) Minor comments

jm-clius:
Forgot to add: please update [CHANGELOG.md](https://github.com/status-im/nim-waku/blob/master/CHANGELOG.md) as well. Needed for release documentation.

tests/v2/test_wakunode.nim
rshiv: Our implementation for check has changed and now we don't check the same and libp2p handles the scenario like other scenarios.

waku/v2/node/wakunode2.nim
rshiv: This is an old comment will remove the same.

waku/v2/utils/peers.nim
```diff
@@ -44,22 +44,23 @@ proc init*(p: typedesc[RemotePeerInfo],
...
+## Check if wire Address is supported
+proc validWireAddr*(ma: MultiAddress): bool =
+  const
+    ValidTransports = mapOr(TCP, WebSockets)
+  if ValidTransports.match(ma) == true:
+    return true
   else:
-    raise newException(ValueError,
-                       "Invalid bootstrap node multi-address")
+    return false
```
jm-clius, line 54: You can replace this whole conditional with:
```
return ValidTransports.match(ma)
```
This is simpler, since you are testing a boolean condition and returning the result of that condition.

waku/v2/utils/wakuswitch.nim
```diff
@@ -0,0 +1,57 @@
+# Waku Switch utils.
+import
+  std/[options, tables, strutils, sequtils, os],
```
jm-clius, line 3: Thanks for splitting this out! Are all of these imports necessary in the new module? Not sure, but I don't think e.g. `ping` and `enr` is used?
Imports can be messy, but good idea to try our best to not have unused/unnecessary imports.

jm-clius (commented):
I've added a couple more comments, and would like to see the CHANGELOG updated too, but think we're v close to merging. 👍

waku/v2/utils/wakuswitch.nim
```diff
@@ -0,0 +1,57 @@
...
+      ],
+    transportFlags: set[ServerFlags] = {},
+    rng = crypto.newRng(),
+    inTimeout: Duration = 5.minutes,
+    outTimeout: Duration = 5.minutes,
+    maxConnections = MaxConnections,
+    maxIn = -1,
+    maxOut = -1,
+    maxConnsPerPeer = MaxConnectionsPerPeer,
+    nameResolver: NameResolver = nil,
```
jm-clius, line 35: This is not yet added in the `SwitchBuilder` below.

oskarth (approved):
LGTM

jm-clius (approved):
This looks good! Let's get it merged. Make sure you squash all commits into a single commit based on `master` (Github's default "Squash and Merge" should work).

PR #758, opened 2021-11-02
Fix: store dimensioning issues

### LINKED ISSUE #702: Improvements to `store` memory and disk management
Relates to https://github.com/status-im/nim-waku/issues/695

## Background

The `store` protocol currently has no upper limit for the number of messages it will attempt to store (both in-memory and on-disk). This could lead to out-of-memory issues, long IO pauses and `HistoryResponse` messages that exceed the maximum length.

Furthermore, for each message received, the `store` will:
- write that message to persistent storage
- if it fails to do the above, log at `warn` level that it failed to do so.

Writing to persistent storage is relatively slow, so we may do better by inserting bundles (rather than individual messages) in the SQLite DB. This may, of course, lead to some in-memory message losses after a restart, but with a FT store protocol the risks are likely acceptable.

Moreover, it seems as if the `store` often gets into a state where it fails to store multiple messages (see #695). Logging each of these at `warn` level further exacerbates the problem at high message rates and clutters the logs. I'd suggest lowering the log level, or also only logging bundled write operations.

## Acceptance criteria

- [ ] A configurable maximum number of messages that can be stored, with the oldest elements removed first (e.g. a circular FIFO buffer).

## Further suggested improvements

- [ ] Bundled writes to persistent storage. I'd suggest writing bundled messages every, say, 10 seconds or so. That way we only have ~6 DB operations per minute.
- [ ] Per-message logs should preferably not exceed `trace` level (even if logging an error condition).

cc @staheri14 @oskarth

jm-clius:
This PR closes #702. It is a prerequisite for the [next release](https://github.com/status-im/nim-waku/issues/725)

### What it does

It adds the following fixes for `store`:
- limiting the number of messages that can be queued ("stored") in-memory
- priming the in-memory storage with a limited set returned from the persistent database
- making this maximum store capacity configurable
- lowering per-message log rates

Note that the store will be initialised from persistent storage by reading only the _most recent_ records from the database, up to the configured capacity. No records are deleted from persistent storage (yet).

### What can we still do better?

This is a minimal fix for the current `store` capacity issues. We can still improve by:
- replacing the capacity-limited `StoreQueue` buffer with a circular/ring buffer. This way each `add` operation won't have to be accompanied by a `delete` for buffers that are at capacity
- truncating the persistent storage to remain in sync with the in-memory buffer
- batched writing to persistent storage

### DIFF
--- a/waku/v2/node/storage/message/waku_message_store.nim
+++ b/waku/v2/node/storage/message/waku_message_store.nim
@@ -1,7 +1,7 @@
 {.push raises: [Defect].}
 
 import 
-  std/tables,
+  std/[options, tables],
   sqlite3_abi,
   stew/[byteutils, results],
   ./message_store,
@@ -74,8 +74,9 @@ method put*(db: WakuMessageStore, cursor: Index, message: WakuMessage, pubsubTop
 
   ok()
 
-method getAll*(db: WakuMessageStore, onData: message_store.DataProc): MessageStoreResult[bool] =
+method getAll*(db: WakuMessageStore, onData: message_store.DataProc, limit = none(int)): MessageStoreResult[bool] =
   ## Retrieves all messages from the storage.
+  ## Optionally limits the number of rows returned.
   ##
   ## **Example:**
   ##
@@ -114,7 +115,15 @@ method getAll*(db: WakuMessageStore, onData: message_store.DataProc): MessageSto
            WakuMessage(contentTopic: contentTopic, payload: payload , version: uint32(version), timestamp: senderTimestamp.float64), 
                        pubsubTopic)
 
-  let res = db.database.query("SELECT receiverTimestamp, contentTopic, payload, pubsubTopic, version, senderTimestamp FROM " & TABLE_TITLE & " ORDER BY receiverTimestamp ASC", msg)
+  var selectQuery = "SELECT receiverTimestamp, contentTopic, payload, pubsubTopic, version, senderTimestamp " &
+                    "FROM " & TABLE_TITLE & " " & 
+                    "ORDER BY receiverTimestamp ASC"
+  if limit.isSome():
+    # Optional limit applies. This works because SQLITE will perform the time-based ORDER BY before applying the limit.
+    selectQuery &= " LIMIT " & $(limit.get()) &
+                   " OFFSET cast((SELECT count(*)  FROM " & TABLE_TITLE & ") AS INT) - " & $(limit.get()) # offset = total_row_count - limit
+  
+  let res = db.database.query(selectQuery, msg)
   if res.isErr:
     return err("failed")
 
--- a/waku/v2/protocol/waku_store/waku_store.nim
+++ b/waku/v2/protocol/waku_store/waku_store.nim
@@ -44,6 +44,7 @@ logScope:
 
 const
   WakuStoreCodec* = "/vac/waku/store/2.0.0-beta3"
+  DefaultStoreCapacity* = 50000 # Default maximum of 50k messages stored
 
 # Error types (metric label values)
 const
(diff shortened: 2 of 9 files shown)


rshiv (approved):
LGTM

waku/v2/node/storage/message/waku_message_store.nim
```diff
@@ -114,7 +115,15 @@ method getAll*(db: WakuMessageStore, onData: message_store.DataProc): MessageSto
            WakuMessage(contentTopic: contentTopic, payload: payload , version: uint32(version), timestamp: senderTimestamp.float64), 
                        pubsubTopic)
 
-  let res = db.database.query("SELECT receiverTimestamp, contentTopic, payload, pubsubTopic, version, senderTimestamp FROM " & TABLE_TITLE & " ORDER BY receiverTimestamp ASC", msg)
+  var selectQuery = "SELECT receiverTimestamp, contentTopic, payload, pubsubTopic, version, senderTimestamp " &
+                    "FROM " & TABLE_TITLE & " " & 
+                    "ORDER BY receiverTimestamp ASC"
```
oskarth, line 120: Might need index for this if we don't have this already

Are we logging before and after this operation so we can debug long access time in cluster logs?

Even if debug logs, if we only log info that might not be visible

oskarth (approved):
Nice one!

waku/v2/protocol/waku_store/waku_store.nim
```diff
@@ -487,7 +491,7 @@ proc handleMessage*(w: WakuStore, topic: string, msg: WakuMessage) {.async.} =
 
   let res = w.store.put(index, msg, topic)
   if res.isErr:
-    warn "failed to store messages", err = res.error
+    trace "failed to store messages", err = res.error
```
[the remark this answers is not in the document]
jm-clius: > I see the implementation is read through only, this might lead to data inconsistency and i don't see any timestamp or write through strategies, is our assumption messages are not to be modified once in store?

I may not get your exact meaning, but when we query/retrieve data we will only ever consult our in-memory store/cache. This will never in turn query the persistent data storage (we _only_ read from this during startup), which to me is not quite a "read through" strategy? For each new message we receive, we'll store a copy in-memory and on disk (somewhat of a "write through" strategy). The disk/persistent copy is only meant to serve as backup to prime the store after restarts. Every query (read/write) will primarily interface with the in-memory store. Not sure if that answers your question? Also, very happy to receive any suggestions here if there are ways we can do it better - I already think we may have to do batched reads/writes to persistent storage, but that's for a follow-up PR. LMK what you think. :)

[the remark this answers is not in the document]
jm-clius: > is there a reason store failure message is trace?

Whenever we enter a failure condition, this will be logged _per message_. At high message rates the act of logging itself can cause further processing bottle necks and make the logs unusable, as we have indeed seen happening. Even `debug` level has caused such log overflows before. One way of addressing this is to introduce some log filter that will log this at `warn` level once and at `trace`  for every subsequent, say, 10000 messages. Since we do have visibility of these errors in a metric, a first step is to remove per-message `warn` logs that are currently clogging up `store` fleet nodes.

waku/v2/node/storage/message/waku_message_store.nim
jm-clius: > Might need index for this if we don't have this already

There is a primary ID per record taking the message index, if that is what you're referring to? We just don't access these when reading.

> Are we logging before and after this operation so we can debug long access time in cluster logs? Even if debug logs, if we only log info that might not be visible

Good point! I've added some before and after logs now.

oskarth: Nevermind actually, I forgot we only do this select at startup time :man_facepalming:

PR #760, opened 2021-11-03
Release v0.6

jm-clius:
This PR relates to https://github.com/status-im/nim-waku/issues/725

It adds the release notes for `nim-waku` release `v0.6`.

Please go through the changelog and LMK if there are any missing items. @staheri14 there were a couple of commits related to RLN, but I'm not sure if there's anything concrete to add to the changelog?

cc @oskarth @rshiv

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,13 @@
-# Changelog
+## 2021-11-05 v0.6
 
-## Next version
+Some useful features and fixes in this release, include:
+- two methods for Waku v2 node discovery
+- support for unsecure websockets, which paves the way for native browser usage
+- a fix for `nim-waku` store nodes running out of memory due to store size: the number of stored messages can now easily be configured
+- a fix for densely connected nodes refusing new connections: the maximum number of allowed connections can now easily be configured
+- support for larger message sizes (up from 64kb to 1Mb per message)
 
-This release contains the following:
+The full list of changes is below.
 
 ### Features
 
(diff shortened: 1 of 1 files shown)


rshiv (approved):
LGTM

staheri14:
>  @staheri14 there were a couple of commits related to RLN, but I'm not sure if there's anything concrete to add to the changelog?

The situation with RLN is a bit complicated, there are changes that I have made though they never get compiled unless the RLN flag is set to true. For instance, I have defined the `RateLimitProof` protobuf message, and it would be one of the `WakuMessage` fields in case RLN flag is true during compilation, otherwise, not. Considering this, I think, there is nothing to include in the changelog at this point, until we enable RLN by default. I am open to suggestions!

oskarth:
> there is nothing to include in the changelog at this point, until we enable RLN by default

I think we should still mention it. Just preface it is off by default and under a compiler flag

CHANGELOG.md
```diff
@@ -1,6 +1,4 @@
-# Changelog
-
-## Next version
+## 2021-11-05 v0.6
```
oskarth, line 1: I wonder if we should add a brief description of _benefits_ for users? I.e. we mention "store configurable" but we don't say what this fixes. Ditto re websockets paving way for native browser usage. Could be a paragraph at top of release notes that focuses on benefits to end users. WDYT?

Also I think fine to include RLN changes, prefacing with that it is experimental and under compiler flag.

Otherwise LGTM!

jm-clius: Good idea. RLN changes and top-level summary of benefits added in 6bc46d6

jm-clius:
@staheri14, based on Oskar's suggestions, I've added a short list of RLN changes in 6bc46d6. LMK if you think anything else required/wording should be different.

oskarth (approved):

oskarth:
LGTM, if we get this out tomorrow I can tweet as a nice Friday News

jm-clius:
> LGTM, if we get this out tomorrow I can tweet as a nice Friday News

Agree! If nothing comes up in my local tests, it seems likely.

PR #763, opened 2021-11-08
Proposal: ENR filtering on Waku-field in `discv5`

jm-clius:
This PR proposes adding a `waku2` field to `wakunode2` ENRs for `discv5` purposes.

The field takes the following format:

```
waku2 : <flags>
```

where `waku2` is the key
and `<flags>` is an 8-bit flag field. Currently only the LSB is used: specified to be `1` to indicate `waku2` as `true`.

It is conceivable that the field value, `<flags>`, can be extended with other identifying/filterable flags in future.

All nodes add this to their discoverable ENR and _only_ discover nodes with the `waku` field and version as above.

Note that this only allows `wakunode2`s participating in `discv5` to filter _discovered_ nodes, it does not apply any filter on what actually gets added to the routing table. For the latter we'd have to fork/modify the existing nim `discv5` implementation.

## What should be reviewed?

- the format of the added field: if we can get rough consensus on this, we can add this to specifications e.g. [RFC31](https://github.com/vacp2p/rfc/pull/465)
- the logic: whether a filter on discovered nodes is sensible and enough - i.e. if we should also filter what goes into our routing tables, to address the needle-in-a-haystack problem

> UPDATE 2021-11-09: Based on recommendations, changed the waku2 field value to 8-bit flag field with only LSB currently used.


arnetheduck:
given the limited space that enr has, wouldn't it be more appropriate to use a bit flag field here with waku-related capabilities ?

kdeme (approved):
lgtm, @arnetheduck does have a point that perhaps using immediately a bitlist could be a good idea, even if only 1 bit is used for now. (could change the field name to "waku2")

rshiv (approved):
LGTM

jm-clius:
UPDATE: Based on recommendations above, I changed the field name to `waku2` and the value to an 8-bit flag field with only LSB currently used and set to `1`.

oskarth (approved):

oskarth:
I wonder if it makes sense to sketch out bitfield usage now already? I thought using `waku` instead of `wakuv2` might make sense, but then that would require two bits for Waku version...

More of an RFC discussion, but sketch of information that could be useful at this initial filtering step: `waku, relay, store, filter, lightpush, reputation_registered`. Possibly something more on store side (ft store / state sync / "community" sync). Unclear if we need this for RLN. This also wouldn't capture things like pubsub topics, but that might not be feasible, especially in 8 bit field. Possibly things like WS/RTC support? Though that might be a bit of a catch 22...

What we are really describing here are capabilities related to Waku, so that might be more in line with something like https://rfc.vac.dev/spec/30/

LGTM as initial cut though and can iterate on it in spec discussion before it becomes stable.

jm-clius:
UPDATE: brought implementation in line with recent additions in [RFC 31](https://rfc.vac.dev/spec/31/).

waku/v2/node/discv5/waku_discv5.nim
```diff
@@ -63,12 +93,14 @@ proc findRandomPeers*(wakuDiscv5: WakuDiscoveryV5): Future[Result[seq[RemotePeer
   ## Find random peers to connect to using Discovery v5
   
   ## Query for a random target and collect all discovered nodes
-  ## @TODO: we could filter nodes here
   let discoveredNodes = await wakuDiscv5.protocol.queryRandom()
   
+  ## Filter based on our needs
+  let filteredNodes = discoveredNodes.filter(isWakuNode) # Currently only a single predicate
```
oskarth, line 99: Does this mean Discovery v5 logic is stopped for non-filter nodes? Trying to understand how we'll keep the network cleanly separated in case of happy little accident

kdeme: Are you referring to keeping the networks separated from the eth2 network(s) (and others)?

If so, no, this is not what it does. It only filters the nodes it just found through a query (querying a bunch of other nodes for more nodes). The underlying routing table will not get filtered and contains all sort of nodes.

jm-clius: Yeah, this only uses the flag to filter the discovered nodes for usefulness elsewhere. If we want to limit what goes into the routing table we'll have to modify the underlying routing table logic (likely forking the implementation/creating our own).

Issue #764, opened 2021-11-09
Publish both TCP and websocket address


rshiv:
In current implementation if websocket is enabled , we don't publish the tcp address.
![image](https://user-images.githubusercontent.com/12139926/140928211-b5416a27-9749-4a1a-ad9b-3288d938560c.png)

Publish both the addresses when websocket is enabled.

Issue #744, opened 2021-10-14
Use libp2p PeerInfo struct


Menduist:
Here, you're using you own PeerInfo instance to show the listening address:
https://github.com/status-im/nim-waku/blob/165e23515807dee923569455fd40f5b7ac8fd225/waku/v2/node/wakunode2.nim#L720-L725

This is not good, because libp2p will rewrite some addresses in some cases (for instance, when the port is 0, or when you use circuit-relay (in the future), or when you try to use a WSS address on a WS transport)

You should only use the `PeerInfo` which is stored in the Switch, not have your own copy

The flow should be:
- You put your wanted listen addresses has the switch input (eg, `withAddresses(@[...])`)
- You start the switch
- You can see your usable addresses in `switch.peerInfo.addrs`

LMK if I missed something ;)

PR #766, opened 2021-11-10
Publish multiple address when websocket enabled.

rshiv:
Closes Issue #764. Increment towards #744 
Will publish multiple address when websocket config is enabled.
This PR also adds announced addresses in node object and references it for display of the node address.

Screenshot of the multiple address:
image

### DIFF
--- a/waku/v2/node/wakunode2.nim
+++ b/waku/v2/node/wakunode2.nim
@@ -820,7 +820,7 @@ proc start*(node: WakuNode) {.async.} =
   info "PeerInfo", peerId = peerInfo.peerId, addrs = peerInfo.addrs
   var listenStr = ""
   for address in node.announcedAddress:
-    var fulladdr = "["& $address & "/p2p/" & $peerInfo.peerId & "]" 
+    var fulladdr = "[" & $address & "/p2p/" & $peerInfo.peerId & "]" 
     addf(listenStr, fulladdr)
                 
   ## XXX: this should be /ip4..., / stripped?
--- a/waku/v2/node/wakunode2.nim
+++ b/waku/v2/node/wakunode2.nim
@@ -55,6 +55,7 @@ const clientId* = "Nimbus Waku v2 node"
 # Default topic
 const defaultTopic = "/waku/2/default-waku/proto"
 
+
 # key and crypto modules different
 type
   KeyPair* = crypto.KeyPair
@@ -87,6 +88,7 @@ type
     filters*: Filters
     rng*: ref BrHmacDrbgContext
     wakuDiscv5*: WakuDiscoveryV5
+    webSocketEnabled*: bool 
     started*: bool # Indicates that node has started listening
 
 proc protocolMatcher(codec: string): Matcher =
@@ -168,8 +170,11 @@ proc new*(T: type WakuNode, nodeKey: crypto.PrivateKey,
     enrTcpPort = if extPort.isSome(): extPort
                  else: some(bindPort)
     enr = createEnr(nodeKey, enrIp, enrTcpPort, none(Port))
+    webSocketEnabled = if wsEnabled or wssEnabled: true
+                       else: false
+    
   
-  if wsEnabled == true or wssEnabled == true:
+  if wsEnabled == true or wssEnabled:
     info "Initializing networking", hostAddress, wsHostAddress,
                                     announcedAddresses
     peerInfo.addrs.add(wsHostAddress)
@@ -194,10 +199,11 @@ proc new*(T: type WakuNode, nodeKey: crypto.PrivateKey,
   let wakuNode = WakuNode(
     peerManager: PeerManager.new(switch, peerStorage),
     switch: switch,
-    rng: rng,
+    rng: rng, 
     peerInfo: peerInfo,
     enr: enr,
-    filters: initTable[string, Filter]()
+    filters: initTable[string, Filter](),
+    webSocketEnabled: webSocketEnabled
   )
 
   return wakuNode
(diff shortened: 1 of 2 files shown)


waku/v2/node/wakunode2.nim
```diff
@@ -789,7 +789,7 @@ proc stopDiscv5*(node: WakuNode): Future[bool] {.async.} =
 
     debug "Successfully stopped discovery v5 service"
 
-proc start*(node: WakuNode) {.async.} =
+proc start*(node: WakuNode, websocketConn: bool = false) {.async.} =
```
oskarth, line 792: Is there a reason the websocketConn isn't part of WakuNode object? Seems a bit odd to add another argument to start here but maybe it's fine. It is false by default so that's good regardless for compatibility.

waku/v2/node/wakunode2.nim
```diff
@@ -800,7 +800,9 @@ proc start*(node: WakuNode) {.async.} =
   # TODO Get this from WakuNode obj
   let peerInfo = node.peerInfo
   info "PeerInfo", peerId = peerInfo.peerId, addrs = peerInfo.addrs
-  let listenStr = $peerInfo.addrs[^1] & "/p2p/" & $peerInfo.peerId
+  let listenStr = if websocketConn: @[$peerInfo.addrs[^1] & "/p2p/" & $peerInfo.peerId,
```
oskarth, line 803: This seems a bit error-prone, especially if we have more transports. What else is in addrs field? Can we iterate over this and add them all?

Might be worth to see what Nimbus or other libp2p consumers do here

waku/v2/node/wakunode2.nim
```diff
@@ -1025,8 +1027,14 @@ when isMainModule:
     ## Resume history, connect to static nodes and start
     ## keep-alive, if configured.
 
+    #if websocket or secure websocket enabled,
+    #start websocket server with websocketConn as true.
+    if(conf.websocketSupport):
+      waitFor node.start(websocketConn = true)
```
oskarth, line 1033: I feel like this should be part of the node object at config time probably

oskarth (commented):
Btw, if you paste link to issue it is easier to click in issue :) https://github.com/status-im/nim-waku/issues/764

waku/v2/node/wakunode2.nim
rshiv: We have host address along with announced address.
Iterating over them may not be the best idea as we have to add peerID separately anyway, however it might be a good idea to make the index constants .

waku/v2/node/wakunode2.nim
rshiv: makes sense

waku/v2/node/wakunode2.nim
rshiv: fixed

jm-clius: I think this still needs to be removed from the `start()` arguments.

waku/v2/node/wakunode2.nim
jm-clius: We have this outstanding issue to remove reliance on our own cached `peerInfo`: https://github.com/status-im/nim-waku/issues/744

We don't have to remove `peerInfo` in this PR, but I think a better intermediate step would be to cache the `announcedAddresses` on the node to keep them separate from the host addresses. This way you can just iterate over the announcedAddresses for the listenStr (`info` would have to be updated as well), without requiring the extra state variable in for the node.

jm-clius (changes requested):
Some changes related to our reliance on `peerInfo`. See https://github.com/status-im/nim-waku/issues/744.

waku/v2/node/wakunode2.nim
rshiv: 39ee44c60e9ed111d87a4733e7aadc8efd6119be fixes the same.

waku/v2/node/wakunode2.nim
rshiv: resolved in 7c7f9fb106ac162e7efef4a3ae29daa5bc9110fe

waku/v2/node/wakunode2.nim
```diff
@@ -87,6 +88,7 @@ type
     filters*: Filters
     rng*: ref BrHmacDrbgContext
     wakuDiscv5*: WakuDiscoveryV5
+    announcedAddress* : seq[MultiAddress]
```
jm-clius, line 91:
```suggestion
    announcedAddresses* : seq[MultiAddress]
```
Plural variable name for sequences. If you prefer using abbreviations, we've settled on `addr` for a singular address and `addrs` for more than one. Just make sure it is consistent with other uses of `announcedAddresses`.

waku/v2/node/wakunode2.nim
```diff
@@ -814,7 +818,11 @@ proc start*(node: WakuNode) {.async.} =
   # TODO Get this from WakuNode obj
   let peerInfo = node.peerInfo
   info "PeerInfo", peerId = peerInfo.peerId, addrs = peerInfo.addrs
-  let listenStr = $peerInfo.addrs[^1] & "/p2p/" & $peerInfo.peerId
+  var listenStr = ""
+  for address in node.announcedAddress:
+    var fulladdr = "[" & $address & "/p2p/" & $peerInfo.peerId & "]" 
+    addf(listenStr, fulladdr)
```
jm-clius, line 824: Not sure I see the value of using `addf` rather than simply `&=` if we don't apply any particular formatting? It might just be because I don't know the `addf` function very well. :)

waku/v2/node/wakunode2.nim
```diff
@@ -168,8 +170,9 @@ proc new*(T: type WakuNode, nodeKey: crypto.PrivateKey,
     enrTcpPort = if extPort.isSome(): extPort
                  else: some(bindPort)
     enr = createEnr(nodeKey, enrIp, enrTcpPort, none(Port))
+    
   
-  if wsEnabled == true or wssEnabled == true:
+  if wsEnabled == true or wssEnabled:
```
jm-clius, line 175:
```suggestion
  if wsEnabled or wssEnabled:
```
Minor consistency :)

jm-clius (approved):
LGTM. Some minor comments.

oskarth (commented):
Nothing to add to Hanno's comments

oskarth (approved):

waku/v2/node/wakunode2.nim
rshiv: fixed in 5eff5597eabec4afaca39814bcd04d751c41ed2f

waku/v2/node/wakunode2.nim
rshiv: fixed in 5eff5597eabec4afaca39814bcd04d751c41ed2f

waku/v2/node/wakunode2.nim
rshiv: fixed in 5eff5597eabec4afaca39814bcd04d751c41ed2f

PR #775, opened 2021-11-18
publish addresses from announced address sequence.

rshiv:
The JSON-RPC info published the address from PeerInfo, rather than getting the address from announced address sequence, causing inconsistency in published list of multi-addresses, as PeerInfo also stores host address which is not needed to be published.

This PR changes published addresses location from peerInfo to announced address sequence.


rshiv:
Screenshot of the output of JSON-RPC info:
image

waku/v2/node/wakunode2.nim
```diff
@@ -392,9 +392,11 @@ proc info*(node: WakuNode): WakuInfo =
...
   ##
 
-  # TODO Generalize this for other type of multiaddresses
   let peerInfo = node.peerInfo
-  let listenStr = $peerInfo.addrs[^1] & "/p2p/" & $peerInfo.peerId
+  var listenStr = ""
+  for address in node.announcedAddresses:
+    var fulladdr = "[" & $address & "/p2p/" & $peerInfo.peerId & "]" 
+    listenStr &= fulladdr
   let wakuInfo = WakuInfo(listenStr: listenStr)
```
oskarth, line 400: Considering this is used in WakuInfo response, e.g. here https://rfc.vac.dev/spec/16/#wakuinfo It might make sense to make this response more programmatic. I.e. if we return multiple addresses, it should probably return a proper sequence as opposed to a constructed string. I don't know what other libp2p users do here but I'm sure others have the equivalent of WakuInfo somewhere? Ethereum, IPFS, etc. Could be pulled out as an issue but probably not high priority.

However, this is better than what exists right now and would be useful for logs, so seems fine to me.

oskarth (approved):

waku/v2/node/wakunode2.nim
```diff
@@ -392,9 +392,11 @@ proc info*(node: WakuNode): WakuInfo =
   ## Status: Implemented.
   ##
 
-  # TODO Generalize this for other type of multiaddresses
   let peerInfo = node.peerInfo
-  let listenStr = $peerInfo.addrs[^1] & "/p2p/" & $peerInfo.peerId
+  var listenStr = ""
+  for address in node.announcedAddresses:
+    var fulladdr = "[" & $address & "/p2p/" & $peerInfo.peerId & "]"
```
jm-clius, line 398: Don't feel too strongly about this, but it seems a bit more idiomatic to me if the output was something like
> "[`listenstrAddr1`, `listenstrAddr2`, ...]"

rather than

> "[`listenstrAddr1`][`listenstrAddr2`]..."

I think to attain that you can create a `seq` of `listenstr`s and just use the canonical `$` stringify for the sequence.

jm-clius (approved):
LGTM. Two comments:
1. This requires updating [the spec documentation](https://rfc.vac.dev/spec/16/#debug-api) with a one liner to explain the format of the output.
2. I think this is parsed somewhere in `infra-nim-waku` and used to populate the fleets website. We must ensure that things don't break there and that infra understands the new format.

staheri14 (approved):
LGTM, agree with the comments of Oskar and Hanno

waku/v2/node/wakunode2.nim
rshiv: changed format in 08fd37dabd54bcd0286890d5e610a8688ac6af2a

jakubgs (changes requested):
Are we really putting a string representation of a JSON list inside of a JSON?

rshiv:
image
Changed the JSON output from string to seq of strings.

jakubgs:
A list is much better, but now the `listenStr` name does not make sense. I see 3 options:

1. Leave it as `listenStr` even though data is a list, just bad.
2. Rename it to something that makes sense, like just `listen` or `listenAddresses` or `listenAddr` or `listenList`
3. Leave `listenStr` with one value as it was and add `listenList` beside it with all values

Number two probably is simplest and makes most sense, not sure about naming tho.

rshiv:
Changed the same
```
{"jsonrpc":"2.0","id":"id","result":{"listenAddresses":["/ip4/86.26.190.13/tcp/60000/p2p/16Uiu2HAm2QFuvpXxtHxEmNHakASyjXgE1QgFbHryF7rRrdcRZcNY","/ip4/86.26.190.13/tcp/8000/ws/p2p/16Uiu2HAm2QFuvpXxtHxEmNHakASyjXgE1QgFbHryF7rRrdcRZcNY"]}}
```

jakubgs (approved):
Yeah, that's much better. But there's another option I was thinking about, which would make this easier to use:
```json
{                                                                                                                                                                                                                                                          
  "jsonrpc": "2.0",
  "id": "id",
  "result": {
    "listen": {
      "libp2p": "/ip4/86.26.190.13/tcp/60000/p2p/16Uiu2HAm2QFuvpXxtHxEmNHakASyjXgE1QgFbHryF7rRrdcRZcNY",
      "websocket": "/ip4/86.26.190.13/tcp/8000/ws/p2p/16Uiu2HAm2QFuvpXxtHxEmNHakASyjXgE1QgFbHryF7rRrdcRZcNY"
    }
  }
}
```
This way you can easily extract the address you want with `jq` or whatever. What do you think @oskarth @staheri14 ?

But as it is is fine too.

oskarth:
As a spec, the address of a node should be a multiaddress. This is specified here https://multiformats.io/multiaddr/ and there are multiple implementations.

There may be cases where we want to split up the address for easier parsing with tools like jq etc. If we want to turn it into a map for JSON parsing reasons, the index of each transport should probably correspond with network protocol in address. But at that point, maybe it is easier to just leave it as a multiaddr and parse that somehow?

PR #782, opened 2021-11-23
Handles stale clients in FILTER protocol

### LINKED ISSUE #781: WAKU-FILTER: Handle stale client.
The implementation in waku_filter failure needs to improve as:

Filter will keep attempting to push messages indefinitely until a client node unsubscribes or a node restarts, even if the node client goes offline. 

Proposed solution is introduced in [VAC](https://github.com/vacp2p/rfc/issues/475), which will maintain the state of the client during message push.

rshiv:
This PR closes #781

The PR deletes the peer info if after timeout period the peer is not available. We maintain a map of peerid and first unavailable time to determine its departure.

This PR achieves
- Store the map of the failed peers and its first failure epoch
- Receives the config for timeout from user in float, by default 24 hours.
- Removal of subscriber if failure continues.
- Test to check the above features.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,10 @@
 ##  Next version
 
-This patch release contains the following fix:
-- This Patch introduces Filter Protocol timeout.
-- Support for full secure websockets.
+This release contains the following:
+
+### Features
+- Waku v2 node timeout for Filter nodes.
+- Waku v2 node support for secure websockets.
 
 ### Changes
 - The WakuInfo Object field of `listenStr` is deprecated and is now replaced with `listenAddresses`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ This patch release contains the following fix:
 - Support for full secure websockets.
 
 ### Changes
-- The WakuInfo Object field of listenStr is depreciated and is now replaced with listenAddresses
+- The WakuInfo Object field of `listenStr` is deprecated and is now replaced with `listenAddresses`
 which is a sequence of string.
 
 ## 2021-11-05 v0.6
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,24 @@
+## 2021-07-26 v0.6.1
+
+This patch release contains the following fix:
+- This Patch introduces Filter Protocol timeout.
+- Support for full secure websockets.
+
+### Changes
+- The WakuInfo Object field of listenStr is depreciated and is now replaced with listenAddresses
+which is a sequence of string.
+
+It supports the same [libp2p protocols](https://docs.libp2p.io/concepts/protocols/):
+| Protocol | Spec status | Protocol id |
+| ---: | :---: | :--- |
+| [`17/WAKU-RLN`](https://rfc.vac.dev/spec/17/) | `raw` | `/vac/waku/waku-rln-relay/2.0.0-alpha1` |
+| [`11/WAKU2-RELAY`](https://rfc.vac.dev/spec/11/) | `stable` | `/vac/waku/relay/2.0.0` |
+| [`12/WAKU2-FILTER`](https://rfc.vac.dev/spec/12/) | `draft` | `/vac/waku/filter/2.0.0-beta1` |
+| [`13/WAKU2-STORE`](https://rfc.vac.dev/spec/13/) | `draft` | `/vac/waku/store/2.0.0-beta3` |
+| [`18/WAKU2-SWAP`](https://rfc.vac.dev/spec/18/) | `draft` | `/vac/waku/swap/2.0.0-beta1` |
+| [`19/WAKU2-LIGHTPUSH`](https://rfc.vac.dev/spec/19/) | `draft` | `/vac/waku/lightpush/2.0.0-beta1` |
+
+The Waku v1 implementation is stable but not under active development.
 ## 2021-11-05 v0.6
 
 Some useful features and fixes in this release, include:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 2021-07-26 v0.6.1
+##  Next version
 
 This patch release contains the following fix:
 - This Patch introduces Filter Protocol timeout.
@@ -8,17 +8,6 @@ This patch release contains the following fix:
 - The WakuInfo Object field of listenStr is depreciated and is now replaced with listenAddresses
 which is a sequence of string.
 
-It supports the same [libp2p protocols](https://docs.libp2p.io/concepts/protocols/):
-| Protocol | Spec status | Protocol id |
-| ---: | :---: | :--- |
-| [`17/WAKU-RLN`](https://rfc.vac.dev/spec/17/) | `raw` | `/vac/waku/waku-rln-relay/2.0.0-alpha1` |
-| [`11/WAKU2-RELAY`](https://rfc.vac.dev/spec/11/) | `stable` | `/vac/waku/relay/2.0.0` |
-| [`12/WAKU2-FILTER`](https://rfc.vac.dev/spec/12/) | `draft` | `/vac/waku/filter/2.0.0-beta1` |
-| [`13/WAKU2-STORE`](https://rfc.vac.dev/spec/13/) | `draft` | `/vac/waku/store/2.0.0-beta3` |
-| [`18/WAKU2-SWAP`](https://rfc.vac.dev/spec/18/) | `draft` | `/vac/waku/swap/2.0.0-beta1` |
-| [`19/WAKU2-LIGHTPUSH`](https://rfc.vac.dev/spec/19/) | `draft` | `/vac/waku/lightpush/2.0.0-beta1` |
-
-The Waku v1 implementation is stable but not under active development.
 ## 2021-11-05 v0.6
 
 Some useful features and fixes in this release, include:
--- a/waku/v2/node/config.nim
+++ b/waku/v2/node/config.nim
@@ -131,7 +131,7 @@ type
     
     filterTimeout* {.
       desc: "Timeout for filter node in seconds.",
-      defaultValue: 86400
+      defaultValue: 14400 # 4 hours
       name: "filter-timeout" }: int64
     
     ## Swap config
--- a/waku/v2/node/config.nim
+++ b/waku/v2/node/config.nim
@@ -7,7 +7,7 @@ import
   nimcrypto/utils,
   eth/keys,
   ../protocol/waku_rln_relay/[waku_rln_relay_types]
-
+   
 type
   WakuNodeConf* = object
     ## General node config
@@ -128,6 +128,11 @@ type
       desc: "Peer multiaddr to request content filtering of messages.",
       defaultValue: ""
       name: "filternode" }: string
+    
+    filterTimeout* {.
+      desc: "Timeout for filter node in seconds.",
+      defaultValue: 60 * 60 * 24
+      name: "filter-timeout" }: float
     
     ## Swap config
 
--- a/waku/v2/node/config.nim
+++ b/waku/v2/node/config.nim
@@ -131,8 +131,8 @@ type
     
     filterTimeout* {.
       desc: "Timeout for filter node in seconds.",
-      defaultValue: 60 * 60 * 24
-      name: "filter-timeout" }: float
+      defaultValue: 86400
+      name: "filter-timeout" }: int64
     
     ## Swap config
 
--- a/waku/v2/node/wakunode2.nim
+++ b/waku/v2/node/wakunode2.nim
@@ -56,7 +56,7 @@ const clientId* = "Nimbus Waku v2 node"
 const defaultTopic = "/waku/2/default-waku/proto"
 
 # Default Waku Filter Timeout
-const WakuFilterTimeout: Duration = chronos.days(1)
+const WakuFilterTimeout: Duration = 1.days
 
 
 # key and crypto modules different
--- a/waku/v2/node/wakunode2.nim
+++ b/waku/v2/node/wakunode2.nim
@@ -55,6 +55,9 @@ const clientId* = "Nimbus Waku v2 node"
 # Default topic
 const defaultTopic = "/waku/2/default-waku/proto"
 
+# Default Waku Filter Timeout
+const WakuFilterTimeout = 60 * 60 * 24
+
 
 # key and crypto modules different
 type
@@ -398,7 +401,7 @@ proc info*(node: WakuNode): WakuInfo =
   let wakuInfo = WakuInfo(listenStr: listenStr)
   return wakuInfo
 
-proc mountFilter*(node: WakuNode, filterTimeout: float) {.raises: [Defect, KeyError, LPError]} =
+proc mountFilter*(node: WakuNode, filterTimeout: float = WakuFilterTimeout) {.raises: [Defect, KeyError, LPError]} =
   info "mounting filter"
   proc filterHandler(requestId: string, msg: MessagePush)
     {.gcsafe, raises: [Defect, KeyError].} =
--- a/waku/v2/node/wakunode2.nim
+++ b/waku/v2/node/wakunode2.nim
@@ -398,7 +398,7 @@ proc info*(node: WakuNode): WakuInfo =
   let wakuInfo = WakuInfo(listenStr: listenStr)
   return wakuInfo
 
-proc mountFilter*(node: WakuNode) {.raises: [Defect, KeyError, LPError]} =
+proc mountFilter*(node: WakuNode, filterTimeout: float) {.raises: [Defect, KeyError, LPError]} =
   info "mounting filter"
   proc filterHandler(requestId: string, msg: MessagePush)
     {.gcsafe, raises: [Defect, KeyError].} =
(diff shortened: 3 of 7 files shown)


jm-clius:
Thanks, Reeshav. Not sure we're on the same page re this issue, so maybe you have some more background.

Removing a peer from the peer store/persistent storage may indeed be required in future, but as far as I can see that is a separate concern from `filter` protocol and the issues we see there. In fact, a peer may be connected for more than just `filter` protocol and we wouldn't want to remove it purely based on its `filter` behaviour (for example, other protocols may have much longer timeouts than `filter`).
As it stands, I think the filternode would still view the now-removed-client as a `subscriber` and attempt to send messages matching a filter to that subscriber. It should therefore be removed from the [`subscribers` list in the `WakuFilter` object](https://github.com/status-im/nim-waku/blob/master/waku/v2/protocol/waku_filter/waku_filter_types.nim#L46) and only there, for now. Or am I missing something?

rshiv:
> Thanks, Reeshav. Not sure we're on the same page re this issue, so maybe you have some more background.
> 
> Removing a peer from the peer store/persistent storage may indeed be required in future, but as far as I can see that is a separate concern from `filter` protocol and the issues we see there. In fact, a peer may be connected for more than just `filter` protocol and we wouldn't want to remove it purely based on its `filter` behaviour (for example, other protocols may have much longer timeouts than `filter`). As it stands, I think the filternode would still view the now-removed-client as a `subscriber` and attempt to send messages matching a filter to that subscriber. It should therefore be removed from the [`subscribers` list in the `WakuFilter` object](https://github.com/status-im/nim-waku/blob/master/waku/v2/protocol/waku_filter/waku_filter_types.nim#L46) and only there, for now. Or am I missing something?

Thanks @jm-clius , the original issue with filter and subsequent discussion for the stale connections was to handle the message tries as well as handle the memory overload for the same. Hence, the thought behind implementation was to clear the peer from the persistent storage, and remove the peer from peer manager. If we do remove the peer only from the subscriber list, this would still maintain the peer in the peer manager, even if it is the filter node. 

However, For maintaining the connection for the other protocols was not something I had considered and assumed its just the filter protocol we would connect with, thanks for the insight , I will make the required changes.

waku/v2/node/config.nim
```diff
@@ -129,6 +129,11 @@ type
       defaultValue: ""
       name: "filternode" }: string
     
+    filterTimeout* {.
+      desc: "Timeout for filter node in seconds.",
+      defaultValue: 60 * 60 * 24
+      name: "filter-timeout" }: float
```
oskarth, line 135: Any specific reason this is a float and not e.g. uint16? I don't think we need to specify subseconds :P

waku/v2/protocol/waku_filter/waku_filter.nim
```diff
@@ -187,44 +188,59 @@ method init*(wf: WakuFilter) =
...
-  
+  wf.init(filterTimeout = timeout)
   return wf
 
 proc setPeer*(wf: WakuFilter, peer: RemotePeerInfo) =
   wf.peerManager.addPeer(peer, WakuFilterCodec)
   waku_filter_peers.inc()
 
+proc handleClientError*(wf: WakuFilter, subscriber: Subscriber){.raises: [Defect, KeyError], async.} = 
+  var stringkey: string = $(subscriber)
```
oskarth, line 206: A better variable name might be something like `subscriberIndex` `subKey`. `stringkey` refers to data type, not what it actually is.

waku/v2/protocol/waku_filter/waku_filter.nim
```diff
@@ -187,44 +188,59 @@ method init*(wf: WakuFilter) =
...
+  if not wf.failedPeers.hasKey(stringkey):
```
oskarth, line 207: Seems easier to read if this if condition is inverted so `if wf.failedPeers..`

waku/v2/protocol/waku_filter/waku_filter.nim
```diff
@@ -187,44 +188,59 @@ method init*(wf: WakuFilter) =
...
-  wf.init()
-  
+  wf.init(filterTimeout = timeout)
   return wf
 
 proc setPeer*(wf: WakuFilter, peer: RemotePeerInfo) =
   wf.peerManager.addPeer(peer, WakuFilterCodec)
   waku_filter_peers.inc()
 
+proc handleClientError*(wf: WakuFilter, subscriber: Subscriber){.raises: [Defect, KeyError], async.} =
```
oskarth, line 205: Can you add a comment with what is happening here and expected behavior? There are a lot of comparisons and times and mutations so would make it easier to reason about.

waku/v2/protocol/waku_filter/waku_filter.nim
```diff
@@ -187,44 +188,59 @@ method init*(wf: WakuFilter) =
...
+      wf.subscribers.delete(index)
+  return
+
 
 proc handleMessage*(wf: WakuFilter, topic: string, msg: WakuMessage) {.async.} =
   # Handle WakuMessage according to filter protocol
   trace "handle message in WakuFilter", topic=topic, msg=msg
-
+  var handleMessageFailed = false
+  var failedSubscriber: Subscriber
```
oskarth, line 222: Why do we need this intermittent variable if we are just going to pass subscriber here?

waku/v2/protocol/waku_filter/waku_filter.nim
```diff
@@ -187,44 +188,59 @@ method init*(wf: WakuFilter) =
...
           await connOpt.get().writeLP(push.encode().buffer)
         else:
           # @TODO more sophisticated error handling here
+          handleMessageFailed = true
+          failedSubscriber = subscriber
           error "failed to push messages to remote peer"
           waku_filter_errors.inc(labelValues = [dialFailure])
         break
+  if handleMessageFailed:
+    discard handleClientError(wf, failedSubscriber)
```
oskarth, line 243: Should probably log here

waku/v2/node/config.nim
rshiv: epochTime() or getTime() in nim is represented in float, hence the config is in float as well.

waku/v2/protocol/waku_filter/waku_filter.nim
rshiv: We are changing the subscriber sequence , nim doesn't allow the same while iterating over the same.

waku/v2/protocol/waku_filter/waku_filter.nim
rshiv: fixed in b17f17957452dc78ac1dd16be5f11b50c5f6448a

waku/v2/protocol/waku_filter/waku_filter.nim
rshiv: fixed in b17f17957452dc78ac1dd16be5f11b50c5f6448a

waku/v2/protocol/waku_filter/waku_filter.nim
rshiv: fixed in b17f17957452dc78ac1dd16be5f11b50c5f6448a

waku/v2/protocol/waku_filter/waku_filter.nim
rshiv: added log in b17f17957452dc78ac1dd16be5f11b50c5f6448a

waku/v2/protocol/waku_filter/waku_filter.nim
```diff
@@ -187,44 +188,63 @@ method init*(wf: WakuFilter) =
...
 proc setPeer*(wf: WakuFilter, peer: RemotePeerInfo) =
   wf.peerManager.addPeer(peer, WakuFilterCodec)
   waku_filter_peers.inc()
 
+# If we have already failed to send message to this peer,
+# check for elapsed time and if it's been too long, remove the peer.
+proc handleClientError*(wf: WakuFilter, subscriber: Subscriber){.raises: [Defect, KeyError], async.} = 
+  var subKey: string = $(subscriber)
+  if wf.failedPeers.hasKey(subKey):
+    var elapsedTime = epochTime() - wf.failedPeers[subKey]
```
jm-clius, line 210: I think it should be possible to use https://github.com/status-im/nim-chronos/blob/master/chronos/timer.nim functions herr, rather than using `std/times` functions e.g. `epochTime()`. These are generally faster and less prone to errors.

For example, you could measure the `elapsedTime` as a `Duration` from `Momen.now()`

waku/v2/node/config.nim
jm-clius: Couldn't this be a `Duration`? (https://github.com/status-im/nim-chronos/blob/master/chronos/timer.nim#L185)

jm-clius (changes requested):
Thanks, Reeshav.

This is now closer to what I understood the goal to be. We prefer using `chronos/timer` functions above `std/times`. There are only few cases where we really need a globally significant epoch as reference and where the chronos library won't be sufficient.

waku/v2/node/config.nim
rshiv: solved in 2f0e923b3e6c2494bdd639ee6ad352f56bd2f434

waku/v2/protocol/waku_filter/waku_filter.nim
rshiv: solved in 2f0e923b3e6c2494bdd639ee6ad352f56bd2f434

CHANGELOG.md
```diff
@@ -1,3 +1,24 @@
+## 2021-07-26 v0.6.1
```
oskarth, line 1: Are we doing a patch release in this PR? @jm-clius 

Usually we add it to the file and then once we cut release we add the actual release version and name. This process is decoupled from any individual PR.

jm-clius: This should just be under the heading `## Next version`. No need to provide release summary or summary of supported `libp2p protocols`. We just add our `Changes` as we go along and compile them to release notes once we determine a release is necessary.

rshiv: resolves in 5a7869620e2a3dbd9d63e441b7cc21f29bb8c1ee.

CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
+##  Next version
+
+This patch release contains the following fix:
+- This Patch introduces Filter Protocol timeout.
+- Support for full secure websockets.
```
jm-clius, line 5:
```suggestion
This release contains the following:
```

A good idea is to take a look at one of the past versions of CHANGELOG before a release, e.g. [here](https://github.com/status-im/nim-waku/blob/d1e06fa17ab7989e487b095548543df0936399fb/CHANGELOG.md), to see how the `Next version` section is structured.

The specifics about `filter` protocol and `websockets` should be moved to a subheading below.

CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
...
+
+### Changes
+- The WakuInfo Object field of listenStr is depreciated and is now replaced with listenAddresses
```
jm-clius, line 8:
```suggestion
- The WakuInfo Object field of `listenStr` is deprecated and is now replaced with `listenAddresses`
```

waku/v2/node/config.nim
```diff
@@ -129,6 +129,11 @@ type
       defaultValue: ""
       name: "filternode" }: string
     
+    filterTimeout* {.
+      desc: "Timeout for filter node in seconds.",
+      defaultValue: 86400
```
jm-clius, line 134: Maybe add a comment to say this translates to `24h`. Was this value chosen arbitrarily, or do we have good reason to believe this is a reasonable default? My own gut-feeling would be a default of ~4h, which is more or less the max that most mobile networks allow devices to be offline before forcing them to reregister. Of course, some applications (especially with high message rates) may set this to something much, much shorter.

waku/v2/node/wakunode2.nim
```diff
@@ -55,6 +55,9 @@ const clientId* = "Nimbus Waku v2 node"
 # Default topic
 const defaultTopic = "/waku/2/default-waku/proto"
 
+# Default Waku Filter Timeout
+const WakuFilterTimeout: Duration = chronos.days(1)
```
jm-clius, line 59: I think you should be able to assign the const using simpler syntax, e.g.
`WakuFilterTimeout = 1.days` or similar

waku/v2/protocol/waku_filter/waku_filter.nim
```diff
@@ -187,44 +188,63 @@ method init*(wf: WakuFilter) =
...
   wf.handler = handle
   wf.codec = WakuFilterCodec
+  wf.timeout = filterTimeout
 
-proc init*(T: type WakuFilter, peerManager: PeerManager, rng: ref BrHmacDrbgContext, handler: MessagePushHandler): T =
+proc init*(T: type WakuFilter, peerManager: PeerManager, rng: ref BrHmacDrbgContext, handler: MessagePushHandler,timeout: Duration = WakuFilterTimeout): T =
   let rng = crypto.newRng()
   var wf = WakuFilter(rng: rng,
                       peerManager: peerManager, 
                       pushHandler: handler)
```
jm-clius, line 197: If you add `timeout` to the `WakuFilter(...` object creation, you don't have to pass the argument to the second `init()` constructor :)

waku/v2/protocol/waku_filter/waku_filter.nim
```diff
@@ -187,44 +188,63 @@ method init*(wf: WakuFilter) =
...
         if connOpt.isSome:
           await connOpt.get().writeLP(push.encode().buffer)
         else:
           # @TODO more sophisticated error handling here
+          handleMessageFailed = true
+          failedSubscriber = subscriber
           error "failed to push messages to remote peer"
           waku_filter_errors.inc(labelValues = [dialFailure])
         break
+  if handleMessageFailed:
```
jm-clius, line 245: One thing I'm missing is how subscriber's state can be cleared in the `failedSubscriber` list? In other words, if one connection attempt fails, but the next is successful, how does the subscriber get removed from the list?

jm-clius (changes requested):
Thanks for the progress! Some comments. Most important, I think, is the question of clearing failed state after a success.

waku/v2/node/config.nim
rshiv: The time was arbitary, changed the same to 4h in commit bfc3104ca29a499cc4512975ba8e90e7f6071342

waku/v2/protocol/waku_filter/waku_filter.nim
rshiv: resolved in bfc3104ca29a499cc4512975ba8e90e7f6071342

waku/v2/protocol/waku_filter/waku_filter.nim
rshiv: resolved in bfc3104ca29a499cc4512975ba8e90e7f6071342

rshiv:
> Thanks for the progress! Some comments. Most important, I think, is the question of clearing failed state after a success.

Yes, a single failure would have started the counter and wouldn't have removed from failed peer list , fixed in bfc3104ca29a499cc4512975ba8e90e7f6071342

oskarth (approved):

waku/v2/protocol/waku_filter/waku_filter.nim
```diff
@@ -188,43 +189,72 @@ method init*(wf: WakuFilter) =
...
   wf.init()
-  
   return wf
 
 proc setPeer*(wf: WakuFilter, peer: RemotePeerInfo) =
   wf.peerManager.addPeer(peer, WakuFilterCodec)
   waku_filter_peers.inc()
 
+#clear the failed peer table if subscriber was able to connect.
+proc handleClientSuccess*(wf: WakuFilter, subscribers: seq[Subscriber]){.raises: [Defect, KeyError], async.} =
```
jm-clius, line 206: Any reason these are public (`*`) and `async`? Afaics this can be done synchronously? In general, functions and procs should be private, unless there's a specific reason to expose this on the public API.

waku/v2/protocol/waku_filter/waku_filter.nim
```diff
@@ -188,43 +189,72 @@ method init*(wf: WakuFilter) =
...
           await connOpt.get().writeLP(push.encode().buffer)
+          connectedSubscribers.add(subscriber)
         else:
           # @TODO more sophisticated error handling here
+          handleMessageFailed = true
+          failedSubscriber.add(subscriber)
           error "failed to push messages to remote peer"
           waku_filter_errors.inc(labelValues = [dialFailure])
         break
+  discard handleClientSuccess(wf, connectedSubscribers)
```
jm-clius, line 255: Futures (implicit return of `{.async.}`) should not be discarded (see: https://hackmd.io/@vac/BJbgn4QT_). Use `asyncSpawn` instead. However, I'm not sure that this should be a future at all. See my comment elsewhere.

jm-clius (approved):
In general, lgtm. One or two minor comments. When addressed, I think this can be merged. Thanks!

waku/v2/protocol/waku_filter/waku_filter.nim
rshiv: resolves in d91eaeb6f12c7deb56d65fb0e275f119d2e56158

waku/v2/protocol/waku_filter/waku_filter.nim
rshiv: resolves in d91eaeb6f12c7deb56d65fb0e275f119d2e56158

PR #784, opened 2021-11-24
Update announced addrs in switch

### LINKED ISSUE #783: Announced addresses in `identify`
There is [an open issue in `nim-libp2p`](https://github.com/status-im/nim-libp2p/issues/429) where only the bound addresses of a host are included during the `identify` protocol exchange. This leads to the [meaningless AnyAddress `0.0.0.0`](https://github.com/status-im/nim-libp2p/issues/450) being exposed to peers in `identify`, rather than the external listening address.

For `status-go`, this prohibits using `identify` in a meaningful way to update peer information after DNS discovery: https://github.com/status-im/status-go/pull/2438

### Possible workaround

A temporary workaround in `nim-waku` would be to update the local `switch` `peerInfo` with the external announced/listening addresses. This has to be done _after_ the `switch` has started and initialised all its transports.

jm-clius:
This PR closes #783 

It adds a workaround for https://github.com/status-im/nim-libp2p/issues/429 where the bound addresses (instead of externally significant addresses) are announced by a `nim-libp2p` switch, specifically as part of the `identify` procedure.

This allows other clients to use the info received during `identify` to update local addresses, where necessary. One example where this is necessary is for status-go integration of Waku v2, where a table with remote peer info is populated after DNS discovery using `identify` (https://github.com/status-im/status-go/pull/2438).

cc @richard-ramos

### DIFF
--- a/waku/v2/node/wakunode2.nim
+++ b/waku/v2/node/wakunode2.nim
@@ -125,6 +125,22 @@ proc removeContentFilters(filters: var Filters, contentFilters: seq[ContentFilte
   
   debug "filters modified", filters=filters
 
+proc updateSwitchPeerInfo(node: WakuNode) =
+  ## TODO: remove this when supported upstream
+  ## 
+  ## nim-libp2p does not yet support announcing addrs
+  ## different from bound addrs.
+  ## 
+  ## This is a temporary workaround to replace
+  ## peer info addrs in switch to announced
+  ## addresses.
+  ## 
+  ## WARNING: this should only be called once the switch
+  ## has already been started.
+  
+  if node.announcedAddresses.len > 0:
+    node.switch.peerInfo.addrs = node.announcedAddresses
+
 template tcpEndPoint(address, port): auto =
   MultiAddress.init(address, tcpProtocol, port)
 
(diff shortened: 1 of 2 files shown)


richard-ramos (approved):

waku/v2/node/wakunode2.nim
```diff
@@ -125,6 +125,22 @@ proc removeContentFilters(filters: var Filters, contentFilters: seq[ContentFilte
...
+  ## TODO: remove this when supported upstream
+  ## 
+  ## nim-libp2p does not yet support announcing addrs
+  ## different from bound addrs.
+  ## 
+  ## This is a temporary workaround to replace
+  ## peer info addrs in switch to announced
+  ## addresses.
+  ## 
+  ## WARNING: this should only be called once the switch
```
rshiv, line 138: In peerInfo.addrs an empty addrs would be indication that the switch is not built before and can be used to prevent update if switch is not built.

rshiv (approved):
LGTM

waku/v2/node/wakunode2.nim
jm-clius: Right, I don't think it will indicate though if the switch has been `start`ed or not? Which is what we want to avoid, as the switch needs to start its transports on the bind addresses, not external.

oskarth (approved):

PR #785, opened 2021-11-24
Upped Jenkins CI `make` verbosity to 1

jm-clius:
Upped Jenkins CI verbosity.


jakubgs (commented):
If we want to change that for all builds that's fine, but you can do that manually for a specific build too:

image

image

Not sure if we want to flood build logs for ALL builds. But up to you guys.

jakubgs (approved):
Fine by me.

PR #789, opened 2021-12-01
Add support for RFC31 ENR

### LINKED ISSUE #755: Support for `multiaddrs` in ENR
## Background

[RFC 31 introduced](https://github.com/vacp2p/rfc/pull/465) a method of encoding `multiaddrs` in ENR.

## Acceptance criteria

`nim-waku` needs to implement support for these, to be able to:
1. Parse ENR records containing `multiaddrs` keys
2. Construct ENR from `multiaddrs` according to RFC 31 - this does not affect traditional TCP transport addresses. which are still constructed according to conventional EIP-778 keys.

jm-clius:
This PR closes #755

As a side-effect it fixes https://github.com/status-im/nim-waku/issues/794

It introduces support for RFC31 compliant ENR creation for Waku v2 nodes.

Changes include:
- support and tests for `multiaddrs` field creation
- support and tests for `multiaddrs` field decoding
- support and tests for `waku2` flag field
- refactoring of address management in wakunode2 to improve readability
- bugfix for `listenStr` when there are no external addresses

@rshiv in order to populate the ENR `multiaddrs`, I've had to refactor the way announcedAddresses gets populated.
I think we can probably simplify things much more still, but for now I've done the minimal required for this PR.
See if the changes make sense to you.
Changes include:
- making sure we announce host addresses if no external addresses available
- change how the ws/wss flag gets added
- making wsHostAddress optional
- removing log on ws and wss both being enabled: I don't think this is an error condition: it should simply mean we prefer wss and continue from there on.

### DIFF
--- a/tests/v2/test_enr_utils.nim
+++ b/tests/v2/test_enr_utils.nim
@@ -1,0 +1,65 @@
+{.used.}
+
+import
+  testutils/unittests,
+  std/options,
+  stew/byteutils,
+  chronos,
+  ../../waku/v2/utils/wakuenr,
+  ../test_helpers
+
+procSuite "ENR utils":
+
+  asyncTest "Parse multiaddr field":
+    let
+      reasonable = "0x000A047F0000010601BADD03".hexToSeqByte()
+      tooLong = "0x000B047F0000010601BADD03".hexToSeqByte()
+      tooShort = "0x000A047F0000010601BADD0301".hexToSeqByte()
+      gibberish = "0x3270ac4e5011123c".hexToSeqByte()
+      empty = newSeq[byte]()
+    
+    ## Note: we expect to fail optimistically, i.e. extract
+    ## any addresses we can and ignore other errors.
+    ## Worst case scenario is we return an empty `multiaddrs` seq.    
+    check:
+      # Expected case
+      reasonable.toMultiAddresses().contains(MultiAddress.init("/ip4/127.0.0.1/tcp/442/ws")[])
+      # Buffer exceeded
+      tooLong.toMultiAddresses().len() == 0
+      # Buffer remainder
+      tooShort.toMultiAddresses().contains(MultiAddress.init("/ip4/127.0.0.1/tcp/442/ws")[])
+      # Gibberish
+      gibberish.toMultiAddresses().len() == 0
+      # Empty
+      empty.toMultiAddresses().len() == 0
+
+  asyncTest "Init ENR for Waku Usage":
+    let
+      enrIp = ValidIpAddress.init("127.0.0.1")
+      enrTcpPort, enrUdpPort = Port(60000)
+      enrKey = PrivateKey.random(Secp256k1, rng[])[]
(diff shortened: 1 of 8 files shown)


jm-clius:
Note on the CI failures: These seem to be unrelated to this specific PR. Being tracked here: https://github.com/status-im/nim-waku/issues/738

oskarth (approved):

rshiv (approved):
LGTM

tests/v2/test_enr_utils.nim
```diff
@@ -0,0 +1,65 @@
...
+      tooShort = "0x000A047F0000010601BADD0301".hexToSeqByte()
+      gibberish = "0x3270ac4e5011123c".hexToSeqByte()
+      empty = newSeq[byte]()
+    
+    ## Note: we expect to fail optimistically, i.e. extract
+    ## any addresses we can and ignore other errors.
+    ## Worst case scenario is we return an empty `multiaddrs` seq.    
+    check:
+      # Expected case
+      reasonable.toMultiAddresses().contains(MultiAddress.init("/ip4/127.0.0.1/tcp/442/ws")[])
```
D4nte, line 26: Maybe worth adding a `dns4` value.

tests/v2/test_enr_utils.nim
```diff
@@ -0,0 +1,65 @@
...
+      # Empty
+      empty.toMultiAddresses().len() == 0
+
+  asyncTest "Init ENR for Waku Usage":
+    let
+      enrIp = ValidIpAddress.init("127.0.0.1")
+      enrTcpPort, enrUdpPort = Port(60000)
+      enrKey = PrivateKey.random(Secp256k1, rng[])[]
+      wakuFlags = initWakuFlags(false, true, false, true)
+      multiaddrs = @[MultiAddress.init("/ip4/127.0.0.1/tcp/442/ws")[], MultiAddress.init("/ip4/127.0.0.1/tcp/443/wss/p2p/16Uiu2HAm4v86W3bmT1BiH6oSPzcsSr31iDQpSN5Qa882BCjjwgrD")[]]
```
D4nte, line 42: [31/WAKU2-ENR](https://rfc.vac.dev/spec/31/) defines that no peer id should be encoded in the `mutliaddrs` field. Hence I would recommend to not have a peer id in the happy path test.

If you want to stripe the peer id and test it then I'd do a separate test for it.

D4nte (approved):
I just looked at the tests.

As it is the implementation of a spec, I think it's better if the happy path tests follows the spec.
This makes it easier for future dev that look at this implementation to do their own.

tests/v2/test_enr_utils.nim
jm-clius: Makes sense. Done in fffcdb2.

tests/v2/test_enr_utils.nim
jm-clius: Done in fffcdb2

PR #790, opened 2021-12-02
fix: return db error message

richard-ramos:


jakubgs (approved):

jm-clius (approved):
Thanks, Richard. Now to investigate why it was failing in the first place.

PR #795, opened 2021-12-04
Fix empty scripts condition. Logging.

jm-clius:
This PR addresses https://github.com/status-im/nim-waku/issues/791

It ensures that a DB migration fails if it finds no scripts to run for a particular migration.

In addition:
- some extra/more detailed logging, to assist with debugging
- stricter column name requirement in `Message` migration script

### DIFF
--- a/waku/v2/node/storage/sqlite.nim
+++ b/waku/v2/node/storage/sqlite.nim
@@ -247,6 +247,7 @@ proc migrate*(db: SqliteDatabase, path: string, targetVersion: int64 = migration
     ok(true)
   
   else:
+    info "database user_version outdated. migrating.", userVersion=userVersion, targetVersion=targetVersion
     # TODO check for the down migrations i.e., userVersion.value > tragetVersion
     # fetch migration scripts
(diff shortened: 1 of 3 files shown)


oskarth (approved):

waku/v2/node/storage/sqlite.nim
```diff
@@ -260,6 +261,9 @@ proc migrate*(db: SqliteDatabase, path: string, targetVersion: int64 = migration
       return err("failed to filter migration scripts")
     
     let scripts = scriptsRes.value
+    if (scripts.len == 0):
```
staheri14, line 264: Sorry for being late to this PR, could not review it yesterday. Thanks Hanno for addressing the problem.

As we spoke in the call, maybe a more proper check is to see whether there is at least one script for all the versions that fall between `userVersion` and `targetVersion` (excluding `userVersion`), we will probably need a helper function to check this. This also implies that `scripts.len` should be at least `targetVersion-userVersion`.
Having this said, since this PR is particularly to address the fleet issue, maybe the current check `scripts.len == 0` is sufficient.

PR #796, opened 2021-12-07
Solves issue 793

rshiv:
Nim web-socket cannot handle encryption in key for SSL verification, hence Readme should add nodes flag
Closes issue #793


oskarth (approved):
@D4nte can you confirm this addresses issue for you?

oskarth:
I notice https://github.com/status-im/nim-waku/issues/793#issuecomment-988340127 so seems like partially.

Do we want to fix this in the same PR or as separate issue? I'm not clear on what the problem is here

jm-clius (approved):

D4nte:
> @D4nte can you confirm this addresses issue for you?

Yes confirmed it fixes one of the issue I encountered. 

More investigation needed for other issues I am encountering as it may come from the JS side.

PR #797, opened 2021-12-07
Fix migration failure

jm-clius:
PR addresses https://github.com/status-im/nim-waku/issues/791

It fixes the underlying issue causing DB migration failure, by adding the migration scripts to the docker image.


staheri14 (approved):
LGTM!
interesting issue and a nice catch! Scripts were not accessible by the docker container?

oskarth (approved):

jm-clius:
> LGTM! interesting issue and a nice catch! Scripts were not accessible by the docker container?

Correct. I've opened this issue, because I think we can perform resource management much better: https://github.com/status-im/nim-waku/issues/798

PR #800, opened 2021-12-21
Fix incorrect store read buffer limit

### LINKED ISSUE #787: Store node: limit size of `HistoryResponse` messages
Many `store` clients assumes a hard limit of 64kb when reading a `HistoryResponse` from the wire, including `go-waku` and `nim-waku` (for example, [here](https://github.com/status-im/nim-waku/blob/master/waku/v2/protocol/waku_store/waku_store.nim#L526) and [here](https://github.com/status-im/nim-waku/blob/master/waku/v2/protocol/waku_store/waku_store.nim#L551)).

However, a store node does not similarly limit the size of `HistoryResponse` messages and, particularly for open history requests with no pagination, this results in various RPC/buffer errors if the response message size exceeds this maximum.

One solution may be to keep track of history size and to "auto-paginate" once a history response nears 64kb.

cc @staheri14 @richard-ramos

jm-clius:
This PR closes #787

I've decided against the initial proposed solution of varying page size based on some unrelated maximum response size.
Rather, I've explicitly set the `MaxWakuMessageSize` (corresponds to the recommended PubSub default) and derived the input stream limit by multiplying with `MaxPageSize`. A safety factor of 64kB is built in to account for protocol overhead. This seems much simpler than what I originally proposed. LMK what you think @staheri14?

I've used the same approach to find a sensible read limit for other protocols. All protocols, other than `Swap` use an RPC message for request/response, so I've adjusted the limit to whichever of the Request or Response are likely the largest, together with a significant safety factor.

### DIFF
--- a/waku/v2/protocol/waku_message.nim
+++ b/waku/v2/protocol/waku_message.nim
@@ -13,6 +13,9 @@ import
 when defined(rln):
   import waku_rln_relay/waku_rln_relay_types
 
+const
+  MaxWakuMessageSize* = 1024 * 1024 # In bytes. Corresponds to PubSub default
+
 type
   ContentTopic* = string
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ This release contains the following:
 - The WakuInfo Object field of `listenStr` is deprecated and is now replaced with `listenAddresses`
 which is a sequence of string.
 
+### Fixes
+- Increased maximum length for reading from a libp2p input stream to allow largest possible protocol messages, including `HistoryResponse` messages at max size.
+
 ## 2021-11-05 v0.6
 
 Some useful features and fixes in this release, include:
--- a/waku/v2/node/wakunode2.nim
+++ b/waku/v2/node/wakunode2.nim
@@ -640,7 +640,8 @@ proc mountRelay*(node: WakuNode,
     #msgIdProvider = msgIdProvider,
     triggerSelf = triggerSelf,
     sign = false,
-    verifySignature = false
+    verifySignature = false,
+    maxMessageSize = MaxWakuMessageSize
   )
   
   info "mounting relay", relayMessages=relayMessages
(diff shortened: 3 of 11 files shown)


staheri14 (approved):
Thanks for the PR, LGTM!

staheri14:
> A safety factor of 64kB is built in to account for protocol overhead. This seems much simpler than what I originally proposed. LMK what you think @staheri14?

I think the solution is good enough given that the size of a waku message always meets the pubsub message size limit. But if this assumption does not hold then there would be the probability of running to the same error (hit the max allowed stream size).

waku/v2/protocol/waku_message.nim
```diff
@@ -13,6 +13,9 @@ import
 when defined(rln):
   import waku_rln_relay/waku_rln_relay_types
 
+const
+  MaxWakuMessageSize* = 1024 * 1024 # In bytes. Corresponds to PubSub default
```
staheri14, line 17: Does libp2p have a hardcoded const/variable corresponding to this default? thought might be good to use that one to make sure our approximation is always consistent with the most recent default size of pubsub messages.

jm-clius: > I think the solution is good enough given that the size of a waku message always meets the pubsub message size limit. But if this assumption does not hold then there would be the probability of running to the same error (hit the max allowed stream size).

Right. I agree, so I've ensured that we initialise PubSub with this max value (overriding the default max message size). This way, even if for some unlikely reason the underlying PubSub default changes it won't affect any of the Waku protocols. It's a bit hidden in between the other changes, but here I initialise PubSub with our `MaxWakuMessageSize`: https://github.com/status-im/nim-waku/blob/d4eddfced48c1b6e323d495948eb6b8d5c35bd52/waku/v2/node/wakunode2.nim#L644

> Does libp2p have a hardcoded const/variable corresponding to this default? thought might be good to use that one to make sure our approximation is always consistent with the most recent default size of pubsub messages.

It does, but since not all our protocols are PubSub protocols I thought it better to define our own maximum, informed by the PubSub default, which we then apply to all protocols including PubSub (see initialisation line above). This way we're in control of any changes.

PR #801, opened 2021-12-22
Remove node peerInfo copy

### LINKED ISSUE #744: Use libp2p PeerInfo struct

jm-clius:
Closes #744

Removes the local `peerInfo` instance/node property and replaces all uses of this copy with the underlying libp2p `switch.peerInfo`.

A local copy on the node is a bad idea, as the underlying addresses may be manipulated by the `switch` under certain circumstances. The `wakunode2` is not the authority on local peer information.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ This release contains the following:
 ### Changes
 - The WakuInfo Object field of `listenStr` is deprecated and is now replaced with `listenAddresses`
 which is a sequence of string.
+- Removed cached `peerInfo` on local node. Rely on underlying libp2p switch instead.
 
 ## 2021-11-05 v0.6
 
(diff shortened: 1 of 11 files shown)


staheri14 (approved):
LGTM!

oskarth (approved):

PR #803, opened 2021-12-23
Add MessageIdProvider for `WakuRelay`

### LINKED ISSUE #420: MessageIdProvider is not configured
https://github.com/status-im/nim-waku/blob/b3208b343f35baebd2adea0da28bc9bf664d45ee/waku/v2/protocol/waku_relay.nim#L40

If `anonimyze` is set to true to not use an author in the message, a message id function should be provided, otherwise it will use the default message id which will return an incorrect message Id as seen in this function:

```nim
func defaultMsgIdProvider*(m: Message): MessageID =
  let mid = 
    if m.seqno.len > 0 and m.fromPeer.data.len > 0:
      byteutils.toHex(m.seqno) & $m.fromPeer
    else:
      # This part is irrelevant because it's not standard,   <----------------------------------
      # We use it exclusively for testing basically and users should  <--------------------
      # implement their own logic in the case they use anonymization <-----------------
      $m.data.hash & $m.topicIDs.hash   <------------------------------------------------------
  mid.toBytes()
  ```

A suggestion from @dryajov is to use a similar approach to eth2 which is to hash the message and use that as the messageId

This is the function used in nimbus
```nim
func gossipId(data: openArray[byte], valid: bool): seq[byte] =
  # https://github.com/ethereum/eth2.0-specs/blob/v1.0.0/specs/phase0/p2p-interface.md#topics-and-messages
  const
    MESSAGE_DOMAIN_INVALID_SNAPPY = [0x00'u8, 0x00, 0x00, 0x00]
    MESSAGE_DOMAIN_VALID_SNAPPY = [0x01'u8, 0x00, 0x00, 0x00]
  let messageDigest = withEth2Hash:
    h.update(
      if valid: MESSAGE_DOMAIN_VALID_SNAPPY else: MESSAGE_DOMAIN_INVALID_SNAPPY)
    h.update data

  result = newSeq[byte](20)
  result[0..19] = messageDigest.data.toOpenArray(0, 19)

func msgIdProvider(m: messages.Message): seq[byte] =
  try:
    let decoded = snappy.decode(m.data, GOSSIP_MAX_SIZE)
    gossipId(decoded, true)
  except CatchableError:
    gossipId(m.data, false)
 ```

jm-clius:
Closes #420

Adds a `MessageIdProvider` for `relay` messages, based on a hash of the message data (as recommended in the libp2p specs [here](https://github.com/libp2p/specs/blob/master/pubsub/README.md#signature-policy-options)).

### DIFF
--- a/waku/v2/node/wakunode2.nim
+++ b/waku/v2/node/wakunode2.nim
@@ -1,7 +1,7 @@
 {.push raises: [Defect].}
 
 import
-  std/[options, tables, strutils, sequtils, os],
+  std/[hashes, options, tables, strutils, sequtils, os],
   chronos, chronicles, metrics,
   stew/shims/net as stewNet,
   stew/byteutils,
@@ -9,9 +9,9 @@ import
   eth/p2p/discoveryv5/enr,
   libp2p/crypto/crypto,
   libp2p/protocols/ping,
-  libp2p/protocols/pubsub/gossipsub,
+  libp2p/protocols/pubsub/[gossipsub, rpc/messages],
   libp2p/nameresolving/dnsresolver,
-  libp2p/builders,
+  libp2p/[builders, multihash],
   libp2p/transports/[transport, tcptransport, wstransport],
   ../protocol/[waku_relay, waku_message],
   ../protocol/waku_store/waku_store,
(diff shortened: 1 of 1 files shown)


richard-ramos (approved):
:100:

D4nte:
I am not familiar with this topic (no issue encountered with js-waku). Looks good tho.

oskarth (approved):

PR #804, opened 2021-12-23
Added protocol message counters

jm-clius:
Adds a counter for protocol messages (e.g. counting the number of store/filter queries received).

### Why?

We currently have no visibility in how often these protocols get invoked, especially in fleet nodes. There are still some unexplained CPU spikes in the `prod` fleet, for example, and being able to see if this correlates to e.g. a store query will go a long way towards prioritising improvements.

### DIFF
--- a/waku/v2/protocol/waku_store/waku_store.nim
+++ b/waku/v2/protocol/waku_store/waku_store.nim
@@ -38,6 +38,7 @@ export
 declarePublicGauge waku_store_messages, "number of historical messages", ["type"]
 declarePublicGauge waku_store_peers, "number of store peers"
 declarePublicGauge waku_store_errors, "number of store protocol errors", ["type"]
+declarePublicGauge waku_store_queries, "number of store queries received"
 
 logScope:
   topics = "wakustore"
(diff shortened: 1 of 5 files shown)


waku/v2/protocol/waku_store/waku_store.nim
```diff
@@ -671,9 +673,9 @@ proc resume*(ws: WakuStore, peerList: Option[seq[RemotePeerInfo]] = none(seq[Rem
           continue
         
       ws.messages.add(indexedWakuMsg)
-      waku_store_messages.inc(labelValues = ["stored"])
-      
       added = added + 1
+    
+    waku_store_messages.set(ws.messages.len.int64, labelValues = ["stored"])
```
staheri14, line 678: It may not be in the scope of the current PR, though, some of the messages may fail to be stored in the DB, therefore the number set in here would be inconsistent with the actual size of DB.

staheri14 (approved):
Looks good!

waku/v2/protocol/waku_store/waku_store.nim
jm-clius: Good point. This gauge is meant to track the total number of messages in _memory_. With the recently adopted store limits, we expect this to reach a maximum and remain constant - a bit boring to watch on a graph, but a good indication that our memory management works as expected. You're right though that we do not have a good way to keep track of messages in the DB, which won't be similarly limited. I'll see what I can add.

PR #811, opened 2022-01-06
Ensure store response never exceeds `MaxPageSize`

jm-clius:
Small but significant change.

It ensures that, even with no paging info specified in the store query, the store will never return a response that exceeds the specified `MaxPageSize`. In other words, _all HistoryResponses will now be auto-paginated_.

This is necessary, because:
- https://github.com/status-im/nim-waku/pull/800 will not work properly if we don't limit _all_ response sizes
- at least some of the performance issues on the fleet is caused by un-paginated query responses attempting to stream the entire history in a single HistoryResponse (https://github.com/status-im/nim-waku/issues/736)

More store changes coming up: https://github.com/status-im/nim-waku/issues/805

@staheri14, keen to know what you think and if there are any gotchas that I've missed. LMK if you think this should be a provision in the spec as well?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ which is a sequence of string.
 - Metrics: added counters for protocol messages
 
 ### Fixes
-- `HistoryResponse` messages now auto-paginated to a maximum of 100 messages per page
+- All `HistoryResponse` messages are now auto-paginated to a maximum of 100 messages per response
 - Increased maximum length for reading from a libp2p input stream to allow largest possible protocol messages, including `HistoryResponse` messages at max size.
 
 ## 2021-11-05 v0.6
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ which is a sequence of string.
 - Metrics: added counters for protocol messages
 
 ### Fixes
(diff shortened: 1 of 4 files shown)


staheri14 (approved):
The auto-pagination makes sense to me! Thanks!

staheri14:
> @staheri14, keen to know what you think and if there are any gotchas that I've missed. LMK if you think this should be a provision in the spec as well?

Thanks for adding the auto-pagination, makes sense and is essential indeed!
The spec also needs to be updated, here is the PR for it https://github.com/vacp2p/rfc/pull/478/files (feel free to edit)

kaiserd:
Should we add a message to the changelog extending the item for the increased maximum length fix?

Something like:

* Increased maximum length for reading from a libp2p input stream to allow largest possible protocol messages, including HistoryResponse messages at max size. To make this work for all response sizes, all HistoryResponses will now be auto-paginated (i.e. pagination will kick-in even for store queries with pageSize set to 0).

jm-clius:
> Should we add a message to the changelog extending the item for the increased maximum length fix?

Yeah, we should! Thanks.

I added the line in f76e3b4 and shortened it a bit. By adding it before the `"Increased maximum length..."` line the causality should be implicit. LMK if you think we should be clearer.

PR #812, opened 2022-01-07
Improved `store` query performance

jm-clius:
## Background

This closes https://github.com/status-im/nim-waku/issues/805 (albeit using a slightly different approach than envisioned in that issue).
It is a prerequisite for the release (https://github.com/status-im/nim-waku/issues/808) and likely addresses https://github.com/status-im/nim-waku/issues/736 as well, though longer term testing will be required to confirm.

It significantly decreases store query times, especially for large message stores as in our fleets.
Recently fleet performance has deteriorated resulting in unusably slow response times.
In local tests query response time improved from >10min to under 100ms.

## What does it change?

- tries to limit calls to more computationally expensive procs (such as `findIndex`)
- extracts all query criteria before filtering, and then filters only once as a single predicate (this saves as from making multiple copies of the in-memory messages)
- pre-sorts messages before pagination (saves us from making another copy)

## What can we still improve?

Changing the underlying data structure (`WakuStore.messages`) to one that sorts on insert will save us from having to sort for each query.  This would also mean that we would only have to find/filter messages until we have populated one full page of history - currently we filter and sort the entire history for every query, and then extract only a single page to respond to the querying node. I don't believe this is urgent yet, as query times seems to be acceptable for now.

cc @D4nte

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ which is a sequence of string.
 ### Fixes
 - All `HistoryResponse` messages are now auto-paginated to a maximum of 100 messages per response
 - Increased maximum length for reading from a libp2p input stream to allow largest possible protocol messages, including `HistoryResponse` messages at max size.
+- Significantly improved store node query performance
 
 ## 2021-11-05 v0.6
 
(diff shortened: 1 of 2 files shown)


oskarth (approved):
Nice work!

staheri14 (commented):
Though the PR is already merged, I reviewed it (as I had promised), and it LGTM!

PR #815, opened 2022-01-14
Ensure consistent resolution of receiver timestamp

jm-clius:
This PR provides a (very) minimal solution for https://github.com/status-im/nim-waku/issues/813

It does two things:
1. ensures that the receiver timestamp in `WakuMessage` indices are consistently millisecond resolution
2. uses stored `receiverTimestamp` when loading from persistent storage rather than system time (this keeps the index consistent after restarts)

## Longer term solution:

- Timestamps should be `int64` epoch milliseconds rather than `float64`, which is bound to lead to precision errors. Since this requires a data scheme change, protocol change and an update in other Waku v2 clients, I'll postpone this change until after the release.
- I would still propose that the index itself be a simple, time-sortable hash over receiver timestamp and message content. If prepended by the epoch received timestamp, the hash is guaranteed to be unique forever and can be reused even if the algorithm for index computation changes in future. Following this, the message index should be computed only once. Even while loading from the DB, the index should simply be read from the corresponding record and not re-computed. This will allow us to index messages in-memory in the store using a hash table (much faster to traverse), and for better uniqueness guarantee as PRIMARY KEY in the `Message` table in the DB.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ which is a sequence of string.
 - Increased maximum length for reading from a libp2p input stream to allow largest possible protocol messages, including `HistoryResponse` messages at max size.
 - Significantly improved store node query performance
 - Added GossipSub `MessageIdProvider` for `11/WAKU2-RELAY` messages.
+- Store: timestamps of message reception, used for indexing, now have consistent millisecond resolution.
 
 ## 2021-11-05 v0.6
 
--- a/waku/v2/protocol/waku_store/waku_store.nim
+++ b/waku/v2/protocol/waku_store/waku_store.nim
@@ -7,7 +7,7 @@
 # Group by std, external then internal imports
 import
   # std imports
-  std/[tables, times, sequtils, algorithm, options],
+  std/[tables, times, sequtils, algorithm, options, math],
   # external imports
   bearssl,
   chronicles,
@@ -55,15 +55,18 @@ const
 # TODO Move serialization function to separate file, too noisy
 # TODO Move pagination to separate file, self-contained logic
 
-proc computeIndex*(msg: WakuMessage): Index =
-  ## Takes a WakuMessage and returns its Index
+proc computeIndex*(msg: WakuMessage, receivedTime = getTime().toUnixFloat()): Index =
+  ## Takes a WakuMessage with received timestamp and returns its Index.
+  ## Received timestamp will default to system time if not provided.
   var ctx: sha256
   ctx.init()
   ctx.update(msg.contentTopic.toBytes()) # converts the contentTopic to bytes
   ctx.update(msg.payload)
   let digest = ctx.finish() # computes the hash
   ctx.clear()
-  var index = Index(digest:digest, receiverTime: epochTime(), senderTime: msg.timestamp)
+
+  let receiverTime = receivedTime.round(3) # Ensure timestamp has (only) millisecond resolution
+  var index = Index(digest:digest, receiverTime: receiverTime, senderTime: msg.timestamp)
   return index
 
 proc encode*(index: Index): ProtoBuffer =
@@ -466,7 +469,7 @@ proc init*(ws: WakuStore, capacity = DefaultStoreCapacity) =
 
   proc onData(receiverTime: float64, msg: WakuMessage, pubsubTopic:  string) =
     # TODO index should not be recalculated
-    ws.messages.add(IndexedWakuMessage(msg: msg, index: msg.computeIndex(), pubsubTopic: pubsubTopic))
+    ws.messages.add(IndexedWakuMessage(msg: msg, index: msg.computeIndex(receiverTime), pubsubTopic: pubsubTopic))
 
   info "attempting to load messages from persistent storage"
 


oskarth (approved):
> int64

Agree. Also see https://github.com/vacp2p/rfc/issues/448

> the index itself be a simple, time-sortable hash over receiver timestamp and message content.

Agree, is there any strong argument against this?

jm-clius:
> > the index itself be a simple, time-sortable hash over receiver timestamp and message content.
> 
> Agree, is there any strong argument against this?

None that I can think of. @staheri14?

kaiserd (approved):
Looks fine to me :). I also agree with the future plan of moving to a hash based index and eventually switching to a hash table based store. The full timestamp information could still be kept in a data structure like this

```Nim
  StoredWakuMessage* = object
    msg*: WakuMessage
    receiverTime: int64
    senderTime: int64
    pubsubTopic*: string
    index*: ?       # a digest (preserving time order) of msg, receiverTime, senderTime, pubsubTopic
```

staheri14:
Thanks for the PR Hanno!
A few comments:
- For consistency sake, the `senderTime` should also be in milliseconds resolution otherwise we will get to the same issue as the `receiverTime`.
- Spec also needs an update 

- A few points re this comment
> I would still propose that the index itself be a simple, time-sortable hash over receiver timestamp and message content. If prepended by the epoch received timestamp, 

1) If we aim to do so, then it should be the hash of the sender timestamp with the content, not the receiver timestamp as the receiver timestamp is not a correct indication of the time the message is originally generated (especially if in the future we aim to allow cross-store node queries)

2) I am not still clear on how this can improve performance? is it about finding the `Cursor` in O(1)? Or to improve the message retrieval time e.g., for time-range queries? 
I think what you are suggesting is entirely an internal implementation detail and does not need to touch the spec and the structure of the `Index` and `Cursor`. 
Do you want a hash map whose keys are the `Index` object? so a custom hash function is sufficient for that implementation. Can you please elaborate on your proposed solution? @jm-clius 

cc: @oskarth @kaiserd

jm-clius:
Thanks for the questions, @staheri14! Good to think things through. 😄 My answer below may have gone into too much details, so LMK if anything's confusing or if you have other thoughts/questions.

> For consistency sake, the senderTime should also be in milliseconds resolution otherwise we will get to the same issue as the receiverTime.

It may be a good idea to use the same resolution where `senderTime` is generated (outside the `store`), but I don't think the `store` should ever modify the `senderTime` - which should remain consistent across node hops. Since the `store` simply uses the `senderTime` in whatever resolution it receives it, we don't run into the same issue. It's only `receiverTime` which is _generated_ in the `store` that may mismatch between what we write on the wire and what we store in memory, because of how unpredictable `epochTime()` is. Of course, once we move to `int64` milliseconds, it will be applied to both `senderTime` and `receiverTime` and there will be no ambiguity.

> Spec also needs an update

The spec currently states that `receiverTime` is "the UNIX time at which the waku message is received". According to me this still holds exactly. We have only identified that in the nim implementation there could be an _internal_ mismatch between the timestamp representation in-memory (because we use `epochTime()`, which is inconsistent) and the one we "advertise" in the wire. Using millisecond resolution is simply a way to generate the timestamp (i.e. to follow the spec) in a way that does not cause inconsistencies in `nim-waku`. IMO this is then an implementation-only fix, or wdyt?

> If we aim to do so, then it should be the hash of the sender timestamp with the content, not the receiver timestamp as the receiver timestamp is not a correct indication of the time the message is originally generated.

It should be guaranteed unique inside the `store` node, so we can use it as the Primary Key field in the SQL database and as the index/hash in our internal hash table (index:message). According to me the only way to guarantee uniqueness is to use `receiverTime`, as `senderTime` is outside our control and are most often not used in any case. I don't think the index need to have any significance outside the `store` node, so no need for this to be globally "correct".

>  (especially if in the future we aim to allow cross-store node queries)

Index is only a way for us to efficiently structure the stored messages _internally_ and to provide as a cursor to the next page inside _this_ store node. I don't think it will affect cross-store node queries (it's only used as a cursor for "next page" queries, which will be unique to each store node in the current implementation too, or am I missing something?)

> I am not still clear on how this can improve performance?

Currently, each time we receive a query:
1. We make a copy of the entire message history
2. We filter the entire message history from this copy to extract the messages that match a filter into a new sequence.
3. We sort this filtered sequence into yet another copy (at this point we may even have to deduplicate, which will again have to be done across everything and will create another copy of the entire history, but I think if we only apply the filter in `2` _once_ we should be fine without deduplicating :) ).
4. We iterate through the resulting sequence from (3) to find the message that matches the cursor in our query.
5. After all these copies, we only extract a max of 100 messages to fill the page response.

If we have a hash table that's naturally sorted by an index key (for example: https://github.com/status-im/nim-stew/blob/master/stew/sorted_set.nim)
1. No need to make a copy of the history when we receive a query, as the in-memory store becomes read-only
2. Finding the cursor becomes a hash table lookup, which scales well.
3. No need to deduplicate (unique hashes take care of that) or sort (messages are naturally sorted when they're inserted).
4. No need to filter entire history: start at wherever the cursor is pointing and read in the direction of paging (backwards/forwards) until you have filled out a single page of history that matches the filter criteria. Then stop.

> I think what you are suggesting is entirely an internal implementation detail and does not need to touch the spec

Agreed! Changing the timestamps to `int64` will be a spec change, but this will probably be done beforehand in any case. It may make sense to provide a simpler `cursor` to clients though - a simple hash value that clients can use to request the next page of history without having to worry about how that cursor is composed. It will also help us not to have to recompute the hash (needed for lookup) when receiving a query.

> Do you want a hash map whose keys are the Index object? so a custom hash function is sufficient for that implementation.

Yeah, exactly. The idea is centered around this custom hash function so the Index can be used as an efficient, time-sorted key in a hash table.

staheri14:
> It may be a good idea to use the same resolution where `senderTime` is generated (outside the `store`), but I don't think the `store` should ever modify the `senderTime` - which should remain consistent across node hops. Since the `store` simply uses the `senderTime` in whatever resolution it receives it, we don't run into the same issue. It's only `receiverTime` which is _generated_ in the `store` that may mismatch between what we write on the wire and what we store in memory, because of how unpredictable `epochTime()` is. Of course, once we move to `int64` milliseconds, it will be applied to both `senderTime` and `receiverTime` and there will be no ambiguity.

My general point is that all the waku messages should be indexed identically across all the store nodes, this means a `Cursor` obtained from one node can be used to query from another node (or at least discrepancies are not due to some unnoticed type conversions in the json rpc API or other places).
As for the `senderTimeStamp`, assume the case that a node A uploads a message with timestamp t.wxyz to a store node B via json rpc API, and in the string conversions happening under the hood, B receives t.wxy . If later A happens to want to query from `B` for the messages published after its own message, A creates a `Cursor` using its own message data i.e., sets the sender time stamp as t.wxyz (and calculates hashe based on spec description), then its cursor renders invalid on the B side because t.wxyz != t.wxy.  (note that in this scenario I am disregarding the receiver time which by default different at each store node)

I do not see any harm to specify the desired time resolution for sender and receiver time and I think it is prudent to do so.

staheri14:
> The spec currently states that receiverTime is "the UNIX time at which the waku message is received". According to me this still holds exactly. We have only identified that in the nim implementation there could be an internal mismatch between the timestamp representation in-memory (because we use epochTime(), which is inconsistent) and the one we "advertise" in the wire. Using millisecond resolution is simply a way to generate the timestamp (i.e. to follow the spec) in a way that does not cause inconsistencies in nim-waku. IMO this is then an implementation-only fix, or wdyt?


Again, for the cursor reusability (explained in the previous comment), I'd suggest indicating the time resolution in the spec.

staheri14:
> It should be guaranteed unique inside the store node, so we can use it as the Primary Key field in the SQL database and as the index/hash in our internal hash table (index:message). According to me the only way to guarantee uniqueness is to use receiverTime, as senderTime is outside our control and are most often not used in any case. I don't think the index need to have any significance outside the store node, so no need for this to be globally "correct".

From my PoV, store nodes while being independent entities, they serve the same set of waku messages in a fully decentralized fashion. Later, we want them to synchronize and reach an identical store state. Using waku message's receiver time disallows this.
Secondly, I do not see why sender time plus hash does not provide uniqueness?
If two messages have the same payload and are published at the same time, w.h.p. are duplicates. We can also bring the content topic into this equation for extra caution.

>  senderTime is ... most often not used in any case

Why not? all the time-range queries target sender time.

staheri14:
> If we have a hash table that's naturally sorted by an index key (for example: https://github.com/status-im/nim-stew/blob/master/stew/sorted_set.nim)

No need to make a copy of the history when we receive a query, as the in-memory store becomes read-only
Finding the cursor becomes a hash table lookup, which scales well.
No need to deduplicate (unique hashes take care of that) or sort (messages are naturally sorted when they're inserted).
No need to filter entire history: start at wherever the cursor is pointing and read in the direction of paging (backwards/forwards) until you have filled out a single page of history that matches the filter criteria. Then stop

I think in the above scenario the query based on the content topic is disregarded. Even if messages are sorted based on time, they should be filtered based on the queries `contetTopic`.

staheri14:
>  a simple hash value that clients can use to request the next page of history without having to worry about how that cursor is composed. It will also help us not to have to recompute the hash (needed for lookup) when receiving a query.

@jm-clius 

I do not see how knowing how a cursor is composed would worry clients? 

Also, recomputing the hash is not related to the Cursor/Index protobuf def. Right now hash of payloads is not recomputed per query, they are computed once at the message arrival and saved for future queries. The same hash is used to populate the Cursor of history responses.
We shall do a similar thing if we want to update the inputs to the hash i.e.,  we use the `digest` field of the `Index` to represent the hash of payload + content topic + sender time instead.

jm-clius:
Thanks, @staheri14. Let's continue the discussion here: https://forum.vac.dev/t/timestamping-and-indexing-in-the-wakustore/118

PR #816, opened 2022-01-14
result aggregation in resume API

kaiserd:
This PR addresses https://github.com/status-im/nim-waku/issues/809

- updates `resume` and `queryLoop` implementing the logic described in https://github.com/status-im/nim-waku/issues/809
- updates the existing resume test and adds a test for resuming from an empty `peerList`
- updates the [changelog](https://github.com/status-im/nim-waku/blob/master/CHANGELOG.md)
- updates the waku v2 API [description](https://github.com/status-im/nim-waku/blob/master/docs/api/v2/node.md)

The code for awaiting the futures in `resume` is clunky. I could not find a version of `allFutures` that returns a `Future[seq[T]]`.
What I ended up using is `allFinished*[T](futs: varargs[Future[T]]): Future[seq[Future[T]]]` which requires to go through the already finshed futures again.
At least, this allows the futures to be awaited in parallel.
Did I overlook something?

Another point that came up:
Currently, if no peer in the peerList answers, resume exits with an error.
It could instead ask known peers, which is the case when peerList is empty.
What do you think @staheri14 @jm-clius ?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ The full list of changes is below.
 - Removed cached `peerInfo` on local node. Rely on underlying libp2p switch instead
 - Metrics: added counters for protocol messages
 - Waku v2 node discovery now supports [`31/WAKU2-ENR`](https://rfc.vac.dev/spec/31/)
+- resuming the history via `resume` now takes the answers of all peers in `peerList` into consideration and consolidates them into one deduplicated list
 
 ### Fixes
 
--- a/tests/v2/test_waku_store.nim
+++ b/tests/v2/test_waku_store.nim
@@ -715,7 +715,7 @@ procSuite "Waku Store":
         messagesResult.isOk
         messagesResult.value.len == 4
 
-    asyncTest "resume history from an empty list of candidate peers":
+    asyncTest "resume history from a list of offline peers":
       var offListenSwitch = newStandardSwitch(some(PrivateKey.random(ECDSA, rng[]).get()))
       var dialSwitch3 = newStandardSwitch()
       await dialSwitch3.start()
@@ -736,9 +736,12 @@ procSuite "Waku Store":
 
       let successResult = await proto3.resume(some(@[offListenSwitch.peerInfo.toRemotePeerInfo(),
                                                      listenSwitch.peerInfo.toRemotePeerInfo(),
-                                                     listenSwitch.peerInfo.toRemotePeerInfo(),
                                                      listenSwitch2.peerInfo.toRemotePeerInfo()]))
       check:
+        # `proto3` is expected to retrieve 14 messages because:
+        # - the store mounted on `listenSwitch` holds 10 messages (`msgList`)
+        # - the store mounted on `listenSwitch2` holds 7 messages (see `msgList2`)
+        # - both stores share 3 messages, resulting in 14 unique messages in total
         proto3.messages.len == 14
         successResult.isOk
         successResult.value == 14
--- a/tests/v2/test_waku_store.nim
+++ b/tests/v2/test_waku_store.nim
@@ -534,6 +534,8 @@ procSuite "Waku Store":
     let
       key = PrivateKey.random(ECDSA, rng[]).get()
       peer = PeerInfo.new(key)
+      key2 = PrivateKey.random(ECDSA, rng[]).get()
+      # peer2 = PeerInfo.new(key2)
     var
       msgList = @[WakuMessage(payload: @[byte 0], contentTopic: ContentTopic("2"), timestamp: float(0)),
         WakuMessage(payload: @[byte 1],contentTopic: ContentTopic("1"), timestamp: float(1)),
@@ -545,7 +547,19 @@ procSuite "Waku Store":
         WakuMessage(payload: @[byte 7],contentTopic: ContentTopic("1"), timestamp: float(7)),
         WakuMessage(payload: @[byte 8],contentTopic: ContentTopic("2"), timestamp: float(8)),
         WakuMessage(payload: @[byte 9],contentTopic: ContentTopic("1"),timestamp: float(9))]
-            
+
+      msgList2 = @[WakuMessage(payload: @[byte 0], contentTopic: ContentTopic("2"), timestamp: float(0)),
+        WakuMessage(payload: @[byte 11],contentTopic: ContentTopic("1"), timestamp: float(1)),
+        WakuMessage(payload: @[byte 12],contentTopic: ContentTopic("2"), timestamp: float(2)),
+        WakuMessage(payload: @[byte 3],contentTopic: ContentTopic("1"), timestamp: float(3)),
+        WakuMessage(payload: @[byte 4],contentTopic: ContentTopic("2"), timestamp: float(4)),
+        WakuMessage(payload: @[byte 5],contentTopic: ContentTopic("1"), timestamp: float(5)),
+        WakuMessage(payload: @[byte 13],contentTopic: ContentTopic("2"), timestamp: float(6)),
+        WakuMessage(payload: @[byte 14],contentTopic: ContentTopic("1"), timestamp: float(7))]
+
+    #--------------------
+    # setup default test store
+    #--------------------
     var dialSwitch = newStandardSwitch()
     await dialSwitch.start()
 
@@ -562,6 +576,28 @@ procSuite "Waku Store":
     for wakuMsg in msgList:
       # the pubsub topic should be DefaultTopic
       await proto.handleMessage(DefaultTopic, wakuMsg)
+
+    #--------------------
+    # setup 2nd test store
+    #--------------------
+    var dialSwitch2 = newStandardSwitch()
+    await dialSwitch2.start()
+
+    # to be connected to
+    var listenSwitch2 = newStandardSwitch(some(key2))
+    await listenSwitch2.start()
+
+    let proto2 = WakuStore.init(PeerManager.new(dialSwitch2), crypto.newRng())
+
+    proto2.setPeer(listenSwitch2.peerInfo.toRemotePeerInfo())
+
+    listenSwitch2.mount(proto2)
+
+    for wakuMsg in msgList2:
+      # the pubsub topic should be DefaultTopic
+      await proto2.handleMessage(DefaultTopic, wakuMsg)
+
+
     
     asyncTest "handle temporal history query with a valid time window":
       var completionFut = newFuture[bool]()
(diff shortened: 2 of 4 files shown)


waku/v2/protocol/waku_store/waku_store.nim
```diff
@@ -612,14 +612,23 @@ proc queryFromWithPaging*(w: WakuStore, query: HistoryQuery, peer: RemotePeerInf
...
-
-  debug "failed to resolve the query"
-  return err("failed to resolve the query")
+  ## loops through the candidateList in order and sends the query to each
+  ## once all responses have been received, the retrieved messages are consolidated into one deduplicated list
+  ## if no messages have been retrieved, the returned future will resolve into a MessagesResult result holding an empty seq.
+  var futureList: seq[Future[MessagesResult]]
+  for peer in candidateList.items:
+    futureList.add(w.queryFromWithPaging(query, peer))
+  futureList = await allFinished(futureList) # TODO: is there a proc that returns a Future[seq[T]?
```
jm-clius, line 621: Any advantage to doing this rather than just use `await allFutures()`? Seems like that will at least avoid double-assigning to `futureList`? It will still be necessary to retrieve the completed values iteratively, though, as you did below.

waku/v2/protocol/waku_store/waku_store.nim
```diff
@@ -612,14 +612,23 @@ proc queryFromWithPaging*(w: WakuStore, query: HistoryQuery, peer: RemotePeerInf
...
+  var messagesList: seq[WakuMessage]
+  # all futures have already been awaited, so the following loop runs without actually waiting
+  for fut in futureList:
+    let successResult = await fut
```
jm-clius, line 625: Since we "know" that these futures have all completed, why not rather just `read` the value (with a sanity check for completeness too). I think this will improve readability.
For example:
```
if fut.completed():
  let successResult = fut.read()
```
or something similar should be safe to do.

waku/v2/protocol/waku_store/waku_store.nim
```diff
@@ -612,14 +612,23 @@ proc queryFromWithPaging*(w: WakuStore, query: HistoryQuery, peer: RemotePeerInf
...
+    if successResult.isOk: messagesList.insert(successResult.value)
```
jm-clius, line 626: I think it may be possible to map `futureList` to `messagesList` using the [`mapIt()`](https://nim-lang.org/docs/sequtils.html#mapIt.t%2Ctyped%2Cuntyped) template and `concat`ting the resulting sequence-of-sequences. Compressing too much logic into a single line is probably not a good idea, so feel free to ignore this suggestion, or just applying it to some steps. :)

jm-clius (approved):
This LGTM! Thanks. I've left some comments about possible ways to make the code more readable, but feel free to apply only where you deem necessary/helpful.

waku/v2/protocol/waku_store/waku_store.nim
jm-clius: You could also encompass this logic into the `op proc` parameter of a `mapIt` function and directly map your `seq[Future[MessagesResult]]` to a `seq[MessagesResult]`. With a bit more logic, and concatenating the different responses, the original `seq` of futures could also be mapped to a `seq[WakuMessage]` using a single `mapIt` function.

jm-clius:
> The code for awaiting the futures in resume is clunky. I could not find a version of allFutures that returns a Future[seq[T]].

Seems like this did exist, but was deprecated for a number of reasons: https://github.com/status-im/nim-chronos/blob/7dc58d42b6905a7fd7531875fa76060f8f744e4e/chronos/asyncfutures2.nim#L689

You could use the `mapIt` function for sequences to map to your desired `seq`. See my comment elsewhere on the PR.

> Currently, if no peer in the peerList answers, resume exits with an error.

I think this is fine. We could decide on a future "enhancement" where we use an existing peer, but IMO if someone calls the API with an explicit list of peers it's better to return an error than transparently contacting a peer that was _not_ provided.

waku/v2/protocol/waku_store/waku_store.nim
kaiserd: Thank you. I will apply this change. As I am very new to Nim, I did not know better 😅. At first I thought I could save the following iteration by directly getting a list with the results of all successfully resolved futures, and then replace the allFInished call with the respective call. But as you pointed out, using allFinished, I still have to iterate.

waku/v2/protocol/waku_store/waku_store.nim
kaiserd: Thank you. I did not know about `read()`. I will apply that.

waku/v2/protocol/waku_store/waku_store.nim
kaiserd: Thank you. I was looking for this, but then decided to take syntactically simple approach due to lack of Nim knowledge.
I will look into it an apply the change you suggested.

kaiserd:
Thank you for the feedback.
In pushed commits addressing your feedback.
The second commit addresses the functional style of constructing the `messagesList`.
In Rust, there are combinators like filterMap, which would allow making this more concise. I will check later, if Nim offers something similar.

waku/v2/protocol/waku_store/waku_store.nim
```diff
@@ -618,12 +618,15 @@ proc queryLoop(w: WakuStore, query: HistoryQuery, candidateList: seq[RemotePeerI
...
-    let successResult = await fut
-    if successResult.isOk: messagesList.insert(successResult.value)
+  await allFutures(futureList) # all(), which returns a Future[seq[T]], has been deprecated
+
+  let messagesList = futureList
+    .filterIt(it.completed()) # just a sanity check. These futures have been awaited before using allFutures()
+    .mapIt(it.read())
+    .filterIt(it.isOk)
+    .mapIt(it.value)
+    .concat()
```
jm-clius, line 628: I like how explicit and readable this is, but it may be worth considering that each `mapIt` and `filterIt` returns a new `seq` which for very large sequences could be inefficient. Not really a problem here, as the `futureList` will be small.
The way to fix that would be to call [a `map` proc](https://nim-lang.org/docs/sequtils.html#map%2CopenArray%5BT%5D%2Cproc%28T%29) only once, but with a slightly more advanced logic. Something like:
```
proc mapToWakuMessages(in: Future[MessagesResult]): seq[WakuMessages] =
  if in.completed() and in.read().isOk():
    in.read().value
  else:
    @[]

let messagesList = futureList.map(mapToWakuMessages).concat()
```
(Might be some errors here, but the rough idea is to minimise seq copies)

kaiserd:
I updated to a single map proc. However, not giving the proc directly as a parameter raises the following error:
```
Error: can raise an unlisted exception: CatchableError
```
So, for now, I defined the proc inline. (Using the `=>` sugar syntax would be more readable, imho.)
I will check, what caused that error.

kaiserd:
Thanks @jm-clius for the tip to annotate the exception.
With that, we could also use this code if perferred:

```nim
  proc mapToWakuMessages(it: Future[MessagesResult]): seq[WakuMessage] {.raises: [CatchableError, Defect].}=
    if it.completed() and it.read().isOk():
      it.read().value
    else:
      @[]

  let messagesList = futureList.map(mapToWakuMessages).concat()
```

staheri14:
> It could instead ask known peers, which is the case when peerList is empty.

tests/v2/test_waku_store.nim
```diff
@@ -679,6 +715,16 @@ procSuite "Waku Store":
         messagesResult.isOk
         messagesResult.value.len == 4
 
+    asyncTest "resume history from an empty list of candidate peers":
```
staheri14, line 718: `empty list of candidate peers` or `offline peers`?

tests/v2/test_waku_store.nim
```diff
@@ -690,12 +736,16 @@ procSuite "Waku Store":
 
       let successResult = await proto3.resume(some(@[offListenSwitch.peerInfo.toRemotePeerInfo(),
                                                      listenSwitch.peerInfo.toRemotePeerInfo(),
-                                                     listenSwitch.peerInfo.toRemotePeerInfo()]))
+                                                     listenSwitch.peerInfo.toRemotePeerInfo(),
+                                                     listenSwitch2.peerInfo.toRemotePeerInfo()]))
       check:
-        proto3.messages.len == 10
+        proto3.messages.len == 14
```
staheri14, line 742: Wondering why 14? (ok, I found it :), `listenSweitch2` has 4 waku messages that `listenSwitch` does not have, though it would be good to add a comment and explain that, would be helpful for debugging)

tests/v2/test_waku_store.nim
```diff
@@ -690,12 +736,16 @@ procSuite "Waku Store":
 
       let successResult = await proto3.resume(some(@[offListenSwitch.peerInfo.toRemotePeerInfo(),
                                                      listenSwitch.peerInfo.toRemotePeerInfo(),
-                                                     listenSwitch.peerInfo.toRemotePeerInfo()]))
+                                                     listenSwitch.peerInfo.toRemotePeerInfo(),
```
staheri14, line 739: Wondering why did you add `listenSwitch` twice? if you have a specific test scanrio, please add a small comment

staheri14 (approved):
Congratulations @kaiserd for your first nim-waku PR!
I have left a few comments. Going to approve the PR, but please address them before merging :)

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ This release contains the following:
 which is a sequence of string.
 - Removed cached `peerInfo` on local node. Rely on underlying libp2p switch instead.
 - Metrics: added counters for protocol messages
+- resuming the history via `resume` now takes the answers of all peers in `peerList` into consideration and consolidates them into one deduplicated list
```
oskarth, line 14:
```suggestion
- Resuming history via `resume` now takes the answers of all peers in `peerList` and consolidates them into one deduplicated list
```

oskarth (approved):
LGTM, nothing to add to above. Congrats on your first PR!

kaiserd:
@staheri14 thank you for the feedback. I addressed your feedback and added comments.

> Can you please elaborate on this?
> 
> > It could instead ask known peers, which is the case when peerList is empty.

I thought it might make sense to ask a random peer like here
https://github.com/status-im/nim-waku/blob/024128b39e77935268e049769ffbe4a6987bf5aa/waku/v2/protocol/waku_store/waku_store.nim#L723
in case no `peerList` is provided to `resume`.
But I agree with @jm-clius   comment above, that this behavior might not be expected.
So I would just leave it as it is.

kaiserd:
@jm-clius @staheri14 is your feedback sufficiently addressed? Just want to make sure before merging :).
@jm-clius Should I keep the current version of building `messagesList?

jm-clius:
Yeah, you can squash and merge - I notice not all commits are signed, which you'll need to do beforehand, but afaict you addressed all comments from all reviewers, CI is passing (as well as it does :) ) and the `messagesList` building makes sense to me. 😄

kaiserd:
Yes. The signatures were lost during a rebase :(. I reapplied them.

kaiserd:
I rebased again on the v0.7 commit.
@jm-clius Should this PR be part of v0.7. If not, I have to update the changelog.

jm-clius:
No, I think it's OK to merge after the release

kaiserd:
OK. I am waiting for @staheri14 to check if I applied her feedback as she expected and then merge :).

staheri14:
> OK. I am waiting for @staheri14 to check if I applied her feedback as she expected and then merge :).

@kaiserd LGTM!

PR #817, opened 2022-01-17
ci: add architecture specification to agent label

jakubgs:
Should fix errors like:
```
clang: error: the clang compiler does not support '-march=native'
```
Caused by hitting the Apple M1 `arm64` host.


jm-clius (approved):
Thanks!

PR #829, opened 2022-01-24
stop sockets after use in test_waku_store.nim

kaiserd:
This PR closes switches after use in test_waku_store.
[Stopping switches](https://github.com/status-im/nim-libp2p/blob/58f383e661521314df314e7096c24db5a7490372/libp2p/switch.nim#L209) frees the transport resources, which means closing TCP ports in waku_store_test.
(A TCP-based switch holds a set of `StreamServer`s in its `TcpTransport` object, which when [stopped](https://github.com/status-im/nim-chronos/blob/519ca463dfd060dfec64a221b8e39e06a70860d3/chronos/transports/stream.nim#L1753),
 waits to be [closed](https://github.com/status-im/nim-chronos/blob/519ca463dfd060dfec64a221b8e39e06a70860d3/chronos/transports/stream.nim#L1806), which then [closes](https://github.com/status-im/nim-chronos/blob/519ca463dfd060dfec64a221b8e39e06a70860d3/chronos/transports/stream.nim#L1804) the socket.)

### DIFF
--- a/tests/v2/test_waku_store.nim
+++ b/tests/v2/test_waku_store.nim
@@ -54,6 +54,11 @@ procSuite "Waku Store":
 
     check:
       (await completionFut.withTimeout(5.seconds)) == true
+
+    # free resources
+    await allFutures(dialSwitch.stop(),
+      listenSwitch.stop())
+
   asyncTest "handle query with multiple content filters":
     let
       key = PrivateKey.random(ECDSA, rng[]).get()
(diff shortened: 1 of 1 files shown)


kaiserd:
I added a minor change renaming in `asyncTest "resume message history"`
dialSwitch2 -> dialSwitch3
proto2 -> proto3
It worked before because the local proto2 and dialSwitch2 shadow the outer ones; however, imho, it is clearler like this.

jm-clius:
Mmm, seems like CI is now failing for `jenkins/prs/macos` and `jenkins/prs/windows`. I've restarted those jobs, just in case it was due to a temporary resource contention error. I wouldn't expect them to fail anymore after the recent CI fixes. (`Test / windows` failure still expected.)

jm-clius (approved):
LGTM. Thanks.

oskarth (approved):
Thanks for the cleanup PR!

Issue #828, opened 2022-01-24
`nwaku` for operators trial


jm-clius:
# Background

One of the main takeaways from the [Q4 2022 Vac Sustainability and Business Workshop](https://forum.vac.dev/t/vac-sustainability-and-business-workshop/116#q1-goals-draft-37) was _the need for `nwaku` to settle into a niche of being an easy-to-install, easy-to-monitor, lightweight and robust option for user-run Waku v2 nodes._

Currently the main obstacles to achieving this goal is:
- incomplete or unclear documentation for users on how to set up and dimension their own `nwaku` nodes for different protocols
- limited pre-compiled binaries for easy setup (users usually need to clone the repo and build `nwaku` themselves, seldomly a smooth process)
- limited insight into actual robustness of running a `nwaku` node in the long term, especially for lesser-used protocols, such as `filter` and `lightpush`
- limited visibility (e.g. a metric) of the number of user-run nodes in the default Waku v2 network and number of applications using Waku v2
- lack of ambient peer discovery

# What do we want to achieve as a first step?

This issue tracks the work necessary to launch a Vac-internal operator-run node trial. Specifically, we want:

## Goal 1: Team-internal operators set up own `nwaku` nodes

Operators to follow instructions to set up their own `nwaku` node using a precompiled binary that:
- participates and is discoverable in the default `relay` network
- provides a dimensioned `store`
- provide `lightpush` and `filter` capabilities

Between team members we could probably run a hybrid network of nodes on at least Windows, Linux and MacOS.

## Goal 2: Operators monitor their nodes

Operators to follow instructions to monitor their nodes, e.g. using an included metrics dashboard, looking at logs, etc.

## Goal 3: Operators to provide informal, weekly updates

This should track two main qualitative metrics:
1. "How easy is it to operate a node?"
- feedback on documentation/instructions. What are the gaps in knowledge that should be bridged/obscured from end users?
- are the included tooling enough for operation?
- feedback re available metrics/logs and their usefulness
- feedback re metrics dashboard
2. "How well does the node run?"
- i.e. how stable has it been running as far as you can tell from metrics/logs
- how many stable connections does the node maintain
- how many new, useful peers has the node discovered (only PX during trial)

As a team we can use the various `chat2` clients to simulate different scenarios during the trial, to ensure we properly test `filter`, `lightpush`, etc.

# What do we have already?

[...]

PR #831, opened 2022-01-25
Improve node metrics

jm-clius:
Minor PR towards #828 

It adds:
1. A dimension to the message counter to keep track of the number of `contentTopic`s in the network. This should enable us to build Prometheus queries to view i.a. the number of applications using the network.
2. A metric to track the number and type of connections initiated by the nwaku node. This will allow us to monitor number of connections initiated due to static configuration, dns discovery, discv5, API calls, and more.
3. Some error handling for failed connections initiated by this node.

For (2) it was necessary to clean up the `connect` and `dial` API in `wakunode2` (somewhat).

There is some parallel work ongoing to update the bundled Grafana dashboard accordingly.

### DIFF
--- a/waku/v2/node/wakunode2.nim
+++ b/waku/v2/node/wakunode2.nim
@@ -252,7 +252,10 @@ proc subscribe(node: WakuNode, topic: Topic, handler: Option[TopicHandler]) =
       if (not node.wakuStore.isNil):
         await node.wakuStore.handleMessage(topic, msg.value())
 
-      waku_node_messages.inc(labelValues = ["relay", msg.value().contentTopic])
+      # Increase message counter
+      let ctLabel = if msg.value().contentTopic.len > 0: msg.value().contentTopic
+                    else: "none"
+      waku_node_messages.inc(labelValues = ["relay", ctLabel])
 
   let wakuRelay = node.wakuRelay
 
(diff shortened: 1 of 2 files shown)


staheri14 (approved):
Looks good!

waku/v2/node/wakunode2.nim
```diff
@@ -433,7 +434,7 @@ proc mountFilter*(node: WakuNode, filterTimeout: Duration = WakuFilterTimeout) {
     info "push received"
     for message in msg.messages:
       node.filters.notify(message, requestId) # Trigger filter handlers on a light node
-      waku_node_messages.inc(labelValues = ["filter"])
+      waku_node_messages.inc(labelValues = ["filter", message.contentTopic])
```
oskarth, line 437: What if contentTopic isn't filled out? IIRC it is a SHOULD, do we want to assert this?

oskarth (approved):

waku/v2/node/wakunode2.nim
jm-clius: Good point. It will write an empty string to Prometheus, which I imagined would still be query-able, but [quick search](https://github.com/grafana/grafana/issues/19410) makes it seem like it won't be straightforward to extract label values. Probably best to add a placeholder label for messages without content topic, like `"none"`.

jm-clius: Addressed in ae3e1cc

Issue #834, opened 2022-01-26
Changing timestamp type to int64


staheri14:
Following the discussion in https://github.com/vacp2p/rfc/issues/448, the decision has been made to use `int64` type to represent `timestamp` values across wakuv2 protocols where the time values going to indicate the Unix epoch time in milliseconds. This includes
-  [`timestamp`](https://github.com/status-im/nim-waku/blob/4421b8de0074574c3740b71a197b4f6eeb90c1c5/waku/v2/protocol/waku_message.nim#L27) field of [`WakuMessage`](https://rfc.vac.dev/spec/14/#wakumessage). 
- [`senderTime`](https://github.com/status-im/nim-waku/blob/4421b8de0074574c3740b71a197b4f6eeb90c1c5/waku/v2/utils/pagination.nim#L14) and [`receiverTime`](https://github.com/status-im/nim-waku/blob/4421b8de0074574c3740b71a197b4f6eeb90c1c5/waku/v2/utils/pagination.nim#L13) fields of [`Index`](https://rfc.vac.dev/spec/13/#index)  object used for pgaination in the store protocol.

This change will touch the following parts of the codebase (non-exclusive):
- [Waku store protocol](https://github.com/status-im/nim-waku/tree/master/waku/v2/protocol/waku_store)
- [WakuMessage](https://github.com/status-im/nim-waku/blob/master/waku/v2/protocol/waku_message.nim)
- Message store interface: https://github.com/status-im/nim-waku/blob/master/waku/v2/node/storage/message/message_store.nim
- Waku message store and the `Message` table schema: https://github.com/status-im/nim-waku/blob/master/waku/v2/node/storage/message/waku_message_store.nim
- A subset of wakunode2 JSON RPC APIs: https://github.com/status-im/nim-waku/tree/master/waku/v2/node/jsonrpc

# Acceptance Criteria
- [ ] The codebase should be refactored to accommodate this change (making sure all the tests pass successfully)
- [ ] The corresponding specs i.e., [Waku2-Store](https://rfc.vac.dev/spec/13/) and [Waku2-Message](https://rfc.vac.dev/spec/14/) should reflect this change
- [ ] [Changelog](https://github.com/status-im/nim-waku/blob/master/CHANGELOG.md) should have an entry briefly explaning this update
- [ ] A migration script should be added to [migration scripts directory](https://github.com/status-im/nim-waku/tree/master/waku/v2/node/storage/migration/migrations_scripts/message) to handle the schema update in the `Message` table

PR #842, opened 2022-02-03
Refactoring timestamps

s1fr0:
We introduces a new type `Timestamp` for all timestamps in waku v2.  The type `Timestamp` is currently set to be an alias for `int64` (see [utils/time](https://github.com/status-im/nim-waku/blob/int64-timestamps/waku/v2/utils/time.nim)) and all times have now nanosecond resolution.

In particular this PR addresses the following items of [#834](https://github.com/status-im/nim-waku/issues/834):

- [x] The codebase should be refactored to accommodate this change (making sure all the tests pass successfully)
- [x] [Changelog](https://github.com/status-im/nim-waku/blob/int64-timestamps/CHANGELOG.md) should have an entry briefly explaning this update
- [x] A migration script should be added to [migration scripts directory](https://github.com/status-im/nim-waku/tree/int64-timestamps/waku/v2/node/storage/migration/migrations_scripts/message) to handle the schema update in the `Message` table


arnetheduck:
As a general recommendation, going with nanoseconds-since-epoch for timestamps is the "standard" these days, and has few downsides (except perhaps that you need to copypaste a bigger number).

jm-clius:
> As a general recommendation, going with nanoseconds-since-epoch for timestamps is the "standard" these days, and has few downsides (except perhaps that you need to copypaste a bigger number).

Right, from my side I have no strong preference i.t.o. resolution, but it may make sense to do the more standard thing. It's probably a good idea to get input from the other waku clients here as well, in case they have a reason to prefer a specific timestamp resolution. @richard-ramos @D4nte 

cc @staheri14

tests/v2/test_jsonrpc_waku.nim
```diff
@@ -102,7 +102,7 @@ procSuite "Waku v2 JSON-RPC API":
       response == true
     
     # Publish a message on the default topic
-    response = await client.post_waku_v2_relay_v1_message(defaultTopic, WakuRelayMessage(payload: @[byte 1], contentTopic: some(defaultContentTopic), timestamp: some(epochTime())))
+    response = await client.post_waku_v2_relay_v1_message(defaultTopic, WakuRelayMessage(payload: @[byte 1], contentTopic: some(defaultContentTopic), timestamp: some(int64(epochTime()))))
```
jm-clius, line 105: I know for tests we don't care as much about the accuracy of the timestamps, but `int64(epochTime())` will convert the epochTime to a timestamp with _second_ resolution by simply stripping off the fractional part.
In order to avoid having to convert by `*1000` (or `*1000000000` if we move to ns), I think it's probably best to add a utility function somewhere that takes a `float64` timestamp (seconds resolution, with sub-seconds after the decimal) and convert it to `int64` in the desired resolution. In fact, if such a util doesn't already exist in a submodule, we can provide a `getNanosecondTime(): int64` function that even reads the epoch time on behalf of the module calling the function. This way the logic resides in one place and can be used by all tests and modules.

jm-clius (commented):
Good to see this coming along!

Minor comment below about moving the utility for getting (and converting) the int64 timestamp to a utility module. This way it's unit testable and we don't need to have a lot of "magic multipliers of *1000" across the code.

staheri14:
> As a general recommendation, going with nanoseconds-since-epoch for timestamps is the "standard" these days, and has few downsides (except perhaps that you need to copypaste a bigger number).

I do also have no strong preference, the choice of millisecond was basically to make sure we can break the tie between concurrent messages (when ordering in the store protocol) and thought millisecond is good enough. I am fine with moving to nanosecond if that is standard. 

As Hanno said, would be good to also know @D4nte  and @richard-ramos thoughts.

richard-ramos:
I do like the idea of using nanoseconds and `int64`, because I'd only have to call https://pkg.go.dev/time#Time.UnixNano, instead of the way it is right now (https://github.com/status-im/go-waku/blob/master/waku/v2/utils/time.go#L7-L9) where I have to convert from nanoseconds to seconds, with the decimal part being the nanoseconds, to be able to match nim `epochTime()`

s1fr0:
Added in utils/time.nim some time utility functions to convert Unix float to ms, us, ns resolution (default nanoseconds).

D4nte:
Milliseconds is the standard in JavaScript so we'll pad right with zeroes.

waku/common/wakubridge.nim
```diff
@@ -91,7 +91,7 @@ func toWakuMessage(env: Envelope): WakuMessage =
   # Translate a Waku v1 envelope to a Waku v2 message
   WakuMessage(payload: env.data,
               contentTopic: toV2ContentTopic(env.topic),
-              timestamp: float64(env.expiry - env.ttl),
+              timestamp: int64(env.expiry - env.ttl),
```
staheri14, line 94: If `env.expiry` and `env.ttl` have a different resolution, then this needs to be converted to nanosecond

waku/v2/node/storage/migration/migrations_scripts/message/00002_addSenderTimeStamp.up.sql
```diff
@@ -26,4 +26,4 @@ INSERT INTO Message (id, receiverTimestamp, contentTopic, pubsubTopic, payload,
     SELECT id, timestamp, contentTopic, pubsubTopic, payload, version, 0
     FROM Message_backup;
 
-DROP TABLE Message_backup;
\ No newline at end of file
+DROP TABLE Message_backup;
```
staheri14, line 29: This file does not need to be changed, it is the migration script of an old version of db

waku/v2/protocol/waku_message.nim
```diff
@@ -10,6 +10,7 @@
 
 import
   libp2p/protobuf/minprotobuf
+import libp2p/varint
```
staheri14, line 13: You can combine these two imports into one like:
```
import
  libp2p/protobuf/minprotobuf
  libp2p/varint
```

waku/v2/node/storage/message/waku_message_store.nim
```diff
@@ -61,7 +61,7 @@ method put*(db: WakuMessageStore, cursor: Index, message: WakuMessage, pubsubTop
   ## 
   let prepare = db.database.prepareStmt(
     "INSERT INTO " & TABLE_TITLE & " (id, receiverTimestamp, contentTopic, payload, pubsubTopic, version, senderTimestamp) VALUES (?, ?, ?, ?, ?, ?, ?);",
-    (seq[byte], float64, seq[byte], seq[byte], seq[byte], int64, float64),
+    (seq[byte], int64, seq[byte], seq[byte], seq[byte], int64, int64),
     void
   )
```
staheri14, line 67: Lines 94 and 110 of this file also need change (The receiver timestamp and sender timestamp should be treated as int not double)

```
receiverTimestamp = sqlite3_column_double(s, 0)
```

```
senderTimestamp = sqlite3_column_double(s, 5)
```

waku/v2/protocol/waku_store/waku_store.nim
```diff
@@ -663,12 +670,12 @@ proc resume*(ws: WakuStore, peerList: Option[seq[RemotePeerInfo]] = none(seq[Rem
...
   
-  var currentTime = epochTime()
-  var lastSeenTime: float = findLastSeen(ws.messages.allItems())
+  var currentTime = getNanosecondTime(epochTime())
+  var lastSeenTime: int64 = findLastSeen(ws.messages.allItems())
   debug "resume", currentEpochTime=currentTime
   
   # adjust the time window with an offset of 20 seconds
-  let offset: float64 = 200000
+  let offset: int64 = 200000
```
staheri14, line 678: The time literals need to be converted to the nanosecond, in here `20000` is in millisecond (supposed to represent 20 seconds), without conversion, it is rendered as `20000` nanoseconds.

staheri14: BTW, `200000` has an issue, it is 200 seconds not 20 seconds, please drop one zero from it (it is not related to your PR, I just realized this is a bug)

waku/v2/utils/time.nim
```diff
@@ -0,0 +1,15 @@
+## Contains types and utilities for timestamps.
+
```
staheri14, line 2: I'd suggest defining a `Timestamp` type in this module (as an alias for int64), and replacing all the int64 types in the code with that. It makes it easier to keep track of all the timestamp values in the code and potentially makes future changes and refactoring easier.

staheri14 (commented):
Thanks for the PR, and congratulation!
I have left some comments. Once done, please let me know and I'll review again.

waku/v2/node/storage/migration/migrations_scripts/message/00002_addSenderTimeStamp.up.sql
s1fr0: Indeed, I accidentally removed the line and later restored

waku/v2/node/storage/message/waku_message_store.nim
s1fr0: In light of the new type Timestamp definition, is it worth to define a sqlite3_column_timestamp?

jm-clius (approved):
LGTM! Thanks, this goes a long way to clearing up how we use timestamps. :)

I think a good future follow-up (not for this PR) would be to also abstract the `std/times` functions, such as `epochTime()` or `getTime().toUnix()`, to our own `time` util. It could provide for example a `proc getCurrentNano*(): Timestamp` export that provides calling modules with the current epoch time as a nanosecond `Timestamp`, without having to understand or import `std/times` themselves.

waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
```diff
@@ -426,7 +426,6 @@ proc compare*(e1, e2: Epoch): int64 =
     epoch2 = fromEpoch(e2)
   return int64(epoch1) - int64(epoch2)
 
-
```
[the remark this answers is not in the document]
s1fr0: Sure, I was trying to trigger Jenkins again, since yesterday it seems it had some network issues. I would have simply re-run the tests, sorry!

jm-clius:
Two things I forgot to add:
1. Depending on a discussion in https://github.com/vacp2p/rfc/pull/483, we likely need to bump the [`protocol-id`/Codec](https://github.com/status-im/nim-waku/blob/master/waku/v2/protocol/waku_store/waku_store.nim#L46) to indicate us breaking backwards compatibility on the protocol.
2. All commits need to be signed for the PR to be mergeable (see [e.g. this doc](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits))

staheri14:
Thanks, @s1fr0 for addressing the comments, I'll review it during the day. A quick note about the PR description, it is conventional and good practice to have a PR description as well as to link the PR to the original GitHub issue (if any, which in your case is https://github.com/status-im/nim-waku/issues/834). This would allow reviewers to have a better context about the problem that the PR aims to solve :)
As for the PR description, you can get inspiration from old PRs like https://github.com/status-im/nim-waku/pull/844 and https://github.com/status-im/nim-waku/pull/835

waku/v2/node/storage/message/waku_message_store.nim
staheri14: No need for that, the underlying `sqlite3_gen` already has support for `sqlite3_column_int64`

waku/v2/utils/time.nim
```diff
@@ -0,0 +1,24 @@
...
+proc getMicrosecondTime*[T](timeInSeconds: T): Timestamp = 
+  var us = Timestamp(timeInSeconds*1000000)
+  return us
+
+proc getMillisecondTime*[T](timeInSeconds: T): Timestamp = 
+  var ms = Timestamp(timeInSeconds*1000)
+  return ms
+
+proc column_timestamp*(a1: ptr sqlite3_stmt, iCol: cint): int64 =
```
staheri14, line 23: Thanks for defining this proc, though I think it is also ok to use the `sqlite3_column_int64` directly in the code. I'd like to hear @jm-clius opinion as well

staheri14:
> 1. Depending on a discussion in [Refactoring timestamps vacp2p/rfc#483](https://github.com/vacp2p/rfc/pull/483), we likely need to bump the [`protocol-id`/Codec](https://github.com/status-im/nim-waku/blob/master/waku/v2/protocol/waku_store/waku_store.nim#L46) to indicate us breaking backwards compatibility on the protocol.

Good point, and also we may need to update the protocol id of other protocols that directly use `WakuMessage` e.g., filter and lightpush? @jm-clius

waku/v2/node/storage/message/waku_message_store.nim
s1fr0: Using `sqlite3_column_int64` might partially defeat the changes proposed in [this](https://github.com/status-im/nim-waku/pull/842#discussion_r801149390) comment. I defined a `column_timestamp` which calls `sqlite3_column_int64` in utils/time.

staheri14:
@s1fr0 The PR looks good to me! just awaiting the result of the decision about the protocol id https://github.com/status-im/nim-waku/pull/842#issuecomment-1034305222 (after that I'll approve :) ).

waku/v2/utils/time.nim
s1fr0: [Here](https://github.com/status-im/nim-waku/pull/842#discussion_r803178993) a related comment to this.

staheri14: Fair enough, I cannot see any issue with having a separate proc `column_timestamp`

oskarth:
Temporarily blocked, see https://github.com/vacp2p/rfc/pull/483#issuecomment-1042529641

D4nte:
I can't see the change of the waku store codec string from beta3 to beta4 in this PR, is this expected? @s1fr0

PR #848, opened 2022-02-11
Update websocket documentation to use domain certificates

D4nte:
See the investigation done in https://github.com/status-im/js-waku/issues/50

An operator needs to use a proper domain certificate or the user's browser will silently reject any connection to it.

### DIFF
--- a/waku/v2/README.md
+++ b/waku/v2/README.md
@@ -134,23 +134,40 @@ You can change this to `wakunode2`, the Waku v2 node like this:
 make docker-image MAKE_TARGET=wakunode2
 docker run --rm -it statusteam/nim-waku:latest --help
 ```
-## Using Websocket and secure Websockets
+## Enabling Websocket
 
-Websocket support is hidden under a feature flag and must be explicitly enabled in order to get Websockets support. The default port is 8000.
+Websocket is currently the only Waku transport supported by browser nodes that uses [js-waku](https://github.com/status-im/js-waku).
+Setting up websocket enables your node to directly serve browser peers.
 
+A valid certificate is necessary to serve browser nodes,
+you can use [`letsencrypt`](https://letsencrypt.org/):
+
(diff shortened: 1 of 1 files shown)


waku/v2/README.md
```diff
@@ -134,23 +134,40 @@ You can change this to `wakunode2`, the Waku v2 node like this:
...
 ```
-mkdir -p ../ssl_dir/
 
-openssl req -x509 -newkey rsa:4096 -keyout ../ssl_dir/key.pem -out ../ssl_dir/cert.pem -sha256 -nodes
+### Self-signed certificates
+
+Self-signed certificates are not recommended for production setups because:
+
+- Browsers do not accept self-signed certificates
+- Browser do not display an error when rejecting a certificate for websocket.
```
jm-clius, line 163:
```suggestion
- Browsers do not display an error when rejecting a certificate for websocket.
```

waku/v2/README.md
```diff
@@ -134,23 +134,40 @@ You can change this to `wakunode2`, the Waku v2 node like this:
 make docker-image MAKE_TARGET=wakunode2
 docker run --rm -it statusteam/nim-waku:latest --help
 ```
-## Using Websocket and secure Websockets
+## Enabling Websocket
 
-Websocket support is hidden under a feature flag and must be explicitly enabled in order to get Websockets support. The default port is 8000.
+Websocket is currently the only Waku transport supported by browser nodes that uses [js-waku](https://github.com/status-im/js-waku).
+Settings up websocket enables your node to directly serve browser peers.
```
jm-clius, line 140:
```suggestion
Setting up websocket enables your node to directly serve browser peers.
```

jm-clius (approved):
Thanks for this, @D4nte. Feel free to merge this, as it's certainly better than what we've had.

PR #852, opened 2022-02-17
Domain name config for `nwaku` node

jm-clius:
This PR introduces the ability to configure a `nwaku` node with an (ipv4) domain name.

This allows for the node's publically announced `multiaddrs` to use the `/dns4/` scheme. In addition, nodes with domain name and secure websocket configured, will generate a discoverable ENR containing the `wss` `multiaddr` with `/dns4/` domain name. This is necessary to verify domain certificates when connecting to this node over secure websocket. 

Related issues: https://github.com/status-im/nim-waku/pull/848 https://github.com/status-im/nim-waku/issues/845

### DIFF
--- a/tests/v2/test_wakunode.nim
+++ b/tests/v2/test_wakunode.nim
@@ -1481,3 +1481,22 @@ procSuite "WakuNode":
       node.switch.peerInfo.addrs.contains(announcedEndpoint)
 
     await node.stop()
+
+  asyncTest "Node can use dns4 in announced addresses":
+    let
+      nodeKey = crypto.PrivateKey.random(Secp256k1, rng[])[]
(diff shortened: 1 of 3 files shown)


oskarth (approved):
> This allows for the node's publically announced multiaddrs to use the /dns4/ scheme. In addition, nodes with domain name and secure websocket configured, will generate a discoverable ENR containing the wss multiaddr with /dns4/ domain name. This is necessary to verify domain certificates when connecting to this node over secure websocket.

I wonder if we should add a paragraph or so about this in docs?

jm-clius:
> I wonder if we should add a paragraph or so about this in docs?

Agree! Done in https://github.com/status-im/nim-waku/pull/856

PR #856, opened 2022-02-18
Add docs on configuring domain and DNS discovery

jm-clius:
Adds documentation:
- explaining how to configure a `dns4` domain name for a `nwaku` node (see https://github.com/status-im/nim-waku/pull/852)
- linking the DNS discovery instructions in the separate tutorial to the main `README`
- listing the current available URLs for connecting to one of the existing fleets using DNS discovery

### DIFF
--- a/waku/v2/README.md
+++ b/waku/v2/README.md
@@ -134,6 +134,48 @@ You can change this to `wakunode2`, the Waku v2 node like this:
 make docker-image MAKE_TARGET=wakunode2
 docker run --rm -it statusteam/nim-waku:latest --help
 ```
+
+## Configuring a domain name
+
+It is possible to configure an IPv4 DNS domain name that resolves to the node's public IPv4 address.
+
+```shell
+wakunode2 --dns4-domain-name=mynode.example.com
+```
+
+This allows for the node's publically announced `multiaddrs` to use the `/dns4` scheme.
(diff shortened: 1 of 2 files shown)


waku/v2/README.md
```diff
@@ -134,6 +134,48 @@ You can change this to `wakunode2`, the Waku v2 node like this:
...
+--dns-discovery-name-server  DNS name server IPs to query. Argument may be repeated.
+```
+
+- `--dns-discovery` is used to enable DNS discovery on the node.
+Waku DNS discovery is disabled by default.
+- `--dns-discovery-url` is mandatory if DNS discovery is enabled.
+It contains the URL for the node list.
+The URL must be in the format `enrtree://<key>@<fqdn>` where `<fqdn>` is the fully qualified domain name and `<key>` is the base32 encoding of the compressed 32-byte public key that signed the list at that location.
+- `--dns-discovery-name-server` is optional and contains the IP(s) of the DNS name servers to query.
+If left unspecified, the Cloudflare servers `1.1.1.1` and `1.0.0.1` will be used by default.
```
D4nte, line 168: Any reason why it does not use the system's default DNS resolution?

D4nte (approved):

oskarth (approved):

PR #858, opened 2022-02-21
Remove ref to fleets.status.im

jm-clius:
Closes https://github.com/status-im/nim-waku/issues/850

Removes the remaining reference to _fleets.status.im_ from the `nim-waku` codebase. This was used by `chat2` to select random nodes to connect to from a configured fleet. In the new implementation, `chat2` uses DNS discovery to discover and connect to fleet nodes (or any list of nodes published to a user-configurable `--dns-discovery-url`).

From a user's perspective the behaviour is the same as before: by default a `chat2` client will attempt to connect to `prod`, although this is configurable using the `--fleet` option and can be disabled using `--fleet:none`.

### DIFF
--- a/examples/v2/chat2.nim
+++ b/examples/v2/chat2.nim
@@ -154,7 +154,6 @@ proc getChatLine(c: Chat, msg:WakuMessage): Result[string, string]=
                 else: string.fromBytes(msg.payload)
     return ok(chatline)
 
-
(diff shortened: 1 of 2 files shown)


jakubgs (approved):
Nice.

oskarth (approved):

PR #861, opened 2022-02-22
Moved PB Timestamp index to 10

s1fr0:
This is a fix to maintain retro-compatibility with `timestamp` protobuffer fields.

Timestamps now have `Timestamp` type (currently an alias for `int64`) and nanoseconds resolution. Their new protobuffer index is `10`.

Previous versions had timestamps as `float` with subseconds contained in the fractional part and protobuffer index `4`.

This PR is related to https://github.com/vacp2p/rfc/pull/491, https://github.com/status-im/nim-waku/pull/842 and https://github.com/vacp2p/rfc/pull/483.

### DIFF
--- a/waku/v2/protocol/waku_message.nim
+++ b/waku/v2/protocol/waku_message.nim
@@ -46,7 +46,7 @@ proc init*(T: type WakuMessage, buffer: seq[byte]): ProtoResult[T] =
   discard ? pb.getField(3, msg.version)
 
   var timestamp: zint64
-  discard ? pb.getField(4, timestamp)
+  discard ? pb.getField(10, timestamp)
   msg.timestamp = Timestamp(timestamp)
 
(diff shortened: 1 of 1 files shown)


jm-clius:
Thanks! This will buy some time for other clients to implement the new `WakuMessage` proto without breaking.

jm-clius (approved):
Thanks!

staheri14 (approved):
Looks good to me, what triggered this change? what was the issue with index 4?

Update: found the answer in 

> Thanks! This will buy some time for other clients to implement the new `WakuMessage` proto without breaking.

Are we going to bring it back to index 4 once the timestamp type and resolution are updated by other clients?

D4nte (approved):

oskarth (approved):

Issue #837, opened 2022-01-28
Discovery v5: Separate Waku Network via selectable `protocol-id`


kaiserd:
## Background

This issue is part of the [Waku v2 discv5 Roadmap](https://forum.vac.dev/t/waku-v2-discv5-roadmap-discussion/121).

As a first step for establishing a separate Waku discv5 network, #770 proposed filtering nodes by ENR using the [waku-specific ENR field](https://rfc.vac.dev/spec/31/#waku2-enr-key)
However, we believe that this will not be enough to sufficiently separate the Waku discv5 network, because ENR information is not necessarily available for all messages and thus might lead to leakage into the main discv5 network (or other discv5 networks).
A separate `protocol-id` allows validating each incoming and outgoing message.

## Proposed Solution

This issue proposes to generalize [nim-eth/discv5](https://github.com/status-im/nim-eth/blob/master/doc/discv5.md) allowing to select a `protocol-id` different from the standard `discv5`.
With this generalization, it seems reasonable to make the `protocol-version` selectable as well.

### steps
- [ ] Implement selectable `protocol-id` and `protocol-version` as a feature-branch in a fork of nim-eth.
  - `protocol-id` and `protocol-version` as members of [Protocol object](https://github.com/status-im/nim-eth/blob/73d9bf4c80812a889f240a9ba39ef260cc940e39/eth/p2p/discoveryv5/protocol.nim#L119-L137)
- [ ] Use `"d5waku"`as `protocol-id` for nim-waku discv5 in a feature-branch and adjust the implementation accordingly
- [ ] adapt/extend testcases
- [ ] test Waku discv5 in a local test setup

As this is an experimental feature for now, this issue does not track updating the documentation.

After sufficient testing and the approval of the nim-eth developers, we plan to merge the selectable `protocol-id` feature into nim-eth.
Larger scale tests of the benefits of having a separate Waku discv5 network will be tracked by future issues.

### alternative implementation
Having `protocol-id` and `protocol-version` as members of the protocol object makes it more "dynamic" but requires quite a few code changes. The `protocol-id` and `protocol-version` are expected to be fixed per application implementation, so a more static implementation with global variables might be preferable.
@kdeme wdyt?

cc @jm-clius @oskarth @kdeme

kdeme:
> The `protocol-id` and `protocol-version` are expected to be fixed per application implementation, so a more static implementation with global variables might be preferable.

[...]

PR #862, opened 2022-02-22
discv5 stage1 beta test

kaiserd:
This PR serves as a basis for a first beta test of Waku ambient discovery via discv5. It is part of the first stage of the [Waku discv5 roadmap](https://forum.vac.dev/t/waku-v2-discv5-roadmap-discussion) and addresses #837.

##  Info 
This PR does not use the master branch of [nim-eth/discv5](https://github.com/status-im/nim-eth/tree/master/eth/p2p/discoveryv5) but uses this [experimental feature branch](https://github.com/kaiserd/nim-eth/tree/add-selectable-protocol-id-static) which implements a static selectable `protocol-id`. The selectable `protocol-id` is used to span a separate Waku discv5 network (see [Waku RFC 33 draft PR](https://github.com/vacp2p/rfc/pull/487)).

## Edit / Updates
- back in draft mode for now, see my last comment


waku/v2/node/discv5/waku_discv5.nim
```diff
@@ -77,7 +77,8 @@ proc findRandomPeers*(wakuDiscv5: WakuDiscoveryV5): Future[Result[seq[RemotePeer
   let discoveredNodes = await wakuDiscv5.protocol.queryRandom()
   
   ## Filter based on our needs
-  let filteredNodes = discoveredNodes.filter(isWakuNode) # Currently only a single predicate
+  # let filteredNodes = discoveredNodes.filter(isWakuNode) # Currently only a single predicate
+  let filteredNodes = discoveredNodes # we do not filter based on ENR in the first waku discv5 beta stage
```
jm-clius, line 81: I'd rather not make the copy here. In other words, just use `discoveredNodes`, until some time in future we decide to filter again. For now `filteredNodes` can be removed.

waku/v2/node/wakunode2.nim
```diff
@@ -1168,9 +1172,13 @@ when isMainModule:
...
     ## Resume history, connect to static nodes and start
     ## keep-alive, if configured.
-
-   # Start Waku v2 node
+    
+    # Start Waku v2 node
     waitFor node.start()
+
+    # start discv5 and connect to discovered nodes
+    if not waitFor node.startDiscv5():
```
jm-clius, line 1180: This will attempt to start discv5, even if it's disabled. Even though this will fail quickly, it does trigger code which is unexpected. Perhaps only call this if `conf.discv5Discovery`?

waku/v2/node/config.nim
```diff
@@ -265,6 +265,24 @@ type
...
+      desc: "Maximum amount of nodes with the same IP in discv5 routing table buckets"
+      defaultValue: 2
+      name: "discv5-bucket-ip-limit" .}: uint
+
+    discv5BitsPerHop* {.
+      hidden
+      desc: "Kademlia's b variable, increase for less hops per lookup"
+      defaultValue: 1
+      name: "discv5-bits-per-hop" .}: int
+
```
jm-clius, line 285: I'm not convinced that these need to be exposed as user-configurable items, at least for now with our very basic requirements. Unless we can define a clear use case where "normal" operators would need to fine tune this, it may be prudent to set these defaults inside the code for now. As we encounter more and more exotic use cases we can expose config items as necessary. Wdyt?

jm-clius (changes requested):
Looks good and very close to mergeable. One or two minor comments, and a question about what configuration we need to expose.

waku/v2/node/config.nim
jm-clius: Update: Reviewing https://github.com/status-im/nim-waku/pull/863 I notice that at least some of these items are useful when setting up local tests, which I think is a good enough reason for their existence.

waku/v2/node/discv5/waku_discv5.nim
kaiserd: Done. I left a comment about filtering.

waku/v2/node/wakunode2.nim
kaiserd: done. 😅

kaiserd:
When waku discv5 discovers a new peers, it generates the peer info based on the ENR used by discv5 [see here](
https://github.com/status-im/nim-waku/blob/c27086e1865ce42ba1b318ceef634ace6b0f07f1/waku/v2/node/discv5/waku_discv5.nim#L86).
This peer info is then given to the waku relay / libp2p dialer, which in turn tries to dial the discv5 multiaddress instead of the waku relay multiaddress.

Because of this issue, I will switch this PR to draft mode for now.

jm-clius (approved):
This lgtm, thanks! I would imagine that, since discv5 is over udp, that the libp2p dialer would skip over these as udp is not a supported transport?

kaiserd:
> I would imagine that, since discv5 is over udp, that the libp2p dialer would skip over these as udp is not a supported transport?

Yes. But it still continuously tries to connect:

```
Failed to connect to peer                  topics="wakunode" tid=4160 file=wakunode2.nim:779 wireAddr=/ip4/0.0.0.0/udp/9003 peerId=16U*X3XYsB
```
(The actually used IP addr in this test is a static public IP address.)
I will dig deeper into this.

Imho, the UDP address in the ENR should not be handed to the Waku relay protocol at all because it does not correspond to a Waku relay endpoint, but to a discv5 endpoint. Once we have a final solution, I will document the following in the RFC.

* the TCP based multiaddress extracted from an ENR correpsonds to a Waku relay endpoint
* the UDP based multiaddress extracted from an ENR correpsonds to a Waku discv5 endpoint

@jm-clius thank you for the off-issue `--keep-alive=true` hint. I was not aware of this, with this the experiment runs stable so far. I will let it run overnight one more time.

kaiserd:
Update on the issue:

The ENRs distributed by discv5 did not contain a TCP port, which is necessary to build the Waku relay multiaddress.
The reason was the port was not passed to the constructor when disv5 object for wakunode2, if an external address was manually specified (which I did for the experiment).

My last commit here fixes this issue. The test runs stable now. I re-open this issue for review after some more testing.

waku/v2/node/wakunode2.nim
```diff
@@ -1076,18 +1076,22 @@ when isMainModule:
...
     if conf.discv5Discovery:
-      let discv5UdpPort = Port(uint16(conf.discv5UdpPort) + conf.portsShift)
+      let
+        discv5UdpPort = Port(uint16(conf.discv5UdpPort) + conf.portsShift)
+        discoveryConfig = DiscoveryConfig.init(
+          conf.discv5TableIpLimit, conf.discv5BucketIpLimit, conf.discv5BitsPerHop)
 
       node.wakuDiscv5 = WakuDiscoveryV5.new(
-        extIP, extTcpPort, some(discv5UdpPort),
+        extIP, extPort, some(discv5UdpPort),
```
jm-clius, line 1085: Thanks for catching this! Would have been a nasty bug to find later.

jm-clius (approved):
Thanks, Daniel. LGTM. I think this can be merged and included in the release.

kaiserd:
Thank you!
A first interoperabilty test we just ran was successful, too.

kaiserd:
The interoperability test was successful :).

.gitmodules
```diff
@@ -1,8 +1,8 @@
 [submodule "vendor/nim-eth"]
 	path = vendor/nim-eth
-	url = https://github.com/status-im/nim-eth.git
+	url = https://github.com/kaiserd/nim-eth.git
```
oskarth, line 3: Can we get this to point to the branch `add-selectable-protocol-id-static` on status-im nim-eth instead? Easier to keep up to date with permissions etc as opposed to at your individual account. You should have permissions to push branches

Is there an issue for tracking this upstream?

waku/v2/node/config.nim
oskarth: At some point we might want to factor out admin/dev options vs things users are likely to want to specify. For example with a "namespace"

oskarth (approved):
Nice!

jm-clius:
I suggest we merge this so we can start smoke testing the bulk of the changes that will form part of the release. We can switch the submodule branch in a subsequent PR. wdyt @kaiserd ?

PR #863, opened 2022-02-23
discv5 stage1 testnet

kaiserd:
This draft PR adds discv5 to `start_network2` which serves as a local test for #862.

## Running the test

```
make sim2
./build/start_network2 --topology=bootstrapFleet
```

## Info

`bootstrapFleet` is a new topology type which contains a full-mesh core of `fleet nodes` and a set of `full nodes` using this fleet as discv5 bootstrap nodes.
These `full nodes` retrieve a random set of nodes via discv5 and then use these as waku peers.
The number of fleet nodes can be set via `--numFleetNodes` (default = 3), the number of `full nodes` can be set via `--amount` (default = 16).

The local IP address limit of the discv5 routing table is set to 1024 and the IP address limit for table buckets is set to 16 to allow larger scale local tests. (PR #862 introduces these IP limits as config options.)


waku/v2/node/start_network2.nim
```diff
@@ -9,10 +12,11 @@ import
 import strutils except fromHex
 
 const
-  defaults ="--log-level:TRACE --metrics-logging --metrics-server --rpc"
+  defaults ="--log-level:TRACE --metrics-logging --metrics-server --rpc --discv5-discovery --discv5-table-ip-limit=1024 --discv5-bucket-ip-limit=16 --discv5-bits-per-hop=1"
```
jm-clius, line 15: Does this mean the nodes spawned by `start_network2` will now always support and run `discv5`?

jm-clius (commented):
Interesting to see this coming along! Since this is only a `Draft`, perhaps you could merge to your other feature branch in https://github.com/status-im/nim-waku/pull/862 so the diff introduced in this branch is clearer?

waku/v2/node/start_network2.nim
kaiserd: For this experiment yes. I did not try to make the testnet more general so far.
Should I add a setup with just a subset of the nodes supporting discv5, or  should I generalize the whole setup?
I was planning to add `mobile nodes`  later once we agree on a frist synchronization protocol for `mobile nodes`.

oskarth (commented):
Nice! What is left for this to be ready-for-review and mergeable? Would you mind adding the minimal TODO (s)that's left in PR description?

kaiserd:
> What is left for this to be ready-for-review and mergeable? 

Atm, I did not plan to merge this PR and keep it as experimental. I focus on the interoperability test for now.

> Would you mind adding the minimal TODO (s)that's left in PR description?

I added a todo item that had to be address if we want to merge it :); adding more items if needed.

waku.nimble
```diff
@@ -65,7 +65,7 @@ task test1, "Build & run Waku v1 tests":
 
 ### Waku v2 tasks
 task wakunode2, "Build Waku v2 (experimental) cli node":
-  buildBinary "wakunode2", "waku/v2/node/", "-d:chronicles_log_level=DEBUG"
+  buildBinary "wakunode2", "waku/v2/node/", "-d:chronicles_log_level=TRACE"
```
oskarth, line 68: This is fine for local checkout but for merging can remove this

waku/v2/node/start_network2.nim
```diff
@@ -153,12 +212,15 @@ when isMainModule:
 
   # Scenario xx2 14
   let amount = conf.amount
+  let numFleetNodes = conf.numFleetNodes
```
oskarth, line 215: If this is a new scenario I'd just copy paste existing code and add it there instead of overwriting

>  Generalize the setup to allow for various disovery methods, especially static nodes. In its current state, this PR simply overwrites the former experiment, using discv5 instead of static nodes.

Shouldn't need to generalize too much, just add another code section for discv5 scenario. Let's KISS

oskarth (commented):
I think we should merge it so it can be used by others. Don't need any general setup here, it is just simulation code and easiest is to just add a scenario and KISS

waku.nimble
kaiserd: done

waku/v2/node/start_network2.nim
kaiserd: Done :).

kaiserd:
Thank you. I addressed the feedback.
Waiting for one more review on  #862 before making this PR mergeable, because this PR built on that.

waku/v2/node/start_network2.nim
```diff
@@ -68,19 +90,68 @@ proc initNodeCmd(shift: int, staticNodes: seq[string] = @[], master = false, lab
...
+    some(Port(discv5Port)), # tcp-port
+    some(Port(discv5Port))) # udp-port
+    .expect("Record within size limits")
+  result.enrUri = "enr:" & enr.toBase64
+
   info "Node command created.", cmd=result.cmd, address = result.address
 
+
+# # Scenarios without discv5
+# proc starNetwork(amount: int): seq[NodeInfo] =
```
oskarth, line 121: Is there a reason to comment this out? As opposed to defining a new function with other name like, `discv5Network` and have it be listed as a topology here https://github.com/status-im/nim-waku/blob/7168bf6606938fff97435251238aece802606793/waku/v2/node/start_network2.nim#L232-L238

I don't quite understand `bootstrapFleetNetwork` vs the changes in `starNetwork`. How come they aren't a single topology discv5Something?

kaiserd: These topologies are ment to be independent of the scenarios.
`Star` consists of a single `master node` which all other nodes use as bootstrap node.
`BootstrapFleet` consists of a fully connected fleet (whose size can be set via a parameter). All nodes get the set of fleet nodes as bootstrap nodes. `BootstrapFleet` can be seen as a generalization of `Star`.

I used `BootstrapFleet` as the topology for the discv5 tests to mimic the real world fleet setup.

> Is there a reason to comment this out? 

There is one default `Star` setup and one for discv5.
Because `BootstrapFleet` is new, and I just focused on the discv5 setup, I did not add a default version for it.
I could add one, if desired.

kaiserd: (resolved in 1:1 meeting)

jm-clius (approved):
LGTM!

PR #864, opened 2022-02-23
Store data model updates and fixes

jm-clius:
This PR updates the store data model (both in-memory and in the SQLite schema).
It is the last necessary step to close https://github.com/status-im/nim-waku/issues/838 and fix https://github.com/status-im/nim-waku/issues/841.
This required a minor change in the RFC, tracked in https://github.com/vacp2p/rfc/pull/492.

## Data model

The data model defines an `index` for each message that defines:

### 1. The minimal vector to _uniquely_ identify a message.
This is necessary e.g. for deduplication, finding a message based on a `cursor` (an `index` provided in `HistoryQuery`), defining a suitable PRIMARY KEY in the database, etc:

A message is uniquely identified by the combination of `senderTimestamp`, `digest` (over both the message `payload` and `contentTopic`) and `pubsubTopic`. A `receiverTimestamp` is not part of the unique identifier of a message, but can be used as fallback for sorting order when `senderTimestamp` is unset.

### 2. The default sorting order for messages.
This determines the default order in which pages of history is retrieved and returned.
The sorting order, from most to least significant, is based on:
1. senderTimestamp
2. receiverTimestamp (only when senderTimestamp is unused)
3. message digest
4. pubsubTopic

## Summary of code changes

- use `receiverTimestamp` as a fallback for sort order, if the `senderTimestamp` is unset/unused. The `receiverTimestamp` should not be considered for message equality, which results in a bit of code nastiness but seems to be the best we can do for now.
- add `pubsubTopic` to `Index`. This defaults to the default pubsub topic.
- fix `PRIMARY KEY` constraint on SQLite `Message` table
- some miscellaneous migration fixes: (cc @s1fr0 here)
  - there's a script for `user_version` `3`, but the target version was never set in `migration_types.nim`, which means it wouldn't have run
  - SQLite does not support `FLOOR` to convert `REAL` types to `INTEGER` afaict, but `CAST TO INTEGER` essentially does the same (i.e. truncates the fractional part)

### DIFF
--- a/waku/v2/node/storage/message/waku_message_store.nim
+++ b/waku/v2/node/storage/message/waku_message_store.nim
@@ -38,7 +38,7 @@ proc init*(T: type WakuMessageStore, db: SqliteDatabase): MessageStoreResult[T] 
         payload BLOB,
         version INTEGER NOT NULL,
         senderTimestamp """ & TIMESTAMP_TABLE_TYPE & """  NOT NULL,
-        CONSTRAINT messageIndex PRIMARY KEY (senderTimestamp, receiverTimestamp, id, pubsubTopic)
+        CONSTRAINT messageIndex PRIMARY KEY (senderTimestamp, id, pubsubTopic)
     ) WITHOUT ROWID;
     """, NoParams, void)
 
--- a/waku/v2/node/storage/message/waku_message_store.nim
+++ b/waku/v2/node/storage/message/waku_message_store.nim
@@ -31,13 +31,14 @@ proc init*(T: type WakuMessageStore, db: SqliteDatabase): MessageStoreResult[T] 
 
   let prepare = db.prepareStmt("""
     CREATE TABLE IF NOT EXISTS """ & TABLE_TITLE & """ (
-        id BLOB PRIMARY KEY,
+        id BLOB,
         receiverTimestamp """ & TIMESTAMP_TABLE_TYPE & """ NOT NULL,
         contentTopic BLOB NOT NULL,
         pubsubTopic BLOB NOT NULL,
         payload BLOB,
         version INTEGER NOT NULL,
-        senderTimestamp """ & TIMESTAMP_TABLE_TYPE & """  NOT NULL
+        senderTimestamp """ & TIMESTAMP_TABLE_TYPE & """  NOT NULL,
+        CONSTRAINT messageIndex PRIMARY KEY (senderTimestamp, receiverTimestamp, id, pubsubTopic)
     ) WITHOUT ROWID;
     """, NoParams, void)
 
@@ -133,5 +134,3 @@ method getAll*(db: WakuMessageStore, onData: message_store.DataProc, limit = non
 proc close*(db: WakuMessageStore) = 
   ## Closes the database.
   db.database.close()
-
-
--- a/waku/v2/node/storage/migration/migration_types.nim
+++ b/waku/v2/node/storage/migration/migration_types.nim
@@ -7,7 +7,7 @@ const MESSAGE_STORE_MIGRATION_PATH* = sourceDir / "migrations_scripts/message"
 const PEER_STORE_MIGRATION_PATH* = sourceDir / "migrations_scripts/peer"
 const ALL_STORE_MIGRATION_PATH* = sourceDir / "migrations_scripts"
 
-const USER_VERSION* = 2 # increase this when there is an update in the database schema
+const USER_VERSION* = 4 # increase this when there is an update in the database schema
 
 type MigrationScriptsResult*[T] = Result[T, string]
 type
--- a/waku/v2/utils/pagination.nim
+++ b/waku/v2/utils/pagination.nim
@@ -14,26 +14,52 @@ export hash
 type
   Index* = object
     ## This type contains the  description of an Index used in the pagination of WakuMessages
-    digest*: MDigest[256]
+    digest*: MDigest[256] # calculated over payload and content topic
     receiverTime*: Timestamp
     senderTime*: Timestamp # the time at which the message is generated
+    pubsubTopic*: string
 
 proc `==`*(x, y: Index): bool =
-  ## receiverTime plays no role in index comparison
-  (x.senderTime == y.senderTime) and (x.digest == y.digest)
+  ## receiverTime plays no role in index equality
+  (x.senderTime == y.senderTime) and
+  (x.digest == y.digest) and
+  (x.pubsubTopic == y.pubsubTopic)
 
 proc cmp*(x, y: Index): int =
   ## compares x and y
   ## returns 0 if they are equal 
   ## returns -1 if x < y
   ## returns 1 if x > y
-  ## receiverTime plays no role in index comparison
+  ## 
+  ## Default sorting order priority is:
+  ## 1. senderTimestamp
+  ## 2. receiverTimestamp (a fallback only if senderTimestamp unset and all other fields unequal)
+  ## 3. message digest
+  ## 4. pubsubTopic
   
   # Timestamp has a higher priority for comparison
   let timecmp = cmp(x.senderTime, y.senderTime)
   if timecmp != 0: 
     return timecmp
 
-  # Only when timestamps are equal 
-  let digestcm = cmp(x.digest.data, y.digest.data)
-  return digestcm
+  # Continue only when sender timestamps are equal 
+  let
+    digestcmp = cmp(x.digest.data, y.digest.data)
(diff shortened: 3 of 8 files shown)


waku/v2/utils/pagination.nim
```diff
@@ -14,26 +14,52 @@ export hash
...
 
-  # Only when timestamps are equal 
-  let digestcm = cmp(x.digest.data, y.digest.data)
-  return digestcm
+  # Continue only when sender timestamps are equal 
+  let
+    digestcmp = cmp(x.digest.data, y.digest.data)
+    pstopiccmp = cmp(x.pubsubTopic, y.pubsubTopic)
+
+  if x.senderTime == 0 and
```
staheri14, line 50: Just a suggestion, it could have been easier to do an equality check at the beginning of the `cmp` proc  using `==` with an immediate return (if x and y are equal). Then the rest of the logic would become a bit simpler.

waku/v2/utils/pagination.nim
```diff
@@ -14,26 +14,52 @@ export hash
...
+     (digestcmp != 0 or pstopiccmp != 0):
+    
+    ## We sort on receiver time only if:
+    ## - senderTime is unset (== 0)
+    ## - message digests and pubsubTopics aren't equal (i.e. receiver time should not affect index equality, which
+    ##   is based only on sender time, pubsubTopic and digest)
+    let receiverTimeCmp = cmp(x.receiverTime, y.receiverTime)
```
staheri14, line 57: what if `y` has a `senderTimestamp`? do we totally ignore the `senderTimestamp` if not present in both x and y?

waku/v2/utils/pagination.nim
staheri14, line 50: What will happen if `y.senderTime` is zero?

staheri14 (commented):
LGTM!
I have a few questions about the fallback mechanism, please see my comments.

waku/v2/node/storage/message/waku_message_store.nim
```diff
@@ -31,13 +31,14 @@ proc init*(T: type WakuMessageStore, db: SqliteDatabase): MessageStoreResult[T]
...
-        id BLOB PRIMARY KEY,
+        id BLOB,
         receiverTimestamp """ & TIMESTAMP_TABLE_TYPE & """ NOT NULL,
         contentTopic BLOB NOT NULL,
         pubsubTopic BLOB NOT NULL,
         payload BLOB,
         version INTEGER NOT NULL,
-        senderTimestamp """ & TIMESTAMP_TABLE_TYPE & """  NOT NULL
+        senderTimestamp """ & TIMESTAMP_TABLE_TYPE & """  NOT NULL,
+        CONSTRAINT messageIndex PRIMARY KEY (senderTimestamp, receiverTimestamp, id, pubsubTopic)
```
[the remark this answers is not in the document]
jm-clius: Right, very good point. The primary key indeed acts as a data integrity constraint that would cause DB operations to fail with an exception when violated. This is expensive and should be avoided. In other words, we should deduplicate messages _before_ attempting to store them in the DB. This is what I've now done with [this return](https://github.com/status-im/nim-waku/pull/864/commits/d29d0e44eb3aca5dbfabde343ee8d4e7ac4a1481#diff-c574072f2cd8c3e7a6fbc74ba4aa947f6296e0fe18d455053f45b23447afcdb7R399). Of course, then there's also no need for the DB scheme to be permissive of duplicates, so I've also removed the `receiverTimestamp` from the PRIMARY KEY in [ed09cce](https://github.com/status-im/nim-waku/pull/864/commits/ed09cce6b890ba5ff13486d1f85df0b7c36db7e6). Thanks for catching this. 👍

waku/v2/utils/pagination.nim
jm-clius: 😆 You're right! Much simpler this way. Done in [d29d0e4](https://github.com/status-im/nim-waku/pull/864/commits/d29d0e44eb3aca5dbfabde343ee8d4e7ac4a1481)

waku/v2/utils/pagination.nim
jm-clius: Another good point and oversight from my side. According to me the behaviour should be to fall back to `receiverTimestamp` comparison on both x and y if `senderTimestamp` is unset on _either_  x _or_ y. I've added this in [d29d0e4](https://github.com/status-im/nim-waku/pull/864/commits/d29d0e44eb3aca5dbfabde343ee8d4e7ac4a1481) and also included unit tests for this case.

waku/v2/utils/pagination.nim
jm-clius: See comment above. It should now fall back to `receiverTime` even if only one of x or y has unset `senderTime`.

waku/v2/utils/pagination.nim
```diff
@@ -33,33 +33,28 @@ proc cmp*(x, y: Index): int =
...
+  if x == y:
+    # Quick exit ensures receiver time does not affect index equality
+    return 0
+  
   # Timestamp has a higher priority for comparison
-  let timecmp = cmp(x.senderTime, y.senderTime)
+  var timecmp: int
+  if (x.senderTime == 0 or y.senderTime == 0):
+    # if either side does not have a senderTime set, we fall back to receiverTime comparison
+    timecmp = cmp(x.receiverTime, y.receiverTime)
```
staheri14, line 48: Re https://github.com/status-im/nim-waku/pull/864#discussion_r813805368
I'd say let's only fall back to `receivertime` for messages that don't have `senderTime`. I mean if x has sender time but y does not have, then compare x.senderTime with y.receiverTime. In the current implementation, a message with a valid `senderTime` sometimes gets compared with its `senderTime` and sometimes with its `receiverTime` which causes a non-deterministic behavior in the sorting output.

waku/v2/node/storage/migration/migration_types.nim
```diff
@@ -7,7 +7,7 @@ const MESSAGE_STORE_MIGRATION_PATH* = sourceDir / "migrations_scripts/message"
 const PEER_STORE_MIGRATION_PATH* = sourceDir / "migrations_scripts/peer"
 const ALL_STORE_MIGRATION_PATH* = sourceDir / "migrations_scripts"
 
-const USER_VERSION* = 2 # increase this when there is an update in the database schema
+const USER_VERSION* = 4 # increase this when there is an update in the database schema
```
oskarth, line 10: What is USER_VERISON 3?

oskarth: Nevermind answered in PR description

waku/v2/utils/pagination.nim
```diff
@@ -14,26 +14,47 @@ export hash
...
 proc cmp*(x, y: Index): int =
   ## compares x and y
   ## returns 0 if they are equal 
   ## returns -1 if x < y
   ## returns 1 if x > y
-  ## receiverTime plays no role in index comparison
+  ## 
+  ## Default sorting order priority is:
+  ## 1. senderTimestamp
+  ## 2. receiverTimestamp (a fallback only if senderTimestamp unset on either side, and all other fields unequal)
```
oskarth, line 36: What's the rationale for not breaking ties on senderTimestamp with receiverTimestamp? I.e. if you receive two messages in same second. Seems like the special casing makes it more complicated, but maybe I'm missing something here?

jm-clius: The same message on different store nodes will have slightly differing `receiverTimestamp`s. If a store node tries to synchronise/resume from multiple other store nodes, it won't be able to deduplicate any messages if it takes `receiverTimestamp` into consideration. Something similar could happen during normal `relay` if different traversal paths for the same message has enough latency diff to fall outside the underlying gossipsub deduplication buffer - i.e. two identical messages with the exact same `senderTimestamp` should be considered duplicates. Of course, this would mean that `store` duplicate detection on messages without `senderTimestamp` may not be 100% accurate, but I think this is our best option.

waku/v2/utils/pagination.nim
jm-clius: Right, I had it like this and abandoned this approach because I thought comparing `receiverTime` only with `receiverTime` would be somewhat more accurate. However, you're right, I can now think of ways in which this may lead to (serious) inconsistencies in the underlying data structure. Have changed to suggested approach in [edfb68b](https://github.com/status-im/nim-waku/pull/864/commits/edfb68bdcde8bb2385210e0876158b2b7c1c6466) and modified unit tests to cover this case.

staheri14 (approved):
Thanks Hanno  👍

PR #865, opened 2022-02-24
fix: message when connecting to fleet nodes

kaiserd:
This PR fixes the message echoed when connecting to a fleet.
It used to say "No static peers configured [...]".
However, the corresponding branch will be executed regardless if static peers are configured or not.

This PR also adds an echo printing a message when connecting to static nodes, because a similiar messages is printed when connecting to nodes learned via other methods.

### DIFF
--- a/examples/v2/chat2.nim
+++ b/examples/v2/chat2.nim
@@ -378,13 +378,14 @@ proc processInput(rfd: AsyncFD, rng: ref BrHmacDrbgContext) {.async.} =
                   symKey: generateSymKey(conf.contentTopic))
 
   if conf.staticnodes.len > 0:
+    echo "Connecting to static peers..."
     await connectToNodes(chat, conf.staticnodes)
   
   var dnsDiscoveryUrl = none(string)
 
   if conf.fleet != Fleet.none:
     # Use DNS discovery to connect to selected fleet
-    echo "No static peers configured. Connecting to " & $conf.fleet & " fleet using DNS discovery..."
+    echo "Connecting to " & $conf.fleet & " fleet using DNS discovery..."
     
     if conf.fleet == Fleet.test:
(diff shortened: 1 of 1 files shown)


examples/v2/chat2.nim
```diff
@@ -378,13 +378,14 @@ proc processInput(rfd: AsyncFD, rng: ref BrHmacDrbgContext) {.async.} =
                   symKey: generateSymKey(conf.contentTopic))
 
   if conf.staticnodes.len > 0:
+    echo "Connecting to static peers..."
     await connectToNodes(chat, conf.staticnodes)
   
   var dnsDiscoveryUrl = none(string)
 
   if conf.fleet != Fleet.none:
```
staheri14, line 386: Shouldn't it be inside an `elif` to split the two execution branches?
```suggestion
  elif conf.fleet != Fleet.none:
```

staheri14 (commented):
Looks good! just left a comment.

examples/v2/chat2.nim
jm-clius: No strong opinion here, but I think it should be valid to have both static nodes and fleet option configured.

jm-clius (approved):
LGTM, thanks!

examples/v2/chat2.nim
kaiserd: @staheri14 I thought the same, but did not change the control flow to keep my change minimal. Imho, too, this behaviour would be more intuitive.
@jm-clius I agree, but it might also be counter-intuitive to connect to the fleet automatically when specifying static nodes.
For me, it would be more natural to have fleet=node as a default if I specify static node.
Also, no strong opinion on my side here.

kaiserd: On the other hand, It might be better to leave the control flow as is as some scripts might assume this logic...

jm-clius:
Going ahead and merging this. Any changes in logic can be addressed in future PRs. This is simply a log message change.

PR #878, opened 2022-03-03
Fixes getNanoSeconds() time conversion 

staheri14:
A fix to the conversion of seconds to nanoseconds in the `getNanoSeconds()` proc. 
The faulty version would cause messages to be timestamped (sender or receiver) with some value pointing to the past.

### DIFF
--- a/waku/v2/utils/time.nim
+++ b/waku/v2/utils/time.nim
@@ -9,11 +9,11 @@ type Timestamp* = int64
 const TIMESTAMP_TABLE_TYPE* = "INTEGER"
 
 proc getNanosecondTime*[T](timeInSeconds: T): Timestamp = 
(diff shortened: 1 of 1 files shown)


s1fr0 (approved):
Much better! Thanks!

kaiserd (approved):
LGTM!

oskarth (approved):

jm-clius (approved):
Thanks! This is a very important fix. It will cause issues with the sorting order of `store` messages. 
Afaict it affects:
-nim-waku chat2 client (new messages will be dropped by `store`s, as the message timestamps are in the past)
-store receiver timestamps (an issue if the clients don't provide sender timestamps)
-bridged messages from v1 to v2 (arguably the most critical)

I propose we fast-track this merge please and get this deployed as a patch on all fleets.

cc @oskarth

Issue #880, opened 2022-03-04
Discv5: DNS bootstrap


kaiserd:
## Background

[DNS discovery](https://rfc.vac.dev/spec/10/#discovery-methods) is currently used for retrieving Waku2 relay bootstrap nodes.
The only way of providing discv5 bootstrap nodes, however, is passing their [ENRs](https://eips.ethereum.org/EIPS/eip-778) as command line arguments.
This makes setting up a discv5-capable Waku node cumbersome.

## Proposed solution

Use nodes retrieved via DNS discovery not only as bootstrap nodes for Waku2 relay but also as bootstrap nodes for discv5.

## Acceptance criteria

* [x] implementation of DNS bootstrap nodes for Waku discv5 (PR: #888)
* [x] successful test in experimental fleet

PR #888, opened 2022-03-10
discv5 dns bootstrap

kaiserd:
## Background

This PR addresses #880.
The current draft implementation passed a test on an experimental fleet.
This fleet server can be used for testing:
`enrtree://AKRUVB7WRORHGL5IIHKHEXGALD5IWWAL7UIVOUZ2FKE4ORV5WOX76@waku.337.ovh`
(I will keep this running during the testing phase; might be down sometimes during update phases.)

The following explains my implementation choices.
(No strong opinion on these, and happy to change if desired :))

### Introducing a separate setup phase for bootstrap nodes

Bootstrap nodes are passed to the discv5 object during the `initialize node` phase while they are dialed by Waku relay in the `start node` phase.
To avoid retrieving bootstrap nodes twice, having a separate phase for retrieving them before `initialize node` seemed like a solution with not too much restructuring.
Also, the new `retrieve dynamic bootstrap nodes` phase can be extended to retrieve dynamic bootstrap nodes from other sources, too (e.g. mDNS).

### ENR in RemotePeerInfo

This PR extends `RemotePeerInfo` storing ENRs of discovered nodes.
[DNS discovery](https://rfc.vac.dev/spec/10/#discovery-methods) retrieves ENRs anyway. While storing the ENR was not useful before because Waku only uses multiaddrs, it is now useful with the introduction of discv5, which uses the ENR format.
Storing the ENRs also avoids additional conversions.

(I am not sure if `Option` is the idiomatic Nim way for introducing an optional field. I did not find information on this in the [style guide](https://status-im.github.io/nim-style-guide/00_introduction.html). Options feel a bit clunky in Nim.)

## Todo

If the current implementation choices are fine, I would address the following

* [x] remove double conversion of ENRs from string to Record format and vice versa
* ~extend local peerInfo with ENR~ `peerInfo` is in libp2p
   - a future PR could generate an ENR [here](https://github.com/status-im/nim-waku/blob/b8da352a56b49642f198f30c45946669e055c0c1/waku/v2/utils/peers.nim#L160)

### DIFF
--- a/waku/v2/node/wakunode2.nim
+++ b/waku/v2/node/wakunode2.nim
@@ -1263,12 +1263,12 @@ when isMainModule:
   else:
     (pStorage, mStorage) = setupStorageRes.get()
 
-  debug "2/7 Retrieve dynamic boostrap nodes"
+  debug "2/7 Retrieve dynamic bootstrap nodes"
   
   var dynamicBootstrapNodes: seq[RemotePeerInfo]
   let dynamicBootstrapNodesRes = retrieveBootstrapNodes(conf)
   if dynamicBootstrapNodesRes.isErr:
-    error "2/7 Retrieving dynamic boostrap nodes failed. Continuing without dynamic boostrap nodes."
+    error "2/7 Retrieving dynamic bootstrap nodes failed. Continuing without dynamic bootstrap nodes."
   else:
     dynamicBootstrapNodes = dynamicBootstrapNodesRes.get()
 
--- a/waku/v2/node/wakunode2.nim
+++ b/waku/v2/node/wakunode2.nim
@@ -1014,7 +1014,7 @@ when isMainModule:
     ok(storeTuple)
 
   # 2/7 Retrieve dynamic bootstrap nodes
-  proc retrieveBootstrapNodes(conf: WakuNodeConf): Result[seq[RemotePeerInfo], cstring] =
+  proc retrieveDynamicBootstrapNodes(conf: WakuNodeConf): Result[seq[RemotePeerInfo], cstring] =
   # DNS discovery
     if conf.dnsDiscovery and conf.dnsDiscoveryUrl != "":
       debug "Discovering nodes using Waku DNS discovery", url=conf.dnsDiscoveryUrl
@@ -1099,16 +1099,22 @@ when isMainModule:
         discv5UdpPort = Port(uint16(conf.discv5UdpPort) + conf.portsShift)
         discoveryConfig = DiscoveryConfig.init(
           conf.discv5TableIpLimit, conf.discv5BucketIpLimit, conf.discv5BitsPerHop)
-      var discv5BootstrapNodes = conf.discv5BootstrapNodes
+
+      # select dynamic bootstrap nodes that have an ENR containing a udp port
+      var discv5BootstrapEnrs: seq[enr.Record]
       for n in dynamicBootstrapNodes:
-        if n.enr.isSome():
-          discv5BootstrapNodes.add("enr:" & n.enr.get().toBase64)
+        if n.enr.isSome() and n.enr.get().toTypedRecord().tryGet().udp.isSome(): # TODO: idiomatic error handling
+          discv5BootstrapEnrs.add(n.enr.get())
+    
+      # parse enrURIs from the configuration and add the resulting ENRs to the discv5BootstrapEnrs seq
+      for enrUri in conf.discv5BootstrapNodes:
+        addBootstrapNode(enrUri, discv5BootstrapEnrs)
 
       node.wakuDiscv5 = WakuDiscoveryV5.new(
         extIP, extPort, some(discv5UdpPort),
         conf.listenAddress,
         discv5UdpPort,
-        discv5BootstrapNodes,
+        discv5BootstrapEnrs,
         conf.discv5EnrAutoUpdate,
         keys.PrivateKey(conf.nodekey.skkey),
         wakuFlags,
@@ -1266,7 +1272,7 @@ when isMainModule:
   debug "2/7 Retrieve dynamic bootstrap nodes"
   
   var dynamicBootstrapNodes: seq[RemotePeerInfo]
-  let dynamicBootstrapNodesRes = retrieveBootstrapNodes(conf)
+  let dynamicBootstrapNodesRes = retrieveDynamicBootstrapNodes(conf)
   if dynamicBootstrapNodesRes.isErr:
     error "2/7 Retrieving dynamic bootstrap nodes failed. Continuing without dynamic bootstrap nodes."
   else:
--- a/waku/v2/node/wakunode2.nim
+++ b/waku/v2/node/wakunode2.nim
@@ -1014,7 +1014,7 @@ when isMainModule:
     ok(storeTuple)
 
   # 2/7 Retrieve dynamic bootstrap nodes
-  proc retrieveDynamicBootstrapNodes(conf: WakuNodeConf): Result[seq[RemotePeerInfo], cstring] =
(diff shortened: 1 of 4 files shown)


waku/v2/node/wakunode2.nim
```diff
@@ -1257,44 +1265,53 @@ when isMainModule:
   let setupStorageRes = setupStorage(conf)
 
   if setupStorageRes.isErr:
-    error "1/6 Setting up storage failed. Continuing without storage."
+    error "1/7 Setting up storage failed. Continuing without storage."
   else:
     (pStorage, mStorage) = setupStorageRes.get()
 
-  debug "2/6 Initializing node"
+  debug "2/7 Retrieve dynamic boostrap nodes"
```
jm-clius, line 1272:
```suggestion
  debug "2/7 Retrieve dynamic bootstrap nodes"
```
:)

waku/v2/node/wakunode2.nim
```diff
@@ -1257,44 +1265,53 @@ when isMainModule:
...
+  
+  var dynamicBootstrapNodes: seq[RemotePeerInfo]
+  let dynamicBootstrapNodesRes = retrieveBootstrapNodes(conf)
+  if dynamicBootstrapNodesRes.isErr:
+    error "2/7 Retrieving dynamic boostrap nodes failed. Continuing without dynamic boostrap nodes."
```
jm-clius, line 1277:
```suggestion
    error "2/7 Retrieving dynamic bootstrap nodes failed. Continuing without dynamic boostrap nodes."
```
:)

jm-clius:
This approach makes sense to me! Feel free to mark as ready for review once it is. :)

kaiserd:
I now opened the issue for review after addressing the double ENR conversion.
The tests on my experimental fleet still run successfully.

I added a second constructor for WakuDiscoveryV5, which takes `seq[enr.Record]` instead of a `seq[string]` (EnrUri).
This saves conversations. (The old constructor is not used anymore; but it might still make sense to keep it for potential future use.)

Imo, generating an ENR in [`toRemotePeerInfo`](https://github.com/status-im/nim-waku/blob/b8da352a56b49642f198f30c45946669e055c0c1/waku/v2/utils/peers.nim#L160) could be addressed in a separate Issue/PR. Let me know, if I should do this in this PR. Currently, this generated ENR would not be used.


[Here](https://github.com/status-im/nim-waku/blob/c838b487ab02359119991dfcee83c8a8a783738a/waku/v2/node/wakunode2.nim#L1104), I tried to do something like

```nim
var discv5BootstrapEnrs = dynamicBootstrapNodes
                            .filterIt(it.enr.isSome() and
                              it.enr.get().toTypedRecord().tryGet().udp.isSome()) # TODO: idiomatic error handling. 
                            .mapIt(it.enr.get())
```

However, filter and map are not composable without copying, and I could not find a way of addressing the potential error on tryGet() inline.
Does Nim provide something like result combinators in Rust?

I also found collect, which would allow doing something along the lines of:

```nim
let dynamicBootstrapEnrs = collect(newSeq):
                            for n in dynamicBootstrapNodes:
                              if n.enr.isSome():
                                n.enr.get()
```

However, this requires `std/sugar`, uses a macro, and is not much more concise compared to a regular for loop (in this example, I I did not do the udp port check, which is why it looks more concise.)

kaiserd:
Because the `WakuDiscoveryV5` constructores only differ in the type of the ENR seq, passing an empty seq without type does not work anymore. The type has to be specified like  [here](https://github.com/status-im/nim-waku/blob/f15990d93149ea90d1f397f08332c8e3495ddf60/tests/v2/test_waku_discv5.nim#L55).

waku/v2/node/wakunode2.nim
```diff
@@ -1013,9 +1013,34 @@ when isMainModule:
     
     ok(storeTuple)
 
-  # 2/6 Initialize node
+  # 2/7 Retrieve dynamic bootstrap nodes
+  proc retrieveDynamicBootstrapNodes(conf: WakuNodeConf): Result[seq[RemotePeerInfo], cstring] =
```
oskarth, line 1017: Is there a reason we are using `cstring` here as opposed to hiding this in a lower level API? Seems like an implementation detail that'd be better to abstract away to either a string or some other relevant type. (I realize this might be a consequence of lower level API)

oskarth, line 1017: Does this block? One concern would be if the node gets busy discovering nodes and isn't able to start functioning earlier

waku/v2/node/wakunode2.nim
```diff
@@ -1075,11 +1100,25 @@ when isMainModule:
         discoveryConfig = DiscoveryConfig.init(
           conf.discv5TableIpLimit, conf.discv5BucketIpLimit, conf.discv5BitsPerHop)
 
+      # select dynamic bootstrap nodes that have an ENR containing a udp port
```
oskarth, line 1103: Maybe brief comment mentioning why this is necessary, i.e. Discovery v5 is only UDP, or link to spec

waku/v2/utils/peers.nim
```diff
@@ -145,11 +150,12 @@ proc toRemotePeerInfo*(enr: enr.Record): Result[RemotePeerInfo, cstring] =
...
 
-  return ok(RemotePeerInfo.init(peerId, addrs))
+  return ok(RemotePeerInfo.init(peerId, addrs, some(enr)))
 
 ## Converts the local peerInfo to dialable RemotePeerInfo
 ## Useful for testing or internal connections
 proc toRemotePeerInfo*(peerInfo: PeerInfo): RemotePeerInfo =
   RemotePeerInfo.init(peerInfo.peerId,
                       peerInfo.addrs,
+                      none(enr.Record), # TODO
```
oskarth, line 160: What's the TODO referring to here?

oskarth (commented):
Nice one! Some minor comments

waku/v2/node/wakunode2.nim
kaiserd: There is no reason. I converted the Result to `SetupResult`, which all other setup methods use as well. This is much clearer.  Thank you :).

waku/v2/utils/peers.nim
kaiserd: I replaced the TODO with a comment.

waku/v2/node/wakunode2.nim
kaiserd: Yes it blocks. Agreed.
The blocking stems form [waku_dnsdisc](https://github.com/status-im/nim-waku/blob/e6fdbdb42ac8592c02811d4a2709d4d17e065d0c/waku/v2/node/dnsdisc/waku_dnsdisc.nim#L52).
Imo, we should address this in a separate issue.

For now, this should not be too severe, because it is "just" blocking DNS queries. These DNS records should be available in the next setup step and would be waited for anyways.
I think the DNS disc API should be async, but the setup functions should be sync because the steps sequentially depend on each other.

Should I open a new issue or take further steps within this PR?

kaiserd:
I addressed the feedback and also [allowed a `udp6` port in the ENR](https://github.com/status-im/nim-waku/blob/305735b250802810a05b758d8a521ab04a952e37/waku/v2/node/wakunode2.nim#L1112), because the [ENR spec](https://eips.ethereum.org/EIPS/eip-778) does not exclude ENRs that only have `udp6` but no `udp`.

waku/v2/node/wakunode2.nim
oskarth: Agree and opening a new issue sounds good, thanks!

oskarth (approved):

jm-clius (approved):
LGTM! Thanks.

Issue #866, opened 2022-02-25
Build error - ubuntu


D4nte:
Trying to build nim-waku master on ubuntu machine.

```
$ sudo apt install g++ make
$ make
Building: Nim compiler
Building: install/usr/lib/libbacktrace.a
Building: install/usr/lib/libbacktracenim.a
Building: build/wakunode1
[NimScript] exec: nim c --out:build/wakunode1 -d:chronicles_log_level=DEBUG -d:chronosStrictException --verbosity:0 --hints:off -d:release waku/v1/node/wakunode1.nim
Building: build/sim1
[NimScript] exec: nim c --out:build/quicksim -d:chronicles_log_level=INFO -d:chronosStrictException --verbosity:0 --hints:off -d:release waku/v1/node/quicksim.nim
[NimScript] exec: nim c --out:build/start_network -d:chronicles_log_level=DEBUG -d:chronosStrictException --verbosity:0 --hints:off -d:release waku/v1/node/start_network.nim
Building: build/example1
[NimScript] exec: nim c --out:build/example -d:chronicles_log_level=DEBUG -d:chronosStrictException --verbosity:0 --hints:off -d:release examples/v1/example.nim
Building: build/wakunode2
[NimScript] exec: nim c --out:build/wakunode2 -d:chronicles_log_level=DEBUG --verbosity:0 --hints:off -d:release waku/v2/node/wakunode2.nim
/home/fryorcraken/src/nim-waku/waku/v2/protocol/waku_relay.nim(60, 8) Warning: use {.base.} for base methods; baseless methods are deprecated [UseBase]
/home/fryorcraken/src/nim-waku/waku/v2/protocol/waku_relay.nim(75, 8) Warning: use {.base.} for base methods; baseless methods are deprecated [UseBase]
/home/fryorcraken/src/nim-waku/waku/v2/protocol/waku_relay.nim(81, 8) Warning: use {.base.} for base methods; baseless methods are deprecated [UseBase]
/home/fryorcraken/src/nim-waku/waku/v2/protocol/waku_store/waku_store.nim(10, 6) Warning: imported and not used: 'algorithm' [UnusedImport]
/home/fryorcraken/src/nim-waku/waku/v2/utils/wakuswitch.nim(6, 7) Warning: imported and not used: 'byteutils' [UnusedImport]
[...]

PR #889, opened 2022-03-14
add RAM requirement to README

kaiserd:
This PR addresses #866.
I also added a short explanation on how to run wakunod2 on the smaller DO droplet.
Not sure if this is too specific for the main README, but might help folks running into the same issue.

### DIFF
--- a/waku/v2/README.md
+++ b/waku/v2/README.md
@@ -29,6 +29,15 @@ make wakunode2
 #./build/wakunode2 --log-level:debug --discovery:off --fleet:test --log-metrics
 ```
 
+Note: building `wakunode2` requires 2GB of RAM. The build will fail on systems not fulfilling this requirement.
+
+Setting up a `wakunode2` on the smallest [digital ocean](https://docs.digitalocean.com/products/droplets/how-to/) droplet, you can either
+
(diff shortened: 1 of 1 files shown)


waku/v2/README.md
```diff
@@ -29,6 +29,14 @@ make wakunode2
 #./build/wakunode2 --log-level:debug --discovery:off --fleet:test --log-metrics
 ```
 
+Note: building `wakunode2` requires 2GB of RAM. The build will fail on systems not fulfilling this requirement.
+
+Setting up a `wakunode2` on the smallest [digital ocean](https://docs.digitalocean.com/products/droplets/how-to/) droplet, you can either
+
+* compile on a stronger droplet featuring the same CPU architecture and downgrade after compiling, or
+* activate swap on the smallest droplet.
```
D4nte, line 37: Or use docker :)

D4nte (approved):

waku/v2/README.md
kaiserd: added :).

jm-clius (approved):
Thanks! This will also be important when eventually compiling a more comprehensive guide.

PR #893, opened 2022-03-15
Updates the type of the `proof` field of wakumessage to `RateLimitProof`

staheri14:
This PR is to update the type of the proof field of the waku message to rate limit proof. It was previously disabled under the rln  compiler flag due to which spam-protected messages would not be routed by relay nodes with no support for rln. This will aid the rln live testnet where the test fleets can merely act as a router for the spam-protected messages needless to compile the rln rust lib and other utility functions.

### DIFF
--- a/waku/v2/protocol/waku_message.nim
+++ b/waku/v2/protocol/waku_message.nim
@@ -11,9 +11,8 @@
 import
   libp2p/protobuf/minprotobuf,
   libp2p/varint,
-  ../utils/time
-when defined(rln):
-  import waku_rln_relay/waku_rln_relay_types
+  ../utils/time,
+  waku_rln_relay/waku_rln_relay_types
 
 const
   MaxWakuMessageSize* = 1024 * 1024 # In bytes. Corresponds to PubSub default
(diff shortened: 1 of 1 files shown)


oskarth:
> where the test fleets can merely act as a router for the spam-protected messages needless to compile the rln rust lib and other utility functions.

I don't understand why the RLN library wouldn't have to be compiled for testnet? Does that mean test fleet wouldn't validate messages? Is there an underlying GH issue for this?

oskarth (approved):

jm-clius (approved):

staheri14:
> > where the test fleets can merely act as a router for the spam-protected messages needless to compile the rln rust lib and other utility functions.
> 
> I don't understand why the RLN library wouldn't have to be compiled for testnet? Does that mean test fleet wouldn't validate messages? Is there an underlying GH issue for this?

The end goal is to enable test fleets to run rln hence filtering spam-protected messages. 
However, as we discussed in our last 1:1,  since all chat users connect to the fleets, the network will have a star topology hence if we activate rln on the fleets then spam protection occurs at the fleets side and we cannot demonstrate spam-detection at the chat2 users' side.

Due to this: 1) I'll start using fleets just as routers so that spam detection takes place at the chat2 clients' side. 2) The next step is to activate rln on test fleets for chat2 spam-protected content topics in which spam-protection occurs centrally on the fleet nodes.

This will also help in debugging potential issues as changes are deployed to the test fleets incrementally.

oskarth:
Oh my bad, now I remember the conversation and rationale :)

PR #897, opened 2022-03-16
fix: prevent empty store response due to blocked walker

jm-clius:
Fixes https://github.com/status-im/nim-waku/issues/882

Previously a store node could run into a state where it would return an empty history response to each store query. This was because the walker used to populate a page of history would not get properly destroyed under certain error conditions - specifically when the node received a query with an invalid cursor. Afterwards all history responses would be empty.

Fix has been soak tested on the `prod` fleet where this issue was most observed previously.


oskarth (approved):
I'm generally in favor of pedestrians, but I guess you gotta break a few eggs to make an omelette

staheri14 (approved):

PR #900, opened 2022-03-20
fix(store): capacity limit for persistent message store

kaiserd:
This PR provides a simple solution to https://github.com/status-im/infra-status/issues/5 and https://github.com/status-im/nim-waku/issues/767

Currently, the DB grows boundless, and, while only a subset of messages is loaded into the in-memory store, an `ORDER-BY` operation is run on the whole DB.
This might lead to very long loading times https://github.com/status-im/infra-status/issues/5 

This PR adds `numMessages` (a message counter) and `maxMessages` (max number of messages to be held in the sqlite DB) to the `WakuMessageStore`.
When `put`ting a message into the store, it checks if `maxMessages` is reached and if so, deletes a configurable percentage of oldest messages.

This PR introduces two new config options:

* `--db-capacity` the maximum number of messages stored in the DB
* `--db-truncate-percentage` the percentage of messages to be deleted once the max is reached

These are added for flexibility when testing on the test fleet and might be removed in the future.

This PR further changes the first part of the sqlite primary key from `senderTimestamp` to `receiverTimestamp`.
Rationale: Most costly sort operations are on `receiverTimestamp`, so this change increases performance.
I would suggest, building a dedicated index on `receiverTimestamp` as even with this change, sqlite has to build a temporary B-tree when ordering by `receiverTimestamp`.
(For this, I could introduce another issue. As we plan to update to a more efficient DB structure anyway, we might wait for the respective issue.)

## Alternatives

These are alternatives to solving the issues mentioned above:

* don't change the implementation but simply run a chron job truncating the DB from time to time
* trigger the `trunctate` proc introduced by this PR within a timer instead of on `put`.

We are planning to update the DB structure to make it more efficient, which will solve these issues, too.
So this PR is more of a quick fix.

cc @jakubgs 

## Todo

These todo's should be completed before making the PR mergeable / "ready for review".

* [x] fix test error : reverted the primary key change for now
* [x] add tests
* ~adapt the migration test to the new primary key~ no primary key update in this PR
* [ ] update documentation

### DIFF
--- a/waku/v2/node/storage/message/waku_message_store.nim
+++ b/waku/v2/node/storage/message/waku_message_store.nim
@@ -4,6 +4,7 @@ import
   std/[options, tables],
   sqlite3_abi,
   stew/[byteutils, results],
+  chronicles,
   ./message_store,
   ../sqlite,
   ../../../protocol/waku_message,
@@ -12,18 +13,59 @@ import
 
 export sqlite
 
+logScope:
+  topics = "wakuMessageStore"
+
 const TABLE_TITLE = "Message"
+const MaxStoreOverflow = 1.3 # has to be > 1.0
 
 # The code in this file is an adaptation of the Sqlite KV Store found in nim-eth.
 # https://github.com/status-im/nim-eth/blob/master/eth/db/kvstore_sqlite3.nim
-#
-# Most of it is a direct copy, the only unique functions being `get` and `put`.
 
 type
+  # WakuMessageStore implements auto deletion as follows:
+  # The sqlite DB will store up to `storeMaxLoad = storeCapacity` * `MaxStoreOverflow` messages, giving an overflow window of (storeCapacity*MaxStoreOverflow - storeCapacity).
+  # In case of an overflow, messages are sorted by `receiverTimestamp` and the oldest ones are deleted. The number of messages that get deleted is (overflow window / 2) = deleteWindow,
+  # bringing the total number of stored messages back to `storeCapacity + (overflow window / 2)`. The rationale for batch deleting is efficiency.
+  # We keep half of the overflow window in addition to `storeCapacity` because we delete the oldest messages with respect to `receiverTimestamp` instead of `senderTimestamp`.
+  # `ReceiverTimestamp` is guaranteed to be set, while senders could omit setting `senderTimestamp`.
+  # However, `receiverTimestamp` can differ from node to node for the same message.
+  # So sorting by `receiverTimestamp` might (slightly) prioritize some actually older messages and we compensate that by keeping half of the overflow window.
   WakuMessageStore* = ref object of MessageStore
     database*: SqliteDatabase
+    numMessages: int
+    storeCapacity: int # represents both the number of messages that are persisted in the sqlite DB (excl. the overflow window explained above), and the number of messages that get loaded via `getAll`.
+    storeMaxLoad: int  # = storeCapacity * MaxStoreOverflow
+    deleteWindow: int  # = (storeCapacity * MaxStoreOverflow - storeCapacity)/2; half of the overflow window, the amount of messages deleted when overflow occurs
  
-proc init*(T: type WakuMessageStore, db: SqliteDatabase): MessageStoreResult[T] =
+proc messageCount(db: SqliteDatabase): MessageStoreResult[int64] =
+  var numMessages: int64
+  proc handler(s: ptr sqlite3_stmt) = 
+    numMessages = sqlite3_column_int64(s, 0)
+  let countQuery = "SELECT COUNT(*) FROM " & TABLE_TITLE
+  let countRes = db.query(countQuery, handler)
+  if countRes.isErr:
+    return err("failed to count number of messages in DB")
+  ok(numMessages)
+
+proc deleteOldest(db: WakuMessageStore): MessageStoreResult[void] =
+  var deleteQuery = "DELETE FROM " & TABLE_TITLE & " " &
+                          "WHERE id NOT IN " &
+                              "(SELECT id FROM " & TABLE_TITLE & " " &
+                                "ORDER BY receiverTimestamp DESC " &
+                                "LIMIT " & $(db.storeCapacity + db.deleteWindow) & ")"
+  let res = db.database.query(deleteQuery, proc(s: ptr sqlite3_stmt) = discard)
+  if res.isErr:
+    return err(res.error)
+  db.numMessages = db.storeCapacity + db.deleteWindow # sqlite3 DELETE does not return the number of deleted rows; Ideally we would subtract the number of actually deleted messages. We could run a separate COUNT.
+
+  when defined(debug):
+    let numMessages = messageCount(db.database).get() # requires another SELECT query, so only run in debug mode
+    debug "Oldest messages deleted from DB due to overflow.", storeCapacity=db.storeCapacity, maxStore=db.storeMaxLoad, deleteWindow=db.deleteWindow, messagesLeft=numMessages
+
+  ok()
+
+proc init*(T: type WakuMessageStore, db: SqliteDatabase, storeCapacity: int = 50000): MessageStoreResult[T] =
   ## Table is the SQL query for creating the messages Table.
   ## It contains:
   ##  - 4-Byte ContentTopic stored as an Integer
@@ -45,11 +87,28 @@ proc init*(T: type WakuMessageStore, db: SqliteDatabase): MessageStoreResult[T] 
   if prepare.isErr:
     return err("failed to prepare")
 
-  let res = prepare.value.exec(())
-  if res.isErr:
+  let prepareRes = prepare.value.exec(())
+  if prepareRes.isErr:
     return err("failed to exec")
 
-  ok(WakuMessageStore(database: db))
+  let numMessages = messageCount(db).get()
+  debug "number of messages in sqlite database", messageNum=numMessages
+
+  let storeMaxLoad = int(float(storeCapacity) * MaxStoreOverflow)
+  let deleteWindow = int(float(storeMaxLoad - storeCapacity) / 2)
+  let wms = WakuMessageStore(database: db,
+                      numMessages: int(numMessages),
+                      storeCapacity: storeCapacity,
+                      storeMaxLoad: storeMaxLoad,
+                      deleteWindow: deleteWindow)
+
+  # If the loaded db is already over max load, delete the oldest messages before returning the WakuMessageStore object
+  if wms.numMessages >= wms.storeMaxLoad:
+    let res = wms.deleteOldest()
+    if res.isErr: return err("deleting oldest messages failed")
+
+  ok(wms)
+
 
 method put*(db: WakuMessageStore, cursor: Index, message: WakuMessage, pubsubTopic: string): MessageStoreResult[void] =
   ## Adds a message to the storage.
@@ -74,11 +133,17 @@ method put*(db: WakuMessageStore, cursor: Index, message: WakuMessage, pubsubTop
   if res.isErr:
     return err("failed")
 
+  db.numMessages += 1
+  if db.numMessages >= db.storeMaxLoad:
+    let res = db.deleteOldest()
+    if res.isErr: return err("deleting oldest failed")
+
   ok()
 
-method getAll*(db: WakuMessageStore, onData: message_store.DataProc, limit = none(int)): MessageStoreResult[bool] =
-  ## Retrieves all messages from the storage.
-  ## Optionally limits the number of rows returned.
+
+
+method getAll*(db: WakuMessageStore, onData: message_store.DataProc): MessageStoreResult[bool] =
+  ## Retrieves `storeCapacity` many  messages from the storage.
   ##
   ## **Example:**
   ##
@@ -120,10 +185,10 @@ method getAll*(db: WakuMessageStore, onData: message_store.DataProc, limit = non
   var selectQuery = "SELECT receiverTimestamp, contentTopic, payload, pubsubTopic, version, senderTimestamp " &
                     "FROM " & TABLE_TITLE & " " & 
                     "ORDER BY receiverTimestamp ASC"
-  if limit.isSome():
-    # Optional limit applies. This works because SQLITE will perform the time-based ORDER BY before applying the limit.
-    selectQuery &= " LIMIT " & $(limit.get()) &
-                   " OFFSET cast((SELECT count(*)  FROM " & TABLE_TITLE & ") AS INT) - " & $(limit.get()) # offset = total_row_count - limit
+
+  # Apply limit. This works because SQLITE will perform the time-based ORDER BY before applying the limit.
+  selectQuery &= " LIMIT " & $db.storeCapacity &
+                 " OFFSET cast((SELECT count(*)  FROM " & TABLE_TITLE & ") AS INT) - " & $db.storeCapacity # offset = total_row_count - limit
   
   let res = db.database.query(selectQuery, msg)
   if res.isErr:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 - Waku v1 <> v2 bridge now supports DNS `multiaddrs`
 - Waku v1 <> v2 bridge now validates content topics before attempting to bridge a message from Waku v2 to Waku v1
+- Message store now auto deletes messages once over specified `--store-capacity`. This can significantly improve node start-up times.
 
 ### Fixes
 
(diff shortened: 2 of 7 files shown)


jakubgs (commented):
Wouldn't setting `storeCapacity` to a value lower than `dbCapacity` invalidate the first setting?
If so, should that at least trigger some warning if not an error?

Seems like both of those parameters are for the same thing, but work differently, that could be quite confusing. Also, doesn't this approach mean that at times we'd have less messages than `dbCapacity`accounts for?

kaiserd:
Thank you for the feedback.

> Wouldn't setting `storeCapacity` to a value lower than `dbCapacity` invalidate the first setting?
> If so, should that at least trigger some warning if not an error?

I could add such an error. As an alternative, the two new options could also be removed and automatically set to `storeCapacity * 2` and `20` for example.

> Seems like both of those parameters are for the same thing, but work differently, that could be quite confusing.

One is for in-memory, the other for persistence. I agree, this is confusing with the current naming.  Most likely, it would be better to just have one option controlling both. With these paramaters, I wanted to give more flexibility in a testing phase.

>  Also, doesn't this approach mean that at times we'd have less messages than `dbCapacity` accounts for?

Yes. I think, deleting in batches makes more sense because of efficiency.
But yes, this adds to the confusion you pointed out.

My suggestion would be not exposing `dbCapacty` and `dbTruncatePercentage`. This would solve the confusion issue you pointed out. Maybe it would make sense, to do a few tests checking which values we should choose relateive to `storeCapacity`. Also, we could add a sentence in the doc explaining that `storeCapacity` just refers to in-memory storage.
(Will discuss this in today's meeting.)

kaiserd:
I revertet the primary key for now, because it made problems with the current migration test.

kaiserd:
> If we want query cache then it should be done using something dedicated, like memcached, or similar.

Regarding this, I was thinking about using something like Redis because most of the operations happen on the in-memory store and the DB is just for persisting, which Redis can do, too, on regular intervals.
Also, afaik, the focus for the store is more on efficiency than transactional safety and it should not be a big deal if a few messages get lost on a few nodes, because others would still likely hold the respective messages. I will soon have a 1:1 with @jm-clius planning our store efficiency issue.

jakubgs:
>This PR is just a quick-fix for the very long loading times.

Of course, just trying to clear some things up.

>Regarding this, I was thinking about using something like Redis because most of the operations happen on the in-memory store and the DB is just for persisting, which Redis can do, too, on regular intervals.

Yeah, agree, Redis is a nice choice, I quite like it as a cache.

jakubgs:
@oskarth @s1fr0 @staheri14 @jm-clius can we please get some reviews going?

This issue is causing nodes to go down at every deployment and config change for an hour or more now.

waku/v2/node/storage/message/waku_message_store.nim
```diff
@@ -45,11 +52,38 @@ proc init*(T: type WakuMessageStore, db: SqliteDatabase): MessageStoreResult[T]
...
+    return err("failed to count number of messages in DB")
+
+  debug "number of messages in sqlite database", messageNum=numMessages
+
+  ok(WakuMessageStore(database: db,
+                      numMessages: int(numMessages),
+                      maxNumMessages: 2000,
+                      truncateWindowSize: int(maxNumMessages * truncateWindowPercentage/100)))
+
+proc truncate(db: WakuMessageStore): MessageStoreResult[void] =
```
jm-clius, line 75: A bit of a nitpick, but doesn't `truncate` in SQL terms usually mean deleting _all_ rows from a table (not just the oldest)? In other words, if we want to add such a function in future we may run into clashing terminology. Perhaps calling this something closer to `deleteOldest` or `resize` could avoid confusion in future.

jm-clius (approved):
Thanks, @kaiserd. I think this is a mergeable minimal fix for the DB capacity issue! As you mentioned in the conversation above, we can work on improving our DB solution and efficiency in subsequent PRs.

oskarth:
Please mark PR as "ready for review" once it is ready for proper review

waku/v2/node/storage/message/waku_message_store.nim
jm-clius: Ah, yes! That's on me. :)

kaiserd:
I updated the PR addressing @jakubgs' feedback.
Further, I removed the `limit` parameter from the `getAll` method.
The only configurable limit is now `storeCapacity` which can be set via `--store-capacity`.
Also, auto-delete now happens on node start, too, when loading a DB that exceeds the limit already.

Auto-delete now works like this (I added this explanation as a comment to the `WakuMessageStore` type):

The sqlite DB will store up to `storeCapacity` * `maxStoreOverflow` messages. `maxStoreOverflow` is a `const` set to `1.2`.
If more messages are stored, the messages are sorted by `receiverTimestamp` and the oldest ones are deleted,
bringing the total number of stored messages back to `storeCapacity*maxStoreOverflow/2`. The rationale for batch deleting is efficiency.
We keep half of the overflow window because we delete the oldest messages with respect to `receiverTimestamp` instead of `senderTimestamp`. `ReceiverTimestamp` is guaranteed to be set, while senders could omit setting `senderTimestamp`. However, `receiverTimestamp` can differ from node to node for the same message.
So, sorting by `receiverTimestamp` might (slightly) prioritize some actually older messages and we compensate that by keeping half of the overflow window.

(I got delayed because I ran into a nasty bug. I forgot to adapt the signature of `getAll` in `message_store` which was hard to find due to Nim's dynamic dispatch.)

I also commented out a test for the old limit parameter.

This PR will become mergeable / "ready for review" after adding/adapting tests and adding docu which I will do asap.

I will open a new issue for
* introducing an index on `receiverTimestamp`, and
* using [sqlite3 vaccum](https://www.sqlite.org/lang_vacuum.html)

for increasing efficiency.

waku/v2/node/storage/message/waku_message_store.nim
```diff
@@ -12,22 +13,49 @@ import
...
   WakuMessageStore* = ref object of MessageStore
     database*: SqliteDatabase
+    numMessages: int
+    storeCapacity: int # represents both the number of messages that are persisted in the sqlite DB (excl. the overflow window explained above), and the number of messages that get loaded via `getAll`.
+    storeMaxLoad: int  # = --store-capacity * maxStoreOverflow
  
-proc init*(T: type WakuMessageStore, db: SqliteDatabase): MessageStoreResult[T] =
+proc deleteOldest(db: WakuMessageStore): MessageStoreResult[void] =
+  let limit = int(float(db.storeCapacity) * maxStoreOverflow/2)
+  var deleteQuery = "DELETE FROM " & TABLE_TITLE & " " &
```
staheri14, line 42: It seems an expensive query that can take a lot of time to conclude (of course depends on the internals of the db, but in the worst case it can take `n`*`m` comparisons where `n` is the DB size and `m` is the selected portion), how about selecting the portion you want, then erasing the entire table and replacing its content with that already selected portion.

waku/v2/node/storage/message/waku_message_store.nim
```diff
@@ -12,22 +13,49 @@ import
...
+  db.numMessages = db.storeCapacity # TODO: subtract the number of actually deleted messages (sqlite delete does not return the num of deleted rows)
+  debug "Oldest messages deleted from DB due to overflow.", storeCapacity=db.storeCapacity, maxStore=db.storeMaxLoad, messagesRemaining=limit
+  ok()
+
+proc init*(T: type WakuMessageStore, db: SqliteDatabase, storeCapacity: int = 50000): MessageStoreResult[T] =
   ## Table is the SQL query for creating the messages Table.
   ## It contains:
   ##  - 4-Byte ContentTopic stored as an Integer
-  ##  - Payload stored as a blob
+  #  - Payload stored as a blob
```
staheri14, line 58:
```suggestion
  ##  - Payload stored as a blob
```

waku/v2/node/storage/message/waku_message_store.nim
```diff
@@ -12,22 +13,49 @@ import
...
 # The code in this file is an adaptation of the Sqlite KV Store found in nim-eth.
 # https://github.com/status-im/nim-eth/blob/master/eth/db/kvstore_sqlite3.nim
-#
-# Most of it is a direct copy, the only unique functions being `get` and `put`.
 
 type
+  # WakuMessageStore implements auto deletion as follows:
+  # The sqlite DB will store up to `storeCapacity` * `maxStoreOverflow` messages.
+  # If more messages are stored, the messages are sorted by `receiverTimestamp` and the oldest ones are deleted,
+  # bringing the total number of stored messages back to `storeCapacity*maxStoreOverflow/2`. The rationale for batch deleting is efficiency.
```
staheri14, line 29: > bringing the total number of stored messages back to `storeCapacity*maxStoreOverflow/2`. The rationale for batch deleting is efficiency. 

I assume the `storeCapacity` indicates the in-memory limit, right? if yes, then I think the `maxStoreOverflow` should be a value larger than `2`, otherwise, while truncating, you may end up deleting DB messages that are still available in the memory. 

For example, for `maxStoreOverflow = 1.3`, the persisted db messages will be `storeCapacity`*0.65, which means the db does not contain all the in-mem messages.

waku/v2/node/storage/message/waku_message_store.nim
```diff
@@ -12,22 +13,49 @@ import
...
+  # We keep half of the overflow window because we delete the oldest messages with respect to `receiverTimestamp` instead of `senderTimestamp`.
```
staheri14, line 30:
```suggestion
  # We keep half of the db capacity because we delete the oldest messages with respect to `receiverTimestamp` instead of `senderTimestamp`.
```

staheri14 (commented):
Looks fine in general. Please see my comments inline. 
A suggestion about the `storeCapacity` name, you may want to consider `inMemCapacity` instead (might be a bit less confusing and more self-explanatory).

waku/v2/node/storage/message/waku_message_store.nim
jm-clius: We are indeed planning several DB and efficiency improvements for subsequent PRs.

waku/v2/node/storage/message/waku_message_store.nim
jm-clius: Right, if the idea is to only remove half of the _overflow_ (not half of the entire capacity), shouldn't the formula for total messages after deletion be something like `storeCapacity*(1 + (maxStoreOverflow - 1)/2)`. A bit of a complicated way to express this though. Perhaps it's easier to just determine the absolute number of overflow messages allowed `(maxStoreOverflow*storeCapacity - storeCapacity)` when initiating the store and always removing half of that?

kaiserd:
Thank you very much for your feedback @staheri14 . I will add tests and then address your feedback.

waku/v2/node/storage/message/waku_message_store.nim
kaiserd: @staheri14 Yes. Good point. The store does not sync between DB and in-memory store.
This is also one of the points we plan to address in the future.
The currently implemented formula is as @jm-clius stated.
Only half of the overflow window gets deleted. The full `storeCapacity` and half of the overflow window are kept.

waku/v2/node/storage/message/waku_message_store.nim
kaiserd: Yes. In the current implementation, most of the messages are kept and only a comparatively small part is deleted.
So, selecting the messages that are kept would take most of the time.
An exception would be if a very large DB of an older version is loaded.

My suggestion for making this query more efficient would be adding an index on `receiverTimestamp`,
which would reduce the cost for the select part to log(m).
I will open an issue for this (+ vacuum) as soon as this issue is closed.

While Index + vacuum will be addressed timely, future steps we envisioned are:

* pure SQL store
  - large storage capacity but slow
  - in-line with status-desktop requirement of keeping messages for 30 day
* adding an efficient cache

I could also add a store efficiency roadmap issue tracking these future goals.

waku/v2/node/storage/message/waku_message_store.nim
kaiserd: We actually keep capactiy + half of the overflow window. See comment above.
I will write this more clearly.

kaiserd:
I pushed a new commit addressing feedback and adding tests.

There is still an issue I have to figure out before opening the PR.
In the [test]() I added, when setting capacity and overload to a larger number, e.g.

```
let
  capacity = 1000
  overload = 1000
```

the messages do not get deleted properly anymore.
I added a debug statement printing the number of messages actually in the DB after the deletion.
This could be removed later for efficiency, because the count has to go over the DB one more time.

waku/v2/node/storage/message/waku_message_store.nim
kaiserd: > Perhaps it's easier to just determine the absolute number of overflow messages allowed (maxStoreOverflow*storeCapacity - storeCapacity) when initiating the store and always removing half of that?

I now added a `deleteWindow` to `WakuMessageStore`.

waku/v2/node/storage/message/waku_message_store.nim
```diff
@@ -12,18 +13,59 @@ import
 
 export sqlite
 
+logScope:
+  topics = "wakuMessageStore"
+
 const TABLE_TITLE = "Message"
+const maxStoreOverflow = 1.3
```
staheri14, line 20: I suggest using capital case instead (it is easier to distinguish local vars from const)
```suggestion
const MAX_STORE_OVERFLOW = 1.3
```

waku/v2/node/storage/message/waku_message_store.nim
```diff
@@ -12,18 +13,59 @@ import
...
 
 # The code in this file is an adaptation of the Sqlite KV Store found in nim-eth.
 # https://github.com/status-im/nim-eth/blob/master/eth/db/kvstore_sqlite3.nim
-#
-# Most of it is a direct copy, the only unique functions being `get` and `put`.
 
 type
+  # WakuMessageStore implements auto deletion as follows:
+  # The sqlite DB will store up to `storeMaxLoad = storeCapacity` * `maxStoreOverflow` messages, giving an overflow window of (storeCapacity*maxStoreOverflow - storeCapacity).
```
staheri14, line 27: Thanks, this new formulation addresses my past comment

staheri14 (approved):
LGTM! approved it though if it is still a WIP and needed another review, ping me again.

waku/v2/node/storage/message/waku_message_store.nim
kaiserd: I did not use capital case because of [NEP1](https://nim-lang.org/docs/nep1.html#introduction-naming-conventions), and I did not find anything overriding this in [status style guide](https://status-im.github.io/nim-style-guide/00_introduction.html).
No strong opinion on my side. Wdyt?

kaiserd:
Thank you for reviewing the PR again :). I will open the PR as soon as I figure out the problem in the test described above.

kaiserd:
The PR is now open :).

The issue in the test mentioned above occurred because the content of the messages in the test was only a single byte, which wrapped around when storing more than 256 messages.
Since the id only depends on the message digest, which in turn depends on the message payload and `pubsubTopic` (only one in the test), sets of messages shared the same id.

A second issue was the second-resolution of the receiverTimestamp which I overlooked.
A comment in the test said that 1ms of waiting would result in different timestamps. I removed that comment and the 1ms wait.

waku/v2/node/storage/message/waku_message_store.nim
jm-clius: Yeah, no strong opinions either, other than that we've certainly not been consistent so far. :/ I suggest we go with the linked Nim convention and specifically PascalCase - this is also what `nim-libp2p` uses. A good exercise for our next focused refactoring effort. Thanks for highlighting, @staheri14.

jm-clius (approved):
LGTM! Thanks :)

jm-clius:
@kaiserd can this be merged, or are you still waiting for one more review?

waku/v2/node/storage/message/waku_message_store.nim
```diff
@@ -12,18 +13,59 @@ import
 
 export sqlite
 
+logScope:
+  topics = "wakuMessageStore"
+
 const TABLE_TITLE = "Message"
+const MaxStoreOverflow = 1.3
```
staheri14, line 20: Although it would never happen probably, it is prudent to add a comment that this value must be larger than `1`, otherwise the math at line 98, i.e., 
> let deleteWindow = int(float(storeMaxLoad - storeCapacity) / 2)
will yield a negative value.

staheri14 (approved):
LGTM! feel free to apply my last comment in your subsequent PRs, nothing urgent.

waku/v2/node/storage/message/waku_message_store.nim
kaiserd: I will do this. Thanks :).

PR #907, opened 2022-03-23
Fixed building on Apple clang and M1

jemboh:
Apple's clang (v13) doesn't support "-march=native" on M1.
Checking `clang --print-supported-cpus` shows that it 
_does_ support `apple-m1`.


jemboh:
Thanks to @richard-ramos!

jm-clius (approved):
Not an expert, but seems reasonable to me. :)

PR #909, opened 2022-03-23
Copy rln lib into wakunode2 docker container

staheri14:
This change is needed to enable rln-relay protocol (which relies on rln lib) on waku test fleets. The rln lib is dynamically linked hence needs to be loaded separately. 

The `Copy` operation for the rln lib is made optional (by appending the `*` at the end of the source file) to account for the cases that the rln lib does not exist e.g., in waku prod fleets which are built with the rln compiler flag turned off hence the rln lib `librln.so` never gets produced.


oskarth (approved):
TIL about making COPY operation optional with * wildcard!

jm-clius:
> TIL about making COPY operation optional with * wildcard!

Same.

jakubgs (approved):
Looks good, though I wonder, what's the reason we can't link it statically and avoid this entirely?

staheri14:
> what's the reason we can't link it statically and avoid this entirely?

That is a future work;  just has not been prioritized yet as the dynamic version also works fine.

PR #910, opened 2022-03-24
chore(rln): add cleanrln make rule + docu

kaiserd:
During the rln test, I got a compile error due to an outdated crate that did not get recompiled.

This PR 

* adds a trouble shoot section to the docu
* adds a `cleanrln` rule to the Makefile
* adds cleaning the rln crate to the `clean` rule

Rationale for separate `cleanrln` rule:  avoid having to rebuild the whole project.
Later, this could also be added to the `update` rule.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -223,6 +223,11 @@ clean: | clean-common
 ifneq ($(USE_LIBBACKTRACE), 0)
 	+ $(MAKE) -C vendor/nim-libbacktrace clean $(HANDLE_OUTPUT)
 endif
+	cargo clean --manifest-path vendor/rln/Cargo.toml
+
+# clean the rln build (forces recompile of old crates on next build)
+cleanrln:
+	cargo clean --manifest-path vendor/rln/Cargo.toml
 
 endif # "variables.mk" was not included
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -223,6 +223,11 @@ clean: | clean-common
 ifneq ($(USE_LIBBACKTRACE), 0)
 	+ $(MAKE) -C vendor/nim-libbacktrace clean $(HANDLE_OUTPUT)
 endif
+	cargo clean --manifest-path vendor/rln/Cargo.toml
+
+# clean the rln build (forces recompile of old crates on next build)
+cleanrln:
```
staheri14, line 229: Thanks for running the test and sharing the issue! 
There is a similar target `clean` which is supposed to clean the old artifacts of the rln compilation, have you tried it? let me know whether that works or not

docs/tutorial/rln-chat2-live-testnet.md
```diff
@@ -123,3 +123,21 @@ Connecting to storenode: 16Uiu2HAmPLe7Mzm8TsYUubgCAW1aJoeFScxrLj8ppHFivPo97bUZ
 >> /exit
 quitting...
 ```
+
+# Trouble shooting
+
+## compilation error: found possibly newer version of crate
```
staheri14, line 129: That is a good idea to have a troubleshooting section, if the `make clean` worked, please update the command in this description, and then we are good to merge the PR.

staheri14 (approved):
Thanks for reporting and resolving the issue!

kaiserd:
Thank you for the reviews :). Merging this PR.

PR #912, opened 2022-03-24
docs: added bug reporting section to main readme

jm-clius:
Adds a section to the `README.md` on bug reporting.

I've also added a [`critical`](https://github.com/status-im/nim-waku/labels/critical) label, which can be used on high-priority issues to assist with triaging. We could also consider adding a special column to [the project board](https://github.com/status-im/nim-waku/projects/1) to increase visibility of new critical issues/bug reports.


oskarth (approved):
Good idea :)

Issue #906, opened 2022-03-23
ENR is missing the UDP port


richard-ramos:
While attempting to use the status.test fleet for DiscV5, go-waku reported that the ENR was not valid due to missing the UDP port
![image](https://user-images.githubusercontent.com/1106587/159733750-7591d06b-b905-498c-9cd5-e4858f832072.png)

```
bad bootstrap node "enr:-Iu4QNUaBXrXvciIzPUgsPgiqXwA9LZgPE2y6B5VSxKFy0klIeYx8lm0xPH8zRM3OAA5aFSuJnHIwfklc0Tzd_AFMywBgmlkgnY0gmlwhMCoARSJc2VjcDI1NmsxoQKhU8PVgaVNi3-fbU-6OiKWdrYVu2fjCPj2Q30HqbAi_IN0Y3CC6mCFd2FrdTID": missing UDP port
```

D4nte:
So is discv5 executed over UDP only? Both TCP and UDP?

kaiserd:
Discv5 only uses UDP. From the [spec](https://github.com/ethereum/devp2p/blob/master/discv5/discv5-theory.md) :

> For the purposes of this protocol, the node must at least provide an IP address ("ip" or "ip6" key) and UDP port ("udp" key) in order to have it's record relayed in the DHT.

jm-clius:
@kaiserd I think this issue can be closed?

jm-clius:
Closing, as the linked fix has been merged.

PR #913, opened 2022-03-24
fix(discv5): info print ENR containing discv5 udp port

kaiserd:
This PR addresses #906.

This is a minimal fix printing the actually discoverable ENR managed by the discv5 subsystem.

Wakunode holds another ENR that got printed before.
I will look into consolidating these two ENR instances into one in the future (not critial for now).
This PR should be a minimal fix solving this `citical` issue.

(successfully tested setting up a local fleet using the printed ENR)

### DIFF
--- a/waku/v2/node/wakunode2.nim
+++ b/waku/v2/node/wakunode2.nim
@@ -893,6 +893,7 @@ proc startDiscv5*(node: WakuNode): Future[bool] {.async.} =
     asyncSpawn node.runDiscv5Loop()
 
     debug "Successfully started discovery v5 service"
+    info "Discv5: discoverable ENR ", enr = node.wakuDiscV5.protocol.localNode.record.toURI()
     return true
 
   return false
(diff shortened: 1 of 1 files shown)


richard-ramos (approved):

oskarth (approved):

jm-clius (changes requested):
We do still need this to be logged even if discv5 is not enabled for DNS discovery purposes. Alternatives for quick fix then:
a) log both ENRs (with slightly different log messages)
b) if discv5 is enabled, just replace the `node.enr` with the discv5 version once that gets created.

Although (a) is confusing (so should be temporary until we have consolidated the enrs), it's probably our best option as we need to include `dns4` `multiaddrs` fields in the enr for DNS discovery, and this has not been introduced for the discv5 enr yet.

kaiserd:
Thank you! 
I implemented solution (a).

Because, as you said, this might introduce confusion, I will add an issue regarding consolidating these ENRs soon.
Still, with this solution the fact that there are two ENRs is explicit.

jm-clius (approved):

Issue #859, opened 2022-02-21
RPC Debug API call that returns version and commit


jakubgs:
For debugging purposes it's very useful to be able to verify the current running version of Node via the JSON RPC API.

Something like:
```
 > ./rpc.sh get_waku_v2_debug_v1_version
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": {
    "version": "0.1.0-beta",
    "commit": "1234abcd"
  }
}
```
Or possibly just extend the info returned by `get_waku_v2_debug_v1_info`?

jm-clius:
Good idea. Thanks.

jakubgs:
I've made a PR that adds commit to Docker images as a label: https://github.com/status-im/nim-waku/pull/860

jakubgs:
Looks like we are getting nothing on `wakuv2.test`:
```
admin@node-01.do-ams3.wakuv2.test:~ % /docker/nim-waku-v2/rpc.sh get_waku_v2_debug_v1_version
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": "n/a"
}
```

PR #919, opened 2022-03-28
feat(jsonrpc): add call returning git version

kaiserd:
This PR addresses #859.

There are other ways of getting access to the git version number from within Nim.
This requires minimal changes and no further build deps.

The current format is simple, just returing a single string of the form `<tag>-<commit-hash>` if there is a tag for the respective commit hash. (See output of `git describe --abbrev=6 --always --tags`).
If desired, I could split it into the format that is given as an example in the issue.
Either by

* calling `git describe` twice during `make`
* splitting the string in Nim

Alternatively, we could also print the most recent tag on any branch (this would require calling `git describe` twice).
Right now, this might be confusing, because we only have tags up to v0.6.

I chose to introduce a new call `get_waku_v2_debug_v1_version`, because `get_waku_v2_debug_v1_info` returns `node.info()` and making the git-version part of `WakuInfo` felt off.

If you agree with this implementation, I will add the call to the docs.

## TODO:

- [ ] add docs

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -99,6 +99,10 @@ endif
 # use a separate waku discv5 network with `protocol-id="d5waku"`
 NIM_PARAMS := $(NIM_PARAMS) -d:discv5_protocol_id:d5waku
 
+# git version for JSON RPC call
+GIT_VERSION := "$(shell git describe --abbrev=6 --always --tags)"
+NIM_PARAMS := $(NIM_PARAMS) -d:git_version:\"$(GIT_VERSION)\"
+
 deps: | deps-common nat-libs waku.nims rlnlib
 ifneq ($(USE_LIBBACKTRACE), 0)
 deps: | libbacktrace
--- a/waku/v2/node/jsonrpc/debug_api.nim
+++ b/waku/v2/node/jsonrpc/debug_api.nim
@@ -8,6 +8,9 @@ import
 logScope:
   topics = "debug api"
 
+const 
+  git_version {.strdefine.} = "n/a"
+
 proc installDebugApiHandlers*(node: WakuNode, rpcsrv: RpcServer) =
 
   ## Debug API version 1 definitions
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -99,6 +99,10 @@ endif
 # use a separate waku discv5 network with `protocol-id="d5waku"`
 NIM_PARAMS := $(NIM_PARAMS) -d:discv5_protocol_id:d5waku
 
+# git version for JSON RPC call
+GIT_VERSION := "$(shell git describe --abbrev=6 --always --tags)"
+NIM_PARAMS := $(NIM_PARAMS) -d:git_version:\"$(GIT_VERSION)\"
```
jakubgs, line 104: I would not recommend setting variables that call commands with `shell` using immediate assignment(`:=`), you should be using lazy assignment(`=`) instead to avoid calling this command EVERY SINGLE TIME `make` is called.

For more details see:

* https://www.gnu.org/software/make/manual/html_node/Flavors.html
* https://stackoverflow.com/questions/448910/

kaiserd: I could do that for `GIT_VERSION`, but `NIM_PARAMS` is already used quite a lot with `:=` (and doing a single assignment with = for a recursive variable does not work). So `GIT_VERSION` would still be eagerly evaluated.

`NIM_PARAMS += -d:git_version:\"$(GIT_VERSION)\"` I did this in another PR before, but this will evaluate in the same way the other uses of `NIM_PARAMS` do and we reverted that to be consistent with the other assignments.

Should we rethink this for the whole Makefile and see which vars can actually be lazily evaluated? But imo, this should be in a new PR. Wdyt?

kaiserd: I will put this on my list :).

jm-clius (approved):
LGTM

oskarth (approved):

PR #930, opened 2022-03-30
feat(noise): add support to ChaChaPoly encryption

s1fr0:
This PR adds support to the ChaChaPoly symmetric cipher for encrypting Waku payload message contents. 

Specifically, it adds:
- Encryption/Decryption primitives
- Some unit tests

It partially addresses https://github.com/status-im/nim-waku/issues/881.

### DIFF
--- a/tests/v2/test_waku_noise.nim
+++ b/tests/v2/test_waku_noise.nim
@@ -1,0 +1,6 @@
+{.used.}
+
+import
+  testutils/unittests,
+  ../../waku/v2/protocol/waku_noise/noise,
+  ../test_helpers
--- a/tests/v2/test_waku_noise.nim
+++ b/tests/v2/test_waku_noise.nim
@@ -4,3 +4,20 @@ import
   testutils/unittests,
   ../../waku/v2/protocol/waku_noise/noise,
   ../test_helpers
+
+
+procSuite "Waku Noise":
+  
+  let rng = rng()
+
+  test "ChaChaPoly Encryption/Decryption":
+
+    let cipherState = randomChaChaPolyCipherState(rng[])
+
+    let
+      plaintext: seq[byte] = randomSeqByte(rng[], 128)
+      ciphertext: ChaChaPolyCiphertext = encrypt(cipherState, plaintext)
+      dec_ciphertext: seq[byte] = decrypt(cipherState, ciphertext)
+
+    check: 
+      plaintext == dec_ciphertext
--- a/tests/v2/test_waku_noise.nim
+++ b/tests/v2/test_waku_noise.nim
@@ -17,7 +17,7 @@ procSuite "Waku Noise":
     let
       plaintext: seq[byte] = randomSeqByte(rng[], 128)
       ciphertext: ChaChaPolyCiphertext = encrypt(cipherState, plaintext)
-      dec_ciphertext: seq[byte] = decrypt(cipherState, ciphertext)
+      decryptedCiphertext: seq[byte] = decrypt(cipherState, ciphertext)
 
     check: 
-      plaintext == dec_ciphertext
+      plaintext == decryptedCiphertext
--- a/waku/v2/protocol/waku_noise/noise.nim
+++ b/waku/v2/protocol/waku_noise/noise.nim
@@ -20,11 +20,8 @@ import libp2p/utility
 import libp2p/errors
 import libp2p/crypto/[crypto, chacha20poly1305]
 
-when defined(libp2p_dump):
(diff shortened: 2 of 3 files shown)


waku/v2/protocol/waku_noise/noise.nim
```diff
@@ -0,0 +1,76 @@
+## Nim-LibP2P
```
oskarth, line 1: Leftover copy paste? I don't think we have this part in other nim files

Maybe just mention it was taken/inspired by nim-libp2p with brief link?

waku/v2/protocol/waku_noise/noise.nim
```diff
@@ -0,0 +1,76 @@
...
+
+  NoiseError* = object of LPError
+  NoiseDecryptTagError* = object of NoiseError
+
+# ChaChaPoly encryption
+proc encrypt*(
+    state: ChaChaPolyCipherState,
+    plaintext: openArray[byte]): ChaChaPolyCiphertext
+    {.noinit, raises: [Defect].} =
+  #TODO: add padding
```
oskarth, line 52: Can we raise a Github issue for this maybe?

waku/v2/protocol/waku_noise/noise.nim
oskarth: I'd also add a link to RFC

waku/v2/protocol/waku_noise/noise.nim
oskarth: OK not to address straight away, as long as we capture it somewhere as it may impact security

oskarth (approved):

waku/v2/protocol/waku_noise/noise.nim
s1fr0: Sure, in fact padding is a requirement we have for encrypted after-handshake transport messages. Here we're still adding vanilla ChaChaPoly encryption, so padding is not necessarily a requirement, but it would be nice to have it as an option later on.

waku/v2/protocol/waku_noise/noise.nim
s1fr0: Yes, my fault with copy paste from my bigger WIP source! No problem adding links in the source, although this part doesn't directly come from libp2p noise implementation.

waku/v2/protocol/waku_noise/noise.nim
```diff
@@ -0,0 +1,75 @@
...
+import libp2p/crypto/[crypto, chacha20poly1305]
+
+when defined(libp2p_dump):
+  import libp2p/debugutils
+
+logScope:
+  topics = "nim-waku noise"
+
+const
+  # Empty is a special value which indicates k has not yet been initialized.
```
staheri14, line 30: what is `k` in the comment?

tests/v2/test_waku_noise.nim
```diff
@@ -0,0 +1,6 @@
+{.used.}
+
+import
+  testutils/unittests,
```
staheri14, line 4: the test file looks empty, is it intentional?

s1fr0: Yes it was intentionally left blank since this PR comes from a process of cherry picking commits to split the big WIP implementation into smaller PRs. All changes to test_waku_noise.nim can in fact be discarded from this PR, but if we quickly merge this and all the other noise related PR then would not be necessary.

jm-clius: Not important to address in this PR, but in future I think it's preferable having each incremental PR be reasonable as a stand-alone that can be reviewed in its entirety. It makes sense to have open TODOs, of course, but since empty modules can't really be reviewed we increase the risk of having dead code in future that we forget about.

jm-clius (changes requested):
Looks good. Some minor comments about Nim conventions.

s1fr0:
@jm-clius addressed comments in https://github.com/status-im/nim-waku/commit/36b1104ed0a4056448f00bccd2c4b4ef182a5c34, https://github.com/status-im/nim-waku/commit/9d12464f1d0aaa17ff99ec02c324ce710702f725 and https://github.com/status-im/nim-waku/commit/852d8d91c897a4a28ecb92eb6db21b8a33414ef3

I added a new test to test encryption/decryption (it was missing before) and the file is no more empty. Removed debug tools import used in libp2p implementation. Restyled code according to suggested Nim style guide.

jm-clius (approved):
LGTM, thanks!

PR #941, opened 2022-04-11
chore|feat (waku-rln-relay): modules reorganization|Initial test for capturing events using nim-web3

staheri14:
As the first step towards enabling [dynamic group management](https://github.com/status-im/nim-waku/issues/894) in the rln-relay protocol, we need to be able to collect events emitted from the membership contract. One of the available nim libraries for this sake is nim-web3.  In the current PR, the initial test for interaction with the RLN contract using nim-web3 functions is introduced.

Further updates included in the PR:
- This PR also restructures rln relay tests by splitting the on-chain and off-chain tests. Furthermore, the rln-relay solidity contract and its binary codes are isolated from the test files and moved to a distinct module.
- Due to a missing capability in nim-web3, i.e., lacking support for decoding indexed event arguments, the rln-relay contract has been modified and recompiled to circumvent this issue (by amending the events types from `indexed` to unindexed). The corresponding new bytecodes are incorporated in this PR.  The new byte-codes are available in the `MEMBERSHIP_CONTRACT_CODE` variable defined in the `rln_relay_contract.nim` file.
- Beautifies the rln-relay modules.

# Nest Steps
In the following PRs, an async component will be added to the rln-relay protocol which will utilize the event subscription capability to listen to the member addition and deletion of the contract and reflect those on the local Merkle tree.


waku/v2/protocol/waku_rln_relay/waku_rln_relay_types.nim
```diff
@@ -77,49 +77,249 @@ type WakuRLNRelay* = ref object
...
 type MessageValidationResult* {.pure.} = enum
-    Valid, Invalid, Spam
+  Valid, Invalid, Spam
 
 # inputs of the membership contract constructor
 # TODO may be able to make these constants private and put them inside the waku_rln_relay_utils
-const 
-  MEMBERSHIP_FEE* = 5.u256
+const
+  MembershipFee* = 5.u256
```
jm-clius, line 92: For interest's sake, was this `const` identifier changed to `PascalCase` by `nimpretty`? No strong feelings, but I indeed prefer PascalCase for constants as is done here, and this is also what the official style guide suggests: https://nim-lang.org/docs/nep1.html#introduction-naming-conventions

I know we've been very inconsistent with this, though, but would suggest that we try to remain consistent at least within a module (i.e. either PascalCase or CAPITAL_LETTERS in this one). We can then see it as a single refactoring task to refactor all `const`s across the codebase to `PascalCase`, once we agree on a convention.

jm-clius (approved):
LGTM!

waku/v2/protocol/waku_rln_relay/waku_rln_relay_types.nim
staheri14: > For interest's sake, was this const identifier changed to PascalCase by nimpretty?

I changed it myself, it was referred to in some other parts of the code as `MembershipFee`, I  wanted them to be consistent so I changed it.

>  No strong feelings, but I indeed prefer PascalCase for constants as is done here, and this is also what the official style guide suggests: https://nim-lang.org/docs/nep1.html#introduction-naming-conventions

I also prefer the Camal case and almost followed this in my implementation. Will revert this one also back to Upper case.

staheri14: Addressed in https://github.com/status-im/nim-waku/pull/941/commits/3d1038a7789d18c868e18b3d328817952f035947

waku/v2/protocol/waku_rln_relay/rln_relay_contract.nim
```diff
@@ -0,0 +1,107 @@
...
...e60608f01518a090101019050828c8886098d8a88098e8c8a0901010193508b82830992508b828d8586090991508b81820992508b818d8586090990508b84850992508b848d8586090993507f1def7475769de3cd372283c7d36bd153637723ff14615542bfe8b27108ad4de79250828c60408d015186098d60208e015184098e8e5186090101019550828c8b86098d60808e015184098e60608f015186090101019450828c8886098d8a84098e8c860901010193508b86870992508b868d8586090995508b85860992508b858d8586090994508b84850992508b848d8586090993507f119217aded5f1ebf525a69e23a08f6a976178e5939646f6c87033299d2d3c8749250828c60408d015186098d60208e015188098e8e518a090101019150828c8b86098d60808e015188098e60608f01518a090101019050828c8886098d8a88098e8c8a0901010193508b82830992508b828d8586090991508b81820992508b818d8586090990508b84850992508b848d8586090993507f036bd8311510a8d03c5b354f03c664f9f2d23f57b1e4286d1938f8dbe4a7c28c9250828c60408d015186098d60208e015184098e8e5186090101019550828c8b86098d60808e015184098e60608f015186090101019450828c8886098d8a84098e8c86090101019350505089848509905089848b83840909935089838409905089838b83840909925089828309905089828b838409505050929b9a505050505050505050505056fea2646970667358221220eef5419e4124debe6900ddea7f02a5dfbb1c6b867ba8e61b34edbc913728191664736f6c63430007040033"
```
oskarth, line 102: I think this is fine considering it is fairly small, but another option here is to stick this bytecode in a `resource` folder and load it from the file system. This makes it more amendable to scripting etc, and separates code from data better.

waku/v2/protocol/waku_rln_relay/waku_rln_relay_types.nim
```diff
@@ -77,49 +77,249 @@ type WakuRLNRelay* = ref object
...
+  STATIC_GROUP_KEYS* = @[("e9a4d05b1f539d65c59015a079ee89aabeafbcfc9734342d9559f81601e85417",
```
oskarth, line 112: Same re resource folder here

oskarth (approved):

kaiserd (approved):
LGTM. We could manage the solidity code in a vendor submodule and compile it as part of the build process. This would be more transparent.

tests/v2/test_waku_rln_relay_onchain.nim
```diff
@@ -0,0 +1,263 @@
...
+    let accounts = await web3.provider.eth_accounts()
+    web3.defaultAccount = accounts[1]
+    let add = web3.defaultAccount
+    debug "contract deployer account address ", add
+
+    # prepare a contract sender to interact with it
+    var sender = web3.contractSender(MembershipContract,
+        contractAddress) # creates a Sender object with a web3 field and contract address of type Address
+
+    # send takes three parameters, c: ContractCallBase, value = 0.u256, gas = 3000000'u64 gasPrice = 0
```
s1fr0, line 149: `c: ContractCallBase` is one of the parameters? (if yes then are 4)
`value` should be `5.u256 ` instead of `0.u256`?

```suggestion
    # send takes three parameters, c: ContractCallBase, value = 0.u256, gas = 3000000'u64, gasPrice = 0
```

tests/v2/test_waku_rln_relay_onchain.nim
```diff
@@ -0,0 +1,263 @@
...
+    # should use send proc for the contract functions that update the state of the contract
+    let tx = await sender.register(20.u256).send(value = MEMBERSHIP_FEE)
+    debug "The hash of registration tx: ", tx # value is the membership fee
+
+    # var members: array[2, uint256] = [20.u256, 21.u256]
```
s1fr0, line 154: can be removed?

tests/v2/test_waku_rln_relay_onchain.nim
```diff
@@ -0,0 +1,263 @@
...
+    let
+      web3 = await newWeb3(ETH_CLIENT)
+      accounts = await web3.provider.eth_accounts()
+      # choose one of the existing accounts for the rln-relay peer
+      ethAccountAddress = accounts[9]
+    await web3.close()
+
+    # create an RLN instance
+    var rlnInstance = createRLNInstance()
+    check: rlnInstance.isOk == true
```
s1fr0, line 178: Maybe worth aligning the source to check all `isOk` alone or against true
```suggestion
    check:
      rlnInstance.isOk == true
```

tests/v2/test_waku_rln_relay_onchain.nim
```diff
@@ -0,0 +1,263 @@
...
+
+    # generate the membership keys
+    let membershipKeyPair = membershipKeyGen(rlnInstance.value)
+    check: membershipKeyPair.isSome
```
s1fr0, line 182:
```suggestion
    check:
      membershipKeyPair.isSome
```

tests/v2/test_waku_rln_relay_onchain.nim
```diff
@@ -0,0 +1,263 @@
...
+
+    # deploy the contract
+    let membershipContractAddress = await uploadRLNContract(ETH_CLIENT)
+
+    # prepare rln-relay inputs
+    let
+      web3 = await newWeb3(ETH_CLIENT)
+      accounts = await web3.provider.eth_accounts()
+      # choose one of the existing account for the rln-relay peer
+      ethAccountAddress = accounts[9]
```
s1fr0, line 212: The index `9` is platform/account independent? Is 0 safer?

tests/v2/test_waku_rln_relay_onchain.nim
```diff
@@ -0,0 +1,263 @@
...
+    let index = uint(5)
+
+    # Create a group of 10 members
+    var group = newSeq[IDCommitment]()
+    for i in 0..10:
+      var member_is_added: bool = false
+      if (uint(i) == index):
+        #  insert the current peer's pk
+        group.add(keypair.get().idCommitment)
+        member_is_added = rln.insertMember(keypair.get().idCommitment)
```
s1fr0, line 233: Is it better CamelCase?

waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
```diff
@@ -422,23 +445,24 @@ proc getCurrentEpoch*(): Epoch =
...
 
-proc validateMessage*(rlnPeer: WakuRLNRelay, msg: WakuMessage, timeOption: Option[float64] = none(float64)): MessageValidationResult =
+proc validateMessage*(rlnPeer: WakuRLNRelay, msg: WakuMessage,
+    timeOption: Option[float64] = none(float64)): MessageValidationResult =
   ## validate the supplied `msg` based on the waku-rln-relay routing protocol i.e.,
   ## the `msg`'s epoch is within MAX_EPOCH_GAP of the current epoch
   ## the `msg` has valid rate limit proof
   ## the `msg` does not violate the rate limit
-  ## `timeOption` indicates Unix epoch time (fractional part holds sub-seconds) 
+  ## `timeOption` indicates Unix epoch time (fractional part holds sub-seconds)
```
s1fr0, line 462: Why times in RLN Relay are intended as floats? Would it better to align everything to int64 or there are some particular reason to keep them as floats?

waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
```diff
@@ -422,23 +445,24 @@ proc getCurrentEpoch*(): Epoch =
 
 proc compare*(e1, e2: Epoch): int64 =
```
s1fr0, line 446: To my intuition "compare" would return -1,0,1 depending if e1 is lower, equal, greater than e2 (as happens for the `cmp` function common to many languages). Definitely not relevant, but maybe clearer to call it diff, epochGap or something like this?

s1fr0 (approved):
Great job and very instructive going through it! Left some comments.

waku/v2/protocol/waku_rln_relay/rln_relay_contract.nim
oskarth: I think this gets generated when running https://github.com/kilic/rlnapp

In a production environment, the bytecode for deployed contract (and source) would be available e.g. here https://etherscan.io/address/0x00000000219ab540356cbb839cbe05303d7705fa#code

staheri14: > I think this is fine considering it is fairly small, but another option here is to stick this bytecode in a resource folder and load it from the file system. This makes it more amendable to scripting etc, and separates code from data better.

Thanks for the comment. A few points about the current contract bytecode: 
1) The byte code is only used for testing purposes. In the final deployment, the contract is to be deployed externally (using e.g., truffle) and not from our nim codebase.   
2) The bytecode was initially inside a separate file, but I figured it incurs unnecessary I/O operation
3) Defining it as a const is similar to what nim-web3 tests look like  (https://github.com/status-im/nim-web3/blob/2b9e8166d444441847de3e8b35d442e540d356db/tests/depositcontract.nim)

Please let me know wdyt?

staheri14: @s1fr0 As Oskar said,  if you compile the RLN.sol contract code in the rlnapp repository https://github.com/kilic/rlnapp you will get to the same bytecode.

tests/v2/test_waku_rln_relay.nim
```diff
@@ -495,44 +246,44 @@ suite "Waku rln relay":
...
       get_root_successful2 = get_root(rlnInstance.value, rootPtr2)
     check:
       get_root_successful2
       root2.len == 32
 
-    # delete the first member 
+    # delete the first member
     var deleted_member_index = MembershipIndex(0)
     let deletion_success = delete_member(rlnInstance.value, deleted_member_index)
     check deletion_success
```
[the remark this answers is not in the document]
staheri14: Thanks for the suggestion, though wondering whether it is a new convention to split the `check` from the condition? otherwise, the current version works, why the split?

tests/v2/test_waku_rln_relay_onchain.nim
staheri14: Good catch, my intention was to type `0` but `9` was hit mistakenly!

waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
staheri14: Fair enough, tracking issue https://github.com/status-im/nim-waku/issues/946

tests/v2/test_waku_rln_relay_onchain.nim
staheri14: sure, tracking issue https://github.com/status-im/nim-waku/issues/946

waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
staheri14: Makes sense, it is now renamed to `diff`.  https://github.com/status-im/nim-waku/pull/941/commits/fbd5d53b55aafc2e2b936792b415c8f2d4ae83e0

tests/v2/test_waku_rln_relay.nim
```diff
@@ -279,68 +29,69 @@ procSuite "Waku rln relay":
...
       parametersBuffer = Buffer(`ptr`: addr(pbytes[0]), len: len)
     check:
       # check the parameters.key is not empty
       pbytes.len != 0
 
-    var 
+    var
       rlnInstance: RLN[Bn256]
-    let res = new_circuit_from_params(merkleDepth, addr parametersBuffer, addr rlnInstance)
+    let res = new_circuit_from_params(merkleDepth, addr parametersBuffer,
```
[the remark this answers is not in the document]
staheri14: (Not based on an agreed-upon conversion) I used snake_case here to differentiate between the rln lib  Nim bindings and other utility procs.

waku/v2/protocol/waku_rln_relay/rln_relay_contract.nim
staheri14: I created this tracking issue for all the refactoring tasks raised in this PR https://github.com/status-im/nim-waku/issues/946

waku/v2/protocol/waku_rln_relay/waku_rln_relay_types.nim
staheri14: sure, please see the tracking issue https://github.com/status-im/nim-waku/issues/946

tests/v2/test_waku_rln_relay_onchain.nim
```diff
@@ -0,0 +1,263 @@
...
+  let
+    hasherReceipt = await web3.deployContract(POSEIDON_HASHER_CODE)
+    hasherAddress = hasherReceipt.contractAddress.get
+  debug "hasher address: ", hasherAddress
+
+
+  # encode membership contract inputs to 32 bytes zero-padded
+  let
+    membershipFeeEncoded = encode(MEMBERSHIP_FEE).data
+    depthEncoded = encode(MERKLE_TREE_DEPTH.u256).data
```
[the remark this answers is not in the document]
staheri14: The registration fee is supposed to be set so that an attacker cannot exhaust a large portion of the leaves for free. I suspect such an entity with lots of coins is a reasonable adversarial assumption. An attacker like that needs to possess almost 2^20*30 USD = 30M USD! (see this issue for the registration cost https://github.com/vacp2p/research/issues/56)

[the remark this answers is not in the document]
staheri14: At the termination of the DoS attack, the attacker will lose half of his deposit so he would not be able to recover his entire money back. In the preceding example, he will burn 15M USD!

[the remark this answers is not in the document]
staheri14: >As we already discussed https://github.com/vacp2p/rfc/pull/484#pullrequestreview-879074788, it may be worth brainstorm on alternative constructions which provide a good decentralized/efficiency tradeoff.Nevertheless, the ability to have a more decentralized group management + good incentives would be a nice idea regardless.

tests/v2/test_waku_rln_relay.nim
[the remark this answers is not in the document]
s1fr0: For these, I had in mind the suggestion provided [here](https://github.com/status-im/nim-waku/pull/931#discussion_r841574139), to have some consistency among single-check tests. While there there were inconsistencies between single and two-lines single-checks (like the ones in the suggestion), here I see you used either `check: condition` and `check condition`. I personally think the most versatile version which cover all these cases for single/multi checks is 

```
check:
  condition(s)
```

since it further allows to easily add new checks when/if needed by future PRs. But this is minor suggestion and we might indeed agree on some common convention.

tests/v2/test_waku_rln_relay.nim
```diff
@@ -776,25 +530,26 @@ suite "Waku rln relay":
     check verified == false
   test "toEpoch and fromEpoch consistency check":
     # check edge cases
-    let 
+    let
       time = uint64.high
```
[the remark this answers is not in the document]
staheri14: to resolve confusion it is renamed, see https://github.com/status-im/nwaku/pull/941/commits/81174aa5e7816b20140466a6efe08baca9dd2f76

tests/v2/test_waku_rln_relay_onchain.nim
```diff
@@ -0,0 +1,263 @@
...
+    # create an RLN instance
+    var rlnInstance = createRLNInstance()
+    check: rlnInstance.isOk == true
+    # generate the membership keys
+    let membershipKeyPair = membershipKeyGen(rlnInstance.value)
+    check: membershipKeyPair.isSome
+    let pk = membershipKeyPair.get().idCommitment.toUInt256()
+    debug "membership commitment key", pk = pk
+
+    # test ------------------------------
```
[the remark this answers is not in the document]
staheri14: Which part is unclear?

tests/v2/test_waku_rln_relay_onchain.nim
staheri14: > value should be 5.u256 instead of 0.u256?

0 is its default value but can be anything if the caller sets a different value

staheri14: Also edited the comment https://github.com/status-im/nwaku/pull/941/commits/be242a607c57eea03c9aa3027e0640406cc3a545

tests/v2/test_waku_rln_relay_onchain.nim
staheri14: this part is commented out since the batch registration is not tested yet, but is conceivable to be used in the future

tests/v2/test_waku_rln_relay_onchain.nim
```diff
@@ -0,0 +1,263 @@
...
+    debug "contract deployer account address ", add
+
+    # prepare a contract sender to interact with it
+    var sender = web3.contractSender(MembershipContract,
+        contractAddress) # creates a Sender object with a web3 field and contract address of type Address
+
+    # send takes three parameters, c: ContractCallBase, value = 0.u256, gas = 3000000'u64 gasPrice = 0
+    # should use send proc for the contract functions that update the state of the contract
+    let tx = await sender.register(20.u256).send(value = MEMBERSHIP_FEE)
+    debug "The hash of registration tx: ", tx # value is the membership fee
```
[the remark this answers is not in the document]
staheri14: right, fixed https://github.com/status-im/nwaku/pull/941/commits/dc2a57563d7180e1f3fde952724dfbc88eaa5de6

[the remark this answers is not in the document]
staheri14: fixed https://github.com/status-im/nwaku/pull/941/commits/dc2a57563d7180e1f3fde952724dfbc88eaa5de6

tests/v2/test_waku_rln_relay.nim
```diff
@@ -669,7 +669,7 @@ suite "Waku rln relay":
 
 
     check:
-      msgValidate1 == MessageValidationResult.Valid
+      msg_Validate1 == MessageValidationResult.Valid
```
s1fr0, line 672: Snake case was intentional?

s1fr0 (approved):

tests/v2/test_waku_rln_relay.nim
staheri14: I was testing `styleCheck` compiler option, and forgot to rever it back, fixed now https://github.com/status-im/nwaku/pull/941/commits/de6aa8ff4dd7a8cac29972ad1c538681da8b6bea

tests/v2/test_waku_rln_relay.nim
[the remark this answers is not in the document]
staheri14: I see, I think it is an implicit agreement not explicit, all the rln-relay tests use one-line checks to assert a single condition.  I have split them now https://github.com/status-im/nwaku/pull/941/commits/f9417a30cab96e0ec4b035a434aaa56b42081a54, though it would make sense to consolidate all such internal and implicit conventions to make sure everyone is on the same page.

PR #943, opened 2022-04-12
WIP: Safe timestamps arithmetic

s1fr0:
This PR implements a possible mitigation to https://github.com/status-im/nim-waku/issues/939 in order to have a timestamp arithmetic which is safe from under and over-flows.

### DIFF
--- a/tests/all_tests_v2.nim
+++ b/tests/all_tests_v2.nim
@@ -23,7 +23,8 @@ import
   ./v2/test_waku_store_queue,
   ./v2/test_pagination_utils,
   ./v2/test_peer_exchange,
-  ./v2/test_waku_noise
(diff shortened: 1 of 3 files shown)


oskarth:
What's the state of this PR? If it isn't being worked on I suggest we close it

cc @jm-clius when back

jm-clius:
@s1fr0 IIRC we were unsure if this condition can be triggered by corrupt/malicious _external_ input (receiver timestamps as used here are generated internally)? I think if this can only be caused by internal memory corruption of receiver timestamps (with malicious sender timestamp injections) we can close without merging. Otherwise if we can show an external input can trigger the under/overflow condition we should merge the safe operators.

jm-clius:
Thanks, @s1fr0 :)

PR #945, opened 2022-04-15
fix(Jenkins): remove caching

s1fr0:
Since version `2.343`, Jenkins implements a [security fix](https://issues.jenkins.io/browse/JENKINS-67173) which breaks the [Job Cacher](https://plugins.jenkins.io/jobcacher/) plugin used to cache dependencies like the Nim compiler.

In the meantime that the Job Cacher plugins is updated to address this fix, or other alternatives are found, this PR disables implemented Jenkins caching in `nim-waku`.

Related issues are https://github.com/status-im/status-desktop/pull/5560 and https://github.com/status-im/nimbus-eth2/pull/3594

### DIFF
--- a/ci/Jenkinsfile.prs
+++ b/ci/Jenkinsfile.prs
@@ -43,8 +43,8 @@ pipeline {
   stages {
     stage('Deps') { steps { script {
       /* Avoid checking multiple times. */
-      v1changed = true //versionWasChanged('v1')
-      v2changed = true //versionWasChanged('v2')
+      v1changed = versionWasChanged('v1')
+      v2changed = versionWasChanged('v2')
       /* TODO: Re-add caching of Nim compiler. */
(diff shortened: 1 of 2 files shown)


jakubgs (commented):
Looks good, but why are there so many commits in this branch?

s1fr0:
Thanks @jakubgs! As soon as I found Jenkins was not working properly, I created this branch to do some tests in order to identify the reason. Then, after talking with you, I implemented the right fix :)

jakubgs (approved):
Squash it to one commit and clean it up so it's not full of noise.

jm-clius (approved):
Thanks. Using the Github `Squash and merge` option should work for commit history clean-up.

s1fr0:
@jakubgs @jm-clius sure, as we usually do! I'll merge then! Thanks!

Issue #914, opened 2022-03-24
store efficiency: add index on receiverTimestamp + vacuum on delete


kaiserd:
Extending on the efficiency improvements introduced by #900, further easy-to-implement efficiency upgrades are:

* adding an index to `receiverTimestamp`
* using [vacuum](https://www.sqlite.org/lang_vacuum.html) after batch delete operations

## Background

Expsensive sorting queries in `waku_message_store` sort by `receiverTimestamp`, but there is no index on `receiverTimestamp`.

The primary key of the database consists of (`senderTimestamp`, `id` ,`pubsubTopic`).
This is because the `senderTimestamp` is unique per message, while various `receiverTimestamp`s can exist for the same message on different nodes.
However, ordering is done by `receiverTimestamp` on a given node, because the `senderTimestamp` might not be set by the sender.

So, an easy solution for now is keeping the primary key as it is, but adding an index to `receiverTimestamp`, which improves lookup from O(n) to O(log(n)).
It comes at the cost of a bit of additional storage and takes 2 times longer compared to a primary key lookup. But it is a quick-to-implement improvement of the current state.

In addition to the index, adding vacuum after the batch delete introduced by #900 will keep the DB file compact.
It is a very quick to add improvement.

Both suggested efficiency upgrades do not effect the API of the store.

## Acceptance criteria

- [x] index on `receiverTimestamp` #947
- [x] vacuum on delete #928

cc @jm-clius @jakubgs

PR #947, opened 2022-04-19
perf(store): add index on receiverTimestamp

kaiserd:
This PR addresses the missing part of #914 adding an index on `receiverTimestamp`.

cc @jakubgs


jakubgs (approved):
Sensible.

PR #958, opened 2022-05-09
Disable GossipSub PX by default

jm-clius:
This PR disables relay (gossipsub) peer exchange by default.

### Why?

After a couple of discussions, including with @Menduist, concluded that there are too many security risks in a system whereby a peer has the _default_ ability to inform pruned peers about all its own connected peers.
Furthermore, without proper peer management and a standardised discovery interface, we always attempt connection to _all_ peers received via peer-exchange, resulting in a large amount of churn while stabilising the mesh.

Note that https://github.com/status-im/nim-libp2p/pull/715 will disable peer exchange by default in the library as well.


kaiserd (approved):
LGTM. This is in line with the discussion in the new ambient peer discovery post and our privacy/anonymity goals:).
In the future, we could think about providing a peer list to pruned peers which is not part of the actively used peers... but that would come with changes to PX and might be a new protocol.

staheri14 (approved):
👍

Issue #925, opened 2022-03-30
Makefile: `NIM_PARAMS` override


kaiserd:
This issue aims to solve a generalization of the problem pointed out in #921.

## Motivation

Makefile variables can be [overridden](https://www.gnu.org/software/make/manual/make.html#Overriding) by command line arguments.
[The Dockerfile](https://github.com/status-im/nim-waku/blob/d0cf3ed1f9299618bbab920c66ee06372f24c1fd/Dockerfile#L21), for instance, overrides `NIM_PARAMS`.

However, some of the parameters added to `NIM_PARAMS` should be fixed for `nwaku`.
These are especially parameters like `-d:discv5_protocol_id:d5waku` which set values that are generic in a vendor module (in this case [nim-eth/discv5](https://github.com/status-im/nim-eth/tree/master/eth/p2p/discoveryv5)) but fixed specified by a Waku spec (in this case [33/WAKU2-DISCV5](https://rfc.vac.dev/spec/33/)).

Parameters of this kind have to be manually set by scripts like the Dockerfile.
Overriding `NIM_PARAMS` but not setting important parameters might lead to confusing hard-to-find errors.

## Suggested solution

Add a new variable holding fixed parameters to the Makefile. This variable would be marked with the  [`override` directive](https://www.gnu.org/software/make/manual/make.html#Override-Directive).

The name could be `NIM_PARAMS_CONST`.

### Alternatives

* document the behaviour but don't introduce changes

## Discussion

Wdyt? I would prefer adding the new variable.
I will implement the preferred solution as soon as I get feedback.

cc @staheri14 @jakubgs @richard-ramos @jm-clius

jm-clius:
I think the issue is that we've (incorrectly) used `NIM_PARAMS` outside the Makefile, where we should have used `NIMFLAGS`: https://github.com/status-im/nimbus-build-system#nimflags. Properly differentiating between `NIMFLAGS` (overrideable) and `NIM_PARAMS` (non-overrideable) is more or less your suggested solution.

kaiserd:
@jm-clius Yes :). My preferred solution would be having two separate Variables.

Should we now change the external uses of `NIM_PARAMS` to `NIM_FLAGS`?
I would also suggest adding the `override` directive to `NIM_PARAMS` then.

@jakubgs wdyt?

[...]

PR #961, opened 2022-05-11
chore(docker): use NIMFLAGS instead of NIM_PARAMS

kaiserd:
This PR addresses #925.
Also enables the `d5waku` protocol ID for discv5 on fleets without additional config.


jakubgs:
This was built in https://ci.status.im/job/nim-waku/job/deploy-status-test/9/ and I deplyed it on `node-01.do-ams3.wakuv2.test` but:
```
INF 2022-05-11 21:36:27.177+00:00 WakuRLNRelay is enabled                    topics="wakunode.setup" tid=1 file=wakunode2.nim:1196
DBG 2022-05-11 21:36:27.177+00:00 rln-relay membership index                 topics="wakurlnrelayutils" tid=1 file=waku_rln_relay_utils.nim:328 rlnRelayMemIndex=0
/app/waku/v2/node/wakunode2.nim(1341) wakunode2
/app/waku/v2/node/wakunode2.nim(1204) setupProtocols
/app/waku/v2/node/wakunode2.nim(541) mountRlnRelay
/app/vendor/nim-chronos/chronos/asyncfutures2.nim(365) futureContinue
/app/waku/v2/node/wakunode2.nim(603) mountRlnRelay
/app/waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim(44) createRLNInstance
/app/vendor/nimbus-build-system/vendor/Nim/lib/system/io.nim(693) readFile
[[reraised from:
/app/waku/v2/node/wakunode2.nim(1341) wakunode2
/app/waku/v2/node/wakunode2.nim(1204) setupProtocols
]]
Error: unhandled exception: cannot open: waku/v2/protocol/waku_rln_relay/parameters.key [IOError]
```
Seems broken.

kaiserd:
Ill address the RLN part tomorrow. Thanks for the help!

kaiserd:
I think I fixed the issue. Still have to test it.
Here is a rough description (I can provide more detail if desired):

The file pointed to by the error message
`waku/v2/protocol/waku_rln_relay/parameters.key` was not copied into the container.
I did not consider this to be the issue, because without this PR the image seemed to work.
However, with overriding `NIM_PARAMS`, `-d:rln` was not set even with `RLN=true`.
`RLN=true` would normally trigger `-d:rln` to be set, but not when overriding `NIM_PARAMS` from outside (as happened here).
So, the rln code requiring this file was never executed, which silently seemed to work.

cc @staheri14

kaiserd:
I also added `RLN=true` to the deps make rule call, because `RLN=true` triggers options there, too, which also can be cached.

jakubgs (approved):
Nice, I tested the image on one host and it appears to work fine:
```
admin@node-01.do-ams3.wakuv2.test:/docker/nim-waku-v2 % d
CONTAINER ID   NAMES         IMAGE                                                 CREATED         STATUS
3dcf250e65e8   websockify    statusteam/websockify:v0.10.0                         2 minutes ago   Up 2 minutes
d0a9f7f2cfce   nim-waku-v2   statusteam/nim-waku:deploy-status-test-nimflag-test   2 minutes ago   Up 2 minutes (healthy)
```
Well done.

kaiserd:
Thank you very much @jakubgs !

staheri14:
@kaiserd thanks for resolving the issue! it might also explain why despite mounting the waku-rln-relay protocol on the test fleets, the interaction was showing otherwise.

kaiserd:
@staheri14 Yes :). RLN should work properly on the fleet nodes after this PR gets deployed. In my local tests, the RLN protocol was mounted after applying these changes, but without, it was not mounted becasue `-d:rln` was not set.

staheri14:
@kaiserd curious to know whether it was finally needed to use the `override` directive for the `NIM_PARAMS` of the Makefile?

kaiserd:
@staheri14 
> @kaiserd curious to know whether it was finally needed to use the override directive for the NIM_PARAMS of the Makefile?

`override` is not needed and this PR does not introduce it, but `override` could be used on `NIM_PARAMS` to enforce that nothing from the outside overrides it.

> Update: I think the answer lies in this comment #925 (comment)?

Yes. With this PR, the Jenkins pipeline uses `NIMFLAGS` instead of `NIM_PARAMS` to add compiler flags. This is now in line with the nimbus build system doc.
This pattern of having a variable for internal use and one for external use makes sense, because make does not allow adding to a variable from outside the Makefile.

jm-clius (approved):
Thanks for this, @kaiserd. Will save a lot of future headaches.

Issue #964, opened 2022-05-14
--dns4-domain-name can lead to an unhandeled exception


kaiserd:
Using the  `--dns4-domain-name` cmd option can cause an unhandeled exception: 

```
waku_1  | /app/waku/v2/node/wakunode2.nim(1266) wakunode2
waku_1  | /app/waku/v2/node/wakunode2.nim(1063) initNode
waku_1  | /app/waku/v2/node/wakunode2.nim(195) new
waku_1  | /app/vendor/nimbus-build-system/vendor/Nim/lib/pure/options.nim(185) get
waku_1  | Error: unhandled exception: Can't obtain a value from a `none` [UnpackError]
```

(This issue was reported by filoozom in Discord)

kaiserd:
Fixed in #965.

PR #965, opened 2022-05-14
fix(dns4name): dns4name + unset port resolved

kaiserd:
This PR solves #964

`nwaku` assumed that when `--dns4-domain-name` is set, an external port is available.
This is not the case when `setupNat` cannot establish a port mapping.
The solution in this PR is simple: in this case, the dns domain name is treated as an external IP address, and port mapping is assumed to be manually resolved.


jm-clius (approved):
LGTM. Thanks! Good catch.

oskarth (approved):

Issue #969, opened 2022-05-16
Add --version flag


dao:
It can be difficult to recall which version of nwaku you're running where. A `--version` flag would make it easy to check.

jm-clius:
@kaiserd would the `get_waku_v2_debug_v1_version()` RPC call be useful here? It seems as if it's not added to RFC16 yet?

kaiserd:
@jm-clius yes. I could add a `--version` flag that prints the same info as the RPC call. And yes, RFC16 has not been updated. Can do that, too.

dao:
This RPC does indeed return "v0.9". This, however, is of little help given that all changes since march 31st are still considered v0.9, with no finer grained version numbers in between. This would be ok if you could return the commit hash as well

kaiserd:
But as you requested, having a `--version` flag is nice to have, too :).

kaiserd:
See #970

Example:

```
./build/wakunode2 --version
version / git commit hash: 062668
```

It will print the version, if the current commit is tagged.

kaiserd:
@dao When testing the RPC call, did you compile the current master branch? It should return the commit hash in this case.
Could it be that you actually have version v0.9? It only prints the version, if the current commit hash is tagged with it.

dao:
> @dao When testing the RPC call, did you compile the current master branch? It should return the commit hash in this case. Could it be that you actually have version v0.9? It only prints the version, if the current commit hash is tagged with it.

Ah ok yes, I must have v0.9 release. Thanks!

kaiserd:
Also added the specification for the version call in RFC 16: https://github.com/vacp2p/rfc/pull/505

kaiserd:
Added feature in #970

PR #970, opened 2022-05-16
feat(info): add --version cmd flag

kaiserd:
Solves #969 in a minimal way.

Prints only the commit hash, if the current version has not been tagged with a version tag.
(That is true for the json RPC call, too.)
If desired, we could add the last version on the current branch as additional info. This would require a new compiler flag in the Makefile.
This PR just prints the info that is available via JSON rpc at the moment.

Example output for untagged version:

```
./build/wakunode2 --version
version / git commit hash: 062668
```


jm-clius (approved):
Neat! Thanks.

kaiserd:
resolved merge conflicts

PR #974, opened 2022-05-18
chore: update submodules

jm-clius:
This PR updates submodules.

The main purpose is to include [a high-priority fix in nim-websock](https://github.com/status-im/nim-websock/pull/109) that causes slow performance on the prod fleet: https://github.com/status-im/infra-nim-waku/issues/50.

Although `nim-websock` is the target, we update all submodules in tandem to avoid compatibility issues between submodules.

### DIFF
--- a/tests/v2/test_waku_keepalive.nim
+++ b/tests/v2/test_waku_keepalive.nim
@@ -48,6 +48,5 @@ procSuite "Waku Keepalive":
     check:
       (await completionFut.withTimeout(5.seconds)) == true
 
-    # await allFutures([node1.stop(), node2.stop()])
     await node2.stop()
     await node1.stop()
--- a/tests/v2/test_waku_keepalive.nim
+++ b/tests/v2/test_waku_keepalive.nim
@@ -48,4 +48,6 @@ procSuite "Waku Keepalive":
     check:
       (await completionFut.withTimeout(5.seconds)) == true
 
-    await allFutures([node1.stop(), node2.stop()])
+    # await allFutures([node1.stop(), node2.stop()])
+    await node2.stop()
+    await node1.stop()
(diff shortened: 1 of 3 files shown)


Menduist:
The CI of libp2p is failing with the new version of nim-websock, need to investigate..

Menduist:
Green in libp2p

tests/v2/test_waku_keepalive.nim
```diff
@@ -48,4 +48,6 @@ procSuite "Waku Keepalive":
     check:
       (await completionFut.withTimeout(5.seconds)) == true
 
-    await allFutures([node1.stop(), node2.stop()])
+    # await allFutures([node1.stop(), node2.stop()])
```
LNSD, line 51: Is there any reason for keeping this code commented? Just asking.

LNSD (approved):
Just a little comment about "commented but not used code".

> From my experience, if you want actually to keep commented code, add a comment explaining why that commented code should not be deleted. Otherwise, just remove it (you can always recover it from git history).

The rest, LGTM ✅

tests/v2/test_waku_keepalive.nim
jm-clius: Good question. No, I commented it out as an experiment and forgot to remove when the experiment worked. :) Will address subsequently, as I don't want the CI chain to kick off again before I can merge. :)

jm-clius: And you're right - we shouldn't have dead code and legacy comments cluttering the code.

LNSD: It happens to me often, too 😆

kaiserd (approved):
LGTM!

PR #975, opened 2022-05-20
feat(store): ability to decouple store from relay

### LINKED ISSUE #937: Store: ability to decouple `store` from `relay`
## Background

Currently the `store` and `relay` implementations are tightly coupled in the `nwaku` implementation. Although it is possible to run a `relay` node without `store` capabilities, a `store` node has to operate as a `relay` node subscribed to one or more `pubsubTopic`s to store historical messages for those `pubsubTopic`s.

For certain applications this has at least two drawbacks:
1. Each `store` node needs to be a well-connected `relay` node with accompanying bandwidth and connectivity requirements, although its primary function is simply to _receive_ and store messages.
2. `store` nodes cannot be selective about the `contentTopic`s (or applications) for which they're willing to store messages. A `store` node will store all messages for all `pubsubTopic`s that it participates in, with no ability to express interest only in certain applications.

A solution for the above would be to implement a `store` that could operate only as a "listener" and historical store for messages on specific `pubsubTopic`s/`contentTopic`s without having to relay messages.

## Suggested solution

A fairly straightforward way to achieve the above functionality is to have the `store` implementation interact with the `filter` implementation. A `store` node could then choose to act as a `filter` client and subscribe only to receive message from those `pubsubTopic`s and `contentTopic`s it's interested in storing.

Implications of such a solution:
1. Capability discovery, especially negotiating with a `store` node in terms of what historical messages it is storing, would need to bear in mind that `store` nodes can now also be hyper-selective in terms of which content topics they're storing.
2. `filter` protocol currently fails silently. A keep-alive and failsafe mechanism (perhaps redundant connections to other `filter` nodes) may be required as `store` nodes will now rely solely on short-lived connections to `filter` nodes to receive messages, which may be less reliable than operating in a well-connected `relay` mesh.

cc @sbc64 @D4nte @staheri14

LNSD:
This PR resolves #937 

Now, nodes can combine store protocol either with the relay protocol or with filter protocol. Having the relay mounted is no longer a requirement to support store protocol.

### DIFF
--- a/tests/v2/test_wakunode.nim
+++ b/tests/v2/test_wakunode.nim
@@ -349,6 +349,65 @@ procSuite "WakuNode":
     await node1.stop()
     await node2.stop()
 
+  asyncTest "Store protocol returns expected message when relay is disabled and filter enabled":
+    # See nwaku issue #937: 'Store: ability to decouple store from relay'
+
+    let
+      nodeKey1 = crypto.PrivateKey.random(Secp256k1, rng[])[]
+      node1 = WakuNode.new(nodeKey1, ValidIpAddress.init("0.0.0.0"), Port(60000))
+      nodeKey2 = crypto.PrivateKey.random(Secp256k1, rng[])[]
+      node2 = WakuNode.new(nodeKey2, ValidIpAddress.init("0.0.0.0"), Port(60002))
+
+      pubSubTopic = "/waku/2/default-waku/proto"
+      contentTopic = ContentTopic("/waku/2/default-content/proto")
+      message = WakuMessage(payload: "hello world".toBytes(), contentTopic: contentTopic)
+
+      filterComplFut = newFuture[bool]()
+      storeComplFut = newFuture[bool]()
+
+    await node1.start()
+    node1.mountStore(persistMessages = true)
+    node1.mountFilter()
+
+    await node2.start()
+    node2.mountStore(persistMessages = true)
+    node2.mountFilter()
+
+    node2.wakuFilter.setPeer(node1.switch.peerInfo.toRemotePeerInfo())
+    node1.wakuStore.setPeer(node2.switch.peerInfo.toRemotePeerInfo())
+
+    proc filterReqHandler(msg: WakuMessage) {.gcsafe, closure.} =
+      check:
+        msg == message
+      filterComplFut.complete(true)
+
+    await node2.subscribe(FilterRequest(pubSubTopic: pubSubTopic, contentFilters: @[ContentFilter(contentTopic: contentTopic)], subscribe: true), filterReqHandler)
+
+    await sleepAsync(2000.millis)
+
+    # Send filter push message to node2
+    await node1.wakuFilter.handleMessage(pubSubTopic, message)
+
+    await sleepAsync(2000.millis)
+
+    # Wait for the node2 filter to receive the push message
+    check:
+      (await filterComplFut.withTimeout(5.seconds)) == true
+
+    proc node1StoreQueryRespHandler(response: HistoryResponse) {.gcsafe, closure.} =
+      check:
+        response.messages.len == 1
+        response.messages[0] == message
+      storeComplFut.complete(true)
+
+    await node1.query(HistoryQuery(contentFilters: @[HistoryContentFilter(contentTopic: contentTopic)]), node1StoreQueryRespHandler)
+
+    check:
+      (await storeComplFut.withTimeout(5.seconds)) == true
+
+    await node1.stop()
+    await node2.stop()
+
   asyncTest "Messages are correctly relayed":
     let
       nodeKey1 = crypto.PrivateKey.random(Secp256k1, rng[])[]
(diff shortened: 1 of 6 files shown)


kaiserd:
Congratulations on your first `nwaku` PR :).

kaiserd (changes requested):
There is an error because of ambiguous types:

```
  wakunode2.Topic: Topic

  waku_filter_types.Topic: Topic
```

waku/v2/protocol/waku_filter/waku_filter_types.nim
```diff
@@ -15,21 +15,24 @@ const
   MaxRpcSize* = 10 * MaxWakuMessageSize + 64*1024
 
 type
+  Topic* = string
```
kaiserd, line 18: This type causes a name colision (see my comment above).
Maybe call it PubSubTopic to make clear it is not the ContentTopic.

Wakunode2.nim and filter could use the same type. You could  use the type declared here in filter also in wakunode2.

Also, if you want to make sure it is not any kind of string but this type in specific, you coud use

```nim
type
  PubSubTopic* = distinct string
```

LNSD: I agree that we would need to have:
* either a shared `PubSubTopic` and `ContentTopic` type aliases between all modules
* or a "module-level" type aliases and one at the waku node module level that are compatible but not interfering

jm-clius: WDYT of adding this type (`PubSubTopic`) export to the `waku_message` module? Although not strictly part of the message, it is so related in concept that I'd be comfortable having it there. This is also where we export `ContentTopic*` which would make the difference and similarity a bit more apparent and it already serves as the "common types" module between all protocols. We could even consider renaming the module to common_types or something, but I don't feel that's necessary.

tests/v2/test_wakunode.nim
```diff
@@ -349,6 +349,65 @@ procSuite "WakuNode":
     await node1.stop()
     await node2.stop()
 
+  asyncTest "Store protocol returns expected message when relay is disabled and filter enabled":
+    # See nwaku issue #937: 'Store: ability to decouple store from relay'
```
jm-clius, line 353: I'm fine with fully qualified links in code comments, btw. :)

tests/v2/test_wakunode.nim
```diff
@@ -1469,10 +1528,10 @@ procSuite "WakuNode":
       # Check that underlying peer info contains only bindIp before starting
       node.switch.peerInfo.addrs.len == 1
       node.switch.peerInfo.addrs.contains(bindEndpoint)
-      
+
       node.announcedAddresses.len == 1
       node.announcedAddresses.contains(announcedEndpoint)
-        
+
```
jm-clius, line 1534: No need to change anything for this PR, but as a general comment: a large number of whitespace-only changes (often caused by prettifying plugins or editors) can make a PR a bit difficult to review and "hide" the changes that are relevant to the issue. Formatting changes are often helpful and necessary, but in future I'd suggest opening a separate PR for formatting only if they become a significant portion of the total changes.

waku/v2/node/wakunode2.nim
```diff
@@ -457,12 +457,16 @@ proc info*(node: WakuNode): WakuInfo =
...
   proc filterHandler(requestId: string, msg: MessagePush)
-    {.gcsafe, raises: [Defect, KeyError].} =
+    {.async, gcsafe, raises: [Defect, KeyError].} =
     
     info "push received"
     for message in msg.messages:
       node.filters.notify(message, requestId) # Trigger filter handlers on a light node
-      
+
+      if not node.wakuStore.isNil and (requestId in node.filters):
```
jm-clius, line 466: Perhaps a very brief comment to explain what this check does. Something like

```
# Ensures `store` is mounted and that we have an active subscription on this `requestId`
```

jm-clius (changes requested):
This looks good! Congratulations on your first PR. :) In fact, I'm ready to approve, but will wait for the changes necessary to address the `Topic` type clash. Please "re-request" review when you've addressed these.

tests/v2/test_wakunode.nim
LNSD: Oh, that was unintentional. I agree, that only adds unnecessary noise. I have to disable the "remove trailing spaces" option from the IDE.

waku/v2/protocol/waku_filter/waku_filter_types.nim
LNSD: We need to give it a thought. Maybe I can tackle it as part of my next PR, and live it for now as @kaiserd suggested.

jm-clius: Sounds reasonable to me.

kaiserd: I agree on adding PubSubTopic to `waku_message`  in a follow up PR. My first intuition was searching the type in waku_message :)

jm-clius:
Seems like CI passes. Remember to click on re-request review if you think it's ready.

jm-clius (approved):
LGTM! When you have one more approving review, feel free to "Squash and Merge". Ensure the single commit message follows semantic commit style, but it seems you have done this already. :)

Issue #894, opened 2022-03-15
Dynamic rln-relay groups with local tx management (no wallet)


staheri14:
# Problem
This issue details the first step towards adding support for dynamic group management in rln-relay (aligned with https://github.com/status-im/nim-waku/issues/827). 
Everything will be tested against a test Eth network e.g., Ropsten.  Since nodes will transact on a test network where no real ether and fun is involved, we can safely delay the wallet integration in the favor of doing more focused work on contract event listening (a bit of diversion from the initial plan set out in #827). We assume rln-relay nodes will pass their Eth private keys for the Ropsten test network to the wakunode. The nim-waku node handles the registration and deletion transactions. Part of this logic i.e., tx management from a nim-waku node using nim-web3 is already implemented, so this allows us to concentrate on the event listening component.

The goal of this step is 1) to replace the hardcoded membership tree with a membership contract 2) to implement the necessary logic that enables a nim-waku node to listen to the events emitted from a membership contract. 

# Acceptance Criteria
- [x] Nodes running rln-relay are able to fetch the latest state of the contract and construct their tree
- [x] Nodes running rln-relay are able to listen to the contract and dynamically update their membership tree locally

# Next
In the next phase, the wallet integration will be investigated i.e., https://github.com/status-im/nim-waku/issues/819

cc: @oskarth

jm-clius:
@staheri14 seems like all related tasks are done. Can this be closed?

staheri14:
@jm-clius Right, going to close it.

PR #976, opened 2022-05-20
Feat (Rln relay): update membership tree dynamically

staheri14:
This PR is in line with https://github.com/status-im/nwaku/issues/894. It introduces the utility functions to aid an rln-relay peer in capturing the registration events emitted from the membership contract and reflecting them on the membership Merkle tree.


waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
```diff
@@ -21,8 +21,12 @@ type SpamHandler* = proc(wakuMessage: WakuMessage): void {.gcsafe, closure,
 
 # membership contract interface
 contract(MembershipContract):
-  # TODO define a return type of bool for register method to signify a successful registration
   proc register(pubkey: Uint256) # external payable
+  proc MemberRegistered(pubkey: Uint256, index: Uint256) {.event.}
```
jm-clius, line 25:
```suggestion
  proc memberRegistered(pubkey: Uint256, index: Uint256) {.event.}
```

jm-clius (approved):
LGTM

oskarth (approved):

waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
staheri14: Thanks for the suggestion, though, it is not possible to change it as it needs to exactly match the name of the corresponding even on the contract. The reason is that when communicating with the contract the hash of the corresponding event i.e., `MemberRegistered` will be used to filter the transaction logs.

Issue #950, opened 2022-05-02
store: answer history queries from sqlite DB


kaiserd:
## Background

While `nwaku` features a sqlite3 database for persisting messages,
history queries are only answered from the in-memory store.
When starting a `nwaku` node, `--store-capacity` messages are loaded from the  sqlite3 DB into the in-memory database.

We also limited the capacity limit for the persistent storage in #900, to avoid overly long starting times for `nwaku` nodes.

## Goal / Motivation

For some applications, storing more messages is desired. For instance, keeping a history of one month of messages.
To achieve this, a first simple solution is introducing a (large capacity) sqlite3-only store.

A follow up issue will track a feasibility test of the sqlite3-only store.
(Performance might not be sufficient.)

## Acceptance Criteria

* [x] sqlite3-only store implemented
* [x] tests added

## Future Work

* generic store interface

jakubgs:
We seriously need this done. Right now it takes ~17 minutes for a node with `684365` messages in the DB to start:
```
INF 2022-05-23 17:37:46.236+00:00 attempting to load messages from persistent storage topics="wakustore" tid=1 file=waku_store.nim:370
INF 2022-05-23 17:53:53.236+00:00 successfully loaded from store             topics="wakustore" tid=1 file=waku_store.nim:377
```
And it appears it's fully utilizing the available disk I/O while doing it:

![image](https://user-images.githubusercontent.com/2212681/169879316-ee0a197b-839d-4d0c-9cd3-ea9c00e76502.png)

jakubgs:
I see the issue, it's because it's SWAPing due to container memory limits, which bumps the disk usage further:

![image](https://user-images.githubusercontent.com/2212681/169880302-fd69144c-bd23-47cb-9c4c-2d248a3d6e21.png)

This is not sustainable for any kind of large history.

PR #982, opened 2022-05-29
feat(store): sqlite only store

kaiserd:
This PR addresses #950.

(I am still working on performance improvments, but the functionality is ready for review.)

cc @jakubgs

# info

I added both a new test suite that tests the sql store in isolation, and made the waku_store test compatible [f5d9e4822](https://github.com/status-im/nwaku/commit/f5d9e482271fd177471e5c32deae2048d5370c26).
Let's discuss if we want to include that, too. I ran the sql-store against these tests, too, and it worked.

## efficiency issue

I tested the sql only store on a droplet fleet with a large DB (1.6GB, >800k messages).
Connecting with chat2 takes a long a time, especially for backward queries.

The DB does not contain many messages with the chat2 content topic. So, the respective query has a sparse filter
which causes `getPage` to retrieve and filter a lot of DB pages in order to fill the desired filtered return page.
Also, store nodes block when answering queries.

Steps:
* further investigation / testing
* make filtering more efficient (see todo items)
* don't block
* mitigation: send more efficient query (e.g. smaller page sizes)

Further efficiency improvements:

* DESC index (thanks @richard-ramos)

Still, there is DoS potential with this.

# todo
* [x] merge #981 (once this is merged) and adapt accordingly
* [x] testing
* [x] retention time: clean up old messages after a specified time (e.g. one month)
* ~~make `msg` handler proc more efficient (don't recalculate the index)~~ this is not a priority
* [ ] fix efficiency issues with spare queries (the item above is one part of that)
    - [x] partially addressed by implementing adaptive DB page sizing
* [ ] more testing

# future work (follow up PRs)

* [ ] more investigation regarding efficiency (caching, etc....)
* [ ] investigate and fix blocking while handling a store query
* ~~clean up old messages after a specified time (e.g. one month)~~ this took prio (see todo list above), first version already implemented
* [ ] new store interface
* [ ] reuse getPage related prepared statements beyond a getPage call (see discussion below)
* [ ] memory cache
* [ ] ...

## testing frame work

After introducing the store interface, it would be nice to have a test frame work that can automatically run the tests against various implementations of the interface. (Similar to dependency injection based testing in Java.)

### DIFF
--- a/waku/v2/node/config.nim
+++ b/waku/v2/node/config.nim
@@ -160,6 +160,16 @@ type
       defaultValue: 50000
       name: "store-capacity" }: int
     
+    sqliteStore* {.
+      desc: "Enable sqlite-only store: true|false",
+      defaultValue: false
+      name: "sqlite-store" }: bool
+
+    sqliteRetentionTime* {.
+      desc: "time the sqlite-only store keeps messages (in seconds)",
+      defaultValue: 30.days.seconds
+      name: "sqlite-retention-time" }: int64 # TODO: Duration
+
     ## Filter config
 
     filter* {.
--- a/waku/v2/node/storage/message/waku_message_store.nim
+++ b/waku/v2/node/storage/message/waku_message_store.nim
@@ -1,13 +1,15 @@
 {.push raises: [Defect].}
 
 import 
-  std/[options, tables],
+  std/[options, tables, times],
   sqlite3_abi,
   stew/[byteutils, results],
   chronicles,
+  chronos,
   ./message_store,
   ../sqlite,
   ../../../protocol/waku_message,
+  ../../../protocol/waku_store/waku_store,
   ../../../utils/pagination,
   ../../../utils/time
 
@@ -37,6 +39,9 @@ type
     storeCapacity: int # represents both the number of messages that are persisted in the sqlite DB (excl. the overflow window explained above), and the number of messages that get loaded via `getAll`.
     storeMaxLoad: int  # = storeCapacity * MaxStoreOverflow
     deleteWindow: int  # = (storeCapacity * MaxStoreOverflow - storeCapacity)/2; half of the overflow window, the amount of messages deleted when overflow occurs
+    isSqliteOnly: bool
+    retentionTime: chronos.Duration
+    oldestReceiverTimestamp: int64
     insertStmt: SqliteStmt[(seq[byte], Timestamp, seq[byte], seq[byte], seq[byte], int64, Timestamp), void]
  
 proc messageCount(db: SqliteDatabase): MessageStoreResult[int64] =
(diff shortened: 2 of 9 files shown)


jm-clius:
Good to see this progress. Thanks! Fix for memory leak issue now open for review.
You're aware of this, but it may be a good idea to cache and reuse frequently used prepared statements to lower the amount of memory churn.

kaiserd:
I added dispose and some error handling.
So far, I don't store the prepared paging statement yet.
Currently there are 4 kinds of statements for paging:

* forward without index
* forward with index
* backward without index
* backward with index

I will look into condensing that and then store them in the store object.
Maybe the savings of this are not so big tough. There are at most 2 statements per paging query. Iteration within a `getPage` request will not allocate new statements. Further, paging queries will not be done in parallel, which keeps memory usage at just 2 statements. Wdyt @jm-clius ?
I will still look into that :), but first do some basic efficiency testing with large DBs.

jm-clius:
> I will look into condensing that and then store them in the store object. Maybe the savings of this are not so big though. There are at most 2 statements per paging query. Iteration within a `getPage` request will not allocate new statements. Further, paging queries will not be done in parallel, which keeps memory usage at just 2 statements. Wdyt @jm-clius ? 

I agree. Even relatively frequent store queries are not nearly as often used as, say, the `insert` statement for new messages. I don't think it's a high priority to cache the associated statements.

> I will still look into that :), but first do some basic efficiency testing with large DBs.

I agree.

waku/v2/node/storage/message/waku_message_store.nim
```diff
@@ -51,6 +54,36 @@ proc messageCount(db: SqliteDatabase): MessageStoreResult[int64] =
     return err("failed to count number of messages in DB")
   ok(numMessages)
 
+
+proc getOldestDbReceiverTimestamp(db: SqliteDatabase): MessageStoreResult[int64] =
+  var oldestReceiverTimestamp: int64
+  proc handler(s: ptr sqlite3_stmt) =
+    oldestReceiverTimestamp = column_timestamp(s, 0)
+  let query = "SELECT receiverTimestamp FROM " & TABLE_TITLE & " ORDER BY receiverTimestamp LIMIT 1;";
```
richard-ramos, line 62: Is there a difference in performance between SELECT ... ORDER BY ... LIMIT 1 and: 
```suggestion
  let query = "SELECT MIN(receiverTimestamp) as minReceiverTimestamp FROM " & TABLE_TITLE;
```

kaiserd: `MIN` is more efficient here. I  uses the primary key for searching. Thank you :). And its also more readable.

Ill do 

```nim
let query = "SELECT MIN(receiverTimestamp)  FROM " & TABLE_TITLE;
```

For now, I just wanted the retention timestamp feature to work. There is quite a lot in the code that could be improved / made more efficient  :sweat_smile: 
The goal is to have the feature in the next release, so that it can be tested to get earlier feedback.

kaiserd:
There are still quite a few points that could be improved / condensed / made more elegant... But because this PR should be part of version 0.10 as an experimental new feature, I opened this PR for review. Imo, the SQL-only store with retention time feature is ready for field tests after the reviews :).
I will work on efficiency improvements in parallel.

waku/v2/node/config.nim
```diff
@@ -160,6 +160,16 @@ type
...
     
+    sqliteOnlyStore* {.
+      desc: "Enable sqlite-only store: true|false",
+      defaultValue: false
+      name: "sqlite-only-store" }: bool
+
+    retentionTime* {.
+      desc: "time the sqlite-only store keeps messages (in seconds)",
+      defaultValue: 30.days.seconds
+      name: "retention-time" }: int64 # TODO: Duration
```
jm-clius, line 171: We really need sub-commands :). For now, wdyt about having sqlite-related items be prefixed similarly? Perhaps then the conf items could be: `sqlite-retention-time` and `sqlite-store` (not sure, but imo `only` can be dropped - this is a SQLite implementation of store and the description still clarifies that this is different from the memory store with persistence in SQLite).

jm-clius (approved):
Thanks for this, @kaiserd! IMO this can be merged. My comments below mostly relate to indentation and formatting, which is often a matter of personal preference.

waku/v2/node/storage/message/waku_message_store.nim
```diff
@@ -211,6 +266,156 @@ method getAll*(db: WakuMessageStore, onData: message_store.DataProc): MessageSto
...
+      # TODO: we should consolidate WakuMessage, Index, and IndexedWakuMessage; reason: avoid unnecessary copying and recalculation
+      index = retMsg.computeIndex(receiverTimestamp, pubsubTopic) # TODO: retrieve digest from DB
+      indexedWakuMsg = IndexedWakuMessage(msg: retMsg, index: index, pubsubTopic: pubsubTopic) # TODO: constructing indexedWakuMsg requires unnecessary copying
+
+    lastIndex = index
+    numRecordsVisitedPage += 1
+    try:
+      if pred(indexedWakuMsg): #TODO throws unknown exception
+        numRecordsMatchingPred += 1
+        messages.add(retMsg)
```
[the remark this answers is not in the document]
kaiserd: This is on my todo list :).  It would make the code cleaner as we would not need the adaptive DB pages anymore.
This would, however, limit the expressiveness of the predicate.
If we want to offer an easy to use interface to the store, we have to translate the nim closure into a WHERE compatible predicate. 
For now, I think there are larger performance gains in other areas (e.g. caching), so I postponed this.

waku/v2/node/config.nim
kaiserd: Agreed on both points :). (commit after next call)

oskarth (commented):
Thanks for this and in general looks good to me!

One of the more important thing here is knowledge transfer.

@LNSD are you able to review this PR? See if things makes sense to you.

LNSD (approved):
LGTM ✅

LNSD:
Sorry, @kaiserd, I reviewed it this morning but I forgot to send my approval 😅

jm-clius:
@kaiserd seems like all that's missing here to be merged is to `amend` commits that were not signed?

Issue #622, opened 2021-06-14
Waku Peer Store: Integrated libp2p peer store


jm-clius:
### Background

Waku v2 currently defines its [own peer store and manager](https://github.com/status-im/nim-waku/tree/master/waku/v2/node/peer_manager). `nim-libp2p` has recently [implemented a peer store](https://github.com/status-im/nim-libp2p/commit/93156447ba99fa9f70a29c885c5a1509da3969b0) that replicates much of the functionality provided by the `nwaku` peer store.

### Proposed solution

Waku needs to integrate and use the libp2p peer store in the peer manager, where possible, and extend the store for its self-defined books (such as the `ConnectionBook` and `DisconnectBook` - both `nwaku`-specific).

Some proposals for what the Waku Peer Store should support:
- a generic interface through which peers can be added, although the source can be tracked (e.g. peers added manually vs peers added via one of the discovery methods)
- peer selection mechanism for each protocol (prioritising manually-added peers). Peer scoring will eventually play a role here.
- peer persistence (while providing enough information for the connection manager to respect gossipsub rules here, e.g. the mandatory backoff after a PRUNE before reconnecting)
- proper housekeeping (stale peers which `CannotConnect` should eventually be removed)

This store should be the basis from which a separate **connection manager** can:
- select the best peers to maintain continuously a rich set of `relay` connections
- select peers for service protocols (e.g. `store`, `filter`, `lightpush`), while prioritising peers that were manually added for these protocols (e.g. via `--storenode`, `--filternode`, `--lightpushnode`) options. This could benefit from making these config items repeatable (we currently only support a single service node per protocol).
- update stored peer information with information coming from the connection (e.g. the disconnection time, supported protocols from `identify`, etc.)

jm-clius:
Updating milestone from Release v0.11 to next release (v0.12) to limit scope/risk.

alrevuelta:
As I see it, while the title is "Waku Peer Store" we are talking about two different things:
* `PeerStore`: Stores peers and their information: multiaddr, id, source, connection state, etc. With the capability to persist it, useful across restarts.
* `PeerManager`: Takes care of connecting to new peers (i.e. choosing which one) and disconnecting if max connections were reached (chose which one to disconnect from). Perhaps also cleaning the `PeerStore` and removing stale peers.

## Peer Store
Focusing on `PeerStore`, I would divide this issue into the following subtasks:
[...]

PR #987, opened 2022-05-31
chore: update submodules

jm-clius:
Updates submodules.

Main focus is to include at least two fixes related to websockets:
- Setting activity flag to prevent websocket connections timing out: https://github.com/status-im/nim-libp2p/commit/1696d0c7077fc1d4bec740546a296b3d400f1375
- Several improvements/fixes in nim-websock: https://github.com/status-im/nim-websock/commits/main

Some fixes were required for compatibility with underlying changes in the `PeerStore`. This is still a workaround until we prioritise work to replace our own `WakuPeerStore` with the underlying libp2p `PeerStore` as set out in https://github.com/status-im/nwaku/issues/622

### DIFF
--- a/tests/v2/test_peer_storage.nim
+++ b/tests/v2/test_peer_storage.nim
@@ -19,7 +19,7 @@ suite "Peer Storage":
       peerKey = crypto.PrivateKey.random(ECDSA, rng[]).get()
       peer = PeerInfo.new(peerKey, @[peerLoc])
       peerProto = "/waku/2/default-waku/codec"
-      stored = StoredInfo(peerId: peer.peerId, addrs: toHashSet([peerLoc]), protos: toHashSet([peerProto]), publicKey: peerKey.getPublicKey().tryGet())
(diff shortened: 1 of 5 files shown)


LNSD:
I have one question about #622, do we want to re-prioritize it so we tackle it as part of one of the coming releases?

jm-clius:
> I have one question about #622, @jm-clius, do we want to re-prioritize it so we tackle it as part of one of the coming releases?

It is currently associated with the milestone for v0.11 (mid-August). Do you mean that we should include it in the upcoming release? I think we'd like to keep the scope for v0.10 minimal so we can deliver it in time.

LNSD:
> > I have one question about #622, @jm-clius, do we want to re-prioritize it so we tackle it as part of one of the coming releases?
> 
> It is currently associated with the milestone for v0.11 (mid-August). Do you mean that we should include it in the upcoming release? I think we'd like to keep the scope for v0.10 minimal so we can deliver it in time.

Oh, my bad 🙏🏼 I saw it in the "icebox", but I did not see it as part of the 0.11 milestone. Aug release makes sense, so we do not add unnecessary risk to v0.10 😁

kaiserd (approved):
LGTM!

Issue #727, opened 2021-09-30
HTTP REST API: Debug and Relay APIs


jm-clius:
## Background

`nim-waku` provides a JSON-RPC API for non-Nim clients. The [original plan](https://github.com/status-im/nim-waku/issues/243) was to have this API be REST-like so that it could eventually map to an HTTP REST API, once a secure web server was implemented in Nim.

With the introduction of [`nim-presto`](https://github.com/status-im/nim-presto), this now seems possible. Consequently nimbus-eth2 has [added a REST API](https://github.com/status-im/nimbus-eth2/blob/stable/beacon_chain/rpc/beacon_rest_api.nim). `nim-waku` can follow suit.

## Acceptance Criteria

- An HTTP REST API for Waku v2 translated directly from the existing JSON-RPC REST API.
  + [x] Debug REST API - PRs: 
    - #988 
    - #1007 
  + [x] Relay REST API - PRs: 
    - #1008 
- [x] Integration with Waku V2 node
  + #1022 

In future we also want to add Store, Filter, Admin and Private API functionality for parity with the existing JSON-RPC API. This worked is tracked here: https://github.com/status-im/nwaku/issues/1076

LNSD:
From my initial conversation with @jm-clius, one of the requirements for the HTTP RESTful API is to be as consistent as we can with [nimbus HTTP REST RPC API](https://github.com/status-im/nimbus-eth2/blob/stable/beacon_chain/rpc/beacon_rest_api.nim). So I did a little analysis on the _nimbus beacon_chain_. These are some highlights of the nimbus implementation that we should follow in _nwaku_:

- Nimbus *beacon_node* exposes the "control interface" in two forms: RPC and REST
- REST server request handlers are installed in the *nim-presto* server via the `installRestHandlers` methods. Handlers are grouped under a common namespace (e.g. *debug, beacon* or *node*).
- API versioning strategy is in the path (e.g. `/eth/v1/debug/beacon/states/{state_id}`) and each resource can have a different version. Within a group of resources, each resource is versioned independently.

LNSD:
HTTP RESTful is a widespread way to expose service APIs over HTTP. There is a standardized way of specifying a RESTful API named OpenAPI.

- The [[OpenAPI Specification](https://swagger.io/specification/)](https://swagger.io/specification/) (OAS) defines a standard, language-agnostic interface to RESTful APIs.
- There is a series of tools built around this standard specification format that enables:
[...]

Issue #983, opened 2022-05-31
General log clean-up


jm-clius:
## Background

Most nwaku nodes run with `DEBUG` level logs.

Although this is a useful (and necessary) log level for debugging purposes, the logs are very noisy even under stable operation, making it difficult to find and investigate unusual conditions

Very frequent logs should either be lowered to `TRACE` level or filtered.

## Examples of frequent logs

These can be individually investigated and filtered/suppressed:

```
DBG 2022-05-31 10:06:24.129+00:00 Accepted an incoming connection            topics="libp2p switch" tid=1 file=switch.nim:203 conn=:6295e8a094fbd5702e8ffc07
DBG 2022-05-31 10:06:24.130+00:00 Exception upgrading incoming               topics="libp2p muxedupgrade" tid=1 file=muxedupgrade.nim:155 exc="Stream EOF!"
```

On systems with RLN enabled:

```
DBG 2022-05-31 09:18:54.305+00:00 rln-relay topic validator is called        topics="wakunode" tid=1 file=wakunode2.nim:499
DBG 2022-05-31 09:18:54.306+00:00 content topic did not match:               topics="wakunode" tid=1 file=wakunode2.nim:508 contentTopic=/railgun/v1/137/fees/json payload="{"data":"7b2266656573223a7b22307830643530306231643865386566333165323163393964316462396136343434643361646631323730223a22307830666635396565383333623330303030222c22307832373931626361316632646534363631656438386133306339396137613934343961613834313734223a223078306239336...
```

LNSD:
I think that it would be interesting to define a set of rules to choose the right logging level. Defining these rules will help us to have consistent logging. I think this should be part of the acceptance criteria for this issue.

For example, for `warn` level what makes sense for me is (from: https://github.com/status-im/nwaku/pull/991#issuecomment-1145902891):

> For example, I would set it to **warn** level in the case that reaching that state was non-critical and uncommon but could affect the app stability.


Let's not forget proper (and well thought) logging is a basic tool for debugging in production environments 😁

jm-clius:
Also investigate, first reported in https://github.com/status-im/nwaku/issues/703

```
failed to store peers topics="wakupeers" tid=1 file=peer_manager.nim:44 err="failed to encode: Failed to encode public key"
```

PR #988, opened 2022-06-02
feat(rest): Add HTTP REST API (#727)

LNSD:
This PR addresses https://github.com/status-im/nwaku/issues/727.

**⚠️ IMPORTANT:** This PR is WIP. Check #727 discussion for context.


LNSD:
**NOTE:** Test files are collocated to the actual sources. This collocation is for convenience but temporary. As the PR evolves I will relocate the test files to the tests folder.

jm-clius:
> **NOTE:** Test files are collocated to the actual sources. This collocation is for convenience but temporary. As the PR evolves I will relocate the test files to the tests folder.

Ah, just as I was going to ask :D

waku/v2/node/rest/debug_api.yaml
```diff
@@ -0,0 +1,34 @@
+openapi: 3.1.0
```
jm-clius, line 1: To me it makes sense to have the `yaml` files separate. Perhaps a `/node/rest/yaml/` or `/node/rest/openapi/` subdirectory. WDYT?

waku/v2/node/rest/test_debug_api_serdes.nim
```diff
@@ -0,0 +1,38 @@
...
+  suite "DebugWakuInfo - decode":
+    test "optional field is not provided":
+      # Given
+      let jsonString = JsonString("""{ "listenAddresses":["123"] }""")
+
+      # When
+      let res = decodeJsonString(DebugWakuInfo, jsonString, requireAllFields = true)
+
+      # Then
+      require(res.isOk)
```
jm-clius, line 20: Really like this logical structuring of unit tests. Good job!

waku/v2/node/rest/json_serdes.nim
```diff
@@ -0,0 +1,57 @@
+import std/typetraits
+import stew/results,
+       serialization,
+       json_serialization,
+       json_serialization/std/[options, net, sets],
+       chronicles
+
+logScope: topics = "api_rest_json_serdes"
```
jm-clius, line 8: Perhaps we should create a general refactoring issue where we fix the `logScope` for all Waku modules to something consistent. I have no specific preference and like this `_underscore_` convention. Other modules, however, are all over the place, with `dot.notation`, `spa ces`, etc. sometimes with `waku.` prefix and sometimes without any prefix. Preferably we want to prefix the scope for related modules so their logs are easily recognisable in a noisy file, as you've done here with `api_`. 👍

waku/v2/node/rest/debug_api.yaml
LNSD: In my opinion, these YAML files should be collocated with the different request handlers—something like this:

```
rest/
├─ debug/
│  ├─ debug.yaml
│  ├─ debug_api_handlers.nim
│  ├─ debug_api_models.nim
├─ store/
│  ├─ store.yaml
│  ├─ store_api_handlers.nim
│  ├─ store_api_models.nim
├─ ...
```

The reason behind this is the [_single responsibility principle_](https://www.oreilly.com/library/view/97-things-every/9780596809515/ch76.html):

> Gather together those things that change for the same reason, and separate those things that change for different reasons.

This cannot be appreciated yet because I am still evaluating all the pros and cons of that approach in Nim and nim-presto. From my experience, collocating everything in that "structured" way boosts maintainability (e.g. removing a resource, or namespace in our case, is as simple as removing one folder).

If you see any other reason (that probably I am not taking into account) to segregate the files, we can have a call and discuss 😁

waku/v2/node/rest/test_debug_api_serdes.nim
LNSD: This comes from my time as QA, structuring tests like this helped me a lot to focus on the _setup-stimulus-assertion_ way of thinking; and I find it very helpful in order to apply the SOLID principles because it helps you to think that a "unit" should do one and only one thing.

From: https://www.browserstack.com/guide/unit-testing-best-practices

> Generally, unit tests comprise three phases:
> 
> * Initialize the application module to be tested AKA the system under test
> * Apply stimulus to the system under tests
> * Observe resultant behavior

And, given that you were a JVM guy 😁 I like very much the way Groovy's Spock models the test cases: https://www.baeldung.com/groovy-spock#2-blocks

waku/v2/node/rest/json_serdes.nim
LNSD: We have this issue: https://github.com/status-im/nwaku/issues/983

From my POC, that one looks to me like the perfect moment to agree and apply the convention.

tests/v2/test_rest_debug_api.nim
```diff
@@ -0,0 +1,51 @@
...
+  ../../waku/v2/node/wakunode2,
+  ../../waku/v2/node/rest/[server, client, debug_api]
+
+
+proc testWakuNode(): WakuNode = 
+  let rng = crypto.newRng()
+  let privkey = crypto.PrivateKey.random(Secp256k1, rng[]).tryGet()
+  let bindIp = ValidIpAddress.init("0.0.0.0")
+  let extIp = ValidIpAddress.init("127.0.0.1")
+  let port = Port(9000)
```
jm-clius, line 17: How do you feel about block `let` statements, e.g.:

```suggestion
  let
    rng = crypto.newRng()
    privkey = crypto.PrivateKey.random(Secp256k1, rng[]).tryGet()
    bindIp = ValidIpAddress.init("0.0.0.0")
    extIp = ValidIpAddress.init("127.0.0.1")
    port = Port(9000)
```

LNSD: I am ok with them. I guess that it's just that I am not used to this Nim particularity 🙂

waku/v2/node/rest/client.nim
```diff
@@ -0,0 +1,4 @@
+import presto/client
```
jm-clius, line 1: Top level `{.push raises: [Defect].}` for all modules. Reasoning [here](https://status-im.github.io/nim-style-guide/errors.exceptions.html).

waku/v2/node/rest/debug_api.nim
```diff
@@ -0,0 +1,94 @@
+import
```
jm-clius, line 1: Top level `{.push raises: [Defect].}` here and elsewhere.

waku/v2/node/rest/debug_api.nim
```diff
@@ -0,0 +1,94 @@
...
+
+####  Server request handlers
+
+proc toDebugWakuInfo(nodeInfo: WakuInfo): DebugWakuInfo = 
+    DebugWakuInfo(
+      listenAddresses: nodeInfo.listenAddresses,
+      enrUri: some(nodeInfo.enrUri)
+    )
+
+const ROUTE_DEBUG_INFOV1* = "/debug/v1/info"
```
jm-clius, line 67: If we believe maintenance on this constant may be necessary in future, can we move it top level (before the `type`s)?

jm-clius (approved):
Very good to see the Rest API taking shape! This PR lgtm. :) Only comment to really address before merging is error handling (top level `{.push raises: [Defect].}` in all modules, except test modules).

waku/v2/node/rest/debug_api.nim
LNSD: I am curious about the maintenance work you foresee with this constant.

The idea is to have a constant per path. So the string literal is not within the handler code. So for example, if we come up with an "info V2" endpoint then we would have another constant collocated to the "info V2" route handler method:

```nim
const ROUTE_DEBUG_INFOV2* = "/debug/v2/info"
```

waku/v2/node/rest/server.nim
```diff
@@ -0,0 +1,86 @@
...
+  RestServerConf(
+      cacheSize: 3,
+      cacheTtl: 60,
+      requestTimeout: 0,
+      maxRequestBodySize: 16_384,
+      maxRequestHeadersSize: 64
+  )
+
+
+template init*(T: type RestServerRef,
```
kaiserd, line 52: Just wondering: What is the reason for using a template here instead of a proc?

waku/v2/node/rest/server.nim
```diff
@@ -0,0 +1,86 @@
...
+    address, 
+    serverFlags = serverFlags,
+    httpHeadersTimeout = headersTimeout,
+    maxHeadersSize = maxHeadersSize,
+    maxRequestBodySize = maxRequestBodySize
+  )
+
+  if res.isErr():
+    notice "Rest server could not be started", address = $address, reason = res.error()
+    nil
```
kaiserd, line 83: Continuing my comment above: Imo, a proc returning a result would be conceptually easier here.

kaiserd (approved):
LGTM! (only one small comment)

waku/v2/node/rest/server.nim
LNSD: I just copied and adapted this server init method from the nimbus codebase.

LNSD:
Seems that rebasing from Github does not sign the commits. So I rebased the branch locally and pushed the rebased branch.

I am waiting for the tests to finish and pass, then I will merge the branch 😁

waku/v2/node/rest/server.nim
kaiserd: oh! There might be a reason for doing this which I don't see :sweat_smile:

PR #989, opened 2022-06-03
chore: disable swap by default

jm-clius:
Minimal PR to disable `swap` protocol by default.

### Why?

`Swap` is not currently being used or investigated by anyone, yet it is enabled by default in `Soft` mode. Impact of keeping it enabled:
1. Many `INFO` level logs that record accounting state in all nodes.
2. Accounting state with no proper capacity management (i.e. a slow memory leak) being kept for each node involved in a `store` transaction.

We can reenable `swap` once it's used for incentivisation.


LNSD (approved):
LGTM ✅

PR #990, opened 2022-06-03
chore: more regular keepalive pings

jm-clius:
Minimal PR to lower the period between keepalive `PING`s from `5.min` to `2.min`.

### Why?

Libp2p TCP connections have an underlying timeout of `10.min`. The keepalive is meant to prevent these and other timeouts from happening, but once every `5.min` leaves very little room for error (if a node misses one `PING` for some reason it might still trigger a timeout).

Note that we do not necessarily want (or need to) make this period either configurable or dependent on a specific underlying constant (e.g. the TCP timeout), as this is mostly a workaround mechanism for fleet nodes to remain connected during idle periods.

### DIFF
--- a/waku/v2/node/wakunode2.nim
+++ b/waku/v2/node/wakunode2.nim
@@ -779,7 +779,7 @@ proc keepaliveLoop(node: WakuNode, keepalive: chronos.Duration) {.async.} =
     await sleepAsync(keepalive)
 
 proc startKeepalive*(node: WakuNode) =
-  let defaultKeepalive = 5.minutes # 50% of the default chronosstream timeout duration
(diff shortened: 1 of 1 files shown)


LNSD (approved):
LGTM ✅

LNSD:
One related/unrelated question. Is this PING mechanism at "application level" or at "transport level"? Do we have something similar for the WebSocket transport?

kaiserd (approved):
LGTM :+1:

jm-clius:
> One question: Is this PING mechanism at "application level" or at "transport level"? Do we have something similar for the WebSocket transport?

Ping is a libp2p protocol that can run over any transport, therefore "application level" from this perspective. It therefore works for any transport, including websocket.

PR #991, opened 2022-06-03
chore: reduce unnecessary logging

jm-clius:
This minimal PR partially addresses https://github.com/status-im/nwaku/issues/983

It cleans up unnecessary logging, correct misleading `ERROR` logs and lower log levels where necessary.

Specifically, it:
- avoids the unnecessary error condition where the peer manager tries to store information for a connecting/disconnecting peer that is not tracked in the store (resulted in both error metrics and logs).
- lowers the per-message log level for nodes with RLN enabled, especially when few messages match the RLN content topic (currently RLN logs all messages at `DEBUG` level on the `wakuv2.test` fleet)
- avoids misleading `ERROR: Retrieving dynamic bootstrap nodes failed` by having an empty bootstrap list be an acceptable condition
- lowers error-level warning for `store` nodes starting with no static `store` peers configured to resume history from - many nodes start like this (i.e. with no desire to `resume`).

### DIFF
--- a/waku/v2/protocol/waku_store/waku_store.nim
+++ b/waku/v2/protocol/waku_store/waku_store.nim
@@ -619,7 +619,7 @@ proc resume*(ws: WakuStore, peerList: Option[seq[RemotePeerInfo]] = none(seq[Rem
     # if no peerList is set then query from one of the peers stored in the peer manager 
     let peerOpt = ws.peerManager.selectPeer(WakuStoreCodec)
     if peerOpt.isNone():
-      error "no suitable remote peers"
+      warn "no suitable remote peers"
       waku_store_errors.inc(labelValues = [dialFailure])
       return err("no suitable remote peers")
 
--- a/waku/v2/node/peer_manager/peer_manager.nim
+++ b/waku/v2/node/peer_manager/peer_manager.nim
@@ -107,6 +107,11 @@ proc loadFromStorage(pm: PeerManager) =
 ##################   
 
 proc onConnEvent(pm: PeerManager, peerId: PeerID, event: ConnEvent) {.async.} =
+  if not pm.peerStore.addressBook.contains(peerId):
+    ## We only consider connection events if we
+    ## already track some addresses for this peer
+    return
+
   case event.kind
   of ConnEventKind.Connected:
     pm.peerStore.connectionBook[peerId] = Connected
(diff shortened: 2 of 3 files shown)


LNSD (approved):
LGTM ✅

I don't know the context of the "logic" related changes. But if tests are passing, I am trusting you, @jm-clius 😉

LNSD:
My two cents on logging levels 🙂 

> avoids misleading `ERROR: Retrieving dynamic bootstrap nodes failed` by having an empty bootstrap list be an acceptable condition

If we want to provide some feedback, the level could be set to `warn` or `info` depending on the severity. For example, I would set it to _warn_ level in the case that reaching that state was non-critical and uncommon but could affect the app stability.

staheri14 (approved):
LGTM!

waku/v2/protocol/waku_store/waku_store.nim
```diff
@@ -619,7 +619,7 @@ proc resume*(ws: WakuStore, peerList: Option[seq[RemotePeerInfo]] = none(seq[Rem
     # if no peerList is set then query from one of the peers stored in the peer manager 
     let peerOpt = ws.peerManager.selectPeer(WakuStoreCodec)
     if peerOpt.isNone():
-      error "no suitable remote peers"
+      warn "no suitable remote peers"
```
kaiserd, line 622: Should this not actually be an error, because at this point, resuming will fail?
Maybe the error message should also say something about resume having failed.

If we want to reserve ERR for things that we cannot recover from, we should also consider lines 622 and 648 for warnings instead of errors.

As @LNSD pointed out, we should agree on which severity level justifies and error.

kaiserd (changes requested):

jm-clius:
> If we want to provide some feedback, the level could be set to `warn` or `info` depending on the severity. For example, I would set it to _warn_ level in the case that reaching that state was non-critical and uncommon but could affect the app stability.

I agree. Could be a section in the contributor guide: https://github.com/status-im/nwaku/issues/994

waku/v2/protocol/waku_store/waku_store.nim
jm-clius: Resuming will indeed fail, but in this case resuming itself failing is not _necessarily_ an error. It simply indicates that the store node has no interest in resuming history from a peer (and many/most store nodes are currently configured in this way). In other words, many correctly configured nwaku nodes will "fail" to resume because they have simply not configured a store peer, indicating that they do not care about resuming past history. I agree that there's probably a better way to handle this (e.g. in configuration), but this change is aimed at keeping ERROR logs from appearing when a correctly-configured node is started up.

kaiserd: Thank you for carifying :).

kaiserd (approved):
LGTM :+1:

Issue #915, opened 2022-03-25
discv5+DNS: consolidating ENR information


kaiserd:
## Background

Currently `nwaku` builds two separate ENRs (in case discv5 is active) holding separate information.
[One](https://github.com/status-im/nim-waku/blob/0131d1437e5f1c729aed0f4d229e4275acecef37/waku/v2/node/wakunode2.nim#L228) when initializing a waku node, and [another one](https://github.com/status-im/nim-waku/blob/0131d1437e5f1c729aed0f4d229e4275acecef37/waku/v2/node/wakunode2.nim#L1133) when initializing the discv5 subsystem.
The former might contain a DNS multiaddress, while the latter contains the UDP port necessary for discv5.

## Goal / Motivation

Both ENR instances should contain the union of their respective current information.

* reduces confusion, because both ENR instances can be used interchangeably
* all information, including dns4 multiaddresses are discoverable via discv5
* sufficient to print a single ENR in the log (see #913)

## non-goals of this issue

For now, the discv5 subsystem holds a separate ENR instance. This issue does not propose consolidating both into a single ENR instance, as that would require more refactoring.
We might address that in the future though.

cc @jm-clius @richard-ramos 

## Acceptance criteria

- [ ] Wakunode2 and discv5 subsystem ENRs contain the same information
- [ ] successfully tested on experimental/local test fleet

jm-clius:
This could be a dependency for https://github.com/waku-org/nwaku/issues/1475

rymnc:
This issue is addressed in #1512

rymnc:
@jm-clius can we close this issue as #1512 has been merged?

PR #996, opened 2022-06-08
feat(enr): add discv5 UDP port to node's main discoverable ENR

jm-clius:
Adds the discv5 UDP port to the "main" discoverable ENR of the node. This way operator nodes referencing/discovering this ENR will have access to both the TCP port for libp2p communication and UDP port for discv5 bootstrapping.

Identified as part of https://github.com/status-im/nwaku/issues/828, in an effort to improve node deployment workflow for operators.

### DIFF
--- a/waku/v2/node/wakunode2.nim
+++ b/waku/v2/node/wakunode2.nim
@@ -1069,7 +1069,8 @@ when isMainModule:
           .mapErr(proc (e: cstring): string = $e)
       else:
         warn "Failed to init Waku DNS discovery"
-    
+
+    debug "No method for retrieving dynamic bootstrap nodes specified."
     ok(newSeq[RemotePeerInfo]()) # Return an empty seq by default
 
   # 3/7 Initialize node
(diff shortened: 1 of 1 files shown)


kaiserd (approved):
LGTM :+1: 

This also addresses part of #915.
Note: WakuInfo, which is used in the info JSON RPC call, uses the [discv5 exclusive ENR](https://github.com/status-im/nwaku/issues/915) if discv5 is enabled.

waku/v2/node/wakunode2.nim
```diff
@@ -172,7 +172,8 @@ proc new*(T: type WakuNode, nodeKey: crypto.PrivateKey,
     wakuFlags = none(WakuEnrBitfield),
     nameResolver: NameResolver = nil,
     sendSignedPeerRecord = false,
-    dns4DomainName = none(string)
+    dns4DomainName = none(string),
+    discv5UdpPort = none[Port]()
```
LNSD, line 176: Nitpick/curiosity: Why are you defining these `none` values using two different notations? 

From my experience, code style consistency is an important aspect of codebase maintainability. And it is usually overlooked. If all the statements are uniform, it is easier to detect things that are not "correct" (e.g.bugs).

In this case: IMHO we should be consistent and use just one format. I like the `none(Port)` format, over the generic parameter format.

LNSD (approved):
Besides my little "nitpicking" comment, the rest LGTM ✅

waku/v2/node/wakunode2.nim
jm-clius: Not at all nitpicky and in fact a very important principle. Did this to be consistent with the way the "other" port and IP address options are initialised in https://github.com/status-im/nwaku/blob/9a16883012cabc84a8f7793eed6f486fc85cbc46/waku/v2/node/wakunode2.nim#L164, but then missed the inconsistency with the parameter right above this line. I've now changed everything to use the `none(type)` format, which I also prefer ([f2fa3ef](https://github.com/status-im/nwaku/pull/996/commits/f2fa3efe9f1a1809a09235973aef1cdcf97e7707)). Thanks for pointing this out.

PR #1006, opened 2022-06-16
fix: encode protobuffers using proto3

richard-ramos:
Currently protobuffers are being encoded using `proto2`, while on the specs `proto3` is the version meant to be used, This results in null/empty fields being written in the buffer. i.e: the following WakuMessage:
```
payload: @[1, 2, 3, 4],
contentTopic: "hola",
version:      0,
timestamp:    12345,
```
Is generating this byte array: 
```
[10, 4, 1, 2, 3, 4, 18, 4, 104, 111, 108, 97, 24, 0, 80, 242, 192, 1, 170, 1, 173, 3, 10, 128, 2, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 18, 32, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 26, 32, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 34, 32, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 42, 32, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 50, 32, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
```
matching `proto2` specs,  while if they were encoded using `proto3`, the following output would be generated:
```
[10, 4, 1, 2, 3, 4, 18, 4, 104, 111, 108, 97, 80, 242, 192, 1]
```
which is more efficient (less bytes being transmitted), since it does not include the empty fields (see:  https://developers.google.com/protocol-buffers/docs/proto3#default).

This PR introduces new functions:  `write3` and `finish3` which should be used instead of `write` and `finish` when encoding a protobuffer with `proto3`. It also fixes some instances of protobuffers being encoded without calling .finish()

@D4nte, this will require a change in js-waku to handle empty/default values, since right now a field that's not present in the byte array for a protobuffer is assumed to be undefined (like in proto2), and it should instead return the default value for a field type.

### DIFF
--- a/waku/v2/utils/protobuf.nim
+++ b/waku/v2/utils/protobuf.nim
@@ -1,0 +1,20 @@
+{.push raises: [Defect].}
+
+import
+  libp2p/protobuf/minprotobuf,
+  libp2p/varint
+
+# Collection of utilities related to protobuffer encoding
+
+proc write3*(proto: var ProtoBuffer, field: int, value: auto) =
+  if default(type(value)) != value:
+    proto.write(field, value)
+
+proc finish3*(proto: var ProtoBuffer) =
+  if proto.buffer.len > 0:
+    proto.finish()
+  else:
+    proto.offset = 0
+
+proc `==`*(a: zint64, b: zint64): bool =
+  int64(a) == int64(b)
--- a/waku/v2/protocol/waku_filter/waku_filter.nim
+++ b/waku/v2/protocol/waku_filter/waku_filter.nim
@@ -13,6 +13,7 @@ import
   libp2p/crypto/crypto,
   waku_filter_types,
   ../../utils/requests,
+  ../../utils/protobuf,
   ../../node/peer_manager/peer_manager
 
 # NOTE This is just a start, the design of this protocol isn't done yet. It
@@ -77,18 +78,23 @@ proc unsubscribeFilters(subscribers: var seq[Subscriber], request: FilterRequest
 proc encode*(filter: ContentFilter): ProtoBuffer =
   var output = initProtoBuffer()
 
-  output.write(1, filter.contentTopic)
+  output.write3(1, filter.contentTopic)
+
+  output.finish3()
+
   return output
 
 proc encode*(rpc: FilterRequest): ProtoBuffer =
   var output = initProtoBuffer()
   
-  output.write(1, uint64(rpc.subscribe))
+  output.write3(1, uint64(rpc.subscribe))
 
-  output.write(2, rpc.pubSubTopic)
+  output.write3(2, rpc.pubSubTopic)
 
   for filter in rpc.contentFilters:
-    output.write(3, filter.encode())
+    output.write3(3, filter.encode())
+
+  output.finish3()
 
   return output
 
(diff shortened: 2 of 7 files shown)


staheri14 (approved):
Thanks! looks good to me!

waku/v2/utils/protobuf.nim
```diff
@@ -0,0 +1,18 @@
+import
```
jm-clius, line 1:
```suggestion
{.push raises: [Defect].}

import
```

PR #1007, opened 2022-06-16
feat(rest): Add HTTP REST API (#727). Amend Debug REST API client

LNSD:
As the title states, this is an amendment to the Debug REST API so:
- We can access the response status code.
- We avoid the presto client raising an exception on non-2XX status codes.
- Modules have been split and moved under the corresponding namespace folder.

This PR is associated with issue #727


jm-clius (approved):
LGTM!

LNSD:
The last commit was for solving the OpenAPI spec issues. Now we can generate the documentation for the Debug REST API too: https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/status-im/nwaku/issue-727-debug/waku/v2/node/rest/debug/debug_api.yaml#tag/relay/operation/deleteSubscriptions

image

waku/v2/node/rest/debug/api_types.nim
```diff
@@ -20,6 +15,15 @@ type
     enrUri*: Option[string]
 
 
+#### Type conversion
+
+proc toDebugWakuInfo*(nodeInfo: WakuInfo): DebugWakuInfo = 
+    DebugWakuInfo(
+      listenAddresses: nodeInfo.listenAddresses,
+      enrUri: some(nodeInfo.enrUri)
```
kaiserd, line 23: I am wondering why we need a wrapper for WakuInfo that wraps enrUri in an Option.
Afaik, WakuInfo always contains an enrUri.

waku/v2/node/rest/debug/debug_api.nim
```diff
@@ -0,0 +1,45 @@
...
+####  Server request handlers
+
+const ROUTE_DEBUG_INFOV1* = "/debug/v1/info"
+
+proc installDebugInfoV1Handler(router: var RestRouter, node: WakuNode) =
+  router.api(MethodGet, ROUTE_DEBUG_INFOV1) do () -> RestApiResponse:
+    let info = node.info().toDebugWakuInfo()
+    let resp = RestApiResponse.jsonResponse(info, status=Http200)
+    if resp.isErr():
+      debug "An error ocurred while building the json respose", error=resp.error()
```
kaiserd, line 25:
```suggestion
      debug "An error occurred while building the json respose", error=resp.error()
```

kaiserd (approved):
LGTM :+1: (minor comments in-line)

waku/v2/node/rest/debug/api_types.nim
LNSD: I was following the [RPC RFC for the WakuInfo](https://rfc.vac.dev/spec/16/#wakuinfo) and it says that it should be optional.

kaiserd: Good point :). We could make it mandatory or change our `WakuInfo` type. Imo, having two types is confusing. @jm-clius wdyt?
When adding the ENR field to the RFC, I assumed that `nawku` always provides the ENR in answers to RPC calls, but not all implementations have to do that. I did not assume that `nwaku` would act as client against the debug API.

LNSD: > I did not assume that nwaku would act as client against the debug API.

Hmmmmm I think I need to do a clarification on the client part. 

I am "implementing" a client for the REST HTTP API mainly for testing purposes. As NIM does not have "interfaces" (like Java, Go, etc.) we took the decision of testing this from an end-to-end point of view. To do that, I had to provide a client implementation that is the one used in the tests.

There's something that we could also do which is to provide a CLI REST API client to interact with the node through this API. Maybe, it is interesting from the operators' point of view; But this is not part of the scope of the #727 issue.

kaiserd: In case the client functionality is only for tests, imo it is fine to wrap the type like this :).
I was not aware of what the scope of it should finally be.

Also agree with your second point. A CLI REST API client would be nice to tackle in a follow up issue/PR.

PR #1014, opened 2022-06-21
Fixed handling of bindParam(uint32) in sqlite.nim

Bulat-Ziganshin:
1. n is already int
2. when val >= 1<<31, using sqlite3_bind_int should store it as negative value


LNSD (approved):
LGTM ✅

PR #1016, opened 2022-06-22
Chore(rln-relay): better documentation for rln nim bindings 

staheri14:
This is a minimal PR merely to add more detailed descriptions to the rln nim bindings to ease its replacement with the new zero-kit.


oskarth (approved):
LGTM

@s1fr0 should approve PR before merged

waku/v2/protocol/waku_rln_relay/rln.nim
```diff
@@ -29,38 +29,61 @@ type Auth* = object
...
 #----------------------------------------------------------------------------------------------
 #-------------------------------- Common procedures -------------------------------------------
 
 proc new_circuit_from_params*(merkle_depth: uint,
                               parameters_buffer: ptr Buffer,
                               ctx: ptr RLN[Bn256]): bool {.importc: "new_circuit_from_params".}
-                              
+## creates an instance of rln object as defined by the rln lib https://github.com/kilic/rln/blob/7ac74183f8b69b399e3bc96c1ae8ab61c026dc43/src/public.rs#L48
+## merkle_depth represent the depth of the Merkle tree
+## parameters_buffer holds prover and verifier keys
```
s1fr0, line 75: To be precise, `parameters_buffer` here corresponds to the user's proving key. The proving key, in turn, contains a copy of the verifying key.

waku/v2/protocol/waku_rln_relay/rln.nim
```diff
@@ -29,38 +29,61 @@ type Auth* = object
...
 #-------------------------------- zkSNARKs operations -----------------------------------------
-
 proc key_gen*(ctx: RLN[Bn256], keypair_buffer: ptr Buffer): bool {.importc: "key_gen".}
+## generates id key and id commitment key serialized inside keypair_buffer as | id_key <32 bytes>| id_commitment_key <32 bytes> |
+## the return bool value indicates the success or failure of the operation
 
 proc generate_proof*(ctx: RLN[Bn256],
                     input_buffer: ptr Buffer,
                     output_buffer: ptr Buffer): bool {.importc: "generate_proof".}
 ## input_buffer serialized as  [ id_key<32> | id_index<8> | epoch<32> | signal_len<8> | signal<var> ]
```
s1fr0, line 55: Why `id_index` is 8 bytes long? Aren't 4 enough? (32 bytes * 2^(8*4) = 128GiB of leaves).
I guess its based on the typical 8-bytes `usize` for most 64bits architecture, but over 32bits architectures such length won't make sense.

waku/v2/protocol/waku_rln_relay/rln.nim
```diff
@@ -29,38 +29,61 @@ type Auth* = object
...
-
+## get_root populates the passed pointer output_buffer with the current tree root
+## the output_buffer holds the Merkle tree root of size 32 bytes
+## the return bool value indicates the success or failure of the operation
+## 
 #----------------------------------------------------------------------------------------------
 #-------------------------------- zkSNARKs operations -----------------------------------------
-
 proc key_gen*(ctx: RLN[Bn256], keypair_buffer: ptr Buffer): bool {.importc: "key_gen".}
+## generates id key and id commitment key serialized inside keypair_buffer as | id_key <32 bytes>| id_commitment_key <32 bytes> |
```
s1fr0, line 49: You may add that id_commitment_key is the poseidon hash of id_key?

waku/v2/protocol/waku_rln_relay/rln.nim
```diff
@@ -29,38 +29,61 @@ type Auth* = object
...
+## the return bool value indicates the success or failure of the operation
 
 proc generate_proof*(ctx: RLN[Bn256],
                     input_buffer: ptr Buffer,
                     output_buffer: ptr Buffer): bool {.importc: "generate_proof".}
 ## input_buffer serialized as  [ id_key<32> | id_index<8> | epoch<32> | signal_len<8> | signal<var> ]
 ## output_buffer holds the proof data and should be parsed as |proof<256>|root<32>|epoch<32>|share_x<32>|share_y<32>|nullifier<32>|
```
s1fr0, line 56: Unrelated to this PR, but the zk proof can be compressed to 128 bytes (only x coordinates are kept for G1,G2,G3 + 1 bit for the sign of y), so this will improve data communication.

waku/v2/protocol/waku_rln_relay/rln.nim
```diff
@@ -29,38 +29,61 @@ type Auth* = object
...
-## sizes are in bytes
+## integers wrapped in <> indicate value sizes in bytes
+## the return bool value indicates the success or failure of the operation
+## 
 proc verify*(ctx: RLN[Bn256],
             proof_buffer: ptr Buffer,
             result_ptr: ptr uint32): bool {.importc: "verify".}
 ## proof_buffer [ proof<256>| root<32>| epoch<32>| share_x<32>| share_y<32>| nullifier<32> | signal_len<8> | signal<var> ]
+## the return bool value indicates the success or failure of the verification
```
s1fr0, line 64: This sounds misleading: the return bool value indicates the success of the verify API call, not if the proof results successfully verified. To verify thatm you need to check `result_ptr` value (which seems that counterintuitively has to be set to 0 for the proof to be valid).

s1fr0 (commented):
Looks good! Left some comments: most relevant probably the one about proof verification.

waku/v2/protocol/waku_rln_relay/rln.nim
staheri14: `uint8` is the smallest among the integer types supported by Rust (and typically by other languages), which also appeared to be large enough for the leaf index. Of course, smaller than that also works but requires a custom type definition.

waku/v2/protocol/waku_rln_relay/rln.nim
staheri14: Curious on how this distinction would matter? In my view, it generates CRS that embodies both the proving keys and the verification keys.

waku/v2/protocol/waku_rln_relay/rln.nim
staheri14: Thanks, clarified in https://github.com/status-im/nwaku/pull/1016/commits/efdfad78cf5d3001887ffd9fef145b610be753f5

waku/v2/protocol/waku_rln_relay/rln.nim
staheri14: Sure, done in https://github.com/status-im/nwaku/pull/1016/commits/efdfad78cf5d3001887ffd9fef145b610be753f5

waku/v2/protocol/waku_rln_relay/rln.nim
```diff
@@ -29,38 +29,61 @@ type Auth* = object
...
   
 #------------------------------ Merkle Tree operations -----------------------------------------
-
 proc update_next_member*(ctx: RLN[Bn256],
                         input_buffer: ptr Buffer): bool {.importc: "update_next_member".}
+## input_buffer points to the id commitment byte seq
+## the return bool value indicates the success or failure of the operation
 
 proc delete_member*(ctx: RLN[Bn256], index: uint): bool {.importc: "delete_member".}
+## index is the position of the id commitment key to be deleted from the tree
```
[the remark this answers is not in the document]
staheri14: Done in https://github.com/status-im/nwaku/pull/1016/commits/efdfad78cf5d3001887ffd9fef145b610be753f5

staheri14:
> Looks good! Left some comments: most relevant probably the one about proof verification.

Thanks @s1fr0 for the comments! they are addressed in https://github.com/status-im/nwaku/pull/1016/commits/efdfad78cf5d3001887ffd9fef145b610be753f5/. Please let me know if further edits are required.

waku/v2/protocol/waku_rln_relay/rln.nim
staheri14: Good point, but it comes with a trade-off where it would save us bandwidth, and will add computation to the verifying peers.

waku/v2/protocol/waku_rln_relay/rln.nim
s1fr0: It doesn't matter indeed, it is just a clarification: the proving key just contains the verification key. IMO a RLN-relay node doesn't need to generate proofs but only verify them, so in some circumstance (optimization?) it might make sense to separate the two (proving keys are bigger than verification keys).

waku/v2/protocol/waku_rln_relay/rln.nim
s1fr0: The overhead is 3 square roots (i.e. 3 exponentiations) in Fr vs. 128 bytes space per proof per node. While the proof expansion can be done in parallel with other proofs, increased proof size will affect bandwidth and thus will slower propagation in the network of messages. It's very unclear how the two interplay though, so hard to give an opinion without proper benchmarks. We can just keep this in mind for the moment and do some tests later.

waku/v2/protocol/waku_rln_relay/rln.nim
```diff
@@ -29,38 +29,64 @@ type Auth* = object
...
-## sizes are in bytes
+## integers wrapped in <> indicate value sizes in bytes
+## the return bool value indicates the success or failure of the operation
+## 
 proc verify*(ctx: RLN[Bn256],
             proof_buffer: ptr Buffer,
             result_ptr: ptr uint32): bool {.importc: "verify".}
 ## proof_buffer [ proof<256>| root<32>| epoch<32>| share_x<32>| share_y<32>| nullifier<32> | signal_len<8> | signal<var> ]
+## the return bool value indicates the success or failure of the call to the verify function
+## the verification of the zk proof  is available in result_ptr, where 0 indicates success and 1 is failure
```
s1fr0, line 67:
```suggestion
## the verification of the zk proof is available in result_ptr, where 0 indicates success and 1 is failure
```

s1fr0 (commented):
Thanks @staheri14. LGTM except the 8 bytes assigned to the `index_id`. 

Besides this, where do you use the Auth object? https://github.com/status-im/nwaku/blob/6f93b219283a489cfdac7219ae126e902549c04d/waku/v2/protocol/waku_rln_relay/rln.nim#L27
It's in kilic's code, but I don't think is used here.

staheri14:
> Thanks @staheri14. LGTM except the 8 bytes assigned to the `index_id`.
> 
> Besides this, where do you use the Auth object?
> 
> https://github.com/status-im/nwaku/blob/6f93b219283a489cfdac7219ae126e902549c04d/waku/v2/protocol/waku_rln_relay/rln.nim#L27
> 
> 
> It's in kilic's code, but I don't think is used here.

In the early version of the lib, AuthObj was used to save the id key and the index of id commitment, but it it was removed from the later versions.

staheri14:
@s1fr0 Please let me know if any further clarification is needed, otherwise, I'll merge the PR.

waku/v2/protocol/waku_rln_relay/rln.nim
staheri14: Proof generation is needed as part of message publishing, it contains both capabilities i.e., relaying and publishing.

s1fr0 (approved):
LGTM! Some details will be addressed in followup PRs when we will replace kilic's implementations with zerokit RLN.

waku/v2/protocol/waku_rln_relay/rln.nim
staheri14: Right, misinterpreted the sizes. Right, the 32 depth of the tree seems reasonable and large enough, and perhaps `uint32` would also do the job. But considering that this data is not being transmitted as part of the proof on the wire, it does not cost bandwidth. Would it be of any help to change it to `uint32`? if yes, then we can do it.

waku/v2/protocol/waku_rln_relay/rln.nim
s1fr0: The delay would be mainly in bandwidth/data transfer if you want it to scale: 1M messages exchanged x 128 extra bytes ~ 128MB ~ 6s at 20MB/s (optimistic or pessimistic?) per node. To be better, square roots if Fp have to take around 200ns each. Will see with the integration of zerokit if we can have some benchmark.

waku/v2/protocol/waku_rln_relay/rln.nim
s1fr0: The problem here is not about bandwidth but the architecture: on 32 bits machines `usize` is 4 bytes and you then need a safe conversion from `u64` to `usize` in order to pass this to the RLN API. The other way around, i.e. when serializing the input, keeping index at 8 bytes, requires casting to a non-native type (`u64`) before serializing it. This extra overhead and extra checks doesn't make sense here if 4 bytes out of 8 will always be 0. We can address this in a later PR.

staheri14: I see, if the concern is about interaction with the Rust lib API, then the rln lib and the nim code will run and be compiled on the same machine with the same architecture, right? then my understanding is that `uint` will be rendered with an identical size both by the nim code and by the underlying lib, so no conversion would be needed. Or are you referring to some other issue?

waku/v2/protocol/waku_rln_relay/rln.nim
staheri14: Thanks for the clarification, I understand your point, but that depends on the messaging rate of the application which in your example is 1 million per second which may or may not be the case for all the apps. Let's see what benchmarking tells us.

waku/v2/protocol/waku_rln_relay/rln.nim
s1fr0: The issue will be then in the deserialization: right now you read a fixed amount of 8 bytes (the common `usize` byte-size over 64 bits architectures); to make doable what you say you need to first get the size of `usize` (4 or 8, but not necessarily) and then read that amount of bytes to deserialize the index. To me makes sense to just work with `u32`.

Issue #1017, opened 2022-06-23
Slow store queries may block fleet nodes


jm-clius:
With the SQLite-only store it seems as if nodes can become unresponsive when building responses to sparse queries.

This has been observed on the `wakuv2.prod` fleet for `node-01.gc-us-central1-a.wakuv2.prod` around 8 UTC on 2022-06-23, though the issue may have been present quite some time before that. New connection attempts to the node time out and consul reports metrics reporting failures.

Successive backtraces on the running node reveals that the node is stuck answering a store query:

```
(gdb) bt
#0  0x000055af8ad2ce93 in pcache1PinPage (pPage=<optimized out>) at /app/vendor/nim-sqlite3-abi/sqlite3.c:50879
#1  pcache1FetchStage2 (pCache=0x7f6894694ef8, iKey=31504, createFlag=2) at /app/vendor/nim-sqlite3-abi/sqlite3.c:50879
#2  0x000055af8ad82ca9 in sqlite3PcacheFetch (createFlag=3, pgno=31504, pCache=<optimized out>) at /app/vendor/nim-sqlite3-abi/sqlite3.c:57397
#3  getPageNormal (pPager=0x7f6894694718, pgno=31504, ppPage=0x7ffec9598190, flags=2) at /app/vendor/nim-sqlite3-abi/sqlite3.c:57397
#4  0x000055af8ad84637 in sqlite3PagerGet (flags=2, ppPage=0x7ffec9598190, pgno=<optimized out>, pPager=<optimized out>) at /app/vendor/nim-sqlite3-abi/sqlite3.c:69670
#5  accessPayload (pCur=0x7f68941dc0d8, offset=0, amt=52524,
    pBuf=0x7f68924b818a "83031613061356561653536663432363633343030222c22307839303239666466616539613033313335383436333831633763653136353935633335353465313061223a2230783036313130613833623965636366313935383030222c223078376230653"..., eOp=0) at /app/vendor/nim-sqlite3-abi/sqlite3.c:4134
#6  0x000055af8ad8595a in sqlite3BtreePayload (pBuf=<optimized out>, amt=65182, offset=107, pCur=0x7f68941dc0d8) at /app/vendor/nim-sqlite3-abi/sqlite3.c:77264
#7  sqlite3VdbeMemFromBtree (pCur=0x7f68941dc0d8, offset=107, amt=65182, pMem=0x7f68941dd588) at /app/vendor/nim-sqlite3-abi/sqlite3.c:11728
[...]

PR #1018, opened 2022-06-23
fix(store): limit dbPageSize

kaiserd:
This PR limits `dbPageSize` avoiding excessive use of memory in case of sparse or non-matching filters.
This PR partly solves #1017

### DIFF
--- a/waku/v2/node/storage/message/waku_message_store.nim
+++ b/waku/v2/node/storage/message/waku_message_store.nim
@@ -273,7 +273,8 @@ proc adjustDbPageSize(dbPageSize: uint64, matchCount: uint64, returnPageSize: ui
   var ret =
     if matchCount < 2: dbPageSize * returnPageSize
     else: dbPageSize * (returnPageSize div matchCount)
-  if ret >= maxDbPageSize: ret = maxDbPageSize
+  if ret >= maxDbPageSize:
+    ret = maxDbPageSize
   trace "dbPageSize adjusted to: ",  ret
   ret
 
--- a/waku/v2/node/storage/message/waku_message_store.nim
+++ b/waku/v2/node/storage/message/waku_message_store.nim
@@ -267,8 +267,12 @@ method getAll*(db: WakuMessageStore, onData: message_store.DataProc): MessageSto
   ok gotMessages
 
 proc adjustDbPageSize(dbPageSize: uint64, matchCount: uint64, returnPageSize: uint64): uint64 {.inline.} =
-  var ret = if matchCount < 2: dbPageSize * returnPageSize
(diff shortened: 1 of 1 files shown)


waku/v2/node/storage/message/waku_message_store.nim
```diff
@@ -267,8 +267,12 @@ method getAll*(db: WakuMessageStore, onData: message_store.DataProc): MessageSto
   ok gotMessages
 
 proc adjustDbPageSize(dbPageSize: uint64, matchCount: uint64, returnPageSize: uint64): uint64 {.inline.} =
-  var ret = if matchCount < 2: dbPageSize * returnPageSize
+  const maxDbPageSize: uint64 = 20000 # the maximum DB page size is limited to prevent excessive use of memory in case of very sparse or non-matching filters. TODO: dynamic, adjust to available memory
```
LNSD, line 270: Could you put this constant outside of the function?

waku/v2/node/storage/message/waku_message_store.nim
```diff
@@ -267,8 +267,12 @@ method getAll*(db: WakuMessageStore, onData: message_store.DataProc): MessageSto
...
+  if dbPageSize >= maxDbPageSize: return maxDbPageSize
```
LNSD, line 271: I prefer that we put early returns in different lines. This one-liner does not help with code readability.
```suggestion
  if dbPageSize >= maxDbPageSize: 
    return maxDbPageSize
```

waku/v2/node/storage/message/waku_message_store.nim
```diff
@@ -267,8 +267,12 @@ method getAll*(db: WakuMessageStore, onData: message_store.DataProc): MessageSto
...
+  var ret =
+    if matchCount < 2: dbPageSize * returnPageSize
     else: dbPageSize * (returnPageSize div matchCount)
+  if ret >= maxDbPageSize: ret = maxDbPageSize
```
LNSD, line 275: Same here, multiline conditional statements help with code readability.
```suggestion
  if ret >= maxDbPageSize:
    ret = maxDbPageSize
```
If you want to have a oneliner, use the _std/math_ `min(x,y)` method:

```suggestion
ret = min(ret, maxDbPageSize)
```

LNSD (changes requested):
Sorry if my comments sound a little nitpicked 🙏🏼

jm-clius (approved):
LGTM. No strong opinion about where the `const` resides, but since it is a workaround "stand-in" for a heuristic I'm fine with it hiding in scope.

waku/v2/node/storage/message/waku_message_store.nim
LNSD: Ok, then LGTM ✅😁

kaiserd:
Just as info: Most likely, the dbPageSize will be replaced by putting the filtering into the WHERE clause in the future.

Issue #998, opened 2022-06-10
Moving the code blocks related to mounting rln-relay from wakunode2 to a separate module 


staheri14:
This is an issue tracking the following comment https://github.com/status-im/nwaku/pull/992#discussion_r893622325
> `wakunode2.nim` is becoming unwieldy and difficult to debug so we are already planning to significantly improve its structure and move much functionality to other modules. 

In line with this comment, `mountRlnRelayStatic` and `mountRlnRelayDynamic` procs should be moved to a separate module, and the wakunode2 module can be revised to:
```
when defined(rln): 
  if conf.rlnRelay:
    mountRlnRelay(node, conf)
```
where `mountRlnRelay` lives in a separate module and reads from the `node` and `conf` to call either `mountRlnRelayStatic()` or `mountRlnRelayDynamic()`.

PR #1019, opened 2022-06-23
Chore(rln-relay): decouples rln-relay and wakunode2 modules

staheri14:
This PR closes https://github.com/status-im/nwaku/issues/998

As part of the code refactoring, a circular dependency occurred between the `waku_rln_relay_utils` module and the `wakunode2` module. To address that I moved the`WakuNode` type def to a separate module. For consistency, all the other types defs in wakunode2 are also moved to the same module.

### DIFF
--- a/tests/v2/test_waku_rln_relay_onchain.nim
+++ b/tests/v2/test_waku_rln_relay_onchain.nim
@@ -3,11 +3,11 @@
 {.used.}
 
 import
-  std/options, sequtils, times,
+  std/options,
   testutils/unittests, chronos, chronicles, stint, web3, json,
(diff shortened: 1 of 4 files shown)


LNSD (approved):
LGTM ✅

It's a great step forward in the direction of reducing the size/cognitive load of `wakunode2.nim` file 😁

jm-clius (approved):
LGTM! Thanks for this @staheri14. Agree with @LNSD that this is a great step forward in disentangling the wakunode2 module and is a pattern that we should extend to all other protocols as well.

s1fr0 (approved):
Looks good!

Issue #1024, opened 2022-06-29
[WIP] Store Roadmap


kaiserd:
# Waku Store Roadmap

> ⚠️  **NOTE:** Work in progress. This issue is incomplete and will be edited.

This issue tracks tasks related to our roadmap for the message store.

The goal is for the message store to evolve from offering the functionality the store protocol needs to an efficient production-grade implementation.

### TODO
- [ ] #1159
- [x] #1292
- [ ] store: document waku store's grafana dashboard based troubleshooting
- [ ] store: limit max concurrent history queries handled by a store node
- [ ] store: SQLite message store common queries prepared statements

### DONE
- [x] #1103
- [x] #1155
  	* #1169 
- [x] #1026
- [x] Remove VACCUM on insertion
	* #1025 
- [x] Store interface that abstracts from concrete store implementations
	* #1043
	* #1070 
	* #1141
	* #1147
	* #1163 
	* #1165
	* #1164 
	* #1167 

### Backlog/Performance Improvements

* Investigate the blocking issue mentioned in #982
* Move from the three-column primary key `(senderTimestamp, id, pubsubTopic)` to a single-column primary key that binary encodes `(senderTimestamp, id, pubsubTopic)`.
* Add more indexes (also backward index)
* ~make  `--sqlite-retention-time` a duration, so that it the retention time can be specified in days and other time units besides seconds.~ Moved to #1103
* Investigate an in-memory cache for the SQLite store
  - Since the Waku message store does not need the relational property of SQL, we should investigate [Redis](https://redis.io/) (and other efficient key-value stores) as an alternative
  - A persistence layer can be added
  - Losing a small number of messages should be acceptable if we gain a significant performance boost
    - Such losses are expected to be rare, if they even occur...
- [ ] #1154

jm-clius:
Excellent roadmap. Thanks @kaiserd. @LNSD may get involved here.

cc @iurimatias @richard-ramos

richard-ramos:
A possible improvement for sqlite could be to execute this pragma on init:
```
pragma temp_store = memory;
```
Depending on the query execution plan, it might improve performance as temporary tables and indices would be kept in memory.

LNSD:
I am taking over #1026

D4nte:
What about being able to configure the store limit in days instead of number of messages?

kaiserd:
For now, I added a retention time option to be specified in seconds: #982
[...]

PR #1025, opened 2022-06-29
perf(store): remove vacuum

kaiserd:
[Vacuum is an expensive operation](https://wiki.mozilla.org/Firefox/Projects/Places_Vacuum#Vacuum_timings).
This PR removes vacuum from the batch delete operation in the store.
For now, we rather vacuum "manually".
In the future we could run vacuum in a timer task.

See #1024


jm-clius (approved):
MSTM!

PR #1029, opened 2022-06-29
Feat(Rln relay): Enables and integrates onchain rln into chat2

staheri14:
This PR enables the on-chain mode of rln-relay in chat2.


examples/v2/config_chat2.nim
```diff
@@ -248,7 +248,42 @@ type
       desc: "the pubsub topic for which rln-relay gets enabled",
       defaultValue: "/waku/2/default-waku/proto"
       name: "rln-relay-pubsub-topic" }: string
-
+      
+    rlnRelayDynamic* {.
+      desc: "Enable  waku-rln-relay with on-chain dynamic group management: true|false",
```
kaiserd, line 253:
```suggestion
      desc: "Enable waku-rln-relay with on-chain dynamic group management: true|false",
```

kaiserd (approved):
LGTM :+1:

s1fr0 (commented):
Looks good to me! I just wonder if it makes sense here to align default values to the defaults nim implicitly assigns to new variables, e.g. an all-0 IdKey, IdCommitment etc.
Even in such case, however, to make default consistent, IdCommitment should be the Poseidon hash of IdKey, which clearly is not the case if values are kept as default (assuming "" is later processed as the all-0 word).

LNSD (approved):

kgrgpg (approved):
LGTM

staheri14:
> Looks fine to me! I guess the idea with echo instead of debug is to make it easier for people to see info in chat without going to log files?

Indeed

staheri14:
> Looks good to me! I just wonder if it makes sense here to align default values to the defaults nim implicitly assigns to new variables, e.g. an all-0 IdKey, IdCommitment etc.

If the id commitment or idKey are left as empty strings, then a new key-pair gets generated in the code, so the empty string does not indicate a zero value of id Key or id commitment key, it is rendered as unassigned.

examples/v2/config_chat2.nim
staheri14: done in 604add4b381d7a1f360f3165f084b0eff93509c7

PR #1030, opened 2022-06-30
Use Zerokit RLN module in nwaku

s1fr0:
This PR replaces [Kilic's RLN](https://github.com/kilic/rln) implementation with [Zerokit RLN](https://github.com/vacp2p/zerokit) module in nwaku.

Minor changes to the nim wrappers were required, as well us update of constants in tests.

### DIFF
--- a/tests/v2/test_waku_rln_relay.nim
+++ b/tests/v2/test_waku_rln_relay.nim
@@ -15,7 +15,9 @@ import
 const RLNRELAY_PUBSUB_TOPIC = "waku/2/rlnrelay/proto"
 const RLNRELAY_CONTENT_TOPIC = "waku/2/rlnrelay/proto"
 
+
 procSuite "Waku rln relay":
+
   asyncTest "mount waku-rln-relay in the off-chain mode":
     let
       nodeKey = crypto.PrivateKey.random(Secp256k1, rng[])[]
@@ -66,33 +68,17 @@ procSuite "Waku rln relay":
 suite "Waku rln relay":
   test "key_gen Nim Wrappers":
     var
-      merkleDepth: csize_t = 32
-      # parameters.key contains the parameters related to the Poseidon hasher
-      # to generate this file, clone this repo https://github.com/kilic/rln
-      # and run the following command in the root directory of the cloned project
-      # cargo run --example export_test_keys
-      # the file is generated separately and copied here
-      parameters = readFile("waku/v2/protocol/waku_rln_relay/parameters.key")
-      pbytes = parameters.toBytes()
-      len: csize_t = uint(pbytes.len)
-      parametersBuffer = Buffer(`ptr`: addr(pbytes[0]), len: len)
-    check:
-      # check the parameters.key is not empty
-      pbytes.len != 0
+      merkleDepth: csize_t = 20
 
-    var
-      rlnInstance: RLN[Bn256]
-    let res = new_circuit_from_params(merkleDepth, addr parametersBuffer,
-        addr rlnInstance)
+    var rlnInstance = createRLNInstance()
     check:
-      # check whether the circuit parameters are generated successfully
-      res == true
+      rlnInstance.isOk == true
 
     # keysBufferPtr will hold the generated key pairs i.e., secret and public keys
     var
       keysBuffer: Buffer
       keysBufferPtr = addr(keysBuffer)
-      done = key_gen(rlnInstance, keysBufferPtr)
+      done = key_gen(rlnInstance.value(), keysBufferPtr)
     check:
       # check whether the keys are generated successfully
       done == true
@@ -436,6 +422,8 @@ suite "Waku rln relay":
       shareX: MerkleNode
       shareY: MerkleNode
       nullifier: Nullifier
+      rlnIdentifier: RlnIdentifier
+
     # populate fields with dummy values
     for x in proof.mitems: x = 1
     for x in merkleRoot.mitems: x = 2
@@ -443,6 +431,7 @@ suite "Waku rln relay":
     for x in shareX.mitems: x = 4
     for x in shareY.mitems: x = 5
     for x in nullifier.mitems: x = 6
+    for x in rlnIdentifier.mitems: x = 7
 
     let
       rateLimitProof = RateLimitProof(proof: proof,
@@ -450,7 +439,8 @@ suite "Waku rln relay":
                           epoch: epoch,
                           shareX: shareX,
                           shareY: shareY,
-                          nullifier: nullifier)
+                          nullifier: nullifier,
+                          rlnIdentifier: rlnIdentifier)
       protobuf = rateLimitProof.encode()
       decodednsp = RateLimitProof.init(protobuf.buffer)
 
--- a/tests/v2/test_waku_rln_relay.nim
+++ b/tests/v2/test_waku_rln_relay.nim
@@ -358,7 +358,7 @@ suite "Waku rln relay":
       hashSuccess
     let outputArr = cast[ptr array[32, byte]](outputBuffer.`ptr`)[]
     check:
-      "efb8ac39dc22eaf377fe85b405b99ba78dbc2f3f32494add4501741df946bd1d" ==
+      "4c6ea217404bd5f10e243bac29dc4f1ec36bf4a41caba7b4c8075c54abb3321e" ==
           outputArr.toHex()
 
     var
@@ -379,7 +379,7 @@ suite "Waku rln relay":
 
     let hash = rln.hash(msg)
     check:
-      "efb8ac39dc22eaf377fe85b405b99ba78dbc2f3f32494add4501741df946bd1d" ==
+      "4c6ea217404bd5f10e243bac29dc4f1ec36bf4a41caba7b4c8075c54abb3321e" ==
           hash.toHex()
 
   test "create a list of membership keys and construct a Merkle tree based on the list":
--- a/tests/v2/test_wakunode.nim
+++ b/tests/v2/test_wakunode.nim
@@ -790,9 +790,9 @@ procSuite "WakuNode":
       await node1.publish(rlnRelayPubSubTopic, message)
       await sleepAsync(2000.millis)
 
-
+      ## Current circom takes more time to load the circuit..
       check:
-        (await completionFut.withTimeout(10.seconds)) == true
+        (await completionFut.withTimeout(200.seconds)) == true
 
       await node1.stop()
       await node2.stop()
@@ -893,9 +893,10 @@ procSuite "WakuNode":
       await node1.publish(rlnRelayPubSubTopic, message)
       await sleepAsync(2000.millis)
 
+      ## Current circom takes more time to load the circuit...
       check:
         # the relayHandler of node3 never gets called
-        (await completionFut.withTimeout(10.seconds)) == false
+        (await completionFut.withTimeout(200.seconds)) == false
 
       await node1.stop()
       await node2.stop()
--- a/waku/v2/protocol/waku_rln_relay/rln.nim
+++ b/waku/v2/protocol/waku_rln_relay/rln.nim
@@ -6,7 +6,7 @@ import
   os,
   waku_rln_relay_types
 
-const libPath = "vendor/rln/target/debug/"
+const libPath = "vendor/zerokit/target/debug/"
 when defined(Windows):
   const libName* = libPath / "rln.dll"
 elif defined(Linux):
@@ -17,50 +17,100 @@ elif defined(MacOsX):
 # all the following procedures are Nim wrappers for the functions defined in libName
 {.push dynlib: libName, raises: [Defect].}
 
-
 ## Buffer struct is taken from
 # https://github.com/celo-org/celo-threshold-bls-rs/blob/master/crates/threshold-bls-ffi/src/ffi.rs
 type Buffer* = object
   `ptr`*: ptr uint8
   len*: uint
 
-type Auth* = object
-  secret_buffer*: ptr Buffer
-  index*: uint
-  
 #------------------------------ Merkle Tree operations -----------------------------------------
+proc update_next_member*(ctx: ptr RLN, input_buffer: ptr Buffer): bool {.importc: "set_next_leaf".}
+## adds an element in the merkle tree to the next available position
+## input_buffer points to the id commitment byte seq
+## the return bool value indicates the success or failure of the operation
 
-proc update_next_member*(ctx: RLN[Bn256],
-                        input_buffer: ptr Buffer): bool {.importc: "update_next_member".}
+proc delete_member*(ctx: ptr RLN, index: uint): bool {.importc: "delete_leaf".}
+## index is the position of the id commitment key to be deleted from the tree
+## the deleted id commitment key is replaced with a zero leaf
+## the return bool value indicates the success or failure of the operation
 
-proc delete_member*(ctx: RLN[Bn256], index: uint): bool {.importc: "delete_member".}
+proc get_root*(ctx: ptr RLN, output_buffer: ptr Buffer): bool {.importc: "get_root".}
+## get_root populates the passed pointer output_buffer with the current tree root
+## the output_buffer holds the Merkle tree root of size 32 bytes
+## the return bool value indicates the success or failure of the operation
 
-proc get_root*(ctx: RLN[Bn256], output_buffer: ptr Buffer): bool {.importc: "get_root".}
+proc get_merkle_proof*(ctx: ptr RLN, index: uint, output_buffer: ptr Buffer): bool {.importc: "get_proof".}
+## populates the passed pointer output_buffer with the merkle proof for the leaf at position index in the tree stored by ctx
+## the output_buffer holds a serialized Merkle proof (vector of 32 bytes nodes)
+## the return bool value indicates the success or failure of the operation
+
+proc set_leaf*(ctx: ptr RLN, index: uint, input_buffer: ptr Buffer): bool {.importc: "set_leaf".}
+## sets the leaf at position index in the tree stored by ctx to the value passed by input_buffer
+## the input_buffer holds a serialized leaf of 32 bytes
+## the return bool value indicates the success or failure of the operation
+
+proc set_leaves*(ctx: ptr RLN, input_buffer: ptr Buffer): bool {.importc: "set_leaves".}
+## sets multiple leaves in the tree stored by ctx to the value passed by input_buffer
+## the input_buffer holds a serialized vector of leaves (32 bytes each)
+## leaves are set one after each other starting from index 0
+## the return bool value indicates the success or failure of the operation
+
+proc reset_tree*(ctx: ptr RLN, tree_height: uint): bool {.importc: "set_tree".}
+## resets the tree stored by ctx to the the empty tree (all leaves set to 0) of height tree_height
+## the return bool value indicates the success or failure of the operation
 
 #----------------------------------------------------------------------------------------------
+
 #-------------------------------- zkSNARKs operations -----------------------------------------
+proc key_gen*(ctx: ptr RLN, output_buffer: ptr Buffer): bool {.importc: "key_gen".}
+## generates id key and id commitment key serialized inside keypair_buffer as | id_key <32 bytes>| id_commitment_key <32 bytes> |
+## id commitment is the poseidon hash of the id key
+## the return bool value indicates the success or failure of the operation
 
-proc key_gen*(ctx: RLN[Bn256], keypair_buffer: ptr Buffer): bool {.importc: "key_gen".}
+proc generate_proof*(ctx: ptr RLN,
+                         input_buffer: ptr Buffer,
+                         output_buffer: ptr Buffer): bool {.importc: "generate_rln_proof".}
+## input_buffer has to be serialized as [ id_key<32> | id_index<8> | epoch<32> | signal_len<8> | signal<var> ]
+## output_buffer holds the proof data and should be parsed as [ proof<128> | share_y<32> | nullifier<32> | root<32> | epoch<32> | share_x<32> | rln_identifier<32> ]
+## integers wrapped in <> indicate value sizes in bytes
+## the return bool value indicates the success or failure of the operation
+ 
+proc verify*(ctx: ptr RLN,
+                       proof_buffer: ptr Buffer,
+                       proof_is_valid_ptr: ptr bool): bool {.importc: "verify_rln_proof".}
+## proof_buffer has to be serialized as [ proof<128> | share_y<32> | nullifier<32> | root<32> | epoch<32> | share_x<32> | rln_identifier<32> | signal_len<8> | signal<var> ]
+## the return bool value indicates the success or failure of the call to the verify function
+## the verification of the zk proof  is available in result_ptr, where 0 indicates success and 1 is failure
 
-proc generate_proof*(ctx: RLN[Bn256],
-                    input_buffer: ptr Buffer,
-                    output_buffer: ptr Buffer): bool {.importc: "generate_proof".}
-## input_buffer serialized as  [ id_key<32> | id_index<8> | epoch<32> | signal_len<8> | signal<var> ]
-## output_buffer holds the proof data and should be parsed as |proof<256>|root<32>|epoch<32>|share_x<32>|share_y<32>|nullifier<32>|
-## sizes are in bytes
-proc verify*(ctx: RLN[Bn256],
-            proof_buffer: ptr Buffer,
-            result_ptr: ptr uint32): bool {.importc: "verify".}
-## proof_buffer [ proof<256>| root<32>| epoch<32>| share_x<32>| share_y<32>| nullifier<32> | signal_len<8> | signal<var> ]
+proc zk_prove*(ctx: ptr RLN,
+            input_buffer: ptr Buffer,
+            output_buffer: ptr Buffer): bool {.importc: "prove".}
+## input_buffer serialized as  input_data is [ id_key<32> | id_index<8> | epoch<32> | signal_len<8> | signal<var> ]
+## output_buffer holds the proof data and should be parsed as [ proof<128> | share_y<32> | nullifier<32> | root<32> | epoch<32> | share_x<32> | rln_identifier<32> ]
+## integers wrapped in <> indicate value sizes in bytes
+## the return bool value indicates the success or failure of the operation
+
+
+
+proc zk_verify*(ctx: ptr RLN,
+             proof_buffer: ptr Buffer,
+             proof_is_valid_ptr: ptr bool): bool {.importc: "verify".}
+
 #----------------------------------------------------------------------------------------------
+
 #-------------------------------- Common procedures -------------------------------------------
+proc new_circuit*(tree_height: uint, input_buffer: ptr Buffer, ctx: ptr (ptr RLN)): bool {.importc: "new".}
+## creates an instance of rln object as defined by the zerokit RLN lib
+## merkle_depth represent the depth of the Merkle tree
+## input_buffer contains a serialization of the path where the circuit resources can be found (.r1cs, .wasm, .zkey and optionally the verification_key.json)
+## ctx holds the final created rln object
+## the return bool value indicates the success or failure of the operation
 
-proc new_circuit_from_params*(merkle_depth: uint,
-                              parameters_buffer: ptr Buffer,
-                              ctx: ptr RLN[Bn256]): bool {.importc: "new_circuit_from_params".}
-                              
-proc hash*(ctx: RLN[Bn256],
-          inputs_buffer: ptr Buffer,
-          output_buffer: ptr Buffer): bool {.importc: "signal_to_field".}
-
-{.pop.}
+proc hash*(ctx: ptr RLN,
+           input_buffer: ptr Buffer,
+           output_buffer: ptr Buffer): bool {.importc: "hash".}
+## it hashes (sha256) the plain text supplied in inputs_buffer and then maps it to a field element
+## this proc is used to map arbitrary signals to field element for the sake of proof generation
+## inputs_buffer holds the hash input as a byte seq 
+## the hash output is generated and populated inside output_buffer 
+## the output_buffer contains 32 bytes hash output 
--- a/waku/v2/protocol/waku_rln_relay/waku_rln_relay_types.nim
+++ b/waku/v2/protocol/waku_rln_relay/waku_rln_relay_types.nim
@@ -318,7 +318,7 @@ const
   # STATIC_GROUP_MERKLE_ROOT is the root of the Merkle tree constructed from the STATIC_GROUP_KEYS above
   # only identity commitments are used for the Merkle tree construction
   # the root is created locally, using createMembershipList proc from waku_rln_relay_utils module, and the result is hardcoded in here
-  STATIC_GROUP_MERKLE_ROOT* = "a1877a553eff12e1b21632a0545a916a5c5b8060ad7cc6c69956741134397b2d"
+  STATIC_GROUP_MERKLE_ROOT* = "2af1e8ca39f5b1c3b097d37c2281da47b5a9b5cdde0525fe51dd929c27cb5400"
 
 const EPOCH_UNIT_SECONDS* = float64(10) # the rln-relay epoch length in seconds
 const MAX_CLOCK_GAP_SECONDS* = 20.0 # the maximum clock difference between peers in seconds
--- a/waku/v2/protocol/waku_rln_relay/waku_rln_relay_types.nim
+++ b/waku/v2/protocol/waku_rln_relay/waku_rln_relay_types.nim
@@ -10,10 +10,8 @@ import
   ../../utils/protobuf
 
 ## Bn256 and RLN are Nim wrappers for the data types used in
-## the rln library https://github.com/kilic/rln/blob/3bbec368a4adc68cd5f9bfae80b17e1bbb4ef373/src/ffi.rs
-type Bn256* = pointer
-type RLN*[E] = pointer
-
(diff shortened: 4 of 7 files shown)


staheri14:
Thanks @s1fr0 for the PR, that's great to see that zerokit is ready! 👍 
May I ask please instead of replacing the current rln-lib, to move the zero-kit part under a compiler flag so that we can switch between the two? The reason is that rln-relay onchain testnet is almost ready for launch, and by having zero-kit integrated it becomes very hard to perform debugging during dogfooding. As soon as the dogfooding finishes and we make sure the onchain rln-relay part is solid, then we can integrate zero-kit and do another test.  
cc: @oskarth

tests/v2/test_waku_rln_relay.nim
```diff
@@ -65,33 +67,17 @@ procSuite "Waku rln relay":
 suite "Waku rln relay":
   test "key_gen Nim Wrappers":
     var
-      merkleDepth: csize_t = 32
-      # parameters.key contains the parameters related to the Poseidon hasher
-      # to generate this file, clone this repo https://github.com/kilic/rln
-      # and run the following command in the root directory of the cloned project
-      # cargo run --example export_test_keys
-      # the file is generated separately and copied here
-      parameters = readFile("waku/v2/protocol/waku_rln_relay/parameters.key")
```
oskarth, line 74: How come this isn't needed anymore?

tests/v2/test_waku_rln_relay.nim
```diff
@@ -371,7 +357,7 @@ suite "Waku rln relay":
       hashSuccess
     let outputArr = cast[ptr array[32, byte]](outputBuffer.`ptr`)[]
     check:
-      "efb8ac39dc22eaf377fe85b405b99ba78dbc2f3f32494add4501741df946bd1d" ==
+      "4c6ea217404bd5f10e243bac29dc4f1ec36bf4a41caba7b4c8075c54abb3321e" ==
```
oskarth, line 360: Any idea why these are different? (Not saying they shouldn't be, just wondering if we understand reason)

waku/v2/protocol/waku_rln_relay/rln.nim
```diff
@@ -17,72 +17,100 @@ elif defined(MacOsX):
...
 ## the return bool value indicates the success or failure of the operation
 
-proc delete_member*(ctx: RLN[Bn256], index: uint): bool {.importc: "delete_member".}
+proc delete_member*(ctx: ptr RLN, index: uint): bool {.importc: "delete_leaf".}
 ## index is the position of the id commitment key to be deleted from the tree
 ## the deleted id commitment key is replaced with a zero leaf
 ## the return bool value indicates the success or failure of the operation
 
-proc get_root*(ctx: RLN[Bn256], output_buffer: ptr Buffer): bool {.importc: "get_root".}
+proc get_root*(ctx: ptr RLN, output_buffer: ptr Buffer): bool {.importc: "get_root".}
```
oskarth, line 37: How come this is difference?

waku/v2/protocol/waku_rln_relay/rln.nim
```diff
@@ -17,72 +17,100 @@ elif defined(MacOsX):
...
 ## get_root populates the passed pointer output_buffer with the current tree root
 ## the output_buffer holds the Merkle tree root of size 32 bytes
 ## the return bool value indicates the success or failure of the operation
-## 
+
+proc get_merkle_proof*(ctx: ptr RLN, index: uint, output_buffer: ptr Buffer): bool {.importc: "get_proof".}
```
oskarth, line 42: I'd probably keep this as qualified merkle proof, as "get proof" could be a ZK proof

waku/v2/protocol/waku_rln_relay/rln.nim
```diff
@@ -17,72 +17,100 @@ elif defined(MacOsX):
...
-proc generate_proof*(ctx: RLN[Bn256],
-                    input_buffer: ptr Buffer,
-                    output_buffer: ptr Buffer): bool {.importc: "generate_proof".}
-## input_buffer serialized as  [ id_key<32> | id_index<8> | epoch<32> | signal_len<8> | signal<var> ]
-## output_buffer holds the proof data and should be parsed as |proof<256>|root<32>|epoch<32>|share_x<32>|share_y<32>|nullifier<32>|
+proc generate_proof*(ctx: ptr RLN,
+                         input_buffer: ptr Buffer,
+                         output_buffer: ptr Buffer): bool {.importc: "generate_rln_proof".}
+## input_buffer has to be serialized as [ id_key<32> | id_index<8> | epoch<32> | signal_len<8> | signal<var> ]
+## output_buffer holds the proof data and should be parsed as [ proof<128> | share_y<32> | nullifier<32> | root<32> | epoch<32> | share_x<32> | rln_identifier<32> ]
```
oskarth, line 74: How come proof size is different?I find the order of share_/share_y/nullifier a bit odd, any particular reason for changing this order?

waku/v2/protocol/waku_rln_relay/waku_rln_relay_types.nim
```diff
@@ -9,23 +9,20 @@ import
   stew/arrayops,
   ../../utils/protobuf
 
-## Bn256 and RLN are Nim wrappers for the data types used in
-## the rln library https://github.com/kilic/rln/blob/3bbec368a4adc68cd5f9bfae80b17e1bbb4ef373/src/ffi.rs
-type Bn256* = pointer
```
oskarth, line 14: How come we don't have Bn256 anymore?

waku/v2/protocol/waku_rln_relay/waku_rln_relay_types.nim
```diff
@@ -55,6 +52,8 @@ type RateLimitProof* = object
   ## nullifier enables linking two messages published during the same epoch
   ## see details in https://hackmd.io/tMTLMYmTR5eynw2lwK9n1w?view#Nullifiers
   nullifier*: Nullifier
+  ## Application specific RLN Identifier
+  rlnIdentifier*: RlnIdentifier
```
oskarth, line 56: Good call, we should probably add this to RLN relay spec @staheri14

waku/v2/protocol/waku_rln_relay/waku_rln_relay_types.nim
```diff
@@ -98,6 +97,9 @@ const
   ETH_CLIENT* = "ws://localhost:8540/"
 
 const
+  # The relative folder where the circuit, proving and verification key for RLN can be found
+  # Note that resources has to be compiled with respect to the above MERKLE_TREE_DEPTH
+  RLN_RESOURCE_FOLDER* = "vendor/zerokit/rln/resources/tree_height_" & $MERKLE_TREE_DEPTH & "/"
```
oskarth, line 102: How come we need this now?

waku/v2/protocol/waku_rln_relay/waku_rln_relay_types.nim
```diff
@@ -362,6 +368,7 @@ proc encode*(nsp: RateLimitProof): ProtoBuffer =
   output.write3(4, nsp.shareX)
   output.write3(5, nsp.shareY)
   output.write3(6, nsp.nullifier)
+  output.write3(7, nsp.rlnIdentifier)
```
oskarth, line 371: Requires spec change I believe

waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
```diff
@@ -181,14 +169,7 @@ proc appendLength*(input: openArray[byte]): seq[byte] =
     output = concat(@len, @input)
   return output
 
-proc toBuffer*(x: openArray[byte]): Buffer =
```
oskarth, line 184: Why don't we need this anymore?

waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
```diff
@@ -234,87 +216,88 @@ proc proofGen*(rlnInstance: RLN[Bn256], data: openArray[byte],
...
   ## a private proc to convert RateLimitProof and data to a byte seq
   ## this conversion is used in the proof verification proc
-  ## the order of serialization is based on https://github.com/kilic/rln/blob/7ac74183f8b69b399e3bc96c1ae8ab61c026dc43/src/public.rs#L205
-  ## [ proof<256>| root<32>| epoch<32>| share_x<32>| share_y<32>| nullifier<32> | signal_len<8> | signal<var> ]
+  ## [ proof<128> | root<32> | epoch<32> | share_x<32> | share_y<32> | nullifier<32> | rln_identifier<32> | signal_len<8> | signal<var> ]
   let lenPrefMsg = appendLength(@data)
   var proofBytes = concat(@(proof.proof),
                           @(proof.merkleRoot),
                           @(proof.epoch),
                           @(proof.shareX),
```
oskarth, line 267: This order of share x / y / nullifier makes more sense to me

oskarth (commented):
Nice one!

oskarth:
@jm-clius re gracefully integrating this option

Switching between:
- Nothing
- RLN kilic
- Zerokit RLN seems reasonable short term to me

Once we know Zerokit RLN works correctly and with OK perf we can replace kilic version

waku/v2/protocol/waku_rln_relay/rln.nim
s1fr0: So is fine like this, or you mean changing the zerokit RLN api name to get_merkle_proof too?

waku/v2/protocol/waku_rln_relay/rln.nim
s1fr0: Proof is 128 bytes since it's compressed: 254 bits for each X coordinate of the 3 proof points + 1 bit for the sign of the y coordinate = 32 bytes per coordinate x 4 coordinates (2x1 over G1, 1x2 over G2) = 128 bytes. Compressed proofs are the default returned by ark-circom create_random_proof_with_reduction and verify_proof, so to make the API the same I should have gone through a decompression and re-compression adding useless overhead. There is probably a way to directly process proofs in uncompressed form (to avoid square roots operation to recover the right y coordinate), but [here](https://github.com/status-im/nwaku/pull/1016#discussion_r905563408) we discussed how we can benchmark if this makes actually sense.

The comment is wrong with respect to the implementation: the current version is 
`[ proof<128> | root<32> | epoch<32> | share_x<32> | share_y<32> | nullifier<32> | rln_identifier<32> ]`

The reason for such order of variables was in order to follow the order of the circuit's variables, cf. https://github.com/privacy-scaling-explorations/rln/blob/616ee9b0b085bdf14e7f39df884496b8e77ddc2f/circuits/rln.circom#L5 and  https://github.com/privacy-scaling-explorations/rln/blob/616ee9b0b085bdf14e7f39df884496b8e77ddc2f/circuits/rln-base.circom#L60-L62. For serialization/deserialization purposes, this order makes more sense to me since this is the order parameters should be passed to the verify_proof function. However, ultimately, I didn't want to change too much nwaku code, so I ultimately decided the keep the previously implemented order and add rln_identifier at the end.
Can be easily changed though, since the zerokit witness structure abstracts the fields order: it is just a matter of redefining the serialization/deserialization order. 

Comment updated in https://github.com/status-im/nwaku/commit/80dddff151187047d98901d13e1e9e2c138f6b2f

waku/v2/protocol/waku_rln_relay/waku_rln_relay_types.nim
s1fr0: In zerokit RLN I used as default Bn254, the same curve used by circom for generating zkSNARK proofs. Hopefully in the next future, RLN object implementation and API can be extended so that can be instantiatied with any curve from ark-ec.

waku/v2/protocol/waku_rln_relay/waku_rln_relay_types.nim
s1fr0: In order to not have duplicates of the circuit in nwaku, I thought that it is better to point to zerokit resource folder directly. Since nwaku is calling zerokit API from a different path than the latter, the resource folders should be passed to correctly locate the circuit (i.e., this string cannot be hardcoded in zerokit)

waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
s1fr0: In order to use it in createRLNInstance, I had to move it at the beginning of the file https://github.com/status-im/nwaku/blob/765e8c5fa6457beb0b878da18a98682c935ac9ac/waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim#L30

waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
s1fr0: See https://github.com/status-im/nwaku/pull/1030/files/765e8c5fa6457beb0b878da18a98682c935ac9ac#r914918440

waku/v2/protocol/waku_rln_relay/rln.nim
oskarth: `get_merkle_proof`

waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
```diff
@@ -43,38 +50,20 @@ proc createRLNInstance*(d: int = MERKLE_TREE_DEPTH): RLNResult
...
-    parameters = readFile("waku/v2/protocol/waku_rln_relay/parameters.key")
-    pbytes = parameters.toBytes()
-    len: csize_t = uint(pbytes.len)
-    parametersBuffer = Buffer(`ptr`: addr(pbytes[0]), len: len)
-
-  # check the parameters.key is not empty
-  if (pbytes.len == 0):
-    debug "error in parameters.key"
-    return err("error in parameters.key")
+    resourcesPathBuffer = RLN_RESOURCE_FOLDER.toOpenArrayByte(0, RLN_RESOURCE_FOLDER.high).toBuffer()
```
oskarth, line 55: I'd do error handling on this, assume rln resource folder can't be found / is null

waku/v2/protocol/waku_rln_relay/waku_rln_relay_types.nim
oskarth: Hm, I don't know how exactly but I suspect this might fail when we do distribution. In general vendoring and static resources should be kept separate. Perhaps OK for initial integration where we compile stuff ourselves.

cc @jm-clius

waku/v2/protocol/waku_rln_relay/rln.nim
s1fr0: This nim wrapper is already named `get_merkle_proof` and with this name the API is used internally in nwaku. If you mean to change zerokit RLN public API name, this has to be addressed with an issue in vacp2p/zerokit.

waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
s1fr0: The error handling is done in zerokit RLN (cf. [1](https://github.com/vacp2p/zerokit/blob/9cc333811b0a44638fe5ea5ac26a11f91a15aa68/rln/src/circuit.rs#L40), [2](https://github.com/vacp2p/zerokit/blob/9cc333811b0a44638fe5ea5ac26a11f91a15aa68/rln/src/circuit.rs#L58-L60), [3](https://github.com/gakonst/ark-circom/blob/a93c8b03d4376e197ceb36aa410a8903c283dab7/src/circom/builder.rs#L28-L29)) but we might add an extra error handling layer in nwaku as well if better.

tests/v2/test_wakunode.nim
```diff
@@ -789,9 +789,9 @@ procSuite "WakuNode":
       await node1.publish(rlnRelayPubSubTopic, message)
       await sleepAsync(2000.millis)
 
-
+      ## Current circom takes more time to load the circuit..
       check:
-        (await completionFut.withTimeout(10.seconds)) == true
+        (await completionFut.withTimeout(200.seconds)) == true
```
jm-clius, line 794: Just double checking that `200` is indeed necessary. :) Any idea what the usual successful loading time is, more or less? I know the circom loading takes a long time, but waiting > 3 mins for the test to fail could be frustrating during regular CI.

jm-clius (commented):
This LGTM, but I assume that, as discussed during offsite, you want to maintain both Kilic and Zerokit implementations until after the testnet? My suggestion would be to simply split the relevant RLN modules into the "old" implementation and the "new" implementation using compiler definitions (`when defined...`). Should be a simple copy-paste operation - the result will be ugly, but we can mitigate it a bit by clearly separating the two implementations per module using comments/dividers. Hopefully this dual-impl will live only for a short time!

jm-clius:
> @jm-clius re gracefully integrating this option
> 
> Switching between:
> 
> * Nothing
> * RLN kilic
> * Zerokit RLN seems reasonable short term to me
> 
> Once we know Zerokit RLN works correctly and with OK perf we can replace kilic version

As commented above, I think separating affected modules between "old" and "new" implementations using compiler definitions seem the fastest, easiest-to-maintain option to me. Only if it's short-lived, of course. :D

s1fr0:
@jm-clius @oskarth @staheri14 what would be the default for CI then? If Kilic's RLN, zerokit integration does not get automatically tested; if, instead, by default we set zerokit, then this might create problems in next stages of RLN testing.

For this reason I think the best would be to have a subfolder `zerokit` in `waku/v2/protocol/waku_rln_relay` with the appropriate edited sources for zerokit and add a `when defined(zerokit)` in `all_tests_v2.nim` to load a separate `test_waku_rln_relay_zerokit.nim` file.

s1fr0:
Pushed commits to add a compiler flag `RLNZEROKIT` to switch among Kilic's RLN (`RLN=true`) and Zerokit RLN (`RLNZEROKIT=true`). Current implementation follows the idea of my previous [comment](https://github.com/status-im/nwaku/pull/1030#issuecomment-1195721446): however, when implemented in practice, many edits where necessary to allow compilation/test of both Kilic's RLN and Zerokit RLN module in same run to avoid function/type redefinitions. Thus, I left only Kilic's RLN to be compiled as default by CI. 

`ONCHAIN_RLN=true` tests passes too when `RLNZEROKIT=true` is set.

tests/v2/test_wakunode.nim
s1fr0: Not necessary anymore with zerokit rln perfomance [fixes](https://github.com/vacp2p/zerokit/pull/26).

oskarth:
@staheri14 would you able to review this PR? It has been awaiting your review for a while now I believe.

(I requested a review from Ksr for general code integration, assuming Hanno/Lorenzo won't be back soon)

tests/v2/test_waku_rln_relay_onchain_zerokit.nim
```diff
@@ -0,0 +1,472 @@
...
+  ./test_utils
+
+const RLNRELAY_PUBSUB_TOPIC = "waku/2/rlnrelay/proto"
+const RLNRELAY_CONTENT_TOPIC = "waku/2/rlnrelay/proto"
+
+#  contract ABI
+contract(MembershipContract):
+  proc register(pubkey: Uint256) # external payable
+  proc MemberRegistered(pubkey: Uint256, index: Uint256) {.event.}
+  # proc registerBatch(pubkeys: seq[Uint256]) # external payable
```
kaiserd, line 23: Will these be implemented later?

kaiserd (approved):
LGTM from an integration point of view. I do not have enough background to give feedback / evaluate details.

staheri14:
> @staheri14 would you able to review this PR? It has been awaiting your review for a while now I believe.
> 
> (I requested a review from Ksr for general code integration, assuming Hanno/Lorenzo won't be back soon)

Based on the conversation, I was under the impression that the PR is still in progress, doing review now.

tests/v2/test_waku_rln_relay_onchain_zerokit.nim
staheri14: Yes, but right now, there is no actual use case for them.

tests/v2/test_wakunode.nim
```diff
@@ -790,7 +790,6 @@ procSuite "WakuNode":
       await node1.publish(rlnRelayPubSubTopic, message)
       await sleepAsync(2000.millis)
 
-
```
staheri14, line 793: I think this line is removed unintentionally.

waku/v2/node/config.nim
```diff
@@ -8,9 +8,15 @@ import
...
-  ../protocol/waku_rln_relay/waku_rln_relay_types,
   ../protocol/waku_message
 
+## TODO: properly address these import once zerokit rln is merged
+when defined(rln) or (not defined(rln) and not defined(rlnzerokit)):
+  import ../protocol/waku_rln_relay/waku_rln_relay_types
+
+when defined(rlnzerokit):
+  import ../protocol/waku_rln_relay/rlnzerokit/waku_rln_relay_types
+
```
staheri14, line 19:
```suggestion
## TODO: properly address these import once zerokit rln is merged
when defined(rln):
  import ../protocol/waku_rln_relay/waku_rln_relay_types
else:
  when defined(rlnzerokit):
     import ../protocol/waku_rln_relay/rlnzerokit/waku_rln_relay_types

```

waku/v2/node/wakunode2_types.nim
```diff
@@ -6,11 +6,16 @@ import
...
+  ../protocol/waku_lightpush/waku_lightpush, 
   ./peer_manager/peer_manager,
   ./discv5/waku_discv5
   
+## TODO: properly address these import once zerokit rln is merged
+when defined(rln) or (not defined(rln) and not defined(rlnzerokit)):
+  import ../protocol/waku_rln_relay/waku_rln_relay_types
+
+when defined(rlnzerokit):
+  import ../protocol/waku_rln_relay/rlnzerokit/waku_rln_relay_types
```
staheri14, line 18:
```suggestion
## TODO: properly address these import once zerokit rln is merged
when defined(rln):
  import ../protocol/waku_rln_relay/waku_rln_relay_types
else:
  when defined(rlnzerokit):
     import ../protocol/waku_rln_relay/rlnzerokit/waku_rln_relay_types
```

waku/v2/protocol/waku_message.nim
```diff
@@ -12,8 +12,15 @@ import
...
-  waku_rln_relay/waku_rln_relay_types
+  ../utils/time
+
+## TODO: properly address these import once zerokit rln is merged
+when defined(rln) or (not defined(rln) and not defined(rlnzerokit)):
+  import waku_rln_relay/waku_rln_relay_types
+
+when defined(rlnzerokit):
+  import waku_rln_relay/rlnzerokit/waku_rln_relay_types
+
```
staheri14, line 23:
```suggestion
```suggestion
## TODO: properly address these import once zerokit rln is merged
when defined(rln):
  import ../protocol/waku_rln_relay/waku_rln_relay_types
else:
  when defined(rlnzerokit):
     import ../protocol/waku_rln_relay/rlnzerokit/waku_rln_relay_types
```

```

File: waku/v2/protocol/waku_rln_relay/rlnzerokit/rln_relay_contract.nim [hunk 19]
```diff
@@ -0,0 +1,107 @@
+# src: https://github.com/kilic/rlnapp/blob/master/packages/contracts/contracts/RLN.sol
```
staheri14, line 1: Why do we have a new file for the contract?

waku/v2/protocol/waku_rln_relay/rlnzerokit/waku_rln_relay_types.nim
```diff
@@ -0,0 +1,376 @@
+when defined(rlnzerokit):
```
staheri14, line 1: Why do we have a new file for the rln_relay_types?

waku/v2/protocol/waku_rln_relay/rlnzerokit/waku_rln_relay_utils.nim
```diff
@@ -0,0 +1,841 @@
+{.push raises: [Defect].}
```
staheri14, line 1: There is no need for a duplicate of waku_rln_relay_utils, you can use the existing one

waku/v2/protocol/waku_rln_relay/waku_rln_relay_types.nim
```diff
@@ -1,368 +1,370 @@
-{.push raises: [Defect].}
+when defined(rln) or (not defined(rln) and not defined(rlnzerokit)):
```
staheri14, line 1: Why another file for the waku_rln_types?

staheri14 (commented):
I left my comments, the major issue is with all the files duplicated from the original rln-relay module which is unnecessary. The only thing that needs to be duplicated (if any) is the rln.lib which contains the binding for a different c lib.

waku/v2/protocol/waku_message.nim
staheri14: Such separation is not needed when the same file is used.

waku/v2/node/wakunode2_types.nim
staheri14: Such separation is not needed when the same file is used.

waku/v2/node/config.nim
staheri14: Such separation is not needed when the same file is used.

waku/v2/protocol/waku_rln_relay/rlnzerokit/waku_rln_relay_utils.nim
staheri14: The RLN construct definition should be moved to the rln.nim module and then all the imports of rln.nim module should be managed under a compiler flag.

waku/v2/protocol/waku_rln_relay/rlnzerokit/waku_rln_relay_types.nim
staheri14: see https://github.com/status-im/nwaku/pull/1030#discussion_r936098900

waku/v2/protocol/waku_rln_relay/rlnzerokit/rln_relay_contract.nim
staheri14: Please see https://github.com/status-im/nwaku/pull/1030#discussion_r936098900

waku/v2/node/config.nim
staheri14: Once we remove duplicate waku-rln-relay-types.nim files, then there is no need for conditional import.

waku/v2/node/wakunode2_types.nim
staheri14: Once we remove duplicate waku-rln-relay-types.nim files, then there is no need for conditional import.

waku/v2/protocol/waku_message.nim
staheri14: Ditto, Once we remove duplicate waku-rln-relay-types.nim files, then there is no need for conditional import.

staheri14 (changes requested):
As mentioned before, there are unnecessary duplicates of the original `waku_rln_relay` module which should be removed. This can be addressed by moving the RLN construct definition  to the `rln.nim` module and then managing all the imports of `rln.nim` module under a compiler flag.

staheri14:
> @staheri14 This would not suffice and nonetheless differs from current implementation. Even by moving the new RLN object definition in rln.nim, there are many differences that would require as many `when defined` here and there to make it work. This translate to lots of effort to remove all such conditions when we finally switch to zerokit. By keeping them in a separate folder, although much code is redundant, it will result in just a move of files out of the rlnzerokit folder and remove of the when defined conditions in header of changed files. Hopefully, both kilic and zerokit module are expected to co-exists in nwaku for a short period only.

@s1fr0 
This is an anti-pattern that violates the DRY principle (Don't Repeat Yourself).
This change is cloning the full copy of the library for just modifying a few parts. This damages the flexibility, paintability, and useability of the code over time. This change also compromises the integrity of the code as we need to manually ensure that all future changes are reflected in both the original module as well as its clone. This is exactly why compiler flags are there in the first place to avoid such damage to our codebase. Hence, please stick to the best practice principles to avoid introducing such antipatterns to our code base and persist its flexibility, mailability, and reusability. Also, nothing guarantees how long this transient period is going to take so it is more important to keep the master branch in compliance with the engineering best practices to avoid any unpredicted damages to the integrity.

oskarth:
> Based on the conversation, I was under the impression that the PR is still in progress. Doing review now.

From what I can tell the PR has been waiting for reviews since here https://github.com/status-im/nwaku/pull/1030#issuecomment-1196049783

---

> By keeping them in a separate folder, although much code is redundant, it will result in just a move of files out of the rlnzerokit folder and remove of the when defined conditions in header of changed files. Hopefully, both kilic and zerokit module are expected to co-exists in nwaku for a short period only.

Agree, this was discussed during the offsite and this is a short-lived change. The only reason not to replace it completely in PR was to not delay the testnet with existing kilic implementation as you suggested @staheri14.

>  This is an anti-pattern that violates the DRY principle (Don't Repeat Yourself).

While this is generally true, there are good exceptions.

1) If it is temporary it is often cleaner and easier to just copy paste and make existing modifications
2) If N=2 usually it doesn't make sense to generalize yet, because you don't know the parameter space so you won't know what patterns you are repeating. In this case I hope we aren't going to introduce a third library that implements RLN, but rather reduce to N=1 over next few weeks.

---

What are the minimal changes, if any, we can make to this PR to meet the following objectives?

1) Doesn't break existing code and delay testnet
2) Makes it easy to switch to Zerokit version and matches, roughly, the existing API?

oskarth:
We also need an issue toggling on this version of Zerokit by default, and some acceptance criteria such as

- no regressions (API works, no insane performance diff)
- dynamic testnet live (? - needs milestone as well)

The sooner we make the shift the better

staheri14:
@oskarth 
- In addition to my previous concerns, by introducing a duplicate module, we will lose the entire commit history of the original `waku_rln_relay` module and its submodules. 
- I again insist that any bug/issues that will be found during dogfooding need to be fixed in two files, in the original and the copy, and it doubles the work and increases the chances of error. 
- I still think that it is a straightforward change, by 1) finding the difference between the APIs of `rlnzerokit/rln.nim` and the original `rln.nim` (Which seem to be only `new_circuit`/`new_circuit_from_params` and `verify`) and 2) finding any caller procs that use those different APIs (which seem to be only `createRLNInstance` and `proofVerify` 3)  make those caller procs conditional to the compiler flags. The `RLN` type definition also should be replaced by the original one i.e. ` type RLN*[E] = pointer` so that the two APIs have identical proc signatures.

Having all this said, if this is the group's decision, although I disagree, I'll commit to it.

oskarth:
> In addition to my previous concerns, by introducing a duplicate module, we will lose the entire commit history of the original waku_rln_relay module and its submodules.

The commits are still there. It is true it makes it slightly more annoying if you are trying to do `git bisect` or something like that, but this is true whenever code is re-structured. I really don't see how this is a major problem.

>   I again insist that any bug/issues that will be found during dogfooding need to be fixed in two files, in the original and the copy, and it doubles the work and increases the chances of error.

That's fair, but I'd expect:

1) Most major bugs have been dealt with before testnet starts, and that we are fairly confident in that it'll work (albeit maybe be slow, etc) - please correct me if I'm wrong

2) This not being a big problem, it is just copy pasting small tweaks in two places. If we really think this is a big problem, I think it's OK to fix bugs only in the "live" version for dogfooding critical bugs. Then in the PR where we enable Zerokit by default, we just make sure those bug fixes are incorporated. Porting bug fixes is usually very quick.

> I still think that it is a straightforward change, by 1) finding the difference between the APIs of rlnzerokit/rln.nim and the original rln.nim (Which seem to be only new_circuit/new_circuit_from_params and verify) and 2) finding any caller procs that use those different APIs (which seem to be only createRLNInstance and proofVerify) 3) make those caller procs conditional to the compiler flags. The RLN type definition also should be replaced by the original one i.e. type RLN*[E] = pointer so that the two APIs have identical proc signatures.

That's a fair point. I don't have any super strong opinions here and think it is largely an implementation detail, and there are pros and cons for both. When we discussed this during the offsite, I believe the consensus was that a "flatter" compiler flag in one place was less error-prone. This especially true for temporary changes, compared to if we expected the two implementations to run concurrently for an extended period of time, in which case I'd agree with you that the "diff" should be kept to a minimum.

Additionally, I believe @jm-clius was also in favor of this flatter approach in terms of integration risk, and as the main code owner of nwaku and I believe this should have extra weight in terms of reducing integration risk (see two questions points) which is important in software engineering and code change management.

LNSD:
> Now that #1060 is ready, we can decide better which approach is more appropriate for the integration of zerokit RLN. As I said above, to avoid maintaining the PRs' branches for long with the changes in master, let's just try to agree as soon as we can on one between #1030 and #1060.

I am ok with both approaches. My concern is that we need to merge these changes as soon as possible to avoid more conflict headaches for @s1fr0.

As I commented in @kgrgpg PR, after this month's release, I would like to allocate some time to review the current RLN code. The current nim file division (utils, types, and main module) is causing some problems (e.g. merge/rebase conflicts, cycling imports, etc.)

staheri14 (approved):
@oskarth and @s1fr0 Thanks for your comments and clarification, with your suggested solutions regarding the debugging and commit history, I am convinced that we can merge this PR. 

@s1fr0 Thanks for opening the other PR, this is exactly what I meant (almost), merge whichever that you think is easier for you.

PR #1034, opened 2022-07-17
refactor(wakunode2): Split wankunode2 JSON-RPC setup to its own file

LNSD:
First of a series of code refactoring commits to split the `wakunode2.nim` file.


kaiserd (approved):
LGTM :+1: 
See comment on the other refactor PR:
The remaining migration should also get its separately named file :).

PR #1035, opened 2022-07-17
refactor(wakunode2): Split wankunode2 metrics setup to its own file

LNSD:
First of a series of code refactoring commits to split the `wakunode2.nim` file.


kaiserd (approved):
LGTM. :+1: 
If we split the setup, imo, we should also split the other functions in `wakunode2-setup-rpc` and `waku-setup-sql-migrations` respectively. 
edit: Saw part of that in another PR :sweat_smile:  . (I'd put that all in one PR)

PR #1038, opened 2022-07-20
Chore/update submodules

jm-clius:
WIP to update submodules before the next release.

Fixes so far:
- eth.keys and libp2p.crypto now use two different random number generators. This is causing some confusion as we use both libraries.
- ambiguity in SIGTERM

Fixes still required before merging:
- libp2p now requires all mounted protocols to be explicitly started and stopped

### DIFF
--- a/examples/v2/chat2.nim
+++ b/examples/v2/chat2.nim
@@ -210,9 +210,10 @@ proc publish(c: Chat, line: string) =
   when PayloadV1:
     # Use Waku v1 payload encoding/encryption
     let
+      rng = keys.newRng()
       payload = Payload(payload: chat2pb.buffer, symKey: some(c.symKey))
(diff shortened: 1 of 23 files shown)


Menduist:
> eth.keys and libp2p.crypto now use two different random number generators

nim-eth should use the same bearssl API since this PR: https://github.com/status-im/nim-eth/pull/510

jm-clius:
> nim-eth should use the same bearssl API since this PR: [status-im/nim-eth#510](https://github.com/status-im/nim-eth/pull/510)

Ah, right thanks! Turns out we're not tracking `master` branch. 😲

jm-clius:
Closing. Replaced by https://github.com/status-im/nwaku/pull/1123

Issue #1026, opened 2022-06-29
integrate the predicate into the `WHERE` clause instead of calling it on each result line


LNSD:
As per my discussions with @jm-clius and @kaiserd, seems that we have a bottleneck in the post-processing of the SQL queries. This is consuming lots of CPU cycles and blocking the other tasks.

The idea is to move this filtering logic to the SQL query itself so we do a more specific query and do not post-process lots of messages (which is potentially consuming lots of CPU cycles).

PR #1043, opened 2022-07-25
refactor(waku-store): major code reorganization, move StoreQueue to message_store folder

LNSD:
> **IMPORTANT:** Don't be scared by the number of files of this PR. Most of them are just changes to the imports 😁 

This is a major code refactoring. Actually, a code reorganization to prepare the codebase for #1026 changes.

* Reorganized the waku store files in: `rpc`/`rpc_codec`, and `protocol`.
* Expose only the desired APIs through the `waku/v2/protocol/waku_store.nim` module.
* Move `StoreQueue` message store to `node/storage/message` folder (next to the `message_store`'s module "interface")
* Move all pagination code (e.g. `PageInfo`) under the `utils/pagination.nim` module.

With this reorganization, the code dependencies should flow in a unique direction:
* `wakunode2` should depend on `protocol/waku_storage`
* `protocol/waku_store` should depend on `node/storage/messages/{message_store,waku_store_queue,waku_message_store}`


kaiserd (approved):
LGTM :sparkle: 
This makes orientation much easier :). Thank you!

This also helps in moving to a general store interface.
Would be nice to have a [single abstract store here](https://github.com/status-im/nwaku/blob/63c8035abc9283bda7dc9d23346b21f7bf279f04/waku/v2/protocol/waku_store/protocol.nim#L66), instead of one for each specific implementation as it is now. 
(This would be for a new store interface issue / PR).

PR #1055, opened 2022-08-01
refactor(waku-store): reorganise pagination test cases

LNSD:
There is some test suite grooming work pending after the previous big PR merged. This is one PR related to that:
* Move the payload-related test cases under the right test suite
* Move the message store test cases (StoreQueue) under its own test suite file
* Group the pagination utils test cases under the same test suite


kaiserd (approved):
LGTM :sparkle:

PR #979, opened 2022-05-25
WIP: split Noise submodule in smaller submodules

s1fr0:
This PR splits the current Noise module in smaller submodules, so that result easier to manage and review.

At the moment nim doesn't allow to import modules with visibility to objects' private field, hence to allow the split of the original noise module, all relevant fields and procedures had to be exposed, in contrast to the need to keep some fields private (e.g. private keys).

In the future, support to the [{.all.}](https://github.com/nim-lang/Nim/pull/17706) directive might allow import of modules with access to private fields with no need to change implemented fields' visibility.

At the moment, this branch is kept updated with changes committed in https://github.com/status-im/nwaku/pull/934 which here are expected to affect only the noise_handshake_processing submodule. The code present in all other submodules, instead, is refactored but already present in master.

Related discussion: https://github.com/status-im/nwaku/pull/934#discussion_r879439024

Issue #936, opened 2022-04-04
Noise: add padding/unpadding to ChaChaPoly encryption/decryption


s1fr0:
For security reasons and to reduce metadata leakage from ciphertext, padding should be added to encryptions done with ChaChaPoly, which is used to encrypt contents of Waku payloads with version 2.

Similarly, unpadding should be added in decryption procedures.

Impacted files include [noise.nim](https://github.com/status-im/nim-waku/blob/master/waku/v2/protocol/waku_noise/noise.nim).

PR #1056, opened 2022-08-01
feat(noise): add padding to transport messages

s1fr0:
This PR addresses https://github.com/status-im/nwaku/issues/936 by adding padding to transport messages exchanged during and after a Noise handshake.

The currently implemented padding scheme is PKCS#7 as specified in [RFC 5652](https://datatracker.ietf.org/doc/html/rfc5652#section-6.3). This padding mechanisms is well defined if and only if the padding block size is less than 256. In current PR, the default padding block size is set to be 248 bytes, hence messages will have a length which is a multiple of this value (empty messages will be padded to the set padding block size).

~~Note that this PR is based on the branch [`noise-split-module`](https://github.com/status-im/nwaku/tree/noise-split-module), whose integration is tracked in https://github.com/status-im/nwaku/pull/979.~~

~~To ease reviewing, the merge request is done over the branch `noise-split-module` as base branch. Once https://github.com/status-im/nwaku/pull/979 is closed and `noise-split-module` is merged, this PR will be edited by changing the merge request to the `master` branch.~~

### DIFF
--- a/tests/v2/test_waku_noise.nim
+++ b/tests/v2/test_waku_noise.nim
@@ -26,17 +26,17 @@ procSuite "Waku Noise":
   test "PKCS#7 Padding/Unpadding":
 
     # We test padding for different message lengths
-    let maxMessageLength = 3 * PaddingBlockSize
+    let maxMessageLength = 3 * NoisePaddingBlockSize
     for messageLen in 0..maxMessageLength:
         
       let
         message = randomSeqByte(rng[], messageLen)
-        padded = pkcs7_pad(message, PaddingBlockSize)
-        unpadded = pkcs7_unpad(padded, PaddingBlockSize)
+        padded = pkcs7_pad(message, NoisePaddingBlockSize)
+        unpadded = pkcs7_unpad(padded, NoisePaddingBlockSize)
 
       check:
         padded.len != 0
-        padded.len mod PaddingBlockSize == 0
+        padded.len mod NoisePaddingBlockSize == 0
         message == unpadded
 
   test "ChaChaPoly Encryption/Decryption: random byte sequences":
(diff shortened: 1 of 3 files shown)


waku/v2/protocol/waku_noise/noise_utils.nim
```diff
@@ -33,6 +33,29 @@ proc randomSeqByte*(rng: var BrHmacDrbgContext, size: int): seq[byte] =
   brHmacDrbgGenerate(rng, output)
   return output
 
+# Pads a payload according to PKCS#7 as per RFC 5652 https://datatracker.ietf.org/doc/html/rfc5652#section-6.3
+proc pkcs7_pad*(payload: seq[byte], paddingSize: int): seq[byte] =
```
staheri14, line 37: How about setting the default value of the `paddingSize` parameter to the `PaddingBlockSize`?

tests/v2/test_waku_noise.nim
```diff
@@ -23,6 +23,22 @@ procSuite "Waku Noise":
   # We initialize the RNG in std/random
   randomize()
 
+  test "PKCS#7 Padding/Unpadding":
+
+    # We test padding for different message lengths
+    let maxMessageLength = 3*PaddingBlockSize
```
staheri14, line 29:
```suggestion
    let maxMessageLength = 3 * PaddingBlockSize
```

waku/v2/protocol/waku_noise/noise_handshake_processing.nim
```diff
@@ -221,8 +221,10 @@ proc processMessagePatternPayload(hs: var HandshakeState, transportMessage: seq[
   # We decrypt the transportMessage, if any
   if reading:
     payload = hs.ss.decryptAndHash(transportMessage)
+    payload = pkcs7_unpad(payload, PaddingBlockSize)
```
staheri14, line 224: I think the padding blocksize should be specified in the Noise specification. Otherwise, how would the two ends know the padding block size?

staheri14 (approved):
LGTM!

PR #1067, opened 2022-08-08
refactor(waku-lightpush): waku lightpush protocol code reorganization

LNSD:
This is a code reorganization to facilitate the integration with the REST API:

* Reorganize and divide the `waku_lightpush*.nim` files.
* Create entry module `protocols/waku_lightpush.nim`
* Move from async+callback style to async+result style to simplify code
* Update test cases to match the new waku lightpush module layout


waku/v2/protocol/waku_lightpush/protocol.nim
```diff
@@ -0,0 +1,156 @@
...
+    waku_lightpush_errors.inc(labelValues = [decodeRpcFailure])
+    return err(decodeRpcFailure)
+
+  let rpcRes = res.get()
+  if rpcRes.response == PushResponse():
+    return err("empty response body")
+
+  return ok(rpcRes.response)
+
+proc request*(wl: WakuLightPush, req: PushRequest): Future[WakuLightPushResult[PushResponse]] {.async, gcsafe.} =
```
jm-clius, line 142: Yeah, much better than the callback, especially i.t.o. lightpush semantics. 👍

waku/v2/protocol/waku_lightpush/protocol.nim
```diff
@@ -0,0 +1,156 @@
...
+  let peerOpt = wl.peerManager.selectPeer(WakuLightPushCodec)
+  if peerOpt.isNone():
+    waku_lightpush_errors.inc(labelValues = [dialFailure])
+    return err(dialFailure)
+
+  return await wl.request(req, peerOpt.get())
+
+proc request*(wl: WakuLightPush, req: PushRequest, handler: PushResponseHandler) {.async, gcsafe,
```
jm-clius, line 150: I think `lightpush*()` on the Nim API now uses the deprecated method [here](https://github.com/status-im/nwaku/blob/c67c4e13546e9c5489055f07638024de8ae3cd3a/waku/v2/node/wakunode2.nim#L364-L375). We can either maintain the existing `lightpush` method on the API (with callback) and just internally use the new `request*`, or also deprecate this method and provide a new `lightpush` API call without the callback. WDYT?

jm-clius (approved):
LGTM! Thanks. I think in principle it would be good to use the new, non-deprecated `request*` method internally as well (at least in `wakunode2.nim`, tests can follow later if necessary) before merging.

waku/v2/protocol/waku_lightpush/protocol.nim
LNSD: It's a two-step process. First, refactor and mark as deprecated; then update all usages and remove the deprecated method.

I wanted to reduce the changes to the minimum amount of files to reduce the noise in the PR 😁

jm-clius: Right. Sounds reasonable then to do it in a next PR. No real concerns. Feel free to go ahead and merge this.

PR #1070, opened 2022-08-10
refactor(waku-store): inherit from message_store in store_queue

LNSD:
There is some test suite grooming work pending after the previous big PR merged. This is another PR related to that:

* Inherit from MessageStore in StoreQueue (a.k.a. in-memory store) implementation
* Unify the message store interface (add new methods that will replace the existing ones in the upcoming PRs)

This is work to accommodate the changes necessary for https://github.com/status-im/nwaku/issues/1026

### DIFF
--- a/waku/v2/node/storage/message/message_store.nim
+++ b/waku/v2/node/storage/message/message_store.nim
@@ -6,18 +6,33 @@
 import
   std/options,
   stew/results,
+  chronos
+import
   ../../../protocol/waku_message,
   ../../../protocol/waku_store/rpc,
   ../../../utils/time,
   ../../../utils/pagination
 
 
-type
-  DataProc* = proc(receiverTimestamp: Timestamp, msg: WakuMessage, pubsubTopic: string) {.closure, raises: [Defect].}
+const 
+  StoreDefaultCapacity* = 25_000
+  StoreMaxOverflow* = 1.3
+  StoreDefaultRetentionTime* = chronos.days(30).seconds
+  StoreMaxPageSize* = 100.uint64
+  StoreMaxTimeVariance* = getNanoSecondTime(20) # 20 seconds maximum allowable sender timestamp "drift" into the future
 
+
+type
   MessageStoreResult*[T] = Result[T, string]
+  
+  MessageStorePage* = (seq[WakuMessage], Option[PagingInfo])
+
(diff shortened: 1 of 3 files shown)


waku/v2/node/storage/message/message_store.nim
```diff
@@ -6,19 +6,34 @@
...
 
 
-type
-  DataProc* = proc(receiverTimestamp: Timestamp, msg: WakuMessage, pubsubTopic: string) {.closure, raises: [Defect].}
+const 
+  STORE_DEFAULT_CAPACITY* = 25_000
+  STORE_MAX_OVERFLOW* = 1.3
+  STORE_DEFAULT_RETENTION_TIME* = chronos.days(30).seconds
+  STORE_MAX_PAGE_SIZE* = 100.uint64
+  STORE_MAX_TIME_VARIANCE* = getNanoSecondTime(20) # 20 seconds maximum allowable sender timestamp "drift" into the future
```
jm-clius, line 22: There's no strong convention, and we've certainly been inconsistent here, but in general we avoid `CAPITAL_CASE` for `const`s following the Nim style guide: https://nim-lang.org/docs/nep1.html#introduction-naming-conventions

I personally prefer `PascalCase` (and I think this was the convention we converged on in nwaku code), but there is a case to be made for using `lowerCamel` (to avoid confusion with types). Also see this conversation:
https://github.com/status-im/nwaku/pull/941#discussion_r848477332

I'm OK if we extract "refactor constants to use consistent case" as a separate/follow-up task.

waku/v2/node/storage/message/waku_store_queue.nim
```diff
@@ -21,23 +21,10 @@ logScope:
...
-
-  QueryResult* = Result[uint64, string]
-
-  MessagesResult* = Result[seq[WakuMessage], string]
-
-type
-  StoreQueueResult*[T] = Result[T, cstring]
-
-  StoreQueueRef* = ref object
+  StoreQueueRef* = ref object of MessageStore
```
jm-clius, line 27: Such a big improvement to have an "interface" extracted. :)

jm-clius (approved):
Big improvement. 👍

waku/v2/node/storage/message/message_store.nim
LNSD: Ok, I will change it 😁👍🏼

PR #1073, opened 2022-08-11
fix(waku-store): integrate the predicate into the WHERE clause instead of calling it on each result line (#1026)

### LINKED ISSUE #1026: integrate the predicate into the `WHERE` clause instead of calling it on each result line

LNSD:
This PR resolves #1026:

* Decouple the SQL queries from the waku message store implementation
* Integrate the predicate into the SQLite WHERE clause instead of calling it on each result line
* Update and extend test cases coverage accordingly
* Message store interface cleanup. Remove deprecated methods
* Update waku store protocol implementation accordingly, simplifying the code


waku/v2/node/storage/message/waku_store_queue.nim
```diff
@@ -298,16 +302,8 @@ proc last*(storeQueue: StoreQueueRef): MessageStoreResult[IndexedWakuMessage] =
...
 
-## --- Queue API ---
-
-method getMostRecentMessageTimestamp*(store: StoreQueueRef): MessageStoreResult[Timestamp] =
-  let message = ?store.last()
-  ok(message.index.receiverTime)
-
-method getOldestMessageTimestamp*(store: StoreQueueRef): MessageStoreResult[Timestamp] =
-  let message = ?store.first()
-  ok(message.index.receiverTime)
```
LNSD, line 309: I added these two methods previously and they were unnecessaries. Removing them

waku/v2/node/storage/message/waku_message_store_queries.nim
```diff
@@ -0,0 +1,365 @@
...
+  ?dbStmt.execSelectMessagesWithLimitStmt(
+    contentTopic,
+    pubsubTopic,
+    cursor,
+    startTime,
+    endTime,
+    queryRowCallback
+  )
+
+  ok(messages)
```
LNSD, line 365: This part is the tricky part 🪄

LNSD, line 365: This part is the tricky part 🪄

waku/v2/node/storage/message/waku_message_store_queries.nim
```diff
@@ -0,0 +1,365 @@
...
```
jm-clius, line 365: Here is a bit of a subtle thing from the spec which is not always immediately clear: the `PagingDirection` and the sort order for the messages in the query result are not necessarily the same. In fact, the sorting order _within_ a page must always be _ascending_ according to [the spec](https://rfc.vac.dev/spec/13/).

> "the retrieved page contains an ordered list of WakuMessages from the oldest message to the most recent one."

This is also why the `StoreQueue` has this silly `reverse()` when paging backwards:
https://github.com/status-im/nwaku/blob/fd5da9e42f8a2fcf3b216540657dd947bbb7c466/waku/v2/node/storage/message/waku_store_queue.nim#L234

It causes a bit of inconvenience, but makes sense from a spec point of view as the default. It's up to applications then to modify this sort order to their own needs, but the spec "contract" is that all pages will be sorted ascending. Anyway, long story short, if the paging direction here is `BACKWARD` (i.e. `ascending == false`) I think the messages need to be sorted once more to be ASC in the returned list. For `FORWARD` pages it will already be in the correct order.

waku/v2/node/storage/message/message_store.nim
```diff
@@ -31,38 +30,11 @@ type
...
-    
-  QueryFilterMatcher* = proc(indexedWakuMsg: IndexedWakuMessage) : bool {.gcsafe, closure.}
-
 
 # MessageStore interface
-method getMostRecentMessageTimestamp*(db: MessageStore): MessageStoreResult[Timestamp] {.base.} = discard
-
-method getOldestMessageTimestamp*(db: MessageStore): MessageStoreResult[Timestamp] {.base.} = discard
-
 method put*(db: MessageStore, cursor: Index, message: WakuMessage, pubsubTopic: string): MessageStoreResult[void] {.base.} = discard
```
jm-clius, line 35: Perhaps change `db` to just `ms` in the base methods? Makes it a bit clearer that implementations of `MessageStore` are now not necessarily a database.

tests/all_tests_v2.nim
```diff
@@ -11,7 +11,8 @@ import
   ./v2/test_utils_pagination,
   ./v2/test_message_store_queue,
   ./v2/test_message_store_queue_pagination,
-  ./v2/test_message_store,
+  ./v2/test_message_store_sqlite_query,
+  ./v2/test_message_store_sqlite,
```
jm-clius, line 15: May be good (in future) to have general `MessageStore` tests that can instantiate both the `WakuMessageStore` and `StoreQueue` implementations (and whatever else we add in future) and compare query results.

waku/v2/node/storage/message/message_store.nim
LNSD: Agree. I'll change it 😁

waku/v2/node/storage/message/waku_store_queue.nim
```diff
@@ -12,16 +12,20 @@ import
...
 
-# TODO: Remove after resolving nwaku #1026
-export
-  message_store
-  
-
 logScope:
   topics = "message_store.storequeue"
 
+# TODO: Remove after resolving nwaku #1026. Move it back to waku_store_queue.nim
```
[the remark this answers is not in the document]
LNSD: Removed! 😁

waku/v2/node/storage/message/waku_message_store_queries.nim
jm-clius: Sidenote, I think that this logic in the current sqlite-only store has this bug and is why people report reverse chronological messages when connecting to the test fleet using `chat2`.

tests/v2/test_message_store_sqlite.nim
```diff
@@ -0,0 +1,329 @@
...
+  ../../waku/v2/node/storage/sqlite,
+  ../../waku/v2/protocol/waku_message,
+  ../../waku/v2/utils/time,
+  ../../waku/v2/utils/pagination,
+  ./utils
+
+
+const 
+  DEFAULT_PUBSUB_TOPIC = "/waku/2/default-waku/proto"
+  DEFAULT_CONTENT_TOPIC = ContentTopic("/waku/2/default-content/proto")
```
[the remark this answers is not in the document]
LNSD: Changed all over 🙂

tests/v2/test_message_store_sqlite.nim
```diff
@@ -0,0 +1,329 @@
...
+    let res = database.migrate(migrationPath, 10)
+    check:
+      res.isErr == false
+
+    let ver = database.getUserVersion()
+    check:
+      ver.isErr == false
+      ver.value == 10
+
+    test "number of messages retrieved by getAll is bounded by storeCapacity":
```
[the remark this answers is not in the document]
LNSD: It was indented one extra level in comparison to the other test cases. Solved 😁

waku/v2/node/storage/message/waku_message_store_queries.nim
```diff
@@ -0,0 +1,365 @@
...
+## Delete oldest messages not within limit
+
+template deleteOldestMessagesNotWithinLimitQuery*(table: string, limit: int): SqlQueryStr =
+  "DELETE FROM " & table & " WHERE id NOT IN (" &
+  " SELECT id FROM " & table & 
+  " ORDER BY receiverTimestamp DESC" &
+  " LIMIT " & $limit &
+  ");"
+
+proc deleteOldestMessagesNotWithinLimit*(db: SqliteDatabase, limit: int): DatabaseResult[void] {.inline.} =
```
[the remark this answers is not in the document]
LNSD: Comment added 👍🏼 Good point, at the very beginning it was confusing for me the "limit" word in this SQL query

tests/all_tests_v2.nim
LNSD: That sounds like a good idea to me 😁 
 
Indeed, I would like to completely decouple the WakuStore protocol from the two kinds of stores (in-memory aka queue, and the SQLite store-backed store) and just provide one ref to the protocol implementation. 

Then if we want to do some caching in-memory and back it with persistent storage for speed, what I would do is implement a wrapper method that handles that logic.


At this point you might be thinking:

> And why did you not do it yet?

I would have done it but... the "resume store" functions in the WakuStore protocol need some cleanup before starting to do some refactoring in that direction. It is not straightforward to replace it and also the interface might need some extra method to support the resume functionality.

So... one step after the other 😁

jm-clius: Hahaha. I wasn't thinking that (given the size of the open PRs), but thanks for clarifying. :D

waku/v2/node/storage/message/waku_message_store.nim
```diff
@@ -1,425 +1,265 @@
...
-  let res = db.database.query(deleteQuery, proc(s: ptr sqlite3_stmt) = discard)
-  if res.isErr:
-    return err(res.error)
-  db.numMessages = db.storeCapacity + db.deleteWindow # sqlite3 DELETE does not return the number of deleted rows; Ideally we would subtract the number of actually deleted messages. We could run a separate COUNT.
-
-  info "Oldest messages deleted from DB due to overflow.", storeCapacity=db.storeCapacity, maxStore=db.storeMaxLoad, deleteWindow=db.deleteWindow
-  when defined(debug):
-    let numMessages = messageCount(db.database).get() # requires another SELECT query, so only run in debug mode
-    debug "Number of messages left after delete operation.", messagesLeft=numMessages
+proc deleteMessagesExceedingRetentionTime(s: WakuMessageStore): MessageStoreResult[void] =
```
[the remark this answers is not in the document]
LNSD: IMHO we should not have this `isSqliteOnly` flag. The name is not descriptive enough in the context of this implementation and for me was a headache to understand how to refactor it 😅 

I would prefer having two different implementations of the `MessageStore` object that implement two different behaviors if the differences are notable or keep that flag as an SQLite message store configuration flag with a much better name.

I prefer the "two different store types" even if it means that we are duplicating some code in this case. But I kept what we had. I didn't want to diverge too much.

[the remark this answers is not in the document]
LNSD: Regarding where to update the state... Probably it's just my personal opinion/preference:
 
When I am modifying the state of an object I always try to update it at the "top-level functions" of that object (mainly, the methods that form the public API of that object). I try to avoid changing the state in helper methods because it makes it harder to follow the execution and data/state manipulation flow.

Said that, I am open to any suggestions 😁 If necessary we can iterate again and improve this implementation

waku/v2/node/storage/message/waku_message_store.nim
```diff
@@ -1,425 +1,265 @@
...
+   
+  let now = getNanosecondTime(getTime().toUnixFloat()) 
+  let retentionTimestamp = now - s.retentionTime.nanoseconds
+  let thresholdTimestamp = retentionTimestamp - s.retentionTime.nanoseconds div 10
+  if thresholdTimestamp <= s.oldestReceiverTimestamp: 
+    return ok()
 
+  s.db.deleteMessagesOlderThanTimestamp(ts=retentionTimestamp)
+
+proc deleteMessagesOverflowingTotalCapacity(s: WakuMessageStore): MessageStoreResult[void] =
```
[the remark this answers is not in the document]
LNSD: See the comments in the thread above: https://github.com/status-im/nwaku/pull/1073#discussion_r943457431

waku/v2/node/storage/message/waku_message_store_queries.nim
```diff
@@ -0,0 +1,365 @@
...
+  let dbStmt = ?db.prepareSelectMessagesWithlimitStmt(query)
+  ?dbStmt.execSelectMessagesWithLimitStmt(
+    contentTopic,
+    pubsubTopic,
+    cursor,
+    startTime,
+    endTime,
+    queryRowCallback
+  )
+
```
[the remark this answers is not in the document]
LNSD: Absolutely! Great catch 👏🏼

waku/v2/node/storage/message/waku_message_store_queries.nim
LNSD: I added the `reverse()` statement when ascending is set to `false`. And updated the test cases accordingly. Now, the issue should be solved

waku/v2/node/storage/message/waku_message_store.nim
```diff
@@ -1,425 +1,269 @@
...
+  let pagingInfo = PagingInfo(
+    pageSize: uint64(messages.len),
+    cursor: lastIndex,
+    direction: if ascendingOrder: PagingDirection.FORWARD
+               else: PagingDirection.BACKWARD
+  )
+
+  # The retrieved messages list should always be in chronological order
+  if not ascendingOrder:
+    messages.reverse()
```
LNSD, line 257: These lines, as per @jm-clius comment, should solve the message ordering issue detected during the "dog-fooding" session.

jm-clius: Great! Should fix it.

jm-clius (approved):
LGTM, thanks!

PR #1060, opened 2022-08-04
Zerokit RLN integration in nwaku: alternative approach with inline compilation flag

s1fr0:
In light of the discussion in https://github.com/status-im/nwaku/pull/1030, this PR proposes to integrate the [zerokit RLN](https://github.com/vacp2p/zerokit/tree/master/rln) module by using inline compilation flag only on part that differ from current [master](https://github.com/status-im/nwaku/commit/32f91cb26b725f686e52ac165ff3de7e032763a7), in order to ultimately allow to switch between [Kilic's RLN](https://github.com/kilic/rln) module and Zerokit RLN module employed by RLN-RELAY at compilation time.

To compile nwaku with Kilic's RLN add the usual flags: `RLN=true ONCHAIN_RLN=true`
To compile nwaku with zerokit RLN use instead: `RLNZEROKIT=true ONCHAIN_RLN=true`

PR #1077, opened 2022-08-16
fix(RLN): add support in chat2 to zerokit RLN

s1fr0:
Following https://github.com/status-im/nwaku/pull/1060, this PR integrates zerokit RLN module in `chat2`. 

It further fixes a bug arising while using the hard-coded secrets from [`STATIC_GROUP_KEYS`](https://github.com/status-im/nwaku/blob/8a14b5b2438d61f11b65ec91b0b969a36c6d8954/waku/v2/protocol/waku_rln_relay/waku_rln_relay_types.nim#L182) when generating the RLN proof attached to messages generated during [local tests](https://github.com/status-im/nwaku/blob/master/docs/tutorial/rln-chat2-local-test.md). Such values were indeed generated for Kilic's RLN module, which uses the Bn256 curve; zerokit RLN, instead, uses the Bn254 curve arithmetic, hence such secrets couldn't match as hash pre-images when using zerokit RLN tree implementation and were thus upgraded with new random values.

To compile chat2 with zerokit RLN module enabled it suffices to

`make chat2 RLNZEROKIT=true`

Then, `chat2` can be locally tested by following the corresponding [tutorial](https://github.com/status-im/nwaku/blob/master/docs/tutorial/rln-chat2-local-test.md), starting from the "Create a local network of chat2 clients" section.


staheri14 (approved):
LGTM! also followed the tutorial and it worked.

jm-clius (approved):
LGTM!

Issue #1069, opened 2022-08-10
RLN-Relay testnet feedbacks


staheri14:
Created this issue to track comments collected from on-chain rln-relay testnet.

**Write up edits:**  (corresponding PR https://github.com/status-im/nwaku/pull/1068)
- [x] (G) one screenshot has "endpoint" and "secret" in clear, while in the next one are blurred
- [x] (G) there is an extra line after the first registration command
- [x] Hiding the membership index so that users know they need to input their own index
- [x] Spacing inconsistency in the chat2 commands
- [x] Emphasize that the ethereal client address should be a web socket address 

**Rln credentials** (corresponding PR https://github.com/status-im/nwaku/pull/1080)
- [x] (R) Updating the tutorial to signify that RLN credentials are being persisted in a separate file. Otherwise, all the chat2 instances read and use the same credentials. 
- [x] (G) warn users with a message when a local credential file is loaded, specifying that no membership regsitration was executed. 
- [x] (G) Allow re-registration even when the rlncredentials.txt file is present.  Maybe via an extra parameter --force-registration=true  ((S)the cleanest path would be to just rely on one input path, either the file or cli arguments, but not both.)
- [x] (R) Allowing multiple  rlnCredentials.txt files for distinct chat2 apps running on the same machine. Potentially prefixing the filenames with the Ethereum account address (good but not sufficient)

**Chat interface:** (corresponding PR https://github.com/status-im/nwaku/pull/1082)
- [x] (R) It would be nice to see an indication that a message you are sending is considered spam. The only way to know is by opening two chat2 apps and seeing that one does not receive the messages from the other
- [x] (R) After rln-relay preparation is in progress ...,  print a link to etherscan to see the transaction details

**Error handling:**
- [x] (F) chat2 should handle the error when a http address for the Ethereum client address is passed instead of printing the unhandled exception
- [x]  ~~(F) Validate RLN credentials and let the user know if invalid.~~ no longer valid since we removed the manual rln credential config option and replaced it with a file, so the code write and reads the data to/from a file  with no user intervention
- [x] (R) Proper error handling when using an account without funds
- [x] ~~(G) Resources are not released when chat2 is terminated by `ctrl+c`, this should be fixed.~~ seems already supported

**Requested features:**
-  (G) When reusing rln credentials, user should not need to pass the ethereum address/private key. Tracking issue #1116 

# Out of scope
[...]

PR #1082, opened 2022-08-19
feat(Rln relay): Improves rln-chat2 interface, displays links to the registration tx, warns about spam messages

staheri14:
Thie PR is a progress towards https://github.com/status-im/nwaku/issues/1069
Addresses a subset pf comments re "chat interface", namely:
- Warns users when about their spamming activity
- Provides users with a link to their registration tx on Goerli testnet


LNSD:
I understand that at this moment the goal is to have a prototype where the user is using the command-line. So the comments below are just suggestions 😁 

IMO, writing to the stdout (via echo, or via any other means) should only happen from the main application, the nodewaku2 in our case. 

Can we make the "warning" an event that can be consumed in the future by other APIs (e.g., an android/ios app that wraps the nwaku implementation). In that case, you cannot use the stdout to notify the user. 

So in this case I would create a callback to the "warning event" and hook it in nodewaku2 and call `echo` inside the callback method.

jm-clius (commented):
LGTM. Feel free to mark as ready for review when it is. Agree with @LNSD that the `chat2` toy application can do with a cleanup, though this is an effort separate from RLN.

staheri14:
> IMO, writing to the stdout (via echo, or via any other means) should only happen from the main application, the nodewaku2 in our case.
> 
> Can we make the "warning" an event that can be consumed in the future by other APIs (e.g., an android/ios app that wraps the nwaku implementation). In that case, you cannot use the stdout to notify the user.
> 
> So in this case I would create a callback to the "warning event" and hook it in nodewaku2 and call `echo` inside the callback method.

I believe that the current implementation is also event-driven. The `mountRlnRelay` proc accepts a callback function to be triggered when the registration tx gets mined. The callback function is defined and passed by the chat2 user (the main program) and contains the print statement. Do you suggest something different than this?

staheri14:
@jm-clius Thanks, forgot to mark it as ready when added the reviewers. It is now marked as ready!

staheri14:
@LNSD I see, I think your suggestion is about the `publish` proc which prints a warning about spamming. I can also update the interface of that proc to accept a callback function. Already on it.

LNSD (approved):
LGTM ✅

PR #1084, opened 2022-08-22
refactor(waku-filter): waku filter protocol code reorganisation

LNSD:
This is a code reorganization to facilitate the integration with the REST API:

* Reorganize and divide the `waku_filter.nim` file
* Create entry module `protocols/waku_filter.nim`
* Move from async+callback style to async+result style to simplify code
* Update test cases accordingly


tests/v2/test_waku_filter.nim
```diff
@@ -2,286 +2,316 @@
...
+  ../test_helpers, 
+  ./utils
 
-procSuite "Waku Filter":
 
-  asyncTest "handle filter":
-    const defaultTopic = "/waku/2/default-waku/proto"
+const 
+  DEFAULT_PUBSUB_TOPIC = "/waku/2/default-waku/proto"
+  DEFAULT_CONTENT_TOPIC = ContentTopic("/waku/2/default-content/proto")
```
jm-clius, line 21: CamelCase, perhaps? Although we can re-discuss our style choice here if you want, as long as we come up with something consistent and Nim-like for `const` :)

LNSD: Oh, man... Sometimes I go in auto-pilot mode and I forget about those things, sorry 😅 Changing it!

waku/v2/protocol/waku_filter/protocol.nim
```diff
@@ -0,0 +1,282 @@
...
+  let rpc = FilterRPC(
+    requestId: id,
+    request: FilterRequest(
+      subscribe: true,
+      pubSubTopic: pubsubTopic,
+      contentFilters: contentTopics.mapIt(ContentFilter(contentTopic: it))
+    )
+  )
+
+  let res = await wf.sendFilterRpcToRemotePeer(rpc, peer)
```
jm-clius, line 242: I vaguely remember that it may not be as simple as this, but why couldn't we use `sendFilterRpcToPeer()` here and just use `peer.peerId`? IIRC the `dialPeer*()` method with only a `peerId` on the peer manager is to dial existing peers - in our case, we are guaranteed that all peers we dial here already exists. Or am I missing something?

waku/v2/protocol/waku_filter/protocol.nim
```diff
@@ -0,0 +1,282 @@
...
+  let rpc = FilterRPC(
+    requestId: id, 
+    request: FilterRequest(
+      subscribe: false,
+      pubSubTopic: pubsubTopic,
+      contentFilters: contentTopics.mapIt(ContentFilter(contentTopic: it))
+    )
+  )
+
+  let res = await wf.sendFilterRpcToRemotePeer(rpc, peer)
```
jm-clius, line 269: See comment above about using `sendFilterRpcToPeer()`

waku/v2/protocol/waku_filter/protocol.nim
```diff
@@ -0,0 +1,282 @@
...
+proc sendFilterRpcToRemotePeer(wf: WakuFilter, rpc: FilterRPC, peer: RemotePeerInfo): Future[WakuFilterResult[void]] {.async, gcsafe.}=
+  let connOpt = await wf.peerManager.dialPeer(peer, WakuFilterCodec)
+  if connOpt.isNone():
+    return err(dialFailure)
+
+  let connection = connOpt.get()
+
+  await connection.writeLP(rpc.encode().buffer)
+
+  return ok()
```
jm-clius, line 168: Couldn't we get away with removing this entire proc and only using `sendFilterRpcToPeer()` with `peerId`? PeerID should be available from wherever we want to dial and we're only dialing existing peers. Similar comments under `subscribe` and `unsubscribe` below.

jm-clius (approved):
A drastic improvement. Thanks! Clearing the filter client clutter from wakunode2, much better separation of concerns, clearer types, returns, etc. Comments below are not major:
- I think some logging disappeared which we already use quite a bit for debugging (in fact, the current implementation could probably do with more `trace` logging).
- Perhaps we can use only the `peerId` for dialing (since peers must already exist before being used in `filter`), which removes some of the redundancy when trying to accommodate dialing either on `RemotePeer` or only on `PeerId`.

waku/v2/protocol/waku_filter/protocol.nim
LNSD: I am glad that you both noticed. I also saw it, the refactoring and cleanup work has these good things. You notice that there are things that can be simplified and optimized even more. 

It took me like 1 day to notice that there are two methods called `dialPeer()` in the `peer_manager.nim` file:

* With `RemotePeerInfo` as an argument ([here](https://github.com/status-im/nwaku/blob/master/waku/v2/node/peer_manager/peer_manager.nim#L269)):

```nim
proc dialPeer*(pm: PeerManager, remotePeerInfo: RemotePeerInfo, proto: string, dialTimeout = defaultDialTimeout): Future[Option[Connection]] {.async.} =
  # Dial a given peer and add it to the list of known peers
  # @TODO check peer validity and score before continuing. Limit number of peers to be managed.
  
  # First add dialed peer info to peer store, if it does not exist yet...
  if not pm.hasPeer(remotePeerInfo.peerId, proto):
    trace "Adding newly dialed peer to manager", peerId = remotePeerInfo.peerId, addr = remotePeerInfo.addrs[0], proto = proto
    pm.addPeer(remotePeerInfo, proto)
  
  if remotePeerInfo.peerId == pm.switch.peerInfo.peerId:
    # Do not attempt to dial self
    return none(Connection)

  return await pm.dialPeer(remotePeerInfo.peerId, remotePeerInfo.addrs, proto, dialTimeout)
```

* With `PeerID` as an argument ([here](https://github.com/status-im/nwaku/blob/master/waku/v2/node/peer_manager/peer_manager.nim#L284)):

```nim
proc dialPeer*(pm: PeerManager, peerId: PeerID, proto: string, dialTimeout = defaultDialTimeout): Future[Option[Connection]] {.async.} =
  # Dial an existing peer by looking up it's existing addrs in the switch's peerStore
  # @TODO check peer validity and score before continuing. Limit number of peers to be managed.
   
  if peerId == pm.switch.peerInfo.peerId:
    # Do not attempt to dial self
    return none(Connection)

  let addrs = pm.switch.peerStore[AddressBook][peerId]
  if addrs.len == 0:
    return none(Connection)

  return await pm.dialPeer(peerId, addrs, proto, dialTimeout)
```

If you inspect the content of the methods... both methods do two different things depending on the type of the peer argument 🎉

That "method overload trap" is exactly the issue that drove me crazy during the refactoring; and why I have been saying that we need to do a review/refactoring/cleanup in the PeerManager code.

waku/v2/protocol/waku_filter/protocol.nim
LNSD: Please, check my answer to your other comment :)

waku/v2/protocol/waku_filter/protocol.nim
LNSD: Please, check my answer to your other comment :)

waku/v2/protocol/waku_filter/protocol.nim
LNSD: DISCLAIMER: I did not go into the rabbit hole to see why the `PeerId` method does not work as the other method.

rymnc (approved):

waku/v2/protocol/waku_filter/client.nim
```diff
@@ -0,0 +1,69 @@
...
+  std/[tables, sequtils],
+  chronicles
+import
+  ../waku_message,
+  ./rpc
+
+type
+  ContentFilterHandler* = proc(msg: WakuMessage) {.gcsafe, closure, raises: [Defect].}
+
+  Filter* = object
```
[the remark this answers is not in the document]
LNSD: I limited the work here to move the code here from `wakunode2.nim` file and the original `waku_filter.nim` file.

I agree that this names are confusing, I am still looking for some nems that describe better the functionality. 

IMO, this code should integrated with the `WakuFilter` structure, somehow. The protocol should be "self contained", with methods that allow using it as a subscriber and a publisher. 

This "client" part is something that I think needs a review/rework. I am planning to fix that part in the following days.

waku/v2/protocol/waku_filter/client.nim
```diff
@@ -0,0 +1,69 @@
...
+  for key, filter in filters.pairs:
+    # We do this because the key for the filter is set to the requestId received from the filter protocol.
+    # This means we do not need to check the content filter explicitly as all MessagePushs already contain
+    # the requestId of the coresponding filter.
+    if requestId != "" and requestId == key:
+      filter.handler(msg)
+      continue
+
+    # TODO: In case of no topics we should either trigger here for all messages,
+    # or we should not allow such filter to exist in the first place.
```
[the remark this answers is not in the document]
LNSD: Same as above, limited the work to move code to here. Needs to review/rework.

@jm-clius @kaiserd This was one of the "undefined behaviour" areas that I did not want to touch. If the subscription's content topics list is empty, should be push all the messages to the subscriptor? Should we define a "wildcard"? WDYT?

IMO, we should make it mandatory, together with the pubsub topic.

LNSD:
If there are no more concerns, after the GH actions pass, this PR can be merged. 

There are some minor discussions still open, but any action that comes out from there will be done in a future PR since they are out of this PR's scope.

fryorcraken:
I am happy to run this code against js-waku for added verification. Let me know when it reaches a stable stage

Issue #948, opened 2022-04-20
Waku Filter In The Browser


neekolas:
## The Problem

Waku Filter conceptually seems like an ideal protocol for browser-based environments. Unfortunately, the way the 2.0 version of the filter is specified, it does not seem to be designed with browser-based light clients in mind.

Here's my understanding of how Waku Filter works currently:
1. A client node sends a `FilterRequest` to a node that implements the Filter protocol, specifying some predicates for the filter (`topic`, `contentTopics`)
2. The node updates an in-memory mapping of Peer IDs to Subscriptions to store these predicates
3. The node closes the stream with the client
4. When a message comes in that matches the predicates, the node attempts to [dial the PeerID](https://github.com/status-im/nwaku/blob/450e0494f1cd8a4d20cc01dc4d2b7b3d3fad8988/waku/v2/protocol/waku_filter/waku_filter.nim#L247) and send the message.
5. The client node may later send another `FilterRequest` to unsubscribe, which will delete the mapping. Alternatively, if replies to the client fail consecutively over a given time period the Filter Node may automatically delete the dead client from its records.

This falls apart for browser clients at Step 4, as browsers are only capable of initiating outbound connections but not listening for inbound requests.

I discovered this when [attempting to implement Waku Filter in `js-waku`](https://github.com/status-im/js-waku/pull/711). My implementation works fine in node.js tests when the following configuration options are passed to the Waku Client, but fails otherwise. 
```ts
{ libp2p: { addresses: { listen: ["/ip4/0.0.0.0/tcp/0/ws"] } }
```
I haven't actually validated that this also fails in the browser, but my overall understanding of LibP2P tells that it should. Feel free to correct me if I'm wrong there.

## Proposed Solution
We would implement a V3 of the Waku Filter protocol that would keep the stream open upon receiving a `FilterRPC` request and send any messages that match the predicate back down the same stream.

I would be happy to put together a PR that proposes an alternate version of the Waku Filter Spec, and would be able to implement it in `go-waku` and `js-waku` if everyone was on board. I'd need someone from the Vac/Status side to take on `nwaku` support.

## Alternate solutions
- [WebRTC-star](https://github.com/libp2p/js-libp2p-webrtc-star) could solve the problem of making the client reachable, but it is not widely supported in LibP2P right now

D4nte:
> I would be happy to put together a PR that proposes an alternate version of the Waku Filter Spec,

[...]

PR #1090, opened 2022-08-25
fix: peer manager dial logic when dialing an existing peer

jm-clius:
Implements the fix by @Menduist for https://github.com/status-im/nwaku/issues/948.

The peer manager no longer expects an address entry in the `switch` peer store when dialing existing peers. In some cases, such as on existing incoming connections from browser clients, there may be no stored addresses for the peer.


LNSD (approved):
LGTM ✅

rymnc (approved):

fryorcraken:
Confirmed, it fixes the problem with js-waku :tada:

PR #1093, opened 2022-08-26
fix|feat(rln-relay|rln-chat2): Handling on-chain transactions errors and informing chat2 users

staheri14:
Covers comments under the "Error handling" section #1069 
With the changes in this PR: If any error occurs during the on-chain registration, it will be handled; the chat2 users will get informed about the cause and the installation of rln-relay gets discarded and the chat program continues with no crashing.


LNSD (approved):
LGTM ✅

jm-clius (approved):
LGTM

PR #1094, opened 2022-08-29
Periodic logging for additional metrics

### LINKED ISSUE #984: Logging: Periodic metric/info dump on running nodes
## Problem

It's difficult to glean purely from logs how healthy a `nwaku` node is running or any other information about its connections, traffic rates, subscribed topics, etc. Metrics provide better insight into node operation, but requires interpretation (preferably visualization) to be useful.

##  Suggested solution

One way in which logs can be made significantly more useful is to periodically log the most important performance and health metrics for a running `nwaku` node. We already have [optional periodic metrics logging](https://github.com/status-im/nwaku/blob/8a6b1cfba6dbddd69da56765605199c4f6a441a3/waku/v2/node/wakunode2_setup_metrics.nim#L28-L46) but this should be extended with information such as:
1. Message rate (over the last reporting period)
2. Number of connected peers
3. Number of subscribed topics
4. Number of errors (since last log)
5. etc.

## Acceptance criteria

- [ ] metrics logging enabled by default for `wakunode2` with a reasonable period (30 seconds?)
- [ ] periodically logging at least the following metrics:
  - [ ] total messages _since last report_
  - [ ] number of peers in address book
  - [ ] number of subscribed topics

rymnc:
Should resolve #984

- chore: enable metrics logging by default
- feat(logging): add additional logging metrics, change period to 30 seconds

We now log the following fields:
- Total connections initiated (peer count) since last log
- Total Messages
- Total SWAP peers
- Total FILTER peers
- Total errors since last log
- Total subscribed topics

Should the docs mention which fields refer to values over the last reporting period vs since the node setup?

### DIFF
--- a/waku/v2/node/wakunode2_setup_metrics.nim
+++ b/waku/v2/node/wakunode2_setup_metrics.nim
@@ -9,7 +9,9 @@ import
   metrics/chronos_httpserver,
   ./config,
   ./wakunode2,
-  ../protocol/waku_filter
+  ../protocol/waku_filter,
+  ../protocol/waku_store,
+  ../protocol/waku_lightpush
 
 logScope:
   topics = "wakunode.setup.metrics"
@@ -56,11 +58,13 @@ proc startMetricsLog*() =
       cumulativeErrors = totalErrors
       cumulativeConns = totalConnections
       info "Total connections initiated", count = freshConnCount
-      info "Total Messages", count = parseCollectorIntoF64(waku_node_messages)
-      info "Total SWAP peers", count = parseCollectorIntoF64(waku_swap_peers_count)
-      info "Total FILTER peers", count = parseCollectorIntoF64(waku_filter_peers)
+      info "Total messages", count = parseCollectorIntoF64(waku_node_messages)
+      info "Total swap peers", count = parseCollectorIntoF64(waku_swap_peers_count)
+      info "Total filter peers", count = parseCollectorIntoF64(waku_filter_peers)
+      info "Total store peers", count = parseCollectorIntoF64(waku_store_peers)
+      info "Total lightpush peers", count = parseCollectorIntoF64(waku_lightpush_peers)
       info "Total errors", count = freshErrorCount
-      info "Total subscribed topics", count = parseCollectorIntoF64(waku_filter_subscribers)
+      info "Total active filter subscriptions", count = parseCollectorIntoF64(waku_filter_subscribers)
 
     discard setTimer(Moment.fromNow(30.seconds), logMetrics)
   
--- a/waku/v2/node/wakunode2_setup_metrics.nim
+++ b/waku/v2/node/wakunode2_setup_metrics.nim
@@ -8,7 +8,8 @@ import
   metrics,
   metrics/chronos_httpserver,
   ./config,
-  ./wakunode2
+  ./wakunode2,
+  ../protocol/waku_filter
 
 logScope:
   topics = "wakunode.setup.metrics"
--- a/waku/v2/node/wakunode2_setup_metrics.nim
+++ b/waku/v2/node/wakunode2_setup_metrics.nim
@@ -45,12 +45,12 @@ proc startMetricsLog*() =
       # TODO: libp2p_pubsub_peers is not public, so we need to make this either
       # public in libp2p or do our own peer counting after all.
 
-      var totalErrors = parseCollectorIntoF64(waku_node_errors)
-      var totalConnections = parseCollectorIntoF64(waku_node_conns_initiated)
+      let totalErrors = parseCollectorIntoF64(waku_node_errors)
+      let totalConnections = parseCollectorIntoF64(waku_node_conns_initiated)
 
       # track cumulative, and then max.
-      var freshErrorCount = max(totalErrors - cumulativeErrors, 0)
-      var freshConnCount = max(totalConnections - cumulativeConns, 0)
+      let freshErrorCount = max(totalErrors - cumulativeErrors, 0)
+      let freshConnCount = max(totalConnections - cumulativeConns, 0)
       
       cumulativeErrors = totalErrors
       cumulativeConns = totalConnections
--- a/waku/v2/node/wakunode2_setup_metrics.nim
+++ b/waku/v2/node/wakunode2_setup_metrics.nim
@@ -24,24 +24,44 @@ proc startMetricsServer*(serverIp: ValidIpAddress, serverPort: Port) =
 
     info "Metrics HTTP server started", serverIp, serverPort
 
+proc parseCollectorIntoF64(collector: Collector): float64 = 
+  var total = 0.float64
+  for key in collector.metrics.keys():
+    try:
+      total = total + collector.value(key)
+    except KeyError:
+      discard
+  return total
 
 proc startMetricsLog*() =
   # https://github.com/nim-lang/Nim/issues/17369
   var logMetrics: proc(udata: pointer) {.gcsafe, raises: [Defect].}
 
+  var cumulativeErrors = 0.float64
+  var cumulativeConns = 0.float64
+
   logMetrics = proc(udata: pointer) =
     {.gcsafe.}:
       # TODO: libp2p_pubsub_peers is not public, so we need to make this either
       # public in libp2p or do our own peer counting after all.
-      var totalMessages = 0.float64
-      for key in waku_node_messages.metrics.keys():
-        try:
-          totalMessages = totalMessages + waku_node_messages.value(key)
-        except KeyError:
-          discard
 
-    info "Node metrics", totalMessages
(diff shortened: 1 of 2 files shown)


rymnc:
Additionally, should we log out the metrics using `metrics/chronicles_support`, as mentioned [here](https://github.com/status-im/nim-metrics/blob/master/README.md#logging), instead of casting them to float64's ourselves?

waku/v2/node/wakunode2_setup_metrics.nim
```diff
@@ -24,24 +24,44 @@ proc startMetricsServer*(serverIp: ValidIpAddress, serverPort: Port) =
...
+      info "Total Messages", count = parseCollectorIntoF64(waku_node_messages)
+      info "Total SWAP peers", count = parseCollectorIntoF64(waku_swap_peers_count)
+      info "Total FILTER peers", count = parseCollectorIntoF64(waku_filter_peers)
+      info "Total errors", count = freshErrorCount
+      info "Total subscribed topics", count = parseCollectorIntoF64(waku_filter_subscribers)
+
+    discard setTimer(Moment.fromNow(30.seconds), logMetrics)
   
-  discard setTimer(Moment.fromNow(2.seconds), logMetrics)
+  discard setTimer(Moment.fromNow(30.seconds), logMetrics)
```
LNSD, line 66: This is on nim-chronos side, but it would be great if we had an "interval timer" API like in the browser's APIs.

At first sight, I didn't get why you need 2 setTimer 😅

waku/v2/node/wakunode2_setup_metrics.nim
```diff
@@ -24,24 +24,44 @@ proc startMetricsServer*(serverIp: ValidIpAddress, serverPort: Port) =
...
-
-    info "Node metrics", totalMessages
-    discard setTimer(Moment.fromNow(2.seconds), logMetrics)
+
+      var totalErrors = parseCollectorIntoF64(waku_node_errors)
+      var totalConnections = parseCollectorIntoF64(waku_node_conns_initiated)
+
+      # track cumulative, and then max.
+      var freshErrorCount = max(totalErrors - cumulativeErrors, 0)
+      var freshConnCount = max(totalConnections - cumulativeConns, 0)
```
LNSD, line 53: If you are not modifying the variable value within the scope, please use let. It helps to know what is modified and what is not at first sight. And if you change mistakenly a `let` the compiler will warn you.

LNSD (approved):
A little nitpicking. 

LGTM ✅

waku/v2/node/wakunode2_setup_metrics.nim
rymnc: I believe the one on line 66 is used to start off the metrics logging, and the one on line 64 triggers the next log

waku/v2/node/wakunode2_setup_metrics.nim
rymnc: Will do, thanks!

rymnc: resolved in https://github.com/status-im/nwaku/pull/1094/commits/848ca0937f6de6ea7b387f1f0870c6273414cb9a

waku/v2/node/wakunode2_setup_metrics.nim
rymnc: We could have the one on line 66 immediately call into the logMetrics proc, but it wouldn't have any useful information.

Should we have the interval configurable?

rymnc: @LNSD I believe setInterval's equivalent for nim-chronos is a part of this pr - https://github.com/status-im/nim-chronos/pull/62

LNSD: Yes! That (https://github.com/status-im/nim-chronos/pull/62) was what I was referring. Seems that the issue is stale. No progress for a long time 😞

LNSD: > Should we have the interval configurable?

Well, we could do it configurable, but I don't see much value in doing it. @jm-clius WDYT?

LNSD:
Tip: Rebase your branch on top of `master` instead of merging `master` into your branch to update it with the latest changes. It makes the PRs cleaner.

waku/v2/node/wakunode2_setup_metrics.nim
jm-clius: No, no need to have this configurable at this stage. I prefer keeping configuration options as simple as possible and only making something configurable if it provides real value to an end user.

waku/v2/node/wakunode2_setup_metrics.nim
```diff
@@ -24,24 +25,44 @@ proc startMetricsServer*(serverIp: ValidIpAddress, serverPort: Port) =
...
+      # track cumulative, and then max.
+      let freshErrorCount = max(totalErrors - cumulativeErrors, 0)
+      let freshConnCount = max(totalConnections - cumulativeConns, 0)
+      
+      cumulativeErrors = totalErrors
+      cumulativeConns = totalConnections
+      info "Total connections initiated", count = freshConnCount
+      info "Total Messages", count = parseCollectorIntoF64(waku_node_messages)
+      info "Total SWAP peers", count = parseCollectorIntoF64(waku_swap_peers_count)
+      info "Total FILTER peers", count = parseCollectorIntoF64(waku_filter_peers)
```
jm-clius, line 61: For consistency, could we also log the number of `store` and `lightpush` peers?

waku/v2/node/wakunode2_setup_metrics.nim
```diff
@@ -24,24 +25,44 @@ proc startMetricsServer*(serverIp: ValidIpAddress, serverPort: Port) =
...
+      info "Total errors", count = freshErrorCount
+      info "Total subscribed topics", count = parseCollectorIntoF64(waku_filter_subscribers)
```
jm-clius, line 63:
```suggestion
      info "Total active filter subscriptions", count = parseCollectorIntoF64(waku_filter_subscribers)
```
This counts the number of subscriptions that this node services, not the number of content topics it is subscribed to.

waku/v2/node/wakunode2_setup_metrics.nim
jm-clius, line 61:
```suggestion
      info "Total messages", count = parseCollectorIntoF64(waku_node_messages)
      info "Total swap peers", count = parseCollectorIntoF64(waku_swap_peers_count)
      info "Total filter peers", count = parseCollectorIntoF64(waku_filter_peers)
```
No need to capitalise swap, filter, lightpush, store, etc. in code (we do that when referring to a full RFC name, e.g. RFC 18/WAKU2-SWAP).

jm-clius (changes requested):
Good job! Congratulations on your first PR. 🥳 I have added some minor comments on log completion, consistency, etc. If possible, could you add an example log output to this PR to get an idea of how it will look when printed to stdout?

rymnc:
Thanks for your comments, I'll get those changes in tomorrow

rymnc:
cc: @jm-clius 

- When running `wakunode2` with the default config
    image
    
- When running `wakunode2` with `--ports-shift:1 --dns-discovery:true --dns-discovery-url:enrtree://ANTL4SLG2COUILKAPE7EF2BYNL2SHSHVCHLRD5J7ZJLN5R3PRJD2Y@prod.waku.nodes.status.im --discv5-discovery:true`   
    image

waku/v2/node/wakunode2_setup_metrics.nim
rymnc: @jm-clius, in this [snippet](https://github.com/status-im/nwaku/blob/0deedf0d7ab494d41a09861b44c0e823aa2e0651/waku/v2/node/wakunode2.nim#L243-L271), the number of contentFilters is equivalent to the number of contentTopics. Can we log this out as the total subscribed topics?

jm-clius: Indeed! Not sure though if it's necessary to periodically log this, as it's (usually) unlikely to change during the lifetime of a filter client: filter clients most often start with a handful of content topic interests, subscribe to these, and don't really change their subscription over their lifetime. They'd be able to monitor this number using metrics/dashboards, but not sure how valuable it will be to repeat this number every 30s. I think we can consider this in subsequent PRs, but for now the scope of what we log periodically seems 👍 for me. WDYT?

jm-clius (approved):
LGTM, thanks! I think this can be merged :)

waku/v2/node/wakunode2_setup_metrics.nim
rymnc: Agreed! If the need arises for it, we can definitely log it in the future

PR #1095, opened 2022-08-29
fix(rest): encode waku message payload in base64

LNSD:
During Bologna's offsite, this issue was raised by @richard-ramos.

* Now the waku message payload is specified to be base64 encoded (see: [OpenAPI types](https://swagger.io/specification/#:~:text=string-,string,base64%20encoded%20characters,-string)).
* Base64 encoding wrapper.
* Encode/decode the waku messages payload when converted into/from with the Relay REST API types.
* Updated the test cases accordingly.

### DIFF
--- a/waku/v2/node/rest/relay/relay_api.yaml
+++ b/waku/v2/node/rest/relay/relay_api.yaml
@@ -118,6 +118,7 @@ components:
       properties:
         payload:
           type: string
+          format: byte
         contentTopic:
           $ref: '#/components/schemas/ContentTopic'
         version:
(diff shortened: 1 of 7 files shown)


richard-ramos (approved):

waku/v2/node/rest/relay/relay_api.yaml
```diff
@@ -118,6 +118,7 @@ components:
       properties:
         payload:
           type: string
+          format: byte
```
LNSD, line 121: As per OpenAPI spec, this means that the payload should be a base64 string.

rymnc (approved):

jm-clius (approved):
LGTM! Thanks.

PR #1104, opened 2022-08-31
chore(noise): minor refactor of default checks of keypairs

### LINKED ISSUE #1098: Minor refactor of noise's implementation
## Proposed changes

1. In the noise handshake process, the sender and receiver can only *either* read *or* write the next message. In [noise_handshake_processing.nim](https://github.com/status-im/nwaku/blob/842cfb3476e8867ef44854e5198922388b2161f3/waku/v2/protocol/waku_noise/noise_handshake_processing.nim#L34-L55), there are 2 variables, `reading`, and `writing`. This could be replaced by just one variable (or an enum), `isReading` or `isWriting`, which would avoid the usages of 

    ```nim
    elif writing:
    ```
    in many [places](https://github.com/status-im/nwaku/blob/842cfb3476e8867ef44854e5198922388b2161f3/waku/v2/protocol/waku_noise/noise_handshake_processing.nim#L140)

2. The [message direction](https://github.com/status-im/nwaku/blob/842cfb3476e8867ef44854e5198922388b2161f3/waku/v2/protocol/waku_noise/noise_types.nim#L94-L96) enum could be replaced with either `inbound` or `outbound`, making it more readable

3. Conversion of the following [conditions](https://github.com/status-im/nwaku/blob/842cfb3476e8867ef44854e5198922388b2161f3/waku/v2/protocol/waku_noise/noise_handshake_processing.nim#L389) into a function called `exists()`, which takes in the keypair, and returns if the keypair is equal to the default (for readability purposes)

## Acceptance criteria
- [ ] Reading/Writing state as one variable
- [ ] Converting D_l and D_r into inbound and outbound, respectively
- [ ] Cleaning up default keypair checks

cc: @s1fr0

rymnc:
Should resolve #1098

### DIFF
--- a/waku/v2/protocol/waku_noise/noise_utils.nim
+++ b/waku/v2/protocol/waku_noise/noise_utils.nim
@@ -59,8 +59,8 @@ proc pkcs7_unpad*(payload: seq[byte], paddingSize: int): seq[byte] =
 
 # Simple utility that checks if the given variable is "default", 
 # Therefore, it has not been initialized
-proc isDefault*[T](value: T): bool {.inline.} = 
-  value == default(T)
+proc isDefault*[T](value: T): bool = 
+  value == static(default(T))
 
 #################################################################
 
--- a/waku/v2/protocol/waku_noise/noise_utils.nim
+++ b/waku/v2/protocol/waku_noise/noise_utils.nim
@@ -411,4 +411,5 @@ proc deserializePayloadV2*(payload: seq[byte]): Result[PayloadV2, cstring]
 # Simple utility that checks if the given variable is "default", 
 # Therefore, it has not been initialized
 proc isDefault*[T](defaultyVar: T): bool = 
-  result = defaultyVar == default(T)
+  {.inline.}
+  defaultyVar == default(T)
--- a/waku/v2/protocol/waku_noise/noise_utils.nim
+++ b/waku/v2/protocol/waku_noise/noise_utils.nim
@@ -408,7 +408,7 @@ proc deserializePayloadV2*(payload: seq[byte]): Result[PayloadV2, cstring]
 
   return ok(payload2)
 
-# Simple utility that checks if the given keypair is "default", 
+# Simple utility that checks if the given variable is "default", 
 # Therefore, it has not been initialized
-proc exists*[T](keypair: T): bool = 
-  result = keypair == default(T)
+proc exists*[T](defaultyVar: T): bool = 
+  result = defaultyVar == default(T)
--- a/waku/v2/protocol/waku_noise/noise_utils.nim
+++ b/waku/v2/protocol/waku_noise/noise_utils.nim
@@ -57,6 +57,12 @@ proc pkcs7_unpad*(payload: seq[byte], paddingSize: int): seq[byte] =
   let unpadded = payload[0..payload.high-k.int]
   return unpadded
 
+# Simple utility that checks if the given variable is "default", 
+# Therefore, it has not been initialized
+proc isDefault*[T](value: T): bool = 
+  {.inline.}
+  value == default(T)
+
 #################################################################
 
 #################################
@@ -407,9 +413,3 @@ proc deserializePayloadV2*(payload: seq[byte]): Result[PayloadV2, cstring]
   i += transportMessageLen
 
   return ok(payload2)
-
-# Simple utility that checks if the given variable is "default", 
-# Therefore, it has not been initialized
-proc isDefault*[T](defaultyVar: T): bool = 
-  {.inline.}
-  defaultyVar == default(T)
--- a/waku/v2/protocol/waku_noise/noise_utils.nim
+++ b/waku/v2/protocol/waku_noise/noise_utils.nim
@@ -407,3 +407,8 @@ proc deserializePayloadV2*(payload: seq[byte]): Result[PayloadV2, cstring]
   i += transportMessageLen
 
   return ok(payload2)
+
+# Simple utility that checks if the given keypair is "default", 
+# Therefore, it has not been initialized
+proc exists*[T](keypair: T): bool = 
+  result = keypair == default(T)
--- a/waku/v2/protocol/waku_noise/noise_utils.nim
+++ b/waku/v2/protocol/waku_noise/noise_utils.nim
@@ -410,5 +410,5 @@ proc deserializePayloadV2*(payload: seq[byte]): Result[PayloadV2, cstring]
 
 # Simple utility that checks if the given variable is "default", 
 # Therefore, it has not been initialized
-proc exists*[T](defaultyVar: T): bool = 
+proc isDefault*[T](defaultyVar: T): bool = 
   result = defaultyVar == default(T)
--- a/waku/v2/protocol/waku_noise/noise_utils.nim
+++ b/waku/v2/protocol/waku_noise/noise_utils.nim
@@ -59,8 +59,7 @@ proc pkcs7_unpad*(payload: seq[byte], paddingSize: int): seq[byte] =
 
 # Simple utility that checks if the given variable is "default", 
 # Therefore, it has not been initialized
-proc isDefault*[T](value: T): bool = 
-  {.inline.}
+proc isDefault*[T](value: T): bool {.inline.} = 
   value == default(T)
 
 #################################################################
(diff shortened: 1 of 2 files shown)


waku/v2/protocol/waku_noise/noise_utils.nim
```diff
@@ -406,4 +406,9 @@ proc deserializePayloadV2*(payload: seq[byte]): Result[PayloadV2, cstring]
   payload2.transportMessage = payload[i..i+transportMessageLen-1]
   i += transportMessageLen
 
-  return ok(payload2)
\ No newline at end of file
+  return ok(payload2)
+
+# Simple utility that checks if the given variable is "default", 
+# Therefore, it has not been initialized
+proc isDefault*[T](defaultyVar: T): bool =
```
LNSD, line 413: Please, either turn this into a `{.inline.}` _proc_ or turn this into a _template_. So we ensure that this is optimized at compile time. Personally, I prefer the _template_ option.

waku/v2/protocol/waku_noise/noise_utils.nim
```diff
@@ -406,4 +406,9 @@ proc deserializePayloadV2*(payload: seq[byte]): Result[PayloadV2, cstring]
...
   i += transportMessageLen
 
-  return ok(payload2)
\ No newline at end of file
+  return ok(payload2)
+
+# Simple utility that checks if the given variable is "default", 
+# Therefore, it has not been initialized
+proc isDefault*[T](defaultyVar: T): bool = 
+  result = defaultyVar == default(T)
```
LNSD, line 414: We do not use `result`. Please, check the following Status' nim style guide entry: https://status-im.github.io/nim-style-guide/language.result.html

LNSD (changes requested):
Please, check my comments :)

waku/v2/protocol/waku_noise/noise_utils.nim
jm-clius: Nim's expression based returns should allow you to do something like

```suggestion
  defaultyVar == default(T)
```

Is it meant to be `defaultyVar` rather than `defaultVar`? :)

jm-clius (approved):
Agree with @LNSD's comments, but otherwise looks good. Once addressed and approved by him and @s1fr0, feel free to merge.

waku/v2/protocol/waku_noise/noise_utils.nim
LNSD: The rationale for using an _inline proc_ or _template_ is because if you use a normal _proc_ it is up to the compiler to optimize it. 

Using a normal proc, if it is not optimized by the nim/C compiler, may mean that you are creating a new stack frame for just doing a comparison every time you do this comparison. If you use a template/inline proc, at compile time the comparison is embedded in the statement code during the code preprocessing.

waku/v2/protocol/waku_noise/noise_utils.nim
rymnc: After seeing @s1fr0 's suggestion, replacing `defaultyVar` with `value`

waku/v2/protocol/waku_noise/noise_utils.nim
LNSD: Here a thread from the Nim forum about _inline proc vs template_: https://forum.nim-lang.org/t/3948

waku/v2/protocol/waku_noise/noise_utils.nim
```diff
@@ -57,6 +57,12 @@ proc pkcs7_unpad*(payload: seq[byte], paddingSize: int): seq[byte] =
   let unpadded = payload[0..payload.high-k.int]
   return unpadded
 
+# Simple utility that checks if the given variable is "default", 
+# Therefore, it has not been initialized
+proc isDefault*[T](value: T): bool = 
+  {.inline.}
```
LNSD, line 63: Note that the inline pragma should be between the type and the equal sign: https://nim-lang.org/docs/manual.html#:~:text=proc%20host*(s%3A%20Socket)%3A%20int%20%7B.inline.%7D%20%3D

rymnc:
Apologies, I'll do that now

waku/v2/protocol/waku_noise/noise_utils.nim
rymnc: Resolved in [587a037](https://github.com/status-im/nwaku/pull/1104/commits/587a03781ad9f38ef9d1b534294f0e87125c918f)

waku/v2/protocol/waku_noise/noise_utils.nim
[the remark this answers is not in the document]
rymnc: Resolved in [dc0bc0b](https://github.com/status-im/nwaku/pull/1104/commits/dc0bc0b83102f3d224f8f1b4e6abd64852f56ec2)

waku/v2/protocol/waku_noise/noise_utils.nim
rymnc: Resolved in [dc0bc0b](https://github.com/status-im/nwaku/pull/1104/commits/dc0bc0b83102f3d224f8f1b4e6abd64852f56ec2)

waku/v2/protocol/waku_noise/noise_utils.nim
rymnc: Resolved in [fb2b652](https://github.com/status-im/nwaku/pull/1104/commits/fb2b6528d0b1c3ea324df31e3e8c19b7498314fb)

arnetheduck: we generally avoid `inline`: https://status-im.github.io/nim-style-guide/language.inline.html (LTO is better almost always)

LNSD: FWIW @s1fr0 Check the heavy usage of the templates in _nim-stew's results_ package 😁
https://github.com/status-im/nim-stew/blob/master/stew/results.nim

arnetheduck: lol, sorry to crash the party unannounced .. that said, `proc` without inline is usually fine - LTO will inline it at the linking stage, if it's suitable for inlining (small enough and not recursive) - if I was worried about performance, I'd be more worried about `default(T)` which has to create an all-zero instance on the fly which can be slow if `T` is large - but... it's rare for performance problems to be found in these kind of functions / cases because they don't get called often enough that it matters - `proc` is the safer choice as you correctly point out above, because `template` has some gotchas, and until proven otherwise with strong reasoning or a profiler, safety is more interesting

arnetheduck: >  Check the heavy usage of the templates

so - templates are useful sometimes indeed - from a functional perspective, you get access to symbols from the instantiation context which in the case of `nim-result` is necessary to deduce the type of the special `result` variable - the other case is when dealing with return types: functions that return large types have historically been inefficient, and some of the templates in nim-result sometimes "return" large types - finally, some of the nim-result code was written pre-LTO so it still sort of lives with that legacy. 

Now, there's plenty of nuance if you go looking for it, _but_ in cases like this it wouldn't make much difference no matter what you do - compilers are pretty clever these days, and profilers are even better - more importantly though, as a matter of style, this should be a `func` to indicate that it is side-effect free - this helps future developers trust that the function won't do weird things.

arnetheduck: for default, there's a trick I think - one can do `static(default(T))` and then a constant will be generated such that the runtime zero:ing of memory is avoided

rymnc: @arnetheduck thanks for your insight, addressed in [a42e509](https://github.com/status-im/nwaku/pull/1104/commits/a42e509cf242956159133d9d38974afe6b1e72b0)

LNSD (approved):
It was a very instructive and interesting discussion 😁 

LGTM ✅

waku/v2/protocol/waku_noise/noise_utils.nim
arnetheduck: no problem :)

if you want to continue tuning the code, switching most `proc` to `func` will help the compiler help you in the future -  the next step is to not rely on `default` at all, but rather use a type like `results.Opt` which keeps track of "unset" values more clearly (as far as the reader is concerned). 

For many types of values, depending on the serialization format, all-zeroes _might_ be a valid object (a keypair is unlikely to suffer from this, but...) - `Opt` removes the risk of falling into this trap.

PR #1106, opened 2022-09-01
chore: change results error type to cstring

LNSD:
To conform with Status Nim style guide's [results rules](https://status-im.github.io/nim-style-guide/libraries.results.html?highlight=cstring#results-librariesresults:~:text=Use%20cstring%20errors%20to%20provide%20diagnostics%20without%20expectation%20of%20error%20differentiation.) I am updating all the non-complying cases in the codebase to the following rule:

> Use `cstring` errors to provide diagnostics without expectation of error differentiation.


arnetheduck:
hm, we might need to update that guide to also allow `string` - ie `cstring` is good start when you don't need to "compose" a reasonabl error string - `Result[T, string]` is one "complexity level" above on the ladder of error handling richness

LNSD:
> hm, we might need to update that guide to also allow string - ie cstring is good start when you don't need to "compose" a reasonabl error string - Result[T, string] is one "complexity level" above on the ladder of error handling richness

It makes sense 🙂 In these cases, we are using them as "constant" strings so I think we should go for the `cstring` type. WDYT @arnetheduck?

arnetheduck:
yes, it's mostly a reminder to myself that the style guide needs to evolve sometimes too ;)

Issue #1076, opened 2022-08-15
HTTP REST API: Store, Filter, Lightpush, Admin and Private APIs


jm-clius:
**Planned start date**:
**Due date**:

# Summary

We've previously added [a basic HTTP REST API](https://github.com/status-im/nwaku/issues/727) for `nwaku`, covering Relay and Debug functionality. This issue tracks the effort to add basic Store, Filter, Admin and Private HTTP REST API functionality.

# Acceptance Criteria

- An HTTP REST API for Waku v2 translated directly from the existing JSON-RPC REST API.
  + [x] Store REST API
  + [x] Filter REST API: 
    - https://github.com/status-im/nwaku/pull/1096
    - https://github.com/waku-org/nwaku/issues/1851
    - https://github.com/waku-org/nwaku/issues/1872
  + [x] Health REST API
  + [x] Lightpush REST API
  + [x] Admin REST API
  + [ ] Private REST API

## Tasks
  - [x] https://github.com/waku-org/nwaku/issues/1851
  - [x] https://github.com/waku-org/nwaku/issues/1872
  - [x] https://github.com/waku-org/nwaku/issues/1909 - Added for RLN support of Relay Rest Api
  - [x] https://github.com/waku-org/nwaku/issues/1988 - Health api 
  - [x] https://github.com/waku-org/nwaku/issues/2040
  - [x] https://github.com/waku-org/nwaku/issues/2075
  - [ ] https://github.com/waku-org/nwaku/issues/2120
  - [ ] https://github.com/waku-org/nwaku/issues/2121

# RAID (Risks, Assumptions, Issues and Dependencies)

jm-clius:
Note the issue raised in https://github.com/status-im/nwaku/issues/1032 when implementing the Admin API.

NagyZoltanPeter:
achieved: Legacy Filter - v1 - interface Rest Api support added.
next: Extend Rest Api interface for new v2 filter. Get v2 filter service supported from node.
blocker: -

NagyZoltanPeter:
**Weekly Update**

- _achieved_: Legacy Filter - v1 - interface Rest Api support added.
- _next_: Extend Rest Api interface for new v2 filter. Get v2 filter service supported from node. Add more tests.

NagyZoltanPeter:
**Weekly Update**

- _achieved_: Legacy Filter - v1 - interface Rest Api support added. V2 implementation done wait for PR review
[...]

PR #1108, opened 2022-09-01
refactor(wakunode2): deprecate non-async protocol api methods

LNSD:
Continuing with the wakunode2 file refactoring and protocol refactoring:

* Removed the deprecated methods from the protocols
* Use the async version of the deprecated methods in wakunode2
* Deprecated the old handler (non-async) methods

This is part of the groundwork necessary for #1076


jm-clius (approved):
Great to see!

PR #1119, opened 2022-09-05
feat(store): Allow messages to be marked as "don't store"

### LINKED ISSUE #985: Allow messages to be marked as "don't store"
This forms the implementation part of https://github.com/vacp2p/rfc/issues/441

## Problem description (copied from: https://github.com/vacp2p/rfc/issues/441)

Sometimes the application might produce messages that don't need to be stored, an example would be a ping to show that you are online, or for example messages that are time sensitive (say a push notification, through our current push notification system, which would be ignored if retrieved at a later date).

In those cases it would be useful to mark them as "don't store", so that they are not stored by store nodes. This would improve bandwidth performance of the overall network and of the clients, at the cost of some metadata leaking (though can be optional).

rymnc:
Should resolve #985.

Follows the RFC change described in [vacp2p/rfc](https://github.com/vacp2p/rfc/pull/532)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,6 +39,12 @@ node_modules/
 .idea/
 rlnCredentials.txt
 testPath.txt
+<<<<<<< HEAD
+
+# Nimbus Build System
+nimbus-build-system.paths
+=======
+>>>>>>> d2713745 (Revert "feat(store): replace ephemeral with ttl model")
 
 # Nimbus Build System
 nimbus-build-system.paths
--- a/.gitignore
+++ b/.gitignore
@@ -39,12 +39,6 @@ node_modules/
 .idea/
 rlnCredentials.txt
 testPath.txt
-<<<<<<< HEAD
-
-# Nimbus Build System
-nimbus-build-system.paths
-=======
->>>>>>> d2713745 (Revert "feat(store): replace ephemeral with ttl model")
 
 # Nimbus Build System
 nimbus-build-system.paths
--- a/tests/v2/test_waku_store.nim
+++ b/tests/v2/test_waku_store.nim
@@ -30,13 +30,17 @@ proc newTestDatabase(): SqliteDatabase =
 proc fakeWakuMessage(
   payload = "TEST-PAYLOAD",
   contentTopic = DefaultContentTopic, 
-  ts = getNanosecondTime(epochTime())
+  ts = getNanosecondTime(epochTime()),
+  ephemeral = false,
+  storeTTL = getNanosecondTime(60*5) # 5 minutes
 ): WakuMessage = 
   WakuMessage(
     payload: toBytes(payload),
     contentTopic: contentTopic,
     version: 1,
-    timestamp: ts
+    timestamp: ts,
+    ephemeral: ephemeral,
+    storeTTL: storeTTL
   )
 
 proc newTestSwitch(key=none(PrivateKey), address=none(MultiAddress)): Switch =
@@ -445,6 +449,49 @@ suite "Waku Store":
 
     ## Cleanup
     await allFutures(clientSwitch.stop(), serverSwitch.stop())
+
+  asyncTest "handle ephemeral messages":
+    ## Setup
+    let 
+      serverSwitch = newTestSwitch()
+      clientSwitch = newTestSwitch()
+    
+    await allFutures(serverSwitch.start(), clientSwitch.start())
+      
+    let 
+      serverProto = newTestWakuStore(serverSwitch)
+      clientProto = newTestWakuStore(clientSwitch)
+
+    clientProto.setPeer(serverSwitch.peerInfo.toRemotePeerInfo())
+
+    ## Send 5 ephemeral messages. A message is stateful by default.
+    let msgList = @[
+        fakeWakuMessage(ephemeral = true),
+        fakeWakuMessage(ephemeral = true),
+        fakeWakuMessage(ephemeral = true),
+        fakeWakuMessage(ephemeral = true),
+        fakeWakuMessage(ephemeral = true),
+      ]
+
+    for msg in msgList:
+      await serverProto.handleMessage("foo", msg)
+
+    ## Query the RPC for all messages in store. Since we are testing the behaviour of the store handler,
+    ## making the rpc should be OK.
+    let rpc = HistoryQuery(contentFilters: @[HistoryContentFilter(contentTopic: DefaultContentTopic)])
+    let resQuery = await clientProto.query(rpc)
+
+    ## Then
+    check:
+      resQuery.isOk()
+
+    # Make sure that the response has no messages
+    let response = resQuery.tryGet() 
+    check:
+      response.messages.len == 0
+
+    ## Cleanup
+    await allFutures(serverSwitch.stop(), clientSwitch.stop())
 
 
 # TODO: Review this test suite test cases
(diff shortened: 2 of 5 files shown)


fryorcraken:
Regarding (4) I agree it's invalid. In practice I'd say the message should be treated as ephemeral.

rymnc:
Thanks @fryorcraken

waku/v2/utils/protobuf.nim
```diff
@@ -17,4 +17,10 @@ proc finish3*(proto: var ProtoBuffer) =
...
 proc `==`*(a: zint64, b: zint64): bool =
-  int64(a) == int64(b)
\ No newline at end of file
+  int64(a) == int64(b)
+
+proc uintToBool*(value: uint64): bool =
+  value != 0
+
+proc boolToUint*(value: bool): uint =
+  if value: 1 else: 0
```
LNSD, line 26: I don't understand why you need this 🤔 Protocol buffers supports the boolean type: https://developers.google.com/protocol-buffers/docs/proto3#:~:text=Int64-,bool,bool,-string

waku/v2/protocol/waku_message.nim
```diff
@@ -31,6 +31,14 @@ type
...
     # XXX Experimental, this is part of https://rfc.vac.dev/spec/17/ spec and not yet part of WakuMessage spec
     proof*: RateLimitProof
+    # The ephemeral field indicates if the message should
+    # be stored. bools and uints are 
+    # equivalent in serialization of the protobuf
+    ephemeral*: bool
+    # The storeTTL field indicates for how long a given message
+    # should be considered "valid", after which it may still
+    # be fetched, but should not be used in any business logic
+    storeTTL*: Timestamp
```
LNSD, line 41: What does this timestamp represent?

 * The point in time after which the message will be no longer valid? This is what I think when I see the "timestamp" type.
 * The amount of time that the message will be valid. Then which is the starting point in time? Note that the _senderTimestamp_ field in the message is optional.

LNSD (commented):
Before jumping into implementation details, I would like to focus on having a well-thought design/specification. Then we can talk about how we should implement this feature (e.g., to minimize the impact on the store performance).

waku/v2/utils/protobuf.nim
rymnc: While protobufs do support bools, our upstream implementation (minprotobuf) in nim-libp2p does not. Upon serialisation bools and uints are the same, hence the usage of uint :)

waku/v2/protocol/waku_message.nim
rymnc: If the senderTimestamp is optional, then storeTTL can be used to indicate the timestamp after which the message is invalid. wdyt?

waku/v2/utils/protobuf.nim
rymnc: @Menduist, does it make sense to do it this way, or would we rather create the bool field upstream?

Some additional usage of bool here - https://github.com/status-im/nwaku/blob/master/waku/v2/protocol/waku_filter/rpc_codec.nim#L45-L47

jm-clius: As discussed, how we did it previously: https://github.com/status-im/nwaku/blob/master/waku/v2/protocol/waku_filter/rpc_codec.nim#L47

waku/v2/protocol/waku_message.nim
jm-clius: Agree that this should be part of a design discussion centred around RFC changes in https://github.com/vacp2p/rfc. Have discussed this out of band and will continue the conversation there once the proposed changes are ready for review.

waku/v2/utils/protobuf.nim
Menduist: You can open a PR to nim-libp2p

waku/v2/protocol/waku_message.nim
```diff
@@ -31,12 +31,16 @@ type
...
+    # be stored. bools and uints are 
+    # equivalent in serialization of the protobuf
+    ephemeral*: bool
     
    
 
 # Encoding and decoding -------------------------------------------------------
 proc init*(T: type WakuMessage, buffer: seq[byte]): ProtoResult[T] =
-  var msg = WakuMessage()
+  var msg = WakuMessage(ephemeral: false)
```
jm-clius, line 43: Is initialising the field here necessary? Empty initialisation should initialise all fields with a default value, including `false` for booleans.

jm-clius (approved):
LGTM, thanks!

richard-ramos (approved):

waku/v2/protocol/waku_message.nim
rymnc: It isn't necessary, just added it in to be explicit

jm-clius:
@LNSD if you have some time to review :)

tests/v2/test_waku_store.nim
```diff
@@ -446,6 +448,49 @@ suite "Waku Store":
     ## Cleanup
     await allFutures(clientSwitch.stop(), serverSwitch.stop())
 
+  asyncTest "handle ephemeral messages":
```
LNSD, line 451: I see what you did here 😁 

I feel that this is overkill. The goal of the test should be to test the `if msg.ephemeral` statement in the `protocol.nim` file.

What I would do is something simpler:

1. Create a `StoreQueueRef` instance.
```nim
let store = StoreQueueRef.init(...)
```
2. Create a `WakuStore` instance passing the `StoreQueueRef` instance as the `store` parameter.
```nim
let proto = WakuStore.init(..., store=store)
```
3. Handle five messages with ephemeral and non-ephemeral flag:
```nim
let msgList = @[
  fakeWakuMessage(ephemeral = false),
  fakeWakuMessage(ephemeral = true),
  fakeWakuMessage(ephemeral = true),
  fakeWakuMessage(ephemeral = true),
  fakeWakuMessage(ephemeral = false),
]

for msg in msgList:
  await proto.handleMessage(DefaultPubsubTopic, msg)
```
5. Assert that the number of messages in the `StoreQueueRef`:
```nim
check:
  store.len == 2
```

LNSD (changes requested):

LNSD:
@rymnc Please, check my comments above 🙂

tests/v2/test_waku_store.nim
rymnc: Addressed in [5c542e3](https://github.com/status-im/nwaku/pull/1119/commits/5c542e3f6b0ee7ebbbc2022e5df278dc3a5fe5cf)

.gitignore
```diff
@@ -39,6 +39,12 @@ node_modules/
 .idea/
 rlnCredentials.txt
 testPath.txt
+<<<<<<< HEAD
+
+# Nimbus Build System
+nimbus-build-system.paths
+=======
+>>>>>>> 639713f0 (Revert "feat(store): replace ephemeral with ttl model")
```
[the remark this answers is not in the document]
rymnc: Addressed in [dea7861](https://github.com/status-im/nwaku/pull/1119/commits/dea7861cf71f431e638b3e5113bd08ef3f76c596)

LNSD (approved):
LGTM ✅

Well done 😁

rymnc:
Jenkins tests pass, merging.

PR #1120, opened 2022-09-05
fix(waku-store): added an index to improve messages query time

LNSD:
This PR contains surgical changes to unblock Status's waku store testing. Some deficiencies have been identified during the analysis and will be addressed shortly.

The existing SQLite index of the messages table, `i_rt`, was not helpful for the store queries. 

- [x] Added a DB migration and increased the database schema version: 
	* Drop the previous `i_rt` index
	* Create a new index, `i_msg`, that contains the fields relevant to the store query
- [x] Update the messages index creation query

### DIFF
--- a/waku/v2/node/storage/message/waku_message_store_queries.nim
+++ b/waku/v2/node/storage/message/waku_message_store_queries.nim
@@ -76,7 +76,7 @@ proc createTable*(db: SqliteDatabase): DatabaseResult[void] {.inline.} =
 ## Create index
 
 template createIndexQuery(table: string): SqlQueryStr = 
-  "CREATE INDEX IF NOT EXISTS i_rt ON " & table & " (receiverTimestamp);"
+  "CREATE INDEX IF NOT EXISTS i_msg ON " & table & " (contentTopic, pubsubTopic, senderTimestamp, id);"
 
 proc createIndex*(db: SqliteDatabase): DatabaseResult[void] {.inline.} =
   let query = createIndexQuery(DbTable)
--- a/waku/v2/node/storage/migration/migration_types.nim
+++ b/waku/v2/node/storage/migration/migration_types.nim
@@ -7,7 +7,7 @@ const MESSAGE_STORE_MIGRATION_PATH* = sourceDir / "migrations_scripts/message"
 const PEER_STORE_MIGRATION_PATH* = sourceDir / "migrations_scripts/peer"
 const ALL_STORE_MIGRATION_PATH* = sourceDir / "migrations_scripts"
 
-const USER_VERSION* = 4 # increase this when there is an update in the database schema
+const USER_VERSION* = 5 # increase this when there is an update in the database schema
 
 type MigrationScriptsResult*[T] = Result[T, string]
 type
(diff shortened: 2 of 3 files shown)


rymnc (approved):

waku/v2/node/storage/message/waku_message_store_queries.nim
```diff
@@ -76,7 +76,7 @@ proc createTable*(db: SqliteDatabase): DatabaseResult[void] {.inline.} =
 ## Create index
 
 template createIndexQuery(table: string): SqlQueryStr = 
-  "CREATE INDEX IF NOT EXISTS i_rt ON " & table & " (receiverTimestamp);"
+  "CREATE INDEX IF NOT EXISTS i_msg ON " & table & " (contentTopic, pubsubTopic, senderTimestamp, id);"
```
felicio, line 79: Note: nice 👍.

felicio, line 79: Non-blocking:

> The queried node MUST sort the WakuMessages based on their Index, where the senderTime constitutes the most significant part and the digest comes next, and then perform pagination on the sorted result. As such, the retrieved page contains an ordered list of WakuMessages from the oldest message to the most recent one.
>
> – https://rfc.vac.dev/spec/13/

> `ReceiverTimestamp` is guaranteed to be set, while senders could omit setting `senderTimestamp`.
>
> – https://github.com/status-im/nwaku/blob/2cb7123df84b18ab7e4df0be0d77353cca4820f2/waku/v2/node/storage/message/waku_message_store.nim

1. Since the field is optional, where would messages like that end up in the resulting list? At the end?
2. In case of a null value, isn't worth falbackcing to `receiverTimestamp`, thus keeping it in the index?
3. Or shouldn't `senderTimestamp` be required?

LNSD: > 1. Since the field is optional, where would messages like that end up in the resulting list? At the end?

I am not sure about that. I need to review the current implementation.

> 2. In case of a null value, isn't worth fallbacking to _receiverTimestamp_, thus keeping it in the index?

What I am thinking is to simplify the current implementation by keeping a single column with a timestamp. 

The new column will be named `storeTimestamp` or `storedAt`. This timestamp will be set to _senderTimesamp_ if present and to the epoch time value at the moment of message insertion. The message will be dropped and not inserted into the store if the message is "too old" (older than the current time minus ~10s (?)) or "from the future" (current time plus more than ~10s (?)).

Optionally we can keep the _senderTimestamp_ column just in case we want to use it to extract metrics of the message propagation. But the reference timestamp column will be the `storeTimestamp` column for query purposes (and store restore functionality).

> 3. Or shouldn't _senderTimestamp_ be required?

IIRC For privacy-preserving reasons, the sender timestamp can be omitted. So makes no sense to make it mandatory.

@kaiserd @jm-clius  WDYT?

LNSD:
As the load tests' results indicate a substantial improvement in the query times, I merge these changes into the `master` branch. Further improvements in the waku store protocol (e.g., readability, metrics, implementation details, etc.) will be in future pull requests.

PR #1126, opened 2022-09-07
chore(wakunode): collocate wakuswitch file to the depending modules

LNSD:
Just maintenance work, as usual 🧹 Relocating a file that is not next to the module using it.


rymnc (approved):

jm-clius (approved):
🥇

PR #1128, opened 2022-09-07
chore: always include node ENR in response to info requests

jm-clius:
Blocks: https://github.com/status-im/infra-status/issues/15

Since the `node.enr` is:
- always available
- more complete than the `wakuDiscv5` enr
I believe it's sensible to always include this ENR in responses to info requests. Once [the ENRs are consolidated](https://github.com/status-im/nwaku/issues/915), we can include only the consolidated ENR here (which is likely to be `node.enr` in any case).

This is necessary to extract ENRs for DNS discovery purposes: https://github.com/status-im/infra-status/issues/15


LNSD (approved):
LGTM ✅

P.S. I love this kind of PRs 😛

Issue #1127, opened 2022-09-07
bug: CPU spikes/stalls in some deployed `nwaku` instances


jm-clius:
### Problem

One some deployed fleets, notably `status.test` CPU spikes are visible and nodes become unresponsive from time to time.

### Impact

Significant. Impact includes:
- failure to process, relay and store messages during periods of unresponsiveness
- failure to accept incoming connections

### To reproduce

Not yet reproduced locally. Visible on higher traffic fleets, such as status.test

### Screenshots/logs

Shows 90 day period, where increase in CPU usage since ~23 Aug becomes clear:

![image](https://user-images.githubusercontent.com/68783915/188914188-3ec48355-2287-4fca-b15a-0eef18f9d1e4.png)

although message rates have not changed too much:

![image](https://user-images.githubusercontent.com/68783915/188914433-1e50b567-bdf2-4a7b-87c2-16769f1aad76.png)

### nwaku version/commit hash

Reported since `v0.11`.

### Additional context

Relates to https://github.com/status-im/status-web/issues/305

fryorcraken:
proposed work around: use `status.prod` fleet.

LNSD:
@jm-clius identified yesterday that the issue is very likely to be caused by a regression introduced in #1120. The `i_rt` index was dropped. PR #1136 restores the `i_rt` index, so we expect insertion time and CPU usage to return to normal values.

LNSD:
We can conclude that the changes done in #1136 have solved the CPU spikes and have solved the regression. I proceed to resolve this issue.

PR #1136, opened 2022-09-09
fix(store): add a table index to improve message insertion time

LNSD:
This PR fixes a regression introduced in #1120, and solves the #1127 issue.

* Revert `i_rt` index. The index was dropped, translating into a heavy CPU load when a message was inserted.
* Replace `waku_message_store_queries.nim` templates with `proc` and remove `{.inline.}` as per [this discussion](https://github.com/status-im/nwaku/pull/1104#discussion_r959559319) in a previous PR and the [Status' nim styleguide](https://status-im.github.io/nim-style-guide/language.inline.html?highlight=inline#inline-functions-languageinline).
* Raise, temporarily, to debug the level of the insertion failures.

### DIFF
--- a/waku/v2/node/storage/message/waku_message_store.nim
+++ b/waku/v2/node/storage/message/waku_message_store.nim
@@ -88,15 +88,19 @@ proc init*(T: type WakuMessageStore, db: SqliteDatabase,
   
   ## Database initialization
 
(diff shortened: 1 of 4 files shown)


rymnc (approved):

jm-clius (approved):
LGTM!

PR #1141, opened 2022-09-12
refactor(store): rename to sqlite store and introduce retention policies interface

LNSD:
Before doing deeper improvements in the "message uniqueness" logic, I have refactored the `WakuMessageStore` implementation:

* Rename to `SqliteStore`. Keep the previous imports to keep the number of files in the present PR small.
* Move all implementation files under `sqlite_store` module.
* Extract `isSqliteOnly` logic into the `MessageRetentionPolicy` type/interface.
* Add Time based and Capacity based message retention policies.


waku/v2/node/storage/message/sqlite_store/retention_policy_time.nim
```diff
@@ -0,0 +1,45 @@
...
+  let now = getNanosecondTime(getTime().toUnixFloat()) 
+  let retentionTimestamp = now - p.retentionTime.nanoseconds
+  let thresholdTimestamp = retentionTimestamp - p.retentionTime.nanoseconds div 10
+
+  if thresholdTimestamp <= oldestReceiverTimestamp: 
+    return ok()
+
+  let res = db.deleteMessagesOlderThanTimestamp(ts=retentionTimestamp)
+  if res.isErr(): 
+    return err("failed to delete oldes messages: " & res.error())
```
rymnc, line 43:
```suggestion
    return err("failed to delete oldest messages: " & res.error())
```

waku/v2/node/storage/message/sqlite_store/retention_policy_capacity.nim
```diff
@@ -0,0 +1,70 @@
+{.push raises: [Defect].}
```
rymnc, line 1: This retention policy would be applicable to all types of stores that we have in the future, right?
Would it be beneficial to have this one directory up?

waku/v2/node/storage/message/sqlite_store/retention_policy_time.nim
```diff
@@ -0,0 +1,45 @@
+{.push raises: [Defect].}
```
rymnc, line 1: Same query re: other stores making use of these functions

waku/v2/node/storage/message/sqlite_store/retention_policy.nim
```diff
@@ -0,0 +1,14 @@
+{.push raises: [Defect].}
+
+import
+  stew/results
+import
+  ../../sqlite
+
+
+type RetentionPolicyResult*[T] = Result[T, string]
```
rymnc, line 9: Same query re: other stores making use of these functions and types

rymnc (approved):
LGTM, just a couple questions!

LNSD:
From @rymnc:

> This retention policy would be applicable to all types of stores that we have in the future, right?
Would it be beneficial to have this one directory up?

Note the `execute` method of `MessageRetentionPolicy` interface depends on the `db: SqliteDatabase` parameter. So it is not reusable with any other store implementation as it is.

I'd rather keep it like this because, at this moment, those retention policies are only relevant if tied to the SQLite message store implementation. We could abstract them more, but it is beyond the scope of these changes.

waku/v2/node/storage/message/sqlite_store/retention_policy.nim
LNSD: Check my comment: https://github.com/status-im/nwaku/pull/1141#issuecomment-1243721470

waku/v2/node/storage/message/sqlite_store/retention_policy_time.nim
LNSD: Check my comment: https://github.com/status-im/nwaku/pull/1141#issuecomment-1243721470

waku/v2/node/storage/message/sqlite_store/retention_policy_capacity.nim
LNSD: Check my comment: https://github.com/status-im/nwaku/pull/1141#issuecomment-1243721470

jm-clius (approved):
Yes, thanks! :)

waku/v2/node/storage/message/sqlite_store/retention_policy_capacity.nim
rymnc: Absolutely. resolving.

rymnc (approved):

PR #1145, opened 2022-09-12
fix macos builds

### LINKED ISSUE #1131: chore: Github CI failures
### Background

Github CI test workflows for nwaku have been failing since a [recent submodule update](https://github.com/status-im/nwaku/actions/runs/3008298998). This is because the tests are now so slow that they time out after 1 hour on the `macos-latest` platform and get cancelled.

### Details

There are several avenues to explore here:
- considering moving exclusively to a more configurable Jenkins CI environment (we already have the tests set up here: https://ci.status.im/job/nim-waku/job/prs/)
- improve individual test run times, remove unnecessary `sleep`s etc.
- parallelize tests (or schedule tests better)

### Acceptance criteria

- [ ] Regular CI passes for each PR

rymnc:
Should resolve #1131.

Reverts the `nim-json-rpc` submodule update, which moved from `nim-websock` to `news` between updates. Relevant [diff](https://github.com/status-im/nim-json-rpc/compare/335f292a5816910aebf215e3a88db8a665133e0e...32ba2d16b919310a8c6797ac45747595f301a207)

This regression only affects macOS.

Experimentation can be done by updating to the latest version of `nim-json-rpc`, there seems to be some work done on this [PR](https://github.com/status-im/nim-json-rpc/pull/147) to better support macOS.

### DIFF
--- a/ci/Jenkinsfile.prs
+++ b/ci/Jenkinsfile.prs
@@ -108,7 +108,7 @@ def getAgentLabel() {
         return params.AGENT_LABEL
     }
     def tokens = env.JOB_NAME.split('/')
-    for (platform in ['linux', 'macos', 'windows']) {
+    for (platform in ['linux', 'macos']) {
       if (tokens.contains(platform)) { return platform }
     }
     throw new Exception('No agent provided or found in job path!')
--- a/ci/Jenkinsfile.prs
+++ b/ci/Jenkinsfile.prs
@@ -108,7 +108,7 @@ def getAgentLabel() {
         return params.AGENT_LABEL
     }
     def tokens = env.JOB_NAME.split('/')
-    for (platform in ['linux', 'macos']) {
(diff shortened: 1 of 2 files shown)


rymnc:
cc: @LNSD @jm-clius

Is windows a target we support?

LNSD:
> cc: @LNSD @jm-clius
> 
> Is windows a target we support?

No, It was not supported (read passing) before the dependencies upgrade.

rymnc:
Can we disable it on Jenkins, then?

jm-clius:
Really good catch! The easiest workaround is therefore to downgrade nim-json-rpc for now. @LNSD wdyt? I think we can probably leave it "downgraded" for now since we're planning to deprecate `nim-json-rpc` in any case? When we do a next submodule update we can focus on this submodule specifically as there'll hopefully be an upstream fix already.

jm-clius (approved):

PR #1149, opened 2022-09-13
refactor(wakunode): remove exports from wakunode

LNSD:
Another multi-commit grooming PR. In this case, the goal is to continue removing the exports to avoid implicit and cyclic imports.

### DIFF
--- a/examples/v2/config_chat2.nim
+++ b/examples/v2/config_chat2.nim
@@ -6,8 +6,8 @@ import
   libp2p/crypto/secp,
   nimcrypto/utils,
   eth/keys,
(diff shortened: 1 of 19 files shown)


rymnc (approved):

jm-clius (approved):
LGTM. I take it the `MembershipIndex` change to `uint` for RLN config will have no impact on how that config is used.

LNSD:
> LGTM. I take it the `MembershipIndex` change to `uint` for RLN config will have no impact on how that config is used.

Yes, that's correct. No impact. It has been moved up where the config value is used to untangle the imports graph. 

As a rule of thumb, in configuration parsing/definition, we should work with "primitive" types as much as we can.

Issue #1057, opened 2022-08-02
feat: Implement 34/WAKU2-PEER-EXCHANGE


kaiserd:
This issue tracks the implemenation of  [34/WAKU2-PEER-EXCHANGE](https://rfc.vac.dev/spec/34/).
It is part of the [Ambient Peer Discovery Roadmap](https://github.com/vacp2p/research/issues/116) and the [Waku Peer Exchange Milestone](https://github.com/vacp2p/research/issues/117).

## action items

* [ ] implement [34/WAKU2-PEER-EXCHANGE](https://rfc.vac.dev/spec/34/) requester functionality
* [ ] implement [34/WAKU2-PEER-EXCHANGE](https://rfc.vac.dev/spec/34/) responder functionality
* [ ] add test cases

fryorcraken:
@kaiserd During the Vac PM call last week you mentioned that there are known limitations with the implementation that could be tackled at a later stage after some dog fooding.

Could you please update this issue with an executive summary of:
- the known limitations
- the expected/likely next steps once dogfooding is started.

PR #1152, opened 2022-09-14
feat: waku peer exchange (RFC34)

kaiserd:
This PR addresses #1057, adding a first rudimentary implementation of [34/WAKU2-PEER-EXCHANGE](https://rfc.vac.dev/spec/34/).

CC @D4nte 

### cache implementation / retrieving discv5 nodes

The current version features a simple `seq` based cache (cmp. simple ring buffer).
Once the cache size is over a threshold, the oldest elements are deleted in a batch.

The cache is filled and maintained in an async loop running discv5 (independent of the discv5 loop that retrieves peers for the node's own use, e.g. for relay.)

## open points

* remove `client.nim`?
  - for now it is just a place holder
* MaxRpcSize

## future work (follow up PR)

* introduce a peer exchange flag to [31/WAKU2-ENR](https://rfc.vac.dev/spec/31/#references)
* error handling (for now, analogous to other protocol implementations)
  - use enum error type instead of string
  - use `?` operator
* add more a sophisticated cache (see [34/WAKU2-PEER-EXCHANGE](https://rfc.vac.dev/spec/34/))
* separate requester and responder functionality
  - it should be possible, to start a request without having to mount the responder side of the protocol
    (for now, a request-only node mounts the protocol, but does not respond)
  - this would allow to start requesting nodes in `retrieveDynamicBootstrapNodes`
* check / improve anonymity properties


waku/v2/protocol/waku_peer_exchange/protocol.nim
```diff
@@ -0,0 +1,209 @@
...
+type
+  WakuPeerExchangeResult*[T] = Result[T, string]
+
+  WakuPeerExchange* = ref object of LPProtocol
+    peerManager*: PeerManager
+    wakuDiscv5: Option[WakuDiscoveryV5]
+    enrCache: seq[enr.Record] # todo: next step: ring buffer; future: implement cache satisfying https://rfc.vac.dev/spec/34/
+
+proc sendPeerExchangeRpcToPeer(wpx: WakuPeerExchange, rpc: PeerExchangeRpc, peer: RemotePeerInfo | PeerId): Future[WakuPeerExchangeResult[void]] {.async, gcsafe.} =
+  let connOpt = await wpx.peerManager.dialPeer(peer, WakuPeerExchangeCodec)
```
LNSD, line 57: I think that without #1161 changes, the `PeerId` version will fail.

waku/v2/protocol/waku_peer_exchange/protocol.nim
```diff
@@ -0,0 +1,209 @@
...
+proc getEnrsFromCache(px: WakuPeerExchange, numPeers: uint64): seq[enr.Record] {.gcsafe.} =
+  randomize()
+  if px.enrCache.len() == 0:
+    debug "peer exchange ENR cache is empty"
+    return @[]
+  for i in 0..<min(numPeers, px.enrCache.len().uint64()):
+    let ri = rand(0..<px.enrCache.len())
+    result.add(px.enrCache[ri])
+
+proc init(px: WakuPeerExchange) =
```
LNSD, line 154: I would rather rename this procedure: `initProtocolHandler(...)` to avoid mixing it with the exported `init` procedure. It is a more descriptive name.

waku/v2/node/wakunode2.nim
```diff
@@ -1131,7 +1120,12 @@ when isMainModule:
     if dynamicBootstrapNodes.len > 0:
       info "Connecting to dynamic bootstrap peers"
       waitFor connectToNodes(node, dynamicBootstrapNodes, "dynamic bootstrap")
-    
+
+    # retrieve and connect to peer exchange peers
+    if conf.pxNode != "":
+      info "Retrieving peer info via peer exchange protocol"
+      discard waitFor node.wakuPeerExchange.request(6) # todo: relay mesh degree paramter
```
LNSD, line 1127: Can you put the `6`, magic number, in a constant?

LNSD (approved):
I see some details to polish, but in general LGTM ✅

waku/v2/node/wakunode2.nim
```diff
@@ -34,15 +35,15 @@ import
   ./wakunode2_types
 
 export
-  wakunode2_types
+  wakunode2_types,
+  peer_manager
```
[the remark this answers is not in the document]
LNSD: > Another quick fix would be importing peer manager in chat2 and wakubridge.

This is what I suggest, exactly. Let's make imports explicit rather than implicit.

waku/v2/protocol/waku_peer_exchange/protocol.nim
LNSD: I think that #1161 is ready to merge, so if we merge the PR before this one, we should be good.

waku/v2/protocol/waku_peer_exchange/protocol.nim
LNSD: I am already doing it for waku store, see: 
 
https://github.com/status-im/nwaku/blob/refactor-store-protocol/waku/v2/protocol/waku_store/protocol.nim#L201

waku/v2/protocol/waku_peer_exchange/protocol.nim
jm-clius: Ah, apologies. Fixing signing on #1161 and will merge afterwards. Not sure thought that I see why those changes affect this?

waku/v2/node/config.nim
```diff
@@ -390,6 +390,17 @@ type
       defaultValue: 1
       name: "discv5-bits-per-hop" .}: int
 
+    ## waku peer exchange config
+    wakuPeerExchange* {.
+      desc: "Enable waku peer exchange protocol (responder side): true|false",
+      defaultValue: false
+      name: "waku-px" }: bool
+
+    pxNode* {.
```
jm-clius, line 399: Minor comment, but any way we can have these config items have a similar prefix? There is indeed a future refactor task to simplify and properly group config items, but at some point we started prefixing similar config items to achieve a little more order. For example we have `store`, `storeNode`, `store*`, `filter`, `filterNode`, etc. I imagine `wakuPeerExchange` is to differentiate it from `relayPeerExchange`, but perhaps the latter is already clear enough and we can just use `peerExchange` and `peerExchangeNode`?

jm-clius (approved):
LGTM!

waku/v2/protocol/waku_peer_exchange/protocol.nim
jm-clius: #1161 is merged

waku/v2/node/config.nim
jm-clius: Indeed. And there is an implied "Waku..." prefix to all our protocols (WakuStore, WakuFilter, etc.). Thanks!

Issue #946, opened 2022-04-18
Refactoring the waku-rln-relay module


staheri14:
This issue is to address a few refactoring tasks:
1) to replace all the float64 time values in the rln-relay module with int64 type to make it consistent with the rest of the nim-waku codebase. 
2) change the variables and types cases and make them consistent: 2) camalCase for variables 2) PascalCase for types 3) remove all the snake case uses in function names and variables if any 
3) Separate data from logic, move rln-relay static keys and solidity byte codes to separate files, and load them from their files

jm-clius:
@rymnc I'm not sure if this issue is still necessary/relevant. Some of these items may have been addressed or are already captured in other issues?

rymnc:
yes, we can close this issue now

PR #1158, opened 2022-09-15
feat(rln-relay): validate the merkle root in the RateLimitProof

### LINKED ISSUE #1049: Validating the Merkle tree root of the rln-relay proofs
# Problem
In the waku-rln-relay protocol, the tree root in the rln proof is not checked against the current state of the membership group. This entails two issues:
- Peers can make proof against some other rln group
- Peers who are deleted from the group can still generate a valid membership proof pointing to an old version of the membership tree and pass the proof verification part

# Acceptance Criteria
- [ ] The Merkle root embedded in the rln-relay proofs should be compared against the routing peer's local Merkle tree root, and discarded in case of discrepancy.  
- [ ] In a more advanced solution, routing peer shall maintain a window of acceptable tree roots, for some recent valid states of the group, and compares each incoming message against that buffer of valid roots.

rymnc:
Should resolve #1049.

Note: There is no functionality for maintaining a window of acceptable tree roots. Should we track this in a separate issue? 

Edit: As @jm-clius pointed out in https://github.com/status-im/nwaku/pull/1158#discussion_r972237018, errors should be better handled when dealing with the `proofVerify` function. Instead of returning `false` when one of the rln libs throw errors, we throw errors. This PR also includes this change.

### DIFF
--- a/tests/v2/test_waku_rln_relay.nim
+++ b/tests/v2/test_waku_rln_relay.nim
@@ -633,14 +633,19 @@ suite "Waku rln relay":
     # verify the proof (should not be verified)
     let verified = rln.proofVerify(data = messageBytes,
                                   proof = proof)
-    check:
+
+    require:
       verified.isOk()
+    check:
       verified.value() == false
 
   test "invalidate messages with a valid, but stale root":
+    # Setup: 
+    # This step consists of creating the rln instance,
+    # Inserting members, and creating a valid proof with the merkle root
     var rlnInstance = createRLNInstance()
-    check:
-      rlnInstance.isOk == true
+    require:
+      rlnInstance.isOk() == true
     var rln = rlnInstance.value
 
     let
@@ -663,6 +668,8 @@ suite "Waku rln relay":
       check:
         memberIsAdded
 
+    # Given: 
+    # This step includes constructing a valid message with the latest merkle root
     # prepare the message
     let messageBytes = "Hello".toBytes()
 
@@ -675,7 +682,7 @@ suite "Waku rln relay":
                                     memKeys = memKeys,
                                     memIndex = MembershipIndex(index),
                                     epoch = epoch)
-    check:
+    require:
       validProofRes.isOk()
     let validProof = validProofRes.value
 
@@ -683,26 +690,36 @@ suite "Waku rln relay":
     let verified = rln.proofVerify(data = messageBytes,
                                   proof = validProof)
 
-    check:
+    require:
       verified.isOk()
       verified.value() == true
 
+    # When: 
+    # This test depends on the local merkle tree root being different than a
+    # new message with an older/different root
+    # This can be simulated by removing a member, which changes the root of the tree
+    # Which is equivalent to a member being removed upon listening to the events emitted by the contract
     # Progress the local tree by removing a member
     discard rln.removeMember(MembershipIndex(0))
 
     # Ensure the local tree root has changed
     let currentMerkleRoot = rln.getMerkleRoot()
 
-    check:
+    require:
       currentMerkleRoot.isOk()
       currentMerkleRoot.value() != validProof.merkleRoot
 
+    # Then: 
+    # we try to verify a proof against this new merkle tree,
+    # which should return false
     # Try to send a message constructed with an older root
     let olderRootVerified = rln.proofVerify(data = messageBytes,
                                             proof = validProof)
 
-    check:
+    require:
       olderRootVerified.isOk()
+
+    check:
       olderRootVerified.value() == false
 
   test "toEpoch and fromEpoch consistency check":
--- a/tests/v2/test_waku_rln_relay.nim
+++ b/tests/v2/test_waku_rln_relay.nim
@@ -686,9 +686,8 @@ suite "Waku rln relay":
       validProofRes.isOk()
     let validProof = validProofRes.value
 
-    # verify the proof (should be verified)
-    let verified = rln.proofVerify(data = messageBytes,
-                                  proof = validProof)
+    # validate the root (should be true)
+    let verified = rln.validateRoot(validProof.merkleRoot)
 
     require:
       verified.isOk()
@@ -713,8 +712,7 @@ suite "Waku rln relay":
     # we try to verify a proof against this new merkle tree,
     # which should return false
     # Try to send a message constructed with an older root
-    let olderRootVerified = rln.proofVerify(data = messageBytes,
-                                            proof = validProof)
+    let olderRootVerified = rln.validateRoot(validProof.merkleRoot)
 
     require:
       olderRootVerified.isOk()
--- a/tests/v2/test_waku_rln_relay.nim
+++ b/tests/v2/test_waku_rln_relay.nim
@@ -631,6 +631,71 @@ suite "Waku rln relay":
                                  proof = proof)
     check:
       verified == false
+
+  test "invalidate messages with a valid, but stale root":
+    var rlnInstance = createRLNInstance()
+    check:
+      rlnInstance.isOk == true
+    var rln = rlnInstance.value
+
+    let
+      # create a membership key pair
+      memKeys = membershipKeyGen(rln).get()
+      # peer's index in the Merkle Tree
+      index = 5
+
+    # Create a Merkle tree with random members
+    for i in 0..10:
+      var member_is_added: bool = false
+      if (i == index):
+        # insert the current peer's pk
+        member_is_added = rln.insertMember(memKeys.idCommitment)
+      else:
+        # create a new key pair
+        let memberKeys = rln.membershipKeyGen()
+        member_is_added = rln.insertMember(memberKeys.get().idCommitment)
+      # check the member is added
+      check:
+        member_is_added
+
+    # prepare the message
+    let messageBytes = "Hello".toBytes()
+
+    # prepare the epoch
+    var epoch: Epoch
+    debug "epoch in bytes", epochHex = epoch.toHex()
+
+    # generate proof
+    let validProofRes = rln.proofGen(data = messageBytes,
+                                memKeys = memKeys,
+                                memIndex = MembershipIndex(index),
+                                epoch = epoch)
+    check:
+      validProofRes.isOk()
+    let validProof = validProofRes.value
+
+    # verify the proof (should be verified)
+    let verified = rln.proofVerify(data = messageBytes,
+                                   proof = validProof)
+    check:
+      verified == true
+
+    # Progress the local tree by removing a member
+    discard rln.removeMember(MembershipIndex(0))
+
+    # Ensure the local tree root has changed
+    check:
+      rln.getMerkleRoot().value() != validProof.merkleRoot
+
+    warn "current, invalid", current=rln.getMerkleRoot().value().toHex, invalid=validProof.merkleRoot.toHex
+
+    # Try to send a message constructed with an older root
+    let olderRootVerified = rln.proofVerify(data = messageBytes,
+                                            proof = validProof)
+
+    check:
+      olderRootVerified == false
+
   test "toEpoch and fromEpoch consistency check":
     # check edge cases
     let
--- a/tests/v2/test_waku_rln_relay.nim
+++ b/tests/v2/test_waku_rln_relay.nim
@@ -579,8 +579,14 @@ suite "Waku rln relay":
     # verify the proof
     let verified = rln.proofVerify(data = messageBytes,
                                     proof = proof)
+
+    # Ensure the proof verification did not error out
+
     check:
-      verified == true
+      verified.isOk()
+
+    check:
+      verified.value() == true
 
   test "test proofVerify and proofGen for an invalid proof":
     var rlnInstance = createRLNInstance()
@@ -628,9 +634,11 @@ suite "Waku rln relay":
 
     # verify the proof (should not be verified)
     let verified = rln.proofVerify(data = messageBytes,
-                                 proof = proof)
+                                  proof = proof)
     check:
-      verified == false
+      verified.isOk()
+    check:
+      verified.value() == false
 
   test "invalidate messages with a valid, but stale root":
     var rlnInstance = createRLNInstance()
@@ -646,17 +654,17 @@ suite "Waku rln relay":
 
     # Create a Merkle tree with random members
     for i in 0..10:
-      var member_is_added: bool = false
+      var memberIsAdded: bool = false
       if (i == index):
         # insert the current peer's pk
-        member_is_added = rln.insertMember(memKeys.idCommitment)
+        memberIsAdded = rln.insertMember(memKeys.idCommitment)
       else:
         # create a new key pair
         let memberKeys = rln.membershipKeyGen()
-        member_is_added = rln.insertMember(memberKeys.get().idCommitment)
+        memberIsAdded = rln.insertMember(memberKeys.get().idCommitment)
       # check the member is added
       check:
-        member_is_added
+        memberIsAdded
 
     # prepare the message
     let messageBytes = "Hello".toBytes()
@@ -667,34 +675,42 @@ suite "Waku rln relay":
 
     # generate proof
     let validProofRes = rln.proofGen(data = messageBytes,
-                                memKeys = memKeys,
-                                memIndex = MembershipIndex(index),
-                                epoch = epoch)
+                                    memKeys = memKeys,
+                                    memIndex = MembershipIndex(index),
+                                    epoch = epoch)
     check:
       validProofRes.isOk()
     let validProof = validProofRes.value
 
     # verify the proof (should be verified)
     let verified = rln.proofVerify(data = messageBytes,
-                                   proof = validProof)
+                                  proof = validProof)
+
     check:
-      verified == true
+      verified.isOk()
+    check:
+      verified.value() == true
 
     # Progress the local tree by removing a member
     discard rln.removeMember(MembershipIndex(0))
 
     # Ensure the local tree root has changed
-    check:
-      rln.getMerkleRoot().value() != validProof.merkleRoot
+    let currentMerkleRoot = rln.getMerkleRoot()
 
-    warn "current, invalid", current=rln.getMerkleRoot().value().toHex, invalid=validProof.merkleRoot.toHex
+    check:
+      currentMerkleRoot.isOk()
+
+    check:
+      currentMerkleRoot.value() != validProof.merkleRoot
 
     # Try to send a message constructed with an older root
     let olderRootVerified = rln.proofVerify(data = messageBytes,
                                             proof = validProof)
 
     check:
-      olderRootVerified == false
+      olderRootVerified.isOk()
+    check:
+      olderRootVerified.value() == false
 
   test "toEpoch and fromEpoch consistency check":
     # check edge cases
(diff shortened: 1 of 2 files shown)


tests/v2/test_waku_rln_relay.nim
```diff
@@ -631,6 +631,71 @@ suite "Waku rln relay":
...
+
+    let
+      # create a membership key pair
+      memKeys = membershipKeyGen(rln).get()
+      # peer's index in the Merkle Tree
+      index = 5
+
+    # Create a Merkle tree with random members
+    for i in 0..10:
+      var member_is_added: bool = false
```
jm-clius, line 649: Is there a reason (here and some places elsewhere) why `camelCase` is not used for variable naming?

rymnc: There's an issue which tracks some of the refactoring required - https://github.com/status-im/nwaku/issues/946.

I'll fix the code which was touched in the PR though.

waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
```diff
@@ -386,8 +386,28 @@ when defined(rln) or (not defined(rln) and not defined(rlnzerokit)):
...
     return proofBytes
 
-  proc proofVerify*(rlnInstance: RLN[Bn256], data: openArray[byte],
-    proof: RateLimitProof): bool =
+  proc getMerkleRoot*(rlnInstance: RLN[Bn256]): MerkleNodeResult =
+    # read the Merkle Tree root after insertion
+    var
+      root {.noinit.}: Buffer = Buffer()
+      rootPtr = addr(root)
+      get_root_successful = get_root(rlnInstance, rootPtr)
```
jm-clius, line 394: Why not `camelCase`?

waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
```diff
@@ -386,8 +386,28 @@ when defined(rln) or (not defined(rln) and not defined(rlnzerokit)):
...
+    if (not get_root_successful): return err("could not get the root")
+    if (not (root.len == 32)): return err("wrong output size")
+
+    var rootValue = cast[ptr MerkleNode] (root.`ptr`)[]
+    return ok(rootValue)
+
+  proc validateVerifiedProof*(rlnInstance: RLN[Bn256], proof: RateLimitProof): bool =
+    # Validate against the local merkle tree
+    let localTreeRoot = rlnInstance.getMerkleRoot().value()
```
jm-clius, line 403: If you have an expectation that `getMerkleRoot()` could return an error result (and based on the above it seems like it could happen), you shouldn't access the `MerkleNodeResult` value directly (which would raise a fatal `Defect`). I assume an error Result would be another indication that `validateVerifiedProof` fails?

tests/v2/test_waku_rln_relay.nim
jm-clius: Oh right! Thanks. Didn't see that issue.

waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
rymnc: Addressed in [a0b3761](https://github.com/status-im/nwaku/pull/1158/commits/a0b376199382ddf0174a2319187d5d5399ac4f1d)

waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
rymnc: Addressed in [a0b3761](https://github.com/status-im/nwaku/pull/1158/commits/a0b376199382ddf0174a2319187d5d5399ac4f1d)

tests/v2/test_waku_rln_relay.nim
rymnc: Addressed in [a0b3761](https://github.com/status-im/nwaku/pull/1158/commits/a0b376199382ddf0174a2319187d5d5399ac4f1d)

tests/v2/test_waku_rln_relay.nim
```diff
@@ -579,8 +579,14 @@ suite "Waku rln relay":
...
                                     proof = proof)
+
+    # Ensure the proof verification did not error out
+
+    check:
+      verified.isOk()
+
     check:
-      verified == true
+      verified.value() == true
```
staheri14, line 589: You can group all the checks together:
```
check:
      verified.isOk()
      verified.value() == true
```

tests/v2/test_waku_rln_relay.nim
```diff
@@ -628,9 +634,84 @@ suite "Waku rln relay":
 
     # verify the proof (should not be verified)
     let verified = rln.proofVerify(data = messageBytes,
-                                 proof = proof)
+                                  proof = proof)
+    check:
+      verified.isOk()
+    check:
+      verified.value() == false
```
staheri14, line 641: Ditto here.

waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
```diff
@@ -397,11 +420,13 @@ when defined(rln) or (not defined(rln) and not defined(rlnzerokit)):
...
+      return err("could not verify proof")
     # f = 0 means the proof is verified
-    if f == 0:
-      return true
-    return false
+    if f != 0:
+      return ok(false)
+
+    # Compare merkle tree roots for a valid proof
+    return validateVerifiedProof(rlnInstance, proof)
```
staheri14, line 429: We can save computation time if we check the root before proof verification. I'd move `validateVerifiedProof` before the call to the `verify` proc in line `420`. Also, the proc name `validateVerifiedProof` is a bit general while it only validates the root.

staheri14 (commented):
Looks good, and nice job with the quick PR! 
Please find my comments below.

waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
rymnc: Addressed in [fefb7e7](https://github.com/status-im/nwaku/pull/1158/commits/fefb7e706182ee964835d1393c5a9f7dc480ad9d)

tests/v2/test_waku_rln_relay.nim
rymnc: Addressed in [fefb7e7](https://github.com/status-im/nwaku/pull/1158/commits/fefb7e706182ee964835d1393c5a9f7dc480ad9d)

tests/v2/test_waku_rln_relay.nim
rymnc: Addressed in [fefb7e7](https://github.com/status-im/nwaku/pull/1158/commits/fefb7e706182ee964835d1393c5a9f7dc480ad9d)

waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
```diff
@@ -386,22 +387,53 @@ when defined(rln) or (not defined(rln) and not defined(rlnzerokit)):
...
     var
       proofBytes = serialize(proof, data)
       proofBuffer = proofBytes.toBuffer()
       f = 0.uint32
     trace "serialized proof", proof = proofBytes.toHex()
 
+    let rootIsValid = validateRoot(rlnInstance, proof)
+
+    if rootIsValid.isErr():
+      return err(rootIsValid.error())
```
rymnc, line 423: Please note that this could have been

```nim
let rootIsValid = ?validateRoot(rlnInstance, proof)
```

However, to enable future refactoring, for tighter error enums, it has been handled manually

waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
```diff
@@ -32,6 +32,7 @@ when defined(rlnzerokit):
 
 
 type MerkleNodeResult* = Result[MerkleNode, string]
+type BoolResult* = Result[bool, string]
 type RateLimitProofResult* = Result[RateLimitProof, string]
```
LNSD, line 36: A little nit-picking, but I think that having all these result types can be improved. And more if they are not going to evolve separately.

I suggest doing the following:
* Have a generic base result type alias for the RLN relay:
```nim
type RlnRelayResult*[T] = Result[T, string]
```
And then define special result types aliases, only in the case they are used in more than one place:

```nim
type MerkleNodeResult* = RlnRelayResult[MerkleNode]
type RateLimitProofResult* = RlnRelayResult[RateLimitProof]
```
> Disclaimer: I have not checked the number of usages of the following type aliases. If they have been used only in one place, then it is preferable to go with the generic type directly.

I do not like the `BoolResult` 🤔; with my proposal, the result of the method would become:
```nim
proc foo*(): Future[RlnRelayResult[bool]] =
  # [...]
```

waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
```diff
@@ -386,22 +387,53 @@ when defined(rln) or (not defined(rln) and not defined(rlnzerokit)):
...
-  proc proofVerify*(rlnInstance: RLN[Bn256], data: openArray[byte],
-    proof: RateLimitProof): bool =
+  proc getMerkleRoot*(rlnInstance: RLN[Bn256]): MerkleNodeResult =
+    # read the Merkle Tree root after insertion
+    var
+      root {.noinit.}: Buffer = Buffer()
+      rootPtr = addr(root)
+      getRootSuccessful = get_root(rlnInstance, rootPtr)
+    if (not getRootSuccessful): return err("could not get the root")
+    if (not (root.len == 32)): return err("wrong output size")
```
LNSD, line 397: For easier readability and consistency with nwaku's codebase style, please make these if statements multi-line.

```suggestion
    if not getRootSuccessful: 
      return err("could not get the root")
    if root.len != 32: 
      return err("wrong output size")
```

tests/v2/test_waku_rln_relay.nim
```diff
@@ -628,9 +632,79 @@ suite "Waku rln relay":
...
+                                  proof = proof)
+    check:
+      verified.isOk()
+      verified.value() == false
+
+  test "invalidate messages with a valid, but stale root":
+    var rlnInstance = createRLNInstance()
     check:
-      verified == false
+      rlnInstance.isOk == true
```
LNSD, line 643: If this is a condition that if it fails, it will fail in many tests, then it is preferable to use the `require()` assertion method.

tests/v2/test_waku_rln_relay.nim
```diff
@@ -628,9 +632,79 @@ suite "Waku rln relay":
 
     # verify the proof (should not be verified)
     let verified = rln.proofVerify(data = messageBytes,
-                                 proof = proof)
+                                  proof = proof)
+    check:
+      verified.isOk()
+      verified.value() == false
+
+  test "invalidate messages with a valid, but stale root":
```
LNSD, line 640: When designing a test, you should try to follow this structure:

* **Setup:** Setting up the test harness (e.g., async runtime, waku node). This step is optional.
* **Given:** You declare all the pieces that will interact in the test. These are the stimulus (e.g., test content topic string, prepopulated Merkle tree) and the SUT (System Under Test).
* **When:** Then you stimulate the SUT. Execute the action that you want to test and keep the system outputs.
* **Then:** Here, you do the assertions of the system outputs.
* **Cleanup/Teardown:** The last step is to clean up everything that was set up in the first stage; or restore the state for something that got mutated. This step is also optional.

This will make a huge difference in the quality of the code, as you'll start thinking about doing more testable code, and also, it will make the tests way easier to understand, review and maintain.

LNSD: As a rule of thumb, `check()` can only go in the **Then** section. And if there is a condition _sine qua non_ the test should halt, then you should use `require` that finishes the test execution. This second method can be used in any test stage. It is not harmful to use `require` in the _Then_ section, but in that section, it is preferable to just use `check()`.

LNSD (commented):
Please, check my comments 😁

waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
rymnc: Addressed in [87661ba](https://github.com/status-im/nwaku/pull/1158/commits/87661ba1d572a90bbba441fca568e2dad9fb5fa9)

We will probably update this in one of the refactoring PRs to include a second generic type, the Error enum

waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
rymnc: Addressed in [ac3975b](https://github.com/status-im/nwaku/pull/1158/commits/ac3975b05d4ba26addd6b7544c4828147253d4cf)

tests/v2/test_waku_rln_relay.nim
rymnc: Addressed in [f75eef3](https://github.com/status-im/nwaku/pull/1158/commits/f75eef32439c50edb2b8395da8b3f3d425abe110)

LNSD (approved):
LGTM ✅

Awesome! Good job 🤝

tests/v2/test_waku_rln_relay.nim
rymnc: Addressed in [5d48c46](https://github.com/status-im/nwaku/pull/1158/commits/5d48c46018f964874d9404006398d3505ebf3728)

waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
```diff
@@ -386,22 +386,55 @@ when defined(rln) or (not defined(rln) and not defined(rlnzerokit)):
...
+
+  proc validateRoot*(rlnInstance: RLN[Bn256], proof: RateLimitProof): RlnRelayResult[bool] =
+    # Validate against the local merkle tree
+    let localTreeRoot = rlnInstance.getMerkleRoot()
+    if not localTreeRoot.isOk():
+      return err(localTreeRoot.error())
+    if localTreeRoot.value() == proof.merkleRoot:
+      return ok(true)
+    else:
+      trace "Local tree root does not match the root sent by peer", localTreeRoot=localTreeRoot, messageRoot=proof.merkleRoot
```
staheri14, line 411:
```suggestion
      trace "Local tree root does not match the supplied proof's root", localTreeRoot=localTreeRoot, messageRoot=proof.merkleRoot
```

waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
```diff
@@ -386,22 +386,55 @@ when defined(rln) or (not defined(rln) and not defined(rlnzerokit)):
...
+      return ok(false)
+
+  proc proofVerify*(rlnInstance: RLN[Bn256], data: openArray[byte], proof: RateLimitProof): RlnRelayResult[bool] =
```
staheri14, line 414: I remember during our call we agreed that the root validation shall  be integrated into the [`validateMessage` proc](https://github.com/status-im/nwaku/blob/master/waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim#L725) (did not notice in the first review that it is added to the `proofVerify`). The reason is that the `proofVerify` proc acts as a wrapper for the [`verify` proc](https://github.com/status-im/nwaku/blob/f73477be6e4ee3c8f2b7f96bc7f7746a5071ec7a/waku/v2/protocol/waku_rln_relay/rln.nim#L68) and its sole intention is to verify the validity of the [zkSNARK proof](https://github.com/status-im/nwaku/blob/master/waku/v2/protocol/waku_rln_relay/waku_rln_relay_types.nim#L56)  of the `RateLimitProof` object regardless of the state of the Merkle tree. Validation of other fields of `RateLimitProof` object shall be added to the `validateMessage` proc.

staheri14 (commented):
Thanks for addressing the comments. One major part was overlooked which I explained in my comment :)

waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
rymnc: Decided to remove this line altogether, in [d367f91](https://github.com/status-im/nwaku/pull/1158/commits/d367f9197738026ca13071f797de91a9ac30b3ad)

waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
rymnc: Addressed in [d367f91](https://github.com/status-im/nwaku/pull/1158/commits/d367f9197738026ca13071f797de91a9ac30b3ad)

LNSD (approved):
LGTM ✅

waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
```diff
@@ -811,7 +792,18 @@ proc validateMessage*(rlnPeer: WakuRLNRelay, msg: WakuMessage,
   let
     contentTopicBytes = msg.contentTopic.toBytes
     input = concat(msg.payload, contentTopicBytes)
-    proofVerificationRes = rlnPeer.rlnInstance.proofVerify(input, msg.proof)
+
```
staheri14, line 795: These lines were grouped together with line `806`  i.e.,
```
let proofVerificationRes = rlnPeer.rlnInstance.proofVerify(input, msg.proof)
```

waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
```diff
@@ -811,7 +792,18 @@ proc validateMessage*(rlnPeer: WakuRLNRelay, msg: WakuMessage,
   let
     contentTopicBytes = msg.contentTopic.toBytes
     input = concat(msg.payload, contentTopicBytes)
-    proofVerificationRes = rlnPeer.rlnInstance.proofVerify(input, msg.proof)
+  
+  let merkleRootIsValidRes = rlnPeer.rlnInstance.validateRoot(msg.proof.merkleRoot)
+
+  if merkleRootIsValidRes.isErr():
+      debug "invalid message: could not compute root"
```
staheri14, line 799:
```
debug "invalid message: could not validate the root"
```

staheri14 (approved):
Looks good, just some nitpicks, then it is ready to merge.

jm-clius (approved):
LGTM! Good work.

waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
rymnc: Addressed in [9f76dab](https://github.com/status-im/nwaku/pull/1158/commits/9f76dab6e712b951feb10f10af19f1700d9c6712)

waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
rymnc: Addressed in [9f76dab](https://github.com/status-im/nwaku/pull/1158/commits/9f76dab6e712b951feb10f10af19f1700d9c6712)

PR #1161, opened 2022-09-15
fix: nodes in AddressBook are not necessarily connected

jm-clius:
One-liner fix for issue that prevented the node to reconnect to peers from the discv5 loop. Peers that appear in the `AddressBook` are not necessarily connected anymore (e.g. they may have restarted) and a reconnect should be attempted when they are discovered "again" via Discovery v5.

Can confirm improvement on my own deployed node - previously if the fleet restarted the running node would never attempt to reconnect.


LNSD (approved):
LGTM ✅

jm-clius:
Wait, what happened to signing on this PR? :) Fixing that and will merge after.

PR #1162, opened 2022-09-15
fix: no need to start protocols twice

jm-clius:
This was a bit of a silly oversight when the submodules were updated: 
there's no need to start protocols explicitly before the `switch` is started, as all mounted protocols will be started [here](https://github.com/status-im/nim-libp2p/blob/master/libp2p/multistream.nim#L217).

It's only necessary to explicitly start protocols when mounted after the switch has already started.


jm-clius:
Investigating the CI failures. Converting to draft until then.

LNSD (approved):
LGTM ✅

Issue #1155, opened 2022-09-14
Execute retention policies periodically instead of at insertion time


LNSD:
## Requirement
Reduce the message insertion duration by running the retention policy in a periodic task instead running it at insertion time.

## Background/context
The starting point is the following:

![image](https://user-images.githubusercontent.com/3949095/190628802-c4e6cb45-359c-4d81-bdd5-e6d284e0c61f.png)

LNSD:
After the refactoring was performed in #1163, #1165, and #1164, as expected, the message store insertion time stayed the same as before.

![image](https://user-images.githubusercontent.com/3949095/190629230-c57894d3-0f02-4b8e-a2f7-7801c9ba4b8c.png)

The next step is to stop running the retention policy at insertion time and move it to a periodic task.

LNSD:
I did a draft implementation and deployed it to `wakuv2.test` fleet with the following results:

* CPU usage consumption decrease:
![image](https://user-images.githubusercontent.com/3949095/190993380-d2414736-c2d3-40fa-aa94-36d075c02afc.png)

* Insertion time has decreased too:
![image](https://user-images.githubusercontent.com/3949095/190993434-4108047c-05f4-4db3-80dc-a2503aa6236a.png)

PR #1163, opened 2022-09-15
refactor(store): don't cache the message count in sqlite message store

LNSD:
The present PR forms part of a series of refactoring PRs that aim to decouple the retention policy from the message store interface completely.

* Make message store interface `getMessagesCount` method fallible
* Do not cache SQLite's store message count

This is part of the groundwork necessary for #1155


jm-clius (approved):
LGTM! Looking forward to the rest of the series and the cliffhanging finale. :)

PR #1165, opened 2022-09-15
refactor(store): extend message store interface with store manipulation methods

LNSD:
The present PR forms part of a series of refactoring PRs that aim to decouple the retention policy from the message store interface completely.

* Extend message store interface
* Implement the message store interface methods for `SqliteStore`
* Implement the message store interface methods for `StoreQueueRef`

**NOTE:** A couple of in-memory store methods have not been implemented. Some extra work is needed there to support message retention policies.

This is part of the groundwork necessary for #1155

### DIFF
--- a/waku/v2/node/storage/message/message_store.nim
+++ b/waku/v2/node/storage/message/message_store.nim
@@ -46,4 +46,16 @@ method getMessagesByHistoryQuery*(
   ascendingOrder = true
 ): MessageStoreResult[MessageStorePage] {.base.} = discard
 
+
+# Store manipulation
+
 method getMessagesCount*(ms: MessageStore): MessageStoreResult[int64] {.base.} = discard
+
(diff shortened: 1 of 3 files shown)


rymnc (approved):

jm-clius:
> NOTE: A couple of in-memory store methods have not been implemented. Some extra work is needed there to support message retention policies.

I think having these be not-implemented is the most reasonable way to go for now. It's debatable whether a standalone StoreQueue should (a) be considered a MessageStore at all and (b) should therefore have retention policy applied to it directly. In fact, the most efficient implementation for the StoreQueue would be a circular buffer of fixed capacity, which would make retention policy redundant. That said, if we abstract the StoreQueue together with its on-disck SQLite backup into a single MessageStore, retention policy would make sense.

jm-clius (approved):
Makes sense to me. See comment on the future of retention policy as applied to the StoreQueue

PR #1172, opened 2022-09-21
refactor(wakunode): remove deprecated non-async methods

LNSD:
The present PR aims to remove the deprecated non-async methods from `wakunode2` module:

* Remove waku lightpush deprecated method
* Remove waku waku store query deprecated method
* Update tests and examples accordingly

### DIFF
--- a/tests/v2/test_waku_swap.nim
+++ b/tests/v2/test_waku_swap.nim
@@ -1,9 +1,12 @@
 {.used.}
 
 import
-  std/[options, tables, sets],
+  std/[options, tables, sets, times],
+  stew/byteutils,
+  stew/shims/net as stewNet, 
   testutils/unittests,
-  chronos, chronicles, stew/shims/net as stewNet, stew/byteutils,
+  chronos, 
+  chronicles, 
   libp2p/switch,
   libp2p/protobuf/minprotobuf,
   libp2p/stream/[bufferstream, connection],
@@ -17,7 +20,31 @@ import
   ../../waku/v2/node/storage/message/waku_store_queue,
   ../../waku/v2/node/wakunode2,
   ../../waku/v2/utils/peers,
-  ../test_helpers, ./utils
+  ../../waku/v2/utils/time,
+  ../test_helpers, 
+  ./utils
+
+
+const 
+  DefaultPubsubTopic = "/waku/2/default-waku/proto"
+  DefaultContentTopic = ContentTopic("/waku/2/default-content/proto")
+
+
+proc now(): Timestamp = 
+  getNanosecondTime(getTime().toUnixFloat())
+
+proc fakeWakuMessage(
+  payload = "TEST-PAYLOAD",
+  contentTopic = DefaultContentTopic, 
+  ts = now()
+): WakuMessage = 
+  WakuMessage(
+    payload: toBytes(payload),
+    contentTopic: contentTopic,
+    version: 1,
+    timestamp: ts
+  )
+
 
 procSuite "Waku SWAP Accounting":
   test "Handshake Encode/Decode":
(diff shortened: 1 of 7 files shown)


tests/v2/test_waku_swap.nim
```diff
@@ -50,98 +77,91 @@ procSuite "Waku SWAP Accounting":
...
+    ## When
+    let queryRes = await client.query(HistoryQuery(contentFilters: @[HistoryContentFilter(contentTopic: DefaultContentTopic)]))
 
-    proc storeHandler(response: HistoryResponse) {.gcsafe, closure.} =
-      debug "storeHandler hit"
-      check:
-        response.messages[0] == message
-      completionFut.complete(true)
+    ## Then
+    check queryRes.isOk()
```
rymnc, line 107:
```suggestion
    require queryRes.isOk()
```

tests/v2/test_waku_swap.nim
```diff
@@ -50,98 +77,91 @@ procSuite "Waku SWAP Accounting":
...
 
-    await node1.query(HistoryQuery(contentFilters: @[HistoryContentFilter(contentTopic: contentTopic)]), storeHandler)
+    let response = queryRes.get()
+    check:
```
rymnc, line 110:
```suggestion
    require:
```

tests/v2/test_waku_swap.nim
```diff
@@ -50,98 +77,91 @@ procSuite "Waku SWAP Accounting":
...
+    ## When
+    # TODO: Handshakes - for now we assume implicit, e2e still works for PoC
+    let res1 = await client.query(HistoryQuery(contentFilters: @[HistoryContentFilter(contentTopic: DefaultContentTopic)]))
+    let res2 = await client.query(HistoryQuery(contentFilters: @[HistoryContentFilter(contentTopic: DefaultContentTopic)]))
 
-    # TODO Handshakes - for now we assume implicit, e2e still works for PoC
-    await node1.query(HistoryQuery(contentFilters: @[HistoryContentFilter(contentTopic: contentTopic)]), handler1)
-    await node1.query(HistoryQuery(contentFilters: @[HistoryContentFilter(contentTopic: contentTopic)]), handler2)
+    ## Then
+    check:
```
rymnc, line 156:
```suggestion
    require:
```

tests/v2/test_wakunode_store.nim
```diff
@@ -29,172 +29,185 @@ import
...
+  getNanosecondTime(getTime().toUnixFloat())
+
 proc newTestMessageStore(): MessageStore =
   let database = SqliteDatabase.init("", inMemory = true)[]
   SqliteStore.init(database).tryGet()
 
+proc fakeWakuMessage(
+  payload = "TEST-PAYLOAD",
+  contentTopic = DefaultContentTopic, 
+  ts = now()
```
rymnc, line 47: Maybe add the ephemeral flag here?

tests/v2/test_wakunode_store.nim
```diff
@@ -29,172 +29,185 @@ import
...
-        response.messages[0] == message
-      completionFut.complete(true)
-
-    await node1.query(HistoryQuery(contentFilters: @[HistoryContentFilter(contentTopic: contentTopic)]), storeHandler)
+    ## When
+    let req = HistoryQuery(contentFilters: @[HistoryContentFilter(contentTopic: DefaultContentTopic)])
+    let queryRes = await client.query(req)
+    
+    ## Then
+    check queryRes.isOk()
```
rymnc, line 83:
```suggestion
    require queryRes.isOk()
```

waku/v2/node/wakunode2.nim
```diff
@@ -354,19 +357,8 @@ proc lightpush*(node: WakuNode, topic: Topic, message: WakuMessage): Future[Waku
...
-  ## functionality.
-
-  let rpc = PushRequest(pubSubTopic: topic, message: message)
-  let res = await node.wakuLightPush.request(rpc)
-  if res.isOk():
-    handler(res.value)
-  else:
-    error "Message lightpush failed", error=res.error()
+proc lightpush2*(node: WakuNode, topic: Topic, message: WakuMessage) {.async, gcsafe.} =
+  discard await node.lightpush(topic, message)
```
LNSD, line 361: I had to create this _fire-and-forget_ method because it is required by `chat2` example

rymnc (approved):
LGTM, minor nitpicks!

LNSD:
@rymnc Regarding your suggestions, I will refer to the [other PR conversation](https://github.com/status-im/nwaku/pull/1158#discussion_r972828110): 

> As a rule of thumb, `check()` can only go in the **Then** section. And if there is a condition sine qua non the test should halt, then you should use `require()` that finishes the test execution. This second method can be used in any test stage. It is not harmful to use `require()` in the **Then** section, but in that section, it is preferable just to use `check()`.

And I want to add a clarification:

> Inside **Then** section we should only use the `check()` statement.


We should write down these "rules" to have them as the project code style guide (that extends the status nim style guide and only applies to this specific project)

tests/v2/test_wakunode_store.nim
LNSD: We need to extract all these test helper procedures and constants in a common module under the `tests` folder to reuse them. This is a big TODO 😄 

- [ ] Extract all test helper procedures and constants in a common module

rymnc: Agreed! should we track it in an issue?

LNSD: I have created an issue: https://github.com/status-im/nwaku/issues/1173

waku/v2/node/wakunode2.nim
jm-clius: Wouldn't it be easier to just dispatch the future in `chat2` and discard it? `chat2` is a bit of a mess, so I don't think we should design around its idiosyncrasies.

jm-clius (approved):
LGTM!

waku/v2/node/wakunode2.nim
LNSD: Indeed, `chat2` is a nightmare regarding this kind of PRs.

Here the problem is that if I remove the method from here, I need to wrap it in an anonymous proc and discard the result in 2 places in the `chat2` code:

* https://github.com/status-im/nwaku/blob/refactor-wakunode-remove-deprecated-methods/examples/v2/chat2.nim#L244
* https://github.com/status-im/nwaku/blob/refactor-wakunode-remove-deprecated-methods/examples/v2/chat2.nim#L273

Because we are spawning directly the `Future` returned by the waku node `lightpush()` proc 😞

PR #1175, opened 2022-09-21
chore(store): move constants to retention policies modules

LNSD:
Small PR moving the store message retention constants from the message store module to the right module


rymnc (approved):
LGTM!

jm-clius (approved):
👍

PR #1178, opened 2022-09-22
fix(store): move insert criteria logic to waku store protocol module

LNSD:
This PR aims to address the inconsistencies based [the discussion](https://github.com/vacp2p/rfc/pull/535#discussion_r973520241) about what should be considered a valid message based on the metadata and the payload.

The gist of the changes:

> As we now have more message store implementations, in this PR, I have changed the `MessageStore` interface dependence on the `Index` type when a message is inserted in the DB (see `message_store`'s interface `put()` API). And moved all the common insertion logic to the waku protocol module.
> 
> Now the messages digest is calculated before inserting it into the store. Calculating a hash is a computational "costly" operation, so the lesser number of times that we compute it, the better.

### DIFF
--- a/tests/v2/test_message_store_queue.nim
+++ b/tests/v2/test_message_store_queue.nim
@@ -88,34 +88,6 @@ procSuite "Sorted store queue":
     check:
       store.len == capacity
 
-  test "Sender time can't be more than MaxTimeVariance in future":
-    ## Given
-    let capacity = 5
-    let store = StoreQueueRef.new(capacity)
-    let
-      receiverTime = getNanoSecondTime(10)
-      senderTimeOk = receiverTime + StoreMaxTimeVariance
-      senderTimeErr = senderTimeOk + 1
-    
-    let invalidMessage = IndexedWakuMessage(
-      msg: WakuMessage(
-        payload: @[byte 1], 
-        timestamp: senderTimeErr
-      ),
-      index: Index(
-        receiverTime: receiverTime, 
-        senderTime: senderTimeErr
-      )
-    )
-    
-    ## When
-    let addRes = store.add(invalidMessage)
-    
-    ## Then
-    check:
-      addRes.isErr()
-      addRes.error() == "future_sender_timestamp"
-    
   test "Store queue sort-on-insert works":    
     ## Given
     let
--- a/tests/v2/test_message_store_sqlite.nim
+++ b/tests/v2/test_message_store_sqlite.nim
@@ -25,13 +25,16 @@ const
 proc newTestDatabase(): SqliteDatabase =
   SqliteDatabase.init("", inMemory = true).tryGet()
 
+proc now(): Timestamp =
+  getNanosecondTime(getTime().toUnixFloat())
+
 proc getTestTimestamp(offset=0): Timestamp = 
-  Timestamp(getNanosecondTime(epochTime()))
+  Timestamp(getNanosecondTime(getTime().toUnixFloat() + offset.float))
 
 proc fakeWakuMessage(
   payload = "TEST-PAYLOAD",
   contentTopic = DefaultContentTopic, 
-  ts = getNanosecondTime(epochTime())
+  ts = now()
 ): WakuMessage = 
   WakuMessage(
     payload: toBytes(payload),
@@ -80,10 +83,9 @@ suite "SQLite message store - insert messages":
       store = SqliteStore.init(database).tryGet()
 
     let message = fakeWakuMessage(contentTopic=contentTopic)
-    let messageIndex = Index.compute(message, getNanosecondTime(epochTime()), DefaultPubsubTopic)
 
     ## When
-    let resPut = store.put(messageIndex, message, DefaultPubsubTopic)
+    let resPut = store.put(DefaultPubsubTopic, message)
 
     ## Then
     check:
@@ -123,8 +125,7 @@ suite "SQLite message store - insert messages":
 
     ## When
     for msg in messages:
-      let index = Index.compute(msg, msg.timestamp, DefaultPubsubTopic)
-      require store.put(index, msg, DefaultPubsubTopic).isOk()
+      require store.put(DefaultPubsubTopic, msg).isOk()
       require retentionPolicy.execute(store).isOk()
 
     ## Then
@@ -147,25 +148,24 @@ suite "Message Store":
     let 
       database = newTestDatabase()
       store = SqliteStore.init(database).get()
-      topic = DefaultContentTopic
-      pubsubTopic = DefaultPubsubTopic
 
+    let
       t1 = getTestTimestamp(0)
       t2 = getTestTimestamp(1)
       t3 = high(int64)
 
     var msgs = @[
-      WakuMessage(payload: @[byte 1, 2, 3], contentTopic: topic, version: uint32(0), timestamp: t1),
-      WakuMessage(payload: @[byte 1, 2, 3, 4], contentTopic: topic, version: uint32(1), timestamp: t2),
+      WakuMessage(payload: @[byte 1, 2, 3], contentTopic: DefaultContentTopic, version: uint32(0), timestamp: t1),
+      WakuMessage(payload: @[byte 1, 2, 3, 4], contentTopic: DefaultContentTopic, version: uint32(1), timestamp: t2),
       # high(uint32) is the largest value that fits in uint32, this is to make sure there is no overflow in the storage
-      WakuMessage(payload: @[byte 1, 2, 3, 4, 5], contentTopic: topic, version: high(uint32), timestamp: t3),
+      WakuMessage(payload: @[byte 1, 2, 3, 4, 5], contentTopic: DefaultContentTopic, version: high(uint32), timestamp: t3),
     ]
 
     var indexes: seq[Index] = @[]
     for msg in msgs:
-      var index = Index.compute(msg, msg.timestamp, DefaultPubsubTopic)
-      let resPut = store.put(index, msg, pubsubTopic)
-      require resPut.isOk
+      require store.put(DefaultPubsubTopic, msg, computeDigest(msg), msg.timestamp).isOk()
+
+      let index = Index.compute(msg, msg.timestamp, DefaultPubsubTopic)
       indexes.add(index)
 
     ## When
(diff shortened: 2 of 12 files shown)


waku/v2/protocol/waku_store/protocol.nim
```diff
@@ -213,19 +225,26 @@ proc handleMessage*(w: WakuStore, pubsubTopic: string, msg: WakuMessage) =
...
-  trace "handling message", topic=pubsubTopic, index=index
+    trace "handling message", pubsubTopic=pubsubTopic, contentTopic=msg.contentTopic, timestamp=msg.timestamp, digest=msgDigest
 
-  # Add messages to persistent store, if present
-  let putStoreRes = w.store.put(index, msg, pubsubTopic)
-  if putStoreRes.isErr():
-    debug "failed to insert message to persistent store", index=index, err=putStoreRes.error
-    waku_store_errors.inc(labelValues = [insertFailure])
-    return
+    let putStoreRes = w.store.put(pubsubTopic, msg, msgDigest, msgReceivedTime)
```
LNSD, line 243: This is the logic that sets the message store timestamp:

* If the message timestamp (a.k.a. `senderTimestamp`) is present, use that timestamp as the store timestamp.
* If the message timestamp is equal to 0 (which means either the timestamp is not present or was set to 0), we insert the message with a store timestamp equal to the arrival time.

Note that at this point, we have already checked if the message is valid.

LNSD (commented):
Based on the discussion of https://github.com/vacp2p/rfc/pull/535#discussion_r973520241 this is the message validation criteria and message insertion logic:

> In case the timestamp is not present, let's fall back to the receiver time stamp (i.e., storedAt). The sender timestamp will not be altered, and the receiver timestamp is only used for sake of message ordering and is an internal logic.

waku/v2/node/storage/message/sqlite_store/queries.nim
```diff
@@ -247,11 +247,11 @@ proc timeRangeWhereClause(startTime: Option[Timestamp], endTime: Option[Timestam
...
   if startTime.isSome():
-    where &= "senderTimestamp >= (?)"
+    where &= "receiverTimestamp >= (?)"
   if startTime.isSome() and endTime.isSome():
     where &= " AND "
   if endTime.isSome():
-    where &= "senderTimestamp <= (?)"
+    where &= "receiverTimestamp <= (?)"
   where &= ")"
   some(where)
```
LNSD, line 256: The SQLite store implementation now will query by `receiverTimestamp` field. 

This field will be renamed `storeTimestamp`  or `storedAt` in one of the subsequent PRs to avoid any future confusion.

tests/v2/test_message_store_queue.nim
```diff
@@ -88,34 +88,6 @@ procSuite "Sorted store queue":
     check:
       store.len == capacity
 
-  test "Sender time can't be more than MaxTimeVariance in future":
```
staheri14, line 91: It is deleted because of the support for zero sender timestamp?

tests/v2/test_message_store_sqlite.nim
```diff
@@ -25,13 +25,16 @@ const
 proc newTestDatabase(): SqliteDatabase =
   SqliteDatabase.init("", inMemory = true).tryGet()
 
+proc now(): Timestamp =
+  getNanosecondTime(getTime().toUnixFloat())
+
 proc getTestTimestamp(offset=0): Timestamp = 
-  Timestamp(getNanosecondTime(epochTime()))
+  Timestamp(getNanosecondTime(getTime().toUnixFloat() + offset.float))
```
staheri14, line 32: Curious why this replacement. I remember it was recommended to use `getTime()` instead of `epochTime()`, but thought maybe you had a different reason?

tests/v2/test_message_store_sqlite.nim
```diff
@@ -25,13 +25,16 @@ const
...
 
 proc fakeWakuMessage(
   payload = "TEST-PAYLOAD",
   contentTopic = DefaultContentTopic, 
-  ts = getNanosecondTime(epochTime())
+  ts = now()
```
staheri14, line 37: From the documentation, `now` does not return the epoch time, but we need epoch time for the sender timestamp.

staheri14 (commented):
Thanks for addressing the timestamp issue! I left some comments and will do a more thorough review tomorrow. 
If you could please give a summary of the changes in the PR + the rationale (e.g., why `Index.compute` calls are replaced with `computeDigest`) that would help for a speedy review process!

tests/v2/test_message_store_sqlite.nim
LNSD: From nim documentation for `epochTime()`:

> Gets time after the UNIX epoch (1970) **in seconds**. It is a float because sub-second resolution is likely to be supported (depending on the hardware/OS).


> **`getTime` should generally be preferred over this proc.**

We want sub-second precision which is only provided by `getTime()` as I have found out empirically.

tests/v2/test_message_store_sqlite.nim
LNSD: If you check the defined functions in this test suite module, there's a method called `now` added to improve readability readability:

```nim
proc now(): Timestamp =
  getNanosecondTime(getTime().toUnixFloat())
```

tests/v2/test_message_store_queue.nim
LNSD: No, this test case has been moved to the `test_waku_store.nim` module and extended with the following test cases:

```Nim
  asyncTest "it should store a message with no sender timestamp":
	# [...]

  asyncTest "it should not store a message with a sender time variance greater than max time variance (future)":
	# [...]

  asyncTest "it should not store a message with a sender time variance greater than max time variance (past)":
  # [...]
```

Since the logic, as this PRs title says, has been moved from the different store implementations to the waku store protocol.

LNSD:
> Thanks for addressing the timestamp issue! I left some comments and will do a more thorough review tomorrow. If you could please give a summary of the changes in the PR + the rationale (e.g., why `Index.compute` calls are replaced with `computeDigest`) that would help for a speedy review process!

As we now have more message store implementations, in this PR, I have changed the `MessageStore` interface dependence on the `Index` type when a message is inserted in the DB (see `message_store`'s interface `put()` API). And moved all the common insertion logic to the waku protocol module.

Now the messages digest is calculated before inserting it into the store. Calculating a hash is a computational "costly" operation, so the lesser number of times that we compute it, the better.

>  why `Index.compute` calls are replaced with `computeDigest`

It has been refactored. I have extracted it out of the compute function. And moved the `Index` object initialization inside the `StoreQueueRef.put()` function:

```Nim
method put*(store: StoreQueueRef, pubsubTopic: string, message: WakuMessage, digest: MessageDigest, receivedTime: Timestamp): MessageStoreResult[void] =
  let index = Index(pubsubTopic: pubsubTopic, senderTime: message.timestamp, receiverTime: receivedTime, digest: digest)
  let message = IndexedWakuMessage(msg: message, index: index, pubsubTopic: pubsubTopic)
  store.add(message)
```

Using all around `Index` is problematic:

* It leaked the `Index` implementation detail outside the `StoreQueueRef` module. This makes any code maintenance/modification/extension painful.
* When calling `put` on the message store, we provided the `pubsubTopic` twice: one through the index and another through the procedure parameter. With the potential consequences of having two different pub-sub topics if not used properly.
* Not all message store implementations should depend on the same `Index` type. Different technologies will have different ways of doing indexing (e.g., SQLite vs. RbTree is a clear example).

Consequently, one of the upcoming refactorings will decouple the Pagination `Index` type from the `StoreQueueRef` `Index` type. So different parts can evolve independently.

waku/v2/protocol/waku_store/protocol.nim
```diff
@@ -213,19 +225,26 @@ proc handleMessage*(w: WakuStore, pubsubTopic: string, msg: WakuMessage) =
     # The message is ephemeral, should not be stored
     return
   
+  if not isValidMessage(msg):
+    waku_store_errors.inc(labelValues = [invalidMessage])
```
jm-clius, line 229: Not necessarily for this PR: I know this messes with the very clean boolean logic, but in practice messages usually arrive "late" or "early" for quite different reasons and it may be worth differentiating between these two invalid cases for debugging purposes. If the majority of messages arrive "late", this may be a case of network congestion. If the majority of messages arrive "early", it may be that our node is lagging, struggling to keep up or that you require an NTP sync,

waku/v2/protocol/waku_store/protocol.nim
```diff
@@ -213,19 +225,26 @@ proc handleMessage*(w: WakuStore, pubsubTopic: string, msg: WakuMessage) =
...
   let insertStartTime = getTime().toUnixFloat()
-    
-  let now = getNanosecondTime(getTime().toUnixFloat())
-  let index = Index.compute(msg, receivedTime=now, pubsubTopic=pubsubTopic)
+  
+  block:
+    let
+      msgDigest = computeDigest(msg) 
+      msgReceivedTime = if msg.timestamp > 0: msg.timestamp
+                        else: getNanosecondTime(getTime().toUnixFloat())
```
jm-clius, line 239: ~Not sure of real impact here, but I think that getting current epoch time is a relatively expensive operation. Should we consider doing this once at the start of `handleMessage` and use it both for the validity check and received time?~

On second thought, probably not that important as this second `getNanosecondTime()` will only be called in the presumably rare case when `msg.timestamp == 0`

jm-clius (approved):
Thanks, Lorenzo! I think this is the simplest and most consistent solution - using a single timestamp for storage which in 99% cases will be equal to sender timestamp. LGTM (though it would be good to see `receiverTimestamp` renamed to `storeTimestamp` in future.)

waku/v2/protocol/waku_store/protocol.nim
LNSD: I did not think about that possibility, thanks for that. It is a very good point to differentiate why a message is invalid. We never get enough metrics about how things are performing.

I will add it to my TODO list and tackle it in one of the upcoming PRs as it is low-hanging fruit.

LNSD:
> LGTM (though it would be good to see `receiverTimestamp` renamed to `storeTimestamp` in future.)

Waiting for this PR to be merged to open that PR 😉

rymnc (approved):
awesome, LGTM!

LNSD:
@staheri14 I am merging this PR since we need this and the next one (#1183) to be part of Friday's release.

Feel free to write your comments here, and I will gladly answer your questions and address your change suggestions in another PR.

PR #1181, opened 2022-09-22
fix: fail to start switch with ext ip

jm-clius:
Fixes regression introduced when submodules were updated.

In order for the node to "advertise" external, announced addresses during the `identify` procedure, the switch's addresses need to be explicitly updated.  However, the switch should be started on the bind address before this is done.

This resulted in `TransportError` being raised when starting the switch with an `extip`.


rymnc (approved):
great catch, LGTM!

Issue #1182, opened 2022-09-22
chore(rln-relay): metrics collection


staheri14:
### Problem
This issue is to collect performance metrics of the waku-rln-relay protocol. 
This would give us more insights into the network status and the protocol performance in general.

### Acceptance criteria
The following metrics are collected and logged periodically e.g., every minute.

- [x] Number of messages published on the content topic of `/toy-chat/2/luzhou/proto` 
- [x] Number of non-spam messages published on the same content topic
- [x] Number of messages detected as spam within the same content topic
- [x] Number of incoming  messages with invalid proof field
- [x] Number of incoming messages with invalid Epoch
- [x] Number of incoming messages with stale tree root
- [x] Time required for mounting waku-rln-relay
- [x] Proof verification time (and average)
- [x] Proof generation time (and  average)
- [x] The number of times the proof verification is checked

### Possible Solutions
Can use nim-metrics  https://github.com/status-im/nim-metrics

### Research track
Rln-Track

#### Next
Visualizing the metrics collected on the waku test fleets on the waku dashboard.

staheri14:
@rymnc feel free to put yourself as the assignee of this issue.

rymnc:
@staheri14 should we log these metrics to stdout as well?

staheri14:
No need for stdout, I guess when you add periodic logging, they will be part of the logs, and that is fine.

PR #1185, opened 2022-09-26
feat(rln-relay): metrics

rymnc:
Should resolve https://github.com/status-im/nwaku/issues/1182
Before rln-testnet 2 it is important to gather metrics on the performance of waku-rln-relay for key things like -
- proof generation time
- number of spam messages detected
- number of valid messages detected
- proof verification time
etc. The rest of the metrics are tracked in the issue.

Additional changes:
- Renamed the constants from screaming case to PascalCase
- Minor refactor of constants 

Note: I do understand this PR has a refactor as well as a feature, but this sets up the other PRs that have to do with the maintenance of waku-rln-relay. Thanks!


waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
```diff
@@ -1098,7 +1116,7 @@ proc readPersistentRlnCredentials*(path: string) : RlnMembershipCredentials {.ra
   debug "Deserialized Rln credentials", rlnCredentials=deserializedRlnCredentials
   result = deserializedRlnCredentials
 
-proc mountRlnRelay*(node: WakuNode, conf: WakuNodeConf|Chat2Conf, spamHandler: Option[SpamHandler] = none(SpamHandler), registrationHandler: Option[RegistrationHandler] = none(RegistrationHandler)): RlnRelayResult[bool] {.raises: [Defect, ValueError, IOError, CatchableError, Exception].} =
+proc mount(node: WakuNode, conf: WakuNodeConf|Chat2Conf, spamHandler: Option[SpamHandler] = none(SpamHandler), registrationHandler: Option[RegistrationHandler] = none(RegistrationHandler)): RlnRelayResult[bool] {.raises: [Defect, ValueError, IOError, CatchableError, Exception].} =
```
jm-clius, line 1119: Could we perhaps make this proc def multiline with indents (and the same for `mountRlnRelay` further down)?

waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
```diff
@@ -1183,3 +1201,15 @@ proc mountRlnRelay*(node: WakuNode, conf: WakuNodeConf|Chat2Conf, spamHandler: O
       if res.isErr:
         return err("dynamic rln-relay could not be mounted: " & res.error())
       return ok(true)
+
+
+proc mountRlnRelay*(node: WakuNode, conf: WakuNodeConf|Chat2Conf, spamHandler: Option[SpamHandler] = none(SpamHandler), registrationHandler: Option[RegistrationHandler] = none(RegistrationHandler)): RlnRelayResult[bool] {.raises: [Defect, ValueError, IOError, CatchableError, Exception].} =
+  waku_rln_relay_mounting_time.time:
```
jm-clius, line 1207: For interest sake, if we are splitting the `mount` proc just to measure the time to mount, what are the individual actions during mounting that we believe will take a long time? Registration? Would it make sense to measure these items individually rather than the more vague "mount time"?

waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
```diff
@@ -1183,3 +1201,15 @@ proc mountRlnRelay*(node: WakuNode, conf: WakuNodeConf|Chat2Conf, spamHandler: O
       if res.isErr:
         return err("dynamic rln-relay could not be mounted: " & res.error())
       return ok(true)
+
+
+proc mountRlnRelay*(node: WakuNode, conf: WakuNodeConf|Chat2Conf, spamHandler: Option[SpamHandler] = none(SpamHandler), registrationHandler: Option[RegistrationHandler] = none(RegistrationHandler)): RlnRelayResult[bool] {.raises: [Defect, ValueError, IOError, CatchableError, Exception].} =
```
jm-clius, line 1206: Unrelated to this PR, but procs should not really raise `CatchableError` or `Exception`. If they are raised from an underlying module/procedure they should be caught and translated locally. This can be addressed in future as it's not related to adding metrics. :) See: https://status-im.github.io/nim-style-guide/errors.exceptions.html for more.

jm-clius (approved):
LGTM! Minor comments below, but can be merged.

waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
rymnc: You're right! registration is will be where most time is spent, While this may not necessarily be an important metric, we're measuring broadly the time to mount the relay, to see how it would affect the overall node startup time (IIUC). Unsure if we're measuring total node setup time.

cc: @staheri14

waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
rymnc: Addressed in [cec453f](https://github.com/status-im/nwaku/pull/1185/commits/cec453f9df432ed6bd898a31cdbcbf9c20c8861c)

waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
```diff
@@ -795,36 +799,50 @@ proc validateMessage*(rlnPeer: WakuRLNRelay, msg: WakuMessage,
...
     return MessageValidationResult.Invalid
 
   # verify the proof
   let
     contentTopicBytes = msg.contentTopic.toBytes
     input = concat(msg.payload, contentTopicBytes)
-    proofVerificationRes = rlnPeer.rlnInstance.proofVerify(input, msg.proof)
+
+  waku_rln_proof_verification.inc()
+  waku_rln_proof_verification_time.time:
```
staheri14, line 816: My understanding from the nim-metrics description of `time` is that it measures time in seconds. As verification is very fast and in the order of milliseconds, it may mean that the measured time is always 0. Wondering whether you have tested this?

waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
```diff
@@ -795,36 +799,50 @@ proc validateMessage*(rlnPeer: WakuRLNRelay, msg: WakuMessage,
...
+    debug "invalid message: message is spam", payload = string.fromBytes(msg.payload)
+    waku_rln_spam_messages.inc()
     return MessageValidationResult.Spam
 
   # insert the message to the log
   # the result of `updateLog` is discarded because message insertion is guaranteed by the implementation i.e.,
   # it will never error out
   discard rlnPeer.updateLog(msg)
   debug "message is valid", payload = string.fromBytes(msg.payload)
+  let rootLabel = $rlnPeer.validMerkleRoots.find(msg.proof.merkleRoot)
```
staheri14, line 842: That is an interesting metric, how would you think of using the `root` value in the analysis?

waku/v2/protocol/waku_rln_relay/waku_rln_relay_metrics.nim
```diff
@@ -0,0 +1,17 @@
...
+declarePublicCounter(waku_rln_spam_messages, "number of spam messages detected")
+declarePublicCounter(waku_rln_invalid_messages, "number of invalid messages detected", ["type"])
+# This metric will be useful in detecting the index of the root in the acceptable window of roots
+declarePublicCounter(waku_rln_valid_messages, "number of valid messages with their roots tracked", ["index"])
+declarePublicCounter(waku_rln_errors, "number of errors detected while operating the rln relay", ["type"])
+declarePublicCounter(waku_rln_proof_verification, "number of times the rln proofs are verified")
+
+# Timing metrics
+declarePublicHistogram(waku_rln_proof_verification_time, "time taken to verify a proof")
+declarePublicGauge(waku_rln_relay_mounting_time, "time taken to mount the waku rln relay")
```
staheri14, line 16: Why not use the same collection type for all the time values i.e., right now, the verification time and proof generation are of `histogram` type and the mounting time is a `Gauge` type,  why are they different?
Also, have you considered using [summary](https://github.com/status-im/nim-metrics#summary) instead for calculating the average time? Does histogram calculate average? (I guess not)

waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
staheri14: Thanks for the comment @jm-clius. As the mounting procedure is done only once, we are mostly interested its overall running time. Yet, I agree we can have more fine-grained metrics for other specific parts. However, the registration is somewhat out of the control of the rln-relay as it is impacted by the underlying Blockchain traffic and the network throughput, thus even if we know how long it takes, there is not much we can do about it. Nevertheless, it would be good to understand what portion of the setup time is due to registration or other internal mechanics of the rln-relay.
@rymnc please do not delete the overall mounting time, but please please add the following new metrics to the PR:
- The registration time 
- The time required to create an instance of an RLN (it is to measure the Merkle tree performance)
- The time required for the insertion of a new commitment to the tree, with the index of the commitment (it is to measure the Merkle tree performance)
- The time for importing credentials from the file

waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
rymnc: This is an oversight on my end, will resolve!

waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
rymnc: It would be interesting to know how often peers are receiving messages with roots older than the most recent one.

waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
rymnc: Added a `granularTime` template that resolves this. Ideally, we would move this template up to a common utils file so that the other protocols can use it. - [e18de66](https://github.com/status-im/nwaku/pull/1185/commits/e18de6673aa6501443390d3ccef7ff4ef267db0d)

waku/v2/protocol/waku_rln_relay/waku_rln_relay_metrics.nim
rymnc: `waku_rln_relay_mounting_time` is a gauge since mounting happens only once in the lifetime of the process. The others run frequently throughout the lifetime of the process

rymnc: However, since nodes are allowed to be online sporadically, it could be useful to track it in a histogram instead. Thanks for the tip. 
Addressed in [c30d7aa](https://github.com/status-im/nwaku/pull/1185/commits/c30d7aa3ea2946bd835da194630fd7692745192f)

waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
rymnc: Addressed in [676c689](https://github.com/status-im/nwaku/pull/1185/commits/676c689b4afaaac698a159eb4b371e750c704e1c)

waku/v2/protocol/waku_rln_relay/waku_rln_relay_metrics.nim
```diff
@@ -0,0 +1,30 @@
...
+declarePublicCounter(waku_rln_proof_verification, "number of times the rln proofs are verified")
+
+# Timing metrics
+declarePublicHistogram(waku_rln_proof_verification_seconds, "time taken to verify a proof")
+declarePublicHistogram(waku_rln_relay_mounting_seconds, "time taken to mount the waku rln relay")
+declarePublicHistogram(waku_rln_proof_generation_seconds, "time taken to generate a proof")
+declarePublicHistogram(waku_rln_registration_seconds, "time taken to register to a rln membership set")
+declarePublicHistogram(waku_rln_instance_creation_seconds, "time taken to create an rln instance")
+declarePublicHistogram(waku_rln_membership_insertion_seconds, "time taken to insert a new member into the local merkle tree")
+declarePublicHistogram(waku_rln_membership_credentials_import_seconds, "time taken to import membership credentials")
```
LNSD, line 22: All of these are durations, so following Jakub's advice, the metric names should finish with `*_duration_seconds`

LNSD: This is the Prometheus _best practices and recommendations_ page: https://prometheus.io/docs/practices/naming/

waku/v2/protocol/waku_rln_relay/waku_rln_relay_metrics.nim
```diff
@@ -0,0 +1,30 @@
...
+
+export metrics
+
+declarePublicCounter(waku_rln_messages, "number of messages published on the rln content topic")
+declarePublicCounter(waku_rln_spam_messages, "number of spam messages detected")
+declarePublicCounter(waku_rln_invalid_messages, "number of invalid messages detected", ["type"])
+# This metric will be useful in detecting the index of the root in the acceptable window of roots
+declarePublicCounter(waku_rln_valid_messages, "number of valid messages with their roots tracked", ["index"])
+declarePublicCounter(waku_rln_errors, "number of errors detected while operating the rln relay", ["type"])
+declarePublicCounter(waku_rln_proof_verification, "number of times the rln proofs are verified")
```
LNSD, line 13: Let's follow Prometheus' best practices when defining metric names: http_requests_total (for a unit-less accumulating count)

I will have to do it in waku store sooner than later. But now is the right moment for the RLN relay 👍🏼

LNSD (approved):
LGTM ✅

One comment: please, invite @jakubgs to this PR so he can provide you feedback

waku/v2/protocol/waku_rln_relay/waku_rln_relay_metrics.nim
rymnc: Addressed in [2a32c82](https://github.com/status-im/nwaku/pull/1185/commits/2a32c82d6e68db06514570c6939afe70bd59fa93)

waku/v2/protocol/waku_rln_relay/waku_rln_relay_metrics.nim
rymnc: Addressed in [2a32c82](https://github.com/status-im/nwaku/pull/1185/commits/2a32c82d6e68db06514570c6939afe70bd59fa93)

waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
```diff
@@ -795,36 +806,50 @@ proc validateMessage*(rlnPeer: WakuRLNRelay, msg: WakuMessage,
...
+  waku_rln_valid_messages_total.inc(labelValues=[
+    rootLabel
```
jakubgs, line 851: What is the cardinality of this label? High cardinality labels have terrible effect on query performance and any prometheus backend existing today. Labels should have a capped number of possible values not exceeding hundreds. Label values that are uncapped can easily explode and cause major performance issues for the whole cluster, as there's no easy ways to clean up specific metrics with specific labels.

What are the example values of `rootLabel`? How man of those values can be? Is there a limit?

rymnc: Examples of root label: 0,1,2..AcceptableRootWindowSize

Ideally we would add a check to make sure that the operator does not set the AcceptableRootWindowSize to a very large number, but due to the nature of the validations, we can expect any value between 0 & 4.

jakubgs: Yes, some validation would be good if it is actually possible to set this to any number, since that could easily be used to kill our(or any) metrics backend with an absurd number of labels.

rymnc: Addressed in [0656a0a](https://github.com/status-im/nwaku/pull/1185/commits/0656a0aeb3afc11427e9c33048f76692872a1da6). Not sure if it's the most nim way to do it. 

cc: @LNSD / @jm-clius wdyt?

rymnc: Addressed properly in [ee1ee0a](https://github.com/status-im/nwaku/pull/1185/commits/ee1ee0a31b7835c6d46c84857129d994011edd4d), specifically - 


https://github.com/status-im/nwaku/blob/ee1ee0a31b7835c6d46c84857129d994011edd4d/waku/v2/protocol/waku_rln_relay/waku_rln_relay_metrics.nim#L7-L14 and https://github.com/status-im/nwaku/blob/ee1ee0a31b7835c6d46c84857129d994011edd4d/waku/v2/protocol/waku_rln_relay/waku_rln_relay_metrics.nim#L20-L22

rymnc: @jakubgs, is this solution acceptable?

jm-clius (approved):
In general, lgtm! I think we need to decide in future whether the `nanosecondTime` template belongs to the `time` utility or should rather go to a new `metrics` util. Let's get this merged though, if the other reviewers do not have major concerns.

rymnc:
Can confirm that the metrics are being exported correctly -
image
re: regression fixed in [e23a3cd](https://github.com/status-im/nwaku/pull/1185/commits/e23a3cd6c5539ca54718ffbd0658ec89c9f0fd80)

PR #1192, opened 2022-09-28
Makefile: allow overriding GIT_VERSION with env

jakubgs:
This is necessary for Nix builds because the sandbox there does not contain `.git` directory or `git` command.

Also the quotes are pointless.


LNSD (approved):
LGTM ✅

jm-clius (approved):

PR #1209, opened 2022-09-30
chore(rln): Update zerokit submodule and RLN APIs

s1fr0:
This PRs updates the zerokit RLN module used in nwaku to the latest version.

A new API `seeded_key_gen` was added in zerokit, hence the nwaku RLN bindings were updated as well to support this new API.

cc @fryorcraken


rymnc (approved):
LGTM, thanks 😁

PR #1211, opened 2022-10-01
refactor(store): simplify store queue query implementation

LNSD:

- [ ] Update SQLite DB table primary key
- [ ] Decouple SQLite queries module from pagination types
- [ ] Move page info logic to waku store protocol

In this PR:

- Unified into `getPage` the `fwdPage` and the `bwdPage` methods
- Replace the returned tuple with a result type for better error handling
- Update test suites accordingly


rymnc (approved):
LGTM 😁

PR #1212, opened 2022-10-01
fix(store): update sqlite db table primary key and decouple queries module from pagination types

LNSD:

- [x] Update SQLite DB table primary key
- [x] Decouple SQLite queries module from pagination types
- [ ] Move page info logic to waku store protocol

In this PR:

- Update SQLite DB table primary key
- Rename `senderTimestamp` column to `timestamp`
- Extend history query order by clause for results consistency
- Decouple `sqlite_store/queries` module from `pagination` module types
- Update test cases accordingly

### DIFF
--- a/waku/v2/node/storage/message/sqlite_store/queries.nim
+++ b/waku/v2/node/storage/message/sqlite_store/queries.nim
@@ -7,13 +7,14 @@ import
 import
   ../../sqlite, 
   ../../../../protocol/waku_message,
-  ../../../../protocol/waku_store/pagination,
   ../../../../utils/time
 
 
 const DbTable = "Message"
 
 type SqlQueryStr = string
+
+type DbCursor* = (Timestamp, seq[byte], string)
 
 
 ### SQLite column helper methods
@@ -227,7 +228,7 @@ proc contentTopicWhereClause(contentTopic: Option[seq[ContentTopic]]): Option[st
   contentTopicWhere &= ")"
   some(contentTopicWhere)
 
-proc cursorWhereClause(cursor: Option[PagingIndex], ascending=true): Option[string] =
+proc cursorWhereClause(cursor: Option[DbCursor], ascending=true): Option[string] =
   if cursor.isNone():
     return none(string)
 
@@ -292,7 +293,7 @@ proc prepareSelectMessagesWithlimitStmt(db: SqliteDatabase, stmt: string): Datab
 proc execSelectMessagesWithLimitStmt(s: SqliteStmt, 
                           contentTopic: Option[seq[ContentTopic]], 
                           pubsubTopic: Option[string],
-                          cursor: Option[PagingIndex],  
+                          cursor: Option[DbCursor],  
                           startTime: Option[Timestamp],
                           endTime: Option[Timestamp],
                           onRowCallback: DataProc): DatabaseResult[void] =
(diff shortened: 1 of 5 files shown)


waku/v2/node/storage/message/sqlite_store/queries.nim
```diff
@@ -57,14 +58,14 @@ proc queryRowPubsubTopicCallback(s: ptr sqlite3_stmt, pubsubTopicCol: cint): str
...
-  " id BLOB," &
-  " storedAt INTEGER NOT NULL," &
-  " contentTopic BLOB NOT NULL," &
   " pubsubTopic BLOB NOT NULL," &
+  " contentTopic BLOB NOT NULL," &
   " payload BLOB," &
   " version INTEGER NOT NULL," &
-  " senderTimestamp INTEGER NOT NULL," &
-  " CONSTRAINT messageIndex PRIMARY KEY (senderTimestamp, id, pubsubTopic)" &
+  " timestamp INTEGER NOT NULL," &
```
rymnc, line 65: What does `timestamp` indicate? as per my understanding of the discussion around senderTimestamps, should this field be nullable?

LNSD: I thought the same, but when the field is not present in the protobuffer, it is equal to zero. So it won't be null.

BTW, I renamed the column to match the waku message field for consistency reasons.

rymnc: Ah! okay. Can we then make this field nullable, and if `timestamp` == 0, set it to null (would be 1 byte vs 4 bytes)?

rymnc: This can be completed in a follow up PR rather, ignore!

rymnc (approved):
LGTM, thanks!😁

LNSD:
Doing a manual local test, I noticed that migration 0007 failed in the last statement. I am investigating.

waku/v2/node/storage/migration/migrations_scripts/message/00007_updatePrimaryKey.up.sql
```diff
@@ -0,0 +1,18 @@
...
+  contentTopic BLOB NOT NULL,
+  payload BLOB,
+  version INTEGER NOT NULL,
+  timestamp INTEGER NOT NULL,
+  id BLOB,
+  storedAt INTEGER NOT NULL,
+  CONSTRAINT messageIndex PRIMARY KEY (storedAt, id, pubsubTopic)
+) WITHOUT ROWID;
+
+INSERT OR IGNORE INTO message(pubsubTopic, contentTopic, payload, version, timestamp, id, storedAt)
```
LNSD, line 14: Added the `OR IGNORE` clause to skip the stored messages that do not comply with the new constraint during the migration to the new scheme.

LNSD: In the local database copy that I have, with schema version 4, there were ~7000 messages that did not comply with the new constraint. With the ignore clause, those messages are dropped. For now, this is acceptable since we are not sure why those messages are not complying with the new contraint

LNSD: The Previous constraint was:
```SQL
  CONSTRAINT messageIndex PRIMARY KEY (senderTimestamp, id, pubsubTopic)
```
That constriant allowed that some messages had the same `receivedTimestamp` (now renamed to `storedAt`).

-------

With the new constraint as:
```SQL
  CONSTRAINT messageIndex PRIMARY KEY (storedAt, id, pubsubTopic)
```
While migrating and inserting the previous messages in the new table (with the new constraint), the messages with the same `storedAt` value failed; making the migration fail.

------

The question here is: _under which circumstances did those messages got inserted in the database._ 

In previous releases (v0.9 and v0.10) the `receivedTimestamp` was set to `getNanosecondTime(getTime().toUnixFloat())`. And with a nanosecond resolution, it is disturbing how the uniqueness constraint failed.

PR #1215, opened 2022-10-03
fix(store): Move page info logic to waku store protocol

LNSD:

- [x] Update SQLite DB table primary key
- [x] Decouple SQLite queries module from pagination types
- [x] Move page info logic to waku store protocol

In this PR:

- Simplify the message store interface and update message store implementations accordingly
- Move paging info calculation to the protocol module
- Fix pagination logic: `Always search for _max page size_ + 1. If the extra row does not exist, do not return pagination info.` 
- Update test cases accordingly

### DIFF
--- a/waku/v2/protocol/waku_store/protocol.nim
+++ b/waku/v2/protocol/waku_store/protocol.nim
@@ -4,7 +4,7 @@
 {.push raises: [Defect].}
 
 import
-  std/[tables, times, sequtils, options],
+  std/[tables, times, sequtils, options, algorithm],
   stew/results,
   chronicles,
   chronos, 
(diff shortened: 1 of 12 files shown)


LNSD:
Some lower priority work is pending:

- [ ] Review the waku store protocol and message store test cases. Remove the redundant test cases and extend the test coverage.
- [ ] Add test coverage for the message retention policies

waku/v2/protocol/waku_store/protocol.nim
```diff
@@ -135,11 +135,46 @@ proc findMessages(w: WakuStore, query: HistoryQuery): HistoryResponse {.gcsafe.}
...
+  if rows.len <= 0:
+    return HistoryResponse(messages: @[], error: HistoryResponseError.NONE)
+  
+  var messages = if rows.len <= int(qMaxPageSize): rows.mapIt(it[1])
+                 else: rows[0..^2].mapIt(it[1])
+  var pagingInfo = none(PagingInfo)
+
+  # The retrieved messages list should always be in chronological order
+  if not qAscendingOrder:
+    messages.reverse()
```
jm-clius, line 149: Not the main benefit of moving paging logic to the protocol, but I like that the message stores don't have to take care of this protocol-specific idiosyncrasy. 👍

jm-clius (approved):
Great work!

PR #1223, opened 2022-10-04
nix: don't pull in gcc explicitly, stdenv has it

jakubgs:
If we pull it explicitly we affect variables like `CC_FOR_TARGET` which can break builds on MacOS due to need for `clang` use.


jakubgs:
Looks like it's failing on macOS with:
```
13:05:09  + nix-shell --show-trace --run 'make V=1 v2' --pure --no-out-link --keep LOCALE_ARCHIVE_2_27 --option sandbox relaxed /Users/jenkins/workspace/nim-waku_prs_macos_PR-1223/shell.nix
13:05:11  which gcc &>/dev/null || { echo "C compiler (gcc) not installed. Aborting."; exit 1; }
13:05:11  C compiler (gcc) not installed. Aborting.
13:05:11  make: *** [vendor/nimbus-build-system/makefiles/targets.mk:41: sanity-checks] Error 1
```
Which is caused by this `sanity-check` target:
```Makefile
sanity-checks:
	which $(CC) &>/dev/null || { echo "C compiler ($(CC)) not installed. Aborting."; exit 1; }
```
https://github.com/status-im/nimbus-build-system/blob/c7977002/makefiles/targets.mk#L40-L41
Which in turn is caused by this assignment:
```Makefile
CC := gcc
```
https://github.com/status-im/nimbus-build-system/blob/c7977002/makefiles/variables.mk#L10

jakubgs:
Attempting to fix this in:

* https://github.com/status-im/nimbus-build-system/pull/51

By using `?=` assignment so `CC` can be overridden by environment.

jm-clius (approved):

PR #1216, opened 2022-10-03
[WIP] ci: use Nix shell to provide build dependencies

jakubgs:
Because we need Rust version newer than `1.62.0` and Ubuntu 22.04 only comes with `1.61.0`:
https://packages.ubuntu.com/jammy/rustc

And locking dependencies using Nix would make it easier for developers to track them and update them:
https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/compilers/rust/1_63.nix

PR #1230, opened 2022-10-05
ci: install 1.64 rust in Docker image

jakubgs:
Necessary to fix errors like:
```
error: package `ethers-core v0.17.0 (https://github.com/gakonst/ethers-rs#454449b0)` cannot be built
because it requires rustc 1.62 or newer, while the currently active rustc version is 1.60.0
```
Also upgraded Alpine from `3.15` to `3.16`.

Related:

* https://github.com/status-im/nwaku/pull/1216

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,13 +1,21 @@
 # BUILD IMAGE --------------------------------------------------------
 
-FROM alpine:3.15 AS nim-build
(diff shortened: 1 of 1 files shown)


jm-clius (approved):
LGTM! @s1fr0 are all the external resources required by Zerokit during runtime included?

PR #1233, opened 2022-10-05
fix(store): fix waku store resume tests

LNSD:
This PR fixes the store resume test suite flakiness:

- Moved the time query tests under the store query test suite
- Split the remaining waku store resume test suite into another file
- Fixed both test suite test cases

It is in the interest of everybody to merge this PR before any other, so we don't suffer from more CI failures due to these test cases.


jm-clius (approved):
Thanks! Approving this in my own interest. :)

rymnc (approved):
LGTM, thanks! 😁

Issue #1125, opened 2022-09-07
chore: add `version` gauge to metrics


jm-clius:
### Background

To increase the visibility of the nwaku `version` from a monitoring perspective, we need to add this as a metric label somewhere. We could follow the same route as nimbus and simply [declare a gauge specifically for this purpose](https://github.com/status-im/nimbus-eth2/blob/3a8abd60108ff94b340620e2b12fd2cb7229ffbe/beacon_chain/nimbus_beacon_node.nim)

### Acceptance criteria
- [ ] nwaku version and/or commit hash logged as label in metric
- [ ] Grafana dashboard updated to make this visible

jakubgs:
Still waiting...

PR #1234, opened 2022-10-05
chore: publish nwaku version to metric

jm-clius:
Closes https://github.com/status-im/nwaku/issues/1125.

Publishes the nwaku `git_version` to a metric with the label value matching the version. For now the label format matches the `git describe` format, though we should consider more explicit, parsed versioning information similar to nimbus in future.


LNSD (approved):
LGTM ✅

jakubgs (approved):
Thanks! It's a start. A separate `version` and `commit` would be even better tho.
But we can extend it later, this should give us better insight into deployments in Grafana.

PR #1237, opened 2022-10-05
WIP: zerokit as static library

s1fr0:


tests/all_tests_v2.nim
```diff
@@ -1,47 +1,47 @@
-import
+#import
```
jm-clius, line 1: I think this file should not have been commited :D

PR #1239, opened 2022-10-06
ci: use rust:1.64.0-slim-bullseye image without musl

jakubgs:
This is necessary for generating the `zerokit` `librln` dynamic library. Otherwise we get:
```
warning: dropping unsupported crate type `cdylib` for target `x86_64-unknown-linux-musl`
```

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,7 +1,11 @@
-/README.md
-/Dockerfile
-/.*ignore
-/LICENSE*
-/tests
-/metrics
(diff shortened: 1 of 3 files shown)


rymnc:
> @rymnc you know, I am also reading the same Discord chat as you.

context for someone new to the conversation

jm-clius:
@rymnc @jakubgs is this still relevant or can we close without merging?

PR #1244, opened 2022-10-06
fix(store): fix store validity timestamp variance constant

LNSD:
Fix the constant definition. Seems that using chronos' `seconds` and `nanoseconds` procs set the constant to 0. The changes revert to the original way of defining the constant.


jm-clius (approved):

jm-clius:
(We may want to make sure this is covered by unit testing - not sure how it fell through, as we have a lot of store unit tests)

PR #1249, opened 2022-10-07
fix(rln-relay): replaces verify_rln_proof with verify

staheri14:


rymnc (approved):
LGTM, thanks 🥳

oskarth:
I believe @staheri14 is off on vacation, so could you guys help get PR merged so we can get live demo working? @s1fr0 @rymnc 

If there are problems with cluster deployment (saw something about it not being updated?) @LNSD can probably assist

PR #1250, opened 2022-10-07
feat(RLN): Add verify against multiple roots

s1fr0:
This PR initiates the implementation of RLN proof verification against multiple acceptable roots, by using the new zerokit API (added in this PR) `verify_with_roots`.

At the moment `proofVerify` takes as default an empty sequence of acceptable roots and this internally implies that any check on the validity of the RLN proof's root value is skipped. In other words, **any RLN proof with valid proof values except the root will be successfully verified** by default.

In follow-up PRs, all calls to `proofVerify` should be done by properly passing the acceptable roots sequence, hence enforcing the RLN proof's root to be checked against these values.

Note that since now root verification is skipped as default, the unit test "testing rln-relay with invalid proof" had to be changed since was previously invalidated by changing the membership index which affects only root (now, payload is changed). In case of the unit test "test proofVerify and proofGen for an invalid proof", it was possible instead to pass directly the internal root of the RLN object.

As additional change of this PR, the toBuffer proc is changed to account for empty buffers too.

### DIFF
--- a/tests/v2/test_waku_rln_relay.nim
+++ b/tests/v2/test_waku_rln_relay.nim
@@ -633,9 +633,10 @@ suite "Waku rln relay":
       proofRes.isOk()
     let proof = proofRes.value
 
-    # verify the proof (should not be verified)
+    # verify the proof (should not be verified) against the internal RLN tree root
     let verified = rln.proofVerify(data = messageBytes,
-                                  proof = proof)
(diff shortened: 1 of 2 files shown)


richard-ramos (approved):
good work!

rymnc (approved):
LGTM, thanks 😁

PR #1259, opened 2022-10-11
refactor(rln): display byte arrays in hex as uints and fix Kilic's proofVerify proc

s1fr0:
After the recent [refactor/fix](https://github.com/status-im/nwaku/pull/1256) that changed the encoding of ID commitments to LE, this PR implements a `inHex` proc to print 32 bytes arrays (that is most of RLN proof fields) to hex representation corresponding to the uint value of the original field (before, bytes were shown in reverse order).

This will ease debugging (especially by looking at logs) and allows direct comparison with values registered in the membership smart contract.

This PR further fixes a disalignament with the `proofVerify` API used when Kilic's RLN is used: a stub `validRoots` empty field is added to make tests pass (`verify` API in Kilic's lib doesn't support verification against a set of roots).


rymnc (approved):
LGTM!

Issue #1203, opened 2022-09-30
chore(chat2): enable metrics


rymnc:
### Background
The metricsServer and metricsLogging flags are available as configuration parameters but are not setup in the chat2 initialization procedure. Metrics would be useful for those running chat2.

### Details
Enable metricsLogging, similar to -
https://github.com/status-im/nwaku/blob/master/waku/v2/node/wakunode2.nim#L1160-L1161

### Acceptance criteria
- [x] Create a procedure that both wakunode2 and chat2 can use to periodically log rln metrics. 
  - https://github.com/status-im/nwaku/pull/1262
- [ ] Enable the metrics server for chat2
  - https://github.com/status-im/nwaku/pull/1279

staheri14:
Is this issue in progrees? if yes, then please move it to the `in progress` column.

rymnc:
Yes, this issue is in progress. The second step is a *little* more involved, and should be resolved soon

PR #1262, opened 2022-10-13
feat(rln-relay): periodically log metrics

rymnc:
This PR serves as the initial groundwork to allow metric logging on chat2 (https://github.com/status-im/nwaku/issues/1203)

Creates a `startRlnMetricsLog` procedure which periodically logs out certain counter-based metrics, like -
- total messages
- total spam messages
- total invalid messages
- total valid messages
- total errors
- total proofs verified

Also creates a template to avoid reuse of logic to get the metrics only for the last `period`

This proc is then imported into `wakunode_setup_metrics`, and called after.

### DIFF
--- a/waku/v2/node/wakunode2_setup_metrics.nim
+++ b/waku/v2/node/wakunode2_setup_metrics.nim
@@ -49,9 +49,9 @@ proc startMetricsLog*() =
       let totalErrors = parseCollectorIntoF64(waku_node_errors)
       let totalConnections = parseCollectorIntoF64(waku_node_conns_initiated)
 
-      # track cumulative, and then max.
-      let freshErrorCount = max(totalErrors - cumulativeErrors, 0)
-      let freshConnCount = max(totalConnections - cumulativeConns, 0)
+      # track cumulative values
+      let freshErrorCount = totalErrors - cumulativeErrors
+      let freshConnCount = totalConnections - cumulativeConns
       
       cumulativeErrors = totalErrors
       cumulativeConns = totalConnections
@@ -71,5 +71,6 @@ proc startMetricsLog*() =
 
   # Start protocol specific metrics logging
   when defined(rln) or defined(rlnzerokit):
-    logRlnMetrics()
+    let logRlnMetrics = getRlnMetricsLogger()
+    logRlnMetrics(nil)
   
--- a/waku/v2/node/wakunode2_setup_metrics.nim
+++ b/waku/v2/node/wakunode2_setup_metrics.nim
@@ -41,6 +41,9 @@ proc startMetricsLog*() =
   var cumulativeErrors = 0.float64
   var cumulativeConns = 0.float64
 
+  when defined(rln) or defined(rlnzerokit):
+    let logRlnMetrics = getRlnMetricsLogger()
+
   logMetrics = proc(udata: pointer) =
     {.gcsafe.}:
       # TODO: libp2p_pubsub_peers is not public, so we need to make this either
@@ -60,12 +63,12 @@ proc startMetricsLog*() =
       info "Total errors", count = freshErrorCount
       info "Total active filter subscriptions", count = parseCollectorIntoF64(waku_filter_subscribers)
 
+      # Start protocol specific metrics logging
+      when defined(rln) or defined(rlnzerokit):
+        logRlnMetrics()
+
     discard setTimer(Moment.fromNow(30.seconds), logMetrics)
   
   discard setTimer(Moment.fromNow(30.seconds), logMetrics)
 
-  # Start protocol specific metrics logging
-  when defined(rln) or defined(rlnzerokit):
-    let logRlnMetrics = getRlnMetricsLogger()
-    logRlnMetrics(nil)
   
--- a/waku/v2/node/wakunode2_setup_metrics.nim
+++ b/waku/v2/node/wakunode2_setup_metrics.nim
@@ -71,5 +71,5 @@ proc startMetricsLog*() =
 
   # Start protocol specific metrics logging
   when defined(rln) or defined(rlnzerokit):
-    startRlnMetricsLog()
+    logRlnMetrics()
   
--- a/waku/v2/node/wakunode2_setup_metrics.nim
+++ b/waku/v2/node/wakunode2_setup_metrics.nim
@@ -14,7 +14,9 @@ import
   ../protocol/waku_store,
   ../protocol/waku_lightpush,
   ../protocol/waku_swap/waku_swap,
-  ../protocol/waku_peer_exchange
+  ../protocol/waku_peer_exchange,
+  ../protocol/waku_rln_relay/waku_rln_relay_metrics,
+  ../utils/collector
 
 logScope:
   topics = "wakunode.setup.metrics"
@@ -30,14 +32,7 @@ proc startMetricsServer*(serverIp: ValidIpAddress, serverPort: Port) =
 
     info "Metrics HTTP server started", serverIp, serverPort
 
-proc parseCollectorIntoF64(collector: Collector): float64 = 
-  var total = 0.float64
-  for key in collector.metrics.keys():
-    try:
-      total = total + collector.value(key)
-    except KeyError:
-      discard
-  return total
+
 
 proc startMetricsLog*() =
   # https://github.com/nim-lang/Nim/issues/17369
@@ -73,4 +68,8 @@ proc startMetricsLog*() =
     discard setTimer(Moment.fromNow(30.seconds), logMetrics)
   
   discard setTimer(Moment.fromNow(30.seconds), logMetrics)
+
+  # Start protocol specific metrics logging
+  when defined(rln) or defined(rlnzerokit):
+    startRlnMetricsLog()
   
--- a/waku/v2/node/wakunode2_setup_metrics.nim
+++ b/waku/v2/node/wakunode2_setup_metrics.nim
@@ -46,15 +46,10 @@ proc startMetricsLog*() =
       # TODO: libp2p_pubsub_peers is not public, so we need to make this either
       # public in libp2p or do our own peer counting after all.
 
-      let totalErrors = parseCollectorIntoF64(waku_node_errors)
-      let totalConnections = parseCollectorIntoF64(waku_node_conns_initiated)
-
       # track cumulative values
-      let freshErrorCount = totalErrors - cumulativeErrors
-      let freshConnCount = totalConnections - cumulativeConns
+      let freshErrorCount = parseAndAccumulate(waku_node_errors, cumulativeErrors)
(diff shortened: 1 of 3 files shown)


waku/v2/node/wakunode2_setup_metrics.nim
```diff
@@ -73,4 +68,8 @@ proc startMetricsLog*() =
     discard setTimer(Moment.fromNow(30.seconds), logMetrics)
   
   discard setTimer(Moment.fromNow(30.seconds), logMetrics)
+
+  # Start protocol specific metrics logging
+  when defined(rln) or defined(rlnzerokit):
+    startRlnMetricsLog()
```
jm-clius, line 74: Maybe I'm missing something obvious, but why not log the RLN metrics with the other metrics? I.e. on the same timer as above (under `logMetrics`), you just do:

```suggestion
  when defined(rln) or defined(rlnzerokit):
    logRlnMetrics()
```

jm-clius: In other words, I don't see the need for a separately timed task

rymnc: Agreed, will fix soon

waku/v2/protocol/waku_rln_relay/waku_rln_relay_metrics.nim
```diff
@@ -31,3 +42,48 @@ declarePublicGauge(waku_rln_registration_duration_seconds, "time taken to regist
 declarePublicGauge(waku_rln_instance_creation_duration_seconds, "time taken to create an rln instance")
 declarePublicGauge(waku_rln_membership_insertion_duration_seconds, "time taken to insert a new member into the local merkle tree")
 declarePublicGauge(waku_rln_membership_credentials_import_duration_seconds, "time taken to import membership credentials")
+
+template parseAndAccumulate(collector: Collector, cumulativeValue: float64): float64 =
```
rymnc, line 46: @jm-clius , do you think this would be useful in the utils directory? It can be used in `wakunode_setup_metrics.nim`: https://github.com/status-im/nwaku/blob/371016b04fc8467f6a4c09564b26d96d1acd0999/waku/v2/node/wakunode2_setup_metrics.nim#L54

waku/v2/node/wakunode2_setup_metrics.nim
rymnc: Addressed in [2a805e0](https://github.com/status-im/nwaku/pull/1262/commits/2a805e06ed02f591490bdfb7010353aed935e0bd)

rymnc: Maybe some protocols would like tighter control on how often their metrics are logged, that's why I've retained the `setTimer` for the rlnMetrics as well.

waku/v2/protocol/waku_rln_relay/waku_rln_relay_metrics.nim
```diff
@@ -1,9 +1,20 @@
...
+  chronos,
   metrics,
-  waku_rln_relay_constants
+  metrics/chronos_httpserver,
+  waku_rln_relay_constants,
+  ../../utils/collector
 
 export metrics
 
+const LogPeriod = 15
```
staheri14, line 13: Please indicate the unit.

waku/v2/protocol/waku_rln_relay/waku_rln_relay_metrics.nim
```diff
@@ -31,3 +42,48 @@ declarePublicGauge(waku_rln_registration_duration_seconds, "time taken to regist
 declarePublicGauge(waku_rln_instance_creation_duration_seconds, "time taken to create an rln instance")
 declarePublicGauge(waku_rln_membership_insertion_duration_seconds, "time taken to insert a new member into the local merkle tree")
 declarePublicGauge(waku_rln_membership_credentials_import_duration_seconds, "time taken to import membership credentials")
+
+template parseAndAccumulate(collector: Collector, cumulativeValue: float64): float64 =
+  let total = parseCollectorIntoF64(collector)
+  let freshCount = max(total - cumulativeValue, 0)
```
staheri14, line 48: Curious to know whether `total < cumulativeValue` can ever happen?

waku/v2/protocol/waku_rln_relay/waku_rln_relay_metrics.nim
```diff
@@ -31,3 +42,48 @@ declarePublicGauge(waku_rln_registration_duration_seconds, "time taken to regist
...
+      info "Total messages", count = freshMsgCount
+      info "Total spam messages", count = freshSpamCount
+      info "Total invalid messages", count = freshInvalidMsgCount
+      info "Total valid messages", count = freshValidMsgCount
+      info "Total errors", count = freshErrorCount
+      info "Total proofs verified", count = freshProofCount
+
+    discard setTimer(Moment.fromNow(LogPeriod.seconds), logMetrics)
+  
+  logRlnMetrics()
```
staheri14, line 88: Shouldn't it be `logMetrics()`?

waku/v2/protocol/waku_rln_relay/waku_rln_relay_metrics.nim
```diff
@@ -31,3 +42,48 @@ declarePublicGauge(waku_rln_registration_duration_seconds, "time taken to regist
...
 declarePublicGauge(waku_rln_membership_insertion_duration_seconds, "time taken to insert a new member into the local merkle tree")
 declarePublicGauge(waku_rln_membership_credentials_import_duration_seconds, "time taken to import membership credentials")
+
+template parseAndAccumulate(collector: Collector, cumulativeValue: float64): float64 =
+  let total = parseCollectorIntoF64(collector)
+  let freshCount = max(total - cumulativeValue, 0)
+  cumulativeValue = total
+  freshCount
+
+proc logRlnMetrics*() =
```
staheri14, line 52: We also need metrics logging in chat2. please add it there as well.

staheri14 (commented):
Thanks, Looks good, a couple of comments are inline.

waku/v2/protocol/waku_rln_relay/waku_rln_relay_metrics.nim
staheri14: Why not use the same timer duration as the other waku metrics logging i.e., `30`s?

staheri14 (commented):
I already added a comment in https://github.com/vacp2p/rfc/pull/539#discussion_r995056265
Once clarified and addressed, we are good to merge the PR.

waku/v2/protocol/waku_rln_relay/waku_rln_relay_metrics.nim
rymnc: Just providing flexibility if it needs to be changed later.

waku/v2/protocol/waku_rln_relay/waku_rln_relay_metrics.nim
rymnc: This will happen in a subsequent PR :) Tracked in https://github.com/status-im/nwaku/issues/1203

waku/v2/protocol/waku_rln_relay/waku_rln_relay_metrics.nim
rymnc: Addressed in [eb9101f](https://github.com/status-im/nwaku/pull/1262/commits/eb9101ff1b19478b500921cbb93433e2f140071c)

waku/v2/protocol/waku_rln_relay/waku_rln_relay_metrics.nim
rymnc: It cannot. Fixed in [eb9101f](https://github.com/status-im/nwaku/pull/1262/commits/eb9101ff1b19478b500921cbb93433e2f140071c)

waku/v2/protocol/waku_rln_relay/waku_rln_relay_metrics.nim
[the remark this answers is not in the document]
rymnc: Addressed in [eb9101f](https://github.com/status-im/nwaku/pull/1262/commits/eb9101ff1b19478b500921cbb93433e2f140071c)

jm-clius: Yeah, I think this can go to the collector util. :)

waku/v2/node/wakunode2_setup_metrics.nim
```diff
@@ -73,4 +68,9 @@ proc startMetricsLog*() =
     discard setTimer(Moment.fromNow(30.seconds), logMetrics)
   
   discard setTimer(Moment.fromNow(30.seconds), logMetrics)
+
+  # Start protocol specific metrics logging
+  when defined(rln) or defined(rlnzerokit):
+    let logRlnMetrics = getRlnMetricsLogger()
+    logRlnMetrics(nil)
```
jm-clius, line 75: Any reason not to move this into the metrics logging block above and removing the need for adding a separate timed task for RLN metrics?

waku/v2/node/wakunode2_setup_metrics.nim
jm-clius: Right, I think I don't quite see the need for this flexibility if we don't have any current reason for logging metrics for different protocols at different intervals. In my mind this just creates unnecessary complexity in the logic (and the possibility of somewhat confusing logs)? My pattern is to always go for the very simplest thing - this is easy to change to something a bit more intricate in future if the need arise, but changing complex code back to something simpler is much more difficult.

waku/v2/protocol/waku_rln_relay/waku_rln_relay_metrics.nim
jm-clius: See my comment above on flexibility. I don't think we should be eager to add "in case" flexibility if it's unclear that such a need will ever arise.

waku/v2/protocol/waku_rln_relay/waku_rln_relay_metrics.nim
rymnc: Addressed in [6915cd0](https://github.com/status-im/nwaku/pull/1262/commits/6915cd09ba7e3ce55c1b6804a97ddedcc2c95648)

waku/v2/node/wakunode2_setup_metrics.nim
rymnc: Addressed in [d18a025](https://github.com/status-im/nwaku/pull/1262/commits/d18a0251a1fadb17d51671f8302a3316989f503a)

jm-clius (approved):
LGTM! Thanks :)

Issue #1157, opened 2022-09-15
bug: ERROR_INVALID_CURSOR


fryorcraken:
### Problem

Getting an `Invalid cursor` error from prod fleet after using a cursor returned in a previous reply.

### Impact

Unsure as messages are being returned as expected at first.

### To reproduce

1. Save https://gist.github.com/fryorcraken/0dfe1c83b74b2373ca8236f640fa4bfc locally
2. Open it with a browser (Firefox or Chrom*, doesn't matter(
3. be sure to have `debug=waku*` in local storage
4. Refresh, check the logs in console.

### Expected behavior
No `invalid cursor` error when the cursor used was previously returned by nwaku.

### Screenshots/logs
console logs:

```
MetaMask: Content script lost connection to "MetaMask Background Multiplex". undefined
contentscript.js:1 MetaMask: Muxed traffic for channel "metamask-provider" failed. undefined
contentscript.js:1 MetaMask: Content script lost connection to "MetaMask Background Legacy Multiplex". undefined
contentscript.js:1 MetaMask: Muxed traffic between channels "provider" and "metamask-provider" failed. undefined
contentscript.js:1 MetaMask: Muxed traffic for channel "publicConfig" failed. undefined
index-a538767c.js:287 waku:peer-discovery-static-list Use provided list of peers (reduced to maxPeers) +0ms ['[object Object]']
index-a538767c.js:287 waku:waku Waku node created +0ms 12D3KooWHc8NPEq7bARMYceqfuzaKLpk9BCiwHdddnctMWt9rhFP relay: true, store: true, light push: true, filter: true
index-a538767c.js:287 waku:peer-discovery-static-list Starting to emit static peers. +91ms
index-a538767c.js:287 waku:wait-for-remote-peer Resolving for +0ms /vac/waku/store/2.0.0-beta4 (6) ['/ipfs/id/1.0.0', '/ipfs/ping/1.0.0', '/vac/waku/filter/2.0.0-beta1', '/vac/waku/lightpush/2.0.0-beta1', '/vac/waku/relay/2.0.0', '/vac/waku/store/2.0.0-beta4']
index-a538767c.js:287 waku:store Querying history with the following options +0ms {peerId: undefined, pageDirection: 'backward', pageSize: 1, timeFilter: {…}}
index-a538767c.js:287 waku:select-peer Using codec /vac/waku/store/2.0.0-beta4 +0ms
[...]

PR #1263, opened 2022-10-13
fix: store cursors should be exclusive and match a DB item

jm-clius:
This fixes two related bugs in nwaku v0.12.0 with regards to cursors in the store protocol:

### Summary of unexpected behaviour:
- a history response returns a `cursor` pointing to the beginning of the _next_ page, if such a page exists
- the message pointed to by this `cursor` is excluded from a subsequent history response as per RFC [13](https://rfc.vac.dev/spec/13/) and is therefore excluded from the retrievable history

### Fix

This fix:
- reverts to using the index of the last message _within_ the returned response as the cursor (as it was pre `v0.12.0` and is described in the RFC)
- continues with the post `v0.12.0` behaviour of _only returning a cursor when a next page is available_. This is how most clients are currently implemented and is necessary to fix a fairly urgent bug.

### Comments:

- an alternative solution would be to continue using the index of the _next_ message as cursor to retrieve the next page and simply change the cursor logic (for all stores) to be inclusive of the cursor rather than exclusive. I prefer not doing this, as it violates the RFC (and requires the store client to make use of the index of a message it has not yet received)
- after some deliberation, I'm not sure that I agree with the logic of "no cursor returned if no next page available". Why wouldn't we? A cursor is a pointer to a place in history, not some contractual agreement that more history exists. This is up to the client application to determine. If someone, for example, is querying forward in history they may reach "real time", with no newer messages available in history _yet_. However, they should still receive a cursor from which they can continue their forward query at some point when they believe new messages may have arrived. This logic is also the only way the RFC can really be interpreted currently. Since several clients rely on this "empty pagination/cursor info" condition to stop querying, we can temporarily continue implementing the cursor this way, but in future we either need to amend the RFC or change the clients' behaviour.

### Summary of unexpected behaviour

- when reading the message digest (id) from the DB, the wrong number of bytes were read, resulting in `HistoryResponse`s containing the wrong cursor
- this implies that subsequent queries would use a cursor with a digest component that doesn't match any message in the DB. The impact here was somewhat mitigated in that a cursor was never directly matched, but only compared to other cursors (with the `storedAt` component playing an important role).

### Fix

- use correct column when determining the byte length of DB columns

## Tests

Since these bugs were missed in tests, I've modified the unit tests to verify that we can page through an entire history in both directions.
This:
- ensures that the message pointed to by a `pagingInfo` cursor are excluded as expected from subsequent queries
- ensures that an exact match does exist for the cursor (otherwise one of the paging directions would wrongly _reinclude_ the message supposedly pointed to by the non-matching cursor)


jm-clius:
Note that this bug has the side effect of allowing a workaround for the issue reported in https://github.com/status-im/nwaku/issues/1157. We therefore first need to fix that issue, before merging this.

rymnc (approved):
LGTM!

LNSD (approved):
LGTM ✅ 

It took me some time to review it. Thanks for the fixes :)

LNSD:
> after some deliberation, I'm not sure that I agree with the logic of "no cursor returned if no next page available". Why wouldn't we? A cursor is a pointer to a place in history, not some contractual agreement that more history exists. This is up to the client application to determine. If someone, for example, is querying forward in history they may reach "real time", with no newer messages available in history yet. However, they should still receive a cursor from which they can continue their forward query at some point when they believe new messages may have arrived. This logic is also the only way the RFC can really be interpreted currently. Since several clients rely on this "empty pagination/cursor info" condition to stop querying, we can temporarily continue implementing the cursor this way, but in future we either need to amend the RFC or change the clients' behaviour.

As I see it, this is a tradeoff decision. With current behavior, we avoid doing an extra query to see that the next page is empty; and I think that all implementations should be able to build a cursor based on the last message they received. 

The rationale behind this decision was: Having the clients to query once more to stop the queries loop is terrible for the waku store nodes as it implies a new request to process and a new DB query with the corresponding CPU usage.

alrevuelta (commented):
lgtm!

as a nitpick/question, any reason behind using a non deterministic timestamp `getNanosecondTime(getTime().toUnixFloat())` instead of hardcoded value?

In this case it shouldn't matter but tests should be reproduzable/deterministic/idempotent/etc.

jm-clius:
> In this case it shouldn't matter but tests should be reproduzable/deterministic/idempotent/etc.

I agree! In this case using real time was necessary, because the underlying store [message validity check](https://github.com/status-im/nwaku/blob/44e49a023587257a9d275dc653689f825d05bbf6/waku/v2/protocol/waku_store/protocol.nim#L244-L253) uses real time to determine if a message will be inserted or not. The test would therefore fail if it uses timestamps that do not fall within 20 seconds of current time. Out of scope for this PR, but the way to fix it would be to change the underlying validity check to take a reference timestamp as parameter (which defaults to current time).

PR #1267, opened 2022-10-14
[WIP] feat(lightpush): Dandelion support

kaiserd:
This PR adds basic Waku Dandelion stem support to lightpush,
and addresses part of the [Dandelion++ for Waku v2 Milestone](https://github.com/vacp2p/research/issues/119).

Currently, this PR is WIP and mainly serves as an example for the discussion in: https://github.com/vacp2p/rfc/pull/541

This PR does not feature the fail-safe.
I'd suggest adding the fail-safe in a future PR.

## Open Tasks

* [ ] randomly select peers from the mesh peer set
  - left this open because of the planned peer store restructuring
* [ ] select two peers per epoch, and implement the connection to stem-relay mapping 
* [ ] more analysis
  - e.g. lightpush replies with confirmations. Do we want this for the stem phase?

### DIFF
--- a/waku/v2/node/config.nim
+++ b/waku/v2/node/config.nim
@@ -251,6 +251,11 @@ type
       desc: "Peer multiaddr to request lightpush of published messages.",
       defaultValue: ""
       name: "lightpushnode" }: string
+
+    dandelion* {.
+      desc: "Enable dandelion stem relaying: true|false",
+      defaultValue: false
+      name: "dandelion" }: bool
     
     ## JSON-RPC config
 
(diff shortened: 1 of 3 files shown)


LNSD:
@kaiserd I don't like mixing the dandelion++ code with the _lightpush_ protocol code. 

Is there any possibility that we could do some kind of "composition" with dandelion and _lightpush_ protocol? Maybe something like a dependency that gets injected in the _lightpush_ protocol?

Note that the `protocol` module will be split into client and server to respect the single responsibility principle.

LNSD:
Please, check PR #1295

LNSD:
I think that this PR won't be merged as it is. Can we close it? We can always open a new one based on this one, once when we resume the lightpush dandelion implementation.

PR #1273, opened 2022-10-18
refactor(wakunode2): move config random private key generation to a procedure

LNSD:
First gardening PR 🌳 of the _split wakunode2.nim and move wakunode2 main to apps folder_ series.

- [x] Generate config default values within a procedure
- [x] Remove unused imports

### DIFF
--- a/waku/v2/node/config.nim
+++ b/waku/v2/node/config.nim
@@ -1,14 +1,15 @@
 import
   std/strutils,
-  confutils, confutils/defs, confutils/std/net,
+  confutils, 
+  confutils/defs, 
+  confutils/std/net,
   confutils/toml/defs as confTomlDefs,
   confutils/toml/std/net as confTomlNet,
-  chronicles, chronos,
+  chronicles, 
+  chronos,
   libp2p/crypto/crypto,
   libp2p/crypto/secp,
-  nimcrypto/utils,
-  eth/keys,
-  ../protocol/waku_message
+  nimcrypto/utils
 
 export
   confTomlDefs,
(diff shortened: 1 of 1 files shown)


waku/v2/node/config.nim
```diff
@@ -458,11 +459,13 @@ proc parseCmdArg*(T: type Port, p: TaintedString): T =
 proc completeCmdArg*(T: type Port, val: TaintedString): seq[string] =
   return @[]
 
-func defaultListenAddress*(conf: WakuNodeConf): ValidIpAddress =
```
rymnc, line 461: Can we continue to use `func` here? 
context: https://status-im.github.io/nim-style-guide/language.proc.html

waku/v2/node/config.nim
```diff
@@ -458,11 +459,13 @@ proc parseCmdArg*(T: type Port, p: TaintedString): T =
...
+proc defaultListenAddress*(): ValidIpAddress =
   # TODO: How should we select between IPv4 and IPv6
   # Maybe there should be a config option for this.
   (static ValidIpAddress.init("0.0.0.0"))
 
+proc defaultPrivateKey*(): PrivateKey =
```
rymnc, line 467:
```suggestion
func defaultPrivateKey*(): PrivateKey =
```
Not sure if the generation of the key has side effects!

rymnc (approved):
LGTM, thanks!

waku/v2/node/config.nim
LNSD: Yes, it has side effects. That's why I changed all to `proc`: looking for consistency.

jm-clius (approved):
👍

PR #1274, opened 2022-10-18
refactor(node): split wakunode into waku_node and wakunode2

LNSD:
Another gardening PR 🌳 part of the split wakunode2.nim and move wakunode2 main to apps folder series.

- [x] Split waku node types from the `wakunode2` app module
- [x] Update imports accordingly


waku/v2/node/wakunode2.nim
```diff
@@ -827,6 +52,10 @@ when isMainModule:
     ./storage/message/dual_message_store,
```
jm-clius, line 52: Curious if there's still a reason to keep this second group of imports are still necessary now that this module has only a single purpose?

LNSD: No, there is no technical reason behind it. I wanted this PR to keep this PR "as simple as possible". If I had moved them, it would have been harder to follow the changes.

I am planning to improve this file code, too. These imports will be merged in one of the subsequent PRs 😁

LNSD:
Merged this PR due to the high risk of having to rewok things (waku_node is one core module)

PR #1275, opened 2022-10-18
feat(rln-relay): add metrics dashboard

rymnc:
Adds a dashboard of rln-relay with variables related to the test fleets. A single node dashboard is added as well.

Screenshots - 
image
image
image


jm-clius (approved):
Thanks!

PR #1276, opened 2022-10-18
chore(discv5): bump nim-eth version

kaiserd:
This PR bumps nim-eth, which contains discv5, to the latest version.

(We are using a nim-eth feature branch that implements [33/WAKU2-DISCV5](https://rfc.vac.dev/spec/33/). I just rebased [this nim-eth feature branch](https://github.com/status-im/nim-eth/tree/selectable-protocol-id) onto the current master.)


jm-clius (approved):

jm-clius:
Approved, but I notice that there are multiple CI failures - either some new incompatibility, or because we have not updated all other submodules in tandem. We usually update all submodules together to avoid any interdependency mismatches :)

PR #1279, opened 2022-10-19
feat(chat2): metrics server + metrics logging

rymnc:
Should resolve https://github.com/status-im/nwaku/issues/1203

This change includes the following -
- starting the metrics server in chat2
- enabling metrics logging in chat2
- some movement of code so that the metrics logging can be accessed by both apps, wakunode2 and chat2 (cc: @LNSD)

image


apps/wakunode2/wakunode2_setup_metrics.nim
```diff
@@ -1,78 +1,8 @@
...
-        logRlnMetrics()
-
-    discard setTimer(Moment.fromNow(30.seconds), logMetrics)
-  
-  discard setTimer(Moment.fromNow(30.seconds), logMetrics)
-
-  
+export
+  startMetricsServer,
+  startMetricsLog
```
jm-clius, line 8: Seems like the function of this module is now simply to import the single util and export all its symbols? I agree that the metrics logging may work bettter as a util, but then we can probably just remove this module.

waku/v2/utils/metrics_logging.nim
```diff
@@ -0,0 +1,79 @@
...
+  ../node/peer_manager/peer_manager,
+  ../node/waku_node
+
+when defined(rln) or defined(rlnzerokit):
+  import ../protocol/waku_rln_relay/waku_rln_relay_metrics
+
+const LogInterval = 30.seconds
+
+logScope:
+  topics = "wakunode.setup.metrics"
```
jm-clius, line 26: Now rather something like `waku.utils.metrics`

apps/wakunode2/wakunode2_setup_metrics.nim
rymnc: I agree, wanted the thoughts of someone from the production team re: this.

rymnc: Addressed in [52852fb](https://github.com/status-im/nwaku/pull/1279/commits/52852fb3cf863a0b0cb58f2102bb0178c489e357)

waku/v2/utils/metrics_logging.nim
rymnc: Addressed in [9f2d9a9](https://github.com/status-im/nwaku/pull/1279/commits/9f2d9a92de06f2f740f3c4722b5a6a797a89a089)

apps/chat2/config_chat2.nim
```diff
@@ -180,7 +180,7 @@ type
 
     metricsLogging* {.
       desc: "Enable metrics logging: true|false"
-      defaultValue: false
```
LNSD, line 183: I also disagree that the console metrics logging task should be enabled by default.

Chat2 is a "toy app"/example, and we use it to do some sanity tests. I am okay with adding metrics support but not enabling it by default.

LNSD (changes requested):
I don't think that metrics are a "utility". They are an entity by themselves. We should avoid importing anything from `node` or `protocols` in `utils` directory.

This is a partial list of rules to avoid problematic import cycles:

* `utils` MUST NOT import any module from `protocol` or `node`.
* `protocol` CAN import modules from `utils`.
* `protocol` MUST NOT import modules from `node`.
* `node` CAN import modules from `utils` and `protocol`.
* `apps` modules (e.g. `config.nim`) MUST NOT be imported by any module.

If you want to _generalize_ the metrics logic, then please add the as a module under `waku/v2/node/metrics`. They are tightly related to waku `node` code (`chat2` should be seen as a wrapper around `WakuNode`).

jm-clius:
@LNSD is right and I should have caught this in my review. A `utils` class is more specific than the node modules and should therefore not import node logic. I've added @LNSD's partial list of rules to the [coding guide for nwaku](https://hackmd.io/@vac/main/%2F5oIVOuQcSM2WKrnDLUPcpg).

apps/chat2/config_chat2.nim
rymnc: Imo, we should enable it by default, because the next testnet would involve users/operators looking at proof verification times etc, and/or sending the metrics data to a central server for researchers to analyze. wdyt?

rymnc:
> If you want to _generalize_ the metrics logic, then please add the as a module under `waku/v2/node/metrics`. They are tightly related to waku `node` code (`chat2` should be seen as a wrapper around `WakuNode`).

Addressed in [ab4c99f](https://github.com/status-im/nwaku/pull/1279/commits/ab4c99f39dfc192f240902e9f461317e6d4f43d5)

LNSD (approved):
LGTM ✅

waku/v2/node/waku_metrics.nim
```diff
@@ -8,22 +8,22 @@ import
...
 
 when defined(rln) or defined(rlnzerokit):
-  import ../../waku/v2/protocol/waku_rln_relay/waku_rln_relay_metrics
+  import ../protocol/waku_rln_relay/waku_rln_relay_metrics
 
+const LogInterval = 30.seconds
 
 logScope:
-  topics = "wakunode.setup.metrics"
+  topics = "waku.metrics"
```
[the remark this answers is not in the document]
LNSD: To comply with the Waku Code conventions compiled by @jm-clius:

See: https://hackmd.io/5oIVOuQcSM2WKrnDLUPcpg#Logging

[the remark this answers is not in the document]
rymnc: https://github.com/status-im/nwaku/pull/1279#discussion_r999201949 and https://hackmd.io/@vac/main/%2F5oIVOuQcSM2WKrnDLUPcpg#Logging

apps/chat2/config_chat2.nim
```diff
@@ -165,7 +165,7 @@ type
 
     metricsServer* {.
       desc: "Enable the metrics server: true|false"
-      defaultValue: false
+      defaultValue: true
```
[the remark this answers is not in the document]
rymnc: Addressed in [faccda0](https://github.com/status-im/nwaku/pull/1279/commits/faccda03cede24cf0540d9e34b93cb3f92b9497d)

waku/v2/utils/collector.nim
```diff
@@ -1,19 +1,29 @@
...
-  freshCount
\ No newline at end of file
+  {.gcsafe.}:
+    let total = parseCollectorIntoF64(collector)
+    let freshCount = total - cumulativeValue
+    cumulativeValue = total
+    freshCount
+
+template collectorAsF64*(collector: Collector): float64 =
+  ## This template is used to get metrics from 0
```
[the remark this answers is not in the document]
rymnc: Addressed in [208b265](https://github.com/status-im/nwaku/pull/1279/commits/208b26533683d4005f6ebe37dd0ffde7c942c2fa)

staheri14 (approved):
LGTM!

PR #1280, opened 2022-10-19
chore(rln-relay): removes the Goerli account address config option from the onchain-rln-relay tutorial

staheri14:
The Goerli account address is an optional flag, and this PR updates the onchain-rln-relay tutorial by removing the Goerli account address from the list of necessary config options.


rymnc:
Merging since the PR has had 2 approvals for a while now.

PR #1281, opened 2022-10-20
chore(store): disable temporarily the store resume flaky test case

LNSD:
As the test has been failing for a month, causing many inconveniences, no clear root cause has been identified so far, and the resume store is a second-priority functionality; this PR disables it temporarily, and the following issue has been created to track the investigation/fix:

- #1282


jm-clius (approved):
Thanks!

PR #1283, opened 2022-10-20
fix(store): cursor retrieval from sqlite DB using incorrect col number

LNSD:
This PR aims to merge the SQLite DB query fixes of PR #1263:

> ## Bug 2: cursor retrieval from sqlite DB using incorrect length
> 
> ### Summary of unexpected behaviour
> 
> - when reading the message digest (id) from the DB, the wrong number of bytes were read, resulting in `HistoryResponse`s containing the wrong cursor
> - this implies that subsequent queries would use a cursor with a digest component that doesn't match any message in the DB. The impact here was somewhat mitigated in that a cursor was never directly matched, but only compared to other cursors (with the `storedAt` component playing an important role).
> 
> ### Fix
> 
> - use correct column when determining the byte length of DB columns


LNSD:
PR #1263 was merged. Closing this PR

Issue #1278, opened 2022-10-18
feat(rln): How to securely store RLN credentials cross-client


s1fr0:
### Background
In 
- https://github.com/status-im/nwaku/issues/1021
- https://github.com/status-im/nwaku/issues/1238
- https://github.com/vacp2p/rfc/issues/543

is raised multiple times the question of standardizing the RLN credential format and securely persist it.

### Details
`nwaku`, `js-waku` and `go-waku` generates RLN credentials through `zerokit` FFI/WASM APIs.

While standardization of a format for RLN credentials can be pursued in https://github.com/vacp2p/rfc/issues/543 (a tentative solution was proposed [here](https://github.com/status-im/nwaku/issues/1238#issuecomment-1278114079)), their secure persistence and cross-client usage (e.g. decryption) is non-trivial.

As suggested [here](https://github.com/status-im/nwaku/issues/1021#issuecomment-1222081649), `nwaku` comes with a `nimbus` keystore implementation that allows to securely encrypt messages (in fact, *keys*) using a password/mnemonic sentence. However, to allow cross-client usage of encrypted credentials, the underlying employed primitives (`scrypt, pbkdf, sha256, aes-128-ctr`) should be ported/implemented in `js-waku` and `go-waku` too. 

On top of this, nimbus keystores don't seem to allow enough flexibility on fields structure to reach something similar to what  proposed [here](https://github.com/status-im/nwaku/issues/1238#issuecomment-1278114079) and come with primitives that should probably be replaced at least to provide 256 bits of security (I would personally use Argon2 + ChaChaPoly).

Given this, a possible solution could be to implement a WASM-friendly rust module that we can interface, similarly as zerokit, with FFI/WASM implementing all the logic on our custom standardized RLN credential format (most probably JSON with fields content encrypted), so that we avoid multiple implementations of the same crypto primitives.

Nevertheless, if implementation/usage of a browser extension is envisioned for js-rln in order to safely store credentials (e.g., [crypt-keeper](https://github.com/Rate-Limiting-Nullifier/crypt-keeper)), then we might opt for separate implementations in `nwaku` and `go-waku` of RLN credentials' field encryption, since both already support Noise primitives. In any case, import/export of credentials should be supported too by the browser extension and crypt-keeper doesn't seem to have such feature yet.

### Pro and cons
[...]

Issue #999, opened 2022-06-10
Using sub-command to group rln-relay config options 


staheri14:
This issue is to track the following [comment](https://github.com/status-im/nwaku/pull/992#discussion_r893633559) that is to use [sub-commands](https://github.com/status-im/nim-confutils) to group rln-relay config options together.

Ivansete-status:
@rymnc - can this be closed?

rymnc:
Yes, we don't need this anymore.

PR #1285, opened 2022-10-20
feat: add keyfile support for RLN credentials secure storage

s1fr0:
This PR adds support to keyfile defined according to [Web3 Secret Storage Definition](https://github.com/ethereum/wiki/wiki/Web3-Secret-Storage-Definition) but extended in order to allow 
- secure storage of arbitrary-long input data rather than fixed-size ethereum private key;
- multiple keyfiles stored in same file, which could be encrypted with different passwords; 
- a retrieval mechanism to automatically decrypt the first keyfile encrypted with the input password among many;
- a skip mechanism to iterate among multiple keyfiles stored in same file and encrypted with same password;

The keyfile implementation is adapted from [nim-eth](https://github.com/status-im/nim-eth/blob/master/eth/keyfile) for this purpose. In order to allow flexibility/compatibility with `go-waku` implementation, the keyfile fields `id` and `version` are omitted by default, but can be re-enabled by changing their respective [flags](https://github.com/status-im/nwaku/blob/7c2c1f9ebe20234b175c850de65bf4cd315de219/waku/v2/utils/keyfile.nim#L103-L105).

Since the main purpose of keyfile introduction is to encrypt RLN credentials, this PR further implements their storage through keyfiles (re)definining the procs `writeRlnCredentials` and `readRlnCredentials`.  Given that keyfiles process RLN credentials as arbitrary byte data, the format of RLN credentials can be changed anytime with minor modification to the code. [Here](https://github.com/status-im/nwaku/issues/1278#issuecomment-1286250519) a related comment.

In order to set a password to encrypt/decrypt RLN credentials, it suffices to pass the `--rln-relay-cred-password=<PASSWORD>` flag to the chat2 or wakunode2 instances. When this flag is not passed, the encryption/decryption password is set to "" (empty string).

This PR addresses and solve with respect to `nwaku` and `go-waku` cross-clients credentials the issue https://github.com/status-im/nwaku/issues/1278

Note: keyfile.nim, by being a code adaptation coming from nim-eth doesn't seem to follow rigorously our nim guidelines, e.g. note use the `result` variable. Let me know if it is worth to align it to guidelines (would require some effort though).

### DIFF
--- a/docs/tutorial/onchain-rln-relay-chat2.md
+++ b/docs/tutorial/onchain-rln-relay-chat2.md
@@ -49,7 +49,7 @@ make chat2 RLN=true
 Run the following command to set up your chat2 client. 
 
 ```bash
-./build/chat2 --fleet:test --content-topic:/toy-chat/2/luzhou/proto --rln-relay:true --rln-relay-dynamic:true --rln-relay-eth-contract-address:0x4252105670fe33d2947e8ead304969849e64f2a6 --rln-relay-eth-account-address:xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx --rln-relay-eth-account-private-key:xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx --rln-relay-eth-client-address:xxxx --ports-shift=1 
+./build/chat2 --fleet:test --content-topic:/toy-chat/2/luzhou/proto --rln-relay:true --rln-relay-dynamic:true --rln-relay-eth-contract-address:0x4252105670fe33d2947e8ead304969849e64f2a6 --rln-relay-eth-account-private-key:xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx --rln-relay-eth-client-address:xxxx --ports-shift=1 
 ```
 
 In this command
@@ -60,14 +60,12 @@ In this command
 - the `--rln-relay-eth-contract-address` option gets the address of the membership contract.
  The current address of the contract is `0x4252105670fe33d2947e8ead304969849e64f2a6`.
  You may check the state of the contract on the [Goerli testnet](https://goerli.etherscan.io/address/0x4252105670fe33d2947e8ead304969849e64f2a6).
-- the `rln-relay-eth-account-address` option is for your account address on the Goerli testnet.
- It is a hex string of length 40 (not sensitive to the `0x` prefix). 
 - the `rln-relay-eth-account-private-key` option is for your account private key on the Goerli testnet. 
  It is made up of 64 hex characters (not sensitive to the `0x` prefix).
 - the `rln-relay-eth-client-address` is the WebSocket address of the hosted node on the Goerli testnet. 
  You need to replace the `xxxx` with the actual node's address.
 
-For the last three config options i.e., `rln-relay-eth-account-address`, `rln-relay-eth-account-private-key`, and `rln-relay-eth-client-address`, if you do not know how to obtain those, you may use the following tutorial on the [prerequisites of running on-chain spam-protected chat2](./pre-requisites-of-running-on-chain-spam-protected-chat2.md).
+For the last two config options i.e., `rln-relay-eth-account-private-key` and `rln-relay-eth-client-address`, if you do not know how to obtain those, you may use the following tutorial on the [prerequisites of running on-chain spam-protected chat2](./pre-requisites-of-running-on-chain-spam-protected-chat2.md).
 
 You may set up more than one chat client,
 just make sure that you increment the `--ports-shift` value for each new client you set up e.g., `--ports-shift=2`.
@@ -167,10 +165,10 @@ If this file does not already exist under the supplied path, then a new credenti
 Otherwise, the chat client does not generate a new credential and will use, instead, the persisted RLN credential.
 
 ```bash
-./build/chat2  --fleet:test --content-topic:/toy-chat/2/luzhou/proto --rln-relay:true --rln-relay-dynamic:true --rln-relay-eth-contract-address:0x4252105670fe33d2947e8ead304969849e64f2a6  --rln-relay-eth-account-address:your_eth_account --rln-relay-eth-account-private-key:your_eth_private_key  --rln-relay-eth-client-address:your_goerli_node  --ports-shift=1  --rln-relay-cred-path:./
+./build/chat2  --fleet:test --content-topic:/toy-chat/2/luzhou/proto --rln-relay:true --rln-relay-dynamic:true --rln-relay-eth-contract-address:0x4252105670fe33d2947e8ead304969849e64f2a6  --rln-relay-eth-account-private-key:your_eth_private_key  --rln-relay-eth-client-address:your_goerli_node  --ports-shift=1  --rln-relay-cred-path:./
 ```
 
-Note: If you are reusing credentials, you can omit the `rln-relay-eth-account-address` and `rln-relay-eth-account-private-key` flags
+Note: If you are reusing credentials, you can omit the `rln-relay-eth-account-private-key` flag.
 
 Therefore, the command to start chat2 would be -
 
@@ -180,7 +178,7 @@ Therefore, the command to start chat2 would be -
 
 # Sample test output
 In this section, a sample test of running two chat clients is provided.
-Note that the values used for `rln-relay-eth-account-address`, `rln-relay-eth-account-private-key`, and `rln-relay-eth-client-address` in the following code snippets are junk and not valid.
+Note that the values used for  `rln-relay-eth-account-private-key` and `rln-relay-eth-client-address` in the following code snippets are junk and not valid.
 
 The two chat clients namely `Alice` and `Bob` are connected to the test fleets.
 `Alice` sends 4 messages i.e., `message1`, `message2`, `message3`, and `message4`.
@@ -192,7 +190,7 @@ You can check this fact by looking at `Bob`'s console, where `message3` is missi
 
 **Alice**
 ```bash
-./build/chat2 --fleet:test --content-topic:/toy-chat/2/luzhou/proto --rln-relay:true --rln-relay-dynamic:true --rln-relay-eth-contract-address:0x4252105670fe33d2947e8ead304969849e64f2a6 --rln-relay-eth-account-address:0x1234567890123456789012345678901234567890 --rln-relay-eth-account-private-key:0x1234567890123456789012345678901234567890123456789012345678901234 --rln-relay-eth-client-address:wss://goerli.infura.io/ws/v3/12345678901234567890123456789012 --ports-shift=1 
+./build/chat2 --fleet:test --content-topic:/toy-chat/2/luzhou/proto --rln-relay:true --rln-relay-dynamic:true --rln-relay-eth-contract-address:0x4252105670fe33d2947e8ead304969849e64f2a6 --rln-relay-eth-account-private-key:0x1234567890123456789012345678901234567890123456789012345678901234 --rln-relay-eth-client-address:wss://goerli.infura.io/ws/v3/12345678901234567890123456789012 --ports-shift=1 
 ```
 
 ```
@@ -236,7 +234,7 @@ your rln identity commitment key is: bd093cbf14fb933d53f596c33f98b3df83b7e9f7a19
 
 **Bob**
 ```bash
-./build/chat2 --fleet:test --content-topic:/toy-chat/2/luzhou/proto --rln-relay:true --rln-relay-dynamic:true --rln-relay-eth-contract-address:0x4252105670fe33d2947e8ead304969849e64f2a6 --rln-relay-eth-account-address:0x1234567890123456789012345678901234567890 --rln-relay-eth-account-private-key:0x1234567890123456789012345678901234567890123456789012345678901234 --rln-relay-eth-client-address:wss://goerli.infura.io/ws/v3/12345678901234567890123456789012 --ports-shift=2 
+./build/chat2 --fleet:test --content-topic:/toy-chat/2/luzhou/proto --rln-relay:true --rln-relay-dynamic:true --rln-relay-eth-contract-address:0x4252105670fe33d2947e8ead304969849e64f2a6 --rln-relay-eth-account-private-key:0x1234567890123456789012345678901234567890123456789012345678901234 --rln-relay-eth-client-address:wss://goerli.infura.io/ws/v3/12345678901234567890123456789012 --ports-shift=2 
 ```
 
 ```


waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
```diff
@@ -1145,19 +1157,31 @@ proc mountRlnRelayDynamic*(node: WakuNode,
   node.wakuRlnRelay = rlnPeer
   return ok(true)
 
-proc readPersistentRlnCredentials*(path: string) : RlnMembershipCredentials {.raises: [Defect, OSError, IOError, Exception].} =
-  info "Rln credentials exist in file"
+proc writeRlnCredentials*(path: string, credentials: RlnMembershipCredentials, password: string) : KfResult[void] {.raises: [Defect, OSError, IOError, Exception].} =
```
LNSD, line 1160: Here you are using two ways of passing errors:
- Result: `KfResult[void]`
- Exceptions `{.raises: [Defect, OSError, IOError, Exception].}`

We should use a single approach, which should be the `Result[T, R]` for consistency. And convert the exceptions to `err()` by wrapping the code with a `try/except` block.

I understand that this might be out of the scope of this PR, but this is something to consider when evolving the RLN module.

LNSD: Additionally, each module should return the `Result` type of the module. This function, which is exported and is part of the public API of the RLN relay module, returns the result type of the `utils/keyfile` module.

- Either you make this procedure non-public (remove the `*` next to the `proc` name)
- Or return the RLN relay module's `Result` type (e.g. `RlnResult`)

waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
```diff
@@ -1220,9 +1244,10 @@ proc mount(node: WakuNode,
       let rlnRelayCredPath = joinPath(conf.rlnRelayCredPath, RlnCredentialsFilename)
       debug "rln-relay credential path", rlnRelayCredPath
       # check if there is an rln-relay credential file in the supplied path
-      if fileExists(rlnRelayCredPath): 
+      if fileExists(rlnRelayCredPath):
+        info "A RLN credential file exists in provided path"
```
LNSD, line 1248: Can you provide the path as extra logged info?

```suggestion
        info "A RLN credential file exists in provided path", path=rlnRelayCredPath
```

waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
```diff
@@ -1145,19 +1157,31 @@ proc mountRlnRelayDynamic*(node: WakuNode,
...
-  return deserializedRlnCredentials
+    let entireRlnCredentialsFile = loadKeyFile(path, password)
+    if entireRlnCredentialsFile.isOk():
+      let jsonObject = parseJson(toString(entireRlnCredentialsFile.get()))
+      let deserializedRlnCredentials = to(jsonObject, RlnMembershipCredentials)   
+      debug "Deserialized RLN credentials", rlnCredentials=deserializedRlnCredentials
+      return some(deserializedRlnCredentials)
+    else:
+      debug "Unable to decrypt RLN credentials with provided password. ", error=entireRlnCredentialsFile.error
+      echo "Unable to decrypt RLN credentials with provided password. ", entireRlnCredentialsFile.error
```
LNSD, line 1183: Is this a leftover?

waku/v2/utils/keyfile.nim
```diff
@@ -0,0 +1,517 @@
...
+  if c.cipher.kind == CipherNoSupport:
+    return err(CipherNotSupported)
+  if len(c.cipher.text) == 0:
+    return err(EmptyCiphertext)
+  if len(c.mac) == 0:
+    return err(EmptyMac)
+  if isNil(c.kdfParams):
+    return err(MalformedError)
+
+  result = ok(c)
```
LNSD, line 400: We are avoiding using the `result` special variable. Either:

```suggestion
  ok(c)
```
Or:

```suggestion
  return ok(c)
```

waku/v2/utils/keyfile.nim
```diff
@@ -0,0 +1,517 @@
...
+
+  p.dklen = params.getOrDefault("dklen").getInt()
+  p.c = params.getOrDefault("c").getInt()
+  p.prf = getPrfHash(params.getOrDefault("prf").getStr())
+
+  if p.prf == HashNoSupport:
+    return err(PrfNotSupported)
+  if p.dklen == 0 or p.dklen > MaxDKLen:
+    return err(IncorrectDKLen)
+  result = ok(p)
```
LNSD, line 416: Idem:

```suggestion
  return ok(p)
```

waku/v2/utils/keyfile.nim
```diff
@@ -0,0 +1,517 @@
...
+
+  p.dklen = params.getOrDefault("dklen").getInt()
+  p.n = params.getOrDefault("n").getInt()
+  p.p = params.getOrDefault("p").getInt()
+  p.r = params.getOrDefault("r").getInt()
+
+  if p.dklen == 0 or p.dklen > MaxDKLen:
+    return err(IncorrectDKLen)
+
+  result = ok(p)
```
LNSD, line 432: Idem:

```suggestion
  return ok(p)
```

waku/v2/utils/keyfile.nim
```diff
@@ -0,0 +1,517 @@
...
+
+  case crypto.kind
+  of PBKDF2:
+    let res = decodePbkdf2Params(crypto.kdfParams)
+    if res.isErr:
+      return err(res.error)
+
+    let params = res.get()
+    let dkey = ? deriveKey(password, params.salt, PBKDF2, params.prf, params.c)
+    result = decryptSecret(crypto, dkey)
```
LNSD, line 464:
```suggestion
    return decryptSecret(crypto, dkey)
```

waku/v2/utils/keyfile.nim
```diff
@@ -0,0 +1,517 @@
...
+
+  of SCRYPT:
+    let res = decodeScryptParams(crypto.kdfParams)
+    if res.isErr:
+      return err(res.error)
+
+    let params = res.get()
+    let dkey = ? deriveKey(password, params.salt, params.n, params.r, params.p)
+    result = decryptSecret(crypto, dkey)
```
LNSD, line 473:
```suggestion
    return decryptSecret(crypto, dkey)
```

waku/v2/utils/keyfile.nim
```diff
@@ -0,0 +1,517 @@
...
+      result = newString(len(sarr))
+      copyMem(addr result[0], addr sarr[0], len(sarr))
+    except CatchableError:
+      result = ""
+  else:
+    result = ""
+
+proc compareMac(m1: openArray[byte], m2: openArray[byte]): bool =
+  if len(m1) == len(m2) and len(m1) > 0:
+    result = equalMem(unsafeAddr m1[0], unsafeAddr m2[0], len(m1))
```
LNSD, line 299:
```suggestion
    return equalMem(unsafeAddr m1[0], unsafeAddr m2[0], len(m1))
```

waku/v2/utils/keyfile.nim
```diff
@@ -0,0 +1,517 @@
...
+
+proc decodeSalt(m: string): string =
+  var sarr: seq[byte]
+  if len(m) > 0:
+    try:
+      sarr = utils.fromHex(m)
+      result = newString(len(sarr))
+      copyMem(addr result[0], addr sarr[0], len(sarr))
+    except CatchableError:
+      result = ""
```
LNSD, line 293:
```suggestion
      return ""
```

waku/v2/utils/keyfile.nim
```diff
@@ -0,0 +1,517 @@
...
+  else:
+    result = ""
```
LNSD, line 295:
```suggestion
    return ""
```

waku/v2/utils/keyfile.nim
```diff
@@ -0,0 +1,517 @@
...
+        "salt": salt
+      }
+    )
+  else:
+    err(NotImplemented)
+
+proc decodeHex(m: string): seq[byte] =
+  if len(m) > 0:
+    try:
+      result = utils.fromHex(m)
```
LNSD, line 279:
```suggestion
      return utils.fromHex(m)
```

waku/v2/utils/keyfile.nim
```diff
@@ -0,0 +1,517 @@
...
+    except CatchableError:
+      result = newSeq[byte]()
```
LNSD, line 281:
```suggestion
      return newSeq[byte]()
```

waku/v2/utils/keyfile.nim
```diff
@@ -0,0 +1,517 @@
...
+  else:
+    result = newSeq[byte]()
```
LNSD, line 283:
```suggestion
    return newSeq[byte]()
```

waku/v2/utils/keyfile.nim
```diff
@@ -0,0 +1,517 @@
...
+    return err(IncorrectDKLen)
+
+  result = ok(p)
+
+func decryptSecret(crypto: Crypto, dkey: DKey): KfResult[seq[byte]] =
+  var ctx: keccak256
+  ctx.init()
+  ctx.update(toOpenArray(dkey, 16, 31))
+  ctx.update(crypto.cipher.text)
+  var mac = ctx.finish()
```
LNSD, line 439: Is the `ctx.clear()` missing? 
```suggestion
  var mac = ctx.finish()
  ctx.clear()
```

waku/v2/utils/keyfile.nim
```diff
@@ -0,0 +1,517 @@
...
+      discard ctx.pbkdf2(password, salt, c, output)
+      ok(output)
+    of HashSHA3_384:
+      var ctx: HMAC[sha3_384]
+      discard ctx.pbkdf2(password, salt, c, output)
+      ok(output)
+    of HashSHA3_512:
+      var ctx: HMAC[sha3_512]
+      discard ctx.pbkdf2(password, salt, c, output)
+      ok(output)
```
LNSD, line 198: Are the `ctx.clear()` missing?

waku/v2/utils/keyfile.nim
```diff
@@ -0,0 +1,517 @@
...
+      result = SupportedHashesKinds[res]
+    else:
+      result = HashNoSupport
+
+proc getCipher(c: string): CryptKind =
+  var cl = c.toLowerAscii()
+  if cl == "aes-128-ctr":
+    result = AES128CTR
+  else:
+    result = CipherNoSupport
```
LNSD, line 140:
```suggestion
  if cl == "aes-128-ctr":
    return AES128CTR
  else:
    return CipherNoSupport
```

waku/v2/utils/keyfile.nim
```diff
@@ -0,0 +1,517 @@
...
+proc getPrfHash(prf: string): HashKind =
+  result = HashNoSupport
+  let p = prf.toLowerAscii()
+  if p.startsWith("hmac-"):
+    var hash = p[5..^1]
+    var res = SupportedHashes.find(hash)
+    if res >= 0:
+      result = SupportedHashesKinds[res]
+    else:
+      result = HashNoSupport
```
LNSD, line 133:
```suggestion
    if res >= 0:
      return SupportedHashesKinds[res]
    else:
      return HashNoSupport
```

waku/v2/utils/keyfile.nim
```diff
@@ -0,0 +1,517 @@
...
+      result = "scrypt"
+    else:
+      result = "pbkdf2"
+
+proc `$`(k: CryptKind): string =
+  case k
+    of AES128CTR:
+      result = "aes-128-ctr"
+    else:
+      result = "aes-128-ctr"
```
LNSD, line 122:
```suggestion
    of AES128CTR:
      return "aes-128-ctr"
    else:
      return "aes-128-ctr"
```

waku/v2/utils/keyfile.nim
```diff
@@ -0,0 +1,517 @@
...
+
+proc mapErrTo[T, E](r: Result[T, E], v: static KeyFileError): KfResult[T] =
+  r.mapErr(proc (e: E): KeyFileError = v)
+
+proc `$`(k: KdfKind): string =
+  case k
+    of SCRYPT:
+      result = "scrypt"
+    else:
+      result = "pbkdf2"
```
LNSD, line 115:
```suggestion
    of SCRYPT:
      return "scrypt"
    else:
      return "pbkdf2"
```

waku/v2/utils/keyfile.nim
```diff
@@ -0,0 +1,517 @@
...
+      result = newSeq[byte]()
+  else:
+    result = newSeq[byte]()
+
+proc decodeSalt(m: string): string =
+  var sarr: seq[byte]
+  if len(m) > 0:
+    try:
+      sarr = utils.fromHex(m)
+      result = newString(len(sarr))
```
LNSD, line 290:
```suggestion
      return newString(len(sarr))
```

waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
```diff
@@ -1145,19 +1157,31 @@ proc mountRlnRelayDynamic*(node: WakuNode,
...
+proc writeRlnCredentials*(path: string, credentials: RlnMembershipCredentials, password: string) : KfResult[void] {.raises: [Defect, OSError, IOError, Exception].} =
+  info "Storing RLN credentials"
+  var jsonString: string
+  jsonString.toUgly(%credentials)
+  let keyfile = createKeyFileJson(toSeqByte(jsonString), password)
+  if keyfile.isErr():
+    return err(keyfile.error)
+  return saveKeyFile(path, keyfile.get())
+
+proc readRlnCredentials*(path: string, password: string) : Option[RlnMembershipCredentials] {.raises: [Defect, OSError, IOError, Exception].} =
```
LNSD, line 1169: Same here, we should return a result type:
```nim
proc readRlnCredentials*(path: string, password: string): RlnResult[Option[RlnMembershipCredentials]] =
```

waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
```diff
@@ -1220,9 +1244,10 @@ proc mount(node: WakuNode,
       let rlnRelayCredPath = joinPath(conf.rlnRelayCredPath, RlnCredentialsFilename)
       debug "rln-relay credential path", rlnRelayCredPath
       # check if there is an rln-relay credential file in the supplied path
-      if fileExists(rlnRelayCredPath): 
+      if fileExists(rlnRelayCredPath):
+        info "A RLN credential file exists in provided path"
         # retrieve rln-relay credential
-        credentials = some(readPersistentRlnCredentials(rlnRelayCredPath))
+        credentials = readRlnCredentials(rlnRelayCredPath, conf.rlnRelayCredentialsPassword)
```
LNSD, line 1250: Here you are accessing a `conf` field. Using the `conf` object everywhere makes refactoring/decoupling work harder.

Can you specify a parameter in this function named `rlnRelayCredentialsPassword` and use it here?

waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
```diff
@@ -1256,8 +1281,8 @@ proc mount(node: WakuNode,
         # persist rln credential
         credentials = some(RlnMembershipCredentials(rlnIndex: node.wakuRlnRelay.membershipIndex, 
                                                     membershipKeyPair: node.wakuRlnRelay.membershipKeyPair))
-        writeFile(rlnRelayCredPath, pretty(%credentials.get()))
-
+        if writeRlnCredentials(rlnRelayCredPath, credentials.get(), conf.rlnRelayCredentialsPassword).isErr():
```
LNSD, line 1284: Same here:

> Can you specify a parameter in this function named `rlnRelayCredentialsPassword` and use it here?

apps/chat2/config_chat2.nim
```diff
@@ -290,6 +290,11 @@ type
       desc: "Address of membership contract on an Ethereum testnet", 
       defaultValue: ""
       name: "rln-relay-eth-contract-address" }: string
+
+    rlnRelayCredentialsPassword* {.
+      desc: "Password for encrypting RLN credentials", 
+      defaultValue: ""
+      name: "rln-relay-cred-password" }: string
```
LNSD, line 297: Can this configuration option be within a `when defined(rln):` compilation guard?

The rest of the RLN conf options should also be under the `when` compilation guard; if this is allowed by Nim.

apps/wakunode2/config.nim
```diff
@@ -174,7 +174,12 @@ type
       desc: "Address of membership contract on an Ethereum testnet", 
       defaultValue: ""
       name: "rln-relay-eth-contract-address" }: string
-    
+
+    rlnRelayCredentialsPassword* {.
+      desc: "Password for encrypting RLN credentials", 
+      defaultValue: ""
+      name: "rln-relay-cred-password" }: string
```
LNSD, line 181: Can this configuration option be within a `when defined(rln):` compilation guard?

The rest of the RLN conf options should also be under the `when` compilation guard; if this is allowed by Nim.

LNSD (changes requested):
Please, check my comments (and sorry for the nitpicking 🙏🏼)

tests/v2/test_keyfile.nim
```diff
@@ -0,0 +1,119 @@
...
+      keyfile.isOk()
+      saveKeyFile(filepath, keyfile.get()).isOk()
+
+    var decodedSecret = loadKeyFile("test.keyfile", password)
+
+    check:
+      decodedSecret.isOk()
+      secret == decodedSecret.get()
+
+    removeFile(filepath)
```
alrevuelta, line 34: As a nitpick, aren't we risking here that `removeFile` is not executed if an exception is thrown before? i.e. if `loadKeyFile` raises an exception, the file will never be removed.

I would use `defer` at the begining:

```
test "Create/Save/Load single keyfile":

    let password = "randompassword"
    let filepath = "./test.keyfile"

    defer: removeFile(filepath)

    ...
```

tests/v2/test_waku_rln_relay.nim
```diff
@@ -1003,16 +1003,21 @@ suite "Waku rln relay":
...
+    var credentials = readRlnCredentials(filepath, password)
 
     check:
-      credentials.membershipKeyPair == k
-      credentials.rlnIndex == index
+      credentials.isSome()
+      credentials.get().membershipKeyPair == k
+      credentials.get().rlnIndex == index
+
+    removeFile(filepath)
```
alrevuelta, line 1020: same, use of defer?

apps/chat2/config_chat2.nim
LNSD: I'm ok moving this to another PR. Not a blocker on this PR, but certainly, in the near future, we should put a `when` guard to RLN-specific configuration options (the same way we have it in the implementation).

waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
LNSD: The advantage is that you decouple the `mount()` procedure from the `WakuConf` type (char2 or wakunode2 conf type), which will ease the task of decoupling the RLN module from the chat2 and wakunode2 configuration types.

tests/v2/test_keyfile.nim
alrevuelta: I can see nimbus extensively uses it in testing for [closing files](https://github.com/status-im/nimbus-eth2/blob/v22.10.1/tests/test_keystore_management.nim#L42) and [closing databases](https://github.com/status-im/nimbus-eth2/blob/v22.10.1/tests/slashing_protection/test_slashing_protection_db.nim#L56) and in production code for [removing secrets](https://github.com/status-im/nimbus-eth2/blob/v22.10.1/beacon_chain/spec/keystore.nim#L804) and [closing files](https://github.com/status-im/nimbus-eth2/blob/v22.10.1/beacon_chain/validators/keystore_management.nim#L411). Would be a pitty if Nim removes it, but I don't see any immediate plans of doing it.

imho `try` `finally` adds an extra ugly indentation level and `defer` solves the same issue more elegantly. Nimbus using it gives me confidence, so its why I would prefer `defer`, but feel free to use `finally`, it also solves the problem I pointed out of the file not being removed :)

alrevuelta: ow nice reference! i actually didn't know it was translated to a try/finally.
resolving thanks!

apps/chat2/config_chat2.nim
staheri14: One more relevant issue about the visibility of config options is https://github.com/status-im/nwaku/issues/999 where it was suggested to use sub-commands to group all the rln-relay-related configs and make them available only when the rln-relay config option is set to true. Would that address your comment @LNSD ?

tests/v2/test_utils_keyfile.nim
```diff
@@ -0,0 +1,120 @@
...
+    defer: removeFile(filepath)
+
+    var secret = randomSeqByte(rng[], 300)
+    let keyfile = createKeyFileJson(secret, password)
+
+    check:
+      keyfile.isOk()
+      saveKeyFile(filepath, keyfile.get()).isOk()
+
+    var decodedSecret = loadKeyFile("test.keyfile", password)
```
staheri14, line 31:
```suggestion
    var decodedSecret = loadKeyFile(filepath, password)
```

tests/v2/test_utils_keyfile.nim
```diff
@@ -0,0 +1,120 @@
...
+
+  test "Create/Save/Load single keyfile":
+
+    let password = "randompassword"
+
+    let filepath = "./test.keyfile"
+    defer: removeFile(filepath)
+
+    var secret = randomSeqByte(rng[], 300)
+    let keyfile = createKeyFileJson(secret, password)
```
staheri14, line 25: A one-line comment on this part would be nice.

tests/v2/test_utils_keyfile.nim
```diff
@@ -0,0 +1,120 @@
...
+      saveKeyFile(filepath, keyfile.get()).isOk()
+
+    # We encrypt secret5 with password1
+    let secret5 = randomSeqByte(rng[], 300)
+    keyfile = createKeyFileJson(secret5, password1)
+    check:
+      keyfile.isOk()
+      saveKeyFile(filepath, keyfile.get()).isOk()
+
+    # We encrypt secret5 with password1
```
staheri14, line 79:
```suggestion
    # We encrypt secret6 with password1
```

waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
```diff
@@ -1145,19 +1146,39 @@ proc mountRlnRelayDynamic*(node: WakuNode,
...
+  var jsonString: string
+  jsonString.toUgly(%credentials)
+  let keyfile = createKeyFileJson(toBytes(jsonString), password)
+  if keyfile.isErr():
+    return err("Error while creating keyfile for RLN credentials")
+  if saveKeyFile(path, keyfile.get()).isErr():
+    return err("Error while saving keyfile for RLN credentials")
+  return ok()
+
+proc readRlnCredentials*(path: string, password: string) : RlnRelayResult[Option[RlnMembershipCredentials]] {.raises: [Defect, IOError, Exception].} =
```
staheri14, line 1160:
```suggestion
proc readRlnCredentials*(path: string, password: string): RlnRelayResult[Option[RlnMembershipCredentials]] {.raises: [Defect, IOError, Exception].} =
```

tests/v2/test_utils_keyfile.nim
```diff
@@ -0,0 +1,120 @@
...
+
+    var decodedSecret = loadKeyFile("test.keyfile", password)
+
+    check:
+      decodedSecret.isOk()
+      secret == decodedSecret.get()
+
+  test "Create/Save/Load multiple keyfiles in same file":
+
+    let password1 = "password1"
```
staheri14, line 39: May we also have a test for an empty password? it should work in general, but as it is a corner case, it may result in some abrupt behavior.

waku/v2/utils/keyfile.nim
```diff
@@ -0,0 +1,532 @@
+# This implementation is originally taken from nim-eth keyfile module https://github.com/status-im/nim-eth/blob/master/eth/keyfile and adapted to
```
staheri14, line 1: I guess there are some parts from the `keyfile` module that are imported here without any change, right? if yes, what do you think about importing such untouched parts directly from the `nim-eth/keyfile` module and just implementing the differences (which enable arbitrary data size encryption and other features) in the current module?  some immediate benefits are -it makes the review of new edits easier, and -it is less error-prone 
wdyt?

staheri14 (commented):
Good one! Please find my comments below.

waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
staheri14, line 1160: In this proc, we do not expose the `skip` feature, where the user can select which decrypted credential to use. 
Another useful utility is to return the number of available credentials under the supplied password. 
If it is easy to integrate them into this PR, I'd suggest adding them here, otherwise, in a follow-up PR.

staheri14, line 1160: In this proc, we do not expose the `skip` feature, where the user can select which decrypted credential to use. 
Another useful utility is to return the number of available credentials under the supplied password. 
If it is easy to integrate them into this PR, I'd suggest adding them here, otherwise, in a follow-up PR.

waku/v2/utils/keyfile.nim
```diff
@@ -0,0 +1,532 @@
+# This implementation is originally taken from nim-eth keyfile module https://github.com/status-im/nim-eth/blob/master/eth/keyfile and adapted to 
+# - create keyfiles for arbitrary-long input byte data (rather than fixed-size private keys)
+# - allow storage of multiple keyfiles (encrypted with different passwords) in same file and iteration among successful decryptions
+# - enable/disable at compilation time the keyfile id and version fields
+
+{.push raises: [Defect].}
```
staheri14, line 6: If we are going to have our own `keyfile` module as it is, then I'd add documentation for the constituent procs.

alrevuelta:
slightly related, just fyi related to storing keys in files [link](https://rauljordan.com/2020/11/01/custom-static-analysis-in-go-part-1.html). See "enforcing file permission best practices in our application section"

rymnc (approved):
LGTM after other reviewers' comments are addressed!

waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
jm-clius, line 1160: I don't think this `IOError` should be raised by the `proc` if it's already caught in the body. Also, do we know where the `Exception` comes from?

apps/chat2/config_chat2.nim
LNSD: Currently, we have the `RLN=true` feature flag (which translates into `-d:rln` and `-d:rlnzerokit`). But this compilation feature flag is not being applied in the configuration. 

If you compile with `RLN=false`, you still see the RLN configuration options in the help. That can be misleading because no RLN functionality was compiled.

So the solution would be to wrap in a compilation guard all the RLN options:

```Nim
      # [...]
      
      rlnRelayCredentialsPassword* {.
        desc: "Password for encrypting RLN credentials", 
        defaultValue: ""

      # [...]
```

---- 

On a side note, I don't understand how subcommands will help here. We are not using subcommands in the nwaku node app (aka `wakunode2`). And I don't see any use case for them in `wakunode2`, at least in the near future. 

We are working on grouping/namespacing the configuration options with a prefix (e.g., `--store-message-db-url`, `--store-message-retention-policy`, etc.). But this is already present in the current implementation of RLN 😁

LNSD: >  but nim seems a bit unhandy for these tasks.

Now I see why you said that. It is an issue in `nim-confutils` macros 😫 Then, let's leave it as it is until we have a better way of doing this.

LNSD (approved):
LGTM ✅

apps/chat2/config_chat2.nim
staheri14: @LNSD I see your point, that makes sense. 

> On a side note, I don't understand how subcommands will help here. We are not using subcommands in the nwaku node app (aka wakunode2). And I don't see any use case for them in wakunode2, at least in the near future.

It was suggested in one of the PRs and my understanding was that it is part of the plan to use subcommands: https://github.com/status-im/nwaku/pull/992#discussion_r893633559 @jm-clius

staheri14: Please let us know if the plan re subcommands has changed so that we deprioritize that issue i.e., #999 cc: @jm-clius @LNSD

staheri14: > I don't understand how subcommands will help here.

I'd say, then all the rln-relay-related config options will get encapsulated (and only accessible) under the rln-relay command/subcommand, hence the CLI interface becomes less confusing for the users (I assume this is the desirable feature we are looking for)

waku/v2/utils/keyfile.nim
```diff
@@ -0,0 +1,550 @@
...
+# Note that the keyfile is open in Append mode so that multiple credentials can be stored in same file
+proc saveKeyFile*(pathname: string,
+                  jobject: JsonNode): KfResult[void] =
+  ## Save JSON object ``jobject`` to file with pathname ``pathname``.
+  var
+    f: File
+  if not f.open(pathname, fmAppend):
+    return err(OsError)
+  try:
+    # To avoid other users/attackers to be able to read keyfiles, we make the make the file readable/writable only by the running user
```
staheri14, line 541:
```suggestion
    # To avoid other users/attackers to be able to read keyfiles, we make the file readable/writable only by the running user
```

tests/v2/test_utils_keyfile.nim
```diff
@@ -0,0 +1,135 @@
+{.used.}
+
+import
```
staheri14, line 3: Does this file cover all the unit tests included in https://github.com/status-im/nim-eth/blob/master/tests/keyfile/test_keyfile.nim ?

staheri14 (commented):
Thanks for addressing the comments. It LGTM, I left my last comment/question about the keyfile tests. If we have covered all the tests, then I have no further comments.

waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
jm-clius: Ah, no. Not like that 😄.

If an error is caught it is _not_ raised by the function anymore (unless explicitly re-raised). In other words, the `raises:` pragma is to list the exceptions that this function is "allowed" to still raise if called elsewhere. If someone calls `readRlnCredentials()` there should be no way that an `IOError` is raised if it's already caught and handled in the func body. If a specific error is not handled in the func body, it should be listed as specifically as possible in the `raises:` list, so that calling functions don't have to implement catch-all blocks (which may obscure serious and unrelated errors).

The changes in https://github.com/status-im/nwaku/commit/01ebc549f5000d3e0021fc63b2c815e1a8165af2 adds a catch-all, which would mean that any unrelated errors will quietly be swallowed.
In other words:
- where reasonable, catch and handle specific errors inside the func body and use the `Result` return type to indicate `error` or `ok` results.
- where it makes sense for an error to go uncaught in a body proc (so that it's raised in the calling function), list it explicitly in the `raises:` list.

tests/v2/test_utils_keyfile.nim
```diff
@@ -0,0 +1,379 @@
...
+    let password2 = ""
+    let password3 = "password3"
+    var keyfile: KfResult[JsonNode] 
+
+    let filepath = "./test.keyfile"
+    defer: removeFile(filepath)
+
+    # We generate 6 different secrets and we encrypt them using 3 different passwords, and we store the obtained keystore
+
+    let secret1 = randomSeqByte(rng[], 300)
```
alrevuelta, line 63: Since there is a clear pattern here, perhaps just use a for loop?

```nim
for pass in @[pass1, passn]:
  let secret = randomSeqByte(rng[], 300)
  let keyfile = createKeyFileJson(secret, pass)
  check:
      keyfile.isOk()
      saveKeyFile(filepath, keyfile.get()).isOk()
```

you can even randomize the password for each iteration, or even include %,Aa-? and weird stuff just in case?

tests/v2/test_utils_keyfile.nim
```diff
@@ -0,0 +1,379 @@
...
+    keyfile = createKeyFileJson(secret6, password1)
+    check:
+      keyfile.isOk()
+      saveKeyFile(filepath, keyfile.get()).isOk()
+
+    # Now there are 6 keyfiles stored in filepath encrypted with 3 different passwords
+    # We decrypt the keyfiles using the respective passwords and we check that the number of 
+    # successful decryptions corresponds to the number of secrets encrypted under that password
+
+    var decodedKeyfilesPassword1 = loadKeyFiles(filepath, password1)
```
alrevuelta, line 106: same, loop perhaps?

tests/v2/test_utils_keyfile.nim
```diff
@@ -0,0 +1,379 @@
...
+
+
+# The following tests are originally from the nim-eth keyfile tests module https://github.com/status-im/nim-eth/blob/master/tests/keyfile/test_keyfile.nim
+# and are slightly adapted to test our customized version of the utils/keyfile module
+# Note: the original nim-eth "Create/Save/Load test" is redefined and expanded above in "KeyFile test suite"
+suite "KeyFile test suite (adapted from nim-eth keyfile tests)":
+
+  let rng = newRng()
+
+  let TestVectors = [
```
alrevuelta, line 144: long test vector should perhaps go into a file? it muds the test imho

tests/v2/test_utils_keyfile.nim
```diff
@@ -0,0 +1,379 @@
...
+          "id": "3c8efdd6-d538-47ec-b241-36783d3418b9",
+          "version": 3
+      },
+      "password": "moomoocow",
+      "priv": "21eac69b9a52f466bfe9047f0f21c9caf3a5cdaadf84e2750a9b3265d450d481",
+      "name": "eth-keyfile-conftest"
+    }
+  ]
+
+  test "KeyStoreTests/basic_tests.json test1":
```
alrevuelta, line 292: not sure i understand this test, what are you testing?

i mean seckey.get() == expectkey but both values come from the same testvector?

tests/v2/test_utils_keyfile.nim
```diff
@@ -0,0 +1,379 @@
...
+  test "KeyStoreTests/basic_tests.json mycrypto":
+    var expectkey = decodeHex(TestVectors[4].getOrDefault("priv").getStr())
+    let seckey =
+      decodeKeyFileJson(TestVectors[4].getOrDefault("keyfile"),
+                        TestVectors[4].getOrDefault("password").getStr())
+    check:
+      seckey.isOk()
+      seckey.get() == expectkey
+
+  test "eth-key/conftest.py":
```
alrevuelta, line 344: perhaps being more descriptive with the test? something more meaningul and then a link to i.e. https://github.com/ethereum/eth-keys/blob/master/tests/backends/conftest.py? which i belive is the source?

waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
```diff
@@ -1145,19 +1146,45 @@ proc mountRlnRelayDynamic*(node: WakuNode,
...
+  let keyfile = createKeyFileJson(toBytes(jsonString), password)
+  if keyfile.isErr():
+    return err("Error while creating keyfile for RLN credentials")
+  if saveKeyFile(path, keyfile.get()).isErr():
+    return err("Error while saving keyfile for RLN credentials")
+  return ok()
+
+# Attempts decryptions of all keyfiles with the provided password. 
+# If one or more credentials are successfully decrypted, the max(min(index,number_decrypted),0)-th is returned.
+proc readRlnCredentials*(path: string, password: string, index: int = 0): RlnRelayResult[Option[RlnMembershipCredentials]] {.raises: [Defect, IOError, OSError].} =
```
alrevuelta, line 1162: shouldn't we avoid mixing both Result and exceptions?

Why `{.raises: [Defect, IOError, OSError].}`?

Since you have a `try` and `except` this proc should never rise an exception right?

alrevuelta (commented):
looking great, just some final questions/comments

waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
```diff
@@ -1145,19 +1146,45 @@ proc mountRlnRelayDynamic*(node: WakuNode,
...
+        # We should return the index-th decrypted credential, but we ensure to not overflow
+        let credentialIndex = max(min(index, decodedRlnCredentials.len - 1), 0)
+        debug "Picking credential with (adjusted) index", inputIndex=index, adjustedIndex=credentialIndex
+        let jsonObject = parseJson(string.fromBytes(decodedRlnCredentials[credentialIndex].get()))
+        let deserializedRlnCredentials = to(jsonObject, RlnMembershipCredentials)   
+        debug "Deserialized RLN credentials", rlnCredentials=deserializedRlnCredentials
+        return ok(some(deserializedRlnCredentials))
+      else:
+        debug "Unable to decrypt RLN credentials with provided password. ", error=decodedKeyfiles.error
+        return ok(none(RlnMembershipCredentials))
```
alrevuelta, line 1184: Shouldn't this be an error? And return error(xxx). wdyt?

On the other hand perhaps instead of `RlnRelayResult[Option[RlnMembershipCredentials]]` you can use `RlnRelayResult[RlnMembershipCredentials]`? As I understand if it returns ok() it should always reeturn RlnMembershipCredentials right?

tests/v2/test_utils_keyfile.nim
alrevuelta: I see. I'm asking because looks like you are testing `decodeKeyFileJson` function, and as far as I can see this is a nim-eth `proc`. can't we just assume that the function works as expected?

They already have their own tests, which I just realized are the same as these ones.
https://github.com/status-im/nim-eth/blob/2c236f64951d5c55095f8dc8af6328bb06f1df60/tests/keyfile/test_keyfile.nim#L170

or what is the reason behind copy pasting them here?

jm-clius (approved):
LGTM! Thanks for addressing the error handling. :)

tests/v2/test_utils_keyfile.nim
alrevuelta: my bad, indeed makes sense.

alrevuelta (approved):
lgtm! thank you for addressing the comments. I left [this](https://github.com/status-im/nwaku/pull/1285#discussion_r1006892737) one open as a nice to have but not blocking.

staheri14 (commented):
Looks good, I just left some minor comments.

staheri14 (approved):

PR #1286, opened 2022-10-20
refactor(wakunode2): extract setup proc definitions from main section

LNSD:
Another gardening PR 🌳 Preparing the ground for improving the configuration management.

- [x] Move `wakunode2` setup procs out of `isMainModule` section
- [x] Add try/catch clauses to the exception-raising procs and return a result
- [x] Turn setup methods using `waitFor` into `{.async.}`. 

**NOTE:** As a rule of thumb, `waitFor` should be used in the outermost scope (in this case, `main()`, in the tests the `asyncTest` already performs the `waitFor` for us.


alrevuelta (approved):
lgtm! the main module now is way easier to read!

PR #1297, opened 2022-10-25
fix: call runPeerExchangeDiscv5Loop

kaiserd:
This PR adds a call to `runPeerExchangeDiscv5Loop` in `wakunode2.nim`.
`runPeerExchangeDiscv5Loop` asynchronously fills the px peer list via discv5.
Thanks @richard-ramos for pointing this out :).


LNSD (approved):
LGTM ✅

jm-clius (approved):
Makes sense!

PR #1299, opened 2022-10-25
ci(jenkins): use MAKEFLAGS and NIMFLAGS pipeline params in build and test stages

LNSD:
The `NIMFLAGS` pipeline parameter was ignored previously in Jenkins' CI PR pipeline. I have added it back, together with the `MAKEFLAGS` parameter, to the _build_ and _test_ stages.


jm-clius (approved):

jakubgs (commented):
Thanks.

PR #1293, opened 2022-10-24
feat(store): simplify wakunode2 configuration options

LNSD:
The first part of the work simplifies the store configuration. Added new configuration options and deprecated the old ones.

Check for more details: #1103

The refactoring of the `wakunode2` setup sequence to support the new simplified store and persistence configuration options will be performed as part of the subsequent PR.

PR #1300, opened 2022-10-26
feat(wakunode2): simplify wakunode2 config and decouple peer persistence

### LINKED ISSUE #1103: store: simplify configuration options
## Background

Currently `store` configuration is quite confusing:
`--store` needs to be enabled, even if a node is only a store client
`--persist-messages` needs to be enabled to store historical messages as a store service node, even if messages are not persisted in DB
`--db-path` is required to actually persist messages on disk

See [this guide](https://github.com/status-im/nwaku/blob/ad7f9ed5bfe5c47cd581149a20def03d019b82a6/docs/operators/how-to/configure-store.md#configuring-a-store-client) for more.

## Suggested solution

It makes much more sense to me this way: `store` should in fact only be mounted if you are storing messages yourself and this implies:
- only the `--store` config option is necessary (the confusing `--persist-messages` can be removed <-- this can probably be done in any case)
- store query API methods should always be available, as it's unrelated to whether `store` is mounted or not.

This has been discussed before in the context of [advertising store protocol](https://github.com/status-im/go-waku/issues/292#issuecomment-1227474687) and [handling asymmetrical protocols in libp2p](https://github.com/status-im/nim-libp2p/issues/568)

LNSD:
This is the second part of #1293, and resolves #1103:

- [x] Decouple peer persistence database from message store database
- [x] Decouple `--storenode` from `resumeStoreNode`
- [x] Simplified `wakunode2` storage setup procedures
- [x] Made `SqliteDatabase` compatible with `dbUrl` path
- [x] Removed the deprecated configuration options from `wakunode2/config.nim`
- [x] Updated the tests accordingly

**NOTE:** Some extra work is necessary to improve the message store options (e.g., implement the unimplemented methods in the in-memory store, decouple the message store from the waku store protocol, etc.). This will be done in future PRs.


apps/wakunode2/wakunode2.nim
```diff
@@ -46,98 +49,157 @@ when defined(rln) or defined(rlnzerokit):
...
-    if dbRes.isErr():
-      warn "failed to init database connection", err = dbRes.error
-      waku_node_errors.inc(labelValues = ["init_db_failure"])
-      return err("failed to init database connection")
     else:
-      sqliteDatabase = dbRes.value
+      return err("unknown database engine")
 
+  if connRes.isErr():
+    return err("failed to init database connection:" & connRes.error)
```
rymnc, line 79:
```suggestion
    return err("failed to init database connection: " & connRes.error)
```

apps/wakunode2/wakunode2.nim
```diff
@@ -46,98 +49,157 @@ when defined(rln) or defined(rlnzerokit):
...
+      return err("failed to init sqlite message store: " & res.error)
+    
+    return ok(res.value)
+
+  else:
+    debug "setting up in-memory message store"
+    let store = StoreQueueRef.new(storeCapacity)
+    return ok(store)
+
+proc setupMessageStoreRetentionPolicy(retentionPolicy: string): SetupResult[Option[MessageRetentionPolicy]] =
```
rymnc, line 142: Should we use an optional enum for the retentionPolicy argument since we have only 2, or *no* retention policy at all?

apps/wakunode2/wakunode2.nim
```diff
@@ -46,98 +49,157 @@ when defined(rln) or defined(rlnzerokit):
...
+    # TODO: Run this only if the database engine is SQLite
+    let (pageSize, pageCount, freelistCount) = ?gatherSqlitePageStats(db.get())
+    debug "sqlite database page stats", pageSize=pageSize, pages=pageCount, freePages=freelistCount
+
+    if vacuum and (pageCount > 0 and freelistCount > 0):
+      ?performSqliteVacuum(db.get())
+
+  # Database migration
+  if db.isSome() and migrate:
+    ?performDbMigration(db.get(), migrationPath=MessageStoreMigrationPath)
```
rymnc, line 189:
```suggestion
    if migrate:
      ?performDbMigration(db.get(), migrationPath=MessageStoreMigrationPath)
```

apps/wakunode2/wakunode2.nim
```diff
@@ -431,27 +490,42 @@ proc startNode(node: WakuNode, conf: WakuNodeConf,
...
+    return err("failed to start the json-rpc server: " & getCurrentExceptionMsg())
+
+  ok()
   
+proc startRestServer(node: WakuNode, address: ValidIpAddress, port: uint16, portsShift: uint16, conf: WakuNodeConf): SetupResult[void] =
+  startRestServer(node, address, Port(port + portsShift), conf)
+  ok()
+
+proc startMetricsServer(node: WakuNode, address: ValidIpAddress, port: uint16, portsShift: uint16): SetupResult[void] =
+  startMetricsServer(address, Port(port + portsShift))
```
rymnc, line 523: a try-except here maybe?

rymnc (approved):
LGTM 😁

jm-clius (approved):
LGTM! LMK if you want to divide and conquer the subsequent tasks (infra updates, doc updates, etc.)

apps/wakunode2/wakunode2.nim
LNSD: Can you elaborate on that? Where are you suggesting using the enum?

- Return type?
- Input argument?
- Config?

apps/wakunode2/wakunode2.nim
LNSD: Well... the compiler is not complaining, so I don't think we should add a try/catch there.

But I agree that the error management in the metrics, JSON-RPC, and rest servers setup procedure has to be improved.

LNSD:
> LGTM! LMK if you want to divide and conquer the subsequent tasks (infra updates, doc updates, etc.)

@jakubgs In this PR, we are dropping the deprecated options. So in the next release, v0.13.0, the store configuration will change. Do you have any concerns before merging the PR?

jakubgs:
Yes, we'll need to coordinate the deployment to fleets or shit will break. I'm in Istanbul next week and off most of week after that, so if we want this done it would probably have to be tomorrow. Or on the 9th possibly.

apps/wakunode2/wakunode2.nim
rymnc: in the input, sorry. `retentionPolicy: string`

LNSD:
No need to rush. AFAIK, we are not planning to do any rollout for now. We can wait until you are back. 

Probably it is a better idea because I am planning to add support for configuring via environment variables. So we do not re-work things twice.

PR #1309, opened 2022-10-27
refactor(node): use store client in waku_node

LNSD:
The second part of the _spit store client_ PR:

- [x] Remove deprecated methods from store `protocol.nim` and `client.nim`
- [x] Move store's peer count metric to waku node
- [x] Waku node store API refactoring and simplification:
	* Now store client logic is decoupled from "server" logic
	* Waku swap is decoupled from waku store client
- [x] Updated wkunode2's `--storenode` logic to allow using client without mounting store protocol
- [x] Updated tests accordingly


LNSD:
> LGTM! We may have to think about peer management, but in principle I like specifying the queried peer directly.

Absolutely! For now, I am extracting the peer management capabilities outside the different protocol clients (to the waku node) as an intermediate movement. With the goal of simplifying the client code.

tests/v2/test_waku_swap.nim
```diff
@@ -117,26 +121,31 @@ procSuite "Waku SWAP Accounting":
     await server.mountStore(store=StoreQueueRef.new())
     await client.mountSwap(swapConfig)
     await client.mountStore()
+    client.mountStoreClient()
 
-    client.wakuStore.setPeer(server.peerInfo.toRemotePeerInfo())
     client.wakuSwap.setPeer(server.peerInfo.toRemotePeerInfo())
```
alrevuelta, line 126: Beyond the scope of the PR but should we also add `server.setSwapPeer` instead of `server.wakuSwap.setPeer`? Can see we already have `setLightPushPeer` and `setPeerExchangePeer`.

alrevuelta (approved):
lgtm

tests/v2/test_waku_swap.nim
LNSD: One protocol at a time. Check PR #1308 for lightpush protocol.

alrevuelta: great, just saw it!

LNSD:
Merging with a CI failure. The [offending test](https://github.com/status-im/nwaku/actions/runs/3346900385/jobs/5544272240#step:7:5667) case has already been reported to the RLN team.

PR #1315, opened 2022-10-28
feat(testlib): add enable/disable test debug utils

LNSD:
Following Javascript's Jest idea of _disabling a test by adding an 'x' before/after the test keyword_ (see [Jest docs](https://jestjs.io/docs/next/api#describeskipname-fn:~:text=Also%20under%20the%20aliases%3A%20it.skip(name%2C%20fn)%2C%20xit(name%2C%20fn)%2C%20and%20xtest(name%2C%20fn))):

> When you are maintaining a large codebase, you may sometimes find a test that is temporarily broken for some reason. If you want to skip running this test, but you don't want to delete this code, you can use test.skip to specify some tests to skip.

So in our case this is how to use the new test debug utils:

1. Import the `./testlib/testutils` module in your test suite file
2. Add a 'x' either before or after the test case/suite keyword: `xtest`, `xasyncTest`, `xsuite`, `xprocSuite`, etc.
3. Run the test suite. The tests you marked with an ✖️ won't be executed.

**IMPORTANT:** Do not forget to re-enable the test cases disabled during your debugging session before committing your changes.

### DIFF
--- a/tests/v2/testlib/testutils.nim
+++ b/tests/v2/testlib/testutils.nim
@@ -1,0 +1,12 @@
+template suitex*(name: string, body: untyped) = discard
+template xsuite*(name: string, body: untyped) = discard
+
+template testx*(name: string, body: untyped) = discard
+template xtest*(name: string, body: untyped) = discard
+
+template procSuitex*(name: string, body: untyped) = discard
+template xprocSuite*(name: string, body: untyped) = discard
+
+template asyncTestx*(name: string, body: untyped) = discard
+template xasyncTest*(name: string, body: untyped) = discard
+  


jm-clius (approved):
Yes, great feature, thanks! Shall we add this in a README.md somewhere? (And perhaps the coding guide?)

alrevuelta (approved):
Great! +1 to documenting it.

Fun fact I just learned that nim allows for comments inside comments, which C does't. Helpful for commenting the tests also.

```nim
#[


let x = 3
# comments
echo x
#[
long comment
]#


#[
```

Slightly related. Is someone looking into running specific tests or testsuites?

LNSD:
> Slightly related. Is someone looking into running specific tests or testsuites?

From our documentation: https://github.com/status-im/nwaku/blob/16c85db43ca492fb762e6acf61888e16812f883a/waku/v1/README.md#L51:~:text=You%20can%20also%20run%20a%20specific%20test%20(and%20alter%20compile%20options%20as%20you%20want)%3A`

> You can also run a specific test (and alter compile options as you want):

LNSD:
Merging this without waiting for checks to pass since the module is not imported, there is no risk.

alrevuelta:
@LNSD #1302?

jm-clius:
@alrevuelta 🤣 go ahead and merge #1302 by bypassing the checks.

PR #1318, opened 2022-10-31
fix(rln-relay): RangeDefect test

rymnc:
This should resolve the broken ci tests for the epoch comparison in rln-relay.

Compares the epochs and returns the appropriate value.

### DIFF
--- a/waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
+++ b/waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
@@ -785,8 +785,8 @@ proc getCurrentEpoch*(): Epoch =
   ## gets the current rln Epoch time
   return calcEpoch(epochTime())
 
-proc diff*(e1, e2: Epoch): uint64 =
-  ## returns the difference between the two rln `Epoch`s `e1` and `e2`
+proc absDiff*(e1, e2: Epoch): uint64 =
+  ## returns the absolute difference between the two rln `Epoch`s `e1` and `e2`
   ## i.e., e1 - e2
 
   # convert epochs to their corresponding unsigned numerical values
@@ -826,7 +826,7 @@ proc validateMessage*(rlnPeer: WakuRLNRelay, msg: WakuMessage,
   let
     msgEpoch = msg.proof.epoch
     # calculate the gaps
-    gap = diff(epoch, msgEpoch)
+    gap = absDiff(epoch, msgEpoch)
 
   debug "message epoch", msgEpoch = fromEpoch(msgEpoch)
 
--- a/waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
+++ b/waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
@@ -785,20 +785,20 @@ proc getCurrentEpoch*(): Epoch =
   ## gets the current rln Epoch time
   return calcEpoch(epochTime())
 
-proc safeIntCast(x: uint64): int64 =
-  if x < uint64(int64.high): return int64(x)
-  else: return -int64(bitnot(x)) - 1
-
-proc diff*(e1, e2: Epoch): int64 =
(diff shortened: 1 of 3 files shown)


waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
```diff
@@ -793,7 +793,10 @@ proc diff*(e1, e2: Epoch): int64 =
   let
     epoch1 = fromEpoch(e1)
     epoch2 = fromEpoch(e2)
-  return int64(epoch1) - int64(epoch2)
+  # Cannot use `abs` here because it is not available for `uint64`
+  if epoch1 > epoch2: return int64(epoch1 - epoch2)
+  else: return int64(epoch2 - epoch1) * -1
```
LNSD, line 798:
```suggestion
  if epoch1 > epoch2: 
    return int64(epoch1 - epoch2)
  else: 	
    return int64(epoch2 - epoch1) * -1
```

LNSD (approved):
LGTM ✅

LNSD:
> It doesn't really solve if epoch1 is uint64.high and epoch2 is uint64.low = 0: epoch1-epoch2 would have uint64 size and the cast to int64 (even if doesn't fail) will chop some bits. The only assumption that saves the test is that an epoch is 10s hence you reduce the epoch-width by few bits when converting from a timestamp. But the solution in general doesn't work (i.e. epoch = 1s)

Agree, it does not cover all the corner cases. See an example: https://play.nim-lang.org/#ix=4eAg

waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
rymnc: Does it apply to the `safeIntCast` proc in [730e352](https://github.com/status-im/nwaku/pull/1318/commits/730e352b07edd2cea9e856f595965482b025025c) as well?

rymnc:
> It doesn't really solve if epoch1 is uint64.high and epoch2 is uint64.low = 0: epoch1-epoch2 would have uint64 size and the cast to int64 (even if doesn't fail) will chop some bits. The only assumption that saves the test is that an epoch is 10s hence you reduce the epoch-width by few bits when converting from a timestamp. But the solution in general doesn't work (i.e. epoch = 1s)

This commit [730e352](https://github.com/status-im/nwaku/pull/1318/commits/730e352b07edd2cea9e856f595965482b025025c) ensures there is no overflow. wdyt?

rymnc:
The only line we use `diff` on is - https://github.com/status-im/nwaku/blob/77465653415b1ca9cf58e2d0f69f9f09e36ae8a5/waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim#L829

and subsequently, we use the absolute value of the int64 returned, here - https://github.com/status-im/nwaku/blob/master/waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim#L828

A simple fix would be just to return the uint64 after checking which epoch is bigger, which is in [7746565](https://github.com/status-im/nwaku/pull/1318/commits/77465653415b1ca9cf58e2d0f69f9f09e36ae8a5)

staheri14 (approved):
Thanks, LGTM!

waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
```diff
@@ -1,7 +1,7 @@
 {.push raises: [Defect].}
 
 import
-  std/[sequtils, tables, times, streams, os, deques],
+  std/[sequtils, tables, times, streams, os, deques, bitops],
```
staheri14, line 4: Where the `bitops` module is used?

rymnc: Addressed in [a42eb55](https://github.com/status-im/nwaku/pull/1318/commits/a42eb55c4a70df3d310471c58d810a4b262028cf), and removed `streams` as well.

PR #1319, opened 2022-10-31
feat(filter): split waku filter protocol into server and client

LNSD:
- [x] Split waku filter into client (subscriber) and server (publisher)
- [x] Remove deprecated methods from filter `protocol.nim` and `client.nim`
- [x] Move waku filter's peer count metric to the waku node module
- [x] Waku node filter API refactoring and simplification:
	* No more subscription management in waku node module
	* Simplified and renamed the expose APIs
- [x] Updated wkunode2's `--filternode` logic to allow using client without mounting filter protocol
- [x] Updated tests accordingly

### DIFF
--- a/apps/wakunode2/wakunode2.nim
+++ b/apps/wakunode2/wakunode2.nim
@@ -418,17 +418,18 @@ proc setupProtocols(node: WakuNode, conf: WakuNodeConf,
       return err("failed to set node waku lightpush peer: " & getCurrentExceptionMsg())
   
   # Filter setup. NOTE Must be mounted after relay
-  if (conf.filternode != "") or (conf.filter):
+  if conf.filter:
     try:
       await mountFilter(node, filterTimeout = chronos.seconds(conf.filterTimeout))
     except:
       return err("failed to mount waku filter protocol: " & getCurrentExceptionMsg())
 
-    if conf.filternode != "":
-      try:
-        setFilterPeer(node, conf.filternode)
-      except:
-        return err("failed to set node waku filter peer: " & getCurrentExceptionMsg())
+  if conf.filternode != "":
+    try:
+      await mountFilterClient(node)
+      setFilterPeer(node, conf.filternode)
+    except:
+      return err("failed to set node waku filter peer: " & getCurrentExceptionMsg())
   
   # waku peer exchange setup
   if (conf.peerExchangeNode != "") or (conf.peerExchange):
--- a/tests/v2/test_jsonrpc_waku.nim
+++ b/tests/v2/test_jsonrpc_waku.nim
@@ -5,7 +5,7 @@ import
   chronicles,
   testutils/unittests, stew/shims/net as stewNet,
   json_rpc/[rpcserver, rpcclient],
-  eth/[keys, rlp], eth/common/eth_types,
+  eth/keys, eth/common/eth_types,
   libp2p/[builders, switch, multiaddress],
   libp2p/protobuf/minprotobuf,
   libp2p/stream/[bufferstream, connection],
@@ -27,6 +27,7 @@ import
   ../../waku/v2/protocol/waku_store,
   ../../waku/v2/protocol/waku_swap/waku_swap,
   ../../waku/v2/protocol/waku_filter,
+  ../../waku/v2/protocol/waku_filter/client,
   ../../waku/v2/utils/peers,
   ../../waku/v2/utils/time,
   ./testlib/common
@@ -51,7 +52,7 @@ procSuite "Waku v2 JSON-RPC API":
 
     # RPC server setup
     let
-      rpcPort = Port(8545)
+      rpcPort = Port(8546)
       ta = initTAddress(bindIp, rpcPort)
       server = newRpcHttpServer([ta])
 
@@ -78,7 +79,7 @@ procSuite "Waku v2 JSON-RPC API":
 
     # RPC server setup
     let
-      rpcPort = Port(8545)
+      rpcPort = Port(8547)
       ta = initTAddress(bindIp, rpcPort)
       server = newRpcHttpServer([ta])
 
@@ -150,7 +151,7 @@ procSuite "Waku v2 JSON-RPC API":
 
     # RPC server setup
     let
-      rpcPort = Port(8545)
+      rpcPort = Port(8548)
       ta = initTAddress(bindIp, rpcPort)
       server = newRpcHttpServer([ta])
     
@@ -164,7 +165,7 @@ procSuite "Waku v2 JSON-RPC API":
     # First see if we can retrieve messages published on the default topic (node is already subscribed)
     await node2.publish(DefaultPubsubTopic, message1)
 
-    await sleepAsync(2000.millis)
+    await sleepAsync(100.millis)
 
     var messages = await client.get_waku_v2_relay_v1_messages(DefaultPubsubTopic)
 
@@ -182,7 +183,7 @@ procSuite "Waku v2 JSON-RPC API":
 
     var response = await client.post_waku_v2_relay_v1_subscriptions(@[pubSubTopic])
 
-    await sleepAsync(2000.millis)
+    await sleepAsync(100.millis)
 
     check:
       # Node is now subscribed to pubSubTopic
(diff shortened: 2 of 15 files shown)


tests/v2/test_jsonrpc_waku.nim
```diff
@@ -51,7 +52,7 @@ procSuite "Waku v2 JSON-RPC API":
 
     # RPC server setup
     let
-      rpcPort = Port(8545)
+      rpcPort = Port(8546)
```
rymnc, line 55: Is there a better way we can handle getting an open port?

rymnc (approved):
LGTM, just curious about port selection in tests.

tests/v2/test_waku_filter.nim
```diff
@@ -5,213 +5,191 @@ import
...
-      serverPeerManager = PeerManager.new(serverSwitch) 
-      serverProto = WakuFilter.init(serverPeerManager, rng, dummyHandler)
-    await serverProto.start()
-    serverSwitch.mount(serverProto)
-
-    # Client
-    let handlerFuture = newFuture[(string, MessagePush)]()
-    proc handler(requestId: string, push: MessagePush) {.async, gcsafe, closure.} =
-      handlerFuture.complete((requestId, push))
+    let
```
alrevuelta, line 52: In general, perhaps you could reuse more code across the tests? move some stuff to a common place? I see some boilerplate code that makes the tests more difficult to read and distracts from whats being tested.

waku/v2/node/waku_node.nim
```diff
@@ -409,80 +411,123 @@ proc mountRelay*(node: WakuNode,
...
-  waku_node_filters.set(node.filters.len.int64)
+  let subRes = await node.wakuFilterClient.subscribe(pubsubTopic, contentTopics, handlerWrapper, peer=remotePeer)
+  if subRes.isOk():
+    info "subscribed to topic", pubsubTopic=pubsubTopic, contentTopics=contentTopics
+  else:
+    error "failed filter subscription", error=subRes.error
+    waku_node_errors.inc(labelValues = ["subscribe_filter_failure"])
 
-proc unsubscribe*(node: WakuNode, request: FilterRequest) {.async, gcsafe.} =
+proc filterUnsubscribe*(node: WakuNode, pubsubTopic: PubsubTopic, contentTopics: ContentTopic|seq[ContentTopic],
```
alrevuelta, line 467: since `ContentTopic` is a subset of `seq[ContentTopic]` can we have just one type `contentTopic:seq[ContentTopic]` instead of `contentTopics: ContentTopic|seq[ContentTopic]` and if its a single one just use `@[topic]`?

waku/v2/protocol/waku_filter/client.nim
```diff
@@ -1,69 +1,207 @@
...
+          peerManager: PeerManager,
+          rng: ref rand.HmacDrbgContext): T = 
+          
+  let wf = WakuFilterClient(
+      peerManager: peerManager,
+      rng: rng,
+      subManager: SubscriptionManager.init()
+    )
+  wf.initProtocolHandler()
+  wf
```
alrevuelta, line 126: i would enforce using `return` in every function. thoughts?

alrevuelta (commented):
lgtm, just some minor comments.

waku/v2/node/waku_node.nim
LNSD: It is for convenience. The nim compiler compiles one version of the code or another depending on the argument type. Why should we remove it?

apps/wakunode2/wakunode2.nim
```diff
@@ -418,17 +418,18 @@ proc setupProtocols(node: WakuNode, conf: WakuNodeConf,
...
 
-    if conf.filternode != "":
-      try:
-        setFilterPeer(node, conf.filternode)
-      except:
-        return err("failed to set node waku filter peer: " & getCurrentExceptionMsg())
+  if conf.filternode != "":
+    try:
+      await mountFilterClient(node)
+      setFilterPeer(node, conf.filternode)
```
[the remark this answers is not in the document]
LNSD: This is like this for consistency with the rest of the code in this file. Feel free to change the code in this file

tests/v2/test_jsonrpc_waku.nim
LNSD: What do you mean? In this PR, I have just put random numbers per test to avoid chained failures.

I guess what you are referring to is to allow putting a `Port(0)` and getting a random TCP port. This is work that needs to be done. Feel free to jump on that task and do it :)

tests/v2/test_waku_filter.nim
LNSD: Can you be more specific? I don't understand what you are calling "boilerplate code"

waku/v2/protocol/waku_filter/client.nim
LNSD: What is the rationale behind the rule you are suggesting?

alrevuelta: 1) consistency and 2) readability.

i.e. we use `return ok()` but then `wf`.

do you have any strong reason for avoiding the `return` here?

tests/v2/test_waku_filter.nim
alrevuelta: the first 30-ish lines of each test case look almost identical (`timeout=200.milliseconds` variation in some cases). wondering if we could move more code to a common function, just as you do with `newTestWakuFilterNode`. just for the sake of readability.

tests/v2/test_jsonrpc_waku.nim
alrevuelta: perhaps @rymnc refers to a getRandomPort() function that we can use instead of manually setting a random port? this function could use a hardcoded seed so that ports are generated deterministically, but randomly.

waku/v2/node/waku_node.nim
alrevuelta: yes, that's clear. but why doing this? is it some kind of optimization? There are tons of `seq[xxx]` in our code but not in all cases we have `xxx|seq[xxx]` so I guess there is a "why" that I'm missing?

waku/v2/protocol/waku_filter/client.nim
LNSD: We use `return ok()` because we cannot do just `ok()`. But this is only true if the is annotated with `{.async.}` pragma and returns a result. In that case, the compilation fails, and it refuses to compile. So the only solution is to do a `return ok()`.

As reference: https://status-im.github.io/nim-style-guide/language.result.html#practical-notes

alrevuelta: okay great input. I only tried **without** {.async.} and `ok()` works, but yeah not with {.async.}, where `return` is required.

But this kind of backs the decision of enforcing `return` since in some cases without it, it doesn't work (like ok() inside {.async.}

jm-clius (approved):
LGTM! Feel free to merge.
Two comments:
1. Not related to this PR, but do we know if the failing Jenkins Linux CI test is related to RLN?
2. Note that client teams are starting to dogfood `filter` and `lightpush` from Thursday noon-ish (everyone will be installing Status Desktop at the same time, some will force it to filter and lightpush and the usual communication channels on Discord will be set to read only). I know we're not deploying due to config changes, but probably worth focusing on deploying these changes to Status fleets next week before the release so they form part of some dogfooding.

LNSD:
> * Not related to this PR, but do we know if the failing Jenkins Linux CI test is related to RLN?

The `RangeDefect` issue was patched in RLN by @rymnc a couple of days ago. That test case should not fail.

The test cases have been failing intermittently, but I couldn't pinpoint the root cause since the failures are not consistent at first sight. It will need a more thorough investigation.

Additionally, I have been doing these code refactoring PRs to give random TCP port numbers to the different nodes in the test cases. Hence, we have a temporary patch for the test failure chains (e.g., `TransportError`). Of course, there are other solutions to this failure chain, but this was the simplest 😁

> * Note that client teams are starting to dogfood filter and lightpush from Thursday noon-ish (everyone will be installing Status Desktop at the same time, some will force it to filter and lightpush and the usual communication channels on Discord will be set to read only).

I was foreseeing it; that's one reason I prioritized these PRs. 

>  I know we're not deploying due to config changes, but probably worth focusing on deploying these changes to Status fleets next week before the release so they form part of some dogfooding.

Before trying to deploy, we need @jakubgs assistance updating configuration management in the infra repository. We can prioritize this, sure.

Also, I am having some issues with the `libpcre.so` in my machine (Linux Arch); the `wakunode2` binary cannot start because it is not dynamically finding the shared library in my OS at runtime because the library name in my OS is different than in Ubuntu.

Issue #1189, opened 2022-09-27
chore(rln-relay): handle Exceptions and CatchableErrors


rymnc:
### Background
context: https://github.com/status-im/nwaku/pull/1185#discussion_r981043763

### Details
Refactor functions into using the Result type, and raise the error as is

### Acceptance criteria
- [x] Convert all functions that may raise an error into a result type

Issue #1245, opened 2022-10-06
chore(rln-relay): tighten error results with enums


rymnc:
### Background
In the waku-rln-relay module, there are multiple usages of regular errors (strings), which makes it difficult to handle each type of error thrown from a proc.

### Details
Create error enums for procedures and handle all cases appropriately

### Acceptance criteria
- [ ] Error enums for each procedure
- [ ] Multiple steps: Handle each type of error thrown from a proc

Issue #1266, opened 2022-10-14
feat(rln-relay): The window of acceptable roots should change per block, not per event 


rymnc:
### Problem
Currently, the window of acceptable roots changes per event that is processed. With the current acceptable root window size (5), this would mean that if a block has more than 5 registrations, the previous acceptable window of roots would be invalidated, and a *poorly* connected node would not be able to send messages till it catches up. 

### Suggested solution
The acceptable window of roots must be updated only *after* all the events in the block are processed. This would also ensure that there would not be any inconsistency while processing multiple events at the same time (MemberRegistrations, MemberDeletions). This would then allow *poorly* connected nodes to still participate in messaging, since they still hold a valid root.

### Alternatives considered
We could expand the size of the acceptable window of roots dynamically depending on the count of tree changing events emitted per block.

### Additional context
None

### Acceptance criteria
- [ ] After all events are processed in a block, update the window of roots

rymnc:
> On top of this, I would also probably wait transactions to be confirmed, i.e. wait ~6 blocks confirmations. If a block containing registrations is not confirmed, we have no way to revert except restarting nodes. Waiting would be beneficial for confirming the registrations and to allow a bit more time to poorly connected nodes to recover the block.

I don't think we should wait for 6 confirmations, we should have additional logic to handle forks.

PR #1321, opened 2022-11-01
chore(rln-relay): remove all raises and replace with Result types

rymnc:
Should resolve https://github.com/status-im/nwaku/issues/1189, and allow work to be started on https://github.com/status-im/nwaku/issues/1245

List of changes -
  - `mountRlnRelay` is now `async`: Removed `waitFor`
  - Procs that do not raise exceptions, but return their errors with the `RlnRelayResult` type -
  	- `mountRlnRelayDynamic`  
  	- `mountRlnRelayStatic`  
  	- `mount`
  	- `toMembershipKeyPairs`
  	- `membershipKeyGen`
  	- `createMembershipList`
  	- ... and more

  - Replaced many `var` with `let`, can create a new issue to clean up outliers


LNSD (approved):
LGTM ✅

jm-clius (approved):
Good stuff! :)

waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
```diff
@@ -1340,26 +1387,25 @@ proc mount(node: WakuNode,
...
 
 
 proc mountRlnRelay*(node: WakuNode,
                     conf: WakuNodeConf|Chat2Conf,
                     spamHandler: Option[SpamHandler] = none(SpamHandler),
                     registrationHandler: Option[RegistrationHandler] = none(RegistrationHandler)
-                   ): RlnRelayResult[bool] {.raises: [Defect, ValueError, IOError, CatchableError, Exception].} =
-  waku_rln_relay_mounting_duration_seconds.nanosecondTime:
-    let res = mount(
+                   ): Future[RlnRelayResult[void]] {.async.} =
```
staheri14, line 1403: why do we need this change?

waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
```diff
@@ -614,28 +612,33 @@ proc validateRoot*(wakuRlnRelay: WakuRLNRelay, root: MerkleNode): bool =
...
+        idCommitment = hexToUint[IDCommitment.len*8](groupKeys[i][1]).toBytesLE()
+      groupKeyPairs.add(MembershipKeyPair(idKey: idKey,
+          idCommitment: idCommitment))
+    except ValueError as err:
+      warn "could not convert the group key to bytes", err = err.msg
+      return err("could not convert the group key to bytes: " & err.msg)
+  return ok(groupKeyPairs)
+
+proc calcMerkleRoot*(list: seq[IDCommitment]): RlnRelayResult[string] =
   ## returns the root of the Merkle tree that is computed from the supplied list
```
staheri14, line 635: The proc description should change to indicate that the Result output may contain an error if the operation does not go through. This shall be done for all the procs in this PR.

waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
```diff
@@ -171,12 +168,13 @@ when defined(rlnzerokit):
     var
       keypair = MembershipKeyPair(idKey: secret, idCommitment: public)
 
-    return some(keypair)
+    return ok(keypair)
 
-proc createRLNInstance*(d: int = MerkleTreeDepth): RLNResult {.raises: [Defect, IOError].} =
+proc createRLNInstance*(d: int = MerkleTreeDepth): RLNResult =
```
staheri14, line 173: There seems to be no possibility for errors in this proc, right? then why result type for the output?

staheri14 (commented):
LGTM in general, please see my comments below.

apps/chat2/chat2.nim
```diff
@@ -545,8 +545,8 @@ proc processInput(rfd: AsyncFD) {.async.} =
           echo "You are registered to the rln membership contract, find details of your registration transaction in https://goerli.etherscan.io/tx/0x", txHash
         
         echo "rln-relay preparation is in progress..."
-        let res = node.mountRlnRelay(conf = conf, spamHandler = some(spamHandler), registrationHandler = some(registrationHandler))
-        if res.isErr:
+        let res = await node.mountRlnRelay(conf = conf, spamHandler = some(spamHandler), registrationHandler = some(registrationHandler))
```
staheri14, line 548: Why treat it as an async proc?

waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
rymnc: When the rln instance creation fails, etc we do return an error as part of the RlnResult. This would allow appropriate error handling.

waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
rymnc: 1. We don't really need to use bool for a result unless necessary. In this case, we just need to check if it mounted successfully, which is indicated by a success result, i.e ok()
2. It is a Future now since we moved from using waitFor to async processing

apps/chat2/chat2.nim
rymnc: Because the dynamic mounting proc is async. We're trying to avoid use of waitFor unless absolutely necessary (i.e in entrypoints)

waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
rymnc: Addressed in [3570dd1](https://github.com/status-im/nwaku/pull/1321/commits/3570dd18f9dbf444614b7eb5b767e0b11ce2a4da)

apps/chat2/chat2.nim
LNSD: > Because the dynamic mounting proc is async. We're trying to avoid use of waitFor unless absolutely necessary (i.e in entrypoints)

Exactly, this is the rule we are following. Only use `waitFor` at the application level (e.g., `chat2` or `wakunode2`)

The reason is that using `waitFor` indiscriminately can lead to a live-lock situation where the chronos' futures won't progress. This situation can happen if we wait for a method that (at some point deep in the call hierarchy) also contains a `waitFor`. The clearest example is `asyncTest` method: which contains a `waitFor` keeping the chronos' scheduler polling until all the asynchronous methods/futures have been completed.

rymnc:
Merged since this was blocking #1266

PR #1325, opened 2022-11-02
refactor(sqlite): move migrations logic to sqlite common module

LNSD:
As an intermediate step, I decided to do a minor cleanup before we take another decision about what to do with the SQLite migrations. I moved the logic to the `waku/common` module where SQLite database code was moved previously.

- [x] Move migrations to `common/sqlite` module since this code is agnostic to any implementation.


waku/common/sqlite/database.nim
```diff
@@ -0,0 +1,342 @@
...
+  ## two between 512 and 65536 inclusive.
+  var size: int64
+  proc handler(s: RawStmtPtr) = 
+    size = sqlite3_column_int64(s, 0)
+
+  let res = db.query("PRAGMA page_size;", handler)
+  if res.isErr():
+      return err("failed to get page_size")
+
+  ok(size)
```
rymnc, line 271: Is there a format for returning results? Maybe an addition to the style guide - `ok` vs `return ok`

waku/common/sqlite/migrations.nim
```diff
@@ -0,0 +1,138 @@
...
+  try: 
+    for scriptPath in walkDirRec(path):
+      if isSqlScript(scriptPath):
+        scripts.add(scriptPath)
+      else:
+        debug "invalid migration script", file=scriptPath
+  except OSError:
+    return err("failed to list migration scripts: " & getCurrentExceptionMsg())
+
+  return ok(scripts)
```
rymnc, line 43: ditto

rymnc (approved):
LGTM!

waku/common/sqlite/migrations.nim
LNSD: I missed this one. IMO we should go with no return unless necessary. Solving it

waku/common/sqlite.nim
```diff
@@ -1,334 +1,7 @@
-{.push raises: [Defect].}
+{.push raises: [].}
```
jm-clius, line 1: Thanks for starting to add these. Can you see any (good) reason for us to rather use:

```
when (NimMajor, NimMinor) < (1, 4):
  {.push raises: [Defect].}
else:
  {.push raises: [].}
```
?

Perhaps if someone compiles nwaku into a system with an older nim version (unlikely and should probably not be allowed) or (perhaps more likely) that the nimbus-build-system gets downgraded due to some unforeseen issues?

jm-clius (approved):
LGTM

waku/common/sqlite.nim
LNSD: I removed the `Defect` because the compiler added extra noise because of that pragma. I did it as an experiment, and I forgot to restore it.

Let's do this in a unique PR where we change all the occurrences. WDYT? 🤔

PR #1326, opened 2022-11-02
fix: replace TaintedString type deprecated in nim 1.5

LNSD:
Fixes for the following warnings:

> /Users/runner/work/nwaku/nwaku/apps/wakunode2/config.nim(439, 49) Warning: Deprecated since 1.5; TaintedString is deprecated [Deprecated]


alrevuelta (approved):
great!

PR #1327, opened 2022-11-02
fix: replace nim's std/nre with nim-regex library

LNSD:
Nim's standard library `std/nre` loads at runtime (via `dlopen()`) the `libpcre.so.3`. 

This is not portable enough. This shared library name differs depending on the Linux distributions (e.g., in Linux Arch, the shared object is named `libpcre.so.0`).

- [x] Replace `std/nre` with `nitely/nim-regex` library

For more context, check [this forum thread](https://forum.nim-lang.org/t/6008) and [this comment](https://forum.nim-lang.org/t/6008#:~:text=We%20could%20maybe%20adopt%20nim%2Dregex%20into%20the%20stdlib%20and%20deprecate%20both%20re%20and%20nre...)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -153,3 +153,13 @@
 	url = https://github.com/vacp2p/zerokit.git
 	ignore = dirty
 	branch = master
+[submodule "vendor/nim-regex"]
+	path = vendor/nim-regex
+	url = https://github.com/nitely/nim-regex.git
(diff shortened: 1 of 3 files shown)


jm-clius (approved):
Don't know much about this issue, but based on forum thread this seems to be the accepted solution.

Issue #798, opened 2021-12-08
`nim-waku` external files and resources


jm-clius:
### Problem

In some places `nim-waku` code references versioned controlled files/resources in an ad-hoc manner. For example, [migration loads migrations scripts](https://github.com/status-im/nim-waku/blob/9eb532171e5f15eee11bbe7b7ef24d00e7d678ae/waku/v2/node/storage/sqlite.nim#L253) from the source code directory, while [RLN references a file in the `/vendor/` subdirectory](https://github.com/status-im/nim-waku/blob/master/waku/v2/protocol/waku_rln_relay/rln.nim#L9-L15). These are sometimes added, sometimes not, to the docker image, leading to bugs and inconsistencies.

### Solution

Proper resources management, with all resources being made available in a `resources/` subdirectory that is always included in the docker image and PATH.

### Acceptance criteria

- [ ] External resources for `wakunode2` (including for RLN) all reside in a single place.
- [ ] Docker image updated to reference this path

jm-clius:
Closing. RLN can now be statically compiled. Migration scripts are being moved to external process by DB operator. Similar issue with postgresql dependency being addressed here: https://github.com/waku-org/nwaku/issues/1839

PR #1330, opened 2022-11-02
chore(migrations): move migration scripts to project root

LNSD:
Following the spirit of #1325 and aligned with #798:

> As an intermediate step, I decided to do a minor cleanup before we make another decision about what to do with the SQLite migrations.

- [x] Move the SQLite database migration scripts outside the project code. Move them to the `migrations` directory in the project root
- [x] Update the script's path to the new location


jm-clius (approved):
👍

PR #1329, opened 2022-11-02
refactor(wakunode2): move configuration load logic to config module

LNSD:
In my road to support configuring `wakunode2` via environment variables, I bring this gardening PR 🌳

- [x] Move the `WakuNodeConf.load()` to the `config` module
- [x] Make `load()` to return a `ConfResult`
- [x] Manage the `--version` option with _confutils_

PR #1331, opened 2022-11-02
feat(wakunode2): support configuration via environment variables

### LINKED ISSUE #1198: feat: configuration via environment variables
### Problem
I'm always frustrated by the need to use command line arguments in order to configure `nwaku`. Different configuration management / orchestration mechanisms have a multitude of special character considerations, limitations etc, that leads to vendor specific restrictions with respect to command line arguments.

### Suggested solution
Implement a concise configuration engine, including:

- [ ] Sane defaults - no configuration for the base-line is better than having to specify anything.
- [ ] Command line arguments
- [ ] Environment variables
- [ ] Configuration file (TOML, ini, YAML etc).

The method of configuration should take into account the above order of priority to allow for overwriting configuration options when deemed necessary by the system administrator / packager.

### Additional context
This issue has arisen when implementing `dappnode` packaging for `nwaku`. As `dappnode` uses `docker`, it's heavily biased towards environment variable configuration (with command argument configuration being more fragile).

### Acceptance criteria

- [ ] Configuration order of priority (config file < env var < command line argument)
- [ ] Environment variable configuration

LNSD:
As part of the effort to conform to the [twelve-factor app](https://12factor.net/) principles. And regarding the configuration, [the III principle](https://12factor.net/config), this PR adds support to configure the `wakunode2` app via environment variables.

This PR resolves #1198 

------

Environment variables should be prefixed by the app's name. For example, to set the `--tcp-port` configuration we should call `wakunode2` binary as follows:

```bash
WAKUNODE2_TCP_PORT=65000 ./build/wakunode2
```

And this is the configuration precedence order (each item takes precedence over the item below it):

1. Commandline options/flags
2. Environment variables
3. Configuration file (currently, only TOML is supported)
4. Default value

This precedence order also aligns with other languages' de facto standards like [Go's Viper](https://github.com/spf13/viper#:~:text=Viper%20uses%20the,default).

-----

In `nim-cofutils` the environment variables support is intended to be provided using `nim-serialization` (actually, `envvar_serialization`). So yes, `nim-confutils` documentation (a.k.a. README.md) is outdated. And added to that, the code that parses the environment variables in the `master` branch is not working as expected.

This left me with only one option: "fork" the code. At some point, we'll want to upstream it back to `nim-confutils` repo. 

So, the result is:

- [x] Add the "envvar" config file in `wakunode2` as another configuration source
- [x] Fix environment variables name construction (in `waku/common/envvar_serialization` fork)
- [x] Mimick Toml's modules in `nim-confutils` to support deserialization (in `waku/common/confutils/envvar`)

### DIFF
--- a/apps/wakunode2/config.nim
+++ b/apps/wakunode2/config.nim
@@ -12,10 +12,15 @@ import
   libp2p/crypto/crypto,
   libp2p/crypto/secp,
   nimcrypto/utils
+import
+  ../../waku/common/confutils/envvar/defs as confEnvvarDefs,
+  ../../waku/common/confutils/envvar/std/net as confEnvvarNet
 
 export
   confTomlDefs,
-  confTomlNet
+  confTomlNet,
+  confEnvvarDefs,
+  confEnvvarNet
 
 
 type ConfResult*[T] = Result[T, string]
@@ -506,6 +511,12 @@ proc readValue*(r: var TomlReader, value: var crypto.PrivateKey) {.raises: [Seri
   except CatchableError:
     raise newException(SerializationError, getCurrentExceptionMsg())
 
+proc readValue*(r: var EnvvarReader, value: var crypto.PrivateKey) {.raises: [SerializationError].} =
+  try: 
+    value = parseCmdArg(crypto.PrivateKey, r.readValue(string))
+  except CatchableError:
+    raise newException(SerializationError, getCurrentExceptionMsg())
+
 
 {.push warning[ProveInit]: off.}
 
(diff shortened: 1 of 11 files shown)


LNSD:
The present PR depends on #1329

alrevuelta (commented):
Can we document how env flags are used somewhere?  i.e. that flags follow the `WAKUNODE2_flag` pattern. I usually prefer to add the doc with the feature in the same PR, in case I forgot in the future.

Since @mfw78 brought this up, maybe you have some comments?

LNSD:
> Can we document how env flags are used somewhere? i.e. that flags follow the `WAKUNODE2_flag` pattern. I usually prefer to add the doc with the feature in the same PR, in case I forgot in the future.

I prefer doing it in a separate PR where I only change the documentation. Otherwise, I have to wait for quite a lot for all checks to pass every time that I change something in the docs part of the PR.

jm-clius (approved):
Thanks! This is really great. Your prediction was right - I do like this PR. :)

alrevuelta (approved):
lgtm! thanks for the comments! indeed waiting again for ci would be too much for just a readme update.

Issue #1010, opened 2022-06-21
Waku v2 monitoring node


jm-clius:
## Problem

We need a node that can be deployed to a specific Waku v2 network that can be used to gather various network metrics and monitor the overall health of the network. Note that this should not be confused with the ["Waku v2 canary service"](https://github.com/status-im/nwaku/issues/754) which will function more like a tool that can be used for checking the health of specific nodes on an ad-hoc basis.

## Suggestion

This node could, for example,
- "walk" the discv5 DHT in the network to get an idea of the number of nodes participating in the network
- try to get an idea of the number of unique peer IDs seen in the network (using various discovery methods)
- keep track of the number of messages per:
  - protocol
  - content topic
  - application (first part of content topic)

## Tracking:

- [x] metrics on discovered and connected peers: https://github.com/status-im/nwaku/pull/1290
- [x] metrics on messages (application, topic, etc): #1335 
- [ ] run an instance of `networkmonitor` exposing its data in a grafana dashboard.
  - [x] https://github.com/status-im/infra-misc/issues/104#issuecomment-1315622948
  - [ ] https://github.com/status-im/nwaku/pull/1401

jm-clius:
@D4nte I think this node is closer to what you had in mind in this comment: https://github.com/status-im/nwaku/issues/754#issuecomment-1161443548

alrevuelta:
@jm-clius This is something that interests me and I think it would be a great way of starting with both nim and libp2p, perhaps using `nim-libp2p`. I've done similar work in the Ethereum consensus layer with [armiarma](https://github.com/migalabs/armiarma). This project:
* Discovers peers using `discv5`, estimating the total peers the network has.
* Pings each peer attempting a connection in order to fetch some metadata (mainly the UserAgent) to see:
  * Which client software (prysm, nimbus, teku, lighthouse) the node it running.
  * Build version and OS (usually part of the UserAgent)
  * If the node is synced to the latest head
* The crawler also listens to all pub-sub messages in the network (topics: `beacon_block`, `voluntary_exit`, etc)

jm-clius:
@alrevuelta, indeed! With the operator trial having just launched, the value that a monitoring node will add has grown significantly. A basic monitoring node is indeed a good next step after the basic canary service to estimate at least two things as a start: estimate network size using `discv5` and estimate the number of active applications (using `contentTopic` prefix) on the default network. Liveness checks, a way of determining client type, version, etc. would be a very valuable bonus!

[...]

PR #1335, opened 2022-11-03
chore(networkmonitor): add metric listing content topics + messages

alrevuelta:
New feature for `networkmonitor`:
- [x] Subscribes to `DefaultPubsubTopic` and counts amount of messages rx in each content topic
- [x] Add a new endpoint to get the metrics on content topics and message count. `http://127.0.0.1:8009/contenttopics`
- [x] Stop storing content topics after the table reaches x size.

Notes:
* In the future we may want to count a message rate, measured in msg per unit of time. A simple counter will overflow at some point.
* Everything is in memory, which might be inviable depending on the number of content topics existing in the network. Note that there table has a limit. When exceeded, it removes the content topics with the least messages as protection.

Related #1010 
Depends on #1290


jm-clius (approved):
Great stuff!

PR #1336, opened 2022-11-03
chore: randomize waku tcp ports in tests

LNSD:
Test cases have been failing with `[LPError] Address in use` because most of the test cases' nodes were using port 60000. 

This is a temporary patch to decouple the test cases until we support getting a random port when initializing the waku node.


rymnc (approved):
LGTM :)

PR #1337, opened 2022-11-03
feat(testlib): mark commented test cases as skipped

LNSD:
Previosly, the test cases were removed from the execution, and there was no way to notice if any test case was still commented. 

Now, the test cases marked to be skipped with an ✖️ will appear as `[SKIPPED]` in the execution logs.

### DIFF
--- a/tests/v2/testlib/testutils.nim
+++ b/tests/v2/testlib/testutils.nim
@@ -1,12 +1,25 @@
-template suitex*(name: string, body: untyped) = discard
+import
+  testutils/unittests
+
 template xsuite*(name: string, body: untyped) = discard
+template suitex*(name: string, body: untyped) = discard
 
-template testx*(name: string, body: untyped) = discard
-template xtest*(name: string, body: untyped) = discard
-
-template procSuitex*(name: string, body: untyped) = discard
 template xprocSuite*(name: string, body: untyped) = discard
+template procSuitex*(name: string, body: untyped) = discard
 
-template asyncTestx*(name: string, body: untyped) = discard
-template xasyncTest*(name: string, body: untyped) = discard
-  
+
+template xtest*(name: string, body: untyped) =
+  test name:
+    skip()
+
+template testx*(name: string, body: untyped) =
+  test name:
+    skip()
+
+template xasyncTest*(name: string, body: untyped) =
+  test name:
+    skip()
+
+template asyncTestx*(name: string, body: untyped) =
+  test name:
+    skip()


alrevuelta (commented):
great! can we also mark as SKIPPED the testsuites? Actually when I tried to test this PR that's what I tried and then I realised that its only for testcases. Both are important imho

LNSD:
> can we also mark as SKIPPED the testsuites?

No. AFAIK `nim-unittest2` (or `nim-testutils`) does not support "skipping" the test suites

alrevuelta:
> > can we also mark as SKIPPED the testsuites?
> 
> No. AFAIK `nim-unittest2` (or `nim-testutils`) does not support "skipping" the test suites

Am I missing something?

```nim
template xprocSuite*(name: string, body: untyped) =
  test name:
    skip()
```

```nim
xprocSuite "WakuNode":
...
```

```console
[SKIPPED] WakuNode
```

LNSD:
> Am I missing something?

Yes

> 
> ```nim
> template xprocSuite*(name: string, body: untyped) =
>   test name:
>     skip()
> ```

You are hacking the thing. You are using a `test` template. You are not using a `suite` template.

---

As I explained above, `suite` templates fail to compile if you do the following:

```nim
template xprocSuite*(name: string, body: untyped) =
  suite name:
    skip()
```

alrevuelta:
> You are hacking the thing. You are using a test template. You are not using a suite template.

it marks the suite as skipped. so unless its doing something else that we don't want, whats wrong with the hack?

rymnc:
Slightly unrelated, is there a way we can just grep tests without altering code? see [rust, running single tests](https://doc.rust-lang.org/book/ch11-02-running-tests.html#running-single-tests)

LNSD:
> Slightly unrelated, is there a way we can just grep tests without altering code? see [rust, running single tests](https://doc.rust-lang.org/book/ch11-02-running-tests.html#running-single-tests)

This is what I am doing: https://github.com/status-im/nwaku/blob/master/waku/v1/README.md#:~:text=You%20can%20also%20run%20a%20specific%20test%20(and%20alter%20compile%20options%20as%20you%20want)%3A

There is [`nim-testutils` testrunner](https://github.com/status-im/nim-testutils), but we are not using it. And sincerely, I am not sure of the state of the tool.

LNSD:
> whats wrong with the hack?

It is a hack.

----

If we replace a test suite with a test case template, the Junit reports that unittes2 supports generating will be incorrect; and as a consequence, the following won't work as expected either: https://github.com/marketplace/actions/junit-report-action

Having said that, feel free to investigate a non-hackish solution to cover this use case 🙂

alrevuelta (approved):
great, thanks for the comments!

jm-clius (approved):
LGTM!

Issue #1002, opened 2022-06-14
`nwaku` binaries name change


jm-clius:
### Problem

The `nim-waku` repo and project has recently been renamed to `nwaku`.
However, the binaries compiled from the repo still use a naming scheme where `nwaku` is never mentioned.
Examples of binaries compiled from the `nwaku` repo:
- `wakunode1`
- `wakunode2`
- `chat2`
- `wakubridge`
- `sim2`
etc.

This may be confusing, e.g. when:
- any of the binaries are published as a standalone
- a binary is embedded as a service node in another application
- from a marketing perspective

### Proposal

This issue proposes that we rethink the binary naming scheme, preferably adopting a scheme similar to [nimbus-eth2](https://github.com/status-im/nimbus-eth2) for organisational consistency.

PR #1339, opened 2022-11-03
refactor(peer_manager): move peer_store under peer_manager module

LNSD:
As part of the work done under #1293, the peer persistence database was decoupled from the message store database connection. This PR continues that work:

- [x] Moved `peer_store` module under `peer_manager` module. The code has been collocated with the peer manager code, as it is the only module depending on the peer store.
- [x] Decoupled `migrations.migrate()` procedure from the message store's.
- [x] Used a different migrations `SchemaVersion` than the message's store database. 
- [x] Updated the `wakunode2` module, the tests, and imports in general accordingly.

### DIFF
--- a/apps/wakunode2/wakunode2_setup_rest.nim
+++ b/apps/wakunode2/wakunode2_setup_rest.nim
@@ -13,7 +13,7 @@ import
 
 
 logScope:
-  topics = "wakunode.setup.rest"
+  topics = "wakunode rest"
 
 
 proc startRestServer*(node: WakuNode, address: ValidIpAddress, port: Port, conf: WakuNodeConf) = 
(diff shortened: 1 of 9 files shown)


apps/wakunode2/wakunode2_setup_rest.nim
```diff
@@ -13,7 +13,7 @@ import
 
 
 logScope:
-  topics = "wakunode.setup.rest"
+  topics = "wakunode rest"
```
jm-clius, line 16: I know not part of this PR, but FWIW I like that the app is just `wakunode` like here - I'm getting a bit wary of having a confusing `2` everywhere, when we're practically always working with version 2.

jm-clius (approved):

apps/wakunode2/wakunode2_setup_rest.nim
LNSD: Yes... I also have that concern. To achieve that, we need to coordinate with the infra changes. I have avoided mixing too many infra changes for this release, we are already changing the configuration.

The related issue is: https://github.com/status-im/nwaku/issues/1002

PR #1340, opened 2022-11-03
chore: remove unused imports

LNSD:
A gardening PR 🌳 cleaning up all the `[UnusedImport]` warnings


jm-clius (approved):
Issa lot. 😅 Do we know much yet about why unused import detection seems to be better in the 1.6 compiler? Haven't looked into this yet.

rymnc (approved):
LGTM!

alrevuelta (approved):
wow, thanks for this. actually i wouldn't mind if it errors when having unused imports. but based on the compiler flags the closest i find is warningAsError:X:on|off, and that would be too strict imho.

LNSD:
Regarding _Warnings_ in general (not only the `[UnusedImport]` warning). It should be the developer's responsibility to avoid adding more warnings. And it should be a continuous maintenance task.

We should reduce the warnings to the minimum number to avoid [alert fatigue](https://www.atlassian.com/incident-management/on-call/alert-fatigue). With those many warnings, we end up not paying attention to them, which could mask more severe issues.

I took the task this time and did it, but please try to keep the warnings number low.

LNSD:
> Do we know much yet about why unused import detection seems to be better in the 1.6 compiler?

No idea at all 🤷🏼 But it is excellent! Because before the upgrade, I knew that there were modules that were not used, and I had to do the test manually (remove, compile and repeat).

Issue #1033, opened 2022-07-16
Upgrade nwaku to Nim `v1.6.6`


jm-clius:
This issue may be more of an investigation than a singular task.

Quick exploration by @LNSD shows that compilation fails if setting the nimbus-build-system to Nim 1.6.6. This issue tracks effort to make this work and upgrade to Nim 1.6.6.

LNSD:
@Menduist I remember from one of our offsite conversations that there's some limitation right now that prevents us from moving to Nim `v1.6.6`. Is libp2p supporting `v1.6.6` yet?

Menduist:
libp2p supports 1.6.6
Nimbus briefly switched to 1.6.6, but it got rolled back because of a showstopper https://github.com/nim-lang/Nim/issues/20107 that has since been fixed, but not yet backported

narimiran:
> fixed, but not yet backported

It is backported now.

LNSD:
As per the discussion with @jm-clius yesterday, we are postponing the Nim upgrade for one of the subsequent releases. Removing it from the release/milestone v0.12.

jm-clius:
Nim version bumped in submodule update: https://github.com/status-im/nwaku/pull/1307

Outstanding tasks for 1.6.6:
- [ ] top level {.push raises: [].}
- [ ] work through most important deprecated methods and other compiler warnings

alrevuelta:
Closed in https://github.com/status-im/nwaku/pull/1307?

jm-clius:
@LNSD I think we can consider this done (thanks to the extra tasks you did to improve compiler messages, correct raises requirements, etc.)? May still be minor tasks that come up, but we can deal with them as they arise. Feel free to reopen if disagree.

PR #1341, opened 2022-11-03
chore: update push raises pragma according to nim 1.6

LNSD:
This PR covers part of the work pending after the Nim upgrade to 1.6 (#1033). It replaces the top level `{.push raises: [Defect].}` with `{.push raises: [].}` keeping backward compatibility with lower nim versions.


jm-clius (approved):
Thanks for jumping in and just doing it! :)

alrevuelta:
@LNSD beyond the pr but related, any reason why `wakunode2/config.nim` doesn't have it? guess its the only place where we allow to just raise and uncatched exception?

PR #1342, opened 2022-11-04
refactor(message_store): move message_store to node module

LNSD:
Similar to #1339, this time with the `message_store` module.

- [x] Move the `message_store` module to the `node` module
- [x] Decoupled the `migrations.migrate()` procedure from the peer store's proc and the "shared migrations" module.
- [x] Used an independent migrations `SchemaVersion`. 
- [x] Updated the `wakunode2` module, the tests, and imports in general accordingly.
- [x] Removed the common migrations code.


LNSD:
Merging this PR since it solves the node startup issue reported by @jakubgs

PR #1343, opened 2022-11-04
ci: fix path to migrations in Dockerfile

jakubgs:
Broken by moving the `migrations_scripts` folder:
- https://github.com/status-im/nwaku/pull/1330


LNSD (approved):
IMO much cleaner! Thanks for your patch 👍🏼

PR #1344, opened 2022-11-04
chore: build rln dependecies only when building v2

### LINKED ISSUE #1313: chore(makefile): Control NIM_PARAMS and compiled deps for waku v1 and v2 targets
### Background
Currently, both v1 and v2 targets receive the same NIM_PARAMS variable and `deps` dependencies, that collects all flags to pass to executables and resources to be compiled, respectively.

These include, for example, RLN-related flags and libraries, but v1 doesn't support RLN-RELAY, hence RLN resources would be compiled even if will not be used. This is the case also for GitHub CI test1 workflows, where roughly 10 mins are spent in the compilation of rln resources. 

### Possible Solution
To give more control over what is passed to v1 and v2 targets, we can populate accordingly two distinct NIM_PARAMS_V1 and NIM_PARAMS_V2 variables in the makefile and separate `deps` into `depsv1` and `depsv2`.

### Acceptance criteria
- [ ] Make v1 and v2 targets to compile/include only necessary resources/flags.

LNSD:
This is a gardening 🌳 and CI optimization PR:

- [x] Tidy up the `Makefile`
- [x] Split v2 dependencies from `deps` target into `deps2` target
- [x] Build v2 dependencies only when building waku v2 code
- [x] Add a per binary `nim.cfg` for static nim defines and flags

This PR resolves #1313

----

There is some work pending to decouple the `NIM_PARAMS` as @s1fr0 pointed out in #1313, but this work should be taken in a separate piece for the sake of simplicity.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,24 +18,47 @@ EXCLUDED_NIM_PACKAGES := vendor/nim-dnsdisc/vendor
 
 LINK_PCRE := 0
 
+
+# Get git version
+GIT_VERSION ?= $(shell git describe --abbrev=6 --always --tags)
+
+# Detecting the os
+ifeq ($(OS),Windows_NT) # is Windows_NT on XP, 2000, 7, Vista, 10...
+ detected_OS := Windows
+else ifeq ($(strip $(shell uname)),Darwin)
+ detected_OS := macOS
+else
+ # e.g. Linux
+ detected_OS := $(strip $(shell uname))
+endif
+
+
 # we don't want an error here, so we can handle things later, in the ".DEFAULT" target
 -include $(BUILD_SYSTEM_DIR)/makefiles/variables.mk
 
 .PHONY: \
 	all \
-	deps \
-	update \
-	sim1 \
-	wakunode1 \
-	wakunode2 \
-	example1 \
-	example2 \
-	wakubridge \
 	test \
+	update \
 	clean \
+	libbacktrace \
+	deps \
+	deps2 \
+	v1 \
+	v2 \
+	test1 \
+	test2 \
+	wakunode1 \
+	example1 \
+	sim1 \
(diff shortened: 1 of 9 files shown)


LNSD:
> what kind of flags should go to each app nim.cfg file and what instead should go in [waku.nimble](https://github.com/status-im/nwaku/blob/master/waku.nimble?rgh-link-date=2022-11-04T18%3A29%3A57Z)?

I do not have a clear answer right now, at least to establish a set of good practices. But, I will do a second iteration on the configuration and build scripts in the following weeks.

But, this is how I see it right now: we should put the static nim flags in a single place (e.g., the application `nim.cfg` file). The dynamic options/flags are the part that I do not have a clear idea of how to manage, yet.

Makefile
```diff
@@ -55,118 +78,156 @@ else # "variables.mk" was included. Business as usual until the end of this file
...
+ifeq ($(detected_OS), macOS)
+NIM_PARAMS := $(NIM_PARAMS) -d:onchain_rln
+endif
+endif
+endif
+
+
 installganache:
 ifeq ($(ONCHAIN_RLN), true) 
 	npm install ganache-cli; npx ganache-cli -p	8540	-g	0	-l	3000000000000&
```
LNSD, line 233: @staheri14 @s1fr0 `ganache-cli` node package is deprecated. Check the NPM page for more context: https://www.npmjs.com/package/ganache-cli

Additionally, we should move the "ganache-cli" logic in the Makefile to the test suite itself.

How we should proceed with this?

LNSD:
Merging this PR as #1347 depends on it. Moving [this comment](https://github.com/status-im/nwaku/pull/1344#discussion_r1014332523) to #1347 since it is still relevant. Waiting for a response.

PR #1345, opened 2022-11-04
chore(wakucanary): remove port hardcoding + add cli config

alrevuelta:
* listening port for wakunode was hardcoded to 60000
* remove hardcoding and add a new cli flag `--node-port` to configure the port


LNSD (approved):
LGTM ✅

jakubgs (approved):
Nice, thanks for quick response.

PR #1346, opened 2022-11-04
chore: rename protocols rpc_codec procs from init to decode

LNSD:
This is another gardening PR 🌳 

- [x] Rename RPC types codec methods from init to decode
- [x] Move `PubsubTopic` type from `waku_node` module to `waku_message` module


rymnc (approved):
LGTM

PR #1347, opened 2022-11-06
chore(rln): update ganache

s1fr0:
The npm package `ganache-cli` employed by onchain RLN tests is [deprecated](https://www.npmjs.com/package/ganache-cli) and is now replaced by `ganache`. 

This issue was originally raised in https://github.com/status-im/nwaku/pull/1344#discussion_r1014434448.

### Scope
This PR:
- [x] replaces `ganache-cli` with `ganache`. 
This update introduced some breaking changes which required some parameters previously passed to ganache-cli to be revised.
- [x] moves ganache installation logic from Makefile to the waku_rln_relay_onchain test suite. 
This allows to run ganache only during execution of onchain tests and not during the whole duration of tests2. Furthermore, onchain RLN tests will be executed by CI on both linux and macOS platform (before only macOS was ran these tests in CI).
- [x] adds gracefully termination of `ganache` by sending a SIGINT signal.
- [x] uninstalls `ganache` when tests are terminated.

### Out of scope
Any other extra feature.

### DIFF
--- a/apps/chat2/chat2.nim
+++ b/apps/chat2/chat2.nim
@@ -542,7 +542,21 @@ proc processInput(rfd: AsyncFD) {.async.} =
           echo "You are registered to the rln membership contract, find details of your registration transaction in https://goerli.etherscan.io/tx/0x", txHash
         
         echo "rln-relay preparation is in progress..."
-        let res = await node.mountRlnRelay(conf = conf, spamHandler = some(spamHandler), registrationHandler = some(registrationHandler))
+        
+        let rlnConf = WakuRlnConfig(
+          rlnRelayDynamic: conf.rlnRelayDynamic,
+          rlnRelayPubsubTopic: conf.rlnRelayPubsubTopic,
+          rlnRelayContentTopic: conf.rlnRelayContentTopic,
+          rlnRelayMembershipIndex: conf.rlnRelayMembershipIndex,
+          rlnRelayEthContractAddress: conf.rlnRelayEthContractAddress,
+          rlnRelayEthClientAddress: conf.rlnRelayEthClientAddress,
+          rlnRelayEthAccountPrivateKey: conf.rlnRelayEthAccountPrivateKey,
+          rlnRelayEthAccountAddress: conf.rlnRelayEthAccountAddress,
+          rlnRelayCredPath: conf.rlnRelayCredPath,
+          rlnRelayCredentialsPassword: conf.rlnRelayCredentialsPassword
+        )
+
+        let res = await node.mountRlnRelay(conf=rlnConf, spamHandler = some(spamHandler), registrationHandler = some(registrationHandler))
         if res.isErr():
           echo "failed to mount rln-relay: " & res.error()
         else:
--- a/apps/wakunode2/wakunode2.nim
+++ b/apps/wakunode2/wakunode2.nim
@@ -358,8 +358,22 @@ proc setupProtocols(node: WakuNode, conf: WakuNodeConf,
   
   when defined(rln) or defined(rlnzerokit): 
     if conf.rlnRelay:
+      
+      let rlnConf = WakuRlnConfig(
+        rlnRelayDynamic: conf.rlnRelayDynamic,
+        rlnRelayPubsubTopic: conf.rlnRelayPubsubTopic,
+        rlnRelayContentTopic: conf.rlnRelayContentTopic,
+        rlnRelayMembershipIndex: conf.rlnRelayMembershipIndex,
+        rlnRelayEthContractAddress: conf.rlnRelayEthContractAddress,
+        rlnRelayEthClientAddress: conf.rlnRelayEthClientAddress,
+        rlnRelayEthAccountPrivateKey: conf.rlnRelayEthAccountPrivateKey,
+        rlnRelayEthAccountAddress: conf.rlnRelayEthAccountAddress,
+        rlnRelayCredPath: conf.rlnRelayCredPath,
+        rlnRelayCredentialsPassword: conf.rlnRelayCredentialsPassword
+      )
+
       try: 
-        let res = await node.mountRlnRelay(conf)
+        let res = await node.mountRlnRelay(rlnConf)
         if res.isErr():
           return err("failed to mount waku RLN relay protocol: " & res.error)
       except:
--- a/waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
+++ b/waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils.nim
@@ -20,9 +20,7 @@ import
   waku_rln_relay_metrics,
   ../../utils/time,
   ../../utils/keyfile,
-  ../../node/waku_node, 
-  ../../../../../apps/wakunode2/config,  ## TODO: Decouple the protocol code from the app configuration
-  ../../../../../apps/chat2/config_chat2,  ## TODO: Decouple the protocol code from the app configuration
+  ../../node/waku_node,
   ../waku_message
 
 logScope:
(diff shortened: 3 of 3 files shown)


rymnc (approved):
LGTM, please add package.json to the .gitignore, thanks!

tests/v2/test_waku_rln_relay_onchain.nim
```diff
@@ -104,6 +106,31 @@ proc createEthAccount(): Future[(keys.PrivateKey, Address)] {.async.} =
...
 
 procSuite "Waku-rln-relay":
+
+  ########################
+  ## Ganache installation
+  ########################
+
+  # We install Ganache
+  let installGanache = startProcess("npm", args = ["install",  "ganache"], options = {poUsePath})
+  debug "Ganache installation completed. Printing install log", returnCode=installGanache.waitForExit(), log=installGanache.outputstream.readAll()
```
LNSD, line 116: Chronicles at compile time remove all the statements below the `-d:chronicles_log_level=<level>` log level. So, in this case, if we set the log level to `INFO`, for example, it won't call `waitForExit()` nor `readAll()`.

```suggestion
  let installGanache = startProcess("npm", args = ["install",  "ganache"], options = {poUsePath})
  
  let installGanacheRc = installGanache.waitForExit()
  let installGanacheOutStream = installGanache.outputstream.readAll()
  debug "Ganache installation completed. Printing install log", returnCode=installGanacheRc, log=installGanacheOutStream
```

tests/v2/test_waku_rln_relay_onchain.nim
```diff
@@ -533,3 +560,17 @@ procSuite "Waku-rln-relay":
...
     await node2.stop()
+
+  ################################
+  ## Terminating/removing Ganache
+  ################################
+
+  # We gracefully terminate Ganache daemon
+  # We send a SIGINT signal to the runGanache PID to trigger RPC server termination and clean-up
+  debug "Sending SIGINT to Ganache", returnCode=kill(ganachePID.int32, SIGINT)
+  debug "Ganache daemon terminated. Printing run log", returnCode=runGanache.waitForExit(), log=runGanache.outputstream.readAll()
```
LNSD, line 571: Same here.

```suggestion
  # We gracefully terminate Ganache daemon
  # We send a SIGINT signal to the runGanache PID to trigger RPC server termination and clean-up
  let ganacheKillRc = kill(ganachePID.int32, SIGINT)
  debug "Sent SIGINT to Ganache", returnCode=ganacheKillRc

  let ganacheDaemonRc = runGanache.waitForExit()
  let ganacheDaemonOutStream = runGanache.outputstream.readAll()
  debug "Ganache daemon terminated. Printing run log", returnCode=ganacheDaemonRc, log=ganacheDaemonOutStream
```

tests/v2/test_waku_rln_relay_onchain.nim
```diff
@@ -533,3 +560,17 @@ procSuite "Waku-rln-relay":
...
+
+  # We uninstall Ganache
+  let uninstallGanache = startProcess("npm", args = ["uninstall",  "ganache"], options = {poUsePath})
+  debug "Ganache uninstall completed. Printing uninstall log", returnCode=uninstallGanache.waitForExit(), log=uninstallGanache.outputstream.readAll()
```
LNSD, line 575: And here.

```suggestion
  let uninstallGanacheRc = uninstallGanache.waitForExit()
  let uninstallGanacheOutStream = uninstallGanache.outputstream.readAll()
  debug "Ganache uninstall completed. Printing uninstall log", returnCode=uninstallGanacheRc, log=uninstallGanacheOutStream`
```

LNSD:
Merged #1344 into `master`. The [following comment](https://github.com/status-im/nwaku/pull/1344#discussion_r1014332523) remained unresponded, I am bringing it here since it is still relevant.

> I think we can expand this test to both linux and macOS, I see no reason to test it only on the latter. @staheri14

tests/v2/test_waku_rln_relay_onchain.nim
```diff
@@ -104,6 +106,31 @@ proc createEthAccount(): Future[(keys.PrivateKey, Address)] {.async.} =
   return (pk, acc)
 
 procSuite "Waku-rln-relay":
+
+  ########################
+  ## Ganache installation
+  ########################
```
LNSD, line 112: Can you wrap the ganache install and start logic in a `proc` (e.g., `proc setupGanache()`) and call it here?

tests/v2/test_waku_rln_relay_onchain.nim
```diff
@@ -533,3 +560,17 @@ procSuite "Waku-rln-relay":
 
     await node.stop()
     await node2.stop()
+
+  ################################
+  ## Terminating/removing Ganache
+  ################################
```
LNSD, line 566: Same here. Can you wrap the ganache kill and uninstall logic in a `proc` (e.g., `proc cleanupGanache()`) and call it here?

LNSD (approved):
LGTM: Looks Great To Me 😁 

Thank you for taking the time to solve this issue. Very much appreciated 🙏🏼

LNSD:
> LGTM, please add package.json to the .gitignore, thanks!

@rymnc Sorry for my ignorance, but why do you need to add `package.json` to the `.gitignore`? Is ganache generating a `package.json` when it is installed? 🤔

rymnc:
> @rymnc Sorry for my ignorance, but why do you need to add `package.json` to the `.gitignore`? Is ganache generating a `package.json` when it is installed? 🤔

Yeah, the npm install adds one by default :)

LNSD:
> I think we can expand this test to both linux and macOS, I see no reason to test it only on the latter. @staheri14

My only concern about running the on-chain RLN tests in Linux is that it takes more than 1h to execute each task in the GH Actions CI. The macOS jobs were already failing due to timeout.

We need to investigate why it is taking too much time (on-chain tests are doubling the current tests duration)

tests/v2/test_waku_rln_relay_onchain.nim
alrevuelta: +1

alrevuelta (commented):
looks great! I always had the impression the Makefile wasn't the right place for ganache.

tests/v2/test_waku_rln_relay_onchain.nim
```diff
@@ -104,6 +106,31 @@ proc createEthAccount(): Future[(keys.PrivateKey, Address)] {.async.} =
...
+  ## Ganache installation
+  ########################
+
+  # We install Ganache
+  let installGanache = startProcess("npm", args = ["install",  "ganache"], options = {poUsePath})
+  debug "Ganache installation completed. Printing install log", returnCode=installGanache.waitForExit(), log=installGanache.outputstream.readAll()
+
+  # We run Ganache daemon
+  # Note that we run directly "node node_modules/ganache/dist/node/cli.js" rather than using "npx ganache", so that the daemon does not spawn in a new child process. 
+  # In this way, we can directly send a SIGINT signal to the corresponding PID to gracefully terminate Ganache without dealing with multiple processes.
```
staheri14, line 120: I'd add the description of the argument as well as a link to the ganache documentation https://github.com/trufflesuite/ganache#documentation:
`-p` stands for port, `-l` sets the gas limit, and `-e` sets the default account balance, specified in ether.

staheri14 (approved):
Good one! left a comment, also I agree with @LNSD comment about creating separate procs for ganache setup and termination.

staheri14:
> onchain RLN tests will be executed by CI on both linux and macOS platform (before only macOS was ran these tests in CI).

@s1fr0 
I think the on-chain tests will be run in CIs regardless of the OS since the OS check on the CI is removed as part of this PR.  I

rymnc:
> > I think we can expand this test to both linux and macOS, I see no reason to test it only on the latter. @staheri14
> 
> My only concern about running the on-chain RLN tests in Linux is that it takes more than 1h to execute each task in the GH Actions CI. The macOS jobs were already failing due to timeout.
> 
> We need to investigate why it is taking too much time (on-chain tests are doubling the current tests duration)

Actually, the tests fail, and I guess the ganache process is not killed appropriately (even though there is a pkill in the Makefile for it). That's why they live for an hour. This PR should resolve that!

tests/v2/test_waku_rln_relay_onchain.nim
LNSD: Indeed! This is a very good point, I completely missed it:

 @s1fr0 instead of using the short version of the flags, can you use the long version (e.g., `--port` instead of `-p`?

rymnc:
> @rymnc The tests went fine today (now jenkins seems to fail for an unrelated one) but I obseverd an absurd amount of time while building macOS (not even in test): did you notice some specific test failing that relates to this? Of course it is likely affecting something, but I'm not completely sure it's the (only) reason.

Unsure, macOS always took longer than linux to build and test

Issue #1159, opened 2022-09-15
store: Improve error reporting on waku store protocol


LNSD:
## Context/Background
Current error responses are limited, and there is some work pending. See: 
* https://github.com/status-im/nwaku/blob/87479a603be4bb3d0eb63d8c5a31bf7c6613b58a/waku/v2/protocol/waku_store/rpc.nim#L22-L25
* https://github.com/status-im/nwaku/blob/87479a603be4bb3d0eb63d8c5a31bf7c6613b58a/waku/v2/protocol/waku_store/protocol.nim#L127-L130

## Acceptance criteria
- [ ] Extend current **13/WAKU2-STORE** RFC protobuf definition to match the new error types
- [ ] Extend the error types enum
- [ ] Update the Waku Store RPC codec implementation
- [ ] Map the errors reported store query method to store enum errors

LNSD:
This task depends on the protocol wire format and RFC ongoing work. This should be moved to the next release, v0.14.0.

cc @jm-clius

LNSD:
The RFC issue tracking the Waku store RFC stabilization can be found here: https://github.com/vacp2p/rfc/issues/552

LNSD:
This issue depends on https://github.com/waku-org/pm/issues/9

PR #1350, opened 2022-11-07
refactor(store): decouple waku store public api types from rpc types

LNSD:
The Waku Store protocol wire format will be reworked in the next release. To avoid having to refactor the whole codebase because the RPC types are used everywhere, decoupling the public-facing types from the internal implementation details is a good practice. This should apply to any public-facing API (e.g., c-bindings, REST, JSON-RPC, etc.).

- [x] Define a new set of types for the waku store and the message store API in the `common` module (common to the protocol handler and client)
- [x] Add the corresponding type mappings between the RPC and the API types
- [x] Add a wrapper for the Waku Store protocol results/errors
- [x] Remove RPC types exports. If needed, RPC types should be imported explicitly


waku/v2/protocol/waku_store/common.nim
```diff
@@ -0,0 +1,104 @@
...
+  of HistoryErrorKind.PEER_DIAL_FAILURE:
+    "PEER_DIAL_FAILURE: " & err.address 
+  of HistoryErrorKind.BAD_RESPONSE:
+    "BAD_RESPONSE: " & err.cause
+  of HistoryErrorKind.BAD_REQUEST:
+    "BAD_REQUEST: " & err.cause
+  of HistoryErrorKind.SERVICE_UNAVAILABLE:
+    "SERVICE_UNAVAILABLE"
+  of HistoryErrorKind.UNKNOWN:
+    "UNKNOWN"
```
LNSD, line 104: @rymnc This is, more or less, how I am envisioning error management in the `nwaku` codebase, in case we need to differentiate between error causes. Note the enum and the object variants type (this might remember you Rust's `io:Error` 🙂 )

rymnc: very clean!!

tests/v2/test_waku_store.nim
```diff
@@ -189,18 +187,17 @@ procSuite "Waku Store - history query":
...
-    let resQuery = await client.query(rpc, peer=serverPeerInfo)
+    let queryRes = await client.query(req, peer=serverPeerInfo)
 
     ## Then
     check:
-      resQuery.isOk()
+      queryRes.isOk()
 
-    let response = resQuery.tryGet() 
+    let response = queryRes.tryGet()
```
rymnc, line 200:
```suggestion
    let response = queryRes.get() 
```
Because we've already checked if the queryRes is OK?

waku/v2/node/message_store/queue_store/index.nim
```diff
@@ -33,19 +33,19 @@ proc compute*(T: type Index, msg: WakuMessage, receivedTime: Timestamp, pubsubTo
   )
 
 
-proc toPagingIndex*(index: Index): PagingIndex =
-  PagingIndex(
+proc tohistoryCursor*(index: Index): HistoryCursor =
```
rymnc, line 36:
```suggestion
proc toHistoryCursor*(index: Index): HistoryCursor =
```

waku/v2/protocol/waku_store/client.nim
```diff
@@ -40,28 +43,45 @@ proc new*(T: type WakuStoreClient,
...
+  let respDecodeRes = HistoryRPC.decode(buf)
+  if respDecodeRes.isErr():
     waku_store_errors.inc(labelValues = [decodeRpcFailure])
-    return err(decodeRpcFailure)
+    return err(HistoryError(kind: HistoryErrorKind.BAD_RESPONSE, cause: decodeRpcFailure))
+
+  let respRpc = respDecodeRes.get()
+
+
+  # Disabled ,for now, since the default response is a possible case (no messages, pagesize = 0, error = NONE(0))
```
rymnc, line 69: Can we create a tracking issue for this?

waku/v2/protocol/waku_store/protocol.nim
```diff
@@ -78,25 +74,68 @@ proc reportStoredMessagesMetric*(w: WakuStore) =
...
+
+  let 
+    now = getNanosecondTime(getTime().toUnixFloat())
+    lowerBound = now - MaxMessageTimestampVariance
+    upperBound = now + MaxMessageTimestampVariance
+
+  return lowerBound <= msg.timestamp and msg.timestamp <= upperBound
+
+# TODO: Move to a message store wrapper
+proc handleMessage*(w: WakuStore, pubsubTopic: string, msg: WakuMessage) =
```
rymnc, line 90: Should this proc return a Result as well, since there are cases of failed message handling?

waku/v2/protocol/waku_store/protocol.nim
```diff
@@ -78,25 +74,68 @@ proc reportStoredMessagesMetric*(w: WakuStore) =
...
+    # The message is ephemeral, should not be stored
+    return
+  
+  if not isValidMessage(msg):
+    waku_store_errors.inc(labelValues = [invalidMessage])
+    return
+
 
-proc findMessages(w: WakuStore, query: HistoryQuery): HistoryResponse {.gcsafe.} =
+  let insertStartTime = getTime().toUnixFloat()
```
rymnc, line 104: https://github.com/status-im/nwaku/blob/d40636d03be11dcac22d36c3d9aa0dab0fcf97db/waku/v2/utils/time.nim#L28-L34
you can use this for Histograms now

rymnc: Can be addressed in a follow up PR :)

rymnc (approved):
LGTM, minor nits!

waku/v2/protocol/waku_store/protocol.nim
LNSD: All the "message handling and storage" logic, in other words, the logic that is not related to historical message retrieval, will be extracted into a separate entity together with the changes necessary to support the "custom" message validation.

-------

That's a good question. And the answer is "It depends": for now, I would like to leave it as is since this is used in the waku relay handler (and the filter push handler), which cannot take any action if the handling is unsuccessful. They only can log the failure/error.

I also thought about that, but it will be necessary to implement the message validation functionality. Now it provides little/no value.

rymnc: Okay :)

waku/v2/protocol/waku_store/protocol.nim
LNSD: All the "message handling and storage" logic, in other words, the logic that is not related to historical message retrieval, will be extracted into a separate entity together with the changes necessary to support the "custom" message validation.

-----

Yes, I agree. Thanks for the heads up. We'll take it when doing the message store split.

JFYI relate to the `nowInUnixFloat` and all the `times` related primitives: There is a piece of work pending to investigate and implement to use `nim-chronos` timers' time primitives to define the `Timestamp` type and all the operations.

tests/v2/test_waku_store.nim
LNSD: Yes, thx 👍🏼

waku/v2/protocol/waku_store/protocol.nim
rymnc: Okay

waku/v2/protocol/waku_store/client.nim
LNSD: It already exists: #1159

rymnc: Okay

jm-clius (approved):
LGTM. Thanks!

PR #1352, opened 2022-11-08
refactor: unify pubsub topic type and default value

LNSD:
This is a gardening PR 🌳 A long-awaited refactoring to unify the `PubsubTopic` type and some default constants: 

- [x] Move `PubsubTopic` type,`DefaultPubsubTopic` and `DefaultContentTopic` constants to `waku_message` module
- [x] Replace some string types with `PubsubTopic` type for better readability

### DIFF
--- a/tests/v2/test_rest_relay_api.nim
+++ b/tests/v2/test_rest_relay_api.nim
@@ -45,7 +45,7 @@ suite "REST API - Relay":
     restServer.start()
 
     let pubSubTopics = @[
-      PubSubTopicString("pubsub-topic-1"), 
+      PubSubTopicString("pubsub-topic-1"),
       PubSubTopicString("pubsub-topic-2"),
       PubSubTopicString("pubsub-topic-3")
     ]
@@ -204,7 +204,7 @@ suite "REST API - Relay":
     ]
     discard await client.relayPostSubscriptionsV1(newTopics)
     
-    let response = await client.relayPostMessagesV1(defaultTopic, RelayWakuMessage(
+    let response = await client.relayPostMessagesV1(DefaultPubsubTopic, RelayWakuMessage(
       payload: Base64String.encode("TEST-PAYLOAD"), 
       contentTopic: some(ContentTopicString(defaultContentTopic)), 
       timestamp: some(int64(2022))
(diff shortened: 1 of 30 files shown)


rymnc (approved):
LGTM

tests/v2/test_rest_relay_api.nim
```diff
@@ -45,7 +45,7 @@ suite "REST API - Relay":
     restServer.start()
 
     let pubSubTopics = @[
-      PubSubTopicString("pubsub-topic-1"), 
+      PubSubTopicString("pubsub-topic-1"),
```
alrevuelta, line 48: can we include in this PR removing this `PubSubTopicString`? I guess the correct way would be PubsubTopic("pubsub-topic-1)?

waku/v2/protocol/waku_relay.nim
```diff
@@ -23,7 +25,7 @@ const
 
 type
   WakuRelay* = ref object of GossipSub
-    defaultTopics*: seq[string] # Default configured PubSub topics
+    defaultTopics*: seq[PubsubTopic] # Default configured PubSub topics
```
alrevuelta, line 28: i liked that you were enforcing "topic" to be "pubsubtopic" when applicable. Does it make sense to enforce here `defaultPubsubTopic`? When having also contenttopics this seems useful to me.

alrevuelta (commented):
lgtm! great PR, really alliviates removing that many duplicated DefaultTopics. Left some comments :)

tests/v2/test_rest_relay_api.nim
LNSD: That's too much work. I think this is out of the scope of this PR. As part of one of the subsequent releases, the REST API will be finished, this will be the moment when this code will be cleaned up.

But feel free to do it yourself if you have the bandwidth.

----

That means refactoring the following types:

https://github.com/status-im/nwaku/blob/7ef4eaf46d56bbd08c5a72baed9fdd6d2d3784aa/waku/v2/node/rest/relay/api_types.nim#L20-L22

And all the associated code.

waku/v2/protocol/waku_relay.nim
LNSD: I don't want to add more cognitive load to this review, it already has 30 files.

Same, feel free to do it yourself if you have the bandwidth.

alrevuelta (approved):
lgtm! great, thanks for fixing the comment. will investigate the two comments that we have pending tomorrow now that we have the momemtum with this :)

PR #1355, opened 2022-11-08
docs: extend docs with the different node configuration mechanisms

LNSD:
I am not 100% sure if this is the right place (the correct markdown file) to provide these details. If you know a better place where to specify this information, do not hesitate to suggest it in the comments.

- [x] Extend operator's documentation with the different configuration methods supported by `wakunode2`


LNSD:
@mfw78, it would be interesting to hear your opinion on this and if this covers your requirements hence #1198

docs/operators/how-to/configure.md
```diff
@@ -1,12 +1,135 @@
...
+## Node configuration methods
+
+One node can be configured using a combination of the following methods:
+
+1. Command line options and flags
+2. Environment variables
+3. Configuration file (currently, only TOML format is supported)
+4. Default value
+
+Note the precedence order, each item takes precedence over the item below it.
```
alrevuelta, line 16: I had to read this couple of times `precedence over the item below it`. Perhaps be more explicit? "each item overrides the one below."

alrevuelta (approved):
lgtm! thanks for the PR! really relevant. left a minor comment.

docs/operators/how-to/configure.md
LNSD: How would you rephrase it? What do you suggest?

alrevuelta: > "each item overrides the one below."

Note the precedence order, where "each item overrides the one below."

mfw78 (approved):
LGTM. Great work on the configuration options! This fits perfectly with the issue that was raised 👍

docs/operators/how-to/configure.md
```diff
@@ -1,12 +1,135 @@
...
+```toml
+log-level = "DEBUG"
+tcp-port = 65000
+```
+
+The path to the TOML file can be specified usin one of the previous configuration mechanisms:
+
+* By passing the `--config-file` command line option:
+  ```shell
+  waknode2 --config-file=<path-to-toml-config-file>
```
rymnc, line 97:
```suggestion
  wakunode2 --config-file=<path-to-toml-config-file>
```

rymnc (approved):
LGTM, this looks like a good place to put it, maybe can go in its own how-to later

alrevuelta (approved):
lgtm! thanks for the example on the overrides: i.e. cli overrides env.

PR #1359, opened 2022-11-08
docs: extend waku store config tutorial of the operator docs

LNSD:
This PR covers the work pending of #1293 

Same here:

> I am not 100% sure if this is the right place (the correct markdown file) to provide these details. If you know a better place where to specify this information, do not hesitate to suggest it in the comments.

- [x] Extend the operator's documentation (the store configuration tutorial) with the changes in the waku store protocol configuration


docs/operators/how-to/configure-store.md
```diff
@@ -1,56 +1,57 @@
...
+* `--store-message-db-url`: The message store database url option controls the message storage engine. This option follows the [_SQLAlchemy_ database URL format](https://docs.sqlalchemy.org/en/14/core/engines.html#database-urls).
+
+  + SQLite engine: The only database engine supported by the nwaku node. The database URL has this shape: `sqlite://<database-file-path>`. If the `<database-file-path>` is not an absolute path (preceded by a `/` character), the file will be created in the current working directory. The SQLite engine also supports to select a non-persistent in-memory database by setting the `<database-file-path>` to `:memory:`.
+  + In the case you don't want to use a persistent message store; set the `--store-message-db-url` to an empty string, `""`. This will instruct the node to use the fallback in-memory message store.
+
+By default the node message store will be configured with a time retention policy set to `14400` seconds (4 hours). Additionaly, by default, the node message store will use the SQLite database engine to store historical messages in order to persist these between restarts.
```
rymnc, line 43:
```suggestion
By default the node message store will be configured with a time retention policy set to `14400` seconds (4 hours). Additionaly, the node message store will use the SQLite database engine to store historical messages in order to persist these between restarts. 
```

rymnc (approved):
LGTM, thanks!

docs/operators/how-to/configure-store.md
```diff
@@ -1,56 +1,57 @@
...
-./build/wakunode2 \
-  --store:true \
-  --persist-messages:true \
-  --db-path:/mnt/nwaku/data/db1/ \
-  --store-capacity:150000
+from the configured store service node. One way to trigger such queries is asking your node for historical messages using the [Waku v2 JSON RPC API](https://rfc.vac.dev/spec/16/).
+
+## Configuring the node as a store service node
+
+If the waku store node is enabled (the `--store` option is set to `true`) the node will store historical messages and will be able to serve those messages to the waku store clients.
```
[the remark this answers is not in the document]
LNSD: > Does that mean storing messages is default? (Because `--store=true` is default)

Yes, that's what I meant with _"the node will store historical messages"_.

> I'd make it explicit above that you need to set `--store=false` and set `--storenode` to setup a client only store-node.

This is not the case. Both things (server and client) are independent. You can have both on, both off, or one on and the other off.

This is what I meant with this: https://github.com/status-im/nwaku/pull/1359/files?diff=split&w=0#diff-a2d58beafcccb9e2b626c20cb0e31383f1107c520e7a427e20d2896763f71eb7R16

alrevuelta (approved):
lgtm! thanks!

docs/operators/how-to/configure-store.md
```diff
@@ -1,56 +1,57 @@
...
 
-Currently store service nodes use an in-memory key-value store as primary storage with the disk-based database only used for backups.
-Most Waku messages average a size of 1KB - 2KB,
-implying a minimum memory requirement of at least ~250MB
-for a medium capacity store of 100k messages.
-Note, however, that the allowable maximum size for Waku messages is up to 1MB.
-We are working on a disk-only and hybrid store to lower the memory requirement.
-It will soon also be possible to configure store capacity on maximum store size or number of days' history to keep.
\ No newline at end of file
+Currently store service nodes use, by default, a message store backed by an in-disk SQLite database. Most Waku messages average a size of 1KB - 2KB, implying a minimum memory requirement of at least ~250MB
```
jm-clius, line 56: Great to have some numbers here!

docs/operators/how-to/configure-store.md
```diff
@@ -1,56 +1,57 @@
...
+## Configuring the node as a store service node
+
+If the waku store node is enabled (the `--store` option is set to `true`) the node will store historical messages and will be able to serve those messages to the waku store clients.
+
+There is a set of configuration options to customize the waku store protocol's message store. These are the most relevant:
+
+* `--store-message-retention-policy`: This option controls the retention policy i.e., how long certain messages will be persisted. Two different retention policies are supported:
+  + The time retention policy,`time:<duration-in-seconds>` (e.g., `time:14400`)
+  + The capacity retention policy,`capacity:<messages-count>` (e.g, `capacity:25000`)
+  + To disable the retention policy, explicitly, set this option to to `""`, an empty string.
```
jm-clius, line 37: Can we spell out the implications? Something like: `Disabling the retention policy implies that, from nwaku's perspective, messages will either be persisted indefinitely or that retention is managed by a third party outside of the node.`

docs/operators/how-to/configure-store.md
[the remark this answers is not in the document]
jm-clius: True, but it may be worth spelling out in the client section that to configure a client _only_ `store` must explicitly be set to `false`. This is clearly implied, of course, but we should assume that operators will just skip to the section they're interested in and may be surprised that following the client guidelines lead to a node that's also persisting messages.

Not part of this conversation, but wondering if shouldn't rather make the default for `store` `false`. Otherwise many operators upgrading after the release will suddenly start persisting messages. We can warn them, of course, but I think the sensible default for a nwaku node should just be a `relay` node.

jm-clius (commented):
LGTM! Thanks. Apologies for only reviewing after merging. Comments below are not critical at all and can be addressed in future. A good idea to follow [semantic line breaks](https://sembr.org/) in future. :)

PR #1362, opened 2022-11-09
refactor: continue pubsub/content types started in #1352

alrevuelta:
* Resolves [comment1](https://github.com/status-im/nwaku/pull/1352/files#r1016901351) from #1352
* Resolves [comment2](https://github.com/status-im/nwaku/pull/1352/files#r1016897936) from #1352

Summary:
* General refactor using `PubsubTopic` and `ContentTopic` fields.
* Reuse default values for both fields introduced in #1352
* Removes `PubsubTopicString` and its usages.
* Enforce `defaultPubsubTopics` instead of `defaultTopics`

As a general note, which I think @LNSD agrees with me, we should be specific using `pubsub` or `content` instead of just using `topic`. As it may be confusing.

### DIFF
--- a/waku/v2/protocol/waku_message.nim
+++ b/waku/v2/protocol/waku_message.nim
@@ -27,8 +27,8 @@ type
   ContentTopic* = string
 
 const 
-  DefaultPubsubTopic*: PubsubTopic = "/waku/2/default-waku/proto"
-  DefaultContentTopic*: ContentTopic = "/waku/2/default-content/proto"
+  DefaultPubsubTopic*: PubsubTopic = PubsubTopic("/waku/2/default-waku/proto")
+  DefaultContentTopic*: ContentTopic = ContentTopic("/waku/2/default-content/proto")
 
 
 type WakuMessage* = object
--- a/tests/v2/test_message_cache.nim
+++ b/tests/v2/test_message_cache.nim
@@ -10,16 +10,12 @@ import
   ./testlib/common
 
 
-
-type PubsubTopicString = string
-
-type TestMessageCache = MessageCache[(PubsubTopicString, ContentTopic)]
-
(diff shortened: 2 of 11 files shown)


waku/v2/protocol/waku_message.nim
```diff
@@ -27,8 +27,8 @@ type
   ContentTopic* = string
 
 const 
-  DefaultPubsubTopic*: PubsubTopic = "/waku/2/default-waku/proto"
-  DefaultContentTopic*: ContentTopic = "/waku/2/default-content/proto"
+  DefaultPubsubTopic*: PubsubTopic = PubsubTopic("/waku/2/default-waku/proto")
+  DefaultContentTopic*: ContentTopic = ContentTopic("/waku/2/default-content/proto")
```
LNSD, line 31: This is redundant and not necessary, the types are not `distinct string`

LNSD (approved):
LGTM ✅

rymnc (approved):
LGTM!

waku/v2/protocol/waku_message.nim
alrevuelta: yep, thought it was nice for reference, so that people declaring topics they would see this and follow the pattern:

```nim
let x = PubsubTopic("x/y/z/a")
```

instead of `let y = "x/y/z/a"` which I would say its bad practise?

no strong opinion about it, though :)

LNSD: It is redundant since the variable is already type annotated:

```nim
 DefaultPubsubTopic*: PubsubTopic = PubsubTopic("/waku/2/default-waku/proto")
 #~~~~~~~~~~~~~~~~~~~~~~~~~^
````

Issue #1357, opened 2022-11-08
maintenance: failing tests


s1fr0:
## Problem
In the development of `nwaku`, we occasionally observe some CI/Jenkins test failures that are usually not triggered when tests are re-executed. 

This issue aims to collect observed test failures in order to:
- track context in which a failure happened;
- discuss/speculate possible causes that triggered the failure;
- make developers aware if a failure is new (and likely coming from recently changed code) or was previously known.

Everyone is encouraged to edit this issue and add new tests that fail during checks. 
Once the causes of a test failure are identified and addressed, please remove the failing test from this post and comment to this issue by reporting it description, the list of identified causes and link the PR that fixed it.

Any feedback is welcome!

## Test report template
**Action type**: CI/Jenkins
**OS**: linux/macOS
**Test name/version:** (v1/v2)
**nwaku commit/tree:**
**Log excerpt**:
**Causes:** (not identified/speculated/identified)
**TimesSeen:** (times the test was seen)
**Temporary fix/workaround:** (if any)

## Failing tests

**Action type**: CI
**OS**: macOS
**Test name/version:** ["Messages are bridged between Waku v1 and Waku v2"](https://github.com/status-im/nwaku/blob/7bc657b545c1b6eb54d5b70f8a926fb208aad939/tests/v2/test_waku_bridge.nim#L117) (v2)
**nwaku commit/tree:** https://github.com/status-im/nwaku/tree/7bc657b545c1b6eb54d5b70f8a926fb208aad939 
**Log excerpt**:
**TimeSeen**: 4
```
2022-11-08T12:27:30.0160480Z     /Users/runner/work/nwaku/nwaku/tests/v2/test_waku_bridge.nim(152, 49): Check failed: v1Node.protocolState(Waku).queue.items.len == 1
2022-11-08T12:27:30.0161010Z     v1Node.protocolState(Waku).queue.items.len was 0
2022-11-08T12:27:30.0161840Z /Users/runner/work/nwaku/nwaku/vendor/nimbus-build-system/vendor/Nim/lib/system/fatal.nim(53) sysFatal
2022-11-08T12:27:30.0162530Z /Users/runner/work/nwaku/nwaku/vendor/nim-chronos/chronos/asyncfutures2.nim(391) futureContinue
2022-11-08T12:27:30.0162820Z 
2022-11-08T12:27:30.0163040Z     Unhandled exception: index out of bounds, the container is empty [IndexDefect]
[...]

PR #1364, opened 2022-11-09
fix: move stopGanache/uninstallGanache to teardown

alrevuelta:
Looks like there are some problems with macOs tests and RLN. Haven't been able to reproduce the issue, but looking at a ok build and nok build, everything seems to point to `uninstallGanache()`.

NOK build
image

OK build
image

So everything seems to point out to `uninstallGanache()`.

WIP: This PR is debugging the issue.
Ideas:
* using `teardown` might help?

### DIFF
--- a/tests/v2/test_waku_rln_relay_onchain.nim
+++ b/tests/v2/test_waku_rln_relay_onchain.nim
@@ -605,12 +605,13 @@ procSuite "Waku-rln-relay":
     await node.stop()
     await node2.stop()
 
+
   ################################
   ## Terminating/removing Ganache
   ################################
+  teardown:
+    # We stop Ganache daemon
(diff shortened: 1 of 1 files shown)


LNSD:
@alrevuelta Please, check this issue for details on which test cases are failing: https://github.com/status-im/nwaku/issues/1357

From my POV, if you want to help fix the failing test cases, then start with the test case that @s1fr0 identified in that issue. The issue is not with _ganache_ as @s1fr0 has pointed out here (and also in Vac's discord).
 

Regarding the `teardown` usage in tests. I agree that it is needed, but it is necessary to add it in all test cases that need to do a cleanup after the test finishes, not only in this case. This is a pending task that the lack of bandwidth has not let me tackle yet.

PR #1367, opened 2022-11-10
feat(rln-relay): group manager api

staheri14:
This PR is in draft mode and in progress (a lot of clean up and updating the tests) but is open to suggestions and comments.

The major change is to separate the group manager logic from waku-rln-relay part.
Group management is a component whose sole goal is to obtain group-relevant updates (whether statically or by onchain interaction, or even using a p2p protocol).
Then, group-relevant updates are consumed by a `WakuRLNRelay` object which reflects them in its internal state.
To enable this separation, a new data type `GroupManager` with a specific set of methods is defined under a newly introduced module of `group_manager`.  Methods include `start`, `stop`, `register`, `setEventsHandlers`, and `setRegistrationHandler`.
A `WakuRLNRelay` has a member field of `GroupManager` type which can be initialized to any datatype that implements the methods of a `GroupManager`.

Following this new design, all the logic related to the on-chain rln group management is moved to a separate module, `rln_group_manager`, and defined as methods of an `OnChainRlnGroupManager` type that satisfies the `GroupManager` interface.

**Note** tests are not yet updated, but you can make a local build for `wakunode2` and `chat2` with `RLN` flag set to true and they all will be successful.

# Next
Later, we can do a similar refactoring for the static group management.


waku/v2/protocol/waku_rln_relay/rln_types.nim
```diff
@@ -0,0 +1,23 @@
+import
```
staheri14, line 1: This module is created to resolve a circular dependency issue.

staheri14:
cc: @LNSD @s1fr0 @rymnc

apps/chat2/chat2.nim
```diff
@@ -34,11 +34,13 @@ import
...
   ./config_chat2
 
+
 when defined(rln) or defined(rlnzerokit):
   import
     libp2p/protocols/pubsub/rpc/messages,
     libp2p/protocols/pubsub/pubsub,
-    ../../waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils
+    ../../waku/v2/protocol/waku_rln_relay/waku_rln_relay_utils,
+    ../../waku/v2/protocol/waku_rln_relay/rln_types
```
LNSD, line 43: We need to put all RLN modules behind a unique module. Please, check the other protocols (e.g., `protocols/waku_store.nim`, `protocols/waku_filter.nim`, etc.):

NB the "entry point" module is in the `protocols` folder. It imports the internal modules and reexports only those that should be accessible from outside).

With that approach, the internal modules can evolve without affecting/making the imports change in the app code (like here).

waku/v2/protocol/waku_rln_relay/rln_types.nim
LNSD: Can you put all types + logic associated (all `proc`s whose first argument is `RLN`) inside of an `rln.nim` module?

_The code that changes together should be together._

staheri14:
This PR contains some features and refactorings that would be easier to review in their own self-contained smaller PRs, so I'm closing this one and will open individual PRs after that. 

Thank @LNSD for your comments so fat, I'll take them into account in the follow up PRs.

PR #1368, opened 2022-11-10
fix(wakubridge): fix test failure while bridging messages between Waku v1 and v2 (WIP) 

s1fr0:
This PR partially covers the issues collected by the flaky tests umbrella issue #1357.

> ⚠️ WIP


jm-clius:
@s1fr0 with the merging of https://github.com/waku-org/nwaku/pull/1429 perhaps we can close this PR? The former is not necessarily a fix, but if we still encounter the exact same error, we'd have to find another workaround than just changing timing, I believe. :)

PR #1375, opened 2022-11-14
fix(tests): add missing import for tests

alrevuelta:
Due to a missing import, some tests could not be run isolated. The issue can be reproduced as follows, and it applies to all modified tests where the `chronicles` import was added.

```console
$ ./env.sh bash
$ nim c -r ./tests/v2/test_waku_filter.nim

/Users/xxx/Github/nwaku/tests/v2/test_waku_filter.nim(18, 95) template/generic instantiation of `async` from here
/Users/xxx/Github/nwaku/tests/v2/test_waku_filter.nim(20, 30) template/generic instantiation of `new` from here
/Users/xxx/Github/nwaku/waku/v2/node/peer_manager/peer_manager.nim(149, 11) template/generic instantiation of `debug` from here
/Users/xxx/Github/nwaku/vendor/nim-chronicles/chronicles.nim(359, 10) template/generic instantiation of `log` from here
/Users/xxx/Github/nwaku/vendor/nim-chronicles/chronicles.nim(332, 21) Error: undeclared identifier: 'activeChroniclesStream'
candidates (edit distance, scope distance); see '--spellSuggest':
 (13, 7): 'IfDvbrccUpstream' [enumField declared in /Users/xxx/Github/nwaku/vendor/nim-chronos/chronos/transports/osnet.nim(176, 5)]
 (13, 7): 'asyncstream' [module declared in /Users/xxx/Github/nwaku/vendor/nim-chronos/chronos/streams/chunkstream.nim(12, 8)]
 (13, 7): 'asyncstream' [module declared in /Users/xxx/Github/nwaku/vendor/nim-chronos/chronos/transport.nim(10, 17)]
 (13, 7): 'chunkstream' [module declared in /Users/xxx/Github/nwaku/vendor/nim-chronos/chronos/transport.nim(10, 17)]
 (13, 7): 'initStream' [method declared in /Users/xxx/Github/nwaku/vendor/nim-libp2p/libp2p/protocols/secure/secure.nim(63, 8)]
 (13, 7): 'initStream' [method declared in /Users/xxx/Github/nwaku/vendor/nim-libp2p/libp2p/protocols/secure/secure.nim(63, 8)]
 (13, 7): 'initStream' [method declared in /Users/xxx/Github/nwaku/vendor/nim-libp2p/libp2p/stream/connection.nim(75, 8)]
 (13, 7): 'initStream' [method declared in /Users/xxx/Github/nwaku/vendor/nim-libp2p/libp2p/stream/connection.nim(75, 8)]
 (13, 7): 'initStream' [method declared in /Users/xxx/Github/nwaku/vendor/nim-libp2p/libp2p/stream/lpstream.nim(154, 8)]
 (13, 7): 'initStream' [method declared in /Users/xxx/Github/nwaku/vendor/nim-libp2p/libp2p/stream/lpstream.nim(154, 8)]
```


rymnc (approved):
LGTM, thanks!

waku/v2/README.md
```diff
@@ -45,14 +45,19 @@ Setting up a `wakunode2` on the smallest [digital ocean](https://docs.digitaloce
...
-You can also run a specific test (and alter compile options as you want):
+You can also run a specific test.
 ```bash
 # Get a shell with the right environment variables set
 ./env.sh bash
 # Run a specific test
 nim c -r ./tests/v2/test_waku_filter.nim
 ```
 
+And and alter compile options as you want. For example, if you want a less verbose output you can do the following. For more, refer to the [compiler flags](https://nim-lang.org/docs/nimc.html#compiler-usage) and [chronicles documentation](https://github.com/status-im/nim-chronicles#compile-time-configuration).
```
rymnc, line 56:
```suggestion
and alter compile options as you want. For example, if you want a less verbose output you can do the following. For more, refer to the [compiler flags](https://nim-lang.org/docs/nimc.html#compiler-usage) and [chronicles documentation](https://github.com/status-im/nim-chronicles#compile-time-configuration).
```

LNSD (commented):
Changes LGTM ✅

One question: are this changes adding an `UnusedImport` warning to the `test2` compilation logs? If so, then I would rather investigate a better approach that do not introduce those warnings and allows compiling the module individually.

Some context: 

> The amount of warnings we have is so high that is already causing "alert fatigue", and we can miss relevant issues for the PR. We need to address them as a maintenance chore.

LNSD:
> are these changes adding an `UnusedImport` warning to the `test2` compilation logs?

I answer my question: Yes, they do.

https://github.com/status-im/nwaku/actions/runs/3460521398/jobs/5777145106#step:7:44

Then, I would investigate a better approach that does not introduce those warnings and allows compiling the module individually.

jm-clius (approved):
LGTM! Thanks for fixing this, as these individual tests have been broken for a while.

I agree with @LNSD about perhaps finding a way of doing this that does not increase compiler warnings (we've been struggling to stay ahead here due to Nim limitations). Is it possible to use the `{.used.}` pragma (explained [here](https://nim-lang.org/docs/manual.html#pragmas-used-pragma))  on the specific offending imports?
If there's no easy and clean way of avoiding the warnings, I think we should track it in a separate issue and come up with a strategy, but that would be subsequent to fixing the tests.

alrevuelta:
@LNSD good catch. A simple approach could be to use [this](https://forum.nim-lang.org/t/5461) but I wouldn't like to have the test code full of these off/on statements. Another approach is to find the root cause of the issue and fix it properly, but have the feeling this will take some time. As I see it this has a really low prio, so I would manually disable the warnings as @jm-clius kind of suggested.

LNSD:
> so I would manually disable the warnings as @jm-clius kind of suggested.

What do you mean by "manually disable" the warnings? 👀

alrevuelta:
> What do you mean by "manually disable" the warnings? 👀

Disregard my comment. I was expecting this to work, but it does not. 

```nim
{. warning[UnusedImport]:off .}

import "sequtils"

{. warning[UnusedImport]:on .}

import "tables"

echo "hi there"
```

But: 
```
Warning: imported and not used: 'sequtils' [UnusedImport]
Warning: imported and not used: 'tables' [UnusedImport]
```

I was ofc expecting to warn only about `tables`.

alrevuelta:
@LNSD will give it a shot before the end of the week. if it turns out to be too much, i would merge the PR and open an issue to track the issue. I think its better a "unused import" than a compile error. wdyt?

jm-clius:
Let's merge this. The issue here is _incorrect_ compiler message of `unused import` for an import that's clearly used. First step is to fix the tests, which indeed affects development (I ran into this as well). Next step is to see if there is a solution around this limitation of the Nim compiler.

alrevuelta:
@jm-clius great! tracking it here: https://github.com/status-im/nwaku/issues/1398

PR #1382, opened 2022-11-15
fix: store clients with `--store=false` does not install Store Client JSON-RPC API handlers

### LINKED ISSUE #1381: bug: store clients with `--store=false` does not install Store Client JSON-RPC API handlers
### Problem

nwaku nodes that does not have `store` enabled does not install the correct JSON-RPC handlers for store clients. This is related to recent config changes that now requires only `store` service nodes (not clients) to enable the protocol.

### Impact

Store clients are now required to also act as store services (`--store=true`) to install and expose the store client JSON-RPC API functions.

### Expected behavior

Nwaku nodes should always install store client API methods, regardless of whether `--store` is enabled or not.

### nwaku version/commit hash
`v0.13.0`

LNSD:
This PR resolves #1381 

- [x] The JSON RPC handlers for filter and store are installed if the corresponding `storenode` and `filternode` configuration parameters are provided. 

This fix affects to `wakunode2` and `wakubridge` apps.

### DIFF
--- a/apps/wakubridge/wakubridge.nim
+++ b/apps/wakubridge/wakubridge.nim
@@ -11,10 +11,12 @@ import
   stew/shims/net as stewNet, json_rpc/rpcserver,
   libp2p/errors,
   libp2p/peerstore,
-  # Waku v1 imports
-  eth/[keys, p2p], eth/common/utils,
+  eth/[keys, p2p], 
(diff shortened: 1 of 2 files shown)


jm-clius (approved):
Thanks! Should do the trick.

PR #1383, opened 2022-11-15
refactor: reuse nim-libp2p peerstore + move peermanager logic

### LINKED ISSUE #622: Waku Peer Store: Integrated libp2p peer store

alrevuelta:
Closes #622

- [x] Reuse `nim-libp2p` [peerstore](https://github.com/status-im/nim-libp2p/blob/unstable/libp2p/peerstore.nim) and remove duplicated fields.
- [x] Move all logic related to accessing peers, retrieving peers with specific conditions (like supported protocol) from `peer_manager` to the peer store.
- [x] Add new cli flag to configure the number of peers stored by the nim-libp2p peerstore.
- [x] Add new field tracking peer origin (discv5, static, dns). Unused.
- [x] Add test coverage in the peermanager, since its being modified and some functions were not unit tested.

### DIFF
--- a/tests/v2/test_peer_manager.nim
+++ b/tests/v2/test_peer_manager.nim
@@ -35,7 +35,7 @@ procSuite "Peer Manager":
       nodeKey2 = crypto.PrivateKey.random(Secp256k1, rng[])[]
       node2 = WakuNode.new(nodeKey2, ValidIpAddress.init("0.0.0.0"), Port(60802))
       peerInfo2 = node2.switch.peerInfo
-    
+
     await allFutures([node1.start(), node2.start()])
 
     await node1.mountRelay()
@@ -48,7 +48,7 @@ procSuite "Peer Manager":
     check:
       conn.activity
       conn.peerId == peerInfo2.peerId
-    
+
     # Check that node2 is being managed in node1
     check:
       node1.peerManager.peerStore.peers().anyIt(it.peerId == peerInfo2.peerId)
@@ -56,9 +56,9 @@ procSuite "Peer Manager":
     # Check connectedness
     check:
       node1.peerManager.peerStore.connectedness(peerInfo2.peerId) == Connectedness.Connected
-    
+
     await allFutures([node1.stop(), node2.stop()])
-  
+
   asyncTest "Dialing fails gracefully":
     let
       nodeKey1 = crypto.PrivateKey.random(Secp256k1, rng[])[]
@@ -66,7 +66,7 @@ procSuite "Peer Manager":
       nodeKey2 = crypto.PrivateKey.random(Secp256k1, rng[])[]
       node2 = WakuNode.new(nodeKey2, ValidIpAddress.init("0.0.0.0"), Port(60812))
       peerInfo2 = node2.switch.peerInfo
-    
+
     await node1.start()
     # Purposefully don't start node2
 
@@ -79,7 +79,7 @@ procSuite "Peer Manager":
     # Check connection failed gracefully
     check:
       connOpt.isNone()
-    
+
     await node1.stop()
 
   asyncTest "Adding, selecting and filtering peers work":
@@ -98,7 +98,7 @@ procSuite "Peer Manager":
       storeLoc = MultiAddress.init("/ip4/127.0.0.3/tcp/4").tryGet()
       storeKey = crypto.PrivateKey.random(ECDSA, rng[]).get()
       storePeer = PeerInfo.new(storeKey, @[storeLoc])
-    
+
     await node.start()
 
     await node.mountFilterClient()
@@ -106,7 +106,7 @@ procSuite "Peer Manager":
     node.mountStoreClient()
 
     node.wakuSwap.setPeer(swapPeer.toRemotePeerInfo())
-    
+
     node.setStorePeer(storePeer.toRemotePeerInfo())
     node.setFilterPeer(filterPeer.toRemotePeerInfo())
 
@@ -122,9 +122,9 @@ procSuite "Peer Manager":
       node.peerManager.peerStore.peers(WakuStoreCodec).allIt(it.peerId == storePeer.peerId and
                                                              it.addrs.contains(storeLoc) and
                                                              it.protos.contains(WakuStoreCodec))
-    
+
     await node.stop()
-  
+
 
   asyncTest "Peer manager keeps track of connections":
     let
@@ -133,7 +133,7 @@ procSuite "Peer Manager":
       nodeKey2 = crypto.PrivateKey.random(Secp256k1, rng[])[]
       node2 = WakuNode.new(nodeKey2, ValidIpAddress.init("0.0.0.0"), Port(60832))
       peerInfo2 = node2.switch.peerInfo
-    
+
     await node1.start()
 
     await node1.mountRelay()
@@ -164,7 +164,7 @@ procSuite "Peer Manager":
     check:
       # Not currently connected to node2, but had recent, successful connection.
       node1.peerManager.peerStore.connectedness(peerInfo2.peerId) == CanConnect
-    
+
     await node2.stop()
 
   asyncTest "Peer manager can use persistent storage and survive restarts":
@@ -174,9 +174,9 @@ procSuite "Peer Manager":
       nodeKey1 = crypto.PrivateKey.random(Secp256k1, rng[])[]
       node1 = WakuNode.new(nodeKey1, ValidIpAddress.init("0.0.0.0"), Port(60840), peerStorage = storage)
       nodeKey2 = crypto.PrivateKey.random(Secp256k1, rng[])[]
-      node2 = WakuNode.new(nodeKey2, ValidIpAddress.init("0.0.0.0"), Port(60842)) 
+      node2 = WakuNode.new(nodeKey2, ValidIpAddress.init("0.0.0.0"), Port(60842))
       peerInfo2 = node2.switch.peerInfo
-    
+
     await node1.start()
     await node2.start()
 
@@ -194,7 +194,7 @@ procSuite "Peer Manager":
     let
       nodeKey3 = crypto.PrivateKey.random(Secp256k1, rng[])[]
       node3 = WakuNode.new(nodeKey3, ValidIpAddress.init("0.0.0.0"), Port(60844), peerStorage = storage)
-    
+
     await node3.start()
     check:
       # Node2 has been loaded after "restart", but we have not yet reconnected
@@ -203,15 +203,15 @@ procSuite "Peer Manager":
       node3.peerManager.peerStore.connectedness(peerInfo2.peerId) == NotConnected
 
     await node3.mountRelay()  # This should trigger a reconnect
-    
+
     check:
       # Reconnected to node2 after "restart"
       node3.peerManager.peerStore.peers().len == 1
       node3.peerManager.peerStore.peers().anyIt(it.peerId == peerInfo2.peerId)
       node3.peerManager.peerStore.connectedness(peerInfo2.peerId) == Connected
-    
+
     await allFutures([node1.stop(), node2.stop(), node3.stop()])
-  
+
   # TODO: nwaku/issues/1377
   xasyncTest "Peer manager support multiple protocol IDs when reconnecting to peers":
     let
@@ -224,7 +224,7 @@ procSuite "Peer Manager":
       peerInfo2 = node2.switch.peerInfo
       betaCodec = "/vac/waku/relay/2.0.0-beta2"
       stableCodec = "/vac/waku/relay/2.0.0"
-    
+
     await node1.start()
     await node2.start()
 
@@ -245,7 +245,7 @@ procSuite "Peer Manager":
     let
       nodeKey3 = crypto.PrivateKey.random(Secp256k1, rng[])[]
       node3 = WakuNode.new(nodeKey3, ValidIpAddress.init("0.0.0.0"), Port(60854), peerStorage = storage)
-    
+
     await node3.mountRelay()
     node3.wakuRelay.codec = stableCodec
     check:
@@ -257,7 +257,7 @@ procSuite "Peer Manager":
       node3.peerManager.peerStore.peers().anyIt(it.peerId == peerInfo2.peerId)
       node3.peerManager.peerStore.peers().anyIt(it.protos.contains(betaCodec))
       node3.peerManager.peerStore.connectedness(peerInfo2.peerId) == NotConnected
-    
+
     await node3.start() # This should trigger a reconnect
 
     check:
@@ -267,5 +267,5 @@ procSuite "Peer Manager":
       node3.peerManager.peerStore.peers().anyIt(it.protos.contains(betaCodec))
       node3.peerManager.peerStore.peers().anyIt(it.protos.contains(stableCodec))
       node3.peerManager.peerStore.connectedness(peerInfo2.peerId) == Connected
-    
+
     await allFutures([node1.stop(), node2.stop(), node3.stop()])
--- a/tests/v2/test_peer_manager.nim
+++ b/tests/v2/test_peer_manager.nim
@@ -24,7 +24,8 @@ import
   ../../waku/v2/protocol/waku_store,
   ../../waku/v2/protocol/waku_filter,
   ../../waku/v2/protocol/waku_swap/waku_swap,
-  ../test_helpers
+  ../test_helpers,
+  ./testlib/testutils
 
 procSuite "Peer Manager":
   asyncTest "Peer dialing works":
@@ -50,11 +51,11 @@ procSuite "Peer Manager":
     
     # Check that node2 is being managed in node1
     check:
-      node1.peerManager.peers().anyIt(it.peerId == peerInfo2.peerId)
+      node1.peerManager.peerStore.peers().anyIt(it.peerId == peerInfo2.peerId)
 
     # Check connectedness
     check:
-      node1.peerManager.connectedness(peerInfo2.peerId) == Connectedness.Connected
+      node1.peerManager.peerStore.connectedness(peerInfo2.peerId) == Connectedness.Connected
     
     await allFutures([node1.stop(), node2.stop()])
   
@@ -111,16 +112,16 @@ procSuite "Peer Manager":
 
     # Check peers were successfully added to peer manager
     check:
-      node.peerManager.peers().len == 3
-      node.peerManager.peers(WakuFilterCodec).allIt(it.peerId == filterPeer.peerId and
-                                                    it.addrs.contains(filterLoc) and
-                                                    it.protos.contains(WakuFilterCodec))
-      node.peerManager.peers(WakuSwapCodec).allIt(it.peerId == swapPeer.peerId and
-                                                  it.addrs.contains(swapLoc) and
-                                                  it.protos.contains(WakuSwapCodec))
-      node.peerManager.peers(WakuStoreCodec).allIt(it.peerId == storePeer.peerId and
-                                                   it.addrs.contains(storeLoc) and
-                                                   it.protos.contains(WakuStoreCodec))
+      node.peerManager.peerStore.peers().len == 3
+      node.peerManager.peerStore.peers(WakuFilterCodec).allIt(it.peerId == filterPeer.peerId and
+                                                              it.addrs.contains(filterLoc) and
+                                                              it.protos.contains(WakuFilterCodec))
+      node.peerManager.peerStore.peers(WakuSwapCodec).allIt(it.peerId == swapPeer.peerId and
+                                                            it.addrs.contains(swapLoc) and
+                                                            it.protos.contains(WakuSwapCodec))
+      node.peerManager.peerStore.peers(WakuStoreCodec).allIt(it.peerId == storePeer.peerId and
+                                                             it.addrs.contains(storeLoc) and
+                                                             it.protos.contains(WakuStoreCodec))
     
     await node.stop()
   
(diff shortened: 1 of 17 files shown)


waku.nimble
```diff
@@ -34,7 +34,7 @@ proc buildBinary(name: string, srcDir = "./", params = "", lang = "c") =
     extra_params &= " " & paramStr(i)
   exec "nim " & lang & " --out:build/" & name & " " & extra_params & " " & srcDir & name & ".nim"
 
-proc test(name: string, params = "-d:chronicles_log_level=DEBUG", lang = "c") =
+proc test(name: string, params = "-d:chronicles_log_level=WARN", lang = "c") =
```
jm-clius, line 37: I take it this is temporary? :)

waku/v2/node/peer_manager/peer_manager.nim
```diff
@@ -203,36 +167,27 @@ proc addPeer*(pm: PeerManager, remotePeerInfo: RemotePeerInfo, proto: string) =
...
-
-  if peers.len >= 1:
-     # TODO: proper heuristic here that compares peer scores and selects "best" one. For now the first peer for the given protocol is returned
-    let peerStored = peers[0]
-
-    return some(peerStored.toRemotePeerInfo())
-  else:
-    return none(RemotePeerInfo)
-
 proc reconnectPeers*(pm: PeerManager,
```
jm-clius, line 191: I'm unsure if we should keep this reconnection logic, at least in the medium term: this was added before discovery methods, mainly to reconnect fleet nodes to each other after a restart. The main difference between the previous peer store and the underlying libp2p one, is that the previous peer store _only_ contained peers that we initiated a connection to, whereas the libp2p peer store also contains all nodes connecting to us. This implies that we only attempted to reconnect to peers that we explicitly connected to before (e.g. using the `connectToNodes` API call). Because GossipSub has a mandatory backoff time before GRAFTing a PRUNEd connection (implicitly then reconnecting after a restart), we had to wait before attempting to reconnect to each node. IIRC this can block for quite a while when the node starts (and maybe much more so now that _all_ nodes will be attempted with reconnection). Now that nodes generally use DNS discovery on boot to reconnect to nodes, perhaps the whole idea of peer persistence and reconnection should be rethought?

waku/v2/node/peer_manager/peer_manager.nim
```diff
@@ -242,15 +197,15 @@ proc reconnectPeers*(pm: PeerManager,
   
   debug "Reconnecting peers", proto=proto
   
-  for storedInfo in pm.peers(protocolMatcher):
+  for storedInfo in pm.peerStore.peers(protocolMatcher):
     # Check if peer is reachable.
-    if pm.peerStore.connectionBook[storedInfo.peerId] == CannotConnect:
+    if pm.peerStore[ConnectionBook][storedInfo.peerId] == CannotConnect: #TODO what if it doesnt exist?
```
jm-clius, line 202: I'm also curious about this TODO. :D

waku/v2/node/peer_manager/peer_manager.nim
```diff
@@ -29,7 +29,7 @@ logScope:
 type
   PeerManager* = ref object of RootObj
     switch*: Switch
```
jm-clius, line 31: Looking forward to having the `switch` removed here in future (not this PR, of course :))

waku/v2/node/peer_manager/peer_manager.nim
```diff
@@ -203,36 +167,27 @@ proc addPeer*(pm: PeerManager, remotePeerInfo: RemotePeerInfo, proto: string) =
...
   debug "Adding peer to manager", peerId = remotePeerInfo.peerId, addr = remotePeerInfo.addrs[0], proto = proto
-  
+
   # ...known addresses
   for multiaddr in remotePeerInfo.addrs:
-    pm.peerStore.addressBook.add(remotePeerInfo.peerId, multiaddr)
-  
+    # TODO: unsure why this does not work. understand and fix
+    #pm.peerStore[AddressBook][remotePeerInfo.peerId].add(multiaddr)
+    pm.peerStore[AddressBook][remotePeerInfo.peerId] = pm.peerStore[AddressBook][remotePeerInfo.peerId] & multiaddr
```
jm-clius, line 175: Wouldn't we now be adding peers that already exist in the peerStore with the same information? (Since we only add peers that we dial currently.)

waku/v2/node/peer_manager/peer_manager.nim
Menduist: default(T), which is debatable
```nim
if storedInfo.peerId in pm.peerStore[ConnectionBook] and pm.peerStore[ConnectionBook][storedInfo.peerId] == CannotConnect:

# or a pattern I start to like, even though it's ugly, {} returning an option
if pm.peerStore[ConnectionBook]{storedInfo.peerId} == some(CannotConnect)
```

arnetheduck:
how is application-specific per-peer data stored when using libp2p-based peer manager?

Menduist:
See the new peer store that enables to create custom books with custom types:
https://github.com/status-im/nim-libp2p/blob/unstable/libp2p/peerstore.nim#L11-L23

waku/v2/node/peer_manager/peer_manager.nim
alrevuelta: Good point. If that's fine I will just change what is needed to not break anything and leave this for another PR.

> perhaps the whole idea of peer persistence and reconnection should be rethought?

Yep. Regarding reconnections, I think `reconnectPeers` can be rescoped to be `runPeeringStrategy` (not happy with the name) with the following functionalities:
* Infinite loop.
* If connectedPeers is below a threshold, try to connect to new peers, from the peerstore.
* It respects the backoff (as its done)
* Some criteria TBD can be added on which peers to select, like score, keeping a given inbound/outbound ratio, keeping a store/relay/etc nodes ratio, etc.
* Forces disconnections if needed.
* Its totally agnostic of peer discovery.
* And does not add peers to the peer store as done now: `pm.addPeer(remotePeerInfo, proto)`

Beyond `runPeeringStrategy`, we will have `runPrunningStrategy` that will remove peers from the peerstore and `runDiscoveryStrategy` that will populate the peerstore with new peers. Names TBD.

waku/v2/node/peer_manager/peer_manager.nim
alrevuelta: I love snippets that can be run `nim r file.nim`. I'm surprised with the result, a bit dangerous isn't it?

```nim
import libp2p/peerstore
import libp2p/peerid
let myPeerStore = PeerStore.new(capacity = 5)

type ConnectionBook = ref object of PeerBook[string]

var p1, p2: PeerId
discard p1.init("peer1")
discard p2.init("peer2")

myPeerStore[ConnectionBook][p1] = "Connected"

echo myPeerStore[ConnectionBook][p1]
echo myPeerStore[ConnectionBook][p2]

#Output:
#Connected
#Connected
```

alrevuelta (commented):
thanks @jm-clius really appreciate the comments. left some answers :)

waku/v2/node/peer_manager/peer_manager.nim
alrevuelta: > (note that there is a minor bug with the latter see that i plan to fix).

https://github.com/status-im/nim-libp2p/pull/804

waku/v2/node/peer_manager/peer_manager.nim
jm-clius: Right, from the perspective of connection management it does make sense to have a `switch` as dialer. That said, I'm still leaning towards separating these two paradigms: a peer manager that operates on a peer store (and provides, e.g. API for discovery methods to add peers) and a connection manager that reads from said peer store and uses the `switch` as dialer to maintain certain connectivity goals. If not physically separated into separate modules it may still makes sense to have these logically separate?

Would `PeerPool` be roughly equivalent to what I described as the peer manager above?

alrevuelta: > a peer manager that operates on a peer store (and provides, e.g. API for discovery methods to add peers)

What I had in mind is that the methods for adding peers are part of the PeerStore and not the PeerManager. The source is specified when adding the peer, and can be bootstrap/discv5 or other.

Will raise this topic in the meeting we have pending, to gather everyone's input.

> Would PeerPool be roughly equivalent to what I described as the peer manager above?

* nimbus: PeerPool is similar to our peerstore, but doesn't maintain any connectivity goals. It allows to for example [addPeer](https://github.com/status-im/nimbus-eth2/blob/stable/beacon_chain/networking/peer_pool.nim#L462) but has no knowledge of discovery. Connectivity goals are handled in eth2_network.nim.
* nim-eth: PeerPool both stores peers and connects to them, for example, [trying to connect to more peers](https://github.com/status-im/nim-eth/blob/master/eth/p2p/peer_pool.nim#L148) if below minPeers. [node disvoery](https://github.com/status-im/nim-eth/blob/master/eth/p2p/peer_pool.nim#L103) is inside the peer pool.

tests/v2/test_peer_store_extended.nim
```diff
@@ -0,0 +1,268 @@
...
+  ./testlib/testutils
+
+
+suite "Extended nim-libp2p Peer Store":
+  # Valid peerId missing the last digit. Useful for creating new peerIds
+  # basePeerId & "1"
+  # basePeerId & "2"
+  let basePeerId = "QmeuZJbXrszW2jdT7GdduSjQskPU3S7vvGWKtKgDfkDvW"
+  
+  # Given
```
LNSD, line 23: _Given_ is for the test cases. It is optional for the test cases like "setup" and "cleanup". On the other hand, "when" and "then" are mandatory. 

So, "Given" is not intended to be added globally. For that, you have _Setup_. And it is usually for doing things like what you are doing here: _configuring shared fixtures_.

tests/v2/test_peer_store_extended.nim
```diff
@@ -0,0 +1,268 @@
...
+
+    # Then
+    check:
+      remotePeerInfo1.peerId == p1
+      remotePeerInfo1.addrs == @[MultiAddress.init("/ip4/127.0.0.1/tcp/1").tryGet()]
+      remotePeerInfo1.protocols == @["/vac/waku/relay/2.0.0-beta1", "/vac/waku/store/2.0.0"]
+
+  test "connectedness() returns the connection status of a given PeerId":
+    # Then
+    check:
```
LNSD, line 214: There's no "when" clause here, like in other test cases.

tests/v2/test_peer_store_extended.nim
```diff
@@ -0,0 +1,268 @@
...
+      pMatcherStorePeers.filterIt(it.peerId == p5)[0].protos == @["/vac/waku/swap/2.0.0", "/vac/waku/store/2.0.0-beta2"]
+
+    # When
+    let pMatcherSwapPeers = peerStore.peers(protocolMatcher("/vac/waku/swap/2.0.0"))
+
+    # Then
+    check:
+      pMatcherSwapPeers.len == 1
+      pMatcherSwapPeers.anyIt(it.peerId == p5)
+      pMatcherSwapPeers[0].protos == @["/vac/waku/swap/2.0.0", "/vac/waku/store/2.0.0-beta2"]
```
LNSD, line 197: The order should always be: "Given" -> "When" -> "Then".

All the checks should happen at the end of the test case under the _Then_ clause. This makes tests easier to maintain, and it will avoid chain reactions.

If there are any conditions _sine qua non_ that you need to assert within the _When_ block, then you must use `require`, which will fail the test case immediately. No `check` assertions must happen within a _When_ block.

tests/v2/test_peer_manager.nim
```diff
@@ -24,7 +25,8 @@ import
   ../../waku/v2/protocol/waku_store,
   ../../waku/v2/protocol/waku_filter,
   ../../waku/v2/protocol/waku_swap/waku_swap,
-  ../test_helpers
+  ../test_helpers,
+  ./testlib/testutils
```
LNSD, line 29: If you don't have any `xtest` or `xsuite`, this will raise an "UnusedImport" warning.

waku/v2/node/jsonrpc/admin_api.nim
```diff
@@ -64,34 +64,38 @@ proc installAdminApiHandlers*(node: WakuNode, rpcsrv: RpcServer) =
 
     if not node.wakuRelay.isNil:
       # Map managed peers to WakuPeers and add to return list
-      wPeers.insert(node.peerManager.peers(WakuRelayCodec)
+      wPeers.insert(node.peerManager.peerStore
+                                    .peers(WakuRelayCodec)
                                     .mapIt(WakuPeer(multiaddr: constructMultiaddrStr(toSeq(it.addrs.items)[0], it.peerId),
                                                     protocol: WakuRelayCodec,
-                                                    connected: node.peerManager.connectedness(it.peerId) == Connectedness.Connected)),
+                                                    connected: it.connection == Connectedness.Connected)),
```
LNSD, line 71: This should be `it.connectedness`, given that you are comparing it with an enum named `Connectedness`.

waku/v2/node/peer_manager/peer_manager.nim
```diff
@@ -99,49 +94,55 @@ proc loadFromStorage(pm: PeerManager) =
...
-    pm.peerStore.connectionBook[peerId] = CanConnect
+    pm.peerStore[ConnectionBook][peerId] = CanConnect
     if not pm.storage.isNil:
       pm.storage.insertOrReplace(peerId, pm.peerStore.get(peerId), CanConnect, getTime().toUnix)
     return
 
 proc new*(T: type PeerManager, switch: Switch, storage: PeerStorage = nil): PeerManager =
   let pm = PeerManager(switch: switch,
-                       peerStore: WakuPeerStore.new(),
+                       peerStore: switch.peerStore,
```
LNSD, line 140: `peerStorage` and `peerStore` names are confusing. Is there any plan to rename any of them?

waku/v2/node/peer_manager/waku_peer_store.nim
```diff
@@ -4,12 +4,17 @@ else:
...
+  std/[tables, sequtils, sets, options],
   libp2p/builders,
   libp2p/peerstore
 
+import
+  ../../utils/peers
+
 export peerstore, builders
 
+# TODO rename to peer_store_extended to emphasize its a nimlibp2 extension
```
LNSD, line 16: Please don't. Look for a better name or keep the same name.

waku/v2/node/peer_manager/waku_peer_store.nim
```diff
@@ -20,74 +25,107 @@ type
...
+    Discv5,
+    Static,
+    Dns
+
+  # Keeps track of the Connectedness state of a peer
   ConnectionBook* = ref object of PeerBook[Connectedness]
 
-  DisconnectBook* = ref object of PeerBook[int64] # Keeps track of when peers were disconnected in Unix timestamps
+  # Keeps track of when peers were disconnected in Unix timestamps
+  DisconnectBook* = ref object of PeerBook[int64]
```
LNSD, line 39: This clearly demonstrates that we need a common time module for timestamps. We should use the same timestamp format.

cc @jm-clius

waku/v2/node/peer_manager/waku_peer_store.nim
```diff
@@ -20,74 +25,107 @@ type
...
+  return peerStore[ConnectionBook].book.getOrDefault(peerId, NotConnected)
+
+proc hasPeer*(peerStore: PeerStore, peerId: PeerID, proto: string): bool =
+  # Returns `true` if peer is included in manager for the specified protocol
+  # TODO: What if peer does not exist in the peerStore?
+  peerStore.get(peerId).protos.contains(proto)
+
+proc hasPeers*(peerStore: PeerStore, proto: string): bool =
+  # Returns `true` if the peerstore has any peer for the specified protocol
+  toSeq(peerStore[ProtoBook].book.values()).anyIt(it.anyIt(it == proto))
```
LNSD, line 115: These two methods differ just with an "s" and have completely different behavior. Please consider renaming one or merging both of the into a single one and allow multiple input types using the `when` statement.

waku/v2/node/peer_manager/waku_peer_store.nim
```diff
@@ -20,74 +25,107 @@ type
...
 
   return allKeys.mapIt(peerStore.get(it))
+
+proc peers*(peerStore: PeerStore, proto: string): seq[StoredInfo] =
+  # Return the known info for all peers registered on the specified protocol
+  peerStore.peers.filterIt(it.protos.contains(proto))
+
+proc peers*(peerStore: PeerStore, protocolMatcher: Matcher): seq[StoredInfo] =
+  # Return the known info for all peers matching the provided protocolMatcher
+  peerStore.peers.filterIt(it.protos.anyIt(protocolMatcher(it)))
```
LNSD, line 94: This proc's name is a "noun". We use nouns for accessing properties (e.g., a property that we did not make public because we don't want to allow modifying it) or "dynamic properties" (virtual properties that are derived from other properties). These are functions as they have parameters. Please, consider renaming the methods to something like `getPeersByProtocol` or something similar.

LNSD (changes requested):
Please check my comments.

Looking forward to understanding what your plan for the peer manager is.

waku/v2/node/peer_manager/waku_peer_store.nim
alrevuelta: +1 noted.

waku/v2/node/peer_manager/waku_peer_store.nim
alrevuelta: +1 `getPeersByProtocol`. this proc already existed, so just reused the name. will followup with a pr with renamings once the heavy stuff is done.

waku/v2/node/peer_manager/waku_peer_store.nim
alrevuelta: i was just moving this procs, but yes. notes.

LNSD (approved):

waku/v2/node/peer_manager/peer_manager.nim
jm-clius: Ok, makes sense to me! We can discuss functionality in more detail.

waku/v2/node/peer_manager/peer_manager.nim
jm-clius: Ok, thanks for this answer! Makes sense. Peers that were "hardcoded" for certain protocols (e.g. `store`, `filter`, etc.) should be the preferred (or only?) peers for those protocols, but perhaps in future we can verify that these peers indeed support the statically-supplied protocols before adding them via `identify`?

waku/v2/node/peer_manager/peer_manager.nim
jm-clius: > Good point. If that's fine I will just change what is needed to not break anything and leave this for another PR.

~Absolutely. I think, though, this PR will change the current behaviour, unless I'm missing something. Previously on startup a peer would reconnect _only_ to peers that were statically added or connected via DNS discovery. Now _all_ peers will both be persisted and a reconnection attempted.~

Changed my thinking halfway through typing this: the peers we _persist_ will still only be the ones added via `addPeers()` (i.e. the ones we ourselves statically configured, discovered via DNS, etc) and not include all peers in the peer store (which would include incoming connections too)? My worry was that we'll now keep persisting all these peers and reattempt connecting to all of them forever (since the peer storage is never truncated/pruned as it stands).

waku/v2/node/peer_manager/waku_peer_store.nim
jm-clius: Could we make this now a `Timestamp` (which is currently just an alias for `int64`) to save some refactoring in future.

jm-clius (approved):
LGTM! Thanks. May be worth doing some log analyses after this gets deployed to the fleets (should now be auto-deployed to both `wakuv2.test` and `status.test`) to double check that e.g. reconnection works as before.

waku/v2/node/peer_manager/peer_manager.nim
alrevuelta: > we'll now keep persisting all these peers and reattempt connecting to all of them forever

As I can see in [kibana](https://kibana.infra.status.im/app/discover#/?_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:now-7d,to:now))&_a=(columns:!(message,severity),filters:!(('$state':(store:appState),meta:(alias:!n,disabled:!f,index:ffcc22b0-0116-11ed-9719-cdd3b483481c,key:fleet,negate:!f,params:!(status.prod),type:phrases),query:(bool:(minimum_should_match:1,should:!((match_phrase:(fleet:status.prod)))))),('$state':(store:appState),meta:(alias:!n,disabled:!f,index:ffcc22b0-0116-11ed-9719-cdd3b483481c,key:severity,negate:!f,params:(query:6),type:phrase),query:(match_phrase:(severity:6))),('$state':(store:appState),meta:(alias:!n,disabled:!f,index:ffcc22b0-0116-11ed-9719-cdd3b483481c,key:logsource,negate:!f,params:(query:node-01.do-ams3.status.prod),type:phrase),query:(match_phrase:(logsource:node-01.do-ams3.status.prod)))),index:ffcc22b0-0116-11ed-9719-cdd3b483481c,interval:auto,query:(language:kuery,query:'%22Reconnecting%20peers%22%20OR%20%22Dialing%20peer%22'),sort:!(!('@timestamp',desc)))) we are already attempting to connect to all peers forever, so after checking it I would say the behaviour is the same. But I have this in mind and planning to further investigate it, also related to the issue in the disv5 loop that we discussed. We are constanly discovering peers, connecting to them. But since we don't have that many peers, we are always trying to connect to the same 10-20 nodes.

> which would include incoming connections too

Yes! Will address all this in another PR, with a new field indicating `Direction` in/out. Then that would be used to complete https://github.com/waku-org/nwaku/issues/1206 (showing number of in/out connections)

PR #1384, opened 2022-11-15
tools: move readmes into their respective folders

jakubgs:
Makes it easier to link to just one folder in the repo for documentation purposes without having to also link the parent folder.

Also added a meta file linking to the tools. Not sure what the others ones do.


alrevuelta (approved):
lgtm! thanks for the fix!

Issue #1380, opened 2022-11-15
feat(RLN): load zerokit RLN as static library


s1fr0:
### Problem
At the moment zerokit RLN is loaded by the waku-rln-relay module as dynamic library. 
This creates path issues similar to https://github.com/status-im/nwaku/issues/1372 (i.e. wakunode has to be loaded from root folder). Furthermore, from a Waku product perspective, [it is required](https://github.com/status-im/nwaku/issues/1372#issuecomment-1313921001) that zerokit RLN is loaded as static library instead.

### Acceptance criteria

- [x] Support compiling Zerokit RLN in nwaku as a static library
- [x] Add the required compilation flags to the nwaku build scripts (e.g., Makefile)
- [x] Update the Jenkins CI and GH Actions pipelines accordingly
- [x] Update the Dockerfile accordingly (cc @jakubgs)

LNSD:
Added the issue tentatively to the next release, v0.14.0.

PR #1385, opened 2022-11-15
feat(rln): use zerokit as static lib

s1fr0:
This PR addresses the first, second, forth item of https://github.com/status-im/nwaku/issues/1380

The RLN flags management in the makefile has been refactored so that rln flags (`rln` and `onchain_rln`) together with rlnlib compilation and library linking are processed only when `rlnlib` target is compiled. 

This avoid to pass unnecessary rln flags to e.g. v1 targets when targets are **not** compiled in parallel, and ensures the RLN library to be eventually linked only after it has been effectively compiled. See https://github.com/status-im/nwaku/pull/1385#issuecomment-1317816904 for more details/limitations of current implementation.

This branch has been deployed on test fleet and the chat2 app with RLN enabled worked as expected.

Minor: I added a yellow warning that tells the user that onchain and RLN tests will not be compiled when only the flag `ONCHAIN_RLN` is set to true and `RLN` remains unset.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -60,6 +60,5 @@ CMD ["--help"]
 
 FROM prod AS experimental
 
-# If RLN enabled, copy the librln artifacts
-COPY --from=nim-build /app/vendor/zerokit/target/release/librln.so vendor/zerokit/target/release/librln.so
+# If RLN enabled, copy RLN resources (WASM ZK circuit, proving and verification keys) used by tests
 COPY --from=nim-build /app/vendor/zerokit/rln/resources/ vendor/zerokit/rln/resources/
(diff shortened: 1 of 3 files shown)


LNSD (approved):
LGreatTM ✅ 🧙🏼‍♂️

LNSD:
Taking over this PR, as agreed with @s1fr0

rymnc (approved):
LGTM

.github/workflows/test2.yml
```diff
@@ -61,10 +61,7 @@ jobs:
           key: ${{ runner.os }}-${{ matrix.env.NPROC }}-nim-${{ hashFiles('.gitmodules') }}
 
       - name: Update dependencies
-        run: make -j${NPROC} V=1 update
-
-      - name: Build V2 binaries
```
rymnc, line 66: Shouldn't we always build the V2 binaries because some change in the waku/v2/ dir could break code which isn't part of the testing suite?

LNSD:
ℹ️ This PR now depends on #1437

rymnc:
Awesome :)

PR #1386, opened 2022-11-16
chore(rln-relay): barrel imports

rymnc:
This PR introduces barrel imports for waku-rln-relay, resolving part of #1376.

Out of scope for this PR -
- Resolving cyclic dependencies by updating rln-relay imports
- Mounting procedure update in waku_node

PR #1387, opened 2022-11-16
fix(ci): paths-ignore filter

rymnc:
The paths-ignore filter ignores changes if code and docs are updated together in one commit.

- fix(ci): when docs and code is changed, workflows run
- fix(ci): add more paths when workflow should run

### DIFF
--- a/.github/workflows/test1.yml
+++ b/.github/workflows/test1.yml
@@ -11,6 +11,10 @@ on:
       - 'metrics/**'
       - 'docs/**'
       - '**.md'
+
+      # Should run the workflow if changes are made to the following files
+      - '!waku/v1/**.nim'
+      - '!tests/v1/**.nim'
   push:
     branches:
       - master
@@ -19,6 +23,8 @@ on:
       - 'metrics/**'
       - 'docs/**'
       - '**.md'
+      - '!waku/v1/**.nim'
+      - '!tests/v1/**.nim'
 
 concurrency: 
   group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
--- a/.github/workflows/test2.yml
+++ b/.github/workflows/test2.yml
@@ -15,6 +15,9 @@ on:
       - 'metrics/**'
       - 'docs/**'
       - '**.md'
+      # Should run the workflow if changes are made to the following files
+      - '!waku/v2/**.nim'
+      - '!tests/v2/**.nim'
   push:
     branches:
       - master
@@ -23,6 +26,8 @@ on:
       - 'metrics/**'
       - 'docs/**'
       - '**.md'
+      - '!waku/v2/**.nim'
+      - '!tests/v2/**.nim'
 
 concurrency: 
   group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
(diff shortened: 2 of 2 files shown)


.github/workflows/test1.yml
```diff
@@ -11,6 +11,10 @@ on:
       - 'metrics/**'
       - 'docs/**'
       - '**.md'
+
+      # Should run the workflow if changes are made to the following files
+      - '!waku/v1/**.nim'
+      - '!tests/v1/**.nim'
```
LNSD, line 17: This was implicit. Please remove this. For example, the "whisper" protocol changes are not included here.

.github/workflows/test1.yml
```diff
@@ -19,6 +23,8 @@ on:
       - 'metrics/**'
       - 'docs/**'
       - '**.md'
+      - '!waku/v1/**.nim'
+      - '!tests/v1/**.nim'
```
LNSD, line 27: This was deliberate. We want all tests to be executed when merging into `master`. Please, remove this too.

.github/workflows/test2.yml
```diff
@@ -15,6 +15,12 @@ on:
       - 'metrics/**'
       - 'docs/**'
       - '**.md'
+      # Should run the workflow if changes are made to the following files
+      - '!waku/v2/**.nim'
+      - '!tests/v2/**.nim'
+      - '!tests/all_tests_v2.nim'
+      - '!apps/**/**.nim'
+      - '!tools/**/**.nim'
```
LNSD, line 23: Same here. This was implicit. No need to specify it. Please remove it.

.github/workflows/test2.yml
```diff
@@ -23,6 +29,8 @@ on:
       - 'metrics/**'
       - 'docs/**'
       - '**.md'
+      - '!waku/v2/**.nim'
+      - '!tests/v2/**.nim'
```
LNSD, line 33: Same here. This was deliberate. We want all tests to be executed when merging into the master branch. Please, remove this too.

LNSD (changes requested):
I don't see the increment compared to the current CI workflow definitions. Can you clarify the following statement? What are you trying to solve with it?

> fix(ci): when docs and code is changed, workflows run

Additionally, when you change the `ignore-paths` section in `test2.yaml`, you have to do the complementary with `test2-ingore.yml` otherwise, the checks will remain not executed since there will be files that are not covered by any of the workflows.

.github/workflows/test1.yml
rymnc: > I don't see the increment compared to the current CI workflow definitions. Can you clarify the following statement? What are you trying to solve with it?
> 
> > fix(ci): when docs and code is changed, workflows run
> 
> Additionally, when you change the `ignore-paths` section in `test2.yaml`, you have to do the complementary with `test2-ingore.yml` otherwise, the checks will remain not executed since there will be files that are not covered by any of the workflows.

Refer to this PR - https://github.com/status-im/nwaku/pull/1386. Tests are skipped even though there are changes made to the code.

rymnc:
Closing to be revisited later

PR #1388, opened 2022-11-16
refactor(protocols): decouple protocol rpc types from api types

LNSD:
Protocol wire format types (a.k.a. "RPC types") should be decoupled from the exposed API and application types. The reasoning can be found [here](https://github.com/vacp2p/waku/pull/1#issuecomment-1313879596). 

As I already did previously with the Waku store protocol code, this PR aims to start the work of decoupling the "RPC types" from the "API/application types".

- [x] Do not re-export the RPC types. Or import them explicitly so, in the near future, the code can be refactored and decoupled.


rymnc (approved):
LGTM :)

PR #1389, opened 2022-11-16
chore(networkmonitor): use nim-presto + add timeout

alrevuelta:
- [x] Use `nim-presto` client for making http requests
- [x] Add timeout for requests and warn instead of error.

Related #1010 
Related https://github.com/status-im/infra-misc/issues/104

### DIFF
--- a/tools/networkmonitor/networkmonitor.nim
+++ b/tools/networkmonitor/networkmonitor.nim
@@ -4,17 +4,18 @@ else:
   {.push raises: [].}
 
 import
-  std/[tables,strutils,times,sequtils,httpclient],
+  std/[tables,strutils,times,sequtils],
   chronicles,
   chronicles/topics_registry,
   chronos,
+  chronos/timer as ctime,
   confutils,
   eth/keys,
   eth/p2p/discoveryv5/enr,
   libp2p/crypto/crypto,
   metrics,
   metrics/chronos_httpserver,
-  presto/[route, server],
+  presto/[route, server, client],
   stew/shims/net
 
 import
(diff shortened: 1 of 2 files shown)


tools/networkmonitor/networkmonitor.nim
```diff
@@ -81,13 +82,19 @@ proc setConnectedPeersMetrics(discoveredNodes: seq[Node],
     allPeers[peerId].ip = ip
 
     # get more info the peers from its ip address
-    let location = await ipToLocation(ip, client)
-    if not location.isOk():
+    var location: NodeLocation
+    try:
+      # IP-API endpoints are now limited to 45 HTTP requests per minute
+      # TODO: As network grows, find a better way to now block the whole app
```
jm-clius, line 88:
```suggestion
      # TODO: As network grows, find a better way to not block the whole app
```
:) Just in case someone in future believes they should _now_ really block everything.

tools/networkmonitor/networkmonitor.nim
```diff
@@ -4,17 +4,18 @@ else:
   {.push raises: [].}
 
 import
-  std/[tables,strutils,times,sequtils,httpclient],
+  std/[tables,strutils,times,sequtils],
   chronicles,
   chronicles/topics_registry,
   chronos,
+  chronos/timer as ctime,
```
jm-clius, line 11: For interest's sake, I remember that I also earlier needed both `chronos/timer` and `std/times` for some use cases before (though I can't remember details). Generally we avoid `std/times`, but there are still places where it may be necessary. I'm hoping we can abstract all of these use cases to our own `common/time` in future, as there's been some confusing issues that arose from different time impls in the past.

jm-clius (approved):
LGTM

tools/networkmonitor/networkmonitor.nim
alrevuelta: Interesting will take it into account :)

In my case I imported `std/times` to use `getTime()`. The chronos alternative would be `Moment.now()` I guess, but since I needed a readable time, getTime() made the work.

```nim
import chronos
import std/times

echo Moment.now() #533979900656875ns
echo getTime() # 2022-11-16T16:01:20+01:00
```

Issue #1361, opened 2022-11-09
feat: move protocol wire format codecs to nim-protobuf-serialization


LNSD:
### Background

TBD

### Details

TBD

### Acceptance criteria

TBD

arnetheduck:
keep in mind that there's a major rewrite due here: https://github.com/status-im/nim-protobuf-serialization/pull/36 - it needs more unit tests above all, in case you want to help push it along ;)

LNSD:
> keep in mind that there's a major rewrite due here: [status-im/nim-protobuf-serialization#36](https://github.com/status-im/nim-protobuf-serialization/pull/36) - it needs more unit tests above all, in case you want to help push it along ;)

I added this as a placeholder. Currently, I am concerned about how we manage protobufs in `nwaku` and the type mapping. I was thinking of putting some effort into improving it, but it may be worth pushing a joint initiative instead of having multiple implementations. Thanks for the heads up

PR #1390, opened 2022-11-16
chore: move protobuf module from utils to common

LNSD:
Following previous work with `sqlite.nim` and `envvar_serialziation.nim` modules, this PR moves the protobuf module to the `common` module. This module should, in the future, act as a "facade" for the protobuf codec APIs.

- [x] Move `protobuf.nim` under `common` module
- [x] Re-export libp2p's protobuf codec APIs
- [x] Update imports accordingly

### DIFF
--- a/waku/common/protobuf.nim
+++ b/waku/common/protobuf.nim
@@ -1,3 +1,5 @@
+# Extensions for libp2p's protobuf library implementation
+
 when (NimMajor, NimMinor) < (1, 4):
   {.push raises: [Defect].}
 else:
@@ -6,8 +8,11 @@ else:
 import
   libp2p/protobuf/minprotobuf,
   libp2p/varint
+ 
+export
+  minprotobuf,
+  varint
 
-# Collection of utilities related to protobuffer encoding
 
 proc write3*(proto: var ProtoBuffer, field: int, value: auto) =
   if default(type(value)) != value:
--- a/waku/v2/protocol/waku_filter/rpc_codec.nim
+++ b/waku/v2/protocol/waku_filter/rpc_codec.nim
@@ -4,10 +4,7 @@ else:
   {.push raises: [].}
 
 import
-  libp2p/protobuf/minprotobuf,
-  libp2p/varint
-import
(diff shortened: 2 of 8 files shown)


jm-clius (approved):
Great! Worth adding a line to the Coding Guide re "use `common/protobuf` and don't import libp2p minprotobuf directly"?

alrevuelta (commented):
Using @jm-clius guideline "use common/protobuf and don't import libp2p minprotobuf directly"?, perhaps we are missing updating `waku_peer_storage.nim` as well?

We can leave that for another PR, but do we also want to remove `libp2p/protobuf/minprotobuf` from the tests?

LNSD:
> Great! Worth adding a line to the Coding Guide re "use common/protobuf and don't import libp2p minprotobuf directly"?

Note that I have only touched the protocol modules. I am not 100% sure if we should do it. I have not analyzed the impact of moving everything to import this common module. 

I had something in mind: do that when we make a decision about the `protobuf` library (see: https://github.com/status-im/nwaku/issues/1361) 

> Using @jm-clius guideline "use common/protobuf and don't import libp2p minprotobuf directly"?, perhaps we are missing updating `waku_peer_storage.nim` as well?

I am unfamiliar with `waku_peer_store.nim`; that's why I left it out. To minimize the risk. The `waku_peer_store` needs a cleanup. Changing something that will change in the near future, IMO, is out of the scope of this PR.

> We can leave that for another PR, but do we also want to remove `libp2p/protobuf/minprotobuf` from the tests?

Same thing here. I don't know the implication of removing the types. And I am pretty sure that, in some cases, the analysis will lead us to a deeper refactoring of the test suite. I want to avoid this extra work to be part of this PR.

rymnc (approved):
LGTM!

LNSD:
The check failure can be ignored. It is not related to the changes.

The [offending test case](https://github.com/status-im/nwaku/actions/runs/3479719256/jobs/5818611674#step:7:4155) is already tracked in #1357.

PR #1393, opened 2022-11-17
fix(store): waku store rpc codec support optional fields

LNSD:
To follow the [Waku store `2.0.0-beta4` protobuf definition](https://github.com/vacp2p/waku/blob/main/waku/store/v2beta4/store.proto):

- [x] Add support for the `Option[x]` type to the `protobuf` module at serialization time. 
- [x] Remove the behavior of not serializing a value when it matches the type's default.
- [x] Update RPC types to match the protobuf definition. Use explicit `Option[T]` types when the field is optional.
- [x] Update API-RPC type mappings accordingly
- [x] Update the Waku store protocol and tests accordingly

This is a dependency of the _Make `waku_message.nim` comply with the protobuf definition_ task.

### DIFF
--- a/waku/v2/protocol/waku_store/rpc.nim
+++ b/waku/v2/protocol/waku_store/rpc.nim
@@ -49,9 +49,9 @@ type
 
   PagingInfoRPC* = object
     ## This type holds the information needed for the pagination
-    pageSize*: uint64
-    cursor*: PagingIndexRPC
-    direction*: PagingDirectionRPC
+    pageSize*: Option[uint64]
+    cursor*: Option[PagingIndexRPC]
+    direction*: Option[PagingDirectionRPC]
 
 
 type
@@ -60,10 +60,10 @@ type
 
   HistoryQueryRPC* = object
     contentFilters*: seq[HistoryContentFilterRPC]
-    pubsubTopic*: PubsubTopic
-    pagingInfo*: PagingInfoRPC # used for pagination
-    startTime*: Timestamp # used for time-window query
-    endTime*: Timestamp # used for time-window query
+    pubsubTopic*: Option[PubsubTopic]
+    pagingInfo*: Option[PagingInfoRPC]
+    startTime*: Option[int64]
+    endTime*: Option[int64]
 
   HistoryResponseErrorRPC* {.pure.} = enum
     ## HistoryResponseErrorRPC contains error message to inform  the querying node about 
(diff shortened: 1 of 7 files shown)


waku/v2/protocol/waku_store/rpc.nim
```diff
@@ -112,55 +112,53 @@ proc toAPI*(rpc: PagingIndexRPC): HistoryCursor =
...
-                else: some(rpc.startTime)
+    startTime = rpc.startTime
 
-    endTime = if rpc.endTime == default(Timestamp): none(Timestamp)
-              else: some(rpc.endTime)
+    endTime = rpc.endTime
     
-    pageSize = if rpc.pagingInfo == default(PagingInfoRPC): 0.uint64
-               else: rpc.pagingInfo.pageSize
+    pageSize = if rpc.pagingInfo.isNone() or rpc.pagingInfo.get().pageSize.isNone(): 0'u64
```
rymnc, line 157: Is there a better way to check if a nested option isNone, i.e a proc/template like -
```nim
nestedFieldIsNone(pagingInfo, pageSize)
```
no real opinion on this, just curious :)

waku/v2/protocol/waku_store/rpc_codec.nim
```diff
@@ -33,180 +33,222 @@ proc encode*(index: PagingIndexRPC): ProtoBuffer =
...
+    return err(ProtoError.RequiredFieldMissing)
 
   var queryBuffer: seq[byte]
-  discard ?pb.getField(2, queryBuffer)
-  rpc.query = ?HistoryQueryRPC.decode(queryBuffer)
+  if not ?pb.getField(2, queryBuffer):
+    rpc.query = none(HistoryQueryRPC)
+  else:
+    let query = ?HistoryQueryRPC.decode(queryBuffer)
+    rpc.query = some(query)
```
rymnc, line 245: I see this structure repeated a little bit in the decoding procs. I know we aren't supposed to use templates extensively, but this looks like a good place to use them. Doesn't have to be a part of this PR though :)

rymnc (approved):
LGTM!

waku/v2/protocol/waku_store/rpc.nim
LNSD: I know in other languages it exists, like Kotlin's `?` operator. I don't know about Nim.

In any case, here, the solution is to simplify and relocate the protobuf message fields.

waku/v2/protocol/waku_store/rpc_codec.nim
LNSD: First, establish the pattern, then extract and apply the template. This is how I usually create "extensible code" 😉

>  but this looks like a good place to use them.

Here we are using `libp2p`'s protobuf library. I don't know yet if we'll stick with it or if we'll move to `nim-protobuf-serialization` (see #1361).

waku/v2/protocol/waku_store/rpc.nim
rymnc: Sounds good!

waku/v2/protocol/waku_store/rpc_codec.nim
rymnc: Sounds good, thanks!

jm-clius (approved):
LGTM!

PR #1396, opened 2022-11-17
fix(filter): waku filter rpc codec support optional fields

LNSD:
> ~~Waiting for https://github.com/vacp2p/waku/pull/9~~

To follow the [Waku filter `2.0.0-beta1` protobuf definition](https://github.com/vacp2p/waku/blob/main/waku/filter/v2beta1/filter.proto):

- [x] Update RPC types to match the protobuf definition. Use explicit `Option[T]` types when the field is optional.
- [x] Update the protocol accordingly

This is a dependency of the _Make `waku_message.nim` comply with the protobuf definition_ task.

### DIFF
--- a/waku/common/protobuf.nim
+++ b/waku/common/protobuf.nim
@@ -19,6 +19,8 @@ proc write3*(proto: var ProtoBuffer, field: int, value: auto) =
   when value is Option:
     if value.isSome():
       proto.write(field, value.get())
+  elif value is bool:
+    proto.write(field, zint(value))
   else:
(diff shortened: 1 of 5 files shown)


LNSD:
macOS test2 job failing due to the [same offending test case.](https://github.com/status-im/nwaku/actions/runs/3495100003/jobs/5856552806#step:7:4155). Test failure already tracked in #1357. Merging without re-running the test cases and Ubuntu test2 job is passing.

PR #1399, opened 2022-11-17
refactor(rln): decouple rln types from waku message type

LNSD:
From this comment https://github.com/status-im/nwaku/pull/1386#discussion_r1025370353 related to the cyclic import issues in RLN:

> We need to break that cycle. And it should be done this way:
> 
> 1. Waku message's `proof` field should be changed to `seq[byte]` type. With that, we break the `waku_rln_relay -> waku_message` dependency.
> 2. Consequently, we'll have to decode that buffer within the `waku_rln_relay` module. The buffer should be decoded whenever you need to access the `proof` field.
> 
> Given the current RLN code layout, this last part will be cumbersome. But this is what we need.

### DIFF
--- a/waku/v2/protocol/waku_message.nim
+++ b/waku/v2/protocol/waku_message.nim
@@ -15,18 +15,14 @@ import
   ../../common/protobuf,
   ../utils/time
 
-when defined(rln):
-  import
-    ./waku_rln_relay/protocol_types
-
-
 const MaxWakuMessageSize* = 1024 * 1024 # In bytes. Corresponds to PubSub default
+
 
 type
   PubsubTopic* = string
   ContentTopic* = string
 
-const 
+const
   DefaultPubsubTopic*: PubsubTopic = PubsubTopic("/waku/2/default-waku/proto")
   DefaultContentTopic*: ContentTopic = ContentTopic("/waku/2/default-content/proto")
 
@@ -41,9 +37,9 @@ type WakuMessage* = object
     # this field will be used in the rln-relay protocol
     # XXX Experimental, this is part of https://rfc.vac.dev/spec/17/ spec and not yet part of WakuMessage spec
     when defined(rln):
-      proof*: RateLimitProof
+      proof*: seq[byte]
     # The ephemeral field indicates if the message should
-    # be stored. bools and uints are 
+    # be stored. bools and uints are
     # equivalent in serialization of the protobuf
     ephemeral*: bool
 
@@ -58,7 +54,7 @@ proc encode*(message: WakuMessage): ProtoBuffer =
   buf.write3(3, message.version)
   buf.write3(10, zint64(message.timestamp))
   when defined(rln):
-    buf.write3(21, message.proof.encode())
+    buf.write3(21, message.proof)
   buf.write3(31, uint64(message.ephemeral))
   buf.finish3()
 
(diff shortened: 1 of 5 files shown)


rymnc (approved):
LGTM

LNSD:
@rymnc Build is failing in the CI. Would you mind taking over this PR and fixing the issue? 🙏🏼

rymnc:
> @rymnc Build is failing in the CI. Would you mind taking over this PR and fixing the issue? 🙏🏼

will do

LNSD:
LGTM ✅

waku/v2/protocol/waku_message.nim
```diff
@@ -41,9 +37,9 @@ type WakuMessage* = object
     # this field will be used in the rln-relay protocol
     # XXX Experimental, this is part of https://rfc.vac.dev/spec/17/ spec and not yet part of WakuMessage spec
     when defined(rln):
-      proof*: RateLimitProof
+      proof*: seq[byte]
```
staheri14, line 40: There was an initial intention to define `RateLimitProof` and not directly `seq[bytes]`, it is an alias and will make it swappable with other future types and easier and clearer for interfacing with other procs/modules. Also, it is inefficient to have recurring decoding/encoding when we can do the decoding once and keep using the resultant `RateLimitProof` object. 

An alternative approach to breaking the current circular dependency is: Given that the `WakuMessage` module only depends on the `RateLimitProof` data type, I suggest moving this type i.e., `RateLimitProof` out to a separate module like `proof_types` and then changing `WakuMessage` dependent to that module.

staheri14 (commented):
Thanks for looking into this issue @LNSD! I have concerns about the way the dependency issue is tackled. Please see my comment below. I appreciate your reconsideration of the solution.

waku/v2/protocol/waku_message.nim
LNSD: I understand this was intentional from the beginning. But I don't understand the following:

> There was an initial intention to define RateLimitProof and not directly seq[bytes], **it is an alias and will make it swappable with other future types** and easier and clearer for interfacing with other procs/modules.

What do you mean by "it is an alias"? How do you expect to swap this?

----

> Also, it is inefficient to have recurring decoding/encoding when we can do the decoding once and keep using the resultant RateLimitProof object.

Regarding the state of the RLN code, performance in decoding is one of the less important things to worry about at this moment. Decoding a complex field, like `proof: RateLimitProof`, that is not used in other protocols adds overhead to the rest of the protocols.
 
Different strategies could be used to optimize the implementation (e.g., duplicate the WakuMessage type within the RLN module and decode in the Waku relay handler with the new type). But in any case, the optimization should be done once the code is in better shape.

Additionally, using a simple buffer (a.k.a. `seq[byte]`) opens the possibility to change the field format if the `RateLimitProof` format changes without changing it in a thousand places in the code. We will be able to modify the RLN code freely without having to touch common parts of the codebase. 

In other words, this approach is less risky and future-proof.

> An alternative approach to breaking the current circular dependency is: Given that the WakuMessage module only depends on the `RateLimitProof` data type, I suggest moving this type, i.e., `RateLimitProof` out to a separate module like `proof_types` and then changing WakuMessage dependent to that module.

Sorry, but this is not a good option. 

We are using the `waku_node` as a leave of the imports graph. All modules import the `waku_message`. So this module should not import any other module.

With that approach, we are adding unnecessary complexity (e.g., unnecessary dependencies between modules, scattering code that is not collocated around instead of keeping it within the module that uses it). 

This is one of my software principles for maintainable code:

> **All the code that changes together should be together.**

PR #1403, opened 2022-11-20
chore: add editorconfig file

LNSD:
This PR addresses a long-existing concern: trailing whitespaces.

- [x] Add `.editorconfig` file for editor configuration homogeneity. Configure to remove the trailing whitespaces automatically. 

Every time a file is modified, the trailing whitespaces will be removed. This will allow us to follow an incremental approach toward a cleaner codebase.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,18 @@
+root = true
+
+[*]
+charset = utf-8
+indent_style = space
+indent_size = 2
+end_of_line = lf
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+
+[{Makefile, *.sh}]
+indent_style = tab
+
+# Trailing spaces in markdown indicate word wrap
+[{*.markdown,*.md}]
+trim_trailing_spaces = false
+max_line_length = 80


alrevuelta:
I assume this involves installing an extension and loading this config file? like "EditorConfig for VS Code"? Should we document this somewhere?

.editorconfig
```diff
@@ -0,0 +1,18 @@
...
+insert_final_newline = true
+
+
+[{Makefile, *.sh}]
+indent_style = tab
+
+# Trailing spaces in markdown indicate word wrap
+[{*.markdown,*.md}]
+trim_trailing_spaces = false
+max_line_length = 80
```
jm-clius, line 18: What is the effect of this `max_line_length`? Will longer lines be auto-wrapped?

LNSD:
> I assume this involves installing an extension and loading this config file? like "EditorConfig for VS Code"? Should we document this somewhere?

I thought VS Code was already shipped with the editorconfig extension installed by default. Given that you are the one that joined most recently, and I am very busy with the Waku store investigation, @alrevuelta can you document it?

jm-clius (approved):
Thanks! Happy with this, but I assume it's Visual Studio-specific? Since some contributors use other editors (e.g. vim), we can perhaps just provide them with style guidelines.

.editorconfig
LNSD: This is from the editorconfig's [default values](https://github.com/editorconfig/editorconfig-defaults/blob/master/.editorconfig#L13). I just copied it here, so we don't have any surprises with implicit configurations.

> Will longer lines be auto-wrapped?

VS code only supports a small subset of editorconfig parameters: https://marketplace.visualstudio.com/items?itemName=EditorConfig.EditorConfig#supported-properties 
So the answer is: no, it is not supported by the vscode/plugin.

LNSD:
> Thanks! Happy with this, but I assume it's Visual Studio-specific? Since some contributors use other editors (e.g. vim), we can perhaps just provide them with style guidelines.

In that case, they'll have to install the `editorconfig` support for their editor. For example, vim has [an official plugin](https://github.com/editorconfig/editorconfig-vim).

alrevuelta:
@LNSD Documented in our coding guidelines, miscellaneous [section](https://hackmd.io/@vac/main/%2F5oIVOuQcSM2WKrnDLUPcpg#Miscellaneous). Using VSCode I had to install that extension for it to work.

PR #1407, opened 2022-11-21
fix(peer-exchange): start px discv5 loop after discv5 has started

kaiserd:
The peer exchange discv5 loop was stared before discv5 was listening, causing the loop to immediately stop.
This PR fixes this issue.


alrevuelta (approved):
thanks! lgtm!

LNSD (approved):
LGTM ✅

jm-clius (approved):
LGTM. Seems indeed like your failing test is the one investigated in https://github.com/waku-org/nwaku/pull/1368

PR #1406, opened 2022-11-21
feat(archive): add waku archive implementation

LNSD:
This PR got precipitated due to #1400.

- [x] Split the message storage logic out of the Waku store protocol module.
- [x] Decouple the message validation logic. 
	+ Extract the current message validation logic into the "default" message validator.
	+ Improve the "invalid message" metrics by differentiating invalid message types.
- [x] Collocate under the `waku_archive` module the "driver" (formerly `message_store`) and "retention_policy" modules.
- [x] Ported and added new test case covering the functionality.

ℹ️ Instead of performing the fixes and improvements for #1400 in the `message_store` code, they will be performed in the `waku_archive` module (e.g., the `sqlite_driver` module).

Issue #1282, opened 2022-10-20
fix and reenable the store resume flaky test case


LNSD:
### Background
One store resume test case has been filing intermittently causing many inconveniences:

```
[2022-10-19T19:50:59.747Z]     /home/jenkins/workspace/nim-waku_prs_linux_PR-1229/tests/v2/test_waku_store_resume.nim(198, 28): Check failed: restoredMessagesCount == 14
[2022-10-19T19:50:59.747Z]     restoredMessagesCount was 13
[2022-10-19T19:50:59.747Z]     /home/jenkins/workspace/nim-waku_prs_linux_PR-1229/tests/v2/test_waku_store_resume.nim(199, 26): Check failed: storedMessagesCount == 14
[2022-10-19T19:50:59.747Z]     storedMessagesCount was 13
[2022-10-19T19:50:59.747Z]   [FAILED] resume history from a list of candidates - online and offline peers
```

### Details
The test case has been disabled by #1281

### Acceptance criteria
- [ ] Reenable the test case, by reverting the commit
- [ ] The test shall pass

Issue #1129, opened 2022-09-08
Review nwaku's resume store on boot mechanism


LNSD:
### Problem
> A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]

TBD
### Suggested solution
> A clear and concise description of what you want to happen.

TBD

### Alternatives considered
> A clear and concise description of any alternative solutions or features you've considered.

TBD

### Additional context
Nwaku's _resume store on boot_ mechanism is implemented using the timeframe query mechanism specified by the [21/WAKU2-FTSTORE RFC](https://rfc.vac.dev/spec/21/).

### Acceptance criteria
> A list of tasks that need to be done for the issue to be considered resolved.

TBD

LNSD:
In _#nwaku_ channel at VAC's discord server @fryorcraken has asked the following questions:

> 1. It is currently not enabled in any fleet, right?
> 2. It only synchronize in-memory messages, right?
> 3. In what instance would a message be in the sqlite DB but not in memory?
> 4. I assume that for now, we would prefer to investigate missing message (ie, issue with relay?) before turning on FT-Store that may hide defect, right?
> 5. Anything else I need to know?

> One more question: when enabled, what triggers a sync?

> Also, would it make sense at a later stage to enable ft-store so that a node that is being reboot for upgrade/config change can catch up on messages?

LNSD:
My answer @fryorcraken to your questions:

> It is currently not enabled in any fleet, right?

As far as I have seen, it is not. It requires the nodes to have a "store peer" configured, and at the moment, it is not the case. In Kibana, I observe warnings saying there's no peer to retrieve messages.

> It only synchronize in-memory messages, right?

In the current implementation, we only "resume" (the verb used in the code) the messages other nodes have stored during the time node was offline. And then it inserts them in the in-memory store. So if the node store protocol is working in "sqlite-only" mode, it won't insert the retrieved messages in the persistent store.

This is the remaining work from unifying the in-memory (aka message queue) and the persistent (aka message store) interface. After that, we'll be able to use the in-memory and the persistent store interchangeability.

> In what instance would a message be in the sqlite DB but not in memory?

If the node store protocol runs in "sqlite-only" mode. Everything interacts with the SQLite DB. The messages are inserted in the DB, and the history queries are performed to the DB too.

> I assume that for now, we would prefer to investigate missing message (ie, issue with relay?) before turning on FT-Store that may hide defect, right?

[...]

PR #1409, opened 2022-11-22
feat(archive): add waku archive implementation

LNSD:
> ~~This PR depends on #1406~~

This PR takes the last step to replace the current waku store implementation with the new Waku archive.

⚠️ **IMPORTANT:** The "resume store on boot" functionality and the tests need some work. Due to the circumstances, this cannot be performed in this PR. Two issues exist to track that work.

* The functionality has been moved behind the `waku_exp_resume_store` compilation flag in this PR.
	- #1129 
* The "resume store on boot" tests do not compile.
	- #1282


jm-clius (approved):
LGTM as building on #1406! :)

PR #1410, opened 2022-11-23
feat(ci): run experimental features in diff ci

rymnc:
This PR adds 2 new workflows, which are the same as the `test2` and `test2-ignore` workflows, but run with experimental flags enabled.

- Ensures that when `ONCHAIN_RLN` is set, even rln is passed as NIM_PARAMS
- Adds `EXPERIMENTAL` flag which encompasses all experimental features
- Removes RLN default compilation if `CI` is set
- They are defined separately so that even if the experimental CI fails, PRs can be merged without bypassing branch protections.


Makefile
```diff
@@ -204,25 +204,31 @@ docs: | build deps
...
 ifeq ($(RLN), true)
-NIM_PARAMS := $(NIM_PARAMS) -d:rln
-else ifeq ($(CI), true)
-NIM_PARAMS := $(NIM_PARAMS) -d:rln
+compile_rln = true
 endif
 
 # control compilation of rln tests that require on chain interaction
 ifeq ($(ONCHAIN_RLN), true) 
+compile_rln = true
```
LNSD, line 216: What about putting everything under `compile_rln` and renaming it to `experimental_rln`?

This is what I was planning for the static compilation PR.

Makefile
```diff
@@ -204,25 +204,31 @@ docs: | build deps
...
 # control rln code compilation
 ifeq ($(RLN), true)
-NIM_PARAMS := $(NIM_PARAMS) -d:rln
-else ifeq ($(CI), true)
-NIM_PARAMS := $(NIM_PARAMS) -d:rln
+compile_rln = true
 endif
 
 # control compilation of rln tests that require on chain interaction
 ifeq ($(ONCHAIN_RLN), true)
```
LNSD, line 215: Why are we still differentiating in the makefile between on-chain and off-chain RLN? Aren't we running all tests always?

LNSD (approved):
LGTM ✅

LNSD (approved):

Makefile
rymnc: Agreed, addressed in [055b387](https://github.com/waku-org/nwaku/pull/1410/commits/055b387a3d35467ad85fe50f7576a9af1683e2b9)

Makefile
rymnc: Addressed in [055b387](https://github.com/waku-org/nwaku/pull/1410/commits/055b387a3d35467ad85fe50f7576a9af1683e2b9)

rymnc:
Agreed, although I think we agreed before that while some experimental features are broken, that should not delay the nwaku release, etc. Having a separate CI just for experimental features allows this. I can add additional paths to the paths filter so that we run the experimental ci only when rln code is changed. wdyt?

Makefile
```diff
@@ -204,25 +204,24 @@ docs: | build deps
...
 endif
 
-# control compilation of rln tests that require on chain interaction
-ifeq ($(ONCHAIN_RLN), true) 
-NIM_PARAMS := $(NIM_PARAMS) -d:onchain_rln
-else ifeq ($(CI), true) 
-NIM_PARAMS := $(NIM_PARAMS) -d:onchain_rln
+ifdef experimental_rln
+	NIM_PARAMS := $(NIM_PARAMS) -d:rln
 endif
```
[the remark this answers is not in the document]
rymnc: Addressed in afaf31cd

LNSD:
> I can add additional paths to the paths filter so that we run the experimental ci only when rln code is changed. wdyt?

Make sense to me, but... It is not scalable. We'll have to change the CI definition every time an experimental feature is added. Maybe we should investigate how to do this within the Makefile/waku.nimble scripts.

We are increasing the complexity of the GH CI actions, which is not easy to manage. 

I'd instead merge this increment without filtering the RLN code. And investigate the possible solution of not filtering using the `paths`/`paths-ignore` clauses.

PR #1411, opened 2022-11-23
test(archive): randomize archive driver query test suites' fixtures

LNSD:
To ensure that we are not testing only the "happy path" of the different queries, some randomness has been added to the test fixtures to verify the message ordering.

### DIFF
--- a/tests/v2/waku_archive/test_driver_queue_query.nim
+++ b/tests/v2/waku_archive/test_driver_queue_query.nim
@@ -1,16 +1,24 @@
 {.used.}
 
 import
-  std/[options, sequtils],
+  std/[options, sequtils, random],
   testutils/unittests,
-  chronos
+  chronos,
+  chronicles
 import
   ../../../waku/v2/protocol/waku_archive,
   ../../../waku/v2/protocol/waku_archive/driver/queue_driver,
   ../../../waku/v2/protocol/waku_message,
   ../utils,
-  ../testlib/common,
-  ../testlib/testutils
+  ../testlib/common
+
+
+logScope:
+  topics = "test archive queue_driver"
+
+
+# Initialize the random number generator
+common.randomize()
 
 
 proc newTestSqliteDriver(): ArchiveDriver =
@@ -33,7 +41,7 @@ suite "Queue driver - query by content topic":
 
     let driver = newTestSqliteDriver()
 
-    let messages = @[
+    let expected = @[
       fakeWakuMessage(@[byte 0], contentTopic=DefaultContentTopic, ts=ts(00)),
       fakeWakuMessage(@[byte 1], contentTopic=DefaultContentTopic, ts=ts(10)),
       fakeWakuMessage(@[byte 2], contentTopic=contentTopic, ts=ts(20)),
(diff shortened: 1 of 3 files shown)


tests/v2/waku_archive/test_driver_queue_query.nim
```diff
@@ -61,7 +67,7 @@ suite "Queue driver - query by content topic":
     let filteredMessages = res.tryGet().mapIt(it[1])
     check:
       filteredMessages.len == 5
-      filteredMessages == messages[0..4]
+      filteredMessages == expected[0..4]
```
jm-clius, line 70: Should we swap these checks around or remove the `len` check which is a bit redundant (here and elsewhere)? Reason: since the message order will now be random, if one test case fails we should understand what ordering made it fail. The `.len` check will simply log unexpected length on failure and quit, whereas the `expected` set should log the test set on failure.

jm-clius (approved):
Approving for two reasons:
1. We already have some indeterminacy in CI testing (a.k.a. sleeps) and I see this as a separate issue.
2. I understand the urgency now to increase our test coverage.

However, I think we should bear in mind that CI should preferably be deterministic with enumerated test cases. Fuzzing-type tests and randomizations should of course also be done regularly, but we don't have the infra yet.

My comment below is re making the failing case visible so we can trace back to whichever ordering made the CI fail.

tests/v2/waku_archive/test_driver_queue_query.nim
LNSD: > remove the `len` check which is a bit redundant (here and elsewhere)?

Agree. Doing it before merging.

LNSD: > The `.len` check will simply log unexpected length on failure and quit, whereas the `expected` set should log the test set on failure.

That's not 100% accurate. The `len` and the expected messages assertions are under the same `check` block so if it fails, both failing assertions will be displayed. So we still see the length mismatch and the messages list mismatch.

As said above, I am removing the length issue since it is redundant.

LNSD:
> Approving for two reasons:
> 
> 1. We already have some indeterminacy in CI testing (a.k.a. sleeps) and I see this as a separate issue.
> 2. I understand the urgency now to increase our test coverage.

Yes, in some way this is a "shortcut" for testing the drivers.

> However, I think we should bear in mind that CI should preferably be deterministic with enumerated test cases. Fuzzing-type tests and randomizations should of course also be done regularly, but we don't have the infra yet.
> 
> My comment below is re making the failing case visible so we can trace back to whichever ordering made the CI fail. 

I partially agree with the deterministic test cases idea.

There are special cases where having non-deterministic test cases can provide more value. But I understand that it adds some difficulty when debugging since the test inputs will be non-deterministic. To solve that something else we should add to better understand how the test case was executed (e.g., a debug statement printing the insertion order in the database).

In this case, the functionality should behave exactly the same way no matter the message insertion order. With a non-deterministic message insertion we are testing a scenario closer to the real world (e.g., SQLite does not insert messages sequentially). So if the test case fails at any point, we need to investigate the issue since this is critical.

alrevuelta (approved):
lgtm! great idea.

> I think we should bear in mind that CI should preferably be deterministic with enumerated test cases

Totaly agree and correct me if I'm missing something but `randomize` inits the seed right? So the tests are already deterministic? `shuffle` should shuffle identically in all runs since it uses the same seed?

rymnc (approved):
LGTM

LNSD:
> Totaly agree and correct me if I'm missing something but `randomize` inits the seed right? So the tests are already deterministic? `shuffle` should shuffle identically in all runs since it uses the same seed?

You missed this part of the PR: 

https://github.com/waku-org/nwaku/blob/test-archive-reandomize-fixtures/tests/v2/testlib/common.nim#L41-L47

I am calling `testlib`'s `common.randomize()`, which wraps `std/random`'s `randomize` proc and uses time as a seed. This will provide a pseudo-random sequence, hence non-deterministic, between executions.

LNSD:
The [same test case](https://github.com/waku-org/nwaku/actions/runs/3532196074/jobs/5926255475#step:7:3755) is failing. As always.

It is already tracked in #1357

Issue #1400, opened 2022-11-17
bug: inconsistent / missing messages in status.prod fleet


richard-ramos:
I created the following project to help debug this scenario: https://github.com/richard-ramos/test-waku-query

- Topic: `/waku/2/default-waku/proto`
- ContentTopic: `/waku/1/0xd6861a81/rfc26`
- StartTime: `1668613188`, // Convert it to nano
- EndTime: `1668696055`, // Convert to nano

and got these results:

```
>>>> 119 messages found in 16Uiu2HAkvEZgh3KLwhLwXg95e5ojM8XykJ4Kxi2T7hk22rnA7pJC in the first attempt (Used cursor 2 times)
>>>> 242 messages found in 16Uiu2HAkvEZgh3KLwhLwXg95e5ojM8XykJ4Kxi2T7hk22rnA7pJC in the second attempt (Used cursor 3 times)
>>>> 154 messages found in 16Uiu2HAkvEZgh3KLwhLwXg95e5ojM8XykJ4Kxi2T7hk22rnA7pJC in the third attempt (Used cursor 2 times)
>>>> 25 messages found in 16Uiu2HAm6HZZr7aToTvEBPpiys4UxajCTU97zj5v7RNR2gbniy1D in the first attempt (Used cursor 1 times)
>>>> 20 messages found in 16Uiu2HAm6HZZr7aToTvEBPpiys4UxajCTU97zj5v7RNR2gbniy1D in the second attempt (Used cursor 1 times)
>>>> 24 messages found in 16Uiu2HAm6HZZr7aToTvEBPpiys4UxajCTU97zj5v7RNR2gbniy1D in the third attempt (Used cursor 1 times)
>>>> 227 messages found in 16Uiu2HAkwBp8T6G77kQXSNMnxgaMky1JeyML5yqoTHRM8dbeCBNb in the first attempt (Used cursor 3 times)
>>>> 227 messages found in 16Uiu2HAkwBp8T6G77kQXSNMnxgaMky1JeyML5yqoTHRM8dbeCBNb in the second attempt (Used cursor 3 times)
>>>> 227 messages found in 16Uiu2HAkwBp8T6G77kQXSNMnxgaMky1JeyML5yqoTHRM8dbeCBNb in the third attempt (Used cursor 3 times)
[...]

PR #1415, opened 2022-11-24
fix(archive): patch the sqlite driver to solve results inconsistencies

LNSD:
This is a temporary patch until the SQLite library is improved. Making the SQLite manage a copy is far from ideal. Binding the parameters with this "destructor configuration" option,  implies that SQLite is making a copy of the buffer. This is inefficient. A more performant solution should be investigated.

From SQLite C documentation ([link](https://www.sqlite.org/c3ref/bind_blob.html)):

> The fifth argument to the BLOB and string binding interfaces controls or indicates the lifetime of the object referenced by the third parameter. These three options exist: 
> 1. A destructor to dispose of the BLOB or string after SQLite has finished with it may be passed. It is called to dispose of the BLOB or string even if the call to the bind API fails, except the destructor is not called if the third parameter is a NULL pointer or the fourth parameter is negative.
> 2.The special constant, [SQLITE_STATIC](https://www.sqlite.org/c3ref/c_static.html), may be passsed to indicate that the application remains responsible for disposing of the object. In this case, the object and the provided pointer to it must remain valid until either the prepared statement is finalized or the same SQL parameter is bound to something else, whichever occurs sooner. 
> 3. **The constant, [SQLITE_TRANSIENT](https://www.sqlite.org/c3ref/c_static.html), may be passed to indicate that the object is to be copied prior to the return from `sqlite3_bind_*()`. The object and pointer to it must remain valid until then. SQLite will then manage the lifetime of its private copy.**

The tests I have performed so far seem to confirm that these changes solve issue #1400. But to be 100% sure, more thorough testing needs to be done.

### DIFF
--- a/waku/common/sqlite/database.nim
+++ b/waku/common/sqlite/database.nim
@@ -4,12 +4,12 @@
 #
 # Most of it is a direct copy, the only unique functions being `get` and `put`.
 
-import 
+import
   std/os,
   stew/results,
   chronicles,
   sqlite3_abi
-  
+
 
 logScope:
   topics = "sqlite"
@@ -54,7 +54,7 @@ template checkErr*(op) =
 
 type
   DatabaseResult*[T] = Result[T, string]
-  
+
   SqliteDatabase* = ref object of RootObj
     env*: Sqlite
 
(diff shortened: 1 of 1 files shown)


waku/common/sqlite/database.nim
```diff
@@ -127,10 +127,15 @@ template prepare*(env: Sqlite, q: string, cleanup: untyped): ptr sqlite3_stmt =
 
 proc bindParam*(s: RawStmtPtr, n: int, val: auto): cint =
   when val is openarray[byte]|seq[byte]:
+    # The constant, SQLITE_TRANSIENT, may be passed to indicate that the object is to be copied
+    #  prior to the return from sqlite3_bind_*(). The object and pointer to it must remain valid
+    #  until then. SQLite will then manage the lifetime of its private copy.
```
jm-clius, line 132: I assume this means it handles the `malloc` but always `free` it again - more memory but (hopefully) not leaky memory?

jm-clius (approved):
Really great work. Quite something to find this. Well done!

waku/common/sqlite/database.nim
LNSD: We call the [`sqlite3_finalize`](https://www.sqlite.org/c3ref/finalize.html) function later in the code to release the prepared statement memory. 

This is what the documentation says about the SQLite statement destructor:

> The application must finalize every [prepared statement](https://www.sqlite.org/c3ref/stmt.html) in order **to avoid resource leaks**.

I interpret this as deallocating any memory associated with the prepared statement. Bindings included.

LNSD:
macOS test2 job failing due to the same offending test case. Test failure already tracked in #1357. Merging without re-running the checks and Ubuntu test2 job is passing.

rymnc (approved):
LGTM

PR #1417, opened 2022-11-24
chore(store): remove obsolete metrics and error labels

LNSD:
These metrics were moved to the `waku_archive_` namespace. Removing those that are obsolete.


jm-clius (approved):
LGTM

PR #1424, opened 2022-11-25
feat(peerstore): store peer direction

alrevuelta:
- [x] Store connection direction of each peer in the peerstore: `Inbound` or `Outbound`.
- [x] Add function in the peer store to retrieve peers by direction `getPeersByDirection`.
- [x] Add new metric `waku_connected_peers` with `direction` label.
- [x] Set `MaxConnectionsPerPeer` to `1` to ensure we only support one connection per peer. Note that this value was already set to `1` in `nim-libp2p` but it may change. If in the future we decide to support multiple connections, we would need to revisit this. Note that there is no change in here, we were not supporting multiple connections before.

Useful for debugging. A reachable node should have some `Inbound` peers.

```console
$ curl localhost:8003/metrics | grep waku_connected_peers

# HELP waku_connected_peers number of connected peers per direction inbound|outbound
# TYPE waku_connected_peers gauge
waku_connected_peers{direction="Outbound"} 12.0
waku_connected_peers_created{direction="Outbound"} 1669729760.0
```

Related https://github.com/waku-org/nwaku/issues/1206

### DIFF
--- a/tests/v2/test_peer_manager.nim
+++ b/tests/v2/test_peer_manager.nim
@@ -314,3 +314,56 @@ procSuite "Peer Manager":
       nodes[0].peerManager.peerStore[ConnectionBook][nodes[3].switch.peerInfo.peerId] == Connected
 
     await allFutures(nodes.mapIt(it.stop()))
+
+  asyncTest "Peer store keeps track of incoming connections":
+    # Create 4 nodes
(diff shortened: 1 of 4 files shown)


jm-clius (approved):
LGTM.

PR #1425, opened 2022-11-25
feat: enable AutoNAT and libp2p circuit relay

### LINKED ISSUE #1402: feat: Enable circuit relay in nim-libp2p
### Problem

Circuit relay is a useful protocol when two nodes can't communicate directly, but can do so via another node. Nodes could be prevented from creating direct connection by e.g. restrictive NATs, incompatible transports, etc. If two such nodes can both connect to a third node, this node can act as a relayer to establish communication between them.
See detailed description [here](https://docs.libp2p.io/concepts/nat/circuit-relay/).

### Suggested solution

Circuit relay must be enabled in nim-libp2p so that nwaku can act as a relayer for other nodes. [This tutorial](https://status-im.github.io/nim-libp2p/docs/circuitrelay/) explains more. It should be as simple as building the nwaku switch with `.withCircuitRelay`.

jm-clius:
## Summary

Closes #1402 
Enables two features on the nwaku switch and tests each functionality:

### 1. libp2p circuit relay

nwaku can serve as a libp2p circuit-relay server. Using a nwaku node as a circuit-relay server is an intermediate step within the NAT hole punching procedure. See [this document](https://status-im.github.io/nim-libp2p/docs/circuitrelay/) for more on circuit relay.

### 2. AutoNAT

Peers can request dialbacks from nwaku nodes to verify that they are publicly reachable. This is the first step before attempting a NAT hole punching procedure. AutoNAT is also a requirement towards: https://github.com/waku-org/nwaku/issues/1206

## Configurability

For now I don't see a good reason to make either protocol configurable. A node can enable circuit relay and AutoNAT with default configuration by default.
See [this issue](https://github.com/waku-org/pm/issues/7#issuecomment-1324013387) for more on the use of AutoNAT and libp2p circuit relay in the NAT hole punching procedure.

### DIFF
--- a/tests/all_tests_v2.nim
+++ b/tests/all_tests_v2.nim
@@ -52,7 +52,7 @@ import
   ./v2/test_peer_exchange,
   ./v2/test_waku_noise,
   ./v2/test_waku_noise_sessions,
-  ./v2/test_circuitrelay,
+  ./v2/test_waku_switch,
   # Utils
   ./v2/test_utils_keyfile
 
--- a/tests/all_tests_v2.nim
+++ b/tests/all_tests_v2.nim
@@ -52,6 +52,7 @@ import
   ./v2/test_peer_exchange,
   ./v2/test_waku_noise,
   ./v2/test_waku_noise_sessions,
+  ./v2/test_circuitrelay,
   # Utils
   ./v2/test_utils_keyfile
 
--- a/tests/v2/test_circuitrelay.nim
+++ b/tests/v2/test_circuitrelay.nim
@@ -1,0 +1,82 @@
+{.used.}
+
+import
+  testutils/unittests,
+  chronos,
+  libp2p,
+  libp2p/protocols/connectivity/relay/relay,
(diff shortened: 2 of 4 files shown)


Menduist:
@lchenut can you just confirm that this will just enable circuit v2, we don't want v1 enabled by default

lchenut:
It's V2 by default. For the V1, you have to do `withRelay(Relay.new(circuitRelayV1=true))` while building your switch.

rymnc:
Curious if we need to mention these two features, or AutoNAT at least, in the operator guide

alrevuelta (approved):
lgtm!

alrevuelta:
> Curious if we need to mention these two features, or AutoNAT at least, in the operator guide

@rymnc 
really good point. Since feature will be added per se here: https://github.com/waku-org/nwaku/issues/1206 I can document it there.

LNSD (approved):
LGTM ✅

PR #1429, opened 2022-11-28
fix: waku bridge test timing

jm-clius:
PR to partially address https://github.com/waku-org/nwaku/issues/1357.

This fix seems to at least make things better. If we still get intermittent failures, we can consider reworking this test or coming up with a different Future/Result mechanism that will circumvent the need for async sleeps. Let's monitor in conjunction with other CI tests.


LNSD (approved):
LGTM ✅

jm-clius:
@s1fr0 indeed, we'd have to monitor and perhaps disable this test. Note that I also _removed_ one timer.

PR #1431, opened 2022-11-28
fix(makefile): fix NIM_PARAMS flag update

s1fr0:
Test fleet is not correctly loading the rln flag passed to NIM_PARAMS. This PR fixes it by updating its value with `eval` command,


rymnc (approved):
LGTM, thanks!

jm-clius (approved):

PR #1432, opened 2022-11-28
fix(noise): `T_ss` should be `ss` not `se`

s1fr0:
This PR fixes a bug in the definition of the `T_ss` type in Noise module, that should be equal to the string `ss` and not `se`.

A great catch from @richard-ramos !


rymnc (approved):
LGTM

PR #1436, opened 2022-11-30
chore: set debug mode by default in Makefile

alrevuelta:
* When compiling, `-d:debug -d:disable_libbacktrace` is only set when `USE_LIBBACKTRACE=0`.
* Problem is that neither `make wakunode2` nor `Dockerfile` set this variable to **any** value (variable was not set)
* This causes that `-d:release` to be the default compile mode, which disables the stack traces for performance.
* I believe we should build at least our testing/staging environments with `d:debug`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -78,7 +78,7 @@ else # "variables.mk" was included. Business as usual until the end of this file
 # default target, because it's the first one that doesn't start with '.'
 all: | v1 v2
 
-v1: | wakunode1 example1 sim1 
+v1: | wakunode1 example1 sim1
 v2: | wakunode2 example2 sim2 wakubridge chat2 chat2bridge
 
 # Builds and run the test suite (Waku v1 + v2)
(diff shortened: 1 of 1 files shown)


jm-clius (commented):
Thanks. Still considering this and trying to figure out some details about nim builds. I don't think the issue is that we build in release mode. Although this would result in `--stacktrace:off`, we want to build with debugger symbols even for releases and libbacktrace should allow us to do that. Afaict this would be the case if we build with `--debugger:native` as explained [here](https://github.com/status-im/nim-libbacktrace#usage) even with `-d:release`. Our config should [enable this by default](https://github.com/waku-org/nwaku/blob/master/config.nims#L53) except on mac. So either I'm missing something related to libbacktrace or there is another unexplained misconfiguration.

jm-clius:
Or perhaps things are working as expected and we should just use `getBacktrace()` rather than attempt `getStacktrace()` as per https://github.com/status-im/nim-libbacktrace#usage? We may still need to explicitly create debug builds for certain environments, unless libbacktrace provides us with enough info as is (and allows gdb usage with sensible symbols).

alrevuelta:
Closing since we want to build in `release` mode and while `writeStacktrace` is indeed "disabled" in release mode, we can use `getBacktrace()` from libbacktrace instead, which less performance degradation https://github.com/waku-org/nwaku/pull/1440

PR #1437, opened 2022-11-30
chore(build): simplify project makefile and update dockerfile

LNSD:
Splitting the Makefile/Dockerfile changes in a separate PR from #1385:

- [x] Move all experimental features under `EXPERIMENTAL=true` flag. 
    + We are keeping `RLN=true` for local development purposes.
    + The `test2-experimental` CI job is already passing the `EXPERIMENTAL=true` env variable.
- [x] Simplify experimental dependencies flow (check the `librln` section in the `Makefile`)
- [x] Renamed RLN targets: `rlnlib` to `librln`, `cleanrln` to `clean-librln`
- [x] Pass the shared library artifact path using Nim compile-time defines and `{.strdefine.}` to `rln.nim` file. This is temporary; after the #1385 issues are solved, we'll remove this.
- [x] Pass `EXPERIMENTAL` build argument to `docker-image` build target.
- [x] Pass the `GIT_VERSION` to `docker-image` target as the `commit` container image label.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,64 +1,16 @@
-# Copyright (c) 2020 Status Research & Development GmbH. Licensed under
+# Copyright (c) 2022 Status Research & Development GmbH. Licensed under
 # either of:
 # - Apache License, version 2.0
 # - MIT license
 # at your option. This file may not be copied, modified, or distributed except
 # according to those terms.
-
-SHELL := bash # the shell used internally by Make
-
-# used inside the included makefiles
 BUILD_SYSTEM_DIR := vendor/nimbus-build-system
-
-# -d:insecure - Necessary to enable Prometheus HTTP endpoint for metrics
-# -d:chronicles_colors:none - Necessary to disable colors in logs for Docker
-DOCKER_IMAGE_NIMFLAGS ?= -d:chronicles_colors:none -d:insecure
-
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -99,93 +77,106 @@ else
...
+chat2bridge: | build deps
 	echo -e $(BUILD_MSG) "build/$@" && \
 		$(ENV_SCRIPT) nim chat2bridge $(NIM_PARAMS) waku.nims
 
-## Waku v2 tooling targets
+
+###################
+## Waku v2 tools ##
+###################
+.PHONY: wakucanary wakumonitor
```
alrevuelta, line 179: can we use `networkmonitor` instead of `wakumonitor` to be consistent with the target. https://github.com/waku-org/nwaku/pull/1437/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R185

alrevuelta (commented):
thanks for the PR! just a minor comment in `networkmonitor`

Makefile
LNSD: My bad, changing it 👍🏼

alrevuelta (approved):
lgtm! 🚀

jm-clius (approved):
LGTM. The use of libbacktrace is not yet clear to me, but I'm busy trying to figure it out.

rymnc (approved):

PR #1438, opened 2022-11-30
ci: merge test1 and tes2 github workflows

LNSD:
The previous GH actions did not scale. I had some limitations. Here we are merging the different workflow files into two and running the different tasks as jobs.

- [x] Use _Path Changes Filter_ Github actions to run dynamically the different jobs.
	+ Legacy code (whisper and v1) changes will trigger only the legacy builds and tests.
	+ Waku v2 changes will trigger only the v2 production and experimental builds and tests.
	+ Any dependencies changes (submodule version change) will trigger all jobs.
- [x] Cache the nim compiler to avoid rebuilding it. 
	+ Use the _nimbus-build-system_ submodule hash as the cache key. 
	+ If the submodule version changes, the cache will be invalidated.
- [x] In the experimental features workflow, cache the _zerokit_ build artifacts. 
	+ Use the nimbus-build-system submodule hash as the cache key. 
	+ If the submodule version changes, the cache will be invalidated.
- [x] Split workflows into build and test. This is just a test. We need to investigate and develop a better nim build-cache strategy.

The following screenshot shows the new checks' build times:

image

Some optimization can be done to reduce even more the nim build times. But it is out of this PR's scope.


rymnc (approved):
This is awesome! LGTM
You'll have to update branch protection rules for the new ci names

PR #1440, opened 2022-11-30
chore: print getBacktrace() after sigsegv

alrevuelta:
* Import `libbacktrace` and prints `getBacktrace()` after a `SIGSEGV`.
* Note that this requires `--debugger:native` which is enabled by default [but in mac](https://github.com/waku-org/nwaku/blob/master/config.nims#L51).
* This should help debug segfaults, providing a more complete stacktrace.

### DIFF
--- a/apps/wakunode2/wakunode2.nim
+++ b/apps/wakunode2/wakunode2.nim
@@ -9,6 +9,7 @@ import
   chronicles,
   chronos,
   metrics,
+  libbacktrace,
   system/ansi_c,
   eth/keys,
   eth/p2p/discoveryv5/enr,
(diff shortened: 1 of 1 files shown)


alrevuelta:
_(could have reused the prev PR, but too late :))_

jm-clius (approved):

apps/wakunode2/wakunode2.nim
```diff
@@ -709,12 +710,13 @@ when isMainModule:
...
   when defined(posix):
     proc handleSigsegv(signal: cint) {.noconv.} =
-      fatal "Shutting down after receiving SIGSEGV"
-      waitFor node.stop()
+      # Require --debugger:native
+      fatal "Shutting down after receiving SIGSEGV", stacktrace=getBacktrace()
 
-      # Only available with --stacktrace:on --linetrace:on
+      # Not available in -d:release mode
       writeStackTrace()
```
jm-clius, line 717: For interest's sake. If not in `-d:release` I imagine we'll now print the libbacktrace and stacktrace. Are both necessary/helpful?

PR #1442, opened 2022-12-01
ci(jenkins): support experimental flag in the pr and release pipelines

LNSD:
Recently the `EXPERIMENTAL` flag was added to the `Makefile` to compile the experimental features into the nwaku node binary conditionally.

- [x] Add a UI parameter to allow building/testing an experimental image
- [x] Pass the experimental flag to the make processes and the docker build command
- [x] Default to false the `EXPERIMENTAL` build argument in the Dockerfile 

-------

### Context

The problem is that we have been deploying EXPERIMENTAL images to the production/test fleets. Causing the fleets to fail with the following error:

```
could not load: /app/vendor/zerokit/target/release/librln.so
```

### DIFF
--- a/ci/Jenkinsfile.prs
+++ b/ci/Jenkinsfile.prs
@@ -29,6 +29,11 @@ pipeline {
       description: 'Makefile flags.',
       defaultValue: params.MAKEFLAGS ?: '-j6'
     )
+    booleanParam(
+      name: 'EXPERIMENTAL',
+      description: 'Enable experimental features.',
+      defaultValue: false
+    )
   }
 
   options {
@@ -59,7 +64,7 @@ pipeline {
 
     stage('Binaries') {
       parallel {
-        stage('V1') { 
+        stage('V1') {
           when { expression { v1changed } }
           steps { script {
             nix.shell("make ${params.MAKEFLAGS} NIMFLAGS=\"${params.NIMFLAGS}\" V=${params.VERBOSITY} v1")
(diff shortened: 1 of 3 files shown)


rymnc (approved):
LGTM

ci/Jenkinsfile.prs
```diff
@@ -29,6 +29,11 @@ pipeline {
       description: 'Makefile flags.',
       defaultValue: params.MAKEFLAGS ?: '-j6'
     )
+    booleanParam(
+      name: 'EXPERIMENTAL',
+      description: 'Enable experimental features.'
```
jakubgs, line 34: You're missing a comma.

ci/Jenkinsfile.release
```diff
@@ -36,6 +36,11 @@ pipeline {
         '-d:insecure',
       ].join(' ')
     )
+    booleanParam(
+      name: 'EXPERIMENTAL',
+      description: 'Enable experimental features.'
```
jakubgs, line 41: Also missing a comma.

jakubgs (approved):

jakubgs:
Actually, this doesn't work because the Docker image fails to build:
```
Step 21/27 : COPY --from=nim-build /app/vendor/zerokit/target/release/librln.so vendor/zerokit/target/release/librln.so
COPY failed: stat app/vendor/zerokit/target/release/librln.so: file does not exist
```
I should have tested this locally. I kinda expected this to fail.

PR #1443, opened 2022-12-01
ci(jenkins): split dockerfile into production and experimental stages

LNSD:
With the new `EXPERIMENTAL` build argument, as the `librln` artifacts are not built, the docker image build fails with the following error:

```
COPY failed: stat app/vendor/zerokit/target/release/librln.so: file does not exist
```


jakubgs (approved):
Looks good. Though i don't think we need the `stage` variable in the `Jenkinsfile`. Just inline it.

PR #1448, opened 2022-12-06
chore: remove unused imports

alrevuelta:
- [x] remove unused imports


jm-clius (approved):
Thanks!

LNSD (dismissed):
The checks are not passing. This PR is not ready to merge. We should convert it to draft.

alrevuelta:
will call this the *schrödinger import*, you truly don't know if its used or not until you remove it and compile. some imports were being labeled as unused but they are indeed used. while I agree that seeing unused imports its not nice, we will have to live with some.

related https://github.com/waku-org/nwaku/issues/1398

PR #1452, opened 2022-12-07
chore(rln-relay): address laundry list for tests

rymnc:
Addresses https://github.com/waku-org/nwaku/issues/1358

The tests now adhere to the style guide

- chore(rln-relay): s/var/let
- chore(rln-relay): s/var/let
- chore(rln-relay): s/check/require
- chore(rln-relay): s/check/require
- fix(rln-relay): s/check/require


LNSD (approved):
LGTM ✅

PR #1454, opened 2022-12-08
fix(ci): repo not cloned on startup

rymnc:
dorny/paths-filter needs the repo to be checked out for push events


LNSD (approved):
LGTM ✅

Thanks for that 👍🏼

PR #1456, opened 2022-12-09
fix(ci): changes to nim doc input

rymnc:
Args to the `nim doc` commmand have changed. `--accept` has been replaced by `--run`


rymnc:
> I think we should remove this CI job since we are not using it at all. Makes no sense to me to build this if we are not serving it.

https://waku-org.github.io/nwaku/ It's supposed to be here but it looks like its broken

rymnc:
on investigation it looks like the correct directory is not being served. https://github.com/waku-org/nwaku/blob/51bb92c3b622b5fe366af7ecdbfd44c7a50979e6/.github/workflows/docs.yml#L39
`publish_dir` is not set

jm-clius:
Hah. There's a (very old) issue for this: https://github.com/waku-org/nwaku/issues/337. I agree that we can disable for now while docs are not (yet) prioritized.

PR #1458, opened 2022-12-12
chore(networking): unify dialpeer functions

alrevuelta:
Refactors as part of the networking roadmap.

* Remove metric label `waku_peers_errors.inc["conn_init_failure"]` as its duplicated from `waku_peers_dials.inc(labelValues = ["failed"])`
* Unify `dialPeer` functions where all end up calling the same one but with a different input: `peerInfo`, `peerId`.
* Remove `connectToNode`, as its duplicated from `dialPeer`.
* Merge all two `connectToNodes` into a single one.
* Modify `DefaultDialTimeout` to 10 seconds.

### DIFF
--- a/waku/v2/node/peer_manager/peer_manager.nim
+++ b/waku/v2/node/peer_manager/peer_manager.nim
@@ -18,6 +18,7 @@ import
 export waku_peer_store, peer_storage, peers
 
 declareCounter waku_peers_dials, "Number of peer dials", ["outcome"]
+# TODO: Populate from PeerStore.Source when ready
 declarePublicCounter waku_node_conns_initiated, "Number of connections initiated", ["source"]
 declarePublicGauge waku_peers_errors, "Number of peer manager errors", ["type"]
 declarePublicGauge waku_connected_peers, "Number of connected peers per direction: inbound|outbound", ["direction"]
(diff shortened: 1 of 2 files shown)


waku/v2/node/peer_manager/peer_manager.nim
```diff
@@ -233,59 +248,31 @@ proc dialPeer*(pm: PeerManager, remotePeerInfo: RemotePeerInfo, proto: string, d
     trace "Adding newly dialed peer to manager", peerId= $remotePeerInfo.peerId, address= $remotePeerInfo.addrs[0], proto= proto
     pm.addPeer(remotePeerInfo, proto)
```
jm-clius, line 249: Perhaps for the next PR, but is this still necessary if dialing adds the peer in the underlying peer store in any case?

jm-clius (approved):
LGTM!

waku/v2/node/peer_manager/peer_manager.nim
alrevuelta: Yes, good point.

I'm actually preparing a PR where I will be decoupling **dialing** from **adding**. So just discv5,px,rpc,bootrap will be the ones calling `addPeer`. So dial peer won't be adding peers anymore.

LNSD (approved):
LGTM ✅

PR #1462, opened 2022-12-13
feat(networking): add backoff period after failed dial

alrevuelta:
Closes https://github.com/waku-org/nwaku/issues/1414

**Summary:**
After failing to dial a peer, it adds a backoff to it so that we don't attempt to dial the same peer again during some time. The time to wait depends on the amount of consecutive failures, calculated with the following formula:

```
initialBackoffInSec*(backoffFactor^(failedAttempts-1))
```

Note that `initialBackoffInSec` and `backoffFactor` are configurable values that allow to configure how aggressive the backoffs are. Using `initialBackoffInSec=120` and `backoffFactor=4` the times to wait would be:

```
120s, 480s, 1920, 7680s
```

This PR helps nodes to rapidly increase their number of connections, since less time is wasted in trying to connect to nodes that fail. This improvement is even more noticeable in networks whith a high ratio of non reachable peers. Note that this backoff only applies to `relay` peers.

**Changes**:
- [x] Add exponential backoff after a failed dial.
- [x] Some minor test refactoring.
- [x] Some minor logging improvements.
- [x] Directly use `switch.dial` in ping (keep alive)

### DIFF
--- a/waku/v2/node/peer_manager/peer_manager.nim
+++ b/waku/v2/node/peer_manager/peer_manager.nim
@@ -85,7 +85,7 @@ proc dialPeer(pm: PeerManager, peerId: PeerID,
 
   var reasonFailed = ""
   try:
-    if (await dialFut.withTimeout(DefaultDialTimeout)):
+    if (await dialFut.withTimeout(dialTimeout)):
       waku_peers_dials.inc(labelValues = ["successful"])
       # TODO: This will be populated from the peerstore info when ready
       waku_node_conns_initiated.inc(labelValues = [source])
@@ -316,6 +316,9 @@ proc connectToNodes*(pm: PeerManager,
     futConns.add(pm.dialPeer(RemotePeerInfo(node), proto, dialTimeout, source))
 
   await allFutures(futConns)
+  let successfulConns = futConns.mapIt(it.read()).countIt(it.isSome)
+
+  info "Finished dialing multiple peers", successfulConns=successfulConns, attempted=nodes.len
 
   # The issue seems to be around peers not being fully connected when
   # trying to subscribe. So what we do is sleep to guarantee nodes are
@@ -341,14 +344,14 @@ proc relayConnectivityLoop*(pm: PeerManager) {.async.} =
       await sleepAsync(ConnectivityLoopInterval)
       continue
 
-    var disconnectedPeers = pm.peerStore.getNotConnectedPeers().mapIt(RemotePeerInfo.init(it.peerId, it.addrs))
-    var withinBackoffPeers = disconnectedPeers.filterIt(pm.peerStore.canBeConnected(it.peerId, pm.initialBackoffInSec))
+    var notConnectedPeers = pm.peerStore.getNotConnectedPeers().mapIt(RemotePeerInfo.init(it.peerId, it.addrs))
+    var withinBackoffPeers = notConnectedPeers.filterIt(pm.peerStore.canBeConnected(it.peerId, pm.initialBackoffInSec))
     let numPeersToConnect = min(min(maxConnections - numConPeers, withinBackoffPeers.len), MaxParalelDials)
 
     info "Relay connectivity loop",
       connectedPeers = numConPeers,
       targetConnectedPeers = maxConnections,
-      disconnectedPeers = disconnectedPeers.len,
+      notConnectedPeers = notConnectedPeers.len,
       withinBackoffPeers = withinBackoffPeers.len
 
     await pm.connectToNodes(withinBackoffPeers[0..<numPeersToConnect], WakuRelayCodec)
--- a/waku/v2/node/peer_manager/peer_manager.nim
+++ b/waku/v2/node/peer_manager/peer_manager.nim
@@ -27,21 +27,30 @@ declarePublicGauge waku_connected_peers, "Number of connected peers per directio
 logScope:
   topics = "waku node peer_manager"
 
-type
-  PeerManager* = ref object of RootObj
-    switch*: Switch
-    peerStore*: PeerStore
-    storage: PeerStorage
-
 const
   # TODO: Make configurable
   DefaultDialTimeout = chronos.seconds(10)
+
+  # Max attempts before removing the peer
+  MaxFailedAttempts = 5
+
+  # InitialBackoffInSec^failedAttempts
+  # 15s, 225s, 3375s (1h), 50625s (14h)
+  InitialBackoffInSec = 15
 
   # limit the amount of paralel dials
   MaxParalelDials = 10
 
   # delay between consecutive relayConnectivityLoop runs
   ConnectivityLoopInterval = chronos.seconds(30)
+
+type
+  PeerManager* = ref object of RootObj
+    switch*: Switch
+    peerStore*: PeerStore
+    initialBackoffInSec*: int
+    maxFailedAttempts*: int
+    storage: PeerStorage
 
 ####################
 # Helper functions #
@@ -61,45 +70,60 @@ proc insertOrReplace(ps: PeerStorage,
 proc dialPeer(pm: PeerManager, peerId: PeerID,
               addrs: seq[MultiAddress], proto: string,
               dialTimeout = DefaultDialTimeout,
-              source = "api"
+              source = "api",
               ): Future[Option[Connection]] {.async.} =
 
   # Do not attempt to dial self
   if peerId == pm.switch.peerInfo.peerId:
     return none(Connection)
 
-  info "Dialing peer from manager", wireAddr = addrs, peerId = peerId
+  let failedAttempts = pm.peerStore[NumberFailedConnBook][peerId]
+  debug "Dialing peer", wireAddr = addrs, peerId = peerId, failedAttempts=failedAttempts
 
   # Dial Peer
+  var deadline = sleepAsync(dialTimeout)
   let dialFut = pm.switch.dial(peerId, addrs, proto)
 
+  var reasonFailed = ""
   try:
-    # Attempt to dial remote peer
-    if (await dialFut.withTimeout(DefaultDialTimeout)):
+    await dialFut or deadline
+    if dialFut.finished():
+      if not deadline.finished():
+        deadline.cancel()
       waku_peers_dials.inc(labelValues = ["successful"])
       # TODO: This will be populated from the peerstore info when ready
       waku_node_conns_initiated.inc(labelValues = [source])
+      pm.peerStore[NumberFailedConnBook][peerId] = 0
       return some(dialFut.read())
     else:
-      # TODO: any redial attempts?
-      debug "Dialing remote peer timed out"
-      waku_peers_dials.inc(labelValues = ["timeout"])
+      reasonFailed = "timeout"
+      await cancelAndWait(dialFut)
+  except CatchableError as exc:
+    reasonFailed = "failed"
 
-      pm.peerStore[ConnectionBook][peerId] = CannotConnect
-      if not pm.storage.isNil:
-        pm.storage.insertOrReplace(peerId, pm.peerStore.get(peerId), CannotConnect)
+  # Dial failed
+  pm.peerStore[NumberFailedConnBook][peerId] = pm.peerStore[NumberFailedConnBook][peerId] + 1
+  pm.peerStore[LastFailedConnBook][peerId] = Moment.init(getTime().toUnix, Second)
+  pm.peerStore[ConnectionBook][peerId] = CannotConnect
 
-      return none(Connection)
-  except CatchableError as e:
-    # TODO: any redial attempts?
-    debug "Dialing remote peer failed", msg = e.msg
-    waku_peers_dials.inc(labelValues = ["failed"])
+  debug "Dialing peer failed", peerId = peerId, reason = reasonFailed, failedAttempts=failedAttempts
+  waku_peers_dials.inc(labelValues = [reasonFailed])
 
-    pm.peerStore[ConnectionBook][peerId] = CannotConnect
-    if not pm.storage.isNil:
-      pm.storage.insertOrReplace(peerId, pm.peerStore.get(peerId), CannotConnect)
+  # Update storage
+  if not pm.storage.isNil:
+    pm.storage.insertOrReplace(peerId, pm.peerStore.get(peerId), CannotConnect)
 
-    return none(Connection)
+  return none(Connection)
+
+# TODO: To be addressed in nwaku/pull/1473. Do not prune service peers
+# TODO: Currently unused
+proc prunePeerStore(pm: PeerManager) =
+  # iterate peers in peerstore
+  # skip service peers
+  #if pm.peerStore[NumberFailedConnBook][peerId] >= pm.maxFailedAttempts:
+  #  debug "Removing peer from peer store", peerId = peerId, failedAttempts=failedAttempts
+  #  pm.peerStore.del(peerId)
+  doAssert(false, "Not implemented!")
 
 proc loadFromStorage(pm: PeerManager) =
   debug "loading peers from storage"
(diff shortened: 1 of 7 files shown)


jm-clius (commented):
Although I agree with the idea of exponential backoffs, I'm not sure I agree that this should be part of (every) `dialPeer`. In my mind, if an "application" (or just a protocol) calls `dialPeer`, it should attempt to connect to that peer immediately without considering delays/backoffs (except maybe a _very_ recent failure). The protocol/application may "know better" than the peer manager about when a peer  has become available again - especially when we're in the 14h backoff period. If a protocol attempts to continuously dial an unreachable peer, there is a problem with that protocol that should be addressed. The only protocol that _currently_ requires continuous connection attempts, however, is Relay. In my mind then there should be a connectivity loop that continuously attempts to connect to some peers from the peer store and respects the backoff period for these peers _within this connectivity loop_. In this approach, `dialPeer` will be available to protocols that want to make an ad-hoc connection to a specific peer and should remain a simple way to attempt to dial the peer (there may be scope for some sanity checks here, but I don't think a protocol should attempt to dial a peer and then wait ~14 hours for the dial to be attempted).

jm-clius:
Of course, similar connectivity guarantees could be useful for service protocols in future too. I just think this connectivity attempts and backoffs should be done in parallel and not serially/ad-hoc whenever attempting to dial.

alrevuelta:
Thanks @jm-clius, fair comments. How about only respecting the backoff for the relay protocol, and having a direct path for the service protocols? Also related to "service protocol slots" https://github.com/waku-org/nwaku/issues/1461.

Some comments:

> The protocol/application may "know better" than the peer manager about when a peer has become available again

mm not sure I follow. Can you provide an example? Not sure how the application layer can have knowledge of this.

> The only protocol that currently requires continuous connection attempts, however, is Relay. In my mind then there should be a connectivity loop that continuously attempts to connect to some peers from the peer store and respects the backoff period for these peers within this connectivity loop

Nice, as stated I will implement that, and have a direct path for service protocols.

> it should attempt to connect to that peer immediately without considering delays/backoffs (except maybe a very recent failure)

Sure. Linking this to the "service protocol slots" https://github.com/waku-org/nwaku/issues/1461. I can implement n retries for "slotted" peers. If its fine will leave that for the PR fixing that issue.

alrevuelta:
Will leave this PR on hold since with the existing code, I can't differenciate between service (store, lp,...) and relay peers. Once I implement this feature https://github.com/waku-org/nwaku/issues/1461 I will be able to know which peers are "slotted" and don't apply the backoff to them.

alrevuelta:
@jm-clius Fixed your comment by adding a flag to make respecting the backoff optional. Let me know what you think :)

This feature is currently unused and will be used in https://github.com/waku-org/nwaku/issues/1477 in the `connectivity loop` that you also refer to, and only for relay peers.

waku/v2/node/peer_manager/peer_manager.nim
```diff
@@ -54,45 +64,65 @@ proc insertOrReplace(ps: PeerStorage,
...
-  info "Dialing peer from manager", wireAddr = addrs, peerId = peerId
+  let failedAttempts = pm.peerStore[NumberFailedConnBook][peerId]
+
+  # Respect a backoff period for peers that have failed to connect
+  if respectBackoff and not pm.peerStore.canBeConnected(peerId, pm.initialBackoffInSec):
+    debug "Peer recently failed to connect, skipping dial",
+          peerId = peerid,
+          failedAttempts = failedAttempts,
+          lastFailed = pm.peerStore[LastFailedConnBook][peerId]
+    return none(Connection)
```
jm-clius, line 83: Afaict this will not allow any differentiation between "still backing off" and other connection failures. Would this be something applications require? In other words, this would not/should not stop the application from attempting to dial this peer continuously as there is no indication that under the hood we're still backing off?

waku/v2/node/peer_manager/peer_manager.nim
```diff
@@ -54,45 +64,65 @@ proc insertOrReplace(ps: PeerStorage,
...
   let dialFut = pm.switch.dial(peerId, addrs, proto)
 
+  var reasonFailed = ""
   try:
-    # Attempt to dial remote peer
-    if (await dialFut.withTimeout(DefaultDialTimeout)):
+    await dialFut or deadline
+    if dialFut.finished():
+      if not deadline.finished():
+        deadline.cancel()
```
jm-clius, line 96: Any reason not to use `dialFut.withTimeout()`?

waku/v2/node/peer_manager/peer_manager.nim
```diff
@@ -54,45 +64,65 @@ proc insertOrReplace(ps: PeerStorage,
...
+  # Dial failed
+  pm.peerStore[NumberFailedConnBook][peerId] = pm.peerStore[NumberFailedConnBook][peerId] + 1
+  pm.peerStore[LastFailedConnBook][peerId] = Moment.init(getTime().toUnix, Second)
+  pm.peerStore[ConnectionBook][peerId] = CannotConnect
+
+  debug "Dialing remote failed", peerId = peerId, reason = reasonFailed, failedAttempts=failedAttempts
+  waku_peers_dials.inc(labelValues = [reasonFailed])
+
+  # If failed too many times, remove peer from peer store
+  if respectBackoff and pm.peerStore[NumberFailedConnBook][peerId] >= pm.maxFailedAttempts:
```
jm-clius, line 116: This seems a bit weird to me - the fact that me dialing some peer could result in it being removed from the peer store (and doing it again would result in it being added again, presumably?). Of course, we know that for the _application_ of "continuously attempt to connect to all available `relay` peers" this would make sense, i.e. to eventually stop attempting to dial peers that continue to fail. But this highlights my concern that this logic should not be part of the dialer, even if behind a `respectBackoff` flag.

jm-clius (commented):
Thanks! I will review in more detail after the weekend, in case I've missed some intricacies. :)

Adding the `respectBackoff` flag is indeed better, but I'm still not convinced that this logic should be part of the dialer.

As I see it, we have two distinct needs here:
1. A dialer interface that must allow any protocol/application to attempt to dial peers, maintain connectivity related peer books, etc.
2. At least one "application" of the peer manager/store (the "connect loop") that wants to use this dialer to continuously attempt to connect to all relevant peers in the peer store. This application should only attempt connection to peers that are not being backed off from, remove peers that it has the authority to do (i.e. not static peers), etc.

Mixing logic from (2) into (1) seems to me to create some confusion. What if an application wants to `respectBackoff`? Currently it will simply receive a `none(Connection)` in return if the peer is being backed off from. In my mind it has no reason not to continue attempting to connect to this peer and doesn't gain any information to help it make better decisions in future. Furthermore, the dialer will now make decisions such as removing peers from the peer store after max failed attempts. It seems to me to be doing that because it "knows" that this is what the connect loop would expect of it, since the connect loop is selecting peers from the peer store for attempted connection. This has no bearing, however, on other applications/protocols that may be managing their own peers.

alrevuelta:
@jm-clius I agree with your 1. and 2. needs. Main reason I added the `respectBackoff` to `dialPeer` and mixed the logic is so that its easier to unit test. If I respect the backoff directly in the loop, that makes it more difficult to unit test.

Plan b is to have a separate function, but that involves duplicating lots of logic: update metrics, dial + timeouts, etc.

Will convert the PR to draft and implement a vanila connectivity loop in a separate PR. Then I will add the backoff to that connectivity loop, which should comply with your "not mixing logic" requirement.

waku/v2/node/peer_manager/peer_manager.nim
alrevuelta: just reverted my changes, using again withTimeout.

i thought there was a possible race condition and wanted also to cancel the dial if the timer timed out, but noticed it didnt made much sense.

waku/v2/node/peer_manager/peer_manager.nim
alrevuelta: fixed

waku/v2/node/waku_node.nim
```diff
@@ -911,16 +906,15 @@ proc runDiscv5Loop(node: WakuNode) {.async.} =
...
-    ## Query for a random target and collect all discovered nodes
-    ## TODO: we could filter nodes here
-    let discoveredPeers = await node.wakuDiscv5.findRandomPeers()
-    if discoveredPeers.isOk():
-      ## Let's attempt to connect to peers we
-      ## have not encountered before
+    let discoveredPeersRes = await node.wakuDiscv5.findRandomPeers()
 
-      trace "Discovered peers", count=discoveredPeers.get().len()
+    if discoveredPeersRes.isOk:
```
rymnc, line 911:
```suggestion
    if discoveredPeersRes.isOk():
```
nit, but i believe this is the style guide we're adopting

rymnc: maybe we should handle an error if `findRandomPeers` fails. An error log atleast

waku/v2/node/peer_manager/waku_peer_store.nim
```diff
@@ -62,12 +67,41 @@ type
...
+  let failedAttempts = peerStore[NumberFailedConnBook][peerId]
+
+  # if it never errored, we can try to connect
+  if failedAttempts == 0:
+    return true
+
+  # If it errored we wait an exponential backoff from last connection
+  # the more failed attemps, the greater the backoff since last attempt
+  let now = Moment.init(getTime().toUnix, Second)
+  let lastFailed = peerStore[LastFailedConnBook][peerId]
```
rymnc, line 94:
```suggestion
  let lastFailed = peerStore.getLastFailedPeer(peerId)
```
or similar, which would allow us to change the underlying datastructure if required in the futute

waku/v2/node/peer_manager/peer_manager.nim
```diff
@@ -27,22 +27,34 @@ declarePublicGauge waku_connected_peers, "Number of connected peers per directio
...
   DefaultDialTimeout = chronos.seconds(10)
 
+  # Max attempts before removing the peer
+  MaxFailedAttempts = 5
+
+  # Time to wait before attempting to dial again is calculated as:
+  # initialBackoffInSec*(backoffFactor^(failedAttempts-1))
+  # 120s, 480s, 1920, 7680s
+  InitialBackoffInSec = 120
+  BackoffFactor = 4
```
rymnc, line 41: Just wondering what people's views are on making this configurable by the operator. To me, it allows for more aggressive dialing behaviour. wdyt?

rymnc (commented):
LGTM, just a couple questions :)

waku/v2/node/waku_node.nim
alrevuelta: sure thanks!

waku/v2/node/peer_manager/waku_peer_store.nim
alrevuelta: I agree with this, but I'm just trying to follow nimlibp2p peerstore https://github.com/status-im/nim-libp2p/blob/unstable/libp2p/peerstore.nim#L148  pattern, of using custom books.

Doing this will require to have 6-7 new getter functions, with just one line of code, and not sure I see the benefit right now.

But I'm totally open for suggestions.

waku/v2/node/peer_manager/peer_manager.nim
alrevuelta: mm these are some default safe values that i've tested. what do you mean by configurable? new cli flags? not sure if that would be to low level for an operator? note though that they can be changed when creating the peermanager.

waku/v2/node/peer_manager/waku_peer_store.nim
rymnc: Sounds good! just something to keep in mind if we decide to have more complex functionality later.

waku/v2/node/peer_manager/peer_manager.nim
rymnc: yeah, I mean cli flags :)

waku/v2/node/peer_manager/waku_peer_store.nim
```diff
@@ -62,12 +67,41 @@ type
...
+  let backoff = chronos.seconds(initialBackoffInSec*(backoffFactor^(failedAttempts-1)))
+  if now >= (lastFailed + backoff):
+    return true
+  return false
+
+proc delete*(peerStore: PeerStore,
```
jm-clius, line 100: Afaict, this is not yet used anywhere? Given the fact that in existing deployments peer IDs are cycled very often (I think), we should add a mechanism to manage the size of the peer store fairly urgently - unsure of the implication if this memory essentially leaks in the meantime.

jm-clius (approved):
Thanks! Makes sense to me now that the `dialPeer` mechanism does not make any decisions on whether to dial a peer or not. Some minor comments below. My biggest concern is prioritising some mechanism to manage the number of peers kept in the store to avoid leaks. Would be good to monitor peer management behaviour closely once this is merged (and auto-deployed to `wakuv2.test`)

waku/v2/node/peer_manager/peer_manager.nim
```diff
@@ -310,17 +349,18 @@ proc relayConnectivityLoop*(pm: PeerManager) {.async.} =
       await sleepAsync(ConnectivityLoopInterval)
       continue
 
-    # TODO: Respect backoff before attempting to connect a relay peer
-    var disconnectedPeers = pm.peerStore.getNotConnectedPeers().mapIt(RemotePeerInfo.init(it.peerId, it.addrs))
-    shuffle(disconnectedPeers)
-
-    let numPeersToConnect = min(min(maxConnections - numConPeers, disconnectedPeers.len), MaxParalelDials)
+    var notConnectedPeers = pm.peerStore.getNotConnectedPeers().mapIt(RemotePeerInfo.init(it.peerId, it.addrs))
+    var withinBackoffPeers = notConnectedPeers.filterIt(pm.peerStore.canBeConnected(it.peerId,
```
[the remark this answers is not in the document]
alrevuelta: ouch, leftover. will fix.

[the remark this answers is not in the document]
alrevuelta: ah right! will fix.

waku/v2/node/waku_node.nim
```diff
@@ -879,17 +879,12 @@ proc keepaliveLoop(node: WakuNode, keepalive: chronos.Duration) {.async.} =
...
     for peer in peers:
-      let connOpt = await node.peerManager.dialPeer(peer, PingCodec)
-
-      if connOpt.isNone():
-        # TODO: more sophisticated error handling here
-        debug "failed to connect to remote peer", peer=peer
+      try:
+        let conn = await node.switch.dial(peer.peerId, peer.addrs, PingCodec)
+        let pingDelay = await node.libp2pPing.ping(conn)
+      except CatchableError as exc:
```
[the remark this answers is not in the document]
alrevuelta: The main change here is to use `switch.dial` instead of `peerManager.dialPeer` for the ping. Main reason is that `dialPeer` updates metrics on ok/nok connections, failed attempts, last time failed, Can/Cannot be connected etc.

And since here we are just pinging this peer, I don't think using that function makes sense. For example, we will be updating the metrics with ok connections, every time we ping a peer.

And actually this should be more like `getConnection` because we are not dialing/connecting to any peer, but getting an already existing connection and sending a ping.

I agree with explicir error handling, but here I had to add the `try` `except` because `switch.dial` doesn't return `Result[xx]`.

waku/v2/node/peer_manager/waku_peer_store.nim
alrevuelta: Yep, not used in this PR but tracked here in "Prune peers from peerstore".
shouldn't leak since its limited by `.withPeerStore(capacity=xxx)` but yes, we have to handle it more gracefully.

LNSD (approved):
LGTM ✅

PR #1467, opened 2022-12-15
refactor(networking): create networking folder + move files

alrevuelta:
Summary:
* No changes in the code.
* Create `networking` folder in `waku/v2/node`
* Move all networking functionalities inside such as:
  * `waku_discv5`
  * `waku_dnsdisc`
  * `waku_switch`
  * `waku_peer_store`
  * `waku_peer_manager`


LNSD:
I'm sorry, but I disagree with this. 

We should go in the opposite direction: Flattening the `waku/v2/node` folder by moving the discovery mechanisms outside the `node` module (e.g. `waku/v2/waku_discv5` or `waku/v2/waku_dnsdisc`). And I would do the same with the `protocols`, but that's another discussion.

Adding more subdirectories makes the project structure more complex and harder to understand. We need to go in the direction of simplification.

alrevuelta:
> We should go in the opposite direction: Flattening the waku/v2/node folder by moving the discovery mechanisms outside the node module (e.g. waku/v2/waku_discv5 or waku/v2/waku_dnsdisc). And I would do the same with the protocols, but that's another discussion.

If `node` is the problem can we do `waku/v2/networking`?

Almost every p2p/blockchain protocol has a `networking` or `p2p` folder containing all related logic. This makes it very easy to navigate.

> Adding more subdirectories makes the project structure more complex and harder to understand. We need to go in the direction of simplification.

This PR is actually removing subdirectories and makes the project structure more simple and easier to understand (less folders). So not sure what you mean. Before/After.

image

image


wdyt?

alrevuelta:
@jm-clius mind also giving your opinion here? TLDR: move all related logic into a `networking` folder like other projects. Easier to locate stuff and not scattered across many folders.
Not very opinionated on it, so happy to go with the already established module structure if it doesn't fit.

jm-clius:
My view is a `/waku/v2/networking` subdirectory is the way to go (not nested in `/node`). It's a trade-off between simplification (moving unrelated things from the artificial `/node/`, trying to keep nested subdirectories to a minimum).

> The direction we should follow: fewer nested modules and fewer nested directories

Definitely agree, but keeping networking related stuff together is both idiomatic across p2p repos and prevents network-specific modules, which are often treated separately by maintainers and users alike, from being spread across node constructs, protocols and APIs. Agree though that other artificial nesting such as `/protocol/` etc may be revised.

alrevuelta:
Great! As long as its inside `networking` i dont mind where that folder is placed :)

LNSD:
> My view is a `/waku/v2/networking` subdirectory is the way to go (not nested in `/node`). It's a trade-off between simplification (moving unrelated things from the artificial `/node/`, trying to keep nested subdirectories to a minimum).
> 
> > The direction we should follow: fewer nested modules and fewer nested directories
> 
> Definitely agree, but keeping networking related stuff together is both idiomatic across p2p repos and prevents network-specific modules, which are often treated separately by maintainers and users alike, from being spread across node constructs, protocols and APIs. Agree though that other artificial nesting such as `/protocol/` etc may be revised.

Sorry, if I come across as too insistent. Besides that, we should not move the `wakuswitch` submodule out of the `node` module since the Waku node code ONLY uses the code.

My structure proposal is not based on an opinion. It is based on the following rationale:

> It is not the concepts like "networking" that should dictate the project module structure; it should be the code usability and maintainability.

The dependencies/imports between modules should be the constraints that we should pay attention to before structuring our code. The import graph should look like a tree. If we do not respect this, the capacity of our code to evolve will be seriously affected. The Waku RLN relay module was a clear example of a module with cyclic imports, and it was almost impossible to understand and extend. Yes, the structure had its "logic", but it did not respect the import constraints producing import cycles.

Another aspect to consider when defining a project structure is the API that a library wants to expose. A simple question: _Do we want a person that wants to interact with the store client to import a module that is inside of 3/4 module levels?_

```Nim
import waku/v2/protocols/waku_store
```

My answer: No. As a developer, I find of better quality a module that flatly exposes the API, i.e., `import waku/v2/waku_store` or `import waku/v2/peer_manager`.

I have checked different languages (not only Nim) and the way that they expose these sub-modules, and most of them follow this approach: a flat hierarchy.

alrevuelta:
I'm sure we can met halfway. imho this is important because as Hanno says is quite idiomatic to have everything in a `networking` or `p2p` folder. Examples: [charon](https://github.com/ObolNetwork/charon/tree/main/p2p), [nimbus](https://github.com/status-im/nimbus-eth2/tree/stable/beacon_chain/networking), [teku](https://github.com/ConsenSys/teku/tree/master/networking), [geth](https://github.com/ethereum/go-ethereum/tree/master/p2p), [prysm](https://github.com/prysmaticlabs/prysm/tree/develop/beacon-chain/p2p), [celestia](https://github.com/celestiaorg/celestia-core/tree/v0.34.x-celestia/p2p). So if we all acknowledge that this is what we want, we just need to think on the best way to do it, and tradeoffs. Do all of these projcets "break" your rationale? Or how are they diferent from nwaku?

LNSD:
Sorry, but I don't see here any trade-off. Even though it is "common" (I don't like the term _idiomatic_) in other projects to have a "networking" module, I think that the module import graph "straightforwardness" should prevail. Of course, we can always rework the modules in a way that meets both criteria: a straightforward imports tree and a logical module structure.

I don't intend to make my criteria prevail. I want to inform you of the consequences I see and enable you to make an informed decision, assuming all the consequences. If the criteria used in this PR is to mimic what other projects are doing, please, go ahead 😁👍🏼✅.

----

> Examples: [charon](https://github.com/ObolNetwork/charon/tree/main/p2p?rgh-link-date=2023-01-26T14%3A25%3A36Z), [nimbus](https://github.com/status-im/nimbus-eth2/tree/stable/beacon_chain/networking?rgh-link-date=2023-01-26T14%3A25%3A36Z), [teku](https://github.com/ConsenSys/teku/tree/master/networking?rgh-link-date=2023-01-26T14%3A25%3A36Z), [geth](https://github.com/ethereum/go-ethereum/tree/master/p2p?rgh-link-date=2023-01-26T14%3A25%3A36Z), [prysm](https://github.com/prysmaticlabs/prysm/tree/develop/beacon-chain/p2p?rgh-link-date=2023-01-26T14%3A25%3A36Z), [celestia](https://github.com/celestiaorg/celestia-core/tree/v0.34.x-celestia/p2p?rgh-link-date=2023-01-26T14%3A25%3A36Z).

I cannot analyze the codebase of all those projects to demonstrate how those projects do not follow the rationale I was describing. Often, good practices are ignored or bypassed (e.g., lack of experience from the developer in this aspect of software engineering, tight deadlines, etc.), making the code less readable, navigatable, understandable and, as an ultimate consequence, unmaintainable and non-extendable.

PR #1476, opened 2023-01-04
fix(rpc): return error if peer cant be connected

### LINKED ISSUE #1469: bug(rpc): connecting to a peer via RPC returns true even when failed
### Problem

Connecting to a peer via RPC returns true even when failed

### Impact
In the RPC and difficults kurtosis testing, since its how peers are currently being added.

### To reproduce

Run `nwaku`
```
./build/wakunode2 --relay=true --filter=true --lightpush=true --store=true --rpc-admin=true --keep-alive=true --max-connections=150 --dns-discovery=true --dns-discovery-url=enrtree://AOGECG2SPND25EEFMAJ5WF3KSGJNSGV356DSTL2YVLLZWIV6SAYBM@prod.nodes.status.im --discv5-discovery=true --discv5-udp-port=9005 --discv5-enr-auto-update=True --log-level=INFO --rpc-port=8545 --rpc-address=0.0.0.0 --tcp-port=30304 --metrics-server=True --metrics-server-port=8003 --metrics-server-address=0.0.0.0
```

When trying to connect to a peer via the RPC. Note `fake-peerid` is not a valid peerId:
```
curl -d '{"jsonrpc":"2.0","method":"post_waku_v2_admin_v1_peers","params":[["/ip4/43.74.128.5/tcp/60000/p2p/fake-peerid"]],"id":1}' -H 'Content-Type: application/json' 127.0.0.1:8545
```

Returns, but it didn't connect.
```
{"jsonrpc":"2.0","id":1,"result":true}
```

### Expected behavior
It returns the following when fails: `false`.
```
{"jsonrpc":"2.0","id":1,"result":false}
```

### nwaku version/commit hash
3347f8030c74d12c50204155012558238d293614

cc @AlbertoSoutullo

alrevuelta:
Closes #1469

Summary:
* When trying to connect to a i) peer not listening or ii) malformed multiaddress via the RPC, the api was returning `result:true` which is missleading since no connection was made.
* Moreover `connectToNodes` was used with a timeout but `dialPeer` already has a timeout. And note that `connectToNodes` discards the dialing result, so we can't know if the connection failed.
* This PR uses `dialPeer` instead and raises an exception if we can't connect to the peer and adds a unit test for the bug it fixes.
* Relevant for `wakurtosis` testing efforts.

### DIFF
--- a/tests/v2/test_jsonrpc_waku.nim
+++ b/tests/v2/test_jsonrpc_waku.nim
@@ -389,6 +389,15 @@ procSuite "Waku v2 JSON-RPC API":
       getRes.anyIt(it.protocol == WakuRelayCodec and
                    it.multiaddr == constructMultiaddrStr(peerInfo3))
 
+    # Verify that raises an exception if we can't connect to the peer
+    let nonExistentPeer = "/ip4/0.0.0.0/tcp/10000/p2p/16Uiu2HAm6HZZr7aToTvEBPpiys4UxajCTU97zj5v7RNR2gbniy1D"
+    expect(ValueError):
+      discard await client.post_waku_v2_admin_v1_peers(@[nonExistentPeer])
+
+    let malformedPeer = "/malformed/peer"
+    expect(ValueError):
+      discard await client.post_waku_v2_admin_v1_peers(@[malformedPeer])
+
     await server.stop()
     await server.closeWait()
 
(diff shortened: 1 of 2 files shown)


LNSD (approved):
LGTM ✅

jm-clius (approved):
Thanks!

PR #1482, opened 2023-01-10
feat(networking): add relay connectivity loop

alrevuelta:
WIP: Ignore.


waku/v2/node/peer_manager/waku_peer_store.nim
```diff
@@ -143,3 +143,6 @@ proc selectPeer*(peerStore: PeerStore, proto: string): Option[RemotePeerInfo] =
 
 proc getPeersByDirection*(peerStore: PeerStore, direction: Direction): seq[StoredInfo] =
   return peerStore.peers().filterIt(it.direction == direction)
+
+proc getDisconnectedPeers*(peerStore: PeerStore): seq[StoredInfo] =
```
jm-clius, line 147: Can we rather call this something else, like `UnconnectedPeers` or `NotConnectedPeers`. Disconnected seems a bit misleading.

apps/wakunode2/wakunode2.nim
```diff
@@ -498,6 +498,9 @@ proc startNode(node: WakuNode, conf: WakuNodeConf,
   if conf.keepAlive:
     node.startKeepalive()
 
+  # Maintain relay connections
+  node.startRelayConnectivityLoop()
```
jm-clius, line 502: Does it make sense to start this when relay is not enabled?

jm-clius (commented):
Generally makes sense to me! Would like an opportunity to revisit this tomorrow. Can we get one more reviewer? In the meantime see two minor comments below,

waku/v2/node/peer_manager/peer_manager.nim
```diff
@@ -284,3 +287,38 @@ proc connectToNodes*(pm: PeerManager,
...
+    # limit the amount of paralel dials
+    let maxParalelDials = 10
+    let numPeersToConnect = min(min(maxConnections - numConPeers, disconnectedPeers.len), maxParalelDials)
+
+    info "Relay connectivity loop",
+      connectedPeers = numConPeers,
+      targetConnectedPeers = maxConnections,
+      availableDisconnectedPeers = disconnectedPeers.len
+
+    await pm.connectToNodes(disconnectedPeers[0..<numPeersToConnect], WakuRelayCodec)
```
Menduist, line 322: This is not a good strategy imo, if every peers try to dial peers until it reaches maxConnections, you will naturally have more outgoing connections than incoming connections, and that cannot work at the network scale (every outgoing connection is someone else's incoming connection)

Ideally, every node leaves `> maxConnections / 2` open for incoming connections (also need to account for peers that are "outbound only" and are just using slots without giving any)

What is done in nimbus is that we only dial when we have to (while we are unhealthy), and once we are healthy, we leave all the remaining slots for incoming connections. This makes nimbus very nice to the network (ie on our fleet, we generally have ~30 outgoing connections and >200 incoming). But if we have to, we'll kick every one of this incoming connection to get healthy (in case of attack or whatev)

apps/wakunode2/wakunode2.nim
alrevuelta: good catch. will fix.

waku/v2/node/peer_manager/peer_manager.nim
alrevuelta: great input thanks.

with this PR i dont intend to add this in/out balance (see above "Enforce a given in/out peers ratio") but just decouple this logic from discovery without extra functionality.

but plan a follow up PR with this. Mind pointing to where this is done in nimbus? Can't immediatly see looking into `connectWorker` and `runDiscoveryLoop`

Menduist: https://github.com/status-im/nimbus-eth2/blob/f6a5a5b1da51863b2db9f812e5c682ca4eaa40ab/beacon_chain/networking/eth2_network.nim#L1562-L1563

It will not connect to anyone if we are already healthy

Not sure what the current behavior is in nwaku, but if it is similar to what is done here, it should be fixed ASAP, otherwise, as soon as there will be more than `maxConnections` peers in the network, the bootstrap nodes will be full of useless outgoing connections, and new peers won't be able to connect to them

alrevuelta: > Not sure what the current behavior is in nwaku, but if it is similar to what is done here, it should be fixed ASAP, otherwise, as soon as there will be more than maxConnections peers in the network, the bootstrap nodes will be full of useless outgoing connections, and new peers won't be able to connect to them

yep, this is the current behaviour. luckily we have more maxAmountOfConnections that nodes exists in any network. but yes, looking into it to fix it asap :)

waku/v2/node/peer_manager/waku_peer_store.nim
alrevuelta: fixed

apps/wakunode2/wakunode2.nim
alrevuelta: fixed

rymnc (approved):
LGTM

waku/v2/node/peer_manager/peer_manager.nim
alrevuelta: to be addressed here:
https://github.com/waku-org/nwaku/issues/1495

waku/v2/node/peer_manager/peer_manager.nim
```diff
@@ -284,3 +287,38 @@ proc connectToNodes*(pm: PeerManager,
...
+
+    # TODO: Enforce a given in/out peers ratio
+
+    # Leave some room for service peers
+    if numConPeers >= (maxConnections - 5):
+      await sleepAsync(defaultInterval)
+      continue
+
+    # TODO: Respect backoff before attempting to connect a relay peer
+    var disconnectedPeers = pm.peerStore.getNotConnectedPeers().mapIt(RemotePeerInfo.init(it.peerId, it.addrs))
```
jm-clius, line 310: Minor: `notConnectedPeers`

jm-clius (approved):
LGTM as an initial decoupling, thanks!

A suggestion for subsequent PRs is to extract the connectivity logic and decisions that we've made (e.g. number of slots for service peers, max parallel dials, number of peers to connect). This is not only to avoid magic numbers, but so that we can select which of these to specify in an RFC. An incremental process, of course.

waku/v2/node/peer_manager/peer_manager.nim
```diff
@@ -284,3 +287,38 @@ proc connectToNodes*(pm: PeerManager,
   # This issue was known to Dmitiry on nim-libp2p and may be resolvable
   # later.
   await sleepAsync(chronos.seconds(5))
+
+# Ensures a healthy amount of connected relay peers
+proc relayConnectivityLoop*(pm: PeerManager) {.async.} =
```
LNSD, line 292: Functions usually have "verb" names, e.g., `**get**ConnectedPeersById`. This function name has a "noun" name. Could you rename it? 

For example: if it spawns a coroutine that contains a loop, I would rename it to `spawnRelayConnectivityLoop()`.

waku/v2/node/peer_manager/peer_manager.nim
```diff
@@ -179,19 +180,21 @@ proc addPeer*(pm: PeerManager, remotePeerInfo: RemotePeerInfo, proto: string) =
...
-    pm.peerStore[AddressBook][remotePeerInfo.peerId] = pm.peerStore[AddressBook][remotePeerInfo.peerId] & multiaddr
-
   # ...public key
   var publicKey: PublicKey
   discard remotePeerInfo.peerId.extractPublicKey(publicKey)
 
+  if pm.peerStore[AddressBook][remotePeerInfo.peerId] == remotePeerInfo.addrs and
+     pm.peerStore[KeyBook][remotePeerInfo.peerId] == publicKey:
+    # Peer already managed
+    return
```
LNSD, line 190: For readability reasons, this complex if statement condition should be replaced by a `pm.addressBook.isPeerManaged()` proc returning a boolean. If you do that, the comment on line 189 will be unnecessary.

waku/v2/node/peer_manager/waku_peer_store.nim
```diff
@@ -143,3 +143,6 @@ proc selectPeer*(peerStore: PeerStore, proto: string): Option[RemotePeerInfo] =
 
 proc getPeersByDirection*(peerStore: PeerStore, direction: Direction): seq[StoredInfo] =
   return peerStore.peers().filterIt(it.direction == direction)
+
+proc getNotConnectedPeers*(peerStore: PeerStore): seq[StoredInfo] =
+  return peerStore.peers().filterIt(it.connectedness != Connected)
```
LNSD, line 148: Nit: You can remove the parenthesis from the `peers` property, as it is a noun.

```suggestion
  return peerStore.peers.filterIt(it.connectedness != Connected)
```

waku/v2/node/peer_manager/peer_manager.nim
```diff
@@ -284,3 +287,38 @@ proc connectToNodes*(pm: PeerManager,
   # This issue was known to Dmitiry on nim-libp2p and may be resolvable
   # later.
   await sleepAsync(chronos.seconds(5))
+
+# Ensures a healthy amount of connected relay peers
+proc relayConnectivityLoop*(pm: PeerManager) {.async.} =
+  let defaultInterval = chronos.seconds(30)
```
LNSD, line 293: Is this a parametrizing constant? Can you move it outside of the proc? To the top of the module with the rest of the constants.

LNSD (commented):
Please, check my comments.

waku/v2/node/peer_manager/peer_manager.nim
```diff
@@ -284,3 +287,38 @@ proc connectToNodes*(pm: PeerManager,
...
+    if numConPeers >= (maxConnections - 5):
+      await sleepAsync(defaultInterval)
+      continue
+
+    # TODO: Respect backoff before attempting to connect a relay peer
+    var disconnectedPeers = pm.peerStore.getNotConnectedPeers().mapIt(RemotePeerInfo.init(it.peerId, it.addrs))
+    shuffle(disconnectedPeers)
+
+    # limit the amount of paralel dials
+    let maxParalelDials = 10
```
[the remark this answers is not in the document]
alrevuelta: sure, done

waku/v2/node/waku_node.nim
```diff
@@ -392,6 +392,9 @@ proc startRelay*(node: WakuNode) {.async.} =
                                           protocolMatcher(WakuRelayCodec),
                                           backoffPeriod)
 
+  # Maintain relay connections
+  asyncSpawn node.peerManager.relayConnectivityLoop()
```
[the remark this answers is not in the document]
alrevuelta: yeap, agree. adding it to the todo list.

waku/v2/node/peer_manager/peer_manager.nim
```diff
@@ -284,3 +287,38 @@ proc connectToNodes*(pm: PeerManager,
...
+    let numPeersToConnect = min(min(maxConnections - numConPeers, disconnectedPeers.len), maxParalelDials)
+
+    info "Relay connectivity loop",
+      connectedPeers = numConPeers,
+      targetConnectedPeers = maxConnections,
+      availableDisconnectedPeers = disconnectedPeers.len
```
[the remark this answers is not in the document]
alrevuelta: imho this log is quite important, and its not spammy. quite common in blockchains to have this constant feedback on connected peers. Also useful to quickly see if there are any problems, lets say that i don't have enough peers, well looking into `availableDisconnectedPeers` might help. And always good to see the difference between my target amount of peers and the ones that are actually connected.

waku/v2/node/peer_manager/waku_peer_store.nim
alrevuelta: sure fixed.

waku/v2/node/peer_manager/peer_manager.nim
alrevuelta: yup thanks, done.

waku/v2/node/peer_manager/peer_manager.nim
LNSD: > Why thats not part of remotePeerInfo? well perhaps it should, but beyond this PR.

🤷🏼 

If I were working in Waku's peer management, I would modify the `RemotePeerInfo` type and add it. This is a fundamental issue that should be fixed if it helps improve code readability and reduces the cognitive load of the code.

As a temporary solution until the remote peer type includes the public key, I don't see any problem with extracting the key twice.

waku/v2/node/peer_manager/peer_manager.nim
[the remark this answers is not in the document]
LNSD: Ok.

One observation about "quite common in blockchains to have this constant feedback on connected peers.":

> Logs, exceptions, and other "out of the code execution flow" elements are part of an application's public-facing API. Good API design states that a library module, like the Waku node, should not make assumptions about this matter.

A blockchain node is an application, not a library.

PR #1494, opened 2023-01-16
fix(chat2): warn when rln relay is enabled but not compiled

rymnc:
Warns the user when chat2 has rln-relay options passed in, but was not compiled with the appropriate rln flags. Addresses https://github.com/waku-org/nwaku/issues/1201


LNSD (approved):
LGTM ✅

PR #1509, opened 2023-01-25
feat(wakunode): advertise custom multiaddresses

### LINKED ISSUE #1475: feat: ability to manually specify additional multiaddrs
### Problem
As a node operator, if I have followed the ['listening on websocket to enable connections with waku v2 browser peers ](https://github.com/waku-org/nwaku/blob/019f357dc948da11ec2eae7f6403adfb347e4dce/docs/tutorial/websocket.md) tutorial, my node will not advertise `wss` on disc.

This is a problem in all cases where SSL termination is handled upstream of the `nwaku` / `websockify` node combination, such as:

* Kubernetes deployments using ingress proxies.
* Dappnodes (my case) that use an NGINX HTTPS proxy that abstracts away SSL termination / letsencrypt / DNS config.

### Suggested solution
Provide a configuration option (CLI and environment variable) that allows for manually specifying additional `multiaddrs` that are to be advertised on disc protocols.

### Alternatives considered
It was also considered to use the `--websocket-support:true` flag for normal `ws` (albeit on port `8000`), though in this configuration, using the `light-js` client to attempt to dial the node (being behind a dappnode's HTTPS NGINX proxy, forwarded to the websocket port of `8000`), the `light-js` client failed to dial. It was determined that only with `websockify` forwarding to port `60000` on `nwaku` was I able to dial using `light-js`.

Nonetheless, this alternative would still not solve the issue of correctly advertising `wss` and `ws` on the correct IANA-specified ports.

### Acceptance criteria
For this feature to be deemed accepted, a node operator should be able to:

- [ ] Specify additional `multiaddrs` on which their node is dialable (`ws`, and `wss`).

rymnc:
Should allow the operator to set custom multiaddresses if the node's TLS termination happens upstream of nwaku (k8s ingress, dappnodes).
Fixes #1475

It is possible to set it while running wakunode2 -
```sh
./build/wakunode2 --ext-multiaddr:/ip4/200.19.20.21/tcp/60012/wss
```

image

This could allow operators to introduce bad inputs (wrong ip addresses/ports), so not sure if I should add a mechanism to test for it


LNSD:
> This could allow operators to introduce bad inputs (wrong ip addresses/ports), so not sure if I should add a mechanism to test for it

IMO, it depends on the consequences of adding a "bad multiaddress" (e.g., malformed). E.g., if it crashes the node abruptly, it should be tested and fail graciously.

apps/wakunode2/config.nim
```diff
@@ -185,6 +186,10 @@ type
       defaultValue: ""
       name: "rln-relay-cred-password" }: string
 
+    extMultiAddrs* {.
+      desc: "External multiaddresses to advertise to the network. Argument may be repeated."
+      name: "ext-multiaddr" }: seq[string]
```
LNSD, line 191: I think this option should be located after the `nat` option. Since it is related to "transport".

waku/v2/node/waku_node.nim
```diff
@@ -136,6 +136,7 @@ proc new*(T: type WakuNode,
           bindPort: Port,
           extIp = none(ValidIpAddress),
           extPort = none(Port),
+          extMultiAddrs: seq[MultiAddress] = @[],
```
LNSD, line 139:
```suggestion
          extMultiAddrs = newSeq[MultiAddress](),
```

apps/wakunode2/wakunode2.nim
```diff
@@ -249,6 +249,14 @@ proc initNode(conf: WakuNodeConf,
                 some(Port(uint16(conf.tcpPort) + conf.portsShift))
               else:
                 extTcpPort
+    extMultiAddrs = if (conf.extMultiAddrs.len > 0):
+                      let extMultiAddrsValidationRes = validateExtMultiAddrs(conf.extMultiAddrs)
+                      if extMultiAddrsValidationRes.isErr():
+                        return err("invalid external multiaddress: " & extMultiAddrsValidationRes.error)
```
LNSD, line 255: Here you are writing twice the "invalid external multiaddress": one in the validation function (returned as part of the error message) and another in this line.

I would suggest returning just the error in the validation and adding here the "invalid external multiaddress" message.

waku/v2/node/waku_node.nim
rymnc: Ah, very idiomatic, thanks!

apps/wakunode2/config.nim
rymnc: Addressed in [b21f1ed](https://github.com/waku-org/nwaku/pull/1509/commits/b21f1ed0f4126e1166c234a9b9a0893f8dc84121). I thought I would colocate the multiaddrs options together (with staticnode)

apps/wakunode2/wakunode2.nim
rymnc: Addressed in [82eff08](https://github.com/waku-org/nwaku/pull/1509/commits/82eff08658bedf1ea81474bdcf5c055f5358c349)

Removed the empty string validation too, the first check in MultiAddress.init() checks if the string starts with '/'. It covers the test case.

rymnc:
> > This could allow operators to introduce bad inputs (wrong ip addresses/ports), so not sure if I should add a mechanism to test for it
> 
> IMO, it depends on the consequences of adding a "bad multiaddress" (e.g., malformed). E.g., if it crashes the node abruptly, it should be tested and fail graciously.

It wouldn't crash the node, but advertise incorrectly to other peers.

jm-clius (commented):
Thanks for this! Will review more detailed tomorrow, but note that https://github.com/waku-org/nwaku/issues/1475 is not about announced addresses in `identify`, which I think you're changing here, but about adding manually specified `multiaddrs` to the discoverable ENR (both for DNS discovery and discv5). Doesn't mean this can't be used, just that there may be a follow-up PR before that issue can be closed :)

rymnc:
> Thanks for this! Will review more detailed tomorrow, but note that #1475 is not about announced addresses in `identify`, which I think you're changing here, but about adding manually specified `multiaddrs` to the discoverable ENR (both for DNS discovery and discv5). Doesn't mean this can't be used, just that there may be a follow-up PR before that issue can be closed :)

Was a simple fix, Addressed in [10cc174](https://github.com/waku-org/nwaku/pull/1509/commits/10cc174c83d8f047fdc9fdd625bcac8f5341f703) :)

tests/v2/test_wakunode.nim
```diff
@@ -250,4 +250,37 @@ procSuite "WakuNode":
...
+    let
+      # custom multiaddress
+      expectedMultiaddress1 = MultiAddress.init("/ip4/200.200.200.200/tcp/1234").get()
+
+    # Note: this could have been done with a single node, but it is useful to
+    # have two nodes to check that the multiaddress is advertised correctly
+    let
+      # node with custom multiaddress
+      nodeKey1 = crypto.PrivateKey.random(Secp256k1, rng[])[]
+      node1 = WakuNode.new(nodeKey1, ValidIpAddress.init("0.0.0.0"), Port(61018),
```
alrevuelta, line 265: im starting to use Port(0) which should get one among the available ones. just an fyi :)
https://github.com/waku-org/nwaku/blob/master/tests/v2/test_peer_manager.nim#L157

alrevuelta (approved):
thanks for this! lgtm!

> This could allow operators to introduce bad inputs (wrong ip addresses/ports), so not sure if I should add a mechanism to test for it

Guess this won't be new, I mean with `--nat=extip` you can also provide a wrong ip.

```
./build/wakunode2 --ext-multiaddr:/ip4/200.19.20.21/tcp/60012/ws --nat=extip:9.9.9.9
```

jm-clius (approved):
Approving, because I think subsequent steps can be left for next PRs. This still does not close https://github.com/waku-org/nwaku/issues/1475 . :D This is because the ENR we use on discv5 is different from the one you modified here. See: https://github.com/waku-org/nwaku/issues/915

Probably the way to solve that once and for all is to move wakuDiscv5 creation from the app to the node and use that `enr` for the node (unless discv5 is disabled, in which case we call `initEnr` like before). Happy to chat about a strategy to do this, as it will require more config to be available to the node during creation.

rymnc:
> Approving, because I think subsequent steps can be left for next PRs. This still does not close #1475 . :D This is because the ENR we use on discv5 is different from the one you modified here. See: #915
> 
> Probably the way to solve that once and for all is to move wakuDiscv5 creation from the app to the node and use that `enr` for the node (unless discv5 is disabled, in which case we call `initEnr` like before). Happy to chat about a strategy to do this, as it will require more config to be available to the node during creation.

Just saw https://github.com/waku-org/nwaku/blob/10cc174c83d8f047fdc9fdd625bcac8f5341f703/apps/wakunode2/wakunode2.nim#L320! 
Agree that we should move the wakuDiscv5 creation to the node

tests/v2/test_wakunode.nim
rymnc: Will address in a subsequent pr! Don't want to retrigger ci :)

PR #1512, opened 2023-01-26
feat(discv5): advertise custom multiaddresses

### LINKED ISSUE #1475: feat: ability to manually specify additional multiaddrs

rymnc:
Should fix #1475, as a follow up to https://github.com/waku-org/nwaku/pull/1509. Makes discv5 broadcast custom-set multiaddrs


waku/v2/node/waku_node.nim
```diff
@@ -148,6 +148,13 @@ proc new*(T: type WakuNode,
           nameResolver: NameResolver = nil,
           sendSignedPeerRecord = false,
           dns4DomainName = none(string),
+          discv5Enabled = false,
+          discv5TableIpLimit = none(uint),
+          discv5BucketIpLimit = none(uint),
+          discv5BitsPerHop = none(int),
```
rymnc, line 154: Hard to constrain these types more, since the upstream lib doesn't expose types like `Discv5TableIpLimit`

waku/v2/node/waku_node.nim
```diff
@@ -258,10 +265,34 @@ proc new*(T: type WakuNode,
     peerManager: PeerManager.new(switch, peerStorage),
     switch: switch,
     rng: rng,
-    enr: enr,
+    enr: populatedEnr,
     announcedAddresses: announcedAddresses
   )
 
+  if discv5Enabled:
+    var discv5Enrs: seq[enr.Record]
```
rymnc, line 273: `addBootstrapNodes` needs a mutable variable. The fn can be changed to be without side-effects in a different pr, imo

waku/v2/utils/wakuenr.nim
```diff
@@ -34,21 +34,25 @@ type
...
       maRaw = multiaddr.data.buffer # binary encoded multiaddr
       maSize = maRaw.len.uint16.toBytes(Endianness.bigEndian) # size as Big Endian unsigned 16-bit integer
-    
+
     assert maSize.len == 2
 
     fieldRaw.add(concat(@maSize, maRaw))
-  
+
+  return fieldRaw
```
rymnc, line 49: Need the raw field so that we can pass in the appropriate type to discv5 protocol setup. (for additional multiaddresses)

LNSD:
@rymnc I do not agree with moving discv5 to the waku_node module. 

I think that all the discovery mechanisms should be managed and orchestrated by the peer manager. As a consequence, these mechanisms should be:
* Either the `wakunode2` app should register the event handlers to the discovery services and handle them by the peer manager instance.
* Or dependency of the peer manager and passed into the peer manager constructor.

rymnc:
> @rymnc I do not agree with moving discv5 to the waku_node module.
> 
> I think that all the discovery mechanisms should be managed and orchestrated by the peer manager. As a consequence, these mechanisms should be:
> 
> * Either the `wakunode2` app should register the event handlers to the discovery services and handle them by the peer manager instance.
> * Or dependency of the peer manager and passed into the peer manager constructor.

I am not sure if the PeerManager was intended to have this functionality, i.e discovery. It is currently used to manage dialing, adding and removing peers.

Maybe @alrevuelta can weigh in here, was this something you intended for the PeerManager to do?

LNSD:
> I am not sure if the PeerManager was intended to have this functionality, i.e discovery. It is currently used to manage dialing, adding and removing peers.

Note the following from my comment:

> Either the wakunode2 app should register the event handlers to the discovery services and handle them by the peer manager instance.

This would be something like:

```Nim
discoveryMechanism.onPeerDiscovered do (peer: PeerId):
  peerManager.addPeer(peer)
```

tests/v2/test_waku_discv5.nim
```diff
@@ -119,3 +119,65 @@ procSuite "Waku Discovery v5":
...
+                           discv5Enabled = true,
+                           discv5UdpPort = some(nodeUdpPort1),
+                           extMultiAddrs = @[expectedMultiAddr],
+                           discv5TableIpLimit = some(10'u),
+                           discv5BucketIpLimit = some(10'u),
+                           discv5BitsPerHop = some(10),
+                           wakuFlags = some(flags))
+
+
+      nodeKey2 = crypto.PrivateKey.random(Secp256k1, rng[])[]
```
alrevuelta, line 151: minor suggestion, direction use `generateKey()`
`WakuNode.new(nodeKey=generateKey()))`

(see test_helpers.nim)

tests/v2/test_waku_discv5.nim
```diff
@@ -119,3 +119,65 @@ procSuite "Waku Discovery v5":
...
+                           bindPort = nodeTcpPort2,
+                           discv5Enabled = true,
+                           discv5UdpPort = some(nodeUdpPort2),
+                           discv5BootstrapNodes = @[node1.wakuDiscv5.protocol.localNode.record.toURI()],
+                           discv5TableIpLimit = some(10'u),
+                           discv5BucketIpLimit = some(10'u),
+                           discv5BitsPerHop = some(10),
+                           wakuFlags = some(flags))
+
+    await node1.mountRelay()
```
alrevuelta, line 167: nitpick? do we need to mount the relay? with just dicovery should be fine?

waku/v2/node/waku_node.nim
```diff
@@ -148,6 +148,13 @@ proc new*(T: type WakuNode,
           nameResolver: NameResolver = nil,
           sendSignedPeerRecord = false,
           dns4DomainName = none(string),
+          discv5Enabled = false,
+          discv5TableIpLimit = none(uint),
```
alrevuelta, line 152: perhaps 7 fields for discv5 is too much? WakuNode.new() is starting to have too many parameters.

regarding discv5TableIpLimit, discv5BucketIpLimit, discv5BitsPerHop not sure if having these configurable is too much. default values should be fine, and whatever is not default should be carefully studied. but out of scope now i guess.


Using `DiscoveryConfig` can perhaps reduce the number of fields there? and if empty defaults to `defaultDiscoveryConfig`
```
  DiscoveryConfig(
    tableIpLimits: TableIpLimits(
      tableIpLimit: tableIpLimit,
      bucketIpLimit: bucketIpLimit),
    bitsPerHop: bitsPerHop
```

waku/v2/node/waku_node.nim
```diff
@@ -258,10 +265,34 @@ proc new*(T: type WakuNode,
...
+      extIp, extPort, discv5UdpPort,
+      bindIp,
+      discv5UdpPort.get(),
+      discv5Enrs,
+      discv5EnrAutoUpdate,
+      keys.PrivateKey(nodekey.skkey),
+      wakuFlags.get(),
+      discv5EnrMultiAddrs,
+      wakuNode.rng,
+      DiscoveryConfig.init(discv5TableIpLimit.get(),
```
alrevuelta, line 292: related to my other comments can we pass directly the DiscoveryConfig? should simplify things a bit?

alrevuelta (commented):
thanks! some comments

waku/v2/node/waku_node.nim
rymnc: Addressed in [185a10a](https://github.com/waku-org/nwaku/pull/1512/commits/185a10ad6637c8c13f2e58c5d7305d290c6a3982), thanks!

waku/v2/node/waku_node.nim
rymnc: Addressed in [185a10a](https://github.com/waku-org/nwaku/pull/1512/commits/185a10ad6637c8c13f2e58c5d7305d290c6a3982), thanks!

tests/v2/test_waku_discv5.nim
rymnc: Addressed in [185a10a](https://github.com/waku-org/nwaku/pull/1512/commits/185a10ad6637c8c13f2e58c5d7305d290c6a3982), thanks!

tests/v2/test_waku_discv5.nim
rymnc: Addressed in [185a10a](https://github.com/waku-org/nwaku/pull/1512/commits/185a10ad6637c8c13f2e58c5d7305d290c6a3982), thanks!

apps/wakunode2/wakunode2.nim
```diff
@@ -286,45 +286,19 @@ proc initNode(conf: WakuNodeConf,
                         dnsResolver,
                         conf.relayPeerExchange, # We send our own signed peer record when peer exchange enabled
                         dns4DomainName,
+                        conf.discv5Discovery,
+                        DiscoveryConfig.init(conf.discv5TableIpLimit,
+                                             conf.discv5BucketIpLimit,
+                                             conf.discv5BitsPerHop),
```
jm-clius, line 292: May be more readable to assign this to a variable before calling `.new()`

waku/v2/node/waku_node.nim
```diff
@@ -148,6 +148,11 @@ proc new*(T: type WakuNode,
           nameResolver: NameResolver = nil,
           sendSignedPeerRecord = false,
           dns4DomainName = none(string),
+          discv5Enabled = false,
+          discv5Config = defaultDiscoveryConfig,
+          discv5BootstrapNodes = newSeq[string](),
+          discv5BootstrapEnrs = newSeq[enr.Record](),
```
jm-clius, line 154: Why two arguments here? Perhaps better to have only the `discv5BootstrapEnrs` as a `seq[enr.Record]` and leave it to the application to extract that from config.

waku/v2/node/waku_node.nim
```diff
@@ -258,10 +263,32 @@ proc new*(T: type WakuNode,
...
+    enr: populatedEnr,
     announcedAddresses: announcedAddresses
   )
 
+  if discv5Enabled:
+    var discv5Enrs: seq[enr.Record]
+    # parse enrURIs from the configuration and add the resulting ENRs to the discv5BootstrapEnrs seq
+    for enrUri in discv5BootstrapNodes:
+      addBootstrapNode(enrUri, discv5Enrs)
+    discv5Enrs.add(discv5BootstrapEnrs)
```
jm-clius, line 275: I think we can perhaps keep the node initialisation a bit cleaner by having this population done by the application? I.e. application provides seq of `enr.Record`.

waku/v2/node/waku_node.nim
```diff
@@ -258,10 +263,32 @@ proc new*(T: type WakuNode,
...
+
+    let rawMultiAddrField = enrMultiaddrs.getRawField()
+    # construct openArray[(string, seq[byte])]
+    let discv5EnrMultiAddrs = @[(MULTIADDR_ENR_FIELD, rawMultiAddrField)]
+    wakuNode.wakuDiscv5 = WakuDiscoveryV5.new(
```
jm-clius, line 280: Perhaps change the `WakuDiscoveryV5` constructor to rather take a `seq[MultiAddress]` so the node does not need to know about enr construction?

waku/v2/node/waku_node.nim
```diff
@@ -258,10 +263,32 @@ proc new*(T: type WakuNode,
     peerManager: PeerManager.new(switch, peerStorage),
     switch: switch,
     rng: rng,
-    enr: enr,
+    enr: populatedEnr,
```
jm-clius, line 266: This can be set to the `wakuDiscv5` enr if we have that available (in other words, move the construction to before node construction). We want to consolidate ENRs, so only need to create our own with `initEnr` if `wakuDiscv5` did not provide us with one.

jm-clius, line 266: Provide constructed `wakuDiscv5` as an argument here.

jm-clius (commented):
Thanks, this is more or less the direction I had in mind too. I've added some comments re keeping the config vs initialisation a bit cleaner and consolidating the two different ENRs into one.

waku/v2/node/waku_node.nim
rymnc: Addressed in [74c3622](https://github.com/waku-org/nwaku/pull/1512/commits/74c36223864222f0da14cee19f4a8f9b1f7f3654)

rymnc: Addressed in [74c3622](https://github.com/waku-org/nwaku/pull/1512/commits/74c36223864222f0da14cee19f4a8f9b1f7f3654)

waku/v2/node/waku_node.nim
rymnc: Addressed in [74c3622](https://github.com/waku-org/nwaku/pull/1512/commits/74c36223864222f0da14cee19f4a8f9b1f7f3654)

waku/v2/node/waku_node.nim
rymnc: Addressed in [74c3622](https://github.com/waku-org/nwaku/pull/1512/commits/74c36223864222f0da14cee19f4a8f9b1f7f3654)

waku/v2/node/waku_node.nim
rymnc: I think it is better set this way, since https://github.com/vacp2p/rfc/pull/566 refers to setting custom shard numbers. In a future PR, we can probably create an enum of `enrFieldPairs` which can be passed into the WakuDiscoveryV5 constructor, which all implement `getRawField` for use with the discv5 protocol.

```nim
type EnrFields = enum
  multiaddrs,
  shards
  
type EnrField* = object
  case key: EnrFields
  of multiaddrs:
    multiaddrs: seq[MultiAddress]
  of shards:
    shards: seq[ShardId]
  
getRawField(enrField: EnrField): seq[byte] =
  # same as getRawField implemented, but for ShardId and MultiAddress both
  if enrField.key == multiaddrs: # or a switch case
    return getRawField(enrField.multiaddrs)
  elif enrField.key == shards
    return getRawField(enrField.shards)

proc new*(T: type WakuDiscoveryV5,
          extIp: Option[ValidIpAddress],
          extTcpPort, extUdpPort: Option[Port],
          bindIP: ValidIpAddress,
          discv5UdpPort: Port,
          bootstrapNodes: seq[string],
          enrAutoUpdate = false,
          privateKey: keys.PrivateKey,
          flags: WakuEnrBitfield,
          enrFields: seq[EnrFields],
          rng: ref HmacDrbgContext,
          discv5Config: protocol.DiscoveryConfig = protocol.defaultDiscoveryConfig): T =
          var enrInitFields = @[(WAKU_ENR_FIELD, @[flags.byte])]
          for enrField in enrFields:
            let rawField = enrField.getRawField()
            enrInitFields.add((enrField.key, rawField))
          ...
```

Since the scaling solution will be put in place soon, I will not change the constructor to accept a seq[Multiaddress], wdyt?

apps/wakunode2/wakunode2.nim
rymnc: Addressed in [63878f0](https://github.com/waku-org/nwaku/pull/1512/commits/63878f0b5d3d554a36e64368ba8b9c67d495ce8d)

waku/v2/node/waku_node.nim
jm-clius: In that case I would either explicitly send the shard index and cluster number to the WakuDiscoveryV5 constructor or provide a field constructor elsewhere. The issue here is that we're mixing where the fields are initialised - `flags` for example is passed in as argument and added the enr fields internally, whereas the `multiaddrs` are not.

rymnc: Okay, I think I understand what you mean now. Addressed in [bffcca7](https://github.com/waku-org/nwaku/pull/1512/commits/bffcca70c8259ef80d2c88813d49074fef3197e6). I do think we should init the fields elsewhere in the future, though.

waku/v2/node/waku_node.nim
LNSD: Indeed, this is one of the reasons for not having a discv5 within the waku node. You configure discv5 outside, so you don't have to thread down any configuration.

waku/v2/node/waku_node.nim
```diff
@@ -12,7 +12,7 @@ import
   eth/keys,
   nimcrypto,
   bearssl/rand,
-  eth/p2p/discoveryv5/enr,
+  eth/p2p/discoveryv5/enr as enr,
```
LNSD, line 15: Isn't this redundant?

LNSD (commented):
As I said in my initial comment, I see a bad idea in moving the discovery mechanisms inside the waku_node module. 

Some comments second this idea: moving the code inside the node module means you have to thread down the configuration. But it also implies that the `WakuNode` type has more than one responsibility (besides of orchestrating the different protocols and configuring the switch it also generates the local ENR, configures the discv5 service and holds the discv5 state). So... you cannot test the `WakuNode` without the discovery instantiated.

In summary, this couples the node code with the discovery code.

----

If everybody agrees, I won't block the PR. But I think we should go in the opposite direction: splitting the content of the big modules instead of merging them into big modules.

rymnc:
Moving this PR to draft to decouple disc and `WakuNode` initialization.

jm-clius:
Agree with @LNSD here. The issue is currently our configuration and initialisation is very convoluted.
- WakuDiscv5 requires some addresses only available in WakuNode initialisation for its own initialisation
- WakuNode requires an initialised WakuDiscv5 to be constructed with a consolidated ENR

Perhaps an increment to make it a bit better is extracting those parts of initialisation to do with constructing addresses and making sure that that's available to both WakuDiscv5 and WakuNode during initialisation? That way WakuDiscv5 initialisation can be decoupled and performed before WakuNode initialisation.

For this, the constructor arguments for WakuDiscv5 may also have to change so that it can be initialised from config.

I know from previously looking at this that there are some intricacies and gotchas, so I'm okay with incrementing here in reasonable steps to something cleaner and decoupled.

alrevuelta:
thanks for addressing the comments, resolved them all.

> Maybe @alrevuelta can weigh in here, was this something you intended for the PeerManager to do?

missed this comment. I think we can keep the peer manager and discovery separated. Peer manager should be agnostic on how the peers are discovered/added to the peer manager.

rymnc:
@LNSD, I've decoupled the discv5 and waku_node setup now. Added a deprecated flag to the original `Wakunode.new()` proc in favor of the `WakuNodeAddrMeta` variant (because there are a lot of usages of `Wakunode.new()`).

LNSD (changes requested):
Please, check my comments :)

waku/v2/utils/peers.nim
```diff
@@ -182,3 +182,16 @@ proc hasProtocol*(ma: MultiAddress, proto: string): bool =
...
+  for peer in peers:
+    if peer.enr.isSome():
+      let
+        enr = peer.enr.get()
+        typedEnrRes = enr.toTypedRecord()
+      if typedEnrRes.isOk():
+        let typedEnr = typedEnrRes.get()
+        if typedEnr.udp.isSome() or typedEnr.udp6.isSome():
+          filteredEnrs.add(enr)
+  return filteredEnrs
```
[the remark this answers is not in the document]
rymnc: Addressed in 8f854af7

waku/v2/node/waku_node.nim
```diff
@@ -130,36 +130,46 @@ template wsFlag(wssEnabled: bool): MultiAddress =
...
-          nameResolver: NameResolver = nil,
-          sendSignedPeerRecord = false,
-          dns4DomainName = none(string),
-          discv5UdpPort = none(Port),
-          agentString = none(string),    # defaults to nim-libp2p version
-          peerStoreCapacity = none(int), # defaults to 1.25 maxConnections
-          ): T {.raises: [Defect, LPError, IOError, TLSStreamProtocolError].} =
-  ## Creates a Waku Node instance.
-
+type WakuNodeAddrMeta* = object
```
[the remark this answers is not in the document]
rymnc: Addressed in 70e34bae

waku/v2/node/waku_node.nim
```diff
@@ -195,26 +205,67 @@ proc new*(T: type WakuNode,
...
+    extUdpPort = addressMetadata.discv5UdpPort,
+    bindIp = addressMetadata.bindIp,
+    discv5UdpPort = addressMetadata.discv5UdpPort.get(),
+    bootstrapEnrs = discv5BootstrapEnrs,
+    enrAutoUpdate = discv5EnrAutoUpdate,
+    privateKey = keys.PrivateKey(nodekey.skkey),
+    flags = addressMetadata.wakuFlags.get(),
+    multiaddrs = addressMetadata.enrMultiaddrs,
+    rng = rng,
+    discv5Config = discv5Config)
```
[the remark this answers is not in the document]
rymnc: Addressed in 8f854af7

apps/wakunode2/wakunode2.nim
```diff
@@ -271,60 +271,53 @@ proc initNode(conf: WakuNodeConf,
...
+      discv5BootstrapEnrs.add(dynamicBootstrapEnrs)
+      let discv5Config = DiscoveryConfig.init(conf.discv5TableIpLimit,
+                                              conf.discv5BucketIpLimit,
+                                              conf.discv5BitsPerHop)
+      wakuDiscv5 = some(getWakuDiscoveryV5(addressMetadata = addressMetadata,
+                                           nodekey = conf.nodekey,
+                                           discv5Config = discv5Config,
+                                           discv5BootstrapEnrs = discv5BootstrapEnrs,
+                                           discv5EnrAutoUpdate = conf.discv5EnrAutoUpdate,
+                                           rng = rng))
```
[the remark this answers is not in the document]
rymnc: Addressed in 29a3f116

LNSD (approved):
LGTM ✅

jm-clius (approved):
Thanks! I think this is a great step forward to also clean up the confusing address configuration and initialisation. It may be worth adding a unit test suite for other aspects of `NetConfig` (e.g. populating sensible `announcedAddresses`) in future PRs.

PR #1513, opened 2023-01-26
feat(networking): prune peers from peerstore exceeding capacity

alrevuelta:
Closes https://github.com/waku-org/nwaku/issues/1504

**Summary**: Every `PrunePeerStoreInterval` the peerstore is pruned, removing peers from it if we have exceeded its capacity. The criteria is the following: peers that failed in the past are removed first, and if we are still over the capacity, not connected peers are removed.

Other changes:
* New metric to track peerstore size.
* Asserts that capacity > maxConnections, fails otherwise.

### DIFF
--- a/waku/v2/node/peer_manager/peer_manager.nim
+++ b/waku/v2/node/peer_manager/peer_manager.nim
@@ -23,6 +23,7 @@ declareCounter waku_peers_dials, "Number of peer dials", ["outcome"]
 declarePublicCounter waku_node_conns_initiated, "Number of connections initiated", ["source"]
 declarePublicGauge waku_peers_errors, "Number of peer manager errors", ["type"]
 declarePublicGauge waku_connected_peers, "Number of connected peers per direction: inbound|outbound", ["direction"]
+declarePublicGauge waku_peer_store_size, "Number of peers managed by the peer store"
 
 logScope:
   topics = "waku node peer_manager"
@@ -45,6 +46,9 @@ const
 
   # Delay between consecutive relayConnectivityLoop runs
   ConnectivityLoopInterval = chronos.seconds(30)
+
+  # How often the peer store is pruned
+  PrunePeerStoreInterval = chronos.minutes(5)
 
 type
   PeerManager* = ref object of RootObj
@@ -119,16 +123,6 @@ proc dialPeer(pm: PeerManager, peerId: PeerID,
 
   return none(Connection)
 
-# TODO: To be addressed in nwaku/pull/1473. Do not prune service peers
-# TODO: Currently unused
-proc prunePeerStore(pm: PeerManager) =
-  # iterate peers in peerstore
-  # skip service peers
-  #if pm.peerStore[NumberFailedConnBook][peerId] >= pm.maxFailedAttempts:
-  #  debug "Removing peer from peer store", peerId = peerId, failedAttempts=failedAttempts
-  #  pm.peerStore.del(peerId)
-  doAssert(false, "Not implemented!")
-
 proc loadFromStorage(pm: PeerManager) =
   debug "loading peers from storage"
   # Load peers from storage, if available
@@ -191,6 +185,14 @@ proc new*(T: type PeerManager,
           backoffFactor = BackoffFactor,
           maxFailedAttempts = MaxFailedAttempts,): PeerManager =
 
+  let capacity = switch.peerStore.capacity
+  let maxConnections = switch.connManager.inSema.size
+  if maxConnections > capacity:
+    error "Max number of connections can't be greater than PeerManager capacity",
+         capacity = capacity,
+         maxConnections = maxConnections
+    doAssert(false, "Max number of connections can't be greater than PeerManager capacity")
+
   let pm = PeerManager(switch: switch,
                        peerStore: switch.peerStore,
                        storage: storage,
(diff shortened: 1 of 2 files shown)


waku/v2/node/peer_manager/peer_manager.nim
```diff
@@ -386,6 +394,41 @@ proc relayConnectivityLoop*(pm: PeerManager) {.async.} =
...
+    for peerId in pm.peerStore[NumberFailedConnBook].book.keys:
+      if peersToPrune - prunned == 0:
+        break
+      if pm.peerStore[NumberFailedConnBook][peerId] >= pm.maxFailedAttempts:
+        pm.peerStore.del(peerId)
+        prunned += 1
+
+    # if we still need to prune, prune peers that are not connected
+    let notConnecteed = pm.peerStore.peers.filterIt(it.connectedness != Connected).mapIt(it.peerId)
+    for peerId in notConnecteed:
```
LNSD, line 419: Typo:
```suggestion
    for peerId in notConnected:
```

waku/v2/node/peer_manager/peer_manager.nim
```diff
@@ -386,6 +394,41 @@ proc relayConnectivityLoop*(pm: PeerManager) {.async.} =
 
     await sleepAsync(ConnectivityLoopInterval)
 
+proc prunePeerStore(pm: PeerManager) {.async.}  =
+  while pm.started:
+    let numPeers = toSeq(pm.peerStore[AddressBook].book.keys).len
+    let capacity = pm.peerStore.capacity
+    if numPeers < capacity:
+      await sleepAsync(PrunePeerStoreInterval)
+      continue
```
LNSD, line 403: Split things into sub-procedures so you can `return` instead of repeating the `sleepAsync(...)` code. That will make the code more readable.

The sub-procedure should be a linear function wrapped and called by the periodic execution function. Mixing the loop code with the linear code makes things more complex to understand.

waku/v2/node/peer_manager/peer_manager.nim
```diff
@@ -191,6 +185,14 @@ proc new*(T: type PeerManager,
           backoffFactor = BackoffFactor,
           maxFailedAttempts = MaxFailedAttempts,): PeerManager =
 
+  let capacity = switch.peerStore.capacity
+  let maxConnections = switch.connManager.inSema.size
+  if maxConnections > capacity:
+    error "Max number of connections can't be greater than PeerManager capacity",
+         capacity = capacity,
+         maxConnections = maxConnections
+    doAssert(false, "Max number of connections can't be greater than PeerManager capacity")
```
LNSD, line 194: Raising a `Defect` is the preferred way to finish a program in Nim. It is the `panic` equivalent of Go or Rust.

LNSD (changes requested):
Please, check my comments

waku/v2/node/peer_manager/peer_manager.nim
```diff
@@ -386,6 +394,41 @@ proc relayConnectivityLoop*(pm: PeerManager) {.async.} =
...
+    let capacity = pm.peerStore.capacity
+    if numPeers < capacity:
+      await sleepAsync(PrunePeerStoreInterval)
+      continue
+
+    debug "Peer store capacity exceeded", numPeers = numPeers, capacity = capacity
+    let peersToPrune = numPeers - capacity
+
+    # prune peers with too many failed attempts
+    var prunned = 0
```
jm-clius, line 409:
```suggestion
    var pruned = 0
```
:)

waku/v2/node/peer_manager/peer_manager.nim
```diff
@@ -386,6 +394,41 @@ proc relayConnectivityLoop*(pm: PeerManager) {.async.} =
...
+    for peerId in pm.peerStore[NumberFailedConnBook].book.keys:
+      if peersToPrune - prunned == 0:
+        break
+      if pm.peerStore[NumberFailedConnBook][peerId] >= pm.maxFailedAttempts:
+        pm.peerStore.del(peerId)
+        prunned += 1
+
+    # if we still need to prune, prune peers that are not connected
+    let notConnecteed = pm.peerStore.peers.filterIt(it.connectedness != Connected).mapIt(it.peerId)
```
jm-clius, line 418: Checking my understanding: because maxConnections <= capacity we should always be able to reach our `peersToPrune` quota?

jm-clius (approved):
LGTM.

waku/v2/node/peer_manager/peer_manager.nim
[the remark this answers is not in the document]
alrevuelta: good one thanks, fixed in 3df173c6b77edf7bcfd049ed197dffc675bccede

waku/v2/node/peer_manager/peer_manager.nim
alrevuelta: fixed thanks 3df173c6b77edf7bcfd049ed197dffc675bccede

waku/v2/node/peer_manager/peer_manager.nim
alrevuelta: indeed! should be fixed here. 3df173c6b77edf7bcfd049ed197dffc675bccede
makes it easier to unit test, added one in a94325ab694bb38d34e655a7ba01a68dec76600b

waku/v2/node/peer_manager/peer_manager.nim
alrevuelta: sure, i think i saw doAssert somewhere in libp2p and nimbus, but switched to raising Defect here.
3df173c6b77edf7bcfd049ed197dffc675bccede

added also a test case, since its important.

waku/v2/node/peer_manager/peer_manager.nim
alrevuelta: exactly!

slightly related, since enforcing this maxConnections <= capacity is quite important added a unit test for this here a94325ab694bb38d34e655a7ba01a68dec76600b, otherwise, that wont be true.

alrevuelta:
@LNSD comments were addressed, mind checking and approving if ok?

LNSD (approved):
LGTM, thanks ✅

PR #1522, opened 2023-01-31
fix(rng): use only one instance of rng

rymnc:
! WIP: DO NOT MERGE !

as per 
https://github.com/status-im/nim-libp2p/blob/c11772c94e44f0987b79d6593f34d6a4f92f07c5/libp2p/crypto/crypto.nim#L165-L167

We're only supposed to use one instance of the rng per app/library.

As per [nim-libp2p's](https://github.com/status-im/nim-libp2p/blob/4bce8f38c9b9c55c8bf0dc88cf5843662d85d78f/tests/helpers.nim#L67-L74) implementation, this may cause compilation errors on macos. This PR tests if that is the case using CI.

This PR replaces all usages of `crypto.newRng()` with `wakuCrypto.getRng()`

### DIFF
--- a/examples/v2/publisher.nim
+++ b/examples/v2/publisher.nim
@@ -11,6 +11,7 @@ import
 
 import
   ../../../waku/common/logging,
+  ../../../waku/common/crypto as wakuCrypto,
   ../../../waku/v2/node/discv5/waku_discv5,
   ../../../waku/v2/node/peer_manager/peer_manager,
   ../../../waku/v2/node/waku_node,
@@ -33,7 +34,7 @@ proc setupAndPublish() {.async.} =
     setupLogLevel(logging.LogLevel.NOTICE)
     notice "starting publisher", wakuPort=wakuPort, discv5Port=discv5Port
     let
-        nodeKey = crypto.PrivateKey.random(Secp256k1, crypto.newRng()[])[]
+        nodeKey = crypto.PrivateKey.random(Secp256k1, wakuCrypto.getRng()[])[]
         ip = ValidIpAddress.init("0.0.0.0")
         node = WakuNode.new(nodeKey, ip, Port(wakuPort))
         flags = initWakuFlags(lightpush = false, filter = false, store = false, relay = true)
--- a/waku/common/crypto.nim
+++ b/waku/common/crypto.nim
@@ -1,0 +1,3 @@
+import ./utils/crypto
+
+export crypto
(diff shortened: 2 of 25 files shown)


examples/v2/publisher.nim
```diff
@@ -33,7 +34,7 @@ proc setupAndPublish() {.async.} =
     setupLogLevel(logging.LogLevel.NOTICE)
     notice "starting publisher", wakuPort=wakuPort, discv5Port=discv5Port
     let
-        nodeKey = crypto.PrivateKey.random(Secp256k1, crypto.newRng()[])[]
+        nodeKey = crypto.PrivateKey.random(Secp256k1, wakuCrypto.getRng()[])[]
```
alrevuelta, line 37: Perhaps we can leverage the occasion to start using `generateKey()` util function and saving some lines of code? Would be great to replace all occurrences.

node = WakuNode.new(generateKey(), ip, Port(wakuPort))

alrevuelta (commented):
left an unsolicitated comment :) in case you want to levarage the occassion for this related low hanging fruit fix.

waku/common/crypto.nim
```diff
@@ -0,0 +1,3 @@
+import ./utils/crypto
+
+export crypto
```
LNSD, line 3: Please, move the utils code here. We should reduce the code inside the `utils` module to the minimum.

waku/common/utils/crypto.nim
```diff
@@ -0,0 +1,17 @@
...
+else:
+  {.push raises: [].}
+
+import libp2p/crypto/crypto as libp2pCrypto
+
+export libp2pCrypto
+
+type RngCtx = ref HmacDrbgContext
+
+var rng {.threadvar.}: RngCtx
```
LNSD, line 12: These "module variables" are often described as code smell. This is Nim's equivalent for the C/C++ global variables.

We should instantiate the RNG in the application top-level proc (in the `wakunode2` `main()` proc) and then pass it down as a constructor dependency argument. Similar to all the dependencies we create (e.g., Waku Archive drivers, Waku node, Peer manager, etc.).

-------

I recommend [this book](https://www.manning.com/books/dependency-injection-principles-practices-patterns) to understand the Dependency Injection patterns and good practices, and more importantly, the antipatterns that should be avoided. It is a .Net book, but the important part is the principles.

LNSD (changes requested):
Please, check my comments.

rymnc:
Closing this PR for revision.

PR #1523, opened 2023-01-31
fix(rng): create rng only within isMainModule blocks

rymnc:
As a follow up to #1522, this method is more surgical, and does not make use of global variables (sort of).

Makes the rng required by all the procs which consume it, except for WakuNode.new(), because refactoring this will require updating a lot of tests, which can be done in a follow up PR.

### DIFF
--- a/apps/chat2/chat2.nim
+++ b/apps/chat2/chat2.nim
@@ -370,10 +370,10 @@ proc readInput(wfd: AsyncFD) {.thread, raises: [Defect, CatchableError].} =
     discard waitFor transp.write(line & "\r\n")
 
 {.pop.} # @TODO confutils.nim(775, 17) Error: can raise an unlisted exception: ref IOError
-proc processInput(rfd: AsyncFD) {.async.} =
+proc processInput(rfd: AsyncFD, rng: ref HmacDrbgContext) {.async.} =
   let
     transp = fromPipe(rfd)
-    conf = Chat2Conf.load()
+    conf = Chat2Conf.load(rng)
 
   # set log level
   if conf.logLevel != LogLevel.NONE:
@@ -593,7 +593,7 @@ proc processInput(rfd: AsyncFD) {.async.} =
 
   runForever()
 
-proc main() {.async.} =
+proc main(rng: ref HmacDrbgContext) {.async.} =
   let (rfd, wfd) = createAsyncPipe()
   if rfd == asyncInvalidPipe or wfd == asyncInvalidPipe:
     raise newException(ValueError, "Could not initialize pipe!")
@@ -601,7 +601,7 @@ proc main() {.async.} =
   var thread: Thread[AsyncFD]
   thread.createThread(readInput, wfd)
   try:
-    await processInput(rfd)
+    await processInput(rfd, rng)
   # Handle only ConfigurationError for now
   # TODO: Throw other errors from the mounting procedure
   except ConfigurationError as e:
@@ -609,8 +609,9 @@ proc main() {.async.} =
 
 
 when isMainModule: # isMainModule = true when the module is compiled as the main file
+  let rng = crypto.newRng()
   try:
-    waitFor(main())
+    waitFor(main(rng))
   except CatchableError as e:
     raise e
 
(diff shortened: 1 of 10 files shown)


apps/chat2/chat2.nim
```diff
@@ -370,10 +370,10 @@ proc readInput(wfd: AsyncFD) {.thread, raises: [Defect, CatchableError].} =
     discard waitFor transp.write(line & "\r\n")
 
 {.pop.} # @TODO confutils.nim(775, 17) Error: can raise an unlisted exception: ref IOError
-proc processInput(rfd: AsyncFD) {.async.} =
+proc processInput(rfd: AsyncFD, rng: ref HmacDrbgContext) {.async.} =
```
rymnc, line 373: ideally this type would be aliased to something like `RngCtx`, but currently it does not justify a new file and import everywhere.

apps/wakunode2/config.nim
```diff
@@ -53,7 +53,6 @@ type
 
     nodekey* {.
       desc: "P2P node private key as 64 char hex string.",
-      defaultValue: defaultPrivateKey()
       name: "nodekey" }: PrivateKey
```
LNSD, line 56: If we are not providing a default value, then let's use a `Option[PrivateKey]`:
```suggestion
      name: "nodekey" }: Option[PrivateKey]
```

apps/wakunode2/config.nim
```diff
@@ -543,16 +539,19 @@ proc readValue*(r: var EnvvarReader, value: var crypto.PrivateKey) {.raises: [Se
...
-    let conf = WakuNodeConf.load(
+    var conf = WakuNodeConf.load(
       version=version,
       secondarySources = proc (conf: WakuNodeConf, sources: auto) =
         sources.addConfigFile(Envvar, InputFile("wakunode2"))
 
         if conf.configFile.isSome():
           sources.addConfigFile(Toml, conf.configFile.get())
     )
+    if $conf.nodekey == "":
```
LNSD, line 552: Then check if the value is `None`.

apps/wakunode2/wakunode2.nim
```diff
@@ -215,7 +215,8 @@ proc retrieveDynamicBootstrapNodes*(dnsDiscovery: bool, dnsDiscoveryUrl: string,
 
 proc initNode(conf: WakuNodeConf,
               peerStore: Option[WakuPeerStorage],
-              dynamicBootstrapNodes: openArray[RemotePeerInfo] = @[]): SetupResult[WakuNode] =
+              dynamicBootstrapNodes: openArray[RemotePeerInfo] = @[],
+              rng: ref HmacDrbgContext): SetupResult[WakuNode] =
```
LNSD, line 219: As this parameter is mandatory, it should be located before the parameters with default values.

Doing that, you will avoid passing the RNG to this proc.

apps/wakunode2/config.nim
```diff
@@ -543,16 +539,19 @@ proc readValue*(r: var EnvvarReader, value: var crypto.PrivateKey) {.raises: [Se
...
+      conf.nodekey = crypto.PrivateKey.random(Secp256k1, rng[]).tryGet()
```
LNSD, line 553: This default value assignment should be in the wakunode2 module, like the CLI argument validations.

LNSD (approved):
Please, check my comments 🙂

apps/wakunode2/config.nim
rymnc: Do you mean, after reading the config, we check if nodekey is `none`, and then assign a random one? 

Something like this? 

```nim
let nodekey = if conf.nodekey.isSome(): conf.nodekey.get()
              else: crypto.PrivateKey.random(Secp256k1, rng[]).tryGet()
```

apps/wakunode2/config.nim
rymnc: Addressed in 1b9566c8

apps/wakunode2/config.nim
rymnc: Addressed in 1b9566c8

apps/wakunode2/wakunode2.nim
rymnc: Addressed in 1b9566c8

apps/wakunode2/config.nim
rymnc: Addressed in 1b9566c8. Please let me know if this is what you thought of.

rymnc: Thanks!

apps/chat2/chat2.nim
LNSD: I thought the same 👍🏼

LNSD (approved):
LGTM, thanks ✅ 

P.S.: Note that there are conflicts after other PRs were merged.

PR #1524, opened 2023-01-31
docs: mention cloning of submodules in README.md

felicio:


README.md
```diff
@@ -27,6 +27,13 @@ detailed instructions, see Waku v1 and v2 home above.
 
 More information on the installation of these can be found [here](https://github.com/status-im/nimbus#prerequisites).
 
+### Clone
+
+```bash
+# Ensure submodules are cloned as well
+git clone --recurse-submodules <REPOSITORY_REMOTE_URL>
```
alrevuelta, line 34: Thanks, indeed missing. Can we add it right before `make wakunode1 wakunode2` in "Wakunode" section?

`make update` should e fine instead.

LNSD:
AFAIK this is not necessary, the makefile already does it for you if you run `make update` for the first time.

Here: https://github.com/waku-org/nwaku/blob/master/Makefile#L17-L26

LNSD (changes requested):
AFAIK this is not necessary, the makefile already does it for you if you run `make update` for the first time.

You can clone the repo normally, and then run `make update`. The submodules will be cloned recursively.

----

The makefile relevant part: https://github.com/waku-org/nwaku/blob/master/Makefile#L17-L26

README.md
alrevuelta: yup as @LNSD I think we should just add `make update`

alrevuelta: @felicio Mind adding `make update`, so we can merge this? Thanks!

PR #1526, opened 2023-02-01
chore: add better e2e trace logging for relay

jm-clius:
Hopefully an easy fix for https://github.com/waku-org/nwaku/issues/1394

Improves relay trace logging on publishing and receiving a message, to log:

- a hash of the message
- publish/arrival timestamp
- pubsubTopic

@Daimakaimura my understanding is that this will cover your needs? Anything else that would be required/nice to have?

This will result in the following example trace log on the relaying node:

```
TRC 2023-02-01 14:48:52.270+02:00 waku.relay received                        topics="waku node" tid=22467 file=waku_node.nim:310 pubsubTopic=/waku/2/default-waku/proto hash=0x1220b9cb650d6af10c34d9d882a609d9aa6d506b0831f8f01a8e27be3c665d889974 receivedTime=1675255732000000000
```

and on the publishing node

```
TRC 2023-02-01 14:48:52.270+02:00 waku.relay published                        topics="waku node" tid=22467 file=waku_node.nim:310 pubsubTopic=/waku/2/default-waku/proto hash=0x1220b9cb650d6af10c34d9d882a609d9aa6d506b0831f8f01a8e27be3c665d889974 publishTime=1675255732000000000
```

## Note:

- log format and hash selected to be roughly similar to go-waku: https://github.com/waku-org/go-waku/commit/a247e8346d458ce3d4ac6726ae98fb7a2428c5d1
- hash computation currently done inside trace log statement (to skip compilation when trace logs not enabled), should probably replaced by the message UID concept being developed

### DIFF
--- a/waku/v2/node/waku_node.nim
+++ b/waku/v2/node/waku_node.nim
@@ -38,6 +38,7 @@ import
   ../protocol/waku_peer_exchange,
   ../utils/peers,
   ../utils/wakuenr,
+  ../utils/time,
   ./peer_manager/peer_manager,
   ./dnsdisc/waku_dnsdisc,
   ./discv5/waku_discv5,
@@ -300,13 +301,16 @@ proc subscribe(node: WakuNode, topic: PubsubTopic, handler: Option[TopicHandler]
 
   proc defaultHandler(topic: string, data: seq[byte]) {.async, gcsafe.} =
     # A default handler should be registered for all topics
-    trace "Hit default handler", topic=topic, data=data
 
     let msg = WakuMessage.decode(data)
     if msg.isErr():
       # TODO: Add metric to track waku message decode errors
       return
 
+    trace "waku.relay received",
+      pubsubTopic=topic,
+      hash=MultiHash.digest("sha2-256", data).expect("valid hash").data.buffer.to0xHex(), # TODO: this could be replaced by a message UID
+      receivedTime=getNowInNanosecondTime()
 
     # Notify mounted protocols of new message
     if not node.wakuFilter.isNil():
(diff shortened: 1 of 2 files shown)


jm-clius:
@alrevuelta Indeed, I did consider this. We can either separate this logging out completely with a compiler flag, e.g. `waku_expensive_logging` (and log on `INFO`) or enable only the `waku node` topic TRACE logging (which should be clean) with a compiler flag (`-d:chronicles_enabled_topics`). Both would require changes to the compilation. It is not an option to compile and log all `TRACE` logs, as you've said.

LNSD:
From discord conversation: 

> There is this chronicles option that could be used to generate a special build that only logs certain topics with TRACE level: https://github.com/status-im/nim-chronicles#chronicles_enabled_topics

alrevuelta:
Perhaps for another PR, but do we plan to compile and push a new docker image for this variant? I would say it would facilitate things for the DS team, but unsure about starting to push many variants of nwaku, can end up being very convoluted.

@Daimakaimura

jm-clius:
Of course happy to discuss what is practical with the DS team, but I think it's reasonable to have a different compilation and docker image for this. I see this special logging use case basically as a specialised application of Nwaku. I would avoid compiling this logging into our usual image and then only enable this "application" via some runtime configuration. We can address during next steps. The idea now is to get consensus on adding those logs.

LNSD:
Nobody said anything about pushing any docker image to docker-hub. The image could be built by docker/kurtosis before running any tests. A docker build argument can be added to the current Dockefile to support a perf-test image variant.

While I understand the concerns, please, let's focus the discussion on the information to be displayed in the log message.

jm-clius:
From https://github.com/waku-org/nwaku/issues/1394#issuecomment-1413364319

> I checked the PR Draft and adding those 2 messages to the logs should be okay for us to calculate delays. Regarding the volume in the logs, indeed that could be also another issue and probably the best option would likely to have a flag to enable only these messages (independently of the log level).

Leaving configuration and compilation TBD, but having gotten the 👍 from DS opening for review and merging.

PR #1527, opened 2023-02-01
refactor(px): refactor peer exchange + tests

alrevuelta:
Last item to close https://github.com/waku-org/nwaku/issues/1463

Some refactoring in the px protocol + more tests + moving the "what to do with the peers we got via px" logic to the application. Some discussions started in https://github.com/waku-org/nwaku/issues/1463 and this prepares the ground for https://github.com/waku-org/nwaku/issues/1521.

**Summary of changes**:
* Improve error handling using `Result` and catch exceptions (mainly `readLp` and `writeLp` since can raise an Exception if the stream is prematurely closed)
* No longer handle the px response on the protocol. The response should be handled by the application, which now adds the peers to the peerstore intead of trying to connect to them.
* Changes `respond` proc, reusing the existing connection instead of attempting to create a new one.
* Fixes possible vulnerability [here](https://github.com/waku-org/nwaku/blob/master/waku/v2/protocol/waku_peer_exchange/protocol.nim#L170-L187), since one could inject unsolicitatede responses to a node that never requested px. In other words, you could send a `PeerExchangeResponse` to a px node without that node ever requesting anything from you.
* Add new tests for all `request` proc variants both succesfull and not.

Test:

- [x] Once approved manually test it.

## End to end test

Node that discovers peers and runs peer exchange:
```
./build/wakunode2 \
--dns-discovery=true \
--dns-discovery-url=enrtree://AOGECG2SPND25EEFMAJ5WF3KSGJNSGV356DSTL2YVLLZWIV6SAYBM@test.waku.nodes.status.im \
--discv5-discovery=true \
--discv5-enr-auto-update=True \
--nodekey=161c2836197e8918d7eb1039758423795f0dc697a018ed55d5931aa9b0e5efb3 \
--peer-exchange
```

Node that sets the above one as px node
```
./build/wakunode2 \
--tcp-port=30305 \
--ports-shift=1 \
--peer-exchange-node=/ip4/127.0.0.1/tcp/60000/p2p/16Uiu2HAm2ZuRHGZUzG6JGJeY1nGKHXLGrQYnZ7mZToymHtFfYFFf
```

Peers are succesfuly retrieved in the second node and added to the peerstore.

### DIFF
--- a/apps/wakunode2/wakunode2.nim
+++ b/apps/wakunode2/wakunode2.nim
@@ -529,10 +529,13 @@ proc startNode(node: WakuNode, conf: WakuNodeConf,
   if conf.peerExchangeNode != "":
     info "Retrieving peer info via peer exchange protocol"
     let desiredOutDegree = node.wakuRelay.parameters.d.uint64()
-    try:
-      discard await node.wakuPeerExchange.request(desiredOutDegree)
-    except:
-      return err("failed to retrieve peer info via peer exchange protocol: " & getCurrentExceptionMsg())
+    let pxPeersRes = await node.wakuPeerExchange.request(desiredOutDegree)
+    if pxPeersRes.isOk:
+      let pxPeers = pxPeersRes.get().peerInfos
+      echo "----pxPeers, ", pxPeers
+    else:
+      #error is tosevere?
+      warn "failed to retrieve peer info via peer exchange protocol"
 
   # Start keepalive, if enabled
   if conf.keepAlive:
--- a/apps/wakunode2/wakunode2.nim
+++ b/apps/wakunode2/wakunode2.nim
@@ -526,21 +526,10 @@ proc startNode(node: WakuNode, conf: WakuNodeConf,
   if conf.peerExchange:
     asyncSpawn runPeerExchangeDiscv5Loop(node.wakuPeerExchange)
 
-  # retrieve and connect to peer exchange peers
+  # retrieve px peers and add the to the peer store
   if conf.peerExchangeNode != "":
-    info "Retrieving peer info via peer exchange protocol"
     let desiredOutDegree = node.wakuRelay.parameters.d.uint64()
-    let pxPeersRes = await node.wakuPeerExchange.request(desiredOutDegree)
-    if pxPeersRes.isOk:
-      var record: enr.Record
-      var validPeers = 0
-      for pi in pxPeersRes.get().peerInfos:
-        if enr.fromBytes(record, pi.enr):
-          node.peerManager.addPeer(record.toRemotePeerInfo().get, WakuRelayCodec)
-          validPeers += 1
-      info "Retrieved peer info via peer exchange protocol", validPeers = validPeers
-    else:
-      warn "Failed to retrieve peer info via peer exchange protocol"
+    await node.fetchPeerExchangePeers(desiredOutDegree)
 
   # Start keepalive, if enabled
   if conf.keepAlive:
--- a/apps/wakunode2/wakunode2.nim
+++ b/apps/wakunode2/wakunode2.nim
@@ -41,6 +41,7 @@ import
   ../../waku/v2/protocol/waku_archive/retention_policy/retention_policy_capacity,
   ../../waku/v2/protocol/waku_archive/retention_policy/retention_policy_time,
   ../../waku/v2/protocol/waku_store,
+  ../../waku/v2/protocol/waku_relay,
   ../../waku/v2/protocol/waku_filter,
   ../../waku/v2/protocol/waku_lightpush,
   ../../waku/v2/protocol/waku_peer_exchange,
@@ -531,11 +532,15 @@ proc startNode(node: WakuNode, conf: WakuNodeConf,
     let desiredOutDegree = node.wakuRelay.parameters.d.uint64()
     let pxPeersRes = await node.wakuPeerExchange.request(desiredOutDegree)
     if pxPeersRes.isOk:
-      let pxPeers = pxPeersRes.get().peerInfos
-      echo "----pxPeers, ", pxPeers
+      var record: enr.Record
+      var validPeers = 0
+      for pi in pxPeersRes.get().peerInfos:
+        if enr.fromBytes(record, pi.enr):
+          node.peerManager.addPeer(record.toRemotePeerInfo().get, WakuRelayCodec)
+          validPeers += 1
+      info "Retrieved peer info via peer exchange protocol", validPeers = validPeers
     else:
-      #error is tosevere?
-      warn "failed to retrieve peer info via peer exchange protocol"
+      warn "Failed to retrieve peer info via peer exchange protocol"
 
   # Start keepalive, if enabled
   if conf.keepAlive:
--- a/waku/v2/protocol/waku_peer_exchange/protocol.nim
+++ b/waku/v2/protocol/waku_peer_exchange/protocol.nim
@@ -142,22 +142,19 @@ proc initProtocolHandler(wpx: WakuPeerExchange) =
       waku_px_errors.inc(labelValues = [exc.msg])
       return
 
-    let res = PeerExchangeRpc.decode(buffer)
-    if res.isErr():
+    let decBuf = PeerExchangeRpc.decode(buffer)
+    if decBuf.isErr():
       waku_px_errors.inc(labelValues = [decodeRpcFailure])
       return
 
-    let rpc = res.get()
-
-    # If we got a request (request field is not empty)
-    if rpc.request != default(PeerExchangeRequest):
-      trace "peer exchange request received"
-      let enrs = wpx.getEnrsFromCache(rpc.request.numPeers)
-      let res = await wpx.respond(enrs, conn)
-      if res.isErr:
-        waku_px_errors.inc(labelValues = [res.error])
-      else:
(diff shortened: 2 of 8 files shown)


waku/v2/protocol/waku_peer_exchange/protocol.nim
```diff
@@ -147,44 +130,34 @@ proc getEnrsFromCache(wpx: WakuPeerExchange, numPeers: uint64): seq[enr.Record]
...
     if res.isErr():
       waku_px_errors.inc(labelValues = [decodeRpcFailure])
       return
 
     let rpc = res.get()
 
-    # handle peer exchange request
-    if rpc.request != PeerExchangeRequest():
+    # If we got a request (request field is not empty)
+    if rpc.request != default(PeerExchangeRequest):
```
LNSD, line 153: While I know this is in other places in the code, we should avoid doing it. This is a defect in this protocol's RPC codec and RPC types: 

- If the field is mandatory, return a "required field missing error"
- If the field is optional, the codec should return an `Option`; here, you should check if it is present (or not) and act in consequence.

LNSD (changes requested):
Waku peer-exchange wire formats must be added to the [vacp2p/waku](https://github.com/vacp2p/waku) repository.

Please add them before proceeding with modifying the implementation. Flaws in the wire format and the RPC types should be fixed before moving forward.

waku/v2/protocol/waku_peer_exchange/protocol.nim
alrevuelta: Yep, I agree. But since the rpc doesn't allow us to add an error message, this would involve changing the px proto, changing the spec, and perhaps bumping the px version? All of this beyond the PR.

alrevuelta (commented):
> Waku peer-exchange wire formats must be added to the [vacp2p/waku](https://github.com/vacp2p/waku) repository.

> Please add them before proceeding with modifying the implementation. Flaws in the wire format and the RPC types should be fixed before moving forward.

Sure, wont consider it a blocker for this, but this should cover it:
https://github.com/vacp2p/waku/pull/12
https://github.com/vacp2p/rfc/pull/570

apps/wakunode2/wakunode2.nim
```diff
@@ -513,10 +514,17 @@ proc startNode(node: WakuNode, conf: WakuNodeConf,
...
+    if pxPeersRes.isOk:
+      var record: enr.Record
+      var validPeers = 0
+      for pi in pxPeersRes.get().peerInfos:
+        if enr.fromBytes(record, pi.enr):
+          node.peerManager.addPeer(record.toRemotePeerInfo().get, WakuRelayCodec)
+          validPeers += 1
+      info "Retrieved peer info via peer exchange protocol", validPeers = validPeers
+    else:
+      warn "Failed to retrieve peer info via peer exchange protocol"
```
jm-clius, line 527: This is a bit weird for me. While I agree with removing this from the `protocol` module, why not into `waku_node` when set up as a peer exchange client? It may make sense during node `init` to call some function that sets up peer exchange as a client and for now does only a single request. Basically we init and start all other discovery mechanisms when initiating the `WakuNode`, so this belongs there IMO.

waku/v2/protocol/waku_peer_exchange/protocol.nim
```diff
@@ -49,70 +49,53 @@ type
   WakuPeerExchange* = ref object of LPProtocol
     peerManager*: PeerManager
     wakuDiscv5: Option[WakuDiscoveryV5]
-    enrCache: seq[enr.Record] # todo: next step: ring buffer; future: implement cache satisfying https://rfc.vac.dev/spec/34/
+    enrCache*: seq[enr.Record] # todo: next step: ring buffer; future: implement cache satisfying https://rfc.vac.dev/spec/34/
 
-proc sendPeerExchangeRpcToPeer(wpx: WakuPeerExchange, rpc: PeerExchangeRpc, peer: RemotePeerInfo | PeerId): Future[WakuPeerExchangeResult[void]] {.async, gcsafe.} =
+proc request*(wpx: WakuPeerExchange, numPeers: uint64, conn: Connection): Future[WakuPeerExchangeResult[PeerExchangeResponse]] {.async, gcsafe.} =
```
jm-clius, line 54: Why not split this (and `respond*`) into a `client` module?

jm-clius (changes requested):
Agree with the general approach of:
- decoupling peer exchange response from connecting to peers
- what to do with peers in peer exchange response not being a concern of the protocol but the "application", but think we can have `WakuNode` be setup as a basic peer exchange client application: e.g. calling request once on init, similar to DNS lookup being done once when configured.

apps/wakunode2/wakunode2.nim
alrevuelta: yes indeed, not sure what i was thinking. Addressed in https://github.com/waku-org/nwaku/pull/1527/commits/9083938d51fe615c47f08a37f12b1271832cd273

Added also a new test which also showcases how this is expected to be used.

https://github.com/waku-org/nwaku/blob/9083938d51fe615c47f08a37f12b1271832cd273/tests/v2/test_wakunode.nim#L293-L317

waku/v2/protocol/waku_peer_exchange/protocol.nim
alrevuelta: mm, since the px protocol is quite simple, i would prefer to keep everything in the same file if its fine. easier to navigate.

jm-clius (approved):
LGTM! Thanks. I'm still uneasy about the wakunode2 performing the API call, but I think this is clearly beyond the scope of this PR. :)

alrevuelta:
@LNSD I addresses your comments (rfc update and proto updates) which are already merged, and replied. Is there anything left? Would like to merge this today unless you need more time.

waku/v2/protocol/waku_peer_exchange/protocol.nim
LNSD: Honestly, I am with @jm-clius. 

Keeping everything in the same module opens the doors for the code to be modified in a way that makes it convoluted and, in the end, non-extensible and unmaintainable.

waku/v2/protocol/waku_peer_exchange/protocol.nim
LNSD: This is not correct. 

I am not asking to change the protocol. I am concerned about the implementation and the way this equality comparison uses the `default(...)` to detect if the field was missing:

```Nim
    if rpc.request != default(PeerExchangeRequest):
```

You can return the [this "required field missing" error](https://github.com/waku-org/nwaku/blob/master/waku/v2/protocol/waku_store/rpc_codec.nim#L63) from the `rpc_codec` the same way other protocols do, and this check can be removed from the protocol handler because if the field is not present, the message will be considered invalid.

This doesn't require any protocol modification. It is an implementation detail, and it should be in the scope of the PR given that this PR aims to improve the implementation.

waku/v2/protocol/waku_peer_exchange/protocol.nim
```diff
@@ -147,44 +130,34 @@ proc getEnrsFromCache(wpx: WakuPeerExchange, numPeers: uint64): seq[enr.Record]
...
 
 proc initProtocolHandler(wpx: WakuPeerExchange) =
   proc handler(conn: Connection, proto: string) {.async, gcsafe, closure.} =
-    let buff = await conn.readLp(MaxRpcSize.int)
+    var buffer: seq[byte]
+    try:
+      buffer = await conn.readLp(MaxRpcSize.int)
+    except CatchableError as exc:
+      waku_px_errors.inc(labelValues = [exc.msg])
+      return
```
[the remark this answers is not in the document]
LNSD: The first time I hear about that 🤷🏼 

I am curious. Where do you get that conclusion?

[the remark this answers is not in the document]
alrevuelta: Something related to this. Happens if the connection is closed. `Stream Remotely Closed!`
https://github.com/waku-org/nwaku/blob/1af3a0d656c309aeae2851bf0058a729a74c475f/tests/v2/test_waku_peer_exchange.nim#L235-L255

waku/v2/protocol/waku_peer_exchange/protocol.nim
alrevuelta: Should be fixed in 7cb54339fe20cf991a9816a62e6447b94ae13212

> and it should be in the scope of the PR

note thought that the initial goal of this PR was to close https://github.com/waku-org/nwaku/issues/1463 and some refactoring was done along the way, but the priority was to decouple discovery from connection.

alrevuelta:
@LNSD thanks, addresses the changes.

LNSD (approved):
I see some details that do not follow the agreed style guide, but you can merge it if you want ✅

waku/v2/protocol/waku_peer_exchange/protocol.nim
jm-clius: As a general note, PRs can and should have a limited scope. This PR had a clear focus listed in the description and addressed those items.

Not approving a PR generally means that there is a critical bug/regression/technical debt _introduced_ (not existing prior to PR) or a critical question which needs answering. Technical debt which existed before the PR can be pointed out, but to speed up PR process should not stop approval.

PR #1529, opened 2023-02-02
fix: improve filter test race condition

jm-clius:
Partially addresses https://github.com/waku-org/nwaku/issues/1357

Currently the most common cause for test failures in the `CI` pipeline, seems to be related to filter tests.

This is caused by a race condition between a subscribing client and a publishing filter service node. In each case I looked at, the race condition was a matter of ~100ms or less, so extra time padding should improve this.

Longer term this race condition should be avoided by using [the improved filter specification](https://github.com/waku-org/nwaku/issues/1357) and waiting for the subscribe transaction to be complete.

NB: I think .editorconfig changed quite a lot on this one! Please ignore whitespace to see the diff.

### DIFF
--- a/tests/v2/test_waku_filter.nim
+++ b/tests/v2/test_waku_filter.nim
@@ -112,7 +112,7 @@ suite "Waku Filter":
     require (await client.subscribe(pubsubTopic, contentTopic, pushHandler, peer=serverAddr)).isOk()
 
     # WARN: Sleep necessary to avoid a race condition between the subscription and the handle message proc
-    await sleepAsync(1.seconds)
+    await sleepAsync(500.milliseconds)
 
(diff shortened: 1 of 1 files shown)


rymnc (approved):
LGTM

alrevuelta (approved):
lgtm! thanks!

did all ci checks passed on the first run :O?

LNSD (approved):
> Longer term this race condition should be avoided by using https://github.com/waku-org/nwaku/issues/1357 and waiting for the subscribe transaction to be complete.

LGTM 💯 

Indeed, we need to review the test strategy after the new filter version is implemented.

jm-clius:
> did all ci checks passed on the first run :O?

Very first run was cancelled as I pushed a new version. But this version did!

PR #1536, opened 2023-02-06
refactor(utils): moved noise/compat payload encryption code to utils

LNSD:
The `waku_payload` is an odd module inside the `node` module. This is an intermediate and temporary refactoring to isolate the backward compatibility code.

- [x] Split the `waku_payload` module into two `utils` submodules: `noise` and `compat` (whisper and waku v1 payload encryption compatibility).
- [x] Renamed the `test_waku_payload` test suite to `test_utils_compat`.
- [x] Updated all the imports accordingly.

This is another gardening PR 🪴, part of the upcoming MUID implementation.

### DIFF
--- a/apps/chat2/chat2.nim
+++ b/apps/chat2/chat2.nim
@@ -27,17 +27,21 @@ import
   ../../waku/v2/protocol/waku_lightpush/rpc,
   ../../waku/v2/protocol/waku_filter,
   ../../waku/v2/protocol/waku_store,
-  ../../waku/v2/node/[waku_node, waku_payload, waku_metrics],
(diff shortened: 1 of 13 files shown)


rymnc (approved):
LGTM!

alrevuelta (approved):
lgtm!

wondering though if the new `waku/v2/utils/noise.nim` file should go in `waku_noise` folder?

LNSD:
I was thinking of moving it inside, but as far as I know `waku_noise` is agnostic to the `waku_message` types.

That's why I moved it "temporarily" to utils folder.

PR #1537, opened 2023-02-06
feat: add WakuMessage validation in gossipsub

alrevuelta:
**Summary**
* Add new libp2p validator to `WakuRelay`, so that messages that are not `WakuRelay` (aka fail to decode into pb) are rejected and not further gossiped.
* Add unit test testing this new feature.

### DIFF
--- a/examples/v2/publisher.nim
+++ b/examples/v2/publisher.nim
@@ -53,6 +53,7 @@ proc setupAndPublish(rng: ref HmacDrbgContext) {.async.} =
 
     await node.start()
     await node.mountRelay()
+    node.peerManager.start()
     if not await node.startDiscv5():
       error "failed to start discv5"
       quit(1)
--- a/examples/v2/subscriber.nim
+++ b/examples/v2/subscriber.nim
@@ -49,6 +49,7 @@ proc setupAndSubscribe(rng: ref HmacDrbgContext) {.async.} =
 
     await node.start()
     await node.mountRelay()
+    node.peerManager.start()
     if not await node.startDiscv5():
       error "failed to start discv5"
       quit(1)
--- a/tests/v2/test_wakunode_relay.nim
+++ b/tests/v2/test_wakunode_relay.nim
@@ -1,12 +1,12 @@
 {.used.}
 
 import
-  std/os,
+  std/[os,sysrand,sequtils,tables,math],
   stew/byteutils,
   stew/shims/net as stewNet,
   testutils/unittests,
-  chronicles, 
-  chronos, 
+  chronicles,
+  chronos,
   libp2p/crypto/crypto,
   libp2p/crypto/secp,
   libp2p/peerid,
@@ -19,8 +19,11 @@ import
   ../../waku/v2/protocol/waku_message,
   ../../waku/v2/node/peer_manager,
   ../../waku/v2/utils/peers,
-  ../../waku/v2/node/waku_node
-
+  ../../waku/v2/node/waku_node,
+  ../../waku/v2/protocol/waku_relay,
+  ../test_helpers,
+  ./testlib/common
+  #./testlib/testutils
 
 template sourceDir: string = currentSourcePath.parentDir()
 const KEY_PATH = sourceDir / "resources/test_key.pem"
@@ -28,7 +31,7 @@ const CERT_PATH = sourceDir / "resources/test_cert.pem"
 
 procSuite "WakuNode - Relay":
   let rng = crypto.newRng()
-  
+
   asyncTest "Relay protocol is started correctly":
     let
       nodeKey1 = crypto.PrivateKey.random(Secp256k1, rng[])[]
@@ -61,7 +64,7 @@ procSuite "WakuNode - Relay":
       GossipSub(node2.wakuRelay).heartbeatFut.isNil == false
 
     await allFutures([node1.stop(), node2.stop()])
-  
+
   asyncTest "Messages are correctly relayed":
     let
       nodeKey1 = crypto.PrivateKey.random(Secp256k1, rng[])[]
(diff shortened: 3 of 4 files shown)


LNSD:
What is the validator chaining strategy? Have you thought about it?

alrevuelta:
> What is the validator chaining strategy?

Not sure what you mean, mind elaborating?

LNSD:
> > What is the validator chaining strategy?
> 
> Not sure what you mean, mind elaborating?

There are several things to consider when talking about Waku Relay validators:

- Which validators are we going to add? (I can think about three at least: message decoding, message integrity and the RLN validator).
- How do we plan to add those validators? Per topic? Per content topic? To all topics?
- Which order should the validators' execution follow? Which validator should be the first?

Validators act as middleware in an HTTP server. The order is important, as well as what is done inside of each validator.

alrevuelta:
> What is the validator chaining strategy? Have you thought about it?

Perhaps out of scope for this PR but now that you bring that up, these were my initial thoughts applying to this PR:
* message decoding (aka checking that the bytes can be decoded into a `WakuMessage`) should be added to all gossipsub/content topics and the first check to be executed.

And then beyond this:
* RLN validation already exists, but I see it as somewhat "opinionated" so would leave it aside (experimental as it is right now).
* There is a validation that i see us implementing in the short term, related to [this](https://github.com/vacp2p/research/issues/164), that prevents spaming. Some kind of validation that maps the message to a given identifier that i) protects us from sybils and ii) that we can use to rate limit. So if we see that this given identifier sent more than x bytes in a given time, then we reject them.
* Regarding "message integrity", we can add some check as well here. Perhaps you have some input since I guess it will depend on your message uid changes?
* Another check we could add is timestamps. So not relay messages with +- xx time to current timestamp. Run some DoS tests last week and was getting echoes 1 hour later, quite odd, will write more about these findings.
* Regarding to which gossipsub topics we should add the validation, some of this will depend on ongoing work in sharding. But imho all gossipsub/content topics should have some non opinionated validation, like the existing in this PR and some other to come.

LNSD:
> * There is a validation that i see us implementing in the short term, related to [this](https://github.com/vacp2p/research/issues/164), that prevents spaming. Some kind of validation that maps the message to a given identifier that i) protects us from sybils and ii) that we can use to rate limit. So if we see that this given identifier sent more than x bytes in a given time, then we reject them.
> * Regarding "message integrity", we can add some check as well here. Perhaps you have some input since I guess it will depend on your message uid changes?

Indeed, this is what I was about to say: I am working on the Message UID. 

This PR overlaps with the work I am doing. I already went through all the Gossipsub implementations to understand the best validation strategy for the MUID. And have this validator as part of an upcoming PR.

alrevuelta:
> This PR overlaps with the work I am doing. I already went through all the Gossipsub implementations to understand the best validation strategy for the MUID. And have this validator as part of an upcoming PR.

Great! I dont plan to add more changes to this PR, so feel free to add the MUID validation when its ready.

So at some point the validation pipeline would be:
`decode-message -> validate-muid -> some-tobedefined-dos-protection`

LNSD:
Let me take over the Waku Relay protocol.

I have a couple of refactoring PRs waiting to be pushed (and to be reviewed) that fix the current implementation issues it has. And after that I will add the MUID.

alrevuelta:
> Let me take over the Waku Relay protocol.

This will need to be a joint effort since for DoS mitigation I would need to do changes in the relay. See "Simple Waku Relay DoS mitigation" from [here](https://github.com/waku-org/pm/issues/12#issuecomment-1403577619) and [here](https://github.com/vacp2p/research/issues/164). Most likely all of them will be related to just message validation.

Ofc happy to review the PRs and leaving to you everything related to MUID validation and other changes you may have.

jm-clius:
In general I would suggest keeping our validation strategy as simple as possible.
The GossipSub validations that are currently roadmapped that I know of:
1. Short term: Application-level Validation that requires a message being validated to have originated within a specific Community (this would cover, but go further, than the message integrity use case and requires key material being provided by each community). This is being worked on by @alrevuelta, @kaiserd and the SeM team.
2. Longer term: RLN validation. Being worked on by several people in the SeM team.

(1) and (2) are preferably mutually exclusive (i.e. you run only the one and not the other type of validation), so may not require validator chaining.

jm-clius (approved):
I take it this is an increment towards app-level validation? Or do we believe there's an existing problem with invalid `WakuMessage`s being passed around?

fryorcraken:
In this approach, the message is validated if there either `payload` or `contentTopic` are not present on the wire:

https://github.com/waku-org/nwaku/blob/1995bbec6cc8dc798ce8a8aee3845c68586759fa/waku/v2/protocol/waku_message/codec.nim#L41

However, in `proto3`, for a standard field (`optional` not present), then the absence of the field should be interpreted as the default value being set.

Which means that nwaku rejects messages with:
- content topic as an empty string
- OR payload is an empty byte array

For example, ping relays message have an empty payload: they would be rejected.

I think this behaviour is fine but I would suggest that we make it explicit with possibly an RFC update.

fryorcraken:
From Discord:

In nwaku decoding and application logic are mixed up in nwaku: application error is thrown if protobuf field missing from wire.
nwaku makes a difference between field is absent from wire (error thrown) and field is set to default value (no error thrown) which is not as per the proto3 language spec for standard fields.

In go-waku and nwaku, there is no tri-state on these fields (because no `optional` argument) and hence absent value and default value cannot be differentiated.
Which is why if we want to error on absent value than we also have to error on default value.

Note that the RFC says that the fields MUST be present. But again, this statement is no different from the fields MUST NOT HAVE default value from a proto3 POV.

Two ways forward:
1. Makes it explicit in the spec that the fields MUST NOT HAVE default value.
2. OR, use optional on these fields to enable different behaviour between absent values and default values.

PR #1541, opened 2023-02-07
refactor(message): split waku message module into rpc and codec

LNSD:
Another gardening PR 🪴

- [x] Split module into `codec` and `types` modules.


alrevuelta (commented):
Understand that leaving `waku_message.nim` is to import `codec` and `message` in one go, but I think having and almost empty `waku_message.nim` and then `waku_message` folder muds the repo.

I was fine with having everything into `waku_message.nim`, but just an opinion.

other than that, lgtm.

jm-clius (approved):
LGTM

LNSD:
> I was fine with having everything into `waku_message.nim`, but just an opinion.

I think this is necessary to include here the MUID module. If I add the MUID inside the `waku_message`, we'll end up with a big file. 

In a follow-up PR, I will add more code to this module.

PR #1543, opened 2023-02-08
refactor(jsonrpc): deep code and tests reorganization

LNSD:
For a long time, the JSON-RPC was a little "abandoned". I have given it a little love and refactored all the code here. Although it is planned to be deprecated and be superseded by the Rest API, some nwaku stakeholders are actively using it (e.g., Logos perf-tests, etc.).

- [x] Changed code organization. The different APIs share the same structure (modules, types, clients, etc.)
- [x] The Private API has been merged with the Waku Relay API since both have a clear dependency.
- [x] The Waku Relay and the Waku Filter APIs now use the `message_cache` (LRU cache) data structure.
- [x] Tests have been split into different test suites (matching the `node/jsonrpc` submodules).

The follow-up PR will do the same with the Rest API code.


alrevuelta (changes requested):
Mind elaborating on why having 8 new files? 4x `callsigs.nim` and 4x`client.nim`. I would remove all 8 since it adds a substantial amount of files that muds code navigation.

I also liked having a single `jsonrpc_callsigs.nim` with all the function prototypes, being able to see all in one look is a good to have imho. Just an opinion.

Perhaps instead of creating one folder per api type, we could have a single `api` folder with `admin_api.nim`, `relay_api.nim`, `store_api.nim` inside. With this we differenciate api code than utils and other things. Also, just an idea, but with the goal in mind of reducing the amount of files we have, specially if they are almost empty with boilerplate stuff.

Really like splitting `test_jsonrpc_waku.nim` into multiple `admin/debug/filter/relay/store` as you did!

LNSD:
> Mind elaborating on why having 8 new files? 4x `callsigs.nim` and 4x`client.nim`. I would remove all 8 since it adds a substantial amount of files that muds code navigation.
> 
> I also liked having a single `jsonrpc_callsigs.nim` with all the function prototypes, being able to see all in one look is a good to have imho. Just an opinion.
> 
> Perhaps instead of creating one folder per api type, we could have a single `api` folder with `admin_api.nim`, `relay_api.nim`, `store_api.nim` inside. With this we differenciate api code than utils and other things. Also, just an idea, but with the goal in mind of reducing the amount of files we have, specially if they are almost empty with boilerplate stuff.
> 
> Really like splitting `test_jsonrpc_waku.nim` into multiple `admin/debug/filter/relay/store` as you did!

TLDR: For maintenance and extensibility reasons: it decouples the code. 

Fewer files do not necessarily mean better code. All the code is contained inside a module. If you are extending an API, all your code should be inside that module. 

This helps enormously with code maintenance and extensibility, and as a consequence with PRs. If one decides to use code that shouldn't be used, like an API-specific function, we'll notice because it will have to **import explicitly** this function. If someone claims to extend a certain API but is modifying a different module, this will be evident.

LNSD:
And I almost forgot the most important reason! 

> If we are all working on the same file, conflicts in PRs will be unavoidable.

jm-clius (approved):
Thanks! Indeed, the JSON-RPC API has been neglected for a while.

I'm not necessarily convinced that splitting each API module into 3/4 submodules adds _enough_ value to justify the directory explosion, especially since we plan on deprecating this and don't expect much parallel work/maintenance on each API. The principle is sound, though, that things which functionally belong together should be located together and functionality should be separated to keep maintenance, expansion and import paths and use cases (e.g. client vs server) clean. Too many files, on the other hand, can start obscuring the abstracted big picture. This may just require a balance between being purist and pragmatic, which may be different for different people, so leaving the decision to you.

rymnc (approved):
LGTM, thanks, personally, I find it easier to understand the json-rpc api now, and looks extensible for future protocols

LNSD:
> This may just require a balance between being purist and pragmatic, which may be different for different people, so leaving the decision to you.

Well, my intention is far from puristic. It's more pragmatic (e.g., avoid conflict in shared files like the types). but I suppose that you got a different impression 😅 

Honestly, I would merge it like it is for two main reasons:

- I have some work that already depends on this refactoring (part of the relay refactoring). Changing this will mean reworking those things.
- While we aim to sunset the JSON-RPC API, people are already using it. Some conversations are pending: the nwaku node API design (what we should expose through Nim API and what through the Rest API) and the JSON-RPC deprecation plan (now that people are using it).

I take notes so that I won't introduce so many files next time I do a refactoring.

LNSD:
Said that, waiting for @alrevuelta re-review.

jm-clius:
>  my intention is far from puristic. It's more pragmatic

Hahaha, I guess depending on POV these two terms can mean anything, so they're not very helpful. You're reasoning makes sense, so I'm happy for this to be merged.

alrevuelta (approved):
lgtm then, feel free to go ahead and merge.

I understand your reasons but just to give an example: See this: https://github.com/status-im/nimbus-eth2/tree/stable/beacon_chain/rpc

one folder, all rpc files in there, with meaningul names and very easy to understand and reason about. ofc this is a rest api and not a jsonrpc.
image

So I'm fine with having the jsonrpc as you suggest, but whenever we continue developing the rest one, I would like having our rest similar to nimbus one. Also a nice to have following similar structures across status projects imho. But yep no strictly related to this, just saying :)

Issue #1545, opened 2023-02-09
bug: subscribe to default pubsub topic even when not configured


jm-clius:
### Problem

Nwaku will always subscribe to the default pubsub topic `/waku/2/default-waku/proto` even when explicitly configured not to.

### Impact

This prevents nwaku nodes unable/unwilling to participate on the default pubsub topic (with high bandwidth requirements) from subscribing only to a different pubsub topic of their choice. This is a current requirement for developers: https://github.com/waku-org/pm/issues/13

### To reproduce

`./build/wakunode2 --topics:/waku/2/dev-waku/proto`

Note logs. Node subscribes to both `/waku/2/dev-waku/proto` and (incorrectly) the default `waku/2/default-waku/proto`.

### Expected behavior

`./build/wakunode2` (no arguments): subscribe to default pubsub topic `waku/2/default-waku/proto`.
`./build/wakunode2 --topics:/waku/2/dev-waku/proto`: subscribe _only_ to the configured topic `/waku/2/dev-waku/proto`
`./build/wakunode2 --topics:"/waku/2/default-waku/proto /waku/2/dev-waku/proto"`: subscribe to both configured topics, namely `waku/2/default-waku/proto` and `/waku/2/dev-waku/proto`.

Note this config item itself can be revisited to perhaps be a repeated item rather than space separated list, but this falls beyond the scope of fixing this bug.

### nwaku version/commit hash

nwaku `v0.14.0` and earlier.

PR #1547, opened 2023-02-10
refactor(relay): improve waku_relay apis and add tests

LNSD:
This is an API cleanup PR for the `waku_relay` module. The module was leaking implementation details all around (e.g., access to fields of the parent object directly), amongst other deficiencies.

- [x] Added APIs exposing the functionality accessed through calling the parent object.
	+ Added a `addValidator` procedure.
    + Added a `isSubscribed` and `subscribedTopics` procedures (in a previous PR)
- [x] Removed the `defaultPubsubTopis` state from within the `WakuRelay` instance.
- [x] Fixed a missing import issue.
- [x] Added a test suite that does not depend on the `waku_node` module (see: `test_waku_relay.nim`).
- [x] Updated tests accordingly.

> **Note** 
> The validator added in a previous PR has been temporarily commented out and will be added back once a clear validator implementation strategy is defined (in the code).


waku/v2/protocol/waku_rln_relay/utils.nim
```diff
@@ -1026,9 +1025,11 @@ proc new*(T: type WakuRlnRelay,
...
   if wakuRelay.isNil():
     return err("WakuRelay protocol is not mounted")
-  # check whether the pubsub topic is supported at the relay level
-  if conf.rlnRelayPubsubTopic notin wakuRelay.defaultPubsubTopics:
-    return err("The relay protocol does not support the configured pubsub topic")
+
+  # TODO: Review this. The Waku Relay instance is no longer keeping track of the default pubsub topics
+  # # check whether the pubsub topic is supported at the relay level
+  # if conf.rlnRelayPubsubTopic notin wakuRelay.defaultPubsubTopics:
+  #   return err("The relay protocol does not support the configured pubsub topic")
```
LNSD, line 1032: @rymnc I need your input here. We no longer keep the "default pub-sub topics" as part of the `WakuRelay` instance. Would you like me to remove it?

waku/v2/protocol/waku_relay.nim
```diff
@@ -78,15 +77,12 @@ method initPubSub(w: WakuRelay) {.raises: [InitializationError].} =
...
-          switch: Switch,
-          defaultPubsubTopics: seq[PubsubTopic] = @[],
-          triggerSelf: bool = true): WakuRelayResult[T] =
+proc new*(T: type WakuRelay, switch: Switch, triggerSelf: bool = true): WakuRelayResult[T] =
 
   proc msgIdProvider(msg: messages.Message): Result[MessageID, ValidationResult] =
     let hash = MultiHash.digest("sha2-256", msg.data)
     if hash.isErr():
-      ok(($msg.data.hash).toBytes())
+      ok(toBytes($hashes.hash(msg.data)))
```
LNSD, line 85: I have no idea how the `waku_relay` module was compiled 🤯  

The `std/hashes` import was missing for this line. It took me 2 days to find the right import that fixed this line's compilation error.

waku/v2/node/waku_node.nim
```diff
@@ -553,6 +548,13 @@ proc mountRelay*(node: WakuNode,
 
   info "relay mounted successfully"
 
+  # TODO: As part of #1545, remove this and update the tests cases
+  node.subscribe(DefaultPubsubTopic)
```
LNSD, line 552: This line will be removed as part of a follow-up PR that will resolve #1545

tests/v2/waku_relay/test_wakunode_relay.nim
```diff
@@ -208,7 +213,8 @@ procSuite "WakuNode - Relay":
 
     await allFutures(node1.stop(), node2.stop(), node3.stop())
 
-  asyncTest "Stats of peer sending wrong WakuMessages are updated":
+  # TODO: Add a function to validate the WakuMessage integrity
+  xasyncTest "Stats of peer sending wrong WakuMessages are updated":
```
LNSD, line 217: This test is disabled because the recently added validator has been commented out. Waiting to come up with a solution to re-enable it.

waku/v2/protocol/waku_rln_relay/utils.nim
rymnc: Yes, that works. As long as we can assert that the rln relay pubsub topic belongs to the configured pubsub topics, it should be ok

waku/v2/protocol/waku_relay.nim
```diff
@@ -100,25 +96,28 @@ proc new*(T: type WakuRelay,
...
+    # # Rejects messages that are not WakuMessage
+    # proc validator(topic: string, message: messages.Message): Future[ValidationResult] {.async.} =
+    #   let msg = WakuMessage.decode(message.data)
+    #   if msg.isOk():
+    #     return ValidationResult.Accept
+    #   return ValidationResult.Reject
+
+    # # Add validator to all default pubsub topics
+    # for pubSubTopic in defaultPubsubTopics:
+    #   wr.addValidator(pubSubTopic, validator)
```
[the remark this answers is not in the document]
alrevuelta: I still don't see the problem with this. imho its better an existing feature that solves a real problem (with room to improve) than a non existing feature.

The initial concerns were
> The Waku message integrity validator is only being applied to the default pub-sub topics ([here](https://github.com/waku-org/nwaku/blob/master/waku/v2/protocol/waku_relay.nim#L105-L113)) since the validator is only applied at the WakuRelay instance creation.

Dont see a problem here, was the intention of the PR.

> If any user subscribes to a pub-sub topic via the subscribe API ([here](https://github.com/waku-org/nwaku/blob/master/waku/v2/protocol/waku_relay.nim#L138-L151)) (e.g., via the JSON-RPC API), the message integrity validator won't be added to that topic, leaving those topics out of the claimed "DoS protection".

Also, dont see a problem. Having some validation is better than none at all. If api subs are not validated, then we have to add that, but not remove all validations. Better some than none.

> As they are executed on every message's arrival, validators must be fast and efficient to not impact the message latency and overall system performance

Sure, we will have two decodings, one for validating and one for the actual decoding. Can be improved, yes. But aren't we overoptimizing? Whats even worse is wrong messages being gossiped in the network. This do imply way more validations. For inspiration we can use [nimbus](https://github.com/status-im/nimbus-eth2/blob/ba7c0bc091161f261148dbfcb27b21cc48f1fdf8/beacon_chain/networking/eth2_network.nim#L2423).

**tldr**: i would leave this validation, as its better having some than none. unless im missing something i dont see anything critical here.

alrevuelta (changes requested):

waku/v2/node/waku_node.nim
jm-clius: Thanks!

jm-clius (approved):
LGTM.

waku/v2/protocol/waku_relay.nim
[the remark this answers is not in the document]
LNSD: I am not against the "claimed" DoS protection. I said that it was not the moment of merging it. And the way it was implemented, had flaws.

I commented it out because it uses the `defaultPubsubTopics` field. After this PR, this field is no longer available. So, the snippet cannot work. This will be included in one of the upcoming PRs: **once the validator implementation strategy is defined (in code)**.

LNSD:
@alrevuelta I have answered your comment. If you don't have any other concerns besides the commented-out code snippet, I will merge this PR.

alrevuelta:
> I commented it out because it uses the defaultPubsubTopics field. After this PR, this field is no longer available. So, the snippet cannot work.

That's fine, but when modifying master one has to adapt to the changes that others have made. In order to make your changes compatible, the `validator` should be added to DefaultPubsubTopic + the ones configured by `topics` flag just like it was before. That would be my suggestion. My future work is based on this.

As others have approved, I won't block this PR but don't think we should remove functionalities without a clear reason. I see no clear flaw and a functionality is being removed here. We should follow an incremental approach.

This validator thingy already caused multiple backs and forth, so I don't have anything else to add.

LNSD:
> This validator thingy already caused multiple backs and forth, so I don't have anything else to add.

You rushed and merged #1537. I told you to allow me to implement this in a proper way what you wanted to add in this comment: https://github.com/waku-org/nwaku/pull/1537#issuecomment-1420474361

-----

The validator will be merged sooner than you think, but... technical debt :)

image

PR #1552, opened 2023-02-13
fix(tests): do not specify port number

LNSD:
This PR should help mitigate the _"LPError: Address already in use"_ issues reported in #1357.

From [this](https://discord.com/channels/864066763682218004/865466680129880074/1052594544400678983) conversation at discord: 
> FWIW. The issue could be solved by allowing the Waku node/switch to be configured with a `/ip4/0.0.0.0/tcp/0` (then it would look for a free TCP/UDP port and bind to it). Then each test case would use its own TCP port, avoiding the "port reuse".

### DIFF
--- a/tests/v2/test_peer_exchange.nim
+++ b/tests/v2/test_peer_exchange.nim
@@ -22,11 +22,11 @@ procSuite "Peer Exchange":
     let
       bindIp = ValidIpAddress.init("0.0.0.0")
       nodeKey1 = generateSecp256k1Key()
-      node1 = WakuNode.new(nodeKey1, bindIp, Port(61100))
+      node1 = WakuNode.new(nodeKey1, bindIp, Port(0))
       nodeKey2 = generateSecp256k1Key()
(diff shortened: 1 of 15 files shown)


alrevuelta (approved):
lgtm! thanks!

(ingnore this if CI is ok)
found in the past some cases where the supplied port was 0 and some tests were failing because the assert was (==0) but in the multiaddress what was being stored was a "real" available port (eg 2000) so 0!=2000.

rymnc (approved):
Thanks for this! LGTM if ci passes in the first go :)

LNSD:
> found in the past some cases where the supplied port was 0 and some tests were failing because the assert was (==0) but in the multiaddress what was being stored was a "real" available port (eg 2000) so 0!=2000.

I explicitly decided to not modify those test cases exactly for that reason. Those test cases will have to be fixed in a different piece of work.

LNSD:
Passing ✅ 🎸

PR #1553, opened 2023-02-13
fix: remove unused imports

LNSD:
Another 🪴 PR. Removed the unused import warnings (except the `chronicles` one).

### DIFF
--- a/apps/wakubridge/wakubridge.nim
+++ b/apps/wakubridge/wakubridge.nim
@@ -4,7 +4,7 @@ else:
   {.push raises: [].}
 
 import
-  std/[os, tables, hashes, sequtils],
+  std/[tables, hashes, sequtils],
   stew/byteutils,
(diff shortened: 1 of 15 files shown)


LNSD:
It seems that some `peer_manager` test cases are failing intermittently: https://github.com/waku-org/nwaku/actions/runs/4163008348/jobs/7202871447#step:5:5917

cc #1357

alrevuelta:
> It seems that some peer_manager test cases are failing intermittently: https://github.com/waku-org/nwaku/actions/runs/4163008348/jobs/7202871447#step:5:5917

mm will check. seems to be the `prunePeerStore` test.

alrevuelta (approved):
lgtm thanks!

double checked that tests can be executed individually, since sometimes the unused warning lies a bit (like the one you mention chronicles). but all good :)

PR #1556, opened 2023-02-13
fix: change message ID provider to match go-waku

jm-clius:
The existing `messageIdProvider` in nwaku uses `Multihash` format, while in go-waku straightforward `sha256` hashes are used. This implies that message IDs (e.g. in the seen cache, in control messages, etc.) do not match between go-waku and nwaku, resulting in a lot of duplication and unnecessary control messages.

Since nwaku is more easily upgraded than go-waku, this quickfix changes nwaku message IDs to match those of go-waku.

A longer term solution is to specify a hashing scheme in an RFC.

### DIFF
--- a/waku/v2/protocol/waku_relay.nim
+++ b/waku/v2/protocol/waku_relay.nim
@@ -8,7 +8,7 @@ else:
   {.push raises: [].}
 
 import
-  std/[tables, sequtils, hashes],
+  std/[tables, sequtils],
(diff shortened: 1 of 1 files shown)


richard-ramos (approved):

LNSD (approved):
Thanks! LGTM ✅

Menduist (approved):

PR #1560, opened 2023-02-14
fix(networking): fix wrong peer connected state

alrevuelta:
* Fixes an important issue, causing the connectedness state of the peers to be stored wrongly `Connected`/`NotConnected`. This had further implications, since the connected state was wrong, we were trying to connect to already connected peers, leading to `Max streams for protocol reached` error after 10 streams were created.
* Also closes https://github.com/waku-org/nwaku/issues/1503
* TLDR: We can't use `ConnEventHandler` to track the peer connected state, we should use `PeerEventHandler` instead. This
* Connection state cant be tracked using `ConnEventHandler` when more than one conn is allowed and when using autonat (which we currently are). For example, if a peer has 2 conns and one get disconnected we cant assume that the peer is disconnected, so it doesnt make sense to update the state in this handler. The other con might still be active. So `PeerEventHandler` should be used instead.

### DIFF
--- a/waku/v2/node/peer_manager/peer_manager.nim
+++ b/waku/v2/node/peer_manager/peer_manager.nim
@@ -45,7 +45,7 @@ const
   MaxParalelDials = 10
 
   # Delay between consecutive relayConnectivityLoop runs
-  ConnectivityLoopInterval = chronos.seconds(30)
+  ConnectivityLoopInterval = chronos.seconds(15)
 
   # How often the peer store is pruned
   PrunePeerStoreInterval = chronos.minutes(5)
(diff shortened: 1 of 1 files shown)


waku/v2/node/peer_manager/peer_manager.nim
```diff
@@ -45,7 +45,7 @@ const
   MaxParalelDials = 10
 
   # Delay between consecutive relayConnectivityLoop runs
-  ConnectivityLoopInterval = chronos.seconds(30)
+  ConnectivityLoopInterval = chronos.seconds(15)
```
jm-clius, line 48: Any particular reason for this change?

jm-clius (approved):

waku/v2/node/peer_manager/peer_manager.nim
alrevuelta: yep, its intended. 30 seconds was to much.
since right now we have the backoff, it doesnt hurt to run it more often, since we no longer spam connections

PR #1561, opened 2023-02-14
chore: lower per-message log level to trace

jm-clius:
On nodes where the relay RPC cache is not polled (e.g. fleet nodes), the cache will log on every message. This log must therefore be `TRACE` and not `DEBUG`.


LNSD (approved):
LGTM ✅

PR #1563, opened 2023-02-15
feat(protobuf): added error wrappers for invalid length validation

LNSD:
## Context

At this moment, no protobuf max field length validation is performed. In 14/WAKU2-MESSAGE RFC, no maximum length for the `content_topic` attribute is specified. This is potentially problematic. Additionally, this is not supported by _libp2p's minprotobuf_ implementation.

This requirement motivates the present PR to add support in the `common/protobuf` wrapper module for validating the protobuf field length.

## Acceptance criteria

- [x] Added a maximum `content_topic` attribute length: 128 bytes (128 ASCII chars).
- [x] Added new wrapper result and error types to the minprotobuf's ProtoError. Added some error details for better debugging.
- [x] Updated the different protocols codec modules with the new result type.

## Other Considerations

Note that the validation is only executed during the deserialization phase. The content topic length is assumed to be valid in the serialisation phase.

Initially, the minprotobuf's `getRequirefField` was also considered. But as there isn't a max length validation equivalent, for homogeneity, I discarded its usage and added also the "required field missing" error variant.

### DIFF
--- a/tests/common/test_protobuf_validation.nim
+++ b/tests/common/test_protobuf_validation.nim
@@ -1,0 +1,103 @@
+
+{.used.}
+
+import
+  testutils/unittests
+import
+  ../../waku/common/protobuf
+
+
+## Fixtures
+
+const MaxTestRpcFieldLen = 5
+
+type TestRpc = object
+    testField*: string
+
+proc init(T: type TestRpc, field: string): T =
+  T(testField: field)
+
+proc encode(rpc: TestRpc): ProtoBuffer =
+  var pb = initProtoBuffer()
+  pb.write3(1, rpc.testField)
+  pb.finish3()
+  pb
+
+proc encodeWithBadFieldId(rpc: TestRpc): ProtoBuffer =
+  var pb = initProtoBuffer()
+  pb.write3(666, rpc.testField)
+  pb.finish3()
+  pb
+
+proc decode(T: type TestRpc, buf: seq[byte]): ProtobufResult[T] =
+  let pb = initProtoBuffer(buf)
+
+  var field: string
+  if not ?pb.getField(1, field):
+    return err(ProtobufError.missingRequiredField("test_field"))
+  if field.len > MaxTestRpcFieldLen:
+    return err(ProtobufError.invalidLengthField("test_field"))
+
+  ok(TestRpc.init(field))
+
+
+## Tests
+
+suite "Waku Common - libp2p minprotobuf wrapper":
+
+  test "serialize and deserialize - valid length field":
+    ## Given
+    let field = "12345"
+
+    let rpc = TestRpc.init(field)
+
+    ## When
+    let encodedRpc = rpc.encode()
+    let decodedRpcRes = TestRpc.decode(encodedRpc.buffer)
+
+    ## Then
+    check:
+      decodedRpcRes.isOk()
+
+    let decodedRpc = decodedRpcRes.tryGet()
+    check:
+      decodedRpc.testField == field
+
+  test "serialize and deserialize - missing required field":
+    ## Given
+    let field = "12345"
+
+    let rpc = TestRpc.init(field)
+
+    ## When
+    let encodedRpc = rpc.encodeWithBadFieldId()
+    let decodedRpcRes = TestRpc.decode(encodedRpc.buffer)
+
+    ## Then
+    check:
+      decodedRpcRes.isErr()
+
+    let error = decodedRpcRes.tryError()
+    check:
+      error.kind == ProtobufErrorKind.MissingRequiredField
+      error.field == "test_field"
+
+
+  test "serialize and deserialize - invalid length field":
+    ## Given
+    let field = "123456"  # field.len = MaxTestRpcFieldLen + 1
+
+    let rpc = TestRpc.init(field)
+
+    ## When
+    let encodedRpc = rpc.encode()
+    let decodedRpcRes = TestRpc.decode(encodedRpc.buffer)
+
+    ## Then
+    check:
+      decodedRpcRes.isErr()
+
+    let error = decodedRpcRes.tryError()
+    check:
+      error.kind == ProtobufErrorKind.InvalidLengthField
+      error.field == "test_field"
--- a/waku/v2/protocol/waku_message/message.nim
+++ b/waku/v2/protocol/waku_message/message.nim
@@ -12,7 +12,8 @@ else:
 import
   ../../utils/time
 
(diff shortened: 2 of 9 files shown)


rymnc (approved):
LGTM, will create a tracking issue to use the new error in rln-relay's protobuf decoding too.

Maybe the [RFC](https://rfc.vac.dev/spec/14/) should include the 128 byte size constraint on content_topic?

waku/v2/protocol/waku_message/message.nim
```diff
@@ -12,7 +12,10 @@ else:
 import
   ../../utils/time
 
-const MaxWakuMessageSize* = 1024 * 1024 # In bytes. Corresponds to PubSub default
+const
+  MaxTopicAttributeLength* = 128 # 128 bytes
```
alrevuelta, line 16: perhaps nitpick but can we add "content topic" to it? to differenciate from the pubsub?

waku/v2/protocol/waku_store/rpc_codec.nim
```diff
@@ -31,14 +31,14 @@ proc encode*(index: PagingIndexRPC): ProtoBuffer =
...
-proc decode*(T: type PagingIndexRPC, buffer: seq[byte]): ProtoResult[T] =
+proc decode*(T: type PagingIndexRPC, buffer: seq[byte]): ProtobufResult[T] =
   ## creates and returns an Index object out of buffer
   var rpc = PagingIndexRPC()
   let pb = initProtoBuffer(buffer)
 
   var data: seq[byte]
   if not ?pb.getField(1, data):
-    return err(ProtoError.RequiredFieldMissing)
+    return err(ProtobufError.missingRequiredField("index"))
```
alrevuelta, line 41: isn't this field the "digest"? so missingRequiredField("digest")?

alrevuelta (commented):
looks good, left some minor comments.

mind adding some testcases? at least something basic testing this new MaxTopicAttributeLength check?

LNSD:
> Maybe the [RFC](https://rfc.vac.dev/spec/14/) should include the 128 byte size constraint on content_topic?

Honestly, I don't know if we should. IIRC Gossipsub RFC does not specify a maximum topic length; hence it is up to the implementations to set limits. I see here the same kind of situation. Wdyt?

rymnc:
> > Maybe the [RFC](https://rfc.vac.dev/spec/14/) should include the 128 byte size constraint on content_topic?
> 
> Honestly, I don't know if we should. IIRC Gossipsub RFC does not specify a maximum topic length; hence it is up to the implementations to set limits. I see here the same kind of situation. Wdyt?

Works either way, maybe this is better noted in a "Recommendations" section

tests/common/test_protobuf_validation.nim
```diff
@@ -0,0 +1,103 @@
+
```
alrevuelta, line 1: nice. I actually was thinkinig about testing the `encode`, `decode` functions of the protos that were changed, like `PagingInfoRPC`.

A simple back and forth converstion like [this](https://github.com/status-im/nimbus-eth2/blob/5b13b7cc9bd52617f81c2b0ce5eae06827741404/tests/test_eth2_ssz_serialization.nim#L28-L30) to assert that encode()/decode() work as expected and protect the code from future modifications. Also, test this extra logic `if topic.len > MaxPubsubTopicLen:` trying to encode eg a `FilterRequest` with a longer topic, and asserting it fails.

But we can leave that for another PR.

alrevuelta (approved):
> Maybe the [RFC](https://rfc.vac.dev/spec/14/) should include the 128 byte size constraint on content_topic?

I would say whatever can create an incompatibility issue should be added to the spec. I mean, if one implementation has that 128bytes limit and another doesn't, a message sent with a content_topic longer than 128byte, may not be routed by the second implementation (depending on how we do message validation).

regarding the PR, lgtm!

tests/common/test_protobuf_validation.nim
alrevuelta: (ofc testing all encode/decode functions would be a long PR) beyond this, but perhaps we are lacking some testing in protos in general?

LNSD: > (ofc testing all encode/decode functions would be a long PR) beyond this, but perhaps we are lacking some testing in protos in general?

Protocol-specific protobuf tests can be found in the different protocol folders (see the `tests/v2/waku_store` module). Additionally, the different protocols' integration/end-to-end tests implicitly test protocol RPC serialisation. 

I don't see the need to add test suites for unit testing/covering specifically the RPC serialization of the different protocols. If we had to add tests to cover this area: Interoperability tests (a.k.a integration-tests) would be the ones providing additional value.

LNSD:
@jm-clius Looking forward to your feedback on the size limit of the topic and if we should specify it as part of the 14/WAKU2-MESSAGE RFC 😁

jm-clius (commented):
Indeed! Since these can introduce incompatibilities/breaking change in how the protocol is used, it should be part of an RFC and agreed upon before merging, especially since we don't know how this will affect current applications. My suggestion is https://rfc.vac.dev/spec/23/ as recommendations (i.e. the pubsubTopic SHOULD be less than 128 characters. Individual Waku nodes MAY choose to discard messages that does not adhere to....). To get this PR merged, unstaging the changes specific to max topic length and leaving that for a next PR (after RFC agreement) will work. 👍

LNSD:
> To get this PR merged, unstaging the changes specific to max topic length and leaving that for a next PR (after RFC agreement) will work.

I have split the different codecs' topic length validation changes into a separate commit. Waiting for approval @jm-clius @alrevuelta @rymnc

I will open the "max topic length recommendation" RFC PR with the split-out changes after this PR gets merged.

waku/v2/protocol/waku_message/message.nim
```diff
@@ -12,7 +12,8 @@ else:
 import
   ../../utils/time
 
-const MaxWakuMessageSize* = 1024 * 1024 # In bytes. Corresponds to PubSub default
+const
+  MaxWakuMessageSize* = 1024 * 1024 # 1 Mibytes. Corresponds to PubSub default
```
rymnc, line 16:
```suggestion
const MaxWakuMessageSize* = 1024 * 1024 # 1 Mibytes. Corresponds to PubSub default
```

rymnc (approved):
LGTM

jm-clius (approved):
LGTM!

waku/v2/protocol/waku_message/message.nim
LNSD: As I plan to restore the max topic length constants, I opted not to restore the single line const :)

PR #1570, opened 2023-02-20
fix: allow higher resolution timestamps

jm-clius:
This fix allows higher precision `Timestamp`s

Wakurtosis reported that all timestamps logged by nwaku were in second resolution (no fractional seconds), which is too imprecise for testing latency, identifying messages etc.

It turns out that our `Timestamp` construction converted the more precise `float`seconds (with fraction) to `int64` before multiplying, resulting in the fractional part always being truncated

I've also:
- removed the unused util functions for micro- and milliseconds. I believe a nwaku `Timestamp` should always be nanoseconds.
- made the generic generation function more strict in requiring `SomeNumber` types in input.
- added basic test suite

### DIFF
--- a/tests/v2/test_utils_time.nim
+++ b/tests/v2/test_utils_time.nim
@@ -1,0 +1,35 @@
+{.used.}
+
+import
+  stew/results,
+  testutils/unittests
+import
+  ../../waku/v2/utils/time
+
+suite "Utils - Time":
+
+  test "Test timestamp conversion":
+    ## Given
+    let
+      nanoseconds = 1676562429123456789.int64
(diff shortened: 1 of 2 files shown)


LNSD (approved):
LGTM ✅

jm-clius:
Whoops. Naively assumed after local testing that CI will pass. Will investigate the failing tests.

alrevuelta (approved):
apart from the ci issue, lgtm!

jm-clius:
Ci fixed, with higher resolution one test failed because it assumed several instructions can be completed in 1 nanosecond. :)

PR #1572, opened 2023-02-22
fix(jsonrpc): use padded-base64 for payload encoding

LNSD:
This PR includes code and API cleanup and adds tests covering the `base64` module. A typical gardening PR 🪴

> **Note**
> The v0.15.0 is working as expected: **Payload must be encoded using regular base64 (with padding).**

----

Regarding the RFC discussion about the JSON-RPC message payload encoding using base64 ([here](https://github.com/vacp2p/rfc/pull/574#issuecomment-1440393179)):

> Ok... There was some confusion on my side: I thought the payload was encoded in base64 URL-friendly format without padding. This is not what was happening. It is encoded in regular base64 with padding. It seems I implemented it the right way from the very beginning.
> 
> So the v0.15.0 is working as expected: Payload must be encoded using regular base64 (with padding).
> 
> I have updated this PR so it is clear that the encoding must be with padding.

### DIFF
--- a/tests/common/test_base64_codec.nim
+++ b/tests/common/test_base64_codec.nim
@@ -1,0 +1,40 @@
+
+{.used.}
+
+import
+  stew/[results, byteutils],
+  testutils/unittests
+import
+  ../../waku/common/base64
+
+
+suite "Waku Common - stew base64 wrapper":
+
+  test "encode into base64 (with padding)":
+    ## Given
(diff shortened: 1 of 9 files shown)


fryorcraken (approved):

fryorcraken:
I can confirmed this fixes the issues previously reported in https://github.com/waku-org/nwaku/issues/1564
With this PR, nwaku know accepts NodeJS Buffer's `base64` string encoding.

jm-clius (approved):
LGTM thanks!

PR #1574, opened 2023-02-23
chore: log node peer ID with relay trace logging

jm-clius:
Logs the peer ID when trace logging relay messages.

Requested by Wakurtosis as a requirement/improvement when scraping logs.

### DIFF
--- a/waku/v2/node/waku_node.nim
+++ b/waku/v2/node/waku_node.nim
@@ -373,6 +373,9 @@ proc new*(T: type WakuNode,
 proc peerInfo*(node: WakuNode): PeerInfo =
   node.switch.peerInfo
 
+proc peerId*(node: WakuNode): PeerId =
+  node.peerInfo.peerId
+
 # TODO: Extend with more relevant info: topics, peers, memory usage, online time, etc
 proc info*(node: WakuNode): WakuInfo =
   ## Returns information about the Node, such as what multiaddress it can be reached at.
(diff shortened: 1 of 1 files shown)


waku/v2/node/waku_node.nim
```diff
@@ -401,6 +401,7 @@ proc registerRelayDefaultHandler(node: WakuNode, topic: PubsubTopic) =
 
   proc traceHandler(topic: PubsubTopic, data: seq[byte]) {.async, gcsafe.} =
     trace "waku.relay received",
+      peerId=node.switch.peerInfo.peerId,
```
LNSD, line 404: I see that this leaks the "switch" implementation detail. Could you add a `proc peerId(): PeerId =` to `WakuNode`? 
 
IIRC, I already added a `peerInfo` proc. It would be as simple as adding a `PeerId` accessor, so we don't leak the "switch" in different places.

LNSD (commented):
Please check my comment 🙂

waku/v2/node/waku_node.nim
jm-clius: Oh, yeah makes sense.

Done in [6429b90](https://github.com/waku-org/nwaku/pull/1574/commits/6429b903bbc393ce4e7dd4d30a41f5878b42e212)

LNSD (approved):
LGTM ✅

PR #1578, opened 2023-02-27
feat(rln-relay): use precompiled resources w/ rln

rymnc:
Companion PR: https://github.com/vacp2p/zerokit/pull/118
This pr loads rln as a static library, without the need for additional resources (verifier, prover etc). It is an incremental update, and the next pr will download the static lib from zerokit's nightly release directly(blocked at the moment). 
can be merged after zerokit's pr is merged.


.gitmodules
```diff
@@ -143,7 +143,7 @@
 	path = vendor/zerokit
 	url = https://github.com/vacp2p/zerokit.git
 	ignore = dirty
-	branch = master
+	branch = staticlib
```
rymnc, line 146: I'll update this after zerokit's pr is merged.

jm-clius (approved):
Thanks!

LNSD (changes requested):
I noticed that you removed the experimental stage from the Dockerfile. That's a great improvement 👏🏼 

There are some remains in the `Makefile` (e.g., [these lines should be removed](https://github.com/waku-org/nwaku/pull/1578/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L233-L235)). Could you review the Makefile tasks?

LNSD (approved):
LGTM ✅

rymnc:
will merge since zerokit's pr has been merged

PR #1584, opened 2023-02-28
feat: new filter protocol increment - subscribe request handling

jm-clius:
## Background

This is a first increment towards the server-side implementation of the filter protocol changes proposed in https://github.com/vacp2p/rfc/pull/562 with wire protocol in https://github.com/vacp2p/waku/pull/16.

It only covers basic subscriber request handling.

Note that this is still an incomplete feature and isolated in code. Once feature-complete, the old `filter` implementation will be removed and replaced by this one (with simultaneous merging of the RFC). Test cases are also still mainly focused on happy-path cases, to keep this increment as simple as possible.

The major outstanding features are:
- [ ] message handling within the filter service node
- [ ] a client implementation

Dogfooding can likely start without a client implementation in nwaku, as nwaku is most often configured as a service node only.

### Next steps

Further steps, roughly in order:
- [ ] message handling within filter service node
- [ ] a subscriptions sweeper to handle subscription maintenance (e.g. removing stale subscriptions, managing capacity, etc.)
- [ ] test cases that cover error and boundary conditions
- [ ] improved metrics and logging
- [ ] client implementation

### DIFF
--- a/waku/v2/protocol/waku_filter_v2/protocol.nim
+++ b/waku/v2/protocol/waku_filter_v2/protocol.nim
@@ -159,7 +159,7 @@ proc initProtocolHandler(wf: WakuFilter) =
     let buf = await conn.readLp(MaxSubscribeSize)
 
     let decodeRes = FilterSubscribeRequest.decode(buf)
-    if decodeRes.isErr:
+    if decodeRes.isErr():
       error "Failed to decode filter subscribe request", peerId=conn.peerId
       waku_filter_errors.inc(labelValues = [decodeRpcFailure])
       return
--- a/waku/v2/protocol/waku_filter_v2/protocol.nim
+++ b/waku/v2/protocol/waku_filter_v2/protocol.nim
@@ -1,0 +1,159 @@
+## Waku Filter protocol for subscribing and filtering messages
+
+when (NimMajor, NimMinor) < (1, 4):
+  {.push raises: [Defect].}
+else:
+  {.push raises: [].}
+
+import
+  std/[sets,tables],
+  libp2p/peerid
+import
+  ../../node/peer_manager,
+  ../waku_message,
+  ./common,
+  ./codec
+
+const
+  MaxSubscriptions* = 1000 # TODO make configurable
+  MaxCriteriaPerSubscription = 1000
+
+type
+  FilterCriterion* = (PubsubTopic, ContentTopic) # a single filter criterion is fully defined by a pubsub topic and content topic
+  FilterCriteria* = HashSet[FilterCriterion] # a sequence of filter criteria
+
+  WakuFilter* = object
+    subscriptions*: Table[PeerID, FilterCriteria] # a mapping of peer ids to a sequence of filter criteria
+    peerManager: PeerManager
+
+proc pingSubscriber(wf: WakuFilter, peerId: PeerID): FilterSubscribeResult =
+  if peerId notin wf.subscriptions:
+    return err(FilterSubscribeError(
+      kind: FilterSubscribeErrorKind.NOT_FOUND,
+      cause: "NOT_FOUND: peer has no subscriptions"
+    ))
+
+  ok()
+
+proc subscribe(wf: WakuFilter, peerId: PeerID, pubsubTopic: Option[PubsubTopic], contentTopics: seq[ContentTopic]): FilterSubscribeResult =
+  if pubsubTopic.isNone() or contentTopics.isEmpty():
+    return err(FilterSubscribeError(
+      kind: FilterSubscribeErrorKind.BAD_REQUEST,
+      cause: "BAD_REQUEST: pubsubTopic and contentTopics must be specified"
+    ))
+
+  let filterCriteria = toHashSet(contentTopics.mapIt((pubsubTopic.get(), it)))
+
+  if peerId in wf.subscriptions:
+    let peerSubscription = wf.subscriptions[peerId]
+    if peerSubscription.len() + filterCriteria.length >= MaxCriteriaPerSubscription:
+      return err(FilterSubscribeError(
+        kind: FilterSubscribeErrorKind.BAD_REQUEST,
+        cause: "BAD_REQUEST: peer has reached maximum number of filter criteria"
+      ))
+
+    wf.subscriptions[peerId].incl(filterCriteria)
+  else:
+    if wf.subscriptions.len() >= MaxSubscriptions:
+      return err(FilterSubscribeError(
+        kind: FilterSubscribeErrorKind.BAD_REQUEST,
+        cause: "BAD_REQUEST: node has reached maximum number of subscriptions"
+      ))
+    wf.subscriptions[peerId] = filterCriteria
+
+  ok()
+
+proc unsubscribe(wf: WakuFilter, peerId: PeerID, pubsubTopic: Option[PubsubTopic], contentTopics: seq[ContentTopic]): FilterSubscribeResult =
+  if pubsubTopic.isNone() or contentTopics.isEmpty():
+    return err(FilterSubscribeError(
+      kind: FilterSubscribeErrorKind.BAD_REQUEST,
+      cause: "BAD_REQUEST: pubsubTopic and contentTopics must be specified"
+    ))
+
+  let filterCriteria = toHashSet(contentTopics.mapIt((pubsubTopic.get(), it)))
+
+  if peerId notin wf.subscriptions:
+    return err(FilterSubscribeError(
+      kind: FilterSubscribeErrorKind.NOT_FOUND,
+      cause: "NOT_FOUND: peer has no subscriptions"
+    ))
+
+  let peerSubscription = wf.subscriptions[peerId]
+  # TODO: consider error response if filter criteria does not exist
+  wf.subscriptions[peerId].excl(filterCriteria)
+
+  ok()
+
+proc unsubscribeAll(wf: WakuFilter, peerId: PeerID): FilterSubscribeResult =
+  if peerId notin wf.subscriptions:
+    return err(FilterSubscribeError(
+      kind: FilterSubscribeErrorKind.NOT_FOUND,
+      cause: "NOT_FOUND: peer has no subscriptions"
+    ))
+
+  wf.subscriptions.del(peerId)
+
+  ok()
+
+
+proc handleSubscribeRequest*(wf: WakuFilter, request: FilterSubscribeRequest): FilterSubscribeResponse =
+  var subscribeResult: FilterSubscribeResult
+
+  case request.filterSubscribeType
+  of SubscribeType.SUBSCRIBER_PING:
+    subscribeResult = wf.pingSubscriber(request.peerId)
+  of SubscribeType.SUBSCRIBE:
+    subscribeResult = wf.subscribe(request.peerId, request.pubsubTopic, request.contentTopics)
+  of SubscribeType.UNSUBSCRIBE:
+    subscribeResult = wf.unsubscribe(request.peerId, request.pubsubTopic, request.contentTopics)
+  of SubscribeType.UNSUBSCRIBE_ALL:
+    subscribeResult = wf.unsubscribeAll(request.peerId)
+
+  if subscribeResult.isErr():
+    return FilterSubscribeResponse(
+      requestId: requestId,
+      statusCode: subscribeResult.error.kind,
+      statusMessage: some($subscribeResult.error)
+    )
+  else:
+    return FilterSubscribeResponse(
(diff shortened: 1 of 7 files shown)


rymnc:
Can we feature flag this, similar to `rln`? Would save some CI time. It would also need to be under the `EXPERIMENTAL` flag, defined https://github.com/waku-org/nwaku/blob/43b3f1cef4c28d5f493efaa0687fc5e1fdf8a821/Makefile#L98-L104

jm-clius:
> Can we feature flag this, similar to `rln`? Would save some CI time. It would also need to be under the `EXPERIMENTAL` flag, defined

This is currently not integrated anywhere or part of any CI, while incrementing towards the feature. The idea is that this implementation will completely replace the existing filter implementation within the next week or two, so there may be little value in having a separate build or flag (the existing `filter` flags will be used), although we could consider doing it if both implementations need to co-exist for a while (hopefully not).

waku/v2/protocol/waku_filter_v2/protocol.nim
```diff
@@ -0,0 +1,184 @@
...
+
+  ok()
+
+proc unsubscribeAll(wf: WakuFilter, peerId: PeerID): FilterSubscribeResult =
+  if peerId notin wf.subscriptions:
+    debug "unsubscibing peer has no subscriptions", peerId=peerId
+    return err(FilterSubscribeError(
+      kind: FilterSubscribeErrorKind.NOT_FOUND,
+      cause: "peer has no subscriptions"
+    ))
```
LNSD, line 117: Same here. I would add a convenience wrapper proc for the errors:

```Nim
FilterSubscribeError.notFound(cause="peer has no subscriptions")
```

LNSD (approved):
I see some details but overall makes sense to me. Feel free to take my comments as part of the upcoming increments.

waku/v2/protocol/waku_filter_v2/protocol.nim
```diff
@@ -0,0 +1,184 @@
...
+
+  trace "subscribing peer to filter criteria", peerId=peerId, filterCriteria=filterCriteria
+
+  if peerId in wf.subscriptions:
+    var peerSubscription = wf.subscriptions.mgetOrPut(peerId, initHashSet[FilterCriterion]())
+    if peerSubscription.len() + filterCriteria.len() >= MaxCriteriaPerSubscription:
+      return err(FilterSubscribeError(
+        kind: FilterSubscribeErrorKind.SERVICE_UNAVAILABLE,
+        cause: "peer has reached maximum number of filter criteria"
+      ))
```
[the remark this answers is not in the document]
jm-clius: > In my opinion, we shouldn't.

Agreed! I also considered this scenario and think a request should either be fully honoured or fail (partially honouring a request is a nightmare in error handling from both client and service node's perspective)

> here, despite erroring out, we are inserting in the subscriptions map

Wait, I don't see it or am missing something obvious. 😅 We return in L63 without doing the insert in L68 and L69. Good idea to return earlier though.

> e.g. MAX 20 subscriptions per query

Good point. Will also clarify this as a recommendation so that we get reasonable filter requests from the beginning.

> I would calculate the remaining "subscription slots

As mentioned above, sounds reasonable to me too to return earlier here, although I don't quite see that we're currently exceeding max criteria per peer subscription. Will address in a follow-up (the subscriptions map is likely to be split to a separate module soon in any case, as I can imagine we may want to make it a bit more complex to e.g. build a reverse index, add timestamps, a sweeper, etc.)

waku/v2/protocol/waku_filter_v2/protocol.nim
```diff
@@ -0,0 +1,184 @@
...
+    peerSubscription.incl(filterCriteria)
+    wf.subscriptions[peerId] = peerSubscription
+  else:
+    if wf.subscriptions.len() >= MaxSubscriptions:
+      return err(FilterSubscribeError(
+        kind: FilterSubscribeErrorKind.SERVICE_UNAVAILABLE,
+        cause: "node has reached maximum number of subscriptions"
+      ))
+    debug "creating new subscription", peerId=peerId
+    wf.subscriptions[peerId] = filterCriteria
```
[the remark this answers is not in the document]
jm-clius: I agree re early checking (e.g. proc that quickly check if a new request can be honoured given current subscriptions/criteria-per-subscription capacity). Not sure though I see how we

> permit surpassing the MaxSubscriptions limit

if we return immediately once we reach the `==` limit?

I will add the early check in any case in a follow-up though.

waku/v2/protocol/waku_filter_v2/protocol.nim
```diff
@@ -0,0 +1,184 @@
...
+
+  ok()
+
+proc unsubscribe(wf: WakuFilter, peerId: PeerID, pubsubTopic: Option[PubsubTopic], contentTopics: seq[ContentTopic]): FilterSubscribeResult =
```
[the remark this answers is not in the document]
jm-clius: Good point. Will add to follow-up!

waku/v2/protocol/waku_filter_v2/protocol.nim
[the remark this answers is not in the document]
LNSD: > Wait, I don't see it or am missing something obvious.  We return in L63 without doing the insert in L68 and L69. Good idea to return earlier though.

My bad, I misread this part. I thought that `mgetOrPut` was modifying already the table's value. True. You bail out before setting the value back to the table.

waku/v2/protocol/waku_filter_v2/protocol.nim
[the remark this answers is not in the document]
LNSD: > if we return immediately once we reach the == limit?
> 
> I will add the early check in any case in a follow-up though.

I believe that doing a precondition check is the safest approach. So you avoid some necessary operations (and you save some CPU cycles).

waku/v2/protocol/waku_filter_v2/protocol.nim
jm-clius: Added in [4b6854c](https://github.com/waku-org/nwaku/pull/1584/commits/4b6854c3d4edc63f7796122a23f91ae8c0d5702f)

waku/v2/protocol/waku_filter_v2/protocol.nim
```diff
@@ -0,0 +1,159 @@
...
+  ./common,
+  ./protocol_metrics,
+  ./rpc_codec,
+  ./rpc
+
+logScope:
+  topics = "waku filter"
+
+const
+  MaxSubscriptions* = 1000 # TODO make configurable
```
alrevuelta, line 26: perhaps a nitpick but

`MaxTotalSubscriptions`
`MaxPeerSubscriptions`

waku/v2/protocol/waku_filter_v2/protocol.nim
```diff
@@ -0,0 +1,159 @@
...
+proc pingSubscriber(wf: WakuFilter, peerId: PeerID): FilterSubscribeResult =
+  trace "pinging subscriber", peerId=peerId
+
+  if peerId notin wf.subscriptions:
+    debug "pinging peer has no subscriptions", peerId=peerId
+    return err(FilterSubscribeError.notFound())
+
+  ok()
+
+proc subscribe(wf: WakuFilter, peerId: PeerID, pubsubTopic: Option[PubsubTopic], contentTopics: seq[ContentTopic]): FilterSubscribeResult =
```
alrevuelta, line 46: any reason why `pubsubTopic` is optional?

waku/v2/protocol/waku_filter_v2/protocol.nim
```diff
@@ -0,0 +1,159 @@
...
+
+logScope:
+  topics = "waku filter"
+
+const
+  MaxSubscriptions* = 1000 # TODO make configurable
+  MaxCriteriaPerSubscription = 1000
+
+type
+  FilterCriterion* = (PubsubTopic, ContentTopic) # a single filter criterion is fully defined by a pubsub topic and content topic
```
alrevuelta, line 30: just an idea:
* make `FilterCriterion` a table pubsubtopic->contenttopic
* easier to reason perhaps since contenttopic is a "subtopic" of gossipsub topic
* less likely that we store duplicates since someone subscribing to ctopic_1, ctopic_2 in gtopic_1 will require: (gtopic_1, ctopic_1), (gtopic_1, ctopic_2).
* faster access time perhaps?
* but the most important, easier to access subscriptions
* one type less

```
# renamed FilterCriterion to PeerSubscriptions
type
   PeerSubscriptions* = Table[PubsubTopic, HashSet[ContentTopic]]
```

and then `WakuFilter`
```
 WakuFilter* = ref object of LPProtocol
    subscriptions*: Table[PeerID, PeerSubscriptions]
    peerManager: PeerManager
```

wdyt?

waku/v2/protocol/waku_filter_v2/protocol.nim
```diff
@@ -0,0 +1,159 @@
...
+    if decodeRes.isErr():
+      error "Failed to decode filter subscribe request", peerId=conn.peerId
+      waku_filter_errors.inc(labelValues = [decodeRpcFailure])
+      return
+
+    let request = decodeRes.value #TODO: toAPI() split here
+
+    let response = wf.handleSubscribeRequest(conn.peerId, request)
+
+    await conn.writeLp(response.encode().buffer) #TODO: toRPC() separation here
```
alrevuelta, line 146: not very likely to happen in reality, but if the connection is closed by the counterparty right before, this will throw an exception. I just verified empirically when modifying px.

https://github.com/waku-org/nwaku/blob/master/tests/v2/test_waku_peer_exchange.nim#L247

alrevuelta (approved):
lgtm! some minor comments/ideas.

rymnc (approved):
LGTM

waku/v2/protocol/waku_filter_v2/protocol.nim
jm-clius: I agree. Will change in next PR.

waku/v2/protocol/waku_filter_v2/protocol.nim
jm-clius: It's an optional in the request RPC since it's only mandatory when following the SUBSCRIBE workflow. I made the choice to do all request consistency checking within the utility function belonging to that request. In this case it is mandatory to have a `pubsubTopic`, so `subscribe` returns a bad request error if optional is none.

waku/v2/protocol/waku_filter_v2/protocol.nim
jm-clius: Mmm. This is indeed how we previously modelled it - a map of peer IDs to a map of pubsub topics to content topics - but I made the (perhaps not final) decision to flatten this as much as possible. Reasoning: the subscription is actually against a `contentTopic`, so in this model the "real" subscription would be nested two levels deep - we just (mandatorily) require each `contentTopic` to be qualified against a specific `pubsubTopic`. Explicitly modelling the pubsub topic separately requires a bit more maintenance. In that case it would be invalid, for example, to subscribe/unsubscribe with filter criteria that contains only a `pubsubTopic` or for a subscription where a peer's Peer Subscription has an empty content topic set against any pubsub topic. We can build in rules around this, of course, but importantly we want to move away from a model where `pubsubTopic` is necessary as a qualifier at all.

I am planning possible changes to this model, though, as we may need something of a reverse map/index when looking up peers for a specific subscription. In this case the reverse index may be modelled as you suggested. Will still think about this, as you make a good point.

waku/v2/protocol/waku_filter_v2/protocol.nim
```diff
@@ -0,0 +1,159 @@
...
+
+proc subscribe(wf: WakuFilter, peerId: PeerID, pubsubTopic: Option[PubsubTopic], contentTopics: seq[ContentTopic]): FilterSubscribeResult =
+  if pubsubTopic.isNone() or contentTopics.len() == 0:
+    return err(FilterSubscribeError.badRequest("pubsubTopic and contentTopics must be specified"))
+
+  let filterCriteria = toHashSet(contentTopics.mapIt((pubsubTopic.get(), it)))
+
+  trace "subscribing peer to filter criteria", peerId=peerId, filterCriteria=filterCriteria
+
+  if peerId in wf.subscriptions:
```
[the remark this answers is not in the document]
jm-clius: True. But to e.g. count number of total subscriptions and enforce limit you'd need to count the values of the second map - i.e. sum of content topics for all pubsub topics (easily doable of course, but slightly more roundabout when modelled).

PR #1587, opened 2023-03-02
chore(rln-relay): move rln-relay test suite to own dir

rymnc:
As mentioned by @LNSD, this dir structure makes more sense for the waku_rln_relay test suite


tests/all_tests_v2.nim
```diff
@@ -48,7 +48,6 @@ import
   ./v2/test_utils_peers,
   ./v2/test_message_cache,
   ./v2/test_peer_manager,
-  ./v2/test_web3, # TODO  remove it when rln-relay tests get finalized
```
rymnc, line 51: This is why the `test_web3.nim` file was removed

PR #1591, opened 2023-03-06
fix: clean and normalize module imports

LNSD:
From discord (@arnetheduck):

> hm, looks like some waku imports are a bit off - take for example https://github.com/waku-org/nwaku/blob/master/waku/v2/node/peer_manager/peer_manager.nim#L15 - the ../.. works because of an unfortunate quirk in Nim when compiling a top-level project in a suitable folder, but the file is actually not available in that location - instead, it actually resides in ../../../../waku/v2/protocol/waku_relay (or simply ../protocol/waku_relay) preventing the use of the module from "outside", ie when importing waku in a stand-alone application - I think this might be the leftover of some refactoring? cc @LNSD 

This PR aims to fix this issue.

### DIFF
--- a/waku/v2/node/jsonrpc/admin/handlers.nim
+++ b/waku/v2/node/jsonrpc/admin/handlers.nim
@@ -9,11 +9,11 @@ import
   json_rpc/rpcserver,
   libp2p/[peerinfo, switch]
 import
-  ../../../../waku/v2/protocol/waku_store,
-  ../../../../waku/v2/protocol/waku_filter,
-  ../../../../waku/v2/protocol/waku_relay,
-  ../../../../waku/v2/node/peer_manager,
-  ../../../../waku/v2/node/waku_node,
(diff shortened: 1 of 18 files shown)


jakubgs:
Thanks! I tested this with our Nix derivation from:

* https://github.com/NixOS/nixpkgs/pull/193679

And it works:
```
 > nix-build -A nim-waku .
/nix/store/z6x53di4xpk9j2lx0z2m75iznfjsyb57-nwaku-fb506e9196e2838b2dad08de011f120c62f46e90-fb506e91

 > $(nix-build -A nim-waku .)/bin/wakunode2 --version                                                                           
version / git commit hash: fb506e9196e2838b2dad08de011f120c62f46e90
```

jakubgs (approved):

jm-clius (approved):
LGTM, assuming it fixes the issue. (Sounds like it does.)

PR #1595, opened 2023-03-07
chore(enr): move waku_enr to protocols module

LNSD:
A new gardening PR 🪴 This is part of the ENR and discovery refactoring groundworks for static sharding:

- [x] Move `wakuenr` out of the `utils` module and into the `protocol` module.
- [x] Update imports accordingly.


rymnc (approved):
LGTM, like the change of waku_enr into a protocol :)

jm-clius (approved):
MSTM

LNSD:
> LGTM, like the change of waku_enr into a protocol :)

There are things inside the "protocol" directory that are actual protocol implementations (e.g., Waku Relay, Waku Store, etc.), but others are services (e.g., Waku Archive, Waku Keystore, etc.). Then there is the "core" module (a.k.a. Waku Message).

As I see it, we should flatten the hierarchy (remove the `protocol` folder) and talk about "modules" in general.

PR #1596, opened 2023-03-07
refactor(enr): node capabilities code clean up and reorganization

LNSD:
This is part of the ENR and discovery refactoring groundworks for static sharding:

- [x] Refactor the capabilities bitfield and the associated procedures
- [x] Support specifying the ENR sequence number in the extension method
- [x] Updated tests and dependent modules accordingly

### DIFF
--- a/apps/wakunode2/wakunode2.nim
+++ b/apps/wakunode2/wakunode2.nim
@@ -269,10 +269,12 @@ proc initNode(conf: WakuNodeConf,
                     else:
                       @[]
 
-    wakuFlags = initWakuFlags(conf.lightpush,
-                              conf.filter,
-                              conf.store,
-                              conf.relay)
+    wakuFlags = CapabilitiesBitfield.init(
+        lightpush = conf.lightpush,
+        filter = conf.filter,
+        store = conf.store,
+        relay = conf.relay
+      )
 
   var node: WakuNode
 
--- a/tests/v2/test_waku_enr.nim
+++ b/tests/v2/test_waku_enr.nim
@@ -1,14 +1,113 @@
 {.used.}
 
 import
-  std/[options, sequtils],
-  stew/byteutils,
+  std/options,
+  stew/[results, byteutils],
   testutils/unittests
 import
   ../../waku/v2/protocol/waku_enr,
   ./testlib/waku2
 
-suite "Waku ENR":
+
+suite "Waku ENR -  Capabilities bitfield":
+  test "check capabilities support":
+    ## Given
+    let bitfield: CapabilitiesBitfield = 0b0000_1101u8  # Lightpush, Filter, Relay
+
+    ## Then
+    check:
+      bitfield.supportsCapability(Capabilities.Relay)
+      not bitfield.supportsCapability(Capabilities.Store)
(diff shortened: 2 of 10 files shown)


tests/v2/test_waku_enr.nim
```diff
@@ -126,92 +221,3 @@ suite "Waku ENR":
...
-    let
-      enrIp = ValidIpAddress.init("127.0.0.1")
-      enrTcpPort, enrUdpPort = Port(60000)
-      enrKey = generateSecp256k1Key()
-      multiaddrs = @[MultiAddress.init("/ip4/127.0.0.1/tcp/442/ws")[]]
-
-      # TODO: Refactor enr.Record.init, provide enums as inputs enr.Record.init(capabilites=[Store,Filter])
-      # TODO: safer than a util function and directly using the bits
-      # test all flag combinations 2^4 = 16 (b0000-b1111)
-      records = toSeq(0b0000_0000'u8..0b0000_1111'u8)
```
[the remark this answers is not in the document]
LNSD: The test is redundant. The code is already covered in the "Waku ENR - Capabilities" suite (above)

waku/v2/protocol/waku_enr.nim
```diff
@@ -147,11 +180,16 @@ func toMultiAddresses*(multiaddrsField: seq[byte]): seq[MultiAddress] =
 
   return multiaddrs
 
+
+## ENR
+
 func init*(T: type enr.Record,
+           seqNum: uint64,
```
jm-clius, line 187: Do we plan to implement some heuristic around the `seq` number at some point?

jm-clius (approved):
LGTM!

waku/v2/protocol/waku_enr.nim
LNSD: I am exposing it since the discv5 protocol uses it to exchange records and update the stored ENR. If the number is stuck in `1`, the querying node stored records may not be updated.

IMO, we should develop the ENR sequence number heuristic after working on an integrated discovery strategy.

apps/wakunode2/wakunode2.nim
```diff
@@ -269,10 +269,12 @@ proc initNode(conf: WakuNodeConf,
                     else:
                       @[]
 
-    wakuFlags = initWakuFlags(conf.lightpush,
-                              conf.filter,
-                              conf.store,
-                              conf.relay)
+    wakuFlags = CapabilitiesBitfield.init(
```
alrevuelta, line 272: Now that we are refactoring this, would be good to abstract this away. I mean, wakunode should have no knowledge about wakuFlags since it's too low-level and specific to discv5.

I would suggest doing:

```
WakuDiscoveryV5.new(..., capabilities=[Store, Lightpush])
```

which is less likely to mess up.

fine if you want to address in a followup PR

alrevuelta (approved):

apps/wakunode2/wakunode2.nim
LNSD: Sure, you will see how we will do the ENR extension in the next PR.

PR #1597, opened 2023-03-07
refactor(networking): unify peer data models, remove StoredInfo

alrevuelta:
Summary:
* `RemotePeerInfo` and `StoredInfo` store very similar content, being totally redundant.
* Back and forth conversions between both types are all over the codebase.
* This PR unifies `RemotePeerInfo` and `StoredInfo`, removing the later leaving only `RemotePeerInfo`.

### DIFF
--- a/waku/v2/utils/peers.nim
+++ b/waku/v2/utils/peers.nim
@@ -17,7 +17,6 @@ import
           multicodec,
           peerid,
           peerinfo,
-          peerstore,
           routing_record]
 
 #import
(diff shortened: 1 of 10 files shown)


waku/v2/utils/peers.nim
```diff
@@ -16,15 +17,73 @@ import
...
+  DisconnectBook* = ref object of PeerBook[int64]
+
+  # Keeps track of the origin of a peer
+  SourceBook* = ref object of PeerBook[PeerOrigin]
+
+  # Direction
+  DirectionBook* = ref object of PeerBook[PeerDirection]
+
+  # ENR Book
+  ENRBook* = ref object of PeerBook[enr.Record]
```
jm-clius, line 68: Even though the `peers` util module itself should probably move closer to the `peer_manager`, I don't think the concept of a `PeerBook` should mix with the concept of a `Peer`. In other words, I'd leave the `Book`s part of `waku_peer_store` for now.

jm-clius (approved):
Thanks! This improves consistency. I wouldn't couple concepts around `PeerBook`s with the `Peer` though and leave those as part of the `waku_peer_store`.

Side note: Fwiw, I've never been fully convinced of the term `RemotePeerInfo` - it was introduced to differentiate it from `PeerInfo`, which became (but wasn't always) solely _local_ peer info. Perhaps we can think of a better term at some point...

LNSD (approved):
Overall, LGTM ✅ 

We should have a conversation about the "Peer" data model. I concur with @jm-clius. The `RemotePeerInfo` type overlaps with libp2p's `PeerId`. 

As I see it:

> We should come up with a "waku peer type" (e.g., as simple as `Peer`) that should be compatible ad convertible to/from and contain the following peer types and peer info formats:
> 
> - nim-libp2p's `PeerId`
> - nim-ibp2p's Multiaddress
> - nim-eth `PeerId`
> - nim-eth ENR `Record` 
> 
> This also affects the key types. Additionally, we should come up with a "waku key type" (e.g., as simple as `Key`) that should be compatible ad convertible to/from and contain the following key types:
> 
> - nim-eth key types
> - nim-libp2p key types

waku/v2/utils/peers.nim
alrevuelta: You are totally right, fixed c5a699c6e848cf76d158143605876a5f03c32d55

(I kind of regret using the peerbook, but that's another story :P)

PR #1599, opened 2023-03-09
feat(enr): added enr builder waku capabilities extension

LNSD:
This PR introduces the Waku ENR capabilities field extension to the `EnrBuilder` type.

- [x] Added the `EnrBuilder` Waku node capabilities extension.
- [x] Added tests covering this functionality.

### DIFF
--- a/waku/v2/protocol/waku_enr.nim
+++ b/waku/v2/protocol/waku_enr.nim
@@ -12,9 +12,10 @@ import
   stew/[endians2, results],
   stew/shims/net,
   eth/keys,
-  eth/p2p/discoveryv5/enr,
   libp2p/[multiaddress, multicodec],
   libp2p/crypto/crypto
+import
+  ../../common/enr
 
 export enr, crypto, multiaddress, net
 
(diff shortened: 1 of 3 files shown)


waku/v2/protocol/waku_enr.nim
```diff
@@ -66,9 +68,19 @@ func toCapabilities*(bitfield: CapabilitiesBitfield): seq[Capabilities] =
...
-  toFieldPair(CapabilitiesEnrField, @[caps.uint8])
+# ENR builder extension
+
+proc withWakuCapabilities*(builder: var EnrBuilder, caps: CapabilitiesBitfield) =
+  builder.addFieldPair(CapabilitiesEnrField, @[caps.uint8])
+
+proc withWakuCapabilities*(builder: var EnrBuilder, caps: varargs[Capabilities]) =
+  withWakuCapabilities(builder, CapabilitiesBitfield.init(caps))
+
+proc withWakuCapabilities*(builder: var EnrBuilder, caps: openArray[Capabilities]) =
```
alrevuelta, line 79: just an opinion. I generally prefer one function rather than multiple ones. less code to read, more simple and clearer. not sure I see the value of having all three.

Isn't this too low level? Cool thing about the builder is abstracting away all this
`withWakuCapabilities(..., caps: CapabilitiesBitfield))`

Both are very similar.
`withWakuCapabilities(Relay, Store)`
`withWakuCapabilities(@[Relay, Store])`

tldr: I would just leave `withWakuCapabilities(@[Relay, Store])`

alrevuelta (approved):
lgtm! left a non blocking comment.

waku/v2/protocol/waku_enr.nim
LNSD: Different `proc` signatures, cover different use cases. 

At first sight, you can see it as too low level, for example. But IIRC this is the exact API we need to create an ENR in the Waku discv5 module. A similar thing happens with the multiple capabilities methods: it is not the same to pass a `seq[T]` than a `varargs[T]`. You can can convert easily a `varargs` instance into a sequence, but the opposite has some compilation-time implications (varargs can be understood as an array of known size at compile time).

jm-clius (approved):
LGTM, thanks!

PR #1600, opened 2023-03-09
feat(filter-v2): new filter protocol increment - message handling and clients

jm-clius:
## Background

This is a further increment towards the implementation of the filter protocol changes proposed in https://github.com/vacp2p/rfc/pull/562 with wire protocol in https://github.com/vacp2p/waku/pull/16.

It builds on https://github.com/waku-org/nwaku/pull/1584 and add:
- [ ] message handling on the server side
- [ ] a basic filter client API implementation
- [ ] a simple subscriptions maintenance loop

Note that this is still an incomplete feature and isolated in code.

There are also some comments on the previous PR which I will address separately (but would have bloated this one even more).

## Next steps

- [ ] test cases that cover error and boundary conditions
- [ ] improved consistency checks (e.g. max number of content topics) and error handling
- [ ] more sophisticated policy for managing and storing client peers in the peer store
- [ ] more metrics, especially duration of servicing requests (server and client side)


waku/v2/protocol/waku_filter/client.nim
```diff
@@ -3,7 +3,7 @@ when (NimMajor, NimMinor) < (1, 4):
 else:
```
jm-clius, line 3: The changes in this file is purely whitespace. Please ignore.

tests/v2/waku_filter_v2/test_waku_filter.nim
```diff
@@ -0,0 +1,287 @@
...
+    # When
+    let msg4 = fakeWakuMessage(contentTopic=contentTopic2)
+    pushHandlerFuture = newFuture[(string, WakuMessage)]() # Clear previous future
+    await wakuFilter.handleMessage(DefaultPubsubTopic, msg4)
+
+    # Then
+    check:
+      not (await pushHandlerFuture.withTimeout(2.seconds)) # No message should be pushed
+
+  asyncTest "subscribe to multiple content topics and unsubscribe all":
```
alrevuelta, line 214: Perhaps missing a testcase subscribing to multiple gossipsub topics and multiple content topics?

tests/v2/waku_filter_v2/test_waku_filter_protocol.nim
```diff
@@ -237,3 +235,64 @@ suite "Waku Filter - handling subscribe requests":
...
+        filterSubscribeType = FilterSubscribeType.SUBSCRIBE,
+        pubsubTopic = some(DefaultPubsubTopic),
+        contentTopics = @[DefaultContentTopic]
+      )
+
+    # When
+    switch.peerStore[ProtoBook][peerId1] = @[WakuFilterPushCodec]
+    switch.peerStore[ProtoBook][peerId2] = @[WakuFilterPushCodec]
+    switch.peerStore[ProtoBook][peerId3] = @[WakuFilterPushCodec]
+    discard wakuFilter.handleSubscribeRequest(peerId1, filterSubscribeRequest)
```
alrevuelta, line 258: maybe nitpick?

`require wakuFilter.handleSubscribeRequest(peerId1, filterSubscribeRequest).isOk`

waku/v2/protocol/waku_filter_v2/protocol.nim
```diff
@@ -125,8 +119,61 @@ proc handleSubscribeRequest*(wf: WakuFilter, peerId: PeerId, request: FilterSubs
   else:
     return FilterSubscribeResponse.ok(request.requestId)
 
-proc handleMessage*(wf: WakuFilter, message: WakuMessage) =
-  raiseAssert "Unimplemented"
+proc pushToPeer(wf: WakuFilter, peer: PeerId, buffer: seq[byte]) {.async.} =
+  trace "pushing message to subscribed peer", peer=peer
+
+  if not wf.peerManager.peerStore.hasPeer(peer, WakuFilterPushCodec):
+    # Check that peer has not been removed from peer store
```
alrevuelta, line 126: Beyond this pr, but this makes me think that the peermanager should have knowledge of subscriptions/filter to not remove these peers from the peerstore.

waku/v2/protocol/waku_filter_v2/protocol.nim
```diff
@@ -157,3 +204,24 @@ proc new*(T: type WakuFilter,
   )
   wf.initProtocolHandler()
   wf
+
+const MaintainSubscriptionsInterval* = 1.minutes
+
+proc startMaintainingSubscriptions*(wf: WakuFilter, interval: Duration) =
```
alrevuelta, line 210: beyond this pr, not blocking.

since we have a few of these timers, I would suggest creating a [heartbeat](https://github.com/status-im/nim-libp2p/blob/c11772c94e44f0987b79d6593f34d6a4f92f07c5/libp2p/utils/heartbeat.nim#L20) so that we can reuse code.

and then just.

```
heartbeat "sometext", interval:
  await functionToRun.
```

ofc it depends if:
* we want to run it exactly every x time.
* or wait x time between the last completed run.

alrevuelta (approved):
lgtm! just some minor comments.

waku/v2/protocol/waku_filter_v2/protocol.nim
jm-clius: I agree! My current thinking is that we may implement some kind of policy which could be set by clients of the peer store. Needs more thinking, though.

tests/v2/waku_filter_v2/test_waku_filter.nim
jm-clius: Good point! Added in [9105f88](https://github.com/waku-org/nwaku/pull/1600/commits/9105f88a45688599ea8ee580a873aa17a3af21ac) (I want to significantly extend test cases, but at least cover the most basic happy paths for now, so this was indeed an oversight).

tests/v2/waku_filter_v2/test_waku_filter_protocol.nim
jm-clius: Agree. Done in [9105f88](https://github.com/waku-org/nwaku/pull/1600/commits/9105f88a45688599ea8ee580a873aa17a3af21ac)

tests/v2/waku_filter_v2/test_waku_filter_protocol.nim
```diff
@@ -237,3 +235,64 @@ suite "Waku Filter - handling subscribe requests":
...
+    # Then
+    check:
+      wakuFilter.subscriptions.len == 3
+      wakuFilter.subscriptions.hasKey(peerId1)
+      wakuFilter.subscriptions.hasKey(peerId2)
+      wakuFilter.subscriptions.hasKey(peerId3)
+
+    # When
+    # Remove peerId1 and peerId3 from peer store
+    discard switch.peerStore[ProtoBook].del(peerId1)
```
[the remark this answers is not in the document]
jm-clius: Thanks! [9105f88](https://github.com/waku-org/nwaku/pull/1600/commits/9105f88a45688599ea8ee580a873aa17a3af21ac)

waku/v2/protocol/waku_filter_v2/protocol.nim
LNSD: > the peermanager should have knowledge of subscriptions/filter to not remove these peers from the peerstore.

It makes sense, but... How would you implement that without coupling the peer management code to the filter protocol code?

waku/v2/protocol/waku_filter_v2/protocol.nim
```diff
@@ -125,8 +119,61 @@ proc handleSubscribeRequest*(wf: WakuFilter, peerId: PeerId, request: FilterSubs
...
+  trace "pushing message to subscribed peers", peers=peers, messagePush=messagePush
+
+  let bufferToPublish = messagePush.encode().buffer
+
+  var pushFuts: seq[Future[void]]
+  for peerId in peers:
+    let pushFut = wf.pushToPeer(peerId, bufferToPublish)
+    pushFuts.add(pushFut)
+
+  await allFutures(pushFuts)
```
LNSD, line 149: I think that we need a timeout when pushing the messages to the subscribers. I don't know if it should be per push, per peer, or a unique timeout for the `allFutures` call.

LNSD (approved):
LGTM ✅

waku/v2/protocol/waku_filter_v2/protocol.nim
alrevuelta: > It makes sense, but... How would you implement that without coupling the peer management code to the filter protocol code?

mm interesting. I would say its not possible without coupling it. I mean, the cleanest thing imho would be to have a new field in the Peer (subscribedToFilter) that can be set or unset via a proc used by filter. Then the peer manager will never? drop this connection.

waku/v2/protocol/waku_filter_v2/protocol.nim
jm-clius: Good point! The individual peer _dials_ already have timeouts, but not the stream writes. Since these should be spawned concurrently, I think a global timeout for `allFutures` will be enough safety for now.

PR #1603, opened 2023-03-13
feat(rln-relay): fetch release from zerokit ci, or build

rymnc:
This PR introduces a script, `build_rln.sh` that extracts the host triple, checks if a zerokit release asset exists for it, and if not, builds zerokit from scratch.

Zerokit supports 6 target architectures at the moment, and it is possible that some users will need to build it manually when compiling nwaku with experimental features enabled. I propose that we continue to use zerokit as a submodule for this reason.

### DIFF
--- a/build_rln.sh
+++ b/build_rln.sh
@@ -1,0 +1,31 @@
+#!/usr/bin/env bash
+
+# This script is used to build the rln library for the current platform, or download it from the
+# release page if it is available.
+
+set -e
+
+# first argument is the build directory
+build_dir=$1
+
+if [ -z "$build_dir" ]; then
+    echo "No build directory specified"
+    exit 1
+fi
+
+# Get the host triplet
+host_triplet=$(rustup show | grep "Default host: " | cut -d' ' -f3)
+
+# Download the prebuilt rln library if it is available
+if wget https://github.com/vacp2p/zerokit/releases/download/nightly/$host_triplet-rln.tar.gz
+then
+    echo "Downloaded $host_triplet-rln.tar.gz"
+    tar -xzf $host_triplet-rln.tar.gz
+    mv release/librln.a .
+    rm -rf $host_triplet-rln.tar.gz release
+else
+    echo "Failed to download $host_triplet-rln.tar.gz"
+    # Build rln instead
+    cargo build --release --manifest-path $build_dir/rln/Cargo.toml
+    cp $build_dir/rln/target/release/librln.a .
+fi
--- a/.gitignore
+++ b/.gitignore
@@ -41,6 +41,7 @@ node_modules/
 # RLN / keystore
 rlnCredentials.txt
 rlnKeystore.json
+*.tar.gz
 
 # Nimbus Build System
 nimbus-build-system.paths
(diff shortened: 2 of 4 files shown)


build_rln.sh
```diff
@@ -0,0 +1,31 @@
...
+if [ -z "$build_dir" ]; then
+    echo "No build directory specified"
+    exit 1
+fi
+
+# Get the host triplet
+host_triplet=$(rustup show | grep "Default host: " | cut -d' ' -f3)
+
+# Download the prebuilt rln library if it is available
+if wget https://github.com/vacp2p/zerokit/releases/download/nightly/$host_triplet-rln.tar.gz
```
rymnc, line 20:
```suggestion
if wget https://github.com/vacp2p/zerokit/releases/download/v0.2/$host_triplet-rln.tar.gz
```
note: we will use a static version some time in the future, keeping nightly in for now will allow us to detect breaking changes faster

build_rln.sh
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
jm-clius: Mmm. Good question. Since there are no obvious alternatives, I think the root folder is the best option _for now_, since this is where other build and environment scripts are.

jm-clius (approved):
Right, I'm still not sure how we should prevent cluttering in build resources and scripts, but I agree that this is better than building everytime.
One worry about using `nightly` is that it may cause CI builds to become indeterminate?

rymnc:
Yes you're right about experimental ci becoming indeterminate, however, using `nightly` is just temporary, till the next release of zerokit is created (which includes the build artifacts for different targets). it should be soon!

LNSD (approved):
LGTM ✅

jakubgs (approved):
Maybe we should have a `scripts` folder for `build_rln.sh`?
Leaving stuff at repo root unless absolutely necessary should be avoided.

And yes, we can add `wget` to `shell.nix`, though honestly I'd prefer `curl`, as it's more extendable and has more option, like the combination of `--silent --fail-with-body`.

build_rln.sh
```diff
@@ -0,0 +1,31 @@
...
+
+# This script is used to build the rln library for the current platform, or download it from the
+# release page if it is available.
+
+set -e
+
+# first argument is the build directory
+build_dir=$1
+
+if [ -z "$build_dir" ]; then
```
jakubgs, line 11: If we are using bash just use `[[ ]]` for `if` clauses. You should probably just run `shellcheck` on the whole script.

build_rln.sh
jakubgs, line 20: If we are saving such files at repo root we should also also add something like `*.tar.gz` to `.gitignore`.

build_rln.sh
rymnc: Addressed in [2ea1ca0](https://github.com/waku-org/nwaku/pull/1603/commits/2ea1ca007a0923c8cd07f85ada244a1ee07cf1f6)

build_rln.sh
rymnc: Addressed in [2ea1ca0](https://github.com/waku-org/nwaku/pull/1603/commits/2ea1ca007a0923c8cd07f85ada244a1ee07cf1f6)

PR #1608, opened 2023-03-14
fix: only pass -msse3 if i386 or amd64

jm-clius:
Fixes https://github.com/waku-org/nwaku/issues/1607

Only pass `-msse3` to the compiler (and linker) if `i386` or `amd64` and not `macosx`.

Cf. [nimbus for similar logic](https://github.com/status-im/nimbus-eth2/blob/stable/config.nims#L93-L123)

### DIFF
--- a/config.nims
+++ b/config.nims
@@ -24,17 +24,25 @@ if defined(windows):
 # engineering a more portable binary release, this should be tweaked but still
 # use at least -msse2 or -msse3.
 if defined(disableMarchNative):
-  switch("passC", "-msse3")
+  if defined(i386) or defined(amd64):
+    if defined(macosx):
+      switch("passC", "-march=haswell -mtune=generic")
+      switch("passL", "-march=haswell -mtune=generic")
+    else:
+      switch("passC", "-msse3")
+      switch("passL", "-msse3")
 elif defined(macosx) and defined(arm64):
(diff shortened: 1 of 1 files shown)


rymnc (approved):
LGTM, will have to test manually though since ci hasn't run

jakubgs (approved):
I have tested ARM compilation using Nix with this patch included and it works:
```
 > uname -a
Linux sachiel 6.0.12 #1-NixOS SMP Thu Dec 8 10:30:22 UTC 2022 aarch64 GNU/Linux

 > nix-build -A nwaku                                              
/nix/store/xa6h6aws3vmsz60w8l41iacmnljjq4dl-nwaku-0.15.0-e12b7cb4

 > file /nix/store/xa6h6aws3vmsz60w8l41iacmnljjq4dl-nwaku-0.15.0-e12b7cb4/bin/wakunode2
/nix/store/xa6h6aws3vmsz60w8l41iacmnljjq4dl-nwaku-0.15.0-e12b7cb4/bin/wakunode2: ELF 64-bit LSB executable, ARM aarch64, version 1 (SYSV), dynamically linked, interpreter /nix/store/d28cdic91i8y33297zhmm3ykvlz7md2w-glibc-2.35-224/lib/ld-linux-aarch64.so.1, for GNU/Linux 3.10.0, not stripped
```
Thanks for the fix.

LNSD (approved):
LGTM ✅

PR #1609, opened 2023-03-14
chore(networking): get streams and protos

alrevuelta:
WIP: Do not merge


alrevuelta:
cc @Menduist

waku/v2/node/peer_manager/peer_manager.nim
```diff
@@ -530,8 +556,22 @@ proc relayConnectivityLoop*(pm: PeerManager) {.async.} =
     await pm.connectToRelayPeers()
     await sleepAsync(ConnectivityLoopInterval)
 
+proc updateMetrics(pm: PeerManager) {.async.} =
+  # TODO: Hardcoded due to circular dependency
+  let WakuFilterCodec = "/vac/waku/filter/2.0.0-beta1"
```
jm-clius, line 561: This protocol will very soon be deprecated, so it may be worth rather using the `filter-subscribe` and `filter-push` protocols under filter-v2 instead.

waku/v2/node/peer_manager/peer_manager.nim
```diff
@@ -530,8 +556,22 @@ proc relayConnectivityLoop*(pm: PeerManager) {.async.} =
...
+  let WakuLightPushCodec = "/vac/waku/lightpush/2.0.0-beta1"
+
+  heartbeat "Scheduling updateMetrics run", UpdateMetricsInterval:
+    for dir in @[Direction.In, Direction.Out]:
+      for proto in @[WakuRelayCodec, WakuStoreCodec, WakuFilterCodec, WakuLightPushCodec]:
```
jm-clius, line 566: To me it's a bit weird that the peer manager would have knowledge of these protocols. Why couldn't it just report against the unique set of protocols retrievable from the `Protobook`? This way this list can dynamically change. If need be, we may even considering filtering only for protocols prefixed `/vac/waku`, so that we're sure we don't report any unrelated libp2p proto streams.

jm-clius: Another option, which may be a bit complex for this PR, could be to provide a method by which protocols can register themselves on the peer manager. That way they could not only register their codec for reporting, but perhaps also a retention policy for peers in the store (e.g. max number of filter peers to store).

jm-clius (changes requested):
Cool feature and will be great to have this accurately reported. Change request relates to knowledge of codecs within peer manager, which I think we can avoid if I'm not missing something.

waku/v2/node/peer_manager/peer_manager.nim
alrevuelta: resolving as c3d7a3b9584a7d7a64346966fb0e886a8980f18a removed this

waku/v2/node/peer_manager/peer_manager.nim
alrevuelta: good point. fixed in c3d7a3b9584a7d7a64346966fb0e886a8980f18a

now the list of protocols can dynamically change and the peer manager doesnt need knowledge of them

thanks!

waku/v2/utils/flatten.nim
```diff
@@ -0,0 +1,12 @@
+when (NimMajor, NimMinor) < (1, 4):
```
jm-clius, line 1: Perhaps `sequence.nim` for this module so that we can add other seq-related utilities here?

jm-clius (approved):
Last request before merging is to move the new `utils` to `/waku/common`. We're trying to slowly deprecate the somewhat hotchpotch mix of `utils`: if a utility is closely related to a specific module e.g. `peers`, it should be moved closer to the peer modules in future, if it is truly general or `common`, as I think both of yours are, it should go to `/waku/common`. Apologies for missing in the first review. Otherwise lgtm.

waku/v2/utils/flatten.nim
alrevuelta: sure! renamed and moved to /waku/common
53795549e81905251bd8acaf25acdb46a8001c5c

PR #1613, opened 2023-03-17
docker: add dash between target and version

jakubgs:
Instead of ugly:
```
statusteam/nim-waku:wakunode2v0.16.0-1-gb495dd
```
Create better looking:
```
statusteam/nim-waku:wakunode2-v0.16.0-1-gb495dd
```
Annoys me every time I use it.


jm-clius (approved):
Thanks!

Issue #1612, opened 2023-03-17
feat: opt-in spam protected topic with signed messages


alrevuelta:
### Problem
Publishing messages to gossipsub topics has no limits, meaning that anyone can send messages at a high rate and DoS the network. This would elevate the bandwidth consumtion of all nodes subscribed to said pubsub topic, making it prohivitive (in terms of bandwidth) to be subscribed to it.

### Suggested solution
This issue tracks a solution where certain gossipsub topics can be configured to only accept messages signed with a given key, that only trusted entities know. This key can be pre-shared among a set of participants, that are trusted to make a fair usage of the network, publishing messages at a reasonable rate/size. Note that this key can be shared/reused among multiple participants. Note that this is an opt-in solution that operators can chose to deploy in their topics, but its not enforced in the default pubsub topics that waku has.

Note also that a node only requires the public keys to validate messages, so that participation can be kept open. Only private key holders will be able to sign messages and get the gossiped in the network.

Summary of the solution:
* Make spam protected topics configurable: `[pubsub-topic-name, public-key]` via cli flags.
* Don't break any backwards compatibility with existing non protected topics, which should still allow all traffic (eg default pubsub topic)
* Add a new optional `signature` message field to include the signature of the message hash.
* All messages whose signatures weren't produced by `public-key` for the configured `topic-name` will be rejected.
* A peer sending multiple invalid messages for said topics, gets its reputation lowered and eventually is disconnected from.

### Acceptance criteria

- [x] Implement libp2p message validation for the configured topics, only accepting messaes signed with configured key. https://github.com/waku-org/nwaku/pull/1614
- [x] Add configuration via cli + metrics: https://github.com/waku-org/nwaku/pull/1696
- [x] Add metrics: https://github.com/waku-org/nwaku/pull/1704
- [x] Roll out the solution to our fleets, segregating nodes containing non protected topic.

LNSD:
I find the problem description too broad. Could you add to the problem description a specific use case that you are trying to cover with this solution? (e.g., the status community scenario or a "generic group of nodes subscribed to a pub-sub topic that share a group key pair").

I also miss a list of assumptions (e.g., a pre-shared private-public key pair) that the solution relies on. Could you add it too?

alrevuelta:
@LNSD made some claritications, hope it helps.

LNSD:
[...]

PR #1614, opened 2023-03-17
feat: dos protected topic poc

alrevuelta:
WorkInProgress

Partially addresses #1612


LNSD:
Are you planning to use the `meta` field for DoS protection? There is a discussion related to that here: https://github.com/vacp2p/rfc/pull/573#issuecomment-1446294482

> This is the case for RLN. The RLN relay validator precise a `proof` Waku message attribute. Maybe that field could be generalized/standardised in a particular manner so it can be used for different purposes.
> 
> > And for app layer tagging, I'd suggest app protocols to add this tag as part of their specification of the structure of the Waku messages payload field.
> > (Maybe this should be added to the RFC to make it explicit.)

The idea is to use a generalized version of the `proof` field for DoS purposes instead of the `meta` field which should be **an application-specific field**.

LNSD:
Additionally, the "formalization" of the `proof` field was also discussed in this PR: https://github.com/vacp2p/waku/pull/13#issuecomment-1446327988

> > In case we should need specific data to be associated with a message for relay layer DoS protections, I'd suggest introducing a new field for this.
> 
> This is the case for RLN. The RLN relay validator precise a `proof` Waku message attribute. Maybe that field could be generalized/standardised in a particular manner so it can be used for different purposes.

alrevuelta:
@LNSD Thanks for the input. I actually haven't thought about it, was using meta since I initially thought it was a good idea, but goes against `I would suggest using meta only for data the app layer wants to hand down` and I have to do a couple of tricks to make it work (eg removing meta from the hash calculation).

So I was initially using `meta` because imho this message signing shouldn't be part of the protocol, and I saw it as some kinf of application level thingy (I see the key as an application level thing) but ofc mixed with the message hash, which clearly belong to the protocol.

So I'm fine with both of your suggestions:
* Using `proof` or a generalized version of it. A variable size array with a max and just try to decode it into an rlnProof or a a signature. Following [this](https://github.com/vacp2p/waku/pull/13#issuecomment-1446327988)
* Adding a new field, `msgSignature`. However I'm not sure having a field for this would be nice, since it kind of expresses an intention, and we don't want messages to be signed.

So perhaps I would lean towards a generalized version of `proof`, but not 100% sure of the implications. cc @rymnc

tests/v2/waku_relay/test_wakunode_relay.nim
```diff
@@ -243,7 +247,219 @@ suite "WakuNode - Relay":
...
+        let unsignedMessage = WakuMessage(
+          payload: urandom(1*(10^3)),
+          contentTopic: spamProtectedTopic,
+          version: 2,
+          timestamp: now(),
+          ephemeral: true)
+        await nodes[i].publish(spamProtectedTopic, unsignedMessage, signMessage=true)
+
+    # Wait for gossip
+    await sleepAsync(500.millis)
```
Ivansete-status, line 294: The PR looks really great! 

While learning from it, I saw this "wait" that worries me a little. Would it be possible to make a more precise wait so that we make sure the node is in a valid "Gossip" state?

Having that kind of "waits" make me think about a possible random error, but of course I might be missing some details :)

tests/v2/waku_relay/test_wakunode_relay.nim
```diff
@@ -243,7 +247,219 @@ suite "WakuNode - Relay":
...
+    # Create 5 nodes
+    let nodes = toSeq(0..<5).mapIt(WakuNode.new(generateSecp256k1Key(), ValidIpAddress.init("0.0.0.0"), Port(0)))
+
+    # Protected topic and key to sign
+    let spamProtectedTopic = PubSubTopic("some-spam-protected-topic")
+    let secretKey = SkSecretKey.fromHex("5526a8990317c9b7b58d07843d270f9cd1d9aaee129294c1c478abf7261dd9e6").expect("valid key")
+    let publicKey = secretKey.toPublicKey()
+    let topicsPrivateKeys = {spamProtectedTopic: secretKey}.toTable
+
+    # Start all the nodes and mount relay with with protected topic
```
Ivansete-status, line 260: Very tiny issue: double word

jm-clius:
I think it's important that (apart from the new field) the actual _validation_ functionality gets removed from the protocol itself as much as possible. This imply some mechanism to pass validation-like functionality from the configured app (`wakunode2` in this case) down to where relay is mounted, but I don't think the protocol should be aware of how app-level validations work (or even the configured public key). I know there may be some complexities that I'm missing here in getting this concept to work, but I would strive for a simple mechanism that works in this way.

The implementation itself should be much simpler if we don't try to do signing ourselves and don't have to deal with public keys.

jm-clius:
> Note that I added message signing for completeness, since it feels weird that the node can validate signed messages but not publish them.

I think the concept here is that from the nodes perspective it simply receives messages from the app (e.g. via the API) and publishes as-is. I also don't think the "node" should therefore do the validation - just be configurable to accept validation vectors from the app of which it is agnostic, therefore allowing app-defined validation.

alrevuelta:
@jm-clius removed signing of messages and moved everything to the app.

jm-clius:
Thanks for addressing the app-node separation concerns. I think this will work and keeps the node clean of application-level knowledge. One thing to note is that this still require the inverse (node/protocol-level knowledge on the application, e.g. the app "knowing" about `node.wakuRelay` and relay validators). This may just be an OK tradeoff to keep things simple for now?

If we need a more general approach I'd suggest something like defining the concept of an app-level validation function, which the node gets constructed with. When relay then gets constructed and mounted, it "translates" these app-defined validations to relay validators and adds them to its set. This would allow the app to remain ignorant of protocol/relay concepts and the protocol/relay to remain ignorant of app concepts (as it is in your implementation).

This idea would require a bit more structure so is perhaps overkill for now, but perhaps useful if we need general app-level validation in future.

The most important concern left then, I think, is to address the naming of the field we add to the `WakuMessage`. I don't think it should be a `signature`, as the protocol should be unaware of what this field contains. What do you think about `appdata` (or any better suggestions)?

LNSD:
Please, @alrevuelta, could you answer my questions in [the associated issue](https://github.com/waku-org/nwaku/issues/1612)?

alrevuelta:
> The most important concern left then, I think, is to address the naming of the field we add to the WakuMessage. I don't think it should be a signature, as the protocol should be unaware of what this field contains. What do you think about appdata (or any better suggestions)?

Very open for namings. Have the feeling that `appdata` would be interpreted as the same as the existing `meta` defined as `# Application specific metadata.`. Perhaps it can be named `validityProof` and reused by also rln or future validators.

> This idea would require a bit more structure so is perhaps overkill for now, but perhaps useful if we need general app-level validation in future.

Since the future of this feature is not very clear and I'm also figuring things out along the way, I would prefer to prioritize validating it + moving forward with scoring + trimming connections. Actually I wouldn't mind having a dedicated branch for this complete feature, where functionality >> convenctions, processes, patterns, etc. This would allow to iterate faster and first validate the idea before spending time in things that may be discarded. Once the idea is validated with a bunch of learnings along the way, it can be materialized in production ready code. wdyt?

alrevuelta:
As agreed in a meeting, this PR will use the new meta field to store the signuture instead of creating a new field. Once the new meta is in place, this PR will use it.

alrevuelta:
Using now the `meta` field + new app hash function as discussed, tested and ready to rumble.
can I get a review @jm-clius @LNSD ?
will follow up with metrics + cli config in upcoming PRs.

apps/wakunode2/wakunode2.nim
```diff
@@ -383,6 +385,15 @@ proc setupProtocols(node: WakuNode, conf: WakuNodeConf,
...
     except CatchableError:
+
+      # Get this from cli
+      var topicsPublicKeys = initTable[string, SkPublicKey]()
+
+      # Add validation keys to protected topics
+      for topic, publicKey in topicsPublicKeys.pairs:
+        info "routing only signed traffic", topic=topic, publicKey=publicKey
+        node.wakuRelay.addSignedTopicValidator(Pubsubtopic(topic), publicKey)
+    except:
```
jm-clius, line 396: I'm pretty sure this is a faulty merge conflict resolution. This should not be part of the `except CatchableError` block and the error-less `except` in L396 should be `except CatchableError`.

jm-clius (approved):
Approving to not block and require re-review, but comment below should be addressed first - think it's just something that went wrong when merging

PR #1618, opened 2023-03-21
feat(common): added enr typed record replacement

LNSD:
The `TypedEnr` type from _nim-eth_ modules is not extensible enough. Here I introduce an alternate implementation that allows extending the field access via "getter" functions.

- [x] Turned `enr.nim` into a module with the `builder.nim` and `typed_record.nim` submodules.
- [x] Added the new `TypedRecord` type and the `toTyped()` procedure to the current ENR `Record` implementation.
- [x] Added an enum type for ENR identity scheme (`id` field)
- [x] Added the standard Ethereum field accessors: `id`, `secp2561k`, `ip/ip6`, `tcp/tcp6` and `udp/udp6`.

> **Note**
> Due to nim not having an "Any" type (the `Object` class in Java, or `interface{}` in Go), having an extensible field value caching mechanism using a `Table` is not trivial. I am leaving it for the future.

### DIFF
--- a/waku/common/enr/builder.nim
+++ b/waku/common/enr/builder.nim
@@ -1,0 +1,103 @@
+when (NimMajor, NimMinor) < (1, 4):
+  {.push raises: [Defect].}
+else:
+  {.push raises: [].}
+
+
+import
+  std/options,
+  stew/results,
+  stew/shims/net,
+  eth/keys as eth_keys,
+  eth/p2p/discoveryv5/enr,
+  libp2p/crypto/crypto as libp2p_crypto
+
+
+## Builder
+
+type EnrBuilder* = object
+    seqNumber: uint64
+    privateKey: eth_keys.PrivateKey
+    fields: seq[FieldPair]
+
+
+proc init*(T: type EnrBuilder, key: eth_keys.PrivateKey, seqNum: uint64 = 1): T =
+  EnrBuilder(
+    seqNumber: seqNum,
+    privateKey: key,
+    fields: newSeq[FieldPair]()
+  )
+
+proc init*(T: type EnrBuilder, key: libp2p_crypto.PrivateKey, seqNum: uint64 = 1): T =
+  # TODO: Inconvenient runtime assertion. Move this assertion to compile time
+  if key.scheme != PKScheme.Secp256k1:
+    raise newException(Defect, "invalid private key scheme")
+
+  let
+    bytes = key.getRawBytes().expect("Private key is valid")
+    privateKey = eth_keys.PrivateKey.fromRaw(bytes).expect("Raw private key is of valid length")
+
+  EnrBuilder.init(key=privateKey, seqNum=seqNum)
+
+proc addFieldPair*(builder: var EnrBuilder, pair: FieldPair) =
+  builder.fields.add(pair)
+
+proc addFieldPair*[V](builder: var EnrBuilder, key: string, value: V) =
+  builder.addFieldPair(toFieldPair(key, value))
+
+proc build*(builder: EnrBuilder): EnrResult[enr.Record] =
+  # Note that nim-eth's `Record.init` does not deduplicate the field pairs.
+  # See: https://github.com/status-im/nim-eth/blob/4b22fcd/eth/p2p/discoveryv5/enr.nim#L143-L144
+  enr.Record.init(
+    seqNum = builder.seqNumber,
+    pk = builder.privateKey,
+    ip = none(ValidIpAddress),
+    tcpPort = none(Port),
+    udpPort = none(Port),
+    extraFields = builder.fields
+  )
+
+
+## Builder extension: IP address and TCP/UDP ports
+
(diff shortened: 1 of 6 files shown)


waku/common/enr/typed_record.nim
```diff
@@ -0,0 +1,91 @@
...
+    return none(RecordId)
+
+  let fieldRes = toRecordId(fieldOpt.get())
+  if fieldRes.isErr():
+    return none(RecordId)
+
+  some(fieldRes.value)
+
+func secp256k1*(record: TypedRecord): Option[array[33, byte]] =
+  record.tryGet("secp256k1", array[33, byte])
```
rymnc, line 73: probably can be addressed in a follow up pr, can we write a getter for `multiaddrs` too?

waku/common/enr/builder.nim
```diff
@@ -0,0 +1,103 @@
...
+
+## Builder
+
+type EnrBuilder* = object
+    seqNumber: uint64
+    privateKey: eth_keys.PrivateKey
+    fields: seq[FieldPair]
+
+
+proc init*(T: type EnrBuilder, key: eth_keys.PrivateKey, seqNum: uint64 = 1): EnrBuilder =
```
rymnc, line 24:
```suggestion
proc init*(T: type EnrBuilder, key: eth_keys.PrivateKey, seqNum: uint64 = 1): T =
```
should work?

waku/common/enr/builder.nim
```diff
@@ -0,0 +1,103 @@
...
+    bytes = key.getRawBytes().expect("Private key is valid")
+    privateKey = eth_keys.PrivateKey.fromRaw(bytes).expect("Raw private key is of valid length")
+
+  EnrBuilder.init(key=privateKey, seqNum=seqNum)
+
+proc addFieldPair*(builder: var EnrBuilder, pair: FieldPair) =
+  builder.fields.add(pair)
+
+proc addFieldPair*[V](builder: var EnrBuilder, key: string, value: V) =
+  builder.addFieldPair(toFieldPair(key, value))
```
rymnc, line 46: we may want to add a check that `value` is not null here - this caused enrs to be misread by other implementations

rymnc (approved):
LGTM, couple queries!

waku/common/enr/typed_record.nim
LNSD: Coming in a follow-up PR, yes. It will be part of the `waku_enr` module, as it is a non-Ethereum field. It is a Waku-specific field.

waku/common/enr/builder.nim
LNSD: Yes, it should work. Thanks :)

waku/common/enr/builder.nim
LNSD: What do you mean by "is not `null`"? Could you provide a snippet illustrating the scenario where the `value` argument could be null?

rymnc: err, not directly related, but https://github.com/waku-org/nwaku/pull/1559

do you intend `addFieldPair` to only be used by the enr builder module?

LNSD: > err, not directly related, but #1559
> 
> do you intend `addFieldPair` to only be used by the enr builder module?

I see. This field should not be used outside the "extension modules". If used, it is the responsibility of the developer to perform the checks.

Regarding #1559, I plan to rework the multiaddress extension as part of the static sharding implementation. I will also add the multiaddress validation to the builder extension. Thanks for the heads-up 👍🏼

rymnc: Sounds good! Thanks

alrevuelta (approved):
lgtm.

Guess its matter of taste, but while I agree with the initial intention of making it extensible, I think its starting the get convoluted.

As an example, nimbus doesn't wrap the enr.Record type, its used in a very simple way.

Created as:
```
bootstrapEnr = enr.Record.init(
        1, # sequence number
        networkKeys.seckey.asEthKey,
        some(config.bootstrapAddress),
        some(config.bootstrapPort),
        some(config.bootstrapPort),
        [
          toFieldPair(enrForkIdField, SSZ.encode(forkId)),
          toFieldPair(enrAttestationSubnetsField, SSZ.encode(netMetadata.attnets))
        ])
```

Accessed as eg. `let eth2FieldBytes = n.record.tryGet(enrForkIdField, seq[byte])`

We can have ofc some getters to avoid further parsing, but conceptually we are using the ENR in a very similar way to ethereum, with just some extra fields on top (waku, multiaddrs and shard). Also would say 4 new files for the enr are too much.

Not action point though, just some thoughts.

jm-clius (approved):
LGTM

LNSD:
> Guess its matter of taste, but while I agree with the initial intention of making it extensible, I think its starting the get convoluted.

It is not a matter of taste. Here, I am respecting the "open-close" principle to make the code extensible (as I explained in a previous PR), and leveraging the type system to check the different field types at compilation time.

As I said in the description, the `nim-eth` implementation of the `TypedRecord` is not flexible enough: it uses explicit object fields to leverage the type system and perform compilation time checks. This is limiting for us.

PR #1621, opened 2023-03-22
bug: fix hash size greater than 32

alrevuelta:
* Uses the message `digest` to calculate the message hash when logging it.

### DIFF
--- a/waku/v2/node/waku_node.nim
+++ b/waku/v2/node/waku_node.nim
@@ -394,7 +394,7 @@ proc registerRelayDefaultHandler(node: WakuNode, topic: PubsubTopic) =
     trace "waku.relay received",
       peerId=node.peerId,
       pubsubTopic=topic,
-      hash="0x" & $topic.digest(msg).toHex(),
+      hash=topic.digest(msg).to0xHex(),
       receivedTime=getNowInNanosecondTime()
 
     waku_node_messages.inc(labelValues = ["relay"])
@@ -483,7 +483,7 @@ proc publish*(node: WakuNode, topic: PubsubTopic, message: WakuMessage) {.async,
   trace "waku.relay published",
       peerId=node.peerId,
       pubsubTopic=topic,
-      hash="0x" & $topic.digest(message).toHex(),
+      hash=topic.digest(message).to0xHex(),
       publishTime=getNowInNanosecondTime()
 
 proc startRelay*(node: WakuNode) {.async.} =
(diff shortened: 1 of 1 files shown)


waku/v2/node/waku_node.nim
```diff
@@ -392,10 +392,11 @@ proc registerRelayDefaultHandler(node: WakuNode, topic: PubsubTopic) =
     return
 
   proc traceHandler(topic: PubsubTopic, data: seq[byte]) {.async, gcsafe.} =
+    let mh = MultiHash.digest("sha2-256", data).expect("valid hash")
```
LNSD, line 395: Why don't we use the std/hash `sha256` function directly? Is there any requirement for using multihash here?

jm-clius: Yes, good catch - we should use only `sha256`. We used Multihash before to mimic what we used in the messageIdProvider, but that has also since changed to pure `sha256` to match with go-waku.

alrevuelta: btw can we use `digest` from `digest*(pubsubTopic: PubsubTopic, msg: WakuMessage): WakuMessageDigest =`? or what is stopping is from using it? `timestamp` is not included but that should be encoded in `meta`? [as per this](https://github.com/waku-org/nwaku/blob/master/waku/v2/protocol/waku_message/message.nim#L32)?

alrevuelta:
using now digest as suggested :)

waku/v2/node/waku_node.nim
```diff
@@ -391,11 +390,11 @@ proc registerRelayDefaultHandler(node: WakuNode, topic: PubsubTopic) =
   if node.wakuRelay.isSubscribed(topic):
     return
 
-  proc traceHandler(topic: PubsubTopic, data: seq[byte]) {.async, gcsafe.} =
+  proc traceHandler(topic: PubsubTopic, msg: WakuMessage) {.async, gcsafe.} =
     trace "waku.relay received",
       peerId=node.peerId,
       pubsubTopic=topic,
-      hash=MultiHash.digest("sha2-256", data).expect("valid hash").data.buffer.to0xHex(), # TODO: this could be replaced by a message UID
+      hash="0x" & $topic.digest(msg).toHex(),
```
LNSD, line 397: A minor request. There is a [`to0xHex()` proc](https://github.com/status-im/nim-stew/blob/e18f5a62af2ade7a1fd1d39635d4e04d944def08/stew/byteutils.nim#L232) in `stew/byteutils`. Could you replace the string concat here with that method?

LNSD (approved):
A minor request. In general, LGTM ✅

waku/v2/node/waku_node.nim
alrevuelta: sure! fair request. done
e41d04253d2a4c03eb72c73a603ae53f69c9af3f

PR #1622, opened 2023-03-24
fix: connect instead dial relay peer

alrevuelta:
* As detected by @lchenut we were using`dial` for connecting to relay peers.
* Using `dial` creates a stream (that can be used to read and write data)
* However since relay is based on gossipsub, the streams are created under the hood for us by libp2p. And since we just use the protocol, there is no need to create new streams, since we don't write/read data for the relay protocol.
* This PR now enforces usig `connect` when using relay protocol and `dial` for the rest.

As a rule of thumb:
* Use `dial` when a stream to read/write is required (eg all protocols but relay)

### DIFF
--- a/waku/v2/node/peer_manager/peer_manager.nim
+++ b/waku/v2/node/peer_manager/peer_manager.nim
@@ -87,29 +87,108 @@ proc insertOrReplace(ps: PeerStorage,
     warn "failed to store peers", err = res.error
     waku_peers_errors.inc(labelValues = ["storage_failure"])
 
-proc dialPeer(pm: PeerManager, peerId: PeerID,
-              addrs: seq[MultiAddress], proto: string,
-              dialTimeout = DefaultDialTimeout,
-              source = "api",
-              ): Future[Option[Connection]] {.async.} =
+proc addPeer*(pm: PeerManager, remotePeerInfo: RemotePeerInfo) =
+  # Adds peer to manager for the specified protocol
+
+  if remotePeerInfo.peerId == pm.switch.peerInfo.peerId:
+    # Do not attempt to manage our unmanageable self
+    return
+
+  # ...public key
+  var publicKey: PublicKey
+  discard remotePeerInfo.peerId.extractPublicKey(publicKey)
+
+  if pm.peerStore[AddressBook][remotePeerInfo.peerId] == remotePeerInfo.addrs and
+     pm.peerStore[KeyBook][remotePeerInfo.peerId] == publicKey:
+    # Peer already managed
+    return
+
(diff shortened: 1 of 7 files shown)


waku/v2/node/peer_manager/peer_manager.nim
```diff
@@ -87,50 +87,116 @@ proc insertOrReplace(ps: PeerStorage,
     warn "failed to store peers", err = res.error
     waku_peers_errors.inc(labelValues = ["storage_failure"])
 
-proc dialPeer(pm: PeerManager, peerId: PeerID,
-              addrs: seq[MultiAddress], proto: string,
-              dialTimeout = DefaultDialTimeout,
-              source = "api",
-              ): Future[Option[Connection]] {.async.} =
+proc addPeer*(pm: PeerManager, remotePeerInfo: RemotePeerInfo) =
```
alrevuelta, line 90: no changes here, just  moving the function

waku/v2/node/waku_node.nim
```diff
@@ -380,8 +380,8 @@ proc info*(node: WakuNode): WakuInfo =
 
 proc connectToNodes*(node: WakuNode, nodes: seq[RemotePeerInfo] | seq[string], source = "api") {.async.} =
   ## `source` indicates source of node addrs (static config, api call, discovery, etc)
-  # NOTE This is dialing on WakuRelay protocol specifically
-  await peer_manager.connectToNodes(node.peerManager, nodes, WakuRelayCodec, source=source)
+  # NOTE Connects to the node without a give protocol, which automatically creates streams for relay
```
jm-clius, line 383:
```suggestion
  # NOTE Connects to the node without a given protocol., Creates streams for any mounted pubsub protocols, e.g. `relay`
```

LNSD (approved):
LGTM ✅

waku/v2/node/peer_manager/peer_manager.nim
```diff
@@ -87,50 +87,116 @@ proc insertOrReplace(ps: PeerStorage,
...
+  var deadline = sleepAsync(dialTimeout)
+  var workfut = pm.switch.connect(peerId, peer.addrs)
   var reasonFailed = ""
+
   try:
-    if (await dialFut.withTimeout(dialTimeout)):
+    await workfut or deadline
+    if workfut.finished():
+      if not deadline.finished():
+        deadline.cancel()
```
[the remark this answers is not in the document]
alrevuelta: happy to get input on this. nothing against `withTimeout()` besides this (which imho its enough) but perhaps im missing something.

PR #1627, opened 2023-03-28
fix(rln-relay): on chain registration

rymnc:
Fixes on-chain registration with custom gas limit. Uncaught on ganache tests, need to revisit arguments later.

### DIFF
--- a/waku/v2/protocol/waku_rln_relay/group_manager/on_chain/group_manager.nim
+++ b/waku/v2/protocol/waku_rln_relay/group_manager/on_chain/group_manager.nim
@@ -111,8 +111,10 @@ method register*(g: OnchainGroupManager, identityCredentials: IdentityCredential
   var txHash: TxHash
   try: # send the registration transaction and check if any error occurs
     txHash = await rlnContract.register(idCommitment).send(value = membershipFee,
-                                                           gasPrice = gasPrice)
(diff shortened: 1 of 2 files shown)


waku/v2/protocol/waku_rln_relay/group_manager/on_chain/group_manager.nim
```diff
@@ -364,6 +366,15 @@ method init*(g: OnchainGroupManager): Future[void] {.async.} =
   let chainId = await ethRpc.provider.eth_chainId()
   g.chainId = some(chainId)
 
+  if g.ethPrivateKey.isSome():
+    let pk = g.ethPrivateKey.get()
+    let pkParseRes = keys.PrivateKey.fromHex(pk)
+    if pkParseRes.isErr():
+      raise newException(ValueError, "could not parse the private key")
+    ethRpc.privateKey = some(pkParseRes.get())
+    ethRpc.defaultAccount = ethRpc.privateKey.get().toPublicKey().toCanonicalAddress().Address
```
rymnc, line 375: weird that we need to set the defaultAccount here, without which transactions use the zero address as the from address instead of deriving it from the pk.

LNSD (approved):
LGTM ✅

jm-clius (approved):

PR #1620, opened 2023-03-22
refactor(wakunode2): encapsulate wakunode2 setup and fix warnings

LNSD:
This is a small gardening PR 🪴 

~~- [x] Fix several warnings about the `except` clause.~~
- [x] Wrap the Wakunode2 setup and main logic in two procedures.

PR #1628, opened 2023-03-29
fix(wakunode2): fix main warnings and drop swap support

LNSD:
A spin off PR from #1620:

- [x] Fix several warnings like this: `The bare except clause is deprecated; use ``except CatchableError:`` instead [BareExcept]`.
- [x] Drop apps support for Waku Swap protocol (`wakunode2` and `chat2`).


jm-clius (approved):
Thanks!

PR #1629, opened 2023-03-29
chore: increase meta size to 64 bytes + tests

alrevuelta:
Related https://github.com/vacp2p/rfc/pull/589

* Increases the maximum size if the `meta` field from 32 to 64 bytes.
* Adds a test vector the the 64 bytes case, since it has a very specific use case.
* Adds a test to ensure the message can't be decoded if `meta` field is over the limit.

### DIFF
--- a/tests/v2/test_waku_message_digest.nim
+++ b/tests/v2/test_waku_message_digest.nim
@@ -1,9 +1,12 @@
 {.used.}
 
 import
+  std/sequtils,
   stew/byteutils,
   testutils/unittests
 import
+  ../../waku/v2/protocol/waku_message,
+  ../../waku/v2/protocol/waku_message/codec,
   ../../waku/v2/protocol/waku_message/digest,
   ./testlib/waku2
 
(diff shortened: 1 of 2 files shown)


jm-clius (approved):
LGTM

tests/v2/test_waku_message_digest.nim
```diff
@@ -57,8 +64,41 @@ suite "Waku Message - Deterministic hashing":
 
     ## Then
     check:
+      byteutils.toHex(pubsubTopic.toBytes()) == "2f77616b752f322f64656661756c742d77616b752f70726f746f"
+      byteutils.toHex(message.contentTopic.toBytes()) == "2f77616b752f322f64656661756c742d636f6e74656e742f70726f746f"
+      byteutils.toHex(message.payload) == "010203045445535405060708"
+      byteutils.toHex(message.meta) == "73757065722d736563726574"
```
LNSD, line 70: I don't think this is necessary. It adds more work in case we want to update the test case 🤷🏼

LNSD (approved):
LGTM

PR #1631, opened 2023-03-30
feat(common): added postgress async pool wrapper

LNSD:
> **Warning**
> This PR is a work in progress

## Context

In this comment: https://github.com/waku-org/nwaku/pull/1590#issuecomment-1460046028
And in this issue: https://github.com/waku-org/nwaku/issues/1604

## Description

To overcome the blocking nature of the standard library's `db_postgres`, I mimicked _treeform's pg_ package implementation. I wrapped the connections in an asynchronous connection pool.

### DIFF
--- a/waku/common/postgres/asyncpool.nim
+++ b/waku/common/postgres/asyncpool.nim
@@ -1,0 +1,176 @@
+# Simple async pool driver for postgress.
+# Inspired by: https://github.com/treeform/pg/
+when (NimMajor, NimMinor) < (1, 4):
+  {.push raises: [Defect].}
+else:
+  {.push raises: [].}
+
+import
+  std/sequtils,
+  stew/results,
+  chronicles,
+  chronos
+import
+  ./common,
+  ./connection
+
+logScope:
+  topics = "postgres asyncpool"
+
+
+## Database connection pool options
+
+type PgAsyncPoolOptions* = object
+    minConnections: int
+    maxConnections: int
+
+func init*(T: type PgAsyncPoolOptions, minConnections: Positive = 1, maxConnections: Positive = 5): T =
+  if minConnections > maxConnections:
+    raise newException(Defect, "maxConnections must be greater or equal to minConnections")
+
+  PgAsyncPoolOptions(
+    minConnections: minConnections,
+    maxConnections: maxConnections
+  )
+
+func minConnections*(options: PgAsyncPoolOptions): int =
+  options.minConnections
+
+func maxConnections*(options: PgAsyncPoolOptions): int =
+  options.maxConnections
+
+
+## Database connection pool
+
+type PgAsyncPoolState {.pure.} = enum
+    Live,
+    Closing,
+    Closed
+
+type
+  ## Database connection pool
+  PgAsyncPool* = ref object
+    connOptions: PgConnOptions
+    poolOptions: PgAsyncPoolOptions
+
+    state: PgAsyncPoolState
+    conns: seq[DbConn]
+    busy: seq[bool]
+
+func isClosing*(pool: PgAsyncPool): bool =
+  pool.state == PgAsyncPoolState.Closing
+
+func isLive*(pool: PgAsyncPool): bool =
+  pool.state == PgAsyncPoolState.Live
+
+func isBusy*(pool: PgAsyncPool): bool =
+  pool.busy.allIt(it)
+
+
+
+proc close*(pool: var PgAsyncPool): Future[PgResult[void]] {.async.} =
+  ## Gracefully wait and close all openned connections
+  if pool.state == PgAsyncPoolState.Closing:
+    while true:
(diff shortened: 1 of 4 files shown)


LNSD:
I started porting [treeform's pg](https://github.com/treeform/pg) library from `asyncdispatch` to `nim-chronos`, but things went out of hand...

@arnetheduck @jm-clius @cammellos Can I get a first review from you?

waku/common/postgres/asyncpool.nim
```diff
@@ -0,0 +1,176 @@
...
+      pool.conns.add(conn)
+      pool.busy.add(true)
+
+      return ok(conn)
+    else:
+      warn "failed to stablish a new connection", msg = connRes.error
+
+  # wait for a free connection without blocking the async runtime
+  while pool.isBusy():
+    await sleepAsync(0.milliseconds)
```
Menduist, line 144: 5 ms better, I think

waku/common/postgres/asyncpool.nim
```diff
@@ -0,0 +1,176 @@
...
+    Closed
+
+type
+  ## Database connection pool
+  PgAsyncPool* = ref object
+    connOptions: PgConnOptions
+    poolOptions: PgAsyncPoolOptions
+
+    state: PgAsyncPoolState
+    conns: seq[DbConn]
```
Menduist, line 57: Personally I would have created a new type to hold the DbConn & bool instead of two seq

waku/common/postgres/connection.nim
```diff
@@ -0,0 +1,106 @@
...
+
+
+## Connection management
+
+proc error(db: DbConn): string =
+  ## Extract the error message from the database connection.
+  $pqErrorMessage(db)
+
+
+proc open*(options: PgConnOptions): common.PgResult[DbConn] =
```
Menduist, line 56: This proc is not actually async, right?

waku/common/postgres/connection.nim
```diff
@@ -0,0 +1,106 @@
...
+  if success != 1:
+    return err(db.error)
+
+  while true:
+    let success = pqconsumeInput(db)
+    if success != 1:
+      return err(db.error)
+
+    if pqisBusy(db) == 1:
+      await sleepAsync(0.milliseconds) # Do not block the async runtime
```
Menduist, line 90: That's good for a PoC, but ideally we would register the FD in chronos and wait for data to be available

waku/common/postgres/asyncpool.nim
Menduist: Actually, you can probably do a poll-free version by using an Queue that would you check in `relaseConn` (or maybe AsyncQueue)

waku/common/postgres/connection.nim
LNSD: No... It is _sync_. Just a convenience wrapper that takes the Pg options and returns a result.

waku/common/postgres/asyncpool.nim
LNSD: It's another possibility, yes 🙂 Some parts come from _treeforms_'s implementation. This is one of them.

waku/common/postgres/connection.nim
Menduist: You don't have to reimplement postgres

https://nim-lang.org/docs/postgres.html#pqsocket%2CPPGconn
This proc will give you the FD of the underlying postgres connection

You would you have to do something like:
```nim
while pqIsBusy(db) == 1:
  await waitForRead(AsyncSocket(pqsocket(db))
  pqconsumeInput(db)
```

Unfortunately, chronos doesn't has a `waitForRead`, so that would need to be added
Again, fine for a first version, but something to improve on later IMO

More info on the async usage of pg here: https://www.postgresql.org/docs/current/libpq-async.html
And an example in C : https://github.com/Menduist/sample_coroutine_api/blob/master/src/db/re_pgsql.c (callback based)

waku/common/postgres/connection.nim
Menduist: So personally I would just have a `numConnections` instead of `min/max`, don't want my program to randomly do blocking IO when I reach an arbitrary limit. Since the connections >min are not closed, you would probably end up with always `maxConnections` anyway
Though would also have to check what happens if for instance, the server restarts

LNSD: > Though would also have to check what happens if, for instance, the server restarts (or worst, just shuts down)

Absolutely! There are a thousand corner cases that I am not covering here. But for time's sake, I am trying to have a little MVP covering the "happy-path" cases, and we can iteratively improve the implementation.

Ivansete-status:
I close this PR as the issue has been addressed by the next PR, which is fundamentally inspired by that one being closed now. I submitted a different PR in order to respect the original changes

https://github.com/waku-org/nwaku/pull/1779

Thanks so much for that @LNSD !

PR #1635, opened 2023-03-31
chore: remove waku swap protocol remainings

LNSD:
With this PR, I am removing the remaining code of the Waku Swap protocol. We can always resurrect it by reverting this commit 🙂

PR #1634, opened 2023-03-31
build: added testcommon target to makefile

LNSD:
I have added a new target to the project makefile to run the `waku/common` test suites independently from the command line.

- [x] Added `testcommon` target to Makefile and `waku.nims` file.
- [x] Added `testcommon` target as a dependency for `test1` and `test2` targets. Removed the import from `all_tests_v2.nim`.

PR #1636, opened 2023-03-31
fix(rln-relay): buildscript bad cp

rymnc:
Copies the binary from the repo/target instead of package/target


rymnc:
Unblocks #1635 #1634

LNSD (approved):
LGTM ✅

Thanks!

PR #1637, opened 2023-03-31
feat: integrate new filter protocol, other improvements

jm-clius:
### What is in this PR?

This PR does the following:
- integrates the new filter service protocol in the waku node
- extend test cases to include error conditions
- some improvements to metrics, maintenance and monitoring

### What about the legacy filter protocol?

For now, when filter is configured, the node will mount _both_ the new and legacy filter protocols. This is to maintain backwards compatibility until filter clients have dogfooded and switched to the new filter protocol. The idea is to remove all legacy code in the future.

### What about the nwaku filter client?

Note that the new filter client has not been integrated yet. Since nwaku is rarely used as a filter client, this client protocol can be integrated and the Filter APIs updated once dogfooding has progressed without supporting the legacy filter client.


jm-clius:
Update: although tests pass locally, they fail during CI due to a segfault getting triggered on the underlying peerStore attempting to access its `peerStore.books` field. At first glance this should be impossible, but I'm investigating.

waku/v2/protocol/waku_filter/rpc.nim
jm-clius, line 1: Whitespace changes only. Hide whitespace in GH settings to ignore this file.

waku/v2/protocol/waku_filter/protocol.nim
jm-clius, line 1: Hide whitespace in GH settings to view only effective changes in this file.

LNSD (approved):
LGTM ✅ 💯

PR #1639, opened 2023-03-31
build(deps): upgrade nim-chronos and nim-presto to latest version

LNSD:
Bump `nim-presto` to the latest version


jm-clius:
Probably requires bumping nim-chronos too (and possibly other submodules).

LNSD:
> Probably requires bumping nim-chronos too (and possibly other submodules).

It seems that there was some dependency on `nim-chonos` latest release. As you pointed out 👍🏼

jm-clius (approved):

PR #1640, opened 2023-03-31
Bump container image versions to v0.16.0 in quickstart

### LINKED ISSUE #1667: chore: Update quickstart with v0.16.0
### Background
Update operator quickstarts (https://github.com/waku-org/nwaku/blob/master/docs/operators/docker-quickstart.md) to point to the latest release (v0.16.0)

### Details
The quickstarts points to outdated release. Would be good to update it.

### Acceptance criteria

- [ ]  Quickstart docs point to current latest release

vpavlin:
closes #1667


jm-clius (approved):
Thanks for this! I agree that we could use `latest` here, though this is an example on how to use a specific release. I don't necessarily think we need to update it every time, though it's good that it's not way too old.

PR #1642, opened 2023-04-04
bug: bump libp2p with traffic metrics fix

alrevuelta:
* This PR contains a nim-libp2p version bump, containing an important fix included in https://github.com/status-im/nim-libp2p/pull/879
* It fixes a bug in the `in` `libp2p_network_bytes` metric, used to calculate the incoming bandwidth of the node. Due to this bug, this metric was reporting values of x5 times (or even more) the real bandwidth of the node.
* After this fix, both `libp2p_network_bytes` `in/out` metrics now match with other ones (provided by cadvisor). So now we should have better obseravbility on the consumed bandwdith on nwaku nodes.


jm-clius (approved):

jm-clius:
Btw, we've not been very consistent with this, but we try to stick to the prefixes defined by semantic commits/conventional commits for commit messages (i.e. `fix:` rather than `bug:`): https://www.conventionalcommits.org/en/v1.0.0/ This will help us write e.g. automated release tools.

alrevuelta:
You are very right! Fixed.

PR #1645, opened 2023-04-04
fix: fixed multiple bare except warnings

LNSD:
This is a gardening PR 🪴 This fixes multiple `BareExcept` warnings. Some others cannot be fixed without changing the function, which raises the exception: this should be addressed in a separate PR.


jm-clius (approved):
Thanks for this!

alrevuelta (approved):

PR #1646, opened 2023-04-04
refactor(node): added waku_node barrel import and split config module

LNSD:
Since we have new colleagues in the nwaku team, it makes more sense to continue the work I started some time ago of chunking the `waku_node` module. These changes aim to split some parts (e.g., the network configuration type).

- [x] Added a `waku_node` barrel import at `waku/v2/waku_node`. It re-exports the `WakuNode` types, the `WakuNetConfig` configuration, and the `WakuSwitch` types.
- [x] Updated the test and some imports accordingly.
- [x] Removed unnecessary or outdated files from the `waku/v2/node` folder.
- [x] Moved `NetConfig.init` from exceptions to results. Updated the code accordingly.

More changes will follow these (e.g., a WakuNode builder). This is the initial PR to establish the "barrel" import module.


waku/v2/node/waku_node.nim
```diff
@@ -247,7 +127,7 @@ proc getEnr*(netConfig: NetConfig,
 
   return ok(recordRes.get())
 
-proc getAutonatService*(rng = crypto.newRng()): AutonatService =
+proc getAutonatService*(rng: ref HmacDrbgContext): AutonatService =
```
Ivansete-status, line 130: The PR looks awesome! For learning purposes, what is the benefit of this? Does it go to the heap instead of the stack?

Ivansete-status (approved):

waku/v2/node/waku_node.nim
LNSD: The `ref HmacDrbgContext` type is a reference of a random number generator (usually, when you use the `ref` keyword, it is allocated at the heap). This is of type `ref` because we want to instantiate and share a single RNG instance.

With these changes, I am removing the proc's argument's default value (a new instance of RNG). I want it to be explicitly passed to the method.

The relevant part of the Nim documentation: [here](https://nim-lang.github.io/Nim/manual.html#procedures:~:text=A%20parameter%20may%20be%20declared%20with%20a%20default%20value%20which%20is%20used%20if%20the%20caller%20does%20not%20provide%20a%20value%20for%20the%20argument.%20The%20value%20will%20be%20reevaluated%20every%20time%20the%20function%20is%20called.)

Ivansete-status: Understood thanks!

alrevuelta (approved):

PR #1649, opened 2023-04-05
feat(node): added waku node builder type

LNSD:
This PR introduces a `WakuNode` compatible builder. The idea is to establish here the builder type and then evolve it in future iterations. This is just the first PR introducing the type and the API. It maps the `WakuNode.new` to the `WakuNodeBuilder` API.

- [x] Added the `WakuNodeBuilder` type and methods.
- [x] Added a set of `WakuNode.new` compatible `withXXX()` methods.
- [x] Moved all apps, tools, and examples to the builder pattern.

### DIFF
--- a/waku/v2/node/builder.nim
+++ b/waku/v2/node/builder.nim
@@ -1,0 +1,167 @@
+when (NimMajor, NimMinor) < (1, 4):
+  {.push raises: [Defect].}
+else:
+  {.push raises: [].}
+
+import
+  std/options,
+  stew/results,
+  stew/shims/net,
+  chronicles,
+  libp2p/crypto/crypto,
+  libp2p/builders,
+  libp2p/nameresolving/nameresolver
+import
+  ../protocol/waku_enr,
+  ../protocol/waku_discv5,
+  ./config,
+  ./peer_manager,
+  ./waku_node
+
+
+type
+  WakuNodeBuilder* = object
+    # General
+    nodeRng: Option[ref crypto.HmacDrbgContext]
+    nodeKey: Option[crypto.PrivateKey]
+    netConfig: Option[NetConfig]
+
+    # Peer storage and peer manager
+    peerStorage: Option[PeerStorage]
+    peerStorageCapacity: Option[int]
+    peerManager: Option[PeerManager]
+
+    # Libp2p switch
+    switchMaxConnections: Option[int]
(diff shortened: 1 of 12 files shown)


waku/v2/node/builder.nim
```diff
@@ -0,0 +1,167 @@
...
+  builder.switchAgentString = agentString
+
+  if not nameResolver.isNil():
+    builder.switchNameResolver = some(nameResolver)
+
+
+## Waku discv5
+
+proc withWakuDiscv5*(builder: var WakuNodeBuilder, instance: WakuDiscoveryV5) =
+  if not instance.isNil():
```
Ivansete-status, line 129: Should we do anything for the case 'isNil' ?

LNSD: I added this for compatibility reasons (with the deprecated `WakuNode.new()` API). Accepting a `nil` makes 8 test cases crash 😅

waku/v2/node/builder.nim
```diff
@@ -0,0 +1,167 @@
...
+    dns4DomainName = dns4DomainName,
+    discv5UdpPort = discv5UdpPort,
+  )
+  builder.withNetworkConfiguration(netConfig)
+  ok()
+
+
+## Peer storage and peer manager
+
+proc withPeerStorage*(builder: var WakuNodeBuilder, peerStorage: PeerStorage, capacity = none(int)) =
```
alrevuelta, line 97: perhaps merge with withPeerManager?

alrevuelta (approved):
lgtm! really like the pattern

LNSD:
Just for the record, this PR is an intermediate step of a more extensive refactoring. 

The builder APIs here are not final. Some of the APIs will change after the `WakuNode` procedures change.

waku/v2/node/builder.nim
LNSD: Yes! This is the idea. It will be merged after we can change the `WakuNode.new()` proc without having to change 120 test cases :)

LNSD:
I have added some changes to the PR that will make it easier to understand than in the vacuum:

- [x] Moved all apps, tools, and examples to the builder pattern.

jm-clius (approved):
LGTM! Great improvement. Thanks.

richard-ramos:
Cool PR! :100:

PR #1650, opened 2023-04-05
chore: do not mount relay more than once

alrevuelta:
* Prevents the relay from being mounted more than once.
* This prevents mounting more relay protocols in the switch, which created unnecessary streams under the hood.


LNSD (approved):
LGTM ✅

alrevuelta:
@jm-clius I found this issue while testing, but i think its a nice to have due to the implications it has.

what do you mean by MSTM? can't find it online.

jm-clius:
> but i think its a nice to have due to the implications it has

Indeed. I was just worried that there may be some bug in `wakunode2` that causes relay to be mounted twice, which would be unexplained behaviour. It makes sense as a safety check.

Oh, MSTM = Makes Sense To Me. I thought it was used more widely, but I also now can't find it. 😄

Approving.

jm-clius (approved):

PR #1651, opened 2023-04-05
refactor(node): remove deprecated constructor and extend testlib with builder

LNSD:
As most of the test cases depend on the `WakuNode.new()` constructor, it is not easy to refactor and remove the deprecated constructor.

- [x] Added a `testlib/wakunode` submodule. It is not possible to add these methods to the already existing `testlib/waku2` submodule without triggering some cryptic and unrelated compilation errors.
- [x] Renamed `testlib/waku2` to `testlib/wakucore`, for consistency with the `testlib/wakunode` subpackage.
- [x] Added a  shim`newTestWakuNode` method to the `testlib` module that acts as a `WakuNodeBuilder` wrapper.
- [x] Updated all test cases and imports accordingly. 
- [x] Removed the deptercated `WakuNode.new()` method.

### DIFF
--- a/tests/v2/testlib/wakunode.nim
+++ b/tests/v2/testlib/wakunode.nim
@@ -1,0 +1,73 @@
+import
+  std/options,
+  stew/results,
+  stew/shims/net,
+  chronos,
+  libp2p/switch,
+  libp2p/builders,
+  libp2p/nameresolving/nameresolver,
+  libp2p/crypto/crypto as libp2p_keys,
+  eth/keys as eth_keys
+import
+  ../../../waku/v2/waku_node,
+  ../../../waku/v2/node/peer_manager,
+  ../../../waku/v2/protocol/waku_enr,
+  ../../../waku/v2/protocol/waku_discv5,
+  ./common
+
+
+# Waku node
+
+proc newTestWakuNode*(nodeKey: crypto.PrivateKey,
+                      bindIp: ValidIpAddress,
+                      bindPort: Port,
+                      extIp = none(ValidIpAddress),
+                      extPort = none(Port),
+                      extMultiAddrs = newSeq[MultiAddress](),
+                      peerStorage: PeerStorage = nil,
+                      maxConnections = builders.MaxConnections,
+                      wsBindPort: Port = (Port)8000,
+                      wsEnabled: bool = false,
+                      wssEnabled: bool = false,
+                      secureKey: string = "",
+                      secureCert: string = "",
+                      wakuFlags = none(CapabilitiesBitfield),
+                      nameResolver: NameResolver = nil,
+                      sendSignedPeerRecord = false,
+                      dns4DomainName = none(string),
+                      discv5UdpPort = none(Port),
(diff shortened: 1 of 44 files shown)


tests/v2/testlib/wakunode.nim
```diff
@@ -0,0 +1,73 @@
...
+  ../../../waku/v2/waku_node,
+  ../../../waku/v2/node/peer_manager,
+  ../../../waku/v2/protocol/waku_enr,
+  ../../../waku/v2/protocol/waku_discv5,
+  ./common
+
+
+# Waku node
+
+proc newTestWakuNode*(nodeKey: crypto.PrivateKey,
```
alrevuelta, line 21: mind in another PR adding some default?

so for exameple the boilerplate widely used:
newTestWakuNode(generateSecp256k1Key(), ValidIpAddress.init("0.0.0.0"), Port(0))

becomes just newTestWakuNode, since most of the times we are using 0.0.0.0, port 0 and a random key?

alrevuelta (approved):
lgtm! left a comment, but feel free to address it in other prs.

tests/v2/testlib/wakunode.nim
LNSD: Makes sense. Feel free to open a PR and add it :)

tests/v2/testlib/wakunode.nim
```diff
@@ -0,0 +1,73 @@
...
+    extMultiAddrs = extMultiAddrs,
+    wsBindPort = wsBindPort,
+    wsEnabled = wsEnabled,
+    wssEnabled = wssEnabled,
+    wakuFlags = wakuFlags,
+    dns4DomainName = dns4DomainName,
+    discv5UdpPort = discv5UdpPort,
+  )
+  if netConfigRes.isErr:
+    raise newException(Defect, "Invalid network configuration")
```
Ivansete-status, line 56: Shall we have $netConfigRes.error appended in this point?

Also, I wonder if we could have a future PR where we convert all the "raise exception" approaches into Result[...] ones.

Ivansete-status (approved):

tests/v2/testlib/wakunode.nim
LNSD: > Also, I wonder if we could have a future PR where we convert all the "raise exception" approaches into Result[...] ones.

The problem is how the node instance creation, the `newTestWakuNode()`, is used around the test suites. The test cases expect either the `WakuNode.new()` method to succeed and return the instance or raise an exception. Changing the API for a result would mean many more complex changes in the test suites' codebase.

I defend that, in general, `Result`s should prevail over exceptions. This is true for "production code". 

In this case, these changes are on the test suites codebase, so it is ok to raise a `Defect` (the equivalent of a `panic` in Rust/Go) or an exception if there is any issue with the instantiation of the node as it will halt the test case and it will mark it as failed.

LNSD, line 56:
```suggestion
  if netConfigRes.isErr():
    raise newException(Defect, "Invalid network configuration: " & $netConfigRes.error)
```

PR #1653, opened 2023-04-06
fix(wakunode2): made setup nat return errors

LNSD:
The `setupNat` method from `waku/common/utils/nat` module is unsuitable for production. Instead of returning an error, it quits the application with return code `1`.

- [x] Deprecated the `waku/common/nat#setupNat()` method.
- [x] Added an error returning version of `setupNat` to `wakunode2` module (as a temporary solution).

This has been discovered during the refactoring process of the WakuNode and the Wakunode2 app.

### DIFF
--- a/apps/wakunode2/wakunode2.nim
+++ b/apps/wakunode2/wakunode2.nim
@@ -12,6 +12,7 @@ import
   libbacktrace,
   system/ansi_c,
   eth/keys,
+  eth/net/nat,
   eth/p2p/discoveryv5/enr,
   libp2p/builders,
   libp2p/multihash,
@@ -23,7 +24,6 @@ import
   libp2p/nameresolving/dnsresolver
 import
   ../../waku/common/sqlite,
-  ../../waku/common/utils/nat,
   ../../waku/common/logging,
   ../../waku/v2/node/peer_manager,
   ../../waku/v2/node/peer_manager/peer_store/waku_peer_storage,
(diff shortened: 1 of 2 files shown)


apps/wakunode2/wakunode2.nim
```diff
@@ -214,6 +214,51 @@ proc retrieveDynamicBootstrapNodes*(dnsDiscovery: bool, dnsDiscoveryUrl: string,
...
+      # because it obtains the address of a non-gcsafe proc?
+      var extPorts: Option[(Port, Port)]
+      try:
+        extPorts = ({.gcsafe.}: redirectPorts(tcpPort = tcpPort,
+                                              udpPort = udpPort,
+                                              description = clientId))
+      except CatchableError:
+        # TODO: nat.nim Error: can raise an unlisted exception: Exception. Isolate here for now.
+        error "unable to determine external ports"
+        extPorts = none((Port, Port))
```
Ivansete-status, line 243: Should it return `error(...)` in this case ? I'm afraid it'd return `ok(...)`

LNSD: It sets the `extPorts` to none, a valid state. So no need to break the execution here.

waku/common/utils/nat.nim
```diff
@@ -14,14 +14,15 @@ logScope:
 proc setupNat*(natConf, clientId: string, tcpPort, udpPort: Port):
     tuple[ip: Option[ValidIpAddress],
           tcpPort: Option[Port],
-          udpPort: Option[Port]] {.gcsafe.} =
+          udpPort: Option[Port]] {.gcsafe, deprecated:
+  "this methos quits the app if something is not ok".} =
```
Ivansete-status, line 18: tiny typo

apps/wakunode2/wakunode2.nim
```diff
@@ -214,6 +214,51 @@ proc retrieveDynamicBootstrapNodes*(dnsDiscovery: bool, dnsDiscoveryUrl: string,
...
+
+      if extPorts.isSome():
```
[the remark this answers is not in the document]
LNSD: AFAICT, this is a valid state. This piece of code is just mapping the `tuple(option(port), option(port))` to `option(tuple(port, port))

jm-clius (approved):
Thanks!

PR #1657, opened 2023-04-07
refactor(wakunode2): split setup logic into app module (pt. 3)

LNSD:
This is the third PR of the `refactor(wakunode2): split setup logic into app module` refactoring series.

> **Note**
> This PR is not merging into the `master` branch. It is merging into `wakunode-develop`.

- [x] Moved some of the setup methods from the main file to the app module:
	+ The protocols mounting logic


apps/wakunode2/app.nim
```diff
@@ -31,6 +31,12 @@ import
   ../../waku/v2/protocol/waku_dnsdisc,
   ../../waku/v2/protocol/waku_enr,
   ../../waku/v2/protocol/waku_discv5,
+  ../../waku/v2/protocol/waku_peer_exchange,
```
Ivansete-status, line 34: LGTM! We may be able to delete these imports from the wakunode2.nim at some point in future PR :)

Ivansete-status (approved):

apps/wakunode2/app.nim
LNSD: Yes, that's the idea 😁

PR #1658, opened 2023-04-07
refactoring: Adding parsePeerInfo and deprecating 'parseRemotePeerInfo'

Ivansete-status:
As per @LNSD's recommendation, we are adding a 'parseRemotePeerInfo'-equivalent function that doesn't raises any expection but deals with Restult[...] which is a better approach to handle errors.

### DIFF
--- a/tests/v2/test_peer_manager.nim
+++ b/tests/v2/test_peer_manager.nim
@@ -4,7 +4,6 @@ import
   std/[options, sequtils],
   stew/shims/net as stewNet,
   testutils/unittests,
-  chronicles,
   chronos,
   json_rpc/rpcserver,
   json_rpc/rpcclient,
@@ -76,7 +75,10 @@ procSuite "Peer Manager":
     await allFutures(nodes.mapIt(it.start()))
     await allFutures(nodes.mapIt(it.mountRelay()))
 
-    let nonExistentPeer = parseRemotePeerInfo("/ip4/0.0.0.0/tcp/1000/p2p/16Uiu2HAmL5okWopX7NqZWBUKVqW8iUxCEmd5GMHLVPwCgzYzQv3e")
+    let nonExistentPeerRes = parsePeerInfo("/ip4/0.0.0.0/tcp/1000/p2p/16Uiu2HAmL5okWopX7NqZWBUKVqW8iUxCEmd5GMHLVPwCgzYzQv3e")
+    require nonExistentPeerRes.isOk()
+
+    let nonExistentPeer = nonExistentPeerRes.value
 
     # Dial non-existent peer from node1
     let conn1 = await nodes[0].peerManager.dialPeer(nonExistentPeer, WakuFilterCodec)
@@ -136,9 +138,14 @@ procSuite "Peer Manager":
       nodes[0].peerManager.peerStore.connectedness(nodes[1].peerInfo.peerId) == NotConnected
 
     # Failed connection
-    let nonExistentPeer = parseRemotePeerInfo("/ip4/0.0.0.0/tcp/1000/p2p/16Uiu2HAmL5okWopX7NqZWBUKVqW8iUxCEmd5GMHLVPwCgzYzQv3e")
+    let nonExistentPeerRes = parsePeerInfo("/ip4/0.0.0.0/tcp/1000/p2p/16Uiu2HAmL5okWopX7NqZWBUKVqW8iUxCEmd5GMHLVPwCgzYzQv3e")
+    require:
+      nonExistentPeerRes.isOk()
+
+    let nonExistentPeer = nonExistentPeerRes.value
     require:
       (await nodes[0].peerManager.connectRelay(nonExistentPeer)) == false
+
     check:
       # Cannot connect to node2
       nodes[0].peerManager.peerStore.connectedness(nonExistentPeer.peerId) == CannotConnect
@@ -165,7 +172,10 @@ procSuite "Peer Manager":
     await allFutures(nodes.mapIt(it.start()))
     await allFutures(nodes.mapIt(it.mountRelay()))
 
-    let nonExistentPeer = parseRemotePeerInfo("/ip4/0.0.0.0/tcp/1000/p2p/16Uiu2HAmL5okWopX7NqZWBUKVqW8iUxCEmd5GMHLVPwCgzYzQv3e")
+    let nonExistentPeerRes = parsePeerInfo("/ip4/0.0.0.0/tcp/1000/p2p/16Uiu2HAmL5okWopX7NqZWBUKVqW8iUxCEmd5GMHLVPwCgzYzQv3e")
+    require nonExistentPeerRes.isOk()
+
+    let nonExistentPeer = nonExistentPeerRes.value
 
     nodes[0].peerManager.addPeer(nonExistentPeer)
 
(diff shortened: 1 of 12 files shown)


waku/v2/utils/peers.nim
```diff
@@ -112,7 +112,49 @@ func getTransportProtocol(typedR: TypedRecord): Option[IpTransportProtocol] =
...
+        wsPart = addrPart.tryGet()
+      of "wss":
+        wssPart = addrPart.tryGet()
+
+    # nim-libp2p dialing requires remote peers to be initialised with a peerId and a wire address
+    let
+      peerIdStr = p2pPart.toString()[].split("/")[^1]
+      wireAddr = nwPart & tcpPart & wsPart & wssPart
+
+    if (not wireAddr.validWireAddr()):
```
rymnc, line 141:
```suggestion
    if not wireAddr.validWireAddr():
```

waku/v2/utils/peers.nim
```diff
@@ -112,7 +112,49 @@ func getTransportProtocol(typedR: TypedRecord): Option[IpTransportProtocol] =
...
+      return err("Invalid node multi-address")
```
rymnc, line 142:
```suggestion
      return err("Invalid node multiaddress")
```

rymnc (approved):
LGTM, minor nits

waku/v2/utils/peers.nim
Ivansete-status: Thanks for the suggestions @rymnc ! I've applied them.

waku/v2/utils/peers.nim
```diff
@@ -112,7 +112,49 @@ func getTransportProtocol(typedR: TypedRecord): Option[IpTransportProtocol] =
...
+
+    return ok(RemotePeerInfo.init(peerIdStr, @[wireAddr]))
+  except CatchableError:
+    return err(getCurrentExceptionMsg())
+
+## Parses a fully qualified peer multiaddr, in the
+## format `(ip4|ip6)/tcp/p2p`, into dialable PeerInfo
+## Notice that we deprecated this function in favor
+## of 'parsePeerInfo' which does the same but just
+## returs a Result[...] instead of raising an exception.
```
LNSD, line 152: I know that this is not part of your changes, but JFYI: In nim, documentation goes inside of the function (similar to Python): https://nim-lang.org/docs/docgen.html#introduction-documentation-comments

waku/v2/utils/peers.nim
```diff
@@ -112,7 +112,49 @@ func getTransportProtocol(typedR: TypedRecord): Option[IpTransportProtocol] =
...
+proc parseRemotePeerInfo*(address: string):
+          RemotePeerInfo
+            {.raises: [Defect, ValueError, LPError],
+            deprecated: "Raises exceptions which is difficult to maintain".}=
```
LNSD, line 156: Something I would do to avoid having duplicate functionality and to test that the function is working as expected is: 

> Call inside of the deprecated method the result method, and in the case, the result returned is an error, we raise a `ValueError`. 

With this approach, the deprecated method acts as a shim to the new result-returning implementation and we can detect with the tests if we missed anything (e.g., an unexpected exception) on the new implementation.

waku/v2/utils/peers.nim
```diff
@@ -112,7 +112,49 @@ func getTransportProtocol(typedR: TypedRecord): Option[IpTransportProtocol] =
...
 ## Parses a fully qualified peer multiaddr, in the
 ## format `(ip4|ip6)/tcp/p2p`, into dialable PeerInfo
-proc parseRemotePeerInfo*(address: string): RemotePeerInfo {.raises: [Defect, ValueError, LPError].}=
+proc parsePeerInfo*(address: string):
+                    Result[RemotePeerInfo, string] =
+  var
+    nwPart, tcpPart, p2pPart, wsPart, wssPart: MultiAddress
+
+  try:
+    let multiAddr = MultiAddress.init(address).tryGet()
```
LNSD, line 121: Note that the `tryGet()` method is a `stew/results` package procedure/template that converts an error into an exception (similar to `unwrap` in Rust).

Here the correct thing would be:

- Either bubble up the error using the `?` operator:
```suggestion
    let multiAddr = ? MultiAddress.init(address)
```
- Or bubble up the error with a wrapped error info via `?` operator and the `mapErr` method:

```suggestion
    let multiAddr = ? MultiAddress.init(address).mapErr(proc(err: string): string = "invalid multiaddress format: " & err)
```

The option depends on what error message do you want to provide.

waku/v2/utils/peers.nim
```diff
@@ -112,7 +112,49 @@ func getTransportProtocol(typedR: TypedRecord): Option[IpTransportProtocol] =
...
+      of "ip4", "ip6", "dns", "dnsaddr", "dns4", "dns6":
+        nwPart = addrPart.tryGet()
+      of "tcp":
+        tcpPart = addrPart.tryGet()
+      of "p2p":
+        p2pPart = addrPart.tryGet()
+      of "ws":
+        wsPart = addrPart.tryGet()
+      of "wss":
+        wssPart = addrPart.tryGet()
```
LNSD, line 134: Same here. Do not use `tryGet()`, bubble up the error.

waku/v2/utils/peers.nim
```diff
@@ -112,7 +112,49 @@ func getTransportProtocol(typedR: TypedRecord): Option[IpTransportProtocol] =
...
+    let
+      peerIdStr = p2pPart.toString()[].split("/")[^1]
+      wireAddr = nwPart & tcpPart & wsPart & wssPart
+
+    if not wireAddr.validWireAddr():
+      return err("Invalid node multiaddress")
+
+    return ok(RemotePeerInfo.init(peerIdStr, @[wireAddr]))
+  except CatchableError:
+    return err(getCurrentExceptionMsg())
```
LNSD, line 146: As a rule of thumb, it is a good practice to keep `try/except` clauses as narrow as possible.

Said that, if you bubble up the errors using the `?` operator, I think there is no exception source so maybe you can remove the `try/except` clause.

LNSD (changes requested):
Please, check my comments 🙂

waku/v2/utils/peers.nim
Ivansete-status: Yes, absolutely right!

waku/v2/utils/peers.nim
Ivansete-status: I like more the second option, which gives more info.

waku/v2/utils/peers.nim
Ivansete-status: Sorry I misunderstood this recommendation and I made a wider refactoring. With that I think the error cases are covered better and it gives better feedback.

Of course, if you don't agree with the changes I will roll them back and just call the "return" function inside the "exception" one as you stated in the comment :)

tests/v2/test_peer_manager.nim
```diff
@@ -520,23 +534,25 @@ procSuite "Peer Manager":
       storage = nil)
 
     # Create 15 peers and add them to the peerstore
-    let peers = toSeq(1..15).mapIt(parseRemotePeerInfo("/ip4/0.0.0.0/tcp/0/p2p/" & $PeerId.random().get()))
-    for p in peers: pm.addPeer(p)
+    let peers = toSeq(1..15).mapIt(parsePeerInfo("/ip4/0.0.0.0/tcp/0/p2p/" & $PeerId.random().get()))
+    for p in peers:
+      require p.isOk()
+      pm.addPeer(p.value)
```
LNSD, line 540:
```suggestion
    let peers = toSeq(1..15)
    	.mapIt(parsePeerInfo("/ip4/0.0.0.0/tcp/0/p2p/" & $PeerId.random().get()))
    	.filterIt(it.isOk()).mapIt(it.value)
```

waku/v2/utils/peers.nim
```diff
@@ -110,37 +110,59 @@ func getTransportProtocol(typedR: TypedRecord): Option[IpTransportProtocol] =
...
+  ## format `(ip4|ip6)/tcp/p2p`, into dialable PeerInfo
+  ## Notice that we deprecated this function in favor
+  ## of 'parsePeerInfo' which does the same but just
+  ## returs a Result[...] instead of raising an exception.
+
+  let remoteInfo = parsePeerInfo(address)
+  if remoteInfo.isOk():
+    return remoteInfo.value
+  else:
+    raise newException(ValueError, remoteInfo.error)
```
LNSD, line 164: If you removed all usages of this proc, we could remove it.

LNSD (approved):
LGTM ✅ 

A couple of minor comments

waku/v2/node/waku_node.nim
```diff
@@ -427,10 +427,12 @@ proc filterSubscribe*(node: WakuNode, pubsubTopic: PubsubTopic, contentTopics: C
     error "cannot register filter subscription to topic", error="waku filter client is nil"
     return
 
-  let remotePeer = when peer is string: parseRemotePeerInfo(peer)
-                   else: peer
+  var remotePeer = parsePeerInfo(peer)
```
jm-clius, line 430: Any reason this is a `var` and not `let`?

jm-clius (approved):
LGTM. Thanks for this!

waku/v2/node/waku_node.nim
Ivansete-status: ah yes, you are right, thanks !

tests/v2/test_peer_manager.nim
```diff
@@ -76,10 +76,11 @@ procSuite "Peer Manager":
     await allFutures(nodes.mapIt(it.start()))
     await allFutures(nodes.mapIt(it.mountRelay()))
 
-    let nonExistentPeer = parseRemotePeerInfo("/ip4/0.0.0.0/tcp/1000/p2p/16Uiu2HAmL5okWopX7NqZWBUKVqW8iUxCEmd5GMHLVPwCgzYzQv3e")
+    let nonExistentPeer = parsePeerInfo("/ip4/0.0.0.0/tcp/1000/p2p/16Uiu2HAmL5okWopX7NqZWBUKVqW8iUxCEmd5GMHLVPwCgzYzQv3e")
+    require nonExistentPeer.isOk()
 
     # Dial non-existent peer from node1
-    let conn1 = await nodes[0].peerManager.dialPeer(nonExistentPeer, WakuFilterCodec)
+    let conn1 = await nodes[0].peerManager.dialPeer(nonExistentPeer.value, WakuFilterCodec)
```
alrevuelta, line 83: mind doing the following?

```
let nonExistentPeerRes = parsePeerInfo("/ip4/0.0.0.0/tcp/1000/p2p/16Uiu2HAmL5okWopX7NqZWBUKVqW8iUxCEmd5GMHLVPwCgzYzQv3e")

... checks

let nonExistentPeer = nonExistentPeerRes.value
```


Like this you dont need to have the boilerplate `.value` `.value`, so less lines of code to modify, less diff and perhaps clearer?

We kind of have this pattern in other places. `xxxRes`, then checks and then `xxx = xxxRes.value` (or get)

tests/v2/test_peer_manager.nim
```diff
@@ -413,37 +416,44 @@ procSuite "Peer Manager":
...
-    node.peerManager.addServicePeer(peer5, WakuRelayCodec)
+    node.peerManager.addServicePeer(peer5.value, WakuRelayCodec)
 
     # all peers are stored in the peerstore
     check:
-      node.peerManager.peerStore.peers().anyIt(it.peerId == peer1.peerId)
-      node.peerManager.peerStore.peers().anyIt(it.peerId == peer2.peerId)
-      node.peerManager.peerStore.peers().anyIt(it.peerId == peer3.peerId)
-      node.peerManager.peerStore.peers().anyIt(it.peerId == peer4.peerId)
+      node.peerManager.peerStore.peers().anyIt(it.peerId == peer1.value.peerId)
```
alrevuelta, line 443: see my other comments for all this file. instead of having `.value` perhaps use `xxRes`

tests/v2/test_peer_manager.nim
Ivansete-status: I will do that thanks !

alrevuelta (approved):
lgtm. thanks for the changes!

PR #1660, opened 2023-04-07
refactor(wakunode2): split setup logic into app module (pt. 5)

LNSD:
This is the fifth PR of the `refactor(wakunode2): split setup logic into app module` refactoring series.

> **Note**
> This PR is not merging into the `master` branch. It is merging into `wakunode-develop`.

- [x] Moved some of the setup methods from the main file to the app module:
	+ Start HTTP JSON-RPC server procedure
	+ Start HTTP REST server procedure
	+ Start metrics server procedure
	+ Start console metrics procedure

### DIFF
--- a/apps/wakunode2/wakunode2.nim
+++ b/apps/wakunode2/wakunode2.nim
@@ -47,26 +47,17 @@ import
   ../../waku/v2/protocol/waku_peer_exchange,
   ../../waku/v2/protocol/waku_relay/validators,
   ../../waku/v2/utils/peers,
-  ./wakunode2_setup_rest,
-  ./wakunode2_setup_rpc,
   ./config
-
-when defined(rln):
-  import
-    ../../waku/v2/protocol/waku_rln_relay
-
 
 logScope:
   topics = "wakunode"
-
-type SetupResult[T] = Result[T, string]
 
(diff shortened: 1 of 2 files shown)


apps/wakunode2/wakunode2.nim
```diff
@@ -47,25 +47,18 @@ import
...
-type SetupResult[T] = Result[T, string]
-
 # Temporarily merge `app.nim` into `wakunode2.nim` until we encapsulate the
 #  state and the logic in an object instance.
 include ./app
 
 when defined(waku_exp_store_resume):
+  type SetupResult[T] = Result[T, string]
+
   proc resumeMessageStore(node: WakuNode, address: string): Future[SetupResult[void]] {.async.} =
```
rymnc, line 62: if resumeMessageStore is using SetupResult , shouldn't it also be wrapped within the `when defined` block?

if not, maybe we shouldn't define the SetupResult type in the `when defined ` block

rymnc (approved):
LGTM, small question

Ivansete-status (approved):

apps/wakunode2/wakunode2.nim
LNSD: And it is under the `when defined` block 🤔

PR #1661, opened 2023-04-11
refactor(wakunode2): drop experimental store resume at node boot

LNSD:
The "resume waku archive on boot" functionality has been behind a compilation flag and has not been released and tested for a long time. I suggest dropping the integration with the application until this functionality becomes part of the roadmap again.

> **Note**
> This PR is not against the `master` branch. It is against `wakunode-develop` feature branch.

- [x] Removed `resume` functionality wrapper functions from the `wakunode2` app.

### DIFF
--- a/apps/wakunode2/wakunode2.nim
+++ b/apps/wakunode2/wakunode2.nim
@@ -50,28 +50,11 @@ import
   ./config
 
 logScope:
-  topics = "wakunode"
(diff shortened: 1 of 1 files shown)


jm-clius (approved):
Yes, I think it's cleaner this way for unused/untested functionality.

PR #1663, opened 2023-04-11
fix: wrap untracked protocol handler exceptions

LNSD:
During the `wakunode2` refactoring, some issues were detected. The following changes aim to fix them:

- [x] Surround the `waku_store` protocol handler with a try/except block since it can raise an  `Exception`.
- [x] Surround the `waku_lightpush` protocol handler with a try/except block since it can raise an `Exception`.

### DIFF
--- a/waku/v2/protocol/waku_lightpush/protocol.nim
+++ b/waku/v2/protocol/waku_lightpush/protocol.nim
@@ -27,7 +27,7 @@ const WakuLightPushCodec* = "/vac/waku/lightpush/2.0.0-beta1"
 
 type
   WakuLightPushResult*[T] = Result[T, string]
-  
+
   PushMessageHandler* = proc(peer: PeerId, pubsubTopic: PubsubTopic, message: WakuMessage): Future[WakuLightPushResult[void]] {.gcsafe, closure.}
 
   WakuLightPush* = ref object of LPProtocol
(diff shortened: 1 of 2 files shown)


waku/v2/protocol/waku_store/protocol.nim
```diff
@@ -72,7 +72,16 @@ proc initProtocolHandler(ws: WakuStore) =
...
     waku_store_queries.inc()
 
-    let responseRes = ws.queryHandler(request)
+    var responseRes: HistoryResult
+    try:
+      responseRes = ws.queryHandler(request)
+    except Exception:
+      error "history query failed", peerId= $conn.peerId, requestId=requestId, error=getCurrentExceptionMsg()
+
+      let response = responseRes.toRPC()
```
jm-clius, line 81: What would be the content of response here, since `responseRes` raised an exception. Just default values?

jm-clius (commented):
Makes sense to me in general, though I have a question re the expected content of the history response when the query raised an exception.

waku/v2/protocol/waku_store/protocol.nim
```diff
@@ -72,7 +72,16 @@ proc initProtocolHandler(ws: WakuStore) =
...
+      let rpc = HistoryRPC(requestId: requestId, response: some(response))
+      await conn.writeLp(rpc.encode().buffer)
+      return
```
Ivansete-status, line 84: LGTM!
Just a couple of doubts:

* Is this part intended to be inside the 'except' or should it be outside?

* Is 'conn' properly set at this point?

waku/v2/protocol/waku_lightpush/protocol.nim
```diff
@@ -57,7 +57,15 @@ proc initProtocolHandler*(wl: WakuLightPush) =
...
 
     var response: PushResponse
-    let handleRes = await wl.pushHandler(conn.peerId, pubsubTopic, message)
+    var handleRes: WakuLightPushResult[void]
+    try:
+      handleRes = await wl.pushHandler(conn.peerId, pubsubTopic, message)
+    except Exception:
+      response = PushResponse(is_success: false, info: some($getCurrentExceptionMsg()))
+      waku_lightpush_errors.inc(labelValues = [messagePushFailure])
+      error "pushed message handling failed", error= $getCurrentExceptionMsg()
```
rymnc, line 66:
```suggestion
      error "pushed message handling failed", error= getCurrentExceptionMsg()
```

rymnc (approved):
LGTM

waku/v2/protocol/waku_store/protocol.nim
LNSD: > * Is this part intended to be inside the 'except' or should it be outside?

It is intended to be part of the except block since we respond to the node by writing to the connection.

> * Is 'conn' properly set at this point?

Yes, it should be available to write the response. The callback/handler function raises the exception here, but the connection is available.

PR #1666, opened 2023-04-11
refactor(wakunode2): added peer storage app setup proc and tests (pt. 2)

LNSD:
This is the second part of the `App` implementation and test cases.

> **Note**
> This PR is not merging into the `master` branch. It is merging into the `wakunode-develop` feature branch.

- [x] Added the setup logic for the JSON-RPC server, the REST server, and the metrics server.
- [x] Added shutdown logic for the JSON-RPC and REST servers.
- [x] Added a smoke test that asserts the correct initialization of the Wakunode2 App instance with the default configuration.
- [x] Enabled the `testwakunode2` tests (a.k.a. Wakunode2 App tests) in the CI pipeline.

### DIFF
--- a/apps/wakunode2/app.nim
+++ b/apps/wakunode2/app.nim
@@ -14,7 +14,9 @@ import
   libp2p/protocols/pubsub/gossipsub,
   libp2p/peerid,
   eth/keys,
-  eth/net/nat
+  eth/net/nat,
+  json_rpc/rpcserver,
+  presto
 import
   ../../waku/common/sqlite,
   ../../waku/v2/waku_node,
@@ -64,12 +66,14 @@ type
 
     node: WakuNode
 
+    rpcServer: Option[RpcHttpServer]
+    restServer: Option[RestServerRef]
+
   AppResult*[T] = Result[T, string]
 
 
-# TODO: Uncomment after the refactoring work is done
-# func node*(app: App): WakuNode =
-#   app.node
+func node*(app: App): WakuNode =
+  app.node
 
 func version*(app: App): string =
   app.version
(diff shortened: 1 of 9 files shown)


apps/wakunode2/wakunode2.nim
LNSD, line 1: I had to rename the `node` variable to `nodeInstance` to solve the conflict of the `App`'s `node` getter.

tests/wakunode2/test_app.nim
```diff
@@ -53,3 +56,29 @@ suite "Wakunode2 - App initialization":
 
     ## Then
     check res.isOk()
+
+  test "node setup is successful":
```
LNSD, line 60: I cannot use `asyncTest` here due to some non-GC-safe issue. I am using `waitFor` as a temporary workaround.

apps/wakunode2/wakunode2.nim
LNSD, line 1: I had to rename the `node` variable to `nodeInstance` to solve the conflict of the `App`'s `node` getter.

tests/wakunode2/test_app.nim
```diff
@@ -53,3 +55,28 @@ suite "Wakunode2 - App initialization":
 
     ## Then
     check res.isOk()
+
+  test "node setup is successful with default configuration":
```
LNSD, line 59: I cannot use `asyncTest` here due to some non-GC-safe issue that makes the `wakunode2.setupWakuArchive()` incompatible with it. 

I am using `waitFor` as a temporary workaround.

apps/wakunode2/app.nim
```diff
@@ -222,6 +226,41 @@ proc setupWakuArchiveDriver(dbUrl: string, vacuum: bool, migrate: bool): AppResu
...
+                                                migrate = app.conf.storeMessageDbMigration)
+  if archiveDriverRes.isOk():
+    app.archiveDriver = some(archiveDriverRes.get())
+  else:
+    return err("failed to configure archive driver" & archiveDriverRes.error)
+
+  # Message store retention policy
+  let storeMessageRetentionPolicyRes = validateStoreMessageRetentionPolicy(app.conf.storeMessageRetentionPolicy)
+  if storeMessageRetentionPolicyRes.isErr():
+    return err("failed to configure the message retention policy" & storeMessageRetentionPolicyRes.error)
```
Ivansete-status, line 251: The PR LGTM!
Shall we add a separator (e.g. colon or `" [" & error & "]"`) ?
There few other cases in this file that also need to separate errors.

Ivansete-status (approved):

apps/wakunode2/app.nim
LNSD: Added. Note that these errors are here for backward compatibility and temporarily until the rest of the `App` is reworked and refactored.

apps/wakunode2/app.nim
```diff
@@ -252,6 +291,17 @@ proc retrieveDynamicBootstrapNodes*(dnsDiscovery: bool, dnsDiscoveryUrl: string,
...
   ok(newSeq[RemotePeerInfo]()) # Return an empty seq by default
 
+proc setupDyamicBootstrapNodes*(app: var App): AppResult[void] =
+  let dynamicBootstrapNodesRes = retrieveDynamicBootstrapNodes(app.conf.dnsDiscovery,
+                                                               app.conf.dnsDiscoveryUrl,
+                                                               app.conf.dnsDiscoveryNameServers)
+  if dynamicBootstrapNodesRes.isOk():
+    app.dynamicBootstrapNodes = dynamicBootstrapNodesRes.get()
+  else:
+    warn "2/7 Retrieving dynamic bootstrap nodes failed. Continuing without dynamic bootstrap nodes.", error=dynamicBootstrapNodesRes.error
```
vpavlin, line 301: What is this 2/7?

vpavlin:
Just a thought - any reason why we use `withXYZ` in node/builder and `setupXYZ` in the app? Would it make sense to keep this consistent?

apps/wakunode2/app.nim
LNSD: Step 2 of 7. See:

https://github.com/waku-org/nwaku/blob/73cbafa658295213298b06ef7b349683f0d1f858/apps/wakunode2/wakunode2.nim#L719-L721

vpavlin: Ah, I see:D

LNSD:
> Just a thought - any reason why we use `withXYZ` in node/builder and `setupXYZ` in the app? Would it make sense to keep this consistent?

Both things are different:

* The `WakuNodeBuilder` creates an instance of an embeddable `WakuNode`. This instance should not contain any logic related to the `wakunode2` infra node application.
* The `App`, when instantiated, on the other hand, would create a reference to the wakunode2 infra node application.

-------

Note that `WakuNode` and `wakunode2` are not the same.

The `wakunode2` app may include only relevant functionality if executed as a fleet infra node.
 
Opposed to that, the `waku/v2/waku_node#WakuNode` should be an embeddable node that contains the minimal functionality a node should perform; the idea is to provide the extra behavior to this minimal node via the builder methods (e.g., via Waku Relay handlers).

PR #1668, opened 2023-04-11
feat: Add tools maket target and build tools in CI

vpavlin:


alrevuelta (approved):
thanks for this PR! lgtm.

Very needed, specially since we broke wakucanary in the past due to not building it in CI.

Issue #1665, opened 2023-04-11
bug: unhandled overflow in backoff


vpavlin:
### Problem
A node randomly crashes with unhandled exception - OverflowDefect

### Impact
Not sure, potentially disruptive for production/reliable runtime, but does not seem critical

### To reproduce

1. Clone https://github.com/alrevuelta/nwaku-compose
2. Update the image version to v0.16.0
4. Run `docker compose up`
5. Observe the following error after some time

### Expected behavior
The node does not crash with unhandled exception

### Screenshots/logs
```
INF 2023-04-08 11:50:42.807+00:00 Peer reachability status                   topics="waku node" tid=1 file=waku_node.nim:257 networkReachability=Reachable confidence=1.0
INF 2023-04-08 11:50:44.196+00:00 Finished dialing multiple peers            topics="waku node peer_manager" tid=1 file=peer_manager.nim:382 successfulConns=1 attempted=4
INF 2023-04-08 11:50:44.862+00:00 Discovered peers                           topics="waku node" tid=1 file=waku_node.nim:1016 discovered=36 new=1
INF 2023-04-08 11:50:47.139+00:00 Total connections initiated                topics="waku node metrics" tid=1 file=waku_metrics.nim:67 count=6.0
INF 2023-04-08 11:50:47.139+00:00 Total messages                             topics="waku node metrics" tid=1 file=waku_metrics.nim:68 count=29737.0
INF 2023-04-08 11:50:47.139+00:00 Total store peers                          topics="waku node metrics" tid=1 file=waku_metrics.nim:69 count=0.0
INF 2023-04-08 11:50:47.139+00:00 Total peer exchange peers                  topics="waku node metrics" tid=1 file=waku_metrics.nim:70 count=0.0
INF 2023-04-08 11:50:47.139+00:00 Total lightpush peers                      topics="waku node metrics" tid=1 file=waku_metrics.nim:71 count=0.0
INF 2023-04-08 11:50:47.139+00:00 Total filter peers                         topics="waku node metrics" tid=1 file=waku_metrics.nim:72 count=0.0
INF 2023-04-08 11:50:47.139+00:00 Total active filter subscriptions          topics="waku node metrics" tid=1 file=waku_metrics.nim:73 count=46.0
INF 2023-04-08 11:50:47.139+00:00 Total errors                               topics="waku node metrics" tid=1 file=waku_metrics.nim:74 count=0.0
[...]

PR #1670, opened 2023-04-12
bug: move canBeConnected to PeerManager and check for potential overflow

vpavlin:
This PR moves `canBeConnected` to PeerManager. It also adds a check for input parameters for backoff calculation to make sure we do not hit overflow once the node is up and running - it checks that the max backoff is lower than 1 week and raises an exception otherwise.

It also adds tests to verify the potential overflow is caught at PeerManager initialization.

closes: #1665


vpavlin:
@alrevuelta I could also see another approach where we add `maxFailedAttempts` to peerStore object, so that we do not need to pass it in each `canBeConnected` call. Let me know if I should change the PR

alrevuelta (changes requested):
thanks for the PR!

Indeed, perhaps its better to add `maxFailedAttempts` to the peerstore, and avoid passing it in `canBeConnected`. Actually it may be also a good moment to do the same with `backoffFactor` and `initialBackoffInSec`. Would be very clean to have just `.canBeConnected(peerId)`.

The thing is that `PeerStore` doest really have a new or an init proc. This is becaue that type is take from nim-libp2p.

So I would say we have two options:
* a: add a `new` proc for the peerstore and pass all these variables. We would need also our own `PeerStore` extending the nim-libp2p with fields such as maxFailedAttempts. then `canBeConnected` can use these values without having to pass them. but not sure i like having this values both in peer manager and peerstore.
* b: merge peerstore and peermanager so that `canBeConnected` is part of the peermanager, which already has all the fields we need (backofffactors, maxFailedAttempts, etc). or well a simpler approach is to move `canBeConnected` to the peermanager, since now that i think about it, its more related to the peer manager that the peer store.

tldr: what do you thinnk about moving `canBeConnected` to the peermanager, so that all the parameters that are needed dont need to be passed (just peerId)?

waku/v2/node/peer_manager/waku_peer_store.nim
```diff
@@ -55,6 +56,10 @@ proc canBeConnected*(peerStore: PeerStore,
   if failedAttempts == 0:
     return true
 
+  # if there are too many failed attemps, do not reconnect
+  if failedAttempts >= maxFailedAttempts:
+    return false
```
rymnc, line 61: we should probably add a check earlier to ensure that the maxFailedAttempts is < 14

waku/v2/node/peer_manager/peer_manager.nim
```diff
@@ -235,6 +235,31 @@ proc loadFromStorage(pm: PeerManager) =
...
+  if failedAttempts == 0:
+    return true
+
+  # if there are too many failed attemps, do not reconnect 
+  # the check for 14 comes from overflow described in https://github.com/waku-org/nwaku/issues/1665)
+  if failedAttempts >= pm.maxFailedAttempts or failedAttempts > 14:
+    return false
+
+  # If it errored we wait an exponential backoff from last connection
+  # the more failed attemps, the greater the backoff since last attempt
```
rymnc, line 255:
```suggestion
  # the more failed attempts, the greater the backoff since last attempt
```

waku/v2/node/peer_manager/peer_manager.nim
```diff
@@ -235,6 +235,31 @@ proc loadFromStorage(pm: PeerManager) =
...
+  let now = Moment.init(getTime().toUnix, Second)
+  let lastFailed = pm.peerStore[LastFailedConnBook][peerId]
+  let backoff = chronos.seconds(pm.initialBackoffInSec*(pm.backoffFactor^(failedAttempts-1)))
```
rymnc, line 258: Hmm.. can we abstract this into a function and fuzz it in the test suite? Will help catch these errors in the futures too

tests/v2/test_peer_manager.nim
```diff
@@ -195,20 +195,14 @@ procSuite "Peer Manager":
...
       # Right after failing there is a backoff period
-      nodes[0].peerManager.peerStore.canBeConnected(
-        nonExistentPeer.peerId,
-        nodes[0].peerManager.initialBackoffInSec,
-        nodes[0].peerManager.backoffFactor) == false
+      nodes[0].peerManager.canBeConnected(nonExistentPeer.peerId) == false
 
     # We wait the first backoff period
-    await sleepAsync(2100.milliseconds)
+    await sleepAsync(2100)
```
rymnc, line 201: I think the better practice is to use the unit explicitly, i.e .seconds/.millis etc

tests/v2/test_peer_manager.nim
```diff
@@ -655,3 +649,52 @@ procSuite "Peer Manager":
...
+
+  asyncTest "canBeConnected() returns correct value":
+    let pm = PeerManager.new(
+      switch = SwitchBuilder.new().withRng(rng).withMplex().withNoise()
+      .withPeerStore(10)
+      .withMaxConnections(5)
+      .build(),
+      initialBackoffInSec = 1, # with InitialBackoffInSec = 1 backoffs are: 1, 2, 4, 8secs.
+      backoffFactor = 2,
+      maxFailedAttempts = 10,
```
rymnc, line 661: This would test only the happy path, can we randomize these values?
I think tests should be deterministic, but wherever math is we should be fuzzing too

tests/v2/test_peer_manager.nim
vpavlin: That makes sense, but when I try to use `milliseconds`, I get this error:

```
/home/vpavlin/devel/github.com/waku-org/nwaku/tests/v2/test_peer_manager.nim(201, 21) Error: type mismatch: got <TimeInterval>
but expected one of:
proc sleepAsync(duration: Duration): Future[void]
  first type mismatch at position: 1
  required type for duration: Duration
  but expression 'milliseconds(2100)' is of type: TimeInterval
proc sleepAsync(ms: int): Future[void]
  first type mismatch at position: 1
  required type for ms: int
  but expression 'milliseconds(2100)' is of type: TimeInterval

expression: sleepAsync(milliseconds(2100))
```

which is why I removed it

tests/v2/test_peer_manager.nim
vpavlin: That's a good point, I am testing for failure by modifying the `maxFailedAttempts` on the line 698, but I guess some fuzzing would make sense

waku/v2/node/peer_manager/peer_manager.nim
```diff
@@ -235,6 +235,31 @@ proc loadFromStorage(pm: PeerManager) =
...
+  # wait more before trying again.
+  let failedAttempts = pm.peerStore[NumberFailedConnBook][peerId]
+
+  # if it never errored, we can try to connect
+  if failedAttempts == 0:
+    return true
+
+  # if there are too many failed attemps, do not reconnect 
+  # the check for 14 comes from overflow described in https://github.com/waku-org/nwaku/issues/1665)
+  if failedAttempts >= pm.maxFailedAttempts or failedAttempts > 14:
```
alrevuelta, line 251: perhaps i would enforce the chech that `maxFailedAttempts` < some_value when we init the peer manager.

note also that with the current values 14 is ok, but if i set `backoffFactor` to 10000 (nonsense perhaps) i will hit the overflow way earlier.

waku/v2/node/peer_manager/peer_manager.nim
alrevuelta: not against fuzzing but perhaps overkill? related to my other comment, perhaps we can just assert that the backoff values makes sense when we init the peermanager with initialBackoffInSec and backoffFactor. just run the math in there to ensure values are withing some reasonable limits.

alrevuelta (changes requested):
looking really good, moving it to the peermanager was a great idea. some comments.

waku/v2/node/peer_manager/peer_manager.nim
rymnc: that works too, just something we should consider next time :)

tests/v2/test_peer_manager.nim
rymnc: This works too, or as @alrevuelta mentioned above, this is a non blocker :)

vpavlin:
@rymnc @alrevuelta I added a check to the `new` proc  - it calculates maximum backoff and checks if it is over 1 week (rather arbitrary value, so we can agree on a different one:) )

This results in either success, exception due to being over 1 week or exception due to overflow. In any case this should prevent the overflow happening later when the node is running.

alrevuelta (changes requested):
looking good, some minor comments and almost ready to go.

can you run a node for some time to ensure it works as expected? if you connect to the existing fleets you should see multiple failed attempts, as many nodes in the network are not reachable (eg behind nat)

waku/v2/node/peer_manager/peer_manager.nim
```diff
@@ -235,6 +242,31 @@ proc loadFromStorage(pm: PeerManager) =
...
+  # It uses an exponential backoff. Each connection attempt makes us
+  # wait more before trying again.
+  let failedAttempts = pm.peerStore[NumberFailedConnBook][peerId]
+
+  # if it never errored, we can try to connect
+  if failedAttempts == 0:
+    return true
+
+  # if there are too many failed attempts, do not reconnect
+  # the check for 14 comes from overflow described in https://github.com/waku-org/nwaku/issues/1665)
```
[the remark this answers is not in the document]
vpavlin: Yeah, I had a check for 14 there initially, that is gone now

tests/v2/test_peer_manager.nim
vpavlin: Ok, I get the error now, glad I looked into Chronos a bit more today:D Fixed

waku/v2/node/peer_manager/peer_manager.nim
```diff
@@ -286,12 +318,20 @@ proc new*(T: type PeerManager,
          maxConnections = maxConnections
     raise newException(Defect, "Max number of connections can't be greater than PeerManager capacity")
 
+  # attempt to calculate max backoff to prevent potential overflows or unreasonably high values
```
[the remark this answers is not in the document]
alrevuelta: just saw the test thanks!

alrevuelta (approved):
lgtm. thanks for fixing it and addressing the comments!

mind editing the description "as long as maxFailedAttempts is lower than 15 in the default setting"?

vpavlin:
Rewrote the description

PR #1671, opened 2023-04-12
feat: curate peers shared over px protocol

alrevuelta:
closes https://github.com/waku-org/nwaku/issues/1521

- [x] Remove the peer exchange discv5 loop and reuse the peers discovered via discv5 by the peer manager. [ref](https://github.com/waku-org/nwaku/issues/1463#issuecomment-1350649456)
- [x] Curate peers shared over peer exchange, sharing only reachable peers https://github.com/waku-org/nwaku/issues/1521 (peers that are connected (or were connected in the past) and come from discv5). This helps the requester to not waste time with unreachable peers.
- [x] Fix issue were duplicated peers were returned by peer exchange protocol within the same rpc call. (eg if 2 peers were requested, both could be the same)
- [x] Update px cache every x time, to avoid attacks such as graph learning. With this, a peer requesting peers via px will get at max `MaxPeersCacheSize` peers (see [this](https://rfc.vac.dev/spec/34/#exchange-peer-cache-size)).

Beyond unit tests, can be tested with:

```
./build/wakunode2 --dns-discovery=true --dns-discovery-url=enrtree://AOGECG2SPND25EEFMAJ5WF3KSGJNSGV356DSTL2YVLLZWIV6SAYBM@prod.nodes.status.im --discv5-discovery=true --discv5-enr-auto-update=True --nodekey=161c2836197e8918d7eb1039758423795f0dc697a018ed55d5931aa9b0e5efb3 --peer-exchange
```

```
./build/wakunode2 \
--tcp-port=30305 \
--ports-shift=1 \
--peer-exchange-node=/ip4/127.0.0.1/tcp/60000/p2p/16Uiu2HAm2ZuRHGZUzG6JGJeY1nGKHXLGrQYnZ7mZToymHtFfYFFf
```

### DIFF
--- a/tests/v2/test_waku_peer_exchange.nim
+++ b/tests/v2/test_waku_peer_exchange.nim
@@ -126,17 +126,15 @@ procSuite "Waku Peer Exchange":
     await allFutures([node1.start(), node2.start(), node3.start()])
     await allFutures([node1.startDiscv5(), node2.startDiscv5()])
 
-    # node2 can be connected, so will be returned peer exchange
-    require (await node1.peerManager.connectRelay(node2.switch.peerInfo.toRemotePeerInfo()))
-
     # Give disv5 some time to discover each other
     await sleepAsync(5000.millis)
+
+    # node2 can be connected, so will be returned by peer exchange
+    require (await node1.peerManager.connectRelay(node2.switch.peerInfo.toRemotePeerInfo()))
 
     # Mount peer exchange
     await node1.mountPeerExchange()
     await node3.mountPeerExchange()
-
-
 
     let connOpt = await node3.peerManager.dialPeer(node1.switch.peerInfo.toRemotePeerInfo(), WakuPeerExchangeCodec)
     check:


waku/v2/node/peer_manager/peer_manager.nim
```diff
@@ -113,6 +113,10 @@ proc addPeer*(pm: PeerManager, remotePeerInfo: RemotePeerInfo) =
 
   pm.peerStore[AddressBook][remotePeerInfo.peerId] = remotePeerInfo.addrs
   pm.peerStore[KeyBook][remotePeerInfo.peerId] = publicKey
+  pm.peerStore[SourceBook][remotePeerInfo.peerId] = origin
+
+  if remotePeerInfo.enr.isSome:
```
Ivansete-status, line 118: Shall we use () as it is a verb?
```suggestion
  if remotePeerInfo.enr.isSome():
```

Ivansete-status (approved):
LGTM! :1st_place_medal:

waku/v2/node/waku_node.nim
```diff
@@ -780,13 +776,13 @@ proc fetchPeerExchangePeers*(node: Wakunode, amount: uint64) {.async, raises: [D
   let pxPeersRes = await node.wakuPeerExchange.request(amount)
   if pxPeersRes.isOk:
     var validPeers = 0
-    for pi in pxPeersRes.get().peerInfos:
+    let peers = pxPeersRes.get().peerInfos
```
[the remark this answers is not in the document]
alrevuelta: tbh it was for readability. out of curiosity why could it be more efficient?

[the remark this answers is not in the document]
Ivansete-status: ah ok! If guessed that you probably did that for efficiency for any Nim's issue beyond my current knowledge. It should be equally efficient as it just reads a value

waku/v2/protocol/waku_peer_exchange/protocol.nim
```diff
@@ -96,41 +96,39 @@ proc respond(wpx: WakuPeerExchange, enrs: seq[enr.Record], conn: Connection): Fu
...
+
+proc updatePxEnrCache(wpx: WakuPeerExchange) {.async.} =
+  # try more aggressively to fill the cache at startup
+  while wpx.enrCache.len < MaxPeersCacheSize:
+    populateEnrCache(wpx)
+    await sleepAsync(5.seconds)
+
+  heartbeat "Updating px enr cache", CacheRefreshInterval:
+    wpx.cleanCache()
+    wpx.populateEnrCache()
```
jm-clius, line 131: Wouldn't this cause a race condition between when someone is querying the cache and the heartbeat is cleaning the cache in order to populate it again? I would suggest that the cache being populated and the one queried be different with an atomic swap between the two after the former is populated.

jm-clius (commented):
Approach looks good, though I have a question about race conditions.

tests/v2/test_waku_peer_exchange.nim
```diff
@@ -126,14 +126,17 @@ procSuite "Waku Peer Exchange":
     await allFutures([node1.start(), node2.start(), node3.start()])
     await allFutures([node1.startDiscv5(), node2.startDiscv5()])
 
+    # node2 can be connected, so will be returned peer exchange
+    require (await node1.peerManager.connectRelay(node2.switch.peerInfo.toRemotePeerInfo()))
+
+    # Give disv5 some time to discover each other
+    await sleepAsync(5000.millis)
```
vpavlin, line 133: Is there a way to limit these "arbitrary" sleeps in our tests and rather check for the state (e.g. in a loop with some smaller sleep duration) or would that be an overkill? IMO that could potentially speed up some test runs and at the same time make them more resilient in case the node is slower.

alrevuelta: indeed, totally agree that in some tests we should follow this pattern, waitFor instead of arbitrary sleep. we would still need a timeout but ofc tests would finish before, and we can cut some time.

perhaps not for this specific test, but will take it into account for further tests/refactors thanks!

jm-clius (approved):
Thanks! LGTM now.

PR #1672, opened 2023-04-12
fix(rln-relay): scope of getEvents

rymnc:
Fixes a bug where removed events were included in the valid event block table, thanks @richard-ramos

Also, fixes https://github.com/waku-org/nwaku/pull/1623#pullrequestreview-1362558825

### DIFF
--- a/waku/v2/protocol/waku_rln_relay/group_manager/on_chain/group_manager.nim
+++ b/waku/v2/protocol/waku_rln_relay/group_manager/on_chain/group_manager.nim
@@ -60,7 +60,7 @@ type
 const DefaultKeyStorePath* = "rlnKeystore.json"
 const DefaultKeyStorePassword* = "password"
 
-template initializedGuard*(g: OnchainGroupManager): untyped =
+template initializedGuard(g: OnchainGroupManager): untyped =
   if not g.initialized:
     raise newException(ValueError, "OnchainGroupManager is not initialized")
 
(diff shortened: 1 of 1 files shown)


waku/v2/protocol/waku_rln_relay/group_manager/on_chain/group_manager.nim
```diff
@@ -196,7 +196,16 @@ proc backfillRootQueue*(g: OnchainGroupManager, blockTable: BlockTable): Future[
         # add the backfilled root
         g.validRoots.addLast(g.validRootBuffer.popLast())
 
-proc getEvents*(g: OnchainGroupManager, fromBlock: BlockNumber, toBlock: Option[BlockNumber] = none(BlockNumber)): Future[BlockTable] {.async.} =
+proc insert(blockTable: var BlockTable, blockNumber: BlockNumber, member: Membership) =
+  if blockTable.hasKeyOrPut(blockNumber, @[member]):
+    try:
+      blockTable[blockNumber].add(member)
+    except KeyError: # qed
+      error "could not insert member into block table", blockNumber=blockNumber, member=member
```
jm-clius, line 204: Would an `error` log here be enough error handling (vs. e.g. adding metrics returning an `err` `Result`, etc.)?

jm-clius (approved):
LGTM!

waku/v2/protocol/waku_rln_relay/group_manager/on_chain/group_manager.nim
rymnc: the `hasKeyOrPut` takes care of the error. nim throws an error without having context of the previous line (thats why an error log is OK imo in this case)

PR #1673, opened 2023-04-12
refactor:optimize getting number of connections and streams

vpavlin:
I saw PR #1609  after merge and noticed it could be potentially optimized to eliminate multiple loops over all connections and streams.

I've updated the procedures to not accept directions and rather return a tupplo of `(connectionsIn, connectionsOut)` and `(streamsIn, streamsOut)`

It makes code a bit less explicit, but we only need to loop over all connections and streems twice rather than 4 times (in and out direction for connections, in and out for streams)

@alrevuelta please review carefully, in case I misunderstood how things work in your code:)

### DIFF
--- a/waku/v2/node/peer_manager/peer_manager.nim
+++ b/waku/v2/node/peer_manager/peer_manager.nim
@@ -477,31 +477,41 @@ proc connectToNodes*(pm: PeerManager,
   # later.
   await sleepAsync(chronos.seconds(5))
 
-# Returns the amount of physical connections for a given direction
+# Returns the amount of physical connections (in and out)
 # containing at least one stream with the given protocol.
-proc getNumConnections*(pm: PeerManager, dir: Direction, protocol: string): int =
-  var numConns = 0
+proc getNumConnections*(pm: PeerManager, protocol: string): (int, int) =
+  var
+    numConnsIn = 0
+    numConnsOut = 0
   for peerId, muxers in pm.switch.connManager.getConnections():
     for peerConn in muxers:
       let streams = peerConn.getStreams()
-      if peerConn.connection.transportDir == dir:
-        if streams.anyIt(it.protocol == protocol):
-          numConns += 1
-  return numConns
+      if streams.anyIt(it.protocol == protocol):
+        if peerConn.connection.transportDir == Direction.In:
+          numConnsIn += 1
+        elif peerConn.connection.transportDir == Direction.Out:
+          numConnsOut += 1
 
-proc getNumStreams*(pm: PeerManager, dir: Direction, protocol: string): int =
-  var numConns = 0
+  return (numConnsIn, numConnsOut)
+
+proc getNumStreams*(pm: PeerManager, protocol: string): (int, int) =
+  var
+    numStreamsIn = 0
+    numStreamsOut = 0
   for peerId, muxers in pm.switch.connManager.getConnections():
     for peerConn in muxers:
         for stream in peerConn.getStreams():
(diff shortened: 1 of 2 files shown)


alrevuelta (changes requested):
thanks! really like the idea! left a comment, let me know what you think :)

waku/v2/node/peer_manager/peer_manager.nim
```diff
@@ -455,10 +455,41 @@ proc getNumStreams*(pm: PeerManager, dir: Direction, protocol: string): int =
             numConns += 1
   return numConns
 
+proc getNumConnsAndStreams*(pm: PeerManager, protocol: string): (int, int, int, int) =
```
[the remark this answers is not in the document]
vpavlin: Well, I don't have a strong preference, my goal here was to come up with the most optimal code possible (which I am pretty sure my code is not anyway:D). So if you prefer to have it separated, I can update the PR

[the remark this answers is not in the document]
alrevuelta: since imho this optimization is not very critical (runs every few seconds on a few dozen of peers), i would say its better to sacrifice some iterations in favor of better readability and have a more generic function. eg someone may want to know the connections but not the streams. Having everything in one place seems to specific for this use case. That's why i think it's better to keep it separate.

So I would have 2 functions but return both directions as in your idea.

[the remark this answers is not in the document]
vpavlin: See the latest push, I also updated the PR description and removed WIP flag

.github/workflows/release-assets.yml
```diff
@@ -31,12 +31,18 @@ jobs:
       - name: Submodules
         run: |
           git submodule update --init --recursive
+
+      - name: Get submodules versions
```
alrevuelta, line 35: perhaps a leftover?

waku/v2/node/peer_manager/peer_manager.nim
```diff
@@ -477,31 +477,41 @@ proc connectToNodes*(pm: PeerManager,
...
       let streams = peerConn.getStreams()
-      if peerConn.connection.transportDir == dir:
-        if streams.anyIt(it.protocol == protocol):
-          numConns += 1
-  return numConns
-
-proc getNumStreams*(pm: PeerManager, dir: Direction, protocol: string): int =
-  var numConns = 0
+      if streams.anyIt(it.protocol == protocol):
+        if peerConn.connection.transportDir == Direction.In:
```
alrevuelta, line 490: perhaps nitpick but can be have "if In" and "else if Out"? Just in case there is an uninitialized value or something weird coming from libp2p.

if peerConn.connection.transportDir == Direction.In:
  xxx
else if peerConn.connection.transportDir == Direction.Out:
  xxx

alrevuelta (changes requested):
@vpavlin great thanks! looking very good. a minor nitpick and we are good to go :)

.github/workflows/release-assets.yml
vpavlin: huh..yeah, this bubbled in by mistake

waku/v2/node/peer_manager/peer_manager.nim
vpavlin: Yeah, absolutely, I was originally gonna do that, but then was not sure if it can be uninitialized (cause it did not seem like being a refernce), but maybe I misunderstood how that works in Nim:)

alrevuelta (approved):
lgtm thanks!

PR #1675, opened 2023-04-14
fix(logs): fix log reporting wrong ok connected peers

alrevuelta:
* Fix log reporting wrong ok connected peers
* `countIt(true)` != `countIt(it == true)` 🤦


jm-clius (approved):
Ah, thanks! I noticed an issue with this log, but haven't had time to take a closer look.

PR #1676, opened 2023-04-17
feat(enr): added support for relay shards field

LNSD:
This PR introduces the Static Sharding ENR extension.

- [x] Split the `waku_enr` module into three submodules: capabilities, multiaddr, and sharding.
- [x] Added the `rs` and `rsv` extension implementation.
- [x] Added some tests covering the sharding extension.


alrevuelta (commented):
left some comments.

@kaiserd Remember we talked last week about sharding. Do we consider the spec ready and moving to implementation?

I raised my concerns [here](https://github.com/vacp2p/research/issues/174#issuecomment-1410749909) and [here](https://github.com/vacp2p/rfc/pull/566) but as a tldr:
* I think waku should handle discovery only on waku shards.
* I dont think we need this hierarchy of cluster + index. just shard index.
* I think we should cover way less shards. quality and not quantity.
* I dont think we need static sharding, just automatic. I wouldn't justify the need of static wharding with the MVP.
* imho static sharding would be waku's dystopia. companies pushing their users to shards they control with no guarantees on decentralization, privacy, etc. all shards should be equal = managed by the protocol = auto sharding.
* if an app wants static shards, fine, fork the network (like bsc, gnosis, ethclassic or even eth networks). wouldn't be difficult with a network-id in the ENR (which acts as kind of the shard-cluster but more generic)

These discussions, belong to here, so lets reply in there.
https://github.com/vacp2p/research/issues/174

But raising it again here since we started with the implementation.

tests/v2/test_waku_enr.nim
```diff
@@ -251,3 +252,172 @@ suite "Waku ENR - Multiaddresses":
...
+      enrSeqNum = 1u64
+      enrPrivKey = generatesecp256k1key()
+
+    let
+      shardCluster: uint16 = 22
+      shardIndices: seq[uint16] = @[1u16, 1u16, 2u16, 3u16, 5u16, 8u16]
+
+    let shards = RelayShards.init(shardCluster, shardIndices)
+
+    ## When
```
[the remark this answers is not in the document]
LNSD: > perhaps a nitpick but `withSharding` instead `withWakuRelaySharding`?
> so `.withSharding().withCapabilities().withMultiAddress()` etc.

If you add this chained builder pattern, you lose the possibility of one of those methods to return a `Result`, which is very convenient.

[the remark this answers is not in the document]
LNSD: > so we avoid having to init `RelayShards` first? and then passing it?

You can always extend the implementation once this part is used. I added the methods I will need as part of the static sharding discv5 implementation. I am reluctant to add more convenience methods just because they "might be handy".

waku/v2/protocol/waku_enr/sharding.nim
```diff
@@ -0,0 +1,219 @@
...
+    raise newException(Defect, "invalid index")
+
+  let indicesSeq = deduplicate(@indices)
+  if indices.len < 1:
+    raise newException(Defect, "invalid index count")
+
+  RelayShards(cluster: cluster, indices: indicesSeq)
+
+func init*(T: type RelayShards, cluster: uint16, indices: seq[uint16]): T =
+  if indices.anyIt(it > 1023):
```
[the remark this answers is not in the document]
LNSD: I can add it as a constant, but reusing it outside of this module makes no sense.

waku/v2/protocol/waku_enr/sharding.nim
```diff
@@ -0,0 +1,219 @@
...
+  for index in rs.indices:
+    vec[index div 8].setBit(index mod 8)
+
+  res.add(vec)
+
+  res
+
+func fromBitVector(buf: seq[byte]): Result[RelayShards, string] =
+  if buf.len != 130:
+    return err("invalid data")
```
Ivansete-status, line 135: In all the error cases I'd give as much info as possible.
```suggestion
    return err("fromBitVecto - invalid data [" & string(buf.len) & "]" )    
```

waku/v2/protocol/waku_enr/capabilities.nim
```diff
@@ -0,0 +1,103 @@
...
+  let field = record.tryGet(CapabilitiesEnrField, seq[uint8])
+  if field.isNone():
+    return none(CapabilitiesBitfield)
+
+  some(CapabilitiesBitfield(field.get()[0]))
+
+proc supportsCapability*(r: Record, cap: Capabilities): bool =
+  let recordRes = r.toTyped()
+  if recordRes.isErr():
+    return false
```
Ivansete-status, line 84: Shall we use Result[...] so that we can handle errors better?

waku/v2/protocol/waku_enr/sharding.nim
```diff
@@ -0,0 +1,222 @@
...
+# ENR record accessors (e.g., Record, TypedRecord, etc.)
+
+func relayShardingIndicesList*(record: TypedRecord): Option[RelayShards] =
+  let field = record.tryGet(ShardingIndicesListEnrField, seq[byte])
+  if field.isNone():
+    return none(RelayShards)
+
+  let indexList = fromIndicesList(field.get())
+  if indexList.isErr():
+    return none(RelayShards)
```
Ivansete-status, line 172: In this case we are losing the possible error given by `fromIndicesList`.
I'd try to chain the possible errors back to the caller as much as possible by returning a `Result[Option[..], string]`

waku/v2/protocol/waku_enr/sharding.nim
```diff
@@ -0,0 +1,222 @@
...
+    return false
+
+  rs.contains(parseRes.value)
+
+
+# ENR builder extension
+
+func toIndicesList(rs: RelayShards): seq[byte] =
+  if rs.indices.len > 255:
+    raise newException(Defect, "indices list too long")
```
Ivansete-status, line 89: In this case I'd rather return a `Result[seq[byte], string]` and give as much detail as possible in case of error.

Ivansete-status (approved):
The PR looks great!
However, I think we could have some more details in error handling that maybe worth adding.

waku/v2/protocol/waku_enr/capabilities.nim
LNSD: It is not necessary. But it is a trade-off. 

Following the function logic: 

> If we cannot parse the ENR record, we can conclude that it won't contain any capabilities field.

If we return a result, it will force us to move the error logic one level up. We'll be able to handle the error here and take care of it. This is ok for all the use cases of this function. Note that performing error handling within a `sequtils.map()` (or any other `sequtils` procedure) can be cumbersome due to Nim's "orthopedic lambda function" syntax.

waku/v2/protocol/waku_enr/sharding.nim
LNSD: Agree, changing it 👍🏼

jm-clius (approved):
LGTM! Thanks for making progress on this.

PR #1680, opened 2023-04-18
chore: flatten waku v2 protocols folder

LNSD:
The `protocols` folder has been growing for a long time containing modules that are not protocols (e.g., waku_archive, waku_keystore, etc.). 

- [x] Flattened the `waku/v2` structure by moving the `protocols` content to its parent root: `waku/v2`.


Ivansete-status:
Thanks for the PR @LNSD !
If we apply this change we will have generic concepts like 'utils' or 'node' at the same level as 'waku_rekay', for instance, which is a quite specific concept.
IMO I liked the previous structure but I may be missing sth :)

jm-clius (approved):
LGTM. I agree that `protocol` was a bit of a catch-all with no clear meaning.

rymnc (approved):
LGTM, thanks :)

LNSD:
> Thanks for the PR @LNSD ! If we apply this change we will have generic concepts like 'utils' or 'node' at the same level as 'waku_rekay', for instance, which is a quite specific concept. IMO I liked the previous structure but I may be missing sth :)

It is a fair point. We have been doing an effort to relocate things within the `utils` folder within other modules. The idea is to remove `utils`. On the other hand, `node` is a specific concept, such as archive or discv5. It is the "embeddable node".

Another topic is the things within the `node` folder that are unrelated to it. For example, the peer manager can be moved into its own module; or the JSON-RPC and REST APIs, which I think are application-specific and should be moved to the `wakunode2` app module. But this second part is a different discussion.

PR #1681, opened 2023-04-18
feat(mem-analysis): Adding Dockerfile_with_heaptrack

Ivansete-status:
## Issue
https://github.com/waku-org/nwaku/issues/1662

## Description
Adding a Dockerfile that allows to generate a Docker image with both waku and heaptrack. The generated docker containers will run a `wakunode2` instance underneath `heaptrack`.

Once the container is running, it will generate periodic mem-allocation stats in the container's `/` folder with the name `heaptrack.wakunode.1.gz`. If the container is restarted, then the previous allocation file is lost so it is important to copy out the file before the container restarts or crashes.

[Heaptrack](https://github.com/KDE/heaptrack) is the tool used and it has two [operational modes](https://github.com/KDE/heaptrack#using-heaptrack). In this case, the 'attach' mode makes the underlying program to crash (SYGSEGV). I tried this with a simple C program and it happens the same so it seems to be a Docker-related issue. On the other hand, when running `wakunode` throughout `heaptrack`, the stats are properly written periodically.

This docker file also enforces `nimbus-build-system` to use the nim compiler that points at the [heaptrack_support](https://github.com/status-im/Nim/tree/heaptrack_support) branch. This is achieved by passing `NIM_COMMIT=heaptrack_support` to the `make wakunode2` call.

## How to use it
0. Go to the nwaku root folder.
1. Create a docker image with a command like the next:
   `sudo make docker-image DOCKER_IMAGE_NAME=docker_repo:docker_tag HEAPTRACKER=1`
2. Run the docker container as usual, giving the needed wakunode2 params.
3. Wait for some minutes until the mem-stats file gets populated.
4. copy out the stats file: `sudo docker cp 768e7de52d3c:/heaptrack.wakunode.1.gz .` (replace 768.. with your docker container id.).
5. Analyse the file with: `bin/heaptrack heaptrack.wakunode.1.gz` ([Heaptrack](https://github.com/KDE/heaptrack) should be built locally beforehand.).

## Additional hints
https://github.com/waku-org/nwaku/pull/1682#issuecomment-1513260444


.dockerignore
```diff
@@ -1,5 +1,6 @@
 /README.md
 /Dockerfile
+/Dockerfile_with_heaptrack
```
LNSD, line 3: I think this line can be removed now :)

Ivansete-status:
I've applied @vpavlin's [recommendation](https://github.com/waku-org/nwaku/pull/1682#discussion_r1170066461) and now we only install libunwind (`apk add libunwind`). Thanks!

Dockerfile
```diff
@@ -52,5 +52,35 @@ COPY --from=nim-build /app/migrations/ /app/migrations/
...
+
+WORKDIR /heaptrack/build
+RUN cmake -DCMAKE_BUILD_TYPE=Release ..
+RUN make -j$(nproc)
+
+
+# Debug image
+FROM prod AS debug
+
+RUN apk add --no-cache gdb
```
vpavlin, line 77: It would be good to only call apk once

```suggestion
RUN apk add --no-cache gdb libunwind
```

Dockerfile
```diff
@@ -52,5 +52,35 @@ COPY --from=nim-build /app/migrations/ /app/migrations/
...
+
+# Add heaptrack
+COPY --from=heaptrack-build /heaptrack/build/ /heaptrack/build/
+RUN apk add libunwind
```
vpavlin, line 81:
```suggestion
```

vpavlin (changes requested):
Suggested a change, but otherwise LGTM

jm-clius (commented):
Thanks for this! No strong opinion, but since we'll still have to add `-d:heaptrack` or `-d:heaptrack -d:heaptrack_inject` to the [`make` command] I think we at least need another flag/`ARG` for `HEAPTRACK` or `DEBUG`. Since this could pollute the Dockerfile, wouldn't an alternative be to create a new Dockerfile specifically for these debug builds under say `/tools/nwaku-heaptrack` ? cc @vpavlin here for advice - not sure what standard/best practice is here re number of Dockerfiles: opting for separate files for different use cases or a single Dockerfile, but with flags/ARGs for the different build scenarios?

vpavlin:
If the Dockerfiles are essentially copies of each other, then that feels like a bad practice because we may forget to update one or another and then we potentially get different images, hence different behaviour of waku node. For such case it would make sense to "polute" the Dockerfile with more ARGs and instruct Makefile.

If one Dockerfile is just an exetension of another (i.e. building on top of another image build) or is separate then it makes sense to have separate Dockerfiles to keep the main one as clean as possible.

I was not aware we need to add `-d:heaptrack` to the wakunode build based on @Ivansete-status comments

jm-clius:
> If the Dockerfiles are essentially copies of each other, then that feels like a bad practice

Good point and thanks. I think proper use of ARGs here in the same Dockerfile is then the way to go.

Dockerfile
Ivansete-status: Good point! I'll do that thanks !

Ivansete-status:
Actually I was preparing the next commit by adding a combination of new 'heaptrack' args to the Dockerfile plus Makefile.
I'll commit once I confirm it works.

Dockerfile
```diff
@@ -1,10 +1,11 @@
-# BUILD IMAGE ------------------------------------------------------------------
+# BUILD NIM APP ----------------------------------------------------------------
 
 FROM alpine:edge AS nim-build
 
 ARG NIMFLAGS
 ARG MAKE_TARGET=wakunode2
 ARG EXPERIMENTAL=false
+ARG NIM_COMMIT_ARG
```
vpavlin, line 8: I don't think the ARG shoul have the suffix ARG:)

Ivansete-status: ah yes, I change it now. thanks :)

Makefile
```diff
@@ -54,11 +54,31 @@ clean:
...
 ## Git version
 GIT_VERSION ?= $(shell git describe --abbrev=6 --always --tags)
 NIM_PARAMS := $(NIM_PARAMS) -d:git_version=\"$(GIT_VERSION)\"
 
+## Heaptracker options
+HEAPTRACKER ?= 0
+HEAPTRACKER_INJECT ?= 0
+ifeq ($(HEAPTRACKER), 1)
+# Needed to make nimbus-build-system use the Nim's 'heaptrack_support' branch
+NIM_COMMIT := NIM_COMMIT=heaptrack_support
```
vpavlin, line 69: Would this result in `NIM_COMMIT` having a value `NIM_COMMIT=heaptrack_support` - i.e. trying to checkout branch `NIM_COMMIT=heaptrack_support`, rather than `heaptrack_support`?

Ivansete-status: Ok, I've applied your suggestions and I think it now looks much better. Thanks @vpavlin !

vpavlin (approved):
LGTM!

PR #1682, opened 2023-04-18
build: merge heaptrack dockerfile with production dockerfile

LNSD:
Both docker files can be consolidated into a single file by adding a `debug` stage and slightly modifying the `Makefile`.

> **Note**
> I am having local issues with my docker setup and I couldn't test them. I am pretty sure these changes should work. Feel free to test them and accept the PR.


Dockerfile
```diff
@@ -52,5 +52,35 @@ COPY --from=nim-build /app/migrations/ /app/migrations/
...
+
+
+# Debug image
+FROM prod AS debug
+
+RUN apk add --no-cache gdb
+
+# Add heaptrack
+COPY --from=heaptrack-build /heaptrack/build/ /heaptrack/build/
+COPY --from=heaptrack-build /usr/lib/ /usr/lib/
```
vpavlin, line 81: @Ivansete-status 

This is weird, I would not blindly copy `/usr/lib/` between images (especially different versions of images) as you might mess up some libraries.

I assume you are doing this because heaptrack actually drops some files in there? It would be good to know which files and only copy those rather than whole  `/usr/lib`

Ivansete-status: I see the point. I will check it asap to have only the needed files copied.
Thanks @vpavlin !

Ivansete-status:
Thanks for the PR @LNSD ! I like it :)

vpavlin:
From #1681:

> copy out the stats file: sudo docker cp 768e7de52d3c:/heaptrack.wakunode.1.gz . (replace 768.. with your docker container id.).

I would recommend using `volume` rather than copying the file from container - that way you make sure that the file will be persisted even if the container or machine reboots.

Something like

```
mkdir -p heaptrack/
docker ... --workdir /tmp/heaptrack -v $PWD/heaptrack:/tmp/heaptrack:z ...
```

We could also add `WORKDIR /tmp/heaptrack` into the Dockerfile, so that we can skip that option in `docker run`

Sadly it does not seem we can influence the name of the file heaptrack produces - so we cannot keep files from previous runs around easily.

The only way I was able to "simulate" this was

```
touch heaptrack/heaptrack.123.gz
docker run ... --workdir /tmp/heaptrack -v $PWD/heaptrack/heaptrack.123.gz:/tmp/heaptrack/heaptrack.wakunode.1.gz:z ...
```

But not sure how useful that is - whether it is worth the complications when running it:)

vpavlin:
In general LGTM, I was able to run it and it produced the heaptrack file, but sadly,I was not able to open it - heaptrack gave me a qt5 related error and also the produced files were only 10bytes large, so probably no content - not sure if I did something wrong, or it just did not ran long enough?

Dockerfile
vpavlin: Disregard, there are some libs missing when you don't include /usr/lib :)

Ivansete-status (approved):
The changes LGTM. Thanks !

PR #1684, opened 2023-04-18
peers.nim: adding more detail in case the p2pPart doesn't contain '/'

Ivansete-status:
Merely adds the content of `p2pPartStr` to the returned error in `parsePeerInfo` if the next error occurs:

` Error in parsePeerInfo: p2p part should contain /`


alrevuelta (approved):
lgtm! always nice to be more verbose with errors 👍

PR #1687, opened 2023-04-19
Adding more detail in an error produced by 'parsePeerInfo'

Ivansete-status:


waku/v2/utils/peers.nim
```diff
@@ -136,8 +136,8 @@ proc parsePeerInfo*(peer: RemotePeerInfo|string):
 
   let p2pPartStr = p2pPart.toString()[]
   if not p2pPartStr.contains("/"):
-    let msg = "Error in parsePeerInfo: p2p part should contain / [" &
-          p2pPartStr & "]"
+    let msg = "Error in parsePeerInfo: p2p part should contain / [p2pPartStr:" &
+          p2pPartStr & "] [peer:" & cast[string](peer) & "]"
```
alrevuelta, line 140: mm not sure this casts are a good practice in nim. if we used `when` it should be safer and there is no need to cast it. saw we have others in this function.

something like this should examplify it.

```
proc twoTypes(a: string|int) = 
  when a is int:
    echo a + 3

  when a is string:
    echo a & "..."

twoTypes(3)
twoTypes("10")
```

what do you think?

Ivansete-status: Thanks for the hint @alrevuelta !
I think in this case the case is not needed because we previously control the case where `peer` is of type `RemotePeerInfo` so maybe we should just use the `peer` value without cast?

alrevuelta: yep, with my example using `when` there is not need to cast it, so its safer imho.

LNSD: @Ivansete-status Avoid using `cast[x](...)` operator.
```suggestion
          p2pPartStr & "] [peer:" & string(peer) & "]"
```

Or:

```suggestion
          p2pPartStr & "] [peer:" & peer.string & "]"
```

Or, in this special case (convert to string):

```suggestion
          p2pPartStr & "] [peer:" & $peer & "]"
```

Ivansete-status: Got it thank guys ! I will apply these now :)

LNSD (approved):
LGTM ✅

alrevuelta (approved):

PR #1690, opened 2023-04-19
refactor(wakunode2): split setup logic into app module

LNSD:
To be able to test the `wakunode2` application, the following changes were performed:

- [x] Moved all `wakunode2` setup logic to `app.nim` module.
- [x] Encapsulated the setup logic within an `App` type.
- [x] Added support for gracefully shutting down the app node and the associated servers (REST and JSON-RPC).
- [x] Added smoke tests and a `testwakunode2` make target, covering the `App` type setup logic.
- [x] Migrated the `main` procedure to use the `App` setup logic.

There is still work to be done in the `App,` for example, to better reorganize the setup logic and expose the different APIs or extend the current test coverage.


Ivansete-status (approved):

rymnc (approved):
LGTM, thanks

jm-clius (approved):
Thanks! Looks reasonable to me.

PR #1692, opened 2023-04-19
fix(networkmonitor): break import dependency with wakunode2 app

LNSD:
Importing "app" (or tool) modules within other modules is not a good practice. App modules should be final (roots of the imports tree). This PR breaks the `networkmonitor` tool's import dependency on the `wakunode2` app.

- [x] Copied the `retrieveDynamicBootstrapNodes` from `wakunode2` to the `networkmonitor` module to break the dependency.
- [x] Fixed a couple of minor deprecation warnings (e.g., use `Duration` as `asyncSleep` argument).

### DIFF
--- a/tools/networkmonitor/networkmonitor.nim
+++ b/tools/networkmonitor/networkmonitor.nim
@@ -5,6 +5,7 @@ else:
 
 import
   std/[tables,strutils,times,sequtils],
+  stew/results,
   stew/shims/net,
   chronicles,
   chronicles/topics_registry,
@@ -14,12 +15,11 @@ import
   eth/keys,
   eth/p2p/discoveryv5/enr,
   libp2p/crypto/crypto,
+  libp2p/nameresolving/dnsresolver,
   metrics,
   metrics/chronos_httpserver,
   presto/[route, server, client]
-
 import
-  ../../apps/wakunode2/wakunode2,
   ../../waku/v2/node/peer_manager,
   ../../waku/v2/waku_node,
   ../../waku/v2/waku_core,
@@ -148,7 +148,7 @@ proc populateInfoFromIp(allPeersRef: CustomPeersTableRef,
     var location: NodeLocation
     try:
       # IP-API endpoints are now limited to 45 HTTP requests per minute
-      await sleepAsync(1400)
+      await sleepAsync(1400.millis)
       let response = await restClient.ipToLocation(allPeersRef[peer].ip)
       location = response.data
     except CatchableError:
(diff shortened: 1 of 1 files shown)


tools/networkmonitor/networkmonitor.nim
```diff
@@ -193,7 +193,33 @@ proc crawlNetwork(node: WakuNode,
     # we dont run ipMajorityLoop
     # we dont run revalidateLoop
 
-    await sleepAsync(crawlInterval)
+    await sleepAsync(crawlInterval.millis)
+
+proc retrieveDynamicBootstrapNodes(dnsDiscovery: bool, dnsDiscoveryUrl: string, dnsDiscoveryNameServers: seq[ValidIpAddress]): Result[seq[RemotePeerInfo], string] =
```
rymnc, line 198: maybe a comment here that this is the same as the proc in wakunode2?

rymnc (approved):
LGTM

tools/networkmonitor/networkmonitor.nim
```diff
@@ -193,7 +193,33 @@ proc crawlNetwork(node: WakuNode,
...
+      nameServers.add(initTAddress(ip, Port(53))) # Assume all servers use port 53
+
+    let dnsResolver = DnsResolver.new(nameServers)
+
+    proc resolver(domain: string): Future[string] {.async, gcsafe.} =
+      trace "resolving", domain=domain
+      let resolved = await dnsResolver.resolveTxt(domain)
+      return resolved[0] # Use only first answer
+
+    var wakuDnsDiscovery = WakuDnsDiscovery.init(dnsDiscoveryUrl, resolver)
```
Ivansete-status, line 214: Not bit deal but shall use let instead?
```suggestion
    let wakuDnsDiscovery = WakuDnsDiscovery.init(dnsDiscoveryUrl, resolver)
```

Ivansete-status (approved):

alrevuelta (approved):

tools/networkmonitor/networkmonitor.nim
LNSD: This method probably will change in `wakunode2` app after the refactoring feature branch is merged into the `master` branch. So... I don't think a comment will add any value 🤷🏼

tools/networkmonitor/networkmonitor.nim
LNSD: Actually, this broke the build 🤣 Reverting it!

LNSD:
Merging this PR since [the CI checks failure](https://github.com/waku-org/nwaku/actions/runs/4745380376/jobs/8427616948?pr=1692#step:5:4306) is unrelated to this PR's changes.

cc @alrevuelta

PR #1697, opened 2023-04-20
feat: add metrics with msg size histogram

### LINKED ISSUE #1626: feat: add metric with average message size
We currently have `waku_node_messages` metric that counts the number of rx messages. This metric allows to be processed in tools like grafana to extract the amount of messages per second or minute, which gives great observability on the node, network and pubsub topic in general.

I would like to suggest adding a `waku_avg_message_size` metric that keeps track of the average message size of the messages that are rx by the node. Just like the already existing metric, on every message that is received, we update the metric (see `traceHandler`)

This can be archieved by an inexpensive rolling moving average algorithm, that doesn't require all samples to be stored to calculate the averages. There is some deviation from the traditional average (where are values must be known), but totally neglectable for our use case.

`new_average = old_average * (n-1)/n + new_value /n`

Where:
* `old_average`: is the existing `waku_avg_message_size` value
* `new_average`: is the new calculated value of `waku_avg_message_size`.
* `new_value` is the new sample size, in our case `len(msg.bytes)`
* `n`: is the moving average factor (amount of messages that we average)

There are some nuances like if no messages are rx, the metric will remain unchanged with the latest value.

Very useful for simulations imho, as we should not only look at msg rate but also msg size.

alrevuelta:
closes #1626 

* Add a new metric `waku_histogram_message_size` containing the message size histogram.
* This allows to have better observability on the message distribution size.

Suggested usage:

image

* 98% of the messages are between 5 and 15 kBytes
* 0.7% of the messages are between 50 and 100 kBytes.
* 99% of the messages are lower than 37.6 kBytes.

### DIFF
--- a/waku/v2/node/waku_node.nim
+++ b/waku/v2/node/waku_node.nim
@@ -48,7 +48,8 @@ when defined(rln):
     ../waku_rln_relay
 
 declarePublicCounter waku_node_messages, "number of messages received", ["type"]
-declarePublicHistogram waku_histogram_message_size, "message size histogram in kB", buckets = [0.0, 5.0, 15.0, 50.0, 100.0, 300.0, 700.0, 1000.0]
+declarePublicHistogram waku_histogram_message_size, "message size histogram in kB",
+  buckets = [0.0, 5.0, 15.0, 50.0, 100.0, 300.0, 700.0, 1000.0, Inf]
 
 declarePublicGauge waku_version, "Waku version info (in git describe format)", ["version"]
 declarePublicGauge waku_node_errors, "number of wakunode errors", ["type"]
--- a/waku/v2/node/waku_node.nim
+++ b/waku/v2/node/waku_node.nim
@@ -47,8 +47,10 @@ when defined(rln):
   import
     ../waku_rln_relay
 
-declarePublicGauge waku_version, "Waku version info (in git describe format)", ["version"]
 declarePublicCounter waku_node_messages, "number of messages received", ["type"]
+declarePublicHistogram waku_histogram_message_size, "message size histogram in kB", buckets = [0.0, 5.0, 15.0, 50.0, 100.0, 300.0, 700.0, 1000.0]
+
+declarePublicGauge waku_version, "Waku version info (in git describe format)", ["version"]
 declarePublicGauge waku_node_errors, "number of wakunode errors", ["type"]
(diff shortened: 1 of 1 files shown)


waku/v2/node/waku_node.nim
```diff
@@ -47,8 +47,10 @@ when defined(rln):
   import
     ../waku_rln_relay
 
-declarePublicGauge waku_version, "Waku version info (in git describe format)", ["version"]
 declarePublicCounter waku_node_messages, "number of messages received", ["type"]
+declarePublicHistogram waku_histogram_message_size, "message size histogram in kB", buckets = [0.0, 5.0, 15.0, 50.0, 100.0, 300.0, 700.0, 1000.0]
```
jm-clius, line 51: Perhaps I'm missing something about how `nim-metrics` work, but should the `Inf` bucket be declared?

jm-clius (approved):
LGTM

waku/v2/node/waku_node.nim
alrevuelta: Indeed, added in e27702c89b8ed7cc06d67bd2a6f029915e029195 to be explicit.

But looks like its automatically added. Have seen examples that both provide `Inf` and not, but seems that even without it its automatically added. Example without having `Inf` in bucket.

```console 
$ curl localhost:8004/metrics | grep histogram
...
waku_histogram_message_size_sum 0.0
waku_histogram_message_size_count 0.0
waku_histogram_message_size_created 1681987077.0
waku_histogram_message_size_bucket{le="0.0"} 0.0
waku_histogram_message_size_bucket{le="5.0"} 0.0
waku_histogram_message_size_bucket{le="15.0"} 0.0
waku_histogram_message_size_bucket{le="50.0"} 0.0
waku_histogram_message_size_bucket{le="100.0"} 0.0
waku_histogram_message_size_bucket{le="300.0"} 0.0
waku_histogram_message_size_bucket{le="700.0"} 0.0
waku_histogram_message_size_bucket{le="1000.0"} 0.0
waku_histogram_message_size_bucket{le="+Inf"} 0.0   <---
```

jm-clius: Thanks! I thought it may be added automatically, but wasn't sure.

vpavlin (approved):
Awesome! LGTM

PR #1709, opened 2023-05-02
feat(discv5): added find random nodes with predicate

LNSD:
# Description

Waku discovery5 implementation does random walks through the DHT and reports the discovered nodes. To support Waku relay static shards discovery, we want to announce in the ENR the shards the node is subscribed to and keep only the discovered nodes that match specific criteria (a.k.a. predicate).

# Changes

- [x] Simplified and refactored some parts of the `WakuDiscv5` implementation (e.g., moved from exceptions to `Result`s) 
- [x] Added a _predicate_ argument to the "find random peers" method to filter out records that do not match the predicate.
- [x] Reworked the Waku Discv5 test suites.

### DIFF
--- a/waku/v2/node/waku_node.nim
+++ b/waku/v2/node/waku_node.nim
@@ -853,59 +853,48 @@ proc startKeepalive*(node: WakuNode) =
   asyncSpawn node.keepaliveLoop(defaultKeepalive)
 
 proc runDiscv5Loop(node: WakuNode) {.async.} =
-  ## Continuously add newly discovered nodes
-  ## using Node Discovery v5
-  if (node.wakuDiscv5.isNil):
+  ## Continuously add newly discovered nodes using Node Discovery v5
+  if node.wakuDiscv5.isNil():
     warn "Trying to run discovery v5 while it's disabled"
     return
 
-  info "Starting discovery loop"
+  info "starting discv5 discovery loop"
 
   while node.wakuDiscv5.listening:
-    trace "Running discovery loop"
-    let discoveredPeersRes = await node.wakuDiscv5.findRandomPeers()
+    trace "running discv5 discovery loop"
+    let discoveredRecords = await node.wakuDiscv5.findRandomPeers()
+    let discoveredPeers = discoveredRecords.mapIt(it.toRemotePeerInfo()).filterIt(it.isOk()).mapIt(it.value)
 
-    if discoveredPeersRes.isOk:
-      let discoveredPeers = discoveredPeersRes.get
-      let newSeen = discoveredPeers.countIt(not node.peerManager.peerStore[AddressBook].contains(it.peerId))
-      info "Discovered peers", discovered=discoveredPeers.len, new=newSeen
+    for peer in discoveredPeers:
+      let isNew = not node.peerManager.peerStore[AddressBook].contains(peer.peerId)
+      if isNew:
+        debug "new peer discovered", peer= $peer, origin= "discv5"
 
-      # Add all peers, new ones and already seen (in case their addresses changed)
-      for peer in discoveredPeers:
-        node.peerManager.addPeer(peer, Discv5)
+      node.peerManager.addPeer(peer, PeerOrigin.Discv5)
 
     # Discovery `queryRandom` can have a synchronous fast path for example
     # when no peers are in the routing table. Don't run it in continuous loop.
     #
-    # Also, give some time to dial the discovered nodes and update stats etc
+    # Also, give some time to dial the discovered nodes and update stats, etc.
     await sleepAsync(5.seconds)
 
-proc startDiscv5*(node: WakuNode): Future[bool] {.async.} =
+proc startDiscv5*(node: WakuNode): Future[Result[void, string]] {.async.} =
   ## Start Discovery v5 service
+  if node.wakuDiscv5.isNil():
+    return err("discovery v5 is disabled")
 
   info "Starting discovery v5 service"
+  let res = node.wakuDiscv5.start()
+  if res.isErr():
+    return err(res.error)
 
-  if not node.wakuDiscv5.isNil():
-    ## First start listening on configured port
-    try:
-      trace "Start listening on discv5 port"
-      node.wakuDiscv5.open()
-    except CatchableError:
-      error "Failed to start discovery service. UDP port may be already in use"
-      return false
+  trace "Start discovering new peers using discv5"
+  asyncSpawn node.runDiscv5Loop()
 
-    ## Start Discovery v5
-    trace "Start discv5 service"
-    node.wakuDiscv5.start()
-    trace "Start discovering new peers using discv5"
+  debug "Successfully started discovery v5 service"
+  info "Discv5: discoverable ENR ", enr = node.wakuDiscV5.protocol.localNode.record.toUri()
+  return ok()
 
-    asyncSpawn node.runDiscv5Loop()
-
-    debug "Successfully started discovery v5 service"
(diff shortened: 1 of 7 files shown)


waku/v2/node/waku_node.nim
```diff
@@ -853,59 +853,47 @@ proc startKeepalive*(node: WakuNode) =
...
+    let discoveredRecords = await node.wakuDiscv5.findRandomPeers()
+    let discoveredPeers = discoveredRecords.mapIt(it.toRemotePeerInfo()).filterIt(it.isOk()).mapIt(it.value)
 
-    if discoveredPeersRes.isOk:
-      let discoveredPeers = discoveredPeersRes.get
-      let newSeen = discoveredPeers.countIt(not node.peerManager.peerStore[AddressBook].contains(it.peerId))
-      info "Discovered peers", discovered=discoveredPeers.len, new=newSeen
+    for peer in discoveredPeers:
+      let isNew = not node.peerManager.peerStore[AddressBook].contains(peer.peerId)
+      debug "peer discovered", peer= $peer, origin= "discv5", is_new= $isNew
```
alrevuelta, line 870: isnt this to much to log one line for every peer? perhaps we can trace it, but debug is too much imho.

i wouldnt remove the "Discovered peers" as its quite relevant to see if we are finding new peers. Im fine with moving it to debug if info is too verbose, but wouldnt remove it.

alrevuelta (changes requested):

waku/v2/waku_discv5.nim
```diff
@@ -85,49 +73,55 @@ proc new*(T: type WakuDiscoveryV5,
...
+  let record = block:
+        var builder = EnrBuilder.init(privateKey)
+        builder.withIpAddressAndPorts(
+            ipAddr = extIp,
+            tcpPort = extTcpPort,
+            udpPort = extUdpPort,
+        )
+        builder.withWakuCapabilities(flags)
+        builder.withMultiaddrs(multiaddrs)
+        builder.build().expect("Record within size limits")
```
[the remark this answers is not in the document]
LNSD: Note that this method is deprecated in favor of the other constructor that does not build the record within the proc. Once all calls to this method are replaced with the other one, this should be removed.

In my opinion, records should be constructed in the App initialization phase and passed to this protocol already constructed. Adding another responsibility (creating an ENR record instance) is not a good practice.

I opted for a `expect` as the underlying implementation (nim-eth's Discv5) does. In the underlying implementation, if the "created" record length exceeds the EIP's maximum allowed length, it "panics". This one behaves exactly the same way, one level above.

waku/v2/node/waku_node.nim
LNSD: > isnt this to much to log one line for every peer? perhaps we can trace it, but debug is too much imho.

Maybe. We can just log the new peers in debug level.


> i wouldnt remove the "Discovered peers" as its quite relevant to see if we are finding new peers. Im fine with moving it to debug if info is too verbose, but wouldnt remove it.

It is not about verbosity. `info` level should never be used in libraries. This goes against standard practices and, as a consequence, the expectations of the library users. The standard is that a library must not emit logs unless the `debug` level is enabled.

Typically, `info`, `warn`, and `error` are application-specific log levels. If one library uses the `error` level clearly shows that there is an API design flaw. We have `Result`s and `Exception`s to convey these "erroneous states" up in the call chain.

Based on this reasoning: `waku/v2/waku_node` is a library module. It should not use `info`, `warn`, or `error` logs. If used, those are code smells and should be converted to either `Result`/`Defect` or pushed one layer up to the application: `wakunode2/app.nim#App`.

----

Note that this is not just my opinion; it is based on facts and good practices and aligns with a discussion about logging that happened in the Nimbus server (for example).

LNSD: > > isnt this to much to log one line for every peer? perhaps we can trace it, but debug is too much imho.
> 
> Maybe. We can just log the new peers in debug level.

Changed here: https://github.com/waku-org/nwaku/compare/975924dd1b6c1e7c67aa49a6ef7a14448571af5d..b972dd88a66f7afa97410b133e9825bec817dd18

waku/v2/waku_discv5.nim
```diff
@@ -136,26 +130,51 @@ proc closeWait*(wd: WakuDiscoveryV5) {.async.} =
   wd.listening = false
   await wd.protocol.closeWait()
 
-proc findRandomPeers*(wd: WakuDiscoveryV5): Future[Result[seq[RemotePeerInfo], cstring]] {.async.} =
+proc findRandomPeers*(wd: WakuDiscoveryV5, pred: WakuDiscv5Predicate = nil): Future[seq[waku_enr.Record]] {.async.} =
```
jm-clius, line 133: This `pred` is not set yet anywhere. In other words, the remaining work here is to add the static topic configuration and construct a predicate that checks ENRs against that. LMK if you planned to do this or if we should schedule it our side (no expectations, of course, just don't want to interfere if you have more PRs lined up :) )

jm-clius (approved):
LGTM, thanks!

alrevuelta (approved):

waku/v2/node/waku_node.nim
alrevuelta: not sure I follow.

This log has been removed, which I think is important.

`info "Discovered peers", discovered=discoveredPeers.len, new=newSeen`

If you dont want to use `info`, fine, use `debug`.

and regarding the log you have added, having `debug "peer discovered"` (which is logged for every discovered peer every discovery hearbeat) is going to add too many logs.

not a blocker though.

Ivansete-status (approved):

vpavlin (approved):
LGTM

vpavlin:
Seems like macos test is failing with

```
2023-05-02T15:07:34.3574730Z     /Users/runner/work/nwaku/nwaku/tests/v2/test_waku_discv5.nim(145, 48): Check failed: node2.wakuDiscv5.protocol.nodesDiscovered == 2
2023-05-02T15:07:34.3575280Z     node2.wakuDiscv5.protocol.nodesDiscovered was 1
```

Is it due to "missing" sleep?:)

LNSD:
> Seems like macos test is failing with
> 
> ```
> 2023-05-02T15:07:34.3574730Z     /Users/runner/work/nwaku/nwaku/tests/v2/test_waku_discv5.nim(145, 48): Check failed: node2.wakuDiscv5.protocol.nodesDiscovered == 2
> 2023-05-02T15:07:34.3575280Z     node2.wakuDiscv5.protocol.nodesDiscovered was 1
> ```
> 
> Is it due to "missing" sleep?:)

Hehehe 😆 

No, I already increased the "sleep time," and the issue still happens. I have reduced the test checks to the bare minimum, just "sanity checks". It should be sufficient.

waku/v2/waku_discv5.nim
jm-clius: @LNSD, I apologise for not having seen this comment earlier. At some point I must have added it to a TODO and then lost track :)

For now I think the best idea is to use the `--topics` config option with the app then identifying this as a static shard configuration and (somehow) passing in the predicate for discv5 filtering. This indeed points to the fact that at least certain types of stand-beside discovery should happen as a service separate to the node. For now, depending on capacity, we may do something simpler first, but I agree that this is the direction we should be taking.

That said, you've helped us a lot so happy to take ownership to get this and the subsequent features merged. To make it a bit easier fort us to rebase (which may help fix some unrelated tests) I've merged your changes to a base repo branch in https://github.com/waku-org/nwaku/pull/1763. If you agree that we take over from here (thanks for getting this done!), we can close this PR without merging and continue the feature from there?

jm-clius:
Closing as duplicated and merged elsewhere.

PR #1710, opened 2023-05-02
fix: unstable peers in mesh

alrevuelta:
Closes https://github.com/waku-org/nwaku/issues/1669

Summary:
* Due to a bug in nim-libp2p, amount of peers in mesh were unstable, fluctuating a lot and even dropping below D_low.
* This was fixed upstream in https://github.com/status-im/nim-libp2p/pull/883 and this PR bumps to the latest version containing the fix.
* Tested in the sandbox machine [see](https://github.com/waku-org/nwaku/issues/1669#issuecomment-1522963335)


rymnc (approved):
LGTM, great catch

jm-clius (approved):
Great catch!

PR #1713, opened 2023-05-04
chore: add timestamp and ephemeral for opt-in dos validator

alrevuelta:
- [x] Include the `timestamp` and `ephemeral` of the message in the app hash calculation.
- [x] Reject messages that dont containg a valid timestamp (=0)
- [x] Reject messages not matching timestamp (with some tollerance)
- [x] Update test vectors accordingly
- [x] Update spec accordingly: https://github.com/vacp2p/rfc/pull/598

### DIFF
--- a/apps/wakunode2/wakunode2_validator_signed.nim
+++ b/apps/wakunode2/wakunode2_validator_signed.nim
@@ -4,6 +4,7 @@ else:
   {.push raises: [].}
 
 import
+  std/[math,times],
   chronicles,
   chronos,
   metrics,
@@ -14,6 +15,8 @@ import
   libp2p/protocols/pubsub/errors,
   nimcrypto/sha2,
   secp256k1
+
+const MessageWindowInSec = 5*60 # +- 5 minutes
 
 import
   ../../waku/v2/waku_relay/protocol,
@@ -35,6 +38,21 @@ proc msgHash*(pubSubTopic: string, msg: WakuMessage): array[32, byte] =
 
   return ctx.finish()
 
+proc withinTimeWindow*(msg: WakuMessage): bool =
+  # Returns true if the message timestamp is:
+  # abs(now - msg.timestamp) < MessageWindowInSec
+  let ts = msg.timestamp
+  let now = getNanosecondTime(getTime().toUnixFloat())
+  let window = getNanosecondTime(MessageWindowInSec)
+
+  if now > ts:
+    if now - ts < window:
+      return true
+  else:
+    if ts - now < window:
+      return true
+  return false
+
 proc addSignedTopicValidator*(w: WakuRelay, topic: PubsubTopic, publicTopicKey: SkPublicKey) =
   debug "adding validator to signed topic", topic=topic, publicTopicKey=publicTopicKey
 
@@ -44,11 +62,12 @@ proc addSignedTopicValidator*(w: WakuRelay, topic: PubsubTopic, publicTopicKey: 
 
     if msg.isOk():
       if msg.get.timestamp != 0:
-        let msgHash = SkMessage(topic.msgHash(msg.get))
-        let recoveredSignature = SkSignature.fromRaw(msg.get.meta)
-        if recoveredSignature.isOk():
-          if recoveredSignature.get.verify(msgHash, publicTopicKey):
-            outcome = errors.ValidationResult.Accept
+        if msg.get.withinTimeWindow():
+          let msgHash = SkMessage(topic.msgHash(msg.get))
+          let recoveredSignature = SkSignature.fromRaw(msg.get.meta)
+          if recoveredSignature.isOk():
+            if recoveredSignature.get.verify(msgHash, publicTopicKey):
+              outcome = errors.ValidationResult.Accept
 
     waku_msg_validator_signed_outcome.inc(labelValues = [$outcome])
     return outcome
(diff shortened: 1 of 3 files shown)


apps/wakunode2/wakunode2_validator_signed.nim
```diff
@@ -29,9 +33,26 @@ proc msgHash*(pubSubTopic: string, msg: WakuMessage): array[32, byte] =
...
   return ctx.finish()
 
+proc withinTimeWindow*(msg: WakuMessage): bool =
+  # Returns true if the message timestamp is:
+  # abs(now - msg.timestamp) < MessageWindowInSec
+  let ts = msg.timestamp
+  let now = getNanosecondTime(getTime().toUnixFloat())
+  let window = getNanosecondTime(MessageWindowInSec)
+
+  if now > ts:
```
alrevuelta, line 48: open for other ways.

cant do abs(now-ts) since im working with **u**int64 and if I cast to **s**igned, I will lose precision. So its the way I found.

richard-ramos: Maybe https://nim-lang.org/docs/times.html#abs%2CDuration from `times` can be used

richard-ramos (approved):

apps/wakunode2/wakunode2_validator_signed.nim
vpavlin: Sorry, mabe it is a completely stupid question, but would losing precision matter? Is the concern that someone could make the difference between TS and now overflow and thus the result of abs() be < window?

Can we check for overflow?

vpavlin (approved):
LGTM

apps/wakunode2/wakunode2_validator_signed.nim
```diff
@@ -29,9 +33,26 @@ proc msgHash*(pubSubTopic: string, msg: WakuMessage): array[32, byte] =
...
+  ctx.update(msg.timestamp.uint64.toBytes(Endianness.littleEndian))
+  ctx.update(if msg.ephemeral: @[1.byte] else: @[0.byte])
 
   return ctx.finish()
 
+proc withinTimeWindow*(msg: WakuMessage): bool =
+  # Returns true if the message timestamp is:
+  # abs(now - msg.timestamp) < MessageWindowInSec
+  let ts = msg.timestamp
+  let now = getNanosecondTime(getTime().toUnixFloat())
```
jm-clius, line 45: `waku_core` (exporting `time`) already provides `getNowInNanosecondTime()` so you don't have to import `std/times`

apps/wakunode2/wakunode2_validator_signed.nim
```diff
@@ -4,16 +4,20 @@ else:
   {.push raises: [].}
 
 import
+  std/[math,times],
```
jm-clius, line 7: I think we should try to avoid importing `std/times` (especially when we use `chronos.timer` constructs in the same module). In this case we need it for absolute system time, but `waku_core/time` provides the wrappers for that.

jm-clius (approved):
LGTM. We need overflow-proof arithmetic for waku `Timestamp`s. In general, though, we want to stop people from directly having to choose between `std/times` and `chronos/timer` and just use `waku_core/time` which should eventually wrap what is needed from both the above. This is beyond the scope of this PR, of course.

apps/wakunode2/wakunode2_validator_signed.nim
alrevuelta: ou `between` is indeed what I needed, but since in nwaku we generally use our own `Timestamp` (and not te `Duration`) I think this may be simpler. Less castings, etc.

> Sorry, mabe it is a completely stupid question, but would losing precision matter? Is the concern that someone could make the difference between TS and now overflow and thus the result of abs() be < window?

mm havent done the maths if matters or not. Difference between `ts` and `now` will never overflow, since it will be always a smaller value. But since both values are `uint64` I need to do `ts-now` or `now-ts` to ensure the result is not negative. Hence why i compare.

Mathematically is just `abs(now-ts)` but in practice thats not possible afaik without converting to int64 with a possible lose in precission.

apps/wakunode2/wakunode2_validator_signed.nim
alrevuelta: yes right, i keep forgeting.

fixed 3092a519137b5db1663a479d6562da6ecfe98cac

apps/wakunode2/wakunode2_validator_signed.nim
alrevuelta: fixed 3092a519137b5db1663a479d6562da6ecfe98cac

apps/wakunode2/wakunode2_validator_signed.nim
jm-clius: Wait, but `Timestamp` is an alias for `int64` not `uint64`, unless I'm missing something?

alrevuelta: ouch indeed. Totally assumed it was `uint64`, since otherwise it would be a waste of bytes (assuming negative timestamps doesnt make sense). my bad.

fixed, way easier da7814a4dd33f9b1120f419763f100bbf67c8339

some quick napkin math @vpavlin int64 overflow with `9223372036854775807`, since its nano seconds thats `9223372036 . 854775807` and `9223372036` in unix timestamp its `Fri Apr 11 2262 23:47:16 GMT+0000` so we are covered.

thanks all!

alrevuelta:
Merging, failing experimental test is a known issue.

Issue #1632, opened 2023-03-30
feat: C bindings for Waku Relay


Ivansete-status:
### Problem
nwaku can't be easily exported to other environments, such as NodeJS or Rust.

### Suggested solution
Have nwaku C bindings so that a NodeJS user can easily start a Waku node in Relay mode.

### Additional context
This is a first step of having a full operative Waku node in other platforms. For now we focus on the Waku Relay protocol.
We want to tackle the nwaku C-bindings option of: "use nwaku with NodeJS" - https://github.com/waku-org/nwaku/issues/1332
The final goal that we chase: "C-bindings for Kotlin, Rust, etc" -  https://github.com/waku-org/nwaku/issues/904

### Acceptance criteria
- [x] Have dynamic library (waku.so) for unix systems.
- [x] Have a static library (waku.a) for unix systems.
- [x] Have a C program that links nwaku libraries. That program should operate as a Relay node, allowing:
   - subscribe/unsubscribe from a PubSub topic.
   - send a message to a PubSub topic.
- [x] Have a npm package so that a "Waku Relay" node can be started from NodeJS.
- [x] Waku relay example in Python.
- [x] Waku relay example in Golang.

PR #1714, opened 2023-05-04
feat(cbindings): first commit - waku relay (#1632)

Ivansete-status:
# Description
This allows the creation of a `libcwakuv2.a` library which exposes the next Waku functions:
```
waku_version()
waku_set_event_callback(..)
waku_content_topic(..)
waku_pubsub_topic(..)
waku_default_pubsub_topic(..)
waku_relay_publish(..)
waku_start()
waku_stop()
waku_relay_subscribe(..)
waku_relay_unsubscribe(..)
waku_connect(..)
waku_poll()

```
# Changes

- [x] Have a static library (libwaku.a) for unix systems.
- [x] Have a C program that links the Nwaku library. That program should mount the Relay protocol, allowing:
          - subscribe/unsubscribe from a PubSub topic.
          - send a message to a PubSub topic.

## How to test

1. Build the `cwaku_example` with:
  `make cwaku_example -j$(cat /proc/cpuinfo | grep processor | wc -l)`
   note: on completion, the `cwaku_example` binary and the `libwaku.a` library will be in the `build` folder.
   
2. Open a terminal and run:
  `./build/cwaku_example --peers=/ip4/127.0.0.1/tcp/60001/p2p/16Uiu2HAmVFXtAfSj4EiR7mL2KvL4EE2wztuQgUSBoj2Jx2KeXFLN --port=60000 --key=364d111d729a6eb6d2e6113e163f017b5ef03a6f94c9b5b7bb1bb36fa5cb07a9`
3. Open a terminal and run:
  `./build/cwaku_example --peers=/ip4/127.0.0.1/tcp/60000/p2p/16Uiu2HAm2eqzqp6xn32fzgGi8K4BuF88W4Xy6yxsmDcW8h1gj6ie --port=60001 --key=0d714a1fada214dead6dc9c7274585eca0ff292451866e7d6d677dc818e8ccd2`
4. Subscribe both nodes to the same pubsubtopic and from one node, publish a message. For that, follow the instructions that are shown in the naïve menu.
i.e.
```
Please, select an option:
	1.) Subscribe to topic
	2.) Connect to other node
	3.) Publish a message
```

## Issue

## Pending tasks
1. Generate dynamic library
3. Pending to know how to retrieve the message_ID when publishing a new message
5. Pending to implement the publish_enc_symmetric and publish_enc_asymmetric, which is already done in go-waku.
6. Pending to implement `waku_relay_enough_peers(char* pubsubTopic)`. This feature is not  currently implemented in 
 `libp2p-1.0.0/libp2p/protocols/pubsub/gossipsub.nim`.
7. @richard-ramos - It might be interesting to offer a `show_version` function in the [rfc](https://rfc.vac.dev/spec/36/). That would be helpful for users to know the library's version. This function is given in this PR.


library/cwakuv2.nim
```diff
@@ -0,0 +1,160 @@
...
+  # {
+  #   error: string;
+  # }
+  return $(%* { "error": message })
+
+################################################################################
+### Exported components
+
+proc show_version(): cstring {.dynlib, exportc.} =
+  return wakuNode2VersionString
```
richard-ramos, line 59: Nice!

library/cwakuv2.nim
```diff
@@ -0,0 +1,160 @@
...
+                          else:
+                            pubSubTopic
+
+  let pubMsgFut = wakunode2.publishMessage(targetPubSubTopic, wakuMessage)
+
+  # With the next loop we convert an asynchronous call into a synchronous one
+  for i in 0 .. timeoutMs:
+    if pubMsgFut.finished():
+      break
+    sleep(1)
```
richard-ramos, line 124: `chronos` has a `withTimeout` function that i think could be used here instead of `sleep`:
https://github.com/status-im/nim-chronos/blob/master/chronos/asyncloop.nim#L1116

library/cwakuv2.nim
```diff
@@ -0,0 +1,160 @@
...
+      version = (uint32) jsonContent["version"].getInt()
+
+    wakuMessage = WakuMessage(
+        # Visit https://rfc.vac.dev/spec/14/ for further details
+        payload: jsonContent["payload"].getStr().toSeq().mapIt(byte (it)),
+        contentTopic: $jsonContent["content_topic"].getStr(),
+        version: version,
+        timestamp: getTime().toUnix(),
+        ephemeral: false
+    )
```
richard-ramos, line 108: It should be possible to specify these values instead of having them hardcoded. The `meta` field is also missing https://github.com/waku-org/nwaku/blob/master/waku/v2/waku_core/message/message.nim#L28 (this field should probably be added to the bindings rfc)

library/events/json_message_event.nim
```diff
@@ -0,0 +1,42 @@
...
+
+  return JsonMessageEvent(
+    eventType: "message",
+    pubSubTopic: pubSubTopic,
+    messageId: "TODO",
+    wakuMessage: JsonMessage(
+        payload: payload,
+        contentTopic: msg.contentTopic,
+        version: msg.version,
+        timestamp: int64(msg.timestamp)
```
richard-ramos, line 37: Some fields from WakuMessage are missing. Maybe it also makes sense to include the `proof` if RLN is enabled? 

What do you think if instead of creating a JsonMessage with the WakuMessage content, to instead serialize it with https://github.com/status-im/nim-json-serialization? I imagine it should be possible to deserialize it too in `waku_relay_publish`. The advantage of doing that is that if in the future new fields are added to `WakuMessage` protobuffer, they'll be serialized/unserialized automatically in the c-bindings API functions

library/events/json_message_event.nim
```diff
@@ -0,0 +1,42 @@
...
+    wakuMessage*: JsonMessage
+
+proc new*(T: type JsonMessageEvent,
+          pubSubTopic: string,
+          msg: WakuMessage): T =
+  # Returns a WakuMessage event as indicated in
+  # https://rfc.vac.dev/spec/36/#jsonmessageevent-type
+
+  var payload = newString(len(msg.payload))
+  copyMem(addr payload[0], unsafeAddr msg.payload[0], len(msg.payload))
```
richard-ramos, line 27: I'm curious about this line. Why is it necessary to copy the payload like this instead of doing it directly?

richard-ramos (commented):
Very cool, i like this PR.
I just left some comments with minor observations and items for discussion

apps/wakunode2/wakunode2.nim
```diff
@@ -149,3 +150,30 @@ when isMainModule:
...
+
+proc unsubscribeCallbackFromTopic*(pubSubTopic: cstring,
+                                   callback: proc(pubsubTopic: string, data: seq[byte]): Future[void] {.gcsafe, raises: [Defect].}) =
+  wakunode2.unsubscribeCallbackFromTopic(pubSubTopic, callback)
+
+proc unsubscribeAllCallbacksFromTopic*(pubSubTopic: cstring) =
+  wakunode2.unsubscribeAllCallbackFromTopic(pubSubTopic)
+
+proc publishMessage*(pubSubTopic: cstring, message: WakuMessage): Future[int] {.gcsafe, async.} =
+  return await wakunode2.publishMessage(pubSubTopic, message)
```
jm-clius, line 166: The way I see it, these do not belong here. The C bindings should be a wrapper around the Nim API defined in `waku_node.nim`. As such the app modules should remain unchanged and preferably unimported elsewhere. Any additional API functionality should go to `waku_node.nim`. Of course, since we need a configured `WakuNode` object for those bindings, we may have to provide `App`-like functionality when initiating the node. For now, though I think the configuration will be much, much simpler than is needed for a `wakunode2` application and providing the basic configuration in a `JsonConfig` via `waku_new()` will be enough.

library/cwakuv2.nim
jm-clius: I think this can be `waku_version()`. `waku_node` already provides `git_version`, so I don't think this needs to be exported elsewhere. If we need to construct something specific, that should be done on the Nim API in `waku_node.nim`.

apps/wakunode2/wakunode2.nim
```diff
@@ -13,30 +14,32 @@ import
...
-  ## 3. Mount and initialize configured protocols
-  ## 4. Start node and mounted protocols
-  ## 5. Start monitoring tools and external interfaces
-  ## 6. Setup graceful shutdown hooks
+var wakunode2 {.threadvar.}: App
+var confRes:ConfResult[WakuNodeConf]
 
-  const versionString = "version / git commit hash: " & app.git_version
+{.pop.} # @TODO confutils.nim(775, 17) Error: can raise an unlisted exception: ref IOError
+proc init*(configFilePath = "") =
```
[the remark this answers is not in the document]
jm-clius: I don't think we should `init` an app here, but operate on the `WakuNode` object as defined in `waku_node.nim` (see comments elsewhere). `wakunode2` is simply an application written around the Nim API. As stated elsewhere, we want to provide C bindings wrapping the Nim API, not the `wakunode2` application

jm-clius (changes requested):
Great to see progress here and thanks for including an example! Most important directional change I suggest is to keep the interface between the `WakuNode` and the `wakunode2` application very clean. Our library is the `WakuNode` (defined in `waku_node.nim`) and it already exposes a Nim API (as well as JSON, REST, etc.). We want to provide C bindings for this library (more or less just wrapping the Nim API) not the `wakunode2` application.

apps/wakunode2/wakunode2.nim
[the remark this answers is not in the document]
danielSanchezQ: > I see `init` here uses a path vs the rfc which suggests a json string. We should decide which approach to use to unify behaviors between implementations. cc: @danielSanchezQ for his thoughts as he has worked with the rust+go bindings before

IMO using a configuration is better. A config file should be used for a binary, a configuration object for a library.

@jm-clius Where can I find the nim api documentation? Don't know if Im understanding correctly, but, does it mean that the `wakunode2` is an external application running and the bindings just communicate with it. Or do this works in a similar way to go-waku in where we do embedded the node within the application?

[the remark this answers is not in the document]
jm-clius: The application simply provides a way to configure and instantiate the node, so I guess more similar to "embedded within the application". The Nim API documentation is sparse and outdated (since it's only used internally), but in theory lives here: https://github.com/waku-org/nwaku/blob/master/docs/api/v2/node.md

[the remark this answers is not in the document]
jm-clius: Oh, it's very outdated, in fact. 😟 @Ivansete-status perhaps a good idea that we clean the Nim API doc up at some point during the development of the C bindings, since we're likely to expand/change it in any case.

library/cwakuv2.nim
Ivansete-status: Super interesting @richard-ramos. Thanks for that!
However, we cannot invoke `await`, and therefore, return a Future[..] in this proc because it is exported to C and this call needs to be synchronous.

apps/wakunode2/wakunode2.nim
```diff
@@ -92,20 +94,19 @@ when isMainModule:
     error "4/7 Mounting protocols failed", error=res5.error
     quit(QuitFailure)
 
+proc startNode*() =
```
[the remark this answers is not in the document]
Ivansete-status: Exaclty, that is a simple split. However that might not be needed, and I will likely revert it back, when I wrap the `waku_node.nim` library instead of the wakunode2 app.

library/events/json_message_event.nim
Ivansete-status: Good point! I will try to directly use `seq[byte]` instead

Ivansete-status: I think I will leave it as is because if I cahnge it to use `seq[byte]` we don't get a readable base64 string. In this case, it appears sth like `"payload":[97,71,57,115,89,81,61,61]`

library/events/json_message_event.nim
Ivansete-status: Yes, I absolutely agree. Nevertheless, I added the fields indicated in the next:
https://rfc.vac.dev/spec/36/#jsonmessage-type

Ivansete-status:
Thanks for the comments ! It is now ready to be reviewed again :)

library/waku.nim
```diff
@@ -0,0 +1,399 @@
...
+import
+  std/[sequtils,times,strformat,json,options],
+  strutils,
+  os
+import
+  chronicles,
+  chronos,
+  libp2p/crypto/secp,
+  stew/shims/net,
+  eth/net/nat as todo_delete_this_module
```
jm-clius, line 11: Leftover? :)

library/waku.nim
```diff
@@ -0,0 +1,399 @@
...
+import
+  ../vendor/nim-libp2p/libp2p/crypto/crypto,
+  ../../waku/common/utils/nat,
+  ../../waku/v2/waku_enr/capabilities,
+  ../../waku/v2/waku_core/message/codec,
+  ../../waku/v2/waku_core/message/message,
+  ../../waku/v2/waku_core/topics/pubsub_topic,
+  ../../waku/v2/node/peer_manager/peer_manager,
+  ../../waku/v2/node/waku_node as waku_node_module,
```
jm-clius, line 20: Nitpick: is this alias necessary?

library/waku.nim
```diff
@@ -0,0 +1,399 @@
+
```
jm-clius, line 1: I would consider calling this module something like `libwaku`.

library/waku.nim
```diff
@@ -0,0 +1,399 @@
...
+                        appVersion: uint,
+                        contentTopicName: cstring,
+                        encoding: cstring,
+                        outContentTopic: var string) {.dynlib, exportc.} =
+  # https://rfc.vac.dev/spec/36/#extern-char-waku_content_topicchar-applicationname-unsigned-int-applicationversion-char-contenttopicname-char-encoding
+  outContentTopic = fmt"{appName}/{appVersion}/{contentTopicName}/{encoding}"
+
+proc waku_pubsub_topic(topicName: cstring, encoding: cstring, outPubsubTopic: var string) {.dynlib, exportc.} =
+  # https://rfc.vac.dev/spec/36/#extern-char-waku_pubsub_topicchar-name-char-encoding
+  outPubsubTopic = fmt"/waku/2/{topicName}/{encoding}"
```
jm-clius, line 263: Note that for pubsub topics we've deprecated the `{encoding}` suffix (https://rfc.vac.dev/spec/23/#pubsub-topic-format).

library/waku.nim
```diff
@@ -0,0 +1,399 @@
...
+proc waku_connect(peerMultiAddr: cstring,
+                  timeoutMs: uint = 10000,
+                  jsonResp: var string): bool
+                  {.dynlib, exportc.} =
+  # peerMultiAddr: comma-separated list of fully-qualified multiaddresses.
+  let peers = ($peerMultiAddr).split(",").mapIt(strip(it))
+
+  let connectFut = node.connectToNodes(peers, source="static")
+  while not connectFut.finished():
+    poll()
```
jm-clius, line 387: Wait, where does `timeoutMs` come in here?

library/waku.nim
```diff
@@ -0,0 +1,399 @@
...
+  $(%* { "result": message })
+
+proc errResp(message: string): string =
+  # Convers an error message into an error-JsonResponse
+  # {
+  #   error: string;
+  # }
+  return $(%* { "error": message })
+
+proc setupNat(natConf, clientId: string, tcpPort, udpPort: Port):
```
jm-clius, line 84: I must be missing some details here, but why wouldn't [`setupNat`](https://github.com/waku-org/nwaku/blob/b277ce1013563b4e76a3f8694f0eefb1f3528a4d/waku/common/utils/nat.nim#L14) in the `common` utility work?

waku/v2/node/waku_node.nim
```diff
@@ -71,6 +70,7 @@ const clientId* = "Nimbus Waku v2 node"
 # Default Waku Filter Timeout
 const WakuFilterTimeout: Duration = 1.days
 
+const wakuNode2VersionString* = "version / git commit hash: " & git_version
```
jm-clius, line 73: It's not followed everywhere, but we want constants to be in `PascalCase` (the exception are for those overridden by compiler definitions, such as `git_version` above).
This would just be `WakuNodeVersionString` (not WakuNode2)

jm-clius (approved):
Really great work with this. Well done! I've added a couple of non-blocking comments. Feel free to merge once those are addressed. 

I would suggest thinking of ways in which we can split the logic into separate modules in future (the main library module is going to grow very large!) I'd suggest using the REST API as inspiration and separate relay, store and other protocol bindings as well as `types`, `serdes`, etc. For now this is OK as a starting point.

jm-clius (approved):
Really great work with this. Well done! I've added a couple of non-blocking comments. Feel free to merge once those are addressed. 

I would suggest thinking of ways in which we can split the logic into separate modules in future (the main library module is going to grow very large!) I'd suggest using the REST API as inspiration and separate relay, store and other protocol bindings as well as `types`, `serdes`, etc. For now this is OK as a starting point.

library/waku.nim
Ivansete-status: The `setupNat` in the `common` is deprecated with the next comment: `deprecated:
  "Unsafe: this proc quits the app if something is not ok".`
  
 Once this PR is fixed I want to raise a separate PR to fix the `setupNat` proc from `common` and only use it. Also, clean the `eth/net/nat as todo_delete_this_module` :)

library/waku.nim
Ivansete-status: Indeed, this is something we need to fix from lower levels. I'll add a `TODO` statement here.

library/waku.nim
Ivansete-status: True, is not needed. Thanks!

library/waku.nim
Ivansete-status: I will clean the `setupNat`, once this is merged, in a separate PR :)

library/libwaku.nim
```diff
@@ -0,0 +1,400 @@
...
+    poll()
+
+  if not connectFut.completed():
+    jsonResp = errResp("Timeout expired.")
+    return false
+
+  return true
+
+proc waku_poll() {.dynlib, exportc.} =
+  poll()
```
richard-ramos, line 397: What is this function used for?

library/libwaku.nim
```diff
@@ -0,0 +1,400 @@
...
+                jsonResp: var string): bool
+                {.dynlib, exportc.} =
+  # @params
+  #  topic: Pubsub topic to subscribe to. If empty, it unsubscribes to the default pubsub topic.
+  if node.wakuRelay == nil:
+    jsonResp = errResp("""Cannot unsubscribe without a callback.
+Kindly set it with the 'waku_set_event_callback' function""")
+    return false
+
+  if node.wakuRelay.isNil():
```
richard-ramos, line 369: for learning purposes, what's the difference between  `wakuRelay == nil` and `wakuRelay.isNil()`?

richard-ramos (approved):
LGTM!

library/libwaku.nim
Ivansete-status: Good point my friend! Actually you discovered a bug hehe.

It seems `isNil()` is more efficient, so I'd rather use it from now on.

> proc isNil*[T](x: ptr T): bool {.noSideEffect, magic: "IsNil".}
> proc isNil*(x: pointer): bool {.noSideEffect, magic: "IsNil".}
> proc isNil*(x: cstring): bool {.noSideEffect, magic: "IsNil".}
> proc isNil*[T: proc](x: T): bool {.noSideEffect, magic: "IsNil".}
>  Fast check whether `x` is nil. This is sometimes more efficient than ## `== nil`.

PR #1715, opened 2023-05-04
test: set LOG_LEVEL to 'ERROR' in order to have less logs in ci jobs

Ivansete-status:
# Description
When a job fails I find it awkward to look for what has happened exactly. This change seeks for making it easier.

Therefore, if I need to debug I would do it locally.

### DIFF
--- a/.github/workflows/ci-experimental.yml
+++ b/.github/workflows/ci-experimental.yml
@@ -84,7 +84,7 @@ jobs:
           key: ${{ runner.os }}-zerokit-${{ steps.submodules.outputs.zerokit-hash }}
 
       - name: Build binaries
-        run: make V=1 LOG_LEVEL=DEBUG QUICK_AND_DIRTY_COMPILER=1 v2
+        run: make V=1 LOG_LEVEL=ERROR QUICK_AND_DIRTY_COMPILER=1 v2
 
   test-v2:
     needs: changes
@@ -122,4 +122,4 @@ jobs:
           key: ${{ runner.os }}-zerokit-${{ steps.submodules.outputs.zerokit-hash }}
 
       - name: Run tests
-        run: make V=1 LOG_LEVEL=DEBUG QUICK_AND_DIRTY_COMPILER=1 test2 testwakunode2
+        run: make V=1 LOG_LEVEL=ERROR QUICK_AND_DIRTY_COMPILER=1 test2 testwakunode2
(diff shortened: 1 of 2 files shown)


vpavlin:
@Ivansete-status may I ask why is the symlink needed?

Ivansete-status:
Morning @vpavlin ! I don't fully control this but I guess is required for any kind of platform specific issues. Not sure though.
On the other hand I will simplify this PR and merely make it to reduce the number of logs in CI jobs.

Ivansete-status:
I've simplified the PR.
@vpavlin, on his comment, mentions a change that I did initially in this PR

```
waku.nims:
	ln -sf waku.nimble $@
``` 

The purpose of that change was to make sure the next Makefile task can be run:
```
testwakunode2: | build deps librln
		$(ENV_SCRIPT) nim testwakunode2 $(NIM_PARAMS) $(EXPERIMENTAL_PARAMS) waku.nims
```
However, I simplified the PR until I fully understand how the `waku.nims` and `waku.nimble` files are managed and whether we may have a better solution :)

jm-clius:
Mmm. I'm not sure I fully agree. Indeed, it is a bit slow and time-consuming to search through `DEBUG` logs in CI, but on the other hand it _does_ allow us to get at least _some_ quick understanding of what went wrong without having to try to reproduce locally. This has allowed me to fix obvious bugs in the past, some of which only happened during CI (because of timing differences between platforms). Of course we want to find a good balance here, but I haven't found it prohibitively difficult to search through the `DBG` level logs so far. Happy to go with majority opinion here, though.

jm-clius (commented):
See comment above. :)

alrevuelta:
No strong opinion on this, debug might be useful to debug possible issues but agree that it is a pita to search logs in ci. Perhaps better to leave it as it is since as Hanno says, it helped me debug issues in the past.

vpavlin:
Same here - it is hard to find anything in the logs, so I usually resort to downloading them and grepping locally. But that said, it would be potentially way less informative with only ERROR logs.

Should we switch to ERROR for builds and keep DEBUG for tests? I am not sure if that would help

Ivansete-status:
> Same here - it is hard to find anything in the logs, so I usually resort to downloading them and grepping locally. But that said, it would be potentially way less informative with only ERROR logs.
> 
> Should we switch to ERROR for builds and keep DEBUG for tests? I am not sure if that would help

Actually, for me it would be nice to reduce the tests logs as well :)

However, I will dismiss this.

What is the best way to look for the particular point that caused a job to fail ?

Thanks for the comments guys!

vpavlin:
I usually download the logs and then search for `[FAILED]` :)

jm-clius:
I just search for `[FAILED]` - it takes a couple of seconds, but the `DBG` surrounding it sometimes help that I can see what went wrong at a glance.

PR #1716, opened 2023-05-05
fix: fix filter v2 proto fields

jm-clius:
Fixes the protofield numbering in the nwaku implementation of https://rfc.vac.dev/spec/12/.

The `status_code` and `status_desc` fields were using field numbers `2` and `3`, whereas the specified field numbers are `10` and `11`.

cc @fryorcraken @danisharora099


vpavlin (approved):
LGTM

Ivansete-status (approved):
Shall we have comments in the code to pin point where these numbers come from?

alrevuelta (approved):

PR #1717, opened 2023-05-05
fix(rln-relay): handle invalid deletes

rymnc:
# Description

After deletion behaviour in zerokit's master was changed, it caused ci errors in nwaku. This PR addresses that change in behaviour. 

# Changes

- [x] Updated tests to include a valid deletion operation


alrevuelta (approved):
thanks for the quick fix! lgtm!

PR #1720, opened 2023-05-05
feat: example using filter and lightpush

jm-clius:
While working on testing the new filter protocol, I realised that it will have some value to package a lightpush publisher and filter subscriber as an example.

I've tried to keep to the spirit of the current subscriber-publisher example, but without adding to our current MAKE targets.

### DIFF
--- a/examples/v2/filter_subscriber.nim
+++ b/examples/v2/filter_subscriber.nim
@@ -49,7 +49,7 @@ proc maintainSubscription(wfc: WakuFilterClient,
 
       let subscribeRes = await wfc.subscribe(filterPeer, filterPubsubTopic, @[filterContentTopic])
 
-      if subscribeRes.isErr:
+      if subscribeRes.isErr():
         notice "subscribe request failed. Quitting.", err=subscribeRes.error
         break
       else:
--- a/examples/v2/filter_subscriber.nim
+++ b/examples/v2/filter_subscriber.nim
@@ -43,7 +43,7 @@ proc maintainSubscription(wfc: WakuFilterClient,
     notice "maintaining subscription"
     # First use filter-ping to check if we have an active subscription
     let pingRes = await wfc.ping(filterPeer)
-    if pingRes.isErr:
+    if pingRes.isErr():
       # No subscription found. Let's subscribe.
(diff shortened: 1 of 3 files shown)


examples/v2/filter_subscriber.nim
```diff
@@ -0,0 +1,82 @@
...
+
+proc maintainSubscription(wfc: WakuFilterClient,
+                          filterPeer: RemotePeerInfo,
+                          filterPubsubTopic: PubsubTopic,
+                          filterContentTopic: ContentTopic) {.async.} =
+  while true:
+    notice "maintaining subscription"
+    # First use filter-ping to check if we have an active subscription
+    let pingRes = await wfc.ping(filterPeer)
+    if pingRes.isErr:
```
Ivansete-status, line 46: Tiny detail: shall we use () for verbs and non-() for nouns?
```suggestion
    if pingRes.isErr():
```

Ivansete-status (approved):
It does look great! Thanks for this PR!

examples/v2/filter_subscriber.nim
```diff
@@ -0,0 +1,82 @@
...
+    switch = newStandardSwitch()
+    pm = PeerManager.new(switch)
+    wfc = WakuFilterClient.new(rng, messagePushHandler, pm)
+
+  # Mount filter client protocol
+  switch.mount(wfc)
+
+  # Start maintaining subscription
+  asyncSpawn maintainSubscription(wfc, filterPeer, FilterPubsubTopic, FilterContentTopic)
+
```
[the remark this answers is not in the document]
jm-clius: So this is to dispatch a future without awaiting its result. However, if you simply discard the future it may contain errors which gets "swallowed". For that reason, you should always use `asyncSpawn`, which will convert any error in the future to a `Defect`. See https://github.com/status-im/nim-chronos/blob/8563c936733a28bc1e773542e9741b5b55dfcaeb/chronos/asyncloop.nim#L130-L137

examples/v2/filter_subscriber.nim
jm-clius: Good point. Thanks!

PR #1723, opened 2023-05-10
chore: use address instead of public key for dos protection

alrevuelta:
* Adapts to changes introduced in: https://github.com/vacp2p/rfc/pull/601
* With this change we now use the address (shorter and easier to read) rather than the public key to perform validation.
* This allows to "encode" the address into the pubsub topic, so if you know the topic you have all the info you need to perform validation on it. No need to share the public key by other means.
* Signed topics follow this structrure: `/waku/2/signed:0x2ea1f2ec2da14e0e3118e6c5bdfa0631785c76cd/proto`
* Meta size is changes from 64 bytes to 65 bytes, so that a recoverable signature can be stored.

Example:
```
./build/wakunode2 \
--topics="/waku/2/signed:0x2ea1f2ec2da14e0e3118e6c5bdfa0631785c76cd/proto" \
--protected-topic=/waku/2/signed:0x2ea1f2ec2da14e0e3118e6c5bdfa0631785c76cd/proto
```

### DIFF
--- a/apps/wakunode2/wakunode2_validator_signed.nim
+++ b/apps/wakunode2/wakunode2_validator_signed.nim
@@ -4,17 +4,19 @@ else:
   {.push raises: [].}
 
 import
-  std/math,
+  std/[math, strutils],
   chronicles,
   chronos,
   metrics,
   stew/byteutils,
   stew/endians2,
+  stew/results,
   libp2p/protocols/pubsub/gossipsub,
   libp2p/protocols/pubsub/rpc/messages,
   libp2p/protocols/pubsub/errors,
   nimcrypto/sha2,
-  secp256k1
+  nimcrypto/keccak,
+  eth/keys
 
 const MessageWindowInSec = 5*60 # +- 5 minutes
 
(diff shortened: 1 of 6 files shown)


apps/wakunode2/wakunode2_validator_signed.nim
```diff
@@ -49,23 +51,53 @@ proc withinTimeWindow*(msg: WakuMessage): bool =
...
+    return err("invalid address, expected 0x prefix and size of 42 chars")
+
+  return ok(address)
+
+proc addSignedTopicValidator*(w: WakuRelay, protectedTopic: PubsubTopic) =
+
+  # address (hashed pubkey) to validate the message is encoded in the pubsub topic
+  let address0x = getKeyFromTopic(protectedTopic)
+  if address0x.isErr():
+    raise newException(Defect, address0x.error)
```
Ivansete-status, line 81: Would it be possible to avoid raising an exception and make the `proc` to return a Result[] ?

```suggestion
  let address0x = ? getKeyFromTopic(protectedTopic)
```

apps/wakunode2/wakunode2_validator_signed.nim
```diff
@@ -49,23 +51,53 @@ proc withinTimeWindow*(msg: WakuMessage): bool =
...
-          if recoveredSignature.isOk():
-            if recoveredSignature.get.verify(msgHash, publicTopicKey):
-              outcome = errors.ValidationResult.Accept
+          let signature = Signature.fromRaw(msg.get.meta)
+          if signature.isOk():
+            let msgHash = protectedTopic.msgHash(msg.get)
+            let recoveredPublic = signature.get.recover(msgHash)
+            if recoveredPublic.isOk():
+              if recoveredPublic.get.toAddress() == address0x.get:
+                outcome = errors.ValidationResult.Accept
```
Ivansete-status, line 98: it seems as if we are only considering the "positive" case but should we consider the error cases and return them appropriately? Just to learn about it :)

Ivansete-status (approved):
The PR looks great! Congrats for it!
If possible I'd try to avoid raising exceptions :)

apps/wakunode2/wakunode2_validator_signed.nim
```diff
@@ -49,23 +51,53 @@ proc withinTimeWindow*(msg: WakuMessage): bool =
...
+  let topicName = parts[3]
+  let topicParts = topicName.split(":")
+  if topicParts.len != 2:
+    return err("invalid topic format for a signed protected topic, : missing")
+
+  if topicTag notin topicName:
+    return err("invalid topic format expected signed: tag")
+
+  let address = topicParts[1]
+  if not address.startsWith("0x") or address.len != 42:
```
richard-ramos, line 71: other criteria to validate: characters from idx 2 to 41 are hex characters

apps/wakunode2/wakunode2_validator_signed.nim
```diff
@@ -49,23 +51,53 @@ proc withinTimeWindow*(msg: WakuMessage): bool =
...
+    return err("invalid address, expected 0x prefix and size of 42 chars")
+
+  return ok(address)
```
richard-ramos, line 74:
```suggestion
  return ok(address.toLowerAscii)
```

apps/wakunode2/wakunode2_validator_signed.nim
```diff
@@ -49,23 +51,53 @@ proc withinTimeWindow*(msg: WakuMessage): bool =
...
-          let recoveredSignature = SkSignature.fromRaw(msg.get.meta)
-          if recoveredSignature.isOk():
-            if recoveredSignature.get.verify(msgHash, publicTopicKey):
-              outcome = errors.ValidationResult.Accept
+          let signature = Signature.fromRaw(msg.get.meta)
+          if signature.isOk():
+            let msgHash = protectedTopic.msgHash(msg.get)
+            let recoveredPublic = signature.get.recover(msgHash)
+            if recoveredPublic.isOk():
+              if recoveredPublic.get.toAddress() == address0x.get:
```
richard-ramos, line 97:
```suggestion
              if recoveredPublic.get.toAddress().toLowerAscii == address0x.get:
```

apps/wakunode2/wakunode2_validator_signed.nim
richard-ramos: Maybe it makes sense to return an error if the address is not in lowercase at all?

apps/wakunode2/wakunode2_validator_signed.nim
```diff
@@ -49,23 +51,53 @@ proc withinTimeWindow*(msg: WakuMessage): bool =
...
   return false
 
-proc addSignedTopicValidator*(w: WakuRelay, topic: PubsubTopic, publicTopicKey: SkPublicKey) =
-  debug "adding validator to signed topic", topic=topic, publicTopicKey=publicTopicKey
+proc getKeyFromTopic*(protectedTopic: PubsubTopic): Result[string, string] =
+  # example: /waku/2/signed:0x2ea1f2ec2da14e0e3118e6c5bdfa0631785c76cd/proto
+  # see https://rfc.vac.dev/spec/57/#dos-protection
+  let topicTag = "signed"
+  let parts = protectedTopic.split("/")
+  if parts.len != 5:
```
jm-clius, line 59: I've added some comments on the [RFC PR](https://github.com/vacp2p/rfc/pull/601) that would make this an invalid check. For one, we've deprecated the use of the superfluous `/proto` on pubsub topic names. We also want people to still be able to define their own topic names for disambiguation too (what if multiple topics need to be signed with the same key?).  We also want to be able to define the static shard in the topic name for Status Communities.

alrevuelta:
Thanks all for the comments, however we wont be continuing with this feature. Main reason is what was already discussed here and in https://github.com/vacp2p/rfc/pull/601

PR #1725, opened 2023-05-11
fix(ci): remove target flag from docker command

rymnc:
# Description

Includes a debug flag in the jenkinsfile

# Changes

- [x] Adds the debug target, and replaces the experimental target since it doesn't exist anymore

### DIFF
--- a/ci/Jenkinsfile.release
+++ b/ci/Jenkinsfile.release
@@ -41,6 +41,11 @@ pipeline {
       description: 'Enable experimental features.',
       defaultValue: false
     )
+    booleanParam(
+      name: 'DEBUG',
+      description: 'Enable debug features (heaptrack).',
+      defaultValue: false
+    )
   }
 
   stages {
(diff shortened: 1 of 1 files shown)


LNSD:
@rymnc Note that I suggested having a "debug tools" target in the Dockerfile (e.g., for an image including the `heaptrack` tool), and that would require it.

cc @Ivansete-status

rymnc:
@LNSD, thanks for the catch - 
I've added an additional flag in https://github.com/waku-org/nwaku/pull/1725/commits/7013911b077facde0a59c1bc41db69612548bddf for the `debug` target, so we can rollout the debug containers to all nodes in the test fleet at once if required then, if it is unset, will use the prod image.

Ivansete-status:
Thanks for the PR @rymnc ! However, I miss the DEBUG flag in our Makefile so that we can clearly have `release` by default and `debug` on demand.

Btw, we are generating the release binaries with `debug` information. We have the next for the `statusteam/nim-waku:v0.16.0` package:
```
/ # file /usr/local/bin/wakunode2 
/usr/local/bin/wakunode2: ELF 64-bit LSB pie executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib/ld-musl-x86_64.so.1, with debug_info, not stripped
```

Was that made on purpose @jm-clius ?

jm-clius:
> Btw, we are generating the release binaries with debug information. 

Mmm. There may be something here, though I think the default build will be with `-d:release` as it should. Release builds _should_ contain debugger info, just with the lighter weight libbacktrace stack tracing: https://github.com/status-im/nim-libbacktrace

vpavlin (approved):

PR #1727, opened 2023-05-12
fix: wakunode2 config. adding new 'topic' config parameter.

### LINKED ISSUE #1726: bug: inconsistent usage of topic flags
### Problem
There's two ways to use topics flags:

* `--topics` - Takes a list of comma separated topics, CANNOT be repeated.
* `--protected-topic` - Does NOT take a list, CAN be repeate.

Makes no sense.

### Impact
Almost made me vomit.

### Expected behavior
Not inducing nauseousness. Probably just make `--topics` into `--topic` and allow it to be repeated.

### Additional context
https://github.com/waku-org/nwaku/blob/b277ce1013563b4e76a3f8694f0eefb1f3528a4d/apps/wakunode2/config.nim#L210-L213
https://github.com/waku-org/nwaku/blob/b277ce1013563b4e76a3f8694f0eefb1f3528a4d/apps/wakunode2/config.nim#L40-L43

Ivansete-status:
# Description
Adding a new `topic` parameter can be repeated.
The `topics` parameter becomes deprecated because it is not consistent with the other "repeatable" ones as it expects a space-separated list of pubsub topics.

## How to use the new parameter
When setting different pubsub topics through the CLI, they can be repeated:
`./build/wakunode2 --topic=t1 --topic=t2 --topic=t3 ...`

However, when setting them from within a config file, they should be in an array format within the cfg file:
```
topic = [ "t1" "t2" "t3" ]
```

## Issue
closes #1726

### DIFF
--- a/apps/wakunode2/app.nim
+++ b/apps/wakunode2/app.nim
@@ -549,11 +549,12 @@ proc setupProtocols(node: WakuNode, conf: WakuNodeConf,
   if conf.relay:
 
     var pubsubTopics = @[""]
-    if conf.topicsDeprecated != "":
+    if conf.topicsDeprecated != "/waku/2/default-waku/proto":
+      warn "The 'topics' parameter is deprecated. Better use the 'topic' one instead."
       if conf.topics != @["/waku/2/default-waku/proto"]:
         return err("Please don't specify 'topics' and 'topic' simultaneously. Only use the 'topic' parameter")
 
-      # This clause (if conf.topicsDeprecated != "") should disapear in >= v0.18.0
+      # This clause (if conf.topicsDeprecated ) should disapear in >= v0.18.0
       pubsubTopics = conf.topicsDeprecated.split(" ")
     else:
       pubsubTopics = conf.topics
--- a/apps/wakunode2/app.nim
+++ b/apps/wakunode2/app.nim
@@ -547,7 +547,16 @@ proc setupProtocols(node: WakuNode, conf: WakuNodeConf,
     peerExchangeHandler = some(handlePeerExchange)
 
   if conf.relay:
-    let pubsubTopics = conf.topics.split(" ")
+
+    var pubsubTopics = @[""]
+    if conf.topicsDeprecated != "":
+      if conf.topics != @["/waku/2/default-waku/proto"]:
+        return err("Please don't specify 'topics' and 'topic' simultaneously. Only use the 'topic' parameter")
+
+      # This clause (if conf.topicsDeprecated != "") should disapear in >= v0.18.0
+      pubsubTopics = conf.topicsDeprecated.split(" ")
+    else:
+      pubsubTopics = conf.topics
     try:
       await mountRelay(node, pubsubTopics, peerExchangeHandler = peerExchangeHandler)
     except CatchableError:
--- a/apps/wakunode2/config.nim
+++ b/apps/wakunode2/config.nim
@@ -42,7 +42,6 @@ type
       defaultValue: newSeq[ProtectedTopic](0)
       name: "protected-topic" .}: seq[ProtectedTopic]
 
-
     ## Log configuration
     logLevel* {.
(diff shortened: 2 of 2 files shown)


apps/wakunode2/app.nim
```diff
@@ -547,7 +547,16 @@ proc setupProtocols(node: WakuNode, conf: WakuNodeConf,
     peerExchangeHandler = some(handlePeerExchange)
 
   if conf.relay:
-    let pubsubTopics = conf.topics.split(" ")
+
+    var pubsubTopics = @[""]
+    if conf.topicsDeprecated != "":
```
vpavlin, line 552: How about logging a warning if `conf.topicsDeprecated != ""` ?

vpavlin (approved):

apps/wakunode2/config.nim
```diff
@@ -207,11 +205,16 @@ type
       defaultValue: false
       name: "keep-alive" }: bool
 
-    topics* {.
-      desc: "Default topics to subscribe to (space separated list)."
-      defaultValue: "/waku/2/default-waku/proto"
+    topicsDeprecated* {.
+      desc: "Default topics to subscribe to (space separated list). DEPRECATED: please use repeated --topic argument instead."
+      defaultValue: ""
```
alrevuelta, line 210: should we keep the `/waku/2/default-waku/proto` here? otherwise old deployments using "topics" will break?

jm-clius (approved):
Thanks! After this is merged the next step is to change this across infra repos, make sure we include this in the release notes as an upgrade instruction and then remove the deprecated option.

alrevuelta (changes requested):
thanks for this, left a comment which i think its important, otherwise we will be introducing breaking changes.

and out of curiosity, why from file its like this? i find it confusing that the behaviour is different. cant we have the same? topic=a, topic=b?
`topic = [ "t1" "t2" "t3" ]`

mind having a look on how `staticnode(s)` is implemented?

apps/wakunode2/app.nim
Ivansete-status: Good point! I'll add it

alrevuelta: agree.

```
if conf.topicsDeprecated != "" and conf.topics.len !=0.
  error

if conf.topicsDeprecated != "":
  warn
```

apps/wakunode2/config.nim
jm-clius: Oh! Important point that I missed. @alrevuelta is right here. I think we should keep the default in both config options.

Ivansete-status: It shouldn't break because if nothing is specified, the default pubsub topic will get used.

Ivansete-status:
> thanks for this, left a comment which i think its important, otherwise we will be introducing breaking changes.
> 
> and out of curiosity, why from file its like this? i find it confusing that the behaviour is different. cant we have the same? topic=a, topic=b? `topic = [ "t1" "t2" "t3" ]`
> 
> mind having a look on how `staticnode(s)` is implemented?

This behavior comes from 'confutils'. It behaves equally for any other repeatable parameter, i.e. `staticnodes`.

apps/wakunode2/config.nim
alrevuelta: you are right:
https://github.com/waku-org/nwaku/blob/493f082975e7b1bef231acf0299453d645388b8e/apps/wakunode2/app.nim#L559

but perhaps a bit convoluted? can see you applied the changes thanks!

alrevuelta (approved):
now lgtm thanks!

apps/wakunode2/app.nim
Ivansete-status: Now, if `topicsDeprecated` is given, it prints a `warn`. If both `topicsDeprecated` and `topics` are given, then an error is returned and the execution is halted.

PR #1731, opened 2023-05-15
refactor(ci): cache all of submodules/deps to speed up build time

vpavlin:
# Description
I've been experimenting with how we use cache inour CI builds - we only cache the nim compiler binary, which leads to potentially unnecessary rebuilds of some libs and recursive submodule initialization on every build.

I believe I was able to get `vendor/` submodules caching properly, so that we can reuse all of the content without submodule init and rebuilds of deps.

I've experimented with this in a fork of nwaku repo (https://github.com/nwaku-test-org/nwaku/blob/master/.github/workflows/test-cache.yml) and the change seemed to result in change of build time from 21-25 mins down to 13-17 mins.

Potential drawback is that previously when one submodule changed we'd still use cached nim compiler, but here it will trigger a compiler rebuild on the PR which is updating a dependency - hence longer builds on PR changing the deps, but those seem to be relatively rare, so hopefully the benefits outweight that.

# Changes

- [ ] all GH actions jobs now use cached `vendor/`
- [ ] removed explicit caching of zerokit artifacts as they fall under the `vendor/`

### DIFF
--- a/.github/workflows/ci-experimental.yml
+++ b/.github/workflows/ci-experimental.yml
@@ -61,27 +61,19 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v3
-        with:
-          submodules: recursive
 
-      - name: Get submodule versions
+      - name: Get submodules hash
         id: submodules
-        shell: bash
         run: |
-          echo "nim-hash=$(git rev-parse HEAD:vendor/nimbus-build-system)" >> $GITHUB_OUTPUT
-          echo "zerokit-hash=$(git rev-parse HEAD:vendor/zerokit)" >> $GITHUB_OUTPUT
+          echo "hash=$(git submodule status | sha256sum | sed 's/[ -]*//g')" >> $GITHUB_OUTPUT
 
-      - name: Cache nim compiler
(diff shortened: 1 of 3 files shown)


Ivansete-status (approved):
I think it looks great! Thanks for it :)

.github/workflows/ci-experimental.yml
```diff
@@ -61,27 +61,19 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v3
-        with:
-          submodules: recursive
 
-      - name: Get submodule versions
+      - name: Get Submodules Hash
```
jm-clius, line 65: Pure nitpick from my side, but can we get these names to follow the same capitalisation (i.e. only first letter capitalised)?

```suggestion
      - name: Get submodules hash
```

Here and elsewhere.

jm-clius (approved):
Thanks! See my nitpick below. :)

.github/workflows/ci-experimental.yml
vpavlin: Pardon my morning ignorance. Fixed

PR #1739, opened 2023-05-17
chore(networking): set and use target outbound connections + prune

alrevuelta:
# Description
* Before this PR nwaku nodes constantly tried to create new outcoming relay connections up to max connections. In theory, this could cause a node to only have out connections, which is a selfish behaviour since this takes a lot of incoming slots (of other peers) in the network.
* This PR enforces a given in/out ratio of connections, leaving more incoming slots than outgoing, which is more altruistic with the network, something quite important in the fleets. It archieves so by:
  * prunning exceding inbound connections
  * not attempting mroe out connections if the target was reach. 

closes https://github.com/waku-org/nwaku/issues/1495

### DIFF
--- a/waku/v2/node/peer_manager/peer_manager.nim
+++ b/waku/v2/node/peer_manager/peer_manager.nim
@@ -57,6 +57,9 @@ const
   # How often the peer store is updated with metrics
   UpdateMetricsInterval = chronos.seconds(15)
 
+  # How often to log peer manager metrics
+  LogSummaryInterval = chronos.seconds(60)
+
 type
   PeerManager* = ref object of RootObj
     switch*: Switch
@@ -66,6 +69,7 @@ type
     maxFailedAttempts*: int
     storage: PeerStorage
     serviceSlots*: Table[string, RemotePeerInfo]
+    outPeersTarget*: int
     started: bool
 
 proc protocolMatcher*(codec: string): Matcher =
@@ -333,6 +337,7 @@ proc new*(T: type PeerManager,
                        storage: storage,
                        initialBackoffInSec: initialBackoffInSec,
                        backoffFactor: backoffFactor,
+                       outPeersTarget: max(maxConnections div 10, 10),
                        maxFailedAttempts: maxFailedAttempts)
 
   proc connHook(peerId: PeerID, event: ConnEvent): Future[void] {.gcsafe.} =
@@ -477,22 +482,22 @@ proc connectToNodes*(pm: PeerManager,
   # later.
   await sleepAsync(chronos.seconds(5))
 
-# Returns the amount of physical connections (in and out)
+# Returns the peerIds of physical connections (in and out)
 # containing at least one stream with the given protocol.
-proc getNumConnections*(pm: PeerManager, protocol: string): (int, int) =
-  var
-    numConnsIn = 0
-    numConnsOut = 0
+proc connectedPeers*(pm: PeerManager, protocol: string): (seq[PeerId], seq[PeerId]) =
+  var inPeers: seq[PeerId]
+  var outPeers: seq[PeerId]
+
   for peerId, muxers in pm.switch.connManager.getConnections():
     for peerConn in muxers:
       let streams = peerConn.getStreams()
       if streams.anyIt(it.protocol == protocol):
         if peerConn.connection.transportDir == Direction.In:
-          numConnsIn += 1
+          inPeers.add(peerId)
         elif peerConn.connection.transportDir == Direction.Out:
-          numConnsOut += 1
+          outPeers.add(peerId)
 
-  return (numConnsIn, numConnsOut)
+  return (inPeers, outPeers)
 
 proc getNumStreams*(pm: PeerManager, protocol: string): (int, int) =
   var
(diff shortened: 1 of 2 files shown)


waku/v2/node/peer_manager/peer_manager.nim
```diff
@@ -508,28 +513,33 @@ proc getNumStreams*(pm: PeerManager, protocol: string): (int, int) =
...
   return (numStreamsIn, numStreamsOut)
 
+proc pruneInRelayConns(pm: PeerManager, amount: int) {.async.} =
+  let (inRelayPeers, outRelayPeers) = pm.connectedPeers(WakuRelayCodec)
+  let connsToPrune = min(amount, inRelayPeers.len)
+
+  for p in inRelayPeers[0..<connsToPrune]:
+    await pm.switch.disconnect(p)
+
 proc connectToRelayPeers*(pm: PeerManager) {.async.} =
```
Ivansete-status, line 523: I'd rename this proc to something like `updateRelayPeersConnections`

waku/v2/node/peer_manager/peer_manager.nim
```diff
@@ -508,28 +513,33 @@ proc getNumStreams*(pm: PeerManager, protocol: string): (int, int) =
...
-  let (inRelayPeers, outRelayPeers) = pm.getNumConnections(WakuRelayCodec)
-  let totalRelayPeers = inRelayPeers + outRelayPeers
+  let totalRelayPeers = inRelayPeers.len + outRelayPeers.len
+  let inPeersTarget = maxConnections - pm.outPeersTarget
+
+  if inRelayPeers.len > inPeersTarget:
+    await pm.pruneInRelayConns(inRelayPeers.len-inPeersTarget)
+
+  if outRelayPeers.len >= pm.outPeersTarget:
+    return
```
Ivansete-status, line 533: Shall we also prune the exceeding `out` connection at this point?

waku/v2/node/peer_manager/peer_manager.nim
```diff
@@ -607,13 +617,30 @@ proc relayConnectivityLoop*(pm: PeerManager) {.async.} =
     await pm.connectToRelayPeers()
     await sleepAsync(ConnectivityLoopInterval)
 
+proc logSummary*(pm: PeerManager) {.async.} =
+  heartbeat "Log peer manager summary", LogSummaryInterval:
+    let (inRelayPeers, outRelayPeers) = pm.connectedPeers(WakuRelayCodec)
+    let maxConnections = pm.switch.connManager.inSema.size
```
Ivansete-status, line 623: I wonder if it may be also interesting to print the in/out ratio, for debugging purposes.

waku/v2/node/peer_manager/peer_manager.nim
```diff
@@ -508,28 +513,33 @@ proc getNumStreams*(pm: PeerManager, protocol: string): (int, int) =
               numStreamsOut += 1
   return (numStreamsIn, numStreamsOut)
 
+proc pruneInRelayConns(pm: PeerManager, amount: int) {.async.} =
+  let (inRelayPeers, outRelayPeers) = pm.connectedPeers(WakuRelayCodec)
+  let connsToPrune = min(amount, inRelayPeers.len)
```
Ivansete-status, line 518:
```suggestion
proc pruneInRelayConns(pm: PeerManager, connsToPrune: int) {.async.} =
```
As per how this proc is used, the `min(amount, inRelayPeers.len)` will always return `amount`.

Ivansete-status (commented):
Thanks for the PR! Really interesting!
I added a few suggestions that I hope they help a little :)

jm-clius (approved):
LGTM. I have one comment on the selected out:in ratio, but non-blocking as I don't have a better suggestion. May have to come from experiments.

waku/v2/node/peer_manager/peer_manager.nim
alrevuelta: good point, but the thing is that since we control the out connections that we create, that shouldnt be necessary. i mean, we should never exceed the max out limit, since we do this check at `connectToRelayPeers`.

note that on top of this in prunning, we will have a prunning based on libp2p scoring, which will prune in/out connections from our mesh based on a given score. But this comes from gossipsub.

waku/v2/node/peer_manager/peer_manager.nim
alrevuelta: its there right? or wdym? we have `inPeersTarget` and `outPeersTarget`

```
inRelayConns = $inRelayPeers.len & "/" & $inPeersTarget,
outRelayConns = $outRelayPeers.len & "/" & $pm.outPeersTarget,
```

waku/v2/node/peer_manager/peer_manager.nim
alrevuelta: yup, as its used yep but you never know hoy a function will be used. good sanity check to avoid accessing out of bound elements.

waku/v2/node/peer_manager/peer_manager.nim
```diff
@@ -333,6 +337,7 @@ proc new*(T: type PeerManager,
                        storage: storage,
                        initialBackoffInSec: initialBackoffInSec,
                        backoffFactor: backoffFactor,
+                       outPeersTarget: max(maxConnections div 10, 10),
```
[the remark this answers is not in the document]
alrevuelta: yup, the eclipse attack was also my concern. took this ratio from both nimbus (1:9) and prysm (1:8) but we can revisit later on.
* [nimbus](https://github.com/status-im/nimbus-eth2/blob/40253a76ddcfd2aa29ef662bb8ee61306093fd1c/beacon_chain/networking/eth2_network.nim#L1559)
* [prysm](https://github.com/prysmaticlabs/prysm/blob/537236e1c9cbecf863a391a7b5f097313b8043c0/beacon-chain/p2p/peers/status.go#LL68C2-L68C14)

unsure how they justify this number given the possibility of eclipse attacks. Perhaps from a practical point of view its more complex since the "eclipser" is compeeting with the whole network to get these slots, so it would be very difficult for the attacker to end up having all slots?

waku/v2/node/peer_manager/peer_manager.nim
Ivansete-status: Morning! I mean to print the ratio that we expect. This is the real ratio that is being applied. I would print the theoretical ratio just in case it is not accomplished by the real one, for any reason.
Not a big deal though. If you consider we have enough info, then go ahead :)

alrevuelta: I see. if you dont mind i leave it as it is, since whats being printed here its more complete imho. eg: "...5/10, 30/50" which indicates that 5 `in` slots out of 10 are taken and 30 out of 50 `out` are taken.

Ivansete-status: Ok cool, is perfect then! thanks !

waku/v2/node/peer_manager/peer_manager.nim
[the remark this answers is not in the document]
vpavlin: Probably unrelated to the eclipse attack, but would it make sense to make to ratio configurable?

waku/v2/node/peer_manager/peer_manager.nim
Ivansete-status: Cool, makes sense!

Ivansete-status (approved):

waku/v2/node/peer_manager/peer_manager.nim
alrevuelta: wops i missed this. isn't `connectToRelayPeers` shorter and more descriptive?

waku/v2/node/peer_manager/peer_manager.nim
vpavlin: Would it make sense to pass list of peerIds to prune here though? Instead of `amount` - that way we could avoid counting/getting the peers again (you already do it before calling this function) and it would allow for potentially "smarter" ways of choosing the peers to prune - this just goes from first peer, but by passing an array of peerIds to prune the caller could come up with a different way how to choose the peers to prune

waku/v2/node/peer_manager/peer_manager.nim
Ivansete-status: Hey @alrevuelta ! It is shorter but this proc not only `connectToNodes` but also `pruneInRelayConns`. This is why I see it more convenient to have the "update" word on its name. Maybe sth like `updateRelayPeers` can suit better, which is even shorter ;P

PR #1740, opened 2023-05-17
refactor: proper use of setupNat

Ivansete-status:
# Description

This PR is aimed to give a better structure to the `setupNat` proc usage, plus update the reference to `nim-eth` submodule.
The update of `nim-eth` submodule is important to break the `confutils` dependency, and therefore, allow the generation of the waku dynamic library, as intended in: https://github.com/waku-org/nwaku/pull/1730

Notice that I had to use `rlpx_connected_peers` instead of `connected_peers` in `wakunode1.nim` due to the update of the `vendor/nim-eth`, which not only adds the dependency-break with 'confutils' but also includes another changes.


jm-clius (approved):
LGTM! :)

apps/chat2bridge/chat2bridge.nim
```diff
@@ -248,11 +248,15 @@ when isMainModule:
   if conf.logLevel != LogLevel.NONE:
     setLogLevel(conf.logLevel)
 
+  let natRes = setupNat(conf.nat, clientId,
+                        Port(uint16(conf.libp2pTcpPort) + conf.portsShift),
+                        Port(uint16(conf.udpPort) + conf.portsShift))
+  if natRes.isErr():
+    error "Error in setupNat", error = natRes.error
+
```
vpavlin, line 256: You are calling `quit()` in the `wakubridge`, but not here - should the reaction be consistent accross apps?

tests/v1/test_waku_connect.nim
```diff
@@ -58,9 +58,9 @@ procSuite "Waku connections":
       p2 = await n2.rlpxConnect(newNode(n3.toENode()))
       p3 = await n4.rlpxConnect(newNode(n3.toENode()))
     check:
-      p1.isNil
-      p2.isNil == false
-      p3.isNil == false
+      p1.isErr()
```
vpavlin, line 61: In other cases it is xplicitly checking against `true` - should we be consistent and do it here as well?

vpavlin (approved):
Few comments, but in general LGTM

apps/chat2bridge/chat2bridge.nim
Ivansete-status: I agree indeed. However, this is consistent within the chat2bridge.nim itself. I think it is better if we raise a separate PR for that.

Ivansete-status: @vpavlin - Next week I will create a separate PR for adding consistency in the parameters checks for all the apps.

PR #1742, opened 2023-05-18
fix(ci): make experimental default to true in fleet deployment

rymnc:
# Description

During the period of rln-relay testnet 3, we will make the experimental flag default to true for all fleet deployments.

# Changes

- [x] Updated `Jenkinsfile.release` to make `EXPERIMENTAL` default
- [ ] ...

### DIFF
--- a/ci/Jenkinsfile.release
+++ b/ci/Jenkinsfile.release
@@ -39,7 +39,7 @@ pipeline {
     booleanParam(
       name: 'EXPERIMENTAL',
       description: 'Enable experimental features.',
-      defaultValue: false
+      defaultValue: true
     )
     booleanParam(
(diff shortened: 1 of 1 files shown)


jm-clius:
Why not simply change it on the build parameters for the deployments where it's necessary? I think this configuration is transferred to the next deployments? https://ci.infra.status.im/job/nim-waku/job/deploy-wakuv2-test/

rymnc:
it isn't transferred to the next deployments afaik, and any merge to master overwrites the flag :/

jm-clius (approved):
Ok, happy to approve if this is not easily configurable from Jenkins. Good idea to revert back after the testnet and to bear this in mind when we make non-test deployments.

PR #1746, opened 2023-05-19
chore: close px connections after resp

alrevuelta:
* Since peer exchange protocol is req/resp and after exchanging the req/resp it's not very likely that the stream will be reused, we shall close the stream.
* Note that `.closeWithEOF()` is used instead of `.close()` [see](https://github.com/status-im/nim-libp2p/blob/unstable/libp2p/stream/lpstream.nim#L327-L335).
* A test is added that will fail without the changes introduced in this PR.
* Note that the stream is closed but not the connection.

### DIFF
--- a/tests/v2/test_waku_peer_exchange.nim
+++ b/tests/v2/test_waku_peer_exchange.nim
@@ -1,7 +1,7 @@
 {.used.}
 
 import
-  std/[options, sequtils],
+  std/[options, sequtils, tables],
   testutils/unittests,
   chronos,
   chronicles,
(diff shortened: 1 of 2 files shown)


jm-clius (approved):
Thanks! And good catch. Should we do the same for other req/resp protocols?

alrevuelta:
> Should we do the same for other req/resp protocols?

@jm-clius I would lightpush for sure and perhaps also store?

But unsure given:

> I imagine the expected behaviour would be to close streams for "pure" req/resp protocols after every resp and to keep it open if there's a high likelihood for follow-up requests (e.g. filter push, subsequent store queries, etc.)

PR #1747, opened 2023-05-19
fix: fix flaky test for dos topic

alrevuelta:
# Description
* Fix flaky test for dos topic. Too many messages were being sent, and in some cases the `check` was done before all the messages were gossiped, hence failing.
* This is fixed by sending less messages, that should take less time to gossip. If it happens again, plan b is to increase the timeouts.
* Thanks @rymnc for the report.


rymnc (approved):
LGTM

PR #1748, opened 2023-05-19
chode(makefile): unify where chronicles_log_level is set

vpavlin:
# Description
There were multiple places/ways how `chronicles_log_level` was configured. This PR unifies it and sets default to TRACE via Makefile.

# Changes

- [ ] remove chnicles_log_level setting from waku.nimble
- [ ] set LOG_LEVEL=TRACE at the top of Makefile to leverage code in nimbus-build-system makefile
- [ ] add missing QUICK_AND_DIRTY_COMPILER=1 to Dockerfile to keep builds consistent


Ivansete-status:
That looks great! Thanks for it! However, I'd remove all the `-d:chronicles_log_level=TRACE` entries from `waku.nimble`.

vpavlin:
> I'd remove all the `-d:chronicles_log_level=TRACE` entries from `waku.nimble`.

I think/hope I did that - did I miss any?

jm-clius (approved):
Thanks for trying to bring some consistency to all this! Approving as this is definitely an improvement and simplification. However, I think we probably have to review the CI workflows that override the `LOG_LEVEL`. Perhaps we want the `build`-tasks to compile the default (`TRACE`) and then only set `LOG_LEVEL` to `DEBUG` for the unit tests? E.g. here: https://github.com/waku-org/nwaku/blob/master/.github/workflows/ci.yml#L88

vpavlin:
@jm-clius 100% agree - that is what I tried to suggest on Discord - let's build with default (TRACE) log level and only run tests with DEBUG (to not over-polute the logs).

Ivansete-status:
> > I'd remove all the `-d:chronicles_log_level=TRACE` entries from `waku.nimble`.
> 
> I think/hope I did that - did I miss any?

Yes you did it :) I meant to remove all the `-d:chronicles_log_level=...` from the waku.nimble so that everything is built with 'TRACE', but I don't know the impact on the other components, so let's better leave it as it for now.

vpavlin:
> waku.nimble so that everything is built with 'TRACE'

Ah, right, yeah, I thought about that and decided not to do it, since the option is otherwise mostly used for waku v1 targets, which I would rather not touch:) and then you have it in cbindings targets as ERROR, assuming for a reason

rymnc (approved):
LGTM

alrevuelta (approved):
thanks!

Ivansete-status (approved):
LGTM ! thanks !

rymnc (approved):
LGTM

PR #1753, opened 2023-05-23
fix(test): wait more for gossip

alrevuelta:
* Fix flaky test by adding more delay so that all messages are gossiped.


rymnc (approved):
LGTM

PR #1758, opened 2023-05-24
feat(ci): add nightly builds

vpavlin:
# Description
It might be useful to build releasable artifacts nightly and create/update a Github Release with included commits and artifacts.

This gives anyone a place to simply get and test latest binary artifacts and to us a place to experiment with how artifacts are produced and packaged.

The GH Actions job is triggered every night + can be triggered manually, it builds `wakunode2` and tools (`networkmonitor` and `wakucanary`)  and creates or updates the `nightly` tag and release.

See an example here: https://github.com/nwaku-test-org/nwaku/releases/tag/nightly

# Changes

- [x] added a CI job which for nightly builds

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -1,0 +1,97 @@
+name: Nightly Release
+
+on:
+  schedule:
+    - cron: 13 3 * * *
+  workflow_dispatch:
+
+env:
+  RELEASE_NAME: nightly
+
+  NPROC: 2
+  MAKEFLAGS: "-j${NPROC}"
+  NIMFLAGS: "--parallelBuild:${NPROC}"
+
+jobs:
+  build-and-publish:
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+
+      - name: Get submodules hash
+        id: submodules
+        run: |
+          echo "hash=$(git submodule status | awk '{print $1}' | sort | shasum -a 256 | sed 's/[ -]*//g')" >> $GITHUB_OUTPUT
+
+      - name: Cache submodules
+        uses: actions/cache@v3
+        with:
+          path: |
+            vendor/
+            .git/modules
+          key: ${{ runner.os }}-vendor-modules-${{ steps.submodules.outputs.hash }}
+
+
+      - name: build artifacts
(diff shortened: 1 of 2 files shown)


rymnc:
nice, we do this in zerokit too!

vpavlin:
@rymnc hmm, we should probably add an artifact (e.g. nwaku-experimental...) where we also build with `EXPERIMENTAL=true`. WDYT?

rymnc:
That's a great idea!

jm-clius (approved):
This is useful add! We may want to point people to this somewhere in our documentation (perhaps in the [operator manual](https://github.com/waku-org/nwaku/blob/master/docs/operators/overview.md)) and make sure that people are aware of this on the Discord channels, etc.

vpavlin:
> perhaps in the [operator manual](https://github.com/waku-org/nwaku/blob/master/docs/operators/overview.md)

added a note in the overview, once the first nightly is built, I'll announce it in nwaku channel

alrevuelta (approved):
lgtm thanks!

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,97 @@
...
+          gh release view ${RELEASE_NAME} &>/dev/null &&\
+            gh release delete -y ${RELEASE_NAME} --cleanup-tag &&\
+            git tag -d ${RELEASE_NAME}
+
+          docker run \
+              -t \
+              --rm \
+              -v ${PWD}:/opt/sv4git/repo:z \
+              -u $(id -u) \
+              docker.io/wakuorg/sv4git:latest \
```
Ivansete-status, line 87: Maybe sth to be considered in the future: I think it could be interesting to have separate `nightly` artifacts. One per each day the `master` had changes, instead of having only the latest one. That way we might be able to debug possible issues by comparing artifacts from different dates.


https://hub.docker.com/r/statusteam/nim-waku/tags?page=1

Ivansete-status (approved):
This PR is awesome, thanks for it!
I just leave a comment for a possible future consideration.

.github/workflows/nightly.yml
vpavlin: We can introduce a step where we upload the nightly artifact somewhere external (IPFS?), but GH releases are not a good place for that - if we do like `nightly-$(date)`, our release page would get cluttered really quickly.

Container registries are designed to keep and serve large number of versions/tags, but the release page on GH is not, IMO.

PR #1764, opened 2023-05-26
feat(postgresql): align Andrea's PR#1590 changes into master

Ivansete-status:
# Summary
This PR is aimed to align the great progress made by @cammellos in terms of PostgreSQL adoption, into our `master` branch. With this, we add a `postres` driver that allows to interact with a PostgreSQL database.

# Details
The important changes were introduced in the next PR.
https://github.com/waku-org/nwaku/pull/1590

Apart from the adaptation I add a tiny test aimed to explicitly test the asynchronicity, which should work once we add the Lorenzo's changes.

My idea was to align Andrea's branch first and then work on the PR#1590 but I rebased it from `origin/master` as ususal so I need to merge into `master`. Sorry for that Andrea.

# How to test it
1. Start a PostgreSQL instance by mean of the included docker compose file
   ```code
   sudo docker compose -f tests/docker-compose.yml up
   ```
2. Run the postgresql tests:
   ```code
   . env.sh
   ```
   ```code
   nim c -r -d:chronicles_log_level=WARN --verbosity=0 --hints=off --outdir=build ./tests/v2/waku_archive/test_driver_postgres.nim
   ```
# Pending tasks
- Extend the `postgresql` driver to cover all the "archive" features.
- Include @LNSD changes, https://github.com/waku-org/nwaku/pull/1631, so that we can make asynchronous calls to the postgres driver.
- Allow the postgres unit tests to run without a running local PostgreSQL database.
- Move the included docker compose file to a more appropriate location.

## Issue


waku/v2/waku_archive/driver/postgres_driver/postgres_driver.nim
```diff
@@ -0,0 +1,288 @@
...
+  return err("not implemented")
+
+method close*(s: PostgresDriver):
+              Future[ArchiveDriverResult[void]] {.async.} =
+  ## Close the database connection
+  s.connection.close()
+  return ok()
+
+method sleep*(s: PostgresDriver, seconds: int):
+              Future[ArchiveDriverResult[void]] {.base, async.} =
```
jm-clius, line 277: If this is not overridden/inherited anywhere else, it should not be annotated `base` and be a `proc` rather than a `method`.

jm-clius (approved):
Afaict this looks good. I imagine a tricky step will be integrating this with configuration and the rest of the node :)

tests/all_tests_v2.nim
```diff
@@ -19,6 +19,10 @@ import
   ./v2/waku_archive/test_retention_policy,
   ./v2/waku_archive/test_waku_archive
 
+# TODO: add the postgres test when we can mock the database.
+# The postgres tests now need a running postgresql database running locally.
+#  ./v2/waku_archive/test_driver_postgres,
```
vpavlin, line 24: If you wanted the tests enabled, you could just execute a `docker run` command to start postgres container and then clean it up.

GH Actions support docker, so it should work and most of people (us) running tests also have docker installed and working.

tests/all_tests_v2.nim
```diff
@@ -19,6 +19,10 @@ import
   ./v2/waku_archive/test_retention_policy,
   ./v2/waku_archive/test_waku_archive
 
+# TODO: add the postgres test when we can mock the database.
```
vpavlin, line 22: In case you wanted to have tests enabled in CI, you could just exec a `docker run` command from tests to setup DB prior execution the tests themselves - it should work fine in CI

waku/v2/waku_archive/driver/postgres_driver.nim
```diff
@@ -0,0 +1,8 @@
+when (NimMajor, NimMinor) < (1, 4):
+  {.push raises: [Defect].}
+else:
+  {.push raises: [].}
+
+import ./postgres_driver/postgres_driver
+
+export postgres_driver
```
vpavlin, line 8: missing new line:) Would you like to add it?

waku/v2/waku_archive/driver/postgres_driver/postgres_driver.nim
```diff
@@ -0,0 +1,288 @@
...
+  " payload VARCHAR," &
+  " version INTEGER NOT NULL," &
+  " timestamp BIGINT NOT NULL," &
+  " id VARCHAR NOT NULL," &
+  " storedAt BIGINT NOT NULL," &
+  " CONSTRAINT messageIndex PRIMARY KEY (storedAt, id, pubsubTopic)" &
+  ");"
+
+proc insertRow(): string =
+ """INSERT INTO messages (id, storedAt, contentTopic, payload, pubsubTopic,
```
alrevuelta, line 42: will this insert work if we try to insert a duplicated message? I mean, insert a message twice. Perhaps we need ON CONFLICT?

mind adding a testcase if its the case?

alrevuelta (approved):
lgtm! thanks! left a comment that might be important, let me know otherwise :)

waku/v2/waku_archive/driver/postgres_driver/postgres_driver.nim
```diff
@@ -0,0 +1,288 @@
...
+      return err("failed to reset database")
+  except DbError:
+    return err("failed to reset database")
+  ok()
+
+method init*(s: PostgresDriver): ArchiveDriverResult[void] {.base.} =
+  try:
+    let res = s.connection.tryExec(sql(createTableQuery()))
+    if not res:
+      return err("failed to migrate database")
```
vpavlin, line 85: would an error saying `failed to initialize` make more sense here? Or is this a prep for adding migrations in the future?

waku/v2/waku_archive/driver/postgres_driver/postgres_driver.nim
```diff
@@ -0,0 +1,288 @@
...
+    rows = s.connection.getAllRows(sql("""SELECT storedAt, contentTopic,
+                                  payload, pubsubTopic, version, timestamp,
+                                  id FROM messages ORDER BY storedAt ASC"""))
+  except DbError:
+    return err("failed to query rows")
+
+  for r in rows:
+    let rowRes = extractRow(r)
+    if rowRes.isErr():
+      return err("failed to extract row")
```
vpavlin, line 164: Will this also print/pass the actual error from `extractRow`?

waku/v2/waku_archive/driver/postgres_driver/postgres_driver.nim
```diff
@@ -0,0 +1,288 @@
...
+    let cstmt = "contentTopic IN (" & "?".repeat(contentTopic.len).join(",") & ")"
+    statements.add(cstmt)
+    for t in contentTopic:
+      args.add(t)
+
+  if pubsubTopic.isSome():
+    statements.add("pubsubTopic = ?")
+    args.add(pubsubTopic.get())
+
+  if cursor.isSome():
```
vpavlin, line 194: Can you please explain what this does?

vpavlin (approved):
Few questions/comments mainly to feed my curiosity:) Otherwise, LGTM!

waku/v2/waku_archive/driver/postgres_driver/postgres_driver.nim
Ivansete-status: Thanks for the comment @alrevuelta 
I think it should be added correctly as the primary key would be different, given different timestamps (nanosecconds precision)
The expected behavior is that the duplicate messages are properly added right ?

waku/v2/waku_archive/driver/postgres_driver/postgres_driver.nim
```diff
@@ -0,0 +1,288 @@
...
+export postgres_driver
+
+type PostgresDriver* = ref object of ArchiveDriver
+  connection: DbConn
+  preparedInsert: SqlPrepared
+
+proc dropTableQuery(): string =
+  "DROP TABLE messages"
+
+proc createTableQuery(): string =
```
rymnc, line 29: Curious if we can manage migrations in plain SQL, so that other implementations can use the same migrations (not a blocker, just want to know your thoughts)

Ivansete-status: Good point @rymnc ! I'll set that as a pending action item in the issue

cammellos (commented):
Code looks good to me, I won't approve since I have co-written it, thank you @Ivansete-status for taking care of it!!

waku/v2/waku_archive/driver/postgres_driver/postgres_driver.nim
Ivansete-status: In case of completely equal messages, the `put` call will return an error. I will add a little test with the next:
```code
putRes = await driver.put(DefaultPubsubTopic,
                              msg2, computeDigest(msg2), msg2.timestamp)
    ## Then
    require:
      not putRes.isOk()
```

waku/v2/waku_archive/driver/postgres_driver/postgres_driver.nim
Ivansete-status: Sure! The concept of "cursor" is part of the pagination feature when retrieving messages. 

The client can optionally provide with a "PagingInfo" in the store protocol request that will therefore split the whole list of messages in pages (if page size is smaller than the # of messages retrieved after applying the search conditions.) 

Therefore, if a store response represents a page, that response will include a cursor that can be used to retrieve the next page. For the last page, the cursor is empty in the response.

An example of this can be seen at:
https://github.com/waku-org/nwaku/blob/ffac77611cd244980fbb98e24627186fb5f54d63/tests/v2/wakunode_rest/test_rest_store.nim#L219

vpavlin: thanks!

tests/all_tests_v2.nim
Ivansete-status: Hey @vpavlin ! In the end I will tackle that in a separate PR because I need to tweak it in a way that the postgres tests are only run in a Linux platform. This is because the GH Actions only allow starting a docker container within Linux runners.
Thanks for the hint !

Ivansete-status:
Special kudos to @cammellos for submiting the original PR!
Btw Andrea, kindly close your original PR, if you consider it appropriate, as we included your changes in that one:
https://github.com/waku-org/nwaku/pull/1590

If you consider that we are missing any important change we can create a separate issue to tackle it.
Thanks!

PR #1765, opened 2023-05-29
feat(networking): prune peers from same ip beyond colocation limit

### LINKED ISSUE #1749: feat: blacklist peers sharing more than x IPs
### Problem
Each node has a maximum amount of `in`/`out` connections. If an attacker creates multiple peers under the same IP and connects to us, we can run out of `in` slots to serve other honest peers. This attack can be extended to not just one node but the whole network, and will end up leaving very few available `in` slots for honest peers to connect. This attack has a huge impact on service protocols like store, because the network (depending on its size) can run out of slots to serve this protocols.

There is no single solution to prevent this attack, but more like a combination of:
* 1) churn: rotating peers from time to time (not implemented)
* 2) scoring: keeping the peers with high score, and disconnecting from the ones with low. Trickier in service protocols since its altruistic nature makes it not possible to score the client (as its not giving anything in return) (not implemented)
* 3) limit ips: limit the amount of peers we can see for each ip (not implemented)
* 4) enforce in/out: leave some slots for `out` peers, so that the node can control some of the peers that is connected to (implemented)

### Suggested solution
This issue suggests to implement 3) as follows. It aims to mitigate the above explained attack by relying on the fact that sybil attacking with IPs is way harder than with peerIds (which is trivial). It suggests to limit the amount of peers that we see for each IP, so that if a given IP has multiple peers behind, we ignore them (exceeding a threshold):

Solution:
* Define a `collocationFactor` that limits the amount of peers we allow from each ip. Example: 5.
* Add a new table to track the amount of peers that we see behind each IP in peerstore. If we detect that ip `ip_1` is shared among `p1..5` peers then this table will store `ip1->5`.
* Use this metric in the peerstore so that:
  * If a new `p6` is discovered and has `ip1`, don't add it to the peerstore and skip it (since `collocationFactor` = 5). This protects our peerstore to be full of possible spammers.
  * Reject incomming connections from peers if we have > `collocationFactor` conns from the same IP (unsure if this can be done in nim-libp2p, in golibp2p there is `ConnectionGater` that allows to define a handle that is executed before completing the connection)
  * Dont attempt connections to peers if their IP is tracked and shared by > `collocationFactor`. This shouldn't happen if they are never added into the peerstore.

Note. This can ben enforced on two ways:
* Just for connections. So that we never connect to more than `collocationFactor` sharing the same IP.
* For the peerstore in general. So we dont even track/store peers that share the same up (just up to `collocationFactor`)

Inspiration:
* prysm [follows this approach](https://github.com/prysmaticlabs/prysm/blob/537236e1c9cbecf863a391a7b5f097313b8043c0/beacon-chain/p2p/peers/status.go#L785) mixes it with other [scores](https://github.com/prysmaticlabs/prysm/blob/537236e1c9cbecf863a391a7b5f097313b8043c0/beacon-chain/p2p/peers/status.go#L336) and enforces it [here](https://github.com/prysmaticlabs/prysm/blob/537236e1c9cbecf863a391a7b5f097313b8043c0/beacon-chain/p2p/connection_gater.go#L36) and [here](https://github.com/prysmaticlabs/prysm/blob/537236e1c9cbecf863a391a7b5f097313b8043c0/beacon-chain/sync/rpc_status.go#L49)

### Alternatives considered

Gossipsub scoring takes this into account. This parameter is given a weight and is part of a final score with other parameters. imho limiting the peers we see from each IP should be either ok or ban (disconnect). It should be possible to tweak the weights to simulate this behaviour, so that if the amount of peers behind a given ip is > x, then the score automatically drops bellow the threshold and a disconnection is triggered.

Main problem with gossipsub scoring, is that it only applies to gossipsub, and we need protection also for service protocols. Gossipsub scoring can be integrated later on, but I would considering adding this layer on top that restricts the peers per IP for all connections.

alrevuelta:
closes #1749

Description:
* Allows a maximum amount of `colocationLimit` connected peers from the same ip.
* Runs a periodic task to trigger disconnections from peers not meeting the criteria.

Note that this PR does not prevent connection attempts from peers violating the `colocationLimit`. These connections will be succesful but cleaned up later on.

### DIFF
--- a/waku/v2/node/peer_manager/peer_manager.nim
+++ b/waku/v2/node/peer_manager/peer_manager.nim
@@ -10,7 +10,8 @@ import
   chronicles,
   metrics,
   libp2p/multistream,
-  libp2p/muxers/muxer
+  libp2p/muxers/muxer,
+  libp2p/nameresolving/nameresolver
 import
   ../../../common/nimchronos,
   ../../waku_core,
@@ -57,8 +58,14 @@ const
   # How often the peer store is updated with metrics
   UpdateMetricsInterval = chronos.seconds(15)
 
+  # Prune by ip interval
+  PruneByIpInterval = chronos.seconds(30)
+
   # How often to log peer manager metrics
   LogSummaryInterval = chronos.seconds(60)
+
+  # Max peers that we allow from the same IP
+  ColocationLimit = 5
 
 type
   PeerManager* = ref object of RootObj
@@ -70,6 +77,8 @@ type
     storage: PeerStorage
     serviceSlots*: Table[string, RemotePeerInfo]
     outPeersTarget*: int
+    ipTable*: Table[string, seq[PeerId]]
+    colocationLimit*: int
     started: bool
 
 proc protocolMatcher*(codec: string): Matcher =
@@ -278,13 +287,8 @@ proc canBeConnected*(pm: PeerManager,
 # Initialisation #
 ##################
 
-# currently disabled. note that peer connection state connected/disconnected
-# cant be tracked using this handler when more than one conn is allowed and
-# when using autonat. eg if a peer has 2 conns and one is disconnected we cant
-# assume that the peer is disconnected, because the other one might still be active.
-# note that even with maxconn = 1, autonat forces more than one connection.
+# called when a connection i) is created or ii) is closed
 proc onConnEvent(pm: PeerManager, peerId: PeerID, event: ConnEvent) {.async.} =
-
   case event.kind
   of ConnEventKind.Connected:
     let direction = if event.incoming: Inbound else: Outbound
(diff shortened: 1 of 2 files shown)


waku/v2/node/peer_manager/peer_manager.nim
```diff
@@ -278,44 +287,60 @@ proc canBeConnected*(pm: PeerManager,
...
-
   case event.kind
   of ConnEventKind.Connected:
     let direction = if event.incoming: Inbound else: Outbound
     discard
   of ConnEventKind.Disconnected:
     discard
 
+# called when a peer i) first connects to us ii) disconnects all connections from us
 proc onPeerEvent(pm: PeerManager, peerId: PeerId, event: PeerEvent) {.async.} =
```
alrevuelta, line 300: minor refactor to avoid some duplicated code. no funcionality change.

waku/v2/node/peer_manager/peer_manager.nim
```diff
@@ -520,6 +547,22 @@ proc pruneInRelayConns(pm: PeerManager, amount: int) {.async.} =
   for p in inRelayPeers[0..<connsToPrune]:
     await pm.switch.disconnect(p)
 
+proc pruneConnsByIp*(pm: PeerManager) {.async.} =
+  ## prunes connections based on ip colocation, allowing no more
+  ## than ColocationLimit inbound connections from same ip
+  ##
+
+  # update the table tracking ip and the connected peers
+  pm.updateIpTable()
```
jm-clius, line 556: Mmm. This rebuilds the IP table every time, so I don't see any reason to cache the `ipTable` on the `pm` object. In other words, why not have a `getPeersPerIp` that simply creates and populates such a table every time?

waku/v2/node/peer_manager/peer_manager.nim
```diff
@@ -649,6 +698,7 @@ proc start*(pm: PeerManager) =
   asyncSpawn pm.updateMetrics()
   asyncSpawn pm.relayConnectivityLoop()
   asyncSpawn pm.prunePeerStoreLoop()
+  asyncSpawn pm.pruneConnsByIpLoop()
   asyncSpawn pm.logSummary()
```
jm-clius, line 702: We have quite a large number of spawned tasks here. Wouldn't a single maintenance heartbeat be a better approach?

jm-clius (approved):
Thanks for this! Generally makes sense to me. See comments/suggestions below.

waku/v2/node/peer_manager/peer_manager.nim
alrevuelta: yup, but in the future I plan to use this table for other things. still unclear but some ideas:
* intercept a connection. before even allowing a connection to go through, check if colocation limit is ok.
* discovery. if we discover a peer violating the colocation limit, do not add it to the peer store.
* out connections. dont start connections to peers violating the colocation limit.

so would like to leave it if its fine.

waku/v2/node/peer_manager/peer_manager.nim
```diff
@@ -520,6 +547,22 @@ proc pruneInRelayConns(pm: PeerManager, amount: int) {.async.} =
...
+  # update the table tracking ip and the connected peers
+  pm.updateIpTable()
+
+  # trigger disconnections based on colocationLimit
+  for ip, peersInIp in pm.ipTable.pairs:
+    if peersInIp.len > pm.colocationLimit:
+      let connsToPrune = peersInIp.len - pm.colocationLimit
+      for peerId in peersInIp[0..<connsToPrune]:
+        debug "Pruning connection due to ip colocation", peerId = peerId, ip = ip
+        await pm.switch.disconnect(peerId)
```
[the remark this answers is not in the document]
alrevuelta: actually it wasnt obvious, good catch, the peer was not being removed. added f9c25b4783b9caf43034485856f9dd1bda7577fb removing + test case.

right now prunning is only for in connections, which is the attack this PR aims to mitigate. but yep i might implement it later on for out connections.

waku/v2/node/peer_manager/peer_manager.nim
alrevuelta: yup, merged two of them in this pr
49db93e98f68c2d8c8f6258e5e7f10f66e8be764

PR #1766, opened 2023-05-29
feat: Enable Hole Punching service

diegomrsantos:
Description:

Enable the Hole Punching Service.

### DIFF
--- a/waku/common/utils/nat.nim
+++ b/waku/common/utils/nat.nim
@@ -54,6 +54,9 @@ proc setupNat*(natConf, clientId: string,
         endpoint.udpPort = some(extUdpPort)
 
   else: # NatNone
+    if natConf == "none":
+      return ok((none(ValidIpAddress), none(Port), none(Port)))
+
     if not natConf.startsWith("extip:"):
       return err("not a valid NAT mechanism: " & $natConf)
 
@@ -64,4 +67,3 @@ proc setupNat*(natConf, clientId: string,
       return err("not a valid IP address: " & $natConf[6..^1])
 
   return ok(endpoint)
-
--- a/apps/wakunode2/app.nim
+++ b/apps/wakunode2/app.nim
@@ -11,6 +11,7 @@ import
   libp2p/crypto/crypto,
   libp2p/nameresolving/dnsresolver,
   libp2p/protocols/pubsub/gossipsub,
+  libp2p/discovery/discoverymngr,
   libp2p/peerid,
   eth/keys,
   json_rpc/rpcserver,
@@ -80,6 +81,9 @@ type
     rpcServer: Option[RpcHttpServer]
     restServer: Option[RestServerRef]
     metricsServer: Option[MetricsHttpServerRef]
+
+    topics: seq[string]
+    dm: DiscoveryManager
 
   AppResult*[T] = Result[T, string]
 
@@ -160,7 +164,8 @@ proc init*(T: type App, rng: ref HmacDrbgContext, conf: WakuNodeConf): T =
     rng: rng,
     key: key,
     record: record,
-    node: nil
+    node: nil,
+    topics: topics,
   )
 
 
@@ -329,13 +334,15 @@ proc setupWakuApp*(app: var App): AppResult[void] =
 
 ## Mount protocols
 
-proc setupProtocols(node: WakuNode,
-                    conf: WakuNodeConf,
-                    nodeKey: crypto.PrivateKey):
-                    Future[AppResult[void]] {.async.} =
+proc setupProtocols(app: App): Future[AppResult[void]] {.async.} =
   ## Setup configured protocols on an existing Waku v2 node.
   ## Optionally include persistent message storage.
   ## No protocols are started yet.
+
+  let
+    node = app.node
+    conf = app.conf
+    key = app.key
 
   # Mount relay on all nodes
   var peerExchangeHandler = none(RoutingRecordsHandler)
@@ -355,14 +362,7 @@ proc setupProtocols(node: WakuNode,
     peerExchangeHandler = some(handlePeerExchange)
 
   if conf.relay:
-    let pubsubTopics =
-      if conf.pubsubTopics.len > 0 or conf.contentTopics.len > 0:
-        # TODO autoshard content topics only once.
-        # Already checked for errors in app.init
-        let shards = conf.contentTopics.mapIt(getShard(it).expect("Valid Shard"))
-        conf.pubsubTopics & shards
-      else:
-        conf.topics
+    let pubsubTopics = app.topics
 
     try:
       await mountRelay(node, pubsubTopics, peerExchangeHandler = peerExchangeHandler)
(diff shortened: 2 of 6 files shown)


waku/v2/node/waku_node.nim
```diff
@@ -186,7 +197,8 @@ proc new*(T: type WakuNode,
     sendSignedPeerRecord = sendSignedPeerRecord,
     agentString = agentString,
     peerStoreCapacity = peerStoreCapacity,
-    services = @[Service(getAutonatService(rng))],
+    services = @[Service(hpservice)],
```
alrevuelta, line 200: just to make sure i got it right: autonatservice is included indirectly in `hpservice`? so no need for `services=[hpservice, autonatservice]`?

wondering what would happen if i try:
`services=[hpservice, autonatservice]`

waku/common/utils/nat.nim
```diff
@@ -54,6 +54,9 @@ proc setupNat*(natConf, clientId: string,
         endpoint.udpPort = some(extUdpPort)
 
   else: # NatNone
+    if natConf == "none":
```
alrevuelta, line 57: isnt this redundant? its the else case of `if strategy != NatNone:` so it will always be none?

waku/v2/node/waku_node.nim
```diff
@@ -42,6 +46,8 @@ import
   ./peer_manager,
   ./waku_switch
 
+export chronos
```
alrevuelta, line 49: is this needed?

waku/v2/node/waku_node.nim
```diff
@@ -172,6 +178,11 @@ proc new*(T: type WakuNode,
 
   info "Initializing networking", addrs= $netConfig.announcedAddresses
 
+  let relayClient = RelayClient.new()
+  let autoRelayService = AutoRelayService.new(1, relayClient, nil, rng)
+  let autonatService = getAutonatService(rng)
+  let hpservice = HPService.new(autonatService, autoRelayService)
```
alrevuelta, line 184: wondering if we should have hole punching behind a flag or always on by default. specially now at the begining that its new. opinion? @jm-clius

alrevuelta (commented):
thanks for this @diegomrsantos!

does a private node with this code get inbound connections out of the box as it is in the waku network?

waku/v2/node/waku_node.nim
diegomrsantos: Yes, `autonatservice` is added at line 184. Regarding your second question, it'd create two `autonatservice`, I didn't test it tbh, but it shouldn't be done.

waku/common/utils/nat.nim
diegomrsantos: Yes, but before it executed the code below:

```nim
    if not natConf.startsWith("extip:"):
      return err("not a valid NAT mechanism: " & $natConf)
```

waku/v2/node/waku_node.nim
diegomrsantos: @Menduist do you have more info about this issue? Is this a PR related to that?

waku/common/utils/nat.nim
Ivansete-status: Shall we just return an error in this case (` else: # NatNone `)?

I have the impression that the next is always run:
```nim
return err("not a valid NAT mechanism: " & $natConf)
```

Ivansete-status:
Thanks so much indeed for submitting this PR @diegomrsantos !
May kindly elaborate a bit more on its description please? For example, the next sections might help:
- summary.
- deeper details about what this PR addresses.
- issue. If exists, a link to the issue that needs to be sorted out.

You'll need to set an appropriate PR's title.

And, in order to fix the conflicts I would make this in your working branch:
- ```code
   git fetch origin
   ```
- ```code
   git rebase origin/master
   ```
- fix the possible conflicst :)

waku/v2/node/waku_node.nim
Menduist: Can't remember, what happens without this line?

waku/v2/node/waku_node.nim
Menduist: Maybe a flag that is true by default

waku/v2/node/waku_node.nim
diegomrsantos: I can't remember the error now, but it is related to generic procs. I think adding the `export` is the only way to fix it. I believe in this case it is caused by the `new` proc.

Menduist: Right, if you use a package inside a generic proc, you need to export it
Not a bug, just a design quirk of nim

waku/v2/node/waku_node.nim
Ivansete-status: I see the `hpservice` as a wrapper around `autonatservice`, so I think we don't need to add it twice.

https://github.com/status-im/nim-libp2p/blob/74c402ed9db4652a455c00c8d1713b222e3ef3d5/libp2p/services/hpservice.nim#L110

Ivansete-status (commented):
It looks great indeed!
I've added a little nitpicky comment.

How could we test it?

diegomrsantos:
@Ivansete-status @jm-clius I'm putting this work on hold for now. The Hole Punching Service requires that a private peer's relay address be advertised to peers interested in connecting to it. We are going to use the Rendezvous protocol for that, but the current nwaku version is just mounting the protocol. Advertising the peer's signed record, requesting other peers ones, and using it to discover new peers need to be implemented. It seems those things are already in place in go-waku but not sure if they have been fully tested.

Ivansete-status:
> @Ivansete-status @jm-clius I'm putting this work on hold...

Thanks for the comment @diegomrsantos! Is there anything that we should do to help this feature progress?
Kind regards

diegomrsantos:
@Ivansete-status If I get it right, circuit-relay addresses from private nodes aren't discoverable through discv5, because they are too big. For that reason, nwaku nodes need to use Rendezvous to advertise such addresses when they are private. Likewise, they also need to ask Rendezvous Points for such node's addresses.

Ivansete-status:
Thanks for the answer @diegomrsantos!
Then, in order to make this _Hole Punching service_ PR progress we first need to achieve the next items right?

- `nwaku` should use Rendezvous to advertise "circuit-relay addresses" when they are private.
- `nwaku` should ask Rendezvous Points for "circuit-relay addresses".

diegomrsantos:
That's correct. Regarding the first point, nwaku could use Rendezvous to advertise its addresses all the time, not only when it's private. This would probably make the implementation easier, but I'm not sure about the interaction with the current discv5 usage. This can be decided by the team, the only requirement for Hole Punching is that circuit-relay addresses of private nodes are advertised, discoverable, and dialed by other nodes.

diegomrsantos:
I have added the advertising part (in all cases, not only when the node is private) here https://github.com/waku-org/nwaku/pull/1766/commits/f64d12b445c78bdfff3195446881421bb11ea58e. It can be used as inspiration, I'm not sure I did it properly. 

It requires some fixes being done here https://github.com/status-im/nim-libp2p/pull/951.

Ivansete-status:
Thank you so much indeed for this PR @diegomrsantos!
It has been very inspirational but discontinued because it diverged quite a lot from the current master branch.
Therefore, we are tackling it in the following PR: https://github.com/waku-org/nwaku/pull/3112

PR #1769, opened 2023-05-30
feat(networking): integrate gossipsub scoring

### LINKED ISSUE #1756: feat: integrate gossipsub scoring
### Problem

Right now there is no mechanism to trigger disconections from bad relay peers, and on top of that we don't have any means to measure each peer repuation/score. Without this, a node can end up being connected to either bad quality peers or non honest peers.

### Suggested solution
* Integrate [gossipsub scoring](https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.1.md#the-score-function) and trigger disconnections based on it
* Use some safe weights until further research is done.

alrevuelta:
closes #1756 

Changes:

- [x] Only allow handling messages containing a valid `WakuMessage`. Just one type of handler. No more handling `seq[byte]` and parsing. Waku relay message handler takes directly a `WakuMesssage`.
- [x] Add a default validator to all subscribed topics, which enforces being able to decode the message into a valid `WakuMessage`. If ok the message is accepted, otherwise rejected. With this, "upper layers" now always get a `WakuMessage`.
- [x] Integrate gossipsub scoring + trigger dissconnections from bad peers.
- [x] Set some default safe parameters for gossipsub scoring to use.
- [x] Write some tests to ensure the feature is integrated ok, where disconnections based on score are triggered.
- [x] Change the in/out peer relation, enforcing 50/50%. 10% of outgoing peers seems to have some issues in simulations, *to be continued*.
- [x] Test in sandbox machine with 200 nodes.

### DIFF
--- a/apps/wakubridge/wakubridge.nim
+++ b/apps/wakubridge/wakubridge.nim
@@ -264,7 +264,7 @@ proc start*(bridge: WakuBridge) {.async.} =
 
   # Always mount relay for bridge.
   # `triggerSelf` is false on a `bridge` to avoid duplicates
-  await bridge.nodev2.mountRelay(triggerSelf = false)
+  await bridge.nodev2.mountRelay()
 
   # Bridging
   # Handle messages on Waku v1 and bridge to Waku v2
--- a/apps/wakubridge/wakubridge.nim
+++ b/apps/wakubridge/wakubridge.nim
@@ -275,12 +275,11 @@ proc start*(bridge: WakuBridge) {.async.} =
   bridge.nodev1.registerEnvReceivedHandler(handleEnvReceived)
 
   # Handle messages on Waku v2 and bridge to Waku v1
-  proc relayHandler(pubsubTopic: PubsubTopic, data: seq[byte]) {.async, gcsafe.} =
-    let msg = WakuMessage.decode(data)
-    if msg.isOk() and msg.get().isBridgeable():
+  proc relayHandler(topic: PubsubTopic, msg: WakuMessage): Future[void] {.async, gcsafe.} =
+    if msg.isBridgeable():
       try:
-        trace "Bridging message from V2 to V1", msg=msg.tryGet()
-        bridge.toWakuV1(msg.tryGet())
+        trace "Bridging message from V2 to V1", msg=msg
+        bridge.toWakuV1(msg)
       except ValueError:
         trace "Failed to convert message to Waku v1. Check content-topic format.", msg=msg
         waku_bridge_dropped.inc(labelValues = ["value_error"])
--- a/apps/wakubridge/wakubridge.nim
+++ b/apps/wakubridge/wakubridge.nim
@@ -265,6 +265,7 @@ proc start*(bridge: WakuBridge) {.async.} =
   # Always mount relay for bridge.
   # `triggerSelf` is false on a `bridge` to avoid duplicates
   await bridge.nodev2.mountRelay()
+  bridge.nodev2.wakuRelay.triggerSelf = false
 
   # Bridging
   # Handle messages on Waku v1 and bridge to Waku v2
--- a/tests/v2/waku_relay/test_wakunode_relay.nim
+++ b/tests/v2/waku_relay/test_wakunode_relay.nim
@@ -483,7 +483,7 @@ suite "WakuNode - Relay":
     # all nodes lower the score of nodes[0] (will change if gossipsub params or amount of msg changes)
     for i in 1..<5:
       check:
-        nodes[1].wakuRelay.peerStats[nodes[0].switch.peerInfo.peerId].score == -25009.0
+        nodes[1].wakuRelay.peerStats[nodes[0].switch.peerInfo.peerId].score == -250004.9
 
     # nodes[0] was blacklisted from all other peers, no connections
     check:
--- a/tests/v2/waku_relay/test_wakunode_relay.nim
+++ b/tests/v2/waku_relay/test_wakunode_relay.nim
@@ -483,7 +483,7 @@ suite "WakuNode - Relay":
     # all nodes lower the score of nodes[0] (will change if gossipsub params or amount of msg changes)
     for i in 1..<5:
       check:
-        nodes[1].wakuRelay.peerStats[nodes[0].switch.peerInfo.peerId].score == -250004.9
+        nodes[i].wakuRelay.peerStats[nodes[0].switch.peerInfo.peerId].score == -250004.9
 
     # nodes[0] was blacklisted from all other peers, no connections
     check:
--- a/tests/v2/waku_relay/test_wakunode_relay.nim
+++ b/tests/v2/waku_relay/test_wakunode_relay.nim
@@ -483,7 +483,7 @@ suite "WakuNode - Relay":
     # all nodes lower the score of nodes[0] (will change if gossipsub params or amount of msg changes)
     for i in 1..<5:
       check:
-        nodes[i].wakuRelay.peerStats[nodes[0].switch.peerInfo.peerId].score == -250004.9
+        nodes[i].wakuRelay.peerStats[nodes[0].switch.peerInfo.peerId].score == -249999.9
 
     # nodes[0] was blacklisted from all other peers, no connections
     check:
--- a/tests/v2/waku_relay/test_wakunode_relay.nim
+++ b/tests/v2/waku_relay/test_wakunode_relay.nim
@@ -94,14 +94,11 @@ suite "WakuNode - Relay":
     )
 
     var completionFut = newFuture[bool]()
-    proc relayHandler(topic: string, data: seq[byte]) {.async, gcsafe.} =
-      let msg = WakuMessage.decode(data)
-      if msg.isOk():
-        let val = msg.value()
-        check:
-          topic == pubSubTopic
-          val.contentTopic == contentTopic
-          val.payload == payload
+    proc relayHandler(topic: PubsubTopic, msg: WakuMessage): Future[void] {.async, gcsafe.} =
+      check:
+        topic == pubSubTopic
+        msg.contentTopic == contentTopic
+        msg.payload == payload
       completionFut.complete(true)
 
     node3.subscribe(pubSubTopic, relayHandler)
@@ -182,18 +179,13 @@ suite "WakuNode - Relay":
     node2.wakuRelay.addValidator(pubSubTopic, validator)
 
     var completionFut = newFuture[bool]()
-    proc relayHandler(topic: string, data: seq[byte]) {.async, gcsafe.} =
-      debug "relayed pubsub topic:", topic
-      let msg = WakuMessage.decode(data)
-      if msg.isOk():
-        let val = msg.value()
-        check:
-          topic == pubSubTopic
-          # check that only messages with contentTopic1 is relayed (but not contentTopic2)
-          val.contentTopic == contentTopic1
+    proc relayHandler(topic: PubsubTopic, msg: WakuMessage): Future[void] {.async, gcsafe.} =
+      check:
+        topic == pubSubTopic
+        # check that only messages with contentTopic1 is relayed (but not contentTopic2)
+        msg.contentTopic == contentTopic1
       # relay handler is called
       completionFut.complete(true)
-
 
     node3.subscribe(pubSubTopic, relayHandler)
     await sleepAsync(500.millis)
@@ -269,14 +261,11 @@ suite "WakuNode - Relay":
     await node1.connectToNodes(@[node2.switch.peerInfo.toRemotePeerInfo()])
 
     var completionFut = newFuture[bool]()
-    proc relayHandler(topic: string, data: seq[byte]) {.async, gcsafe.} =
-      let msg = WakuMessage.decode(data)
-      if msg.isOk():
-        let val = msg.value()
-        check:
-          topic == pubSubTopic
-          val.contentTopic == contentTopic
-          val.payload == payload
+    proc relayHandler(topic: PubsubTopic, msg: WakuMessage): Future[void] {.async, gcsafe.} =
+      check:
+        topic == pubSubTopic
+        msg.contentTopic == contentTopic
+        msg.payload == payload
       completionFut.complete(true)
 
     node1.subscribe(pubSubTopic, relayHandler)
@@ -313,14 +302,11 @@ suite "WakuNode - Relay":
     await node1.connectToNodes(@[node2.switch.peerInfo.toRemotePeerInfo()])
 
     var completionFut = newFuture[bool]()
-    proc relayHandler(topic: string, data: seq[byte]) {.async, gcsafe.} =
-      let msg = WakuMessage.decode(data)
-      if msg.isOk():
-        let val = msg.value()
-        check:
-          topic == pubSubTopic
-          val.contentTopic == contentTopic
-          val.payload == payload
+    proc relayHandler(topic: PubsubTopic, msg: WakuMessage): Future[void] {.async, gcsafe.} =
+      check:
+        topic == pubSubTopic
+        msg.contentTopic == contentTopic
+        msg.payload == payload
       completionFut.complete(true)
 
     node1.subscribe(pubSubTopic, relayHandler)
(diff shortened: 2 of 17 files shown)


apps/wakubridge/wakubridge.nim
```diff
@@ -264,7 +264,7 @@ proc start*(bridge: WakuBridge) {.async.} =
 
   # Always mount relay for bridge.
   # `triggerSelf` is false on a `bridge` to avoid duplicates
-  await bridge.nodev2.mountRelay(triggerSelf = false)
+  await bridge.nodev2.mountRelay()
```
vpavlin, line 267: Is the `  bridge.nodev2.wakuRelay.triggerSelf = false` missing here? 
￼

waku/v2/waku_relay/protocol.nim
```diff
@@ -29,19 +30,100 @@ logScope:
...
+    meshMessageDeliveriesWeight: -5.0,
+    meshMessageDeliveriesDecay: 0.5,
+    meshMessageDeliveriesCap: 10,
+    meshMessageDeliveriesThreshold: 1,
+    meshMessageDeliveriesWindow: 5.milliseconds,
+    meshMessageDeliveriesActivation: 10.seconds,
+
+    # p3b: tracks history of prunes
+    meshFailurePenaltyWeight: -1.0,
+    meshFailurePenaltyDecay: 0.5,
```
Menduist, line 57: These parameters are the one I talked about that are not fit for waku, because they set an expected rate of messages per topic, and we can't estimate the actual rate of messages in waku

alrevuelta: mm i see. reading the docs i didnt got that impression. shall we zero both values?

> P₃b: Mesh Message Delivery Failures for a topic. This is a sticky parameter that counts the number of mesh message delivery failures. Whenever a peer is pruned with a negative score, the parameter is augmented by the rate deficit at the time of prune. This is intended to keep history of prunes so that a peer that was pruned because of underdelivery cannot quickly get re-grafted into the mesh. The parameter is mixed with negative weight.

apps/wakubridge/wakubridge.nim
alrevuelta: indeed!

waku/v2/waku_relay/protocol.nim
Menduist: Yeah, 0 values will work

All of these parameters are very obscure, but this parameter is basically a value that keep increasing as time passes, and every time a peer sends a message, it's decreased by some amount
So if the rate at which it decreases is < than time passing, all of the peers in the mesh will be penalized (because of the negative weight)

That's a convoluted way to impose a certain cadence to the messages in the topic

apps/wakubridge/wakubridge.nim
alrevuelta: done 3d8ba07d5944d5374ccb981c06716db46e727518

waku/v2/waku_relay/protocol.nim
alrevuelta: done 3d8ba07d5944d5374ccb981c06716db46e727518

waku/v2/waku_relay/protocol.nim
```diff
@@ -131,31 +190,44 @@ iterator subscribedTopics*(w: WakuRelay): lent PubsubTopic =
...
+  let wrappedHandler = proc(pubsubTopic: string, data: seq[byte]): Future[void] {.gcsafe, raises: [Defect].} =
+    let decMsg = WakuMessage.decode(data)
+    if decMsg.isErr():
+      # fine if triggerSelf enabled, since validators are bypassed
+      error "failed to decode WakuMessage, validator passed a wrong message"
+      let fut = newFuture[void]()
+      fut.complete()
+      return fut
+    else:
+      return handler(pubsubTopic, decMsg.get)
```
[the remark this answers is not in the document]
alrevuelta: sure b1df0dd55a42434f45b2fbfe4fad45272574e676

waku/v2/waku_relay/protocol.nim
```diff
@@ -131,31 +190,44 @@ iterator subscribedTopics*(w: WakuRelay): lent PubsubTopic =
...
 
-  var subsHandler: PubsubRawHandler
-  when handler is SubscriptionHandler:
-    subsHandler = proc(pubsubTopic: PubsubTopic, data: seq[byte]): Future[void] {.gcsafe.} =
-        let decodeRes = WakuMessage.decode(data)
-        if decodeRes.isErr():
-          debug "message decode failure", pubsubTopic=pubsubTopic, error=decodeRes.error
-          return
+  # rejects messages that are not WakuMessage
+  proc validator(pubsubTopic: string, message: messages.Message): Future[ValidationResult] {.async.} =
```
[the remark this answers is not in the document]
alrevuelta: tldr: indeed, done b1df0dd55a42434f45b2fbfe4fad45272574e676

longer version. related to your other comment [here](https://github.com/waku-org/nwaku/pull/1769#pullrequestreview-1455422015) I had the validator inside since I was trying to optimize this, to avoid decoding the message twice (or as many times as validators we have). This optimization was similar to the one in [nimbus](https://github.com/status-im/nimbus-eth2/blob/c22962f3821d069f86ee38826443f9830931cc7c/beacon_chain/networking/eth2_network.nim#L2385-L2386) , where the validator acts also as the handler. Meaning that inside the validator, you also call the handler, and then you dont have to register the handle (just pass nil in subscribe) because the validator calls it for you. Like this you dont have to decode the message i) in the validation stage and ii) in the handler stage, just once.

something like:
```nim
proc validator(pubsubTopic: string, message: messages.Message): Future[ValidationResult] {.async.} =
  let msg = WakuMessage.decode(message.data)
  if msg.isOk():
    yourHandler(yyy) <----
    return ValidationResult.Accept
  return ValidationResult.Reject
```

Buut, at the end I discarded this solution as it has some implications. For example, when having multiple validators (rln, signed topic,e tc). This hacked validator containing also the handler would need to be executed always the last, and this hack on top of a hack seemed too much.

I haven't done a detailed analysis on the impact of decoding the message twice, but haven't seen a significant increase in the simulations on any parameter. So at this stage I would say its overoptimizing. Leaving this idea here in case we want to implement it later on.

alrevuelta:
> Thanks. Minor nits below, but LGTM! Not approving as I know you want to test the selected parameters first. One more uncertainty is what the performance/memory knock would be if we decode all messages twice, i.e. for both validation and handling. Assumption here is negligible/nothing, but may be worth double checking in simulation.

@jm-clius thanks for the comments. Replied to this [here](https://github.com/waku-org/nwaku/pull/1769#discussion_r1214245921). Indeed, seems neglible and left an idea in that comment in case in the future we want to optimize this.

waku/v2/waku_relay/protocol.nim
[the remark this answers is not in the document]
Menduist: We have an issue somewhere (https://github.com/status-im/nimbus-eth2/issues/3043) about adding a "user data" field to the Message type, to let the app decode a single time and store the decoded version there.
That will require some trickery since libp2p doesn't know the type of the user data, but feel free to pick that up

waku/v2/waku_relay/protocol.nim
```diff
@@ -62,51 +144,28 @@ proc initProtocolHandler(w: WakuRelay) =
...
-    # for pubSubTopic in defaultPubsubTopics:
-    #   wr.addValidator(pubSubTopic, validator)
+    procCall GossipSub(w).initPubSub()
+    w.initProtocolHandler()
 
-  ok(wr)
+  except InitializationError:
+    return err("initialization error: " & getCurrentExceptionMsg())
 
+  ok(w)
```
Ivansete-status, line 168: Not a big deal at all but for I personally prefer being explicit and always use `return`. Just for new incomers to Nim :)
```suggestion
  return ok(w)
```

Ivansete-status (approved):
LGTM! This PR is very cool! I just have a doubt re the `unsubscribe` change.

waku/v2/waku_relay/protocol.nim
```diff
@@ -131,31 +196,37 @@ iterator subscribedTopics*(w: WakuRelay): lent PubsubTopic =
...
+
+  procCall GossipSub(w).unsubscribe(topics, nil)
 
 proc unsubscribeAll*(w: WakuRelay, pubsubTopic: PubsubTopic) =
   debug "unsubscribeAll", pubsubTopic=pubsubTopic
 
   procCall GossipSub(w).unsubscribeAll(pubsubTopic)
 
-
 proc publish*(w: WakuRelay, pubsubTopic: PubsubTopic, message: WakuMessage|seq[byte]): Future[int] {.async.} =
```
LNSD, line 230: If the idea is not to return a `seq[byte]` only `WakuMessage`s in the subscription handler anymore, then to be consistent, you should not allow publishing a `seq[byte]`, only `WakuMessage`s.

waku/v2/waku_relay/protocol.nim
```diff
@@ -131,31 +196,37 @@ iterator subscribedTopics*(w: WakuRelay): lent PubsubTopic =
...
-        let decodeRes = WakuMessage.decode(data)
-        if decodeRes.isErr():
-          debug "message decode failure", pubsubTopic=pubsubTopic, error=decodeRes.error
-          return
+  # we need to wrap the handler since gossipsub doesnt understand WakuMessage
+  let wrappedHandler = proc(pubsubTopic: string, data: seq[byte]): Future[void] {.gcsafe, raises: [Defect].} =
+    let decMsg = WakuMessage.decode(data)
+    if decMsg.isErr():
+      # fine if triggerSelf enabled, since validators are bypassed
+      error "failed to decode WakuMessage, validator passed a wrong message"
```
LNSD, line 204: This can be extremely noisy. Think of a "DoS" scenario. It will blow up the logging system.

waku/v2/waku_relay/protocol.nim
```diff
@@ -120,6 +179,12 @@ method stop*(w: WakuRelay) {.async.} =
   debug "stop"
   await procCall GossipSub(w).stop()
 
+# rejects messages that are not WakuMessage
+proc validator(pubsubTopic: string, message: messages.Message): Future[ValidationResult] {.async.} =
+  let msg = WakuMessage.decode(message.data)
```
LNSD, line 184: **Optimization note:** 

You don't need to decode a message to know if the message is correct. Although typically Protocol buffer libraries perform both at the same time when decoding, validation and deserialization can be performed independently. For example, using nim-libp2p's protobuf library, you can check if the required fields are present without allocating memory.

apps/wakubridge/wakubridge.nim
alrevuelta: by now i would say `triggerSelf = true` should be the default value. the bridge is a particular case.

tests/v2/waku_relay/test_wakunode_relay.nim
```diff
@@ -468,3 +445,54 @@ suite "WakuNode - Relay":
...
+    for i in 0..<5:
+      check:
+        nodes[i].peerManager.switch.connManager.getConnections().len == 4
+
+    # node[0] publishes wrong messages (random bytes not decoding into WakuMessage)
+    for j in 0..<50:
+      discard await nodes[0].wakuRelay.publish(topic, urandom(1*(10^3)))
+
+    # long wait, must be higher than the configured decayInterval (how often score is updated)
+    await sleepAsync(20.seconds)
```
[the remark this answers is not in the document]
alrevuelta: yep, this is a high delay but would like to leave it as it is, so that we dont end up testing something different than how it looks in reality.

waku/v2/waku_relay/protocol.nim
```diff
@@ -29,19 +30,100 @@ logScope:
...
+
+    # remember peer score during x after it disconnects
+    retainScore: chronos.minutes(10),
+
+    # p5: application specific, unset
+    appSpecificWeight: 0.0,
+
+    # p6: penalizes peers sharing more than threshold ips
+    ipColocationFactorWeight: -50.0,
+    ipColocationFactorThreshold: 5.0,
```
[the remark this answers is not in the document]
alrevuelta: this indeed lowers the peer score based on the ip (with more than 5 ips it kicks in). but note that if 6 peers are behind the same ip and they perform great in other scores, we wont really prune any peer.

note that this doesnt replace our peer prunning based on ips, since this score just applies to peers participating in gossipsub.

waku/v2/waku_relay/protocol.nim
alrevuelta: interesting, added a note here: dc70df97f859a4fa907db0be13bfd56d7a23dce8 for future optimizations.

waku/v2/waku_relay/protocol.nim
alrevuelta: done dc70df97f859a4fa907db0be13bfd56d7a23dce8

waku/v2/waku_relay/protocol.nim
```diff
@@ -29,19 +30,100 @@ logScope:
...
+    meshFailurePenaltyDecay: 0.5,
+
+    # p4: penalizes invalid messages. highly penalize
+    # peers sending wrong messages
+    invalidMessageDeliveriesWeight: -100.0,
+    invalidMessageDeliveriesDecay: 0.5
+  )
+
+# see: https://rfc.vac.dev/spec/29/#gossipsub-v10-parameters
+const gossipsubParams = GossipSubParams(
```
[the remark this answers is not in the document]
alrevuelta: missed the comment, done b1df0dd55a42434f45b2fbfe4fad45272574e676

waku/v2/node/waku_node.nim
```diff
@@ -308,10 +304,10 @@ proc unsubscribe*(node: WakuNode, topic: PubsubTopic, handler: WakuRelayHandler)
     error "Invalid API call to `unsubscribe`. WakuRelay not mounted."
     return
 
-  debug "unsubscribe", oubsubTopic= topic
+  debug "unsubscribe", pubsubTopic=topic
 
   let wakuRelay = node.wakuRelay
-  wakuRelay.unsubscribe(@[(topic, handler)])
+  wakuRelay.unsubscribe(topic)
```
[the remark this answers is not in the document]
alrevuelta: good catch. this was a leftover. i have replaced unsubscribeAll by unsubcribe. So whenever we call now unsubscribe, we unsubscribe all handlers from the topic, meaning 100% unsubscribing to the topic.

unsubscribing individual handlers seemed very specific and since now we wrap the handler to only accept a WakuMessage, it wasnt trivial to adapt it.

If in the future we need to unsubscribe individual handlers, we can revisit this, but imho it would be better to incentivice people to use a single handler that does everything.

fixed dc70df97f859a4fa907db0be13bfd56d7a23dce8

waku/v2/waku_relay/protocol.nim
```diff
@@ -29,19 +30,100 @@ logScope:
...
+    dOut: 3,
+    dLazy: 6,
+
+    heartbeatInterval: chronos.seconds(1),
+    historyLength: 6,
+    historyGossip: 3,
+    fanoutTTL: chronos.minutes(1),
+    seenTTL: chronos.minutes(2),
+
+    # no gossip is send to peers below this score
```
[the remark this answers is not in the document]
alrevuelta: thanks dc70df97f859a4fa907db0be13bfd56d7a23dce8

waku/v2/waku_relay/protocol.nim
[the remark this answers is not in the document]
alrevuelta: thanks dc70df97f859a4fa907db0be13bfd56d7a23dce8

waku/v2/waku_relay/protocol.nim
alrevuelta: yep, i actually prefer to be explicit with this
dc70df97f859a4fa907db0be13bfd56d7a23dce8

waku/v2/waku_relay/protocol.nim
alrevuelta: yup, done dc70df97f859a4fa907db0be13bfd56d7a23dce8

waku/v2/waku_relay/protocol.nim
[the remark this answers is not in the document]
jm-clius: Ok, very interesting about both possible approaches to avoiding double decoding. Agree with doing what is simple and supported in libp2p for now.

jm-clius (approved):
LGTM, thanks!

PR #1770, opened 2023-05-31
feat(wakunode2): enable libp2p rendezvous protocol by default

### LINKED ISSUE #1605: feat: integrate libp2p rendezvous discovery
This issue relates to https://github.com/vacp2p/research/issues/176

It tracks the integration of libp2p rendezvous discovery into nwaku. An [underlying implementation](https://github.com/status-im/nim-libp2p/blob/8d5ea43e2bbd9efdb34703fe6249efe2a88cc00f/libp2p/protocols/rendezvous.nim) already exists in nim-libp2p.

vpavlin:
# Description

Rendezvous Discovery protocol allows peers to advertise their addresses at a rendezvous point and other peers to query the rendezvous point to discover new peers.

This will be particularly useful for peers using circuit relay as they can then easily advertise their circuit relay multiaddress and be discovered by other network participants.

This PR uses libp2p Rendezvous implementation and mounts the protocol by default when relay protocol is enabled.

# Changes

- [x] mount Rendezvous protocol when relay is mounted
- [x] add simple test of libp2p rendezvous protocol to waku_switch test suite
- [x] add a check to wakunode2 app tests that rendezvous is enabled along with relay

## Issue

closes #1605

### DIFF
--- a/waku/v2/node/waku_node.nim
+++ b/waku/v2/node/waku_node.nim
@@ -19,6 +19,7 @@ import
   libp2p/protocols/pubsub/rpc/messages,
   libp2p/protocols/connectivity/autonat/client,
   libp2p/protocols/connectivity/autonat/service,
+  libp2p/protocols/rendezvous,
   libp2p/nameresolving/nameresolver,
   libp2p/builders,
   libp2p/transports/tcptransport,
@@ -101,6 +102,7 @@ type
     libp2pPing*: Ping
     rng*: ref rand.HmacDrbgContext
     wakuDiscv5*: WakuDiscoveryV5
+    rendezvous*: RendezVous
     announcedAddresses* : seq[MultiAddress]
     started*: bool # Indicates that node has started listening
 
@@ -920,6 +922,16 @@ proc stopDiscv5*(node: WakuNode): Future[bool] {.async.} =
       await node.wakuDiscv5.closeWait()
 
     debug "Successfully stopped discovery v5 service"
+
+proc mountRendezvous*(node: WakuNode) {.async, raises: [Defect, LPError].} =
+  info "mounting rendezvous discovery protocol"
+
+  node.rendezvous = RendezVous.new(node.switch)
+
+  if node.started:
+    await node.rendezvous.start()
+
+  node.switch.mount(node.rendezvous)
 
 
 proc start*(node: WakuNode) {.async.} =
--- a/waku/v2/node/waku_switch.nim
+++ b/waku/v2/node/waku_switch.nim
@@ -10,6 +10,7 @@ import
   eth/keys,
   libp2p/crypto/crypto,
   libp2p/protocols/pubsub/gossipsub,
+  libp2p/protocols/rendezvous,
   libp2p/nameresolving/nameresolver,
   libp2p/builders,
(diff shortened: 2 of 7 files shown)


tests/v2/test_waku_switch.nim
```diff
@@ -100,3 +111,37 @@ suite "Waku Switch":
...
+      destSwitch = newRendezvousClientSwitch(destClient) #client
+
+    # Setup client rendezvous
+    wakuClient.setup(wakuSwitch)
+    sourceClient.setup(sourceSwitch)
+    destClient.setup(destSwitch)
+
+    await allFutures(wakuSwitch.start(), sourceSwitch.start(), destSwitch.start())
+
+    # Connect clients to the rendezvous pint
```
Ivansete-status, line 133: small typo
```suggestion
    # Connect clients to the rendezvous point
```

Ivansete-status (approved):
LGTM! Just a little comment

tests/v2/test_waku_switch.nim
```diff
@@ -100,3 +111,37 @@ suite "Waku Switch":
 
     ## Teardown
     await allFutures(wakuSwitch.stop(), sourceSwitch.stop(), destSwitch.stop())
+
+  asyncTest "Waku Switch uses Rendezvous":
```
jm-clius, line 115: I wouldn't include this in the `switch` tests. Since this is a separate protocol I suggest creating a new `test_waku_rendezvous` suite (even if it contains only the one test for now).

waku/v2/node/waku_switch.nim
```diff
@@ -9,7 +9,7 @@ import
   chronos, chronicles,
   eth/keys,
   libp2p/crypto/crypto,
-  libp2p/protocols/pubsub/gossipsub,
+  libp2p/protocols/[pubsub/gossipsub, rendezvous],
```
jm-clius, line 12: We generally have imports fully qualified and in separate lines (collective imports like this are difficult to "find and replace" at later stage). See [coding guidelines](https://hackmd.io/@vac/main/%2F5oIVOuQcSM2WKrnDLUPcpg#Importsexports)

waku/v2/node/waku_node.nim
```diff
@@ -21,6 +21,7 @@ import
   libp2p/protocols/connectivity/autonat/service,
   libp2p/nameresolving/nameresolver,
   libp2p/builders,
+  libp2p/protocols/rendezvous,
```
jm-clius, line 24: Very much a nitpick: while `builders` above clearly doesn't follow this practice, we tend to sort imports more or less alphabetically (with the exception of common imports (`std` `chronos` `metrics` etc.) which are usually top-level). Not a strict rule, but would be great for this import to fit in with the other `libp2p/protocols` above :)

jm-clius (approved):
LGTM! All my comments below are nitpicks (except perhaps moving the test to a different suite), so feel free to address/ignore as you see fit.

alrevuelta (approved):

richard-ramos:
Can confirm that go-waku can use nwaku as a rendezvous point in this PR! :rocket:

PR #1771, opened 2023-05-31
test(postgres): Allow run postgres tests only on ci Linux platforms

Ivansete-status:
## Summary
Allowing the execution of postgres unit tests on Linux GitHub runners

## Note
The next file is only left in order to run local tests manually:
tests/postgres-docker-compose.yml

The ci launches its postgres instance with `sudo docker run ...`

## Issue

### DIFF
--- a/.github/workflows/ci-experimental.yml
+++ b/.github/workflows/ci-experimental.yml
@@ -47,7 +47,6 @@ jobs:
       common: ${{ steps.filter.outputs.common }}
       v2: ${{ steps.filter.outputs.v2 }}
 
-
   build-v2:
     needs: changes
     if: ${{ needs.changes.outputs.v2 == 'true' || needs.changes.outputs.common == 'true' }}
@@ -88,6 +87,7 @@ jobs:
     timeout-minutes: 60
 
     name: test-v2-${{ matrix.os }}
+
     steps:
       - name: Checkout code
         uses: actions/checkout@v3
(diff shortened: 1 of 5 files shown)


.github/workflows/ci-experimental.yml
```diff
@@ -106,4 +106,9 @@ jobs:
           key: ${{ runner.os }}-vendor-modules-${{ steps.submodules.outputs.hash }}
 
       - name: Run tests
-        run: make V=1 LOG_LEVEL=DEBUG QUICK_AND_DIRTY_COMPILER=1 test2 testwakunode2
+        run: |
+          if [ ${{ runner.os }} == "Linux" ]; then
```
rymnc, line 110: Hey, people use https://github.com/abiosoft/colima as a workaround to the docker and mac image licensing issues, do you think we can try the same as well? 
non-blocker, just curious!

Ivansete-status: Morning @rymnc ! thanks so much for the suggestion. I take a note to try it in the future when I have more time.

jm-clius (approved):

.github/workflows/ci-experimental.yml
```diff
@@ -106,4 +106,9 @@ jobs:
...
+            sudo docker run --rm -d -e POSTGRES_PASSWORD=test123 -p 5432:5432 postgres:9.6-alpine
```
vpavlin, line 111: This should wait for the image pull, so slow connection should not be an issue, but in theory it could take some small amount of time to start the postgres itself, but since the postgres test is not the first to run, we should be fine. Just noting this in case we get into some "race condition" errors in the future:)

Another thing to consider is the Docker Hub rate limits - since you docker client is not authenticated, we could get rate limited on something like IP-range in case many people pull images from Docker Hub in the same time window (adding a docker login with token in secret should solve this in case it happens)

Ivansete-status: Thanks for the comments @vpavlin !
Do you want me to tackle the auth issue now or can I merge and take this point into consideration if the problem arises?

vpavlin (approved):

PR #1779, opened 2023-06-02
feat(postgres): Adding a postgres async pool to make the db interactions asynchronous

Ivansete-status:
## Summary
Allowing asynchronous calls 

## Further details

This PR:
   - continues the postgres asynchronous adoption started by @LNSD in https://github.com/waku-org/nwaku/pull/1631. I submit a different PR in order to adapt it better to the current code state and respect the Lorenzo's PR.
   - makes the postgres asynchronous but is not yet integrated in the wakunode app. So far is only used by the tests.
   - pending to implement the enhancement mentioned by @Menduist - https://github.com/waku-org/nwaku/pull/1631#discussion_r1154267054

## How to test
1. Start postgres database locally
   ```code
   sudo docker compose -f tests/docker-compose.yml up
   ```
2. Run tests locally
   ```code
   . env.sh
   ```
   ```code
   nim c -r -d:chronicles_log_level=WARN --verbosity=0 --hints=off --outdir=build ./tests/v2/waku_archive/test_driver_postgres.nim
   ```

## Issue


waku/v2/waku_archive/driver/postgres_driver/asyncpool.nim
```diff
@@ -0,0 +1,202 @@
...
+      await sleepAsync(0.milliseconds) # Do not block the async runtime
+    return ok()
+
+  pool.state = PgAsyncPoolState.Closing
+
+  # wait for the connections to be released and close them, without
+  # blocking the async runtime
+  if pool.conns.anyIt(it.busy):
+    while pool.conns.anyIt(it.busy):
+      await sleepAsync(0.milliseconds)
```
vpavlin, line 75: Why is the non-sleeping sleep here?

vpavlin: Or maybe a better question - what does it do?

tests/v2/waku_archive/test_driver_postgres.nim
```diff
@@ -27,76 +27,62 @@ suite "Postgres driver":
...
 
-    ## When
-    let driverRes = PostgresDriver.new(storeMessageDbUrl)
+    assert driverRes.isOk(), driverRes.error
 
-    ## Then
-    require:
-      driverRes.isOk()
+    let driver = driverRes.value
+    let _ = await driver.reset()
```
rymnc, line 36:
```suggestion
    discard await driver.reset()
```
maybe?

tests/v2/waku_archive/test_driver_postgres.nim
```diff
@@ -27,76 +27,62 @@ suite "Postgres driver":
...
-    ## Given
+  asyncTest "Insert a message":
     const contentTopic = "test-content-topic"
 
     let driverRes = PostgresDriver.new(storeMessageDbUrl)
+    assert driverRes.isOk(), driverRes.error
 
-    require:
-      driverRes.isOk()
+    let driver = driverRes.value
```
rymnc, line 72:
```suggestion
    let driver = driverRes.get()
```

tests/v2/waku_archive/test_driver_postgres.nim
```diff
@@ -27,76 +27,62 @@ suite "Postgres driver":
...
 
-    discard driverRes.get().reset()
-    discard driverRes.get().init()
+    let _ = await driver.reset()
 
-    let driver: ArchiveDriver = driverRes.tryGet()
-    require:
-      not driver.isNil()
+    let initRes = await driver.init()
+    assert initRes.isOk(), initRes.error
```
rymnc, line 77: Maybe we should follow this pattern in other test files too so we can see why the result is an error!
non-blocker

waku/v2/waku_archive/driver/postgres_driver/asyncpool.nim
```diff
@@ -0,0 +1,202 @@
...
+  {.push raises: [].}
+
+import
+  std/sequtils,
+  stew/results,
+  chronicles,
+  chronos
+import
+  ../../driver,
+  connection
```
rymnc, line 15: should this import be with the std lib import group?

waku/v2/waku_archive/driver/postgres_driver/asyncpool.nim
```diff
@@ -0,0 +1,202 @@
...
+  ## Runs the SQL query getting results.
+
+  let connIndexRes = await pool.getConnIndex()
+  if connIndexRes.isErr():
+    return err("connRes.isErr in query: " & connIndexRes.error)
+
+  let conn = pool.conns[connIndexRes.value].dbConn
+  defer: pool.releaseConn(conn)
+
+  return await conn.rows(sql(query), args)
```
rymnc, line 142: Maybe we should try-except here to bubble the error as a result instead of a failed future?

waku/v2/waku_archive/driver/postgres_driver/asyncpool.nim
```diff
@@ -0,0 +1,202 @@
...
+  let conn = pool.conns[connIndexRes.value].dbConn
+  defer: pool.releaseConn(conn)
+
+  var preparedStmt = pool.conns[connIndexRes.value].insertStmt
+  if cast[string](preparedStmt) == "":
+    # The connection doesn't have insertStmt set yet. Let's create it.
+    # Each session/connection should have its own prepared statements.
+    try:
+      pool.conns[connIndexRes.value].insertStmt =
+                                    conn.prepare("insertRow", sql(baseStmt), 7)
```
rymnc, line 184: I'm probably missing something here, what is 7?

waku/v2/waku_archive/driver/postgres_driver/connection.nim
```diff
@@ -0,0 +1,105 @@
...
+
+  if message.len > 0:
+    return err($message)
+
+  return ok()
+
+proc open*(connString: string):
+           Result[DbConn, string] =
+  ## Opens a new connection.
+  var conn:DbConn = nil
```
rymnc, line 30:
```suggestion
  var conn: DbConn = nil
```

waku/v2/waku_archive/driver/postgres_driver/postgres_driver.nim
```diff
@@ -39,85 +38,81 @@ proc createTableQuery(): string =
...
+                        Future[ArchiveDriverResult[void]] {.async.}  =
 
-method reset*(s: PostgresDriver): ArchiveDriverResult[void] {.base.} =
-  try:
-    let res = s.connection.tryExec(sql(dropTableQuery()))
-    if not res:
-      return err("failed to reset database")
-  except DbError:
-    return err("failed to reset database")
+  let ret = await s.connPool.exec(createTableQuery(), newSeq[string](0))
```
rymnc, line 74: try-except here and bubble the error within the result?

waku/v2/waku_archive/driver/postgres_driver/postgres_driver.nim
```diff
@@ -282,9 +280,9 @@ proc sleep*(s: PostgresDriver, seconds: int):
   # database for the amount of seconds given as a parameter.
   try:
     let params = @[$seconds]
-    s.connection.exec(sql"SELECT pg_sleep(?)", params)
+    let _ = await s.connPool.query("SELECT pg_sleep(?)", params)
```
rymnc, line 283:
```suggestion
    asyncCheck await s.connPool.query("SELECT pg_sleep(?)", params)
```

rymnc: https://forum.nim-lang.org/t/5279#33208

discard is OK for tests imo (in some situations)

rymnc (commented):
Great PR, left a few comments for style and couple queries, thanks :)

waku/v2/waku_archive/driver/postgres_driver/asyncpool.nim
Ivansete-status: Hi @vpavlin ! The main purpose of this `await` is to bring the Nim's asynchornous runtime the opportunity to attend other futures that are being "awaited". However, we might need to follow a better approach in further PR's (this comment is considered as a ToDo in the offending issue:)

https://github.com/waku-org/nwaku/pull/1631#discussion_r1154267054

waku/v2/waku_archive/driver/postgres_driver/asyncpool.nim
Ivansete-status: Actually this refers to a module set in the same folder. I'll make it more explicit with `./connection`

waku/v2/waku_archive/driver/postgres_driver/asyncpool.nim
Ivansete-status: 7 is the number of parameters to be handled.
In this case, we are only setting an unique "insertRow" prepared statement and the `baseStmt` has the next value:

```
INSERT INTO messages (id, storedAt, contentTopic, payload, pubsubTopic, version, timestamp) VALUES ($1, $2, $3, $4, $5, $6, $7);
```

The concept of "prepared statement" is interesting from a performance point of view.

rymnc: I see! thanks, its much clearer now
Can we make this value a constant near the "prepared statement", in case the value changes in the future?

waku/v2/waku_archive/driver/postgres_driver/postgres_driver.nim
Ivansete-status: Thanks for the recommendation! However, the test got stuck with that.
Therefore, in the end I'll suggest sth like:
```code
    let sleepRes = await s.connPool.query("SELECT pg_sleep(?)", params)
    if sleepRes.isErr():
      return err("error in postgres_driver sleep: " & sleepRes.error)
```

jm-clius (approved):
LGTM and thanks for taking this over!

tests/v2/waku_archive/test_driver_postgres.nim
Ivansete-status: Morning @rymnc ! After checking the code to apply this change I see that the change would be quite wide but we won't get very great investment return by doing so, as the tests don't tend to fail very often.

Therefore, I think is better to just apply this technique from now on in the upcoming tests.

Of course, if you consider it worth the effort, then I will make it for the current tests :)

alrevuelta (approved):

vpavlin (approved):

rymnc (approved):
LGTM, awesome PR!

PR #1780, opened 2023-06-05
chore(ci): properly set os and architecture for nightly and release

### LINKED ISSUE #1760: chore: asses and configure target CPU architecture properly
### Background
After hitting an issue where GH Action runners ran on CPU with "exotic" instructions, we introduced an explicit setting to enforce the most common instruction set (https://github.com/waku-org/nwaku/pull/1759)

This is not optimal when we consider Waku would benefit from more modern instructions considering it's using crypto and hashing algorithms relatively extensively.

Nimbus pretty explicitly controls the target CPU architectures here https://github.com/status-im/nimbus-eth2/blob/stable/config.nims#L94

We should incorporate similar approach.

### Details

Config section to decided on the arch to use: https://github.com/status-im/nimbus-eth2/blob/stable/config.nims#L94
CI builds run with `native`
Release builds run with `disableMarchNative`: https://github.com/status-im/nimbus-eth2/blob/stable/docker/dist/entry_point.sh#LL39C31-L39C31

### Acceptance criteria

Arch `native` is only used for non-release builds, realeasable binary artifacts must be built with limited/generic CPU instruction set, but not with basic `x86-64` as that could hurt performance.

vpavlin:
# Description
This PR updates `config.nims` according to findings by Nimbus about various CPU archs and options (https://github.com/status-im/nimbus-eth2/blob/stable/config.nims#L94). It also updates release workflows (Nightly and actual releases) to leverage these options.

Next, it adds `arm64` architecture for Mac OSX system to make the released assets run on Apple M1 hardware (cc @alrevuelta)

This also reverts #1759 which was an interim solution to enable compatibility between various CPU architectures.

Lastly, it changes the name of the uploaded binary artifacts to 

```
nwaku-${{matrix.arch}}-${{runner.os}}.tar.gz
```

i.e.

```
nwaku-arm64-macos.tar.gz
nwaku-amd64-linux.tar.gz
```

See the example nightly release here: https://github.com/nwaku-test-org/nwaku/releases/tag/nightly

# Changes

- [x] Add `arm64` architecture for Mac OSX system in the build matrix
- [x] Update the CPU configuration when `native` architecture is disabled (according to Nimbus)
- [x] Remove the default `-march=x86_64` from `Makefile`
- [x] Rename binary release artifact tarball from  `nim-waku-${OS}-latest.tar.gz` to `nwaku-${{arch}}-${{os}}.tar.gz`

## Issue

closes #1760

### DIFF
--- a/config.nims
+++ b/config.nims
@@ -18,19 +18,34 @@ if defined(windows):
   # because these require direct manipulations of the stdout File object.
   switch("define", "chronicles_colors=off")
 
-# This helps especially for 32-bit x86, which sans SSE2 and newer instructions
-# requires quite roundabout code generation for cryptography, and other 64-bit
-# and larger arithmetic use cases, along with register starvation issues. When
-# engineering a more portable binary release, this should be tweaked but still
-# use at least -msse2 or -msse3.
+# https://github.com/status-im/nimbus-eth2/blob/stable/docs/cpu_features.md#ssse3-supplemental-sse3
+# suggests that SHA256 hashing with SSSE3 is 20% faster than without SSSE3, so
+# given its near-ubiquity in the x86 installed base, it renders a distribution
+# build more viable on an overall broader range of hardware.
+#
 if defined(disableMarchNative):
   if defined(i386) or defined(amd64):
     if defined(macosx):
+      # macOS Catalina is EOL as of 2022-09
+      # https://support.apple.com/kb/sp833
+      # "macOS Big Sur - Technical Specifications" lists current oldest
+      # supported models: MacBook (2015 or later), MacBook Air (2013 or later),
+      # MacBook Pro (Late 2013 or later), Mac mini (2014 or later), iMac (2014
+      # or later), iMac Pro (2017 or later), Mac Pro (2013 or later).
+      #
+      # These all have Haswell or newer CPUs.
+      #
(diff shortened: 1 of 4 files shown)


config.nims
```diff
@@ -41,8 +56,9 @@ else:
   if defined(windows):
     # https://gcc.gnu.org/bugzilla/show_bug.cgi?id=65782
     # ("-fno-asynchronous-unwind-tables" breaks Nim's exception raising, sometimes)
-    switch("passC", "-mno-avx512vl")
-    switch("passL", "-mno-avx512vl")
+    switch("passC", "-mno-avx512f")
```
Ivansete-status, line 59: Do we have a way to test this works?

Ivansete-status (approved):
LGTM! Thanks for it!
Just a question whether we could somehow test the build in a Windows machine. Maybe we would need to set up a Windows runner for that ?

config.nims
vpavlin: We are no building for Windows right now, so no, we do not really have a reasonable way to fix this. On the other hand this is "just" to prevent the compiler to enable these abx512f instructions which was the initial reason to look into this. Once we start building for Windows, we'll also want to verify the build artifacts work:) For now it is just a "dumb" copy of best practices from Nimbus:)

jm-clius (approved):
Thanks!

PR #1784, opened 2023-06-07
chore(ci): extend and rename nightly workflow to support RC builds

vpavlin:
# Description
This PR updates Nightly build workflow to also support building and create release candidate (RC) tags and releases.

It also adds `version` (e.g. `nightly`, `v0.18.0-rc.0`...) to artifact filenames.

I also removed the cache as it might make sense to "build from scratch" for releases, but I don't have a strong opinion on this

You can see the example pre-releases here: https://github.com/nwaku-test-org/nwaku/releases

I've also RC release steps to the release process docs

# Changes

- [ ] rename nightly to pre-release workflow
- [ ] add support for `*-rc.X` releases to the workflo
- [ ] add info about RC releases to release process doc


jm-clius (approved):
LGTM. Thanks for improving this process.

rymnc:
Let me check and get back to you @vpavlin

.github/workflows/pre-release.yml
```diff
@@ -28,41 +31,41 @@ jobs:
...
+            update
+
+          make QUICK_AND_DIRTY_COMPILER=1 V=1 CI=false EXPERIMENTAL=${{matrix.experimental}} NIMFLAGS="-d:disableMarchNative --os:${OS} --cpu:${{matrix.arch}}" \
+            wakunode2\
+            chat2\
+            tools
 
-          tar -cvzf ${{steps.vars.outputs.nwaku}} ./build/wakunode2
+          make QUICK_AND_DIRTY_COMPILER=1 V=1 CI=false NIMFLAGS="-d:disableMarchNative --os:${OS} --cpu:${{matrix.arch}}" \
+            wakunode1
```
Ivansete-status, line 65: I think we are almost deprecating the wakunode1.
I wouldn't add it to the release package, just to avoid confusion.

Ivansete-status (approved):
This looks awesome! Thanks for it!
I think it is a good idea to generate RC from scratch.
On the other hand, I wouldn't do anything around `wakunode1` (build & package) as we are going to deprecate it someday soon.

.github/workflows/pre-release.yml
vpavlin: It is currently part of it (https://github.com/waku-org/nwaku/blob/master/.github/workflows/release-assets.yml#L57), so I wanted to make sure we keep things relatively unchanged until the repo refactoring is done

vpavlin:
@rymnc seems like the test passed this time (3rd run) - would be probably good to check if there is a possibility of some kind of race condition (or what other "flaky" CI behaviour could cause this - e.g. if resource limited environment can influence the outcome of the test

.github/workflows/pre-release.yml
Ivansete-status: Ok cool, thanks for letting me know!

PR #1785, opened 2023-06-08
feat(postgres): complete implementation of driver and apply more tests

Ivansete-status:
## Summary
postgres_driver retention policy methods plus more tests

## Changes

- [x] postgres_driver: Completing retention policy methods and corresponding unit tests.
- [x] Addition of 'test_driver_postgres_query.nim' which performs the same validation as in '*sqlite_query.nim' but for PostgreSQL.

## Issue

### DIFF
--- a/waku/v2/waku_archive/driver/postgres_driver/asyncpool.nim
+++ b/waku/v2/waku_archive/driver/postgres_driver/asyncpool.nim
@@ -126,10 +126,10 @@ proc releaseConn(pool: PgAsyncPool, conn: DbConn) =
     if pool.conns[i].dbConn == conn:
       pool.conns[i].busy = false
 
-proc runQuery*(pool: PgAsyncPool,
-               query: string,
-               args: seq[string] = newSeq[string](0)):
-               Future[Result[seq[Row], string]] {.async.} =
+proc query*(pool: PgAsyncPool,
+            query: string,
+            args: seq[string] = newSeq[string](0)):
+            Future[Result[seq[Row], string]] {.async.} =
   ## Runs the SQL query getting results.
   ## Retrieves info from the database.
 
--- a/waku/v2/waku_archive/driver/postgres_driver/asyncpool.nim
+++ b/waku/v2/waku_archive/driver/postgres_driver/asyncpool.nim
@@ -126,11 +126,12 @@ proc releaseConn(pool: PgAsyncPool, conn: DbConn) =
     if pool.conns[i].dbConn == conn:
       pool.conns[i].busy = false
 
-proc query*(pool: PgAsyncPool,
-            query: string,
-            args: seq[string] = newSeq[string](0)):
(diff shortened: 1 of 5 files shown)


waku/v2/waku_archive/driver/postgres_driver/asyncpool.nim
```diff
@@ -126,11 +126,12 @@ proc releaseConn(pool: PgAsyncPool, conn: DbConn) =
     if pool.conns[i].dbConn == conn:
       pool.conns[i].busy = false
 
-proc query*(pool: PgAsyncPool,
-            query: string,
-            args: seq[string] = newSeq[string](0)):
-            Future[Result[seq[Row], string]] {.async.} =
+proc runQuery*(pool: PgAsyncPool,
```
vpavlin, line 129: why the rename?

waku/v2/waku_archive/driver/postgres_driver/postgres_driver.nim
```diff
@@ -234,41 +234,85 @@ method getMessages*(s: PostgresDriver,
...
-                                           limit: int):
-                                           Future[ArchiveDriverResult[void]] {.async.} =
-  return err("not implemented")
+method deleteOldestMessagesNotWithinLimit*(
+                                 s: PostgresDriver,
+                                 limit: int):
+                                 Future[ArchiveDriverResult[void]] {.async.} =
+
+  let execRes = await s.connPool.exec(
+                     """DELETE FROM messages WHERE id NOT IN
```
vpavlin, line 307: Probably not a problem, but this could potentially result in large set of `IN`, not sure if there could be some performance penalty when a lot of messages (thousands?) would need to be deleted. On the other hand, my SQL-fu is weak, so not sure if there is another way:)

vpavlin (approved):

jm-clius (approved):
Thanks!

waku/v2/waku_archive/driver/postgres_driver/asyncpool.nim
Ivansete-status: hehe, the reason was to avoid the next :)

`let rowsRes = await s.connPool.query(query, args)`
and instead have:
`let rowsRes = await s.connPool.runQuery(query, args)`

vpavlin: Hmm..no strong feelings TBH:)

But I am a sucker for consistency, so I checked the SQLite implementation: https://github.com/waku-org/nwaku/blob/master/waku/common/sqlite/database.nim#L223

Ivansete-status: Good point! I'll rename it back to keep it consistent. With less priority though, I'll submit a separate PR to try to enhance that proc name.
Thanks!

Issue #1776, opened 2023-06-01
feat: configure and filter discovered records for static shards


jm-clius:
### Problem

This is a subtask for https://github.com/waku-org/nwaku/issues/1643

[RFC 51](https://rfc.vac.dev/spec/51/#discovery) introduces the idea of static sharding for Waku and specifies how static shards should be encoded in ENRs. A Waku client should be configurable to relay messages on one or more static shards, encode this information in its own advertised ENR and filter discovered peers to only consider those subscribed to the static shards that it's configured for.

https://github.com/waku-org/nwaku/pull/1763 added the following functionality towards achieving this:
- ability to encoding static shards into an ENR, check if the record contains a specific shard and other utilities [see module]( https://github.com/waku-org/nwaku/blob/7df6f4c851ee4994c2e6cc3fc9163ea2d9844024/waku/v2/waku_enr/sharding.nim)
- ability to filter discovered discv5 nodes using a predicate function

However, to finish this feature, the following still needs to be finished:
- the `wakunode2` app should be configurable with static shards
- a node configured for static shards should encode this in its own advertised ENR
- a node configured for static shards should filter its discovered nodes to only consider those connected to one or more of its own configured static shards

### Suggested solution

To achieve the above, I can see at least three self-contained steps, preferably completed in at least three incremental PRs:

#### Step 1: allow static shard configuration on `wakunode2` application

This entails allowing the node to be configured for static shards via the app's `--topic` option.
Once configured for one or more static shards, the app should verify that:
- there are no configured non-static shard topics
- all configured static shards are from the same cluster (see RFC 51)
A node with static shard configuration should then encode this information in its ENR.

#### Step 2: move discv5 discovery loop into `wakunode2` application

In order to sensibly create a predicate function by which to filter discovered nodes, the recurring task of discv5 lookups should move from the node to the application. This in turn would allow us to remove `wakuDiscv5` as a node component.

#### Step 3: create predicate to filter discovered nodes on static shard

Once the discv5 discovery loop is part of the application (and not the underlying node), the application should create a predicate function by which to filter discovered node records (ENRs), if it detects a static shard configuration. This should use the already provided functionality to filter ENRs based on the static shard field.

### Acceptance criteria

- [x] nodes configured with static shard topics advertise this in their ENR according to RFC51
- [ ] discv5 no longer belongs to the `waku_node`, but to the `wakunode2` application
- [x] nodes configured with static shard topics filter discovered nodes according to their own configuration

[...]

PR #1795, opened 2023-06-12
feat: named & sharded topic validation and ENR update

SionoiS:
# Description
Process node config for either static sharding or named topics, verify and update the ENR.

# Changes
- Parse topics from config then check their validity according to https://rfc.vac.dev/spec/51/
- Update ENR with sharded topic info.

Step 1 in closing #1776 

### Feedback

All feedback is welcome but more specifically on;

- ~~Had to make some func public. Could be undone in next step when discovery is re-org. Not sure yet. Otherwise maybe we want to change the ergonomics around RelayShard <-> Record.~~
- ~~Not using parsed topics internally seams inelegant. What do?~~
- ~~Could leave the new code where it is and re-org once discovery is changed. Otherwise, where does this new code fit IDK?~~
- ~~Nim "style", tips & tricks. egg. L522 I tried ```return err("Pubsub topic parsing error: " & topic.error)``` I don't understand why it doesn't work? ParsingError has the concat operator non?~~

### DIFF
--- a/apps/wakunode2/app.nim
+++ b/apps/wakunode2/app.nim
@@ -439,7 +439,8 @@ proc initNode(conf: WakuNodeConf,
         flags = netConfig.wakuFlags.get(),
         multiaddrs = netConfig.enrMultiaddrs,
         rng = rng,
-        discv5Config = discv5Config,
+        conf.topics,
+        discv5Config = discv5Config
       ))
     except CatchableError:
       return err("failed to create waku discv5 instance: " & getCurrentExceptionMsg())
@@ -512,6 +513,7 @@ proc setupProtocols(node: WakuNode, conf: WakuNodeConf,
       pubsubTopics = conf.topicsDeprecated.split(" ")
     else:
       pubsubTopics = conf.topics
+
     try:
       await mountRelay(node, pubsubTopics, peerExchangeHandler = peerExchangeHandler)
     except CatchableError:
--- a/tests/v2/test_waku_discv5.nim
+++ b/tests/v2/test_waku_discv5.nim
@@ -1,11 +1,10 @@
 {.used.}
 
 import
-  std/[sequtils, sets],
+  std/[sequtils],
   stew/results,
   stew/shims/net,
   chronos,
-  chronicles,
   testutils/unittests,
   libp2p/crypto/crypto as libp2p_keys,
   eth/keys as eth_keys
@@ -280,3 +279,33 @@ procSuite "Waku Discovery v5":
 
     # Cleanup
     await allFutures(node1.stop(), node2.stop(), node3.stop(), node4.stop())
+
+  asyncTest "get relayShards from topics":
+    ## Given
+    let mixedTopics = @["/waku/2/thisisatest", "/waku/2/rs/0/2", "/waku/2/rs/0/8"]
+    let shardedTopics = @["/waku/2/rs/0/2", "/waku/2/rs/0/4", "/waku/2/rs/0/8"]
+    let namedTopics = @["/waku/2/thisisatest", "/waku/2/atestthisis", "/waku/2/isthisatest"]
+    let gibberish = @["aedyttydcb/uioasduyio", "jhdfsjhlsdfjhk/sadjhk", "khfsd/hjfdsgjh/dfs"]
+    let empty: seq[string] = @[]
+
+    let relayShards = RelayShards.init(0, @[uint16(2), uint16(4), uint16(8)])
+
+    ## When
+
+    let mixedRes = topicsToRelayShards(mixedTopics)
+    let shardedRes = topicsToRelayShards(shardedTopics)
+    let namedRes = topicsToRelayShards(namedTopics)
+    let gibberishRes = topicsToRelayShards(gibberish)
+    let emptyRes = topicsToRelayShards(empty)
+
+    ## Then
+    assert mixedRes.isErr(), $mixedRes.value
+    assert shardedRes.isOk(), shardedRes.error
+    assert shardedRes.value.isSome()
+    assert shardedRes.value.get() == relayShards, $shardedRes.value.get()
+    assert namedRes.isOk(), namedRes.error
+    assert namedRes.value.isNone(), $namedRes.value
+    assert gibberishRes.isErr(), $gibberishRes.value
+    assert emptyRes.isOk(), emptyRes.error
+    assert emptyRes.value.isNone(), $emptyRes.value
+
--- a/waku/v2/waku_discv5.nim
+++ b/waku/v2/waku_discv5.nim
@@ -48,6 +48,27 @@ type WakuDiscoveryV5* = ref object
     protocol*: protocol.Protocol
     listening*: bool
 
+func topicsToRelayShards*(topics: seq[string]): Result[Option[RelayShards], string] =
+  if topics.len < 1:
+    return ok(none(RelayShards))
+
+  let parsedTopicsRes = topics.mapIt(NsPubsubTopic.parse(it))
+
+  for res in parsedTopicsRes:
+    if res.isErr():
+      return err("failed to parse topic: " & $res.error)
+
+  if parsedTopicsRes.allIt(it.get().kind == NsPubsubTopicKind.NamedSharding):
+    return ok(none(RelayShards))
+
+  if parsedTopicsRes.anyIt(it.get().kind == NsPubsubTopicKind.NamedSharding):
+    return err("use named topics OR sharded ones not both.")
+
+  if parsedTopicsRes.anyIt(it.get().cluster != parsedTopicsRes[0].get().cluster):
+    return err("use sharded topics within the same cluster.")
+
+  return ok(some(RelayShards.init(parsedTopicsRes[0].get().cluster, parsedTopicsRes.mapIt(it.get().shard))))
+
 proc new*(T: type WakuDiscoveryV5, rng: ref HmacDrbgContext, conf: WakuDiscoveryV5Config, record: Option[waku_enr.Record]): T =
   let protocol = newProtocol(
     rng = rng,
@@ -77,8 +98,18 @@ proc new*(T: type WakuDiscoveryV5,
           flags: CapabilitiesBitfield,
           multiaddrs = newSeq[MultiAddress](),
           rng: ref HmacDrbgContext,
-          discv5Config: protocol.DiscoveryConfig = protocol.defaultDiscoveryConfig): T {.
+          topics: seq[string],
+          discv5Config: protocol.DiscoveryConfig = protocol.defaultDiscoveryConfig
+          ): T {.
   deprecated: "use the config and record proc variant instead".}=
+
+  let relayShardsRes = topicsToRelayShards(topics)
+
+  let relayShard =
+    if relayShardsRes.isErr():
+      debug "pubsub topic parsing error", reason = relayShardsRes.error
+      none(RelayShards)
+    else: relayShardsRes.get()
 
   let record = block:
         var builder = EnrBuilder.init(privateKey)
@@ -89,6 +120,15 @@ proc new*(T: type WakuDiscoveryV5,
         )
         builder.withWakuCapabilities(flags)
         builder.withMultiaddrs(multiaddrs)
+
+        if relayShard.isSome():
+          let res = builder.withWakuRelaySharding(relayShard.get())
+
+          if res.isErr():
+            debug "building ENR with relay sharding failed", reason = res.error
+          else:
+            debug "building ENR with relay sharding information", cluster = $relayShard.get().cluster(), shards = $relayShard.get().indices()
+
         builder.build().expect("Record within size limits")
 
   let conf = WakuDiscoveryV5Config(
--- a/apps/networkmonitor/networkmonitor.nim
+++ b/apps/networkmonitor/networkmonitor.nim
@@ -274,7 +274,7 @@ proc initAndStartNode(conf: NetworkMonitorConf): Result[WakuNode, string] =
     node.wakuDiscv5 = WakuDiscoveryV5.new(
         some(extIp), some(nodeTcpPort), some(nodeUdpPort),
         bindIp, nodeUdpPort, discv5BootstrapEnrs, false,
-        keys.PrivateKey(nodeKey.skkey), flags, @[], node.rng)
+        keys.PrivateKey(nodeKey.skkey), flags, @[], node.rng, @[])
 
     node.wakuDiscv5.protocol.open()
     return ok(node)
--- a/examples/v2/publisher.nim
+++ b/examples/v2/publisher.nim
@@ -61,7 +61,9 @@ proc setupAndPublish(rng: ref HmacDrbgContext) {.async.} =
         bootstrapEnrs = @[bootstrapNodeEnr],
         privateKey = keys.PrivateKey(nodeKey.skkey),
         flags = flags,
-        rng = node.rng)
+        rng = node.rng,
+        topics = @[],
+        )
 
     await node.start()
     await node.mountRelay()
--- a/examples/v2/subscriber.nim
+++ b/examples/v2/subscriber.nim
@@ -56,7 +56,9 @@ proc setupAndSubscribe(rng: ref HmacDrbgContext) {.async.} =
         bootstrapEnrs = @[bootstrapNodeEnr],
         privateKey = keys.PrivateKey(nodeKey.skkey),
         flags = flags,
-        rng = node.rng)
+        rng = node.rng,
+        topics = @[],
+        )
 
     await node.start()
     await node.mountRelay()
--- a/tests/v2/test_waku_peer_exchange.nim
+++ b/tests/v2/test_waku_peer_exchange.nim
@@ -110,7 +110,8 @@ procSuite "Waku Peer Exchange":
         keys.PrivateKey(nodeKey1.skkey),
         flags,
         newSeq[MultiAddress](), # Empty multiaddr fields, for now
-        node1.rng
+        node1.rng,
+        newSeq[string]()
       )
 
     node2.wakuDiscv5 = WakuDiscoveryV5.new(
@@ -122,7 +123,8 @@ procSuite "Waku Peer Exchange":
         keys.PrivateKey(nodeKey2.skkey),
         flags,
         newSeq[MultiAddress](), # Empty multiaddr fields, for now
-        node2.rng
+        node2.rng,
+        newSeq[string]()
       )
 
     ## Given


apps/wakunode2/app.nim
```diff
@@ -512,6 +512,46 @@ proc setupProtocols(node: WakuNode, conf: WakuNodeConf,
       pubsubTopics = conf.topicsDeprecated.split(" ")
     else:
       pubsubTopics = conf.topics
+
+    # Validate topics and update a node's ENR accordingly
+    let parsed_topics = pubsubTopics.mapIt(NsPubsubTopic.parse(it))
```
alrevuelta, line 517: minor, we use `camelCase`.
feel free to have a look
https://hackmd.io/@vac/main/%2F5oIVOuQcSM2WKrnDLUPcpg

apps/wakunode2/app.nim
```diff
@@ -512,6 +512,46 @@ proc setupProtocols(node: WakuNode, conf: WakuNodeConf,
...
+      let key = node.wakuDiscv5.protocol.privKey()
+      
+      let pairs = if (let idxList = relayShard.toIndicesList(); idxList.isOk):
+        [toFieldPair(ShardingIndicesListEnrField, idxList.value)]
+      else:
+        let bitVector = relayShard.toBitVector()
+
+        [toFieldPair(ShardingBitVectorEnrField, bitVector)]
+
+      if record.update(key, pairs).isErr():
```
alrevuelta, line 552: wondering if this logic should go in `initNode` or `WakuDiscoveryV5.new`.

`flags = netConfig.wakuFlags.get()` is similar, see [this](https://rfc.vac.dev/spec/31/#waku2-enr-key), so perhaps its a good idea to do all enr field modification in the same place: so far it would be shards + wakuFlags.

apps/wakunode2/app.nim
```diff
@@ -512,6 +512,46 @@ proc setupProtocols(node: WakuNode, conf: WakuNodeConf,
...
+
+    # Validate topics and update a node's ENR accordingly
+    let parsed_topics = pubsubTopics.mapIt(NsPubsubTopic.parse(it))
+
+    var prevTopic: NsPubsubTopic
+    for i, topic in parsed_topics:
+      if topic.isErr():
+        return err("Pubsub topic parsing error")
+      
+      if i == 0:
```
alrevuelta, line 524: mind elaborating on this?

apps/wakunode2/app.nim
alrevuelta, line 552: wondering if there is any other higher level function we can use instead¿

not sure if `withWakuRelayShardingIndicesList` or `withWakuRelaySharding` are meant for this?

perhaps these tests can serve as an example
https://github.com/waku-org/nwaku/blob/0111b88b9b3c13065de6f09844f03c8c0259fcdc/tests/v2/test_waku_enr.nim#L337

alrevuelta (changes requested):
left some comments, thanks!

apps/wakunode2/app.nim
SionoiS: Rust habits... come back when I don't pay attention! good spot.

SionoiS: Any Nim tools to automate this?

apps/wakunode2/app.nim
SionoiS: The discv5 re-org changes a lot of thing. Notably where the ENR will be created. IDK where it fit yet.

SionoiS: Now that I'm changing where discv5 will sit. I will use the EnrBuilder yes.

apps/wakunode2/app.nim
```diff
@@ -512,6 +512,46 @@ proc setupProtocols(node: WakuNode, conf: WakuNodeConf,
...
+        prevTopic = topic.value
+        continue
+      
+      if prevTopic.kind != topic.value.kind:
```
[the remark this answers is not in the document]
jm-clius: Indeed, this would require iterating twice, but given that it improves readability quite a bit, I think it's worth it.
The 99% use case would be nodes not using any static shards. I'd also suggest extracting most of this logic into a separate unit-testable function, perhaps taking the `parsedTopics` as argument and returning a `Result[RelayShards]`. Then you could do something like:

```
if parsedTopics.anyIt(it.kind == NsPubsubTopicKind.StaticSharding):
  let relayShardsRes = parsedTopics.getRelayShards()
  ## error handling here in case result is err
```

The `getRelayShards` func could then check that `allIt` is StaticSharding and of the same cluster before returning the value.

apps/wakunode2/app.nim
jm-clius: Understand there may be some complexity if we don't do the code reorg now, but I'd still prefer (temporarily) extending the constructor of `wakuDiscv5` than exposing the underlying functions.

apps/wakunode2/app.nim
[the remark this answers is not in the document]
SionoiS: Is there a way to handle error when using functional style Nim?
```
if parsedTopicsRes.anyIt(it.isErr()):
    return err(it.error)
```
Something like that?

SionoiS:
I still don't get how to handle error functional style but I'm getting better at this!

jm-clius:
I think I've missed a couple of your queries before. :) I hope the following helps:

>  a way to handle error when using functional style Nim

In this case your `Result` has a `string` error type, so you have to stringify (`$`) the error before you can return it. If you had defined an underlying error type or used an existing one, you could in theory define that as the error `Return` type and simply return the underlying error. Quick check shows this being done in e.g. https://github.com/status-im/nim-chronos/blob/0035f4fa6692e85756aa192b4df84c21d3cacacb/chronos/handles.nim#L81-L104

>  I tried return err("Pubsub topic parsing error: " & topic.error) I don't understand why it doesn't work?

Mmm. Perhaps simply needing to stringify: `$topic.error`?

> Not using parsed topics internally seams inelegant. What do?

Not sure parsed topics should have any meaning internally - it's only of use to the application. From the `waku_node` module's perspective a `pubsubTopic` is a non-semantic string.

I see you've made some changes since my last review - I'll still review these ASAP. :)

apps/wakunode2/app.nim
alrevuelta: > Any Nim tools to automate this? (formatting & style)

no that i know 😢

waku/v2/waku_discv5.nim
```diff
@@ -48,6 +48,36 @@ type WakuDiscoveryV5* = ref object
     protocol*: protocol.Protocol
     listening*: bool
 
+func topicsToRelayShards*(topics: seq[string]): Result[Option[RelayShards], string] =
```
alrevuelta, line 51: perhaps its me but i find this code a bit confusing. there are some special cases i == 0, i > 0, and not sure i like the use of prevTopic.


im for sure missing something so take this as just **pseudocode**. we iterate multiple times but performance here is not an issue, but imho it improves readibility?

```nim
# check errors
if parsedTopicsRes.anyIt(it.isErr()):
    return err(it.error)

# check all are named sharding
if parsedTopicsRes.anyIt(it.get.kind == NsPubsubTopicKind.StaticSharding):
    return error(xxx)

# check all shards belong to the same cluster
if !parsedTopicsRes.allIt(it.get == parsedTopicsRes[0].get.cluster):
   return errors(xxx)

# return
return RelayShards.init(
                 cluster=parsedTopicsRes[0].get.cluster,
                 shards=parsedTopicsRes.mapIt(it.get.shard)

```

alrevuelta (commented):
left some comments, thanks!

waku/v2/waku_discv5.nim
jm-clius: Agree with Alvaro here re readability. And I think it's possible to do only one iteration if you expand your predicate, e.g.:

```
if not parsedTopicsRes.allIt(
  it.isOk() &&
  it.get.kind == NsPubsubTopicKind.StaticSharding &&
  it.get.cluster == parsedTopicsRes[0].get.cluster
  )
  return error(....)
```

If you want to you could even extract the three conditions above into their own predicate function (e.g. `isValid()`), but I think it's readable enough without having to do this. Up to you :)

waku/v2/waku_discv5.nim
```diff
@@ -48,6 +48,36 @@ type WakuDiscoveryV5* = ref object
...
+        return err("use sharded topics within the same cluster.")
+
+    if i > 0:
+      prevTopic = topic.value
+
+  if prevTopic.kind == NsPubsubTopicKind.StaticSharding:
+    let indices = parsedTopicsRes.mapIt(it.value.shard)
+    let relayShards = RelayShards.init(prevTopic.cluster, indices)
+
+    return ok(some(relayShards))
```
jm-clius, line 77: Good idea to have the final `return` be a default and not subject to a condition. In other words, the "happy path" for the function returns the final return statement. Every other return is either an error condition or a quick success.

tests/v2/test_waku_discv5.nim
```diff
@@ -280,3 +280,30 @@ procSuite "Waku Discovery v5":
 
     # Cleanup
     await allFutures(node1.stop(), node2.stop(), node3.stop(), node4.stop())
+
+  asyncTest "get relayShards from topics":
+    ## Given
+    let mixedTopics = @["/waku/2/thisisatest", "/waku/2/rs/0/2", "/waku/2/rs/0/8"]
+    let shardedTopics = @["/waku/2/rs/0/2", "/waku/2/rs/0/4", "/waku/2/rs/0/8"]
+    let namedTopics = @["/waku/2/thisisatest", "/waku/2/atestthisis", "/waku/2/isthisatest"]
+    let gibberish = @["aedyttydcb/uioasduyio", "jhdfsjhlsdfjhk/sadjhk", "khfsd/hjfdsgjh/dfs"]
```
jm-clius, line 289: Perhaps also test an empty sequence?

waku/v2/waku_discv5.nim
```diff
@@ -89,6 +126,15 @@ proc new*(T: type WakuDiscoveryV5,
         )
         builder.withWakuCapabilities(flags)
         builder.withMultiaddrs(multiaddrs)
+
+        if relayShardsRes.isOk():
+          let op = relayShardsRes.value
+          if op.isSome():
+            let shards = relayShardsRes.value.get()
```
jm-clius, line 133:
```suggestion
            let shards = op.get()
```
perhaps?

waku/v2/waku_discv5.nim
```diff
@@ -89,6 +126,15 @@ proc new*(T: type WakuDiscoveryV5,
...
+            let res = builder.withWakuRelaySharding(shards)
```
jm-clius, line 134: I don't necessarily think it's necessary to define `shards` as a separate variable here.
Something like:

```suggestion
            let res = builder.withWakuRelaySharding(op.get())
```

If you want to be more explicit you could rename `op` to e.g. `shardsOp`

tests/v2/test_waku_discv5.nim
```diff
@@ -280,3 +280,30 @@ procSuite "Waku Discovery v5":
...
+
+    let relayShards = RelayShards.init(0, 2, 4, 8)
```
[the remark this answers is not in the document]
SionoiS: I was confused as well the first thing I tried was indeed ```let relayShards = RelayShards.init(0, @[2, 4, 8])```

waku/v2/waku_discv5.nim
```diff
@@ -77,9 +107,16 @@ proc new*(T: type WakuDiscoveryV5,
...
-          discv5Config: protocol.DiscoveryConfig = protocol.defaultDiscoveryConfig): T {.
+          topics: seq[string],
+          discv5Config: protocol.DiscoveryConfig = protocol.defaultDiscoveryConfig
+          ): T {.
   deprecated: "use the config and record proc variant instead".}=
 
+  let relayShardsRes = topicsToRelayShards(topics)
+
+  if relayShardsRes.isErr():
+    debug "pubsub topic parsing error", reason = relayShardsRes.error
```
jm-clius, line 118: May be possible to skip one conditional check later by getting the `shards` option here already.
Something like:
```suggestion
  var relayShardsOp: Option[RelayShard]
  if relayShardsRes.isOk():
    relayShardsOp = relayShardsRes.get()
  else
    debug "pubsub topic parsing error", reason = relayShardsRes.error
    relayShardsOp = none(RelayShard)
```

That way the `record` block below can be a bit cleaner by just checking the option before modifying the builder. Wdyt?

jm-clius (commented):
Thanks! I think this is now much better separated. Some of my comments below are nitpicks, so feel free to address as you see fit.

waku/v2/waku_discv5.nim
SionoiS:
```
if parsedTopicsRes.anyIt(it.isErr()):
    return err(it.error)
```
That is what I wanted to do but seams like you can't use ```it``` outside of scope no?

tests/v2/test_waku_discv5.nim
SionoiS: Good Idea!

waku/v2/waku_discv5.nim
SionoiS: I might rewrite this it look funky but yes your right!

waku/v2/waku_discv5.nim
jm-clius: Where do you mean out of scope? The predicate function could be defined inside the scope of the outer `topicsToRelayShards()` func.

The following should work structurally:

```
proc myOuterFunction()...
  
  func isValidPredicate(someVar: VariableType): bool =
    return someBooleanPredicateCondition

  if not myList.allIt(it.isValidPredicate()):
   return error(...)
```

Not great pseudocode, but hope it illustrates what I mean 😅

Of course, in this instance I don't even think the separate predicate `func` is necessary and you could just add the simple conditions within the `allIt()` argument directly.

waku/v2/waku_discv5.nim
SionoiS: Agree

waku/v2/waku_discv5.nim
SionoiS: In Rust I would do
```
let result: Result<Vec<_>, Error> = topics.into_iter().map(|item| parse(item)).collect;
let parsed_topics = result?;
```
Can't seams to do the same in Nim... How do you return an Error inside a sequence in a functional style?
```
for res in parsedTopics
   if res.IsErr()
      return err(res.error)
```
That works but it's imperative style and looks ugly :P

SionoiS:
I still don't know if there's a way to do this like in Rust lang
```
let seqOfResults: seq[Result[T, Error]]
```
Into
```
let resultOfSeq: Result[seq[T], Error]
```
Might be something I have talk on the Nim forums

waku/v2/waku_discv5.nim
```diff
@@ -77,9 +95,18 @@ proc new*(T: type WakuDiscoveryV5,
...
+          discv5Config: protocol.DiscoveryConfig = protocol.defaultDiscoveryConfig
+          ): T {.
   deprecated: "use the config and record proc variant instead".}=
 
+  let relayShardsRes = topicsToRelayShards(topics)
+
+  let relayShard: Option[RelayShards] = if relayShardsRes.isErr():
+    debug "pubsub topic parsing error", reason = relayShardsRes.error
+    none(RelayShards)
+  else: relayShardsRes.get()
```
jm-clius, line 108: Nitpick:

Perhaps possible to indent the `if` and `else` to be under each other and indented from the `let`.
The following is just a suggestion but may be one way to do it:

```suggestion
  let relayShard =
    if relayShardsRes.isErr():
      debug "pubsub topic parsing error", reason = relayShardsRes.error
      none(RelayShards)
    else: relayShardsRes.get()
```

waku/v2/waku_discv5.nim
```diff
@@ -89,6 +116,12 @@ proc new*(T: type WakuDiscoveryV5,
         )
         builder.withWakuCapabilities(flags)
         builder.withMultiaddrs(multiaddrs)
+
+        if relayShard.isSome():
+          let res = builder.withWakuRelaySharding(relayShard.get())
```
jm-clius, line 121: Perhaps a good idea to add `debug` log here to indicate that we're building ENR with relay shard information?

jm-clius (approved):
Great work! I have some nits below, but it's up to you if you want to address in this or a next PR. I think this incremental PR is ready to be merged, so feel free to unmark as Draft when you agree.

waku/v2/waku_discv5.nim
SionoiS: I agree, your version is more readable.

waku/v2/waku_discv5.nim
SionoiS: Agreed

SionoiS:
I asked @arnetheduck about ```seq[Result[T, Error]]``` => ```Result[seq[T], Error]```
He came up with this
```
proc transposeFirstError[T, E](values: openArray[Result[T, E]]): Result[seq[T], E] =
  let errors = mySeq.filterIt(it.isErr())
  if errors.len>0: err(errors[0]) # or some other aggregation strategy
  else: ok(mySeq.mapIt(it.value))
```
Promising!

Ivansete-status (approved):
That looks awesome from my point of view. Thanks for it!
Just a minor comments :)

alrevuelta (commented):
Before approving, can we get a "proof" that it works end to end? For example, configure some shards via cli arguments, run nwaku and paste here the ENR that is generated. Then would like to see in https://enr-viewer.com/ that the fields are the expected ones.

waku/v2/waku_discv5.nim
```diff
@@ -48,6 +48,24 @@ type WakuDiscoveryV5* = ref object
...
+    if res.isErr():
+      return err("failed to parse topic: " & $res.error)
+
+  if parsedTopicsRes.allIt(it.get().kind == NsPubsubTopicKind.NamedSharding):
+    return ok(none(RelayShards))
+
+  if parsedTopicsRes.anyIt(it.get().kind == NsPubsubTopicKind.NamedSharding):
+    return err("use named topics OR sharded ones not both.")
+
+  if parsedTopicsRes.anyIt(it.get().cluster != parsedTopicsRes[0].get().cluster):
```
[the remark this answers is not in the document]
SionoiS: I believe .allIt returns default true for empty sequence but I guess explicitly checking is more readable.

waku/v2/waku_discv5.nim
```diff
@@ -48,6 +48,24 @@ type WakuDiscoveryV5* = ref object
...
+    return err("use sharded topics within the same cluster.")
+
+  return ok(some(RelayShards.init(parsedTopicsRes[0].get.cluster, parsedTopicsRes.mapIt(it.get().shard))))
```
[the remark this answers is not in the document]
SionoiS: Annoyingly these are not marked as errors. Good catch!

tests/v2/test_waku_discv5.nim
```diff
@@ -280,3 +279,34 @@ procSuite "Waku Discovery v5":
...
+    let mixedRes = topicsToRelayShards(mixedTopics)
+    let shardedRes = topicsToRelayShards(shardedTopics)
+    let namedRes = topicsToRelayShards(namedTopics)
+    let gibberishRes = topicsToRelayShards(gibberish)
+    let emptyRes = topicsToRelayShards(empty)
+
+    ## Then
+    check:
+      mixedRes.isErr()
+      shardedRes.isOk()
```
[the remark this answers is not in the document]
SionoiS: I gather that ```check:``` is more for simpler booleans.

SionoiS:
> Before approving, can we get a "proof" that it works end to end? For example, configure some shards via cli arguments, run nwaku and paste here the ENR that is generated. Then would like to see in https://enr-viewer.com/ that the fields are the expected ones.

```enr:-Le4QFjM3fZPD1QR9YdhwD8KXRHfGJMRAFKtSWk6xj-jB5GkTuC5YKZFe9TqDHTFAwCTM_S1OldCiEsiYY0VRwxjvnYBgmlkgnY0gmlwhMCoAZiKbXVsdGlhZGRyc4CCcnOVAAAJAAIAAwAEAAUABgAHAAgACQAKiXNlY3AyNTZrMaED_vsoLc-A90e9uS29Mmd4SKckuakcBnncVHdeztxFfFGDdGNwgupgg3VkcIIjKIV3YWt1MgE```

LGTM

alrevuelta (approved):
thanks for the ENR, ready to go when CI passes.

PR #1798, opened 2023-06-13
fix(rln-relay): flaky static group manager test

### LINKED ISSUE #1796: chore: test_rln_group_manager_static.nim fails randomly
### Background
Tests fail randomly in PR's.

### Details
When a PR is submitted, the `experimental` workflow failed on MacOS but after re-running the job, it worked.

The issue happened in:
https://github.com/waku-org/nwaku/blob/940206b439f58494887abc0f25b48b9e514980f8/tests/v2/waku_rln_relay/test_rln_group_manager_static.nim#L38

Maybe we could have sth like the next to get a better idea about the error when it happens:

```code
assert rlnInstanceRes.isOk(), rlnInstanceRes.error
```

This is a screenshot showing the test that failed:
image

rymnc:
# Description

Generates a new temp path for the tree db, as it was failing intermittently due to same path being used between tests.

# Changes

- [x] imports std/tempfiles and generates a new temp path

## Issue

closes #1796

### DIFF
--- a/tests/v2/waku_rln_relay/test_rln_group_manager_static.nim
+++ b/tests/v2/waku_rln_relay/test_rln_group_manager_static.nim
@@ -21,6 +21,9 @@ import
   eth/keys,
   discovery/dnsdisc/builder
 
+import
+  std/tempfiles
+
 proc generateCredentials(rlnInstance: ptr RLN): IdentityCredential =
   let credRes = membershipKeyGen(rlnInstance)
   return credRes.get()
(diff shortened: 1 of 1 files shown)


Ivansete-status (approved):
LGTM, thanks !
What does `createTempPath` do under the hood?

rymnc:
https://nim-lang.org/docs/tempfiles.html#genTempPath%2Cstring%2Cstring%2Cstring

PR #1803, opened 2023-06-15
feat(rln-relay): metadata ffi api

rymnc:
# Description

This PR introduces 2 new api's from zerokit, `set_metadata`, and `get_metadata`. They are both used to persist metadata about merkle tree state, like last block synced, etc. This is very useful to resume sync from an existing tree database.

# Changes

- [x] new ffi api's `get_metadata`, `set_metadata`
- [x] Helpers for both
- [x] Tests for both

## Issue

Unblocks https://github.com/waku-org/nwaku/issues/1772


SionoiS (approved):
LGTM

PR #1805, opened 2023-06-16
feat(rln-relay): resume onchain sync from persisted tree db

### LINKED ISSUE #1772: feat(rln-relay): allow passing in a path to the tree db
### Problem
With the integration of pmtree into zerokit and the downstream to nwaku, we should now accept a path to the tree db to avoid onchain resyncs.

### Suggested solution
- A new cmd line arg to nwaku `--rln-relay-tree-db` which accepts an absolute path to the tree db. 
- Validation to ensure that the same db config is used (already handled on the zerokit end)
- If the db doesn't exist at the given path, we need to persist the db there
- Set metadata like `LAST_BLOCK_PROCESSED` into the tree upon each block being processed so that nwaku knows where to resync from.

### Alternatives considered
-

### Acceptance criteria
- [x] Operators are able to pass in the path to an existing db

rymnc:
# Description

This feature makes use of the new ffi api's `set_metadata` and `get_metadata` introduced in #1803. This allows us to insert the block number we sync 
to, and to resume sync when we re-use the database. reduces complexity and improves node startup time *significantly*

# Changes

- [x] Use `setMetadata` in `handleEvents`
- [x] Use `getMetadata` in `init` for the group manager

## Issue

closes #1772

### DIFF
--- a/apps/wakunode2/app.nim
+++ b/apps/wakunode2/app.nim
@@ -552,7 +552,8 @@ proc setupProtocols(node: WakuNode, conf: WakuNodeConf,
         rlnRelayEthAccountPrivateKey: conf.rlnRelayEthAccountPrivateKey,
         rlnRelayEthAccountAddress: conf.rlnRelayEthAccountAddress,
         rlnRelayCredPath: conf.rlnRelayCredPath,
-        rlnRelayCredentialsPassword: conf.rlnRelayCredentialsPassword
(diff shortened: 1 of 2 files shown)


SionoiS (approved):

Ivansete-status (approved):
LGTM! Thanks!

PR #1807, opened 2023-06-16
fix(rln-relay): tree race condition upon initialization

rymnc:
# Description

Resolves the race condition in the tests other than static group manager. Thanks for the report @IvanSete-Status

# Changes

- [ ] Updates tree_path in onchain, waku_rln_relay tests


Ivansete-status (approved):
LGTM! Thanks!

PR #1808, opened 2023-06-16
feat(postgres): integration of postgres in wakunode2

Ivansete-status:
## Summary
This PR is a combination of:
   1. Integration of Postgres in wakunode2.
   2. refactoring of Waku Archive, database, etc.

## Details
First, apologies for that big PR:S. I started the integration of Postgres into wakinode2 but during the process, I ended up doing a more generic refactoring, sorry about that. I can explain everything until we are all happy with it.

The main purpose of this PR is to allow a Nwaku to mount the Waku Archive protocol by using Postgres as the underlying database. For that, run the WakuNode with:

`./build/wakunode2 --store-message-db-url="postgres://postgres:test123@localhost:5432/postgres" ...`

## How to test
As usual, we can run two Nwaku nodes, `a`, and `b`.
`b` will act as the store node and will connect to the Postgres database.

1. Start local Postgres database. Go to the repo's root folder and run the next:
   ```code
   sudo docker compose -f tests/postgres-docker-compose.yml up
   ```
   
   where the `postgres-docker-compose.yml` contains:
   ```code
   version: "3.8"

   services:
     db:
       image: postgres:9.6-alpine
       restart: always
       environment:
         POSTGRES_PASSWORD: test123
       ports:
         - "5432:5432"
   ```
     
2. Start a node that will act as a client
   ```code
   ./build/wakunode2 --config-file=cfg_node_a.txt
   ```   
[cfg_node_a.txt](https://github.com/waku-org/nwaku/files/11772154/cfg_node_a.txt)
   
   
3. Start a node that will act as the "store", which will connect to the Postgres database.
   ```code
      ./build/wakunode2 --config-file=cfg_node_b.txt
   ```   
[cfg_node_b.txt](https://github.com/waku-org/nwaku/files/11772166/cfg_node_b.txt)

4. Send a couple of requests to store two messages.
   The next command sends requests to node b (started in 3.)
   ```code
   curl -d "{\"jsonrpc\":\"2.0\",\"id\":"$(date +%s%N)",\"method\":\"post_waku_v2_relay_v1_message\", \"params\":[\"/waku/2/default-waku/proto\", {\"timestamp\":"$(date +%s%N)", \"payload\":\"VGhpcyBpcyBhIG1lc3NhZ2UK\"}]}" --header "Content-Type: application/json" http://localhost:8546
   ```
   
5. Finally, send a "get_waku_v2_store_v1_messages" request to node `a` (started in 2.)
   ```code
   curl -d '{"jsonrpc":"2.0","id":"id","method":"get_waku_v2_store_v1_messages"}' --header "Content-Type: application/json" http://localhost:8545
   ```   
   And this should bring the two messages stored in point 4.
   
## Issue

### DIFF
--- a/waku/v2/waku_archive/driver/builder.nim
+++ b/waku/v2/waku_archive/driver/builder.nim
@@ -13,11 +13,13 @@ import
   ../../../common/databases/db_sqlite,
   ./sqlite_driver,
   ./sqlite_driver/migrations as archive_driver_sqlite_migrations,
-  ./queue_driver
+  ./queue_driver,
+  ./postgres_driver
 
 export
   sqlite_driver,
-  queue_driver
+  queue_driver,
+  postgres_driver
 
 proc new*(T: type ArchiveDriver,
           url: string,
@@ -68,6 +70,18 @@ proc new*(T: type ArchiveDriver,
       return err("failed to init sqlite archive driver: " & res.error)
 
     return ok(res.get())
+
+  of "postgres":
+    const MaxNumConns = 5 #TODO: we may need to set that from app args (maybe?)
+    let res = PostgresDriver.new(url, MaxNumConns)
+    if res.isErr():
+      return err("failed to init postgres archive driver: " & res.error)
+
+    let driver = res.get()
+    # The table should exist beforehand.
+    discard driver.createMessageTable()
+
+    return ok(driver)
 
   else:
     debug "setting up in-memory waku archive driver"
--- a/waku/v2/waku_archive/driver/builder.nim
+++ b/waku/v2/waku_archive/driver/builder.nim
@@ -6,7 +6,8 @@ else:
 
 import
   stew/results,
-  chronicles
+  chronicles,
+  chronos
 import
   ../driver,
   ../../../common/databases/dburl,
@@ -78,8 +79,14 @@ proc new*(T: type ArchiveDriver,
       return err("failed to init postgres archive driver: " & res.error)
 
     let driver = res.get()
-    # The table should exist beforehand.
-    discard driver.createMessageTable()
+
+    try:
+      # The table should exist beforehand.
+      let newTableRes = waitFor driver.createMessageTable()
+      if newTableRes.isErr():
+        return err("error creating table: " & newTableRes.error)
+    except CatchableError:
+      return err("exception creating table: " & getCurrentExceptionMsg())
 
     return ok(driver)
 
--- a/waku/v2/waku_archive/driver/postgres_driver/postgres_driver.nim
+++ b/waku/v2/waku_archive/driver/postgres_driver/postgres_driver.nim
@@ -238,9 +238,10 @@ proc getInt(s: PostgresDriver,
   if fields.len != 1:
     return err("failed in getRow: Expected one field but got " & $fields.len)
 
-  var retInt: int64
+  var retInt = 0'i64
   try:
-    retInt = parseInt(fields[0])
+    if fields[0] != "":
+      retInt = parseInt(fields[0])
   except ValueError:
     return err("exception in getRow, parseInt: " & getCurrentExceptionMsg())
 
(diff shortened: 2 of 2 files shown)


Ivansete-status:
Thanks for the comments @jm-clius !

Now, in the `app.nim` module, from `master`, we have the next procs:

```code
proc setupDatabaseConnection(dbUrl: string): AppResult[Option[SqliteDatabase]] =

proc gatherSqlitePageStats(db: SqliteDatabase): AppResult[(int64, int64, int64)] =

proc performSqliteVacuum(db: SqliteDatabase): AppResult[void] =
```

Those procs refer to SQLite, which is very specific, so it needs to be moved into a "common/databases" folder. And I'm not keen on just adding other procs for Postgres in `app.nim`.
    
IMO, the only places that should refer to "SQLite" are the `archive.nim` module and the logic around the peer storage feature.

--------

> why it was necessary to move all the archive setup, migration and maintenance logic into the archive module itself?

We currently have the next in the App object (master:)
```code
  App* = object
    version: string
    conf: WakuNodeConf

    rng: ref HmacDrbgContext
    peerStore: Option[WakuPeerStorage]
    archiveDriver: Option[ArchiveDriver]
    archiveRetentionPolicy: Option[RetentionPolicy]
    dynamicBootstrapNodes: seq[RemotePeerInfo]

    node: WakuNode

    rpcServer: Option[RpcHttpServer]
    restServer: Option[RestServerRef]
    metricsServer: Option[MetricsHttpServerRef]
```
Looking at it, we have an archive/store-protocol concept, `archiveDriver` and `archiveRetentionPolicy`, at the same level as `node`. This IMHO, is not correct and especially considering that the `node` itself already has a reference to `wakuArchive` -> `driver`, as shown below:

```code
type
  WakuNode* = ref object
...
    wakuRelay*: WakuRelay
    wakuArchive*: WakuArchive
    wakuStore*: WakuStore

... many other protocols ...

type
  WakuArchive* = ref object
    driver*: ArchiveDriver  # TODO: Make this field private. Remove asterisk
    validator: MessageValidator
    retentionPolicy: RetentionPolicy
```
------
 
> This was specifically removed from the node modules before as the node should be unaware of which archive drivers the application choose to set up and how it manages these. According to me this all belongs to the application. 

I absolutely agree about that. Within the feature branch, the `node` is still unaware of the selected driver with the next considerations:
   - The decision about which driver to create comes from the app.
   - The `WakuNode` only has a reference to `wakuArchive` and therefore doesn't care if the underlying driver is SQLite or Postgres.
   - The responsibility to create the appropriate `ArchiveDriver` falls under the `archive` module because this module is the one that knows about drivers. This module is the father/mother of the drivers. The app knows about config parameters but doesn't speak the "driver language", so it delegates that factory-driver responsibility to that `archive` module.

------
> Of course, if it becomes too large a part of the app module, it could be split into a separate module specifically aimed at how the application sets up its archive drivers. I'd still not extend the node's archive driver modules, though, as it's designed to be agnostic.

IMO, there should be only one place where the drivers are considered. The only module that has interest on drivers is the `archive` one so any driver management outside it is not correct for me.

IMHO, the app module itself should be as simple as possible and it should be just the root of the "tree of responsibility".

I hope that justifies the changes :)

Thanks again for the comment and let's have a quick call if there is sth I'm missing.

Ivansete-status:
The original PR that I raised was too generic. Now, this is simpler as we already could make a proper refactor in previous PRs.

Thanks for your patience!

waku/v2/waku_archive/driver/builder.nim
```diff
@@ -69,6 +71,18 @@ proc new*(T: type ArchiveDriver,
...
 
+  of "postgres":
+    const MaxNumConns = 5 #TODO: we may need to set that from app args (maybe?)
+    let res = PostgresDriver.new(url, MaxNumConns)
+    if res.isErr():
+      return err("failed to init postgres archive driver: " & res.error)
+
+    let driver = res.get()
+    # The table should exist beforehand.
+    discard driver.createMessageTable()
```
alrevuelta, line 82: the table should exist but perhaps its still valuable to check isErr?

alrevuelta (approved):

waku/v2/waku_archive/driver/builder.nim
Ivansete-status: I thought that was not important but now that you mention I'd check it :)
Thanks !

SionoiS (approved):

PR #1811, opened 2023-06-20
refactor(databases): Creation of the databases folder to keep the logic for sqlite and postgres

Ivansete-status:
# Description
Simple refactoring to move all the database logic in one place (common/databases)


waku/common/databases/db_sqlite.nim
```diff
@@ -345,3 +345,141 @@ proc setUserVersion*(database: SqliteDatabase, version: int64): DatabaseResult[v
...
+  # Run the migration scripts
+  for script in scripts:
+
+    for statement in script.breakIntoStatements():
+      debug "executing migration statement", statement=statement
+
+      let execRes = db.query(statement, NoopRowHandler)
+      if execRes.isErr():
+        error "failed to execute migration statement", statement=statement, error=execRes.error
+        return err("failed to execute migration statement")
```
SionoiS, line 464: Question: what happen if it fail in the middle of a migration does it need to be rolled back or no?

waku/common/databases/db_postgres/pgasyncpool.nim
```diff
@@ -0,0 +1,226 @@
...
+    while pool.state == PgAsyncPoolState.Closing:
+      await sleepAsync(0.milliseconds) # Do not block the async runtime
+    return ok()
+
+  pool.state = PgAsyncPoolState.Closing
+
+  # wait for the connections to be released and close them, without
+  # blocking the async runtime
+  if pool.conns.anyIt(it.busy):
+    while pool.conns.anyIt(it.busy):
```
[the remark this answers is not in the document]
Ivansete-status: I think you are absolutely right. Thanks for it! I'll check how it works but is redundant indeed.

waku/common/databases/db_sqlite.nim
Ivansete-status: Good point. We are assuming that either all the query works or all the query fails but we are not considering the case where the query partially worked in which case we should make sure the database is rolled back to the initial state. However, I'd consider that optimization for further PRs. I'd like to keep this PR with minimal changes, i.e. only rename files or delete the `if` statement that you suggested :)

Ivansete-status: Btw, I just said "just rename files" but precisely at this point, I'm adding content from "waku/common/sqlite/migrations.nim", which I removed. If you are not happy with that I can restore back the 'migrations.nim'.

SionoiS: Totally fine to not consider this edge case for now.

SionoiS (approved):

jm-clius (approved):
LGTM! Thanks for your patience :)

PR #1813, opened 2023-06-21
feat: limit relay connections below max conns

alrevuelta:
# Description
* Prunes incoming relay connections exceeding `maxInRelayPeers`
* Does not attempt to connect to more than `maxOutRelayPeers`
* This leaves `maxConnections`-`maxInRelayPeers`-`maxOutRelayPeers` slots for service peers
* Amount of `maxRelayPeers` defaults to 80% of the `maxConnections`

image
[ref](https://excalidraw.com/#json=wmsjdjnZUWn0GKnoLy0mI,BG_ExLzzYByNYxdVBMN9MQ)

closes https://github.com/waku-org/nwaku/issues/1567

### DIFF
--- a/waku/v2/node/peer_manager/peer_manager.nim
+++ b/waku/v2/node/peer_manager/peer_manager.nim
@@ -344,7 +344,7 @@ proc onPeerEvent(pm: PeerManager, peerId: PeerId, event: PeerEvent) {.async.} =
 
 proc new*(T: type PeerManager,
           switch: Switch,
-          maxRelayPeers: int = 50,
+          maxRelayPeers: Option[int],
           storage: PeerStorage = nil,
           initialBackoffInSec = InitialBackoffInSec,
           backoffFactor = BackoffFactor,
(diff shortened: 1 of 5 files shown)


waku/v2/node/peer_manager/peer_manager.nim
```diff
@@ -95,6 +93,37 @@ proc calculateBackoff(initialBackoffInSec: int,
     return chronos.seconds(0)
   return chronos.seconds(initialBackoffInSec*(backoffFactor^(failedAttempts-1)))
 
+# Returns the peerIds of physical connections (in and out)
+# containing at least one stream with the given protocol.
+proc connectedPeers*(pm: PeerManager, protocol: string): (seq[PeerId], seq[PeerId]) =
```
alrevuelta, line 98: no change here, just moving it to the top

nim doesnt like this.
```
proc there() = hi()
proc hi() = echo "hi" <---has to be defined before.
there()
```

waku/v2/node/peer_manager/peer_manager.nim
```diff
@@ -358,16 +359,22 @@ proc new*(T: type PeerManager,
...
-    raise newException(Defect, "Max number of relay peers can't be greater the max amount of connections")
-
-  if maxRelayPeers == maxConnections:
-    warn "Max number of relay peers is equal to max amount of connections, peer wont be contribute to service peers",
-         maxConnections = maxConnections,
-         maxRelayPeers = maxRelayPeers
+  var maxRelayPeersValue = 0
+  if maxRelayPeers.isSome():
+    if maxRelayPeers.get() > maxConnections:
+      error "Max number of relay peers can't be greater the max amount of connections",
```
SionoiS, line 365:
```suggestion
      error "Max number of relay peers can't be greater than the max amount of connections",
```
phrasing

waku/v2/node/peer_manager/peer_manager.nim
```diff
@@ -358,16 +359,22 @@ proc new*(T: type PeerManager,
...
+           maxConnections = maxConnections,
+           maxRelayPeers = maxRelayPeers.get()
+      raise newException(Defect, "Max number of relay peers can't be greater the max amount of connections")
```
SionoiS, line 368:
```suggestion
      raise newException(Defect, "Max number of relay peers can't be greater than the max amount of connections")
```
phrasing

waku/v2/node/peer_manager/peer_manager.nim
```diff
@@ -358,16 +359,22 @@ proc new*(T: type PeerManager,
...
+
+    if maxRelayPeers.get() == maxConnections:
+      warn "Max number of relay peers is equal to max amount of connections, peer wont be contribute to service peers",
```
SionoiS, line 371:
```suggestion
      warn "Max number of relay peers is equal to max amount of connections, peer won't be contributing to service peers",
```
phrasing

SionoiS (approved):
LGTM expect the weird phrasing.

alrevuelta:
> One thing I thought of after reviewing: it seems we only prune relay peers (which makes sense). However, nothing would therefore stop non-relay peers (i.e. client-only peers) from creating inbound connections occupying all slots? Or am I missing something?

@jm-clius Indeed you are right. For this PR I was considering just relay protocol. Will followup with a PR addressing this.

waku/v2/node/peer_manager/peer_manager.nim
alrevuelta: thanks, all addressed c32672fb07ba44606c7cc02f9c32596e9ba2fbcd

PR #1816, opened 2023-06-22
chore: add peer manager config to builder

alrevuelta:
# Description
* Currently the `withPeerManager()` in the builder was not implemented nor used. This PR moves the switch creation from `waku_node` to the `builder` and simplifies `WakuNode.new()` having less parameters.
* This is done because otherwise, `withPeerManger` can't be used since to build the PeerManager, we need the Switch, since its built inside.
* This PR adds a new **unused** flag `max-relay-peers` to be used in future PRs.
* Note that it doesnt introduce any breaking change, unless `maxConnections` is configured <50, but this is not the case in the fleets.
* Prerequisite for https://github.com/waku-org/nwaku/pull/1813

### DIFF
--- a/waku/v2/node/builder.nim
+++ b/waku/v2/node/builder.nim
@@ -180,8 +180,8 @@ proc build*(builder: WakuNodeBuilder): Result[WakuNode, string] =
       peerStoreCapacity = builder.peerStorageCapacity,
       services = @[Service(getAutonatService(rng))],
     )
-  except:
-    return err("failed to create switch")
+  except CatchableError:
+    return err("failed to create switch: " & getCurrentExceptionMsg())
 
   let peerManager = PeerManager.new(
     switch = switch,
--- a/waku/v2/node/builder.nim
+++ b/waku/v2/node/builder.nim
@@ -10,13 +10,15 @@ import
   chronicles,
   libp2p/crypto/crypto,
   libp2p/builders,
-  libp2p/nameresolving/nameresolver
+  libp2p/nameresolving/nameresolver,
+  libp2p/transports/wstransport
 import
   ../waku_enr,
   ../waku_discv5,
   ./config,
   ./peer_manager,
-  ./waku_node
+  ./waku_node,
+  ./waku_switch
 
 
 type
@@ -30,7 +32,9 @@ type
     # Peer storage and peer manager
     peerStorage: Option[PeerStorage]
     peerStorageCapacity: Option[int]
-    peerManager: Option[PeerManager]
+
+    # Peer manager config
+    maxRelayPeers: Option[int]
 
     # Libp2p switch
     switchMaxConnections: Option[int]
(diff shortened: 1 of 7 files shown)


waku/v2/node/builder.nim
```diff
@@ -145,21 +151,45 @@ proc build*(builder: WakuNodeBuilder): Result[WakuNode, string] =
...
-      agentString = builder.switchAgentString,
-      secureKey = builder.switchSslSecureKey.get(""),
-      secureCert = builder.switchSslSecureCert.get(""),
       sendSignedPeerRecord = builder.switchSendSignedPeerRecord.get(false),
+      agentString = builder.switchAgentString,
+      peerStoreCapacity = builder.peerStorageCapacity,
+      services = @[Service(getAutonatService(rng))],
+    )
+  except:
+    return err("failed to create switch")
```
Ivansete-status, line 176:
```suggestion
    return err("failed to create switch: " & getCurrentExceptionMsg())
```

waku/v2/node/builder.nim
```diff
@@ -145,21 +151,45 @@ proc build*(builder: WakuNodeBuilder): Result[WakuNode, string] =
...
       nameResolver = builder.switchNameResolver.get(nil),
-      agentString = builder.switchAgentString,
-      secureKey = builder.switchSslSecureKey.get(""),
-      secureCert = builder.switchSslSecureCert.get(""),
       sendSignedPeerRecord = builder.switchSendSignedPeerRecord.get(false),
+      agentString = builder.switchAgentString,
+      peerStoreCapacity = builder.peerStorageCapacity,
+      services = @[Service(getAutonatService(rng))],
+    )
+  except:
```
Ivansete-status, line 175: I think we need to add a specific catch type to avoid warning :)
```suggestion
  except CatchableError:
```

waku/v2/node/peer_manager/peer_manager.nim
```diff
@@ -347,6 +349,17 @@ proc new*(T: type PeerManager,
          maxConnections = maxConnections
     raise newException(Defect, "Max number of connections can't be greater than PeerManager capacity")
 
+  if maxRelayPeers > maxConnections:
+    error "Max number of relay peers can't be greater the max amount of connections",
+         maxConnections = maxConnections,
+         maxRelayPeers = maxRelayPeers
+    raise newException(Defect, "Max number of relay peers can't be greater the max amount of connections")
```
Ivansete-status, line 356: This comment applies for further possible enhancements:
What do you think about returning a `Result` rather than raising Exception?

Ivansete-status (approved):
Just a minor comments but it looks great to me!
Thanks!

waku/v2/node/builder.nim
alrevuelta: fixed 5bf87652a36b72f5d9429480bdbb538952b187e4

waku/v2/node/builder.nim
alrevuelta: fixed 5bf87652a36b72f5d9429480bdbb538952b187e4

waku/v2/node/peer_manager/peer_manager.nim
alrevuelta: yep you are right. not as part of this pr since this just follows how it was done before but definetly something to consider.

jm-clius (approved):
LGTM!

PR #1817, opened 2023-06-22
refactor(archive): Moving waku archive logic from app.nim to the archive module

Ivansete-status:
# Description

This PR performs a refactoring to simplify the `app.nim`, and move all the _waku archive_ logic into the `archive.nim` module.

The App object shouldn't directly know about the "archive" protocol or the "retention policy". IMHO, these concepts should be managed by the archive module.

This movement is interesting from the perspective of new incoming archive drivers.

This is how I see a natural chain of responsibility:

`app` -> `node` -> `wakuArchive` -> `driver` (sqlite now but postgres in the next PR to be merged)


tests/v2/waku_archive/test_waku_archive.nim
```diff
@@ -467,7 +440,6 @@ procSuite "Waku Archive - find messages":
       response.cursor.isNone()
 
   test "handle temporal history query with a valid time window":
-    ## Given
```
SionoiS, line 470: Why remove this?

waku/v2/node/waku_node.nim
```diff
@@ -507,43 +507,40 @@ proc unsubscribe*(node: WakuNode, pubsubTopic: PubsubTopic, contentTopics: Conte
...
+    try:
+      debug "executing message retention policy"
+      let retPolRes = waitFor node.wakuArchive.executeMessageRetentionPolicy()
+      if retPolRes.isErr():
+        return err("error in mountArchive: " & retPolRes.error)
+    except CatchableError:
+      return err("exception in mountArch-ret-pol: " & getCurrentExceptionMsg())
 
-  discard setTimer(Moment.fromNow(interval), executeRetentionPolicy)
+    node.wakuArchive.startMessageRetentionPolicyPeriodicTask(30.minutes)
```
SionoiS, line 541: Hardcoded 30m? Do we have an issue for making this configurable?

SionoiS (approved):
LGTM

waku/v2/waku_archive/archive.nim
```diff
@@ -70,15 +74,57 @@ type
...
-          retentionPolicy = none(RetentionPolicy)): T =
-  WakuArchive(
-    driver: driver,
-    validator: validator.get(nil),
-    retentionPolicy: retentionPolicy.get(nil)
-  )
+          storeMessageDbUrl: string,
+          storeMessageDbVacuum: bool = false,
+          storeMessageDbMigration: bool = false,
+          storeMessageRetentionPolicy: string = "none"):
```
jm-clius, line 89: I disagree that the Archive Driver should have any knowledge about underlying implementations. An Archive does not even have to be a DB at all - it could simply be an internal memory queue. The point of having a separate archive driver is exactly so that applications can set up and provide their own archive backend to the generic archive driver, with no information about application specific choices (do I have harddrives? do I have my own proprietary DB?) being passed down.

waku/v2/node/waku_node.nim
```diff
@@ -507,43 +507,40 @@ proc unsubscribe*(node: WakuNode, pubsubTopic: PubsubTopic, contentTopics: Conte
 ## Waku archive
 
 proc mountArchive*(node: WakuNode,
-                   driver: Option[ArchiveDriver],
-                   messageValidator: Option[MessageValidator],
-                   retentionPolicy: Option[RetentionPolicy]) =
+                   storeMessageDbUrl: string,
+                   storeMessageDbVacuum: bool = false,
+                   storeMessageDbMigration: bool = false,
+                   storeMessageRetentionPolicy: string = "none"):
```
jm-clius, line 513: The node does not receive a "DB" it receives an Archive as provided by the application - it should not have to have any knowledge about this external component. This may in fact not even be a DB, but simply a chunk of memory.

waku/v2/waku_archive/archive.nim
```diff
@@ -186,25 +231,140 @@ proc findMessages*(w: WakuArchive, query: ArchiveQuery): Future[ArchiveResult] {
...
+    except CatchableError:
+      waku_archive_errors.inc(labelValues = [retPolicyFailure])
+      error "exception in periodic retention policy",
+            error = getCurrentExceptionMsg()
+
+    discard setTimer(Moment.fromNow(interval), executeRetentionPolicy)
+
+  discard setTimer(Moment.fromNow(interval), executeRetentionPolicy)
+
+proc setupWakuArchiveDriver(url: string, vacuum: bool, migrate: bool):
```
jm-clius, line 284: This is application logic: decisions about how the configured url should be interpreted and which specific backend storage systems should be set up for the application lies with the application. When adding new drivers (e.g. someone now supports Codex distributed storage) the point of a driver interface is exactly that the backend operation would not have to be changed or even be aware of this fact. In this case, the application would have to request changes in the setup instructions which couples the driver interface with application-defined underlying implementations.

jm-clius (commented):
Ah! Apologies for not being able to approve this yet. :) Many of the changes seem reasonable, I mainly think my issue here is that the archive driver interface is _designed_ to be generic and to leave the choice and set up of specific driver backends to the application. I agree that we can move the driver-specific setup code to not clutter `app.nim`, but I'd put them in a place close to the application (or at least not in the generic `archive.nim` - which should be seen as similar to an interface). I hope my comments below make sense, but happy to discuss in more detail if perspectives diverge a lot here. :)

jm-clius:
> The App object shouldn't directly know about the "archive" protocol or the "retention policy". 

I disagree. I think _only_ the App should know what e.g. "postgresql" means. Database technologies (and other technologies for storage) are not Waku protocols so a Waku p2p node should not know about the internals here. It simply provides a plugin interface and an application can then configure a service of its choice that works with the plugin.

>  or the "retention policy". 

Mmm. There may indeed be some generic concepts here that can be further abstracted from the application. Thanks for pointing this out! I do think, however, that the "simplest" change here may be to keep the setup code here similarly close to the application and ensure that it works with postgresql as well. But happy to consider other ideas here. :)

> `app` -> `node` -> `wakuArchive` -> `driver`

<-- I think this is true. What shouldn't be added here is that the archive is aware of different driver _implementations_, which moving the implementation-specific setup code into the archive interface would achieve. :)

waku/v2/node/waku_node.nim
Ivansete-status: We don't have an issue for that but I think is a good idea indeed :)

tests/v2/waku_archive/test_waku_archive.nim
Ivansete-status: ah yes, this is totally beyond the scope of this PR. I will set it back

Ivansete-status:
I've re-submitted the PR seeking to follow your suggestions, @jm-clius :) Thanks for them!

waku/v2/node/waku_node.nim
```diff
@@ -491,43 +491,36 @@ proc unsubscribe*(node: WakuNode, pubsubTopic: PubsubTopic, contentTopics: Conte
...
+    try:
+      debug "executing message retention policy"
+      let retPolRes = waitFor node.wakuArchive.executeMessageRetentionPolicy()
+      if retPolRes.isErr():
+        return err("error in mountArchive: " & retPolRes.error)
+    except CatchableError:
+      return err("exception in mountArch-ret-pol: " & getCurrentExceptionMsg())
 
-  discard setTimer(Moment.fromNow(interval), executeRetentionPolicy)
+    node.wakuArchive.startMessageRetentionPolicyPeriodicTask(30.minutes)
```
jm-clius, line 521: Nit: Better to extract this to a `const` just before the function than to make this a "magic value" inline with the code: easier to debug; clear where to add config if needed.

apps/wakunode2/app.nim
```diff
@@ -136,44 +130,14 @@ proc init*(T: type App, rng: ref HmacDrbgContext, conf: WakuNodeConf): T =
...
-  ok(some(connRes.value))
-
-
 ## Peer persistence
 
-const PeerPersistenceDbUrl = "sqlite://peers.db"
-
 proc setupPeerStorage(): AppResult[Option[WakuPeerStorage]] =
-  let db = ?setupDatabaseConnection(PeerPersistenceDbUrl)
+  let db = ? SqliteDatabase.new("peers.db")
```
jm-clius, line 136: Nit: Better to extract this to a `const` just before the function than to make this a "magic value" inline with the code: easier to debug; clear where to add config if needed.

waku/v2/waku_archive/driver/postgres_driver/postgres_driver.nim
```diff
@@ -11,7 +11,7 @@ import
 import
   ../../../waku_core,
   ../../common,
-  ../../driver,
+  ../../driver_base,
```
jm-clius, line 14: Nit: I wouldn't call this `_base`. We didn't previously follow this pattern (see e.g. "node" and "builder"). It's clear enough that the builder for the driver is distinct from the driver. I'd either go for `driver.nim` and `driver_builder` or, preferably, `driver.nim` and `builder.nim` as it's fully qualified in import path: `driver/driver.nim` and `driver/builder.nim`. The latter would be the established pattern.

waku/v2/waku_archive/retention_policy/retention_policy_capacity.nim
```diff
@@ -8,8 +8,8 @@ import
   chronicles,
   chronos
 import
-  ../driver,
-  ../retention_policy
+  ../driver_base,
+  ../retention_policy_base
```
jm-clius, line 12: Nit: I wouldn't call this `_base`. See my comment elsewhere on the established pattern here.

jm-clius (approved):
Thanks! I think this is a much better pattern than we had before. Thanks for your patience and taking the time to clean this up. Note my comments/nitpicks below. One comment is that this signifies a significant number of changes (also for SQLite DBs) which we'll only be able to fully test in a production environment. When rolling this out in a next release, I'd suggest (1) testing basic functions in collaboration with go-waku and (2) deploying only to a few production nodes at first.

jakubgs:
Looks like this resulted in all nodes on `wakuv2.test` restarting due to:
```
admin@node-01.do-ams3.wakuv2.test:~ % d run --rm -it statusteam/nim-waku:deploy-wakuv2-test --version
could not load: libpq.so(.5|)
(compile with -d:nimDebugDlOpen for more information)
```
So I've made a PR to sanity-check the Docker image startup:

* https://github.com/waku-org/nwaku/pull/1825

PR #1827, opened 2023-06-27
fix(docker): Adding 'libpq-dev' to the resulting docker image

Ivansete-status:
# Description
This is needed to avoid the `wakunode2` app not starting due to lack of the 'libpq.so.5' library.

## Issue


jakubgs (approved):
Thanks.

PR #1830, opened 2023-06-28
fix(rln-relay): chunk event fetching

### LINKED ISSUE #1773: fix(rln-relay): chunk event fetching
### Problem
With a large difference in the block numbers between which we have to fetch all the `MemberRegistered` events, there is a chance that the rpc provider would reject the requests made, simply because it is too large a range.

### Impact
`medium`

### To reproduce
- (cannot repro, since this issue will exist on mainnet)

### Expected behavior
- Chunk block syncs (10 blocks per fetch)

### Additional context
- Follow how nimbus did it here - https://github.com/status-im/nimbus-eth2/blob/55ae7d3b00fcb7029ced737eef8868690cbaa756/beacon_chain/eth1/eth1_monitor.nim#L1852

Thanks for the report: @weboko @richard-ramos (ref: https://github.com/waku-org/js-rln/pull/57#discussion_r1179789824)

rymnc:
# Description

This PR introduces "chunking" of events that are being fetched from the rpc provider. This allows using resource restrained rpc providers, 
abiding by their block request limits.

# Changes

- [ ] Checks if the latest block > BlockChunkSize (100), and chunks events till then
- [ ] if not, it gets all the events in one request

## Issue

closes #1773


SionoiS (approved):
LGTM!

PR #1832, opened 2023-06-28
refactor(internalconfig): avoid passing a 'WakuNodeConf' item to the 'createRecord' proc

Ivansete-status:
# Description
This change is needed so that the `libwaku.nim` code gets simpler. The `libwaku.nim` can't import the 'WakuNodeConf' type because this is related to the wakunode2 app, and the library shouldn't know about it.


SionoiS (approved):
If you want to change the location of ```createRecord``` I'm all for it. In retrospect it should not have been in ```app```

Ivansete-status:
> If you want to change the location of `createRecord` too, I'm all for it. In retrospect it should not have been in `app`

Thanks for the review!
For me is fine as is it now but of course feel free to submit a refactor that you consider interesting

PR #1844, opened 2023-07-06
chore(rln-relay): Verify proofs based on bandwidth usage

### LINKED ISSUE #1837: chore(rln-relay): Verify proofs based on bandwidth usage
### Background
Verifying each proof that is sent may not be possible on resource-restricted nodes, and should be done only when bandwidth exceeds a set figure.

### Details
Check the libp2p-bandwidth metric, and then verify the proof if the bandwidth exceeds the set limit.

### Acceptance criteria
- [x] Proofs are verified when bandwidth exceeds an operator-set amount.

rymnc:
# Description

Runs rln verification only if bandwidth exceeds 1mbps. @jm-clius, what is the bandwidth cutoff you'd like to be the default?

# Changes

- [x] Uses TokenBucket from chronos/ratelimit to check if bandwidth has been exceeded

## Issue

closes #1837

### DIFF
--- a/apps/wakunode2/external_config.nim
+++ b/apps/wakunode2/external_config.nim
@@ -214,7 +214,7 @@ type
     rlnRelayBandwidthThreshold* {.
       desc: "Message rate in bytes/sec after which verification of proofs should happen",
       defaultValue: 0 # to maintain backwards compatibility
-      name: "rln-relay-bandwidth-cutff" }: int
+      name: "rln-relay-bandwidth-cutoff" }: int
 
     staticnodes* {.
       desc: "Peer multiaddr to directly connect with. Argument may be repeated."
--- a/apps/wakunode2/external_config.nim
+++ b/apps/wakunode2/external_config.nim
@@ -214,7 +214,7 @@ type
     rlnRelayBandwidthThreshold* {.
       desc: "Message rate in bytes/sec after which verification of proofs should happen",
       defaultValue: 0 # to maintain backwards compatibility
-      name: "rln-relay-bandwidth-cutoff" }: int
+      name: "rln-relay-bandwidth-threshold" }: int
 
     staticnodes* {.
       desc: "Peer multiaddr to directly connect with. Argument may be repeated."
--- a/apps/wakunode2/external_config.nim
+++ b/apps/wakunode2/external_config.nim
@@ -213,7 +213,7 @@ type
 
     rlnRelayBandwidthThreshold* {.
       desc: "Message rate in bytes/sec after which verification of proofs should happen",
-      defaultValue: 1_000_000
+      defaultValue: 0 # to maintain backwards compatibility
       name: "rln-relay-bandwidth-cutff" }: int
 
     staticnodes* {.
--- a/apps/wakunode2/external_config.nim
+++ b/apps/wakunode2/external_config.nim
@@ -211,6 +211,11 @@ type
       defaultValue: ""
       name: "rln-relay-tree-path" }: string
 
+    rlnRelayBandwidthThreshold* {.
+      desc: "Message rate in bytes/sec after which verification of proofs should happen",
+      defaultValue: 1_000_000
+      name: "rln-relay-bandwidth-cutff" }: int
+
     staticnodes* {.
       desc: "Peer multiaddr to directly connect with. Argument may be repeated."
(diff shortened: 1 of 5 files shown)


SionoiS (approved):
nitpick:
At first, I got confused  by the word cutoff. I thought it was stopping the rate limiting then I read the code.

Call it threshold maybe?

Otherwise :+1:

rymnc:
Thanks @SionoiS and @jm-clius, addressed both concerns in [9974535](https://github.com/waku-org/nwaku/pull/1844/commits/99745351983b15000d83ad7cfab41469e4143f50)

alrevuelta (approved):
great! lgtm.

perhaps we are missing a simple unit test? since a missconfoguration of this feature can totally "short circuit" rln, totally disabling it. Even if it works now, its always nice to protect the code from future changes.

on the other hand (no action required) i usually prefer when things **always** works or **never** works. this introduces a *sometimes* works.

rymnc:
Thanks @alrevuelta, addressed in [a633f30](https://github.com/waku-org/nwaku/pull/1844/commits/a633f3071d5f6fc8d73678e5815de278e9e42e3a) and [63fa8dd](https://github.com/waku-org/nwaku/pull/1844/commits/63fa8dd271c120b037cb7f26f2a19fa37555fbbf)

apps/wakunode2/external_config.nim
```diff
@@ -211,6 +211,11 @@ type
       defaultValue: ""
       name: "rln-relay-tree-path" }: string
 
+    rlnRelayBandwidthThreshold* {.
```
Ivansete-status, line 214: From a user perspective, how could I set the proper value of this setting?

Ivansete-status (approved):
LGTM! :clap:

apps/wakunode2/external_config.nim
rymnc: Hmm, it's mentioned bytes/sec, so it should be according to their personal bandwidth constraints.

alrevuelta: im fine with having it as a cli flag, but imho this parameter should be fixed for the network and not "according to their personal bandwidth constraints". wouldn't this be dangerous?

if different nodes in the same network set different thresholds, then message validity (valid/invalid) wont be an objective truth, but something that will depend on each node (each node will have its own view on wether the message is valid or invalid). and for waku as a network, i thinnk this can cause troubles.

rymnc: I think this problem can be solved if we broadcast capabilities like waku_relay, however if you'd like to set it to a global value, it'll be a matter of prioritising resource restrained nodes vs an overall global stable network. wdyt?

alrevuelta: just to ensure we are on the same page:
* do we agree that these value has to be fixed for the whole network because of the above reasons?
* mind elaborating on "it'll be a matter of prioritising resource restrained nodes vs an overall global stable network"?

rymnc: 1. yes, agree that the value has to be fixed for the whole network
2. follow up comment

rymnc: the network will be stable if we set a fixed bandwidth threshold for all nodes, however, some nodes which have a bandwidth lesser than that will blindly forward messages to other nodes. and if we implement peer scoring for rln, this opens another issue where we need to adjust peer score based on the other's bandwidth threshold

waku/v2/waku_rln_relay/rln_relay.nim
```diff
@@ -286,6 +291,18 @@ proc generateRlnValidator*(wakuRlnRelay: WakuRLNRelay,
...
     trace "rln-relay topic validator is called"
+
+    ## Check if enough tokens can be consumed from the message bucket
+    try:
+      if wakuRlnRelay.messageBucket.isSome() and
+         wakuRlnRelay.messageBucket.get().tryConsume(message.data.len):
+        return pubsub.ValidationResult.Accept
+      else:
+        info "message bandwidth limit exceeded, running rate limit proof validation"
+    except OverflowDefect: # not a problem
```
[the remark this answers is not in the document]
rymnc: yeah, if you pass in 0 to the TokenBucket, it underflows. sorry for not creating an issue!

PR #1845, opened 2023-07-06
refactor(cbindings): exposing a `libwaku` library with raw C types

Ivansete-status:
# Description
We are using only raw `C` types with the aim of exposing the `libwaku` in a more idiomatic way.

# Changes

- Start to git-control the `libwaku.h` and force C types in there. Avoiding Nim types.
- The `libwaku` users can pass information from outside with a `char *`.
- The `libwaku` can retrieve information outside by means of callbacks (`void (*WakuCallBack) (char* msg, size_t len_0);`).

## Issue


jm-clius (approved):
LGTM with the raw types and callbacks. Thanks for your effort here.

richard-ramos (approved):
:rocket: very cool. I like the callback idea for handling success/error responses

examples/cbindings/waku_example.c
```diff
@@ -11,22 +10,26 @@
...
-static NimStringDesc wakuString;
-NimStringDesc* mResp = &wakuString;
+#define WAKU_CALL(call)                                                        \
+do {                                                                           \
+  int ret = call;                                                              \
+  if (ret != 0) {                                                              \
+    printf("Failed the call to: %s. Returned code: %d\n", #call, ret);         \
+    exit(1);                                                                   \
+  }                                                                            \
+} while (0)
```
richard-ramos, line 23: I'm curious on why is the `do while` needed here :thinking:

Ivansete-status: Thanks for the review!

I agree that this seems pointless but its main purpose is to make sure the macro always behaves as it is expected to behave.

We have good examples in the next:
https://hownot2code.wordpress.com/2016/12/05/do-while-0-in-macros/

PR #1852, opened 2023-07-15
test(rln-relay): rpc handler to support waku rln relay

rymnc:
WIP


alrevuelta:
should we modify the rpc endpoint `post_waku_v2_relay_v1_message` to include something like `useRln`? The behaviour would be:
* If `useRln=true` but rln is not enabled in the node, return error.
* If `useRln=true` and rln is enabled in the node, publish the msg including an rln proof
* If `useRln=false` do not include the rln proof (legacy behaviour)

would also require changing the [spec](https://rfc.vac.dev/spec/16/)

vpavlin:
Would this introduce unnecessary level of complexity? As in if the node has RLN membership info configured and RLN relay enabled, always attach the proof, otherwise, do not

Is there particular reason/need for not sending proof when RLN is enabled on the node?

rymnc:
I think it's to maintain slight backwards compatibility

alrevuelta:
@vpavlin yeap agree with the complexity here. The main reason for adding this complexity would be to have always the same input, same output. Without the `useRln` we have different behavours for the same input.

But well, as an advocate of simplicity im actually leaning towards your suggestion. Same api call but proof is attached or not depending on how the node is configured.

jm-clius:
Unless there's some hidden complexity, my suggestion would be to leave the API untouched and simply attach proof based on whether node is RLN-configured/compiled or not.

Ivansete-status:
I agree with @vpavlin's suggestion.
On the other hand, should we do that in the REST handler? I thought we were moving out from rpc and the idea is to only have REST.

alrevuelta:
@rymnc Seems we have quorum. Mind rebasing and implementing the changes? Ideally we can have it for the REST api but I'm fine with leaving it for another PR.

Thanks!

alrevuelta (approved):

waku/node/jsonrpc/relay/handlers.nim
```diff
@@ -77,14 +84,21 @@ proc installRelayApiHandlers*(node: WakuNode, server: RpcServer, cache: MessageC
...
         timestamp: msg.timestamp.get(Timestamp(0)),
         ephemeral: msg.ephemeral.get(false)
       )
+    
+    when defined(rln):
+      if not node.wakuRlnRelay.isNil():
+        let success = node.wakuRlnRelay.appendRLNProof(message, 
+                                                      float64(getTime().toUnix()))
+        if not success:
+          raise newException(ValueError, "Failed to append RLN proof to message")
```
vpavlin, line 101: Will the `appendRLNProof` print out any more specific error log? This seems to be too ambiguous for users to know what is going on:)

vpavlin (approved):
lgtm, just one comment/question

waku/node/jsonrpc/relay/handlers.nim
rymnc: I agree, we're working on it on the zerokit side where we return an error enum atleast so the error code is known

Issue #1846, opened 2023-07-06
[Milestone]  Autosharding v1


SionoiS:
**Planned start date**: 28 Jun 2023
**Due date**: 18 August 2023

# Summary
 
Design and implementation of autosharding. Autosharding is the automatic assignment of content topics to shards.

# Acceptance Criteria

 - The autosharding function distribute topics to shards.
 - The ENR of node using autoshard is created/updated with the shards info.
 - Nodes can be configured for content topics and/or pubsub topics.
 - The FILTER protocol code has been updated for autosharding.
 - The LIGHTPUSH protocol code has been updated for autosharding.
 - The [RFC](https://rfc.vac.dev/spec/51/#automatic-sharding) has been updated with the [new specifications](https://github.com/vacp2p/rfc/tree/feat-autosharding/content/docs/rfcs/51#automatic-sharding).

## Tasks
- [x] #1854 
- [x] #1856
- [ ] #1857 

# RAID (Risks, Assumptions, Issues and Dependencies)

Autosharding will affect various other protocols and may have unintended consequences.

https://github.com/waku-org/nwaku/issues/1846#issuecomment-1663651873

chaitanyaprem:
@SionoiS , @alrevuelta  Was going through peer manager code and noticed that there seems to be a limit of 1 connection towards a peer as per libp2p. 
I am thinking we will have to improve peer management to prune/manage connections towards peers considering shards as well. Basically a more intelligent pruning.
Otherwise a node may end up not having connections to a shard at all as connection limit is reached.
There are 2 ways to address this problem: 

1. Open a separate connection towards same peer for each shard. This would be simpler, as we could just enforce peer/connectionLimits per shard. But i have noticed that in nwaku there is a setting to limit 1 connection per peer. https://github.com/waku-org/nwaku/blob/2fc48842b7747ccc372a1e4ad09dc632c8db218e/waku/v2/node/waku_switch.nim#L20
2. Another alternative is share the single connection between peers for communicating across multiple shards. In this case we may have to ensure that a min relayConnections are maintained per shard the node is interested in. Not sure how we would decide percentage of peers to be maintained per shard though.

Thoughts.

SionoiS:
That is a good point! @chaitanyaprem In theory It's possible, my question would be how often does that happen in practice?

I don't think we want multiple connections per peer. Although, I'm not as familiar with this part of the code than @alrevuelta 

I'll have to investigate further.

SionoiS:
[...]

PR #1856, opened 2023-07-19
feat: autosharding content topic in config

SionoiS:
# Description

Added content topics in waku external config (pubsub VS content is now explicit). When used, a shard is picked and ENR is updated. Pubsub topics can also be used at the same time but only on the same cluster.

~~I also remove the default pubsub topic from config. Do we want to keep it until autoshard is working properly?~~

Let's not forget to deprecate the default pubsub topic when autosharding is :+1: 

# Changes

- [x] Added pubsub & content topic to external config.
- [x] Update ENR based on all topics. 

Tracking #1846

### DIFF
--- a/apps/wakunode2/external_config.nim
+++ b/apps/wakunode2/external_config.nim
@@ -227,6 +227,7 @@ type
 
     pubsubTopics* {.
       desc: "Default pubsub topic to subscribe to. Argument may be repeated."
+      defaultValue: @["/waku/2/default-waku/proto"]
       name: "pubsub-topic" .}: seq[string]
 
     contentTopics* {.
--- a/apps/wakunode2/external_config.nim
+++ b/apps/wakunode2/external_config.nim
@@ -225,10 +225,13 @@ type
       defaultValue: false
       name: "keep-alive" }: bool
 
-    topics* {.
-      desc: "Default topic to subscribe to. Argument may be repeated."
-      defaultValue: @["/waku/2/default-waku/proto"]
-      name: "topic" .}: seq[string]
+    pubsubTopics* {.
+      desc: "Default pubsub topic to subscribe to. Argument may be repeated."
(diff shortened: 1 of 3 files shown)


vpavlin (approved):
LGTM

I feel like we may not need to keep the default, but we need to make sure all the docs/how-tos are updated and we mention it in release notes, from this point, it might be easier to keep the default topic until we switch fully to autosharding.

apps/wakunode2/external_config.nim
```diff
@@ -225,10 +225,13 @@ type
       defaultValue: false
       name: "keep-alive" }: bool
 
-    topics* {.
-      desc: "Default topic to subscribe to. Argument may be repeated."
-      defaultValue: @["/waku/2/default-waku/proto"]
-      name: "topic" .}: seq[string]
+    pubsubTopics* {.
+      desc: "Default pubsub topic to subscribe to. Argument may be repeated."
+      name: "pubsub-topic" .}: seq[string]
```
jm-clius, line 230: I wouldn't remove the default here for now. For some applications this will be a breaking change (for example, I think DappNode may assume subscription to default topic).

jm-clius (approved):
LGTM. Note comment on keeping default pubsub topic for now. We should announce deprecation in following release in upcoming release notes.

SionoiS:
I restored the default pubsub topic.

Where would we track deprecation notices for the release notes @vpavlin ?

vpavlin:
I am thinking we should have a label for things we want to have explicitely mentioned in release notes - WDYT?

SionoiS:
> I am thinking we should have a label for things we want to have explicitely mentioned in release notes - WDYT?

Yes that's good. I'll edit the top and mention the deprecation.

jakubgs:
Please communicate infra breaking changes to the infra team before merging.

PR #1857, opened 2023-07-21
feat: update various protocols to autoshard

SionoiS:
# Description

Updating LIGHTPUSH & FILTER protocols to handle autosharding.

# Changes

- [x] LIGHTPUSH  take optional pubsub topics and compute shards.
- [x] FITLER  take optional pubsub topics and compute shards.

Tracking #1846


SionoiS:
I'm not sure how to change filter to handle multiple shards. The code ATM assumes one pubsub and multiple content topics.

waku/v2/node/waku_node.nim
```diff
@@ -601,11 +601,28 @@ proc mountLightPush*(node: WakuNode) {.async.} =
...
   if node.wakuRelay.isNil():
     debug "mounting lightpush without relay (nil)"
-    pushHandler = proc(peer: PeerId, pubsubTopic: string, message: WakuMessage): Future[WakuLightPushResult[void]] {.async.} =
+    pushHandler = proc(peer: PeerId, pubsubTopic: Option[string], message: WakuMessage): Future[WakuLightPushResult[void]] {.async.} =
       return err("no waku relay found")
   else:
-    pushHandler = proc(peer: PeerId, pubsubTopic: string, message: WakuMessage): Future[WakuLightPushResult[void]] {.async.} =
-      discard await node.wakuRelay.publish(pubsubTopic, message.encode().buffer)
+    pushHandler = proc(peer: PeerId, pubsubTopic: Option[string], message: WakuMessage): Future[WakuLightPushResult[void]] {.async.} =
+      let topic =
```
vpavlin, line 608: Should this code be extracted somewhere into a proc? It seems exact same code repeats in `waku_node.nim` and `waku_filter/client.nim` - not a fan of premature optimization, but this seems like a generally useful piece of code 

`pubsubTopicOrShard(pubsubTopic: Option[string], contentTopic: string): string` or something like this?

SionoiS: Good point. It's already used in 5 places.

SionoiS:
Added a proc for optional autosharding, changed the cluster index and count for gen0 and added back ```topics``` in config with a deprecation note.

SionoiS:
@jakubgs When we merge this to master the config will be back to normal. ```--topic``` will be back as deprecated with ```--pubsub-topic``` and ```--content-topic``` present.

waku/v2/waku_filter/rpc.nim
```diff
@@ -15,7 +15,7 @@ type
 
   FilterRequest* = object
     contentFilters*: seq[ContentFilter]
-    pubsubTopic*: string
+    pubsubTopic*: Option[string]
```
jm-clius, line 18: Not sure I agree with where these changes are made. Let me explain:
This essentially changes the underlying protocol to have pubsubTopic/shard as optional, which would mean the _protocol_ would have to become aware of autosharding and mapping rather than this simply being a running use case for the application. In this case, for example, I think the filter service node would have to be able to receive a FilterRequest without `pubsubTopic` and somehow "know" about autosharding, even though it's most probably unaware of the application(s) it's servicing.
I think the "cleaner" (and simpler) way to introduce autosharding is to simply change the most external API (the one used by applications) to fill out shards/pubsubTopics where it's necessary but not populated by the application. This way the protocols remain untouched and still works as before.

jm-clius (changes requested):
Made a comment below re `filter`, but it applies to all the protocols: I think we don't want the underlying protocols to change, just provide _applications_ with some "middleware" logic when they interact with the node that populates shards in API calls to the underlying protocols. IMO protocols should remain completely unaware of autosharding, which occurs at a higher layer.

SionoiS:
> Made a comment below re `filter`, but it applies to all the protocols: I think we don't want the underlying protocols to change, just provide _applications_ with some "middleware" logic when they interact with the node that populates shards in API calls to the underlying protocols. IMO protocols should remain completely unaware of autosharding, which occurs at a higher layer.

~~The protocols are not the code in nwaku. They are in the RFCs and for [FILTER](https://rfc.vac.dev/spec/12/#protobuf) for example the pubsub topic is optional. They were not following the protocol, now they do.~~

I totally miss this part ```If the request contains filter criteria, it MUST contain a pubsub_topic and the content_topics set MUST NOT be empty.```

As for where we decide to compute the shard, my idea was that light node should have the least amount of computing to do.

I'll make the changes.

waku/v2/waku_filter_v2/client.nim
```diff
@@ -77,25 +80,71 @@ proc ping*(wfc: WakuFilterClient, servicePeer: RemotePeerInfo): Future[FilterSub
...
-proc unsubscribe*(wfc: WakuFilterClient, servicePeer: RemotePeerInfo, pubsubTopic: PubsubTopic, contentTopics: seq[ContentTopic]): Future[FilterSubscribeResult] {.async.} =
-  let requestId = generateRequestId(wfc.rng)
-  let filterSubscribeRequest = FilterSubscribeRequest.unsubscribe(
-    requestId = requestId,
-    pubsubTopic = pubsubTopic,
-    contentTopics = contentTopics
-  )
-
-  return await wfc.sendSubscribeRequest(servicePeer, filterSubscribeRequest)
+proc subscribe*(wfc: WakuFilterClient, servicePeer: RemotePeerInfo, pubsubTopic: Option[PubsubTopic], contentTopics: seq[ContentTopic]): Result[seq[Future[FilterSubscribeResult]], FilterSubscribeError] =
```
SionoiS, line 83: I'm not sure I like this return type but would returning only Futures be better?

waku/v2/waku_filter_v2/client.nim
```diff
@@ -77,25 +80,71 @@ proc ping*(wfc: WakuFilterClient, servicePeer: RemotePeerInfo): Future[FilterSub
...
+        requestId = requestId,
+        pubsubTopic = pubsubTopic,
+        contentTopics = contentTopics
+      )
+
+      wfc.sendSubscribeRequest(servicePeer, filterSubscribeRequest)
+
+  ok(futures)
+  
+proc unsubscribe*(wfc: WakuFilterClient, servicePeer: RemotePeerInfo, pubsubTopic: Option[PubsubTopic], contentTopics: seq[ContentTopic]): Result[seq[Future[FilterSubscribeResult]], FilterSubscribeError] =
```
SionoiS, line 116: Same here

waku/v2/waku_lightpush/client.nim
```diff
@@ -66,6 +66,14 @@ proc sendPushRequest(wl: WakuLightPushClient, req: PushRequest, peer: PeerId|Rem
 
   return ok()
 
-proc publish*(wl: WakuLightPushClient, pubsubTopic: PubsubTopic, message: WakuMessage, peer: PeerId|RemotePeerInfo): Future[WakuLightPushResult[void]] {.async, gcsafe.} =
-  let pushRequest = PushRequest(pubsubTopic: pubsubTopic, message: message)
+proc publish*(wl: WakuLightPushClient, pubsubTopic: Option[PubsubTopic], message: WakuMessage, peer: PeerId|RemotePeerInfo): Future[WakuLightPushResult[void]] {.async, gcsafe.} =
```
SionoiS, line 69: Like this???

waku/v2/waku_store/client.nim
```diff
@@ -81,7 +84,40 @@ proc sendHistoryQueryRPC(w: WakuStoreClient, req: HistoryQuery, peer: RemotePeer
 proc query*(w: WakuStoreClient, req: HistoryQuery, peer: RemotePeerInfo): Future[HistoryResult] {.async, gcsafe.} =
   return await w.sendHistoryQueryRPC(req, peer)
 
+proc autosharding_query*(w: WakuStoreClient, req: HistoryQuery, peer: RemotePeerInfo): Result[seq[Future[HistoryResult]], HistoryError]  =
```
SionoiS, line 87: This is weird too.

SionoiS:
It's still a bit weird how you have to send multiple request in the case of the content topic not using the same shards.

apps/wakunode2/external_config.nim
```diff
@@ -225,9 +225,13 @@ type
       defaultValue: false
       name: "keep-alive" }: bool
 
+    topics* {.
```
alrevuelta, line 228: No action required, just a thought.

afaik its agreed to remove `topics` (which is pubsub topic) in favour of `content-topic`, since we want to abstract the user away from that.

so perhaps this change is not neccesary? i mean, why renaming `topics` to `pubsub-topic` if we are removing in it soon?

examples/v2/filter_subscriber.nim
```diff
@@ -22,7 +23,7 @@ proc unsubscribe(wfc: WakuFilterClient,
                  filterPubsubTopic: PubsubTopic,
                  filterContentTopic: ContentTopic) {.async.} =
   notice "unsubscribing from filter"
-  let unsubscribeRes = await wfc.unsubscribe(filterPeer, filterPubsubTopic, @[filterContentTopic])
+  let unsubscribeRes = await wfc.unsubscribe(filterPeer, some(filterPubsubTopic), @[filterContentTopic])
```
alrevuelta, line 26: imho pubsub topic should be derived directly from content topic.
* simpler
* less error prone (what if the provided pubsub topic doesnt match the one derived from content topic? it introduces a new not neccesary error path)

in other words

unsubscribe(filterPeer, @[filterContentTopic])

examples/v2/filter_subscriber.nim
```diff
@@ -47,7 +48,7 @@ proc maintainSubscription(wfc: WakuFilterClient,
       # No subscription found. Let's subscribe.
       notice "no subscription found. Sending subscribe request"
 
-      let subscribeRes = await wfc.subscribe(filterPeer, filterPubsubTopic, @[filterContentTopic])
+      let subscribeRes = await wfc.subscribe(filterPeer, some(filterPubsubTopic), @[filterContentTopic])
```
alrevuelta, line 51: same, just content topic?

waku/v2/waku_filter_v2/client.nim
```diff
@@ -77,25 +79,79 @@ proc ping*(wfc: WakuFilterClient, servicePeer: RemotePeerInfo): Future[FilterSub
...
-proc unsubscribe*(wfc: WakuFilterClient, servicePeer: RemotePeerInfo, pubsubTopic: PubsubTopic, contentTopics: seq[ContentTopic]): Future[FilterSubscribeResult] {.async.} =
-  let requestId = generateRequestId(wfc.rng)
-  let filterSubscribeRequest = FilterSubscribeRequest.unsubscribe(
-    requestId = requestId,
-    pubsubTopic = pubsubTopic,
-    contentTopics = contentTopics
-  )
-
-  return await wfc.sendSubscribeRequest(servicePeer, filterSubscribeRequest)
+proc subscribe*(wfc: WakuFilterClient, servicePeer: RemotePeerInfo, pubsubTopic: Option[PubsubTopic], contentTopics: seq[ContentTopic]): Future[FilterSubscribeResult] {.async.} =
```
alrevuelta, line 82: why do we need the pubsub topic if its derived from the content topic?

alrevuelta (changes requested):
left some comments, all of them related to pubsub vs content topic. let me know what you think or if im missing something :)

waku/v2/waku_filter/rpc.nim
alrevuelta: not sure if this is what you want or dont

> I think the filter service node would have to be able to receive a FilterRequest without pubsubTopic and somehow "know" about autosharding, even though it's most probably unaware of the application(s) it's servicing

imho pubsub topic should not be part of the request if content topic is already there, since it would be redundant. or am i missing something?

apps/wakunode2/external_config.nim
alrevuelta: yup, we are aligned on the first part.

> None of this, however, necessarily determines which shards/pubsub topic a node is configured by default to relay on. I might want to support all public waku network shards without even knowing a single content topic, so would continue needing a pubsubTopic config item for that.

If you want to to support all shard, perhaps what we need is as `all-shards` flag. This issue https://github.com/waku-org/nwaku/issues/1892 might also show my intentions.

But yeah, by now not against leaving the pubsub flag there.

alrevuelta:
@SionoiS 

> It's still a bit weird how you have to send multiple request in the case of the content topic not using the same shards.

If we remove the pubsub topic from the request then you dont need to send multiple requests? The full node will "infer" it.

apps/wakunode2/external_config.nim
jm-clius: For now at least, I think the light protocols should run the same "middleware" and populate the pubsub topic in e.g. the filter subscribe request before sending it. This may only be a first phase, but allows the application to gain from the abstraction of not having to use pubsub topics on the API, while remaining "safe" by not violating any existing protocols or having implicit assumption that the the full node is on the same version/generation of autosharding.

jm-clius:
> It's still a bit weird how you have to send multiple request in the case of the content topic not using the same shards.

True. If we don't change the underlying light protocols themselves, the "middleware" on the light client that populates the pubsub topic in requests might have to create multiple requests instead of one. To me this still seems safer as an initial step, rather than removing the pubsub topic from the protocol itself and having implicit assumption that the service node the client is contacting is on the same version/generation of autosharding as the client (in other words, on protocol level we remain explicit about shards/pubsub topics). Happy to have people disagree with me here, as I do think we eventually want to simplify how the protocols are used too. This just seems to me the simplest/safest increment to get there.

apps/wakunode2/external_config.nim
alrevuelta: resolving

happy to keep the pusub topic by now, but in the future i dont think it makes sense. i get the "remaining safe" due to generation, but this is something that i think we have to discuss. exposing the app/users to this pubsub topic + generation concept adds more load and moving parts that should be abstracted away.

in other words, it solves the problem from a technical perspective but paying in complexity for the app (developers building on top)

SionoiS:
> implicit assumption that the service node the client is contacting is on the same version/generation of autosharding as the client

The generation to use is in the content topic there's not assumptions. unsupported generation = bad request.

As for the version, any autosharding algo. change is breaking change. unsupported version = bad request.
This one is weak since the version is not in the protocol.

SionoiS:
The scope of this PR it for autosharding to work with the existing protocols.

SionoiS:
I can't really replace the store client proc with a autosharding aware version so I added another proc.

jm-clius (changes requested):
Thanks for moving sharding logic to the clients to keep the wire protocol the same. However, I'm still confused why the autosharding logic needs to be in the protocol clients. This effectively changes the protocol behaviour too (e.g. a filter subscribe now spinning off multiple subscribes). I would say the only thing that needs to be changed is where the application interacts with the public API, in other words: either in the JSON-RPC/REST APIs (e.g. [here](https://github.com/waku-org/nwaku/blob/b9d5d28af5afcc2589357a132c9b5beb57202828/waku/node/jsonrpc/filter/handlers.nim#L43-L50)) or, if we want to be a bit more universal, in the Nim API for the node, e.g. [here](https://github.com/waku-org/nwaku/blob/master/waku/node/waku_node.nim#L368-L397). If we follow the latter route, we'll still have a problem with store `query`, but the way to fix it IMO, would be to provide an API `query*` call where `HistoryQuery` creation is moved into the API call, while query criteria (content topics, optional pubsub, etc.) are provided separately as arguments.

SionoiS:
> Thanks for moving sharding logic to the clients to keep the wire protocol the same. However, I'm still confused why the autosharding logic needs to be in the protocol clients. This effectively changes the protocol behaviour too (e.g. a filter subscribe now spinning off multiple subscribes). I would say the only thing that needs to be changed is where the application interacts with the public API, in other words: either in the JSON-RPC/REST APIs (e.g. [here](https://github.com/waku-org/nwaku/blob/b9d5d28af5afcc2589357a132c9b5beb57202828/waku/node/jsonrpc/filter/handlers.nim#L43-L50)) or, if we want to be a bit more universal, in the Nim API for the node, e.g. [here](https://github.com/waku-org/nwaku/blob/master/waku/node/waku_node.nim#L368-L397). If we follow the latter route, we'll still have a problem with store `query`, but the way to fix it IMO, would be to provide an API `query*` call where `HistoryQuery` creation is moved into the API call, while query criteria (content topics, optional pubsub, etc.) are provided separately as arguments.

I got confused by `node` again. JSONRPC handler call `node.lightpushPublish()` then it calls `client.publish()`, I though it called it directly.

I am not a fan of adding the "sharding middleware" to the APIs handlers and I also don't like node.

I guess I'll put it in node for now...

SionoiS:
Is there a difference between omitting pubsub topic for a STORE request or one request per shards?

I see three different intent when making a request.
1. I don't care about pubsub topics.
2. I want autosharding
3. I want messages for one pubsub topic.

1 and 2 are functionally the same but maybe we want to differentiate?

Also, like a said there's no easy way to merge results when sending multiple requests. How would you page?

Maybe we could error when requesting multiple shards at the same time?

WDYT? @jm-clius @alrevuelta

jm-clius:
> I am not a fan of adding the "sharding middleware" to the APIs handlers and I also don't like node.

I understand, though I think both options are better than adding it into the protocols themselves. Another option may be just to create a new API for applications using autosharding? Up to you, especially since we can increment here.

> Maybe we could error when requesting multiple shards at the same time?

Mmm. Yes, I think Store behaviour here is different than Filter, in that it's possible to query with an open (i.e. unpopulated) pubsub topic which would mean "don't care". But in the case of autosharding you assume your content topics are unique (i.e. doesn't span shards), so there's no reason to change anything for autosharding (the pubsub topic can just remain empty in the `HistoryQuery`). Where things will get interesting for store and other protocols is in how we select the store peer that is able to service our request - we'd need to know that it's a store node for the shard we're querying on.

waku/waku_core/topics/sharding.nim
```diff
@@ -91,15 +72,66 @@ proc weightedShardList*(topic: NsContentTopic, shardCount: int, weightList: seq[
...
+  let weights = repeat(1.0, count)
 
   let list = ? weightedShardList(topic, count, weights)
 
   let (pubsub, _) = list[list.len - 1]
 
-  ok(pubsub)
+  ok(pubsub) ]#
+
+proc genZeroSharding*(topic: NsContentTopic, count: int): NsPubsubTopic =
```
jm-clius, line 88: Does this function need to be exported `*` or is it simply internal call of `autosharding*()`?

waku/waku_core/topics/sharding.nim
```diff
@@ -91,15 +72,66 @@ proc weightedShardList*(topic: NsContentTopic, shardCount: int, weightList: seq[
...
+
+  # We only use the last 64 bits of the hash as having more shards is unlikely.
+  let hashValue = uint64.fromBytesBE(hash.data[24..31])
+
+  # This is equilavent to modulo shard count but faster
+  let shard = hashValue and uint64((count - 1))
+ 
+  NsPubsubTopic.staticSharding(ClusterIndex, uint16(shard))
+
+proc autosharding*(topic: NsContentTopic): Result[NsPubsubTopic, string] =
```
jm-clius, line 101: We _generally_ try to stick to verbs for function names to be clearer that it performs an action. In this case, I'd go for something like `getAutoshard()` or similar. Same applies to `genZeroSharding` above.

waku/node/waku_node.nim
```diff
@@ -389,14 +387,45 @@ proc filterSubscribe*(node: WakuNode, pubsubTopic: PubsubTopic, contentTopics: C
 
       await handler(pubsubTopic, message)
```
jm-clius, line 389: If `pubsubTopic.isSome()` we want to skip the unnecessary iteration and autoshard considerations and simply publish the filter request as it is.

waku/node/waku_node.nim
```diff
@@ -389,14 +387,45 @@ proc filterSubscribe*(node: WakuNode, pubsubTopic: PubsubTopic, contentTopics: C
...
+    let (pubsub, content) =
+      if res.isErr():
+        error "parsing error", error = res.error
+        return
+      else: res.get()
+    
+    if not topicMap.hasKey(pubsub):
+      topicMap[pubsub] = @[]
+    
+    topicMap[pubsub].add(content)
```
jm-clius, line 409: Can we extract all this logic into a simple helper function in topics/sharding.nim (could overload `parseSharding`) that takes a `pubsubTopic` option and sequence of content topics and returns a map from pubsub topic to content topics. Will prevent waku_node.nim from bloating even further and can be reused for the same logic in the `unsubscribe` below.

jm-clius (approved):
Thanks for your patience. I have added some comments below but have also approved, since I think that when addressed this can be merged.

waku/waku_core/topics/sharding.nim
SionoiS: Is there a way to not export when used in testing? That would be ideal.

waku/waku_core/topics/sharding.nim
SionoiS: hmm. `computeShard()` and `computeGen0Shard` then?

waku/node/waku_node.nim
SionoiS: that would be better yes

waku/node/waku_node.nim
SionoiS: true

waku/waku_core/topics/sharding.nim
jm-clius: yeah, works for me!

waku/waku_core/topics/sharding.nim
jm-clius: Unfortunately not. If you need to test this, it would indeed have to remain exported.

alrevuelta (approved):

PR #1859, opened 2023-07-25
fix(rln-relay): sync from a recent block number, not 0

rymnc:
!WIP!
- fix(rln-relay): sync from a recent block number, not 0

# Description

# Changes

- [ ] ...
- [ ] ...

### DIFF
--- a/waku/v2/node/waku_node.nim
+++ b/waku/v2/node/waku_node.nim
@@ -828,4 +828,8 @@ proc stop*(node: WakuNode) {.async.} =
   await node.switch.stop()
   node.peerManager.stop()
 
+  when defined(rln):
+    if not node.wakuRlnRelay.isNil():
+      await node.wakuRlnRelay.stop()
+
   node.started = false
--- a/waku/v2/waku_rln_relay/group_manager/group_manager_base.nim
+++ b/waku/v2/waku_rln_relay/group_manager/group_manager_base.nim
@@ -89,6 +89,9 @@ method withdrawBatch*(g: GroupManager, identitySecretHashes: seq[IdentitySecretH
 method atomicBatch*(g: GroupManager, idCommitments: seq[IDCommitment], toRemoveIndices: seq[MembershipIndex]): Future[void] {.base,gcsafe.} =
   raise newException(CatchableError, "atomicBatch proc for " & $g.type & " is not implemented yet")
 
+method close*(g: GroupManager): Future[void] {.base,gcsafe.} =
+  raise newException(CatchableError, "close proc for " & $g.type & " is not implemented yet")
+
 # This proc is used to set a callback that will be called when an identity commitment is withdrawn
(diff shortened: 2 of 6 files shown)


rymnc:
This PR is no longer required since it was fixed with the decay introduction in https://github.com/waku-org/nwaku/pull/1858

alrevuelta:
can see its closed but out of curiosity. which block is it used to sync the registration events? from 0?

rymnc:
> can see its closed but out of curiosity. which block is it used to sync the registration events? from 0?

Yes, as of now, its 0 - but when `waku-org/waku-rln-contract` is a submodule, we will fetch from the block number in which the contract was deployed

alrevuelta:
@rymnc Leaving the idea in case you find it useful. afaik in solidity you can access the block `block.number`. So in the constructor that is called on deployment, just store `block.number` and then add a getter to it.

So then we can just `getDeployedBlockNumber()` and sync from there.

Is that what you had in mind by "we will fetch from the block number in which the contract was deployed" ?

rymnc:
Sure, we can add block.number in the constructor, it's unnecessary storage though. (open to this solution)
by 
"we will fetch from the block number in which the contract was deployed", i meant from the deployment stored in the `deployments` directory in waku-org/waku-rln-contract

alrevuelta:
> it's unnecessary storage though

shouldn't be expensive, just one variable? isnt it neglectable?

> i meant from the deployment stored in the deployments directory in waku-org/waku-rln-contract

but how will people access it? manually? or how can this be automated?

rymnc:
> shouldn't be expensive, just one variable? isnt it neglectable?

agree, it can be neglected - I'll make the changes upstream for this 

> but how will people access it? manually? or how can this be automated?

since it is a submodule in nwaku, it will be automated - check [#1884](https://github.com/waku-org/nwaku/pull/1884), where its automatically fetched

alrevuelta:
> since it is a submodule in nwaku, it will be automated - check https://github.com/waku-org/nwaku/pull/1884, where its automatically fetched

ah i see. but then if someone deploys its own rln contract, this wont work, and we would need a flag to specify it and override the default?

rymnc:
> > since it is a submodule in nwaku, it will be automated - check #1884, where its automatically fetched
> 
> ah i see. but then if someone deploys its own rln contract, this wont work, and we would need a flag to specify it and override the default?

Correct - I will use your approach moving forward, but there still may be some merit in using waku-rln-contract as a submodule, wdyt?

alrevuelta:
> but there still may be some merit in using waku-rln-contract as a submodule, wdyt?

having the ABI is needed i guess? so that would be a merit.

rymnc:
`nim-web3` doesn't need the ab I believe ~ it wouldn't be too useful for nwaku except for using the deployedBytecode from the artifacts (but thats only for tests)

PR #1863, opened 2023-07-27
fix(ci): do not depend on number of procesors with job name

vpavlin:
# Description
For some reason actions would omit the `env` when templating the matrix for the `include` section (i.e. Mac OS + arm64 combination), which would result in the workflow failing.

Since the `NPROC` is same for all the combinations anyway, this PR moves it out to `env` block and then uses `arch` in the job name (which also makes more sense)

# Changes

- [x] move NPROC to `env` block
- [x] change the job name to use `arch`

### DIFF
--- a/.github/workflows/release-assets.yml
+++ b/.github/workflows/release-assets.yml
@@ -6,24 +6,23 @@ on:
     - 'v*' # "e.g. v0.4"
   
   workflow_dispatch:
-    
+
+env:
+  NPROC: 2
+
 jobs:
   build-and-upload:
     strategy:
       matrix:
-        env:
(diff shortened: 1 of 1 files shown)


Ivansete-status (approved):
LGTM! Thanks! :)

Where is the NPROC being used btw?

vpavlin:
It is used to instruct `make` and nim compiler to use more than 1 core

https://github.com/waku-org/nwaku/blob/8ee97b20b51bcc7fa99125604b76dea5caed3c20/.github/workflows/release-assets.yml#L55

NagyZoltanPeter (approved):
Looks very reasonable!

Issue #1851, opened 2023-07-13
feat: HTTP REST API: Filter 


jm-clius:
### Background

This forms a subtask of https://github.com/waku-org/nwaku/issues/1076.
Nwaku nodes operating as [12/WAKU2-FILTER](https://rfc.vac.dev/spec/12/) clients can use the filter protocol to subscribe to service nodes to receive messages adhering to a set of subscription criteria.
Currently these filter operations can only be triggered in clients in an [internal Nim API call](https://github.com/waku-org/nwaku/blob/master/waku/v2/waku_filter_v2/client.nim#L74-L106).
Although there exists an external filter client JSON-RPC API, this was designed for a deprecated version of the filter protocol. There is therefore no external API for filter client calls.
The JSON-RPC API is clunky to use and call signatures outdated. As such, we've started migrating to [an HTTP REST API](https://github.com/waku-org/nwaku/issues/727). 

### Suggested solution

Create and expose a Filter Client HTTP REST API functionality. It should use the existing [REST APIs](https://github.com/waku-org/nwaku/blob/master/waku/v2/node/rest/) as inspiration and follow the same patterns and structure. Although, the JSON-RPC API can be referred to, the new API method should follow the new filter API specifications to construct the calls.

### Acceptance criteria

- [ ] working HTTP Filter Client REST API to perform filter client operations on remote service nodes
- [ ] sufficient unit tests for basic functionality
- [ ] integration of Filter Client API into the `wakunode2` app

PR #1864, opened 2023-07-28
feat: Rest API interface for filter service. 

NagyZoltanPeter:
# Description

Draft PR for preliminary review, I appreciate any comment on it. Specially on the test_rest_filter.nim tests as 
peers and not finding each other. Probably overlooked something in setting up.

- Interface added
- tests are added but not yet functional. 

# Changes

- [x] Filter rest api is added
- [x] Tests are added for sub/unsub basic functionality.

## How to test
nim c -r -d:chronicles_log_level=DEBUG --verbosity=0 --hints=off --outdir=build ./tests/v2/wakunode_rest/test_rest_filter.nim test "Subscribe a node to an array of topics - POST /filter/v1/subscriptions"

tracking: [#1851](https://github.com/waku-org/nwaku/issues/1851)

### DIFF
--- a/apps/wakunode2/app.nim
+++ b/apps/wakunode2/app.nim
@@ -45,6 +45,7 @@ import
   ../../waku/v2/node/rest/debug/handlers as rest_debug_api,
   ../../waku/v2/node/rest/relay/handlers as rest_relay_api,
   ../../waku/v2/node/rest/relay/topic_cache,
+  ../../waku/v2/node/rest/filter/handlers as rest_filter_api,
   ../../waku/v2/node/rest/store/handlers as rest_store_api,
   ../../waku/v2/node/jsonrpc/admin/handlers as rpc_admin_api,
   ../../waku/v2/node/jsonrpc/debug/handlers as rpc_debug_api,
@@ -565,6 +566,11 @@ proc startRestServer(app: App, address: ValidIpAddress, port: Port, conf: WakuNo
   if conf.relay:
     let relayCache = TopicCache.init(capacity=conf.restRelayCacheCapacity)
     installRelayApiHandlers(server.router, app.node, relayCache)
+
+  ## Filter REST API
+  if conf.filter:
+    let filterCache = rest_filter_api.MessageCache.init(capacity=rest_filter_api.filterMessageCacheDefaultCapacity)
+    installFilterApiHandlers(server.router, app.node, filterCache)
 
   ## Store REST API
   installStoreApiHandlers(server.router, app.node)
--- a/waku/v2/node/rest/filter/handlers.nim
+++ b/waku/v2/node/rest/filter/handlers.nim
@@ -46,21 +46,25 @@ type
 func decodeRequestBody[T](contentBody: Option[ContentBody]) : Result[T, RestApiResponse] =
   # Check the request body
   if contentBody.isNone():
-    return err(RestApiResponse.badRequest())
+    return err(RestApiResponse.badRequest("Missing content body"))
 
   let reqBodyContentType = MediaType.init($contentBody.get().contentType)
   if reqBodyContentType != MIMETYPE_JSON:
-    return err(RestApiResponse.badRequest())
+    return err(RestApiResponse.badRequest("Wrong Content-Type, expected application/json"))
 
   let reqBodyData = contentBody.get().data
 
-  let reqResult = decodeFromJsonBytes(T, reqBodyData)
-  if reqResult.isErr():
-    return err(RestApiResponse.badRequest())
+  let requestResult = decodeFromJsonBytes(T, reqBodyData)
+  if requestResult.isErr():
+    return err(RestApiResponse.badRequest("Invalid content body, could not decode. " & $requestResult.error))
 
-  return ok(reqResult.get())
+  return ok(requestResult.get())
+
 
 proc installFilterPostSubscriptionsV1Handler*(router: var RestRouter, node: WakuNode, cache: MessageCache) =
+
+  let pushHandler: FilterPushHandler = proc(pubsubTopic: PubsubTopic, msg: WakuMessage) {.async, gcsafe, closure.} =
+      cache.addMessage(msg.contentTopic, msg)
 
   router.api(MethodPost, ROUTE_FILTER_SUBSCRIPTIONSV1) do (contentBody: Option[ContentBody]) -> RestApiResponse:
     # ## Subscribes a node to a list of contentTopics of a pubsubTopic
@@ -69,22 +73,20 @@ proc installFilterPostSubscriptionsV1Handler*(router: var RestRouter, node: Waku
     let decodedBody = decodeRequestBody[FilterSubscriptionsRequest](contentBody)
 
     if decodedBody.isErr():
-      error "Failed to decode body", error=decodedBody.error()
-      return decodedBody.error()
+      return decodedBody.error
 
     let req: FilterSubscriptionsRequest = decodedBody.value()
 
     let peerOpt = node.peerManager.selectPeer(WakuFilterCodec)
 
     if peerOpt.isNone():
-      raise newException(ValueError, "no suitable remote filter peers")
+      return RestApiResponse.internalServerError("No suitable remote filter peers")
 
-    let handler: FilterPushHandler = proc(pubsubTopic: PubsubTopic, msg: WakuMessage) {.async, gcsafe, closure.} =
-        cache.addMessage(msg.contentTopic, msg)
 
-    let subFut = node.filterSubscribe(req.pubsubTopic, req.contentFilters, handler, peerOpt.get())
+    let subFut = node.filterSubscribe(req.pubsubTopic, req.contentFilters, pushHandler, peerOpt.get())
     if not await subFut.withTimeout(futTimeout):
-      raise newException(ValueError, "Failed to subscribe to contentFilters")
+      error "Failed to subscribe to contentFilters do to timeout!"
+      return RestApiResponse.internalServerError("Failed to subscribe to contentFilters")
 
     # Successfully subscribed to all content filters
     for cTopic in req.contentFilters:
@@ -101,13 +103,14 @@ proc installFilterDeleteSubscriptionsV1Handler*(router: var RestRouter, node: Wa
     let decodedBody = decodeRequestBody[FilterSubscriptionsRequest](contentBody)
 
     if decodedBody.isErr():
-      return decodedBody.error()
+      return decodedBody.error
 
     let req: FilterSubscriptionsRequest = decodedBody.value()
 
     let unsubFut = node.unsubscribe(req.pubsubTopic, req.contentFilters)
     if not await unsubFut.withTimeout(futTimeout):
-      raise newException(ValueError, "Failed to unsubscribe from contentFilters")
+      error "Failed to unsubscribe from contentFilters due to timeout!"
+      return RestApiResponse.internalServerError("Failed to unsubscribe from contentFilters")
 
     for cTopic in req.contentFilters:
       cache.unsubscribe(cTopic)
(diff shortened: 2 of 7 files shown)


jm-clius (commented):
Thanks for this. Functionally, this looks good. Two comments:

1. I apologise that this wasn't clear enough in https://github.com/waku-org/nwaku/issues/1851, but we want this effort to also consider the "new" filter client API. This still follows the same "legacy" filter API functions as the existing JSON-RPC API for filter. We've since significantly improved the filter specification and the client API now defines different functions as documented in https://github.com/waku-org/nwaku/blob/master/waku/v2/waku_filter_v2/client.nim#L74-L106 and https://rfc.vac.dev/spec/12/ (you can still view the old version if you click on the link at the top of the RFC). These client API functions are not yet integrated in the `node` object, so we'll have to first expand the node to set up such a new filter client (in a next PR) and then expose API functions for these as well. Will provide more details in a 1:1. For now I propose, rough roadmap:
- get this version merged and tests fixed/integrated
- integrate "new" filter client in node and expose client functions in node API
- expose new filter client API functions in REST API

2. Could you provide more details as to what you see in terms of the tests failing? Perhaps some logs? Are all tests failing? We don't expect a filter client to connect to its service peer on startup - it makes ad-hoc connections when needed (i.e. only when it posts the subscription request). Perhaps you just need to add a `sleep` to allow the client to connect to the service node after `filterPost` or other action that requires the connection to be setup first?

NagyZoltanPeter:
Update:

Tests refactored a bit and now works ok. 
Basic functionality tested: subscribe/unsubscribe (original way, copying jsonrpc)

Branch squashed and rebased.

NagyZoltanPeter:
Update:

Tests refactored a bit and now works ok. 
Basic functionality tested: subscribe/unsubscribe (original way, copying jsonrpc)

Branch squashed and rebased.

NagyZoltanPeter:
Hi Hanno,
Yes it was not clear for me. Looking at the code and the existing openapi spec it seemed just an interface change.
I agree with the proposed steps and yet we need to discuss through due some unclear points remained even after more careful read of rfc. I have questions to the interface as well.
The latter, test issues are solved with help of Ivan already.


> Thanks for this. Functionally, this looks good. Two comments:
> 
> 1. I apologise that this wasn't clear enough in [feat: HTTP REST API: Filter  #1851](https://github.com/waku-org/nwaku/issues/1851), but we want this effort to also consider the "new" filter client API. This still follows the same "legacy" filter API functions as the existing JSON-RPC API for filter. We've since significantly improved the filter specification and the client API now defines different functions as documented in https://github.com/waku-org/nwaku/blob/master/waku/v2/waku_filter_v2/client.nim#L74-L106 and https://rfc.vac.dev/spec/12/ (you can still view the old version if you click on the link at the top of the RFC). These client API functions are not yet integrated in the `node` object, so we'll have to first expand the node to set up such a new filter client (in a next PR) and then expose API functions for these as well. Will provide more details in a 1:1. For now I propose, rough roadmap:
> 
> * get this version merged and tests fixed/integrated
> * integrate "new" filter client in node and expose client functions in node API
> * expose new filter client API functions in REST API
> 
> 2. Could you provide more details as to what you see in terms of the tests failing? Perhaps some logs? Are all tests failing? We don't expect a filter client to connect to its service peer on startup - it makes ad-hoc connections when needed (i.e. only when it posts the subscription request). Perhaps you just need to add a `sleep` to allow the client to connect to the service node after `filterPost` or other action that requires the connection to be setup first?

SionoiS (approved):
LGTM!

apps/wakunode2/app.nim
```diff
@@ -548,6 +549,11 @@ proc startRestServer(app: App, address: ValidIpAddress, port: Port, conf: WakuNo
     let relayCache = TopicCache.init(capacity=conf.restRelayCacheCapacity)
     installRelayApiHandlers(server.router, app.node, relayCache)
 
+  ## Filter REST API
+  if conf.filter:
+    let filterCache = rest_filter_api.MessageCache.init(capacity=30)
```
vpavlin, line 554: Should this be configurable like for `restRelayCacheCapacity` above, or at least defined in a constant? (I saw there are other "magic numbers" used below this when starting RPC server, but still:)

NagyZoltanPeter: You're absolutely right about it. I will quick fix it, also I need to sign the commit that I had not set up earlier though

NagyZoltanPeter:
Fixed magic number, first not with conf but with global const. Let's see in v2 if we need a config for that value.

I had to come over some hurdles with signing the commit and I had to renew my ssh creds too...

vpavlin (approved):
LGTM

waku/v2/node/rest/filter/handlers.nim
```diff
@@ -0,0 +1,151 @@
...
+
+    let req: FilterSubscriptionsRequest = decodedBody.value()
+
+    let peerOpt = node.peerManager.selectPeer(WakuFilterCodec)
+
+    if peerOpt.isNone():
+      raise newException(ValueError, "no suitable remote filter peers")
+
+    let handler: FilterPushHandler = proc(pubsubTopic: PubsubTopic, msg: WakuMessage) {.async, gcsafe, closure.} =
+        cache.addMessage(msg.contentTopic, msg)
```
jm-clius, line 83: Minor note, but any reason why we'd create a new `FilterPushHandler` for every new `post`? Wouldn't it be possible to create it outside the main method body and reuse the same one?

jm-clius (approved):
This LGTM! Well done on a thorough first PR. Minor comment/question below, but afaict it can be merged as-is.

waku/v2/node/rest/filter/handlers.nim
```diff
@@ -0,0 +1,151 @@
...
+
+    let msgRes = cache.getMessages(contentTopic, clear=true)
+    if msgRes.isErr():
+      raise newException(ValueError, "Not subscribed to topic: " & contentTopic)
+
+    let data = FilterGetMessagesResponse(msgRes.get().map(toFilterWakuMessage))
+    let resp = RestApiResponse.jsonResponse(data, status=Http200)
+    if resp.isErr():
+      debug "An error ocurred while building the json respose", error=resp.error
+      return RestApiResponse.internalServerError()
```
Ivansete-status, line 143: Shall we bring better feedback to the client?

```suggestion
     msg = "...."
     error msg
     return RestApiResponse.internalServerError(msg)
```

waku/v2/node/rest/filter/handlers.nim
```diff
@@ -0,0 +1,151 @@
...
+
+  let reqBodyContentType = MediaType.init($contentBody.get().contentType)
+  if reqBodyContentType != MIMETYPE_JSON:
+    return err(RestApiResponse.badRequest())
+
+  let reqBodyData = contentBody.get().data
+
+  let reqResult = decodeFromJsonBytes(T, reqBodyData)
+  if reqResult.isErr():
+    return err(RestApiResponse.badRequest())
```
Ivansete-status, line 59: Little nitpick to help debug possible error:

```suggestion
    return err(RestApiResponse.badRequest(reqResult.error))
```

waku/v2/node/rest/filter/types.nim
```diff
@@ -0,0 +1,157 @@
...
+      contentTopic = some(reader.readValue(ContentTopic))
+    of "version":
+      version = some(reader.readValue(Natural))
+    of "timestamp":
+      timestamp = some(reader.readValue(int64))
+    else:
+      unrecognizedFieldWarning()
+
+  if payload.isNone():
+    reader.raiseUnexpectedValue("Field `payload` is missing")
```
Ivansete-status, line 112: In other comments, I mentioned that we should avoid raising exceptions. Is fine for now in the case of the `types.nim` as we are doing the same in another protocols :)

waku/v2/node/rest/filter/handlers.nim
```diff
@@ -0,0 +1,151 @@
...
+const filterMessageCacheDefaultCapacity* = 30
+
+type
+  MessageCache* = message_cache.MessageCache[ContentTopic]
+
+
+func decodeRequestBody[T](contentBody: Option[ContentBody]) : Result[T, RestApiResponse] =
+  # Check the request body
+  if contentBody.isNone():
+    return err(RestApiResponse.badRequest())
```
Ivansete-status, line 49: Shall we have more feedback in case of error?

```suggestion
    return err(RestApiResponse.badRequest("content body is none"))
```

waku/v2/node/rest/filter/handlers.nim
```diff
@@ -0,0 +1,151 @@
...
+
+  router.api(MethodPost, ROUTE_FILTER_SUBSCRIPTIONSV1) do (contentBody: Option[ContentBody]) -> RestApiResponse:
+    # ## Subscribes a node to a list of contentTopics of a pubsubTopic
+    # debug "post_waku_v2_filter_v1_subscriptions"
+
+    let decodedBody = decodeRequestBody[FilterSubscriptionsRequest](contentBody)
+
+    if decodedBody.isErr():
+      error "Failed to decode body", error=decodedBody.error()
+      return decodedBody.error()
```
Ivansete-status, line 73: Shall we have better feedback in case of error?

```suggestion
      let msg = "Failed to decode body: " & decodedBody.error()
      error msg
      return RestApiResponse.badRequest(msg)
```

waku/v2/node/rest/filter/handlers.nim
```diff
@@ -0,0 +1,151 @@
...
+
+    let req: FilterSubscriptionsRequest = decodedBody.value()
+
+    let peerOpt = node.peerManager.selectPeer(WakuFilterCodec)
+
+    if peerOpt.isNone():
+      raise newException(ValueError, "no suitable remote filter peers")
```
Ivansete-status, line 80: I wouldn't raise Exception in this case. Instead, log the error and 

`return RestApiResponse.badRequest( msg )`

waku/v2/node/rest/filter/handlers.nim
```diff
@@ -0,0 +1,151 @@
...
+    let decodedBody = decodeRequestBody[FilterSubscriptionsRequest](contentBody)
+
+    if decodedBody.isErr():
+      return decodedBody.error()
+
+    let req: FilterSubscriptionsRequest = decodedBody.value()
+
+    let unsubFut = node.unsubscribe(req.pubsubTopic, req.contentFilters)
+    if not await unsubFut.withTimeout(futTimeout):
+      raise newException(ValueError, "Failed to unsubscribe from contentFilters")
```
Ivansete-status, line 110:
```suggestion
      return RestApiResponse.internalServerError("Failed to unsubscribe from contentFilters")
```

waku/v2/node/rest/filter/handlers.nim
```diff
@@ -0,0 +1,151 @@
...
+
+proc installFilterGetMessagesV1Handler*(router: var RestRouter, node: WakuNode, cache: MessageCache) =
+  router.api(MethodGet, ROUTE_RELAY_MESSAGESV1) do (contentTopic: string) -> RestApiResponse:
+    # ## Returns all WakuMessages received on a specified content topic since the
+    # ## last time this method was called
+    # ## TODO: ability to specify a return message limit
+    # debug "get_waku_v2_filter_v1_messages", contentTopic=contentTopic
+
+    if contentTopic.isErr():
+      return RestApiResponse.badRequest()
```
Ivansete-status, line 129: In this proc, all the possible errors should be retrieved back to the client

```suggestion
      return RestApiResponse.badRequest("Error in content topic: " & contentTopic.error)
```

Ivansete-status (commented):
Thanks for that! you made it super fast and very well!
I add small details to mostly enhance the feedback in case of error.
Great job!

NagyZoltanPeter:
@Ivansete-status, @jm-clius 
Thanks for the valuable comments, I fixed those. I did not want to left for v2.
Also added some more tests. In v2 I will add even more, planned.

Ivansete-status (approved):
LGTM! Thanks for it and for your patience!

PR #1867, opened 2023-07-31
fix(rln-relay): mark duplicated messages as spam

### LINKED ISSUE #1862: bug(rln-relay): duplicated messages within same epoch
In https://github.com/waku-org/nwaku/blob/master/waku/v2/waku_rln_relay/rln_relay.nim#L105-L107 we are returning `ok(false)` if there's already an identical record. We should return instead a `true`, because this means that a message is duplicated within the same epoch. 

Currently `hasDuplicate` considers a message to be duplicated if there exists 2 messages in the log with the same `nullifier` but different `share_x` and `share_y` values. The problem is that the RLN Signal is built using only the content topic and the payload, meaning that there is the possibility of sending more than one message containing the same information within the same epoch and have these messages be considered not duplicate. (This is not necessarily malicious behaviour as it could be a common scenario in the case of machine to machine messages).

Changing the return of https://github.com/waku-org/nwaku/blob/master/waku/v2/waku_rln_relay/rln_relay.nim#L105-L107 to `ok(true)` will capture these scenarios and mark the message as invalid.

To reproduce this scenario the following test unit can be built:
1. Create a loop that every second publishes via relay a `WakuMessage` with the same payload, version, content topic, and proof, and replace the timestamp with the current time in nanoseconds.
2. Create a subscriber on the pubsub topic where the messages are being broadcasted
Current behavior: a message is received every second
Expected Result: only 1 message should be received per epoch (10s), with 9 messages being marked as spam.

I believe that another possibility could be for building the RLN Signal to take into account more fields from the `WakuMessage` like the timestamp, meta, version etc, but the possibility of duplication will be reintroduced if new attributes are added to the message protobuffer.

rymnc:
# Description

If we have the same externalNullifier and proofMetadata, and same `share_x` and `share_y` we should mark the message as spam.

# Changes

- [x] Marks duplicated messages with the same proof as spam

## Issue

closes #1862

### DIFF
--- a/waku/v2/waku_rln_relay/rln_relay.nim
+++ b/waku/v2/waku_rln_relay/rln_relay.nim
@@ -110,8 +110,8 @@ proc hasDuplicate*(rlnPeer: WakuRLNRelay,
     return ok(false)
   try:
     if rlnPeer.nullifierLog[externalNullifier].contains(proofMetadata):
-      # there is an identical record, ignore rhe mag
-      return ok(false)
+      # there is an identical record, mark it as spam
+      return ok(true)
 
     # check for a message with the same nullifier but different secret shares
     let matched = rlnPeer.nullifierLog[externalNullifier].filterIt((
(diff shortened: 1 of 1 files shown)


vpavlin:
Is the description of the `hasDuplicate` function still correct with this change?

vpavlin (approved):
The change makes sense, just consider updating the description of the function

rymnc:
> The change makes sense, just consider updating the description of the function

Addressed in [54c1d9c](https://github.com/waku-org/nwaku/pull/1867/commits/54c1d9c8965273b240f1324575b22411acedd8d9), thanks!

PR #1875, opened 2023-08-02
feat: ENR can now be updated

### LINKED ISSUE #1843: feat: update ENR based on RPC topic subscribe
### Problem
Subscribing to a new topics via RPC does not update the ENR when using shards.

### Suggested solution
discv5 has a way to update the record. Create the same in waku discv5.

### Alternatives considered
Restarting the node with new topics always works but is annoying.

### Additional context
https://github.com/status-im/nim-eth/blob/285da12bf318a2e21182dc2453a4a30c58f73067/eth/p2p/discoveryv5/protocol.nim#L247

### Acceptance criteria
- [ ] Subbing to a shard topic like ```/waku/2/rs/16/0``` via RPC should update the ENR in waku discv5.

SionoiS:
# Description

I added a way to update the Discv5 ENR and used a `AsyncEventQueue` to broadcast topic subscriptions.

# Changes
- [x] Discv5 add/remove shard from ENR
- [x] event system for topic subscriptions  

Fix #1843

Followed by #1918

### DIFF
--- a/apps/wakunode2/app.nim
+++ b/apps/wakunode2/app.nim
@@ -541,12 +541,14 @@ proc startNode(node: WakuNode, conf: WakuNodeConf,
 
 proc startApp*(app: App): Future[AppResult[void]] {.async.} =
   if app.wakuDiscv5.isSome():
-    let res = app.wakuDiscv5.get().start()
+    let wakuDiscv5 = app.wakuDiscv5.get()
 
+    let res = wakuDiscv5.start()
     if res.isErr():
       return err("failed to start waku discovery v5: " & $res.error)
 
-    asyncSpawn app.wakuDiscv5.get().searchLoop(app.node.peerManager, some(app.record))
+    asyncSpawn wakuDiscv5.searchLoop(app.node.peerManager, some(app.record))
+    asyncSpawn wakuDiscv5.subscriptionsListener(app.node.topicSubscriptionQueue)
 
   return await startNode(
     app.node,
--- a/waku/waku_discv5.nim
+++ b/waku/waku_discv5.nim
@@ -4,7 +4,7 @@ else:
   {.push raises: [].}
 
 import
-  std/[sequtils, strutils, options],
+  std/[sequtils, strutils, options, sugar, sets],
   stew/results,
   stew/shims/net,
   chronos,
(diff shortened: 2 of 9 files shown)


apps/wakunode2/app.nim
```diff
@@ -570,7 +570,7 @@ proc startRestServer(app: App, address: ValidIpAddress, port: Port, conf: WakuNo
   ## Relay REST API
   if conf.relay:
     let relayCache = TopicCache.init(capacity=conf.restRelayCacheCapacity)
-    installRelayApiHandlers(server.router, app.node, relayCache)
+    installRelayApiHandlers(server.router, app.node, app.wakuDiscv5, relayCache)
```
alrevuelta, line 573: It seems odd to pass here `wakuDiscv5`. Perhaps `wakuDiscv5` should be part of the `node`.

Not a blocker though.

SionoiS: It is weird but it's the simplest thing I could do.

A refactor of how handlers send info. internally would be good.

waku/waku_enr/sharding.nim
```diff
@@ -68,7 +68,7 @@ func init*(T: type RelayShards, cluster: uint16, indices: seq[uint16]): T =
 
   RelayShards(cluster: cluster, indices: indicesSeq)
 
-func topicsToRelayShards*(topics: seq[string]): Result[Option[RelayShards], string] =
+proc topicsToRelayShards*(topics: seq[string]): Result[Option[RelayShards], string] =
```
jm-clius, line 71: Not major, but any specific reason for this change?

waku/waku_discv5.nim
```diff
@@ -122,6 +122,75 @@ proc new*(T: type WakuDiscoveryV5,
 
   WakuDiscoveryV5.new(rng, conf, some(record))
 
+proc enrSharding(wd: WakuDiscoveryV5, newTopics: seq[PubsubTopic], add: bool):  Result[void, string] =
```
jm-clius, line 125: Nitpick: perhaps a more verb-y proc name (`modifyEnrShardField` but something better 😛 )

jm-clius (approved):
Approving to not be a bottleneck, but I'm not sure I agree with passing in discv5 into the API. The API should have no knowledge about discovery in general. Wouldn't it be possible (and simpler) to just have a "watcher" task on `wakunode2/app.nim` that monitors the subscribed topics on the node and triggers the discv5 ENR update when it notices a new/removed topic?

waku/waku_enr/sharding.nim
SionoiS: More in line with everything else. Also I'm still not sure I understand the diff? Is it no-side effect the only diff.

waku/waku_discv5.nim
SionoiS: Naming is hard...

SionoiS:
> Approving to not be a bottleneck, but I'm not sure I agree with passing in discv5 into the API. The API should have no knowledge about discovery in general. Wouldn't it be possible (and simpler) to just have a "watcher" task on `wakunode2/app.nim` that monitors the subscribed topics on the node and triggers the discv5 ENR update when it notices a new/removed topic?

Yes it is a weird thing to do. I don't like your watcher idea either. Why watch for changes when the info could be sent directly.

I'll try to find something better. I wish I was more knowledgeable about NIm :(

SionoiS:
I discussed with @richard-ramos go-waku way of sending the info. from the relay api to various waku sub-systems and it seams like more or less like I would have done it in Rust.

I still don't know what to do in Nim thought...

SionoiS:
OK!

I'm now satisfied with this solution and we can expand upon it later too.

tests/wakunode_rest/test_rest_relay.nim
```diff
@@ -95,7 +96,7 @@ suite "Waku v2 Rest API - Relay":
     topicCache.subscribe("pubsub-topic-3")
     topicCache.subscribe("pubsub-topic-x")
 
-    installRelayDeleteSubscriptionsV1Handler(restServer.router, node, topicCache)
+    installRelayDeleteSubscriptionsV1Handler(restServer.router, node,topicCache)
```
Ivansete-status, line 99: I think this file doesn't need to be changed

Ivansete-status (commented):
Is this PR duplicated? I feel like a dejavu :)
( https://github.com/waku-org/nwaku/pull/1918 )

SionoiS:
> Is this PR duplicated? I feel like a dejavu :) ( #1918 )

Yeah sorry should not have ask to review the follow-up PR before merging this.

Issue #1877, opened 2023-08-03
chore: build a container image from existing binaries


vpavlin:
### Background

When we build our container (Docker) images we rebuild wakunode from scratch. This is generally ok as the whole build process is mostly self-contained, but there is still a tiny chance of the resulting binary in the container to differ from the released binaries and it wastes resources/time. 

It would make more sense to build the container images from existing binaries - especially for releases (vX.Y.Z), but also for commits in `master` and we can even use this to build images per PR

### Details

* Come up with an alternative Dockerfile (`Dockerfile.binaries` ?) which would pick up a `wakunode2` binary from predefined location rather than rebuild it
* Use this new Dockefile in Github Actions workflows - CI (for PRs and master merges), pre-release, release...
* Setup pushes to some registry (for PRs, it would be ideal to separate from the main `statusteam/nim-waku` docker repo, and ideall use quay.io as that allows to automatically clean up the images after X days)

### Acceptance criteria

- [ ] New Dockerfile working with pre-built binaries exists
- [ ] The build of this image is integrated into ci.yml (ideally comment is posted on the issue with a link to the newly built image)
- [ ] The build of this image is integrated into pre-release.yml
- [ ] The build of this image is integrated into release-assets.yml
- [ ] Images are generally available in some docker registry/repository and documentation is updated to where to find these images

Ivansete-status:
Thanks for that one @vpavlin !
I think the following is already implemented as we enjoy the PR quay.io images ;)

> and ideall use quay.io as that allows to automatically clean up the images after X days

-----------------

Besides, is interesting to start using some registry for permanent binaries. I was thinking of [Artifactory](https://jfrog.com/artifactory/) . Or do you know another fancy option ;P?

PR #1881, opened 2023-08-03
feat(ci): add docker image builds per PR

vpavlin:
# Description

This PR adds a separate Dockerfile which creates an image from a binary built in CI.

It also adds a job which pushes it to a Docker repository (quay.io/wakuorg/nwaku-pr) separate from the one where standard releases are pushed (to keep the latter clean of these "ephemeral" images) - the PR images are cleaned up after 7 days.

Additional step comments the PR to present the newly created image to contributor and reviewers.

You can see the example here: https://github.com/nwaku-test-org/nwaku/pull/12

There is one caveat - to get the image served as quickly as possible, there is an extra `wakunode2` build running, this is not ideal, but it would increase "interactivity" of the PR. It is also easy to just "append" these 2 steps to existing build job if that is preferred - interested in thoughts.

Another option would be to split the build job and first build `wakunode2` and the image and then the rest of the binaries (other apps and tools), which are not used any further anyway.

# Changes 

- [ ] additional Dockerfile under `docker/binaries/` (following nimbus folder structure here)
- [ ] additional job to build `wakunode2`, build the container image, push it to registry and provide comment

## Issue

related to #1877

### DIFF
--- a/docker/binaries/Dockerfile.bn.amd64
+++ b/docker/binaries/Dockerfile.bn.amd64
@@ -1,0 +1,34 @@
+# Dockerfile to build a distributable container image from pre-existing binaries
+FROM debian:stable-slim as prod
+
+ARG MAKE_TARGET=wakunode2
+
+LABEL maintainer="vaclav@status.im"
+LABEL source="https://github.com/waku-org/nwaku"
+LABEL description="Wakunode: Waku client"
+LABEL commit="unknown"
+
+# DevP2P, LibP2P, and JSON RPC ports
+EXPOSE 30303 60000 8545
+
+# Referenced in the binary
+RUN apt-get update &&\
+    apt-get install -y libpcre3 libpq-dev &&\
+    apt-get clean && rm -rf /var/lib/apt/lists/*
+
+# Fix for 'Error loading shared library libpcre.so.3: No such file or directory'
+RUN ln -s /usr/lib/libpcre.so /usr/lib/libpcre.so.3
+
+# Copy to separate location to accomodate different MAKE_TARGET values
+ADD ./build/$MAKE_TARGET /usr/local/bin/
+
+# Copy migration scripts for DB upgrades
+ADD ./migrations/ /app/migrations/
+
+# Symlink the correct wakunode binary
+RUN ln -sv /usr/local/bin/$MAKE_TARGET /usr/bin/wakunode
+
+ENTRYPOINT ["/usr/bin/wakunode"]
+
+# By default just show help if called without arguments
+CMD ["--help"]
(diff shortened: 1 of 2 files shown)


jm-clius (approved):
LGTM and this adds a valuable extra layer of CI. Wondering though if we shouldn't start adding short description of each CI job and their purpose in our README?

> to get the image served as quickly as possible, there is an extra wakunode2 build running

Mmm. I don't see this as a major issue, at least for now. Assuming we'll often see both failing, which is a bit of wasted effort, but at least we test the docker image building process from start to finish.

vpavlin:
Ha! It actually works!

docker/binaries/Dockerfile.bn.amd64
```diff
@@ -0,0 +1,33 @@
+FROM debian:stable-slim as prod
```
Ivansete-status, line 1: Shall we add a comment at the top indicating the expected purpose of this Dockerfile?

docker/binaries/Dockerfile.bn.amd64
```diff
@@ -0,0 +1,33 @@
...
+
+ARG MAKE_TARGET=wakunode2
+
+LABEL maintainer="vaclav@status.im"
+LABEL source="https://github.com/waku-org/nwaku"
+LABEL description="Wakunode: Waku and Whisper client"
```
Ivansete-status, line 7: I never realized that but maybe we should avoid mentioning "Whisper client" :)

Ivansete-status (commented):
Thanks for that! This is awesome!

Is it possible to use the existing Dockerfile? I have the impression that we are adding extra complexity :)

vpavlin:
> Is it possible to use the existing Dockerfile? I have the impression that we are adding extra complexity :)

We could, but I would like to gradually move away from rebuilding the binary in the docker build and rather just import some existing binary. We will still use Docker to build the binaries, but the goal is "build once, distribute in many ways". This is a first step where we introduce (and test) a Dockerfile which can take an existing binary and only use the Docker build as a way to produce a distributable package.

Hope it makes sense:)

docker/binaries/Dockerfile.bn.amd64
vpavlin: I can add a comment there - something like "A Dockerfile to build a distributable container image from pre-existing binaries" ??

Ivansete-status:
> > Is it possible to use the existing Dockerfile? I have the impression that we are adding extra complexity :)
> 
> We could, but I would like to gradually move away from rebuilding the binary in the docker build and rather just import some existing binary. We will still use Docker to build the binaries, but the goal is "build once, distribute in many ways". This is a first step where we introduce (and test) a Dockerfile which can take an existing binary and only use the Docker build as a way to produce a distributable package.
> 
> Hope it makes sense:)

Lovely, thanks for the answer!

Ivansete-status (approved):

docker/binaries/Dockerfile.bn.amd64
Ivansete-status: Yes that sounds great!

PR #1890, opened 2023-08-07
feat: HTTP REST API: Filter support v2

NagyZoltanPeter:
# Description
WIP PR to allow pre-check.

- Add support for Waku Filter v2 specification on Rest API interface.
- Integrate use of v2 filter into node.

# Changes

- [X] Updated openapi.yaml doc for v1 and v2 support
- [x] Separation of legacy filter rest api
- [x] Add client interface for v2 filter rest api
- [x] Add handler for processing v2 rest requests
- [x] Integrate v2 filter into node
- [x] Add tests for v2
- [x] Extend tests with failure cases
- [x] Rebased to latest master with applied autoshard feat on FilterV2 subscribe/unsubscribe 

# Tracks
https://github.com/waku-org/nwaku/issues/1872


NagyZoltanPeter:
@jm-clius : Please check if the interface is ok.

jm-clius (commented):
Added some comments :)

waku/v2/node/rest/filter/openapi.yaml
```diff
@@ -82,14 +104,197 @@ paths:
...
+          Returns HTTP200 if exists and HTTP404 if not.
+          Client must not fill anything but requestId in the request body.
+      operationId: subscriberPing
+      tags:
+        - filter
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/FilterSubscriptionsPing'
```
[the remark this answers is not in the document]
NagyZoltanPeter: Would be good, but each of those functions have different request data validation criterias. Therefore I decided to separate them, this way I can inform the user better about the api requirements.

NagyZoltanPeter:
Hi, @jm-clius, @Ivansete-status, @SionoiS

Please review this PR, I intend to be the finalized Filter V2 - Rest API interface.
Although there are number of changes in the protocol level, but mainly about conversion of types.

@SionoiS I would like to ask your special attention on Autoshard part of legacy filter and filter v2 adaptation.  I hope I can apply it carefully while rebasing/merging. Also I would like to discuss to apply some more test cases on sub/unsub using sharding. WDYT?

waku/node/rest/filter/handlers.nim
```diff
@@ -50,106 +52,188 @@ func decodeRequestBody[T](contentBody: Option[ContentBody]) : Result[T, RestApiR
...
     ## Subscribes a node to a list of contentTopics of a PubSub topic
-    # debug "delete_waku_v2_filter_v1_subscriptions"
+    debug "delete", ROUTE_FILTER_SUBSCRIPTIONS, contentBody
 
-    let decodedBody = decodeRequestBody[FilterSubscriptionsRequest](contentBody)
+    let decodedBody = decodeRequestBody[FilterUnsubscribeRequest](contentBody)
 
     if decodedBody.isErr():
-      return decodedBody.error
+      return makeRestResponse("unknown", FilterSubscribeError.badRequest("Failed to decode request"))
```
Ivansete-status, line 197:
```suggestion
      return makeRestResponse("unknown", FilterSubscribeError.badRequest("Failed to decode request: " & decodedBody.error))
```

Ivansete-status (changes requested):
Thanks for the PR!
I'm adding a bunch of reviews so that you can start checking it. Tomorrow morning I will study this PR again, and potentially add more comments
Cheers

waku/node/rest/filter/openapi.yaml
```diff
@@ -82,22 +104,208 @@ paths:
...
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/FilterSubscriptionResponse'
+
+
+    delete: # delete_waku_v2_filter_v2_subscription
+      summary: Unsubscribe a peer from a pubsobTopic filtering by content topics
+      description: |
+          Unsubscribe a peer from a pubsobTopic filtering by content topics
```
SionoiS, line 262: Typo `pubsubTopic` not `pubsobTopic`

waku/node/waku_node.nim
```diff
@@ -361,22 +365,50 @@ proc filterHandleMessage*(node: WakuNode, pubsubTopic: PubsubTopic, message: Wak
...
-    await node.wakuFilterClientLegacy.start()
+    await allFutures(node.wakuFilterClientLegacy.start(), node.wakuFilterClient.start())
 
+  node.switch.mount(node.wakuFilterClient, protocolMatcher(WakuFilterSubscribeCodec))
   node.switch.mount(node.wakuFilterClientLegacy, protocolMatcher(WakuFilterCodec))
 
-proc filterSubscribe*(node: WakuNode, pubsubTopic: Option[PubsubTopic], contentTopics: ContentTopic|seq[ContentTopic],
-                handler: FilterPushHandler, peer: RemotePeerInfo|string) {.async, gcsafe, raises: [Defect, ValueError].} =
+# TODO: review this if it is sufficient app interface or not.
+proc getFilterV2PushMessageCache*(node: WakuNode): Option[TopicCache] =
```
SionoiS, line 394: Fine for now but I'm thinking about adding a `waku_cache` sub-system in the near future.

waku/node/waku_node.nim
```diff
@@ -431,9 +464,76 @@ proc filterSubscribe*(node: WakuNode, pubsubTopic: Option[PubsubTopic], contentT
     for pubsub, topics in topicMap.pairs:
       info "subscribed to topic", pubsubTopic=pubsub, contentTopics=topics
 
-proc filterUnsubscribe*(node: WakuNode, pubsubTopic: Option[PubsubTopic], contentTopics: ContentTopic|seq[ContentTopic],
-                  peer: RemotePeerInfo|string) {.async, gcsafe, raises: [Defect, ValueError].} =
-  ## Unsubscribe from a content filter.
+proc filterV2Subscribe*(node: WakuNode,
```
SionoiS, line 467: Should be just called `filterSubscribe` because the other is called now `legacyFilterSubscribe` no?

waku/node/waku_node.nim
```diff
@@ -482,35 +582,86 @@ proc filterUnsubscribe*(node: WakuNode, pubsubTopic: Option[PubsubTopic], conten
...
       info "unsubscribed from topic", pubsubTopic=pubsub, contentTopics=topics
 
-# TODO: Move to application module (e.g., wakunode2.nim)
-proc subscribe*(node: WakuNode, pubsubTopic: Option[PubsubTopic], contentTopics: ContentTopic|seq[ContentTopic], handler: FilterPushHandler) {.async, gcsafe,
-  deprecated: "Use the explicit destination peer procedure. Use 'node.filterSubscribe()' instead.".} =
-  ## Registers for messages that match a specific filter. Triggers the handler whenever a message is received.
-  if node.wakuFilterClientLegacy.isNil():
-    error "cannot register filter subscription to topic", error="waku filter client is nil"
-    return
+proc filterV2Unsubscribe*(node: WakuNode,
```
SionoiS, line 585: I would drop the V2 part.

SionoiS (approved):
LGTM

See comments but nothing major.

NagyZoltanPeter:
Sorry, it was my bad to screw up zerokit submodule commit while doing the rebase. Not it is fixed and align with master. 
Thanks for pointing out @Ivansete-status

waku/node/waku_node.nim
```diff
@@ -361,22 +365,50 @@ proc filterHandleMessage*(node: WakuNode, pubsubTopic: PubsubTopic, message: Wak
...
+  if filterV2MessagePushHandler.isNone():
+    # TODO: tie it to config, but from proper place (rest_legacy_filter_api.filterMessageCacheDefaultCapacity)
+    node.filterV2PushMessageCache = TopicCache.init(capacity=30)
+
+    let internalPushMessageHandler: filter_client.MessagePushHandler = proc(pubsubTopic: string, message: WakuMessage) {.gcsafe, closure.} =
+      node.filterV2PushMessageCache.addMessage(pubsubTopic, message)
+
+    node.wakuFilterClient = filter_client.WakuFilterClient.new(node.rng, internalPushMessageHandler, node.peerManager)
+  else:
+    node.wakuFilterClient = filter_client.WakuFilterClient.new(node.rng, filterV2MessagePushHandler.get(), node.peerManager)
```
jm-clius, line 385: Not sure I see the need here for this internal vs external message cache and handlers? For one, I think we're overcomplicating the responsibility of the `WakuNode` object, which should not really contain more logic or properties than what is necessary to mount the different protocols on the `switch` it encapsulates. In other words, it makes sense to me to have the application layer always provide its own message handler when mounting a filter client. It's up to this application layer if it needs to cache messages or not. In this case, the application layer is the REST API, which needs a handler and a cache to allow for polling functionality. I'd rather simplify this by removing the cache from the node and the internal handler unless there is some need/reasoning behind it that requires it?

jm-clius:
Btw, especially if a PR remains open for quite some time, it's helpful to "resolve" older comments that are no longer relevant or that you've addressed. This way when a reviewer comes back to the PR it's easy to navigate through the history and see what has been resolved and what is still to be addressed. :)

waku/node/waku_node.nim
```diff
@@ -431,9 +464,76 @@ proc filterSubscribe*(node: WakuNode, pubsubTopic: Option[PubsubTopic], contentT
...
+                {.async, gcsafe, raises: [Defect, ValueError].} =
+
+  ## Registers for messages that match a specific filter. Triggers the handler whenever a message is received.
+  if node.wakuFilterClient.isNil():
+    error "cannot register filter subscription to topic", error="waku filter client is not set up"
+    return err(FilterSubscribeError.serviceUnavailable())
+
+  let remotePeerRes = parsePeerInfo(peer)
+  if remotePeerRes.isErr():
+    error "Couldn't parse the peer info properly", error = remotePeerRes.error
```
Ivansete-status, line 483: Something that we should enhance in the future is to make the `waku_node` trace the minimum possible (or nothing ideally)

From my point of view, we should always deal with `Result[T]` types and leave the traces for the upper layers. There shouldn't be any trace/log in `waku_node`. wdyt about this statement @jm-clius ?

waku/node/rest/filter/handlers.nim
```diff
@@ -50,106 +52,188 @@ func decodeRequestBody[T](contentBody: Option[ContentBody]) : Result[T, RestApiR
...
+
+    let req: FilterUnsubscribeRequest = decodedBody.value()
+
+    let peerOpt = node.peerManager.selectPeer(WakuFilterCodec)
```
SionoiS, line 201: Doesn't this select a random peer?

When unsubscribing we should use the same peer than for subscribing no?

NagyZoltanPeter:
@jm-clius , @Ivansete-status , @SionoiS 
Hereby I would like to come to conclusion on approach mounting 
Filter V2 client vs. legacy one.

All of you found this part questionable. I expected about it but would like to come up with some sort of solution.
Mostly because it derives from the same question the led to SP's cache handling proposal.

Unlike realy's and legacy filte's approach where we add message callback per subscription, FilterV2 changes it to one global handler to be provided by its user at creation. So it makes it a bit more difficult to provide it at the stage of creation. Maybe less flexible. So for easy use, I gave an option to handle it internally for the app level and give a generic caching of received messages from the push by the filter service peer.

I feel both option a bit limiting the app level possibilities and not that consistent. 
Maybe a bit better abstraction could be more handy / happy to discuss /.

But for shorthand solution we may think it is enough that app level provides a handler for the messages pushed and do what it wants about it.

Noticing here the discussion just recently popped out on discord that it may needs a bit wider thinking where - what layer - to provide any guaranties over messages delivered to the light client.

I agree it is clumsy as now it is, so some fix is necessary.

Wdyt?

apps/wakunode2/app.nim
```diff
@@ -468,7 +471,12 @@ proc setupProtocols(node: WakuNode,
   if conf.filternode != "":
     let filterNode = parsePeerInfo(conf.filternode)
     if filterNode.isOk():
-      await mountFilterClient(node)
+      let filterMessageCache = TopicCache.init(capacity=rest_legacy_filter_api.filterMessageCacheDefaultCapacity)
+
+      let filterPushMessageHandler : waku_filter_client.MessagePushHandler = proc(pubsubTopic: string, message: WakuMessage) {.gcsafe, closure.} =
+        filterMessageCache.addMessage(message.contentTopic, message)
```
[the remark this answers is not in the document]
jm-clius: Well, as commented elsewhere, I don't really see the need for a cache here. The REST API has a need for a cache to allow for polling.

jm-clius:
> FilterV2 changes it to one global handler to be provided by its user at creation.

Right, I see your point. I still wouldn't add an intermediate handler and cache though, as it provides functionality which is only really needed by the API. In this case the API has some application-like requirements of how the pushed messages should be handled, but cannot effect this as it is not responsible for creating the `filterClient`. This indeed points out a flaw in the architecture. My suggestion here would be to change the filter client to have a list of global push handlers rather than just one set at creation and to expose an `addPushHandler()`-like method. This way applications can add handlers to an already-mounted filter client. The _API_ should then add a push handler that populates a cache and expose a read function that returns the cached messages to REST clients (and clears the cache). This is to provide a polling mechanism via REST, similar to what we have for JSON-RPC API. WDYT?

SionoiS:
Seams like the simplest thing to do for now would be what @jm-clius wrote.

Allow registering handlers in `filterClient`. Rest API register handlers for caches. APPs can also register handlers.

NagyZoltanPeter:
@Ivansete-status @jm-clius 
I've done with fixing the findings (do hope). 
Please re-check it as the filter push handling involved quite number of change also in interfaces. 

Note: I tried different approaches, but was not happy with them. So I went back the approach used with Filter v1, and providing callback at subscribe. So does rest interface is extended as well with get message for V2.

Note2.: I will see if ci goes ok and check if needs any further action.

Note3.: I did not rebased again for ease your review now for better compare against previous.

jm-clius (commented):
Right, I would really like to approve and thanks for your patience with this, but I don't quite see the benefit of introducing complex subscription management logic into the filter protocol. What particular benefit does this provide above the far simpler approach of adding handlers to the client object itself? Perhaps I'm missing part of the picture here, so happy to set a call to clear this up.

tests/wakunode_rest/test_rest_filter.nim
```diff
@@ -36,51 +41,63 @@ proc testWakuNode(): WakuNode =
...
+  restServerForService: RestServerRef
   messageCache: filter_api.MessageCache
   client: RestClientRef
+  clientTwdServiceNode: RestClientRef
 
 
 proc setupRestFilter(): Future[RestFilterTest] {.async.} =
-  result.node1 = testWakuNode()
-  result.node2 = testWakuNode()
+  result.serviceNode = testWakuNode()
```
SionoiS, line 54: It is suggested to not use the implicit `result`. see https://status-im.github.io/nim-style-guide/errors.result.html

NagyZoltanPeter:
@Ivansete-status @jm-clius 
Although I'm now finished with all fixes and catch ups where tests / examples / chat app were broken I have put back to draft while I will make agreed changes with Hanno.
We discussed to shoot for a simpler less restrictive interface on FilterPushHandling, while keep the possibility of poll messages through REST API if.
It means, it will be not necessary to add handler per subscription but apart from that. 
I will notice you when ready. But /health endpoint has now higher prio first.

waku/node/waku_node.nim
```diff
@@ -415,25 +443,96 @@ proc filterSubscribe*(node: WakuNode, pubsubTopic: Option[PubsubTopic], contentT
...
+        subRes = FilterSubscribeResult.err(res.error)
 
     for pubsub, topics in topicMap.pairs:
       info "subscribed to topic", pubsubTopic=pubsub, contentTopics=topics
 
-proc filterUnsubscribe*(node: WakuNode, pubsubTopic: Option[PubsubTopic], contentTopics: ContentTopic|seq[ContentTopic],
-                  peer: RemotePeerInfo|string) {.async, gcsafe, raises: [Defect, ValueError].} =
-  ## Unsubscribe from a content filter.
+    # return the last error or ok
+    return subRes
```
Ivansete-status, line 528: This is in combination with the two previous suggestions. _GH_ didn't allow me to make it in only one.
```suggestion
    return ok()
```

waku/node/waku_node.nim
```diff
@@ -482,35 +581,86 @@ proc filterUnsubscribe*(node: WakuNode, pubsubTopic: Option[PubsubTopic], conten
...
-    return
+  let remotePeerRes = parsePeerInfo(peer)
+  if remotePeerRes.isErr():
+    error "couldn't parse remotePeerInfo", error = remotePeerRes.error
+    return err(FilterSubscribeError.serviceUnavailable("No peers available"))
+
+  let remotePeer = remotePeerRes.value
+
+  if pubsubTopic.isSome():
+    if contentTopics.isSome():
```
Ivansete-status, line 606: Tiny nitpick:
```suggestion
  if pubsubTopic.isSome() and contentTopics.isSome():
```

waku/node/waku_node.nim
```diff
@@ -482,35 +581,86 @@ proc filterUnsubscribe*(node: WakuNode, pubsubTopic: Option[PubsubTopic], conten
...
+      let unsubRes = await node.wakuFilterClient.unsubscribe(remotePeer, pubsubTopic.get(), contentTopics.get())
+      if unsubRes.isOk():
+        info "unsubscribed from topic", pubsubTopic=pubsubTopic.get(), contentTopics=contentTopics.get()
+      else:
+        error "failed filter unsubscription", error=unsubRes.error
+        waku_node_errors.inc(labelValues = ["unsubscribe_filter_failure"])
+
+      return unsubRes
+
+  elif contentTopics.isNone(): # none of pubsubTopic and contentTopics are not provided, go for unsub all
```
Ivansete-status, line 618: I'd be more explicit as this depends somehow to the previous `if`.
```suggestion
  elif pubsubTopic.isNone() and contentTopics.isNone(): # go for unsub all
```

waku/waku_core/subscription.nim
```diff
@@ -0,0 +1,7 @@
+import
```
Ivansete-status, line 1: The changes in `waku/waku_filter/client.nim`, and the addition of the `subscription` module, I think can go in a separate PR because it is a refactoring. That way, we will make this PR simpler :)

Ivansete-status (commented):
It looks really great in general! Good job!
However, I've added a few nitpicks and a comment where I suggest to submit a separate PR to refactor the `waku/waku_filter/client.nim` file.

waku/node/waku_node.nim
```diff
@@ -90,8 +92,10 @@ type
     wakuStore*: WakuStore
     wakuStoreClient*: WakuStoreClient
     wakuFilter*: waku_filter_v2.WakuFilter
+    wakuFilterClient*: filter_client.WakuFilterClient
+    filterPushMessageCache: TopicCache #Exposed by proc intentionally to reflect inner or outer pushMessageHandler used
```
jm-clius, line 96: Shouldn't this now be removed?

waku/node/waku_node.nim
```diff
@@ -482,35 +581,86 @@ proc filterUnsubscribe*(node: WakuNode, pubsubTopic: Option[PubsubTopic], conten
...
-# TODO: Move to application module (e.g., wakunode2.nim)
-proc subscribe*(node: WakuNode, pubsubTopic: Option[PubsubTopic], contentTopics: ContentTopic|seq[ContentTopic], handler: FilterPushHandler) {.async, gcsafe,
-  deprecated: "Use the explicit destination peer procedure. Use 'node.filterSubscribe()' instead.".} =
-  ## Registers for messages that match a specific filter. Triggers the handler whenever a message is received.
-  if node.wakuFilterClientLegacy.isNil():
-    error "cannot register filter subscription to topic", error="waku filter client is nil"
-    return
+proc filterUnsubscribe*(node: WakuNode,
+                          pubsubTopic: Option[PubsubTopic],
+                          contentTopics: Option[seq[ContentTopic]],
```
jm-clius, line 586: I don't think this should be an `Option`. `seq[ContentTopic]` can be empty, but that would result in an `err` returned by the underlying client protocol. We don't want to build much logic into `waku_node` belonging to underlying protocols - it simply exists to mount, integrate and expose a common API for underlying protocols. The exception here is that we do provide autosharding "middleware" in the `waku_node` - in other words, logic to populate the `pubsubTopic` if absent from API calls. But even this should probably be extracted to another module. The `waku_node` should in theory just provide a `filterSubscribe`, `filterUnsubscribe` and `filterUnsubscribeAll`- not interpret certain API calls as one or the other based on added logic.

waku/node/waku_node.nim
jm-clius, line 618: This logic should not exist in the `waku_node`. It simply passes the `pubsubTopic` and `contentTopics` to the underlying client. The client has the mandate to implement the protocol here, which states that and empty `contentTopics` is an error. In fact, this would not even allow the possibility for a client to `unsubscribeAll` for all `pubsubTopics`. The node simply exposes this underlying functionality in an integrated API (in other words, should add a `filterUnsubscribeAll*` method). Even the fact that the node is applying autosharding rules here should probably be abstracted away, but that is too much for this PR.

jm-clius (changes requested):
I've requested some changes. After they are implemented I'll be ready to approve the rest of this PR. :) I think it will be too much overhead to change this now, but there could have been a great many individual PRs here. For complex PRs, even basic refactoring (such as changing line lengths) should be split into separate PRs. The API handlers and types could be a PR separate from the actual integration with the node, etc. I know this has been mentioned before, but since then the PR has still grown quite a bit.

waku/node/waku_node.nim
jm-clius: (The `filterPushMessageCache` that is)

NagyZoltanPeter:
I needed to rebase to latest master to kick in ci checks, hence I did a squash (that would be done anyway by GH).

NagyZoltanPeter:
@jm-clius, @Ivansete-status : 
Please look at the final stage of this long long running PR. 
Last changes:
- Addressed interface change on waku_node, introduced filterUnsubscribeAll.
- It involves changes on Rest interface, new endpoint added: /filter/v2/subscriptions/all with HTTP DELETE.

jm-clius (approved):
Happy with the changes now. Good job. :) Let's get this merged! Thanks for your patience.

Ivansete-status (approved):
LGTM! Thanks for it and for the patience! :100:

PR #1891, opened 2023-08-07
chore: Adding milestone issue template inspired by the one present in the `pm` repo

Ivansete-status:
# Description
Adding a milestone-issue template

### DIFF
--- a/.github/ISSUE_TEMPLATE/milestone.md
+++ b/.github/ISSUE_TEMPLATE/milestone.md
@@ -1,0 +1,41 @@
+---
+name: Milestone Issue Template
+about: Track Milestones
+title: "[Milestone] "
+labels: milestone
+assignees: ''
+
+---
+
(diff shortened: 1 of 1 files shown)


vpavlin:
Hmm, could this result in "random" community members creating Milestones (by mistake or on purpose?)...and does it matter?:D

Ivansete-status:
> Hmm, could this result in "random" community members creating Milestones (by mistake or on purpose?)...and does it matter?:D

haha I think there shouldn't be many milestones created randomly. They will be created by us at a quarter or monthly pace.
For example, we might have one "maintenance" milestone per each release, and we'd create them monthly.

Ivansete-status:
> Happy with this. Assuming there's no way for us to simply "inherit" this from another repo?

I don't know if that's feasible tbh but it could be great indeed

vpavlin (approved):

PR #1895, opened 2023-08-08
fix: fixes out of bounds crash when waku2 is not set

alrevuelta:
* Fixes an out of bounds crash when `waku2` field in the ENR is not set. This shouldn't really happen as thats not compliant with the spec and no node should have that field unset, but ofc we can't control that.
* Closes https://github.com/status-im/infra-misc/issues/164
* Note that while it was detected when using `networkmonitor` this could affect `wakunode2` if using waku2 field.
* Thanks @jakubgs for the report.


jm-clius (approved):

jakubgs:
Thanks for the quick fix.

PR #1902, opened 2023-08-10
fix: bring back default topic in config

### LINKED ISSUE #1896: bug(wakunode): bring back the --topic option for backwards compatibility
### Problem

With this change https://github.com/waku-org/nwaku/commit/afb93e299536bfdfe791a3ec88fb8608edce98bd  we lost the `--topic` option. We need to bring it back with the note on deprecation for a few releases to make sure users have time to migrate

### Impact
Potential to disrupt any existing deployments

### To reproduce

### Expected behavior

Run cmd with `--topic` still works for 2 future releases

SionoiS:
If #1857 is not merged in time for v0.19

Fix #1896

### DIFF
--- a/apps/wakunode2/app.nim
+++ b/apps/wakunode2/app.nim
@@ -360,7 +360,12 @@ proc setupProtocols(node: WakuNode,
     peerExchangeHandler = some(handlePeerExchange)
 
   if conf.relay:
-    let pubsubTopics = conf.topics
+    # TODO autoshard content topics only once.
+    # Already checked for errors in app.init
+    let contentTopics = conf.contentTopics.mapIt(NsContentTopic.parse(it).expect("Parsing"))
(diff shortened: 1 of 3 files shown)


Ivansete-status (changes requested):
Morning and thanks for this!
However, I have the impression that we are missing changes in the `app.nim` file to also support the `topic` param:

https://github.com/waku-org/nwaku/blob/b9d5d28af5afcc2589357a132c9b5beb57202828/apps/wakunode2/app.nim#L363

SionoiS:
Now it should work with `topics`, `pubsub-topics` and `content-topics` of course you can't use all of them at once.

vpavlin (approved):

jm-clius (approved):
Lgrm

PR #1904, opened 2023-08-11
fix(rln-relay): should error out on rln-relay mount failure

rymnc:
# Description

Previously, we would let the node continue as usualy if the rln-relay failed to mount ~ this PR addresses that issue by bubbling up 
the error to the `setupProtocols` proc in `wakunode2.nim`

Thanks @alrevuelta for the suggestion

# Changes

- [x] Raises an exception instead of error logging on mount failure

Maybe we should convert all the mounting procs to return a result instead of raising exceptions?

### DIFF
--- a/waku/node/waku_node.nim
+++ b/waku/node/waku_node.nim
@@ -663,8 +663,7 @@ when defined(rln):
     info "mounting rln relay"
 
     if node.wakuRelay.isNil():
-      error "WakuRelay protocol is not mounted, cannot mount WakuRlnRelay"
-      return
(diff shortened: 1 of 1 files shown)


alrevuelta:
Thanks for this! Its indeed very important to crash the node if rln was not configured properly and don't continue running the node.

> Maybe we should convert all the mounting procs to return a result instead of raising exceptions?

Don't have a strong opinion on this. Guess its matter on where you draw the line between: till here is Result and from here is exception. Not a blocker for this PR I assume?

vpavlin:
I would agree this is a good start and we should do an "analysis" of where wakunode should fail and where it is safe to continue

vpavlin (approved):

PR #1907, opened 2023-08-11
fix(chat2): no enr record in chat2

richard-ramos:
# Description
Fixes an `node record is required ` error which crashes chat2.

# Changes
- [x] build enr record and provide it to waku node


vpavlin (approved):
lgtm

PR #1908, opened 2023-08-11
fix(ci): only add comment on PR and do not duplicate it

vpavlin:
# Description

This PR makes sure the `comment` step only runs on PRs (as it fails CI for `master` otherwise)

It also configures the action to only update the comment and not post a new comment after each CI run

# Changes

- [ ] add condition to only run on PRs
- [ ] update the comment rather than add new comment on each run


NagyZoltanPeter (approved):
Lgtm!

Ivansete-status (approved):
Gorgeous! Thanks for it!

PR #1913, opened 2023-08-16
fix(rln-relay): timeout on macos runners, use fixed version of ganache

rymnc:
# Description

The macos runners were timing out consistently due to a version change in ganache. This pr addresses it by locking the 
version of ganache used, and killing the child processes that ganache creates

# Changes

- [x] kills all `ganache`'s
- [x] version lock


alrevuelta (approved):

vpavlin (approved):
oh man :facepalm:

PR #1915, opened 2023-08-17
fix(rln-relay): invalid start index being set results in invalid proofs 

rymnc:
# Description

This pr introduces a couple small changes for logging and fixes the startIndex starting from 0 instead of 1 like the contract

# Changes

- [x] logging improvements
- [x] fixed startIndex

### DIFF
--- a/waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
+++ b/waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
@@ -69,40 +69,39 @@ template initializedGuard(g: OnchainGroupManager): untyped =
   if not g.initialized:
     raise newException(ValueError, "OnchainGroupManager is not initialized")
 
-method register*(g: OnchainGroupManager, idCommitment: IDCommitment): Future[void] {.async.} =
-  initializedGuard(g)
-
-  await g.registerBatch(@[idCommitment])
-
 method atomicBatch*(g: OnchainGroupManager, 
+                    start: MembershipIndex,
                     idCommitments = newSeq[IDCommitment](), 
                     toRemoveIndices = newSeq[MembershipIndex]()): Future[void] {.async.} =
   initializedGuard(g)
 
-  let startIndex = g.latestIndex
   waku_rln_membership_insertion_duration_seconds.nanosecondTime:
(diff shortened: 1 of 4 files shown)


rymnc:
cc: @alrevuelta @richard-ramos thanks for the report!

waku/waku_rln_relay/rln/wrappers.nim
```diff
@@ -89,7 +89,7 @@ proc createRLNInstanceLocal(d = MerkleTreeDepth,
       cache_capacity: 15_000,
       mode: "high_throughput",
       compression: false,
-      flush_interval: 12_000,
+      flush_interval: 500,
```
richard-ramos, line 92: What is the flush_interval attribute used for?

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
```diff
@@ -501,6 +504,8 @@ method init*(g: OnchainGroupManager): Future[void] {.async.} =
     except CatchableError:
       error "failed to restart group sync", error = getCurrentExceptionMsg()
 
+  # Contract starts from the first index
```
richard-ramos, line 507: Why are we calculating the startIndex manually (setting g.latestIndex to be `1` here, and then using that value in `atomicBatch`, instead of passing the index retrieved from the event logs?

I'm probably wrong, so please correct me if that's the case, but I think this might be problematic in case of a partial sync or if a lot of time has passed since the last sync,  since in `getAndHandleEvents` the function `setMetadata` is called for each block chunk, and if for some reason we don't sync the full chain, the next time we start nwaku,  it will load the metadata to retrieve the latest block that was synced, and will call atomicOperation passing the index `1`, instead of the index from the logs that were emitted in the blocks that are being processed, thus the leaves from the merkle tree will be overwritten instead of inserting new elements.

rymnc: Hi, thanks for this issue - fixed in 048cdc29

waku/waku_rln_relay/rln/wrappers.nim
rymnc: This is describes how often we flush pmtree cache to disk. just experimenting with waku-simulator here

alrevuelta (approved):
Tested the latest commit. lgtm.

PR #1917, opened 2023-08-18
chore(rln): run rln in all relay pubsubtopics + remove cli flags

alrevuelta:
# Changes

* Remove `rln-relay-pubsub-topic` and `rln-relay-content-topic` cli flags.
* Remove all references to pubsub and content topics in rln.
* Mount rln in all pubsub topics that relay is configured with.
* Adds tests to ensure that rln is applied in all pubsub/content topics configured in relay.

Related https://github.com/waku-org/nwaku/issues/1906

### DIFF
--- a/waku/waku_rln_relay/rln_relay.nim
+++ b/waku/waku_rln_relay/rln_relay.nim
@@ -31,8 +31,6 @@ logScope:
 
 type WakuRlnConfig* = object
   rlnRelayDynamic*: bool
-  rlnRelayPubsubTopic*: PubsubTopic
-  rlnRelayContentTopic*: ContentTopic
   rlnRelayCredIndex*: uint
   rlnRelayMembershipGroupIndex*: uint
   rlnRelayEthContractAddress*: string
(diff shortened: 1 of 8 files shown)


jm-clius (approved):
Makes sense to me!

waku/waku_rln_relay/rln_relay.nim
```diff
@@ -308,21 +300,12 @@ proc generateRlnValidator*(wakuRlnRelay: WakuRLNRelay,
...
     if decodeRes.isOk():
-      let
-        wakumessage = decodeRes.value
-        payload = string.fromBytes(wakumessage.payload)
-
-      # check the contentTopic
-      if (wakumessage.contentTopic != "") and (contentTopic != "") and (wakumessage.contentTopic != contentTopic):
-        trace "content topic did not match:", contentTopic=wakumessage.contentTopic, payload=payload
-        return pubsub.ValidationResult.Accept
-
```
rymnc, line 323: how should we prevent replay attacks on different topics if we do not perform this check?

rymnc (commented):
LGTM in general, small query

waku/waku_rln_relay/rln_relay.nim
alrevuelta: You mean using a message (with a given content topic) + proof created by someone else to spam other pubsub topics?

Mind ellaborating also on this part? I don't fully understand why its Accepting the message when !=

rymnc: yeah, so imagine that I craft a message for a given content-topic/pubsub-topic pair, and someone takes that message and sends it on a different content-topic/pubsub-topic pair

now with sharding, this may not be an issue, but it will still be possible to send messages from one shard onto another (where it doesn't belong)

alrevuelta: With vanila relay (just for status by now), the idea was to use the timestamps to prevent this, so we only accept a window of x around the timestamp. But well, timestamp is currently not mandatory.

https://github.com/waku-org/nwaku/blob/b3bb7a1113c6d584e30eecfd724175461f0ada9c/apps/wakunode2/wakunode2_validator_signed.nim#L61

But I'm missing something in here.

rymnc: we can just use the epoch for this, its mandatory!

rymnc: I think that would solve the problem

rymnc (approved):
LGTM

Issue #1906, opened 2023-08-11
chore(rln-relay): Requirements to consider RLN ready (non experimental)


rymnc:
### Background
As discussed with @alrevuelta, following are the discussion points for rln-relay

### Details
- [x] 1. run rln-relay on all configured pubsub topics and content topics (passed in w/ `--pubsub-topic`) - 
  - [x] https://github.com/waku-org/nwaku/pull/1917
- [x] 2. separate out bandwidth constraint validator 
  - [x] https://github.com/waku-org/nwaku/pull/1966
- [x] #1967
  1. WakuMessage decoding
  2. WakuRelay
  3. WakuRlnRelay
  - [x] https://github.com/waku-org/nwaku/pull/1966
- [x] 4. trace for bandwidth info log
  - 1 The [log](https://github.com/waku-org/nwaku/blob/b9d5d28af5afcc2589357a132c9b5beb57202828)
  - 2 Remove DEBUG from [here](https://github.com/waku-org/nwaku/blob/master/waku/waku_rln_relay/rln_relay.nim#L346) and leave just TRACE (all 3 places)
  - [x] https://github.com/waku-org/nwaku/pull/1986#issuecomment-1705196322
- [x] 5. freeze nwaku startup till rln-relay is ready (otherwise it would reject valid proofs)
  - [x] https://github.com/waku-org/nwaku/pull/1920 
- [x] 6. review metrics (few too many one-time metrics, like startup time etc)
  - [x] https://github.com/waku-org/nwaku/pull/1926
- [x] 7. clear up the nullifier table (clear the table once in 10 epochs or so) (reported by @richard-ramos)
  - [x] https://github.com/waku-org/nwaku/pull/1994 
- [x] 8. health check api (`/health` should denote if the nwaku node is ready for use)
- [x] 9. Sync from the deployment block instead of from genesis. Related to discussions in https://github.com/waku-org/nwaku/pull/1859
  - [x] https://github.com/waku-org/nwaku/pull/1955
- [ ] 10. The `rln` feature should be available in the non-experimental context
- [x] 11. Add metric to show the amout of registered memberships in the tree.
- [x] 12. rename rlnkeystore application to `waku-rln-relay`, thanks @richard-ramos 
  - [x] https://github.com/waku-org/nwaku/pull/1924 
- [x] 13. remove registration capability from nwaku
  - [x] https://github.com/waku-org/nwaku/pull/1916
[...]

PR #1920, opened 2023-08-21
fix(rln-relay): waitFor startup, otherwise valid proofs will be marked invalid

rymnc:
# Description

This PR ensures that messages are processed only after the rln-relay is ready to verify proofs, i.e updated till latest 
head

# Changes

- [x] `s/await/waitFor` in chat2/wakunode2/waku_node 

## Issue

Addresses a part of #1906


alrevuelta (approved):
lgtm thanks!

Indeed, this should solve the problem that a node that just started (is out of sync with the contract, hence unaware of the tree and valid registrations) applies validation to messages with an outdated (or even empty tree). This would cause this node to flag valid messages as invalid (since its tree is not up to date) and this may result in descoring other peers + disconnecting for no reason.

jm-clius (approved):

PR #1923, opened 2023-08-21
fix(test): fix flaky rln test

alrevuelta:
fix flaky rln test


rymnc (approved):
LGTM, thanks for the quick fix

PR #1925, opened 2023-08-22
feat(rln): init rln_keystore_generator

rymnc:
# Description

Initializes a new app called `rln_keystore_generator` which will take care of the following - 
1. generate identity_credentials
2. write to local keystore
3. (optional) register on-chain

# Changes

- [x] Updated Makefile, waku.nimble and created the barebones app/


jm-clius (approved):
LGTM, except that I would suggest having this under `/tools` and not `/apps`. Apps are applications built on top of a Waku node.

alrevuelta (approved):
lgtm!

Used https://github.com/waku-org/go-waku/pull/654 and works like charm. Guess for completiness its nice to have it also in nwaku, so thanks for this.

rymnc:
> LGTM, except that I would suggest having this under /tools and not /apps. Apps are applications built on top of a Waku node.

Addressed in [8f26627](https://github.com/waku-org/nwaku/pull/1925/commits/8f26627017edaffebc020a6ebc921f1f9e964f53)

PR #1926, opened 2023-08-22
chore(rln): remove old and add new rln metric

alrevuelta:
* Removes unused RLN metrics.
* Add new RLN metric to track the amount of active RLN memberships.


rymnc (approved):
LGTM, thanks for this!

PR #1927, opened 2023-08-22
chore(ci): add a job checking config options and db schema

### LINKED ISSUE #1873: chore: add an action to check for config option changes
### Background

As nwaku is still under heavy development, we need to be (more) careful when changing config options (renaming, removing, but also adding) and make sure we announce this, keep backwards compatibility and notify various parties involved in running Waku nodes to make sure they have enough information and time to prepare for potentially breaking changes

### Details

We'd like to create a GH Action workflow which will check if any of the changes in PR is touching files defining config options (e.g. https://github.com/waku-org/nwaku/blob/master/apps/wakunode2/external_config.nim) and add a specific label and a comment to the PR

E.g. 

label: `breaking-change`
comment: This PR seems to be changing some configuration options in `$file`. Make sure this change is clearly highlighted in release notes and node operators are aware of required migration steps. 

You can take a look at our CI workflow - it also checks the modified files to only execute necessary steps - https://github.com/waku-org/nwaku/blob/master/.github/workflows/ci.yml#L27

### Acceptance criteria

- [ ] a workflow described above runs on each new or updated PR

vpavlin:
# Description

This PR adds a CI job checking changes in the PR and matching them agains files which contain

1. configuration options for apps
2. DB schema for archive drivers

If it matches, it will post a comment nudging the submitter to check whether there is a risk for a breaking change and also suggests to use `release-notes` label, to make sure the change is covered in Release Notes and changelog

I tried to articulate the comments clearly, but maybe there is a better wording/suggestion to submitter, please let me know what would be the information you'd like to see in the comment.

Also, let me know if I missed any files that should make it into the lists:)

# Changes

- [ ] new CI job checking for a fixed list of files in a PR and commenting on it in case it matches

## Issue

closes #1873

### DIFF
--- a/.github/workflows/pr-lint.yml
+++ b/.github/workflows/pr-lint.yml
@@ -48,3 +48,49 @@ jobs:
         with:   
           header: pr-title-lint-error
           delete: true
+
+  labels:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v3
+      name: Checkout code
+      id: checkout
+    - uses: dorny/paths-filter@v2
+      id: filter
+      with:
+        filters: |
+          config:
+          - 'apps/wakunode2/external_config.nim'
+          - 'apps/networkmonitor/networkmonitor_config.nim'
+          - 'apps/chat2/config_chat2.nim'
+          - 'apps/chat2bridge/config_chat2bridge.nim'
+
+          db_schema:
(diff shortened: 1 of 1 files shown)


Ivansete-status (approved):
Beautiful PR! Thanks for it!

NagyZoltanPeter (approved):
Cool stuff!

PR #1928, opened 2023-08-22
chore(rln_keystore_generator): generate and persist credentials

rymnc:
# Description

A follow up PR to #1925

# Changes

- [x] Creates an instance of the rln object
- [x] Generates credentials
- [x] Uses dummy chainId, contractAddress and treeIndex to write to a keystore

The next follow up PR will actually execute the registration if `--execute` is true

### DIFF
--- a/tools/rln_keystore_generator/external_config.nim
+++ b/tools/rln_keystore_generator/external_config.nim
@@ -51,7 +51,7 @@ type
       defaultValue: "",
       name: "rln-relay-eth-contract-address" }: string
 
-    rlnRelayCredentialsPassword* {.
+    rlnRelayCredPassword* {.
       desc: "Password for encrypting RLN credentials",
       defaultValue: "",
       name: "rln-relay-cred-password" }: string
@@ -59,6 +59,12 @@ type
 proc loadConfig*(T: type RlnKeystoreGeneratorConf): Result[T, string] =
   try:
     let conf = RlnKeystoreGeneratorConf.load()
+    if conf.rlnRelayCredPath == "":
+      return err("--rln-relay-cred-path must be set")
+    if conf.rlnRelayEthContractAddress == "":
+      return err("--rln-relay-eth-contract-address must be set")
+    if conf.rlnRelayCredPassword == "":
+      return err("--rln-relay-cred-password must be set")
     ok(conf)
   except CatchableError:
     err(getCurrentExceptionMsg())
--- a/tools/rln_keystore_generator/rln_keystore_generator.nim
+++ b/tools/rln_keystore_generator/rln_keystore_generator.nim
@@ -5,16 +5,21 @@ else:
 
 import
   chronicles,
-  stew/[results]
+  stew/[results],
+  std/tempfiles
 
 import
-    ./external_config
+  ../../waku/waku_keystore,
+  ../../waku/waku_rln_relay/rln,
+  ../../waku/waku_rln_relay/conversion_utils,
+  ./external_config
 
 logScope:
   topics = "rln_keystore_generator"
 
 when isMainModule:
   {.pop.}
+  # 1. load configuration
   let confRes = RlnKeystoreGeneratorConf.loadConfig()
   if confRes.isErr():
     error "failure while loading the configuration", error=confRes.error()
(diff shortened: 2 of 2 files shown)


tools/rln_keystore_generator/external_config.nim
```diff
@@ -51,14 +51,20 @@ type
...
       name: "rln-relay-cred-password" }: string
 
 proc loadConfig*(T: type RlnKeystoreGeneratorConf): Result[T, string] =
   try:
     let conf = RlnKeystoreGeneratorConf.load()
+    if conf.rlnRelayCredPath == "":
+      return err("--rln-relay-cred-path must be set")
+    if conf.rlnRelayEthContractAddress == "":
+      return err("--rln-relay-eth-contract-address must be set")
+    if conf.rlnRelayCredPassword == "":
```
Ivansete-status, line 66: Shall we also ask for a minimum password strength? ( e.g. minimum length, a-zA-Z, alphanumeric symbols, ... )

tools/rln_keystore_generator/rln_keystore_generator.nim
```diff
@@ -24,7 +29,48 @@ when isMainModule:
   
   debug "configuration", conf = $conf
 
-  # initialize keystore
+  # 2. initialize rlnInstance
+  let rlnInstanceRes = createRLNInstance(d=20, 
+                                         tree_path = genTempPath("rln_tree", "rln_keystore_generator"))
+  if rlnInstanceRes.isErr():
+    error "failure while creating RLN instance", error=rlnInstanceRes.error()
```
Ivansete-status, line 36: I think we tend not to use `()` for "name" attributes, although is being used in many places in this file
```suggestion
    error "failure while creating RLN instance", error=rlnInstanceRes.error
```

tools/rln_keystore_generator/rln_keystore_generator.nim
```diff
@@ -24,7 +29,48 @@ when isMainModule:
...
+    membershipGroups: @[MembershipGroup(
+      membershipContract: MembershipContract(
+        chainId: "1155511",
+        address: conf.rlnRelayEthContractAddress,
+      ),
+      treeIndex: 0,
+    )]
+  )
+
+  let persistRes = addMembershipCredentials(conf.rlnRelayCredPath, @[keystoreCred], conf.rlnRelayCredPassword, RLNAppInfo)
```
Ivansete-status, line 67: Better try to have the code as vertical as possible and each line less than 80 characters :)
```suggestion
  let persistRes = addMembershipCredentials(conf.rlnRelayCredPath,
                                                                         @[keystoreCred],
                                                                         conf.rlnRelayCredPassword,
                                                                         RLNAppInfo)
```

tools/rln_keystore_generator/rln_keystore_generator.nim
```diff
@@ -24,7 +29,48 @@ when isMainModule:
...
+                       idCommitment = credential.idCommitment.inHex()
+
+  # 4. write to keystore
+  ## TODO: after hooking up to the OnchainGroupManager, 
+  ## obtain chainId and treeIndex from the contract
+  let keystoreCred = MembershipCredentials(
+    identityCredential: credential,
+    membershipGroups: @[MembershipGroup(
+      membershipContract: MembershipContract(
+        chainId: "1155511",
```
Ivansete-status, line 60: Shall we have a comment on this value? Can this come from config as well?

Ivansete-status (approved):
LGTM as far as I could understand :P!
Just added minor comments

tools/rln_keystore_generator/rln_keystore_generator.nim
rymnc: This is a dummy value that will be replaced with the next iteration!

tools/rln_keystore_generator/external_config.nim
rymnc: can I cover this in a future iteration? tracking issue for it here - https://github.com/waku-org/nwaku/issues/1930

tools/rln_keystore_generator/rln_keystore_generator.nim
rymnc: Addressed in fed49b55

tools/rln_keystore_generator/rln_keystore_generator.nim
rymnc: Addressed in 131519d7

tools/rln_keystore_generator/external_config.nim
Ivansete-status: Yes of course:) Thanks!

alrevuelta: unsure if we can do that. is the keystore taken from some spec (eg EIP)? if so, if the EIP does not enforce any password strength, then we can enforce it on top.

eg, not the one we are using but https://eips.ethereum.org/EIPS/eip-2335

> The password is a string of arbitrary unicode characters. The password is first converted to its NFKD representation, then the control codes (specified below) are stripped from the password and finally it is UTF-8 encoded.

there are no requirements beyond that

alrevuelta (approved):

PR #1931, opened 2023-08-23
feat(rln_keystore_generator): wired to onchain group manager

rymnc:
# Description

Hooks up the rln_keystore_generator to the OnchainGroupManager. 

# Changes

- [ ] Integrates OnchainGroupManager
- [ ] Allows broadcasting transactions
- [ ] Writes to keystore appropriately


Makefile
```diff
@@ -186,7 +186,7 @@ chat2: | build deps librln
 
 rln_keystore_generator: | build deps librln
```
jm-clius, line 187: I'd go for `rln-keystore-generator`. General Makefile convention is that words in target names are not separated at all or separated by a hyphen. See https://style-guides.readthedocs.io/en/latest/makefile.html#targets

jm-clius (approved):
LGTM. Minor style comment

tools/rln_keystore_generator/rln_keystore_generator.nim
```diff
@@ -50,17 +51,44 @@ when isMainModule:
...
+                                         saveKeystore: false)
+  try:
+    waitFor groupManager.init()
+  except CatchableError:
+    error "failure while initializing OnchainGroupManager", error=getCurrentExceptionMsg()
+    quit(1)
+
+  # 5. register on-chain
+  try:
+    waitFor groupManager.register(credential)
```
alrevuelta, line 76: Solving this is not related to this PR but if we have a tool to register a membership in nwaku perhaps we should fix this https://github.com/status-im/nim-web3/issues/79 ? May take some time though.

I don't think you are able to broadcast a tx without a chaindId so this `groupManager.register` will most likely fail unless your Ethereum node uses `rpc.allow-unprotected-txs`, and afaik thats disabled in most of the cases. So someone using this tool will most likely run into this.

alrevuelta (approved):
lgtm

tools/rln_keystore_generator/rln_keystore_generator.nim
rymnc: very nice catch! will try to make some time to fix that!

PR #1932, opened 2023-08-23
fix(rln-relay): RLN DB should be aware of chain and contract address

### LINKED ISSUE #1921: bug(rln-relay): RLN DB should be aware of chain and contract address
### Problem
Currently the RLN metadata only saves the last processed block. If we change the RPC endpoint to point to a different chainID, or change the contract address, the information stored in RLN SledDB then the information stored in the database is not useful anymore, as the event logs that populate the database are associated to an specific chainID and contract address. However, rln-relay in its current version does not verify the chainID, so it will reuse the same database information even tho it is incorrect.

### Impact
Merkle tree with incorrect data is generated. RLN validation will be done incorrectly.

### Expected behavior
- [ ] SetMetadata should store the last processed block, the chainID, and contract address
- [ ] When mounting rln-relay, it should compare the value from `--rln-relay-eth-contract-address` flag against the contract address stored in the metadata, AND the result from calling `eth_chainId` RPC method with the chainID stored in the metadata. If there is no match, drop the information of the DB
- [ ] Consider having different databases per chainID and contract address (?)  (perhaps it's possible by specifying a different tree path, and instead of dropping the database information, to log a FATAL error indicating a wrong database path is being used?)

rymnc:
# Description

If an operator runs waku-rln-relay in 2 different runs, once with chainId x and another with chainId y, then the tree 
used would be the same. This leads to inconsistencies, and hence, we use the `metadata` feature to store the chainId, as 
well as the contractAddress. If there is a mismatch in any of the 2 fields, we raise an exception and prevent the node 
from starting up.

# Changes

- [x] added 2 fields `chainId`, `contractAddress` to `RlnMetadata`
- [x] updated tests, and serde

## Issue

closes #1921

### DIFF
--- a/tests/waku_rln_relay/test_rln_group_manager_onchain.nim
+++ b/tests/waku_rln_relay/test_rln_group_manager_onchain.nim
@@ -197,6 +197,29 @@ suite "Onchain group manager":
       manager.membershipFee.isSome()
       manager.initialized
 
+  asyncTest "should error on initialization when loaded metadata does not match":
+    let manager = await setup()
+    await manager.init()
+
+    let metadataSetRes = manager.setMetadata()
+    assert metadataSetRes.isOk(), metadataSetRes.error
+    let metadataRes = manager.rlnInstance.getMetadata()
+    assert metadataRes.isOk(), metadataRes.error
+    let metadata = metadataRes.get()
+    require:
+      metadata.chainId == 1337
+      metadata.contractAddress == manager.ethContractAddress
+    
+    await manager.stop()
+    
+    # simulating a change in the contractAddress
+    let manager2 = OnchainGroupManager(ethClientUrl: EthClient,
+                                       ethContractAddress: "0x0000000000000000000000000000000000000000",
+                                       ethPrivateKey: manager.ethPrivateKey,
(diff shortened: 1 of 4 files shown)


alrevuelta:
Thanks for this, indeed, very important. Can we have an extra unit test where we test this "a bit more end to end" persisting some info with chainId=a1, contractAddress=a2 and then attempting to load it from an rln instance that is configured with chainId=b2, contractAddress=b2 and assert it fails?

rymnc:
> Thanks for this, indeed, very important. Can we have an extra unit test where we test this "a bit more end to end" persisting some info with chainId=a1, contractAddress=a2 and then attempting to load it from an rln instance that is configured with chainId=b2, contractAddress=b2 and assert it fails?

Yeah, in progress! :)

alrevuelta (approved):
lgtm thanks!

Understand that this change is breaking right? I mean a node with this version will fail when loading an old database? Not a problem, since its an experimental feature. But just that others are aware.

rymnc:
> Understand that this change is breaking right? I mean a node with this version will fail when loading an old database? Not a problem, since its an experimental feature. But just that others are aware.

Correct, this is a breaking change ~ will communicate for the next release notes!

PR #1955, opened 2023-08-26
fix(rln-relay): sync from deployed block number

### LINKED ISSUE #1942: bug + chore: Unhandled exception + fix reducing blockChunkSize
This issue reports a bug and suggests a solution to fix it with some data to back it.

## Problem

When syncing a waku node with RLN with the following configuration. Note that YOUR_ENDPOINT is the nimbus.sepolia one.

```
--relay=true\
--rpc-admin=true\
--log-level=DEBUG\
--rpc-address=0.0.0.0\
--metrics-server=True\
--metrics-server-address=0.0.0.0\
--rln-relay=true \
--rln-relay-dynamic=true \
--rln-relay-cred-path=/data/rlnKeystore.json \
--rln-relay-tree-path=/data/rln_tree.db \
--rln-relay-eth-contract-address=0x2992c7bFD42729991b614e95F4C2C78619f49c50  \
--rln-relay-eth-client-address=YOUR_ENDPOINT
```

I get the following unhandled exception.

```
DBG 2023-08-24 07:39:22.999+00:00 rln metadata persisted                     topics="waku rln_relay onchain_group_manager" tid=1 file=group_manager.nim:316 blockNumber=4149736
ERR 2023-08-24 07:39:23.025+00:00 Ethereum client disconnected               topics="waku rln_relay onchain_group_manager" tid=1 file=group_manager.nim:523
INF 2023-08-24 07:39:23.025+00:00 reconnecting with the Ethereum client, and restarting group sync topics="waku rln_relay onchain_group_manager" tid=1 file=group_manager.nim:525 fromBlock=4149736
INF 2023-08-24 07:39:23.025+00:00 resuming onchain sync from block           topics="waku rln_relay onchain_group_manager" tid=1 file=group_manager.nim:352 fromBlock=4149736
/app/apps/wakunode2/wakunode2.nim(83) wakunode2
/app/apps/wakunode2/app.nim(492) setupAndMountProtocols
/app/vendor/nim-chronos/chronos/asyncfutures2.nim(373) futureContinue
/app/apps/wakunode2/app.nim(493) setupAndMountProtocols
/app/apps/wakunode2/app.nim(340) setupProtocols
/app/vendor/nim-chronos/chronos/asyncfutures2.nim(376) futureContinue
/app/apps/wakunode2/app.nim(411) setupProtocols
/app/waku/node/waku_node.nim(726) mountRlnRelay
/app/vendor/nim-chronos/chronos/asyncfutures2.nim(373) futureContinue
/app/waku/node/waku_node.nim(735) mountRlnRelay
/app/vendor/nim-chronos/chronos/asyncloop.nim(1251) waitFor
/app/vendor/nim-chronos/chronos/asyncloop.nim(269) poll
/app/vendor/nim-chronos/chronos/asyncfutures2.nim(376) futureContinue
/app/vendor/nim-json-rpc/json_rpc/clients/websocketclient.nim(76) processData
/app/waku/waku_rln_relay/group_manager/on_chain/group_manager.nim(527) colonanonymous
[[reraised from:
/app/apps/wakunode2/wakunode2.nim(83) wakunode2
/app/apps/wakunode2/app.nim(492) setupAndMountProtocols
/app/vendor/nim-chronos/chronos/asyncfutures2.nim(373) futureContinue
/app/apps/wakunode2/app.nim(493) setupAndMountProtocols
/app/apps/wakunode2/app.nim(340) setupProtocols
/app/vendor/nim-chronos/chronos/asyncfutures2.nim(376) futureContinue
/app/apps/wakunode2/app.nim(411) setupProtocols
/app/waku/node/waku_node.nim(726) mountRlnRelay
/app/vendor/nim-chronos/chronos/asyncfutures2.nim(373) futureContinue
/app/waku/node/waku_node.nim(735) mountRlnRelay
/app/vendor/nim-chronos/chronos/asyncloop.nim(1251) waitFor
/app/vendor/nim-chronos/chronos/asyncloop.nim(269) poll
/app/vendor/nim-chronos/chronos/asyncfutures2.nim(395) futureContinue
]]
[[reraised from:
/app/apps/wakunode2/wakunode2.nim(83) wakunode2
/app/apps/wakunode2/app.nim(492) setupAndMountProtocols
/app/vendor/nim-chronos/chronos/asyncfutures2.nim(373) futureContinue
/app/apps/wakunode2/app.nim(493) setupAndMountProtocols
/app/apps/wakunode2/app.nim(340) setupProtocols
/app/vendor/nim-chronos/chronos/asyncfutures2.nim(376) futureContinue
/app/apps/wakunode2/app.nim(411) setupProtocols
/app/waku/node/waku_node.nim(726) mountRlnRelay
/app/vendor/nim-chronos/chronos/asyncfutures2.nim(395) futureContinue
]]
[[reraised from:
/app/apps/wakunode2/wakunode2.nim(83) wakunode2
/app/apps/wakunode2/app.nim(492) setupAndMountProtocols
/app/vendor/nim-chronos/chronos/asyncfutures2.nim(373) futureContinue
/app/apps/wakunode2/app.nim(493) setupAndMountProtocols
/app/apps/wakunode2/app.nim(340) setupProtocols
/app/vendor/nim-chronos/chronos/asyncfutures2.nim(395) futureContinue
]]
[[reraised from:
/app/apps/wakunode2/wakunode2.nim(83) wakunode2
/app/apps/wakunode2/app.nim(492) setupAndMountProtocols
/app/vendor/nim-chronos/chronos/asyncfutures2.nim(395) futureContinue
]]
Error: unhandled exception: Asynchronous task [startOnchainSync() at group_manager.nim:342] finished with an exception "ValueError"!
Message: Transport is not initialised (missing a call to connect?)
Stack trace: /app/apps/wakunode2/wakunode2.nim(83) wakunode2
/app/apps/wakunode2/app.nim(492) setupAndMountProtocols
/app/vendor/nim-chronos/chronos/asyncfutures2.nim(373) futureContinue
/app/apps/wakunode2/app.nim(493) setupAndMountProtocols
/app/apps/wakunode2/app.nim(340) setupProtocols
/app/vendor/nim-chronos/chronos/asyncfutures2.nim(376) futureContinue
/app/apps/wakunode2/app.nim(411) setupProtocols
/app/waku/node/waku_node.nim(726) mountRlnRelay
/app/vendor/nim-chronos/chronos/asyncfutures2.nim(373) futureContinue
/app/waku/node/waku_node.nim(735) mountRlnRelay
/app/vendor/nim-chronos/chronos/asyncloop.nim(1251) waitFor
/app/vendor/nim-chronos/chronos/asyncloop.nim(269) poll
/app/vendor/nim-chronos/chronos/asyncfutures2.nim(376) futureContinue
/app/vendor/nim-json-rpc/json_rpc/clients/websocketclient.nim(76) processData
/app/waku/waku_rln_relay/group_manager/on_chain/group_manager.nim(527) colonanonymous
/app/waku/waku_rln_relay/group_manager/on_chain/group_manager.nim(342) startOnchainSync
/app/vendor/nim-chronos/chronos/asyncfutures2.nim(373) futureContinue
/app/waku/waku_rln_relay/group_manager/on_chain/group_manager.nim(358) startOnchainSync
/app/vendor/nimbus-build-system/vendor/Nim/lib/core/macros.nim(573) eth_blockNumber
/app/vendor/nim-chronos/chronos/asyncfutures2.nim(373) futureContinue
/app/vendor/nim-json-rpc/json_rpc/client.nim(157) eth_blockNumber
/app/vendor/nim-json-rpc/json_rpc/clients/websocketclient.nim(33) call
/app/vendor/nim-chronos/chronos/asyncfutures2.nim(373) futureContinue
/app/vendor/nim-json-rpc/json_rpc/clients/websocketclient.nim(38) call
[[reraised from:
/app/apps/wakunode2/wakunode2.nim(83) wakunode2
/app/apps/wakunode2/app.nim(492) setupAndMountProtocols
/app/vendor/nim-chronos/chronos/asyncfutures2.nim(373) futureContinue
/app/apps/wakunode2/app.nim(493) setupAndMountProtocols
/app/apps/wakunode2/app.nim(340) setupProtocols
/app/vendor/nim-chronos/chronos/asyncfutures2.nim(376) futureContinue
/app/apps/wakunode2/app.nim(411) setupProtocols
/app/waku/node/waku_node.nim(726) mountRlnRelay
/app/vendor/nim-chronos/chronos/asyncfutures2.nim(373) futureContinue
/app/waku/node/waku_node.nim(735) mountRlnRelay
/app/vendor/nim-chronos/chronos/asyncloop.nim(1251) waitFor
/app/vendor/nim-chronos/chronos/asyncloop.nim(269) poll
/app/vendor/nim-chronos/chronos/asyncfutures2.nim(376) futureContinue
/app/vendor/nim-json-rpc/json_rpc/clients/websocketclient.nim(76) processData
/app/waku/waku_rln_relay/group_manager/on_chain/group_manager.nim(527) colonanonymous
/app/waku/waku_rln_relay/group_manager/on_chain/group_manager.nim(342) startOnchainSync
/app/vendor/nim-chronos/chronos/asyncfutures2.nim(373) futureContinue
/app/waku/waku_rln_relay/group_manager/on_chain/group_manager.nim(358) startOnchainSync
/app/vendor/nimbus-build-system/vendor/Nim/lib/core/macros.nim(573) eth_blockNumber
/app/vendor/nim-chronos/chronos/asyncfutures2.nim(376) futureContinue
/app/vendor/nim-chronos/chronos/asyncmacro2.nim(300) eth_blockNumber
]]
[[reraised from:
/app/apps/wakunode2/wakunode2.nim(83) wakunode2
/app/apps/wakunode2/app.nim(492) setupAndMountProtocols
/app/vendor/nim-chronos/chronos/asyncfutures2.nim(373) futureContinue
/app/apps/wakunode2/app.nim(493) setupAndMountProtocols
/app/apps/wakunode2/app.nim(340) setupProtocols
/app/vendor/nim-chronos/chronos/asyncfutures2.nim(376) futureContinue
/app/apps/wakunode2/app.nim(411) setupProtocols
/app/waku/node/waku_node.nim(726) mountRlnRelay
/app/vendor/nim-chronos/chronos/asyncfutures2.nim(373) futureContinue
/app/waku/node/waku_node.nim(735) mountRlnRelay
/app/vendor/nim-chronos/chronos/asyncloop.nim(1251) waitFor
/app/vendor/nim-chronos/chronos/asyncloop.nim(269) poll
/app/vendor/nim-chronos/chronos/asyncfutures2.nim(376) futureContinue
/app/vendor/nim-json-rpc/json_rpc/clients/websocketclient.nim(76) processData
/app/waku/waku_rln_relay/group_manager/on_chain/group_manager.nim(527) colonanonymous
/app/waku/waku_rln_relay/group_manager/on_chain/group_manager.nim(342) startOnchainSync
/app/vendor/nim-chronos/chronos/asyncfutures2.nim(376) futureContinue
/app/vendor/nim-chronos/chronos/asyncmacro2.nim(300) startOnchainSync
]]
 [FutureDefect]
```

In spite of saying "Ethereum client disconnected" I suspect the root cause of this issue is because the amount of events `fromBlock` and `toBlock` is too high, so that the Ethereum node raises an exeption, and this exception is not handled properly. Actually I tried lowering the `blockChunkSize` and this error didn't appear.

## Suggested Solution

Context:
* When fetching events from Ethereum, you must specify a block range `fromBlock` `toBlock`. This issue is about choosing both values so that its diff works in most of the cases.
  * If `toBlock-fromBlock` is too **big**, you i) may choke the node ii) will get a massive amount of events impossible to handle or iii) the node will simple error saying it can't handle that.
  * If `toBlock-fromBlock` is too **low**, then it will take ages to sync.

I think we are in the case that `toBlock-fromBlock` is **too big**. Note that its not only about the diff between both values but also about the events that are inside. Even with 1 000 000 blocks, if there are no events it will most likely work, but since  I'm using a contract with thousands of registrations made in few days `0x2992c7bFD42729991b614e95F4C2C78619f49c50` if you hit the time interval of this thousand of registrations, then it will fail.

So I think [blockChunkSize](https://github.com/waku-org/nwaku/blob/master/waku/waku_rln_relay/group_manager/on_chain/group_manager.nim#L349C7-L349C21) is too big. Even with the `DecayFactor` that lowers the range 1.2 times on every run as:
* 1 000 000
* 833 333
* 694 444
* 482 252
* 159 238
* ...

The block ranges i) are too high and ii) it doesn't make sense to have such decay. Taking as starting point when the contract was deployed, why having such decay? Each time range will have the same probability of having events in it, so why having a variable chunk size? Once we have `9 Sync from the deployment block instead of from genesis.` from https://github.com/waku-org/nwaku/issues/1906 then the decay wont be needed imho.

Ideally people will run their own Ethereum node, but we must also pick a `blockChunkSize` so that waku can be synced with any external rpc provider such as infura or alchemy. And here we hit some limitations, that are very reasonable imho. See below:
* alchemy: `In order to ensure maximum reliability for your calls, you can make eth_getLogs requests with up to a 2K block range and a 150MB limit on the response size. You can also request any block range with a cap of 10K logs in the response.` [source](https://docs.alchemy.com/discuss/62d157025ddb500013a2d796)
* quicknode: `eth_getLogs and eth_newFilter are limited to a 10,000 blocks range`
* infura: To prevent queries from consuming too many resources, eth_getLogs requests are currently limited by three constraints: [source](https://docs.infura.io/networks/ethereum/json-rpc-methods/eth_getlogs)
  * A maximuim of 5,000 parameters in a single request
  * A maximum of 10,000 results can be returned by a single query
  * Query duration must not exceed 10 seconds

### 5k benchmark
* Did a test syncing from block 0 to latest sepolia block 4150223 (more than 5 years of history) **using blockChunkSize=5_000** and it took 5 minutes to sync. Ofc once we sync from deployment, we won't need to sync from genesis and will save up a lot of time. This is around 800 requests, so any free tier would be ok with that (infura limits 100k/day). And the node didn't failed (no exception raised).

## Action points:
- [ ] Wait until "9 Sync from the deployment block instead of from genesis" to be completed is completed. See https://github.com/waku-org/nwaku/issues/1906
- [ ] Remove `DecayFactor`
- [ ] Set `blockChunkSize` to 5k (or 2k to be ok with alchemy)
- [ ] Handle the shown exception

rymnc:
# Description

Syncs from the deployed block number to prevent excess network calls

# Changes

- [x] removed decay factor, static chunk size of 2000
- [x] handles eth rpc reconnection
- [x] syncs from deployment block of the storage contract

## Issue

closes #1942

### DIFF
--- a/waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
+++ b/waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
@@ -39,6 +39,7 @@ contract(RlnStorage):
   proc MEMBERSHIP_DEPOSIT(): Uint256 {.pure.}
   proc members(idCommitment: Uint256): Uint256 {.view.}
   proc idCommitmentIndex(): Uint256 {.view.}
+  proc deployedBlockNumber(): Uint256 {.view.}
 
 type
   RegistryContractWithSender = Sender[WakuRlnRegistry]
@@ -49,6 +50,7 @@ type
     ethContractAddress*: string
     ethRpc*: Option[Web3]
     rlnContract*: Option[RlnContractWithSender]
+    rlnContractDeployedBlockNumber*: Option[BlockNumber]
     registryContract*: Option[RegistryContractWithSender]
     usingStorageIndex: Option[Uint16]
     membershipFee*: Option[Uint256]
@@ -67,9 +69,6 @@ type
 
 const DefaultKeyStorePath* = "rlnKeystore.json"
 const DefaultKeyStorePassword* = "password"
-
-const DecayFactor* = 1.2
-const DefaultChunkSize* = 1000
 
 template initializedGuard(g: OnchainGroupManager): untyped =
   if not g.initialized:
@@ -353,16 +352,17 @@ proc startOnchainSync(g: OnchainGroupManager): Future[void] {.async.} =
 
   let ethRpc = g.ethRpc.get()
 
-  # the block chunk size decays exponentially with the number of blocks
-  # the minimum chunk size is 100
-  var blockChunkSize = 1_000_000
+  # static block chunk size
+  let blockChunkSize = 2_000
 
-  var fromBlock = if g.latestProcessedBlock.isSome():
+  var fromBlock = if g.latestProcessedBlock.isSome() and 
+                  g.latestProcessedBlock.get() > g.rlnContractDeployedBlockNumber.get():
     info "resuming onchain sync from block", fromBlock = g.latestProcessedBlock.get()
     g.latestProcessedBlock.get() + 1
   else:
-    info "starting onchain sync from scratch"
-    BlockNumber(0)
+    let deployedBlockNumber = g.rlnContractDeployedBlockNumber.get()
+    info "starting onchain sync from deployed block number", deployedBlockNumber = deployedBlockNumber
+    deployedBlockNumber
 
   let latestBlock = cast[BlockNumber](await ethRpc.provider.eth_blockNumber())
   try:
@@ -378,8 +378,6 @@ proc startOnchainSync(g: OnchainGroupManager): Future[void] {.async.} =
         fromBlock = toBlock + 1
         if fromBlock >= currentLatestBlock:
           break
-        let newChunkSize = float(blockChunkSize) / DecayFactor
-        blockChunkSize = max(int(newChunkSize), DefaultChunkSize)
     else:
       await g.getAndHandleEvents(fromBlock, some(BlockNumber(0)))
   except CatchableError:
@@ -477,10 +475,25 @@ method init*(g: OnchainGroupManager): Future[void] {.async.} =
                        "could not get the membership deposit: " & getCurrentExceptionMsg())
   g.membershipFee = some(membershipFee)
 
+  var deployedBlockNumber: Uint256
+  try:
+    deployedBlockNumber = await rlnContract.deployedBlockNumber().call()
+  except CatchableError:
+    raise newException(ValueError, 
+                       "could not get the deployed block number: " & getCurrentExceptionMsg())
+  g.rlnContractDeployedBlockNumber = some(cast[BlockNumber](deployedBlockNumber))
+
   ethRpc.ondisconnect = proc() =
     error "Ethereum client disconnected"
-    let fromBlock = g.latestProcessedBlock.get()
+    let fromBlock = max(g.latestProcessedBlock.get(), g.rlnContractDeployedBlockNumber.get())
     info "reconnecting with the Ethereum client, and restarting group sync", fromBlock = fromBlock
+    try:
+      let newEthRpc = waitFor newWeb3(g.ethClientUrl)
+      newEthRpc.ondisconnect = ethRpc.ondisconnect
+      g.ethRpc = some(newEthRpc)
+    except CatchableError:
+      error "failed to reconnect with the Ethereum client", error = getCurrentExceptionMsg()
+      return
     try:
       asyncSpawn g.startOnchainSync()
     except CatchableError:
--- a/waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
+++ b/waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
@@ -76,8 +76,6 @@ template initializedGuard(g: OnchainGroupManager): untyped =
 
 
 proc setMetadata*(g: OnchainGroupManager): RlnRelayResult[void] =
-  if g.latestProcessedBlock == 0:
-    return err("latest processed block is not set")
   try:
     let metadataSetRes = g.rlnInstance.setMetadata(RlnMetadata(
                             lastProcessedBlock: g.latestProcessedBlock,
@@ -358,9 +356,8 @@ proc startOnchainSync(g: OnchainGroupManager): Future[void] {.async.} =
     info "resuming onchain sync from block", fromBlock = g.latestProcessedBlock
     g.latestProcessedBlock + 1
   else:
-    let deployedBlockNumber = g.rlnContractDeployedBlockNumber
-    info "starting onchain sync from deployed block number", deployedBlockNumber = deployedBlockNumber
-    deployedBlockNumber
+    info "starting onchain sync from deployed block number", deployedBlockNumber = g.rlnContractDeployedBlockNumber
+    g.rlnContractDeployedBlockNumber
 
   let latestBlock = cast[BlockNumber](await ethRpc.provider.eth_blockNumber())
   try:
--- a/waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
+++ b/waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
@@ -50,11 +50,11 @@ type
     ethContractAddress*: string
     ethRpc*: Option[Web3]
     rlnContract*: Option[RlnContractWithSender]
-    rlnContractDeployedBlockNumber*: Option[BlockNumber]
+    rlnContractDeployedBlockNumber*: BlockNumber
     registryContract*: Option[RegistryContractWithSender]
     usingStorageIndex: Option[Uint16]
     membershipFee*: Option[Uint256]
-    latestProcessedBlock*: Option[BlockNumber]
+    latestProcessedBlock*: BlockNumber
     registrationTxHash*: Option[TxHash]
     chainId*: Option[Quantity]
     keystorePath*: Option[string]
@@ -76,11 +76,11 @@ template initializedGuard(g: OnchainGroupManager): untyped =
 
 
 proc setMetadata*(g: OnchainGroupManager): RlnRelayResult[void] =
-  if g.latestProcessedBlock.isNone():
+  if g.latestProcessedBlock == 0:
     return err("latest processed block is not set")
   try:
     let metadataSetRes = g.rlnInstance.setMetadata(RlnMetadata(
-                            lastProcessedBlock: g.latestProcessedBlock.get(),
+                            lastProcessedBlock: g.latestProcessedBlock,
                             chainId: uint64(g.chainId.get()),
                             contractAddress: g.ethContractAddress,
                             validRoots: g.validRoots.toSeq()))
@@ -313,15 +313,14 @@ proc getAndHandleEvents(g: OnchainGroupManager,
   await g.handleEvents(blockTable)
   await g.handleRemovedEvents(blockTable)
 
-  let latestProcessedBlock = if toBlock.isSome(): toBlock.get()
+  g.latestProcessedBlock = if toBlock.isSome(): toBlock.get()
                              else: fromBlock
-  g.latestProcessedBlock = some(latestProcessedBlock)
   let metadataSetRes = g.setMetadata()
   if metadataSetRes.isErr():
     # this is not a fatal error, hence we don't raise an exception
     warn "failed to persist rln metadata", error=metadataSetRes.error()
   else:
-    trace "rln metadata persisted", blockNumber = latestProcessedBlock
+    trace "rln metadata persisted", blockNumber = g.latestProcessedBlock
 
 proc getNewHeadCallback(g: OnchainGroupManager): BlockHeaderHandler =
   proc newHeadCallback(blockheader: BlockHeader) {.gcsafe.} =
@@ -355,12 +354,11 @@ proc startOnchainSync(g: OnchainGroupManager): Future[void] {.async.} =
   # static block chunk size
   let blockChunkSize = 2_000
 
-  var fromBlock = if g.latestProcessedBlock.isSome() and 
-                  g.latestProcessedBlock.get() > g.rlnContractDeployedBlockNumber.get():
-    info "resuming onchain sync from block", fromBlock = g.latestProcessedBlock.get()
-    g.latestProcessedBlock.get() + 1
+  var fromBlock = if g.latestProcessedBlock > g.rlnContractDeployedBlockNumber:
+    info "resuming onchain sync from block", fromBlock = g.latestProcessedBlock
+    g.latestProcessedBlock + 1
   else:
-    let deployedBlockNumber = g.rlnContractDeployedBlockNumber.get()
+    let deployedBlockNumber = g.rlnContractDeployedBlockNumber
     info "starting onchain sync from deployed block number", deployedBlockNumber = deployedBlockNumber
     deployedBlockNumber
 
@@ -455,7 +453,7 @@ method init*(g: OnchainGroupManager): Future[void] {.async.} =
   let metadataGetRes = g.rlnInstance.getMetadata()
   if metadataGetRes.isErr():
     warn "could not initialize with persisted rln metadata"
-    g.latestProcessedBlock = some(BlockNumber(0))
+    g.latestProcessedBlock = BlockNumber(0)
   else:
     let metadata = metadataGetRes.get()
     if metadata.chainId != uint64(g.chainId.get()):
@@ -463,7 +461,7 @@ method init*(g: OnchainGroupManager): Future[void] {.async.} =
   
     if metadata.contractAddress != g.ethContractAddress.toLower():
       raise newException(ValueError, "persisted data: contract address mismatch")
-    g.latestProcessedBlock = some(metadata.lastProcessedBlock)
+    g.latestProcessedBlock = metadata.lastProcessedBlock
     g.validRoots = metadata.validRoots.toDeque()
 
   # check if the contract exists by calling a static function
@@ -481,11 +479,11 @@ method init*(g: OnchainGroupManager): Future[void] {.async.} =
   except CatchableError:
     raise newException(ValueError, 
                        "could not get the deployed block number: " & getCurrentExceptionMsg())
-  g.rlnContractDeployedBlockNumber = some(cast[BlockNumber](deployedBlockNumber))
+  g.rlnContractDeployedBlockNumber = cast[BlockNumber](deployedBlockNumber)
 
   ethRpc.ondisconnect = proc() =
     error "Ethereum client disconnected"
-    let fromBlock = max(g.latestProcessedBlock.get(), g.rlnContractDeployedBlockNumber.get())
+    let fromBlock = max(g.latestProcessedBlock, g.rlnContractDeployedBlockNumber)
     info "reconnecting with the Ethereum client, and restarting group sync", fromBlock = fromBlock
     try:
       let newEthRpc = waitFor newWeb3(g.ethClientUrl)
--- a/tests/waku_rln_relay/test_rln_group_manager_onchain.nim
+++ b/tests/waku_rln_relay/test_rln_group_manager_onchain.nim
@@ -191,8 +191,7 @@ suite "Onchain group manager":
       manager.rlnContract.isSome()
       manager.membershipFee.isSome()
       manager.initialized
-      manager.rlnContractDeployedBlockNumber.isSome() 
-      manager.rlnContractDeployedBlockNumber.get() > 0
+      manager.rlnContractDeployedBlockNumber > 0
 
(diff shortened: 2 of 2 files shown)


waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
```diff
@@ -477,10 +475,25 @@ method init*(g: OnchainGroupManager): Future[void] {.async.} =
...
+                       "could not get the deployed block number: " & getCurrentExceptionMsg())
+  g.rlnContractDeployedBlockNumber = some(cast[BlockNumber](deployedBlockNumber))
+
   ethRpc.ondisconnect = proc() =
     error "Ethereum client disconnected"
-    let fromBlock = g.latestProcessedBlock.get()
+    let fromBlock = max(g.latestProcessedBlock.get(), g.rlnContractDeployedBlockNumber.get())
     info "reconnecting with the Ethereum client, and restarting group sync", fromBlock = fromBlock
+    try:
+      let newEthRpc = waitFor newWeb3(g.ethClientUrl)
```
rymnc, line 491: maybe we should have a maximum number of reconnection attempts here to prevent connecting to a dead node forever - can address in a follow up pr

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
```diff
@@ -49,6 +50,7 @@ type
     ethContractAddress*: string
     ethRpc*: Option[Web3]
     rlnContract*: Option[RlnContractWithSender]
+    rlnContractDeployedBlockNumber*: Option[BlockNumber]
```
alrevuelta, line 53: perhaps this should be mandatory?

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
```diff
@@ -353,16 +352,17 @@ proc startOnchainSync(g: OnchainGroupManager): Future[void] {.async.} =
...
 
-  var fromBlock = if g.latestProcessedBlock.isSome():
+  var fromBlock = if g.latestProcessedBlock.isSome() and 
+                  g.latestProcessedBlock.get() > g.rlnContractDeployedBlockNumber.get():
     info "resuming onchain sync from block", fromBlock = g.latestProcessedBlock.get()
     g.latestProcessedBlock.get() + 1
   else:
-    info "starting onchain sync from scratch"
-    BlockNumber(0)
+    let deployedBlockNumber = g.rlnContractDeployedBlockNumber.get()
```
alrevuelta, line 363: related to my other comment, of having `rlnContractDeployedBlockNumber` as an Option.

here we are not handling if `isSome` so technically it could raise an unhandled exception? Perhaps having the field as mandatory also simplifies this.

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
```diff
@@ -353,16 +352,17 @@ proc startOnchainSync(g: OnchainGroupManager): Future[void] {.async.} =
...
   let ethRpc = g.ethRpc.get()
 
-  # the block chunk size decays exponentially with the number of blocks
-  # the minimum chunk size is 100
-  var blockChunkSize = 1_000_000
+  # static block chunk size
+  let blockChunkSize = 2_000
 
-  var fromBlock = if g.latestProcessedBlock.isSome():
+  var fromBlock = if g.latestProcessedBlock.isSome() and
```
alrevuelta, line 358: similar comment regarding `latestProcessedBlock`. Why being an option? In this case every nwaku rln instance will have a latest processed block.

Have the feeling that you are treating `0` as `.isNone` and perhaps that can be simplified?

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
```diff
@@ -49,6 +50,7 @@ type
     ethContractAddress*: string
     ethRpc*: Option[Web3]
```
alrevuelta, line 51: not strictly related to the PR but related to my other comments.

any reason for having `ethRpc` as an option? I believe every onchain group manager would need one, or?

the less options the less we have to check if `isSome` and the less risk we have of doing `.get` in a `none`.

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
alrevuelta: yep, mind tracking it?

Not that concern about infinite reconections as long as the node is aware that "its not ready" (eg via the `/health` endpoint) so not a high prio.

But perhaps we need some delay? to avoid an infinite connect-disconnect-connect loop?

alrevuelta (approved):
lgtm! Thanks for this, a great improvement that will i) reduce rln tree sync time by a lot and ii) avoid choking eth nodes by requesting less blocks at once (something quite important when using your own non infura node)

left some comments, nothing critical but relevant imho.

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
rymnc: The reason the ethRpc is an option is because it's set in the `init*` method. all methods which utilize the optional fields, like `latestProcessedBlock` and `rlnContractDeployedBlockNumber`, are guarded by https://github.com/waku-org/nwaku/blob/3674879b0e1c072303b9d6524462c20656c9a82b/waku/waku_rln_relay/group_manager/on_chain/group_manager.nim#L73-L75 which ensures that the groupManager is initialized before any other operations are performed on it (and by extension, the `ethRpc`, `latestProcessedBlock`, and `rlnContractDeployedBlockNumber`)
I could change them to be non-options, but they'd be nil for a while, and if someone accessed them before init, then would throw a segfault, right?

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
rymnc: see https://github.com/waku-org/nwaku/pull/1955#discussion_r1307034035

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
alrevuelta: ah right, feel free to leave the `ethRpc` as it is.

it was a personal preference. i prefer to check for `isNil` which then avoids `.get()`, its cleaner due to `ethRpc*: Web3` and easier to read (no templates). biased by golang i guess :)

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
alrevuelta: in this case I think its a bit different. Isn't this treating 0 as `isNone`? When we can just remove the option?

rymnc: hmm here we're checking if it's safe to `.get()` the value from g.latestProcessedBlock

you're right, we can update it to not be an option here!

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
rymnc: Addressed in 7ca355e9!

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
rymnc: Addressed in 7ca355e9!

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
rymnc: yup max-retries + backoff

alrevuelta:
🚀 thanks!

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
```diff
@@ -355,12 +354,11 @@ proc startOnchainSync(g: OnchainGroupManager): Future[void] {.async.} =
...
-  var fromBlock = if g.latestProcessedBlock.isSome() and 
-                  g.latestProcessedBlock.get() > g.rlnContractDeployedBlockNumber.get():
-    info "resuming onchain sync from block", fromBlock = g.latestProcessedBlock.get()
-    g.latestProcessedBlock.get() + 1
+  var fromBlock = if g.latestProcessedBlock > g.rlnContractDeployedBlockNumber:
+    info "resuming onchain sync from block", fromBlock = g.latestProcessedBlock
+    g.latestProcessedBlock + 1
   else:
-    let deployedBlockNumber = g.rlnContractDeployedBlockNumber.get()
+    let deployedBlockNumber = g.rlnContractDeployedBlockNumber
```
alrevuelta, line 361: nitpick, but not needed?

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
```diff
@@ -76,11 +76,11 @@ template initializedGuard(g: OnchainGroupManager): untyped =
 
 
 proc setMetadata*(g: OnchainGroupManager): RlnRelayResult[void] =
-  if g.latestProcessedBlock.isNone():
+  if g.latestProcessedBlock == 0:
```
alrevuelta, line 79: mmm do we still need to error here? guess 0 could be an ok value? eg at startup?

alrevuelta (commented):
couple of comments, one might be important?

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
rymnc: there shouldn't be a reason to save the metadata if the latest block number is 0, hence there were 2 options here, either a noop or an error, went for the latter

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
rymnc: Addressed in 110a8a84

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
```diff
@@ -39,6 +39,7 @@ contract(RlnStorage):
   proc MEMBERSHIP_DEPOSIT(): Uint256 {.pure.}
   proc members(idCommitment: Uint256): Uint256 {.view.}
   proc idCommitmentIndex(): Uint256 {.view.}
+  proc deployedBlockNumber(): Uint256 {.view.}
```
Ivansete-status, line 42: Looks nice indeed! However, I wonder if we could add comments above each of these procs to elaborate more about their purposes.

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
```diff
@@ -477,10 +470,25 @@ method init*(g: OnchainGroupManager): Future[void] {.async.} =
...
+                       "could not get the deployed block number: " & getCurrentExceptionMsg())
+  g.rlnContractDeployedBlockNumber = cast[BlockNumber](deployedBlockNumber)
+
   ethRpc.ondisconnect = proc() =
     error "Ethereum client disconnected"
-    let fromBlock = g.latestProcessedBlock.get()
+    let fromBlock = max(g.latestProcessedBlock, g.rlnContractDeployedBlockNumber)
     info "reconnecting with the Ethereum client, and restarting group sync", fromBlock = fromBlock
+    try:
+      let newEthRpc = waitFor newWeb3(g.ethClientUrl)
```
Ivansete-status, line 486: I wonder if makes sense to perform some retrial mechanism?

Ivansete-status (approved):
LGTM! thanks!

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
rymnc: hi! https://github.com/waku-org/nwaku/issues/1958 tracks handling delays, this piece of code does handle the reconnections!

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
rymnc: will address in a follow up pr :)

PR #1957, opened 2023-08-28
chore(cbindings): Thread-safe libwaku. WakuNode instance created directly from the Waku Thread

Ivansete-status:
# Description
This is aimed at having a safer c-bindings `libwaku`. With that regard, given that the [WakuNode](https://github.com/waku-org/nwaku/blob/01634f57f0386efebfcfcb1d6e6c88de345a17e7/waku/node/waku_node.nim#L85) is a GC-managed type, we need to make sure the instance of the _WakuNode_ is directly created from within the _Waku Thread_ itself.

note: each `ref`, `seq[T]`, or `string` are GC-manged types

# Changes

- [x] library/waku_thread/inter_thread_communication/node_lifecycle_request.nim -> New `CREATE_NODE` operation type.
- [x] library/libwaku.nim -> Splitting the _Waku Node_ creation in two parts: 1) create the _Waku Thread_. 2) send a `CREATE_NODE` req to the _Waku Thread_ so that it creates the _Waku Node_ instance.
- [x] examples/cbindings/waku_example.c -> fix so that `make cwaku_example` actually compiles.

## Issue
https://github.com/waku-org/nwaku/issues/1878

### DIFF
--- a/library/alloc.nim
+++ b/library/alloc.nim
@@ -2,6 +2,6 @@
 proc alloc*(str: cstring): cstring =
   # Byte allocation from the given address.
   # There should be the corresponding manual deallocation with deallocShared !
-  let ret = cast[cstring](allocShared(len(str) + 1))
+  let ret = cast[cstring](allocShared0(len(str) + 1))
   copyMem(ret, str, len(str) + 1)
   return ret
--- a/library/alloc.nim
+++ b/library/alloc.nim
@@ -2,6 +2,6 @@
 proc alloc*(str: cstring): cstring =
   # Byte allocation from the given address.
   # There should be the corresponding manual deallocation with deallocShared !
-  let ret = cast[cstring](allocShared0(len(str) + 1))
+  let ret = cast[cstring](allocShared(len(str) + 1))
   copyMem(ret, str, len(str) + 1)
   return ret
--- a/library/libwaku.nim
+++ b/library/libwaku.nim
@@ -73,9 +73,8 @@ proc waku_new(configJson: cstring,
     return RET_MISSING_CALLBACK
 
   ## Create the Waku thread that will keep waiting for req from the main thread.
-  let createThRes = waku_thread.createWakuThread()
-  if createThRes.isErr():
-    let msg = "Error in createWakuThread: " & $createThRes.error
+  waku_thread.createWakuThread().isOkOr:
+    let msg = "Error in createWakuThread: " & $error
     onErrCb(unsafeAddr msg[0], cast[csize_t](len(msg)))
     return RET_ERR
 
--- a/library/libwaku.nim
+++ b/library/libwaku.nim
@@ -72,9 +72,19 @@ proc waku_new(configJson: cstring,
   if isNil(onErrCb):
     return RET_MISSING_CALLBACK
 
-  let createThRes = waku_thread.createWakuThread(configJson)
+  ## Create the Waku thread that will keep waiting for req from the main thread.
+  let createThRes = waku_thread.createWakuThread()
   if createThRes.isErr():
     let msg = "Error in createWakuThread: " & $createThRes.error
+    onErrCb(unsafeAddr msg[0], cast[csize_t](len(msg)))
+    return RET_ERR
+
+  let sendReqRes = waku_thread.sendRequestToWakuThread(
+                                NodeLifecycleRequest.new(
+                                          NodeLifecycleMsgType.CREATE_NODE,
+                                          configJson))
+  if sendReqRes.isErr():
+    let msg = $sendReqRes.error
     onErrCb(unsafeAddr msg[0], cast[csize_t](len(msg)))
     return RET_ERR
 
--- a/library/waku_thread/inter_thread_communication/node_lifecycle_request.nim
+++ b/library/waku_thread/inter_thread_communication/node_lifecycle_request.nim
@@ -81,13 +81,16 @@ proc createNode(configJson: cstring): Result[WakuNode, string] =
 
     else: recordRes.get()
 
+  ## TODO: make the next const configurable from 'configJson'.
+  const MAX_CONNECTIONS = 50.int
+
   var builder = WakuNodeBuilder.init()
   builder.withRng(crypto.newRng())
   builder.withNodeKey(privateKey)
   builder.withRecord(record)
   builder.withNetworkConfiguration(netConfig)
   builder.withSwitchConfiguration(
-    maxConnections = some(50.int)
+    maxConnections = some(MAX_CONNECTIONS)
   )
 
   let wakuNodeRes = builder.build()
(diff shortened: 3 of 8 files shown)


library/waku_thread/inter_thread_communication/node_lifecycle_request.nim
```diff
@@ -6,32 +6,118 @@ import
...
+
+    else: recordRes.get()
+
+  var builder = WakuNodeBuilder.init()
+  builder.withRng(crypto.newRng())
+  builder.withNodeKey(privateKey)
+  builder.withRecord(record)
+  builder.withNetworkConfiguration(netConfig)
+  builder.withSwitchConfiguration(
+    maxConnections = some(50.int)
```
vpavlin, line 90: Should this be really hardcoded? Can you at least make it a constant?

library/waku_thread/inter_thread_communication/node_lifecycle_request.nim
```diff
@@ -6,32 +6,118 @@ import
...
-                node: WakuNode): Future[Result[string, string]] {.async.} =
+                node: ptr WakuNode): Future[Result[string, string]] {.async.} =
 
   case self.operation:
+    of CREATE_NODE:
+      let newNodeRes = createNode(self.configJson)
+      if newNodeRes.isErr():
+        return err(newNodeRes.error)
+
+      node[] = newNodeRes.get()
```
vpavlin, line 115: What does this notation (the `[]`) mean/do?

library/alloc.nim
```diff
@@ -2,6 +2,6 @@
 proc alloc*(str: cstring): cstring =
   # Byte allocation from the given address.
   # There should be the corresponding manual deallocation with deallocShared !
-  let ret = cast[cstring](allocShared(len(str) + 1))
+  let ret = cast[cstring](allocShared0(len(str) + 1))
```
arnetheduck, line 5: there's no need for 0 here since the full string will be overwritten

library/waku_thread/inter_thread_communication/node_lifecycle_request.nim
Ivansete-status: Thanks for the comments! I'll fix them asap.
This `[]` is used for `ptr` types. I understand the `[]` operator as a way to dereference the ptr, i.e. a way to access the value or object it points to.

Something interesting I've learned recently:

| ref obj A | ptr A |
|--------|--------|
| Points to data stored in the heap | Points to data stored either in the heap or stack |
| GC'ed | The developer should take care of allocating and deallocating manually |
| No thread-safe | Thread-safe |

library/libwaku.nim
```diff
@@ -72,12 +72,22 @@ proc waku_new(configJson: cstring,
   if isNil(onErrCb):
     return RET_MISSING_CALLBACK
 
-  let createThRes = waku_thread.createWakuThread(configJson)
+  ## Create the Waku thread that will keep waiting for req from the main thread.
+  let createThRes = waku_thread.createWakuThread()
   if createThRes.isErr():
     let msg = "Error in createWakuThread: " & $createThRes.error
     onErrCb(unsafeAddr msg[0], cast[csize_t](len(msg)))
     return RET_ERR
```
arnetheduck, line 80:
```suggestion
  let createTh = waku_thread.createWakuThread().valueOr:
    let msg = "Error in createWakuThread: " & $error
    onErrCb(unsafeAddr msg[0], cast[csize_t](len(msg)))
    return RET_ERR
```

library/waku_thread/inter_thread_communication/node_lifecycle_request.nim
```diff
@@ -6,32 +6,121 @@ import
...
 
     of START_NODE:
-      waitFor node.start()
+      waitFor node[].start()
```
arnetheduck, line 121: `{.async.}` functions must never call `waitFor` - instead, they should use `await`

library/libwaku.nim
arnetheduck: or.. if this is `Result[void]`, use `isOkOr:`

library/waku_thread/inter_thread_communication/node_lifecycle_request.nim
```diff
@@ -6,32 +6,121 @@ import
...
+
+  if relay:
+    waitFor newNode.mountRelay()
+    newNode.peerManager.start()
+
+  return ok(newNode)
 
 method process*(self: NodeLifecycleRequest,
-                node: WakuNode): Future[Result[string, string]] {.async.} =
+                node: ptr WakuNode): Future[Result[string, string]] {.async.} =
```
arnetheduck, line 110: `WakuNode` is a `ref object` - the previous code without the `ptr` was correct - a consequence is that `process` is not a "thread-safe" function - it can only be called from the thread that created the `WakuNode` instance (ie the processing loop below)

arnetheduck: ah, I see now - this looks as if you wanted to return the node instance and `async` prevents `var` from being used - ok, `ptr` can be used as a workaround for that limitation as long as you keep in mind that a ptr might become dangling - this is a bit of a general problem here and for this PR it's an ok tradeoff to improve the status quo - later, I would probably look for solutions that avoid this quirk

library/waku_thread/inter_thread_communication/node_lifecycle_request.nim
Ivansete-status: ah yes, my bad!

library/waku_thread/inter_thread_communication/node_lifecycle_request.nim
Ivansete-status: Well, the idea of using `ptr` was to alter the node instance: https://github.com/waku-org/nwaku/blob/f29105b6fcb617b975a9815d9c14db82d2fc1148/library/waku_thread/waku_thread.nim#L48

This instance will live enough for all the calls to any `process` proc and will always be owned by the _Waku Thread_.

library/libwaku.nim
Ivansete-status: Super interesting! Thanks for that! Notice that I will only apply that to this snippet and I'm planning to extend this approach to further places at `nwaku`.

library/waku_thread/inter_thread_communication/node_lifecycle_request.nim
vpavlin: Thank you very much!

vpavlin (approved):
lgtm

Ivansete-status:
I will merge this as we have another PR that is blocked by that one.
Thank you so much indeed for the comments guys!
@arnetheduck - Although the PR will get merged, I can revisit any further suggestions that may arise. Thank you very much for the comments so far. I'll set you as a reviewer for the upcoming PR, where we avoid the use of the `Channel` type.

PR #1960, opened 2023-08-28
fix(Makefile): error out if rln-keystore-generator not compiled with rln flag

rymnc:
# Description

This pr is associated to https://github.com/waku-org/nwaku/pull/1956#issuecomment-1695562089
We need to ensure that RLN is set to true when compiling the `rln-keystore-generator` target

# Changes

- [x] Prevents rebuilding of rln by fixing the LIBRLN_BUILDDIR
- [x] conditional compilation of rln-keystore-generator

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -135,11 +135,14 @@ $(LIBRLN_BUILDDIR)/$(LIBRLN_FILE):
 	echo -e $(BUILD_MSG) "$@" && \
 		./scripts/build_rln.sh $(LIBRLN_BUILDDIR)
 
+librln-experimental:
+EXPERIMENTAL_PARAMS += -d:rln --passL:$(LIBRLN_FILE) --passL:-lm
+librln: $(LIBRLN_BUILDDIR)/$(LIBRLN_FILE)
+
 ifneq ($(RLN), true)
 librln: ; # noop
 else
-EXPERIMENTAL_PARAMS += -d:rln --passL:$(LIBRLN_FILE) --passL:-lm
-librln: $(LIBRLN_BUILDDIR)/$(LIBRLN_FILE)
+librln: | librln-experimental
 endif
 
 clean-librln:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -185,8 +185,12 @@ chat2: | build deps librln
 		$(ENV_SCRIPT) nim chat2 $(NIM_PARAMS) $(EXPERIMENTAL_PARAMS) waku.nims
 
 rln-keystore-generator: | build deps librln
-	echo -e $(BUILD_MSG) "build/$@" && \
+ifeq ($(RLN),true)
+		echo -e $(BUILD_MSG) "build/$@" && \
 		$(ENV_SCRIPT) nim rln_keystore_generator $(NIM_PARAMS) $(EXPERIMENTAL_PARAMS) waku.nims
+else
+		$(error RLN is not true. To build rln-keystore-generator, set RLN=true.)
+endif
```
rymnc, line 193: not sure if there's a better way to make RLN default to true while building this target - thoughts @vpavlin?

alrevuelta (approved):
instead of erroring out, cant we just set RLN=true if "make rln-keystore-generator"?

rymnc:
> instead of erroring out, cant we just set RLN=true if "make rln-keystore-generator"?

yeah, not sure how to

alrevuelta:
seems that by default its compiled with TRACE, so perhaps we should add this to the instructions.
```
make rln-keystore-generator RLN=true LOG_LEVEL=info
```

rymnc:
> seems that by default its compiled with TRACE, so perhaps we should add this to the instructions.
> 
> ```
> make rln-keystore-generator RLN=true LOG_LEVEL=info
> ```

Yeah im working on the readme for it

alrevuelta:
btw, it works fine but I see `execution error: metadata does not exist` in every run.

rymnc:
> btw, it works fine but I see `execution error: metadata does not exist` in every run.

because when the tree has just been created, the metadata does not exist when we try to fetch it.

Makefile
Ivansete-status: Hey @rymnc !
Maybe we could do sth like the next:
https://gist.github.com/Ivansete-status/fbd0c075658f942b2f9a5c111ae4ef1b#file-makefile-L138
https://gist.github.com/Ivansete-status/fbd0c075658f942b2f9a5c111ae4ef1b#file-makefile-L145
https://gist.github.com/Ivansete-status/fbd0c075658f942b2f9a5c111ae4ef1b#file-makefile-L190

rymnc: oh! good idea, i'll add that
thanks!

rymnc: Addressed in 3a7ebc43

Ivansete-status (approved):
LGTM! Thanks!

PR #1964, opened 2023-08-29
fix(Makefile): rln was enabled by default

rymnc:
# Description

With https://github.com/waku-org/nwaku/pull/1960, we introduced a change that made all compilations (even wakunode2) with rln enabled. This PR fixes that regression

# Changes

- [x] appropriate handling of librln target
- [x] fixed target for librln.a file

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -131,13 +131,12 @@ else
 LIBRLN_FILE := librln.a
 endif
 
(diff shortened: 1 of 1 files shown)


Ivansete-status (approved):
Awesome! Thanks for it!

PR #1965, opened 2023-08-29
chore: add debug log indicating succesful message pushes and also log the message hash

richard-ramos:
# Description
one suggestion that was made during today's message reliability call was to add this information to help debug cases where a message appears to have been pushed succesfully to all subscribers yet the message appears to not have been delivered.

To see this change in action, it would require the fleet to be deployed and also use `debug` level for logs

Notice that might add some noise on the logs for full filter nodes, due to changing the log level for one of the messages from TRACE to DEBUG, and also due to adding an extra log line, but it should be an acceptable tradeoff while dogfooding for filter is in progress. We can do a separate PR in the future to change the affected log lines to TRACE

### DIFF
--- a/waku/waku_filter_v2/protocol.nim
+++ b/waku/waku_filter_v2/protocol.nim
@@ -7,6 +7,7 @@ else:
 
 import
   std/[options,sequtils,sets,strutils,tables],
+  stew/byteutils,
   chronicles,
   chronos,
   libp2p/peerid,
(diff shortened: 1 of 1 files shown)


cammellos (approved):

jm-clius (approved):
Thanks! This will indeed lead to high log rates when there are many filter clients. Perhaps we can open an issue to revert log level to `trace` in future?

Ivansete-status (approved):

Ivansete-status:
> Thanks! This will indeed lead to high log rates when there are many filter clients. Perhaps we can open an issue to revert log level to `trace` in future?

I agree with that.
I think that in the future we could also strengthen the logic around metrics and keep the ability to debug any possible issues while keeping the log level back to `trace`.

richard-ramos:
> Perhaps we can open an issue to revert log level to trace in future?

Issue: https://github.com/waku-org/nwaku/issues/1974

PR #1966, opened 2023-08-30
feat(relay): ordered validator execution

### LINKED ISSUE #1967: chore(relay): ordered validator execution
(unordered by default in [nim-libp2p](https://github.com/status-im/nim-libp2p/blob/66f9dc9167f5ba8e5b6867be54f3e63052b93797/libp2p/protocols/pubsub/pubsub.nim#L520-L526))

## Acceptance criteria
- [ ] Remove unnecessary WakuMessage decodings
- [ ] Validators are executed in the order they are inserted in

rymnc:
# Description

> [!IMPORTANT]  
> This PR modifies waku_relay code and must be reviewed strictly :)
> Maybe we should run it in waku-simulator?

nim-libp2p does not order the execution of validators 
https://github.com/status-im/nim-libp2p/blob/66f9dc9167f5ba8e5b6867be54f3e63052b93797/libp2p/protocols/pubsub/pubsub.nim#L520-L526

but for our validators (signedTopicValidator, Rln), we need to guarantee order of execution. This PR sets the foundation 
for ordering the validators (they should all be appended sequentially)

# Changes

- [x] Removed bandwidth constraint validator from rln relay
- [x] left the config option in though, since it is still relevant (`--rln-relay-bandwidth-threshold`)
- [x] Modified WakuRelay.addValidator to accept validators that accept `WakuMessage`'s instead of nim-libp2p `Message`'s (thereby saving the extra decoding)
- [x] Updated SignedTopicValidator and RlnRelay to reflect the same
- [x] Tests updated

## Issue

closes #1967


waku/waku_rln_relay/rln_relay.nim
```diff
@@ -22,6 +22,7 @@ import
   ./protocol_types,
   ./protocol_metrics
 import
+  ../waku_relay, # for WakuRelayHandler
```
rymnc, line 25: not too sure if we should cause this dependency with Waku Relay, but it is tightly coupled for now

rymnc:
hmm looks like the tests fail where it shouldn't - valid rln messages are not hitting the relayHandlers - https://github.com/waku-org/nwaku/blob/0ac8a7f0cd46f345cbeeff10fb255200824e4ee0/tests/waku_rln_relay/test_wakunode_rln_relay.nim#L230-L234

waku/waku_relay/protocol.nim
```diff
@@ -167,9 +168,16 @@ proc new*(T: type WakuRelay, switch: Switch): WakuRelayResult[T] =
...
   return ok(w)
 
-method addValidator*(w: WakuRelay, topic: varargs[string], handler: ValidatorHandler) {.gcsafe.} =
-  procCall GossipSub(w).addValidator(topic, handler)
-
+proc addValidator*(w: WakuRelay, 
+                   topic: varargs[string], 
+                   handler: WakuValidatorHandler) {.gcsafe.} =
+  for t in topic:
+    if not w.wakuValidators.hasKey(t):
```
alrevuelta, line 175: perhaps we can simplify with
```
w.wakuValidators.mgetOrPut(topic, @[]).add(handler)
```

waku/waku_relay/protocol.nim
```diff
@@ -210,8 +209,25 @@ proc subscribe*(w: WakuRelay, pubsubTopic: PubsubTopic, handler: WakuRelayHandle
...
+    let msgRes = WakuMessage.decode(message.data)
+    if msgRes.isErr():
+      return ValidationResult.Reject
+    let msg = msgRes.get()
+
+    # now sequentially validate the message
+    if w.wakuValidators.hasKey(pubsubTopic):
+      let validatorsForTopic = w.wakuValidators[pubsubTopic]
+      for validator in validatorsForTopic:
+        return await validator(pubsubTopic, msg)
```
alrevuelta, line 226: not sure i understnad this return. if the first validator returns Accept, this `wrappedValidator` will return `Accept` and that it? no matter if the second or third validators reject the message? or am i missing something?

alrevuelta: i would expect something like 

```
for validator in validatorsForTopic:
  if (await validator(pubsubTopic, msg)) == ValidationResult.Reject:
    return ValidationResult.Reject
```

alrevuelta (changes requested):
left some comments, perhaps one is related to your problem?

waku/waku_relay/protocol.nim
rymnc: Addressed in 449e96fd ,thanks!

waku/waku_relay/protocol.nim
rymnc: Addressed in 449e96fd, thanks

waku/waku_relay/protocol.nim
```diff
@@ -210,8 +204,25 @@ proc subscribe*(w: WakuRelay, pubsubTopic: PubsubTopic, handler: WakuRelayHandle
...
+    let msgRes = WakuMessage.decode(message.data)
+    if msgRes.isErr():
+      return ValidationResult.Reject
+    let msg = msgRes.get()
+
+    # now sequentially validate the message
+    if w.wakuValidators.hasKey(pubsubTopic):
+      for validator in w.wakuValidators[pubsubTopic]:
+        if (await validator(pubsubTopic, msg)) == ValidationResult.Reject:
+          return ValidationResult.Reject
```
alrevuelta, line 221: almost forgot, guess we should handle `Ignore`?

https://github.com/status-im/nim-libp2p/blob/b76bac752ff0305730472a334bd2aaba1bd29210/libp2p/protocols/pubsub/errors.nim#L8

so if a validator returns `Ignore`, also return that result

alrevuelta:
```
let result = await validator(pubsubTopic, msg)
if result != ValidationResult.Accept:
   return result
```

or something like this. `Reject` or `Ignore` returns early. `Acept` forces to jump in the next iteration

waku/waku_relay/protocol.nim
```diff
@@ -122,7 +121,10 @@ const GossipsubParameters = GossipSubParams(
 type
   WakuRelayResult*[T] = Result[T, string]
   WakuRelayHandler* = proc(pubsubTopic: PubsubTopic, message: WakuMessage): Future[void] {.gcsafe, raises: [Defect].}
+  WakuValidatorHandler* = proc(pubsubTopic: PubsubTopic, message: WakuMessage): Future[ValidationResult] {.gcsafe, raises: [Defect].}
   WakuRelay* = ref object of GossipSub
+    wakuValidators*: Table[string, seq[WakuValidatorHandler]]
+    validatorInserted*: Table[string, bool]
```
Ivansete-status, line 127: These can be private.
Can we also have a short line describing them a little? Specially interesting the expected values for Keys, or maybe explicitly set `Pubsubtopic` as a type for the keys.
```suggestion
    wakuValidators: Table[string, seq[WakuValidatorHandler]]
    validatorInserted: Table[string, bool]
```

waku/waku_relay/protocol.nim
```diff
@@ -179,22 +183,33 @@ method stop*(w: WakuRelay) {.async.} =
...
-  return ValidationResult.Reject
-
 proc isSubscribed*(w: WakuRelay, topic: PubsubTopic): bool =
   GossipSub(w).topics.hasKey(topic)
 
 iterator subscribedTopics*(w: WakuRelay): lent PubsubTopic =
   for topic in GossipSub(w).topics.keys():
     yield topic
 
+proc generateOrderedValidator*(w: WakuRelay): auto {.gcsafe.} =
```
Ivansete-status, line 193: The 'Ordered' in this case refers to first validating the basic _Waku Message_ encoding, and then, the rest?

waku/waku_relay/protocol.nim
```diff
@@ -122,7 +121,10 @@ const GossipsubParameters = GossipSubParams(
 type
   WakuRelayResult*[T] = Result[T, string]
   WakuRelayHandler* = proc(pubsubTopic: PubsubTopic, message: WakuMessage): Future[void] {.gcsafe, raises: [Defect].}
+  WakuValidatorHandler* = proc(pubsubTopic: PubsubTopic, message: WakuMessage): Future[ValidationResult] {.gcsafe, raises: [Defect].}
```
Ivansete-status, line 124:
```suggestion
  WakuValidatorHandler = proc(pubsubTopic: PubsubTopic, message: WakuMessage): Future[ValidationResult] {.gcsafe, raises: [Defect].}
```

tests/waku_rln_relay/test_waku_rln_relay.nim
```diff
@@ -710,61 +710,6 @@ suite "Waku rln relay":
       msgValidate3 == MessageValidationResult.Valid
       msgValidate4 == MessageValidationResult.Invalid
 
-  asyncTest "should validate invalid proofs if bandwidth is available":
```
Ivansete-status, line 713: Ooh what a pitty. Why we are removing that? I am happy with it although not fully understanding the test logic ;P

Ivansete-status (approved):
LGTM! I've added a few comments/questions :)

Btw, I believe we should move the "validator" logic to a more appropriate place.
Given that the validation is part of the `Relay` protocol, I think we need to make the next refactoring:

- 1) Create a `validators` folder in "nwaku/waku/waku_relay/"
- 2) Move the logic of `apps/wakunode2/wakunode2_validator_signed.nim` into `nwaku/waku/waku_relay/validators/signed_topic_validator.nim`
- 3) Move the RLN validation logic to `nwaku/waku/waku_relay/validators/rln_validator.nim`

WDYT @jm-clius , @alrevuelta , @rymnc ?

alrevuelta:
> LGTM! I've added a few comments/questions :)
> 
> Btw, I believe we should move the "validator" logic to a more appropriate place. Given that the validation is part of the `Relay` protocol, I think we need to make the next refactoring:
> 
>     * 1. Create a `validators` folder in "nwaku/waku/waku_relay/"
> 
>     * 2. Move the logic of `apps/wakunode2/wakunode2_validator_signed.nim` into `nwaku/waku/waku_relay/validators/signed_topic_validator.nim`
> 
>     * 3. Move the RLN validation logic to `nwaku/waku/waku_relay/validators/rln_validator.nim`
> 
> 
> WDYT @jm-clius , @alrevuelta , @rymnc ?

Kind of agree but its a bit more complex than that. While rln is very coupled with relay, they are still kind of separate things. I would be happy to "merge" both since I don't conceive relay without rln, but not sure everyone agrees. Until we agree on this, 1. would be blocked. Same applies to 3.

2. This validator is a bit different, since its not part of waku protocol. Is some kind of special validator app level validator, thats why its in the app folder. If you ask me, I would completely remove this validator. But since some people may be using it, not sure until when, guess we can't.

Ivansete-status:
> > LGTM! I've added a few comments/questions :)
> > Btw, I believe we should move the "validator" logic to a more appropriate place. Given that the validation is part of the `Relay` protocol, I think we need to make the next refactoring:
> > ```
> > * 1. Create a `validators` folder in "nwaku/waku/waku_relay/"
> > 
> > * 2. Move the logic of `apps/wakunode2/wakunode2_validator_signed.nim` into `nwaku/waku/waku_relay/validators/signed_topic_validator.nim`
> > 
> > * 3. Move the RLN validation logic to `nwaku/waku/waku_relay/validators/rln_validator.nim`
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > WDYT @jm-clius , @alrevuelta , @rymnc ?
> 
> Kind of agree but its a bit more complex than that. While rln is very coupled with relay, they are still kind of separate things. I would be happy to "merge" both since I don't conceive relay without rln, but not sure everyone agrees. Until we agree on this, 1. would be blocked. Same applies to 3.
> 
> 2. This validator is a bit different, since its not part of waku protocol. Is some kind of special validator app level validator, thats why its in the app folder. If you ask me, I would completely remove this validator. But since some people may be using it, not sure until when, guess we can't.

Thanks for the answers @alrevuelta !
I personally vote for having both validators added to _Relay_. For sure, the _RLN_ one. And re the "signed-pubsubtopics" validator I'd add it to _Relay_ and, if we decide in the future not to have it, then we remove it. I believe that the `libwaku` should have also the possibility to benefit from those validators.

alrevuelta:
trying to upstream this https://github.com/status-im/nim-libp2p/pull/945/ so perhaps we can simplify this PR. Our case would be the `Sequential` one.

jm-clius:
Even without the upstream change, I quite like explicitly decoupling "Waku Validators" from the internal relay validators. This is exactly because validators rely so much on application-specific configuration, decisions and ordering.
This PR is probably fine for now, but in future I would suggest something like:
- move all validators to the node level (can be in a subfolder) and convert them to `WakuValidator` (as has been done here)
- let the _app_ (`wakunode2`) do some `setupValidators` logic that add validators according to the external configuration, etc. and combine them into a single validator
- this single validator are then added as relay validator on `WakuNode` creation

> Given that the validation is part of the Relay protocol

Don't quite agree. The validation internals is not part of relay protocol. It exists exactly because the application (in this case the Waku Node) has knowledge about message validity that is not available (and should not be available) to the relay protocol. In other words, validators are a mechanism for the application to pass down validation knowledge (which belongs to them) to the relay protocol. See for example [this discussion](https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.0.md#message-processing) on message processing in the gossipsub spec.

rymnc:
> Even without the upstream change, I quite like explicitly decoupling "Waku Validators" from the internal relay validators. This is exactly because validators rely so much on application-specific configuration, decisions and ordering.
> This PR is probably fine for now, but in future I would suggest something like:
> - move all validators to the node level (can be in a subfolder) and convert them to `WakuValidator` (as has been done here)
> - let the _app_ (`wakunode2`) do some `setupValidators` logic that add validators according to the external configuration, etc. and combine them into a single validator
> - this single validator are then added as relay validator on `WakuNode` creation
> 
> > Given that the validation is part of the Relay protocol
> 
> Don't quite agree. The validation internals is not part of relay protocol. It exists exactly because the application (in this case the Waku Node) has knowledge about message validity that is not available (and should not be available) to the relay protocol. In other words, validators are a mechanism for the application to pass down validation knowledge (which belongs to them) to the relay protocol. See for example [this discussion](https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.0.md#message-processing) on message processing in the gossipsub spec:
> 
> > Payload processing will validate the message according to application-defined rules...

I like this idea - this way the application (wakunode2) defines the validators for waku relay, in the order it wants. reduces dependency issues too.

so I guess we have to define the "minimum validation" for waku_relay, which could be the waku message decoding.

jm-clius:
> so I guess we have to define the "minimum validation" for waku_relay, which could be the waku message decoding.

Indeed, but even that could in theory be on the "node" layer - relay shouldn't necessarily have to know it's passing around `WakuMessage`. From it's perspective it's a routing protocol for "payload" with some security properties. It's the waku node layer that requires this payload to be valid decodable waku messages. But, of course, it's not always possible to refactor architecture in one leap, so this is just an idea.

alrevuelta:
> trying to upstream this https://github.com/status-im/nim-libp2p/pull/945 so perhaps we can simplify this PR. Our case would be the Sequential one.

Lets try to wrap this up this PR. Not sure if we can upstream the feature of sequential validators [see](https://github.com/status-im/nim-libp2p/pull/945#issuecomment-1702530329) so don't want to block this PR :)

tests/waku_rln_relay/test_waku_rln_relay.nim
rymnc: this pr also removes the bandwidth constraint validator, which should be a separate validator!

waku/waku_relay/protocol.nim
rymnc: Sort of, we describe here 1 single validator that gossipsub calls back to - and we control the order of

waku/node/waku_node.nim
```diff
@@ -749,7 +749,7 @@ when defined(rln):
     # register rln validator for all subscribed relay pubsub topics
     for pubsubTopic in node.wakuRelay.subscribedTopics:
       debug "Registering RLN validator for topic", pubsubTopic=pubsubTopic
-      procCall GossipSub(node.wakuRelay).addValidator(pubsubTopic, validator)
+      node.wakuRelay.addValidator(pubsubTopic, validator)
```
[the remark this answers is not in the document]
rymnc: yes, this is. as @jm-clius described below, we will create a new `waku_validators` directory where we initialize them in the order we wish for them to execute

waku/waku_relay/protocol.nim
rymnc: I believe this should be public so other protocols may use this to better describe the handlers

alrevuelta (commented):
minor comment, looking almost ready.

@rymnc As far as I understand we must ensure that `addValidator` is called always before `subscribe.` Otherwise the wrapped validator created with `generateOrderedValidator` won't contain the ones added later.

This behaviour is ok, but can we ensure that `addValidator` is not called after `subscribe()` is called? So that we can avoid people adding a validator that is not really applied.

waku/waku_relay/protocol.nim
```diff
@@ -210,8 +225,10 @@ proc subscribe*(w: WakuRelay, pubsubTopic: PubsubTopic, handler: WakuRelayHandle
     else:
       return handler(pubsubTopic, decMsg.get())
 
-  # add the default validator to the topic
-  procCall GossipSub(w).addValidator(pubSubTopic, validator)
+  # add the ordered validator to the topic
+  if not w.validatorInserted.hasKey(pubSubTopic):
+    procCall GossipSub(w).addValidator(pubSubTopic, w.generateOrderedValidator())
+    w.validatorInserted[pubSubTopic] = true
```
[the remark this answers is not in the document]
rymnc: Thanks, addressed in 893ddf5c

rymnc:
> minor comment, looking almost ready.
> 
> @rymnc As far as I understand we must ensure that `addValidator` is called always before `subscribe.` Otherwise the wrapped validator created with `generateOrderedValidator` won't contain the ones added later.
> 
> This behaviour is ok, but can we ensure that `addValidator` is not called after `subscribe()` is called? So that we can avoid people adding a validator that is not really applied.

not sure I understand "ensure that `addValidator` is called always before `subscribe.`"
https://github.com/waku-org/nwaku/blob/893ddf5c271ddf42a46635bb750547016e5e1274/waku/waku_relay/protocol.nim#L217-L241

it is called before subscribe 🤔

alrevuelta:
> not sure I understand "ensure that addValidator is called always before subscribe."

@rymnc  mm the `addValidator` you are referring to is the gossipsub one. I'm refering to wakuRelay `addValidator`. Unless I'm missing something, what would happen if:

```
node.subscribe(topic)
node.addValidator(validator)
```

That validator wont be executed right? So its added but it never "reaches" gossipsub

alrevuelta (approved):
lgtm!

PR #1968, opened 2023-08-30
chore(api): validate rln message before sending (rest + rpc)

alrevuelta:
Description:
* Adds validation to rest+rpc api so that an rln message proof is verified before sending the message. If the proof is wrong or the message is considered spam, the message won't be sent and it will error out so that the user is aware that the message didn't made it.
* Adds validation to rest+rpc api so that it errors out when trying to publish to a topic that the node is not subcribed, since it risks that there are no connected peers "listening" in that topic.
* Update tests so that they work with rln enabled/disabled.

How to test it:

```
./build/wakunode2 \
--rln-relay=true \
--rpc-address=0.0.0.0 \
--rpc-admin=true \
--rest=true \
--rest-admin=true \
--rest-private=true \
--rest-address=0.0.0.0 \
--rln-relay-dynamic=true \
--rln-relay-cred-path=rlnKeystore_420.json \
--rln-relay-membership-index=420 \
--rln-relay-tree-path=rln_tree_1.db \
--rln-relay-eth-contract-address=0x8e1F3742B987d8BA376c0CBbD7357fE1F003ED71  \
--rln-relay-eth-client-address=wss://you_endpoint \
--log-level=debug
```

Rest API
```
curl -X 'POST'   'localhost:8645/relay/v1/messages/%2Fwaku%2F2%2Fdefault-waku%2Fproto'   -H 'accept: */*'   -H 'Content-Type: application/json'   -d '{
  "payload": "string",
  "contentTopic": "string",
  "version": 0,
  "timestamp": 0
}'
```

RPC API

```
curl http://localhost:8545 -d '{"jsonrpc":"2.0","method":"post_waku_v2_relay_v1_message","params":["/waku/2/default-waku/proto", {"payload": "SGkgdGhlcmUgZXRoY2M=", "timestamp": 1, "contentTopic": "some-topic", "ephemeral": false}],"id":1}' -H 'Content-Type: application/json'
```

### DIFF
--- a/waku/node/jsonrpc/relay/handlers.nim
+++ b/waku/node/jsonrpc/relay/handlers.nim
@@ -100,6 +100,7 @@ proc installRelayApiHandlers*(node: WakuNode, server: RpcServer, cache: MessageC
     if pubsubTopic notin node.wakuRelay.subscribedTopics():
       raise newException(ValueError, "Failed to publish: Node not subscribed to pubsubTopic: " & pubsubTopic)
 
+    # if RLN is mounted, append the proof to the message
     when defined(rln):
       if not node.wakuRlnRelay.isNil():
         # append the proof to the message
@@ -107,7 +108,6 @@ proc installRelayApiHandlers*(node: WakuNode, server: RpcServer, cache: MessageC
                                                       float64(getTime().toUnix()))
         if not success:
           raise newException(ValueError, "Failed to publish: error appending RLN proof to message")
-
         # validate the message before sending it
         let result = node.wakuRlnRelay.validateMessage(message)
         if result == MessageValidationResult.Invalid:
@@ -115,13 +115,22 @@ proc installRelayApiHandlers*(node: WakuNode, server: RpcServer, cache: MessageC
         elif result == MessageValidationResult.Spam:
           raise newException(ValueError, "Failed to publish: limit exceeded, try again later")
         elif result == MessageValidationResult.Valid:
+          debug "Publishing message WITH RLN proof", pubSubTopic=pubSubTopic
           let publishFut = node.publish(pubsubTopic, message)
-
           if not await publishFut.withTimeout(futTimeout):
             raise newException(ValueError, "Failed to publish: timed out")
         else:
           raise newException(ValueError, "Failed to publish: unknown RLN proof validation result")
+      else:
+        raise newException(ValueError, "Failed to publish: RLN enabled but not mounted")
 
+    # if RLN is not mounted, publish the message as is
+    else:
+      debug "Publishing message WITHOUT RLN proof", pubSubTopic=pubSubTopic
+      let publishFut = node.publish(pubsubTopic, message)
+
+      if not await publishFut.withTimeout(futTimeout):
+        raise newException(ValueError, "Failed to publish: timed out")
 
     return true
 
--- a/waku/node/jsonrpc/relay/handlers.nim
+++ b/waku/node/jsonrpc/relay/handlers.nim
@@ -115,22 +115,17 @@ proc installRelayApiHandlers*(node: WakuNode, server: RpcServer, cache: MessageC
         elif result == MessageValidationResult.Spam:
           raise newException(ValueError, "Failed to publish: limit exceeded, try again later")
         elif result == MessageValidationResult.Valid:
-          debug "Publishing message WITH RLN proof", pubSubTopic=pubSubTopic
-          let publishFut = node.publish(pubsubTopic, message)
-          if not await publishFut.withTimeout(futTimeout):
-            raise newException(ValueError, "Failed to publish: timed out")
+          debug "RLN proof validated successfully", pubSubTopic=pubSubTopic
         else:
           raise newException(ValueError, "Failed to publish: unknown RLN proof validation result")
       else:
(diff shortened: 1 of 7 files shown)


waku/node/jsonrpc/relay/handlers.nim
```diff
@@ -93,18 +94,34 @@ proc installRelayApiHandlers*(node: WakuNode, server: RpcServer, cache: MessageC
...
+        let success = node.wakuRlnRelay.appendRLNProof(message,
                                                       float64(getTime().toUnix()))
         if not success:
-          raise newException(ValueError, "Failed to append RLN proof to message")
-
-    let publishFut = node.publish(topic, message)
+          raise newException(ValueError, "Failed to publish: error appending RLN proof to message")
+
+        # validate the message before sending it
+        let result = node.wakuRlnRelay.validateMessage(message)
```
alrevuelta, line 112: unsure if its ok to use `validateMessage` here, since inside it updates a bunch of metrics and call `updateLog`. Any opinion @rymnc

rymnc: yeah I don't think we should use validateMessage, because if you try publishing the same message it would get stuck since the proofMetadata would exist in the nullifierLog afaik. maybe try using hasDuplicate?

alrevuelta: Something like this 74d38f530c253c7ab8f9745de7649b42de307d06 ?

rymnc: yes!

vpavlin (approved):
lgtm

waku/waku_rln_relay/rln_relay.nim
```diff
@@ -241,19 +240,39 @@ proc validateMessage*(rlnPeer: WakuRLNRelay,
...
-  # the result of `updateLog` is discarded because message insertion is guaranteed by the implementation i.e.,
-  # it will never error out
-  discard rlnPeer.updateLog(proofMetadataRes.get())
-  debug "message is valid", payload = string.fromBytes(msg.payload)
+  debug "message is valid", payloadLen = msg.payload.len
   let rootIndex = rlnPeer.groupManager.indexOfRoot(proof.merkleRoot)
   waku_rln_valid_messages_total.observe(rootIndex.toFloat())
   return MessageValidationResult.Valid
 
+proc validateMessageAndUpdateLog*(
```
alrevuelta, line 252: @rymnc created this function, which is like the old `validateMessage` but now we have:
* validateMessage
* validateMessageAndUpdateLog

waku/waku_relay/protocol.nim
```diff
@@ -191,9 +191,8 @@ proc validator(pubsubTopic: string, message: messages.Message): Future[Validatio
 proc isSubscribed*(w: WakuRelay, topic: PubsubTopic): bool =
   GossipSub(w).topics.hasKey(topic)
 
-iterator subscribedTopics*(w: WakuRelay): lent PubsubTopic =
-  for topic in GossipSub(w).topics.keys():
-    yield topic
+proc subscribedTopics*(w: WakuRelay): seq[PubsubTopic] =
+  return toSeq(GossipSub(w).topics.keys())
```
Ivansete-status, line 195: Morning! Is this directly related to the purpose of the PR?

Ivansete-status (approved):
LGTM! Thanks for it! I just added a question to learn sth :)

waku/waku_relay/protocol.nim
alrevuelta: nope, but imho there is no need to use generators for this. the amount of topics is not enough to justify it. also using poorly documented features such as lent `lent` is a bad idea, unless strictly needed.

waku/node/jsonrpc/relay/handlers.nim
```diff
@@ -93,18 +94,43 @@ proc installRelayApiHandlers*(node: WakuNode, server: RpcServer, cache: MessageC
...
+      else:
+        raise newException(ValueError, "Failed to publish: RLN enabled but not mounted")
+
+    # if RLN is not mounted, publish the message as is
+    else:
+      debug "Publishing message WITHOUT RLN proof", pubSubTopic=pubSubTopic
+      let publishFut = node.publish(pubsubTopic, message)
+
+      if not await publishFut.withTimeout(futTimeout):
+        raise newException(ValueError, "Failed to publish: timed out")
```
rymnc, line 133: I think these `ValueError`'s can be `CatchableError`'s

rymnc (approved):
LGTM, minor comment

waku/node/jsonrpc/relay/handlers.nim
rymnc: also, maybe we can operate on the same `message` in both cases? i.e rln is enabled, rln is disabled?
then we can save some loc by having just one 
```nim
let publishFut = node.publish(pubsubTopic, message)
if not await publishFut.withTimeout(futTimeout):
    raise newException(ValueError, "Failed to publish: timed out")
```

alrevuelta: ow indeed, that was a bit spageti.
c964916c2f5260bd79c96ab6a247446fea1a92ba

PR #1972, opened 2023-08-30
fix(tests): fix flaky test

alrevuelta:
* Fixes flaky test `testing rln-relay is applied in all rln pubsub/content topics`.
* This is caused most likely because generating rln proof in github actions take a while. Guess the resources are quite low.


tests/waku_rln_relay/test_wakunode_rln_relay.nim
```diff
@@ -150,17 +150,26 @@ procSuite "WakuNode - RLN relay":
...
+    var messages1: seq[WakuMessage] = @[]
+    var messages2: seq[WakuMessage] = @[]
+
     for i in 0..<3:
-      var message1 = WakuMessage(payload: ("Payload_" & $i).toBytes(), contentTopic: contentTopics[0])
-      doAssert(nodes[0].wakuRlnRelay.appendRLNProof(message1, epochTime()))
-      await nodes[0].publish(pubsubTopics[0], message1)
+      var message = WakuMessage(payload: ("Payload_" & $i).toBytes(), contentTopic: contentTopics[0])
+      doAssert(nodes[0].wakuRlnRelay.appendRLNProof(message, epochTime()))
+      messages1.add(message)
```
vpavlin, line 162: I am not sure what the difference is - does `appendRLNProof` somehow "cache" the proof when called in the same epoch? Because it seems like the only difference is in when does the "await" happen - here vs. a bit later, but the proof generation does not really change. Or what am I missing?

alrevuelta: mm as it was before:
* await generate proof
* await publish message
* await generate proof
* await publish message

the problem is that generate proof takes a bit of time. If this time grows > Epoch, then the second published message wont be considered spam, since enough time passed so you are not rate limited. This was the reason the test were failing. I was expecting only 1 message to arrive, but 2 arrives, since the second was sent too late, hence skipping the rate limit.

vpavlin: So would it make sense to wrap this in an async proc and execute the proof generation in parallel and then just await all futures? I feel like it's highly probable that if the proof generation takes a lot of time, we could still hit the "too slow" issue this way (even though with lower probability)

rymnc (approved):
LGTM

tests/waku_rln_relay/test_wakunode_rln_relay.nim
alrevuelta: you are very right, but found an easier solution. "paralellizing" the proof generation would imply more chanes, since its not async, buttt just realized we can use the same epoch time for all of them. so with this it shouldn't matter that proof generation takes 1 hour, since all messages are using the same epoch.

b8b58e22a1320959b1401c1a5577d28093b1de8f

vpavlin: Ha! Perfect!

vpavlin (approved):

PR #1975, opened 2023-08-31
chore(Makefile): Small addon to be able to list all possible target in our Makefile

NagyZoltanPeter:
Just a small addon, was useful for me initially (idea coming from how cmake generates Makefiles).
I hope it is still useful for others.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -300,3 +300,11 @@ release-notes:
 			sed -E 's@#([0-9]+)@[#\1](https://github.com/waku-org/nwaku/issues/\1)@g'
 # I could not get the tool to replace issue ids with links, so using sed for now,
 # asked here: https://github.com/bvieira/sv4git/discussions/101
+
+#############################
+# List all possible Targets #
+#############################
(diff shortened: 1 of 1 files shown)


vpavlin:
What is the difference between this and the built-in shell completion? It seems to output the same list of targets:)

NagyZoltanPeter:
> What is the difference between this and the built-in shell completion? It seems to output the same list of targets:)

The difference is in my lack of knowledge having such feature :-D
So in this case we may can drop this one.

Ivansete-status (approved):
LGTM! Thanks for it!

jm-clius:
If I understand correctly we should close without merging as the feature already exists? @NagyZoltanPeter please go ahead and do so, if that is the case.

Ivansete-status:
Yes, we can close it. I didn't know that feature too xD
Very useful

NagyZoltanPeter:
Agree, lessons learned. Therefore close this.

PR #1976, opened 2023-08-31
chore(ci): enable experimental for the PR image builds

vpavlin:
# Description

Spliting out the container image build into a separate workflow and invoking it from both `ci` and `ci-experimental`. This results in 2 images being pushed - `${PR_NUMBER}` and `${PR_NUMBER}-experimental` - thought it might be useful with regards to the push on RLN lately

# Changes

- [ ] split out `container-image.yml` workflow
- [ ] execute it from both `ci` and `ci-experimental` workflows


.github/workflows/container-image.yml
```diff
@@ -0,0 +1,71 @@
...
+            vendor/
+            .git/modules
+          key: ${{ runner.os }}-vendor-modules-${{ steps.submodules.outputs.hash }}
+
+      - name: Build binaries
+        id: build
+        run: |
+          EXPERIMENTAL=${{ inputs.experimental }}
+
+          make -j${NPROC} V=1 QUICK_AND_DIRTY_COMPILER=1 NIMFLAGS="-d:disableMarchNative" wakunode2 EXPERIMENTAL=${{EXPERIMENTAL}}
```
rymnc, line 45: curious if we can have this workflow depend on the [build job](https://github.com/waku-org/nwaku/blob/b785b6ba1588c808e0c5d78c2c8aa4c7cc5794b7/.github/workflows/ci.yml#L57-L76) in the `ci` workflow

we can probably skip the docker image building if the `ci` workflow fails, right?

vpavlin: We could, my thinking initially was to provide the image ASAP rather than wait for the CI to finish, but we definitely can change that

rymnc (approved):
LGTM

.github/workflows/ci-experimental.yml
```diff
@@ -116,3 +116,11 @@ jobs:
           fi
 
           make V=1 LOG_LEVEL=DEBUG QUICK_AND_DIRTY_COMPILER=1 test testwakunode2
+
+  build-docker-image:
+    needs: changes
+    if: ${{ needs.changes.outputs.v2 == 'true' || needs.changes.outputs.common == 'true' || needs.changes.outputs.docker == 'true' }}
```
Ivansete-status, line 122: Is is needed to add the next snippet in the "filter change detection" above?
```
          docker:
          - 'docker/**'
```

Ivansete-status (approved):
LGTM! Thanks for it!

.github/workflows/ci-experimental.yml
vpavlin: Ah, yes, I added that for the ci, but not for the ci-experimental, good catch

PR #1977, opened 2023-08-31
fix: deprecated --topic config option should be ignored when using newer options

SionoiS:
The deprecated `--topic` config has a default.

When using the newer options `--content-topic` & `--pubsub-topic` it should be ignored.

Also added `getShard` that take and return `string`


jm-clius (approved):
LGTM

PR #1978, opened 2023-08-31
refactor(cbindings): Thread-safe communication between the main thread and the Waku Thread

Ivansete-status:
# Description
This PR is part of a PR suite aimed to follow Jacek's recommendations re thread-safe communication. 

Concretely, 
- Start using `ChannelSPSCSingle` so that the main thread and the _Waku Thread_ communicate safely.
- The two threads communicate by sending each other a `ptr` whose memory is being allocated in the thread-shared space.
- One thread allocates in transmission, and the other thread deallocates in reception.
- A request-response communication takes place.

This change is motivated by the next comment:
https://github.com/waku-org/nwaku/pull/1865#discussion_r1282738778

# Changes

- [x] Stop using `Channel` type in favour of `ChannelSPSCSingle` ( https://github.com/status-im/nim-taskpools/tree/15e23ef1cf0860330dcc32f50fcce5f840031e28 )
- [x] Adding `nim-taskpools`  as a new submodule/vendor ( https://github.com/status-im/nim-taskpools/tree/15e23ef1cf0860330dcc32f50fcce5f840031e28 )

## Issue

Closes https://github.com/waku-org/nwaku/issues/1878


SionoiS (approved):

library/waku_thread/inter_thread_communication/requests/node_lifecycle_request.nim
```diff
@@ -108,8 +111,10 @@ proc createNode(configJson: cstring):
 
   return ok(newNode)
 
-method process*(self: NodeLifecycleRequest,
-                node: ptr WakuNode): Future[Result[string, string]] {.async.} =
+proc process*(self: ptr NodeLifecycleRequest,
+              node: ptr WakuNode): Future[Result[string, string]] {.async.} =
+
+  defer: deallocShared(self)
```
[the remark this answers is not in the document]
Ivansete-status: Morning and good question!
I understand the `defer` statement as a way to make sure something will get called when the context goes out of the scope the `defer` is in. So, in this case, I'm assuming that the `deallocShared(self)` will get called in any case the `process` ends.

library/waku_thread/inter_thread_communication/waku_thread_request.nim
```diff
@@ -0,0 +1,55 @@
...
+              node: ptr WakuNode):
+              Result[string, string] =
+  ## Processes the request and deallocates its memory
+  defer: deallocShared(request)
+
+  echo "Request received: " & $request[].reqType
+
+  case request[].reqType
+    of LIFECYCLE:
+      waitFor cast[ptr NodeLifecycleRequest](request[].reqContent).process(node)
```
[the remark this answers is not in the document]
Ivansete-status: Well, in this case, the `waitFor` is "sync", and it keeps there making the dispatcher progress until the `process` is done. The `reqContent` pointer will only be deallocated by the `process` proc when it completes.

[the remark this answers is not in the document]
Ivansete-status: Well, afaik, the `waitFor` is "sync", and it keeps there making the dispatcher progress until the `process` is done. The `reqContent` pointer will only be deallocated by the `process` proc when it completes.

library/waku_thread/inter_thread_communication/requests/node_lifecycle_request.nim
```diff
@@ -29,18 +28,22 @@ type
...
     configJson: cstring ## Only used in 'CREATE_NODE' operation
 
 proc new*(T: type NodeLifecycleRequest,
           op: NodeLifecycleMsgType,
-          configJson: cstring = ""): T =
+          configJson: cstring = ""): ptr NodeLifecycleRequest =
 
-  return NodeLifecycleRequest(operation: op, configJson: configJson)
+  var ret = cast[ptr NodeLifecycleRequest](
+              allocShared0(sizeof(NodeLifecycleRequest)))
```
arnetheduck, line 40:
```suggestion
  var ret = createShared(NodeLifecycleRequest)
```

library/waku_thread/inter_thread_communication/requests/node_lifecycle_request.nim
```diff
@@ -29,18 +28,22 @@ type
...
 type
-  NodeLifecycleRequest* = ref object of InterThreadRequest
+  NodeLifecycleRequest* = object
     operation: NodeLifecycleMsgType
     configJson: cstring ## Only used in 'CREATE_NODE' operation
 
 proc new*(T: type NodeLifecycleRequest,
           op: NodeLifecycleMsgType,
-          configJson: cstring = ""): T =
+          configJson: cstring = ""): ptr NodeLifecycleRequest =
```
arnetheduck, line 37:
```
proc create*(T: type Nodeli...): ptr T = ...
```

library/waku_thread/inter_thread_communication/requests/node_lifecycle_request.nim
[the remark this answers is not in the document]
arnetheduck: indeed, `defer` is syntactic sugar for `try/finally`

library/waku_thread/inter_thread_communication/requests/peer_manager_request.nim
```diff
@@ -7,27 +7,29 @@ import
...
-          dialTimeout: Duration): T =
+          dialTimeout: Duration): ptr PeerManagementRequest =
 
-  return PeerManagementRequest(operation: op,
-                               peerMultiAddr: peerMultiAddr,
-                               dialTimeout: dialTimeout)
+  var ret = cast[ptr PeerManagementRequest](
+                      allocShared0(sizeof(PeerManagementRequest)))
+  ret[].operation = op
+  ret[].peerMultiAddr = peerMultiAddr
```
arnetheduck, line 30: `string` is a garbage-collected type - here, a copy must be taken with the shared allocator (and later it must be deallocated) - no garbage collected types allowed in objects constructed with `create`: https://status-im.github.io/nim-style-guide/interop.html#garbage-collected-types

arnetheduck: (type needs changing to `cstring`)

library/waku_thread/inter_thread_communication/requests/peer_manager_request.nim
```diff
@@ -46,13 +48,15 @@ proc connectTo(node: WakuNode,
 
   return ok()
 
-method process*(self: PeerManagementRequest,
-                node: ptr WakuNode): Future[Result[string, string]] {.async.} =
+proc process*(self: ptr PeerManagementRequest,
+              node: ptr WakuNode): Future[Result[string, string]] {.async.} =
```
arnetheduck, line 52: this can be plain `WakuNode` because the function does not write to `node`

library/waku_thread/inter_thread_communication/requests/node_lifecycle_request.nim
arnetheduck: actually, `createShared` because it's taken from the shared allocator

library/waku_thread/inter_thread_communication/waku_thread_request.nim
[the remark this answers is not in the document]
arnetheduck: `waitFor` turns async code into sync code indeed - but a problem with this setup is that the async loop will only be running while `waitFor` is running - this means that it relies on a stead flow of requests in order to process networking buffers and other waku activity (timers, etc).

To solve this, `process` itself must be `async` and this should be turned into `await process` - the "main" loop then needs to call `poll` / `runForever`

library/waku_thread/inter_thread_communication/requests/node_lifecycle_request.nim
Ivansete-status: I changed that, thanks! ( as a return type I set `ptr type T` )

library/waku_thread/inter_thread_communication/requests/peer_manager_request.nim
Ivansete-status: Fixed. Thanks!

library/waku_thread/inter_thread_communication/requests/peer_manager_request.nim
Ivansete-status: Okay, this is the trickiest part. 

I was assuming that a thread-safe communication was ensured by just sending `ptr` types, given that the `ptr` type is not tracked by the GC by default. I also wanted to avoid the overhead of serializing/parsing JSON objects.

My assumption was that the next was secure:
1. `Thread A` creates a `ptr` of the request in the shared space.
2. `Thread A` sends the address of the request object to `Thread B`.
3. `Thread B` handles the request and deallocates the memory from the shared space.

The [next](https://github.com/waku-org/nwaku/blob/78ebb3af77bdeca1d62761f79ecb2c3b26c08728/library/waku_thread/waku_thread.nim#L23C58-L23C58) is used to communicate both threads:
```
    reqChannel: ChannelSPSCSingle[ptr InterThreadRequest]
    respChannel: ChannelSPSCSingle[ptr InterThreadResponse]
```

The `RelayRequest` type is the most complex example:
```
type
  RelayRequest* = object
    operation: RelayMsgType
    pubsubTopic: PubsubTopic
    relayEventCallback: WakuRelayHandler
    message: WakuMessage
```
... which is created by `Thread A` in https://github.com/waku-org/nwaku/blob/78ebb3af77bdeca1d62761f79ecb2c3b26c08728/library/waku_thread/inter_thread_communication/requests/protocols/relay_request.nim#L34

... and deallocated by `Thread B` in: https://github.com/waku-org/nwaku/blob/78ebb3af77bdeca1d62761f79ecb2c3b26c08728/library/waku_thread/inter_thread_communication/requests/protocols/relay_request.nim#L44

Isn't that a thread-safe approach to sending `ptr` types over?

arnetheduck: `peerMultiAddr` is a `string` which is a garbage-collected type (together with seq, ref and closures). Both the "main object" and all its fields need to be thread safe, ie non-req.

This means that we manually allocate a copy of everything on `createShared` and release it - I suggest implementing a `destroyShared` function for every `createShared` which deallocates all fields and finally the main object - in the future, these can be turned into proper destructors but they are good for now.

library/waku_thread/inter_thread_communication/waku_thread_request.nim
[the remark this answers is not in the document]
arnetheduck: right - this runs the risk of getting stuck in `poll` because `poll` itself will block until there is activity, and if there is no activity it will simply block forever - in fact, I'm not sure how this works at all - it should get stuck there and never perform any loop iterations - I'm probably missing some other detail which wakes up poll but this looks like a significant risk with the setup.

There are two ways to solve this: introducing a timer / `sleepAsync` (`waitFor sleepAsync(1.millis)` instead of `poll`) or using a [`ThreadSignal`]. Introducing a timer is the easier way to make this code correct.

Using a signal has the advantage of using fewer resources, but is slightly more difficult to implement. I recommend leaving it for a separate PR - the way to use that is to create a signal for every channel (this is the way channels are normally implemented, with a "notification mechanism"), and every time an item is added to the channel, the signal is fired - here's a simple example: https://gist.github.com/arnetheduck/b6a7ac8f4b85490d26d464674e09d57d#file-threadsynctask-nim

library/waku_thread/waku_thread.nim
```diff
@@ -83,20 +89,24 @@ proc createWakuThread*(): Result[void, string] =
...
+  ## Sending the request
+  let sentOk = ctx.reqChannel.trySend(req)
+  if not sentOk:
+    return err("Couldn't send a request to the waku thread: " & $req[])
 
-  return resp[1]
+  ## Waiting for the response
+  var response: ptr InterThreadResponse
+  var recvOk = ctx.respChannel.tryRecv(response)
+  while recvOk == false:
```
arnetheduck, line 107: instead of this loop, a signal can be used here too

library/waku_thread/inter_thread_communication/waku_thread_request.nim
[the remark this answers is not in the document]
arnetheduck: "in general" I would look for ways / strive to avoid calling `waitFor` in "inner" proc's and rather structure the code in such a way that the `waitFor` happens only at the "outer" layer - the waitFor strategy is not incorrect, but it's unusual and runs the risk of accidentally ending up with nested `waitFor` calls (calling `waitFor` from an `async` function) which is not supported.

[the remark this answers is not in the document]
Ivansete-status: > "in general" I would look for ways / strive to avoid calling `waitFor` in "inner" proc's...

Okay perfect. I enhanced that in my last commit.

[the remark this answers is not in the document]
Ivansete-status: > right - this runs the risk of getting stuck in `poll` because `poll` itself will block...

Ok thanks! We'll apply `waitFor sleepAsync(1)` for now and will implement the enhancement with signal in further PRs.

library/waku_thread/waku_thread.nim
Ivansete-status: Okay!, we'll apply the `signal` enhancements in a future PR.

library/waku_thread/waku_thread.nim
```diff
@@ -83,20 +90,24 @@ proc createWakuThread*(): Result[void, string] =
...
 
-  ctx.reqChannel.send(req)
+  let req = InterThreadRequest.createShared(reqType, reqContent)
 
-  var resp = ctx.respChannel.tryRecv()
-  while resp[0] == false:
-    resp = ctx.respChannel.tryRecv()
-    os.sleep(1)
+  ## Sending the request
+  let sentOk = ctx.reqChannel.trySend(req)
```
arnetheduck, line 101: can it happen that requests are answered out-of-order because of `async`?

Ivansete-status: > can it happen that requests are answered out-of-order because of `async`?

Good point! I believe this couldn't happen as requests are attended sequentially, thanks to the `waitFor` in https://github.com/waku-org/nwaku/blob/01793d2156bf39162995459a59777a36dc4e0b29/library/waku_thread/waku_thread.nim#L58

On the other hand, once we will apply the "signal" approach we'll have a better synchronization between both threads.

arnetheduck:
LGTM! looking forward to the next round

arnetheduck (approved):

PR #1980, opened 2023-09-01
chore(rln-relay): confirm that the provided credential is correct using onchain query

rymnc:
# Description

Updates tests to use zero-based indexing in the contract deployed at `0x59af93Bb30ED234418E6fC2324597e3FD72FB90e` on sepolia
Also checks that the decrypted credential's commitment does in fact belong to the tree by calling `memberExists()` on the rln storage contract

# Changes

- [x] Adds comments about contract fns
- [x] checks that the decrypted credential's commitment belongs to the tree
- [x] updated tests


alrevuelta (commented):
ah just realized that `memberExists` should actually be `memberExistsOrExisted` (forget the naming). I mean, since the contract is dummy and just emit events, it can't really know if the member exists. It just can know if it exists or existed in the past. Because at some point someone can "unregister" the membership and it won't exist anymore.

Perhaps right now it doesn't matter as we dont "revoke" credentials, but in the future I guess we can't realy on this?

rymnc:
> ah just realized that `memberExists` should actually be `memberExistsOrExisted` (forget the naming). I mean, since the contract is dummy and just emit events, it can't really know if the member exists. It just can know if it exists or existed in the past. Because at some point someone can "unregister" the membership and it won't exist anymore.
> 
> Perhaps right now it doesn't matter as we dont "revoke" credentials, but in the future I guess we can't realy on this?

Err, the contract does store commitments onchain so it is possible to know if the member exists

alrevuelta:
> Err, the contract does store commitments onchain so it is possible to know if the member exists

ah in that case we're good. Thought it was pure event based with no state kept in the contract. noice.

alrevuelta (approved):
lgtm

richard-ramos:
Is this is the PR that introduced these contract changes? 
https://github.com/waku-org/waku-rln-contract/pull/8

rymnc:
> Is this is the PR that introduced these contract changes? [waku-org/waku-rln-contract#8](https://github.com/waku-org/waku-rln-contract/pull/8)

yessir

Issue #1427, opened 2022-11-28
bug: `0.0.0.0` included in `listenAddrs` of identify message


fryorcraken:
### Problem

When the identify protocol is execute between js-waku and nwaku v0.13.0, the nwaku peer includes a multiaddrs with `/ip4/0.0.0.0/...` in the `listenAddrs` field of the identify message.

Note that js-libp2p uses `listenAddrs` when `signedPeerRecord` is not present in the identify message: https://github.com/libp2p/js-libp2p/blob/bae32bafce75a3801a7a96f77a9ccf43b3208f9c/src/identify/index.ts#L361

AFAIK, `signedPeerRecord` is not use by nwaku.

Note that according to libp2p dev, only _valid_ addresses are expected:
- https://github.com/libp2p/js-libp2p/issues/1484#issuecomment-1327641899
- https://github.com/libp2p/js-libp2p/pull/1485#pullrequestreview-1194583511

### Impact

This is the cause for https://github.com/waku-org/js-waku/issues/751 which is the highest priority bug for js-waku

### To reproduce

This happens for any connection, including connection to a local nwaku node or a wakuv2 prod fleet nwaku node.

1.  https://examples.waku.org/light-js
2. set `debug=waku*,libp2p*` in local storage
3. click `dial`

You will see:

```
 libp2p:identify no signed peer record received from 16Uiu2HAkvWiyFsgRhuJEb9JfjYxEkoHLgnUQmr1N5mKWnYjxYRVm +0ms index.js:29065:10
22:45:42.181 libp2p:identify falling back to legacy addresses from 16Uiu2HAkvWiyFsgRhuJEb9JfjYxEkoHLgnUQmr1N5mKWnYjxYRVm +0ms
```

To get the `0.0.0.0` value it's trickier as you would need to use a modify version of js-llbp2p that logs the content of the identify message (see libp2p link in first section).

### Expected behavior

1. `listenAddrs` only contains values that are expected to be reachable by remote node
2. `signedPeerRecord` is used or equivalent functionality in ENR format https://rfc.vac.dev/spec/31/

### nwaku version/commit hash

`v0.13`

fryorcraken:
Cc @Menduist. Not sure if it's nwaku usage of nim-libp2p or an issue on nim-libp2p side.

jm-clius:
How it currently works:
[...]

PR #1982, opened 2023-09-01
fix: preventing IP 0.0.0.0 from being published

gabrielmer:
# Description

Currently, the IP address `0.0.0.0` is being published in the Identify message when there's no external addresses found. This happens in cases when Waku nodes are being run locally.

This goes against [RFC 1122](https://www.rfc-editor.org/rfc/rfc1122#page-29), section 3.2.1.3 where referring to the `0.0.0.0` IP it states:
>This host on this network. MUST NOT be sent, except as a source address as part of an initialization procedure by which the host learns its own IP address.

In that case, we will publish localhost's loopback interface's address `127.0.0.1` instead.
# Changes

- [x] Added function `formatListenAddress`, which checks if IP `0.0.0.0` is present in the node's published address and if so, replaces it for `127.0.0.1`.
- [x] Added an unit test suite for this functionality
- [x] Allowing the function `addPeer` to modify an existing peer entry whenever it's missing ENR data
- [x] Modified tests that used and expected the IP `0.0.0.0` to use IP `127.0.0.1`  

## How to test

When running the node locally, if no IP address is specified (or if the 0 IP is provided as an input), the node should publish the `127.0.0.1` IP instead.

image

If a custom IP is provided, it should be published

image

## Issue

closes [#1427](https://github.com/waku-org/nwaku/issues/1427)

### DIFF
--- a/tests/waku_core/test_published_address.nim
+++ b/tests/waku_core/test_published_address.nim
@@ -1,0 +1,27 @@
+{.used.}
+
+import
+  stew/shims/net as stewNet,
+  std/strutils,
+  testutils/unittests
+import
+  ../testlib/wakucore,
+  ../testlib/wakunode
+
+suite "Waku Core - Published Address":
+  
+  test "Test IP 0.0.0.0":  
+    let 
+      node = newTestWakuNode(generateSecp256k1Key(), ValidIpAddress.init(
+        "0.0.0.0"),Port(0))
+ 
+    check:
+      ($node.announcedAddresses).contains("127.0.0.1")
+
+  test "Test custom IP":  
+    let 
+      node = newTestWakuNode(generateSecp256k1Key(), ValidIpAddress.init(
+        "8.8.8.8"),Port(0))
+ 
+    check:
+      ($node.announcedAddresses).contains("8.8.8.8")
--- a/waku/node/peer_manager/peer_manager.nim
+++ b/waku/node/peer_manager/peer_manager.nim
@@ -121,8 +121,9 @@ proc addPeer*(pm: PeerManager, remotePeerInfo: RemotePeerInfo, origin = UnknownO
   discard remotePeerInfo.peerId.extractPublicKey(publicKey)
 
   if pm.peerStore[AddressBook][remotePeerInfo.peerId] == remotePeerInfo.addrs and
-     pm.peerStore[KeyBook][remotePeerInfo.peerId] == publicKey:
-    # Peer already managed
+     pm.peerStore[KeyBook][remotePeerInfo.peerId] == publicKey and
+     pm.peerStore[ENRBook][remotePeerInfo.peerId].raw.len > 0:
+    # Peer already managed and ENR info is already saved
     return
 
   trace "Adding peer to manager", peerId = remotePeerInfo.peerId, addresses = remotePeerInfo.addrs
--- a/tests/all_tests_waku.nim
+++ b/tests/all_tests_waku.nim
@@ -5,7 +5,8 @@ import
   ./waku_core/test_namespaced_topics,
   ./waku_core/test_time,
   ./waku_core/test_message_digest,
-  ./waku_core/test_peers
+  ./waku_core/test_peers,
+  ./waku_core/test_published_address
 
 
 # Waku archive test suite
--- a/tests/wakunode_jsonrpc/test_jsonrpc_admin.nim
+++ b/tests/wakunode_jsonrpc/test_jsonrpc_admin.nim
@@ -31,10 +31,10 @@ procSuite "Waku v2 JSON-RPC API - Admin":
   asyncTest "connect to ad-hoc peers":
     # Create a couple of nodes
     let
-      node1 = newTestWakuNode(generateSecp256k1Key(), ValidIpAddress.init("0.0.0.0"), Port(60600))
-      node2 = newTestWakuNode(generateSecp256k1Key(), ValidIpAddress.init("0.0.0.0"), Port(60602))
+      node1 = newTestWakuNode(generateSecp256k1Key(), ValidIpAddress.init("127.0.0.1"), Port(60600))
+      node2 = newTestWakuNode(generateSecp256k1Key(), ValidIpAddress.init("127.0.0.1"), Port(60602))
       peerInfo2 = node2.switch.peerInfo
-      node3 = newTestWakuNode(generateSecp256k1Key(), ValidIpAddress.init("0.0.0.0"), Port(60604))
+      node3 = newTestWakuNode(generateSecp256k1Key(), ValidIpAddress.init("127.0.0.1"), Port(60604))
       peerInfo3 = node3.switch.peerInfo
 
     await allFutures([node1.start(), node2.start(), node3.start()])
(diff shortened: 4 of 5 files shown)


waku/node/peer_manager/peer_manager.nim
```diff
@@ -121,6 +121,7 @@ proc addPeer*(pm: PeerManager, remotePeerInfo: RemotePeerInfo, origin = UnknownO
   discard remotePeerInfo.peerId.extractPublicKey(publicKey)
 
   if pm.peerStore[AddressBook][remotePeerInfo.peerId] == remotePeerInfo.addrs and
+     not ($remotePeerInfo.addrs).contains("127.0.0.1") and
```
vpavlin, line 124: Can you explain why this is needed, please? I cannot wrap my head around it, sorry:D

alrevuelta:
> When running the node locally, if no IP address is specified (or if the 0 IP is provided as an input), the node should publish the 127.0.0.1 IP instead.

Unsure I undestand why we want to advertise 127.0.0.1? If no `nat=extip:<IP>` is provided, perhaps we should default to the public ip of the node? Which well, we would need a 3rd party to give it to us. How is nimbus doing this? Perhaps we can use it as inspiration.

waku/node/peer_manager/peer_manager.nim
gabrielmer: Yes! So we were having issues with the peer exchange tests in which we weren't able to add peers when running locally after the initial change we made. `pm.peerStore[AddressBook][remotePeerInfo.peerId]` equals `127.0.0.1` in local runs, and `remotePeerInfo.addrs` used to equal `0.0.0.0` but after our fix to the Identify message, it equals `127.0.0.1`.

Because this if statement was always yielding true, it considered the IP `127.0.0.1` as duplicate, returned early from the addPeer procedure and we weren't able to add local peers. So we added a special case to bypass that check when it comes to localhost.

vpavlin: But that is strange, isn't it? We are also checking the publicKey, so that would mean that not only the address is the loopback interface, but also the pubKey is the same - IOW this check should be fals if you have multiple local node peers as long as they have different peerId, so the explicit check for 127.0.0.1 should not be necessary. Or I am still missing something:D

Can you share which tests were failing?

jm-clius:
@alrevuelta just to give some background: this only addresses the marginal case where no external IP of the node was configured or could be determined in any other way (NAT-PMP, UPnP, etc.). In other words, it really only affects the case where you are running node(s) locally. The issue was that you would then wrongly advertise the default `0.0.0.0` (in `identify`), which causes unexpected behaviour and even crashes in e.g. js-waku. There are two possible solutions to this:
1. The most correct thing to do is probably to not advertise any addresses. However, this makes local building and experiments a bit difficult (e.g. you won't have a usable address logged for `Listening on...`).
2. Default to localhost/loopback when you really couldn't determine an external IP by any means. This should be the exception, but at least allows proper local setups, consistent logging/debugging.

waku/node/peer_manager/peer_manager.nim
gabrielmer: Now that I look further into it I agree that there's something weird in there.
Maybe the public key check is reduntant/wrong? If `pm.peerStore[KeyBook][remotePeerInfo.peerId]` has to have the value of the public key for a peerId, and `publicKey` is a variable with the public key of the peer we are checking - aren't they supposed to be always the same?

The test that fails without that line is `nwaku/tests/test_waku_peer_exchange test 'retrieve and provide peer exchange peers from discv5'`

gabrielmer:
Labeled the PR again as draft, as it is still WIP.

We're investigating why the following [test](https://github.com/waku-org/nwaku/blob/master/tests/test_waku_peer_exchange.nim#L74-L194) is not passing without the special exception for the the loopback address.

waku/node/peer_manager/peer_manager.nim
gabrielmer: > Weird, I commented out this line and ran the test you mentioned locall and everything passed:)

That's really weird, without that line I couldn't get the test to pass neither locally nor in the CI. I'll DM you to have a look :)

Menduist:
The proper way to do this is to use an addressMapper that would remove every addresses whos IP is 0.0.0.0, I think
cc @diegomrsantos

gabrielmer:
We tried the approach of using an addressMapper to remove every `0.0.0.0` IP. However, the previous workaround seems necessary as it doesn't affect `Identify` when we return an empty set of IPs. The address field is required in `Identify` and if we have an empty set, `0.0.0.0` is populated.
cc @diegomrsantos

waku/node/peer_manager/peer_manager.nim
```diff
@@ -121,8 +121,9 @@ proc addPeer*(pm: PeerManager, remotePeerInfo: RemotePeerInfo, origin = UnknownO
   discard remotePeerInfo.peerId.extractPublicKey(publicKey)
 
   if pm.peerStore[AddressBook][remotePeerInfo.peerId] == remotePeerInfo.addrs and
-     pm.peerStore[KeyBook][remotePeerInfo.peerId] == publicKey:
-    # Peer already managed
+     pm.peerStore[KeyBook][remotePeerInfo.peerId] == publicKey and
+     pm.peerStore[ENRBook][remotePeerInfo.peerId].raw.len > 0:
```
jm-clius, line 125: Ah, could you add a comment (to the issue, not the code) on why this is necessary? Is this because these peers were explicitly connected to in addition to being added via discovery?

jm-clius (commented):
Approach lgtm now. :) Feel free to mark as ready for review when you feel it's mergeable and I'll do a full review, but should be an approve.

gabrielmer:
The issue found was the following:

We are calling `addPeer` multiple times for the same `peerId`. Only the first call adds the peer and subsequent calls for the same peer are dropped in the duplicate check.
The problem is that for every peer, its first call to `addPeer` doesn't contain the peer's ENR data, but subsequent calls do. Because we see the peer as duplicate and return immediately, we are not filling the peer's ENR information which is later needed for establishing connections.

With the new fix, if we call `addPeer` for a `peerId` that we already have saved, if we don't have its ENR information yet we don't fail the duplicate check and continue to update the peer's data with the latest information. It means that `addPeer` will work as an addition function when the peer doesn't exist, and as an update function in the case it exists but information is incomplete.

Why did it work before? Out of luck it seems.
Before we changed the published IP from `0.0.0.0` to `127.0.0.1`, the first call for adding each peer assigned the peer the IP `0.0.0.0` and subsequent calls for the same peer were done with the IP `127.0.0.1`.
As the calls that contained the ENR information were done with a different IP than the first call, the check didn't see it as duplicate and let it update the peer information correctly.

After changing the published IP to `127.0.0.1`, the first call for adding a peer is done with `127.0.0.1`, the same IP as the later calls. So in this case it was correctly flagged as duplicate earlier and did not let us update the ENR information once we had it.

SionoiS (approved):
LGTM

The only question I have is unrelated, why do we call `addPeer()` twice?

I'm looking into it...

gabrielmer:
> LGTM
> 
> The only question I have is unrelated, why do we call `addPeer()` twice?
> 
> I'm looking into it...

According to stack traces I added, the calls come from the functions `connectRelay` and `searchLoop`, so it seems to be exactly the reason Hanno pointed out - that nodes are being both and added by discovery and explicitly in the test

tests/waku_core/test_published_address.nim
```diff
@@ -0,0 +1,27 @@
+{.used.}
+
+import
+  stew/shims/net as stewNet,
+  std/[strutils],
```
Ivansete-status, line 5: Minor nitpick
```suggestion
  std/strutils,
```

Ivansete-status (approved):
LGTM! Thanks for it!

tests/waku_core/test_published_address.nim
gabrielmer: Done! Thank you! :)

jm-clius (approved):

jm-clius:
LGTM! Well done. :)

vpavlin (approved):
Lgtm! Good job, sir!

PR #1984, opened 2023-09-04
chore(rln-relay): use the only key from keystore if only 1 exists

rymnc:
# Description

from the perspective of an end user, i will most likely have only one membership per chainId-contractAddress pair.
To account for this, we are removing `rln-relay-membership-index` as a compulsory argument to wakunode2, and instead, using the `only` available key in the keystore

# Changes

- [x] Changes to the keystore to account for this
- [x] Updated tests 
- [x] Updated config of wakunode2 and chat2 to make it an `Option` type
- [x] Updated OnchainGroupManager init process to set the membershipIndex appropriately

## How to test

1. Create a fresh keystore using rln-keystore-generator - contractAddress: `0x0A988fd9CA5BAebDf098b8A73621b2AaDa6492E8` 
1. Run wakunode2 with the following args 
```shell
./build/wakunode2 \
--rln-relay=true \
--rln-relay-dynamic=true \
--rln-relay-cred-password=$KEYSTORE_PASSWORD \
--rln-relay-cred-path=rlnKeystore.json \
--rln-relay-tree-path=rln_tree_new_1.db \
--rln-relay-eth-contract-address=0x0A988fd9CA5BAebDf098b8A73621b2AaDa6492E8  \
--rln-relay-eth-client-address=$SEPOLIA_URL \
--nodekey=161c2836197e8918d7eb1039758423795f0dc697a018ed55d5931aa9b0e5efb3 \
--log-level=debug
```

You should see that it is able to fetch the membership index from the decrypted keystore. Additionally, the rpc api is able to append proofs to messages as well.

## Issue

Addresses a part of #1906

### DIFF
--- a/waku/waku_keystore/keystore.nim
+++ b/waku/waku_keystore/keystore.nim
@@ -4,7 +4,9 @@ else:
   {.push raises: [].}
 
 import
-  options, json, strutils,
+  options,
+  json,
+  strutils,
   std/[tables, os]
 
 import
@@ -144,7 +146,7 @@ proc addMembershipCredentials*(path: string,
       let encodedMembershipCredential = membership.encode()
       let keyfileRes = createKeyFileJson(encodedMembershipCredential, password)
       if keyfileRes.isErr():
-        return err(AppKeystoreError(kind: KeystoreCreateKeyfileError, 
+        return err(AppKeystoreError(kind: KeystoreCreateKeyfileError,
                                     msg: $keyfileRes.error))
 
       # We add it to the credentials field of the keystore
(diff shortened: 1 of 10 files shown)


waku/waku_keystore/keystore.nim
```diff
@@ -183,13 +185,22 @@ proc getMembershipCredentials*(path: string,
...
-      if not keystoreCredentials.hasKey(key):
+      if keystoreCredentials.len == 0:
         # error
         return err(AppKeystoreError(kind: KeystoreCredentialNotFoundError,
-                                    msg: "Credential not found in keystore"))
+                                    msg: "No credentials found in keystore"))
+      var keystoreCredential: JsonNode
+      if keystoreCredentials.len == 1:
+        for v in keystoreCredentials.getFields().values():
+          keystoreCredential = v
```
rymnc, line 195: this is quite ugly - the compiler was throwing errors when i tried doing this -
```nim
keystoreCredential = keystoreCredentials.getFields().values()[0]
```

vpavlin: Isn't that because values() returns iterator, which means you'd have to turn it into `seq` first before you can access the Nth element?

vpavlin (approved):

waku/waku_keystore/keystore.nim
rymnc: thanks, addressed in 99b3fe6c!

alrevuelta (approved):
lgtm! thanks!

PR #1990, opened 2023-09-04
chore(ci): use commit instead of master for docker image

vpavlin:
# Description

per @alrevuelta's reuest - replacing `master` tag with git commit for traceability 

# Changes

- [ ] replace `master` tag with `git rev-parse --short HEAD` output


alrevuelta (approved):
thanks! unsure if ci failure is related.

vpavlin:
> thanks! unsure if ci failure is related.

No, I don't think it is

Ivansete-status (approved):
That awesome! Thanks for it !

PR #1991, opened 2023-09-05
Add utils to Dockerfile

alrevuelta:
Add utils to Dockerfile


vpavlin (approved):
Not a fan of adding both curl and wget since both serve basically the same purpose, but probably not that big of a deal

PR #1994, opened 2023-09-06
chore(rln-relay): clean up nullifier table every MaxEpochGap

rymnc:
# Description

This PR clears the nullifier table every time there are more than `MaxEpochGap` epochs stored in it

# Changes

- [x] Updated rln-relay validator to clear nullifier log on invocation
- [x] Added test

## Issue

Addresses a part of #1906

### DIFF
--- a/tests/waku_rln_relay/test_wakunode_rln_relay.nim
+++ b/tests/waku_rln_relay/test_wakunode_rln_relay.nim
@@ -7,17 +7,12 @@ import
   testutils/unittests,
   chronicles,
   chronos,
-  libp2p/crypto/crypto,
-  libp2p/peerid,
-  libp2p/multiaddress,
   libp2p/switch,
-  libp2p/protocols/pubsub/pubsub,
-  eth/keys
+  libp2p/protocols/pubsub/pubsub
 import
   ../../../waku/waku_core,
   ../../../waku/waku_node,
   ../../../waku/waku_rln_relay,
-  ../../../waku/waku_keystore,
   ../testlib/wakucore,
   ../testlib/wakunode
 
@@ -206,7 +201,6 @@ procSuite "WakuNode - RLN relay":
     await node1.mountRlnRelay(WakuRlnConfig(rlnRelayDynamic: false,
       rlnRelayCredIndex: some(1.uint),
       rlnRelayTreePath: genTempPath("rln_tree", "wakunode_4"),
-      rlnRelayBandwidthThreshold: 0,
     ))
 
     await node1.start()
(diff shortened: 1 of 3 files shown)


tests/waku_rln_relay/test_wakunode_rln_relay.nim
```diff
@@ -1,23 +1,18 @@
 {.used.}
 
 import
-  std/[sequtils, tempfiles],
+  std/[tempfiles, sequtils],
```
vpavlin, line 4: isn't s before t in the alphabet?:D

rymnc: err, my bad - that change shouldn't be in the diff

vpavlin (approved):
lgtm

waku/waku_rln_relay/rln_relay.nim
```diff
@@ -300,13 +299,26 @@ proc appendRLNProof*(rlnPeer: WakuRLNRelay,
   msg.proof = proofGenRes.get().encode().buffer
   return true
 
+proc clearNullifierLog(rlnPeer: WakuRlnRelay) =
```
alrevuelta, line 302: just to understand it. what we really want to keep is the `seq[ProofMetadata]` of the epochs that are no older than `MaxEpochGap` right?

I mean, in your case you keep `MaxEpochGap` at max, no matter if they meet this condition. But we would be interested in:

```nim
for epoch in rlnPeer.nullifierLog.keys():
  if (epoch + MaxEpochGap) < now_epoch:
    rlnPeer.nullufierLog.del(epoch)
```

this doesn't require an `OrderedTable` (unsure if ordered tables impact performance).

not a blocker

alrevuelta (approved):
lgtm thanks! left a comment in case you find it relevant.

side questions:
* so as it was before, we should have a memory leak? if the table was not being emptied, that should grow to infinity? just asking because i didnt saw the leak in the simulations.
* what would be the napkin math on the worst case size of this table? I assume `seq[ProofMetadata]` can be as long as the max amount of memberships, since we would only store 1 proof per epoch per membership (if it sent a message). And then if we allow a diff between clocks of 20 seconds and 1 message every 1 second MaxEpochGap=20/1. Having 10k memberhips, is it fair to assume that we can have 10.000 * 20 entries at worse?

rymnc:
> lgtm thanks! left a comment in case you find it relevant.
> 
> side questions:
> * so as it was before, we should have a memory leak? if the table was not being emptied, that should grow to infinity? just asking because i didnt saw the leak in the simulations.
> * what would be the napkin math on the worst case size of this table? I assume `seq[ProofMetadata]` can be as long as the max amount of memberships, since we would only store 1 proof per epoch per membership (if it sent a message). And then if we allow a diff between clocks of 20 seconds and 1 message every 1 second MaxEpochGap=20/1. Having 10k memberhips, is it fair to assume that we can have 10.000 * 20 entries at worse?

dependa on how long were your simulations - but yes, it would've grown to infinity

PR #1996, opened 2023-09-06
fix(p2p): fix possible connectivity issue

alrevuelta:
Description:
* Temporally comment inbound relay peer prunning, since it may be causing connectivity issues. Note that prunning by IP is kept, this just affects the general prunning.
* Increase default peer store size. If the max amount of peers is `x` the peerstore size should be way bigger than that. If no, we wont benefit much from the exponential backoff, since after prunning, this state is "forgotten".
* Increase peerstore prunning interval to 10 minutes.

Test:
* Simulated in an environment with `waku-simulator`

### DIFF
--- a/waku/node/peer_manager/peer_manager.nim
+++ b/waku/node/peer_manager/peer_manager.nim
@@ -10,7 +10,6 @@ import
   chronicles,
   metrics,
   libp2p/multistream,
-  libp2p/connmanager,
   libp2p/muxers/muxer,
   libp2p/nameresolving/nameresolver
 import
--- a/waku/node/peer_manager/peer_manager.nim
+++ b/waku/node/peer_manager/peer_manager.nim
@@ -10,6 +10,7 @@ import
   chronicles,
   metrics,
   libp2p/multistream,
+  libp2p/connmanager,
   libp2p/muxers/muxer,
   libp2p/nameresolving/nameresolver
 import
@@ -53,7 +54,7 @@ const
   ConnectivityLoopInterval = chronos.seconds(15)
 
   # How often the peer store is pruned
-  PrunePeerStoreInterval = chronos.minutes(5)
+  PrunePeerStoreInterval = chronos.minutes(10)
 
   # How often metrics and logs are shown/updated
   LogAndMetricsInterval = chronos.minutes(3)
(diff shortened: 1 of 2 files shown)


waku/node/peer_manager/peer_manager.nim
```diff
@@ -583,15 +584,16 @@ proc connectToRelayPeers*(pm: PeerManager) {.async.} =
   let totalRelayPeers = inRelayPeers.len + outRelayPeers.len
   let inPeersTarget = maxConnections - pm.outRelayPeersTarget
 
-  if inRelayPeers.len > pm.inRelayPeersTarget:
-    await pm.pruneInRelayConns(inRelayPeers.len - pm.inRelayPeersTarget)
+  # TODO: Temporally disabled. Might be causing connection issues
+  #if inRelayPeers.len > pm.inRelayPeersTarget:
+  #  await pm.pruneInRelayConns(inRelayPeers.len - pm.inRelayPeersTarget)
```
Ivansete-status, line 589: I vote to remove that directly. If we need it in the future we can review this commit and easily restore it back. The TODO is likely to be there forever :)
```suggestion
```

waku/node/waku_switch.nim
```diff
@@ -101,7 +101,7 @@ proc newWakuSwitch*(
     if peerStoreCapacity.isSome():
       b = b.withPeerStore(peerStoreCapacity.get())
     else:
-      let defaultPeerStoreCapacity = int(round(float64(maxConnections)*1.25))
+      let defaultPeerStoreCapacity = int(maxConnections)*5
```
Ivansete-status, line 104: Shall we have a comment indicating why we set the max PeerStoreCapacity to that?
From the PR's comment is clear, but I'd mention that in the code as well

waku/node/peer_manager/peer_manager.nim
```diff
@@ -10,6 +10,7 @@ import
   chronicles,
   metrics,
   libp2p/multistream,
+  libp2p/connmanager,
```
Ivansete-status, line 13: Is this really needed?

Ivansete-status (approved):
LGTM! Thanks for it!

waku/node/waku_switch.nim
alrevuelta: will address it in the future. maybe this shouldn't even be configurable and always harcoded depending on the amount of maxConnections.

waku/node/peer_manager/peer_manager.nim
alrevuelta: would like to leave it there, since its something I would like to revisit. Some peer prunning in inbound connections is needed, the open question is how.

waku/node/peer_manager/peer_manager.nim
alrevuelta: good catch, fixed
47e3a4942cbf2ed42d41b2093e99733e2329dde0

PR #1997, opened 2023-09-06
chore(postgres): Adding healtcheck and reconnection mechanism to the postgres archive driver

Ivansete-status:
# Description
It starts an asynchronous infinite task that checks the connectivity with the database.

In case of error, the `postgres_healthcheck` task tries to reconnect for a while, and if it determines that the connection cannot be resumed, then it invokes a callback indicating that situation.

For the case of the `wakunode2` app, that callback quits the application itself and adds a log trace indicating the connectivity issue with the database.

# Changes

- [x] New `proc resetConnPool*(pool: PgAsyncPool)` to force close the pool if either the database got down or lost the connection with it.
- [x] `wakunode2` quits if it has _Store_ mounted and the _Postgres_ database goes down for more than 30''.
- [x] New `waku/waku_archive/driver/postgres_driver/postgres_healthcheck.nim` that contains the infinite health check proc.

### Test case: `wakunode2` stops if the database gets stopped.
1. Run node A - `./build/wakunode2 --config-file=cfg_node_a.txt`
[cfg_node_a.txt](https://github.com/waku-org/nwaku/files/12537414/cfg_node_a.txt)

2. Run node B - `./build/wakunode2 --config-file=cfg_node_b.txt`
[cfg_node_b.txt](https://github.com/waku-org/nwaku/files/12537417/cfg_node_b.txt)

3. Run a _Postgres_ instance: `docker compose -f postgres-docker-compose.yml up`
Content of postgres-docker-compose.yml:
```
version: "3.8"

services:
  db:
    image: postgres:9.6-alpine
    restart: always
    environment:
      POSTGRES_PASSWORD: test123
    ports:
      - "5432:5432"
```
4. Close the _Postgres_ instance started in 3. After less than 30'' the `wakunode2` should stop.

### Test case: `wakunode2` resumes connection with the _Postgres_ database.
1. Run node A - `./build/wakunode2 --config-file=cfg_node_a.txt`
[cfg_node_a.txt](https://github.com/waku-org/nwaku/files/12537414/cfg_node_a.txt)

2. Run node B - `./build/wakunode2 --config-file=cfg_node_b.txt`
[cfg_node_b.txt](https://github.com/waku-org/nwaku/files/12537417/cfg_node_b.txt)

3. Run a _Postgres_ instance: `docker compose -f postgres-docker-compose.yml up`
Content of postgres-docker-compose.yml:
```
version: "3.8"

services:
  db:
    image: postgres:9.6-alpine
    restart: always
    environment:
      POSTGRES_PASSWORD: test123
    ports:
      - "5432:5432"
```
4.
  - Perform a _Store_ request: `curl -d '{"jsonrpc":"2.0","id":"id","method":"get_waku_v2_store_v1_messages"}' --header "Content-Type: application/json" http://localhost:8546`
  In this case, a valid response should be returned.
  - Close the database started in 3.
  - Send another _Store_ request.
  Now, the next error should appear: `{"jsonrpc":"2.0","id":"id","error":{"code":-32000,"message":"get_waku_v2_store_v1_messages raised an exception","data":"BAD_REQUEST: invalid cursor"}}`
 - Start the database again (point 3.) and wait for 30''.
 - Send another _Store_ request. A valid response should come.

## Issue
closes https://github.com/waku-org/nwaku/issues/1893

### DIFF
--- a/apps/wakunode2/app.nim
+++ b/apps/wakunode2/app.nim
@@ -411,10 +411,17 @@ proc setupProtocols(node: WakuNode,
         return err("failed to mount waku RLN relay protocol: " & getCurrentExceptionMsg())
 
   if conf.store:
+    var onErrAction = proc(msg: string) {.gcsafe, closure.} =
+      ## Action to be taken when an internal error occurs during the node run.
+      ## e.g. the connection with the database is lost and not recovered.
+      error "Unrecoverable error occurred", error = msg
+      quit(QuitFailure)
+
     # Archive setup
     let archiveDriverRes = ArchiveDriver.new(conf.storeMessageDbUrl,
                                              conf.storeMessageDbVacuum,
-                                             conf.storeMessageDbMigration)
+                                             conf.storeMessageDbMigration,
+                                             onErrAction)
     if archiveDriverRes.isErr():
       return err("failed to setup archive driver: " & archiveDriverRes.error)
 
--- a/waku/common/databases/db_postgres/pgasyncpool.nim
+++ b/waku/common/databases/db_postgres/pgasyncpool.nim
@@ -136,6 +136,20 @@ proc getConnIndex(pool: PgAsyncPool):
     pool.conns[index].busy = true
     return ok(index)
 
+proc resetConnPool*(pool: PgAsyncPool): Future[DatabaseResult[void]] {.async.} =
+  ## Forces closing the connection pool.
+  ## This proc is intended to be called when the connection with the database
+  ## got interrupted from the database side or a connectivity problem happened.
+
+  for i in 0..<pool.conns.len:
+    pool.conns[i].busy = false
+
+  (await pool.close()).isOkOr:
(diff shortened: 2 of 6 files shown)


jm-clius (approved):
Thanks! Curious to know - it seems that on failure, we will only continue trying to reconnect for 20 seconds and then quit the application. Is that enough to allow e.g. for a postgresql restart without crashing the node?

vpavlin:
Well, it is something between 20s and 50s (healthcheck every 30s + 20s of retries, so if PG fails right before the healthcheck, then you have 20s, otherwise it is more:D), but I do agree that this might be highly dependent on the platform you run, so we should turn this into an option eventually (locally 20s+ should be enouch, in Kubernetes under high load, it could take a bit longer)

vpavlin (approved):

SionoiS (approved):

PR #2007, opened 2023-09-07
libwaku: annotate thread-shared callback with .nimcall.

Ivansete-status:
## Description
This is needed to ensure the `proc relayEventCallback(...` proc is not considered a closure. Therefore, no hidden GC'ed environment context is kept when sharing or copying its address to another thread.

This PR is motivated by the following Jacek's comment on Discord:
> other types of "hidden" references are closures - these appear a bit here and there but must also be avoided - the closure environment itself is a hidden GC:d type, so passing closures between threads is also not allowed - thus, any proc types used with threads must be annotated with {.nimcall.} at least which prevents them from capturing a garbage-collected environment

## Changes

- [x] Annotating `proc relayEventCallback(...` with `.nimcall.`.

### DIFF
--- a/library/libwaku.nim
+++ b/library/libwaku.nim
@@ -45,7 +45,8 @@ type
 var extEventCallback*: WakuCallBack = nil
 
 proc relayEventCallback(pubsubTopic: PubsubTopic,
-                        msg: WakuMessage): Future[void] {.async, gcsafe.} =
+                        msg: WakuMessage):
+                        Future[void] {.async, gcsafe, nimcall.} =
   # Callback that hadles the Waku Relay events. i.e. messages or errors.
   if not isNil(extEventCallback):
     try:


arnetheduck:
ah - top-level functions don't become closures so they don't need callbacks.. 
the comment applies only to "nested" `proc`'s that capture variables from their surrounding environment, for example:


```nim
prox xxx() = echo "hello" # not a closure

proc yyy(): proc() =
  let hello = "hello"
  proc zzz = echo hello # captures the hello variable

  return zzz # closure is created here
```

in this case, the `nimcall` I referred to applies to proc _type declarations_`, ie
```nim
type MyCallback = proc() {.nimcall, gcsafe.}

type MyThreadSafeObject =
   callback: MyCallback

let data = createShared(MyThreadSafeObject)

data[].callback = xxx # this works because "normal" functions are `nimcall` by default

data[].callback = yyy() # this doesn't work - cannot assign a closure to a variable whose proc type includes `nimcall`
```

Ivansete-status:
> ah - top-level functions don't become closures so they don't need callbacks.. the comment applies only to "nested" `proc`'s...

Beautiful! Then, we are thread-safe as long as we only share "top-defined" `proc`s with the _Waku Thread_

PR #2011, opened 2023-09-07
feat: Rest endoint /health for rln

NagyZoltanPeter:
# Description
For RLN readiness /health rest endpoint is added.
It checks if wakunode is ready - currently it means rln is mounted correctly. returns proper http status upon.

# Changes

- [x] New rest endpoint /health introduced
- [x] Test that supports rln is added to verify

## How to test

1. make test EXPERIMENTAL=true

or

1. ./scripts/build_rln.sh ./vendor/zerokit
2. nim c -r -d:chronicles_log_level=DEBUG -d:os=Linux --parallelBuild:2 --verbosity:1 -d:chronicles_log_level=DEBUG -d:release -d:rln --passL:./vendor/zerokit/target/release/librln.a --passL:-lm ./tests/wakunode_rest/test_rest_health.nim 

## Issue
[feat(rest): Add /health endpoint to rest api](https://github.com/waku-org/nwaku/issues/1988)

### DIFF
--- a/tests/wakunode_rest/test_rest_health.nim
+++ b/tests/wakunode_rest/test_rest_health.nim
@@ -1,0 +1,77 @@
+{.used.}
+
+import
+  std/tempfiles,
+  stew/shims/net,
+  testutils/unittests,
+  presto,
+  presto/client as presto_client,
+  libp2p/peerinfo,
+  libp2p/multiaddress,
+  libp2p/crypto/crypto
+import
+  ../../waku/waku_node,
+  ../../waku/node/waku_node as waku_node2,  # TODO: Remove after moving `git_version` to the app code.
+  ../../waku/node/rest/server,
+  ../../waku/node/rest/client,
+  ../../waku/node/rest/responses,
+  ../../waku/node/rest/health/handlers as health_api,
+  ../../waku/node/rest/health/client as health_api_client,
+  ../../waku/waku_rln_relay,
+  ../testlib/common,
+  ../testlib/wakucore,
+  ../testlib/wakunode
+
+
+proc testWakuNode(): WakuNode =
+  let
+    privkey = crypto.PrivateKey.random(Secp256k1, rng[]).tryGet()
+    bindIp = ValidIpAddress.init("0.0.0.0")
+    extIp = ValidIpAddress.init("127.0.0.1")
+    port = Port(58000)
+
+  newTestWakuNode(privkey, bindIp, port, some(extIp), some(port))
+
+
+suite "Waku v2 REST API - health":
+  asyncTest "Get node health info - GET /health":
+    # Given
+    let node = testWakuNode()
+    await node.start()
+    await node.mountRelay()
+
+    let restPort = Port(58001)
+    let restAddress = ValidIpAddress.init("0.0.0.0")
+    let restServer = RestServerRef.init(restAddress, restPort).tryGet()
+
+    installHealthApiHandler(restServer.router, node)
+    restServer.start()
+    let client = newRestHttpClient(initTAddress(restAddress, restPort))
+
+    # When
+    var response = await client.healthCheck()
+
+    # Then
+    check:
+      response.status == 503
+      $response.contentType == $MIMETYPE_TEXT
+      response.data == "Node is not inititialized"
+
+    # now kick in rln (currently the only check for health)
+    await node.mountRlnRelay(WakuRlnConfig(rlnRelayDynamic: false,
+      rlnRelayCredIndex: some(1.uint),
+      rlnRelayTreePath: genTempPath("rln_tree", "wakunode"),
+    ))
+
+    # When
+    response = await client.healthCheck()
+
+    # Then
+    check:
+      response.status == 200
+      $response.contentType == $MIMETYPE_TEXT
+      response.data == "Node is healthy"
+
+    await restServer.stop()
+    await restServer.closeWait()
+    await node.stop()
--- a/tests/wakunode_rest/test_rest_health.nim
+++ b/tests/wakunode_rest/test_rest_health.nim
@@ -55,7 +55,7 @@ suite "Waku v2 REST API - health":
     check:
       response.status == 503
       $response.contentType == $MIMETYPE_TEXT
-      response.data == "Node is not inititialized"
+      response.data == "Node is not ready"
 
     # now kick in rln (currently the only check for health)
     await node.mountRlnRelay(WakuRlnConfig(rlnRelayDynamic: false,
--- a/tests/wakunode_rest/test_rest_health.nim
+++ b/tests/wakunode_rest/test_rest_health.nim
@@ -28,7 +28,7 @@ proc testWakuNode(): WakuNode =
     privkey = crypto.PrivateKey.random(Secp256k1, rng[]).tryGet()
     bindIp = ValidIpAddress.init("0.0.0.0")
     extIp = ValidIpAddress.init("127.0.0.1")
-    port = Port(58000)
+    port = Port(0)
 
   newTestWakuNode(privkey, bindIp, port, some(extIp), some(port))
 
--- a/waku/node/rest/health/client.nim
+++ b/waku/node/rest/health/client.nim
@@ -1,0 +1,30 @@
+when (NimMajor, NimMinor) < (1, 4):
+  {.push raises: [Defect].}
+else:
+  {.push raises: [].}
+
+import
+  chronicles,
+  json_serialization,
+  json_serialization/std/options,
+  presto/[route, client]
+import
+  ../serdes,
+  ../responses
+
+logScope:
+  topics = "waku node rest health_api"
+
+proc decodeBytes*(t: typedesc[string], value: openArray[byte],
+                  contentType: Opt[ContentTypeData]): RestResult[string] =
+  if MediaType.init($contentType) != MIMETYPE_TEXT:
+    error "Unsupported contentType value", contentType = contentType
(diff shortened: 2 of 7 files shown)


waku/node/waku_node.nim
```diff
@@ -910,4 +910,5 @@ proc isReady*(node: WakuNode): Future[bool] {.async.} =
       return false
     return await node.wakuRlnRelay.isReady()
   ## TODO: add other protocol `isReady` checks
+  echo "***NZP*** rln not defined!"
```
alrevuelta, line 913: leftover?

tests/wakunode_rest/test_rest_health.nim
```diff
@@ -0,0 +1,77 @@
...
+      response.data == "Node is not inititialized"
+
+    # now kick in rln (currently the only check for health)
+    await node.mountRlnRelay(WakuRlnConfig(rlnRelayDynamic: false,
+      rlnRelayCredIndex: some(1.uint),
+      rlnRelayTreePath: genTempPath("rln_tree", "wakunode"),
+    ))
+
+    # When
+    response = await client.healthCheck()
```
alrevuelta, line 67: Since this is a `static` membership (not onchain) here we are not really testing anyhing, since the healthcheck is aways true.
https://github.com/waku-org/nwaku/blob/master/waku/waku_rln_relay/group_manager/static/group_manager.nim#L113-L117

but yeah, testing it with onchain might be tricky with many things to mock. I'm fine with leaving it like this but leaving a todo for the future.

waku/node/rest/health/client.nim
```diff
@@ -0,0 +1,30 @@
+when (NimMajor, NimMinor) < (1, 4):
```
alrevuelta, line 1: not sure if its too much to have another folder `health`? Would try to keep the minimum amount of folders. But yeah, not sure it fits in the existing ones.

waku/node/rest/health/openapi.yaml
```diff
@@ -0,0 +1,41 @@
+openapi: 3.0.3
```
alrevuelta, line 1: can't we have a single `openapi.yaml` for all waku apis?

waku/node/rest/health/handlers.nim
```diff
@@ -0,0 +1,37 @@
...
+
+    let isReadyStateFut = node.isReady()
+    if not await isReadyStateFut.withTimeout(futIsReadyTimout):
+       return RestApiResponse.internalServerError("Health check timed out")   
+
+    var msg = "Node is healthy"
+    var status = Http200
+
+    if not isReadyStateFut.read(): 
+      msg = "Node is not inititialized"
```
alrevuelta, line 34: perhaps `node is not ready`?

waku/node/rest/health/handlers.nim
```diff
@@ -0,0 +1,37 @@
...
+      status = Http503
+
+    return RestApiResponse.textResponse(msg, status)
```
alrevuelta, line 37: unsure about the best practices of `health` endpoints, but would it make sense to return a json that we can extend in the future? In case we want more granularity on the state of the services. For example, we may want to say that is not healthy, but provide a list on which services are not healthy.

No need to define it right now but allow for future extension. return json instead of textResponse?

alrevuelta (commented):
left some comments thanks! will manually test it later with onchain rln

waku/node/waku_node.nim
NagyZoltanPeter: Ah... sht... I will remove it right now.

tests/wakunode_rest/test_rest_health.nim
NagyZoltanPeter: I would say unit test of rest endpoint just tests if it can return proper answer based upon node's interface.
That onchain test should test isReady interface of waku_node and not through rest interface test. WDYT?

waku/node/rest/health/client.nim
NagyZoltanPeter: It currently fits with existing rest api structures. I would stick to it for now.

waku/node/rest/health/openapi.yaml
NagyZoltanPeter: I had the same idea while working on filter rest api... but it would involve dir / code changes, like I would move relevant types to higher level due to duplications. I would leave it for later restructure work.

waku/node/rest/health/handlers.nim
NagyZoltanPeter: If you wish so?! :-) I will change!

waku/node/rest/health/handlers.nim
NagyZoltanPeter: I can change it but in this case http status holds the information rather. 
For meaningful message, text can be enough. Not sure what information we can structure into json later.
For checking status of different submodules or protocols sounds more like an admin interface than a health. Wdyt?

tests/wakunode_rest/test_rest_health.nim
alrevuelta: right, closing

waku/node/rest/health/client.nim
alrevuelta: sure, closing.

waku/node/rest/health/openapi.yaml
alrevuelta: yep makes sense. out of scope for this pr but would love to have everything in the same file at some point.

waku/node/rest/health/handlers.nim
alrevuelta: yep indeed. http status code has the information.

i was suggesting to return a json in case we want to add more information in the future. so if for example its failing, you have information such as (example taken from prysm node): so you know which service is not healthy

```
/healthz

*rpc.Service: OK
*prometheus.Service: OK
*p2p.Server: OK
*powchain.Web3Service: OK
*attestation.Service: OK
*operations.Service: OK
*blockchain.ChainService: OK
*sync.Service: ERROR not initially synced

```

no need to do it now, but be prepared to extend it.

Ivansete-status (approved):
LGTM! Thanks for it!. Just minor comments

waku/node/rest/health/handlers.nim
```diff
@@ -0,0 +1,37 @@
...
+  presto/route
+import
+  ../../waku_node,
+  ../responses,
+  ../serdes
+
+logScope:
+  topics = "waku node rest health_api"
+
+const ROUTE_HEALTH* = "/health"
```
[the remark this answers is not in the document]
NagyZoltanPeter: Well, I would stick to previous habits found in other Rest endpoint path consts...
So either change all or none.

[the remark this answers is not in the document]
Ivansete-status: We need to change all then :D

jm-clius (approved):
LGTM! Thanks.

rymnc (approved):
LGTM, thanks for this!

tests/wakunode_rest/test_rest_health.nim
```diff
@@ -0,0 +1,77 @@
...
+  ../testlib/wakucore,
+  ../testlib/wakunode
+
+
+proc testWakuNode(): WakuNode =
+  let
+    privkey = crypto.PrivateKey.random(Secp256k1, rng[]).tryGet()
+    bindIp = ValidIpAddress.init("0.0.0.0")
+    extIp = ValidIpAddress.init("127.0.0.1")
+    port = Port(58000)
```
vpavlin, line 31: Let's not hardcode ports unless necessary (JSON-RPC or REST at the moment) - please use `Port(0)`

waku/node/rest/health/handlers.nim
vpavlin: I was gonna add the same comment - let's use JSON rather than text - it is extensible and easier to process with tools than plain text

waku/node/rest/health/openapi.yaml
vpavlin: It should be possible to combine the existing openapi.yaml into a single one (which we could publish as part of release?). But I'd say having them separate makes them more maintainable than a single massive yaml?

waku/node/rest/health/handlers.nim
```diff
@@ -0,0 +1,37 @@
...
+
+const FutIsReadyTimout = 5.seconds
+
+proc installHealthApiHandler*(router: var RestRouter, node: WakuNode) =
+
+  router.api(MethodGet, ROUTE_HEALTH) do () -> RestApiResponse:
+
+    let isReadyStateFut = node.isReady()
```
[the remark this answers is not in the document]
NagyZoltanPeter: I will add as there is one little stuff to fix in test.

waku/node/rest/health/handlers.nim
NagyZoltanPeter: I'm ok with it, but will do in a separate PR is possible.
Also I think we need to enhance a bit client decoders (not just here) to accept json/text both due error cases returns text type, mixing them can cause exception. It's on my list already, but will do in all rest endpoints where necessary.

vpavlin: Unrelated to this issue, obviously, but since we are talking about it - would it make more sense to always return json (even for errors)?

tests/wakunode_rest/test_rest_health.nim
NagyZoltanPeter: Fixed.

alrevuelta (approved):
lgtm! when ci passes

PR #2013, opened 2023-09-07
fix(rln-relay): missed roots during sync

### LINKED ISSUE #2002: bug(rln): problems syncing the rln tree, different roots
### Problem

Using `0x0A988fd9CA5BAebDf098b8A73621b2AaDa6492E8`
Commit `5638bd06bbab5f52aeb399b2cf162c8b834ce011`

When continuosly injecting new memberships into the rln contract (at a rate of one per block or so) different nodes calculate different merkle roots when syncing. This creates a "fork" in the network, where some nodes reject valid messages and others accept them. This issue seems to happen (or at least more often) when continuosly injecting new memberships into the contract.

After stopping injecting memberships and inspecting the database of different nodes, they seem to split into:
* Nodes that have synced 5673 memberships
* Nodes with 5672 memberships

This off-by-one membership seems to be the rootcause of the issue. If one of the first memberships was lost, then the nodes will calculate completely different trees and won't be able to validate the same messages.

Notes:
* Cleaning the db and resyncing again seems to fix the issue.
* This happens to 1 every 10 nodes or so.

rymnc:
# Description

on a new block header, processes min(g.latestProcessedBlock, newHeadNumber) to account for missed blocks 

# Changes

- [x] newBlockHeader processing strategy

## Issue

closes #2002

### DIFF
--- a/tests/waku_rln_relay/test_rln_group_manager_onchain.nim
+++ b/tests/waku_rln_relay/test_rln_group_manager_onchain.nim
@@ -193,6 +193,8 @@ suite "Onchain group manager":
       manager.initialized
       manager.rlnContractDeployedBlockNumber > 0
 
+    await manager.stop()
+
   asyncTest "should error on initialization when loaded metadata does not match":
     let manager = await setup()
     await manager.init()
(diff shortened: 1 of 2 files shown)


alrevuelta (approved):
lgtm!

PR #2006, opened 2023-09-07
chore(ci): execute js-waku integration tests on image build

vpavlin:
# Description

Depends on https://github.com/waku-org/js-waku/pull/1541 (merged)

This PR adds another step in our CI workflow - it executes js-waku tests on top of the image built from a PR - you can think of this as an integration test attempt

# Changes

- [ ] add reusable workflow from js-waku
- [ ] update PR image build comment to also mention the commit hash (cc @alrevuelta)

PR #2017, opened 2023-09-08
chore(ci): add js-waku test to pre-release workflow

vpavlin:
# Description
This PR builds on top of #2006 and adds image build and js-waku execution to pre-release workflow - i.e. `nightly` and release candidates

# Changes

- [ ] add `input` to docker build job to be able to pass explicit tag
- [ ] add job to decide a tag/release name in pre-release WF


rymnc (approved):
LGTM as far as my knowledge of github actions goes 😅

.github/workflows/pre-release.yml
```diff
@@ -16,7 +16,22 @@ env:
...
+  tag-name:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v3
+
+    - name: Vars
+      id: vars
+      run: |
+        TAG=$([[ "${{github.ref}}" == "refs/heads/master" ]] && echo "${{env.RELEASE_NAME}}" || echo ${{github.ref}} | sed 's#refs/tags/##')
```
Ivansete-status, line 28: We don't tend to have many frequent PRs being merged. However, I suggest creating a separate _release branch_, e.g. `release/0.20`, each time we create a release so that we make sure that nothing else gets added inadvertently during the release process. (A possible improvement for the future.)

.github/workflows/pre-release.yml
```diff
@@ -76,9 +88,23 @@ jobs:
...
+    with:
+      image_tag: ${{ needs.tag-name.outputs.tag }}
+    secrets: inherit
+
+  js-waku:
+    needs: build-docker-image
+    uses: waku-org/js-waku/.github/workflows/ci.yml@master
+    with:
+      nim_wakunode_image: ${{ needs.build-docker-image.outputs.image }}
+      caller: nwaku
```
Ivansete-status, line 103: For me to understand: what does that do?
Thanks in advance!

Ivansete-status (approved):
LGTM! Thanks for it!

.github/workflows/pre-release.yml
vpavlin: Yeah, we do that already - the check for `master` here is because the same PR is used for `nightly`, so if the build comes from `master`, rather than `v*-rc.*`, we want to call it `nightly`

.github/workflows/pre-release.yml
vpavlin: Yes, `uses` at `job` level specifies the workflow YAML file to use for the job to execute (can be local to the repo or external -as seen here), then `with` specifes input parameters for the called workflow - the PR/pre-release image and then also specify the call comes from `nwaku` so that we can limit the jobs in js-waku CI that get executed (e.g. we do not want to do release:) )

weboko (approved):

PR #2018, opened 2023-09-11
fix(rln-relay): waku_rln_number_registered_memberships metrics appropriately handled

### LINKED ISSUE #2008: bug(rln): Wrong rln registered memberships metric
### Problem
Metric `waku_rln_number_registered_memberships` is wrong, since it doesnt take into account previously stored memberships in the database.

See: https://github.com/waku-org/nwaku/blob/1fb13b096710bda1ac73e72027ae25ed36a2b3a4/waku/waku_rln_relay/group_manager/on_chain/group_manager.nim#L108

Fix: Apart from increasing +1 when detecting a new membership, we should also update the metric when loading a state from the db, with the loaded amount of memberships.

rymnc:
# Description

This PR sets the number of registered memberships to the `start` index (i.e last index that was inserted) + number of commitments - commitments removed

Upon startup, if we fetch the tree db, there is no way currently from the zerokit ffi api to fetch the number of members registered - hence, we set the metric based on the event thrown by the contract.

# Changes

- [x] changed counter to gauge, since we want to `set` the value of the collector 

## Issue

closes #2008

### DIFF
--- a/waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
+++ b/waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
@@ -105,7 +105,8 @@ method atomicBatch*(g: OnchainGroupManager,
     let operationSuccess = g.rlnInstance.atomicWrite(some(start), idCommitments, toRemoveIndices)
   if not operationSuccess:
     raise newException(ValueError, "atomic batch operation failed")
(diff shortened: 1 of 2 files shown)


alrevuelta (approved):
lgtm thanks!

rymnc:
merging since the ci failure is unrelated

PR #2023, opened 2023-09-11
Test waku filter subscribe ping

AlejandroCabeza:
# Description
Implement set of tests for waku filter.

# Changes
- [X] Add asynchronous testing utilities: asyncSetup and asyncTeardown
- [X] Move commonly used methods to standalone file.
- [X] Implement first batch of SUBSCRIBER_PING tests.


jm-clius:
Haven't had a chance for a detail review yet, but approach looks good so far! Will take another look tomorrow. Feel free to mark as "ready for review" once comfortable that it can be merged :)

jm-clius (approved):
This looks like a very complete suite of ping tests. Thanks! One thing which we might consider doing is to deprecate the existing tests which are now duplicated (once we've audited that they indeed test the same functionality). An example would be this existing ping test: https://github.com/waku-org/nwaku/blob/f805d67c6c2e6772f8e82d3c1cf5781b53748814/tests/waku_filter_v2/test_waku_filter.nim#L20-L53. Perhaps only after adding the entire new suite?

AlejandroCabeza:
> This looks like a very complete suite of ping tests. Thanks! One thing which we might consider doing is to deprecate the existing tests which are now duplicated (once we've audited that they indeed test the same functionality). An example would be this existing ping test:
> 
> https://github.com/waku-org/nwaku/blob/f805d67c6c2e6772f8e82d3c1cf5781b53748814/tests/waku_filter_v2/test_waku_filter.nim#L20-L53
> . Perhaps only after adding the entire new suite?

At first I also thought to do it afterwards, but now I think it's a better idea to do it as part of this PR so as to make it easier to review. Any objections?

jm-clius:
>Any objections?

None. Can add complete suite and remove old ones at the same time. I'd still suggest incrementing though (i.e. review and merge ping tests, then subscribe, etc.)

AlejandroCabeza:
> > Any objections?
> 
> None. Can add complete suite and remove old ones at the same time. I'd still suggest incrementing though (i.e. review and merge ping tests, then subscribe, etc.)

Of course, this PR is closed in terms of scope. Don't like doing big PRs, they're messy to review.

PR #2025, opened 2023-09-12
chore: move SubscriptionManager under waku_core 

NagyZoltanPeter:
# Description
Cherry pick from[Filter V2 Rest API PR](https://github.com/waku-org/nwaku/pull/1890) to separate one.
FilterV1 client hold FilterPushHandler and SubscriptionManager which was not the right place to import from.
Better moved to under waku_core/subscription to be easily reused.

## How to test
All existing tests on Waku Filter (v1 or legacy) covers the change properly.

### DIFF
--- a/waku/waku_core/subscription.nim
+++ b/waku/waku_core/subscription.nim
@@ -1,0 +1,7 @@
+import
+  ./subscription/subscription_manager,
+  ./subscription/push_handler
+
+export
+  subscription_manager,
+  push_handler
--- a/waku/waku_core/subscription/subscription_manager.nim
+++ b/waku/waku_core/subscription/subscription_manager.nim
@@ -1,0 +1,48 @@
+when (NimMajor, NimMinor) < (1, 4):
+  {.push raises: [Defect].}
+else:
+  {.push raises: [].}
+
+import
+  std/tables,
+  stew/results,
+  chronicles,
+  chronos
+
+import
+  ./push_handler,
+  ../topics,
+  ../message
+
+## Subscription manager
+type SubscriptionManager* = object
+    subscriptions: TableRef[(string, ContentTopic), FilterPushHandler]
+
+proc init*(T: type SubscriptionManager): T =
+  SubscriptionManager(subscriptions: newTable[(string, ContentTopic), FilterPushHandler]())
+
+proc clear*(m: var SubscriptionManager) =
+  m.subscriptions.clear()
+
+proc registerSubscription*(m: SubscriptionManager, pubsubTopic: PubsubTopic, contentTopic: ContentTopic, handler: FilterPushHandler) =
+  try:
+    # TODO: Handle over subscription surprises
+    m.subscriptions[(pubsubTopic, contentTopic)]= handler
+  except CatchableError:
+    error "failed to register filter subscription", error=getCurrentExceptionMsg()
+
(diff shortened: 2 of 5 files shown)


waku/waku_core/subscription/subscription_manager.nim
```diff
@@ -0,0 +1,48 @@
...
+import
+  std/tables,
+  stew/results,
+  chronicles,
+  chronos
+
+import
+  ./push_handler,
+  ../topics,
+  ../message
```
Ivansete-status, line 15: Maybe any of these imports can be deleted from `waku/waku_filter/client.nim`

waku/waku_filter/client.nim
```diff
@@ -27,50 +27,7 @@ logScope:
...
-  except:  # TODO: Fix "BareExcept" warning
-    discard
-
-proc getSubscriptionsCount(m: SubscriptionManager): int =
-  m.subscriptions.len()
-
-
 ## Client
-
-type  MessagePushHandler* = proc(requestId: string, msg: MessagePush): Future[void] {.gcsafe, closure.}
```
Ivansete-status, line 72: This type is not needed, right?

Ivansete-status (approved):
LGTM! :100:

waku/waku_filter/client.nim
NagyZoltanPeter: yes, unused!

waku/waku_core/subscription.nim
gabrielmer, line 1: It might me kind of a dumb question, but what's the purpose of exporting the same things that get imported?
Is it so they can be then can all be imported together in one line?

waku/waku_core/subscription/subscription_manager.nim
SionoiS, line 1: Pretty funny since I came up with the same thing but to track Relay handlers when using content topics in #1983.

Looks fine for now but maybe we could use generics in the future so that it can support any kind of handlers.

I would add a comment and/or an issue to this effect.

SionoiS (approved):

waku/waku_core/subscription.nim
NagyZoltanPeter: Exactly, you don't need to than specify separately but enough to import waku_core.

NagyZoltanPeter: nim does not have that ugly feature of C/C++ that expose everything included through header files.

tests/wakunode_rest/test_rest_filter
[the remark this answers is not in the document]
NagyZoltanPeter: Thanks, sure it should not be here. I will remove of course.... my lazy compilation result.

waku/waku_core/subscription/subscription_manager.nim
NagyZoltanPeter: Yeah, I would agree on a joint effort to reslove such duplications around.
This pr is just about to separate such things I added into the biiiig FilterV2 Rest pr.

waku/waku_core/subscription/subscription_manager.nim
NagyZoltanPeter: I can check if there is any unused ones.

gabrielmer:
LGTM :):)

gabrielmer (approved):

waku/waku_core/subscription/subscription_manager.nim
NagyZoltanPeter: There is no unused import in this client.nim - surprisingly

PR #2028, opened 2023-09-12
chore(postgres): not loading the libpq library by default & better user feedback

Ivansete-status:
# Description
Before this change, when a new `nwaku` user wanted to start a node, the next error appeared if the [libpq](https://www.postgresql.org/docs/9.5/libpq.html) library wasn't installed on his/her system.
```
could not load: libpq.so(.5|)
(compile with -d:nimDebugDlOpen for more information)
```

This is indeed confusing and very little information is given. Especially confusing because it appeared even the user not mounting the _Store_ protocol.

This change aims to enhance that and, by default, if the user doesn't mount the _Store_ protocol, the node will start as usual.

# Changes

- [x] Adding explicit Makefile parameter to load the _postgres_ module, and therefore, add the dependency with the libpq library.
- [x] Removing unused imports from `waku/waku_archive/archive.nim` (not directly related to the issue being enhanced in this PR.)
- [x] Adding more feedback if the user tries to mount the _Store_ protocol and use _Postgres_ as the archive backend.

## How to test

1. Run the `wakunode2` without _Store_ mounted. The node should start properly, even if the `libpq` is not being installed.
2. Run the `wakunode2` mounting _Store_ with _SQLite_. The node should start properly.
3. Run `wakunode2` mounting _Store_ with _Postgres_ and no `libpq` is present in the system. In this case, the next message should appear and the `wakunode2` ends immediately:
 "Postgres has been configured but not been compiled. Check compiler definitions."
4.
  i. `make POSTGRES=1 wakunode2 -j11`   ( In this case, the user has _postgres_ in his/her mindset )
  ii. Run `wakunode2` mounting _Store_ with _Postgres_ and no `libpq` is present in the system.
  iii. The next message should appear, which is a little bit more descriptive than the one that appeared in all the cases:
     `libpq.so.5: cannot open shared object file: No such file or directory`
     `libpq.so: cannot open shared object file: No such file or directory`
     `could not load: libpq.so(.5|)`

5. Follow all the points from 4. but having installed the `libpq`. The node should start properly.


waku/waku_archive/driver/builder.nim
```diff
@@ -78,22 +80,33 @@ proc new*(T: type ArchiveDriver,
...
+          return err("error creating table: " & newTableRes.error)
+      except CatchableError:
+        return err("exception creating table: " & getCurrentExceptionMsg())
+
+      return ok(driver)
+
+    else:
+      ## Error variable to be used if the user is trying to mount Store with a Postgres database
+      var errPgMsg = "ERROR:"
+      errPgMsg &= "The POSTGRES flag should be passed to the make command, e.g. "
```
NagyZoltanPeter, line 104: It sounds me that the message is more like targeting developers not node operators.
It's lucky when it is met.
I would rather say something "This build does not support store with PostgresSQL."

NagyZoltanPeter (approved):
Apart from the error message explanation it looks good to me. Thank you!

waku/waku_archive/driver/builder.nim
```diff
@@ -78,22 +80,33 @@ proc new*(T: type ArchiveDriver,
...
+        if newTableRes.isErr():
+          return err("error creating table: " & newTableRes.error)
+      except CatchableError:
+        return err("exception creating table: " & getCurrentExceptionMsg())
+
+      return ok(driver)
+
+    else:
+      ## Error variable to be used if the user is trying to mount Store with a Postgres database
+      var errPgMsg = "ERROR:"
```
jm-clius, line 103: Ah, the error message here (i.e. the `builder.nim`) can have no knowledge of the encapsulating application, `wakunode2`, or `make` commands, etc. It's good that we're trying to be helpful, but at most the driver knows something like: "Postgres has been configured but not been compiled. Check compiler definitions."
I think it's OK to link to the install guide here, even though strictly speaking this module should not even know it's part of nwaku. :)

(Not for this PR, but the correct way to do it is to enumerate the error types here, rather than just returning error strings. That way the application `wakunode2` receives an error result when setting up the archive and the application can then log more instructions as it has the context to do so. For now, this will be overkill.)

jm-clius (approved):
LGTM! Thanks for this. :) I add one comment re the content of the error message - it should be more generic to not assume knowledge of build environment and `wakunode2` in the driver. Will this also fix the requirement for `libpq` when running `make test` in a clean environment? I think we might have to hide postgres tests behind the same compiler definition too.

Ivansete-status:
> LGTM! Thanks for this. :) I add one comment re the content of the error message - it should be more generic to not assume...

Super good catch re `make test`!. I'll review that.

I'll also make the message more generic as you both suggested :)

Ivansete-status:
**Weekly Update**

- _achieved_: Make the `nwaku` to not directly depend on the `libpq` library unless the user doesn't explicitly configure it as _Store_ with _Postgres_. 
- _next_: -

Issue #1576, opened 2023-02-26
chore: update resolved enr ip when using `dns4-domain-name` flag


alrevuelta:
### Background

When using `dns4-domain-name` eg

```
./build/wakunode2 --dns4-domain-name=google.com --discv5-discovery=true
```

The node creates an ENR with ip `0.0.0.0` and the multi is constructed using that ip `/ip4/0.0.0.0/tcp/60000/p2p/16Uiu2HAmFj9mVkzj81XBFTVK1475iKoPPiakdgZtoPTtPaTYqqRy`.

Example of the generated ENR with the above command (decode it [here](https://enr-viewer.com/))
```
enr:-Ki4QAnkkUPeFFRuIxB9KkPQqO9lKIgFxc84egqGM7mBj7bdPGlUxsnqERtpD1Z2i0-nVxyyDquUlPAq7-u9KcQxbNkBgmlkgnY0gmlwhAAAAACKbXVsdGlhZGRyc5EADzYKZ29vZ2xlLmNvbQbqYIlzZWNwMjU2azGhAy2kCEHyIKJ1iAyAzger0hd5kzbxgpYJyOvLFfzElg9Yg3RjcILqYIV3YWt1MgE
```

The thing is that since discv5 doesn't understand our `multiaddrs` field, this node won't be discoverable by disv5, since the address that is being advertised is `0.0.0.0`.

In other protocols using libp2p and discv5 (eg Ethereum) the address that is advertised in the enr is the ip that results of resolving `dns4-domain-name`.

Example. In prysm using the following, will lead to an ENR that contains the IP that resolved from google.com. See logs:

```
docker run prysmaticlabs/prysm-beacon-chain:latest --p2p-host-dns=google.com --accept-terms-of-use
```

### Acceptance criteria
* When running `./build/wakunode2 --dns4-domain-name=some_domain` the constructed ENR `ip` field contains the resolved IP instead of just `0.0.0.0`.

gabrielmer:
**Weekly Update**

- _next_: analyze and fix issue

gabrielmer:
**Weekly Update**

- _achieved_: implemented solution that does DNS IP resolution during node bringup when no external IP is found but a DNS address is provided.
Validated and tested "happy paths" of the solution, raised draft PR and got feedback about the solution
-_next_: discuss and define the system's behavior on errors, implement error handling and adding tests for this feature.

gabrielmer:
**Weekly Update**

[...]

PR #2030, opened 2023-09-13
chore: resolving DNS IP and publishing it when no extIp provided

gabrielmer:
# Description

When running a node and no external IP is provided but a DNS4 address is, we want the ENR to use the resolved DNS IP instead of using `0.0.0.0`, which is currently the case.

# Changes

- [x] If there's no external IP but a DNS address was provided, do a DNS IP lookup and use the result as external IP in the node's network configuration
- [x] Updated procedure `newTestWakuNode` to account for the above scenario
- [x] Added test cases for the success and failure flows of the DNS lookup

## Issue

closes [#1576](https://github.com/waku-org/nwaku/issues/1576)

### DIFF
--- a/apps/wakunode2/internal_config.nim
+++ b/apps/wakunode2/internal_config.nim
@@ -1,9 +1,12 @@
 import
-  std/options,
-  stew/shims/net,
-  stew/results,
+  chronicles,
+  chronos,
   libp2p/crypto/crypto,
-  libp2p/multiaddress
+  libp2p/multiaddress,
+  libp2p/nameresolving/dnsresolver,
+  std/options,
+  stew/results,
+  stew/shims/net
 import
   ../../waku/common/utils/nat,
   ../../waku/node/config,
@@ -18,6 +21,23 @@ proc validateExtMultiAddrs*(vals: seq[string]):
     multiaddrs.add(multiaddr)
   return ok(multiaddrs)
 
+proc dnsResolve*(domain: string, conf: WakuNodeConf): Future[Result[string, string]] {.async} =
+    
+  # Use conf's DNS servers
+  var nameServers: seq[TransportAddress]
+  for ip in conf.dnsAddrsNameServers:
+    nameServers.add(initTAddress(ip, Port(53))) # Assume all servers use port 53
+  
+  let dnsResolver = DnsResolver.new(nameServers)
+
+  # Resolve domain IP
+  let resolved = await dnsResolver.resolveIp(domain, 0.Port, Domain.AF_UNSPEC)
+
+  if resolved.len > 0:
+    return ok(resolved[0].host) # Use only first answer
+  else:
+    return err("Could not resolve IP from DNS: empty response")
+
 proc networkConfiguration*(conf: WakuNodeConf,
                            clientId: string,
                            ): NetConfigResult =
@@ -30,7 +50,7 @@ proc networkConfiguration*(conf: WakuNodeConf,
   if natRes.isErr():
     return err("failed to setup NAT: " & $natRes.error)
 
-  let (extIp, extTcpPort, _) = natRes.get()
+  var (extIp, extTcpPort, _) = natRes.get()
 
   let
     dns4DomainName = if conf.dns4DomainName != "": some(conf.dns4DomainName)
@@ -69,6 +89,18 @@ proc networkConfiguration*(conf: WakuNodeConf,
         relay = conf.relay
       )
 
+  # Resolve and use DNS domain IP
+  if dns4DomainName.isSome() and extIp.isNone():
+    try:
+      let dnsRes = waitFor dnsResolve(conf.dns4DomainName, conf)
+      
+      if dnsRes.isErr():
+        return err($dnsRes.error) # Pass error down the stack
+      
+      extIp = some(ValidIpAddress.init(dnsRes.get()))
+    except CatchableError:
+      return err("Could not update extIp to resolved DNS IP: " & getCurrentExceptionMsg())
+  
   # Wrap in none because NetConfig does not have a default constructor
   # TODO: We could change bindIp in NetConfig to be something less restrictive
   # than ValidIpAddress, which doesn't allow default construction
--- a/tests/test_wakunode.nim
+++ b/tests/test_wakunode.nim
@@ -1,7 +1,7 @@
 {.used.}
 
 import
-  std/sequtils,
+  std/[sequtils,strutils],
   stew/byteutils,
   stew/shims/net as stewNet,
   testutils/unittests,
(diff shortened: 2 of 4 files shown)


apps/wakunode2/internal_config.nim
```diff
@@ -18,6 +21,20 @@ proc validateExtMultiAddrs*(vals: seq[string]):
...
+    
+  # Use conf's DNS servers
+  var nameServers: seq[TransportAddress]
+  for ip in conf.dnsAddrsNameServers:
+    nameServers.add(initTAddress(ip, Port(53))) # Assume all servers use port 53
+  
+  let dnsResolver = DnsResolver.new(nameServers)
+
+  # Resolve domain IP
+  let resolved = await dnsResolver.resolveIp(domain, 0.Port, Domain.AF_UNSPEC)
```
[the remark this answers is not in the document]
gabrielmer: True, great catch!

We should add a check to see if the returned sequence is empty, and either throw an exception or continue the program as it currently works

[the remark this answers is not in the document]
gabrielmer: I like the option of throwing an exception better so the user knows that something went wrong and the DNS provided is not working properly.

What do you think?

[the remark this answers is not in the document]
vpavlin: I agree, always fail during the setup if something is not according to expectation

apps/wakunode2/internal_config.nim
```diff
@@ -3,7 +3,10 @@ import
   stew/shims/net,
   stew/results,
   libp2p/crypto/crypto,
-  libp2p/multiaddress
+  libp2p/multiaddress,
```
vpavlin, line 6: We try to keep the lit of import alphabetically ordered as much as possible

apps/wakunode2/internal_config.nim
```diff
@@ -18,6 +21,23 @@ proc validateExtMultiAddrs*(vals: seq[string]):
     multiaddrs.add(multiaddr)
   return ok(multiaddrs)
 
+proc dnsResolve*(domain: string, conf: WakuNodeConf): Future[string] {.async} =
```
vpavlin, line 24: Can we return `Result` here and either return `ok(resolved[0].host)` or `err("Could not resolve IP from DNS: empty response")`

This way the logic is encapsulated in the proc and does not leak to the caller

tests/test_wakunode.nim
```diff
@@ -215,6 +215,64 @@ suite "WakuNode":
       node.announcedAddresses.len == 1
       node.announcedAddresses.contains(expectedDns4Addr)
 
+  asyncTest "Node uses dns4 resolved ip in announced addresses if no extIp is provided":
+    let
+      nodeKey = generateSecp256k1Key()
+      bindIp = ValidIpAddress.init("0.0.0.0")
+      bindPort = Port(61010)
```
vpavlin, line 222: Use `Port(0)` to get the port auto assigned to avoid potential confilicts in CI

tests/test_wakunode.nim
```diff
@@ -215,6 +215,64 @@ suite "WakuNode":
...
+    check:
+      ipStr.len() > 0
+      not ipStr.contains("0.0.0.0")
+      not ipStr.contains("127.0.0.1")
+
+  asyncTest "Node creation fails when invalid dns4 address is provided":
+    let
+      nodeKey = generateSecp256k1Key()
+      bindIp = ValidIpAddress.init("0.0.0.0")
+      bindPort = Port(61010)
```
vpavlin, line 247: Same here - use `Port(0)`

tests/testlib/wakunode.nim
```diff
@@ -38,10 +40,26 @@ proc newTestWakuNode*(nodeKey: crypto.PrivateKey,
                       discv5UdpPort = none(Port),
                       agentString = none(string),
                       peerStoreCapacity = none(int)): WakuNode =
+  
+  var resolvedExtIp = extIp
+
+  # Update extPort to default value if it's missing and there's an extIp or a DNS domain 
+  let extPort = if (extIp.isSome() or dns4DomainName.isSome()) and
```
vpavlin, line 47: Why do you need to do this?

gabrielmer: It is based on the following code:
https://github.com/waku-org/nwaku/blob/dc25057a2632baf487a9ae00533db291f2791dc6/apps/wakunode2/internal_config.nim#L48-L52

I copied the behavior we currently have to create a node to `newTestWakuNode` so what we create the test node under the same logic that we use to create production nodes.

If we provide a DNS address and no port to create a Node, we use the default port.
But in `newTestWakuNode` that logic didn't exist, and it attempted to create a node without an extIp provided and without having a default one set, so an exception was thrown

tests/wakunode2/test_app.nim
```diff
@@ -15,12 +15,13 @@ import
...
   ../testlib/wakucore
 
-proc defaultTestWakuNodeConf(): WakuNodeConf =
+proc defaultTestWakuNodeConf*(): WakuNodeConf =
   WakuNodeConf(
     listenAddress: ValidIpAddress.init("127.0.0.1"),
     rpcAddress: ValidIpAddress.init("127.0.0.1"),
     restAddress: ValidIpAddress.init("127.0.0.1"),
     metricsServerAddress: ValidIpAddress.init("127.0.0.1"),
+    dnsAddrsNameServers: @[ValidIpAddress.init("1.1.1.1"), ValidIpAddress.init("1.0.0.1")],
```
SionoiS, line 24: Both use cloudflare maybe add google in there?

SionoiS (approved):
LGTM!

tests/wakunode2/test_app.nim
gabrielmer: Yes! I added only those because they are the ones we configure by default, so tests have a more accurate representation of our nodes.
https://github.com/waku-org/nwaku/blob/6cea737109223bbd2900519d8d958593086b9798/apps/wakunode2/external_config.nim#L115-L118

What do you think is better to do?
Add Google DNS servers to the test anyways, discuss whether to add Google to the default config and if so adding it to the tests, or just leave it the way it is?

SionoiS: You are right it should be consistent. Leave it like this it's fine.

vpavlin: At the same time, there is this issue: https://github.com/waku-org/nwaku/issues/1490

But yeah, let's keep it until someone actually works on ^

jm-clius (approved):
Approving to not be a bottleneck here and the general idea makes sense. The important thing is to get approving reviews from the reviewers that left some comments by re-requesting reviews once you've addressed their concerns. You could also mark specific comments as "resolved" as you go along.

vpavlin (approved):

PR #2035, opened 2023-09-14
test(waku-filter): Subscribe tests batch (2/4)

AlejandroCabeza:
# Description
Implement second batch of subscribe tests for waku filter.

# Changes
- [X] Implement subscribe tests.

# Dependencies
https://github.com/waku-org/nwaku/pull/2034


tests/waku_filter_v2/test_waku_client.nim
```diff
@@ -9,720 +9,941 @@ import
...
-      # Then
-      check:
-        unsubscribedPingResponse.isErr() # Not subscribed
-        unsubscribedPingResponse.error().kind == FilterSubscribeErrorKind.NOT_FOUND
-
-    asyncTest "After Unsubscription":
-      # Given
-      let 
-        subscribeResponse = await wakuFilterClient.subscribe(
+        assert subscribeResponse.isOk(), $subscribeResponse.error
```
NagyZoltanPeter, line 109: Isn't it be unsubscribeResponse here?

tests/waku_filter_v2/test_waku_client.nim
```diff
@@ -9,720 +9,941 @@ import
...
+
+        # Given a valid unsubscription to an existing subscription
+        let unsubscribeResponse = await wakuFilterClient.unsubscribe(
+          serverRemotePeerInfo, pubsubTopic, contentTopicSeq
+        )
+        require:
+          unsubscribeResponse.isOk()
+          wakuFilter.subscriptions.len == 0
+
+        # When sending a message to the previously subscribed content topic
```
NagyZoltanPeter, line 165: Just a little typo here.
```suggestion
        # When sending a message to the previously unsubscribed content topic
```

tests/waku_filter_v2/test_waku_client.nim
```diff
@@ -9,720 +9,941 @@ import
...
+        
+        # Given a valid unsubscription to an existing subscription
+        let unsubscribeResponse = await wakuFilterClient.unsubscribe(
+          serverRemotePeerInfo, pubsubTopic, contentTopicsSeq
+        )
+        require:
+          unsubscribeResponse.isOk()
+          wakuFilter.subscriptions.len == 0
+        
+        # When sending a message to the previously subscribed content topic
```
NagyZoltanPeter, line 239:
```suggestion
        # When sending a message to the previously unsubscribed content topic
```

tests/waku_filter_v2/test_waku_client.nim
```diff
@@ -9,720 +9,941 @@ import
...
+        pushHandlerFuture = newPushHandlerFuture() # Clear previous future
+        let msg4 = fakeWakuMessage(contentTopic=contentTopic)
+        await wakuFilter.handleMessage(pubsubTopic, msg4)
+
+        # Then the message is not pushed to the client
+        check:
+          not await pushHandlerFuture.withTimeout(FUTURE_TIMEOUT)
+
+        # When sending a message to the other previously subscribed content topic
```
NagyZoltanPeter, line 248:
```suggestion
        # When sending a message to the other previously unsubscribed content topic
```

tests/waku_filter_v2/test_waku_client.nim
```diff
@@ -9,720 +9,941 @@ import
...
+        )
+
+        # Then the unsubscription is successful
+        check:
+          unsubscribeResponse1.isOk()
+          wakuFilter.subscriptions.len == 1
+          wakuFilter.subscriptions.hasKey(clientPeerId)
+          wakuFilter.getSubscribedContentTopics(clientPeerId) == otherContentTopicSeq
+        
+        # When sending a message to the previously subscribed content topic
```
NagyZoltanPeter, line 340:
```suggestion
        # When sending a message to the previously unsubscribed content topic
```

tests/waku_filter_v2/test_waku_client.nim
```diff
@@ -9,720 +9,941 @@ import
...
+        let unsubscribeResponse2 = await wakuFilterClient.unsubscribe(
+          serverRemotePeerInfo, otherPubsubTopic, otherContentTopicSeq
+        )
+
+        # Then the unsubscription is successful
+        check:
+          unsubscribeResponse2.isOk()
+          wakuFilter.subscriptions.len == 0
+
+        # When sending a message to the previously subscribed content topic
```
NagyZoltanPeter, line 371:
```suggestion
        # When sending a message to the previously unsubscribed content topic
```

NagyZoltanPeter (changes requested):
I think it is all good! Thank you for the great work! 
I found only some typos and maybe one missed check.

tests/waku_filter_v2/test_waku_client.nim
AlejandroCabeza: Indeed it is. Fixed, along the other typos you mentioned.

The `assert` is a suggestion from @Ivansete-status. What this does is, if the check fails it prints the value on the right, namely `$subscribeResponse.error`, which is useful for debugging.

NagyZoltanPeter (approved):
LGTM!
Its a good idea having that assert like this. I did not know.

tests/waku_filter_v2/test_waku_client.nim
```diff
@@ -9,720 +9,941 @@ import
...
+    var contentTopicSeq {.threadvar.}: seq[ContentTopic]
+    var clientPeerId {.threadvar.}: PeerId
+    var messagePushHandler {.threadvar.}: FilterPushHandler
+    var pushHandlerFuture {.threadvar.}: Future[(string, WakuMessage)]
+
+    asyncSetup:
+      pushHandlerFuture = newPushHandlerFuture()
+      messagePushHandler = proc(
+        pubsubTopic: PubsubTopic, message: WakuMessage
+      ): Future[void] {.async, closure, gcsafe.} =
```
Ivansete-status, line 47: Small nitpick:
I think is not needed to use `Future[void]`. This is done implicitly by the `async` pragma.

tests/waku_filter_v2/test_waku_client.nim
```diff
@@ -9,720 +9,941 @@ import
...
+
+    suite "Subscriber Ping":
+      asyncTest "Active Subscription Identification":
+        # Given
+        let 
+          subscribeResponse = await wakuFilterClient.subscribe(
+            serverRemotePeerInfo, pubsubTopic, contentTopicSeq
+          )
+        require:
+          subscribeResponse.isOk()
```
Ivansete-status, line 74: For that kind of validation, I think it is better to do the next so that we have a better detail in case of failure:
`assert subscribeResponse.isOk(), $subscribeResponse.error`

That applies elsewhere a `Result` is being checked.

tests/waku_filter_v2/test_waku_client.nim
```diff
@@ -9,720 +9,941 @@ import
...
-  
-  asyncTeardown:
-    await allFutures(wakuFilter.stop(), wakuFilterClient.stop(), serverSwitch.stop(), clientSwitch.stop())
-
-  suite "Subscriber Ping":
-    asyncTest "Active Subscription Identification":
-      # Given
-      let 
-        subscribeResponse = await wakuFilterClient.subscribe(
+  suite "MessagePushHandler - Void":
```
Ivansete-status, line 30: For me to understand better. What does "Void" mean in this context?

Ivansete-status (approved):
LGTM! Thanks for it! :100:

tests/waku_filter_v2/test_waku_client.nim
AlejandroCabeza: Good to know, thanks!

tests/waku_filter_v2/test_waku_client.nim
AlejandroCabeza: Agree, must've slipped by when changing. 
I'm actually moving away from require, as if a require doesn't work, it makes the whole test suite fail, so I'd rather use checks or asserts.

Issue #1883, opened 2023-08-03
chore: do not advertise multiaddr with port 0


vpavlin:
### Background

When you explicitly set the `--tcp-port=0`, the resulting MA used by the waku node will contain the port 0:

```
$ ./build/wakunode2 --relay --tcp-port=0
INF 2023-08-03 18:03:15.264+02:00 Initializing networking                    tid=234166 file=waku_node.nim:142 addrs=@[/ip4/10.1.6.152/tcp/0]
INF 2023-08-03 18:03:15.264+02:00 mounting relay protocol                    topics="waku node" tid=234166 file=waku_node.nim:309
INF 2023-08-03 18:03:15.264+02:00 relay mounted successfully                 topics="waku node" tid=234166 file=waku_node.nim:328
INF 2023-08-03 18:03:15.264+02:00 mounting rendezvous discovery protocol     topics="waku node" tid=234166 file=waku_node.nim:778
INF 2023-08-03 18:03:15.264+02:00 mounting libp2p ping protocol              topics="waku node" tid=234166 file=waku_node.nim:736
INF 2023-08-03 18:03:15.264+02:00 mounting store client                      topics="waku node" tid=234166 file=waku_node.nim:543
INF 2023-08-03 18:03:15.264+02:00 Starting Waku node                         topics="waku node" tid=234166 file=waku_node.nim:793 version=v0.19.0-rc.0-5-gd2b607
INF 2023-08-03 18:03:15.264+02:00 PeerInfo                                   topics="waku node" tid=234166 file=waku_node.nim:796 peerId=16U*2SdE95 addrs=@[]
INF 2023-08-03 18:03:15.264+02:00 Listening on                               topics="waku node" tid=234166 file=waku_node.nim:803 full=[/ip4/10.1.6.152/tcp/0/p2p/16Uiu2HAkzF2Q9v7EUJQyiqGbV2GhrkKKGSp5Do3ZtiWpR92SdE95]
```

which is obviously not true. Waku should advertise the real port  it is listening on in the multiaddress

### Acceptance criteria

- [ ] waku advertises the real port it is listening on instead of port 0 when instructed to use a random port

gabrielmer:
**Weekly Update**

- _next_: analyze and fix issue

gabrielmer:
**Weekly Update**

[...]

PR #2037, opened 2023-09-14
chore: not allowing tcp-port 0 input  by the user

gabrielmer:
# Description

When the user chooses port 0, the node's published port is 0. However, that's not the port that we are getting bound to by the kernel.
Therefore, to avoid inconsistencies and imprecisions, we won't allow the user to choose port 0. Either the default port or a port explicitly chosen by the user will be used, avoiding non-deterministic outcomes.

# Changes

- [x] Throwing an error in case port 0 is selected by the user
- [x] Updating tests that configure Waku nodes with port 0 as default, and change them so they use our current default 60000 port

## Issue

closes [#1883](https://github.com/waku-org/nwaku/issues/1883)

### DIFF
--- a/apps/wakunode2/internal_config.nim
+++ b/apps/wakunode2/internal_config.nim
@@ -24,6 +24,9 @@ proc networkConfiguration*(conf: WakuNodeConf,
 
   ## `udpPort` is only supplied to satisfy underlying APIs but is not
   ## actually a supported transport for libp2p traffic.
+  if conf.tcpPort == Port(0):
+    return err("Choosing port 0 is not allowed")
+  
   let natRes = setupNat(conf.nat, clientId,
                         Port(uint16(conf.tcpPort) + conf.portsShift),
                         Port(uint16(conf.tcpPort) + conf.portsShift))
--- a/tests/wakunode2/test_app.nim
+++ b/tests/wakunode2/test_app.nim
@@ -21,6 +21,7 @@ proc defaultTestWakuNodeConf(): WakuNodeConf =
     rpcAddress: ValidIpAddress.init("127.0.0.1"),
     restAddress: ValidIpAddress.init("127.0.0.1"),
     metricsServerAddress: ValidIpAddress.init("127.0.0.1"),
+    tcpPort: Port(60000),
     nat: "any",
     maxConnections: 50,
     topics: @["/waku/2/default-waku/proto"],


gabrielmer:
> Port 0 allows the system to pick a port, which is a feature in many cases

Yes, the main issue here is us publishing the 0 port when that's not really the port allocated by the kernel during runtime.
The final solution we aim for is to update all our internal data structures to reflect the port chosen by the system.

Currently, `WakuNode` is getting built and filled before the port binding happens, so updating all the internal data structures consistently with the actual port might require some major refactoring.
We agreed that initially, restricting the user from choosing port 0 can be our initial solution to avoid this situation. So the user can either use the default port, or choose one explicitly themselves.

@arnetheduck What do you think about it? Are there cases in which this could be problematic?

arnetheduck:
Not sure - seems easier to just get it done than to confuse users for a while and then get it done - that said, I'm not familiar with the scope of refactoring needed - as long as there's a plan, sgtm, though it's always sad when applications are missing this little piece of convenience (specially because we run discovery and auto-nat-config and a number of other surrounding services all built for the purpose of not locking a particular port)

gabrielmer:
> Not sure - seems easier to just get it done than to confuse users for a while and then get it done - that said, I'm not familiar with the scope of refactoring needed - as long as there's a plan, sgtm, though it's always sad when applications are missing this little piece of convenience (specially because we run discovery and auto-nat-config and a number of other surrounding services all built for the purpose of not locking a particular port)



> Not sure - seems easier to just get it done than to confuse users for a while and then get it done - that said, I'm not familiar with the scope of refactoring needed - as long as there's a plan, sgtm, though it's always sad when applications are missing this little piece of convenience (specially because we run discovery and auto-nat-config and a number of other surrounding services all built for the purpose of not locking a particular port)

True, makes sense - it might just confuse users. 
The rationale for postponing it was simply because of inexperience - we felt we needed to know better the codebase in order to make this change properly and efficiently. In the future it should be an easier task to handle for the team. As it wasn't seen enough as a priority for the resources it might currently take, we thought of the temporary solution of not allowing the user to choose the 0 port.
At the end of the day, it is just a matter of resources and priorities. But we might have underestimated its importance, I'll double check and discuss again how we should proceed with this :)

gabrielmer:
After discussing the issue deeper, we decided not to move forward with approach. The plan is to work is to allow port 0 to be chosen by the user and properly handle the kernel's port selection. Therefore, closing this PR.
Opened issue #2042 for tracking

PR #2038, opened 2023-09-14
docs: Clarify running instructions

s-tikhomirov:
Extend README to as it relates to running a node.

# Description

Add an example for running a node with clarifications regarding bootstrapping.

# Changes

- [ ] Add an example command that runs `nwaku` with DNS bootstrapping.
- [ ] Clarify that knowing at least one node is necessary for connecting to the network.
- [ ] Add a link to hard-coded bootstrap nodes.


Ivansete-status (approved):
LGTM! Thanks so much for it!

README.md
```diff
@@ -26,9 +26,14 @@ The standard developer tools, including a C compiler, GNU Make, Bash, and Git. M
 # You'll run `make update` after each `git pull` in the future to keep those submodules updated.
 make wakunode2
 
+# Run with DNS bootstrapping
+./build/wakunode2 --dns-discovery --dns-discovery-url=DNS_BOOTSTRAP_NODE_URL
+
 # See available command line options
 ./build/wakunode2 --help
 ```
+You can use nodes specified [here](https://github.com/waku-org/nwaku/blob/master/apps/chat2/chat2.nim#L375-L378) as bootstrap nodes.
```
vpavlin, line 35: Would it make more sense to point to the https://fleets.status.im/ for the DNS discovery infor?

Ivansete-status:
This can be merged @s-tikhomirov. Thanks!

README.md
gabrielmer, line 35: Notice that if `chat2.nim` file is changed in the future, lines `L375-L378` might point to something else and the readme should be updated.
Other option is to link to the code of a specific commit instead of latest master, but in that case we must also update the readme if the `dnsDiscoveryUrl` is modified.

I think that the best option is to link to a specific commit, and avoid the possibility of pointing in the future to something unrelated, which might confuse users

gabrielmer (changes requested):

README.md
s-tikhomirov: I agree that linking to particular lines in some source file is a bit ugly. Is there a canonical list of bootstrap nodes in some documentation file? If not, then I'll change the link in question to a specific commit, for the lack of better options.

gabrielmer: I found the address in other readme files such as https://github.com/waku-org/nwaku/blob/c74abe131d00988cc69933525ad2d8c206bf0e63/waku/README.md?plain=1#L197 and https://github.com/waku-org/nwaku/blob/c74abe131d00988cc69933525ad2d8c206bf0e63/docs/operators/how-to/configure-dns-disc.md?plain=1#L27

I don't know if any of them is considered canonical though, so I'm not sure whether it's better to link to one of them or to a specific commit of the code.

What do you think?

PR #2044, opened 2023-09-18
refactor(cbindings): Adding 'librln' to `libwaku` and cleaning waku_example.c

Ivansete-status:
* When adding the dependency with 'librln' to 'libwaku', it is required to remove the dependency with 'confutils' because the 'nim-confutils' module prevents the creation of a dynamic library (libwaku.so.)

* waku_example.c: less code is needed because the Waku Thread attends any Waku event whereas the main thread can have a blocking scanf to retrieve user inputs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -251,7 +251,7 @@ docker-push:
 
 STATIC ?= false
 
-libwaku: | build deps
+libwaku: | build deps librln
 		rm -f build/libwaku*
 ifeq ($(STATIC), true)
 		echo -e $(BUILD_MSG) "build/$@.a" && \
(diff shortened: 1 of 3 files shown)


SionoiS (approved):
LGTM

Why does nim-confutils preclude dyn libs ?

SionoiS:
> > Why does nim-confutils preclude dyn libs ?
> 
> Because it uses `commandLineParams()` which doesn't allow dyn libs:
> 
> https://github.com/status-im/nim-confutils/blob/c8063eb8142aeb6489ded24f3da908bc7f921b7b/confutils.nim#L867
> 
> https://github.com/status-im/Nim/blob/71ba2e7f3c5815d956b1ae0341b0743242b8fec6/lib/pure/os.nim#L3001

Ah! OS dependent functions would be hard I'm guessing. Thanks!

PR #2045, opened 2023-09-18
refactor(cbindings): waku_thread.nim - Using 'ThreadSignalPtr' instead of loop to handle req/resp

Ivansete-status:
# Description
Use of a better approach to synchronize threads. Use of `ThreadSignalPtr` instead of an intense loop.
This PR is motivated by the following comment: https://github.com/waku-org/nwaku/pull/1978#discussion_r1318154604

### DIFF
--- a/library/waku_thread/waku_thread.nim
+++ b/library/waku_thread/waku_thread.nim
@@ -8,9 +8,10 @@ import
 import
   chronicles,
   chronos,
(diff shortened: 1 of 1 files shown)


NagyZoltanPeter (approved):
Pretty nice stuff, thank you!

jm-clius (approved):

PR #2057, opened 2023-09-20
test(waku-filter): Subscribe tests batch (4/4)

AlejandroCabeza:
# Description
Implement fourth (and last) batch of subscribe tests for waku filter, the service to service ones.

# Changes
- [X] Implement waku filter service to service tests.
- [X] Remove legacy waku filter tests.

# Dependencies
https://github.com/waku-org/nwaku/pull/2046


tests/node/test_wakunode_filter.nim
```diff
@@ -0,0 +1,116 @@
...
+  chronos,
+  libp2p/crypto/crypto
+
+import
+  std/[options, tables, sequtils],
+  testutils/unittests,
+  chronos,
+  chronicles,
+  os,
+  libp2p/peerstore
```
gabrielmer, line 18: Is there a reason for having two separate import statements and with some repeated modules?
I think it's better to combine them into one import statement without duplicates

tests/testlib/futures.nim
```diff
@@ -0,0 +1,13 @@
+import
+#   std/sequtils,
+#   std/options,
+#   stew/shims/net as stewNet,
+#   testutils/unittests,
+  chronicles,
+  chronos
+#   libp2p/crypto/crypto
```
gabrielmer, line 8: Here I would delete the lines with unused modules instead of commenting them

tests/waku_filter_v2/test_waku_client.nim
```diff
@@ -18,6 +18,7 @@ import
   ../testlib/wakucore,
   ../testlib/testasync,
   ../testlib/testutils,
+  ../testlib/futures,
```
gabrielmer, line 21: Why is a new import required if no code is being added to this file?

gabrielmer (commented):
Commented some questions, mostly regarding style and not the logic itself.

For the rest I still have some hw to do, need to first get more familiar with the filter protocol. Working on it and hopefully I'll be able to add more feedback on it soon :)

I think it's best to add one or two more reviewers to the PR, ideally including someone that has already worked with filter.

tests/node/test_wakunode_filter.nim
AlejandroCabeza: I followed the style I saw in sibling files. I think the reasoning is to separate `builtin` and `third-party` imports from local ones.

tests/testlib/futures.nim
AlejandroCabeza: Ah, yes. Oversight when testing. Thanks!

tests/waku_filter_v2/test_waku_client.nim
AlejandroCabeza: Because I moved the `newPushHandlerFuture` in `waku_filter_utils` (already imported) into the new `futures` module, as that function will also be used outside waku filter tests.

tests/node/test_wakunode_filter.nim
gabrielmer: Yes, but in this case there's 3 import statements instead of 2 and `chronos`, `chronicles` and some other modules are being imported twice:

https://github.com/waku-org/nwaku/blob/b354235eba77e29911f7a6e641932cba3eeea61c/tests/node/test_wakunode_filter.nim#L3-L31

AlejandroCabeza: Oh, that. It seems my brain just read the first half and assummed the second part. You're completely right, changing that.

tests/node/test_wakunode_filter.nim
```diff
@@ -0,0 +1,109 @@
...
+  var server {.threadvar.}: WakuNode
+  var serverRemotePeerInfo {.threadvar.}: RemotePeerInfo
+  var pubsubTopic {.threadvar.}: PubsubTopicy
+  var contentTopic {.threadvar.}: ContentTopic
+  var contentTopicSeq {.threadvar.}: seq[ContentTopic]
+  var pushHandlerFuture {.threadvar.}: Future[(string, WakuMessage)]
+  var messagePushHandler {.threadvar.}: FilterPushHandler
+  
+  asyncSetup:
+    pushHandlerFuture = newFuture[(string, WakuMessage)]()
```
gabrielmer, line 40: It's a super nit but for consistency I would use here `newPushHandlerFuture()` like in the unsubscribed test case

gabrielmer (approved):
LGTM! Added one comment with a small nitpick - tests looks great 🔥

Thank you!

tests/node/test_wakunode_filter.nim
```diff
@@ -0,0 +1,109 @@
...
+  asyncTest "Full Client Node to Full Service Node Subscription":
+    # When a full client node subscribes to a full service node
+    let subscribeResponse = await client.filterSubscribe(
+      some(pubsubTopic), contentTopicSeq, serverRemotePeerInfo
+    )
+
+    # Then the subscription is successful
+    check:
+      subscribeResponse.isOk()
+      server.wakuFilter.subscriptions.len == 1
```
Ivansete-status, line 78:
```suggestion
    
    assert subscribeResponse.isOk(), $subscribeResponse.error
    check server.wakuFilter.subscriptions.len == 1
```
If possible, I'd apply that elsewhere.

Ivansete-status (approved):
Great PR! LGTM! Just added a tiny comment

tests/node/test_wakunode_filter.nim
AlejandroCabeza: Indeed! Must've slipped my mind. I'm planning to do a "standardising" PR after I finish the current one I'm working on to mend all places I skipped.

Issue #1941, opened 2023-08-24
feat: Service peer selection on specific shards


chair28980:
Currently a service peer (e.g. a store node) is either provided by the application or it expects the peer manager to suggest a suitable service peer (selectPeer()). This task focuses on the latter: peer manager should provide ability to find service peers for a specific shard, based on the content topic that the application is interested in (e.g. an API call to selectPeer(Protocol, ContentTopic)), or kick off an ad-hoc discovery process until a suitable service peer is found. Client APIs (for filter, lightpush, peer-exchange) should automatically make use of this mechanism if the application does not explicitly set the desired service peer.

Priority: critical for launch

SionoiS:
I've looked at the code and it seams easy to do.

The hard part is the architecture doesn't make it easy to access `waku_dscv5`.

Handlers would call `selectPeer()` then if none is found launch some discovery process.

SionoiS:
Architecturally is seams that calling discv5 to find peers with specific shard and caps should be done at the app level.

Protocols clients impl. should not know about the discovery process.

SionoiS:
Extracting all the protocols from `waku_node` is harder than I thought. The external API sit at the node level and would have to change to app level.....

SionoiS:
**Weekly Update**

- _achieved_: peer manager can filter peer by shard, filter discv5 bootstrap nodes by shard, external APIs moved out of node folder
- _next_: refactor APIs handlers to discover peers if none is found in peer manager with the required capability

SionoiS:
**Weekly Update**

- _achieved_: newly refactored STORE REST API handler that trigger discv5 peer search when needed. 
- _next_: refactor other APIs

SionoiS:
**Weekly Update**

- _achieved_: Filter, Store, Light push REST APIs discovery handler (a rework of the previous solution)

SionoiS:
**Weekly Update**

- _achieved_: REST APIs discovery handlers PR merged

SionoiS:
Done

PR #2063, opened 2023-09-21
feat: added optional filtering by shard to selectPeer()

SionoiS:
# Description
Tiny PR adding an optional shard filter when selecting a peer via peer manager.

Tracking #1941

### DIFF
--- a/waku/node/peer_manager/peer_manager.nim
+++ b/waku/node/peer_manager/peer_manager.nim
@@ -16,6 +16,7 @@ import
   ../../common/nimchronos,
   ../../waku_core,
   ../../waku_relay,
+  ../../waku_enr/sharding,
   ./peer_store/peer_storage,
   ./waku_peer_store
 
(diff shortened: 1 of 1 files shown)


waku/node/peer_manager/peer_manager.nim
```diff
@@ -631,11 +632,14 @@ proc prunePeerStore*(pm: PeerManager) =
                                        capacity = capacity,
                                        pruned = pruned
 
-proc selectPeer*(pm: PeerManager, proto: string): Option[RemotePeerInfo] =
+proc selectPeer*(pm: PeerManager, proto: string, shard: Option[PubsubTopic] = none(PubsubTopic)): Option[RemotePeerInfo] =
```
jm-clius, line 635: In the spirit of `shard` being the number encoded into the `pubsubTopic`, could we call this something like `shardedPubsubTopic`?

jm-clius (approved):
LGTM for now, but wondering if we'll always have the ENR for interesting peers available? The idea of some metadata exchange protocol, for example, would allow us to discover the shards for any peer even outside of ENR encoding. In other words, we may want to also index against the shard, cluster in the peerStore so that it's not a separate filter loop.

SionoiS:
> LGTM for now, but wondering if we'll always have the ENR for interesting peers available? The idea of some metadata exchange protocol, for example, would allow us to discover the shards for any peer even outside of ENR encoding. In other words, we may want to also index against the shard, cluster in the peerStore so that it's not a separate filter loop.

I agree that sharding is not well integrated into peer management yet. I don't think that ENRs will be missing in many cases but we def should have something in place for those.

waku/node/peer_manager/peer_manager.nim
SionoiS: > In the spirit of `shard` being the number encoded into the `pubsubTopic`, 

Is it? I always thought of shards as pubsub topic that conform to static or autoshard specs. The numbers are named cluster and index.

> could we call this something like `shardedPubsubTopic`?
The term `shardedPubsubTopic` has never been used before. I don't like adding more terminology.

PR #2068, opened 2023-09-21
feat(networkmonitor): add ping latencies, optimize reconnections

vpavlin:
# Description

This PR leverages `ping` protocol to measure latency in between the network monitor instance and discovered nodes.

It records the latency as a
* histogram metric
* last measured latency per node
* average latency per node with sliding window of 10 

It also adds retry limit for failed connections and adds a cliff to repeatedly discovered nodes to only reconnect to them after 60s

# Changes

- [ ] use `int64` for timestamps instead of strings
- [ ] add PingDuration (i.e. latency) fields
- [ ] add retries (how many times we failed to connect) and discovered (how many times we discovered the node) field
- [ ] only dial and ping the node if `ReconnectTime` passed
- [ ] only dial and ping unless we are over the `MaxConnectionRetries`
- [ ] use PeerId instead of public key
- [ ] add `ping()` proc to wrap dialling and executing ping protocol
- [ ] set default log level to INFO

### DIFF
--- a/apps/networkmonitor/networkmonitor.nim
+++ b/apps/networkmonitor/networkmonitor.nim
@@ -16,6 +16,7 @@ import
   eth/p2p/discoveryv5/enr,
   libp2p/crypto/crypto,
   libp2p/nameresolving/dnsresolver,
+  libp2p/protocols/ping,
   metrics,
   metrics/chronos_httpserver,
   presto/[route, server, client]
@@ -33,6 +34,10 @@ import
 logScope:
   topics = "networkmonitor"
 
+const ReconnectTime = 60
+const MaxConnectionRetries = 10
+const AvgPingWindow = 10.0
+
 proc setDiscoveredPeersCapabilities(
   routingTableNodes: seq[Node]) =
   for capability in @[Relay, Store, Filter, Lightpush]:
@@ -47,11 +52,13 @@ proc setConnectedPeersMetrics(discoveredNodes: seq[Node],
                               restClient: RestClientRef,
                               allPeers: CustomPeersTableRef) {.async.} =
 
-  let currentTime = $getTime()
+  let currentTime = getTime().toUnix()
 
   # Protocols and agent string and its count
   var allProtocols: Table[string, int]
   var allAgentStrings: Table[string, int]
+
+  var newPeers = 0
 
   # iterate all newly discovered nodes
   for discNode in discoveredNodes:
@@ -65,15 +72,25 @@ proc setConnectedPeersMetrics(discoveredNodes: seq[Node],
       warn "could not get secp256k1 key", typedRecord=typedRecord.get()
       continue
 
+    let peerRes = toRemotePeerInfo(discNode.record)
+
+    let peerInfo = peerRes.valueOr():
+      warn "error converting record to remote peer info", record=discNode.record
+      continue
+
     # create new entry if new peerId found
-    let peerId = secp256k1.get().toHex()
+    let peerId = $peerInfo.peerId
     let customPeerInfo = CustomPeerInfo(peerId: peerId)
     if not allPeers.hasKey(peerId):
       allPeers[peerId] = customPeerInfo
+      newPeers += 1
+    else:
+      info "already seen", peerId=peerId
 
     allPeers[peerId].lastTimeDiscovered = currentTime
     allPeers[peerId].enr = discNode.record.toURI()
     allPeers[peerId].enrCapabilities = discNode.record.getCapabilities().mapIt($it)
+    allPeers[peerId].discovered += 1
 
     if not typedRecord.get().ip.isSome():
       warn "ip field is not set", record=typedRecord.get()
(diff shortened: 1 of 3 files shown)


apps/networkmonitor/networkmonitor.nim
```diff
@@ -82,43 +99,68 @@ proc setConnectedPeersMetrics(discoveredNodes: seq[Node],
...
+        except CatchableError:
+          var msg = getCurrentExceptionMsg()
+          if msg == "Future operation cancelled!":
+            msg = "timedout"
+          warn "failed to ping the peer", peer=peerInfo, err=msg
+
+          allPeers[peerId].connError = msg
+          return err("could not ping peer: " & msg)
+
+      let timedOut = not await ping().withTimeout(timeout)
```
alrevuelta, line 124: dont fully understand `proc ping(): Future[Result[void, string]] {.async, gcsafe.} =`

and this `let timedOut = not await ping().withTimeout(timeout)`


isnt something like this better to handle the timeout?
```
if not (waitFornode.libp2pPing.ping(conn)).withTimeout(timeout)):
   xxx
```

and if it errors before timeout, that is handled in the `except`?

apps/networkmonitor/networkmonitor.nim
```diff
@@ -82,43 +99,68 @@ proc setConnectedPeersMetrics(discoveredNodes: seq[Node],
...
+        continue
+
+      info "successfully pinged peer", peer=peerInfo, duration=pingDelay.millis
+      peer_ping.observe(pingDelay.millis)
+
+      if allPeers[peerId].avgPingDuration == 0.millis:
+        allPeers[peerId].avgPingDuration = pingDelay
+
+      # TODO: check why the calculation ends up losing precision
+      allPeers[peerId].avgPingDuration = int64((float64(allPeers[peerId].avgPingDuration.millis) * (AvgPingWindow - 1.0) + float64(pingDelay.millis)) / AvgPingWindow).millis
```
alrevuelta, line 137: is this formula correct?
think its missing a `/AvgPingWindow`
old average * (n-1)/n + new value /n

something like this (unsure about the castings)
```
allPeers[peerId].avgPingDuration = int64((float64(allPeers[peerId].avgPingDuration.millis) * (AvgPingWindow - 1.0)/AvgPingWindow + float64(pingDelay.millis)) / AvgPingWindow).millis
```

alrevuelta (commented):
thanks! left couple of comments.

curious about the how the histogram looks like in the current network.

apps/networkmonitor/networkmonitor.nim
vpavlin: Hmm.. You produced exactly the same formula if I see correctly:D

vpavlin: No, that would not be correct IMO - you need to make avg represent 9 values in the window + new value, what you have there is make avg*0.9 + new value and then divide by 10 (check the parentheses)

I believe my formula is correct, just some weird issue with casting/rounding

apps/networkmonitor/networkmonitor.nim
vpavlin: Maybe:D This was the first thing that came to mind when trying to overcome the weirdness of withTimeout which would not return any result from ping.

I'll rewrite it

apps/networkmonitor/networkmonitor.nim
```diff
@@ -82,43 +99,68 @@ proc setConnectedPeersMetrics(discoveredNodes: seq[Node],
...
-    if not allAgentStrings.hasKey(nodeUserAgent):
-      allAgentStrings[nodeUserAgent] = 0
-    allAgentStrings[nodeUserAgent] += 1
+      # store available user-agents in the network
+      if not allAgentStrings.hasKey(nodeUserAgent):
+        allAgentStrings[nodeUserAgent] = 0
+      allAgentStrings[nodeUserAgent] += 1
 
-    debug "connected to peer", peer=allPeers[customPeerInfo.peerId]
+      debug "connected to peer", peer=allPeers[customPeerInfo.peerId]
```
jm-clius, line 161: Afaict this extra indentation will now result in these being only set if the ping condition is met. Perhaps intentional?

apps/networkmonitor/networkmonitor.nim
jm-clius: Feels to me that this inner `ping()` function may be unnecessary and you can just dial and await the `libp2pPing.ping()` directly? May be missing something here.

apps/networkmonitor/networkmonitor.nim
```diff
@@ -82,43 +99,68 @@ proc setConnectedPeersMetrics(discoveredNodes: seq[Node],
     let ip = $typedRecord.get().ip.get().join(".")
     allPeers[peerId].ip = ip
 
-    let peer = toRemotePeerInfo(discNode.record)
-    if not peer.isOk():
-      warn "error converting record to remote peer info", record=discNode.record
-      continue
+    # try to ping the peer
+    if getTime().toUnix() >= allPeers[peerId].lastTimeConnected + ReconnectTime and allPeers[peerId].retries < MaxConnectionRetries:
```
jm-clius, line 103: I'd suggest moving the logic under this out into a separate function. `setConnectedPeersMetrics` is becoming diffult to follow and maintain.
Something like:
```
if dueForPing(allPeers[PeerId]):
  let pingDelay = ping(allPeers[PeerId]).onError ...
```

My not be as simple as that, but I guess we should start separating concerns here as much as possible.

apps/networkmonitor/networkmonitor.nim
```diff
@@ -65,15 +72,25 @@ proc setConnectedPeersMetrics(discoveredNodes: seq[Node],
...
     if not allPeers.hasKey(peerId):
       allPeers[peerId] = customPeerInfo
+      newPeers += 1
+    else:
+      info "already seen", peerId=peerId
 
     allPeers[peerId].lastTimeDiscovered = currentTime
     allPeers[peerId].enr = discNode.record.toURI()
     allPeers[peerId].enrCapabilities = discNode.record.getCapabilities().mapIt($it)
+    allPeers[peerId].discovered += 1
```
jm-clius, line 93: Perhaps time to assign `allPeers[peerId]` here to a variable? Just to improve readability rather than repeated `allPeers[peerId]` throughout the rest of the proc.

jm-clius (approved):
Approving as to not be a bottleneck, but made some comments below. Direction makes sense to me :)

apps/networkmonitor/networkmonitor.nim
alrevuelta: missread the parenthesis in your formula. yours and mine are the same.

> what you have there is make avg*0.9 + new value and then divide by 10 

nope, since division has priority over sum, 
`old_average * (n-1)/n + new_value /n`

is equivalent to

``old_average * (n-1)/n + (new_value /n)``

alrevuelta (approved):

apps/networkmonitor/networkmonitor.nim
vpavlin: Ok, so looking at the code again, my problem was that I need the `Connection` from `dial()` to be able to call the `ping`, but without `withTimeout` on the `dial` it will hang for some nodes. But then if I add `withTimeout` I don't see a way to extract the returned `Connection`.

This is why I wrapped it in a `proc` and just set the variables directly from inside of the inner proc. If there is another (better) way to do it, I am happy to refactor, but I don't see

cc @jm-clius

vpavlin:
@jm-clius I do agree with your comments and I am planning to address them, but I did not want to make this particular PR more complex than necessary. So I'd like to merge the addition of `ping` and then come up with a proper refactoring PR

PR #2069, opened 2023-09-21
refactor: Moved external APIs out of node

SionoiS:
# Description

The JSONRPC and REST APIs are no longer part of `node`. I created a `waku_api` folder for them.

This would be the first step in a bigger refactoring effort. Next step would be to move peer manager out of `node` as well.

Tracking #1941


waku/waku_api/rest/filter/handlers.nim
```diff
@@ -14,13 +14,14 @@ import
...
 import
   ../../../waku_core,
+  ../../../waku_node,
+  ../../../node/peer_manager,
   ../../../waku_filter,
   ../../../waku_filter_v2,
   ../../../waku_filter_v2/client as filter_protocol_client,
-  ../../../waku_filter_v2/common as filter_protocol_type,
+  ../../../waku_filter_v2/common as filter_protocol_type
+import
```
Ivansete-status, line 23: Why this line?

Ivansete-status (approved):
LGTM! Thanks for it! :100: 

Just a comment: I don't see the need to move `peer_manager` out of `node`. Could elaborate a bit more on why this is interesting plz?

jm-clius (approved):
Seems to work :)

waku/waku_api/rest/filter/handlers.nim
SionoiS: Some left overs. Good catch!

SionoiS:
> LGTM! Thanks for it! 💯
> 
> Just a comment: I don't see the need to move `peer_manager` out of `node`. Could elaborate a bit more on why this is interesting plz?

I haven't looked into it yet but either reorganize some files only OR move the logic to the app layer.

As for why, managing peer is always app specific. A light node that only use light push and filter is not going to manage peer the same as a full node with relay enabled for multiple shards.

PR #2070, opened 2023-09-22
chore: Makefile - making the 'libwaku.so' to be built with 'make all'

Ivansete-status:
# Description
This is to make sure the upcoming refactoring tasks ensure that the `libwaku` still builds properly.


SionoiS (approved):
yep would prevent breaking things.

PR #2072, opened 2023-09-22
fix: container-image.yml: Adding -d:postgres to support Postgres

Ivansete-status:
## Description

Making the generated image in the PRs to support Postgres.


Ivansete-status:
I close this because is duplicated

PR #2074, opened 2023-09-25
feat(wakucanary): add latency measurement using ping protocol

vpavlin:
# Description

Since this is very similar to #2068, I thought it might be interesting and useful to have the latency measured by wakucanary. 

Set it on by default with a switch `--ping=false` to disable it

```
INF 2023-09-25 12:53:31.315+02:00 Peer response time (ms)                    tid=2064003 file=wakucanary.nim:110 peerId=16U*ZsXy69 ping=59
```

# Changes

- [ ] add `--ping` option
- [ ] mount `ping` protocol
- [ ] measure latency is `--ping=true`

### DIFF
--- a/apps/wakucanary/wakucanary.nim
+++ b/apps/wakucanary/wakucanary.nim
@@ -47,7 +47,7 @@ type
 
     logLevel* {.
       desc: "Sets the log level",
-      defaultValue: LogLevel.DEBUG,
+      defaultValue: LogLevel.INFO,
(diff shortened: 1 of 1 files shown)


alrevuelta (approved):
lgtm! nice idea.

jakubgs (approved):
I like it.

vpavlin:
Merging even though js-waku test not passing, but that is for sure unrelated

PR #2079, opened 2023-09-26
chore(cbindings): Adding cpp example that integrates the 'libwaku'

Ivansete-status:
# Description
In this example, we can see how to integrate a `nwaku` node from within a `cpp` code. The code is very similar to the `C` version. The interesting point is the usage of cpp-lambdas and how they can seamlessly integrate with the execution context.

e.g.
```
    std::string defautPubsubTopic;
    WAKU_CALL(waku_default_pubsub_topic(cify([&](const char* msg, size_t len) {
        defautPubsubTopic = msg;
    })));
```

# Changes

- [x] Adding `cpp`-`nwaku` example.

### DIFF
--- a/examples/cpp/waku.cpp
+++ b/examples/cpp/waku.cpp
@@ -300,7 +300,7 @@ int main(int argc, char** argv) {
         defaultPubsubTopic = msg;
     })));
 
-    std::cout << "Default pubsub topic: " << defautPubsubTopic << std::endl;
+    std::cout << "Default pubsub topic: " << defaultPubsubTopic << std::endl;
 
     WAKU_CALL(waku_version(cify([&](const char* msg, size_t len) {
         std::cout << "Git Version: " << msg << std::endl;
--- a/examples/cpp/waku.cpp
+++ b/examples/cpp/waku.cpp
@@ -98,14 +98,6 @@ void show_main_menu() {
     printf("\t3.) Publish a message\n");
 }
 
-void set_scanf_to_not_block() {
-    fcntl(0, F_SETFL, fcntl(0, F_GETFL) | O_NONBLOCK);
-}
-
-void set_scanf_to_block() {
-    fcntl(0, F_SETFL, fcntl(0, F_GETFL) ^ O_NONBLOCK);
-}
-
 void handle_user_input() {
     char cmd[1024];
     memset(cmd, 0, 1024);
@@ -122,7 +114,6 @@ void handle_user_input() {
     case SUBSCRIBE_TOPIC_MENU:
     {
         printf("Indicate the Pubsubtopic to subscribe:\n");
-        set_scanf_to_block();
         char pubsubTopic[128];
         scanf("%127s", pubsubTopic);
         // if (!waku_relay_subscribe(pubsubTopic, &mResp)) {
@@ -130,7 +121,6 @@ void handle_user_input() {
         // }
         // printf("Waku Relay subscription response: %s\n", mResp->data);
 
-        set_scanf_to_not_block();
         show_main_menu();
     }
     break;
@@ -138,19 +128,16 @@ void handle_user_input() {
     case CONNECT_TO_OTHER_NODE_MENU:
         printf("Connecting to a node. Please indicate the peer Multiaddress:\n");
         printf("e.g.: /ip4/127.0.0.1/tcp/60001/p2p/16Uiu2HAmVFXtAfSj4EiR7mL2KvL4EE2wztuQgUSBoj2Jx2KeXFLN\n");
-        set_scanf_to_block();
         char peerAddr[512];
         scanf("%511s", peerAddr);
         // if (!waku_connect(peerAddr, 10000 /* timeoutMs */, &mResp)) {
         //     printf("Couldn't connect to the remote peer: %s\n", mResp->data);
         // }
-        set_scanf_to_not_block();
         show_main_menu();
     break;
 
     case PUBLISH_MESSAGE_MENU:
     {
-        set_scanf_to_block();
         printf("Indicate the Pubsubtopic:\n");
         char pubsubTopic[128];
         scanf("%127s", pubsubTopic);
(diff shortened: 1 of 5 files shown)


SionoiS (approved):
I think I understand what's happening (unfamiliar with C++). Having another example is great.

Thank you!

NagyZoltanPeter:
Before leaving review comments, I would have some questions or maybe open a little discussion on it due I'm embarrassed a bit about it.

What is the exact aim of such example over the current ansi C example?
What is the minimum C++ standard we targeting (of coz its ok to still use C++98, though a bit uncommon)?

Lets put in other way around. If I would like to show some c++ example, maybe I would like to provide a c++ wrapper on the c-bindings first. Also maybe include base64 encode/decode into the library (just to be confirm fully our requirements).

You know I'm quite ok with such usage, but I would like to give a mean or value over the C version.
I hope you, @Ivansete-status will get the point, I would not like to be destructive on it surely and I know it's just a helper for the crowd, whom may use it differently as their taste.

gabrielmer:
Really like the use of lambdas for the callbacks!

To use them we need at least C++11. Even though they are not strictly necessary I think C++11 is a good minimum standard to target - even if we didn't use lambdas, it's one of the most widely used standards, and the ones before it are now pretty rarely used.

I do agree that we should either add the base64 encode/decode functions to the library or at least import those functions in the example from some existing external library.

I think that implementing the functions in the example may add noise and take attention from more important parts, and I don't know if we should encourage users to implement their own encoding/decoding functions instead of taking one from a library

Ivansete-status:
Thanks for the comments @NagyZoltanPeter @gabrielmer :fire: !
This is a simple example of how to use `nwaku` from cpp. Of course, it can be optimized. Nevertheless, cpp is not our priority right now and this is just a simple example.

Re the `base64` stuff, I think we have it separated in the `C` example.
Re the C++ I think C++11 is good as a minimum.

I wouldn't worry very much about this `cpp` example for now.

This is the roadmap I'd like to follow:
- Make the `libwaku` implementation more solid and work properly in Rust, NodeJs, Python. For that, a simple `relay` example is enough.
- Once we have `libwaku` super solid and thread-safe, create separate `waku` repos for each language, and from there, make each of them use the `nwaku` as a submodule and therefore compile and link the `libwaku` library. In those repos we will be able to invest our efforts to have better performance, test, etc.

Thanks again for the comments :)

gabrielmer (approved):
Sounds great! Thanks so much!

examples/cpp/waku.cpp
```diff
@@ -0,0 +1,336 @@
...
+#include <vector>
+#include <iostream>
+
+#include <sys/types.h>
+#include <unistd.h>
+#include <sys/syscall.h>
+
+#include "../../library/libwaku.h"
+
+#define WAKU_CALL(call)                                                        \
```
NagyZoltanPeter, line 18: If it can be done to outcast the return value to the caller would be nice.
I'm thinking of a templatized function wrapper, but definitely that is more complicated.

examples/cpp/waku.cpp
```diff
@@ -0,0 +1,336 @@
...
+		ret += 3 - (inlen % 3);
+	ret /= 3;
+	ret *= 4;
+
+	return ret;
+}
+
+const char b64chars[] = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/";
+
+void b64_encode(char* in, size_t len, std::vector<char>& out)
```
NagyZoltanPeter, line 99: If C++11 we should simple return with std::vector<char>. Move semantics and temporary object optimization will solve it for free.

examples/cpp/waku.cpp
```diff
@@ -0,0 +1,336 @@
...
+
+        printf("Type the message tp publish:\n");
+        char msg[1024];
+        scanf("%1023s", msg);
+
+        char jsonWakuMsg[1024];
+        std::vector<char> msgPayload;
+        b64_encode(msg, strlen(msg), msgPayload);
+
+        // waku_content_topic("appName",
```
NagyZoltanPeter, line 217: Is it something intentionally left commented our not working?

examples/cpp/waku.cpp
```diff
@@ -0,0 +1,336 @@
...
+    printf("Receiving message %s\n", msg);
+}
+
+void handle_error(const char* msg, size_t len) {
+    printf("Error: %s\n", msg);
+    exit(1);
+}
+
+template <class F>
+auto cify(F&& f) {
```
NagyZoltanPeter, line 259: I'm not pretty sure that wrapping around lambdas is necessary. Maybe enough to simple give lamba in place to the call of C function.

examples/cpp/waku.cpp
```diff
@@ -0,0 +1,336 @@
...
+                                    cfgNode.key,
+                                    cfgNode.relay ? "true":"false");
+
+    WAKU_CALL(waku_new(jsonConfig, cify([](const char* msg, size_t len) {
+        std::cout << "Error: " << msg << std::endl;
+        exit(1);
+    })));
+
+    // example on how to retrieve a value from the `libwaku` callback.
+    std::string defautPubsubTopic;
```
NagyZoltanPeter, line 298:
```suggestion
    std::string defaultPubsubTopic;
```

NagyZoltanPeter (approved):
Found some tiny stuff but all OK, main target is already discussed!
Thanks for this nice work!

jm-clius (approved):
Great to have this too!

examples/cpp/waku.cpp
Ivansete-status: Good catch! I leave those comments to indicate which `libwaku` should be invoked. If we uncomment they might not work because the signatures are changing due to making the `libwaku` thread-safe. We should revisit that once we have time to tackle it.

examples/cpp/waku.cpp
Ivansete-status: Good point! Afaik, this is needed. We'll revisit that point in the future.

Ivansete-status:
Thanks for the comments!
I'll merge it now and leave it for the future to apply all the recommendations needed to have a proper `cpp` example.

PR #2080, opened 2023-09-26
chore(networkmonitor): refactor setConnectedPeersMetrics, make it partially concurrent, add version

vpavlin:
# Description

This PR refactors `setConnectedPeersMetrics` proc a bit

* splits out the `shouldReconnect` check
* splits out the reconnection (i.e. dial and ping) to separate proc
* uses seq of futures to make the dials to newly discovered peers concurrent
* adds `ResetRetriesAfter` time (in seconds) to retry peers we failed to connect (if they are re-discovered)

It also adds `--version` (cc @jakubgs)

# Changes

- [x] add `shouldReconnect`
- [x] add `analyzePeer`
- [x] add concurrency to peer analysis
- [x] add `--version`

## Issue

partially based on comments from #2068

### DIFF
--- a/apps/networkmonitor/networkmonitor.nim
+++ b/apps/networkmonitor/networkmonitor.nim
@@ -35,8 +35,11 @@ logScope:
   topics = "networkmonitor"
 
 const ReconnectTime = 60
-const MaxConnectionRetries = 10
+const MaxConnectionRetries = 5
+const ResetRetriesAfter = 1200
 const AvgPingWindow = 10.0
+
+const git_version* {.strdefine.} = "n/a"
 
 proc setDiscoveredPeersCapabilities(
   routingTableNodes: seq[Node]) =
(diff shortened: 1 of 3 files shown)


apps/networkmonitor/networkmonitor.nim
```diff
@@ -80,103 +134,109 @@ proc setConnectedPeersMetrics(discoveredNodes: seq[Node],
...
+    let nodeUserAgent = lp2pPeerStore[AgentBook][peerId]
+    customPeerInfo.userAgent = nodeUserAgent
 
   info "number of newly discovered peers", amount=newPeers
   # inform the total connections that we did in this round
-  let nOfOkConnections = allProtocols.len()
-  info "number of successful connections", amount=nOfOkConnections
+  info "number of successful connections", amount=successfulConnections
+
+proc updateMetrics(allPeersRef: CustomPeersTableRef) {.gcsafe, raises: [KeyError].} =
```
alrevuelta, line 194: nitpick, use `Result` instead of raising except?

apps/networkmonitor/networkmonitor_metrics.nim
```diff
@@ -25,22 +25,30 @@ logScope:
...
     "Number of peers with each user agent",
     labels = ["user_agent"]
 
-declarePublicHistogram peer_ping,
+declarePublicHistogram networkmonitor_peer_ping,
     "Histogram tracking ping durations for discovered peers",
     buckets = [100.0, 200.0, 300.0, 400.0, 500.0, 600.0, 700.0, 800.0, 900.0, 1000.0, 2000.0, Inf]
 
+declarePublicGauge networkmonitor_peer_count,
+    "Number of discovered peers",
```
alrevuelta, line 45: is this "discovered"? based on the code i think its something like "attempted to connect"?

alrevuelta (approved):
lgtm! two minor non blocking things

apps/networkmonitor/networkmonitor_metrics.nim
vpavlin: So it track number of all discovered peers, but distinguishes between successful and unsuccessful connections based on `connected` label

apps/networkmonitor/networkmonitor.nim
alrevuelta: yeap, the thing would be to catch that and return Result[void] error. No big deal since this is not part of the core codebase, but usually what we do in nwaku.

vpavlin: Ok, I can add that, just still don't understad how it could throw that - feels like false positive

jm-clius (approved):
Thanks! Definitely better. :)

PR #2083, opened 2023-09-26
fix: return erring response if Lightpush content topic is empty

### LINKED ISSUE #1641: bug: light push response is empty when protobuf decode fails
### Problem

When js-waku client tries to use light push with a waku message that has no content topic, nwaku sends an empty response back.

Expected: an error in the light push response to point towards decoding issue.

### To reproduce
1. Start nwaku with light push
2. Send a light push rpc request that contains a missing protobuf field such as content topic
3. get an empty rpc light push response

### Expected behavior

rpc light push response contains an error

### Screenshots/logs

[nwaku_Waku_Light_Push_[node_only]_Push_successfully.log](https://github.com/waku-org/nwaku/files/11137022/nwaku_Waku_Light_Push_.node_only._Push_successfully.log)

```
ERR 2023-04-03 20:00:14.214+10:00 failed to decode rpc                       topics="waku lightpush" tid=147839 file=protocol.nim:43
```

### nwaku version/commit hash
v0.16.0

s-tikhomirov:
# Description
The Lightpush PFC prohibits empty content topics. The implementation did not reflect this. This PR fixes this: the server now checks if the content topic is empty, and if it is, sends an erring response to the client (no interaction with the P2P network occurs). A new test is added to test this behavior.

Note: this PR only considers the "empty content topic" erring case. There are other erring cases (e.g., decoding error), for which the server does not return a response. I think these are to be addressed in a separate issue / branch.

P.S. I wanted to put a link to the RFC that says that content topic must not be empty, but cannot find it (apparently, it's not [19/WAKU2-LIGHTPUSH](https://rfc.vac.dev/spec/19/)). Can anyone point to the spec that specifies this restriction?

# Changes

- [x] in Lightpush protocol: check if the content topic is empty and return an erring response if so
- [x] in Lightpush tests: add a test that tests this behavior (existing tests are not modified).

## How to test
```
./env.sh bash
nim c -r ./tests/test_waku_lightpush.nim
```

## Issue
closes #1641

### DIFF
--- a/tests/test_waku_lightpush.nim
+++ b/tests/test_waku_lightpush.nim
@@ -1,6 +1,8 @@
 {.used.}
 
 import
+  std/options,
+  std/strscans,
   testutils/unittests,
   chronicles,
   chronos,
@@ -10,9 +12,10 @@ import
   ../../waku/waku_core,
   ../../waku/waku_lightpush,
   ../../waku/waku_lightpush/client,
+  ../../waku/waku_lightpush/protocol_metrics,
+  ../../waku/waku_lightpush/rpc,
   ./testlib/common,
   ./testlib/wakucore
-
 
 proc newTestWakuLightpushNode(switch: Switch, handler: PushMessageHandler): Future[WakuLightPush] {.async.} =
   let
@@ -116,3 +119,35 @@ suite "Waku Lightpush":
 
     ## Cleanup
     await allFutures(clientSwitch.stop(), serverSwitch.stop())
+
+  asyncTest "incorrectly encoded request should return an erring response":
+    ## Setup
+    let
+      serverSwitch = newTestSwitch()
+      handler = proc(peer: PeerId, pubsubTopic: PubsubTopic, message: WakuMessage): Future[WakuLightPushResult[void]] {.async.} =
+        ## this handler will never be called: request must fail earlier
+        return ok()
+      server = await newTestWakuLightpushNode(serverSwitch, handler)
+
+    ## Given
+    let
+      fakeBuffer = @[byte(42)]
+      fakePeerId = PeerId.init(PrivateKey.random(ECDSA, (newRng())[]).tryGet()).tryGet()
+    
+    ## When
+    let
+      pushRpcResponse = await server.handleRequest(fakePeerId, fakeBuffer)
+      requestId = pushRpcResponse.requestId
+    
+    ## Then
+    check:
+      requestId == ""
+      pushRpcResponse.response.isSome()
+    
+    let resp = pushRpcResponse.response.get()
+
+    check:
+      resp.isSuccess == false
+      resp.info.isSome()
+      ## the error message should start with decodeRpcFailure
+      scanf(resp.info.get(), decodeRpcFailure)
--- a/waku/waku_lightpush/protocol.nim
+++ b/waku/waku_lightpush/protocol.nim
@@ -35,47 +35,46 @@ type
     peerManager*: PeerManager
     pushHandler*: PushMessageHandler
 
-proc initProtocolHandler*(wl: WakuLightPush) =
-  proc handle(conn: Connection, proto: string) {.async, gcsafe, closure.} =
-    let buffer = await conn.readLp(MaxRpcSize.int)
-    let reqDecodeRes = PushRPC.decode(buffer)
-    if reqDecodeRes.isErr():
-      error "failed to decode rpc"
-      waku_lightpush_errors.inc(labelValues = [decodeRpcFailure])
-      return
+proc handleRequest*(wl: WakuLightPush, peerId: PeerId, buffer: seq[byte]): Future[PushRPC] {.async.} = 
+  let reqDecodeRes = PushRPC.decode(buffer)
+  var
+      isSuccess = false
+      pushResponseInfo = ""
+      requestId = ""
 
-    let req = reqDecodeRes.get()
-    if req.request.isNone():
-      error "invalid lightpush rpc received", error=emptyRequestBodyFailure
-      waku_lightpush_errors.inc(labelValues = [emptyRequestBodyFailure])
-      return
+  if reqDecodeRes.isErr():
+    pushResponseInfo = decodeRpcFailure & ": " & $reqDecodeRes.error
+  elif reqDecodeRes.get().request.isNone():
+    pushResponseInfo = emptyRequestBodyFailure
+  else:
+    let pushRpcRequest = reqDecodeRes.get();
 
-    waku_lightpush_messages.inc(labelValues = ["PushRequest"])
+    requestId = pushRpcRequest.requestId
+
     let
-      pubSubTopic = req.request.get().pubSubTopic
-      message = req.request.get().message
-    debug "push request", peerId=conn.peerId, requestId=req.requestId, pubsubTopic=pubsubTopic
+      request = pushRpcRequest.request
 
-    var response: PushResponse
-    var handleRes: WakuLightPushResult[void]
-    try:
-      handleRes = await wl.pushHandler(conn.peerId, pubsubTopic, message)
-    except Exception:
-      response = PushResponse(is_success: false, info: some(getCurrentExceptionMsg()))
-      waku_lightpush_errors.inc(labelValues = [messagePushFailure])
-      error "pushed message handling failed", error= getCurrentExceptionMsg()
+      pubSubTopic = request.get().pubSubTopic
+      message = request.get().message
+    waku_lightpush_messages.inc(labelValues = ["PushRequest"])
+    debug "push request", peerId=peerId, requestId=requestId, pubsubTopic=pubsubTopic
+    
+    let handleRes = await wl.pushHandler(peerId, pubsubTopic, message)
+    isSuccess = handleRes.isOk()
+    pushResponseInfo = (if isSuccess: "OK" else: handleRes.error)
 
+  if not isSuccess:
+    waku_lightpush_errors.inc(labelValues = [pushResponseInfo])
+    error "failed to push message", error=pushResponseInfo
+  let response = PushResponse(isSuccess: isSuccess, info: some(pushResponseInfo))
+  let rpc = PushRPC(requestId: requestId, response: some(response))
+  return rpc
 
-    if handleRes.isOk():
-      response = PushResponse(is_success: true, info: some("OK"))
-    else:
-      response = PushResponse(is_success: false, info: some(handleRes.error))
-      waku_lightpush_errors.inc(labelValues = [messagePushFailure])
-      error "pushed message handling failed", error=handleRes.error
-
-    let rpc = PushRPC(requestId: req.requestId, response: some(response))
+proc initProtocolHandler*(wl: WakuLightPush) =
+  proc handle(conn: Connection, proto: string) {.async.} =
+    let buffer = await conn.readLp(MaxRpcSize.int)
+    let rpc = await handleRequest(wl, conn.peerId, buffer)
     await conn.writeLp(rpc.encode().buffer)
-
   wl.handler = handle
   wl.codec = WakuLightPushCodec
 


SionoiS (approved):
LGTM

Thank you!

s-tikhomirov:
I see that two CI tests are failing...

```
2023-09-26T15:23:36.7503236Z   1) Waku Light Push [node only] - custom pubsub topic
2023-09-26T15:23:36.7503689Z        Push message:
2023-09-26T15:23:36.7504342Z      TypeError: Cannot read properties of undefined (reading 'toString')
2023-09-26T15:23:36.7505745Z       at Context.<anonymous> (file:///home/runner/work/nwaku/nwaku/packages/tests/tests/light-push/custom_pubsub.node.spec.ts:43:39)
2023-09-26T15:23:36.7510507Z       at processTicksAndRejections (node:internal/process/task_queues:95:5)
2023-09-26T15:23:36.7511700Z 
2023-09-26T15:23:36.7512046Z   2) Waku Light Push [node only]
2023-09-26T15:23:36.7512614Z        "before each" hook for "Push message with short payload":
2023-09-26T15:23:36.7513932Z      Error: Timeout of 15000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/home/runner/work/nwaku/nwaku/packages/tests/tests/light-push/index.spec.ts)
2023-09-26T15:23:36.7514733Z       at listOnTimeout (node:internal/timers:569:17)
2023-09-26T15:23:36.7515243Z       at processTimers (node:internal/timers:512:7)
```

The second error may be connected to the fact that in the test that I've added (`"request with an empty content topic should fail"`) I got rid or asynchronicity because a) it caused some other issue; b) it's not really needed. In this case, there is no "the server pushes a message to the network, let's wait what comes back" scenario. The server rejects the request before interacting with the Relay network. Hence, my test has no `handlerFuture`.

I'm not sure how to explain the first failing test.

waku/waku_lightpush/protocol.nim
```diff
@@ -54,24 +54,31 @@ proc initProtocolHandler*(wl: WakuLightPush) =
...
-      error "pushed message handling failed", error= getCurrentExceptionMsg()
-
-
-    if handleRes.isOk():
-      response = PushResponse(is_success: true, info: some("OK"))
+    
+    if contentTopic == "":
+      response = PushResponse(is_success: false, info: some(emptyContentTopicFailure))
+      waku_lightpush_errors.inc(labelValues = [emptyContentTopicFailure])
+      error "empty content topic", error=emptyContentTopicFailure
```
jm-clius, line 65: I don't quite interpret the bug in https://github.com/waku-org/nwaku/issues/1641 in this way. The issue is not that the content topic is an "empty string" it's that that field is _unset_ in the protocol buffer, resulting in a wrongly encoded RPC that fails to decode. The same issue states:

> Send a light push rpc request that contains a missing protobuf field such as content topic

So it's not related to the content topic specifically, but any missing field that causes an rpc decode failure. Since we still return in lines 45 and 51 on invalid RPC (protobufs), afaict the original issue won't be fixed.

tests/test_waku_lightpush.nim
```diff
@@ -115,4 +116,40 @@ suite "Waku Lightpush":
...
+    
+    let handler = proc(peer: PeerId, pubsubTopic: PubsubTopic, message: WakuMessage): Future[WakuLightPushResult[void]] {.async.} =
+        ## this handler will never be called: request will fail due to empty content topic
+        return ok()
+
+    let
+      server = newTestWakuLightpushNode(serverSwitch, handler)
+      client = newTestWakuLightpushClient(clientSwitch)
+      serverPeerId = serverSwitch.peerInfo.toRemotePeerInfo()
+      message = fakeWakuMessage(contentTopic="")
```
jm-clius, line 140: The bug is that an empty response is received, not on content topic being an empty string, but that the Waku Message and PushRPC is constructed without this field and decoding fails.

jm-clius (changes requested):
Thanks! I think the approach here of trying to have a default response is the right one. Changes requested is that the bug is about an unset field and RPC decoding failure (not content topic being an "empty string"). My suggestion is to first try to trigger this error condition in the unit test and then make sure that these failure cases also lead to a properly constructed response with error message. The same condition is triggered when the WakuMessage is encoded in any other non-supported way and can't be decoded, as reported in https://github.com/waku-org/nwaku/issues/2059. This PR should also close that issue.

jm-clius:
> P.S. I wanted to put a link to the RFC that says that content topic must not be empty

Ah, I think it's not up to the lightpush service node to interpret this. In other words, it's not about content topics, but about properly constructed protocol buffers and fields. The issue is that the message being published MUST be a well-formed WakuMessage (as in https://rfc.vac.dev/spec/14/). If for any reason this message field (or other field in the PushRequest) fails to decode properly, we should return a proper response.

waku/waku_lightpush/protocol.nim
s-tikhomirov: Thank you, now I think i understand the issue properly!

If we want to return a `PushResponse` after a decoding failure, what should its `requestId` be set to? Normally, it is copied from the corresponding field in `PushRequest`, but if the request cannot be decoded properly, we can't count on that. Can we just omit the `requestId` field in that case? It is not wrapped in `Option` in the definition of `PushRPC`... Or shall we use some default value?

jm-clius: Mmm. This is a good point. I would actually guess that in almost all cases decoding fail because of an inner field (either `PushRequest` or `WakuMessage`) failing to decode, rather than the outer `PushResponse`. In other words, it may be possible to change to a two stage decoding process where the `requestId` can be extracted before attempting to decode the encapsulated request.

However, for now this may be overkill and the best option is probably to initialise the `requestId` as an empty string.

@fryorcraken @danisharora099 since the PushRPC and PushRequest fail to decode properly in this bug, it's difficult to extract a `requestId`. Would a decent solution (from js-waku's perspective) be to return a well-formed response, with error code, but with the `requestId` set to an empty string?

s-tikhomirov:
Thinking more about this, I feel that I don't grasp some fundamentals of what's happening here.

What exactly does it mean to "decode" a message? In particular, how can the following line not fail, if there is no, say, `contentTopic` inside `message`?

https://github.com/waku-org/nwaku/blob/e85f05b0b60c86d4a24c3f61298acd9d99d90a63/waku/waku_lightpush/protocol.nim#L56

After all, the left-hand-side `message` is type-inferred to `WakuMessage`, and `WakuMessage` must have a content topic by definition... How come a statically typed language allows us to put a value into a variable that is not guaranteed to be compatible with that variable's type definition?

s-tikhomirov:
Experimenting a bit more, I am not sure I understand how to reproduce the behavior in question. AFAICT, Nim assigns default values to object fields if they are not provided. I modified a playground example with a simplified version of `WakuMessage` ([live plauground](https://play.nim-lang.org/#ix=4HEG)):

```
import strformat

type WakuMessage = object
  contentTopic: string
  ephemeral: bool

let messages = [
  WakuMessage(contentTopic: "SomeTopic", ephemeral: true),
  WakuMessage(contentTopic: "DefaultEphemeral"),
  WakuMessage()
]

for message in messages:
  echo(fmt"Message for content topic {message.contentTopic} is ephemeral: {message.ephemeral}")
```

The result is:
```
Message for content topic SomeTopic is ephemeral: true
Message for content topic DefaultEphemeral is ephemeral: false
Message for content topic  is ephemeral: false
```

My point is: how to I generate a `WakuMessage` with a _missing_ field (which is not the same as a field equal to an empty string)? If it is not possible, then how should I reproduce the behavior we're trying to fix?

s-tikhomirov:
> simply generate a set of raw bytes

Makes sense. It's not clear yet how to use these raw bytes. I can no longer use `client.publish()`, as passing raw bytes as `message` doesn't type-check, and rightly so:

https://github.com/waku-org/nwaku/blob/e85f05b0b60c86d4a24c3f61298acd9d99d90a63/tests/test_waku_lightpush.nim#L104

Similarly, `wl.sendPushRequest()` expects a `PushRequest` and not raw bytes:

https://github.com/waku-org/nwaku/blob/e85f05b0b60c86d4a24c3f61298acd9d99d90a63/waku/waku_lightpush/client.nim#L71

Going down the call stack, it seems that here I can insert random bytes instead of `rpc.encode().buffer`:

https://github.com/waku-org/nwaku/blob/e85f05b0b60c86d4a24c3f61298acd9d99d90a63/waku/waku_lightpush/client.nim#L46

I'm unsure though, what is the proper way to bring this logic into the test? Copy-pasting the whole `sendPushRequest` from `client.nim` to the test with one minor modification (push raw bytes instead of a valid encoding) just feels wrong. Is there a better way?

jm-clius:
Mmm. I see your point. I think this is a good indication that our code is not very unit testable (a sign that concerns aren't well separated).

I wouldn't test this from the client perspective, but rather try to test the lightpush (server-side) protocol directly. One way to do this is to extract the code inside the handler

https://github.com/waku-org/nwaku/blob/master/waku/waku_lightpush/protocol.nim#L39-L77

that returns a `PushRPC` response into a separate function, taking the raw bytes as argument and returning a `PushRPC` response. You could then simply unit test this new function. The benefit is that the logic will now be much better separated and maintainable. The `handle` call will then become something like:

```Nim
proc handle(conn: Connection, proto: string) {.async, gcsafe, closure.} =
    let buffer = await conn.readLp(MaxRpcSize.int)
    let response = handleRequest(buffer)
    await conn.writeLp(response.encode().buffer)
 ```
 
 with `handleRequest` being unit testable with any raw buffer.
 
 WDYT of this option?

**Comment by s-tikhomirov** [comment]
Thank you @jm-clius for the tip, I agree with the approach.

> taking the raw bytes as argument and returning a PushRPC response

It should also probably take a `WakuLightPush` as an argument, as producing a response implies, among other things, pushing a message into the network? (If decoding is successful, that is.)

File: waku/waku_lightpush/protocol.nim (hunk 1 above)
Comment (fryorcraken) [reply waku/waku_lightpush/protocol.nim]: > @fryorcraken @danisharora099 since the PushRPC and PushRequest fail to decode properly in this bug, it's difficult to extract a `requestId`. Would a decent solution (from js-waku's perspective) be to return a well-formed response, with error code, but with the `requestId` set to an empty string?

This is a fair solution IMO.

We currently do not use the `requestId`, we are able to assume the response is on the same multiplex stream than the request (or at least it translates this way in js-libp2p API).

**Comment by s-tikhomirov** [comment]
> minimize rightward drift

What would be the best way to reconcile this with avoiding early returns? My idea was that as `handleRequest` progresses with the checks, it sets the appropriate values for the error message, but the actual response object with that error message is just constructed just once, in the end. I want something like: "if decoding fails, set error message to `decoding error` and skip other checks". Goto-like behavior, if you will. What's the Nim way to achieve this?

File: waku/waku_lightpush/protocol_metrics.nim [hunk 3]
```diff
@@ -17,4 +17,5 @@ const
   peerNotFoundFailure* = "peer_not_found_failure"
   emptyRequestBodyFailure* = "empty_request_body_failure"
   emptyResponseBodyFailure* = "empty_response_body_failure"
+  emptyContentTopicFailure* = "empty_content_topic_failure"
```
jm-clius, line 20: This would not be an "empty content topic failure" anymore, but a "decode rpc failure". Empty content topic is just one reason why RPC decoding would fail.

tests/test_waku_lightpush.nim
```diff
@@ -4,15 +4,17 @@ import
   testutils/unittests,
   chronicles,
   chronos,
-  libp2p/crypto/crypto
+  libp2p/crypto/crypto,
+  std/options
```
jm-clius, line 8: Nitpick: in general our imports are alphabetical, but we add `std` imports at the very top of the import list, followed by `testutils`

waku/waku_lightpush/protocol.nim
```diff
@@ -35,47 +35,50 @@ type
...
+      try:
+        handleRes = await wl.pushHandler(peerId, pubsubTopic, message)
+        isSuccess = handleRes.isOk()
+        pushResponseInfo = (if isSuccess: "OK" else: handleRes.error)
+      except Exception:
+        pushResponseInfo = getCurrentExceptionMsg()
+
+      if not isSuccess:
+        waku_lightpush_errors.inc(labelValues = [messagePushFailure])
+        error "pushed message handling failed", error=pushResponseInfo
```
jm-clius, line 72: Minor comment: seems like this can be outdented one level to apply to the entire outer `if...else`. In other words, the error metric increment and log in lines 45, 46 can be removed and error logging done only here.

waku/waku_lightpush/protocol.nim
```diff
@@ -35,47 +35,50 @@ type
...
-      waku_lightpush_errors.inc(labelValues = [messagePushFailure])
-      error "pushed message handling failed", error=handleRes.error
+      requestId = pushRpcRequest.requestId
+      let request = pushRpcRequest.request
+      waku_lightpush_messages.inc(labelValues = ["PushRequest"])
+      let
+        pubSubTopic = request.get().pubSubTopic
+        message = request.get().message
+      debug "push request", peerId=peerId, requestId=requestId, pubsubTopic=pubsubTopic
+      var handleRes: WakuLightPushResult[void]
```
jm-clius, line 62: I'd add a newline above this line.

waku/waku_lightpush/protocol.nim
```diff
@@ -35,47 +35,50 @@ type
...
-      response = PushResponse(is_success: true, info: some("OK"))
+      pushResponseInfo = emptyRequestBodyFailure
     else:
-      response = PushResponse(is_success: false, info: some(handleRes.error))
-      waku_lightpush_errors.inc(labelValues = [messagePushFailure])
-      error "pushed message handling failed", error=handleRes.error
+      requestId = pushRpcRequest.requestId
+      let request = pushRpcRequest.request
+      waku_lightpush_messages.inc(labelValues = ["PushRequest"])
+      let
```
jm-clius, line 58: I'd add a newline above this line

jm-clius (approved):
LGTM! Some minor comments and nitpicks below.

SionoiS:
> > minimize rightward drift
> 
> What would be the best way to reconcile this with avoiding early returns? My idea was that as `handleRequest` progresses with the checks, it sets the appropriate values for the error message, but the actual response object with that error message is just constructed just once, in the end. I want something like: "if decoding fails, set error message to `decoding error` and skip other checks". Goto-like behavior, if you will. What's the Nim way to achieve this?

This avoid `var` response and is more compact. Also, no need for try/catch, `PushMessageHandler` returns an error.
```
let response =
      if contentTopic == "":
        error "empty content topic", error=emptyContentTopicFailure
        waku_lightpush_errors.inc(labelValues = [emptyContentTopicFailure])
        PushResponse(is_success: false, info: some(emptyContentTopicFailure))
      elif (let res = await wl.pushHandler(conn.peerId, pubsubTopic, message); res.isErr()):
          waku_lightpush_errors.inc(labelValues = [messagePushFailure])
          error "pushed message handling failed", error=res.error
          PushResponse(is_success: false, info: some(res.error))
      else: PushResponse(is_success: true, info: some("OK"))
```

waku/waku_lightpush/protocol.nim
s-tikhomirov: I'm assuming you mean above `let` on the preceding line?

s-tikhomirov:
> > > minimize rightward drift
> > 
> > 
> > What would be the best way to reconcile this with avoiding early returns? My idea was that as `handleRequest` progresses with the checks, it sets the appropriate values for the error message, but the actual response object with that error message is just constructed just once, in the end. I want something like: "if decoding fails, set error message to `decoding error` and skip other checks". Goto-like behavior, if you will. What's the Nim way to achieve this?
> 
> This avoids `var` response and is more compact. Also, no need for try/catch, `PushMessageHandler` returns an error.
> 
> ```
> let response =
>       if contentTopic == "":
>         error "empty content topic", error=emptyContentTopicFailure
>         waku_lightpush_errors.inc(labelValues = [emptyContentTopicFailure])
>         PushResponse(is_success: false, info: some(emptyContentTopicFailure))
>       elif (let res = await wl.pushHandler(conn.peerId, pubsubTopic, message); res.isErr()):
>           waku_lightpush_errors.inc(labelValues = [messagePushFailure])
>           error "pushed message handling failed", error=res.error
>           PushResponse(is_success: false, info: some(res.error))
>       else: PushResponse(is_success: true, info: some("OK"))
> ```

(Judging by `if contentTopic == ""`, it seems to me that this code is modified based on an outdated commit, but anyway.)

We don't only generate a `PushResponse` here, but also we need to extract the `requestId` (but only if the buffer is decoded properly), and log the error (if it occurs). With that in mind, I'm not sure moving `if-elif-else` into `let response = ` applies well here: we still need to remember what error occurred to log it later. And as long as we're doing that, why not also construct `PushResponse` in the end as well (before it is in turn used to construct `PushRPC`).

SionoiS:
> (Judging by `if contentTopic == ""`, it seems to me that this code is modified based on an outdated commit, but anyway.)
> 
> We don't only generate a `PushResponse` here, but also we need to extract the `requestId` (but only if the buffer is decoded properly), and log the error (if it occurs). With that in mind, I'm not sure moving `if-elif-else` into `let response = ` applies well here: we still need to remember what error occurred to log it later. And as long as we're doing that, why not also construct `PushResponse` in the end as well (before it is in turn used to construct `PushRPC`).

Feel free to ignore it was just a suggestion.

waku/waku_lightpush/protocol.nim
```diff
@@ -35,47 +35,46 @@ type
...
-      return
+proc handleRequest*(wl: WakuLightPush, peerId: PeerId, buffer: seq[byte]): Future[PushRPC] {.async.} = 
+  let reqDecodeRes = PushRPC.decode(buffer)
+  var
+      isSuccess = false
+      pushResponseInfo = ""
+      requestId = ""
+
+  if reqDecodeRes.isErr():
+    pushResponseInfo = decodeRpcFailure
```
Ivansete-status, line 46: Shall we append the decode error to the response?
```suggestion
  if reqDecodeRes.isErr():
    pushResponseInfo = decodeRpcFailure & ": " & $reqDecodeRes.error
```

tests/test_waku_lightpush.nim
```diff
@@ -115,4 +117,36 @@ suite "Waku Lightpush":
...
+    
+    ## When
+    let
+      pushRpcResponse = await server.handleRequest(fakePeerId, fakeBuffer)
+      requestId = pushRpcResponse.requestId
+    
+    ## Then
+    check:
+      requestId == ""
+      pushRpcResponse.response.isSome()
```
Ivansete-status, line 145: Shall we also check the received message response in this case?

Ivansete-status (changes requested):
It looks great indeed! I simply add a recommendation to enrich a little bit the message in case or error.

tests/test_waku_lightpush.nim
s-tikhomirov: In the following lines, we check that `isSuccess` is false, and that `info` is (more precisely, after the suggested change above, contains)  `decodeRpcFailure`. What else do you think we should check here?

Ivansete-status: ah ok! I missed that. I was kind of expecting a hardcoded value. Didn't notice you were properly using that `const`.

Ivansete-status (approved):
LGTM! Thanks!

s-tikhomirov:
I tried to retroactively sign my older commits (which GitHub required) but apparently messed something up. Now all my commits in this branch and a prior commit by @vpavlin are squashed together.

Shall I squash-and-merge anyway? If so, should a commit message describe my changes on top of the older changes squashed here? The whole thing looks ugly, I must admit, and I'm sorry for that.

s-tikhomirov:
> I would start with rebasing the PR :)

That was a great idea, thanks @alrevuelta ! Now at least I only see _relevant_ changes in this PR. However, they are now seemingly "authored by Vaclav and committed by me", and this commit is not signed, blocking the merge again (which is why I started this local-squash business in the first place). Any ideas on how to go about this?

alrevuelta:
> Any ideas on how to go about this?

mm as an easy fix I would say:
* undo the last commit. git reset HEAD^1
* commit again. "git add -A", "git commit -m "your commit name"
* if you have configured your keys in the prev step, that new commit should be signed)
* "git push --force-with-lease" to update the PR

PR #2089, opened 2023-09-28
chore(cbindings): cbindings rust simple libwaku integration example

Ivansete-status:
# Description
The purpose of this PR is to show how the `libwaku` can be used from Rust.

So far, it performs the following points:
- `waku_new` (create waku node.)
- `waku_version` (extract default pubsub topic to the Rust space.)
- `waku_default_pubsub_topic` (extract default pubsub topic to the Rust space.)

The "trampoline" technique is being used in order to properly pass Rust closures as callbacks to the `libwaku`. Therefore, we don't have the same concise code that we can achieve in cpp but it is indeed feasible. The code can become more concise by defining `macros!` and the idea for the next step would be to properly expose a safe API to the user and leave the unsafe code managed internally.

Notice that this PR also includes changes in `libwaku` that make it avoid the use of global variables, as stated in the next recommendation: https://github.com/waku-org/nwaku/pull/1865#discussion_r1282722954. This should definitely go in a separate PR.

cc - @jm-clius @fryorcraken @alrevuelta @vpavlin @NagyZoltanPeter @gabrielmer


examples/rust/src/main.rs
```diff
@@ -0,0 +1,128 @@
...
+        };
+        let cb = get_trampoline(&closure);
+        waku_set_user_data(&mut ctx, &closure as *const _ as *const c_void);
+        let _ret = waku_version(
+            &mut ctx,
+            cb,
+        );
+
+        // Extracting the default pubsub topic
+        let default_pubsub_topic: Arc<RefCell<String>> = Arc::new(RefCell::new(String::from("")));
```
SionoiS, line 110: Same here

SionoiS (approved):
LGTM but could be simplified a bit.

examples/rust/src/main.rs
```diff
@@ -0,0 +1,128 @@
...
+            cb,
+        );
+
+        // Extracting the current waku version
+        let version: Arc<RefCell<String>> = Arc::new(RefCell::new(String::from("")));
+        let closure = |data: &str| {
+            *version.borrow_mut() = data.to_string();
+        };
+        let cb = get_trampoline(&closure);
+        waku_set_user_data(&mut ctx, &closure as *const _ as *const c_void);
```
richard-ramos, line 103: I think it should be possible to get rid of this. See experiment here: https://github.com/richard-ramos/test_rust_ffi/blob/af098508acfae181df508eee524d75f70af68010/rustlib/src/lib.rs#L29-L38

Ivansete-status: It looks beautiful indeed!

examples/rust/src/main.rs
```diff
@@ -0,0 +1,128 @@
...
+        let config_json_str = CString::new(config_json).unwrap();
+        waku_set_user_data(&mut ctx, &closure as *const _ as *const c_void);
+        let _ret = waku_new(
+            &mut ctx,
+            config_json_str.as_ptr() as *const u8,
+            cb,
+        );
+
+        // Extracting the current waku version
+        let version: Arc<RefCell<String>> = Arc::new(RefCell::new(String::from("")));
```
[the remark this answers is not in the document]
Ivansete-status: Thanks so much! This has been applied in https://github.com/waku-org/nwaku/pull/2089/commits/36df31b6a5d3b2a7dcf60a87ca6afa772257e659

richard-ramos (approved):

PR #2090, opened 2023-09-28
feat: filter service peers by shard

SionoiS:
# Description
When selecting a peer via `peer_manager` if a service slot is chosen, the shard that the peer support is checked.

### DIFF
--- a/waku/node/peer_manager/peer_manager.nim
+++ b/waku/node/peer_manager/peer_manager.nim
@@ -652,8 +652,14 @@ proc selectPeer*(pm: PeerManager, proto: string, shard: Option[PubsubTopic] = no
 
   # For other protocols, we select the peer that is slotted for the given protocol
   pm.serviceSlots.withValue(proto, serviceSlot):
-    debug "Got peer from service slots", peerId=serviceSlot[].peerId, multi=serviceSlot[].addrs[0], protocol=proto
-    return some(serviceSlot[])
+    if shard.isSome() and serviceSlot[].enr.isSome() and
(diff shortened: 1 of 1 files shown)


Ivansete-status (commented):
I suggest a possible change. If too complicated then it is okay.

Btw, what do "service slots" contain?

SionoiS:
> Btw, what do "service slots" contain?

A Peer that has the capability required.

SionoiS:
Fixed the logic flaw.

jm-clius (commented):
Wondering if this is necessary or would work in practice. The service slots are for manually configured service peers. Presumably then the node operator has some external information to "know" that this service peer is the one to use for whichever shard it has configured, even if the node itself cannot verify this info. In fact, these service peers are not discovered via any existing discovery means (statically configured `multiaddrs`) so are very unlikely to ever have an ENR associated.

SionoiS:
> Wondering if this is necessary or would work in practice. The service slots are for manually configured service peers. Presumably then the node operator has some external information to "know" that this service peer is the one to use for whichever shard it has configured, even if the node itself cannot verify this info. In fact, these service peers are not discovered via any existing discovery means (statically configured `multiaddrs`) so are very unlikely to ever have an ENR associated.

Maybe it would be better to check shard support when service peers are added? OR like you said scrap it and assume the service peers do have the shard and capability the node need.

SionoiS:
Closing this. Service peers have to be add manually to the config so it's safe to assume the user will pick the peer they want.

PR #2091, opened 2023-09-28
chore: adding NetConfig test suite

### LINKED ISSUE #1540: chore: improve test coverage on NetConfig generation
### Background
In #1512, the waku node network config was abstracted out.
Test coverage on address generation is low at the moment.

### Details
Tests for the network config generation to improve coverage to detect regressions.

### Acceptance criteria
- [ ] Identify common network configs
- [ ] Tests for the same

cc: @jm-clius

gabrielmer:
# Description
Added test suite to cover common NetConfig settings

# Changes

- [x] Created new file `./tests/test_waku_netconfig.nim` with a NetConfig test suite
- [x] Moved `defaultTestWakuNodeConf` from `test_app.nim` to `external_config.nim`
- [x] Updated `defaultTestWakuNodeConf`to use our current default `listenAddress` `0.0.0.0` and added more default parameters

## Issue

closes #1540

### DIFF
--- a/apps/wakunode2/external_config.nim
+++ b/apps/wakunode2/external_config.nim
@@ -552,19 +552,4 @@ proc load*(T: type WakuNodeConf, version=""): ConfResult[T] =
   except CatchableError:
     err(getCurrentExceptionMsg())
 
-proc defaultTestWakuNodeConf*(): WakuNodeConf =
-  WakuNodeConf(
-    tcpPort: Port(60000),
-    websocketPort: Port(8000),
-    listenAddress: ValidIpAddress.init("0.0.0.0"),
-    rpcAddress: ValidIpAddress.init("127.0.0.1"),
-    restAddress: ValidIpAddress.init("127.0.0.1"),
-    metricsServerAddress: ValidIpAddress.init("127.0.0.1"),
-    dnsAddrsNameServers: @[ValidIpAddress.init("1.1.1.1"), ValidIpAddress.init("1.0.0.1")],
-    nat: "any",
-    maxConnections: 50,
-    topics: @["/waku/2/default-waku/proto"],
-    relay: true
-  )
-
 {.pop.}
--- a/apps/wakunode2/external_config.nim
+++ b/apps/wakunode2/external_config.nim
@@ -552,4 +552,19 @@ proc load*(T: type WakuNodeConf, version=""): ConfResult[T] =
   except CatchableError:
     err(getCurrentExceptionMsg())
 
+proc defaultTestWakuNodeConf*(): WakuNodeConf =
+  WakuNodeConf(
+    tcpPort: Port(60000),
+    websocketPort: Port(8000),
+    listenAddress: ValidIpAddress.init("0.0.0.0"),
+    rpcAddress: ValidIpAddress.init("127.0.0.1"),
+    restAddress: ValidIpAddress.init("127.0.0.1"),
+    metricsServerAddress: ValidIpAddress.init("127.0.0.1"),
+    dnsAddrsNameServers: @[ValidIpAddress.init("1.1.1.1"), ValidIpAddress.init("1.0.0.1")],
+    nat: "any",
+    maxConnections: 50,
+    topics: @["/waku/2/default-waku/proto"],
+    relay: true
+  )
+
 {.pop.}
--- a/tests/test_waku_netconfig.nim
+++ b/tests/test_waku_netconfig.nim
@@ -133,6 +133,31 @@ asyncTest "AnnouncedAddresses includes extMultiAddrs when provided":
       netConfig.announcedAddresses[1] == extMultiAddrs[0]
 
 
+asyncTest "AnnouncedAddresses uses dns4DomainName over extIp when both are provided":
+
+  let 
+    conf = defaultTestWakuNodeConf()
+    dns4DomainName = "example.com"
+    extIp = ValidIpAddress.init("1.2.3.4")
+    extPort = Port(1234)
+        
+  let netConfigRes = NetConfig.init(
+    bindIp = conf.listenAddress,
+    bindPort = conf.tcpPort,
+    dns4DomainName = some(dns4DomainName),
+    extIp = some(extIp),
+    extPort = some(extPort)
+  )
+     
+  require:
+    netConfigRes.isOk()
+
+  let netConfig = netConfigRes.get()
+
+  check:
+    netConfig.announcedAddresses.len == 1  # DNS address
+    netConfig.announcedAddresses[0] == dns4TcpEndPoint(dns4DomainName, extPort) 
+
 asyncTest "AnnouncedAddresses includes WebSocket addresses when enabled":
 
     var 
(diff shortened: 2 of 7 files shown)


tests/testlib/wakunode.nim
```diff
@@ -13,8 +13,8 @@ import
   ../../../waku/node/peer_manager,
   ../../../waku/waku_enr,
   ../../../waku/waku_discv5,
+  ../../apps/wakunode2/external_config,
   ../../apps/wakunode2/internal_config,
-  ../wakunode2/test_app,
```
gabrielmer, line 17: When importing `defaultTestWakuNodeConf` from `test_app` into other test suites, the `test_app` suite was being re-run as part of the suites importing it. For easier `defaultTestWakuNodeConf`'s reusability, moved the function to `external_config`, mimicking  `newTestWakuNode`, which is reused by many test suites and is defined in `wakunode.nim`

tests/test_waku_netconfig.nim
```diff
@@ -0,0 +1,297 @@
...
+     
+  require:
+    netConfigRes.isOk()
+
+  let netConfig = netConfigRes.get()
+
+  check:
+    netConfig.announcedAddresses.len == 2  # External address + wsHostAddress
+    netConfig.announcedAddresses[1] == (ip4TcpEndPoint(extIp, 
+      conf.websocketPort) & wsFlag(wssEnabled))
```
gabrielmer, line 205: Note that in order to configure WS using an `extIp`, the published port isn't `extPort` but `conf.websocketPort`. For websockets we bind to and announce the same port number.

tests/test_waku_netconfig.nim
```diff
@@ -0,0 +1,351 @@
...
+      netConfigRes.isOk()
+
+    let netConfig = netConfigRes.get()
+
+    check:
+      netConfig.announcedAddresses.len == 1  # Only bind address should be present
+      netConfig.announcedAddresses[0] == formatListenAddress(ip4TcpEndPoint(conf.listenAddress, conf.tcpPort))
+
+
+asyncTest "AnnouncedAddresses contains external address if extIp/Port are provided":
```
jm-clius, line 66: Ditto in terms of indentation and for other tests (I think they should be single indented to be considered part of "Waku Netconfig" `suite`

jm-clius (commented):
Looks good and covers quite a lot! Thanks for this. Much better than what we have (which was noting 😅 ). Feel free to mark as ready for review once you think it can be merged.

tests/test_waku_netconfig.nim
gabrielmer: Good catch! For some reason the tests were still being run as part of the test suite with that indentation, which is pretty interesting. Just pushed a fix for it - thank you!

apps/wakunode2/external_config.nim
```diff
@@ -552,4 +552,19 @@ proc load*(T: type WakuNodeConf, version=""): ConfResult[T] =
...
+    listenAddress: ValidIpAddress.init("0.0.0.0"),
+    rpcAddress: ValidIpAddress.init("127.0.0.1"),
+    restAddress: ValidIpAddress.init("127.0.0.1"),
+    metricsServerAddress: ValidIpAddress.init("127.0.0.1"),
+    dnsAddrsNameServers: @[ValidIpAddress.init("1.1.1.1"), ValidIpAddress.init("1.0.0.1")],
+    nat: "any",
+    maxConnections: 50,
+    topics: @["/waku/2/default-waku/proto"],
+    relay: true
+  )
```
jm-clius, line 568: Why has this been moved from `test_app.nim`? I wouldn't move any test code into "production" code.

tests/testlib/wakunode.nim
jm-clius: Ah, made a comment about this elsewhere before seeing this. `wakunode.nim` is a `testlib` - in other words, it's designed to be a utility module for functions useful in tests. `newTestWakuNode` `defaultTestWakuNodeConf` etc. shouldn't be defined in any "production" modules (test code should remain completely separate. The way to prevent `test_app` from being run multiple times for each import is to move `defaultTestWakuNodeConf` to a test helper module. I'd suggest just adding it to `testlib/wakunode.nim` and importing in both `test_app` and `test_waku_netconfig`

jm-clius (approved):
LGTM. `defaultTestWakuNodeConf` should not be defined in a "production" module llike `external_config.nim`, but once this has been moved I'll be happy for the suite to be merged. No need to re-request review from my side. Great job!

gabrielmer:
> LGTM. `defaultTestWakuNodeConf` should not be defined in a "production" module llike `external_config.nim`, but once this has been moved I'll be happy for the suite to be merged. No need to re-request review from my side. Great job!

Ooh now everything makes much more sense! I didn't notice that `wakunode.nim` was a `testlib` 😶
Will move `defaultTestWakuNodeConf`  to `wakunode.nim` then :)

Thanks so much!

tests/test_waku_netconfig.nim
```diff
@@ -0,0 +1,351 @@
...
+
+    let conf = defaultTestWakuNodeConf()
+        
+    let netConfigRes = NetConfig.init(
+      bindIp = conf.listenAddress,
+      bindPort = conf.tcpPort
+    )
+     
+    require:
+      netConfigRes.isOk()
```
Ivansete-status, line 58: I'd always apply the next technique for such cases:

```suggestion
    assert netConfigRes.isOk(), $netConfigRes.error
```

That applies elsewhere.

tests/test_waku_netconfig.nim
```diff
@@ -0,0 +1,351 @@
...
+import
+  chronos,
+  libp2p/multiaddress,
+  testutils/unittests
+
+import
+    ../../waku/node/config,
+    ../../waku/waku_enr/capabilities,
+    ../../apps/wakunode2/external_config
+
```
Ivansete-status, line 13: Sorry, this is a personal obsession hehe. If possible I recommend always to keep one empty line max.
```suggestion
```

waku/node/config.nim
```diff
@@ -36,7 +36,7 @@ type NetConfig* = object
 type NetConfigResult* = Result[NetConfig, string]
 
 
-template ip4TcpEndPoint(address, port): MultiAddress =
+template ip4TcpEndPoint*(address, port): MultiAddress =
```
Ivansete-status, line 39: Isn't there a way to make the tests without changing the scope of that?
If possible, I recommend only testing the public procs and don't change the scope of private ones. If you still need to use these private procs you may need to `include` the `../../waku/node/config,` in the test, instead of `import`.

Ivansete-status (changes requested):
Thanks for that! You created many more test cases than I could imagine :1st_place_medal: 
Nevertheless, there are a few things that need to be revisited :)
It is especially important the comment about `apps/wakunode2/external_config.nim`. I'd prevent changing it.
Cheers!

SionoiS (approved):
No further comments than what was already said.

I just feel like we could use the type system in cases like this to enforce some "rules".

My 0.02$

tests/test_waku_netconfig.nim
gabrielmer: Agree that it's a better technique! Done :)

waku/node/config.nim
gabrielmer: Makes sense! I changed it to `import` instead. I noticed that for some reason really few test files use `import` - I wonder how many private procs have been turned into public for testing, probably a lot. Anyway, I agree on the principle of making the least amount possible of changes in prod code that aren't related to prod logic.

tests/test_waku_netconfig.nim
gabrielmer: Hahaha done, deleted that line and also a couple more instances in the PR with extra empty lines :)

gabrielmer:
> No further comments than what was already said.
> 
> I just feel like we could use the type system in cases like this to enforce some "rules".
> 
> My 0.02$

I'm not sure I fully understand the idea. Is it about setting explicitly the type when declaring variables instead of having the types being automatically inferred so we enforce the return types of procs?

Or what kind of rules are you thinking of?

Ivansete-status (approved):
LGTM! Thanks for it! :100:

SionoiS:
> > No further comments than what was already said.
> > I just feel like we could use the type system in cases like this to enforce some "rules".
> > My 0.02$
> 
> I'm not sure I fully understand the idea. Is it about setting explicitly the type when declaring variables instead of having the types being automatically inferred so we enforce the return types of procs?
> 
> Or what kind of rules are you thinking of?

Might not be practical to do this here but the type system can help enforcing requirement like
`AnnouncedAddresses contains only bind address when no external addresses are provided`

For example you could use some type to represent the state.
- `only bind address when no external addresses are provided`
- `only external address if extIp/Port are provided`

PR #2096, opened 2023-10-02
test(waku-filter): Waku node filter privacy and security tests

AlejandroCabeza:
# Description
Implement privacy and security tests for waku node filter.

# Changes
- [X] Implement privacy and security tests.

# Dependencies
https://github.com/waku-org/nwaku/pull/2095

### DIFF
--- a/tests/node/test_wakunode_filter.nim
+++ b/tests/node/test_wakunode_filter.nim
@@ -25,12 +25,12 @@ import
 
 let FUTURE_TIMEOUT = 1.seconds
 
-suite "Full Node - Waku Filter - End to End":
+suite "Waku Filter - End to End":
   var client {.threadvar.}: WakuNode
   var clientPeerId {.threadvar.}: PeerId
   var server {.threadvar.}: WakuNode
   var serverRemotePeerInfo {.threadvar.}: RemotePeerInfo
-  var pubsubTopic {.threadvar.}: PubsubTopicy
+  var pubsubTopic {.threadvar.}: PubsubTopic
   var contentTopic {.threadvar.}: ContentTopic
   var contentTopicSeq {.threadvar.}: seq[ContentTopic]
   var pushHandlerFuture {.threadvar.}: Future[(string, WakuMessage)]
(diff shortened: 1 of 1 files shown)


tests/node/test_wakunode_filter.nim
```diff
@@ -25,12 +25,12 @@ import
...
 let FUTURE_TIMEOUT = 1.seconds
 
-suite "Full Node - Waku Filter - End to End":
+suite "Waku Filter - End to End":
   var client {.threadvar.}: WakuNode
   var clientPeerId {.threadvar.}: PeerId
   var server {.threadvar.}: WakuNode
   var serverRemotePeerInfo {.threadvar.}: RemotePeerInfo
-  var pubsubTopic {.threadvar.}: PubsubTopicy
+  var pubsubTopic {.threadvar.}: PubsubTopic
```
Ivansete-status, line 33: What?! How this could even compile xD?
That makes me suspect that maybe we are not running this test at all?

Ivansete-status (approved):
LGTM! Thanks!

tests/node/test_wakunode_filter.nim
AlejandroCabeza: I arrived at the same thought you did 😂 And I do think, although I'm not sure, that some tests are not ran; will check that later.

tests/node/test_wakunode_filter.nim
```diff
@@ -107,3 +107,117 @@ suite "Full Node - Waku Filter - End to End":
...
+    let msg1 = fakeWakuMessage(contentTopic=contentTopic)
+    await server.filterHandleMessage(pubsubTopic, msg1)
+
+    # Then the message is not pushed to the client
+    check (not await pushHandlerFuture.withTimeout(FUTURE_TIMEOUT))
+
+    # Given the client refreshes the subscription
+    # TODO: CHECK IF THIS IS NECESSARY. 
+    # AT FIRST GLANCE IT SEEMS TO COLLIDE WITH WAKU_FILTER_CLIENT'S BEHAVIOUR: SHOULD NOT NEED?
+    let subscribeResponse2 = await client.filterSubscribe(
```
jm-clius, line 173: Is this required for the test to pass? I would expect that it shouldn't be needed as long as:
1. the service node didn't restart
2. the client starts up again with the same peerId and transport as before

jm-clius (approved):
Good set of tests!

tests/node/test_wakunode_filter.nim
AlejandroCabeza: I think so too, even more given the client doesn't do any unsubscribing, apparently. I'll open a ticket.

PR #2098, opened 2023-10-02
chore: added size based retention policy

### LINKED ISSUE #1885: Add retention policy with GB or MB limitation
### Background
We came across an issue in HK-prod node where the SQLite database reached 39GB.
We need a mechanism to apply a retention policy so that the database doesn't grow more than a certain amount of GB, or MB.

### Acceptance criteria
The Waku node with _Store_ mounted should be able to limit the maximum space occupied by the underlying database.
This applies to both sqlite and postgres.

The new retention policy should support the next formats:
```
--store-message-retention-policy=size:25GB
--store-message-retention-policy=size:25gb
--store-message-retention-policy=size:250MB
--store-message-retention-policy=size:250mb
... any of the case combinations
```

ABresting:
# Description
Size retention Policy has been added. Based on the size limit provided in MB or GB. Once the size limit overflows by messages, then the 20% of the outdated messages are dropped and a vacuum of the database for defragmentation is triggered.

# Changes

- [x] Size-related retention policy.
- [x] Vacuum process after the deletion of overflowed database entries in size-based retention policy.
- [x] Vacuum process in capacity-based retention policy.

## Issue

#1885


waku/waku_archive/retention_policy/retention_policy_size.nim
```diff
@@ -0,0 +1,87 @@
...
+                Future[RetentionPolicyResult[void]] {.async.} =
+  ## when db size overshoots the database limit, shread 20% of outdated messages 
+  
+  # to get the size of the database, pageCount and PageSize is required
+  # get page count in "messages" database
+  var pageCountRes = await driver.getPagesCount()
+  if pageCountRes.isErr():
+    return err("failed to get Pages count: " & pageCountRes.error)
+
+  var pageCount: int64 = pageCountRes.value
```
Ivansete-status, line 51: This is great indeed!
However, we can make it a bit shorter with:
```suggestion
  let pageCount = await driver.getPagesCount().isOkOr:
    return err("failed to get Pages count: " & $error)
```

This technique is preferable whenever possible.

waku/waku_archive/retention_policy/retention_policy_size.nim
```diff
@@ -0,0 +1,87 @@
...
+
+  # check if current databse size crosses the db size limit
+  if totalSizeOfDB < p.sizeLimit:
+    return ok()
+
+  # to shread/delete messsges, get the total row/message count
+  var numMessagesRes = await driver.getMessagesCount()
+  if numMessagesRes.isErr():
+    return err("failed to get messages count: " & numMessagesRes.error)
+  var numMessages = numMessagesRes.value
```
Ivansete-status, line 71:
```suggestion
  let numMessages = await driver.getMessagesCount().isOkOr:
    return err("failed to get messages count: " & $error)
```
Notice that I suggest to also use a `let` type because we don't need a mutable `var` at this point.

waku/waku_archive/retention_policy/retention_policy_size.nim
```diff
@@ -0,0 +1,87 @@
...
+
+  let res = await driver.deleteOldestMessagesNotWithinLimit(limit=pageDeleteWindow)
+  if res.isErr():
+      return err("deleting oldest messages failed: " & res.error)
+
+  # vacuum to get the deleted pages defragments to save storage space
+  # this will resize the database size
+  let resVaccum = await driver.performsVacuum()
+  if resVaccum.isErr():
+    return err("vacuumming failed: " & resVaccum.error)
```
Ivansete-status, line 84:
```suggestion
  await driver.performsVacuum().isOkOr:
    return err("vacuumming failed: " & $error)
```

tests/waku_archive/test_retention_policy.nim
```diff
@@ -53,6 +54,45 @@ suite "Waku Archive - Retention policy":
...
+    var putFutures = newSeq[Future[ArchiveDriverResult[void]]]()
+    var retentionFutures = newSeq[Future[ArchiveDriverResult[void]]]()
+
+    for i in 1..excess:
+        let msg = fakeWakuMessage(payload= @[byte i], contentTopic=DefaultContentTopic, ts=Timestamp(i))
+        putFutures.add(driver.put(DefaultPubsubTopic, msg, computeDigest(msg), msg.timestamp))
+        retentionFutures.add(retentionPolicy.execute(driver))
+
+    # waitFor is used to synchronously wait for the futures to complete.
+    discard waitFor allFinished(putFutures & retentionFutures)
```
Ivansete-status, line 80: This is great and I don't see a need to change it.
However, I think it can be reduced a little bit with:
```suggestion
    for i in 1..excess:
        let msg = fakeWakuMessage(payload= @[byte i], contentTopic=DefaultContentTopic, ts=Timestamp(i))
        waitFor driver.put(DefaultPubsubTopic, msg, computeDigest(msg), msg.timestamp)
        waitFor retentionPolicy.execute(driver)
```

Ivansete-status (changes requested):
Thanks for that PR! It is awesome!
I'm just adding a few comments that I hope you find useful.
Especially interesting to me is to only run the retention policy once, and make the limit validation before and after the retention policy execution.

waku/waku_archive/driver.nim
```diff
@@ -45,6 +45,15 @@ method getMessages*(driver: ArchiveDriver,
 method getMessagesCount*(driver: ArchiveDriver):
                          Future[ArchiveDriverResult[int64]] {.base, async.} = discard
 
+method getPagesCount*(driver: ArchiveDriver):
+                         Future[ArchiveDriverResult[int64]] {.base, async.} = discard
+
+method getPagesSize*(driver: ArchiveDriver):
+                         Future[ArchiveDriverResult[int64]] {.base, async.} = discard
+
+method performsVacuum*(driver: ArchiveDriver):
```
jm-clius, line 54:
```suggestion
method performVacuum*(driver: ArchiveDriver):
```

Nitpick. :) (function names should be verbs in imperative form, i.e. without the `s`)

waku/waku_archive/retention_policy/retention_policy_size.nim
```diff
@@ -0,0 +1,87 @@
...
+  # get page count in "messages" database
+  var pageCountRes = await driver.getPagesCount()
+  if pageCountRes.isErr():
+    return err("failed to get Pages count: " & pageCountRes.error)
+
+  var pageCount: int64 = pageCountRes.value
+
+  # get page size of database
+  let pageSizeRes = await driver.getPagesSize()
+  var pageSize: int64 = int64(pageSizeRes.valueOr(0) div 1024)
```
jm-clius, line 55:
```suggestion
  let pageSize: int64 = int64(pageSizeRes.valueOr(0) div 1024)
```

Unless you have a good reason to modify a variable later on, it should be final assigned with `let`. A good rule of thumb is to always use `let`, unless you are forced to create a variable that needs to be modifiable.

waku/waku_archive/retention_policy/retention_policy_size.nim
```diff
@@ -0,0 +1,87 @@
...
+
+  if pageSize == 0:
+    return err("failed to get Page size: " & pageSizeRes.error)
+
+  # database size in megabytes (Mb)
+  var totalSizeOfDB: float = float(pageSize * pageCount)/1024.0
```
jm-clius, line 61:
```suggestion
  let totalSizeOfDB: float = float(pageSize * pageCount)/1024.0
```

jm-clius: (and elsewhere)

jm-clius (approved):
Thanks. Approving from approach POV (will let Ivan take a look at the detail of implementation). Some minor comments on code.

Importantly, though, this doesn't fix https://github.com/waku-org/nwaku/issues/1885, as it still needs to be integrated into the `wakunode2` app configuration. This should be done in a separate PR before that issue can be closed (your comment in the issue description will auto close the issue once this is merged).

tests/waku_archive/test_retention_policy.nim
```diff
@@ -53,6 +54,45 @@ suite "Waku Archive - Retention policy":
...
+            
+    ## Then
+    # calculate the current database size
+    let pageSize = (waitFor driver.getPagesSize()).tryGet()
```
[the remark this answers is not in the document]
ABresting: Good point Ivan,
Changes made to compare the current db size with the size limit, retention policy enabled only in case of overflow.
After the retention policy is executed, again comparison is made between the current db size and the size limit.

waku/waku_archive/retention_policy/retention_policy_size.nim
ABresting: since the getPagesCount method is of future return type `.sOkOr` is not possible.

waku/waku_archive/retention_policy/retention_policy_size.nim
ABresting: same as before, future return type.

tests/waku_archive/test_retention_policy.nim
ABresting: as per the previous review with Simone, the original method is more efficient?

waku/waku_archive/driver.nim
ABresting: good catch!

tests/waku_archive/test_retention_policy.nim
```diff
@@ -53,6 +54,51 @@ suite "Waku Archive - Retention policy":
...
+    discard waitFor allFinished(putFutures)
+
+    ## Then
+    # calculate the current database size
+    var pageSize = (waitFor driver.getPagesSize()).tryGet()
+    var pageCount = (waitFor driver.getPagesCount()).tryGet()
+    var sizeDB = float(pageCount * pageSize) / (1024.0 * 1024.0)
+    # execute policy if the current db size oveflows 
+    if sizeDB >= sizeLimit:
+      require (waitFor retentionPolicy.execute(driver)).isOk()
```
Ivansete-status, line 87: This is great indeed! However, would it be possible to enforce a `require` instead?
```suggestion
    require sizeDB >= sizeLimit
    require (waitFor retentionPolicy.execute(driver)).isOk()
```

waku/waku_archive/retention_policy/retention_policy_size.nim
```diff
@@ -0,0 +1,98 @@
...
+
+  # check if current databse size crosses the db size limit
+  if totalSizeOfDB < p.sizeLimit:
+    return ok()
+
+  # keep deleting until the current db size falls within size limit 
+  while totalSizeOfDB > p.sizeLimit:
```
Ivansete-status, line 66: This logic looks great to me! However, given the possible issue of stepping into an infinite loop, I think we should have a PR in the near future where we prevent that possible blocking issue. For example, if the measured `totalSizeOfDB` doesn't change in two consecutive iterations, then we should break the loop by returning the appropriate error.

Ivansete-status (approved):
LGTM! Thanks for it and for the patience! :100: 

I just added a comment to add a further protection but for a separate PR.

@apentori - this feature is very interesting to be applied in the current `wakuv2.test` and `wakuv2.prod` fleets. `wakuv2.test` will receive that once this PR is merged, and `wakuv2.prod` will have this feature in the next release (I think during next week.)

waku/waku_archive/retention_policy/retention_policy_size.nim
ABresting: yes, I do not like loops in the code, what I can imagine we can do is, since we have a max size/threshold of the database, we can have a lower threshold of the database as well, then inside the loop, we introduce a notation, i.e. max number of times the loop should run so that the size of the database will reduce to the lowest,  

$$ n > \frac{\log(\text{lowerThresholdInMbs}/\text{maxDbSizeOrCurrentDbSize})}{\log(\text{reducedToUponEachIteration})} $$

Here: 
lowerThresholdInMbs is the lowest possible database size possible
maxDbSizeOrCurrentDbSize is the maximum database size possible
reducedToUponEachIteration is the fraction of size the database reduced to upon each iteration

For eg., in the case of 30720 MBs(30 GB) as max size, 0.1 MBs as a lower threshold, and 0.80 as a fraction of the database retained upon each iteration. We get 57 iterations. SO if we use this then there is a guarantee that the loop will not cross 57 times at max.

This way there is a good chance that the forever loop is not encountered.
Post this we can affirm a warning/error if the database size is still not less than the maximum permitted size.

WDYT? @Ivansete-status @alrevuelta

Ivansete-status: I vote to follow the simplest approach. However it is fine for this PR for now and we can revisit that in a separate PR in the near future.

waku/waku_archive/retention_policy/retention_policy_size.nim
```diff
@@ -0,0 +1,98 @@
...
+    return err("failed to get Page size: " & pageSizeRes.error)
+
+  # database size in megabytes (Mb)
+  var totalSizeOfDB: float = float(pageSize * pageCount)/1024.0
+
+  # check if current databse size crosses the db size limit
+  if totalSizeOfDB < p.sizeLimit:
+    return ok()
+
+  # keep deleting until the current db size falls within size limit
```
alrevuelta, line 65: I see in this function a pattern that is usually solved with a `do while`. A "do while" can remove aroud 25 lines of code, since `totalSizeOfDB`, `pageSize` and `pageCount` are a bit boilerplate.

Since nim doesnt have "do while" I would suggest:

```nim
while true:
    ...(get page get count)
    var totalSizeOfDB: float = float(pageSize * pageCount)/1024.0
    if totalSizeOfDB > p.sizeLimit:
      break
    let res = await driver.deleteOldestMessagesNotWithinLimit(limit=pageDeleteWindow)
    if res.isErr():
        return err("deleting oldest messages failed: " & res.error)
```

So with something like this you just need 1 single call to `totalSizeOfDB`, `pageSize` and `pageCount`

alrevuelta (changes requested):
minor comment

alrevuelta (approved):
lgtm

PR #2099, opened 2023-10-02
chore: Revert lightpush error handling to allow zero peer publish again succeed

NagyZoltanPeter:
# Description
Previous change on lightpush error handling caused js-waku tests failing and thus ci errors on nwaku side as well.
There had been a discussion around and decision is made to enrich the protocol for the future.
[[Epic] Enhance light push protocol ](https://github.com/waku-org/pm/issues/93)

Up until than we do not return error in case there is no suitable peer to publish message comes from lightpush.

# Changes

- [X] Changed return error to debug log in lightpush callback in waku_node
- [X] Commented out relevant test that would fail.

# Check
- CI's js-waku should not fail

### DIFF
--- a/tests/wakunode_rest/test_rest_lightpush.nim
+++ b/tests/wakunode_rest/test_rest_lightpush.nim
@@ -171,24 +171,27 @@ suite "Waku v2 Rest API - lightpush":
 
     await restLightPushTest.shutdown()
 
-  asyncTest "Push message request service not available":
-    # Given
-    let restLightPushTest = await RestLightPushTest.init()
+  ## TODO: Re-work this test when lightpush protocol change is done: https://github.com/waku-org/pm/issues/93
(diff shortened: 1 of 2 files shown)


fbarbu15:
I've looked at the js-waku tests and the problem is gone
There's still one failing test but it needs fixing on js side (missing ensureSubscriptions call)
Thanks

fbarbu15 (approved):

jm-clius (approved):
Thanks for this!

NagyZoltanPeter:
> I've looked at the js-waku tests and the problem is gone There's still one failing test but it needs fixing on js side (missing ensureSubscriptions call) Thanks

Hi Floring!
Great.... at least I would not worry about still failing js-waku tests.

NagyZoltanPeter:
Merging this PR to allow js-waku CI jobs succeed. Adjusted with Florin that this change fixes js-waku relevant test fail.

PR #2100, opened 2023-10-02
refactor: relay REST API into two parts; core logic and API specific

SionoiS:
# Description

I refactored the relay REST API separating the logic in two. Same way I did for STORE in #2092

The first part is REST related code, encode, decode, http errors, etc...
The second part is the logic that the request triggers; (un)subscribing, fetching & publishing messages.

In the future tests could be separated the same way, REST and logic but it was not done here.

# Changes

- [x] Added new handlers called when receiving a RELAY request.
- [x] Moved the core logic to app, where handlers are declared.
- [x] Fixed tests

Tracking #1941

### DIFF
--- a/apps/wakunode2/app.nim
+++ b/apps/wakunode2/app.nim
@@ -57,6 +57,9 @@ import
   ./internal_config,
   ./external_config
 
+from std/times import getTime
+from std/times import toUnix
+
 logScope:
   topics = "wakunode app"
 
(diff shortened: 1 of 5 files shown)


NagyZoltanPeter:
@SionoiS I would like to discuss the [option/proposal on the other PR](https://github.com/waku-org/nwaku/pull/2092#issuecomment-1746416867) before taking this one.

SionoiS:
Closing as https://github.com/waku-org/nwaku/pull/2109 is a better approach.

PR #2101, opened 2023-10-03
test(waku-relay): Relay

AlejandroCabeza:
# Description
Implement waku filter relay tests for message id generation

# Changes
- [X] Implement message id tests.
- [X] Implement relay tests.
- [X] Update import paths to use `test_all`.

# Dependencies
https://github.com/waku-org/nwaku/pull/2096

### DIFF
--- a/tests/waku_filter_v2/test_all.nim
+++ b/tests/waku_filter_v2/test_all.nim
@@ -2,5 +2,4 @@
 
 import
   ./test_waku_client,
-  ./test_waku_filter_protocol,
-  ./test_waku_filter
+  ./test_waku_filter_protocol
--- a/tests/waku_relay/crypto_utils.nim
+++ b/tests/waku_relay/crypto_utils.nim
@@ -1,0 +1,44 @@
+import nimcrypto
+
+
+proc cfbEncode*(key: string, iv: string, data: string): seq[byte] =
+  var context: CFB[aes256]
+  var pKey: array[aes256.sizeKey, byte]
+  var pIv: array[aes256.sizeBlock, byte]
+  var pData = newSeq[byte](len(data))
+  var encodedData = newSeq[byte](len(data))
+
+  copyMem(addr pData[0], unsafeAddr data[0], len(data))
+  # WARNING! Do not use 0 byte padding in applications, this is done as example.
+  copyMem(addr pKey[0], unsafeAddr key[0], len(key))
+  copyMem(addr pIv[0], unsafeAddr iv[0], len(iv))
+
+  # Initialization of CFB[aes256] context with encryption key
+  context.init(pKey, pIv)
+  # Encryption process
+  context.encrypt(pData, encodedData)
+  # Clear context of CFB[aes256]
+  context.clear()
+
+  return encodedData
+
+
+proc cfbDecode*(key: string, iv: string, encodedData: seq[byte]): seq[byte] =
+  var context: CFB[aes256]
+  var pKey: array[aes256.sizeKey, byte]
+  var pIv: array[aes256.sizeBlock, byte]
+  var decodedData = newSeq[byte](len(encodedData))
+
+  # copyMem(addr _data[0], addr data[0], len(data))
+  # WARNING! Do not use 0 byte padding in applications, this is done as example.
+  copyMem(addr pKey[0], unsafeAddr key[0], len(key))
+  copyMem(addr pIv[0], unsafeAddr iv[0], len(iv))
+
+  # Initialization of CFB[aes256] context with encryption key
+  context.init(pKey, pIv)
+  # Decryption process
+  context.decrypt(encodedData, decodedData)
+  # Clear context of CFB[aes256]
+  context.clear()
+  
+  return decodedData
--- a/tests/waku_relay/crypto_utils.nim
+++ b/tests/waku_relay/crypto_utils.nim
@@ -1,3 +1,6 @@
+# Source: nimcrypto/examples/cfb.nim
+
+
 import nimcrypto
 
 
--- a/tests/waku_relay/test_message_id.nim
+++ b/tests/waku_relay/test_message_id.nim
@@ -1,0 +1,37 @@
+import 
+  unittest,
+  stew/shims/net,
+  stew/[results, byteutils]
+
+import
+  stew/results,
+  nimcrypto/sha2,
+  libp2p/protocols/pubsub/rpc/messages,
+  ../../../waku/waku_relay/message_id,
+  ../testlib/sequtils
+
+
+suite "Message ID Provider":
+  test "Non-empty string":
+    let message = Message(data: "Hello, world!".toBytes())
+    let result = defaultMessageIdProvider(message)
+    let expected = MDigest[256].fromHex("315f5bdb76d078c43b8ac0064e4a0164612b1fce77c869345bfc94c75894edd3")
+    check:
+      result.isOk()
+      result.get() == expected.data
+
+  test "Empty string":
+    let message = Message(data: "".toBytes())
+    let result = defaultMessageIdProvider(message)
+    let expected = MDigest[256].fromHex("E3B0C44298FC1C149AFBF4C8996FB92427AE41E4649B934CA495991B7852B855")
+    check:
+      result.isOk()
+      result.get() == expected.data
+  
+  test "Empty array":
+    let message = Message(data: @[])
+    let result = defaultMessageIdProvider(message)
+    let expected = MDigest[256].fromHex("E3B0C44298FC1C149AFBF4C8996FB92427AE41E4649B934CA495991B7852B855")
+    check:
+      result.isOk()
+      result.get() == expected.data
--- a/tests/waku_relay/test_protocol.nim
+++ b/tests/waku_relay/test_protocol.nim
@@ -1,0 +1,1367 @@
+{.used.}
+
+import
+  std/[options, sequtils, strutils],
+  stew/shims/net as stewNet,
+  testutils/unittests,
+  chronicles,
+  chronos,
+  libp2p/protocols/pubsub/[
+      pubsub,
+      gossipsub
+  ],
+  libp2p/[
+    multihash,
+    stream/connection,
+    switch
+  ],
+  ./crypto_utils,
+  std/json
+
+import
+  ../../../waku/[
+    node/peer_manager,
+    waku_relay/protocol,
+    waku_relay,
+    waku_core
+  ],
+  ../testlib/[
+    wakucore,
+    testasync,
+    testutils,
+    futures,
+    sequtils
+  ],
+  ./utils,
+  ../resources/payloads
+
+
+suite "Waku Relay":
+  var messageSeq {.threadvar.}: seq[(PubsubTopic, WakuMessage)]
+  var handlerFuture {.threadvar.}: Future[(PubsubTopic, WakuMessage)]
+  var simpleFutureHandler {.threadvar.}: WakuRelayHandler
+  var switch {.threadvar.}: Switch
+  var peerManager {.threadvar.}: PeerManager
+  var node {.threadvar.}: WakuRelay
+  var remotePeerInfo {.threadvar.}: RemotePeerInfo
+  var peerId {.threadvar.}: PeerId
+  var contentTopic {.threadvar.}: ContentTopic
+  var pubsubTopic {.threadvar.}: PubsubTopic
+  var pubsubTopicSeq {.threadvar.}: seq[PubsubTopic]
+  var testMessage {.threadvar.}: string
+  var wakuMessage {.threadvar.}: WakuMessage
+
+  asyncSetup:
+    messageSeq = @[]
+    handlerFuture = newPushHandlerFuture()
+    simpleFutureHandler = proc(
+      topic: PubsubTopic, msg: WakuMessage
+    ): Future[void] {.async, closure, gcsafe.} =
+      messageSeq.add((topic, msg))
+      handlerFuture.complete((topic, msg))
+
+    switch = newTestSwitch()
+    peerManager = PeerManager.new(switch)
+    node = await newTestWakuRelay(switch)
+
+    testMessage = "test-message"
+    contentTopic = DefaultContentTopic
+    pubsubTopic = DefaultPubsubTopic
+    pubsubTopicSeq = @[pubsubTopic]
+    wakuMessage = fakeWakuMessage(testMessage, pubsubTopic)
+
+    await allFutures(switch.start(), node.start())
+
+    remotePeerInfo = switch.peerInfo.toRemotePeerInfo()
+    peerId = remotePeerInfo.peerId
+
+  asyncTeardown:
+    await allFutures(switch.stop(), node.stop())
+
+  suite "Subscribe":
+    asyncTest "Publish without Subscription":
+      # When publishing a message without being subscribed
+      discard await node.publish(pubsubTopic, wakuMessage)
+
+      # Then the message is not published
+      check:
+        not await handlerFuture.withTimeout(3.seconds)
+
+    asyncTest "Publish with Subscription (Network Size: 1)":
+      # When subscribing to a Pubsub Topic
+      discard node.subscribe(pubsubTopic, simpleFutureHandler)
+
+      # Then the node is subscribed
+      check:
+        node.isSubscribed(pubsubTopic)
+        node.subscribedTopics == pubsubTopicSeq
+
+      # When publishing a message
+      discard await node.publish(pubsubTopic, wakuMessage)
+
+      # Then the message is published
+      assert (await handlerFuture.withTimeout(3.seconds))
+      let (topic, msg) = handlerFuture.read()
+      check:
+        topic == pubsubTopic
+        msg == wakuMessage
+    
+    asyncTest "Pubsub Topic Subscription (Network Size: 2, only one subscribed)":
+      # Given a second node connected to the first one
+      let 
+        otherSwitch = newTestSwitch()
+        otherNode = await newTestWakuRelay(otherSwitch)
+
+      await allFutures(otherSwitch.start(), otherNode.start())
+      let otherRemotePeerInfo = otherSwitch.peerInfo.toRemotePeerInfo()
+      check await peerManager.connectRelay(otherRemotePeerInfo)
+
+      var otherHandlerFuture = newPushHandlerFuture()
+      proc otherSimpleFutureHandler(topic: PubsubTopic, message: WakuMessage) {.async, gcsafe.} =
+        otherHandlerFuture.complete((topic, message))
+
+      # When subscribing the second node to the Pubsub Topic
+      discard otherNode.subscribe(pubsubTopic, otherSimpleFutureHandler)
+
+      # Then both nodes are subscribed
+      check:
+        not node.isSubscribed(pubsubTopic)
+        node.subscribedTopics != pubsubTopicSeq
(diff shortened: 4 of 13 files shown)


tests/waku_relay/test_protocol.nim
```diff
@@ -0,0 +1,1288 @@
...
+        node.isSubscribed(pubsubTopic)
+        node.subscribedTopics == pubsubTopicSeq
+      let fromOtherWakuMessage = fakeWakuMessage("fromOther")
+      discard await node.publish(pubsubTopic, fromOtherWakuMessage)
+
+      # When refreshing the subscription
+      var otherHandlerFuture = newPushHandlerFuture()
+      proc otherSimpleFutureHandler(topic: PubsubTopic, message: WakuMessage) {.async, gcsafe.} =
+        otherHandlerFuture.complete((topic, message))
+      # FIXME: Refreshing a subscription triggers a SEGFAULT.
```
AlejandroCabeza, line 250: Related Ticket: https://github.com/waku-org/nwaku/issues/2114

tests/waku_relay/test_protocol.nim
```diff
@@ -0,0 +1,1288 @@
...
+
+      # When sending multiple messages from node
+      let
+        msg1 = fakeWakuMessage("msg1", pubsubTopic)
+        msg2 = fakeWakuMessage("msg2", pubsubTopic)
+        msg3 = fakeWakuMessage("msg3", pubsubTopic)
+        msg4 = fakeWakuMessage("msg4", pubsubTopic)
+      
+      discard await node.publish(pubsubTopic, msg1)
+      # FIXME: Test SEGFAULTS when publishing after first publish in pubsub.nim:handleData:317
```
AlejandroCabeza, line 1116: Not the same reason, but the error's origin lies in the same place as: https://github.com/waku-org/nwaku/issues/2114

tests/waku_relay/test_protocol.nim
```diff
@@ -0,0 +1,1288 @@
...
+      check:
+        node.subscribedTopics == pubsubTopicSeq
+        otherNode.subscribedTopics == pubsubTopicSeq
+      
+      await sleepAsync(500.millis)
+
+      # Given other node is stopped and restarted
+      await allFutures(otherSwitch.stop(), otherNode.stop())
+      await allFutures(otherSwitch.start(), otherNode.start())
+      # FIXME: Once stopped and started, nodes are not considered connected, nor do they reconnect after running connectRelay, as below
```
AlejandroCabeza, line 1172: Needs investigating.

tests/waku_relay/test_protocol.nim
```diff
@@ -0,0 +1,1288 @@
...
+
+      discard otherNode.subscribe(pubsubTopic, otherSimpleFutureHandler)
+      discard node.subscribe(pubsubTopic, simpleFutureHandler)
+      check:
+        node.subscribedTopics == pubsubTopicSeq
+        otherNode.subscribedTopics == pubsubTopicSeq
+      
+      await sleepAsync(500.millis)
+
+      # FIXME: Inconsistent behaviour with Filter protocol.
```
AlejandroCabeza, line 1259: Filter is able to receive messages even if it has been stopped.

tests/waku_relay/test_protocol.nim
```diff
@@ -0,0 +1,1303 @@
...
```
AlejandroCabeza, line 1274: Filter is able to receive messages even if it has been stopped.

tests/waku_relay/crypto_utils.nim
```diff
@@ -0,0 +1,44 @@
+import nimcrypto
+
+
+proc cfbEncode*(key: string, iv: string, data: string): seq[byte] =
```
vpavlin, line 4: Why is this called `encode` if it actually `encrypts` ? (same for the proc below)

tests/waku_relay/test_protocol.nim
```diff
@@ -0,0 +1,1303 @@
...
+        otherNode.isSubscribed(pubsubTopic)
+        otherNode.subscribedTopics == pubsubTopicSeq
+
+      await sleepAsync(500.millis)
+
+      # When publishing a message in node
+      let fromOtherWakuMessage = fakeWakuMessage("fromOther")
+      discard await node.publish(pubsubTopic, fromOtherWakuMessage)
+
+      # Then the message is published only in both nodes
```
vpavlin, line 199: I think the comment has a copy&paste error:)

tests/waku_relay/test_protocol.nim
```diff
@@ -0,0 +1,1303 @@
...
+      await allFutures(otherSwitch.stop(), otherNode.stop(), anotherSwitch.stop(), anotherNode.stop())
+
+  suite "Unsubscribe":
+    asyncTest "Without Subscription":
+      # Given an external topic handler
+      let 
+        otherSwitch = newTestSwitch()
+        otherNode = await newTestWakuRelay(otherSwitch)
+      await allFutures(otherSwitch.start(), otherNode.start())
+      let otherTopicHandler: TopicHandler = otherNode.subscribe(pubsubTopic, simpleFutureHandler)
```
vpavlin, line 720: Why do you subscribe with `otherNode` and then checking `node` - what would be the potential side-effect of the `otherNode` being subscribed?

tests/waku_relay/test_protocol.nim
```diff
@@ -0,0 +1,1303 @@
...
+      # Given a node without subscriptions
+      check node.subscribedTopics == []
+
+      # When unsubscribing from all pubsub topics
+      node.unsubscribeAll(pubsubTopic)
+      
+      # Then the node is still not subscribed
+      check node.subscribedTopics == []
+
+    asyncTest "Single Node with Single Pubsub Topic":
```
vpavlin, line 781: Is this a duplicate of the one on L737?

tests/waku_relay/test_protocol.nim
```diff
@@ -0,0 +1,1303 @@
...
+      # Then the message is received in both nodes
+      check:
+        await handlerFuture.withTimeout(3.seconds)
+        not await otherHandlerFuture.withTimeout(3.seconds)
+        (pubsubTopic, msg6) == handlerFuture.read()
+
+      # Finally stop the other node
+      await allFutures(otherSwitch.stop(), otherNode.stop())
+
+    asyncTest "Multiple messages at once":
```
vpavlin, line 1083: Are you actually testing the protocol or testing your handler (and whether it successfully adds to `seq`)?

Because this is not "sending messages at once", this is sending them sequentially and the only change from other tests is custom handled adding to seq

tests/waku_relay/test_protocol.nim
vpavlin: Does this actually depend on persistence of peer store? (Just a hunch, but maybe something to consider)...although peer store is probably note even enabled in this case

vpavlin:
I am mildly confused about this PR - the title say `waku-filter` but all the tests are testing `relay`

tests/waku_relay/crypto_utils.nim
AlejandroCabeza: These were sourced (and slightly adapted) from `nimcrypto/examples/cfb.nim`, where they used the `encode` and `decode` keywords in the respective variables. Thus, I assumed those keywords would be understood.

tests/waku_relay/test_protocol.nim
AlejandroCabeza: Indeed 😅 You repeat one thing too many times and you start missing obvious errors

tests/waku_relay/test_protocol.nim
```diff
@@ -0,0 +1,1303 @@
...
+      let (otherTopic1, otherMsg1) = otherHandlerFuture.read()
+      check:
+        topic1 == pubsubTopic
+        msg1 == encryptedWakuMessage
+        otherTopic1 == pubsubTopic
+        otherMsg1 == encryptedWakuMessage
+      
+      # When decrypting the message
+      let 
+        decryptedText = cfbDecode(key, iv, msg1.payload)
```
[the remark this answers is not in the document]
AlejandroCabeza: True, updating :)

tests/waku_relay/test_protocol.nim
AlejandroCabeza: To run an `unsubscribe` method I need a `TopicHandler`, which you get as return value from running `subscribe`. Thus, to test "Unsubscribe without subscription" I need a valid `TopicHandler` that the target node is unrelated to.

tests/waku_relay/test_protocol.nim
AlejandroCabeza: Nope, one tests `unsubscribe` while the other `unsubscribeAll`.

AlejandroCabeza:
> I am mildly confused about this PR - the title say `waku-filter` but all the tests are testing `relay`

That was a typo on my side, updating it.

tests/waku_relay/test_protocol.nim
AlejandroCabeza: Unfortunately, this is the closest I managed to get to safely testing "send at once" because of an issue with the future (used for assertions) in the handler. The problem is calling `complete` on a `completed` future causes a `SEGFAULT`.

I found two ways of proceeding with this test:

1. Remove the future. This is the ideal solution. With this I'd be able to "send at once" (without any awaits) all messages, and after that, check whether they exist in the `seq`. The problem is, because of asynchronicity, I need to await an arbitrary number of seconds (which is not ideal for tests). Also, I can't remember off the top of my head, but for some reason, this didn't work sometimes.

2. Add the future. This works flawlessly, but it doesn't actually check for "send at once" because, as you mentioned, we await before sending each messages.

Any suggestion?

tests/waku_relay/test_protocol.nim
```diff
@@ -0,0 +1,1302 @@
...
+      check await peerManager.connectRelay(otherRemotePeerInfo)
+
+      var otherHandlerFuture = newPushHandlerFuture()
+      proc otherSimpleFutureHandler(topic: PubsubTopic, message: WakuMessage) {.async, gcsafe.} =
+        otherHandlerFuture.complete((topic, message))
+
+      # When subscribing the second node to the Pubsub Topic
+      discard otherNode.subscribe(pubsubTopic, otherSimpleFutureHandler)
+
+      # Then both nodes are subscribed
```
gabrielmer, line 126: Might be an error? isn't here only on node subscribed?

tests/waku_relay/test_protocol.nim
```diff
@@ -0,0 +1,1302 @@
...
+        otherMessage1 == fromOtherWakuMessage
+      
+      # When publishing a message in the other node
+      handlerFuture = newPushHandlerFuture()
+      otherHandlerFuture = newPushHandlerFuture()
+      let fromNodeWakuMessage = fakeWakuMessage("fromNode")
+      discard await node.publish(pubsubTopic, fromNodeWakuMessage)
+      discard await otherNode.publish(pubsubTopic, fromNodeWakuMessage)
+
+      # Then the message is published only in the subscribed node
```
gabrielmer, line 220: Here the message is published in both nodes (both are subscribed)

tests/waku_relay/test_protocol.nim
```diff
@@ -0,0 +1,1302 @@
...
+      await allFutures(otherSwitch.stop(), otherNode.stop())
+
+    asyncTest "Refreshing subscription":
+      # Given a subscribed node
+      node.subscribe(pubsubTopic, simpleFutureHandler)
+      check:
+        node.isSubscribed(pubsubTopic)
+        node.subscribedTopics == pubsubTopicSeq
+      let fromOtherWakuMessage = fakeWakuMessage("fromOther")
+      discard await node.publish(pubsubTopic, fromOtherWakuMessage)
```
gabrielmer, line 244: Shouldn't it be `fromNodeWakuMessage`? There's no other node in this test. Same for the other names in this test that use the word other.

tests/waku_relay/test_protocol.nim
gabrielmer: How can messages be received and sent with the node and the switch stopped? I'm missing probably missing something :)

tests/waku_relay/test_protocol.nim
```diff
@@ -0,0 +1,1302 @@
...
+      
+      # Then the message is received in both nodes
+      check:
+        await handlerFuture.withTimeout(3.seconds)
+        await otherHandlerFuture.withTimeout(3.seconds)
+        (pubsubTopic, msg2) == handlerFuture.read()
+        (pubsubTopic, msg2) == otherHandlerFuture.read()
+      
+      # Finally stop the other node
+      await allFutures(otherSwitch.stop(), otherNode.stop())
```
gabrielmer, line 1302: I think otherSwitch and otherNode have already been stopped in line 1275

tests/waku_relay/test_protocol.nim
```diff
@@ -0,0 +1,1302 @@
...
+      # When sending the 1025KiB message
+      handlerFuture = newPushHandlerFuture()
+      otherHandlerFuture = newPushHandlerFuture()
+      discard await node.publish(pubsubTopic, msg6)
+
+      # Then the message is received in both nodes
+      check:
+        await handlerFuture.withTimeout(3.seconds)
+        not await otherHandlerFuture.withTimeout(3.seconds)
+        (pubsubTopic, msg6) == handlerFuture.read()
```
gabrielmer, line 1077: Same here

gabrielmer (commented):
Thanks so much, looks great! :)
Left some comments/questions

tests/waku_relay/test_protocol.nim
AlejandroCabeza: Here you go, [issue comment](https://github.com/waku-org/nwaku/issues/2178#issuecomment-1791338580) that answers your question.

tests/waku_relay/test_protocol.nim
AlejandroCabeza: I can't, because there's just future, and it needs to be awaited and resetted sequentially. Otherwise, you'd be calling `complete` on a `completed` future, which triggers the aforementioned `SEGFAULT`.

tests/waku_filter_v2/test_all.nim
```diff
@@ -2,5 +2,4 @@
 
 import
   ./test_waku_client,
-  ./test_waku_filter_protocol,
-  ./test_waku_filter
```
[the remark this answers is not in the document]
AlejandroCabeza: Those were files that don't exist anymore, containing tests that were replaced (for another files).

tests/waku_relay/test_message_id.nim
```diff
@@ -0,0 +1,37 @@
...
+  ../../../waku/waku_relay/message_id,
+  ../testlib/sequtils
+
+
+suite "Message ID Provider":
+  test "Non-empty string":
+    let message = Message(data: "Hello, world!".toBytes())
+    let result = defaultMessageIdProvider(message)
+    let expected = MDigest[256].fromHex("315f5bdb76d078c43b8ac0064e4a0164612b1fce77c869345bfc94c75894edd3")
+    check:
```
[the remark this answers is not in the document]
AlejandroCabeza: It actually does! It avoids the risk that, would the sha256 algorithm fail at some point for any reason, we would still detect those anomalies.

By having the `expected` value be caculated by a function (the same one used for the `result`), we would not be checking for any issues in the hash generation; only for failures between the generation and encapsulation in the `MDigest` object.

tests/waku_relay/test_protocol.nim
AlejandroCabeza: Indeed, too much copy-pasting, thanks.

tests/waku_relay/test_protocol.nim
AlejandroCabeza: You're right, I don't know why I named it that way.

tests/waku_relay/test_protocol.nim
```diff
@@ -0,0 +1,1302 @@
...
+      
+      var anotherHandlerFuture2 = newPushHandlerFuture()
+      proc anotherSimpleFutureHandler2(topic: PubsubTopic, message: WakuMessage) {.async, gcsafe.} =
+        anotherHandlerFuture2.complete((topic, message))
+      
+      discard otherNode.subscribe(pubsubTopic, otherSimpleFutureHandler1)
+      discard otherNode.subscribe(pubsubTopicC, otherSimpleFutureHandler2)
+      discard anotherNode.subscribe(pubsubTopicB, anotherSimpleFutureHandler1)
+      discard anotherNode.subscribe(pubsubTopicC, anotherSimpleFutureHandler2)
+      await sleepAsync(500.millis)
```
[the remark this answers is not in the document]
AlejandroCabeza: Subscriptions are async operations, and there's no way (that I know of) to smartly await until they are subscribed; hence an arbitrary small explicit `await`.

tests/waku_relay/test_protocol.nim
```diff
@@ -0,0 +1,1302 @@
...
+      
+      await sleepAsync(500.millis)
+
+      # Given some valid payloads
+      let
+        msg1 = fakeWakuMessage(contentTopic=contentTopic, payload=getByteSequence(1024)) # 1KiB
+        msg2 = fakeWakuMessage(contentTopic=contentTopic, payload=getByteSequence(10*1024)) # 10KiB 
+        msg3 = fakeWakuMessage(contentTopic=contentTopic, payload=getByteSequence(100*1024)) # 100KiB
+        msg4 = fakeWakuMessage(contentTopic=contentTopic, payload=getByteSequence(1023*1024)) # 1MiB - 1B -> Max Size (Inclusive Limit)
+        msg5 = fakeWakuMessage(contentTopic=contentTopic, payload=getByteSequence(1024*1024)) # 1MiB      -> Max Size (Exclusive Limit)
```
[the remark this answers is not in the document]
AlejandroCabeza: Yes, I realised this later so I changed it in the following PR.

tests/waku_relay/test_protocol.nim
AlejandroCabeza: Oh, yeah! Did that out of custom

tests/waku_relay/test_protocol.nim
AlejandroCabeza: I opened a ticket for this.

tests/waku_relay/test_protocol.nim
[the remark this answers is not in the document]
gabrielmer: Makes sense, and normal `await` in each of the subscribe statements doesn't work? or is it important to have them done concurrently? If it's for efficiency compared to awaiting each statement separately, not sure it's worth the possible indeterminism. Let me know what you think :))

gabrielmer (approved):
LGTM, thank you!

tests/waku_relay/test_protocol.nim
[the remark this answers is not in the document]
AlejandroCabeza: Actually, I said something that's not right: `subscribe` is a sync operation. The thing is, for some reason I don't know, existing tests already had an await in place, and they don't work without it; which leads me to think there's some `async` behaviour to "formalise" the subscription and there's no way to smartly await for it.

tests/waku_relay/test_protocol.nim
vpavlin: Interesting:D I guess that is why I am not a QE as I would never think about testing this:D

vpavlin (approved):
LGTM

PR #2108, opened 2023-10-04
chore(rln): add more hardcoded memberhips to static group

alrevuelta:
Description:
* We currently had a hardcoded tree with 100 memberships for the static version of rln.
* This PR adds more harcoded memberships, a total of 10000.
* Useful for simulations and tests.
* New file is 2.7Mb which should be okay.

```
works
./build/wakunode2 --rln-relay-dynamic=false --rln-relay=true --rln-relay-membership-index=9999

does not work (we just have 10000)
./build/wakunode2 --rln-relay-dynamic=false --rln-relay=true --rln-relay-membership-index=10001
```


Ivansete-status (approved):
Thanks for that!

I approve because now we need two approvers. However, my approval shouldn't have lot of weight :P

A couple of comments:
Isn't the next constant redundant? Maybe it can be obtained from the `.len` of the array?
`StaticGroupSize* = 10000`

The next comment should be updated to reflect 10000:
  `#  StaticGroupKeys is a static list of 100 membership keys in the form of (identity key, identity commitment)`

In case we are interested in having only private constants, maybe it worth to just `include` that "constants" module instead of `import`ing it.

rymnc (approved):
LGTM!!

alrevuelta:
@Ivansete-status indeed, fixed 5a8e020381c7c5411a98c0f3a407507e2e184e22

>  Isn't the next constant redundant? Maybe it can be obtained from the .len of the array?

A bit yeap, but it serves as a way to know how many static memberships there are. There are checks ensuring that == .len so imho its nice to have.

PR #2111, opened 2023-10-05
fix: sqlite limited delete query bug

ABresting:
# Description
Observed a bug while testing the retention policy where the database, if filled beyond its size limit, reaches a point where the deletion query `proc deleteOldestMessagesNotWithinLimitQuery(table: string, limit: int): SqlQueryStr = `ceases to function. The database size remains unchanged, leading to an endless loop in the deletion process, especially when the deletion is expected to be achieved via a loop.

This impacts the retention policy process. Turns out that the sqlite DB query assumes that the `id` column alone can uniquely identify rows in the table. If two different rows have the same `id` but different `storedAt` values, then using only `id` in the subquery may not produce the intended results. Given the schema has a composite primary key (`storedAt, id, pubsubTopic`), just checking against id might not be precise enough.

For the size-based retention policy, there isn't a direct correlation between the database's size and its number of rows. Also, deletion based on page count isn't feasible. To manage database size effectively:

Ascertain the total number of messages/rows.
Delete 20% of the messages and then perform a database Vacuum.
Confirm if the adjusted database size aligns with the intended target size.
If not, iterate the process.

So in this process, if the database size increases beyond a limit then a forever loop is encountered. 

# Changes

Only the query has been modified.

- [x] Able to delete the outdated database rows correctly

## Issue

#1885

### DIFF
--- a/waku/waku_archive/driver/sqlite_driver/queries.nim
+++ b/waku/waku_archive/driver/sqlite_driver/queries.nim
@@ -181,9 +181,9 @@ proc deleteMessagesOlderThanTimestamp*(db: SqliteDatabase, ts: int64):
 ## Delete oldest messages not within limit
 
 proc deleteOldestMessagesNotWithinLimitQuery(table: string, limit: int): SqlQueryStr =
-  "DELETE FROM " & table & " WHERE id NOT IN (" &
-  " SELECT id FROM " & table &
-  " ORDER BY storedAt DESC" &
+  "DELETE FROM " & table & " WHERE (storedAt, id, pubsubTopic) NOT IN (" &
+  " SELECT storedAt, id, pubsubTopic FROM " & table &
+  " ORDER BY storedAt DESC, id DESC" &
   " LIMIT " & $limit &
   ");"
 


Ivansete-status (approved):
LGTM! Thanks!
Very well explained the issue.

Btw, this issue would happen if we start applying the upcoming size-based retention policy, right? 

We never faced it because we are not applying a loop in the currently existing retention policies, right?

ABresting:
> LGTM! Thanks! Very well explained the issue.
> 
> Btw, this issue would happen if we start applying the upcoming size-based retention policy, right?
Yes
> 
> We never faced it because we are not applying a loop in the currently existing retention policies, right?
Yes that's correct!

jm-clius (approved):
Makes sense to me. This shows another reason why we need a unique "index" column in the schema that is derived according to the deterministic message hashing algorithm here: https://rfc.vac.dev/spec/14/#deterministic-message-hashing
This deterministic message hashing must be applied everywhere as the authoritative message ID. The existing `id` column can likely be dropped.

ABresting:
> lgtm thanks.
> 
> Can we write a unit test to catch this? Don't want to block so another PR is fine.

Actually, this was caught while extending the retention policy test case, I stress tested (not on GB scale but slightly more than 2x the original data), how about keeping the existing test case that broke it?

PR #2113, opened 2023-10-05
fix: Properly divide num_bytes by 1024 to get KB

Ivansete-status:
# Description

Fix conversion from bytes to kilobytes in `waku_histogram_message_size.observe(msgSizeKB)`

### DIFF
--- a/waku/node/waku_node.nim
+++ b/waku/node/waku_node.nim
@@ -193,7 +193,7 @@ proc registerRelayDefaultHandler(node: WakuNode, topic: PubsubTopic) =
       receivedTime=getNowInNanosecondTime(),
       payloadSizeBytes=msg.payload.len
 
-    let msgSizeKB = msg.payload.len/1000
+    let msgSizeKB = msg.payload.len/1024
 
     waku_node_messages.inc(labelValues = ["relay"])
     waku_histogram_message_size.observe(msgSizeKB)


gabrielmer (approved):
Looks good! Thanks :):)

ABresting (approved):

alrevuelta:
> fix: Properly divide num_bytes by 1024 to get KB

As it is its actually correct.
* /1000 KB
* /1024 KiB

So "Properly divide num_bytes by 1024 to get **KiB**"

Unsure though if we want KB or KiB.

Ivansete-status:
> > fix: Properly divide num_bytes by 1024 to get KB
> 
> As it is its actually correct.
> 
> * /1000 KB
> * /1024 KiB
> 
> So "Properly divide num_bytes by 1024 to get **KiB**"
> 
> Unsure though if we want KB or KiB.

Very good point @alrevuelta !
That definitely breaks the ambiguity. Nevertheless, I don't know if this terminology is fully extended.

Ivansete-status:
I will dismiss this PR for now.
As a mental note, I think it is interesting that all of us consider that size unit convention:
```
1000 bytes == 1 KB
1024 bytes == 1 KiB
```

arnetheduck:
why are kilobytes being used? there's no advantage in metrics really where the UI can scale to whatever size if bytes are used instead

Ivansete-status:
> why are kilobytes being used? there's no advantage in metrics really where the UI can scale to whatever size if bytes are used instead

Good point! We could only use bytes and then let the Grafana accommodate the graph properly depending on the dimensions.

PR #2118, opened 2023-10-09
chore(cbindings): avoid using global var in libwaku.nim

Ivansete-status:
### Description
Applying Jacek's recommendation which is to prevent using global variable. That way we provide more flexibility to the library users. For example, we could start multiple `wakunode` instances within the same app.

Original recommendation: https://github.com/waku-org/nwaku/pull/1865#discussion_r1282722954

### Changes
- [x] Remove global variable from `libwaku.nim`.
- [x] Extract the `ctx` variable to outside the library so that the Context can be passed to any other libwaku function.
- [x] Addition of `int callerRet` parameter to `WakuCallback`. This parameter allows the caller to indicate whether the callback call was a successful or erroneous one.
- [x] Adding `void* userData` parameter to every library function. This is useful to attach closure reference in Rust.
- [x] Adapting the C code example.

### Issue
https://github.com/waku-org/nwaku/issues/1878


NagyZoltanPeter (approved):
Pretty cool step ahead! Thank you!

I think it is quite well polished overall.
I just have some notice for the future to think of but handle it as a personal flavour.

I think in general library internal memory shall not be exposed to client. I personally more like to name it as handle (e.g. ContextHandle that can be checked if that maps to a real object allocated in the library, rather believe caller gives us the right pointer back). Main reason that any kind of mess in the usage will point to library failure, like crash or something.

Similar but even harder to solve in C the thingy with returning strings through callback.
We should trust client code that our message pointer will not outlive the callback call or GC (does GC enabled in libwaku?) can free or reuse it causing weird working in the user code.

In the example program I would encourage use strncpy not to trust ever 0 ending.
Also malloc call is better "malloc((len+1)*sizeof(char))" in case sometime ever it turns to wchar_t. But this one is really for habits, not a bug in this particular case.

Ivansete-status:
Thanks for the comment and for checking it! :fire: 

I reply to a couple of points:

> Similar but even harder to solve in C the thingy with returning strings through callback. We should trust client code that our message pointer will not outlive the callback call or GC (does GC enabled in libwaku?) can free or reuse it causing weird working in the user code.

Good point! I'll revisit all points where we return information through the callback approach to make sure the strings are properly allocated, live during the callback call, and are freed afterwards.

> In the example program I would encourage use strncpy not to trust ever 0 ending. Also malloc call is better "malloc((len+1)*sizeof(char))" in case sometime ever it turns to wchar_t. But this one is really for habits, not a bug in this particular case.

Very good point indeed! Something to enhance but not urgent.

library/libwaku.h
```diff
@@ -15,45 +15,73 @@
...
+typedef void (*WakuCallBack) (int callerRet, const char* msg, size_t len);
+
+// Initializes the waku library and returns a pointer to the Context.
+void* waku_init(WakuCallBack callback,
+                void* userData);
 
 // Creates a new instance of the waku node.
 // Sets up the waku node from the given configuration.
-int waku_new(const char* configJson, WakuCallBack onErrCb);
+int waku_new(void* ctx,
```
arnetheduck, line 26: for the context, a useful trick is to define an opaque type for it (`typedef struct WakuContext WakuContext;`) and pass around `WakuContext*` instead of `void*` - this documents intent and gives some rudimentary type safety (even if it's limited in C).

richard-ramos (approved):

Issue #1797, opened 2023-06-13
feat: provide a way to define advertised addresses


filoozom:
### Problem

I'm running `nwaku` nodes behind an SSL gateway that manages all the encryption and certificates. Basically, the `nwaku` node listens to WebSocket requests without any encryption, and I add the right multi address via `--ext-multiaddr`.

However, there's no way to disable the default `/ip4/.../tcp/8000/ws/p2p/...` address, meaning that `js-waku` will try to connect to it for example.

I end up with the following advertised addresses:
```
- /ip4/.../tcp/60001/p2p/peerId
- /dns4/.../tcp/443/wss/p2p/peerId (added with `--ext-multiaddr`)
- /ip4/.../tcp/8000/ws/p2p/peerId (invalid, port not exposed, should be removed)
```

### Suggested solution

Add a `--advertised-multiaddr` flag that makes it possible to override all advertised addresses. In my example, I'd need to remove `--ext-multiaddr` and to add the following flags:
```
--ext-multiaddr:/dns4/fqdn/tcp/443/wss
--ext-multiaddr:/ip4/ip/tcp/60001
```

I could also then presumably remove `--nat:extip:ip`.

### Alternatives considered

An alternative that at least doesn't really break this is to also advertise the WebSocket address and forward that on the reverse proxy, so that `/ip4/.../tcp/8000/ws/p2p/...` actually works. But that's not what I want.

An alternative solution to implement could be not to add `/ip4/.../tcp/8000/ws/p2p/...` by default when enabling WebSocket without making it much more flexible.

### Additional context

There are many reasons why one could want this:
- Not requiring heavy changes to an existing infrastructure (Nomad, Kubernetes, Docker Swarm, ...)
- If I trust my infra's SSL capabilities, and see no need to involve some additional SSL "dependencies" that might be less secure or slower
- If I want my certificates to be managed automatically by my existing infra (Let's Encrypt integrations for example)
- If I don't want the entire service to restart just to rotate keys
- It might not be possible to do this without a reverse proxy (DAppNode?)

Being able to precisely confuse the advertised addresses also makes everything more flexible. For example, I can have my `nwaku` node listen on port 60000 in a Docker container, VM or whatever, and NAT that to any IP:PORT combination without having to listen on that specific port and setting `--nat:extip:IP`

### Acceptance criteria

I can configure exactly which multi addresses are being advertised by my node.

Ivansete-status:
We may need to adjust around the next:

[...]

PR #2122, opened 2023-10-11
fix: consider WS extMultiAddrs before publishing host address

gabrielmer:
# Description

When the user runs a Waku node with the flag `--ext-multiaddr` using a WS/WSS address, our current logic does not notice we specified a WS/WSS address and incorrectly adds the default `wsHostAddress` to the `announcedAddresses`

# Changes

- [x] Checking for announced WS/WSS addresses in `extMultiAddrs` before announcing the default `wsHostAddress`
- [x] Adding test case accounting for the mentioned scenario 

## How to test

When WS/WSS is enabled and no WS/WSS ext-multiaddr provided, default WS/WSS address is published:
without_extma

When WS/WSS is enabled and a WS/WSS ext-multiaddr is provided, the default WS/WSS is not published:
with_extma

## Issue

closes [#1797](https://github.com/waku-org/nwaku/issues/1797)


gabrielmer:
Notice that currently the same case happens with external addresses that aren't WS/WSS: if an external address is set with the `--ext-multiaddr` flag and no `extIp` or `dns` are provided, both the host's bind address and the external multiaddress are announced.

We can add the same logic for that case. However I do think that it's better to require users to add `extIp` and `extPort` properly through its respective CLI flags instead of trying to infer them.

The way I see it is: `extIp` and `extPort` have their own flags which should be used. There's no flag for a separate WS/WSS address so it's either inferred from `extIp`/`dns` or passed through the `ext-multiaddr` flag.

Let me know what you think about this approach

waku/node/config.nim
```diff
@@ -59,6 +59,20 @@ proc formatListenAddress(inputMultiAdd: MultiAddress): MultiAddress =
...
+    isWss = ma.contains(multiCodec("wss")).get()
+
+  return isWs or isWss
+
+proc findWsAddress(extMultiAddrs: seq[MultiAddress]): Option[MultiAddress] =
+  for a in extMultiAddrs:
+    if isWsAddress(a):
+      return some(a)
+  
+  return none(MultiAddress)
```
Ivansete-status, line 74: I think this can be simpler. Apparently we only need to check where any MA contains ws or wss.

```suggestion
proc containsWsAddress(extMultiAddrs: seq[MultiAddress]): bool =
  return extMultiAddrs.filterIt( it.isWsAddress() ).len > 0
```
Then only use that proc

Ivansete-status (commented):
Thanks for this!
I just added a comment mentioning how we could do the same (not a big deal though.)

However, shall we have in the PR description or somewhere an example on how to validate that the issue is sorted out with that? Maybe a snippet log will help as well.

Cheers!

waku/node/config.nim
gabrielmer: Thank you! I'll add this change now, it's much more elegant :)

My intention of returning the MA instead of a Bool was that no extra computation has to be done to return the MA, so by returning more information "for free" the function could be reused in more cases in the future.

However, using `filterIt` the way you proposed is much better, simpler, and can be used in cases where you need the MA :)

vpavlin (approved):
LGTM, seems to correctly produce the list of listen addresses

NagyZoltanPeter (approved):
LGTM! Thank you!!!

PR #2123, opened 2023-10-12
chore: remove js-node tests as release candidate dependencies

gabrielmer:
# Description

Lightpush tests in `js-waku-node` tests have been failing for quite some time and are still a work in progress. Removing the dependency for the release-candidate workflow until the issue is addressed.

# Changes

- [x] removed `js-waku-node` tests from release-candidate workflows


vpavlin (approved):

Ivansete-status (approved):
Thanks for this!

gabrielmer:
@fbarbu15 please let us know in the future when `js-waku-node` tests are stabilized so we add again the dependencies

fbarbu15:
> @fbarbu15 please let us know in the future when `js-waku-node` tests are stabilized so we add again the dependencies

@gabrielmer  I have a [fix](https://github.com/waku-org/js-waku/pull/1645) for this for some time now, but I can't merge it because then the tests will fail on wakuorg/nwaku:v0.20.0 (which doesn't contain the change)

gabrielmer:
> > @fbarbu15 please let us know in the future when `js-waku-node` tests are stabilized so we add again the dependencies
> 
> @gabrielmer I have a [fix](https://github.com/waku-org/js-waku/pull/1645) for this for some time now, but I can't merge it because then the tests will fail on wakuorg/nwaku:v0.20.0 (which doesn't contain the change)

Makes sense! Is there anything already merged to master that would make it pass the tests in next releases? Or how can we make it compatible for next releases?

fbarbu15:
> > > @fbarbu15 please let us know in the future when `js-waku-node` tests are stabilized so we add again the dependencies
> > 
> > 
> > @gabrielmer I have a [fix](https://github.com/waku-org/js-waku/pull/1645) for this for some time now, but I can't merge it because then the tests will fail on wakuorg/nwaku:v0.20.0 (which doesn't contain the change)
> 
> Makes sense! Is there anything already merged to master that would make it pass the tests in next releases? Or how can we make it compatible for next releases?

Unfortunately no. For such cases where there is a fix that impacts tests (usually for issues found by the tests themselves) it's expected for the tests to fail on one version and fail on the other. 
The alternative is to put all kinds of conditions in the tests based on versions but I don't think that's a good idea because it's prone to mistakes and we could overlook actual issues. 
In this particular case, when there is a new nwaku release that includes this fix (2023-10-18 according to @chair28980's  comment), I will update the js-waku tests to use that one and also merge this [PR](https://github.com/waku-org/js-waku/pull/1645).

gabrielmer:
> Unfortunately no. For such cases where there is a fix that impacts tests (usually for issues found by the tests themselves) it's expected for the tests to fail on one version and fail on the other. The alternative is to put all kinds of conditions in the tests based on versions but I don't think that's a good idea because it's prone to mistakes and we could overlook actual issues. In this particular case, when there is a new nwaku release that includes this fix (2023-10-18 according to @chair28980's comment), I will update the js-waku tests to use that one and also merge this [PR](https://github.com/waku-org/js-waku/pull/1645).

Makes sense! Thank you :)

PR #2124, opened 2023-10-13
chore: Refactored openapi sepcifications and build doc

NagyZoltanPeter:
# Description
This PR is an effort towards a slim and comprehensive Rest API specification and documentation.
Please take it as a reasoning on choices we can have as well as a source of proposal. (In this manner please feel free to invite others to give opinions)

Current Waku node interface specifications are living next to each endpoint implementation and not really used to provide a real help for waku users on how they can access their node running.

To reach this I think best move is to merge these separate specifications and use as single source of truth.
Although from maintenance point of view i think it is better to have it well splitted up yet being able to organized in hierarchy.
I found unfortunate the OpenApi specification rules are a bit too restrictive on this but was able to reorganize under one openapi.yaml - as a root of the specification - while having endpoints defined alone.
Schema specification was also aimed to get common and eliminate duplications.

This approach is presented here.

There are few things to get decided.

- Is this the right place where Rest-api interface specification should live?
  - My proposal is the keep it in the nwaku repository as nwaku is the reference implementation of wakunode. Having it in a separate repository bears the risk to get forgotten updated.
  - Having them duplicated in go-waku also has the risks of branching from each other.
- Is this the right form of having a maintainable api specification?
- What form we should use it as documentation for waku users?
  - My proposal is to use it as a single source of truth and generate our doc page out of it right from nwaku master branch.
  - Where to fit this generated doc page?

For getting an impression I created a generated api doc under my gh pages:
https://nagyzoltanpeter.github.io/nwaku-api/index.html

- This one now is in my repository and fed from my [dev branch's head](https://raw.githubusercontent.com/waku-org/nwaku/chore/refactor-openapi/waku/waku_api/rest/doc/openapi.yaml).
  - This is intended to move where we decide to place such doc and shall be fed from master's head (or?)
- The renderer engine is from RapiDoc and used in a local copy in my repository. This has MIT license.
- Of course there is a lot of chance to make it nicer and customized as needed.

# Changes

- [X] All openapi.yaml files are moved under waku/waku_api/rest/doc
- [X] There is one openapi.yaml as a root.
- [X] Each endpoint yaml's are named accordingly and following the rules of Openapi spec referencing conventions
- [X] schema types is in one common file under .../rest/doc/schemas
- [X] No code changes done now
  - [X] There are some renaming in the spec (in store endpoint - agreed with @Ivansete-status - to get common naming)
  - [ ] This one should be follow up in later PR in the code as well. 

## Issue
[https://github.com/waku-org/nwaku/issues/2120](https://github.com/waku-org/nwaku/issues/2120)


fbarbu15:
Good job! This facilitates the creation of automated tests based on specs, utilizing OpenAPI tools. From [schema validators](https://github.com/python-openapi/openapi-schema-validator) to [automated test generation](https://github.com/schemathesis/schemathesis)

chaitanyaprem:
> Is this the right place where Rest-api interface specification should live?
    My proposal is the keep it in the nwaku repository as nwaku is the reference implementation of wakunode. Having it in a separate repository bears the risk to get forgotten updated.
    Having them duplicated in go-waku also has the risks of branching from each other.

I think we should place this spec in a separate repo under wak-org similar to how we are storing proto files at https://github.com/waku-org/waku-proto . 
This way, spec is maintained independent of implementations and each implementation can refer to a version of the spec which can be specified in the implementations documentation. 
The process to update REST API's should be first to update the spec, get it reviewed and then implement the spec. This is a process which i have seen followed across many orgs and works well.

waku/waku_api/rest/doc/schemas/apitypes.yaml
```diff
@@ -0,0 +1,189 @@
...
+    - multiaddr
+    - protocols
+  properties:
+    multiaddr:
+      type: string
+    protocols:
+      type: array
+      items:
+        type: object
+        required:
```
chaitanyaprem, line 13: Maybe it would be good to also show the shardInfo supported by a peer.

waku/waku_api/rest/doc/openapi.yaml
```diff
@@ -0,0 +1,66 @@
...
+     $ref: "./debugapi_version.yaml"
+  /health:
+    $ref: "./healthapi.yaml"
+  /lightpush/v1/message:
+    $ref: "./lightpushapi.yaml"
+  /relay/v1/messages/{pubsubTopic}:
+    $ref: "./relayapi_messages.yaml"
+  /relay/v1/subscriptions:
+    $ref: "./relayapi_subscriptions.yaml"
+  /relay/v1/auto/messages/{contentTopic}:
```
chaitanyaprem, line 44: Would be good to indicate somewhere the difference between relay and relay/auto API's, it is not very obvious by looking at URL paths.

waku/waku_api/rest/doc/openapi.yaml
```diff
@@ -0,0 +1,66 @@
...
+  - name: lightpush
+    description: Interface to request a message relay. See [19/WAKU2-LIGHTPUSH](https://rfc.vac.dev/spec/19/) RFC
+  - name: relay
+    description: Control of the relaying of messages. See [11/WAKU2-RELAY](https://rfc.vac.dev/spec/11/) RFC
+  - name: store
+    description: Retrieve the message history. See [13/WAKU2-STORE](https://rfc.vac.dev/spec/13/) RFC
+  - name: filter
+    description: Control of the content filtering. See [12/WAKU2-FILTER](https://rfc.vac.dev/spec/12/) RFC
+  - name: filter_legacy
+    description: Obsolate Filter interface kept for compatibility reason. Will be removed in future.
```
chaitanyaprem, line 28: Typo Obsolate --> Obsolete.

waku/waku_api/rest/doc/relayapi_auto_content_topic.yaml
```diff
@@ -0,0 +1,25 @@
+# /relay/v1/auto/messages/{contentTopic}:  # Note the plural in messages
```
chaitanyaprem, line 1: Same comment as for filter.

waku/waku_api/rest/doc/relayapi_auto_content_topic.yaml
```diff
@@ -0,0 +1,25 @@
...
+        type: string
+      description: The user ID
+  responses:
+    '200':
+      description: The latest messages on the polled topic.
+      content:
+        application/json:
+          schema:
+            $ref: './schemas/apitypes.yaml#/RelayGetMessagesResponse'
+    '4XX':
```
chaitanyaprem, line 22: Would 4XX be returned in case the node is not subscribed to that particular contentTopic?
Hope we have an error message being returned for all error responses indicating the specific error that occured.

waku/waku_api/rest/doc/relayapi_auto_messages.yaml
```diff
@@ -0,0 +1,19 @@
+# /relay/v1/auto/messages:  # Note the plural in messages
```
chaitanyaprem, line 1: Only a single message can be posted at once, URL should be `<>/message`

waku/waku_api/rest/doc/relayapi_auto_subscriptions.yaml
```diff
@@ -0,0 +1,50 @@
...
+  operationId: postSubscriptions
+  tags:
+    - relay
+  requestBody:
+    content:
+      application/json:
+        schema:
+          type array:
+          items:
+            $ref: './schemas/apitypes.yaml#/ContentTopic'
```
chaitanyaprem, line 14: We don't seem to provide an option to specify minPeers required to publish a message. This is available in the relay API via programming interface. 
It would be good to specify this either in postMessage or subscribe as well.

waku/waku_api/rest/doc/relayapi_messages.yaml
```diff
@@ -0,0 +1,51 @@
...
+      schema:
+        type: string
+      description: The user ID
+  responses:
+    '200':
+      description: The latest messages on the polled topic.
+      content:
+        application/json:
+          schema:
+            $ref: './schemas/apitypes.yaml#/RelayGetMessagesResponse'
```
chaitanyaprem, line 21: I am wondering if on the server side any chunking is done in case messages are too many and each message is large in size. 
AFAIK 5MB is the limit for REST API response size.

waku/waku_api/rest/doc/relayapi_subscriptions.yaml
```diff
@@ -0,0 +1,50 @@
...
+  tags:
+    - relay
+  requestBody:
+    content:
+      application/json:
+        schema:
+          type array:
+          items:
+            $ref: './schemas/apitypes.yaml#/PubsubTopic'
+  responses:
```
chaitanyaprem, line 15: How is partial error handled? 
Let's say if we request to subscribe to 2 topics and one of them fails and other succeeds, how is the response returned to caller? Better to include this information as well.

Similar case needs to be handled in all API's where multiple operations are executed on server side.

waku/waku_api/rest/doc/storeapi.yaml
```diff
@@ -0,0 +1,138 @@
...
+get:
+  summary: Gets message history
+  description: >
+    Retrieves WakuV2 message history. The returned history
+    can be potentially filtered by optional request parameters.
+  operationId: getMessageHistory
+  tags:
+    - store
+  parameters:
+    - name: peerAddr
```
chaitanyaprem, line 11: Would it be good to define peerAddr as a type, as this maybe requried to be used in other API's as well.

waku/waku_api/rest/doc/storeapi.yaml
```diff
@@ -0,0 +1,138 @@
...
+      schema:
+        type: string
+      description: >
+        The pubsub topic on which a WakuMessage is published.
+        If left empty, no filtering is applied.
+        It is also intended for pagination purposes.
+        It should be a URL-encoded string.
+      example: 'my%20pubsub%20topic'
+
+    - name: contentTopics
```
chaitanyaprem, line 32: Why not an array here? 
Wouldn't that be a better approach to take in multiple contentTopics as argument.
Also, we should reuse contentTopic type defined.

waku/waku_api/rest/doc/storeapi.yaml
```diff
@@ -0,0 +1,138 @@
...
+      in: query
+      schema: string
+      description: >
+        Comma-separated list of content topics. When specified,
+        only WakuMessages that are linked to any of the given
+        content topics will be delivered in the get response.
+        It should be a URL-encoded-comma-separated string.
+      example: 'my%20first%20content%20topic%2Cmy%20second%20content%20topic%2Cmy%20third%20content%20topic'
```
chaitanyaprem, line 40: Also, not sure why the example has spaces. 
Content-topics as i understand don't have spaces in them rather of the format "toychat/1/abc/proto". Example should reflect the same.

waku/waku_api/rest/doc/storeapi.yaml
```diff
@@ -0,0 +1,138 @@
...
+        type: string
+      description: >
+        Cursor field intended for pagination purposes.
+        Represents the Unix time in nanoseconds at which a message was generated.
+        It could be empty for retrieving the first page,
+        and will be returned from the GET response so that
+        it can be part of the next page request.
+      example: '1680590947000000000'
+
+    - name: storeTime
```
chaitanyaprem, line 76: Not sure if this and senderTime should be used. 
Because each store node in the network may store the message at a different time, whereas senderTime would be same. 
But maybe this is also more of a store protocol change.

waku/waku_api/rest/doc/storeapi.yaml
```diff
@@ -0,0 +1,138 @@
...
+        type: string
+      description: >
+        Cursor field intended for pagination purposes.
+        Represents the Unix time in nanoseconds at which a message was stored.
+        It could be empty for retrieving the first page,
+        and will be returned from the GET response so that
+        it can be part of the next page request.
+      example: '1680590945000000000'
+
+    - name: digest
```
chaitanyaprem, line 88: What would happen if i specify multiple cursor fields?
better to document that behaviour so that users of the API are clear as well.

waku/waku_api/rest/doc/storeapi.yaml
```diff
@@ -0,0 +1,138 @@
...
+      description: >
+        Cursor field intended for pagination purposes.
+        URL-base64-encoded string computed as a hash of the
+        a message content topic plus a message payload.
+        It could be empty for retrieving the first page,
+        and will be returned from the GET response so that
+        it can be part of the next page request.
+      example: 'Gc4ACThW5t2QQO82huq3WnDv%2FapPPJpD%2FwJfxDxAnR0%3D'
+
+    - name: pageSize
```
chaitanyaprem, line 101: If messageSize is too big to fit page size messages in a response, does the server truncate and chunk or send less than requested pageSize?

waku/waku_api/rest/doc/storeapi.yaml
```diff
@@ -0,0 +1,138 @@
...
+      in: query
```
chaitanyaprem, line 102: Also curious to know how paging works, does the user need to keep querying until empty response is returned?

waku/waku_api/rest/doc/schemas/apitypes.yaml
```diff
@@ -0,0 +1,189 @@
...
+          - protocol
+          - connected
```
chaitanyaprem, line 15: Connected need not be per protocol, this is per peer. 
I am assuming this indicates whether node is connected with this peer or not and not per protocol.

waku/waku_api/rest/doc/relayapi_auto_content_topic.yaml
chaitanyaprem, line 1: Another reason for not having contentTopic as a URL param is since contentTopic format recommends having `/`, a user will have to URLencode the contentTopic before invoking the API. It is non-standard but if we include contentTopic in requestBody then this issue can be avoided.

While testing with postman for example with contentTopic `/toychat/1/huilong/proto`, i had to pass `%2Ftoychat%2F1%2Fhuilong%2Fproto` in order for this to work.

NagyZoltanPeter: We had exactly this conversation in nwaku. 
So far our conclusion was, that even though it requires url encoding, having GET request in content body is against rest convention and Http recommendation. But I could live with it having to do so.
Major issue was that the nim-presto implementation we use is - in align with the http spec - forbids having content body for a GET request.
So in order to do so we need to modify our HttpServer implementation.

I'm not sure how it is in your go impl. Does that allows it?

waku/waku_api/rest/doc/schemas/apitypes.yaml
NagyZoltanPeter: Good idea!!!

waku/waku_api/rest/doc/schemas/apitypes.yaml
NagyZoltanPeter: Our idea was to group protocol and info under peers. Of course having a peer with no protocol does not make sense, hence the required.

waku/waku_api/rest/doc/storeapi.yaml
NagyZoltanPeter: Agree, I will incorporate.

waku/waku_api/rest/doc/relayapi_auto_content_topic.yaml
chaitanyaprem: In golang, the way we are implementing REST methods i don't see any limitation with this. 
Will test once to confirm though.

waku/waku_api/rest/doc/schemas/apitypes.yaml
chaitanyaprem: I do agree that protocol is required, but what i meant is connected status need not be at same level with protocol. 

What i am suggesting is that response should be in the following format. If you noticed the peer connection status is shown outside protocols. Also, one more query on this point, does this endpoint list all peers in the peerStore or only peers that we are currently connected to?

```
{
    "multiaddr":[
        "/ip4/10.0.0.133/tcp/60000/p2p/16Uiu2HAkymj3jUpK1eGaymZbGftwffVYtoMRgnd8sngC8sSJzabo", 
        "/ip4/127.0.0.1/tcp/60000/p2p/16Uiu2HAkymj3jUpK1eGaymZbGftwffVYtoMRgnd8sngC8sSJzabo"
    ],
    "connected": true,
    "protocols":[
        "/vac/waku/filter-push/2.0.0-beta1",
        "/vac/waku/relay/2.0.0"
    ]
}


```

waku/waku_api/rest/doc/schemas/apitypes.yaml
```diff
@@ -0,0 +1,189 @@
+WakuPeer:
+  type: object
+  required:
+    - multiaddr
+    - protocols
+  properties:
+    multiaddr:
+      type: string
```
chaitanyaprem, line 8: Also, this multiaddr should be an array of multiaddresses for the peer...as a peer can have more than 1 multiaddress advertising.

chaitanyaprem: Implemented structure here for reference
https://github.com/waku-org/go-waku/blob/9000314447e3450754dbc667a2e2049accb758d1/cmd/waku/server/rest/admin_api.yaml#L55-L74


Haven't added shardInfo yet, but will do.

waku/waku_api/rest/doc/schemas/apitypes.yaml
NagyZoltanPeter: My understanding we list all peers we know and their supported protocols we maybe connected at query time.
Suppose that we can have several peers that supports filtering but currently we are using only one of them. (or same with store protocol).
@Ivansete-status WDYT?

waku/waku_api/rest/doc/relayapi_auto_content_topic.yaml
NagyZoltanPeter: It is even more extensive in store api. But as I see its only a problem if one would issue such and api call by hand, in all other programmatic case url encoding shall not be an issue.

waku/waku_api/rest/doc/adminapi.yaml
```diff
@@ -0,0 +1,37 @@
+get:
```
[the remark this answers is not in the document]
NagyZoltanPeter: I accept it!

waku/waku_api/rest/doc/debugapi_info.yaml
```diff
@@ -0,0 +1,17 @@
...
+  tags:
+    - debug
+  responses:
+    '200':
+      description: Information about a Waku v2 node.
+      content:
+        application/json:
+          schema:
+            $ref: './schemas/apitypes.yaml#/WakuInfo'
+    '4XX':
```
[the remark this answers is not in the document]
NagyZoltanPeter: Hehe, nice catch!

waku/waku_api/rest/doc/debugapi_version.yaml
```diff
@@ -0,0 +1,18 @@
...
+  tags:
+    - debug
+  responses:
+    '200':
+      description: The version of a Waku v2 node.
+      content:
+        text/plain:
+          schema:
+            type: string
+    '4XX':
```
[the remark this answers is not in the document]
NagyZoltanPeter: Sure, not!

waku/waku_api/rest/doc/filterapi_v1_messages.yaml
```diff
@@ -0,0 +1,41 @@
+# Legacy support for v1 waku filter
+# /filter/v1/messages/{contentTopic}:
+get: # get_waku_v2_filter_v1_messages
```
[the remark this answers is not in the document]
NagyZoltanPeter: Yes, this is the only way... at least that I know ... open api specification allows referencing files. So there is no way, that I would prefer more either, to move the paths also to the ref files.
I do not like it much, but - imho - still better maintainable separated than in one big.

waku/waku_api/rest/doc/filterapi_v1_messages.yaml
```diff
@@ -0,0 +1,41 @@
...
+          schema:
+            $ref: './schemas/apitypes.yaml#/FilterGetMessagesResponse'
+    # TODO: Review the possible errors of this endpoint
+    '400':
+      description: Bad request.
+      content:
+        text/plain:
+          schema:
+            type: string
+    '404':
```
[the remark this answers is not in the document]
NagyZoltanPeter: Hm, I need to test it, although I think we should check if we have subscription for the given contentTopic only. 
Than its a bad request.
This can be an issue for implementation rather. But here I will give a better explanation!

waku/waku_api/rest/doc/filterapi_v1_subscriptions.yaml
```diff
@@ -0,0 +1,71 @@
+# Legacy support for v1 waku filter
+# /filter/v1/subscriptions:
+post: # post_waku_v2_filter_v1_subscription
```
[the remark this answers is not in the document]
NagyZoltanPeter: It will be removed yes, probably together with json-rpc counterpart.
I think there can only be users of json-rpc filter v1 and none for rest api filter v1.

waku/waku_api/rest/doc/filterapi_v2_messages.yaml
```diff
@@ -0,0 +1,40 @@
...
+      description: Content topic of message
+  responses:
+    '200':
+      description: The latest messages on the polled topic.
+      content:
+        application/json:
+          schema:
+            $ref: './schemas/apitypes.yaml#/FilterGetMessagesResponse'
+    # TODO: Review the possible errors of this endpoint
+    '400':
```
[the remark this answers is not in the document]
NagyZoltanPeter: You're right, although I'm not pretty sure if there is an exact contentTopic format spec or just recommendation?

waku/waku_api/rest/doc/filterapi_v2_messages.yaml
```diff
@@ -0,0 +1,40 @@
...
+      description: Bad request.
+      content:
+        text/plain:
+          schema:
+            type: string
+    '404':
```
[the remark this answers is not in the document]
NagyZoltanPeter: Yes, description shall be more explanatory.

waku/waku_api/rest/doc/filterapi_v1_messages.yaml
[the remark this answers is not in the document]
chaitanyaprem: One of my reason for suggesting this kind of organization is a user who wants to use a protocol has to load/work with only 1 file, hence suggesting to merge filterv1 related API's into 1.
It doesn't cause that big of an impact, just a suggestion.

waku/waku_api/rest/doc/filterapi_v2_messages.yaml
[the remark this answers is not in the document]
chaitanyaprem: There is a recommendation, and when autosharding is used the format has to follow the recommendation otherwise an error is thrown.

waku/waku_api/rest/doc/filterapi_v1_messages.yaml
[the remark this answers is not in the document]
NagyZoltanPeter: @chaitanyaprem what do you mean by load only 1 file?
This setup now follows the only possible file split in open-api structure.
User will face only one file: openapi.yaml
All sub yamls are referenced from there. Through that root openapi.yaml it is possible to generate reference doc/webpage, also can be imported to Postman if one needs it.

NagyZoltanPeter:
@chaitanyaprem @harsh-98, @fryorcraken , @chair28980 
I'm closig this PR as it this will live in a separate repository. However I will make a follow up issue to track these comments.
Some of them effect only api specification in semantic way but many other has consequences on nwaku and go-waku api implementations. Those must be tracked in separate issues.

New repository: https://github.com/waku-org/waku-rest-api
Generated api reference is living in the same repository's GH pages: https://waku-org.github.io/waku-rest-api/

PR #2125, opened 2023-10-13
feat: implementing port 0 support

### LINKED ISSUE #2042: feat: allowing users to choose port 0 for dynamically allocated ports
### Problem
Currently, if we allow the user to choose port 0, the node's data structures are built with 0 as its port.
The data structures do not get updated once we get a port allocated by the kernel and bind, which means that the Waku node does not have the real port it's bound to saved in its internal data structures.

### Suggested solution
We want to fill the Waku node's data structures after the binding and not beforehand, which is currently the case. This would mean that we can build the node, the ENR and all the internal data using the port we were bound to.
It would imply changing the order in which the different structures are getting built, and bumping the port binding to be earlier in the initialization process.

### Alternatives considered
We considered requiring the user to specify a port and not allow delegating the task to the kernel. However, we do want to have this feature.

### Additional context
We will first raise a [PR ](https://github.com/waku-org/nwaku/pull/2037) to disallow the use of port 0 so we don't run into the situation of nodes running and publishing incorrect port information. Once this feature is developed, we will disable the temporary solution, and allow the use of port 0.

### Acceptance criteria
- [ ] The user should be able to run a Waku node selecting port 0
- [ ] All the internal data structures should consistently contain the port which the node is bound to
- [ ] The published ENR and addresses should contain the port which the node is bound to

gabrielmer:
# Description

Updating data structures to correctly reflect port number when the port is dynamically allocated by the kernel.

# Changes

- [x] Updating NetConfig
- [x] Updating ENR record
- [x] Updating wakuDiscV5
- [x] Updating announcedAddresses 

## Issue

closes #2042


apps/wakunode2/app.nim
```diff
@@ -329,6 +300,78 @@ proc setupWakuApp*(app: var App): AppResult[void] =
...
+              AppResult[tuple[tcpPort, websocketPort: Option[Port]]] = 
+  
+  var tcpPort, websocketPort = none(Port)
+  
+  for a in listenAddrs:
+    if a.isWsAddress() and websocketPort.isNone():
+      let wsAddress = initTAddress(a).valueOr:
+        return err(error)
+      websocketPort = some(wsAddress.port)
+    elif tcpPort.isNone():
```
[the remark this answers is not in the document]
gabrielmer: Great catch! Improved it :)

apps/wakunode2/app.nim
```diff
@@ -545,7 +588,18 @@ proc startNode(node: WakuNode, conf: WakuNodeConf,
...
+  
+  try:
+    (waitFor startNode(app.node,app.conf,app.dynamicBootstrapNodes)).isOkOr:
+      return err(error)
+  except CatchableError:
+    return err("exception starting node: " & getCurrentExceptionMsg())
+
+  # Update app data that is set dynamically on node start
+  app.updateApp().isOkOr:
+    return err(error)
```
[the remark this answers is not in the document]
gabrielmer: Super, done! Thank you!

apps/wakunode2/app.nim
```diff
@@ -545,7 +588,18 @@ proc startNode(node: WakuNode, conf: WakuNodeConf,
 
   return ok()
 
-proc startApp*(app: App): Future[AppResult[void]] {.async.} =
+proc startApp*(app: var App): AppResult[void] =
+  
+  try:
+    (waitFor startNode(app.node,app.conf,app.dynamicBootstrapNodes)).isOkOr:
```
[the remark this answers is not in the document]
Ivansete-status: Yes, for this case it is fine. It is quite in the surface :)

apps/wakunode2/app.nim
```diff
@@ -329,6 +300,71 @@ proc setupWakuApp*(app: var App): AppResult[void] =
...
+              AppResult[tuple[tcpPort, websocketPort: Option[Port]]] = 
+  
+  var tcpPort, websocketPort = none(Port)
+  
+  for a in listenAddrs:
+    if a.isWsAddress() and websocketPort.isNone():
+      let wsAddress = initTAddress(a).valueOr:
+        return err(error)
+      websocketPort = some(wsAddress.port)
+    elif not a.isWsAddress() and tcpPort.isNone():
```
vpavlin, line 313: It's a nitpick, but how about

```
if a.WsAddress():
  ...handle wsPort...
else:
  ...handle nonWsPort...
```

Just to skip doing the ws address check twice

vpavlin: Another thing - this will obviously take a port from the first listen address it encounters - is that possible an issue? Could it mean we use non-public port instead of a public one, or is that not  a problem anymore with all the previous changes regarding listen addresses?

vpavlin (approved):
Tested locally with few options combinations and it worked as promised:)

apps/wakunode2/app.nim
gabrielmer: Thank you! Just refactored it to avoid doing the ws address check twice :)

Regarding the ports, the proc takes the tcp and ws(s) ports the switch got bound to - there should be only one of each (this `listenAddrs` seq is part of the switch's struct).
Afterwards we recreate the App's and WakuNode's netConfig as if we had manually configured to bind to those ports.

These lines should not concern public ports: if we configured previously an `extPort`, then NetConfig will be recreated with the same `extPort` as before. And on the other hand, if we use the `ext-multi-addr-only` CLI flag, we would still only (and blindly) announce the multiaddresses  passed by the user.

apps/wakunode2/app.nim
```diff
@@ -329,6 +300,72 @@ proc setupWakuApp*(app: var App): AppResult[void] =
...
+proc getPorts(listenAddrs: seq[MultiAddress]):
+              AppResult[tuple[tcpPort, websocketPort: Option[Port]]] = 
+  
+  var tcpPort, websocketPort = none(Port)
+  
+  for a in listenAddrs:
+    if a.isWsAddress():
+      if websocketPort.isNone():
+        let wsAddress = initTAddress(a).valueOr:
+          return err(error)
```
Ivansete-status, line 312: I'd add a bit more detail in case of error:
```suggestion
          return err("getPorts wsAddr error:" & $error)
```

apps/wakunode2/app.nim
```diff
@@ -329,6 +300,72 @@ proc setupWakuApp*(app: var App): AppResult[void] =
...
+        websocketPort = some(wsAddress.port)
+    elif tcpPort.isNone():
+      let tcpAddress = initTAddress(a).valueOr:
+        return err(error)
```
Ivansete-status, line 316: I'd add a bit more detail in case of error:
```suggestion
        return err("getPorts tcpAddr error:" & $error)
```

apps/wakunode2/app.nim
```diff
@@ -329,6 +300,72 @@ proc setupWakuApp*(app: var App): AppResult[void] =
...
+  let record = enrConfiguration(app.conf, app.netConf, app.key).valueOr:
+    return err(error)
+
+  app.record = record
+  app.node.enr = record
+
+  if app.conf.discv5Discovery:
+    app.wakuDiscV5 = some(app.setupDiscoveryV5())
+
+  ok()
```
Ivansete-status, line 352: Little nitpick to keep consistency ;p
```suggestion
  return ok()
```

apps/wakunode2/app.nim
```diff
@@ -329,6 +300,72 @@ proc setupWakuApp*(app: var App): AppResult[void] =
...
+      return err(error)
+    
+    updateEnr(app).isOkOr:
+      return err(error)
+
+    app.node.announcedAddresses = app.netConf.announcedAddresses
+
+    printNodeNetworkInfo(app.node)
+
+  ok()
```
Ivansete-status, line 368: A small nitpick and adding a bit more feedback in case of error:

```suggestion
    updateNetConfig(app).isOkOr:
      return err("error calling updateNetConfig: " & &error)
    
    updateEnr(app).isOkOr:
      return err("error calling updateEnr: " & $error)

    app.node.announcedAddresses = app.netConf.announcedAddresses

    printNodeNetworkInfo(app.node)

  return ok()
```

Ivansete-status (approved):
Just added a few nitpicks ;P

gabrielmer:
> Wonderful PR! Thanks for it! 💯 Just added a few nitpicks ;P

Done! Thanks for the feedback! :)

Issue #2112, opened 2023-10-05
feat: add new message_hash column to the archive protocol


Ivansete-status:
### Problem
We have the `id` column in sqlite and postgres databases that is not unique.

### Suggested solution (defined by @jm-clius)

Create a unique field, `message_hash`, which will be computed following RFC: https://rfc.vac.dev/spec/14/#deterministic-message-hashing.
The `message_hash` will become the PRIMARY KEY.
This should also be how the CURSOR gets computed when doing a database lookup. So querying for the next page will become much simpler - find the PRIMARY KEY that matches the CURSOR and query from that point
Drop the current `id` field from the current databases schemas.

### Acceptance criteria
- [x] The `message_hash` attribute should be present SQLite.
- [x] The `message_hash` attribute should be present Postgres.
- [x] The current `id` column should be ignored and removed from the database schema.
- [x] Store protocol behaves well and pagination works well.

### Epic
https://github.com/waku-org/pm/issues/57

CC: @chair28980 @fryorcraken

ABresting:
achieved
- worked on the current issue-related PRs this week, launched initial PR already, others lined up.
- finalized all remaining tasks associated with the retention policy, including infra-testing.

next
- (short-term) aim to finalize the current issue by the end of next week.
- (long-term) continuing focus on developing the sync mechanism for archive clients. 

blocker
- NA.

ABresting:
achieved

- worked on the current issue-related PRs this week, and refactored PRs based on components.
- worked on pagination in the light of new attribute

next
- PRs are lined up but linearly dependent so one at a time
- migration of new attribute

blocking
- not exactly blocking but awaiting reviews.

ABresting:
### achieved

On SQLite's schema transition (i.e. this PR) to `messageHash` feature complete PR posted (awaiting reviews), Gained insight into the connection and interplay between the store and archive components, and how they may be leveraged into making a sync protocol. Small stuff - bug fix on the jsWaku which was this PR dependent (that too was time-consuming since my first time interacting with JS code of waku), PR on vacuum on time-based retention policy, thought through the nitty gritty details of node based roles and incentives.

### next

the sync protocol formulation totally based on the messages sync without any external factors into POV. additional stuff such as PostgreSQL PRs by Ivan to gain more aspects of the storage/archive.

ABresting:
**Weekly Update**

[...]

PR #2127, opened 2023-10-13
feat: add new message_hash column

ABresting:
# Description
Adding the message_hash (messageHash) as a new column in the Waku archive protocol. To compute the `messageHash`, we refer to [rfc_guide](https://rfc.vac.dev/spec/14/#deterministic-message-hashing). `computeDigest` function has been modified to support the conjunction of pubSubToic without breaking the existing test cases. In future PRs, the existing `id` attribute/column of the databases (SQLite and PostgreSQL) shall be removed with due testing and pagination support.

The change will help attain a message index type attribute which will be used in sync-related protocols for archive Waku node.

# Changes

- [x] The message_hash attribute should be present SQLite.
- [x] The message_hash attribute should be present Postgres.

## Issue
#2112

### DIFF
--- a/waku/waku_archive/common.nim
+++ b/waku/waku_archive/common.nim
@@ -9,7 +9,8 @@ import
   stew/byteutils,
   nimcrypto/sha2
 import
-  ../waku_core
+  ../waku_core,
+  ../waku_core/topics
 
 
 ## Waku message digest
@@ -18,14 +19,16 @@ import
 
 type MessageDigest* = MDigest[256]
 
-proc computeDigest*(msg: WakuMessage): MessageDigest =
+proc computeDigest*(msg: WakuMessage, pubSubTopic: string = DefaultPubsubTopic): MessageDigest =
   var ctx: sha256
   ctx.init()
   defer: ctx.clear()
 
-  ctx.update(msg.contentTopic.toBytes())
+  ctx.update(pubSubTopic.toBytes())
   ctx.update(msg.payload)
-
+  ctx.update(msg.contentTopic.toBytes())
+  ctx.update(msg.meta)
+  
   # Computes the hash
   return ctx.finish()
 
--- a/waku/waku_archive/archive.nim
+++ b/waku/waku_archive/archive.nim
@@ -100,7 +100,7 @@ proc handleMessage*(w: WakuArchive,
 
   block:
     let
-      msgDigest = computeDigest(msg)
+      msgDigest = computeDigest(msg, pubsubTopic)
       msgReceivedTime = if msg.timestamp > 0: msg.timestamp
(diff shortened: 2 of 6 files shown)


alrevuelta (changes requested):
Approach looks good. Can we add a simple unit test?

Also not sure on the migration strategy, since it modifies the schema.

waku/waku_archive/common.nim
```diff
@@ -18,14 +19,16 @@ import
 
 type MessageDigest* = MDigest[256]
 
-proc computeDigest*(msg: WakuMessage): MessageDigest =
+proc computeDigest*(msg: WakuMessage, pubSubTopic: string = DefaultPubsubTopic): MessageDigest =
```
vpavlin, line 22: Since every message would come through a pubSub topic, what is the point of the default here? (Feels like it would only lead to issues in the future if we forget to pass in a pubSub topic and then get inconsistent behaviour)

ABresting: Yes, but the pubSub topic needs to be passed to the `computeDigest` function as it is not a WakuMessage attribute.

The default is used since we are using the existing function in so many test cases already, one way to get rid of that (without changing all the other test cases) is:
- extend the function with a default (what I've used) or,
- add another function with some sort of similar name and redirect it from the queue driver and Waku archive code for production workflow

better of the two seems to be the former. WDYT @vpavlin?

ABresting:
> Approach looks good. Can we add a simple unit test?

I am wondering what a simple unit test for `messageHash` looks like. Just like `id` it is computationally generated, and there is no existing test case for `id` as well.

If I allow my imagination to run, Hash implies something unique, so maybe a hash collision test? to make sure that the hash computed is indeed unique? IDK if this is an overkill..? WDYT @alrevuelta?

waku/waku_archive/common.nim
vpavlin: Why is "update the tests" not an option?:)

vpavlin:
> I am wondering what a simple unit test for `messageHash` looks like. Just like `id` it is computationally generated, and there is no existing test case for `id` as well.

You could just test that if a different pubsub topic is used, or there is different metadata field value that the hashes are different.

I'd say it is less about testing the hash function output, but rather making sure the digest proc inputs actually influence the output.

I.e. if someone comes back and changes the compueDigest to previous version by removing the pubSub from it, your test should catch the regression.

Testing if the hash collides does not make sense, unless we are actualyl implementing the hash algo

ABresting:
closing this PR so that new PRs are split more targetted towards the specific components such as Postgres, Sqlit, Queue schema etc. Thanks, @Ivansete-status for the suggestions and feedback/review from Alvaro and Vaclav!

PR #2132, opened 2023-10-17
feat: amending computeDigest func. + related test cases

ABresting:
# Description

Amending the `computeDigest` function to add pubSubTopic as a function parameter, will help introduce a new `messageHash` attribute. The corresponding queue driver code is updated to accommodate the amended function.

Amending the existing test cases that use the older version of the `computeDigest` function.

# Changes

- [x] `computeDigest` function amended
- [x] related existing test cases amended
- [x] queue driver code adjusted to the new change

## Issue

#2112

### DIFF
--- a/tests/waku_archive/test_driver_postgres_query.nim
+++ b/tests/waku_archive/test_driver_postgres_query.nim
@@ -37,7 +37,7 @@ proc computeTestCursor(pubsubTopic: PubsubTopic, message: WakuMessage): ArchiveC
     pubsubTopic: pubsubTopic,
     senderTime: message.timestamp,
     storeTime: message.timestamp,
-    digest: computeDigest(message)
+    digest: computeDigest(message, pubsubTopic)
   )
 
 suite "Postgres driver - query by content topic":
@@ -65,7 +65,7 @@ suite "Postgres driver - query by content topic":
     debug "randomized message insertion sequence", sequence=messages.mapIt(it.payload)
 
     for msg in messages:
-      require (await driver.put(DefaultPubsubTopic, msg, computeDigest(msg), msg.timestamp)).isOk()
+      require (await driver.put(DefaultPubsubTopic, msg, computeDigest(msg, DefaultPubsubTopic), msg.timestamp)).isOk()
 
     ## When
     let res = await driver.getMessages(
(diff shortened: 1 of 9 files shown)


vpavlin (approved):
If I am not mistaken, this does not really change the DB schema or how messages are stored or searched for. It may result in some duplicate messages being stored after the node upgrades - as the unique index values will change for the same message after upgrade, but that should not be a problem. 

Thanks for updating the tests!

LGTM

alrevuelta (approved):
lgtm.

I actually dont understant why the pubsub topic wasnt part of the digest if the RFC specifies so: https://rfc.vac.dev/spec/14/#deterministic-message-hashing

And related in case you want to have a look. Seems we have two hashed, one mainly used for store (the one you modified) and one for relay `defaultMessageIdProvider`. Wondering if we should unify both.
https://github.com/waku-org/nwaku/blob/459331e3a6c72a711003700e4c89f5d834e677e5/waku/waku_relay/message_id.nim#L31-L37

ABresting:
> If I am not mistaken, this does not really change the DB schema or how messages are stored or searched for. It may result in some duplicate messages being stored after the node upgrades - as the unique index values will change for the same message after upgrade, but that should not be a problem.

The schema is going to be changed, we aim to replace `id` attribute to the `messageHash` attribute this will improve the storage and searching of messages. The schema change code I will push today as it was suggested by Ivan to push changes incrementally based on some layer that they impact, such as Queue driver, SQLite, Postgres, today I will push SQLite related changes. 

> 
> Thanks for updating the tests!
> 
> LGTM

ABresting:
> lgtm.
> 
> I actually dont understant why the pubsub topic wasnt part of the digest if the RFC specifies so: https://rfc.vac.dev/spec/14/#deterministic-message-hashing
> 
> And related in case you want to have a look. Seems we have two hashed, one mainly used for store (the one you modified) and one for relay `defaultMessageIdProvider`. Wondering if we should unify both.
> 
> https://github.com/waku-org/nwaku/blob/459331e3a6c72a711003700e4c89f5d834e677e5/waku/waku_relay/message_id.nim#L31-L37

AFAIK, the messageID or `id` attribute will be replaced by `messageHash`, i.e. the way messageID is computed is being changed along with it's name. Indeed, we add pubSubTopic inside the hash as referred by this [rfc](https://rfc.vac.dev/spec/14/#deterministic-message-hashing)

PR #2133, opened 2023-10-18
fix(restapi init): wrong install of filter rest api

NagyZoltanPeter:
# Description
Filter rest api installation was rely on wrong configuration argument. --filter was checked (which is intended to filter service) instead of --filternode (that one is for filter client support).

# Changes
- [X] Correct configuration argument check
- [X] added additional check whether filter client protocols are properly mounted.
- [X] As realized that lightpush rest api also lacks this check I added it here.

## How to test

1. run: wakunode2 --relay --filter --lightpush --rest=true --rest-port=11111
  a. should not crash, but start properly
2. run: curl -i -X GET "http://127.0.0.1:11111/filter/v2/subscriptions/111" -H "accept: application/json"
  a. should return Http404
3. run:  wakunode2 --relay --filternode=<address of the first node> --rest=true --rest-port=11112
4. run:  curl -i -X GET "http://127.0.0.1:11112/filter/v2/subscriptions/111" -H "accept: application/json"
  a. must return Some meaing full json response with or without error code.
 
## Issue
[bug: SIGSEGV when Filter and REST are configured](https://github.com/waku-org/nwaku/issues/2131)

### DIFF
--- a/apps/wakunode2/app.nim
+++ b/apps/wakunode2/app.nim
@@ -473,9 +473,12 @@ proc setupProtocols(node: WakuNode,
   if conf.filternode != "":
     let filterNode = parsePeerInfo(conf.filternode)
     if filterNode.isOk():
-      await node.mountFilterClient()
-      node.peerManager.addServicePeer(filterNode.value, WakuLegacyFilterCodec)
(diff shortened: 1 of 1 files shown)


gabrielmer (approved):
LGTM, thanks so much!

vpavlin (approved):

PR #2138, opened 2023-10-18
test(relay-filter): cleanup

AlejandroCabeza:
# Description
Cleanup relay and filter tests.

# Changes
- [X] Remove legacy relay tests.
- [X] Remove unused imports.
- [X] Add missing payload size test for filter.

# Dependencies
https://github.com/waku-org/nwaku/pull/2101


tests/waku_filter_v2/test_waku_client.nim
```diff
@@ -1952,164 +2142,8 @@ suite "Waku Filter - End to End":
...
 
         # Then the client receives the message
-        require await pushHandlerFuture.withTimeout(FUTURE_TIMEOUT)
+        check await pushHandlerFuture.withTimeout(FUTURE_TIMEOUT)
         let (pushedMsgPubsubTopic, pushedMsg) = pushHandlerFuture.read()
         check:
           pushedMsgPubsubTopic == pubsubTopic
           pushedMsg == msg
-
-  suite "MessagePushHandler - Msg List":
```
Ivansete-status, line 1961: Where this suite is being moved?

tests/waku_relay/test_waku_relay.nim
```diff
@@ -1,211 +0,0 @@
-{.used.}
-
-import
-  std/[options, sequtils, strutils],
```
Ivansete-status, line 4: Why this file is being removed?

Ivansete-status (commented):
Thanks for the PR! It is looking good as tests are expected to wait less time!
Besides, I've added a couple of questions :D

jm-clius (approved):
LGTM, thanks!

tests/waku_relay/test_waku_relay.nim
AlejandroCabeza: Those were legacy tests that have been replaced for the new ones.

tests/waku_filter_v2/test_waku_client.nim
AlejandroCabeza: In the beginning I split tests in two to "optimise" the handlers. That is, one set of tests needed a less complex handler, so I created two different handlers, and split the tests between the two. 
After a while, I realised the advantage of having two handlers (a close to zero optimisation), was not worth the simplicity of having just one handler.

Ivansete-status (approved):
LGTM! :100:

PR #2141, opened 2023-10-19
feat: adding ext-multiaddr-only CLI flag

### LINKED ISSUE #1797: feat: provide a way to define advertised addresses

gabrielmer:
# Description
Adding `ext-multiaddr-only` CLI flag so user can override the announced address with their addresses passed to the `ext-multiaddr` flags. This option will allow us to set up a node behind a proxy and publish the proxy's address.

# Changes

- [x] adding `ext-multiaddr-only` CLI flag
- [x] updating the network configuration logic to take into account the new flag
- [x] adding unit test for the use of the flag
- [x] improving ws+extMultiaddr test case to also include wss scenario   

## Issue

closes #1797


vpavlin (approved):
LGTM, would like to know opinion of @alrevuelta, @Ivansete-status, @NagyZoltanPeter or anyone else:)

Discussion in https://discord.com/channels/1110799176264056863/1164492374672678922

PR #2142, opened 2023-10-19
feat: messageHash attribute added in SQLite + testcase

ABresting:
# Description

`messageHash` attributed added to SQLite, which aims to remove the `id` attribute in upcoming PRs. An associated test case also added. 

migration scripts will be changed after the deletion of `id` attribute in upcoming PRs.

# Changes

- [x] `messageHash` attribute added
- [x] added test case to test the `messageHash` attribute

## Issue

#2112

### DIFF
--- a/waku/waku_archive/driver/sqlite_driver/sqlite_driver.nim
+++ b/waku/waku_archive/driver/sqlite_driver/sqlite_driver.nim
@@ -66,6 +66,7 @@ method put*(s: SqliteDriver,
   ## Inserts a message into the store
   let res = s.insertStmt.exec((
     @(digest.data),                # id
+    @(digest.data),                # messageHash
     receivedTime,                  # storedAt
     toBytes(message.contentTopic), # contentTopic
     message.payload,               # payload
--- a/tests/waku_archive/test_driver_postgres.nim
+++ b/tests/waku_archive/test_driver_postgres.nim
@@ -87,7 +87,7 @@ suite "Postgres driver":
     require:
       storedMsg.len == 1
       storedMsg.all do (item: auto) -> bool:
-        let (pubsubTopic, actualMsg, digest, storeTimestamp) = item
+        let (pubsubTopic, actualMsg, digest, messageHash, storeTimestamp) = item
         actualMsg.contentTopic == contentTopic and
         pubsubTopic == DefaultPubsubTopic and
         toHex(computedDigest.data) == toHex(digest) and
--- a/tests/waku_archive/test_driver_sqlite.nim
+++ b/tests/waku_archive/test_driver_sqlite.nim
@@ -60,7 +60,7 @@ suite "SQLite driver":
     check:
       storedMsg.len == 1
       storedMsg.all do (item: auto) -> bool:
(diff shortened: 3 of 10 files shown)


richard-ramos (approved):
LGTM

waku/waku_archive/driver/sqlite_driver/sqlite_driver.nim
```diff
@@ -66,6 +66,7 @@ method put*(s: SqliteDriver,
   ## Inserts a message into the store
   let res = s.insertStmt.exec((
     @(digest.data),                # id
+    @(digest.data),                # messageHash
```
Ivansete-status, line 69: Maybe there is something that I'm missing but `messageHash` looks the same as `id`.

I think we need to create a new `computeDeterministicMessageHash` proc in `waku_archive.nim` where we create the following attribute: https://rfc.vac.dev/spec/14/#deterministic-message-hashing

Ivansete-status (changes requested):
Thanks for the PR! It is getting a look shape. However, I have the impression that we need to properly implement the `computeDeterministicMessageHash` proc, as I mentioned in a comment.

Ivansete-status:
@jm-clius - something that I realized while reviewing this PR is that the messageHash won't be unique and therefore we need to also consider the `storedAt` attribute as part of the PK, as @ABresting is properly applying now.
wdyt?

Ivansete-status:
@jm-clius - something that I realized while reviewing this PR is that the messageHash won't be unique and therefore we need to also consider the `storedAt` attribute as part of the PK, as @ABresting is properly applying now.
wdyt?

waku/waku_archive/driver/sqlite_driver/sqlite_driver.nim
ABresting: well, as per the initial approach, we aim to remove the `id` attribute and replace it with `messageHash`, as not only nomenclature is better but the way digest/hash is computed it makes sense.

the computation logic behind `messageHash` follows the [RFC](https://rfc.vac.dev/spec/14/#deterministic-message-hashing). The digest logic is implemented in the previous PR for the same base issue #2112

ABresting:
> @jm-clius - something that I realized while reviewing this PR is that the messageHash won't be unique and therefore we need to also consider the `storedAt` attribute as part of the PK, as @ABresting is properly applying now. wdyt?

Well, the problem with `storedAt` is that it is very local since the `storedAt` time is the local node time, i.e. for every node this time attribute can be different. The aim of `messageHash` is to make sure we have a unique message identifier across the decentralized network which then be used in the new sync protocol for efficient synchronization.

waku/waku_archive/driver/sqlite_driver/sqlite_driver.nim
Ivansete-status: You are absolutely right that we have it implemented. I got confused because we have another `computeDigest` definition, which is different. We need to revisit it and check if we can only use one.

https://github.com/waku-org/nwaku/blob/817a7b2e68cdd4e702bb2638531f55f063da6647/waku/waku_store/common.nim#L30

Ivansete-status (approved):

PR #2144, opened 2023-10-20
chore: Enhance rest request error handling

NagyZoltanPeter:
# Description
Issue https://github.com/waku-org/nwaku/issues/2135
describes the case when intentionally we do not install certain endpoints.
In such cases we shall report nicely that the offered endpoint is missing due configuration not by error.

# Changes
- [x] Eugen enhance HttpServer and RestHttpServer with an error callback hook on request.
- [x] nim-presto version is elevated to new commit (we had been in the HEAD of master, after new feature implemented we got to the HEAD now again.
- [x] requestErrorHandler added for wakunode2 app.nim for RestServer.
- [x] Error handler takes care of intentionally not installed endpoints.

## How to test

1. run: wakunode2 --relay --filter --lightpush --rest=true --rest-port=11111
2. run: curl -i -X GET "http://127.0.0.1:11111/filter/v2/subscriptions/111" -H "accept: application/json"
3. Check response, it shall tell the endpoint is not installed, configuration shall be checked.
4. curl -i -X GET "http://127.0.0.1:11111/admin/v1/peers" -H "accept: application/json"
5. Response shall be Http200 and may list connected peers or may return empty list.
6. curl -i -X GET "http://127.0.0.1:11111/aaaadmin/v1/peers" -H "accept: application/json"
7. Response shall be Http400 but with bad request message
8. run ./wakunode2 --relay --filternode=/ip4/127.0.0.1/tcp/60000/p2p/16Uiu2HAm8n7MSAr59Mae7K7WoKiyqtKstJhh5RvvGWYx1ZDJ4HGQ --rest=true --rest-port=11112
9. run: curl -i -X GET "http://127.0.0.1:11112/filter/v2/subscriptions/111" -H "accept: application/json"
10. Response shall be Http200 with json response from filter service.
 
## Issue
https://github.com/waku-org/nwaku/issues/2135

### DIFF
--- a/apps/wakunode2/app.nim
+++ b/apps/wakunode2/app.nim
@@ -614,7 +614,33 @@ proc startApp*(app: var App): AppResult[void] =
 ## Monitoring and external interfaces
 
 proc startRestServer(app: App, address: ValidIpAddress, port: Port, conf: WakuNodeConf): AppResult[RestServerRef] =
-  let server = ? newRestHttpServer(address, port)
+
+  # Used to register api endpoints that are not currently installed as keys,
+  # values are holding error messages to be returned to the client
+  var notInstalledTab: Table[string, string] = initTable[string, string]()
+
+  proc requestErrorHandler(error: RestRequestError,
+                                request: HttpRequestRef):
+                                Future[HttpResponseRef] {.async.} =
+    case error
+    of RestRequestError.Invalid:
+      return await request.respond(Http400, "Invalid request", HttpTable.init())
+    of RestRequestError.NotFound:
+      let rootPath = request.rawPath.split("/")[1]
+      if notInstalledTab.hasKey(rootPath):
+        return await request.respond(Http404, notInstalledTab[rootPath], HttpTable.init())
+      else:
+        return await request.respond(Http400, "Bad request initiated. Invalid path or method used.", HttpTable.init())
+    of RestRequestError.InvalidContentBody:
+      return await request.respond(Http400, "Invalid content body", HttpTable.init())
+    of RestRequestError.InvalidContentType:
+      return await request.respond(Http400, "Invalid content type", HttpTable.init())
+    of RestRequestError.Unexpected:
+      return defaultResponse()
+
+    return defaultResponse()
+
+  let server = ? newRestHttpServer(address, port, requestErrorHandler = requestErrorHandler)
 
   ## Admin REST API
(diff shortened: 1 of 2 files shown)


apps/wakunode2/app.nim
```diff
@@ -596,6 +619,8 @@ proc startRestServer(app: App, address: ValidIpAddress, port: Port, conf: WakuNo
       app.node.subscribe((kind: ContentSub, topic: contentTopic), some(autoHandler))
 
     installRelayApiHandlers(server.router, app.node, cache)
+  else:
+    notInstalledTab["relay"] = "/relay endpoints are not available"
```
vpavlin, line 623: Do we have protocol names anywhere as constants? It would be better/nicer than having plain strings as keys (although also probably ok for this singular use case)

NagyZoltanPeter: You're right with that, but currently there are no such consts only for each endpoint paths.
The other option But still would not be more clear coded as we don't have mechanism to structure request handling in hierarchy, so it would be error prone even to duplicate similar code to each handler.
For now I would keep it short and simple here but will add some better response message here I see I missed it though.

vpavlin (approved):

NagyZoltanPeter:
I have made a rebase along with a small enhancement on how the proper error message are gathered.
I would appreciate if @gabrielmer, @Ivansete-status can take a look.
Thanks.

gabrielmer (approved):
Oh apologies, I thought I had already approved it.
Went over it again anyways and looks great :)

Thank you!

apps/wakunode2/app.nim
```diff
@@ -614,7 +614,33 @@ proc startApp*(app: var App): AppResult[void] =
...
+      else:
+        return await request.respond(Http400, "Bad request initiated. Invalid path or method used.", HttpTable.init())
+    of RestRequestError.InvalidContentBody:
+      return await request.respond(Http400, "Invalid content body", HttpTable.init())
+    of RestRequestError.InvalidContentType:
+      return await request.respond(Http400, "Invalid content type", HttpTable.init())
+    of RestRequestError.Unexpected:
+      return defaultResponse()
+
+    return defaultResponse()
```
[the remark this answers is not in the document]
NagyZoltanPeter: Well, I wanted to keep it close where the information exists about the installed / not installed api endpoints.

Issue #1940, opened 2023-08-24
feat: Peer management with shard as a dimension


chair28980:
If we assume that peer discovery takes care of filtering peers by shard, these peers must now be managed in a way that makes sense in a dynamic auto-sharded environment. Some requirements/ideas:

- peer store must preferably try to keep at least a target n peers for each subscribed shard (note that dynamically subscribing/unsubscribing from shards could complicate this)
- peer manager must attempt to maintain a healthy relay connectivity for each subscribed shard
- a mechanism to perform ad-hoc discovery (or another solution) to cater for when a node subscribes to a shard for which there are no tracked peers

Priority: Critical for launch

SionoiS:
**Weekly Update**

- _achieved_: Waku Metadata shard subscriptions, Sharded relay peer management, draft sharded peer store pruning
- _next_: finalize sharded peer store pruning & run simulations

SionoiS:
**Weekly Update**

- _achieved_: discv5 filter peer by capability, misc. improvement w.r.t sharding and tests, sharded peer management improvement
- _next_:  run more simulations

SionoiS:
**Weekly Update**

- _achieved_: sharded peer management final version in review
- _next_:  review feedback

SionoiS:
**Weekly Update**

- _achieved_: sharded peer management and store pruning PR merged

PR #2149, opened 2023-10-23
feat: metadata protocol shard subscription

### LINKED ISSUE #2105: feat: automatically update shards in waku metadata protocol
Automatically update shards in waku metadata protocol. They are currently empty.

SionoiS:
# Description

The shards a node support can dynamically changes and the metadata protocol must be updated at run-time to reflect this.

# Changes

- [x] Added shard subscription listener to metadata protocol
- [x] Refactor/clean-up

Tracking #1940

Fix #2105

### DIFF
--- a/waku/node/waku_node.nim
+++ b/waku/node/waku_node.nim
@@ -155,7 +155,7 @@ proc new*(T: type WakuNode,
   )
 
   # mount metadata protocol
-  let metadata = WakuMetadata.new(netConfig.clusterId)
+  let metadata = WakuMetadata.new(netConfig.clusterId, queue)
   node.switch.mount(metadata, protocolMatcher(WakuMetadataCodec))
   node.wakuMetadata = metadata
   peerManager.wakuMetadata = metadata
@@ -1126,6 +1126,8 @@ proc start*(node: WakuNode) {.async.} =
   await node.switch.start()
 
   node.started = true
+
+  node.wakuMetadata.start()
 
   info "Node started successfully"
 
--- a/waku/waku_metadata/protocol.nim
+++ b/waku/waku_metadata/protocol.nim
@@ -4,7 +4,7 @@ else:
   {.push raises: [].}
 
 import
-  std/[options, sequtils, random],
+  std/[options, sequtils, random, sets],
   stew/results,
   chronicles,
   chronos,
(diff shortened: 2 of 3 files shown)


waku/waku_metadata/protocol.nim
```diff
@@ -27,57 +27,51 @@ const RpcResponseMaxBytes* = 1024
...
-      shards: m.shards
-      ).encode().buffer)
-  except CatchableError as exc:
-    return err(exc.msg)
+  let response = WakuMetadataResponse(
+    clusterId: some(m.clusterId),
+    shards: toSeq(m.shards)
+  )
+  
+  let res = catch: await conn.writeLP(response.encode().buffer)
```
alrevuelta, line 38: afaik `writeLP` can raise an exception, why removing the except?

waku/waku_metadata/protocol.nim
```diff
@@ -27,57 +27,51 @@ const RpcResponseMaxBytes* = 1024
...
-    ).encode().buffer)
-    buffer = await conn.readLp(RpcResponseMaxBytes)
-  except CatchableError as exc:
-    error = $exc.msg
-  finally:
-    # close, no more data is expected
-    await conn.closeWithEof()
+  let request = WakuMetadataRequest(clusterId: some(m.clusterId), shards: toSeq(m.shards))
+
+  let writeRes = catch: await conn.writeLP(request.encode().buffer)
```
alrevuelta, line 47: same, i think it can raise an exception

alrevuelta (changes requested):
really good, thanks for this.
* minor comment regarding exception raising
* can we add some units tests for this?

waku/waku_metadata/protocol.nim
SionoiS: the `catch` bundle the exception if present into a `Result` with an error containing the exception message.

waku/waku_metadata/protocol.nim
SionoiS: Same here `catch:` take care of the exception

SionoiS:
> * can we add some units tests for this?

For the subscription part?

waku/waku_metadata/protocol.nim
alrevuelta: i see. not sure i like using this kind of non-standard sugar. ok we already use nim-results, but in the repo we are not using this "way" of `catch`.

SionoiS: Maybe we should :wink:

waku/waku_metadata/protocol.nim
```diff
@@ -27,57 +27,51 @@ const RpcResponseMaxBytes* = 1024
...
+  return ok(response)
 
 proc initProtocolHandler*(m: WakuMetadata) =
   proc handle(conn: Connection, proto: string) {.async, gcsafe, closure.} =
-    var buffer: seq[byte]
-    try:
-      buffer = await conn.readLp(RpcResponseMaxBytes)
-    except CatchableError as exc:
+    let res = catch: await conn.readLp(RpcResponseMaxBytes)
+    let buffer = res.valueOr:
```
Ivansete-status, line 69: I wonder if we should add an `error` log before `return`. Likewise for the next statement.

waku/waku_metadata/protocol.nim
```diff
@@ -27,57 +27,51 @@ const RpcResponseMaxBytes* = 1024
...
+  let readRes = catch: await conn.readLp(RpcResponseMaxBytes)
+  let buffer = 
+    if readRes.isErr():
+      return err("read failed: " & readRes.error.msg)
+    else: readRes.get()
 
-  let decodedBuff = WakuMetadataResponse.decode(buffer)
-  if decodedBuff.isErr():
-    return err("decode failed: " & $decodedBuff.error)
+  let closeRes = catch: await conn.closeWithEof()
```
Ivansete-status, line 57: Just for me to learn something, what is the benefit of using that technique? (`catch:`)

Ivansete-status (approved):
LGTM! Thanks for it!
Just added a few nitpicks but not blocking ones :)

On the other hand, I realized that the metadata protocol is being mounted within the creation of the Waku node. Why is that needed @alrevuelta? I wonder if we could follow the standard approach that we are following in other protocols so that all is more consistent.

waku/waku_metadata/protocol.nim
SionoiS: It's more concise than `try: catch:` and less rightward drift too. It even bundle the error msg in a `Result`

waku/waku_metadata/protocol.nim
SionoiS: I agree that would be better

waku/node/waku_node.nim
```diff
@@ -1127,6 +1127,10 @@ proc start*(node: WakuNode) {.async.} =
 
   node.started = true
 
+  node.wakuMetadata.start()
+
+  asyncSpawn node.wakuMetadata.subscriptionsListener(node.topicSubscriptionQueue)
```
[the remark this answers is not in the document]
SionoiS: It would make more sense yes. I'll change it.

jm-clius (approved):
Thanks! Good step forward.

PR #2151, opened 2023-10-23
feat: sharded peer manager

SionoiS:
# Description

Peer manager now check each shard for a target number of in/out peers and connect/prune accordingly.

# Changes

- [x] Per shard pruning of connections
- [x] Per shard management of in/out connections
- [x] Refactor/clean-up

Tracking #1940


waku/node/peer_manager/peer_manager.nim
```diff
@@ -619,30 +606,58 @@ proc getNumStreams*(pm: PeerManager, protocol: string): (int, int) =
...
+      peersToDisconnect += abs(inPeerDiff)
 
-  if outRelayPeers.len >= pm.outRelayPeersTarget:
+    if outPeerDiff <= 0:
+      continue
+
+    let connectablePeers = peers.filterIt(
+      (not pm.peerStore.isConnected(it.peerId) and pm.canBeConnected(it.peerId)))
+
+    peersToConnect.insert(connectablePeers[0..outPeerDiff])
```
gabrielmer, line 640: If peers are inserted to the peersToConnect seq by shard based on the order in `pm.wakuMetadata.shards.items` seq, if one of the first shards on that array always have a lot of peers to connect to, will other shards get starved? If so, is that an issue?

waku/node/peer_manager/peer_manager.nim
```diff
@@ -49,10 +49,10 @@ const
   BackoffFactor = 4
 
   # Limit the amount of paralel dials
-  MaxParalelDials = 10
+  MaxParallelDials = 10
 
   # Delay between consecutive relayConnectivityLoop runs
-  ConnectivityLoopInterval = chronos.seconds(15)
+  ConnectivityLoopInterval = chronos.seconds(60)
```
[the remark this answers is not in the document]
SionoiS: Nah I don't think so. I made it longer because the node may have to connect to many node.

waku/node/peer_manager/peer_manager.nim
SionoiS: Each shard should be independent, and with an in/out target # of peers. If some shard has less peers to connect to it's a discovery problem, not management.

To answer your question, yes they could but can't fix that since discovery is random.

SionoiS: Just notice that `connectablePeer` could be shorter than `outPeerDif`

gabrielmer: Makes sense, thank you!

> Just notice that `connectablePeer` could be shorter than `outPeerDif`

True! Great catch! :)

waku/node/peer_manager/peer_manager.nim
alrevuelta, line 55: note that this will increase the time it takes to get a reasonable amount of peers at startup. it will only attempt 10 peers each 1 minute, so if we set eg 50 peers, it will take 5 minutes, which seems quite a lot.

waku/node/peer_manager/peer_manager.nim
```diff
@@ -283,108 +277,101 @@ proc canBeConnected*(pm: PeerManager,
...
 
+proc onPeerMetadata(pm: PeerManager, peerId: PeerId) {.async.} =
+  # To prevent metadata protocol from breaking prev nodes, by now we only
+  # disconnect if the clusterid is specified.
+  if pm.wakuMetadata.clusterId == 0:
+    return
+
+  let res = catch: await pm.switch.dial(peerId, WakuMetadataCodec)
+
+  let reason =
```
alrevuelta, line 319: good one

waku/node/peer_manager/peer_manager.nim
```diff
@@ -619,30 +606,59 @@ proc getNumStreams*(pm: PeerManager, protocol: string): (int, int) =
...
 
 proc connectToRelayPeers*(pm: PeerManager) {.async.} =
-  let (inRelayPeers, outRelayPeers) = pm.connectedPeers(WakuRelayCodec)
-  let maxConnections = pm.switch.connManager.inSema.size
-  let totalRelayPeers = inRelayPeers.len + outRelayPeers.len
-  let inPeersTarget = maxConnections - pm.outRelayPeersTarget
+  var peersToConnect: seq[RemotePeerInfo]
+  var peersToDisconnect: int
+
+  for shard in pm.wakuMetadata.shards.items:
```
alrevuelta, line 619: not sure that we can use `wakuMetadata.shards` right now. afaik `shards` is not populated nor updated yet?

waku/node/peer_manager/peer_manager.nim
```diff
@@ -619,30 +606,59 @@ proc getNumStreams*(pm: PeerManager, protocol: string): (int, int) =
...
+    let peers = pm.peerStore.getPeersByShard(uint16(pm.wakuMetadata.clusterId), uint16(shard))
+
+    let connectedInPeers = peers.filterIt(
```
alrevuelta, line 622: main reason for using `pm.connectedPeers(WakuRelayCodec)` is that it returns the peers connected with that codec.

in other words, `.isConnected(it.peerId)` might give you peers that are connected but with other protocol that is not relay, and afaik here we need to count relay peers.

waku/node/peer_manager/peer_manager.nim
```diff
@@ -619,30 +606,59 @@ proc getNumStreams*(pm: PeerManager, protocol: string): (int, int) =
...
 
 proc pruneInRelayConns(pm: PeerManager, amount: int) {.async.} =
-  let (inRelayPeers, outRelayPeers) = pm.connectedPeers(WakuRelayCodec)
+  let (inRelayPeers, _) = pm.connectedPeers(WakuRelayCodec)
   let connsToPrune = min(amount, inRelayPeers.len)
 
   for p in inRelayPeers[0..<connsToPrune]:
     asyncSpawn(pm.switch.disconnect(p))
 
 proc connectToRelayPeers*(pm: PeerManager) {.async.} =
```
alrevuelta, line 615: since you have added `pruneInRelayConns` perhaps this has to be renamed? because its not `connectToRelayPeers` anymore buy also "disconnect"?

waku/node/peer_manager/peer_manager.nim
```diff
@@ -619,30 +606,59 @@ proc getNumStreams*(pm: PeerManager, protocol: string): (int, int) =
...
+  if peersToConnect.len == 0:
     return
 
-  let notConnectedPeers = pm.peerStore.getNotConnectedPeers().mapIt(RemotePeerInfo.init(it.peerId, it.addrs))
-  let outsideBackoffPeers = notConnectedPeers.filterIt(pm.canBeConnected(it.peerId))
-  let numPeersToConnect = min(outsideBackoffPeers.len, MaxParalelDials)
+  # Even with duplicate, after a couple of iteration the target will be reached
+  peersToConnect.deduplicate()
+
+  let chunks = peersToConnect.len div MaxParallelDials
```
alrevuelta, line 651: can we simplify this a bit? understand the goal but maybe we don't need an special case for `mod MaxParallelDials` (the remainder)

waku/node/peer_manager/peer_manager.nim
alrevuelta: can see that `connectToRelayPeers` has been modified, so nothing to address then :)

alrevuelta (changes requested):
thanks! left some comments

waku/node/peer_manager/peer_manager.nim
SionoiS: The PRs are splitted but I intent to merge the shard subscription one first.

waku/node/peer_manager/peer_manager.nim
SionoiS: Ah good catch!

waku/node/peer_manager/peer_manager.nim
SionoiS: True

waku/node/peer_manager/peer_manager.nim
SionoiS: If I remove it, we can't connect to less than `MaxParallelDials` peers. That would be weird no?

waku/node/peer_manager/peer_manager.nim
alrevuelta: @SionoiS cant see it fixed in the latest commit?

waku/node/peer_manager/peer_manager.nim
alrevuelta: did it quickly but perhaps this can prove my point

```
let test = @[0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12]
let paralelDials = 5

for i in countup(0, test.len, paralelDials):
  echo test[i..min((i+paralelDials), test.len-1)]
  
#@[0, 1, 2, 3, 4, 5]
#@[5, 6, 7, 8, 9, 10]
#@[10, 11, 12]
```

this doesn't handle the last iteration as a special case

@SionoiS

waku/node/peer_manager/peer_manager.nim
SionoiS: @alrevuelta  L#622 the peers are filtered by protocol (WakuRelayCodec)

waku/node/peer_manager/peer_manager.nim
SionoiS: > did it quickly, may be missing something, but perhaps this can prove my point
> 
> ```
> let test = @[0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12]
> let paralelDials = 5
> 
> for i in countup(0, test.len, paralelDials):
>   echo test[i..min((i+paralelDials), test.len-1)]
>   
> #@[0, 1, 2, 3, 4, 5]
> #@[5, 6, 7, 8, 9, 10]
> #@[10, 11, 12]
> ```
> 
> this doesn't handle the last iteration as a special case
> 
> @SionoiS

I get it now!

SionoiS:
I'm still looking at the tests. Some are easier to update than others.

alrevuelta (approved):
lgtm! Thanks for the changes.

Since it involves some networking changes, sometimes difficult to unit test, I would verify with waku-simulator (or related setup) that connectivity is working fine.

SionoiS:
> lgtm! Thanks for the changes.
> 
> Since it involves some networking changes, sometimes difficult to unit test, I would verify with waku-simulator (or related setup) that connectivity is working fine.

Yes I won't be merging this without some simulation. These changes could have unforeseen consequences.

waku/node/peer_manager/peer_manager.nim
```diff
@@ -619,30 +614,53 @@ proc getNumStreams*(pm: PeerManager, protocol: string): (int, int) =
...
+    peers.keepItIf(it.protocols.contains(WakuRelayCodec))
+
+    let connectedInPeers = peers.filterIt(
+      pm.peerStore.isConnected(it.peerId) and it.direction == Inbound)
+    
+    let connectedOutPeers = peers.filterIt(
+      pm.peerStore.isConnected(it.peerId) and it.direction == Outbound)
+
+    let outPeerDiff = pm.outRelayPeersTarget - connectedOutPeers.len
+    let inPeerDiff = pm.inRelayPeersTarget - connectedInPeers.len
```
jm-clius, line 639: Do I understand correctly that this now becomes the _per-shard_ targets? That makes sense, except that I think these values are deduced from the underlying absolute maximum connections setting in libp2p. In other words, the switch would have to be initialised to accommodate for the connection targets for each shard (plus allowing some service protocol client slots). This may require having to dynamically altering these targets? In any case, unless I'm missing this, a per-shard target would have to reflect in the switch max-connections configuration.

waku/node/peer_manager/peer_manager.nim
```diff
@@ -619,30 +614,53 @@ proc getNumStreams*(pm: PeerManager, protocol: string): (int, int) =
...
 
   for p in inRelayPeers[0..<connsToPrune]:
     asyncSpawn(pm.switch.disconnect(p))
 
-proc connectToRelayPeers*(pm: PeerManager) {.async.} =
-  let (inRelayPeers, outRelayPeers) = pm.connectedPeers(WakuRelayCodec)
-  let maxConnections = pm.switch.connManager.inSema.size
-  let totalRelayPeers = inRelayPeers.len + outRelayPeers.len
-  let inPeersTarget = maxConnections - pm.outRelayPeersTarget
+proc manageRelayPeers*(pm: PeerManager) {.async.} =
```
jm-clius, line 623: Could we somewhat annotate this function with (simple) comments to explain the heuristic. Once we have some confidence in our approach, we need to be able to extract these as a simple set of shard connection management "rules" to help us maintain and extend this in future.

jm-clius (approved):
Great job. Approving as not to be a blocker, though I think the current targets per shard may quickly cause issues if underlying max-connections for the libp2p switch does not account for the total number of connections. Comment re this below.

waku/node/peer_manager/peer_manager.nim
SionoiS: Totally forgot to check where these target value come from. I assumed it was config based. I'll verify.

SionoiS: I divided the targets by the number of shards. There's many way to config the node wrong but with the default value it should work fine.

waku/node/peer_manager/peer_manager.nim
```diff
@@ -283,108 +289,101 @@ proc canBeConnected*(pm: PeerManager,
...
+      if (let metadata = (await pm.wakuMetadata.request(conn.get())); metadata.isOk()):
+        if (let clusterId = metadata.get().clusterId; clusterId.isSome()):
+          if pm.wakuMetadata.clusterId == clusterId.get():
+            if metadata.get().shards.anyIt(pm.wakuMetadata.shards.contains(it)):
+              return
+            else: "no shards in common"
+          else: "different clusterId reported: " & $pm.wakuMetadata.clusterId & " vs " & $clusterId
+        else: "empty clusterId reported"
+      else: "failed waku metadata codec request"
+    else: "waku metadata codec not supported"
```
Ivansete-status, line 342: This snippet is great indeed! The only thing is that we may bring a little more information because we are missing a couple of result error messages. The following is aimed to enhance that a little :)

```suggestion
  var reason: string
  if res.isErr():
    reason = "waku metadata codec not supported: " & res.error[].msg
  else:
    let metadata = await pm.wakuMetadata.request(res.get())
    if metadata.isErr():
      reason = "failed waku metadata codec request: " & $metadata.error
    else:
      if (let clusterId = metadata.get().clusterId; clusterId.isSome()):
        if pm.wakuMetadata.clusterId == clusterId.get():
          if metadata.get().shards.anyIt(pm.wakuMetadata.shards.contains(it)):
            return
          else: reason = "no shards in common"
        else: reason = "different clusterId: " & $pm.wakuMetadata.clusterId & " vs " & $clusterId
      else: reason = "empty clusterId reported"
```

Ivansete-status (commented):
I'm late to this PR. Just added a comment aiming to bring more debug detail. No blocking

waku/node/peer_manager/peer_manager.nim
SionoiS: Yes indeed. I already changed it because I wanted more details. commit incoming

SionoiS:
In order to unblock this work I've opened #2237

SionoiS:
I've done simulations and this should be the final version of this features. Sorry for the re-review!

waku/node/peer_manager/peer_manager.nim
```diff
@@ -115,21 +119,38 @@ proc addPeer*(pm: PeerManager, remotePeerInfo: RemotePeerInfo, origin = UnknownO
...
+          protos.add(WakuRelayCodec)
+        of Store:
+          protos.add(WakuStoreCodec)
+        of Filter:
+          protos.add(WakuFilterSubscribeCodec)
+          protos.add(WakuFilterPushCodec)
+        of LightPush:
+          protos.add(WakuLightPushCodec)
+
+    pm.peerStore[ProtoBook][remotePeerInfo.peerId] = protos
```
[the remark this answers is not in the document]
SionoiS: We have a chicken and egg problem here. If the protos are not populated we can't filter non-Relay peers then we don't connect and don't get the protos from identify.

Alternative would be to connect to all peer we discover regardless of their (ENR) capability.

waku/node/peer_manager/peer_manager.nim
```diff
@@ -293,105 +318,106 @@ proc canBeConnected*(pm: PeerManager,
...
 proc onConnEvent(pm: PeerManager, peerId: PeerID, event: ConnEvent) {.async.} =
-  case event.kind
-  of ConnEventKind.Connected:
-    let direction = if event.incoming: Inbound else: Outbound
-    discard
-  of ConnEventKind.Disconnected:
-    discard
+  discard
+
+proc onPeerMetadata(pm: PeerManager, peerId: PeerId) {.async.} =
```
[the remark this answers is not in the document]
SionoiS: I'll rewrite this.

log.txt
jm-clius, line 1: Should this have been committed?

waku/node/peer_manager/peer_manager.nim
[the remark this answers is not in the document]
jm-clius: Indeed. We shouldn't mix the two concepts here:
- the Protobook tracks what libp2p protocols you have _identified_ peers as supporting
- ENR capabilities provide an optional, "convenience flag" to filter for interesting peers before having to connect

Nodes may support more protocols than they're willing to advertise in their ENRs (which may be their way to say "I support Store, but not for you") and they may even lie about what they support. TL;DR the ENR protocols should be used for initial filtering, but should not be seen as authoritative on the libp2p level.

waku/node/peer_manager/peer_manager.nim
```diff
@@ -633,30 +662,88 @@ proc getNumStreams*(pm: PeerManager, protocol: string): (int, int) =
...
+  
+    let shardCount = connectablePeers.len
+
+    connectablePeers.keepItIf(
+      not pm.peerStore.isConnected(it.peerId) and
+      pm.canBeConnected(it.peerId))
+
+    let connectableCount = connectablePeers.len
+
+    connectablePeers.keepItIf(it.protocols.contains(WakuRelayCodec))
```
jm-clius, line 719: Based on comments elsewhere, if we do not populate the `ProtoBook` with the proto flags from the ENR, perhaps we can retrieve the ENR here for non-connected peers here and only consider those advertising Relay as viable?

waku/node/peer_manager/peer_manager.nim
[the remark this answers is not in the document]
jm-clius: (That said, I can see the complexity this would cause elsewhere as we do not pre-filter on protocol when adding the peers, only when maintaining the connections. It may be possible to retrieve the ENR from the ENR book at the point of retrieving relay peers to connect to, but I understand if there are other complexities that make this a necessary workaround.)

jm-clius (approved):
LGTM!

waku/node/peer_manager/peer_manager.nim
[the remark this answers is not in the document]
SionoiS: As soon as you connect the protos gets overwritten by identify anyway.

You either have to remember that protos are never populated before the first connection and the ENR caps should be used OR remember that before the first connection the protos maybe not be the same as after.

I feel like the second case is less of a gotcha than the first.

PR #2159, opened 2023-10-25
feat: messageHash attaribute added in SQLite + migration script ready

ABresting:
# Description
`messageHash` attribute added to SQLite, which removes the `id` attribute. An associated test case also added. A migration script also added.

Also also unifies the use of `computeDigest` as per the RFC. Since it was inconsistent with the remaining code. All the instances of `id` attribute replaced by `messageHash` which includes data types, testcases and library code.

It has to be done using a single PR to avoid adding another migration script to remove the legacy `id` attribute related code.
Wherever possible digest is replaced by `messageHash` for uniformity.

Adding folks from goWaku and jsWaku as well since this PR also modifies the nodeRestAPIs and some associated types.

# Changes

- [x] `id` attribute replaced by `messageHash`
- [x] `id`/`messageHash` computation made uniform protocol wide.
- [x] all instances of `id` attribute replaced by `messageHash`
- [x] migration script added for `id` to `messageHash` replacement
- [x] primary key of SQLite updated

## Issue

#2112


waku/waku_archive/archive.nim
```diff
@@ -104,7 +104,7 @@ proc handleMessage*(w: WakuArchive,
       msgReceivedTime = if msg.timestamp > 0: msg.timestamp
                         else: getNanosecondTime(getTime().toUnixFloat())
 
-    trace "handling message", pubsubTopic=pubsubTopic, contentTopic=msg.contentTopic, timestamp=msg.timestamp, digest=msgDigest
+    trace "handling message", pubsubTopic=pubsubTopic, contentTopic=msg.contentTopic, timestamp=msg.timestamp, messageHash=msgDigest
```
Ivansete-status, line 107:
```suggestion
    trace "handling message", pubsubTopic=pubsubTopic, contentTopic=msg.contentTopic, timestamp=msg.timestamp, messageHash=msgHash
```

waku/waku_archive/driver/postgres_driver/postgres_driver.nim
```diff
@@ -93,12 +93,12 @@ proc reset*(s: PostgresDriver): Future[ArchiveDriverResult[void]] {.async.} =
 method put*(s: PostgresDriver,
             pubsubTopic: PubsubTopic,
             message: WakuMessage,
-            digest: MessageDigest,
+            messageHash: MessageDigest,
```
Ivansete-status, line 96: We need to change that type name elsewhere.
```suggestion
            messageHash: MessageHash,
```

waku/waku_archive/driver/sqlite_driver/migrations.nim
```diff
@@ -14,12 +14,12 @@ logScope:
...
 
 
-const SchemaVersion* = 7 # increase this when there is an update in the database schema
+const SchemaVersion* = 8 # increase this when there is an update in the database schema
 
 template projectRoot: string = currentSourcePath.rsplit(DirSep, 1)[0] / ".." / ".." / ".." / ".."
 const MessageStoreMigrationPath: string = projectRoot / "migrations" / "message_store"
 
-proc isSchemaVersion7*(db: SqliteDatabase): DatabaseResult[bool] =
+proc isSchemaVersion8*(db: SqliteDatabase): DatabaseResult[bool] =
```
Ivansete-status, line 22: This proc is a temporary workaround that we will keep until version 0.30.0.
Therefore, we need to keep considering the hard-coded logic for the version 7 :D
```suggestion
proc isSchemaVersion7*(db: SqliteDatabase): DatabaseResult[bool] =
```

waku/waku_archive/driver/sqlite_driver/sqlite_driver.nim
```diff
@@ -65,7 +65,7 @@ method put*(s: SqliteDriver,
             Future[ArchiveDriverResult[void]] {.async.} =
   ## Inserts a message into the store
   let res = s.insertStmt.exec((
-    @(digest.data),                # id
+    @(digest.data),                # messageHash
```
Ivansete-status, line 68:
```suggestion
    @(messageHash.data),                # messageHash
```

waku/waku_store/common.nim
```diff
@@ -27,13 +27,15 @@ type WakuStoreResult*[T] = Result[T, string]
 
 type MessageDigest* = MDigest[256]
 
-proc computeDigest*(msg: WakuMessage): MessageDigest =
+proc computeDigest*(msg: WakuMessage, pubSubTopic: string): MessageDigest =
```
Ivansete-status, line 30: What if we completely remove this and only have the `computeDigest` from `waku_archive`?
I suggest doing that in a separate PR.

Ivansete-status (changes requested):
It's getting a good shape indeed!
I added a few comments that we need to resolve.
Especially interesting is the one around the "version 7" issue, where we should emphasize the upgrade correctness. i.e. we need to make sure the database gets well updated from sqlite files created in version 0.15.0, for example.

tests/waku_store/test_resume.nim
```diff
@@ -11,7 +11,7 @@ import
   ../../waku/node/message_store/sqlite_store,
   ../../waku/node/peer_manager,
   ../../waku/waku_core,
-  ../../waku/waku_store,
+  ../../waku/waku_store
```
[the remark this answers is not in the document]
ABresting: I think it's one of those typos :D

waku/waku_store/common.nim
ABresting: indeed, this created a whole lot of mess for me earlier.

Ivansete-status (changes requested):
Apologies that I didn't realize that before but I think it is better to stick to the "digest" concept. Trying to modify so many files to accommodate the name to _messageHash_ isn't great because we would need also to change the _Store_ protocol itself.

Therefore I think it is better to keep using the `digest` name for that field in the _Message_ database.

@jm-clius - wdyt about naming the new field `messageDigest` instead of `messageHash`? I think that would make it more aligned with our current code base. Maybe easier to just rename the "message_hash" to "messageDigest" in https://rfc.vac.dev/spec/14/#test-vectors :)

waku/waku_store/common.nim
```diff
@@ -23,17 +23,19 @@ const
...
 
 
-## Waku message digest
+## Waku message hash
 
-type MessageDigest* = MDigest[256]
+type MessageHash* = MDigest[256]
 
-proc computeDigest*(msg: WakuMessage): MessageDigest =
+proc computeDigest*(msg: WakuMessage, pubSubTopic: string): MessageHash =
```
jm-clius, line 30: This should not be what's used to compute the message hash - this function exists for various historical reasons that I can't fully remember. It must indeed eventually be deprecated, but the authoritative message digest computation is: https://github.com/waku-org/nwaku/blob/master/waku/waku_core/message/digest.nim

jm-clius: In other words, changing this function is unnecessarily going to break backwards compatibility. We simply want to add the message digest as defined in the RFC and implemented in https://github.com/waku-org/nwaku/blob/master/waku/waku_core/message/digest.nim as the primary key index column. This will allow us to extend the Store protocol in future to operate on these unique message digests. These will also replace our current cursors, but we'll do so in a way that doesn't break backwards compatbility. IMO for now the existing query behaviour, and especially cursors, shouldn't change as it will be very high-impact, unless I'm missing something?

jm-clius (changes requested):
I think this introduces significant breaking changes. All Store clients operate currently with a view of how the cursors work and should be computed and this will cause queries to fail, afaics.

I don't see the need for introducing breaking changes. The main idea is to make sure that we introduce the authoritative deterministic message hash as set out in RFC 14 (https://rfc.vac.dev/spec/14/#deterministic-message-hashing) and implemented in https://github.com/waku-org/nwaku/blob/master/waku/waku_core/message/digest.nim as a queryable "column" (at least for postgresql, but it probably makes sense in the other archive impls too). To not break backwards compatibility, this should be a new column (I'm happy with `messageHash`). Since the deterministic message hash is used to populate this column, it MUST be unique and can serve as the (only) primary key. The `computeDigest` and other functions will indeed eventually be deprecated once we extend the Store protocol itself, but should remain untouched for now.

ABresting:
closing this since as per new research [direction](https://github.com/waku-org/research/issues/46), a new [PR](https://github.com/waku-org/nwaku/pull/2202) with targeted scope is drafted.

Issue #2164, opened 2023-10-27
chore: time retention policy should perform vacuum as well as size and capacity retention policy


Ivansete-status:
### Background
The time retention policy reduces the number of rows but does not apply size reductions in the _SQLite_ database file.

### Details
The following doesn't apply vacuum: https://github.com/waku-org/nwaku/blob/master/waku/waku_archive/retention_policy/retention_policy_time.nim

Whereas the following does:
https://github.com/waku-org/nwaku/blob/master/waku/waku_archive/retention_policy/retention_policy_size.nim
https://github.com/waku-org/nwaku/blob/master/waku/waku_archive/retention_policy/retention_policy_capacity.nim

### Acceptance criteria
The "vacuum" operation should be performed more regularly so that the database file size is actually reduced after reducing some rows.

PR #2165, opened 2023-10-27
chore: vacuum in time based retention policy

ABresting:
# Description
Enables the vacuum of DB in Time-based retention policy. 

# Changes

- [x] vacuuming of DB in time-based retention policy

## Issue

#2164

### DIFF
--- a/waku/waku_archive/retention_policy/retention_policy_time.nim
+++ b/waku/waku_archive/retention_policy/retention_policy_time.nim
@@ -49,4 +49,10 @@ method execute*(p: TimeRetentionPolicy,
   if res.isErr():
     return err("failed to delete oldest messages: " & res.error)
 
+  # vacuum to get the deleted pages defragments to save storage space
+  # this will resize the database size
+  let resVaccum = await driver.performVacuum()
+  if resVaccum.isErr():
+    return err("vacuumming failed: " & resVaccum.error)
+
   return ok()


Ivansete-status (approved):
LGTM! Thanks for that @ABresting !
Would it be possible to test that this actually reduces the sqlite file size?
Cheers

SionoiS (approved):
Approved BUT like Ivan said. Are we sure it works?

ABresting:
> Approved BUT like Ivan said. Are we sure it works?

Now that's actually a `good first issue` material!

jm-clius (approved):
Approving, although I think we have to monitor how long this vacuuming can stall nodes with large DBs (e.g. Status with 30 days of history), or do we already have data on how long a vacuum can take?

richard-ramos:
Would be good to test this PR with a copy of a database like status.test or status.prod. I'm worried that this will end up impacting negatively the node. `VACUUM` seems more of an activity that should be executed by infra periodically by taking the node offline than something that could be executed as part of normal operation of the node

richard-ramos:
The go-waku prod fleet is currently connected to status.prod and keeps 7d of history. VACUUM took 11mins to execute, (because of this, this PR was introduced to indicate the user that something is happening https://github.com/waku-org/go-waku/pull/838)
cc: @yakimant

jm-clius:
I agree with @richard-ramos here. Approved previously, but with some hard numbers of time it takes to vacuum, I don't think this is a good idea. @ABresting @Ivansete-status I'm not sure that periodic vacuuming should be part of retention policy at all - this seems to be something specifically required by SQLite and not any other type of backend. In this case a better approach is indeed to have infra perform the vacuum "out of band" from time to time so as not to stall production nodes. In our case, this is only necessary until the migration to postgresql, which should be imminent.

Ivansete-status:
Good point indeed!
Thanks for bringing this @richard-ramos !

How long does it take if you perform _VACUUM_ regularly? For example, daily, or two consecutive _VACUUM_ calls? I have the impression that it took 11' because the vacuum operation needed to handle a lot of data.

I believe that the _VACUUM_ concept is widely used in other backends. In _Postgres_, the "standard" _VACUUM_ operation doesn't necessarily block other operations (see [this](https://www.postgresql.org/docs/current/routine-vacuuming.html#VACUUM-BASICS).) And we are performing it asynchronously.

Finally, if we decide that the _VACUUM_ operation should be made manually, then we will need to adapt the other retention policies (capacity and size.) as well.

CC: @NagyZoltanPeter @apentori

Ivansete-status:
Thanks for checking @richard-ramos , @yakimant !
Wow, it is super slow!

In this case, even though this PR is approved we can't merge it @ABresting. I think we need to close this PR and submit a different one where we remove the `VACUUM` calls from the "capacity" and "size" retention policies. We cannot perform VACUUM when working with _SQLite_. There shouldn't be issues in _Postgres_ because all is async there.

On the other hand, we need to create a separate chron script to perform the vacuum task separately under certain conditions. WDYT @richard-ramos , @yakimant ?

yakimant:
I have simmilar feelings, we need to address sqlite vacuuming differently.
Either by cron job or https://sqlite.org/pragma.html#pragma_auto_vacuum, but not a part of waku code.

richard-ramos:
I'm actually thinking of removing the vacuum functionality of go-waku because seems to me that having it as part of the node is too problematic.

ABresting:
closing this PR since due to SQLite it is better to not use automatic vacuuming.

PR #2166, opened 2023-10-27
chore: Minor Postgres optimizations

Ivansete-status:
# Description
This PR is intended to add some enhancements regarding the use of parallel connections.

# Changes

- [x] Increase the max limit of connections from 5 to 50 (hard-coded.)
- [x] Split the connection pool in two so that we ensure the select/read connection can benefit from the caching mechanisms provided by _Postgres_.
- [x] Increase the _Postgres_ health check time from 30'' to 60'' (waku/waku_archive/driver/postgres_driver/postgres_healthcheck.nim.)

## Issue


SionoiS (approved):
LGMT

If I understand correctly this would increase max connection count from 5 to 50x2 right?

Ivansete-status:
> If I understand correctly this would increase max connection count from 5 to 100 (50x2) right?

Yes, the database has 100 as max connections by default.

waku/waku_archive/driver/postgres_driver/postgres_driver.nim
```diff
@@ -233,7 +239,7 @@ proc getInt(s: PostgresDriver,
             Future[ArchiveDriverResult[int64]] {.async.} =
   # Performs a query that is expected to return a single numeric value (int64)
 
-  let rowsRes = await s.connPool.query(query)
+  let rowsRes = await s.readConnPool.query(query, newSeq[string](0))
```
gabrielmer, line 242: Why is it necessary to add here `newSeq[string](0)`?

waku/waku_archive/driver/postgres_driver/postgres_driver.nim
```diff
@@ -312,8 +318,13 @@ method deleteOldestMessagesNotWithinLimit*(
...
   ## Close the database connection
-  let result = await s.connPool.close()
-  return result
+  (await s.writeConnPool.close()).isOkOr:
+    return err("error closing write pool: " & $error)
+
+  (await s.readConnPool.close()).isOkOr:
+    return err("error closing read pool: " & $error)
+
+  return ok()
```
gabrielmer, line 328: If `s.writeConnPool.close()` fails, shouldn't we still attempt to run `s.readConnPool.close()` before returning?

waku/waku_archive/driver/postgres_driver/postgres_driver.nim
```diff
@@ -40,29 +42,33 @@ proc insertRow(): string =
  """INSERT INTO messages (id, storedAt, contentTopic, payload, pubsubTopic,
   version, timestamp) VALUES ($1, $2, $3, $4, $5, $6, $7);"""
 
-const MaxNumConns = 5 #TODO: we may need to set that from app args (maybe?)
+const MaxNumConns = 50 #TODO: we may need to set that from app args (maybe?)
```
gabrielmer, line 45: Random question. In most systems, there's significantly more read operations than write ones.

Does it makes sense to instead of dividing the max connections equally, to give a higher maximum to read ones and a lower one for write? Would it help in anything?

gabrielmer (commented):
Thanks so much! Looks great!
Added a few questions :):)

waku/waku_archive/driver/postgres_driver/postgres_driver.nim
```diff
@@ -312,8 +318,13 @@ method deleteOldestMessagesNotWithinLimit*(
...
   ## Close the database connection
-  let result = await s.connPool.close()
-  return result
+  (await s.writeConnPool.close()).isOkOr:
+    return err("error closing write pool: " & $error)
+
+  (await s.readConnPool.close()).isOkOr:
+    return err("error closing read pool: " & $error)
+
+  return ok()
```
gabrielmer, line 327: If `s.writeConnPool.close()` fails, shouldn't we still attempt to run `s.readConnPool.close()` before returning?

waku/waku_archive/driver/postgres_driver/postgres_driver.nim
Ivansete-status: Good point! :heart: 
However, in the tests made so far, it never reached the parallel 50 connection so for now I'd leave it in with that naïve separation until we find an empirical reason for that.

waku/waku_archive/driver/postgres_driver/postgres_driver.nim
Ivansete-status: Good point! I've pushed a change around that. Let me know if it looks better now :)

waku/waku_archive/driver/postgres_driver/postgres_driver.nim
Ivansete-status: Good catch! Not needed indeed :)
I'll change it now

vpavlin (approved):
LGTM

waku/waku_archive/driver/postgres_driver/postgres_driver.nim
gabrielmer: Looks amazing! Thanks! 😄

gabrielmer (approved):
LGTM! Thanks so much! :):)

PR #2167, opened 2023-10-27
prune shards with many peers

SionoiS:
# Description

# Changes

- [ ] ...
- [ ] ...


waku/node/peer_manager/peer_manager.nim
```diff
@@ -645,38 +646,79 @@ proc connectToRelayPeers*(pm: PeerManager) {.async.} =
...
+  let avg = min(1, total div max(1, peersByShard.len))
+
+  # prune peers from shard with higher than avg count
+  for shard, peers in peersByShard.pairs:
+    let count = max(peers.len - avg, 0)
+    for peer in peers[0..count]:
+      if peersToPrune.len >= pruningCount:
+        break
+
+      peersToPrune.incl(peer)
```
jm-clius, line 711: On thing to remember is that these peers more than likely also support other shards that we may be interested in, so removing them here also removes them from the other supported shards' slots. Not sure if it's a good idea to make this more complicated in this PR, but a possible issue I foresee is that peers that support more shards (good!) are more likely to be pruned overall. Perhaps a future improvement is to sort the peers here by number of supported shards and prune the peers first that support fewer/no other shards?

jm-clius (approved):
Roughly looks fine to me. I cannot immediately think of another simply mechanism to prune. However, simulating this (and other conn/peer behaviour) becomes more and more important to really understand the impact here. For example, mentioned in comment below that I think this will make us more likely to prune peers that support multiple shards, which should in fact be a reason for us to prefer to keep them.

waku/node/peer_manager/peer_manager.nim
SionoiS: > One thing to remember is that these peers more than likely also support other shards that we may be interested in, so removing them here also removes them from the other supported shards' slots. Not sure if it's a good idea to make this more complicated in this PR, but a possible issue I foresee is that peers that support more shards (good!) are more likely to be pruned overall. Perhaps a future improvement is to sort the peers here by number of supported shards and prune the peers first that support fewer/no other shards?

On the flip side, peers that support less shard might be more decentralized (not running in data center with beefy hardware).

Peer that support all shards == central server (exaggeration but still). We have to be careful about what we incentivize.

waku/node/peer_manager/peer_manager.nim
```diff
@@ -645,38 +646,79 @@ proc connectToRelayPeers*(pm: PeerManager) {.async.} =
...
+    if peersToPrune.len >= pruningCount:
       break
-    pm.peerStore.del(peerId)
-    pruned += 1
 
-  let afterNumPeers = toSeq(pm.peerStore[AddressBook].book.keys).len
+    peersToPrune.incl(peer)
+
+  # calculate the avg peers per shard
+  var total = 0
```
alrevuelta, line 699: just for the fun

`let total = sum(toSeq(y.values).mapIt(it.len))`

alrevuelta (approved):
lgtm!

waku/node/peer_manager/peer_manager.nim
SionoiS: Thank you! I tried so many thing that didn't work or were awkward.

SionoiS:
Blocked by #2151

PR #2168, opened 2023-10-30
chore(networking): lower dhigh to limit amplification factor

alrevuelta:
This PR lowers gossipsub's `Dhigh` value so that nodes have a more predictable amplification factor (in terms of bandwidth). Before `4 to 12` now `4 to 8`.

Used [this tool](https://github.com/waku-org/research/tree/master/rln-delay-simulations) to run simulations using 1000 nodes and propagation time should not be affected, since average `D` in the network stays the same.

**before**: mean of D among 1000 randomly connected nodes 6.129. Message latency average 509 ms (8991 samples).
```
Statistics. mean=6.129 max=12 min=4
Statistics. mean=509.4595706817929 max=695 min=263
```

**after this PR**: mean of D among 1000 randomly connected nodes 6.124. Message latency average 506 ms  (8991 samples). But with the main difference that max D is `8` instead of `12`.
```
Statistics. mean=6.124 max=8 min=4
Statistics. mean=506.74274274274273 max=677 min=263
```

In other words, it shouldn't affect the network performance, but it will decrease bandwidth consumption in some nodes by around 30% (12 -> 8)

**note**: In the future we may want to lower both Dlow/Dhigh even more, to limit bandwidth consumption, or at least to some "profiles". But this indeed has an impact in propagation delay. More https://github.com/waku-org/research/issues/44

### DIFF
--- a/waku/waku_relay/protocol.nim
+++ b/waku/waku_relay/protocol.nim
@@ -72,7 +72,7 @@ const GossipsubParameters = GossipSubParams(
 
     d: 6,
     dLow: 4,
-    dHigh: 12,
+    dHigh: 8,
     dScore: 6,
     dOut: 3,
     dLazy: 6,
@@ -174,8 +174,8 @@ proc new*(T: type WakuRelay, switch: Switch): WakuRelayResult[T] =
 
   return ok(w)
 
-proc addValidator*(w: WakuRelay, 
-                   topic: varargs[string], 
+proc addValidator*(w: WakuRelay,
+                   topic: varargs[string],
                    handler: WakuValidatorHandler) {.gcsafe.} =
   for t in topic:
     w.wakuValidators.mgetOrPut(t, @[]).add(handler)
@@ -196,7 +196,7 @@ proc subscribedTopics*(w: WakuRelay): seq[PubsubTopic] =
 
 proc generateOrderedValidator*(w: WakuRelay): auto {.gcsafe.} =
   # rejects messages that are not WakuMessage
-  let wrappedValidator = proc(pubsubTopic: string, 
+  let wrappedValidator = proc(pubsubTopic: string,
                               message: messages.Message): Future[ValidationResult] {.async.} =
     # can be optimized by checking if the message is a WakuMessage without allocating memory
     # see nim-libp2p protobuf library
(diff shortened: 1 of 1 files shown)


SionoiS:
I'm a bit uneasy about this change. I feel like the intent of `d` is more than just amplification, it protect mesh connectivity from degrading too much (also protect from eclipse attack).

What is the relation between churn rate and propagation time and what about message loss rate?

SionoiS (approved):

jm-clius (approved):
LGTM.

alrevuelta:
> I'm a bit uneasy about this change. I feel like the intent of d is more than just amplification, it protect mesh connectivity from degrading too much (also protect from eclipse attack). What is the relation between; churn rate, propagation time and message loss rate?

Note that Dlow and D have not changed, and in the simulations reported above, I can verify that the average D in the network stays the same. So I would say there is no impact, or at least not significant. The main difference is that before there were on average 12-6=6 "mesh slots" per node and now 8-6=2. Guess this affects new joiners, but as long as it is > 0, we should be okay.

I would feel uneasy about going as low as in https://github.com/waku-org/research/issues/44, but the change of this PR is quite convervative.

SionoiS:
> Note that Dlow and D have not changed, and in the simulations reported above, I can verify that the average D in the network stays the same. So I would say there is no impact, or at least not significant. The main difference is that before there were on average 12-6=6 "mesh slots" per node and now 8-6=2. Guess this affects new joiners, but as long as it is > 0, we should be okay.
> 
> I would feel uneasy about going as low as in [waku-org/research#44](https://github.com/waku-org/research/issues/44), but the change of this PR is quite convervative.

Perfect! Thanks.

alrevuelta:
Interesting to see the effect of this live at: simulator.waku.org. Seems that D changes a bit more in some nodes. It stays within the expected values but is worth noting.
image

jm-clius:
> Seems that D changes a bit more in some nodes

This seems strange and I have no immediate hypothesis about why this might happen. The worry would be that more churn = more bandwidth (due to control messages) which may diminish the gains we get from a lower D.

alrevuelta:
> Seems that D changes a bit more in some nodes

image

It looks like it was temporal, unsure why though. Right now the behavior seems like it should. Will monitor just in case.

jm-clius:
Ah, much better 😌

PR #2169, opened 2023-10-30
fix(rest): fix bug in rest api when sending rln message

alrevuelta:
Description:
* Fix bug in rest api causing rln messages to be rejected.
* Message without proof was being included due to a minor bug.


gabrielmer (approved):
LGTM. Tried it and fixes the issue :)

Thanks so much!

vpavlin (approved):
Nice catch!

PR #2171, opened 2023-10-30
chore: Optimize postgres - use of rowCallback approach

Ivansete-status:
# Description
This represents a quite important time reduction in _Postgres_ queries. However, still slower than _SQLite_.
The main issue happened because the we were processing the obtained rows multiple times, and there were an overhead due to changing the rows objects from one type to another.
In this PR we aim to reduce the type conversion overhead by providing a `rowCallback(result: ptr PGresult)` to the `dbconn` object.

( Thanks to @NagyZoltanPeter for the help to achieve that. )

# Changes

- [x] Use of `rowCalback` to retrieve rows in a more direct way.
- [x] Change logic on how to retrieve the first free connection - waku/common/databases/db_postgres/pgasyncpool.nim

## Issue

### DIFF
--- a/waku/common/databases/db_postgres/pgasyncpool.nim
+++ b/waku/common/databases/db_postgres/pgasyncpool.nim
@@ -167,13 +167,13 @@ proc releaseConn(pool: PgAsyncPool, conn: DbConn) =
     if pool.conns[i].dbConn == conn:
       pool.conns[i].busy = false
 
-proc query*(pool: PgAsyncPool,
-            query: string,
-            args: seq[string] = newSeq[string](0),
-            rowCallback: DataProc):
-            Future[DatabaseResult[void]] {.async.} =
-  ## Runs the SQL query getting results.
-  ## Retrieves info from the database.
+proc pgQuery*(pool: PgAsyncPool,
+              query: string,
+              args: seq[string] = newSeq[string](0),
+              rowCallback: DataProc = nil):
+              Future[DatabaseResult[void]] {.async.} =
+  ## rowCallback != nil when it is expected to retrieve info from the database.
+  ## rowCallback == nil for queries that change the database state.
 
   let connIndexRes = await pool.getConnIndex()
   if connIndexRes.isErr():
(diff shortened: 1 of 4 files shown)


NagyZoltanPeter (approved):
Read through and totally awesome work! Thank you!!!

NagyZoltanPeter:
@Ivansete-status: One concern I thought about, but not connected directly to this PR.
The query where clause processing (aka string replacing arguments) maybe harmful. I did not look through the whole data flow, but as this is fully exposed to the interface level, it might be possible that one can achieve SQL injection attack against the db.
Maybe it is worth to track down later.
Also maybe to consider to switch to prepared statements and execution if possible, that would mean some little perf gain on DB side either.

Ivansete-status:
> Also maybe to consider to switch to prepared statements and execution if possible, that would mean some little perf gain on DB side either.

Thanks for the comment!
In fact, today I was planning to do exactly that and use prepared statements in the `select` queries. The thing is that when we have concurrent queries, each connection stays in "busy" state (`db.pqisBusy() == 1`) for a lot of time, once the query has been sent (`db.pqsendQuery(...)`.)

I agree to review the "SQL injection" issue in the future.

NagyZoltanPeter:
> > Also maybe to consider to switch to prepared statements and execution if possible, that would mean some little perf gain on DB side either.
> 
> Thanks for the comment!
> In fact, today I was planning to do exactly that and use prepared statements in the `select` queries. The thing is that when we have concurrent queries, each connection stays in "busy" state (`db.pqisBusy() == 1`) for a lot of time, once the query has been sent (`db.pqsendQuery(...)`.)
> 
> I agree to review the "SQL injection" issue in the future.
> 
> 

Indeed, prepared statement secure that one as well.

waku/common/databases/db_postgres/pgasyncpool.nim
```diff
@@ -8,15 +8,11 @@ else:
...
   std/[sequtils,nre, strformat],
   stew/results,
-  chronicles,
   chronos
 import
   ./dbconn,
   ../common
 
-logScope:
-  topics = "postgres asyncpool"
```
SionoiS, line 18: We don't log anything anymore? Or was it not used.

SionoiS (approved):
LGTM

Thanks!

waku/common/databases/db_postgres/pgasyncpool.nim
Ivansete-status: Good point! It was not used. In fact, all the exposed `procs` return `Result` so it is pretty likely that we won't need many longs in this file.

PR #2176, opened 2023-11-01
fix: lightpush rest 

Ivansete-status:
## Description
This PR aims to enhance the `lightpush rest` API a little.

This was detected by @weboko. Thanks for spotting it !

## Changes

- [x] Fix error response
- [x] Fix typo in openapi.yaml definition

## Future notes
- We need to review all the open API definitions and make sure all the cases work as expected.
- Important to define examples of how to use each API. Maybe something that can be written within each open API.
- It is not very handy to import openapi.yaml file into _Postman_. The parameter types are not properly described. For example, in the case of `lightpush`, we should make it clear that the `payload` should be base64-encoded. On the other hand, the numeric fields are imported (in Postman) as `"<number>"` but that confuses a little and it should be `<number>` instead, or having an example would help.

### DIFF
--- a/waku/waku_api/rest/lightpush/openapi.yaml
+++ b/waku/waku_api/rest/lightpush/openapi.yaml
@@ -76,7 +76,7 @@ components:
     PushRequest:
       type: object
       properties:
-        pusbsubTopic:
+        pubsubTopic:
           $ref: '#/components/schemas/PubsubTopic'
         message:
           $ref: '#/components/schemas/WakuMessage'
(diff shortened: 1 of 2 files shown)


gabrielmer (approved):
LGTM, thanks! 🤩

waku/waku_api/rest/lightpush/openapi.yaml
```diff
@@ -76,7 +76,7 @@ components:
     PushRequest:
       type: object
       properties:
-        pusbsubTopic:
+        pubsubTopic:
```
NagyZoltanPeter, line 79: Great catch!
Notice that we must correct it here as well:
   In  Waku-org/waku-rest-api repository.

NagyZoltanPeter (approved):
Thanks for it

PR #2180, opened 2023-11-02
Revert "feat: amending computeDigest func. + related test cases (#2132)"

ABresting:
This reverts commit 1669f710ce8f48d1b7fb06ca98433f2b25a44e6d.

# Description

As discussed with Hanno earlier, this will revert the `computeDigest` introduced in Waku archive and associated test cases changes as well.
We are doing this to make sure that we keep original behavior intact and use waku_core https://github.com/waku-org/nwaku/blob/master/waku/waku_core/message/digest.nim#L29 for computing `messageHash` attribute of database.

# Changes

- [x] revert the commit

## Issue

#2112

### DIFF
--- a/tests/waku_archive/test_driver_postgres.nim
+++ b/tests/waku_archive/test_driver_postgres.nim
@@ -19,7 +19,7 @@ proc computeTestCursor(pubsubTopic: PubsubTopic,
     pubsubTopic: pubsubTopic,
     senderTime: message.timestamp,
     storeTime: message.timestamp,
-    digest: computeDigest(message, pubsubTopic)
+    digest: computeDigest(message)
   )
 
 suite "Postgres driver":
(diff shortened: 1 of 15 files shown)


jm-clius (approved):
I'm a bit confused, as I didn't realise that this was already merged into master. These changes seemed to still be part of the recently opened PR that changed the DB schema, but perhaps it was simply a polluted diff?
It would be good to understand what the actual impact is in terms of client implementation backwards compatibility if the underlying cursor is changed.

Ivansete-status (approved):

ABresting:
> I'm a bit confused, as I didn't realise that this was already merged into master. These changes seemed to still be part of the recently opened PR that changed the DB schema, but perhaps it was simply a polluted diff? It would be good to understand what the actual impact is in terms of client implementation backwards compatibility if the underlying cursor is changed.

This was a separate PR since on for issue #2112 tasks are done serially to have a visible separation component-wise merging in, that PR modified the `computeDigest` functionality.

AFAIK, this change only improved the cursor functionality since with new `digest`/`id` computed as per the [RFC](https://rfc.vac.dev/spec/14/#deterministic-message-hashing) value, in js-Waku it failed a test case value and it was due to the existing digest functionality not updated in the jsWaku client. 

in jsWaku and goWaku client it passed all test cases. About backwards compatibility idk tbh.

PR #2182, opened 2023-11-03
chore: Optimize postgres - prepared statements in select

Ivansete-status:
# Description
The [prepared statements](https://www.postgresql.org/docs/current/sql-prepare.html#:~:text=A%20prepared%20statement%20is%20a,statement%20is%20planned%20and%20executed.) is a mechanism that allows saving time for common queries.

We are applying that approach for the "select" queries, i.e. queries that come from the _Store_ protocol requests. Notice that we were already applying this in "insert" operations.

With that, we enhance the performance of "select" queries. However, bear in mind that there is an underlying time consumption when performing concurrent queries. I've added a long comment about that in `dbconn.nim`.

# Changes

- [x] Use of prepared statements in select queries. 

## Issue

### DIFF
--- a/waku/common/databases/db_postgres/dbconn.nim
+++ b/waku/common/databases/db_postgres/dbconn.nim
@@ -138,7 +138,7 @@ proc waitQueryToFinish(db: DbConn,
     if success != 1:
       let checkRes = db.check()
       if checkRes.isErr():
-        return err("failed pqconsumeInput: " & checkRes.error)
+        return err("failed pqconsumeInput: " & $checkRes.error)
 
       return err("failed pqconsumeInput: unknown reason")
 
@@ -149,6 +149,10 @@ proc waitQueryToFinish(db: DbConn,
     let pqResult = db.pqgetResult()
 
     if pqResult == nil:
+      let checkRes = db.check()
+      if checkRes.isErr():
+        return err("error in query: " & $checkRes.error)
+
       return ok() # reached the end of the results
 
     if not rowCallback.isNil():
--- a/waku/common/databases/db_postgres/dbconn.nim
+++ b/waku/common/databases/db_postgres/dbconn.nim
@@ -53,9 +53,8 @@ proc sendQuery(db: DbConn,
   ## This proc can be used directly for queries that don't retrieve values back.
 
   if db.status != CONNECTION_OK:
-    let checkRes = db.check()
-    if checkRes.isErr():
-      return err("failed to connect to database: " & checkRes.error)
+    db.check().isOkOr:
+      return err("failed to connect to database: " & $error)
 
     return err("unknown reason")
 
@@ -68,9 +67,8 @@ proc sendQuery(db: DbConn,
 
   let success = db.pqsendQuery(cstring(wellFormedQuery))
   if success != 1:
-    let checkRes = db.check()
-    if checkRes.isErr():
-      return err("failed pqsendQuery: " & checkRes.error)
+    db.check().isOkOr:
+      return err("failed pqsendQuery: " & $error)
 
     return err("failed pqsendQuery: unknown reason")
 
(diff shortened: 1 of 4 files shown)


jm-clius (approved):
LGTM! :)

waku/common/databases/db_postgres/dbconn.nim
```diff
@@ -75,32 +76,82 @@ proc sendQuery(db: DbConn,
...
+               paramValues: openArray[string],
+               paramLengths: openArray[int32],
+               paramFormats: openArray[int32]):
+               Result[void, string] =
+  ## This proc can be used directly for queries that don't retrieve values back.
+
+  if paramValues.len != paramLengths.len or paramValues.len != paramFormats.len or
+     paramLengths.len != paramFormats.len:
+    let lengthsErrMsg = $paramValues.len & " " & $paramLengths.len & " " & $paramFormats.len
+    return err("lengths discrepancies in sendQueryPrepared: " & $lengthsErrMsg)
```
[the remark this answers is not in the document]
Ivansete-status: Good catch! Thanks for it!

The thing is that we need to keep those three arrays separately so that we can pass their addresses to https://www.postgresql.org/docs/current/libpq-async.html#LIBPQ-PQSENDQUERYPREPARED

If you agree, we could clean that up by defining a `ParamList` type and then store the lists there.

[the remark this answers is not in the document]
SionoiS: > The thing is that we need to keep those three arrays separately so that we can pass their addresses to https://www.postgresql.org/docs/current/libpq-async.html#LIBPQ-PQSENDQUERYPREPARED

:thinking: 

> If you agree, we could clean that up by defining a `ParamList` type and then store the lists there.

That would kinda defeat the purpose.

Disregard my comment, it's not really practical.

SionoiS (approved):

waku/common/databases/db_postgres/dbconn.nim
```diff
@@ -67,40 +67,85 @@ proc sendQuery(db: DbConn,
...
-  while true:
-
+  while db.pqisBusy() == 1:
+    ## TODO: Enhance performance in concurrent queries.
+    ## The connection keeps busy for quite a long time when performing intense concurrect queries.
+    ## For example, a given query can last 11 milliseconds within from the database point of view
+    ## but, on the other hand, the connection remains in "db.pqisBusy() == 1" for 100ms more.
+    ## I think this is because `nwaku` is single-threaded and it has to handle many connections (20)
+    ## simultaneously. Therefore, there is an underlying resource sharing (cpu) that makes this
+    ## to happen. Notice that the _Postgres_ database spawns one process per each connection.
```
NagyZoltanPeter, line 131: I appreciate this great, explanatory comment here!!!

waku/waku_archive/driver/postgres_driver/postgres_driver.nim
```diff
@@ -240,6 +295,114 @@ method getMessages*(s: PostgresDriver,
...
+                                   int32(endTimeStr.len),
+                                   int32(limit.len)],
+                                 @[int32(0), int32(0), int32(0), int32(0), int32(0)],
+                                  rowCallback)
+    ).isOkOr:
+      return err("failed to run query without cursor: " & $error)
+
+  return ok(rows)
+
+method getMessages*(s: PostgresDriver,
```
NagyZoltanPeter, line 373: How come PostgresDriver reference named as "s"? :-)

NagyZoltanPeter (approved):
All good! Special thanks that this approach prevents sql injection attacks.

waku/waku_archive/driver/postgres_driver/postgres_driver.nim
Ivansete-status: Thanks for the comments! I'll review that in upcoming refactoring PRs :)

PR #2187, opened 2023-11-06
chore: improving node setup w.r.t sharding information

### LINKED ISSUE #2183: chore: improve cluster id, shards, topics flow
### Background
I feel like the information flow for cluster and shard could be simplified.

### Details
Cluster id and shards are both specified in the config. What happen if they disagree?
Waku metadata, relay and discv5 must be updated and not conflict with each other WRT shards.

### Acceptance criteria
- [ ] Disallow conflicting cluster Id and topics
- [ ] Refactor testing so that cluster and shard are easy to setup.
- [ ] Simplify cluster and shard setup for all protocols.

SionoiS:
# Description

Various misc. fixes and changes related to shards, Waku metadata and testings.

Motivated by #2151 which made me realized testing was not ergonomic w.r.t. sharding.

# Changes

- [x] added shards info to Waku metadata protocol setup
- [x] refactor Discv5 setup
- [x] added check for cluster id mismatch between config and ENR
- [x] change default topic for tests to shard 0 in cluster 2
- [x] added param to tests nodes allowing extra shards
- [x] tests nodes ENRs are now updated with sharding info  

Close #2183

### DIFF
--- a/apps/wakunode2/app.nim
+++ b/apps/wakunode2/app.nim
@@ -127,6 +127,21 @@ proc init*(T: type App, rng: ref HmacDrbgContext, conf: WakuNodeConf): T =
       quit(QuitFailure)
     else: recordRes.get()
 
+  # Check the ENR sharding info for matching config cluster id
+  if conf.clusterId != 0:
+    let res = record.toTyped()
+    if res.isErr():
+      error "ENR setup failed", error = $res.get()
+      quit(QuitFailure)
+
+    let relayShard = res.get().relaySharding().valueOr:
+      error "no sharding info"
+      quit(QuitFailure)
+
+    if conf.clusterId != relayShard.clusterId:
+      error "cluster id mismatch"
+      quit(QuitFailure)
+
   App(
     version: git_version,
     conf: conf,
@@ -234,7 +249,13 @@ proc setupDiscoveryV5*(app: App): WakuDiscoveryV5 =
     autoupdateRecord: app.conf.discv5EnrAutoUpdate,
   )
 
-  WakuDiscoveryV5.new(app.rng, discv5Conf, some(app.record))
+  WakuDiscoveryV5.new(
+    app.rng, 
+    discv5Conf,
+    some(app.record),
+    some(app.node.peerManager),
+    app.node.topicSubscriptionQueue,
+  )
 
 ## Init waku node instance
 
@@ -286,17 +307,16 @@ proc initNode(conf: WakuNodeConf,
   ok(node)
 
 proc setupWakuApp*(app: var App): AppResult[void] =
-
-  ## Discv5
-  if app.conf.discv5Discovery:
-    app.wakuDiscV5 = some(app.setupDiscoveryV5())
-
   ## Waku node
   let initNodeRes = initNode(app.conf, app.netConf, app.rng, app.key, app.record, app.peerStore, app.dynamicBootstrapNodes)
   if initNodeRes.isErr():
     return err("failed to init node: " & initNodeRes.error)
 
   app.node = initNodeRes.get()
+
+  ## Discv5
+  if app.conf.discv5Discovery:
+    app.wakuDiscV5 = some(app.setupDiscoveryV5())
 
   ok()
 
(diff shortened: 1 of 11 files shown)


apps/wakunode2/app.nim
```diff
@@ -587,25 +620,24 @@ proc startNode(node: WakuNode, conf: WakuNodeConf,
...
 
-  if app.wakuDiscv5.isSome():
-    let wakuDiscv5 = app.wakuDiscv5.get()
-
-    let res = wakuDiscv5.start()
-    if res.isErr():
-      return err("failed to start waku discovery v5: " & $res.error)
+  if (let wakuDiscv5 =  app.wakuDiscv5.get(); app.wakuDiscv5.isSome()):
+    let wakudiscv5Res = catch: (waitFor wakuDiscv5.start())
+    if wakudiscv5Res.isErr():
```
alrevuelta, line 636: Isn't this weird?

First we check `wakudiscv5Res.isErr()`

And the `wakudiscv5Res.get().isOkOr:`

Can't we have just one? Either it errors or not. But checking if getting the error itself errored, seems weird?

apps/wakunode2/app.nim
```diff
@@ -127,6 +127,21 @@ proc init*(T: type App, rng: ref HmacDrbgContext, conf: WakuNodeConf): T =
...
+      error "ENR setup failed", error = $res.get()
+      quit(QuitFailure)
+
+    let relayShard = res.get().relaySharding().valueOr:
+      error "no sharding info"
+      quit(QuitFailure)
+
+    if conf.clusterId != relayShard.clusterId:
+      error "cluster id mismatch"
+      quit(QuitFailure)
```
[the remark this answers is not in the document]
SionoiS: It's a bit weird to have 2 ways to specify the cluster Id.

`conf.clusterId` 0 is a special case that mean disregard cluster Ids. It's for backward compatibility.
For context: https://github.com/waku-org/nwaku/pull/2062#issuecomment-1746328936

The conf should have precedence IMO.

apps/wakunode2/app.nim
SionoiS: Good catch! It's weird indeed

gabrielmer (approved):
LGTM! Thank you :)

waku/waku_metadata/protocol.nim
```diff
@@ -91,11 +93,27 @@ proc initProtocolHandler*(m: WakuMetadata) =
   m.handler = handle
   m.codec = WakuMetadataCodec
 
+proc clusterAndShards(enr: Record): Option[(uint32, HashSet[uint32])] =
```
jm-clius, line 96: Does this belong in the waku_enr module?

apps/wakunode2/app.nim
```diff
@@ -127,6 +127,21 @@ proc init*(T: type App, rng: ref HmacDrbgContext, conf: WakuNodeConf): T =
       quit(QuitFailure)
     else: recordRes.get()
 
+  # Check the ENR sharding info for matching config cluster id
```
jm-clius, line 130: We may want to figure out some method of configuration that doesn't allow for inconsistencies here (i.e. only configuring cluster ID and shards, deprecating any `--pubsub-topic` configuration)

jm-clius (approved):
Looks reasonable to me. :)

apps/wakunode2/app.nim
SionoiS: Yes having 2 ways of setting the cluster id is not super ergonomic.

waku/waku_metadata/protocol.nim
SionoiS: Well no, maybe a rename? I needed a proc to either give the cluster and shards in the ENR OR just the cluster from the config.

SionoiS:
I don't get it! All test green locally but it always fail here......

SionoiS:
**Weekly Update**

- _achieved_: Various tests updates and fixes.
- _next_: Figure out why CI passes locally only.

SionoiS:
Seams like the bug is fixed. Merging even with JS failing as it's not related.

PR #2199, opened 2023-11-08
fix: rln_keystore_generator asks for creating credentials folder if doesn't exist

Ivansete-status:
# Description
The `rln_keystore_generator` tool stores the user credentials in a file passed by the `rln-relay-cred-path` param.
If the users passes a path which folder doesn't exist, then the next error appears and the credentials aren't properly stored:

```
ERR 2023-11-08 17:25:50.598+01:00 failed to persist credentials              topics="rln_keystore_generator" tid=240725 file=rln_keystore_generator.nim:100 error="keystore error: OS specific error: Cannot open file for writing"
```

This PR enhances that situation by ending the app immediately and informing the user that the credentials folder doesn't exist.

# Changes

- [x] rln_keystore_generator informs the user when the credential folder doesn't exist and ends.

### DIFF
--- a/tools/rln_keystore_generator/rln_keystore_generator.nim
+++ b/tools/rln_keystore_generator/rln_keystore_generator.nim
@@ -4,9 +4,10 @@ else:
   {.push raises: [].}
 
 import
+  os,
   chronicles,
   stew/[results],
-  std/tempfiles
+  std/[tempfiles,strutils]
 
 import
   ../../waku/waku_keystore,
@@ -91,6 +92,11 @@ when isMainModule:
     treeIndex: groupManager.membershipIndex.get(),
     identityCredential: credential,
   )
+
+  let baseDir = conf.rlnRelayCredPath.splitFile().dir
+  if baseDir.len > 0 and not existsDir(baseDir):
+    info "Creating credentials keystore folder", dir=baseDir
+    createDir(baseDir)
 
   let persistRes = addMembershipCredentials(conf.rlnRelayCredPath,
                                             keystoreCred,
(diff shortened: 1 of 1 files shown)


alrevuelta:
Not sure if this is a thing that nwaku should do. If you pass a path that doesn't exist, IMHO the expected behavior is that it should fail.

Ivansete-status:
> In addition to this, maybe instead of failing _after_ the membership transaction, we should check if the path exists before kicking off any operations (id credential generation, etc) I can take this up @Ivansete-status

This is exactly what is happening in this PR. That validation is made at the beginning :)

tools/rln_keystore_generator/rln_keystore_generator.nim
```diff
@@ -30,6 +31,12 @@ when isMainModule:
 
   trace "configuration", conf = $conf
 
+  let baseDir = conf.rlnRelayCredPath.splitFile().dir
+  if baseDir.len > 0 and not existsDir(baseDir):
+    let errMsg = "The next folder doesn't exist: " & baseDir
+    error "failure missing credentials folder", error=errMsg
+    quit(1)
+
```
SionoiS, line 39: When running for the first time is the dir created before this check? Otherwise it can never pass.

Ivansete-status: Yes, the folder should be created manually beforehand.

rymnc: while I agree that the check should be made, we should not create the directory, it is not the concern of the application to do it

SionoiS (approved):

Ivansete-status:
I close this PR because this is not needed in the end.
It now works the `./register_rln.sh` script from within the next `nwaku-compose` commit: https://github.com/waku-org/nwaku-compose/commit/3d58809874940493afaa2f2a605f49364a834107

I probably was having issues because my local `register_rln.sh` script wasn't using $(pwd)

PR #2203, opened 2023-11-09
chore: decoupling announced and listen addresses

### LINKED ISSUE #2148: chore: decouple listen and announced addresses
### Details
Currently our  "Listening on..." log does not show the addresses we are actually listening on - it prints the addresses that we announce.
This is a bit misleading and therefore, we want to separate it into two different log prints: one that includes the addresses our node is listening on and another with the addresses the node wants to announce to the outside world.

For the listen addresses, we also want to include the node's local IP, which is currently not displayed in nwaku. The local IP is logged in go-waku and it's an address where our node is reachable. Therefore, it makes sense to also include it in nwaku, both for consistency and convenience.

### Acceptance criteria
- [x] two different logs are printed to the console, one with listen addresses and other with announced addresses
- [x] local IP is included in the listen addresses log

gabrielmer:
# Description

Currently, our logs don't distinguish between the addresses we announce to the outside world vs the addresses our node is actually listening on. We used to only print "Listening on", and included there only the addresses we announced.

Fixed wording and added a log with the addresses our node is bound to.
In addition, included the LAN IP in the new log for convenience.

# Changes

- [x] moved network information logs to be printed after the switch binds and starts listening
- [x] separated logging for announced addresses and listening addresses
- [x] added LAN IP to the logs 

image

## Issue

closes #2148


waku/node/waku_node.nim
```diff
@@ -1082,16 +1083,29 @@ proc isBindIpWithZeroPort(inputMultiAdd: MultiAddress): bool =
...
 proc printNodeNetworkInfo*(node: WakuNode): void =
   let peerInfo = node.switch.peerInfo
+  var announcedStr = ""
   var listenStr = ""
+  var localIp = ""
+
+  try:
+    localIp = $getPrimaryIPAddr()
+  except Exception as e:
+    warn "Could not retrieve localIp", msg=e.msg
```
SionoiS, line 1093:
```suggestion
  let res = catch: getPrimaryIPAddr()
  localIp = $res.valueOr:
    warn "Could not retrieve localIp", msg=error.msg
```
I like `catch:` it makes `try except` more concise and less drift. https://github.com/arnetheduck/nim-results/blob/f3c666a272c69d70cb41e7245e7f6844797303ad/results.nim#L723

SionoiS (approved):
Thank you!

waku/node/waku_node.nim
gabrielmer: Ooh that looks better, will do it that way! Thank you!

NagyZoltanPeter (approved):
Nice one! LGTM!

PR #2205, opened 2023-11-09
fix(rest): properly check if rln is used

vpavlin:
# Description

Fixing leftover check for `define(rln)` from when RLN was compiled in conditionally.

Using the option to get a boolean instead

# Changes

- [ ] ...
- [ ] ...

### DIFF
--- a/waku/waku_api/rest/relay/handlers.nim
+++ b/waku/waku_api/rest/relay/handlers.nim
@@ -304,7 +304,7 @@ proc installRelayApiHandlers*(router: var RestRouter, node: WakuNode, cache: Mes
         return RestApiResponse.internalServerError("Failed to publish: unknown RLN proof validation result")
 
     # if we reach here its either a non-RLN message or a RLN message with a valid proof
-    debug "Publishing message", contentTopic=message.contentTopic, rln=defined(rln)
+    debug "Publishing message", contentTopic=message.contentTopic, rln=not node.wakuRlnRelay.isNil()
     if not (waitFor node.publish(none(PubSubTopic), message).withTimeout(futTimeout)):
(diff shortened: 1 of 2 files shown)


waku/waku_api/rest/relay/handlers.nim
```diff
@@ -174,7 +174,7 @@ proc installRelayApiHandlers*(router: var RestRouter, node: WakuNode, cache: Mes
         return RestApiResponse.internalServerError("Failed to publish: unknown RLN proof validation result")
 
     # if we reach here its either a non-RLN message or a RLN message with a valid proof
-    debug "Publishing message", pubSubTopic=pubSubTopic, rln=defined(rln)
+    debug "Publishing message", pubSubTopic=pubSubTopic, rln=not node.wakuRlnRelay.isNil()
     if not (waitFor node.publish(some(pubSubTopic), message).withTimeout(futTimeout)):
       error "Failed to publish message to topic", pubSubTopic=pubSubTopic
       return RestApiResponse.internalServerError("Failed to publish: timedout")
```
alrevuelta, line 180: can we fix it below as well?
https://github.com/waku-org/nwaku/blob/a8763a3525b4439a87f7fb119f57f8d5289c0657/waku/waku_api/rest/relay/handlers.nim#L307

vpavlin: Good point, missed that. It was also present in jsonrpc, so I fixed that too.

the `defined(rln)` is still used in `libwaku` (cc @Ivansete-status) - not touching that for this PR:)

alrevuelta (approved):
lgtm thanks!

SionoiS (approved):

PR #2209, opened 2023-11-09
chore: Allow text/plain content type descriptor for json formatted content body

NagyZoltanPeter:
# Description
@weboko requested that from browser no-coors mode we need to supprt text/plain content-body requests.

# Changes

- [x] For decoding content body we now allow text/plain content-type descriptor as well
- [x] Refactored rest type encode/decode functions to remove tons of code duplication.

## How to test

1. Setup a light client node.
2. Send normal rest requests (POST/PUT) with curl as text/plain
3. Should work as normal as with application/json content-type (notice no json format changed)

## Issue
https://github.com/waku-org/nwaku/issues/2207

### DIFF
--- a/waku/waku_api/rest/admin/client.nim
+++ b/waku/waku_api/rest/admin/client.nim
@@ -13,6 +13,7 @@ import
 import
   ../serdes,
   ../responses,
+  ../rest_serdes,
   ./types
 
 export types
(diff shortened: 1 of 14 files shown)


vpavlin (approved):
Didn't have time to actually run it, but from visual sanity check perspective it looks great (thanks for the refactoring!)

gabrielmer (approved):
LGTM, thanks so much!

NagyZoltanPeter:
As @weboko pointed out, I missed converting /relay apis to accept text/plain requests, it is done now with some simplifications

weboko:
Amazing, now it works. I noticed some of improvements that can be done:
- browsers sometimes for legacy support send `Content-Type` as `text/plain;charset=UTF-8` and since it is made to accept only [text/plain](https://github.com/waku-org/nwaku/blob/2cb0989a28e3b4e4e4f495f8cf05f88173fe6d81/waku/waku_api/rest/responses.nim#L16) it would fail with `400`;
- as additional precaution can we set `Access-Control-Allow-Origi` to `*`?

fryorcraken:
> as additional precaution can we set Access-Control-Allow-Origin to * or make it configurable by config flag;

Do not set it to `*` as it makes it a security issue as now any website can try to access a local running nwaku node using default port and would be authorized to then manipulate the nwaku node because the CORS would not stop it.

The CORS origin should only be `localhost`/`127.0.0.1`. @weboko, how are you hitting your webserver? I assume it's running in docker now? are you using the docker container ip to access the webserver? that might the issue, need to use the port exposed on localhost.

NagyZoltanPeter:
Added support for decoding content body when content-type header argument has more parameters above media type.

weboko (approved):

weboko:
> The CORS origin should only be localhost/127.0.0.1

Browser recognizes CORS based on [Same origin policy](https://developer.mozilla.org/en-US/docs/Web/Security/Same-origin_policy) which unfortunately takes port into account, so event if it is same domain but ports are different it violates `same origin` policy. 

If REST is intended to be user only locally - it should be enough with MIME change in this PR.

gabrielmer (approved):

fryorcraken:
> If REST is intended to be user only locally - it should be enough with MIME change in this PR.

Yes, REST should be user only locally.

PR #2212, opened 2023-11-10
fix(rln): error in api when rate limit

alrevuelta:
When using the rest api to publish messages with rln, eg:

```
curl -X POST "http://127.0.0.1:8645/relay/v1/auto/messages" \
 -H "content-type: application/json" \
 -d '{"payload":"'$(echo -n "Hello Waku Network - from Anonymous User" | base64)'","contentTopic":"/my-app/2/chatroom-1/proto"}'
```

It should error when a message is expected to be considered as spam by the network. This is very important, because otherwise the user would expect the message to arrive, while it was rejected due to being rate limited.

Using `validateMessage` does not work, because that doesn't update the nullifier log. Using `validateMessageAndUpdateLog` instead, which does the same + update the nullifier.

cc @rymnc


Ivansete-status (approved):
LGTM! Thanks :100: !

rymnc (approved):
LGTM, thanks for this!

Issue #2137, opened 2023-10-18
chore: enable rest api interface by default on wakunode


NagyZoltanPeter:
### Background
As rest interface is now close to feature ready this issue derived from a research discussion.
By having enable rest endpoint by default would help operators manage their nodes while still keep the opt for switch it off of set certain ip and port.

Also maybe related to two other chores:
- [chore: Rest API requests on not installed endpoints shall return meaningful message along with Http404](https://github.com/waku-org/nwaku/issues/2135)
- [chore: wakunode2: allow filter and lightpush clients without pre-specify service node](https://github.com/waku-org/nwaku/issues/2136)

NagyZoltanPeter:
> Not sure what would be the action item here. Just set `true` as the default value for `rest` flag?

I guess mostly yes! This is more a tracking issue and there are other related ones, mentioned above.
It is useful for track release content as well.
Maybe it can be connect also to removing json-rpc as planned? Any thoughts?

jm-clius:
With the REST API complete, I agree that it should be enabled by default.
Afaik, with the possible exception of the "Admin API", all our APIs are client-side APIs (e.g. filter _client_, store _client_, etc.). 
Since IMO all waku nodes (regardless of configuration) should be capable of acting as filter, store, lightpush, peer-exchange _clients_ (in other words, the necessary protocols should be mounted/configured) the APIs can all be enabled for any waku node.

PR #2218, opened 2023-11-15
fix: admin REST API to be enabled only if config is set

chaitanyaprem:
# Description
Admin REST API was enabled even if restadmin flag was set to false.

Note: Tested by running node locally and enabling/disabling the config flag to verify the availability of Admin REST API.

# Changes

- [ ] Included config check before installing Admin REST API handlers


NagyZoltanPeter (commented):
I have some generic thoughts on this PR.
One littel, and yes it is a general notice to myself also as me also running into this time-to-time. It is better avoid general formatting changes when having a fix PR. Maybe you have some auto-formatting enabled, idk. But it makes harder to find and review the meaningful change.

On the topic of the PR. 
I think this change is opposed to https://github.com/waku-org/nwaku/issues/2137
Also it might be my fault and did not checked with enough care those cli switches, but did not expect that for a generic endpoint like admin (and health, debug, ...) we need such distinction. It is ok that some endpoint not mounted as the underlying feature not present (lightpush, filter,...). 
So I think in general we can have /admin by default if having --rest=true.

@jm-clius , @fryorcraken : WDYT?

chaitanyaprem:
> I have some generic thoughts on this PR.
> One littel, and yes it is a general notice to myself also as me also running into this time-to-time. It is better avoid general formatting changes when having a fix PR. Maybe you have some auto-formatting enabled, idk. But it makes harder to find and review the meaningful change.
> 
> On the topic of the PR. 
> I think this change is opposed to https://github.com/waku-org/nwaku/issues/2137
> Also it might be my fault and did not checked with enough care those cli switches, but did not expect that for a generic endpoint like admin (and health, debug, ...) we need such distinction. It is ok that some endpoint not mounted as the underlying feature not present (lightpush, filter,...). 
> So I think in general we can have /admin by default if having --rest=true.
> 
> @jm-clius , @fryorcraken : WDYT? 

Sorry, did not realise my IDE did all changes. 
Let me revert and only push the 2 lines change I actually did.

NagyZoltanPeter:
> > I have some generic thoughts on this PR.
> > One littel, and yes it is a general notice to myself also as me also running into this time-to-time. It is better avoid general formatting changes when having a fix PR. Maybe you have some auto-formatting enabled, idk. But it makes harder to find and review the meaningful change.
> > On the topic of the PR.
> > I think this change is opposed to #2137
> > Also it might be my fault and did not checked with enough care those cli switches, but did not expect that for a generic endpoint like admin (and health, debug, ...) we need such distinction. It is ok that some endpoint not mounted as the underlying feature not present (lightpush, filter,...).
> > So I think in general we can have /admin by default if having --rest=true.
> > @jm-clius , @fryorcraken : WDYT?
> 
> Sorry, did not realise my IDE did all changes. Let me revert and only push the 2 lines change I actually did.
Yeah I thought that was... :-)
You don't need I think. I found the relevant changes.

fryorcraken:
I have no opinion on the matter as we are yet to understand how node operator would (or would not) use the REST API.
If anything, I'd say default to false to avoid one exposing the admin API by mistake

Ivansete-status (changes requested):
Thanks for submitting this @chaitanyaprem !
Just asking for changes to reduce its scope and avoid the automatic IDE changes.

From my point of view I think is fine to have the `admin` enabled by default. Although I understand this PR is aimed to allow disabling it if needed.

Thanks again, and I'll re-check once the PR is simplified.
Cheers

Ivansete-status (approved):
LGTM! Thanks so much indeed!

gabrielmer (approved):
LGTM, thank you!

Issue #2192, opened 2023-11-07
bug: now receiving messages via auto/messages REST API when subscribed to pubsub topic but not to its content topic


gabrielmer:
### Problem
We run a node subscribed to all 8 pubsub topics:
![image](https://github.com/waku-org/nwaku/assets/101006718/637cfad9-4e44-4734-8236-a18eed994f84)

We publish a message with the following command:
```
curl -X POST "http://127.0.0.1:8645/relay/v1/auto/messages"  -H "content-type: application/json"  -d '{"payload":"ZXhhbXBsZQ==","contentTopic":"/toy-chat/2/luzhou/proto"}'
```

and we then try to retrieve the messages for the given content topic from the node:

```
curl -X GET "http://127.0.0.1:8645/relay/v1/auto/messages/%2Ftoy-chat%2F2%2Fluzhou%2Fproto"  -H "accept: application/json" \
```

We expected to find the message in our node because it is already subscribed to its pubsub topic, but we see the following error in the logs:

```
nwaku_1              | DBG 2023-11-07 16:44:43.671+00:00 Publishing message                         topics="waku node rest relay_api" tid=1 file=handlers.nim:307 contentTopic=/toy-chat/2/luzhou/proto rln=false
nwaku_1              | DBG 2023-11-07 16:44:43.674+00:00 Received response from handler             tid=1 file=serverprivate.nim:253 status=200 meth=POST peer=172.30.0.1:46146 uri=/relay/v1/auto/messages content_type=text/plain content_size=2
nwaku_1              | DBG 2023-11-07 16:44:52.254+00:00 Received request                           tid=1 file=serverprivate.nim:96 peer=172.30.0.1:46160 meth=GET uri=/relay/v1/auto/messages/%2Ftoy-chat%2F2%2Fluzhou%2Fproto
nwaku_1              | DBG 2023-11-07 16:44:52.255+00:00 Serving API request                        tid=1 file=serverprivate.nim:143 peer=172.30.0.1:46160 meth=GET uri=/relay/v1/auto/messages/%2Ftoy-chat%2F2%2Fluzhou%2Fproto path_params="[(contenttopic, /toy-chat/2/luzhou/proto)]" query_params=[] content_body=(None)
nwaku_1              | DBG 2023-11-07 16:44:52.255+00:00 Not subscribed to topic                    topics="waku node rest relay_api" tid=1 file=handlers.nim:253 topic=/toy-chat/2/luzhou/proto
[...]

Issue #2201, opened 2023-11-08
chore: allow fetching cached messages by content or pubsub topic


SionoiS:
### Background
We currently have 2 type of REST API for relay, one for content and pubsub topic. Message cached using one can't be fetched using the other.

### Details
The message cache should allow fetching by content OR pubsub topic. The REST Relay API should also allow fetching message both ways.

### Acceptance criteria
- [x] Updated cache logic
- [x] Updated Relay Rest API

SionoiS:
**Weekly Update**

- _achieved_: failed refactor of message cache 
- _next_: a better and simpler message cache

SionoiS:
**Weekly Update**

- _achieved_: PR merged DONE!

PR #2221, opened 2023-11-15
chore(rest): improved message cache logic

SionoiS:
# Description

I refactored the message cache with the intent to allow fetching messages by pubsub OR content topic. Adding a message now add the missing topic either pubsub or content (still need to be subscribed to either). I also added a check to prevent duplicate messages.

The capacity is now TOTAL instead of PER topic.

# Changes

- [x] new message cache
- [x] updated APIs
- [x] new tests for new behaviour 

Tracking: #2201
Fix: #2192

### DIFF
--- a/waku/waku_api/message_cache.nim
+++ b/waku/waku_api/message_cache.nim
@@ -4,7 +4,7 @@ else:
   {.push raises: [].}
 
 import
-  std/[tables, sequtils],
+  std/[sequtils, sugar, algorithm],
   stew/results,
   chronicles,
   chronos,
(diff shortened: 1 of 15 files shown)


waku/waku_api/message_cache.nim
```diff
@@ -15,68 +15,287 @@ import
...
+
+proc isPubsubSubscribed*(self: MessageCache, pubsubTopic: PubsubTopic): bool =
+  self.pubsubSearch(pubsubTopic).isOk()
+
+proc isContentSubscribed*(self: MessageCache, contentTopic: ContentTopic): bool =
+  self.contentSearch(contentTopic).isOk()
+
+proc pubsubSubscribe*(self: MessageCache, pubsubTopic: PubsubTopic) =
+  let pubsubIdx = self.pubsubSearch(pubsubTopic).valueOr:
+    error
```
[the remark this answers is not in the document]
SionoiS: it's a weird way of getting the value in either the `err` OR the `ok` regardless.

Not sure if there's an equally concise but more legible way to write this....

waku/waku_api/message_cache.nim
```diff
@@ -15,68 +15,287 @@ import
...
+
+  self.pubsubTopics.insert(pubsubTopic, pubsubIdx)
+
+proc contentSubscribe*(self: MessageCache, contentTopic: ContentTopic) =
+  let contentIdx = self.contentSearch(contentTopic).valueOr:
+    error
+
+  self.contentTopics.insert(contentTopic, contentIdx)
```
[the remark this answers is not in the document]
SionoiS: :thinking: That's a mistake `insert` doesn't do what I was thinking. Will change.

waku/waku_api/message_cache.nim
```diff
@@ -15,68 +15,287 @@ import
...
-  if messages.len >= t.capacity.int:
-    trace "Topic cache capacity reached", topic=topic
-    # Message cache on this topic exceeds maximum. Delete oldest.
-    # TODO: this may become a bottle neck if called as the norm rather than
-    #  exception when adding messages. Performance profile needed.
-    messages.delete(0,0)
+  if pubsubIdxRes.isErr():
+    self.pubsubTopics.add(pubsubTopic)
+  else:
+    self.pubsubTopics[pubsubIdx] = pubsubTopic
```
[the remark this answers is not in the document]
SionoiS: True! Will remove.

waku/waku_api/message_cache.nim
```diff
@@ -15,68 +15,287 @@ import
...
-  t.table[topic] = messages
+  # add the message, make space if needed
+  if self.messages.len >= self.capacity:
+    # replace the oldest message
+    self.messages[self.writeIndex] = msg
 
-proc clearMessages*[K](t: MessageCache[K], topic: K) =
-  if not t.isSubscribed(topic):
+    # replace the indexes of the replaced message
+    for (pId, mId) in self.pubsubIndex.mitems:
```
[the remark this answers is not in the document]
SionoiS: mutable iteration!

waku/waku_api/message_cache.nim
```diff
@@ -15,68 +15,287 @@ import
...
+      if mId == self.writeIndex:
+        pId = pubsubIdx
+
+    for (pId, mId) in self.contentIndex.mitems:
```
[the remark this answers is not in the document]
SionoiS: Yes, wrong copy pasta

waku/waku_api/message_cache.nim
```diff
@@ -15,68 +15,286 @@ import
...
+
+    if mId == idx:
+      self.contentIndex.del(j)
+    elif mId == lastIndex:
+      self.contentIndex[j] = (cId, idx)
+
+    dec(j)
+
+  if self.writeIndex > self.messages.high:
+    self.writeIndex = self.messages.high
```
gabrielmer, line 104: If you delete a message mid-array, the indexes of the messages after it get decreased by one and then `mId` in `pubsubIndex` and `contentIndex` can point to different messages than what they initially intended.

Is that taken care of somewhere? or am I missing something?

I feel that the logic should be much simpler and less bug-prone with a different selection of data structures.
I think that using a hash table instead of a seq, and using messageIds as identifiers instead of their indexes should simplify a lot the logic and avoid the problem of moving indexes

Let me know what you think :):)

waku/waku_api/message_cache.nim
```diff
@@ -15,68 +15,286 @@ import
...
+proc isContentSubscribed*(self: MessageCache, contentTopic: ContentTopic): bool =
+  self.contentSearch(contentTopic).isSome()
+
+proc pubsubSubscribe*(self: MessageCache, pubsubTopic: PubsubTopic) =
+  if self.pubsubSearch(pubsubTopic).isNone():
+    self.pubsubTopics.add(pubsubTopic)
+
+proc contentSubscribe*(self: MessageCache, contentTopic: ContentTopic) =
+  if self.contentSearch(contentTopic).isNone():
+    self.contentTopics.insert(contentTopic)
```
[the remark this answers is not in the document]
SionoiS: Opss.

waku/waku_api/message_cache.nim
SionoiS: I agree it's a bit over engineered. :D I just normalized the data and ran with it...... was fun to write though. :P

In this case the `del` is a swap to last index then a decrement of the seq length. O(1) but doesn't preserve ordering.
https://nim-lang.org/docs/system.html#del,seq[T],Natural

gabrielmer: Oh makes sense! Great then :) Thank you!

waku/waku_api/message_cache.nim
```diff
@@ -15,68 +15,292 @@ import
...
+      self.writeIndex = 0
+    
     return
-  t.table[topic] = @[]
+ 
+  let msgIdx = self.messages.len
+  self.messages.add(msg)
+
+  self.pubsubIndex.add((pubsubIdx, msgIdx))
+  self.contentIndex.add((contentIdx, msgIdx))
```
gabrielmer, line 238: Do we need to separate it in two cases?
I think that the code inside the `self.messages.len >= self.capacity` block would work in both cases

waku/waku_api/message_cache.nim
gabrielmer: Now that I think, if ordering is not preserved then when we add a message when the cache is full we may have the unintended consequence of removing a message that is not necessarily the oldest one.

Is that ok?

waku/waku_api/message_cache.nim
```diff
@@ -15,68 +15,292 @@ import
...
+  
+  # update indexes
+  var j = self.pubsubIndex.high
+  while j > -1:
+    let (pId, mId) = self.pubsubIndex[j]
+
+    if mId == idx:
+      self.pubsubIndex.del(j)
+    elif mId == lastIndex:
+      self.pubsubIndex[j] = (pId, idx)
```
gabrielmer, line 94: what are we doing here? not sure I understand this special case of the last index

waku/waku_api/message_cache.nim
SionoiS: :thinking: we could check the message timestamp instead?

waku/waku_api/message_cache.nim
SionoiS: `del` is a swap to last index. So we need to also update `pubsubIndex` to point to the message after the swap.

`messages[last]` swap place with `messages[idx]`

waku/waku_api/message_cache.nim
SionoiS: Is it possible to `self.messages[self.messages.len] = msg` ?

SionoiS:
Converted it back to draft. The first solution was not good.

SionoiS:
This is annoying tests passes locally only...............

waku/waku_api/message_cache.nim
gabrielmer: Oh sorry, just saw your response. My bad then :)

jm-clius:
> I don't want to spend even more time on this

Fair enough 😄

gabrielmer (approved):
LGTM! Thanks so much! 🤩

PR #2225, opened 2023-11-16
chore: creating prepare_release template

gabrielmer:
# Description

Creating GitHub issue template for release work

### DIFF
--- a/.github/ISSUE_TEMPLATE/prepare_release.md
+++ b/.github/ISSUE_TEMPLATE/prepare_release.md
@@ -1,0 +1,27 @@
+---
+name: Prepare release
+about: Execute tasks for the creation and publishing of a new release
+title: 'Prepare release 0.0.0'
+labels: release
+assignees: ''
+
+---
+
+<!--
+Add appropriate release number to title!
+
+For detailed info on the release process refer to https://github.com/waku-org/nwaku/blob/master/docs/contributors/release-process.md
+ -->
+ 
+### Items to complete
+- [ ] create release branch
+- [ ] assign release candidate tag
+- [ ] validate release candidate
+- [ ] generate and edit releases notes
+- [ ] open PR and merge release notes to master
+- [ ] cherry-pick release notes to the release branch
+- [ ] assign release tag to the cherry-picked release notes commit
+- [ ] create GitHub release
+- [ ] deploy the release to DockerHub
+- [ ] deploy release to `wakuv2.prod` fleet
+- [ ] announce the release


.github/ISSUE_TEMPLATE/prepare_release.md
```diff
@@ -0,0 +1,27 @@
+---
+name: Prepare release
+about: Execute tasks for the creation and publishing of a new release
+title: 'Prepare release 0.0.0'
+labels: release
```
gabrielmer, line 5: I don't think this label exists, should we create it? didn't find an existing label that seemed fit

chair28980: @gabrielmer the labels across the waku-org are managed in the labels.yml `pm/.github/labels.yml`, I added the label here: https://github.com/waku-org/pm/commit/d73c241f8785e5572cbd3ac6e6b21368352dd633

chair28980 (approved):
LGTM :green_apple:

.github/ISSUE_TEMPLATE/prepare_release.md
```diff
@@ -0,0 +1,27 @@
...
+- [ ] assign release candidate tag
+- [ ] validate release candidate
+- [ ] generate and edit releases notes
+- [ ] open PR and merge release notes to master
+- [ ] cherry-pick release notes to the release branch
+- [ ] assign release tag to the cherry-picked release notes commit
+- [ ] create GitHub release
+- [ ] deploy the release to DockerHub
+- [ ] deploy release to `wakuv2.prod` fleet
+- [ ] announce the release
```
Ivansete-status, line 27: That's awesome indeed! :partying_face: !
Additionally, I think it would be outstanding if we could have links to pages where we can have more elaborated information re each step.

Ivansete-status (changes requested):
Thanks for that ! Just added a nitpick

.github/ISSUE_TEMPLATE/prepare_release.md
gabrielmer: Yes! I added a comment in the template with the url of the doc with the steps explained in detail

https://github.com/waku-org/nwaku/blob/c65f3e886fca696c94d24883fabf9e8e9c14fcba/.github/ISSUE_TEMPLATE/prepare_release.md?plain=1#L13

How do you think this can be improved?

Ivansete-status: > How do you think this can be improved?

Cool ! I didn't notice that !
IMHO, I think it would be better if we could have a link right after each bullet point, to make it even more explicit

Ivansete-status: Hey @gabrielmer !
Given that we are planning to move all the info into docs.waku.org, I think is better to postpone that detailed link assignation until we have this information there.

PR #2230, opened 2023-11-21
chore: peer_manager.nim - better feedback if can't dial peer with WakuMetadaCodec

Ivansete-status:
Very simple PR to have better insight if the dial against a node supposed to support the "WakuMetadataCodec".


gabrielmer (approved):
Thanks so much!

PR #2234, opened 2023-11-22
chore(store-archive): Remove duplicated code

strinnityk:
# Description
Remove duplicated functions defined in several test files into one. Also improve function naming so as to differentiate between return types with very similar names.

# Changes
- [X] Join duplicated functions
- [X] Improve naming


Ivansete-status (approved):
LGTM! Great refactoring!

Ivansete-status:
ah btw, kindly make sure only the `js-waku-node` test set fails :)

PR #2235, opened 2023-11-22
test(store): Implement store tests (1/2)

AlejandroCabeza:
# Description
Implement store tests

# Changes
- [X] Implement batch of store tests.
- [X] Fix issue with default values for HistoryQuery.

### DIFF
--- a/tests/node/test_wakunode_store.nim
+++ b/tests/node/test_wakunode_store.nim
@@ -277,17 +277,6 @@ suite "Waku Store - End to End - Sorted Archive":
         check:
           queryResponse1.get().messages == archiveMessages[0..<10]
 
-      asyncTest "Pagination with Zero Page Size":
-        # Given the first query (1/1)
-        historyQuery.pageSize = 0
-
-        # When making a history query
-        let queryResponse1 = await client.query(historyQuery, serverRemotePeerInfo)
-
-        # Then the response contains the messages
-        check:
-          queryResponse1.get().messages == archiveMessages[0..<10]
-
       asyncTest "Pagination with Mixed Page Size":
         # Given the first query (1/3)
         historyQuery.pageSize = 2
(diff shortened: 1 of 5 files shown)


tests/node/test_wakunode_store.nim
```diff
@@ -0,0 +1,662 @@
...
+        # When making a history query
+        let queryResponse1 = await client.query(historyQuery, serverRemotePeerInfo)
+
+        # Then the response contains the messages
+        check:
+          queryResponse1.get().messages == archiveMessages[0..<10]
+
+      asyncTest "Pagination with Zero Page Size":
+        # Given the first query (1/1)
+        historyQuery.pageSize = 0
```
gabrielmer, line 281: What does page size 0 mean? Is it something defined?

tests/node/test_wakunode_store.nim
```diff
@@ -0,0 +1,662 @@
...
+          contentTopics: contentTopicSeq,
+          ascending: true
+        )
+
+        # When making a history query
+        let queryResponse = await client.query(historyQuery, serverRemotePeerInfo)
+
+        # Then the response contains the messages
+        check:
+          queryResponse.get().messages == archiveMessages[0..<10]
```
gabrielmer, line 346: Maybe we should change something in the check so if we change the default page size for a number less than 10, that the right number of messages is received and the test passes?

tests/node/test_wakunode_store.nim
Ivansete-status: Very interesting question and test :partying_face: !
@AlejandroCabeza - I think we shouldn't allow pageSize == 0

https://rfc.vac.dev/spec/13/

image

waku/waku_store/rpc.nim
```diff
@@ -156,7 +158,7 @@ proc toAPI*(rpc: HistoryQueryRPC): HistoryQuery =
     pageSize = if rpc.pagingInfo.isNone() or rpc.pagingInfo.get().pageSize.isNone(): 0'u64
                else: rpc.pagingInfo.get().pageSize.get()
 
-    ascending = if rpc.pagingInfo.isNone() or rpc.pagingInfo.get().direction.isNone(): true
+    ascending = if rpc.pagingInfo.isNone() or rpc.pagingInfo.get().direction.isNone(): HistoryQueryAscendingDefaultValue
```
Ivansete-status, line 161: Would it be possible to have the changes in `rpc.nim` in a separate PR? I think that will give better scope to the change. Even though this seems a simple change, I think is convenient to separate it.

On the other hand, I'd change the attribute's name from `ascending` to `direction`.

Ivansete-status (changes requested):
LGTM! Thanks for it!
Just added a couple of comments. The most relevant IMHO is to put the `rpc.nim` change in a separate PR.
Cheers

tests/waku_store/test_client.nim
```diff
@@ -0,0 +1,209 @@
...
+
+    serverPeerInfo = serverSwitch.peerInfo.toRemotePeerInfo()
+    clientPeerInfo = clientSwitch.peerInfo.toRemotePeerInfo()
+
+  asyncTeardown:
+    await allFutures(serverSwitch.stop(), clientSwitch.stop())
+  
+  suite "HistoryQuery Creation and Execution":
+    asyncTest "Valid Queries":
+      assert false
```
gabrielmer, line 77: What does this assert false do? doesn't it fail the test?

tests/node/test_wakunode_store.nim
AlejandroCabeza: I was considering that to be equivalent to `None`, a.k.a. "give me everything".
Should we ask some RFC guy?

tests/node/test_wakunode_store.nim
AlejandroCabeza: Sure, fine by me.

tests/waku_store/test_client.nim
AlejandroCabeza: That's funny, I can't recall how that ended up there 😂 Might be some piece of code left from testing one bug out. Good catch!

tests/node/test_wakunode_store.nim
gabrielmer: Yes! Now that I read again I didn't express well the idea in the comment.
What I mean is that if for some reason we decide to change the default page size, this test can fail because of the hardcoded 10.
Not sure if it's the intention or not

gabrielmer (approved):
Overall looks great! Thanks so much!

tests/node/test_wakunode_store.nim
AlejandroCabeza: Where do you mean "hardcoded 10"?

gabrielmer: The 10 in `queryResponse.get().messages == archiveMessages[0..<10]`

If for some reason we decide to change in our codebase the default page size for a number less than 10, then this will fail. This test assumes that the default page size is >= 10 and I think we shouldn't make any assumptions about the default size.

waku/waku_store/rpc.nim
AlejandroCabeza: Moved to https://github.com/waku-org/nwaku/pull/2263.

tests/node/test_wakunode_store.nim
jm-clius: Mmm. I definitely think we'd want to be explicit here in future versions of the RFC, but in protobuf if you leave an int field unset it would default to `0`. In other words, if someone leaves the `pageSize` unset it would be interpreted as page size `0` and nwaku would just use the default page size: `DefaultPageSize` - this seems reasonable to me.

tests/node/test_wakunode_store.nim
AlejandroCabeza: Scratch all I wrote above, I was assuming the `Default Page Size` was equivalent to `0`. Hanno just answered another thread and I realised what you truly meant. Updating.

Ivansete-status (approved):
LGTM! Thanks for it!

PR #2236, opened 2023-11-22
fix: waku_filter_v2/common: PEER_DIAL_FAILURE ret code change: 200 -> 504

Ivansete-status:
Changing the return code to a more meaningful value when `PEER_DIAL_FAILURE` happens.

### DIFF
--- a/waku/waku_filter_v2/common.nim
+++ b/waku/waku_filter_v2/common.nim
@@ -13,7 +13,7 @@ const
 type
   FilterSubscribeErrorKind* {.pure.} = enum
     UNKNOWN = uint32(000)
-    PEER_DIAL_FAILURE = uint32(200) # TODO shouldn't this be an error code, e.g. 504 Gateway Timeout?
+    PEER_DIAL_FAILURE = uint32(504)
     BAD_RESPONSE = uint32(300)
     BAD_REQUEST = uint32(400)
     NOT_FOUND = uint32(404)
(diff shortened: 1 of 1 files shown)


jm-clius (approved):
LGTM and this should definitely be an error code.

Note however that it was chosen to correspond with what is already implemented for Store: https://github.com/waku-org/nwaku/blob/59ee3c69c8fb1733bda2875d1c2b771e2b87b02d/waku/waku_store/common.nim#L66 and would have to be fixed there as well.

Less important, but something to consider:
504 is also the HTTP error code for Gateway Timeout, which is very similar/equivalent to peer dial failure. Should we consider sticking to HTTP terminology? All other error codes are chosen to mirror the HTTP codes

waku/waku_filter_v2/common.nim
```diff
@@ -13,11 +13,11 @@ const
 type
   FilterSubscribeErrorKind* {.pure.} = enum
     UNKNOWN = uint32(000)
-    PEER_DIAL_FAILURE = uint32(200) # TODO shouldn't this be an error code, e.g. 504 Gateway Timeout?
```
SionoiS, line 16: There are a couples of TODOs in the code base and I always think to myself. Why not just change it instead of adding a TODO your touching this part of the code anyway. If you need to remember to do something later should open an issue IMO.

SionoiS (approved):
Much better thanks!

PR #2260, opened 2023-11-30
chore: waku_store/common.nim: correct ret code in PEER_DIAL_FAILURE

Ivansete-status:
## Description
Simple PR to have proper return error codes in Store protocol.
This is inspired by the following comment: https://github.com/waku-org/nwaku/pull/2236#pullrequestreview-1744220535


gabrielmer (approved):
LGTM, thank you!

SionoiS (approved):
Thx!

PR #2263, opened 2023-11-30
refactor(store): HistoryQuery.direction

AlejandroCabeza:
# Description
Update `HistoryQuery.ascending` (and related) references to  `direction` so they are consistent with RFC naming. Also, fix a default value issue with it.

# Changes
- [X] Update `ascending` references to `direction`
- [X] Fix default value issue with `HistoryQuery.direction`

### DIFF
--- a/tests/waku_archive/test_waku_archive.nim
+++ b/tests/waku_archive/test_waku_archive.nim
@@ -341,7 +341,7 @@ procSuite "Waku Archive - find messages":
     ## Given
     let req = ArchiveQuery(
       pageSize: 4,
-      ascending: true
+      direction: true
     )
 
     ## When
@@ -377,7 +377,7 @@ procSuite "Waku Archive - find messages":
     ## Given
     let req = ArchiveQuery(
       pageSize: 4,
-      ascending: false  # backward
+      direction: false  # backward
     )
 
     ## When
@@ -454,7 +454,7 @@ procSuite "Waku Archive - find messages":
       contentTopics: @[ContentTopic("1")],
       startTime: some(ts(15, timeOrigin)),
       endTime: some(ts(55, timeOrigin)),
-      ascending: true
+      direction: true
     )
 
     ## When
--- a/tests/waku_archive/test_waku_archive.nim
+++ b/tests/waku_archive/test_waku_archive.nim
@@ -9,6 +9,7 @@ import
 
 import
   ../../../waku/common/databases/db_sqlite,
+  ../../../waku/common/paging,
   ../../../waku/waku_core,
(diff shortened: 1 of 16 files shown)


Ivansete-status (changes requested):
Thanks for this PR!
It looks great! Besides, I think it would be a bit better if we could use the following type when we refer to a "pagination direction", if that's feasible of course

https://github.com/waku-org/nwaku/blob/103d3981ad80b83dba48fe88cb5f2dcf715104b2/waku/waku_store/rpc.nim#L44

tests/waku_archive/test_waku_archive.nim
```diff
@@ -341,7 +341,7 @@ procSuite "Waku Archive - find messages":
     ## Given
     let req = ArchiveQuery(
       pageSize: 4,
-      ascending: true
+      direction: true
```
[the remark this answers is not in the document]
AlejandroCabeza: I completely agree, but same reason as before: I didn't wanna change some code you guys might be used to. But, given your comment, I'll understand this as an approval for the change 😁

waku/waku_api/jsonrpc/store/handlers.nim
```diff
@@ -27,16 +28,16 @@ proc toPagingInfo*(pagingOptions: StorePagingOptions): PagingInfoRPC =
   PagingInfoRPC(
     pageSize: some(pagingOptions.pageSize),
     cursor: pagingOptions.cursor,
-    direction: if pagingOptions.forward: some(PagingDirectionRPC.FORWARD)
-               else: some(PagingDirectionRPC.BACKWARD)
+    direction: if pagingOptions.forward: some(PagingDirection.FORWARD)
+               else: some(PagingDirection.BACKWARD)
```
Ivansete-status, line 32: Good morning!
Is this conversion needed?

Ivansete-status (approved):
LGTM! Thanks! Just added a minor comment

waku/waku_api/jsonrpc/store/handlers.nim
AlejandroCabeza: I don't think so, I likely missed it because of tunnel vision when updating stuff. Good catch

AlejandroCabeza: Nevermind, that was actually well done. `pageOptions.forward` is a bool 😂

NagyZoltanPeter (approved):
It is great, thank you. I realized how rest and rpc namings are differ but its not an issue to change. We will remove sometime rpc thus we should not break any interface just because of it.

PR #2267, opened 2023-12-01
fix(discv5): add bootnode filter exception

### LINKED ISSUE #2162: bug: can't discover peers when using DNS Discovery URL + shards
The nodes returned by DNS DIscovery in `shards.test` don't have the information about the shards: 
```
enrtree://AMOJVZX4V6EXP7NTJPMAYJYST2QP6AJXYW76IU6VGJS7UVSNDYZG4@boot.test.shards.nodes.status.im

Nodes returned by this URL
1 - enr:-Ne4QJKpiQqwYpo0p1yDW6opKFYzh801nhSzX65S_x892UXABVYzFBrdFwCPiWwXlKqVz5sXkTzYtUuX1wg2sW5DZnwBgmlkgnY0gmlwhCIfDu-KbXVsdGlhZGRyc7g4ADY2MWJvb3QtMDIuZ2MtdXMtY2VudHJhbDEtYS5zaGFyZHMudGVzdC5zdGF0dXNpbS5uZXQGdl-Jc2VjcDI1NmsxoQJm8YcPIYhI5rvlLJJRlpebApk6w4uOLdFgAeHN2wO9N4N0Y3CCdl-DdWRwgiMohXdha3UyDQ
2 - enr:-Ne4QIvHiMe1Gf7h22jygL1kPFVAcQ0RkDYNk1PNA52KUKElBSPuPy-HSD1pRX-rCx2A2Qqh0GtkzFUyL8NQEiL15P0BgmlkgnY0gmlwhAjaF0yKbXVsdGlhZGRyc7g4ADY2MWJvb3QtMDEuYWMtY24taG9uZ2tvbmctYy5zaGFyZHMudGVzdC5zdGF0dXNpbS5uZXQGdl-Jc2VjcDI1NmsxoQM_sJtGT5gonA4UUzhn2d7LQY9ztY8loLAaSk1HKVruYIN0Y3CCdl-DdWRwgiMohXdha3UyDQ
3 - enr:-Ne4QHOpWLyVVZMzJwXcc00CNp16vB5x2WFy6WQAEKyaOf_UMWKvz2a0HN9QCoSyBYmudBKspqYa_U6tJ64B0TqLzy0BgmlkgnY0gmlwhAjarmyKbXVsdGlhZGRyc7g4ADY2MWJvb3QtMDIuYWMtY24taG9uZ2tvbmctYy5zaGFyZHMudGVzdC5zdGF0dXNpbS5uZXQGdl-Jc2VjcDI1NmsxoQNeQXcyqdYwEjflVdLKYAusuZJ93fpGiFwqK1jU9ISQC4N0Y3CCdl-DdWRwgiMohXdha3UyDQ
4 - enr:-M24QJDZfhB_wN_PHOAQuzgnta20xKUsZl5kdhBeQJM16gdldCJNAKQp6dgbwo-MTRJxYVNCr85cHRAJxtNLR4vTbP0BgmlkgnY0gmlwhKdjEy-KbXVsdGlhZGRyc68ALTYoYm9vdC0wMS5kby1hbXMzLnNoYXJkcy50ZXN0LnN0YXR1c2ltLm5ldAZ2X4lzZWNwMjU2azGhAt60bRUEoHNuLlnsM12sU2PIQwBwfLIJ8a_ZPEY2-Rnkg3RjcIJ2X4N1ZHCCIyiFd2FrdTIN
5 - enr:-M24QAsRRxoLDnnXFGnbHGUKjtqgXOVxb2Cian1vegc1rtY0Yk5wXDF7NeBzPl7frvyxo3Vt-xSL0vUa2jazchNIS_oBgmlkgnY0gmlwhLKAj_GKbXVsdGlhZGRyc68ALTYoYm9vdC0wMi5kby1hbXMzLnNoYXJkcy50ZXN0LnN0YXR1c2ltLm5ldAZ2X4lzZWNwMjU2azGhAtsXOrELG9R5LlIbF6bqeLC0tg7bmNzQ0JkSmEO3zxqzg3RjcIJ2X4N1ZHCCIyiFd2FrdTIN
6 - enr:-Ne4QINS7SZiUk9oN3mcLpOrdQrFWS-AUDjyq5F9__8iTUT_H8ExnAj5qDWmG4qbLaz4NKvDtmIU3Ycu9sP_Ixk6hn4BgmlkgnY0gmlwhCKHDVeKbXVsdGlhZGRyc7g4ADY2MWJvb3QtMDEuZ2MtdXMtY2VudHJhbDEtYS5zaGFyZHMudGVzdC5zdGF0dXNpbS5uZXQGdl-Jc2VjcDI1NmsxoQLGOqANDRbJFI6KVhTfYMDmT9c2UOKzebVV1eQr3EzqQ4N0Y3CCdl-DdWRwgiMohXdha3UyDQ
```

These nodes are currently subscribed to shards:
- `/waku/2/rs/16/32`
- `/waku/2/rs/16/64`
- `/waku/2/rs/16/128`
- `/waku/2/rs/16/256`

However if you check the ENRs in https://enr-viewer.com/ these nodes lack the `rs` or `rsv` attributes. Also, if you compare the current ENR of the nodes (using the nodes' RPC server), against those from the DNS Discovery URL, you'd see that the latter are outdated compared to the former, which makes sense since the DNS Discovery URL was manually created when the fleet was setup, while the subscription to shards is something that happens 'dynamically'.

This is problematic because If i have a node subscribed to shards 32,64,128 and/or 256, if I use the dns discovery URL, I'll not be able to find new peers, because they'll get filtered out. DiscV5 will populate the routing tables with those shardless ENRs retrieved from the URL, and then they get filtered out, since Discv5 (at least in go-waku implementation, and looking at nwaku implementation seems to be similar) will not ask those nodes for their current ENR, and the filtering logic defined in this predicate: https://github.com/waku-org/nwaku/blob/3be6163639d4141a6ee51ae8f8e83635f541f783/waku/waku_discv5.nim#L53-L75 , meaning that the nodes will not be able to discover new peers using the dns discovery URL.

This is particularly problematic in the case of Status, since the DNS discovery URL is hardcoded in the node configuration.

SionoiS:
# Description
Bootstrapping nodes should not be filtered out.

# Changes
- [x] Removed bootnode filtering

Closes #2162

### DIFF
--- a/waku/waku_discv5.nim
+++ b/waku/waku_discv5.nim
@@ -52,7 +52,7 @@ type WakuDiscoveryV5* = ref object
     peerManager: Option[PeerManager]
     topicSubscriptionQueue: AsyncEventQueue[SubscriptionEvent]
 
-proc shardingPredicate*(record: Record): Option[WakuDiscv5Predicate] =
+proc shardingPredicate*(record: Record, bootnodes: seq[Record] = @[]): Option[WakuDiscv5Predicate] =
   ## Filter peers based on relay sharding information
   let typedRecord = record.toTyped().valueOr:
     debug "peer filtering failed", reason=error
@@ -65,8 +65,9 @@ proc shardingPredicate*(record: Record): Option[WakuDiscv5Predicate] =
   debug "peer filtering updated"
 
   let predicate = proc(record: waku_enr.Record): bool =
-      record.getCapabilities().len > 0 and #RFC 31 requirement
-      nodeShard.shardIds.anyIt(record.containsShard(nodeShard.clusterId, it)) #RFC 64 guideline
+      bootnodes.contains(record) or # Temp. Bootnode exception
+      (record.getCapabilities().len > 0 and #RFC 31 requirement
+      nodeShard.shardIds.anyIt(record.containsShard(nodeShard.clusterId, it))) #RFC 64 guideline
 
   return some(predicate)
 
@@ -78,32 +79,23 @@ proc new*(
   peerManager: Option[PeerManager] = none(PeerManager),
   queue: AsyncEventQueue[SubscriptionEvent] = newAsyncEventQueue[SubscriptionEvent](30),
   ): T =
-  let shardPredOp =
-    if record.isSome(): shardingPredicate(record.get())
-    else: none(WakuDiscv5Predicate)
-  
-  var bootstrapRecords = conf.bootstrapRecords
-
-  # Remove bootstrap nodes with which we don't share shards.
-  if shardPredOp.isSome():
-    bootstrapRecords.keepIf(shardPredOp.get())
-  
-  if conf.bootstrapRecords.len > 0 and bootstrapRecords.len == 0:
-    warn "No discv5 bootstrap nodes share this node configured shards"
-
   let protocol = newProtocol(
     rng = rng,
     config = conf.discv5Config.get(protocol.defaultDiscoveryConfig),
     bindPort = conf.port,
     bindIp = conf.address,
     privKey = conf.privateKey,
-    bootstrapRecords = bootstrapRecords,
+    bootstrapRecords = conf.bootstrapRecords,
     enrAutoUpdate = conf.autoupdateRecord,
     previousRecord = record,
     enrIp = none(IpAddress),
     enrTcpPort = none(Port),
     enrUdpPort = none(Port),
   )
+
+  let shardPredOp =
+    if record.isSome(): shardingPredicate(record.get(), conf.bootstrapRecords)
+    else: none(WakuDiscv5Predicate)
 
   WakuDiscoveryV5(
     conf: conf,
@@ -239,7 +231,7 @@ proc subscriptionsListener(wd: WakuDiscoveryV5) {.async.} =
 
     debug "ENR updated successfully"
 
-    wd.predicate = shardingPredicate(wd.protocol.localNode.record)
+    wd.predicate = shardingPredicate(wd.protocol.localNode.record, wd.protocol.bootstrapRecords)
 
   wd.topicSubscriptionQueue.unregister(key)
 


jm-clius (approved):
Seems reasonable.

waku/waku_discv5.nim
```diff
@@ -78,33 +79,24 @@ proc new*(
...
     enrAutoUpdate = conf.autoupdateRecord,
     previousRecord = record,
     enrIp = none(ValidIpAddress),
     enrTcpPort = none(Port),
     enrUdpPort = none(Port),
   )
 
+  let shardPredOp =
+    if record.isSome(): shardingPredicate(record.get(), conf.bootstrapRecords)
+    else: none(WakuDiscv5Predicate)
```
Ivansete-status, line 98: Not a big deal at all but maybe we can simplify a little by setting a default argument to the `new` proc:
Something like:
```
record: Option[waku_enr.Record] = none(WakuDiscv5Predicate)
```

waku/waku_discv5.nim
```diff
@@ -52,7 +52,7 @@ type WakuDiscoveryV5* = ref object
     peerManager: Option[PeerManager]
     topicSubscriptionQueue: AsyncEventQueue[SubscriptionEvent]
 
-proc shardingPredicate*(record: Record): Option[WakuDiscv5Predicate] =
+proc shardingPredicate*(record: Record, bootnodes: seq[Record] = @[]): Option[WakuDiscv5Predicate] =
```
Ivansete-status, line 55: This comment is out of the scope of the PR but, shall we make it private?
```suggestion
proc shardingPredicate(record: Record, bootnodes: seq[Record] = @[]): Option[WakuDiscv5Predicate] =
```

Ivansete-status (approved):
LGTM! Thanks for it!
However, let's better wait for the @richard-ramos 's opinion before merging :)
I also added minor nitpick comments

richard-ramos (approved):

waku/waku_discv5.nim
SionoiS: The types would be different no?

waku/waku_discv5.nim
SionoiS: it's required for testing

PR #2276, opened 2023-12-08
chore: Cbindings allow mounting the Store protocol from libwaku

Ivansete-status:
# Description
Allows to mount _Store_ in the node started from `libwaku`. With that PR, the node can receive the following configuration items as a parameter:

```
{
      "store": 1,
      "storeDbUrl": "postgres://postgres:test123@localhost:5432/postgres",
      "storeRetentionPolicy": "time:6000000",
      "storeMaxNumDbConnections": 30
}
```

In upcoming PRs, we will introduce pending changes to allow the "self" node to make requests to other _Store_ nodes, and we will introduce an additional API function for that: `waku_store_query(char* queryJSON, char* peerID, int ms, WakuCallBack cb, void* userData);`

Note: ignore the library/waku_thread/inter_thread_communication/requests/protocols/store_request.nim for now. It will be completed shortly.

# Changes

- [x] Allow mount Store protocol from libwaku-cbindings 
- [x] Extend the C example to mount _Store_ in the "self" node.
- [x] Move the `WakuCallBack` definition in a separate file, `library/callback.nim`

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -4,4 +4,6 @@
 /LICENSE*
 /tests
 /metrics
+/nimcache
+librln*
 **/vendor/*
--- a/.github/ISSUE_TEMPLATE/prepare_release.md
+++ b/.github/ISSUE_TEMPLATE/prepare_release.md
@@ -1,0 +1,27 @@
+---
+name: Prepare release
+about: Execute tasks for the creation and publishing of a new release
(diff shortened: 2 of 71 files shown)


jm-clius (approved):
Thanks! :)

SionoiS (approved):

PR #2280, opened 2023-12-11
chore: peer_manager.nim - add protection to avoid crashes if wakuMetadata is nil

Ivansete-status:
Simple PR to avoid crashes if `wakuMetadata` is not properly set

### DIFF
--- a/waku/node/waku_node.nim
+++ b/waku/node/waku_node.nim
@@ -369,7 +369,12 @@ proc startRelay*(node: WakuNode) {.async.} =
 
 proc mountRelay*(node: WakuNode,
                  pubsubTopics: seq[string] = @[],
-                 peerExchangeHandler = none(RoutingRecordsHandler)) {.async, gcsafe.} =
+                 peerExchangeHandler = none(RoutingRecordsHandler),
+                 metadataClusterId: uint32 = 1) {.async, gcsafe.} =
+
+  ## metadataClusterId: Param needed by the WakuMetadata protocol.
+  ##                    Defaults to 1 to indicate The Waku Network, aka TWN.
+
   if not node.wakuRelay.isNil():
     error "wakuRelay already mounted, skipping"
     return
@@ -399,6 +404,13 @@ proc mountRelay*(node: WakuNode,
   # Subscribe to topics
   for pubsubTopic in pubsubTopics:
     node.subscribe((kind: PubsubSub, topic: pubsubTopic))
+
+  ## Mount the WakuMetadata protocol
+  ## Notice that we mount the WakuMetadata at this point because there couldn't exist
+  ## WakuRelay without WakuMetadata.
+  node.mountMetadata(metadataClusterId).isOkOr:
+    error "failed to mount waku metadata protocol", error=error
+    return
 
 ## Waku filter
 
--- a/waku/node/waku_node.nim
+++ b/waku/node/waku_node.nim
@@ -369,11 +369,7 @@ proc startRelay*(node: WakuNode) {.async.} =
 
 proc mountRelay*(node: WakuNode,
                  pubsubTopics: seq[string] = @[],
-                 peerExchangeHandler = none(RoutingRecordsHandler),
-                 metadataClusterId: uint32 = 1) {.async, gcsafe.} =
-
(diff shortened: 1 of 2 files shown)


alrevuelta (changes requested):
not sure i understand the implications. If someone doesn't mount it, this function will always return and no relay connections will be attempted. That doesn't seem desirable.

Wondering also if we should support clients not supporting this protocol, meaning having `nil` here is a bug or feature?

jm-clius:
> not sure i understand the implications. If someone doesn't mount it, this function will always return and no relay connections will be attempted. That doesn't seem desirable.

As far as I can see, though, this entire function requires a mounted `wakuMetadata` as main source of truth for supported shards. This nil check doesn't change this strict requirement, but perhaps this approach should be rethought - why not have the shards be a node property?

> Wondering also if we should support clients not supporting this protocol, meaning having `nil` here is a bug or feature?

@Ivansete-status @SionoiS are there legitimate scenarios where someone could mount Relay without mounting WakuMetadata?

Ivansete-status:
Thanks for the comments!
@alrevuelta - what change do you want me to apply?:)
@jm-clius - IMO, the `wakuMetadata` is related only to "static sharding", and therefore I assume we should support `wakuMetadata == nil` when the node uses "named sharding" or "auto sharding" (@SionoiS please correct me if I'm wrong.)

alrevuelta:
> UPDATE: in other words, if we believe that WakuMetadata is essential for Relay, we shouldn't really be able to get to this point with wakuMetadata set to nil.

I agree with this. and as far as i can see, its always mounted: https://github.com/waku-org/nwaku/blob/15c1f974df17bdd686704286a6a5f035a2391fb1/apps/wakunode2/app.nim#L410-L411

so perhaps this check is not necessary here?

we can add it in `new` https://github.com/waku-org/nwaku/blob/a02498d40346df578d1369598597647486d23c5b/waku/node/peer_manager/peer_manager.nim#L413 and error/raise if nil.

SionoiS:
> Thanks for the comments! @alrevuelta - what change do you want me to apply?:) @jm-clius - IMO, the `wakuMetadata` is related only to "static sharding", and therefore I assume we should support `wakuMetadata == nil` when the node uses "named sharding" or "auto sharding" (@SionoiS please correct me if I'm wrong.)

Named sharding is deprecated no? As for autosharding yes Waku Metadata is required.

Ivansete-status:
I tried to apply your suggestion @alrevuelta and mount the `WakuMetadata` within `proc new*(T: type PeerManager,...` but I considered that would require too many changes. Therefore, I applied a different approach, considering the premise that the WakuMetadata protocol will be always mounted when WakuRelay is mounted. Let me know if that looks good ;P

waku/node/waku_node.nim
```diff
@@ -401,6 +406,13 @@ proc mountRelay*(node: WakuNode,
   for pubsubTopic in pubsubTopics:
     node.subscribe((kind: PubsubSub, topic: pubsubTopic))
 
+  ## Mount the WakuMetadata protocol
+  ## Notice that we mount the WakuMetadata at this point because there couldn't exist
+  ## WakuRelay without WakuMetadata.
+  node.mountMetadata(metadataClusterId).isOkOr:
```
alrevuelta, line 412: not sure I agree with this. A node can have relay disabled but still mount metadata protocol (eg light client).

My suggestion would be to:
* leave it as it was
* just raise an exception in the peer manager if wakumetadata is nil.

Ivansete-status: Okay perfect!
The thing is that  I need the `WakuMetadata` to be mounted somehow so that I don't have an exception thrown from c-bindings. IMO, that shouldn't throw an exception because of trying to access a `nil` ref object.

I'll re-think the approach then!

alrevuelta: mm i mean something like this:
https://github.com/waku-org/nwaku/blob/master/waku/node/peer_manager/peer_manager.nim#L429

Ivansete-status:
This PR doesn't make sense after applying this revert: https://github.com/waku-org/nwaku/pull/2312

We will likely need to resume this again once the `feat: shard aware peer management (#2151)` feature is resumed too.

PR #2281, opened 2023-12-11
chore: examples/nodejs - adapt code to latest callback and ctx/userData definitions

Ivansete-status:
NodeJs example adaptation to the latest cbindings approach, where we use a Context and userData.

### DIFF
--- a/examples/nodejs/waku.js
+++ b/examples/nodejs/waku.js
@@ -27,15 +27,18 @@ function event_handler(event) {
     console.log("evento NodeJs: " + event)
 }
 
+wakuMod.wakuNew(cfg)
+
 wakuMod.wakuVersion(function(msg){ console.log("Waku Version: " + msg) })
 
 // Example on how to retrieve a value from the waku library
 var defaultPubsubTopic = ""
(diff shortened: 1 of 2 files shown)


NagyZoltanPeter (approved):
Seems all good!
Only remark that we discussed offline. 
I see no exact reason for double dereference ctx on the subsequent api calls in C after creation.
I saw that libwaku.nim api procs are waiting for ptr ptr Context while libwaku.h just has the signature void* ctx. This might not be obvious for the lib user from C side to pass a void** ctx. Maybe there is room for future improvement here.
I would consider to use an artificial handle (unique id) as context descriptor not a raw pointer. Such a handle can be checked on the library side to avoid missuse of pointers.

gabrielmer:
I think that the number of arguments to call a function do not match with the definitions in https://github.com/waku-org/nwaku/blob/c9d558f878f89319ddf397304a475ddb480305cf/library/libwaku.nim for the functions `waku_new` and `waku_set_event_callback`

Otherwise LGTM :)

gabrielmer (approved):

PR #2283, opened 2023-12-12
chore: archive - move error to trace level when insert row fails

Ivansete-status:
## Description

That is helpful to prevent the node to spam the logs when it shares connection to the same Postgres database with other nodes, in which case the following log appears too much:

topics="waku archive" tid=1 file=archive.nim:113 err="error in runStmt: error in dbConnQueryPrepared calling waitQueryToFinish: error in query: ERROR: duplicate key value violates unique constraint "messageindex" DETAIL: Key
(messagehash)=(88f4ee115eef6f233a7dceaf975f03946e18666adda877e38d61be98add934e8) already exists. "

### DIFF
--- a/waku/waku_archive/archive.nim
+++ b/waku/waku_archive/archive.nim
@@ -111,7 +111,7 @@ proc handleMessage*(w: WakuArchive,
 
     let putRes = await w.driver.put(pubsubTopic, msg, msgDigest, msgHash, msgReceivedTime)
     if putRes.isErr():
-      error "failed to insert message", err=putRes.error
+      trace "failed to insert message", err=putRes.error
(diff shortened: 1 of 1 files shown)


richard-ramos (approved):

NagyZoltanPeter (approved):
The code change is ok as is, but my first think are we sure we are just hiding such dupl instert issues only? Cannot we distinguish somehow between such failures?

jm-clius (approved):

Ivansete-status:
> The code change is ok as is, but my first think are we sure we are just hiding such dupl instert issues only? Cannot we distinguish somehow between such failures?

I've applied this. Now only change the log level for "duplicate insert" errors and leave the rest of errors in `debug` level.

PR #2292, opened 2023-12-14
chore(rln_db_inspector): include in wakunode2 binary

rymnc:
# Description

This PR includes the rln-db-inspector tool within the wakunode2 binary for ease of use.

# Changes

- [x] Removes rln_db_inspector `external_config.nim`, and `nim.cfg`
- [x] `proc`ifies the rln_db_inspector for use with wakunode2
- [x] adds `inspectRlnDb` command to wakunode2

## How to test

1. First, we compile the binary
    
    ```bash
    make -j16 wakunode2
    ```
    This command will fetch the rln static library and link it automatically.

2. Define the arguments you wish to use

    ```bash
    export RLN_TREE_DB_PATH="xxx"
    ```

3. Run the db inspector 

    ```bash
    ./build/wakunode2 inspectRlnDb \
    --rln-relay-tree-path:$RLN_TREE_DB_PATH 
    ```

    What this does is - 
    a. loads the tree db from the path provided
    b. Logs out the metadata, including, number of leaves set, past 5 merkle roots, last synced block number


apps/wakunode2/external_config.nim
```diff
@@ -30,9 +30,12 @@ type ProtectedTopic* = object
   topic*: string
   key*: secp256k1.SkPublicKey
 
+proc noop() = discard
```
alrevuelta, line 33: ?

alrevuelta (approved):
minor comment but lgtm

apps/wakunode2/external_config.nim
rymnc: still a wip, need to figure out how we can have a command for wakunode2 which doesn't accept additional args, see my message here - https://discord.com/channels/1110799176264056863/1151319265694199829/1184765875329433681

Ivansete-status:
While I understand this approach is followed by famous apps/projects such as _Geth_, _Docker_, _Git_, etc., I think for now it is better to have the tools separated.

It sounds a bit weird to me having to use it with something like: `./build/wakunode2 inspectRlnDb --rln-relay-tree-path:$RLN_TREE_DB_PATH`

imo, I think it would be better to have the following subcommand set:

`./build/waku node ...`
`./build/waku rln ...`
`./build/waku canary ...` ***
`./build/waku monitor ...` ***

*** Notice that I've added a line for to the current [tools](https://github.com/waku-org/nwaku/releases/download/v0.23.0-rc.1/nwaku-tools-amd64-linux-v0.23.0-rc.1.tar.gz) that we are packaging (wakucanary and networkmonitor.) 

My suggestion is to postpone that `rln`-app unification for now and leave it for a future global unification where we will have the above commands. If not possible to postpone, I suggest following that direction so that we achieve that command-calling-convention in the future.

wdyt @alrevuelta @rymnc @jm-clius @NagyZoltanPeter @gabrielmer @SionoiS?

Also, very important to align this approach with `go-waku` colleagues (cc @richard-ramos @chaitanyaprem .)

alrevuelta:
@Ivansete-status Tools such as `monitor` seems very specific.

PR #2294, opened 2023-12-14
fix: make rln rate limit spec compliant

alrevuelta:
fix: make rln rate limit spec compliant
closes https://github.com/waku-org/nwaku/issues/2289

### DIFF
--- a/tests/waku_rln_relay/test_wakunode_rln_relay.nim
+++ b/tests/waku_rln_relay/test_wakunode_rln_relay.nim
@@ -398,7 +398,7 @@ procSuite "WakuNode - RLN relay":
     await node3.stop()
 
   asyncTest "clearNullifierLog: should clear epochs > MaxEpochGap":
-    
+
     let
       # publisher node
       nodeKey1 = generateSecp256k1Key()
@@ -445,7 +445,7 @@ procSuite "WakuNode - RLN relay":
       proofAdded1 = node1.wakuRlnRelay.appendRLNProof(wm1, time)
       # another message in the same epoch as wm1, it will break the messaging rate limit
       wm2 = WakuMessage(payload: "message 2".toBytes(), contentTopic: contentTopic)
-      proofAdded2 = node1.wakuRlnRelay.appendRLNProof(wm2, time + EpochUnitSeconds)
+      proofAdded2 = node1.wakuRlnRelay.appendRLNProof(wm2, time)
       #  wm3 points to the next epoch
       wm3 = WakuMessage(payload: "message 3".toBytes(), contentTopic: contentTopic)
       proofAdded3 = node1.wakuRlnRelay.appendRLNProof(wm3, time + EpochUnitSeconds * 2)
@@ -455,7 +455,7 @@ procSuite "WakuNode - RLN relay":
       proofAdded1
       proofAdded2
       proofAdded3
-    
+
     #  relay handler for node2
     var completionFut1 = newFuture[bool]()
     var completionFut2 = newFuture[bool]()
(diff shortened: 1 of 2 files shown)


fryorcraken (approved):

jm-clius (approved):

tests/waku_rln_relay/test_wakunode_rln_relay.nim
```diff
@@ -445,7 +445,7 @@ procSuite "WakuNode - RLN relay":
       proofAdded1 = node1.wakuRlnRelay.appendRLNProof(wm1, time)
       # another message in the same epoch as wm1, it will break the messaging rate limit
       wm2 = WakuMessage(payload: "message 2".toBytes(), contentTopic: contentTopic)
-      proofAdded2 = node1.wakuRlnRelay.appendRLNProof(wm2, time + EpochUnitSeconds)
+      proofAdded2 = node1.wakuRlnRelay.appendRLNProof(wm2, time)
```
alrevuelta, line 448: @rymnc Unsure I understand why `wm2` had `time + EpochUnitSecods` if its expected to "break the messaging rate limit". Isn't it better to set the same time?

tests/waku_rln_relay/test_wakunode_rln_relay.nim
```diff
@@ -469,25 +469,25 @@ procSuite "WakuNode - RLN relay":
...
           completionFut3.complete(true)
-    
+
     # mount the relay handler for node2
     node2.subscribe((kind: PubsubSub, topic: DefaultPubsubTopic), some(relayHandler))
     await sleepAsync(2000.millis)
 
     await node1.publish(some(DefaultPubsubTopic), wm1)
-    await sleepAsync(10.seconds)
     await node1.publish(some(DefaultPubsubTopic), wm2)
```
alrevuelta, line 478: @rymnc Mind clarifying why this sleep was needed? Epoch is hardcoded above, so perhaps not needed?

tests/waku_rln_relay/test_wakunode_rln_relay.nim
```diff
@@ -469,25 +469,25 @@ procSuite "WakuNode - RLN relay":
...
       res2 = await completionFut2.withTimeout(10.seconds)
       res3 = await completionFut3.withTimeout(10.seconds)
 
     check:
-      (res1 and res2 and res3) == true # all 3 are valid
-      node2.wakuRlnRelay.nullifierLog.len() == 1 # after clearing, only 1 is stored
+      res1 == true
+      res2 == false
+      res3 == true
+      node2.wakuRlnRelay.nullifierLog.len() == 2
```
alrevuelta, line 490: @rymnc The change from `10` to `1` in `EpochUnitSeconds` triggered this edge case (I belive). Mind clarifying what would be expected here?

tests/waku_rln_relay/test_wakunode_rln_relay.nim
```diff
@@ -469,25 +469,25 @@ procSuite "WakuNode - RLN relay":
...
     let
       res1 = await completionFut1.withTimeout(10.seconds)
       res2 = await completionFut2.withTimeout(10.seconds)
       res3 = await completionFut3.withTimeout(10.seconds)
 
     check:
-      (res1 and res2 and res3) == true # all 3 are valid
-      node2.wakuRlnRelay.nullifierLog.len() == 1 # after clearing, only 1 is stored
+      res1 == true
+      res2 == false
```
alrevuelta, line 488: @rymnc Not sure why `rest2` was expected to complete. Since as stated above the second message "it will break the messaging rate limit" and that implies that the handler wont be triggered? So `res==false` would be expected isn't it?

alrevuelta (commented):
@rymnc This change triggered some edge cases? Mind checking my questions?

tests/waku_rln_relay/test_wakunode_rln_relay.nim
rymnc: doesn't matter too much, just showing that 2 messages were generated within the same epoch window.

tests/waku_rln_relay/test_wakunode_rln_relay.nim
rymnc: it was for message propagation i believe, due to some race condition we had earlier

tests/waku_rln_relay/test_wakunode_rln_relay.nim
rymnc: agree, oversight on my part

tests/waku_rln_relay/test_wakunode_rln_relay.nim
rymnc: this is correct, thanks for the fix 😁

rymnc (approved):
LGTM, thanks for the fixes in the edge cases!

PR #2308, opened 2023-12-19
chore: reverting shard aware connections

gabrielmer:
# Description

Reverting sharded peer management to allow shard-agnostic connections. This feature has been the root cause of different bugs and misunderstanding, so postponing it for the next release when it will have more maturity and better adoption.

# Changes

- [x] replaced `manageRelayPeers()` for its previous version `connectToRelayPeers()`
- [x] changed connection loop interval from 1 minute to 15 seconds


Ivansete-status (approved):
LGTM! Thanks for it!

PR #2313, opened 2023-12-21
feat: postgres vacuum enabled with test case

### LINKED ISSUE #1885: Add retention policy with GB or MB limitation

ABresting:
# Description
With this change, Waku Store protocol now supports retention policy on PostgreSQL. Outdated messages are deleted and DB enters a non-blocking VACUUM state. Effectively it reduces the size of the DB on disk while allowing parallel read/write operation on database.

# Changes

Apart from vacuum functionality in PostgreSQL database, I have extended the ArchiveDriver so that using the driver one can know which type of Database driver it is currently using i.e. Sqlite or Postgres or in-memory (Queue driver) etc. It was required to disable SQLite-based vacuuming smoothly.

Retention policy test cases have also been updated to support Postgres instead of SQLite. We make this choice since SQLite Vacuum process blocks the read/write operations, so it is decided to do Vacuum manually on Waku nodes/client running SQLite as store archive.

Changed the Size based retention policy test case such that it fulfills the purpose that after performing the vacuum, DB size is reduced.

- [x] Postgres Vacuum function created
- [x] SQLite focused retention policy test cases replaced by Postgres
- [x] Vacuum function added after each type of retention policy 
- [x] ArchiveDriver extended to support database type feature

## Issue
closes #1885

### DIFF
--- a/tests/waku_archive/test_retention_policy.nim
+++ b/tests/waku_archive/test_retention_policy.nim
@@ -11,7 +11,7 @@ import
   ../../../waku/waku_core,
   ../../../waku/waku_core/message/digest,
   ../../../waku/waku_archive,
-    ../../../waku/waku_archive/driver/postgres_driver,
+  ../../../waku/waku_archive/driver/postgres_driver,
   ../../../waku/waku_archive/retention_policy,
   ../../../waku/waku_archive/retention_policy/retention_policy_capacity,
   ../../../waku/waku_archive/retention_policy/retention_policy_size,
@@ -57,9 +57,8 @@ suite "Waku Archive - Retention policy":
     ## Then
     let numMessages = (waitFor driver.getMessagesCount()).tryGet()
     check:
-      # Expected number of messages is 120 because
-      # (capacity = 100) + (half of the overflow window = 15) + (5 messages added after after the last delete)
-      # the window size changes when changing `const maxStoreOverflow = 1.3 in sqlite_store
+      # Expected number of messages is 115 because
+      # (capacity = 100) + (half of the overflow window = 15)
       numMessages == 115
 
     ## Cleanup
@@ -68,52 +67,51 @@ suite "Waku Archive - Retention policy":
   test "size retention policy - windowed message deletion":
     ## Given
     let driver = newTestPostgresDriver()
-    let dbEngine = driver.getDbType()
 
     # make sure that the db is empty to before test begins
     let storedMsg = (waitFor driver.getAllMessages()).tryGet()
-    # if there are messages in db, empty them
+    # if there are messages in db, delete them before the test begins
     if storedMsg.len > 0:
       let now = getNanosecondTime(getTime().toUnixFloat())
       require (waitFor driver.deleteMessagesOlderThanTimestamp(ts=now)).isOk()  
       require (waitFor driver.performVacuum()).isOk()
 
-    # in bytes
-    let sizeLimit:int64 = 10422851
-    let excess = 800
+    # get the minimum/empty size of the database
+    let sizeLimit = int64((waitFor driver.getDatabaseSize()).tryGet())
+    let num_messages = 100
+    let retryLimit = 4
 
     let retentionPolicy: RetentionPolicy = SizeRetentionPolicy.init(size=sizeLimit)
     var putFutures = newSeq[Future[ArchiveDriverResult[void]]]()
 
     ## When
-    ##
-
-    # create a number of messages so that the size of the DB overshoots
-    for i in 1..excess:
-        let msg = fakeWakuMessage(payload= @[byte (i*10)], contentTopic=DefaultContentTopic, ts=Timestamp(i))
-        putFutures.add(driver.put(DefaultPubsubTopic, msg, computeDigest(msg), computeMessageHash(DefaultPubsubTopic, msg), msg.timestamp))
+    # create a number of messages to increase DB size
+    for i in 1..num_messages:
+      let msg = fakeWakuMessage(payload= @[byte i], contentTopic=DefaultContentTopic, ts=Timestamp(i))
+      putFutures.add(driver.put(DefaultPubsubTopic, msg, computeDigest(msg), computeMessageHash(DefaultPubsubTopic, msg), msg.timestamp))
 
     # waitFor is used to synchronously wait for the futures to complete.
     discard waitFor allFinished(putFutures)
+    sleep(150)
 
     ## Then
     # calculate the current database size
-    var sizeDB = int64((waitFor driver.getDatabaseSize()).tryGet())
+    let sizeBeforeRetPolicy = int64((waitFor driver.getDatabaseSize()).tryGet())
+    var sizeAfterRetPolicy:int64 = sizeBeforeRetPolicy
+    var retryCounter = 0
 
-    if (sizeDB < sizeLimit):
-      # we put a warning/debug that the size of the DB is less than the limit already
-      echo ("WARNING: DB size is less than the limit. This test may not work as expected")
+    while (sizeAfterRetPolicy >= sizeBeforeRetPolicy)  and (retryCounter < retryLimit):
+      # execute the retention policy
+      require (waitFor retentionPolicy.execute(driver)).isOk()
 
-    require (waitFor retentionPolicy.execute(driver)).isOk()
-
-    # get the updated DB size post vacuum
-    sizeDB = int64((waitFor driver.getDatabaseSize()).tryGet())
+      # get the updated DB size post vacuum
+      sizeAfterRetPolicy = int64((waitFor driver.getDatabaseSize()).tryGet())
+      retryCounter += 1
+      sleep(150)
 
     check:
-      # size of the database is used to check if the storage limit has been preserved
-      # check the current database size with the limitSize provided by the user
-      # it should be lower
-      sizeDB <= sizeLimit
+      # check if the size of the database has been reduced after executing the retention policy
+      sizeAfterRetPolicy < sizeBeforeRetPolicy
 
     ## Cleanup
     (waitFor driver.close()).expect("driver to close")
--- a/tests/waku_archive/test_retention_policy.nim
+++ b/tests/waku_archive/test_retention_policy.nim
@@ -4,19 +4,32 @@ import
   std/[sequtils,times],
   stew/results,
   testutils/unittests,
-  chronos
+  chronos,
+  os
 import
   ../../../waku/common/databases/db_sqlite,
   ../../../waku/waku_core,
   ../../../waku/waku_core/message/digest,
   ../../../waku/waku_archive,
-  ../../../waku/waku_archive/driver/sqlite_driver,
+    ../../../waku/waku_archive/driver/postgres_driver,
   ../../../waku/waku_archive/retention_policy,
   ../../../waku/waku_archive/retention_policy/retention_policy_capacity,
   ../../../waku/waku_archive/retention_policy/retention_policy_size,
   ../waku_archive/archive_utils,
   ../testlib/common,
   ../testlib/wakucore
+
+
+const storeMessageDbUrl = "postgres://postgres:test123@localhost:5432/postgres"
+
+proc newTestPostgresDriver(): ArchiveDriver =
+  let driver = PostgresDriver.new(dbUrl = storeMessageDbUrl).tryGet()
+  discard waitFor driver.reset()
+
+  let initRes = waitFor driver.init()
+  assert initRes.isOk(), initRes.error
+
+  return driver
 
 
 suite "Waku Archive - Retention policy":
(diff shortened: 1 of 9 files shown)


waku/waku_archive/retention_policy/retention_policy_size.nim
```diff
@@ -41,29 +41,44 @@ proc init*(T: type SizeRetentionPolicy, size=DefaultRetentionSize): T =
...
-  (await driver.deleteOldestMessagesNotWithinLimit(limit=pageDeleteWindow)).isOkOr:
-    return err("deleting oldest messages failed: " & error)
+    # perform vacuum
+    let resVaccum = await driver.performVacuum()
+    if resVaccum.isErr():
+      return err("vacuumming failed: " & resVaccum.error)
+    # recompute the DB size
+    dbSize = (await driver.getDatabaseSize()).valueOr:
+      return err("failed to get database size: " & $error)
+    totalSizeOfDB = int64(dbSize)
```
AlejandroCabeza, line 82: What purpose do these two last statements serve?

waku/waku_archive/driver/postgres_driver/postgres_driver.nim
```diff
@@ -24,6 +24,9 @@ type PostgresDriver* = ref object of ArchiveDriver
   writeConnPool: PgAsyncPool
   readConnPool: PgAsyncPool
 
+method getDbType*(s: PostgresDriver): string =
+    return "postgres"
```
AlejandroCabeza, line 28: Would it be an overkill using an `enum` for this?

AlejandroCabeza (approved):
LGTM

waku/waku_archive/driver/postgres_driver/postgres_driver.nim
```diff
@@ -535,3 +538,17 @@ proc sleep*(s: PostgresDriver, seconds: int):
...
   return ok()
+
+method performVacuum*(s: PostgresDriver):
+                     Future[ArchiveDriverResult[void]] {.async.} =
+  ## Perform a VACUUM operation to clean up the database.
+  debug "starting Postgres database vacuuming"
+  try:
+    let execRes = await s.writeConnPool.pgQuery("VACUUM")
+    if execRes.isErr:
+      return err("error in Postgres Vacuum operation: " & execRes.error)
```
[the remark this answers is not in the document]
ABresting: I am following error handling the way Ivan has just above this method. Unless I a missing something?

waku/waku_archive/retention_policy/retention_policy_size.nim
ABresting: this is to recompute the size of the DB since we need to ask Postgres the current size because with number rows we can not estimate the exact size of the data to delete so we delete 20% first and then check if the size is below the limit if not then we again slash 20%.

tests/waku_archive/test_retention_policy.nim
```diff
@@ -72,6 +78,13 @@ suite "Waku Archive - Retention policy":
       require (waitFor driver.deleteMessagesOlderThanTimestamp(ts=now)).isOk()  
       require (waitFor driver.performVacuum()).isOk()
 
+    # in bytes
+    let sizeLimit:int64 = 10122851
+    let excess = 69
```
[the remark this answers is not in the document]
ABresting: oh `sizeLimit` should be above 8 MB since that's the size even an empty DB has. so just to be safe I have used around 10 MB size. `excess` is just a random number :eyes:

waku/waku_archive/driver/postgres_driver/postgres_driver.nim
ABresting: based on the use-case yes, since we are not doing a case type conditional logic.

waku/waku_archive/retention_policy/retention_policy_size.nim
```diff
@@ -41,29 +41,49 @@ proc init*(T: type SizeRetentionPolicy, size=DefaultRetentionSize): T =
...
+  let retryLimit = 2
+  var retryCounter:int = 0
 
   if totalSizeOfDB < p.sizeLimit:
     return ok()
+  
+    # NOTE: Using SQLite vacuuming is done manually, we delete a percentage of rows
+    # if vacumming is done automatically then we aim to check DB size periodially for efficient
+    # retention policy implementation.
+  while (totalSizeOfDB > p.sizeLimit) and (retryLimit > retryCounter):
```
Ivansete-status, line 66: This check, `while (totalSizeOfDB > p.sizeLimit)`, will always be `true` after removing and applying `VACUUM`. We need to think about another approach.

Ivansete-status (commented):
Thanks for this!
Nevertheless, the behaviour isn't 100% valid because the database size is not changed at all.

I think we need to properly configure the "autocavuum" in the database ( suggested by @yakimant ) and simplify this logic. I'd remove the `while`.

On the other hand, I've tested locally with the following settings in order to configure a quite intense vacuum service, and the database gets reduced properly:

```
autovacuum = on			# Enable autovacuum subprocess?  'on'
					# requires track_counts to also be on.

# Aggressive settings for frequent autovacuum operations
autovacuum_vacuum_scale_factor = 0.01   # Trigger vacuum at 1% of dead tuples
autovacuum_vacuum_threshold = 50        # Minimum number of updated or deleted tuples before vacuum
autovacuum_analyze_scale_factor = 0.01   # Trigger analyze at 1% of changed tuples
autovacuum_analyze_threshold = 50        # Minimum number of inserted, updated, or deleted tuples before analyze
autovacuum_freeze_max_age = 200000000   # Maximum age before forced vacuum freeze

# Optionally adjust vacuum cost delay to control vacuuming speed
autovacuum_vacuum_cost_delay = 10       # Vacuuming cost delay in milliseconds
```

yakimant:
@Ivansete-status, just to mention, that I've never used `autovacuum` before and don't know the pros/cons and how it is applicable for our case. I've just seen this option somewhere.

Thanks for trying it! Would be great to know, how well does it work for us.

waku/waku_archive/driver/postgres_driver/postgres_driver.nim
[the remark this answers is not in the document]
AlejandroCabeza: I would vouch for the other way because of the reasons I wrote, but I'm definitely not going to complain if we go with the current one.

tests/waku_archive/test_retention_policy.nim
```diff
@@ -72,38 +78,42 @@ suite "Waku Archive - Retention policy":
...
-    let sizeDB = int64((waitFor driver.getDatabaseSize()).tryGet())
-
-    # NOTE: since vacuumin is done manually, this needs to be revisited if vacuuming done automatically
+    var sizeDB = int64((waitFor driver.getDatabaseSize()).tryGet())
 
-    # get the rows count pre-deletion
-    let rowsCountBeforeDeletion = (waitFor driver.getMessagesCount()).tryGet()
+    if (sizeDB < sizeLimit):
+      # we put a warning/debug that the size of the DB is less than the limit already
+      echo ("WARNING: DB size is less than the limit. This test may not work as expected")
```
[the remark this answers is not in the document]
ABresting: changed the test case logic, so now this change is not required.

waku/waku_archive/retention_policy/retention_policy_size.nim
ABresting: moved the while logic to test case and not the core retention policy.

ABresting:
> @Ivansete-status, just to mention, that I've never used `autovacuum` before and don't know the pros/cons and how it is applicable for our case. I've just seen this option somewhere.
> 
> Thanks for trying it! Would be great to know, how well does it work for us.

Basically, the `autovacuum` process in Postgres is to ensure that deleted tuples/rows space is reusable and the database size is utilized properly. There are certainly thresholds based on which a timely non-blocking simple DB vacuum is done automatically. `autovacuum` is good to have provided the parameters of `autovacuum` are set carefully.

yakimant:
I've read a bit about `vacuum`, `autovacuum`, `vacuum full` and `pg_repack`.

Looks like `autovacuum` is a must for most of the systems, which do update and or delete operations. This should make them emptyed space available for new edits.

But I don't think this will actually let the space available to the filesystem.
Probably we don't neet to take of that unless there is an instant jump of db activity.
In that case `pg_repack` is suggested by community, rather than `vacuum full`

Ivansete-status (commented):
Thanks @ABresting !
It looks nice. I've added some comments. Ping me when done and I'll double-check again.

Regarding the disk space consumption, I've noticed that the `VACUUM` doesn't work well in a normal scenario. In other words, it only worked well when I forced `VACUUM` every two seconds.

I've also tried the `autovacuum` and it didn't work well either. I couldn't manage to reduce the database size.

IMO, the only working solution is to use `pg_repack` tool, even though at first I was quite reluctant to use it (it requires installing a non-standard extension and having the `pg_repack` utility installed in the system as well.)

As you properly mentioned, this PR is not enough and we need to perform additional actions to keep the database size bounded.

ABresting:
Thanks for the nice review Ivan!
I am well aware of the situation you are mentioning. That's a Postgres
async issue where when you delete a large number of rows then they take
time to first delete and then on top if you vacuum then too it takes time.
But good part is that all those operations will eventually take place if
not instantly, since Postgres operations are best effort. Not a lot of
things we can do there. Open to hear how pg_repack tool's trade off in our
case since I have a hunch that it comes with some baggage, performance
degradation, data corruption to name some I could find.

yakimant:
@Ivansete-status vacuum and autovacuum should not make disk usage lower.
It only allows cleaned data to give space for new records.

But probably it's fine in many cases.

Ivansete-status:
> @Ivansete-status vacuum and autovacuum should not make disk usage lower. It only allows cleaned data to give space for new records.
> 
> But probably it's fine in many cases.

Thanks for the comment @yakimant ! You are absolutely right.

In order to have the "size" retention policy in _Postgres_ working correctly we need to take additional action to reduce the disk space occupied by the database. If not, every time the retention policy is applied, it will drop rows until the table gets empty.

If we cannot work on such an external tool (i.e. `pg_repack`) then we shouldn't support "size" retention policy when _Postgres_ is used (cc @ABresting .)

ABresting:
> > @Ivansete-status vacuum and autovacuum should not make disk usage lower. It only allows cleaned data to give space for new records.
> > But probably it's fine in many cases.
> 
> Thanks for the comment @yakimant ! You are absolutely right.
> 
> In order to have the "size" retention policy in _Postgres_ working correctly we need to take additional action to reduce the disk space occupied by the database. If not, every time the retention policy is applied, it will drop rows until the table gets empty.
> 
Thanks for the comment Ivan :100: 
Well, IMO if we devise the retention policy in such as way that a drop of rows only happens once and then after x amount of time (conditional to if the size is above the threshold), in that case, we should not be worried about table getting emptied. It should not work aggressively but rather best-effort policy. Now you may find it non-reliable/probable but if not having out of a box tool/solution then the current retention policy should work. If there is no harm then why not?

About the size of the disk space, I believe that from a user perspective, if cleaned space is not reclaimed to the file system, but still be used by DB to insert new messages then again it is a win-win anyway. I do not think we have run infra tests this way where the retention policy doesn't work in an aggressive way (risk of emptying DB). I think just the Table insertions were not the major issue, it was the ever-growing log size?

BTW this running/checking retention policy after x amount of time is how it is working rn in the app code I believe.

Ivansete-status:
Thanks for the comment @ABresting !

You are right that the database won't get empty in the happy scenario where the node gets messages continuously. However, we cannot always guarantee that, and there will be periods of inactivity. We cannot deliver something that won't work well in 100% of the cases.

The only solution I see, if we want to support the "size" retention policy for _Postgres_, is that we start using the `pg_repack` utility. For that, we need to use a _postgres_ docker image with that extension installed, e.g. `hartmutcouk/pg-repack-docker:1.4.8`, and on the other hand, we will need to have an external app (another docker service) that invokes the `pg_repack` regularly (`PGPASSWORD=test123 ~/utils/pg_repack-1.4.8/bin/pg_repack -U postgres -h 127.0.0.1 -p 5432 -d postgres --table messages`) so that the database size gets bounded properly.

Ivansete-status:
Thanks so much for the PR @ABresting !
In the end, we've applied a different approach and we can keep the database size controlled by utilizing partitions. See https://github.com/waku-org/nwaku/pull/2506
Excuse me, I am getting the license to close this PR according to the mentioned above.
Big hug!

PR #2314, opened 2023-12-21
feat: adding filter data admin endpoint (REST)

### LINKED ISSUE #2290: feat: Have additional Admin REST API endpoints that helps node operator in monitoring
### Problem
While debugging [issue](https://github.com/status-im/status-go/issues/4459) faced in status app, realized that it would be helpful to have some admin REST API's or some form for node operators to inspect their node in case of issues. This would also inturn help Waku team analyze an issue further(e.g in the waku network as well). 

### Suggested solution
Have a set of Admin API's for service protocols that help in inspecting some data-structures used by various protocols in a service node. 
-  For filter protocol, it would be good if we can query the list of current subscriptions that are currently active at the node. The response can include details like contentTopics and/or pubsubTopics subscribed to, peer details,  and probably current state of connection to the peer. 
- Similar analysis can be done for other service protocols as well and have such API's that can help introspect into the node further.

Note: Also observed that we are not logging contentTopics when a filter subscription is created in the logs, it would help to log it as well. https://github.com/waku-org/nwaku/blob/0cc0c80563339c9466c5e3711735a331fbc8ac2c/waku/waku_filter_v2/protocol.nim#L115

### Alternatives considered
None

### Additional context
None

### Acceptance criteria

- [ ] A Admin query REST API to query list of active filter subscriptions.

### QA Test Scenario Description
Scenario 1
1. 
2. 
Scenario 2
1. 
2.

gabrielmer:
# Description

Adding admin REST endpoint to retrieve the serving node's subscriber filter clients and their subscribed pubsub and content topics.

Endpoint: GET /admin/v1/filter/subscriptions
# Changes

- [x] mounted endpoint to retrieve filter subscriber data
- [x] added unit test verifying its proper functioning
- [x] updated `openapi.yaml` documentation 

image

## Issue

closes #2290

### DIFF
--- a/waku/waku_api/rest/admin/handlers.nim
+++ b/waku/waku_api/rest/admin/handlers.nim
@@ -114,12 +114,18 @@ proc installAdminV1PostPeersHandler(router: var RestRouter, node: WakuNode) =
 proc installAdminV1GetFilterHandler(router: var RestRouter, node: WakuNode) =
   router.api(MethodGet, ROUTE_ADMIN_V1_FILTER) do () -> RestApiResponse:
     
-    var subscriptions: seq[tuple[peerId: string, filterCriteria: string]] = @[]
+    var subscriptions: seq[FilterSubscription] = @[]
 
     if not node.wakuFilter.isNil():
 
+      var filterCriteria: seq[FilterTopic]
+
       for (peerId, criteria) in node.wakuFilter.subscriptions.pairs():
-        subscriptions.add(($peerId, $criteria.toSeq()))
+      
+        filterCriteria = criteria.toSeq().mapIt(FilterTopic(pubsubTopic: it[0],
+          contentTopic: it[1]))
+
+        subscriptions.add(FilterSubscription(peerId: $peerId, filterCriteria: filterCriteria))
 
     let resp = RestApiResponse.jsonResponse(subscriptions, status=Http200)
     if resp.isErr():
--- a/waku/waku_api/rest/admin/handlers.nim
+++ b/waku/waku_api/rest/admin/handlers.nim
@@ -121,8 +121,6 @@ proc installAdminV1GetFilterHandler(router: var RestRouter, node: WakuNode) =
       for (peerId, criteria) in node.wakuFilter.subscriptions.pairs():
         subscriptions.add(($peerId, $criteria))
 
-      echo subscriptions
-
     let resp = RestApiResponse.jsonResponse(subscriptions, status=Http200)
     if resp.isErr():
       error "An error ocurred while building the json respose: ", error=resp.error
--- a/waku/waku_api/rest/admin/handlers.nim
+++ b/waku/waku_api/rest/admin/handlers.nim
@@ -113,7 +113,10 @@ proc installAdminV1PostPeersHandler(router: var RestRouter, node: WakuNode) =
 
 proc installAdminV1GetFilterSubsHandler(router: var RestRouter, node: WakuNode) =
   router.api(MethodGet, ROUTE_ADMIN_V1_FILTER_SUBS) do () -> RestApiResponse:
-    
+
+    if node.wakuFilter.isNil():
+      return RestApiResponse.badRequest("Error: Filter Protocol is not mounted to the node")
+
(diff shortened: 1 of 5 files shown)


Ivansete-status (approved):
nitpick: I wonder if we could add more test cases and not only the "happy path" :D

gabrielmer:
> Wonderful PR! Thanks for it! 💯 nitpick: I wonder if we could add more test cases and not only the "happy path" :D

Oh yess, my bad! Copied from another endpoint, adapted it for this case and forgot about adding more 😳
Will add more comprehensive tests now! Thanks for the remark! 🤩

NagyZoltanPeter (approved):
Thanks for it, will be great for debugging as well.

waku/waku_api/rest/admin/openapi.yaml
```diff
@@ -49,6 +49,24 @@ paths:
           description: Cannot connect to one or more peers.
         '5XX':
           description: Unexpected error.
+  /admin/v1/filter:
```
chaitanyaprem, line 52: Would recommend appending `/subscriptions` to this endpoint. 
This would help in future to add some other admin sub endpoints for filter.

chaitanyaprem (approved):
Reviewed only the API details and left a minor comment.
LGTM once that is addressed.

waku/waku_api/rest/admin/openapi.yaml
gabrielmer: Great point! Will add it now :))
Thanks!

SionoiS (approved):
LGTM Thank you!

I would add an error when filter is not mounted but it's ok for now.

gabrielmer:
> LGTM Thank you!
> 
> I would add an error when filter is not mounted but it's ok for now.

Yes! I was 50-50 whether to return an error stating that filter isn't mounted or to return an empty array of subscribers in that case. I followed the second option just to be consistent with the approach of `installAdminV1GetPeersHandler`.

But I do agree that returning an error stating that the protocol is not even mounted is more informative. @Ivansete-status @NagyZoltanPeter, what do you think?

NagyZoltanPeter:
> > LGTM Thank you!
> > I would add an error when filter is not mounted but it's ok for now.
> 
> Yes! I was 50-50 whether to return an error stating that filter isn't mounted or to return an empty array of subscribers in that case. I followed the second option just to be consistent with the approach of `installAdminV1GetPeersHandler`.
> 
> But I do agree that returning an error stating that the protocol is not even mounted is more informative. @Ivansete-status @NagyZoltanPeter, what do you think?

My first thought is to return with error and correct explanation. Due this admin endpoint is strictly depends on Filter service to be mounted. Otherwise it has no meaning. Emtpy array would mean to me there is no subscriptions and Filter service is running.

waku/waku_api/rest/admin/handlers.nim
```diff
@@ -111,6 +111,32 @@ proc installAdminV1PostPeersHandler(router: var RestRouter, node: WakuNode) =
...
 
+proc installAdminV1GetFilterSubsHandler(router: var RestRouter, node: WakuNode) =
+  router.api(MethodGet, ROUTE_ADMIN_V1_FILTER_SUBS) do () -> RestApiResponse:
+
+    if node.wakuFilter.isNil():
+      return RestApiResponse.badRequest("Error: Filter Protocol is not mounted to the node")
+
+    var subscriptions: seq[FilterSubscription] = @[]
+
+    if not node.wakuFilter.isNil():
```
SionoiS, line 122: Can be remove since it's checked above

SionoiS (approved):
LGTM Thanks!

waku/waku_api/rest/admin/handlers.nim
gabrielmer: Oh yes! Was supposed to remove it but for sure got distracted and forgot 😶

Will fix it now, thanks so much!

NagyZoltanPeter (approved):
It's great!
Just to note to myself, I need to touch here as soon it gets merged - surely before my rework on managing filter subscriptions. that makes filter subscription structure upside-down. :-)

PR #2316, opened 2023-12-22
docs: add benchmarks around postgres adoption

Ivansete-status:
# Description
Adding benchmark info related to the _Postgres_ adoption.

This is expected to be consumed by https://github.com/waku-org/docs.waku.org/

Notice that all the benchmarks that we will want to make public on https://docs.waku.org should follow this approach.


gabrielmer (approved):
Woow! Really interesting read! Thanks so much for such a comprehensive investigation! 😍

NagyZoltanPeter (approved):
As being already familiar with this awesome document but still amazed! Thank a lot for it!

Ivansete-status:
Thanks for the comments! We are adding all "benchmark" documentation to `/docs` folder so that this page can be consumed and published by `docs.waku.org`.

alrevuelta:
Not sure this repo is the best place for this. Seems more suitable for a blogpost, or research issue.

Ivansete-status:
> Not sure this repo is the best place for this. Seems more suitable for a blogpost, or research issue.

The idea to add the benchmark details in the `/docs` folder of this repo is to later on consume it and produce documentation on `docs.waku.org` from _Docusaurus_.

PR #2321, opened 2023-12-27
test(peer-connection-managenent): Functional Tests

AlejandroCabeza:
# Description
Peer Connection Management Tests

# Changes
- [X] Add simple mock mechanism
- [X] Implement migration tests
- [X] Peer Storage tests
- [X] Simple Protobuf serialisation tests
  - Expanding these will be left for a later date due to a priority shift.


tests/node/peer_manager/peer_store/test_migrations.nim
```diff
@@ -0,0 +1,56 @@
...
+  ../../waku_archive/archive_utils,
+  ../../testlib/[simple_mock]
+
+import std/[tables, strutils, os], stew/results, chronicles
+
+import ../../../../waku/[common/databases/db_sqlite, common/databases/common]
+
+suite "Migrations":
+  test "migrate ok":
+    # Given the db_sqlite.migrate function returns an error
```
SionoiS, line 14: copy pasta

tests/node/peer_manager/peer_store/test_migrations.nim
```diff
@@ -0,0 +1,56 @@
...
+          db: SqliteDatabase, targetVersion: int64, migrationsScriptsDir: string
+      ): DatabaseResult[void] =
+        ok()
+
+      mockedMigrate
+
+    # When we call the migrate function
+    let migrationResult = migrations.migrate(newSqliteDatabase(), 1)
+
+    # Then we expect the result to be an error
```
SionoiS, line 27: copy pasta

SionoiS (approved):
I'm not sure I understand the point in mocking the migration.

If the mock returns `ok` and the test check for `ok` then obviously it's going to work no?

Otherwise LGTM

AlejandroCabeza:
> I'm not sure I understand the point in mocking the migration.
> 
> If the mock returns `ok` and the test check for `ok` then obviously it's going to work no?
> 
> Otherwise LGTM

The idea is to test the top-level `migrate` only, not the `db_sqlite.migrate`. For this particular instance, I don't care (and don't have the specifics) about the internals of what happens when a migration occurs: What files are created, in which format, what errors may be triggered. The top-level function's only a wrapper to provide a different (better?) error handling, and that's precisely what I wanna test: That the top-level function behaves as expected for the different results of `db_sqlite.migrate`.

tests/node/peer_manager/peer_store/test_migrations.nim
```diff
@@ -0,0 +1,56 @@
...
```
gabrielmer, line 27: Comment is wrong I think, not an error. Right?

gabrielmer (approved):
Looks great! Thank you! :))

tests/node/peer_manager/peer_store/test_migrations.nim
AlejandroCabeza: Yes, fixed.

PR #2323, opened 2023-12-29
chore: ip colocation is parameterizable. If set to 0, it is disabled

Ivansete-status:
## Description
The "ip colocation" concept refers to the maximum allowed peers from the same IP address. We allow disable this limit when the node works behing a reverse proxy.

## How to test

1. Run one node with `ip-colocation-limit` set to 0.
2. The node set in 1. can accept multiple connections from nodes run locally (127.0.0.1)

---

1. Run one node with `ip-colocation-limit` set to 2.
2. The node set in 1. only accepts two peers from 127.0.0.1.

:information_source:  in both cases remember to set `rest-admin = true` so that you can check the peers with `curl -X GET 127.0.0.1:8645/admin/v1/peers`.

## Issue

closes https://github.com/waku-org/nwaku/issues/2306

### DIFF
--- a/apps/wakunode2/app.nim
+++ b/apps/wakunode2/app.nim
@@ -303,6 +303,7 @@ proc initNode(conf: WakuNodeConf,
       sendSignedPeerRecord = conf.relayPeerExchange, # We send our own signed peer record when peer exchange enabled
       agentString = some(conf.agentString)
   )
+  builder.withColocationLimit(conf.colocationLimit)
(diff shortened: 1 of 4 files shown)


gabrielmer (approved):
Looks great! Thanks so much!

NagyZoltanPeter (approved):
LGTM! Great, I hope it also helps Filter subscriptions flood, tested by @chaitanyaprem.
Due that scenario will not work out if PeerManager can be configured to limit peers per IP.

PR #2333, opened 2024-01-04
chore: bump vendors for 0.24.0

Ivansete-status:
# Description
The following vendors have changes but are not being updated for the reason explained.

nim-web3: not updated because unit tests started to fail and no straightforward solution found.

nim-toml-serialization: not updated because it introduced a breaking change on how the --config-file attribute is parsed. The array attributes now need a comma. For example, the following attribute from within the config file:

pubsub-topic = [ "/waku/2/default-waku/proto" "/waku/2/testing-store" ]

... should be converted to:

pubsub-topic = [ "/waku/2/default-waku/proto", "/waku/2/testing-store" ]

and we cannot accept that breaking change

## Issue
closes https://github.com/waku-org/nwaku/issues/2329

### DIFF
--- a/waku/waku_api/rest/debug/types.nim
+++ b/waku/waku_api/rest/debug/types.nim
@@ -34,8 +34,8 @@ proc writeValue*(writer: var JsonWriter[RestJson], value: DebugWakuInfo)
   {.raises: [IOError].} =
   writer.beginRecord()
   writer.writeField("listenAddresses", value.listenAddresses)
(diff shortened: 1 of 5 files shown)


gabrielmer (approved):
LGTM, thanks so much! 🔥

NagyZoltanPeter (approved):
LGTM! Thanks for taking care!

SionoiS (approved):

PR #2334, opened 2024-01-04
test(autosharding): Mocked test help

AlejandroCabeza:
# Description
Get the last test case working.
This case requires either:
- Being able to raise a CatchableError in the `system.add` function.
- Mocking.

Assuming `sytem.add` works flawlessly, which I kinda do; I thought the best way forward would be to mock the function but I haven't been successful in that task. 
Does anyone know how to make this whole ordeal work? I'm lacking some Nim proficiency to get know my way around it.
There are a couple of issues that I'll try to address in comments. All help is welcome :)

# Changes
- [ ] Implement the `catchable error on topicMap.add` test case with mock


tests/waku_core/topics/test_sharding.nim
```diff
@@ -148,6 +148,23 @@ suite "Autosharding":
       check:
         topicMap.error() == "Cannot autoshard content topic: Generation > 0 are not supported yet"
 
-    xtest "catchable error on add to topicMap":
-      # TODO: Trigger a CatchableError or mock
-      discard
+    test "catchable error on add to topicMap":
+      # Given the sequence.add function returns a CatchableError
+      let backup = system.add
```
AlejandroCabeza, line 153: This is used to keep a reference to the original function, which is necessary for cleanup later. 
Produces this error: `Error: invalid type: 'None' for let`

tests/waku_core/topics/test_sharding.nim
```diff
@@ -148,6 +148,23 @@ suite "Autosharding":
...
+      mock(system.add):
```
AlejandroCabeza, line 154: And then, this attempt at mocking produces:
```
Error: ambiguous identifier: 'add' -- you need a helper proc to disambiguate the following:
  system.add: proc (x: var seq[T], y: sink T){.noSideEffect.}
  system.add: proc (x: var string, y: string){.noSideEffect, gcsafe, locks: 0.}
  system.add: proc (x: var string, y: char){.noSideEffect.}
  system.add: proc (x: var seq[T], y: openArray[T]){.noSideEffect.}
  system.add: proc (x: var string, y: cstring){.noSideEffect, gcsafe, locks: 0.}
  ```

**Review by AlejandroCabeza — COMMENTED** [review]
I've made quite a few attempts at fixing it but I've been utterly unsuccessful. Anyone has an idea on how to circumvent these errors?

**Comment by AlejandroCabeza** [comment]
# Attempts
1.  ```nim
    let nsAdd = system.add(seq[NsContentTopic])  # Error: type mismatch: got <typedesc[seq[NsContentTopic]]
    ```
2. ```nim
    type
      NsContentTopicAdd =
        proc(x: seq[NsContentTopic], y: sink NsContentTopic) {.noSideEffect.}

    let 
      nsAdd: NsContentTopicAdd = system.add  # Error: type mismatch: got 'None' for 'add' but expected 'NsContentTopicAdd = proc (x: seq[NsContentTopic], y: NsContentTopic){.closure, noSideEffect.}
    ```

# Notes
## Attempt 1
Calling add like so doesn't match, because the expected first parameter is a type instance, not a type.
## Attempt 2
One of the possible disambiguations shown by the compiler is the following:
```
Error: ambiguous identifier: 'add' -- you need a helper proc to disambiguate the following:
  ...
  system.add: proc (x: var seq[T], y: sink T){.noSideEffect.}
  ...
```
When following the same pattern as above, but with a different disambiguation, the compiler doesn't complain:
```nim
type
  NsContentTopicSeqAdd =
    proc(x: var seq[NsContentTopic], y: openArray[NsContentTopic]) {.
      noSideEffect
    .}

let
  nsSeqAdd: NsContentTopicSeqAdd = system.add
  nsSeqAddUnsafeAddr = unsafeAddr(nsSeqAdd)

echo repr(nsSeqAdd)  # [Field0 = 0x55dc4e228490, Field1 = nil]
echo repr(nsSeqAddUnsafeAddr)  # ptr 0x7ffc0384d030 --> [Field0 = 0x55dc4e228490, Field1 = nil]
```
I've also tried running `mock` with the `system.add: NsContentTopicSeqAdd` but I get to a `SEGFAULT`.

AlejandroCabeza:
Posted question to Nim Forum: [Link](https://forum.nim-lang.org/t/10854)

PR #2341, opened 2024-01-09
chore: Refactor of FilterV2 subscription management with Time-to-live maintenance

NagyZoltanPeter:
# Description
As during Status dogfooding it turned out we need to handle situation with stale filter subscriptions to ensure  service quality.

NOTE: This become a relative large refactor on management filter subscription. Sorry for that to issue such a big PR.
The structure changes involved a lot changes in filter v2 test.
But because of the I would like to ask you to start reviewing it, if there is something not ok should come early as possible.
THX!!!

NOTE2: Still a work in progress, as realized need to add some more test cases.

# Changes

- [X] New CLI configuration options are added for filter v2: subscription timeout, max peers served and max criterion per peer. 
- [X] From now, a filter client needs to periodically refresh it's subscriptions with PING (within the timeout period) in order to receive message push
- [X] Filter v2 subscriptions are fully reworked. New structure is introduced for better speed. Now subscription management is now encapsulated into its own module.
- [X] test_waku_client adapted to changes.
 - [X] Mount of legacy and filter v2 are now separated to ease future removal of legacy filter.
 
## Issue
https://github.com/waku-org/nwaku/issues/2293

covers also: https://github.com/waku-org/nwaku/issues/2338

### DIFF
--- a/apps/wakunode2/app.nim
+++ b/apps/wakunode2/app.nim
@@ -254,7 +254,7 @@ proc setupDiscoveryV5*(app: App): WakuDiscoveryV5 =
   )
 
   WakuDiscoveryV5.new(
-    app.rng, 
+    app.rng,
     discv5Conf,
     some(app.record),
     some(app.node.peerManager),
@@ -326,7 +326,7 @@ proc setupWakuApp*(app: var App): AppResult[void] =
   ok()
 
 proc getPorts(listenAddrs: seq[MultiAddress]):
-              AppResult[tuple[tcpPort, websocketPort: Option[Port]]] = 
+              AppResult[tuple[tcpPort, websocketPort: Option[Port]]] =
 
   var tcpPort, websocketPort = none(Port)
 
@@ -548,7 +548,15 @@ proc setupProtocols(node: WakuNode,
   # Filter setup. NOTE Must be mounted after relay
   if conf.filter:
     try:
-      await mountFilter(node, filterTimeout = chronos.seconds(conf.filterTimeout))
+      await mountLegacyFilter(node, filterTimeout = chronos.seconds(conf.filterTimeout))
+    except CatchableError:
+      return err("failed to mount waku legacy filter protocol: " & getCurrentExceptionMsg())
+
+    try:
+      await mountFilter(node,
+                        subscriptionTimeout = chronos.seconds(conf.filterSubscriptionTimeout),
+                        maxFilterPeers = conf.filterMaxPeersToServe,
+                        maxFilterCriteriaPerPeer = conf.filterMaxCriteria)
     except CatchableError:
       return err("failed to mount waku filter protocol: " & getCurrentExceptionMsg())
 
@@ -724,7 +732,7 @@ proc startRestServer(app: App, address: IpAddress, port: Port, conf: WakuNodeCon
 
     let filterCache = MessageCache.init()
 
-    let filterDiscoHandler = 
+    let filterDiscoHandler =
       if app.wakuDiscv5.isSome():
         some(defaultDiscoveryHandler(app.wakuDiscv5.get(), Filter))
       else: none(DiscoveryHandler)
@@ -739,7 +747,7 @@ proc startRestServer(app: App, address: IpAddress, port: Port, conf: WakuNodeCon
     notInstalledTab["filter"] = "/filter endpoints are not available. Please check your configuration: --filternode"
 
   ## Store REST API
-  let storeDiscoHandler = 
+  let storeDiscoHandler =
     if app.wakuDiscv5.isSome():
       some(defaultDiscoveryHandler(app.wakuDiscv5.get(), Store))
     else: none(DiscoveryHandler)
@@ -749,7 +757,7 @@ proc startRestServer(app: App, address: IpAddress, port: Port, conf: WakuNodeCon
   ## Light push API
   if conf.lightpushnode  != "" and
      app.node.wakuLightpushClient != nil:
-    let lightDiscoHandler = 
+    let lightDiscoHandler =
       if app.wakuDiscv5.isSome():
         some(defaultDiscoveryHandler(app.wakuDiscv5.get(), Lightpush))
       else: none(DiscoveryHandler)
--- a/apps/wakunode2/external_config.nim
+++ b/apps/wakunode2/external_config.nim
@@ -319,9 +319,8 @@ type
         defaultValue: 1000
         name: "filter-max-peers-to-serve" }: uint32
 
-
       filterMaxCriteria* {.
-        desc: "Maximum number of criteria per peers at a time. Only for v2 filter protocol.",
+        desc: "Maximum number of pubsub- and content topic combination per peers at a time. Only for v2 filter protocol.",
         defaultValue: 1000
         name: "filter-max-criteria" }: uint32
 
--- a/apps/wakunode2/external_config.nim
+++ b/apps/wakunode2/external_config.nim
@@ -95,6 +95,7 @@ type
         defaultValue: false,
         name: "execute" .}: bool
 
+
     of noCommand:
       ##  Application-level configuration
       protectedTopics* {.
@@ -221,7 +222,7 @@ type
         desc: "Rln relay identity commitment key as a Hex string",
         defaultValue: ""
         name: "rln-relay-id-commitment-key" }: string
-      
+
       rlnRelayTreePath* {.
         desc: "Path to the RLN merkle tree sled db (https://github.com/spacejam/sled)",
         defaultValue: ""
@@ -304,9 +305,25 @@ type
         name: "filternode" }: string
 
       filterTimeout* {.
-        desc: "Timeout for filter node in seconds.",
+        desc: "Filter clients will be wiped out if not able to receive push messages within this timeout. In seconds.",
         defaultValue: 14400 # 4 hours
         name: "filter-timeout" }: int64
+
+      filterSubscriptionTimeout* {.
+        desc: "Timeout for filter subscription without ping or refresh it, in seconds. Only for v2 filter protocol.",
+        defaultValue: 300 # 5 minutes
+        name: "filter-subscription-timeout" }: int64
+
+      filterMaxPeersToServe* {.
+        desc: "Maximum number of peers to serve at a time. Only for v2 filter protocol.",
+        defaultValue: 1000
+        name: "filter-max-peers-to-serve" }: uint32
+
+
+      filterMaxCriteria* {.
+        desc: "Maximum number of criteria per peers at a time. Only for v2 filter protocol.",
+        defaultValue: 1000
+        name: "filter-max-criteria" }: uint32
 
       ## Lightpush config
 
--- a/tests/waku_filter_v2/test_waku_client.nim
+++ b/tests/waku_filter_v2/test_waku_client.nim
@@ -2,10 +2,10 @@
 
 import
   std/[
-    options, 
-    tables, 
-    sequtils, 
-    strutils, 
+    options,
+    tables,
+    sequtils,
+    strutils,
     json
   ],
   testutils/unittests,
@@ -39,7 +39,6 @@ import
   ],
   ./waku_filter_utils,
   ../resources/payloads
-
 
 suite "Waku Filter - End to End":
   suite "MessagePushHandler - Void":
@@ -77,27 +76,27 @@ suite "Waku Filter - End to End":
       wakuFilterClient.registerPushHandler(messagePushHandler)
       serverRemotePeerInfo = serverSwitch.peerInfo.toRemotePeerInfo()
       clientPeerId = clientSwitch.peerInfo.toRemotePeerInfo().peerId
-      
+
     asyncTeardown:
       await allFutures(wakuFilter.stop(), wakuFilterClient.stop(), serverSwitch.stop(), clientSwitch.stop())
 
     suite "Subscriber Ping":
       asyncTest "Active Subscription Identification":
         # Given
-        let 
+        let
           subscribeResponse = await wakuFilterClient.subscribe(
             serverRemotePeerInfo, pubsubTopic, contentTopicSeq
           )
         assert subscribeResponse.isOk(), $subscribeResponse.error
-        check wakuFilter.subscriptions.hasKey(clientPeerId)
+        check wakuFilter.subscriptions.isSubscribed(clientPeerId)
+        check wakuFilter.subscriptions.peersSubscribed.hasKey(clientPeerId)
 
         # When
         let subscribedPingResponse = await wakuFilterClient.ping(serverRemotePeerInfo)
-
         # Then
         assert subscribedPingResponse.isOk(), $subscribedPingResponse.error
         check:
-          wakuFilter.subscriptions.hasKey(clientPeerId)
+          wakuFilter.subscriptions.isSubscribed(clientPeerId)
 
       asyncTest "No Active Subscription Identification":
         # When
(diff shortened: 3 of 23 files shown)


apps/wakunode2/external_config.nim
```diff
@@ -304,10 +305,26 @@ type
         name: "filternode" }: string
 
       filterTimeout* {.
-        desc: "Timeout for filter node in seconds.",
+        desc: "Filter clients will be wiped out if not able to receive push messages within this timeout. In seconds.",
```
SionoiS, line 308: I'm a bit confused.

Does this means that a Filter node will try to push messages to a client for 4 hours before giving up?

If a client has to ping every 5m or the sub is dropped, how does that make sense? I don't understand.

waku/waku_filter_v2/subscriptions.nim
```diff
@@ -6,48 +6,161 @@ else:
...
+proc cleanUp*(fs: var FilterSubscriptions) =
+  ## Remove all subscriptions for peers that have not been seen for a while
+  let now = Moment.now()
+  fs.peersSubscribed.keepItIf(now - val.lastSeen <= fs.subscriptionTimeout)
+
+  var filtersToRemove: seq[FilterCriterion] = @[]
+  for filterCriterion, subscribedPeers in fs.subscriptions.mpairs:
+    subscribedPeers.keepItIf(fs.isSubscribed(it)==true)
+
+  fs.subscriptions.keepItIf(val.len() > 0)
```
SionoiS, line 110: Minor thing but I would put the clean-up in a timed loop instead of calling it when events happens.

That way you can "forget" about it.

SionoiS (commented):
Awesome stuff! LGTM.

waku/waku_filter_v2/subscriptions.nim
```diff
@@ -6,48 +6,161 @@ else:
...
+  s.peersSubscribed.withValue(peerId, data):
+    if data.criteriaCount + cast[uint](filterCriteria.len()) > s.maxCriteriaPerPeer:
+      return err("Max number of criteria for peer (" & $s.maxCriteriaPerPeer & ") reached")
+
+    data.lastSeen = Moment.now()
+    peerData = data
+
+  do:
+    ## not yet subscribed
+    if cast[uint](s.peersSubscribed.len()) >= s.maxPeers:
```
AlejandroCabeza, line 129: There's a different usage in the equalities for both `max` values:
```nim
> s.maxCriteriaPerPeer  # line 121
>= s.maxPeers # line 129
```
Is this due to some particular reason? It feels like we should be using the same "method"

AlejandroCabeza (commented):
Liking how the changes are looking 😄

apps/wakunode2/app.nim
```diff
@@ -548,7 +548,15 @@ proc setupProtocols(node: WakuNode,
   # Filter setup. NOTE Must be mounted after relay
   if conf.filter:
     try:
-      await mountFilter(node, filterTimeout = chronos.seconds(conf.filterTimeout))
+      await mountLegacyFilter(node, filterTimeout = chronos.seconds(conf.filterTimeout))
+    except CatchableError:
+      return err("failed to mount waku legacy filter protocol: " & getCurrentExceptionMsg())
+
+    try:
```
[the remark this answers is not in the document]
SionoiS: I did try something like this at one point.
```
let myValue = (catch: await myProc(stuff)).isOkOr:
   #handler error here
```
but it didn't work, I don't remember why.

apps/wakunode2/external_config.nim
NagyZoltanPeter: Well, its truely confusing, hence the thread: https://discord.com/channels/1110799176264056863/1192407282420551742

So filterTimeout is for legacy filter only. 
There is the point where we started diverge the two kind of filter so much in terms of behavior.
Only solution I see is to soon remove legacy filter functionality.

waku/waku_filter_v2/subscriptions.nim
NagyZoltanPeter: Well its there: https://github.com/waku-org/nwaku/blob/7e57b67933a21deac9d42db5571a6357cc0f54bb/waku/waku_filter_v2/protocol.nim#L183

There is a maintenance loop running every 1min...
But due to the nature of subscription limits we must remove from the store as expected the client will re-subscribe in a second call.

tests/waku_filter_v2/test_waku_client.nim
```diff
@@ -483,10 +480,10 @@ suite "Waku Filter - End to End":
         let unsubscribeResponse = await wakuFilterClient.unsubscribeAll(serverRemotePeerInfo)
 
         # Then the unsubscription is successful
-        assert unsubscribeResponse.isOk(), $unsubscribeResponse.error
+        # assert unsubscribeResponse.isOk(), $unsubscribeResponse.error
```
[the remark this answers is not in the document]
NagyZoltanPeter: Great catch! Thank you. 
Its a leftover from fixing tests. Unfortunately some cases result in segfault in tests... and tried to find the root cause, hence it is commented out and forgot to uncomment.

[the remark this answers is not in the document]
AlejandroCabeza: That's likely due to an issue with the unittest module (the segfault hides the actual exception). @Ivansete-status submitted a PR yesterday that should fix it.

waku/waku_filter_v2/subscriptions.nim
SionoiS: Ah I thought we didn't have that NVM then, all good!

apps/wakunode2/external_config.nim
SionoiS: I can't wait to remove legacy filter it's gonna be so much cleaner!

tests/waku_filter_v2/waku_filter_utils.nim
```diff
@@ -41,8 +43,20 @@ proc newTestWakuFilterClient*(switch: Switch): Future[WakuFilterClient] {.async.
...
 
   return contentTopics
+
+proc cmpSeqNoOrder*[T](seq1, seq2: seq[T]): bool {.noSideEffect.} =
+  if seq1.len() != seq2.len():
+    return false
+
+  for item in seq1:
+    if item notin seq2:
+      return false
```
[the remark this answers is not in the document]
NagyZoltanPeter: Ah yes. I will fix this... it was in my head that I need to compare here unique items as they are coming from a set originally.

waku/utils/tableutils.nim
```diff
@@ -0,0 +1,31 @@
...
+    var itemsToDelete: seq[A]
+    let key {.inject.} : A
+    let val {.inject.} : B
+
+    for key, val in t:
+      if not itPredicate:
+        itemsToDelete.add(key)
+
+    for item in itemsToDelete:
+      t.del(item)
```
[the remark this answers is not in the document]
NagyZoltanPeter: Not, this last one is working on TableRef, not Table.
It's not cpp unfortunately that could handle with ease through a template.
Nim template is closer to C macro.

waku/waku_filter_v2/subscriptions.nim
```diff
@@ -6,48 +6,161 @@ else:
...
+  PeerData* = tuple
+    lastSeen: Moment
+    criteriaCount: uint
+
+  FilterSubscriptions* = object
+    peersSubscribed*    : Table[PeerID, PeerData]
+    subscriptions*      : Table[FilterCriterion, SubscribedPeers]
+    subscriptionTimeout*: Duration
+    maxPeers*           : uint
+    maxCriteriaPerPeer* : uint
```
Ivansete-status, line 43: The following attributes can be private (unless tests use them)
```suggestion
    subscriptions      : Table[FilterCriterion, SubscribedPeers]
    subscriptionTimeout: Duration
    maxPeers           : uint
    maxCriteriaPerPeer : uint
```

waku/waku_filter_v2/subscriptions.nim
```diff
@@ -6,48 +6,161 @@ else:
...
+            )
+
+proc isSubscribed*(s: var FilterSubscriptions, peerId: PeerID): bool =
+  s.peersSubscribed.withValue(peerId, data):
+    return Moment.now() - data.lastSeen <= s.subscriptionTimeout
+
+  return false
+
+proc subscribedPeerCount*(s: FilterSubscriptions): uint =
+  return cast[uint](s.peersSubscribed.len())
```
Ivansete-status, line 65: Tiny detail. When using a "noun", better use it without parenthesis. Applies elsewhere.
```suggestion
  return cast[uint](s.peersSubscribed.len)
```

waku/waku_filter_v2/subscriptions.nim
```diff
@@ -6,48 +6,161 @@ else:
...
 
-  var subscribedPeers: seq[PeerID]
+proc getPeerSubscriptions*(s: var FilterSubscriptions, peerId: PeerID): seq[FilterCriterion] =
+  ## Get all pubsub-content topics a peer is subscribed to
+  var subscribedContentTopics: seq[FilterCriterion] = @[]
+  s.peersSubscribed.withValue(peerId, data):
```
Ivansete-status, line 70: What happens if the `peerId` doesn't exist in the `peersSubscribed` table? Does it crash?

Ivansete-status (commented):
Thanks for this great job! Given that is quite a large PR I think is better to comment it little by little.
I hope you find my comments useful. Ping me in case of doubt or whatever :)

apps/wakunode2/external_config.nim
```diff
@@ -304,10 +305,26 @@ type
...
+        name: "filter-subscription-timeout" }: int64
+
+      filterMaxPeersToServe* {.
+        desc: "Maximum number of peers to serve at a time. Only for v2 filter protocol.",
+        defaultValue: 1000
+        name: "filter-max-peers-to-serve" }: uint32
+
+
+      filterMaxCriteria* {.
+        desc: "Maximum number of criteria per peers at a time. Only for v2 filter protocol.",
```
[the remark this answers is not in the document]
AlejandroCabeza: Could be useful, I certainly had issues when I first read that 😆

tests/waku_filter_v2/test_waku_client.nim
```diff
@@ -2150,3 +2154,356 @@ suite "Waku Filter - End to End":
...
+      let msg1 = fakeWakuMessage(contentTopic=contentTopic)
+      await wakuFilter.handleMessage(pubsubTopic, msg1)
+
+      check await pushHandlerFuture.withTimeout(FUTURE_TIMEOUT)
+      let (pushedMsgPubsubTopic1, pushedMsg1) = pushHandlerFuture.read()
+      check:
+        pushedMsgPubsubTopic1 == pubsubTopic
+        pushedMsg1 == msg1
+
+      sleep(2500)
```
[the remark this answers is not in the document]
AlejandroCabeza: I'd be down to reducing it. Didn't wanna go too small in the first pass as that could cause some errors to fail due to timeouts

waku/waku_filter_v2/subscriptions.nim
```diff
@@ -6,48 +6,161 @@ else:
...
+
+  for filterCriterion in filterCriteria:
+    var peersOfSub = addr(s.subscriptions.mgetOrPut(filterCriterion, SubscribedPeers()))
+    if peerId notin peersOfSub[]:
+      peersOfSub[].incl(peerId)
+      peerData.criteriaCount += 1
+
+  return ok()
+
+proc removeSubscription*(s: var FilterSubscriptions, peerId: PeerID, filterCriteria: FilterCriteria): Result[void, string] =
```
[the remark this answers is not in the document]
SionoiS: @Ivansete-status and me made a deal, it's 100 chars not one more :P

waku/waku_filter_v2/subscriptions.nim
```diff
@@ -6,48 +6,161 @@ else:
...
+  var peerData: ptr PeerData
+
+  s.peersSubscribed.withValue(peerId, data):
+    if data.criteriaCount + cast[uint](filterCriteria.len()) > s.maxCriteriaPerPeer:
+      return err("peer has reached maximum number of filter criteria")
+
+    data.lastSeen = Moment.now()
+    peerData = data
+
+  do:
```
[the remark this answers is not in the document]
SionoiS: I would like to know as well. I never seen that before.

SionoiS (approved):
Amazing! Thank you.

waku/waku_filter_v2/subscriptions.nim
SionoiS: > What happens if the `peerId` doesn't exist in the `peersSubscribed` table? Does it crash?

I think in that case the scope (context?) is not entered.

gabrielmer: exactly, seems that it doesn't enter the following block
https://nim-lang.org/docs/tables.html#withValue.t%2CTable%5BA%2CB%5D%2CA%2Cuntyped%2Cuntyped%2Cuntyped

tests/wakunode_rest/test_rest_admin.nim
```diff
@@ -107,10 +107,11 @@ suite "Waku v2 Rest API - Admin":
       pubsubTopicNode2 = DefaultPubsubTopic
       pubsubTopicNode3 = PubsubTopic("/waku/2/custom-waku/proto")
 
+      ## TODO: Note that such checks may depend heavily on the order of the returned data!
```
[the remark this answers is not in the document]
gabrielmer: Oh yes, that's my bad. I hardcoded the output for each peer and checked with `contains` for each peer in the returned string, so the test does not depend on the order of the peers https://github.com/waku-org/nwaku/blob/8bd9c0449438991adde13d9c2f7dd3b1d6a54fde/tests/wakunode_rest/test_rest_admin.nim#L137-L138

What I missed is that the pubsub and content topics for each peer can also come in different orders.

I'll think a way of refactoring it, pretty sure a set is not enough. We have to check that there's N unordered entries in the final response and each can contain a variable number of unordered entries within their bodies.

NagyZoltanPeter:
I can confirm that only one js-waku failed tests relates to this PR only. We talked with @fbarbu15 about it. He is about to catch up with new limit on subscription count per requests from 30 to 100. Due that was the cause of not matched expected result.
Other tests are related to older discrepancy with nwaku.

With this in mind please review this PR.
@AlejandroCabeza @SionoiS @gabrielmer. Of course any further comments are welcome!

tests/waku_filter_v2/test_waku_client.nim
```diff
@@ -2150,3 +2154,356 @@ suite "Waku Filter - End to End":
         check:
           pushedMsgPubsubTopic == pubsubTopic
           pushedMsg == msg
+
+  suite "Subscription timeout":
```
gabrielmer, line 2158: What's the Filter timeout value used in this test? Can't seem to find it. Seems that `newTestWakuFilterClient` uses the default 5 mins and that the timeout used in the tests are the Future's timeouts and not Filter's.

Or am I missing something?

gabrielmer (commented):
Looks amazing! 🔥 🔥 🔥 

Added some small comments/questions :)

waku/waku_filter_v2/subscriptions.nim
```diff
@@ -6,48 +6,163 @@ else:
...
 
 logScope:
   topics = "waku filter subscriptions"
 
 const
-  MaxTotalSubscriptions* = 1000 # TODO make configurable
-  MaxCriteriaPerSubscription* = 1000
+  MaxFilterPeers* = 1000
+  MaxFilterCriteriaPerPeer* = 1000
+  DefaultSubscriptionTimeToLiveSec* = 5.minutes
```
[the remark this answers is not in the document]
gabrielmer: Makes sense! Sounds like a good idea :))

Thank you!

gabrielmer (approved):
Thanks so much! 😍

waku/waku_filter_v2/subscriptions.nim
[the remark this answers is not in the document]
SionoiS: > Unfortunately I found nim std containers has an insane unique approach to containers and algorithms on them, that does not follow any other - well polished - language approaches.

In Rust, all collections can be iterated and anything that impl. iterator can be filtered, mapped, etc...

In Nim, not so much.

SionoiS (approved):

PR #2355, opened 2024-01-15
fix: making filter admin data test order independent

gabrielmer:
# Description

Generalized test to not rely on order of neither peers subscriptions nor of subscription criteria within each peer.

### DIFF
--- a/tests/wakunode_rest/test_rest_admin.nim
+++ b/tests/wakunode_rest/test_rest_admin.nim
@@ -107,16 +107,6 @@ suite "Waku v2 Rest API - Admin":
       pubsubTopicNode2 = DefaultPubsubTopic
       pubsubTopicNode3 = PubsubTopic("/waku/2/custom-waku/proto")
 
-      ## TODO: Note that such checks may depend heavily on the order of the returned data!
-      expectedFilterData2 = fmt"(peerId: ""{$peerInfo2}"", filterCriteria:" &
-        fmt" @[(pubsubTopic: ""{pubsubTopicNode2}"", contentTopic: ""{contentFiltersNode2[1]}""), " &
-        fmt"(pubsubTopic: ""{pubsubTopicNode2}"", contentTopic: ""{contentFiltersNode2[2]}""), " &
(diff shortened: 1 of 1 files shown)


tests/wakunode_rest/test_rest_admin.nim
```diff
@@ -133,8 +124,16 @@ suite "Waku v2 Rest API - Admin":
...
+
+    let
+      peers = @[getRes.data[0].peerId, getRes.data[1].peerId]
+      numCriteria = @[getRes.data[0].filterCriteria.len, getRes.data[1].filterCriteria.len]
+
+    check:
+      $peerInfo2 in peers
+      $peerInfo3 in peers
+      2 in numCriteria
+      3 in numCriteria
```
SionoiS, line 136: Is it enough to only check the number of criteria? I guess the filter tests have this covered right?

SionoiS (approved):
Thanks!

LGTM

tests/wakunode_rest/test_rest_admin.nim
gabrielmer: If the data array is of size 2, then numCriteria (as defined now) will always have size 2, there's no need to check that.

So what we check here is that of the 2 peers that we have, one has 2 subscriptions and the other 3.

I agree that Filter tests kind of cover this scenario, but not completely - because here we're testing the output from the REST API handler, and not the internal data structures. The REST API can return wrong values even if Filter itself works fine 😵‍💫

gabrielmer:
> We may think of how to compare resultset unordered. 

Yes! I changed the case to check for the number of content topics subscribed and not their values to avoid complexity, as I think the check is strong enough in this case.

The problem here why I didn't use a set is that there's two layers of order: there's the list of peers, and within each peer there's a list of content topics. So we would have to use a set of sets, or something like that, and might overcomplicate things.

There might be a more elegant solution, we can think it through :). Sorting is also a option but kind of tedious in this case, for the same reason.

> Also now this change is in contradict with my PR.. but that can be solved as soon as this one gets merged.

Yes! Once this is merged we'll probably have to rebase your PR to include this fix :))

SionoiS (approved):

PR #2360, opened 2024-01-17
fix: Fix test for filter client receiving messages after restart

jm-clius:
Fixes https://github.com/waku-org/nwaku/issues/2175. 

This fixes one test for filter clients: filter clients should continue to receive messages from active subscriptions after a restart. Although the behaviour in the Waku node was as expected, the test failed for two reasons:
1. a Waku node will not update its advertised addresses if its listening port was dynamically allocated after instantiating the node with port `0`. This behaviour was recently fixed for the `wakunode2` app, but not for the `WakuNode` object itself. Issue/feature request reported here: https://github.com/waku-org/nwaku/issues/2359. For now this was fixed by using statically allocated ports.
2. for some reason just `stop`ping an instantiated `node` object does not fully clean up the underlying libp2p switch and the node is unreachable from other nodes. I've investigated as to the source of this, but haven't come up with a satisfying explanation yet. For now an easier fix is to use a node "clone" with the same peerId and listening port to mimic a node going offline and coming back online.

Note that I also changed all `waitFor`s to `await` within async blocks.

### DIFF
--- a/tests/node/test_wakunode_filter.nim
+++ b/tests/node/test_wakunode_filter.nim
@@ -37,6 +37,7 @@ import
 suite "Waku Filter - End to End":
   var client {.threadvar.}: WakuNode
   var clientPeerId {.threadvar.}: PeerId
+  var clientClone {.threadvar.}: WakuNode
   var server {.threadvar.}: WakuNode
   var serverRemotePeerInfo {.threadvar.}: RemotePeerInfo
   var pubsubTopic {.threadvar.}: PubsubTopic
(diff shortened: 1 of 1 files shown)


gabrielmer (approved):
LGTM, thanks so much!

Will also take a look on the issue of instantiating `WakuNode` with port 0 :))

IIRC, when fixing the issue in `wakunode2` the first attempted approach was for `WakuNode` to have this feature independent of the application.

I think that at that moment we decided not to do it because the announced addresses are computed at the application level based on configurations that `WakuNode` does not have access to. Will take a deeper look, there should for sure be a better solution :)

tests/node/test_wakunode_filter.nim
```diff
@@ -192,9 +197,8 @@ suite "Waku Filter - End to End":
       server.wakuFilter.subscriptions.subscribedPeerCount() == 1
 
     # And the client node reboots
-    waitFor client.stop()
-    waitFor client.start()
-    discard client.mountFilterClient()
+    await client.stop()
+    await clientClone.start() # Mimic restart by starting the clone
```
AlejandroCabeza, line 201: Why do this through the clone?

tests/node/test_wakunode_filter.nim
```diff
@@ -204,7 +208,7 @@ suite "Waku Filter - End to End":
     check (not await pushHandlerFuture.withTimeout(FUTURE_TIMEOUT))
 
     # Given the client refreshes the subscription
-    let subscribeResponse2 = await client.filterSubscribe(
+    let subscribeResponse2 = await clientClone.filterSubscribe(
```
AlejandroCabeza, line 211: Same here

tests/node/test_wakunode_filter.nim
jm-clius: `client.stop()` does not seem to clean the libp2p switch context completely refuses connections after a `start()`. I investigated but couldn't identify the cause, but using a clone seems like a reasonable workaround to test the same filter functionality. The programmatic starting-stopping-restarting of a node only occurs within test cases and this does not affect app restarts afaict.

PR #2363, opened 2024-01-19
chore: changing digest and hash log format from bytes to hex

### LINKED ISSUE #2362: bug: restart loop of current master
### Problem
Current builds of `master` get stuck in a restart loop on `status.prod` nodes.

The current setup involves PostgreSQL database, for flags see [this](https://github.com/status-im/infra-status/blob/master/ansible/group_vars/status-node.yml).

### Llogs
Here's the loop in logs:
```
DBG 1/7 Setting up storage                     topics="wakunode main" tid=1 file=wakunode2.nim:63
DBG 2/7 Retrieve dynamic bootstrap nodes       topics="wakunode main" tid=1 file=wakunode2.nim:71
DBG Discovering nodes using Waku DNS discovery topics="wakunode app" tid=1 file=app.nim:192 url=enrtree://AL65EKLJAUXKKPG43HVTML5EFFWEZ7L4LOKTLZCLJASG4DSESQZEC@prod.status.nodes.status.im
DBG init WakuDnsDiscovery                      topics="wakunode app" tid=1 file=waku_dnsdisc.nim:93 locationUrl=enrtree://AL65EKLJAUXKKPG43HVTML5EFFWEZ7L4LOKTLZCLJASG4DSESQZEC@prod.status.nodes.status.im
DBG init success                               topics="wakunode app" tid=1 file=waku_dnsdisc.nim:99
INF Finding peers using Waku DNS discovery     topics="waku dnsdisc" tid=1 file=waku_dnsdisc.nim:52
INF Successfully discovered ENR                topics="waku dnsdisc" tid=1 file=waku_dnsdisc.nim:66 count=6
INF Successfully discovered nodes              topics="waku dnsdisc" tid=1 file=waku_dnsdisc.nim:83 count=6
DBG 3/7 Initializing node                      topics="wakunode main" tid=1 file=wakunode2.nim:78
INF Initializing networking                    tid=1 file=waku_node.nim:143 addrs="@[/dns4/node-02.do-ams3.status.prod.statusim.net/tcp/30303, /dns4/node-02.do-ams3.status.prod.statusim.net/tcp/443/wss]"
DBG no relay sharding information, peer filtering disabled topics="waku discv5" tid=1 file=waku_discv5.nim:62
DBG 4/7 Mounting protocols                     topics="wakunode main" tid=1 file=wakunode2.nim:85
INF Created WakuMetadata protocol              topics="waku node" tid=1 file=protocol.nim:115 clusterId=0
DBG Setting max message size                   topics="wakunode app" tid=1 file=app.nim:448 num_bytes=153600
INF mounting relay protocol                    topics="waku node" tid=1 file=waku_node.nim:377
INF relay mounted successfully                 topics="waku node" tid=1 file=waku_node.nim:396
DBG subscribe                                  topics="waku node" tid=1 file=waku_node.nim:269 pubsubTopic=/waku/2/default-waku/proto
DBG subscribe                                  topics="waku relay" tid=1 file=protocol.nim:222 pubsubTopic=/waku/2/default-waku/proto
INF mounting rendezvous discovery protocol     topics="waku node" tid=1 file=waku_node.nim:1079
INF mounting libp2p ping protocol              topics="waku node" tid=1 file=waku_node.nim:1037
ERR 4/7 Mounting protocols failed              topics="wakunode main" tid=1 file=wakunode2.nim:89 error="failed to setup archive driver: Postgres has been configured but not been compiled. Check compiler definitions."
```
Then it just restarts:
```
admin@node-02.do-ams3.status.prod:/docker/nim-waku % d
CONTAINER ID   NAMES      IMAGE                                                     CREATED         STATUS
364fa9231fb2   nim-waku   harbor.status.im/wakuorg/nwaku:deploy-status-prod-trace   8 minutes ago   Restarting (1) 54 seconds ago
```
### nwaku version/commit hash
`8a9fad29`

gabrielmer:
# Description
Changing `digest` and `messageHash` formats in log from bytes to hex for readability.

## Issue

closes #2362


SionoiS (approved):
LGTM!

Would it make sense to truncate the hex string as well?

apentori:
> I agree 100% that having the hash truncated is better/most comfortable. The issue is that this was requested so messages could be easily matched with Postgres errors such as

Actually it is requested to help status Desktop and Mobile team identify message relialbility issues.

From what @chaitanyaprem  told me,the Postgres error is like to the fleet architecture : 2 nodes trying to write on 1 db.
The second node will always fail due to the uniqueness constraint in the db

waku/waku_archive/archive.nim
```diff
@@ -111,7 +111,7 @@ proc handleMessage*(w: WakuArchive,
       msgReceivedTime = if msg.timestamp > 0: msg.timestamp
                         else: getNanosecondTime(getTime().toUnixFloat())
 
-    trace "handling message", pubsubTopic=pubsubTopic, contentTopic=msg.contentTopic, timestamp=msg.timestamp, digest=msgDigest, messageHash=msgHash
+    trace "handling message", pubsubTopic=pubsubTopic, contentTopic=msg.contentTopic, timestamp=msg.timestamp, digest=toHex(msgDigest.data), messageHash=toHex(msgHash)
```
NagyZoltanPeter, line 114: Upon latest ligging research, can we just move any conversion/calculation away from logging line?
Until we don't know what cause double call of any proc appears in the log arguments.
It is also involves memory allocation for the hex string, I would avoid that.

NagyZoltanPeter (changes requested):
I hope you agree, Doing hex conversion separately makes is more robust and tempt less side effect.

waku/waku_archive/archive.nim
chaitanyaprem: But this leads to another problem, if trace is disabled the hash would still be calculated which would reduce performance even in normal scenario. 
Unless we can access and use the logging macro and put the hash calculation statement under it.

gabrielmer: I actually just discussed this point with Zoltán :))

In such cases we can check the log level and compute depending on it. I just added this fix without doing a specific check for log level, as I'm doing an extremely cheap computation - converting a constant size hash from bytes to hex.

For bigger computations, I guess the best thing is to compute it based on the log level.

Most important thing might be to figure out why computation happens twice when added as a log argument. The biggest problem here is the double memory allocation.

NagyZoltanPeter: Yeah exactly. Agreed to enter an issue for chronicles lib. That can solve this issue in an optimal way.
Within this in mind, I would agree to keep it if you think so. I would have like to claim the attention on hidden costs of such calls.

gabrielmer: That's a good approach, let's check it directly with Chronicles. In the meantime, I'll leave the computation outside the log. It's a super simple/cheap one, so if we can avoid having them inside then better

Issue #2345, opened 2024-01-10
PoC: rln-v2 in waku


rymnc:
### Problem

We do not have arbitrary messaging rates with rln in waku, leading to only 1 message per epoch (set to 1 second)

### Suggested solution

We make use of rln-v2 instead of rln-v1, which brings variable messaging rates depending on the user's stake. 

Till we have a viable tokenomics model for rln-v2, we will just proceed with linearly scaling the stake with the messaging rate.

### Sub-issues for incremental changes

- [x] chore(rln-relay): remove websocket from OnchainGroupManager
  - [x] https://github.com/waku-org/nwaku/pull/2364
- [x] #2375
  - [x] https://github.com/waku-org/nwaku/pull/2381 
- [x] #2376
- [x] #2377
  - [x] https://github.com/waku-org/nwaku/pull/2385
- [x] #2378
  - [x] https://github.com/waku-org/nwaku/pull/2385 
  - the rate commitment now has to be calculated within nwaku: `poseidon([id_commitment, user_message_limit])`
- [x] rln-v2: Update command line arguments to rln-keystore-generator, which now accepts a message rate that the user registers themself with
  - [x] https://github.com/waku-org/nwaku/pull/2392 
- [x] #2415
  - [x] https://github.com/waku-org/nwaku/pull/2413
  - Additionally, the node must clear this nonce every epoch
- [x] #2379
- [x] Appropriate nullifierLog struct: https://github.com/waku-org/nwaku/pull/2508 
- [x] #2481

### Additional context

Zerokit release: https://github.com/vacp2p/zerokit/releases/tag/v0.4.1

### Acceptance criteria

- [ ] A user is able to register themself to the rln-v2 membership set and send multiple messages per epoch

rymnc:
cc: @jm-clius @fryorcraken @alrevuelta @kaiserd

fryorcraken:
> We make use of rln-v2 instead of rln-v1, which brings variable messaging rates depending on the user's stake.

> Till we have a viable tokenomics model for rln-v2, we will just proceed with linearly scaling the stake with the messaging rate.

Can you clarify what exact number we could look into?e,g, you lock 1DAI and you get 1000msgs/day, 2DAY you get 2000msgs/day? This is cool.

rymnc:
yes, something along those lines. however, a day is too large of an epoch size to use - consider the calculations below


```
users = 10_000
nullifier_size = 128 bytes (with metadata)
epoch size = 1 day
[...]

PR #2364, opened 2024-01-22
chore(rln-relay): remove websocket from OnchainGroupManager

rymnc:
# Description

Removes dependence on eth_subscribe since websocket connections can be flaky.

# Changes

- [x] Removes eth_subscribe call and replaces it with block polling every `1` second. default can be changed

## How to test

1. Pass in a http rpc provider instead of ws while running nwaku / `make -j16 test`
2. Generate an rln keystore using
  ```
  ./build/wakunode2 generateRlnKeystore \
    --rln-relay-cred-path:./rlnKeystore.json \
    --rln-relay-eth-client-address:https:<> \
    --rln-relay-cred-password:sup3rsecure \
    --rln-relay-eth-contract-address:0xF471d71E9b1455bBF4b85d475afb9BB0954A29c4 \
    --rln-relay-eth-private-key:<> \
    --execute
  ```
  insert your https rpc provider as well as private key in the required positions

## Issue
addresses part of #2345

### DIFF
--- a/waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
+++ b/waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
@@ -74,13 +74,17 @@ type
     # in event of a reorg. we store 5 in the buffer. Maybe need to revisit this,
     # because the average reorg depth is 1 to 2 blocks.
     validRootBuffer*: Deque[MerkleNode]
+    # interval loop to shut down gracefully
+    blockFetchingActive*: bool
 
 const DefaultKeyStorePath* = "rlnKeystore.json"
 const DefaultKeyStorePassword* = "password"
 
+const DefaultBlockPollRate* = 1.seconds
+
 template initializedGuard(g: OnchainGroupManager): untyped =
   if not g.initialized:
-    raise newException(ValueError, "OnchainGroupManager is not initialized")
+    raise newException(CatchableError, "OnchainGroupManager is not initialized")
 
 
 proc setMetadata*(g: OnchainGroupManager): RlnRelayResult[void] =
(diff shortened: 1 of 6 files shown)


waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
```diff
@@ -316,12 +320,15 @@ proc handleRemovedEvents(g: OnchainGroupManager, blockTable: BlockTable):
 
 proc getAndHandleEvents(g: OnchainGroupManager,
                         fromBlock: BlockNumber,
-                        toBlock: BlockNumber): Future[void] {.async: (raises: [Exception]).} =
+                        toBlock: BlockNumber): Future[bool] {.async: (raises: [Exception]).} =
```
gabrielmer, line 323: I think that this proc either returns true or raises an exception.

Am I missing something or is it supposed to be like that? If so, what's the point of returning `bool` if the return value is always true?

rymnc: that's for compatibility with the retryWrapper template, it expects a return value to assign the result of an async operation, just set to true here since we don't use it (discarded)

gabrielmer (approved):
LGTM! Thanks so much! 🤩

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
```diff
@@ -74,13 +74,17 @@ type
     # in event of a reorg. we store 5 in the buffer. Maybe need to revisit this,
     # because the average reorg depth is 1 to 2 blocks.
     validRootBuffer*: Deque[MerkleNode]
+    # interval loop to shut down gracefully
+    blockFetchingActive*: bool
 
 const DefaultKeyStorePath* = "rlnKeystore.json"
 const DefaultKeyStorePassword* = "password"
 
+const DefaultBlockPollRate* = 1.seconds
```
alrevuelta, line 83: Perhaps this is too often? Would suggest 6s (half-block time). Thinking here of people that may be using freemium providers. Once per second is 86.400 requests a day (and that just for knowing the current block).

On the other hand, since we consider as valid multiple roots, we could relax this further (thinking of >30sec). But lets leave this by now, as it may open up some edge cases. 6s should be ok by now.

alrevuelta (approved):
lgtm thanks! minor comment regarding `DefaultBlockPollRate`

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
rymnc: addressed in https://github.com/waku-org/nwaku/pull/2364/commits/dbd6cc3617634314295c11f4b19ab064de5eb25a

PR #2367, opened 2024-01-24
fix: adding rln validator as default

### LINKED ISSUE #2365: bug: RLN validator is only added for statically configured pubsub topics
### Problem

Currently, when RLN is enabled, RLN validator is added only for pubsub topics added at node initialization https://github.com/waku-org/nwaku/blob/8a9fad2905ef4b7415994ce89b2b4479bfbaf20c/waku/node/waku_node.nim#L983

For dynamically added topics, RLN validator is not set even if node is running with RLN enabled.

In go-waku, if rln is enabled the RLN validator is added to defaultTopicValidators which are set to any pubsubTopics subscribed after initialization:
- https://github.com/waku-org/go-waku/blob/190d8e8e0881bdcfba0d6fa220abfb05066b9f44/waku/v2/node/wakunode2_rln.go#L94
- https://github.com/waku-org/go-waku/blob/190d8e8e0881bdcfba0d6fa220abfb05066b9f44/waku/v2/protocol/relay/validators.go#L42

### Expected behavior
If RLN is enabled in the node, all the pubsub topics must run the RLN validator

gabrielmer:
# Description

It was noticed that RLN validator wasn't configured for dynamically subscribed pubsub topics.
Therefore, we decided to implement the same approach to validators as in go-waku: we're adding now support for default validators, which will be run for all pubsub topics. We're adding RLN validator to the default validators whenever RLN is enabled.

# Changes

- [x] added default validators, which run for all pubsub topics
- [x] included RLN validator to the default validators when RLN is mounted

## How to test

Simulated the scenario in which we send an invalid message on a dynamically added pubsub topics with and without this fix.

With this fix:
Screenshot 2024-01-26 at 17 42 07

Without it:
Screenshot 2024-01-26 at 17 42 46

Thanks @AlejandroCabeza for it!
Not adding the test in this PR as it is added in Alex's PR of the RLN test suite

Also, thanks @chaitanyaprem for helping from go-waku side to maintain consistency between both :)

## Issue

closes #2365

### DIFF
--- a/waku/waku_relay/protocol.nim
+++ b/waku/waku_relay/protocol.nim
@@ -130,6 +130,8 @@ type
     # a map that stores whether the ordered validator has been inserted
     # for a given PubsubTopic
     validatorInserted: Table[PubsubTopic, bool]
+    # seq of validators that are called for every pubsub topic
+    wakuDefaultValidators: seq[WakuValidatorHandler]
 
 proc initProtocolHandler(w: WakuRelay) =
   proc handler(conn: Connection, proto: string) {.async.} =
@@ -183,6 +185,10 @@ proc addValidator*(w: WakuRelay,
   for t in topic:
     w.wakuValidators.mgetOrPut(t, @[]).add(handler)
 
+proc addDefaultValidator*(w: WakuRelay,
+                   handler: WakuValidatorHandler) {.gcsafe.} =
+  w.wakuDefaultValidators.add(handler)
+
 method start*(w: WakuRelay) {.async.} =
   debug "start"
   await procCall GossipSub(w).start()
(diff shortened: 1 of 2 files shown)


chaitanyaprem (approved):
LGTM

rymnc (approved):
LGTM

waku/waku_relay/protocol.nim
```diff
@@ -130,6 +130,8 @@ type
     # a map that stores whether the ordered validator has been inserted
     # for a given PubsubTopic
     validatorInserted: Table[PubsubTopic, bool]
+    # seq of validators that are called for every pubsub topic
+    wakuDefaultValidators: seq[WakuValidatorHandler]
```
alrevuelta, line 134: not sure i understand this.
imho if rln is enabled, it should be enabled in all topics. same applies for all validators.

so not sure i understand the difference between validators and default validators. Having `addValidator` and `addDefaultValidator` creates some cognitive load for a use case that IMHO we don't have?

PR #2370, opened 2024-01-25
chore: adding nwaku compose image update to release process

gabrielmer:
Adding to the release process doc the step of updating nwaku-compose's image after release


NagyZoltanPeter (approved):
Right!

SionoiS (approved):
Nice! Makes a lot of sense.

Issue #2284, opened 2023-12-12
feat: REST API - large messages does not seem to be rejected by relay auto api


fryorcraken:
### Problem

Messages that are over 140kB does not seem to be rejected by the relay REST API

### Suggested solution

Reject messages at the API level if they do not match validity criteria (missing fields or size)

Issue #1700, opened 2023-04-24
feat: Pluggable validator/QoS interface


jm-clius:
### Problem

Waku currently implements and plans to implement various message "evaluation" mechanisms. Two examples of this would b:
- Application-level DoS protection via application signature validation: https://github.com/waku-org/nwaku/issues/1612
- Spam protection via RLN relay: https://rfc.vac.dev/spec/17/

Both these examples tie into the Waku Relay _validation_ mechanism and drops messages that violate some configured rule criteria within a Waku network, subnetwork or application. Although message validation (especially for DoS protection) is an obvious use case for evaluating messages against some criteria, we can imagine more general evaluation scenarios (such as determining a preferred QoS for a message) which wouldn't strictly be a simple validation. For this issue I'll just refer to "validation" as it matches what we have already.

From the examples it's also clear that message validation is configured somewhat separately from the protocol (Waku Relay in both cases) and even sometimes belong to the application. Implementing this in nwaku therefore requires introducing (sometimes complex) validator logic which could in theory be more cleanly maintained separately or application-level constructs which more appropriately belongs elsewhere. There are more good reasons to want to separate validation logic, including communicating about Waku's internal vs external DoS protection capabilities, allowing third parties to provide their own validation/QoS, etc.

### Suggested solution

Define a plugin interface and strategy for Waku validators. To keep things simple, this could for now be:
- a simple C API that takes a `WakuMessage` as input and returns a value between `0` and `1` indicating the underlying validator's evaluation of the message (`0` being lowest preferred QoS and `1` highest).
- a strategy to dynamically load such validation plugins into the nwaku node. As a start we could e.g. allow only a single `.so` file within a configured `lib/` subdirectory. This way we don't have to define strategy for loading multiple plugins and order of validation for now.
- tie this validation output to the Waku Relay validator. Simplifying again, for now, we could drop everything below `0.5` and relay every message with validation result `0.5` and above.
- both RLN-relay and the Status MVP DoS protection mechanism can be moved to such plugins.

cc @arnetheduck @fryorcraken @alrevuelta

rymnc:
We could probably look into using wasm too, in the future

jm-clius:
Another suggestion was to define this as a JSON-RPC API. This way we get the benefits of separation of application-level concerns as set out in the description, but the validation can also be run as a web service that can be provisioned (we can, for example, provide RLN-validation-as-a-service).

fryorcraken:
[...]

PR #2373, opened 2024-01-26
feat: running validators in REST API

gabrielmer:
# Description
We currently add message validators to libp2p's gossipsub, which are run before sending a message. In this PR, we're running the validators in  POST `/relay/v1/messages/{pubsubTopic}` REST handler in order to catch earlier any wrong message and be able to return the user an error in case it's invalid and the message is not published.

In addition to that, a check for messages bigger than the maximum configured size was added to the handler. That's also a check currently made at the libp2p layer before relaying the message, and we want to catch these messages earlier in the flow.

A big change of this PR is that we got rid of topic-specific validators. Now the same validators will run for every pubsub topic.

After completing this PR, a new PR will be opened implementing the same validation logic for other autosharding and lightpush's REST handlers.

# Changes

- [x] removed topic-specific validators and converted them to general validators
- [x] running validators in POST `/relay/v1/messages/{pubsubTopic}` REST handler
- [x] checking for max message size in the above REST handler
- [x] implemented test to check that the max size validator is run at the handler 

## Issue

Part of #2284

### DIFF
--- a/waku/node/waku_node.nim
+++ b/waku/node/waku_node.nim
@@ -397,7 +397,7 @@ proc mountRelay*(node: WakuNode,
     await node.startRelay()
 
   node.switch.mount(node.wakuRelay, protocolMatcher(WakuRelayCodec))
-  
+
   info "relay mounted successfully"
 
   # Subscribe to topics
--- a/waku/waku_api/rest/relay/handlers.nim
+++ b/waku/waku_api/rest/relay/handlers.nim
@@ -4,7 +4,7 @@ else:
   {.push raises: [].}
 
 import
-  std/[sequtils, strformat],
+  std/sequtils,
   stew/[byteutils, results],
   chronicles,
   json_serialization,
--- a/waku/waku_relay/protocol.nim
+++ b/waku/waku_relay/protocol.nim
@@ -196,7 +196,7 @@ proc isSubscribed*(w: WakuRelay, topic: PubsubTopic): bool =
 proc subscribedTopics*(w: WakuRelay): seq[PubsubTopic] =
   return toSeq(GossipSub(w).topics.keys())
 
-proc generateOrderedValidator*(w: WakuRelay): auto {.gcsafe.} =  
+proc generateOrderedValidator*(w: WakuRelay): auto {.gcsafe.} =
   # rejects messages that are not WakuMessage
   let wrappedValidator = proc(pubsubTopic: string,
                               message: messages.Message): Future[ValidationResult] {.async.} =


SionoiS (commented):
:+1:  This approach seams sensible. No strong opinion either way.

tests/waku_relay/test_protocol.nim
```diff
@@ -297,14 +297,14 @@ suite "Waku Relay":
       proc otherSimpleFutureHandler(topic: PubsubTopic, message: WakuMessage) {.async, gcsafe.} =
         otherHandlerFuture.complete((topic, message))
 
-      otherNode.addValidator(pubsubTopic, len4Validator)
+      otherNode.addValidator(pubsubTopic, len4Validator, "")
```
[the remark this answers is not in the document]
Ivansete-status: The issue apparently comes from how the `varargs[string]` is being passed. It compiles well with the following:
`otherNode.addValidator([pubsubTopic], len4Validator)`

tests/waku_relay/test_protocol.nim
```diff
@@ -297,14 +297,14 @@ suite "Waku Relay":
...
       discard otherNode.subscribe(pubsubTopic, otherSimpleFutureHandler)
       await sleepAsync(500.millis)
       check:
         otherNode.isSubscribed(pubsubTopic)
 
       # Given a subscribed node with a validator
-      node.addValidator(pubsubTopic, len4Validator)
+      node.addValidator(pubsubTopic, len4Validator, "")
```
Ivansete-status, line 307:
```suggestion
      node.addValidator([pubsubTopic], len4Validator)
```

tests/waku_relay/test_wakunode_relay.nim
```diff
@@ -163,7 +163,7 @@ suite "WakuNode - Relay":
       completionFutValidatorAcc.complete(true)
       return ValidationResult.Accept
 
-    node2.wakuRelay.addValidator(pubSubTopic, validator)
+    node2.wakuRelay.addValidator(pubSubTopic, validator, "")
```
Ivansete-status, line 166:
```suggestion
    node2.wakuRelay.addValidator([pubSubTopic], validator)
```

waku/node/waku_node.nim
```diff
@@ -344,6 +344,17 @@ proc publish*(
   
   return ok()
 
+proc messageSizeValidator(topic: string, message: WakuMessage): Future[pubsub.ValidationResult] {.async.} =
+  trace "message size validator is called"
+
+  let messageSizeBytes = uint64(message.encode().buffer.len)
+
+  if(messageSizeBytes > MaxWakuMessageSize):
+      trace fmt"Message size exceeded maximum of {DefaultMaxWakuMessageSizeStr}"
```
Ivansete-status, line 353: In the future we might want to use metrics instead so that we can get some stats on why messages are being rejected

waku/node/waku_node.nim
```diff
@@ -397,6 +408,9 @@ proc mountRelay*(node: WakuNode,
 
   node.switch.mount(node.wakuRelay, protocolMatcher(WakuRelayCodec))
 
+  node.wakuRelay.addDefaultValidator(messageSizeValidator,
```
Ivansete-status, line 411: Nitpick. I think is a bit confusing to use `addDefaultValidator`. IMO, when I read this code, I have the feeling that we will have only one default validator but we are keeping an array of default validators.

```suggestion
  node.wakuRelay.addValidator(messageSizeValidator,
```

For the sake of simplicity, would it be possible to stick only to the concept of the "validators" list?

waku/node/waku_node.nim
```diff
@@ -984,10 +998,10 @@ proc mountRlnRelay*(node: WakuNode,
   let rlnRelay = rlnRelayRes.get()
   let validator = generateRlnValidator(rlnRelay, spamHandler)
 
-  # register rln validator for all subscribed relay pubsub topics
-  for pubsubTopic in node.wakuRelay.subscribedTopics:
-    debug "Registering RLN validator for topic", pubsubTopic=pubsubTopic
-    node.wakuRelay.addValidator(pubsubTopic, validator)
+  # register rln validator as default validator
+  debug "Registering RLN validator"
+  node.wakuRelay.addDefaultValidator(validator, "RLN validation failed")
```
Ivansete-status, line 1003: Is this change out of the scope of the PR?

gabrielmer:
> 👍 This approach seams sensible. No strong opinion either way. Maybe `WakuMessage`s could check their own size?

Makes sense! Thought about it and found 2 issues to this.

First one, is that we don't use a `WakuMessage.init()` proc or something similar that runs logic when creating a WakuMessage, we just assign values to the struct as they are. We could create one, but not sure if refactoring it in the whole code only for this use case is worth it.

Apart from that, the easiest way to calculate the size in bytes of a `WakuMessage` is to create the struct and then check its size. We could calculate the size of all its separate fields and add them together, but that might be prone to bugs in case we add a field or make any change in the type.

LMK what you all think :))

waku/waku_relay/protocol.nim
```diff
@@ -218,6 +244,24 @@ proc generateOrderedValidator*(w: WakuRelay): auto {.gcsafe.} =
     return ValidationResult.Accept
   return wrappedValidator
 
+proc validateMessage*(w: WakuRelay, pubsubTopic: string, msg: WakuMessage):
+  Future[Result[void, string]] {.async.} =
+    
+    for validator in w.wakuDefaultValidators:
+        let validatorRes = await validator(pubsubTopic, msg)
```
Ivansete-status, line 251: Sorry but I can't quite understand why we need the "default validators" concept because we end up validating the message per its `pubsubTopic`

Ivansete-status (commented):
Thanks for the PR!
I've added a few comments. If possible, I'd try to simplify the approach and just have one kind of validator (avoiding the use of the generic/default validators list) but ofc I might be missing something :)

SionoiS:
> Apart from that, the easiest way to calculate the size in bytes of a `WakuMessage` is to create the struct and then check its size. We could calculate the size of all its separate fields and add them together, but that might be prone to bugs in case we add a field or make any change in the type.

What I was thinking is something like
```
if not myMessage.isWakuValid():
   return err("Max size 150KB! This message is too large per Waku RFC 64.")

if not myMessage.isGossipSubValid():
   return err("Max size 1MiB! This message is too large for GossipSub.")
```

Plus some tests.

tests/waku_relay/test_protocol.nim
gabrielmer: That's amazing! Thank you! 😍

waku/node/waku_node.nim
gabrielmer: Makes sense! Not sure I understand instead of what though. Wouldn't it just be adding a line of code updating the metrics?

waku/node/waku_node.nim
gabrielmer: So we already have the `addValidator` method which is for pubsub topic specific validators.

https://github.com/waku-org/nwaku/blob/c1121dd1db27344ad38d6ba85c072f0e75eb4e8e/waku/waku_relay/protocol.nim#L182-L186

The concept of the default validator list comes from go-waku, which are the validators that are supposed to run for all pubsub topics. So we need to save in different data structures the validators per pubsub topic and the validators that run for all topics

waku/node/waku_node.nim
Ivansete-status: > Makes sense! Not sure I understand instead of what though. Wouldn't it just be adding a line of code updating the metrics?

Yes exactly :) I was thinking of having something similar to the following, where we can count the number of errors by its type:

https://github.com/waku-org/nwaku/blob/c1121dd1db27344ad38d6ba85c072f0e75eb4e8e/waku/node/peer_manager/peer_manager.nim#L109

To be considered as a future enhancement :)

waku/node/waku_node.nim
gabrielmer: Yes, this is out of scope of this PR. This was already merged. I basically had 2 PRs, one adding the functionality of default validators and using it for RLN, and this one using the validators to check messages in the REST handlers.

So this PR came on top of what was built on the other one. Probably it would have made more sense to open this PR once the other one was merged. Because the other one was small, I had both open in parallel to make things faster, but might have been better to wait for the other one to be merged before opening this one.

waku/waku_relay/protocol.nim
gabrielmer: My reasoning for this was first, to have exactly the same type of validator in the pubsub specific list and in the default validators list, using in both cases the type `WakuValidatorHandler`.

We could create a type `WakuDefaultValidatorHandler` without the pubsub topic parameter.
But I thought it's even better to have the pubsub topic as an argument. Notice that a default validator in this case is a validator that runs for all pubsub topics but doesn't need to behave the same way for all the topics. And that's the advantage of having the pubsub parameter in this case, it adds more possibilities to what we can (and might want to) do.

That's the reasoning behind it, not sure if it makes sense or if it just overcomplicates things

gabrielmer:
> > Apart from that, the easiest way to calculate the size in bytes of a `WakuMessage` is to create the struct and then check its size. We could calculate the size of all its separate fields and add them together, but that might be prone to bugs in case we add a field or make any change in the type.
> 
> What I was thinking is something like
> 
> ```
> if not myMessage.isWakuValid():
>    return err("Max size 150KB! This message is too large per Waku RFC 64.")
> 
> if not myMessage.isGossipSubValid():
>    return err("Max size 1MiB! This message is too large for GossipSub.")
> ```
> 
> Plus some tests.

Oooh got it! So basically having the check inside a proc. That's cool, 100% can be useful and nicer :)

NagyZoltanPeter (approved):
Overall thank you for this great stuff. I agree with the comments added before me, but I can approve it from my side, bet you will answer the rest. ;-)

waku/waku_relay/protocol.nim
alrevuelta: > Notice that a default validator in this case is a validator that runs for all pubsub topics but doesn't need to behave the same way for all the topics

based on this i understand that the goal behind this "default validator" is to allow having different types of validators, the ones affecting to all pubusb topics, and the ones affecting to specific pubsub topics?

If that's the case, I would say there is no need to have this. A given validator shall be applied to all pubsub topics.

waku/waku_api/rest/relay/handlers.nim
```diff
@@ -135,16 +135,8 @@ proc installRelayApiHandlers*(router: var RestRouter, node: WakuNode, cache: Mes
...
-      let result = node.wakuRlnRelay.validateMessageAndUpdateLog(message)
-      if result == MessageValidationResult.Invalid:
-        return RestApiResponse.internalServerError("Failed to publish: invalid RLN proof")
-      elif result == MessageValidationResult.Spam:
-        return RestApiResponse.badRequest("Failed to publish: limit exceeded, try again later")
-      elif result == MessageValidationResult.Valid:
-        debug "RLN proof validated successfully", pubSubTopic=pubSubTopic
-      else:
-        return RestApiResponse.internalServerError("Failed to publish: unknown RLN proof validation result")
+      (await node.wakuRelay.validateMessage(pubsubTopic, message)).isOkOr:
```
alrevuelta, line 138: imho you should check also the result. Its not only about the result being Ok, its also about being `Valid`. If this returns `Spam` or `Invalid` we should error.  (see code that was removed)

waku/waku_relay/protocol.nim
```diff
@@ -130,6 +130,10 @@ type
     # a map that stores whether the ordered validator has been inserted
     # for a given PubsubTopic
     validatorInserted: Table[PubsubTopic, bool]
+    # seq of validators that are called for every pubsub topic
+    wakuDefaultValidators: seq[WakuValidatorHandler]
+    # a map of validators to error messages to return when validation fails
+    wakuValidatorsErrors: Table[WakuValidatorHandler, string]
```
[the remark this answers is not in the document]
gabrielmer: Great idea! I also think it's better :)) Will change it

waku/node/waku_node.nim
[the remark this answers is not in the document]
gabrielmer: 100%

waku/waku_api/rest/relay/handlers.nim
gabrielmer: `validateMessage()` returns `ok()` if the message is validated by all the subscribed validators, and `err("error message")` if the message is not validated. The result doesn't contain anything, its return type is `Result[void, string]`.

The generic validators can't return the results `Spam` or `Invalid`, their return type is `ValidationResult` which is used by libp2p (as the validators are also passed to libp2p layer), so it does not have the RLN specific return values.

This is one of the big tradeoffs of this approach using generic validators instead of hardcoded checks in the REST handlers. Wrote about this on the following comment: https://github.com/waku-org/nwaku/pull/2373/files#r1467933272

Not sure what's the right thing to do. But if we want to run our registered validators in the REST handler to notify the user if there's a validation error, and the RLN validator is part of our validator list (so it will be run), then the best I think we can do is logging the specific result and return the values that they allow. LMK if I'm missing something or if doesn't seem worth it

alrevuelta: right, apologies I overlooked it.

waku/waku_relay/protocol.nim
alrevuelta: mm I mean every validator is per topic. But my point is that we should enforce that whatever amount of validators are set, they are all applied to all subscribed topics.

I don't see a reason for having topica->validator_a, topicb->validator_b.

SionoiS:
@gabrielmer you are aware of https://github.com/waku-org/nwaku/issues/1700 right?

waku/waku_relay/protocol.nim
alrevuelta: > You mean that the same set of validators should run for all pubsub topics, meaning that each validator should be a "big proc" that handles the cases of all pubsub topics instead of validators being "small procs" specific to how to deal with a specific pubsub topic?

My point is that I don't see the need of having validators and default validators. Since IMHO every gossipsub topic that you are subscribed to, should apply the same set of validators.

> meaning that each validator should be a "big proc"

Note that from libp2p perspective, this is what we are doing. In nim-libp2p the order of execution of the validators is not deterministic (all of them are spawn async). So in waku we wrap them all into a single one, and we add that single one to nim-libp2p. In our case we do this because some validators consume more resources (eg rln one) so we don't want to execute that validator if previous one fail (eg decode into a WakuMessage).

gabrielmer: Got it! That makes lots of sense, didn't take into account that last point. So yes, no need of it. Thanks so much! 🤩

gabrielmer:
> @gabrielmer you are aware of #1700 right?

Ooooh, was definitely not aware of that GitHub issue! I was aware of the overall idea, that's why I took this approach here with validators instead of hardcoding the checks.

Thanks!

gabrielmer:
Changing PR to draft as there's still many changes to be done

waku/waku_relay/protocol.nim
```diff
@@ -216,19 +209,38 @@ proc generateOrderedValidator*(w: WakuRelay): auto {.gcsafe.} =
...
-
-    if w.wakuValidators.hasKey(pubsubTopic):
-      for validator in w.wakuValidators[pubsubTopic]:
-        let validatorRes = await validator(pubsubTopic, msg)
-        if validatorRes != ValidationResult.Accept:
-          return validatorRes
     return ValidationResult.Accept
   return wrappedValidator
 
+proc isValidSize(message: WakuMessage): Future[Result[void, string]] {.async.} =
```
gabrielmer, line 219: I'll also add to this proc a check that the message size is smaller than GossipSub's max.

Currently the max size is defined in `nim-libp2p` here: https://github.com/status-im/nim-libp2p/blob/e3c967ad1939fb33b8e13759037d193734acd202/libp2p/protocols/pubsub/pubsub.nim#L556

I don't want to have it hardcoded, so I'll first open a PR in `nim-libp2p` defining it as a global `const` and will then add an `if` condition in this proc

gabrielmer:
I think I implemented all the feedback received, please let me know if I missed anything or if there's anything else we can improve :)

So this PR doesn't get way too big, I'll implement the same validation logic in a separate PR for lightpush's and autosharding's REST handlers.

SionoiS (approved):
LGTM thanks!

apps/wakunode2/wakunode2_validator_signed.nim
```diff
@@ -49,21 +50,28 @@ proc withinTimeWindow*(msg: WakuMessage): bool =
     return true
   return false
 
-proc addSignedTopicValidator*(w: WakuRelay, topic: PubsubTopic, publicTopicKey: SkPublicKey) =
-  debug "adding validator to signed topic", topic=topic, publicTopicKey=publicTopicKey
+proc addSignedTopicsValidator*(w: WakuRelay, protectedTopics: seq[ProtectedTopic]) =
+  debug "adding validator to signed topics"
```
Ivansete-status, line 54: Not a big deal but we are losing a bit of detail because we stopped logging the actual topic

waku/waku_api/rest/relay/handlers.nim
```diff
@@ -135,16 +135,8 @@ proc installRelayApiHandlers*(router: var RestRouter, node: WakuNode, cache: Mes
       if not success:
         return RestApiResponse.internalServerError("Failed to publish: error appending RLN proof to message")
 
-      # validate the message before sending it
-      let result = node.wakuRlnRelay.validateMessageAndUpdateLog(message)
```
Ivansete-status, line 139: Why are we removing this `wakuRlnRelay` validation? This is quite important :) Or are we making this validation in another place?

Ivansete-status (commented):
Looks great indeed! However, I can't approve yet as per the doubt I have re `rest/relay/handlers.nim` :)

apps/wakunode2/wakunode2_validator_signed.nim
gabrielmer: The issue here is that we changed the function so instead of adding one validator for a signed topic, it adds one validator function that works for all signed topics. This comes from the new approach of having the same validators running for all topics.

So I got rid of the topics parameter in the log because it's now a seq of structs and didn't know if it was worth it/spammy to print the whole sequence. I can 100% add it if useful, but the change comes that before it was easier/nicer because there was only one topic. Now it's a whole seq of structs.

waku/waku_api/rest/relay/handlers.nim
gabrielmer: Yes!

So now we register the validator here:

https://github.com/waku-org/nwaku/blob/6171f31e896ce3a8aecaaa5f3ad382866e720313/waku/node/waku_node.nim#L990

And this runs in the REST handler here:
https://github.com/waku-org/nwaku/blob/6171f31e896ce3a8aecaaa5f3ad382866e720313/waku/waku_api/rest/relay/handlers.nim#L138

So instead of hardcoding again the RLN validation in the handler, this gets run when going over all the registered validators

Ivansete-status: Beautiful!, I see that the validation is performed in the following place:
https://github.com/waku-org/nwaku/blob/3e65cc18f64c7b8fa0920026e64dbb9507ab0c2e/waku/waku_rln_relay/rln_relay.nim#L331

Ivansete-status (approved):
LGTM! Thanks for the master-piece PR and the patience! :partying_face:

apps/wakunode2/wakunode2_validator_signed.nim
Ivansete-status: Yes makes sense, I think we can leave it as is. On the other hand, I think the "signed topics" come from a parameter and thus we could determine them if needed, AFAIK.

PR #2384, opened 2024-01-29
feat: supporting meta field in WakuMessage

### LINKED ISSUE #2214: bug: relay publish fails with 400 Bad Request when message contains meta field
Sending this curl on wakuorg/nwaku:latest:
`curl -vv -X POST "http://localhost:43393/relay/v1/messages/%2Fwaku%2F2%2Fdefault-waku%2Fproto" -H "accept: text/plain" -H "content-type: application/json" -d '{"payload": "SGVsbG8gV29ybGQh", "contentTopic": "/test/1/waku-relay/proto", "timestamp": 1699891642113786880, "meta": "dGVzdA=="}'`
yields 400 Bad Request 

On the logs side I can see 
```
DBG 2023-11-13 16:09:39.667+00:00 Serving API request                        tid=1 file=serverprivate.nim:186 peer=172.17.0.1:45944 meth=POST uri=/relay/v1/messages/%2Fwaku%2F2%2Fdefault-waku%2Fproto path_params="[(pubsubtopic, /waku/2/default-waku/proto)]" query_params=[] content_body="(application/json, 129 bytes)"
DBG 2023-11-13 16:09:39.667+00:00 JSON field not recognized by the current version of nwaku. Consider upgrading tid=1 file=types.nim:99 fieldName=meta typeName=None
DBG 2023-11-13 16:09:39.667+00:00 Received error response from handler       tid=1 file=serverprivate.nim:307 status=400 meth=POST peer=172.17.0.1:45944 uri=/relay/v1/messages/%2Fwaku%2F2%2Fdefault-waku%2Fproto error_status="400 Bad Request" error_content_type=text/plain error_message=
```

Notes!!
- On gowaku this works fine
- Using RPC instead of REST will return 200 and it will publish the message. However the message reaches peers without the meta field.

gabrielmer:
# Description

The "meta" field is late addition to the Waku Message specification. It is currently supported in go-waku and not in nwaku. We're implementing its support in this PR.

# Changes

- [x] adding `meta` field to Relay, Filter and Store Waku Messages
- [x] testing for correct returned fields in all 3 types of messages
- [x] fixed validator test that timed out early 

## Issue

closes #2214 

## Note
We also wanted to avoid responding with an error when an extra field is sent in a POST request.
There seems to be an issue in the `nim-json-serialization` dependency that throws an exception when this happens. Will report and track the issue to make sure it gets fixed

### DIFF
--- a/tests/wakunode2/test_validators.nim
+++ b/tests/wakunode2/test_validators.nim
@@ -195,7 +195,7 @@ suite "WakuNode2 - Validators":
     var msgRejected = 0
     
     # Active wait for the messages to be delivered across the mesh
-    for i in 0..<10:
+    for i in 0..<100:
       msgRejected = 0
       for i in 0..<5:
         for k, v in nodes[i].wakuRelay.peerStats.mpairs:
@@ -204,7 +204,7 @@ suite "WakuNode2 - Validators":
       if msgReceived == 125 and msgRejected == 500:
         break
       else:
-        await sleepAsync(1.seconds)
+        await sleepAsync(100.milliseconds)
 
     check:
       msgReceived == 125
--- a/tests/wakunode2/test_validators.nim
+++ b/tests/wakunode2/test_validators.nim
@@ -188,7 +188,7 @@ suite "WakuNode2 - Validators":
         discard await nodes[i].publish(some(spamProtectedTopic), unsignedMessage)
 
     # Wait for gossip
-    await sleepAsync(4.seconds)
+    await sleepAsync(6.seconds)
 
     # Since we have a full mesh with 5 nodes and each one publishes 25+25+25+25+25 msgs
     # there are 625 messages being sent.
--- a/tests/wakunode2/test_validators.nim
+++ b/tests/wakunode2/test_validators.nim
@@ -187,23 +187,28 @@ suite "WakuNode2 - Validators":
           version: 2, timestamp: afterTimestamp, ephemeral: true)
         discard await nodes[i].publish(some(spamProtectedTopic), unsignedMessage)
 
-    # Wait for gossip
-    await sleepAsync(6.seconds)
-
     # Since we have a full mesh with 5 nodes and each one publishes 25+25+25+25+25 msgs
     # there are 625 messages being sent.
     # 125 are received ok in the handler (first hop)
     # 500 are are wrong so rejected (rejected not relayed)
+    
+    var msgRejected = 0
+    
+    # Active wait for the messages to be delivered across the mesh
+    for i in 0..<10:
+      msgRejected = 0
+      for i in 0..<5:
+        for k, v in nodes[i].wakuRelay.peerStats.mpairs:
+          msgRejected += v.topicInfos[spamProtectedTopic].invalidMessageDeliveries.int
+      
+      if msgReceived == 125 and msgRejected == 500:
+        break
+      else:
+        await sleepAsync(1.seconds)
+
     check:
       msgReceived == 125
-
-    var msgRejected = 0
-    for i in 0..<5:
-      for k, v in nodes[i].wakuRelay.peerStats.mpairs:
-        msgRejected += v.topicInfos[spamProtectedTopic].invalidMessageDeliveries.int
-
-    check:
-      msgRejected == 500
+      msgRejected == 500      
 
     await allFutures(nodes.mapIt(it.stop()))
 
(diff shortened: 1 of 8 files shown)


tests/wakunode2/test_validators.nim
```diff
@@ -187,23 +187,28 @@ suite "WakuNode2 - Validators":
...
+    for i in 0..<10:
+      msgRejected = 0
+      for i in 0..<5:
+        for k, v in nodes[i].wakuRelay.peerStats.mpairs:
+          msgRejected += v.topicInfos[spamProtectedTopic].invalidMessageDeliveries.int
+      
+      if msgReceived == 125 and msgRejected == 500:
+        break
+      else:
+        await sleepAsync(1.seconds)
```
Ivansete-status, line 207: This is awesome! I suggest to change a little the wait to break sooner

```suggestion
    for i in 0..<100:
      msgRejected = 0
      for i in 0..<5:
        for k, v in nodes[i].wakuRelay.peerStats.mpairs:
          msgRejected += v.topicInfos[spamProtectedTopic].invalidMessageDeliveries.int
      
      if msgReceived == 125 and msgRejected == 500:
        break
      else:
        await sleepAsync(100.milliseconds)
```

Ivansete-status (commented):
It looks great so far. Good work! 
I just added a minor comment

tests/wakunode2/test_validators.nim
gabrielmer: Done! Thank you!

SionoiS (approved):
LGTM!

What happens with the messages in DB do they have `meta` already but nwaku was discarding the field?

gabrielmer:
> LGTM!
> 
> What happens with the messages in DB do they have `meta` already but nwaku was discarding the field?

Great question! So it seems that `meta` was already defined in the `WakuMessage` type https://github.com/waku-org/nwaku/blob/849d76d6d479eddbe273f8b9f713daa2a6a2eced/waku/waku_core/message/message.nim#L20-L38

And this is the one being used in the db

https://github.com/waku-org/nwaku/blob/849d76d6d479eddbe273f8b9f713daa2a6a2eced/waku/waku_store/common.nim#L61-L63

But apparently the field wasn't getting filled when sending neither relay, filter nor store messages

SionoiS:
> But apparently the field wasn't getting filled when sending neither relay, filter nor store messages

Good to know! but...

Let me rephrase my question. Are messages stored in Postgres as blobs? If not, are the messages previously created by nwaku show no value in the DB for that field? Lastly, is this a problem or not?

gabrielmer:
> Good to know! but...
> 
> Let me rephrase my question. Are messages stored in Postgres as blobs? If not, are the messages previously created by nwaku show no value in the DB for that field? Lastly, is this a problem or not?

Ooh, I think there's actually an issue with that. I think it's not getting saved in the DB at all 😶 
It doesn't seem to be in the schema.
https://github.com/waku-org/nwaku/blob/849d76d6d479eddbe273f8b9f713daa2a6a2eced/waku/waku_archive/driver/postgres_driver/postgres_driver.nim#L30-L41

What I don't understand is it working on the Store test then. Don't we use any DB in those tests? @Ivansete-status

Ivansete-status:
> What I don't understand is it working on the Store test then. Don't we use any DB in those tests? @Ivansete-status

Yup! We use a database in _Postgres_ tests.

Ivansete-status (approved):
LGTM! Thanks!

I think we can merge this PR, stating clearly that is missing the proper storage of this field, and submit another PR in the future where we start storing the  `meta` field in the database.

gabrielmer:
> 
> Yup! We use a database in _Postgres_ tests.

So in https://github.com/waku-org/nwaku/blob/0dac9f9dd49e2047f3988f4cf0332712028efd0e/tests/wakunode_rest/test_rest_store.nim, are store messages saved in memory?

If so, is it done different in a way only for testing? There's something I don't quite understand - I thought store messages were always retrieved from a db

Ivansete-status:
> So in https://github.com/waku-org/nwaku/blob/0dac9f9dd49e2047f3988f4cf0332712028efd0e/tests/wakunode_rest/test_rest_store.nim, are store messages saved in memory?

Yes, this particular test module stores messages in memory.

----

> If so, is it done different in a way only for testing? There's something I don't quite understand - I thought store messages were always retrieved from a db

That represents a test gap that should be filled with a `test_rest_store_postgres.nim` module.

These are the test modules that interact with a local _Postgres_ database:
https://github.com/waku-org/nwaku/blob/master/tests/waku_archive/test_driver_postgres_query.nim
https://github.com/waku-org/nwaku/blob/master/tests/waku_archive/test_driver_postgres.nim
... and notice that these are compiled in the following location: https://github.com/waku-org/nwaku/blob/0dac9f9dd49e2047f3988f4cf0332712028efd0e/tests/all_tests_waku.nim#L26-L28

gabrielmer:
> 
> I think we can merge this PR, stating clearly that is missing the proper storage of this field, and submit another PR in the future where we start storing the `meta` field in the database.

Super, sounds good! Will merge it and implement in a different PR support for db storage of the `meta` field :)

PR #2385, opened 2024-01-31
feat(rln-relay-v2): update C FFI api's and serde

### LINKED ISSUE #2377: rln-v2: Update C FFI api's

rymnc:
- feat(rln-relay-v2): integrate new ffi bindings, serde
- chore: use ExtendedRateLimitProof to describe epoch + rlnIdentifier

# Description

This PR introduces the changes required for serde, and creates a new type ExtendedRateLimitProof which contains all metadata required to verify a proof.

# Changes

- [x] updates the c ffi api
- [x] updates serde to include userMessageLimit and messageId
- [x] new type `ExtendedRateLimitProof` which includes `epoch` and `rlnIdentifier` in calculation of the `externalNullifier`

## Issue

closes #2377 and #2378

### DIFF
--- a/waku/waku_rln_relay/constants.nim
+++ b/waku/waku_rln_relay/constants.nim
@@ -18,6 +18,7 @@ const
 # inputs of the membership contract constructor
 # TODO may be able to make these constants private and put them inside the waku_rln_relay_utils
 const
+  # in wei
   MembershipFee* = 0.u256
   #  the current implementation of the rln lib supports a circuit for Merkle tree with depth 20
   MerkleTreeDepth* = 20
--- a/waku/waku_rln_relay/constants.nim
+++ b/waku/waku_rln_relay/constants.nim
@@ -2,6 +2,9 @@ import
   stint
 
 import
+  ./protocol_types
+
+import
   ../waku_keystore
 
 
@@ -15,7 +18,7 @@ const
 # inputs of the membership contract constructor
 # TODO may be able to make these constants private and put them inside the waku_rln_relay_utils
 const
-  MembershipFee* = 1000000000000000.u256
+  MembershipFee* = 0.u256
   #  the current implementation of the rln lib supports a circuit for Merkle tree with depth 20
   MerkleTreeDepth* = 20
   EthClient* = "http://127.0.0.1:8540"
@@ -28,6 +31,14 @@ const
 
 const
   DefaultRlnTreePath* = "rln_tree.db"
+
+when defined(rln_v2):
+  const
+    # pre-processed "rln/waku-rln-relay/v2.0.0" to array[32, byte]
+    DefaultRlnIdentifier*: RlnIdentifier = [114, 108, 110, 47, 119, 97, 107, 117, 
+                                            45, 114, 108, 110, 45, 114, 101, 108, 
+                                            97, 121, 47, 118, 50, 46, 48, 46, 
+                                            48, 0, 0, 0, 0, 0, 0, 0]
 
 # temporary variables to test waku-rln-relay performance in the static group mode
 const
--- a/waku/waku_rln_relay/conversion_utils.nim
+++ b/waku/waku_rln_relay/conversion_utils.nim
@@ -53,18 +53,36 @@ proc encodeLengthPrefix*(input: openArray[byte]): seq[byte] =
     output = concat(@len, @input)
   return output
 
-proc serialize*(idSecretHash: IdentitySecretHash, 
+when defined(rln_v2):
+  proc serialize*(idSecretHash: IdentitySecretHash, 
+                  memIndex: MembershipIndex, 
+                  userMessageLimit: UserMessageLimit,
+                  messageId: MessageId,
+                  externalNullifier: ExternalNullifier,
+                  msg: openArray[byte]): seq[byte] =
+    ## a private proc to convert RateLimitProof and the data to a byte seq
+    ## this conversion is used in the proofGen proc
+    ## the serialization is done as instructed in  https://github.com/kilic/rln/blob/7ac74183f8b69b399e3bc96c1ae8ab61c026dc43/src/public.rs#L146
+    ## [ id_key<32> | id_index<8> | epoch<32> | signal_len<8> | signal<var> ]
+    let memIndexBytes = toBytes(uint64(memIndex), Endianness.littleEndian)
+    let userMessageLimitBytes = toBytes(uint64(userMessageLimit), Endianness.littleEndian)
+    let messageIdBytes = toBytes(uint64(messageId), Endianness.littleEndian)
+    let lenPrefMsg = encodeLengthPrefix(msg)
+    let output = concat(@idSecretHash, @memIndexBytes, @userMessageLimitBytes, @messageIdBytes, @externalNullifier, lenPrefMsg)
+    return output
+else:
+  proc serialize*(idSecretHash: IdentitySecretHash, 
                 memIndex: MembershipIndex, 
                 epoch: Epoch,
                 msg: openArray[byte]): seq[byte] =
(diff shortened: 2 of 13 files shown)


waku/waku_rln_relay/constants.nim
```diff
@@ -29,6 +32,14 @@ const
 const
   DefaultRlnTreePath* = "rln_tree.db"
 
+when defined(rln_v2):
+  const
+    # pre-processed "rln/waku-rln-relay/v2.0.0" to array[32, byte]
+    DefaultRlnIdentifier*: RlnIdentifier = [114, 108, 110, 47, 119, 97, 107, 117,
```
rymnc, line 38: imo this is the best way to have the DefaultRlnIdentifier stored - if we do decide to have it as a command line argument for rln-relay we can process it into an array of 32 bytes at that point.

waku/waku_rln_relay/conversion_utils.nim
```diff
@@ -53,33 +53,60 @@ proc encodeLengthPrefix*(input: openArray[byte]): seq[byte] =
     output = concat(@len, @input)
   return output
 
-proc serialize*(idSecretHash: IdentitySecretHash, 
+when defined(rln_v2):
```
rymnc, line 56: had to dupe the function because args are different

waku/waku_rln_relay/protocol_types.nim
```diff
@@ -49,8 +53,21 @@ type RateLimitProof* = object
...
+    epoch*: Epoch
+    ## Application specific RLN Identifier
+    rlnIdentifier*: RlnIdentifier
+
+when defined(rln_v2):
+  type ExtendedRateLimitProof* = ref object of RateLimitProof
+    ## epoch is the epoch for which the proof is generated
+    epoch*: Epoch
+    ## rlnIdentifier is the identifier of the RLN application
+    rlnIdentifier*: RlnIdentifier
```
rymnc, line 70: The distinction between the two is made because now, `RateLimitProof` contains the proof with the public inputs/outputs. epoch and rlnIdentifier are no longer public inputs, and must not be confused this way. `ExtendedRateLimitProof` is what should be sent on the wire.

waku/waku_rln_relay/protocol_types.nim
```diff
@@ -49,8 +53,21 @@ type RateLimitProof* = object
...
+  else:
+    ## the epoch used for the generation of the `proof`
+    epoch*: Epoch
+    ## Application specific RLN Identifier
+    rlnIdentifier*: RlnIdentifier
+
+when defined(rln_v2):
+  type ExtendedRateLimitProof* = ref object of RateLimitProof
+    ## epoch is the epoch for which the proof is generated
+    epoch*: Epoch
```
alrevuelta, line 68: why is the `epoch` inside `ExtendedRateLimitProof` instead of `RateLimitProof`. I mean, both rln v1 and v2 should have this field, so? same for `rlnIdentifier`?

waku/waku_rln_relay/conversion_utils.nim
alrevuelta: have the hunch that having rlnv2 behind a flag will end up being more problematic than using a branch (where we completely replace v1 with v2)

alrevuelta (approved):
lgtm

waku/waku_rln_relay/protocol_types.nim
rymnc: the difference between rln-v1 and v2 also includes that the epoch and rlnIdentifier is not passed into the circuit, instead the externalNullifier is directly passed in after computing it from the epoch and rlnIdentifier. the RateLimitProof struct properly defines this behaviour, and ExtendedRateLimitProof includes them so that a receiving node can ensure the epoch is not too old, and that the rlnIdentifier is the same as they one they're validating for.

waku/waku_rln_relay/protocol_types.nim
```diff
@@ -67,54 +84,105 @@ type
...
-  output.write3(7, nsp.rlnIdentifier)
+    var rlnIdentifier: seq[byte]
+    discard ? pb.getField(7, rlnIdentifier)
+    discard nsp.rlnIdentifier.copyFrom(rlnIdentifier)
 
-  output.finish3()
+    return ok(nsp)
 
-  return output
+when defined(rln_v2):
```
Ivansete-status, line 158: Would it be possible to only have one `encode` implementation if we only had `RateLimitProof`?

waku/waku_rln_relay/protocol_types.nim
```diff
@@ -49,8 +53,21 @@ type RateLimitProof* = object
   ## nullifier enables linking two messages published during the same epoch
   ## see details in https://hackmd.io/tMTLMYmTR5eynw2lwK9n1w?view#Nullifiers
   nullifier*: Nullifier
-  ## Application specific RLN Identifier
-  rlnIdentifier*: RlnIdentifier
+  when defined(rln_v2):
+    ## the external nullifier used for the generation of the `proof` (derived from poseidon([epoch, rln_identifier]))
+    externalNullifier*: ExternalNullifier
```
Ivansete-status, line 58: Is is possible to use a different attribute name? `externalNullifier` is also part of the `ProofMetadata` and it is more difficult to follow its references.

waku/waku_rln_relay/rln/rln_interface.nim
```diff
@@ -101,43 +101,43 @@ proc seeded_key_gen*(ctx: ptr RLN, input_buffer: ptr Buffer, output_buffer: ptr
...
 ## the verification of the zk proof is available in proof_is_valid_ptr, where a value of true indicates success and false a failure
 
 proc zk_prove*(ctx: ptr RLN,
-            input_buffer: ptr Buffer,
-            output_buffer: ptr Buffer): bool {.importc: "prove".}
+               input_buffer: ptr Buffer,
+               output_buffer: ptr Buffer): bool {.importc: "prove".}
 ## Computes the zkSNARK proof and stores it in output_buffer for input values stored in input_buffer
-## input_buffer is serialized as input_data as [ id_key<32> | path_elements<Vec<32>> | identity_path_index<Vec<1>> | x<32> | epoch<32> | rln_identifier<32> ]
+## input_buffer is serialized as input_data as [ identity_secret<32> | user_message_limit<32> | message_id<32> | path_elements<Vec<32>> | identity_path_index<Vec<1>> | x<32> | external_nullifier<32> ]
```
Ivansete-status, line 131: We are changing the param definition but this actually depends on the `zerokit/rln` implementation being used, right?

Ivansete-status (commented):
Thanks for this! I've added a few comments/questions that I hope you find useful :)

waku/waku_rln_relay/protocol_types.nim
rymnc: it's the same field/value though

waku/waku_rln_relay/protocol_types.nim
rymnc: addressed in 561a967e

waku/waku_rln_relay/rln/rln_interface.nim
rymnc: addressed in 561a967e

waku/waku_rln_relay/protocol_types.nim
```diff
@@ -31,16 +31,20 @@ type
...
 
+when defined(rln_v2):
+  type
+    UserMessageLimit* = uint64
+    MessageId* = uint64
+    ExternalNullifier* = array[32, byte]
+
 # Custom data types defined for waku rln relay -------------------------
-type RateLimitProof* = object
+type RateLimitProof* = object of RootObj
```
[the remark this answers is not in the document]
rymnc: addressed in 561a967e

PR #2387, opened 2024-01-31
chore: REST store - get msgs from self node when store is mounted and no peerAddr is passed

Ivansete-status:
## Description

A node that handles REST-Store requests it normally acted as a Store-client and therefore it retrieved the messages from another Store-node.
With these changes, we allow a node that has Store mounted, to retrieve its own messages. In other words, the node can act as a Store-server of its own messages.

# Changes

* test_rest_store.nim: add new test to validate that the self-node can retrieve its own messages to the REST client.

* rest/store/client.nim: add new proc to allow making a GET store request without peerAddr.

* rest/store/handle.nim: add logic to handle requests that doesn't provide peerAddr but the self/local node has Store mounted. In this case the self/local node will retrieve its local stored messages.

* waku_store/self_req_handler.nim: logic to handle "store" requests allowing the REST-store node to act as a Store-server node itself. The 'self_req_handler.nim' helps to bypass the store protocol and directly retrieve the messages from the local/self node. I added this logic in a separate file from 'protocol.nim' because it doesn't participate in any libp2p comunication.

* waku_store/protocol.nim: make 'queryHandler' attribute public so that it can be used from the 'self_req_handler.nim' module.

## How to test

Run the new unittest that is being added:
```
nim c -r -d:chronicles_log_level=DEBUG -d:release -d:postgres  -d:rln --passL:librln_v0.3.4.a --passL:-lm -d:nimDebugDlOpen ./tests/wakunode_rest/test_rest_store.nim test "retrieve historical messages from a self-store-node"
```

## Issue

closes https://github.com/waku-org/nwaku/issues/2196

### DIFF
--- a/tests/wakunode_rest/test_rest_store.nim
+++ b/tests/wakunode_rest/test_rest_store.nim
@@ -534,3 +534,67 @@ procSuite "Waku v2 Rest API - Store":
     await restServer.stop()
     await restServer.closeWait()
     await node.stop()
+
+
+  asyncTest "retrieve historical messages from a self-store-node":
+    ## This test aims to validate the correct message retrieval for a store-node which exposes
+    ## a REST server.
+
+    # Given
+    let node = testWakuNode()
+    await node.start()
+
+    let restPort = Port(58014)
+    let restAddress = parseIpAddress("0.0.0.0")
+    let restServer = RestServerRef.init(restAddress, restPort).tryGet()
+
+    installStoreApiHandlers(restServer.router, node)
+    restServer.start()
+
+    # WakuStore setup
+    let driver: ArchiveDriver = QueueDriver.new()
+    let mountArchiveRes = node.mountArchive(driver)
+    assert mountArchiveRes.isOk(), mountArchiveRes.error
+
+    await node.mountStore()
+
+    # Now prime it with some history before tests
+    let msgList = @[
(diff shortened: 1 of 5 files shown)


alrevuelta (approved):
lgtm. great feature.

tests/wakunode_rest/test_rest_store.nim
```diff
@@ -534,3 +534,67 @@ procSuite "Waku v2 Rest API - Store":
...
+
+  asyncTest "retrieve historical messages from a self-store-node":
+    ## This test aims to validate the correct message retrieval for a store-node which exposes
+    ## a REST server.
+
+    # Given
+    let node = testWakuNode()
+    await node.start()
+
+    let restPort = Port(58014)
```
SionoiS, line 547:
```suggestion
    let restPort = Port(0)
```

SionoiS (approved):
LGTM! Thank you

tests/wakunode_rest/test_rest_store.nim
Ivansete-status: Thanks for the suggestion! However, I won't apply it as it resulted in an exception:
```
    Unhandled error: Communication failed while sending/receiving request, http error [HttpConnectionError]: Could not connect to remote host [RestCommunicationError]

```

PR #2389, opened 2024-02-01
chore: ci/Jenkinsfile.release: enforce -d:postgres flag is always used

Ivansete-status:
## Description
Simple PR to enforce the `-d:postgres` flag is always being set when creating a release image.
Jenkins is not 100% reliable in terms of its parameters and therefore, we need to enforce the usage of this compulsory compilation flag in releases.

With that, we want to prevent errors like the following:
```
2024-01-26 16:05:49 ERR 2024-01-26 14:05:49.204+00:00 4/7 Mounting protocols failed              topics="wakunode main" tid=1 file=wakunode2.nim:89 error="failed to setup archive driver: Postgres has been configured but not been compiled. Check compiler definitions."
```

## Issue
closes https://github.com/waku-org/nwaku/issues/2372

### DIFF
--- a/ci/Jenkinsfile.release
+++ b/ci/Jenkinsfile.release
@@ -44,7 +44,6 @@ pipeline {
         '-d:disableMarchNative',
         '-d:chronicles_colors:none',
         '-d:insecure',
-        '-d:postgres',
       ].join(' ')
     )
     choice(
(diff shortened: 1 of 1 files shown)


gabrielmer (approved):
Yess! Thank you!

SionoiS (approved):
Thanks!

PR #2390, opened 2024-02-01
feat: Added simple, configurable rate limit for lightpush and store-query

NagyZoltanPeter:
# Description
As a first step for rate and bandwidth caping of non relay protocol, hereby a very simple request rate / period limitation is added to light-push and store protocols.

# Changes
- [X] with use of chronos TokenBucket type of rate limit check is added for lightpush
- [X] with use of chronos TokenBucket type of rate limit check is added for store query
- [X] global cli configuration is added to define request limit and the associated time period (same applied for both protocols) - defaults now set to disable rate limitation in wakunode
- [x] more tests from REST side
- [x] add metrics for total requests and rejected counts.
- [ ] add new dashboard for wakunode grafana
- [x] add status code response for light push (might be better cover it in separate PR.
- [ ] waku docs update on new config.

# issue
https://github.com/waku-org/nwaku/issues/2032

### DIFF
--- a/tests/waku_lightpush/test_ratelimit.nim
+++ b/tests/waku_lightpush/test_ratelimit.nim
@@ -57,9 +57,8 @@ suite "Rate limited push service":
 
       check await handlerFuture.withTimeout(50.millis)
 
-      check:
-        requestRes.isOk()
-        handlerFuture.finished()
+      assert requestRes.isOk(), requestRes.error
+      check handlerFuture.finished()
 
       let (handledMessagePubsubTopic, handledMessage) = handlerFuture.read()
 
@@ -105,7 +104,7 @@ suite "Rate limited push service":
 
     let
       server =
-        await newTestWakuLightpushNode(serverSwitch, handler, some((3, 1000.millis)))
+        await newTestWakuLightpushNode(serverSwitch, handler, some((3, 500.millis)))
       client = newTestWakuLightpushClient(clientSwitch)
 
     let serverPeerId = serverSwitch.peerInfo.toRemotePeerInfo()
@@ -143,7 +142,7 @@ suite "Rate limited push service":
 
     await rejectProc()
 
-    await sleepAsync(1000.millis)
+    await sleepAsync(500.millis)
 
     ## next one shall succeed due to the rate limit time window has passed
     await successProc()
--- a/tests/waku_lightpush/test_ratelimit.nim
+++ b/tests/waku_lightpush/test_ratelimit.nim
@@ -1,0 +1,152 @@
+{.used.}
+
+import
+  std/[options, strscans],
+  testutils/unittests,
+  chronicles,
+  chronos,
+  libp2p/crypto/crypto
+
+import
+  ../../waku/[
+    node/peer_manager,
+    common/ratelimit,
+    waku_core,
+    waku_lightpush,
+    waku_lightpush/client,
+    waku_lightpush/common,
+    waku_lightpush/protocol_metrics,
+    waku_lightpush/rpc,
+    waku_lightpush/rpc_codec,
+  ],
+  ../testlib/[assertions, wakucore, testasync, futures, testutils],
+  ./lightpush_utils,
+  ../resources/[pubsub_topics, content_topics, payloads]
+
+suite "Rate limited push service":
+  asyncTest "push message with rate limit not violated":
+    ## Setup
+    let
+      serverSwitch = newTestSwitch()
+      clientSwitch = newTestSwitch()
+
+    await allFutures(serverSwitch.start(), clientSwitch.start())
+
+    ## Given
+    var handlerFuture = newFuture[(string, WakuMessage)]()
+    let handler: PushMessageHandler = proc(
+        peer: PeerId, pubsubTopic: PubsubTopic, message: WakuMessage
+    ): Future[WakuLightPushResult[void]] {.async.} =
+      handlerFuture.complete((pubsubTopic, message))
+      return ok()
+
+    let
+      tokenPeriod = 500.millis
+      server =
+        await newTestWakuLightpushNode(serverSwitch, handler, some((3, tokenPeriod)))
+      client = newTestWakuLightpushClient(clientSwitch)
+
+    let serverPeerId = serverSwitch.peerInfo.toRemotePeerInfo()
+
+    let sendMsgProc = proc(): Future[void] {.async.} =
+      let message = fakeWakuMessage()
+
+      handlerFuture = newFuture[(string, WakuMessage)]()
+      let requestRes =
+        await client.publish(DefaultPubsubTopic, message, peer = serverPeerId)
+
+      check await handlerFuture.withTimeout(50.millis)
+
+      check:
+        requestRes.isOk()
+        handlerFuture.finished()
+
+      let (handledMessagePubsubTopic, handledMessage) = handlerFuture.read()
+
+      check:
+        handledMessagePubsubTopic == DefaultPubsubTopic
+        handledMessage == message
+
+    let waitInBetweenFor = 20.millis
+
+    # Test cannot be too explicit about the time when the TokenBucket resets
+    # the internal timer, although in normal use there is no use case to care about it.
+    var firstWaitExtend = 300.millis
+
+    for runCnt in 0 ..< 3:
+      let startTime = Moment.now()
+      for testCnt in 0 ..< 3:
+        await sendMsgProc()
+        await sleepAsync(20.millis)
+
+      var endTime = Moment.now()
+      var elapsed: Duration = (endTime - startTime)
+      await sleepAsync(tokenPeriod - elapsed + firstWaitExtend)
+      firstWaitEXtend = 100.millis
+
+    ## Cleanup
+    await allFutures(clientSwitch.stop(), serverSwitch.stop())
+
+  asyncTest "push message with rate limit reject":
+    ## Setup
+    let
+      serverSwitch = newTestSwitch()
+      clientSwitch = newTestSwitch()
+
+    await allFutures(serverSwitch.start(), clientSwitch.start())
+
+    ## Given
+    var handlerFuture = newFuture[(string, WakuMessage)]()
+    let handler = proc(
+        peer: PeerId, pubsubTopic: PubsubTopic, message: WakuMessage
+    ): Future[WakuLightPushResult[void]] {.async.} =
+      handlerFuture.complete((pubsubTopic, message))
+      return ok()
+
+    let
+      server =
+        await newTestWakuLightpushNode(serverSwitch, handler, some((3, 1000.millis)))
+      client = newTestWakuLightpushClient(clientSwitch)
+
+    let serverPeerId = serverSwitch.peerInfo.toRemotePeerInfo()
+    let topic = DefaultPubsubTopic
+
+    let successProc = proc(): Future[void] {.async.} =
+      let message = fakeWakuMessage()
+      handlerFuture = newFuture[(string, WakuMessage)]()
+      let requestRes =
+        await client.publish(DefaultPubsubTopic, message, peer = serverPeerId)
+      discard await handlerFuture.withTimeout(10.millis)
+
+      check:
+        requestRes.isOk()
+        handlerFuture.finished()
+      let (handledMessagePubsubTopic, handledMessage) = handlerFuture.read()
+      check:
+        handledMessagePubsubTopic == DefaultPubsubTopic
+        handledMessage == message
+
+    let rejectProc = proc(): Future[void] {.async.} =
+      let message = fakeWakuMessage()
+      handlerFuture = newFuture[(string, WakuMessage)]()
+      let requestRes =
+        await client.publish(DefaultPubsubTopic, message, peer = serverPeerId)
+      discard await handlerFuture.withTimeout(10.millis)
+
+      check:
+        requestRes.isErr()
+        requestRes.error == "TOO_MANY_REQUESTS"
+
+    for testCnt in 0 .. 2:
+      await successProc()
+      await sleepAsync(20.millis)
+
+    await rejectProc()
+
+    await sleepAsync(1000.millis)
+
+    ## next one shall succeed due to the rate limit time window has passed
+    await successProc()
+
+    ## Cleanup
+    await allFutures(clientSwitch.stop(), serverSwitch.stop())
--- a/tests/waku_store/test_wakunode_store.nim
+++ b/tests/waku_store/test_wakunode_store.nim
@@ -244,9 +244,7 @@ procSuite "WakuNode - Store":
 
     server.wakuFilterClient.registerPushHandler(filterHandler)
     let resp = waitFor server.filterSubscribe(
-      some(DefaultPubsubTopic),
-      DefaultContentTopic,
-      peer = filterSourcePeer,
+      some(DefaultPubsubTopic), DefaultContentTopic, peer = filterSourcePeer
     )
 
     waitFor sleepAsync(100.millis)
@@ -333,7 +331,7 @@ procSuite "WakuNode - Store":
     let mountArchiveRes = server.mountArchive(archiveA)
     assert mountArchiveRes.isOk(), mountArchiveRes.error
 
-    waitFor server.mountStore((4, 1.seconds))
+    waitFor server.mountStore((4, 500.millis))
 
     client.mountStoreClient()
 
@@ -344,7 +342,7 @@ procSuite "WakuNode - Store":
     let requestProc = proc() {.async.} =
       let queryRes = waitFor client.query(req, peer = serverPeer)
 
-      check queryRes.isOk()
+      assert queryRes.isOk(), queryRes.error
 
       let response = queryRes.get()
       check:
@@ -354,7 +352,7 @@ procSuite "WakuNode - Store":
       waitFor requestProc()
       waitFor sleepAsync(20.millis)
 
-    waitFor sleepAsync(1.seconds)
+    waitFor sleepAsync(500.millis)
 
     for count in 0 ..< 4:
       waitFor requestProc()
--- a/tests/waku_store/test_wakunode_store.nim
+++ b/tests/waku_store/test_wakunode_store.nim
@@ -319,3 +319,97 @@ procSuite "WakuNode - Store":
 
     # Cleanup
     waitFor allFutures(client.stop(), server.stop())
+
+  test "Store protocol queries does not violate request rate limitation":
+    ## Setup
+    let
+      serverKey = generateSecp256k1Key()
+      server = newTestWakuNode(serverKey, parseIpAddress("0.0.0.0"), Port(0))
+      clientKey = generateSecp256k1Key()
+      client = newTestWakuNode(clientKey, parseIpAddress("0.0.0.0"), Port(0))
+
+    waitFor allFutures(client.start(), server.start())
+
+    let mountArchiveRes = server.mountArchive(archiveA)
+    assert mountArchiveRes.isOk(), mountArchiveRes.error
+
+    waitFor server.mountStore((4, 1.seconds))
+
+    client.mountStoreClient()
+
+    ## Given
+    let req = HistoryQuery(contentTopics: @[DefaultContentTopic])
+    let serverPeer = server.peerInfo.toRemotePeerInfo()
+
+    let requestProc = proc() {.async.} =
+      let queryRes = waitFor client.query(req, peer = serverPeer)
+
+      check queryRes.isOk()
+
+      let response = queryRes.get()
+      check:
+        response.messages == msgListA
+
+    for count in 0 ..< 4:
+      waitFor requestProc()
+      waitFor sleepAsync(20.millis)
+
+    waitFor sleepAsync(1.seconds)
+
+    for count in 0 ..< 4:
+      waitFor requestProc()
+      waitFor sleepAsync(20.millis)
+
+    # Cleanup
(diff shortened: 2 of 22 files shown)


NagyZoltanPeter:
Notice:
I [issued a bug for nim-chronos](https://github.com/status-im/nim-chronos/issues/500) as I found an opinionated bug in TokenBucket implementation that is used for rate limiting.
It prevented testing rate limit capping through REST API, due to the more natural timing than in other tests.
I think TokenBucket implementation is wrong with updating timestamp for calculating budget that allows to overrun the required budget.

gabrielmer (commented):
Looks amazing! 🔥 🔥 🔥 Thanks so much!

Added some small questions :))

waku/common/ratelimit.nim
```diff
@@ -0,0 +1,28 @@
...
+let DefaultGlobalNonRelayRateLimit*: RateLimitSetting = (60, 1.minutes)
+let OptionalDefaultGlobalNonRelayRateLimit*: Option[RateLimitSetting] = some(DefaultGlobalNonRelayRateLimit)
+
+proc newTokenBucket*(setting: Option[RateLimitSetting]): Option[TokenBucket] =
+
+  if setting.isNone:
+    return none[TokenBucket]()
+
+  let (volume, period) = setting.get()
+  if volume <= 0 or period <= 0.seconds:
```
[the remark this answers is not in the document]
NagyZoltanPeter: Ah, do you mean someone would disable a service with such settings? I think that's a complete missuse.
Why would you mount a service than disable with an impossible rate limit?

At least I stated in the config description that 0 means disable rate limiting.

waku/waku_lightpush/common.nim
```diff
@@ -19,3 +19,6 @@ type PushMessageHandler* = proc(
   pubsubTopic: PubsubTopic,
   message: WakuMessage
   ): Future[WakuLightPushResult[void]] {.async.}
+
+const TooManyRequestMessage* = "TOO_MANY_REQUESTS"
```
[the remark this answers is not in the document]
NagyZoltanPeter: Yes, good point! Will change it.

waku/common/ratelimit.nim
```diff
@@ -0,0 +1,28 @@
...
+  ratelimit
+
+
+type
+  RateLimitSetting* = tuple
+    volume: int
+    period: Duration
+
+let DefaultGlobalNonRelayRateLimit*: RateLimitSetting = (60, 1.minutes)
+let OptionalDefaultGlobalNonRelayRateLimit*: Option[RateLimitSetting] = some(DefaultGlobalNonRelayRateLimit)
```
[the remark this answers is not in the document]
NagyZoltanPeter: Thanks for pointing out... not necessary, its a leftover. Will remove it.

NagyZoltanPeter:
https://github.com/waku-org/js-waku/pull/1952 covers one issue caused by changing protocol interface in js-waku tests.

waku/waku_store/protocol.nim
```diff
@@ -55,6 +63,18 @@ proc initProtocolHandler(ws: WakuStore) =
...
+      trace "store query request rejected due rate limit exceeded",
+        peerId = $conn.peerId, requestId = reqRpc.requestId
+      let error = HistoryError(kind: HistoryErrorKind.TOO_MANY_REQUESTS).toRPC()
+      let response = HistoryResponseRPC(error: error)
+      let rpc = HistoryRPC(requestId: reqRpc.requestId, response: some(response))
+      await conn.writeLp(rpc.encode().buffer)
+      waku_service_requests_rejected.inc(labelValues = ["Lightpush"])
+      return
+
+    waku_service_requests.inc(labelValues = ["Lightpush"])
```
gabrielmer, line 76: shouldn't be "Store" instead of "Lightpush"?

waku/waku_store/protocol.nim
```diff
@@ -55,6 +63,18 @@ proc initProtocolHandler(ws: WakuStore) =
       # TODO: Return (BAD_REQUEST, cause: "empty query")
       return
 
+    if ws.requestRateLimiter.isSome() and not ws.requestRateLimiter.get().tryConsume(1):
+      trace "store query request rejected due rate limit exceeded",
+        peerId = $conn.peerId, requestId = reqRpc.requestId
+      let error = HistoryError(kind: HistoryErrorKind.TOO_MANY_REQUESTS).toRPC()
+      let response = HistoryResponseRPC(error: error)
+      let rpc = HistoryRPC(requestId: reqRpc.requestId, response: some(response))
```
gabrielmer, line 71: Is all this related to the RPC that we deprecated? Same for the similar code block for Lightpush

waku/waku_store/rpc.nim
```diff
@@ -62,6 +62,7 @@ type
     ## the state of its request
     NONE = uint32(0)
     INVALID_CURSOR = uint32(1)
+    TOO_MANY_REQUESTS = uint32(429)
```
gabrielmer, line 65: same in this file, isn't this supposed to be deprecated?

gabrielmer (commented):
Thanks so much! Added some small questions about points that I still don't understand completely

waku/waku_store/protocol.nim
gabrielmer: pretty sure it's not the same that what we deprecated, but ask just in case for further understanding :))

waku/common/tokenbucket.nim
```diff
@@ -0,0 +1,61 @@
...
+type TokenBucket* = ref object
+  budget*: int
+  budgetCap: int
+  lastTimeFull: Moment
+  fillDuration: Duration
+
+proc update(bucket: TokenBucket, currentTime: Moment) =
+  if bucket.fillDuration == default(Duration):
+    bucket.budget = min(bucket.budgetCap, bucket.budget)
+    return
```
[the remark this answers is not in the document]
NagyZoltanPeter: Ah, ok. Defaul of Duration is 0 meaning unlimited, so we won't care about timing of replenish the buffer.

waku/waku_store/protocol.nim
NagyZoltanPeter: Great catch!!!! Thank you.

gabrielmer (approved):
Amaaaazing! Thanks so much!

waku/waku_store/rpc.nim
NagyZoltanPeter: No, it should not definetely. :-) It is store protocol definition.

tests/wakunode_rest/test_rest_store.nim
```diff
@@ -638,3 +638,147 @@ procSuite "Waku v2 Rest API - Store":
...
+      encodeUrl(reqPubsubTopic),
+      "", # content topics. Empty ignores the field.
+      "", # start time. Empty ignores the field.
+      "", # end time. Empty ignores the field.
+      encodeUrl($reqSenderTime), # sender time
+      encodeUrl($reqStoreTime), # store time
+      reqDigest.toRestStringMessageDigest(),
+        # base64-encoded digest. Empty ignores the field.
+      "5", # page size. Empty implies default page size.
+      "true", # ascending
```
Ivansete-status, line 747: I think we can simplify the test by leaving the default params. I got confused why using `5` in this case but I think is not needed to be a different page size

waku/common/tokenbucket.nim
```diff
@@ -0,0 +1,61 @@
...
+import chronos
+
+## This is an extract from chronos/ratelimit.nim due to the found bug in the original implementation.
+## Unfortunately that bug cannot be solved without harm the original features of TokenBucket class.
+## So, this current shortcut is used to enable move ahead with nwaku rate limiter implementation.
+type TokenBucket* = ref object
+  budget*: int
+  budgetCap: int
+  lastTimeFull: Moment
+  fillDuration: Duration
```
Ivansete-status, line 15: May kindly add a comment wrt purpose of each attribute?

waku/common/tokenbucket.nim
```diff
@@ -0,0 +1,61 @@
...
+  if bucket.budget >= tokens:
+    bucket.budget -= tokens
+    return true
+  else:
+    return false
+
+proc replenish*(bucket: TokenBucket, tokens: int, now = Moment.now()) =
+  ## Add `tokens` to the budget (capped to the bucket capacity)
+  bucket.budget += tokens
+  bucket.update(now)
```
Ivansete-status, line 52: This `proc` seems not to be used anywhere. Maybe we can delete it?

waku/common/tokenbucket.nim
```diff
@@ -0,0 +1,61 @@
...
+
+  let timeDeltaFromLastFull = currentTime - bucket.lastTimeFull
+
+  if timeDeltaFromLastFull.milliseconds < bucket.fillDuration.milliseconds:
+    return
+
+  bucket.budget = bucket.budgetCap
+  bucket.lastTimeFull = currentTime
+
+proc tryConsume*(bucket: TokenBucket, tokens: int, now = Moment.now()): bool =
```
Ivansete-status, line 30: The `now` parameter seems not to be overridden anywhere. In this case, I think we can just have a local `now` var

waku/factory/external_config.nim
```diff
@@ -589,6 +589,22 @@ type WakuNodeConf* = object
...
+      defaultValue: 0,
+      name: "request-rate-limit"
+    .}: int
+
+    ## Currently default to switch of rate limit until become official
+    requestRatePeriod* {.
+      desc: "Period of request rate limitation in seconds. Set it to 0 for unlimited",
+      defaultValue: 0,
+      name: "request-rate-period"
+    .}: int64
```
Ivansete-status, line 606: Maybe you already commented and I might be missing something but these params seem not to be used :)

waku/common/tokenbucket.nim
```diff
@@ -0,0 +1,61 @@
...
+## Unfortunately that bug cannot be solved without harm the original features of TokenBucket class.
+## So, this current shortcut is used to enable move ahead with nwaku rate limiter implementation.
+type TokenBucket* = ref object
+  budget*: int
+  budgetCap: int
+  lastTimeFull: Moment
+  fillDuration: Duration
+
+proc update(bucket: TokenBucket, currentTime: Moment) =
+  if bucket.fillDuration == default(Duration):
```
Ivansete-status, line 18: Could you kindly add a comment elaborating on the purpose of this `proc` and when is being used?

Ivansete-status (approved):
Masterpiece PR! Thanks for it! :100: 
Just added few comments that I hope you find useful.
Something that we might need to consider in the future is to expose `byte` rate metrics.

For now, we expose all the `libp2p` in/out transmission metrics from:
https://github.com/vacp2p/nim-libp2p/blob/09b3e11956459ad4c364b353b7f1067f42267997/libp2p/stream/chronosstream.nim#L112
and
https://github.com/vacp2p/nim-libp2p/blob/09b3e11956459ad4c364b353b7f1067f42267997/libp2p/stream/chronosstream.nim#L133

It will be very interesting to know the percentage (and absolute values) of the bandwidth (bytes/sec) consumed by each protocol. Something to consider in upcoming PRs :P

Cheers

NagyZoltanPeter:
> Masterpiece PR! Thanks for it! 💯 Just added few comments that I hope you find useful. Something that we might need to consider in the future is to expose `byte` rate metrics.
> 
> For now, we expose all the `libp2p` in/out transmission metrics from: https://github.com/vacp2p/nim-libp2p/blob/09b3e11956459ad4c364b353b7f1067f42267997/libp2p/stream/chronosstream.nim#L112 and https://github.com/vacp2p/nim-libp2p/blob/09b3e11956459ad4c364b353b7f1067f42267997/libp2p/stream/chronosstream.nim#L133
> 
> It will be very interesting to know the percentage (and absolute values) of the bandwidth (bytes/sec) consumed by each protocol. Something to consider in upcoming PRs :P
> 
> Cheers

Yes, we have a separate issue tracking that: https://github.com/waku-org/nwaku/issues/1945

tests/waku_lightpush/test_ratelimit.nim
```diff
@@ -0,0 +1,152 @@
...
+      let message = fakeWakuMessage()
+
+      handlerFuture = newFuture[(string, WakuMessage)]()
+      let requestRes =
+        await client.publish(DefaultPubsubTopic, message, peer = serverPeerId)
+
+      check await handlerFuture.withTimeout(50.millis)
+
+      check:
+        requestRes.isOk()
```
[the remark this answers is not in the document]
NagyZoltanPeter: I'm not pretty sure about this approach, as assert (similarly to 'require') will abort the test execution and prevent other tests to run.
In this case I really would like to check if the response is ok or not.

tests/waku_lightpush/test_ratelimit.nim
```diff
@@ -0,0 +1,152 @@
...
+        requestRes.error == "TOO_MANY_REQUESTS"
+
+    for testCnt in 0 .. 2:
+      await successProc()
+      await sleepAsync(20.millis)
+
+    await rejectProc()
+
+    await sleepAsync(1000.millis)
+
```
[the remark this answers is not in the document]
NagyZoltanPeter: Yes, it was in my mind. But found that the whole test is very sensitive to timing. The base of this problem that there is a quite large gap between the test case starts a timer and make assumption on it and when the exact timing will occur on the protocol an token bucket side. 
This is normally not a problem as user of the protocol will not make assumptions normally on timing the request. This is mostly a testing problem.
But agree to avoid such sleeps as much as possible, but this rate limit tests are time sensitive ones.

tests/waku_store/test_wakunode_store.nim
```diff
@@ -319,3 +319,97 @@ procSuite "WakuNode - Store":
...
+      server = newTestWakuNode(serverKey, parseIpAddress("0.0.0.0"), Port(0))
+      clientKey = generateSecp256k1Key()
+      client = newTestWakuNode(clientKey, parseIpAddress("0.0.0.0"), Port(0))
+
+    waitFor allFutures(client.start(), server.start())
+
+    let mountArchiveRes = server.mountArchive(archiveA)
+    assert mountArchiveRes.isOk(), mountArchiveRes.error
+
+    waitFor server.mountStore((4, 1.seconds))
```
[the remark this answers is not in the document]
NagyZoltanPeter: Ok, I will try to run with lower waits, let's see how stable tests are with that.

tests/wakunode_rest/test_rest_store.nim
NagyZoltanPeter: Agree

tests/waku_lightpush/test_ratelimit.nim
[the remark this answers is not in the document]
NagyZoltanPeter: Ok, done as discussed on discord.

waku/common/tokenbucket.nim
NagyZoltanPeter: Sure!

alrevuelta (dismissed):
Some comments:

1) We are missing some DoS attack vectors. For example [this](https://github.com/waku-org/nwaku/blob/c0c1981c8ce5d7e7ae7604bd263535953d30a868/waku/waku_store/protocol.nim#L51-L56) allows an attacker to DoS the node, since it will naively read from the connection and return. We should also count read bytes, not just requests made. And not only on valid requests but also when RPC decode is not successful. [This](https://github.com/vacp2p/nim-libp2p/blob/unstable/libp2p/protocols/pubsub/gossipsub.nim#L456-L463) can be used for inspiration. I would suggest something similar, where if a given peer has sent more than x bytes/unit of time, `readLp` is not called + we disconnect from that peer without reading from the connection. In short: This code allows an attacker to flood with random bytes the peer without any limits.

```nim
let buf = await conn.readLp(MaxRpcSize.int)

let decodeRes = HistoryRPC.decode(buf)
if decodeRes.isErr():
  error "failed to decode rpc", peerId = $conn.peerId
  waku_store_errors.inc(labelValues = [decodeRpcFailure])
  # TODO: Return (BAD_REQUEST, cause: "decode rpc failed")
  return
```

2) Unless I'm missing something, the rate limit is applied globally (eg to all peers), which means that a given peer can trigger this rate limit and won't allow other nodes to get any service from the node. imho the rate limit should be per peerId (or per IP address) so that we allow x amount of usage (bytes or requests) per peer instead of globally. Without this, we are still exposed to DoS attacks.

3) Can we split this PR into different ones? I see 3 clear different PRs here: i) introduce tocken bucket, ii) rate limit lightpush, iii) rate limit store.

4) Do we really need rate limits at the rest API level? afaik we can assume that the REST API is exposed to trusted parties so not sure if we need to rate limit there. imho we should focus on the libp2p protocol, which is the service exposed globally.

waku/common/tokenbucket.nim
```diff
@@ -0,0 +1,61 @@
+when (NimMajor, NimMinor) < (1, 4):
+  {.push raises: [Defect].}
+else:
+  {.push raises: [].}
+
+import chronos
+
+## This is an extract from chronos/ratelimit.nim due to the found bug in the original implementation.
```
alrevuelta, line 8: Mind pointing to the bug and elaborating more? Especially why it can't be fixed upstream.

waku/common/tokenbucket.nim
NagyZoltanPeter: This is for the same reason as above. Keep the gates open to start using chronos implementation again and remove this code dup in future.

NagyZoltanPeter: This maybe needed for us as well to drive own logic of replenishing the bucket by our own.

waku/common/tokenbucket.nim
NagyZoltanPeter: In libp2p chronos' TockenBucket implementation is used but a bit differently than our expectation. It has a kind of time statistics replenish method.
It heavily depends on the usage and can ruin our usage. Unfortunately I could not - yet - fix it that may securely not affect libp2p usage. That fix needs a bit more wide testing for that I did not want to block this feature in nwaku.
I have an idea how to fix that and already agreed the aproach with Jacek... he encouraged us to take actions and co-work with other projects more easy.

waku/common/tokenbucket.nim
NagyZoltanPeter: Sure.

NagyZoltanPeter:
> > Some comments:
> > 
> > 1. We are missing some DoS attack vectors. For example [this](https://github.com/waku-org/nwaku/blob/c0c1981c8ce5d7e7ae7604bd263535953d30a868/waku/waku_store/protocol.nim#L51-L56) allows an attacker to DoS the node, since it will naively read from the connection and return. We should also count read bytes, not just requests made. And not only on valid requests but also when RPC decode is not successful. [This](https://github.com/vacp2p/nim-libp2p/blob/unstable/libp2p/protocols/pubsub/gossipsub.nim#L456-L463) can be used for inspiration. I would suggest something similar, where if a given peer has sent more than x bytes/unit of time, `readLp` is not called + we disconnect from that peer without reading from the connection. In short: This code allows an attacker to flood with random bytes the peer without any limits.
> > 
> > ```nim
> > let buf = await conn.readLp(MaxRpcSize.int)
> > 
> > let decodeRes = HistoryRPC.decode(buf)
> > if decodeRes.isErr():
> >   error "failed to decode rpc", peerId = $conn.peerId
> >   waku_store_errors.inc(labelValues = [decodeRpcFailure])
> >   # TODO: Return (BAD_REQUEST, cause: "decode rpc failed")
> >   return
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > 2. Unless I'm missing something, the rate limit is applied globally (eg to all peers), which means that a given peer can trigger this rate limit and won't allow other nodes to get any service from the node. imho the rate limit should be per peerId (or per IP address) so that we allow x amount of usage (bytes or requests) per peer instead of globally. Without this, we are still exposed to DoS attacks.
> > 3. Can we split this PR into different ones? I see 3 clear different PRs here: i) introduce tocken bucket, ii) rate limit lightpush, iii) rate limit store.
> > 4. Do we really need rate limits at the rest API level? afaik we can assume that the REST API is exposed to trusted parties so not sure if we need to rate limit there. imho we should focus on the libp2p protocol, which is the service exposed globally.
> 
> Thanks for these comments, all really valuable! The first item is a real catch. To explain a bit more detailed with this first PR we only intend to introduce a basic simple rate limit support, first for two non-relay protocols to be testable. Other tasks will follow in different ways.
> 
> * extend rate limitation to other non-relay protocols.
> * fine tune rate limiting with better scalable thresholds.
> * per peer tracking of request
> * bandwidth limit to be introduced (as you linked like libp2p has it) per shard.
>   
>   * Tracking bw on each shard may allow a node to dynamically decide to support a shard or not based upon its bandwidth profiles.
> 
> Of course there are some open questions along this road as rate and bandwidth limits may contradict each other, at least in priority level. Within all these tasks - as I see - we try to hit two target, giving control over resource usage for node operators and protect nodes from different kind of flood attacks. If you would agree this PR should go as is for now as a initial step that will be followed by more fine tuned solution. Your fist point is an absolute threat to be handled. The item 4 is something to be considered, but in this current solution there is no different from where the request is coming as these non-relay protocol level rate limiting is on full node protocol level. The only question is if we want to support to run a full node and yet let the owner to address rest calls (store/lightpush/...) directly without using a lightnode?
> 
> @alrevuelta : WDYT?

Findings are now address in https://github.com/waku-org/nwaku/issues/2589 for future handle them

PR #2395, opened 2024-02-02
fix: bug fix in ci/Jenkinsfile.release: make -d:postgres part of NIMFLAGS

Ivansete-status:
## Description
This PR fixes an issue that came from a change where we enforced the `-d:postgres` flag to always be used when creating a release.

Before this fix, the `-d:postgres` wasn't considered properly as a `NIMFLAG`. Instead, it was considered as a Docker parameter.

image

@apentori - apologies for the inconvenience and thanks for the hints!

## Issue
https://github.com/waku-org/nwaku/issues/2372


gabrielmer (approved):
Thank you!

apentori (approved):
If at one point you want to build without the flag you won't be abble to remove it without making a change to the Jenkinsfile.
Imo letting the param in the `NIMFLAGS` is better, but if you prefer this

NagyZoltanPeter (approved):
Thank you for it.

PR #2397, opened 2024-02-05
feat: add yamux support

### LINKED ISSUE #2331: chore: use yamux instead of mplex
### Background
The default muxer in go-libp2p is yamux, we should use the same. As seen in https://github.com/waku-org/go-waku/pull/851

### Details
replace this with yamux
https://github.com/waku-org/nwaku/blob/b0c780528b5610f8b6c5e3784dc9faadcb7b6917/waku/node/waku_switch.nim#L93

richard-ramos:
# Description
Added support to yamux. Since the order of multiplexers matter, it's added after mplex so it continues being the default multiplexer (instead of completely removing mplex, as suggested in the og issue). This is required since go-libp2p is removing support for mplex.

Closes #2331

### DIFF
--- a/waku/node/waku_switch.nim
+++ b/waku/node/waku_switch.nim
@@ -91,6 +91,7 @@ proc newWakuSwitch*(
       .withMaxOut(maxOut)
       .withMaxConnsPerPeer(maxConnsPerPeer)
       .withMplex(inTimeout, outTimeout)
+      .withYamux()
       .withNoise()
       .withTcpTransport(transportFlags)
       .withNameResolver(nameResolver)


SionoiS (approved):
LGTM

CI green but I still have doubts.

In theory nim-libp2p should use Yamux for go-libp2p interop and fall back to Mplex if needed right?

Does this need to be tested more? Maybe I'm paranoid :P

richard-ramos:
> In theory nim-libp2p should use Yamux for go-libp2p interop and fall back to Mplex if needed right?

I tried using `wakunode2` and it could connect succesfully against peers that support only yamux or only mplex, or both!

chaitanyaprem:
> Since the order of multiplexers matter, it's added after mplex so it continues being the default multiplexer

If a peer supports both mplex and yamux, then yamux would be used right?
mplex would only be used if a peer doesn't support yamux.

waku/node/waku_switch.nim
```diff
@@ -91,6 +91,7 @@ proc newWakuSwitch*(
       .withMaxOut(maxOut)
       .withMaxConnsPerPeer(maxConnsPerPeer)
       .withMplex(inTimeout, outTimeout)
```
Ivansete-status, line 93: Should we remove this line?

Ivansete-status (approved):
Thanks so much for it! Just some naïve questions :)

How the stream multiplexer is being picked up? Does it first try to use `mplex` and if the other peer doesn't support it, then it tries with `yamux`?

richard-ramos:
@chaitanyaprem @Ivansete-status 
If i understood correctly the code in https://github.com/status-im/nim-libp2p/blob/eb0890cd6f5d36e1b3ae414bffda59d31dfc2e7c/libp2p/upgrademngrs/muxedupgrade.nim#L43-L45 nim-libp2p will try muxers depending on the order they're defined both in identify protocol and when building the switcher, so it is like @Ivansete-status says: it will attempt to use mplex first, and if not available, then yamux. 


So with the changes of this PR, if for ex, if go-waku returns  via identify protocol the muxers protocol ids in this order:  `[yamux_protocol_id, mplex protocol_id]`, nim-libp2p will use yamux first.  But if go-waku were to return `[mplex_protocol_id`, `yamux_protocol_id`], since in the switcher mplex is used as default (we called `withMplex` before `withYamux`), then it will prefer to use mplex over yamux.

We could switch the order, but i went with mplex by default, since that's what nwaku nodes out there in the wild are using, and eventually just use yamux?

PR #2398, opened 2024-02-06
chore: libwaku - simpler ctx mgmt. Param now receiving void* instead of void**

Ivansete-status:
## Description
This change is needed so that interoperability with other languages becomes simpler. Particularly, this simplification is needed from the Python point of view, where it is tricky to pass a void** as a parameter to an FFI function.


NagyZoltanPeter (approved):
Looks much better! Great!

SionoiS (approved):
Thanks!

gabrielmer (approved):
Thanks so much!

PR #2403, opened 2024-02-06
feat: Nwaku Sync

SionoiS:
# Description

This is the new Waku Sync protocol. It works by wrapping the Negentropy protocol which implements RBSR.
See this [research post](https://github.com/waku-org/research/issues/80) for more info.

At configurable intervals (A sync can also be requested ad-hoc), a sync request is sent to a peer chosen by the peer manager. After multiple requests and responses between the 2 peers, a callback containing the list of missing hashes is triggered. With the list of missing hashes, store is used to request the messages from the other node.

WIP spec -> https://github.com/waku-org/specs/blob/feat--waku-sync/standards/core/sync.md

# Changes

- [x] new Waku Sync protocol.
- [x] new config params.
- [x] `waku_node.nim` mount the new protocol.
- [x] Negentropy Nim Bindings.
- [x] tests.

Tracking
- https://github.com/waku-org/pm/issues/132
- https://github.com/waku-org/nwaku/issues/2426


waku/node/waku_node.nim
```diff
@@ -182,6 +184,22 @@ proc connectToNodes*(node: WakuNode, nodes: seq[RemotePeerInfo] | seq[string], s
   # NOTE Connects to the node without a give protocol, which automatically creates streams for relay
   await peer_manager.connectToNodes(node.peerManager, nodes, source=source)
 
+## Waku Sync
+
+proc mountWakuSync*(node: WakuNode): Result[void, string] =
```
alrevuelta, line 189: wondering if sync could be part of store protocol?

SionoiS: Yes but I'm sure there are some use cases where operators might want Sync but not Store.

alrevuelta: Guess in that case you can instantiate the store protocol but with a flag specifying to not serve other nodes and just sync?

No strong opinion, but having sync as another protocol seems odd to me.

SionoiS: Sync will probably have it's own DB and needs to be searchable in the DHT but it could be part of Store for sure.

Code wise it seams best to separate both Store and Sync since they don't have the same responsibility.

I have no strong opinion either both could work. I'll try to think about the pros/cons of both.

waku/waku_sync/protocol.nim
```diff
@@ -0,0 +1,169 @@
...
+  let payload: seq[byte] = @[0]
+
+  # TODO update the hashes
+
+  ok(some(payload))
+
+proc intitialization(self: WakuSync): Future[Result[seq[byte], string]] {.async.} =
+  #TODO call binding to compute the payload
+
+  let payload: seq[byte] = @[0]
```
jm-clius, line 66: Is payload the hash over the entire set? Perhaps something less ambiguous (like hash or range_hash)?

waku/waku_sync/protocol.nim
```diff
@@ -0,0 +1,169 @@
...
+
+  ok(payload)
+
+proc request(self: WakuSync, conn: Connection): Future[Result[seq[WakuMessageHash], string]] {.async, gcsafe.} =
+  let request = (await self.intitialization()).valueOr:
+    return err(error)
+
+  let writeRes = catch: await conn.writeLP(request)
```
jm-clius, line 74: Presumably to be encapsulated within a protobuf at some point?

waku/waku_sync/protocol.nim
jm-clius: Oh nvm. The payload would be the payload for the underlying protocol.

waku/node/waku_node.nim
jm-clius: Although in terms of configuration and operation Store and Sync will function as a unit (and therefore a "single protocol") on the wire the one is concerned with querying a key-value store and the other with reconciliation of a set of keys. This requires different request-responses so would have to be a separate libp2p protocol ID. For that reason it makes sense to me to have some code separation at lowest layer, even if at a higher layer there will be integration.

jm-clius (commented):
I initially imagined writing our own wrapper protocol around the underlying bindings (rather than just passing the payload directly onto the wire). For now, this approach will lead us to faster POC and tests, so seems reasonable to me. Do you have any intuition for what our longer term approach will be?

waku/node/waku_node.nim
SionoiS: > Although in terms of configuration and operation Store and Sync will function as a unit (and therefore a "single protocol") on the wire the one is concerned with querying a key-value store and the other with reconciliation of a set of keys. This requires different request-responses so would have to be a separate libp2p protocol ID. For that reason it makes sense to me to have some code separation at lowest layer, even if at a higher layer there will be integration.

I was thinking that the App layer would be where we tie Store and SYNC.

waku/waku_sync/protocol.nim
SionoiS: > Oh nvm. The payload would be the payload for the underlying protocol.

Yes exactly `payload` is the bad of bytes Negentropy gives us.

waku/waku_sync/protocol.nim
SionoiS: > Presumably to be encapsulated within a protobuf at some point?

Maybe but why? What would be the benefit? One case could be if we want to negotiate `maxFrameSize` between peers. I'm not sure yet.

SionoiS:
> I initially imagined writing our own wrapper protocol around the underlying bindings (rather than just passing the payload directly onto the wire). For now, this approach will lead us to faster POC and tests, so seems reasonable to me. Do you have any intuition for what our longer term approach will be?

Negentropy protocol has only one option `maxFrameSize` that's it. I'm not sure what other information we might want to exchange justifying wrapping the protocol. I'll think about it...

SionoiS:
Sorry I squashed everything and loss your mention @chaitanyaprem

chaitanyaprem:
> Sorry I squashed everything and loss your work attribution @chaitanyaprem

Ah..hope it won't be considered as no contributions from me for the past few weeks :) 

Maybe you can mention me as `Co-author` in comments, like this https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors

SionoiS:
I notice that request are received twice and that connection are not closed properly. I will investigate further.

SionoiS:
I have stashed an update with a new pruning mechanism for after the state machine PR is merged.

waku/waku_sync/protocol.nim
```diff
@@ -0,0 +1,311 @@
...
+  periodicSyncFut: Future[void]
+  periodicPruneFut: Future[void]
+
+proc storageSize*(self: WakuSync): int =
+  self.storage.len
+
+proc ingessMessage*(self: WakuSync, pubsubTopic: PubsubTopic, msg: WakuMessage) =
+  if msg.ephemeral:
+    return
+  #TODO: Do we need to check if this message has already been ingessed?
```
chaitanyaprem, line 52: This may have to be fixed to avoid trouble with syncing, but since it is a minor issue..can be tracked via another github issue so that we don't miss it.

waku/waku_sync/protocol.nim
```diff
@@ -0,0 +1,311 @@
...
+    hash = msgHash.toHex(), timestamp = msg.timestamp
+
+  if self.storage.insert(msg.timestamp, msgHash).isErr():
+    debug "failed to insert message ", hash = msgHash.toHex()
+
+proc calculateRange(jitter: Duration, syncRange: Duration = 1.hours): (int64, int64) =
+  var now = getNowInNanosecondTime()
+
+  # Because of message jitter inherent to Relay protocol
+  now -= jitter.nanos
```
chaitanyaprem, line 67: Do we need to consider nano granularity?
Wondering if a different language sync implementation cannot support nanoSec granularity. 
Can we use second granularity instead?

waku/waku_sync/protocol.nim
```diff
@@ -0,0 +1,311 @@
...
+
+proc request(
+    self: WakuSync, conn: Connection
+): Future[Result[seq[WakuMessageHash], string]] {.async, gcsafe.} =
+  let (start, `end`) = calculateRange(self.relayJitter)
+
+  let initialized =
+    ?clientInitialize(self.storage, conn, self.maxFrameSize, start, `end`)
+
+  debug "sync session initialized",
```
chaitanyaprem, line 84: Wondering if we should have info logs for start and end of sync session with some details and result.

But then again, it may flood logs if syncInterval is too small.

waku/waku_sync/protocol.nim
```diff
@@ -0,0 +1,311 @@
...
+      let completed = error # Result[Reconciled, Completed]
+
+      ?await completed.clientTerminate()
+
+      debug "sync session ended gracefully",
+        client = self.peerManager.switch.peerInfo.peerId, server = conn.peerId
+
+      return ok(hashes)
+
+proc sync*(
```
chaitanyaprem, line 119: Maybe can merge this with function below?
Almost all code seems duplicated...maybe can make peer as optional arg to the Sync method.

waku/waku_sync/protocol.nim
```diff
@@ -0,0 +1,311 @@
...
+    T: type WakuSync,
+    peerManager: PeerManager,
+    maxFrameSize: int = DefaultMaxFrameSize,
+    syncInterval: timer.Duration = DefaultSyncInterval,
+    relayJitter: Duration = DefaultGossipSubJitter,
+    pruneCB: Option[PruneCallBack] = none(PruneCallback),
+    transferCB: Option[TransferCallback] = none(TransferCallback),
+): T =
+  let storage = Storage.new().valueOr:
+    debug "storage creation failed"
```
chaitanyaprem, line 216: Maybe make this an error log...I don't remember why we couldn't log error returned here.
I was facing some weird compilation issue.

waku/factory/external_config.nim
```diff
@@ -357,6 +357,32 @@ type WakuNodeConf* = object
...
+
+    storeSyncRelayJitter* {.
+      desc: "Time offset to account for message propagation jitter. In seconds.",
+      defaultValue: 20,
+      name: "store-sync-relay-jitter"
+    .}: int64
+
+    storeSyncMaxPayloadSize* {.
+      desc:
+        "Max size in bytes of the inner negentropy payload. Cannot be less than 5K, 0 is unlimited.",
```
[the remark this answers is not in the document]
SionoiS: There's no limit AFAIK, this option is more for completeness than anything else. If we get huge payload it means we have much bigger problems than the size of the payload :D

waku/node/waku_node.nim
```diff
@@ -193,6 +195,125 @@ proc connectToNodes*(
...
+  let transfer: TransferCallback = proc(
+      hashes: seq[WakuMessageHash], peerId: PeerId
+  ): Future[Result[void, string]] {.async: (raises: []), closure.} =
+    var query = StoreQueryRequest()
+    query.includeData = true
+    query.messageHashes = hashes
+    query.paginationLimit = some(uint64(100))
+
+    while true:
+      # Do we need a query retry mechanism??
```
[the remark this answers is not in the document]
SionoiS: I feel like the store client should have a retry mechanism TBH.

waku/node/waku_node.nim
```diff
@@ -193,6 +195,125 @@ proc connectToNodes*(
...
+
+      query.paginationCursor = response.paginationCursor
+
+      for kv in response.messages:
+        let handleRes = catch:
+          await node.wakuArchive.handleMessage(kv.pubsubTopic.get(), kv.message.get())
+
+        if handleRes.isErr():
+          trace "message transfer failed", error = handleRes.error.msg
+          # Messages can be synced next time since they are not added to storage yet.
```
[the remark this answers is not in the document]
SionoiS: I meant if we had a retry mechanism.

waku/waku_sync/protocol.nim
SionoiS: I'll add a new github issue

waku/waku_sync/protocol.nim
SionoiS: The precision is in nanoseconds but the actual accuracy is in seconds. This was the simplest thing to do.

waku/waku_sync/protocol.nim
SionoiS: `info` is for non technical users. `debug` about the current state of the mechanism and intended for us. `trace` is for hot paths and when processing many elements.

waku/factory/external_config.nim
```diff
@@ -357,6 +357,32 @@ type WakuNodeConf* = object
...
 
+    ## Sync config
+    storeSync* {.
+      desc: "Enable store sync protocol: true|false",
+      defaultValue: false,
+      name: "store-sync"
+    .}: bool
+
+    storeSyncInterval* {.
+      desc: "Interval between store sync attempts. In seconds.",
```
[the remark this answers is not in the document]
SionoiS: We will have to experiment with different sync intervals.

waku/waku_sync/protocol.nim
```diff
@@ -0,0 +1,307 @@
...
+  if self.transferCallBack.isSome() and self.syncInterval > ZeroDuration:
+    self.periodicSyncFut = self.periodicSync(self.transferCallBack.get())
+
+  if self.pruneCallBack.isSome() and self.syncInterval > ZeroDuration:
+    self.periodicPruneFut = self.periodicPrune(self.pruneCallBack.get())
+
+  info "WakuSync protocol started"
+
+proc stopWait*(self: WakuSync) {.async.} =
+  await self.periodicSyncFut.cancelAndWait()
```
jm-clius, line 305: Should we cancel the `periodicPruneFut`?

waku/node/waku_node.nim
```diff
@@ -193,6 +195,125 @@ proc connectToNodes*(
...
+
+  node.wakuSync = WakuSync.new(
+    peerManager = node.peerManager,
+    maxFrameSize = maxFrameSize,
+    syncInterval = syncInterval,
+    relayJitter = relayJitter,
+    pruneCB = prune,
+    transferCB = some(transfer),
+  ).valueOr:
+    return err("Waku sync initializtion failed: " & error)
```
jm-clius, line 305:
```suggestion
    return err("Waku sync initialization failed: " & error)
```

waku/node/waku_node.nim
```diff
@@ -193,6 +195,125 @@ proc connectToNodes*(
...
+proc mountWakuSync*(
+    node: WakuNode,
+    maxFrameSize: int = DefaultMaxFrameSize,
+    syncInterval: timer.Duration = DefaultSyncInterval,
+    relayJitter: Duration = DefaultGossipSubJitter,
+    enablePruning: bool = true, # For testing purposes
+): Result[void, string] =
+  if not node.wakuSync.isNil():
+    return err("Waku sync already mounted, skipping")
+
```
jm-clius, line 209: It seems to me that, if a node starts up with an existing `archive` we will not prime the `wakuSync` storage on startup with the items already existing in the Store. I imagine the behaviour would be that the sync protocol would believe these items do not exist in the Store and try to sync and store all these items from a neighbour? Or am I missing something?

jm-clius (approved):
Thanks for this. Really great effort and good work (also @chaitanyaprem). I do have a few comments below which I would appreciate your looking into. Approving, though, as I do not want to stand in the way of deployed experiments starting.

waku/waku_sync/raw_bindings.nim
```diff
@@ -0,0 +1,352 @@
...
+  {.push raises: [Defect].}
+else:
+  {.push raises: [].}
+
+from os import DirSep
+
+import std/[strutils], chronicles, std/options, stew/[results, byteutils], confutils
+import ../waku_core/message
+
+const negentropyPath =
```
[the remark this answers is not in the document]
chaitanyaprem: I think it would be possible, but as per recommendations noted in nim language bindings guide in status..I chose share lib.

waku/node/waku_node.nim
chaitanyaprem: This is a valid point, if we had persistent negrnyropy storage then we wouldn't need to do this..but since we are using in memory storage..this has to be done.

waku/waku_sync/storage_manager.nim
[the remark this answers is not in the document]
chaitanyaprem: I see a need for this logic in next phase..so better to keep it.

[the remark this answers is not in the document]
SionoiS: I don't want to loose @chaitanyaprem work. We will use it later.

waku/factory/external_config.nim
```diff
@@ -357,6 +357,32 @@ type WakuNodeConf* = object
...
+      name: "store-sync-interval"
+    .}: int64
+
+    storeSyncRelayJitter* {.
+      desc: "Time offset to account for message propagation jitter. In seconds.",
+      defaultValue: 20,
+      name: "store-sync-relay-jitter"
+    .}: int64
+
+    storeSyncMaxPayloadSize* {.
```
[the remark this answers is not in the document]
chaitanyaprem: Good point..maybe these can be hidden config..removed from help as well.

waku/node/waku_node.nim
SionoiS: Ah yeah totally forgot about that.

waku/factory/external_config.nim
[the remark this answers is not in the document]
jm-clius: Indeed. Not against experimental config items, but I would hide them as `experimental` sub-commands: https://github.com/status-im/nim-confutils?tab=readme-ov-file#introduction

waku/node/waku_node.nim
```diff
@@ -193,6 +195,158 @@ proc connectToNodes*(
...
+  let catchRes = catch:
+    node.switch.mount(node.wakuSync, protocolMatcher(WakuSyncCodec))
+  if catchRes.isErr():
+    return err("mounting waku sync to the switch failed: " & catchRes.error.msg)
+
+  if node.started:
+    node.wakuSync.start()
+
+  if not node.wakuArchive.isNil():
```
Ivansete-status, line 315: Let's add a comment to clarify the purpose of the following block. I just added one but surely you will give a more sensible one :P
```suggestion
  if not node.wakuArchive.isNil():
  ## Let's initialize the btree set that contains the known messages
```

waku/node/waku_node.nim
```diff
@@ -193,6 +195,158 @@ proc connectToNodes*(
...
+      includeData: true,
+      cursor: none(ArchiveCursor),
+      startTime: some(starTime),
+      endTime: some(endTime),
+      pageSize: 100,
+    )
+
+    while true:
+      let catchable = catch:
+        waitfor node.wakuArchive.findMessages(query)
```
Ivansete-status, line 329: We cannot use `waitFor` at this level.
`waitFor` should only be used in the highest levels of the code because that could cause deadlocks.
```suggestion
        await node.wakuArchive.findMessages(query)
```

waku/node/waku_node.nim
```diff
@@ -193,6 +195,158 @@ proc connectToNodes*(
...
+
+      let res =
+        if catchable.isErr():
+          return err(catchable.error.msg)
+        else:
+          catchable.get()
+
+      let response = res.valueOr:
+        return err($error)
```
Ivansete-status, line 338:
```suggestion
        return err("error while fetching previous messages for waku sync: " & $error)
```

waku/waku_sync/common.nim
```diff
@@ -0,0 +1,34 @@
...
+  {.push raises: [].}
+
+import std/[options], chronos, libp2p/peerId
+import ../waku_core
+
+const
+  DefaultSyncInterval*: Duration = 1.hours
+  WakuSyncCodec* = "/vac/waku/sync/1.0.0"
+  DefaultMaxFrameSize* = 153600 #TODO change to something sensible
+  DefaultGossipSubJitter*: Duration = 20.seconds
```
Ivansete-status, line 13: I've commented it somewhere else but I think this could be less
```suggestion
  DefaultGossipSubJitter*: Duration = 1.seconds
```

waku/waku_sync/common.nim
```diff
@@ -0,0 +1,34 @@
...
+else:
+  {.push raises: [].}
+
+import std/[options], chronos, libp2p/peerId
+import ../waku_core
+
+const
+  DefaultSyncInterval*: Duration = 1.hours
+  WakuSyncCodec* = "/vac/waku/sync/1.0.0"
+  DefaultMaxFrameSize* = 153600 #TODO change to something sensible
```
Ivansete-status, line 12: Dumb question: what is this value for?

waku/waku_sync/protocol.nim
```diff
@@ -0,0 +1,311 @@
...
+
+  info "WakuSync protocol initialized"
+
+  return ok(sync)
+
+proc periodicSync(self: WakuSync, callback: TransferCallback) {.async.} =
+  debug "periodic sync initialized", interval = $self.syncInterval
+
+  while true:
+    await sleepAsync(self.syncInterval)
```
Ivansete-status, line 230: Should it better be at the end of the loop?

waku/waku_sync/protocol.nim
```diff
@@ -0,0 +1,311 @@
...
+
+    debug "periodic sync started"
+
+    let (hashes, peer) = (await self.sync()).valueOr:
+      debug "sync failed", error
```
Ivansete-status, line 235:
```suggestion
      error "sync failed", error = $error
```

waku/waku_sync/protocol.nim
```diff
@@ -0,0 +1,311 @@
...
+      continue
+
+    if hashes.len > 0:
+      (await callback(hashes, peer.peerId)).isOkOr:
+        debug "transfer callback failed", error
```
Ivansete-status, line 240:
```suggestion
        error "transfer callback failed", error = $error
```

waku/waku_sync/protocol.nim
```diff
@@ -0,0 +1,311 @@
...
+          continue
+        else:
+          break
+
+      if elements.len == 0:
+        break
+
+      for (hash, timestamp) in elements:
+        self.storage.erase(timestamp, hash).isOkOr:
+          trace "storage erase failed", timestamp, hash, error
```
Ivansete-status, line 281:
```suggestion
          error "storage erase failed", timestamp, msg_hash, error
```

waku/waku_sync/protocol.nim
```diff
@@ -0,0 +1,311 @@
...
+          await sleepAsync(30.seconds)
+          continue
+        else:
+          break
+
+      if elements.len == 0:
+        break
+
+      for (hash, timestamp) in elements:
+        self.storage.erase(timestamp, hash).isOkOr:
```
Ivansete-status, line 280: Cool, I found where we delete data :D
I think it could be interesting to implement a tiny unit-test where we ingest many messages, then delete, and then validate that the occupied memory didn't grow.

waku/waku_sync/protocol.nim
```diff
@@ -0,0 +1,311 @@
...
+    debug "periodic prune done", storageSize = self.storage.len
+
+proc start*(self: WakuSync) =
+  self.started = true
+  self.pruneStart = getNowInNanosecondTime()
+
+  if self.transferCallBack.isSome() and self.syncInterval > ZeroDuration:
+    self.periodicSyncFut = self.periodicSync(self.transferCallBack.get())
+
+  if self.pruneCallBack.isSome() and self.syncInterval > ZeroDuration:
```
Ivansete-status, line 299: I think there should always be a pruneCallback. If not, mem leak

waku/waku_sync/raw_bindings.nim
```diff
@@ -0,0 +1,352 @@
...
+    have_hashes = toBufferSeq(myResult.have_ids_len, myResult.have_ids)
+  if myResult.need_ids_len > 0:
+    need_hashes = toBufferSeq(myResult.need_ids_len, myResult.need_ids)
+
+  trace "have and need hashes ",
+    have_count = have_hashes.len, need_count = need_hashes.len
+
+  for i in 0 .. have_hashes.len - 1:
+    var hash = toWakuMessageHash(have_hashes[i])
+    trace "have hashes ", index = i, hash = hash.to0xHex()
```
Ivansete-status, line 319: The following suggestion applies elsewhere ;P

```suggestion
    trace "have hashes ", index = i, msg_hash = hash.to0xHex()
```

Ivansete-status (commented):
Very smart PR, congrats! :partying_face: 
I've added some comments so far that I hope you find useful
Cheers

waku/factory/external_config.nim
```diff
@@ -357,6 +357,34 @@ type WakuNodeConf* = object
...
+      hidden,
+      desc: "Time offset to account for message propagation jitter. In seconds.",
+      defaultValue: 20,
+      name: "store-sync-relay-jitter"
+    .}: int64
+
+    storeSyncMaxPayloadSize* {.
+      hidden,
+      desc:
+        "Max size in bytes of the inner negentropy payload. Cannot be less than 5K, 0 is unlimited.",
```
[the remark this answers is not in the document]
SionoiS: Recompiling for each simulation is annoying so I added those params but hidden because they will probably be turned into `const`s once we find good values.

waku/factory/external_config.nim
```diff
@@ -357,6 +357,34 @@ type WakuNodeConf* = object
...
+    storeSyncInterval* {.
+      desc: "Interval between store sync attempts. In seconds.",
+      defaultValue: 3600, # 1 hours
+      name: "store-sync-interval"
+    .}: int64
+
+    storeSyncRelayJitter* {.
+      hidden,
+      desc: "Time offset to account for message propagation jitter. In seconds.",
+      defaultValue: 20,
```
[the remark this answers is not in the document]
SionoiS: Yes too long, the correct value would be something like the 95th percentile of message propagation time in the network but IDK what that value is yet.

waku/waku_sync/protocol.nim
SionoiS: No? It would trigger a sync when initialized. We planned a feature for this purpose, some kind of Store resume.

waku/waku_sync/protocol.nim
SionoiS: Ah yes good idea!

waku/waku_sync/protocol.nim
SionoiS: I agree but testing with pruning is very annoying so I made it configurable.

waku/node/waku_node.nim
```diff
@@ -193,6 +195,158 @@ proc connectToNodes*(
...
+          endTime: some(pruneStop),
+          pageSize: 100,
+        )
+
+        let catchable = catch:
+          await node.wakuArchive.findMessages(query)
+
+        let res =
+          if catchable.isErr():
+            return err(catchable.error.msg)
```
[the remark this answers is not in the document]
SionoiS: In this case the caller already add context when an error occurs.

tests/waku_sync/test_protocol.nim
```diff
@@ -0,0 +1,477 @@
...
+      wakuSync.ingessMessage(DefaultPubsubTopic, msg3)
+      wakuSync.ingessMessage(DefaultPubsubTopic, msg4)
+
+      await sleepAsync(3.seconds)
+
+      check:
+        wakuSync.storageSize == 0
+
+  suite "Bindings":
+    asyncTest "test c integration":
```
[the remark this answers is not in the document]
SionoiS: I'll extract the test to another file and clean it up.

tests/waku_sync/sync_utils.nim
```diff
@@ -0,0 +1,27 @@
...
+  let
+    peerManager = PeerManager.new(switch)
+    proto = WakuSync.new(
+      peerManager = peerManager,
+      relayJitter = 0.seconds,
+      transferCB = transfer,
+      pruneCB = prune,
+      syncInterval = interval,
+    )
+  assert proto != nil
```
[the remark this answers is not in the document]
SionoiS: It's not, I'll remove it.

waku/waku_sync/common.nim
SionoiS: TBD

waku/waku_sync/protocol.nim
```diff
@@ -0,0 +1,311 @@
...
+  if msg.ephemeral:
+    return
+
+  let msgHash: WakuMessageHash = computeMessageHash(pubsubTopic, msg)
+
+  trace "inserting message into storage ",
+    hash = msgHash.toHex(), timestamp = msg.timestamp
+
+  if self.storage.insert(msg.timestamp, msgHash).isErr():
+    debug "failed to insert message ", hash = msgHash.toHex()
```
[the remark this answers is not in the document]
SionoiS: Same here, IDK if it would be better to use trace since there may be a lot of logs.

waku/node/waku_node.nim
SionoiS: I extracted this logic into it's own proc called `initFillStorage`

waku/waku_sync/common.nim
SionoiS: It is used to control the size of the negentropy payload. Lower values leads to more range checks and network round-trips. It's configurable but IDK if we will use anything else than default.

waku/waku_sync/protocol.nim
SionoiS: :thinking: any ideas how to check memory size?

SionoiS:
I made some changes;
- Bindings now share the same interface for the main `Negentropy` object 
- Added more retry mechanisms
- Move callbacks out of `wakunode`
- Cleaned-up binding tests

waku/waku_sync/protocol.nim
```diff
@@ -0,0 +1,468 @@
...
+
+proc new*(
+    T: type WakuSync,
+    peerManager: PeerManager,
+    maxFrameSize: int = DefaultMaxFrameSize,
+    syncInterval: timer.Duration = DefaultSyncInterval,
+    relayJitter: Duration = DefaultGossipSubJitter,
+    pruning: bool,
+    wakuArchive: WakuArchive,
+    wakuStoreClient: WakuStoreClient,
```
jm-clius, line 330: Doesn't this create the kind of interdependency that you were trying to avoid? Even if the `WakuSync` object is not composed with `wakuArchive` and `wakuStoreClient`, it now does need to be aware of these during construction. In my mind Store-Sync is a cleaner abstraction layer that understands the interrelationship between these components without forcing interdependencies into any of these. One way in which this could be achieved without defining a `WakuStoreSync` protocol, would be to simply define the handlers in a separate module and having waku_node install the handlers rather than `WakuSync` doing so during construction.

jm-clius (approved):
Indeed much cleaner than having everything done in the node. Thanks! I have added a comment below, because I think we're deepening interdependencies by not creating an abstraction layer that makes the separation of concerns clear. However, this is just a suggestion and I'm happy to leave final architecture to you. :)

waku/waku_sync/protocol.nim
SionoiS: > Doesn't this create the kind of interdependency that you were trying to avoid?

Yes it's meh, but it always been like that, status-quo didn't change. The only benefit would be a cleaner `wakunode` and maybe an easier refactor in the future?

> Even if the WakuSync object is not composed with wakuArchive and wakuStoreClient, it now needs to be aware of these during construction.

Until we have channels or similar, no matter at which layer of abstraction we "construct" it this is always going to be the case.

> In my mind Store-Sync is a cleaner abstraction layer that understands the interrelationship between these components without forcing interdependencies into any of the subcomponents. One way in which this could be achieved without defining a WakuStoreSync protocol, would be to simply define the handlers in a separate module and having waku_node install the handlers rather than WakuSync doing so during construction.

I really don't understand, you should build it and show me. I would love for a better way to exist.

SionoiS:
CI is red but I can reproduce...

waku/waku_sync/protocol.nim
Ivansete-status: > 🤔 any ideas how to check memory size?

We can try with `getOccupiedMem()`, from `system` module, which gives the current memory in bytes.
We can measure at three points:
1. At the beginning
2. After allocating memory 
3. At the end, to make sure the memory is back to the level at 1.

SionoiS: `getOccupiedMem()` seams to count only what is managed by the GC. Since we use Negentropy C++ lib that memory is managed on the C++ side. Any other ideas @Ivansete-status ? IDK what to do...

SionoiS:
After discussing with @jm-clius I will have to add a `SyncRange` and a `SyncInterval` so that we can sync the last hour every 5m for example.

waku/waku_sync/protocol.nim
NagyZoltanPeter: What do we target here with mem size check?
If this lib is optimized it might not shrink memory that way. Also it probably preallocates to some extent as it is pretty costly to grow in small chunks also to shrink. Maybe we need to review the library itself?

Ivansete-status (commented):
Great one! Congrats for it! :raised_hands: 
I've added a bunch of comments, some of them are pedantic and others are potentially more interesting :)
Very insightful PR

waku/waku_core/time.nim
```diff
@@ -33,3 +33,16 @@ template nanosecondTime*(collector: Gauge, body: untyped) =
...
+  let timeStr = $time
+  var timestamp: Timestamp = time
+
+  if timeStr.len() > 16:
+    timestamp = Timestamp(time div Timestamp(1_000_000_000))
+  elif timeStr.len() < 16 and timeStr.len() > 13:
+    timestamp = Timestamp(time div Timestamp(1_000_000))
+  elif timeStr.len() > 10:
+    timestamp = Timestamp(time div Timestamp(1000))
+  return timestamp ]#
```
gabrielmer, line 48: This will eventually be used? or we can get rid of it?

waku/waku_sync/storage_manager.nim
```diff
@@ -0,0 +1,76 @@
+# Unused yet.
+#[ when (NimMajor, NimMinor) < (1, 4):
+  {.push raises: [Defect].}
+else:
+  {.push raises: [].}
+
+import std/[times, tables, options], chronicles, chronos, stew/results
+
+import ./raw_bindings, ../waku_core/time
+
```
gabrielmer, line 10: Same here, is this planned to be used?

gabrielmer (commented):
Woow, thanks so much! Amazing work!

Started reviewing and will be adding batches of comments/questions :)

waku/waku_sync/protocol.nim
```diff
@@ -0,0 +1,475 @@
...
+  pruneStart: Timestamp # Last pruning start timestamp
+  pruneOffset: timer.Duration # Offset to prune a bit more than necessary. 
+
+  periodicSyncFut: Future[void]
+  periodicPruneFut: Future[void]
+
+proc storageSize*(self: WakuSync): int =
+  self.storage.len
+
+proc ingessMessage*(self: WakuSync, pubsubTopic: PubsubTopic, msg: WakuMessage) =
```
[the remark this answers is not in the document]
SionoiS: > Is ingess a word? shouldn't be ingest or ingress?

:rofl: did a search and yes suppose to be `ingress`

waku/waku_core/time.nim
SionoiS: It's planned to be used. So we keep.

waku/waku_sync/codec.nim
```diff
@@ -0,0 +1,65 @@
...
+
+import std/options, stew/arrayops
+import ../common/protobuf, ../waku_core, ./common
+
+proc encode*(req: SyncPayload): ProtoBuffer =
+  var pb = initProtoBuffer()
+
+  if req.rangeStart.isSome() and req.rangeEnd.isSome():
+    pb.write3(31, req.rangeStart.get())
+    pb.write3(32, req.rangeEnd.get())
```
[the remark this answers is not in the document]
SionoiS: Those are not logically equivalent. `rangeStart` and `rangeEnd` must be set.

waku/waku_sync/protocol.nim
```diff
@@ -0,0 +1,475 @@
...
+  topics = "waku sync"
+
+type WakuSync* = ref object of LPProtocol
+  storage: NegentropyStorage
+  peerManager: PeerManager
+  maxFrameSize: int
+
+  syncInterval: timer.Duration # Time between each syncronisation attempt
+  relayJitter: Duration # Time delay until all messages are mostly received network wide
+  transferCallBack: Option[TransferCallback] # Callback for message transfer.
```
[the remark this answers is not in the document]
SionoiS: ATM it is only set IF we have access to `wakuArchive` and `wakusStoreClient`.

We could change it so that the setup fail instead?

waku/waku_sync/protocol.nim
```diff
@@ -0,0 +1,475 @@
...
+      debug "sync session ended",
+        server = self.peerManager.switch.peerInfo.peerId, client = conn.peerId, error
+
+      #TODO send error code and desc to client
+      return
+
+    if hashes.len > 0 and self.transferCallBack.isSome():
+      let callback = self.transferCallBack.get()
+
+      (await callback(hashes, conn.peerId)).isOkOr:
```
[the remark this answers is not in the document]
SionoiS: Callbacks are :hankey: . Yes we can use private procs but `WakuSync` would have to reference `WakuArchive` and  `WakuStoreClient` which is also :hankey: 

It goes back to my discussion with @jm-clius IMO there's no way to abstract away archive and store client....

What flavor of :hankey: do we want?

waku/waku_sync/protocol.nim
```diff
@@ -0,0 +1,475 @@
...
+  periodicPruneFut: Future[void]
+
+proc storageSize*(self: WakuSync): int =
+  self.storage.len
+
+proc ingessMessage*(self: WakuSync, pubsubTopic: PubsubTopic, msg: WakuMessage) =
+  if msg.ephemeral:
+    return
+
+  let msgHash: WakuMessageHash = computeMessageHash(pubsubTopic, msg)
```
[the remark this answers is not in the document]
SionoiS: I'll overload the proc with a version that you can pass the hash.

waku/waku_sync/protocol.nim
```diff
@@ -0,0 +1,475 @@
...
+      peer
+
+  let connOpt = await self.peerManager.dialPeer(peer, WakuSyncCodec)
+
+  let conn: Connection = connOpt.valueOr:
+    return err("Cannot establish sync connection")
+
+  let hashes: seq[WakuMessageHash] = (await self.request(conn)).valueOr:
+    debug "sync session ended",
+      server = self.peerManager.switch.peerInfo.peerId, client = conn.peerId, error
```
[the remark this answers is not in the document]
SionoiS: Not in this case this is Negentropy specific terminology. The client is define as the first payload sender.

WIP spec -> https://github.com/waku-org/specs/blob/feat--waku-sync/standards/core/sync.md

waku/waku_sync/protocol.nim
```diff
@@ -0,0 +1,475 @@
...
+  sync.initProtocolHandler()
+
+  if pruning:
+    sync.initPruningHandler(wakuArchive)
+
+  sync.initTransferHandler(wakuArchive, wakuStoreClient)
+
+  let res = await sync.initFillStorage(wakuArchive)
+  if res.isErr():
+    error "initial storage filling failed", error = res.error
```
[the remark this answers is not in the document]
SionoiS: For testing purposes we need to be able to setup without archive or store client.

waku/waku_sync/protocol.nim
```diff
@@ -0,0 +1,475 @@
...
+
+proc periodicPrune(self: WakuSync, callback: PruneCallback) {.async.} =
+  debug "periodic prune initialized", interval = $self.syncInterval
+
+  await sleepAsync(self.syncInterval)
+
+  var pruneStop = getNowInNanosecondTime()
+
+  while true:
+    await sleepAsync(self.syncInterval)
```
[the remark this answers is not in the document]
SionoiS: Not in this case. I added comments to clarify the process.

waku/waku_sync/protocol.nim
```diff
@@ -0,0 +1,475 @@
...
+      let query = ArchiveQuery(
+        includeData: true,
+        cursor: archiveCursor,
+        startTime: some(pruneStart),
+        endTime: some(pruneStop),
+        pageSize: 100,
+      )
+
+      let catchable = catch:
+        await wakuArchive.findMessages(query)
```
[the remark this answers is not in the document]
SionoiS: I agree it's not the best but it's fine for now. We plan on improving the Negentropy storage in the future so that querying the DB would not be necessary.

waku/node/waku_node.nim
```diff
@@ -190,6 +192,42 @@ proc connectToNodes*(
...
+      peerManager = node.peerManager,
+      maxFrameSize = maxFrameSize,
+      syncInterval = syncInterval,
+      relayJitter = relayJitter,
+      pruning = enablePruning,
+      wakuArchive = node.wakuArchive,
+      wakuStoreClient = node.wakuStoreClient,
+    )
+  ).valueOr:
+    return err("initialization failed: " & error)
```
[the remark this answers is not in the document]
SionoiS: In this case the caller give us the context we need.

`failed to mount waku sync protocol: initialization failed: negentropy storage creation failed`
VS
`failed to mount waku sync protocol: wakus sync initialization failed: waku sync error negentropy storage creation failed`

waku/waku_sync/protocol.nim
```diff
@@ -0,0 +1,475 @@
...
+  let peer =
+    if peerInfo.isSome():
+      peerInfo.get()
+    else:
+      let peer: RemotePeerInfo = self.peerManager.selectPeer(WakuSyncCodec).valueOr:
+        return err("No suitable peer found for sync")
+
+      peer
+
+  let connOpt = await self.peerManager.dialPeer(peer, WakuSyncCodec)
```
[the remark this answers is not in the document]
SionoiS: We may dial the same peer if we only have only one but unlikely if we have more. AFAIK we select a random sync peer since we don't specify which one we want.

SionoiS:
Let me know when I can rebase I don't want to mess your reviews.

waku/waku_core/time.nim
```diff
@@ -33,3 +33,16 @@ template nanosecondTime*(collector: Gauge, body: untyped) =
     metrics.set(collector, nowInUnixFloat() - start)
   else:
     body
+
+# Unused yet
+#[ proc timestampInSeconds*(time: Timestamp): Timestamp =
```
NagyZoltanPeter, line 38: If we think (suppose) the value will be an actual time value (not a diff), we don't need to convert it to string. Isn't that cheaper to just compare to some numeric value? Am I miss something? Btw why we expect the value may vary between milli/micro and nano secs?

waku/waku_sync/protocol.nim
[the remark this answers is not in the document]
NagyZoltanPeter: I feel it, but still it can be an option to move out these refs out of this protocol into a separate callback setup place. That would eliminate the need for refs of WakuArchive and WakuStoreClient.
Now they need to be provided with new*.
In other words by separating setting up the callback from the protocol will move away the decision how that is implemented. Also enables mocking it for testing. WDYT?

waku/node/waku_node.nim
```diff
@@ -249,12 +287,19 @@ proc registerRelayDefaultHandler(node: WakuNode, topic: PubsubTopic) =
 
     await node.wakuArchive.handleMessage(topic, msg)
 
+  proc syncHandler(topic: PubsubTopic, msg: WakuMessage) =
```
NagyZoltanPeter, line 290: Just a question, is there any particular reason it cannot be async? (or just because it is call ~sync~Handler :-) ?)

NagyZoltanPeter (commented):
Highest level, pretty, neat and awesome feature. Thank you!!!
I will love to see it working.
Left some observations, mostly questions.

NagyZoltanPeter:
> Let me know when I can rebase I don't want to mess your reviews.
Left some comments, so from my side you can go with it.

waku/waku_sync/protocol.nim
```diff
@@ -0,0 +1,475 @@
...
+    self: WakuSync, peerInfo: Option[RemotePeerInfo] = none(RemotePeerInfo)
+): Future[Result[(seq[WakuMessageHash], RemotePeerInfo), string]] {.async, gcsafe.} =
+  let peer =
+    if peerInfo.isSome():
+      peerInfo.get()
+    else:
+      let peer: RemotePeerInfo = self.peerManager.selectPeer(WakuSyncCodec).valueOr:
+        return err("No suitable peer found for sync")
+
+      peer
```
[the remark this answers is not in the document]
SionoiS: It seams like `.valueOr` for options works sometimes... I didn't know it resided in `libp2p/utility` that explain why! Thanks.

waku/waku_sync/session.nim
```diff
@@ -0,0 +1,236 @@
...
+      frameSize: self.frameSize,
+      payload: payload,
+    )
+  )
+
+  return ok(res)
+
+proc serverReconcile*(
+    self: Received[ServerSync]
+): Result[Result[Reconciled[ServerSync], Completed[ServerSync]], string] =
```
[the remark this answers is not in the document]
SionoiS: yeah looking at it now it's confusing... I'll try to improve this.

waku/waku_sync/common.nim
```diff
@@ -0,0 +1,34 @@
...
+  ): Future[Result[void, string]] {.async: (raises: []), closure.}
+
+  PruneCallback* = proc(
+    startTime: Timestamp, endTime: Timestamp, cursor = none(WakuMessageHash)
+  ): Future[
+    Result[(seq[(WakuMessageHash, Timestamp)], Option[WakuMessageHash]), string]
+  ] {.async: (raises: []), closure.}
+
+  SyncPayload* = object
+    rangeStart*: Option[uint64]
```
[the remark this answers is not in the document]
SionoiS: Will do! Enforcement like this are good.

waku/waku_sync/protocol.nim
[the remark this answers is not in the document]
SionoiS: It will still work but won't be able to sync right away since the storage will not contain any elements to sync.

Maybe warn? It's not a big deal if we sync often.

waku/waku_sync/protocol.nim
```diff
@@ -0,0 +1,475 @@
...
+    relayJitter: relayJitter,
+    pruneOffset: syncInterval div 100,
+  )
+
+  sync.initProtocolHandler()
+
+  if pruning:
+    sync.initPruningHandler(wakuArchive)
+
+  sync.initTransferHandler(wakuArchive, wakuStoreClient)
```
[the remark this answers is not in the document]
SionoiS: The callback option would be empty.

I'll add comments.

waku/waku_sync/protocol.nim
```diff
@@ -0,0 +1,475 @@
...
+
+      hashes = res[0]
+      peer = res[1]
+      break
+
+    if hashes.len > 0:
+      tries = 3
+      while true:
+        (await callback(hashes, peer.peerId)).isOkOr:
+          error "transfer callback failed", error = $error
```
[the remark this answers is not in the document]
SionoiS: True maybe log error when done trying?

waku/waku_sync/protocol.nim
```diff
@@ -0,0 +1,475 @@
...
+    await sleepAsync(self.syncInterval)
+
+    debug "periodic sync started"
+
+    var
+      hashes: seq[WakuMessageHash]
+      peer: RemotePeerInfo
+      tries = 3
+
+    while true:
```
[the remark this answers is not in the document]
SionoiS: It's called break dance. :rofl:

Does Nim have labels for loop control `break: myloop` or `continue: myloop` ?

waku/waku_core/time.nim
SionoiS: I don't know. @chaitanyaprem wrote this part maybe he can answer.

waku/node/waku_node.nim
SionoiS: :thinking: I assumed that async was not needed but maybe it would be better so that not blocking can occurs.

waku/waku_sync/protocol.nim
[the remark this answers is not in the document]
NagyZoltanPeter: Yes, it does not have, while using `while true:` and breaks seems to me controversial. 
Once upon a time there was an article, [Goto statement considered harmful](https://homepages.cwi.nl/~storm/teaching/reader/Dijkstra68.pdf), which is resonates with this, especially with two embereded `while true:`... reading it is not easy.

waku/waku_sync/protocol.nim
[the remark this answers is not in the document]
NagyZoltanPeter: Yes. Is there threshold when we can consider giving up probing?

waku/waku_sync/protocol.nim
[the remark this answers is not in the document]
NagyZoltanPeter: Then it sounds not an error, just would warn the log reader and explain if it will behave different a bit.

waku/waku_sync/protocol.nim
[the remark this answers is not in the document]
SionoiS: IDK how loops can be avoided. I'm open to suggestion.

waku/waku_sync/protocol.nim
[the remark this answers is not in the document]
SionoiS: No other protocol has retry mechanism it's up to us the define good params. 3 tries was chosen arbitrarily.

waku/waku_sync/protocol.nim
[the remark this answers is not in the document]
SionoiS: > In other words by separating setting up the callback from the protocol will move away the decision how that is implemented.

The only other place would be in `wakunode` and that's not good either. I made some changes but IDK what to do.

waku/waku_sync/protocol.nim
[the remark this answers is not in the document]
Ivansete-status: > We could change it so that the setup fail instead?

I think it makes sense to fail as soon as possible and completely stop the `wakunode2` at the beginning if we consider a wrong initial setup.

> edit: we need to be able to not set the callbacks for testing purposes.

Maybe we can use `isNil()` instead? Not a blocking deal though :)

waku/waku_sync/protocol.nim
```diff
@@ -0,0 +1,512 @@
...
+proc new*(
+    T: type WakuSync,
+    peerManager: PeerManager,
+    maxFrameSize: int = DefaultMaxFrameSize,
+    syncRange: timer.Duration = DefaultSyncRange,
+    syncInterval: timer.Duration = DefaultSyncInterval,
+    relayJitter: Duration = DefaultGossipSubJitter,
+    wakuArchive: WakuArchive,
+    wakuStoreClient: WakuStoreClient,
+    pruneCallback: Option[PruneCallback] = none(PruneCallback),
```
Ivansete-status, line 348: It seems that the `pruneCallback` is never called. Maybe we can remove it?

waku/waku_sync/protocol.nim
```diff
@@ -0,0 +1,512 @@
...
+    transferCallback: Option[TransferCallback] = none(TransferCallback),
```
Ivansete-status, line 349: The `transferCallback` is always unset or set to None. This parameter seems to be not needed

waku/waku_sync/protocol.nim
```diff
@@ -0,0 +1,512 @@
...
+
+  sync.transferCallBack = transferCallback
+
+  if sync.transferCallBack.isNone():
+    let res = sync.createTransferCallback(wakuArchive, wakuStoreClient)
+
+    if res.isErr():
+      error "transfer callback creation error", error = res.error
+    else:
+      sync.transferCallBack = some(res.get())
```
Ivansete-status, line 384: The following seems not needed:
```suggestion
  if wakuArchive.isNil():
    return err("waku archive unavailable")

  if wakuStoreClient.isNil():
    return err("waku store client unavailable")
```

waku/waku_sync/protocol.nim
```diff
@@ -0,0 +1,512 @@
...
+          break
+
+      hashes = res[0]
+      peer = res[1]
+      break
+
+    if hashes.len > 0:
+      tries = 3
+      while true:
+        (await callback(hashes, peer.peerId)).isOkOr:
```
Ivansete-status, line 426: We don't need to pass a callback. I'd suggest having a private proc `performTransfer`.
```suggestion
        (await performTransfer(hashes, peer.peerId)).isOkOr:
```

waku/waku_sync/protocol.nim
```diff
@@ -0,0 +1,512 @@
...
+      for (hash, msg) in response.hashes.zip(response.messages):
+        (hash, msg.timestamp)
+
+    let cursor = response.cursor
+
+    return ok((elements, cursor))
+
+  return ok(callback)
+
+proc createTransferCallback(
```
Ivansete-status, line 256: I think we don't need neither this proc or `createPruneCallback`. Having the local private procs will suffice

Ivansete-status (commented):
Thanks for it :100: ! I think we can have some more simplification :P

waku/waku_sync/protocol.nim
SionoiS: Yes it is at L466

waku/waku_sync/protocol.nim
SionoiS: For tests we can mock the callbacks but for normal operation we just pass archive and store client so that the callback are created automatically.

waku/waku_sync/protocol.nim
SionoiS: We can't error in case archive and store client are nil otherwise we would have to include those in ours tests.

waku/waku_sync/protocol.nim
SionoiS: Only true if we don't care about mocking the callback for testing. Lets keep the callback for QA.

waku/waku_sync/protocol.nim
SionoiS: see my reasoning above

waku/waku_sync/protocol.nim
```diff
@@ -0,0 +1,512 @@
...
+
+    reconciled = (?received.clientReconcile(hashes)).valueOr:
+      let completed = error # Result[Reconciled, Completed]
+
+      ?await completed.clientTerminate()
+
+      debug "sync session ended gracefully",
+        client = self.peerManager.switch.peerInfo.peerId, server = conn.peerId
+
+      return ok(hashes)
```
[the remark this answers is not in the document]
SionoiS: L125 either return `Completed` or `Reconciled`. in the latter case it loops again.

waku/waku_sync/protocol.nim
```diff
@@ -0,0 +1,512 @@
...
+
+  return (syncStart, syncEnd)
+
+proc request(
+    self: WakuSync, conn: Connection
+): Future[Result[seq[WakuMessageHash], string]] {.async.} =
+  let (syncStart, syncEnd) = calculateRange(self.relayJitter)
+
+  let initialized =
+    ?clientInitialize(self.storage, conn, self.maxFrameSize, syncStart, syncEnd)
```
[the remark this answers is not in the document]
SionoiS: Since `request()` return a result and is internal only I wanted to avoid double logging.

In cases where `?` is used it is true that the caller is responsible for logging.

waku/waku_sync/protocol.nim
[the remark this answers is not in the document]
SionoiS: @gabrielmer AFAIK no other protocol close their connection either so it seams fine? :thinking: Maybe they all should.

Ivansete-status (approved):
LGTM! Thanks for the patience! :100:

PR #2404, opened 2024-02-07
chore: Simplify configuration for the waku network

alrevuelta:
# Description
Simplifies configuration, extracting  `cluster-id=1` parameters into a file. When `cluster-id=1` is configured, then all parameters are loaded without having to specify them all. Now someone wanting to join `TheWakuNetwork` can do so with just:
```
./build/wakunode2 --cluster-id=1 --rln-relay-eth-client-address=$YOU_ETH_NODE
```

Or if you want to run just one shard:
```
./build/wakunode2 --cluster-id=1 --shard=5 --rln-relay-eth-client-address=$YOU_ETH_NODE
```

This should simplify [this](https://github.com/waku-org/nwaku-compose/blob/master/run_node.sh#L57-L102) monster.

### DIFF
--- a/apps/wakunode2/external_config.nim
+++ b/apps/wakunode2/external_config.nim
@@ -34,6 +34,8 @@ type ProtectedTopic* = object
   topic*: string
   key*: secp256k1.SkPublicKey
 
+type ShardIdx = distinct uint16 
+
 type StartUpCommand* = enum
     noCommand # default, runs waku
     generateRlnKeystore # generates a new RLN keystore
@@ -255,6 +257,11 @@ type
       pubsubTopics* {.
         desc: "Default pubsub topic to subscribe to. Argument may be repeated."
         name: "pubsub-topic" .}: seq[string]
+
+      shards* {.
+        desc: "Shards index to subscribe to [0..MAX_SHARDS-1]. Argument may be repeated."
+        defaultValue: @[]
+        name: "shard" .}: seq[ShardIdx]
 
       contentTopics* {.
         desc: "Default content topic to subscribe to. Argument may be repeated."
@@ -576,6 +583,15 @@ proc parseCmdArg*(T: type Option[int], p: string): T =
   except CatchableError:
     raise newException(ValueError, "Invalid number")
 
+proc completeCmdArg*(T: type ShardIdx, val: string): seq[ShardIdx] =
+  return @[]
+
+proc parseCmdArg*(T: type ShardIdx, p: string): T =
+  try:
+    ShardIdx(parseInt(p))
+  except CatchableError:
+    raise newException(ValueError, "Invalid shard index")
+
 proc parseCmdArg*(T: type Option[uint], p: string): T =
   try:
     some(parseUint(p))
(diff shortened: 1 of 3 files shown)


apps/wakunode2/networks_config.nim
```diff
@@ -0,0 +1,40 @@
...
+  rlnRelayDynamic*: bool
+  rlnRelayBandwidthThreshold*: int
+  pubsubTopics*: seq[string]
+  discv5Discovery*: bool
+  discv5BootstrapNodes*: seq[string]
+
+# cluster-id=1
+# Cluster configuration corresponding to The Waku Network. Note that it
+# overrides existing cli configuration
+proc TheWakuNetworkConf*(T: type ClusterConf): ClusterConf =
```
Ivansete-status, line 20: Super nitpick, I know xD but comments tend to go below proc definition

apps/wakunode2/external_config.nim
```diff
@@ -576,6 +583,15 @@ proc parseCmdArg*(T: type Option[int], p: string): T =
   except CatchableError:
     raise newException(ValueError, "Invalid number")
 
+proc completeCmdArg*(T: type ShardIdx, val: string): seq[ShardIdx] =
+  return @[]
+
+proc parseCmdArg*(T: type ShardIdx, p: string): T =
+  try:
+    ShardIdx(parseInt(p))
```
Ivansete-status, line 591: Would it be possible to ensure that the given index is within the valid boundaries?

apps/wakunode2/wakunode2.nim
```diff
@@ -34,23 +59,48 @@ when isMainModule:
...
+    # Override configuration
+    conf.maxMessageSize = twnClusterConf.maxMessageSize
+    conf.clusterId = twnClusterConf.clusterId
+    conf.rlnRelay = twnClusterConf.rlnRelay
+    conf.rlnRelayEthContractAddress = twnClusterConf.rlnRelayEthContractAddress
+    conf.rlnRelayDynamic = twnClusterConf.rlnRelayDynamic
+    conf.rlnRelayBandwidthThreshold = twnClusterConf.rlnRelayBandwidthThreshold
+    conf.discv5Discovery = twnClusterConf.discv5Discovery
+    conf.discv5BootstrapNodes =
+      conf.discv5BootstrapNodes & twnClusterConf.discv5BootstrapNodes
```
Ivansete-status, line 86: Is this overriding the configuration or just appending made on purpose?

apps/wakunode2/external_config.nim
```diff
@@ -256,6 +258,11 @@ type
         desc: "Default pubsub topic to subscribe to. Argument may be repeated."
         name: "pubsub-topic" .}: seq[string]
 
+      shards* {.
```
jm-clius, line 261: Thinking that in future the ability to configure ranges may be needed for large subscriptions (e.g. `--shards:0..7,15,21..29`). For now repeated arg makes sense.

jm-clius (approved):
LGTM and definitely improves config.

Ivansete-status (approved):
LGTM! Thanks for it! :100: 
Just added some minor comments that I hope you find useful

apps/wakunode2/wakunode2.nim
alrevuelta: appending, just in case someone wants to add more bootstrap nodes on top of the existing ones.

apps/wakunode2/external_config.nim
alrevuelta: will address this in followup prs. don't want to just add a `MaxAmountShards` here but enforce it all over the place when a given cluster-id is chosen. and that may require some changes out of scope of this pr :)

apps/wakunode2/external_config.nim
alrevuelta: mm i would say node operators don't necessarily need to specify the exact shards but the number of them they want to run. unless they are interested in a specific shard because their content topic falls in there, but in that case i expect them to provide the contentTopic and not the shard.

(but yeap, current behaviour is --shard=5 means shard 5 and not 5 shards)

PR #2408, opened 2024-02-08
chore: automatically generating certs if not provided (Waku Canary)

### LINKED ISSUE #2349: bug: wakucanary requires key path to test client connections
### Problem
The `wakucanary` requires `--websocket-secure-key-path` to test client connections.

### Expected behavior
The flag should not be required.

#### Why

- https://github.com/status-im/infra-hq/issues/92#issue-1769478385
  - https://github.com/status-im/infra-hq/issues/92#issuecomment-1882637386
  - https://github.com/status-im/infra-hq/issues/92#issuecomment-1883040790

gabrielmer:
# Description
Automatically generating a self signed certificate in Waku Canary if not provided and WSS is used.

Notice that the keys and certificate are generated only if `OpenSSL` is installed. It comes preinstalled in Ubuntu, other common Linux distributions and MacOS.
If `openssl` command is not available and certificates are not provided, an error message is shown.

# Changes

- [x] generating key and certificate if needed and not provided

## How to test

1. Build Waku Canary: `make wakucanary`
3. Run it on a WSS address without setting certificates: `./build/wakucanary --address=/dns4/node-01.gc-us-central1-a.status.test.statusim.net/tcp/443/wss/p2p/16Uiu2HAmGDX3iAFox93PupVYaHa88kULGqMpJ7AEHGwj3jbMtt76 --protocol=relay`

## Issue

closes #2349

### DIFF
--- a/apps/wakucanary/certsgenerator.nim
+++ b/apps/wakucanary/certsgenerator.nim
@@ -1,4 +1,17 @@
-import osproc, os
+import 
+  osproc, 
+  os, 
+  httpclient, 
+  strutils
+
+proc getPublicIP(): string =
+  let client = newHttpClient()
+  try:
+    let response = client.get("http://api.ipify.org")
+    return response.body
+  except Exception as e:
+    echo "Could not fetch public IP: " & e.msg
+    return "127.0.0.1"
 
 # Function to generate a self-signed certificate
 proc generateSelfSignedCertificate*(certPath: string, keyPath: string) : int =
@@ -8,10 +21,16 @@ proc generateSelfSignedCertificate*(certPath: string, keyPath: string) : int =
     echo "OpenSSL is not installed or not in the PATH."
     return 1
 
+  let publicIP = getPublicIP()
+  
+  if publicIP != "127.0.0.1":
+    echo "Your public IP address is: ", publicIP 
+  
   # Command to generate private key and cert
   let 
     cmd = "openssl req -x509 -newkey rsa:4096 -keyout " & keyPath & " -out " & certPath &
-            " -sha256 -days 3650 -nodes -subj '/C=XX/ST=StateName/L=CityName/O=CompanyName/OU=CompanySectionName/CN=CommonNameOrHostname'"
+            " -sha256 -days 3650 -nodes -subj '/C=XX/ST=StateName/L=CityName/O=CompanyName/OU=CompanySectionName/CN=" & 
+            publicIP & "'"
     res = execCmd(cmd)
 
   if res == 0:
--- a/apps/wakucanary/certsgenerator.nim
+++ b/apps/wakucanary/certsgenerator.nim
@@ -1,0 +1,23 @@
+import osproc, os
+
+# Function to generate a self-signed certificate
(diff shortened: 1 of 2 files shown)


apps/wakucanary/wakucanary.nim
```diff
@@ -24,6 +26,7 @@ const ProtocolsTable = {
 }.toTable
 
 const WebSocketPortOffset = 1000
+const CertsDirectory = "./apps/wakucanary/certs"
```
vpavlin, line 29: We should make this configurable - what happens if you are not in the nwaku repo root? It will probably fail to write the key and cert, right?

vpavlin, line 29: I would either make this configurable, or used some more general path like `./certs` or `./self-signed-certs` - WDYT?

gabrielmer: Yes! Wasn't sure what was the best approach. Didn't want to make it configurable so we don't require another CLI flag (the whole idea is for this to happen under the hood/on its own). Also wanted to keep it inside the `wakucanary` directory so it doesn't mix with other stuff.

So I assumed that the instructions are being followed literally and it would be run from the root of the repo.

But makes sense, maybe it's better to store the certs in the directory where the executable is run, and avoid errors if it's run from a different place. Fixing it now :)

Ivansete-status (approved):
LGTM! Thanks!

My only concert about self-signed certificate is whether the `nwaku` node would accept a connection from that. I think it worth to try it, if not done already.

gabrielmer:
> LGTM! Thanks!
> 
> My only concert about self-signed certificate is whether the `nwaku` node would accept a connection from that. I think it worth to try it, if not done already.

Yes! It seems to be accepted :)

Ran wakucanary trying to contact one of our fleets with that certificate and the connection was successful

jm-clius (approved):
LGTM!

apps/wakucanary/certsgenerator.nim
```diff
@@ -0,0 +1,23 @@
...
+proc generateSelfSignedCertificate*(certPath: string, keyPath: string) : int =
+  
+  # Ensure the OpenSSL is installed
+  if findExe("openssl") == "":
+    echo "OpenSSL is not installed or not in the PATH."
+    return 1
+
+  # Command to generate private key and cert
+  let 
+    cmd = "openssl req -x509 -newkey rsa:4096 -keyout " & keyPath & " -out " & certPath &
```
chaitanyaprem, line 13: Just wondering if we leave the `CN=CommonNameOrHostname` like this, it may cause problems in case a node is validating the certificate. Can we pick this up from either system's domain name or public IP?

chaitanyaprem (approved):
LGTM except for the comment.

apps/wakucanary/certsgenerator.nim
gabrielmer: Good point!Jjust changed it :)

Now we use the public IP and if for some reason it cannot be retrieved then "127.0.0.1" is used.
With domain name is more problematic as we want for it to be automatic and not require any input from the user, so public IP is used instead.

WDYT?

PR #2409, opened 2024-02-08
chore: add stdef.h to libwaku.h

richard-ramos:
# Description
While generating the bindings for rust, I ran into the error `unknown type name 'size_t'`. This PR adds a header to `libwaku.h` to fix that problem. 

Curiously enough, in the c-bindings example we don't have such error because we include `stdio.h` which includes `stddef.h`


Ivansete-status (approved):
Great catch! Thanks for it!

PR #2410, opened 2024-02-09
feat: Add tool to benchmark RLN proof generation/verification

alrevuelta:
Add simple tool to benchmark RLN proof generation and verification.

Run as:
```
make benchmarks LOG_LEVEL=NOTICE
./build/benchmarks
```

Output:
```
Proof generation times: 85 milliseconds, 723 microseconds, and 313 nanoseconds
Proof verification times: 2 milliseconds, 707 microseconds, and 137 nanoseconds
```

### DIFF
--- a/apps/benchmarks/benchmarks.nim
+++ b/apps/benchmarks/benchmarks.nim
@@ -1,0 +1,47 @@
+import
+  math,
+  std/sequtils,
+  stew/results,
+  options,
+  ../../waku/waku_rln_relay/protocol_types,
+  ../../waku/waku_rln_relay/rln,
+  ../../waku/waku_rln_relay,
+  ../../waku/waku_rln_relay/conversion_utils,
+  ../../waku/waku_rln_relay/group_manager/static/group_manager
+
+import std/[times, os]
+
+proc main(): Future[string] {.async, gcsafe.} =
+  let rlnIns = createRLNInstance(20).get()
+  let credentials = toSeq(0 .. 1000).mapIt(membershipKeyGen(rlnIns).get())
+
+  let manager = StaticGroupManager(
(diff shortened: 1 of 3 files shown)


apps/benchmarks/benchmarks.nim
```diff
@@ -0,0 +1,44 @@
+import
+  math,
+  std/sequtils,
+  stew/results,
+  options,
+  ../../waku/waku_rln_relay/protocol_types,
```
rymnc, line 6: maybe we can just import waku_rln_relay?

apps/benchmarks/benchmarks.nim
```diff
@@ -0,0 +1,44 @@
...
+)
+
+manager.init()
+
+let data: seq[byte] = newSeq[byte](1024)
+
+var proofGenTimes: seq[times.Duration] = @[]
+var proofVerTimes: seq[times.Duration] = @[]
+
+for i in 0 .. 50:
```
rymnc, line 34: in a future iteration can we have this configurable?

rymnc (approved):
LGTM

apps/benchmarks/benchmarks.nim
```diff
@@ -0,0 +1,44 @@
...
+  var time = getTime()
```
Ivansete-status, line 35: In case we want to measue CPU-bounded tasks, it would be interesting to use `cpuTime` instead.

Ivansete-status (approved):
Awesome! Thanks!

apps/benchmarks/benchmarks.nim
alrevuelta: imho no need to complicate things. this is intended for devs so just change the value and recompile :)

PR #2413, opened 2024-02-09
feat(rln-relay-v2): nonce/messageId manager

rymnc:
# Description

This PR introduces a new submodule within `waku_rln_relay`, called the `NonceManager` which provides functionality to fetch the next given `nonce`/`messageId` on a per epoch basis which is required to generate rln-v2 proofs.

It also fixes the bad function signature of `appendRLNProof` which previously returned a boolean upon execution. It now returns a result.

# Changes

- [x] NonceManager
- [x] appendRLNProof sig change
- [x] NonceManager tests 

## How to test

1. `RLN_V2=true make -j16 test` is broken at the moment since the onchain/static group manager tests have not been updated yet. However, the NonceManager is independent of those changes, and can be run, and has not been guarded by the `rln_v2` feature flag upon running `make -j16 test` 

## Issue

closes https://github.com/waku-org/nwaku/issues/2415

### DIFF
--- a/waku/waku_rln_relay/nonce_manager.nim
+++ b/waku/waku_rln_relay/nonce_manager.nim
@@ -1,0 +1,69 @@
+when (NimMajor, NimMinor) < (1, 4):
+  {.push raises: [Defect].}
+else:
+  {.push raises: [].}
+
+import
+  chronos,
+  stew/results,
+  times
+import
+  ./constants
+
+export
+  chronos,
+  times,
+  results,
+  constants
+
+# This module contains the NonceManager interface
+# The NonceManager is responsible for managing the messageId used to generate RLN proofs
+# It should be used to fetch a new messageId every time a proof is generated
+# It refreshes the messageId every `epoch` seconds
+
+type
+  Nonce* = uint64
+  NonceManager* = ref object of RootObj
+    epoch*: float64
+    nextNonce*: Nonce
+    lastNonceTime*: float64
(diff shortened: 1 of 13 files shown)


waku/waku_rln_relay/nonce_manager.nim
```diff
@@ -0,0 +1,69 @@
...
+    epoch: epoch,
+    nextNonce: 0,
+    lastNonceTime: 0,
+    nonceLimit: nonceLimit
+  )
+
+
+proc get*(n: NonceManager): NonceManagerResult[Nonce] =
+  let now = getTime().toUnixFloat()
+  var retNonce = n.nextNonce
```
alrevuelta, line 57: in the if/else you set the value of `retNonce` in both cases. So this assigment will be overridden no matter what? Meaning its not needed?

or well, you can avoid the else, something like?


```
var retNonce = n.nextNonce
if now - n.lastNonceTime >= n.epoch:
  retNonce = 0
```

alrevuelta (approved):
lgtm, minor comment.

waku/waku_rln_relay/nonce_manager.nim
rymnc: addressed in [699a5bc](https://github.com/waku-org/nwaku/pull/2413/commits/699a5bc5c0c5218b8822ae4f76a2e942a63faf41)

PR #2418, opened 2024-02-09
feat: `eventCallback` per wakunode and `userData` 

richard-ramos:
# Description
1. Instead of using a global variable for the event callback, each wakunode instance will have their own event callback
2. Uses some of the code changes from https://github.com/waku-org/nwaku/pull/2089 to pass the `userData` to the callbacks


Ivansete-status (approved):
LGTM! Thanks for it! :100:

NagyZoltanPeter (approved):
Thanks for it, looks just good!!!

Issue #1961, opened 2023-08-28
feat: experimental incentivize store protocol


alrevuelta:
## Context

The goal of this issue is to create a proof of concept branching out nwaku, with the following new feature. Incentivize `store` protocol requests requiring a proof of payment of `x` tokens to an address in order to be processed. After successfully proving that the payment was done, then the node will return the requested messages. Note that this is intended to stay as an experimental branch, and most likely not merged into master in the short term.

For context, [see this](https://github.com/waku-org/research/issues/21).

Store protocol is one of waku's service protocols, based on a simple `req/resp` protocol where a node (full or light) [requests](https://github.com/waku-org/nwaku/blob/master/waku/waku_store/rpc.nim#L60C3-L65) the messages of a given topic from start/end dates and gets them in [return](https://github.com/waku-org/nwaku/blob/master/waku/waku_store/rpc.nim#L74-L77). The implementation is [here](https://github.com/waku-org/nwaku/blob/master/waku/waku_store/protocol.nim#L49-L101), which is triggered when a node receives the request.

## Feature

Some informal requirements. Let `requester` be the node that requests messages from store and `responder` the full node that stores messages and offers store capabilities:
* Introduce a new req/resp interaction that is done before the actual store req/resp (explained above). In this new interaction, both requester and responder agree on a price and amount of messages for given topic that will be paid and served. The requester gets an invoice id to be paid.
* The requester pays the invoice and performs the store request including this proof. This would require modifying [request](https://github.com/waku-org/nwaku/blob/master/waku/waku_store/rpc.nim#L60C3-L65).
* After the responder verifies that the payment was completed, it returns the agreed messages, completing the interaction.

Some notes:
* We can use Ethereum, Sepolia testnet would be fine, as we have nodes to use, and RLN is already integrated with this. Of course in reality it wouldn't be feasible to do this due to fees and processing times, but for the proof of concept its fine. As long as we target EVM (or Ethereum RPC compatible chains, we can then change).
* Nodes must have a new parameter, where they configure the "fee recipient" aka address where they will be getting the fees.
* To cut scope for this proof of concept, we can avoid having smart contracts and just use vanilla `rawTx`, using the `data` field to include the invoice id explained above.

The deliverable would be a branch with such feature, where multiple `wakunode2` instances can be run + interconnected in `requester` and `responder` mode. It must include the instruction (cli commands, etc) on how to see this feature live, where a node pays some ETH in return of a message.

chair28980:
[...]

PR #2419, opened 2024-02-12
feat: Incentivization: add codec for eligibility proofs

s-tikhomirov:
# Description

Encoding and decoding data structures that store proofs of eligibility is the first step towards a PoC implementation of incentivization for Store (and, in the future, other protocols), as outlined in #1961 and RFC [73/WAKU2-INCENTIVIZATION](https://github.com/vacp2p/rfc/tree/master/content/docs/rfcs/73).

# Changes

The PR implements the following:

- [x] new data types: `EligibilityProof` (which, for now, is thought to be a transaction id) and `EligibilityStatus`;
- [ ] ~~data types `DummyRequest` (with `EligibilityProof`) and `DummyResponse` (with `EligibilityStatus`) for testing purposes (to be replaced by the request and response structures of the corresponding protocol, e.g. Store);~~ (left for future work)
- [x] `encode` and `decode` procedures that convert Nim data types into protocol buffers and back, respectively;
- [x] tests to ensure that encoding and decoding work correctly: `decode(encode(X)) == X`.

## How to test

```
./env.sh bash
nim c -r ./tests/incentivization/test_rpc_codec.nim
```

### DIFF
--- a/tests/incentivization/test_rpc_codec.nim
+++ b/tests/incentivization/test_rpc_codec.nim
@@ -1,0 +1,39 @@
+import
+  std/options,
+  std/strscans,
+  testutils/unittests,
+  chronicles,
+  chronos,
+  libp2p/crypto/crypto
+
+import
+  ../../../waku/incentivization/rpc,
+  ../../../waku/incentivization/rpc_codec
+
+proc createRequestWithProof(): DummyRequest = 
+  var ep: EligibilityProof
+  ep.proofType = EligibilityProofType.TX_ID
+  ep.proof = "0xffffffff"
+  result.requestId = "request1"
+  result.eligibilityProof = some(ep)
+
+proc createResponse(): DummyResponse = 
+  var es: EligibilityStatus
+  es.statusCode = 200
+  es.statusDesc = "OK"
+  result.requestId = "request1"
+  result.eligibilityStatus = some(es)
+
+suite "Waku Incentivization Eligibility Codec":
+
+    asyncTest "encode eligibility request with proof":
+      let req = createRequestWithProof()
+      let decoded = DummyRequest.decode(encode(req).buffer).get()
+      check:
+          req == decoded
+        
+    asyncTest "encode eligibility response":
+      let resp = createResponse()
+      let decoded = DummyResponse.decode(encode(resp).buffer).get()
+      check:
+        resp == decoded
--- a/tests/incentivization/test_rpc_codec.nim
+++ b/tests/incentivization/test_rpc_codec.nim
@@ -10,30 +10,23 @@ import
   ../../../waku/incentivization/rpc,
   ../../../waku/incentivization/rpc_codec
 
-proc createRequestWithProof(): DummyRequest = 
-  var ep: EligibilityProof
-  ep.proofType = EligibilityProofType.TX_ID
-  ep.proof = "0xffffffff"
-  result.requestId = "request1"
-  result.eligibilityProof = some(ep)
-
-proc createResponse(): DummyResponse = 
-  var es: EligibilityStatus
-  es.statusCode = 200
-  es.statusDesc = "OK"
-  result.requestId = "request1"
-  result.eligibilityStatus = some(es)
 
 suite "Waku Incentivization Eligibility Codec":
 
-    asyncTest "encode eligibility request with proof":
-      let req = createRequestWithProof()
-      let decoded = DummyRequest.decode(encode(req).buffer).get()
+    asyncTest "encode eligibility proof":
+      var byteSequence: seq[byte] = @[1, 2, 3]
+      let epRpc = EligibilityProof(proof: some(byteSequence))
+      let encoded = encode(epRpc)
+      let decoded = EligibilityProof.decode(encoded.buffer).get()
       check:
-          req == decoded
-        
-    asyncTest "encode eligibility response":
-      let resp = createResponse()
-      let decoded = DummyResponse.decode(encode(resp).buffer).get()
+          epRpc == decoded
+    
+    asyncTest "encode eligibility status":
+      let esRpc = EligibilityStatus(
+        statusCode: some(uint32(200)),
+        statusDesc: some("OK")
+      )
+      let encoded = encode(esRpc)
+      let decoded = EligibilityStatus.decode(encoded.buffer).get()
       check:
-        resp == decoded
+        esRpc == decoded
--- a/tests/incentivization/test_rpc_codec.nim
+++ b/tests/incentivization/test_rpc_codec.nim
@@ -23,7 +23,7 @@ suite "Waku Incentivization Eligibility Codec":
     
     asyncTest "encode eligibility status":
       let esRpc = EligibilityStatus(
-        statusCode: some(uint32(200)),
+        statusCode: uint32(200),
         statusDesc: some("OK")
       )
       let encoded = encode(esRpc)
--- a/tests/incentivization/test_rpc_codec.nim
+++ b/tests/incentivization/test_rpc_codec.nim
@@ -14,8 +14,8 @@ import
 suite "Waku Incentivization Eligibility Codec":
 
     asyncTest "encode eligibility proof":
-      var byteSequence: seq[byte] = @[1, 2, 3]
-      let epRpc = EligibilityProof(proof: some(byteSequence))
+      var byteSequence: seq[byte] = @[1, 2, 3, 4, 5, 6, 7, 8]
+      let epRpc = EligibilityProof(proofOfPayment: some(byteSequence))
       let encoded = encode(epRpc)
       let decoded = EligibilityProof.decode(encoded.buffer).get()
       check:
--- a/waku/incentivization/rpc.nim
+++ b/waku/incentivization/rpc.nim
@@ -10,7 +10,7 @@ import
 type
 
   EligibilityProof* = object
-    proof*: Option[seq[byte]]
+    proofOfPayment*: Option[seq[byte]]
 
   EligibilityStatus* = object
     statusCode*: Option[uint32]
--- a/waku/incentivization/rpc.nim
+++ b/waku/incentivization/rpc.nim
@@ -1,0 +1,27 @@
+import
+  json_serialization,
+  std/options
+import
+  ../waku_core
+
+type
+  EligibilityProofType* {.pure.} = enum
+    # Indicates the type of eligibility proof
+    NONE = uint32(0)
+    TX_ID = uint32(1)
+
+  EligibilityProof* = object
+    proofType*: EligibilityProofType
+    proof*: string # or bytes?
+
+  DummyRequest* = object
+    requestId*: string
+    eligibilityProof*: Option[EligibilityProof]
+
+  EligibilityStatus* = object
+    statusCode*: uint32
+    statusDesc*: string # should descripiton be optional?
+
+  DummyResponse* = object
+    requestId*: string
+    eligibilityStatus*: Option[EligibilityStatus]
--- a/waku/incentivization/rpc.nim
+++ b/waku/incentivization/rpc.nim
@@ -4,24 +4,14 @@ import
 import
   ../waku_core
 
+# Exactly following the RFC:
+# https://github.com/vacp2p/rfc/tree/master/content/docs/rfcs/73
+
 type
-  EligibilityProofType* {.pure.} = enum
-    # Indicates the type of eligibility proof
-    NONE = uint32(0)
-    TX_ID = uint32(1)
 
   EligibilityProof* = object
-    proofType*: EligibilityProofType
-    proof*: string # or bytes?
-
-  DummyRequest* = object
-    requestId*: string
-    eligibilityProof*: Option[EligibilityProof]
+    proof*: Option[seq[byte]]
 
   EligibilityStatus* = object
-    statusCode*: uint32
-    statusDesc*: string # should descripiton be optional?
-
-  DummyResponse* = object
-    requestId*: string
-    eligibilityStatus*: Option[EligibilityStatus]
+    statusCode*: Option[uint32]
+    statusDesc*: Option[string]
--- a/waku/incentivization/rpc.nim
+++ b/waku/incentivization/rpc.nim
@@ -4,7 +4,7 @@ import
 import
   ../waku_core
 
-# Exactly following the RFC:
+# Implementing the RFC:
 # https://github.com/vacp2p/rfc/tree/master/content/docs/rfcs/73
 
 type
@@ -13,5 +13,5 @@ type
     proofOfPayment*: Option[seq[byte]]
 
   EligibilityStatus* = object
-    statusCode*: Option[uint32]
+    statusCode*: uint32
     statusDesc*: Option[string]
--- a/waku/incentivization/rpc_codec.nim
+++ b/waku/incentivization/rpc_codec.nim
@@ -10,9 +10,9 @@ import
 
 proc encode*(epRpc: EligibilityProof): ProtoBuffer = 
   var pb = initProtoBuffer()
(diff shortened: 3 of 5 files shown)


waku/incentivization/rpc.nim
```diff
@@ -0,0 +1,27 @@
...
+    NONE = uint32(0)
+    TX_ID = uint32(1)
+
+  EligibilityProof* = object
+    proofType*: EligibilityProofType
+    proof*: string # or bytes?
+
+  DummyRequest* = object
+    requestId*: string
+    eligibilityProof*: Option[EligibilityProof]
```
jm-clius, line 19: I think both `eligibilityProof` and `eligibilityStatus` should be `seq[bytes]` within their encapsulated Request and Response messages respectively (and in the protobufs `bytes`). This allows for two stage decoding (those decoding `DummyRequest` don't need to also decode `EligibilityProof` if not interesting to them). I see the example encoding into HistoryRequest and HistoryResponse in https://github.com/vacp2p/rfc/tree/master/content/docs/rfcs/73 is inconsistent here - encoding as `bytes` in the Request but not in the Response.

waku/incentivization/rpc_codec.nim
```diff
@@ -0,0 +1,81 @@
...
+  var pb = initProtoBuffer()
+
+  pb.write3(1, rpc.requestId)
+
+  if rpc.eligibilityProof.isSome:
+    let ep = rpc.eligibilityProof.get()
+    pb.write3(2, uint32(ord(ep.proofType)))
+    pb.write3(3, ep.proof)
+  else:
+    pb.write3(2, uint32(ord(EligibilityProofType.NONE)))
```
jm-clius, line 19: I'd decouple this encoding from the encapsulating Request encoding. In other words, there should be a separate `encode*(rpc: EligibilityProof): ProtoBuffer`. Same for the `EligibilityStatus` and for decoding.

waku/incentivization/rpc_codec.nim
```diff
@@ -0,0 +1,81 @@
...
+
+proc encode*(rpc: DummyRequest): ProtoBuffer = 
+  var pb = initProtoBuffer()
+
+  pb.write3(1, rpc.requestId)
+
+  if rpc.eligibilityProof.isSome:
+    let ep = rpc.eligibilityProof.get()
+    pb.write3(2, uint32(ord(ep.proofType)))
+    pb.write3(3, ep.proof)
```
jm-clius, line 17: Since `EligibilityProof` is a separate protobuf message, why not reset the tag numbers here (they cannot clash with the encapsulating Request's).

waku/incentivization/rpc_codec.nim
```diff
@@ -0,0 +1,81 @@
...
+  var code: uint32
+  var desc: string
+  var eligibilityStatus: EligibilityStatus
+  if ?pb.getField(2, code):
+    eligibilityStatus.statusCode = code
+    if ?pb.getField(3, desc):
+      eligibilityStatus.statusDesc = desc
+    else:
+      # assuming description is not optional
+      return err(ProtobufError.missingRequiredField("status_desc"))
```
jm-clius, line 78: I would make the `status_code` mandatory and `status_desc` optional.

jm-clius (approved):
LGTM as a start. I've added some comments below, but these can be addressed in a follow-up PR. Mostly I would decouple the coding/decoding of `EligibilityProof` and `EligibilityStatus` from the underlying req-resp protocol. This would also mean that you probably don't need a `DummyRequest` and `DummyResponse` for now - you can simply build logic and tests on the raw bytes of the eligibility messages.

tests/incentivization/test_rpc_codec.nim
```diff
@@ -0,0 +1,39 @@
...
+import
+  ../../../waku/incentivization/rpc,
+  ../../../waku/incentivization/rpc_codec
+
+proc createRequestWithProof(): DummyRequest = 
+  var ep: EligibilityProof
+  ep.proofType = EligibilityProofType.TX_ID
+  ep.proof = "0xffffffff"
+  result.requestId = "request1"
+  result.eligibilityProof = some(ep)
```
Ivansete-status, line 18: It is recommended not to use the implicit `result` return type:
```suggestion
  return DummyRequest(requestId: "request1", eligibilityProof: some(ep))
```

Ivansete-status (approved):
LGTM! I just added a minor comment

jm-clius:
@s-tikhomirov anything blocking for last changes and to get this merged?

Ivansete-status:
Morning @s-tikhomirov!
Shall this PR be merged or is anything blocking it?

s-tikhomirov:
> Morning @s-tikhomirov! Shall this PR be merged or is anything blocking it?

Morning @Ivansete-status ! Sorry for keeping this open, I've been busy with another line of work until now (academic papers). I've been thinking of addressing existing comments in the same PR, would it be OK?

jm-clius:
@s-tikhomirov I would suggest converting to a `draft` PR until you're happy that the changes are ready to be reviewed and merged. cc @Ivansete-status

waku/incentivization/rpc.nim
```diff
@@ -0,0 +1,27 @@
...
+  ../waku_core
+
+type
+  EligibilityProofType* {.pure.} = enum
+    # Indicates the type of eligibility proof
+    NONE = uint32(0)
+    TX_ID = uint32(1)
+
+  EligibilityProof* = object
+    proofType*: EligibilityProofType
```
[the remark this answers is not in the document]
s-tikhomirov: > Perhaps proof should be tag 10

It is tag 1 [in the RFC](https://github.com/vacp2p/rfc/tree/master/content/docs/rfcs/73#wire-format-specification--syntax) though. Do you think it makes sense to implement the existing RFC first, and then iterate on it? Or shall I submit a change to RFC at the same time? I'm a bit vary of modifying both at the same time.

waku/incentivization/rpc.nim
s-tikhomirov: > both eligibilityProof and eligibilityStatus should be seq[bytes] within their encapsulated Request and Response messages respectively (and in the protobufs bytes)

I find this a bit confusing: how does this suggestion relate to the RFC definitions of `eligibilityProof` and `eligibilityProof`? For example, `eligibilityStatus` is defined in the RFC as an object containing an optional `uint32` and an optional `string`. How does this relate to treating both `eligibilityProof` and `eligibilityProof` as bytes? Do you suggest we change the RFC in that respect?

> the example encoding into HistoryRequest and HistoryResponse in https://github.com/vacp2p/rfc/tree/master/content/docs/rfcs/73 is inconsistent here - encoding as bytes in the Request but not in the Response.

Do you mean, the exampes in the RFC are inconsistent? AFAIR, my reasoning was that the proof can take many forms (txid, zkp, etc), which we don't know in advance and therefore use `bytes` as the lowest common denominator, but the response, in contrast, always has the same form: an error code and an error description.

waku/incentivization/rpc_codec.nim
s-tikhomirov: They are both optional in the RFC... I'm not sure I can recall what the reasoning there was (might have simply been an oversight), but I have the same question now: what way should I go among the following options?

1. implement the RFC as it stands now (which means ignoring some of your suggestions, at least for a while);
2. modify the RFC, and then implement the updated version;
3. implement a not-exactly-compliant version of the code (and perhaps modify the RFC later to reflect the implementation).

waku/incentivization/rpc_codec.nim
```diff
@@ -0,0 +1,63 @@
...
+  pb
+
+proc decode*(T: type EligibilityStatus, buffer: seq[byte]): ProtobufResult[T] = 
+  let pb = initProtoBuffer(buffer)
+  var esRpc = EligibilityStatus()
+  # status code
+  # TODO: write this more concisely: ternary operator, default values?
+  # something like this, if hasField by field number existed:
+  #esRpc.statusCode = pb.hasField(1) ? some(pb.getField(1)) : none(uint32)
+  # the same applies to the EligibilityProof's decode
```
Ivansete-status, line 49: I think we can remove that comment. It looks great as it is now

waku/incentivization/rpc_codec.nim
```diff
@@ -0,0 +1,63 @@
...
+  else:
+    epRpc.proof = some(proof)
+  ok(epRpc)
+
+
+# Codec for EligibilityStatus
+
+proc encode*(esRpc: EligibilityStatus): ProtoBuffer = 
+  var pb = initProtoBuffer()
+  if esRpc.statusCode.isSome:
```
Ivansete-status, line 36: Nitpick comment: the use of `()` in "verbs".
```suggestion
  if esRpc.statusCode.isSome():
```

tests/incentivization/test_rpc_codec.nim
```diff
@@ -0,0 +1,32 @@
...
+  std/strscans,
+  testutils/unittests,
+  chronicles,
+  chronos,
+  libp2p/crypto/crypto
+
+import
+  ../../../waku/incentivization/rpc,
+  ../../../waku/incentivization/rpc_codec
+
```
Ivansete-status, line 12: Kindly add this file within the `tests/all_tests_waku.nim` file so that it is considered in our CI tests when a PR is submitted.

Ivansete-status (approved):
LGTM, thanks!
I just added simple comments

waku/incentivization/rpc_codec.nim
SionoiS: +1

tests/incentivization/test_rpc_codec.nim
```diff
@@ -0,0 +1,32 @@
...
+
+suite "Waku Incentivization Eligibility Codec":
+
+    asyncTest "encode eligibility proof":
+      var byteSequence: seq[byte] = @[1, 2, 3]
```
SionoiS, line 17: Nitpick: I would add more bytes
```suggestion
      var byteSequence = @[0..31, byte]
```

SionoiS (approved):
LGTM

Style wise I prefer to cluster logically related lines together and add spaces in between.
It's only my personal preference though.

waku/incentivization/rpc.nim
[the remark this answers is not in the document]
jm-clius: Generally it's up to you which you do first during the very early stages. I like to keep at least a draft PR with spec changes open which I keep more or less update while working on the POC. This helps with communication since people can be referred to the spec to clarify what you're doing in the code (and it won't get merged, since it's just a draft).

waku/incentivization/rpc_codec.nim
jm-clius: In terms of RFC - same suggestion as above. It's fine to harden concepts in POC before spec change, but personally I'd keep a draft PR with spec changes open in parallel.

Note, however, that these fields being `optional` in the protobuf does not imply that they are optional in the _protocol_. These are just directives on how serialisation is done. Optionality gets specified as the rules (i.e. text) separate to the protobuf wire specification. In order to test if a field is present in serialised protobuf and ensure mandatory fields are present, we have to set it as `optional`. Otherwise the field will always appear exactly like it would have if someone just used the default value for the field. In your example, we can only do the:

```
if not ?pb.getField(1, code)
```

check if the corresponding field is present in the protobuf specification. Otherwise `code` will just default to the equivalent of code `0`, because the field's presence cannot be checked.

For that reason, in the protobuf specification it's good to make all fields `optional` by default other than in some specific corner cases.

jm-clius: In short, all fields (regardless label) in proto3 are optional - you can choose to set these fields or not. Leaving a value unset or explicitly setting it to the default value means that field will not get serialised. Therefore, in order to explicitly check if a field was set (and not unintentionally left out or intentionally set to the default), you have to use `optional` - otherwise it will just appear like someone might have used the default value for that field type.  `optional` keyword should have rather been `you_can_check_if_this_field_was_set`.

https://protobuf.dev/programming-guides/field_presence/

waku/incentivization/rpc.nim
```diff
@@ -0,0 +1,17 @@
...
+import
+  ../waku_core
+
+# Exactly following the RFC:
+# https://github.com/vacp2p/rfc/tree/master/content/docs/rfcs/73
+
+type
+
+  EligibilityProof* = object
+    proof*: Option[seq[byte]]
```
jm-clius, line 13: Not sure if this is a planned change in the RFC, but I'd stick to the same field names. In other words, this can `proofOfPayment` rather than just `proof`. If we add other proof types, these can be new fields.

waku/incentivization/rpc.nim
```diff
@@ -0,0 +1,17 @@
...
+
+  EligibilityStatus* = object
+    statusCode*: Option[uint32]
```
jm-clius, line 16: I think this should be a mandatory field, so should not be an `Option`. See https://github.com/waku-org/nwaku/blob/master/waku/waku_filter_v2/rpc.nim#L23-L26 for analogous implementation for Filter. Note that in the RFC protobuf this field must be marked `optional`. This is protobuf weirdness and orthogonal to whether we consider this field mandatory or not.

waku/incentivization/rpc_codec.nim
```diff
@@ -0,0 +1,63 @@
...
+proc decode*(T: type EligibilityStatus, buffer: seq[byte]): ProtobufResult[T] = 
+  let pb = initProtoBuffer(buffer)
+  var esRpc = EligibilityStatus()
+  # status code
+  # TODO: write this more concisely: ternary operator, default values?
+  # something like this, if hasField by field number existed:
+  #esRpc.statusCode = pb.hasField(1) ? some(pb.getField(1)) : none(uint32)
+  # the same applies to the EligibilityProof's decode
+  var code = uint32(0)
+  if not ?pb.getField(1, code):
```
jm-clius, line 51: If there's a response I think there MUST be a status_code. In other words, raise an error here rather than just setting this to `none`. Generally we do consider the extra information in the `status_desc` to be optional.

jm-clius (approved):
LGTM. I think this is a better approach rather than introducing the dummy protocol. I would like to see my comments addressed, but let's do so in a subsequent PR. This one is good to merge. Comments mostly relate to mandatory vs optional fields (in protobuf all fields should be `optional` for consistent, flexible wire encoding, but we can and should enforce rules around mandatory fields at the point where we decode)

tests/incentivization/test_rpc_codec.nim
s-tikhomirov: Got an error on the modified line, what am I doing wrong?

```
Error: type mismatch: got 'typedesc[byte]' for 'byte' but expected 'HSlice[system.int, system.int]'

```

SionoiS: ah sorry that's a Rust thing. In Nim you can't initialize an array with a range AFAIK.

waku/incentivization/rpc.nim
s-tikhomirov: The reasoning makes perfect sense, thank you! So am I correct that this PR is about the eligibility module that "decodes/encodes/interprets eligibility protocol"? Which means that the codec defined in this PR _should_ understand the inner structure of `EligibilityStatus`, namely, that it contains a status code and status description? (Sorry if this looks confusing in an "outdated" thread.)

jm-clius: Indeed! This comment is outdated in that we don't have the Dummy Protocol anymore, but will be relevant for how this gets encapsulated in future req-resp protocols.

tests/incentivization/test_rpc_codec.nim
s-tikhomirov: OK, I'll then add more bytes manually (not 32, perhaps 8).

PR #2424, opened 2024-02-13
fix: check max message size in validator according to configured value

### LINKED ISSUE #2406: chore: make max message size configurable only for clusterIds different than 1
### Background
In cluster-ids different than 1 (aka not the Waku Network), we want the max size of a WakuMessage to be configurable.
For cluster id one, it should comply with the hard limit of the Waku Network.

We should also update the checks at the REST API handler to use the configurable value and not the default limit.

### Acceptance criteria
- max message size should be configurable outside of cluster-id 1
- message validation checks should use the configurable value instead of the hardcoded default

gabrielmer:
# Description
Maximum message size is a configurable parameter. Therefore, we're updating the validator that we run in our REST handlers to check against the configured value.

Notice that for clusterId=1, the configuration is overriden to the Waku Network's max message size during app startup.
https://github.com/waku-org/nwaku/blob/50308eda0d678ab1a63136d6ed931ba9fc5a700e/apps/wakunode2/wakunode2.nim#L78
# Changes

- [x] running size validator against the configured max message size value
- [x] updated tests

## Issue

closes #2406

### DIFF
--- a/waku/waku_relay/protocol.nim
+++ b/waku/waku_relay/protocol.nim
@@ -8,6 +8,7 @@ else:
   {.push raises: [].}
 
 import
+  std/strformat,
   stew/results,
   sequtils,
   chronos,
@@ -215,21 +216,15 @@ proc generateOrderedValidator*(w: WakuRelay): auto {.gcsafe.} =
     return ValidationResult.Accept
   return wrappedValidator
 
-proc isValidSize(message: WakuMessage, maxMessageSize = MaxWakuMessageSize): Result[void, string] =
-  let messageSizeBytes = uint64(message.encode().buffer.len)
-
-  if(messageSizeBytes > maxMessageSize):
-      let message = "Message size exceeded maximum of " & $maxMessageSize & " bytes"
-      debug "Invalid Waku Message", error=message
-      return err(message)
-    
-  return ok()
-
 proc validateMessage*(w: WakuRelay, pubsubTopic: string, msg: WakuMessage):
   Future[Result[void, string]] {.async.} =
 
-    msg.isValidSize(uint64(w.maxMessageSize)).isOkOr:
-      return err(error)
+    let messageSizeBytes = msg.encode().buffer.len
+
+    if messageSizeBytes > w.maxMessageSize:
+      let message = fmt"Message size exceeded maximum of {w.maxMessageSize} bytes"
+      debug "Invalid Waku Message", error=message
+      return err(message)
 
     for (validator, message) in w.wakuValidators:
         let validatorRes = await validator(pubsubTopic, msg)
--- a/waku/waku_relay/protocol.nim
+++ b/waku/waku_relay/protocol.nim
@@ -8,7 +8,6 @@ else:
   {.push raises: [].}
 
 import
-  std/strformat,
   stew/results,
   sequtils,
   chronos,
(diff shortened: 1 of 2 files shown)


waku/waku_relay/protocol.nim
```diff
@@ -229,7 +228,7 @@ proc isValidSize(message: WakuMessage): Future[Result[void, string]] {.async.} =
 proc validateMessage*(w: WakuRelay, pubsubTopic: string, msg: WakuMessage):
   Future[Result[void, string]] {.async.} =
 
-    (await msg.isValidSize()).isOkOr:
+    msg.isValidSize(uint64(w.maxMessageSize)).isOkOr:
```
alrevuelta, line 231: would it make sense to move the content of `isValidSize` function inside `validateMessage`? Instead of having it as another function?

Like this, also there is no need to pass the `maxMessageSize` since its already part of the relay class `w.maxMessageSize`

gabrielmer: The idea of having it in another function is that validating a message size is a common use case (and it's not just an if condition, there's the encoding of the message which makes it a little less trivial).
So it's something with potential to be useful on its own.

Didn't understand about  passing `maxMessageSize`. If we use `isValidSize()` then we pass `maxMessageSize` that we received in `validateMessage`. Am I missing anything?

LMK what do you think

alrevuelta: > validating a message size is a common use case

i just saw one usage of `isValidSize`. Is there any use case where we would want to check `isValidSize` and not `validateMessage` as a whole?

> there's the encoding of the message which makes it a little less trivial

mm right. unsure how max msg is defined but it used to be the max payload size (not the entire message). in such case, there is no need to decode it, just payload.len. I'm ok with both, at your discrection.

I mean something as simple as:

```
proc validateMessage*(w: WakuRelay, pubsubTopic: string, msg: WakuMessage):
  Future[Result[void, string]] {.async.} =

    if(messageSizeBytes > w.maxMessageSize):
      return err("wrong msgsize: xxx")

    for (validator, message) in w.wakuValidators:
        let validatorRes = await validator(pubsubTopic, msg)
        if validatorRes != ValidationResult.Accept:
            if message.len > 0:
              return err(message)
            else:
              return err("Validator failed")
    return ok()
```

and remove `isValidSize`

wdyt?

gabrielmer: > 
> i just saw one usage of `isValidSize`. Is there any use case where we would want to check `isValidSize` and not `validateMessage` as a whole?

True, actually not sure in which cases someone would run only the size validation instead of all validators. There's something that makes me feel that it's useful to have a function that receives a WakuMessage and tells you if it has a valid size or not. But maybe there's not really a use case at all :)

>
> mm right. unsure how max msg is defined but it used to be the max payload size (not the entire message). in such case, there is no need to decode it, just payload.len. I'm ok with both, at your discrection.

Yes, I don't know where the formal definition of the max size should be. I took the encoding decision based on how I saw the max message size enforced in libp2p, which compares the maximum defined with the encoded size: https://github.com/status-im/nim-libp2p/blob/e0f70b71778bf9613421261b22a56499c913f806/libp2p/protocols/pubsub/pubsubpeer.nim#L300-L324

Also, I thought that considering only the payload for the max size can be an attack vector if we don't explicitly enforce a max size for the other message fields. Someone could probably add a huge payload in the `meta` field. Or maybe a huge string as `contentTopic` (not sure if there's a max size for that)

alrevuelta: perfect. lets leave the max size as it is, which is ok.

regarding the other:
its just that don't like `isValidSize(message: WakuMessage, maxMessageSize = MaxWakuMessageSize):` because in the same code you can end up having `isValidSize(m, size1)` and `isValidSize(m, size2)` and maybe `size1` and `size2` are different from the `w.maxMessageSize`. This mismatch can trigger weird behavior where the message has valid size or not depending on "who asks".

so I'm fine with either: 
* `isValidSize(WakuRelay: w, message: WakuMessage):`. So you don't have to pass the size yourself. Its in `w.`
* or this https://github.com/waku-org/nwaku/pull/2424#discussion_r1487665586

IMHO the ones that judge if the message is valid or not is the relay. and they relay has a configured `maxMessageSize` which IMHO is safer to use.

gabrielmer: Agree! Sounds good! Thanks 🤩 
Just pushed the fixes :))

alrevuelta (approved):
lgtm!

Ivansete-status (approved):
LGTM! Thanks! :100:

PR #2427, opened 2024-02-13
fix: checking for keystore file existence

gabrielmer:
# Description

Checking for file existence and not only for a configured path when attempting to mount RLN.

This will avoid errors when users run nwaku-compose without having created the RLN credentials, as currently the `/keystore/keystore.json` path is passed by default and we try to read from a file that doesn't exist.

# Changes

- [x] checking for keystore file existence before attempting to read from it
- [x] raising exception if path and password are provided but file doesn't exist
- [x] added test case 

## Issue

Helps https://github.com/waku-org/nwaku-compose/issues/32


waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
```diff
@@ -622,7 +623,7 @@ method init*(g: OnchainGroupManager): Future[void] {.async.} =
   g.rlnContract = some(rlnContract)
   g.registryContract = some(registryContract)
 
-  if g.keystorePath.isSome() and g.keystorePassword.isSome():
+  if g.keystorePath.isSome() and existsFile(g.keystorePath.get()) and g.keystorePassword.isSome():
```
rymnc, line 626: we probably want to have a specific error log when a path and password is provided but it doesn't exist, what do you think?

gabrielmer: Sounds good! Added it now :)) lmk what you think about the new version

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
```diff
@@ -623,38 +623,41 @@ method init*(g: OnchainGroupManager): Future[void] {.async.} =
...
+                                                              .idCommitment.toUInt256()).call()
+        if membershipExists == 0:
+          raise newException(CatchableError, "the provided commitment does not have a membership")
+      except CatchableError:
+        raise newException(CatchableError, "could not check if the commitment exists on the contract: " &
+                                          getCurrentExceptionMsg())
+
+      g.idCredentials = some(keystoreCred.identityCredential)
+    else:
+      warn "File provided as keystore path does not exist", path=g.keystorePath.get()
```
rymnc, line 660: probably want to exit here right, this means the node will not be able to generate proofs.
additionally, can we add a test vector to further ensure that it works as intended?

gabrielmer: Perfect - thought we could just continue relaying messages as it's supposed to be optional. But if path and password are passed, I guess that the intention is to be able to generate proofs.

This implies that in nwaku-compose generating the credentials will be mandatory. Or in other words, we won't be able to run a node by simply running `docker-compose up -d` without generating credentials and without manually changing `run_node.sh`

rymnc (approved):
LGTM, thanks for this!

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
```diff
@@ -623,37 +624,41 @@ method init*(g: OnchainGroupManager): Future[void] {.async.} =
   g.registryContract = some(registryContract)
 
   if g.keystorePath.isSome() and g.keystorePassword.isSome():
-    var keystoreQuery = KeystoreMembership(
-      membershipContract: MembershipContract(
-        chainId: $g.chainId.get(),
-        address: g.ethContractAddress
+    if existsFile(g.keystorePath.get()):
```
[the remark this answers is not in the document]
gabrielmer: Done! Much much better 😍  😍  😍 
Thanks!

Ivansete-status (approved):
LGTM! Thanks for it! :1st_place_medal:

PR #2428, opened 2024-02-13
chore: waku_metrics - change log interval from 30'' to 10'

Ivansete-status:
## Description
Simple PR to change the log cadence as per our last PM meeting.

It changes the frequency of "Total *" logs (metrics logs) from 30 seconds to 10 minutes.

### DIFF
--- a/waku/node/waku_metrics.nim
+++ b/waku/node/waku_metrics.nim
@@ -16,7 +16,7 @@ import
   ./peer_manager,
   ./waku_node
 
-const LogInterval = 30.seconds
+const LogInterval = 10.minutes
 
 logScope:
   topics = "waku node metrics"


gabrielmer (approved):
Thank you!

NagyZoltanPeter (approved):
Super fast reaction! Thank you. :-)

alrevuelta (commented):
can we merge some lines as well? here and in `getRlnMetricsLogger`

something like this. That /5 the amount of logs without impacting the frequency.
```
info "Connected peers", storePeers=storePeers, pxPeers=pxPeers, lightpushPeers=lightpushPeers, filterPeers=filterPeers.
```

which btw, might be a peermanager job. but that's for another day.

and well, not relevant if the protocol is not enabled (eg with lp disabled will print always 0) but also, ok by now.

jm-clius (approved):
LGTM. :)

Ivansete-status:
@alrevuelta - I'll submit a separate PR with your proposal, https://github.com/waku-org/nwaku/pull/2428#pullrequestreview-1878464481

PR #2430, opened 2024-02-14
chore: add 150 kB to msg size histogram metric

vpavlin:
# Description

With the change of default message max size to 140 kB, we are missing some details in the message size histogram

image 

# Changes

This PR adds 150 kB bucket, so that we can see number of messages between 100 - 150 kB. I left the rest in there for the case of non-TWN networks

- [ ] add 150 bucket to msg size metrics
- [ ] ...


jm-clius (approved):

alrevuelta (approved):

jm-clius:
@vpavlin anything needed to merge this? Follow-up suggestions can be introduced in a subsequent PR, if necessary.

PR #2431, opened 2024-02-14
feat: store v3

SionoiS:
# Description

The new Store protocol version 3 will work side by side with the old version for some time.

I copied all the code from v2 as a starting point. Store v2 is now known as legacy store.

Store v3 order results a bit differently than legacy. Sorted by time and then by hash (was digest previously).
Everything remain the same except now we have the ability to query by hashes too :tada: 

# Changes
- [x] New Store v3 protocol and client logic
- [x] New de/serialization according to the WIP specs
- [x] update tests
- [x] rest api
- [x] Store v3 and archive tie in logic

Tracking https://github.com/waku-org/nwaku/issues/2425


waku/waku_store_v3/client.nim
```diff
@@ -0,0 +1,210 @@
...
+  import std/[sequtils, times]
+  import ../waku_archive
+  import ../waku_core/message/digest
+
+logScope:
+  topics = "waku store client"
+
+const DefaultPageSize*: uint = 20 # A recommended default number of waku messages per page
+
+type WakuStoreV3Client* = ref object
```
jm-clius, line 30: Since the `V3` is essentially a temporary measure to disambiguate with the "legacy" store protocol, perhaps another pattern is to design type names for the new (v3) Store to be exactly like we envision them to be in the long term and disambiguate on `waku_node` by:
- using aliases for the imports related to legacy Store (e.g. `./waku_store/client as legacy_store_client`)
- renaming where necessary the _legacy_ types, metrics, etc. to `...Legacy` (e.g. `WakuStoreClientLegacy`)

This would basically prepare the codebase to remove the legacy store protocol cleanly in the future without having to change "v3" implementation again.

waku/node/waku_node.nim
```diff
@@ -88,6 +90,8 @@ type
     wakuArchive*: WakuArchive
     wakuStore*: WakuStore
     wakuStoreClient*: WakuStoreClient
```
jm-clius, line 92: See my comment further down on using `v3` in code. Perhaps rather rename these to `wakuLegacyStore`, `wakuLegacyStoreClient` ?

waku/waku_store_v3/common.nim
```diff
@@ -0,0 +1,107 @@
...
+    paginationLimit*: Option[uint64]
+
+  WakuMessageKeyValue* = object
+    messageHash*: WakuMessageHash
+    message*: WakuMessage
+
+  StoreResponse* = object
+    requestId*: string
+
+    statusCode*: Option[uint32]
```
jm-clius, line 64: Don't think this should be an `Option` (it's a mandatory field and should result in decoding error if not present in received response)

waku/waku_store_v3/protocol.nim
```diff
@@ -0,0 +1,107 @@
...
+    let req = StoreRequest.decode(buf).valueOr:
+      error "failed to decode rpc", peerId= $conn.peerId
+      waku_store_errors.inc(labelValues = [decodeRpcFailure])
+      
+      res.statusCode = some(uint32(400))# TODO find a way to use the error enum
+      res.statusDesc = some("decode rpc failed")
+
+      await conn.writeLp(res.encode().buffer)
+      return
+
```
jm-clius, line 59: After decoding, I would pull out `handleStoreRequest` into a separate `proc` (for unit testing and better separation of interests).

jm-clius (commented):
Direction lgtm! Added a couple of comments. Will also be making some minor changes to the draft RFC, but will keep you informed about those. :)

waku/waku_store_v3/client.nim
```diff
@@ -0,0 +1,210 @@
...
+    return err(Error(kind: ErrorKind.BAD_RESPONSE, cause: decodeRpcFailure))
+
+  return ok(res)
+
+proc query*(self: WakuStoreV3Client, req: StoreRequest, peer: RemotePeerInfo): Future[StoreResult] {.async, gcsafe.} =
+  return await self.sendStoreRequest(req, peer)
+
+# TODO: Move outside of the client
+when defined(waku_exp_store_resume):
+  ## Resume store
```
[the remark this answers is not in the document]
SionoiS: I vote yes, I never heard anyone even mention this feature.

SionoiS:
N.B. tests can't pass before archive is updated and store v3 is tied to it.

waku/waku_store/rpc_codec.nim
```diff
@@ -3,257 +3,210 @@ when (NimMajor, NimMinor) < (1, 4):
...
 const MaxRpcSize* = MaxPageSize * MaxWakuMessageSize + 64 * 1024
   # We add a 64kB safety buffer for protocol overhead
 
-## Pagination
+### Request ###
 
-proc encode*(index: PagingIndexRPC): ProtoBuffer =
-  ## Encode an Index object into a ProtoBuffer
-  ## returns the resultant ProtoBuffer
+proc encode*(req: StoreQueryRequest): ProtoBuffer =
```
chaitanyaprem, line 15: General Question:

We are manually encoding protobufs, seems error-prone and high maintenance. 
Can't we use something like https://github.com/PMunch/protobuf-nim to generate the nim code?

chaitanyaprem (commented):
Reviewed at high level only Store protocol changes.

Will do the rest again as i have to understand a lot of existing code to understand changes :)

waku/waku_store_legacy/client.nim
```diff
@@ -0,0 +1,240 @@
+when (NimMajor, NimMinor) < (1, 4):
+  {.push raises: [Defect].}
+else:
+  {.push raises: [].}
+
+import std/options, stew/results, chronicles, chronos, metrics, bearssl/rand
```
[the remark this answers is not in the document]
SionoiS: I agree that's awkward to review. The folder changed that's the reason maybe? Or it may be due to nph IDK.

waku/waku_store/client.nim
```diff
@@ -27,214 +17,41 @@ type WakuStoreClient* = ref object
...
 
-  # Disabled ,for now, since the default response is a possible case (no messages, pagesize = 0, error = NONE(0))
-  # TODO: Rework the RPC protocol to differentiate the default value from an empty value (e.g., status = 200 (OK))
-  #        and rework the protobuf parsing to return Option[T] when empty values are received
-  if respRpc.response.isNone():
-    waku_store_errors.inc(labelValues = [emptyRpcResponseFailure])
-    return err(
-      HistoryError(kind: HistoryErrorKind.BAD_RESPONSE, cause: emptyRpcResponseFailure)
-    )
+  let buf = await connection.readLp(MaxRpcSize.int)
```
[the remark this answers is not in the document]
SionoiS: Your totally right it should have a `catch: `. It always gets me when a proc doesn't return a `Result`....

waku/waku_store/rpc_codec.nim
SionoiS: We should but if I'm not mistaken we use proto2 and not 3?

waku/waku_store/client.nim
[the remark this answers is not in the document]
chaitanyaprem: Then this needs to be fixed for all protocols in nwaku then, because looks like we are not catching it anywhere.

waku/waku_store/rpc_codec.nim
chaitanyaprem: Maybe this can be taken up as an improvement as part of separate PR. 

cc @Ivansete-status

apps/wakunode2/app.nim
```diff
@@ -366,9 +360,7 @@ proc startRestServer(
       "/relay endpoints are not available. Please check your configuration: --relay"
 
   ## Filter REST API
-  if conf.filternode  != "" and
-     app.node.wakuFilterClient != nil:
-
+  if conf.filternode != "" and app.node.wakuFilterClient != nil:
```
jm-clius, line 363: Unnecessary nph formatting revert?

waku/waku_store/rpc_codec.nim
```diff
@@ -3,257 +3,210 @@ when (NimMajor, NimMinor) < (1, 4):
...
-  if not ?pb.getField(1, contentTopic):
-    return err(ProtobufError.missingRequiredField("content_topic"))
-  ok(HistoryContentFilterRPC(contentTopic: contentTopic))
+  pb.write3(3, req.pubsubTopic)
 
-proc encode*(rpc: HistoryQueryRPC): ProtoBuffer =
-  var pb = initProtoBuffer()
-  pb.write3(2, rpc.pubsubTopic)
+  for contentTopic in req.contentTopics:
+    pb.write3(4, contentTopic)
```
jm-clius, line 24: Truly not sure how to keep track of this in future use of proto3 (other than to create our own library wrapper), but for proto3 we have to use `writePacked` for encoding repeated fields and `getPackedRepeatedField` for decoding. See https://github.com/waku-org/nwaku/pull/2509

jm-clius (changes requested):
Thanks! The only blocking comment I have is that repeated fields are packed in proto3 (see https://github.com/waku-org/nwaku/pull/2509). Longer term, we should probably write our own library as wrapper for minprotobuf or use another.

waku/waku_store/rpc_codec.nim
SionoiS: Ah yes. I remember vaguely something about compat. being broken otherwise.

apps/wakunode2/app.nim
SionoiS: Different version of nph maybe? `v0.5.1-0-gde5cd48` won't let me use 2 lines.

waku/waku_core/message/digest.nim
```diff
@@ -11,6 +11,11 @@ import ../topics, ./message
 
 type WakuMessageHash* = array[32, byte]
 
+const EmptyWakuMessageHash*: WakuMessageHash = [
+  0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+  0, 0, 0,
+]
```
Ivansete-status, line 17: Maybe this can work too:
```suggestion
const EmptyWakuMessageHash*: WakuMessageHash = newSeqWith(32, 0.byte)
```

waku/waku_store/common.nim
```diff
@@ -3,104 +3,115 @@ when (NimMajor, NimMinor) < (1, 4):
...
 
-type WakuStoreResult*[T] = Result[T, string]
-
-## Waku message digest
-
-type MessageDigest* = MDigest[256]
+  EmptyCursor*: WakuMessageHash = [
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+    0, 0, 0, 0,
+  ]
```
Ivansete-status, line 19: Maybe
```suggestion
  EmptyCursor*: WakuMessageHash = newSeqWith(32, 0.byte)
```

waku/waku_store/rpc_codec.nim
Ivansete-status: Very interesting @chaitanyaprem ! I think that is something we could look at but I don't see super urgency for now as we don't change the protobuf implementation very often

wdyt @jm-clius ?

tests/node/test_wakunode_store.nim
```diff
@@ -115,22 +119,22 @@ suite "Waku Store - End to End - Sorted Archive":
 
     asyncTest "Backward Pagination":
       # Given the history query is backward
-      historyQuery.direction = PagingDirection.BACKWARD
+      storeQuery.paginationForward = PagingDirection.BACKWARD
```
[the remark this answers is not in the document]
SionoiS: Names were taken as is from https://github.com/vacp2p/rfc/pull/665 but we can change them. @jm-clius ?

tests/node/test_wakunode_store.nim
```diff
@@ -546,81 +567,270 @@ suite "Waku Store - End to End - Unsorted Archive":
...
     check:
-      queryResponse.get().messages ==
-        @[
-          unsortedArchiveMessages[2],
-          unsortedArchiveMessages[0],
-          unsortedArchiveMessages[1],
-          unsortedArchiveMessages[4],
-          unsortedArchiveMessages[3],
-        ]
+      queryResponse.get().messages.len == 5
```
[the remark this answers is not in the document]
SionoiS: We do check the ordering L578-597 but each run the messages have different timestamp and hashes and since the messages returned are ordered by hash then the order is different.

We could hard code the messages timestamp but I feel it would make the test less useful.

tests/waku_store/test_resume.nim
```diff
@@ -1,332 +0,0 @@
-{.used.}
-
-import
```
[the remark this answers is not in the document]
SionoiS: It has been remove for now and this functionality will be part of the new Store-Sync protocol (which will use both Store and Sync).

waku/waku_archive/driver/sqlite_driver/cursor.nim
```diff
@@ -5,7 +5,7 @@ else:
 
 import ../../../waku_core, ../../common
 
-type DbCursor* = (Timestamp, seq[byte], PubsubTopic)
+type DbCursor* = (Timestamp, seq[byte], PubsubTopic, WakuMessageHash)
 
 proc toDbCursor*(c: ArchiveCursor): DbCursor =
-  (c.storeTime, @(c.digest.data), c.pubsubTopic)
+  (c.storeTime, @(c.digest.data), c.pubsubTopic, c.hash)
```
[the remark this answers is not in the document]
SionoiS: Yes for Store v3 but not for v2, so we have to keep everything for compatibility.

waku/waku_archive/driver/sqlite_driver/queries.nim
```diff
@@ -451,6 +487,30 @@ proc selectMessagesByHistoryQueryWithLimit*(
 ): DatabaseResult[
     seq[(PubsubTopic, WakuMessage, seq[byte], Timestamp, WakuMessageHash)]
 ] =
+  # Store v3 cursor is only the hash of the message
+  # Must first get the message timestamp before paginating by time
+  var cursorv2 = cursor
+  var cursorv3 = none((Timestamp, WakuMessageHash))
+  if cursor.isSome() and cursor.get()[3] != EmptyWakuMessageHash:
```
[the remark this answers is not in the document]
SionoiS: :thinking: tuples are initialized with default values no? If yes then it would `== EmptyWakuMessageHash`

waku/waku_core/message/digest.nim
SionoiS: `WakuMessageHash` is an `array` not a `seq`. I've not found a way to init arrays with multiple values sadly.

waku/waku_store/client.nim
```diff
@@ -27,214 +17,48 @@ type WakuStoreClient* = ref object
...
-  return resp.toAPI()
+  return ok(res)
 
 proc query*(
-    w: WakuStoreClient, req: HistoryQuery, peer: RemotePeerInfo
-): Future[HistoryResult] {.async, gcsafe.} =
-  return await w.sendHistoryQueryRPC(req, peer)
-
-# TODO: Move outside of the client
-when defined(waku_exp_store_resume):
```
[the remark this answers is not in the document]
SionoiS: It will be part of Store-Sync protocol!

waku/waku_store/common.nim
SionoiS: Same as above `array` can't be initialized like this AFAIK.

waku/waku_store/common.nim
```diff
@@ -3,104 +3,115 @@ when (NimMajor, NimMinor) < (1, 4):
 else:
   {.push raises: [].}
 
-import std/[options, sequtils], stew/results, stew/byteutils, nimcrypto/sha2
+import std/[options], stew/results
 import ../waku_core, ../common/paging
 
 const
-  WakuStoreCodec* = "/vac/waku/store/2.0.0-beta4"
+  WakuStoreCodec* = "/vac/waku/store-query/3.0.0"
```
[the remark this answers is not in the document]
SionoiS: I used the name from the spec. @jm-clius why?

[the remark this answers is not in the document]
chaitanyaprem: > Nitpick: why add the `-query` part?

Because that actually makes sense. :) 
The protocol is not used to Store a message in the node rather querying for data from a Store node.

waku/waku_store/common.nim
chaitanyaprem: Doesn't nim by default initialize array with 0's (kind of similar to C)?

tests/node/test_wakunode_store.nim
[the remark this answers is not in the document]
Ivansete-status: > We do check the ordering L578-597 but each run the messages have different timestamp and hashes and since the messages returned are ordered by hash then the order is different.
> 
> We could hard code the messages timestamp but I feel it would make the test less useful.

Okay I see,
Aside from the test comment itself, something important is that the following should still work: https://github.com/waku-org/nwaku/issues/2317
( cc @richard-ramos )

waku/waku_store/common.nim
SionoiS: > WakuMessageHash

ah yes! So `EmptyCursor = WakuMessageHash()` would probably work.

waku/waku_store/rpc_codec.nim
SionoiS: `[packed = true] can only be specified for repeated primitive fields.`

This is the error that @richard-ramos found while trying to pack repeated strings.

jm-clius (approved):
Approving, based on proto3 discussion with @SionoiS

waku/waku_store_legacy/client.nim
```diff
@@ -0,0 +1,242 @@
...
+
+proc new*(
+    T: type WakuStoreClient, peerManager: PeerManager, rng: ref rand.HmacDrbgContext
+): T =
+  WakuStoreClient(peerManager: peerManager, rng: rng)
+
+proc sendHistoryQueryRPC(
+    w: WakuStoreClient, req: HistoryQuery, peer: RemotePeerInfo
+): Future[HistoryResult] {.async, gcsafe.} =
+  let connOpt = await w.peerManager.dialPeer(peer, WakuStoreCodec)
```
Ivansete-status, line 42: Maybe sth to check in the future but I can't see why dialing on every query. I think it would be better to separate the `dialIn` from `query` operations :)

Ivansete-status (approved):
LGTM! Thank you for such a great PR! :100: 

Just added minor comment :)

waku/waku_store_legacy/client.nim
SionoiS: What would you envision as an alternative? Dialing for each request makes this part of the code stateless which simplify things.

Do you mean separate for code organization purpose?

PR #2435, opened 2024-02-14
fix(bindings): base64 payload and key for content topic

richard-ramos:
# Description
Various fixes to the bindings to issues identified while attempting to integrate nwaku with [waku-rust-bindings](https://github.com/waku-org/waku-rust-bindings/pull/87)

# Changes
- [ ] Stores the userData that's passed when `waku_set_event_callback` is invoked
- [ ] Adds the messageHash to the response when an event is generated
- [ ] Converts from/to base64 the message payloads
- [ ] Uses the key `contentTopic` instead of `content_topic` in the message JSON


library/events/json_message_event.nim
```diff
@@ -1,16 +1,59 @@
...
+    version: uint32(jsonContent["version"].getInt()),
+    timestamp: int64(jsonContent["timestamp"].getBiggestInt()),
+    ephemeral: jsonContent["ephemeral"].getBool(),
+    meta: Base64String(jsonContent["meta"].getStr()),
+  )
+
+proc toWakuMessage*(self: JsonMessage): WakuMessage =
+  let payloadRes = base64.decode(self.payload)
+  if payloadRes.isErr():
+    raise newException(ValueError, "invalid payload format: " & payloadRes.error)
```
SionoiS, line 37:
```suggestion
  let payload = base64.decode(self.payload).valueOr:
    raise newException(ValueError, "invalid payload format: " & error)
```

library/events/json_message_event.nim
```diff
@@ -1,16 +1,59 @@
...
+
+  let metaRes = base64.decode(self.meta)
+  if metaRes.isErr():
+    raise newException(ValueError, "invalid meta format: " & metaRes.error)
```
SionoiS, line 41:
```suggestion
  let meta = base64.decode(self.meta).valueOr:
    raise newException(ValueError, "invalid meta format: " & error)
```

SionoiS (approved):
LGTM thanks!

library/events/json_message_event.nim
Ivansete-status, line 37: Nitpick, if possible I'd avoid using exceptions:)

library/events/json_message_event.nim
```diff
@@ -1,16 +1,59 @@
...
+
+  WakuMessage(
+    payload: payloadRes.value,
+    meta: metaRes.value,
+    contentTopic: self.contentTopic,
+    version: uint32(self.version),
+    timestamp: self.timestamp,
+    ephemeral: self.ephemeral
+  )
```
Ivansete-status, line 50: At first sight I missed having the rln proof but I believe this is something we will add in the future when we expose some kind of `publish_rln` function

library/libwaku.nim
```diff
@@ -43,14 +45,14 @@ const RET_MISSING_CALLBACK: cint = 2
 proc relayEventCallback(ctx: ptr Context): WakuRelayHandler =
   return proc (pubsubTopic: PubsubTopic, msg: WakuMessage): Future[system.void]{.async.} =
     # Callback that hadles the Waku Relay events. i.e. messages or errors.
-    if not isNil(ctx[].eventCallback):
+    if not isNil(ctx[].eventCallback) and not isNil(ctx[].eventUserData):
```
Ivansete-status, line 48: Given that we now have two conditions that should be met, I wonder if we could split the validation in two, and apply some kind of  "return early pattern". The following article explains the idea much better:
https://medium.com/swlh/return-early-pattern-3d18a41bba8

lmk what do you think :)

Ivansete-status (approved):
LGTM! Thanks for it!
Just added a couple of nitpick comments xD

library/events/json_message_event.nim
richard-ramos: Fixed in https://github.com/waku-org/nwaku/pull/2435/commits/a4f7b88bacdbd30fe7a76654d4e3a732ddeb1f66

library/events/json_message_event.nim
richard-ramos: Fixed in https://github.com/waku-org/nwaku/pull/2435/commits/a4f7b88bacdbd30fe7a76654d4e3a732ddeb1f66

library/events/json_message_event.nim
richard-ramos: Replaced by a Result in https://github.com/waku-org/nwaku/pull/2435/commits/a4f7b88bacdbd30fe7a76654d4e3a732ddeb1f66

library/events/json_message_event.nim
richard-ramos: Ended up adding the proof in https://github.com/waku-org/nwaku/pull/2435/commits/a4f7b88bacdbd30fe7a76654d4e3a732ddeb1f66 so it's now 'future-ready' for when we include rln functionality to bindings

library/libwaku.nim
richard-ramos: Sounds good! I removed the validation from the `if` and executed those following the 'return early' pattern as described in the article you shared.
https://github.com/waku-org/nwaku/pull/2435/commits/a4f7b88bacdbd30fe7a76654d4e3a732ddeb1f66

PR #2445, opened 2024-02-17
fix: Better error logs in sharding mismatch and avoid checking clusterId if clusterId == 0

Ivansete-status:
## Description
1. Enhance error management inside `isClusterMismatched`
2. Avoid validation inside `isClusterMismatched` if the node is using `clusterId == 0`. This aims to allow the node to start successfully and avoid the appearance of "cluster id mismatch configured shards" and the `quit(QuitFailure)`. In other words, the node couldn't start due to this proc even though the `clusterId` param was not being set.

### DIFF
--- a/apps/wakunode2/internal_config.nim
+++ b/apps/wakunode2/internal_config.nim
@@ -57,6 +57,8 @@ proc enrConfiguration*(conf: WakuNodeConf, netConfig: NetConfig, key: crypto.Pri
       return err($recordRes.error)
     else: recordRes.get()
 
+  debug "enrConfiguration result", record = $record
+
   return ok(record)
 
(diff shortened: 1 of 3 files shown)


gabrielmer:
This PR would also reintroduce https://github.com/waku-org/nwaku/issues/2311 , I think we should allow nodes to start without subscribed shards and subscribe later dynamically

SionoiS:
> allow nodes to start without subscribed shards and subscribe later dynamically

I would be very ergonomic indeed.

Ivansete-status:
Thanks for the comments guys!
Disregard this PR.
For some reason, last week we had many restarts in `waku.fleet` due to this proc, and the PR was motivated to overcome that. It also happened to me locally but now it works well :)

image

Anyways, I'll close this.

PR #2446, opened 2024-02-19
feat(bindings): generate a random private key

richard-ramos:
# Description
Generates a random private key if config attribute is not defined as described in https://rfc.vac.dev/spec/36/#jsonconfig-type


SionoiS (approved):

Ivansete-status (approved):
Wonderful! Thanks for it! :100:

PR #2447, opened 2024-02-19
refactor(bindings): simplify error handling

richard-ramos:
# Description
While https://rfc.vac.dev/spec/36/#jsonresponse-type describes a JSON structure meant to contain the error description, having changed the bindings to use callback makes it not necessary to have such JSON, thus we can simplify the code by returning a `string` for the error messages (also makes it simpler for language specific bindings to not worry about a json structure, but just error with the returned message)


SionoiS (approved):
LGTM thanks!

Ivansete-status (approved):
Wow! Thanks for it!

Issue #2425, opened 2024-02-13
feat: reference implementation for Store v3 protocol


jm-clius:
### Problem

This feature forms the reference implementation part of Waku Research's new Store protocol proposal.
The overall research task is tracked in https://github.com/waku-org/research/issues/81
The parallel RFC proposal is documented in https://github.com/vacp2p/rfc/pull/665

### Acceptance criteria

- [x] functioning reference implementation of https://github.com/vacp2p/rfc/pull/665
- [x] the current Store protocol (`/vac/waku/store/2.0.0-beta4`) must be mounted in parallel to allow backwards compatibility

cc @SionoiS

SionoiS:
PR https://github.com/waku-org/nwaku/pull/2431

PR #2451, opened 2024-02-19
feat: archive update for store v3

SionoiS:
# Description
I tried a bunch of things that didn't work. In the end I simply added hashed to requests and responses. This approach will require a good clean-up after store v2 is removed.

I simplified how message are validated since we never exposed the validation even if the intent seamed to be to make it configurable. Do we want fully configurable validation? What kind of messages do we reject exactly?

There are also couple of comments that require reviewers input.

# Changes
- [x] refactored `WakuArchive`
- [x] archive cursor new use timestamp and msg hash
- [x] simplified msg validation
- [x] compatibility with Store v2
- [x] update drivers
- [x] update tests

Tracking #2425

### DIFF
--- a/tests/waku_archive/test_driver_postgres_query.nim
+++ b/tests/waku_archive/test_driver_postgres_query.nim
@@ -8,7 +8,6 @@ import
 import
   ../../../waku/waku_archive,
   ../../../waku/waku_archive/driver as driver_module,
-  ../../../waku/waku_archive/driver/builder,
   ../../../waku/waku_archive/driver/postgres_driver,
   ../../../waku/waku_core,
   ../../../waku/waku_core/message/digest,
@@ -670,10 +669,10 @@ suite "Postgres driver - queries":
     ]
     var messages = expected
 
-    let hashes = messages.mapIt(computeMessageHash(DefaultPubsubTopic, it))
-
     shuffle(messages)
     debug "randomized message insertion sequence", sequence=messages.mapIt(it.payload)
+
+    let hashes = messages.mapIt(computeMessageHash(DefaultPubsubTopic, it))
 
     for (msg, hash) in messages.zip(hashes):
       require (await driver.put(DefaultPubsubTopic, msg, computeDigest(msg), hash, msg.timestamp)).isOk()
--- a/tests/waku_archive/test_driver_postgres_query.nim
+++ b/tests/waku_archive/test_driver_postgres_query.nim
@@ -33,7 +33,8 @@ proc computeTestCursor(pubsubTopic: PubsubTopic, message: WakuMessage): ArchiveC
     pubsubTopic: pubsubTopic,
     senderTime: message.timestamp,
     storeTime: message.timestamp,
-    digest: computeDigest(message)
+    digest: computeDigest(message),
+    hash: computeMessageHash(pubsubTopic, message)
   )
 
 suite "Postgres driver - queries":
@@ -651,6 +652,45 @@ suite "Postgres driver - queries":
     let filteredMessages = res.tryGet().mapIt(it[1])
     check:
       filteredMessages == expectedMessages[4..5].reversed()
+
+  asyncTest "only hashes - descending order":
+    ## Given
+    let timeOrigin = now()
+    var expected = @[
+      fakeWakuMessage(@[byte 0], ts=ts(00, timeOrigin)),
+      fakeWakuMessage(@[byte 1], ts=ts(10, timeOrigin)),
+      fakeWakuMessage(@[byte 2], ts=ts(20, timeOrigin)),
+      fakeWakuMessage(@[byte 3], ts=ts(30, timeOrigin)),
+      fakeWakuMessage(@[byte 4],  ts=ts(40, timeOrigin)),
+      fakeWakuMessage(@[byte 5],  ts=ts(50, timeOrigin)),
+      fakeWakuMessage(@[byte 6],  ts=ts(60, timeOrigin)),
+      fakeWakuMessage(@[byte 7],  ts=ts(70, timeOrigin)),
+      fakeWakuMessage(@[byte 8],  ts=ts(80, timeOrigin)),
+      fakeWakuMessage(@[byte 9],  ts=ts(90, timeOrigin)),
+    ]
+    var messages = expected
+
+    let hashes = messages.mapIt(computeMessageHash(DefaultPubsubTopic, it))
+
+    shuffle(messages)
+    debug "randomized message insertion sequence", sequence=messages.mapIt(it.payload)
+
+    for (msg, hash) in messages.zip(hashes):
+      require (await driver.put(DefaultPubsubTopic, msg, computeDigest(msg), hash, msg.timestamp)).isOk()
+
+    ## When
+    let res = await driver.getMessages(
+      hashes=hashes,
+      ascendingOrder=false
+    )
+
+    ## Then
+    assert res.isOk(), res.error
+
+    let expectedMessages = expected.reversed()
+    let filteredMessages = res.tryGet().mapIt(it[1])
+    check:
+      filteredMessages == expectedMessages
 
   asyncTest "start time only":
     ## Given
--- a/waku/waku_archive/archive.nim
+++ b/waku/waku_archive/archive.nim
@@ -4,22 +4,15 @@ else:
   {.push raises: [].}
 
 import
-  std/[tables, times, sequtils, options, algorithm, strutils],
+  std/[times, options, sequtils, strutils, algorithm],
   stew/[results, byteutils],
   chronicles,
   chronos,
-  regex,
   metrics
 import
-  ../common/[
-    databases/dburl,
-    databases/db_sqlite,
-    paging
-  ],
+  ../common/paging,
   ./driver,
   ./retention_policy,
-  ./retention_policy/retention_policy_capacity,
-  ./retention_policy/retention_policy_time,
   ../waku_core,
   ../waku_core/message/digest,
   ./common,
@@ -29,28 +22,46 @@ logScope:
   topics = "waku archive"
 
 const
-  DefaultPageSize*: uint = 20
-  MaxPageSize*: uint = 100
+  #TODO if waku message are 150KiB max are those numbers still appropriate???
+  DefaultPageSize*: uint = 20 # 2.93 MiB
+  MaxPageSize*: uint = 100 # 14.65 MiB
 
-## Message validation
+# Retention policy
+const WakuArchiveDefaultRetentionPolicyInterval* = chronos.minutes(30)
 
-type
-  MessageValidator* = ref object of RootObj
+# Metrics reporting
+const WakuArchiveDefaultMetricsReportInterval* = chronos.minutes(1)
 
-  ValidationResult* = Result[void, string]
+# Message validation
+# 20 seconds maximum allowable sender timestamp "drift"
+const MaxMessageTimestampVariance* = getNanoSecondTime(20) 
 
-method validate*(validator: MessageValidator, msg: WakuMessage): ValidationResult {.base.} = discard
+## Archive
 
-# Default message validator
+type WakuArchive* = ref object
+  driver: ArchiveDriver
 
-const MaxMessageTimestampVariance* = getNanoSecondTime(20) # 20 seconds maximum allowable sender timestamp "drift"
+  retentionPolicy: Option[RetentionPolicy]
 
-type DefaultMessageValidator* = ref object of MessageValidator
+  retentionPolicyHandle: Future[void]
+  metricsHandle: Future[void]
 
-method validate*(validator: DefaultMessageValidator, msg: WakuMessage): ValidationResult =
-  if msg.timestamp == 0:
-    return ok()
+proc new*(T: type WakuArchive,
+          driver: ArchiveDriver,
+          retentionPolicy = none(RetentionPolicy)):
+          Result[T, string] =
+  if driver.isNil():
+    return err("archive driver is Nil")
 
+  let archive =
+    WakuArchive(
+      driver: driver,
+      retentionPolicy: retentionPolicy,
+    )
+
+  return ok(archive)
+
+proc validate*(self: WakuArchive, msg: WakuMessage): Result[void, string] =
   let
     now = getNanosecondTime(getTime().toUnixFloat())
     lowerBound = now - MaxMessageTimestampVariance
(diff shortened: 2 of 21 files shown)


SionoiS:
Will have to wait for #2506 then resolve and review again.

waku/waku_archive/archive.nim
```diff
@@ -29,28 +22,43 @@ logScope:
...
-type DefaultMessageValidator* = ref object of MessageValidator
+  let archive =
+    WakuArchive(
+      driver: driver,
+      retentionPolicy: retentionPolicy,
+    )
 
-method validate*(validator: DefaultMessageValidator, msg: WakuMessage): ValidationResult =
-  if msg.timestamp == 0:
-    return ok()
```
Ivansete-status, line 52: Is timestamp becoming mandatory in the new Store protocol version (v3.) Sorry I couldn't fine the new Store definition

waku/waku_archive/archive.nim
```diff
@@ -64,186 +72,145 @@ method validate*(validator: DefaultMessageValidator, msg: WakuMessage): Validati
...
-      if "duplicate key value violates unique constraint" in putRes.error:
-        trace "failed to insert message", err=putRes.error
-      else:
-        debug "failed to insert message", err=putRes.error
-      waku_archive_errors.inc(labelValues = [insertFailure])
-
+  (await self.driver.put(pubsubTopic, msg, msgDigest, msgHash, msg.timestamp)).isOkOr:
+    waku_archive_errors.inc(labelValues = [insertFailure])
+    
+    # TODO diff log lvl per error value WTF is this???
```
Ivansete-status, line 102: haha yes I agree that it looks weird.

Let me bring some reasoning about that:

It is mostly to prevent spamming the logs when multiple nodes are connected to the same database. In that case, the message cannot be inserted but is an expected "insert error" and therefore we reduce its visibility by having the log in `trace` level.

The change was introduced in: https://github.com/waku-org/nwaku/commit/574cdf555adcdfc77291902d413bb8884bc0c925
```suggestion
```

waku/waku_archive/driver.nim
```diff
@@ -39,6 +39,7 @@ method getMessages*(driver: ArchiveDriver,
                     cursor = none(ArchiveCursor),
                     startTime = none(Timestamp),
                     endTime = none(Timestamp),
+                    hashes: seq[WakuMessageHash] = @[],
```
Ivansete-status, line 42: A different way to do the same. Not need to change if don't want
```suggestion
                    hashes = newSeq[WakuMessageHash](0),
```

waku/waku_archive/driver/postgres_driver/postgres_driver.nim
```diff
@@ -401,11 +427,15 @@ method getMessages*(s: PostgresDriver,
...
+                    hashes: seq[WakuMessageHash] = @[],
                     maxPageSize = DefaultPageSize,
                     ascendingOrder = true):
                     Future[ArchiveDriverResult[seq[ArchiveRow]]] {.async.} =
-
-  if contentTopicSeq.len == 1 and
+  let hexHashes = hashes.mapIt(toHex(it))
+  
+  # TODO find why prepared statement returns no msg
+  #[ if contentTopicSeq.len == 1 and
```
Ivansete-status, line 437: Interesting. We need to make sure the prepared statements queries still work. DM me if need me to review something :P

waku/waku_archive/driver/sqlite_driver/sqlite_driver.nim
```diff
@@ -90,6 +90,7 @@ method getMessages*(s: SqliteDriver,
                     cursor = none(ArchiveCursor),
                     startTime = none(Timestamp),
                     endTime = none(Timestamp),
+                    hashes: seq[WakuMessageHash] = @[],
```
Ivansete-status, line 93: nitpick comment:
```suggestion
                    hashes = newSeq[WakuMessageHash](0),
```

Ivansete-status (changes requested):
Overall it looks great! I'm just adding a few points that I think we need to double-check

waku/waku_archive/archive.nim
SionoiS: Make sense! I will add a comment.

waku/waku_archive/archive.nim
SionoiS: Yes this is part of the Store v3 specification https://github.com/vacp2p/rfc/tree/docs/new-store-v3-protocol/content/docs/rfcs/13#waku-message-store-eligibility

waku/waku_archive/archive.nim
```diff
@@ -64,186 +72,145 @@ method validate*(validator: DefaultMessageValidator, msg: WakuMessage): Validati
...
+proc loopReportStoredMessagesMetric(self: WakuArchive) {.async.} =
+  if self.driver.isNil(): #TODO does this really happen?
+    error "driver is Nil in loopReportStoredMessagesMetric"
+    return
 
-proc loopReportStoredMessagesMetric*(w: WakuArchive):
-                                     Future[Result[void, string]] {.async.} =
-  if w.driver.isNil():
-    return err("driver is Nil in loopReportStoredMessagesMetric")
+  while self.started:
```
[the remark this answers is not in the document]
SionoiS: I'll refactor.

waku/waku_archive/archive.nim
```diff
@@ -29,28 +22,43 @@ logScope:
   topics = "waku archive"
 
 const
-  DefaultPageSize*: uint = 20
-  MaxPageSize*: uint = 100
+  #TODO if waku message are 150KiB max are those numbers still appropriate???
```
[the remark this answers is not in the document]
SionoiS: If Status msg are max 1MiB that means at max page size that's a 100MiB response + overhead is that ok?

waku/waku_archive/driver/postgres_driver/postgres_driver.nim
SionoiS: ops forgot about this one :P

waku/waku_archive/archive.nim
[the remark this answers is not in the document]
Ivansete-status: > If Status msgs are max 1MiB that means at max page size that's a 100MiB response + overhead is that ok?

Yes correct. Reference for other readers: https://github.com/waku-org/nwaku/blob/a9d0e48164008b15e9669e0001fcf1a2e18767ca/waku/waku_store/rpc_codec.nim#L16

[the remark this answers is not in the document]
SionoiS: Sorry what I meant to ask is, 100Mib response is huge should we not reduce that a bit?

SionoiS:
I messed my rebase...

waku/waku_archive/archive.nim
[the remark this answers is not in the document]
Ivansete-status: > Sorry what I meant to ask is, 100Mib response is huge should we not reduce that?

ah yes, good point and good catch! I agree that it sounds huge but I suggest following that change in a separate PR.

waku/waku_archive/driver/postgres_driver/postgres_driver.nim
```diff
@@ -147,10 +151,16 @@ proc rowCallbackImpl(pqResult: ptr PGresult,
     wakuMessage.contentTopic = contentTopic
     wakuMessage.payload = @(payload.toOpenArrayByte(0, payload.high))
 
+    var msgHash: WakuMessageHash
+    let byteCount = copyFrom(msgHash, hashHex.toOpenArrayByte(0, 31))
+    assert byteCount == 32
```
Ivansete-status, line 156: This seems quite similar as the following , present at `waku_archive/common.nim`. Maybe we can make use it somehow:
```
proc fromBytes*(T: type MessageDigest, src: seq[byte]): T =
```

waku/waku_archive/archive.nim
```diff
@@ -64,186 +75,148 @@ method validate*(validator: DefaultMessageValidator, msg: WakuMessage): Validati
...
-  return ok()
+  if self.retentionPolicy.isSome() and not self.retentionPolicyHandle.isNil():
+    futures.add(self.retentionPolicyHandle.cancelAndWait())
 
-proc start*(self: WakuArchive) {.async.} =
-  ## TODO: better control the Result in case of error. Now it is ignored
-  self.retPolicyFut = self.loopApplyRetentionPolicy()
-  self.retMetricsRepFut = self.loopReportStoredMessagesMetric()
+  if not self.metricsHandle.isNil:
+    futures.add(self.metricsHandle.cancelAndWait())
```
Ivansete-status, line 220: Just to learn, what is the benefit of following that approach?

waku/waku_archive/driver/postgres_driver/postgres_driver.nim
```diff
@@ -379,10 +397,12 @@ method getMessages*(s: PostgresDriver,
                     cursor = none(ArchiveCursor),
                     startTime = none(Timestamp),
                     endTime = none(Timestamp),
+                    hashes: seq[WakuMessageHash] = @[],
```
Ivansete-status, line 400: nitpick comment:)
```suggestion
                    hashes = newSeq[WakuMessageHash](0),
```

waku/waku_archive/driver/sqlite_driver/queries.nim
```diff
@@ -58,6 +58,17 @@ proc queryRowDigestCallback(s: ptr sqlite3_stmt, digestCol: cint): seq[byte] =
...
 
+proc queryRowWakuMessageHashCallback(s: ptr sqlite3_stmt, hashCol: cint): WakuMessageHash =
+  let
+    hashPointer = cast[ptr UncheckedArray[byte]](sqlite3_column_blob(s, hashCol))
+    hashLength = sqlite3_column_bytes(s, hashCol)
+
+  var hash: WakuMessageHash
+
+  let copiedBytes = copyFrom(hash, toOpenArray(hashPointer, 0, hashLength-1))
+  assert copiedBytes == 32
```
Ivansete-status, line 69: I think I've seen this block somewhere else ;P

Ivansete-status (approved):
LGTM! Thanks for the patience!

waku/waku_archive/archive.nim
SionoiS: Seams like a good way to graceful shutdown when there are many futures to cancel.

- concurrent cancel on all futures
- await but prevent cancellation of the shutdown process

I found it at https://github.com/status-im/nim-eth/blob/d66a29db7ca4372dba116928f979e92cb7f7661f/eth/p2p/discoveryv5/protocol.nim#L1039

tests/waku_archive/test_driver_postgres_query.nim
```diff
@@ -652,6 +653,45 @@ suite "Postgres driver - queries":
...
+    ]
+    var messages = expected
+
+    let hashes = messages.mapIt(computeMessageHash(DefaultPubsubTopic, it))
+
+    shuffle(messages)
+    debug "randomized message insertion sequence", sequence=messages.mapIt(it.payload)
+
+    for (msg, hash) in messages.zip(hashes):
+      require (await driver.put(DefaultPubsubTopic, msg, computeDigest(msg), hash, msg.timestamp)).isOk()
```
gabrielmer, line 679: Aren't we inserting here messages with hashes that don't correspond to them?

The hashes maintain the original ordering while the messages get shuffled. And we then create tuples out of them with `zip()`

Is this intentional? Maybe not understanding something

gabrielmer (approved):
Amazing work! 🔥 🔥 🔥 
Thanks so much!

waku/waku_archive/driver/sqlite_driver/queries.nim
```diff
@@ -58,6 +58,17 @@ proc queryRowDigestCallback(s: ptr sqlite3_stmt, digestCol: cint): seq[byte] =
...
+
+  hash
```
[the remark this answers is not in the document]
SionoiS: True, this code has not been touched in some time it seams. I'll refresh it.

tests/waku_archive/test_driver_postgres_query.nim
SionoiS: Ah good catch! It only passed because the set of messages was the same....

waku/waku_archive/archive.nim
```diff
@@ -64,186 +75,148 @@ method validate*(validator: DefaultMessageValidator, msg: WakuMessage): Validati
...
+  debug "executing message retention policy"
 
-proc loopApplyRetentionPolicy*(w: WakuArchive):
-                               Future[Result[void, string]] {.async.} =
+  let policy = self.retentionPolicy.get()
 
-  if w.retentionPolicy.isNil():
-    return err("retentionPolicy is Nil in executeMessageRetentionPolicy")
+  while true:
+    await sleepAsync(WakuArchiveDefaultRetentionPolicyInterval)
```
jm-clius, line 191: Would moving this sleep mean that the retention policy will execute only 30 mins after node startup? Is this the desired behaviour? To me it makes sense to run at startup and then every 30 mins.

waku/waku_archive/driver/postgres_driver/postgres_driver.nim
```diff
@@ -257,6 +266,12 @@ proc getMessagesArbitraryQuery(s: PostgresDriver,
     for t in contentTopic:
       args.add(t)
 
+  if hexHashes.len > 0:
+    let cstmt = "messageHash IN (" & "?".repeat(hexHashes.len).join(",") & ")"
```
jm-clius, line 270: Do we have an index built on this column? Would it be necessary for querying on message hash to be efficient? Perhaps we'll need a benchmark on a rather large DB (future work).

jm-clius (approved):
Thanks - this was quite a detailed effort. I've added a few comments, but approving since none are blocking. One general comment: there do seem to be quite a few changes unrelated to what is exactly needed for archive queries to support v3. I get the impulse, as there's clearly a lot that can be improved, but may be good to separate into separate PRs as much as possible in future to derisk and make reviewing a bit easier. :)

waku/waku_archive/archive.nim
```diff
@@ -64,186 +75,148 @@ method validate*(validator: DefaultMessageValidator, msg: WakuMessage): Validati
...
   let insertStartTime = getTime().toUnixFloat()
 
-  block:
-    let
-      msgDigest = computeDigest(msg)
-      msgHash = computeMessageHash(pubsubTopic, msg)
-      msgDigestHex = toHex(msgDigest.data)
-      msgHashHex = toHex(msgHash)
-      msgReceivedTime = if msg.timestamp > 0: msg.timestamp
-                        else: getNanosecondTime(getTime().toUnixFloat())
```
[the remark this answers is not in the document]
SionoiS: Either we refuse messages without timestamp which may cause problems now OR we have problems later when we try to sync a message with different timestamps.

waku/waku_archive/driver/postgres_driver/postgres_driver.nim
SionoiS: I can't find any indexing.

Yes testing will tell us what to expect performance wise. We may have to use LMDB in the end (I hope not).

SionoiS:
> there do seem to be quite a few changes unrelated to what is exactly needed for archive queries to support v3

Yes, some part of the code was not up to our current standards. 

> I get the impulse, as there's clearly a lot that can be improved, but may be good to separate into separate PRs as much as possible in future to derisk and make reviewing a bit easier. :)

Guilty :P I need to get better at this.

waku/waku_archive/archive.nim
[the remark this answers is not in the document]
SionoiS: I'll change it back to the previous validation method.

PR #2463, opened 2024-02-20
refactor: new proc to foster different size retention policy implementations

Ivansete-status:
## Description
The new proc, `decreaseDatabaseSize`, will have different implementations per driver. For example, in future commits, we will implement a size retention policy thanks to partitions management, in Postgres.

## Changes

- [x] New method `decreaseDatabaseSize` in `driver.nim`.
- [x] Split the `decreaseDatabaseSize` implementation in different drivers. Notice that in upcoming PRs we will change the implementation in `postgres_driver.nim` per partition truncation.

### DIFF
--- a/tests/waku_archive/test_retention_policy.nim
+++ b/tests/waku_archive/test_retention_policy.nim
@@ -29,7 +29,7 @@ suite "Waku Archive - Retention policy":
 
     let driver = newSqliteArchiveDriver()
 
-    let retentionPolicy: RetentionPolicy = CapacityRetentionPolicy.init(capacity=capacity)
+    let retentionPolicy: RetentionPolicy = CapacityRetentionPolicy.new(capacity=capacity)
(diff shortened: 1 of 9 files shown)


SionoiS (approved):
LGTM!

Very useful policy.

Do we know how long does it take? Rough estimate?

Ivansete-status:
> LGTM!
> 
> Very useful policy.
> 
> Do we know how long does it take? Rough estimate?

Thanks for the review!
This week I'll submit a PR to tackle the "partition" approach in _Postgres_. The implementation is done but first of all, I will submit another previous PR with the database migration logic (pending implementation.)

gabrielmer (approved):
Looks great! Thanks so much! 🤩

Ivansete-status:
> @Ivansete-status I meant how long to reduce the DB by 20%. For ~1 million messages.

ah ok! I haven't measured the time but it is much faster than deleting rows. By removing partitions we directly truncate many rows at once.

PR #2467, opened 2024-02-21
chore(tests): Simplify imports

AlejandroCabeza:
# Description
Simplify test imports

# Changes
- [X] Add missing imports to `test_all` files
- [X] Fix failing tests (due to not being called)
- [X] Simplify `all_test_waku` using `test_all` files


tests/waku_node/test_wakunode_relay_rln.nim
rymnc, line 1: my intuition is that this test was never imported anywhere and that is why we could not catch it while merging the rln-v2 changes m

Ivansete-status:
@AlejandroCabeza - thanks for tackling that. In a nwaku PM session we considered that we did something similar in separate PRs and therefore it got a bit obsolete:)
Feel free to reopen if you consider any important missing change

PR #2476, opened 2024-02-23
fix: Make `getNanosecondTime` work on `i386` cpu

EmilIvanichkovv:
# Description
The function [getNanosecondTime](https://github.com/waku-org/nwaku/blob/master/waku/waku_core/time.nim#L12) does not work on i386 CPU, so this is a suggestion of a fix.

# Changes
Check the current CPU in compile time and use the correct type.

## Issue
closes # https://github.com/waku-org/nwaku/issues/2475


Ivansete-status (approved):
LGTM! Thanks for it!

May also include how to properly set the compilation flag in the PR description plz?

NagyZoltanPeter (approved):
Nice one! Thank you.

waku/waku_core/time.nim
```diff
@@ -10,7 +10,11 @@ import
 type Timestamp* = int64 # A nanosecond precision timestamp
 
 proc getNanosecondTime*[T: SomeNumber](timeInSeconds: T): Timestamp =
-  var ns = Timestamp(timeInSeconds * 1_000_000_000.T)
+  var ns: Timestamp
```
arnetheduck, line 13: seems more simple to change this function to `getNanosecondTime*(timeInSeconds: int64)`?

EmilIvanichkovv: Sure, I can do that. However, I was hesitant to make changes in a lot of places to the codebase because there are several instances in the code where different numeric types are passed to the function that will trigger `type mismatch` errors. That's why I preferred to keep it simple and just handle this problem within the function without casting types all over the project.

SionoiS (approved):
LGTM thanks

SionoiS:
One problem I thought about is that now we'll never have precise timestamps. Maybe a problem for Waku Sync in the future, I'm not sure yet. :thinking:

EmilIvanichkovv:
> One problem I thought about is that now we'll never have precise timestamps. Maybe a problem for Waku Sync in the future, I'm not sure yet. 🤔

I think you are right, so I updated my changes

waku/waku_core/time.nim
```diff
@@ -9,8 +9,8 @@ import
 
 type Timestamp* = int64 # A nanosecond precision timestamp
 
-proc getNanosecondTime*[T: SomeNumber](timeInSeconds: T): Timestamp =
-  var ns = Timestamp(timeInSeconds * 1_000_000_000.T)
+proc getNanosecondTime*(timeInSeconds: int64 | float64): Timestamp =
+  let ns = Timestamp(timeInSeconds * 1_000_000_000)
   return ns
```
PetarKirov, line 14: Looking at the previous version of the diff (https://github.com/waku-org/nwaku/commit/9bb2f3bbd55f4b1ce738d099ecd49be7bfb8e3a9), there were 12 instances of `getNanosecondTime(getTime().toUnixFloat())`, which I think can be better served by the following function:

```nim
proc getNowInNanosecondTime*(): Timestamp =
  ## Returns the current time with nanosecond precision and range of
  ## 1970 +/- 200 years.

  let now = getTime()
  #            ([-6.4e18 .. -6.4e18]   + [ 0 .. 10^9 ]) ⊂ [-6.5e18 .. -6.5e18]
  #             [-6.5e18 .. -6.5e18] ⊂ [ -9.2e9 .. 9.2e9 ]
  #             [ -9.2e9 .. -9.2e9 ] ⊂ [ -2^63  .. 2^63  ]
  # which fits in [ -9.2e9 .. -9.2e9 ]
  let ns = now.seconds * 1_000_000_000 + now.nanoseconds
  return ns
```

EmilIvanichkovv: Well, this seems reasonable. I am tempted to use `getNowInNanosecondTime` instead of `getNanosecondTime(getTime().toUnixFloat())`

EmilIvanichkovv:
Closing in favour of https://github.com/waku-org/nwaku/pull/2492

Issue #2441, opened 2024-02-16
chore: migrate node initialization logic from app.nim


gabrielmer:
### Background
There's logic related to node configuration and initialization in `app.nim` when it should be part of the core logic of a Waku Node.

This causes code duplication in `libwaku`.

### Acceptance criteria
A Waku Node's configuration and initialization logic should be decoupled from `app.nim`

PR #2479, opened 2024-02-27
chore: moving node initialization code to node_factory.nim

gabrielmer:
# Description
As a first step towards refactoring `app.nim`, we're moving code related to node initialization to a new factory/middleware layer. 

A new directory named `waku/factory` was created, which will contain all logic related to this functionality.
In addition, introduced a file named `node_factory.nim`, which will contain the core logic of this layer.
The idea is to be able to pass down node configurations and get a node in response.

A PR will follow extending the functionalities in `node_factory` beyond moving existing code.

# Changes

- [x] creating `factory` directory and `factory/node_factory.nim` file
- [x] moving node initialization procs from `app.nim` to `factory/node_factory.nim`
- [x] moving `internal_config.nim`, `external_config.nim` and`wakunode2_validator_signed.nim` from `apps/wakunode2` to `waku/factory`
- [x] moving `builder.nim` from `waku/node` to `waku/factory` directory 
- [x] renaming `wakunode2_validator_signed.nim` to `validator_signed.nim`

## Issue
 #2441


waku/node/node_factory.nim
```diff
@@ -0,0 +1,374 @@
...
+  ../waku_filter,
+  ../waku_filter_v2,
+  ../waku_peer_exchange,
+  ../waku_lightpush/common,
+  ../waku_archive/driver/builder,
+  ../waku_archive/retention_policy/builder,
+  ../common/utils/parse_size_units,
+  ../../apps/wakunode2/external_config,
+  ../../apps/wakunode2/internal_config,
+  ../../apps/wakunode2/wakunode2_validator_signed
```
gabrielmer, line 32: I have mixed feelings regarding importing code from the `apps` directory.
On one hand, the node's configuration parameters are defined there and we need them in order to create a node.

On the other, code related to the app should be separate from the the waku node modules.

Not sure if we should move the configs out of the `app` directory or what would be the most appropriate way to go

Ivansete-status: Good point! What if we move the `external_config`, `internal_config`, and `wakunode2_validator_signed` modules into the folder `waky/node/`?

gabrielmer: Sounds good :))

SionoiS (approved):
LGTM

I would avoid naming things node. I suggest `factory` instead of `node_factory` and maybe naming the folder initialization (or init).

waku/node/node_factory.nim
gabrielmer: So took a look and was able to move `internal_config.nim` and `wakunode2_validator_signed` but moving `external_config.nim` caused some conflicts because of its code related to parsing CLI commands, which also makes sense that is part of the `apps` directory.

Ideally we should decouple a node's configuration options to the application's CLI parameters.
The decoupling would make sense only if we find a way to do it without duplicating everything. So I left `external_config.nim` in `apps` until we find a better solution.

waku/node/validator_signed.nim
```diff
@@ -19,9 +19,9 @@ import
 const MessageWindowInSec = 5*60 # +- 5 minutes
 
 import
```
alrevuelta, line 21: unsure if @jm-clius agrees.

gabrielmer: Better having the signed validator in `apps`? the idea is to have all the pieces to build a Waku node with standard features under `waku/node` (but without adding it to the `WakuNode` object itself).

Thought that a signed validator entered into the set of "standard features"

jm-clius: However, these elements (app-level validation, app's DB config, etc.) were moved out of the node for a reason. This (incorrectly) brings back awareness of the app environment into the node object, while the app environment should know about the node and not the other way around. Amongst other things, you're now seeing a lot of interesting circular imports (such as `internal_config` depending on `waku_node`, while `waku_node` also depends on `internal_config`).

Perhaps the solution is rather to raise the bindings to the app level and, of course, create a convenient app-level node factory or middleware.

SionoiS: > the solution is rather to raise the bindings to the app level and, of course, create a convenient app-level node factory or middleware.

:+1: !

gabrielmer: Makes sense, maybe all of this should be at app level.

Notice that here we did not change the node object at all - we only moved the procs that initialize nodes from the `app` file to a `node_factory` file inside `waku/node` directory which isn't used by any other file related to the node.

Or in other words, we're attaching to `waku_node` a separate "kit" to easily create a node. So once you import `waku_node` you get both the basic procs to do everything from scratch and a separate already built "node factory" with standard implementations.

Regarding the circular imports, notice that those two are different files. `internal_config` depends on `waku/node/waku_node.nim` which has the actual `WakuNode` implementation. And then there's `waku/waku_node.nim` that imports files from the `waku/node` directory and exports them as a bundle so apps don't have to import all the files individually.
Here we're adding the node factory to the second one, to `waku/waku_node.nim`. But we are not touching the node's implementation.

Maybe we should have all of this at the app level. But the only thing that would change here is the file path, the logic is already separate.

jm-clius: Right, good point about the import here - `waku/waku_node.nim` is just a barrel import. However, it remains a bit weird that there are now "app" imports in modules that live very closely with the node. It indeed only affects import and directory structure afaics, but I guess what we need is:
a) node (the wrapper around libp2p switch)
b) app middleware (config, factory, validators, etc. that's used by all apps, including ones using the bindings)
c) specific apps, e.g. wakunode2 - bare minimum to use the app middleware to create a specific type of waku application, such as `wakunode2` binary.

This layered approach should also reflect in the import structure (outer app can import app middleware can import node, but never the other way).

gabrielmer: Sounds great! Really like that idea of a new "middleware" layer - agree 100% that it's a much more logically sound structure.

Will re-organize the it like that.
Thanks so much! 🤩

NagyZoltanPeter:
> 
> > I would avoid naming things node. I suggest `factory` instead of `node_factory` and maybe naming the folder initialization (or init).
> 
> Mmm makes sense, having the word node is kind of redundant. On the other hand, it does make clear that it isn't code specific to a protocol and that it's related to the `WakuNode` object itself. I think it makes things easier to understand for me, but don't have a strong opinion.
> 
> @Ivansete-status @NagyZoltanPeter thoughts?

Looking at the content of the source file, it's not a true factory for me, in terms of factory pattern.
I would rather call it an initializer or configurator.
Also it implies the knowledge how to use, when to call these procs, rather embed all this dependencies. Idk if this opinion helps...

Ivansete-status:
> > I would avoid naming things node. I suggest `factory` instead of `node_factory` and maybe naming the folder initialization (or init).
> 
> Mmm makes sense, having the word node is kind of redundant. On the other hand, it does make clear that it isn't code specific to a protocol and that it's related to the `WakuNode` object itself. I think it makes things easier to understand for me, but don't have a strong opinion.
> 
> @Ivansete-status @NagyZoltanPeter thoughts?

I personally prefer being explicit with `node_factory` , but I agree it sounds redundant to have `node/node_factory`, so either option is good to me :)

gabrielmer:
> Looking at the content of the source file, it's not a true factory for me, in terms of factory pattern. I would rather call it an initializer or configurator. Also it implies the knowledge how to use, when to call these procs, rather embed all this dependencies. Idk if this opinion helps...

Yes! So the idea is that it will be an actual factory. We're first trying to move the existing code and dependencies there so when we add its main code the diff won't be enormous filled with copy-pasted things. It would be difficult to differentiate what's new code and what's copy-pasted procs being moved around.

So this PR is only so the following PRs are more scoped and easier to review

gabrielmer:
Went over everything today with @Ivansete-status today

The new "middleware" layer is now in the directory `waku/factory/`.
We settled for that naming as it describes more precisely its current purpose for creating/initiating nodes. If at any point it becomes a more generalized layer it can be renamed.

All the imports should now follow the hierarchical logical structure.

Notice that this is only the first step of a bigger refactor task. This PR is intended as a reordering of existing logic and no new code is introduced. The next step is to build on top on this and add functionalities to `node_factory.nim` to make it an actual factory.

NagyZoltanPeter (approved):
I think its a great step ahead with in mind it will evolve. Thank you!

jm-clius (approved):
Thanks!

Ivansete-status (approved):
Great movement! Thanks for it! :100: 
I didn't check the detail of every block as I assume is block movement.
Cheers

PR #2483, opened 2024-02-27
chore: remove rln epoch hardcoding

alrevuelta:
* Remove hardcoded `EpochUnitSeconds` and allows configuring it via `rln-relay-epoch-sec`. 
* It also sets TheWakuNetwork default epoch value in [here](https://github.com/waku-org/nwaku/blob/caad9f88100923b6aed2da5b13b8d7a2e30db290/apps/wakunode2/networks_config.nim#L30) as per [spec](https://rfc.vac.dev/spec/64/#rln-parameters)
* This allows anyone to deploy another network with their own RLN configuration.

### DIFF
--- a/apps/benchmarks/benchmarks.nim
+++ b/apps/benchmarks/benchmarks.nim
@@ -30,7 +30,7 @@ proc main(): Future[string] {.async, gcsafe.} =
   var proofVerTimes: seq[times.Duration] = @[]
   for i in 0 .. 50:
     var time = getTime()
-    let proof = manager.generateProof(data, getCurrentEpoch()).get()
+    let proof = manager.generateProof(data, default(Epoch)).get()
     proofGenTimes.add(getTime() - time)
 
     time = getTime()
--- a/apps/chat2/config_chat2.nim
+++ b/apps/chat2/config_chat2.nim
@@ -272,6 +272,11 @@ type
       defaultValue: 1,
       name: "rln-relay-user-message-limit" .}: uint64
 
+    rlnEpochSizeSec* {.
+      desc: "Epoch size in second used to rate limit RLN memberships. Default is 1 second.",
+      defaultValue: 1
+      name: "rln-relay-epoch-sec" .}: uint64
+
 # NOTE: Keys are different in nim-libp2p
 proc parseCmdArg*(T: type crypto.PrivateKey, p: string): T =
   try:
--- a/apps/chat2/config_chat2.nim
+++ b/apps/chat2/config_chat2.nim
@@ -273,7 +273,7 @@ type
       name: "rln-relay-user-message-limit" .}: uint64
 
     rlnEpochSizeSec* {.
-      desc: "Epoch size in second used to rate limit RLN memberships. Default is 1 second.",
+      desc: "Epoch size in seconds used to rate limit RLN memberships. Default is 1 second.",
       defaultValue: 1
       name: "rln-relay-epoch-sec" .}: uint64
 
--- a/apps/networkmonitor/networkmonitor.nim
+++ b/apps/networkmonitor/networkmonitor.nim
@@ -494,6 +494,8 @@ when isMainModule:
     conf.pubsubTopics = twnClusterConf.pubsubTopics
     conf.rlnRelayDynamic = twnClusterConf.rlnRelayDynamic
     conf.rlnRelayEthContractAddress = twnClusterConf.rlnRelayEthContractAddress
+    conf.rlnEpochSizeSec = twnClusterConf.rlnEpochSizeSec
+    conf.rlnRelayUserMessageLimit: twnClusterConf.rlnRelayUserMessageLimit
 
   if conf.logLevel != LogLevel.NONE:
     setLogLevel(conf.logLevel)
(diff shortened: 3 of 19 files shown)


apps/networkmonitor/networkmonitor.nim
```diff
@@ -546,6 +548,7 @@ when isMainModule:
       rlnRelayCredPath: "",
       rlnRelayCredPassword: "",
       rlnRelayTreePath: conf.rlnRelayTreePath,
+      rlnEpochSizeSec: conf.rlnEpochSizeSec
```
alrevuelta, line 551: @vpavlin if you can have a look just in case.

apps/wakunode2/networks_config.nim
```diff
@@ -25,6 +27,9 @@ proc TheWakuNetworkConf*(T: type ClusterConf): ClusterConf =
     rlnRelayEthContractAddress: "0xF471d71E9b1455bBF4b85d475afb9BB0954A29c4",
     rlnRelayDynamic: true,
     rlnRelayBandwidthThreshold: 0,
+    rlnEpochSizeSec: 1,
+    # parameter to be defined with rln_v2
+    rlnRelayUserMessageLimit: 1,
```
alrevuelta, line 32: this belongs to rln_v2 but if I used the `when defined(rln_v2)` I have to duplicate everything. So adding it directly here.

apps/benchmarks/benchmarks.nim
```diff
@@ -30,7 +30,7 @@ proc main(): Future[string] {.async, gcsafe.} =
   var proofVerTimes: seq[times.Duration] = @[]
   for i in 0 .. 50:
     var time = getTime()
-    let proof = manager.generateProof(data, getCurrentEpoch()).get()
+    let proof = manager.generateProof(data, default(Epoch)).get()
```
rymnc, line 33: this will default to 0, is that okay?

apps/wakunode2/external_config.nim
```diff
@@ -83,12 +83,17 @@ type
...
       defaultValue: "",
       name: "rln-relay-eth-private-key" }: string
-  
+
     rlnRelayUserMessageLimit* {.
       desc: "Set a user message limit for the rln membership registration. Must be a positive integer. Default is 1.",
       defaultValue: 1,
       name: "rln-relay-user-message-limit" .}: uint64
 
+    rlnEpochSizeSec* {.
```
rymnc, line 92: Please add a validator for this field - we don't want operators to shoot themselves in the foot due to memory footprint of too large epoch sizes

rymnc (commented):
Not sure if this is a required feature, are there any applications who want to use a larger epoch size?

apps/benchmarks/benchmarks.nim
alrevuelta: yep all good. its just to benchmark proof generation, so the epoch here is not really important

apps/wakunode2/external_config.nim
alrevuelta: you mean to limit it to a given value? which one?

rymnc: Possibly 1 day? the operators currently have no idea about the memory usage by bumping this value up

alrevuelta: note that an operator is expected to adhere to an existing cluster, eg cluster-id=1, but not define their own parameters. and well, if someone forks waku, they should have enough knowledge to know that 1 day is probably a bad decision.

unsure if we should enforce a limit. i mean, does it make sense to have a max number of peers of 1000000? maybe not, but we don't check for it?

apps/chat2/config_chat2.nim
```diff
@@ -272,6 +272,11 @@ type
       defaultValue: 1,
       name: "rln-relay-user-message-limit" .}: uint64
 
+    rlnEpochSizeSec* {.
+      desc: "Epoch size in second used to rate limit RLN memberships. Default is 1 second.",
```
gabrielmer, line 276: Super mega nitpick

```suggestion
      desc: "Epoch size in seconds used to rate limit RLN memberships. Default is 1 second.",
```

apps/networkmonitor/networkmonitor.nim
```diff
@@ -494,6 +494,8 @@ when isMainModule:
     conf.pubsubTopics = twnClusterConf.pubsubTopics
     conf.rlnRelayDynamic = twnClusterConf.rlnRelayDynamic
     conf.rlnRelayEthContractAddress = twnClusterConf.rlnRelayEthContractAddress
+    conf.rlnEpochSizeSec = twnClusterConf.rlnEpochSizeSec
+    conf.rlnRelayUserMessageLimit: twnClusterConf.rlnRelayUserMessageLimit
```
gabrielmer, line 498: Shouldn't it be like this?

```suggestion
    conf.rlnRelayUserMessageLimit = twnClusterConf.rlnRelayUserMessageLimit
```

gabrielmer (commented):
Looks great! Thanks so much!

Added minor comments, probably one of them is the cause of the failed test

gabrielmer:
Ok so the failed test is because we need to add `rlnEpochSizeSec` and  `rlnRelayUserMessageLimit ` to `networkmonitor_config.nim` I think

apps/chat2/config_chat2.nim
alrevuelta: thanks! fixed 857810d4220dd8651793977ea423bbfa6a909ad9

gabrielmer (approved):

jm-clius (approved):

PR #2485, opened 2024-02-28
fix: return message id on `waku_relay_publish`

richard-ramos:
# Description
As defined in the [bindings](https://rfc.vac.dev/spec/36/#extern-unsinged-int-waku_relay_publishchar-messagejson-char-pubsubtopic-int-timeoutms-char-jsonresp), `waku_relay_publish` returns a message id on success

### DIFF
--- a/library/alloc.nim
+++ b/library/alloc.nim
@@ -19,8 +19,9 @@ proc alloc*(str: string): cstring =
   return ret
 
 proc allocSharedSeq*[T](s: seq[T]): SharedSeq[T] =
-  let data = cast[ptr T](allocShared(s.len))
-  copyMem(data, unsafeAddr s, s.len)
+  let data = allocShared(sizeof(T) * s.len)
+  if s.len != 0:
+    copyMem(data, unsafeAddr s[0], s.len)
   return (cast[ptr UncheckedArray[T]](data), s.len)
 
 proc deallocSharedSeq*[T](s: var SharedSeq[T]) =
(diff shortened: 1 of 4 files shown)


gabrielmer (approved):
LGTM, thanks so much!

Ivansete-status (approved):
Great! Thanks for it!

Ivansete-status:
Hey @richard-ramos ! Sorry for confusion but right after approving the PR I thought that it may be more convenient to use the same mechanism that we have to generate the `msd_id` within GossipSub: https://github.com/waku-org/nwaku/blob/d832f92a439eb541f689073b4c40caf02afbf0ad/waku/waku_relay/message_id.nim#L31-L33

richard-ramos:
> I thought that it may be more convenient to use the same mechanism that we have to generate the msd_id within GossipSub

Actually, it is the messageHash we need! so that makes me think that i have to rename the `messageId` field in https://rfc.vac.dev/spec/36/#jsonmessageevent-type to `messageHash`. WDYT?  I can do that change in this PR

Ivansete-status:
> > I thought that it may be more convenient to use the same mechanism that we have to generate the msd_id within GossipSub
> 
> Actually, it is the messageHash we need! so that makes me think that i have to rename the `messageId` field in https://rfc.vac.dev/spec/36/#jsonmessageevent-type to `messageHash`. WDYT? I can do that change in this PR

Ok cool! I think `messageHash` sounds better but no need to rush for that change

richard-ramos:
Implemented the change in last commit :)

PR #2490, opened 2024-02-29
feat(waku-stealth-commitments): waku stealth commitment protocol

rymnc:
# Description

Adds a `wakustealthcommitment` application/tool/service to the `apps/` directory - it should be used for generating stealth addresses/commitments for users.

This can be useful in a variety of ways, but needs to be fleshed out for curve operations, FFI api used will be from [erc-5564-bn254](https://github.com/rymnc/erc-5564-bn254) or native use of [constantine](https://github.com/mratsim/constantine).

# Changes

- [x] Updates `waku.nimble`, `Makefile`
- [x] Creates new app `wakustealthcommitments`

## How to test

1. Instructions are in [readme](https://github.com/waku-org/nwaku/blob/64746d3b4c3e453f947978db9695a6c61142fb28/apps/wakustealthcommitments/README.md)

https://github.com/waku-org/nwaku/assets/43716372/ee11e546-b0d7-4360-b65d-ab463b6af3c3


apps/wakustealthcommitments/stealth_commitment_protocol.nim
```diff
@@ -0,0 +1,76 @@
...
+
+
+proc new*(wakuApp: App, contentTopic = ContentTopic("/wakustealthcommitments/1/app/proto")): StealthCommitmentProtocol =
+  let SCP = StealthCommitmentProtocol(wakuApp: wakuApp, contentTopic: contentTopic)
+
+  proc handler(topic: PubsubTopic, msg: WakuMessage): Future[void] {.async, gcsafe.} =
+    if msg.contentTopic == contentTopic:
+      SCPHandler(msg)
+
+  wakuApp.node.subscribe((kind: PubsubSub, topic: DefaultPubsubTopic), some(handler))
```
rymnc, line 38: hey @alrevuelta, for some reason while running this application, this handler gets called for messages it is sending. am i missing a config field?

rymnc: noticed we set `triggerSelf` to `true` here -> https://github.com/waku-org/nwaku/blob/731dfcbdf67998c11660b6997d811a55af21c5b2/waku/waku_relay/protocol.nim#L166

alrevuelta: yep

Makefile
```diff
@@ -206,6 +206,10 @@ chat2: | build deps librln
 	echo -e $(BUILD_MSG) "build/$@" && \
 		$(ENV_SCRIPT) nim chat2 $(NIM_PARAMS) waku.nims
 
+wakustealthcommitments: | build deps librln
```
jm-clius, line 209: Perhaps remove Make target (and nimble task)?

apps/wakustealthcommitments/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+> ![NOTE]
+> This static library also includes the rln ffi library, so you don't need to build it separately.
+> This is because using both of them separately brings in a lot of duplicate symbols.
+
+2. Build the wakustealthcommitments app
+```sh
+cd <path-to-nwaku>
+source env.sh
+nim c --out:build/wakustealthcommitments  --verbosity:0 --hints:off -d:chronicles_log_level=INFO -d:git_version="v0.24.0-rc.0-62-g7da25c" -d:release --passL:-lm --passL:liberc_5564_bn254.a --debugger:native apps/wakustealthcommitments/wakustealthcommitments.nim
```
jm-clius, line 25: Could have been `make wakustealthcommitments` here, but I think this should remain like this as I suggest removing the Make target in any case (as it will be temporary until we incorporate stealth commitment protocol into the node itself).

jm-clius (changes requested):
Mmm. I'd suggest moving `wakustealthcommitments` for now to the `examples` subfolder. In some sense, this is still an example/POC to illustrate how nodes can run the stealth commitment protocol. I imagine that at some point the correct way to use this would be to fully integrate the protocol into wakunode2, create a request API, etc. Until then the reference can live as an example and I wouldn't create a Make target/nimble task for it (people can build and run it using `nim c -r` as already described in the README). Wdyt?

apps/wakustealthcommitments/README.md
rymnc: It couldn't have, without setting up a proper pipeline for the erc-5564-bn254 static lib.
removing the makefile entry in any case

Makefile
rymnc: addressed in fea64e05

Ivansete-status (approved):
LGTM! Thanks!

For my understanding, how we should use that protocol to enhance the communication privacy?

rymnc:
> LGTM! Thanks!
> 
> For my understanding, how we should use that protocol to enhance the communication privacy?

image
This should be the way commitments are registered to maintain privacy!

jm-clius (approved):
LGTM! Thanks for this - very cool.

PR #2492, opened 2024-02-29
fix: time on 32 bits architecture

SionoiS:
Copy of https://github.com/waku-org/nwaku/pull/2476 for merging purpose.

> Description
> The function [getNanosecondTime](https://github.com/waku-org/nwaku/blob/master/waku/waku_core/time.nim#L12) does not > work on 32 bit architecture, so this is a suggestion of a fix.
> 
> Changes
> Use int64.
> 
> To reproduce the problem
> Pass --cpu:i386 -t:-m32 -l:-m32 to the nim compiler
> 
> Issue
> closes # https://github.com/waku-org/nwaku/issues/2475

### DIFF
--- a/tests/waku_core/test_time.nim
+++ b/tests/waku_core/test_time.nim
@@ -21,14 +21,10 @@ suite "Waku Core - Time":
 
     ## When
     let
-      timeInSecondsInt = secondsPart.int
       timeInSecondsInt64 = secondsPart.int64
-      timeInSecondsFloat = float(secondsFloat)
       timeInSecondsFloat64 = float64(secondsFloat)
 
     ## Then
     check:
-      getNanosecondTime(timeInSecondsInt) == lowResTimestamp
       getNanosecondTime(timeInSecondsInt64) == lowResTimestamp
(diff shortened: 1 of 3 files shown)


jm-clius (approved):

tests/waku_core/test_time.nim
```diff
@@ -21,14 +21,10 @@ suite "Waku Core - Time":
...
     ## When
     let
-      timeInSecondsInt = secondsPart.int
       timeInSecondsInt64 = secondsPart.int64
-      timeInSecondsFloat = float(secondsFloat)
       timeInSecondsFloat64 = float64(secondsFloat)
 
     ## Then
     check:
-      getNanosecondTime(timeInSecondsInt) == lowResTimestamp
```
Ivansete-status, line 31: Why this check is not needed?

Ivansete-status (approved):
LGTM! Just a naïve question:)

tests/waku_core/test_time.nim
EmilIvanichkovv: Removed these tests, as they are in contradiction with the change ( allow using the function on 32bit )

PR #2499, opened 2024-03-04
feat(c-bindings): add function to dealloc nodes

richard-ramos:
# Description
While working with the rust bindings I noticed that when I closed the example program, it segfaulted, and this happened due to the waku thread still being alive. This PR adds a `waku_destroy` function that stops the thread, as well as firing a signal so the code does not stay stuck at https://github.com/waku-org/nwaku/blob/9ef2eccb8aa19e80a9381dcb2ece29718ebce30f/library/waku_thread/waku_thread.nim#L59

### DIFF
--- a/library/libwaku.nim
+++ b/library/libwaku.nim
@@ -100,6 +100,21 @@ proc waku_new(configJson: cstring,
 
   return ctx
 
+proc waku_destroy(ctx: ptr Context,
+               callback: WakuCallBack,
+               userData: pointer): cint {.dynlib, exportc.} =
+
+  if isNil(callback):
+    return RET_MISSING_CALLBACK
+
+  let stopNodeRes = waku_thread.stopWakuNodeThread(ctx)
+  if stopNodeRes.isErr():
+    let msg = $stopNodeRes.error
+    callback(RET_ERR, unsafeAddr msg[0], cast[csize_t](len(msg)), userData)
+    return RET_ERR
+
+  return RET_OK
+
 proc waku_version(ctx: ptr Context,
                   callback: WakuCallBack,
                   userData: pointer): cint {.dynlib, exportc.} =
--- a/library/libwaku.nim
+++ b/library/libwaku.nim
@@ -107,9 +107,8 @@ proc waku_destroy(ctx: ptr Context,
   if isNil(callback):
     return RET_MISSING_CALLBACK
 
-  let stopNodeRes = waku_thread.stopWakuNodeThread(ctx)
-  if stopNodeRes.isErr():
-    let msg = $stopNodeRes.error
+  waku_thread.stopWakuThread(ctx).isOkOr:
+    let msg = $error
     callback(RET_ERR, unsafeAddr msg[0], cast[csize_t](len(msg)), userData)
     return RET_ERR
 
--- a/library/waku_thread/waku_thread.nim
+++ b/library/waku_thread/waku_thread.nim
@@ -95,12 +95,16 @@ proc createWakuThread*(): Result[ptr Context, string] =
 
   return ok(ctx)
 
-proc stopWakuNodeThread*(ctx: ptr Context) =
+proc stopWakuNodeThread*(ctx: ptr Context): Result[void, string] =
   running.store(false)
+  let fireRes = ctx.reqSignal.fireSync()
+  if fireRes.isErr():
+    return err(fireRes.error)
   joinThread(ctx.thread)
   discard ctx.reqSignal.close()
   discard ctx.respSignal.close()
   freeShared(ctx)
+  ok()
 
 proc sendRequestToWakuThread*(ctx: ptr Context,
                               reqType: RequestType,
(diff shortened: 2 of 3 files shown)


library/waku_thread/waku_thread.nim
```diff
@@ -95,12 +95,16 @@ proc createWakuThread*(): Result[ptr Context, string] =
 
   return ok(ctx)
 
-proc stopWakuNodeThread*(ctx: ptr Context) =
+proc stopWakuNodeThread*(ctx: ptr Context): Result[void, string] =
   running.store(false)
+  let fireRes = ctx.reqSignal.fireSync()
```
gabrielmer, line 100: Please let me know if I understood correctly.

If we fire the signal, the thread moves from https://github.com/waku-org/nwaku/blob/9ef2eccb8aa19e80a9381dcb2ece29718ebce30f/library/waku_thread/waku_thread.nim#L59

And it will exit the loop in https://github.com/waku-org/nwaku/blob/9ef2eccb8aa19e80a9381dcb2ece29718ebce30f/library/waku_thread/waku_thread.nim#L55

Because of `running.load` not being `true` anymore?

gabrielmer (approved):
Thanks so much!

library/waku_thread/waku_thread.nim
richard-ramos: Yes, it should move from `waitFor ctx.reqSignal.wait()` and stop in the next line:
```
    let recvOk = ctx.reqChannel.tryRecv(request)
```
`recvOk` is false at this point, so the iteration will end and next loop will not be executed due to `running.load` being false.

I'm actually not sure if `tryRecv` will always be false. Perhaps I should change the code so it's like this? let me know what you think:
```nim
  while running.load == true:
    ## Trying to get a request from the libwaku main thread

    var request: ptr InterThreadRequest
    waitFor ctx.reqSignal.wait()
    if not running.load: break       <------------------------------
    let recvOk = ctx.reqChannel.tryRecv(request)
    if recvOk == true:

```

gabrielmer: Thanks for the explanation!

I personally like that alternative better :) Find it more predictable and easier to understand

NagyZoltanPeter (approved):
LGTM!
Was just thinking of if we can automate this destroy thread process...

library/waku_thread/waku_thread.nim
```diff
@@ -95,12 +95,16 @@ proc createWakuThread*(): Result[ptr Context, string] =
 
   return ok(ctx)
 
-proc stopWakuNodeThread*(ctx: ptr Context) =
+proc stopWakuNodeThread*(ctx: ptr Context): Result[void, string] =
```
Ivansete-status, line 98: Out of the scope of this PR but I just realised it worth keeping consistency with the "start" proc:
```suggestion
proc stopWakuThread*(ctx: ptr Context): Result[void, string] =
```

library/waku_thread/waku_thread.nim
```diff
@@ -95,12 +95,16 @@ proc createWakuThread*(): Result[ptr Context, string] =
...
   running.store(false)
+  let fireRes = ctx.reqSignal.fireSync()
+  if fireRes.isErr():
+    return err(fireRes.error)
   joinThread(ctx.thread)
   discard ctx.reqSignal.close()
   discard ctx.respSignal.close()
   freeShared(ctx)
+  ok()
```
Ivansete-status, line 107: nitpic: even though nim supports that I'd rather be more explicit, specially for new Nim devs
```suggestion
  return ok()
```

library/waku_thread/waku_thread.nim
```diff
@@ -95,12 +95,16 @@ proc createWakuThread*(): Result[ptr Context, string] =
 
   return ok(ctx)
 
-proc stopWakuNodeThread*(ctx: ptr Context) =
+proc stopWakuNodeThread*(ctx: ptr Context): Result[void, string] =
   running.store(false)
+  let fireRes = ctx.reqSignal.fireSync()
+  if fireRes.isErr():
+    return err(fireRes.error)
```
Ivansete-status, line 102: To give a little bit more context in the error
```suggestion
    return err("error in stopWakuThread: " & $fireRes.error)
```

library/waku_thread/waku_thread.nim
```diff
@@ -58,7 +58,7 @@ proc run(ctx: ptr Context) {.thread.} =
     var request: ptr InterThreadRequest
     waitFor ctx.reqSignal.wait()
     let recvOk = ctx.reqChannel.tryRecv(request)
-    if recvOk == true:
+    if running.load and recvOk == true:
```
Ivansete-status, line 61: I have the impression that this is not needed but I didn't debug it.
```suggestion
    if recvOk == true:
```

Ivansete-status (approved):
LGTM! Thanks! :100: 
Just added some minor comments :)

library/libwaku.nim
```diff
@@ -100,6 +100,21 @@ proc waku_new(configJson: cstring,
 
   return ctx
 
+proc waku_destroy(ctx: ptr Context,
```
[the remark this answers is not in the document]
richard-ramos: Yes. Bindings in other languages can automate the call to this function to be executed on some destructor function if available. 
Do you have an example of the exit handler idea? I'm not sure how to implement this, so look at something i could use as a guide.

PR #2500, opened 2024-03-05
fix(postgres): import under feature flag

rymnc:
# Description

Previously, the postgres migration was imported outside of the feature flag and hence caused a dependency on the libpq library

# Changes

- [x] moved import to under the feature flag


Ivansete-status (approved):

NagyZoltanPeter (approved):
Great!!!

PR #2504, opened 2024-03-05
chore: migrating logic from wakunode2.nim to node_factory.nim

gabrielmer:
# Description

# Changes

- [ ] ...
- [ ] ...


SionoiS:
LGTM except moving the peer storage inside of node.

Just to be sure there's no confusion. Peer storage is the DB storing the peer information. Peer store is part to the peer management system and store peer info in memory.

Peer storage should not be part of node IMO.

gabrielmer:
> LGTM except moving the peer storage inside of node.
> 
> Just to be sure there's no confusion. Peer storage is the DB storing the peer information. Peer store is part to the peer management system and store peer info in memory.
> 
> Peer storage should not be part of node IMO.

Makes sense! Notice that this is not part of the node itself and is in the factory/middleware layer. I deleted it from the `App` object because it was only being passed down when creating the node and not being used anywhere else.

Is it better to create it when initializing the `App` and pass it down as a parameter to `setupNode()`? 
https://github.com/waku-org/nwaku/blob/d4370b66f7a45566ed0349e0c462a1d7c5fa2c31/waku/factory/node_factory.nim#L377-L414

This is one of the cases that I'm not sure if we want it in the app layer or in the factory/middleware one.

SionoiS:
> Is it better to create it when initializing the `App` and pass it down as a parameter to `setupNode()`?

:thinking: Yes nodes without peer storage can exist so I would said `Option[PeerStorage]` or similar for setup.

jm-clius (approved):
Generally happy with this approach. Feel free to open for review when ready to merge.

waku/node/peer_manager/peer_manager.nim
```diff
@@ -75,7 +75,7 @@ type
     initialBackoffInSec*: int
     backoffFactor*: int
     maxFailedAttempts*: int
-    storage: PeerStorage
+    storage*: PeerStorage
```
gabrielmer, line 78: For testing in `app.nim`. Wanted to import the file instead but there's already one import for the same purpose and they have conflicts.

gabrielmer:
I want to add a test suite for `node_factory.nim` but I think it should probably be outside of the scope of this PR, as the only proc we actually added is `setupNode` (which is mostly a refactor) and it's included in the tests for app initialization.

lmk what you think

apps/wakunode2/app.nim
```diff
@@ -88,74 +85,43 @@ func node*(app: App): WakuNode =
...
+proc init*(T: type App, conf: WakuNodeConf): Result[App, string] =
+
+  var confCopy = conf
+  let rng = crypto.newRng()
+
+  if not confCopy.nodekey.isSome():
+    let keyRes = crypto.PrivateKey.random(Secp256k1, rng[])
+    if keyRes.isErr():  
+      error "Failed to generate key", error = $keyRes.error
+      return err("Failed to generate key: " & $keyRes.error)
```
Ivansete-status, line 99: I think this should keep stopping the whole node
edited: ok I see this is happening in `wakunode2.nim` :)

Ivansete-status (approved):
LGTM! Thanks for that master piece PR! :100:

waku/factory/node_factory.nim
```diff
@@ -372,4 +372,56 @@ proc startNode*(node: WakuNode, conf: WakuNodeConf,
...
+      if conf.nodeKey.isSome() and rng.isSome():
+        conf.nodeKey.get()
+      else:
+          warn "missing key or rng, generating new set"
+          crypto.PrivateKey.random(Secp256k1, nodeRng[]).valueOr:
+            error "Failed to generate key", error=error
+            return err("Failed to generate key: " & $error)
+      
+    let netConfig = networkConfiguration(conf, clientId).valueOr:
+      error "failed to create internal config", error=error
```
SionoiS, line 392: Would it be useful to create an Error enum for all the possible errors?

For e.g. https://github.com/waku-org/nwaku/blob/1274b15d75d4150889623a0b52dfcb7afd1d48ae/waku/waku_archive/common.nim#L69-L75

apps/wakunode2/app.nim
```diff
@@ -88,74 +85,43 @@ func node*(app: App): WakuNode =
...
+    confCopy.nodekey = some(keyRes.get())
+
+  debug "Retrieve dynamic bootstrap nodes"
+  let dynamicBootstrapNodesRes = retrieveDynamicBootstrapNodes(confCopy.dnsDiscovery,
+                                                              confCopy.dnsDiscoveryUrl,
+                                                              confCopy.dnsDiscoveryNameServers)
+  if dynamicBootstrapNodesRes.isErr():
```
SionoiS, line 106: same as above

apps/wakunode2/app.nim
```diff
@@ -88,74 +85,43 @@ func node*(app: App): WakuNode =
...
+    error "Retrieving dynamic bootstrap nodes failed", error = dynamicBootstrapNodesRes.error
+    return err("Retrieving dynamic bootstrap nodes failed: " & dynamicBootstrapNodesRes.error)
+
+  let nodeRes = setupNode(confCopy, some(rng))
+  if nodeRes.isErr():    
+    error "Failed setting up node", error=nodeRes.error
+    return err("Failed setting up node: " & nodeRes.error)
```
SionoiS, line 113: same here

apps/wakunode2/app.nim
SionoiS, line 99: Maybe use `.valueOr()`

SionoiS (approved):
LGTM thanks!

apps/wakunode2/app.nim
gabrielmer: If I remember correctly I initially tried `valueOr()` but for some reason had compilation errors related to logging that I didn't have anywhere else - so just used the other approach

I'll try again to double check :)

waku/factory/node_factory.nim
SionoiS: If you impl. `$` on the error enum then you could do
```
let myvar = doSomething().valueOr:
   error "this failed bla bla", error = $error
   retrun err($error)
```

IDK maybe not worth it.

apps/wakunode2/app.nim
gabrielmer: Yap, it was that.

For example if I change it to

```
let dynamicBootstrapNodes = retrieveDynamicBootstrapNodes(confCopy.dnsDiscovery,
                                                              confCopy.dnsDiscoveryUrl,
                                                              confCopy.dnsDiscoveryNameServers).valueOr:
    error "Retrieving dynamic bootstrap nodes failed", error = $error
    return err("Retrieving dynamic bootstrap nodes failed: " & $error)
```
    
It has issues printing the variable `error` because it thinks it's of type `None`
    
Not sure why, as the return value of `retrieveDynamicBootstrapNodes()` is `Result[seq[RemotePeerInfo], string]`

I think that after some attempts trying to figure it out I just wrote it the other way and moved forward.

Any ideas?

waku/factory/node_factory.nim
gabrielmer: Yeah I don't think it's worth it for this specific case - It's nicer but I prefer keeping it as simple as possible :)

apps/wakunode2/app.nim
SionoiS: Yeah I too noticed some edge cases (I hate those). If only the compiler was smarter... *cough itwouldworkinrust cough* oh well :shrug:

gabrielmer: loool (but true) 😅 😵

PR #2505, opened 2024-03-05
fix: Enable autosharding in any cluster

alrevuelta:
# Description
* Removes hardcoding of amount of shards.
* Removes hardcoding of cluster-id with enabled autosharding.
* This enables any `cluster-id` != 1 to i) have autosharding if desired and ii) have whatever amount of shards they want.

For example, this PR enables the following. Creating a `cluster-id=10` with `4` shards, where the node only subscribed to `0, 1`

```
./build/wakunode2\
  --rest=true\
  --log-level=DEBUG\
  --relay=true\
  --pubsub-topic=/waku/2/rs/10/0\
  --pubsub-topic=/waku/2/rs/10/1\
  --pubsub-topic=/waku/2/rs/10/2\
  --pubsub-topic=/waku/2/rs/10/3\
  --cluster-id=10\
  --shard=0\
  --shard=1
```

### DIFF
--- a/tests/waku_relay/test_wakunode_relay.nim
+++ b/tests/waku_relay/test_wakunode_relay.nim
@@ -507,7 +507,6 @@ suite "WakuNode - Relay":
     await node.start()
     await node.mountRelay()
     require node.mountSharding(1, 1).isOk
-    echo node.wakuSharding.getShard(DefaultContentTopic)
 
     ## Given
     let
--- a/tests/waku_relay/test_wakunode_relay.nim
+++ b/tests/waku_relay/test_wakunode_relay.nim
@@ -506,10 +506,12 @@ suite "WakuNode - Relay":
 
     await node.start()
     await node.mountRelay()
+    require node.mountSharding(1, 1).isOk
+    echo node.wakuSharding.getShard(DefaultContentTopic)
 
     ## Given
     let
-      shard = "/waku/2/rs/1/1"
+      shard = "/waku/2/rs/1/0"
       contentTopicA = DefaultContentTopic
       contentTopicB = ContentTopic("/waku/2/default-content1/proto")
       contentTopicC = ContentTopic("/waku/2/default-content2/proto")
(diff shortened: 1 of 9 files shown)


waku/factory/internal_config.nim
```diff
@@ -31,24 +31,12 @@ proc enrConfiguration*(conf: WakuNodeConf, netConfig: NetConfig, key: crypto.Pri
...
-
-  let addShardedTopics = enrBuilder.withShardedTopics(topics)
-  if addShardedTopics.isErr():
-      error "failed to add sharded topics to ENR", error=addShardedTopics.error
-      return err($addShardedTopics.error)
+  let shards: seq[uint16] =
+    if conf.shards.len == 0: toSeq(0..8).mapIt(uint16(it))
+    else: toSeq(conf.shards.mapIt(uint16(it)))
+
+  enrBuilder.withWakuRelaySharding(RelayShards(clusterId:uint16(conf.clusterId), shardIds: shards)).isOkOr():
```
alrevuelta, line 38: imho the ENR doesnt require knowledge of topics. just `cluster-id` and `shards`

waku/node/waku_node.nim
```diff
@@ -199,6 +200,11 @@ proc mountMetadata*(node: WakuNode, clusterId: uint32): Result[void, string] =
 
   return ok()
 
+## Waku Sharding
+proc mountSharding*(node: WakuNode, clusterId: uint32, shardCount: uint32): Result[void, string] =
```
alrevuelta, line 204: not sure i like this approach

SionoiS: We had a stateless system before. I not sure we can do better than that.

waku/factory/internal_config.nim
SionoiS: The problem is that pubsub topics in the config can be static shards. Maybe we could disallow that somehow?

alrevuelta: mm i see. in that case we need a flag that is static/auto?

waku/node/waku_node.nim
alrevuelta: sometimes i see sharding as part of relay. but a light client without relay, would need it also. so guess it should be a "node" thing.

fbarbu15 (approved):
LGTM, thanks!

tests/waku_relay/test_wakunode_relay.nim
```diff
@@ -506,10 +506,12 @@ suite "WakuNode - Relay":
 
     await node.start()
     await node.mountRelay()
+    require node.mountSharding(1, 1).isOk
+    echo node.wakuSharding.getShard(DefaultContentTopic)
```
SionoiS, line 510: left overs?

waku/waku_enr/sharding.nim
```diff
@@ -27,8 +27,8 @@ const
 
 type
   RelayShards* = object
-    clusterId: uint16
-    shardIds: seq[uint16]
+    clusterId*: uint16
+    shardIds*: seq[uint16]
 
 func clusterId*(rs: RelayShards): uint16 =
```
SionoiS, line 33: remove the `clusterId` and `shardIds` if they are no longer needed.

SionoiS (approved):
LGTM

Thank you!

tests/waku_relay/test_wakunode_relay.nim
alrevuelta: wops, fixed: c6a446b4502f6ec6a727bae99e3d8d87b7190844

waku/waku_enr/sharding.nim
alrevuelta: sure: c6a446b4502f6ec6a727bae99e3d8d87b7190844

PR #2506, opened 2024-03-05
feat: Postgres partition implementation

Ivansete-status:
## Description
First of all, special thanks to @NagyZoltanPeter for suggesting such a great partition proposal :partying_face: !

In this PR we start managing the partition creation and deletion.
This is interesting from the database size maintenance pov.

There will always be a current partition.
The `loopPartitionFactory` proc responsible for making sure that there will always be a future partition available.

The `loopPartitionFactory` will check every `DefaultDatabasePartitionCheckTimeInterval` (10 minutes hard-coded) whether a new partition should be created or not.

The `PartitionsRangeInterval` is also hard-coded (1 hour.) This value establishes the time window to store messaged based in their `storedAt` (epoch time in nanoseconds.)

Notice that the partitions are *only* removed when `size` retention policy is being applied.

image

## Changes

- [x] Avoid using `require` on postgres tests. I've noticed that it doesn't handle the failure cases very well.
- [x] Bump the Postgres schema version to 2 (`waku/waku_archive/driver/postgres_driver/migrations.nim`.)
- [x] New `waku/waku_archive/driver/postgres_driver/partitions_manager.nim` that just keeps track the partitions. The partitions are being created from within the `postgres_driver.nim` (`loopPartitionFactory`.) and removed from the size retention policy.
- [x] Cleanup in `postgres_driver.nim`. Removes `createMessageTable`, `init`, and `deleteMessageTable` procs.
- [x] Notice that the `forceRemoval: bool = false` param in `decreaseDatabaseSize` is for testing purposes, to allow forcing all the partitions.

### DIFF
--- a/waku/waku_archive/driver/builder.nim
+++ b/waku/waku_archive/driver/builder.nim
@@ -108,7 +108,8 @@ proc new*(T: type ArchiveDriver,
           return err("ArchiveDriver build failed in migration: " & $migrateRes.error)
 
       ## This should be started once we make sure the 'messages' table exists
-      asyncSpawn driver.loopPartitionFactory(onFatalErrorAction)
+      ## Hence, this should be run after the migration is completed.
+      asyncSpawn driver.startPartitionFactory(onFatalErrorAction)
 
       info "waiting for a partition to be created"
       for i in 0..<100:
@@ -117,7 +118,7 @@ proc new*(T: type ArchiveDriver,
         await sleepAsync(chronos.milliseconds(100))
 
       if not driver.containsAnyPartition():
-        return err("a partition could not be created")
+        onFatalErrorAction("a partition could not be created")
 
       return ok(driver)
 
--- a/waku/waku_archive/driver/builder.nim
+++ b/waku/waku_archive/driver/builder.nim
@@ -110,6 +110,15 @@ proc new*(T: type ArchiveDriver,
       ## This should be started once we make sure the 'messages' table exists
       asyncSpawn driver.loopPartitionFactory(onFatalErrorAction)
 
+      info "waiting for a partition to be created"
+      for i in 0..<100:
+        if driver.containsAnyPartition():
+          break
+        await sleepAsync(chronos.milliseconds(100))
+
+      if not driver.containsAnyPartition():
+        return err("a partition could not be created")
+
       return ok(driver)
 
     else:
--- a/waku/waku_archive/driver/builder.nim
+++ b/waku/waku_archive/driver/builder.nim
@@ -107,6 +107,9 @@ proc new*(T: type ArchiveDriver,
         if migrateRes.isErr():
           return err("ArchiveDriver build failed in migration: " & $migrateRes.error)
 
+      ## This should be started once we make sure the 'messages' table exists
+      asyncSpawn driver.loopPartitionFactory(onFatalErrorAction)
+
       return ok(driver)
 
     else:
--- a/waku/waku_archive/driver/postgres_driver/partitions_manager.nim
+++ b/waku/waku_archive/driver/postgres_driver/partitions_manager.nim
@@ -81,17 +81,18 @@ proc removeOldestPartitionName*(self: PartitionManager) =
 proc isEmpty*(self: PartitionManager): bool =
   return self.partitions.len == 0
 
-proc getLastMoment*(partition: Partition): Moment =
+proc getLastMoment*(partition: Partition): int64 =
   ## Considering the time range covered by the partition, this
-  ## returns the `end` 'Moment' of such range.
+  ## returns the `end` time (number of seconds since epoch) of such range.
   let lastTimeInSec = partition.timeRange.`end`
-  return Moment.init(lastTimeInSec, Second)
+  return lastTimeInSec
 
-proc containsMoment*(partition: Partition, moment: Moment): bool =
+proc containsMoment*(partition: Partition, time: int64): bool =
   ## Returns true if the given moment is contained within the partition window,
   ## 'false' otherwise.
-  if partition.timeRange.beginning <= moment.epochSeconds() and
-     moment.epochSeconds() < partition.timeRange.`end`:
+  ## time - number of seconds since epoch
+  if partition.timeRange.beginning <= time and
+     time < partition.timeRange.`end`:
     return true
 
   return false
--- a/waku/waku_archive/driver/postgres_driver/partitions_manager.nim
+++ b/waku/waku_archive/driver/postgres_driver/partitions_manager.nim
@@ -1,0 +1,108 @@
+
+## This module is aimed to handle the creation and truncation of partition tables
+## in order to limit the space occupied in disk by the database.
+##
+## The created partitions are referenced by the 'storedAt' field.
+##
+
+import
+  std/[times, deques],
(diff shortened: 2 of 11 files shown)


waku/waku_archive/driver/postgres_driver/postgres_driver.nim
```diff
@@ -587,6 +534,206 @@ proc sleep*(s: PostgresDriver, seconds: int):
...
+  ## In the following lines is where the partition removal happens.
+  ## Detach and remove the partition concurrently to not block the parent table (messages)
+  let detachPartitionQuery =
+    "ALTER TABLE messages DETACH PARTITION " & oldestPartition.getName() & " CONCURRENTLY;"
+  debug "removeOldestPartition", query = detachPartitionQuery
+  (await self.performWriteQuery(detachPartitionQuery)).isOkOr:
+    return err(fmt"error in {detachPartitionQuery}: " & $error)
+
+  ## Clear all data from the partition and retrieve back space to the operating system
+  let truncateQuery = "TRUNCATE TABLE " & oldestPartition.getName()
```
NagyZoltanPeter, line 687: Why do we need to truncate date from the table before dropping it? Seems double work for me.

waku/waku_archive/driver/postgres_driver/postgres_driver.nim
```diff
@@ -587,6 +534,206 @@ proc sleep*(s: PostgresDriver, seconds: int):
...
+  ## database size in bytes
+  var totalSizeOfDB: int64 = int64(dbSize)
+
+  debug "reduce database size", targetSize = $targetSizeInBytes, currentSize = $totalSizeOfDB
+
+  if totalSizeOfDB <= targetSizeInBytes:
+    return ok()
+
+  while totalSizeOfDB > targetSizeInBytes and driver.containsAnyPartition():
+    (await driver.removeOldestPartition(forceRemoval)).isOkOr:
```
NagyZoltanPeter, line 721: In general it is an ok algorithm I think it might be to consider that at least collect information about peak time-ranges. Probably it will heavily depend on different use cases (what type of applications will use the store service) what will be the message count/size distribution among time periods we defined.
Would we consider to log somehow the size of the partitions (with time-period it contained)?

waku/waku_archive/driver/builder.nim
```diff
@@ -107,6 +107,19 @@ proc new*(T: type ArchiveDriver,
         if migrateRes.isErr():
           return err("ArchiveDriver build failed in migration: " & $migrateRes.error)
 
+      ## This should be started once we make sure the 'messages' table exists
+      ## Hence, this should be run after the migration is completed.
+      asyncSpawn driver.startPartitionFactory(onFatalErrorAction)
```
NagyZoltanPeter, line 112: Is this just an async proc start isn't? 
How do you see the regular partition management blocking overhead on general message processing?
Would it be possible to separate this management into a complete new thread instead? I mean to consider this option.

NagyZoltanPeter (approved):
Overall, I'm amazed how wonderful this PR is!
As we discussed it many times I was excited to see already. This is a real win!
Thank you for it.
Approved! Only left some questions and stuff to consider for the future.
Let's see it live working!

SionoiS (approved):
Amazing. Thanks!

Hope it works as well as it sounds.

waku/waku_archive/driver/postgres_driver/postgres_driver.nim
```diff
@@ -587,6 +534,206 @@ proc sleep*(s: PostgresDriver, seconds: int):
...
+        onFatalError("could not get newest partition: " & $newestPartitionRes.error)
+
+      let newestPartition = newestPartitionRes.get()
+      if newestPartition.containsMoment(now):
+        debug "creating a new partition for the future"
+        ## The current used partition is the last one that was created.
+        ## Thus, let's create another partition for the future.
+        (await self.addPartition(newestPartition.getLastMoment(), PartitionsRangeInterval)).isOkOr:
+          onFatalError("could not add the next partition for 'now': " & $error)
+      elif now >= newestPartition.getLastMoment():
```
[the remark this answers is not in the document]
gabrielmer: Really like this idea of creating always one partition ahead of time.

In that case, there might be partitions without messages at all (if we create it in advance and don't receive anything in the defined period).

If that happens we should probably remove it?

gabrielmer (approved):
Amazing PR! Thanks so much! 😍

jm-clius (approved):
Great work, @Ivansete-status (and @NagyZoltanPeter). :)

waku/waku_archive/driver/postgres_driver/postgres_driver.nim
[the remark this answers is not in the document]
Ivansete-status: > Do we expect this scenario regular? I mean is it possible that we would like to store message while we do not have a right time-range partition prepared for it yet?

Yes, this could happen for example if a node has been stopped for a while and gets connected again to the database (considering only one node connected to the database.)

----

> I feel more safe that rather create partition always one period ahead of become actual. WDYT?

Sorry, I can't quite understand this comment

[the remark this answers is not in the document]
Ivansete-status: > If that happens we should probably remove it?

In this case, the retention policy will eat the older partitions until it gets up to the desired size. If one partition is empty, then it will need to truncate more partitions until reaching the target size.

waku/waku_archive/driver/postgres_driver/postgres_driver.nim
Ivansete-status: > Would we consider to log somehow the size of the partitions (with time-period it contained)?

Good point! Tackled in https://github.com/waku-org/nwaku/commit/dce24e57858d370965cad4766f533bf8cf46397b

e.g.:
```
DBG 2024-03-06 20:12:26.948+01:00 removed partition                          tid=1981932 file=postgres_driver.nim:721 partition_name=messages_1709752346_1709755946 partition_size="32 kB"

```

waku/waku_archive/driver/builder.nim
Ivansete-status: > Is this just an async proc start isn't? How do you see the regular partition management blocking overhead on general message processing? Would it be possible to separate this management into a complete new thread instead? I mean to consider this option.

This doesn't represent a big issue because the partitons info is stored in memory and the interaction with database to create a new partition will happen once per hour with that configuration.

waku/waku_archive/driver/postgres_driver/partitions_manager.nim
```diff
@@ -0,0 +1,105 @@
...
+  for partition in self.partitions:
+    let timeRange = partition.timeRange
+
+    let beginning = timeRange.beginning
+    let `end` = timeRange.`end`
+
+    if beginning <= targetMoment and targetMoment < `end`:
+      return ok(partition)
+
+  return err("Could'nt find a partition table for given time: " & $targetMoment)
```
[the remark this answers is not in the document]
Ivansete-status: > Super nitpick 😶

Thanks! Covered in https://github.com/waku-org/nwaku/commit/62e7b3e58e29fbf7cb6f82e7d7dae26b18e2d477

PR #2508, opened 2024-03-05
fix(rln-relay): make nullifier log abide by epoch ordering

rymnc:
# Description

~Instead of using the `epoch` as the key for the nullifierLog, `externalNullifier` was being used. This could lead to a node only accepting one valid message sent during an epoch `e`, and reject all other messages sent during epoch `e`.~
Made the NullifierLog properly map from Epoch to Nullifiers instead of ExternalNullifier to ProofMetadata.
Optimized by nesting the hashmap to locate duplicates faster instead of searching an array.

# Changes

- [x] nullifierLog is now `OrderedTable[Epoch, Table[Nullifier, ProofMetadata]]`
- [x] updated tests

### DIFF
--- a/tests/waku_rln_relay/test_waku_rln_relay.nim
+++ b/tests/waku_rln_relay/test_waku_rln_relay.nim
@@ -705,6 +705,55 @@ suite "Waku rln relay":
       msgValidate3 == MessageValidationResult.Valid
       msgValidate4 == MessageValidationResult.Invalid
 
+  asyncTest "validateMessageAndUpdateLog: multiple senders with same external nullifier":
+    let index1 = MembershipIndex(5)
+    let index2 = MembershipIndex(6)
+
+    let rlnConf1 = WakuRlnConfig(rlnRelayDynamic: false,
+                                 rlnRelayCredIndex: some(index1),
+                                 rlnEpochSizeSec: 1,
+                                 rlnRelayTreePath: genTempPath("rln_tree", "waku_rln_relay_3"))
+    let wakuRlnRelay1Res = await WakuRlnRelay.new(rlnConf1)
+    if wakuRlnRelay1Res.isErr(): assert false, "failed to create waku rln relay: " & $wakuRlnRelay1Res.error
+    let wakuRlnRelay1 = wakuRlnRelay1Res.get()
+
+    let rlnConf2 = WakuRlnConfig(rlnRelayDynamic: false,
+                                 rlnRelayCredIndex: some(index2),
+                                 rlnEpochSizeSec: 1,
+                                 rlnRelayTreePath: genTempPath("rln_tree", "waku_rln_relay_4"))
+    let wakuRlnRelay2Res = await WakuRlnRelay.new(rlnConf2)
+    if wakuRlnRelay2Res.isErr():
+      assert false, "failed to create waku rln relay: " & $wakuRlnRelay2Res.error
+    let wakuRlnRelay2 = wakuRlnRelay2Res.get()
+    # get the current epoch time
+    let time = epochTime()
+
+    #  create messages from different peers and append rln proofs to them
+    var
+      wm1 = WakuMessage(payload: "Valid message from sender 1".toBytes())
+      # another message in the same epoch as wm1, it will break the messaging rate limit
+      wm2 = WakuMessage(payload: "Valid message from sender 2".toBytes())
+
+
+    let
+      proofAdded1 = wakuRlnRelay1.appendRLNProof(wm1, time)
+      proofAdded2 = wakuRlnRelay2.appendRLNProof(wm2, time)
+
+    # ensure proofs are added
+    assert proofAdded1.isOk(), "failed to append rln proof: " & $proofAdded1.error
+    assert proofAdded2.isOk(), "failed to append rln proof: " & $proofAdded2.error
+
+    # validate messages
+    # validateMessage proc checks the validity of the message fields and adds it to the log (if valid)
+    let
+      msgValidate1 = wakuRlnRelay1.validateMessageAndUpdateLog(wm1, some(time))
+      # since this message is from a different sender, it should be validated successfully
+      msgValidate2 = wakuRlnRelay1.validateMessageAndUpdateLog(wm2, some(time))
+
+    check:
+      msgValidate1 == MessageValidationResult.Valid
+      msgValidate2 == MessageValidationResult.Valid
+
   test "toIDCommitment and toUInt256":
     # create an instance of rln
     let rlnInstance = createRLNInstanceWrapper()
--- a/tests/waku_rln_relay/test_waku_rln_relay.nim
+++ b/tests/waku_rln_relay/test_waku_rln_relay.nim
@@ -713,18 +713,15 @@ suite "Waku rln relay":
                                  rlnRelayCredIndex: some(index1),
                                  rlnEpochSizeSec: 1,
                                  rlnRelayTreePath: genTempPath("rln_tree", "waku_rln_relay_3"))
-    let wakuRlnRelay1Res = await WakuRlnRelay.new(rlnConf1)
-    if wakuRlnRelay1Res.isErr(): assert false, "failed to create waku rln relay: " & $wakuRlnRelay1Res.error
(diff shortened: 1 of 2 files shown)


waku/waku_rln_relay/rln_relay.nim
```diff
@@ -103,58 +103,54 @@ proc stop*(rlnPeer: WakuRLNRelay) {.async: (raises: [Exception]).} =
...
                    proofMetadata: ProofMetadata): RlnRelayResult[bool] =
   ## returns true if there is another message in the  `nullifierLog` of the `rlnPeer` with the same
   ## epoch and nullifier as `proofMetadata`'s epoch and nullifier
   ## otherwise, returns false
   ## Returns an error if it cannot check for duplicates
 
-  let externalNullifier = proofMetadata.externalNullifier
   # check if the epoch exists
-  if not rlnPeer.nullifierLog.hasKey(externalNullifier):
+  let nullifier = proofMetadata.nullifier
```
alrevuelta, line 114: Maybe all this can be replace by something like (disregard types). `getOrDefault` might be useful.

```
return defaultTable.getOrDefault("epoch1", initTable[string, int]()).hasKey("key1")
```

A quick example would be something like this.
```
import tables

var defaultTable: Table[string, Table[string, int]] = {"epoch1":{"key1": 3, "key2": 3}.toTable,
                                                       "epoch2": {"key3": 3, "key4": 3}.toTable}.toTable

echo defaultTable.getOrDefault("epoch1", initTable[string, int]()).hasKey("key1")
# true
  
echo defaultTable.getOrDefault("epoch1", initTable[string, int]()).hasKey("key200")
# false

echo defaultTable.getOrDefault("epoch5", initTable[string, int]()).hasKey("key1")
# false
```

just an idea though, at your discretion.

waku/waku_rln_relay/rln_relay.nim
```diff
@@ -103,58 +103,54 @@ proc stop*(rlnPeer: WakuRLNRelay) {.async: (raises: [Exception]).} =
...
   ## saves supplied proofMetadata `proofMetadata`
   ## in the `nullifierLog` of the `rlnPeer`
   ## Returns an error if it cannot update the log
 
-  let externalNullifier = proofMetadata.externalNullifier
-  # check if the externalNullifier exists
-  if not rlnPeer.nullifierLog.hasKey(externalNullifier):
-    rlnPeer.nullifierLog[externalNullifier] = @[proofMetadata]
+  # check if the epoch exists
+  if not rlnPeer.nullifierLog.hasKey(epoch):
```
alrevuelta, line 136: maybe [hasKeyOrPut proc](https://nim-lang.org/docs/tables.html#hasKeyOrPut,OrderedTable[A,B],A,B) or [mgetOrPut proc](https://nim-lang.org/docs/tables.html#mgetOrPut,OrderedTable[A,B],A,B) could simplify this and make it safer?

at your discretion

tests/waku_rln_relay/test_waku_rln_relay.nim
```diff
@@ -632,30 +632,25 @@ suite "Waku rln relay":
 
     # check whether hasDuplicate correctly finds records with the same nullifiers but different secret shares
     # no duplicate for proof1 should be found, since the log is empty
```
alrevuelta, line 634: if 

> This could lead to a node only accepting one valid message sent during an epoch e, and reject all other messages sent during epoch e.

would it be difficult to reproduce it with a test? would like to i) verify that its indeed the case and ii) protect the code from future modification on such critical thing.

alrevuelta (changes requested):
Can we add a test to protect the code from future modifications and understand that

> This could lead to a node only accepting one valid message sent during an epoch e, and reject all other messages sent during epoch e.

can indeed happen?

tests/waku_rln_relay/test_waku_rln_relay.nim
rymnc: addressed in 6d20aef7
sorry, that statement is a little unclear. This bug is mostly to do with the nullifierLog not being mapped from Epoch to ProofMetadata, but rather ExternalNullifier to ProofMetadata. this line https://github.com/waku-org/nwaku/blob/88ff928213b92d20b54453687823afa510ee8d1a/waku/waku_rln_relay/rln_relay.nim#L117 does in fact allow the node to accept multiple valid messages sent during an epoch `e` from multiple senders.

waku/waku_rln_relay/rln_relay.nim
rymnc: addressed in 6d20aef7

waku/waku_rln_relay/rln_relay.nim
rymnc: addressed in 6d20aef7

tests/waku_rln_relay/test_waku_rln_relay.nim
rymnc: because the proofMetadata will be different from different senders. count this PR as a fix for the Table, and optimization to identify double signaling

waku/waku_rln_relay/rln_relay.nim
```diff
@@ -103,58 +103,49 @@ proc stop*(rlnPeer: WakuRLNRelay) {.async: (raises: [Exception]).} =
...
-    if matched.len != 0:
-      # there is a duplicate
-      return ok(true)
-
     # there is no duplicate
     return ok(false)
 
-  except KeyError as e:
+  except KeyError:
     return err("the epoch was not found")
```
Ivansete-status, line 126: Aside from PR scope but may be interesting:
```suggestion
  except KeyError:
    return err("the epoch was not found: " & getCurrentExceptionMsg())
```

waku/waku_rln_relay/rln_relay.nim
```diff
@@ -103,58 +103,49 @@ proc stop*(rlnPeer: WakuRLNRelay) {.async: (raises: [Exception]).} =
...
+      # be implemented here
       # TODO: slashing logic
       return ok()
-    # add proofMetadata to the log
-    rlnPeer.nullifierLog[externalNullifier].add(proofMetadata)
     return ok()
-  except KeyError as e:
-    return err("the external nullifier was not found") # should never happen
+  except KeyError:
+    return err("the epoch was not found") # should never happen
```
Ivansete-status, line 148: Just in case ;P
```suggestion
    return err("the epoch was not found: " & getCurrentExceptionMsg()) # should never happen
```

tests/waku_rln_relay/test_waku_rln_relay.nim
```diff
@@ -710,6 +705,55 @@ suite "Waku rln relay":
...
+    let index1 = MembershipIndex(5)
+    let index2 = MembershipIndex(6)
+
+    let rlnConf1 = WakuRlnConfig(rlnRelayDynamic: false,
+                                 rlnRelayCredIndex: some(index1),
+                                 rlnEpochSizeSec: 1,
+                                 rlnRelayTreePath: genTempPath("rln_tree", "waku_rln_relay_3"))
+    let wakuRlnRelay1Res = await WakuRlnRelay.new(rlnConf1)
+    if wakuRlnRelay1Res.isErr(): assert false, "failed to create waku rln relay: " & $wakuRlnRelay1Res.error
+    let wakuRlnRelay1 = wakuRlnRelay1Res.get()
```
Ivansete-status, line 718: The `valueOr` and `isOkOr` stew templates are very handy. In this case, we could have:
```suggestion
    let wakuRlnRelay1 = (await WakuRlnRelay.new(rlnConf1)).valueOr:
      assert false, "failed to create waku rln relay: " & $error
```

Ivansete-status (approved):
LGTM! Just minor comments on the surface

waku/waku_rln_relay/rln_relay.nim
alrevuelta: not sure thats addressed there. in any case not blocking.

alrevuelta (approved):

waku/waku_rln_relay/rln_relay.nim
rymnc: we don't actually need to get the value from the nullifierLog table, just check if it exists.

PR #2516, opened 2024-03-06
fix(rln-relay): handle empty metadata returned by getMetadata proc

### LINKED ISSUE #2471: chore: avoid execution error: metadata does not exist from ./register_rln.sh
### Background
nwaku-compose starting a node always shows a confusing error.

### Details
Every time the `./regiser_rln.sh` script is executed, the following error appears and we should avoid that because it causes confusion:
`execution error: metadata does not exist`

This can be tested from: https://github.com/waku-org/nwaku-compose

rymnc:
- **fix(rln-relay): silence error on startup when metadata is not found**
- **chore: fix fetching value from option**

# Description

This PR silences the execution error that is received when trying to fetch the rln metadata from the db.
Zerokit's PR: https://github.com/vacp2p/zerokit/pull/231
If it returns a buf of len 0, then we assume the metadata is not found, and proceed execution as usual.

# Changes

- [x] updated rln static lib version to v0.3.6
- [x] updated zerokit submodule
- [x] added test for empty metadata returned

## How to test

1. `make -j16 test`

## Issue

closes #2471

### DIFF
--- a/waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
+++ b/waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
@@ -669,8 +669,8 @@ method init*(g: OnchainGroupManager): Future[void] {.async.} =
   let metadataGetRes = g.rlnInstance.getMetadata()
   if metadataGetRes.isErr():
     warn "could not initialize with persisted rln metadata"
-  else:
-    let metadata = metadataGetRes.get()
+  elif metadataGetRes.get().isSome():
+    let metadata = metadataGetRes.get().value
     if metadata.chainId != uint64(g.chainId.get()):
       raise newException(ValueError, "persisted data: chain id mismatch")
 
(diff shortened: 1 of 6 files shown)


waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
```diff
@@ -669,8 +669,8 @@ method init*(g: OnchainGroupManager): Future[void] {.async.} =
   let metadataGetRes = g.rlnInstance.getMetadata()
   if metadataGetRes.isErr():
     warn "could not initialize with persisted rln metadata"
-  else:
-    let metadata = metadataGetRes.get()
+  elif metadataGetRes.get().isSome():
+    let metadata = metadataGetRes.get().get()
```
rymnc, line 673: not the cleanest, but oh well

gabrielmer (approved):
LGTM, thanks so much! 🤩

Ivansete-status (approved):
LGTM! Thanks! :100:

PR #2517, opened 2024-03-07
fix: use new negentropy instance everytime client sync is initiated

chaitanyaprem:
# Description

While fixing issue with sync protocol test, realized that the way we are using negentropy is not correct in the client side. 
Changed it to initiate a new instance whenever sync is invoked on client side. 
Also added free functions to cleanup negentropy instance and storage instance. 

# Changes

- [x] Make sync test between 2 nodes work
- [x] Add test for 3 node cyclic sync
- [x] Add more tests like: 

> - empty client <-> empty server
> - empty client <-> full server
> - full client <-> empty server
> - 100K with 1 random diff + simple compute time measure
> - 100K with 10K random diffs + simple compute time measure
- [x] add tests to verify single instance acting as server for more than 1 client.
> - 2 clients and 1 server with different diffs
> - 5 clients and 1 server with 1, 10, 100, 1000, 10000 diffs with time calculation for each sync


tests/waku_sync/test_protocol.nim
```diff
@@ -139,3 +140,73 @@ suite "Waku Sync - Protocol Tests":
         assert hashes.isOk(), $hashes.error
         check:
           hashes.value.len == 0
+
+
+    asyncTest "sync 3 nodes cyclic":
```
SionoiS, line 145: Multiparty sync is planned for later. Simple client <-> server is what we need to tests for now.

As for tests in general we may want to test with more messages ~10-100K with 1, 10, 100, 1K, 10K differences.

I'm not sure I understand what are we testing here?

Here's a couple of ideas;
empty client <-> empty server
empty client <-> full server
full client <-> empty server
100K with 1 random diff + simple compute time measure 
100K with 10K random diffs  + simple compute time measure
etc...

waku/waku_sync/protocol.nim
```diff
@@ -54,25 +54,32 @@ proc serverReconciliation(self: WakuSync, message: seq[byte]): Result[seq[byte],
   ok(payload)
 
 proc clientReconciliation(
-  self: WakuSync, message: seq[byte],
+  self: WakuSync, negentropy: pointer, message: seq[byte],
```
SionoiS, line 57: Don't we already have negentropy in `WakuSync` ?

waku/waku_sync/protocol.nim
```diff
@@ -54,25 +54,32 @@ proc serverReconciliation(self: WakuSync, message: seq[byte]): Result[seq[byte],
...
 
+#[ template freeNegentropyAndReturn(negentropy: pointer, ret: untyped) =
+  negentropyFree(negentropy)
+  return ret ]#
+
 proc request(self: WakuSync, conn: Connection): Future[Result[seq[WakuMessageHash], string]] {.async, gcsafe.} =
-  let request: seq[byte] = (await self.intitialization()).valueOr:
+  let negentropy = negentropyNew(self.storage, DefaultFrameSize)
+  let request: seq[byte] = (await self.intitialization(negentropy)).valueOr:
+    negentropyFree(negentropy)
```
SionoiS, line 77: I don't understand why do we need to free negentropy every request?

`negentropyFree(negentropy)` implie removing the storage as well no?

tests/waku_sync/test_protocol.nim
chaitanyaprem: Ok, just wanted to test 3 nodes syncing with each other, wanted to verify same node working as sync client and server.
Thanks for the input, my idea is to have tests similar to what you had mentioned above as next step:)

waku/waku_sync/protocol.nim
chaitanyaprem: We cannot keep using same negentropy instance each time sync has to be initated. There is a preventive check in the library to avoid that. Which lead me to think that a new instance has to be created on client side everytime we want to initiate a sync. This also takes care of scenarios such as 

If we want to do parallel sync with 2 different nodes, having a single instance of negentropy on client side would not work whereas having a single one on server side should work i guess(something i want to validate as well as part of tests).

waku/waku_sync/protocol.nim
chaitanyaprem: Explained above the need for it. 
negentropy freeing doesn't remove storage. If you notice the way the negentropy API is implemented. 
Storage is for storing hashes and negentropy is for the protocol i.e doing actual sync between client and server.

waku/waku_sync/protocol.nim
SionoiS: Yes, per our 1:1 discussion we do have to create new negentropy instance for each new sync "session".

tests/waku_sync/test_protocol.nim
SionoiS: Per our 1:1, I now understand!

chaitanyaprem:
Ran a sync test comparing sync for different diff sizes on my mac and here are a sample set of results showing sync time in seconds.
```

INF 2024-03-11 13:55:07.086+05:30 sync time in seconds tid=93700124 file=test_protocol.nim:416 msgsTotal=100000 diff=1 syncTime=0.01153300000000002

INF 2024-03-11 13:55:07.098+05:30 sync time in seconds tid=93700124 file=test_protocol.nim:426 msgsTotal=100000 diff=10 syncTime=0.008789000000000158

INF 2024-03-11 13:55:07.110+05:30 sync time in seconds tid=93700124 file=test_protocol.nim:436 msgsTotal=100000 diff=100 syncTime=0.008967000000000169

INF 2024-03-11 13:55:07.123+05:30 sync time in seconds tid=93700124 file=test_protocol.nim:446 msgsTotal=100000 diff=1000 syncTime=0.009673000000000265

INF 2024-03-11 13:55:07.147+05:30 sync time in seconds tid=93700124 file=test_protocol.nim:456 msgsTotal=100000 diff=10000 syncTime=0.02006400000000008
. (3.00s)
[Summary] 1 tests run (3.00s): 1 OK, 0 FAILED, 0 SKIPPED

```

PR #2523, opened 2024-03-11
chore: factory cleanup

gabrielmer:
# Description
Cleaning `node_factory.nim`. Only exporting `setupNode()` and `startNode()`.
# Changes

- [x] moving `retrieveDynamicBootstrapNodes()` to `app.nim`
- [x] not exporting utility procs in `node_factory.nim`

## Issue

extends #2441

### DIFF
--- a/apps/wakunode2/app.nim
+++ b/apps/wakunode2/app.nim
@@ -85,6 +85,38 @@ func node*(app: App): WakuNode =
 func version*(app: App): string =
   app.version
 
+## Retrieve dynamic bootstrap nodes (DNS discovery)
+
+proc retrieveDynamicBootstrapNodes*(dnsDiscovery: bool,
+                                    dnsDiscoveryUrl: string,
+                                    dnsDiscoveryNameServers: seq[IpAddress]):
+                                    Result[seq[RemotePeerInfo], string] =
+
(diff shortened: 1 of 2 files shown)


apps/wakunode2/app.nim
```diff
@@ -85,6 +85,38 @@ func node*(app: App): WakuNode =
 func version*(app: App): string =
   app.version
 
+## Retrieve dynamic bootstrap nodes (DNS discovery)
+
+proc retrieveDynamicBootstrapNodes*(dnsDiscovery: bool,
```
Ivansete-status, line 90: I think is better to keep this in the `node_factory.nim` so that the `libwaku` could potentially benefit from it when we add this feature. wdyt @jm-clius ?

SionoiS (approved):
Less public procs :muscle: 

Thank you!

Ivansete-status (approved):
LGTM! In the future, we may need to revert the mentioned proc movement.
Cheers

apps/wakunode2/app.nim
gabrielmer: Sounds good to me :) Will wait a bit before merging so we confirm it and if so I'll update the PR 🤩

PR #2524, opened 2024-03-11
chore: adding node factory tests

gabrielmer:
# Description
Adding simple test suite for the newly introduced factory layer.
Testing vectors are `setupNode()` and `startNode()`, which are the procs that will be initially exported by the factory.

# Changes

- [x] adding test suite in `tests/factory/test_node_factory.nim`
- [x] giving value to parameter `storeMessageDbUrl` in `defaultTestWakuNodeConf()` in order to test store being mounted

## How to test

run `make testfactory`

## Issue

extends #2441

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -179,7 +179,7 @@ testcommon: | build deps
 ##########
 ## Waku ##
 ##########
-.PHONY: testwaku wakunode2 testwakunode2 testfactory example2 chat2 chat2bridge
+.PHONY: testwaku wakunode2 testwakunode2 example2 chat2 chat2bridge
 
 # install anvil only for the testwaku target
 testwaku: | build deps anvil librln
@@ -197,10 +197,6 @@ benchmarks: | build deps librln
 testwakunode2: | build deps librln
 	echo -e $(BUILD_MSG) "build/$@" && \
 		$(ENV_SCRIPT) nim testwakunode2 $(NIM_PARAMS) waku.nims
-
-testfactory: | build deps librln
-	echo -e $(BUILD_MSG) "build/$@" && \
-		$(ENV_SCRIPT) nim testfactory $(NIM_PARAMS) waku.nims
 
 example2: | build deps librln
 	echo -e $(BUILD_MSG) "build/$@" && \
--- a/Makefile
+++ b/Makefile
@@ -179,7 +179,7 @@ testcommon: | build deps
 ##########
 ## Waku ##
 ##########
-.PHONY: testwaku wakunode2 testwakunode2 example2 chat2 chat2bridge
+.PHONY: testwaku wakunode2 testwakunode2 testfactory example2 chat2 chat2bridge
 
 # install anvil only for the testwaku target
 testwaku: | build deps anvil librln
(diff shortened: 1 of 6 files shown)


SionoiS (approved):
LGTM Thanks

Ivansete-status (changes requested):
Wow! This is great! The tests look nice indeed!
Nevertheless, I suggest simplifying the addition to this new test module. For that, I suggest to just add the `tests/factory/test_node_factory.nim` as an import within `all_tests_waku.nim`. Then, nothing else needs to be changed.

gabrielmer:
> Wow! This is great! The tests look nice indeed! Nevertheless, I suggest simplifying the addition to this new test module. For that, I suggest to just add the `tests/factory/test_node_factory.nim` as an import within `all_tests_waku.nim`. Then, nothing else needs to be changed.

The idea was that if we were going to treat Factory as a whole new layer, then it would make sense to have all the logic separate from wakunode2 and from the node, as it's currently done.

Don't have any objections though, I'm good either way :)

Ivansete-status:
> > Wow! This is great! The tests look nice indeed! Nevertheless, I suggest simplifying the addition to this new test module. For that, I suggest to just add the `tests/factory/test_node_factory.nim` as an import within `all_tests_waku.nim`. Then, nothing else needs to be changed.
> 
> The idea was that if we were going to treat Factory as a whole new layer, then it would make sense to have all the logic separate from wakunode2 and from the node, as it's currently done.
> 
> Don't have any objections though, I'm good either way :)

ah yes! I think that is very sensible reasoning and I don't dislike it completely :)

However, let me elaborate on my thoughts a bit.

We have the following "all_tests" modules:
- `all_tests_common.nim`. This is aimed at testing helper modules, and btw, seems not to be run in CI.
- `all_tests_wakunode2.nim`. For testing the `wakunode2` app modules.
- `all_tests_waku.nim`. We have this module as a big bucket where we test protocol modules, but also REST, and RLN, for example.
 
Reasons why I can't approve :)
- I see a risk that we end up not testing the factory module in CI, equally as it is currently happening with the `all_tests_common` module.

- The `test_node_factory` module's name denotes a clear coupling with the node itself and as we know, it helps to create waku nodes :) Therefore I consider this module should fall under `all_tests_waku` as well.

- I'd like to have the `Makefile` as simple a possible. I think that we could also simplify the `Makefile` more and just have one single test binary: [test_waku](https://github.com/waku-org/nwaku/blob/master/Makefile#L185)

Cheers!

gabrielmer:
Thanks for the detailed answer! ❤️ 

Quick reply to the points you mentioned

> * I see a risk that we end up not testing the factory module in CI, equally as it is currently happening with the `all_tests_common` module.

In the PR we're adding the new test module in the same places as `all_tests_wakunode2.nim`, including the CI files, so with this there shouldn't be an issue. I do agree that it's an extra moving part though 

> * The `test_node_factory` module's name denotes a clear coupling with the node itself and as we know, it helps to create waku nodes :) Therefore I consider this module should fall under `all_tests_waku` as well.

True, `all_tests_waku` contains tests that don't strictly belong to the "node" layer. Or in other words, it's not like the case with the implementations itself, where there's a clear decoupling between app and the node which we made sure not to trespass.

Here it's already mixed in a way. We can use REST as an example to see this.
I agree there's a coupling with the node itself but I see it more as a dependency rather than a coupling. The factory depends on the node but the node doesn't depend on the factory - they are not equal.

Taking into account that it's already mixed and we don't follow the same logical separation as the implementation, I'm also ok on having the test included in `all_tests_waku`. Don't have a strong opinion on it 😇 

> * I'd like to have the `Makefile` as simple a possible. I think that we could also simplify the `Makefile` more and just have one single test binary: [test_waku](https://github.com/waku-org/nwaku/blob/master/Makefile#L185)


Agree 10000% with this approach. Always in for simplicity :)

Ivansete-status (approved):
Wondeful PR! Thank you so much indeed for the patience and flexibility! :100:

PR #2529, opened 2024-03-12
fix(rest): filter/v2/subscriptions response

richard-ramos:
# Description
Returns a "OK" for the response when subscription is succesful, and also removes the status code from the response as it is not part of the API. Ping responses also return "OK" when succesful.

## Issue

closes https://github.com/waku-org/nwaku/issues/2286

### DIFF
--- a/tests/wakunode_rest/test_rest_filter.nim
+++ b/tests/wakunode_rest/test_rest_filter.nim
@@ -246,7 +246,7 @@ suite "Waku v2 Rest API - Filter V2":
       pingResponse.status == 200
       $pingResponse.contentType == $MIMETYPE_JSON
       pingResponse.data.requestId == "9999"
-      pingResponse.data.statusDesc.len() == 0
+      pingResponse.data.statusDesc == "OK"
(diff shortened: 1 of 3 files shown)


NagyZoltanPeter (approved):
LGTM! Thanks for this great catch!
Still wonder how it become like this... probably it was me... :-o

richard-ramos:
I did a quick review and we're good!

- https://github.com/waku-org/waku-rest-api/blob/fd9ee5905850eb110973f5c5c8051d9f6fabe535/api-spec/schemas/apitypes.yaml#L95-L103
- https://github.com/waku-org/waku-rest-api/blob/master/api-spec/filterapi_v2_subscription_ping.yaml#L19-L24
- https://github.com/waku-org/waku-rest-api/blob/master/api-spec/filterapi_v2_subscriptions.yaml#L19-L24
- https://github.com/waku-org/waku-rest-api/blob/master/api-spec/filterapi_v2_subscriptions_all.yaml#L15-L20

PR #2530, opened 2024-03-13
chore: sync optimizations

chaitanyaprem:
# Description

# Changes

- [x] optimize initiate and server-reconcile methods to not use fixed alloc size for output
- [x] ~~optimize client-reconcile to use callback~~ (After trying this out, realized that number of malloc's and memory copies are not changing. Hence decided to leave it in current form and can be revisited if a performance bottleneck is found in future).
- [x] Don't start periodic sync if sync-interval is set to 0 seconds.
- [x] Implement and do error handling in all methods


waku/waku_sync/protocol.nim
```diff
@@ -55,17 +55,19 @@ proc serverReconciliation(self: WakuSync, payload: NegentropyPayload): Result[Ne
...
-proc intitialization(self: WakuSync): Future[Result[NegentropyPayload, string]] {.async.} =
-  return self.negentropy.initiate()
+  return negentropy.clientReconcile(payload, haveHashes, needHashes)
 
 proc request(self: WakuSync, conn: Connection): Future[Result[seq[WakuMessageHash], string]] {.async, gcsafe.} =
-  let payload = (await self.intitialization()).valueOr:
+  let negentropy = Negentropy.new(self.storage, DefaultFrameSize)
+
+  let payload =  negentropy.initiate().valueOr:
+    free(negentropy)
```
SionoiS, line 69: Would have to wrap this up. Unsafe code only in `raw_bindings`

SionoiS (commented):
Progress :muscle: !

waku/waku_sync/protocol.nim
```diff
@@ -66,7 +66,7 @@ proc request(self: WakuSync, conn: Connection): Future[Result[seq[WakuMessageHas
   let negentropy = Negentropy.new(self.storage, DefaultFrameSize)
 
   let payload =  negentropy.initiate().valueOr:
-    free(negentropy)
+    delete(negentropy)
```
SionoiS, line 69: Style wise, we use `negentropy.delete()` usually.

.gitmodules
```diff
@@ -168,4 +168,4 @@
 	ignore = untracked
 	path = vendor/negentropy
 	url = https://github.com/waku-org/negentropy.git
-	branch = feat/c-wrapper
\ No newline at end of file
+	branch = chore/optimizations
```
SionoiS, line 171: I'm a bit confused on the best way to use gitmodule.

Would it not be better to update one branch instead of using multiple branch?

My only experience so far has been pain and suffering. :performing_arts:  :rofl:

waku/waku_sync/raw_bindings.nim
```diff
@@ -80,67 +74,74 @@ proc toBufferSeq(buffLen: uint, buffPtr: ptr Buffer): seq[Buffer] =
...
+
+
 proc new*(T: type Storage): T =
   #TODO db name and path
   let storage = storage_init("", "")
 
-  #TODO error handling
+#[ TODO: Uncomment once we move to lmdb   
+  if storage == nil:
+    raise newException(ValueError, "storage initialization failed") ]#
```
SionoiS, line 132: We favor `Result[T, string]` instead of exceptions

waku/waku_sync/raw_bindings.nim
```diff
@@ -80,67 +74,74 @@ proc toBufferSeq(buffLen: uint, buffPtr: ptr Buffer): seq[Buffer] =
...
 
   return storage
 
+proc delete*(storage: Storage) =
+  free(storage)
+
```
SionoiS, line 138: I tried to look for "destructors" in Nim but didn't find anything. I guess we can't automate this. :disappointed:

waku/waku_sync/raw_bindings.nim
```diff
@@ -152,29 +153,36 @@ proc insert*(storage: Storage, id: int64, hash: WakuMessageHash): Result[void, s
...
-proc new*(T: type Negentropy, storage: Storage, frameSizeLimit: uint64): T =
-  let negentropy = constructNegentropy(storage, frameSizeLimit)
-  
-  #TODO error handling
+proc `==` *(a : Negentropy, b : pointer):bool{.borrow.}
 
+proc new*(T: type Negentropy, storage: Storage, frameSizeLimit: uint64): T {.raises: [ValueError] .} =
+  let negentropy = constructNegentropy(storage, frameSizeLimit)
+  if negentropy == nil:
+    raise newException(ValueError, "negentropy initialization failed due to lower framesize")
```
SionoiS, line 168: `: Result[T, string] = ` instead

SionoiS (approved):
Very nice!

.gitmodules
chaitanyaprem: Yeah, i will switch this to use master. 
Just using branches there in case i want to revert something, but now i can start moving to master.

I agree, My experience setting this sub-module up also has been painful especially with branches.

waku/waku_sync/raw_bindings.nim
chaitanyaprem: I used an exception as this will propagate way up to the app. Did not want to keep sending errors up functions.

SionoiS: I'm not sure when it is appropriate to use exception in Nim. Rust only use `Result`

waku/waku_sync/protocol.nim
```diff
@@ -66,14 +66,14 @@ proc request(self: WakuSync, conn: Connection): Future[Result[seq[WakuMessageHas
   let negentropy = Negentropy.new(self.storage, DefaultFrameSize)
 
   let payload =  negentropy.initiate().valueOr:
-    free(negentropy)
-
+    negentropy.delete()
```
Ivansete-status, line 69: We could have something like:
```
defer:
  negentropy.delete()
```
That way all possible return cases are covered.

Ivansete-status (approved):
Overall LGTM! Thanks!
Just added a couple of comments. I'm a bit worried about any possible mem leak. Make sure we use `defer: free_result` (or the required call) so that always freeing memory before leaving a proc.

Ivansete-status:
@chaitanyaprem - My bad! I didn't notice the tests were failing. All should be green but the `js` tests.

SionoiS:
> @chaitanyaprem - My bad! I didn't notice the tests were failing. All should be green but the `js` tests.

Yes, we are not merging to master so it's less important but we do require CI success. I also didn't notice :|

chaitanyaprem:
> Overall LGTM! Thanks! Just added a couple of comments. I'm a bit worried about any possible mem leak. Make sure we use `defer: free_result` (or the required call) so that always freeing memory before leaving a proc.

Thanks, i was not aware of defer in nim. was looking for something like this we have it in golang. I was thinking i would have to use a template, but that was becoming more complex as Future was being returned.

waku/waku_sync/raw_bindings.nim
chaitanyaprem: I was looking at Store protocol implementation and just followed it.

PR #2531, opened 2024-03-13
chore: content_script_version_2: add simple protection and rm messages_backup if exists

Ivansete-status:
## Description
This is aimed to add a simple protection. If the `messages_backup` exists, then delete it.
Notice that this would only protect the users that are now running a version lower than `v0.26.0`, and update their nodes to a version >= `v0.27.0`.

## How to test

1. Run a node with store protocol enabled for a while and version ' v0.25.0`. Wait until some messages are written (I've tested it with ~3.6GB database.)
2. Run a node from this branch and 

## Issue

closes https://github.com/waku-org/nwaku-compose/issues/75

### DIFF
--- a/migrations/message_store_postgres/content_script_version_2.nim
+++ b/migrations/message_store_postgres/content_script_version_2.nim
@@ -1,4 +1,5 @@
 const ContentScriptVersion_2* = """
+DROP TABLE IF EXISTS messages_backup;
 ALTER TABLE messages RENAME TO messages_backup;
 ALTER TABLE messages_backup DROP CONSTRAINT messageIndex;
 


SionoiS (approved):
Am I understanding correctly that this would fix the case where the migration errors before `DROP TABLE IF EXISTS messages_backup;` at the end?

If yes then it would fix the problem but if it fails again then the backup is lost no?

Up to you if loosing the backup is an acceptable risk.

NagyZoltanPeter (approved):
LGTM!

Ivansete-status:
> Am I understanding correctly that this would fix the case where the migration errors before `DROP TABLE IF EXISTS messages_backup;` at the end?
> 
> If yes then it would fix the problem but if it fails again then the backup is lost no?
> 
> Up to you if loosing the backup is an acceptable risk.

Cool thanks! I adapted it according to your advise :partying_face: 
https://github.com/waku-org/nwaku/pull/2531/commits/39b6d24c47ca1a79da3d3ece687dd70f84c00f93

PR #2532, opened 2024-03-13
fix(store): retention policy regex

### LINKED ISSUE #2522: bug: store size retention policy parsing fails for string define fraction number.
### Problem
When using `  --store-message-retention-policy=size:1.5GB` setting wakunode fails to start with the following error.
```
nwaku-1  | ERR 2024-03-11 13:50:44.734+00:00 4/7 Mounting protocols failed              topics="wakunode main" tid=1 file=wakunode2.nim:150 error="failed to create retention policy: invalid 'store message retention policy' format: size:1.5GB"
```
It seem we cannot handle properly fraction numbers in this case, neither `1.5GB` nor `1,5GB` works.

### Impact
Node may fail to start. User must use different size definition.

### nwaku version/commit hash
Found in 0.26.0-rc.0, but probably it is a much older problem since it fails in regex check of format.
https://github.com/waku-org/nwaku/blob/5b87e12b143479b1d5fd64a6cdc60f9e1ddbe958/waku/waku_archive/retention_policy/builder.nim#L27

richard-ramos:
# Description
Allows the usage of optional decimal values in the storage policy, and limits the size unit in the regexp itself. Since i preeventively lowercase the ret policy argument, I ended up removing the different combinations of uppercase/lowercase and just compare against `gb` or `mb`.

## Issue

closes #2522

### DIFF
--- a/waku/waku_archive/retention_policy/builder.nim
+++ b/waku/waku_archive/retention_policy/builder.nim
@@ -17,12 +17,13 @@ import
 proc new*(T: type RetentionPolicy,
           retPolicy: string):
           RetentionPolicyResult[Option[RetentionPolicy]] =
+  let retPolicy = retPolicy.toLower
 
   # Validate the retention policy format
   if retPolicy == "" or retPolicy == "none":
     return ok(none(RetentionPolicy))
 
-  const StoreMessageRetentionPolicyRegex = re"^\w+:\w+$"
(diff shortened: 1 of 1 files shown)


Ivansete-status (approved):
Wonderful! Thanks! :100:

SionoiS (approved):
thanks!

Issue #2491, opened 2024-02-29
bug: metadata protocol will disconnect light clients


richard-ramos:
### Problem
The function `proc onPeerMetadata(pm: PeerManager, peerId: PeerId) {.async.}` uses the metadata protocol to determine whether a connected peer supports a shard or not, and disconnect the peer otherwise.

This works fine for nodes that support relay. However in the case of nodes that support only light protocols (like mobile on status, which only consumes lightpush/filter/peerexchange,  it will fail.

The metadata protocol relies on the ENR containing the shards, yet in the case of request/response protocols, the ENR is not updated at all, so the peer get disconnected

SionoiS:
Light nodes don't use discovery, so the easy fix it to trigger a shard update in Waku Metadata when using filter subscribe.

Does it make sense to also update ENRs when a node uses Relay AND Filter as client? :thinking:

richard-ramos:
Something else to take into account: When using lightmode, peer exchange will be used instead of discv5 to retrieve peers. At that point no subscription to filter protocol has happened yet, so a shard update triggered by filter subscribe will not have happened. Peer exchange also does not use shard data at all so we can't trigger an update when doing a request, can we? :thinking:

alrevuelta:
mm unsure if bug or feature. even light clients should have some identifier (eg mounted waku metadata) because otherwise they may be interacting with nodes in other networks.

so i would say light clients should mount waku metadata (which is dead simple).

richard-ramos:
> @alrevuelta: so i would say light clients should mount waku metadata (which is dead simple).

Light clients (using go-waku) already mount waku metadata, and the logic that deals with the clusterID is executed correctly (I assume by network you mean nodes sharing the same clusterID?). 

The issue described here happens due to this code https://github.com/waku-org/nwaku/blob/55ff6674780f35a453ec312cdb96a5c0f4f1212b/waku/node/peer_manager/peer_manager.nim#L360-L362 

It assumes that the nodes will have a list of shards at all times, yet this is an undefined behaviour for light clients as they do not update their ENR nor are discoverable, and you can see here https://github.com/waku-org/nwaku/blob/55ff6674780f35a453ec312cdb96a5c0f4f1212b/waku/node/waku_node.nim#L191 how metadata protocol is instantiated using the node's own ENR as well as the list of topics you subscribe to in relay. (This behavior is also shared by go-waku),

fryorcraken:
[...]

PR #2533, opened 2024-03-14
fix(2491): Fix metadata protocol disconnecting light nodes

AlejandroCabeza:
# Description
Fix for metadata protocol disconnecting light nodes: Now non-relay peers won't be checked for `shardId`.
Closes: #2491

# Changes
- [X] Fix the issue
- [X] Add test cases covering the issue


waku/node/waku_node.nim
```diff
@@ -575,6 +575,7 @@ proc filterSubscribe*(node: WakuNode,
   if pubsubTopic.isSome():
     info "registering filter subscription to content", pubsubTopic=pubsubTopic.get(), contentTopics=contentTopics, peer=remotePeer.peerId
 
+    let contentTopics = contentTopics.mapIt($it)
```
SionoiS, line 578: What is this for?

waku/node/peer_manager/peer_manager.nim
```diff
@@ -331,6 +333,11 @@ proc onConnEvent(pm: PeerManager, peerId: PeerID, event: ConnEvent) {.async.} =
     of ConnEventKind.Disconnected:
       discard
 
+proc peerHasProtocol(
+    peerManager: PeerManager, peerId: PeerId, protocol: string
+): bool {.inline, raises: [].} =
+  return peerManager.peerStore[ProtoBook][peerId].contains(protocol)
+
```
SionoiS, line 340: I would put this in `waku_peer_store.nim` instead.

Or maybe use https://github.com/waku-org/nwaku/blob/0894f0cfea9fac17348607cb50768e17bd788660/waku/node/peer_manager/waku_peer_store.nim#L112

tests/node/peer_manager/test_peer_manager.nim
SionoiS, line 1: I would put this with the other tests in `tests/test_peer_manager.nim`

SionoiS (approved):
Logic is good but organizationally I would prefer changes.

Thanks!

waku/node/peer_manager/peer_manager.nim
```diff
@@ -357,7 +364,10 @@ proc onPeerMetadata(pm: PeerManager, peerId: PeerId) {.async.} =
       reason = "different clusterId reported: " & $pm.wakuMetadata.clusterId & " vs " & $clusterId
       break guardClauses
 
-    if not metadata.shards.anyIt(pm.wakuMetadata.shards.contains(it)):
+    if (
+      pm.peerHasProtocol(peerId, WakuRelayCodec) and
```
Ivansete-status, line 368: This is great but I would move this check to line 347, so that we skip the peer earlier and the node doesn't spend effort dialing a node that will be skipped later on.

Sth like that:
```
  if not pm.peerHasProtocol(peerId, WakuRelayCodec):
    debug "skipping peer because it doesn't have relay mounted", peer_id = peerId
    return
```

That would also help to mitigate any possible protocol issue. f.e., we came across a protobuf incompatibility issue that made the shards number not properly parsed.

waku/node/waku_node.nim
Ivansete-status: It seems to be meant to convert the seq to a string seq but seems not to be necessary

Ivansete-status (approved):
LGTM! Thanks for it!
I just added some recommendations that I hope you find useful.
Also interesting to make the test part of the CI by either:
- properly importing the new `test_peer_manager` module.
  or
- Following SP's suggestion.

waku/node/waku_node.nim
AlejandroCabeza: I just realised that is wrong, that doesn't work.

The idea is the parameter `contentTopics`  is of type `ContentTopic | seq[ContentTopic]`. Given the expected type for `subscribe` is `seq[ContentTopic]`, that wraps the `ContentTopic` inside a `seq`, to comply with it.

Fixing it.

tests/node/peer_manager/test_peer_manager.nim
SionoiS: Good point! Open an issue and lets do that instead.

AlejandroCabeza: I don't think there's a need for that for the time being. I'm slowly but surely moving them to where "they belong", unless you guys wanna help out with that 😛

waku/node/peer_manager/peer_manager.nim
AlejandroCabeza: I think the order is intended. The check in question is two-fold:
* The peer being WakuRelay (you could do that earlier, certainly)
* At least one of the peer shards needs to be in the server's metadata. For this one, you need to wait, at least, until 355 is executed, as that's where you retrieve the metadata from the server.

In any case, if what I said is true, you could skip two checks (moving the last one two places up).

Any opinions?

Ivansete-status: > Any opinions?

The idea of my comment was just to make it a bit simpler and return as early as possible. I believe the `metadata` protocol only applies to "relay" connections.

Btw, feel free to merge the PR if you consider it appropriate. I don't want to be a blocker :D

richard-ramos (approved):

PR #2544, opened 2024-03-20
feat(c-bindings): rln relay

### LINKED ISSUE #2454: chore: support RLN-relay in libwaku
### Background
Epic: https://github.com/waku-org/nwaku/issues/2420

### Pre-requisite
The following should be completed before starting this so that we can benefit from having a single configuration entry to the `waku_node`: https://github.com/waku-org/nwaku/issues/2441

### Details
Apply the needed changes in `libwaku.nim` and `libwaku.h` so that we bring support to RLN-relay there.

The `mountRlnRelay` should be invoked in https://github.com/waku-org/nwaku/blob/2060cfab675c1e36e36159e5818c9ce479b97992/library/waku_thread/inter_thread_communication/requests/node_lifecycle_request.nim#L193 when `conf.rlnRelay == true`.

In this case, we might need to allow accepting more config options in the JSON file. Example of how JSON config file is being passed: https://github.com/waku-org/nwaku/blob/2060cfab675c1e36e36159e5818c9ce479b97992/examples/cbindings/waku_example.c#L271

Notice that is interesting to extend the `cwaku_example` so that we can test the new RLN-relay feature from there.

From a very high level, the libwaku node should also support these operations:
- appendRLNProof
- validateMessage

richard-ramos:
# Description
Exposes the rln relay configuration to the bindings.

## Issue

closes #2454

### DIFF
--- a/library/waku_thread/config.nim
+++ b/library/waku_thread/config.nim
@@ -1,4 +1,5 @@
 import std/[json, strformat, options]
+import std/sequtils
 import
   libp2p/crypto/crypto,
   libp2p/crypto/secp,
@@ -6,6 +7,7 @@ import
   ../../waku/waku_enr/capabilities,
   ../../waku/common/utils/nat,
   ../../waku/factory/external_config,
+  ../../waku/waku_core/message/default_values,
   ../../waku/node/waku_node,
   ../../waku/node/config,
   ../events/json_base_event
@@ -82,6 +84,20 @@ proc parseRelay(jsonNode: JsonNode, conf: var WakuNodeConf, errorResp: var strin
 
   return true
 
+proc parseClusterId(jsonNode: JsonNode, conf: var WakuNodeConf, errorResp: var string): bool =
+  if not jsonNode.contains("relay"):
+    errorResp = "relay attribute is required"
+    return false
+
+  if jsonNode.contains("clusterId"):
+    if jsonNode["clusterId"].kind != JsonNodeKind.JInt:
+      errorResp = "The clusterId config param should be an int"
+      return false
+    else:
+      conf.clusterId = uint32(jsonNode["clusterId"].getInt())
+
+  return true
+
 proc parseStore(
     jsonNode: JsonNode,
     conf: var WakuNodeConf,
@@ -149,6 +165,29 @@ proc parseTopics(jsonNode: JsonNode, conf: var WakuNodeConf) =
   else:
     conf.pubsubTopics = @["/waku/2/default-waku/proto"]
 
+proc parseRLNRelay(jsonNode: JsonNode, conf: var WakuNodeConf, errorResp: var string): bool =
+  if not jsonNode.contains("rln-relay"):
+    return true
+
+  let jsonNode = jsonNode["rln-relay"]
+  if not jsonNode.contains("enabled"):
+    errorResp = "rlnRelay.enabled attribute is required"
+    return false
+
+  conf.rlnRelay = jsonNode["enabled"].getBool()
+  conf.rlnRelayCredPath = jsonNode{"cred-password"}.getStr()
+  conf.rlnRelayEthClientAddress = EthRpcUrl.parseCmdArg(jsonNode{"eth-client-address"}.getStr("http://localhost:8540"))
+  conf.rlnRelayEthContractAddress = jsonNode{"eth-contract-address"}.getStr()
+  conf.rlnRelayCredPassword = jsonNode{"cred-password"}.getStr()
+  conf.rlnRelayUserMessageLimit = uint64(jsonNode{"user-message-limit"}.getInt(1))
+  conf.rlnEpochSizeSec = uint64(jsonNode{"epoch-sec"}.getInt(1))
+  conf.rlnRelayCredIndex = some(uint(jsonNode{"membership-index"}.getInt()))
+  conf.rlnRelayDynamic = jsonNode{"dynamic"}.getBool()
+  conf.rlnRelayTreePath = jsonNode{"tree-path"}.getStr()
+  conf.rlnRelayBandwidthThreshold = jsonNode{"bandwidth-threshold"}.getInt()
+
+  return true
+
 proc parseConfig*(
     configNodeJson: string,
     conf: var WakuNodeConf,
@@ -199,6 +238,14 @@ proc parseConfig*(
     errorResp = "Exception calling parseRelay: " & getCurrentExceptionMsg()
     return false
 
+  # clusterId
+  try:
+    if not parseClusterId(jsonNode, conf, errorResp):
+      return false
+  except Exception, KeyError:
+    errorResp = "Exception calling parseClusterId: " & getCurrentExceptionMsg()
+    return false
+
   # topics
   try:
     parseTopics(jsonNode, conf)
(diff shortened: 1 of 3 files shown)


Ivansete-status (commented):
It looks great indeed!

Nevertheless, I wonder if we could split the PR in two:
- refactor node initialization
- addition of rln config

cheers!

richard-ramos:
> refactor node initialization

Done in https://github.com/waku-org/nwaku/pull/2547

> addition of rln config

I'll use this PR for this once I rebase it against #2547

richard-ramos:
Done! PR rebased!

library/waku_thread/config.nim
```diff
@@ -82,6 +84,20 @@ proc parseRelay(jsonNode: JsonNode, conf: var WakuNodeConf, errorResp: var strin
...
   return true
 
+proc parseClusterId(jsonNode: JsonNode, conf: var WakuNodeConf, errorResp: var string): bool =
+  if not jsonNode.contains("relay"):
+    errorResp = "relay attribute is required"
+    return false
+
+  if jsonNode.contains("clusterId"):
+    if jsonNode["clusterId"].kind != JsonNodeKind.JInt:
+      errorResp = "The clusterId config param should be a int"
```
Ivansete-status, line 94: Tiny typo :)
```suggestion
      errorResp = "The clusterId config param should be an int"
```

library/waku_thread/config.nim
```diff
@@ -149,6 +165,29 @@ proc parseTopics(jsonNode: JsonNode, conf: var WakuNodeConf) =
   else:
     conf.topics = @["/waku/2/default-waku/proto"]
 
+proc parseRLNRelay(jsonNode: JsonNode, conf: var WakuNodeConf, errorResp: var string): bool =
+  if not jsonNode.contains("rlnRelay"):
```
Ivansete-status, line 169: I wonder if we should use the same style as the `wakunode` has. Hyphen-separated terms. 
```suggestion
  if not jsonNode.contains("rln-relay"):
```

library/waku_thread/config.nim
```diff
@@ -149,6 +165,29 @@ proc parseTopics(jsonNode: JsonNode, conf: var WakuNodeConf) =
...
+  conf.rlnRelayCredPath = jsonNode{"credentialPassword"}.getStr()
+  conf.rlnRelayEthClientAddress = EthRpcUrl.parseCmdArg(jsonNode{"ethClientAddress"}.getStr("http://localhost:8540"))
+  conf.rlnRelayEthContractAddress = jsonNode{"contractAddress"}.getStr()
+  conf.rlnRelayCredPassword = jsonNode{"credentialPassword"}.getStr()
+  conf.rlnRelayUserMessageLimit = uint64(jsonNode{"userMessageLimit"}.getInt(1))
+  conf.rlnEpochSizeSec = uint64(jsonNode{"epochSizeSec"}.getInt(1))
+  conf.rlnRelayCredIndex = some(uint(jsonNode{"membershipIndex"}.getInt()))
+  conf.rlnRelayDynamic = jsonNode{"dynamic"}.getBool()
+  conf.rlnRelayTreePath = jsonNode{"treePath"}.getStr()
+  conf.rlnRelayBandwidthThreshold = jsonNode{"bandwidthThreshold"}.getInt()
```
Ivansete-status, line 187: Analogue to my previous comment, I suggest using the same names as in `wakunode` app:
```
$ ./build/wakunode2 --help | grep rln
 --rln-relay-cred-path     The path for peristing rln-relay credential.
 --rln-relay-eth-client-address  HTTP address of an Ethereum testnet client e.g., http://localhost:8540/
 --rln-relay-eth-contract-address  Address of membership contract on an Ethereum testnet.
 --rln-relay-cred-password  Password for encrypting RLN credentials.
 --rln-relay-eth-private-key  Private key for broadcasting transactions.
 --rln-relay-user-message-limit  Set a user message limit for the rln membership registration. Must be a positive
 --rln-relay-epoch-sec     Epoch size in seconds used to rate limit RLN memberships. Default is 1 second.
 --rln-relay               Enable spam protection through rln-relay: true|false [=false].
 --rln-relay-membership-index  the index of the onchain commitment to use.
 --rln-relay-dynamic       Enable  waku-rln-relay with on-chain dynamic group management: true|false
 --rln-relay-id-key        Rln relay identity secret key as a Hex string.
 --rln-relay-id-commitment-key  Rln relay identity commitment key as a Hex string.
 --rln-relay-tree-path     Path to the RLN merkle tree sled db (https://github.com/spacejam/sled).
 --rln-relay-bandwidth-threshold  Message rate in bytes/sec after which verification of proofs should happen [=0].
 --rln-relay-tree-path     Path to the RLN merkle tree sled db (https://github.com/spacejam/sled).
warning: could not open directory 'postgres-data/': Permission denied
```

library/waku_thread/inter_thread_communication/requests/node_lifecycle_request.nim
```diff
@@ -63,11 +63,32 @@ proc createNode(configJson: cstring): Future[Result[WakuNode, string]] {.async.}
...
+    conf.clusterId = twnClusterConf.clusterId
+    conf.rlnRelay = twnClusterConf.rlnRelay
+    conf.rlnRelayEthContractAddress = twnClusterConf.rlnRelayEthContractAddress
+    conf.rlnRelayDynamic = twnClusterConf.rlnRelayDynamic
+    conf.rlnRelayBandwidthThreshold = twnClusterConf.rlnRelayBandwidthThreshold
+    conf.discv5Discovery = twnClusterConf.discv5Discovery
+    conf.discv5BootstrapNodes =
+      conf.discv5BootstrapNodes & twnClusterConf.discv5BootstrapNodes
+    conf.rlnEpochSizeSec = twnClusterConf.rlnEpochSizeSec
+    conf.rlnRelayUserMessageLimit = twnClusterConf.rlnRelayUserMessageLimit
```
Ivansete-status, line 89: We have a similar structure in https://github.com/waku-org/nwaku/blob/18a053591d8d18b8f066d4bc919c3910cb4295c3/apps/wakunode2/wakunode2.nim#L91-L109

I wonder if we could encapsulate that somehow. Maybe in a separate PR

Ivansete-status (approved):
LGTM! Thanks!
I just added some nitpick comments ;P
Cheers

library/waku_thread/config.nim
richard-ramos: I can use the hyphens to separate the words but adding the `rln` prefix I'm not sure, because it currently looks like this:

```json
{
   "port": 123,
   "host": "0.0.0.0",
   "relay": true,
   "rlnRelay": {
          "enabled": "true",
         "ethClientAddress": "http://localhost:8545",
         "membershipIndex": 2,
         ...
    }
}
```
And seems to me that adding the `rln` prefix would be redundant, as the values are already part of a `rlnRelay` section. A possibility could be to put everything a single level deep (i.e., move all the items from `rlnRelay` into the root json, but seems to me that it's not benefiting from the possibilities of json allowing more logical groups.

Would something like this work?  (i.e. having the `rln-relay` prefix be instead the key that holds all the rln related configs):
```json
```json
{
   "port": 123,
   "host": "0.0.0.0",
   "relay": true,
   "rln-relay": {
          "enabled": "true",
         "eth-client-address": "http://localhost:8545",
         "membership-index": 2,
         "bandwidth-threshold": 1,
         ...
    }
}
```

Ivansete-status: Yes, I think this is the best balanced option (simplicity vs consistency)
image

SionoiS (approved):
I got a feeling of deja vu...

LGTM

richard-ramos:
> I got a feeling of deja vu...
> 
> LGTM

deja vu

PR #2548, opened 2024-03-24
fix: postgres_driver.nim  make sure the partition list is loaded in correct order

Ivansete-status:
## Description
This PR fixes an issue where the list of partitions wasn't obtained correctly.
A node operator, @redwhiteeee , informed about the error:

image

That was caused because the list of partitions was obtained in random order, as can be seen in the following image:

image

## Changes

- [x] Obtain the current partition set ordered chronologically.


redwhiteeee:
Thank you

SionoiS (approved):
Nice Thanks

richard-ramos (approved):

PR #2551, opened 2024-03-25
fix: remove subscription queue limit

SionoiS:
# Description
I removed the limit on the number of subscription events the queue can handle.

Close https://github.com/waku-org/nwaku/issues/2550


Ivansete-status (approved):
LGTM! Thanks!
As a reference, `0` means "no limit":

https://github.com/status-im/nim-chronos/blob/7b02247ce74d5ad5630013334f2e347680b02f65/chronos/asyncsync.nim#L475

PR #2553, opened 2024-03-25
fix(networkmanager): regularly disconnect from random peers

vpavlin:
# Description
As reported by @chaitanyaprem, NetworkMonitor was stuck at `maxConnections  = 50` and although it discovered new peers, it was not able to connect with internal error `Too many connections` 

This PR bumps macConnections to 150 and adds simple logic to regularly check and disconnect from half of currently connected peers selected randomly.

# Changes

- [x]  Add `MaxConnectedPeers` constant (150) 
- [x] Set `maxConnections` to the constant above
- [x] Disconnect random peers regularly

## How to test

1. `make networkmonitor`
1. `./build/networkmonitor --cluster-id=1 --metrics-server --rln-relay-eth-client-address=https://sepolia.infura.io/v3/...`

### DIFF
--- a/apps/networkmonitor/networkmonitor.nim
+++ b/apps/networkmonitor/networkmonitor.nim
@@ -4,7 +4,7 @@ else:
   {.push raises: [].}
 
 import
-  std/[tables, strutils, times, sequtils],
+  std/[tables, strutils, times, sequtils, random],
   stew/results,
   stew/shims/net,
   chronicles,
@@ -27,6 +27,7 @@ import
   ../../waku/waku_enr,
   ../../waku/waku_discv5,
   ../../waku/waku_dnsdisc,
+  ../../waku/waku_relay,
   ../../waku/waku_rln_relay,
   ../../waku/factory/builder,
   ../wakunode2/networks_config,
@@ -41,6 +42,7 @@ const ReconnectTime = 60
 const MaxConnectionRetries = 5
 const ResetRetriesAfter = 1200
 const AvgPingWindow = 10.0
+const MaxConnectedPeers = 150
 
 const git_version* {.strdefine.} = "n/a"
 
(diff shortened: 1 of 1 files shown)


apps/networkmonitor/networkmonitor.nim
```diff
@@ -128,6 +130,16 @@ proc setConnectedPeersMetrics(
...
 
+  var (inConns, outConns) = node.peer_manager.connectedPeers(WakuRelayCodec)
+  info "connected peers", inConns=inConns.len, outConns=outConns.len
+
+  shuffle(outConns)
+
+  if outConns.len >= toInt(MaxConnectedPeers/2):
+    for p in outConns[0 ..< toInt(outConns.len/2)]:
+      trace "Pruning Peer", Peer = $p
+      asyncSpawn(node.switch.disconnect(p))
```
Ivansete-status, line 141: Out of curiosity, why disconnect only `out` and not `in` too?

Ivansete-status (approved):
LGTM! Just added a naïve question :)

apps/networkmonitor/networkmonitor.nim
```diff
@@ -429,7 +441,8 @@ proc initAndStartApp(
 
   nodeBuilder.withNodeKey(key)
   nodeBuilder.withRecord(record)
-  nodeBuilder.withPeerManagerConfig(maxRelayPeers = none(int), shardAware = true)
+  nodeBUilder.withSwitchConfiguration(maxConnections = some(MaxConnectedPeers))
+  nodeBuilder.withPeerManagerConfig(maxRelayPeers = some(20), shardAware = true)
```
SionoiS, line 445: The shard aware peer manager will try to keep each shard populated with the same average # of peers.
It prevent too many in connections and also keep a % of max connection for service peers.
Are you sure that's what you want in this case?

SionoiS (approved):

apps/networkmonitor/networkmonitor.nim
vpavlin: Good question, I am not sure - what do you think?:) I probably want same average # of peers per shard. I don't really need any in connections since the main goal of NM is to discover nodes and keep track of the network, but not really to be discoverable by others, so limiting in connections might be useful? 

I'll probably metge this PR and we can think about it for next one

apps/networkmonitor/networkmonitor.nim
vpavlin: No particular reason, it is mostly NM discovering and connecting to other nodes, so the `out` is more important, but it is a good point to also make sure `in` is not filling up too much

PR #2554, opened 2024-03-25
chore: android support

richard-ramos:
# Description
Adds android support

To test this:
```
export ANDROID_NDK_HOME=/path/to/Android/ndk/home
make libwaku-android
```

Architecures
- [X] amd64  (used in simulaor)
- [x] armeabi-v7a 
- [x] arm64-v8a
- [x] x86

# Changes
- [X] Update Makefile and waku.nimble
- [x] Crosscompile zerokit (TODO: copy scripts from go-zerokit-rln-*)
- [x] Fix `unhandled exception: value out of range` in `getNanosecondTime` for 32b architecures(arm and x86)
- [x] libnatpmp only compatible witth x86_64
- [x] libminiupnpc only compatible witth x86_64
- [ ] libbacktrace only compatible with x86_64
- [ ] Test makefile on macosx
- [x] Test against HelloWorld app in https://github.com/richard-ramos/gonimffi to make sure we're linking against required android libs
- [ ] Setup CI (?)
- ~~Use static linking for zerokit RLN (might not be a good idea since the resulting lib might be too large?)~~. Decided against it. Resulting libraries are smalller this way and in the case of Android it's suggested that the app size should be < 100MB


richard-ramos:
I ended up disabling linking against libbacktrace when building for android. 
@Ivansete-status, @arnetheduck:
How important is it to be able to crosscompile this library?

richard-ramos:
While attempting to compile for i386, I ran into this error:
```
ld: error: relocation R_386_PC32 cannot be used against symbol 'sendnatpmprequest'; recompile with -fPIC
>>> defined in /home/richard/waku-org/nwaku/vendor/nim-nat-traversal/vendor/libnatpmp-upstream/libnatpmp.a(natpmp.o)
>>> referenced by natpmp.c
>>>               natpmp.o:(sendpublicaddressrequest) in archive /home/richard/waku-org/nwaku/vendor/nim-nat-traversal/vendor/libnatpmp-upstream/libnatpmp.a
```

arnetheduck:
> While attempting to compile for i386, I ran into this error:

I've run into this before - the solution is to add `-fPIC` to the CFLAGS of both libnatpmp and `libminiupnpc` - this should be done for all platforms, not just i386 (it can happen on some linux distros that disable `-fPIC` for some platforms.

A static library (`.a`) is simply a collection of object files - if those object files are going to be used to create a shared library (.so), they need to be compiled with `-fPIC`.

Currently, both natpmp and miniupnp are built using their native build systems (cmake/make) - we've considered switching both so that nim builds them instead which "automatically" would take care of this problem and get rid of the need for cmake / make / etc in the build process significantly simplifying the whole setup - both for android and everyone else.

This is something we do for practically all other libraries, ie secp, bearssl and so on.

Ivansete-status:
> How important is it to be able to crosscompile this library?

I think this is interesting for debugging purposes. Maybe we could check what happens (which error appears) by forcing an unmanaged exception somewhere in the code

richard-ramos:
Just tested this in the example mobile app i created the other day, and was able to link succesfully against libwaku + librln in android! :) , I just had to do some minimal changes in the example to remove calls to libhello, and use instead 
```
            System.loadLibrary("rln")
            System.loadLibrary("waku")
```
Originally It failed to load the libraries because i forgot to link against `-llog ` and `-lrln` so it was complaining about missing functions, but once that was fixed, the app opened, indicating that the shared libs were loaded succesfully.

Now the next steps would be to figure out how to call functions from these libraries, probably by using the work @Ivansete-status has done for the go bindings, by adding a gomobile step, since it would be similar to what we would do in status-go... or perhaps we should use jni? :thinking:.

library/libwaku.nim
```diff
@@ -2,7 +2,10 @@
 {.pragma: callback, cdecl, raises: [], gcsafe.}
 {.passc: "-fPIC".}
 
-import std/[json, sequtils, times, strformat, options, atomics, strutils]
+if defined(linux):
+  {.passl: "-Wl,-soname,libwaku.so"}
```
richard-ramos, line 6: Required so the dynamic linker/loader can resolve symbols at runtime.

library/libwaku.nim
```diff
@@ -39,6 +42,17 @@ const RET_MISSING_CALLBACK: cint = 2
 ################################################################################
 ### Not-exported components
 
+
+template foreignThreadGc(body: untyped) =
```
richard-ramos, line 46: I use this template when executing the callbacks instead of having a single call. We need to ensure that Nim's GC can properly manage memory allocated by Nim code that might be running in other language threads.

library/libwaku.nim
```diff
@@ -54,47 +68,80 @@ proc relayEventCallback(ctx: ptr Context): WakuRelayHandler =
...
+
+# To control when the library has been initialized
+var initialized: Atomic[bool]
+
+if defined(android):
+  # Redirect chronicles to Android System logs
+  when compiles(defaultChroniclesStream.outputs[0].writer):
+    defaultChroniclesStream.outputs[0].writer =
+      proc (logLevel: LogLevel, msg: LogOutputStr) {.raises: [].} =
+        echo logLevel, msg
```
richard-ramos, line 100: Without this, the logs for nwaku can't be seen unless we redirect somehow stdio to adb logcat

waku/waku_core/time.nim
```diff
@@ -7,8 +7,12 @@ import std/times, metrics
...
 
-proc getNanosecondTime*(timeInSeconds: int64 | float64): Timestamp =
-  let ns = Timestamp(timeInSeconds * 1_000_000_000)
+proc getNanosecondTime*(timeInSeconds: int64): Timestamp =
+  let ns = Timestamp(timeInSeconds * int64(1_000_000_000))
+  return ns
+
+proc getNanosecondTime*(timeInSeconds: float64): Timestamp =
+  let ns = Timestamp(timeInSeconds * float64(1_000_000_000))
   return ns
```
richard-ramos, line 16: For some reason `proc getNanosecondTime*(timeInSeconds: int64 | float64): Timestamp` didn't work in Android x86. I think in practice these two functions I defined here have the same behavior :shrug:

library/libwaku.nim
```diff
@@ -54,47 +68,80 @@ proc relayEventCallback(ctx: ptr Context): WakuRelayHandler =
...
+proc waku_setup() {.dynlib, exportc.} =
+  NimMain()
+  if not initialized.load:
+    initialized.store(true)
+
+    # TODO: ask Ivan what is nimGC_setStackBottom for
+    when declared(nimGC_setStackBottom):
+      var locals {.volatile, noinit.}: pointer
+      locals = addr(locals)
+      nimGC_setStackBottom(locals)
```
richard-ramos, line 117: I had to introduce this function `waku_setup` in the bindings.

When I tried to use the original `waku_init` function, Android crashed. Looks like NimMain() (at least on android) needs to be called before any nim logic is executed, because the code `if not initialized.exchange(true):` that existed before did fail. I imagine it might be related to the `initialized` atomic being setup by `NimMain()`? 

Anyway, I think I heard/read somewhere that `NimMain()` function is not required in Nim 2.0? so perhaps on that version we can get rid of the existence of `waku_setup` then.

Also, I couldn't find docs for `nimGC_setStackBottom`. Is it really necessary?

richard-ramos:
I think this is ready for review. (The PR is big enough as it is right now).
In separate PRs I'll do the following:
1. Add a default react native app (unmodified. So should be a simple PR)
2. Modify the default react native app to use libwaku
3. Compile libbacktrace for all the android archs
4. Setup CI?

scripts/build_rln_android.sh
```diff
@@ -0,0 +1,26 @@
...
+
+[[ -z "${nwaku_build_dir}" ]] && { echo "No nwaku build directory specified"; exit 1; }
+[[ -z "${zerokit_dir}" ]]     && { echo "No zerokit directory specified";     exit 1; }
+[[ -z "${rln_version}" ]]     && { echo "No rln version specified";           exit 1; }
+[[ -z "${android_arch}" ]]    && { echo "No android architecture specified";  exit 1; }
+[[ -z "${abi}" ]]             && { echo "No abi specified";                   exit 1; }
+
+export RUSTFLAGS="-Ccodegen-units=1"
+
+cargo install cross --git https://github.com/cross-rs/cross
```
gabrielmer, line 17: I initially got the error: 
```
cannot install package `cross 0.2.5`, it requires rustc 1.77.2 or newer, while the currently active rustc version is 1.75.0
```

I see that rustc 1.77.2 is very recent (2024-04-09). 

Maybe we should add to the script to install that specific version, or at least ```rustup upgrade```. For sure safer to do it with a specific version though.

richard-ramos: Good idea!

Makefile
```diff
@@ -290,6 +290,64 @@ else
...
+				libwaku-android-precheck \
+				libwaku-android-arm64 \
+				libwaku-android-amd64 \
+				libwaku-android-x86 \
+				libwaku-android-arm \
+				rebuild-nat-libs \
+				build-libwaku-for-android-arch
+
+ANDROID_TARGET ?= 30
+ANDROID_TOOLCHAIN_DIR ?= $(ANDROID_NDK_HOME)/toolchains/llvm/prebuilt/linux-x86_64
```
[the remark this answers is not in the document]
richard-ramos: Yes, I'll do it in a separate PR against this branch, as I haven't tried to build this PR yet in macos. I'm not sure if there are other things that need to be changed besides this line!

Makefile
```diff
@@ -131,7 +131,7 @@ clean: | clean-libbacktrace
 ##################
 ##     RLN      ##
 ##################
-.PHONY: librln
+.PHONY: librln-env librln shouldUseRLNV2
```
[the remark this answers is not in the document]
richard-ramos: that target i had added it when i started working on this PR but it is not necessary! i'll remove it

arnetheduck:
this is _really nice_ - does it make sense for any of it be ported back to nimbus-build-system or included in https://status-im.github.io/nim-style-guide/tooling.html ?

gabrielmer (approved):
Went over it again and looks amazing :))

Thanks so much!

richard-ramos:
> this is really nice - does it make sense for any of it be ported back to nimbus-build-system or included in https://status-im.github.io/nim-style-guide/tooling.html ?

In tooling.html might be a good idea to document the process followed to cross compile android using the ndk toolchain. 

For porting back to nimbus-build-system, this part I'm not sure and would likely need your review to understand what can be extracted from waku repo to nimbus-build-system. The changes required for getting nwaku to run in android were few and limited to the Makefile and a config.nims file, as well as having a separate script for cross compiling RLN

I think a good candidate could be to improve the build process for the nat-libs, to not have to require this target:
```
rebuild-nat-libs: | clean-cross nat-libs
```
Which i had to add to delete the nat libs before building nwaku for android in the diff architectures, I'm thinking it should be possible to be able to specify the output dir of the libs somehow?

Ivansete-status (commented):
This is amazing :heart: !

I just added some comments that I hope you find useful.

Not approving yet because few points to clarify:
1. is it needed the change in `nim-chronos` and `nimbus-build-system`?
2. We need to adapt the other examples so that the `waku_setup` is called at first. Or instead, try to avoid `waku_setup`. I will double-check if it is feasible to do so.

Congrats again for such wonderful PR :partying_face:

richard-ramos:
> @Ivansete-status: is it needed the change in nim-chronos and nimbus-build-system?

Yes, it's because of some issues I ran into that were fixed in specific PRs (merged already)
- nim-chronos
  - https://github.com/status-im/nim-chronos/pull/538
- nimbus-build-system
  - https://github.com/status-im/nimbus-build-system/pull/77
  -  https://github.com/status-im/nimbus-build-system/pull/81

examples/mobile/android/app/src/debug/AndroidManifest.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:tools="http://schemas.android.com/tools">
+
+    <application
+        android:usesCleartextTraffic="true"
+        tools:targetApi="28"
```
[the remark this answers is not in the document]
richard-ramos: Ah, good catch! i have no idea to be honest. Will investigate what's that attribute used for

examples/mobile/ios/mobile/AppDelegate.h
```diff
@@ -0,0 +1,6 @@
+#import <RCTAppDelegate.h>
```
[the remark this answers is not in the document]
richard-ramos: This is because the default react native app creates also some code for IOS. No work has been done for libwaku on ios yet, but I think it's safe to leave this example code for now, since it will be the base for the IOS integration, even if we do not support it yet.

(Also because I fear breaking react native if this code does not exist :) )

examples/mobile/android/app/src/main/jni/waku_ffi.c
```diff
@@ -0,0 +1,316 @@
...
+void on_response(int ret, const char *msg, size_t len, void *user_data) {
+  if (ret != 0) {
+    char errMsg[300];
+    snprintf(errMsg, 300, "function execution failed. Returned code: %d, %s\n", ret, msg);
+    if (user_data != NULL) {
+      cb_result **data_ref = (cb_result **)user_data;
+      (*data_ref) = malloc(sizeof(cb_result));
+      (*data_ref)->error = true;
+      (*data_ref)->message = malloc(len * sizeof(char) + 1);
+      strcpy((*data_ref)->message, msg);
```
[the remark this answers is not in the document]
richard-ramos: Very cool. I had no idea of the existence of this function!

examples/mobile/android/app/src/debug/AndroidManifest.xml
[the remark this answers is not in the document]
richard-ramos: According to the docs:

> This tells the tools that you believe this element and any children are used only on the specified API level or higher. This stops lint from warning you if that element or its attributes are not available on the API level you specify as your minSdkVersion.

I'll set it to the same android_target value in the meantime. I think it's okay, i did not see any negative effect due to this change

richard-ramos:
@Ivansete-status I implemented all the suggested changes except by calling the `waku_setup` function in the examples. Let me know if you find a way to avoid having to call that function, otherwise, i'll do the changes on all the examples.

Ivansete-status:
> @Ivansete-status I implemented all the suggested changes except by calling the `waku_setup` function in the examples. Let me know if you find a way to avoid having to call that function, otherwise, i'll do the changes on all the examples.

Hey @richard-ramos ! I've been playing and learning from what you've done :)

I haven't found a way to avoid having the `waku_setup` proc. As you already mentioned, it always crashes with messages like the following: `05-16 11:46:09.675 24060 24697 F libc    : Fatal signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x80 in tid 24697 (mqt_native_modu), pid 24060 (com.mobile)`

Nevertheless, I think is fine to have the `waku_setup` proc and make the library user responsible for invoking it only once (as you implemented in the ReactNative example.)

Let me know if I can help to adapt the examples and include that new `waku_setup` proc.

Besides, it would be useful to define the following macros at the top of `waku_ffi.c` so that anyone could add some logs:
```code
#define APPNAME "waku-jni"
#define LOGD(TAG) __android_log_print(ANDROID_LOG_DEBUG , APPNAME,TAG);
```
Then, the following can be added somewhere: `LOGD("log example for debugging purposes...")`

richard-ramos:
@Ivansete-status I updated all the examples except the nodejs and python. Maybe later during the week we can pair program to add the waku_setup call? in the meantime, as soon as i get green lights i'll merge this PR since it's been open for a while and it's mostly complete :)

Ivansete-status:
> @Ivansete-status I updated all the examples except the nodejs and python. Maybe later during the week we can pair program to add the waku_setup call? in the meantime, as soon as i get green lights i'll merge this PR since it's been open for a while and it's mostly complete :)

Yes go ahead thanks!
We will update the examples next week ;) Feel free to send me an invite for next Thursday, the 30th, for example 

Cheers

PR #2559, opened 2024-03-26
feat: examples/golang/waku.go add new example

Ivansete-status:
## Description
This PR adds a simple example and approach on how to perform cbindings over Golang. In other words, this PR allows the integration of `libwaku.so` in a tiny example to demonstrate how to use it.

As a pre-requisite, the `libwaku.so` must exist. To compile it, `make libwaku -j11`.

## Changes

- [x] Add simple Golang example that integrates `libwaku.so`. This example just starts a node and makes it to connect to an already running local node.

### DIFF
--- a/examples/golang/waku.go
+++ b/examples/golang/waku.go
@@ -7,7 +7,7 @@ package main
 	#include <stdio.h>
 	#include <stdlib.h>
 
-	extern void MyEventCallback(int ret, char* msg, size_t len, void* userData);
+	extern void globalEventCallback(int ret, char* msg, size_t len, void* userData);
 
 	typedef struct {
 		int ret;
@@ -89,10 +89,20 @@ package main
 		WAKU_CALL(waku_version(ctx, (WakuCallBack) callback, resp));
 	}
 
-	void cGoWakuSetEventCallback(void* ctx, void* resp) {
-		// We should pass a resp != NULL in this case to overcome
-		// the 'if isNil(ctx[].eventUserData)' check in libwaku.nim
-		waku_set_event_callback(ctx, (WakuCallBack) MyEventCallback, resp);
+	void cGoWakuSetEventCallback(void* ctx) {
+		// The 'globalEventCallback' Go function is shared amongst all possible WakuNode instances.
+
+		// Given that the 'globalEventCallback' is shared, we pass again the
+		// ctx instance but in this case is needed to pick up the correct method
+		// that will handle the event.
+
+		// In other words, for every call the libwaku makes to globalEventCallback,
+		// the 'userData' parameter will bring the context of the node that registered
+		// that globalEventCallback.
+
+		// This technique is needed because cgo only allows to export Go functions and not methods.
+
+		waku_set_event_callback(ctx, (WakuCallBack) globalEventCallback, ctx);
 	}
 
 	void cGoWakuContentTopic(void* ctx,
@@ -262,23 +272,21 @@ func (self *WakuNode) WakuVersion() (string, error) {
 	return "", errors.New(errMsg)
 }
 
-//export MyEventCallback
-func MyEventCallback(callerRet C.int, msg *C.char, len C.size_t, userData unsafe.Pointer) {
+//export globalEventCallback
+func globalEventCallback(callerRet C.int, msg *C.char, len C.size_t, userData unsafe.Pointer) {
+	// This is shared among all Golang instances
+
+	self := WakuNode{ctx: userData}
+	self.MyEventCallback(callerRet, msg, len)
+}
+
+func (self *WakuNode) MyEventCallback(callerRet C.int, msg *C.char, len C.size_t) {
 	fmt.Println("Event received:", C.GoStringN(msg, C.int(len)))
 }
 
-func (self *WakuNode) WakuSetEventCallback() error {
-	var resp = C.allocResp()
-	// Notice that we are not releasing the `resp` memory in this case
-	// because we leave it available to all the upcoming events
-	C.cGoWakuSetEventCallback(self.ctx, resp)
-
-	if C.getRet(resp) == C.RET_OK {
-		return nil
-	}
-
-	return errors.New("error WakuSetEventCallback: " +
-		C.GoStringN(C.getMyCharPtr(resp), C.int(C.getMyCharLen(resp))))
+func (self *WakuNode) WakuSetEventCallback() {
+	// Notice that the events for self node are handled by the 'MyEventCallback' method
+	C.cGoWakuSetEventCallback(self.ctx)
 }
 
 func (self *WakuNode) FormatContentTopic(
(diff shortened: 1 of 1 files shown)


examples/golang/waku.go
```diff
@@ -0,0 +1,329 @@
...
+func WakuListenAddresses(ctx unsafe.Pointer) string {
+	var str = C.allocMyString()
+	defer C.freeMyString(str)
+	C.cGoWakuListenAddresses(ctx, str)
+	var listenAddresses = C.GoStringN(C.getMyCharPtr(str), C.int(C.getMyCharLen(str)))
+    return listenAddresses
+}
+
+func main() {
+	runtime.LockOSThread()
```
chaitanyaprem, line 299: Is there any thread-local state that is being used in the underlying c bindings implementation? 
Otherwise this would not be required.

examples/golang/waku.go
```diff
@@ -0,0 +1,329 @@
...
+	config := `{
+		"host": "0.0.0.0",
+		"port": 30304,
+		"key": "11d0dcea28e86f81937a3bd1163473c7fbc0a0db54fd72914849bc47bdf78710",
+		"relay": true
+	}`
+
+	ctx := WakuNew(config)
+	WakuSetEventCallback(ctx)
+	WakuRelaySubscribe(ctx, WakuDefaultPubsubTopic(ctx))
```
chaitanyaprem, line 310: Maybe best to use sharding  (static/auto based on if RLN is available) as an example rather than defaultpubsubTopic?

chaitanyaprem (approved):
LGTM

Have few queries, nothing blocker.

examples/golang/waku.go
Ivansete-status: > Is there any thread-local state that is being used in the underlying c bindings implementation? Otherwise this would not be required.

Thanks for that! I got it from the previous approach. I addressed it in https://github.com/waku-org/nwaku/pull/2559/commits/c9319c59c0d84b45e366b7567b77a28267518c0b

examples/golang/waku.go
Ivansete-status: > Maybe best to use sharding (static/auto based on if RLN is available) as an example rather than defaultpubsubTopic?

ah yes, good point! I think for now we can keep this example simpler but is something to consider if we create a Golang "bindings" repo and extend the `libwaku` functions.

examples/golang/waku.go
```diff
@@ -0,0 +1,326 @@
...
+	var str = C.allocMyString()
+
+	defer C.free(unsafe.Pointer(cTopicName))
+	defer C.freeMyString(str)
+
+	C.cGoWakuPubsubTopic(ctx, cTopicName, str)
+
+	var pubsubTopic = C.GoStringN(C.getMyCharPtr(str), C.int(C.getMyCharLen(str)))
+    return pubsubTopic
+}
```
SionoiS, line 247: What does this do? Topic string into the shard format?

SionoiS (approved):
Never read Go before but looks good.

richard-ramos (approved):
Looks nice!
On the waku-go-bindings repository we can try to add more goodies to the API that we can copy from go-waku and would probably make things easier for the dev, as well as making some functions friendlier. Here are some things that come to mind. They can be done either in the new repo or here.

1. Instead of panicing with an exit(1) in WAKU_CALL, we should return a go `error` in the Go API for all functions that might fail.

2. Instead of returning an `unsafe.Pointer` in `WakuNew`, we can define the following:
```go
type WakuContext {
     ptr unsafe.Pointer
}

func WakuNew(jsonConfig string) (*WakuContext, error) {
	// TODO: error handling in case waku_new fails, instead of exit
       ptr := C.cGoWakuNew(C.CString(jsonConfig))
       return &WakuContext{ ptr: ptr }, nil
}
```
for all means and purposes, it's basically the same code, but looks less scary to devs than returning an unsafe.Pointer

3. It seems that setting an event callback requires the dev to also define a `void cGoWakuSetEventCallback(void* ctx, void* myStr)` on the C side? If that's the case, it will probably be necessary to add comments indicating the need to do so.

4. In WakuContentTopic it probably makes sense to have the content topic be its own alias instead of string i.e.:  `type WakuContentTopic = string`, same for the pubsub topic

5. There's something strange going on with the indentation. I guess it's an IDE issue? If using vscode, consider installing the go extension, which will automatically format your code.

6. It might make sense to instead of having separate functions, to have a single struct with methods, i.e., following the previous recommendation, you could have something like this:
```go
package waku

type WakuNode {
     ptr unsafe.Pointer
}

func(w *WakuNode) Start() error {  ....  }
func(w *WakuNode) Stop() error {  ....  }

func New(....) (*WakuNode, error) {  }
...
```
That way, if you want to create a node in some project, you would do something like this:
```go
import waku "github.com/waku-org/waku-go-bindings"
...
wakuNode, err := waku.New(theConfig)
if err != nil {
    // TODO: Handle error
}

wakuNode.Start()

...

wakuNode.Stop()
```

7. In `WakuNew` function, instead of accepting a json, we should accept a struct, and do a json.Marshal to convert the struct to a string. That way we leave less oportunities for the devs to make mistakes building the config. This can be copied from https://github.com/waku-org/go-waku/blob/master/library/config.go#L12

8. `WakuRelayPublish` should return the message hash according to the RFC

9. It probably make sense for `WakuListenAddresses` to return a slice of multiaddresses (using https://github.com/multiformats/go-multiaddr) instead of returning the JSON response

examples/golang/waku.go
```diff
@@ -0,0 +1,326 @@
...
+	fmt.Println("Custom content topic:", WakuContentTopic(ctx, "appName", 1, "cTopicName", "enc"))
+	fmt.Println("Custom pubsub topic:", WakuPubsubTopic(ctx, "my-ctopic"))
+	fmt.Println("Default pubsub topic:", WakuDefaultPubsubTopic(ctx))
+	fmt.Println("Listen addresses:", WakuListenAddresses(ctx))
+	
+	for {
+		// A simple wait to let the Waku node to run.
+		// Notice that the Waku node runs in a separate thread.
+		time.Sleep(time.Second)
+	}
```
richard-ramos, line 324:
```suggestion
        // Wait for a SIGINT or SIGTERM signal
        ch := make(chan os.Signal, 1)
        signal.Notify(ch, syscall.SIGINT, syscall.SIGTERM)
        <-ch
```

examples/golang/waku.go
```diff
@@ -0,0 +1,326 @@
...
+}
```
richard-ramos, line 325: Even if the node will stop anyway once the program exits, it is probably a good idea to manually stop the node, so devs see that the function exists!

```suggestion
  WakuStop()
}
```

examples/golang/waku.go
```diff
@@ -0,0 +1,326 @@
+package main
+
+/*
```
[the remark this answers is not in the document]
Ivansete-status: > That's only to show the C logic we assume exist right?

I'm not an expert at all but in this case, we are using `cgo`, which conforms a tandem between C and Go. 
These comments are handled by the Go compiler and they are part of the C-side.

examples/golang/waku.go
Ivansete-status: > What does this do? Topic string into the shard format?

Yes, this function is aimed to simplify the creation of a well-formatter pubsubtopic.

Ivansete-status:
@richard-ramos - I've applied all your comments but the 9.
I'll merge it for now but I'm happy to address any further suggestions in upcoming PRs :)

Thanks so much for the comments guys!

PR #2561, opened 2024-03-28
feat: sync storage manager

chaitanyaprem:
# Description
Introducing storage manager . 

# Changes

- [x] Have a storage manager that manages storage to be used for sync as per https://github.com/waku-org/research/issues/80#issuecomment-2022107830
- [x] Fix for duplicate invocation of dialPeer due to use of valueOr https://github.com/vacp2p/nim-libp2p/pull/1079
- [x] Refactor to introduce sync session

Implementing sync protocol flow in Sync session as per https://github.com/waku-org/research/issues/80#issuecomment-2018235161 can be taken up in a separate PR. 
cc @SionoiS , I don't think i had time to start on this. Maybe you can take it up from here after merging this PR.


waku/waku_sync/storage.nim
```diff
@@ -0,0 +1,59 @@
...
+
+import std/[times, tables], chronicles, chronos, stew/results
+
+import ./raw_bindings
+
+logScope:
+  topics = "waku sync"
+
+type WakuSyncStorageManager* = object
+  storages: OrderedTable[string, Storage]  # Map of dateTime and Storage objects. DateTime is of the format YYYYMMDDHH
```
SionoiS, line 14: string are unwieldy lets use a timestamp instead.

waku/waku_sync/storage.nim
```diff
@@ -0,0 +1,59 @@
...
+
+#Time should be in YYYYMMDDHH format.
+proc createStorage(
+    self: WakuSyncStorageManager, time: string
+): Result[Storage, string] =
+  let storage:Storage = Storage.new().valueOr:
+    error "storage creation failed"
+    return err("storage creation failed")
+  self.storages[time] = storage
+  return storage
```
SionoiS, line 28: we need to check if the new overlap with an older one.

waku/waku_sync/storage.nim
```diff
@@ -0,0 +1,59 @@
...
+
+proc getRecentStorage(): Result[Storage, string]=
+
+
+#Time should be in YYYYMMDDHH format.
+proc getStorage(self: WakuSyncStorageManager, time: string): Result[Storage, string] =
+    if self.storages.hasKey(time):
+        return self.storages[time]
+  return err("no storage found for given dateTime")
```
SionoiS, line 37: That would be an option `none` not an error in this case.

waku/waku_sync/storage.nim
```diff
@@ -0,0 +1,59 @@
...
+
+proc retrieveStorage*(
+    self: WakuSyncStorageManager, time: int64
+): Result[Storage, string] =
```
SionoiS, line 41: same as above I would use a `Result[Option[T], string]` instead.

waku/waku_sync/storage.nim
```diff
@@ -0,0 +1,59 @@
...
+  let unixf = times.fromUnixFloat(float(time))
+
+  let dateTime = times.format(unixf, "YYYYMMDDHH")
```
SionoiS, line 44: Lets do the opposite. Use a timestamp internally but allow fetching by date & time formatted string.

SionoiS (commented):
General direction looks good but I hate strings...

See my other style comments.

waku/waku_sync/storage.nim
chaitanyaprem: Agreed, Will figure out how to use timestamp with only till hour format.

waku/waku_sync/storage.nim
chaitanyaprem: I kind of did that in the caller of this ` proc retrieveStorage*( `and hence this is not made public. Was trying to modularize too much i think and maybe this proc is not even required. Will refactor.

waku/waku_sync/storage.nim
chaitanyaprem: I was able to retrieve storage by using timestamp. I don't think we need to expose any other way and keep the string part hidden within. I did not want to use timestamp as the granularity with which we want to maintain storage is in terms of hours and timestamp is in seconds.

waku/waku_sync/protocol.nim
```diff
@@ -20,98 +20,68 @@ import
...
     callback: Option[WakuSyncCallback]
     periodicSyncFut: Future[void]
 
 proc ingessMessage*(self: WakuSync, pubsubTopic: PubsubTopic, msg: WakuMessage) =
   if msg.ephemeral:
     return
-
+  #TODO: Do we need to check if this message has already been ingessed? 
+  # because what if messages is received via gossip and sync as well?
+  # Might 2 entries to be inserted into storage which is inefficient.
```
SionoiS, line 51: The system should be idempotent, so i would say yes. Is the negentropy storage insert func not idempotent?

waku/waku_sync/protocol.nim
```diff
@@ -20,98 +20,68 @@ import
...
-  return ok(needHashes)
+  return ok(hashes)
 
 proc sync*(
     self: WakuSync
 ): Future[Result[(seq[WakuMessageHash], RemotePeerInfo), string]] {.async, gcsafe.} =
   let peer: RemotePeerInfo = self.peerManager.selectPeer(WakuSyncCodec).valueOr:
     return err("No suitable peer found for sync")
-
-  let conn: Connection = (await self.peerManager.dialPeer(peer, WakuSyncCodec)).valueOr:
```
SionoiS, line 114: Is it not more concise this way?

waku/waku_sync/protocol.nim
```diff
@@ -209,11 +157,13 @@ proc periodicSync(self: WakuSync) {.async.} =
...
+  if self.syncInterval > ZeroDuration:
+    # start periodic-sync only if interval is set.
     self.periodicSyncFut = self.periodicSync()
 
 proc stopWait*(self: WakuSync) {.async.} =
   await self.periodicSyncFut.cancelAndWait()
 
-proc storageSize*(self: WakuSync): int =
-  return self.storage.size()
+#[ TODO:Fetch from storageManager??
```
SionoiS, line 167: Yes, it would make sense to me.

waku/waku_sync/session.nim
```diff
@@ -0,0 +1,142 @@
...
+import std/options, stew/results, chronicles, chronos, libp2p/stream/connection
+
+import ../common/nimchronos, ../waku_core, ./raw_bindings, ./storage_manager
+
+logScope:
+  topics = "waku sync"
+
+type SyncSessionType* = enum
+  CLIENT = 1
+  SERVER = 2
```
SionoiS, line 15: Do we expect more session type in the future?

waku/waku_sync/session.nim
```diff
@@ -0,0 +1,142 @@
...
+  rangeStart*: int64
+  rangeEnd*: int64
+  negentropy*: Negentropy
+
+#[
+    Session State Machine
+    1. negotiate sync params
+    2. start negentropy sync
+    3. find out local needhashes
+    4. If client, share peer's needhashes to peer
```
SionoiS, line 35: Love state machine!

waku/waku_sync/session.nim
```diff
@@ -0,0 +1,142 @@
...
+    if writeRes.isErr():
+      return err(writeRes.error.msg)
+
+  return ok(needHashes)
+
+proc HandleServerSession*(
+    self: SyncSession, conn: Connection, storageMgr: WakuSyncStorageManager
+) {.async, gcsafe.} =
+  #TODO: Find matching storage based on sync range and continue??
+  #TODO: Return error rather than closing stream abruptly?
```
SionoiS, line 112: Adding standard error code and desc. to the payload would be the way I think

SionoiS (approved):
good direction!

waku/waku_sync/protocol.nim
chaitanyaprem: it is concise but leading to a bug where the handler gets invoked twice. i referred the issue in the pr description.

waku/waku_sync/protocol.nim
chaitanyaprem: Negentropy doesn't seem to be, because I wrote a small test to verify and it shows as a diff. Hence this needs to be handled separately at higher layer.

waku/waku_sync/session.nim
chaitanyaprem: I don't think so

PR #2562, opened 2024-04-02
fix: cluster-id 0 discv5 issue

alrevuelta:
# Description
* [This PR](https://github.com/waku-org/nwaku/pull/2556/files#diff-43f34e84eeb9a4b4061898c3bd78274e149e50bd6b5d24aeb0a982e07629172eL300) deprecated `topics` but introduced a regression.
* Since `topics` flag default value `/waku/2/default-waku/proto` was being used when constructing the ENR, after removing that flag, we no longer had a default value.
* This means that eg `wakunode2 --cluster-id=0 ...` discv5 was not working properly, since the node had no default pubsub topic configured.
* This PR fixes it, by setting the default pubsub topic for `cluster-id=0`.
* Other unknown clusters, would need to provide their desired `pubsub-topic`.

### DIFF
--- a/apps/wakunode2/wakunode2.nim
+++ b/apps/wakunode2/wakunode2.nim
@@ -87,26 +87,34 @@ when isMainModule:
   of inspectRlnDb:
     doInspectRlnDb(conf)
   of noCommand:
-    # The Waku Network config (cluster-id=1)
-    if conf.clusterId == 1:
-      let twnClusterConf = ClusterConf.TheWakuNetworkConf()
-      if len(conf.shards) != 0:
-        conf.pubsubTopics = conf.shards.mapIt(twnClusterConf.pubsubTopics[it.uint16])
-      else:
-        conf.pubsubTopics = twnClusterConf.pubsubTopics
+    case conf.clusterId
+      # cluster-id=0
+      of 0:
+        let clusterZeroConf = ClusterConf.ClusterZeroConf()
+        conf.pubsubTopics = clusterZeroConf.pubsubTopics
+        # TODO: Write some template to "merge" the configs
+      # cluster-id=1 (aka The Waku Network)
+      of 1:
+        let twnClusterConf = ClusterConf.TheWakuNetworkConf()
+        if len(conf.shards) != 0:
+          conf.pubsubTopics = conf.shards.mapIt(twnClusterConf.pubsubTopics[it.uint16])
+        else:
+          conf.pubsubTopics = twnClusterConf.pubsubTopics
 
-      # Override configuration
-      conf.maxMessageSize = twnClusterConf.maxMessageSize
-      conf.clusterId = twnClusterConf.clusterId
-      conf.rlnRelay = twnClusterConf.rlnRelay
-      conf.rlnRelayEthContractAddress = twnClusterConf.rlnRelayEthContractAddress
-      conf.rlnRelayDynamic = twnClusterConf.rlnRelayDynamic
-      conf.rlnRelayBandwidthThreshold = twnClusterConf.rlnRelayBandwidthThreshold
-      conf.discv5Discovery = twnClusterConf.discv5Discovery
-      conf.discv5BootstrapNodes =
-        conf.discv5BootstrapNodes & twnClusterConf.discv5BootstrapNodes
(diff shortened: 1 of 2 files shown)


apps/wakunode2/wakunode2.nim
```diff
@@ -87,26 +87,34 @@ when isMainModule:
...
-      let twnClusterConf = ClusterConf.TheWakuNetworkConf()
-      if len(conf.shards) != 0:
-        conf.pubsubTopics = conf.shards.mapIt(twnClusterConf.pubsubTopics[it.uint16])
+    case conf.clusterId
+      # cluster-id=0
+      of 0:
+        let clusterZeroConf = ClusterConf.ClusterZeroConf()
+        conf.pubsubTopics = clusterZeroConf.pubsubTopics
+        # TODO: Write some template to "merge" the configs
+      # cluster-id=1 (aka The Waku Network)
```
alrevuelta, line 96: no real change from here. just idendation.

apps/wakunode2/wakunode2.nim
```diff
@@ -87,26 +87,34 @@ when isMainModule:
...
-      conf.rlnRelay = twnClusterConf.rlnRelay
-      conf.rlnRelayEthContractAddress = twnClusterConf.rlnRelayEthContractAddress
-      conf.rlnRelayDynamic = twnClusterConf.rlnRelayDynamic
-      conf.rlnRelayBandwidthThreshold = twnClusterConf.rlnRelayBandwidthThreshold
-      conf.discv5Discovery = twnClusterConf.discv5Discovery
-      conf.discv5BootstrapNodes =
-        conf.discv5BootstrapNodes & twnClusterConf.discv5BootstrapNodes
-      conf.rlnEpochSizeSec = twnClusterConf.rlnEpochSizeSec
-      conf.rlnRelayUserMessageLimit = twnClusterConf.rlnRelayUserMessageLimit
+        discard
```
Ivansete-status, line 117: We `discard` at this point because we are not willing to set any particular configuration when either cluster_id == 0 or 1, right?

apps/wakunode2/wakunode2.nim
```diff
@@ -87,26 +87,34 @@ when isMainModule:
...
-    if conf.clusterId == 1:
-      let twnClusterConf = ClusterConf.TheWakuNetworkConf()
-      if len(conf.shards) != 0:
-        conf.pubsubTopics = conf.shards.mapIt(twnClusterConf.pubsubTopics[it.uint16])
+    case conf.clusterId
+      # cluster-id=0
+      of 0:
+        let clusterZeroConf = ClusterConf.ClusterZeroConf()
+        conf.pubsubTopics = clusterZeroConf.pubsubTopics
+        # TODO: Write some template to "merge" the configs
```
Ivansete-status, line 95: What do you mean by "merge the configs" ?

Ivansete-status (approved):
LGTM! Thanks for such a quick fix!

apps/wakunode2/wakunode2.nim
alrevuelta: correct. anyone can define their own clusterId eg =999 and there is no config we set in there.

apps/wakunode2/wakunode2.nim
alrevuelta: imagine
```
type A:
  field1
  field2

typeB:
  field1
  field2
  field3
  field4
```

which is similar to `WakuConf` and `ClusterZeroConf`. Note that `typeB` is a superset of `typeA`.

Right now we are "merging" boh manually.

```
let a = TypeA(...)
let b = TypeB(...)

# we want a config to override b
b.field1 = a.field1
b.field2 = a.field2
```

this is i) error prone, ii) if number of parameters increase, we have to assign them manually.

so I'm suggesting with some templating, writing a `merge()` function so that.
```
let a = TypeA(...)
let b = TypeB(...)

merge(a, b)
```

or something like this. with this we no longer need [all this](https://github.com/waku-org/nwaku/pull/2562/files#diff-fdd56fbeef6ba31430b71ea50a53439fc828b61e1d4291ad18e03906b41c7100R104-R115)

jm-clius (approved):
Thanks!

PR #2563, opened 2024-04-02
fix: regex for rpc endpoint

alrevuelta:
# Description
* The regex matching `--rln-relay-eth-client-address` flag was not allowing some valid RPC urls.
* Example, this should be valid `http://foundry:8545`
* But was failing:
```
Error while processing the rln-relay-eth-client-address=http://foundry:8545 parameter: Invalid HTTP RPC URL
```
* This PR fixes it so that it allows such URLs.

It can be also reproduced with:

```bash
docker run harbor.status.im/wakuorg/nwaku@sha256:d193792202726c51cd656c48f4ffd04f261ec2d8663a00591947a5aab3677dc5 --rln-relay-eth-client-address="http://foundry:3000"

Error while processing the rln-relay-eth-client-address=http://foundry:3000 parameter: Invalid HTTP RPC URL
Try wakunode2 --help for more information
```

### DIFF
--- a/waku/factory/external_config.nim
+++ b/waku/factory/external_config.nim
@@ -680,9 +680,9 @@ proc parseCmdArg*(T: type EthRpcUrl, s: string): T =
   ## disallowed patterns:
   ## any valid/invalid ws or wss url
   var httpPattern =
-    re2"^(https?):\/\/((localhost)|([\w_-]+(?:(?:\.[\w_-]+)+)))(:[0-9]{1,5})?([\w.,@?^=%&:\/~+#-]*[\w@?^=%&\/~+#-])*"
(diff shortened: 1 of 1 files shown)


rymnc (approved):
regex reviews are tricky, but approving so I'm not a blocker.

Ivansete-status (approved):
LGTM! Thanks!
I just checked with another examples:
```
https://www.something.com/
http://www.something.com/
https://www.something.edu.co.in
http://www.url-with-path.com/path
https://www.url-with-querystring.com/?url=has-querystring
http://url-without-www-subdomain.com/
https://mail.google.com
```

alrevuelta:
Thanks, merging since it is blocking, but we need unittests for this kind of thing.

PR #2564, opened 2024-04-02
fix(c-bindings): rln credential path key

richard-ramos:
# Description
The key used to obtain the path to the rln keystore was incorrect


Ivansete-status (approved):
I approve to not block it, but I can't confirm whether the change is correct :P

gabrielmer (approved):

gabrielmer:
> I approve to not block it, but I can't confirm whether the change is correct :P

same 😶

PR #2573, opened 2024-04-08
chore: Extend release process with the need of merging back release branch to master

NagyZoltanPeter:
Just to simple fix of our release notation gone wrong on releases/images made from master, there is a need to merge the release branch - that got tagged with the release tag - back to master. 
This helps git to find the last relevant tag on the path for versioning the build.

### DIFF
--- a/docs/contributors/release-process.md
+++ b/docs/contributors/release-process.md
@@ -57,11 +57,15 @@ Ensure all items in this list are ticked:
     ```
 
 1. Once the release-candidate has been validated, create a final release tag and push it.
+We also need to merge release branch back to master as a final step.
 
     ```
     git checkout release/v0.1.0
     git tag -as v0.1.0 -m "Initial release."
     git push origin v0.1.0
+    git switch master
+    git pull
+    git merge release/v0.1.0
     ```
 
 1. Create a [Github release](https://github.com/waku-org/nwaku/releases) from the release tag.


gabrielmer:
Not sure I understand, why should we merge the release branch to master? I guess we should have the tags in master and then branch out to the release branch, not the other way around. But maybe I'm missing something

NagyZoltanPeter:
> Not sure I understand, why should we merge the release branch to master? I guess we should have the tags in master and then branch out to the release branch, not the other way around. But maybe I'm missing something

Unfortunately without the merge ```git describe --abbrev=6 --always --tags``` command which used to provide version info into the build will find any former 'latest' tag accessible from master, thus leads to misleading versioning.
Now our images made from latest master has v0.25.0-rc.0-1234567 version instead of v0.26.0 as that is our latest release.
Merging back solves this.

gabrielmer:
> Unfortunately without the merge `git describe --abbrev=6 --always --tags` command which used to provide version info into the build will find any former 'latest' tag accessible from master, thus leads to misleading versioning. Now our images made from latest master has v0.25.0-rc.0-1234567 version instead of v0.26.0 as that is our latest release. Merging back solves this.

But I think that the problem here is that the tag wasn't made on master for 0.26.0. If we would have tagged 0.26.0 on master and then forked it to a release branch, we wouldn't have this issue.

Or in other words, I don't think it's a matter of merging back, it's a matter of creating the tags on master and then create the release branch after it.

NagyZoltanPeter:
> > Unfortunately without the merge `git describe --abbrev=6 --always --tags` command which used to provide version info into the build will find any former 'latest' tag accessible from master, thus leads to misleading versioning. Now our images made from latest master has v0.25.0-rc.0-1234567 version instead of v0.26.0 as that is our latest release. Merging back solves this.
> 
> But I think that the problem here is that the tag wasn't made on master for 0.26.0. If we would have tagged 0.26.0 on master and then forked it to a release branch, we wouldn't have this issue.
> 
> Or in other words, I don't think it's a matter of merging back, it's a matter of creating the tags on master and then create the release branch after it.

I believe we shall tag the exact commit the release is made of. 
Let say the releasing process takes a week, during this time there can be new commits on master.... after the cut of release branch. Any change/fix made on release branch will be anyway needed on master, at least the change log.
We cannot duplicate tag but this way all condition is met.

gabrielmer:
> I believe we shall tag the exact commit the release is made of. Let say the releasing process takes a week, during this time there can be new commits on master.... after the cut of release branch. Any change/fix made on release branch will be anyway needed on master, at least the change log. We cannot duplicate tag but this way all condition is met.

My understanding is that we were following Trunk Based Development, according to https://github.com/waku-org/nwaku/blob/3d752b11b95d9c1143de5fa9e8505778af40c5b1/docs/contributors/release-process.md?plain=1#L5

So we shouldn't have changes in the release branch that aren't in master. Or in other words, changes are first made in master, and then cherry-picked to the release branch. If that's the case, there's nothing to merge back to master.

Regarding the tags, I think there's 2 cases.
1. If a release candidate is validated, then we should add the release tag into that commit in master and that's it.
2. If after we create the release candidate we make fixes to master, cherry-pick them to the release branch and in parallel master gets new unrelated commits, linearity has already been lost. There's no commit in master that exactly recreates the release.
Either we should rebase master on the release branch, which seems pretty dangerous and not a good solution, or add a release tag on master at the last commit that was cherry-picked to the release branch, even if it doesn't recreate exactly the release. This issue also happens if we merge the release branch back to master. But if fixes are done to master first and then cherry-picked to the release branch, then there's nothing to merge back.

Again, I might still be missing something - lmk 😶

Ivansete-status (approved):
LGTM! Thanks! :100:

gabrielmer (approved):
Now I understood, that when merging the branch to master the release tag is picked up even if it's in a commit not on master.

Thank you!

PR #2576, opened 2024-04-09
fix: node restart test issue

Ivansete-status:
## Description
It has been detected an issue regarding node restart and reconnection. After reconnecting, the messages didn't get its destination.

# Changes

- [x] The actual fix is provided by nim-libp2p update, and particularly the following commit: https://github.com/vacp2p/nim-libp2p/commit/b30b2656d52ee304bd56f8f8bbf59ab82b658a36
- [x] Simplify the `test_protocol.nim` so that it doesn't start things twice, f.e. the relay

## Issue

Special thanks to @AlejandroCabeza for such a great issue description :partying_face: 

closes https://github.com/waku-org/nwaku/issues/2145

### DIFF
--- a/waku/waku_relay/protocol.nim
+++ b/waku/waku_relay/protocol.nim
@@ -61,7 +61,6 @@ const TopicParameters = TopicParams(
 
 # see: https://rfc.vac.dev/spec/29/#gossipsub-v10-parameters
 const GossipsubParameters = GossipSubParams.init(
-  explicit = true,
   pruneBackoff = chronos.minutes(1),
   unsubscribeBackoff = chronos.seconds(5),
   floodPublish = true,
--- a/tests/waku_relay/test_protocol.nim
+++ b/tests/waku_relay/test_protocol.nim
@@ -60,13 +60,13 @@ suite "Waku Relay":
     pubsubTopicSeq = @[pubsubTopic]
     wakuMessage = fakeWakuMessage(testMessage, pubsubTopic)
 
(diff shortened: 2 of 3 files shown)


NagyZoltanPeter (approved):
LGTM!

waku/waku_relay/protocol.nim
```diff
@@ -61,7 +61,6 @@ const TopicParameters = TopicParams(
 
 # see: https://rfc.vac.dev/spec/29/#gossipsub-v10-parameters
 const GossipsubParameters = GossipSubParams.init(
-  explicit = true,
```
gabrielmer, line 64: What does this do?

gabrielmer (approved):
LGTM, thanks so much!

I wonder if this had the same root cause as https://github.com/waku-org/nwaku/issues/2592

Ivansete-status:
> LGTM, thanks so much!
> 
> I wonder if this had the same root cause as #2592

ah maybe yes, thanks for the heads-up! I'll take a look

waku/waku_relay/protocol.nim
Ivansete-status: Good question!
This is needed to make it compile properly. Is an adaptation required due to the following commit, https://github.com/vacp2p/nim-libp2p/commit/89cad5a3ba9088cc721682469a10917903da25a0, which is out of the original scope of this PR but I needed to adapt it as well because we are bumping `nim-libp2p` with the desired commit plus that additional commit.

PR #2577, opened 2024-04-09
fix(rln-relay): reduce sync time

rymnc:
# Description

Parallelizes RPC calls during initial sync while maintaining an RPC call delay to ensure client is not throttled.
We have to await them sequentially so we can retry the failed RPC call, maybe this can be improved.

Brings down sync time from 15 mins to ~ ~~2~~ 3 mins.
In retrospect, this is how it should have been done from the start, my fault!

# Changes

- [x] Collect futures of rpc calls into seq
- [x] await the futures

### DIFF
--- a/waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
+++ b/waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
@@ -606,6 +606,8 @@ proc startOnchainSync(
 
   # static block chunk size
   let blockChunkSize = 2_000
+  # delay between rpc calls to not overload the rate limit
+  let rpcDelay = 200.milliseconds
 
   var fromBlock =
     if g.latestProcessedBlock > g.rlnContractDeployedBlockNumber:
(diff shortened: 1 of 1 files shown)


waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
```diff
@@ -616,25 +618,42 @@ proc startOnchainSync(
...
 
       let toBlock = min(fromBlock + BlockNumber(blockChunkSize), currentLatestBlock)
       debug "fetching events", fromBlock = fromBlock, toBlock = toBlock
-      var handleBlockRes: bool
-      g.retryWrapper(handleBlockRes, "Failed to handle old blocks"):
-        await g.getAndHandleEvents(fromBlock, toBlock)
+      await sleepAsync(rpcDelay)
+      futs.add(g.getAndHandleEvents(fromBlock, toBlock))
       fromBlock = toBlock + 1
+    for fut in futs:
```
alrevuelta, line 645: am i missing something or is this unbounded? i mean what if the sync distance is 2 million blocks? will we spawn 2mill/2000 rpc calls?
If so we may:
* hit some OS limits? unsure if you can spawn async eg 200 calls such this?
* kill the poor eth node? specially if someone is using its own (which will most likely have limited resources).

rymnc: addressed in [7116bff](https://github.com/waku-org/nwaku/pull/2577/commits/7116bff97b70603feb0b5967fb56c4546a86dee5), moderately affects sync time, it is now ~3 mins on the same connection i tested previously with.

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
```diff
@@ -606,6 +619,10 @@ proc startOnchainSync(
 
   # static block chunk size
   let blockChunkSize = 2_000
+  # delay between rpc calls to not overload the rate limit
```
Ivansete-status, line 622: Do we know what is the rate limit or it depends on every RPC vendor?

Ivansete-status (approved):
LGTM! Thanks! :100: 
Only to mention that it may be interesting to use `Result` instead of exceptions. A refactoring  to consider :)

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
rymnc: depends on every rpc provider, this is quite lenient

gabrielmer (approved):
Thank you!

PR #2585, opened 2024-04-12
chore: don't create docker images for users without org's secrets

gabrielmer:
…crets

# Description

# Changes

- [ ] ...
- [ ] ...

### DIFF
--- a/.github/workflows/container-image.yml
+++ b/.github/workflows/container-image.yml
@@ -29,6 +29,17 @@ jobs:
     outputs:
       image: ${{ steps.build.outputs.image }}
     steps:
+      - name: Check secrets
+        id: secrets
+        run: |
+          if [[ -z "$QUAY_PASSWORD" || -z "$QUAY_USER" ]]; then
+            echo "User does not have access to secrets, skipping workflow"
+            exit 0
+          fi
+        env:
+          QUAY_PASSWORD: ${{ secrets.QUAY_PASSWORD }}
+          QUAY_USER: ${{ secrets.QUAY_USER }}
+      
       - name: Checkout code
         uses: actions/checkout@v3
 


NagyZoltanPeter (approved):
I'm not pretty sure if it is possible to use the local docker repo instead of publishing to quay in case creds are missing....?

.github/workflows/container-image.yml
```diff
@@ -29,15 +29,30 @@ jobs:
...
+          if [[ -z "$QUAY_PASSWORD" || -z "$QUAY_USER" ]]; then
+            echo "User does not have access to secrets, skipping workflow"
+            exit 1
+          fi
+        env:
+          QUAY_PASSWORD: ${{ secrets.QUAY_PASSWORD }}
+          QUAY_USER: ${{ secrets.QUAY_USER }}
+      
       - name: Checkout code
+        if: ${{ steps.secrets.outcome == 'success' }}
```
Ivansete-status, line 45: Aren't we just willing to skip the image deployment part if the user doesn't have access to secrets?

In other words, are we willing to skip the following snippet?
```
          docker login -u ${QUAY_USER} -p ${QUAY_PASSWORD} quay.io
          docker build -t ${IMAGE} -f docker/binaries/Dockerfile.bn.amd64 --label quay.expires-after=30d .
          docker push ${IMAGE}
```

Ivansete-status (commented):
Thanks for this! Just added a question :)

.github/workflows/container-image.yml
gabrielmer: Yes! So that was my initial approach. But then I noticed that the whole workflow doesn't make sense if the image is not being published. We already have separate workflows for builds, so there's no value in running another build, creating a docker-image and deleting it.

We could only skip the image deployment and the subsequent step which is writing in the PR the info of the image deployed, but in that case I don't think none of the previous add any value.

Ivansete-status (approved):

PR #2587, opened 2024-04-15
refactor: start moving discovery modules to waku/discovery

Ivansete-status:
## Description
We need to allow `nwaku` features to be accessible from both `wakunode2` and `libwaku`. In this PR, we start moving `discovery` logic from `wakunode2` app to a more generic location, `waku/node/discovery_manager`.

In future PRs, is likely that we will perform the following movements:

1. `wakuDiscv5` and `dynamicBootstrapNodes`, see [this](https://github.com/waku-org/nwaku/blob/480a62facdecd899c014ecdeedcc3f0be0085fd5/apps/wakunode2/app.nim#L68-L69), will get controlled only by `DiscoveryManager`.
2. The `WakuNode` object will contain a reference to `discoveryManager`, similarly to how it now contains a reference to `peerManager`. See: https://github.com/waku-org/nwaku/blob/4117fe65b774fd4875c10131fbe6630917333e76/waku/node/waku_node.nim#L84-L85

Notice that the `discoveryManager` module is aimed to contain the logic to handle the discovery of other peers plus the logic to make the "self" node discoverable by others.

## Issue

migrate DiscV5 and DNS Discovery from app.nim to waku_node.nim: https://github.com/waku-org/nwaku/issues/2452


gabrielmer:
I wonder if discovery should enter at the node layer, or if it belongs to the "middleware" layer we talked about when creating the factory. 

Definitely love the idea of having a discovery manager :) Having the discovery logic in its own module is a great step 🤩

NagyZoltanPeter (approved):
Yes, I think it is a good step for restructure. 
The only thing if sthg was changed in formatting as seemingly there are changes related only for that purpose without any meaningful change? 
But it's ok, manageable.

Ivansete-status:
> I wonder if discovery should enter at the node layer, or if it belongs to the "middleware" layer we talked about when creating the factory.
> 
> Definitely love the idea of having a discovery manager :) Having the discovery logic in its own module is a great step 🤩

Very good point!
I think the `DiscoveryManager` component can be at the same level as `PeerManager` because both imply some management and interaction with other peers. Another option, to fulfill with your suggestion, would be to move the "discovery" logic one folder up, and then have `waku/discovery_manager/` instead of `waku/node/discovery_manager`.

How do you feel about having the following? Does it sound correct?
```code
 WakuNode* = ref object 
   peerManager*: PeerManager
   discManager*: DiscoveryManager   <------ 
```

gabrielmer:
> 
> Very good point! I think the `DiscoveryManager` component can be at the same level as `PeerManager` because both imply some management and interaction with other peers. Another option, to fulfill with your suggestion, would be to move the "discovery" logic one folder up, and then have `waku/discovery_manager/` instead of `waku/node/discovery_manager`.
> 
> How do you feel about having the following? Does it sound correct?
> 
> ```
>  WakuNode* = ref object 
>    peerManager*: PeerManager
>    discManager*: DiscoveryManager   <------ 
> ```

It's true that both the `PeerManager` and the `DiscoveryManager` imply some management/interaction with other peers, but what I'm not sure about is if they are at the same layer of abstraction.

There's a chance that the `DiscoveryManager` is actually some kind of upper level logic intended for the `PeerManager` to get populated, which in that case they shouldn't be together. But I do lack some background, so not sure if what I'm saying makes sense at all haha, but I think it's important to have that clear before adding it to the `waku_node` module.

@jm-clius what's your take on it?

jm-clius:
While having a Discovery Manager makes sense from a code org POV, I wouldn't compose it into the node object. In fact, we've specifically moved discovery elements out of the node as we realised more and more that we needed the separate layer of abstraction (side note: we've also seriously considered moving peer management outside of the node object, even though it is indeed at a lower abstraction than the discovery manager). In general:
- a node is an object wrapping a libp2p switch, mounts libp2p protocols and manages the APIs to these protocols
- a peer manager is an object that manages a store of possible peers for the node and ensures that the node has the kind of connectivity it needs for every type of protocol.
- a discovery protocol may be any protocol/external provider of peers to the peer manager. These protocols may be discv5, DHTs, rendezvous, supplied lists, DNS lookups, etc. that have very little to do with libp2p transports and communications (in other words, it operates completely separately from the node and simply supplies the outcome of discovered peers to the peer manager).

gabrielmer (approved):
Looks amazing! Thanks so much! 😍

PR #2598, opened 2024-04-17
fix: create options api for cors preflight request.

kaichaosun:
# Description
CORS request triggers an Options request to same endpoint. [More context](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/OPTIONS#preflighted_requests_in_cors).
The changes here will fix send REST API request from web browser to local nwaku node.

# Changes

- [x] Access-Control-Allow-Headers set to Content-Type, this is needed to fix error `Reason: header ‘content-type’ is not allowed according to header ‘Access-Control-Allow-Headers’`
- [x] Options method enabled in relay rest api.

### DIFF
--- a/tests/wakunode_rest/test_rest_cors.nim
+++ b/tests/wakunode_rest/test_rest_cors.nim
@@ -86,7 +86,9 @@ proc checkResponse(
       expectedOrigin.isSome() and
       response.headers.contains("Access-Control-Allow-Origin") and
       response.headers.getLastString("Access-Control-Allow-Origin") ==
-      expectedOrigin.get()
+        expectedOrigin.get() and
+      response.headers.contains("Access-Control-Allow-Headers") and
+      response.headers.getLastString("Access-Control-Allow-Headers") == "Content-Type"
     )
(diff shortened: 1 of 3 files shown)


NagyZoltanPeter (approved):
Awesome!!! Thank you for this extension!
Should we have similar OPTION method handlers for all the enpoint paths?

NagyZoltanPeter:
> @NagyZoltanPeter We can add OPTIONS for other endpoints when people requesting it, not sure if it's a best practice though. Prefer not adding too much boilerplate code.

@kaichaosun Do you think it worth to make some more generic solution for it? I think of, in case allow-origin is configured we should auto add OPTION to each installed endpoint?
Current Presto implementation supports a chain of middleware callbacks can be installed, within such we could handle OPTION calls by checking valid paths.

Ivansete-status (approved):
LGTM! Thanks!
@kaichaosun - let's make sure all tests pass before merging :)
The only ones that we expect to fail are the `js-waku` jobs, which will be fixed soon

PR #2600, opened 2024-04-18
chore: enabling rest api as default

### LINKED ISSUE #2137: chore: enable rest api interface by default on wakunode

gabrielmer:
# Description

Enabling node's REST server as default

# Changes

- [x] changing `--rest` CLI param's default value from `false` to `true`

## Issue

closes #2137

### DIFF
--- a/waku/factory/external_config.nim
+++ b/waku/factory/external_config.nim
@@ -409,9 +409,7 @@ type WakuNodeConf* = object
 
     ## REST HTTP config
     rest* {.
-      desc: "Enable Waku REST HTTP server: true|false",
-      defaultValue: false,
-      name: "rest"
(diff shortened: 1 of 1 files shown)


NagyZoltanPeter (approved):
Thank U for taking care of it!

Ivansete-status (approved):
I think is reasonable :)

PR #2601, opened 2024-04-18
fix: don't use WakuMessageSize in req/resp protocols

chaitanyaprem:
# Description
While debugging https://github.com/waku-org/go-waku/issues/1076, i had noticed that for lightpush protocol, the read from stream is being done by using defaultMaxMsgSize which is 150Kib. This means even if a node sets this value to a higher one, lightpush send fails from a client if messageSize crosses 150Kb+overhead. 

More discussion on this here : https://discord.com/channels/1110799176264056863/1230492037099294720

# Changes

- [x] Rename MaxWakuMessageSize to DefaultMaxWakuMessageSize to avoid confusion. 
~~- [ ] LightPush serviceNode to use MaxWakuMessageSize to calculate RPCSize while reading a message. **Now i am not so sure if this is the right approach, rather let it try to publish via relay which would fail?**~~
- [x] All req/resp protocols to not use MaxWakuMessageSize rather try to read as much as possible by passing -1 to readLP. Not sure of the impact of this yet.

### DIFF
--- a/tests/waku_filter_v2/test_waku_client.nim
+++ b/tests/waku_filter_v2/test_waku_client.nim
@@ -126,8 +126,9 @@ suite "Waku Filter - End to End":
 
       asyncTest "Subscribing to an empty content topic":
         # When subscribing to an empty content topic
-        let subscribeResponse =
-          await wakuFilterClient.subscribe(serverRemotePeerInfo, pubsubTopic, newSeq[ContentTopic]())
+        let subscribeResponse = await wakuFilterClient.subscribe(
+          serverRemotePeerInfo, pubsubTopic, newSeq[ContentTopic]()
+        )
 
         # Then the subscription is not successful
         check:
@@ -1838,8 +1839,9 @@ suite "Waku Filter - End to End":
           wakuFilter.subscriptions.isSubscribed(clientPeerId)
 
         # When unsubscribing from an empty content topic
-        let unsubscribeResponse =
-          await wakuFilterClient.unsubscribe(serverRemotePeerInfo, pubsubTopic, newSeq[ContentTopic]())
+        let unsubscribeResponse = await wakuFilterClient.unsubscribe(
+          serverRemotePeerInfo, pubsubTopic, newSeq[ContentTopic]()
+        )
 
         # Then the unsubscription is not successful
         check:
@@ -2076,10 +2078,11 @@ suite "Waku Filter - End to End":
             contentTopic = contentTopic, payload = getByteSequence(100 * 1024)
           ) # 100KiB
           msg4 = fakeWakuMessage(
-            contentTopic = contentTopic, payload = getByteSequence(MaxPushSize - 1024)
+            contentTopic = contentTopic,
+            payload = getByteSequence(DefaultMaxPushSize - 1024),
           ) # Max Size (Inclusive Limit)
           msg5 = fakeWakuMessage(
-            contentTopic = contentTopic, payload = getByteSequence(MaxPushSize)
+            contentTopic = contentTopic, payload = getByteSequence(DefaultMaxPushSize)
           ) # Max Size (Exclusive Limit)
 
         # When sending the 1KiB message
--- a/tests/waku_lightpush/test_client.nim
+++ b/tests/waku_lightpush/test_client.nim
@@ -209,11 +209,11 @@ suite "Waku Lightpush Client":
         ) # 100KiB
         message4 = fakeWakuMessage(
           contentTopic = contentTopic,
-          payload = getByteSequence(MaxRpcSize - overheadBytes - 1),
+          payload = getByteSequence(DefaultMaxWakuMessageSize - overheadBytes - 1),
         ) # Inclusive Limit
         message5 = fakeWakuMessage(
           contentTopic = contentTopic,
-          payload = getByteSequence(MaxRpcSize - overheadBytes),
+          payload = getByteSequence(DefaultMaxWakuMessageSize - overheadBytes),
         ) # Exclusive Limit
 
       # When publishing the 1KiB payload
--- a/tests/waku_lightpush/test_client.nim
+++ b/tests/waku_lightpush/test_client.nim
@@ -43,6 +43,9 @@ suite "Waku Lightpush Client":
     handler = proc(
         peer: PeerId, pubsubTopic: PubsubTopic, message: WakuMessage
     ): Future[WakuLightPushResult[void]] {.async.} =
+      let msgLen = message.encode().buffer.len
+      if msgLen > int(DefaultMaxWakuMessageSize) + 64 * 1024:
+        return err("length greater than maxMessageSize")
       handlerFuture.complete((pubsubTopic, message))
       return ok()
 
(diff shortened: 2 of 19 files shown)


waku/waku_lightpush/protocol.nim
```diff
@@ -79,7 +80,7 @@ proc handleRequest*(
 
 proc initProtocolHandler(wl: WakuLightPush) =
   proc handle(conn: Connection, proto: string) {.async.} =
-    let buffer = await conn.readLp(MaxRpcSize.int)
+    let buffer = await conn.readLp(wl.maxRpcSize)
```
jm-clius, line 83: I may be missing something, but why are we not using the `-1` default here too?

waku/waku_lightpush/protocol.nim
```diff
@@ -92,12 +93,14 @@ proc new*(
...
     pushHandler: PushMessageHandler,
     rateLimitSetting: Option[RateLimitSetting] = none[RateLimitSetting](),
+    maxMessageSize = int(DefaultMaxWakuMessageSize),
 ): T =
   let wl = WakuLightPush(
     rng: rng,
     peerManager: peerManager,
     pushHandler: pushHandler,
     requestRateLimiter: newTokenBucket(rateLimitSetting),
+    maxRPCSize: calculateRPCSize(maxMessageSize),
```
jm-clius, line 103: I think adding state and dynamic calculation here is overkill. The idea of the max stream read size was simply to have _some_ reasonable safety limit that will prevent unrealistically (read: adversarially) large requests (which presumably could be an attack vector). We could have just used a very, very big magic number, but opted to initialise our constants with some value based off the `MaxWakuMsgSize` with safety margins to show that there was some rhyme and reason behind our thinking. Because this max msg size value is now configurable, our generous high limit may indeed now not be generous enough for differently configured nodes - at least in the lightpush case. However, even here the node could likely not configure message size to anything larger than the underlying libp2p limit (of 1MB). If we want to change this to be future proof, I'd rather keep using the underlying libp2p constant (1 MB) for all nodes and keep the simplicity of pure constants or just ignore the max rpc size limit (set to `-1` all around). The latter could presumably expose node to more attacks, though.

jm-clius (changes requested):
Some comments :)

gabrielmer (commented):
Thanks so much! 😍 Added a couple small questions/comments

waku/waku_lightpush/protocol.nim
chaitanyaprem: i started with changing lightpush this way..and later realised -1 is simpler approach. Will change for lightpush as well, as this much complexity is not needed.

waku/waku_lightpush/protocol.nim
chaitanyaprem: yep, realised this

tests/waku_lightpush/test_client.nim
```diff
@@ -209,11 +209,11 @@ suite "Waku Lightpush Client":
         ) # 100KiB
         message4 = fakeWakuMessage(
           contentTopic = contentTopic,
-          payload = getByteSequence(MaxRpcSize - overheadBytes - 1),
+          payload = getByteSequence(DefaultMaxWakuMessageSize - overheadBytes - 1),
```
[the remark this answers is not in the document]
chaitanyaprem: Ah..looks like I didn't think it through. Will fix.

tests/waku_filter_v2/test_waku_client.nim
```diff
@@ -126,8 +126,9 @@ suite "Waku Filter - End to End":
 
       asyncTest "Subscribing to an empty content topic":
         # When subscribing to an empty content topic
-        let subscribeResponse =
-          await wakuFilterClient.subscribe(serverRemotePeerInfo, pubsubTopic, newSeq[ContentTopic]())
```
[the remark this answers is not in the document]
chaitanyaprem: Interesting, will check which version I am using and update. I remember I set my nph to specific version that is being used, maybe env got messed up.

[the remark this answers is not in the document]
chaitanyaprem: Interesting, my nph version is v0.5.1-0-gde5cd48.

Does nph work differently on each machine?

waku/waku_lightpush/protocol.nim
chaitanyaprem: > The latter could presumably expose node to more attacks, though

I am wondering what kind of attacks, because read would anyways be done by the libp2p layer ir-respective of length passed. It is only the length check for the message that is done based on what is passed to it. 
We could keep it to 1MB, but then it is going to cause issues for protocols like Store where response can include 100*WakuMessageSize bytes which would definitely be more than 1MB even with default value. 

If we really have to deal with attacks, then maybe there should be some sort of length negotiation as part of metadata protocol or some other protocol when nodes connect to each other. This i feel is overkill at this point and can be taken up if we notice any vulnerabilities.

chaitanyaprem: > owever, even here the node could likely not configure message size to anything larger than the underlying libp2p limit (of 1MB).

Also, if this is a hard limit set by libp2p...then maybe we shouln't let users configure anything more than 1MB-WakuHeaders as maxMsgSize.

tests/waku_lightpush/test_client.nim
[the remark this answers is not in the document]
chaitanyaprem: Oddly this specific message check is not failing, the next one is failing. Will look little more deep to analyze what is happening.

tests/waku_filter_v2/test_waku_client.nim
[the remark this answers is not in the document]
NagyZoltanPeter: nph is just picked from the PATH I think. So it should be the same.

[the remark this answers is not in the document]
chaitanyaprem: Yep, so maybe this code was formatted with a different version prior to me touching it.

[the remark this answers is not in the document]
gabrielmer: Yes, I just confirmed and mine behaves the same way. So the issue was that it was previously formatted with a different version.

Apologies for the misunderstanding!

gabrielmer (approved):
Thanks so much!

waku/waku_lightpush/protocol.nim
jm-clius: > I am wondering what kind of attacks

Mmm. Indeed, not sure if possible impacts. It seems to me if someone were to pack the length prefix (or just send a very large amount of bytes), we'd have some potential to DoS the stream. This is a sanity check for the buffer length and if not reasonable the stream reader exits: https://github.com/vacp2p/nim-libp2p/blob/b30b2656d52ee304bd56f8f8bbf59ab82b658a36/libp2p/stream/lpstream.nim#L238-L239

jm-clius: > We could keep it to 1MB, but then it is going to cause issues for protocols like Store where response can include 100*WakuMessageSize bytes which would definitely be more than 1MB even with default value.

Which is why for store the max rpc size is `MaxPageSize*MaxWakuMsgSize + safety_margin`. Agree though. This is complicated by configurable sizes.

jm-clius: > This i feel is overkill at this point and can be taken up if we notice any vulnerabilities.

Agree.

jm-clius (approved):
Not completely sure of possible impact/vulnerabilities, but I think this is a reasonable solution for now.

PR #2609, opened 2024-04-19
feat: supporting meta field in store

### LINKED ISSUE #2432: feat: add WakuMessage's `meta` field to db schema
### Problem
As a continuation of #2214, we need to add support for storing WakuMessage's `meta` field in our db. For that, we need to update our schema and the db inserts to contain the new field.

### Acceptance criteria
`meta` field gets stored in the db

gabrielmer:
# Description
Adding support for saving the meta field in the db.

# Changes

- [x] supporting inserting and querying meta field in sqlite
- [x] supporting inserting and querying meta field in postgres
- [x] added unit tests 

## Issue

closes #2432


gabrielmer:
Currently, messages sent before the migration appear with `meta` field NULL while the ones after the migration are either empty (but not NULL) or with a value. Not sure about this inconsistency. If `meta` field is empty, should we set it to NULL everywhere? or just as empty string?

image

cc @NagyZoltanPeter @Ivansete-status

NagyZoltanPeter:
> Currently, messages sent before the migration appear with `meta` field NULL while the ones after the migration are either empty (but not NULL) or with a value. Not sure about this inconsistency. If `meta` field is empty, should we set it to NULL everywhere? or just as empty string?
> 
> image
> cc @NagyZoltanPeter @Ivansete-status

Hm, from pure DB perspective I vote for set it to NULL if meta has no value.

waku/factory/validator_signed.nim
```diff
@@ -34,6 +34,7 @@ proc msgHash*(pubSubTopic: string, msg: WakuMessage): array[32, byte] =
   ctx.update(msg.payload)
   ctx.update(msg.contentTopic.toBytes())
   ctx.update(msg.timestamp.uint64.toBytes(Endianness.littleEndian))
+  # ctx.update(msg.meta) meta is not included in the message hash, as the signature goes in the meta field
```
gabrielmer, line 37: This comment addresses the doubt if meta field should be included in the hash.
Not sure if to delete the comment or not, it's ugly but I feel that it can be useful

Ivansete-status: In my opinion there's no need to add `meta` to hash calculation but better to confirm with @jm-clius

Ivansete-status (approved):
LGTM! Thanks for it!
The only interesting point of simplification would be the migration script. I wonder if we could skip the "partitions" part in that migration script and only focus on the actual partitioned table schema change

SionoiS (approved):
LGTM

waku/waku_archive/driver/postgres_driver/postgres_driver.nim
```diff
@@ -31,11 +31,11 @@ type PostgresDriver* = ref object of ArchiveDriver
 const InsertRowStmtName = "InsertRow"
 const InsertRowStmtDefinition = # TODO: get the sql queries from a file
   """INSERT INTO messages (id, messageHash, storedAt, contentTopic, payload, pubsubTopic,
-  version, timestamp) VALUES ($1, $2, $3, $4, $5, $6, $7, $8) ON CONFLICT DO NOTHING;"""
+  version, timestamp, meta) VALUES ($1, $2, $3, $4, $5, $6, $7, $8, CASE WHEN $9 = '' THEN NULL ELSE $9 END) ON CONFLICT DO NOTHING;"""
```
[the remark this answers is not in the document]
gabrielmer: Great point! So the drivers have to cast the values to the expected Nim data type. So when retrieving the value from the db and setting it to a Nim object, both `NULL` or `''` are both converted to an empty seq of bytes

waku/factory/validator_signed.nim
SionoiS: Then yes indeed you can't add `meta` to the hash otherwise it would break this feature.

migrations/message_store_postgres/content_script_version_4.nim
```diff
@@ -0,0 +1,71 @@
...
+     max_storedAtSeconds integer = 0;
+     partition_name TEXT;
+     create_partition_stmt TEXT;
+BEGIN
+    SELECT MIN(storedAt) into min_storedAt
+    FROM messages_backup;
+
+    SELECT MAX(storedAt) into max_storedAt
+    FROM messages_backup;
+
```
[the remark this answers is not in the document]
gabrielmer: I just tried it without the partition part (https://github.com/waku-org/nwaku/pull/2609/commits/45a35e7fdfee57ab9d54a5d740e35c165d3ef3e0) but the migration failed :/
So for now I think it's better to leave it this way

waku/factory/validator_signed.nim
jm-clius: Please note that the _hash_ for the signed validator feature is an application-level message hash. It is different from the internal Waku Message hash.
For the signed validator feature, the `meta` field cannot be included (as pointed out, because it contains the very signature over the application-level hash). This is specified here: https://github.com/vacp2p/rfc-index/blob/main/status/raw/simple-scaling.md#dos-protection
For Waku Message hashes (the ones used in the Store and elsewhere) the `meta` field forms an essential part in order to allow applications to differentiate messages that may be exactly similar in all other fields: https://github.com/vacp2p/rfc-index/blob/main/waku/standards/core/14/message.md#deterministic-message-hashing

PR #2610, opened 2024-04-22
refactor: REST - create rest/builder.nim and simplify factory/app.nim

Ivansete-status:
# Description
Extract some REST logic from app.nim, aiming for simplify `app.nim` a little.
I will submit a couple of similar PRs simplifying `app.nim`.

## Issue
With that simplification, this PR indirectly contributes to the following: migrate DiscV5 and DNS Discovery from app.nim to waku_node.nim https://github.com/waku-org/nwaku/issues/2452

### DIFF
--- a/waku/factory/app.nim
+++ b/waku/factory/app.nim
@@ -29,16 +29,8 @@ import
   ../../waku/node/waku_metrics,
   ../../waku/node/peer_manager,
   ../../waku/node/peer_manager/peer_store/waku_peer_storage,
-  ../../waku/waku_api/message_cache,
-  ../../waku/waku_api/handlers,
   ../../waku/waku_api/rest/server,
-  ../../waku/waku_api/rest/debug/handlers as rest_debug_api,
-  ../../waku/waku_api/rest/relay/handlers as rest_relay_api,
-  ../../waku/waku_api/rest/filter/handlers as rest_filter_api,
-  ../../waku/waku_api/rest/lightpush/handlers as rest_lightpush_api,
-  ../../waku/waku_api/rest/store/handlers as rest_store_api,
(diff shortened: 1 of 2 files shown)


SionoiS (approved):
LGTM

I like this a lot!

NagyZoltanPeter (changes requested):
I think we are crossing our works. Can you please take a look at mine. I will draft it now. I'm just adding a final json serialization for a bit extended /health endpoint.
To enable separation of initialization health from other endpoints and start rest server earlier, I needed to rework to order of rest initialization. 
Still, I agree to simplify waku node initialization order due I found its overly complicated.

waku/factory/app.nim
```diff
@@ -456,18 +314,15 @@ proc setupMonitoringAndExternalInterfaces*(app: var App): AppResult[void] =
...
 
   ok()
 
 # App shutdown
 
 proc stop*(app: App): Future[void] {.async: (raises: [Exception]).} =
-  if app.restServer.isSome():
-    await app.restServer.get().stop()
+  if app.conf.rest:
+    await app.restServer.stop()
```
gabrielmer, line 325: Not sure I don't understand this change. Shouldn't we check for the `restServer` to be already mounted and not just in the configurations?

gabrielmer (commented):
Looks amazing! Thanks so much! Love this change 😍 

Just added a tiny question

waku/factory/app.nim
Ivansete-status: Yes makes sense! I'll double-check

gabrielmer (approved):
Thanks so much!

Ivansete-status:
Closing because has been solved in https://github.com/waku-org/nwaku/pull/2623

PR #2613, opened 2024-04-22
fix(ci): unique comment_tag to reference rln version

rymnc:
# Description

There is a race condition for the container-image.yml job that replaces comments with rln-v2 true and rln-v2 false images.
There should be separate comments for both versions.

# Changes

- [x] Alter `comment_tag` in the container-image.yml job


Ivansete-status (approved):

gabrielmer (approved):
Thanks so much!

PR #2614, opened 2024-04-22
fix(rln-relay): increase retries for 1 minute recovery time

rymnc:
# Description

This PR increases the values for `retryDelay` and `retryCount` to allow nodes to recover over a longer duration than before (3 seconds)

# Changes

- [x] increases `retryDelay` to `4000`ms
- [x] increases `retryCount` to 15, giving the nodes 1 minute to recover


chaitanyaprem (approved):
LGTM

PR #2622, opened 2024-04-23
fix(rln-relay): enforce error callback to remove exception raised from retryWrapper

rymnc:
# Description

To address something that is blocking #2607, this is necessary to remove possibilities of exceptions being raised in the `retryWrapper` template.

# Changes

- [x] remove default value for errCallback


gabrielmer (approved):
Thanks!

Ivansete-status (approved):
LGTM! Thanks! :heart:

PR #2626, opened 2024-04-25
chore: Enabling to use a full node for lightpush via rest api without lightpush client configured

NagyZoltanPeter:
# Description
This is helper enhancment for testing (rate limit specifically for me) on lightpush... similar to query self store node via REST API.
With this change one can issue lightpush request via REST API on a service node having lightpush protocol installed.

In contrast it was needed to run a full node with lightpush service and a light client (configured with the service node peerId) to be usable from REST API.
With this change a lightpush service node (with relay of course) can serve such lightpush API request and can relay such messages.

This can help in various way.
- easy develop on REST API with nwaku-compose (where a full node is set up)
- easy test lightpush features via rest api.

# Changes

- [X] /lightpush endpoint is enabled either lightpushclient or lightpush(protocol) mounted
- [X] lightpush protocol can handle such request by self hosted coming from REST API

## How to test

1. Run a full node or nwaku-compose with the right image
2. issue REST /lightpush POST message
3. check that relayed (can depend on various stuff)

## Open question:
- Currently relaying message via REST API will attach rln proof to the message. This is missing from lightpush, thus rest api user should be able to provide the proof.
- I think this can be think as, I'm using a thin layer towards the network via REST API by running a light node with my rln credentials. Such case lightpush service can attach proof to the message as in relay (REST API case).

### DIFF
--- a/waku/node/waku_node.nim
+++ b/waku/node/waku_node.nim
@@ -40,6 +40,7 @@ import
   ../waku_lightpush/client as lightpush_client,
   ../waku_lightpush/common,
   ../waku_lightpush/protocol,
+  ../waku_lightpush/self_req_handler,
   ../waku_enr,
   ../waku_peer_exchange,
   ../waku_rln_relay,
@@ -813,7 +814,7 @@ proc mountLightPush*(
 
       if publishedCount == 0:
         ## Agreed change expected to the lightpush protocol to better handle such case. https://github.com/waku-org/pm/issues/93
-        debug("Lightpush request has not been published to any peers")
+        debug "Lightpush request has not been published to any peers"
 
       return ok()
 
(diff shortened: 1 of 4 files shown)


waku/waku_api/rest/builder.nim
```diff
@@ -175,7 +175,10 @@ proc startRestServerProtocolSupport*(
   installStoreApiHandlers(router, node, storeDiscoHandler)
 
   ## Light push API
-  if conf.lightpushnode != "" and node.wakuLightpushClient != nil:
+  ## Install it either if lightpushnode (lightpush service node) is configured and client is mounted)
+  ## or install it to be used with self-hosted lightpush service
+  if (conf.lightpushnode != "" and node.wakuLightpushClient != nil) or
+      (conf.lightpush and node.wakuLightPush != nil):
```
SionoiS, line 181: Don't we need to check relay and rln too?

waku/node/waku_node.nim
```diff
@@ -842,15 +843,33 @@ proc lightpushPublish*(
...
+        peer = peer.peerId
+      return await node.wakuLightpushClient.publish(pubsubTopic, message, peer)
+
+    if not node.wakuLightPush.isNil():
+      debug "publishing message with self hosted lightpush",
+        pubsubTopic = pubsubTopic, contentTopic = message.contentTopic
+      return await node.wakuLightPush.handleSelfLightPushRequest(pubsubTopic, message)
+
+  if node.wakuLightpushClient.isNil() and node.wakuLightPush.isNil():
+    return err("waku lightpush not available")
```
SionoiS, line 865: I would put that first in the proc

SionoiS (approved):
Approved to not block but this flow seams a bit convoluted.

Maybe in `waku_node` local vs remote light push should be 2 different procs?

waku/waku_lightpush/self_req_handler.nim
```diff
@@ -0,0 +1,54 @@
...
+## any store-peer address.
+##
+## Notice that the REST-store requests normally assume that the REST
+## server is acting as a store-client. In this module, we allow that
+## such REST-store node can act as store-server as well by retrieving
+## its own stored messages. The typical use case for that is when
+## using `nwaku-compose`, which spawn a Waku node connected to a local
+## database, and the user is interested in retrieving the messages
+## stored by that local store node.
+##
```
gabrielmer, line 14: Probably need to change this instances of "store" in this comment for "lightpush", and removing the parts that aren't relevant

waku/waku_lightpush/self_req_handler.nim
```diff
@@ -0,0 +1,54 @@
...
+  ./protocol_metrics,
+  ../utils/requests
+
+proc handleSelfLightPushRequest*(
+    self: WakuLightPush, pubSubTopic: PubsubTopic, message: WakuMessage
+): Future[WakuLightPushResult[void]] {.async.} =
+  ## Handles the lightpush requests made by the node to itself.
+  ## Normally used in REST-lightpush requests
+
+  try:
```
gabrielmer, line 32: Maybe it's better to have the try more scoped?

waku/waku_lightpush/self_req_handler.nim
```diff
@@ -0,0 +1,54 @@
...
+    let req = PushRequest(pubSubTopic: pubSubTopic, message: message)
+    let rpc = PushRPC(requestId: generateRequestId(self.rng), request: some(req))
+
+    let respRpc = await self.handleRequest(selfPeerId, rpc.encode().buffer)
+
+    if respRpc.response.isNone():
+      waku_lightpush_errors.inc(labelValues = [emptyResponseBodyFailure])
+      return err(emptyResponseBodyFailure)
+
+    let response = respRpc.response.get()
```
gabrielmer, line 45: Maybe it's shorter to do something like

```suggestion
   let response = respRpc.response.valueOr:
     waku_lightpush_errors.inc(labelValues = [emptyResponseBodyFailure])
     return err(emptyResponseBodyFailure)
     
```

gabrielmer (approved):
Superr, thanks so much!

Added a couple questions/comments :)

waku/waku_lightpush/self_req_handler.nim
NagyZoltanPeter: Ah, yes I was using store self query as motivation, but did not check the comment careful.

PR #2627, opened 2024-04-25
chore: ci.yml - avoid calling brew link libpq --force on macos

Ivansete-status:
## Description
PRs started to fail due to that but we are not actually running Postgres tests on MacOS


rymnc (approved):

NagyZoltanPeter (approved):
Can't wait to get ci work again! :-)

PR #2631, opened 2024-04-25
fix: missing rate limit setting for legacy store protocol

NagyZoltanPeter:
# Description
https://github.com/waku-org/nwaku/pull/2431 missed setting rate limit configuration for legacy store protocol mount.

Also set default rate limit setting to zero in order to let rate limit checks switched off for now.


SionoiS (approved):
Thank you!

gabrielmer (approved):
Thanks! 😍

PR #2632, opened 2024-04-25
fix: store v3 proto field numbers

SionoiS:
Updated the protobuf field numbers according to the spec. https://github.com/vacp2p/rfc/tree/docs/new-store-v3-protocol/content/docs/rfcs/13#payloads

### DIFF
--- a/waku/node/waku_node.nim
+++ b/waku/node/waku_node.nim
@@ -832,6 +832,7 @@ proc toStoreResult(res: ArchiveResult): StoreQueryResult =
   var res = StoreQueryResponse()
 
   res.statusCode = 200
+  res.statusDesc = "OK"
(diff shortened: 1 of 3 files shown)


richard-ramos (approved):

NagyZoltanPeter (approved):
Looks good, but hard to judge without seeing the proto definition.
I guess this will go with this: https://github.com/waku-org/waku-proto/issues/22

SionoiS:
> Looks good, but hard to judge without seeing the proto definition. I guess this will go with this: [waku-org/waku-proto#22](https://github.com/waku-org/waku-proto/issues/22)

this is the best we have. https://github.com/vacp2p/rfc/tree/docs/new-store-v3-protocol/content/docs/rfcs/13#payloads WIP

PR #2633, opened 2024-04-25
fix: parse shards properly in enr config for non twn

chaitanyaprem:
# Description
Fix for https://github.com/waku-org/nwaku/issues/2618

# Changes

- [x] Parse shardinfo from pubsubTopics to be used in ENR generation

### DIFF
--- a/waku/factory/internal_config.nim
+++ b/waku/factory/internal_config.nim
@@ -37,7 +37,8 @@ proc enrConfiguration*(
       if shardsRes.isOk() and shardsRes.get().isSome():
         shardsLocal = shardsRes.get().get().shardIds
       else:
-        error "failed to parse pubsubTopic", error = shardsRes.error
+        error "failed to parse pubsub topic, please format according to static shard specification",
+          error = shardsRes.error
       shardsLocal
 
     # some shards configured


waku/factory/internal_config.nim
```diff
@@ -32,7 +32,14 @@ proc enrConfiguration*(
   let shards: seq[uint16] =
     # no shards configured
     if conf.shards.len == 0:
-      toSeq(0 ..< conf.pubsubTopics.len).mapIt(uint16(it))
+      var shardsLocal = newSeq[uint16]()
```
NagyZoltanPeter, line 35: I think it can be simpler:

```
(topicsToRelayShards(conf.pubsubTopics).valueOr:
        error "failed to parse pubsubTopic", error = shardsRes.error
       @[]
).get(@[])
```
Something like that is possible. Did not tried for now.

NagyZoltanPeter (approved):
Overall it is ok, maybe can be written in simpler form. You may decide :-)

waku/factory/internal_config.nim
```diff
@@ -32,7 +32,14 @@ proc enrConfiguration*(
...
+      let shardsRes = topicsToRelayShards(conf.pubsubTopics)
+      if shardsRes.isOk() and shardsRes.get().isSome():
+        shardsLocal = shardsRes.get().get().shardIds
+      else:
+        error "failed to parse pubsubTopic", error = shardsRes.error
```
SionoiS, line 40: I would suggest error message to be actionable statement.

egg. Parsing pubsub topic failed. Please format according to static shard specification.

SionoiS (approved):
Thank you!

waku/factory/internal_config.nim
chaitanyaprem: That looks complex to read, will try it.

PR #2636, opened 2024-04-26
fix: store v3 validate cursor & remove messages 

SionoiS:
Fixes https://github.com/waku-org/nwaku/issues/2635 and https://github.com/waku-org/nwaku/issues/2634

### DIFF
--- a/tests/wakunode_rest/test_rest_store.nim
+++ b/tests/wakunode_rest/test_rest_store.nim
@@ -226,7 +226,8 @@ procSuite "Waku Rest API - Store v3":
 
       var wakuMessages = newSeq[WakuMessage](0)
       for j in 0 ..< response.data.messages.len:
-        wakuMessages.add(response.data.messages[j].message)
+        if response.data.messages[j].message.isSome():
+          wakuMessages.add(response.data.messages[j].message.get())
 
       pages[i] = wakuMessages
 
@@ -629,7 +630,7 @@ procSuite "Waku Rest API - Store v3":
       $response.contentType == $MIMETYPE_JSON
       response.data.messages.len == 1
 
-    let storeMessage = response.data.messages[0].message
+    let storeMessage = response.data.messages[0].message.get()
 
     check:
       storeMessage.payload == msg.payload
@@ -713,7 +714,8 @@ procSuite "Waku Rest API - Store v3":
 
       var wakuMessages = newSeq[WakuMessage](0)
       for j in 0 ..< response.data.messages.len:
-        wakuMessages.add(response.data.messages[j].message)
+        if response.data.messages[j].message.isSome():
+          wakuMessages.add(response.data.messages[j].message.get())
 
       pages[i] = wakuMessages
 
@@ -776,7 +778,8 @@ procSuite "Waku Rest API - Store v3":
 
     var wakuMessages = newSeq[WakuMessage](0)
     for j in 0 ..< response.data.messages.len:
-      wakuMessages.add(response.data.messages[j].message)
+      if response.data.messages[j].message.isSome():
+        wakuMessages.add(response.data.messages[j].message.get())
 
     check wakuMessages == msgList[6 .. 9]
 
--- a/tests/wakunode_rest/test_rest_store.nim
+++ b/tests/wakunode_rest/test_rest_store.nim
@@ -620,8 +620,9 @@ procSuite "Waku Rest API - Store v3":
     let client = newRestHttpClient(initTAddress(restAddress, restPort))
 
     # Filtering by a known pubsub topic.
-    var response =
-      await client.getStoreMessagesV3(pubsubTopic = encodeUrl(DefaultPubsubTopic))
+    var response = await client.getStoreMessagesV3(
+      includeData = "true", pubsubTopic = encodeUrl(DefaultPubsubTopic)
+    )
 
     check:
       response.status == 200
--- a/tests/wakunode_rest/test_rest_store.nim
+++ b/tests/wakunode_rest/test_rest_store.nim
@@ -1,7 +1,7 @@
 {.used.}
 
 import
-  std/[options, times],
+  std/[options, times, sugar],
   stew/shims/net as stewNet,
   chronicles,
   testutils/unittests,
@@ -224,10 +224,10 @@ procSuite "Waku Rest API - Store v3":
         "7", # page size. Empty implies default page size.
       )
 
-      var wakuMessages = newSeq[WakuMessage](0)
-      for j in 0 ..< response.data.messages.len:
-        if response.data.messages[j].message.isSome():
-          wakuMessages.add(response.data.messages[j].message.get())
+      let wakuMessages = collect(newSeq):
+        for element in response.data.messages:
+          if element.message.isSome():
+            element.message.get()
 
       pages[i] = wakuMessages
 
@@ -712,10 +712,10 @@ procSuite "Waku Rest API - Store v3":
         "3", # page size. Empty implies default page size.
       )
 
-      var wakuMessages = newSeq[WakuMessage](0)
-      for j in 0 ..< response.data.messages.len:
-        if response.data.messages[j].message.isSome():
-          wakuMessages.add(response.data.messages[j].message.get())
+      let wakuMessages = collect(newSeq):
+        for element in response.data.messages:
+          if element.message.isSome():
+            element.message.get()
 
       pages[i] = wakuMessages
 
@@ -776,10 +776,10 @@ procSuite "Waku Rest API - Store v3":
       response.status == 200
       $response.contentType == $MIMETYPE_JSON
 
-    var wakuMessages = newSeq[WakuMessage](0)
-    for j in 0 ..< response.data.messages.len:
-      if response.data.messages[j].message.isSome():
-        wakuMessages.add(response.data.messages[j].message.get())
+    let wakuMessages = collect(newSeq):
+      for element in response.data.messages:
+        if element.message.isSome():
+          element.message.get()
 
     check wakuMessages == msgList[6 .. 9]
 
--- a/waku/waku_archive/archive.nim
+++ b/waku/waku_archive/archive.nim
@@ -140,6 +140,7 @@ proc findMessages*(
 
   let rows = (
     await self.driver.getMessages(
+      includeData = query.includeData,
       contentTopic = query.contentTopics,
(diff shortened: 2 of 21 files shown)


waku/waku_archive/archive.nim
```diff
@@ -163,7 +166,9 @@ proc findMessages*(
   let pageSize = min(rows.len, int(maxPageSize))
 
   #TODO once store v2 is removed, unzip instead of 2x map
-  messages = rows[0 ..< pageSize].mapIt(it[1])
+  if query.includeData:
```
richard-ramos, line 169: This perhaps can be done in a separate PR, but a potential improvement for this code could be to consider modifying the `getMessages` function in line 139 to also pass the `includeData`, and then, in the drivers define a query that will only do a  `SELECT messageHash FROM messages .......` if the value from `includeData` is `true`.

We should check the query plan from sqlite/postgres, of course, but depending on how the indexes were defined, this should be more efficient, due to reduced disk i/o and data transfer from disk to memory, since the query will be satisfied by only reading the values from the index, without needing to fetch additional data from the table. I imagine this would be the case as the following indexes are defined in the code as far as i could see:
- Primary Key: messageHash
- Index: contentTopic, pubsubTopic, senderTimestamp, id
- Index: storedAt

richard-ramos (changes requested):
For a message that exists in the DB, the response is incorrect. It should have included a WakuMessageKeyValue with just the message hash.

```
REQUEST
{"request_id":"d4b573abb709d067da8afca3ec7341a8fbebee09923b4ade414e6b510bb6e24f","include_data": false, "message_hashes":["PmqobzCu78df5nprhzOUUu0e6pCFcOACjQpJistzVKM="],"pagination_forward":true,"pagination_limit":20}

RESPONSE
{"request_id":"d4b573abb709d067da8afca3ec7341a8fbebee09923b4ade414e6b510bb6e24f","status_code":200,"status_desc":"OK"}
```

If I change the `include_data` attribute to `true`, the response returns as expected (hash + message)

waku/waku_archive/archive.nim
richard-ramos: I'll create an issue to not lose track of this then!

SionoiS: In retrospect, I try to share as much code with v2 as possible an it was a mistake.

richard-ramos:
For `includeData` we can use:
```
  var inclData: hint32
  if not ?pb.getField(2, inclData):
    req.includeData = false
  else:
    req.includeData = int32(inclData) == int32(1)
 ```
because according to https://protobuf.dev/programming-guides/encoding/
> Bools and enums are both encoded as if they were int32s. Bools, in particular, always encode as either 00 or 01. In Protoscope, false and true are aliases for these byte strings.

File: waku/waku_store/rpc_codec.nim [hunk 2]
```diff
@@ -56,11 +56,11 @@ proc decode*(
   if not ?pb.getField(1, req.requestId):
     return err(ProtobufError.missingRequiredField("request_id"))
 
-  var inclData: uint
+  var inclData: zint64
```
richard-ramos, line 59:
```suggestion
  var inclData: hint32
```

because according to https://protobuf.dev/programming-guides/encoding/
 > Bools and enums are both encoded as if they were int32s. Bools, in particular, always encode as either 00 or 01. In Protoscope, false and true are aliases for these byte strings.

waku/waku_store/rpc_codec.nim
```diff
@@ -56,11 +56,11 @@ proc decode*(
...
   if not ?pb.getField(2, inclData):
     req.includeData = false
   else:
-    req.includeData = inclData == 1
+    req.includeData = inclData == zint64(1)
```
richard-ramos, line 63:
```suggestion
    req.includeData = int32(inclData) == int32(1)
```

waku/waku_archive/archive.nim
```diff
@@ -133,6 +133,9 @@ proc findMessages*(
   if query.contentTopics.len > 10:
     return err(ArchiveError.invalidQuery("too many content topics"))
 
+  if query.cursor.isSome() and query.cursor.get().hash.len != 32:
+    return err(ArchiveError.invalidQuery("invalid cursor"))
```
richard-ramos, line 137: Instead of doing this validation here, consider adding in `rcp_codec.nim`, line 103:
```nim
    if cursor.len != 32:
      return err(ProtobufError.missingRequiredField("cursor"))
```

Also, it's probably a good idea to replace line 94 for this, to validate the hashes length
```nim
   for i, buf in buffer:
      if buf.len != 32:
        return err(ProtobufError.invalidLengthField("message_hashes[" & $i & "]"))
```

Also, what do you think of editing `protocol.nim` to return the error detail when decoding a protobuffer:
```nim
    res.statusDesc = "decode rpc failed " & $error
```
For example with these changes the status description would look like this:
```
decode rpc failed (kind: MissingRequiredField, field: "cursor")
```

richard-ramos:
What should happen if I send a request with a valid cursor that does not exist? i.e. `0x0000000000000000000000000000000000000000000000000000000000000000`:
```

{"request_id":"ec560c592484291ca6dfa1491507a45dba55b238c2b271cf0cb81ce1e0590fce","include_data":true,"pubsub_topic":"/waku/2/default-waku/proto","content_topics":["test"],"time_start":1714155079182460704,"time_end":1714155079285053447,"pagination_cursor":"AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=","pagination_forward":true,"pagination_limit":20}
```

Currently it is returning all the records but it seems to me that this behavior is not valid, and perhaps it should return either no results, or an invalid_cursor error.  I think we should do the later, and probably do a `SELECT EXISTS(SELECT 1 FROM messages WHERE message_hash=?)` to determine if the cursor is valid. Since the message_hash is the primary key of the table I imagine this should execute fast with no major impact on the total store query response time

SionoiS:
> Currently it is returning all the records but it seems to me that this behavior is not valid, and perhaps it should return either no results, or an invalid_cursor error. I think we should do the later, and probably do a `SELECT EXISTS(SELECT 1 FROM messages WHERE message_hash=?)` to determine if the cursor is valid. Since the message_hash is the primary key of the table I imagine this should execute fast with no major impact on the total store query response time

Seams like the old code never validated the cursor, it only used to order the results. I agree that we should do it though.

richard-ramos:
> Seams like the old code never validated the cursor

I see! It can be done later in a separate PR then. Will open a feature request issue so we don't lose track of this :)
I just tested this PR with my set of tests in go-waku and works fine!

richard-ramos (approved):

waku/waku_archive/archive.nim
SionoiS: > Instead of doing this validation here, consider adding in `rcp_codec.nim`, line 103:
> 
> ```nim
>     if cursor.len != 32:
>       return err(ProtobufError.invalidLengthField("cursor"))
> ```
> 
> Also, it's probably a good idea to replace line 94 for this, to validate the hashes length
> 
> ```nim
>    for i, buf in buffer:
>       if buf.len != 32:
>         return err(ProtobufError.invalidLengthField("message_hashes[" & $i & "]"))
> ```

I disagree, I feel like it would be best to keep cursor validation in archive and protobuf validation separate.

Is there a way to specify a limited `bytes` field in protobuf?

> What do you think of editing `protocol.nim` to return the error detail when decoding a protobuffer:
> 
> ```nim
>     res.statusDesc = "decode rpc failed " & $error
> ```
> 
> For example with these changes the status description would look like this:
> 
> ```
> decode rpc failed (kind: MissingRequiredField, field: "cursor")
> ```

:100 agree

richard-ramos: > Is there a way to specify a limited bytes field in protobuf?

I think it's not possible unless you use a code generator like https://github.com/bufbuild/protoc-gen-validate/tree/main which reads annotations from protobuffers. We're stuck with manual validation for now :disappointed:

waku/waku_archive/driver/postgres_driver/postgres_driver.nim
```diff
@@ -377,7 +377,7 @@ proc getMessagesV2ArbitraryQuery(
     endTime = none(Timestamp),
     maxPageSize = DefaultPageSize,
     ascendingOrder = true,
-): Future[ArchiveDriverResult[seq[ArchiveRow]]] {.async.} =
+): Future[ArchiveDriverResult[seq[ArchiveRow]]] {.async, deprecated.} =
```
Ivansete-status, line 380: I'm not saying that is incorrect to label those procs as `deprecated` but how we should handle those? Do they still work/compile?

waku/waku_archive/common.nim
```diff
@@ -43,6 +43,7 @@ type
     hash*: WakuMessageHash
 
   ArchiveQuery* = object
+    includeData*: bool
```
Ivansete-status, line 46: May we have a little comment about this field?

waku/waku_archive/archive.nim
```diff
@@ -133,10 +133,14 @@ proc findMessages*(
...
```
Ivansete-status, line 137: nitpick one ;P
```suggestion
    return err(ArchiveError.invalidQuery("invalid cursor hash length"))
```

tests/wakunode_rest/test_rest_store.nim
```diff
@@ -775,7 +778,8 @@ procSuite "Waku Rest API - Store v3":
 
     var wakuMessages = newSeq[WakuMessage](0)
     for j in 0 ..< response.data.messages.len:
-      wakuMessages.add(response.data.messages[j].message)
+      if response.data.messages[j].message.isSome():
+        wakuMessages.add(response.data.messages[j].message.get())
```
Ivansete-status, line 782: Unrelated to the actual change but maybe we could use `filterIt` in this case.

Something like:

`let wakuMessages = response.data.messages.filterIt(it.isSome())`

Ivansete-status (approved):
LGTM! Thanks!
Just added some comments that I hope you find useful

waku/waku_archive/driver/postgres_driver/postgres_driver.nim
SionoiS: I assumed it would only be a visual thing but I can't find any documentation so :shrug: They do compile/work though.

PR #2641, opened 2024-04-28
chore: add default react native app to examples

richard-ramos:
This PR adds the default react native app to the android branch. I used a separate PR so it's easier to review the next PRs since they will only show changes on this example app, and not just the tens of files that the default app contains.


gabrielmer (approved):
LGTM, thank you!

Ivansete-status (approved):
Wow incredible! Thanks!
I approve but I haven't checked or tested it carefully

PR #2645, opened 2024-04-29
refactor: waku_node.nim - simplify stop proc

Ivansete-status:
## Description
There is no need to explicitly stop mounted libp2p protocols because they are already being stopped after the switch.stop() is being called.

## Issue
I've noticed that while debugging:
Peer Reconnection not working - https://github.com/waku-org/nwaku/issues/2592

### DIFF
--- a/waku/node/waku_node.nim
+++ b/waku/node/waku_node.nim
@@ -1217,13 +1217,6 @@ proc start*(node: WakuNode) {.async.} =
   info "Node started successfully"
 
 proc stop*(node: WakuNode) {.async.} =
-  if not node.wakuRelay.isNil():
-    await node.wakuRelay.stop()
-
-  if not node.wakuMetadata.isNil():
-    node.wakuMetadata.stop()
-
-  await node.switch.stop()
   node.peerManager.stop()
 
   if not node.wakuRlnRelay.isNil():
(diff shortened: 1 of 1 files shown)


gabrielmer (commented):
I see that in `switch.stop()` all the services mounted are stopped

https://github.com/vacp2p/nim-libp2p/blob/2b5319622c997ce1c80bc62c863e30f3349ee0d7/libp2p/switch.nim#L291-L292

However, I'm not finding where our protocols are assigned to the `services` field.
I see that protocols are mounted at peerInfo.protocols at

https://github.com/vacp2p/nim-libp2p/blob/2b5319622c997ce1c80bc62c863e30f3349ee0d7/libp2p/switch.nim#L213

Maybe I'm missing something? Where exactly are the protocols being stopped when the switch is being stopped?

SionoiS (commented):
As far as I can see nim-libp2p `Services` != `LPProtocol` and so the `stop` is for 2 different object completely.

Where are services and protocol linked? I can't find it.

Ivansete-status:
Thanks for the comments @gabrielmer @SionoiS !

Services and protocols are different entities.

The protocols are actually stopped in https://github.com/vacp2p/nim-libp2p/blob/b30b2656d52ee304bd56f8f8bbf59ab82b658a36/libp2p/switch.nim#L320

image

Then from within `MultistreamSelect` object, we have the precise point where the mounted `LPProtocol`s are stopped:
https://github.com/vacp2p/nim-libp2p/blob/b30b2656d52ee304bd56f8f8bbf59ab82b658a36/libp2p/multistream.nim#L294

image

----

Aside note: the only service that we are currently attaching to the waku switch is the "autonat" one. See:
image

SionoiS (approved):
Thanks! Now I understand. Much simpler this way.

PR #2646, opened 2024-04-29
fix: rest/store/types: contentTopic -> content_topic

Ivansete-status:
## Description
Simple PR to change to properly return `content_topic` in _Store_ REST reponses

## Issue

closes https://github.com/waku-org/nwaku/issues/2582


chaitanyaprem (commented):
Maybe better to use constants for such field names which are common across all protocols to avoid such issues in the future.

Ivansete-status:
> Maybe better to use constants for such field names which are common across all protocols to avoid such issues in the future.

Thanks! I will consider for upcoming PRs

PR #2649, opened 2024-04-29
fix(rln-relay): persist metadata every batch during initial sync

rymnc:
# Description

This PR offloads the `setMetadata` proc from the `atomicBatch` proc to the callsites of the `getAndHandleEvents` so that it is not called twice, and during the initial sync when we parallelize the rpc calls we want to ensure the metadata is set in increasing block number counts to avoid race conditions and missed blocks.

# Changes

- [x] set metadata appropriately when parallelizing rpc calls

### DIFF
--- a/tests/waku_rln_relay/test_rln_group_manager_onchain.nim
+++ b/tests/waku_rln_relay/test_rln_group_manager_onchain.nim
@@ -328,7 +328,10 @@ suite "Onchain group manager":
     let merkleRootAfter = manager.rlnInstance.getMerkleRoot().valueOr:
       raiseAssert $error
 
+    let metadataOpt = manager.rlnInstance.getMetadata().valueOr:
+      raiseAssert $error
     check:
+      metadataOpt.get().validRoots == manager.validRoots.toSeq()
       merkleRootBefore != merkleRootAfter
     await manager.stop()
 
@@ -481,13 +484,8 @@ suite "Onchain group manager":
     except Exception, CatchableError:
       assert false, "exception raised: " & getCurrentExceptionMsg()
 
-    await fut
+    check await fut.withTimeout(5.seconds)
 
-    let metadataOpt = manager.rlnInstance.getMetadata().valueOr:
-      raiseAssert $error
-    assert metadataOpt.isSome(), "metadata is not set"
-    check:
-      metadataOpt.get().validRoots == manager.validRoots.toSeq()
     await manager.stop()
 
   asyncTest "withdraw: should guard against uninitialized state":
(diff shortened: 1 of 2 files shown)


tests/waku_rln_relay/test_rln_group_manager_onchain.nim
```diff
@@ -481,13 +484,8 @@ suite "Onchain group manager":
     except Exception, CatchableError:
       assert false, "exception raised: " & getCurrentExceptionMsg()
 
-    await fut
+    check await fut.withTimeout(5.seconds)
```
Ivansete-status, line 487: ah, this is too slow. Isn't there a way to speed up the validation?

edited: okay, I guess the `5.seconds` wait is the worst scenary and it will fail the test in such case

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
```diff
@@ -591,6 +585,11 @@ proc getNewBlockCallback(g: OnchainGroupManager): proc =
     var handleBlockRes: bool
     g.retryWrapper(handleBlockRes, "Failed to handle new block"):
       await g.getAndHandleEvents(fromBlock, latestBlock)
+
+    let setMetadataRes = g.setMetadata()
+    if setMetadataRes.isErr():
+      error "failed to persist rln metadata", error = setMetadataRes.error
```
Ivansete-status, line 591: Maybe this work as well :)
```suggestion
    g.setMetadata().isOkOr:
      error "failed to persist rln metadata", error = $error
```

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
```diff
@@ -664,6 +663,9 @@ proc startOnchainSync(
       futs.add(g.getAndHandleEvents(fromBlock, toBlock))
       if futs.len >= maxFutures or toBlock == currentLatestBlock:
         await g.batchAwaitBlockHandlingFuture(futs)
+        let setMetadataRes = g.setMetadata(lastProcessedBlock = some(toBlock))
+        if setMetadataRes.isErr():
+          error "failed to persist rln metadata", error = setMetadataRes.error
```
Ivansete-status, line 668: Analogue to previous nitpick comment:
```suggestion
        g.setMetadata(lastProcessedBlock = some(toBlock)).isOkOr:
          error "failed to persist rln metadata", error = $error
```

Ivansete-status (approved):
LGTM! Thanks for it! :100:

gabrielmer (approved):
LGTM, thanks!

PR #2651, opened 2024-04-29
fix: waku_node - first of all stop the waku-switch when stopping the waku-node

Ivansete-status:
## Description
This is aimed to avoid having flaky tests by keeping the original invocation order that we had before applying the following change: https://github.com/waku-org/nwaku/pull/2645/files

image


gabrielmer (approved):
Thanks so much!

chaitanyaprem (approved):
Don't understand the details fully, but looks like this should be fine.

PR #2652, opened 2024-04-30
feat: sync use subranges

chaitanyaprem:
# Description
Use negentropy subrange which gives flexibility to choose sync time.
Still some optimizations are needed in negentropy side, but functionally this change has passed all existing tests.

@SionoiS SubRange seems to work with just timestamp alone, i think the ID argument may not be required then.

# Changes

- [x] Use a subrange instead of maintaining different storages
- [x] Use negentropy subrange wrappers

### DIFF
--- a/waku/waku_sync/raw_bindings.nim
+++ b/waku/waku_sync/raw_bindings.nim
@@ -119,20 +119,66 @@ proc raw_reconcile(
   negentropy: Negentropy, query: ptr Buffer, r: ptr BindingResult
 ): int {.header: NEGENTROPY_HEADER, importc: "reconcile".}
 
-# https://github.com/waku-org/negentropy/blob/d4845b95b5a2d9bee28555833e7502db71bf319f/cpp/negentropy_wrapper.h#L35
-proc free(
-  negentropy: Negentropy
-) {.header: NEGENTROPY_HEADER, importc: "negentropy_delete".}
-
 # https://github.com/waku-org/negentropy/blob/d4845b95b5a2d9bee28555833e7502db71bf319f/cpp/negentropy_wrapper.h#L51
 proc raw_reconcile_with_ids(
   negentropy: Negentropy, query: ptr Buffer, r: ptr BindingResult
 ): int {.header: NEGENTROPY_HEADER, importc: "reconcile_with_ids_no_cbk".}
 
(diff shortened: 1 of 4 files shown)


SionoiS:
> @SionoiS SubRange seems to work with just timestamp alone, i think the ID argument may not be required then.

Fantastic news! Makes our lives much easier.

waku/waku_sync/raw_bindings.nim
```diff
@@ -119,21 +119,67 @@ proc raw_reconcile(
...
+# https://github.com/waku-org/negentropy/blob/3044a30e4ba2e218aee6dee2ef5b4a4b6f144865/cpp/negentropy_wrapper.h#L69
+proc raw_reconcile_with_ids_subrange(
+  negentropy: NegentropySubRange, query: ptr Buffer, r: ptr BindingResult
+): int {.header: NEGENTROPY_HEADER, importc: "reconcile_with_ids_subrange_no_cbk".}
+
+# https://github.com/waku-org/negentropy/blob/3044a30e4ba2e218aee6dee2ef5b4a4b6f144865/cpp/negentropy_wrapper.h#L63
+proc free(
+  negentropy: NegentropySubRange
+) {.header: NEGENTROPY_HEADER, importc: "negentropy_subrange_delete".}
+
```
SionoiS, line 182: We can't avoid this duplication because of C++ templating ? If not we may be able to "hide" it with some kind of interface.

waku/waku_sync/raw_bindings.nim
```diff
@@ -283,3 +329,21 @@ proc clientReconcile*(
...
 
   return ok(some(NegentropyPayload(output)))
+
+### Subrange specific methods
+
+proc new*(
+    T: type SubRange,
+    storage: Storage,
+    startTime: int64 = 0,
+    endTime: int64, #TODO: Set default to MAX_UINT64
```
SionoiS, line 339:
```suggestion
    startTime: uint64 = uint43.low,
    endTime: uint64 = uint64.high
```

SionoiS (approved):

waku/waku_sync/raw_bindings.nim
chaitanyaprem: Yup, since wrapper is of C..we need to have separate methods for each type for template code.

PR #2658, opened 2024-05-02
refactor: big refactor to add waku component in libwaku instead of onlu waku node

Ivansete-status:
## Description
With this PR we:
  - rename the `app.nim` module to `waku.nim`
  - rename `App` type to `Waku` type. We start considering `Waku` as the type that contains `WakuNode`, discovery elements, and will contain an instance of `PeerManager` in the future. `Waku` type represents Waku ;P (very original name ;P)
  - refactor `libwaku` code so that we start dealing with the `Waku` object from within `libwaku` and therefore, start having a similar behaviour as the `wakunode2` app

## Issue

`migrate DiscV5 and DNS Discovery from app.nim to waku_node.nim` - https://github.com/waku-org/nwaku/issues/2452

### DIFF
--- a/waku/factory/waku.nim
+++ b/waku/factory/waku.nim
@@ -46,37 +46,31 @@ import
   ../../waku/factory/external_config
 
 logScope:
-  topics = "wakunode app"
+  topics = "wakunode waku"
 
 # Git version in git describe format (defined at compile time)
 const git_version* {.strdefine.} = "n/a"
 
-type
-  App* = object
-    version: string
-    conf: WakuNodeConf
-    rng: ref HmacDrbgContext
-    key: crypto.PrivateKey
+type Waku* = object
+  version: string
+  conf: WakuNodeConf
+  rng: ref HmacDrbgContext
+  key: crypto.PrivateKey
 
-    wakuDiscv5*: WakuDiscoveryV5
-    dynamicBootstrapNodes: seq[RemotePeerInfo]
+  wakuDiscv5*: WakuDiscoveryV5
+  dynamicBootstrapNodes: seq[RemotePeerInfo]
 
-    node: WakuNode
+  node*: WakuNode
 
-    restServer*: WakuRestServerRef
-    metricsServer*: MetricsHttpServerRef
+  restServer*: WakuRestServerRef
+  metricsServer*: MetricsHttpServerRef
 
-  AppResult*[T] = Result[T, string]
-
-func node*(app: App): WakuNode =
-  app.node
-
-func version*(app: App): string =
-  app.version
+func version*(waku: Waku): string =
+  waku.version
 
 ## Initialisation
 
-proc init*(T: type App, conf: WakuNodeConf): Result[App, string] =
+proc init*(T: type Waku, conf: WakuNodeConf): Result[Waku, string] =
   var confCopy = conf
   let rng = crypto.newRng()
 
@@ -103,7 +97,7 @@ proc init*(T: type App, conf: WakuNodeConf): Result[App, string] =
     error "Failed setting up node", error = nodeRes.error
     return err("Failed setting up node: " & nodeRes.error)
 
-  var app = App(
+  var waku = Waku(
     version: git_version,
     conf: confCopy,
     rng: rng,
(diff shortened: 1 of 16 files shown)


rymnc (commented):
small comment, there are tools/applications which may use the App* export, can we rename them too?

waku/waku_peer_exchange/protocol.nim
```diff
@@ -54,7 +54,7 @@ type
 
 proc request*(
     wpx: WakuPeerExchange, numPeers: uint64, conn: Connection
-): Future[WakuPeerExchangeResult[PeerExchangeResponse]] {.async, gcsafe.} =
+): Future[WakuPeerExchangeResult[PeerExchangeResponse]] {.async: (raises: []).} =
```
SionoiS, line 57: No longer gcsafe?

waku/waku_peer_exchange/protocol.nim
```diff
@@ -79,15 +79,18 @@ proc request*(
...
+    if connOpt.isNone():
+      return err(dialFailure)
+    return await wpx.request(numPeers, connOpt.get())
+  except CatchableError:
+    return err("exception dialing peer: " & getCurrentExceptionMsg())
 
 proc request*(
     wpx: WakuPeerExchange, numPeers: uint64
-): Future[WakuPeerExchangeResult[PeerExchangeResponse]] {.async, gcsafe.} =
+): Future[WakuPeerExchangeResult[PeerExchangeResponse]] {.async: (raises: []).} =
```
SionoiS, line 93: same here?

SionoiS (approved):
LGTM thank you!

SionoiS:
What about the make file? Needs to be changed too no?

waku/node/waku_node.nim
```diff
@@ -1063,7 +1063,7 @@ proc mountPeerExchange*(node: WakuNode) {.async, raises: [Defect, LPError].} =
 
 proc fetchPeerExchangePeers*(
     node: Wakunode, amount: uint64
-): Future[Result[int, string]] {.async, raises: [Defect].} =
+): Future[Result[int, string]] {.async: (raises: []).} =
```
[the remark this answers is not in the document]
Ivansete-status: > This proc did not actual raised Defect before?

I don't know :) In this case, we are forcing that no exception will get raised from that proc.

Ivansete-status:
> What about the make file? Needs to be changed too no?

If you refer to `wakunode2` I think we can't change that easily and would require coordination with infra, and Dockerfile changes too.

waku/factory/waku.nim
```diff
@@ -132,9 +126,9 @@ proc getPorts(
 
   return ok((tcpPort: tcpPort, websocketPort: websocketPort))
 
-proc getRunningNetConfig(app: App): AppResult[NetConfig] =
-  var conf = app.conf
-  let (tcpPort, websocketPort) = getPorts(app.node.switch.peerInfo.listenAddrs).valueOr:
+proc getRunningNetConfig(waku: ptr Waku): Result[NetConfig, string] =
```
jm-clius, line 129: It's hard for me to see why in this module we have to switch to operations on `ptr`s. It causes some uncomfortable integrations into higher layer applications: `startWaku(addr waku)` is quite a bit less intuitive than `waku.startWaku()`. My assumption is that this is necessary for the thread management in the bindings? However, wouldn't it be better to wrap this into another `ptr` operation layer specifically (and only) for the bindings? May be missing some fundamental limitation here.

jm-clius (approved):
Approving, as I understand the general direction taken here. However, have asked a question on switching to `ptr` operations below.

waku/factory/waku.nim
Ivansete-status: Thanks for the comment @jm-clius ! :raised_hands: 

I agree that it becomes more awkward to use with `startWaku(addr waku)`

Let me elaborate more on why this is needed:
- We should make the `startWaku` proc async because the `waitFor` statement should be only invoked from the highest level to avoid deadlocks. i.e. the `waitFor` should only be invoked from within the `wakunode2` code or highest level of `libwaku`.

   It is not correct to have a `waitFor` as we had before this PR. See:
https://github.com/waku-org/nwaku/blob/853ec1869e950f6a0965debc6bb5cb41570a0615/waku/factory/app.nim#L177-L179

- The  `startWaku` (formerly known as `startApp`) changes the state of the given parameter, `waku`, but we cannot longer pass a `var` parameter because the `startWaku` should be `async` (commented in the previous point) and it doesn't compile ( `<<cannot be captured as it would violate memory safety>>` .)

- Given that we cannot use `var`, the only way to allow changing the state of a variable in an `async` proc, is by using `ptr` type, which is considered unsafe but in our case we can make sure that the passed `Waku` instance has longer lifetime than the call of `startApp` and it won't crash because of incorrect address being used inside the `startApp`.

Nevertheless, there might be a more elegant approach ( cc @arnetheduck :)

PR #2664, opened 2024-05-03
chore(rln-relay): health check should account for window of roots

rymnc:
# Description

This PR modifies the `ready` behaviour of rln-relay's onchain mode. If the node is behind the tip of the blockchain by `window` blocks, it should still be marked as ready IMO.

# Changes

- [x] returns `ready` if the node is behind the tip of the blockchain by `window` blocks

### DIFF
--- a/waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
+++ b/waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
@@ -586,7 +586,7 @@ proc getNewBlockCallback(g: OnchainGroupManager): proc =
     g.retryWrapper(handleBlockRes, "Failed to handle new block"):
       await g.getAndHandleEvents(fromBlock, latestBlock)
 
-    # cannot use isOkOr here because results in a compile-time error that 
+    # cannot use isOkOr here because results in a compile-time error that
     # shows the error is void for some reason
     let setMetadataRes = g.setMetadata()
     if setMetadataRes.isErr():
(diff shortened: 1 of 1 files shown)


waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
```diff
@@ -855,7 +855,8 @@ method isReady*(g: OnchainGroupManager): Future[bool] {.async.} =
   g.retryWrapper(currentBlock, "Failed to get the current block number"):
     cast[BlockNumber](await g.ethRpc.get().provider.eth_blockNumber())
 
-  if g.latestProcessedBlock < currentBlock:
+  # the node is still able to process messages if it is behind the latest block by a factor of the valid roots
+  if uint(g.latestProcessedBlock) < uint(currentBlock) - uint(g.validRoots.len):
```
Ivansete-status, line 859: I wonder if is safer to add parenthesis:
```suggestion
  if uint(g.latestProcessedBlock) < ( uint(currentBlock) - uint(g.validRoots.len) ):
```

Ivansete-status (approved):
LGTM! Thank! :100:

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
alrevuelta, line 859: isn't it better widening rather than narrowing? i mean instead of casting everything to unit and operating with that, using the `latestProcessedBlock` type which has more bits and then "widening" validRoots to do the `-`?

wondering also if in the first run we can hit `currentBlock-g.validRoots.len` where `currentBlock=0` and since we are using unit that will lead to a huge value, meaning that we are ready when we are not?

alrevuelta: doesn't harm, i would add it as well :)

rymnc: currentBlock will never be 0, since that is fetched from the chain (and we can operate under the assumption that it will always be greater than 0). good point about widening vs narrowing

rymnc: casted to `u256` - 442ca43a

alrevuelta (approved):

PR #2665, opened 2024-05-03
fix(filter): log is too large

rymnc:
# Description

the filter protocol logs out the whole message, this is bad

image

identified this is the offending log -
image

# Changes

- [x] removes the message  from the "handling message" filter log


Ivansete-status (approved):
LGTM! Thank you very much! :100:

gabrielmer (approved):
Thank you!

PR #2668, opened 2024-05-03
feat: Added flexible rate limit checks for store, legacy store and lightpush

NagyZoltanPeter:
Also added rate and traffic metrics.

# Description

This PR addresses most of the weakness of phase1 implementation and extend it with bandwidth metrics for the protocols.
For convenience it fixes some nuances on other aspects.

# Changes

- [X] Added template for ease and generalize the use of tokenbucket.
- [X] Rate check is done before anything would happen to request lib2p2 stream (no read/copy/decode)
- [X] Added counting of the individual protocols in/outbound bandwidths.
- [X] Renamed legacy store codec global const to avoid misunderstanding
- [X] Fixed missing legacy store protocol in GET /admin/peers query 

# Missing

Will be applied in a separate PR:
- Rate limiting Filter
   - It might need a bit different approach due to its nature of subscribe and ping
- Trace top contributor peers of over limit use and ban them 

## Issue
https://github.com/waku-org/nwaku/issues/2589
https://github.com/waku-org/nwaku/issues/2032

### DIFF
--- a/tests/waku_store_legacy/test_resume.nim
+++ b/tests/waku_store_legacy/test_resume.nim
@@ -1,332 +1,340 @@
 {.used.}
 
-import
-  std/[options, tables, sets],
-  testutils/unittests,
-  chronos,
-  chronicles,
-  libp2p/crypto/crypto
-import
-  ../../waku/common/databases/db_sqlite,
-  ../../waku/waku_archive/driver,
-  ../../waku/waku_archive/driver/sqlite_driver/sqlite_driver,
-  ../../waku/node/peer_manager,
-  ../../waku/waku_core,
-  ../../waku/waku_core/message/digest,
-  ../../waku/waku_store_legacy,
-  ../waku_store_legacy/store_utils,
-  ../waku_archive/archive_utils,
-  ./testlib/common,
-  ./testlib/switch
+when defined(waku_exp_store_resume):
+  # TODO: Review store resume test cases (#1282)
+  # Ongoing changes to test code base had ruin this test meanwhile, need to investigate and fix
 
-procSuite "Waku Store - resume store":
-  ## Fixtures
-  let storeA = block:
-    let store = newTestMessageStore()
-    let msgList =
-      @[
-        fakeWakuMessage(
-          payload = @[byte 0], contentTopic = ContentTopic("2"), ts = ts(0)
-        ),
-        fakeWakuMessage(
-          payload = @[byte 1], contentTopic = ContentTopic("1"), ts = ts(1)
-        ),
-        fakeWakuMessage(
-          payload = @[byte 2], contentTopic = ContentTopic("2"), ts = ts(2)
-        ),
-        fakeWakuMessage(
-          payload = @[byte 3], contentTopic = ContentTopic("1"), ts = ts(3)
-        ),
-        fakeWakuMessage(
-          payload = @[byte 4], contentTopic = ContentTopic("2"), ts = ts(4)
-        ),
-        fakeWakuMessage(
-          payload = @[byte 5], contentTopic = ContentTopic("1"), ts = ts(5)
-        ),
-        fakeWakuMessage(
-          payload = @[byte 6], contentTopic = ContentTopic("2"), ts = ts(6)
-        ),
-        fakeWakuMessage(
-          payload = @[byte 7], contentTopic = ContentTopic("1"), ts = ts(7)
-        ),
-        fakeWakuMessage(
-          payload = @[byte 8], contentTopic = ContentTopic("2"), ts = ts(8)
-        ),
-        fakeWakuMessage(
-          payload = @[byte 9], contentTopic = ContentTopic("1"), ts = ts(9)
-        ),
-      ]
+  import
+    std/[options, tables, sets],
+    testutils/unittests,
+    chronos,
+    chronicles,
+    libp2p/crypto/crypto
+  import
+    ../../waku/common/databases/db_sqlite,
+    ../../waku/waku_archive/driver,
+    ../../waku/waku_archive/driver/sqlite_driver/sqlite_driver,
+    ../../waku/node/peer_manager,
+    ../../waku/waku_core,
+    ../../waku/waku_core/message/digest,
+    ../../waku/waku_store_legacy,
+    ../waku_store_legacy/store_utils,
+    ../waku_archive/archive_utils,
+    ./testlib/common,
+    ./testlib/switch
 
-    for msg in msgList:
-      require store
-      .put(
-        DefaultPubsubTopic,
-        msg,
-        computeDigest(msg),
-        computeMessageHash(DefaultPubsubTopic, msg),
-        msg.timestamp,
+  procSuite "Waku Store - resume store":
+    ## Fixtures
+    let storeA = block:
+      let store = newTestMessageStore()
+      let msgList =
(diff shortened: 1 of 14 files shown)


waku/common/ratelimit.nim
```diff
@@ -23,3 +23,28 @@ proc newTokenBucket*(setting: Option[RateLimitSetting]): Option[TokenBucket] =
...
+  if not tokenBucket.tryConsume(1):
+    return false
+
+  return true
+
+template checkUsageLimit*(
+    t: var Option[TokenBucket],
+    proto: string,
+    conn: Connection,
+    bodyWithinLimit, bodyRejected: untyped,
```
gabrielmer, line 43: First time y encounter a template with two bodies. Want to make sure I read it correctly, as I find it a bit confusing.

First body goes goes after the `:` as usual and the second in a `do:` block?

NagyZoltanPeter: Exactly. I got the concept from nim std Table withValue template. https://nim-lang.org/docs/tables.html#withValue.t%2CTable%5BA%2CB%5D%2CA%2Cuntyped%2Cuntyped%2Cuntyped
as it perfectly suites the situation.

So first block is success scenario where request is within the limit. The `do` block will catch the over limit situation and allow to tailor the response buffer properly.

The issue was that different protocols have different response approaches (unfortunately) and needed to make it easy to define but still keep some consistent usage of the limit handling as it shall always happen before reading message into our buffer and decode it.

gabrielmer (approved):
Loved this PR! 😍  Thanks so much!

Now I noticed that we don't really write tests to make sure counters work correctly. Maybe it would be good idea to start doing it in the future, counters are underrated IMO :))

waku/waku_store/protocol.nim
```diff
@@ -99,15 +84,33 @@ proc handleQueryRequest*(
   return res.encode().buffer
 
 proc initProtocolHandler(self: WakuStore) =
+  let rejectReposnseBuffer = StoreQueryResponse(
+    ## We will not copy and decode RPC buffer from stream only for requestId
+    ## in reject case as it is comparably too expensive and opens possible
+    ## attack surface
+    requestId: "N/A",
+    statusCode: uint32(ErrorCode.TOO_MANY_REQUESTS),
+    statusDesc: $ErrorCode.TOO_MANY_REQUESTS,
```
Ivansete-status, line 93: I repeated the comment below. I wonder if we could enrich the response by indicating what is the actual limit

waku/waku_store_legacy/protocol.nim
```diff
@@ -44,77 +44,103 @@ type WakuStore* = ref object of LPProtocol
...
+proc initProtocolHandler(ws: WakuStore) =
+  let rejectResponseBuf = HistoryRPC(
+    ## We will not copy and decode RPC buffer from stream only for requestId
+    ## in reject case as it is comparably too expensive and opens possible
+    ## attack surface
+    requestId: "N/A",
+    response: some(
+      HistoryResponseRPC(
+        error: HistoryError(kind: HistoryErrorKind.TOO_MANY_REQUESTS).toRPC()
+      )
```
Ivansete-status, line 107: An analogue to previous comments. It would be interesting to have a better detail about why the limit is exceeded, i.e. what is the current rate and what it the rate-limit.

waku/waku_store_legacy/protocol.nim
```diff
@@ -44,77 +44,103 @@ type WakuStore* = ref object of LPProtocol
...
+    ),
+  ).encode().buffer
+
+  proc handler(conn: Connection, proto: string) {.async, gcsafe, closure.} =
```
Ivansete-status, line 111: `async` implies `gcsafe`
```suggestion
  proc handler(conn: Connection, proto: string) {.async, closure.} =
```

Ivansete-status (approved):
LGTM! Super professional PR and a great use of templates :partying_face: 
I just added some comments that I hope you find useful

waku/common/ratelimit.nim
```diff
@@ -23,3 +23,28 @@ proc newTokenBucket*(setting: Option[RateLimitSetting]): Option[TokenBucket] =
     return none[TokenBucket]()
 
   return some(TokenBucket.new(volume, period))
+
+proc checkUsage*(
```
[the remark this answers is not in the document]
NagyZoltanPeter: Ah, thanks, it was a leftover from an iteration...

waku/common/ratelimit.nim
```diff
@@ -23,3 +23,28 @@ proc newTokenBucket*(setting: Option[RateLimitSetting]): Option[TokenBucket] =
...
+    t: var Option[TokenBucket], proto: string, conn: Connection
+): bool {.raises: [Defect].} =
```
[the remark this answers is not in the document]
NagyZoltanPeter: I need to check this. Thx!

waku/waku_store/protocol.nim
```diff
@@ -99,15 +84,33 @@ proc handleQueryRequest*(
...
 
-    let resBuf = await self.handleQueryRequest(conn.peerId, reqBuf)
+      waku_service_inbound_network_bytes.inc(
+        amount = reqBuf.len().int64, labelValues = [WakuStoreCodec]
+      )
+
+      resBuf = await self.handleQueryRequest(conn.peerId, reqBuf)
+    do:
+      debug "store query request rejected due rate limit exceeded", peerId = conn.peerId
+      resBuf = rejectReposnseBuffer
```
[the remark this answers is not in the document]
NagyZoltanPeter: Well maybe... but I think it worth only when starting up and within logging configuration. I don't think it is necessary for every log when rate limit exceeded.

waku/waku_store/protocol.nim
NagyZoltanPeter: Do you mean in the response for the client? I don't want to expose it to any malicious peer to take advantage on it. WDYT?

waku/waku_store_legacy/protocol.nim
NagyZoltanPeter: I think this detail should not be expose to the outer world.

tests/waku_store_legacy/test_resume.nim
```diff
@@ -1,332 +1,340 @@
...
-
-    for msg in msgList:
-      require store
-      .put(
-        DefaultPubsubTopic,
-        msg,
-        computeDigest(msg),
-        computeMessageHash(DefaultPubsubTopic, msg),
-        msg.timestamp,
+when defined(waku_exp_store_resume):
```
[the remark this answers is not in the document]
NagyZoltanPeter: Ah, I just applied the same compile flag here, due it was excluded only on all_test_waku level but not in waku_store_legacy/test_all.nim... it caused me trouble when was running specific protocol tests instead of all_test_waku.
So github compara a bit stupid on this as only when(...): conditional compile was added, all other are just indentation change.

waku/waku_store/protocol.nim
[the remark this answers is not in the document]
NagyZoltanPeter: Added limit print to the debug logs.

waku/common/ratelimit.nim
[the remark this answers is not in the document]
NagyZoltanPeter: Applied

PR #2671, opened 2024-05-06
feat: Added message size check before relay for lightpush

shash256:
# Description
Added setup for message validation for lightpush service nodes to catch validation failures before relay. Fixes issue https://github.com/waku-org/nwaku/issues/2595

# Changes
- Updates to lighpush protocol
- Message validation in REST API
- Unit test to verify the functionality

### DIFF
--- a/migrations/message_store/00009_addMetaColumn.up.sql
+++ b/migrations/message_store/00009_addMetaColumn.up.sql
@@ -1,0 +1,1 @@
+ALTER TABLE message ADD COLUMN meta BLOB;
--- a/migrations/message_store_postgres/content_script_version_4.nim
+++ b/migrations/message_store_postgres/content_script_version_4.nim
@@ -1,70 +1,8 @@
 const ContentScriptVersion_4* =
   """
-ALTER TABLE IF EXISTS messages_backup RENAME TO messages;
-ALTER TABLE messages RENAME TO messages_backup;
-ALTER TABLE messages_backup DROP CONSTRAINT messageIndex;
+ALTER TABLE messages ADD meta VARCHAR default null;
 
-CREATE TABLE IF NOT EXISTS messages (
-   pubsubTopic VARCHAR NOT NULL,
-   contentTopic VARCHAR NOT NULL,
-   payload VARCHAR,
-   version INTEGER NOT NULL,
-   timestamp BIGINT NOT NULL,
-   id VARCHAR NOT NULL,
-   messageHash VARCHAR NOT NULL,
-   storedAt BIGINT NOT NULL,
-   meta VARCHAR,
-   CONSTRAINT messageIndex PRIMARY KEY (messageHash, storedAt)
-  ) PARTITION BY RANGE (storedAt);
-
-DO $$
-DECLARE
-     min_storedAt numeric;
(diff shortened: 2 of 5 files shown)


waku/waku_lightpush/protocol.nim
```diff
@@ -26,6 +26,25 @@ type WakuLightPush* = ref object of LPProtocol
   pushHandler*: PushMessageHandler
   requestRateLimiter*: Option[TokenBucket]
 
+proc validateMessage*(
```
jm-clius, line 29: Ah, the idea is to use the exact same validation function that Relay would use when it publishes the message - not create a new validation function (which would have to be maintained separately and would invariably at some point miss validation edge cases enforced by the relay validator).
In other words, we should use: https://github.com/waku-org/nwaku/blob/b46226fca0fbc5730cbf3c9899e33fac8b817796/waku/waku_relay/protocol.nim#L225-L252

Of course, that requires a reference to the instantiated `wakuRelay` object that will eventually publish this message. This is currently not available in the lightpush module, but in the pushHandler provided to lightpush by the waku node here: https://github.com/waku-org/nwaku/blob/b46226fca0fbc5730cbf3c9899e33fac8b817796/waku/node/waku_node.nim#L917-L929

Architecturally, I think this indirection created by the pushHandler is unnecessary. I'll add a comment on this for future work, but for now I think this PR should be as simple as adding:
```
node.wakuRelay.validateMessage(
```
in the push handler before pushing the message (and testing)

waku/waku_api/rest/lightpush/handlers.nim
jm-clius, line 1: If validation is done in the `pushHandler` (see my comment elsewhere), changes to this file is likely unnecessary.

jm-clius (changes requested):
Congratulations on your first PR! Great to see that you're already getting the hang of Nim language and syntax. Note my comment below - I think it simplifies what needs to be done in this PR quite significantly. However, this did point out an unnecessary architectural complexity introduced by the `pushHandler` pattern.
My suggestion:
- simplify this PR by adding Relay validation in the `pushHandler` as described in my comment below
- for the RLN-proofs-as-a-service work, first remove the `pushHandler` pattern by adding a reference to `wakuRelay` into the Lightpush module and moving all push logic to this module itself. Let me know if this next step is unclear.

waku/waku_lightpush/protocol.nim
NagyZoltanPeter: @shash256 and @jm-clius : I think the biggest issue with lightpush is not that it does not pre-validates the message but in contrary it will not get noticed by the relay that the message validation done in relay protocol side is failed. It cause that we do not have meaningful answer to the lightpush client that would help error handling.
WDYT?

jm-clius: Well, if we validate the message before attempting to relay it and return an explicit "validation error" to the client on failure, wouldn't it cover - or at least improve - this silent failure? Similar to how we return a "bad request" error to Relay API clients in REST: https://github.com/waku-org/nwaku/blob/master/waku/waku_api/rest/relay/handlers.nim#L163-L164

PR #2672, opened 2024-05-06
fix: content_script_version_4.nim: migration failed when dropping unexisting constraing

Ivansete-status:
## Description
There appeared _Postgres_ schema migration issue when going from schema version 3 to 4.

The problem occurred while trying to remove an index that doesn't seem to exist in version 3.

Aside from that, I'm adding the following line, which is part of the script for version 3, to make sure that index is created in case it doesn't exist. I've notice that when testing in `waku.test`:
`CREATE INDEX IF NOT EXISTS i_query ON messages (contentTopic, pubsubTopic, storedAt, id);`

### DIFF
--- a/migrations/message_store_postgres/content_script_version_4.nim
+++ b/migrations/message_store_postgres/content_script_version_4.nim
@@ -63,6 +63,8 @@ INSERT INTO messages (
                        storedAt
                    FROM messages_backup;
 
+CREATE INDEX IF NOT EXISTS i_query ON messages (contentTopic, pubsubTopic, storedAt, id);
+
 DROP TABLE messages_backup;
 
 UPDATE version SET version = 4 WHERE version = 3;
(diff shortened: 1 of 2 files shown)


SionoiS:
I don't understand. The logs seams to suggest that the error happen because `messages` table already existed.

I don't see how this PR would fix that?

Maybe renaming `messages` if it exist to `messages_backup`, instead of the current first 2 steps.

NagyZoltanPeter:
> I don't understand. The logs seams to suggest that the error happen because `messages` table already existed.
> 
> I don't see how this PR would fix that?
> 
> There's 4 cases here, `messages` and `messages_backup` both can either exist or not. If we don't handle all cases then it can fail.

Yeah, I agree with SP, the issue here that the migration failed at a stage when both table existed. Dropping the constraint will not solve this. 
Or do you think the original issue with the migration happened because the existed constraint?

Ivansete-status:
Yes sorry for the confusion guys. I will unset you from reviewers until I have a better fix.
Thanks for the comments!

SionoiS (approved):
Based on your description of the problem this seams good.

Thanks

gabrielmer (approved):
LGTM, thanks so much! 😍

PR #2676, opened 2024-05-07
feat: store v3 return pubsub topics

SionoiS:
This PR is meant to unblock Waku Sync work.

Before Store v3 did not return the pubsub topics messages were sent on, now it does.

Follows updated https://github.com/vacp2p/rfc/pull/665/commits/5d69ab58b34d32af99e60b109ad784ad1dd9e345

### DIFF
--- a/waku/waku_archive/common.nim
+++ b/waku/waku_archive/common.nim
@@ -56,6 +56,7 @@ type
   ArchiveResponse* = object
     hashes*: seq[WakuMessageHash]
     messages*: seq[WakuMessage]
+    topics*: seq[PubsubTopic]
     cursor*: Option[ArchiveCursor]
 
   ArchiveErrorKind* {.pure.} = enum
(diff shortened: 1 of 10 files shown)


waku/waku_archive/common.nim
```diff
@@ -56,6 +56,7 @@ type
   ArchiveResponse* = object
     hashes*: seq[WakuMessageHash]
     messages*: seq[WakuMessage]
+    topics*: seq[PubsubTopic]
```
jm-clius, line 59: Mmm. Wouldn't it be better to return seq[(message, topic)] to keep the 1:1 association explicit?

jm-clius (approved):
Approving, but I do think the comment below should be addressed.

gabrielmer (approved):
LGTM 🤩

SionoiS:
We got a segfault in CI but i can't reproduce locally. :|

SionoiS:
Will merge even with CI red as it's unrelated and the issue is being investigated.

PR #2686, opened 2024-05-08
fix: adding wait after starting node to avoid segfault

gabrielmer:
## Description
Temporary workaround to avoid having flaky tests in MacOS


Ivansete-status (approved):
LGTM! Thanks for it!

NagyZoltanPeter (approved):
Nice catch!!!

PR #2687, opened 2024-05-09
chore: Postgres enhance get oldest timestamp

Ivansete-status:
## Description
This PR enhances the case where, for any reason, there are old empty partitions. In that case of empty partitions, they are not considered when getting the oldest timestamp. Instead, the oldest timestamp was only considered from the oldest row.

Therefore, when "time" retention policy is applied, we should also consider those old-empty partitions so that they can get erased properly.

## Issue
closes https://github.com/waku-org/nwaku/issues/2689

### DIFF
--- a/waku/waku_archive/driver/postgres_driver/postgres_driver.nim
+++ b/waku/waku_archive/driver/postgres_driver/postgres_driver.nim
@@ -730,7 +730,17 @@ method getOldestMessageTimestamp*(
   if intRes.isErr():
     return err("error in getOldestMessageTimestamp: " & intRes.error)
 
-  return ok(Timestamp(intRes.get()))
+  let oldestPartition = s.partitionMngr.getOldestPartition().valueOr:
+    return err("could not get oldest partition: " & $error)
+
+  let oldestPartitionTimeNanoSec =
+    Timestamp(oldestPartition.getPartitionStartTimeInNanosec())
+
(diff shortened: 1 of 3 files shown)


waku/waku_archive/driver/postgres_driver/postgres_driver.nim
```diff
@@ -701,7 +701,10 @@ proc getInt(
   try:
     retInt = parseInt(str)
   except ValueError:
-    return err("exception in getInt, parseInt: " & getCurrentExceptionMsg())
+    return err(
+      fmt"exception in getInt, parseInt, str:[{str}] query:[{query}]: " &
```
SionoiS, line 705: FYI `fmt` can raise and prevent using `raises: []` on procs.

SionoiS (approved):
LGTM thanks!

NagyZoltanPeter (approved):
LGTM!

gabrielmer (approved):
Thank you!

waku/waku_archive/driver/postgres_driver/postgres_driver.nim
Ivansete-status: > FYI `fmt` can raise and prevent using `raises: []` on procs.

Thanks for the comment! I fixed that even though I cannot link to the actual commit (I had to `force-push` as I rebased from origin.)

PR #2688, opened 2024-05-09
feat: message transfer mechanism

SionoiS:
Add a callback to query the store of the other node for the missing messages and then add them to our own.

### DIFF
--- a/waku/node/waku_node.nim
+++ b/waku/node/waku_node.nim
@@ -223,6 +223,7 @@ proc mountWakuSync*(
             none(ArchiveCursor)
 
         let query = ArchiveQuery(
+          includeData: true,
           cursor: archiveCursor,
           startTime: some(pruneStart),
           endTime: some(pruneStop),
--- a/waku/node/waku_node.nim
+++ b/waku/node/waku_node.nim
@@ -197,51 +197,62 @@ proc connectToNodes*(
 
 ## Waku Sync
 
-proc mountWakuSync*(node: WakuNode): Result[void, string] =
+proc mountWakuSync*(
+    node: WakuNode,
+    maxFrameSize: int = DefaultMaxFrameSize,
+    syncInterval: timer.Duration = DefaultSyncInterval,
+    relayJitter: Duration = DefaultGossipSubJitter,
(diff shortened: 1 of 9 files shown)


waku/node/waku_node.nim
```diff
@@ -243,8 +243,49 @@ proc mountWakuSync*(node: WakuNode): Result[void, string] =
 
     return ok((elements, cursor))
 
-  #TODO add sync callback and options
-  node.wakuSync = WakuSync.new(peerManager = node.peerManager, pruneCB = some(prune))
+  let transfer: TransferCallback = proc(
```
jm-clius, line 246: I'm not a huge fan of us extending the `waku_node` module with (more) protocol integration logic. To me it seems like we're mounting a new type of protocol - `mountWakuStoreSync()` if you will. `WakuStoreSync` contains a reference to the local `wakuStoreClient`, `wakuArchive` and `wakuSync` and handles the coordination logic between these layers. I imagine this coordination will even be extended in future (e.g. with configuration checks to ensure the archive and sync stores are reasonably dimensioned). WDYT?

jm-clius (approved):
LGTM in terms of functionality, though I would urge abstracting the coordination logic away from the node module.

waku/node/waku_node.nim
SionoiS: I don't like it either but a "component" (like `WakuStoreSync`) referencing other components is a BIG no-no in my book and architecturally we don't have hierarchies of components in nwaku and that's a good thing!

I think we have to tolerate a `waku_node` like this until we have proper channels implemented in chronos then we can move away from callbacks, get rid of `waku_node` and move to a channel based architecture a la Go or Rust (multi-threading for "free" when chronos impl. it would be another benefit). The fact that `waku_node` cannot be implemented in Rust should tell you that it's a :foot: :gun:. The only reason `waku_node` can exist is because of GC magic and multi-threading not being designed for in nwaku.

I believe that whatever functionality we want from `WakuStoreSync` can be achieved by coordinating components and proper configuration.

SionoiS:
basic E2E test done!

PR #2690, opened 2024-05-09
fix: use await instead of waitFor in async tests

gabrielmer:
# Description

In async functions, we should use `await` instead of `waitFor`. The usage of `waitFor` in async tests can generate crashes.

# Changes

- [x] change use of `waitFor` for `await` in async tests

### DIFF
--- a/tests/node/peer_manager/test_peer_manager.nim
+++ b/tests/node/peer_manager/test_peer_manager.nim
@@ -34,10 +34,12 @@ suite "Peer Manager":
     asyncTest "light client is not disconnected":
       # Given two nodes with the same shardId
       let
-        server =
-          newTestWakuNode(serverKey, listenAddress, listenPort, pubsubTopics = @[shardTopic0])
-        client =
-          newTestWakuNode(clientKey, listenAddress, listenPort, pubsubTopics = @[shardTopic1])
+        server = newTestWakuNode(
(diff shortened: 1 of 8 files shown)


AlejandroCabeza (approved):

Ivansete-status (approved):
LGTM! Thanks! :100:

SionoiS (approved):
Very nice!

chaitanyaprem:
Interesting, What is the reason for crash? Is there some bug in waitFor macro or the way use it?

NagyZoltanPeter (approved):
Sounds very interesting, would be good understand the reason! Thanks!

gabrielmer:
So in principle,`waitFor` should not be used in `async` procedures as it blocks the current thread and it defeats the purpose of async procedures.

Even on its implementation there's a comment stating not to be called from an `async` proc

https://github.com/status-im/nim-chronos/blob/8a306763cec8105fa83574b56734b0f66823f844/chronos/internal/asyncfutures.nim#L628-L638

Regarding the crash we were having, I haven't yet figured out what exactly was going on that caused it. There were other tests with `waitFor` being used inside `async` procedures that did not crash.

In our case, we figured out that the crash happened when trying to write to global variables right after the `waitFor`, but not sure yet why exactly that's a problem.

PR #2692, opened 2024-05-10
chore: closing ping streams

### LINKED ISSUE #2462: chore: close stale ping connections
### Background
A node pings its peers every 2 minutes in https://github.com/waku-org/nwaku/blob/56ff30ca75100b06cb5157ae2655092b3051ec8c/waku/node/waku_node.nim#L1072-L1073
However the connections are never closed

We rely on libp2p's default connection limits. For example
```proc withMplex*(b: SwitchBuilder, inTimeout = 5.minutes, outTimeout = 5.minutes, maxChannCount = 200): SwitchBuilder```

If that's not set by libp2p (what has happened us with yamux), the number of connections grows indefinitely.
We should close unused connections ourselves.

### Acceptance criteria
Unused connections are deliberately closed

gabrielmer:
# Description

We're opening ping streams every 2 minutes but not closing them. This means we get an ever increasing amount of stale streams. Therefore, closing the stream after the ping.

# Changes

- [x] close ping stream after usage

## Testing
Tested it by calling https://github.com/waku-org/nwaku/blob/a37c9ba91158a90879a9e78f124a6e11e1900279/waku/node/peer_manager/peer_manager.nim#L655-L671 and https://github.com/waku-org/nwaku/blob/a37c9ba91158a90879a9e78f124a6e11e1900279/waku/node/peer_manager/peer_manager.nim#L673-L685

inside `keepaliveLoop()` and logging the outputs

The conclusions were:
- without calling `conn.close()`, the output of `getNumStreams()` grew indefinitely
- calling `conn.close()`, the output of `getNumStreams()` varied but was stable
- calling `conn.close()`did not affect the output of `connectedPeers()`, which means that indeed it didn't affect the physical connections

## Issue

closes #2462

### DIFF
--- a/waku/node/waku_node.nim
+++ b/waku/node/waku_node.nim
@@ -1132,7 +1132,6 @@ proc keepaliveLoop(node: WakuNode, keepalive: chronos.Duration) {.async.} =
     let peers =
       node.peerManager.peerStore.peers().filterIt(it.connectedness == Connected)
 
-      node.peerManager.connectedPeers(PingCodec)
     for peer in peers:
       try:
         let conn = await node.switch.dial(peer.peerId, peer.addrs, PingCodec)
--- a/waku/node/waku_node.nim
+++ b/waku/node/waku_node.nim
@@ -1136,13 +1136,18 @@ proc keepaliveLoop(node: WakuNode, keepalive: chronos.Duration) {.async.} =
     let peers =
       node.peerManager.peerStore.peers().filterIt(it.connectedness == Connected)
 
+    echo "----------- physical connections before: ",
+      node.peerManager.connectedPeers(PingCodec)
     for peer in peers:
       try:
         let conn = await node.switch.dial(peer.peerId, peer.addrs, PingCodec)
         let pingDelay = await node.libp2pPing.ping(conn)
+        await conn.close()
       except CatchableError as exc:
         waku_node_errors.inc(labelValues = ["keep_alive_failure"])
 
+    echo "----------- physical connections after: ",
+      node.peerManager.connectedPeers(PingCodec)
     await sleepAsync(keepalive)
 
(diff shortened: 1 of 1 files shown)


NagyZoltanPeter (approved):
Good catch. Isn't this feature is disabled by default?

gabrielmer:
> Good catch. Isn't this feature is disabled by default?

Yes! So the feature is disabled by default, but in `nwaku-compose` we do start the node with `keep-alive=true`

AlbertoSoutullo:
> Yes! So the feature is disabled by default, but in `nwaku-compose` we do start the node with `keep-alive=true`

Wait, one question so I can understand this. Are you saying that this behavior were happening only if we had the `keep-alive` flag set to True?

alrevuelta (approved):
lgtm.

Was wondering why we were not seeing this in `waku_streams_peers` but just realized we only consider [waku protocols](https://github.com/waku-org/nwaku/blob/master/waku/node/peer_manager/waku_peer_store.nim#L79) (ping is excluded).

good catch

gabrielmer:
> Wait, one question so I can understand this. Are you saying that this behavior were happening only if we had the `keep-alive` flag set to True?

Yes, if `keep-alive` is set to false then we wouldn't open those ping connections every 2 minutes that make the open connections to grow indefinitely

Ivansete-status (approved):
LGTM! Thanks! :100: 
As a possible future optimization we should only `dial` once at the beginning and then keep using that connection on further pings

AlbertoSoutullo:
> Yes, if `keep-alive` is set to false then we wouldn't open those ping connections every 2 minutes that make the open connections to grow indefinitely

Then I will need to doublecheck this in our simulations, because we are not using that flag...

PR #2695, opened 2024-05-13
feat: Added message size check before relay for lightpush

shash256:
# Description
Added setup for message validation for lightpush service nodes to catch validation failures before relay. Fixes issue https://github.com/waku-org/nwaku/issues/2595 . Followup to PR https://github.com/waku-org/nwaku/pull/2671 .

# Changes
- Updates to lighpush protocol
- Message validation in REST API
- Unit test to verify the functionality

### DIFF
--- a/waku/node/waku_node.nim
+++ b/waku/node/waku_node.nim
@@ -920,17 +920,17 @@ proc mountLightPush*(
       let validationRes = await node.wakuRelay.validateMessage(pubSubTopic, message)
       if validationRes.isErr():
         return err(validationRes.error)
-      else:
-        let publishedCount =
-          await node.wakuRelay.publish(pubsubTopic, message.encode().buffer)
 
-        if publishedCount == 0:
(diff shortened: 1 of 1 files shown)


waku/node/waku_node.nim
```diff
@@ -917,16 +917,20 @@ proc mountLightPush*(
     pushHandler = proc(
         peer: PeerId, pubsubTopic: string, message: WakuMessage
     ): Future[WakuLightPushResult[void]] {.async.} =
-      let publishedCount =
-        await node.wakuRelay.publish(pubsubTopic, message.encode().buffer)
+      let validationRes = await node.wakuRelay.validateMessage(pubSubTopic, message)
+      if validationRes.isErr():
+        return err(validationRes.error)
+      else:
```
jm-clius, line 923: Since you return in the previous line, I don't think this `else` is necessary. In other words, if validation failed, this block of code will not be executed, whether in an `else` block or not.

jm-clius (approved):
Great work. Minor comment below which you may address, but no need to re-request review. According to me this can be merged. :)

PR #2696, opened 2024-05-13
chore: generic change to reduce the number of compilation warnings

Ivansete-status:
## Description
Simple but long (sorry for that) PR that aims to reduce the number of compilation warnings.

- Compilation warnings related to "imported module but not used".
- Changes around `async: (raises:[..])` pragma
- `re` is deprecated
- Make proc's not to raise exceptions. This is a little bit beyond the original purpose of the PR. I can submit it in a separate one if you consider appropriate.


waku/waku_filter_v2/client.nim
```diff
@@ -26,44 +26,56 @@ proc sendSubscribeRequest(
...
+        if response.statusDesc.isSome():
+          response.statusDesc.get()
+        else:
+          "filter subscribe error"
+      return err(FilterSubscribeError.parse(response.statusCode, cause = cause))
+  except CatchableError:
+    let errMsg = "exception in waku_filter_v2 client: " & getCurrentExceptionMsg()
+    trace "exception in waku_filter_v2 client", error = getCurrentExceptionMsg()
+    waku_filter_errors.inc(labelValues = [errMsg])
+    return err(FilterSubscribeError.badResponse(errMsg))
```
gabrielmer, line 78: wouldn't it be better for the `try` block to be more scoped?

tests/waku_store_legacy/test_wakunode_store.nim
```diff
@@ -31,7 +31,7 @@ import
   ../testlib/wakucore,
   ../testlib/wakunode
 
-procSuite "WakuNode - Store":
+procSuite "WakuNode - Store Legacy":
```
SionoiS, line 34: ah thanks I totally forgot about renaming tests!

waku/node/waku_node.nim
```diff
@@ -444,9 +443,15 @@ proc mountFilter*(
   )
 
   if node.started:
-    await node.wakuFilter.start()
+    try:
+      await node.wakuFilter.start()
+    except CatchableError:
+      error "failed to start wakuFilter", error = getCurrentExceptionMsg()
```
SionoiS, line 449: You don't like the `catch: ` macro?
```suggestion
    let catchable = catch: await node.wakuFilter.start()
    if catchable.isErr():
      error "failed to start wakuFilter", error = catchable.error.msg
```

waku/waku_filter_v2/protocol.nim
```diff
@@ -312,13 +311,13 @@ proc startMaintainingSubscriptions(wf: WakuFilter, interval: Duration) =
 
   wf.maintenanceTask = setTimer(Moment.fromNow(interval), maintainSubs)
 
-method start*(wf: WakuFilter) {.async.} =
+method start*(wf: WakuFilter) {.async, base.} =
```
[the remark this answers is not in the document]
SionoiS: For proper inheritance I think?

SionoiS (approved):
Such a good clean-up. Amazing!

waku/waku_filter_v2/protocol.nim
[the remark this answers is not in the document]
Ivansete-status: Thanks for the comments @gabrielmer , @SionoiS !
This change actually comes from a compiler warning, where it stated that it was deprecated to use non-base methods (I can't remember the precise message.)

waku/node/waku_node.nim
NagyZoltanPeter: @SionoiS We would need better a valueOrCatch macro... hm, due I don't think catch works together with valueOr template.

NagyZoltanPeter (approved):
This is Spartaaa!!!! ... I would say so after our greek tour, but I better say: It is amazing and very needed PR. Special thank for the depricate net lib warning elimination!

waku/node/waku_node.nim
SionoiS: > @SionoiS We would need better a valueOrCatch macro... hm, due I don't think catch works together with valueOr template.

Yes it's unfortunate, otherwise this would be possible
```
(catch: await node.wakuFilter.start()).isOkOr:
   error "failed to start wakuFilter", error = error.msg
```

NagyZoltanPeter: > > @SionoiS We would need better a valueOrCatch macro... hm, due I don't think catch works together with valueOr template.
> 
> Yes it's unfortunate, otherwise this would be possible
> 
> ```
> (catch: await node.wakuFilter.start()).isOkOr:
>    error "failed to start wakuFilter", error = error.msg
> ```

Well we can add! I may do it somewhere.

Ivansete-status: > You don't like the `catch: ` macro?

Well, I "don't like" maybe because I don't fully understand it xD
To me, it sounds a bit more verbose the traditional approach but I'll give it another glance ;P

waku/waku_filter_v2/client.nim
Ivansete-status: > wouldn't it be better for the `try` block to be more scoped?

Yes, you are absolutely right! I tried to apply that in https://github.com/waku-org/nwaku/pull/2696/commits/06546bbd6cd782139adb3a0944e3b55e34da4996

PR #2699, opened 2024-05-14
fix: refactor shard parsing

gabrielmer:
# Description

When running the invalid configuration

```
./build/wakunode2 --pubsub-topic=/waku/2/rs/2/0 --pubsub-topic=/waku/2/rs/3/0 --cluster-id=2
```

I get

```
Error: unhandled exception: Trying to access value with err Result: use shards with the same cluster Id. [ResultDefect]
```

Refactoring the code to avoid this unhandled exception and handle the error gracefully.

# Changes

- [x] refactor shard parsing logic in order to avoid unhandled exception

### DIFF
--- a/waku/factory/internal_config.nim
+++ b/waku/factory/internal_config.nim
@@ -29,23 +29,21 @@ proc enrConfiguration*(
 
   enrBuilder.withMultiaddrs(netConfig.enrMultiaddrs)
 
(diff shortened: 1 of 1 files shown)


Ivansete-status (approved):
LGTM! Thanks for it! :100:

PR #2703, opened 2024-05-15
chore: test_client - add nil error handling after serverSwitch.start() clientSwitch.start()

Ivansete-status:
## Description
Simple PR to help debug flaky crashes on mac-os tests.
This is based on previous studies made by @gabrielmer

### DIFF
--- a/tests/waku_store/test_client.nim
+++ b/tests/waku_store/test_client.nim
@@ -74,6 +74,14 @@ suite "Store Client":
 
     await allFutures(serverSwitch.start(), clientSwitch.start())
 
+    if serverSwitch.peerInfo.isNil():
+      error "serverSwitch.peerInfo is nil"
+      raise newException(CatchableError, "serverSwitch.peerInfo is nil")
+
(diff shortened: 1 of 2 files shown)


gabrielmer (approved):
Thank you!

NagyZoltanPeter (approved):
I'm not fully convinced about lack if this check causes any fault, due seems impossible to create a switch without valid peerInfo... at least from what I see from the code.
...but maybe I'm overlooking sthg....

Ivansete-status:
> I'm not fully convinced about lack if this check causes any fault, due seems impossible to create a switch without valid peerInfo... at least from what I see from the code. ...but maybe I'm overlooking sthg....



> I'm not fully convinced about lack if this check causes any fault, due seems impossible to create a switch without valid peerInfo... at least from what I see from the code. ...but maybe I'm overlooking sthg....

Well, we have the following error in many occasions:

image

With this PR we are not fixing anything particularly but just giving a bit more clarity when that error happens again in the future

PR #2706, opened 2024-05-16
feat: LiteProtocolTester application and docker compose bundle setup. (WIP)

NagyZoltanPeter:
# NOTICE

My apologizes that it is a bigger PR as during working it grew a lot as new demands are identified to be a working solution. Still it needs more improvement. This PR is aims to complete stage1 (testing within one docker compose)

_Files under apps/liteprotocoltester/monitoring needs no review, copied from nwaku-compose for convenience._

## Update: Now stage 1 functionality works as expected!!!
Unfortunately we will not see any latency in the reports due it seems inside one docker network on the same computed we spend more time between messages when sending than between processing on the receiver side...
This latency is negative!!!

# Description
Intention is to create an application(s) that can monitor light protocols reliability over the network.
This PR adds a new application: liteprotocoltester
That can run as a either a publisher or a receiver light client. 
In a docker compose we boundle a complete setup (stage#1) two instances of the tester application (one publisher, one receiver) and a full service node.

The light client publisher is intended to publish indexed messages via lightpush. On the receiver side client is subscribed to the filter service and creates statistics about message losses and latencies.

# Changes

- [X] new application in apps/liteprotocoltester
- [X] it has its own docker-compose setup
- [X] liteprotocoltester has a docker image builder
- [x] Documentation

# Remaining

Still having issue with the liteprotocoltester application setup. WorkInProgress

## How to test
 
Curretntly optimized for Ubuntu builds...(liteprotocoltester image buids via copy allready built binary from build dir) while WIP.
In the apps/liteprotocol directory:
1. make liteprotocoltester
2. cd apps/liteprotocoltester
3. docker compose build
4. docker compose up -d
5. docker compose logs -f receivernode (/publishernode/servicenode)

## Issue
https://github.com/waku-org/nwaku/issues/2700


apps/liteprotocoltester/tester_config.nim
```diff
@@ -0,0 +1,146 @@
...
+  ../../waku/common/confutils/envvar/defs as confEnvvarDefs,
+  ../../waku/common/confutils/envvar/std/net as confEnvvarNet,
+  ../../waku/common/logging,
+  ../../waku/factory/external_config,
+  ../../waku/waku_core
+
+export confTomlDefs, confTomlNet, confEnvvarDefs, confEnvvarNet
+
+const
+  LitePubsubTopic* = PubsubTopic("/waku/2/default-waku/proto")
```
gabrielmer, line 27: Isn't named sharding going to be deprecated? isn't it better to use static sharding?

apps/liteprotocoltester/tester_config.nim
```diff
@@ -0,0 +1,146 @@
...
+
+  logFormat* {.
+    desc:
+      "Specifies what kind of logs should be written to stdout. Suported formats: TEXT, JSON",
+    defaultValue: logging.LogFormat.TEXT,
+    name: "log-format"
+  .}: logging.LogFormat
+
+  ## Test configuration
+  servicenode* {.desc: "Peer multiaddr to of the service node.", name: "service-node".}:
```
gabrielmer, line 57: super nitpick

```suggestion
  servicenode* {.desc: "Peer multiaddr of the service node.", name: "service-node".}:
```

apps/liteprotocoltester/tester_config.nim
```diff
@@ -0,0 +1,146 @@
...
+    string
+
+  nat* {.
+    desc:
+      "Specify method to use for determining public address. " &
+      "Must be one of: any, none, upnp, pmp, extip:<IP>.",
+    defaultValue: "any"
+  .}: string
```
gabrielmer, line 65: This is for the full node, right? If so, I would specify it in the description

apps/liteprotocoltester/run_service_node.sh
```diff
@@ -0,0 +1,30 @@
...
+      --relay=true\
+      --filter=true\
+      --lightpush=true\
+      --store=false\
+      --rest=true\
+      --rest-admin=true\
+      --rest-private=true\
+      --rest-address=0.0.0.0\
+      --rest-allow-origin="waku-org.github.io"\
+      --rest-allow-origin="*"\
```
gabrielmer, line 18: shouldn't it be enough with the `*`?

apps/liteprotocoltester/run_service_node.sh
```diff
@@ -0,0 +1,30 @@
...
+      --dns-discovery=true\
+      --discv5-discovery=true\
+      --discv5-enr-auto-update=True\
+      --log-level=DEBUG\
+      --metrics-server=True\
+      --metrics-server-address=0.0.0.0\
+      --nodekey=e3f5e64568b3a612dee609f6e7c0203c501dab6131662922bdcbcabd474281d5\
+      --nat=extip:${IP}\
+      --pubsub-topic=/waku/2/default-waku/proto\
+      --cluster-id=0
```
gabrielmer, line 30: same here, probably better to use named sharding + some other cluster?

gabrielmer (commented):
Oooh super excited about this! Looks soo good! Thanks so much 😍 

Will review it by batches and adding comments :))

apps/liteprotocoltester/run_service_node.sh
NagyZoltanPeter: Yeah, probably would do! I will remove the first one.

apps/liteprotocoltester/lightpush_publisher.nim
```diff
@@ -0,0 +1,110 @@
...
+      lightpushContentTopic,
+    )
+    prevMessageAt = getNowInNanosecondTime()
+    let wlpRes = await wakuNode.lightpushPublish(some(lightpushPubsubTopic), message)
+
+    if wlpRes.isOk():
+      info "published message using lightpush",
+        index = messagesSent, count = numMessages
+    else:
+      error "failed to publish message using lightpush", err = wlpRes.error()
```
Ivansete-status, line 70:
```suggestion
      error "failed to publish message using lightpush", err = wlpRes.error
```

apps/liteprotocoltester/lightpush_publisher.nim
```diff
@@ -0,0 +1,110 @@
...
+      inc(failedToSendCount)
+
+    await sleepAsync(delayMessages) # Publish every 5 seconds
+    inc(messagesSent)
+
+  let report = catch:
+    """*----------------------------------------*
+|  Expected  |    Sent    |   Failed   |
+|{numMessages:>11} |{messagesSent-failedToSendCount-1:>11} |{failedToSendCount:>11} |
```
Ivansete-status, line 79: This is great indeed!
In the future, I was envisioning to have a CSV with the following columns:
- use-case: e.g. 1000msg-1sender-1receiver; 1000-big-msgs-1-sender-1-receiver; etc.
- nwaku-version
- expected
- sent
- failed
- avg-latency
- ...

apps/liteprotocoltester/lightpush_publisher.nim
```diff
@@ -0,0 +1,110 @@
...
+
+  if report.isErr:
+    echo "Error while printing statistics"
+  else:
+    echo report.get()
+
+  discard c_raise(ansi_c.SIGTERM)
+
+proc setupAndPublish*(wakuNode: WakuNode, conf: LiteProtocolTesterConf) =
+  if wakuNode.wakuLightpushClient == nil:
```
Ivansete-status, line 90:
```suggestion
  if isNil(wakuNode.wakuLightpushClient):
```

apps/liteprotocoltester/run_tester_node.sh
```diff
@@ -0,0 +1,76 @@
...
+      --cluster-id=0\
+      --num-messages=${NUM_MESSAGES}\
+      --delay-messages=${DELAY_MESSAGES}\
+      --nat=extip:${IP}\
+      ${FUNCTION}\
+      ${PUBSUB}\
+      ${CONTENT_TOPIC}\
+      ${REST_PORT}
+
+      # --config-file=config.toml\
```
Ivansete-status, line 76:
```suggestion
```

apps/liteprotocoltester/run_tester_node.sh
```diff
@@ -0,0 +1,76 @@
...
+  echo "Service node not ready, retrying (retries left: ${RETRIES})"
+  sleep 1
+  RETRIES=$(( $RETRIES - 1 ))
+done
+
+if [ -z "${SERIVCE_NODE_ADDR}" ]; then
+   echo "Could not get SERIVCE_NODE_ADDR and none provided. Failing"
+   exit 1
+fi
+
```
Ivansete-status, line 32:
```suggestion
```

apps/liteprotocoltester/statistics.nim
```diff
@@ -0,0 +1,173 @@
...
+|{self.allMessageCount:>11} |{self.receivedMessages:>11} |{self.lossCount():>11} |{self.misorderCount:>11} |{self.duplicateCount:>11} |
+*----------------------------------------------------------------*
+| Latency stat:                                                  |
+|    avg latency: {self.averageLatency():<47}|
+|    min latency: {self.maxLatency:<47}|
+|    max latency: {self.minLatency:<47}|
+*----------------------------------------------------------------*""".fmt()
+
+  if printable.isErr:
+    echo "Error while printing statistics"
```
Ivansete-status, line 115:
```suggestion
  if printable.isErr():
    echo "Error while printing statistics: " & $printable.error
```

apps/liteprotocoltester/statistics.nim
```diff
@@ -0,0 +1,173 @@
...
+  else:
+    echo printable.get()
+
+proc jsonStat*(self: Statistics): string =
+  let json = catch:
+    """{{\"expected\":{self.allMessageCount},
```
Ivansete-status, line 121: Are the `\` needed when using the triple-quote?

apps/liteprotocoltester/statistics.nim
```diff
@@ -0,0 +1,173 @@
...
+      json.add("{{\"sender\": \"{peerId}\", \"stat\":".fmt())
+      json.add(stats.jsonStat())
+      json.add("}, ")
+    json.add("]}")
+    return json
+    #!fmt: on
+  except CatchableError:
+    return
+      "{\"result:\": \"Error while generating json stats: " & getCurrentExceptionMsg() &
+      "\"}"
```
Ivansete-status, line 161: The `std/json` module is pretty handy for such cases.
f.e. https://github.com/waku-org/nwaku/blob/e7b0777d2599e9d33f407dca8de2b1791bdf8320/library/events/json_message_event.nim#L97


```suggestion
  return $(%*self)
```

apps/liteprotocoltester/statistics.nim
```diff
@@ -0,0 +1,173 @@
...
+  for peerId, stats in self.pairs:
+    let peerLine = catch:
+      "Receiver statistics from peer {peerId}".fmt()
+    if peerLine.isErr:
+      echo "Error while printing statistics"
+    else:
+      echo peerLine.get()
+      stats.echoStat()
+
+proc jsonStats*(self: var PerPeerStatistics): string =
```
Ivansete-status, line 147: Maybe not needed to use `var` in this parameter?
```suggestion
proc jsonStats*(self: PerPeerStatistics): string =
```

apps/liteprotocoltester/tester_config.nim
```diff
@@ -0,0 +1,146 @@
...
+      version = version,
+      secondarySources = proc(
+          conf: LiteProtocolTesterConf, sources: auto
+      ) {.gcsafe, raises: [ConfigurationError].} =
+        sources.addConfigFile(Envvar, InputFile("liteprotocoltester"))
+      ,
+    )
+    ok(conf)
+  except CatchableError:
+    err(getCurrentExceptionMsg())
```
Ivansete-status, line 144: I'd suggest to add a bit more context :)
```suggestion
    err("exception in load: " & getCurrentExceptionMsg())
```

apps/liteprotocoltester/tester_message.nim
```diff
@@ -0,0 +1,111 @@
...
+
+import
+  chronicles,
+  json_serialization,
+  json_serialization/std/options,
+  json_serialization/lexer
+
+import ../../waku/waku_api/rest/serdes
+
+type ProtocolTesterMessage* = object
```
Ivansete-status, line 14: Interesting!
What is this type for?

waku.nimble
```diff
@@ -102,6 +102,10 @@ task chat2bridge, "Build chat2bridge":
   let name = "chat2bridge"
   buildBinary name, "apps/chat2bridge/"
 
+task liteprotocoltester, "Build liteprotocoltester":
+  let name = "liteprotocoltester"
+  buildBinary name, "apps/liteprotocoltester/"
```
Ivansete-status, line 107: Super picky comment, again. Hopefully after commenting too many times I can convince a bit more hehe

```suggestion
task lite-protocol-tester, "Build lite-protocol-tester":
  let name = "lite-protocol-tester"
  buildBinary name, "apps/lite-protocol-tester/"
```

Ivansete-status (commented):
It is looking great so far :)
I just added some nitpick comments that I hope you find useful.
I'm not approving just yet because I guess you are willing to commit more
I'm happy to re-review again in the future.

apps/liteprotocoltester/Dockerfile.liteprotocoltester.compile
```diff
@@ -0,0 +1,61 @@
+# BUILD NIM APP ----------------------------------------------------------------
+  FROM rust:1.77.1-alpine3.18  AS nim-build
+
+  ARG NIMFLAGS
+  ARG MAKE_TARGET=liteprotocoltester
```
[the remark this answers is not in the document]
NagyZoltanPeter: waku.nims takes target name and use it as the main source name for build... so for this we would need to name project main source like this...WDYT?

NagyZoltanPeter:
@gabrielmer , @Ivansete-status : With the latest commit I think the functionality of stage 1 is ready. You can review it or even give a try.

apps/liteprotocoltester/Dockerfile.liteprotocoltester.compile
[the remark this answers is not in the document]
Ivansete-status: > waku.nims takes target name and use it as the main source name for build... so for this we would need to name project main source like this...WDYT?

Yes, I think it would get a bit more readable in hyphen-separated mode

Makefile
```diff
@@ -214,6 +214,10 @@ chat2bridge: | build deps librln
 	echo -e $(BUILD_MSG) "build/$@" && \
 		$(ENV_SCRIPT) nim chat2bridge $(NIM_PARAMS) waku.nims
 
+liteprotocoltester: | build deps librln
+	echo -e $(BUILD_MSG) "build/$@" && \
+		$(ENV_SCRIPT) nim liteprotocoltester $(NIM_PARAMS) waku.nims
```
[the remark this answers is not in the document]
NagyZoltanPeter: ... but none of our target and binary uses hyphen, tough.

apps/liteprotocoltester/lightpush_publisher.nim
NagyZoltanPeter: Yes, it's a good idea! What about into phase2?
Indeed we will need to move to static sharding and variable length messages. Later on randomize used shards for auto sharding.
This table is intended to be human readable. For collecting and analysis I added the json formatted.

apps/liteprotocoltester/run_service_node.sh
NagyZoltanPeter: Yes, that is the next step. This first is the base line of the full concept.

apps/liteprotocoltester/run_tester_node.sh
NagyZoltanPeter: This is left here, I think some debugging needed as the basic idea is that we have a set of cli arguments for configuring the test. For configuring some details of the node itself one can use an config file.
But there was some problem with it when I used it, hence the comment. I will take a look again.

apps/liteprotocoltester/tester_config.nim
NagyZoltanPeter: :-) Wow that something mean having good eyes!!! Thx.

apps/liteprotocoltester/tester_message.nim
NagyZoltanPeter: This is the payload of the messages light push client sends. It fills out on its own side to help calculating losses, mis-ordered receive, latency.
Also we will have multiple sender, to separate messages on the receiver side.

apps/liteprotocoltester/statistics.nim
NagyZoltanPeter: I checked it, but cannot do such easy shortcut unfortunately, that is more for key-value pairs or containers...

apps/liteprotocoltester/statistics.nim
NagyZoltanPeter: Yes, they are not needed!

apps/liteprotocoltester/lightpush_publisher.nim
Ivansete-status: > Yes, it's a good idea! What about into phase2? Indeed we will need to move to static sharding and variable length messages. Later on randomize used shards for auto sharding. This table is intended to be human readable. For collecting and analysis I added the json formatted.

Cool! Sure, let's do it in further PRs :)

apps/liteprotocoltester/run_tester_node.sh
Ivansete-status: Okay perfect!, is fine if you have in controlled :P

gabrielmer (commented):
Last question! 😁

apps/liteprotocoltester/tester_config.nim
NagyZoltanPeter: No, it is for the tester app. Maybe there is a misconfiguration there. I need to look at. Light client is configuring a wakunode instance with only light protocols (lightpush or filter).

apps/liteprotocoltester/lightpush_publisher.nim
```diff
@@ -0,0 +1,112 @@
...
+|  Expected  |    Sent    |   Failed   |
+|{numMessages:>11} |{messagesSent-failedToSendCount-1:>11} |{failedToSendCount:>11} |
+*----------------------------------------*""".fmt()
+
+  if report.isErr:
+    echo "Error while printing statistics"
+  else:
+    echo report.get()
+
+  discard c_raise(ansi_c.SIGTERM)
```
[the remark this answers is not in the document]
NagyZoltanPeter: Well, there is an indefinite end of the tester app, but there is no other way to finish runForever loop which runs the async poll mechanism. We install signal handlers that cares about gracefull shutdown so we force it from here.
But I accept it's a kind of workaround and opened for better solution.

apps/liteprotocoltester/statistics.nim
```diff
@@ -0,0 +1,201 @@
...
+  results
+
+import ./tester_message
+
+type
+  StatHelper = object
+    prevIndex: uint32
+    prevArrivedAt: Moment
+    lostIndicies: HashSet[uint32]
+    seenIndicies: HashSet[uint32]
```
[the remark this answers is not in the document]
NagyZoltanPeter: sht, how could I... will fix it.

gabrielmer:
> @Ivansete-status : Are you determined to use hypen in naming? I can do, but its a bit against our current naming scheme all around the repo, hence I would keep this more dense naming scheme here also. @gabrielmer: WDYT? lite-protocol-tester vs. liteprotocoltester?

I'm pretty neutral about it tbh. I think that with hyphen is a bit nicer but I would probably follow our current conventions and write `liteprotocoltester` :))

apps/liteprotocoltester/run_tester_node.sh
```diff
@@ -0,0 +1,76 @@
...
+
+if [ -n "${PUBSUB}" ]; then
+    PUBSUB=--pubsub-topic="${PUBSUB}"
+fi
+
+if [ -n "${PUBSUB}" ]; then
```
SionoiS, line 38:
```suggestion
if [ -n "${CONTENT_TOPIC}" ]; then
```
?

gabrielmer (approved):
Nothing more to add from my side. Tested it and looks really good! Lots of messages lost though 😶 But that's what this is for :))

Thanks so much!

SionoiS (approved):
LGTM nicely done!

apps/liteprotocoltester/run_tester_node.sh
NagyZoltanPeter: Grear catch! Thank you!

Makefile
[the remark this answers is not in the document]
Ivansete-status: > ... but none of our target and binary uses hyphen, tough.

ah yes I see. Well, we have `cwaku_example` with underscore ;P
This is not a big deal at all. Do what you consider more appropriate. I just shared my nitpick opinion :)

Ivansete-status:
> @Ivansete-status : Are you determined to use hypen in naming? I can do, but its a bit against our current naming scheme all around the repo, hence I would keep this more dense naming scheme here also. @gabrielmer: WDYT? lite-protocol-tester vs. liteprotocoltester?

I think I will only spend a few days without sleeping due to that dense naming hahah
Jokes apart, it is fine to me either option of course :)

I'm not approving it yet because it is in `draft` and I think you may want to push further changes on top of it ( feel free to merge it if  that's not the case .)

------

image

We need to investigate such message loss. I think it may be due to a too-high message rate but we need to pinpoint the causes of that.

Something to consider for the short future:
- Also print the message rate. If possible, maybe also interesting to print the measured bitrate of the test.
- From the loss PoV, I think it could help to print the `senderPeerId` and also the offending `msgIndex` that got lost. Maybe a few log lines more.

------

Congrats on it and great job so far! :partying_face:

NagyZoltanPeter:
@Ivansete-status , @SionoiS , @gabrielmer Thank you for all the valuable comments and great gotchas. I incorporated all I think. I will squash and rebase now for merge. THX.

PR #2708, opened 2024-05-17
chore: libwaku - allow to properly set the log level in libwaku and unify a little

Ivansete-status:
## Description
- Set compilation flags to properly compile chronicles for `libwaku`
- Unify the code in the sense that we only expose one proc from the `logging` module and have a standard logging setup.
- The log is configured in `waku.nim` within `init` proc.
- In `wakunode2` we also set the log before "waku.init" is invoked because we need to setup the logs properly for the "essential rest server" that we start at the beginning

## Issue
closes https://github.com/waku-org/nwaku/issues/2411

### DIFF
--- a/waku.nimble
+++ b/waku.nimble
@@ -109,4 +109,7 @@ task libwakuStatic, "Build the cbindings waku node library":
 
 task libwakuDynamic, "Build the cbindings waku node library":
   let name = "libwaku"
-  buildLibrary name, "library/", "-d:chronicles_log_level=ERROR", "dynamic"
+  buildLibrary name,
+    "library/",
+    """-d:chronicles_line_numbers -d:chronicles_runtime_filtering=on -d:chronicles_sinks="textlines,json" -d:chronicles_default_output_device=Dynamic -d:chronicles_disabled_topics="eth,dnsdisc.client" """,
+    "dynamic"
--- a/waku.nimble
+++ b/waku.nimble
@@ -105,7 +105,10 @@ task chat2bridge, "Build chat2bridge":
 ### C Bindings
 task libwakuStatic, "Build the cbindings waku node library":
   let name = "libwaku"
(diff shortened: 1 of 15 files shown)


waku.nimble
```diff
@@ -105,8 +105,14 @@ task chat2bridge, "Build chat2bridge":
 ### C Bindings
 task libwakuStatic, "Build the cbindings waku node library":
   let name = "libwaku"
-  buildLibrary name, "library/", "-d:chronicles_log_level=ERROR", "static"
+  buildLibrary name,
+    "library/",
+    """-d:chronicles_line_numbers -d:chronicles_runtime_filtering=on -d:chronicles_sinks="textlines,json" -d:chronicles_default_output_device=Dynamic -d:chronicles_disabled_topics="eth,dnsdisc.client" """,
```
NagyZoltanPeter, line 110: Just wonder - although I see it is the default in nim.cfg also - why exactly these and only these topics are disabled??? I have no issue with it, just seems such a leftover...

waku/common/logging.nim
```diff
@@ -90,3 +94,15 @@ proc setupLogFormat*(format: LogFormat, color = true) =
         "the present module should be compiled with '-d:chronicles_default_output_device=dynamic' " &
         "and '-d:chronicles_sinks=\"textlines,json\"' options"
     .}
+
+proc setupLog*(level: LogLevel, format: LogFormat) =
+  ## Logging setup
+  # Adhere to NO_COLOR initiative: https://no-color.org/
+  let color =
+    try:
+      not parseBool(os.getEnv("NO_COLOR", "false"))
```
NagyZoltanPeter, line 103: It is great!!!!

NagyZoltanPeter (approved):
LGTM!!!!

gabrielmer (approved):
Superr, thanks so much!

PR #2711, opened 2024-05-19
chore: Discovery in libwaku

Ivansete-status:
## Description
Finalizing the addition of discovery features to `libwaku`.

## Changes
- Extension of `libwaku` with discovery capabilities.
- Extension of `cwaku_example` so that we can check how the new exposed functions work.
- Enhancement in `node_lifecycle_request.nim` for better error control. That helps to determine when a wrong JSON config parameter is passed to the library.

## Issue
closes https://github.com/waku-org/nwaku/issues/2455

### DIFF
--- a/waku/discovery/waku_discv5.nim
+++ b/waku/discovery/waku_discv5.nim
@@ -4,7 +4,7 @@ else:
   {.push raises: [].}
 
 import
-  std/[sequtils, strutils, options, sets, net],
+  std/[sequtils, strutils, options, sets, net, json],
   stew/results,
   chronos,
   chronicles,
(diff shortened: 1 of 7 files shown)


waku/discovery/waku_discv5.nim
```diff
@@ -373,3 +373,28 @@ proc setupDiscoveryV5*(
   WakuDiscoveryV5.new(
     rng, discv5Conf, some(myENR), some(nodePeerManager), nodeTopicSubscriptionQueue
   )
+
+proc updateBootstrapRecords*(
```
gabrielmer, line 377: is there any value in updating bootstrap nodes on runtime? aren't they only used when connecting to the network?

gabrielmer (approved):
LGTM, thank you! :))

waku/discovery/waku_discv5.nim
Ivansete-status: > is there any value in updating bootstrap nodes on runtime? aren't they only used when connecting to the network?

Good point! This feature comes from what we are already offering from `go-waku` so we offer the same. I guess that was a requirement from any of the current projects that integrates us (cc @richard-ramos .)

arnetheduck: > dns discovery

are there nodes that do not require dns discovery in the bootstrap node list?

Ivansete-status: > > dns discovery
> 
> are there nodes that do not require dns discovery in the bootstrap node list?

In case you asked w.r.t `wakunode2` (`nwaku`), it is mandatory for the app to retrieve bootstrap nodes from DNS:
https://github.com/waku-org/nwaku/blob/5ee4cba5340ac64223e72c7abb204da6944af9de/waku/factory/waku.nim#L140-L161

arnetheduck: don't know the details then, but that sounds like an unnecessary single point of failure - normally, one would bootstrap discv5 from multiple sources (dns, hardcoded nodes, nodes from previous startups etc)

gabrielmer: My understanding is that it's not mandatory. If we don't have DNS discovery enabled `retrieveDynamicBootstrapNodes` will return `ok` with an empty seq and the node will simply set up without dynamic bootstrap nodes.

https://github.com/waku-org/nwaku/blob/5ee4cba5340ac64223e72c7abb204da6944af9de/waku/discovery/waku_dnsdisc.nim#L130-L131

However, if we enable DNS discovery and it fails retrieving the nodes then the whole node setup fails. Not sure if it should be like that or if it would be better to add a warning and continue the setup using other sources as a backup (as if DNS discovery wasn't configured)

SionoiS (approved):
LGTM Thank you!

PR #2721, opened 2024-05-23
fix: use `when` instead of `if` for adding soname on linux

richard-ramos:
Fixes build error on macos


gabrielmer (approved):
Thanks so much!

PR #2723, opened 2024-05-23
chore: deprecating named sharding

### LINKED ISSUE #2772: chore: deprecate support for named sharding:
### Background
Deprecate Support for Named Sharding:
It is recommended to remove support for named sharding.
Peer management should ensure that only peers supporting sharded topics are connected.

### Acceptance criteria
- [ ] Named sharding should not be accepted anymore.
- [ ] `--pubsub-topic` CLI config should be deprecated in favor of `--cluster-id` and `--shard`

gabrielmer:
# Description
Deprecating named sharding.

We won't allow further usage of named sharding and the default pubsub topic will be replaced by each cluster's `0` shard.

Another PR will follow in which we will deprecate the `--pubsub-topic` CLI configuration in favor of `shard`

# Note
This PR will be merged after having announced the deprecation of named sharding during the next two releases (it will enter to `v0.31.0`). However, the general approach should be reviewed from now.

## Issue

closes #2772


tests/testlib/wakunode.nim
```diff
@@ -36,7 +36,7 @@ proc defaultTestWakuNodeConf*(): WakuNodeConf =
     maxConnections: 50,
     maxMessageSize: "1024 KiB",
     clusterId: 0.uint32,
-    pubsubTopics: @["/waku/2/rs/1/0"],
+    pubsubTopics: @["/waku/2/rs/0/0"],
```
Ivansete-status, line 39: Why this change? Seems unrelated to the original purpose of the PR

waku/waku_core/topics/pubsub_topic.nim
```diff
@@ -16,40 +16,23 @@ export parsing
...
 
 proc `$`*(topic: NsPubsubTopic): string =
   ## Returns a string representation of a namespaced topic
   ## in the format `/waku/2/<raw-topic>
-  case topic.kind
-  of NsPubsubTopicKind.NamedSharding:
-    "/waku/2/" & topic.name
-  of NsPubsubTopicKind.StaticSharding:
-    "/waku/2/rs/" & $topic.clusterId & "/" & $topic.shardId
+  return "/waku/2/rs/" & $topic.clusterId & "/" & $topic.shardId
```
Ivansete-status, line 35: I think we should review the comment under this proc

waku/waku_core/topics/pubsub_topic.nim
```diff
@@ -89,49 +72,23 @@ proc parseStaticSharding*(
...
 proc parse*(
     T: type NsPubsubTopic, topic: PubsubTopic | string
 ): ParsingResult[NsPubsubTopic] =
   ## Splits a namespaced topic string into its constituent parts.
   ## The topic string has to be in the format `/<application>/<version>/<topic-name>/<encoding>`
-  if topic.startsWith(StaticShardingPubsubTopicPrefix):
-    NsPubsubTopic.parseStaticSharding(topic)
-  else:
-    NsPubsubTopic.parseNamedSharding(topic)
+  NsPubsubTopic.parseStaticSharding(topic)
```
Ivansete-status, line 80: In this `proc` I think we need to change the `topic` type to only `PubsubTopic`

waku/factory/external_config.nim
```diff
@@ -300,12 +300,13 @@ type WakuNodeConf* = object
 
     pubsubTopics* {.
       desc: "Default pubsub topic to subscribe to. Argument may be repeated.",
+      defaultValue: @[],
```
Ivansete-status, line 303: Shall we mark this `pubsubTopic` config attribute as deprecated? In that case, we will need to keep it for two more versions.

Ivansete-status (commented):
Thanks for it! :partying_face: 
I'm not approving yet because it is in `draft`
Cheers

tests/testlib/wakunode.nim
gabrielmer: I think it is a bug, because we are setting `clusterId: 0.uint32` but the pubsub topic we are subscribing to corresponds to cluster 1.

Found this bug now that I added a check in this PR to throw an error if we try to subscribe to a pubsub topic that corresponds to a different shard we're on. So fixed it so the pubsub topic corresponds to cluster 0.

waku/waku_core/topics/pubsub_topic.nim
gabrielmer: I see that
https://github.com/waku-org/nwaku/blob/8b42f199baf4e00794c4cec4d8601c3f6c330a20/waku/waku_core/topics/pubsub_topic.nim#L17

So will try to change the places where it says
```
PubsubTopic | string
```

to just `PubsubTopic`, if that makes sense

waku/factory/external_config.nim
gabrielmer: Yes, the idea is to deprecate it and stop creating confusion/overlap between `pubsubTopic` and `shard`.
So we will have to keep it for 2 more releases (announcing the deprecation in the release notes) and afterwards create a new PR deprecating it

SionoiS:
Should we take this opportunity to remove the usage of `string`s for pubsub topics internally?

My idea would be to replace `PubsubTopic = string` with a proper type that can the turned into a formatted string if needed.

WDYT?

waku/factory/node_factory.nim
```diff
@@ -117,10 +117,16 @@ proc setupProtocols(
   ## Optionally include persistent message storage.
   ## No protocols are started yet.
 
+  var shardCount: uint32
```
alrevuelta, line 120: Not sure about this. imho `pubsubTopics.len` shall be always `>0`?

`pubsubTopics` will be deprecated soon and this will be eg a cli flag `shard-count` or `network-shards`. but that's for another PR.

waku/factory/waku.nim
```diff
@@ -96,12 +96,6 @@ proc init*(T: type Waku, conf: WakuNodeConf): Result[Waku, string] =
 
   case confCopy.clusterId
 
-  # cluster-id=0
```
alrevuelta, line 99: why removing cluster-id=0 configuration?

maybe we should keep it but with just 1 shard? for backwards compatibility? unsure though if someone is using it.

I mean, cluster-id=0 currently was a single named "shard" (default pubsub topic). From an app PoV that is equivalent to autosharding with a single shard.

waku/factory/node_factory.nim
gabrielmer: True! That was done because it made everything work better when I tested not setting anything in `pubsub-topic` and only setting `shard`.

But if we currently need to use `pubsub-topic` to list all the topics in the network, then it makes sense to add the logic similar to this in the next PR, the one deprecating `pubsub-topic`

waku/factory/waku.nim
gabrielmer: My understanding is that after deprecating named sharding, cluster 0 was going to stop being a "special case".

Is there's any reason why we would want to keep the configuration of cluster 0 hardcoded? Once named sharding is deprecated, there won't be backward compatibility in any case

waku/factory/external_config.nim
jm-clius: Perhaps add `Deprecated` in the `desc`?

jm-clius (approved):
Long overdue! Thanks.

gabrielmer:
> Perhaps add `Deprecated` in the `desc`?

How? Like this?
```
desc: "Deprecated. Default pubsub topic to subscribe to. Argument may be repeated.",
```

We can, although I will remove the `pubsubTopics` configuration in the next PR and both will be merged at the same time (roughly in 2 months), so the change won't really make it to master in practice. Now that I think about it, adding a default value to it also will never make it, I should probably remove that change :))

waku/factory/node_factory.nim
alrevuelta: > That was done because it made everything work better when I tested not setting anything in pubsub-topic and only setting shard.

mm IMHO `pubsub-topic` has to be always set. It could be a default value (like when using clusterid=0 or 1) but it has to be set somewhere.

waku/factory/waku.nim
alrevuelta: im ok with removing it. you are right with cluster 0 not being a special case anymore. But think also about the "known clusters" concept. Cluster configuration is not only about topics but also bootstrap nodes, validation strategy, etc.

so the question here is. do we want cluster-id=0 to be a "known cluster" with some specific configuration that people can use, or we don't care about that? imho nope, i would care only about TWN=1

but ccing @jm-clius

apps/chat2/config_chat2.nim
```diff
@@ -85,7 +85,7 @@ type
 
     topics* {.
       desc: "Default topics to subscribe to (space separated list).",
-      defaultValue: "/waku/2/default-waku/proto",
```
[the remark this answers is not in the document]
gabrielmer: Another PR will follow in which we get rid completely of the pubsub topic concept and will change all the instances of pubsub topic in the code in favor of cluster/shards.

If it's ok, will do this and all other changes related to deprecating the pubsub topic configuration in the next PR, so both PRs remain smaller and scoped

Ivansete-status (approved):
LGTM! Thanks! :100:

PR #2724, opened 2024-05-23
fix: invalid cursor returning messages

SionoiS:
Store v3 queries with invalid cursors will no longer return messages. I also added more/better tests.

Fixing the issue for real this time :100: % pinky promise, I swear! :rofl:

### DIFF
--- a/tests/wakunode_rest/test_rest_store.nim
+++ b/tests/wakunode_rest/test_rest_store.nim
@@ -1,9 +1,10 @@
 {.used.}
 
 import
-  std/[options, times, sugar],
+  std/[options, sugar],
   stew/shims/net as stewNet,
   chronicles,
+  chronos/timer,
   testutils/unittests,
   eth/keys,
   presto,
@@ -24,8 +25,10 @@ import
   ../../../waku/waku_api/rest/store/types,
   ../../../waku/waku_archive,
   ../../../waku/waku_archive/driver/queue_driver,
+  ../../../waku/waku_archive/driver/sqlite_driver,
+  ../../../waku/common/databases/db_sqlite,
+  ../../../waku/waku_archive/driver/postgres_driver,
   ../../../waku/waku_store as waku_store,
-  ../../../waku/common/base64,
   ../testlib/wakucore,
   ../testlib/wakunode
 
@@ -42,7 +45,7 @@ proc put(
       if message.timestamp > 0:
         message.timestamp
       else:
-        getNanosecondTime(getTime().toUnixFloat())
+        getNowInNanosecondTime()
 
   store.put(pubsubTopic, message, digest, msgHash, receivedTime)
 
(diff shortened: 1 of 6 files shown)


gabrielmer:
There's something I don't quite understand. In 3 different tests/files, sending an invalid cursor returns different things. 

- In `tests/waku_archive/test_driver_postgres_query.nim` ok is returned but with no messages
- In `tests/waku_archive/test_driver_queue_query.nim` the error `invalid_cursor` is returned
- In `tests/waku_archive/test_driver_sqlite_query.nim` returns an error with `cursor not found`

Why the different behavior in those cases? Especially the one that doesn't return an error, I guess that the error message depends on the driver

SionoiS:
> There's something I don't quite understand. In 3 different tests/files, sending an invalid cursor returns different things.
> 
> * In `tests/waku_archive/test_driver_postgres_query.nim` ok is returned but with no messages
> * In `tests/waku_archive/test_driver_queue_query.nim` the error `invalid_cursor` is returned
> * In `tests/waku_archive/test_driver_sqlite_query.nim` returns an error with `cursor not found`
> 
> Why the different behavior in those cases? Especially the one that doesn't return an error, I guess that the error message depends on the driver

Yes that is a good point that will be addressed by my driver refactor. I plan to consolidate driver error handling.

gabrielmer (approved):
Thanks so much!

tests/wakunode_rest/test_rest_store.nim
```diff
@@ -84,6 +87,84 @@ procSuite "Waku Rest API - Store v3":
...
+      "5", # empty implies default page size
+    )
+
+    check:
+      response.status == 200
+      $response.contentType == $MIMETYPE_JSON
+      response.data.messages.len == 0
+
+    await restServer.stop()
+    await restServer.closeWait()
```
Ivansete-status, line 165: Interesting. Did you need that to avoid a possible crash?

waku/waku_api/rest/store/types.nim
```diff
@@ -35,6 +35,9 @@ proc parseHash*(input: Option[string]): Result[Option[WakuMessageHash], string]
   let decodedBytes = base64.decode(Base64String(base64Encoded)).valueOr:
     return err("waku message hash parsing error: " & error)
 
+  if decodedBytes.len != 32:
+    return err("waku message hash parsing error: invalid hash length")
```
Ivansete-status, line 39: Maybe we can add a bit more detail :)
```suggestion
    return err("waku message hash parsing error: invalid hash length: " & $decodedBytes.len)
```

Ivansete-status (approved):
LGTM! Thanks for it! :100:

tests/wakunode_rest/test_rest_store.nim
SionoiS: I'm not sure. It's just a copy pasta of another test.

PR #2725, opened 2024-05-23
fix: do not print the db url on error

richard-ramos:
# Description
When there's an error with the dbUrl, the logs will contain the full url, and this is a bad idea, since the connection string might contain the user/password, thus, revealing too much information on the logs

cc: @jakubgs


gabrielmer (approved):
Thank you!

SionoiS (approved):
Thanks!

PR #2734, opened 2024-05-27
docs: release-process.md - add step to merge the release branch back to master

Ivansete-status:
## Description
Adding a step to make it clearer the kind of merge that needs to be applied after a release has been completed.
Notice that this is needed to make sure the following command retrieves the appropriate information when running from `master`:
```code
git describe --abbrev=6 --always --tags
```

In other words, we need to use the following merge option when merging the release branch into master:
image

But, it is important to bear in mind that we should use the "Squash and merge" option for the regular PR's:
image


NagyZoltanPeter (approved):
Yesss!

DarshanBPatel:
It's better !

DarshanBPatel (approved):

PR #2736, opened 2024-05-27
fix(waku_archive): only allow a single instance to execute migrations

### LINKED ISSUE #2726: feat: migrations must have an status and check against it before a migration is executed
### Problem
In the status fleet some nodes share a postgres database. When the nwaku version is upgraded, instances sharing the same database will attempt to execute migrations at the same time, and one of them might fail, or might end up with an incorrect db state. We ran into this problem today.

### Suggested solution
Have a table or something that would indicate whether a migration is being executed or not, and only execute the migration if it's possible to do so.

richard-ramos:
# Description
Uses an advisory lock which acts as a mutex for the postgresql migrations.
This lock will be acquired before any migration is executed. If it's not possible to acquire, the instance will fail to start with these errors:
```
ERR 2024-05-27 17:59:38.155-04:00 failed to acquire lock                     topics="waku archive migration" tid=479763 file=migrations.nim:77 error="another waku instance is currently executing a migration"
ERR 2024-05-27 17:59:38.155-04:00 Mounting protocols failed                  tid=479763 file=node_factory.nim:435 error="failed to setup archive driver: ArchiveDriver build failed in migration: failed to lock the db"
```
Once the migration process is done, the lock will be released.

This is required for scenarios in which a postgresql database is used by more than one node (like in status).

cc: @jakubgs 

## Issue

closes #2726

### DIFF
--- a/waku/waku_archive/driver/postgres_driver/postgres_driver.nim
+++ b/waku/waku_archive/driver/postgres_driver/postgres_driver.nim
@@ -853,6 +853,42 @@ proc sleep*(
 
   return ok()
 
+proc acquireDatabaseLock*(
+    s: PostgresDriver, lockId: int = 841886
+): Future[ArchiveDriverResult[void]] {.async.} =
+  ## Acquire an advisory lock (useful to avoid more than one application running migrations at the same time)
+  let locked = (
+    await s.getStr(
+      fmt"""
+    SELECT pg_try_advisory_lock({lockId})
+    """
+    )
+  ).valueOr:
+    return err("error acquiring a lock: " & error)
+
+  if locked == "f":
+    return err("another waku instance is currently executing a migration")
+
+  return ok()
+
+proc releaseDatabaseLock*(
+    s: PostgresDriver, lockId: int = 841886
+): Future[ArchiveDriverResult[void]] {.async.} =
+  ## Acquire an advisory lock (useful to avoid more than one application running migrations at the same time)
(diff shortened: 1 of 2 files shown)


waku/waku_archive/driver/postgres_driver/postgres_driver.nim
```diff
@@ -853,6 +853,40 @@ proc sleep*(
...
+  let lockedRes = await s.getStr(
+    fmt"""
+    SELECT pg_try_advisory_lock({lockId})
+    """
+  )
+  if lockedRes.isErr():
+    return err("error acquiring a lock: " & $lockedRes.error)
+
+  if lockedRes.value == "f":
+    return err("another waku instance is currently executing a migration")
```
Ivansete-status, line 869: I think it could be a little bit more concise with the `valueOr` approach ( notice that we also use `isOkOr` for `Result[void, ...]` )
```suggestion
  let locked = (await s.getStr(
    fmt"""
    SELECT pg_try_advisory_lock({lockId})
    """
  )).valueOr:
    return err("error acquiring a lock: " & $error)

  if locked == "f":
    return err("another waku instance is currently executing a migration")
```

Ivansete-status (approved):
LGTM! Thanks for it! :100:

gabrielmer (approved):
Thanks so much! 🤩

jakubgs (commented):
Nice, thanks.

PR #2740, opened 2024-05-29
chore: test_waku_store.nim - add logs to better analyse uncertain flaky tests

Ivansete-status:
Very simple PR that aims to better catch where a flaky test is actually failing. That flaky test only happens on CI MacOS


NagyZoltanPeter (approved):
Hope we catch it.

Issue #2744, opened 2024-05-29
chore: extract select to a const


richard-ramos:
In https://github.com/waku-org/nwaku/blob/01577f69a571bcbf207179b78d837a59a8d86d32/waku/waku_archive/driver/postgres_driver/postgres_driver.nim the following SELECT is reused in the queries: 
```sql
SELECT storedAt, contentTopic, payload, pubsubTopic, version, timestamp, id, messageHash, meta FROM messages
```
We ran into an issue on which one of the columns (`meta`) was not added by mistake to some of the queries, causing a failure while attempting to retrieve messages. 

To avoid this issue happening in the future, I suggest extracting this part of the queries to a constant and reuse it in all the queries.

SionoiS:
Should be addressed by my driver refactor I think.

PR #2747, opened 2024-05-30
chore: extract common prefixes into a constant for multiple query

DarshanBPatel:
Solution of #2744 "chore: extract select to a const"

Refactored similar query select parts into a constant for clarity.


gabrielmer (approved):
Niiiiiice! Looks great :))

Thanks so much!

SionoiS (approved):
Much nicer thanks!

Ivansete-status (approved):
LGTM! Thanks !
I'm adding just a tiny suggestion to make sure there is always a space added at the end

PR #2750, opened 2024-05-31
chore: Improving liteprotocolteseter stats

gabrielmer:
# Description

Running `liteprotocoltester`, we have seen lately in the `receivernode` periodic stats that there seems to be a message loss.

For example
```
receivernode-1  | *-----------------------------------------------------------------------------*
receivernode-1  | |  Expected  |  Reveived  |    Loss    |  Misorder  |    Late    |  Duplicate |
receivernode-1  | |        120 |         16 |        104 |          0 |          0 |          0 |
receivernode-1  | *-----------------------------------------------------------------------------*
receivernode-1  | | Latency stat:                                                               |
receivernode-1  | |    avg latency: 0ns                                                         |
receivernode-1  | |    min latency: 0ns                                                         |
receivernode-1  | |    max latency: 0ns                                                         |
receivernode-1  | *-----------------------------------------------------------------------------*
```

This happens because when the stats are printed (every 20 seconds), all messages haven't been sent yet (default is 120 messages, 1 message per second).

Therefore, to avoid confusion, changed the calculation of messages lost to be the diff between the maximum index received (index starts at 1) minus the amount of messages received.
In other words, if we have received up to the message number N, we expect to have received N messages in total.

Added the `Target` column to the stats, so we are also aware of how many messages we expect to receive during the whole run. The output currently looks the following way

```
receivernode-1  | *------------------------------------------------------------------------------------------*
receivernode-1  | |  Expected  |  Received  |   Target   |    Loss    |  Misorder  |    Late    |  Duplicate |
receivernode-1  | |         16 |         16 |        120 |          0 |          0 |          0 |          0 |
receivernode-1  | *------------------------------------------------------------------------------------------*
receivernode-1  | | Latency stat:                                                                            |
receivernode-1  | |    avg latency: 0ns                                                                      |
receivernode-1  | |    min latency: 0ns                                                                      |
receivernode-1  | |    max latency: 0ns                                                                      |
receivernode-1  | *------------------------------------------------------------------------------------------*
```

### DIFF
--- a/apps/liteprotocoltester/statistics.nim
+++ b/apps/liteprotocoltester/statistics.nim
@@ -125,7 +125,7 @@ proc echoStat*(self: Statistics) =
   let printable = catch:
     """*-----------------------------------------------------------------------------*
 |  Expected  |  Received  |    Loss    |  Misorder  |    Late    |  Duplicate |
-|{self.helper.prevIndex+1:>11} |{self.receivedMessages:>11} |{self.lossCount():>11} |{self.misorderCount:>11} |{self.lateCount:>11} |{self.duplicateCount:>11} |
+|{self.helper.prevIndex:>11} |{self.receivedMessages:>11} |{self.lossCount():>11} |{self.misorderCount:>11} |{self.lateCount:>11} |{self.duplicateCount:>11} |
 *-----------------------------------------------------------------------------*
 | Latency stat:                                                               |
 |    avg latency: {$self.averageLatency():<60}|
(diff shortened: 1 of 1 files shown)


NagyZoltanPeter (approved):
Ah, I see now the confusion. There are prints during the test, just to see where we are and there is a final printout before terminating filter client, that one meant to be the result of the test! I assume I forgot to exactly show that last print out is the result of the test while any previous are just informative.
Anyhow thank you for making it clear.

Ivansete-status (approved):
LGTM!

PR #2752, opened 2024-05-31
chore(archive): archive refactor

SionoiS:
I updated the `Archive`, the commonly used types and the driver interface.

### DIFF
--- a/waku/waku_archive/archive.nim
+++ b/waku/waku_archive/archive.nim
@@ -96,14 +96,14 @@ proc handleMessage*(
   (await self.driver.put(msgHash, pubsubTopic, msg)).isOkOr:
     waku_archive_errors.inc(labelValues = [insertFailure])
     trace "failed to insert message",
-      hash = msgHash.to0xHex(),
+      hash_hash = msgHash.to0xHex(),
       pubsubTopic = pubsubTopic,
       contentTopic = msg.contentTopic,
       timestamp = msg.timestamp,
       error = error
 
   trace "message archived",
-    hash = msgHash.to0xHex(),
+    hash_hash = msgHash.to0xHex(),
     pubsubTopic = pubsubTopic,
     contentTopic = msg.contentTopic,
     timestamp = msg.timestamp
@@ -115,9 +115,6 @@ proc findMessages*(
     self: WakuArchive, query: ArchiveQuery
 ): Future[ArchiveResult] {.async, gcsafe.} =
   ## Search the archive to return a single page of messages matching the query criteria
-
-  if query.contentTopics.len > 10:
-    return err(ArchiveError.invalidQuery("too many content topics"))
 
   if query.cursor.isSome():
     let cursor = query.cursor.get()
--- a/waku/waku_archive/archive.nim
+++ b/waku/waku_archive/archive.nim
@@ -4,7 +4,7 @@ else:
   {.push raises: [].}
 
 import
-  std/[times, options, sequtils, strutils, algorithm],
+  std/[times, options, sequtils, algorithm],
   stew/[results, byteutils],
   chronicles,
   chronos,
@@ -55,9 +55,6 @@ proc validate*(msg: WakuMessage): Result[void, string] =
     # Ephemeral message, do not store
     return
 
-  if msg.timestamp == 0:
-    return ok()
-
   let
     now = getNanosecondTime(getTime().toUnixFloat())
     lowerBound = now - MaxMessageTimestampVariance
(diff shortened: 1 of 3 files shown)


waku/waku_archive/archive.nim
```diff
@@ -55,9 +55,6 @@ proc validate*(msg: WakuMessage): Result[void, string] =
     # Ephemeral message, do not store
     return
 
-  if msg.timestamp == 0:
-    return ok()
-
```
Ivansete-status, line 60: I agree that we shouldn't have this validation because the WakuMessage spec doesn't enforce it. https://rfc.vac.dev/waku/standards/core/14/message
Let me know if that statement is correct

waku/waku_archive/archive.nim
```diff
@@ -92,38 +89,24 @@ proc handleMessage*(
...
-        getNanosecondTime(getTime().toUnixFloat())
-
-  trace "handling message",
-    msg_hash = msgHashHex,
-    pubsubTopic = pubsubTopic,
-    contentTopic = msg.contentTopic,
-    msgTimestamp = msg.timestamp,
-    usedTimestamp = msgTimestamp,
-    digest = msgDigestHex
+  let msgHash = computeMessageHash(pubsubTopic, msg)
```
Ivansete-status, line 92: We are changing this considering this is for _StoreV3_?

waku/waku_archive/archive.nim
```diff
@@ -92,38 +89,24 @@ proc handleMessage*(
...
-    msg_hash = msgHashHex,
+    trace "failed to insert message",
+      hash = msgHash.to0xHex(),
+      pubsubTopic = pubsubTopic,
+      contentTopic = msg.contentTopic,
+      timestamp = msg.timestamp,
+      error = error
+
+  trace "message archived",
+    hash = msgHash.to0xHex(),
```
Ivansete-status, line 106: We should always keep the `msg_hash` attribute so that we can always grep easily :)
```suggestion
    msg_hash = msgHash.to0xHex(),
```

waku/waku_archive/archive.nim
```diff
@@ -133,6 +116,19 @@ proc findMessages*(
 ): Future[ArchiveResult] {.async, gcsafe.} =
   ## Search the archive to return a single page of messages matching the query criteria
 
+  if query.contentTopics.len > 10:
+    return err(ArchiveError.invalidQuery("too many content topics"))
+
```
Ivansete-status, line 121: Let's not use that limitation :)
```suggestion
  if query.contentTopics.len > 10:
    return err(ArchiveError.invalidQuery("too many content topics"))

```

waku/waku_archive/archive.nim
```diff
@@ -175,115 +164,33 @@ proc findMessages*(
...
-    reverse(messages)
     reverse(topics)
+    reverse(messages)
 
   return ok(
-    ArchiveResponse(hashes: hashes, messages: messages, topics: topics, cursor: cursor)
+    ArchiveResponse(cursor: cursor, topics: topics, hashes: hashes, messages: messages)
   )
 
-proc findMessagesV2*(
```
Ivansete-status, line 219: We are deprecating _StoreV2_ with that removal?

Ivansete-status (approved):
LGTM! Thanks for it! :100:
I'm approving not to block it but I'd like us to solve all the comments before merging :)
Cheers

gabrielmer (approved):
LGTM, thank you!

Are we deprecating StoreV2 here or the old version will be available in some other legacy files.
In case we're deprecating it, I think we haven't announced the deprecation yet, right?

SionoiS:
note to self -> https://github.com/waku-org/nwaku/pull/2751#discussion_r1622571774

waku/waku_archive/archive.nim
SionoiS: This the the spec section, for reference -> https://github.com/vacp2p/rfc/tree/docs/new-store-v3-protocol/content/docs/rfcs/13#waku-message-store-eligibility

waku/waku_archive/archive.nim
SionoiS: Yes store v3 do not use the previous `digest` concept.

waku/waku_archive/archive.nim
SionoiS: Yes correct. Legacy archive still supports it.

SionoiS:
> Are we deprecating StoreV2 here or the old version will be available in some other legacy files. In case we're deprecating it, I think we haven't announced the deprecation yet, right?

#2751 is adding legacy version of archive and drivers. Deprecation will need be announced (next release maybe?).

PR #2756, opened 2024-05-31
chore(archive): tests refactor

SionoiS:


tests/waku_archive/test_waku_archive.nim
```diff
@@ -58,22 +50,6 @@ suite "Waku Archive - message handling":
...
-    let invalidSenderTime = 0
-    let message = fakeWakuMessage(ts = invalidSenderTime)
-
-    ## When
-    waitFor archive.handleMessage(DefaultPubSubTopic, message)
-
-    ## Then
-    check:
-      (waitFor driver.getMessagesCount()).tryGet() == 1
-
```
gabrielmer, line 76: Instead of deleting the test case, wouldn't it be better to adapt it to the new behavior when time is not set?

gabrielmer (approved):
LGTM, thank you! :))

tests/waku_archive/archive_utils.nim
```diff
@@ -23,26 +23,11 @@ proc newSqliteArchiveDriver*(): ArchiveDriver =
...
 proc put*(
     driver: ArchiveDriver, pubsubTopic: PubSubTopic, msgList: seq[WakuMessage]
 ): ArchiveDriver =
   for msg in msgList:
-    let
-      msgDigest = computeDigest(msg)
-      msgHash = computeMessageHash(pubsubTopic, msg)
-      _ = waitFor driver.put(pubsubTopic, msg, msgDigest, msgHash, msg.timestamp)
-        # discard crashes
+    let _ = waitFor driver.put(computeMessageHash(pubsubTopic, msg), pubsubTopic, msg)
```
Ivansete-status, line 30: Maybe a future simplification would be the following and then `computeMessageHash` within the `put` method.
```suggestion
    let _ = waitFor driver.put(pubsubTopic, msg)
```

tests/waku_archive/test_driver_postgres.nim
```diff
@@ -86,74 +72,64 @@ suite "Postgres driver":
...
 
     # Get both content topics, check ordering
     messagesRes =
-      await driver.getMessages(contentTopic = @[contentTopic1, contentTopic2])
+      await driver.getMessages(contentTopics = @[contentTopic1, contentTopic2])
     assert messagesRes.isOk(), messagesRes.error
 
     assert messagesRes.get().len == 2
-    assert messagesRes.get()[0][1].contentTopic == contentTopic1
+    assert messagesRes.get()[0][2].contentTopic == contentTopic1
```
Ivansete-status, line 101: For a better understanding, why does the index change?

tests/waku_archive/test_driver_queue.nim
```diff
@@ -18,13 +18,8 @@ proc genIndexedWakuMessage(i: int8): (Index, WakuMessage) =
...
     topic = "test-pubsub-topic"
-    cursor = Index(
-      receiverTime: Timestamp(i),
-      senderTime: Timestamp(i),
-      digest: MessageDigest(data: data),
-      pubsubTopic: topic,
-      hash: computeMessageHash(topic, message),
-    )
+    cursor =
+      Index(time: Timestamp(i), hash: computeMessageHash(topic, message), topic: topic)
```
Ivansete-status, line 22: I wonder if it'd make sense to keep using the `pubsubTopic` name to make it more explicit
```suggestion
      Index(time: Timestamp(i), hash: computeMessageHash(topic, message), pubsubTopic: topic)
```

Ivansete-status (approved):
LGTM! Thanks! :100:

tests/waku_archive/archive_utils.nim
SionoiS: Yes we could do that.

Another solution would be to refactor our internal WakuMessage type to include the hash and compute it only once at creation.

tests/waku_archive/test_driver_postgres.nim
SionoiS: The result of archive is now (hash, topic, msg) before it was (topic, msg, digest, sender timestmap, hash)

tests/waku_archive/test_driver_queue.nim
SionoiS: W8 is misread that. Yes i will take note of this and update it in #2761

Issue #2593, opened 2024-04-16
feat: RLN-proofs as a lightpush service


jm-clius:
### Problem

The challenges involved in using RLN in resource-restricted clients is [well-documented elsewhere](https://github.com/waku-org/research/issues/45).
There are currently [several promising proposals/PoCs](https://github.com/orgs/waku-org/projects/18/views/1) we've considered to make RLN compatible with variously limited environments.
However, we've recently decided to deprioritise RLN proof generation in light clients to instead focus on validating RLN + Relay as main routing solution for Waku.
This implies that full/relay nodes would have to attach RLN proofs to messages published by light clients (using [Lightpush](https://github.com/vacp2p/rfc-index/blob/main/waku/standards/core/19/lightpush.md)) as a service to these clients.

This will have several advantages:
- resource-restricted clients can be onboarded and continue using the Waku Network as before, without needing to consider RLN proofs, key generation, membership registration, etc. More onboarded apps/clients help validate RLN+Relay in the Waku Network.
- cost implications of providing the service is relatively clear (as it's mostly tied to the RLN membership costs of the service node), which makes this service a good candidate as first PoC for incentivisation (potentially replacing "Store" as PoC service in https://github.com/waku-org/nwaku/issues/1961).
- service providers can choose to provide a certain level of free publishing in the network, helping us overcome the barrier-of-entry problems associated with acquiring RLN memberships.

### Suggested solution

A simple first version of this could work as follows:
- extend [lightpush service protocol](https://github.com/waku-org/nwaku/blob/a00f350cd12b265905deeb3f8736af2125cd3a78/waku/waku_lightpush/protocol.nim) to attach RLN proofs to messages received from clients
- ensure that the proof is valid (i.e. that the service node is still under the rate limit for its membership). If the service node was able to generate a valid proof publish the message, otherwise return an error in the lightpush response.
- similar logic implemented for the [relay api](https://github.com/waku-org/nwaku/blob/master/waku/waku_api/rest/relay/handlers.nim#L146-L156) can be used as inspiration.
- basic DoS protection need to be in place for this service to be viable. See https://github.com/waku-org/nwaku/issues/2032

### Alternatives considered

In future we will want to enforce some per-client fair usage policy or other control measures, especially after we introduce a paid tier. For now, first-come-first-serve is probably a good way to get the service deployed and battle tested.

1010adigupta:
@jm-clius would like to work on this issue, please assign

jm-clius:
[...]

PR #2768, opened 2024-06-04
feat: RLN proofs as a lightpush service

shash256:
# Description
- Extend lightpush service protocol to attach RLN proofs to messages received from clients (Issue #2593)
- Ensure that the proof is valid. If the service node was able to generate a valid proof, publish the message, otherwise return an error in the lightpush response.

# Changes
- Moved PushHandler to waku_lightpush/callbacks
- It can optionally can take in WakuRLNRelay, and attach RLN proof to messages received.
- If not enabled, the messages can still get published without any additions
- Unit test covering scenario when RLN enabled (and existing e2e test covers the scenario when not enabled)

### DIFF
--- a/waku/waku_lightpush/callbacks.nim
+++ b/waku/waku_lightpush/callbacks.nim
@@ -16,21 +16,22 @@ import
  libp2p/peerid,
  stew/byteutils
 
-proc generateAndValidateRLNProof*(rlnPeer: Option[WakuRLNRelay], message: WakuMessage): Result[WakuMessage, string] =
-  # TODO: check and validate if the message already has RLN proof?
+proc checkAndGenerateRLNProof*(rlnPeer: Option[WakuRLNRelay], message: WakuMessage): Result[WakuMessage, string] =
+  # check if the message already has RLN proof
+  if message.proof.len > 0:
+    return ok(message)
+  
   if rlnPeer.isNone():
-    return ok(message) # publishing message without RLN proof
-
+    notice "Publishing message without RLN proof"
+    return ok(message)
   # generate and append RLN proof
   let
     time = getTime().toUnix()
     senderEpochTime = float64(time)
   var msgWithProof = message
-  let appendProofRes = rlnPeer.get().appendRLNProof(msgWithProof, senderEpochTime)
-  if appendProofRes.isOk():
-    return ok(msgWithProof)
-  
-  return err("RLN proof generation failed: " & appendProofRes.error)
+  rlnPeer.get().appendRLNProof(msgWithProof, senderEpochTime).isOkOr:
+    return err(error)
+  return ok(msgWithProof)
 
 proc getNilPushHandler*(): PushMessageHandler =
   return proc(
@@ -46,13 +47,12 @@ proc getRelayPushHandler*(
     peer: PeerId, pubsubTopic: string, message: WakuMessage
   ): Future[WakuLightPushResult[void]] {.async.} =
     # append RLN proof
-    let msgWithProof = generateAndValidateRLNProof(rlnPeer, message)
+    let msgWithProof = checkAndGenerateRLNProof(rlnPeer, message)
     if msgWithProof.isErr():
       return err(msgWithProof.error)
 
-    let validationRes = await wakuRelay.validateMessage(pubSubTopic, msgWithProof.value)
-    if validationRes.isErr():
-      return err(validationRes.error)
+    (await wakuRelay.validateMessage(pubSubTopic, msgWithProof.value)).isOkOr:
+      return err(error)
 
     let publishedCount = await wakuRelay.publish(pubsubTopic, msgWithProof.value)
     if publishedCount == 0:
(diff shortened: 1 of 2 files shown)


waku/waku_lightpush/callbacks.nim
```diff
@@ -0,0 +1,66 @@
...
+    rlnPeer: Option[WakuRLNRelay] = none[WakuRLNRelay]()
+): PushMessageHandler =
+  if wakuRelay.isNil():
+    debug "mounting lightpush without relay (nil)"
+    return proc(
+      peer: PeerId, pubsubTopic: string, message: WakuMessage
+    ): Future[WakuLightPushResult[void]] {.async.} =
+      return err("no waku relay found")
+  else:
+    debug "mounting lightpush with relay"
```
jm-clius, line 47: Preferably these logs, related to mounting, should be in the `waku_node` not in the callbacks module. I'd move the logic of choosing which push handler to "get" to the `waku_node`: if relay is nil, log this case and `getNilPushHandler()`, else `getRelayPushHandler()`.

waku/waku_lightpush/callbacks.nim
```diff
@@ -0,0 +1,66 @@
...
+    return proc(
+      peer: PeerId, pubsubTopic: string, message: WakuMessage
+    ): Future[WakuLightPushResult[void]] {.async.} =
+      # append RLN proof
+      let (rlnResultInfo, msgWithProof) = generateAndValidateRLNProof(rlnPeer, message)
+      if rlnResultInfo == "":
```
jm-clius, line 53: It's probably better to do an early return-on-error here, e.g.:

```
if rlnResultInfo != "":
  return err(rlnResultInfo)
```

waku/waku_lightpush/callbacks.nim
```diff
@@ -0,0 +1,66 @@
...
+ ./protocol,
+ ../waku_rln_relay,
+ ../waku_rln_relay/protocol_types,
+ ../common/ratelimit
+import
+ std/times,
+ libp2p/peerid,
+ stew/byteutils
+
+proc generateAndValidateRLNProof*(rlnPeer: Option[WakuRLNRelay], message: WakuMessage): (string, WakuMessage) =
```
jm-clius, line 19: Return a `Result[WakuMessage, cstring]`

https://status-im.github.io/nim-style-guide/errors.result.html

jm-clius (changes requested):
Thanks! Some comments below, but generally LGTM

SionoiS (changes requested):
The logic is good but code organisation I don't like.

It seams like the `RelayPushHandler` no longer has only 1 purpose. I would split the handler in 2 (proofs & relay), updating the logic in `protocol.nim` in the process.

Also, get familiar with `.isOkOr:` and `.valueOr:` those are great line savers.

jm-clius (approved):
LGTM. Some code organisation can be done before merging, but no need to re-request review.

@SionoiS it will be good to understand which specific comments are blocking approval? Minor code re-organisation can generally be done without requiring another approval or be left for subsequent PRs. I agree with most of your proposals, but think splitting the attach-RLN-proof and publish-RLN handlers is out of scope here: it would imply, among other things, changing the lightpush protocol to accept multiple push handlers with strong dependency on the order in which these handlers are called. Could be done, but I think it needs a proper design and is best left for when we introduce multiple services to the lightpush service node.

SionoiS (approved):

waku/waku_lightpush/callbacks.nim
```diff
@@ -0,0 +1,63 @@
...
+  
+  return err("RLN proof generation failed: " & appendProofRes.error)
+
+proc getNilPushHandler*(): PushMessageHandler =
+  return proc(
+    peer: PeerId, pubsubTopic: string, message: WakuMessage
+  ): Future[WakuLightPushResult[void]] {.async.} =
+    return err("no waku relay found")
+
+proc getRelayPushHandler*(
```
NagyZoltanPeter, line 41: Is it necessary to have proc producing another proc? 
Isn't it possible to define the necessary callback proc here and just use it as handler there?

tests/node/test_wakunode_lightpush.nim
```diff
@@ -103,4 +104,83 @@ suite "Waku Lightpush - End To End":
...
+      # When the client publishes a message
+      let publishResponse = await lightpushClient.lightpushPublish(
+        some(pubsubTopic), message, serverRemotePeerInfo
+      )
+
+      if not publishResponse.isOk():
+        echo "Publish failed: ", publishResponse.error()
+
+      # Then the message is relayed to the server
+      assertResultOk publishResponse
```
NagyZoltanPeter, line 186: Can we make a/some negative case here?

NagyZoltanPeter (approved):
Thank you for this great work, it is an awaited feature I think, at least I need it for lite-protocol-tester to run on RLN-ed networks.
I approve it with some comment for you to consider.

PR #2776, opened 2024-06-04
fix: postgres partitions: ensure the partition ranges are o'clock

Ivansete-status:
## Description
Also, skip the error "partition already exists" because that happens when multiple nodes interact with the same database.

## Changes

- Skip the error if the creation of a partition failed
- Perform partition creation considering o'clock time ranges and partition length of one hour each

## Issue

closes https://github.com/waku-org/nwaku/issues/2766
closes https://github.com/waku-org/nwaku/issues/2769

### DIFF
--- a/tests/all_tests_waku.nim
+++ b/tests/all_tests_waku.nim
@@ -17,7 +17,8 @@ import
   ./waku_archive/test_driver_sqlite_query,
   ./waku_archive/test_driver_sqlite,
   ./waku_archive/test_retention_policy,
-  ./waku_archive/test_waku_archive
+  ./waku_archive/test_waku_archive,
+  ./waku_archive/test_partition_manager
 
 const os* {.strdefine.} = ""
 when os == "Linux" and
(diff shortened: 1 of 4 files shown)


gabrielmer (approved):
Looks great! Thanks so much!

NagyZoltanPeter (approved):
LGTM! Thank you!

SionoiS (approved):
Thank you!

Are you sure that adding a partition from different nodes at different times does not lead to duplicate?

egg.
- T0 nodeA start and try to create partition 0-3600
- T167 nodeB start try to create partition 167-3600 (Does this return an error since both partition are different?)

There's also the case where 2 nodes try to create a partition at the same time. Is one of those node always going to error?

If yes then we gucci.

Ivansete-status:
> Thank you!
> 
> Are you sure that adding a partition from different nodes at different times does not lead to duplicate?
> 
> egg.
> 
> * T0 nodeA start and try to create partition 0-3600
> * T167 nodeB start try to create partition 167-3600 (Does this return an error since both partition are different?)
> 
> There's also the case where 2 nodes try to create a partition at the same time. Is one of those node always going to error?
> 
> If yes then we gucci.

Well, I will add protection in the upcoming PR based on the following @richard-ramos's one: https://github.com/waku-org/nwaku/pull/2736/files

PR #2777, opened 2024-06-05
chore: unifying clusterId and shardId types to uint16

gabrielmer:
# Description
Simple PR to make `clusterId` and `shardId` of type `uint16` everywhere to simplify the code and avoid casts.

## Issue

Helps the deprecation of named sharding in https://github.com/waku-org/nwaku/issues/2163

### DIFF
--- a/waku/common/envvar_serialization/reader.nim
+++ b/waku/common/envvar_serialization/reader.nim
@@ -72,7 +72,12 @@ proc readValue*[T](r: var EnvvarReader, value: var T) {.raises: [SerializationEr
   elif T is (seq or array):
     when uTypeIsPrimitives(T):
       let key = constructKey(r.prefix, r.key)
-      getValue(key, value)
+      try:
+        getValue(key, value)
+      except ValueError:
+        raise newException(
+          SerializationError, "Couldn't get value: " & getCurrentExceptionMsg()
+        )
     else:
       let key = r.key[^1]
       for i in 0 ..< value.len:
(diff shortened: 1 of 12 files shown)


SionoiS (approved):
LGTM!

NagyZoltanPeter (approved):
Yeah, makes sense.

waku/common/envvar_serialization/reader.nim
```diff
@@ -72,7 +72,12 @@ proc readValue*[T](r: var EnvvarReader, value: var T) {.raises: [SerializationEr
   elif T is (seq or array):
     when uTypeIsPrimitives(T):
       let key = constructKey(r.prefix, r.key)
-      getValue(key, value)
+      try:
+        getValue(key, value)
+      except ValueError:
+        raise newException(
+          SerializationError, "Couldn't get value: " & getCurrentExceptionMsg()
+        )
```
Ivansete-status, line 80: Is it out of the scope of the PR?

Ivansete-status (approved):
LGTM! Thanks! :100: 
Just added a minor doubt comment

waku/common/envvar_serialization/reader.nim
gabrielmer: For some reason we need it for the code to compile. After doing the changes, I get a compilation error saying that the proc `getValue` can raise a `ValueError` exception. Not sure why this error pops up with this change (why is that related to `uint16` or `uint32`).

But still, I guess that the extra fix is also an improvement and a correct usage of the proc

PR #2778, opened 2024-06-05
chore: add new index to optimize the order by storedAt

Ivansete-status:
## Description

Postgres optimization related to the `ORDER BY`

In a session with @cammellos and @richard-ramos , @cammellos recommended to add this index after seeing that `EXPLAIN ANALYZE SELECT ...` spent too much time in the sorting part. Then, after adding the offended index, the query time was enhanced.

## Issue
We don't have clear evidence that this PR will fix the following issue but it helps to have better query performance
https://github.com/waku-org/nwaku/issues/2783

### DIFF
--- a/migrations/message_store_postgres/content_script_version_5.nim
+++ b/migrations/message_store_postgres/content_script_version_5.nim
@@ -1,0 +1,6 @@
+const ContentScriptVersion_5* =
+  """
+CREATE INDEX IF NOT EXISTS i_query_storedAt ON messages (storedAt, id);
+
+UPDATE version SET version = 5 WHERE version = 4;
+"""
--- a/migrations/message_store_postgres/pg_migration_manager.nim
+++ b/migrations/message_store_postgres/pg_migration_manager.nim
@@ -1,6 +1,6 @@
 import
   content_script_version_1, content_script_version_2, content_script_version_3,
-  content_script_version_4
+  content_script_version_4, content_script_version_5
 
 type MigrationScript* = object
   version*: int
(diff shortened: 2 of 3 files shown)


richard-ramos (approved):

SionoiS (approved):

fryorcraken:
@Ivansete-status do we have a parent issue to track the fact that the node was killed/stuck because of the inefficient postgresql queries? if so, can we please link it with this PR?

Also, while I agree it makes sense to add prepare statements for regular queries, we also need to understand why the nwaku node get stuck/killed due to Postgresql DB. Is that something being looked at?
Ideally if Postgresql dies then the store service should return an error and other services such as relay, filter, etc should still be operational.

Ivansete-status:
> @Ivansete-status do we have a parent issue to track the fact that the node was killed/stuck because of the inefficient postgresql queries? if so, can we please link it with this PR?

Thanks for the comment! I've updated the PR description and created the offending bug issue

-----------------------

> Also, while I agree it makes sense to add prepare statements for regular queries, we also need to understand why the nwaku node get stuck/killed due to Postgresql DB.

I completely agree with that. We will also stress the `shards.staging` aiming to replicate the issue faster.

--------------------

>  Is that something being looked at? Ideally if Postgresql dies then the store service should return an error and other services such as relay, filter, etc should still be operational.

Yes absolutely, I'm checking it. The odd thing is that the whole database got blocked until some pending `SELECT` queries were killed. On the other hand, after killing the `nwaku` node it resumes well again when it starts.

PR #2779, opened 2024-06-05
chore: deprecating pubsub topic config

gabrielmer:
# Description
Deprecating the `pubsub-topic` CLI config in favor of `cluster-id`, `shards` and a new config item `network-shards` stating the amount of shards in the given `cluster-id` for autosharding purposes

Renamed the CLI config `ProtectedTopic` to `ProtectedShard`. The idea is to avoid ambiguity so the user only has to be familiar with the terms `cluster` and `shard`.

The word `pubsubTopic` in our codebase will refer solely to the string representation of `clusterId`+`shard` that we use in the `libp2p` layer.


fryorcraken (commented):
.

gabrielmer:
Closing this PR as it will be replaced by https://github.com/waku-org/nwaku/pull/2997

PR #2784, opened 2024-06-06
chore: postgres_driver - acquire/release advisory lock when creating partitions

Ivansete-status:
## Description
Add additional protection, in the shape of an advisory lock, when adding new partitions to the database.
This is interesting from the PoV of multiple nodes accessing the same database.

## Issue
Not evidence that this PR will fix the following issue but it will help
https://github.com/waku-org/nwaku/issues/2783


gabrielmer (approved):
LGTM! Thankss

richard-ramos (approved):

SionoiS (approved):

PR #2794, opened 2024-06-09
fix: more detailed logs to differentiate shards with peers

DarshanBPatel:
Adds more information to the log, allowing users to differentiate shards and their connections.

Example :- 

If shards match the peer, the previous log entry would be :- "reason = no shards in common"  
Now it's look like this :- "reason = no shards in common: my_shards =  [ x, y, z ]  others_shards = [ a, b, c ]"

where the number within brackets represents the shard number.


Ivansete-status (approved):
Gorgeous! Thanks for it! :100: 
May kindly add an example in the description, showing how it may appear please?
Cheers

gabrielmer (approved):
LGTM! Thanks so much 🔥 🔥

PR #2797, opened 2024-06-10
chore: Remove filter-timeout configuration option as it remained after removing legacy filter where it belong to.

NagyZoltanPeter:
# Description

Just found this anomaly in the code while looking up answers during our meeting...
It is a leftover, nowhere used config option, related to legacy filter that was removed versions before.

also in docs: https://github.com/waku-org/docs.waku.org/pull/200

# Changes

- [X] Removed CLI confif filter-timout

### DIFF
--- a/waku/factory/external_config.nim
+++ b/waku/factory/external_config.nim
@@ -366,13 +366,6 @@ type WakuNodeConf* = object
       name: "filternode"
     .}: string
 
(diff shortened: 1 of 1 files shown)


gabrielmer (approved):
Ooh great catch! Thanks so much!

SionoiS (approved):
Nice find!

PR #2799, opened 2024-06-11
fix: multi nat initialization causing dead lock in waku tests + serialize test runs to avoid timing and port occupied issues

NagyZoltanPeter:
# Description

### Multiple NAT module initialization cause dead lock in tests
In case nat setup found proper device to make the port mapping (that does not happens in jenkins CI) could cause multiple initilalization of nim-eth/nat module.
That module is not designed for that and changing this needs a bigger rework of that module.
The root cause of the issue with multiple initialization inside one application run leads to multiple remapping thread created. That thread is responsible for refreshing the port mapping on the router if needed.
As such multiple thread created but only the last one tracked caused dead lock in the shut down mechanism of it as the used Channel[bool] single module variable locking mechanism do not handle such situation and remains blocked.

Simple workaround is applied: waku nat module prevents multiple initialization of nim-eth/nat module and will behave as no proper device would be found (which is still an ok case for testing).

### Reduce tests flakyness

In order to reduce probability of timing issue during CI test runs also possibility of failed tests because of ports already in use we made test execution sequential.

## How to test

1. make testwakunode2

## Issue
https://github.com/waku-org/nwaku/issues/2628

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,8 +12,8 @@ concurrency:
 
 env:
   NPROC: 1
-  MAKEFLAGS: ""
-  NIMFLAGS: "--colors:off -d:chronicles_colors:none"
+  MAKEFLAGS: "-j${NPROC}"
+  NIMFLAGS: "--parallelBuild:${NPROC} --colors:off -d:chronicles_colors:none"
 
 jobs:
   changes: # changes detection
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,7 @@ concurrency:
   cancel-in-progress: true
 
 env:
-  NPROC: 1
+  NPROC: 2
   MAKEFLAGS: "-j${NPROC}"
   NIMFLAGS: "--parallelBuild:${NPROC} --colors:off -d:chronicles_colors:none"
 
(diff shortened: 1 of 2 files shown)


gabrielmer (approved):
Amazingg, thanks so much!

NagyZoltanPeter:
@Ivansete-status, @gabrielmer
cc: @SionoiS , @DarshanBPatel 

In addition to the original PR I still encountered flaky tests mostly on mac-os. 
Like REST ports already in use, and some timing sensitive tests.
with @Ivansete-status we made our bets and tried to serialize the tests by not allowing paralel runs for nim compiler.
This seems eliminates mostly the issues, although seems some variance in the tests can come. 
Finally test run times seems not much affected, still around 27-31 mins.

SionoiS (approved):
Zoltan the :detective:.

Very nice!

.github/workflows/ci.yml
```diff
@@ -111,15 +111,15 @@ jobs:
...
-
           postgres_enabled=0
           if [ ${{ runner.os }} == "Linux" ]; then
+            sudo docker run --rm -d -e POSTGRES_PASSWORD=test123 -p 5432:5432 postgres:15.4-alpine3.18
             postgres_enabled=1
           fi
 
+          export MAKEFLAGS="-j1"
+          export NIMFLAGS="--colors:off -d:chronicles_colors:none"
+
```
Ivansete-status, line 122: That LGTM but shall we add that in a separate PR so that is clear the commit that applied it?

Ivansete-status (approved):
LGTM! Thanks! Just added a comment to split it in separate PRs

.github/workflows/ci.yml
Ivansete-status: > That LGTM but shall we add that in a separate PR so that is clear the commit that applied it?

@NagyZoltanPeter - maybe easier to just update the PR title and description to also reflect the CI test change :)

NagyZoltanPeter: :-) agree! I do it.

PR #2800, opened 2024-06-11
chore: adding observers for message logging

gabrielmer:
# Description

Adding observers so the hash and origin/destination of every message sent and received is logged. This will help with our reliability simulation efforts.

This adds lots of logs, so there's always the option to have it under a compilation flag. Let me know if that's preferable or if it's ok for now and we'll make them optional at a later stage.

### Note
We're finally adding observers only for messages received, as no additional valuable information is currently added by logging sent messages. If necessary at any point, they can be easily activated.

# Changes

- [x] added `nim-libp2p` observers for incoming and outgoing messages to log message data

Example output:

image

### DIFF
--- a/waku/node/waku_node.nim
+++ b/waku/node/waku_node.nim
@@ -394,7 +394,7 @@ proc generateRelayObserver(w: WakuRelay): PubSubObserver =
     for msg in msgs.messages:
       let msg_id = w.msgIdProvider(msg).valueOr:
         warn "Error generating message id",
-          from_peer = peer.peerId, topic = msg.topic, error = $error
+          from_peer_id = peer.peerId, topic = msg.topic, error = $error
         continue
 
       let msg_id_short = shortLog(msg_id)
@@ -402,7 +402,7 @@ proc generateRelayObserver(w: WakuRelay): PubSubObserver =
       let wakuMessage = WakuMessage.decode(msg.data).valueOr:
         warn "Error decoding to Waku Message",
           msg_id = msg_id_short,
-          from_peer = peer.peerId,
+          from_peer_id = peer.peerId,
           topic = msg.topic,
           error = $error
         continue
@@ -410,16 +410,16 @@ proc generateRelayObserver(w: WakuRelay): PubSubObserver =
       let msg_hash = computeMessageHash(msg.topic, wakuMessage).to0xHex()
 
       if onRecv:
-        info "received message",
+        notice "received relay message",
           msg_hash = msg_hash,
           msg_id = msg_id_short,
-          from_peer = peer.peerId,
+          from_peer_id = peer.peerId,
           topic = msg.topic
       else:
-        info "sent message",
+        notice "sent relay message",
           msg_hash = msg_hash,
           msg_id = msg_id_short,
-          to_peer = peer.peerId,
+          to_peer_id = peer.peerId,
           topic = msg.topic
 
   proc onRecv(peer: PubSubPeer, msgs: var RPCMsg) =
--- a/waku/node/waku_node.nim
+++ b/waku/node/waku_node.nim
@@ -389,9 +389,31 @@ proc startRelay*(node: WakuNode) {.async.} =
 
   info "relay started successfully"
 
-proc generateRelayObserver(): PubSubObserver =
-  proc onRecv(peer: PubSubPeer, msgs: var RPCMsg) {.gcsafe, raises: [].} =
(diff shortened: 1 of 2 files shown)


waku/node/waku_node.nim
```diff
@@ -389,6 +389,47 @@ proc startRelay*(node: WakuNode) {.async.} =
...
+          error = $error
+        continue
+
+      let msg_hash = computeMessageHash(msg.topic, wakuMessage).to0xHex()
+
+      if onRecv:
+        info "received message",
+          msg_hash = msg_hash,
+          msg_id = msg_id_short,
+          from_peer = peer.peerId,
```
Ivansete-status, line 416:
```suggestion
          from_peer_id = peer.peerId,
```

waku/node/waku_node.nim
```diff
@@ -389,6 +389,47 @@ proc startRelay*(node: WakuNode) {.async.} =
...
+          topic = msg.topic
+      else:
+        info "sent message",
+          msg_hash = msg_hash,
+          msg_id = msg_id_short,
+          to_peer = peer.peerId,
```
Ivansete-status, line 422:
```suggestion
          to_peer_id = peer.peerId,
```

waku/node/waku_node.nim
```diff
@@ -389,6 +389,47 @@ proc startRelay*(node: WakuNode) {.async.} =
 
   info "relay started successfully"
 
+proc generateRelayObserver(w: WakuRelay): PubSubObserver =
+  proc logMessageInfo(peer: PubSubPeer, msgs: var RPCMsg, onRecv: bool) =
+    for msg in msgs.messages:
+      let msg_id = w.msgIdProvider(msg).valueOr:
+        warn "Error generating message id",
+          from_peer = peer.peerId, topic = msg.topic, error = $error
```
Ivansete-status, line 397:
```suggestion
          from_peer_id = peer.peerId, topic = msg.topic, error = $error
```

waku/node/waku_node.nim
```diff
@@ -389,6 +389,47 @@ proc startRelay*(node: WakuNode) {.async.} =
...
+      let msg_hash = computeMessageHash(msg.topic, wakuMessage).to0xHex()
+
+      if onRecv:
+        info "received message",
+          msg_hash = msg_hash,
+          msg_id = msg_id_short,
+          from_peer = peer.peerId,
+          topic = msg.topic
+      else:
+        info "sent message",
```
Ivansete-status, line 419:
```suggestion
        notice "sent relay message",
```

waku/node/waku_node.nim
```diff
@@ -389,6 +389,47 @@ proc startRelay*(node: WakuNode) {.async.} =
...
+          msg_id = msg_id_short,
+          from_peer = peer.peerId,
+          topic = msg.topic,
+          error = $error
+        continue
+
+      let msg_hash = computeMessageHash(msg.topic, wakuMessage).to0xHex()
+
+      if onRecv:
+        info "received message",
```
Ivansete-status, line 413:
```suggestion
        notice "received relay message",
```

Ivansete-status (approved):
LGTM! Thanks! A very clean PR :partying_face: 
I just added some comments that I hope you find useful.
On the other hand, we might need to review the following snippet because I have the feeling that we will duplicate some logs when a message is rx.

I think we could potentially move that logic to the new one, when a message is received.

https://github.com/waku-org/nwaku/blob/15d578ad87dd06d9ec99d0d4a7595b8bc7ecbeb9/waku/node/waku_node.nim#L227-L238

NagyZoltanPeter (approved):
It is super cool!
About the compilation flag question. I'm on the side not to have such logs when out to production as it is not just overwhelming amount of logs but reveals way toooo much information about the network and messages.

Ivansete-status:
> It is super cool! About the compilation flag question. I'm on the side not to have such logs when out to production as it is not just overwhelming amount of logs but reveals way toooo much information about the network and messages.

I see your point. Nevertheless, I think for now it is interesting to have it in that way because it is much simpler than having a compilation flag and we are just disclosing a bit of information about the neighbors but doesn't bring a picture of the whole network.

We can revisit that in the future if needed but for now is fine IMHO.
( cc @gabrielmer )

gabrielmer:
I think I'll merge it as it is now, and if needed we can add the compilation flag anytime :)

PR #2804, opened 2024-06-12
fix: Use random ports in rest tests instead of fixed ones

NagyZoltanPeter:
# Description
Found that some of the random failures in CI comes from old fashioned REST API tests that use fixed ports for test setup.
Now this changed to OS selected random free ports all remaining places.

I expect more stable CI tests from this change. No functionality changed.

# Changes

- [X] All rest test port setup now uses Port(0)

## How to test

1. Just run all tests


gabrielmer (approved):
Thanks so much!

SionoiS (approved):
Nice. Flaky CI begone!

PR #2809, opened 2024-06-13
fix: postgres_driver - better sync lock in partition creation

Ivansete-status:
# Description
With the previous approach it happened cases where one connection acquired the lock and other connection tried to release the same lock, causing an unrecoverable failure which made the node to end.

## Issue

closes https://github.com/waku-org/nwaku/issues/2783


gabrielmer (approved):
Amazing! Thanks so much! 😍

NagyZoltanPeter (approved):
It is truely great! Thank you!!!

PR #2815, opened 2024-06-17
fix: revert "chore: adding observers for message logging (#2800)"

gabrielmer:
This reverts commit b52286524b243feb70ae5f08cafae26f95b0bf50.

# Description
The approach of using observers for logging seems to be extremely inefficient, as observers are triggered before checking whether we have received a message or not. This implies that they get triggered for repeated messages.

On the other hand, increasing the metrics in the observer is a bug, as repeated messages would be counted.

We will need then to create special versions with extra logs for simulations in a branch separate from master.

Thanks @Ivansete-status for raising the issue!

CC @AlbertoSoutullo 

# Changes

- [x] reverting logger observers commit

### DIFF
--- a/waku/node/waku_node.nim
+++ b/waku/node/waku_node.nim
@@ -225,6 +225,19 @@ proc registerRelayDefaultHandler(node: WakuNode, topic: PubsubTopic) =
   if node.wakuRelay.isSubscribed(topic):
     return
 
+  proc traceHandler(topic: PubsubTopic, msg: WakuMessage) {.async, gcsafe.} =
+    notice "waku.relay received",
+      my_peer_id = node.peerId,
+      pubsubTopic = topic,
+      msg_hash = topic.computeMessageHash(msg).to0xHex(),
+      receivedTime = getNowInNanosecondTime(),
+      payloadSizeBytes = msg.payload.len
+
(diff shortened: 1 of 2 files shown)


Ivansete-status (approved):
LGTM! Thanks for it! :100:

waku/node/waku_node.nim
```diff
@@ -225,6 +225,19 @@ proc registerRelayDefaultHandler(node: WakuNode, topic: PubsubTopic) =
   if node.wakuRelay.isSubscribed(topic):
     return
 
+  proc traceHandler(topic: PubsubTopic, msg: WakuMessage) {.async, gcsafe.} =
+    notice "waku.relay received",
```
NagyZoltanPeter, line 229:
```suggestion
    let msg_hash = topic.computeMessageHash(msg).to0xHex()
    notice "waku.relay received",
```
Nitpick I know to separate compute from log...

NagyZoltanPeter (approved):
Sorry to hear it did not worked out. LGTM btw.

waku/node/waku_node.nim
gabrielmer: Niiice, thank you! Done :))

PR #2816, opened 2024-06-17
feat: Add incentivization PoC with dummy request-response protocol

s-tikhomirov:
# Description

This is a proof-of-concept implementation of a dummy request-response protocol that illustrates how eligibility proofs and eligibility statuses (see [spec 73](https://github.com/vacp2p/rfc/tree/master/content/docs/rfcs/73)) can be used in the context of light protocols. The PoC code is only called from within tests. In the tests, a client creates a dummy request with an eligibility proof attached, and sends it to the server. The server checks the validity of the eligibility proof, and returns a response formed accordingly.

The eligibility proof is a transaction ID. For testing purposes, a random confirmed transaction on Sepolia testnet is used. The server only checks that a transaction with the corresponding txid (i.e., hash) exists. If so, the proof is considered valid.

# Changes

- [x] add `incentivization` directory with a dummy client-server protocol;
- [x] add tests that illustrate the usage of eligibility structures in the context of light protocols.

## How to test

1. `./env.sh bash`
1. `nim c -r ./tests/incentivization/test_all.nim`

## Issue

Related to #1961, although this PoS is unrelated to Store protocol per se.


jm-clius (changes requested):
I think we're overcomplicating this POC. For one, we shouldn't add test-only code to master outside of the test suites. But I also don't see the need to introduce a fully specified dummy protocol. We simply need to mimic request-response interactions with eligibility proof and status checks embedded, but this can be done within a single unit test without introducing any extensive protocol scaffolding. In other words, we show the generic flow of req-resp protocol with eligibility checks:
Within a single test module:
1. sendRequestWithEligibilityProof
2. checkEligibility
3. sendResponseWithEligibilityStatus

(1) and (3) would in reality send only the raw encoded eligibility RPCs. Within the test setup code you may have to introduce a dummy protocol ID, but this should be minimal.
(2) is the main focus of the POC and should show that we can lookup a raw tx_id to "check" that payment has been made. It may introduce some basic helper functions in `waku/incentivization` that will be reusable in future.

This POC simply illustrates our thinking around incentivisation so should be kept simple.

jm-clius (commented):
After discussion with @s-tikhomirov, I suggest that we keep the Dummy protocol, but move this feature only to an experimental feature branch and cut the scope of the POC significantly. Created a new issue to explain reduced scope here: https://github.com/waku-org/nwaku/issues/2847
Happy to approve merging this to a feature branch.

jm-clius (approved):
Approving merge to feature branch. See https://github.com/waku-org/nwaku/issues/2847

Ivansete-status (approved):
LGTM! Thanks!

waku/incentivization/txid_proof.nim
```diff
@@ -0,0 +1,48 @@
...
+      TxHashNonExisting)
+  let txHashAsBytes = @(txHash.bytes())
+  EligibilityProof(proofOfPayment: some(txHashAsBytes))
+
+proc genDummyRequestWithTxIdEligibilityProof*(proofValid: bool, requestId: string = ""): DummyRequest =
+  let eligibilityProof = genTxIdEligibilityProof(proofValid)
+  result.requestId = requestId
+  result.eligibilityProof = eligibilityProof
+
+proc checkTxIdExists(txHash: TxHash): Future[bool] {.async.} =
```
jm-clius, line 33: This function should either take a `web3` provider or the `EthClient` string as argument. This would allow us to move the hardcoded test constants to the test modules.

waku/incentivization/txid_proof.nim
```diff
@@ -0,0 +1,48 @@
...
+const EthClient = "https://sepolia.infura.io/v3/470c2e9a16f24057aee6660081729fb9"
+
+proc genTxIdEligibilityProof*(txIdExists: bool): EligibilityProof = 
+  let txHash: TxHash = (
+    if txIdExists:
+      TxHashExisting
+    else:
+      TxHashNonExisting)
+  let txHashAsBytes = @(txHash.bytes())
+  EligibilityProof(proofOfPayment: some(txHashAsBytes))
```
jm-clius, line 26: This seems to me to be test utilities with some hardcoded test values. If we move this to the unit test module, we'll keep the incentivization code cleaner.

jm-clius (approved):
Thanks! I think this should close the initial POC for now on merge. We can revisit once we add more complex incentivization logic, reputation, etc.

PR #2818, opened 2024-06-18
fix: update peers ENRs in peer store in case they are updated

### LINKED ISSUE #2817: bug: peerStore should update ENRs of existing peers in case they change
### Problem
If a peer is added to the peer store and it dynamically changes its ENR, our peer store won't update it. It will currently check that the multiaddress and public key are matched to the peerId and that some ENR is saved. If so, we exit early.

This happens in the following lines in `addPeer()`:
https://github.com/waku-org/nwaku/blob/93e9ba22aa01121217fd7e44d592ae30b8074599/waku/node/peer_manager/peer_manager.nim#L128-L133

But if ENR changed, we won't update it and still exit early.

We should check if the ENR has changed and if so, update it in the peer store.

gabrielmer:
# Description

Checking if the ENR of an incoming peer is the same as the one we have saved and if not, update it in the peer store.

# Changes

- [x] updating peer store for existing peers with changed ENRs

## Issue

closes #2817

### DIFF
--- a/waku/node/peer_manager/peer_manager.nim
+++ b/waku/node/peer_manager/peer_manager.nim
@@ -128,9 +128,14 @@ proc addPeer*(pm: PeerManager, remotePeerInfo: RemotePeerInfo, origin = UnknownO
   if pm.peerStore[AddressBook][remotePeerInfo.peerId] == remotePeerInfo.addrs and
       pm.peerStore[KeyBook][remotePeerInfo.peerId] == remotePeerInfo.publicKey and
       pm.peerStore[ENRBook][remotePeerInfo.peerId].raw.len > 0:
-    trace "peer already managed and ENR info is already saved",
-      remote_peer_id = $remotePeerInfo.peerId
-    return
+    if remotePeerInfo.enr.isNone():
+      trace "peer already managed and incoming ENR is empty"
+      return
+
+    if remotePeerInfo.enr.get().raw == pm.peerStore[ENRBook][remotePeerInfo.peerId].raw:
+      trace "peer already managed and ENR info is already saved",
+        remote_peer_id = $remotePeerInfo.peerId
+      return
 
   trace "Adding peer to manager",
     peerId = remotePeerInfo.peerId, addresses = remotePeerInfo.addrs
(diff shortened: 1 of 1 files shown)


richard-ramos:
Ah do note that this is a scenario that i have seen in go-waku. I do not know if such thing can happen in the nim implementation

gabrielmer:
> PR looks good to me, but I wonder if it could be the case that when discovering you receive the same peer but with older information. Since we are not looking at the sequence number of the enr, we might end up overriding the peer information with outdated info

Oh great catch! So basically before updating the ENR, we have to also check for the `seqNum` to be equal or greater than the one in the ENR currently saved in the peer store?

If that makes sense, will implement that fix too :)

richard-ramos:
> Oh great catch! So basically before updating the ENR, we have to also check for the `seqNum` to be equal or greater than the one in the ENR currently saved in the peer store?

Yes, that's what we do in go-waku! we check if the sequence number is greater to update the peer info!

richard-ramos (approved):

waku/node/peer_manager/peer_manager.nim
```diff
@@ -128,9 +128,15 @@ proc addPeer*(pm: PeerManager, remotePeerInfo: RemotePeerInfo, origin = UnknownO
   if pm.peerStore[AddressBook][remotePeerInfo.peerId] == remotePeerInfo.addrs and
       pm.peerStore[KeyBook][remotePeerInfo.peerId] == remotePeerInfo.publicKey and
       pm.peerStore[ENRBook][remotePeerInfo.peerId].raw.len > 0:
-    trace "peer already managed and ENR info is already saved",
-      remote_peer_id = $remotePeerInfo.peerId
-    return
+    let incomingEnr = remotePeerInfo.enr.valueOr:
+      trace "peer already managed and incoming ENR is empty"
```
Ivansete-status, line 132: I wonder if we could also print the `remote_peer_id` in this case too

Ivansete-status (approved):
LGTM! Thanks! :100:

waku/node/peer_manager/peer_manager.nim
gabrielmer: Great catch! 😍 Added it.
Thank you!

PR #2821, opened 2024-06-18
test: Spammer

vpavlin:
> DO NOT MERGE!

# Description
This adds a simple async loop which publishes a message in a given interval.

It is based on https://github.com/waku-org/nwaku/pull/2759 and attempts to make minimal changes.

# Changes

- [ ] adds extra config options
- [ ] adds async loop publishing messages
- [ ] automatically registers a new membership on start

### DIFF
--- a/apps/wakunode2/spammer.nim
+++ b/apps/wakunode2/spammer.nim
@@ -6,54 +6,84 @@ else:
 import
   chronos,
   chronicles,
-  confutils,
-  confutils/defs,
-  confutils/std/net,
-  confutils/toml/defs as confTomlDefs,
-  confutils/toml/std/net as confTomlNet,
   stew/[byteutils, results],
   std/times,
-  libp2p/protocols/pubsub/gossipsub
+  libp2p/protocols/pubsub/gossipsub,
+  strutils
 
 import
   ../../waku/factory/waku,
+  ../../waku/factory/external_config,
   ../../waku/waku_core,
+  ../../waku/waku_relay,
   ../../waku/node/waku_node,
-  ../../waku/common/confutils/envvar/defs as confEnvvarDefs,
-  ../../waku/common/confutils/envvar/std/net as confEnvvarNet
+  ../../waku/node/peer_manager/peer_manager,
+  ../../waku/waku_rln_relay/rln_relay,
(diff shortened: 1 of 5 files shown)


apps/wakunode2/spammer.nim
```diff
@@ -0,0 +1,119 @@
...
+  return ok()
+
+proc burstPublish(
+    waku: Waku, conf: WakuNodeConf, contentTopic: ContentTopic
+) {.async.} =
+  var futures: seq[Future[Result[void, string]]]
+  var i: uint64 = 0
+  var start = getTime().toUnixFloat()
+
+  while i < conf.rlnRelayUserMessageLimit:
```
alrevuelta, line 58: maybe `i` not needed?
and here
for i in conf.rlnRelayUserMessageLimit

apps/wakunode2/spammer.nim
```diff
@@ -0,0 +1,119 @@
...
+    waku: Waku, conf: WakuNodeConf, contentTopic: ContentTopic = "/spammer/0/test/plain"
+) {.async.} =
+  if not conf.spammerEnable:
+    return
+
+  if not conf.rlnRelay:
+    error "RLN not configured!"
+    quit(QuitFailure)
+
+  var gotPeers = false
```
alrevuelta, line 97: gotPeers not needed?
while true:
  if xx:
    break

alrevuelta (approved):
not to be merged but lgtm thanks. very useful for testing.

apps/wakunode2/spammer.nim
vpavlin: This is not how Nim works:-P There is a function I could use to turn the msg limit into iterator, but I am not sure it is worth searching for it:D

apps/wakunode2/spammer.nim
vpavlin: Fair enough:) fixed:)

apps/wakunode2/spammer.nim
alrevuelta: ah yeah my bad. gues for i in 0..<conf.rlnRelayUserMessageLimit 
no big deal tho

Ivansete-status:
@vpavlin - what is the outstanding point of this one?

PR #2823, opened 2024-06-18
chore: stop connecting to out peers until target is reached

gabrielmer:
# Description
We're currently attempting establish more `out connections` than our target. We are trying to connect 10 peers at a time, and we still dial 10 even if our target is less than 10.

When investigating https://github.com/waku-org/nwaku/issues/2780, I noticed that the issue started right after we tried to connect to more nodes than we could.

Making a small fix so we stick more precisely to our targets.

# Changes

- [x] only dial to the peers necessary to reach `outRelayPeersTarget`

### DIFF
--- a/waku/node/peer_manager/peer_manager.nim
+++ b/waku/node/peer_manager/peer_manager.nim
@@ -725,7 +725,8 @@ proc connectToRelayPeers*(pm: PeerManager) {.async.} =
   shuffle(outsideBackoffPeers)
 
   var index = 0
-  var numPendingConnReqs = outsideBackoffPeers.len
+  var numPendingConnReqs =
+    min(outsideBackoffPeers.len, pm.outRelayPeersTarget - outRelayPeers.len)
     ## number of outstanding connection requests
 
   while numPendingConnReqs > 0 and outRelayPeers.len < pm.outRelayPeersTarget:


gabrielmer:
> Thanks for that @gabrielmer ! However, there might be something that I don't get 100% because I was assuming that we already achieve that goal from the following line:
> 
> https://github.com/waku-org/nwaku/blob/93e9ba22aa01121217fd7e44d592ae30b8074599/waku/node/peer_manager/peer_manager.nim#L732

So before the fix that I added, `numPendingConnReqs` is initialized to `outsideBackoffPeers.len`, which is all the peers that we have saved and can connect to. But this number does not take into account our target, so if we have more possible peers than our target, we can currently overshoot.

That's why I'm fixing `numPendingConnReqs` to match exactly whatever is left for our target.

Ivansete-status (approved):
LGTM! Thanks for the clarification! :100: :partying_face: 

Thanks to this enhancement, I understand that the precondition controlled in the next point: https://github.com/waku-org/nwaku/blob/317c83dc1deace8a3d763f5efffa1e0e0dced600/waku/node/peer_manager/peer_manager.nim#L724

...still remains at the end of the `connectToRelayPeers` proc. In other words, we make sure that the this condition is met: `outRelayPeers.len < pm.outRelayPeersTarget`

Great catch :raised_hands: !

gabrielmer:
> Thanks to this enhancement, I understand that the precondition controlled in the next point:
> 
> https://github.com/waku-org/nwaku/blob/317c83dc1deace8a3d763f5efffa1e0e0dced600/waku/node/peer_manager/peer_manager.nim#L724
> 
> ...still remains at the end of the `connectToRelayPeers` proc. In other words, we make sure that the this condition is met: `outRelayPeers.len < pm.outRelayPeersTarget`

Exactly! This was still controlled but it wasn't precise, we would overshoot most of the times and could dial up to 9 more peers than the target

NagyZoltanPeter (approved):
Yes, it looks ok.

Issue #2780, opened 2024-06-05
bug: Future operation cancelled flooding logs


AlbertoSoutullo:
### Problem
From DST, we are testing version `v0.28`, and we noticed a "regression" in terms of how our simulations get stable.

As a summary with respect to explaining the simulation: We use 45 nWaku nodes as bootstrap nodes, and then 50 normal nWaku nodes. Each normal nWaku node will have 3 `--discv5-bootstrap-node` at random. They will also use `--relay-peer-exchange`. 
Before we inject traffic in the simulation, we make sure that every node is "healthy". We obtain this information from `libp2p_gossipsub_healthy_peers_topics `. This returns the number of topics with peers in the range `d_low < nºpeers < d_high` (in our case, just 1).

In version `v0.27`, everything is ok. The aforementioned scenario reach healthy state in less than 1 minute in a constant manner.
In version `v0.28`, this behavior is very inconsistent. It can vary from a couple of minutes to more than 10 minutes (cancelled at that point). 
The only difference between these 2 cases is the nWaku version, everything else is exactly the same.

What we found so far is that the following line is very present in the nodes that can't reach a healthy state:
`Dialing canceled                           topics="libp2p dialer" tid=7 file=dialer.nim:67 err="Future operation cancelled!"`

### Impact
It slows down considerably the setup time of the simulations for DST. Also, as in the previous version this was not happening, we assume there is an unexpected behavior / possible bug that might be good to look into.

### To reproduce
1. Instantiate 40~ nWaku nodes that will act as bootstrap.
2. Use 3 ENRs from these nodes in another deployment of 50~ nWaku Nodes.
3. Look at DEBUG logs to see the line previously mentioned.
4. Look how `libp2p_gossipsub_healthy_peers_topics` metric varies.

### Expected behavior
No connection getting cancelled if there is no interruption.

### Screenshots/logs
This log is an example of v0.28:
[nodes-11_waku.log](https://github.com/user-attachments/files/15592417/nodes-11_waku.log)

### nwaku version/commit hash
v0.28.0-rc.0

AlbertoSoutullo:
Hello, I come back with more news about this.

I just tested this again. Maybe this information helps:
- This issue happens on `v0.26` and `v0.28.0-rc.0`. It happens so often that we literally cannot even run simulations.
- Version `v0.27` runs perfectly fine. No problem whatsoever. If this log line appears, it is almost unnoticeable.
[...]

PR #2824, opened 2024-06-19
fix: adding peer exchange peers to the peerStore

gabrielmer:
# Description
As soon we receive peers in Peer Exchange, we currently try to connect to all of them directly in a very aggressive way.

This does not take into account our connection targets and limits. Therefore, we should first add the peers to the Peer Store, and the connectivity loop will be in charge to choose and connect to the amount of peers necessary to reach its targets.

# Changes

- [x] adding Peer Exchange peers to the Peer Store instead of trying to directly connect to all of them 

## Issue

We saw this case happen and potentially cause issues in #2780

### DIFF
--- a/waku/factory/node_factory.nim
+++ b/waku/factory/node_factory.nim
@@ -135,11 +135,12 @@ proc setupProtocols(
         # only peers with populated records
         .mapIt(toRemotePeerInfo(it.record.get()))
 
-      debug "connecting to exchanged peers",
+      debug "adding exchanged peers",
         src = peer, topic = topic, numPeers = exchangedPeers.len
 
-      # asyncSpawn, as we don't want to block here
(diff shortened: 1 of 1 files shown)


gabrielmer:
I tried to also take care of 
https://github.com/waku-org/nwaku/blob/e41aa578b2bb95ffb5b450ea8c4f232417217c1e/waku/factory/node_factory.nim#L133

But noticed that in the handler we don't receive the exchanged peers ENRs so we can't filter by the cluster/shards. Maybe I'm missing something.

In any case, I think that the Peer Manager itself should be in charge of the filtering peers by shards so we don't duplicate the logic. This should be taken care of when we activating sharded peer management.

Ivansete-status (approved):
LGTM! Thanks for it! :100:

NagyZoltanPeter (approved):
Sounds reasonable to me. Thank you!!!

PR #2826, opened 2024-06-20
chore: better descriptive log

darshankabariya:
Previously, there was confusion in distinguishing between the content and the pubsub topic in logs. Now, the logs are more descriptive, clearly specifying which content-topic pair has the issue.

### DIFF
--- a/waku/factory/node_factory.nim
+++ b/waku/factory/node_factory.nim
@@ -149,6 +149,8 @@ proc setupProtocols(
       conf.contentTopics.mapIt(node.wakuSharding.getShard(it).expect("Valid Shard"))
     let pubsubTopics = conf.pubsubTopics & shards
 
+    debug "Shards created for content filter", shards = shards
+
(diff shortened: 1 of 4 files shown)


waku/factory/node_factory.nim
```diff
@@ -148,6 +148,8 @@ proc setupProtocols(
       conf.contentTopics.mapIt(node.wakuSharding.getShard(it).expect("Valid Shard"))
     let pubsubTopics = conf.pubsubTopics & shards
 
+    debug "Shards created for content filter", shards = shards
```
Ivansete-status, line 151: Suggestion to give a bit more detail when we have contentTopics as a parameter
```suggestion
    if conf.contentTopics.len > 0:
      debug "Shards created from content topics", contentTopics = conf.contentTopics, shards = shards
```

gabrielmer (approved):
LGTM, thanks!

PR #2831, opened 2024-06-20
fix: only set disconnect time on left event

gabrielmer:
# Description

Small fix to only set `disconnectTime` when a node is actually disconnecting

# Changes

- [x] only setting `disconnectTime` in peer storage when a node is disconnects

## Issue

advances https://github.com/waku-org/pm/issues/172


Ivansete-status (approved):
LGTM! Thanks for it! :100:

NagyZoltanPeter (approved):
LGTM.

PR #2835, opened 2024-06-20
chore: update content parsing for filter

### LINKED ISSUE #2834: bug: Unexpected behavior due to relay and filter presence in config-file
In some scenarios, an exception is not thrown when there is an issue with the content topic.

**Details:**
         Expected Behavior: The system should throw an exception whenever there is an issue with the content topic.
         Actual Behavior: The system's behavior varies based on the configuration settings of relay and filter.

**Scenarios:**
         Relay is True: An exception is thrown when there is an issue with the content topic.
         Relay is False: No exception is thrown even if there is an issue with the content topic.

**Combined Case:**
         When both relay and filter are present in the configuration file and both are set to false, no exception is thrown when there is an issue with the content topic.
         However, if only filter is present and set to false, and relay is absent (defaulting to false), an exception is thrown when there is an issue with the content topic.

**Contradiction:**
         If both relay and filter are present and set to false, no exception is thrown.
         If only filter is present and set to false (with relay being absent and therefore false), an exception is thrown when there is an issue with the content topic.

darshankabariya:
close #2834

This is a minor change to fix bug #2834.

### DIFF
--- a/waku/factory/node_factory.nim
+++ b/waku/factory/node_factory.nim
@@ -144,7 +144,7 @@ proc setupProtocols(
 
     peerExchangeHandler = some(handlePeerExchange)
 
-  if conf.relay:
+  if conf.relay or conf.filter:
     let shards =
       conf.contentTopics.mapIt(node.wakuSharding.getShard(it).expect("Valid Shard"))
     let pubsubTopics = conf.pubsubTopics & shards


NagyZoltanPeter (approved):
Great catch, thank you!

gabrielmer:
I think this should be run only if relay is configured. If filter is configured and relay isn't, with these changes we would still mount relay

https://github.com/waku-org/nwaku/blob/9691dfb5e2f2e12935045be05f65a6c737b973fe/waku/factory/node_factory.nim#L160-L168

darshankabariya:
> I think this should be run only if relay is configured. If filter is configured and relay isn't, with these changes we would still mount relay
> 
> https://github.com/waku-org/nwaku/blob/9691dfb5e2f2e12935045be05f65a6c737b973fe/waku/factory/node_factory.nim#L160-L168

Thanks for suggesting @gabrielmer !

waku/factory/node_factory.nim
```diff
@@ -144,9 +144,9 @@ proc setupProtocols(
...
     peerExchangeHandler = some(handlePeerExchange)
 
+  let shards =
+    conf.contentTopics.mapIt(node.wakuSharding.getShard(it).expect("Valid Shard"))
   if conf.relay:
-    let shards =
-      conf.contentTopics.mapIt(node.wakuSharding.getShard(it).expect("Valid Shard"))
     let pubsubTopics = conf.pubsubTopics & shards
 
     debug "Shards created from content topics",
```
gabrielmer, line 152: I think we should move this log to be right below

https://github.com/waku-org/nwaku/blob/11b575d3f94de25b38e7a82f5809b7efdeac7460/waku/factory/node_factory.nim#L147-L148

gabrielmer (approved):
Thanks so much!

Added a last detail to align :))

PR #2836, opened 2024-06-21
fix(rln-relay): clear nullifier log only if length is over max epoch gap

rymnc:
# Description

Fixes an edge case where the nullifier log was cleared when maxEpochGap == nullifierLog.len

# Changes

- [x] Adds unit tests for varied epoch sizes
- [x] fixes edge case

### DIFF
--- a/tests/waku_rln_relay/test_waku_rln_relay.nim
+++ b/tests/waku_rln_relay/test_waku_rln_relay.nim
@@ -886,3 +886,48 @@ suite "Waku rln relay":
     check:
       buckets.len == 5
       buckets == [2.0, 4.0, 6.0, 8.0, 10.0]
+  
+  asyncTest "nullifierLog clearing only after epoch has passed":
+    let index = MembershipIndex(0)
+
+    proc runTestForEpochSizeSec(rlnEpochSizeSec: uint) {.async.} =
+      let wakuRlnConfig = WakuRlnConfig(
+        rlnRelayDynamic: false,
+        rlnRelayCredIndex: some(index),
+        rlnRelayUserMessageLimit: 1,
+        rlnEpochSizeSec: rlnEpochSizeSec,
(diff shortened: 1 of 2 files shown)


rymnc:
> Thanks for this. Would it be possible to add a unit test to catch this edge case?

added in [ef242e7](https://github.com/waku-org/nwaku/pull/2836/commits/ef242e7814bb380ad3964fe4b6296864cc2b7982), previously we did not test for multiple epoch sizes, this test ensures it for 1,5,10,30, and 60.

tests/waku_rln_relay/test_waku_rln_relay.nim
```diff
@@ -886,3 +886,48 @@ suite "Waku rln relay":
...
+        let testEpoch =  wakuRlnRelay.calcEpoch(epochTime() + float(rlnEpochSizeSec * i))
+        wakuRlnRelay.nullifierLog[testEpoch] = testProofMetadataTable
+        check: wakuRlnRelay.nullifierLog.len().uint == i + 1
+
+      # clearing it now will be a no-op
+      wakuRlnRelay.clearNullifierLog()
+
+      check: wakuRlnRelay.nullifierLog.len().uint == rlnMaxEpochGap
+
+      # append one more now
```
alrevuelta, line 920: two things:

wondering if all this repeated code can be avoided by reusing the `for i in 0..<rlnMaxEpochGap:`. unsure if removing the `<` may help? afaik the intention here is to add rlnMaxEpochGap + 1 nullifier and see that `clearNullifierLog` doesn't clean them.

and for completeness, perhaps we should check that `clearNullifierLog` does clean the log when expected? meaning i keep adding and `len`()` keeps being `rlnMaxEpochGap`?

alrevuelta (approved):

gabrielmer (approved):
Thank you!

PR #2843, opened 2024-06-24
chore: add TWN parameters for RLNv2

alrevuelta:
* Add The Waku Network parameters for RLN v2.
* Set epoch size in seconds.
* Set amount of messages allowed per epoch.
* Set new contract, present in Polygon zkEVM testnet.


rymnc (commented):
Nice, LGTM, but now that the epoch size is `600`, can you add it in the array for the `nullifierLog` tests?

https://github.com/waku-org/nwaku/blob/c56785327ae457aa0aaf2a3822d22f7090e2fce3/tests/waku_rln_relay/test_waku_rln_relay.nim#L922

jm-clius (approved):
Thanks :)

rymnc (approved):
LGTM if checks pass

PR #2846, opened 2024-06-25
chore: archive.nim - increase the max limit of content topics per query to 100

Ivansete-status:
### Description
The _Status_ app tend to perform _Store_ queries multiple content topics. In this tiny PR we are increasing the max number of content topics in a query from 10 to 100

cc @chaitanyaprem

### DIFF
--- a/tests/waku_archive/test_waku_archive.nim
+++ b/tests/waku_archive/test_waku_archive.nim
@@ -233,13 +233,13 @@ procSuite "Waku Archive - find messages":
       response.messages.anyIt(it == msg1)
       response.messages.anyIt(it == msg3)
 
-  test "handle query with more than 10 content filters":
+  test "handle query with more than 100 content filters":
     ## Setup
     let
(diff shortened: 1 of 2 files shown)


chaitanyaprem:
cc @richard-ramos

NagyZoltanPeter (approved):
Legit! 
My only concern that such amount of content_topics will not work on REST interface because of URL max length... although it is not a major consideration.

gabrielmer (approved):
Thanks so much!

richard-ramos (approved):
I like this PR, but, I'm wondering, what was the reason behind using 10 content topics as a max?

chaitanyaprem (approved):
LGTM

Ivansete-status:
> I like this PR, but, I'm wondering, what was the reason behind using 10 content topics as a max?

Good question but I can't quite remember the original reasons ( cc @SionoiS )

PR #2855, opened 2024-06-27
fix(rln): nullifierlog vulnerability

alrevuelta:
* Fixes bug in `nullifierLog`. Now instead of counting elements in the table and removing when > `epochGap`, we ensure that they correspond to the current epoch +- `epochGap`. Anything than that we remove. This protects the node from messages arriving out-of-order.
* Fixes bug in `validateMessageAndUpdateLog`. The `nullifierLog` table was updated with new epochs no matter if the message was rate-limited or not, and no matter if the message even contained a valid proof. This allowed an attacker to attack the `nullifierLog` table, that in conjunction with the previous bug caused rate-limited messages to be accepted. The attacker at no cost could fill the table with unused epochs, bypassing the rate-limit protection.

### DIFF
--- a/waku/waku_rln_relay/rln_relay.nim
+++ b/waku/waku_rln_relay/rln_relay.nim
@@ -318,13 +318,17 @@ proc clearNullifierLog*(rlnPeer: WakuRlnRelay) =
   # if more than MaxEpochGap epochs are in the log
   let currentEpoch = fromEpoch(rlnPeer.getCurrentEpoch())
 
+  var epochsToRemove: seq[Epoch] = @[]
   for epoch in rlnPeer.nullifierLog.keys():
     let epochInt = fromEpoch(epoch)
 
     # clean all epochs that are +- rlnMaxEpochGap from the current epoch
-    if (currentEpoch+rlnPeer.rlnMaxEpochGap) < epochInt and epochInt < (currentEpoch-rlnPeer.rlnMaxEpochGap):
-      trace "clearing epochs from the nullifier log", cleanedEpoch = epochInt
-      rlnPeer.nullifierLog.del(epoch)
+    if (currentEpoch+rlnPeer.rlnMaxEpochGap) <= epochInt or epochInt <= (currentEpoch-rlnPeer.rlnMaxEpochGap):
+      epochsToRemove.add(epoch)
+  
+  for epochRemove in epochsToRemove:
+    trace "clearing epochs from the nullifier log", currentEpoch = currentEpoch, cleanedEpoch = epochInt
+    rlnPeer.nullifierLog.del(epochRemove)
 
 proc generateRlnValidator*(
     wakuRlnRelay: WakuRLNRelay, spamHandler = none(SpamHandler)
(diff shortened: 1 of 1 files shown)


waku/waku_rln_relay/rln_relay.nim
```diff
@@ -308,21 +308,41 @@ proc appendRLNProof*(
...
   let countToClear = rlnPeer.nullifierLog.len().uint - rlnPeer.rlnMaxEpochGap
   trace "clearing epochs from the nullifier log", count = countToClear
-  let epochsToClear = rlnPeer.nullifierLog.keys().toSeq()[0 ..< countToClear]
+
+  #let epochsToClear = rlnPeer.nullifierLog.keys().toSeq()[0 ..< countToClear]
+  #let epochsToClear = rlnPeer.nullifierLog.keys().toSeq().sort(compareKeys)[0 ..< countToClear]
+
+  var epochsToClear = rlnPeer.nullifierLog.keys().toSeq()
+  epochsToClear.sort(compareKeys)
+  epochsToClear = epochsToClear[0 ..< countToClear]
```
alrevuelta, line 340: as it is, it should work fine now.

however I'm having second thought about cleaning the nullifierlog this way. it has some implicit knowledge that can be error prone.

perhaps a better way could be. which epoch are we at? x? well, remove everything in the table that is beyond x-epochGap.

waku/waku_rln_relay/rln_relay.nim
```diff
@@ -358,6 +378,12 @@ proc generateRlnValidator*(
       payload = string.fromBytes(message.payload)
     case validationRes
     of Valid:
+      let proofMetadataRes = msgProof.extractMetadata()
+      if proofMetadataRes.isErr():
+        # dirty should never happen
+        return pubsub.ValidationResult.Reject
+      # insert the message to the log (never errors)
+      discard wakuRlnRelay.updateLog(msgProof.epoch, proofMetadataRes.get())
```
alrevuelta, line 386: the key here is that we shouldn't update the nullifierlog if the message is wrong (eg proof is invalid). otherwise anyone can "DoS" the nullifieerLog, forcing cleanups and making it useless.

waku/waku_rln_relay/rln_relay.nim
```diff
@@ -280,7 +280,7 @@ proc validateMessageAndUpdateLog*(
     return MessageValidationResult.Invalid
 
   # insert the message to the log (never errors)
-  discard rlnPeer.updateLog(msgProof.epoch, proofMetadataRes.get())
+  #discard rlnPeer.updateLog(msgProof.epoch, proofMetadataRes.get())
```
alrevuelta, line 283: this function is called even in invalid messages. see other comment.

alrevuelta: or maybe a better way use `isValidMessage` here and only `updateLog` if valid.

waku/waku_rln_relay/rln_relay.nim
rymnc: i like this idea

rymnc (approved):
LGTM, thanks!

PR #2856, opened 2024-06-28
doc: clarify max shard value on CLI args

fryorcraken:
I started this PR because it bugged me that I could not simply start `wakunode2` and connect to it via `wakucanary` because `wakucanary` wants shards:

```
▶ ./build/wakucanary -a=...
INF 2024-06-28 13:59:19.881+10:00 Cli flags                                  tid=2317631 file=wakucanary.nim:164 address=/ip4/0.0.0.0/tcp/60000/p2p/16Uiu2HAm6AKL8iczGNguEwwcJGmCmgAS2cpTRve4UkMExtwt45gA timeout=10s protocols=@[] logLevel=INFO
ERR 2024-06-28 13:59:19.883+10:00 Relay shards initialization failed         tid=2317631 file=wakucanary.nim:209 error="invalid shard count"
ERR 2024-06-28 13:59:19.883+10:00 The node has some problems (see logs)      tid=2317631 file=wakucanary.nim:283
```

In the _good default value_ spirit, we should be able to run the binary with minimal config out of the box, without having to fiddle things.

The good default at the moment should either be:
1, TWN (aka RLN Relay protected network)
2. Status prod

There is value in doing either. However, for the sake of dogfooding and pushing RLN further, I believe (1) should be default. Until we can get RLN on Status network.

So, the proposal I'd like to make before I proceed forward is to change all default parameters to TWN, meaning that a user will have to pass a API address for eth client to play with nwaku.

This would mean something similar to nwaku-compose, but without postgres.

Cc @waku-org/nwaku-developers 

cc @gabrielmer some suggestions in there in terms of handlng shards, in reference to #2163

### DIFF
--- a/waku/factory/node_factory.nim
+++ b/waku/factory/node_factory.nim
@@ -5,7 +5,8 @@ import
   libp2p/peerid,
   libp2p/protocols/pubsub/gossipsub,
   libp2p/nameresolving/dnsresolver,
-  libp2p/crypto/crypto
+  libp2p/crypto/crypto,
+  system/ansi_c
 
 import
   ./internal_config,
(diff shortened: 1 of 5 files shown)


waku/factory/node_factory.nim
```diff
@@ -144,14 +145,23 @@ proc setupProtocols(
...
-  let shards =
-    conf.contentTopics.mapIt(node.wakuSharding.getShard(it).expect("Valid Shard"))
-  debug "Shards created from content topics",
-    contentTopics = conf.contentTopics, shards = shards
+  var pubsubTopics: seq[string] = conf.pubsubTopics
+  for shard in conf.shards:
+    pubsubTopics.add($NsPubsubTopic.staticSharding(conf.clusterId, shard))
+
+  if conf.shards.len == 0:
+    # If shards are not configured but content topics are, then assume autosharding is expected
```
fryorcraken, line 153: I am guessing this is incorrect as you'd need to have shards configured (ie 0 to 7 ) to apply an auto sharding logic as the result of the autosharding logic depends on the shard range available.

waku/factory/node_factory.nim
```diff
@@ -144,14 +145,23 @@ proc setupProtocols(
 
     peerExchangeHandler = some(handlePeerExchange)
 
-  let shards =
-    conf.contentTopics.mapIt(node.wakuSharding.getShard(it).expect("Valid Shard"))
-  debug "Shards created from content topics",
-    contentTopics = conf.contentTopics, shards = shards
+  var pubsubTopics: seq[string] = conf.pubsubTopics
+  for shard in conf.shards:
```
fryorcraken, line 149: @gabrielmer I think this is needed in #2723

waku/factory/waku.nim
```diff
@@ -68,7 +68,10 @@ proc logConfig(conf: WakuNodeConf) =
 
   info "Configuration. Network", cluster = conf.clusterId, maxPeers = conf.maxRelayPeers
 
-  for shard in conf.pubsubTopics:
+  for pubsubTopic in conf.pubsubTopics:
+    info "Configuration. pubsub topics", pubsubTopic = pubsubTopic
```
fryorcraken, line 72: Something useful @gabrielmer

waku/node/waku_node.nim
```diff
@@ -423,7 +423,7 @@ proc mountRelay*(
 
   node.switch.mount(node.wakuRelay, protocolMatcher(WakuRelayCodec))
 
-  info "relay mounted successfully"
+  info "relay mounted successfully", pubsubTopics
```
fryorcraken, line 426: Something useful @gabrielmer

PR #2858, opened 2024-06-28
chore(rln-relay): add chain-id flag to wakunode and restrict usage if mismatches rpc provider

rymnc:
# Description

As requested by @alrevuelta to restrict chainId usage and waku node initialization should fail if user erroneously passes in an rpc url for another chain.

# Changes

- [x] All config changes from WakuNodeConf down to GroupManager
- [x] Tests

### DIFF
--- a/waku/factory/external_config.nim
+++ b/waku/factory/external_config.nim
@@ -76,6 +76,12 @@ type WakuNodeConf* = object
     name: "rln-relay-eth-contract-address"
   .}: string
 
+  rlnRelayChainId* {.
+    desc: "Chain ID of the provided contract (optional, will fetch from RPC provider if not used)",
+    defaultValue: 0,
+    name: "rln-relay-chain-id"
+  .}: uint
+
   rlnRelayCredPassword* {.
     desc: "Password for encrypting RLN credentials",
     defaultValue: "",
--- a/tests/waku_rln_relay/test_rln_group_manager_onchain.nim
+++ b/tests/waku_rln_relay/test_rln_group_manager_onchain.nim
@@ -26,6 +26,8 @@ import
   ../testlib/common,
   ./utils
 
+const CHAIN_ID = 1337
+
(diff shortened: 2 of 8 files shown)


waku/factory/external_config.nim
```diff
@@ -76,6 +76,12 @@ type WakuNodeConf* = object
     name: "rln-relay-eth-contract-address"
   .}: string
 
+  rlnRelayChainId* {.
+    desc: "Chain ID of the provided contract (optional, will fetch from RPC provider if not used)",
+    defaultValue: 0,
```
alrevuelta, line 81: So if not provided it will default to `0` but technically that can be a valid value?

alrevuelta (commented):
Not sure i follow. Shouldn't this fail? `rln-relay-chain-id` is wrong

```
docker run quay.io/wakuorg/nwaku-pr:2858 \
--cluster-id=1 \
--rln-relay-eth-client-address=https://rpc.cardona.zkevm-rpc.com \
--log-level=DEBUG \
--rln-relay-chain-id=111111
```

waku/factory/external_config.nim
rymnc: 0 is an invalid value for chainId. avoided using an `Optio n` here because of that

rymnc:
> Not sure i follow. Shouldn't this fail? `rln-relay-chain-id` is wrong
> 
> ```
> docker run quay.io/wakuorg/nwaku-pr:2858 \
> --cluster-id=1 \
> --rln-relay-eth-client-address=https://rpc.cardona.zkevm-rpc.com \
> --log-level=DEBUG \
> --rln-relay-chain-id=111111
> ```

since you pass in the cluster-id as 1, we override the rln-relay-chain-id, shouldn't that be the intended behaviour?

alrevuelta:
> since you pass in the cluster-id as 1,

indeed, my bad.

PR #2867, opened 2024-07-02
fix: handle rln-relay-message-limit 

### LINKED ISSUE #2822: bug: RLN_RELAY_MSG_LIMIT handling
### Problem
Setting `RLN_RELAY_MSG_LIMIT >= 21` triggers a node restart after a while. 
When running the `waku-simulator` compose with that value, containers restart after some 10s~. Right before that, messages like 
```
nwaku_5                   | ERR 2024-06-18 14:56:29.495+00:00 failure while registering credentials on-chain topics="rln_keystore_generator" tid=380 file=rln_keystore_generator.nim:72 error="index out of bounds, the container is empty"
```
 are printed. And after this the containers restart.

@stubbsta Verified that "the max value for user message limit is set to 20 in the contract and it's not configurable".

### To reproduce
Run a RLN+Relay-enabled nwaku node with the environment variable `RLN_RELAY_MSG_LIMIT` set to  `>= 21`.

### Expected behavior
The node handles that scenario gracefully, exiting early and printing an appropriate error message.

### nwaku version/commit hash
- wakunode2: `version / git commit hash 8e7e72`

darshankabariya:
close #2822


waku/waku_rln_relay/rln_relay.nim
```diff
@@ -435,19 +434,20 @@ proc mount(
   (await groupManager.startGroupSync()).isOkOr:
     return err("could not start the group sync: " & $error)
 
+  if (conf.rlnRelayUserMessageLimit > 20):
```
rymnc, line 437: it should be 100 now, as per parameters decided by the waku research team :)

darshankabariya: Thanks for the review, @rymnc. I will updates in the next commit.

waku/waku_rln_relay/rln_relay.nim
```diff
@@ -435,19 +434,20 @@ proc mount(
   (await groupManager.startGroupSync()).isOkOr:
     return err("could not start the group sync: " & $error)
 
+  if (conf.rlnRelayUserMessageLimit > 100):
+    return err("relay message limit (rln) cannot be exceed 100 ")
```
gabrielmer, line 438:
```suggestion
    return err("relay message limit (rln) cannot exceed 100")
```

tests/waku_rln_relay/test_wakunode_rln_relay.nim
```diff
@@ -572,3 +574,24 @@ procSuite "WakuNode - RLN relay":
...
+      rlnRelayCredIndex: some(0.uint),
+      rlnRelayUserMessageLimit: 111,
+      rlnEpochSizeSec: 0,
+      rlnRelayTreePath: genTempPath("rln_tree", "wakunode_9"),
+    )
+
+    await node.mountRlnRelay(wakuRlnConfig)
+
+    check:
+      $error == "relay message limit (rln) cannot be exceed 100 "
```
gabrielmer, line 597:
```suggestion
      $error == "relay message limit (rln) cannot exceed 100"
```

gabrielmer (approved):
Thanks so much! Added just a nitpick change

waku/waku_rln_relay/rln_relay.nim
alrevuelta: mm, why enforcing this limit? the user is free to choose whatever `rlnRelayUserMessageLimit` they want. TheWakuNetwork sets this to a max of 100, but it shouldn't be enforced in the code.

for example, someone may want to deploy their own cluster, with their own contract with rlnRelayUserMessageLimit=9999.

darshankabariya: Hi @AlejandroCabeza, Alvaro makes a good point about why we are enforcing a limit for messages.

rymnc: > mm, why enforcing this limit? the user is free to choose whatever `rlnRelayUserMessageLimit` they want. TheWakuNetwork sets this to a max of 100, but it shouldn't be enforced in the code.
> 
> for example, someone may want to deploy their own cluster, with their own contract with rlnRelayUserMessageLimit=9999.

agree - this limit should probably be enforced in another place -
`OnchainGroupManager`: fetch the `MAX_MESSAGE_LIMIT` from the contract and compare with the user provided `rlnRelayUserMessageLimit`

darshankabariya: Great ! then let's automate that.

darshankabariya: Hi @rymnc, I tried fetching MAX_MESSAGE_LIMIT from the contract based on your suggestion, but it failed in many cases. Upon debugging, I discovered that it returns 0 because it hasn't been initialized yet. Now, I've found another approach: there's an attribute called userMessageLimit in base_groupmanager, which I believe is a better option. What do you think ?

Earlier, as you mentioned that the message limit is set 20 for static. Can we fetch the on-chain limit from the contract?

rymnc: > Can we fetch the on-chain limit from the contract?

that is possible by calling `MAX_MESSAGE_LIMIT()`

rymnc: > I discovered that it returns 0

please share a code snippet?

darshankabariya: I've just updated the PR. plz take a look at.

cc @rymnc

darshankabariya: The associated test case failed with this implementation because the MAX_MESSAGE_LIMIT() API is returning 0. This is likely due to it not being initialized, which is why it returns zero.

tests/waku_rln_relay/test_wakunode_rln_relay.nim
```diff
@@ -595,4 +595,4 @@ procSuite "WakuNode - RLN relay":
       await node.mountRlnRelay(wakuRlnConfig)
     except CatchableError as e:
       check e.msg ==
-        "failed to mount WakuRlnRelay: relay message limit (rln) cannot be exceed 100"
+        "failed to mount WakuRlnRelay: relay-message-limit can't be exceed then user-message-limit set by contract"
```
rymnc, line 598: this file uses the static mode of operation, so there is no point checking the user message limit from the contract

rymnc: This test should actually go in the `tests/waku_rln_relay/test_rln_group_manager_onchain.nim` file

tests/waku_rln_relay/test_wakunode_rln_relay.nim
```diff
@@ -572,3 +574,25 @@ procSuite "WakuNode - RLN relay":
...
+  asyncTest "test rln-relay-message-limit":
+    let
+      nodeKey = generateSecp256k1Key()
+      node = newTestWakuNode(nodeKey, parseIpAddress("0.0.0.0"), Port(0))
+
+    await node.mountRelay(@[DefaultPubsubTopic])
+
+    # mount rlnrelay in off-chain mode
+    let wakuRlnConfig = WakuRlnConfig(
+      rlnRelayDynamic: false,
```
rymnc, line 587: notice that it is not dynamic, aka not onchain, hence we cannot get the user message limit from the contract. I would suggest to use a different mechanism for the static group manager

tests/waku_rln_relay/test_rln_group_manager_static.nim
```diff
@@ -220,3 +224,25 @@ suite "Static group manager":
...
+      rlnRelayUserMessageLimit: 111,
+      rlnEpochSizeSec: 0,
+      rlnRelayTreePath: genTempPath("rln_tree", "wakunode_9"),
+    )
+
+    try:
+      await node.mountRlnRelay(wakuRlnConfig)
+    except CatchableError as e:
+      check e.msg ==
+        "failed to mount WakuRlnRelay: rln-relay-message-limit can't be exceed then MAX_MESSAGE_LIMIT set by contract"
```
rymnc, line 248: this wont work, because you're testing the static group manager here (there is no contract, and therefore, no enforceable max msg limit)

waku/waku_rln_relay/rln_relay.nim
```diff
@@ -444,8 +445,10 @@ proc mount(
   (await groupManager.startGroupSync()).isOkOr:
     return err("could not start the group sync: " & $error)
 
-  if ($conf.rlnRelayUserMessageLimit > $groupManager.rlnRelayMaxMessageLimit):
-    return err("relay-message-limit can't be exceed then user-message-limit set by contract")
+  if (conf.rlnRelayUserMessageLimit > groupManager.rlnRelayMaxMessageLimit):
+    return err(
+      "rln-relay-message-limit can't be exceed then MAX_MESSAGE_LIMIT set by contract"
+    )
```
rymnc, line 451: this can be removed

tests/waku_rln_relay/test_rln_group_manager_onchain.nim
```diff
@@ -775,7 +779,27 @@ suite "Onchain group manager":
       isReady == true
 
     await manager.stop()
+  
+  asyncTest "rln-relay-max-message-limit testing":
```
rymnc, line 783: i am not sure if this test belongs here since we are just testing the behaviour of the onchain group manager in this module
can you add a todo comment mentioning it should be moved later?

rymnc (approved):
LGTM after addressing my comment

PR #2874, opened 2024-07-03
chore: Update master from release v30

Ivansete-status:
### Description
This is needed to update master so that we keep the same decisions that we made in release/v30 to keep using Sepolia and also, use new node keys

We will keep using Sepolia testnet because the zkEVM didn't work properly, i.e., the nonce attribute wasn't consistent during different transactions.


NagyZoltanPeter (approved):
...trust you with these consts :-)

rymnc (approved):

alrevuelta (approved):

gabrielmer (approved):
Thank you!

PR #2878, opened 2024-07-04
chore: setting connectivity loop interval to 30 seconds

gabrielmer:
# Description
@alrevuelta noticed that peers get added to the a node very slowly.

We used to have `ConnectivityLoopInterval` set to 15 seconds, and changed it around half a year ago to 1 minute. At that time, the change also caused issues to QA because of slow times getting peers.

The idea is to eventually set the wait times between two connectivity loops depending on the amount of peers a node has. See https://github.com/waku-org/nwaku/issues/2877

In the meantime, I propose changing it to 30 seconds as some kind of middle ground between the 15 seconds and 1 minute alternatives: twice as long as 15 seconds but half as much as 1 minute :)

LMK if it makes sense or not, it should help us fill our peers faster in the meantime

# Changes

- [x] changed `ConnectivityLoopInterval` to 30 seconds


alrevuelta (approved):
lgtm by now. at some point we may want to have a shorter loop when peer count is very low, and higher when close to peer limit.

SionoiS (approved):
Should be fine to use 15s the only problem that could happen is if sharded peer management is enabled and dials don't finish before the loop start again (i'm guessing at worst some dials would fail because of the hard limit).

Ivansete-status (approved):
LGTM! Thanks! :100:

PR #2881, opened 2024-07-04
fix(rln_keystore_generator): improve error handling for unrecoverable failure

rymnc:
# Description

The `rln_keystore_generator` previously did not pass in an `onFatalErrorHandler`, which was leading to segfaults since we don't have a `not nil` check on the callback. Now, this PR introduces the `not nil` check, and also provides an `onFatalError` callback to the `OnchainGroupManager` when it is constructed.

# Changes

- [x] `rln_keystore_generator` now handles fatal errors appropriately
- [x] `retryWrapper` checks if the `errCallback` is nil, and handles the invariant appropriately


gabrielmer (approved):
LGTM, thanks!

Ivansete-status (approved):

PR #2885, opened 2024-07-06
chore: Bump dependencies for v0.31.0

Ivansete-status:
### Description

- Bumping dependencies for `v0.31.0`.
- Bumping the Nim compiler version, from `1.6.0` to `2.0.8` ( caused by the bump of `nimbus-build-system` dependency .)
- :bell: the `--mm:refc` is needed to be passed explicitly to `nim` when compiling as 2.0 enables a new garbage collector that we're not yet ready to support.
- Removing some warnings ( stew/results -> results .)
- Addition of `db_connector` vendor dependency, which contains _postgres_ types ( and other databases .) This is needed because `nimbus-build-system` no longer includes the `db_postgres` module.

( cc: so that you are aware we are bumping the Nim compiler version :) @fryorcraken, @jm-clius, @alrevuelta, @rymnc, @richard-ramos, @chaitanyaprem , @arnetheduck, @kaiserd, @AlejandroCabeza .)

### Issue
closes https://github.com/waku-org/nwaku/issues/2876


gabrielmer (approved):
LGTM, thanks so much!

NagyZoltanPeter (approved):
Excited about nim version bump!!!
Please notice this PR https://github.com/waku-org/nwaku/pull/2882 has overlap but above that unfortunately does not solve the issue with compile secp256k1 lib on latest MAC. Will tackle that issue further as soon as I have more information.

waku/common/databases/db_sqlite.nim
```diff
@@ -231,7 +233,7 @@ proc query*(db: SqliteDatabase, query: string, onData: DataProc): DatabaseResult
       else:
         return err($sqlite3_errstr(v))
     return ok gotResults
-  finally:
+  except Exception, CatchableError:
```
SionoiS, line 236: Is  `finally`and the code below not still required if no exception raised?

SionoiS (approved):
LGTM

PR #2887, opened 2024-07-07
fix: postgres_driver - better partition creation without exclusive access

Ivansete-status:
### Description

This PR is inspired by the following article: https://supabase.com/blog/postgres-dynamic-table-partitioning

With that, we aim to avoid having database `ACCESS EXCLUSIVE` operations, that use to happen when a new partition is created.

We are currently suffering from quite frequent and severe blocks in the `postgres` database, which in turn, blocks the connected `nwaku` nodes. This currently impacts `shards.test` fleet and _Status_ app.

### Issue

closes https://github.com/waku-org/nwaku/issues/2783


SionoiS (approved):
LGTM

Adding the constraint before attaching the table is a very good idea!

NagyZoltanPeter (approved):
Hm, very interesting finding! Await to see in action!!!

gabrielmer (approved):
Thank you!

PR #2889, opened 2024-07-09
fix: peer-exchange issue

### LINKED ISSUE #2414: bug: peer exchange returns nodes that no longer exist.
@chaitanyaprem while on a debuging session with the status team, discovered that sometimes while attempting to use lightpush to publish a message, it would fail even after 5 retries. On further debugging noticed that each time a random peer was selected for lightpush and the sender was unable to dial to it. All these peers are returned via peer-exchange from the boot nodes.

The most probably cause as considering the connection failed in each scenario with a dial timeout could be that these nodes are no longer even available, otherwise the TCP connection would have been established and p2p level connection would have failed.
```
12220 ERROR[02-09|07:17:44.125|github.com/status-im/status-go/vendor/github.com/waku-org/go-waku/waku/v2/protocol/lightpush/waku_lightpush.go:309] could not publish message                      hash=0x2fe09b470bc8f13be52abd09930a5f3eb6ae274ae4678a985dd679edc6deb7c1 pubsubTopic=/waku/2/rs/16/32 contentTopic=/waku/1/0x4cdd76ba/rfc26 timestamp=1      ,707,463,059,122,842,165 peerID=16Uiu2HAmN2n45zWeNhXtQis1kp7Wrun4pz1NFhDi9uBZQaarmBLq error="failed to dial: failed to dial 16Uiu2HAmN2n45zWeNhXtQis1kp7Wrun4pz1NFhDi9      uBZQaarmBLq:\n  * [/ip4/95.217.224.92/tcp/44647] dial tcp4 0.0.0.0:39663->95.217.224.92:44647: i/o timeout"                                  
```

Further discussion on Discord with @alrevuelta revealed that in nwaku every 15 minutes it caches peers that are reachable (connected succesfully in the past), and serves the to peer exchange requests https://github.com/waku-org/nwaku/blob/master/waku/waku_peer_exchange/protocol.nim#L118-L131 , however, to select reachable peers, the criteria it uses to choose the peers is those whose connectedness is either `Connected` or `CanConnect`, which in https://github.com/waku-org/nwaku/blob/master/waku/node/peer_manager/peer_manager.nim#L395-L397 could mean that a peer that was connected will be considered forever that it was CanConnect thus be able to be potentially selected for the peer exchange response even if the node has not been seen anymore.

A potential solution suggested by @alrevuelta is to cache `Connected` peers and that's it. But this has some privacy concerns (graph learning attacks), or maybe add a `lastSuccesfullConn` (similar to the existing `lastFailedConn`) and then filter based on that, (i.e. only allow few hours of discrepancy)

darshankabariya:
closes #2414 

This implementation will enhance the reliability of the peers shared by the peer-exchange protocol with its clients. because now the cache updates every 5 min.

After the resolution of the peer_manager issue #2905, the reliability of peer exchanges has also been improved.

### DIFF
--- a/waku/waku_peer_exchange/protocol.nim
+++ b/waku/waku_peer_exchange/protocol.nim
@@ -32,7 +32,7 @@ const
   DefaultMaxRpcSize* = 10 * DefaultMaxWakuMessageSize + 64 * 1024
     # TODO what is the expected size of a PX message? As currently specified, it can contain an arbitary number of ENRs...
   MaxPeersCacheSize = 60
-  CacheRefreshInterval = 10.minutes
+  CacheRefreshInterval = 5.minutes
 
   WakuPeerExchangeCodec* = "/vac/waku/peer-exchange/2.0.0-alpha1"
 
--- a/waku/waku_peer_exchange/protocol.nim
+++ b/waku/waku_peer_exchange/protocol.nim
@@ -32,7 +32,7 @@ const
   DefaultMaxRpcSize* = 10 * DefaultMaxWakuMessageSize + 64 * 1024
     # TODO what is the expected size of a PX message? As currently specified, it can contain an arbitary number of ENRs...
   MaxPeersCacheSize = 60
-  CacheRefreshInterval = 5.minutes
+  CacheRefreshInterval = 10.minutes
 
   WakuPeerExchangeCodec* = "/vac/waku/peer-exchange/2.0.0-alpha1"
 
--- a/waku/waku_peer_exchange/protocol.nim
+++ b/waku/waku_peer_exchange/protocol.nim
@@ -32,7 +32,7 @@ const
   DefaultMaxRpcSize* = 10 * DefaultMaxWakuMessageSize + 64 * 1024
     # TODO what is the expected size of a PX message? As currently specified, it can contain an arbitary number of ENRs...
   MaxPeersCacheSize = 60
-  CacheRefreshInterval = 15.minutes
+  CacheRefreshInterval = 5.minutes
 
   WakuPeerExchangeCodec* = "/vac/waku/peer-exchange/2.0.0-alpha1"
 


waku/node/peer_manager/waku_peer_store.nim
```diff
@@ -140,8 +145,10 @@ proc getPeersByProtocol*(peerStore: PeerStore, proto: string): seq[RemotePeerInf
   return peerStore.peers.filterIt(it.protocols.contains(proto))
 
 proc getReachablePeers*(peerStore: PeerStore): seq[RemotePeerInfo] =
+  let threshold = Moment.fromNow(millis(-3*60*60*1000))
```
Ivansete-status, line 148: Are we sure the units are in seconds?

Ivansete-status (approved):
LGTM! Thanks! :100: 
I just added a naïve question :)

waku/node/peer_manager/waku_peer_store.nim
darshankabariya: Actually, the secs and hours functions cause compilation errors. will try to move hourly formula. regarding 3 hours I will check with alvaro before merging.

chaitanyaprem (commented):
is it just me or something got messed up during force push?

I literally don't see any diff at all. Here is what i am seeing

darshankabariya:
> is it just me or something got messed up during force push?
> 
> I literally don't see any diff at all. Here is what i am seeing
> 
> Screenshot 2024-07-12 at 3 58 29 PM

Currently, there's nothing in this PR. I implemented Alvaro's suggested proposal, but Hanno recommended a more feasible long-term solution. so i revert older version. I'm working on that now and will update once it's done. Actually discussion going on this thread, https://discord.com/channels/1110799176264056863/1260493155791405096

jm-clius (approved):
This is a good start. Perhaps we can evaluate if things improve? Afaics this is a very cheap refresh, so we may even lower interval further if needed.

darshankabariya:
> This is a good start. Perhaps we can evaluate if things improve? Afaics this is a very cheap refresh, so we may even lower interval further if needed.

Thank you for your suggestion. I've already created a draft PR to analyze peer exchange behavior: [#2940](https://github.com/waku-org/nwaku/pull/2940).

darshankabariya:
> This is a good start. Perhaps we can evaluate if things improve. Afaics this is a very cheap refresh, so we may even lower interval further if needed.

As you suggest @jm-clius 
I've been developing tools to analyze peer exchange availability for the Waku network. You can find the code in PR #2940

**Tool Overview**
This tool evaluates the effectiveness of peer exchange responses. Typically, the peer list updates every 15 minutes. In my analysis, I requested 5 peers every 2 minutes, checking their availability by attempting to connect at the P2P level.

The process involved 60 iterations, accumulating 300 peer requests over 3 hours. I generated percentage charts to visualize success rates at different cache refresh intervals. The results are as follows:

15-minute interval: 265/300
10-minute interval: 278/300
5-minute interval: 277/300

**Conclusion**
Increasing the frequency by decreasing the cache refresh interval appears effective only for 10 minutes. At 5 minutes, similar to 10 minutes. I recommend updating the cache refresh rate from 15 to 10 minutes.

I’d welcome any suggestions or alternative approaches to this analysis 

cc @gabrielmer @chaitanyaprem @Ivansete-status

Ivansete-status:
Thanks for the analysis in https://github.com/waku-org/nwaku/pull/2889#issuecomment-2307475536 @darshankabariya . It seems reasonable to set it to 10min (considering very few test iterations but it is fine.)
Make sure to enhance the CI before merging :)

PR #2893, opened 2024-07-10
feat: sonda tool

### LINKED ISSUE #2869: chore: create sonda tool
### Background
We are facing issues where the store nodes don't respond on time, or do not respond at all.

With that, we need to have a `sonda` node continuously publishing small messages to the nodes. The publish, and subsequent store request should be performed every minute. For example, we will have the `sonda` node that sends a message `Hi, I'm sonda`, with a random contentTopic, and in turn, the same `sonda` node will perform a small store query to retrieve that last published message. In case of an error, we should properly raise an alarm, or metric, to notify this situation, and it should be easy to pinpoint when the issue started to happen and when, if that's the case, the problem got sorted out.

A strong requisite is that the `sonda` node should be running 24/7, always.

Features that should have the `sonda` tool:
- The code base should go under `apps` folder.
- Should be able to receive a list of store nodes as a parameter.
- Should have a configurable request frequency. We can start with 1 req / minute.
- On every request, the tool will publish a message with a fixed message and random contentTopic.
- After publishing a message, the tool should wait some seconds, and then perform a store query.
- The tool should keep different metrics per each store node
  - The time consumed by the store query
  - A time-series metric indicating when a store node is considered "up" and when it is considered "down".

### Acceptance criteria
We should be able to accurately monitor the availability of any store node hosted by us.

gabrielmer:
# Description
Created a new tool named `Sonda` which regularly polls store nodes and displays a Grafana dashboard with the results.

image

For usage instructions, please refer to the README :)

## Issue

closes #2869


apps/sonda/run_node.sh
```diff
@@ -0,0 +1,111 @@
...
+  --rest=true\
+  --rest-admin=true\
+  --rest-address=0.0.0.0\
+  --rest-port=8645\
+  --rest-allow-origin="waku-org.github.io"\
+  --rest-allow-origin="localhost:*"\
+  --nat=extip:"${MY_EXT_IP}"\
+  --store=false\
+  --pubsub-topic="/waku/2/rs/${CLUSTER_ID}/${SHARD}"\
+  --discv5-bootstrap-node="enr:-QEKuECA0zhRJej2eaOoOPddNcYr7-5NdRwuoLCe2EE4wfEYkAZhFotg6Kkr8K15pMAGyUyt0smHkZCjLeld0BUzogNtAYJpZIJ2NIJpcISnYxMvim11bHRpYWRkcnO4WgAqNiVib290LTAxLmRvLWFtczMuc2hhcmRzLnRlc3Quc3RhdHVzLmltBnZfACw2JWJvb3QtMDEuZG8tYW1zMy5zaGFyZHMudGVzdC5zdGF0dXMuaW0GAbveA4Jyc40AEAUAAQAgAEAAgAEAiXNlY3AyNTZrMaEC3rRtFQSgc24uWewzXaxTY8hDAHB8sgnxr9k8Rjb5GeSDdGNwgnZfg3VkcIIjKIV3YWt1Mg0"\
```
gabrielmer, line 101: I added `shards.test` bootstrap nodes as default because if we use `cluster-id` 1, they get overridden by https://github.com/waku-org/nwaku/blob/e1518cf9ffc6c9ce2230aa5a44c52784bbd9f913/waku/factory/networks_config.nim#L37-L42

And most uses for us will be either `cluster-id` 1 or 16, and this setup works for both.

Happy to hear better alternatives :)

apps/sonda/.env.example
```diff
@@ -0,0 +1,33 @@
...
+# -------------------- SONDA CONFIG ------------------
+CLUSTER_ID=16
+SHARD=32
+# Comma separated list of store nodes to poll
+STORE_NODES="/dns4/store-01.do-ams3.shards.test.status.im/tcp/30303/p2p/16Uiu2HAmAUdrQ3uwzuE4Gy4D56hX6uLKEeerJAnhKEHZ3DxF1EfT,\
+/dns4/store-02.do-ams3.shards.test.status.im/tcp/30303/p2p/16Uiu2HAm9aDJPkhGxc2SFcEACTFdZ91Q5TJjp76qZEhq9iF59x7R,\
+/dns4/store-01.gc-us-central1-a.shards.test.status.im/tcp/30303/p2p/16Uiu2HAmMELCo218hncCtTvC2Dwbej3rbyHQcR8erXNnKGei7WPZ,\
+/dns4/store-02.gc-us-central1-a.shards.test.status.im/tcp/30303/p2p/16Uiu2HAmJnVR7ZzFaYvciPVafUXuYGLHPzSUigqAmeNw9nJUVGeM,\
+/dns4/store-01.ac-cn-hongkong-c.shards.test.status.im/tcp/30303/p2p/16Uiu2HAm2M7xs7cLPc3jamawkEqbr7cUJX11uvY7LxQ6WFUdUKUT,\
+/dns4/store-02.ac-cn-hongkong-c.shards.test.status.im/tcp/30303/p2p/16Uiu2HAm9CQhsuwPR54q27kNj9iaQVfyRzTGKrhFmr94oD8ujU6P"
```
gabrielmer, line 29: Should I change the .env example to use `cluster-id` 1 and some other store nodes? If so, which? didn't see in https://fleets.status.im/ store nodes for TWN

Ivansete-status: I think is fine that way :partying_face:

Ivansete-status (approved):
LGTM! Thanks for it! :100:

NagyZoltanPeter (approved):
Amazed!!! Thank you!!

PR #2894, opened 2024-07-10
chore: simplify migration scripts and remove store legacy code

Ivansete-status:
### Description
- Simplify the migration scripts thanks to already having the `timestamp` data within the `storedAt` column.
- Remove all the store legacy code as we are switching to store-v3 only


SionoiS (approved):
1000% Simpler but is it no longer a requirement to be backward compatible?

As for the migration, old messages `timestamp` would be all different after the migration would that not affect Status?

Ivansete-status:
> 1000% Simpler but is it no longer a requirement to be backward compatible?
> 
> As for the migration, old messages `timestamp` would be all different after the migration would that not affect Status?

I think we are aiming for not backward compatible changes and the same happens in the _Status_ changes. @richard-ramos - correct me if I'm wrong

gabrielmer (commented):
Ooh super happy to finally remove legacy store! We should also make sure tu remove it from the API documentation in https://github.com/waku-org/waku-rest-api

Regarding the migration script, there's something I don't quite understand.
Why are we dropping `timestamp` and renaming `storedAt` to `timestamp` instead of just dropping `storedAt`? are the current values in `timestamp` wrong?

gabrielmer (commented):
Ooh super happy to finally remove legacy store! We should also make sure tu remove it from the API documentation in https://github.com/waku-org/waku-rest-api

Regarding the migration script, there's something I don't quite understand.
Why are we dropping `timestamp` and renaming `storedAt` to `timestamp` instead of just dropping `storedAt`? are the current values in `timestamp` wrong?

Ivansete-status:
> Ooh super happy to finally remove legacy store! We should also make sure tu remove it from the API documentation in https://github.com/waku-org/waku-rest-api

Thanks for the reminder!

> Regarding the migration script, there's something I don't quite understand. Why are we dropping `timestamp` and renaming `storedAt` to `timestamp` instead of just dropping `storedAt`? are the current values in `timestamp` wrong?

Both `timestamp` and `storedAt` fields keep exactly the same values. We can't remove the `storedAt` attribute first because it is part of the PK and hence we apply that movement.

gabrielmer (approved):
Amazing! Thanks so much! 😍

richard-ramos:
Even tho status-go will then use v3 once PR https://github.com/status-im/status-go/pull/5123 is merged, existing users that have not updated their status-app will continue to use storeV2 (since that's what their app's status-go version will use), so I'm not sure we should remove the legacy code yet, but keep it instead for a couple of months.

Ivansete-status:
Thanks for the comment @richard-ramos !
Therefore, we need to keep the support to store-v2 until _Stauts_ users have switched to store-v3 ( we consider at least one 6 weeks ( one _Status_ release period ) of having the option to use _store-v3_ .)

NagyZoltanPeter (approved):
LGTM, yeah would have been to easy step to remove store v2 for now. But will come!

Ivansete-status:
> 1000% Simpler but is it no longer a requirement to be backward compatible?

Thanks again for the comments @SionoiS !
May take another look plz guys?

In the current changes we are simultaneously supporting store-v2 and store-v3

( cc @NagyZoltanPeter @gabrielmer )

waku/factory/node_factory.nim
```diff
@@ -223,53 +223,54 @@ proc setupProtocols(
...
-      return err("failed to mount waku legacy archive protocol: " & mountArcRes.error)
+      return err("failed to mount waku archive protocol: " & mountArcRes.error)
 
-    # Store setup
+    # Store legacy setup
     let rateLimitSetting: RateLimitSetting =
       (conf.requestRateLimit, chronos.seconds(conf.requestRatePeriod))
-
     try:
       await mountLegacyStore(node, rateLimitSetting)
```
gabrielmer, line 268: Here we are calling `mountLegacyStore(node, rateLimitSetting)` even in the cases of `conf.legacyStore` being false. Is this expected?

gabrielmer (commented):
Amazingg PR! Really ingenious workaround to support simultaneously both store versions and keep the code nice and clean :)

Added just a small question and a super nitpick comment

waku/factory/node_factory.nim
Ivansete-status: Thanks for that! It is corrected in https://github.com/waku-org/nwaku/pull/2894/commits/5b9871bed60420656215d3a15f28c3c7c37dcc1a

SionoiS (approved):

gabrielmer (approved):

PR #2897, opened 2024-07-11
feat: DOS protection of non relay protocols - rate limit phase3

NagyZoltanPeter:
# Description
This PR is the final phase for Rate Limiting non relay protocols.

### Aimed three major target:

- Protect filter service on its specific needs. 
   - Allow only a reasonable amount of subscribe/unsubscribe and ping requests for each client peer.
   - Such peer measurement is loosely coupled with maintained subscribed peers.
- Add fair usage of lightpush and store protocols per peers. Do not allow one/few peers to overload the service.
- Extend TokenBucket with load balancing.
   -  Replenish calculation can consider previous period load and can apply more tokens thus compensating lower resource usage of previous period (capped with 25% threshold)

> ***Remark*** in this version while lightpush and store protocols' protection can be configured from CLI by default it is switched off. 

### Notice for reviewers:
- Filter protocol has an arbitrary limit applied per peers (30 req / 1 min / peer)

- RequestRateLimiter applies a ratio calculation for managing fair usage of peers.
  - Thus, while main token bucket applies a global token capacity within a certain period of time. Peer allowed usage is managed as a factor of this. 
> Calculated as ***"main token capacity" * ratio * 75% over (main time period) * ratio***
> In other words when a peer used 75% of the previous n periods token capacity it gets rejected in order to let small users go up until main token capacity is not reached.
> Of course, big usage peers will be replenished also but only after the "n" period of time.

The ratio is calculated currently as - if main time period is set to:
- less than a second -> 10x
- less than a minute -> 3x 
- anything above that -> 2x ratio applied

***This is debatable*** I wanted to apply reasonable scale in time on the applied multiplication factor.

 Peer measurement also time capped, non requester peers will be wiped out from the store.

# Changes

- Enhanced TokenBucket to be able to add compensation tokens based on previous usage percentage,
- per peer rate limiter 'PeerRateLimier' applied on waku_filter_v2 with opinionated default of acceptable request rate
- Add traffic metrics to filter message push
- RequestRateLimiter added to combine simple token bucket limiting of request numbers but consider per peer usage over time and prevent some peers to over use the service (although currently rule violating peers will not be disconnected by this time only their requests will get not served)
- TimedMap utility created (inspired and taken from libp2p TimedCache) which serves as forgiving feature for peers had been overusing the service.
- Applied new RequestRateLimiter to store and lightpush protocols
- Added more tests
- Fix rebase issues

## How to test

Unit tests applied.

# Issue

https://github.com/waku-org/nwaku/issues/2683 covers https://github.com/waku-org/nwaku/issues/2032


SionoiS (approved):
Amazing work!

As for what settings to use, it's hard to say. Real usage would be how we determine that.

waku/common/utils/timedmap.nim
```diff
@@ -0,0 +1,150 @@
+# Nim-LibP2P
+# Copyright (c) 2023 Status Research & Development GmbH
+# Licensed under either of
+#  * Apache License, version 2.0, ([LICENSE-APACHE](LICENSE-APACHE))
+#  * MIT license ([LICENSE-MIT](LICENSE-MIT))
+# at your option.
+# This file may not be copied, modified, or distributed except according to
+# those terms.
+
```
jm-clius, line 9: For a new utility type it may be a good idea to preface the module with a brief description as to what is defined and what its purpose is (see for example https://github.com/status-im/nim-stew/blob/master/stew/keyed_queue.nim#L11-L23)

waku/common/utils/timedmap.nim
jm-clius, line 1: Generally we try and avoid adding more things to the `utils` folder (see: https://github.com/waku-org/nwaku/blob/master/waku/utils/DEPRECATION_NOTICE.md). Utility modules should either be in a package matching their function (e.g. `collections`) or be moved closed to the module where it's used. I know this `TimedMap` will be generally useful, but perhaps for now we can add it as part of the `ratelimit` directory as it's the only place it's used?

waku/common/ratelimit/peerratelimiter.nim
jm-clius, line 1: For the sake of clarity, I suggest using `_underscore_` to separate different words in module and directory names. I would also use common prefixes for similarly functioned modules. Since `ratelimit` is already the parent directory, it does not have to be repeated in module names.

My suggestion here would be:

```plaintext
rate_limit/
├── limiter_peer.nim
├── limiter_request.nim
├── limiter_simple.nim
├── metrics.nim
├── setting.nim
├── timed_map.nim
└── token_bucket.nim
```

I suggest elsewhere moving `timed_map` here (for now).

jm-clius (approved):
Really great work! 🎉

Only very minor comments below.

gabrielmer (approved):
Looks amazing! 🔥 🔥 
Thanks so much!

waku/waku_filter_v2/subscriptions.nim
```diff
@@ -12,6 +12,12 @@ const
   DefaultSubscriptionTimeToLiveSec* = 5.minutes
   MessageCacheTTL* = 2.minutes
 
+  # Acceptable call frequence from one peer using filter service
+  # Assumption is having to set up a subscription with max 30 calls than using ping in every min
+  # While subscribe/unsubscribe events are distributed in time among clients, pings will happen regularly from
+  # all subscribed peers
+  FilterPerPeerRateLimit*: RateLimitSetting = (30, 1.minutes)
```
Ivansete-status, line 19: I'm just curious about where that limit came from :)

waku/waku_filter_v2/protocol.nim
```diff
@@ -267,21 +272,32 @@ proc initProtocolHandler(wf: WakuFilter) =
...
+        waku_filter_errors.inc(labelValues = [decodeRpcFailure])
+        return
 
-    let response = wf.handleSubscribeRequest(conn.peerId, request)
+      let request = decodeRes.value #TODO: toAPI() split here
 
-    debug "sending filter subscribe response",
-      peer_id = shortLog(conn.peerId), response = response
+      response = wf.handleSubscribeRequest(conn.peerId, request)
+
```
Ivansete-status, line 290: I wonder if makes sense to also add something like the following:
```code
      waku_service_network_bytes.inc(
        amount = reqBuf.len().int64, labelValues = [WakuFilterSubscribeCodec, "in"]
      )
```

waku/common/ratelimit/requestratelimiter.nim
```diff
@@ -0,0 +1,108 @@
...
+    bodyWithinLimit
+  else:
+    waku_service_requests.inc(labelValues = [proto, "rejected"])
+    bodyRejected
+
+# TODO: review these ratio assumptions! Debatable!
+func calcPeriodRatio(settingOpt: Option[RateLimitSetting]): int =
+  settingOpt.withValue(setting):
+    if setting.isUnlimited():
+      return 0
```
Ivansete-status, line 67: Instead of using hard-coded values I'd try to give them a more meaningful `const` value.
e.g.,
```suggestion
      return RATE_UNLIMITED
```

waku/common/ratelimit/requestratelimiter.nim
```diff
@@ -0,0 +1,108 @@
...
+  do:
+    return 0.seconds
+
+func calcPeerTokenSetting(
+    setting: Option[RateLimitSetting], ratio: int
+): RateLimitSetting =
+  let s = setting.valueOr:
+    return (0, 0.minutes)
+
+  let peerVolume = trunc((s.volume * ratio).float * 0.75).int
```
Ivansete-status, line 96: Similar to what I suggested, wrt `0.75`, I think it is better to use a `const` value

waku/common/ratelimit/simpleratelimiter.nim
```diff
@@ -0,0 +1,48 @@
...
+import std/[options], chronos/timer, libp2p/stream/connection, libp2p/utility
+
+import ./[tokenbucket, ratelimitsetting, waku_service_metrics]
+export tokenbucket, ratelimitsetting, waku_service_metrics
+
+proc newTokenBucket*(
+    setting: Option[RateLimitSetting],
+    replenishMode: ReplenishMode = ReplenishMode.Compensating,
+): Option[TokenBucket] =
+  if setting.isNone:
```
Ivansete-status, line 12: Super nitpick. When using "verbs" we tend to use parenthesis :)
```suggestion
  if setting.isNone():
```

Ivansete-status (approved):
LGTM! Thanks for it! :100: 
I just added some comments that I hope you find useful :)

PR #2899, opened 2024-07-12
chore: improving logging under debugDiscv5 flag

gabrielmer:
# Description
Small PR improving logging when using the `debugDiscv5` flag

Thanks @AlbertoSoutullo for noticing 🤩

### DIFF
--- a/waku/node/peer_manager/peer_manager.nim
+++ b/waku/node/peer_manager/peer_manager.nim
@@ -667,7 +667,7 @@ proc connectToNodes*(
     let origin = connectedPeers.mapIt(it.origin)
     if peerIds.len > 0:
       notice "established connections with found peers",
-        peerIds = peerIds, origin = origin
+        peerIds = peerIds.mapIt(shortLog(it)), origin = origin
     else:
       notice "could not connect to new peers", attempted = nodes.len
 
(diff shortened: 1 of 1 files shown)


waku/node/peer_manager/peer_manager.nim
```diff
@@ -665,8 +665,11 @@ proc connectToNodes*(
   when defined(debugDiscv5):
     let peerIds = connectedPeers.mapIt(it.peerId)
     let origin = connectedPeers.mapIt(it.origin)
-    notice "established connections with found peers",
-      peerIds = peerIds, origin = origin
+    if peerIds.len > 0:
+      notice "established connections with found peers",
+        peerIds = peerIds, origin = origin
```
Ivansete-status, line 670: Shall we have a shorter log line?
```suggestion
        peerIds = peers = toSeq(peerIds.mapIt(shortLog(it))), origin = origin
```

Ivansete-status (approved):
LGTM! Thanks for it! :100: 
Just a minor suggestion to keep the log lines shorter

waku/node/peer_manager/peer_manager.nim
gabrielmer: I'm down for a shorter log line, but have to check if this would make the life harder for DST team or not. @AlbertoSoutullo WDYT? Is it the same for you if we log the whole peerId or just the compressed version?

AlbertoSoutullo: The compressed version of the peerId is also fine c:
It doesn't really make a difference

PR #2903, opened 2024-07-14
chore: archive legacy run report metrics every 30 minutes instead of 1

Ivansete-status:
### Description
Simple PR to reduce the stress on the database and only count the number of messages every 30 minutes instead of each minute.


gabrielmer (approved):
LGTM, thank you!

NagyZoltanPeter (approved):
It is ok for long checks, but will it give enough resolution for short term testings?

Ivansete-status:
> It is ok for long checks, but will it give enough resolution for short term testings?

You are right that it seems too low rate for short tests. Nevertheless, the underlying query is too much resource-consuming

PR #2908, opened 2024-07-15
chore: Update master from release v0.30

Ivansete-status:
### Description
This PR is needed to aggregate the changes and enhancements made in the `release/v0.30` branch for the version `v0.30.2`.
I'm using a separate branch because a merge from `master` was required and I didn't want to pollute the `v0.30.2` branch.

In other words, we are retrieving the following changes to `master`:
- set max limit to 100 messages per epoch in RLN
- skip Postgres feedback errors that can be caused when another nwaku instance accesses the same database, as in Status.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,22 @@
+## v0.30.2 (2024-07-08)
+
+### Release highlights
+
+* Patch to avoid exclusive access when creating new partitions in the postgres messages table.
+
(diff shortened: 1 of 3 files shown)


gabrielmer (approved):
Thank you!

NagyZoltanPeter (approved):
Yess, it looks good according to previous changes.

PR #2913, opened 2024-07-16
chore: Add new custom built and test target to make in order to enable easy build or test single nim modules

NagyZoltanPeter:
# Description
While chating with @fryorcraken, he popped up, why don't we have an easy way of running single tests.
We all do this in different ways, like me I'm using some alias to achieve this.

So I made this little experiment to add custom targets to our makefile to ease and make common and available this task.

While it's a bit tricky to achieve such with make and nimscript I found a usable solution.

So the target is to check if a single file you're working on builds or you may write unit test and want to run first separately from other tests.

With this enhancement, you just need to run

`make build/<path-to-your-file>` or `make test/<path-to-your-file>`

Why is it good?
- no prerequisite knowledge needed how we build our software. All knowledge is coded into Makefile and waku.nimble files.
- It will always follow the changes made to build arguments anytime in the future.
- Simplify ramp-up for anybody newly contributing to the project.

# Changes

- [x] Makefile - added `build/...` and `test/...` targets
- [x] Extended waku.nimble with new task that can build any arbitrary single file or even it can run it.

## How to test

run:
1a. make build/tests/common/test_tokenbucket.nim
1b. make test/tests/common/test_tokenbucket.nim

Check: ./build/tests/common/test_tokenbucket.nim.bin exists

# Suggestions/ideas are welcome
It is not trivial to add parametric target to make, but still possible I'm not aware of the proper solution. If you have any better idea or addon to this, I'm happy to discuss.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -179,7 +179,7 @@ testcommon: | build deps
 ##########
 ## Waku ##
 ##########
-.PHONY: testwaku wakunode2 testwakunode2 example2 chat2 chat2bridge
+.PHONY: testwaku wakunode2 testwakunode2 example2 chat2 chat2bridge liteprotocoltester
 
 # install anvil only for the testwaku target
 testwaku: | build deps anvil librln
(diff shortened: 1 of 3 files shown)


gabrielmer (approved):
Amazing initiative! Thanks so much! 😍

gabrielmer:
Now that I think of it, we should mention this on the README so this easier method doesn't get lost/ignored

Ivansete-status (approved):
LGTM! Thanks for it ! :100:

SionoiS (approved):
Awesome!

PR #2916, opened 2024-07-17
feat: add last online timestamp to archive and drivers

SionoiS:
# Description

This PR adds a way to save a timestamp periodically so that we can know when the node was last online.

This feature will be used for a future store resume functionality.

# Changes

- [x] sqlite driver set/get last online timestamp
- [x] postgres driver set/get last online timestamp
- [x] postgres migration
- [x] archive set/get last online timestamp
- [x] archive periodically set last online timestamp

### DIFF
--- a/migrations/message_store_postgres/content_script_version_7.nim
+++ b/migrations/message_store_postgres/content_script_version_7.nim
@@ -1,0 +1,11 @@
+const ContentScriptVersion_7* =
+  """
+-- Create a new table to hold the last online timestamp
+CREATE TABLE last_online (timestamp BIGINT NOT NULL);
+
+-- Insert a placeholder first value
+INSERT INTO last_online (timestamp) VALUES(1);
+
+-- Update to new version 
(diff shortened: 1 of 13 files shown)


migrations/message_store_postgres/content_script_version_7.nim
SionoiS, line 1: Do we really need a migration just to add a new table?

SionoiS:
> Thanks for that @SionoiS ! Nevertheless, I'm curious about the underlying deliverable/milestone :)

It's part of Store sync https://github.com/waku-org/pm/issues/131

Ivansete-status:
> It's part of Store sync [waku-org/pm#131](https://github.com/waku-org/pm/issues/131) `a protocol/heuristic to resume store services after an offline period`

Cool thanks! In that case, I think is better to have a separate sqlite database where we store that info. Similar to the `peers.db` where we store peers data. The reason is that in _Status_ we have two `nwaku` nodes that share the same PostgreSQL database.

PR #2917, opened 2024-07-17
chore: postgres_driver - reduce query cost avoid ORDER BY messageHash

Ivansete-status:
### Description
Add small optimization and avoid performing `ORDER BY messageHash` because it doesn't seem to be needed.

## Issue
closes https://github.com/waku-org/nwaku/issues/2902

### DIFF
--- a/waku/waku_archive/driver/postgres_driver/postgres_driver.nim
+++ b/waku/waku_archive/driver/postgres_driver/postgres_driver.nim
@@ -41,7 +41,7 @@ const SelectNoCursorAscStmtDef =
           pubsubTopic = $3 AND
           timestamp >= $4 AND
           timestamp <= $5
-    ORDER BY timestamp ASC, messageHash ASC LIMIT $6;"""
+    ORDER BY timestamp ASC LIMIT $6;"""
 
 const SelectNoCursorNoDataAscStmtName = "SelectWithoutCursorAndDataAsc"
 const SelectNoCursorNoDataAscStmtDef =
@@ -51,7 +51,7 @@ const SelectNoCursorNoDataAscStmtDef =
           pubsubTopic = $3 AND
           timestamp >= $4 AND
           timestamp <= $5
-    ORDER BY timestamp ASC, messageHash ASC LIMIT $6;"""
+    ORDER BY timestamp ASC LIMIT $6;"""
 
 const SelectNoCursorDescStmtName = "SelectWithoutCursorDesc"
 const SelectNoCursorDescStmtDef =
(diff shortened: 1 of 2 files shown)


NagyZoltanPeter (approved):
Yes, it is a good move as hash does not represents any meaningful ordering within same timestamp.

gabrielmer (approved):
Thank you!

Ivansete-status:
Thank you so much indeed for the reviews guys!
Double-thinking about that, my only concern is regarding a possible issue with cursors.
Even though we have unit tests, is there a way to validate the cursor's behaviors from _status_ app @richard-ramos ?

NagyZoltanPeter:
> Thank you so much indeed for the reviews guys! Double-thinking about that, my only concern is regarding a possible issue with cursors. Even though we have unit tests, is there a way to validate the cursor's behaviors from _status_ app @richard-ramos ?

Do you mean the paging of the result set from the client perspective will need a strict ordering of the result set?

Ivansete-status:
> Do you mean the paging of the result set from the client perspective will need a strict ordering of the result set?

That's potentially the case but I think is not needed because we are already ordering by time :)

NagyZoltanPeter:
> > Do you mean the paging of the result set from the client perspective will need a strict ordering of the result set?
> 
> That's potentially the case but I think is not needed because we are already ordering by time :)

Yes, the only case it can cause issue if two or more message has the same timestamp. What is the real resolution we use to timestamp?

richard-ramos:
> Even though we have unit tests, is there a way to validate the cursor's behaviors from _status_ app @richard-ramos ?

Testing in the status app is somewhat annoying.... I guess by sending messages on a channel that exceed the number of records per page. Currently it is set at 50, so for example, you could send 150 messages from an instance and make sure that you receive the messages (i.e. 101 to 150 first, then 51 to 100, then 0 to 50). You could even modify the status app to use less records per page.

The easiest way would be to use the message finder tool, since it basically does the same thing the status app does. 
```
...
--pagesize=2  \
--forward=false
...
```
And then in the results review that the ordering is okay. 

---

I think the reason why the sorting is done by message hash is described in the specs: https://github.com/waku-org/specs/blob/master/standards/core/store.md#waku-message-sorting
> Where two or more key-value entries have identical timestamps, the entries MUST be further sorted by the natural order of their message hash keys. 

So, if this change is merged it is probably a good idea to update the specs for the protocol

jm-clius (changes requested):
Unfortunately we do need a guaranteed global ordering of stored messages, both for the cursor to work and (most likely) for sane Store Sync heuristics. The timestamp is not guaranteed unique, so can't be what we use (besides, we recommend that timestamp have very coarse resolution - e.g. tick only 1 second boundaries - exactly so that it cannot be used as a de facto index/identifier for the message).

Depending on what the actual cost of the double "order by" is (and if other schema changes can help mitigate the impact), we should consider making `messageHash` an epoch uuid so that it provides global sorting order on its own.

Ivansete-status:
Thanks for the comment @jm-clius ! Considering it, we can no longer merge this PR :)
Aside from that, I'd like to apply your suggestion of having a sortable msg hash that will enhance the queries.

PR #2919, opened 2024-07-18
feat: store resume

SionoiS:
# Description

Adds a way to save a timestamp periodically so that we can know when the node was last online.
At node start, call a store node and ask for the messages since the last local online timestamp.

# Changes

- [x] DB
- [x] periodic loop
- [x] new config field
- [x] basic wakunode setup
- [x] msg transfer callback

### DIFF
--- a/waku/node/waku_node.nim
+++ b/waku/node/waku_node.nim
@@ -34,6 +34,7 @@ import
   ../waku_store/protocol as store,
   ../waku_store/client as store_client,
   ../waku_store/common as store_common,
+  ../waku_store/resume,
   ../waku_filter_v2,
   ../waku_filter_v2/client as filter_client,
   ../waku_filter_v2/subscriptions as filter_subscriptions,
@@ -94,6 +95,7 @@ type
     wakuLegacyStoreClient*: legacy_store_client.WakuStoreClient
     wakuStore*: store.WakuStore
     wakuStoreClient*: store_client.WakuStoreClient
+    wakuStoreResume*: StoreResume
     wakuFilter*: waku_filter_v2.WakuFilter
     wakuFilterClient*: filter_client.WakuFilterClient
     wakuRlnRelay*: WakuRLNRelay
@@ -954,6 +956,13 @@ proc query*(
     return ok(res)
 
   return ok(response)
+
+proc setupStoreResume*(node: WakuNode) =
+  info "setting up store resume functionality"
+
+  node.wakuStoreResume = StoreResume.new().valueOr:
+    error "Failed to setup Store Resume", error
+    return 
 
 ## Waku lightpush
 
--- a/waku/node/waku_node.nim
+++ b/waku/node/waku_node.nim
@@ -958,11 +958,11 @@ proc query*(
   return ok(response)
 
 proc setupStoreResume*(node: WakuNode) =
-  info "setting up store resume functionality"
-
-  node.wakuStoreResume = StoreResume.new().valueOr:
-    error "Failed to setup Store Resume", error
-    return 
+  node.wakuStoreResume = StoreResume.new(
+    node.peerManager, node.wakuArchive, node.wakuStoreClient
+  ).valueOr:
+    error "Failed to setup Store Resume", error = $error
+    return
 
 ## Waku lightpush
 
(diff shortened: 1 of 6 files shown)


waku/node/waku_node.nim
```diff
@@ -94,6 +95,7 @@ type
     wakuLegacyStoreClient*: legacy_store_client.WakuStoreClient
     wakuStore*: store.WakuStore
     wakuStoreClient*: store_client.WakuStoreClient
+    wakuStoreResume*: StoreResume
```
Ivansete-status, line 98: Unless I'm missing something I think we should avoid that. Instead, I think the best place is to move it to the `Waku` instance.

Ivansete-status (commented):
Thanks for it! I've added some comments that I hope you find interesting :)

waku/node/waku_node.nim
SionoiS: The logic will be expanded to include calling a Store client query in a callback. Store Client, Archive and Store Resume need to be tied together in `waku_node`.

gabrielmer (approved):
LGTM! Thank you!

gabrielmer (approved):
LGTM! Thanks!

jm-clius (approved):
Looks fine to me as an experimental feature. Would be interesting to see impact both on clients and resume service peers.

NagyZoltanPeter (approved):
It is always insightful to review your code!
Maybe it would worth a bit more commentary how it works as it easy to lost in the many redirection and callbacks. :-)
Thank you for it!

waku/node/waku_node.nim
```diff
@@ -955,6 +957,13 @@ proc query*(
 
   return ok(response)
 
+proc setupStoreResume*(node: WakuNode) =
```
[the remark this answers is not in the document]
SionoiS: I wanted to distinguish from mounting as it's not what is happening, it's not a protocol.

SionoiS:
> It is always insightful to review your code! Maybe it would worth a bit more commentary how it works as it easy to lost in the many redirection and callbacks. :-) Thank you for it!

What part do you feel would benefit most from comments? The callback part?

NagyZoltanPeter:
> > It is always insightful to review your code! Maybe it would worth a bit more commentary how it works as it easy to lost in the many redirection and callbacks. :-) Thank you for it!
> 
> What part do you feel would benefit most from comments? The callback part?

Yes, it is not obvious how and when the transferCallBack runs.

waku/node/waku_node.nim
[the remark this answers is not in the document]
NagyZoltanPeter: Exactly, I agree. I only refer to the same happens with sharding, rlnRelay, which are more decorators not protocols.... I can accept and later we may have a separate PR fixing all these and make them common naming. WDYT?

NagyZoltanPeter (commented):
@SionoiS : I suggest to add a comment that explains when the transferCallback is running by design and how many times.
After reading it carefully I see, it is running at start with 3 trial. Is it so?
But one more thing I could not find, when does the periodicSetLastOnline is running?
It is stored in the handle, but I see no-one calls that proc, but maybe I miss something. If I missed it, please add comment :-)

SionoiS:
> @SionoiS : I suggest to add a comment that explains when the transferCallback is running by design and how many times. After reading it carefully I see, it is running at start with 3 trial. Is it so? But one more thing I could not find, when does the periodicSetLastOnline is running? It is stored in the handle, but I see no-one calls that proc, but maybe I miss something. If I missed it, please add comment :-)

Yes there's a lot of Nim magic in there (which I dislike). AFAIK chrono's futures are eager, so creating one starts it. I'll enhance my comments.

waku/waku_store/resume.nim
```diff
@@ -208,6 +209,7 @@ proc start*(self: StoreResume) {.async.} =
 
     break
 
+  # starting periodic storage of last online timestamp
   self.handle = self.periodicSetLastOnline()
```
NagyZoltanPeter, line 213: Wow, never seen like this.... I used asyncSpawn for such cases.

NagyZoltanPeter (approved):
LGTM, huge thank you for tackling my comments.

waku/waku_store/resume.nim
SionoiS: I like to think about `AsyncSpawn` as tasks that don't need to be canceled. Having a defined `handle` makes cancelling easy.

PR #2920, opened 2024-07-19
chore: Distinction between gross/net trafic in bandwidth per shard metric, a…

NagyZoltanPeter:
...added bandwidths and request rate panels to single node and fleet dashboards

# Description
This was a leftover task after @Ivansete-status comments on https://github.com/waku-org/nwaku/pull/2851
Now gross and net traffic of incoming per shard relay measured separately.
Gross traffic means all incoming relay traffic (measured before validation). It will help us to understand how much of the bandwidth stress on the node vs. useful relay traffic that is actually valid and propagated forward. This is now measured per shard.
 
# Changes

- [X] Introduced new counter filter of "type" for waku_relay_network_bytes metric, that represents distinction of gross/net traffic
- [X] In connection to #2851 I pulled and adjusted metrics panels to single node and fleet dashboards.
    - [X] Added `waku relay traffic per shads in/out`
    - [X] Added `request rate`, `non relay traffic in/out` separately in respective rows of store, filter and lightpush

Unfortunately, I cannot create pictures of working dashboards as it is not simple to try out the changes on fleet dashboard.

## Issue
https://github.com/waku-org/nwaku/issues/1945

## Connected PR
nwaku-compose respective changes: https://github.com/waku-org/nwaku-compose/pull/99

### DIFF
--- a/metrics/waku-fleet-dashboard.json
+++ b/metrics/waku-fleet-dashboard.json
@@ -25,7 +25,6 @@
   "editable": true,
   "fiscalYearStartMonth": 0,
   "graphTooltip": 1,
-  "id": 36,
   "links": [],
   "liveNow": false,
   "panels": [
@@ -108,6 +107,8 @@
       },
       "id": 52,
       "options": {
+        "minVizHeight": 75,
+        "minVizWidth": 75,
         "orientation": "auto",
         "reduceOptions": {
           "calcs": [
@@ -118,9 +119,10 @@
         },
         "showThresholdLabels": false,
         "showThresholdMarkers": true,
+        "sizing": "auto",
         "text": {}
       },
-      "pluginVersion": "9.2.5",
+      "pluginVersion": "10.4.2",
       "targets": [
         {
           "datasource": {
@@ -180,13 +182,15 @@
           "fields": "",
           "values": false
         },
+        "showPercentChange": false,
         "text": {
           "titleSize": 18,
           "valueSize": 20
         },
-        "textMode": "auto"
+        "textMode": "auto",
+        "wideLayout": true
       },
-      "pluginVersion": "9.2.5",
+      "pluginVersion": "10.4.2",
       "targets": [
         {
           "datasource": {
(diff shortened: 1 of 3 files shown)


gabrielmer (commented):
Thank you! Could you please add a screenshot of the new dashboard?

NagyZoltanPeter:
> Thank you! Could you please add a screenshot of the new dashboard?

@gabrielmer : Actually these dashboards, specially the fleet one needs special environment.
But if you have an idea how to run grafana that way to show meaningful data on those panels... I will be thankful.

gabrielmer:
> 
> @gabrielmer : Actually these dashboards, specially the fleet one needs special environment. But if you have an idea how to run grafana that way to show meaningful data on those panels... I will be thankful.

Not sure I understand then, so how did these files get generated? I thought you created the dashboards in Grafana and update the files with the ones of the dashboard you built

NagyZoltanPeter:
> > @gabrielmer : Actually these dashboards, specially the fleet one needs special environment. But if you have an idea how to run grafana that way to show meaningful data on those panels... I will be thankful.
> 
> Not sure I understand then, so how did these files get generated? I thought you created the dashboards in Grafana and update the files with the ones of the dashboard you built

Well I was using nwaku-compose, I put them there, but prometheus under nwaku-compose do not provide the right collection to populate the dashboard with data. Like for fleet we have datacenter, node type and other distinctions, but for that you would need to setup properly the whole, which I did not. I was copying from nwaku-compose's dashboard the panels and adapt the queries. Kinda' hack.

gabrielmer:
> Well I was using nwaku-compose, I put them there, but prometheus under nwaku-compose do not provide the right collection to populate the dashboard with data. Like for fleet we have datacenter, node type and other distinctions, but for that you would need to setup properly the whole, which I did not. I was copying from nwaku-compose's dashboard the panels and adapt the queries. Kinda' hack.

Got it! Yeah I'm also not sure how/where these panels are supposed to be run, I didn't know they existed.

@Ivansete-status do you know where are they used and how they can be tested? Are they specifically for https://grafana.infra.status.im/ ?

Ivansete-status (approved):
LGTM! Thanks for it! :100: 
If possible it'd help to have screenshot of new panels.
I wonder though how can we deploy the new dashboards into the fleet's Grafana

gabrielmer (approved):
Approving also not to block with the monitoring part :)) Thank you!

PR #2923, opened 2024-07-19
chore: return all connected peers from REST API

weboko:
# Description

Currently [GET `/admin/v1/peers`](https://waku-org.github.io/waku-rest-api/#get-/admin/v1/peers) has additional logic to build a list of peers service node is connected to. 
It does go through each group of protocols and builds a list:
https://github.com/waku-org/nwaku/blob/6d385cefef88ad80f41e830d904d68da596cbbde/waku/waku_api/rest/admin/handlers.nim#L43 

We can directly return list of connected peers as it more aligned with description of the endpoint: 
> Retrieve information about connected peers.

### DIFF
--- a/tests/wakunode_rest/test_rest_admin.nim
+++ b/tests/wakunode_rest/test_rest_admin.nim
@@ -21,6 +21,7 @@ import
     waku_api/rest/admin/handlers as admin_api,
     waku_api/rest/admin/client as admin_api_client,
     waku_relay,
+    waku_peer_exchange,
   ],
   ../testlib/wakucore,
   ../testlib/wakunode,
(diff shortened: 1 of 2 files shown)


NagyZoltanPeter (changes requested):
Sorry, but it is not working this way. We have a API response structure we must confirm with. So we return peers with their supported protocols.
This issue here - I think - that we won't return peers with protocol if the same protocol we don't support. (we = the node queried by GET /admin/peers). I think we should return all peers are connected and list all the protocols they support.

NagyZoltanPeter:
@weboko : Please check, I updated your PR with the thought functional change.
My question if this is enough for a client (admin) purposes. Do we really need all peers with all protocols?
cc: @jm-clius

NagyZoltanPeter (approved):

gabrielmer (approved):
LGTM, thank you!

Maybe we should add a test case for this?

Ivansete-status (approved):

NagyZoltanPeter:
> LGTM, thank you!
> 
> Maybe we should add a test case for this?

Yepp, done!

PR #2927, opened 2024-07-23
fix: add back waku discv5 metrics

SionoiS:
I noticed that waku discv5 metrics were not used so I added them back.

### DIFF
--- a/waku/discovery/waku_discv5.nim
+++ b/waku/discovery/waku_discv5.nim
@@ -235,7 +235,7 @@ proc searchLoop(wd: WakuDiscoveryV5) {.async.} =
       let peerInfo = record.toRemotePeerInfo().valueOr:
         ## in case of error, we keep track of it for debugging purposes
         wrongRecordsReasons.add(($record, $error))
-        waku_discv5_errors.inc()
+        waku_discv5_errors.inc(labelValues = [$error])
         continue
 
       discoveredPeers.add(peerInfo)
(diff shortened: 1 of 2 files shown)


waku/discovery/waku_discv5.nim
```diff
@@ -211,6 +211,8 @@ proc findRandomPeers*(
   elif wd.predicate.isSome():
     discoveredRecords = discoveredRecords.filter(wd.predicate.get())
 
+  waku_discv5_discovered.inc(discoveredRecords.len)
```
gabrielmer, line 214: The only concern that I have is if we receive the same peer twice, we will update this metric both times.

Another alternative is increasing it in `addPeer()` if the origin is `PeerOrigin.Discv5` after passing the duplicate checks

SionoiS: :thinking:  I thought we had another metric for that...

gabrielmer: Sounds good :) Thank you!

gabrielmer (approved):
LGTM, thanks so much!

waku/discovery/waku_discv5.nim
```diff
@@ -233,6 +235,7 @@ proc searchLoop(wd: WakuDiscoveryV5) {.async.} =
       let peerInfo = record.toRemotePeerInfo().valueOr:
         ## in case of error, we keep track of it for debugging purposes
         wrongRecordsReasons.add(($record, $error))
+        waku_discv5_errors.inc()
```
Ivansete-status, line 238: That's nice! Maybe we could also add error detail
```suggestion
        waku_discv5_errors.inc(labelValues = [$error])
```

Ivansete-status (approved):
LGTM! Thanks for it :100: !
I just added a minor suggestion

PR #2931, opened 2024-07-25
chore: Simplification of store legacy code

Ivansete-status:
## Description
While enhancing postgres queries I noticed that there is some conflict between legacy-store and future store.
This PR is aimed to mitigate those collisions by removing unnecessary code and giving the responsibility of partition management, and retention policy, only to the new-future store archive driver.


gabrielmer (approved):
Thanks so much! Apologies for the delay. For some reason I had already review it but forgot to mark it as approved 😶

tests/waku_archive_legacy/test_driver_postgres.nim
```diff
@@ -4,11 +4,13 @@ import std/[sequtils, options], testutils/unittests, chronos
 import
   waku/waku_archive_legacy,
   waku/waku_archive_legacy/driver/postgres_driver,
+  waku/waku_archive/driver/postgres_driver as new_postgres_driver,
```
NagyZoltanPeter, line 7: Me personally better like to tag the legacy one with alias instead of the new. Due it will ease the work of removing.

NagyZoltanPeter (approved):
It's great, thank you.

tests/waku_archive_legacy/test_driver_postgres.nim
Ivansete-status: > Me personally better like to tag the legacy one with alias instead of the new. Due it will ease the work of removing.

Yes good point! We will be able to completely remove the whole `tests/waku_archive_legacy/test_driver_postgres.nim` file in the future

PR #2932, opened 2024-07-25
chore: add peer filtering by cluster for waku peer exchange

### LINKED ISSUE #2910: chore: filter peers stored in cache by cluster-id in peer-exchange
### Background
This is inspired by https://github.com/waku-org/go-waku/pull/1139

If the `cluster-id` is specified, then the waku peer exchange protocol should only consider peers with that given `cluster-id`.

SionoiS:
# Description

The peer pool in waku peer exchange no longer accept peers from other cluster.

Closes #2910

### DIFF
--- a/waku/waku_peer_exchange/protocol.nim
+++ b/waku/waku_peer_exchange/protocol.nim
@@ -129,20 +129,25 @@ proc getEnrsFromCache(
   # return numPeers or less if cache is smaller
   return shuffledCache[0 ..< min(shuffledCache.len.int, numPeers.int)]
 
+proc poolFilter*(cluster: Option[uint16], peer: RemotePeerInfo): bool =
+  if peer.origin != Discv5:
+    trace "peer not from discv5", peer = $peer, origin = $peer.origin
+    return false
+
+  if peer.enr.isNone():
+    trace "peer has no ENR", peer = $peer
+    return false
+
+  if cluster.isSome() and peer.enr.get().isClusterMismatched(cluster.get()):
+    trace "peer has mismatching cluster", peer = $peer
+    return false
+
+  return true
+
 proc populateEnrCache(wpx: WakuPeerExchange) =
   # share only peers that i) are reachable ii) come from discv5 iii) share cluster
-  let withEnr = collect(newSeq):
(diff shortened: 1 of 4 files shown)


gabrielmer (commented):
Looks greatt, thank you!

I think we should add a test case in the peer exchange suite, WDYT?

Also, in this PR too the linting check is failing, do you use nph version 0.5.1?

waku/waku_peer_exchange/protocol.nim
```diff
@@ -129,11 +130,19 @@ proc getEnrsFromCache(
...
+  let withEnr = collect(newSeq):
+    for peer in wpx.peerManager.peerStore.getReachablePeers():
+      if peer.origin != Discv5:
+        continue
+
+      if peer.enr.isNone():
+        continue
+
+      if wpx.cluster.isSome() and peer.enr.get().isClusterMismatched(wpx.cluster.get()):
+        continue
```
Ivansete-status, line 143: Maybe we can add trace logs under this conditions to help debugging in the future

Ivansete-status (approved):
LGTM thanks! Approving to not block it and happy to merge after Gabriel's comments are fixed :)

SionoiS:
> Looks greatt, thank you!
> 
> I think we should add a test case in the peer exchange suite, WDYT?
> 
> Also, in this PR too the linting check is failing, do you use nph version 0.5.1?

I'll add a test and check my nph version

gabrielmer:
> sionois@laptop:~/Github/nwaku$ nph --version v0.5.1-0-gde5cd48` is that not the correct version?
> 
> I think switching to the new nim vscode extention has something to do with it. I'll try to fix my setup.

Yes, that's the correct version! When I checkout to your branch I get the file formatted differently, so I think it's something in your setup

SionoiS:
> > sionois@laptop:~/Github/nwaku$ nph --version v0.5.1-0-gde5cd48` is that not the correct version?
> > I think switching to the new nim vscode extention has something to do with it. I'll try to fix my setup.
> 
> Yes, that's the correct version! When I checkout to your branch I get the file formatted differently, so I think it's something in your setup

Should be fixed now!

gabrielmer (approved):
Amazingg, thanks so much!

PR #2933, opened 2024-07-25
chore: Optimize hash queries with lookup table

Ivansete-status:
## Description
This PR enhances the performance of the most expensive query that we have at the moment, i.e., `... WHERE messageHash IN (...)`

The enhancement leverages on the use a lookup table ( `messages_lookup` )

# Changes

- [ ] Change schema to add the `messages_lookup` table
- [ ] ...

## Issue

closes https://github.com/waku-org/nwaku/issues/2895


gabrielmer (approved):
Niiice! LGTM!

Thanks so much! Do we yet have any indication/way to compare how much better this performs than without the lookup table?

SionoiS (approved):
LGTM Thanks! :racehorse: faster!

richard-ramos (approved):
Awesome! looking fwd to see this running in the fleet :)

migrations/message_store_postgres/content_script_version_7.nim
```diff
@@ -0,0 +1,29 @@
...
+
+-- Create lookup table
+CREATE TABLE IF NOT EXISTS messages_lookup (
+   timestamp BIGINT NOT NULL,
+   messageHash VARCHAR NOT NULL,
+   CONSTRAINT messageIndexLookupTable PRIMARY KEY (messageHash, timestamp)
+  );
+
+-- Put data into lookup table
+INSERT INTO messages_lookup (messageHash, timestamp) SELECT messageHash, timestamp from messages;
```
richard-ramos, line 12: I do not know  how slow this operation is, but since it's doing a batch insert, a way to speed up the insertion could be to create the table without the primary key, and then, after the insertion is complete, do:
```sql
ALTER TABLE messages_lookup ADD CONSTRAINT messageIndexLookupTable PRIMARY KEY (messageHash, timestamp)
```
This should be faster because it eliminates the overhead of maintaining the primary key index during each insert operation.

richard-ramos: (of course, if the insertion takes just a couple of minutes then doing this optimization is an overkill)

Ivansete-status:
> Niiice! LGTM!
> 
> Thanks so much! Do we yet have any indication/way to compare how much better this performs than without the lookup table?

Yes, this can be checked through _Status_ . I will suggest to dogfood that in our weekly _Status_ dogfooding session

migrations/message_store_postgres/content_script_version_7.nim
Ivansete-status: cool thanks! I've applied the suggestion in https://github.com/waku-org/nwaku/pull/2933/commits/cb80df4f1b6035040b3fe7328f094553e57448dc

PR #2938, opened 2024-07-26
chore: lock in nph version and add pre-commit hook

fryorcraken:
# Description

- Lock in `nph` version to `v0.5.1` by bringing it as a submodule
- Add convenient make tasks to build and run `nph`
- Add convenient make task to install hook that will format nim files at commit time.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -245,11 +245,21 @@ networkmonitor: | build deps librln
 ############
 ## Format ##
 ############
-.PHONY: build-nph clean-nph
+.PHONY: build-nph clean-nph install-nph
 
 build-nph:
 ifeq ("$(wildcard $(NPH))","")
 	$(ENV_SCRIPT) nim c vendor/nph/src/nph.nim
+endif
+
+GIT_PRE_COMMIT_HOOK := .git/hooks/pre-commit
+
+install-nph: build-nph
+ifeq ("$(wildcard $(GIT_PRE_COMMIT_HOOK))","")
+	cp ./scripts/git_pre_commit_format.sh $(GIT_PRE_COMMIT_HOOK)
+else
+	echo "$(GIT_PRE_COMMIT_HOOK) already present, will NOT override"
+	exit 1
 endif
 
 nph/%: build-nph
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,8 @@ BUILD_SYSTEM_DIR := vendor/nimbus-build-system
 EXCLUDED_NIM_PACKAGES := vendor/nim-dnsdisc/vendor
 LINK_PCRE := 0
 LOG_LEVEL := TRACE
-
+NPH := vendor/nph/src/nph
+FORMAT_MSG := "\\x1B[95mFormatting:\\x1B[39m"
 # we don't want an error here, so we can handle things later, in the ".DEFAULT" target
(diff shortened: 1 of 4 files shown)


NagyZoltanPeter (commented):
Awesome! 
I think we need to eforce explicit nph version.
current one we use is `v0.5.1-0-gde5cd48`

fryorcraken:
> Awesome! I think we need to eforce explicit nph version. current one we use is `v0.5.1-0-gde5cd48`

ok, I should be able to do that. Will update.

Makefile
```diff
@@ -256,6 +256,27 @@ coverage:
 	echo -e $(BUILD_MSG) "build/$@" && \
 		$(ENV_SCRIPT) ./scripts/run_cov.sh -y
 
+################
+## Formatting ##
+################
+.PHONY: check_nph_installed check_precommit_hook install_nph_prehook
+
+check_nph_installed:
+ifeq (, $(shell which nph 2> /dev/null))
```
Ivansete-status, line 265: This is nice indeed! Nevertheless, I think it is better to have this check implicit when the developer performs any `make` command. I mean, this is a pre-condition that should always be met and I think it will be more secure to enforce that in a generic manner.

README.md
```diff
@@ -84,6 +84,15 @@ Binary will be created as `<path to your test file.nim>.bin` under the `build` d
...
 ```
 
+### Formatting Code
+
+We use [`nph`](https://github.com/arnetheduck/nph) to format nim code.
+You can install a convenient pre-commit git hook with the following command:
+
+```
+make install_nph_prehook
+```
```
Ivansete-status, line 94: Considering that the hook will always be triggered, we don't need the developer to explicitly do `make install_nph_prehook`. In case my other comment makes sense of course :)
```suggestion
```

Ivansete-status (commented):
That's awesome! Thanks so much for it!
I just added some comments that I hope you find useful

Makefile
fryorcraken: I would disagree with that. If I am a user that just want to build `wakunode2` because I want latest master or testing a branch, the presence of `nph` should not block me to proceed.

README.md
fryorcraken: I don't think we should install pre-hooks nilly-willy, especially if a dev has already their own prehooks in place. Which is why I set it up as separate step.

fryorcraken:
Having some difficulties building `nph` when checked out as a git submodule: https://github.com/arnetheduck/nph/issues/81

README.md
```diff
@@ -84,6 +85,19 @@ Binary will be created as `<path to your test file.nim>.bin` under the `build` d
 make test/tests/common/test_enr_builder.nim
 ```
 
+## Formatting
+
+Nim files are expected to be formatted using the [`nph`](https://github.com/arnetheduck/nph) version present in `vendor/nph`.
+
+You can easily format file with the `make nph/<relative path to nim> file` command.
```
Ivansete-status, line 92: Shall we have an example of how to format a particular file? That would help to understand the command :)

( maybe it sounds more natural having `make nph <relative-path-to-nim-file>` )

Ivansete-status (approved):
LGTM! Thanks for it! :100: 
I just added a nitpick comment that hope you find useful :)

gabrielmer (approved):
LGTM, thank you!

PR #2944, opened 2024-07-30
chore: update master from release/v0.31 branch

Ivansete-status:
### Description
We cherry picked a couple of commits that are already present in `master`


Ivansete-status:
I closed it because it no longer makes sense as we already created the v0.32.0 and `CHANGELOG` has been properly updated

PR #2945, opened 2024-07-31
chore: prepare_release.md simplify and relax CCs usage in status.staging

Ivansete-status:
## Description
Simple PR to make the CCs dogfooding point less strict in the sense that is not blocking. We should inform and advise other CCs to use _Status_ app against `status.staging` but that shouldn't block the expected release date.


gabrielmer (approved):
Thank you!

darshankabariya (approved):
LGTM !

NagyZoltanPeter (approved):
Great!

PR #2947, opened 2024-07-31
chore: optimizations for discv5 only nodes

### LINKED ISSUE #2892: bug: Node accepts connections when configured for just discovery
### Problem

A node that is configured as a "boostrap" node without mounting any protocol such as `relay`, `store`, etc is meant to act as a "discovery only bootstrap peer". See example configuration:

```
--relay=false\
--discv5-discovery=true\
--discv5-enr-auto-update=True\
--pubsub-topic=/waku/2/rs/66/0\
--cluster-id=66
```

However, with this configuration, this node is still mounting the libp2p switch + protocol such as `metadata` or `identify` and accept external connections. This doesn't make much sense since the node is just providing discovery capabilities on discv5 level and no service is offered on libp2p level.

In other words, a node not mounting any waku protocol, shouldn't accept any connection. The proposed solution would be to not mount the libp2p switch if `relay`, `store`, ... are all false.

This behaviour can be reproduced with [waku-simulator](https://github.com/waku-org/waku-simulator) where the [boostrap node](https://github.com/waku-org/waku-simulator/blob/master/run_bootstrap.sh) acts as discovery-only.

image

@AlbertoSoutullo 
@stubbsta

gabrielmer:
# Description
Adding the `discv5-only` CLI flag for bootstrap/discovery nodes that only run discv5. In those cases, we don't mount the libp2p switch nor additional protocols to avoid unnecessary connections

Notice that I changed the approach from `bootstrap-only` to `discv5-only` because bootstrap nodes can use `waku-peer-exchange`, and that protocol does use the nim-libp2p switch. So this optimization works only in the case that it's running Discv5 and nothing else.

## Issue

closes #2892

### DIFF
--- a/examples/wakustealthcommitments/node_spec.nim
+++ b/examples/wakustealthcommitments/node_spec.nim
@@ -34,7 +34,6 @@ proc setup*(): Waku =
   # Override configuration
   conf.maxMessageSize = twnClusterConf.maxMessageSize
   conf.clusterId = twnClusterConf.clusterId
-  conf.rlnRelay = twnClusterConf.rlnRelay
   conf.rlnRelayEthContractAddress = twnClusterConf.rlnRelayEthContractAddress
   conf.rlnRelayDynamic = twnClusterConf.rlnRelayDynamic
   conf.rlnRelayBandwidthThreshold = twnClusterConf.rlnRelayBandwidthThreshold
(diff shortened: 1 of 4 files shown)


Ivansete-status (approved):
LGTM! Thanks for it! :100:

SionoiS (approved):
LGTM! Thank you.

NagyZoltanPeter (approved):
LGTM!

PR #2951, opened 2024-08-01
chore: lightpush better feedback if lightpush service node without peers

Ivansete-status:
## Description
Give better feedback in case the lightpush server node cannot relay messages to other peers

## Issue

closes https://github.com/waku-org/nwaku/issues/2950

### DIFF
--- a/tests/node/test_wakunode_lightpush.nim
+++ b/tests/node/test_wakunode_lightpush.nim
@@ -1,7 +1,7 @@
 {.used.}
 
 import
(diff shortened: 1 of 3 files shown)


NagyZoltanPeter (approved):
Technically it is ok. I see the request from Status, while original agreement was that for this "legacy" lightpush that it is not considered as an error. 
This will change with lightpush v2, under dev. 
Consider this will break js-waku or at least needs them to adapt.

gabrielmer (approved):
LGTM! Thank you!

PR #2953, opened 2024-08-05
chore: upgrade peer exchange mounting 

darshankabariya:
This PR upgrades the mounting scheme for peer_exchange. With this update, a specific node address is no longer required when mounting peer_exchange from the node side.

### DIFF
--- a/waku/factory/external_config.nim
+++ b/waku/factory/external_config.nim
@@ -579,6 +579,13 @@ type WakuNodeConf* = object
       name: "peer-exchange"
     .}: bool
 
+    peerExchangeNode* {.
+      desc:
+        "Peer multiaddr to send peer exchange requests to. (enables peer exchange protocol requester side)",
+      defaultValue: "",
+      name: "peer-exchange-node"
+    .}: string
+
     ## websocket config
     websocketSupport* {.
       desc: "Enable websocket:  true|false",
--- a/waku/factory/external_config.nim
+++ b/waku/factory/external_config.nim
@@ -579,13 +579,6 @@ type WakuNodeConf* = object
       name: "peer-exchange"
     .}: bool
 
-    peerExchangeNode* {.
-      desc:
-        "Peer multiaddr to send peer exchange requests to. (enables peer exchange protocol requester side)",
-      defaultValue: "",
-      name: "peer-exchange-node"
-    .}: string
-
     ## websocket config
     websocketSupport* {.
       desc: "Enable websocket:  true|false",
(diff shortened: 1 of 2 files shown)


gabrielmer:
Not sure I understand the change.

Before this change there were 2 cases:
1. only `peerExchangeNode` was provided and in that case, we mounted the protocol and set the service peer with the address provided
2. `peerExchange` is set to `true` but without a `peerExchangeNode` provided, in which case we mount the protocol but don't set a service peer
3. none of them are set to `true` and we don't mount peer exchange

Not sure what's wrong with that logic, maybe I'm missing something? But I see that with the changes now, we are eliminating the possibility of configuring the node as a peer exchange client

darshankabariya:
> Not sure I understand the change.
> 
> Before this change there were 2 cases:
> 
> 1. only `peerExchangeNode` was provided and in that case, we mounted the protocol and set the service peer with the address provided
> 2. `peerExchange` is set to `true` but without a `peerExchangeNode` provided, in which case we mount the protocol but don't set a service peer
> 3. none of them are set to `true` and we don't mount peer exchange
> 
> Not sure what's wrong with that logic, maybe I'm missing something? But I see that with the changes now, we are eliminating the possibility of configuring the node as a peer exchange client

This change is based on the chat in [this Discord thread](https://discord.com/channels/1110799176264056863/1268927524315070494). Previously, nodes (server-side) would only add the node address if it provided the peer exchange service. Now, that's not necessary; any client can ask any node that provides the peer exchange service. The old method was somewhat permission-based, where only the node could decide who received the peer exchange service. The new way is more permissionless.

Any suggestions or corrections are appreciated, @gabrielmer. 
Please let me know if I misunderstood anything, @Ivansete-status.

gabrielmer:
But we currently don't need a `peerExchangeNode` in order to mount peer exchange. The condition to mount it is

```
if conf.peerExchangeNode != "" or conf.peerExchange:
```

The condition to require a `peerExchangeNode` to mount it would be with an `and` instead of an `or`. But with the `or`, if there's no `peerExchangeNode` and `peerExchange` is `true` then we do mount it.

Again, I might be missing something 😶

darshankabariya:
> But we currently don't need a `peerExchangeNode` in order to mount peer exchange. The condition to mount it is
> 
> ```
> if conf.peerExchangeNode != "" or conf.peerExchange:
> ```
> 
> The condition to require a `peerExchangeNode` to mount it would be with an `and` instead of an `or`. But with the `or`, if there's no `peerExchangeNode` and `peerExchange` is `true` then we do mount it.
> 
> Again, I might be missing something 😶

I just updated the PR. Can you please look into it and try isolating both things so nothing hurts 😊?

jm-clius (approved):
Looks reasonable to me. :)

gabrielmer (approved):
LGTM! Thanks so much!

PR #2957, opened 2024-08-05
chore: adapt SubscriptionEvents to allow subscribe to pubsub/content topics

Ivansete-status:
This is needed from the PoV of filter so that the delivery monitor can be aware of the pubsub/content topics the node is interested in


Ivansete-status:
I'm closing this as I don't like this approach 100%. I think is better to use observer-observable pattern

PR #2958, opened 2024-08-06
chore: using mplex instead of yamux

gabrielmer:
# Description
Switching our multiplexer from `yamux` to `mplex` after finding the Futures leak in `yamux` reported in https://github.com/vacp2p/nim-libp2p/issues/1165

Once it's taken care of we can switch back to `yamux`

### DIFF
--- a/waku/node/waku_switch.nim
+++ b/waku/node/waku_switch.nim
@@ -86,7 +86,6 @@ proc newWakuSwitch*(
     .withMaxIn(maxIn)
     .withMaxOut(maxOut)
     .withMaxConnsPerPeer(maxConnsPerPeer)
-    .withYamux()
     .withMplex(inTimeout, outTimeout)
     .withNoise()
     .withTcpTransport(transportFlags)


jm-clius (approved):
Thanks for swift action here. I notice a corresponding issue has been raised in nim-libp2p. Has someone from that team been made aware and are they looking into it?

gabrielmer:
> I notice a corresponding issue has been raised in nim-libp2p. Has someone from that team been made aware and are they looking into it?

I just opened the issue, however they have already been informed yesterday through a Discord [message](https://discord.com/channels/613988663034118151/740631022080622745/1269944269318127658)

Not sure if they started taking a look into it yet, I was waiting a bit to get some kind of answer/update on the new issue and if not ask directly. In any case, I will keep in top of this :))

richard-ramos:
We will need to add back to go-waku support to mplex since go-libp2p dropped support to it. Do you think this is something that we should do soon? or do you think the issue with yamux will be fixed before next nwaku release?

gabrielmer:
> We will need to add back to go-waku support to mplex since go-libp2p dropped support to it. Do you think this is something that we should do soon? or do you think the issue with yamux will be fixed before next nwaku release?

Oh thank you, I wasn't aware of this. So I will create a thread on the Vac's `p2p-pm` Discord channel to see if they can estimate how long it will take to fix it. I'm personally not very familiar with the muxers' internals so couldn't propose a fix myself.

Will update on how it goes, but won't merge this until we coordinating it with `go-waku`. Depending on the estimation, either we wait for the fix or add again support for `mplex` in `go-waku`. Hopefully it's the former option.

gabrielmer:
It seems that a fix for the leak was found :)) so it seems that we won't need to switch back to `mplex`. Just in case, I'll close this PR once the fix is merged to `nim-libp2p`

cc @richard-ramos

lchenut:
For your information, the fix is merged (https://github.com/vacp2p/nim-libp2p/pull/1171).
@kaiserd will probably add it to the 1.5 release (https://github.com/vacp2p/nim-libp2p/pull/1166).

gabrielmer:
Closing PR as the leak is fixed in `nim-libp2p`. Thanks you @lchenut !

PR #2965, opened 2024-08-09
chore: pruning excess in relay connections

### LINKED ISSUE #2929: chore: review max-relay-peers config attribute in wakunode2
### Background
We have the `max-relay-peers` parameter which is expected to limit the max number of relay peers that a wakunode2 should support.

This is very interesting from the `waku.sandbox` fleet, f.e., because we usually have there 300 libp2p connections that are only used for relay and therefore, there is no room for other protocol such as store, filter, or lightpush.

```nim
waku_connected_peers{direction="In",protocol="/vac/waku/filter-push/2.0.0-beta1"} 0.0
waku_connected_peers{direction="Out",protocol="/vac/waku/store/2.0.0-beta4"} 0.0
waku_connected_peers{direction="In",protocol="/vac/waku/relay/2.0.0"} 277.0
waku_connected_peers{direction="In",protocol="/vac/waku/filter/2.0.0-beta1"} 0.0
waku_connected_peers{direction="Out",protocol="/vac/waku/filter/2.0.0-beta1"} 0.0
waku_connected_peers{direction="Out",protocol="/vac/waku/filter-subscribe/2.0.0-beta1"} 0.0
waku_connected_peers{direction="Out",protocol="/vac/waku/lightpush/2.0.0-beta1"} 0.0
waku_connected_peers{direction="In",protocol="/vac/waku/filter-subscribe/2.0.0-beta1"} 0.0
waku_connected_peers{direction="In",protocol="/vac/waku/store/2.0.0-beta4"} 0.0
waku_connected_peers{direction="Out",protocol="/vac/waku/peer-exchange/2.0.0-alpha1"} 0.0
waku_connected_peers{direction="In",protocol="/vac/waku/peer-exchange/2.0.0-alpha1"} 0.0
waku_connected_peers{direction="In",protocol="/vac/waku/lightpush/2.0.0-beta1"} 0.0
waku_connected_peers{direction="In",protocol="/vac/waku/store-query/3.0.0"} 0.0
waku_connected_peers{direction="In",protocol="/vac/waku/metadata/1.0.0"} 0.0
waku_connected_peers{direction="Out",protocol="/vac/waku/filter-push/2.0.0-beta1"} 0.0
waku_connected_peers{direction="Out",protocol="/vac/waku/metadata/1.0.0"} 0.0
waku_connected_peers{direction="Out",protocol="/vac/waku/store-query/3.0.0"} 0.0
waku_connected_peers{direction="Out",protocol="/vac/waku/relay/2.0.0"} 22.0
```

Once this is available and works well, we must establish that `max-relay-peers` to 200 in `waku.*` fleets, and then leave ~100 for other protocols.

( cc thanks @alrevuelta for the idea! )

gabrielmer:
# Description
Pruning excess relay connections to respect our relay connection targets

## Issue

closes #2929

### DIFF
--- a/waku/node/peer_manager/peer_manager.nim
+++ b/waku/node/peer_manager/peer_manager.nim
@@ -734,9 +734,8 @@ proc connectToRelayPeers*(pm: PeerManager) {.async.} =
   let totalRelayPeers = inRelayPeers.len + outRelayPeers.len
   let inPeersTarget = maxConnections - pm.outRelayPeersTarget
 
-  # TODO: Temporally disabled. Might be causing connection issues
-  #if inRelayPeers.len > pm.inRelayPeersTarget:
-  #  await pm.pruneInRelayConns(inRelayPeers.len - pm.inRelayPeersTarget)
+  if inRelayPeers.len > pm.inRelayPeersTarget:
+    await pm.pruneInRelayConns(inRelayPeers.len - pm.inRelayPeersTarget)
 
   if outRelayPeers.len >= pm.outRelayPeersTarget:
     return


Ivansete-status (approved):
LGTM! Thanks for it! :100: 
As mentioned somewhere, we would need to check with DST and maybe manually deploy into `waku.test` or `waku.sandbox` and see how behaves the node.
The original purpose of the issue was to prevent `relay` occupying the whole libp2p available connections.

One possible option could be having 1/3 for `in`, 1/3 for `out`, and 1/3 for "others"

gabrielmer:
> One possible option could be having 1/3 for `in`, 1/3 for `out`, and 1/3 for "others"

So we have `max-relay-peers` and `max-connections` parameters.

I understand that if we configure `max-connections` to be a bigger number than `max-relay-peers` then it's enough. Not sure if we need to change something? I think that it's just a matter on the config parameters that we use to run a node.

> As mentioned somewhere, we would need to check with DST and maybe manually deploy into `waku.test` or `waku.sandbox` and see how behaves the node

I did test it in a normal node. What I was thinking is to merge it if tests pass, it will go through our regular regression testing and we will figure out if there's an issue. And if so, we open a specific issue for it and not leave it commented without any info. Although I think it shouldn't bring issues.

We can ask to go through explicit testing by DST before merging, but I think that in case this brings an issue, it will be thoroughly tested by all parties and we will know it before this actually gets released. Not sure if we should add this extra step or not.

LMK how you prefer to proceed :)

gabrielmer:
Oh nice, tests failed! So will start looking at that first :)

Turning the PR into draft

SionoiS (approved):

PR #2966, opened 2024-08-12
fix: faster retention policy when error and use of detach finalize when needed

Ivansete-status:
## Description
There are occasions where the concurrently-detach partition query didn't complete successfully. In those cases, the database shows an error hint recommending to perform a "finalize" detach partition. This PR aims at applying this approach when that error happens

## Changes

- [x] Apply detach FINALIZE when recommended by the database 
- [x] Speed up the retention policy application to once every minute, in case of error, so that we try again faster instead of waiting for 30 minutes each time. 

## Issue

closes https://github.com/waku-org/nwaku/issues/2890

### DIFF
--- a/waku/waku_archive/archive.nim
+++ b/waku/waku_archive/archive.nim
@@ -24,6 +24,7 @@ const
 
   # Retention policy
   WakuArchiveDefaultRetentionPolicyInterval* = chronos.minutes(30)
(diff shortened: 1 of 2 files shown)


gabrielmer (approved):
Thank you!

darshankabariya (approved):
LGTM !

PR #2970, opened 2024-08-14
fix: avoid using the msg key in chronicles

gabrielmer:
# Description

As reported on https://github.com/vacp2p/nim-libp2p/issues/1176, using `msg` as a key in `chronicles` causes duplicate keys when parsing the logs, as the main log message is also saved under the `msg` key.

Fixing the instances of this bug in nwaku


Ivansete-status (approved):
LGTM! Thanks for it! :100:

darshankabariya (approved):
LGTM !

PR #2973, opened 2024-08-14
chore: logging received message info via onValidated observer

gabrielmer:
# Description
Logging information related to received messages by using one of the new `onValidated` `nim-libp2p` observers. This is improves our current approach of logging via the `traceHandler`, which does not provide information on who sent the message a node received.

# Changes

- [x] setting `onValidated` logger observer
- [x] removed message received logging from `traceHandler`
- [x] added message received log when a message is sent via REST, as per DST request
- [x] renamed `initRelayMetricObserver` to `initRelayObservers` 

Notice that a "message received" log were added to the REST endpoints that publish messages. This is per DST's request, so they can easily tag the node that published a message to the network as a node that has seen the message.

CC @AlbertoSoutullo 

### Example

image

### DIFF
--- a/waku/waku_relay/protocol.nim
+++ b/waku/waku_relay/protocol.nim
@@ -151,7 +151,36 @@ proc initProtocolHandler(w: WakuRelay) =
   w.handler = handler
   w.codec = WakuRelayCodec
 
-proc initRelayMetricObserver(w: WakuRelay) =
+proc logMessageInfo*(
+    w: WakuRelay,
+    peerId: string,
+    topic: string,
+    msg_id_short: string,
+    msg: WakuMessage,
+    onRecv: bool,
+) =
+  let msg_hash = computeMessageHash(topic, msg).to0xHex()
+
+  if onRecv:
+    notice "received relay message",
+      my_peer_id = w.switch.peerInfo.peerId,
+      msg_hash = msg_hash,
+      msg_id = msg_id_short,
+      from_peer_id = peerId,
+      topic = topic,
+      receivedTime = getNowInNanosecondTime(),
+      payloadSizeBytes = msg.payload.len
+  else:
+    notice "sent relay message",
+      my_peer_id = w.switch.peerInfo.peerId,
+      msg_hash = msg_hash,
+      msg_id = msg_id_short,
+      to_peer_id = peerId,
+      topic = topic,
+      sentTime = getNowInNanosecondTime(),
+      payloadSizeBytes = msg.payload.len
+
+proc initRelayObservers(w: WakuRelay) =
   proc decodeRpcMessageInfo(
       peer: PubSubPeer, msg: Message
   ): Result[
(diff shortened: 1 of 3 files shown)


AlbertoSoutullo:
To add a bit of context regarding ` added message received log when a message is sent via REST`:
In order to check message reliability via logging, the problem we were facing when messages were being injected in the network via REST is that all nodes were logging the received message but the node that started the publishing. So we wanted a way to make sure that at least this node also "received" this message but from a different way, so we could make sure this happened, instead of assuming it, if it makes sense.

waku/waku_relay/protocol.nim
```diff
@@ -151,7 +151,36 @@ proc initProtocolHandler(w: WakuRelay) =
   w.handler = handler
   w.codec = WakuRelayCodec
 
-proc initRelayMetricObserver(w: WakuRelay) =
+proc logMessageInfo*(
+    w: WakuRelay,
+    peerId: string,
```
Ivansete-status, line 156: nitpick one ;P
```suggestion
    remotePeerId: string,
```

Ivansete-status (approved):
Gorgeous one! Thanks so much for it! :100:

waku/waku_relay/protocol.nim
gabrielmer: Niiice! Done!

waku/waku_relay/protocol.nim
```diff
@@ -208,18 +223,38 @@ proc initRelayMetricObserver(w: WakuRelay) =
...
+
+    logMessageInfo(
+      w, shortLog(peer.peerId), msg.topic, msg_id_short, wakuMessage, onRecv = true
+    )
+
   proc onSend(peer: PubSubPeer, msgs: var RPCMsg) =
     for msg in msgs.messages:
       let (msg_id_short, topic, wakuMessage, msgSize) = decodeRpcMessageInfo(peer, msg).valueOr:
+        warn "onSend: failed decoding RPC info",
+          my_peer_id = w.switch.peerInfo.peerId, to_peer_id = peer.peerId
```
SionoiS, line 249: not_idiomatic_no_? Here and elsewhere.

SionoiS (approved):
LGTM thanks!

waku/waku_relay/protocol.nim
gabrielmer: No strong opinions, as it's mostly about how we log things, but I agree that would be nice to have everything unified in a single format.

On the other hand, DST already build its tools based on those keys, which are the ones we have been using for long.

@Ivansete-status opinions?

Ivansete-status: @gabrielmer - it looks correct to my eyes
@SionoiS - do you mean the use of snake case? I think we tend to use that but if not the case I'm happy to unify

gabrielmer: I'll proceed and merge it this way so it enters the release. If we decide to change this, I'll open a separate PR for it :)

Issue #2806, opened 2024-06-12
chore: remove --pubsub-topic configuration option


gabrielmer:
### Background
Following the deprecation of named sharding, we will configure a node by setting its `cluster-id` and `shards`.
Therefore, to avoid redundancy, we will remove the `--pubsub-topic` configuration option.

### Details
We need to remove the option, and update all the dependent places that use it such as infra, DST and nwaku-compose

### Acceptance criteria
- [ ] `--pubsub-topic` configuration is removed
- [ ] dependent places get updated to use `cluster-id` and `shards` instead

PR #2974, opened 2024-08-15
chore: rename NsPubsubTopic

gabrielmer:
# Description

Now that only static sharding is supported and we want to use a type for clusterId + shardId, we're renaming the `NsPubsubTopic` type to `RelayShard` for it to have a more meaningful name and use it throughout the codebase.

# Changes

- [x] renamed `NsPubsubTopic` to `RelayShard`

## Issue

Advances #2806


waku/waku_core/topics/pubsub_topic.nim
```diff
@@ -67,19 +67,19 @@ proc parseStaticSharding*(
...
 # Pubsub topic compatibility
 
-converter toPubsubTopic*(topic: NsPubsubTopic): PubsubTopic =
+converter toPubsubTopic*(topic: RelayShard): PubsubTopic =
   $topic
 
-proc `==`*[T: NsPubsubTopic](x, y: T): bool =
+proc `==`*[T: RelayShard](x, y: T): bool =
   if x.clusterId != y.clusterId:
     return false
```
SionoiS, line 84: That would only be partial equality. What happen is shard don't match?

SionoiS (changes requested):
LGTM expect the equality bit.

waku/waku_core/topics/pubsub_topic.nim
gabrielmer: I think the other part of the proc was not properly displayed

image


I think it's ok like that right?

SionoiS: Ah! NVM then.!

SionoiS (approved):

Ivansete-status (approved):
LGTM! Thanks for it! :100:

PR #2986, opened 2024-08-19
chore: ANALYZE messages query should be performed regularly

Ivansete-status:
## Description
In a pair session with @richard-ramos, thanks for it :partying_face: :raised_hands: !, we've discovered that the database wasn't using the proper indexes for certain query scenarios.

After applying `ANALYZE messages` to a database we've noticed that the query execution time went from ~1second to ~100ms.

Therefore, we considered that it is interesting to `ANALYZE messages` regularly in the database so that the stats are properly updated and the planner can pick up the optimum indexes for each case.

## Changes

- Run `ANALYZE messages` once a day.

## More context
That was the query under analysis:
```code
SELECT contentTopic, payload, pubsubTopic, version, timestamp, id, messageHash, meta FROM messages WHERE contentTopic IN ('/waku/1/0x9a3fcc34/rfc26','/waku/1/0xc9b649fe/rfc26','/waku/1/0x678be1ba/rfc26','/waku/1/0x87c713d2/rfc26','/waku/1/0x267ad6e5/rfc26','/waku/1/0xb50f4cd5/rfc26','/waku/1/0xe2509252/rfc26','/waku/1/0x8de76c97/rfc26') AND pubsubTopic = '/waku/2/rs/16/32' AND (timestamp, id) < ('1724046108468541771','ed019e15afef1273dabd74e9cc4091fad4a5f0fa054230a8e121f81fcec278c7') AND timestamp >= '1723971035000000000' AND timestamp <= '1724057435000000000' ORDER BY timestamp DESC, id DESC LIMIT '21';
```

### DIFF
--- a/waku/waku_archive/driver/builder.nim
+++ b/waku/waku_archive/driver/builder.nim
@@ -102,6 +102,8 @@ proc new*(
       ## Hence, this should be run after the migration is completed.
       asyncSpawn driver.startPartitionFactory(onFatalErrorAction)
 
+      driver.startAnalyzeTableLoop()
+
       info "waiting for a partition to be created"
       for i in 0 ..< 100:
         if driver.containsAnyPartition():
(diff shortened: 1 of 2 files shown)


richard-ramos (approved):
Looking fwd to see this in action on the fleets!

gabrielmer (approved):
Amazing! Thanks so much!

PR #2987, opened 2024-08-19
chore: libwaku retrieve my enr and adapt golang example

Ivansete-status:
## Description
Add new "get ENR" method available in libwaku

## Changes

- new `waku_get_my_enr` available in `libwaku`
- adapt Golang example to use the new `waku_get_my_enr` function

### DIFF
--- a/examples/golang/waku.go
+++ b/examples/golang/waku.go
@@ -457,8 +457,8 @@ func (self *WakuNode) WakuGetMyENR() (string, error) {
 	C.cGoWakuGetMyENR(self.ctx, resp)
 
 	if C.getRet(resp) == C.RET_OK {
-		var listenAddresses = C.GoStringN(C.getMyCharPtr(resp), C.int(C.getMyCharLen(resp)))
-		return listenAddresses, nil
+		var myENR = C.GoStringN(C.getMyCharPtr(resp), C.int(C.getMyCharLen(resp)))
+		return myENR, nil
 	}
 	errMsg := "error WakuGetMyENR: " +
 		C.GoStringN(C.getMyCharPtr(resp), C.int(C.getMyCharLen(resp)))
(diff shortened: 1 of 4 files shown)


examples/golang/waku.go
```diff
@@ -446,6 +451,20 @@ func (self *WakuNode) WakuListenAddresses() (string, error) {
 	return "", errors.New(errMsg)
 }
 
+func (self *WakuNode) WakuGetMyENR() (string, error) {
+	var resp = C.allocResp()
+	defer C.freeResp(resp)
+	C.cGoWakuGetMyENR(self.ctx, resp)
+
+	if C.getRet(resp) == C.RET_OK {
+		var listenAddresses = C.GoStringN(C.getMyCharPtr(resp), C.int(C.getMyCharLen(resp)))
```
NagyZoltanPeter, line 460: Maybe a bit misleading to call it listenAddresses and not the ENR.
I remember that is different also in /debug/info API result.

NagyZoltanPeter (approved):
LGTM, added one little naming observation...

examples/golang/waku.go
```diff
@@ -446,6 +451,20 @@ func (self *WakuNode) WakuListenAddresses() (string, error) {
...
+		return listenAddresses, nil
```
gabrielmer, line 461: Shouldn't we change the variable name to `enr` or something like that? looks like we kept the same variable name after copy-pasting from `WakuListenAddresses()`

gabrielmer (approved):
LGTM, thank you! Added just a tiny question

examples/golang/waku.go
gabrielmer: ooh @NagyZoltanPeter added the same comment a few minutes before I did lol. Nice!

Ivansete-status: Thank you so much guys ! I'll change that

PR #2991, opened 2024-08-21
fix: copy libnegentropy.so from nim-build image

jakubgs:
We shouldn't assume it exists on the host.

Introduced by:
- https://github.com/waku-org/nwaku/pull/2403


Ivansete-status (approved):

gabrielmer (approved):
Thank you!

PR #2994, opened 2024-08-22
chore: Better timing and requestId detail for slower store db queries 

Ivansete-status:
## Description
Better track the `requestId` that comes from `store` requests to lower layers (archive/database) so that we can link requests with slow queries

## Changes
- Pass down the `requestId` value from store layer to the database layer.
- `pgasyncpool.nim`: Print information for queries slower than 2 seconds.
- `client.nim`: allow sending custom `requestId`.
- Adapt tests a little because now some tests checks the returned `requested` vs the sent `requestId` and they failed because we are adding the `requestId` field in the `ArchiveQuery` object.

### DIFF
--- a/waku/waku_archive_legacy/driver.nim
+++ b/waku/waku_archive_legacy/driver.nim
@@ -41,6 +41,7 @@ method getMessagesV2*(
     endTime = none(Timestamp),
     maxPageSize = DefaultPageSize,
     ascendingOrder = true,
+    requestId: string,
 ): Future[ArchiveDriverResult[seq[ArchiveRow]]] {.base, deprecated, async.} =
   discard
 
@@ -55,6 +56,7 @@ method getMessages*(
     hashes = newSeq[WakuMessageHash](0),
     maxPageSize = DefaultPageSize,
     ascendingOrder = true,
+    requestId = "",
 ): Future[ArchiveDriverResult[seq[ArchiveRow]]] {.base, async.} =
   discard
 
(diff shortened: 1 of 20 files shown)


gabrielmer (approved):
Amazing addition! Thanks so much!

waku/waku_archive_legacy/driver.nim
```diff
@@ -41,6 +41,7 @@ method getMessagesV2*(
     endTime = none(Timestamp),
     maxPageSize = DefaultPageSize,
     ascendingOrder = true,
+    requestId: string,
```
NagyZoltanPeter, line 44: If requestId is mandatory here, would it be better at the front of the args?
Maybe it is just my old habit, I know it works in nim though...

waku/waku_archive_legacy/driver.nim
```diff
@@ -55,6 +56,7 @@ method getMessages*(
     hashes = newSeq[WakuMessageHash](0),
     maxPageSize = DefaultPageSize,
     ascendingOrder = true,
+    requestId = "",
```
NagyZoltanPeter, line 59: Why isn't that mandatory to set here if it is with V2 method?

NagyZoltanPeter (approved):
LGTM, just some questions I made.

waku/waku_archive_legacy/driver.nim
Ivansete-status: > If requestId is mandatory here, would it be better at the front of the args? Maybe it is just my old habit, I know it works in nim though...

Ah yes you are right! Nevertheless let me take the license to leave it as is given that we will deprecate this code soon. Thanks for the comment!

waku/waku_archive_legacy/driver.nim
Ivansete-status: > Why isn't that mandatory to set here if it is with V2 method?

Yes in this case I set it a default value to prevent changing too many tests that we will remove shortly.

PR #2996, opened 2024-08-23
fix: libnegentropy integration

gabrielmer:
# Description

When trying to deploy the newer versions to the fleets, the Jenkins job used to fail because of not finding `libnegentropy.so`

I found that in the CI environment, it only finds `./libnegentropy.so` and in the Jenkins environment it only finds `libnegentropy.so`

# Changes

- [x] Refactoring Makefile
- [x] Using `./libnegentropy.so` or `libnegentropy.so` depending on if the file is found 
- [x] In the Dockerfiles, copying `libnegentropy.so` to `/usr/lib` instead of using relative paths


Ivansete-status (approved):
LGTM! Thanks for it! :100: 
I think only required the `/usr/bin` change but approving to not block

gabrielmer:
> LGTM! Thanks for it! 💯 I think only required the `/usr/bin` change but approving to not block

Thank you! I initially tried with only the`/usr/lib` change but it did not work (https://github.com/waku-org/nwaku/pull/2996/commits/41fac669383a4f805d3ae3cb7310b6e2030ea66c).

Maybe not all changes in the Makefile are necessary, I tried to mimic how `librln` is integrated to be consistent and because that works. But for sure there's room for improvement

NagyZoltanPeter (approved):

darshankabariya (approved):
LGTM !

PR #2997, opened 2024-08-26
chore: deprecating pubsub topic

### LINKED ISSUE #2806: chore: remove --pubsub-topic configuration option

gabrielmer:
# Description
Deprecating the `pubsub-topic` CLI config in favor of `cluster-id`, `shards` and a new config item `network-shards` stating the amount of shards in the given `cluster-id` for autosharding purposes

The word `pubsubTopic` in our codebase will refer solely to the string representation of `clusterId`+`shard` that we use in the `libp2p` layer.

# Changes

- [x] marking pubsub topic configuration as deprecated
- [x] converting pubsub topics to shards for compatibility during the deprecation period
- [x] using `RelayShard` type instead of pubsub topic strings in Waku's higher layers
- [x] Introducing `network-shards` configuration item for autosharding
- [x] Using the max shard number + 1 as `network-shards` if not configured

## Issue

closes #2806

### DIFF
--- a/tests/testlib/wakunode.nim
+++ b/tests/testlib/wakunode.nim
@@ -38,7 +38,7 @@ proc defaultTestWakuNodeConf*(): WakuNodeConf =
     maxConnections: 50,
     maxMessageSize: "1024 KiB",
     clusterId: 0,
-    pubsubTopics: @["/waku/2/rs/0/0"],
+    shards: @[uint16(0)],
     relay: true,
     storeMessageDbUrl: "sqlite://store.sqlite3",
   )
@@ -63,8 +63,9 @@ proc newTestWakuNode*(
     dns4DomainName = none(string),
     discv5UdpPort = none(Port),
     agentString = none(string),
-    pubsubTopics: seq[string] = @["/waku/2/rs/1/0"],
     peerStoreCapacity = none(int),
+    clusterId = DefaultClusterId,
+    shards = @[DefaultShardId],
 ): WakuNode =
   var resolvedExtIp = extIp
 
@@ -77,14 +78,8 @@ proc newTestWakuNode*(
 
   var conf = defaultTestWakuNodeConf()
 
-  let clusterId =
-    if pubsubTopics.len() > 0:
-      RelayShard.parse(pubsubTopics[0]).get().clusterId
-    else:
-      1.uint16
-
   conf.clusterId = clusterId
-  conf.pubsubTopics = pubsubTopics
+  conf.shards = shards
 
   if dns4DomainName.isSome() and extIp.isNone():
     # If there's an error resolving the IP, an exception is thrown and test fails
@@ -95,7 +90,7 @@ proc newTestWakuNode*(
 
   let netConf = NetConfig.init(
     bindIp = bindIp,
-    clusterId = clusterId,
+    clusterId = conf.clusterId,
     bindPort = bindPort,
     extIp = resolvedExtIp,
     extPort = extPort,
@@ -111,8 +106,10 @@ proc newTestWakuNode*(
 
   var enrBuilder = EnrBuilder.init(nodeKey)
 
-  enrBuilder.withShardedTopics(pubsubTopics).isOkOr:
-    raise newException(Defect, "Invalid record: " & error)
+  enrBuilder.withWakuRelaySharding(
+    RelayShards(clusterId: conf.clusterId, shardIds: conf.shards)
+  ).isOkOr:
+    raise newException(Defect, "Invalid record: " & $error)
 
   enrBuilder.withIpAddressAndPorts(
     ipAddr = netConf.enrIp, tcpPort = netConf.enrPort, udpPort = netConf.discv5UdpPort
--- a/tests/testlib/wakunode.nim
+++ b/tests/testlib/wakunode.nim
@@ -37,8 +37,8 @@ proc defaultTestWakuNodeConf*(): WakuNodeConf =
     nat: "any",
     maxConnections: 50,
     maxMessageSize: "1024 KiB",
-    clusterId: 0,
-    shards: @[uint16(0)],
+    clusterId: DefaultClusterId,
+    shards: @[DefaultShardId],
     relay: true,
     storeMessageDbUrl: "sqlite://store.sqlite3",
   )
--- a/waku/factory/external_config.nim
+++ b/waku/factory/external_config.nim
@@ -312,8 +312,11 @@ type WakuNodeConf* = object
       name: "keep-alive"
     .}: bool
 
-    networkShards* {.
-      desc: "Number of shards in the network", defaultValue: 0, name: "network-shards"
+    # If numShardsInNetwork is not set, we use the number of shards configured as numShardsInNetwork
+    numShardsInNetwork* {.
+      desc: "Number of shards in the network",
+      defaultValue: 0,
+      name: "num-shards-in-network"
     .}: uint32
 
     pubsubTopics* {.
--- a/waku/factory/external_config.nim
+++ b/waku/factory/external_config.nim
@@ -312,8 +312,9 @@ type WakuNodeConf* = object
       name: "keep-alive"
     .}: bool
 
-    networkShards* {.desc: "Number of shards in the network", name: "network-shards".}:
-      uint32
+    networkShards* {.
+      desc: "Number of shards in the network", defaultValue: 0, name: "network-shards"
+    .}: uint32
 
     pubsubTopics* {.
       desc:
--- a/waku/factory/external_config.nim
+++ b/waku/factory/external_config.nim
@@ -315,6 +315,13 @@ type WakuNodeConf* = object
     networkShards* {.desc: "Number of shards in the network", name: "network-shards".}:
       uint32
 
+    pubsubTopics* {.
+      desc:
+        "Deprecated. Default pubsub topic to subscribe to. Argument may be repeated.",
+      defaultValue: @[],
+      name: "pubsub-topic"
+    .}: seq[string]
+
     shards* {.
       desc: "Shards index to subscribe to [0..MAX_SHARDS-1]. Argument may be repeated.",
       defaultValue:
--- a/waku/factory/external_config.nim
+++ b/waku/factory/external_config.nim
@@ -312,11 +312,8 @@ type WakuNodeConf* = object
       name: "keep-alive"
     .}: bool
 
-    pubsubTopics* {.
-      desc: "Default pubsub topic to subscribe to. Argument may be repeated.",
-      defaultValue: @[],
-      name: "pubsub-topic"
-    .}: seq[string]
+    networkShards* {.desc: "Number of shards in the network", name: "network-shards".}:
+      uint32
 
     shards* {.
       desc: "Shards index to subscribe to [0..MAX_SHARDS-1]. Argument may be repeated.",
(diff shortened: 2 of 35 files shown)


gabrielmer:
Didn't change the terminology in the `liteprotocoltester` to avoid breaking compatibility in current tests. Should we change there too all references to pubsub topic to cluster and shard? @NagyZoltanPeter

NagyZoltanPeter:
> Didn't change the terminology in the `liteprotocoltester` to avoid breaking compatibility in current tests. Should we change there too all references to pubsub topic to cluster and shard? @NagyZoltanPeter 


If that ok I will pull after next week!

waku/factory/external_config.nim
```diff
@@ -312,8 +312,13 @@ type WakuNodeConf* = object
       name: "keep-alive"
     .}: bool
 
+    networkShards* {.
+      desc: "Number of shards in the network", defaultValue: 0, name: "network-shards"
+    .}: uint32
```
SionoiS, line 317: :thinking: maybe we should be explicit that it pertains to autosharding?

What about `autoshardingRange` or `autoshardingsValues` ?

waku/factory/node_factory.nim
```diff
@@ -113,6 +113,12 @@ proc initNode(
 
 ## Mount protocols
 
+proc getNetworkShards*(conf: WakuNodeConf): uint32 =
+  if conf.networkShards != 0:
+    return conf.networkShards
+  # If conf.networkShards is not set, use the number of shards configured as networkShards
+  return uint32(max(conf.shards) + 1)
```
SionoiS, line 120: I'm not sure that would be intuitive for the user. I would suggest requiring this value if using custom network config.

SionoiS (approved):
LGTM Nice work!

Left some suggestions.

tests/testlib/wakunode.nim
```diff
@@ -38,7 +38,7 @@ proc defaultTestWakuNodeConf*(): WakuNodeConf =
     maxConnections: 50,
     maxMessageSize: "1024 KiB",
     clusterId: 0,
-    pubsubTopics: @["/waku/2/rs/0/0"],
+    shards: @[uint16(0)],
```
Ivansete-status, line 41: Maybe :)?
```suggestion
    shards: @[DefaultShardId],
```

waku/factory/external_config.nim
```diff
@@ -312,8 +312,13 @@ type WakuNodeConf* = object
...
+
     pubsubTopics* {.
-      desc: "Default pubsub topic to subscribe to. Argument may be repeated.",
+      desc:
+        "Deprecated. Default pubsub topic to subscribe to. Argument may be repeated.",
```
Ivansete-status, line 321: Let's create a release milestone and take note of the final deprecation. Maybe in `v0.34.0`?

waku/factory/node_factory.nim
```diff
@@ -113,6 +113,12 @@ proc initNode(
 
 ## Mount protocols
 
+proc getNetworkShards*(conf: WakuNodeConf): uint32 =
+  if conf.networkShards != 0:
+    return conf.networkShards
+  # If conf.networkShards is not set, use the number of shards configured as networkShards
```
Ivansete-status, line 119: Maybe we should also comment that in the `external_config.nim` parameter so that it is clear that the `networkShards` is optional.

waku/factory/external_config.nim
Ivansete-status, line 317: Nitpick one :) Maybe a bit more explicit
```suggestion
    numShardsInNetwork* {.
      desc: "Number of shards in the network", defaultValue: 0, name: "num-shards-in-network"
    .}: uint32
```

waku/factory/node_factory.nim
```diff
@@ -127,7 +133,10 @@ proc setupProtocols(
   node.mountMetadata(conf.clusterId).isOkOr:
     return err("failed to mount waku metadata protocol: " & error)
 
-  node.mountSharding(conf.clusterId, uint32(conf.pubsubTopics.len)).isOkOr:
+    # If conf.networkShards is not set, use the number of shards configured as networkShards
```
Ivansete-status, line 136:
```suggestion
    # If conf.networkShards is not set, use the number of shards configured as numShardsInNetwork
```

waku/factory/waku.nim
```diff
@@ -86,6 +86,18 @@ proc logConfig(conf: WakuNodeConf) =
 func version*(waku: Waku): string =
   waku.version
 
+proc validateShards(conf: WakuNodeConf): Result[void, string] =
+  let networkShards = getNetworkShards(conf)
+
+  for shard in conf.shards:
+    if shard >= networkShards:
+      let msg = "Invalid shard: " & $shard & " when networkShards: " & $networkShards
```
Ivansete-status, line 94:
```suggestion
      let msg = "validateShards invalid shard: " & $shard & " when networkShards: " & $networkShards
```

waku/factory/waku.nim
```diff
@@ -94,16 +106,33 @@ proc init*(T: type Waku, conf: WakuNodeConf): Result[Waku, string] =
...
+        error = shardsRes.error
+      return err("failed to parse pubsub topic: " & $shardsRes.error)
+
+    let shardsOpt = shardsRes.get()
+
+    if shardsOpt.isSome():
+      let relayShards = shardsOpt.get()
+      if relayShards.clusterId != conf.clusterId:
+        error "clusterId of the pubsub topic should match the node's cluster",
+          nodeCluster = conf.clusterId, pubsubCluster = relayShards.clusterId
```
Ivansete-status, line 124: I think is interesting to add a small example although having a bit long line
```suggestion
        error "clusterId of the pubsub topic should match the node's cluster. e.g. --pubsub-topic=/waku/2/rs/22/1 and --cluster-id=22",
          nodeCluster = conf.clusterId, pubsubCluster = relayShards.clusterId
```

Ivansete-status (approved):
LGTM! Thanks for it! :100: 
Just added a few minor comments that I hope you find useful

waku/factory/external_config.nim
gabrielmer: Good question, not sure what should be the best name for it. I wouldn't personally use the word autosharding as even though we use this value for autosharding, it represents a property of the network. But open for opinions :))

waku/factory/node_factory.nim
gabrielmer: We currently use the length of pubsub topics configured. I changed this pattern because if we configure shards [0, 3] and we have 2 network shards it wouldn't make sense, so I went for the max + 1.

I agree it might not be intuitive, but I was afraid of requiring that parameter and making it a breaking change. If we require it, all the tests and uses in shards different than 1 would be broken by this new requirement.

@alrevuelta what's your take on this?

Maybe a good alternative would be to log a warning when we set that number for them if they didn't

tests/testlib/wakunode.nim
gabrielmer: Ooh great one! Will also do the same for the clusterId :)

waku/factory/external_config.nim
gabrielmer: We can make it more explicit that it refers to the number, yes :) Committed the fix

waku/factory/external_config.nim
gabrielmer: We have https://github.com/waku-org/nwaku/issues/2806 open, just changed the milestone to `v0.34.0`

Maybe we shouldn't close this issue after this PR? or create a new one? Any of them works for me

waku/factory/node_factory.nim
SionoiS: I guess I don't like that kind of magic. In this case it's fine the stakes are low anyway.

gabrielmer: I agree 100% that it's better to avoid magic. But as you said, stakes are low and it doesn't break things :)

waku/factory/node_factory.nim
```diff
@@ -113,6 +113,12 @@ proc initNode(
 
 ## Mount protocols
 
+proc getNumShardsInNetwork*(conf: WakuNodeConf): uint32 =
+  if conf.numShardsInNetwork != 0:
+    return conf.numShardsInNetwork
+  # If conf.numShardsInNetwork is not set, use the number of shards configured as numShardsInNetwork
+  return uint32(max(conf.shards) + 1)
```
chaitanyaprem, line 120: is it safe to assume that number of shards in the network is the configured shards count (in case static sharding is being used)? 
what if i am running a node which only supports few of the shards that are there in the network in case of static sharding. will this assumption break any functionality?

chaitanyaprem (commented):
Nice, as config was confusing otherwise

jm-clius (approved):
Generally agree with this direction. The concept of pubsub topic remains useful within the wire protocols itself as it matches more closely what is modelled in libp2p.

waku/factory/node_factory.nim
gabrielmer: We initially had that the number of shards configured was the shard count, but we got to edge cases in which if someone subscribed only to shards 4 and 5 for example, we would get that it subscribed to shards 4 and 5 in a network with 2 shards and made no sense.

So I modified it to use the highest number of a shard configured + 1 as number of shards in the network. So if someone subscribes only to shards 4 and 5 and doesn't set manually the number of shards in the network, we deduce that there's 6 shards in the network (0 to 5)

And with this other approach, functionality isn't broken :)

chaitanyaprem: > So I modified it to use the highest number of a shard configured + 1 as number of shards in the network. So if someone subscribes only to shards 4 and 5 and doesn't set manually the number of shards in the network, we deduce that there's 6 shards in the network (0 to 5)

But that seems like a wrong assumption that could lead to issues. Think of status node , it subscribes only to shards 1,32,64,128,256 as of now. But number of shards in the network is all available 1024 as per the static sharding indicated in the [spec](https://github.com/waku-org/specs/blob/master/standards/core/relay-sharding.md#static-sharding).
This assumption may not have an impact as of now, but was wondering if it will cause future issues. Instead if number of shards is not configured, why not consider it as 1024 since all 1024 can be used.

gabrielmer: > Instead if number of shards is not configured, why not consider it as 1024 since all 1024 can be used.

Makes sense! Yes, I actually think it's a better solution and will lead to more nodes "agreeing" on the number of shards in a network, as it doesn't depend on their shard subscriptions.

Ideally, nodes should configure manually the number of shards in the network they use. But if static sharding specification limits it to 1024, then it makes sense to use 1024 as a default.

Will update the PR :) Thanks so much!

chaitanyaprem: > Ideally, nodes should configure manually the number of shards in the network they use. But if static sharding specification limits it to 1024, then it makes sense to use 1024 as a default.

Yep, but that is only in case of auto-sharding. In case of static sharding, nodes don't specify number of shards which is why 1024 made sense.

PR #3002, opened 2024-08-28
chore: Added metrics to liteprotocoltester

NagyZoltanPeter:
# Description
Added a dashboard to monitor testing performance from grafana dashboard

# Changes

- [X] Added similar metrics as in collected statistics
- [X] Fix negentropy lib reference in docker
- [X] Added prometheus and grafana into docker compose
- [X] Added liteprotocoltester dashboard

## How to test

Check liteprotocoltester readme. 
Run with waku-simulator as in the description
go to http://localhost:3033

image

## Issue
https://github.com/waku-org/nwaku/issues/2999


apps/liteprotocoltester/.env
```diff
@@ -4,7 +4,7 @@ START_PUBLISHING_AFTER=10
 NUM_MESSAGES=0
 # 0 for infinite number of messages
 
-DELAY_MESSAGES=8000
+DELAY_MESSAGES=188
```
Ivansete-status, line 7: I'm curious as to why this figure :D

apps/liteprotocoltester/statistics.nim
```diff
@@ -103,8 +114,17 @@ proc addMessage*(
   if not self.contains(peerId):
     self[peerId] = Statistics.init()
 
+  let shortSenderId = block:
+    let senderPeer = PeerId.init(msg.sender)
+    if senderPeer.isErr:
```
Ivansete-status, line 119: Super nitpick comment :)
```suggestion
    if senderPeer.isErr():
```

Ivansete-status (approved):
LGTM! Thanks for it! :100:

apps/liteprotocoltester/.env
NagyZoltanPeter: Ah, I see... it remained from my last waku-sim run.
While 8000ms is set for TWN... not to harm rln limit... on waku-sim I needed to stress the network... hence the much lower value.
I'd better leave the (my) .env file here for an example.

gabrielmer (approved):
LGTM, thanks so much! 🔥

PR #3003, opened 2024-08-28
chore: libwaku - exposing discv5 store-client and relay connected and mesh peers

Ivansete-status:
# Description
Step further to make nwaku completely accessible from libwaku. Still pending to expose some features such as peer exchange, or filter.

# Changes

- Allow to start or stop discv5
- Expose lightpush request operation
- Expose list of connected and mesh peers
- Expose store client

### DIFF
--- a/library/libwaku.nim
+++ b/library/libwaku.nim
@@ -19,6 +19,7 @@ import
   ./waku_thread/inter_thread_communication/requests/peer_manager_request,
   ./waku_thread/inter_thread_communication/requests/protocols/relay_request,
   ./waku_thread/inter_thread_communication/requests/protocols/store_request,
+  ./waku_thread/inter_thread_communication/requests/protocols/lightpush_request,
   ./waku_thread/inter_thread_communication/requests/debug_node_request,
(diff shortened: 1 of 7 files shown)


gabrielmer (approved):
LGTM, thanks so much!

library/libwaku.nim
```diff
@@ -383,6 +384,119 @@ proc waku_relay_unsubscribe(
...
+    ctx: ptr Context,
+    pubSubTopic: cstring,
+    jsonWakuMessage: cstring,
+    callback: WakuCallBack,
+    userData: pointer,
+): cint {.dynlib, exportc, cdecl.} =
+  ctx[].userData = userData
+
+  if isNil(callback):
+    return RET_MISSING_CALLBACK
```
SionoiS, line 455: Why is this needed here but not in all other procs?

SionoiS (approved):
LGTM Thanks!

library/libwaku.nim
Ivansete-status: > Why is this needed here but not in all other procs?

Good point! Because some checks are left. I will submit an enhancement in a separate PR

PR #3010, opened 2024-08-30
fix: get back health check for postgres legacy

Ivansete-status:
### Description
Recovering the health check async task for the legacy driver.
This is needed to be kept because the inserts to the database are made through the legacy store Postgres driver.
What happened is that, if the database was restarted, then the legacy Postgres driver didn't reset the connections pool and therefore, the upcoming inserts failed.

### Issue
`chore: store node loses connection if database restarts`
closes https://github.com/waku-org/nwaku/issues/2989

### DIFF
--- a/waku/waku_archive_legacy/driver/postgres_driver/postgres_healthcheck.nim
+++ b/waku/waku_archive_legacy/driver/postgres_driver/postgres_healthcheck.nim
@@ -1,0 +1,38 @@
+{.push raises: [].}
+
+import chronos, chronicles, results
+import ../../../common/databases/db_postgres, ../../../common/error_handling
+
+## Simple query to validate that the postgres is working and attending requests
+const HealthCheckQuery = "SELECT version();"
+const CheckConnectivityInterval = 60.seconds
+const MaxNumTrials = 20
+const TrialInterval = 1.seconds
+
+proc checkConnectivity*(
+    connPool: PgAsyncPool, onFatalErrorAction: OnFatalErrorHandler
+) {.async.} =
+  while true:
+    (await connPool.pgQuery(HealthCheckQuery)).isOkOr:
+      ## The connection failed once. Let's try reconnecting for a while.
+      ## Notice that the 'exec' proc tries to establish a new connection.
+
+      block errorBlock:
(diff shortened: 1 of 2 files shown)


waku/waku_archive_legacy/driver/postgres_driver/postgres_healthcheck.nim
```diff
@@ -0,0 +1,38 @@
...
+const MaxNumTrials = 20
+const TrialInterval = 1.seconds
+
+proc checkConnectivity*(
+    connPool: PgAsyncPool, onFatalErrorAction: OnFatalErrorHandler
+) {.async.} =
+  while true:
+    (await connPool.pgQuery(HealthCheckQuery)).isOkOr:
+      ## The connection failed once. Let's try reconnecting for a while.
+      ## Notice that the 'exec' proc tries to establish a new connection.
```
SionoiS, line 18: I don't understand this comment. What `exec`?

waku/waku_archive_legacy/driver/postgres_driver/postgres_healthcheck.nim
```diff
@@ -0,0 +1,38 @@
...
+        ## Force close all the opened connections. No need to close gracefully.
+        (await connPool.resetConnPool()).isOkOr:
+          onFatalErrorAction("checkConnectivity legacy resetConnPool error: " & error)
+
+        var numTrial = 0
+        while numTrial < MaxNumTrials:
+          let res = await connPool.pgQuery(HealthCheckQuery)
+          if res.isOk():
+            ## Connection resumed. Let's go back to the normal healthcheck.
+            break errorBlock
```
SionoiS, line 30: I didn't know you could break named block. Very nice!

SionoiS (approved):
LGTM Thanks!

waku/waku_archive_legacy/driver/postgres_driver/postgres_healthcheck.nim
Ivansete-status: > I don't understand this comment. What `exec`?

Good catch! I updated the comment. Thanks!

gabrielmer (approved):
LGTM, thanks so much!

PR #3011, opened 2024-09-02
chore: update prepare_release.md

Ivansete-status:
add a new step to review https://github.com/waku-org/docs.waku.org/blob/develop/docs/guides/nwaku/config-options.md


SionoiS (approved):
Good idea!

PR #3020, opened 2024-09-05
chore: lightpush - error metric less variable by only setting a fixed string

Ivansete-status:
## Description
Avoid having a too-variable metric by preventing adding the msg_hash on it. In that case, we are just adding a fixed string, _not_published_to_any_peer_, when the condition "_Lightpush request has not been published to any peers_" is met.

Thanks for the detailed description @jakubgs :raised_hands: 

## Issue

bug: absurdly high cardinality of waku_lightpush_errors metrics
closes https://github.com/waku-org/nwaku/issues/3019


NagyZoltanPeter (approved):
Yessss

SionoiS (approved):

jakubgs (approved):

PR #3023, opened 2024-09-05
chore: peer_manager - enhance max in relay connections when a new one

Ivansete-status:
## Description
Better control of the incoming relay connections.
When a new in-relay connection happens, we are checking if we already reached the max allowed in-relay connections parameter ( `max-relay-peers` )

## Changes
All within `peer_manager.nim`
- Immediately disconnect incoming relay connection when reached the max target of in-relay connections
- Moved the `connectedPeers` proc so that it is available in the `onPeerEvent` proc.

## Related PRs that need to be completed too
https://github.com/status-im/infra-role-nim-waku/pull/31
https://github.com/status-im/infra-waku/pull/24

## Issue
closes https://github.com/waku-org/nwaku/issues/3021

### DIFF
--- a/waku/node/peer_manager/peer_manager.nim
+++ b/waku/node/peer_manager/peer_manager.nim
@@ -399,6 +399,24 @@ proc onPeerMetadata(pm: PeerManager, peerId: PeerId) {.async.} =
   asyncSpawn(pm.switch.disconnect(peerId))
   pm.peerStore.delete(peerId)
 
+proc connectedPeers*(pm: PeerManager, protocol: string): (seq[PeerId], seq[PeerId]) =
+  ## Returns the peerIds of physical connections (in and out)
+  ## containing at least one stream with the given protocol.
+
+  var inPeers: seq[PeerId]
+  var outPeers: seq[PeerId]
+
+  for peerId, muxers in pm.switch.connManager.getConnections():
+    for peerConn in muxers:
+      let streams = peerConn.getStreams()
+      if streams.anyIt(it.protocol == protocol):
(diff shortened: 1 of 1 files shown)


SionoiS (approved):
LGTM

I could not think of any negative consequences since it limit incoming connections only.

Nice one!

waku/node/peer_manager/peer_manager.nim
```diff
@@ -412,6 +430,19 @@ proc onPeerEvent(pm: PeerManager, peerId: PeerId, event: PeerEvent) {.async.} =
...
+    let inRelayPeers = pm.connectedPeers(WakuRelayCodec)[0]
+    if inRelayPeers.len > pm.inRelayPeersTarget and
+        pm.peerStore.hasPeer(peerId, WakuRelayCodec):
+      debug "disconnecting relay peer because reached max num in-relay peers",
+        peerId,
+        inRelayPeers = inRelayPeers.len,
+        inRelayPeersTarget = pm.inRelayPeersTarget
+
+      await pm.switch.disconnect(peerId)
+      pm.peerStore.delete(peerId)
```
gabrielmer, line 443: I would personally only handle the disconnections in `connectToRelayPeers()` or `manageRelayPeers()`, that part of their function is to monitor periodically if we are beyond our targets and prune connections.

I understand making sure we don't exceed the targets in different places, but I think that the role of checking if we are already beyond our target and disconnect from existing peers should happen only in one place.

Maybe I'm missing something. Is there a case in which `connectToRelayPeers()` or `manageRelayPeers()` don't handle the peer targets correctly and we need to do the disconnection over here?

PR #3031, opened 2024-09-10
feat: adding a dynamic sleep interval in the connectivity loop

### LINKED ISSUE #2877: chore: dynamically increase ConnectivityLoopInterval value
### Background
We try to connect to peers every fixed amount of time, which is defined by the following const
https://github.com/waku-org/nwaku/blob/49f03f124349250fb29e3b642a0d3f84e4045fc6/waku/node/peer_manager/peer_manager.nim#L62

The more peers we have, the less we need to run this loop. The inverse is also true - the less peers we have, the more we need to run this.

We should dynamically modify this waiting time depending on how many peers we have

### Acceptance criteria
- [ ] the connectivity loop wait time is dynamic based on current amount of peers

gabrielmer:
# Description

Dynamically setting the sleep time for the next run of the procs that manage peer connections and disconnections.
The farther away we are from our targets, the less time it will pass until the next attempt to connect to peers.

If we're already at our target, we will sleep the whole `ConnectivityLoopInterval`

# Changes

- [x] decreasing the interval between connectivity loop iterations as a function of remaining `out` connections to reach our target
- [x] decreasing the interval between connectivity loop iterations as a function of excess `in` connections

## Issue

closes #2877

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -117,7 +117,7 @@ jobs:
 
           export MAKEFLAGS="-j1"
           export NIMFLAGS="--colors:off -d:chronicles_colors:none"
-          
+
           make V=1 LOG_LEVEL=DEBUG QUICK_AND_DIRTY_COMPILER=1 POSTGRES=$postgres_enabled test testwakunode2
 
   build-docker-image:
@@ -141,25 +141,36 @@ jobs:
       nim_wakunode_image: ${{ needs.build-docker-image.outputs.image }}
       test_type: node-optional
       debug: waku*
-  
+
   lint:
     name: "Lint"
     runs-on: ubuntu-latest
+    needs: build
     steps:
-      - name: Checkout
-        uses: actions/checkout@v4
+      - name: Checkout code
+        uses: actions/checkout@v3
+
+      - name: Get submodules hash
+        id: submodules
+        run: |
+          echo "hash=$(git submodule status | awk '{print $1}' | sort | shasum -a 256 | sed 's/[ -]*//g')" >> $GITHUB_OUTPUT
+
+      - name: Cache submodules
+        uses: actions/cache@v3
         with:
-          fetch-depth: 2  # In PR, has extra merge commit: ^1 = PR, ^2 = base
+          path: |
+            vendor/
+            .git/modules
+          key: ${{ runner.os }}-vendor-modules-${{ steps.submodules.outputs.hash }}
+
+      - name: Build nph
+        run: |
+          make build-nph
 
       - name: Check nph formatting
-        # Pin nph to a specific version to avoid sudden style differences.
-        # Updating nph version should be accompanied with running the new
-        # version on the fluffy directory.
         run: |
-          VERSION="v0.5.1"
-          ARCHIVE="nph-linux_x64.tar.gz"
-          curl -L "https://github.com/arnetheduck/nph/releases/download/${VERSION}/${ARCHIVE}" -o ${ARCHIVE}
-          tar -xzf ${ARCHIVE}
           shopt -s extglob  # Enable extended globbing
-          ./nph examples waku tests tools apps *.@(nim|nims|nimble)
+          NPH=$(make print-nph-path)
+          echo "using nph at ${NPH}"
+          "${NPH}" examples waku tests tools apps *.@(nim|nims|nimble)
           git diff --exit-code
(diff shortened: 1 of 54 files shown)


waku/node/peer_manager/peer_manager.nim
```diff
@@ -965,7 +965,21 @@ proc relayConnectivityLoop*(pm: PeerManager) {.async.} =
...
+      excessInConns = max(inRelayPeers.len - pm.inRelayPeersTarget, 0)
+
+      # One minus the percentage of excess connections relative to the target, limited to 100% 
+      # We calculate one minus this percentage because we want the factor to be inversely proportional to the number of excess peers
+      inFactor = 1 - min(excessInConns / pm.inRelayPeersTarget, 1)
+      # Percentage of out relay peers relative to the target
+      outFactor = min(outRelayPeers.len / pm.outRelayPeersTarget, 1)
+      factor = min(outFactor, inFactor)
+      dynamicSleepInterval =
+        chronos.seconds(int(float(ConnectivityLoopInterval.seconds()) * factor))
```
gabrielmer, line 979: Not a big fan of these casts, couldn't think of a simpler and nicer way, but open for suggestions :))

gabrielmer:
@SionoiS suggested in https://github.com/waku-org/nwaku/issues/2877#issuecomment-2213795222 to also trigger the connectivity loop if a node has many peers and just added many other ones recently. Or in other words, make the interval of a function both of the number of peers and of how recently those peers were added.

Thinking it through and looking at the implementation, I think that adds complexity and don't think that there's actual performance benefits to it, I think that what matters is the amount of active connections and not really when they were established. But open for opinions/ideas/corrections :)

Ivansete-status (commented):
Nevertheless, I wonder if we could make it a bit simpler. I was envisioning just a delay factor depending on how far we are from the max target ( in + out ) and, if we are currently beyond the target then keep the fixed big delay.
Cheers

EDITED: actually, the farther we are from the number connection target (in+out) the faster we need to strive to reach the target again :) It is great how you are considering the exceed num

SionoiS:
> @SionoiS suggested in [#2877 (comment)](https://github.com/waku-org/nwaku/issues/2877#issuecomment-2213795222) to also trigger the connectivity loop if a node has many peers and just added many other ones recently. Or in other words, make the interval of a function both of the number of peers and of how recently those peers were added.

The benefit is balancing the # of peers per shards, since we can't know their shard before connecting via Waku Metadata but it's true that is not strictly needed.

gabrielmer:
There was a proposal 

> Thanks so much for it! 🙌 Nevertheless, I wonder if we could make it a bit simpler. I was envisioning just a delay factor depending on how far we are from the max target ( in + out ) and, if we are currently beyond the target then keep the fixed big delay. Cheers
> 
> EDITED: actually, the farther we are from the number connection target (in+out) the faster we need to strive to reach the target again :) It is great how you are considering the exceed num

Yes! So it ended up looking more complicated than what I originally expected, I added the comments so it's clearer because I couldn't think of a way of doing the same with simpler math.

What I did was basically multiply a factor between 0 and 1 that depends on how far away we are from our targets and reduces the wait the farther away we are from the targets. In theory it's simple, but the excess peers case made it a bit complicated, as it should be inversely proportional to the factor, and at the same time excess peers are not bounded (it can be over 100% of the target). These two properties added extra operations that aren't that intuitive.

My first attempt was with the sum of `in` + `out`, but because both have different targets and behave differently (in one we check that we don't have too few and in the other that we don't have too many), I got the idea that it's like adding apples to oranges and that it should be calculated separately and take the minimum factor out of both of them.

Can you please be a bit more specific on how to simplify the math? I couldn't find a simpler way myself with the math working out, but would love to hear a better way :))

gabrielmer:
> The benefit is balancing the # of peers per shards, since we can't know their shard before connecting via Waku Metadata but it's true that is not strictly needed.

Aren't the shards in the ENR? I thought that balancing the connections by shards is what's done by https://github.com/waku-org/nwaku/blob/1713f562353476235a2076941775a1f150d7c615/waku/node/peer_manager/peer_manager.nim#L763-L839

Maybe I'm mixing things up

Ivansete-status (approved):

SionoiS:
> Aren't the shards in the ENR? I thought that balancing the connections by shards is what's done by

Yes it true but Waku Metadata might not reflect the ENR shard advertised but I guess it's not a big deal.

PR #3032, opened 2024-09-11
chore: libwaku reduce repetitive code by adding a template handling resp returns

Ivansete-status:
## Description
This is inspired by SP's suggestion: https://github.com/waku-org/nwaku/pull/3026#discussion_r1750892090

### DIFF
--- a/library/libwaku.nim
+++ b/library/libwaku.nim
@@ -52,8 +52,8 @@ template foreignThreadGc(body: untyped) =
   when declared(tearDownForeignThreadGc):
     tearDownForeignThreadGc()
 
-template handleRes[T: string | void, E](
-    res: Result[T, E], callback: WakuCallBack, userData: pointer
+template handleRes[T: string | void](
+    res: Result[T, string], callback: WakuCallBack, userData: pointer
 ) =
   ## Manages sendRequestToWakuThread responses in a generic way
   if res.isErr():
--- a/library/libwaku.nim
+++ b/library/libwaku.nim
@@ -52,6 +52,21 @@ template foreignThreadGc(body: untyped) =
   when declared(tearDownForeignThreadGc):
     tearDownForeignThreadGc()
 
+template handleRes[T, E](res: Result[T, E], callback: WakuCallBack, userData: pointer) =
+  ## Manages sendRequestToWakuThread responses in a generic way
+  if res.isErr():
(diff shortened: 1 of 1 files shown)


library/libwaku.nim
```diff
@@ -52,6 +52,21 @@ template foreignThreadGc(body: untyped) =
...
 
+template handleRes[T, E](res: Result[T, E], callback: WakuCallBack, userData: pointer) =
+  ## Manages sendRequestToWakuThread responses in a generic way
+  if res.isErr():
+    foreignThreadGc:
+      let msg = "libwaku error: " & $res.error
+      callback(RET_ERR, unsafeAddr msg[0], cast[csize_t](len(msg)), userData)
+      return RET_ERR
+
+  when T is string:
```
NagyZoltanPeter, line 63: It seems me suspicious, but maybe overlooking sthg.
So what if T is not a string?
I see code that the template replaces which use stringizing the value of the return type, whatever it is.
I think it is not exactly the same behavior.
Also if it is still ok for now, I would add an else case to result in compile error thou shall not hide a problem of misuse.

NagyZoltanPeter (changes requested):
All in all it is a super step I do appreciate a lot.
Please have a look at my concern in handling OK case inside the template.

library/libwaku.nim
SionoiS: In Rust trait bounds are used to restrict types when using generics any such thing in Nim?

SionoiS (approved):

library/libwaku.nim
arnetheduck: `template handleRes[T: string, ...`

NagyZoltanPeter (approved):
Thanks for it!!!

library/libwaku.nim
Ivansete-status: Thank you for the comments guys!
Applied it in the following commits:

[libwaku: narrow down the possible types used by handleRes](https://github.com/waku-org/nwaku/pull/3032/commits/bfb241a2c32b3dbc00edb2bf1676773339f0d6a1)

[libwaku: limit the possible types accepted by handleRes](https://github.com/waku-org/nwaku/pull/3032/commits/fa06e3421e0ae898017833d879a6f627803759be)

PR #3036, opened 2024-09-12
fix: setting up node with modified config

gabrielmer:
# Description
We currently modify a copy of our configurations when initializing a node in https://github.com/waku-org/nwaku/blob/1713f562353476235a2076941775a1f150d7c615/waku/factory/waku.nim#L104-L106

However, we don't use this modified configuration when setting up the protocols and initializing the node, which causes mismatches between them.

Added a fix to use the same version of the configuration at the`wakunode2` application layer.

# Changes

- [x] using modified config copy in `wakunode2`

### DIFF
--- a/waku/factory/waku.nim
+++ b/waku/factory/waku.nim
@@ -101,16 +101,15 @@ proc validateShards(conf: WakuNodeConf): Result[void, string] =
 
 ## Initialisation
 
-proc init*(T: type Waku, conf: WakuNodeConf): Result[Waku, string] =
-  var confCopy = conf
+proc init*(T: type Waku, confCopy: var WakuNodeConf): Result[Waku, string] =
   let rng = crypto.newRng()
 
-  logging.setupLog(conf.logLevel, conf.logFormat)
+  logging.setupLog(confCopy.logLevel, confCopy.logFormat)
 
   # TODO: remove after pubsubtopic config gets removed
   var shards = newSeq[uint16]()
-  if conf.pubsubTopics.len > 0:
-    let shardsRes = topicsToRelayShards(conf.pubsubTopics)
+  if confCopy.pubsubTopics.len > 0:
+    let shardsRes = topicsToRelayShards(confCopy.pubsubTopics)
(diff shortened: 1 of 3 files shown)


NagyZoltanPeter (approved):
Hm, nice catch! Thank you for it!

waku/factory/waku.nim
```diff
@@ -101,16 +101,15 @@ proc validateShards(conf: WakuNodeConf): Result[void, string] =
 
 ## Initialisation
 
-proc init*(T: type Waku, conf: WakuNodeConf): Result[Waku, string] =
-  var confCopy = conf
+proc init*(T: type Waku, confCopy: var WakuNodeConf): Result[Waku, string] =
```
Ivansete-status, line 104: I actually think these are the only needed changes.
Did you try to see what happens with just that?

Ivansete-status (commented):
Thanks for it! Just added a question

waku/factory/waku.nim
gabrielmer: Yes, so the need of each change is the following:

1. In this file, apart from this line, changed the name to refer that we are using a copy of the original configuration and that we're not modifying the user's input

2. In `apps/wakunode2/wakunode2.nim`, we create a copy of the original configuration and pass it to all the procs (so we don't modify the original configuration passed by the user)

3. In `tests/wakunode2/test_app.nim`, had to change from `let` to `var` for the configurations passed to `init()` now that we modify the object in the proc

NagyZoltanPeter: Do we use/re-use the original user configuration afterward? Just curious...

gabrielmer: > Do we use/re-use the original user configuration afterward? Just curious...

After that point we actually don't use the original configuration, but did it just to not override the original user configuration which generally isn't a good pattern. In principle we can avoid doing the copy and change the original but idk, it doesn't feel right haha

Ivansete-status (approved):

PR #3046, opened 2024-09-20
fix: static linking negentropy in ARM based mac

gabrielmer:
# Description

Adding `--passL:-L/opt/homebrew/lib/` when linking negentropy to fix error of missing crypto library in ARM based mac

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -172,7 +172,12 @@ clean: | clean-librln
 .PHONY: negentropy
 
 LIBNEGENTROPY_BUILDDIR := $(CURDIR)/vendor/negentropy/cpp
+ifeq ($(detected_OS),Linux)
 LIBNEGENTROPY_FILE := libnegentropy.a
+else
+LIBNEGENTROPY_FILE := libnegentropy.so
+endif
+
 
 deps: | negentropy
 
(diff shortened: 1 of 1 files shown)


NagyZoltanPeter (approved):
It is that! Thanks for fixing it.

Makefile
```diff
@@ -185,7 +185,7 @@ $(LIBNEGENTROPY_FILE):
 negentropy: | $(LIBNEGENTROPY_FILE)
     ## Pass libnegentropy and it's deps to linker.
     $(eval LIBNEGENTROPY_PATH := $(shell if [ -f "$(LIBNEGENTROPY_FILE)" ]; then echo "$(LIBNEGENTROPY_FILE)"; else echo "./$(LIBNEGENTROPY_FILE)"; fi))
-    $(eval NIM_PARAMS += --passL:$(LIBNEGENTROPY_PATH) --passL:-lcrypto --passL:-lssl --passL:-lstdc++)
+    $(eval NIM_PARAMS += --passL:$(LIBNEGENTROPY_PATH) --passL:-lcrypto --passL:-lssl --passL:-lstdc++ --passL:-L/opt/homebrew/lib/)
```
Ivansete-status, line 188: I think we could be a bit more explicit and use `detected_OS` to give one value or another

Ivansete-status (approved):
LGTM! Thanks! Just minor comment

Makefile
gabrielmer: > I think we could be a bit more explicit and use `detected_OS` to give one value or another

I actually asked @NagyZoltanPeter about it and told me it doesn't hurt and no need to add an if condition for it. I have no specific preference, can do any of both ways :))

NagyZoltanPeter: Well this is opted from the negentropy Makefile we had, homebrew is really just a linker hint, doing nothing harm on linux as there are no libs on such path.

Ivansete-status: > Well this is opted from the negentropy Makefile we had, homebrew is really just a linker hint, doing nothing harm on linux as there are no libs on such path.

ah yes you are right that it doesn't cause any harm but I think is better to be explicit ( or nitpick hehe ;P )

PR #3049, opened 2024-09-23
chore: test peer connection management

Ivansete-status:
# Description
Enhance reconnect logic and fix reconnect tests.
The important point is that it is necessary to set a port different than zero in tests.

# Changes

## Issue
closes
- https://github.com/waku-org/nwaku/issues/2592

### DIFF
--- a/tests/node/test_wakunode_peer_manager.nim
+++ b/tests/node/test_wakunode_peer_manager.nim
@@ -8,6 +8,7 @@ import
   chronos,
   # chronos/timer,
   chronicles,
+  times,
   libp2p/[peerstore, crypto/crypto, multiaddress]
 
 from times import getTime, toUnix
@@ -62,9 +63,9 @@ suite "Peer Manager":
       serverKey = generateSecp256k1Key()
       clientKey = generateSecp256k1Key()
 
-    server = newTestWakuNode(serverKey, listenIp, listenPort)
+    server = newTestWakuNode(serverKey, listenIp, Port(3000))
     serverPeerStore = server.peerManager.peerStore
-    client = newTestWakuNode(clientKey, listenIp, listenPort)
+    client = newTestWakuNode(clientKey, listenIp, Port(3001))
     clientPeerStore = client.peerManager.peerStore
 
     await allFutures(server.start(), client.start())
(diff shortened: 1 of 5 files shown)


tests/node/test_wakunode_peer_manager.nim
```diff
@@ -577,77 +578,52 @@ suite "Peer Manager":
...
-          await sleepAsync(FUTURE_TIMEOUT_LONG)
+          let reconnectDurationWithBackoffPeriod =
+            getTime().toUnixFloat() - beforeReconnect
 
           # Then both peers should be marked as Connected
           check:
             clientPeerStore.get(serverPeerId).connectedness == Connectedness.Connected
             serverPeerStore.get(clientPeerId).connectedness == Connectedness.Connected
+            reconnectDurationWithBackoffPeriod >
+              (reconnectDuration + backoffPeriod.seconds.float)
```
gabrielmer, line 626: Not sure I understand why this is true. Shouldn't they be theoretically equal, but in practice any of them be longer than the other one?

Or in other words, why `reconnectDuration + backoffPeriod.seconds.float` can't be greater than `reconnectDurationWithBackoffPeriod`?

gabrielmer (commented):
Looks amazing! Thanks so much!

Added a small question :)

tests/node/test_wakunode_peer_manager.nim
Ivansete-status: > Not sure I understand why this is true. Shouldn't they be theoretically equal, but in practice any of them be longer than the other one?
> 
> Or in other words, why `reconnectDuration + backoffPeriod.seconds.float` can't be greater than `reconnectDurationWithBackoffPeriod`?

Actually, in both cases we are doing the same operation but, in when the `backoff` time assigned is != 0, then we add this extra block:
```
    if backoffTime > ZeroDuration:
      debug "Backing off before reconnect",
        peerId = peerInfo.peerId, backoffTime = backoffTime
      # We disconnected recently and still need to wait for a backoff period before connecting
      await sleepAsync(backoffTime)
```
... where we are adding an extra `backoffTime` plus the time spent in the `debug` call plus the if statement. Hence, the `>` condition should happen.

gabrielmer: Mmm I think that this can be problematic, the only difference is the `if` statement, the log and the time it takes to return from the sleep after the time has passed.

But for example, if the call to `await pm.connectToNodes(@[peerInfo])` in the case without the backoff for some reason takes longer than in the run with the backoff, it can happen that the statement won't be true.

So given that it can be false in certain cases and that I don't think that the check adds much, I would personally remove it. But if not we can keep it and remove it if we see that it gives problems :)

gabrielmer (approved):

tests/node/test_wakunode_peer_manager.nim
Ivansete-status: Morning! ah yes , makes sense. For now I just simplified the test in https://github.com/waku-org/nwaku/pull/3049/commits/772131ebafe2be1a7465f90087a960c13bf3f0f9
Thanks!

AlejandroCabeza (approved):

PR #3051, opened 2024-09-23
refactor: wrap peer store 

darshankabariya:
# Description
Refactored the peer manager to improve stale peer management.

# Changes

-  [ abstract the peer_store functionality by using waku_peer_store] --> This allows for extending features in waku_peer_store without modifying the core peer_store.
- [ Improved WakuPeerStore: unified peer data, enhanced queries, and added Waku-specific features for better management. ]

### DIFF
--- a/waku/node/peer_manager/peer_manager.nim
+++ b/waku/node/peer_manager/peer_manager.nim
@@ -73,7 +73,7 @@ const
 
 type PeerManager* = ref object of RootObj
   switch*: Switch
-  peerStore*: WakuPeerStore
+  wakuPeerStore*: WakuPeerStore
   wakuMetadata*: WakuMetadata
   initialBackoffInSec*: int
   backoffFactor*: int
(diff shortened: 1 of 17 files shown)


waku/node/peer_manager/peer_manager.nim
```diff
@@ -73,7 +73,7 @@ const
 
 type PeerManager* = ref object of RootObj
   switch*: Switch
-  peerStore*: PeerStore
+  peerStore*: WakuPeerStore
```
Ivansete-status, line 76: I wonder if we should also remane the variable name. Maybe do the same in the tests
```suggestion
  wakuPeerStore*: WakuPeerStore
```

waku/node/peer_manager/waku_peer_store.nim
```diff
@@ -16,14 +16,16 @@ import
 export peerstore, builders
 
 type
+  WakuPeerStore* = ref object
+    store*: PeerStore
```
Ivansete-status, line 20: I think this is a great opportunity to make the field private. In the suggestion I'm also renaming the attribute :)
```suggestion
    peerStore: PeerStore
```

waku/node/peer_manager/waku_peer_store.nim
```diff
@@ -32,126 +34,139 @@ type
...
-proc peers*(peerStore: PeerStore, protocolMatcher: Matcher): seq[RemotePeerInfo] =
-  # Return the known info for all peers matching the provided protocolMatcher
-  peerStore.peers.filterIt(it.protocols.anyIt(protocolMatcher(it)))
+proc peers*(wps: WakuPeerStore, protocolMatcher: Matcher): seq[RemotePeerInfo] =
+  wps.peers().filterIt(it.protocols.anyIt(protocolMatcher(it)))
 
-proc connectedness*(peerStore: PeerStore, peerId: PeerID): Connectedness =
-  peerStore[ConnectionBook].book.getOrDefault(peerId, NotConnected)
+proc connectedness*(wps: WakuPeerStore, peerId: PeerId): Connectedness =
+  wps.store[ConnectionBook].book.getOrDefault(peerId, NotConnected)
```
Ivansete-status, line 110: I wonder if would it work the following, for consistency
```suggestion
  wps[ConnectionBook].book.getOrDefault(peerId, NotConnected)
```

waku/node/peer_manager/waku_peer_store.nim
```diff
@@ -32,126 +34,139 @@ type
...
+    wps.store[ENRBook][peer.peerId] = peer.enr.get()
+
+proc delete*(wps: WakuPeerStore, peerId: PeerId) =
+  # Delete all the information of a given peer.
+  wps.store.del(peerId)
 
 # TODO: Rename peers() to getPeersByProtocol()
-proc peers*(peerStore: PeerStore): seq[RemotePeerInfo] =
-  ## Get all the stored information of every peer.
+proc peers*(wps: WakuPeerStore): seq[RemotePeerInfo] =
```
Ivansete-status, line 93: I wonder if we should also remane those to `getPeers` or `getWakuPeers` to make them more explicit and easier to lookup

Ivansete-status (changes requested):
Great PR! Thanks so much for it! :partying_face: 

I've just suggested some minor changes.
On the other hand, I have the impression that we are encapsulating the code a little, which is great indeed, but we are not yet tackling the stale peers issue, isn't it?

darshankabariya:
> Great PR! Thanks so much for it! 🥳
> 
> I've just suggested some minor changes. On the other hand, I have the impression that we are encapsulating the code a little, which is great indeed, but we are not yet tackling the stale peers issue, isn't it?

Yes, this PR does not tackle the stale peer issues. That stale peer issue isn't as pressing as it seemed when I created it. That's why I focused on another aspect of refactoring. Regarding the stale peer issue, I will add another PR to separate the functionality of the peer manager and the connection lifecycle. In that, I will also improve stale peer management.

gabrielmer (approved):
LGTM, thanks so much!

Ivansete-status (approved):

PR #3055, opened 2024-09-24
fix: px protocol decode - do not treat missing response field as error

NagyZoltanPeter:
# Description
With v0.33.0-rc2 on status.staging it was obvious that client using PeerExchange RPC the old way - on request they completely leave out response field (not even default is added) - cause decode error on nwaku side - that now implements changed protocol.

# Changes
- PeerExchangeRpc will not handle missing response field as error.
- By adding Peer Exchange metrics to grafana dashboard it is better to use counters instead of gauge for better dashboard stats.

- [X] rpc_codec of WakePeerExchange
- [X] `waku_px_peers_sent` and `waku_px_errors` have changed to counter from gauge type. 

## Issue
https://github.com/waku-org/nwaku/issues/3054


Ivansete-status (approved):

gabrielmer (approved):
Thank you!

PR #3057, opened 2024-09-24
chore: append current version in agentString which is used by the identify protocol

Ivansete-status:
As suggested by @weboko , it is interesting to add the current nwaku version within the `identify` protocol response.

This PR achieves that. f.e.:

image


gabrielmer (approved):
LGTM, thank you!

NagyZoltanPeter (approved):
Great idea, wow!

Issue #3039, opened 2024-09-17
feat: nwaku in status-go (discv5 and dnsdiscovery)


Ivansete-status:
### Description

status-go should support discv5, dnsdiscovery throughout nwaku

To validate that, we need the `TestBasicWakuV2` to work properly. We will consider this as achieved when the following lines work well (then we need to temporarily comment out the rest of the test:)

https://github.com/status-im/status-go/blob/94dc345d8ef5fb95fd855a665b1d32f3b0e32008/wakuv2/waku_test.go#L180-L247

PR #3058, opened 2024-09-25
chore: disabling metrics for libwaku

gabrielmer:
# Description
When running a node using libwaku, we repeatedly get the following error

```
metrics error: New label values must be added from same thread as the metric was created from - observation dropped: libp2p_gossipsub_peers_per_topic_gossipsub collector
metrics error: New label values must be added from same thread as the metric was created from - observation dropped: libp2p_gossipsub_peers_per_topic_fanout collector.
metrics error: New label values must be added from same thread as the metric was created from - observation dropped: libp2p_gossipsub_peers_per_topic_mesh collector.
```

Disabling by default the metrics for libwaku until the root cause is taken care of

# Changes

- [x] undefining the metrics compilation flag
- [x] modified `nanosecondTime` to accept the type passed when metrics are disabled
- [x] assigning a proc to `logMetrics` only when metrics are defined 

## Issue
#3039

### DIFF
--- a/waku.nimble
+++ b/waku.nimble
@@ -63,12 +63,12 @@ proc buildLibrary(name: string, srcDir = "./", params = "", `type` = "static") =
     extra_params &= " " & paramStr(i)
   if `type` == "static":
     exec "nim c" & " --out:build/" & name &
-      ".a --threads:on --app:staticlib --opt:size --noMain --mm:refc --header " &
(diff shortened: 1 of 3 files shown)


Ivansete-status (approved):
LGTM! Thanks for it! :100:

NagyZoltanPeter (approved):
Awesome!

PR #3059, opened 2024-09-25
fix: PeerExchange rpc decode in order not to take response's status_code mandatory - for support old protocol implementation

NagyZoltanPeter:
# Description
lite-protocol-tester implementation started using PX from the receiver side and found that node that implements old style PeerExchange protocol does not containing status_code and status_desc can cause decode failure.
To handle this false assumption of making status_code for response mandatory, if not included the value is derived from the response enr list length.

+1
I found that PeerExchange tries to aggressively fill the ENR cache at startup. It takes a minimum 60 Peers to fill in the cache. But if there is not many this loop will run - async - till end of program at every 5.secs, instead of running at every 10 minutes as designed.
So to break the loop I added a max 10 times trial of aggressive filling.

# Changes

- [x] PeerExchange rpc_codec - to derive status_code value if not included in the protobuf payload
- [x] PeerExchange protocol initial ENR cache fill limited to 10 trials.

### DIFF
--- a/waku/waku_peer_exchange/protocol.nim
+++ b/waku/waku_peer_exchange/protocol.nim
@@ -217,7 +217,9 @@ proc populateEnrCache(wpx: WakuPeerExchange) =
 
 proc updatePxEnrCache(wpx: WakuPeerExchange) {.async.} =
   # try more aggressively to fill the cache at startup
-  while wpx.enrCache.len < MaxPeersCacheSize:
+  var attempts = 10
+  while wpx.enrCache.len < MaxPeersCacheSize and attempts > 0:
+    attempts -= 1
     wpx.populateEnrCache()
     await sleepAsync(5.seconds)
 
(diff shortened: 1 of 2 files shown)


gabrielmer (approved):
LGTM, thank you!

waku/waku_peer_exchange/protocol.nim
```diff
@@ -217,7 +217,9 @@ proc populateEnrCache(wpx: WakuPeerExchange) =
 
 proc updatePxEnrCache(wpx: WakuPeerExchange) {.async.} =
   # try more aggressively to fill the cache at startup
-  while wpx.enrCache.len < MaxPeersCacheSize:
+  var attempts = 10
+  while wpx.enrCache.len < MaxPeersCacheSize and attempts > 0:
+    attempts -= 1
```
Ivansete-status, line 222: Shall we establish a timeout instead of a fixed num of attempts?

PR #3065, opened 2024-09-27
fix: rejecting excess relay connections

### LINKED ISSUE #3063: bug: excess relay connections are not being properly pruned
### Problem
In the `waku.sandbox` fleet, we noticed that relay peers were consuming all the configured `libp2p` connections.

Looking further at the logs, we found that there's an excess of `in` connections not being properly taken care of

```
    
topics="waku node peer_manager" tid=1 file=peer_manager.nim:981 inRelayConns=252/134 outRelayConns=46/66 totalConnections=300/300 notConnectedPeers=2028 outsideBackoffPeers=2028
```

### Expected behavior
The node should properly prune excess peer connections and respect its targets

gabrielmer:
# Description
Rejecting incoming Relay connections whenever we reached our `in` connections target.

Currently, we accept connections and prune the excess connections periodically. However, this approach is not working out properly, as nodes can reconnect to us right after we disconnect to them, making no difference.

This makes the node vulnerable to attacks, as other nodes can fill all our connections without leaving libp2p connections to serve non-Relay nodes.

This PR addresses the issue, disconnecting from a Relay node right after receiving their request, in case we already reached our target.

The content of this PR was actually proposed by @Ivansete-status some weeks ago :) Thanks so much!

# Changes

- [x] disconnecting from incoming connections after reaching `inRelayPeersTarget`
- [x] don't set an arbitrary minimum of `outRelayPeersTarget`
- [x] remove unused variables 

## Issue

closes #3063

### DIFF
--- a/waku/node/peer_manager/peer_manager.nim
+++ b/waku/node/peer_manager/peer_manager.nim
@@ -440,7 +440,7 @@ proc onPeerEvent(pm: PeerManager, peerId: PeerId, event: PeerEvent) {.async.} =
     if inRelayPeers.len > pm.inRelayPeersTarget and
         pm.peerStore.hasPeer(peerId, WakuRelayCodec):
       debug "disconnecting relay peer because reached max num in-relay peers",
-        peerId,
+        peerId = peerId,
         inRelayPeers = inRelayPeers.len,
         inRelayPeersTarget = pm.inRelayPeersTarget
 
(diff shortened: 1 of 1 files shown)


Ivansete-status (approved):
LGTM! Thanks for it! :100:

richard-ramos (approved):

PR #3077, opened 2024-10-01
fix: out connections leak

### LINKED ISSUE #3063: bug: excess relay connections are not being properly pruned

gabrielmer:
# Description

Once we started promptly disconnecting from excess `in` connections, we began seeing our nodes significantly exceeding their `out` connections targets.

The root cause was a race condition in our keep alive loop https://github.com/waku-org/nwaku/blob/643ab20fc67f251987d594cfb5aa4abb60ccc5b2/waku/node/waku_node.nim#L1241-L1258

The case is the following:
1. A node receives incoming connections beyond its target, and it takes a small amount of time from the moment `nim-libp2p` accepts the connection until our peer manager notices that it's beyond our `in` target and disconnects
2. In the time while we're connected to this `in` connection, we start running the keep alive loop and have that peer in the list of connected peers that we should ping
3. While we're pinging other peers, we disconnect from the `in` connection as we noticed it's beyond our target
4. Because the list of the nodes to ping was generated before we disconnected from the node, we ping the node. As there's no existing connection, we end up creating a new `out` connection towards the node

The proposed change to avoid this race condition is to delegate the responsibility of the periodic ping to the node that originally initiated the connection. Or in other words, whoever initiated a connection is the one responsible to ping periodically to maintain it open - there's no need to have both nodes pinging each other.

# Changes

- [x] extended `connectedPeers()` to allow to get connected peers from all protocols
- [x] modified `keepaliveLoop` so that we only ping nodes in our `out` connections list 

## Issue

closes #3063

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -12,7 +12,7 @@
 	path = vendor/nim-libp2p
 	url = https://github.com/vacp2p/nim-libp2p.git
 	ignore = dirty
-	branch = debug-excess-peer-connections
+	branch = master
 [submodule "vendor/nim-stew"]
 	path = vendor/nim-stew
 	url = https://github.com/status-im/nim-stew.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -12,7 +12,7 @@
 	path = vendor/nim-libp2p
 	url = https://github.com/vacp2p/nim-libp2p.git
 	ignore = dirty
-	branch = master
+	branch = debug-excess-peer-connections
 [submodule "vendor/nim-stew"]
 	path = vendor/nim-stew
 	url = https://github.com/status-im/nim-stew.git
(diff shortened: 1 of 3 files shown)


gabrielmer:
> LGTM
> 
> Do we need to revisit how missed pings are handled? If only one side pings maybe we should be more lenient before disconnecting.
> 
> It may not be a problem in practice, IDK.

Great point! I see that the connection should timeout after 4-5 missed pings (~10 minutes without being reachable)

https://github.com/waku-org/nwaku/blob/e406673c4618f95b8c2536e99eb9b22adbbd2149/waku/node/waku_node.nim#L1261

I think it looks reasonable? Don't think it should give issues, lmk what you think :)

PR #3078, opened 2024-10-01
chore: bump negentropy

richard-ramos:
Use latest version of `negentropy` which would allow it to be compiled when using nix


NagyZoltanPeter (approved):
Yepp!

PR #3093, opened 2024-10-07
fix: changing libwaku's error handling format

gabrielmer:
# Description
For some reason, when immediately chaining `sendRequestToWakuThread`'s response with the `handleRes` template, there's cases in which the errors are not handled properly. See this [message](https://discord.com/channels/1110799176264056863/1197844654218039387/1291722090847670292) for a more detailed description of the issue.

It's probably due to some misusage of Nim templates, but until we further understand what's wrong with out chaining approach, modifying to a working version.

## Update
The template usage was wrong - changed it to a proc and now works great :)

# Changes

## Issue
#3039

### DIFF
--- a/library/libwaku.nim
+++ b/library/libwaku.nim
@@ -52,9 +52,9 @@ template foreignThreadGc(body: untyped) =
   when declared(tearDownForeignThreadGc):
     tearDownForeignThreadGc()
 
-template handleRes[T: string | void](
+proc handleRes[T: string | void](
     res: Result[T, string], callback: WakuCallBack, userData: pointer
-) =
+): cint =
   ## Handles the Result responses, which can either be Result[string, string] or
   ## Result[void, string]. Notice that in case of Result[void, string], it is enough to
   ## just return RET_OK and not provide any additional feedback through the callback.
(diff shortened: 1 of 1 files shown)


Ivansete-status (approved):

gabrielmer:
@arnetheduck whenever you're available could you please take a quick look? maybe you have a better understanding on what's going on 😶 we're not sure why does

```
waku_thread
  .sendRequestToWakuThread(
    ctx,
    RequestType.LIFECYCLE,
    NodeLifecycleRequest.createShared(NodeLifecycleMsgType.START_NODE),
  )
  .handleRes(callback, userData)
 ```

behave differently than

```
let res = waku_thread
  .sendRequestToWakuThread(
    ctx,
    RequestType.LIFECYCLE,
    NodeLifecycleRequest.createShared(NodeLifecycleMsgType.START_NODE),
  )

res.handleRes(callback, userData)
```

`handleRes` is defined here
https://github.com/waku-org/nwaku/blob/0f8e87400024a4f0cb08b2f2db18336b40e1e8c0/library/libwaku.nim#L55-L72

**Review by richard-ramos — APPROVED** [review]

**Comment by arnetheduck** [comment]
```
[12:22]arnetheduck: the way the code is written, the handleRes template is calling sendRequestToWakuThread twice when there's an error, I'm pretty sure that's not what you intended
[12:24]arnetheduck: parameters in templates get their full ast expanded at every point of use - if you use res twice, you get two expansions of sendRequestToWakuThread...
[12:25]arnetheduck: that's why moving to let "fixes" it - instead of two calls, it accesses the variable twice
```

PR #3097, opened 2024-10-07
fix: remove quotes from libwaku's returned enr string

gabrielmer:
# Description
In libwaku, when processing a `RETRIEVE_MY_ENR` request we are adding an extra `"` in the beginning and in the end of the string.

This is necessary for `RETRIEVE_LISTENING_ADDRESSES` because we have to convert a sequence to a string, but for`RETRIEVE_MY_ENR` which already has a return type of type string, this ends up adding extra quotes.
# Changes

- [x] returning string in `RETRIEVE_MY_ENR` as is

## Issue
#3039

### DIFF
--- a/library/waku_thread/inter_thread_communication/requests/debug_node_request.nim
+++ b/library/waku_thread/inter_thread_communication/requests/debug_node_request.nim
@@ -30,7 +30,7 @@ proc process*(
   of RETRIEVE_LISTENING_ADDRESSES:
     return ok($(%*waku.node.getMultiaddresses()))
   of RETRIEVE_MY_ENR:
-    return ok($(%*waku.node.enr.toURI()))
+    return ok(waku.node.enr.toURI())
 
   error "unsupported operation in DebugNodeRequest"
   return err("unsupported operation in DebugNodeRequest")


richard-ramos (approved):

NagyZoltanPeter (approved):
LGTM!

library/waku_thread/inter_thread_communication/requests/debug_node_request.nim
```diff
@@ -30,7 +30,7 @@ proc process*(
   of RETRIEVE_LISTENING_ADDRESSES:
     return ok($(%*waku.node.getMultiaddresses()))
   of RETRIEVE_MY_ENR:
-    return ok($(%*waku.node.enr.toURI()))
+    return ok(waku.node.enr.toURI())
```
Ivansete-status, line 33: Maybe we need to do the same with the other cases

Ivansete-status (approved):
LGTM! Thanks! :raised_hands:

library/waku_thread/inter_thread_communication/requests/debug_node_request.nim
gabrielmer: There's only one additional case, `RETRIEVE_LISTENING_ADDRESSES`, and I think it's needed there because the result of `getMultiaddresses()`is a sequence, so we need to convert the sequence to a string

PR #3108, opened 2024-10-10
fix: segfault for libwaku's empty response messages

gabrielmer:
# Description

Whenever we called `callback(RET_OK, unsafeAddr msg[0], cast[csize_t](len(msg)), userData)` with `msg` being an empty string, we got a segfault.

Changed de data type of `msg` to `cstring` instead of nim strings, which fixes the segfault.

Additionally, fixed the usage of `foreignThreadGc` and added usage of `handleRes` in missing parts
# Changes

- [x] using `cstring` for libwaku's response messages
- [x] fixed the usage of `foreignThreadGc` as part of its code was unreachable when returning 
- [x] using `handleRes` in missing parts

## Issue
#3039


NagyZoltanPeter (approved):
LGTM!!! Thank you.

PR #3109, opened 2024-10-10
chore: improving get_peer_ids_by_protocol in libwaku

gabrielmer:
# Description

Currently, libwaku's `GET_PEER_IDS_BY_PROTOCOL` returns the peer ids of peers that have an open libp2p stream associated with a given protocol.

However, what's most interesting for us is the capabilities of our connected peers and not which protocols are we actually using.

For example, for req-res protocols such as Store, most of the time we don't have an open stream associated with the Store protocol - which means that with the current approach `GET_PEER_IDS_BY_PROTOCOL` won't return anything.

Therefore, changing the approach so we return all our connected peers that support a given protocol, even if we don't have open streams for that protocol in specific.

# Changes

- [x] improving `GET_PEER_IDS_BY_PROTOCOL` so it returns our connected peers with a given capability

## Issue

#3039

### DIFF
--- a/library/waku_thread/inter_thread_communication/requests/peer_manager_request.nim
+++ b/library/waku_thread/inter_thread_communication/requests/peer_manager_request.nim
@@ -81,8 +81,11 @@ proc process*(
     return ok(peerIDs)
   of GET_PEER_IDS_BY_PROTOCOL:
     ## returns a comma-separated string of peerIDs that mount the given protocol
-    let (inPeers, outPeers) = waku.node.peerManager.connectedPeers($self[].protocol)
-    let allPeerIDs = inPeers & outPeers
-    return ok(allPeerIDs.join(","))
+    let connectedPeers = waku.node.peerManager.wakuPeerStore
+      .peers($self[].protocol)
+      .filterIt(it.connectedness == Connected)
+      .mapIt($it.peerId)
+      .join(",")
+    return ok(connectedPeers)
 
   return ok("")


Ivansete-status (approved):
Wonderful! Thanks for it! :100:

library/waku_thread/inter_thread_communication/requests/peer_manager_request.nim
```diff
@@ -81,8 +81,11 @@ proc process*(
     return ok(peerIDs)
   of GET_PEER_IDS_BY_PROTOCOL:
     ## returns a comma-separated string of peerIDs that mount the given protocol
-    let (inPeers, outPeers) = waku.node.peerManager.connectedPeers($self[].protocol)
-    let allPeerIDs = inPeers & outPeers
-    return ok(allPeerIDs.join(","))
+    let connectedPeers = waku.node.peerManager.wakuPeerStore
+      .peers($self[].protocol)
+      .filterIt(it.connectedness == Connected)
```
NagyZoltanPeter, line 86: Just asking due to my missing knowledge here.
What is the aim of this req? Would not we extend the list with also CanConnect state peers?

+1: I learned that peers gathered with PX won't get protocols filled until they are not called. So maybe they are in a kind of stale state in peer manager. We know their capabilities (advertised) but we cannot filter them by protocol.

NagyZoltanPeter (approved):
Seems ok, but left a question to consider.

library/waku_thread/inter_thread_communication/requests/peer_manager_request.nim
gabrielmer: The idea of the request is to get the peers we're connected to that support a given protocol. If for example we want to do a Store query, the first thing we would do is see if we're connected to a Store node and send them a Store query.

I think we can also separately, when needed, add a request that includes peers in the peer manager in CanConnect state, but in this case we want the information of already connected nodes.

I believe that the issue with PX, correct me if I'm wrong, is that we get addresses of peers that we store, but we don't really connect to them. Once we're connected to them, we get the protocols when performing Identify, and they get filled.

So in this case for example, if we are connected to a peer we found via PX, it will have the protocol filled, as we are already connected.

NagyZoltanPeter: I got it, thank you. 
About PX. I think this issue there that we will never select such peers as potential service peers, because they have no protocols filled due not connected. But isn't that a trap of 22? 
In js-waku they inherently fill the protocol derived from capability, so those peers get selectable for service.
This is not future proof but inevitably a better option IMHO.
Or is there any mechanism that will consider auto call those peers from PX randomly?

PR #3117, opened 2024-10-15
fix: add a limit of max 10 content topics per query

richard-ramos:
Adds the same restriction we had on waku archive legacy of a maximum of 10 content topics per query as it currently was limitless. We have seen that the store query performance drops the more content topics we add so adding a limit of 10 will help avoid this situation as well as comply with the current assumptions of status-go

### DIFF
--- a/waku/waku_archive/archive.nim
+++ b/waku/waku_archive/archive.nim
@@ -22,6 +22,7 @@ logScope:
 const
   DefaultPageSize*: uint = 20
   MaxPageSize*: uint = 100
+  MaxContentTopicsPerQuery*: int = 10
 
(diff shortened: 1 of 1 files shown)


NagyZoltanPeter (approved):
In general it is ok and need to protect the interface, but I think it might worth to measure where the perf drops, due otherwise it can lead to too many queries, that will in contradict increase the overall load.
Also will be interesting to see how to properly set rate limit for store that still serve the clients (like Status app using a lots of content-topics) but does not allow DOS the node.

Ivansete-status (approved):
LGTM! Thanks for it! :100:

PR #3118, opened 2024-10-15
chore: ci - force use ubuntu-22.04

Ivansete-status:


gabrielmer (approved):
Graciasss

NagyZoltanPeter (approved):
LGTM!
Do we need `sudo apt-get install -y libpcre3 libpcre3-dev` still?

Ivansete-status:
> LGTM! Do we need `sudo apt-get install -y libpcre3 libpcre3-dev` still?

Very good point indeed! I think it shouldn't be needed now

Issue #3115, opened 2024-10-15
feat: nwaku in status-go (peer exchange)


gabrielmer:
status-go should support peer exchange through nwaku

To validate that, we need [TestPeerExchange](https://github.com/status-im/status-go/blob/62b66580505152be6dc647f90d3b1c0bda99f93e/wakuv2/nwaku_test.go#L324-L402) to work properly

gabrielmer:
Closed by https://github.com/status-im/status-go/pull/5983

PR #3121, opened 2024-10-16
chore: returning seqs in libwaku as comma separated strings

gabrielmer:
# Description
In libwaku, when returning sequences as outputs, in some places we are returning the sequence as a comma-separated string such as in https://github.com/waku-org/nwaku/blob/18c375a5deea569e4bbef09d1f9376ad92190bc5/library/waku_thread/inter_thread_communication/requests/peer_manager_request.nim#L79-L90
while in others we return a string wrapped with the `[ ]` characters, representing an array

For the sake of simplicity and consistence - and mainly so the parsing client-side can be the same in all cases - I propose to return all sequences as comma-separated strings

# Changes

- [x] returning output of `RETRIEVE_LISTENING_ADDRESSES` as comma-separated string
- [x] returning output of `GET_BOOTSTRAP_NODES` as comma-separated string

## Issue
#3115

### DIFF
--- a/library/waku_thread/inter_thread_communication/requests/debug_node_request.nim
+++ b/library/waku_thread/inter_thread_communication/requests/debug_node_request.nim
@@ -1,5 +1,5 @@
 import std/json
-import chronicles, chronos, results, eth/p2p/discoveryv5/enr
+import chronicles, chronos, results, eth/p2p/discoveryv5/enr, strutils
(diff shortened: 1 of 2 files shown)


Ivansete-status (approved):
LGTM! Thanks for it! :100:

NagyZoltanPeter (approved):
Looks great!

PR #3122, opened 2024-10-16
fix: avoid segfault by setting a default num peers requested in PX

gabrielmer:
# Description

When using Peer Exchange in nwaku, we are setting the number of peers requested by using a Relay-specific parameter, which isn't set if we use nwaku as a light client and produces a segfault.

Therefore, defined a `DefaultPXNumPeersReq` constant with a value of 5 to use whenever no argument is provided or whenever Relay is not mounted 

# Changes

- [x] defined and integrated `DefaultPXNumPeersReq` constant
- [x] fixed segfault by not reading Relay's parameters if Relay is not mounted

## Issue
#3115

### DIFF
--- a/waku/factory/node_factory.nim
+++ b/waku/factory/node_factory.nim
@@ -414,7 +414,11 @@ proc startNode*(
 
   # retrieve px peers and add the to the peer store
   if conf.peerExchangeNode != "":
-    let desiredOutDegree = node.wakuRelay.parameters.d.uint64()
+    let desiredOutDegree =
+      if not node.wakuRelay.isNil():
+        node.wakuRelay.parameters.d.uint64()
+      else:
+        DefaultPXNumPeersReq
     (await node.fetchPeerExchangePeers(desiredOutDegree)).isOkOr:
       error "error while fetching peers from peer exchange", error = error
       quit(QuitFailure)
--- a/waku/factory/node_factory.nim
+++ b/waku/factory/node_factory.nim
@@ -414,11 +414,9 @@ proc startNode*(
 
   # retrieve px peers and add the to the peer store
   if conf.peerExchangeNode != "":
-    let desiredOutDegree =
-      if not node.wakuRelay.isNil():
-        node.wakuRelay.parameters.d.uint64()
-      else:
-        DefaultPXNumPeersReq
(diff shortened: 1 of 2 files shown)


Ivansete-status (approved):
Thanks for it!

waku/waku_peer_exchange/protocol.nim
```diff
@@ -57,7 +58,7 @@ type
     requestRateLimiter*: RequestRateLimiter
 
 proc request*(
-    wpx: WakuPeerExchange, numPeers: uint64, conn: Connection
+    wpx: WakuPeerExchange, numPeers = DefaultPXNumPeersReq, conn: Connection
```
NagyZoltanPeter, line 61: Is it necessary to apply default here? ... and for other variants?

waku/factory/node_factory.nim
```diff
@@ -414,7 +414,11 @@ proc startNode*(
 
   # retrieve px peers and add the to the peer store
   if conf.peerExchangeNode != "":
-    let desiredOutDegree = node.wakuRelay.parameters.d.uint64()
+    let desiredOutDegree =
+      if not node.wakuRelay.isNil():
+        node.wakuRelay.parameters.d.uint64()
+      else:
+        DefaultPXNumPeersReq
```
NagyZoltanPeter, line 421:
```suggestion
    var desiredOutDegree = DefaultPXNumPeersReq
    if not node.wakuRelay.isNil() and node.wakuRelay.parameters.d.uint64() > 0:
      desiredOutDegree = node.wakuRelay.parameters.d.uint64()        
```
WDYT, would that make sense?

NagyZoltanPeter (approved):
It is a great catch! Thank you for it.
Left some opinions... :-)

richard-ramos (approved):

waku/waku_peer_exchange/protocol.nim
gabrielmer: Over here it's not necessary! But added it because given that we now defined a default, it would make sense for that number to be used whenever no value is provided

waku/factory/node_factory.nim
gabrielmer: Ooh that's nice! More concise :))

Just applied it! Thank you!

PR #3123, opened 2024-10-16
fix: added randomness to peer selection

### LINKED ISSUE #3120: bug: Peer manager does not return random peers
### Problem
When asking the peer manager for a peer, the first peer for a given protocol is return, which means that at startup in the fleet all nodes dial the same peer.

### Impact
The chosen one receives a lot of connections.

### Expected behavior
The peer manager NEVER return the same peers if it can be avoided.

see
https://github.com/waku-org/nwaku/blob/18c375a5deea569e4bbef09d1f9376ad92190bc5/waku/node/peer_manager/peer_manager.nim#L250C5-L250C26

SionoiS:
Fixes #3120


Ivansete-status (approved):

gabrielmer (approved):
Thank you!

PR #3127, opened 2024-10-17
chore: saving peers enr capabilities

gabrielmer:
# Description
When adding peers, we used to ignore the protocols advertised in the ENR and would fill the protocols in the Peer Store only after performing Identify.

We're now initializing the protocols with the advertised capabilities and will get updated once Identify is performed.
When mapping the ENR's capabilities to the supported protocols, we assume that the peers use the same version of the protocol as ours. If that's not the case, we'll find out during Identify, update the information and act accordingly.

In addition to that, started adding the Discv5 bootstrap nodes to the Peer Store.

I moved all the Waku protocols codec definitions to a separate file, as in the current state we had to import most of a protocol even if we only needed a codec, which caused circular dependency issues. Protocol codecs are public information, so it makes sense to be able to access them without having to import an entire protocol :)

# Changes

- [x] created new file with Waku Protocols' codecs
- [x] reading ENRs' capabilities and using them to fill the supported protocols when building the `RemotePeerInfo` 
- [x] only attempt Relay connections to bootstrap nodes when Relay is enabled
- [x] adding Discv5 bootstrap nodes to the Peer Store 
- [x] including tests for the new functionalities

### DIFF
--- a/tests/test_waku_enr.nim
+++ b/tests/test_waku_enr.nim
@@ -1,7 +1,7 @@
 {.used.}
 
 import std/[options, sequtils], stew/results, testutils/unittests
-import waku/waku_core, waku/waku_enr, ./testlib/wakucore
+import waku/waku_core, waku/waku_enr, ./testlib/wakucore, waku/waku_core/codecs
 
 suite "Waku ENR -  Capabilities bitfield":
   test "check capabilities support":
@@ -96,6 +96,28 @@ suite "Waku ENR -  Capabilities bitfield":
       bitfield.supportsCapability(Capabilities.Filter) == false
       bitfield.supportsCapability(Capabilities.Lightpush) == false
       bitfield.toCapabilities() == @[Capabilities.Relay, Capabilities.Store]
+
+    test "get capabilities codecs from record":
+      ## Given
+      let
+        enrSeqNum = 1u64
+        enrPrivKey = generatesecp256k1key()
+
+      ## When
+      var builder = EnrBuilder.init(enrPrivKey, seqNum = enrSeqNum)
+      builder.withWakuCapabilities(Capabilities.Relay, Capabilities.Store)
+
+      let recordRes = builder.build()
+
+      ## Then
+      check recordRes.isOk()
+      let record = recordRes.tryGet()
+
+      let codecs = record.getCapabilitiesCodecs()
+      check:
+        codecs.len == 2
+        codecs.contains(WakuRelayCodec)
+        codecs.contains(WakuStoreCodec)
 
   test "check capabilities on a non-waku node record":
     ## Given
--- a/tests/test_waku_enr.nim
+++ b/tests/test_waku_enr.nim
@@ -110,7 +110,7 @@ suite "Waku ENR -  Capabilities bitfield":
       let recordRes = builder.build()
 
       ## Then
-      check recordRes.isOk()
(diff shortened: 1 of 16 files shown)


tests/waku_discv5/test_waku_discv5.nim
gabrielmer, line 1: Had to move some variables around as many were unnecessary in a global scope and gave issues related to GC safety.

Also, changed the `validEnr` value because apparently that ENR was missing fields

tests/test_waku_enr.nim
```diff
@@ -97,6 +97,28 @@ suite "Waku ENR -  Capabilities bitfield":
...
+        enrPrivKey = generatesecp256k1key()
+
+      ## When
+      var builder = EnrBuilder.init(enrPrivKey, seqNum = enrSeqNum)
+      builder.withWakuCapabilities(Capabilities.Relay, Capabilities.Store)
+
+      let recordRes = builder.build()
+
+      ## Then
+      check recordRes.isOk()
```
Ivansete-status, line 113:
```suggestion
      assert recordRes.isOk(), recordRes.error
```

tests/waku_core/test_peers.nim
```diff
@@ -141,3 +146,34 @@ suite "Waku Core - Peers":
...
+      ipAddr = some(parseIpAddress("127.0.0.1")),
+      tcpPort = some(Port(0)),
+      udpPort = some(Port(0)),
+    )
+    builder.withWakuCapabilities(Capabilities.Relay, Capabilities.Store)
+
+    let recordRes = builder.build()
+
+    ## Then
+    check recordRes.isOk()
```
Ivansete-status, line 167:
```suggestion
    assert recordRes.isOk(), recordRes.error
```

waku/factory/node_factory.nim
```diff
@@ -398,13 +398,13 @@ proc startNode*(
     return err("failed to start waku node: " & getCurrentExceptionMsg())
 
   # Connect to configured static nodes
-  if conf.staticnodes.len > 0:
+  if conf.relay and conf.staticnodes.len > 0:
```
Ivansete-status, line 401: In this case I'd give some more feedback in case wrong setup:
```suggestion
  if conf.staticnodes.len > 0:
    if not conf.relay:
      return err("waku relay (--relay=true) should be set when configuring staticnodes")
```

waku/factory/node_factory.nim
```diff
@@ -398,13 +398,13 @@ proc startNode*(
...
     try:
       await connectToNodes(node, conf.staticnodes, "static")
     except CatchableError:
       return err("failed to connect to static nodes: " & getCurrentExceptionMsg())
 
-  if dynamicBootstrapNodes.len > 0:
+  if conf.relay and dynamicBootstrapNodes.len > 0:
```
Ivansete-status, line 407: Similarly to the previous comment, I think it'd be interesting to bring some more feedback

Ivansete-status (approved):
Outstanding one! Thanks :) !

waku/factory/node_factory.nim
gabrielmer: Ooh great idea! Thanks so much!

NagyZoltanPeter (approved):
LGTM! Awesome, thank you.

PR #3130, opened 2024-10-19
chore: peer_manager - prevent too intense loop when no peers connected

Ivansete-status:
## Description
PR aimed to prevent the node from running a very intense CPU loop. With that, at least we set a delay of one second, in case no peers are connected.

I came across that the node was running 100% CPU while testing local nodes without peers.


gabrielmer (approved):
Great catch! Thanks so much!

Issue #3132, opened 2024-10-21
chore: log message validate error and return error


stubbsta:
### Background
When message validation fails, the error metric is updated, but no error is logged or returned to the calling function. See https://github.com/waku-org/nwaku/blob/c5a825e206c1cb3e6e0cf8c01410527804cb76c4/waku/waku_archive/archive.nim#L85C1-L90C11

### Details
Without the logging it is unclear that archiving doesn't occur at all.

### Acceptance criteria
The error should be logged, TRACE level should be sufficient. An error should be returned.

PR #3133, opened 2024-10-21
fix: add log and archive message ingress for sync

### LINKED ISSUE #3131: bug:  add a path for waku sync to bypass the timestamp check of 'validate'
### Problem
A store node with sync enabled fails to archive the "missing" messages identified by the sync process, because their timestamps are too old.

### Impact
The store sync nodes indicate that they are synced, because the waku sync storage is updated, but the messages are never inserted into the DB. StoreQueryRequests will return no results.

### To reproduce
1. Run 2 store nodes with their own DB backends, one with relay enabled
2. Send a few relay message to the node which has it enabled.
3. Wait for the other node to sync
4. See that messages that are older than 20s at the time of syncing are not archived

### Expected behavior
It is expected for all messages identified by the store sync process to be required by the store node, should be archived. the timestamps of the messages are already filtered to be within the `store-sync-range` config

### Screenshots/logs
No error logs are produced for this issue

### nwaku version/commit hash
This was specifically seen on nwaku versions: v0.33.0-rc.0-13-gd8e6a5 and v0.33.0-44-gc5a825, but it would have been present since store sync was implemented.

SionoiS:
Closes #3131 & #3132

### DIFF
--- a/waku/waku_archive/archive.nim
+++ b/waku/waku_archive/archive.nim
@@ -53,7 +53,7 @@ type WakuArchive* = ref object
 proc validate*(msg: WakuMessage): Result[void, string] =
   if msg.ephemeral:
     # Ephemeral message, do not store
-    return ok()
+    return
 
   let
     now = getNanosecondTime(getTime().toUnixFloat())
--- a/waku/waku_archive/archive.nim
+++ b/waku/waku_archive/archive.nim
@@ -53,7 +53,7 @@ type WakuArchive* = ref object
 proc validate*(msg: WakuMessage): Result[void, string] =
   if msg.ephemeral:
     # Ephemeral message, do not store
-    return
+    return ok()
 
   let
     now = getNanosecondTime(getTime().toUnixFloat())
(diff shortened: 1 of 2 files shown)


waku/waku_archive/archive.nim
```diff
@@ -87,6 +87,12 @@ proc handleMessage*(
 ) {.async.} =
   self.validator(msg).isOkOr:
     waku_archive_errors.inc(labelValues = [error])
+    trace "failed to insert message",
+      hash_hash = computeMessageHash(pubsubTopic, msg).to0xHex(),
```
gabrielmer, line 91: shouldn't it be `msg_hash` instead of `hash_hash`? especially for consistency with all the other logs in the codebase

same comment for the other logs added here

waku/waku_archive/archive.nim
```diff
@@ -111,6 +117,33 @@ proc handleMessage*(
...
+    self: WakuArchive,
+    msgHash: WakuMessageHash,
+    pubsubTopic: PubsubTopic,
+    msg: WakuMessage,
+) {.async.} =
+  let insertStartTime = getTime().toUnixFloat()
+
+  (await self.driver.put(msgHash, pubsubTopic, msg)).isOkOr:
+    waku_archive_errors.inc(labelValues = [insertFailure])
+    trace "failed to insert message",
```
gabrielmer, line 130: shouldn't we log the failures at `debug` or `error` level? or it's something that happens a lot?

gabrielmer (approved):
LGTM, thanks so much!

waku/waku_archive/archive.nim
SionoiS: good catch probably a copy-pasta mistake.

waku/waku_archive/archive.nim
SionoiS: yeah potentially a lot of messages so I went with trace.

darshankabariya (approved):
LGTM !

PR #3137, opened 2024-10-22
chore: updating available procs in golang example

gabrielmer:
# Description
Updating the golang example with the the newly exposed procedures.

For consistency, copy-pasted the definitions from status-go, which implies the change of some variable names.

# Changes

- [x] updated the golang example's available procedures

## How to test
In nwaku's root directory, build libwaku and run the example with the following commands
```
make libwaku
cd examples/golang/
go run waku.go
```

### DIFF
--- a/examples/golang/waku.go
+++ b/examples/golang/waku.go
@@ -1,7 +1,7 @@
 package main
 
 /*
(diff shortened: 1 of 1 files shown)


richard-ramos (approved):

Ivansete-status (approved):
Awesome! Thanks so much indeed! :rocket:

PR #3145, opened 2024-10-24
chore: dbconn - make sure fd is not registered before registering again

Ivansete-status:
# Description
A node operator came across an issue. I could not replicate locally but this will help to avoid exception while registering a fd in dbconn.


Ivansete-status:
closing as not needed for now

PR #3146, opened 2024-10-24
fix: linting error

gabrielmer:
# Description
I think this is the simplest possible PR lol

It fixes a linting error we get when running nph in the CI


richard-ramos (approved):
This PR is too difficult to review! It has to be splitted into separate PRs

(lgtm :) )

PR #3152, opened 2024-10-25
chore: Enhancement on building nph and made it available naturaly on the path

NagyZoltanPeter:
# Description
Due that vscode nph extension is taking nph binary from the path, to avoid possible discrepancies - maybe of having other nph version avail from somewhere else - this chore aims to build nph binary only if needed and move it next to nim binary - into vendor/nimbus-build-system/vendor/Nim/bin/ directory, which is inherently added to the path with the .env.sh script.

# Changes

Only Makefile changes
- [x] build-nph target only executes if nph binary not exists 
- [x] built nph binary moved next to used nim binary under vendor inside repo
- [x] make update will also build nph binary for vscode extension convenience.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,6 @@ BUILD_SYSTEM_DIR := vendor/nimbus-build-system
 EXCLUDED_NIM_PACKAGES := vendor/nim-dnsdisc/vendor
 LINK_PCRE := 0
 LOG_LEVEL := TRACE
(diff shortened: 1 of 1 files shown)


gabrielmer (approved):
LGTM, thank you!

richard-ramos (approved):

Ivansete-status (approved):
LGTM! Thanks! :100:

PR #3153, opened 2024-10-28
feat: change latency buckets

fryorcraken:
~70% of observed values are under 100ms so usefult to split it further.

Less than 10% of observed values are above 400ms so no need to have that much precision.


vpavlin (approved):
lgtm, thanks

SionoiS (approved):
LGTM

PR #3154, opened 2024-10-28
chore: support ping with multiple multiaddresses and close stream

richard-ramos:
# Description
We've noticed before in status-go that sometimes multiaddresses that use dns names fail due to the ISP dns or the nameservers not being available. A change was done before in status-go to instead of having a single multiaddress per storenode, to have multiple multiaddresses: one with the dns and another one with the IP address.

As such, when go-waku does a ping, it does it by creating a `PeerInfo` (a go-libp2p structure) which contains these multiaddresses and then go-libp2p would use its own heurisitc to determine which multiaddress to use when pinging.

Looking at nim-libp2p, `switch.dial` supports multiple addresses too, so this PR modifies part of the nwaku code to make it possible to ping a peer with multiple multiaddresses

It also fixes a stream that was left open.

### DIFF
--- a/library/waku_thread/inter_thread_communication/requests/ping_request.nim
+++ b/library/waku_thread/inter_thread_communication/requests/ping_request.nim
@@ -1,4 +1,4 @@
-import std/json
+import std/[json, strutils]
 import chronos, results
 import libp2p/[protocols/ping, switch, multiaddress, multicodec]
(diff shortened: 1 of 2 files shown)


gabrielmer (approved):
Amazing! Thanks so much!

PR #3157, opened 2024-10-29
chore: naming connect procedure

gabrielmer:
# Description
Looking deeper at the code, I noticed that the `connectRelay` procedure uses the `nim-libp2p` `connect` procedure that connects to a node without setting a protocol. `connectRelay` was a misnomer and it's more correct to call it`connectPeer` 

# Changes

- [x] renaming `connectRelay` to `connectPeer`
- [x] removing checks for `Relay` being configured before calling `connectPeer`
- [x] in `connectPeer`, add peer if it's not found in the peer store instead of looking specifically for Waku Relay's codec

### DIFF
--- a/waku/factory/node_factory.nim
+++ b/waku/factory/node_factory.nim
@@ -430,9 +430,6 @@ proc startNode*(
       error "error while fetching peers from peer exchange", error = error
       quit(QuitFailure)
 
-  #[   if conf.dnsDiscovery and dynamicBootstrapNodes.len == 0:
-    retrieveDynamicBootstrapNodes() ]#
-
   # Start keepalive, if enabled
   if conf.keepAlive:
     node.startKeepalive()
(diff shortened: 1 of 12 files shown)


waku/factory/node_factory.nim
```diff
@@ -402,18 +402,12 @@ proc startNode*(
...
     try:
       await connectToNodes(node, conf.staticnodes, "static")
     except CatchableError:
       return err("failed to connect to static nodes: " & getCurrentExceptionMsg())
 
   if dynamicBootstrapNodes.len > 0:
-    if not conf.relay:
-      return err(
-        "waku relay (--relay=true) should be set when configuring dynamicBootstrapNodes"
-      )
```
gabrielmer, line 416: We don't need these conditions that I added. When I added them I thought that `connectToNodes`, which called `connectRelay`, were only related to Relay protocol but the function name was a misnomer

Ivansete-status (approved):
LGTM! Thanks for it! :100:

richard-ramos (approved):

PR #3158, opened 2024-10-29
chore: Adding lpt-runner script and assemble into liteprotocoltester image - to ease infra deployment

NagyZoltanPeter:
# Description
In agreement with infra team to ease the deployment of the liteprotocoltester it is better to embed the specific runner script (which existed in lpt-runner repository) right into the image. 
This is due they will use their own docker-compose setup thus by modifying the entrypoint and command, it is easier to run the official liteprotocoltester image directly.

# Changes

- [x] moved run_tester_node.sh from lpt-runner to nwaku as run_tester_node_on_fleet.sh
- [x] adapt docker build files to assemble new script into the image. Original entrypoint is not updated.

### DIFF
--- a/apps/liteprotocoltester/run_tester_node_on_fleet.sh
+++ b/apps/liteprotocoltester/run_tester_node_on_fleet.sh
@@ -1,0 +1,104 @@
+#!/bin/sh
+
+#set -x
+#echo "$@"
+
+if test -f .env; then
+  echo "Using .env file"
+  . $(pwd)/.env
+fi
+
+
+echo "I am a lite-protocol-tester node"
+
+BINARY_PATH=$1
+
+if [ ! -x "${BINARY_PATH}" ]; then
+  echo "Invalid binary path '${BINARY_PATH}'. Failing"
+  exit 1
+fi
+
+if [ "${2}" = "--help" ]; then
+  echo "You might want to check nwaku/apps/liteprotocoltester/README.md"
+  exec "${BINARY_PATH}" --help
+  exit 0
+fi
+
+FUNCTION=$2
+if [ "${FUNCTION}" = "SENDER" ]; then
+  FUNCTION=--test-func=SENDER
+  SERIVCE_NODE_ADDR=${LIGHTPUSH_SERVICE_PEER:-${LIGHTPUSH_BOOTSTRAP:-}}
+  NODE_ARG=${LIGHTPUSH_SERVICE_PEER:+--service-node="${LIGHTPUSH_SERVICE_PEER}"}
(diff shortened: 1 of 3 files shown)


gabrielmer (approved):
Thank you!

apps/liteprotocoltester/run_tester_node_on_fleet.sh
```diff
@@ -0,0 +1,104 @@
...
+echo "I am a lite-protocol-tester node"
+
+BINARY_PATH=$1
+
+if [ ! -x "${BINARY_PATH}" ]; then
+  echo "Invalid binary path '${BINARY_PATH}'. Failing"
+  exit 1
+fi
+
+if [ "${2}" = "--help" ]; then
```
Ivansete-status, line 21: I wonder if we should also show the following help in case no parameter is passed to the script

apps/liteprotocoltester/run_tester_node_on_fleet.sh
```diff
@@ -0,0 +1,104 @@
...
+  NODE_ARG=${NODE_ARG:---bootstrap-node="${LIGHTPUSH_BOOTSTRAP}"}
+  METRICS_PORT=--metrics-port="${PUBLISHER_METRICS_PORT:-8003}"
+fi
+
+if [ "${FUNCTION}" = "RECEIVER" ]; then
+  FUNCTION=--test-func=RECEIVER
+  SERIVCE_NODE_ADDR=${FILTER_SERVICE_PEER:-${FILTER_BOOTSTRAP:-}}
+  NODE_ARG=${FILTER_SERVICE_PEER:+--service-node="${FILTER_SERVICE_PEER}"}
+  NODE_ARG=${NODE_ARG:---bootstrap-node="${FILTER_BOOTSTRAP}"}
+  METRICS_PORT=--metrics-port="${RECEIVER_METRICS_PORT:-8003}"
```
Ivansete-status, line 41: Shall we use a separate default port in case sender and receiver are run from the same machine?

Ivansete-status (approved):
LGTM! Thanks for it! :100: 
I just added two nitpick comments ;P

apps/liteprotocoltester/run_tester_node_on_fleet.sh
NagyZoltanPeter: That is the default entry point of the docker image as it expect you to add arguments.

apps/liteprotocoltester/run_tester_node_on_fleet.sh
NagyZoltanPeter: Well, not really necessary as it is the responsibility of infra-misc deployment settings (they requested it btw). So they will configure it for sure.
While running under lpt-runner docker compose env it cause no issue as prometheus is embedded anyway.

PR #3162, opened 2024-10-30
feat: remain windows support

darshankabariya:
Will add all remain windows support with this PR


scripts/windows_setup.sh
```diff
@@ -32,11 +32,20 @@ build_component() {
...
     local name="$3"
-    
+
     echo "Building $name"
     if [ -d "$dir" ]; then
         change_directory "$dir"
         execute_command "$command"
+        # Execute command but ignore errors containing 'upx'
+        if ! eval "$command" 2>&1 | tee /tmp/build.log; then
+            if grep -q "upx" /tmp/build.log; then
```
Ivansete-status, line 42: Should we remove the `/tmp/build.log` afterwards?

Ivansete-status (approved):
LGTM! Thanks for it! :100:

scripts/windows_setup.sh
darshankabariya: sorry for late reply, it's temporary, and we will remove it eventually

waku/common/databases/db_postgres/dbconn.nim
```diff
@@ -182,12 +182,15 @@ proc waitQueryToFinish(
...
-      return err("failed to remove event reader in waitQueryToFinish: " & $error)
+  when not defined(windows):
+    asyncengine.addReader2(asyncFd, onDataAvailable).isOkOr:
+      dbConnWrapper.futBecomeFree.fail(newException(ValueError, $error))
+      return err("failed to add event reader in waitQueryToFinish: " & $error)
+    defer:
+      asyncengine.removeReader2(asyncFd).isOkOr:
+        return err("failed to remove event reader in waitQueryToFinish: " & $error)
+  else:
+    return err("Postgres not supported on Windows")
```
NagyZoltanPeter, line 193: Just out of curiosity... why is that or what it means here?

darshankabariya: we add this if condition bcz nim's postgres support is still not available on windows. @Ivansete-status right ?

scripts/build_wakunode_windows.sh
```diff
@@ -0,0 +1,60 @@
...
+}
+
+echo "0. -.-.-.-- Set PATH -.-.-.-"
+export PATH="/c/msys64/usr/bin:/c/msys64/mingw64/bin:/c/msys64/usr/lib:/c/msys64/mingw64/lib:$PATH"
+
+echo "1. -.-.-.- Verify dependencies -.-.-.-"
+execute_command "which gcc g++ make cmake cargo upx rustc python"
+
+echo "1. -.-.-.- Updating submodules -.-.-.-"
+execute_command "git submodule update --init --recursive"
```
gabrielmer, line 27: The numbering is wrong I think, it should probably be 1, 2 and then add 1 to all the steps that come next

gabrielmer (approved):
LGTM, thanks so much! 🤩

NagyZoltanPeter (approved):
Awesome work @darshankabariya!!! Huge step to multiplatform nwaku!
I have no objection on this.
Maybe some questions:
- Will it be possible to support postgres in the future on win?
- Does setup script make possible to build other targets also?
- Is it possible to build a Window's based docker image that can embed a full dev environment without any user pre-requisite?

darshankabariya:
> Awesome work @darshankabariya!!! Huge step to multiplatform nwaku! I have no objection on this. Maybe some questions:
> 
> * Will it be possible to support postgres in the future on win?
> * Does setup script make possible to build other targets also?
> * Is it possible to build a Window's based docker image that can embed a full dev environment without any user pre-requisite?

1)- We can plan something after CI integration.
2)- The goal is to support all targets via Makefile, but for now, we’re focused on libwaku and wakunode2. Instead of 20 separate scripts, the aim is to set up Windows requirements so all targets can build using make
3)- I want to make the process feel the same across platforms, but without MSYS, Windows support is tricky. @Ivansete-status says it’s not a priority, but the end goal is a smooth experience like Linux and macOS. 

thanks for asking amd if you have any suggestion i am more than happy give a try !

PR #3166, opened 2024-11-01
feat: Add incentivization PoC for RLNaaS in Lightpush

s-tikhomirov:
# Description

This WIP PR implements a PoC for light protocol incentivization as outlined in:
- [Incentivization spec](https://github.com/waku-org/specs/blob/master/standards/core/incentivization.md)
- [Incentivized Lightpush - Vac post](https://forum.vac.dev/t/incentivized-lightpush-a-roadmap-to-waku-service-incentivization/322)

# Changes

- [x] cherry-pick prior PoC work from [an experimental branch](https://github.com/waku-org/nwaku/tree/feat/incentivization-poc)
- [x] remove the dummy protocol (not needed anymore)
- [x] implement the required transaction checks (payment amount, address)
- [x] avoid hard-coded parameters (in the main code, not in tests)
- [ ] track used txids server-side to avoid double spend
- [ ] integrate with Lightpush

Some items (especially those closer to the end of this list) may be moved to later PRs to keep change sets relatively small.


Ivansete-status:
@s-tikhomirov - is there anything blocking this PR?

s-tikhomirov:
> @s-tikhomirov - is there anything blocking this PR?

Good question.

On the one hand, there are still things to implement before this can be considered a fully-fledged PoC. For example, the properties of a transaction that is checked for "eligibility" are now simply hard-coded. In the final vision, those should be node-specific and stored in a config file or something like that (and I'd have to figure out where such files should be located, how to manage them, etc). On the other hand, the current version is more or less self-contained, and in the spirit of merging things incrementally, I could see this PR at least considered for review.

For context: the overall vision of this PoC is that a client would attach a txid as proof of payment alongside its Lightpush request, and the server would check for eligibility of that txid. By eligibility we mean that the transaction:
- has been confirmed;
- is not a contract creation transaction;
- pays the expected amount;
- pays to the expected destination;
- hasn't been used in a previous request.

 Here is what is implemented now:
- check that the tx has been confirmed;
- check that the tx is a simple transfer (not a contract creation and not a contract call);
- check that the amount is "expected" (a hard-coded amount from a totally random tx used in tests);
- check that the destination is "expected" (also hard-coded, same as with the amount).

These could be the next steps:
- avoid hard-coded values (use server-specific expected amounts and destinations);
- use token transfer transactions and not simple transfers (as described [here](https://forum.vac.dev/t/incentivized-lightpush-a-roadmap-to-waku-service-incentivization/322); I'm using simple transfers now for simplicity);
- account for double-spent txids;
- avoid hard-coded Infura endpoint for querying the blockchain;
- attach this to Lightpush (behind a feature flag).

Given this, do you think it makes sense to consider this PR for review as it stands now?

@jm-clius , WDYT?

jm-clius:
Without having looked at the code itself yet, based on your description I would certainly suggest opening for review - or even selecting a smaller increment that illustrates a specific function and creating a PR for that. This way we can show the POC growing, remain clear about next steps and what is WIP, while making the turnaround from PR to merge shorter due to the easier review burden on nwaku contributors.

tests/incentivization/test_poc.nim
```diff
@@ -0,0 +1,59 @@
...
+const TxHashContractCreation* =
+  TxHash.fromHex("0xa2e39bee557144591fb7b2891ef44e1392f86c5ba1fc0afb6c0e862676ffd50f")
+const TxHashContractCall* =
+  TxHash.fromHex("0x2761f066eeae9a259a0247f529133dd01b7f57bf74254a64d897433397d321cb")
+const TxHashSimpleTransfer* =
+  TxHash.fromHex("0xa3985984b2ec3f1c3d473eb57a4820a56748f25dabbf9414f2b8380312b439cc")
+
+const EthClient = "https://sepolia.infura.io/v3/470c2e9a16f24057aee6660081729fb9"
+
+suite "Waku Incentivization PoC Eligibility Proofs":
```
Ivansete-status, line 28: Maybe we can add a short comment elaborating a bit the term "eligibility"

tests/incentivization/test_poc.nim
```diff
@@ -0,0 +1,59 @@
...
+const TxHashNonExisting* =
+  TxHash.fromHex("0x0000000000000000000000000000000000000000000000000000000000000000")
+const TxHashContractCreation* =
+  TxHash.fromHex("0xa2e39bee557144591fb7b2891ef44e1392f86c5ba1fc0afb6c0e862676ffd50f")
+const TxHashContractCall* =
+  TxHash.fromHex("0x2761f066eeae9a259a0247f529133dd01b7f57bf74254a64d897433397d321cb")
+const TxHashSimpleTransfer* =
+  TxHash.fromHex("0xa3985984b2ec3f1c3d473eb57a4820a56748f25dabbf9414f2b8380312b439cc")
+
+const EthClient = "https://sepolia.infura.io/v3/470c2e9a16f24057aee6660081729fb9"
```
gabrielmer, line 26: Not sure if leaking this might be a security concern, we could either have it as an env variable or use Anvil for testing like in RLN

tests/incentivization/test_poc.nim
```diff
@@ -0,0 +1,59 @@
...
+  
+  asyncTest "incentivization PoC: contract call tx is not eligible":
+    # note: assuming payment in native currency (ETH), not a token
+    let eligibilityProof =
+      EligibilityProof(proofOfPayment: some(@(TxHashContractCall.bytes())))
+    let txIsEligible = await isEligible(eligibilityProof, EthClient)
+    check:
+      not txIsEligible
+  
+  asyncTest "incentivization PoC: simple transfer tx is eligible":
```
Ivansete-status, line 52: nitpick comment:
Aren't "transfer tx" the same concepts :) ?

waku/incentivization/eligibility.nim
```diff
@@ -0,0 +1,6 @@
+import std/options, chronos
+
+import waku/incentivization/[rpc, txid_proof]
+
+proc isEligible*(eligibilityProof: EligibilityProof, ethClient: string): Future[bool] {.async.} =
```
Ivansete-status, line 5: Similar comment as before, I think it would help having a description of "elibible" concept
```suggestion
proc isEligible*(eligibilityProof: EligibilityProof, ethClient: string): Future[bool] {.async.} =
   ## We consider a tx to be eligible when...
```

waku/incentivization/txid_proof.nim
```diff
@@ -0,0 +1,62 @@
...
+
+
+# Function to convert a hex string to Address
+proc toAddress*(hexStr: string): Address =
+  # Remove the "0x" prefix if it exists
+  let cleaned = if hexStr.startsWith("0x"): hexStr[2..^1] else: hexStr
+
+  # Ensure the length is exactly 40 characters (20 bytes)
+  if cleaned.len != 40:
+    raise newException(ValueError, "Invalid hexadecimal string length for Address")
```
Ivansete-status, line 13: We tend to refrain from using exceptions. Instead, it is generally better to return a `Result[Address, string]` and then keep pushing the possible error up within the call stack
```suggestion
   return err("toAddress: invalid hexadecimal string length for Address: " & $(cleaned.len))
```

waku/incentivization/txid_proof.nim
```diff
@@ -0,0 +1,62 @@
...
+
+  var arr: array[20, byte]
+  for i in 0 ..< 20:
+    let byteValue = cleaned[i * 2 ..< i * 2 + 2]  # Get two hex characters
+    arr[i] = byte(parseHexInt(byteValue))
+
+  result = Address(arr)
```
Ivansete-status, line 20: We should avoid the use of the implicit `result` attribute. It is generally better to explicitly return a value. See the suggestion considering that this proc return a `Result[Address, string]`
```suggestion
  return ok(Address(arr))
```

waku/incentivization/txid_proof.nim
```diff
@@ -0,0 +1,62 @@
...
+
+
+proc checkTxIdIsEligible(txHash: TxHash, ethClient: string): Future[bool] {.async.} =
+  let web3 = await newWeb3(ethClient)
+  try:
+    let tx = await web3.provider.eth_getTransactionByHash(txHash)
+    let txReceipt = await web3.getMinedTransactionReceipt(txHash)
+    result = true
+    if result:
```
Ivansete-status, line 29: Sorry but this sounds a bit weird to my dumb eyes :)
Besides and as mentioned above, it is recommended not to use the implicit `result` item.

gabrielmer (commented):
Amazing work! Thanks so much! 🤩 

I added some questions and comments, please let me know if I'm missing anything in any of them :)

I also see that the Lint CI job is failing, please try installing the latest version of https://github.com/arnetheduck/nph/tree/latest and integrate it with your IDE (let me know if you have any issues with it, I can help out!)

waku/incentivization/txid_proof.nim
jm-clius, line 20: No `result` returns, see my comment elsewhere.

waku/incentivization/txid_proof.nim
```diff
@@ -0,0 +1,62 @@
...
+          # check that the amount is "as expected" (hard-coded for now)
+          let txValue = tx.value
+          let hasExpectedValue = (txValue == 200500000000005063.u256)
+          # check that the to address is "as expected" (hard-coded for now)
+          let toAddress = toAddressOption.get()
+          let hasExpectedToAddress = (toAddress == toAddress("0x5e809a85aa182a9921edd10a4163745bb3e36284"))
+          result = true
+  except ValueError as e:
+    result = false
+  await web3.close()
```
Ivansete-status, line 52: This line can be moved to the top and make use of `defer` statement to make sure it will always be called whenever the program goes out of this scope
```suggestion
  defer:
    await web3.close()
```

waku/incentivization/txid_proof.nim
```diff
@@ -0,0 +1,62 @@
...
+
+  result = Address(arr)
+
+
+proc checkTxIdIsEligible(txHash: TxHash, ethClient: string): Future[bool] {.async.} =
+  let web3 = await newWeb3(ethClient)
+  try:
+    let tx = await web3.provider.eth_getTransactionByHash(txHash)
+    let txReceipt = await web3.getMinedTransactionReceipt(txHash)
+    result = true
```
jm-clius, line 28: No `result` returns. See comment elsewhere.

waku/incentivization/txid_proof.nim
jm-clius, line 29: Not sure I understand the conditional here, if `result` will always be true. The following code can just be called without any `if`?

waku/incentivization/txid_proof.nim
```diff
@@ -0,0 +1,62 @@
...
+          result = false
+        else:
+          # check that the amount is "as expected" (hard-coded for now)
+          let txValue = tx.value
+          let hasExpectedValue = (txValue == 200500000000005063.u256)
+          # check that the to address is "as expected" (hard-coded for now)
+          let toAddress = toAddressOption.get()
+          let hasExpectedToAddress = (toAddress == toAddress("0x5e809a85aa182a9921edd10a4163745bb3e36284"))
+          result = true
+  except ValueError as e:
```
Ivansete-status, line 50: Maybe not needed to handle that exception but let me recommend a typical approach that we tend to follow when handling exceptions :)
```suggestion
  except ValueError:
    let errorMsg = "checkTxIdIsEligible failure: " & getCurrentExceptionMsg()
    error "exception in checkTxIdIsEligible", error = $errorMsg
    return err($errorMsg)
```

waku/incentivization/txid_proof.nim
```diff
@@ -0,0 +1,62 @@
...
+  for i in 0 ..< 20:
+    let byteValue = cleaned[i * 2 ..< i * 2 + 2]  # Get two hex characters
+    arr[i] = byte(parseHexInt(byteValue))
+
+  result = Address(arr)
+
+
+proc checkTxIdIsEligible(txHash: TxHash, ethClient: string): Future[bool] {.async.} =
+  let web3 = await newWeb3(ethClient)
+  try:
```
jm-clius, line 25: `try...except` should preferably just be around the block of code that can actually raise the exception you want to catch.

waku/incentivization/txid_proof.nim
```diff
@@ -0,0 +1,62 @@
...
+    let tx = await web3.provider.eth_getTransactionByHash(txHash)
+    let txReceipt = await web3.getMinedTransactionReceipt(txHash)
+    result = true
+    if result:
+      # check that it is not a contract creation tx
+      let toAddressOption = txReceipt.to
+      let isContractCreationTx = toAddressOption.isNone
+      if isContractCreationTx:
+        result = false
+      else:
```
jm-clius, line 35: In most of these cases, it seems to me you don't need lots of nested `if...else` statements. You can simply return early on these conditions.

For example:
```
<normal code....>

if isContractCreationTx: return false

<continue normal code without else>
```

waku/incentivization/txid_proof.nim
```diff
@@ -0,0 +1,62 @@
...
+        # check that it is a simple transfer (not a contract call)
+        # a simple transfer uses 21000 gas
+        let gasUsed = txReceipt.gasUsed
+        let isSimpleTransferTx = (gasUsed == Quantity(21000))
```
jm-clius, line 39: It's better to extract these constants to proper `const` definitions top-level at the module (to avoid "magic numbers")

waku/incentivization/txid_proof.nim
```diff
@@ -0,0 +1,62 @@
...
+          let hasExpectedToAddress = (toAddress == toAddress("0x5e809a85aa182a9921edd10a4163745bb3e36284"))
+          result = true
+  except ValueError as e:
+    result = false
+  await web3.close()
+  result
+
+proc txidEligiblityCriteriaMet*(
+    eligibilityProof: EligibilityProof, ethClient: string
+): Future[bool] {.async.} =
```
Ivansete-status, line 57: This is great indeed!
Nevertheless, the more detail we can give to the caller, the better. Thankfully, we have the `Result` type which is a great tool for that :)
```suggestion
proc txidEligiblityCriteriaMet*(
    eligibilityProof: EligibilityProof, ethClient: string
): Future[Result[void, string]] {.async.} =
```

Then, you would be able to use this proc in the following manner:

```
(await txidEligiblityCriteriaMet( params )).isOkOr:
  error "The bla bla is not eligible", error = $error
  ...
```

Ivansete-status (commented):
Thanks for it!
I just added some nitpick comment that I hope you find useful

waku/incentivization/txid_proof.nim
```diff
@@ -0,0 +1,62 @@
...
+        let isSimpleTransferTx = (gasUsed == Quantity(21000))
+        if not isSimpleTransferTx:
+          result = false
+        else:
+          # check that the amount is "as expected" (hard-coded for now)
+          let txValue = tx.value
+          let hasExpectedValue = (txValue == 200500000000005063.u256)
+          # check that the to address is "as expected" (hard-coded for now)
+          let toAddress = toAddressOption.get()
+          let hasExpectedToAddress = (toAddress == toAddress("0x5e809a85aa182a9921edd10a4163745bb3e36284"))
```
jm-clius, line 48: These hardcoded values should only appear in the test files. In other words, in the code you should have `txValue` and `expectedAddress` as parameters which can then be hardcoded into the tests for now. Even in the test files, however, we should avoid random magic numbers appearing in code - rather have these extracted as top-level consts in the module.

waku/incentivization/common.nim
```diff
@@ -0,0 +1,13 @@
+import std/options
+
+import waku/incentivization/rpc
+
+proc genEligibilityStatus*(isEligible: bool): EligibilityStatus =
```
jm-clius, line 5: This seems to me to be interesting only for tests? In other words, this should form part of the test files as a helper function.

waku/incentivization/eligibility.nim
```diff
@@ -0,0 +1,6 @@
...
+  result = await txidEligiblityCriteriaMet(eligibilityProof, ethClient)
```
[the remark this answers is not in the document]
jm-clius: I think this can probable go to `common.nim` for now rather than create a new module.

jm-clius (changes requested):
Thanks! Added a couple of initial comments to address below.

waku/incentivization/rpc_codec.nim
```diff
@@ -1,6 +1,8 @@
 import std/options
 import ../common/protobuf, ../waku_core, ./rpc
 
+const DefaultMaxRpcSize* = -1
```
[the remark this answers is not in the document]
s-tikhomirov: A leftover from code pasted from elsewhere, removed it. Thanks for spotting!

waku/incentivization/txid_proof.nim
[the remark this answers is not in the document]
s-tikhomirov: Indeed, that's unnecessary, removed.

waku/incentivization/txid_proof.nim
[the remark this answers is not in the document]
s-tikhomirov: Good idea, but I'm a bit confused about the following.

First, the current `nwaku` apparently imports [nim-eth @ bb5cb6a](https://github.com/status-im/nim-eth/tree/bb5cb6a4d0dbb592b395ab41cda3fd3f1fdc443e) which doesn't exactly have the code you quote. That isn't a fundamental roadblock though, I found where the address conversion is defined in the  (older) version of `nim-eth` that `nwaku` uses ([here](https://github.com/status-im/nim-eth/blob/bb5cb6a4d0dbb592b395ab41cda3fd3f1fdc443e/eth/common/utils.nim#L21)), but now I face another issue (quoting my own message from a recent Discord [thread](https://discord.com/channels/1110799176264056863/1197844654218039387/1313413055752704021)).

I've looked up where the address type is defined in the version of `nim-eth` that `nwaku` uses - that type is called `EthAddress` and is defined at `eth/common/eth_types` [here](https://github.com/status-im/nim-eth/blob/bb5cb6a4d0dbb592b395ab41cda3fd3f1fdc443e/eth/common/eth_types.nim#L28) as `EthAddress* = array[20, byte]`. However, the address that I want to compare to a hard-coded value comes from a `ReceiptObject` defined in nim-web3 [here](https://github.com/status-im/nim-web3/blob/428b931e7c4f1284b4272bc2c11fca2bd70991cd/web3/ethtypes.nim#L184), and the type of the address contained in the receipt object is defined [here](https://github.com/status-im/nim-web3/blob/428b931e7c4f1284b4272bc2c11fca2bd70991cd/web3/ethtypes.nim#L30) as `Address* = distinct array[20, byte]`.

TLDR: there seem to be two incompatible (?) types that describe Ethereum addresses, and I'm not sure which one to use and how to convert one to another if it's necessary. Any suggestions?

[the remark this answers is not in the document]
gabrielmer: Oh apologies! At that point I thought we had already updated those dependencies but I was mistaken, they have not yet been updated.

If in the current versions of the dependencies used in nwaku the procedure that does exactly the same thing as what you are doing is in `nim-web3`, I would use it over here. Whenever we upgrade the dependencies, we will get an error and simply switch to the newer version that I referenced.

If in the current dependencies there isn't anything that does exactly the same as what you need, then I would keep the code that you wrote and leave a TODO comment so we know to switch it with the newer proc once we update `nim-eth`

tests/incentivization/test_poc.nim
s-tikhomirov: Added comments clarifying the context of this PoC.

tests/incentivization/test_poc.nim
s-tikhomirov: "Transfer" is an adjective here - as opposed to contract call and contract creation transactions. Added a comment clarifying this.

waku/incentivization/txid_proof.nim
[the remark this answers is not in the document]
s-tikhomirov: Thanks for helping me figure this out! Indeed, there is a `fromHex` function in `eth-web3`, I now use it instead of my own function (05912abe).

waku/incentivization/txid_proof.nim
s-tikhomirov: What would be the preferred way to use the value defined within a `try`-block? In this code, only querying the RPC can raise the exception, so ideally the `try`-block would only contain the two queries. But when I assign, for example, `let tx = await web3.provider.eth_getTransactionByHash(txHash)`, `tx` remains in scope only within the `try`-block. I could define `tx` before the `try`-block and then assign the value to it, but I'm not sure it's a good idea either to initialize a variable with a `none` value... (It should be an Option type then?)

waku/incentivization/common.nim
s-tikhomirov: My idea was, on the contrary, that this function (renamed to `new` in later commits) would be useful for future code that would use our eligibility logic to convert a boolean eligibility value into `EligibilityStatus` that will be attached to the response. Something like this:

```
let proof = ... # extract eligibility proof (not necessarily a txid) from the request
let isValid = isEligible(proof, ...) # check proof eligibility
let eligibilityStatus = new(EligibilityStatus, isValid)
let response = LightpushResponse(eligibilityStatus, ...) # generate response
return response # (send the response back to the client)
```

Does it make sense?

waku/incentivization/txid_proof.nim
s-tikhomirov: > `Future[Result[void, string]]`

Why `void` and not `bool`?

tests/incentivization/test_poc.nim
s-tikhomirov: Moved the URL to an environment variable (b3890ca8).

s-tikhomirov:
Re-requesting reviews after addressing the suggestions (one suggestion remains unresolved, but it shouldn't be a blocker).

waku/incentivization/txid_proof.nim
```diff
@@ -0,0 +1,76 @@
...
+import waku/incentivization/rpc
+
+const SimpleTransferGasUsed = Quantity(21000)
+
+proc eth_getTransactionByHash_async(
+    txHash: TxHash, web3: Web3
+): Future[TransactionObject] {.async.} =
+  await web3.provider.eth_getTransactionByHash(txHash)
+
+proc getMinedTransactionReceipt_async(
```
gabrielmer, line 12: In these two proc names, we don't really tend to add async to the name nor to use snake case.

Maybe we can name them `getTransactionByHash` and `getMinedTransactionReceipt`?

waku/incentivization/txid_proof.nim
```diff
@@ -0,0 +1,76 @@
...
+  await web3.getMinedTransactionReceipt(txHash)
+
+proc getTxAndTxReceipt(
+    txHash: TxHash, web3: Web3
+): Future[(TransactionObject, ReceiptObject)] {.async.} =
+  let txFuture = eth_getTransactionByHash_async(txHash, web3)
+  let receiptFuture = getMinedTransactionReceipt_async(txHash, web3)
+  let tx = await txFuture
+  let txReceipt = await receiptFuture
+  return (tx, txReceipt)
```
gabrielmer, line 24: I think this way is a bit more clear that it's being done in parallel

```suggestion
proc getTxAndTxReceipt(
    txHash: TxHash, web3: Web3
): Future[(TransactionObject, ReceiptObject)] {.async.} =
  let txFuture = eth_getTransactionByHash_async(txHash, web3)
  let receiptFuture = getMinedTransactionReceipt_async(txHash, web3)
  await allFutures(txFuture, receiptFuture)
  return (txFuture.read(), receiptFuture.read())

```

waku/incentivization/txid_proof.nim
```diff
@@ -0,0 +1,76 @@
...
+  except ValueError:
+    let errorMsg =
+      "Failed to set up a web3 provider connection: " & getCurrentExceptionMsg()
+    error "exception in isEligibleTxId", error = $errorMsg
+    return err($errorMsg)
+  var tx: TransactionObject
+  var txReceipt: ReceiptObject
+  let txHash = TxHash.fromHex(byteutils.toHex(eligibilityProof.proofOfPayment.get()))
+  try:
+    (tx, txReceipt) = waitFor getTxAndTxReceipt(txHash, web3)
```
gabrielmer, line 50: This proc is async, so we should use `await` instead of `waitFor`

```suggestion
    (tx, txReceipt) = await getTxAndTxReceipt(txHash, web3)
```

gabrielmer (commented):
Oooh looks great! 🤩 

I added comments on some last small details - overall looks amazing! Thanks so much!

lmk in case anything I proposed doesn't make sense :))

waku/incentivization/txid_proof.nim
s-tikhomirov: Hm, when I applied this suggestion, the [test](https://github.com/waku-org/nwaku/blob/d6ac598eb58b9358769f9c438e1003184c455617/tests/incentivization/test_poc.nim#L37) that checks that a non-existent transaction is not eligible fell into an infinite loop. It seems that the updated code tries to query the zero hash again and again, but the transaction with that hash is never there (as expected). What should be the proper handling of this scenario in terms of async operations?

gabrielmer: Ooh I just reproduced and saw the issue, apologies!

So my understanding is that `eth_getTransactionByHash_async` throws an exception and if we don't await for it we won't catch it immediately.

On the other hand, `getMinedTransactionReceipt_async` keeps retrying if it fails to get the receipt, so in this case we are awaiting for both futures to complete but `receiptFuture` never completes because it keeps retrying indefinitely.

I think we need to add a timeout in `getMinedTransactionReceipt_async` and don't allow it to keep retrying without ever stopping.`getMinedTransactionReceipt_async` can leave you hanging indefinitely because of the future not timing out, this is a issue you will potentially find in other usages of the proc - in this particular test we were lucky that the exception in the other procedure stopped it.

Anyway, in summary - current solution that passes the test also works, but be aware that `getMinedTransactionReceipt_async` not timing out is an issue that can leave you in an infinite loop in cases beyond this test. If by design you do want that behavior of infinite retries, then current solution works perfectly and the change I proposed is not relevant :) if you do want to add a timeout to `getMinedTransactionReceipt_async`, then the proposed solution should work once the timeout is in place

gabrielmer: Looks amazing!Thanks so so much! 😍

gabrielmer (approved):
Looks greatt! Thanks so much! 🤩

waku/incentivization/txid_proof.nim
```diff
@@ -0,0 +1,87 @@
...
+): Future[Result[void, string]] {.async.} =
+  ## We consider a tx eligible,
+  ## in the context of service incentivization PoC,
+  ## if it is confirmed and pays the expected amount to the server's address.
+  ## See spec: https://github.com/waku-org/specs/blob/master/standards/core/incentivization.md
+  if eligibilityProof.proofOfPayment.isNone:
+    return err("Eligibility proof is empty")
+  var web3: Web3
+  try:
+    web3 = await newWeb3(ethClient)
```
Ivansete-status, line 48: Creating a new instance of `web3` for every call to `isEligibleTxId` doesn't sound very efficient. I believe it is better to do that once, when the app starts, and only close the `web3` instance at the app's end.

I'd suggest creating a new type `EligibilityManager` that contains the `web3` instance.

waku/incentivization/txid_proof.nim
```diff
@@ -0,0 +1,87 @@
...
+    txAndTxReceipt.isOkOr:
+      return err("Failed to fetch tx or tx receipt")
+    (tx, txReceipt) = txAndTxReceipt.value()
+  except ValueError:
+    let errorMsg = "Failed to fetch tx or tx receipt: " & getCurrentExceptionMsg()
+    error "exception in isEligibleTxId", error = $errorMsg
+    return err($errorMsg)
+  # check that it is not a contract creation tx
+  let toAddressOption = txReceipt.to
+  if toAddressOption.isNone:
```
Ivansete-status, line 68: Nitpick comment: by convention, the "verbs" should always have parenthesis when using them
```suggestion
  if toAddressOption.isNone():
```

waku/incentivization/txid_proof.nim
```diff
@@ -0,0 +1,87 @@
...
+
+proc getTxAndTxReceipt(
+    txHash: TxHash, web3: Web3
+): Future[Result[(TransactionObject, ReceiptObject), string]] {.async.} =
+  let txFuture = getTransactionByHash(txHash, web3)
+  let receiptFuture = getMinedTransactionReceipt(txHash, web3)
+  await allFutures(txFuture, receiptFuture)
+  let tx = txFuture.read()
+  let txReceipt = receiptFuture.read()
+  if txReceipt.isErr:
```
Ivansete-status, line 30:
```suggestion
  if txReceipt.isErr():
```

waku/incentivization/txid_proof.nim
```diff
@@ -0,0 +1,87 @@
...
+    return err("Cannot get tx receipt")
```
Ivansete-status, line 31: In this particular case we can also append the error description
```suggestion
    return err("Cannot get tx receipt: " & txReceipt.error)
```

waku/incentivization/txid_proof.nim
```diff
@@ -0,0 +1,87 @@
...
+    eligibilityProof: EligibilityProof,
+    expectedToAddress: Address,
+    expectedValue: UInt256,
+    ethClient: string,
+): Future[Result[void, string]] {.async.} =
+  ## We consider a tx eligible,
+  ## in the context of service incentivization PoC,
+  ## if it is confirmed and pays the expected amount to the server's address.
+  ## See spec: https://github.com/waku-org/specs/blob/master/standards/core/incentivization.md
+  if eligibilityProof.proofOfPayment.isNone:
```
Ivansete-status, line 44:
```suggestion
  if eligibilityProof.proofOfPayment.isNone():
```

Ivansete-status (commented):
Thanks for it! I've added some nitpick comments that I hope you find useful
I'm happy to re-review later on :) Ping me if need any clarification :partying_face:

jm-clius (approved):
Approved in terms of functionality. Thanks! I think it's worth paying attention to the Nim/nwaku review comments before merging.

waku/incentivization/txid_proof.nim
```diff
@@ -0,0 +1,87 @@
...
+    let errorMsg = "Failed to fetch tx or tx receipt: " & getCurrentExceptionMsg()
+    error "exception in isEligibleTxId", error = $errorMsg
+    return err($errorMsg)
+  # check that it is not a contract creation tx
+  let toAddressOption = txReceipt.to
+  if toAddressOption.isNone:
+    # this is a contract creation tx
+    return err("A contract creation tx is not eligible")
+  # check that it is a simple transfer (not a contract call)
+  # a simple transfer uses 21000 gas
```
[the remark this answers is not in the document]
s-tikhomirov: Strictly speaking, we cannot.

The number `21000` comes from the Ethereum spec (yellow paper) as the gas cost that every transaction pays. A simple transfer pays just that, and a contract call pays extra costs on top.

Note: the Yellow paper is outdated, as is admitted in [its repo](https://github.com/ethereum/yellowpaper):

> The Yellow Paper is out of date. It reflects the Ethereum specification up to the [Shanghai](https://github.com/ethereum/execution-specs/blob/master/network-upgrades/mainnet-upgrades/shanghai.md) network upgrade, activated on the Ethereum mainnet at block 17_034_870 (April 2023). 
> It does not yet contain changes introduced by the Cancun upgrade.
> An alternative [Python Execution Layer specification](https://ethereum.github.io/execution-specs/) is actively maintained and up to date.

TBH, I don't know how to make sense of that last link. But I would assume that as long as the 21000 constant has remained unchanged throughout Ethereum's history (AFAIK), it's unlikely to be changed.

Another question is that if the payments in question happen on a rollup, technically, the rollup developers can tweak gas costs for their EVM implementation. Linea, for one, also uses 21000 [here](https://github.com/Consensys/linea-besu/blob/ec79c17bc6ec56a3b432318953cadb6edfa16e8f/ethereum/core/src/test/java/org/hyperledger/besu/ethereum/transaction/TransactionSimulatorTest.java#L92).

In summary, the assumption "a tx is a simple transfer if and only if it uses 21000 gas" is somewhat of a hack, but haven't come up with a better way to check whether the tx is a contract call or not.

waku/incentivization/common.nim
```diff
@@ -0,0 +1,9 @@
+import std/options, chronos
+
+import waku/incentivization/[rpc, txid_proof]
+
+proc new*(T: type EligibilityStatus, isEligible: bool): T =
```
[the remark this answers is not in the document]
s-tikhomirov: Done in 133a411b24812820971109df9c2ef247aec1a7d5.

waku/incentivization/txid_proof.nim
s-tikhomirov: I understand the idea high-level and it makes total sense. I started implementing it in ccec7f8e (for now, it replicates the current behavior, just with the `EligibilityManager` instance on top).

I'm not sure about the next steps though. How to make the manager start "when the app starts" and close "at the app's end"? Where are the app's start and end defined? Do we have similar examples in our codebase I should look into? I'd note also that this PR doesn't intend to integrate the funcitonality into the app itself; all eligibility-related functions are only called from tests. Is this detail relevant?

I'd appreciate any pointers!

Ivansete-status: Thanks for that @s-tikhomirov !
I think these are the best points to consider:
1. The `EligibilityManager` (or `IntentivizationManager`) should be part of: https://github.com/waku-org/nwaku/blob/09f05fefe2614d640277f96ae96ab4f2f12c8031/waku/factory/waku.nim#L53
2. The Waku instance is started in: https://github.com/waku-org/nwaku/blob/09f05fefe2614d640277f96ae96ab4f2f12c8031/waku/factory/waku.nim#L354
3. The Waku instance is stopped in: https://github.com/waku-org/nwaku/blob/09f05fefe2614d640277f96ae96ab4f2f12c8031/waku/factory/waku.nim#L391

( cc @jm-clius, in case any of the above comments don't follow a proper design )

waku/incentivization/txid_proof.nim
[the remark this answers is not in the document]
s-tikhomirov: Fixed as of ccec7f8eb506997d155ecc0168764a1613351fec

waku/incentivization/txid_proof.nim
```diff
@@ -0,0 +1,100 @@
...
+  let receiptFuture = getMinedTransactionReceipt(txHash, web3)
+  await allFutures(txFuture, receiptFuture)
+  let tx = txFuture.read()
+  let txReceipt = receiptFuture.read()
+  if txReceipt.isErr():
+    return err("Cannot get tx receipt: " & txReceipt.error)
+  return ok((tx, txReceipt.get()))
+
+proc isEligibleTxId*(
+    eligibilityProof: EligibilityProof,
```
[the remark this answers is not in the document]
s-tikhomirov: Done in [d00e782](https://github.com/waku-org/nwaku/pull/3166/commits/d00e78242a3d1289bf9d9c95c3f25393cbdf896d)

waku/incentivization/txid_proof.nim
```diff
@@ -5,25 +5,39 @@ import waku/incentivization/rpc
...
 
+type EligibilityManager* = ref object
+  web3: Web3
+
+# Initialize the eligibilityManager with a web3 instance
+proc init*(
+    T: type EligibilityManager, ethClient: string
+): Future[EligibilityManager] {.async.} =
+  result = EligibilityManager(web3: await newWeb3(ethClient))
+  # TODO: handle error if web3 instance is not established
```
s-tikhomirov, line 16: Should we handle the potential error here, or wherever this procedure is called? From what I see in the existing codebase, calls to `newWeb3` are not wrapped in a `try` if called from within a test (and we only use eligibility-related logic in tests so far).

waku/incentivization/txid_proof.nim
s-tikhomirov: Eligibility manager is introduced in d00e78242a3d1289bf9d9c95c3f25393cbdf896d. The manager is only used in tests, and is created and destroyed in every test. I'd suggest we postpone the integration to `[nwaku/waku/factory/waku.nim](https://github.com/waku-org/nwaku/blob/09f05fefe2614d640277f96ae96ab4f2f12c8031/waku/factory/waku.nim)` until next PR(s) that would deal with the actual usage of eligibility manager in the app.

tests/incentivization/test_poc.nim
```diff
@@ -0,0 +1,79 @@
...
+const TxHashContractCall* =
+  TxHash.fromHex("0x2761f066eeae9a259a0247f529133dd01b7f57bf74254a64d897433397d321cb")
+const TxHashSimpleTransfer* =
+  TxHash.fromHex("0xa3985984b2ec3f1c3d473eb57a4820a56748f25dabbf9414f2b8380312b439cc")
+const ExpectedToAddress = Address.fromHex("0x5e809a85aa182a9921edd10a4163745bb3e36284")
+const ExpectedValue = 200500000000005063.u256
+
+# To set up the environment variable (replace Infura with your provider if needed):
+# $ export WEB3_RPC_URL="https://sepolia.infura.io/v3/YOUR_API_KEY"
+const EthClient = os.getEnv("WEB3_RPC_URL")
```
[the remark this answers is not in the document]
s-tikhomirov: Does using Anvil here imply that I would have to populate the local blockchain with necessary transactions? This is how I envision the pre-test setup:

1. using Anvil, deploy a simple contract with a callable function to the local blockchain;
2. issue a transaction that calls the contract;
3. issue a transaction that does a simple transfer (with correct and wrong parameters: amount and receiver address);
4. extract the factual transaction IDs (and amount and receiver, when necessary);
5. use those txids and other parameters in tests instead of [hard-coded txids extracted from Sepolia](https://github.com/waku-org/nwaku/blob/d00e78242a3d1289bf9d9c95c3f25393cbdf896d/tests/incentivization/test_poc.nim#L12-L21), which are used currently.

Does the plan look reasonable?

If so, what should I use a deployed contract?

Ideally, I'd like to reuse some contracts that are deployed already as part of the tests ([here](https://github.com/waku-org/nwaku/blob/d932dd10cca054faa0ec431c1b42ae02986dcb17/tests/waku_rln_relay/utils_onchain.nim#L65)) - should I attempt it, or should I add a "hello world" contract just for my tests, to avoid unnecessary dependencies between tests?

Ivansete-status (approved):
LGTM! Thanks for it! :100: 
I'm just adding some nitpick comments that I hope you find useful :)

tests/incentivization/test_poc.nim
```diff
@@ -0,0 +1,178 @@
...
+
+  echo "SETTING UP CONTRACT"
+
+  web3.defaultAccount = accounts[0]
+  let ExampleStorageContract =
+    "6080604052348015600e575f80fd5b506101438061001c5f395ff3fe608060405234801561000f575f80fd5b5060043610610034575f3560e01c80632e64cec1146100385780636057361d14610056575b5f80fd5b610040610072565b60405161004d919061009b565b60405180910390f35b610070600480360381019061006b91906100e2565b61007a565b005b5f8054905090565b805f8190555050565b5f819050919050565b61009581610083565b82525050565b5f6020820190506100ae5f83018461008c565b92915050565b5f80fd5b6100c181610083565b81146100cb575f80fd5b50565b5f813590506100dc816100b8565b92915050565b5f602082840312156100f7576100f66100b4565b5b5f610104848285016100ce565b9150509291505056fea26469706673582212209a0dd35336aff1eb3eeb11db76aa60a1427a12c1b92f945ea8c8d1dfa337cf2264736f6c634300081a0033"
+  let receipt = await web3.deployContract(ExampleStorageContract)
+  let txHashContractCreation = receipt.transactionHash
+
+  echo $txHashContractCreation
```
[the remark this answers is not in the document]
s-tikhomirov: These were temporary debug prints, removed in e2458e8c.

tests/waku_rln_relay/test_rln_group_manager_onchain.nim
```diff
@@ -116,15 +116,15 @@ suite "Onchain group manager":
     check triggeredError
 
   asyncTest "should error when keystore path and password are provided but file doesn't exist":
-    let manager = await setup()
+    let manager = await setupOnchainGroupManager()
```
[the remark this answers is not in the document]
s-tikhomirov: This particular change is in the existing test files, not in the eligibility-related test files I'm working on. I found `setup()` function name somewhat too generic, and renamed it to `setupOnchainGroupManager` so that the name better reflects its essense.

waku/incentivization/txid_proof.nim
```diff
@@ -0,0 +1,91 @@
...
+): Future[Result[void, string]] {.async.} =
+  ## We consider a tx eligible,
+  ## in the context of service incentivization PoC,
+  ## if it is confirmed and pays the expected amount to the server's address.
+  ## See spec: https://github.com/waku-org/specs/blob/master/standards/core/incentivization.md
+  if eligibilityProof.proofOfPayment.isNone():
+    return err("Eligibility proof is empty")
+  var tx: TransactionObject
+  var txReceipt: ReceiptObject
+  let txHash = TxHash.fromHex(byteutils.toHex(eligibilityProof.proofOfPayment.get()))
```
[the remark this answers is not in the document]
s-tikhomirov: This is necessary because `proofOfPayment` is a generic sequence of bytes that may or may not encode a transaction hash. In the current examples, transaction hashes as proof of payment, but it doesn't have to be this way. This line interprets a generic byte sequence as a transaction hash.

tests/incentivization/test_poc.nim
```diff
@@ -0,0 +1,178 @@
...
+
+  echo "TXHASHes:"
+  echo $txHashWrongReceiverRightAmount
+  echo $txHashRightReceiverWrongAmount
+  echo $txHashRightReceiverRightAmount
+
+  echo "SETTING UP CONTRACT"
+
+  web3.defaultAccount = accounts[0]
+  let ExampleStorageContract =
```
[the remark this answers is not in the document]
s-tikhomirov: Clarified in e2458e8.

tests/incentivization/test_poc.nim
```diff
@@ -0,0 +1,178 @@
...
+  var txHashContractCreation: TxHash
+
+  var receiverExpected: Address
+  var receiverNotExpected: Address
+
+  asyncTest "incentivization PoC: non-existent tx is not eligible":
+    ## Test that an unconfirmed tx is not eligible.
+
+    let eligibilityManager = await EligibilityManager.init(EthClient)
+
```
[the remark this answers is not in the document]
s-tikhomirov: Implemented in 570c8b45.

tests/incentivization/test_poc.nim
```diff
@@ -0,0 +1,178 @@
...
+    let eligibilityProof =
+      EligibilityProof(proofOfPayment: some(@(TxHashNonExisting.bytes())))
+    let isEligible = await isEligibleTxId(
+      eligibilityManager, eligibilityProof, receiverExpected, TxValueExpectedWei
+    )
+    check:
+      isEligible.isErr()
+    defer:
+      await eligibilityManager.close()
```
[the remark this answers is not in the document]
s-tikhomirov: Implemented in 570c8b45.

tests/waku_rln_relay/utils_onchain.nim
```diff
@@ -101,28 +102,51 @@ proc uploadRLNContract*(ethClientAddress: string): Future[Address] {.async.} =
...
-  tx.to = some(acc)
-  tx.gasPrice = some(gasPrice)
-
-  # Send ethAmount to acc
-  discard await web3.send(tx)
-  let balance = await web3.provider.eth_getBalance(acc, "latest")
-  assert balance == ethToWei(ethAmount),
-    fmt"Balance is {balance} but expected {ethToWei(ethAmount)}"
+  discard
+    await sendEthTransfer(web3, accounts[0], acc, ethToWei(amountEth), some(0.u256))
```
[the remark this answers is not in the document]
s-tikhomirov: Refactored somewhat in df7a118c, but we don't need the returned value in this use case anyway, so I kept the `discrard`.

s-tikhomirov:
Re-requesting reviews after addressing the comments.

Ivansete-status (approved):
LGTM! Thanks for it! :100:

tests/waku_rln_relay/utils_onchain.nim
```diff
@@ -206,15 +225,17 @@ proc setup*(
   let accounts = await web3.provider.eth_accounts()
   web3.defaultAccount = accounts[0]
 
-  var pk = none(string)
-  let (privateKey, _) = await createEthAccount(ethAmount)
-  pk = some($privateKey)
+  let (privateKey, acc) = createEthAccount(web3)
+
+  discard await sendEthTransfer(
```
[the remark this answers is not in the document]
s-tikhomirov: I've added a comment explaining the use of `discard` here: 76b75f96.

Note that `discard` [was used](https://github.com/waku-org/nwaku/blob/6d167dd2ed3dd74fb1fae9c699e1a41b9259e3eb/tests/waku_rln_relay/utils_onchain.nim#L122) here already. I just extracted some functionality into a separate `sendEthTransfer` procedure to reuse it in my code.

waku/incentivization/txid_proof.nim
jm-clius, line 1: Minor comment, but this module now defines the `type` and member functions for `EligibilityManager`, so could be renamed as `eligibility_manager.nim`

jm-clius (approved):
Thanks! One minor comment, but I think this can be merged.

jm-clius:
Btw, useful to add links to the associated deliverable in the PR description: https://github.com/waku-org/pm/issues/245

PR #3167, opened 2024-11-02
chore: libwaku - better error handling and better waku thread destroy handling

Ivansete-status:
## Description
While playing with [waku-rust-bindings](https://github.com/waku-org/waku-rust-bindings) I noticed that the `waku_destroy` function left a node running forever.

The test that helped to detect the issue is `cargo test default_echo -- --nocapture` from `waku-rust-bindings` repo.

This PR aims to enhance the destroy procedure and avoid deadlocks when `waku_destroy` is invoked.
Also, we carry on with the philosophy of zero acceptance to the use of the `discard` keyword ;P

( there was an nph issue in `apps/networkmonitor/networkmonitor_metrics.nim` )


richard-ramos (approved):

gabrielmer (approved):
LGTM, thanks so much!

PR #3171, opened 2024-11-07
refactor(libwaku): allow several multiaddresses for a single peer in store queries

richard-ramos:
nim-libp2p should choose then which multiaddr to use when dialing the peer. This is useful since for example in status-go the storenode list contains both the multiaddress with a DNS, and also another with an IP (for the cases in which for some reason the nameserver fail to resolve the dns name).

It also fixes a parameter name in the ping requests


gabrielmer (approved):
LGTM, thank you!

NagyZoltanPeter (approved):
LGTM! It's great, thank you.

PR #3173, opened 2024-11-08
chore: flaky rln test

### LINKED ISSUE #3095: bug: flaky RLN test
### Problem

The test `WakuNode - RLN relay::clearNullifierLog: should clear epochs > MaxEpochGap` is extremely flaky in the CI and constantly get this error

```
  /Users/runner/work/nwaku/nwaku/build/all_tests_waku 'WakuNode - RLN relay::clearNullifierLog: should clear epochs > MaxEpochGap'
----------------------------------------------------
    /Users/runner/work/nwaku/nwaku/tests/waku_rln_relay/test_wakunode_rln_relay.nim(585, 44): Check failed: node2.wakuRlnRelay.nullifierLog.len() == 4
    node2.wakuRlnRelay.nullifierLog.len() was 3
    /Users/runner/work/nwaku/nwaku/tests/waku_rln_relay/test_wakunode_rln_relay.nim(594, 44): Check failed: node2.wakuRlnRelay.nullifierLog.len() == 4
    node2.wakuRlnRelay.nullifierLog.len() was 3

  [FAILED ] (  41.8s) clearNullifierLog: should clear epochs > MaxEpochGap
```

### Expected behavior
The test should give consistent results

### nwaku version/commit hash
1d2b910f5467865316e331397394a79c29274ccd

darshankabariya:
Close #3095

Fix flaky behavior in RLN relay nullifier test.

### DIFF
--- a/tests/waku_rln_relay/test_wakunode_rln_relay.nim
+++ b/tests/waku_rln_relay/test_wakunode_rln_relay.nim
@@ -489,6 +489,14 @@ procSuite "WakuNode - RLN relay":
       node2 = newTestWakuNode(nodeKey2, parseIpAddress("0.0.0.0"), Port(0))
       epochSizeSec: uint64 = 5 # This means rlnMaxEpochGap = 4
 
+    # Helper function
+    proc waitForNullifierLog(node: WakuNode, expectedLen: int): Future[bool] {.async.} =
+      for i in 0 .. 10: # Try for up to 5 seconds (10 * 500ms)
+        if node.wakuRlnRelay.nullifierLog.len() == expectedLen:
+          return true
+        await sleepAsync(500.millis)
+      return false
+
     # Given both nodes mount relay and rlnrelay
     await node1.mountRelay(shardSeq)
     let wakuRlnConfig1 = buildWakuRlnConfig(1, epochSizeSec, "wakunode_10")
@@ -591,7 +599,7 @@ procSuite "WakuNode - RLN relay":
     await sleepAsync(publishSleepDuration)
     check:
       node1.wakuRlnRelay.nullifierLog.len() == 0
-      node2.wakuRlnRelay.nullifierLog.len() == 4
+      await waitForNullifierLog(node2, 4)
 
     # Then the node 2 should have cleared the nullifier log for epochs > MaxEpochGap
     # Therefore, with 4 max epochs, the first 4 messages will be published (except wm2, which shares epoch with wm1)


gabrielmer:
Thank you! Small question, why do we need to attempt many times until we get the value that we want?

Isn't there a way to make it more deterministic?

darshankabariya:
> Thank you! Small question, why do we need to attempt many times until we get the value that we want?
> 
> Isn't there a way to make it more deterministic?



> Thank you! Small question, why do we need to attempt many times until we get the value that we want?
> 
> Isn't there a way to make it more deterministic?

The nullifier log works like a capped queue with a maximum size of 4 epochs. When a new message arrives in epoch 5, the log needs to both remove the oldest message (from epoch 1) and add the new message. The flaky test occurs because we're checking the log length immediately after this operation - sometimes we catch it when it's only removed the old message (length = 3) but hasn't yet added the new one (should be 4). Instead of using a fixed wait time, we can make the test more reliable by actively waiting until the log reaches its expected length of 4, confirming both the removal of the old message and addition of the new one is complete. or we can just wait little longer ( like wait 10000 milis ) without loop as you suggest !

darshankabariya:
> Thank you! Small question, why do we need to attempt many times until we get the value that we want?
> 
> Isn't there a way to make it more deterministic?

The nullifier log works like a capped queue with a maximum size of 4 epochs. When a new message arrives in epoch 5, the log needs to both remove the oldest message (from epoch 1) and add the new message. The flaky test occurs because we're checking the log length immediately after this operation - sometimes we catch it when it's only removed the old message (length = 3) but hasn't yet added the new one (should be 4). Instead of using a fixed wait time, we can make the test more reliable by actively waiting until the log reaches its expected length of 4, confirming both the removal of the old message and addition of the new one is complete. or we can do like you suggest just wait little longer like ( 10000 milies ) then that help get rid this flakiness.

gabrielmer:
> The nullifier log works like a capped queue with a maximum size of 4 epochs. When a new message arrives in epoch 5, the log needs to both remove the oldest message (from epoch 1) and add the new message. The flaky test occurs because we're checking the log length immediately after this operation - sometimes we catch it when it's only removed the old message (length = 3) but hasn't yet added the new one (should be 4)

Got it, thank you! Makes sense, yes it's better as you did then! Better to catch earlier and don't wait more time than needed.

Last question then. If the queue has size 4 and a message is removed and the new one is added - isn't it possible for the queue to be size 4 and pass the test even before the message was even sent?

In other words, how do we now that this check is not done before we remove the message and pass without having done any operation?

darshankabariya:
> > The nullifier log works like a capped queue with a maximum size of 4 epochs. When a new message arrives in epoch 5, the log needs to both remove the oldest message (from epoch 1) and add the new message. The flaky test occurs because we're checking the log length immediately after this operation - sometimes we catch it when it's only removed the old message (length = 3) but hasn't yet added the new one (should be 4)
> 
> Got it, thank you! Makes sense, yes it's better as you did then! Better to catch earlier and don't wait more time than needed.
> 
> Last question then. If the queue has size 4 and a message is removed and the new one is added - isn't it possible for the queue to be size 4 and pass the test even before the message was even sent?
> 
> In other words, how do we now that this check is not done before we remove the message and pass without having done any operation?

Great observation! I completely agree – we should also ensure that the length of the nullifier remains consistent before and after. let's do it then.

gabrielmer (approved):
Thanks so much!

Ivansete-status (commented):
Awesome! Thanks for it! :raised_hands: 

I'm not approving it yet because we need to change that:

xasyncTest "clearNullifierLog: should clear epochs > MaxEpochGap":

to 

asyncTest "clearNullifierLog: should clear epochs > MaxEpochGap":

darshankabariya:
> Awesome! Thanks for it! 🙌
> 
> I'm not approving it yet because we need to change that:
> 
> xasyncTest "clearNullifierLog: should clear epochs > MaxEpochGap":
> to
> asyncTest "clearNullifierLog: should clear epochs > MaxEpochGap":

The test has already been excluded; I haven’t converted it to xasynctest yet. Are you suggesting that I should simply convert it to xasynctest for the PR to be approved, or is there something else blocking its approval ?

Ivansete-status:
Morning @darshankabariya et al!
The suggestion is to remove the `x` from `xasyncTest` so that the test is not skipped :)

darshankabariya:
> Morning @darshankabariya et al! The suggestion is to remove the `x` from `xasyncTest` so that the test is not skipped :)

Done! Could you plz review and approve it now?

Ivansete-status (approved):

gabrielmer (approved):

PR #3176, opened 2024-11-10
chore: add supervisor for lpt infra

NagyZoltanPeter:
# Description
This PR is based upon the previous one, and serves the same sole purpose: enable infra deployment.
Prev PR: https://github.com/waku-org/nwaku/pull/3158

# Changes

- After agreement with Marko: https://github.com/status-im/infra-misc/pull/380 we decided that infra deployment image must run continuously. As this is against how this test is designed it is now extended:
- Added lpt_supervisor.py script that takes default deployment arguments and a period time arg, and restarts tester apps in predefined times. (This has to be set carefully to allow both parties (SENDER/RECEIVER) to finish before next period comes.)
- infra.env : predefined deployment arguments to testing, built into the image
- if the image tag is set to deploy for docker build, it will create a specific image for infra deployment (having different entrypoint with lpt_supervisor.py)
- As a plus, as Franck requested, I added message latency metrics (latency was already reported but only to stdout).


gabrielmer (approved):
Thank you!

apps/liteprotocoltester/lpt_supervisor.py
```diff
@@ -0,0 +1,53 @@
...
+    for line in f:
+      if line.strip() and not line.startswith('#'):
+        key, value = line.strip().split('=', 1)
+        env_vars[key] = value
+  return env_vars
+
+def start_tester_node():
+  role = sys.argv[1]
+  env = os.environ.copy()
+  for key, value in env_vars.items():
```
markoburcul, line 20: Where do you forward env_vars dictionary? I can't find it defined globally

apps/liteprotocoltester/lpt_supervisor.py
```diff
@@ -0,0 +1,53 @@
...
+  script_cmd = ['sh', '/usr/bin/run_tester_node_at_infra.sh', '/usr/bin/liteprotocoltester', role]
+  process = Popen(script_cmd, env=env)
+  return process
+
+if __name__ == "__main__":
+  if len(sys.argv) < 2 or sys.argv[1] not in ["RECEIVER", "SENDER"]:
+    print("Error: First argument must be either 'RECEIVER' or 'SENDER'")
+    sys.exit(1)
+
+  env_file = '/usr/bin/infra.env'
```
markoburcul, line 32: Why are you hardcoding path to the env vars file? Wouldn't it be simpler to just check if the env var is defined and read it from the code?

apps/liteprotocoltester/Dockerfile.liteprotocoltester.compile
```diff
@@ -43,15 +42,30 @@ EXPOSE 30303 60000 8545
...
 
 ENTRYPOINT ["/usr/bin/run_tester_node.sh", "/usr/bin/liteprotocoltester"]
 
-# # By default just show help if called without arguments
-CMD ["--help"]
+# Image for infra deployment -------------------------------------------
+FROM base_lpt AS deployment_lpt
+
+COPY --from=nim-build /app/apps/liteprotocoltester/run_tester_node_at_infra.sh /usr/bin/
+COPY --from=nim-build /app/apps/liteprotocoltester/infra.env /usr/bin/
```
markoburcul, line 66: You can just define the env variables in this Dockerfile and set their default values. They can be changed if you pass them to the container when starting it.

Makefile
```diff
@@ -373,16 +373,16 @@ DOCKER_LPT_NIMFLAGS ?= -d:chronicles_colors:none -d:insecure
...
 	docker build \
-	  --no-cache \
 		--build-arg="MAKE_TARGET=liteprotocoltester" \
 		--build-arg="NIMFLAGS=$(DOCKER_LPT_NIMFLAGS)" \
 		--build-arg="NIM_COMMIT=$(DOCKER_NIM_COMMIT)" \
 		--build-arg="LOG_LEVEL=TRACE" \
 		--label="commit=$(shell git rev-parse HEAD)" \
 		--label="version=$(GIT_VERSION)" \
-		--target $(TARGET) \
+		--target $(if $(filter deploy,$(DOCKER_LPT_TAG)),deployment_lpt,standalone_lpt) \
```
Ivansete-status, line 385: what's the diff between `deployment_lpt` and `standalone_lpt`?

apps/liteprotocoltester/infra.env
```diff
@@ -0,0 +1,11 @@
+TEST_INTERVAL_MINUTES=10
+START_PUBLISHING_AFTER=120
+NUM_MESSAGES=300
+DELAY_MESSAGES=1000
```
Ivansete-status, line 4: Maybe we could have a more explicit var name;P
```suggestion
MSG_INTERVAL_MILLIS=1000
```

apps/liteprotocoltester/infra.env
```diff
@@ -0,0 +1,11 @@
+TEST_INTERVAL_MINUTES=10
+START_PUBLISHING_AFTER=120
```
Ivansete-status, line 2:
```suggestion
START_PUBLISHING_AFTER_SECS=120
```

Ivansete-status (approved):
LGTM! Thanks for it! :100: 
I just added some minor comments

Makefile
NagyZoltanPeter: It is the distinction between the different use. `deployment_lpt` stands for running in our infra environment with its own constraints. - main diff is that in deployment we need to run a supervisor ahead of tests that can control restarts.
`standalone_lpt` is the normal usage, so in lpt_runner you can easily run your tests on demand.

apps/liteprotocoltester/infra.env
NagyZoltanPeter: Good idea! I need to apply such change more deep. I will check!

apps/liteprotocoltester/infra.env
NagyZoltanPeter: Just as above, thanks for the suggestion.

Makefile
Ivansete-status: Thanks for clarifying! I was wondering if we could add a comment somewhere, if not done already:)

NagyZoltanPeter: Yeah must add it to lpt readme at least!

markoburcul (approved):

PR #3177, opened 2024-11-19
chore: Filter in libwaku

Ivansete-status:


darshankabariya (approved):
LGTM !

waku/factory/node_factory.nim
```diff
@@ -352,11 +352,11 @@ proc setupProtocols(
     except CatchableError:
       return err("failed to mount waku filter protocol: " & getCurrentExceptionMsg())
 
+  await node.mountFilterClient()
```
NagyZoltanPeter, line 355: Same with filter.

library/libwaku.h
```diff
@@ -92,6 +92,22 @@ int waku_relay_unsubscribe(void* ctx,
...
+int waku_filter_unsubscribe(void* ctx,
+                            const char* pubSubTopic,
+                            const char* contentTopics,
+                            WakuCallBack callback,
+                            void* userData);
+
+int waku_filter_unsubscribe_all(void* ctx,
+                                WakuCallBack callback,
+                                void* userData);
+
```
NagyZoltanPeter, line 110: I think we need ping as well! Is it left out intentionally?
Without ping the client is forced to use subscribe all the time (in each 1-2 minutes) which is significantly higher load on service node than just check with ping.

NagyZoltanPeter (changes requested):
I think we miss filter ping from the list of operation to support for the complete functionality. WDYT?

waku/factory/node_factory.nim
```diff
@@ -331,10 +331,10 @@ proc setupProtocols(
     except CatchableError:
       return err("failed to mount waku lightpush protocol: " & getCurrentExceptionMsg())
 
+  mountLightPushClient(node)
```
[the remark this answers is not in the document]
Ivansete-status: GM @NagyZoltanPeter - by default we need to mount all the clients. We are also following that approach in store-clients.

waku/factory/node_factory.nim
Ivansete-status: Yes indeed, good point!, and I agree that it may sound weird but it is handier from user PoV to always have the "client" components mounted so that is not needed an additional argument such as `filter-client = true`, or `lightpush-client = true`, or `store-client

library/libwaku.h
Ivansete-status: Thanks for that and very good point!
IMO, I think it is better to perform the ping internally and avoid the `libwaku` user handling that. I will submit another commit soon

gabrielmer (approved):
Looks amazing! Thanks so much! 😍

waku/factory/node_factory.nim
NagyZoltanPeter: Well, imho, something more robust configuration option maybe needed. Like --light-protocols (edge-protocols?)
then install filter/lightpush/store-clients. Also we can leave the option to specify the service node(s) for the client protocols - but non mandatory. I think it can be a good balance between ease of use and control over resources.

I just noticed that with this change REST API will not function as expected. filter and lightpush REST endpoints are initialized upon --filternode and --lightpushnode having value.
Then this needs to be added here to keep consistency.
WDYT @Ivansete-status ?

library/libwaku.h
NagyZoltanPeter: Do you mean, hide ping behind the scene? Let say ping every minute if any subscription is added and between unsubscribe_all?

waku/factory/node_factory.nim
Ivansete-status: @NagyZoltanPeter - yes it makes 100% sense to me to add that `--light-protocols` option. That's something we will need shortly. I think we can consider that in a separate PR.

library/libwaku.h
NagyZoltanPeter: I agree!

waku/factory/node_factory.nim
NagyZoltanPeter: Ok fine for me. I will approve this PR now!

PR #3179, opened 2024-11-20
chore: python example - call waku_setup to initialize libwaku properly

Ivansete-status:
# Description
Change motivated by @fbarbu15 
In this case, it was missing to invoke `waku_setup` so that the libwaku library is properly initialized and the Nim runtime and gc.


NagyZoltanPeter (approved):
LGTM!

PR #3184, opened 2024-11-28
chore: capping mechanism for relay and service connections

### LINKED ISSUE #3163: chore: Reserve % of max connections for relay
### Background
`max-connections` is used to set the maximum amount of concurrent connections that the switch accepts.
Currently, it is also being used to set the `max-relay-peers`, [here](https://github.com/waku-org/nwaku/blob/b4d3678ad943d60e96ccb14b4f28edac24474433/waku/node/peer_manager/peer_manager.nim#L994).
This allows the waku node to always have room for other protocols.
However, this is not something that happens in the other way around. This means that other protocols can fill the number of `max-connections`, disabling the ability of relaying messages for the waku node.

### Details
Given the following scenario, in the following order:
1. Setup a bootstrap only node.
2. Setup a filter relay node with `max-connections=5`
3. Setup 5 filter client nodes, connected to the filter relay node. This will cap the number of connections.
4. Setup another filter relay node, and connect them to the bootstrap node.

If we inject messages to the nodes in point 4, the node in point 1 will not be able to receive them, as it will not connect to those nodes due to the maximum number of connections reached.

### Acceptance criteria
In the same way that a node has always 20% connections for other protocols, should also have at least `d_low` connections reserved for relay as well. 
If not, there can happen a lot of bad situations that might be complex to find the root of the issue.
If by any chance a relay node has disconnections to other relay nodes and has less than `d_low` relay connections, and in the meantime new filter client nodes are connected, unless those clients are disconnected again, the node will stop relaying messages.
This can be checked because it will start logging that there are not enough peers to publish, but if it loses connections to ALL other relay nodes, it will not log anything, and it will stop receiving messages.

darshankabariya:
This PR prevents nodes from becoming isolated from the network by implementing a capping mechanism for service and relay peers. This ensures there is always room available for relay peers at any given time.

close #3163


waku/node/peer_manager/peer_manager.nim
```diff
@@ -993,6 +999,24 @@ proc new*(
     # Leave by default 20% of connections for service peers
     maxRelayPeersValue = maxConnections - (maxConnections div 5)
 
+  var maxServicePeersValue = 0
+  var d_low = 4
```
SionoiS, line 1003: I would vote for d high juste to be safe.

waku/node/peer_manager/peer_manager.nim
```diff
@@ -959,6 +964,7 @@ proc new*(
     switch: Switch,
     wakuMetadata: WakuMetadata = nil,
     maxRelayPeers: Option[int] = none(int),
+    maxServicePeers: Option[int] = none(int),
```
SionoiS, line 967: Is the peer manager already initialized with `maxServicePeers` somehow?

SionoiS (approved):
LGTM

waku/node/peer_manager/peer_manager.nim
darshankabariya: Yeah, it's not passed anywhere. Instead of making it a local variable, I thought it would be better to incorporate it into the peer_manager. This would showcase the functionality of peer_manager and also make it easier to add some unittests.

waku/node/peer_manager/peer_manager.nim
gabrielmer, line 1003: I think we shouldn't hardcode this and take it directly from `WakuRelay`

waku/node/peer_manager/peer_manager.nim
```diff
@@ -993,6 +999,24 @@ proc new*(
...
+      raise newException(
+        Defect,
+        "Max number of service peers can't be greater than the max amount of connections minus d_low",
+      )
+    maxServicePeersValue =
+      min(maxServicePeers.get(), maxConnections - maxRelayPeersValue)
+      # relay has higher priority then service peers
+  else:
+    maxServicePeersValue =
+      max(maxConnections - maxRelayPeersValue, maxConnections div 5)
```
gabrielmer, line 1018: what's the logic of this line? Why the `maxConnections div 5`?

gabrielmer (commented):
Looks great, thanks so much!

Added just a couple questions/comments

waku/node/peer_manager/peer_manager.nim
darshankabariya: My logic is to reserve 20% for service peers if the user does not provide maxService, as we discuss in nwaku pm we need to change whole logic

Ivansete-status (commented):
The PR doesn't compile snif snif :s
Re-thinking about this, I suggest doing the following:
We need to keep `maxConnections` config param because this needs to be passed to the nim-libp2p-_Switch_ object.
On the other hand, I would only add an additional parameter, named `protoConnWeights` , a.k.a. `proto-conn-weights`, with the following logic:

If `proto-conn-weights` is not passed, then we apply the internal limits of 70% for relay, and the other 30% is evenly distributed for other protocols.

If `proto-conn-weights=relay:80;services:20`, we apply the internal limits of 80% for relay, and the other 20% is evenly distributed for other protocols.

We need to make sure the relay limit is bigger than DHigh ( we can create a 
public `Constant` in `waku_relay/protocol.nim` ) https://github.com/waku-org/nwaku/blob/09f05fefe2614d640277f96ae96ab4f2f12c8031/waku/waku_relay/protocol.nim#L73

darshankabariya:
> The PR doesn't compile snif snif 

My bad, I pushed the code without a local build. I’ll update this PR until we finalize it.

darshankabariya:
> The PR doesn't compile snif snif :s Re-thinking about this, I suggest doing the following: We need to keep `maxConnections` config param because this needs to be passed to the nim-libp2p-_Switch_ object. On the other hand, I would only add an additional parameter, named `protoConnWeights` , a.k.a. `proto-conn-weights`, with the following logic:
> 
> If `proto-conn-weights` is not passed, then we apply the internal limits of 70% for relay, and the other 30% is evenly distributed for other protocols.
> 
> If `proto-conn-weights=relay:80;services:20`, we apply the internal limits of 80% for relay, and the other 20% is evenly distributed for other protocols.
> 
> We need to make sure the relay limit is bigger than DHigh ( we can create a public `Constant` in `waku_relay/protocol.nim` )
> 
> https://github.com/waku-org/nwaku/blob/09f05fefe2614d640277f96ae96ab4f2f12c8031/waku/waku_relay/protocol.nim#L73
> 
> wdyt @NagyZoltanPeter , @gabrielmer , @darshankabariya ?

From my perspective, the current setup might be a bit complex for users. At the moment, we have parameters like `maxConnections`, `maxRelayPeers`, `maxServicePeers`, and another one called `protoConnWeight`. This can make things more confusing, as users need to ensure that maxConnections is never greater than the sum of `maxRelayPeers` and `maxServicePeers`. Additionally, aligning `proto-conn-weights` with `maxRelayPeers` and `maxServicePeers` adds another layer of complexity.

Instead, I believe users generally don’t think in terms of absolute numbers for services and relays. From a psychological perspective, they might find it more intuitive to think in percentages—e.g., 80% for services, 20% for relays.

My suggestion is to simplify this structure. We should keep `maxConnections` bcz represent hardware capacity, which users are familiar with, and suggest deprecate `maxRelayPeers` and `maxServicePeers`. Instead, we can ask users to specify the percentage they want to allocate for relays and services. This approach feels more straightforward and user-friendly.

If users don’t provide this percentage, we could default to a ratio, as Ivan suggested. I believe this would simplify the experience while maintaining flexibility.

SionoiS:
I like your idea! IMO `maxConnection` and `relayToServiceRatio` (float) would be simpler.

I feel we still need hardcoded minimum still so that you can't, by mistake brick your node.

gabrielmer:
Same opinion here, having 2 params, a maximum and a ratio to know how many to allocate for each should make things simple yet customizable :)

darshankabariya:
> I like your idea! IMO `maxConnection` and `relayToServiceRatio` (float) would be simpler.
> 
> I feel we still need hardcoded minimum still so that you can't, by mistake brick your node.

> Same opinion here, having 2 params, a maximum and a ratio to know how many to allocate for each should make things simple yet customizable :)

Thanks @gabrielmer @SionoiS for reply ! 

If this approach looks good to you, I can start the implementation. Some unit tests might also need to be deprecated or modified based on our refactor.

Regarding your thoughts on the ratio, I have an idea: instead of using a float, we could represent it as a string, like 0.33 becoming 40:60 or 0.43 as 30:70. I feel using a float isn't very intuitive. Let me know what you think, and I’ll implement accordingly.

gabrielmer:
> 
> If this approach looks good to you, I can begin the implementation. Just to confirm, we're only discussing the deprecation of the external configuration parameter, correct? I believe we still need to maxRelayPeer and maxServicePeer within the peerManager.
> 
> Regarding your thoughts on the ratio, I have an idea: instead of using a float, we could represent it as a string, like 0.33 becoming 40:60 or 0.43 as 30:70. it's just need calculator 😂for some ratio, Let me know what you guys think, and I’ll implement accordingly.

Yes! That would only be for the external configuration, we would then translate it into `maxRelayPeers` and `maxServicePeers` when configuring the Peer Manager

Regarding the format, I'm pretty neutral about it - no strong opinions. Would like to hear what others think :))

SionoiS:
> Regarding your thoughts on the ratio, I have an idea: instead of using a float, we could represent it as a string, like 0.33 becoming 40:60 or 0.43 as 30:70. it's just need calculator 😂for some ratio, Let me know what you guys think, and I’ll implement accordingly.

If that's simpler go for it! No strong opinion here either.

darshankabariya:
Just leaving a comment to inform, this PR is ready to be merged. Thank you!

waku/common/utils/parse_size_units.nim
```diff
@@ -49,3 +49,24 @@ proc parseCorrectMsgSize*(input: string): uint64 =
...
+      return err("Invalid format for relay ratio expected relay:service")
+
+    let
+      relayRatio = parseInt(elements[0])
+      serviceRatio = parseInt(elements[1])
+
+    if relayRatio < 0 or serviceRatio < 0:
+      return err("Relay and service ratios must be non-negative integers")
+
+    let total = relayRatio + serviceRatio
```
Ivansete-status, line 66: I think is better to accept percentage values so that the user explicitly specifies the weighing. It can be either values between 0 and 100, or between 0 and 1.

For example, it sounds clearer, from user perspective, to accept the following format: `70:30`; `65:35`, etc

waku/factory/builder.nim
```diff
@@ -108,9 +110,20 @@ proc withPeerStorage*(
...
+    relayServiceRatio: string,
+    shardAware = false,
 ) =
-  builder.maxRelayPeers = maxRelayPeers
+  let relayServiceRatio = parseRelayServiceRatio(relayServiceRatio)
+  if relayServiceRatio.isErr:
+    warn "Invalid relay service ratio", ratio = relayServiceRatio
+    return
+
+  let (relayRatio, serviceRatio) = relayServiceRatio.get
```
Ivansete-status, line 123:
```suggestion
  let (relayRatio, serviceRatio) = parseRelayServiceRatio(relayServiceRatio).valueOr:
    error "Invalid relay service ratio", ratio = relayServiceRatio, error = error
    return ## we should stop the execution in this point but for further PR's
```

waku/node/peer_manager/peer_manager.nim
```diff
@@ -975,42 +987,59 @@ proc new*(
...
-    maxRelayPeersValue = maxRelayPeers.get()
-  else:
-    # Leave by default 20% of connections for service peers
-    maxRelayPeersValue = maxConnections - (maxConnections div 5)
+  let (relayRatio, serviceRatio) = relayServiceRatio.get
+
+  var relayPeers = int(ceil(float(maxConnections) * relayRatio))
+  var servicePeers = int(floor(float(maxConnections) * serviceRatio))
+
+  if maxRelayPeers.isSome() or maxServicePeers.isSome():
```
Ivansete-status, line 1000: Maybe not needed this `if` statement?

waku/node/peer_manager/peer_manager.nim
```diff
@@ -975,42 +987,59 @@ proc new*(
...
+      servicePeers = maxConnections - relayPeers
+    elif maxServicePeers.isSome():
+      servicePeers = maxServicePeers.get()
+      relayPeers = maxConnections - servicePeers
+
+  if relayPeers + servicePeers > maxConnections:
+    warn "Max number of connections can't be greater than maxRelayPeers + maxServicePeers",
+      maxConnections = maxConnections,
+      maxRelayPeers = relayPeers,
+      maxServicePeers = servicePeers
```
Ivansete-status, line 1015: This block shouldn't be needed because we should validate that on startup

waku/node/peer_manager/peer_manager.nim
```diff
@@ -975,42 +987,59 @@ proc new*(
...
+
+  let minRelayPeers = 8
```
Ivansete-status, line 1017: The value is correct nowadays:)
However, instead of using this hard-coded value, we should extract this value from the `dHigh` configured in `WakuRelay`.

Then, we should have something like:
```suggestion
  let minRelayPeers = WakuRelay.getDHigh()
```

And, in `waku_relay/protocol.nim`, we should create the following simple proc that exposes the dHigh configured value:

```nim
proc getDHigh*(T: type WakuRelay): int =
  return GossipsubParameters.dHigh
```

waku/node/peer_manager/peer_manager.nim
```diff
@@ -975,42 +987,59 @@ proc new*(
...
+  if relayPeers < minRelayPeers:
+    let newMaxConnections = int(float(minRelayPeers) / relayRatio)
+    warn "fullfill the minimum criteria you need to increase the max number of connections",
+      newMaxConnections = newMaxConnections
```
Ivansete-status, line 1021: I think this validation should be made earlier and, in case not configured properly, immediately stop the execution.

darshankabariya: Thanks for all the comments! I've tried to address all of them all but instead of replying to each one individually, I'm providing a summary in this message. I agree with most of your comments and have updated the PR accordingly. 

However, regarding this suggestion:

I couldn't find a way to validate it earlier. While it's possible to add the validation in the parse function, doing so would require the maxConnection value within that function, which could overcomplicate the logic and violate the single responsibility principle.

Thank you very much !

waku/common/utils/parse_size_units.nim
```diff
@@ -49,3 +49,22 @@ proc parseCorrectMsgSize*(input: string): uint64 =
   let ret = parseMsgSize(input).valueOr:
     return 0
   return ret
+
+## Parses a relay/service ratio string to [ float, float ]
+proc parseRelayServiceRatio*(ratio: string): Result[(float, float), string] =
```
Ivansete-status, line 54: Sorry, super nitpick comment: the proc comments should go below the proc's signature.
Let me also extend the comment a little bit :)
```suggestion
proc parseRelayServiceRatio*(ratio: string): Result[(float, float), string] =
  ## Parses a relay/service ratio string to [ float, float ]. The total should sum 100%
  ## e.g., (0.4, 0.6) == parseRelayServiceRatio("40:60")
```

waku/factory/builder.nim
```diff
@@ -108,10 +110,21 @@ proc withPeerStorage*(
...
+  try:
+    let (relayRatio, serviceRatio) = parseRelayServiceRatio(relayServiceRatio).get()
+    var relayPeers = int(ceil(float(maxConnections) * relayRatio))
+    var servicePeers = int(floor(float(maxConnections) * serviceRatio))
+
+    builder.maxServicePeers = servicePeers
+    builder.maxRelayPeers = relayPeers
+    builder.shardAware = shardAware
+  except ValueError:
+    error "Invalid relay service ratio format", ratio = relayServiceRatio
```
Ivansete-status, line 127: Extend the error detail a little bit:
```suggestion
    error "Invalid relay service ratio format", ratio = relayServiceRatio, error = getCurrentExceptionMsg()
```

waku/factory/external_config.nim
```diff
@@ -206,10 +212,6 @@ type WakuNodeConf* = object
       name: "ip-colocation-limit"
     .}: int
 
-    maxRelayPeers* {.
-      desc: "Maximum allowed number of relay peers.", name: "max-relay-peers"
-    .}: Option[int]
-
```
Ivansete-status, line 212: We cannot directly remove this optional parameter. Instead, we should inform that we are deprecating it and it will be completely removed after two releases. Therefore, we now just need to "tag" it as deprecated:

    maxRelayPeers* {.
      desc: "Deprecated. Use relay-service-ratio instead. It represents the maximum allowed number of relay peers.", name: "max-relay-peers"
    .}: Option[int]

waku/factory/external_config.nim
```diff
@@ -197,7 +197,13 @@ type WakuNodeConf* = object
       desc: "Maximum allowed number of libp2p connections.",
       defaultValue: 50,
       name: "max-connections"
-    .}: uint16
+    .}: int
+
+    relayServiceRatio* {.
+      desc: "This percentage ratio represents the relay peers to service peers",
+      name: "relay-service-ratio",
+      defaultValue: "60:40" # 60:40 ratio of relay to service peers
```
Ivansete-status, line 205: I suggest to extend the comment
```suggestion
      desc: "This percentage ratio represents the relay peers to service peers. For example, 60:40, tells that 60% of the max-connections will be used for relay protocol and the other 40% of max-connections will be reserved for other service protocols (e.g., filter, lightpush, store, metadata, etc.)",
      name: "relay-service-ratio",
      defaultValue: "60:40" # 60:40 ratio of relay to service peers
```

waku/node/peer_manager/peer_manager.nim
```diff
@@ -975,42 +987,43 @@ proc new*(
...
-    # Leave by default 20% of connections for service peers
-    maxRelayPeersValue = maxConnections - (maxConnections div 5)
+  var relayPeers = int(ceil(float(maxConnections) * relayRatio))
+  var servicePeers = int(floor(float(maxConnections) * serviceRatio))
+
+  let minRelayPeers = WakuRelay.getDHigh()
+
+  if relayPeers < minRelayPeers:
+    warn "Doesn't fulfill minimum criteria for relay (which increases the chance of the node becoming isolated)",
+      relayPeers = relayPeers, minRelayPeers = minRelayPeers
```
Ivansete-status, line 1005: I suggest to elaborate a bit more the error and also make the `new` proc to return a `Result` (maybe we need a separate PR to prepare that.)
```suggestion
    let errorMsg = "Doesn't fulfill minimum criteria for relay (which increases the chance of the node becoming isolated.) relayPeers, $1, should be greater or equal than minRelayPeers, $2" % [relayPeers, minRelayPeers]
    error "Wrong relay peers config", error = errorMsg

    return err(errorMsg)
```

Ivansete-status (commented):
It is getting great! Thanks so much for it and for the patience!
We need to iterate over it a little more :)
I've added some recommendations that I hope you find useful.

waku/factory/external_config.nim
darshankabariya: Good approch ! will revert this delete.

waku/node/peer_manager/peer_manager.nim
darshankabariya: Update the PR !

waku/node/peer_manager/peer_manager.nim
```diff
@@ -986,42 +998,48 @@ proc new*(
...
-      error "Max number of relay peers can't be greater than the max amount of connections",
-        maxConnections = maxConnections, maxRelayPeers = maxRelayPeers.get()
-      raise newException(
-        Defect,
-        "Max number of relay peers can't be greater than the max amount of connections",
-      )
+  var relayRatio: float64
+  var serviceRatio: float64
+  try:
+    (relayRatio, serviceRatio) = parseRelayServiceRatio(relayServiceRatio).get()
```
Ivansete-status, line 1004: If possible, I'd handle the possible exception in `parseRelayServiceRatio`

Ivansete-status (commented):
I just added a latest nitpick comment ;P

gabrielmer (approved):
Looks amazing! Thanks so much! 🔥

waku/node/peer_manager/peer_manager.nim
darshankabariya: Got your point ! just make it more centralized exception handling for parse function

waku/factory/external_config.nim
```diff
@@ -197,7 +197,20 @@ type WakuNodeConf* = object
       desc: "Maximum allowed number of libp2p connections.",
       defaultValue: 50,
       name: "max-connections"
-    .}: uint16
+    .}: int
+
+    maxRelayPeers* {.
+      desc:
+        "Deprecated. Use relay-service-ratio instead. It represents the maximum allowed number of relay peers.",
```
NagyZoltanPeter, line 204: While checking changes it seems to me the deprecated means here we will not use it at all.
In other words if one not moves right away to the new `relayServiceRatio` setting, regardless of `maxRelayPeers` default (60:40) will be applied without notice.
Maybe I'm missing something here, but if not, then it sounds like a breaking change anyway.
Cannot we balance somehow in between until we lead out this option completely?
Also maybe it is ok like this, I cannot decide.
But I can imagine a solution that `relayServiceRatio` has no default so we can decide the configurator intent.
If none of the two settings are set we apply 60:40 but when `maxRelayPeers` is set we derive from it. WDYT?

NagyZoltanPeter (commented):
All in all it is  good approach. Left some questions as some part was not very clear.
I like the complex string definition of cli arg, but here maybe a bit overenginered. Maybe it would be enough to define the percentage of either relay or service peers - as they must complement each other. But it is just my feeling, I do not want you to change your current way of doing.

waku/common/utils/parse_size_units.nim
```diff
@@ -49,3 +49,28 @@ proc parseCorrectMsgSize*(input: string): uint64 =
   let ret = parseMsgSize(input).valueOr:
     return 0
   return ret
+
+proc parseRelayServiceRatio*(ratio: string): Result[(float, float), string] =
```
NagyZoltanPeter, line 53:
```suggestion
proc parseRelayServiceRatio*(ratio: string): Result[(float, float), string] {.raises: [ValueError].}=
```
Let's not hide the exception, so compiler can help using the function properly.

waku/factory/external_config.nim
NagyZoltanPeter: I buy in to this idea. All I wanted to emphasize that do nothide the unintended/old usage of `maxRelayPeers` behind a log warning and a `--help` or docs.
Maybe it is better to stop node early and notice user about bad config, IDK.

NagyZoltanPeter (approved):
Great effort! I just added some tiny last notice on it if you would like to consider.

waku/factory/external_config.nim
darshankabariya: That’s a great idea! Users typically don’t check logs unless there’s an error. will stop after detect wrong config.

Ivansete-status (changes requested):

waku/factory/node_factory.nim
```diff
@@ -101,8 +101,17 @@ proc initNode(
     agentString = some(conf.agentString),
   )
   builder.withColocationLimit(conf.colocationLimit)
+
+  # Backword compatibility for maxRelayPeers
+  if conf.maxRelayPeers.isSome():
+    return err(
+      "maxRelayPeers is deprecated, using relayServiceRatio instead ( recommendation ), if relayServiceRatio is not set, by default it will be 60:40, maxRelayPeers and maxServicePeer is calculated accordingly"
```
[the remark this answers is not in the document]
darshankabariya: Previously I only notify with a debug message. However, I implemented Zoltan's suggestion: if we are not using the user configuration, it's better to stop execution early. This is because we completely bypass maxRelayPeers. If we don't stop during execution, the user might remain unaware of this behavior. thats why i use early stopping approch.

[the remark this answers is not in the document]
Ivansete-status: @darshankabariya - ah okay, I see your point. Nevertheless, every time we deprecate a parameter from `external_config.nim` we should:
- Maintain old and new added parameters within two consecutive versions
- Update both status and waku fleets accordingly. In this particular case, `max-relay-peers` is currently being used by them

This is the safest approach :)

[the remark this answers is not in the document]
darshankabariya: okay !

waku/factory/node_factory.nim
```diff
@@ -101,8 +101,15 @@ proc initNode(
     agentString = some(conf.agentString),
   )
   builder.withColocationLimit(conf.colocationLimit)
+
+  # Backword compatibility for maxRelayPeers
+  if conf.maxRelayPeers.isSome():
```
Ivansete-status, line 106: If this condition is `true`, then we need to infer the `relayServiceRation` based on the given `conf.maxConnections` and `conf.maxRelayPeers`. And we need to keep this temporarily within two consecutive versions

Ivansete-status (commented):
Thanks for it and the patience! :raised_hands: 
I think we still need to keep backward compatibility with the current `max-relay-peers` param.
On the other hand, my dreamed nwaku only deals with `Result` types, doesn't use `discard` and doesn't raise Exception either ;P

waku/common/utils/parse_size_units.nim
```diff
@@ -49,3 +49,28 @@ proc parseCorrectMsgSize*(input: string): uint64 =
...
+  ## Parses a relay/service ratio string to [ float, float ]. The total should sum 100%
+  ## e.g., (0.4, 0.6) == parseRelayServiceRatio("40:60")
+  try:
+    let elements = ratio.split(":")
+    if elements.len != 2:
+      raise newException(ValueError, "expected format 'X:Y', ratio = " & ratio)
```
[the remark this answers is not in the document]
darshankabariya: got it.

waku/factory/external_config.nim
darshankabariya: > While checking changes it seems to me the deprecated means here we will not use it at all. In other words if one not moves right away to the new `relayServiceRatio` setting, regardless of `maxRelayPeers` default (60:40) will be applied without notice. Maybe I'm missing something here, but if not, then it sounds like a breaking change anyway. Cannot we balance somehow in between until we lead out this option completely? Also maybe it is ok like this, I cannot decide. But I can imagine a solution that `relayServiceRatio` has no default so we can decide the configurator intent. If none of the two settings are set we apply 60:40 but when `maxRelayPeers` is set we derive from it. WDYT?

Now I understand your concern. I’ve made it backward compatible with maxRelayPeers, ensuring that users who aren’t aware of the deprecation are covered as well. At that time, I didn’t fully grasp your point.

waku/factory/node_factory.nim
darshankabariya: Thanks for your review and patience. It’s now backward compatible and parse function without raising exceptions.

waku/factory/builder.nim
```diff
@@ -108,10 +110,21 @@ proc withPeerStorage*(
...
 proc withPeerManagerConfig*(
-    builder: var WakuNodeBuilder, maxRelayPeers = none(int), shardAware = false
+    builder: var WakuNodeBuilder,
+    maxConnections: int,
+    relayServiceRatio: string,
+    shardAware = false,
 ) =
-  builder.maxRelayPeers = maxRelayPeers
-  builder.shardAware = shardAware
+  try:
```
Ivansete-status, line 118: Maybe not needed to handle exception now?

waku/node/peer_manager/peer_manager.nim
Ivansete-status: Awesome thanks! I think now we don't need to handle the exception

Ivansete-status (approved):
LGTM! Thanks for it! :100: :partying_face:

waku/factory/builder.nim
darshankabariya: Yes ! we already handle in function.

PR #3187, opened 2024-11-29
fix: lite-protocol-tester receiver exit check

NagyZoltanPeter:
# Description
After deployment liteprotocoltester into infra and see running for some days it turned out that in some cases receiver side cannot detect end of test and keeps waiting for more messages.
This PR fixes it with introducing a max time after last message arrived to consider exiting from test loop.

Also valuable information is added to logging and metrics, tested peers and service node failures now distinguished by agent string also. This way dashboard and provide information of node types and versions of erroneous peers.


apps/liteprotocoltester/statistics.nim
```diff
@@ -274,16 +279,43 @@ proc jsonStats*(self: PerPeerStatistics): string =
...
+
+  if not messageCheck:
+    let lastMessageAt = self.lastMessageArrivedAt().valueOr:
       return false
 
+    if Moment.now() - lastMessageAt < maxWaitForLastMessage:
+      return false
+    else:
+      info "No message since max wait time", maxWait = $maxWaitForLastMessage
+
```
gabrielmer, line 318: I'm struggling to understand this part of the code, could you please explain it in a few words? Apologies, maybe I lack context or today I'm particularly dumb lol

NagyZoltanPeter: No way, it is my fault if it is not straightforward enough. Apologise.

receiver side can collect messages from many publisher at once. It creates statistics separately of them.
This part tries to get the last message arrived among all possible sender/publisher peer.

Than here even if we think not all of the messages arrived we check a time limit, a max wait before we can consider the test ended on the other parties.
So we check this threshold against last message arrival.

The issue I faced that if there were some message lost (maybe because of filter service failure) the receiver though it can expect more messages to come yet did not exit... than it screw up the dashboard statistics.

I will add comment to the code!

gabrielmer: Ooh now I understood it! Thanks so much! 😍

gabrielmer (approved):
LGTM, thanks so much! 🤩

apps/liteprotocoltester/statistics.nim
```diff
@@ -274,16 +279,48 @@ proc jsonStats*(self: PerPeerStatistics): string =
...
+  var lastArrivedAt = Moment.init(0, Millisecond)
+  for stat in self.values:
+    let lastMsgFromPeerAt = stat.lastMessageArrivedAt().valueOr:
+      continue
+
+    if lastMsgFromPeerAt > lastArrivedAt:
+      lastArrivedAt = lastMsgFromPeerAt
+
+  if lastArrivedAt == Moment.init(0, Millisecond):
+    return err()
```
Ivansete-status, line 292: If possible I'd always suggest returning a hint about the error condition

apps/liteprotocoltester/statistics.nim
```diff
@@ -274,16 +279,48 @@ proc jsonStats*(self: PerPeerStatistics): string =
...
+
+  if not isAlllMessageReceived:
+    # if not all message received we still need to check if last message arrived within a time frame
+    # to avoid endless waiting while publishers are already quit.
+    let lastMessageAt = self.lastMessageArrivedAt().valueOr:
       return false
 
+    # last message shall arrived within time limit
+    if Moment.now() - lastMessageAt < maxWaitForLastMessage:
+      return false
```
Ivansete-status, line 320: Instead of returning a bool I think in general is better to return a `Result[T, string]`

Ivansete-status (approved):
LGTM! Thanks for it! :100: 
I just added a couple of nitpick comments:)

gabrielmer (approved):
Thank you!

apps/liteprotocoltester/statistics.nim
NagyZoltanPeter: Well, yes. Although it is not an error, it just shows that there is no message arrived yet. I changed return value to Option[Moment] which describes better the functionality.

apps/liteprotocoltester/statistics.nim
NagyZoltanPeter: Not in this case, the function really decides upon if last message arrived - or in other words can we expect any more messages. Hence it is true or false.

PR #3197, opened 2024-12-04
chore: improve epoch monitoring

darshankabariya:
I discovered a bug while monitoring Grafana metrics in PR #3181 Currently, the **waku_rln_proof_remining** counter is reset inside the **generateProof** function. This approach fails in the following scenario:

If a new epoch starts but no proofs are generated (i.e., no messages are sent), the counter doesn't reset. Ideally, the counter should reset immediately when a new epoch begins, regardless of whether any messages are sent.

This PR addresses the issue by implementing an independent epoch monitoring task that ensures the counter is reset as soon as a new epoch starts.

Note: The spike should occur around 23:50, but it happens around 23:53 because the user does not generate any RLN proof ( msg ) within the 3-minute timeframe.

image

### DIFF
--- a/waku/waku_rln_relay/protocol_metrics.nim
+++ b/waku/waku_rln_relay/protocol_metrics.nim
@@ -63,12 +63,12 @@ declarePublicGauge(
 )
 
 declarePublicGauge(
-  waku_rln_proof_remining,
+  waku_rln_remaining_proofs_per_epoch,
   "number of proofs remaining to be generated for the current epoch",
 )
 
 declarePublicGauge(
-  waku_rln_proofs_generated_total,
+  waku_rln_total_generated_proofs,
   "total number of proofs generated since the node started",
 )
 
@@ -84,6 +84,7 @@ proc getRlnMetricsLogger*(): RLNMetricsLogger =
   var cumulativeValidMessages = 0.float64
   var cumulativeProofsVerified = 0.float64
   var cumulativeProofsGenerated = 0.float64
+  var cumulativeProofsRemaining = 100.float64
 
   when defined(metrics):
     logMetrics = proc() =
@@ -102,7 +103,10 @@ proc getRlnMetricsLogger*(): RLNMetricsLogger =
           waku_rln_proof_verification_total, cumulativeProofsVerified
         )
         let freshProofsGeneratedCount =
-          parseAndAccumulate(waku_rln_proofs_generated_total, cumulativeProofsGenerated)
+          parseAndAccumulate(waku_rln_total_generated_proofs, cumulativeProofsGenerated)
+        let freshProofsRemainingCount = parseAndAccumulate(
+          waku_rln_remaining_proofs_per_epoch, cumulativeProofsRemaining
+        )
 
         info "Total messages", count = freshMsgCount
         info "Total spam messages", count = freshSpamCount
@@ -111,4 +115,6 @@ proc getRlnMetricsLogger*(): RLNMetricsLogger =
         info "Total errors", count = freshErrorCount
         info "Total proofs verified", count = freshProofsVerifiedCount
         info "Total proofs generated", count = freshProofsGeneratedCount
+        info "Total proofs remaining", count = freshProofsRemainingCount
+
   return logMetrics
--- a/waku/waku_rln_relay/rln_relay.nim
+++ b/waku/waku_rln_relay/rln_relay.nim
@@ -96,6 +96,18 @@ proc calcEpoch*(rlnPeer: WakuRLNRelay, t: float64): Epoch =
   let e = uint64(t / rlnPeer.rlnEpochSizeSec.float64)
   return toEpoch(e)
 
+proc nextEpoch*(rlnPeer: WakuRLNRelay, time: float64): float64 =
+  let
+    currentEpoch = uint64(time / rlnPeer.rlnEpochSizeSec.float64)
+    nextEpochTime = float64(currentEpoch + 1) * rlnPeer.rlnEpochSizeSec.float64
+    currentTime = epochTime()
+
+  # Ensure we always return a future time
+  if nextEpochTime > currentTime:
+    return nextEpochTime
+  else:
+    return epochTime()
+
 proc stop*(rlnPeer: WakuRLNRelay) {.async: (raises: [Exception]).} =
   ## stops the rln-relay protocol
   ## Throws an error if it cannot stop the rln-relay protocol
@@ -392,6 +404,19 @@ proc generateRlnValidator*(
 
   return validator
 
+proc monitorEpochs(wakuRlnRelay: WakuRLNRelay) {.async.} =
+  while true:
+    try:
+      waku_rln_remaining_proofs_per_epoch.set(
+        wakuRlnRelay.groupManager.userMessageLimit.get().flaot64
+      )
+    except CatchableError:
+      error "Error in epoch monitoring", error = getCurrentExceptionMsg()
+
+    let nextEpochTime = wakuRlnRelay.nextEpoch(epochTime())
+    let sleepDuration = int((nextEpochTime - epochTime()) * 1000)
+    await sleepAsync(sleepDuration)
+
 proc mount(
     conf: WakuRlnConfig, registrationHandler = none(RegistrationHandler)
 ): Future[RlnRelayResult[WakuRlnRelay]] {.async.} =
@@ -445,16 +470,17 @@ proc mount(
   (await groupManager.startGroupSync()).isOkOr:
     return err("could not start the group sync: " & $error)
 
-  return ok(
-    WakuRLNRelay(
-      groupManager: groupManager,
-      nonceManager:
-        NonceManager.init(conf.rlnRelayUserMessageLimit, conf.rlnEpochSizeSec.float),
-      rlnEpochSizeSec: conf.rlnEpochSizeSec,
-      rlnMaxEpochGap: max(uint64(MaxClockGapSeconds / float64(conf.rlnEpochSizeSec)), 1),
-      onFatalErrorAction: conf.onFatalErrorAction,
-    )
+  wakuRlnRelay = WakuRLNRelay(
+    groupManager: groupManager,
+    nonceManager:
+      NonceManager.init(conf.rlnRelayUserMessageLimit, conf.rlnEpochSizeSec.float),
+    rlnEpochSizeSec: conf.rlnEpochSizeSec,
+    rlnMaxEpochGap: max(uint64(MaxClockGapSeconds / float64(conf.rlnEpochSizeSec)), 1),
+    onFatalErrorAction: conf.onFatalErrorAction,
   )
+
+  wakuRlnRelay.epochMonitorFuture = monitorEpochs(wakuRlnRelay)
+  return ok(wakuRlnRelay)
 
 proc isReady*(rlnPeer: WakuRLNRelay): Future[bool] {.async: (raises: [Exception]).} =
   ## returns true if the rln-relay protocol is ready to relay messages
--- a/waku/waku_rln_relay/rln_relay.nim
+++ b/waku/waku_rln_relay/rln_relay.nim
@@ -408,7 +408,7 @@ proc monitorEpochs(wakuRlnRelay: WakuRLNRelay) {.async.} =
   while true:
     try:
       waku_rln_remaining_proofs_per_epoch.set(
-        wakuRlnRelay.groupManager.userMessageLimit.get().flaot64
+        wakuRlnRelay.groupManager.userMessageLimit.get().float64
       )
     except CatchableError:
       error "Error in epoch monitoring", error = getCurrentExceptionMsg()
--- a/waku/waku_rln_relay/rln_relay.nim
+++ b/waku/waku_rln_relay/rln_relay.nim
@@ -89,6 +89,7 @@ type WakuRLNRelay* = ref object of RootObj
   groupManager*: GroupManager
   onFatalErrorAction*: OnFatalErrorHandler
   nonceManager*: NonceManager
+  epochMonitorFuture*: Future[void]
 
 proc calcEpoch*(rlnPeer: WakuRLNRelay, t: float64): Epoch =
   ## gets time `t` as `flaot64` with subseconds resolution in the fractional part
@@ -478,7 +479,8 @@ proc mount(
     rlnMaxEpochGap: max(uint64(MaxClockGapSeconds / float64(conf.rlnEpochSizeSec)), 1),
     onFatalErrorAction: conf.onFatalErrorAction,
   )
-
+   
+  # Start epoch monitoring in the background
   wakuRlnRelay.epochMonitorFuture = monitorEpochs(wakuRlnRelay)
   return ok(wakuRlnRelay)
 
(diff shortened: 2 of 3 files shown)


waku/waku_rln_relay/rln_relay.nim
```diff
@@ -392,6 +392,23 @@ proc generateRlnValidator*(
...
+
+      if currentEpoch != lastEpoch:
+        waku_rln_proof_remining.set(
+          wakuRlnRelay.groupManager.userMessageLimit.get().float64
+        )
+        lastEpoch = currentEpoch
+
+      await sleepAsync(500)
+    except CatchableError as e:
+      error "Error in epoch monitoring", err = e.msg
```
Ivansete-status, line 410: In this cases we tend to use this approach:
```suggestion
    except CatchableError:
      error "Error in epoch monitoring", error = getCurrentExceptionMsg()
```

waku/waku_rln_relay/rln_relay.nim
```diff
@@ -445,17 +462,20 @@ proc mount(
...
+    groupManager: groupManager,
+    nonceManager:
+      NonceManager.init(conf.rlnRelayUserMessageLimit, conf.rlnEpochSizeSec.float),
+    rlnEpochSizeSec: conf.rlnEpochSizeSec,
+    rlnMaxEpochGap: max(uint64(MaxClockGapSeconds / float64(conf.rlnEpochSizeSec)), 1),
+    onFatalErrorAction: conf.onFatalErrorAction,
   )
 
+  # Start epoch monitoring in the background
+  asyncSpawn monitorEpochs(wakuRlnRelay)
```
Ivansete-status, line 475: Instead of performing a _launch-and-forget_ approach, it is better to get a reference of the async-task future handle so that it can be gracefully cancelled afterwards.
See for example:
https://github.com/waku-org/nwaku/blob/1b532e8ab9efec70b0d8b79ba65e145e2b3cb76b/waku/waku_archive/archive.nim#L256
https://github.com/waku-org/nwaku/blob/1b532e8ab9efec70b0d8b79ba65e145e2b3cb76b/waku/waku_archive/archive.nim#L264

Ivansete-status (commented):
Thanks for it! I added a couple of minor enhancements:)

gabrielmer (approved):
On one hand, I'm not a super big fan of having a procedure running in a loop every 500 ms just to reset one counter. On the other hand, it is indeed a bug and have nothing better to propose right now lol.

So I think it's ok, but it will be cool to eventually design a solution that implies a bit less of brute force. Don't want to overcomplicate things either though.

Thanks so much for catching and fixing this issue!

darshankabariya:
> Thanks for it! I added a couple of minor enhancements:)

Hi @Ivansete-status  thanks for your review! I've updated the PR based on your suggestions. Could you please check it again?

darshankabariya:
> On one hand, I'm not a super big fan of having a procedure running in a loop every 500 ms just to reset one counter. On the other hand, it is indeed a bug and have nothing better to propose right now lol.
> 
> So I think it's ok, but it will be cool to eventually design a solution that implies a bit less of brute force. Don't want to overcomplicate things either though.
> 
> Thanks so much for catching and fixing this issue!

Hi @gabrielmer , thank you so much for your comment! I found a better approach and have updated the PR. Could you please take a look?

waku/waku_rln_relay/rln_relay.nim
```diff
@@ -392,6 +393,33 @@ proc generateRlnValidator*(
...
+  var nextEpochTime = epochTime()
+  var lastEpoch = wakuRlnRelay.calcEpoch(epochTime())
+  var firstChangeDetected = false
+
+  while true:
+    try:
+      let currentEpoch = wakuRlnRelay.calcEpoch(epochTime())
+
+      if currentEpoch != lastEpoch:
+        nextEpochTime = nextEpochTime + float64(wakuRlnRelay.rlnEpochSizeSec)
```
gabrielmer, line 406: Isn't there a way to calculate the epoch time in which the RLN epoch started? Now the time won't drift but we can be off up to 500 ms forever.

If we use as reference the time in which the RLN epoch started instead of the time we are checking, it will make it much more accurate :)

gabrielmer: Ok there isn't a simple function to get the next epoch, but it is also not too hard to do

https://github.com/waku-org/nwaku/blob/8f2bd39f2304628ad27b42c3964b4a614292b56a/waku/waku_rln_relay/rln_relay.nim#L93-L97

Here's how RLN epochs are calculated as a function of time.

To find the time where the next epoch starts, we need the minimum `t` greater than the current time such that

```
uint64(t / rlnPeer.rlnEpochSizeSec.float64)
```

results in a bigger epoch than the current one.

So basically, we need the smallest multiple of `rlnPeer.rlnEpochSizeSec` that is bigger than the current time.

Even the brute force way, which is adding 1 in a loop to `t` until `t % rlnPeer.rlnEpochSizeSec.float64 == 0` does the job hahaha

Anyway, don't want to overcomplicate things, so I leave it to your criteria :)

gabrielmer (approved):

waku/waku_rln_relay/rln_relay.nim
```diff
@@ -89,13 +89,29 @@ type WakuRLNRelay* = ref object of RootObj
...
 
 proc calcEpoch*(rlnPeer: WakuRLNRelay, t: float64): Epoch =
   ## gets time `t` as `flaot64` with subseconds resolution in the fractional part
   ## and returns its corresponding rln `Epoch` value
   let e = uint64(t / rlnPeer.rlnEpochSizeSec.float64)
   return toEpoch(e)
 
+proc nextEpochTime*(rlnPeer: WakuRLNRelay, t: float64): float64 =
+  var leftPtr = t
+  var rightPtr = t + rlnPeer.rlnEpochSizeSec.float64
```
gabrielmer, line 102: I think we should convert both pointers to uint64 to be more precise (as we calculate epochs based on uints) and to avoid bugs related to both pointers not converging properly because of floating point differences.

Apart from that, logic looks good but I would make sure sure to test it well as I'm not smart enough to be 100% sure about it at first sight lol.

Otherwise, we can just brute force it instead of doing this kind of binary search, as it might be easier to follow, less bug prone, and at the end of the day it's just a loop with max 10 iterations which is nothing :)

Ivansete-status (commented):
Thanks for the enhancements!
Nevertheless, I have the impression that we are over complicating it :)
Let me elaborate a bit more.
We are looking for showing the generated proofs , and the remaining-available proofs within an RLN epoch when the node acts as a RLNaaS.

In this case, I think we have the epoch time window very well defined by the `rlnEpochSizeSec` config param (10min in the case of _The Waku Network_.) Then, I believe we only need a while loop that resets the `waku_rln_remaining_proofs_per_epoch` once every ``rlnEpochSizeSec` seconds.

Then, I think we only need:
```nim
while true:
  ...
  await sleepAsync(rlnEpochSizeSec * 1000)
```

On the other hand, it is pending to `cancelAndWait` the `epochMonitorFuture` from within the `stop` proc.

darshankabariya:
> The issue is that a sleep of 10 minutes does not take exactly 10 minutes, there will be a small error and in every iteration of the loop the error will increase and increase, until at some point it will be completely out of shift.
> 
> Maybe it won't be immediately noticeable, but if you have a long-running node it does become an issue I believe

excatly !

darshankabariya:
Hi @gabrielmer @Ivansete-status @NagyZoltanPeter @SionoiS 

I recently noticed something during testing. As we all know, with RLN, we can send up to 100 messages per epoch (10 minutes). Here’s what I observed, and I’m unsure if it’s expected behavior or a bug:

For example, if an epoch starts at 4:00, we can send 100 messages until 4:10. If I try to send more than 100 messages, they’re blocked, which makes sense. But in another scenario, if the epoch starts at 4:00 and I don’t send any messages until 4:06, then send all 100 messages between 4:06 and 4:09, when the new epoch starts at 4:10, I still can’t send any messages and get a ‘message limit exceeded’ error. Interestingly, I’m only able to send messages again at 4:16. 

Is this behavior expected, or could it be a bug? If it’s intentional, I’ll need to adjust the logic to handle this scenario.

gabrielmer:
> but recently i noticed that epoch time always sync across the waku network and triggers on the 10th minute `(e.g., XX:X0)`. It never starts at random times like 13 or 17 (It means it's not just tied to the node's start time). I think we can leverage this to simplify the logic, but just wanted to confirm if I’m missing something. I observed this after several tests.

Yes! My understanding is that the epoch is something global for the whole network and does not depend on one node's start time.

In this comment I explained how I found it is calculated https://github.com/waku-org/nwaku/pull/3197#discussion_r1876440964

We can't assume that the epoch is 10 minutes though, I think we have to use the general `rlnEpochSizeSec` parameter to compute when the next epoch is supposed to start

gabrielmer:
> Hi @gabrielmer @Ivansete-status @NagyZoltanPeter @SionoiS
> 
> I recently noticed something interesting during test this PR. we all know, with RLN, we can send up to 100 messages per epoch (10 minutes). Here’s what I observed, and I’m unsure if it’s expected behavior or a bug:
> 
> For example, if an epoch starts at 4:00, we can send 100 messages until 4:10. If I try to send more than 100 messages, they’re blocked, which makes sense. But in another scenario, if the epoch starts at 4:00 and I don’t send any messages until 4:06, then send all 100 messages between 4:06 and 4:09, when the new epoch starts at 4:10, I still can’t send any messages and get a ‘message limit exceeded’ error. Interestingly, I’m only able to send messages again at 4:16.
> 
> Is this behavior expected, or could it be a bug? If it’s intentional, I’ll need to adjust the logic to handle this scenario.

Ah wow that's super interesting. I would think that it is a bug, my understanding is that the limit should be according to the global epoch and not depending on when our node starts sending messages. But I'm not sure though, @alrevuelta should know better

How are you sending the messages? via REST API? If so, I suspect it might be a bug in our REST layer. I might be wrong, but that's the first thing I would check

darshankabariya:
> We can't assume that the epoch is 10 minutes though, I think we have to use the general `rlnEpochSizeSec` parameter to compute when the next epoch is supposed to start

Thanks! I agree, there's a possibility that sync times could change (currently it starts at the 10th, but this could change) or even the rlnEpochSizeSec could change. In this scenario, the idea would fail. On further thought, I believe the current implementation is much more reliable.

darshankabariya:
> I’m using the chat interface to send messages. There is a possibility of an issue with the chat interface, but let’s confirm with @alrevuelta.


Just adding a comment to finalize the communication related to this PR. I believe it is ready to be merged and I already report an issue based on my founding, which I address eventually.

waku/waku_rln_relay/protocol_metrics.nim
```diff
@@ -84,6 +84,7 @@ proc getRlnMetricsLogger*(): RLNMetricsLogger =
   var cumulativeValidMessages = 0.float64
   var cumulativeProofsVerified = 0.float64
   var cumulativeProofsGenerated = 0.float64
+  var cumulativeProofsRemaining = 100.float64
```
Ivansete-status, line 87: Instead of initializing by a hard-coded value, we need to use the value that comes from [rlnRelayUserMessageLimit](https://github.com/waku-org/nwaku/blob/09f05fefe2614d640277f96ae96ab4f2f12c8031/waku/factory/external_config.nim#L103)

We only have hard-coded value in `ClusterConf`. Then, the config value my get overriden. For example, https://github.com/waku-org/nwaku/blob/09f05fefe2614d640277f96ae96ab4f2f12c8031/waku/factory/waku.nim#L194

Summarizing, we need to avoid this hard-coded value ;P

waku/waku_rln_relay/rln_relay.nim
```diff
@@ -89,19 +89,43 @@ type WakuRLNRelay* = ref object of RootObj
...
   nonceManager*: NonceManager
+  epochMonitorFuture*: Future[void]
 
 proc calcEpoch*(rlnPeer: WakuRLNRelay, t: float64): Epoch =
   ## gets time `t` as `flaot64` with subseconds resolution in the fractional part
   ## and returns its corresponding rln `Epoch` value
   let e = uint64(t / rlnPeer.rlnEpochSizeSec.float64)
   return toEpoch(e)
 
+proc nextEpoch*(rlnPeer: WakuRLNRelay, t: float64): float64 =
```
Ivansete-status, line 100: Excuse me, but I think this seems too complicated. May kindly elaborate why is needed? I was wondering if it may suffice with just a fixed `await sleepAsync(FixedEpochCheckDuration)`?

Ivansete-status (commented):
Thanks for this :raised_hands: !
I just added a comment to avoid hard-coded value and other comment seeking for simplification.
Ping me if there is anything unclear. It is likely that I'm missing something.

waku/waku_rln_relay/rln_relay.nim
gabrielmer: I think this is the same thing we discussed here, right? https://github.com/waku-org/nwaku/pull/3197#issuecomment-2541615283

Ivansete-status: Sorry @gabrielmer , I didn't see your comment. I believe it doesn't need to be super precise and I'd rather simplicity than super precise.
Then I suggest following a simpler path for now and we can add complexity in the future if we need it. For now I think it's fine being simple, unless I'm missing something of course :)

gabrielmer: The important thing is not being super precise, it is not going completely out of shift. If we leave the node running for enough time, that counter will start mid-epoch and will give completely wrong information.

I think that having a counter that will eventually give completely wrong data is way worse than not having a counter at all. It doesn't have to be super-precise but there has to be a cap on how imprecise it can be, and here not only there isn't such a cap - it will always get to such a point

waku/waku_rln_relay/rln_relay.nim
```diff
@@ -89,19 +89,43 @@ type WakuRLNRelay* = ref object of RootObj
...
+  # Increment by seconds to find the exact transition
+  while rlnPeer.calcEpoch(timePtr) == currentEpoch:
+    timePtr += 1 # 1 second
+
+  # Ensure the returned time is in the future
+  if timePtr > epochTime():
+    return timePtr
+  else:
+    return epochTime()
+
```
gabrielmer, line 121: We can simplify this a lot by doing this

```suggestion
proc nextEpoch*(rlnPeer: WakuRLNRelay, t: float64): float64 =
  # Calculates the next epoch time from the given time `t`.  
  (uint64(t / rlnPeer.rlnEpochSizeSec.float64) + 1) * rlnPeer.rlnEpochSizeSec
```

This is much simpler, less bug-prone and even more efficient

There may still be concerns of adding extra complexity. At least this way there's way less extra complexity than the previous approach

waku/waku_rln_relay/rln_relay.nim
```diff
@@ -392,6 +406,18 @@ proc generateRlnValidator*(
...
+proc monitorEpochs(wakuRlnRelay: WakuRLNRelay): Future[void] {.async.} =
+  while true:
+    try:
+      waku_rln_remaining_proofs_per_epoch.set(
+        wakuRlnRelay.groupManager.userMessageLimit.get().float64
+      )
+    except CatchableError:
+      error "Error in epoch monitoring", error = getCurrentExceptionMsg()
+
+    let nextEpochTime = wakuRlnRelay.nextEpoch(epochTime())
```
Ivansete-status, line 418: Maybe not needed?
```suggestion
```

Ivansete-status (approved):
LGTM! Thanks for it! :100: 
I just added some nitpick comments

PR #3198, opened 2024-12-05
fix: filter - enhancements in subscription management

Ivansete-status:
## Description

Railgun notified about filter issues. This PR tackled that.
Railgun informed that after 4 days working properly, the nodes stopped sending messages. There were also issues that caused a max of 100 messages being sent over one connection and the reason for that was that we were `dialPeer`ing on every single message the filter server sent to the filter subscribers.

## Changes

* waku_filter_v2: idiomatic way run periodic subscription manager
* filter subscriptions: add more debug logs
* filter: make sure the custom start and stop procs are called
* make sure filter protocol is started if it is mounted
* filter: dial push connection onsubscribe only
* waku/waku_filter_v2/client.nim: make the client to behave as a server for the FilterPush messages sent by filter server

## How to test

1. Start a nwaku with wss enabled
2. [lab.waku.org](https://github.com/waku-org/lab.waku.org) -> examples/experimental/railgun-repro -> npm run dev
3. Press F12 and you should see messages comming without interruption

## Issue

closes https://github.com/waku-org/nwaku/issues/3191
closes https://github.com/waku-org/nwaku/issues/3192


gabrielmer (approved):
Woooow amazing PR! 😍 

Thanks so much!

waku/waku_filter_v2/protocol.nim
```diff
@@ -236,7 +250,7 @@ proc handleMessage*(
     let subscribedPeers =
       wf.subscriptions.findSubscribedPeers(pubsubTopic, message.contentTopic)
     if subscribedPeers.len == 0:
-      trace "no subscribed peers found",
+      error "no subscribed peers found",
```
NagyZoltanPeter, line 253: I don't think it is an error if there is noone that subscribed to that message. Even further in general it is an annoying log as trace. Rather we log only positives in such cases.

waku/waku_filter_v2/subscriptions.nim
```diff
@@ -120,9 +166,19 @@ proc addSubscription*(
   do:
     ## not yet subscribed
     if cast[uint](s.peersSubscribed.len) >= s.maxPeers:
-      return err("node has reached maximum number of subscriptions")
+      return err("node has reached maximum number of subscriptions: " & $(s.maxPeers))
+
+    let connRes = await peerManager.dialPeer(peerId, WakuFilterPushCodec)
```
NagyZoltanPeter, line 171: I see the intent to avoid unnecessary dials over message push cycles, although opening a stream at subscribe seems to me a over optimization + an attack vector where a harmful actor may subscribe to filters will never match and eat up our connections pool.
Also it can be that a subscription may not result in push in a close future without any intent to harm the service.

waku/waku_filter_v2/client.nim
```diff
@@ -168,22 +175,34 @@ proc registerPushHandler*(wfc: WakuFilterClient, handler: FilterPushHandler) =
...
-    let messagePush = decodeRes.value #TODO: toAPI() split here
-    trace "Received message push", peerId = conn.peerId, messagePush
-
-    for handler in wfc.pushHandlers:
-      asyncSpawn handler(messagePush.pubsubTopic, messagePush.wakuMessage)
-
-    # Protocol specifies no response for now
-    return
+    ## Notice that the client component is acting as a server of WakuFilterPushCodec messages
+    while not conn.atEof():
```
NagyZoltanPeter, line 179: How exactly this helps on being a server like? I though it is libp2p duty to maintain it.

NagyZoltanPeter (commented):
I totally agree on re-thinking how we manage message-push and its requirements.
Can you add some explanation how this solution exactly solves the issue appeared for Railgun? I see some hole in the changed connection management as is now.
Can we brainstorm on this?

NagyZoltanPeter (approved):
With https://github.com/waku-org/nwaku/issues/3254 future tackling the peer management of dialing existing connection, see no problem with these enhancements.

PR #3205, opened 2024-12-09
fix(libwaku): support string and int64 for timestamps

richard-ramos:
The Protocol Buffers JSON Specification for 64 bit numbers, recommends serializing numbers as strings to avoid loss of precision when working with 64-bit integers in JavaScript environments. This causes issues as some json serializers like Go's respect the specification when serializing a protobuffer into Json.
To avoid this issue, in this PR will allow accepting both integers and numeric strings for those cases in which this spec is not followed

### DIFF
--- a/library/utils.nim
+++ b/library/utils.nim
@@ -1,0 +1,18 @@
+import std/[json, options, strutils]
+import results
+import ../waku/waku_core/time
+
+proc getProtoInt64*(node: JsonNode, key: string): Option[int64] =
+  let (value, ok) =
+    if node.hasKey(key):
+      if node[key].kind == JString:
+        (parseBiggestInt(node[key].getStr()), true)
+      else:
+        (node[key].getBiggestInt(), true)
+    else:
+      (0, false)
+
+  if ok:
+    return some(value)
+
+  return none(int64)
--- a/library/utils.nim
+++ b/library/utils.nim
@@ -1,18 +1,20 @@
 import std/[json, options, strutils]
 import results
-import ../waku/waku_core/time
 
-proc getProtoInt64*(node: JsonNode, key: string): Option[int64] =
-  let (value, ok) =
-    if node.hasKey(key):
-      if node[key].kind == JString:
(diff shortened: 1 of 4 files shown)


NagyZoltanPeter (approved):
LGTM!
Maybe we could generalize this approach for easy reuse.

richard-ramos:
> @NagyZoltanPeter: Maybe we could generalize this approach for easy reuse.

Would something like this work?
```nim
import std/[json, strutils]
import results
import ../../../waku/waku_core/time

proc getProtoTimestamp*[T](node: JsonNode, key: string): Opt[T] =
  let (value, ok) =
    if node.hasKey(key):
      if node[key].kind == JString:
        (parseInt(node[key].getStr()), true)
      else:
        (node[key].getBiggestInt(), true)

  if not (T is Timestamp or T is int or T is int64):
    raise newException(ValueError, "Unsupported type")

  return
    if not ok:
      return T.none()
    else:
      when T is Timestamp:
        some(Timestamp(value))
      else:
        some(value)
```

richard-ramos:
@NagyZoltanPeter check the latest commit!
I made it even more generic by returning an `int64` instead. I had not realized that `Timestamp` was an alias for `int64`

library/utils.nim
```diff
@@ -0,0 +1,18 @@
+import std/[json, options, strutils]
+import results
+import ../waku/waku_core/time
+
+proc getProtoInt64*(node: JsonNode, key: string): Option[int64] =
```
Ivansete-status, line 5: I wonder if we should return a `Result` instead so that we can better control any possible error

Ivansete-status (approved):
Awesome! Thanks so much for it! :raised_hands:

library/utils.nim
richard-ramos: I changed it to a result now. Please do check latest commit and let me know your thoughts!

NagyZoltanPeter (approved):
I like it! :-)

Issue #3076, opened 2024-09-17
[Epic: nwaku] add nwaku into status-go


Ivansete-status:
# Summary
https://github.com/status-im/status-go should operate with waku through nwaku

We need to add a new `nwaku` submodule within `status-go` and adapt the current `status-go` code base plus nwaku-libwaku so that we can have  `status-go` working with `nwaku`.

There will be changes in:
- status-go to integrate nwaku
- nwaku to extend libwaku with the features needed by status-go

# Tasks

- [ ] https://github.com/waku-org/nwaku/issues/3039
- [ ] https://github.com/waku-org/nwaku/issues/3052
- [ ] https://github.com/waku-org/nwaku/issues/3053

gabrielmer:
Discv5 works after being disconnected for a long time - the bootstrap nodes are never removed from the routing tables. We can see the code [here](https://github.com/status-im/nim-eth/blob/88e4be4dc40e044834dca68f8b69d144744bf145/eth/p2p/discoveryv5/protocol.nim#L488-L4950)

However, after being online again, we don't attempt to connect to the nodes we already found because we have them in backoff period. When we are disconnected, our node thinks the other nodes are faulty because it can't connect to them, but in fact the faulty node is ours. When online again, because of this issue, our node doesn't attempt to connect to them.

We should create a mechanism to recognize when our node is offline and avoid attempting to connect to other nodes whenever that's the case, or at least avoid mischaracterizing well-behaved peers

PR #3206, opened 2024-12-11
feat: allowing configuration of application level callbacks

gabrielmer:
# Description

We noticed that libwaku doesn't automatically configure a callback for receiving Relay messages when creating a node. Currently, the user has to manually subscribe to each topic providing their callback.

In order to make libwaku more user friendly, we need to be able to register all the event-emitting callbacks automatically when creating a node. Therefore, I added a `WakuCallbacks` type where a user can pass all the different callbacks they need when creating a node. Currently, only `relayHandler` is defined in the type, but in a following PRs new callbacks will be added (for example, a callback that runs every time a topic health changes)

# Changes

- [x] introduced `WakuCallbacks` type for application-level callbacks
- [x] added `setupCallbacks` procedure that adds the configured `WakuCallbacks`
- [x] modified `Waku.new()` to accept `WakuCallbacks` and properly set them up
- [x] introduced a `callbacks` parameter to libwaku's `createWaku`so we can configure in node creation all the event-emitting callbacks

## Issue
#3076

### DIFF
--- a/waku/factory/waku.nim
+++ b/waku/factory/waku.nim
@@ -254,7 +254,7 @@ proc new*(
 
   node.setupCallbacks(confCopy, callbacks).isOkOr:
     error "Failed setting callbacks", error = error
-    return err("Failed setting up node: " & $error)
+    return err("Failed setting up callbacks: " & $error)
 
   ## Delivery Monitor
   var deliveryMonitor: DeliveryMonitor
@@ -277,6 +277,7 @@ proc new*(
     key: confCopy.nodekey.get(),
     node: node,
     deliveryMonitor: deliveryMonitor,
+    callbacks: callbacks,
   )
 
   waku.setupSwitchServices(confCopy, relay, rng)
--- a/waku/factory/waku.nim
+++ b/waku/factory/waku.nim
@@ -148,12 +148,22 @@ proc newCircuitRelay(isRelayClient: bool): Relay =
     return RelayClient.new()
   return Relay.new()
 
-proc setupCallbacks(node: var WakuNode, callbacks: WakuCallbacks) =
+proc setupCallbacks(
+    node: WakuNode, conf: WakuNodeConf, callbacks: WakuCallbacks
+): Result[void, string] =
   if callbacks.isNil():
     return
 
-  if not callbacks.onReceivedMessage.isNil():
-    
+  if not callbacks.onReceivedMessage.isNil() and not node.wakuRelay.isNil():
+    let autoShards = node.getAutoshards(conf.contentTopics).valueOr:
+      return err("Could not get autoshards: " & error)
+
+    let confShards =
+      conf.shards.mapIt(RelayShard(clusterId: conf.clusterId, shardId: uint16(it)))
+    let shards = confShards & autoShards
+
+    for shard in shards:
+      discard node.wakuRelay.subscribe($shard, callbacks.onReceivedMessage)
 
 proc new*(
     T: type Waku, confCopy: var WakuNodeConf, callbacks: WakuCallbacks = nil
@@ -237,7 +247,9 @@ proc new*(
   let node = nodeRes.get()
 
   if not callbacks.isNil():
-    node.setupCallbacks(callbacks)
+    node.setupCallbacks(confCopy, callbacks).isOkOr:
+      error "Failed setting callbacks", error = error
+      return err("Failed setting up node: " & $error)
 
   ## Delivery Monitor
   var deliveryMonitor: DeliveryMonitor
--- a/waku/factory/waku.nim
+++ b/waku/factory/waku.nim
@@ -148,6 +148,9 @@ proc newCircuitRelay(isRelayClient: bool): Relay =
     return RelayClient.new()
   return Relay.new()
 
+proc setupCallbacks(node: var WakuNode, callbacks: WakuCallbacks) =
+  return
+
 proc new*(T: type Waku, confCopy: var WakuNodeConf): Result[Waku, string] =
   let rng = crypto.newRng()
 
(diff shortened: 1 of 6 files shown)


waku/factory/waku.nim
```diff
@@ -146,7 +148,32 @@ proc newCircuitRelay(isRelayClient: bool): Relay =
...
+
+    let autoShards = node.getAutoshards(conf.contentTopics).valueOr:
+      return err("Could not get autoshards: " & error)
+
+    let confShards =
+      conf.shards.mapIt(RelayShard(clusterId: conf.clusterId, shardId: uint16(it)))
+    let shards = confShards & autoShards
+
+    for shard in shards:
+      discard node.wakuRelay.subscribe($shard, callbacks.relayHandler)
```
Ivansete-status, line 170: Let's avoid using `discard` :)

waku/factory/waku.nim
```diff
@@ -146,7 +148,32 @@ proc newCircuitRelay(isRelayClient: bool): Relay =
...
 
-proc new*(T: type Waku, confCopy: var WakuNodeConf): Result[Waku, string] =
+proc setupCallbacks(
+    node: WakuNode, conf: WakuNodeConf, callbacks: WakuCallbacks
+): Result[void, string] =
+  if callbacks.isNil():
+    info "No external callbacks to be set"
+    return ok()
+
+  if not callbacks.relayHandler.isNil():
```
Ivansete-status, line 158: Super nitpick comment but I think the code gets a bit linear if:

```suggestion
  if callbacks.relayHandler.isNil():
     return err("error in setupCallbacks callbacks.relayHandler.isNil")
```

waku/factory/waku_callbacks.nim
```diff
@@ -0,0 +1,4 @@
+import ../waku_relay/protocol
+
+type WakuCallbacks* = ref object
```
Ivansete-status, line 3: I have the impression that this new type is not needed and might cause confusion with an already existing type: `WakuCallback`

Ivansete-status (commented):
Thanks so much for it! I like the concept very much but I think we need to address a couple of minor details :)
Nice PR! :partying_face:

waku/factory/waku.nim
gabrielmer: The thing is that the procedure returns the callback that gets registered

https://github.com/waku-org/nwaku/blob/ab0c1d4aa0e545c2c74b7e1ea8045823370eee4d/waku/waku_relay/protocol.nim#L423-L425

There's actually nothing to do with it, it doesn't give us information and we don't need that proc for anything

We also use discard in other places because of this same reason

https://github.com/waku-org/nwaku/blob/ab0c1d4aa0e545c2c74b7e1ea8045823370eee4d/waku/node/waku_node.nim#L241

we could change the return type of wakuRelay's subscribe to void if it isn't used anywhere, but I guess it is out of the scope of this PR

waku/factory/waku.nim
gabrielmer: The issue is that it isn't an error, and we also shouldn't return because the idea is to configure all the application level callbacks provided from that procedure. So if no application level callback is passed for Relay, then we don't configure it and will check for other ones (for example, the callback to run on topic health change).

It's a logic similar to when we try to setup and mount protocols

waku/factory/waku_callbacks.nim
gabrielmer: Mmm I agree that it can be confused with libwaku's `WakuCallback`, we can change the name to `AppCallbacks` or something like that.

Whether it's needed or not, the reason for its existence is to be able to create a Waku instance and pass different callbacks that we support to set at the application level.

In a following PR I will add one more callback for a different case, and we will probably need to add more in the near future.

So instead of passing a different parameter for every callback we support to set at the application level, I thought it was easier to have a type with all the callbacks that we support and pass only that object, instead of one new parameter for every new app level callback which can get pretty long and messy.

Any idea on a different solution to this?

waku/factory/waku.nim
Ivansete-status: ah yes, I absolutely agree. It seems then that we have something not designed properly and we need to address , in a separate one :)

waku/factory/waku.nim
Ivansete-status: What if we just use the following at the beginning?
```code
  if callbacks.isNil() or callbacks.relayHandler.isNil():
    info "No external callbacks to be set"
    return ok()
```

waku/factory/waku_callbacks.nim
Ivansete-status: Okay perfect, I'm happy with `AppCallbacks`

Ivansete-status (approved):

waku/factory/waku.nim
gabrielmer: but this doesn't scale well when more callbacks are provided, it would look like

```
if callbacks.isNil() or (callbacks.relayHandler.isNil() and callbacks.otherHandler1.isNil() and callbacks.otherHandler2.isNil(), ...):
    info "No external callbacks to be set"
    return ok()
```

gabrielmer:
> Would `set_event_callback` stil lbe necessary with this new `AppCallbacks` feature?

My understanding is that it is still necessary because there's two type of callbacks - the callbacks at the nwaku's application level which in libwaku we use to emit events (the type of callback of this PR), and a callback defined at libwaku's client to handle all these events emitted by the application level callbacks.

Or in other words the callbacks of this PRs are run whenever something in nwaku happens, for example when a Relay message is received.
The nwaku user can execute whatever (Nim) code they want whenever that happens. And in libwaku we take advantage of it by emitting events on those callbacks, which will be then handled by the  callback set in libwaku's `set_event_callback`

Let me know if I might be missing something or if you think of a way to simplify this structure :)

PR #3207, opened 2024-12-11
fix(libwaku): waku_relay_unsubscribe

richard-ramos:
SUBSCRIBE was being called instead of UNSUBSCRIBE


Ivansete-status (approved):

gabrielmer (approved):
Thank you!

PR #3209, opened 2024-12-13
chore: enhance libwaku store

Ivansete-status:
# Description
Enhancements based on the needs found when recovering the `toy-chat` example from https://github.com/waku-org/waku-rust-bindings


Ivansete-status:
This PR is duplicated and not needed

PR #3211, opened 2024-12-16
feat(libwaku): add protected shard

richard-ramos:
## Parent Issue
- https://github.com/waku-org/nwaku/issues/3076

### DIFF
--- a/library/libwaku.h
+++ b/library/libwaku.h
@@ -84,6 +84,13 @@ int waku_relay_subscribe(void* ctx,
                          WakuCallBack callback,
                          void* userData);
 
+int waku_relay_add_protected_shard(void* ctx,
+                                   int clusterId,
+                                   int shardId,
+                                   char* publicKey,
+                                   WakuCallBack callback,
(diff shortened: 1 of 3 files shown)


gabrielmer:
Thanks so much!

Random question - shouldn't we keep with the "protected-shard" naming and input cluster and shard instead of topic for consistency?
Not sure if that would give trouble from `status-go` side, I guess it's just a matter of reformatting the input, but it would make libwaku and nwaku consistent

richard-ramos:
> Random question - shouldn't we keep with the "protected-shard" naming and input cluster and shard instead of topic for consistency?
> Not sure if that would give trouble from `status-go` side, I guess it's just a matter of reformatting the input, but it would make libwaku and nwaku consistent

Sounds like a good idea. Will do that change

gabrielmer (approved):
Thanks so much!

Ivansete-status (approved):
LGTM! Thanks for it! :100: 
I just added a nitpick suggestion

PR #3213, opened 2024-12-17
feat: waku store sync 2.0 common types & codec

SionoiS:
# Description

First PR for the new Waku Store Sync 2.0

Include types and encode/decode of payloads and also tests.

[Specification](https://github.com/waku-org/specs/blob/feat--waku-sync-2/standards/core/sync.md)
[Research issue](https://github.com/waku-org/research/issues/102)

# Changes

- [x] Common types
- [x] Codec
- [x] Tests

Followed by #3215

# Issue
- https://github.com/waku-org/nwaku/issues/3221


waku/waku_store_sync/codec.nim
```diff
@@ -0,0 +1,283 @@
...
+proc encode*(value: WakuMessagePayload): ProtoBuffer =
+  var pb = initProtoBuffer()
+
+  pb.write3(1, value.pubsub)
+  pb.write3(2, value.message.encode())
+
+  return pb
+
+proc deltaEncode*(itemSet: ItemSet): seq[byte] =
+  let capacity = 1 + (itemSet.elements.len * 41)
```
gabrielmer, line 16: where does 41 come from?

waku/waku_store_sync/codec.nim
```diff
@@ -0,0 +1,283 @@
...
+  let (bound, _) = value.ranges[0]
+
+  lastTimestamp = bound.a.time
+  lastHash = bound.a.fingerprint
+
+  # encode first timestamp
+  buf = uint64(lastTimestamp).toBytes(Leb128)
+  output &= @buf
+
+  #echo "First Timestamp: " & $lastTimestamp
```
gabrielmer, line 59: There's lots of echo comments in this file that we should probably remove 😄

gabrielmer (commented):
Amazinggg, thanks so much! 😍 

Honestly, I think I should understand more about range based reconciliation in order to follow what's going on here.

Any article or resource you recommend to read and use as a reference? I probably have some homework to do 😬

waku/waku_store_sync/codec.nim
SionoiS: LOL true! I'll remove those.

SionoiS:
> Amazinggg, thanks so much! 😍
> 
> Honestly, I think I should understand more about range based reconciliation in order to follow what's going on here.
> 
> Any article or resource you recommend to read and use as a reference? I probably have some homework to do 😬

The goto should my research issue. https://github.com/waku-org/research/issues/102

Although a lot of details are implementation specific which is not mentioned anywhere :thinking: maybe more code comment are needed.

waku/waku_store_sync/codec.nim
SionoiS: 32 byte hash + 9 byte varint

waku/waku_store_sync/common.nim
```diff
@@ -0,0 +1,80 @@
...
+  SyncReconciliationCodec* = "/vac/waku/reconciliation/1.0"
+  SyncTransferCodec* = "/vac/waku/transfer/1.0"
+  DefaultGossipSubJitter*: Duration = 20.seconds
+
+type
+  Fingerprint* = array[32, byte]
+
+  ID* = object
+    time*: Timestamp
+    fingerprint*: Fingerprint
```
SionoiS, line 20: I'm thinking of changing this as `fingerprint` should be used only for incremental not msg hashes.

jm-clius (changes requested):
Thanks a lot for the (significant!) effort, also in the follow-up PRs! 🚀  I've requested some changes, mostly related to helping reviewers understand the code/protocol a bit better. For example, there are some magic numbers that I still can't quite figure out, some codec types don't quite match the spec or need some context, etc.

For this PR, I'd be happy to approve if the code is made clearer (with comments, reduced magic numbers, etc.). However, I think it will be easier for this (and follow-up) PRs to expand the wire protocol specification itself and get an agreed-upon design as spec first so that reviewers wouldn't have to infer how the protocol works only from code.

waku/waku_store_sync/codec.nim
SionoiS: I'll add a code comment.

It's just to avoid reallocation, it has nothing to do with the spec, it's not even necessary.

waku/waku_store_sync/common.nim
```diff
@@ -0,0 +1,80 @@
...
+
+  RangeType* = enum
+    skipRange = 0
+    fingerprintRange = 1
+    itemSetRange = 2
+
+  SyncPayload* = object
+    ranges*: seq[(Slice[ID], RangeType)]
+    fingerprints*: seq[Fingerprint]
+    itemSets*: seq[ItemSet]
```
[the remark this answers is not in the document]
SionoiS: At this stage, the impl. is 100% equivalent to the previous one in features. There's no negotiation. We can start designing for limited shard support in my research issue if you want.

waku/waku_store_sync/codec.nim
```diff
@@ -0,0 +1,233 @@
...
+    let id = ID(time: time, fingerprint: fingerprint)
+
+    itemSet.elements.add(id)
+
+  itemSet.reconciled = bool(buffer[idx])
+  idx += 1
+
+  return idx
+
+proc deltaDecode*(T: type SyncPayload, buffer: seq[byte]): T =
```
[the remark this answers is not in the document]
SionoiS: I don't think it's a good idea to add links in the code. Link always end up broken anyway.

Maybe a section at the top explaining that there's a spec?

waku/waku_store_sync/common.nim
```diff
@@ -0,0 +1,78 @@
...
+  DefaultGossipSubJitter*: Duration = 20.seconds
+
+type
+  Fingerprint* = array[32, byte]
+
+  ID* = object
+    time*: Timestamp
+    hash*: WakuMessageHash
+
+  ItemSet* = object
```
Ivansete-status, line 20: I think every type, and its procs, should be contained in a separate file

waku/waku_store_sync/common.nim
```diff
@@ -0,0 +1,80 @@
+{.push raises: [].}
+
+import std/[options], chronos, stew/[byteutils]
+
+import ../waku_core
+
+const
+  DefaultSyncInterval*: Duration = 5.minutes
+  DefaultSyncRange*: Duration = 1.hours
+  RetryDelay*: Duration = 30.seconds
```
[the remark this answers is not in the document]
SionoiS: Ah it's a left over and never used. I'll remove it.

waku/waku_store_sync/common.nim
SionoiS: Come on, the file is not even 100 lines.

waku/waku_store_sync/codec.nim
```diff
@@ -0,0 +1,238 @@
...
+    case rangeType
+    of RangeType.Skip:
+      continue
+    of RangeType.Fingerprint:
+      let fingerprint = value.fingerprints[i]
+      i.inc()
+
+      output &= fingerprint
+    of RangeType.ItemSet:
+      let itemSet = value.itemSets[j]
```
Ivansete-status, line 100: I'm a bit concerned about out-of-bounds issues in both `Fingerprint` and `ItemSet` ranges

waku/waku_store_sync/codec.nim
```diff
@@ -0,0 +1,238 @@
...
+    let id = SyncID(time: time, hash: hash)
+
+    itemSet.elements.add(id)
+
+  itemSet.reconciled = bool(buffer[idx])
+  idx += 1
+
+  return idx
+
+proc deltaDecode*(T: type RangesData, buffer: seq[byte]): T =
```
Ivansete-status, line 145: May it be possible to split this proc a little? It is a little bit large and difficult to follow :)

Ivansete-status (approved):
Thanks for it! Approving to not block it although I added a few nitpick comments that I hope you find interesting
Thanks for the patience!

waku/waku_store_sync/codec.nim
```diff
@@ -0,0 +1,238 @@
+{.push raises: [].}
+
+import std/sequtils, stew/leb128
+
+import ../common/protobuf, ../waku_core/message, ../waku_core/time, ./common
+
+const
+  HashLen = 32
+  VarIntLen = 9
```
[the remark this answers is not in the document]
SionoiS: VarInts are max 9 bytes to express big numbers, since each byte use only 7 bits.

waku/waku_store_sync/codec.nim
```diff
@@ -0,0 +1,238 @@
...
+  # 1 byte for resolved bool and 32 bytes hash plus 9 bytes varint per elements 
+  let capacity = 1 + (itemSet.elements.len * (VarIntLen + HashLen))
+
+  var
+    output = newSeqOfCap[byte](capacity)
+    lastTime = Timestamp(0)
+    buf = Leb128Buf[uint64]()
+
+  for id in itemSet.elements:
+    let timeDiff = uint64(id.time) - uint64(lastTime)
```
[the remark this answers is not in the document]
SionoiS: :thinking: we could use seconds instead of nanos. Time diffs would be smaller but hashes would have to be included more often, so IDK if it's worth it.

waku/waku_store_sync/codec.nim
```diff
@@ -0,0 +1,238 @@
...
+
+  for (bound, rangeType) in value.ranges:
+    let timeDiff = uint64(bound.b.time) - uint64(lastTimestamp)
+    lastTimestamp = bound.b.time
+
+    # encode timestamp
+    buf = timeDiff.toBytes(Leb128)
+    output &= @buf
+
+    if timeDiff == 0:
```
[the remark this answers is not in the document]
SionoiS: It can because our timestamp have nano precision but half seconds accuracy.

waku/waku_store_sync/codec.nim
SionoiS: You can't go out of bound because the outer loop is visiting each ranges.

waku/waku_store_sync/codec.nim
NagyZoltanPeter, line 205: Maybe, not just split this proc, but for me it is not obvious how it prevents malicious flows in the decoded buffer. 
Sure I miss something, but how do we validate the correctness of the input here?
This relates somehow the raised concern by @Ivansete-status that how are we sure the Ranges/Itemsets and fingerprints are correct and no over indexing could happen. (https://github.com/waku-org/nwaku/pull/3213/files/f2f0e648236eb5bad4745208368defe93585329f#r1920192288)

SionoiS: In the case of specifically sending malicious payload?

Yes in that case your right, we have a problem. I'll try to think of something.

SionoiS: @NagyZoltanPeter and @Ivansete-status I made sure that no malformed data can crash the decoder.

NagyZoltanPeter (approved):
Thanks for it, after the revised decoding I think it is ok!

jm-clius (approved):
Thanks!

PR #3214, opened 2024-12-17
chore: modifying relay subscribe return value

gabrielmer:
# Description
Currently, `WakuRelay.subscribe()` has a return value that is discarded everywhere except in one place. It returns the wrapped handler that is passed to gossipsub, which is useless to most cases when we subscribe to a topic.

Therefore, defined a `wrapHandler()` procedure to return the wrapped handler in the specific cases that is needed, and changed the return type of `subscribe` to void so we don't have to discard that value everywhere 

# Changes

- [x] changed return value of `WakuRelay.subscribe()` to void
- [x] defined a `wrapHandler()` procedure in the `waku_relay` module

### DIFF
--- a/library/waku_thread/inter_thread_communication/requests/protocols/relay_request.nim
+++ b/library/waku_thread/inter_thread_communication/requests/protocols/relay_request.nim
@@ -94,7 +94,7 @@ proc process*(
   case self.operation
   of SUBSCRIBE:
     # TO DO: properly perform 'subscribe'
-    discard waku.node.wakuRelay.subscribe($self.pubsubTopic, self.relayEventCallback)
+    waku.node.wakuRelay.subscribe($self.pubsubTopic, self.relayEventCallback)
   of UNSUBSCRIBE:
     # TODO: properly perform 'unsubscribe'
     waku.node.wakuRelay.unsubscribeAll($self.pubsubTopic)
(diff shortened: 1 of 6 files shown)


gabrielmer:
Well, looking at the failed test I noticed that we do need to pass the wrapped handler when calling unsubscribe (as unsubscribing is from a specific handler). So maybe it does sense to return it and this PR is irrelevant? LMK what you think

Ivansete-status:
> Well, looking at the failed test I noticed that we do need to pass the wrapped handler when calling unsubscribe (as unsubscribing is from a specific handler). So maybe it does sense to return it and this PR is irrelevant? LMK what you think

Thanks for this!
IMO, we need to allow subscription and unsubscription but it shouldn't be needed to deal with any handler at all to unsubscribe. The only parameters needed to subscribe/unsubscribe should be the `pubsubTopic` , a.k.a. `cluster-id`+`shard`

gabrielmer:
> Thanks for this! IMO, we need to allow subscription and unsubscription but it shouldn't be needed to deal with any handler at all to unsubscribe. The only parameters needed to subscribe/unsubscribe should be the `pubsubTopic` , a.k.a. `cluster-id`+`shard`

I understand not dealing with handlers when unsubscribing, but I think we do need to have a handler as a parameter when subscribing right? the idea of subscribing is running a custom procedure when a message is received, it seems as it is really important to be able to define and pass that procedure through. But I might be missing something :))

gabrielmer:
After further discussions, opened issue https://github.com/waku-org/nwaku/issues/3222 and closing this PR :)

PR #3215, opened 2024-12-17
feat: waku store sync 2.0 storage and tests

SionoiS:
# Description

Second PR for new Waku store sync 2.0

Include everything Storage related and tests

[Specification](https://github.com/waku-org/specs/blob/feat--waku-sync-2/standards/core/sync.md)
[Research issue](https://github.com/waku-org/research/issues/102)

# Changes

- [x] Storage interface
- [x] Storage impl.
- [x] tests

N.B. Cannot be merged before #3213
Followed by #3216 

# Issue
- https://github.com/waku-org/nwaku/issues/3221

### DIFF
--- a/waku/waku_store_sync/storage/seq_storage.nim
+++ b/waku/waku_store_sync/storage/seq_storage.nim
@@ -1,0 +1,287 @@
+import std/[algorithm, sequtils, math, options], results, chronos, stew/arrayops
+
+import ../../waku_core/time, ../common, ./range_processing, ./storage
+
+type SeqStorage* = ref object of SyncStorage
+  elements: seq[ID]
+
+  partitionCount: int
+  lengthThreshold: int
+
+method insert*(self: SeqStorage, element: ID): Result[void, string] {.raises: [].} =
+  let idx = self.elements.lowerBound(element, common.cmp)
+
+  if idx < self.elements.len and self.elements[idx] == element:
+    return err("duplicate element")
+
+  self.elements.insert(element, idx)
+
+  return ok()
+
+method batchInsert*(
+    self: SeqStorage, elements: seq[ID]
+): Result[void, string] {.raises: [].} =
+  ## Insert the sorted seq of new elements.
+
+  if elements.len == 1:
+    return self.insert(elements[0])
+
+  #TODO custom impl. ???
+
+  if not elements.isSorted(common.cmp):
+    return err("seq not sorted")
+
+  var merged = newSeqOfCap[ID](self.elements.len + elements.len)
+
+  merged.merge(self.elements, elements, common.cmp)
+
+  self.elements = merged.deduplicate(true)
+
+  return ok()
+
+method prune*(self: SeqStorage, timestamp: Timestamp): int {.raises: [].} =
+  ## Remove all elements before the timestamp.
+  ## Returns # of elements pruned.
+
+  let bound = ID(time: timestamp, fingerprint: EmptyFingerprint)
+
+  let idx = self.elements.lowerBound(bound, common.cmp)
+
+  self.elements.delete(0 ..< idx)
+
+  return idx
+
+proc fingerprinting(self: SeqStorage, sliceOpt: Option[Slice[int]]): Fingerprint =
+  ## XOR all hashes of a slice of the storage.
+
+  var fingerprint = EmptyFingerprint
+
+  let idxSlice =
+    if sliceOpt.isNone():
+      return fingerprint
+    else:
+      sliceOpt.get()
+
+  for id in self.elements[idxSlice]:
+    fingerprint = fingerprint xor id.fingerprint
+
+  return fingerprint
+
+proc findIdxBounds(self: SeqStorage, slice: Slice[ID]): Option[Slice[int]] =
+  ## Given bounds find the corresponding indices in this storage
+
+  #TODO can thoses 2 binary search be combined for efficiency ???
+
+  let lower = self.elements.lowerBound(slice.a, common.cmp)
+  var upper = self.elements.upperBound(slice.b, common.cmp)
+
+  if upper < 1:
+    # entire range is before any of our elements
+    return none(Slice[int])
+
+  if lower >= self.elements.len:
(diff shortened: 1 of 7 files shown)


jm-clius (approved):
Right, went through this in some detail. lgtm! Good job :). Would it be useful to have a set of unit tests for each public method (`insert`, `batchInsert`, `prune`) to test the boundary conditions at each?

SionoiS:
> Right, went through this in some detail. lgtm! Good job :). Would it be useful to have a set of unit tests for each public method (`insert`, `batchInsert`, `prune`) to test the boundary conditions at each?

Yes, I will add some.

waku/waku_store_sync/storage/storage.nim
```diff
@@ -0,0 +1,33 @@
+import std/math, results, chronos
+
+import ../../waku_core/time, ../common
+
+type SyncStorage* = ref object of RootObj
+
+method insert*(
+    self: SyncStorage, element: ID
+): Result[void, string] {.base, gcsafe, raises: [].} =
+  discard
```
SionoiS, line 10: Maybe `insert` should be idempotent?

waku/waku_store_sync/storage/seq_storage.nim
```diff
@@ -0,0 +1,290 @@
...
+
+  if upper < 1:
+    # entire range is before any of our elements
+    return none(Slice[int])
+
+  if lower >= self.elements.len:
+    # entire range is after any of our elements
+    return none(Slice[int])
+
+  return some(lower ..< upper)
```
Ivansete-status, line 89: Should the returned `Slice` not include the upper too?
```suggestion
  return some(lower .. upper)
```

waku/waku_store_sync/storage/seq_storage.nim
```diff
@@ -0,0 +1,290 @@
...
+
+  if ourFingerprint == inputFingerprint:
+    output.ranges.add((inputBounds, skipRange))
+    return
+
+  let slice =
+    if idxSlice.isNone():
+      output.ranges.add((inputBounds, itemSetRange))
+      let state = ItemSet(elements: @[], reconciled: true)
+      output.itemSets.add(state)
```
Ivansete-status, line 116: I'm a bit puzzled about these two lines :)
Are we trying to "return" an empty `SyncPayload`?

waku/waku_store_sync/storage/seq_storage.nim
```diff
@@ -0,0 +1,290 @@
...
+    else:
+      idxSlice.get()
+
+  if slice.len <= self.lengthThreshold:
+    output.ranges.add((inputBounds, itemSetRange))
+    let state = ItemSet(elements: self.elements[slice], reconciled: false)
+    output.itemSets.add(state)
+    return
+
+  for partitionBounds in equalPartitioning(inputBounds, self.partitionCount):
```
Ivansete-status, line 127: I think it is better to invoke `equalPartitioning` once and later iterate over the result. I believe we should only use that approach if `equalPartitioning` defined some kind of `iterator` with `yield` inside.

waku/waku_store_sync/storage/seq_storage.nim
```diff
@@ -0,0 +1,290 @@
...
+      if not inputItemSet.reconciled:
+        output.ranges.add((inputBounds, itemSetRange))
+        let state = ItemSet(elements: @[], reconciled: true)
+        output.itemSets.add(state)
+      else:
+        output.ranges.add((inputBounds, skipRange))
+
+      return
+    else:
+      idxSlice.get()
```
Ivansete-status, line 173: I see some room for encapsulation as apparently we have similar approach two or three times

Ivansete-status (approved):
Thanks for it :partying_face: 
Approving to not block
I just added some nitpick comments ;P

waku/waku_store_sync/storage/storage.nim
SionoiS: Currently if you can insert with the same element the first call would be `ok()` but if you call again `err()` since the element was already present.

An idempotent function would always insert the same element at the same index and would result in no side effect when calling the same func with the same params multiple times.

waku/waku_store_sync/storage/seq_storage.nim
SionoiS: > I think it is better to invoke `equalPartitioning` once and later iterate over the result.

Is that not how it works under the hood?

waku/waku_store_sync/storage/seq_storage.nim
SionoiS: :thinking: I'm pretty sure that the `Slice[ID]` exclude the upper bound so we do the same.

waku/waku_store_sync/storage/seq_storage.nim
```diff
@@ -0,0 +1,290 @@
...
+    self: SeqStorage, bounds: Slice[ID]
+): Fingerprint {.raises: [].} =
+  let idxSliceOpt = self.findIdxBounds(bounds)
+  return self.fingerprinting(idxSliceOpt)
+
+proc processFingerprintRange*(
+    self: SeqStorage,
+    inputBounds: Slice[ID],
+    inputFingerprint: Fingerprint,
+    output: var SyncPayload,
```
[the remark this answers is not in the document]
SionoiS: Well there's no way for the proc to fail.

Also, the `var payload` could have been local to the proc but the caller has a loop so it's inconvenient.

waku/waku_store_sync/storage/seq_storage.nim
SionoiS: Yes it's a valid case where one side have an empty range.

Peer 1:
range A to F
fingerprint 0x53...

Peer 2:
range A to F
no elements :(
send back an empty item set

waku/waku_store_sync/storage/seq_storage.nim
SionoiS: They are slightly different. One has a `return` the other `continue`

PR #3225, opened 2024-12-24
feat: connection change event

gabrielmer:
# Description
Adding a connection health change event which gets triggered every time a peer connects or disconnects. When the event is triggered, a new app callback named `connectionChangeHandler` is called. This PR also integrates the event to libwaku.

# Changes

- [x] add a connection change event
- [x] integrating the event to libwaku
- [x] created an `eventCallback` template in libwaku to avoid code duplication 

## Issue
#3076

### DIFF
--- a/library/libwaku.nim
+++ b/library/libwaku.nim
@@ -47,6 +47,29 @@ template checkLibwakuParams*(
   if isNil(callback):
     return RET_MISSING_CALLBACK
 
+template eventCallback(ctx: ptr WakuContext, eventName: string, body: untyped) =
+  if isNil(ctx[].eventCallback):
+    error eventName & " - eventCallback is nil"
+    return
+
+  if isNil(ctx[].eventUserData):
+    error eventName & " - eventUserData is nil"
+    return
+
+  foreignThreadGc:
+    try:
+      let event = body
+      cast[WakuCallBack](ctx[].eventCallback)(
+        RET_OK, unsafeAddr event[0], cast[csize_t](len(event)), ctx[].eventUserData
+      )
+    except Exception, CatchableError:
+      let msg =
+        "Exception " & eventName & " when calling 'eventCallBack': " &
+        getCurrentExceptionMsg()
+      cast[WakuCallBack](ctx[].eventCallback)(
+        RET_ERR, unsafeAddr msg[0], cast[csize_t](len(msg)), ctx[].eventUserData
+      )
+
 proc handleRequest(
     ctx: ptr WakuContext,
     requestType: RequestType,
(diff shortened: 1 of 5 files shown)


library/libwaku.nim
```diff
@@ -45,6 +47,29 @@ template checkLibwakuParams*(
   if isNil(callback):
     return RET_MISSING_CALLBACK
 
+template eventCallback(ctx: ptr WakuContext, eventName: string, body: untyped) =
```
Ivansete-status, line 50: I think it is better to rename it differently from the actual `callback` item
```suggestion
template callEventCallback(ctx: ptr WakuContext, eventName: string, body: untyped) =
```

waku/node/peer_manager/peer_manager.nim
```diff
@@ -676,6 +681,9 @@ proc onPeerEvent(pm: PeerManager, peerId: PeerId, event: PeerEvent) {.async.} =
           debug "Pruning connection due to ip colocation", peerId = peerId, ip = ip
           asyncSpawn(pm.switch.disconnect(peerId))
           pm.wakuPeerStore.delete(peerId)
+    if not pm.onConnectionChange.isNil():
+      # we don't want to await for the callback to finish
```
Ivansete-status, line 685: Why not? I think it is better to avoid the use of `discard` and perform `await`

waku/node/peer_manager/peer_manager.nim
```diff
@@ -87,6 +91,7 @@ type PeerManager* = ref object of RootObj
   colocationLimit*: int
   started: bool
   shardedPeerManagement: bool # temp feature flag
+  onConnectionChange*: ConnectionChangeHandler
```
Ivansete-status, line 94: I think it is better to enforce this parameter to the `new` proc and then we can make sure it is always `not isNil():`
```suggestion
  onConnectionChange: ConnectionChangeHandler
```

Ivansete-status (commented):
Thanks so much for it! Master piece PR! :partying_face: 
I'm not approving just yet as I think we can simplify it a little bit
Congrats for it and good work!

waku/node/peer_manager/peer_manager.nim
gabrielmer: But I think that setting up this callback should be optional right? how can it be always different than nil? there's the option of having an empty proc as a default callback but not sure that's a good idea

On the other hand, what's the benefit of only being able to set it from Peer Manager's `new` proc? As this is something that the application is supposed to be able to set, it makes sense to allow to set it on demand beyond the creation of the Peer Manager

waku/node/peer_manager/peer_manager.nim
gabrielmer: I commented it because we don't need to stop the execution of the procedure waiting for the callback to finish. It is an external callback that we don't know what it does and is not related to anything else, so I think there's no point in waiting for it in order to continue with other things.

I might be missing something though. I know using `discard` is not the best practice, but then how should we avoid waiting for an async proc that returns void (so no value being ignored), and that we don't need to wait for?

Ivansete-status: ah I see, maybe we can use `asyncSpawn`? I think it may work well considering temporary async tasks

gabrielmer: Ooh great idea! Changing the `discard` here and in the other place in the proc for `asyncSpawn`s :))

Thank you!

waku/node/peer_manager/peer_manager.nim
Ivansete-status: Got it! Thanks for clarifying!
At first I thought that enforcing it in `PeerManager.new` would make things simpler but I think that way it gives more flexibility

Ivansete-status (approved):

richard-ramos (approved):

PR #3227, opened 2025-01-02
test: test_wakunode_rln_relay use waitForNullifierLog in all tests avoid flaky

Ivansete-status:
Extend @darshankabariya 's approach to all tests to avoid RLN flaky tests at any point

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -118,7 +118,7 @@ jobs:
           export MAKEFLAGS="-j1"
           export NIMFLAGS="--colors:off -d:chronicles_colors:none"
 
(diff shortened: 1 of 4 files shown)


darshankabariya (approved):
Look great to me ! Thanks ivan

NagyZoltanPeter (approved):
That's pretty cool!
Only concern about how long we will need such amount of debug logs here? Does it really helps this way?
Also maybe it can be interesting what is the average wait time needed for nullifier logs. WDYT?

Ivansete-status:
> That's pretty cool! Only concern about how long we will need such amount of debug logs here? Does it really helps this way?

This test is pretty annoying and I have the impression that these tests may complain in the nearby future and this is why it may help us to debug. I just created a reminder for next month and we can check if we can clean it up.

--------------

Also maybe it can be interesting what is the average wait time needed for nullifier logs. WDYT?

Yes this would be interesting to measure but not for now

PR #3233, opened 2025-01-08
chore: Libwaku tweaks

Ivansete-status:
## Description
- Avoid the use of `string` type in libwaku. In fact, we should avoid any GC'ed type (`string`, `ref`, `closure`, `seq`) when performing ffi operations. Instead, we can only allocate/deallocate memory manually
- Initialize default `cstring` params with `nil` instead of an empty `string` (`""`)
- library/alloc.nim: reserve an empty cstring when `nil` is passed to `alloc` proc
- lightpush_request.nim, self_req_handler.nim: return a `msg_hash` when successful publish
- library/waku_thread/inter_thread_communication/requests/protocols/store_request.nim: rm extra `deallocShared`

### DIFF
--- a/library/alloc.nim
+++ b/library/alloc.nim
@@ -4,6 +4,11 @@ type SharedSeq*[T] = tuple[data: ptr UncheckedArray[T], len: int]
 proc alloc*(str: cstring): cstring =
   # Byte allocation from the given address.
   # There should be the corresponding manual deallocation with deallocShared !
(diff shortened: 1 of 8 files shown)


gabrielmer (approved):
Looks amazing! Thanks so much! 🔥

richard-ramos (approved):
:rocket:

PR #3234, opened 2025-01-08
chore: sending msg hash as string on libwaku message event

gabrielmer:
# Description
As previously discussed, it makes more sense for application clients to use message hashes as hex strings instead of a string representation of a byte array.

In this PR, we're modifying the usage of message hashes in libwaku and in the REST API to receive and return the message hashes as hex strings. This change only applies for the interactions with the message hashes and with pagination cursors (which are themselves message hashes). Everything else remains untouched and other fields such as payload and meta are still sent as base64 strings.

# Changes

- [x] creating `WakuMessageKeyValueHex` and `StoreQueryResponseHex` types to be used when interacting with application clients
- [x] returning message hashes in libwaku's `JsonMessageEvent` and `waku_store_query` as hex strings
- [x] returning message hashes in Store REST API as hex strings
- [x] updating tests 

A PR updating the REST API's docs will be opened once this PR is merged. Or shall we define a new endpoint for backward compatibility? Another option is to accept both byte arrays and hex strings in the API and check in the endpoint which one it is

## Issue

#3076

### DIFF
--- a/library/events/json_message_event.nim
+++ b/library/events/json_message_event.nim
@@ -59,7 +59,7 @@ proc `%`*(value: Base64String): JsonNode =
 
 type JsonMessageEvent* = ref object of JsonEvent
   pubsubTopic*: string
-  messageHash*: WakuMessageHash
+  messageHash*: string
   wakuMessage*: JsonMessage
 
 proc new*(T: type JsonMessageEvent, pubSubTopic: string, msg: WakuMessage): T =
@@ -83,7 +83,7 @@ proc new*(T: type JsonMessageEvent, pubSubTopic: string, msg: WakuMessage): T =
   return JsonMessageEvent(
     eventType: "message",
     pubSubTopic: pubSubTopic,
-    messageHash: msgHash,
+    messageHash: msgHash.to0xHex(),
     wakuMessage: JsonMessage(
       payload: base64.encode(payload),
       contentTopic: msg.contentTopic,
--- a/library/waku_thread/inter_thread_communication/requests/protocols/store_request.nim
+++ b/library/waku_thread/inter_thread_communication/requests/protocols/store_request.nim
@@ -40,7 +40,7 @@ func fromJsonNode(
     else:
       none(string)
 
-  let paginationCursor: Option[WakuMessageHash] =
+  let paginationCursor =
     if jsonContent.contains("pagination_cursor"):
       let hash = jsonContent["pagination_cursor"].getStr().hexToHash().valueOr:
           return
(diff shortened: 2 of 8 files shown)


Ivansete-status:
@gabrielmer - ah, sorry but I can't approve that one now, unless you convince me hehe. I think is fine passing the `msg_hash` as byte array because is more efficient than dealing with string. On the other hand, the host code (Go, Rust, etc) should easily show the msg_hash representation in a readable format. Let me know if that causes a big trouble somewhere :)

gabrielmer:
> @gabrielmer - ah, sorry but I can't approve that one now, unless you convince me hehe. I think is fine passing the `msg_hash` as byte array because is more efficient than dealing with string. On the other hand, the host code (Go, Rust, etc) should easily show the msg_hash representation in a readable format. Let me know if that causes a big trouble somewhere :)

Well, I don't know if I can convince you about the change itself, but at least I might be able to convince you that the opposite is true of what you mentioned - dealing with a string is more efficient than dealing with the byte array lol.

If you see, the events are not serialized and deserialized, they are sent as a json-string: https://github.com/waku-org/nwaku/blob/625c8ee51bc3e065da0e2e8d3a53d3634589f548/library/libwaku.nim#L92-L95

In other words, when we sent that byte-array, we are sending the string representation of a byte array, not the bytes directly.

The host has to then parse that string into a byte array and then convert it to its string representation. It's both less efficient and it adds extra/unnecessary work from the host's code to have to parse the string representation of a byte array

Ivansete-status:
@gabrielmer - you are absolutely right my friend. I think I'm starting to get convinced ;P

Ivansete-status (approved):
Approving after the agreement to focus on retrieving all the information from libwaku in the most readable format, by default :)

gabrielmer:
Moving this PR back to draft, will reopen it for review once the hex string format is used all across libwaku and REST

library/events/json_message_event.nim
```diff
@@ -59,7 +59,7 @@ proc `%`*(value: Base64String): JsonNode =
 
 type JsonMessageEvent* = ref object of JsonEvent
   pubsubTopic*: string
-  messageHash*: WakuMessageHash
```
gabrielmer, line 62: Maybe instead of string it's better to define a `WakuMessageHashHex` or `HexString` type and use it instead of just string?

LMK if it makes sense and if it adds value

library/events/json_message_event.nim
```diff
@@ -59,7 +59,7 @@ proc `%`*(value: Base64String): JsonNode =
...
+  messageHash*: string
```
gabrielmer, line 62: Maybe it's better to define a `WakuMessageHashHex` or `HexString` type and use it instead of just string?

LMK if it makes sense and if it adds value

richard-ramos (approved):

Ivansete-status (approved):
Before merge I think we need to make sure the following is green:
image

gabrielmer:
> Before merge I think we need to make sure the following is green: image

Good point! Let's check, I'm seeing this error

image

Doesn't seem to be related to the PR, but let's fix it first and see if something else fails afterwards

PR #3235, opened 2025-01-09
chore: golang example end using negentropy dependency plus simple readme.md

Ivansete-status:
- examples/golang/waku.go: Stop linking against negentropy in cgo example.
- Add README.md for golang example.
- In the other files: avoid dependency with `libpcre` library by avoid importing `re` and `nre` and instead, import regex. With that, we will only have dependency with `libpq` (for Postgres client) library.

( thanks @AYAHASSAN287 for the help testing and validating the golang example requirements :partying_face: )

### DIFF
--- a/examples/golang/README.md
+++ b/examples/golang/README.md
@@ -1,13 +1,19 @@
 
 ## Pre-requisite
 libwaku.so is needed to be compiled and present in build folder. To create it:
+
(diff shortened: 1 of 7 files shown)


gabrielmer (approved):
Thank you!

NagyZoltanPeter (approved):
It is all good, great improvement! Thanks for it.

PR #3237, opened 2025-01-13
chore: update lightclient examples to use waku prod fleet and readme 

chaitanyaprem:
# Description
While running lightpush and filter examples connectivity to peers was failing. Realized that metadata is not mounted and hence connectivity keeps failing randomly. 
Updating the same to mount metadata and also use waku sandbox fleet.

Thanks @richard-ramos for helping with this.

Note that i had specified default shard as 0 and clusterID as 1 for metadata support to work. But this is not ideal, maybe we should have a way for the user to not specify shards and just use autosharding API.

# Changes

- [x] Update lightpush and filter examples to mount metadata
- [x] Update examples to use a node from waku sandbox fleet.

### DIFF
--- a/examples/filter_subscriber.nim
+++ b/examples/filter_subscriber.nim
@@ -1,29 +1,38 @@
-## Example showing how a resource restricted client may
-## subscribe to messages without relay
+import
+  std/[tables, sequtils],
+  stew/byteutils,
+  stew/shims/net,
+  chronicles,
+  chronos,
+  confutils,
+  libp2p/crypto/crypto,
+  eth/keys,
+  eth/p2p/discoveryv5/enr
 
-import chronicles, chronos, stew/byteutils, results
-import waku/[common/logging, node/peer_manager, waku_core, waku_filter_v2/client]
+import
+  waku/[
+    common/logging,
+    node/peer_manager,
+    waku_core,
+    waku_node,
+    waku_enr,
+    discovery/waku_discv5,
+    factory/builder,
+    waku_relay,
+    waku_filter_v2/client,
+  ]
+
+# careful if running pub and sub in the same machine
+const wakuPort = 50000
+
+const clusterId = 1
+const shardId = @[0'u16]
 
 const
   FilterPeer =
-    "/ip4/34.16.1.67/tcp/30303/p2p/16Uiu2HAmDCp8XJ9z1ev18zuv8NHekAsjNyezAvmMfFEJkiharitG"
-    # node-01.gc-us-central1-a.waku.test.status.im on waku.test
-  FilterPubsubTopic = PubsubTopic("/waku/2/rs/0/0")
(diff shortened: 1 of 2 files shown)


examples/README.md
```diff
@@ -17,7 +17,7 @@ Within `examples/` you can find a `publisher` and a `subscriber`. The first one
 
 **Some notes:**
 * These examples are meant to work even if you are behind a firewall and you can't be discovered by discv5.
-* You only need to provide a reachable bootstrap peer (see our [fleets](https://fleets.status.im/))
+* You only need to provide a reachable bootstrap peer (see our [fleets](https://fleets.waku.org/))
```
NagyZoltanPeter, line 20: Why is this better? I think https://fleets.status.im/ has better information about bootstrap nodes of available fleets. WDYT?

NagyZoltanPeter (changes requested):
I think the intent is good, to keep examples working out of the box.
But I think wakuv2.prod was about to be decommissioned, and if you check the fleets.waku.org info board it is abandoned. The bootstrap version is pretty old, v0.24.0, instead of v0.34.0 - the current latest.
I would recommend waku-sandbox maybe as a better option - although that has rln as being connected with TWN if I'm right.

Also I think at some point we shall invest some time in enhancing these examples as they are way too simple and don't give the feeling of real application needs and possibilities.
In this example using the same service peer for filter and lightpush is just not what a normal light client should do.
@Ivansete-status WDYT?

chaitanyaprem:
> I think the intent is good, to keep examples working out of the box. But I think wakuv2.prod was about to be decommissioned, and if you check the fleets.waku.org info board it is abandoned. The bootstrap version is pretty old, v0.24.0, instead of v0.34.0 - the current latest. I would recommend waku-sandbox maybe as a better option - although that has rln as being connected with TWN if I'm right.
> 
> Also I think at some point we shall invest some time in enhancing these examples as they are way too simple and don't give the feeling of real application needs and possibilities. In this example using the same service peer for filter and lightpush is just not what a normal light client should do. @Ivansete-status WDYT?

I have updated the code with sandbox fleet and it seems to work consistentnly. 

I have tried to use a different node from fleet for subscriber and publisher but noticed that DIAL fails and hence left using same node for both.
This is a good start which makes atleast these examples working. But yes, i agree that we should make them more useful by providing some configs where user can specify their own waku node etc.

Ivansete-status (approved):
LGTM! Thanks for it! :100:

gabrielmer (approved):
Thanks so much!

chaitanyaprem:
@NagyZoltanPeter any other changes would you like to have or can i merge this?

NagyZoltanPeter (approved):
Yeah, all good! Thanks for it.

NagyZoltanPeter:
> Yeah, all good! Thanks for it.

No, you can go of course!

PR #3238, opened 2025-01-14
Add test reporting doc to benchmarks dir

stubbsta:
# Description
This PR is a first pass at adding a nwaku test summary page which aims to provide a quick reference for anyone implementing the waku protocol using nwaku to see the expected performance as well as have quick access to test reports.

# Changes

- Added file `docs/benchmarks/test-results-summary.md`

### DIFF
--- a/docs/benchmarks/test-results-summary.md
+++ b/docs/benchmarks/test-results-summary.md
@@ -1,0 +1,63 @@
+---
+title: Performance Benchmarks and Test Reports
+---
+
+
+## Introduction
+This page summarizes key performance metrics for nwaku and provides links to detailed test reports.
+
+> ## Quick reference
+>
+> - **10%** decrease in bandwidth usage in larger networks (1000 vs 2000 relay nodes)
+> - **0.4s** average, **5.5s** max message propogation latency (2000 relay nodes)
+> - Relevancy to Status App: TODO
+
+
+
+## Insights
+*Some metrics for specific protocols are currently unavailable due to reporting and logging limitations. Improvements are in progress.*
+
+### Bandwidth Usage: nwaku v0.34
+Average `libp2p` bandwidth usage for  1KB message size in a 1000 relay nodes network
+
+| Message Rate | libp2p-in (KB/s) | libp2p-out (KB/s) |
+|--------------|------------------|-------------------|
+| 1 msg/s      | ~10.1            | ~10.3             |
+| 1 msg/10s    | ~1.8             | ~1.9              |
+
+### Message Latency: nwaku v0.34.0-rc1
+Latency results for 1KB messages (1msg/s, 600 total). Click on config to see detailed report:
+
+
+| Config            | Avg Latency | Max Latency |
+|------------------------|-------------|-------------|
+| [Relay](https://www.notion.so/Waku-regression-testing-v0-34-1618f96fb65c803bb7bad6ecd6bafff9) (1000 nodes)     | 0.05s       | 1.6s        |
+| [Mixed](https://www.notion.so/Mixed-environment-analysis-1688f96fb65c809eb235c59b97d6e15b) (210 nodes)      | 0.0125s     | 0.007s      |
+| [Non-persistent Relay](https://www.notion.so/High-Churn-Relay-Store-Reliability-16c8f96fb65c8008bacaf5e86881160c)   | 0.0125s     | 0.25s       |  
+
+
+## Testing
+### DST
+The VAC DST team performs regression testing on all new **nwaku** releases, comparing performance with previous versions. They simulate large Waku networks with a variety of network and protocol configurations that are representative of real-world usage.
+
+**Test Reports**: [DST Reports](https://www.notion.so/DST-Reports-1228f96fb65c80729cd1d98a7496fe6f)  
+
+
+### QA
+The VAC QA team performs interoperability tests for **Nim Waku** and **Go Waku** using the latest main branch images. These tests run daily and verify protocol functionality by targeting specific features of each protocol.  
+
+**Test Reports**: [QA Reports](https://discord.com/channels/1110799176264056863/1196933819614363678)  
+
+### nwaku
+The **nwaku** team follows a structured release procedure for all release candidates. This involves deploying RCs to `status.staging` fleet for validation and performing sanity checks.  
+
+**Release Process**: [nwaku Release Procedure](https://github.com/waku-org/nwaku/blob/master/.github/ISSUE_TEMPLATE/prepare_release.md)  
+
+
+### Research
+The Waku Research team conducts a variety of benchmarking, performance testing, proof-of-concept validations and debugging efforts. They also maintaining a Waku simulator designed for small-scale, single-purpose, on-demand testing.
+
+
+**Test Reports**: [Waku Research Reports](https://www.notion.so/Miscellaneous-2c02516248db4a28ba8cb2797a40d1bb)
+
+**Waku Simulator**: [Waku SImulator Book](https://waku-org.github.io/waku-simulator/)


stubbsta:
This is a draft to get inputs on the formatting and content.
The Quick ref section definitely needs work. I would appreciate some info on how to relate the data back to Status App.

docs/benchmarks/test-results-summary.md
```diff
@@ -0,0 +1,63 @@
+---
+title: Performance Benchmarks and Test Reports
+---
+
+
+## Introduction
+This page summarizes key performance metrics for nwaku and provides links to detailed test reports.
```
jm-clius, line 7:
```suggestion
This page summarises key performance metrics for nwaku and provides links to detailed test reports.
```

Slightly controversial, but we've been asked to switch to British English when we publish. We certainly don't follow this consistently elsewhere, so up to you.

docs/benchmarks/test-results-summary.md
```diff
@@ -0,0 +1,63 @@
...
+
+> ## Quick reference
```
jm-clius, line 9: I'd maybe call this section something like "Executive Summary" or "TL;DR"

docs/benchmarks/test-results-summary.md
```diff
@@ -0,0 +1,63 @@
...
+>
+> - **10%** decrease in bandwidth usage in larger networks (1000 vs 2000 relay nodes)
+> - **0.4s** average, **5.5s** max message propogation latency (2000 relay nodes)
```
jm-clius, line 12:
```suggestion
> - **0.4s** average, **5.5s** max message propagation latency (2000 relay nodes)
```
I'd just be explicit here about what we we mean by "message propagation latency". In other words, "0.4s on average for a message to propagate to 100% of nodes in topologies of up to 2000 nodes".

docs/benchmarks/test-results-summary.md
```diff
@@ -0,0 +1,63 @@
...
+title: Performance Benchmarks and Test Reports
+---
+
+
+## Introduction
+This page summarizes key performance metrics for nwaku and provides links to detailed test reports.
+
+> ## Quick reference
+>
+> - **10%** decrease in bandwidth usage in larger networks (1000 vs 2000 relay nodes)
```
jm-clius, line 11: I think the actual bandwidth values are what's interesting here:
- libp2p bandwidth usage fluctuates between `min` and `max` kbps for topologies of up to 1000 nodes, with average bandwidth usage at `avg` kbps.
- this bandwidth decreases to an average of `avg_2` for larger topologies of.... + why?

docs/benchmarks/test-results-summary.md
```diff
@@ -0,0 +1,63 @@
...
+> - **0.4s** average, **5.5s** max message propogation latency (2000 relay nodes)
+> - Relevancy to Status App: TODO
+
+
+
+## Insights
+*Some metrics for specific protocols are currently unavailable due to reporting and logging limitations. Improvements are in progress.*
```
jm-clius, line 18:
```suggestion
*Some metrics for specific protocols are currently unavailable due to reporting and logging limitations.
Improvements are in progress.*
```

As an illustration of how to use semantic breaks (https://sembr.org/) which we use when publishing writeups, specs or other reviewed documents. However, I think this disclaimer can probably be removed completely.

docs/benchmarks/test-results-summary.md
```diff
@@ -0,0 +1,63 @@
...
+
+### Bandwidth Usage: nwaku v0.34
+Average `libp2p` bandwidth usage for  1KB message size in a 1000 relay nodes network
```
jm-clius, line 21: I think readers will just need a bit more guidance to understand what we mean by these terms, e.g. Whose bandwidth? What is the variable and what are the constants here?

An example suggestion:
Average `libp2p` per-node bandwidth usage for various message _injection_ rates into a Relay network of constant size `1000`. The messages are 1KB in size.

docs/benchmarks/test-results-summary.md
```diff
@@ -0,0 +1,63 @@
...
+
+| Message Rate | libp2p-in (KB/s) | libp2p-out (KB/s) |
```
jm-clius, line 23: For a public write-up, I'd be a bit more explicit in column names and avoid internal knowledge (such as that `libp2p-in` is our metric name): Message _Injection_ Rate, Average libp2p incoming bandwidth, Average libp2p outgoing bandwidth

docs/benchmarks/test-results-summary.md
```diff
@@ -0,0 +1,63 @@
...
+|--------------|------------------|-------------------|
+| 1 msg/s      | ~10.1            | ~10.3             |
+| 1 msg/10s    | ~1.8             | ~1.9              |
+
+### Message Latency: nwaku v0.34.0-rc1
+Latency results for 1KB messages (1msg/s, 600 total). Click on config to see detailed report:
```
jm-clius, line 29:
```suggestion
Latency results for 1KB messages (1msg/s, 600 total).
Click on config to see detailed report:
```

Semantic breaks (here and elsewhere)

jm-clius: Similar as before, I would just add some simple explanation of what we mean by "latency", that we're varying the message injection rates while controlling for total messages and network size, etc.

docs/benchmarks/test-results-summary.md
```diff
@@ -0,0 +1,63 @@
...
+**Release Process**: [nwaku Release Procedure](https://github.com/waku-org/nwaku/blob/master/.github/ISSUE_TEMPLATE/prepare_release.md)  
+
+
+### Research
+The Waku Research team conducts a variety of benchmarking, performance testing, proof-of-concept validations and debugging efforts. They also maintaining a Waku simulator designed for small-scale, single-purpose, on-demand testing.
+
+
+**Test Reports**: [Waku Research Reports](https://www.notion.so/Miscellaneous-2c02516248db4a28ba8cb2797a40d1bb)
+
+**Waku Simulator**: [Waku SImulator Book](https://waku-org.github.io/waku-simulator/)
```
jm-clius, line 63:
```suggestion
**Waku Simulator**: [Waku Simulator Book](https://waku-org.github.io/waku-simulator/)
```

docs/benchmarks/test-results-summary.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Testing
+### DST
+The VAC DST team performs regression testing on all new **nwaku** releases, comparing performance with previous versions. They simulate large Waku networks with a variety of network and protocol configurations that are representative of real-world usage.
+
+**Test Reports**: [DST Reports](https://www.notion.so/DST-Reports-1228f96fb65c80729cd1d98a7496fe6f)  
+
+
+### QA
+The VAC QA team performs interoperability tests for **Nim Waku** and **Go Waku** using the latest main branch images. These tests run daily and verify protocol functionality by targeting specific features of each protocol.
```
jm-clius, line 47:
```suggestion
The VAC QA team performs interoperability tests for **nwaku** and **go-waku** using the latest `master` branch builds.
These tests run daily and verify protocol functionality by targeting specific features of each protocol.
```
- semantic breaks
- even more controversially than British vs American English, there is no more "nim-waku", only "nwaku". Generally also use "go-waku" as convention.
- Perhaps we want to mention that there test suites are based on _specified_ functionality?

docs/benchmarks/test-results-summary.md
```diff
@@ -0,0 +1,63 @@
...
+### Message Latency: nwaku v0.34.0-rc1
+Latency results for 1KB messages (1msg/s, 600 total). Click on config to see detailed report:
+
+
+| Config            | Avg Latency | Max Latency |
+|------------------------|-------------|-------------|
+| [Relay](https://www.notion.so/Waku-regression-testing-v0-34-1618f96fb65c803bb7bad6ecd6bafff9) (1000 nodes)     | 0.05s       | 1.6s        |
+| [Mixed](https://www.notion.so/Mixed-environment-analysis-1688f96fb65c809eb235c59b97d6e15b) (210 nodes)      | 0.0125s     | 0.007s      |
+| [Non-persistent Relay](https://www.notion.so/High-Churn-Relay-Store-Reliability-16c8f96fb65c8008bacaf5e86881160c)   | 0.0125s     | 0.25s       |  
+
```
jm-clius, line 37: I'd add at least a section on the discv5 analyses as well. We can also consider a section on "other protocols" where we combine some insights from the various reports on the impact of non-relay protocols (Store Reliability, Mixed Environment, etc.) even if we just conclude that there is no discernible impact on _measured_ scenarios with links to some of the test results.

jm-clius (changes requested):
Thanks for this! I've added a couple of comments and suggestions on what other sections to add.

docs/benchmarks/test-results-summary.md
stubbsta: Is there not an effort to move away from using 'master' as a branch name (for politically correct reasons, I think). Even if the actual branch is named that currently, it should be clear enough what 'main' refers to in this doc, or am I overthinking things?

docs/benchmarks/test-results-summary.md
jm-clius: Mmm. I think in fact we should mention that it remains roughly _the same_ and that this is expected for Relay networks. Slight fluctuation could be due to simulation artifacts or chance differences in routing/connectivity between test runs.

docs/benchmarks/test-results-summary.md
jm-clius: There was a change on Github here to use `main` by default, but afaik this was reverted. You're right, though - "main" branch should be clear enough and cover either scenario

docs/benchmarks/test-results-summary.md
stubbsta: Changed to TL;DR in https://github.com/waku-org/nwaku/pull/3238/commits/e01a61e418616ed6628131f273fb77cb3068f3c2

docs/benchmarks/test-results-summary.md
stubbsta: REmoved in https://github.com/waku-org/nwaku/pull/3238/commits/e01a61e418616ed6628131f273fb77cb3068f3c2

docs/benchmarks/test-results-summary.md
stubbsta: Fixed in https://github.com/waku-org/nwaku/pull/3238/commits/e01a61e418616ed6628131f273fb77cb3068f3c2

docs/benchmarks/test-results-summary.md
stubbsta: Updated in https://github.com/waku-org/nwaku/pull/3238/commits/e01a61e418616ed6628131f273fb77cb3068f3c2

docs/benchmarks/test-results-summary.md
stubbsta: Updated in https://github.com/waku-org/nwaku/pull/3238/commits/e01a61e418616ed6628131f273fb77cb3068f3c2

docs/benchmarks/test-results-summary.md
stubbsta: Updated in https://github.com/waku-org/nwaku/pull/3238/commits/e01a61e418616ed6628131f273fb77cb3068f3c2

docs/benchmarks/test-results-summary.md
stubbsta: Added in https://github.com/waku-org/nwaku/pull/3238/commits/fb3b409bfc285821f2c699fe395158a1a232701f

stubbsta:
@fryorcraken I wonder if the TL;DR section is not too wordy? Is the requirement not to have it be something very short that can be read quickly and easily remembered such as:
- Relay network average bandwidth usage: x KB/s
- Disv5 average bandwidth usage: y KB/s
- etc.

and then if the reader wants more info (such as the network size and message rate for the simulations where the above values were obtained, they can look at the `Insights` section and if they want even more info they can go look at the reports on notion?

docs/benchmarks/test-results-summary.md
```diff
@@ -0,0 +1,77 @@
...
+title: Performance Benchmarks and Test Reports
+---
+
+
+## Introduction
+This page summarises key performance metrics for nwaku and provides links to detailed test reports.
+
+> ## TL;DR
+>
+> - libp2p bandwidth usage fluctuates between 5 and 15 KB/s for topologies of up to 1000 nodes, with average bandwidth usage at **10 KB/s**.
```
jm-clius, line 11: For the bandwidth numbers to make sense, we need to add the message rate and size. Perhaps just mentioning the average and max bandwidth is enough?

docs/benchmarks/test-results-summary.md
```diff
@@ -0,0 +1,77 @@
...
+> - The average bandwidth usage remains roughly the same at **9 KB/s** for a larger topology of 2000 nodes.  
+This is expected for Relay networks and the slight fluctuation could be due to simulation artifacts or chance differences in routing or connectivity between test runs.
+> - The average time for a message to propagate to 100% of nodes in topologies of up to 2000 Relay nodes is **0.4s**.
+> - The average per-node bandwidth usage of the discv5 protocol is **8 KB/s** for incoming traffic and **7.4 KB/s** for outgoing traffic.  
+ This is for a network with 100 continuously online nodes, sending 1KB messages at 1s intervals.
```
jm-clius, line 16: Perhaps the only relevant detail for discv5 here is the number of nodes and not the message size or rate. However, do we have some understanding if the discv5 bandwidth usage _does_ fluctuate much with number of nodes? If not, we can leave out the number of nodes too.

docs/benchmarks/test-results-summary.md
```diff
@@ -0,0 +1,77 @@
...
+For each simulation 600 messages of 1KB were sent at a message injection rate of 1msg/s.  
+Click on a specific config to see the detailed test report.
+
+
+| Config                                                                                                                       | Avg Message Propagation Latency (s) | Max Message Propagation Latency (s)|
+|------------------------------------------------------------------------------------------------------------------------------|-------------------------------------|------------------------------------|
+| [Relay](https://www.notion.so/Waku-regression-testing-v0-34-1618f96fb65c803bb7bad6ecd6bafff9) (1000 nodes)                   | 0.05                                | 1.6                                |
+| [Mixed](https://www.notion.so/Mixed-environment-analysis-1688f96fb65c809eb235c59b97d6e15b) (210 nodes)                       | 0.0125                              | 0.007                              |
+| [Non-persistent Relay](https://www.notion.so/High-Churn-Relay-Store-Reliability-16c8f96fb65c8008bacaf5e86881160c) (510 nodes)| 0.0125                              | 0.25                               |
```
jm-clius, line 41: I would just add very brief description of what "Relay", "Mixed" and "Non-persistent Relay" means, so that a reader doesn't have to click the links to get an intuitive understanding.

docs/benchmarks/test-results-summary.md
```diff
@@ -0,0 +1,77 @@
...
+
+|Message size         |Average discv5 incoming bandwidth (KB/s)|Average discv5 outgoing bandwidth (KB/s)|
+|-------------------- |----------------------------------------|----------------------------------------|
+| no message injection| 7.88                                   | 6.70                                   |
+| 1KB                 | 8.04                                   | 7.40                                   |
+| 10KB                | 8.03                                   | 7.45                                   |
+
+## Testing
+### DST
+The VAC DST team performs regression testing on all new **nwaku** releases, comparing performance with previous versions. They simulate large Waku networks with a variety of network and protocol configurations that are representative of real-world usage.
```
jm-clius, line 55:
```suggestion
The VAC DST team performs regression testing on all new **nwaku** releases, comparing performance with previous versions.
They simulate large Waku networks with a variety of network and protocol configurations that are representative of real-world usage.
```
Semantic breaks, here and further down. :)

docs/benchmarks/test-results-summary.md
jm-clius, line 11: Do we know the configured degree `D` here - I think this is just the default of `6`? Perhaps not worth mentioning if this is a "well-known fact" about Waku.

So, on second thought I think we can simplify this TL;DR, focus on the critical conclusion and use less domain terms. For example, our first sentence suggests that we have concluded an average of 10 KB/s only up to 1000 nodes, but in the next sentence we say roughly the same but this time for up to 2000 nodes. I'd suggest something like:

> Waku bandwidth (minus traffic related to discv5 Discovery) averages ~10KB/s for a message injection rate of X KB/s for any topology size* (*confirmed up to 2000 nodes).

I think X is 1KB/s (i.e. 1KB message every 1 second)?

jm-clius (changes requested):
A few more comments. :)

docs/benchmarks/test-results-summary.md
stubbsta: According to the results here: https://www.notion.so/Measure-DiscV5-bandwidth-with-Waku-discovery-1698f96fb65c80659fa1fbfdac49b1ef?pvs=4#16a8f96fb65c8060ac93dd35e2b9c464
there is some fluctuation when comparing the bandwidth usage for varying total nodes in the network.
The data from the referenced test includes the data from the start of the simulation, which we have not yet determined how much it impacts the bandwidth usage results. 
Because of this I think it's best to keep the total number of nodes in, until we get more information, if you agree?

jm-clius: Yes, that would make sense. I would still suggest leaving out the message size/rate stats as it describes a domain separate from discv5.

docs/benchmarks/test-results-summary.md
stubbsta: I think `D` is the default of `6` @AlbertoSoutullo could you please confirm this?
This line in the doc refers to these results https://www.notion.so/Waku-regression-testing-v0-34-1618f96fb65c803bb7bad6ecd6bafff9

stubbsta: I updated this section according to the lower comment to simplify it further.

docs/benchmarks/test-results-summary.md
stubbsta: Updated in https://github.com/waku-org/nwaku/pull/3238/commits/adcdbee4398f8040b8342c978a0e51f6a403a2b7

docs/benchmarks/test-results-summary.md
stubbsta: Added in https://github.com/waku-org/nwaku/pull/3238/commits/adcdbee4398f8040b8342c978a0e51f6a403a2b7

docs/benchmarks/test-results-summary.md
stubbsta: Added the breaks in https://github.com/waku-org/nwaku/pull/3238/commits/adcdbee4398f8040b8342c978a0e51f6a403a2b7, but I'm not sure whether they are formatted correctly now?

jm-clius (approved):
Thanks! LGTM now. :) I know there's still an outstanding section on relevance to Status, but happy to see this merged once that is done.

docs/benchmarks/test-results-summary.md
AlbertoSoutullo: Unless from your part you changed default values in the waku version, it is 6 afaik, yes c:

stubbsta:
Hi @plopezlpz, could you please have a look and maybe provide some inputs?
One of the main objectives of the doc is to indicate how these values might be relevant/helpful to Status, but my knowledge of the Status waku config is limited.
For example, it would be good to be able to compare the overall Status bandwidth usage to the average Waku bandwidth usage.
Does the Status app have its own performance benchmarks that we could use to compare?

docs/benchmarks/test-results-summary.md
```diff
@@ -0,0 +1,88 @@
...
+title: Performance Benchmarks and Test Reports
+---
+
+
+## Introduction
+This page summarises key performance metrics for nwaku and provides links to detailed test reports.
+
+> ## TL;DR
+>
+> - Average Waku bandwidth usage: ~**10 KB/s** (minus discv5 Discovery) for 1KB message size and message injection rate of 1msg/s.
```
fryorcraken, line 11: Maybe more interesting to read the bandwidth overhead.

I would not exclude discv5

So something like:

> for 1KB message per second, sent by 2000 publisher nodes (all nodes in network are publisher). The total bandwidth is 3MBps. Knowing that data injected to the network is 2MBs (1KB * 2000 per second). There is an overheard of 1MBps (gossipsub amplification + peer discovery discv5).

docs/benchmarks/test-results-summary.md
```diff
@@ -0,0 +1,88 @@
...
+This page summarises key performance metrics for nwaku and provides links to detailed test reports.
+
+> ## TL;DR
+>
+> - Average Waku bandwidth usage: ~**10 KB/s** (minus discv5 Discovery) for 1KB message size and message injection rate of 1msg/s.  
+Confirmed for topologies of up to 2000 Relay nodes.
+> - Average time for a message to propagate to 100% of nodes: **0.4s** for topologies of up to 2000 Relay nodes.
+> - Average per-node bandwidth usage of the discv5 protocol: **8 KB/s** for incoming traffic and **7.4 KB/s** for outgoing traffic,   
+ in a network with 100 continuously online nodes.
+> - Relevancy to Status App: **TODO**
```
fryorcraken, line 16: The aim is to produce a messaging API, whose default values should be the one used by Status and perform benchmarks using the messaging API so it's all relevant for Status or any app using said API.

docs/benchmarks/test-results-summary.md
```diff
@@ -0,0 +1,88 @@
...
+## Insights
+
+### Relay Bandwidth Usage: nwaku v0.34.0
+The average per-node `libp2p` bandwidth usage in a 1000-node Relay network with 1KB messages at varying injection rates.
+
+
+| Message Injection Rate | Average libp2p incoming bandwidth (KB/s) | Average libp2p outgoing bandwidth (KB/s) |
+|------------------------|------------------------------------------|------------------------------------------|
+| 1 msg/s                | ~10.1                                    | ~10.3                                    |
+| 1 msg/10s              | ~1.8                                     | ~1.9                                     |
```
fryorcraken, line 28: Would be good to state the obvious of application data injected (1msg/s  *1000 nodes) and data measured so we understand the overhead.

fryorcraken (approved):
This looks really good :+1:

docs/benchmarks/test-results-summary.md
stubbsta: @fryorcraken I don't think the regression tests currently measure total bandwidth usage per node (@AlbertoSoutullo could you please confirm this), we can only approximate using the libp2p and discv5 data we have and the payload size and msg rate data:

e.g.
_Knowing that data injected to the network is 2MBs (1KB * 2000 per second) and knowing that
libp2p in and out is 2x 8 KB/s and discv5 in and out is 10.2 + 1.6 KB/s.
We have a total of bandwidth usage of approx. 2.1 MBs_

Is this sufficient for now?

docs/benchmarks/test-results-summary.md
stubbsta: Would it make sense to add something here like:

> Future Improvements: A messaging API is currently in development to streamline interactions with the Waku protocol suite.
> Once completed, Waku will be able to provide benchmarks at the messaging API level of the software stack, enabling applications using the API to more easily compare their own benchmark results.

or rather just leave it out for now?

docs/benchmarks/test-results-summary.md
AlbertoSoutullo: We don't report the total usage but the median one. But still, the total libp2p and discv5 usage can be measured and reported. Actually, in the v0.35 report we found that the bandwidth usage of discv5 in some nodes is extremely high compared with the libp2p one.

jm-clius (approved):
@stubbsta I think we can merge with comments addressed as well as is possible now. Always possible to update this in future. :)

docs/benchmarks/test-results-summary.md
stubbsta: This section will remain as if for now. It can be updated when the total bandwidth measurement becomes available in order to show bandwidth overhead.

docs/benchmarks/test-results-summary.md
stubbsta: This section will remain as if for now. It can be updated when the total bandwidth measurement becomes available in order to show bandwidth overhead.

docs/benchmarks/test-results-summary.md
stubbsta: Added a "Future Improvements" section here in https://github.com/waku-org/nwaku/pull/3238/commits/f66369f8c548e412797b54c0b60fcc8abb92e13f

Issue #3236, opened 2025-01-10
Maintenance Y2025Q1


Ivansete-status:
### Description
Gather all maintenance tasks during first quarter of 2025.
All maintenance PRs should be linked to this so that we don't have orphan PRs.

### PRs
- https://github.com/waku-org/nwaku/pull/3233

Ivansete-status:
Just a side note that we need to keep this issue opened during Q1 xD

gabrielmer:
> Just a side note that we need to keep this issue opened during Q1 xD

Upss lol, my bad, forgot to remove the "closes" word in the PR.

Thanks for fixing it!

PR #3240, opened 2025-01-14
chore: adding extra migration to sqlite and improving error message

gabrielmer:
# Description
Waku Archive with sqlite was broken, as our table required a `storedAt` column being not null, and we are not using that column anymore, so insert statements were failing.

The tests didn't catch the issue because migrations weren't executed before the tests, so the tests were run with a db schema that wasn't the one our nodes were using in practice.

Fixed issue executing insert statements and modified tests in order to catch similar issues in the future

# Changes

- [x] added migration script for sqlite's Waku Archive that replaces the `storedAt` column with `timestamp`
- [x] start running migrations before tests, so tests are executed with correct db schema
- [x] improved error message

## Issue

#3236


migrations/message_store/00010_dropStoredAt.up.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE message DROP COLUMN timestamp;
+
+ALTER TABLE message RENAME COLUMN storedAt TO timestamp;
```
gabrielmer, line 3: Copying the following migration we do for Postgres: https://github.com/waku-org/nwaku/blob/625c8ee51bc3e065da0e2e8d3a53d3634589f548/migrations/message_store_postgres/content_script_version_6.nim#L3-L7

Ivansete-status (approved):
Awesome! Thanks for it! :100:

NagyZoltanPeter (approved):
LGTM!  Thank you!

migrations/message_store/00010_dropStoredAt.up.sql
richard-ramos, line 3: Oh this is cool. 
I thought renaming columns in sqlite wasn't possible but required creating a new table with the new column name and migrating the data from the old table to the new one! 
Is this a feature that was added recently? (for sure it would be useful even for status since there we do the 'manual' renaming procedure i mentioned).

richard-ramos (approved):

migrations/message_store/00010_dropStoredAt.up.sql
gabrielmer: Good question! Not sure if it's a recent honestly, I just copied it from the migration script we use in postgres and it worked lol

It does seem it's a feature supported by sqlite: https://www.sqlitetutorial.net/sqlite-rename-column/

richard-ramos: This was introduced in 2018! wow!

PR #3242, opened 2025-01-15
chore: Optimize libwaku size

Ivansete-status:
### Description
This PR aims for reducing the resulting `libwaku` library size

### improvements to reduce libwaku's size
- Avoid compiling libwaku with `-d:chronicles_log_level='TRACE'` 
- Avoid adding too large seq of strings in `waku/waku_rln_relay/constants.nim` as it is only needed by tests. Change in constants.nim: 
image

### Note
In Linux, the command `size build/libwaku.a` is very interesting to see which module occupies more space.

## Issue
closes https://github.com/waku-org/nwaku/issues/3241


gabrielmer:
Amazinggg! Thanks so much!

gabrielmer (approved):

PR #3249, opened 2025-01-23
chore: separating heaptrack from debug build

gabrielmer:
# Description

Debug builds currently include heaptrack. In this PR, we separate heaptrack from debug builds, and rename the Dockerfile `debug` target to `prod-with-heaptrack` which is more accurate.
# Changes

## Issue
#3236


NagyZoltanPeter (approved):
Looks reasonable! Thank you.

Ivansete-status (approved):

PR #3253, opened 2025-01-24
nix: create initial version of flake

markoburcul:
It includes also androidndk and files needed for nimbus-build-system

Referenced issue: https://github.com/waku-org/nwaku/issues/3232

# Description
Adding Nix flake to build Android app library
# Changes

## How to test

1. `nix build '?submodules=1#'`

## Issue
https://github.com/waku-org/nwaku/issues/3232
closes #


markoburcul:
I've managed to overcome building the Nim compiler using nimbus-build-system. The issue I've encountered is that `nwaku` is using Nim `2.0.12` while for `nim-codex` it is `2.0.14`. The part in the preBuild phase that had to be added is a bit different:
```nix
    pushd vendor/nimbus-build-system/vendor/Nim
    mkdir dist
    cp -r ${callPackage ./nimble.nix {}}    dist/nimble
    chmod 777 -R dist/nimble
    mkdir -p dist/nimble/dist
    cp -r ${callPackage ./checksums.nix {}} dist/checksums  # need both
    cp -r ${callPackage ./checksums.nix {}} dist/nimble/dist/checksums
    cp -r ${callPackage ./atlas.nix {}}     dist/atlas
    chmod 777 -R dist/atlas
    mkdir dist/atlas/dist
    cp -r ${callPackage ./sat.nix {}}       dist/nimble/dist/sat
    cp -r ${callPackage ./sat.nix {}}       dist/atlas/dist/sat
    cp -r ${callPackage ./csources.nix {}}  csources_v2
    chmod 777 -R dist/nimble csources_v2
    popd
```
In `nimbus-build-system` the directory `vendor/Nim` was missing rightly placed `atlas` and `sat` git repositories.

jakubgs:
Seems a bit different than what we had when `nimbus-eth2` was using Nim `2.0.12`, but if it works it works.

markoburcul:
It is possible to compile using system nim, tried it on this branch for both libwaku-android-arm64 and libwaku-android-amd64:
```bash
nix develop
make update
make libwaku-android-arm64 USE_SYSTEM_NIM=1
```

I'm still facing issue when building it:
```bash
nwaku> /build/nim-2.0.12/compiler/nimblecmd.nim(144, 12) compiler msg initiated here [MsgOrigin]
nwaku> /build/source/waku.nimble(1, 2) Error: cannot open file: ""
nwaku> /build/nim-2.0.12/compiler/modulepaths.nim(75, 17) compiler msg initiated here [MsgOrigin]
nwaku> make[1]: *** [Makefile:409: build-libwaku-for-android-arch] Error 1
nwaku> make[1]: Leaving directory '/build/source'
nwaku> make: *** [Makefile:417: libwaku-android-arm64] Error 2
```

nix/create-nimble-link.sh
```diff
@@ -0,0 +1,66 @@
+#!/usr/bin/env bash
+
+output_file="submodule_paths.txt"
+
+IFS=':' read -ra EXCLUDED_PATTERNS <<< "$EXCLUDED_NIM_PACKAGES"
```
jakubgs, line 5: But why?

nix/create-nimble-link.sh
```diff
@@ -0,0 +1,66 @@
...
+    base_dir="${submodule_path%/src}"
+    nimble_files_count=$(find "$base_dir" -maxdepth 1 -type f -name "*.nimble" | wc -l)
+
+    if [ "$nimble_files_count" -gt 0 ]; then
+        submodule_dir="$nimble_pkgs_dir/${submodule_name}-#head"
+        mkdir -p "$submodule_dir"
+
+        nimble_link_file="$submodule_dir/${submodule_name}.nimble-link"
+        echo "$submodule_path" > "$nimble_link_file"
+        echo "$submodule_path" >> "$nimble_link_file"
```
jakubgs, line 60: Why?

nix/create-nimble-link.sh
```diff
@@ -0,0 +1,66 @@
...
+    return 1
+}
+
+find . -name .gitmodules | while read -r gitmodules_file; do
+    module_dir=$(dirname "$(realpath "$gitmodules_file")")
+
+    while IFS= read -r line; do
+        if [[ $line =~ path[[:space:]]*=[[:space:]]*(.*) ]]; then
+            submodule_path="${BASH_REMATCH[1]}"
+            abs_path="$module_dir/$submodule_path"
```
jakubgs, line 23: I don't understand what's happening here. Not a single comment explaining what this regex does or why.

nix/create-nimble-link.sh
```diff
@@ -0,0 +1,66 @@
...
+
+            if should_exclude "$abs_path"; then
+                continue
+            fi
```
jakubgs, line 27: But why? What happens if we don't?

nix/create-nimble-link.sh
markoburcul: When inspecting the `.nimble-link` files after running make update there are two lines with identical content. I tried with just one line and it doesn't work.

nix/create-nimble-link.sh
markoburcul: I'll add a comment but basically with the content of .gitmodules files I'm searching for `path = /some/path` lines and extracting only them

nix/create-nimble-link.sh
markoburcul: Since this script is trying to achieve what [this script ](https://github.com/status-im/nimbus-build-system/blob/8fafcd0bac9f409091b7bcaee62ab6330f57441e/scripts/create_nimble_link.sh#L16C1-L21C7) is doing, maybe it is better if I would just link it in the comment. But again, when using `make update` we are also using the [EXCLUDED_NIM_PACKAGES](https://github.com/waku-org/nwaku/blob/4b444a1da60ad0437ee67e393e8acedfcd998dcb/Makefile#L8) environment variable to not create nimble-links for those packages/submodules. Without it it doesn't work.

nix/create-nimble-link.sh
markoburcul: The variable defined in Makefile: https://github.com/waku-org/nwaku/blob/4b444a1da60ad0437ee67e393e8acedfcd998dcb/Makefile#L8
and then used in nimbus-build-system when generating nimble-links:
https://github.com/status-im/nimbus-build-system/blob/8fafcd0bac9f409091b7bcaee62ab6330f57441e/scripts/create_nimble_link.sh#L16C1-L21C7

nix/shell.nix
```diff
@@ -0,0 +1,29 @@
...
+    darwin.apple_sdk.frameworks.Security
+  ];
+
+  LD_LIBRARY_PATH = pkgs.lib.makeLibraryPath [
+    pkgs.pcre
+  ];
+
+  shellHook = ''
+    export ANDROID_SDK_ROOT="${pkgs.androidPkgs.sdk}"
+    export ANDROID_NDK_HOME="${pkgs.androidPkgs.ndk}"
```
jakubgs, line 26: There's `androidShell`, already, you can just use `inputsFrom` with `androidShell`.

nix/shell.nix
```diff
@@ -0,0 +1,29 @@
...
+pkgs.mkShell {
+  buildInputs = with pkgs; [
+    git
+    cmake
+    openssl
+    which
+    rustup
+    docker
+    cargo
+    nim-unwrapped-2_0
```
jakubgs, line 14: Can't we just use `inputsFrom` and provide our `nim-waku` derivation?

nix/overlay.nix
```diff
@@ -0,0 +1,55 @@
...
+      (self.fetchurl { # https://github.com/golang/go/issues/58426
+        url = "https://github.com/golang/mobile/commit/406ed3a7b8e44dc32844953647b49696d8847d51.patch";
+        sha256 = "sha256-dqbYukHkQEw8npOkKykOAzMC3ot/Y4DEuh7fE+ptlr8=";
+      })
+    ];
+  })).override {
+    # FIXME: No Android SDK packages for aarch64-darwin.
+    withAndroidPkgs = stdenv.system != "aarch64-darwin";
+    androidPkgs = self.androidEnvCustom.compose;
+  };
```
jakubgs, line 49: What do we need all of this for?

flake.nix
```diff
@@ -0,0 +1,58 @@
...
+          config = {
+            android_sdk.accept_license = true;
+            allowUnfree = true;
+          };
+          overlays =  [
+            (final: prev: {
+              androidEnvCustom = prev.callPackage ./nix/pkgs/android-sdk { };
+              androidPkgs = final.androidEnvCustom.pkgs;
+              androidShell = final.androidEnvCustom.shell;
+            })
```
jakubgs, line 35: Don't we already do this in `nix/pkgs.nix`? Maybe we don't need that file then and just do the overlay only here?

Makefile
```diff
@@ -152,6 +152,14 @@ endif
 
 clean: | clean-libbacktrace
 
+### Create nimble links (used when building with Nix)
+
+nimble-dir:
+	NIMBLE_DIR="$(CURDIR)/$(NIMBLE_DIR)" \
+	PWD_CMD="$(PWD)" \
+	EXCLUDED_NIM_PACKAGES="$(EXCLUDED_NIM_PACKAGES)" \
+	BUILD_SYSTEM_DIR="$(BUILD_SYSTEM_DIR)" \
```
jakubgs, line 161: Can't we just add `export` to those at the top of the `Makefile`?

jakubgs (changes requested):
Looking good, could use some cleanup.

nix/default.nix
```diff
@@ -0,0 +1,111 @@
+{
+  config ? {},
+  pkgs ? import ./pkgs.nix { inherit config; },
```
jakubgs, line 3: Wy do we get `pkgs` from `pkgs.nix`? Shouldn't they come form the Flake?

nix/pkgs.nix
```diff
@@ -0,0 +1,34 @@
...
+let
+  # For testing local version of nixpkgs
+  #nixpkgsSrc = (import <nixpkgs> { }).lib.cleanSource "/home/jakubgs/work/nixpkgs";
+
+  # We follow release 24-05 of nixpkgs
+  # https://github.com/NixOS/nixpkgs/releases/tag/24.05
+  nixpkgsSrc = builtins.fetchTarball {
+    url = "https://github.com/NixOS/nixpkgs/archive/df27247e6f3e636c119e2610bf12d38b5e98cc79.tar.gz";
+    sha256 = "sha256:0bbvimk7xb7akrx106mmsiwf9nzxnssisqmqffla03zz51d0kz2n";
+  };
```
jakubgs, line 19: This doesn't make sense. We are already locking `nixpkgs` in `flake.nix`.

nix/shell.nix
markoburcul: I mean from the dev point of usage, they will have to put `'?submodules=1#'` and it will try to build the whole derivation on the first run which will last long. Additionally I would need to override in shellHook the fakeGit fakeCargo and fakeRustup commands because otherwise they will be used from the derivation

jakubgs: Hm, you're right, and loading of versions from `vendor` is also a problem. I guess it's fine as it is for now. At least until we extract Nim building to a separate derivation.

jakubgs (approved):
Amazing work.

.gitmodules
```diff
@@ -179,16 +179,6 @@
...
-[submodule "vendor/nim-quic"]
-	path = vendor/nim-quic
-	url = https://github.com/status-im/nim-quic.git
-	ignore = untracked
-	branch = master
-[submodule "vendor/nim-ngtcp2"]
-	path = vendor/nim-ngtcp2
-	url = https://github.com/vacp2p/nim-ngtcp2.git
-	ignore = untracked
-	branch = master
```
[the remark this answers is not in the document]
markoburcul: I removed it because in Nix build when submodules were being fetched, the build would break because of these two since they are not present on the path `vendor/nim-ngtcp2` and `vendor/nim-quic`

gabrielmer (approved):
Not very familiar with Nix, so I don't understand 95% of the code but whatever I understand LGTM 😄 

Thanks so much!

jakubgs:
Happy to explain how it works. Will be cleaner once we stop using `nimbus-build-system`.

The next step is to integrate this into `status-go` builds.

PR #3257, opened 2025-01-27
chore: move mount store before relay and rln relay

Ivansete-status:
## Description

This is needed to make a quick creation of the `messages` table, before the event-rln sync kicks off. With that, we avoid having errors from postgres-exporter (nwaku-compose) complaining about non-existing messages table.

## Issue

closes https://github.com/waku-org/nwaku/issues/3239


gabrielmer (approved):
Ooh amazing! Thanks so much!

PR #3259, opened 2025-01-27
feat: waku sync shard matching check

SionoiS:
Add a check so that peers supporting different shards don't sync.

### DIFF
--- a/tests/waku_store_sync/test_protocol.nim
+++ b/tests/waku_store_sync/test_protocol.nim
@@ -157,6 +157,40 @@ suite "Waku Sync: reconciliation":
       localWants.contains((clientPeerInfo.peerId, hash3)) == true
       localWants.contains((serverPeerInfo.peerId, hash2)) == true
 
+  asyncTest "sync 2 nodes different shards":
+    let
+      msg1 = fakeWakuMessage(ts = now(), contentTopic = DefaultContentTopic)
+      msg2 = fakeWakuMessage(ts = now() + 1, contentTopic = DefaultContentTopic)
+      msg3 = fakeWakuMessage(ts = now() + 2, contentTopic = DefaultContentTopic)
+      hash1 = computeMessageHash(DefaultPubsubTopic, msg1)
+      hash2 = computeMessageHash(DefaultPubsubTopic, msg2)
+      hash3 = computeMessageHash(DefaultPubsubTopic, msg3)
+
+    server.messageIngress(hash1, msg1)
+    server.messageIngress(hash2, msg2)
+    client.messageIngress(hash1, msg1)
+    client.messageIngress(hash3, msg3)
+
+    check:
+      remoteNeeds.contains((serverPeerInfo.peerId, hash3)) == false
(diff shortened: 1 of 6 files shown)


waku/waku_store_sync/codec.nim
```diff
@@ -209,6 +217,28 @@ proc getReconciled(idx: var int, buffer: seq[byte]): Result[bool, string] =
...
+
+  let slice = buffer[idx ..< idx + VarIntLen]
+  let (val, len) = uint64.fromBytes(slice, Leb128)
+  idx += len
+  let shardsLen = val
+
+  var shards: seq[uint16]
+  for _ in 0 ..< shardsLen:
+    if idx + VarIntLen > buffer.len:
+      return err("Cannot decode shards")
```
Ivansete-status, line 232: I'd add a different text and maybe include some more detail
```suggestion
      return err("Cannot decode shards inside for loop, idx: " & $idx)
```

Ivansete-status (approved):
LGTM! Thanks for it! :100:

tests/waku_store_sync/test_protocol.nim
```diff
@@ -157,6 +157,40 @@ suite "Waku Sync: reconciliation":
...
+    check:
+      remoteNeeds.contains((serverPeerInfo.peerId, hash3)) == false
+      remoteNeeds.contains((clientPeerInfo.peerId, hash2)) == false
+      localWants.contains((clientPeerInfo.peerId, hash3)) == false
+      localWants.contains((serverPeerInfo.peerId, hash2)) == false
+
+    server = await newTestWakuRecon(
+      serverSwitch, idsChannel, localWants, remoteNeeds, @[0.uint16, 1, 2, 3]
+    )
+    client = await newTestWakuRecon(
```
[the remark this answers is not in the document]
SionoiS: Since the protocol is symmetrical both peers (client & server) have the same information about who wants what.

Client:
local want -> 0xA
remote need -> 0xB

Server:
remote need -> 0xA
local want -> 0xB

Because of the differences in hashes and timing you don't have to create 2 for each the result is the same.

waku/waku_store_sync/common.nim
```diff
@@ -26,6 +26,8 @@ type
     ItemSet = 2
 
   RangesData* = object
+    shards*: seq[uint16]
```
jm-clius, line 29: I think we'd need the network ID too to fully qualify the shard.

jm-clius (approved):
Generally LGTM. Would be good to see this in a (raw) wire protocol spec too. I think we need network ID + shard to be fully qualified?

SionoiS:
> Generally LGTM. Would be good to see this in a (raw) wire protocol spec too. I think we need network ID + shard to be fully qualified?

I'm confused... Why add this to the spec? It's just a quick hack to prevent sync in some cases.

jm-clius:
> I'm confused... Why add this to the spec? It's just a quick hack to prevent sync in some cases.

Afaict this changes what is serialised to the wire, which IMO is good to specify for other implementers/maintainers/testers - even if we expect a future version of the protocol to replace this logic. :)

PR #3265, opened 2025-01-30
chore: Update notice of build from source pre-requisites for successful fedora builds

NagyZoltanPeter:
# Description
While original issue https://github.com/waku-org/nwaku/issues/2087 problem seems not applicable anymore. @Ivansete-status found that nimbus build system rely on `which` utility that is not included into the fedora distribution.
Installing it additionally as pre-req solves the build issue.

# Changes

- [x] Add a notice to readme.md
- [x] Updated docs.waku.org simultaneously https://github.com/waku-org/docs.waku.org/pull/230

## Issue
https://github.com/waku-org/nwaku/issues/2087


Ivansete-status (approved):
Wonderful, thanks! :partying_face:

PR #3267, opened 2025-01-30
chore: filter remove all subscription from a peer that is leaving

Ivansete-status:
## Description
@fbarbu15 reported that `waku-interop-tests` started to fail.
The problem appeared when one filter-client completely disconnects, e.g., node shutdown. In this case, the `waku_filter_2/subscriptions.nim` module was still keeping an incorrect `Connection` instance. 

Therefore, the `subscriptions` module should listen for peer left events in order to completely remove all the subscription data for that peer.

## Issue

closes: https://github.com/waku-org/nwaku/issues/3263
closes: https://github.com/waku-org/nwaku/issues/3254

### DIFF
--- a/waku/waku_filter_v2/protocol.nim
+++ b/waku/waku_filter_v2/protocol.nim
@@ -186,6 +186,7 @@ proc pushToPeer(
       if connRes.isNone():
         ## We do not remove this peer, but allow the underlying peer manager
         ## to do so if it is deemed necessary
+        error "pushToPeer no connection to peer", peerId = shortLog(peerId)
         return err("pushToPeer no connection to peer: " & shortLog(peerId))
 
       let newConn = connRes.get()
@@ -231,10 +232,6 @@ proc pushToPeers(
       pushFuts.add(pushFut)
 
     await allFutures(pushFuts)
-
-    for fut in pushFuts:
-      if fut.read().isErr():
-        error "error pushing message", error = fut.read().error
 
 proc maintainSubscriptions*(wf: WakuFilter) {.async.} =
   debug "maintaining subscriptions"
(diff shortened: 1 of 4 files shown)


waku/waku_filter_v2/protocol.nim
```diff
@@ -166,24 +167,37 @@ proc handleSubscribeRequest*(
...
+      wf.peerConnections[peerId]
+    else:
+      ## we never pushed a message before, let's dial then
+      let connRes = await wf.peerManager.dialPeer(peerId, WakuFilterPushCodec)
+      if connRes.isNone():
+        ## We do not remove this peer, but allow the underlying peer manager
+        ## to do so if it is deemed necessary
+        return err("pushToPeer no connection to peer: " & shortLog(peerId))
+
+      connRes.get()
```
NagyZoltanPeter, line 191: Should not we add here the new connection to `peerConnections` table here?

waku/waku_filter_v2/protocol.nim
```diff
@@ -211,9 +225,8 @@ proc pushToPeers(
     var pushFuts: seq[Future[void]]
 
     for peerId in peers:
-      let pushFut = wf.pushToPeer(peerId, bufferToPublish)
-      pushFuts.add(pushFut)
-    await allFutures(pushFuts)
+      (await wf.pushToPeer(peerId, bufferToPublish)).isOkOr:
```
NagyZoltanPeter, line 228: Why is it necessary here to sequentialize pushToPeer?

NagyZoltanPeter (changes requested):
I think we miss adding new connection to our table.

waku/waku_filter_v2/protocol.nim
Ivansete-status: Great catch! thanks for it! :raised_hands:

waku/waku_filter_v2/protocol.nim
NagyZoltanPeter: Hm, but we only notice any error as log. No other consequence of it. Why not add log to the single version `pushToPeer` and let them run concurrently from `pushToPeers`.

NagyZoltanPeter (approved):
I would consider not make pushToPeers sequential, but the other seems ok!

gabrielmer (approved):
Ooh amazing PR! Thanks so much! 🤩

PR #3274, opened 2025-02-03
fix: Bump nim-chronicles to latest and greates - was missing from previous version bumps

NagyZoltanPeter:
# Description
Found that even after latest v0.34 and v0.35 version bumps nim-chronicles we use still points to an 2.5y old version.
This pr bumps it to current latest and greatest. 

# Changes

- [X] version bump of nim-chronicles


Ivansete-status (approved):
Great catch! Thanks for it! :partying_face:

gabrielmer (approved):
Thank you!

PR #3275, opened 2025-02-03
feat: waku sync full topic support

SionoiS:
# Description

When syncing take into account the topics a node is interested in. Basically sync on the topic set intersection of the 2 nodes.

Omitting topics means wildcard, use all topics.

Also, trim the time range a node will sync based on interest.

### DIFF
--- a/waku/node/waku_node.nim
+++ b/waku/node/waku_node.nim
@@ -230,29 +230,24 @@ proc mountAutoSharding*(
 
 proc mountStoreSync*(
     node: WakuNode,
-    storeSyncRange = 3600.uint32,
-    storeSyncInterval = 300.uint32,
-    storeSyncRelayJitter = 20.uint32,
+    cluster: uint16,
+    shards: seq[uint16],
+    contentTopics: seq[string],
+    storeSyncRange = 3600,
+    storeSyncInterval = 300,
+    storeSyncRelayJitter = 20,
 ): Future[Result[void, string]] {.async.} =
   let idsChannel = newAsyncQueue[SyncID](0)
   let wantsChannel = newAsyncQueue[PeerId](0)
   let needsChannel = newAsyncQueue[(PeerId, WakuMessageHash)](0)
 
-  var cluster: uint16
-  var shards: seq[uint16]
-  let enrRes = node.enr.toTyped()
-  if enrRes.isOk():
-    let shardingRes = enrRes.get().relaySharding()
-    if shardingRes.isSome():
-      let relayShard = shardingRes.get()
-      cluster = relayShard.clusterID
-      shards = relayShard.shardIds
+  let pubsubTopics = shards.mapIt($RelayShard(clusterId: cluster, shardId: it))
 
   let recon =
     ?await SyncReconciliation.new(
-      cluster, shards, node.peerManager, node.wakuArchive, storeSyncRange.seconds,
-      storeSyncInterval.seconds, storeSyncRelayJitter.seconds, idsChannel, wantsChannel,
-      needsChannel,
+      cluster, pubsubTopics, contentTopics, node.peerManager, node.wakuArchive,
+      storeSyncRange.seconds, storeSyncInterval.seconds, storeSyncRelayJitter.seconds,
+      idsChannel, wantsChannel, needsChannel,
     )
 
   node.wakuStoreReconciliation = recon
--- a/waku/node/waku_node.nim
+++ b/waku/node/waku_node.nim
@@ -237,8 +237,8 @@ proc mountStoreSync*(
     storeSyncInterval = 300,
     storeSyncRelayJitter = 20,
 ): Future[Result[void, string]] {.async.} =
-  let idsChannel = newAsyncQueue[SyncID](0)
-  let wantsChannel = newAsyncQueue[PeerId](0)
+  let idsChannel = newAsyncQueue[(SyncID, PubsubTopic, ContentTopic)](0)
+  let wantsChannel = newAsyncQueue[(PeerId, WakuMessageHash)](0)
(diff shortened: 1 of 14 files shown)


SionoiS:
Future @SionoiS don't merge yet. Wait for testing. Fixes would be easier without this new code.

waku/waku_store_sync/transfer.nim
```diff
@@ -34,7 +35,7 @@ type SyncTransfer* = ref object of LPProtocol
   peerManager: PeerManager
 
   # Send IDs to reconciliation protocol for storage
-  idsTx: AsyncQueue[SyncID]
+  idsTx: AsyncQueue[(SyncID, uint16)]
```
jm-clius, line 38: Not for this PR, but these `uint16` makes me think that we need a `type Shard* = uint16` somewhere.

jm-clius (approved):
Generally direction makes sense to me. Thanks! Indeed agree that this should be merged only after the "simpler" shards support has been tested alongside basic sync functionality.

One admittedly controversial thing to consider (not for this PR): we couple the protocol very tightly to the understanding of a node being connected to a _single_ cluster with multiple shards. This view is correct, for now, but I'm wondering if this network-level understanding of what "cluster" and "shard" represent is appropriate for lower-layer protocols. In other words, if for any reason in future we want to support multiple clusters organically, the Sync protocol will be affected. Another approach would be to simply keep using the fully qualified pubsub topic (cluster+shard or even pubsub topic hash) as dimension alongside the message ID. This will decouple the sync protocol from any particular understanding of how the network sharding design is done.

SionoiS:
> ... This will decouple the sync protocol from any particular understanding of how the network sharding design is done.

Yes this a very good point! I'm almost done with the content topic support and switching to pubsub topic instead of shard is trivial. I will do that.

waku/node/waku_node.nim
```diff
@@ -208,29 +208,24 @@ proc mountSharding*(
 
 proc mountStoreSync*(
     node: WakuNode,
+    cluster: uint16,
+    shards: seq[uint16],
+    contentTopics: seq[string],
```
jm-clius, line 213: Just to make sure I understand, this means that all participating nodes would need to express a bounded content topic interest to sync on content topic level? In other words, we are not yet building something that will replace a "server-client" model where the service node supports "all" content topics (on set of shards) and only the client expresses an interest over which content topics to sync?

jm-clius (commented):
Thanks for including the fully qualified pubsub topic. Haven't been able to review in detail yet, but already have one question below to make sure I understand the use of content topic here correctly. :)

waku/waku_store_sync/reconciliation.nim
```diff
@@ -249,39 +346,41 @@ proc initFillStorage(
...
 
+    # we assume IDs are already in order
     for i in 0 ..< response.hashes.len:
       let hash = response.hashes[i]
       let msg = response.messages[i]
+      let pubsubTopic = response.topics[i]
 
-      ids.add(SyncID(time: msg.timestamp, hash: hash))
+      let id = SyncID(time: msg.timestamp, hash: hash)
+      discard storage.insert(id, pubsubTopic, msg.contentTopic)
```
gabrielmer, line 362: shouldn't we check the result here?

gabrielmer (approved):
LGTM :)) Amazing feature - thanks so much!

waku/waku_store_sync/reconciliation.nim
SionoiS: The only possible error (duplicate msg) cannot happen here.

I though it was fine to discard in this case.

waku/node/waku_node.nim
Ivansete-status: > Empty list is a wildcard accept all topics.

Shall we add comments in every place this is valid?

Ivansete-status (approved):
LGTM! Just added some nitpick comments ;P

waku/waku_store_sync/common.nim
```diff
@@ -27,7 +27,8 @@ type
 
   RangesData* = object
     cluster*: uint16
-    shards*: seq[uint16]
+    pubsubTopics*: seq[PubsubTopic]
```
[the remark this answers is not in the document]
SionoiS: Yes I agree we are mixing terms from different levels.

We could remove `cluster` since it's included in pubsub topic.

@jm-clius WDYT?

waku/node/waku_node.nim
SionoiS: > > Empty list is a wildcard accept all topics.
> 
> Shall we add comments in every place this is valid?

I'll some comments but the spec should be the place to go for that kind of info.

waku/waku_store_sync/common.nim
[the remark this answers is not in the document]
jm-clius: Sorry, missed this comment before. I'd remove `cluster`, yes.

waku/waku_store_sync/reconciliation.nim
```diff
@@ -75,23 +76,83 @@ proc messageIngress*(
...
+proc preProcessPayload(
+    self: SyncReconciliation, payload: RangesData
+): Option[RangesData] =
+  ## Check the received payload for topics and/or time mismatch.
+
+  var payload = payload
+
+  # Always use the smallest pubsub topic scope possible
+  if payload.pubsubTopics.len > 0 and self.pubsubTopics.len > 0:
+    let pubsubIntersection = self.pubsubTopics * payload.pubsubTopics.toHashSet()
```
fryorcraken, line 118: what does `*` do in this context?

SionoiS: intersection of 2 sets

PR #3276, opened 2025-02-05
fix: avoid sending relay callbacks if relay is disabled

gabrielmer:
# Description

When we attempt to start a node with Relay disabled using libwaku, the node creation fails because we attempt to set Relay app callbacks while having Relay disabled.

We should not attempt to configure Relay related callbacks if Relay is disabled.
Added a fix to send `nil` for the Relay related callbacks in case Relay is disabled.

Notice that I override the callbacks instead of configuring them to `nil` in the first place, and that's because when we set the callbacks, we still don't have the configuration parsed and it's not trivial to parse. I also attempted to set the callbacks after parsing the configuration but I got into a circular dependency error.

Therefore, I'm just adjusting which callbacks to actually send the moment we have the configuration parsed.

# Changes

- [x] setting Relay callbacks to `nil` in case Relay is disabled

## Issue
#3076

### DIFF
--- a/library/waku_thread/inter_thread_communication/requests/node_lifecycle_request.nim
+++ b/library/waku_thread/inter_thread_communication/requests/node_lifecycle_request.nim
@@ -67,6 +67,11 @@ proc createWaku(
             formattedString & ". expected type: " & $typeof(confValue)
         )
 
+  # Don't send relay app callbacks if relay is disabled
+  if not conf.relay and not appCallbacks.isNil():
+    appCallbacks.relayHandler = nil
+    appCallbacks.topicHealthChangeHandler = nil
+
   let wakuRes = Waku.new(conf, appCallbacks).valueOr:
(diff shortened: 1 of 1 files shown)


Ivansete-status (approved):
LGTM! Thanks for it! :100: 
Should we do that in "factory" side instead? That way `wakunode2` also benefits from it.

gabrielmer:
> Should we do that in "factory" side instead? That way `wakunode2` also benefits from it.

So in "factory", if a app callback for Relay is configured without having Relay mounted, we return an error.

https://github.com/waku-org/nwaku/blob/81a19c397c7abd8ff98756814f550276799c4831/waku/factory/waku.nim#L158-L160

I'm more keen on not sending wrong configurations from libwaku than accepting wrong configurations in nwaku, does it make sense?

richard-ramos (approved):

PR #3278, opened 2025-02-05
Revert "chore: waku_archive add protection against queries longer than 24h"

Ivansete-status:
Reverts waku-org/nwaku#3256

See the detail of the decision in: https://github.com/waku-org/nwaku/issues/3170#issuecomment-2637755080


gabrielmer (approved):
Thank you!

NagyZoltanPeter (approved):
It's ok to revert it, but why not just allow "infinite" time range without removing the validator?

Ivansete-status:
> It's ok to revert it, but why not just allow "infinite" time range without removing the validator?

ah yes good point! I think the answer is that "revert" was one button click ;P

PR #3281, opened 2025-02-06
chore: refactor filter to react when the remote peer closes the stream

Ivansete-status:
# Description
Better control when the remote peer closes the WakuFilterPushCodec stream remotely.
For example, go-waku closes the stream for every received message. On the other hand, js-waku keeps the stream opened. Therefore, we support both scenarios.

## Issue

closes https://github.com/waku-org/nwaku/issues/3271

### DIFF
--- a/waku/incentivization/eligibility_manager.nim
+++ b/waku/incentivization/eligibility_manager.nim
@@ -13,7 +13,8 @@ type EligibilityManager* = ref object # FIXME: make web3 private?
 proc init*(
     T: type EligibilityManager, ethClient: string
 ): Future[EligibilityManager] {.async.} =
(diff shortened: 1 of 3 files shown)


gabrielmer (approved):
LGTM, thanks so much!

NagyZoltanPeter (approved):
Awesome! Thank you.
I think it is all good however one time `getStreamByPeerIdAndProtocol` would be better placed at libp2p switch and dialer level, imho.

Ivansete-status:
> Awesome! Thank you. I think it is all good however one time `getStreamByPeerIdAndProtocol` would be better placed at libp2p switch and dialer level, imho.

Good point! Something to consider in the future if other projects may benefit from that as well: @richard-ramos

PR #3284, opened 2025-02-07
feat: mix poc

chaitanyaprem:
# Description

# Changes

- [ ] ...
- [ ] ...


chaitanyaprem:
Note that this is not to be merged to master but just wanted to get a review since the code reached a certain stage. Would help to fix any obvious issues before we go ahead with reply implementation etc.

waku/waku_mix/protocol.nim
```diff
@@ -0,0 +1,160 @@
...
+    mixNodes[peerId] = mixNodePubInfo
+  info "using mix bootstrap nodes ", bootNodes = mixNodes
+  return mixNodes
+ ]#
+
+proc new*(
+    T: type WakuMix,
+    nodeAddr: string,
+    switch: Switch,
+    peermgr: PeerManager,
```
jm-clius, line 135: `PeerManager` already has the `switch` as public member.

waku/waku_mix/protocol.nim
```diff
@@ -0,0 +1,160 @@
...
+    error "Failed to convert multiaddress to string.", err = error
+    return multiaddr
+  maddrStr.add("/p2p/" & $peerId)
+  var cleanAddr = MultiAddress.init(maddrStr).valueOr:
+    error "Failed to convert string to multiaddress.", err = error
+    return multiaddr
+  return cleanAddr
+
+proc populateMixNodePool*(mix: WakuMix) {.async.} =
+  # populate only peers that i) are reachable ii) share cluster iii) support mix
```
jm-clius, line 63: Not sure `clusterId` filtering belongs here? Why do we need mix nodes to be on the same cluster, or am I missing something? We do expect the destination node to belong to the same cluster _plus_ shard, but this is presumably left up to the discovery layer. To me it seems that mix operates at a lower layer.

waku/waku_mix/protocol.nim
```diff
@@ -0,0 +1,160 @@
...
+  info "using mix bootstrap nodes ", bootNodes = mixNodes
+  return mixNodes
+ ]#
+
+proc new*(
+    T: type WakuMix,
+    nodeAddr: string,
+    switch: Switch,
+    peermgr: PeerManager,
+    clusterId: uint16,
```
jm-clius, line 136: See comment elsewhere. I'm not quite sure we should see concepts like `cluster` (and `shard` for that matter) in `WakuMix`.

waku/waku_mix/protocol.nim
```diff
@@ -0,0 +1,160 @@
...
+
+proc populateMixNodePool*(mix: WakuMix) {.async.} =
+  # populate only peers that i) are reachable ii) share cluster iii) support mix
+  let remotePeers = mix.peerManager.switch.peerStore.getReachablePeers().filterIt(
+      mixPoolFilter(some(mix.clusterId), it)
+    )
+  var mixNodes = initTable[PeerId, MixPubInfo]()
+
+  for i in 0 ..< min(remotePeers.len, 100):
+    let remotePeerENR = remotePeers[i].enr.get()
```
jm-clius, line 70: Mmm. I understand why we need it for now, but it would indeed be great to in future not have to know about `ENR` within the Waku Mix protocol

waku/waku_lightpush/client.nim
```diff
@@ -110,3 +114,34 @@ proc publishToAny*(
     obs.onMessagePublished(pubSubTopic, message)
 
   return lightpushSuccessResult(publishedCount)
+
+proc publishWithConn*(
```
jm-clius, line 118: Nitpick and not for this PR: we can probably clean up Lightpush API a little bit by reducing the number of `publishX` methods and using `Option` more judiciously (e.g. `publish(LightpushRequest, Option[Connection])`)

jm-clius (commented):
Great work! I've left some comments/questions below.

waku/waku_mix/protocol.nim
chaitanyaprem: well, the problem is exit node has to open a connection to destination node which may get dropped if the node is in different cluster due to metadata negotiation failing. That being said, if shards mismatch as well connection would get closed. 
Need to think how to overcome that.

SionoiS (approved):
Not much comments except good job!

waku/factory/external_config.nim
```diff
@@ -667,6 +667,23 @@ with the drawback of consuming some more bandwidth.""",
       name: "rendezvous"
     .}: bool
 
+    #Mix config
+    mix* {.
+      desc: "Enable mix protocol: true|false",
+      defaultValue: false,
+      name: "mix"
+    .}: bool
+    mixkey* {.
```
Ivansete-status, line 676: Nitpick one: add extra line space between config attributes. Also, nph needs to be run against that file.
```suggestion
    .}: bool

    mixkey* {.
```

waku/factory/internal_config.nim
```diff
@@ -34,6 +36,9 @@ proc enrConfiguration*(
   ).isOkOr:
     return err("could not initialize ENR with shards")
 
+  if conf.mix and mixPubKey.isSome():
+    enrBuilder.withMixKey(mixPubKey.get())
```
Ivansete-status, line 40: Would it be possible to avoid adding an additional key pair? The node already has its "p2p" key pair and I wonder if we could reuse it to avoid adding more key management layers. That's an important point.

waku/factory/waku.nim
```diff
@@ -279,7 +280,7 @@ proc updateEnr(waku: ptr Waku): Result[void, string] =
   let netConf: NetConfig = getRunningNetConfig(waku).valueOr:
     return err("error calling updateNetConfig: " & $error)
 
-  let record = enrConfiguration(waku[].conf, netConf, waku[].key).valueOr:
+  let record = enrConfiguration(waku[].conf, netConf, waku[].key, none(Curve25519Key)).valueOr:
```
Ivansete-status, line 283: Notice that with this, the possible previous `mix` information present in the ENR, will be set to empty.

Ivansete-status (commented):
Thanks so much for the PR! Overall it looks great even though I don't know mix :raised_hands: 
I'm adding a bunch of nitpick comments plus some other suggestions that I hope you find useful.
Cheers

waku/factory/internal_config.nim
chaitanyaprem: Mix uses sphinx packet encryption which is based on ed25519 and hence different keypair has to be used. We can persist this keypair so that it doesn't get regenerated every run. 

Maybe in the long run we can make single key but at this point there are many other things we need to address before we think of this :)

examples/lightpush_publisher_mix_config.nim
```diff
@@ -0,0 +1,28 @@
...
+    name: "msg-interval"
+  .}: int
+
+  minMixPoolSize* {.
+    desc: "Number of messages to wait for before sending.",
+    defaultValue: 3,
+    name: "min-mix-pool-size"
+  .}: int
+
+  withoutMix* {.
```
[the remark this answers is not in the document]
chaitanyaprem: yea, agree it is named funnily. 
Changing it to mixDisabled, because this publisher is mainly supposed to be used with mix.

waku/factory/external_config.nim
chaitanyaprem: hmm, i thought nph is running when we save the file automatically. 
is there a specific way to run it?

waku/factory/waku.nim
chaitanyaprem: ah, this needs to be fixed then. 
Don't know how to make it pick existing key and pass it again though. Any ideas as to how it can be done. 
As of now the mixKey is generated if it is not passed in the conf at https://github.com/waku-org/nwaku/blob/3612200ea26f1df26c08e069730ba143338be16c/waku/factory/node_factory.nim#L500-L518

waku/factory/external_config.nim
chaitanyaprem: nvm, figured it out with help from @gabrielmer .

waku/node/waku_node.nim
```diff
@@ -203,6 +218,33 @@ proc mountSharding*(
...
+  if protoRes.isErr:
+    error "Waku Mix protocol initialization failed", err = protoRes.error
+    return
+  node.mix = protoRes.value
+
+  let catchRes = catch:
+    node.switch.mount(node.mix)
+  if catchRes.isErr():
+    return err(catchRes.error.msg)
+  node.mix.start()
```
gabrielmer, line 245: We generally start the protocols where we start the node instead of when mounting the protocols

https://github.com/waku-org/nwaku/blob/237f7abbb00c64b0ee66d36795dcf3c9e755abe8/waku/node/waku_node.nim#L1430-L1470

gabrielmer (approved):
Amazing work! 😍 😍 

Thanks so much!

waku/factory/node_factory.nim
```diff
@@ -148,7 +150,10 @@ proc getAutoshards*(
   return ok(autoshards)
 
 proc setupProtocols(
-    node: WakuNode, conf: WakuNodeConf, nodeKey: crypto.PrivateKey
+    node: WakuNode,
+    conf: WakuNodeConf,
+    nodeKey: crypto.PrivateKey,
+    mixPrivKey: Curve25519Key,
```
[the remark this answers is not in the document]
chaitanyaprem: Nope, because as you can see in lines 501 to 510, key maynot be passed in the conf and can get generated if not passed. Initially i just tried to set it in the conf object itself if it is not present, but that doesn't seem to work.

waku/factory/waku.nim
chaitanyaprem: hmm, but i don't see Waku object present in `node_factory` -> `setupNode` proc. So not sure how to access it there and store it once generated.

gabrielmer: > hmm, but i don't see Waku object present in `node_factory` -> `setupNode` proc. So not sure how to access it there and store it once generated.

ooh true true my bad, I see that the key is actually being generated in Waku's `new` proc:
https://github.com/waku-org/nwaku/blob/ad921585574edf0aa4c1288c0615090bc0bbb391/waku/factory/waku.nim#L172-L203

so probably from there we should pass it to `setupNode` or something like that. I might be wrong though, just thinking on the pattern followed by the other key but it might be flawed

waku/waku_mix/protocol.nim
```diff
@@ -0,0 +1,159 @@
...
+  var maddrStr = multiaddr.toString().valueOr:
+    error "Failed to convert multiaddress to string.", err = error
+    return multiaddr
+  maddrStr.add("/p2p/" & $peerId)
+  var cleanAddr = MultiAddress.init(maddrStr).valueOr:
+    error "Failed to convert string to multiaddress.", err = error
+    return multiaddr
+  return cleanAddr
+
+proc populateMixNodePool*(mix: WakuMix) {.async.} =
```
Ivansete-status, line 62: Not needed unless I'm missing something ;P
```suggestion
proc populateMixNodePool*(mix: WakuMix) =
```

waku/waku_mix/protocol.nim
```diff
@@ -0,0 +1,159 @@
...
+  # populate only peers that i) are reachable ii) share cluster iii) support mix
+  let remotePeers = mix.peerManager.switch.peerStore.getReachablePeers().filterIt(
+      mixPoolFilter(some(mix.clusterId), it)
+    )
+  var mixNodes = initTable[PeerId, MixPubInfo]()
+
+  for i in 0 ..< min(remotePeers.len, 100):
+    let remotePeerENR = remotePeers[i].enr.get()
+    # TODO: use the most exposed/external multiaddr of the peer, right now using the first
```
Ivansete-status, line 71: What do you mean by most "exposed/external"? Kindly elaborate a bit more the comment

waku/waku_mix/protocol.nim
```diff
@@ -0,0 +1,159 @@
...
+  mix.setNodePool(mixNodes)
+  trace "mix node pool updated", poolSize = mix.getNodePoolSize()
+  return
+
+proc startMixNodePoolMgr*(mix: WakuMix) {.async.} =
+  info "starting mix node pool manager"
+  # try more aggressively to populate the pool at startup
+  var attempts = 50
+  # TODO: make initial pool size configurable
+  while mix.getNodePoolSize() < 100 and attempts > 0:
```
Ivansete-status, line 93: Expecting the node to have 100 peers at first seems a bit optimistic ;P
I believe this will rarely happen and the node will almost always wait 50 seconds.

I think a better approach would be to make the node keep listening for new incoming nodes supporting mix, and once they reach the desired min num mix-peers, then we can consider that mix protocol is in healthy state.

thoughts @gabrielmer @NagyZoltanPeter @jm-clius?

Ivansete-status (commented):
Some more comments that I hope you find useful
Thanks! :raised_hands:

waku/waku_mix/protocol.nim
chaitanyaprem: well, that is just a number along with number of attempts which would at max delay by 50 seconds. 

is there any impact of this on other initialization steps?
this process happens asynchronously and anyways the protocol would not be ready until this is finished. 
I agree that ideally we want the protocol state to be set to healthy only once we reach this pool size, but as a poc isn't this enough?

waku/waku_mix/protocol.nim
chaitanyaprem: Any waku node may have multiple listening addresses based on how its network configuration is. 
i.e a node can have local LAN based ip address, external address and some other address if it supports circuit-relay. 
But for mix we want to use an address which is external in nature by default and preferably not a circuit-relay one.
Hence this statement.
I remember in go-waku we had to use similar logic somewhere and thought it maybe needed here as well. 
But if we are always sure that node's reported addresses in ENR at always externally available then we may not need to add any filter here.

waku/waku_mix/protocol.nim
chaitanyaprem: Is async not needed because we are already invoking this in a heartbeat function which is in itself async?

waku/factory/internal_config.nim
chaitanyaprem: I am not sure at this point what is the impact of changing encryption scheme on mix and hence do not want to focus on it.

the code you are referring to uses `secp256k1` only for libp2p peerID signing and not for sphinx packet encryption.

It may have an impact on sphinx encryption itself which might need further study and hence it makes sense to take this up at a later stage. 
Maybe @AkshayaMani can shed more light on this.

chaitanyaprem: Also, note that the idea is not to merge this PR into master as this is a POC still using separate libp2p branch and mix protocol branches. 
This PR is only to get it reviewed so that feature branch can be created with a reviewed version where further changes to implementation can be done.

waku/waku_mix/protocol.nim
```diff
@@ -0,0 +1,159 @@
...
+
+  # TODO : ideally mix should not be marked ready until certain min pool of mixNodes are discovered
+  var m = WakuMix(peerManager: peermgr, clusterId: clusterId)
+  m.init(localMixNodeInfo, peermgr.switch, initTable[PeerId, MixPubInfo]())
+  procCall MixProtocol(m).init()
+
+  return ok(m)
+
+proc start*(mix: Wakumix) =
+  discard mix.startMixNodePoolMgr()
```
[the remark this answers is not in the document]
chaitanyaprem: The NodePoolManager is a separate entity that needs to constantly keep running throughout the lifetime of the node. Which is why the proc has been made async and invoked without await. 

If we use await, then `start` proc of WakuMix would never return which is not ideal.

Or am i missing something here?

waku/waku_mix/protocol.nim
chaitanyaprem: ah, understood now. 
Modified callers to use await and then removed async.

Ivansete-status:
Adding a short discord chat as a reference. From there, we concluded that the only blocking issue for that PR is making the nim-libp2p's mix changes available in nim-libp2p's master branch ( cc @chaitanyaprem @richard-ramos )

image

waku/factory/node_factory.nim
```diff
@@ -6,7 +6,8 @@ import
   libp2p/protocols/pubsub/gossipsub,
   libp2p/protocols/connectivity/relay/relay,
   libp2p/nameresolving/dnsresolver,
-  libp2p/crypto/crypto
+  libp2p/crypto/crypto,
+  libp2p/crypto/curve25519
```
fryorcraken: why are those imports needed?

waku/factory/waku.nim
```diff
@@ -9,6 +9,7 @@ import
   libp2p/protocols/connectivity/relay/client,
   libp2p/wire,
   libp2p/crypto/crypto,
+  libp2p/crypto/curve25519,
```
fryorcraken: why is this import needed?

waku/waku_mix/protocol.nim
fryorcraken: I came here to say that indeed, external address may changed as the libp2p identify protocol happens, so not sure it's something you can pass when you instantiate the mix protocol.

waku/factory/node_factory.nim
chaitanyaprem: They are remnants of pre-refactor of the mix code..removing them

waku/factory/waku.nim
chaitanyaprem: They are remnants of pre-refactor of the mix code..removing them

waku/factory/waku.nim
chaitanyaprem: With the new conf structure this is no longer an issue, hence resolving it.

waku/waku_mix/protocol.nim
chaitanyaprem: agree, note that a peer passed to mix doesn't mean that current node will ever establish direct connection with it. It would just be used as a peer in the mix path as this peer was discovered in the network supporting mix. So not sure if we can rely on identify always.
 
Also this is not an initialization for the protocol, rather process to keep the mix node pool up-to-date. As of now a very rudimentery approach was followed for the POC. This will have to be enhanced in future.

I am guessing this would get handled as and when ENR updates are received for a peer. Need to think more on this as i see this being a challenge in dynamic environment.The mix protocol implementation might also have to be adapted to lookup the peer's details in case connection fails rather than giving up on first failure(because the multiaddr is old or not the most external in nature).

waku/waku_mix/protocol.nim
chaitanyaprem: > if shards mismatch as well connection would get closed.

this is no longer an issue since metadata has been updated to not close connections due to mismatched shards.

apps/chat2mix/chat2mix.nim
```diff
@@ -0,0 +1,704 @@
+## chat2 is an example of usage of Waku v2. For suggested usage options, please
```
Ivansete-status, line 1: This file duplicates a lot of code from `cha2.nim`. Would it be possible to extend the current `chat2` with the `mix` feature? (the comment applies to the whole `chat2mix` folder.)

examples/lightpush_mix/lightpush_publisher_mix.nim
```diff
@@ -0,0 +1,195 @@
...
+  if not conf.mixDisabled:
+    let (mixPrivKey, mixPubKey) = generateKeyPair().valueOr:
+      error "failed to generate mix key pair", error = error
+      return
+    (await node.mountMix(clusterId, mixPrivKey)).isOkOr:
+      error "failed to mount waku mix protocol: ", error = $error
+      return
+
+  let dPeerId = PeerId.init(destPeerId).valueOr:
+    error "Failed to initialize PeerId", err = error
```
Ivansete-status:
```suggestion
    error "Failed to initialize PeerId", error = error
```

examples/lightpush_mix/lightpush_publisher_mix.nim
```diff
@@ -0,0 +1,195 @@
...
+    return
+  var conn: Connection
+  if not conf.mixDisabled:
+    conn = node.wakuMix.toConnection(
+      MixDestination.init(dPeerId, pxPeerInfo.addrs[0]),
+      WakuLightPushCodec,
+      Opt.some(MixParameters(expectReply: Opt.some(true), numSurbs: Opt.some(byte(1)))),
```
Ivansete-status: May add a tiny comment to each parameter to understand them better?

examples/lightpush_mix/lightpush_publisher_mix.nim
```diff
@@ -0,0 +1,195 @@
...
+    notice "publisher service started with mix node pool size ",
+      currentpoolSize = node.getMixNodePoolSize()
+
+  var i = 0
+  while i < conf.numMsgs:
+    if conf.mixDisabled:
+      let connOpt = await node.peerManager.dialPeer(dPeerId, WakuLightPushCodec)
+      if connOpt.isNone():
+        error "failed to dial peer with WakuLightPushCodec", target_peer_id = dPeerId
+        return
```
Ivansete-status, line 152: It doesn't seem correct to keep opening streams for every single message to be sent.
I think is better to move this snippet next to:
```nim
  if not conf.mixDisabled:
    conn = node.wakuMix.toConnection(
      MixDestination.init(dPeerId, pxPeerInfo.addrs[0]),
      WakuLightPushCodec,
      Opt.some(MixParameters(expectReply: Opt.some(true), numSurbs: Opt.some(byte(1)))),
    ).valueOr:
      error "failed to create mix connection", error = error
      return
      
```

examples/lightpush_mix/lightpush_publisher_mix_config.nim
```diff
@@ -0,0 +1,28 @@
...
+    int
+
+  msgIntervalMilliseconds* {.
+    desc: "Interval between messages in milliseconds.",
+    defaultValue: 1000,
+    name: "msg-interval"
+  .}: int
+
+  minMixPoolSize* {.
+    desc: "Number of messages to wait for before sending.",
```
Ivansete-status: Maybe? I can't quite get the purpose of this param. 
A higher MixPoolSize means more privacy and less speed?
```suggestion
    desc: "Number of mix nodes to wait for before sending.",
```

waku/factory/conf_builder/mix_conf_builder.nim
```diff
@@ -0,0 +1,38 @@
...
+
+logScope:
+  topics = "waku conf builder mix"
+
+##################################
+## Mix Config Builder ##
+##################################
+type MixConfBuilder* = object
+  enabled*: Option[bool]
+  mixKey*: Option[string]
```
Ivansete-status: Shall we make them private so that it is easier later on to track where they are being used?

waku/factory/conf_builder/waku_conf_builder.nim
```diff
@@ -271,6 +274,9 @@ proc withMaxMessageSize*(builder: var WakuConfBuilder, maxMessageSize: string) =
 proc withStaticNodes*(builder: var WakuConfBuilder, staticNodes: seq[string]) =
   builder.staticNodes = concat(builder.staticNodes, staticNodes)
 
+proc withMix*(builder: var WakuConfBuilder, mix: bool) =
+  builder.mix = some(mix)
```
Ivansete-status: Nitpick comment. Shall we move this snippet above, next to other `bool` protocols are set?

waku/factory/conf_builder/mix_conf_builder.nim
```diff
@@ -0,0 +1,38 @@
...
+    if b.mixKey.isSome():
+      let mixPrivKey = intoCurve25519Key(ncrutils.fromHex(b.mixKey.get()))
+      let mixPubKey = public(mixPrivKey)
+      return ok(some(MixConf(mixKey: mixPrivKey, mixPubKey: mixPubKey)))
+    else:
+      # Generate a new key pair if not provided
+      let keyPairResult = generateKeyPair()
+      if keyPairResult.isErr:
+        return err("Generate key pair error: " & $keyPairResult.error)
+      let (mixPrivKey, mixPubKey) = keyPairResult.get()
```
Ivansete-status: Maybe?
```suggestion
      let (mixPrivKey, mixPubKey) = generateKeyPair().valueOr:
        return err("Generate key pair error: " & $error)
```

waku/factory/external_config.nim
```diff
@@ -621,6 +621,13 @@ with the drawback of consuming some more bandwidth.""",
       name: "rendezvous"
     .}: bool
 
+    #Mix config
+    mix* {.desc: "Enable mix protocol: true|false", defaultValue: false, name: "mix".}:
+      bool
+
+    mixkey* {.desc: "ED25519 private key as 64 char hex string.", name: "mixkey".}:
+      Option[string]
```
Ivansete-status: Perhaps?
```suggestion
    mixkey* {.desc: "ED25519 private key as 64 char hex string, without 0x. If not provided, a random key will be generated.", name: "mixkey".}:
      Option[string]
```

waku/factory/internal_config.nim
```diff
@@ -2,11 +2,19 @@ import
...
-  std/[options, sequtils, net],
+  std/[options, sequtils, net, strutils],
   results
-import ../common/utils/nat, ../node/net_config, ../waku_enr, ../waku_core, ./waku_conf
+
+import
+  ../common/utils/nat,
+  ../node/net_config,
+  ../waku_enr,
+  ../waku_enr/mix,
```
Ivansete-status: Instead of doing that, I think is better to extend the following barrel module and import/export mix there: `waku/waku_enr.nim`

waku/node/waku_node.nim
```diff
@@ -123,6 +134,8 @@ type
     started*: bool # Indicates that node has started listening
     topicSubscriptionQueue*: AsyncEventQueue[SubscriptionEvent]
     rateLimitSettings*: ProtocolRateLimitSettings
+    wakuMix*: WakuMix
+    mixbootNodes*: Table[PeerId, MixPubInfo]
```
Ivansete-status: This seems not to be needed. In any case, if needed, that should belong to `wakuMix`
```suggestion
```

waku/node/waku_node.nim
```diff
@@ -226,6 +239,32 @@ proc mountAutoSharding*(
...
 
+proc getMixNodePoolSize*(node: WakuNode): int =
+  return node.wakuMix.getNodePoolSize()
+
+proc mountMix*(
+    node: WakuNode, clusterId: uint16, mixPrivKey: Curve25519Key
+): Future[Result[void, string]] {.async.} =
+  info "mounting mix protocol", nodeId = node.info #TODO log the config used
+
+  let localaddrStr = node.announcedAddresses[0].toString().valueOr:
```
Ivansete-status, line 250: Just in case :)

```suggestion
  if node.announcedAddresses.len == 0:
    return err("Trying to mount mix without having announced addresses")
    
  let localaddrStr = node.announcedAddresses[0].toString().valueOr:
```

waku/node/waku_node.nim
```diff
@@ -226,6 +239,32 @@ proc mountAutoSharding*(
...
+    return err("Failed to convert multiaddress to string.")
+  info "local addr", localaddr = localaddrStr
+
+  let nodeAddr = localaddrStr & "/p2p/" & $node.peerId
+  # TODO: Pass bootnodes from config,
+  let protoRes = WakuMix.new(nodeAddr, node.peerManager, clusterId, mixPrivKey)
+  if protoRes.isErr:
+    error "Waku Mix protocol initialization failed", err = protoRes.error
+    return
+  node.wakuMix = protoRes.value
```
```suggestion
  node.wakuMix = WakuMix.new(nodeAddr, node.peerManager, clusterId, mixPrivKey).valueOr:
    return err("Waku Mix protocol initialization failed: " & $error)
```

waku/node/waku_node.nim
```diff
@@ -1182,17 +1221,45 @@ proc lightpushPublishHandler(
...
-    return await node.wakuLightpushClient.publish(some(pubsubTopic), message, peer)
+      msg_hash = msgHash,
+      mixify = mixify
+    if mixify:
+      #TODO: How to handle multiple addresses?
+      let conn = node.wakuMix.toConnection(
+        MixDestination.init(peer.peerId, peer.addrs[0]),
+        WakuLightPushCodec,
+        Opt.some(
+          MixParameters(expectReply: Opt.some(true), numSurbs: Opt.some(byte(1)))
```
Ivansete-status, line 1239: I think there should be somewhere a comment explaining `numSurbs`' purpose

waku/node/waku_node.nim
```diff
@@ -1182,17 +1221,45 @@ proc lightpushPublishHandler(
...
+        )
+
+      return await node.wakuLightpushClient.publishWithConn(
+        pubsubTopic, message, conn, peer.peerId
+      )
+    else:
+      return await node.wakuLightpushClient.publish(some(pubsubTopic), message, peer)
 
   if not node.wakuLightPush.isNil():
+    if mixify:
```
Ivansete-status, line 1256: What does `mixify` mean? Maybe we can have a tiny comment

waku/waku_enr/mix.nim
```diff
@@ -0,0 +1,26 @@
...
+const MixKeyEnrField* = "mix-key"
+
+func withMixKey*(builder: var EnrBuilder, mixPubKey: Curve25519Key) =
+  builder.addFieldPair(MixKeyEnrField, getBytes(mixPubKey))
+
+func mixKey*(record: TypedRecord): Option[seq[byte]] =
+  let field = record.tryGet(MixKeyEnrField, seq[byte])
+  if field.isNone():
+    return none(seq[byte])
+  return field
```
Ivansete-status: It seems this can be removed as `lightpushwithmix`, `chat2mix` and `wakunode2` compile well without it
```suggestion
```

waku/waku_lightpush/client.nim
```diff
@@ -30,14 +30,18 @@ proc addPublishObserver*(wl: WakuLightPushClient, obs: PublishObserver) =
...
+    peer: PeerId | RemotePeerInfo,
+    conn: Option[Connection] = none(Connection),
 ): Future[WakuLightPushResult] {.async.} =
-  let connection = (await wl.peerManager.dialPeer(peer, WakuLightPushCodec)).valueOr:
-    waku_lightpush_v3_errors.inc(labelValues = [dialFailure])
-    return lighpushErrorResult(
-      LightPushErrorCode.NO_PEERS_TO_RELAY,
-      dialFailure & ": " & $peer & " is not accessible",
-    )
+  let connection = conn.valueOr:
```
Ivansete-status, line 38: I have the impression that we shouldn't open a new stream/connection for every message to be pushed. In any case, if we open the stream/connection for every message, I believe we need to close it afterwards. @richard-ramos can give much better insight about what is the best approach

waku/waku_mix.nim
```diff
@@ -0,0 +1,3 @@
+import ./waku_mix/[protocol]
```
Ivansete-status: Super nitpick sorry
```suggestion
import ./waku_mix/protocol
```

waku/waku_mix/protocol.nim
Ivansete-status: Yes, I think we can iterate further in the future if needed. Fine for now

waku/waku_mix/protocol.nim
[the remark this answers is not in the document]
Ivansete-status: ah yes, you are right.
For that particular case, see the following examples on how to perform that in a more idiomatic way. The benefit of such techniques is that we will eventually be able to perform graceful shutdown.

E.g1:
https://github.com/waku-org/nwaku/blob/dbecd721f9f86e8d05f1506126033fcfa206c084/waku/waku_relay/protocol.nim#L419
https://github.com/waku-org/nwaku/blob/dbecd721f9f86e8d05f1506126033fcfa206c084/waku/waku_relay/protocol.nim#L425

-----
E.g2:

https://github.com/waku-org/nwaku/blob/dbecd721f9f86e8d05f1506126033fcfa206c084/waku/waku_archive/archive.nim#L278
https://github.com/waku-org/nwaku/blob/dbecd721f9f86e8d05f1506126033fcfa206c084/waku/waku_archive/archive.nim#L286

-----
E.g3:

https://github.com/waku-org/nwaku/blob/dbecd721f9f86e8d05f1506126033fcfa206c084/waku/waku_store_sync/reconciliation.nim#L395
https://github.com/waku-org/nwaku/blob/dbecd721f9f86e8d05f1506126033fcfa206c084/waku/waku_store_sync/reconciliation.nim#L406

Ivansete-status (approved):
LGTM! Thanks for it! :100: :partying_face: 
Just adding some nitpick comments that I hope you find useful

waku/waku_lightpush/client.nim
chaitanyaprem: We have to do that with mix, because each message chooses a random mix path and also note that this is only a logical connection and not an actual transport. But i do get your point and that would require more things to be managed at mix protocol layer which will be taken care in further updates. correct me if i am wrong @richard-ramos

apps/chat2mix/chat2mix.nim
chaitanyaprem: agreed, i think now is the time to try and do that. Will see if i can do that. otherwise will take it up in subsequent PR as this is already very old and tired of rebases :)

examples/lightpush_mix/lightpush_publisher_mix.nim
chaitanyaprem: explained reasoning below and note that this is not actual transport stream but a mix wrapper stream which emulates a transport.

The idea in mix is for each message to choose a random path of nodes to route it and hence anyways underlying transport will get opened to send it.

examples/lightpush_mix/lightpush_publisher_mix_config.nim
chaitanyaprem: > A higher MixPoolSize means more privacy and less speed?

Not quite, a higher mix pool size gaurantess better privacy but has nothing to do with speed. Randomly 3 nodes would be chosen from the pool as mixPath for each message.

The purpose of this param was for test app to decide the pool-size to wait before it starts publishing messages. This is used in simulations like [this](https://forum.vac.dev/t/waku-mix-integration-simulations/469) one where based on simulated set say 20 node mixnet, we want the tool to wait till all 20 nodes are discovered and only then initiate sending messages. It makes calculating things like latency and validating other factors simple.

waku/node/waku_node.nim
chaitanyaprem: just following [terminology](https://rfc.vac.dev/vac/raw/mix/#2-terminology) as per spec

PR #3287, opened 2025-02-10
Update references to RLN contract address

stubbsta:
# Description
This PR is part of https://github.com/waku-org/nwaku/issues/3190, which requires that the RLN contract address is updated to the new contract address.
This PR needs to be merged with all other PRs related to https://github.com/waku-org/nwaku/issues/3190 as well as https://github.com/waku-org/nwaku/issues/3272

# Changes
Contract address changed from `0xCB33Aa5B38d79E3D9Fa8B10afF38AA201399a7e3` to `[0xB9cd878C90E49F797B4431fBF4fb333108CB90e6](https://sepolia.lineascan.build/address/0xB9cd878C90E49F797B4431fBF4fb333108CB90e6)` in the following files:
- https://github.com/waku-org/nwaku/blob/master/apps/sonda/docker-compose.yml#L12
- https://github.com/waku-org/nwaku/blob/master/apps/sonda/register_rln.sh#L27
- https://github.com/waku-org/nwaku/blob/master/waku/factory/networks_config.nim#L25

## How to test
This cannot be tested at this time.
The RLN tests in nwaku first need to be updated as well as the contract registration tools.

## Issue closes
Part of closing https://github.com/waku-org/nwaku/issues/3190


stubbsta:
I think we should wait, I plan to make another PR for the updated registration tool and any other related functionality.
This PR can then be merged at the same time as that one.

jm-clius (commented):
Indeed. LGTM, but since this change will break functionality until we have updated the code to use the new contract, this shouldn't be merged for now. Perhaps you can convert this to a Draft PR, which indicates work for which feedback can be given but that shouldn't be merged?

PR #3293, opened 2025-02-12
feat: incentivization PoC: client-side reputation system basics

s-tikhomirov:
# Description

This PR implements parts of a simple client-side reputation system that don't require modifications in the Lightpush protocol.

This is a countinuaton of https://github.com/waku-org/nwaku/pull/3166 and https://github.com/waku-org/nwaku/pull/3264 with associated deliverable https://github.com/waku-org/pm/issues/245.

The goal of this PR is to merge parts of the reputation system that don't require integration into any real protocol (Lightpush or otherwise). This functionality aims to encompass common functions that all reputation-enhanced protocols will use (Lightpush will be the first but probably not the only protocol to do so).

In a reputation-enhanced protocol, clients remember the results of their interactions with servers and update the servers' reputation scores accordingly. If a server sends a bad request (for some definition of response quality implemented in the respective response evaluation functionality), that server's reputation is set to "bad".

Note that reputation indicators have no relation to gossipsub peer scores. It is possible that reputation indicators from client-server protocols will play a role in peer scoring as an application-specific component (cf. `p_5` in the [peer score formula](https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.1.md#the-score-function)). This idea is left for future work.

# Changes 

List of detailed changes

-  [x] implement a reputation manager that maintains a mapping from peer IDs to reputation indicators (good, bad, unknown)
-  [x] add functions to set, get and set reputation indicators
-  [x] add a function to evaluate a (Lightpush) response (a response is "good" if its `isSuccess` field is `true`)
-  [x] update reputation indicators based on response quality.

The next steps will imply integrating the eligibility- and reputation-related functionality in the Lightpush protocol (behind a feature flag) that involve amending the relevant protobuf structures and peer selection mechanism. In particular, a client will avoid selecting bad servers for all its future requests. Also, response quality evaluation will involve making a Store request to ensure a given message has indeed been published.


gabrielmer (approved):
LGTM, thanks so much!

waku/incentivization/reputation_manager.nim
```diff
@@ -0,0 +1,49 @@
+import tables
+import waku/waku_lightpush/rpc
+
+type
+  PeerId = string
+
+  ReputationIndicator* = enum
+    BadRep
+    GoodRep
+    UnknownRep
```
jm-clius, line 10: Perhaps an `Option[bool]` would be more flexible than an enum, especially if we plan to eventually replace this with an analog reputation or integrate it as a weight in a peer score elsewhere? `Option[bool]` can be "none" (unknown), true (good?) or false (bad?).

jm-clius (approved):
LGTM, minor suggestion below. Feel free to merge without rerequesting review. :)

waku/incentivization/reputation_manager.nim
s-tikhomirov: Fair enough. My reasoning for using a custom enum was mostly readability, but considering that it should be easier to integrate elsewhere with a simple type, I changed it to `Option[bool]` ([ecf94f6](https://github.com/waku-org/nwaku/pull/3293/commits/ecf94f629d5aff42cffb1f37a6249b685373085a)).

PR #3296, opened 2025-02-12
chore: supporting parallel libwaku requests

gabrielmer:
# Description
Adding a lock so multiple requests sent by different threads can be handled in parallel.

The lock applies only from the moment a task enters the queue to the moment it starts getting handled. Many requests can be handled in parallel, however, the requests are communicated to the Waku Thread one at a time.

# Changes

- [x] adding a lock from the moment a request is sent to the Waku Thread, to the moment it's acknowledged by it and started getting handled

## Note:
This is the same PR as https://github.com/waku-org/nwaku/pull/3273 but I wasn't able to reopen it, so created this one instead.

This is a temporary solution until https://github.com/waku-org/nwaku/issues/3295 is completed

## Issue

 - #3076

### DIFF
--- a/library/waku_thread/waku_thread.nim
+++ b/library/waku_thread/waku_thread.nim
@@ -123,7 +123,6 @@ proc sendRequestToWakuThread*(
 
   ## wait until the Waku Thread properly received the request
   let res = ctx.reqReceivedSignal.waitSync()
-
   if res.isErr():
     deallocShared(req)
     return err("Couldn't receive reqReceivedSignal signal")
--- a/library/waku_thread/waku_thread.nim
+++ b/library/waku_thread/waku_thread.nim
@@ -2,12 +2,13 @@
 {.pragma: callback, cdecl, raises: [], gcsafe.}
 {.passc: "-fPIC".}
 
-import std/[options, atomics, os, net]
+import std/[options, atomics, os, net, locks]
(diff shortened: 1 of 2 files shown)


library/waku_thread/waku_thread.nim
```diff
@@ -95,23 +98,34 @@ proc sendRequestToWakuThread*(
     userData: pointer,
 ): Result[void, string] =
   let req = WakuThreadRequest.createShared(reqType, reqContent, callback, userData)
+
+  # This lock is only necessary while we use a SP Channel and while the signalling
+  # between threads assumes that there aren't concurrent requests.
+  # Rearchitecting the signaling + migrating to a MP Channel will allow us to receive
+  # requests concurrently and spare us the need of locks
+  ctx.lock.acquire()
```
Ivansete-status, line 106: Maybe more secure to release in a `defer` block ;P
```suggestion
  ctx.lock.acquire()
  defer:
    ctx.lock.release()
```

Ivansete-status (approved):
LGTM! Thanks for it! :100: 
I just added a nitpick comment

library/waku_thread/waku_thread.nim
gabrielmer: Ooh great idea! Thank you!

richard-ramos (approved):

PR #3297, opened 2025-02-12
chore: lightpush enhance log when handling request

Ivansete-status:
# Description
Enhance lightpush log and show `my_peer_id` and the `receivedTime`
That will help @AlbertoSoutullo


gabrielmer (approved):
Thanks so much!

Issue #3069, opened 2024-02-03
[Epic: nwaku] Node Bandwidth Management Features


chair28980:
## Tasks

- [x] https://github.com/waku-org/nwaku/issues/1945
- [x] https://github.com/waku-org/nwaku/issues/1946
- [x] https://github.com/waku-org/nwaku/issues/2032
- [x] https://github.com/waku-org/nwaku/issues/3060
- [ ] https://github.com/waku-org/nwaku/issues/1947 [Out of scope - postponed]

Issue #3060, opened 2024-09-26
feat: Measure impact between protocols


Ivansete-status:
### Problem
nwaku supports multiple streams/protocols. Nevertheless, we need to measure how they can impact each other.

### Suggested solution
Considering the current tools we have to measure the normal bandwidth and rate request, through Grafana/metrics, we should be able to see the possible impact to a any protocol when increasing the bandwidth/rate of one protocol

### Acceptance criteria
- [ ] See how a protocol can impact other protocols. The protocols to analyze are: store, lightpush, and filter.
- [ ] Measure the current limit of each protocol separately. When it starts to have poorer performance
- [ ] Measure the limit when two or more protocols are being compromised. See if they differ from the previous point.
- [ ] Based on previous measures, apply the limitation per protocol, or globally.

### Epic
- https://github.com/waku-org/pm/issues/117

NagyZoltanPeter:
Test environment for measurements:

- waku-simulator 
   - with 15 releay node
   - 1 edge node
   - 1 service node
     - limited to use a single and specific CPU core
     - limited to 500MB Ram
   - 1 Postgres DB
     - use one and the same CPU core as service node (this simulates VPS environment better)
- lite-protocol-tester
  - different configuration is used for stressing lightpush and filter protocols.
- sonda tool
  - 1 sonda tool to add stress condition on store protocol.

Aim was to visualize timely impact on protocols performance upon different stress condition of other protocols.
We measure per request - processing - response - times of protocols.

NagyZoltanPeter:
.

NagyZoltanPeter:
Repository for performance measurements: https://github.com/waku-org/waku-protocol-perf-test

NagyZoltanPeter:
## 3rd test run - altered with more resources

Exact same test run as 3rd, but service-node's resources are higher
- 2 cpu cores
- 2GB of RAM assigned

### Result

![Image](https://github.com/user-attachments/assets/a26c5ef1-6c48-4830-a9e1-ac12bcbbb367)

### Conclusion

We can clearly see the exact same pattern as with original low resource test previously.
More memory with doubled CPU power just not helped lowering response time in this kind of high load conditions.
This comes with the fact nwaku us built single threaded, all the processing of any requests runs on the same single thread and processing speed is not limited by memory but computing power.

Ivansete-status:
Thanks so much for the anaylysis @NagyZoltanPeter !
Why we see the decrease tendency in the request rate?

[...]

PR #3299, opened 2025-02-14
chore: non-relay protocols cross performance measurement metrics

NagyZoltanPeter:
# Description
For the final stage of #3069 under #3060 we would like to test and visualize cross-effect of the different non-relay protocol performances. This can help fine tune rate limiting of those protocols and also to make recommendations for node operator configuration.

# Changes

- [X] introduce request processing time metrics for non-relay protocols

Following point finally tracked in a separate repository: https://github.com/waku-org/waku-protocol-perf-test
- Create new node performance monitoring dashboard for non-relay protocols
    - Add histograms of processing times per protocol next to each other
    - Add panels of request rate graphs
    - Possible add bandwidth / message size rates.
- Define some stress test sequence that uses non-relay protocols parallel. (possibly using sonda and lite-protocol-tester tools)    

## Issue
#3060

### DIFF
--- a/waku/common/rate_limit/request_limiter.nim
+++ b/waku/common/rate_limit/request_limiter.nim
@@ -22,6 +22,8 @@ import
   libp2p/stream/connection,
   libp2p/utility
 
+import std/times except TimeInterval, Duration, seconds, minutes
+
 import ./[single_token_limiter, service_metrics, timed_map]
 
 export token_bucket, setting, service_metrics
--- a/waku/common/rate_limit/request_limiter.nim
+++ b/waku/common/rate_limit/request_limiter.nim
@@ -76,12 +76,12 @@ template checkUsageLimit*(
     bodyWithinLimit, bodyRejected: untyped,
 ) =
   if t.checkUsage(proto, conn):
-    let requestStartTime = Moment.now()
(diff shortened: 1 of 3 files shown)


waku/common/rate_limit/request_limiter.nim
```diff
@@ -76,8 +76,15 @@ template checkUsageLimit*(
     bodyWithinLimit, bodyRejected: untyped,
 ) =
   if t.checkUsage(proto, conn):
+    let requestStartTime = Moment.now()
     waku_service_requests.inc(labelValues = [proto, "served"])
+
     bodyWithinLimit
+
+    let requestDurationSec = (Moment.now() - requestStartTime).milliseconds.float / 1000
```
Ivansete-status, line 84: Surely not needed but just in case :)?
```suggestion
    let requestDurationSec: float = (Moment.now() - requestStartTime).milliseconds.float / 1000
```

----

Aside from this comment, I think it would be useful to rename the `rejected` case to `rejected-rate-limit`, or something like that

Ivansete-status (approved):
LGTM! Thanks for it! :100:

gabrielmer (approved):
Thank you!

PR #3307, opened 2025-02-25
feat: added store sync dashboard panel

SionoiS:
# Description

I added the metrics collected by store sync code into a new panel.

Screenshot from 2025-03-19 11-08-41

### DIFF
--- a/metrics/waku-fleet-dashboard.json
+++ b/metrics/waku-fleet-dashboard.json
@@ -27,14 +27,9 @@
   "graphTooltip": 1,
   "id": 36,
   "links": [],
-  "liveNow": false,
   "panels": [
     {
       "collapsed": true,
-      "datasource": {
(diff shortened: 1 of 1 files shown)


SionoiS:
> Thank you! Could you please add a photo of how the panel looks like?

Done

gabrielmer (approved):
Thank you!

jm-clius (approved):
LGTM, thanks. Based on screenshot, I'm not sure I understand why there does seem to be round trips for the >2 buckets? Or is this simply a function of how the histogram is rendered?

SionoiS:
> LGTM, thanks. Based on screenshot, I'm not sure I understand why there does seem to be round trips for the >2 buckets? Or is this simply a function of how the histogram is rendered?

Yeah that's how it looks.

PR #3308, opened 2025-02-26
fix: subscribing to RelaydefaultHandler in libwaku

gabrielmer:
# Description

In libwaku nodes, messages aren't being saved in Store for dynamically-subscribed shards. This happens because when subscribing dynamically via libwaku, we weren't setting the `defaultRelayHandler` we rely on for Store and other protocols.

We need to subscribe to the default handler when dynamically subscribing to topics too.

# Changes

- [x] calling `registerRelayDefaultHandler` when calling `waku_relay_subscribe` in libwaku 

## Issue

#3076


NagyZoltanPeter (approved):
What a catch! Great work!

Ivansete-status (approved):

PR #3310, opened 2025-02-27
chore: add simple Qt example that uses libwaku

Ivansete-status:
## Description
Simple Qt example that runs libwaku.
It only contains two buttons: start node; stop node.

image

note: This is part of investigation where we see that _Status_ app crashes when loading libwaku.

## Issue


examples/qt/main_qt.cpp
```diff
@@ -0,0 +1,43 @@
...
+
+int main(int argc, char *argv[]) {
+    QGuiApplication app(argc, argv);
+    QQmlApplicationEngine engine;
+
+    WakuHandler wakuHandler;
+    void* userData = nullptr;
+
+    char jsonConfig[5000];
+    snprintf(jsonConfig, 5000, "{ \
```
NagyZoltanPeter, line 19: nitpick, but you don't need sprintf if this is only static string...or bet you can directly use QString here.

examples/qt/waku_handler.h
```diff
@@ -0,0 +1,44 @@
...
+    Q_OBJECT
+private:
+    static void event_handler(int callerRet, const char* msg, size_t len, void* userData) {
+        printf("Receiving message %s\n", msg);
+    }
+
+public:
+    WakuHandler() : QObject(), ctx(nullptr) {}
+
+    void initialize(const QString& jsonConfig, void* event_handler, void* userData) {
```
NagyZoltanPeter, line 17: I think void* event_handler can directly be WakuCallBack type...

NagyZoltanPeter (approved):
Nothing blocking. LGTM!

gabrielmer (approved):
Amazinggg, thanks so much!

examples/qt/main_qt.cpp
Ivansete-status: Thanks! Addressed in https://github.com/waku-org/nwaku/pull/3310/commits/49b352d5e609c47f1378301c24c3cc132902a08c

examples/qt/waku_handler.h
Ivansete-status: Thanks! Addressed in https://github.com/waku-org/nwaku/pull/3310/commits/49b352d5e609c47f1378301c24c3cc132902a08c

PR #3311, opened 2025-02-28
fix: using nimMainPrefix in libwaku

gabrielmer:
# Description

Compiling libwaku with `--nimMainPrefix` to avoid conflicts with other Nim libraries or applications

## Issue

#3076

### DIFF
--- a/library/libwaku.nim
+++ b/library/libwaku.nim
@@ -107,7 +107,7 @@ proc onTopicHealthChange(ctx: ptr WakuContext): TopicHealthChangeHandler =
 
 # Every Nim library must have this function called - the name is derived from
 # the `--nimMainPrefix` command line option
-proc NimMain() {.importc.}
+proc libwakuNimMain() {.importc.}
 
 # To control when the library has been initialized
 var initialized: Atomic[bool]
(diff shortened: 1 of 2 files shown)


library/libwaku.nim
```diff
@@ -122,7 +122,7 @@ if defined(android):
 
 proc initializeLibrary() {.exported.} =
   if not initialized.exchange(true):
-    NimMain() # Every Nim library needs to call `NimMain` once exactly
+    libwakuNimMain() # Every Nim library needs to call `NimMain` once exactly
```
NagyZoltanPeter, line 125:
```suggestion
    libwakuNimMain() # Every Nim library needs to call `libwakuNimMain` once exactly
```

NagyZoltanPeter (approved):
Yet another stuff I learned, thanks for it! :-)

library/libwaku.nim
gabrielmer: mmm good point about that comment! Not sure how to phrase it 😶 
`libwakuNimMain` only exists in libwaku, so it doesn't apply for other libraries. I think that comment is more general, mentioning that as a library we import that because every library has to call `NimMain` or its (equivalent)

But now that we added a prefix that general comment is a bit weird, and if we add the prefix to the general comment it becomes inaccurate

Opinions on this, anyone?

Ivansete-status: Maybe something like the following?
```
## Every Nim library needs to call `<yourprefix>NimMain` once exactly, to initialize the Nim runtime.
## Being `<yourprefix>` the value given in the optional compilation flag --nimMainPrefix:yourprefix
````

Ivansete-status (approved):
Wonderful! Thanks so much for it! :partying_face:

library/libwaku.nim
gabrielmer: ooh I like it :))

Thank you!

Issue #2924, opened 2024-07-23
chore(rln-relay): deprecate tree sync strategy


rymnc:
### Background
With the changes to the rln-v2 contracts, it is now possible to fetch the merkleProofElements from the contract and use that to generate proofs instead. 

### Details
Add a proc to the contract definition in https://github.com/waku-org/nwaku/blob/8d107b0dedf6129032e817db46999fa512e6bb2e/waku/waku_rln_relay/group_manager/on_chain/group_manager.nim#L31-L50 like so -

```nim
  # this function returns the merkleProof for a given index
  proc merkleProofElements(index: Uint32): seq[Uint256] {.view.}
```

and fetch it every time an event is emitted by the contract. Use this to generate proofs instead.

### Acceptance criteria
A list of tasks that need to be done for the issue to be considered resolved.

- [ ] Cache the root of the tree every time an event is emitted by the contract
- [ ] Use the new api in zerokit to generate proofs by passing in the serialized witness
- [ ] Cache the merkleProof every time an event is emitted by the contract
- [ ] Override the GroupManager.[generateProof](https://github.com/waku-org/nwaku/blob/8d107b0dedf6129032e817db46999fa512e6bb2e/waku/waku_rln_relay/group_manager/group_manager_base.nim#L176) in the OnchainGroupManager by using the custom witness we will pass in

rymnc:
This issue is blocked at the moment due to our nim-web3 dependency being out of date. The version we use does not support returning a sequence of elements from the smart contract.

After bumping to the latest version of nim-web3, transaction signing seems to not work as expected, and hence we're blocked

jm-clius:
@seemenkina I know with this issue being blocked your focus has been elsewhere, but do you know if there's an update on the nim-web3 limitation?

seemenkina:
@jm-clius as I see yesterday they started using a newer version of nim in the dependencies - I suspect it will now be possible to continue with this task

jm-clius:
@seemenkina, thanks for the feedback. My understanding from nwaku is that the dependencies in nwaku will be updated within the next few days. :) cc @NagyZoltanPeter

Ivansete-status:
Hey there! We are considering @darshankabariya to take care of this issue once nim-web3 is resolved.
Do you agree with that @jm-clius , @seemenkina ? Or do you prefer carry on with it yourself?

seemenkina:
hey, I'm also agree, if you will need any help with zerokit integration feel free to ask!

PR #3312, opened 2025-03-02
feat: deprecate sync / local merkle tree 

darshankabariya:
This PR deprecates the sync strategy by removing the need for each node to maintain a local Merkle tree. Previously, every node independently managed a Merkle tree and used it to generate Merkle proofs for RLN proof generation. However, constructing a local Merkle tree required syncing a large amount of blockchain data, which was highly time-consuming.

With the latest RLN contract update, we can now fetch Merkle proofs directly from the smart contract, eliminating the need for local Merkle tree maintenance.

<p align="center">
  Screenshot 2025-03-21 at 12 19 33 PM
</p>

Changes

- [x] Deprecated the syncing and Merkle tree maintenance code from onchainGroupManager.
- [x] Introduced a new group manager that fetches Merkle proofs directly from the smart contract and uses them for RLN proof generation.
- [x] disable effected unittest according 

closes
- #2924

### DIFF
--- a/tests/node/test_wakunode_relay_rln.nim
+++ b/tests/node/test_wakunode_relay_rln.nim
@@ -455,7 +455,7 @@ suite "Waku RlnRelay - End to End - OnChain":
       except CatchableError:
         assert true
 
-    xasyncTest "Unregistered contract":
+    asyncTest "Unregistered contract":
       # This is a very slow test due to the retries RLN does. Might take upwards of 1m-2m to finish.
       let
         invalidContractAddress = "0x0000000000000000000000000000000000000000"
--- a/waku/waku_rln_relay/conversion_utils.nim
+++ b/waku/waku_rln_relay/conversion_utils.nim
@@ -157,20 +157,23 @@ func u256*(n: Quantity): UInt256 {.inline.} =
   n.uint64.stuint(256)
 
 proc uint64ToField*(n: uint64): array[32, byte] =
-  ## Converts uint64 to 32-byte little-endian array with zero padding
-  var bytes = toBytes(n, Endianness.littleEndian)
-  result[0 ..< bytes.len] = bytes
+  var output: array[32, byte]
+  let bytes = toBytes(n, Endianness.littleEndian)
+  output[0 ..< bytes.len] = bytes
+  return output
 
 proc UInt256ToField*(v: UInt256): array[32, byte] =
-  return cast[array[32, byte]](v)
+  return cast[array[32, byte]](v) # already doesn't use `result`
 
 proc seqToField*(s: seq[byte]): array[32, byte] =
-  result = default(array[32, byte])
+  var output: array[32, byte]
   let len = min(s.len, 32)
   for i in 0 ..< len:
-    result[i] = s[i]
+    output[i] = s[i]
+  return output
 
 proc uint64ToIndex*(index: MembershipIndex, depth: int): seq[byte] =
-  result = newSeq[byte](depth)
+  var output = newSeq[byte](depth)
   for i in 0 ..< depth:
-    result[i] = byte((index shr i) and 1) # LSB-first bit decomposition
+    output[i] = byte((index shr i) and 1) # LSB-first bit decomposition
+  return output
--- a/waku/waku_rln_relay/conversion_utils.nim
+++ b/waku/waku_rln_relay/conversion_utils.nim
@@ -116,19 +116,16 @@ proc serialize*(memIndices: seq[MembershipIndex]): seq[byte] =
 
   return memIndicesBytes
 
-proc serialize*(witness: Witness): seq[byte] =
+proc serialize*(witness: RLNWitnessInput): seq[byte] =
   ## Serializes the witness into a byte array according to the RLN protocol format
   var buffer: seq[byte]
-  # Convert Fr types to bytes and add them to buffer
   buffer.add(@(witness.identity_secret))
   buffer.add(@(witness.user_message_limit))
   buffer.add(@(witness.message_id))
-  # Add path elements length as uint64 in little-endian
   buffer.add(toBytes(uint64(witness.path_elements.len), Endianness.littleEndian))
-  # Add each path element
   for element in witness.path_elements:
     buffer.add(@element)
-  # Add remaining fields
+  buffer.add(toBytes(uint64(witness.path_elements.len), Endianness.littleEndian))
   buffer.add(witness.identity_path_index)
   buffer.add(@(witness.x))
   buffer.add(@(witness.external_nullifier))
--- a/waku/waku_rln_relay/conversion_utils.nim
+++ b/waku/waku_rln_relay/conversion_utils.nim
@@ -27,9 +27,6 @@ proc inHex*(
     valueHex = "0" & valueHex
   return toLowerAscii(valueHex)
 
-proc toUserMessageLimit*(v: UInt256): UserMessageLimit =
-  return cast[UserMessageLimit](v)
-
 proc encodeLengthPrefix*(input: openArray[byte]): seq[byte] =
   ## returns length prefixed version of the input
   ## with the following format [len<8>|input<var>]
@@ -148,3 +145,22 @@ func `+`*(a, b: Quantity): Quantity {.borrow.}
 
 func u256*(n: Quantity): UInt256 {.inline.} =
   n.uint64.stuint(256)
+
+proc uint64ToField*(n: uint64): array[32, byte] =
+  ## Converts uint64 to 32-byte little-endian array with zero padding
+  var bytes = toBytes(n, Endianness.littleEndian)
+  result[0 ..< bytes.len] = bytes
+
+proc UInt256ToField*(v: UInt256): array[32, byte] =
+  return cast[array[32, byte]](v)
+
+proc seqToField*(s: seq[byte]): array[32, byte] =
+  result = default(array[32, byte])
+  let len = min(s.len, 32)
+  for i in 0 ..< len:
+    result[i] = s[i]
+
+proc uint64ToIndex*(index: MembershipIndex, depth: int): seq[byte] =
+  result = newSeq[byte](depth)
+  for i in 0 ..< depth:
+    result[i] = byte((index shr i) and 1) # LSB-first bit decomposition
--- a/waku/waku_rln_relay/conversion_utils.nim
+++ b/waku/waku_rln_relay/conversion_utils.nim
@@ -76,7 +76,17 @@ proc serialize*(
   return output
 
 proc serialize*(witness: RLNWitnessInput): seq[byte] =
-  ## Serializes the witness into a byte array according to the RLN protocol format
+  ## Serializes the RLN witness into a byte array following zerokit's expected format.
+  ## The serialized format includes:
+  ## - identity_secret (32 bytes, little-endian with zero padding)
+  ## - user_message_limit (32 bytes, little-endian with zero padding)
+  ## - message_id (32 bytes, little-endian with zero padding)
+  ## - merkle tree depth (8 bytes, little-endian) = path_elements.len / 32
+  ## - path_elements (each 32 bytes, ordered bottom-to-top)
+  ## - merkle tree depth again (8 bytes, little-endian)
+  ## - identity_path_index (sequence of bits as bytes, 0 = left, 1 = right)
+  ## - x (32 bytes, little-endian with zero padding)
+  ## - external_nullifier (32 bytes, little-endian with zero padding)
   var buffer: seq[byte]
   buffer.add(@(witness.identity_secret))
   buffer.add(@(witness.user_message_limit))
--- a/waku/waku_rln_relay/conversion_utils.nim
+++ b/waku/waku_rln_relay/conversion_utils.nim
@@ -78,6 +78,21 @@ proc serialize*(
   )
   return output
 
+proc serialize*(witness: RLNWitnessInput): seq[byte] =
+  ## Serializes the witness into a byte array according to the RLN protocol format
+  var buffer: seq[byte]
+  buffer.add(@(witness.identity_secret))
+  buffer.add(@(witness.user_message_limit))
+  buffer.add(@(witness.message_id))
+  buffer.add(toBytes(uint64(witness.path_elements.len / 32), Endianness.littleEndian))
+  for element in witness.path_elements:
+    buffer.add(element)
+  buffer.add(toBytes(uint64(witness.path_elements.len / 32), Endianness.littleEndian))
+  buffer.add(witness.identity_path_index)
+  buffer.add(@(witness.x))
+  buffer.add(@(witness.external_nullifier))
+  return buffer
+
 proc serialize*(proof: RateLimitProof, data: openArray[byte]): seq[byte] =
   ## a private proc to convert RateLimitProof and data to a byte seq
   ## this conversion is used in the proof verification proc
@@ -115,21 +130,6 @@ proc serialize*(memIndices: seq[MembershipIndex]): seq[byte] =
     memIndicesBytes = concat(memIndicesBytes, @memIndexBytes)
 
   return memIndicesBytes
-
-proc serialize*(witness: RLNWitnessInput): seq[byte] =
-  ## Serializes the witness into a byte array according to the RLN protocol format
-  var buffer: seq[byte]
-  buffer.add(@(witness.identity_secret))
-  buffer.add(@(witness.user_message_limit))
-  buffer.add(@(witness.message_id))
-  buffer.add(toBytes(uint64(witness.path_elements.len), Endianness.littleEndian))
-  for element in witness.path_elements:
-    buffer.add(@element)
-  buffer.add(toBytes(uint64(witness.path_elements.len), Endianness.littleEndian))
-  buffer.add(witness.identity_path_index)
-  buffer.add(@(witness.x))
-  buffer.add(@(witness.external_nullifier))
-  return buffer
 
 proc toEpoch*(t: uint64): Epoch =
   ## converts `t` to `Epoch` in little-endian order
--- a/waku/waku_rln_relay/group_manager/group_manager_base.nim
+++ b/waku/waku_rln_relay/group_manager/group_manager_base.nim
@@ -122,35 +122,29 @@ method onWithdraw*(g: GroupManager, cb: OnWithdrawCallback) {.base, gcsafe.} =
 proc slideRootQueue*(
     rootQueue: var Deque[MerkleNode], root: MerkleNode
 ): seq[MerkleNode] =
+  ## updates the root queue with the latest root and pops the oldest one when the capacity of `AcceptableRootWindowSize` is reached
   let overflowCount = rootQueue.len - AcceptableRootWindowSize + 1
   var overflowedRoots = newSeq[MerkleNode]()
   if overflowCount > 0:
+    # Delete the oldest `overflowCount` roots in the deque (index 0..`overflowCount`)
+    # insert into overflowedRoots seq and return
     for i in 0 ..< overflowCount:
       overFlowedRoots.add(rootQueue.popFirst())
+  # Push the next root into the queue
   rootQueue.addLast(root)
   return overFlowedRoots
 
 method indexOfRoot*(
     g: GroupManager, root: MerkleNode
 ): int {.base, gcsafe, raises: [].} =
+  ## returns the index of the root in the merkle tree.
+  ## returns -1 if the root is not found
   return g.validRoots.find(root)
 
 method validateRoot*(
     g: GroupManager, root: MerkleNode
 ): bool {.base, gcsafe, raises: [].} =
   ## validates the root against the valid roots queue
-  # Print all validRoots in one line with square brackets
-  var rootsStr = "["
-  var first = true
-  for r in g.validRoots.items():
-    if not first:
-      rootsStr.add(", ")
-    rootsStr.add($r)
-    first = false
-  rootsStr.add("]")
-  debug "Valid Merkle roots in validateRoot", roots = rootsStr, root_to_validate = root
-
-  # Check if the root is in the valid roots queue
   if g.indexOfRoot(root) >= 0:
     return true
   return false
--- a/waku/waku_rln_relay/group_manager/group_manager_base.nim
+++ b/waku/waku_rln_relay/group_manager/group_manager_base.nim
@@ -122,23 +122,17 @@ method onWithdraw*(g: GroupManager, cb: OnWithdrawCallback) {.base, gcsafe.} =
 proc slideRootQueue*(
     rootQueue: var Deque[MerkleNode], root: MerkleNode
 ): seq[MerkleNode] =
-  ## updates the root queue with the latest root and pops the oldest one when the capacity of `AcceptableRootWindowSize` is reached
   let overflowCount = rootQueue.len - AcceptableRootWindowSize + 1
   var overflowedRoots = newSeq[MerkleNode]()
   if overflowCount > 0:
-    # Delete the oldest `overflowCount` roots in the deque (index 0..`overflowCount`)
-    # insert into overflowedRoots seq and return
     for i in 0 ..< overflowCount:
       overFlowedRoots.add(rootQueue.popFirst())
-  # Push the next root into the queue
   rootQueue.addLast(root)
   return overFlowedRoots
 
 method indexOfRoot*(
     g: GroupManager, root: MerkleNode
 ): int {.base, gcsafe, raises: [].} =
-  ## returns the index of the root in the merkle tree.
-  ## returns -1 if the root is not found
   return g.validRoots.find(root)
 
 method validateRoot*(
--- a/waku/waku_rln_relay/group_manager/group_manager_base.nim
+++ b/waku/waku_rln_relay/group_manager/group_manager_base.nim
@@ -4,7 +4,7 @@ import
   ../protocol_metrics,
   ../constants,
   ../rln
-import options, chronos, results, std/[deques, sequtils], chronicles
+import options, chronos, results, std/[deques, sequtils]
 
 export options, chronos, results, protocol_types, protocol_metrics, deques
 
--- a/waku/waku_rln_relay/group_manager/group_manager_base.nim
+++ b/waku/waku_rln_relay/group_manager/group_manager_base.nim
@@ -189,8 +189,6 @@ method generateProof*(
   if g.userMessageLimit.isNone():
     return err("user message limit is not set")
 
-  debug "calling proofGen from generateProof from group_manager_base", data = data
-
   waku_rln_proof_generation_duration_seconds.nanosecondTime:
     let proof = proofGen(
       rlnInstance = g.rlnInstance,
--- a/waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
+++ b/waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
@@ -10,10 +10,9 @@ import
   nimcrypto/keccak as keccak,
   stint,
   json,
-  std/tables,
+  std/[strutils, tables],
   stew/[byteutils, arrayops],
-  sequtils,
-  strutils
+  sequtils
 import
   ../../../waku_keystore,
   ../../rln,
@@ -174,11 +173,6 @@ proc fetchMerkleProofElements*(
 
     let responseBytes = await g.ethRpc.get().provider.eth_call(tx, "latest")
 
-    # debug "---- raw response ----",
-    #   total_bytes = responseBytes.len, # Should be 640
-    #   non_zero_bytes = responseBytes.countIt(it != 0),
-    #   response = responseBytes
-
     var i = 0
     var merkleProof = newSeq[array[32, byte]]()
     while (i * 32) + 31 < responseBytes.len:
@@ -188,12 +182,6 @@ proc fetchMerkleProofElements*(
       element = responseBytes.toOpenArray(startIndex, endIndex)
       merkleProof.add(element)
       i += 1
-      # debug "---- element ----",
-      #   startIndex = startIndex,
-      #   startElement = responseBytes[startIndex],
-      #   endIndex = endIndex,
-      #   endElement = responseBytes[endIndex],
-      #   element = element
 
     debug "merkleProof", responseBytes = responseBytes, merkleProof = merkleProof
 
@@ -272,10 +260,10 @@ proc trackRootChanges*(g: OnchainGroupManager) {.async.} =
       error "Failed to fetch Merkle proof", error = proofResult.error
     g.merkleProofCache = proofResult.get()
 
-    debug "--- track update ---",
-      len = g.validRoots.len,
-      validRoots = g.validRoots,
-      merkleProof = g.merkleProofCache
+    # debug "--- track update ---",
+    #   len = g.validRoots.len,
+    #   validRoots = g.validRoots,
+    #   merkleProof = g.merkleProofCache
 
     await sleepAsync(rpcDelay)
 
@@ -387,6 +375,20 @@ proc indexToPath*(membershipIndex: UInt256, tree_depth: int): seq[byte] =
 
   debug "indexToPath", index = membershipIndex, path = result
 
+proc identitySecretToField*(secret: seq[byte]): array[32, byte] =
+  let str = cast[string](secret)
+  var field : StUint[256]
+  try:
+    field = parse(str, StUint[256])
+  except CatchableError:
+    error "Failed to parse identity secret", error = getCurrentExceptionMsg()
+  return field.toBytesLE()
+
+proc uint64ToField*(n: uint64): array[32, byte] =
+  ## Converts uint64 to 32-byte little-endian array with zero padding
+  var bytes = toBytes(n, Endianness.littleEndian)
+  result[0..<bytes.len] = bytes
+
 proc createZerokitWitness(
     g: OnchainGroupManager,
     data: seq[byte],
@@ -394,12 +396,14 @@ proc createZerokitWitness(
     messageId: MessageId,
     extNullifier: array[32, byte],
 ): RLNWitnessInput =
-  let identitySecret = g.idCredentials.get().idSecretHash.toArray32()
-    # seq[byte] to array[32, byte] and convert to little-endian
-  let userMsgLimit = g.userMessageLimit.get().toArray32LE()
-    # uint64 to array[32, byte] and convert to little-endian  
-  let msgId = messageId.toArray32LE()
-    # uint64 to array[32, byte] and convert to little-endian
+  let identitySecret = identitySecretToField(@[1'u8, 2'u8, 3'u8, 4'u8])
+  debug "---- identitySecret conversion", before = @[1'u8, 2'u8, 3'u8, 4'u8], after = identitySecret
+
+  let userMsgLimit = uint64ToField(256'u64)
+  debug "---- userMsgLimit conversion", before = 256'u64, after = userMsgLimit
+
+  let msgId = uint64ToField(1'u64)
+  debug "---- msgId conversion", before = 1'u64, after = msgId
 
   try:
     discard waitFor g.updateRoots()
@@ -426,11 +430,13 @@ proc createZerokitWitness(
   except CatchableError:
     error "Error fetching commitment index", error = getCurrentExceptionMsg()
 
-  debug "---- pathElements & pathIndex -----",
-    pathElements = pathElements,
-    pathIndex = pathIndex,
-    pathElementsLength = pathElements.len,
-    pathIndexLength = pathIndex.len
+  debug "---- pathElements", before = g.merkleProofCache, after = pathElements
+
+  # debug "---- pathElements & pathIndex -----",
+  #   pathElements = pathElements,
+  #   pathIndex = pathIndex,
+  #   pathElementsLength = pathElements.len,
+  #   pathIndexLength = pathIndex.len
 
   # Calculate hash using zerokit's hash_to_field equivalent
   let x = hash_to_field(data).toArray32LE() # convert to little-endian
--- a/waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
+++ b/waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
@@ -49,9 +49,9 @@ contract(WakuRlnContract):
   # this constant describes max message limit of rln contract
   proc MAX_MESSAGE_LIMIT(): UInt256 {.view.}
   # this function returns the merkleProof for a given index 
-  proc merkleProofElements(index: EthereumUInt40): seq[array[32, byte]] {.view.}
+  proc merkleProofElements(index: EthereumUInt40): seq[byte] {.view.}
   # this function returns the merkle root 
-  proc root(): Uint256 {.view.}
+  proc root(): UInt256 {.view.}
 
 type
   WakuRlnContractWithSender = Sender[WakuRlnContract]
@@ -67,7 +67,7 @@ type
     keystorePassword*: Option[string]
     registrationHandler*: Option[RegistrationHandler]
     latestProcessedBlock*: BlockNumber
-    merkleProofCache*: seq[array[32, byte]]
+    merkleProofCache*: seq[byte]
 
 proc setMetadata*(
     g: OnchainGroupManager, lastProcessedBlock = none(BlockNumber)
@@ -92,63 +92,27 @@ proc setMetadata*(
     return err("failed to persist rln metadata: " & getCurrentExceptionMsg())
   return ok()
 
-proc toArray32LE*(x: UInt256): array[32, byte] {.inline.} =
-  ## Convert UInt256 to byte array without endianness conversion
-  when nimvm:
-    for i in 0 ..< 32:
-      result[i] = byte((x shr (i * 8)).truncate(uint8) and 0xff)
-  else:
-    copyMem(addr result, unsafeAddr x, 32)
+proc uint64ToField*(n: uint64): array[32, byte] =
+  ## Converts uint64 to 32-byte little-endian array with zero padding
+  var bytes = toBytes(n, Endianness.littleEndian)
+  result[0 ..< bytes.len] = bytes
 
-proc toArray32LE*(x: array[32, byte]): array[32, byte] =
-  for i in 0 ..< 32:
-    result[i] = x[31 - i]
-  return result
+proc UInt256ToField*(v: UInt256): array[32, byte] =
+  var bytes: array[32, byte]
+  let vBytes = v.toBytesBE()
+  for i in 0 .. 31:
+    bytes[i] = vBytes[31 - i]
+  return bytes
 
-proc toArray32LE*(s: seq[byte]): array[32, byte] =
-  var output: array[32, byte]
-  for i in 0 ..< 32:
-    output[i] = 0
+proc seqToField*(s: seq[byte]): array[32, byte] =
+  result = default(array[32, byte])
   let len = min(s.len, 32)
   for i in 0 ..< len:
-    output[i] = s[31 - i]
-  return output
-
-proc toArray32*(s: seq[byte]): array[32, byte] =
-  var output: array[32, byte]
-  for i in 0 ..< 32:
-    output[i] = 0
-  let len = min(s.len, 32)
-  for i in 0 ..< len:
-    output[i] = s[i]
-  return output
-
-proc toArray32LE*(v: uint64): array[32, byte] =
-  let bytes = toBytes(v, Endianness.littleEndian)
-  var output: array[32, byte]
-  discard output.copyFrom(bytes)
-  return output
-
-proc toArray32*(v: uint64): array[32, byte] =
-  let bytes = toBytes(v)
-  var output: array[32, byte]
-  discard output.copyFrom(bytes)
-  return output
-
-# Hashes arbitrary signal to the underlying prime field.
-proc hash_to_field*(signal: seq[byte]): array[32, byte] =
-  var ctx: keccak256
-  ctx.init()
-  ctx.update(signal)
-  var hash_result = ctx.finish()
-
-  var hash: array[32, byte]
-  copyMem(hash[0].addr, hash_result.data[0].addr, 32)
-  toArray32LE(hash)
+    result[i] = s[i]
 
 proc fetchMerkleProofElements*(
     g: OnchainGroupManager
-): Future[Result[seq[array[32, byte]], string]] {.async.} =
+): Future[Result[seq[byte], string]] {.async.} =
   try:
     let membershipIndex = g.membershipIndex.get()
     let index40 = stuint(membershipIndex, 40)
@@ -173,19 +137,7 @@ proc fetchMerkleProofElements*(
 
     let responseBytes = await g.ethRpc.get().provider.eth_call(tx, "latest")
 
-    var i = 0
-    var merkleProof = newSeq[array[32, byte]]()
-    while (i * 32) + 31 < responseBytes.len:
-      var element: array[32, byte]
-      let startIndex = i * 32
-      let endIndex = startIndex + 31
-      element = responseBytes.toOpenArray(startIndex, endIndex)
-      merkleProof.add(element)
-      i += 1
-
-    debug "merkleProof", responseBytes = responseBytes, merkleProof = merkleProof
-
-    return ok(merkleProof)
+    return ok(responseBytes)
   except CatchableError:
     error "Failed to fetch Merkle proof elements",
       errMsg = getCurrentExceptionMsg(), index = g.membershipIndex.get()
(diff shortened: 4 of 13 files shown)


waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
```diff
@@ -226,6 +239,93 @@ method withdrawBatch*(
 ): Future[void] {.async: (raises: [Exception]).} =
   initializedGuard(g)
 
+method generateProof*(
```
jm-clius, line 242: Minor suggestion, but why not simply create a new group manager type for this new no-sync on-chain group manager (or rename and move the old one to something like SyncedGroupManager which can remain unused for now)?

jm-clius (commented):
Thanks! The bulk of the `generateProof` logic makes sense to me (at least on a high level). I do however have a question below about the caching of `merkleProofsByIndex` and the continuing handling of events to maintain the Merkle tree. Even if we don't deprecate the tree yet, I'm not sure I understand why we maintain a `merkleProofsByIndex` (for state we're about to deprecate). It could well be to speed up proof generation (not having to wait for on-chain proof for our own index), but wouldn't a regular timed task be more efficient than processing all change events?
Happy to discuss if I'm missing something or if there's a good reason to maintain some caching. :)
Then it seems to me that what's outstanding to deprecate the tree sync would be to implement `validateRoot` and `verifyProof` on the group manager?

darshankabariya:
> I do however have a question below about the caching of `merkleProofsByIndex` and the continuing handling of events to maintain the Merkle tree. Even if we don't deprecate the tree yet, I'm not sure I understand why we maintain a `merkleProofsByIndex` (for state we're about to deprecate). It could well be to speed up proof generation (not having to wait for on-chain proof for our own index), but wouldn't a regular timed task be more efficient than processing all change events? 

Hey @jm-clius, I use caching to speed up proof generation. Without it, let say the node would call the smart contract 80 times per epoch, which is inefficient. With caching, we can instantly use Merkle proofs and update the index through event handling. The trade-off is that updates only trigger when new members join, but I don’t think 80 members join every epoch. Let me know if I’m missing something. 

> Then it seems to me that what's outstanding to deprecate the tree sync would be to implement validateRoot and verifyProof on the group manager? 

Thanks for suggestion ! now it's more clear. 

One question: Do we use the index of other members in the group manager cache because light nodes rely on the relay node to generate proofs for them right ? If that’s the case, the trade-off is clear—if only 5 light nodes connect to a relay node, it would need to generate around 400 proofs (or just 80 if I’m mistaken) ( apx :- 400 smart contract call )

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
```diff
@@ -347,6 +447,11 @@ proc handleEvents(
         rateCommitments = rateCommitments,
         toRemoveIndices = removalIndices,
       )
+
+      for i in 0 ..< rateCommitments.len:
+        let index = startIndex + MembershipIndex(i)
+        await g.fetchMerkleProof(stuint(index, 256))
+
```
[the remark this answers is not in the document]
darshankabariya: Actually, if this efficiently targets merkleProofsByIndex, we can use it in generateProof.

From my understanding, this event notifies all remote subscribers whenever there’s an update in the smart contract. Is there a better place where I can move this code?

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
darshankabariya: Interesting idea. I agree with you it’s better to isolate the code, at least temporarily, so it doesn’t affect unit tests or other parts of the code. After review, we can use the new group manager for unit testing and then finally replace replace it with the on-chain group manager. is that you want think right ?

jm-clius: What I'm thinking is that there may eventually be interest in a group manager that does indeed sync the entire tree (as a form of further self-sovereignty). If we keep the current on-chain group manager as-is and simply default to a new group manager (without tree syncing), we leave this option open. No strong opinion, though. :) My thought was simply that it could be simpler to implement in this way - fully implement a new group manager from scratch and only switch to this new manager once it's fully tested and mature, rather than trying to adapt the existing one.

jm-clius:
>  I use caching to speed up proof generation. Without it, let say the node would call the smart contract 80 times per epoch, which is inefficient.

Do you mean if we generate up to 80 messages per epoch? It makes sense to me to cache the proof to avoid having to call the SC for every generated messages. My point is that I don't see why you need to maintain this for every member (or maintain the entire tree). Afaiu you only need the merkle proof for your own membership index.

In other words, all we need from the SC:
- your _own_ merkle proof, whenever you generate a message with a proof attached
- the current root of the tree, whenever you validate a received proof

In other words, there is no need to maintain a proof by index. Perhaps it's an acceptable strategy for now to refresh this single merkle proof whenever there is a change event, but the state can be much simpler.
To validate received messages, you would need the latest root. This can also be cached, I think, whenever there's a tree change, but this would still be the only necessary state to validate.

I haven't looked at it yet, but we can use this comparable go-waku implementation as reference: https://github.com/alrevuelta/go-waku-light

That said, I might be missing something here, so if there's any confusion, let's set up a call or find some expert eyes to help us think. :)

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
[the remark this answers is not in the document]
jm-clius: There are some questions here about when to cache the proof, but in principle I agree that caching the proof on tree change could be a useful strategy, However, the main issue here is that we do not need to cache merkle proofs for every member - we only need our own merkle proof (associated with our membership) to generate proofs. Worst case scenario is to lookup the merkle proof every time we want to generate a proof (i.e. every time we publish a message). Of course caching this would be very helpful, but we don't need the merkle proofs for any other member.

darshankabariya:
These three comments have cleared up a lot of my doubts. I now clearly understand the responsibilities of each group manager and how they differ from one another.

Currently, I’m building parallel implementations of the static and on-chain group managers. This new group manager will directly use the root and Merkle proof from the smart contract, eliminating the need for Merkle tree maintenance. As a result, we no longer have to deal with extensive syncing and event-related tasks.

I’m moving in that direction, and once it’s ready, I’ll ask for another review. Sounds cool, right? After getting an expert review, we can deprecate the current on-chain manager.

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
```diff
@@ -226,304 +254,143 @@ method withdrawBatch*(
...
+  ## Generates an RLN proof using the cached Merkle proof and custom witness
+  # Ensure identity credentials and membership index are set
+  if g.idCredentials.isNone():
+    return err("identity credentials are not set")
+  if g.membershipIndex.isNone():
+    return err("membership index is not set")
+  if g.userMessageLimit.isNone():
+    return err("user message limit is not set")
+
+  let merkleProofResult = await g.fetchMerkleProofElements()
```
jm-clius, line 290: Double checking my understanding here: this will now no longer actually use a cached Merkle proof, but perform a contract call to get the latest proof every time. Am I parsing this right? Not necessarily a problem right now, but we may want to perform _some_ caching (perhaps whenever the root changes?) - otherwise high rate publishing could result in many unnecessary contract calls.

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
```diff
@@ -112,6 +88,67 @@ proc setMetadata*(
...
+  ## Converts a UInt256 value to a MerkleNode (array[32, byte])
+  var merkleNode: MerkleNode
+  let byteArray = uint256Value.toBytesBE()
+
+  for i in 0 ..< min(byteArray.len, merkleNode.len):
+    merkleNode[i] = byteArray[i]
+
+  return merkleNode
+
+proc slideRootQueue*(g: OnchainGroupManager) {.async.} =
```
jm-clius, line 138: Afaict, we only update our window of valid roots when the `register` method is invoked? This will happen only once or never at all (if the node itself isn't registered) or am I missing something? We would need to either periodically update the root or still listen to membership added events and update the root whenever a member is added.

jm-clius (commented):
Thanks a lot for your hard work and patience here! This now looks much closer to what I had in mind, but I do think there are still a few changes necessary. Specifically, we should ensure that we still keep track of the latest root (e.g. by listening to contract events that may change the root or perhaps updating periodically). We may also want to cache the merkle proof for this rln instance membership index instead of doing a contract call every time we `generateProof`. Please do let me know if I'm missing something - quite possible. :)
One way to think of this task is that we stop using any of the underlying `rlnInstance` merkle tree operations (i.e. the rlnInstance kept an "offline" tree, but all of this is now maintained on-chain). However, we may still be interested in keeping "up to date" with on-chain changes that may affect the valid roots (which we use to verify other members' messages) or our own merkle proof (which we use to publish our own messages).

darshankabariya:
> Thanks a lot for your hard work and patience here! This now looks much closer to what I had in mind, but I do think there are still a few changes necessary. Specifically, we should ensure that we still keep track of the latest root (e.g. by listening to contract events that may change the root or perhaps updating periodically). We may also want to cache the merkle proof for this rln instance membership index instead of doing a contract call every time we `generateProof`. Please do let me know if I'm missing something - quite possible. :) 

Thanks for pointing that out. I think I’ve caught a bigger fish here.

About the root tracking problem—we’re no longer constructing Merkle trees locally, so we can’t generate all the intermediate roots ourselves like we used to. (Previously, for every registration event we fetched from the smart contract, we manually added it to the local Merkle tree and then stored the updated root.)

The main issue is that we only get the latest root from the contract via RPC calls. In a busy network, by the time we process one root update using block syncing, 3–4 other members might have already registered, causing us to miss those intermediate roots completely. This happens because the merkleRoot function only fetches the latest root.

We basically have a few options:
1. Add an event to the smart contract that fires specifically when the root changes.
2. Implement a time window (e.g., allowing registrations only every 10 seconds).
3. Use an intermediate indexer like The Graph protocol to track all root changes.

Am I making sense? Or should I explain the issue better? there a chance I’m also missing something?

For now, the temporary polling solution works, but it feels like a band-aid rather than a proper fix. I’ll stick with busy polling to keep it functional for the time being, but we should find a better solution.

cc ( @fryorcraken  @stubbsta  @Ivansete-status )

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
darshankabariya: Yes, it now triggers only once when our node requests RLN registration. Previously, it was triggered on every new registration of network bcz local MT.

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
darshankabariya: Done !!

fryorcraken:
Regarding root tracking. I see 3 solutions:

A. We just fetch `root` as often as possible

Not really practical because:
- it's a lot of RPC calls
- Especially on a L2 where we may have blocks every 1-2 seconds
- I wonder if a newly started node may have some issues as they may not validate/generate message correctly until they have a good root range.

B. Add the new root to exiting contract events

This is simple but has some issues:

- there is only one valid root per block - at least that's how `root` behaves
- which means if several membership are registered in a block, we need to be sure we do the right logic to only cache the root of the last tx in the block on the SC
- also, we need to understand if Infura/RPC API allows to only check "latest receipts". we don't want to sync all the receipts, just the most recent one when starting up
- Finally, it may increase gas cost

C. Cache the last N roots in the same contract, so that you can get all of them with one `roots` call (no need for `getMerkleProof` as we should always generate proof on most recent root.

- This might be the best in terms of RPC usage and nwaku sync cost.
  If we have a block every 2 seconds, and 5 roots available in the smart contract, then we need to call it every 6-8 seconds to be sure we don't miss one (it slides over ~10 seconds).
- the SC change should not be too hard, but not trivial either. Currently the Merkle root is [calculated offchain by the caller](https://github.com/waku-org/waku-rlnv2-contract/blob/5d8fd57ae1cbf391107dff4bdd3eb716de8d6c4e/src/WakuRlnV2.sol#L233), whereas now we will need to call it as insertion 
- there is will be increased gas cost, but likely to be similar to option B

I would suggest we go with `C`, or we try `B` and `C` and evaluate gas costs, and also evaluate if for `B` we can use the RPC API in a smart manner to not sync too many events.

---

Regarding migration, can we document a step by step @jm-clius @Ivansete-status?

1. Web app ready by @waku-org/js-waku 
2. This PR is ready (not yet merged)
3. We decide on smart contract instances we will use (the one already deployed on Linea? Status Network testnet)
4. Deploy smart contract if needed
5. Ensure webapp points to write smart contract
6. nwaku-compose PR to use new smart contract and update doc/readme
7. Merge this PR, release nwaku, merge nwaku-compose PR with new nwaku version, go!

Note that we could also merge this PR ASAP, the actual TWN switch is mostly driven by nwalu-compose

waku/waku_rln_relay/conversion_utils.nim
```diff
@@ -116,6 +116,20 @@ proc serialize*(memIndices: seq[MembershipIndex]): seq[byte] =
 
   return memIndicesBytes
 
+proc serialize*(witness: Witness): seq[byte] =
+  ## Serializes the witness into a byte array according to the RLN protocol format
+  var buffer: seq[byte]
+  buffer.add(witness.identity_secret)
+  buffer.add(witness.user_message_limit.toBytesBE())
```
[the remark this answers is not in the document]
darshankabariya: Thanks @seemenkina for pointing that out—noted !

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
```diff
@@ -226,310 +248,185 @@ method withdrawBatch*(
...
+
+      if g.validRoots.len == 0 or g.validRoots[g.validRoots.len - 1] != currentRoot:
+        let overflowCount = g.validRoots.len - AcceptableRootWindowSize + 1
+        if overflowCount > 0:
+          for i in 0 ..< overflowCount:
+            discard g.validRoots.popFirst()
+
+        g.validRoots.addLast(currentRoot)
+        info "Detected new Merkle root",
+          root = currentRoot.toHex, totalRoots = g.validRoots.len
```
jm-clius, line 420: Minor note, but it seems to me the `slideRootQueue` function already covers most of this functionality? In other words, perhaps we don't have to repeat the same heuristic to update the validRoots window and can use extracted utility function.

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
```diff
@@ -226,310 +248,185 @@ method withdrawBatch*(
...
+
+      let proofResult = await g.fetchMerkleProofElements()
```
jm-clius, line 422: If the `root` didn't change, there's no need to fetch a new merkle proof - it would be the same as before.

jm-clius (approved):
Thanks! From basic functionality perspective, once my comment below is addressed, I think this is covers the bases:
- track on-chain root changes to verify proofs (even if a bit hacky with a regular poll)
- use on-chain merkle proof for proof generation (which is cached whenever we detect a root change - minor changes necessary here)
- deprecate the tree sync

After we get more eyes on the code changes and the PR is merged, the next step would be to deploy this in the waku-simulator and test as well as we can. (cc @stubbsta). I'll respond in a next comment on the question of tracking root changes.

jm-clius:
@darshankabariya

> In a busy network, by the time we process one root update using block syncing, 3–4 other members might have already registered, causing us to miss those intermediate roots completely.

This is a good point. Remember that there is some predictability here as we can have up to one-root-per-block and block time should be more or less predictable (I think Linea Sepolia is currently `2 seconds`). But your point stands in that our active polling might miss intermediate roots.

> Those roots are crucial—we can’t afford to lose them, right?

It depends how we define "crucial" and on other network factors, such as current message rates. If you miss a root it means that you won't successfully validate messages that used that root for proof generation and will drop those messages. Note that you won't actively view the messages as spam (in other words, you won't penalise other nodes for forwarding such messages to you) - you simply won't continue relaying the message. We should of course try to prevent any such losses, but messages being dropped at some relay nodes is more or less expected behaviour in Waku (we're _not_ a reliable transport network). In other words, as long as it's relatively rare, the network can deal with a few misses. It's true that if you miss a lot of roots and therefore drop a lot of messages, eventually other peers will disconnect from your node as they perceive it to be unreliable.

@fryorcraken 

> Cache the last N roots in the same contract, so that you can get all of them with one roots call

Yes, I think this is a good approach. This also means that we are not blocked currently as we can use the current polling to get a single root and simply upgrade to less regular polling with a window of roots in future. It would also be possible to check that every polling call actually has one or two roots that overlap with the existing local window as "proof" that we didn't miss any intermediate roots.

I'm also wondering if longer term some version of your solution `B` is possible where the contract emits only an `onRootChange` event. Presumably if we can cache a series of roots on the contract we can emit when see a change. Not sure how it will affect contract complexity/gas, though, so would need some Vac SC expertise.

> Regarding migration, can we document a step by step

I think we'll just use the existing Linea deployment, which will simplify the steps a bit. Why not merge this PR when it's ready? We can then easily add fixes and the changes will only take effect on TWN once we update nwaku-compose. I'd also add a waku-simulator step to test the deprecated tree sync though, which may highlight a few issues:

1. Web app ready by @waku-org/js-waku and pointing to new contract
2. This PR is merged
3. waku-simulator and other tests for deprecated tree sync. Likely iterate through some fixes here.
4. nwaku-compose PR to use new smart contract and update doc/readme
5. merge nwaku-compose PR with new nwaku version, go!

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
darshankabariya: Noted !

waku/waku_rln_relay/group_manager/group_manager_base.nim
```diff
@@ -201,8 +201,6 @@ method generateProof*(
     ).valueOr:
       return err("proof generation failed: " & $error)
 
-  waku_rln_remaining_proofs_per_epoch.dec()
```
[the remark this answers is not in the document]
darshankabariya: I didn’t remove it — I just changed the location of this metric. It’s related to RLN; we override the base generateProof function, which is why the original one isn’t actually used. So technically, that function is never called. If I hadn’t moved the metric, it wouldn’t have shown the correct data. @fryorcraken

tests/node/test_wakunode_relay_rln.nim
```diff
@@ -452,7 +452,7 @@ suite "Waku RlnRelay - End to End - OnChain":
       except CatchableError:
         assert true
 
-    asyncTest "Unregistered contract":
+    xasyncTest "Unregistered contract":
```
Ivansete-status, line 455: I would not disable the tests. They should work to be able to merge the PR :)

waku/waku_rln_relay/conversion_utils.nim
```diff
@@ -116,6 +116,24 @@ proc serialize*(memIndices: seq[MembershipIndex]): seq[byte] =
 
   return memIndicesBytes
 
+proc serialize*(witness: Witness): seq[byte] =
```
Ivansete-status, line 119: 1. Where is the opposite proc for `serialize`?
2. Why `serialize` is needed? ( not used now )

waku/waku_rln_relay/rln_relay.nim
```diff
@@ -482,6 +479,18 @@ proc mount(
 
   # Start epoch monitoring in the background
   wakuRlnRelay.epochMonitorFuture = monitorEpochs(wakuRlnRelay)
+
+  # Start tracking root changes after successful initialization and registration
+  debug "~~~~~~~~~~~~~ Starting root tracking ~~~~~~~~~~~~~~~~"
+  debug "~~~~~~~~~~~~~ groupManager ~~~~~~~~~~~~~~~~"
```
Ivansete-status: We should keep similar format :)
```suggestion
  debug "Starting root tracking groupManager"
```

waku/waku_rln_relay/rln_relay.nim
```diff
@@ -482,6 +479,18 @@ proc mount(
...
+
+  await sleepAsync(5 * 60 * 1000)
```
Ivansete-status: Why 5min wait?

waku/waku_rln_relay/rln_relay.nim
```diff
@@ -482,6 +479,18 @@ proc mount(
...
+  if conf.rlnRelayDynamic:
+    debug "~~~~~~~~~~~~~ groupManager is dynamic ~~~~~~~~~~~~~~~~"
```
Ivansete-status:
```suggestion
    debug "groupManager is dynamic"
```

Ivansete-status (commented):
Thanks for it!
Some comments/questions

tests/node/test_wakunode_relay_rln.nim
darshankabariya: This is just temporary — I’ll update it after debugging.

waku/waku_rln_relay/conversion_utils.nim
darshankabariya: As per Aryaman’s suggestion, I used the new API `generate_proof_with_witness` for generating the proof instead of `generate_proof`. It looks like this introduces a new structure ( witness ) that combines all the necessary information and serializes it into a string. 

We only need serialization during proof generation or verification, right?
I believe deserialization happens on the Zerokit side — @seemenkina, can you confirm?

waku/waku_rln_relay/rln_relay.nim
darshankabariya: All of this is just for debugging — we’ll get rid of it once we’re confident in the code.

waku/waku_rln_relay/conversion_utils.nim
seemenkina: Yes, we wait for the data in a serialised format and return it in the same format. Deserialisation happens on the zerokit side, both for proof generation and for verification

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
```diff
@@ -112,6 +90,132 @@ proc setMetadata*(
...
   return ok()
 
+proc fetchMerkleProofElements*(
+    g: OnchainGroupManager
+): Future[Result[seq[Uint256], string]] {.async.} =
+  let membershipIndex = g.membershipIndex.get()
+  debug " ------ Fetching merkle proof", index = membershipIndex
+  try:
+    # First check if the index is valid
+    let commitmentIndexInvocation = g.wakuRlnContract.get().commitmentIndex()
```
stubbsta, line 100: According to the contract (https://github.com/waku-org/waku-rlnv2-contract/blob/64df4593c6a14e43b8b0e9b396d2f4772bb08b34/src/WakuRlnV2.sol#L40) this call returns the "The index of the next member to be registered" and not the current one.

darshankabariya: Thanks for that.

waku/waku_rln_relay/group_manager/group_manager_base.nim
```diff
@@ -145,6 +145,17 @@ method validateRoot*(
...
 ): bool {.base, gcsafe, raises: [].} =
   ## validates the root against the valid roots queue
+  # Print all validRoots in one line with square brackets
+  var rootsStr = "["
+  var first = true
+  for r in g.validRoots.items():
+    if not first:
+      rootsStr.add(", ")
+    rootsStr.add($r)
+    first = false
```
Ivansete-status, line 155: Are you sure this is needed? I thought seq[T] could get printed directly.

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
```diff
@@ -112,6 +93,135 @@ proc setMetadata*(
...
+            element[j] = responseBytes[startIndex + j]
+        merkleProof.add(element)
+      else:
+        var element: array[32, byte]
+        merkleProof.add(element)
+
+    return ok(merkleProof)
+  except CatchableError:
+    error "------ Failed to fetch Merkle proof elements ------",
+      errMsg = getCurrentExceptionMsg(), index = g.membershipIndex.get()
```
Ivansete-status, line 139:
```suggestion
    error "Failed to fetch Merkle proof elements",
      error = getCurrentExceptionMsg(), index = g.membershipIndex.get()
```

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
```diff
@@ -112,6 +93,135 @@ proc setMetadata*(
...
+
+proc fetchMerkleRoot*(
+    g: OnchainGroupManager
+): Future[Result[Uint256, string]] {.async.} =
+  try:
+    let merkleRootInvocation = g.wakuRlnContract.get().root()
+    let merkleRoot = await merkleRootInvocation.call()
+    return ok(merkleRoot)
+  except CatchableError:
+    error "Failed to fetch Merkle root", errMsg = getCurrentExceptionMsg()
```
Ivansete-status, line 150:
```suggestion
    error "Failed to fetch Merkle root", error = getCurrentExceptionMsg()
```

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
```diff
@@ -112,6 +93,135 @@ proc setMetadata*(
...
+  let merkleRoot = toMerkleNode(rootRes.get())
+  if g.validRoots.len > 0 and g.validRoots[g.validRoots.len - 1] != merkleRoot:
+    let overflowCount = g.validRoots.len - AcceptableRootWindowSize + 1
+    if overflowCount > 0:
+      for i in 0 ..< overflowCount:
+        discard g.validRoots.popFirst()
+
+    g.validRoots.addLast(merkleRoot)
+    debug "~~~~~~~~~~~~~ Detected new Merkle root ~~~~~~~~~~~~~~~~",
+      root = merkleRoot.toHex, totalRoots = g.validRoots.len
```
Ivansete-status, line 192: That kind of special format can only happen in local debugging ;P
```suggestion
    debug "Detected new Merkle root",
      root = merkleRoot.toHex, totalRoots = g.validRoots.len
```

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
```diff
@@ -112,6 +93,135 @@ proc setMetadata*(
...
+    return true
+  else:
+    debug "~~~~~~~~~~~~~ No new Merkle root ~~~~~~~~~~~~~~~~",
+      root = merkleRoot.toHex, totalRoots = g.validRoots.len
```
Ivansete-status, line 196:
```suggestion
    debug "No new Merkle root",
      root = merkleRoot.toHex, totalRoots = g.validRoots.len
```

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
```diff
@@ -226,304 +320,173 @@ method withdrawBatch*(
...
-      "failed to get the history/reconcile missed blocks: " & getCurrentExceptionMsg(),
-    )
+proc toArray32*(s: seq[byte]): array[32, byte] =
+  var output: array[32, byte]
+  for i in 0 ..< 32:
+    output[i] = 0
+  let len = min(s.len, 32)
+  for i in 0 ..< len:
+    output[i] = s[s.len - 1 - i]
+  return output
```
Ivansete-status, line 330: There should be an existing utility. In any case, that doesn't seem to be the appropriate place for that

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
```diff
@@ -226,304 +320,173 @@ method withdrawBatch*(
...
+  ## Generates an RLN proof using the cached Merkle proof and custom witness
+  # Ensure identity credentials and membership index are set
+  if g.idCredentials.isNone():
+    return err("identity credentials are not set")
+  if g.membershipIndex.isNone():
+    return err("membership index is not set")
+  if g.userMessageLimit.isNone():
+    return err("user message limit is not set")
+
+  debug "------ calling generateProof from generateProof from group_manager onchain ------",
```
Ivansete-status, line 355:
```suggestion
  debug "calling generateProof from generateProof from group_manager onchain",
```

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
```diff
@@ -226,304 +320,173 @@ method withdrawBatch*(
...
+  var inputBuffer = toBuffer(serializedWitness)
+
+  # Generate the proof using the zerokit API
+  var outputBuffer: Buffer
+  let success =
+    generate_proof_with_witness(g.rlnInstance, addr inputBuffer, addr outputBuffer)
+  if not success:
+    return err("Failed to generate proof")
+  else:
+    debug "------ Proof generated successfully --------"
```
Ivansete-status, line 411:
```suggestion
    debug "Proof generated successfully"
```

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
```diff
@@ -226,304 +320,173 @@ method withdrawBatch*(
...
+    identity_secret: g.idCredentials.get().idSecretHash.toArray32(),
+    user_message_limit: serialize(g.userMessageLimit.get()),
+    message_id: serialize(messageId),
+    path_elements: g.merkleProofCache,
+    identity_path_index: indexToPath(g.membershipIndex.get()),
+    x: toArray32(data),
+    external_nullifier: externalNullifierRes.get(),
+  )
+
+  debug "------ Generating proof with witness ------",
```
Ivansete-status, line 383:
```suggestion
  debug "Generating proof with witness",
```

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
```diff
@@ -226,304 +320,173 @@ method withdrawBatch*(
...
+    identity_secret = inHex(witness.identity_secret),
+    user_message_limit = inHex(witness.user_message_limit),
+    message_id = inHex(witness.message_id),
+    path_elements = witness.path_elements.map(inHex),
+    identity_path_index = witness.identity_path_index.mapIt($it).join(", "),
+    x = inHex(witness.x),
+    external_nullifier = inHex(witness.external_nullifier)
+
+  debug "------ Witness parameters ------",
```
Ivansete-status, line 392:
```suggestion
  debug "Witness parameters",
```

Ivansete-status (commented):
Some more comments so far. Thanks for it!
@jm-clius , @darshankabariya - shall we perform these changes against a feature branch instead? For example, `feat/rln-mainnet`?

I have the feeling that there are oustanding tasks aside from this PR in order to be compliant with new smart contract definition (RLN v2 if I'm not mistaken.)

waku/waku_rln_relay/group_manager/group_manager_base.nim
darshankabariya: good point !! but i already override in onchain group manager but will change it.

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
darshankabariya: Hi @Ivansete-status, some code may look unnecessary, but I’m still debugging the ZeroKit API. I added extra comments to separate things from the logs. I’ll clean it up once it’s working.

darshankabariya:
> Some more comments so far. Thanks for it! @jm-clius , @darshankabariya - shall we perform these changes against a feature branch instead? For example, `feat/rln-mainnet`?
> 
> I have the feeling that there are oustanding tasks aside from this PR in order to be compliant with new smart contract definition (RLN v2 if I'm not mistaken.)

I don’t think it works directly with the new RLN contract, since the merkleProofElements API has changed. But it shouldn’t be too difficult once the ZeroKit API is working.

My plan is to first address the tree sync issue, then update to the new contract. We can handle the new contract address and related API changes in a separate PR.

From my understanding, we can migrate to RLNv2 with these steps—right? (cc @jm-clius @stubbsta)
Alternatively, we could merge RLNv2 first and update the API before fixing the tree sync. That should also work, I think.

stubbsta:
> From my understanding, we can migrate to RLNv2 with these steps—right? (cc @jm-clius @stubbsta) Alternatively, we could merge RLNv2 first and update the API before fixing the tree sync. That should also work, I think.

Correct, we can move from off-chain merkle proof validation to on-chain merkle proofs for the current contract implementation.
Then, if we want to, with a separate PR, we can update the contract specific functionality for the new incentivisation contract.

stubbsta:
> @darshankabariya
> 
> > In a busy network, by the time we process one root update using block syncing, 3–4 other members might have already registered, causing us to miss those intermediate roots completely.
> 

Apologies, I know I'm very late to this conversation, I was wondering whether we want to play around with increasing the node's 
`AcceptableRootWindowSize` in addition to what has already been discussed. There might be a sweet spot for this value depending on registration rate, message rate and proof generation time. It is definitely something we can look into with a waku-simulator test scenario, to see just how many messages get rejected for being out of the validroots range depending on what this value is.

```
waku/waku_rln_relay/constants.nim:
# Acceptable roots for merkle root validation of incoming messages
const AcceptableRootWindowSize* = 5
```

darshankabariya:
> Apologies, I know I'm very late to this conversation, I was wondering whether we want to play around with increasing the node's `AcceptableRootWindowSize` in addition to what has already been discussed. There might be a sweet spot for this value depending on registration rate, message rate and proof generation time. It is definitely something we can look into with a waku-simulator test scenario, to see just how many messages get rejected for being out of the validroots range depending on what this value is.
> 
> ```
> waku/waku_rln_relay/constants.nim:
> # Acceptable roots for merkle root validation of incoming messages
> const AcceptableRootWindowSize* = 5
> ```

Interesting idea. With a window size of 5, there’s a high chance it occasionally misses the front root, which likely adds some validation cost. Increasing the window size should improve acceptance. I’ve managed to generate the proof, so you’ll soon be able to test this trade-off in the Waku simulator. i will you know, thanks for head up !!

vendor/zerokit
stubbsta, line 1: The zerokit version also needs to be updated in the Makefile for `make docker-image` to work.
https://github.com/waku-org/nwaku/blob/master/Makefile#L162

waku/waku_rln_relay/rln_relay.nim
```diff
@@ -463,13 +467,11 @@ proc mount(
       membershipIndex: conf.rlnRelayCredIndex,
       onFatalErrorAction: conf.onFatalErrorAction,
     )
+    asyncSpawn trackRootChanges(cast[OnchainGroupManager](groupManager))
```
Ivansete-status, line 470: This is not the idiomatic way we use to spawn an async task.

For example, see how we manage a `Future` object in the following example. Notice that we also need to gracefully shutdown the async tasks:
https://github.com/waku-org/nwaku/blob/72d1fa0d349960280563a1d738efdf6a33ab6d7d/waku/waku_archive/archive.nim#L50

fryorcraken:
> . There might be a sweet spot for this value depending on registration rate, message rate and proof generation time.

The sweet spot will be related to the block time of the chain on which the contract is deployed.

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
```diff
@@ -342,14 +344,12 @@ method withdrawBatch*(
 ): Future[void] {.async: (raises: [Exception]).} =
   initializedGuard(g)
 
-proc indexToPath*(membershipIndex: uint): seq[byte] =
-  const TREE_DEPTH = 20 # RLN uses 20-level Merkle trees
-  result = newSeq[byte](TREE_DEPTH)
-
-  # Convert index to little-endian bit array
+proc indexToPath*(membershipIndex: uint, tree_depth: int): seq[byte] =
+  result = newSeq[byte](tree_depth)
```
Ivansete-status, line 348: We shouldn't use the implicit `result` object

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
```diff
@@ -368,13 +368,32 @@ proc createZerokitWitness(
   let msgId = messageId.toArray32LE()
     # uint64 to array[32, byte] and convert to little-endian
 
-  # Convert path elements to little-endian byte arrays
+  try:
+    discard waitFor g.updateRoots()
```
Ivansete-status, line 372: We cannot block the whole node in this point. I'd rather use `await`

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
```diff
@@ -368,13 +368,32 @@ proc createZerokitWitness(
...
+  except CatchableError:
+    error "Error updating roots", error = getCurrentExceptionMsg()
+
+  try:
+    let proofResult = waitFor g.fetchMerkleProofElements()
```
Ivansete-status, line 377: Similarly to the previous comment ;P

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
```diff
@@ -368,13 +368,32 @@ proc createZerokitWitness(
...
   var pathElements: seq[array[32, byte]]
   for elem in g.merkleProofCache:
     pathElements.add(toArray32LE(elem)) # convert every element to little-endian
 
   # Convert index to byte array (no endianness needed for path index)
-  let pathIndex = indexToPath(g.membershipIndex.get()) # uint to seq[byte]
+  let pathIndex = indexToPath(g.membershipIndex.get(), pathElements.len)
+    # uint to seq[byte]
+
+  debug "---- pathElements & pathIndex -----",
```
Ivansete-status, line 392: Let's try to keep similar log format :)
```suggestion
  debug "pathElements & pathIndex",
```

waku/waku_rln_relay/rln_relay.nim
darshankabariya: Thanks, Ivan! I’ll make the updates accordingly.

waku/waku_rln_relay/constants.nim
```diff
@@ -5,7 +5,7 @@ import ./protocol_types
 import ../waku_keystore
 
 # Acceptable roots for merkle root validation of incoming messages
-const AcceptableRootWindowSize* = 5
```
fryorcraken, line 8: Would be good to document the memory implications

waku/waku_rln_relay/group_manager/group_manager_base.nim
fryorcraken: I think you should be able to use `$` if you import `std/strutils`

waku/waku_rln_relay/protocol_types.nim
```diff
@@ -52,6 +52,18 @@ type RateLimitProof* = object
   ## the external nullifier used for the generation of the `proof` (derived from poseidon([epoch, rln_identifier]))
   externalNullifier*: ExternalNullifier
 
+type
+  Fr = array[32, byte] # Field element representation (256 bits)
+
+  RLNWitnessInput* = object
```
fryorcraken, line 58:
```suggestion
  RlnWitnessInput* = object
```
Consistence across the codebase

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
```diff
@@ -609,53 +564,28 @@ method init*(g: OnchainGroupManager): Future[GroupManagerResult[void]] {.async.}
...
-      await g.startOnchainSync()
-    except CatchableError, Exception:
-      g.onFatalErrorAction(
-        "failed to restart group sync" & ": " & getCurrentExceptionMsg()
-      )
-
   ethRpc.ondisconnect = proc() =
     asyncSpawn onDisconnect()
 
   waku_rln_number_registered_memberships.set(int64(g.rlnInstance.leavesSet()))
```
stubbsta, line 584: We have seen in waku-simulator tests with this PR image that the `waku_rln_number_registered_memberships` metric remains zero even with successful membership insertions.
Can you investigate to see why that is?

waku/waku_rln_relay/constants.nim
```diff
@@ -5,7 +5,7 @@ import ./protocol_types
...
+const AcceptableRootWindowSize* = 50
```
stubbsta, line 8: this should probably be set back to 5

waku/waku_rln_relay/protocol_metrics.nim
```diff
@@ -116,5 +120,6 @@ proc getRlnMetricsLogger*(): RLNMetricsLogger =
         info "Total proofs verified", count = freshProofsVerifiedCount
         info "Total proofs generated", count = freshProofsGeneratedCount
         info "Total proofs remaining", count = freshProofsRemainingCount
+        info "Total registered member", count = freshRegisteredMemberCount
```
stubbsta, line 123: I think this should use the plural "members"

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
```diff
@@ -112,6 +89,127 @@ proc setMetadata*(
     return err("failed to persist rln metadata: " & getCurrentExceptionMsg())
   return ok()
 
+proc uint64ToField*(n: uint64): array[32, byte] =
```
stubbsta, line 92: Would all these procs that are conversions with name "ToField" not be better suited to be placed in the conversion_utils.nim file?

stubbsta (commented):
I've made a few comments about minor changes to consider.
I also want to restate here what I said in discord regarding metrics that need to be checked and maybe modified in a separate PR as they currently show zero or no value:
- waku_rln_proof_generation_duration_seconds
- waku_rln_membership_insertion_duration_seconds

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
```diff
@@ -10,40 +10,45 @@ import
...
 
-from strutils import parseHexInt
-
 export group_manager_base
 
 logScope:
   topics = "waku rln_relay onchain_group_manager"
 
+type UInt40* = StUint[40]
+type UInt32* = StUint[32]
```
Ivansete-status, line 30: Better move these definitions to `protocol_types.nim`

Ivansete-status (commented):
Some more comments.
I also think now is a good opportunity to make tests work before we forget it.

darshankabariya:
> I've made a few comments about minor changes to consider. I also want to restate here what I said in discord regarding metrics that need to be checked and maybe modified in a separate PR as they currently show zero or no value:

reply in thread.

darshankabariya:
> > I've made a few comments about minor changes to consider. I also want to restate here what I said in discord regarding metrics that need to be checked and maybe modified in a separate PR as they currently show zero or no value:
> 
> Thanks for review, reply in thread ( cc @stubbsta )

darshankabariya:
https://github.com/waku-org/nwaku/pull/3312#discussion_r2060369375

Yeah, I did that intentionally — added the related discussion to explain why it’s needed. ( cc @stubbsta )

Screenshot 2025-04-25 at 8 30 18 PM

Ivansete-status (commented):
Adding some more nitpick comments

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
```diff
@@ -112,33 +87,103 @@ proc setMetadata*(
...
+): auto =
+  retryWrapper(res, RetryStrategy.new(), errStr, g.onFatalErrorAction):
+    body
+
+method validateRoot*(g: OnchainGroupManager, root: MerkleNode): bool =
+  if g.validRoots.find(root) >= 0:
+    return true
+  return false
+
+proc updateRoots*(g: OnchainGroupManager): Future[bool] {.async.} =
```
Ivansete-status, line 148: This seems can be private
```suggestion
proc updateRoots(g: OnchainGroupManager): Future[bool] {.async.} =
```

waku/waku_rln_relay/rln_relay.nim
```diff
@@ -463,9 +464,10 @@ proc mount(
   # Initialize the groupManager
   (await groupManager.init()).isOkOr:
     return err("could not initialize the group manager: " & $error)
-  # Start the group sync
-  (await groupManager.startGroupSync()).isOkOr:
-    return err("could not start the group sync: " & $error)
+
+  if groupManager of OnchainGroupManager:
+    let onchainManager = cast[OnchainGroupManager](groupManager)
+    asyncSpawn trackRootChanges(onchainManager)
```
Ivansete-status, line 470: Let's avoid anonymous async tasks. Instead, it is better to follow the same as `wakuRlnRelay.epochMonitorFuture`
```suggestion
    asyncSpawn trackRootChanges(onchainManager)
```

Btw - both the new `rootChangesListenerFut` (or other name) handle and `wakuRlnRelay.epochMonitorFuture` should be cancelled within the `stop` proc. With that, we can achieve graceful shutdown.

darshankabariya:
Hi @Ivansete-status, thanks for the suggestions! I’ve implemented all of them except one—adding err() in trackRootChanges. I believe returning an error therenot be a good idea bcz if any of SC call fails, we still want the root tracking to continue instead of stopping the whole process. Let me know if you see a better way to handle that—happy to discuss!

Ivansete-status (commented):
Some more comments, thanks ! :raised_hands:

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
```diff
@@ -132,7 +132,10 @@ proc fetchMerkleRoot*(
 
 template initializedGuard(g: OnchainGroupManager): untyped =
   if not g.initialized:
-    raise newException(CatchableError, "OnchainGroupManager is not initialized")
+    raise newException(
+      CatchableError,
+      "OnchainGroupManager is not initialized: " & getCurrentExceptionMsg(),
```
Ivansete-status, line 137: I think `getCurrentExceptionMsg()` only makes sense when capturing an exception, not when raising it :D
```suggestion
      "OnchainGroupManager is not initialized"
```

waku/waku_rln_relay/constants.nim
stubbsta: Hi @darshankabariya, this value is still `50`, it's not clear whether this is intentional? When resolving a comment/conversation it is very helpful if the commit ID where the comment was addressed is added to that comment or otherwise commenting to say why no changes were made.

Ivansete-status: Hello @stubbsta !
@darshankabariya answered in the following comment: https://github.com/waku-org/nwaku/pull/3312#issuecomment-2838141022

waku/waku_rln_relay/rln_relay.nim
```diff
@@ -476,6 +475,15 @@ proc mount(
...
   )
 
+  proc runTrackRootChanges(g: OnchainGroupManager) {.async.} =
+    (await g.trackRootChanges()).isOkOr:
+      error "Failed to track root changes in background", error
+
+  # track root changes on smart contract merkle tree
+  if groupManager of OnchainGroupManager:
+    let onchainManager = cast[OnchainGroupManager](groupManager)
+    wakuRlnRelay.rootChangesFuture = runTrackRootChanges(onchainManager)
```
Ivansete-status, line 485: It seems this can be simplified:
```suggestion
  # track root changes on smart contract merkle tree
  if groupManager of OnchainGroupManager:
    let onchainManager = cast[OnchainGroupManager](groupManager)
    (await g.trackRootChanges()).isOkOr:
      return err("failed to track root changes in background: " & error)
```

tests/node/test_wakunode_relay_rln.nim
Ivansete-status: This seems the only pending test :)

Ivansete-status (commented):
Some more comments. Almost there! :)

Ivansete-status (approved):
LGTM! Great work! :partying_face:

stubbsta (approved):
Excellent work on this, LGTM!

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
```diff
@@ -226,304 +284,173 @@ method withdrawBatch*(
...
+    addr proofBuffer, # (proof + signal)
+    addr rootsBuffer, # valid Merkle roots
+    addr validProof # will be set by the FFI call
+    ,
+  )
+
+  if not ffiOk:
+    return err("could not verify the proof")
+  else:
+    debug "Proof verified successfully !"
```
NagyZoltanPeter, line 451: Is it always a relevant debug log? Maybe trace can be enough?

NagyZoltanPeter (approved):
I'm truly amazed of this great work!!! Thank you.
Btw, I saw some unaddressed comments... ;-)

PR #3319, opened 2025-03-11
chore: less logs for rendezvous

SionoiS:
In order to stop spammy rendez-vous errors, errors are now pooled and when an error is encountered the interval between tries is increased.

@fryorcraken

### DIFF
--- a/waku/waku_rendezvous/protocol.nim
+++ b/waku/waku_rendezvous/protocol.nim
@@ -32,6 +32,7 @@ type WakuRendezVous* = ref object
   relayShard: RelayShards
   capabilities: seq[Capabilities]
 
+  registrationInterval: timer.Duration
   periodicRegistrationFut: Future[void]
 
 proc batchAdvertise*(
@@ -62,7 +63,7 @@ proc batchAdvertise*(
         fut.read()
 
       if catchable.isErr():
-        trace "rendezvous dial failed", error = catchable.error.msg
+        error "rendezvous dial failed", error = catchable.error.msg
         continue
 
       let connOpt = catchable.get()
@@ -91,7 +92,7 @@ proc batchRequest*(
 ): Future[Result[seq[PeerRecord], string]] {.async: (raises: []).} =
   ## Request all records from all rendezvous peers matching a namespace
 
-  # rendezvous.request except already opened connections
+  # rendezvous.request expect already opened connections
   # must dial first
   var futs = collect(newSeq):
     for peerId in peers:
@@ -111,7 +112,7 @@ proc batchRequest*(
         fut.read()
 
       if catchable.isErr():
-        trace "rendezvous dial failed", error = catchable.error.msg
+        error "rendezvous dial failed", error = catchable.error.msg
         continue
 
       let connOpt = catchable.get()
@@ -143,13 +144,15 @@ proc advertiseAll(
     for pubsubTopic in pubsubTopics:
       # Get a random RDV peer for that shard
       let rpi = self.peerManager.selectPeer(RendezVousCodec, some($pubsubTopic)).valueOr:
-        trace "could not get a peer supporting RendezVousCodec"
         continue
 
       let namespace = computeNamespace(pubsubTopic.clusterId, pubsubTopic.shardId)
 
       # Advertise yourself on that peer
       self.batchAdvertise(namespace, DefaultRegistrationTTL, @[rpi.peerId])
+
+  if futs.len < 1:
+    return err("could not get a peer supporting RendezVousCodec")
 
   let catchable = catch:
     await allFinished(futs)
@@ -159,7 +162,7 @@ proc advertiseAll(
 
   for fut in catchable.get():
     if fut.failed():
-      trace "rendezvous advertisement failed", error = fut.error.msg
+      error "rendezvous advertisement failed", error = fut.error.msg
 
   debug "waku rendezvous advertisements finished"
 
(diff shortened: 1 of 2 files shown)


NagyZoltanPeter (approved):
I know the issue raised by @fryorcraken and it is valid, but still weird to see failure logged as trace.

Maybe we should consider a kind of circuit breaker approach to log only few times (once) within a certain amount of time an error. It could be chronicles feature maybe.

+ why not add a metric counter to report such errors?

SionoiS:
> I know the issue raised by @fryorcraken and it is valid, but still weird to see failure logged as trace.
> 
> Maybe we should consider a kind of circuit breaker approach to log only few times (once) within a certain amount of time an error. It could be chronicles feature maybe.
> 
> * why not add a metric counter to report such errors?

I like the idea @NagyZoltanPeter ! The node would still need to check for new peers from time to time though. :thinking: 

Maybe increasing the interval if no peer is found would reduce the logs AND still try to find peers.

waku/waku_rendezvous/protocol.nim
```diff
@@ -91,7 +92,7 @@ proc batchRequest*(
 ): Future[Result[seq[PeerRecord], string]] {.async: (raises: []).} =
   ## Request all records from all rendezvous peers matching a namespace
 
-  # rendezvous.request except already opened connections
+  # rendezvous.request expect already opened connections
```
fryorcraken, line 95:
```suggestion
  # rendezvous.request expects already opened connections
```

fryorcraken (commented):
When I start my node, I see this being blurbed 8 lines at once:

```
ERR 2025-03-12 14:37:18.992+11:00 could not get a peer supporting RendezVousCodec topics="waku rendezvous" tid=1760362 file=protocol.nim:181
ERR 2025-03-12 14:37:18.992+11:00 could not get a peer supporting RendezVousCodec topics="waku rendezvous" tid=1760362 file=protocol.nim:181
ERR 2025-03-12 14:37:18.992+11:00 could not get a peer supporting RendezVousCodec topics="waku rendezvous" tid=1760362 file=protocol.nim:181
ERR 2025-03-12 14:37:18.992+11:00 could not get a peer supporting RendezVousCodec topics="waku rendezvous" tid=1760362 file=protocol.nim:181
ERR 2025-03-12 14:37:18.992+11:00 could not get a peer supporting RendezVousCodec topics="waku rendezvous" tid=1760362 file=protocol.nim:181
ERR 2025-03-12 14:37:18.992+11:00 could not get a peer supporting RendezVousCodec topics="waku rendezvous" tid=1760362 file=protocol.nim:181
ERR 2025-03-12 14:37:18.992+11:00 could not get a peer supporting RendezVousCodec topics="waku rendezvous" tid=1760362 file=protocol.nim:181
ERR 2025-03-12 14:37:18.992+11:00 could not get a peer supporting RendezVousCodec topics="waku rendezvous" tid=1760362 file=protocol.nim:181
```

Is it once per shard?

Also I can see it appears every 10 seconds. I assume there is some sort of loop that tries every 10s?

Shouldn't it be event based?

1. New peer is discovered/added to peer store
2. Event is emitted
3. rendezvous protocol listens to event and grabs peer
4. check if rendezvous may be supported (via ENR?if discovered)
5. proceed to request for connection if yes
6. confirm at identify whether rendezvous is supported

SionoiS:
> * New peer is discovered/added to peer store
> * Event is emitted
> * rendezvous protocol listens to event and grabs peer

Yes I would love to but we don't have a system in place to do that AFAIK.
In an ideal world nwaku would be based on some actor model and message passing between components would be that easy.

> Is it once per shard?

Also I can see it appears every 10 seconds. I assume there is some sort of loop that tries every 10s?

Yes once per shard every 10s because the TTL is only 60s and we need a good pool of registrations for rendezvous to be effective in this configuration (rendezvous was not design for short TTL).

SionoiS:
> I think the logic is much better now. But seems the message itself is not very informative to be an error. What can a user/operator do about it? Is it really an error? So maybe within this scope to move it to info or maybe only trace level.

Is `warn` ever used in nwaku maybe it would fit in this case?

NagyZoltanPeter:
> > I think the logic is much better now. But seems the message itself is not very informative to be an error. What can a user/operator do about it? Is it really an error? So maybe within this scope to move it to info or maybe only trace level.
> 
> Is `warn` ever used in nwaku maybe it would fit in this case?

Yes, warn is ok IMHO

fryorcraken:
> Yes I would love to but we don't have a system in place to do that AFAIK.
> In an ideal world nwaku would be based on some actor model and message passing between components would be that easy.

@Ivansete-status , thoughts? We're going to continue adding to nwaku, the architecture needs to enable strong modularity. If protocols can't talk to each other, it's going to be hard to keep it clean.

Ivansete-status:
> @Ivansete-status , thoughts? We're going to continue adding to nwaku, the architecture needs to enable strong modularity. If protocols can't talk to each other, it's going to be hard to keep it clean.

I think we can apply observer-observable pattern in the following proc where any module interested in knowing about added peers can be notified: https://github.com/waku-org/nwaku/blob/f08efa477c336931d0f9c2eb1eaa688e179ee645/waku/node/peer_manager/peer_manager.nim#L132

I believe we need to extend the `Capabilities` object so that we can allow future upcoming protocols. In this particular case, we might need to add `WakuRendezvous` as a capability in the ENR: https://github.com/waku-org/nwaku/blob/f08efa477c336931d0f9c2eb1eaa688e179ee645/waku/waku_enr/capabilities.nim#L17

PR #3320, opened 2025-03-11
docs: update prerequisites

jm-clius:
Add `rustc` and `cargo` as prerequisite to README (required for RLN compilation).

I've recently built nwaku on a clean Ubuntu installation and noticed a missing prerequisite in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,13 @@ The standard developer tools, including a C compiler, GNU Make, Bash, and Git. M
 
 > In some distributions (Fedora linux for example), you may need to install `which` utility separately. Nimbus build system is relying on it.
 
+You'll also need an installation of Rust and its toolchain (specifically `rustc` and `cargo`).
+The easiest way to install these, is using `rustup`:
+
+```bash
+curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh
+```
+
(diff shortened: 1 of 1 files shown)


NagyZoltanPeter (commented):
I doubt we need rust and cargo just to build nwaku.
It is needed for the tests that is using anvil which needs rust and cargo to be in place.
But that is covered in the makefile. So maybe we need to update the makefile though.
https://github.com/waku-org/nwaku/blob/6ab92e34f240c7ea7e204123424d430faac483e8/Makefile#L109

Ivansete-status (approved):
LGTM! Thanks for it! :100:

Ivansete-status:
> I doubt we need rust and cargo just to build nwaku. It is needed for the tests that is using anvil which needs rust and cargo to be in place. But that is covered in the makefile. So maybe we need to update the makefile though.
> 
> https://github.com/waku-org/nwaku/blob/6ab92e34f240c7ea7e204123424d430faac483e8/Makefile#L109

I think the need of Rust comes from RLN (zerokit)
The librln library can be either downloaded from _GH_ artfifacts or built locally, if the download failed for some reason.

NagyZoltanPeter (approved):
It is right to mention even if makefile can install it if it missing.

PR #3321, opened 2025-03-11
fix: waku sync mounting

SionoiS:
@jm-clius noticed that Waku sync protocols are not mounted to the switch.

This fixes this problem.

### DIFF
--- a/waku/node/waku_node.nim
+++ b/waku/node/waku_node.nim
@@ -237,12 +237,12 @@ proc mountStoreSync*(
 
   node.wakuStoreReconciliation = recon
 
-  let catchable = catch:
+  let reconMountRes = catch:
     node.switch.mount(
       node.wakuStoreReconciliation, protocolMatcher(WakuReconciliationCodec)
     )
-  if catchable.isErr():
-    return err(catchable.error.msg)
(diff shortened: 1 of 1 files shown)


gabrielmer (approved):
Thanks so much!

jm-clius (approved):
Thanks!

waku/node/waku_node.nim
```diff
@@ -237,12 +237,24 @@ proc mountStoreSync*(
 
   node.wakuStoreReconciliation = recon
 
+  let reconMountRes = catch:
+    node.switch.mount(
+      node.wakuStoreReconciliation, protocolMatcher(WakuReconciliationCodec)
+    )
+  if reconMountRes.isErr():
+    return err(reconMountRes.error.msg)
```
Ivansete-status, line 245: Maybe we could extend the error detail:

"Error mounting WakuReconciliationCodec: " & ...

SionoiS: The caller has the context. https://github.com/waku-org/nwaku/blob/380d2e3382aa2383c924a7f1930393f4451fb374/waku/factory/node_factory.nim#L409

PR #3322, opened 2025-03-12
feat: introduce reputation in PeerManager

s-tikhomirov:
# Description

This PR moves and adapt the logic related to client-side reputation (in the context of Lightpush) to the PeerManager. This logic has been developed in the Lightpush context (see https://github.com/waku-org/nwaku/pull/3293), now it needs to be moved over to peer manager (to avoid tight coupling with Lightpush) and perhaps generalized.

# Changes

- [x] modify `selectPeer` in the peer manager such that, if reputation is enabled, bad-reputation peers are not selected;
- [x] adjust tests to reflect this logic.

### DIFF
--- a/waku/incentivization/reputation_manager.nim
+++ b/waku/incentivization/reputation_manager.nim
@@ -28,6 +28,8 @@ proc getReputation*(manager: ReputationManager, peer: PeerId): Option[bool] =
   else:
     result = none(bool)
 
+### Lightpush-specific functionality ###
+
 # Evaluate the quality of a LightPushResponse by checking its status code
 proc evaluateResponse*(response: LightPushResponse): ResponseQuality =
   if response.statusCode == LightpushStatusCode.SUCCESS.uint32:
@@ -35,7 +37,7 @@ proc evaluateResponse*(response: LightPushResponse): ResponseQuality =
   else:
     return BadResponse
 
-# Update reputation of the peer based on the quality of the response
+# Update reputation of the peer based on LightPushResponse quality
 proc updateReputationFromResponse*(
     manager: var ReputationManager, peer: PeerId, response: LightPushResponse
 ) =
--- a/waku/waku_lightpush/client.nim
+++ b/waku/waku_lightpush/client.nim
@@ -23,9 +23,7 @@ type WakuLightPushClient* = ref object
   publishObservers: seq[PublishObserver]
 
 proc new*(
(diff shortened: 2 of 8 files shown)


waku/incentivization/reputation_manager.nim
```diff
@@ -28,14 +28,16 @@ proc getReputation*(manager: ReputationManager, peer: PeerId): Option[bool] =
...
     result = none(bool)
 
+### Lightpush-specific functionality ###
+
 # Evaluate the quality of a LightPushResponse by checking its status code
 proc evaluateResponse*(response: LightPushResponse): ResponseQuality =
   if response.statusCode == LightpushStatusCode.SUCCESS.uint32:
     return GoodResponse
   else:
     return BadResponse
```
SionoiS, line 38: We'll need a more generic way of doing this, eventually.

waku/waku_lightpush/client.nim
```diff
@@ -57,29 +46,31 @@ proc sendPushRequest(
     buffer = await connection.readLp(DefaultMaxRpcSize.int)
   except LPStreamRemoteClosedError:
     error "Failed to read responose from peer", error = getCurrentExceptionMsg()
-    if wl.reputationManager.isSome:
-      wl.reputationManager.get().setReputation(peer.peerId, some(false))
+    if wl.peerManager.reputationManager.isSome:
+      wl.peerManager.reputationManager.get().setReputation(peer.peerId, some(false))
```
SionoiS, line 50: Fine for now but in the future if we decide to have a standalone `reputationManager` we could use `AsyncQueue` as a communication channel.

SionoiS (commented):
So far so good.

Maybe one more comment about workflow.

Since this work is pretty experimental I would not bother with tests until later. Say you don't like the code ATM and end up changing it 4 times are you also going to change the tests 4 times?

I write tests when the code is in a good shape and all the parts should work in theory.

It's up to you. I know some people swear by test driven development.

waku/incentivization/reputation_manager.nim
s-tikhomirov: Sure. By generic you mean supporting evaluation of responses of other protocols?

SionoiS: > Sure. By generic you mean supporting evaluation of responses of other protocols?

Yes and also avoid importing types from other part of the code.

PR #3326, opened 2025-03-13
fix: waku sync 2.0 codecs ENR support

SionoiS:
# Description
Removes the old waku sync 1.0 codec and replace it with reconciliation in the ENR capability list.

A workaround for this bug https://github.com/waku-org/nwaku/issues/3324


gabrielmer (approved):
Thank you!

jm-clius (approved):

PR #3329, opened 2025-03-17
fix:lightppush v3 not returning relayed peers count

NagyZoltanPeter:
# Description

During presentation prep I noticed with lightpush v3 rest api always returns 0 as relayed peers even if with success.
This should not be the case.

# Changes

- [x] Added missing lightpush response embedding relyed peer count

## How to test

1. send successful lightpush v3 with rest
2. relayPeerCount must be greater than 0


Ivansete-status (approved):
LGTM! Thanks for it! :100:

gabrielmer (approved):
LGTM, thank you!

PR #3331, opened 2025-03-19
chore: enable filter and lightpush REST endpoints without having filternode/lightpushnode configured

NagyZoltanPeter:
# Description

From https://github.com/waku-org/nwaku/issues/3288 and from Build with Waku presentation it is obvious misleading usage of configuration that REST endpoints of lightpush and filter are only available if --lightpushnode/--filternode arguments are defined.

As lightpush/filter clients are always mounted and can handle service peers from other sources there no reason to restrict respective REST endpoints availability connected to filternode and lightpushnode cli arguments -they can accelerate accessing service nodes from the client side. Still, they shall not be mandatory as discovery methods can provide suitable peers.

# Changes

- [X] Removed check for `filternode` before initializing /filter endpoints
- [X] Removed check for `lightpushnode` before initializing /lightpush endpoints

## How to test

1. Start a node with discovery enabled but no relay or other services
1. Issue filter ping with REST. It should not say "REST endpoint not available" but either No service peers or no subscription found.
1. Same with lightpush.

## Issue
https://github.com/waku-org/nwaku/issues/3288

### DIFF
--- a/waku/waku_api/rest/builder.nim
+++ b/waku/waku_api/rest/builder.nim
@@ -148,7 +148,7 @@ proc startRestServerProtocolSupport*(
       "/relay endpoints are not available. Please check your configuration: --relay"
 
   ## Filter REST API
-  if conf.filternode != "" and node.wakuFilterClient != nil:
+  if node.wakuFilterClient != nil:
     let filterCache = MessageCache.init()
 
     let filterDiscoHandler =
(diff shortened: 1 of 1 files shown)


Ivansete-status (approved):
LGTM! Thanks for it! :100: 

Just a question. In the PR's description:

> they can help but not mandatory to access filter and lightush clients

..do you mean "access filter and lightush service nodes" ? ( maybe I'm missing something .)

NagyZoltanPeter:
> LGTM! Thanks for it! 💯
> 
> Just a question. In the PR's description:
> 
> > they can help but not mandatory to access filter and lightush clients
> 
> ..do you mean "access filter and lightush service nodes" ? ( maybe I'm missing something .)

Yes, you're right, I was not precise.
Correctly it would be: `... they can accelerate accessing service nodes from client side, but shall not be mandatory as discovery methods can provide suitable peers.`

PR #3332, opened 2025-03-19
fix: filter out ephemeral msg from waku sync

SionoiS:
## Description
Ephemeral messages were not being filtered out and caused problem when they were not present to be fetched from the db.

Thanks to @Ivansete-status for noticing this bug.

## Issue


Ivansete-status (approved):
LGTM! Thanks for it! :100:

jm-clius (approved):
Ah, happy this was caught. :)

PR #3333, opened 2025-03-19
chore: Make /debug REST endpoints /version and /info available as root

NagyZoltanPeter:
# Description

To make more convenient to use /version and /info REST API endpoints made available right under root instead of /debug/v1/...

Old endpoint are still available but shall be removed after few release as adaptation period.


gabrielmer (approved):
LGTM!

Ivansete-status (approved):
Awesome thanks! The simpler the better! :raised_hands:

PR #3334, opened 2025-03-19
fix: apply latest nph formating

SionoiS:
# Description

Some files had to be updated.

Otherwise CI would complain about formating


gabrielmer (approved):
Thank you!

PR #3337, opened 2025-03-20
fix:  waku sync timing

SionoiS:
# Description
Because of the async nature of the waku sync code, missing messages were sent before the differences could be computed by the other node.

I removed some async parts of the code so that the reconciliation payload is always sent before anything else.

N.B. I also removed the limit on the queues because the async `addLast` was waiting on available space, that is no longer the case and could raise if no space.


jm-clius (approved):

gabrielmer (approved):
LGTM

PR #3341, opened 2025-03-27
fix: docker gives warning while building 

chaitanyaprem:
# Description
when i try to build docker image in my macos, it gives a warning due to incorrect case usage. 
fixed wherever i have noticed.

` => WARN: FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 30)                                                                                       0.0s
`


gabrielmer (approved):
Thank you!

Ivansete-status (approved):
wow that's super picky hahaha
Thanks !

chaitanyaprem:
> wow that's super picky hahaha Thanks !

well, this message confused me while debugging why image generation is failing as this was being shown as last statement when  `make docker-image` fails. So,wanted to address it :)

PR #3346, opened 2025-03-29
feat: introduce `preset` option

fryorcraken:
# Description

Introduce `--preset` configuration to explicitly override a set of config value based on the target Waku network (e.g. TWN, Status, Railgun).
Adding more preset will be done in separate PR.
Overwriting config from cluster-id will be deprecated as a second step.

# Changes

- [x] introduce `--preset` config option
- [x] introduced tests for config matters

# How to test

1. Run `wakunode2` with `--preset=twn` and confirms it can connect to The Waku Network
2. Run `wakunode2` with `--cluster-id=1` and confirms it can **still** connect to The Waku Network (to be deprecated)
3. Run `wakunode2` with Status network parameters and cofirm it can still connect to Status network

## Notes

A redo of #2859 but with a smaller scope.
I added a number of TODO that I will address in follow-up PRs.

### DIFF
--- a/waku/factory/external_config.nim
+++ b/waku/factory/external_config.nim
@@ -148,7 +148,7 @@ type WakuNodeConf* = object
     ## General node config
     preset* {.
       desc:
-        "Network preset to use. 'default' is The RLN-protected Waku Network (cluster 1).",
+        "Network preset to use. 'twn' is The RLN-protected Waku Network (cluster 1).",
       defaultValue: "",
       name: "preset"
     .}: string
--- a/waku/factory/external_config.nim
+++ b/waku/factory/external_config.nim
@@ -76,7 +76,7 @@ type WakuNodeConf* = object
   .}: EthRpcUrl
 
   rlnRelayEthContractAddress* {.
-    desc: "Address of membership contract on an Ethereum testnet",
+    desc: "Address of membership contract on an Ethereum testnet.",
     defaultValue: "",
     name: "rln-relay-eth-contract-address"
   .}: string
@@ -100,6 +100,7 @@ type WakuNodeConf* = object
     name: "rln-relay-eth-private-key"
   .}: string
 
+  # TODO: Remove "Default is" when it's already visible on the CLI
   rlnRelayUserMessageLimit* {.
     desc:
       "Set a user message limit for the rln membership registration. Must be a positive integer. Default is 1.",
@@ -145,6 +146,12 @@ type WakuNodeConf* = object
     .}: seq[ProtectedShard]
 
     ## General node config
+    preset* {.
+      desc: "Network preset to use." & "Must be one of 'default', ''",
+      defaultValue: "",
+      name: "preset"
+    .}: string
+
     clusterId* {.
       desc:
         "Cluster id that the node is running in. Node in a different cluster id is disconnected.",
@@ -276,7 +283,7 @@ hence would have reachability issues.""",
     .}: bool
 
     rlnRelay* {.
-      desc: "Enable spam protection through rln-relay: true|false",
+      desc: "Enable spam protection through rln-relay: true|false.",
       defaultValue: false,
       name: "rln-relay"
     .}: bool
@@ -287,7 +294,7 @@ hence would have reachability issues.""",
     .}: Option[uint]
 
     rlnRelayDynamic* {.
-      desc: "Enable  waku-rln-relay with on-chain dynamic group management: true|false",
+      desc: "Enable  waku-rln-relay with on-chain dynamic group management: true|false.",
       defaultValue: false,
       name: "rln-relay-dynamic"
     .}: bool
@@ -311,7 +318,8 @@ hence would have reachability issues.""",
     .}: string
 
     rlnRelayBandwidthThreshold* {.
-      desc: "Message rate in bytes/sec after which verification of proofs should happen",
+      desc:
+        "Message rate in bytes/sec after which verification of proofs should happen.",
       defaultValue: 0, # to maintain backwards compatibility
       name: "rln-relay-bandwidth-threshold"
     .}: int
@@ -327,6 +335,7 @@ hence would have reachability issues.""",
       name: "keep-alive"
     .}: bool
 
+    # TODO: This is trying to do too much, this should only be used for autosharding, which itself should be configurable
     # If numShardsInNetwork is not set, we use the number of shards configured as numShardsInNetwork
     numShardsInNetwork* {.
       desc: "Number of shards in the network",
@@ -590,7 +599,7 @@ with the drawback of consuming some more bandwidth.""",
 
     ## Discovery v5 config
     discv5Discovery* {.
-      desc: "Enable discovering nodes via Node Discovery v5",
+      desc: "Enable discovering nodes via Node Discovery v5.",
       defaultValue: false,
       name: "discv5-discovery"
     .}: bool
--- a/waku/factory/external_config.nim
+++ b/waku/factory/external_config.nim
@@ -147,7 +147,8 @@ type WakuNodeConf* = object
 
     ## General node config
     preset* {.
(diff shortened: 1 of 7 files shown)


fryorcraken:
Mac test fail is RLN related so I don't think it's me :)


```
  [FAILED ] (  95.9s) clearNullifierLog: should clear epochs > MaxEpochGap[0m
```

Ivansete-status (commented):
Thanks for it! Some comments so far.

In general, I would avoid adding an additional parameter.

I think we need to make sure the CLI passed properly override the implicitly set of parameters.

For example, when passing `rlnRelayEthContractAddress` CLI should override the implicit value that happens when clusterId==1.

waku/factory/node_factory.nim
```diff
@@ -265,6 +258,7 @@ proc setupProtocols(
 
   if conf.numShardsInNetwork == 0:
     warn "Number of shards in network not configured, setting it to",
+      # TODO: If not configured, it mounts 1024 shards! Make it a mandatory configuration instead
```
gabrielmer, line 261: I think we didn't want to make it a mandatory configuration and chose a default value so it works out of the box for everyone and if someone needs a different value they can change it. But I guess if someone doesn't want to use a preset and has to set up everything manually, then it's ok to make them set up the number of shards in the network.

We got to the 1024 default here: https://github.com/waku-org/nwaku/pull/2997#discussion_r1741720714

waku/factory/node_factory.nim
```diff
@@ -482,18 +476,11 @@ proc startNode*(
   return ok()
 
 proc setupNode*(
-    conf: WakuNodeConf, rng: ref HmacDrbgContext = crypto.newRng(), relay: Relay
+    conf: WakuNodeConf,
+    key: crypto.PrivateKey,
```
gabrielmer, line 480: why adding an additional argument for the key if it should be included in `conf`?

waku/factory/waku.nim
```diff
@@ -184,16 +171,16 @@ proc setupAppCallbacks(
   return ok()
 
 proc new*(
-    T: type Waku, confCopy: var WakuNodeConf, appCallbacks: AppCallbacks = nil
+    T: type Waku, srcConf: var WakuNodeConf, appCallbacks: AppCallbacks = nil
```
gabrielmer, line 174: I think that we should use `finalConf` as an argument and not `srcConf`. Notice we have `var WakuNodeConf` so we can modify the passed variable

We do it because if a user sets `conf`, passes it to `new()` and `new()` overrides configurations, whenever the user passes `conf` to other functions it won't be updated with the actual configurations the node is using and then there's bugs that come from that.

So we figured that whenever we create a node and pass a `conf` variable, if we override values we should do it in the passed variable too, so whenever the same variable is used for subsequent function calls it contains the right values

gabrielmer (commented):
Thanks so much! Really like the idea of having presets!

Added a couple comments/questions :))

waku/factory/node_factory.nim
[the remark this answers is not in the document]
fryorcraken: I will do a follow-up PR to clean this TODO.

IMO, the `numShardsInNetwork` should be `0` by default and it should mean that autosharding is NOT enabled.

then, to enable autosharding, this value must be set, and would be the way to enable it.

So that there is no need to come up with some "default" value in case it's not set when we want autosharding.

fryorcraken: See here: https://github.com/waku-org/nwaku/pull/3346#discussion_r2020826347

IMO:

- if set -> autosharding is enabled
- if not set (eg `0` or 1) -> not auto-sharding.

Working "out of the box" -> preset or network with 1 shard (you don't need more than one shard when you start)

waku/factory/waku.nim
fryorcraken: The real issue here is that `WakuNodeConf` is used in too many context. It is not single-purpose.

It is used to grab value from CLI Args. Meaning a bunch of those are `option`s.
It is used to pass the final config to setup a waku node, after those which can't be option have been overridden by `new`.

And then it's used in a bunch of other places as you mentioned.

A better solution would be to have 2 different types for CLI and for actual node configuration. Where the CLI may have `option` but the one for actual node configuration does not -> all values are set.

This way, we can avoid duplicate code like key generation and we can also avoid the bug you just mentioned -> because all those function who need the "final" config, will get an object that does not have `option` that may be `none`.

I'll tackle your comment in this PR to not introduce bugs, and then apply the solution above in a new PR.

waku/factory/node_factory.nim
fryorcraken: See https://github.com/waku-org/nwaku/pull/3346#discussion_r2020590111

waku/factory/external_config.nim
```diff
@@ -145,6 +146,13 @@ type WakuNodeConf* = object
     .}: seq[ProtectedShard]
 
     ## General node config
+    preset* {.
+      desc:
+        "Network preset to use. 'default' is The RLN-protected Waku Network (cluster 1).",
```
Ivansete-status, line 151: Instead of `default` I think it'd better to use `twn` ( more explicit .)

waku/factory/waku.nim
gabrielmer: Yes I agree that our usage of the configuration is a bit messy and we use the same object for everything.

The way I look at it is that we get the initial input from the user and then we progressively process, set and override whatever we need to in the process of actually setting up the node.

If we can just in one place do all the changes from CLI input to actual values to be configured, it would definitely be much cleaner. Not sure if all the transformation logic fits in a single place or some parts of the logic have to happen during different steps of the node configuration, we would need to attempt and see, but definitely worth trying :)

gabrielmer (approved):
LGTM, thanks so much! 😍

fryorcraken:
> In general, I would avoid adding an additional parameter.

I agree, already too many parameters, many not user friendly at all.

The intent is to define more presets once this is introduced.

I can do a follow PR to remove some parameters.

fryorcraken:
> For example, when passing rlnRelayEthContractAddress CLI should override the implicit value that happens when clusterId==1.

The actual problem is that when you pass cluster id 1, there is a surprising side effect of other parameters being overridden.

Having a dedicate option for presets remove the need for the side effect.
It simplifies code and UX.

Ivansete-status (approved):

waku/factory/internal_config.nim
```diff
@@ -157,3 +158,71 @@ proc networkConfiguration*(conf: WakuNodeConf, clientId: string): NetConfigResul
...
+proc applyPresetConfiguration*(srcConf: WakuNodeConf): Result[WakuNodeConf, string] =
+  var resConf = srcConf
+
+  if resConf.clusterId == 1:
+    warn(
+      "Default (The Waku Network) configuration will not be applied when `--cluster-id=1` is passed in future releases. Use `--preset=sandbox` instead."
+    )
+    resConf.preset = "default"
+
+  case resConf.preset
```
Ivansete-status, line 171: We need to make sure is case insensitive.
```suggestion
  case resConf.preset.toLower()
```

waku/factory/external_config.nim
```diff
@@ -145,6 +146,13 @@ type WakuNodeConf* = object
     .}: seq[ProtectedShard]
 
     ## General node config
+    preset* {.
+      desc:
+        "Network preset to use. 'twn' is The RLN-protected Waku Network (cluster 1).",
+      defaultValue: "",
```
Ivansete-status, line 152: Sorry, I missed this detail earlier on.
Do we want `twn` to be config by default or is something we should explicitly configure?

```suggestion
      defaultValue: "twn",
```

fryorcraken: Explicitly configure for now as there are challenges around RLN and need for web3 RPC endpoint.

PR #3348, opened 2025-03-31
chore: waku.nimble force libwaku compile logs with TRACE log level by default

Ivansete-status:
# Description
When working in waku-rust-bindings I've noticed that setting the `logLevel` to `DEBUG` didn't change anything. After applying the suggested change, I could set it to `DEBUG` properly (or `TRACE`.)

## Issue


gabrielmer (approved):
Thank you!

PR #3354, opened 2025-04-01
chore: retrieve protocols in new added peer from discv5

Ivansete-status:
# Description
- Simplify `PeerStore` management. We should only deal with one `PeerStore` instance.
- Remove the type `WakuPeerStore` in `waku/node/peer_manager/waku_peer_store.nim`. Suggest to only use raw `PeerStore` type.
- Properly update the protocols list when a peer is discovered. The previous implementation just overrode the protos list. In this PR, we update the current protocols list with new discovered protocols. ( see  `addPeer` from `peer_manager.nim` .)
- Changes in `tests/waku_discv5/test_waku_discv5.nim`:
  - Activate discv5 unit test list. They were implemented but we didn't run them in our CI.
  - Add new discv5 unittest to validate the discovery of "exotic" protocols such as "peer-exchange" or "rendezvous".
- Add new test in `tests/test_peer_manager.nim`

## Issue

closes
- https://github.com/waku-org/nwaku/issues/3324

### DIFF
--- a/apps/liteprotocoltester/diagnose_connections.nim
+++ b/apps/liteprotocoltester/diagnose_connections.nim
@@ -42,7 +42,7 @@ proc `$`*(cap: Capabilities): string =
 
 proc allPeers(pm: PeerManager): string =
   var allStr: string = ""
-  for idx, peer in pm.wakuPeerStore.peers():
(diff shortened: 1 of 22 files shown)


NagyZoltanPeter (approved):
LGTM!!!! Thank you!

PR #3355, opened 2025-04-02
fix: use counter instead of gauge for metrics that only increase over time

chaitanyaprem:
# Description
Noticed that gauge is being used for metrics that increase over time like number of msgs received/pushed etc.
Aligning them as per definitions https://prometheus.io/docs/concepts/metric_types/


gabrielmer (approved):
Thanks so much!

NagyZoltanPeter (commented):
Will the grafana panels work without change upon this?

chaitanyaprem:
I am not able to understand why lint is failing.
is there something i need to run locally to figure out what is the issue?

NagyZoltanPeter:
> > Will the grafana panels work without change upon this?
> 
> unless they are using some special property assuming gauge they should work the same. Because i had a time-series panel which was using gauge and had no impact when changed to Counter. Any specific issue do you forsee?

I think counters in prometheus will be postfixed by `_total` all those gauge names must be changed in the boards to refer them as <>_total.
But correct me if I'm wrong.

chaitanyaprem:
> I think counters in prometheus will be postfixed by `_total` all those gauge names must be changed in the boards to refer them as <metric_name>_total.

true, that needs to be changed.  are these part of nwaku repo or the compose repo?

waku/waku_archive/archive_metrics.nim
```diff
@@ -3,8 +3,8 @@
 import metrics
 
 declarePublicGauge waku_archive_messages, "number of historical messages", ["type"]
-declarePublicGauge waku_archive_errors, "number of store protocol errors", ["type"]
-declarePublicGauge waku_archive_queries, "number of store queries received"
+declarePublicCounter waku_archive_errors, "number of store protocol errors", ["type"]
+declarePublicCounter waku_archive_queries, "number of store queries received"
```
Ivansete-status, line 7: This is not used so I think it can be removed

waku/waku_archive_legacy/archive_metrics.nim
```diff
@@ -7,9 +7,9 @@ import metrics
 
 declarePublicGauge waku_legacy_archive_messages,
   "number of historical messages", ["type"]
-declarePublicGauge waku_legacy_archive_errors,
+declarePublicCounter waku_legacy_archive_errors,
   "number of store protocol errors", ["type"]
-declarePublicGauge waku_legacy_archive_queries, "number of store queries received"
+declarePublicCounter waku_legacy_archive_queries, "number of store queries received"
```
Ivansete-status, line 12: This is not used either and can also be removed

Ivansete-status (approved):
LGTM! Thanks so much for it! :100:

Ivansete-status:
Great catch @NagyZoltanPeter !
That happens due to the following: https://github.com/status-im/nim-metrics/blob/25ffd054fd774f8cf7935e75d6cad542306d7802/metrics.nim#L520

@chaitanyaprem - I'm afraid we can't apply that change as this may require too much paperwork

NagyZoltanPeter:
> > I think counters in prometheus will be postfixed by `_total` all those gauge names must be changed in the boards to refer them as <metric_name>_total.
> 
> true, that needs to be changed. are these part of nwaku repo or the compose repo?

Well, yes and no. We used to replicate the dashboard json into the nwaku repo, but also used in nwaku-compose and directly in infra grafana.
So the perfect way is to make a fresh copy of dashboard from grafana to nwaku repo and make the changes and copy it back.

chaitanyaprem:
> That happens due to the following: https://github.com/status-im/nim-metrics/blob/25ffd054fd774f8cf7935e75d6cad542306d7802/metrics.nim#L520

oh, interesting thanks for this...i was wondering this is some prometheus thing where it attaches `_total `. 




> @chaitanyaprem - I'm afraid we can't apply that change as this may require too much paperwork

is it because we have to change compose repo and release nwaku and compose together? or do you see some other challenge?

NagyZoltanPeter:
> > That happens due to the following: https://github.com/status-im/nim-metrics/blob/25ffd054fd774f8cf7935e75d6cad542306d7802/metrics.nim#L520
> 
> oh, interesting thanks for this...i was wondering this is some prometheus thing where it attaches `_total `.
> 
> > @chaitanyaprem - I'm afraid we can't apply that change as this may require too much paperwork
> 
> is it because we have to change compose repo and release nwaku and compose together? or do you see some other challenge?

Well the pitty with grafana boards are that they are not deployed with a release but moves completely independent.
So the easiest to do so is with a new release maybe.

PR #3356, opened 2025-04-02
feat: add waku_get_connected_peers_info to libwaku

gabrielmer:
# Description
Adding a `waku_get_connected_peers_info` function to libwaku, which returns a json-string with all the connected nodes peerId's as keys, and an object with an array of their protocols supported and and an array of their multiaddresses as values.

# Changes

- [x] adding `waku_get_connected_peers_info ` function to libwaku

## Issue
#3076

### DIFF
--- a/library/libwaku.nim
+++ b/library/libwaku.nim
@@ -692,6 +692,22 @@ proc waku_get_peerids_from_peerstore(
     userData,
   )
 
+proc waku_get_peerids_and_protocols_from_peerstore(
+    ctx: ptr WakuContext, callback: WakuCallBack, userData: pointer
+): cint {.dynlib, exportc.} =
+  initializeLibrary()
+  checkLibwakuParams(ctx, callback, userData)
+
+  handleRequest(
+    ctx,
+    RequestType.PEER_MANAGER,
+    PeerManagementRequest.createShared(
+      PeerManagementMsgType.GET_ALL_PEER_IDS_AND_PROTOCOLS
+    ),
+    callback,
+    userData,
+  )
+
 proc waku_get_connected_peers(
     ctx: ptr WakuContext, callback: WakuCallBack, userData: pointer
 ): cint {.dynlib, exportc.} =
--- a/library/libwaku.nim
+++ b/library/libwaku.nim
@@ -692,7 +692,7 @@ proc waku_get_peerids_from_peerstore(
     userData,
   )
 
-proc waku_get_all_peer_info(
+proc waku_get_connected_peers_info(
     ctx: ptr WakuContext, callback: WakuCallBack, userData: pointer
 ): cint {.dynlib, exportc.} =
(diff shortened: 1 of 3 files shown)


Ivansete-status (approved):
LGTM! Thanks for it! :100:

library/libwaku.nim
```diff
@@ -650,6 +650,20 @@ proc waku_get_peerids_from_peerstore(
     userData,
   )
 
+proc waku_get_connected_peers_info(
```
NagyZoltanPeter, line 653: Will it be clean for lib users what is the difference btw `waku_get_connected_peers_info` and `waku_relay_get_connected_peers`?
Sounds very similar but will return different set of peers.

library/waku_thread/inter_thread_communication/requests/peer_manager_request.nim
```diff
@@ -83,6 +88,24 @@ proc process*(
...
+      )
+
+    # Build a map of peer IDs to peer info objects
+    for peer in peers:
+      let peerIdStr = $peer.peerId
+      peersMap[peerIdStr] =
+        PeerInfo(protocols: peer.protocols, addresses: peer.addrs.mapIt($it))
+
+    # Convert the map to JSON string
+    let jsonObj = %*peersMap
```
NagyZoltanPeter, line 106: Is it written anywhere what json structure the lib user can expect?

NagyZoltanPeter (approved):
Having two minor questions, but other than that it does what it tells. :-)

library/libwaku.nim
gabrielmer: So the request of `relay` focuses on nodes that are connected to us for the `waku_relay` protocol, or in other words that are in our gossipsub network.

And `waku_get_connected_peers_info ` returns all the peers we're connected to (for any protocol). They sound very similar yes, but the `relay` is actually the key

library/waku_thread/inter_thread_communication/requests/peer_manager_request.nim
gabrielmer: Not really, someone would have to look at the function in order to see the structure. It's the same way for other json strings that we return for Store and for different events.

I agree that for a user it's a bit hard to figure out, I would think that we should eventually standardize everything in libwaku and add proper documentation to all the exposed libwaku functions, although I would personally do it when libwaku's interfaces become stable as currently it moves too fast

library/libwaku.nim
NagyZoltanPeter: I pretty much know this, my question is more the interface naming... that as if I would need to use an API of a lib, would I know it by just getting function names?

How about naming it `waku_get_all_connected_peers_info`... yepp not much better, how to differentiate it is hard.
So yeah, sometime we need to figure it out, how to differentiate lib interface for different use cases...

PR #3358, opened 2025-04-04
chore: better keystore management

Ivansete-status:
# Description
While working with a keystore created by js-waku team, we've realized that the nwaku complained as it wasn't able to access the keystore file. What was happening was that the keystore file couldn't be parsed at all and the reason was that the keystore content should be given in one single line.

This PR simply adds a small protection to make sure the keystore file is always passed in a compact way to the parse phase.

This is motivated by @danisharora099 while working with a tailored keystore file.


gabrielmer (approved):
Niiiice! Thanks so much!

NagyZoltanPeter (approved):
Great, thank you!

PR #3364, opened 2025-04-09
chore: skip two flaky tests

Ivansete-status:
# Description
Simple PR to skip flaky tests because they need some time to run properly. They are useful for local debug operations but annoying in CI.

- `find random peers with instance predicate`
- `clearNullifierLog: should clear epochs > MaxEpochGap`


gabrielmer (approved):
Thank you!

PR #3365, opened 2025-04-10
fix: avoid performing nil check for userData

gabrielmer:
# Description
When using libwaku, from some languages it's not necessary to set `userData` (in C for example).

Removing the null check for userData in libwaku before our event callbacks. The client should be responsible to set `userData` if their language requires it - or in other words, clients such as `waku-go-bindings` and `waku-rust-bindings` should take care of it.

# Changes
- [x] removing null check for `userData` 

## Issue
https://github.com/waku-org/nwaku/issues/3236


Ivansete-status (approved):
LGTM! Thanks for it! :100:

PR #3380, opened 2025-04-17
fix: libwaku's redundant allocs

gabrielmer:
# Description
Removing redundant memory allocations in `libwaku.nim`
The shared memory is always allocated in the corresponding requests' modules

# Changes

- [x] remove memory allocations in `libwaku.nim`

## Issue

#3236

### DIFF
--- a/library/libwaku.nim
+++ b/library/libwaku.nim
@@ -225,18 +225,10 @@ proc waku_content_topic(
   initializeLibrary()
   checkLibwakuParams(ctx, callback, userData)
 
-  let appStr = appName.alloc()
-  let ctnStr = contentTopicName.alloc()
-  let encodingStr = encoding.alloc()
-
-  let contentTopic = fmt"/{$appStr}/{appVersion}/{$ctnStr}/{$encodingStr}"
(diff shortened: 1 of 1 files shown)


NagyZoltanPeter (approved):
Nice!!!!

library/libwaku.nim
```diff
@@ -225,19 +225,11 @@ proc waku_content_topic(
   initializeLibrary()
   checkLibwakuParams(ctx, callback, userData)
 
-  let appStr = appName.alloc()
-  let ctnStr = contentTopicName.alloc()
-  let encodingStr = encoding.alloc()
-
-  let contentTopic = fmt"/{$appStr}/{appVersion}/{$ctnStr}/{$encodingStr}"
+  let contentTopic = fmt"/{$appName}/{appVersion}/{$contentTopicName}/{$encoding}"
```
narimiran, line 228:
```suggestion
  let contentTopic = fmt"/{$appName}/{$appVersion}/{$contentTopicName}/{$encoding}"
```

?

gabrielmer: It was working in the previous version without the `$` but we can add it, might be better than an implicit cast :)

Thanks so much!

PR #3381, opened 2025-04-17
fix: wrong comments and broken link

gabrielmer:
# Description

Removing comments that are wrong (came from a copy-paste mistake), and fixing broken url

# Changes

- [x] removed wrong comments
- [x] fixed broken link 

## Issue
#3236

### DIFF
--- a/library/events/json_message_event.nim
+++ b/library/events/json_message_event.nim
@@ -71,7 +71,7 @@ type JsonMessageEvent* = ref object of JsonEvent
 
 proc new*(T: type JsonMessageEvent, pubSubTopic: string, msg: WakuMessage): T =
   # Returns a WakuMessage event as indicated in
-  # https://rfc.vac.dev/spec/36/#jsonmessageevent-type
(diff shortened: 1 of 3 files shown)


library/events/json_message_event.nim
```diff
@@ -71,7 +71,7 @@ type JsonMessageEvent* = ref object of JsonEvent
 
 proc new*(T: type JsonMessageEvent, pubSubTopic: string, msg: WakuMessage): T =
   # Returns a WakuMessage event as indicated in
-  # https://rfc.vac.dev/spec/36/#jsonmessageevent-type
+  # https://github.com/vacp2p/rfc/blob/master/content/docs/rfcs/36/README.md#jsonmessageevent-type
```
Ivansete-status, line 74: I wonder if we just remove the comment because it will get broken eventually

gabrielmer: I don't think it will get broken because that repo is archived, so it's not undergoing more changes.
We can take it out or leave it, whatever you prefer. I think it's a bit informative

PR #3394, opened 2025-04-24
chore: speed up local docker image builds

NagyZoltanPeter:
# Description
Running `make docker-image` takes a full build inside the docker image and can take quite long. When we test our code inside a dockerized environment it slows us down.
To speed this process up this PR adds new target `docker-quick-image` and `docker-quick-liteprotocoltester`.
This achieved through copy the locally prebuilt image to the final image.

> Notice the built image is a slim debian. 

# Changes

- [X] new make target: `docker-quick-image` to build wakunode2 docker image
- [X] new make target: `docker-quick-liteprotocoltester` to build liteprotocoltester docker image
- [X] fix for building heaptrack enabled wakunode docker image

## How to test

1. run `make docker-quick-image`
2. run `make docker-quick-liteprotocoltester`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,8 +40,8 @@ ifeq ($(detected_OS),Windows)
   NIM_PARAMS += --passL:"-L$(MINGW_PATH)/lib"
   NIM_PARAMS += --passL:"-Lvendor/nim-nat-traversal/vendor/miniupnp/miniupnpc"
   NIM_PARAMS += --passL:"-Lvendor/nim-nat-traversal/vendor/libnatpmp-upstream"
-  
-  LIBS = -static -lws2_32 -lbcrypt -liphlpapi -luserenv -lntdll -lminiupnpc -lnatpmp -lpq 
+
+  LIBS = -static -lws2_32 -lbcrypt -liphlpapi -luserenv -lntdll -lminiupnpc -lnatpmp -lpq
   NIM_PARAMS += $(foreach lib,$(LIBS),--passL:"$(lib)")
 endif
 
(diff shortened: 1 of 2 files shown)


gabrielmer (approved):
Thanks so much!

darshankabariya (approved):
LGTM ! Very helpful it's take more than 10-15 min in my machine.

Makefile
```diff
@@ -340,6 +340,17 @@ docker-image:
 		--target $(TARGET) \
 		--tag $(DOCKER_IMAGE_NAME) .
 
+docker-quick-image: MAKE_TARGET ?= wakunode2
```
Ivansete-status, line 343: Kindly add a comment to explain why `docker-quick-image` is interesting and what makes it different  from the current `docker-image` option.

Makefile
```diff
@@ -83,7 +83,7 @@ HEAPTRACKER_INJECT ?= 0
 ifeq ($(HEAPTRACKER), 1)
 # Needed to make nimbus-build-system use the Nim's 'heaptrack_support' branch
 DOCKER_NIM_COMMIT := NIM_COMMIT=heaptrack_support
-TARGET := prod-with-heaptrack
+TARGET := heaptrack-build
```
Ivansete-status, line 86: I'd suggest to revert that one as it doesn't fix a particular issue.

Ivansete-status (approved):
LGTM! Thanks for it! :100: 
Just some nitpick comments ;P

Makefile
NagyZoltanPeter: @Ivansete-status : Sorry, I just saw your comment... while already merged to master.
But! It actually solves issue with the build. As the correct target was not matched with Dockerfile -> https://github.com/waku-org/nwaku/blob/1ec9b8d96d765cd5063d0e477f36628663f1fd2c/Dockerfile#L67

Update: I find that this change also not correct as the it should be `TARGET := debug`
With having HEAPTRACKER=1 for the make command it will just build the heaptracker image part but not nwaku... I will fix it in separate pr.

Ivansete-status: No worries @NagyZoltanPeter ! I think I just added the review comments one minute after you merged it xD
Thanks for the feedback! :raised_hands:

NagyZoltanPeter: Well anyhow the original target was not working at all... We need an iteration in this as the used target shark be the final desired image phase. Should we distinguish between debug and debug with heaptrack?

Makefile
NagyZoltanPeter: I will add in a separate PR.

PR #3399, opened 2025-05-05
chore: Enhance error msgs libwaku

Ivansete-status:
# Description
Simple PR to standardize the error format returned by libwaku calls.
This PR is also inspired by Zoltán's comments: https://github.com/waku-org/nwaku/pull/3396#discussion_r2072942215

## Issue


library/waku_thread/inter_thread_communication/requests/protocols/relay_request.nim
```diff
@@ -113,28 +113,25 @@ proc process*(
       (kind: SubscriptionKind.PubsubSub, topic: $self.pubsubTopic),
       handler = some(self.relayEventCallback),
     ).isOkOr:
-      let errorMsg = "Subscribe failed:" & $error
-      error "SUBSCRIBE failed", error = errorMsg
-      return err(errorMsg)
+      error "SUBSCRIBE failed", error
```
NagyZoltanPeter, line 116: Well, my original intent was about the comment to avoid such logs like `SUBSCRIBER failed ... error: Subscribe failed: blablabla` in log files.
I'm not sure if the fixed text here and in the other places holds relevant information when we return with the error text?
So `return err($error)` will hold the same necessary information about where the error is coming exactly?

Ivansete-status: > Well, my original intent was about the comment to avoid such logs like `SUBSCRIBER failed ... error: Subscribe failed: blablabla` in log files.

Ah okay, I don't see any harm in that. The more info the better

-----

> I'm not sure if the fixed text here and in the other places holds relevant information when we return with the error text?

Yes, it sounds a bit redundant but I don't see a big issue having it :)

-----

> So `return err($error)` will hold the same necessary information about where the error is coming exactly?

That doesn't contain the precise place where the error was caused but the library consumer already knows which operation was performed and which failed.

@NagyZoltanPeter -  If you consider appropriate, we can add more redundant detail in `return err($error)`. I'm happy with that as well (in which case, I would refactor the PR to make it consistent evrywhere.)

-----
wdyt @gabrielmer @darshankabariya ?

gabrielmer: I personally think that in logs the more detail the better (generally), so I'm personally fine with the current logs.

I also don't have much of an issue with the logs of the style `SUBSCRIBER failed ... error: Subscribe failed: blablabla` because they give relevant information, are easily searchable, and even though it looks redundant, sometimes we don't know how far into the stack we are when something fails or all the possible errors and formats that we can get from function calls downstream, so we end up having some repeated phrases.

Obviously that can be figured out, but it implies an effort that I don't think is worth it. Maybe for really common errors we can have them all nice and cleaned up so it doesn't show redundant parts, but I think that avoiding redundancy in logs everywhere might easily have more cons than pros

gabrielmer (approved):
Looks great :)

Thanks so much!

library/waku_thread/inter_thread_communication/requests/protocols/relay_request.nim
NagyZoltanPeter: Well, my comment was only about the redundancy in logs, I'm not against adding any relevant - yet debug help - info to log lines.

NagyZoltanPeter (approved):
LGTM!

darshankabariya (approved):
LGTM !

PR #3401, opened 2025-05-07
chore: Simple restyle

Ivansete-status:
# Description
Simple changes to keep the code consistent at some extent.


waku/factory/external_config.nim
```diff
@@ -930,7 +930,7 @@ proc toWakuConf*(n: WakuNodeConf): ConfResult[WakuConf] =
   b.withClusterId(n.clusterId)
 
   let clusterConf = toClusterConf(n.preset, some(n.clusterId)).valueOr:
-    return err($error)
+    return err("Error in toWakuConf: " & $error)
```
NagyZoltanPeter, line 933: maybe?
```suggestion
    return err("Error determining cluster from preset: " & $error)
```

NagyZoltanPeter (approved):
LGTM! 
Just one suggestion if you considerr to take...

PR #3402, opened 2025-05-07
chore: allow multiple rln eth clients

Ivansete-status:
## Description

In the RLN context, nwaku needs to interact with Smart Contracts. For that, it uses Eth clients.
This PR gives the user the opportunity to the user to use multiple Eth clients and therefore, give the node more estability.
Notice that having a valid/alive Eth client connection is a strong constraint to RLN and therefore, before this PR, if the connection to the single Eth client failed, the whole node got stopped.

## Issue

closes https://github.com/waku-org/nwaku/issues/3126

### DIFF
--- a/apps/networkmonitor/networkmonitor_config.nim
+++ b/apps/networkmonitor/networkmonitor_config.nim
@@ -8,7 +8,7 @@ import
   stew/shims/net,
   regex
 
-type EthRpcUrl = distinct string
+type EthRpcUrl* = distinct string
 
 type NetworkMonitorConf* = object
   logLevel* {.
@@ -85,7 +85,7 @@ type NetworkMonitorConf* = object
   ethClientUrls* {.
     desc:
       "HTTP address of an Ethereum testnet client e.g., http://localhost:8540/. Argument may be repeated.",
-    defaultValue: "http://localhost:8540/",
+    defaultValue: newSeq[EthRpcUrl](0),
     name: "rln-relay-eth-client-address"
   .}: seq[EthRpcUrl]
 
--- a/apps/networkmonitor/networkmonitor_config.nim
+++ b/apps/networkmonitor/networkmonitor_config.nim
@@ -82,11 +82,12 @@ type NetworkMonitorConf* = object
     name: "rln-relay-tree-path"
   .}: string
 
-  rlnRelayEthClientAddress* {.
-    desc: "HTTP address of an Ethereum testnet client e.g., http://localhost:8540/",
+  ethClientUrls* {.
+    desc:
+      "HTTP address of an Ethereum testnet client e.g., http://localhost:8540/. Argument may be repeated.",
     defaultValue: "http://localhost:8540/",
     name: "rln-relay-eth-client-address"
-  .}: EthRpcUrl
+  .}: seq[EthRpcUrl]
 
   rlnRelayEthContractAddress* {.
     desc: "Address of membership contract on an Ethereum testnet",
--- a/waku/factory/external_config.nim
+++ b/waku/factory/external_config.nim
@@ -78,7 +78,7 @@ type WakuNodeConf* = object
   ethClientUrls* {.
     desc:
       "HTTP address of an Ethereum testnet client e.g., http://localhost:8540/. Argument may be repeated.",
-    defaultValue: newSeq[EthRpcUrl](0),
+    defaultValue: @[EthRpcUrl("http://localhost:8540/")],
     name: "rln-relay-eth-client-address"
   .}: seq[EthRpcUrl]
 
--- a/waku/factory/external_config.nim
+++ b/waku/factory/external_config.nim
@@ -1,5 +1,5 @@
 import
-  std/[strutils, strformat],
+  std/[strutils, strformat, sequtils],
   results,
   chronicles,
   chronos,
@@ -75,7 +75,7 @@ type WakuNodeConf* = object
     name: "rln-relay-cred-path"
   .}: string
 
-  rlnRelayEthClientAddress* {.
+  ethClientUrls* {.
     desc:
       "HTTP address of an Ethereum testnet client e.g., http://localhost:8540/. Argument may be repeated.",
     defaultValue: newSeq[EthRpcUrl](0),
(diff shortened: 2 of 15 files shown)


waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
```diff
@@ -458,11 +458,30 @@ method onRegister*(g: OnchainGroupManager, cb: OnRegisterCallback) {.gcsafe.} =
...
+    for clientUrl in g.ethClientUrls:
+      ## We give a chance to the user to provide multiple clients
+      ## and we try to connect to each of them
+      try:
+        innerEthRpc = await newWeb3(clientUrl)
+        break
+      except CatchableError:
+        error "failed connect Eth client", error = getCurrentExceptionMsg()
+
+    innerEthRpc
```
NagyZoltanPeter, line 477: This looks very strange at first glance. 
maybe `return innerEthRpc`?

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
```diff
@@ -458,11 +458,30 @@ method onRegister*(g: OnchainGroupManager, cb: OnRegisterCallback) {.gcsafe.} =
...
-    await newWeb3(g.ethClientUrl)
+    var innerEthRpc: Web3
+    for clientUrl in g.ethClientUrls:
+      ## We give a chance to the user to provide multiple clients
+      ## and we try to connect to each of them
+      try:
+        innerEthRpc = await newWeb3(clientUrl)
+        break
+      except CatchableError:
+        error "failed connect Eth client", error = getCurrentExceptionMsg()
```
NagyZoltanPeter, line 475: What happens if none of the connection will succeed? Will it return a default web3 object?
Maybe I miss sthg how it returns error...

waku/factory/waku_conf.nim
```diff
@@ -224,8 +224,8 @@ proc validateNoEmptyStrings(wakuConf: WakuConf): Result[void, string] =
 
     if isEmptyOrWhiteSpace(rlnRelayConf.treePath):
       return err("rlnRelayConf.treepath is an empty string")
-    if isEmptyOrWhiteSpace(rlnRelayConf.ethClientAddress):
-      return err("rlnRelayConf.ethClientAddress is an empty string")
+    if rlnRelayConf.ethClientUrls.len == 0:
+      return err("rlnRelayConf.ethClientUrls is empty")
```
NagyZoltanPeter, line 228:
```suggestion
      return err("rlnRelayConf.ethClientUrls must have at least one url set")
```
?

Also it can be overall misleading that we use here internal variable names of the flags and not referring them as appears on the cli. Final such error will be printed to log, operators must look up doc in such case. WDYT?

apps/networkmonitor/networkmonitor_config.nim
```diff
@@ -82,11 +82,12 @@ type NetworkMonitorConf* = object
     name: "rln-relay-tree-path"
   .}: string
 
-  rlnRelayEthClientAddress* {.
-    desc: "HTTP address of an Ethereum testnet client e.g., http://localhost:8540/",
-    defaultValue: "http://localhost:8540/",
+  ethClientUrls* {.
+    desc:
+      "HTTP address of an Ethereum testnet client e.g., http://localhost:8540/. Argument may be repeated.",
+    defaultValue: newSeq[EthRpcUrl](0),
```
NagyZoltanPeter, line 88: Isn't it a breaking change we don't use that default anymore?

NagyZoltanPeter: ... ok forget, I see its in network monitor... so it is not a breaking change...

waku/factory/external_config.nim
```diff
@@ -75,11 +75,12 @@ type WakuNodeConf* = object
     name: "rln-relay-cred-path"
   .}: string
 
-  rlnRelayEthClientAddress* {.
-    desc: "HTTP address of an Ethereum testnet client e.g., http://localhost:8540/",
-    defaultValue: "http://localhost:8540/",
+  ethClientUrls* {.
+    desc:
+      "HTTP address of an Ethereum testnet client e.g., http://localhost:8540/. Argument may be repeated.",
+    defaultValue: newSeq[EthRpcUrl](0),
```
NagyZoltanPeter, line 81: But here, it maybe a bit disturbing having no default as before? WDYT?

NagyZoltanPeter (approved):
It is great overall, thank you! Added some questions though.
Maybe to extend in the future to track bad clients cannot be connected for x times and drop them from the re-trial list, or shuffle the list when making trials.

darshankabariya (approved):

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
Ivansete-status: Good point!
I addressed it in https://github.com/waku-org/nwaku/pull/3402/commits/54d2d59f6498135067251985b001cf534f168873

The way how the error is handled is a bit different from other places:
```nim
template retryWrapper*(
    res: auto,
    retryStrategy: RetryStrategy,
    errStr: string,
    errCallback: OnFatalErrorHandler,
    body: untyped,
): auto =
  if errCallback == nil:
    raise newException(CatchableError, "Ensure that the errCallback is set")
  var retryCount = retryStrategy.retryCount
  var shouldRetry = retryStrategy.shouldRetry
  var exceptionMessage = ""

  while shouldRetry and retryCount > 0:
    try:
      res = body
      shouldRetry = false
    except:
      retryCount -= 1
      exceptionMessage = getCurrentExceptionMsg()
      await sleepAsync(retryStrategy.retryDelay)
  if shouldRetry:
    errCallback(errStr & ": " & exceptionMessage)   <-------  this is called in case of error
    return
  ```

waku/factory/waku_conf.nim
Ivansete-status: Yes, I agree that is misleading but we'd need to make it in a separate PR because this happens to other params too

waku/factory/external_config.nim
Ivansete-status: Great catch! Thanks for it
Addressed in https://github.com/waku-org/nwaku/pull/3402/commits/59fe5c41866402b9b48454146ee5dad4bde9de66

Issue #3400, opened 2025-05-06
bug: Peer connected status not updated properly


NagyZoltanPeter:
### Problem
Working, dialed and connected peers still have NotConnected state in PeerManager.

### Impact
Not possible to query or distinguish between active - connected peers and truly not connected ones.

### To reproduce
1. Use `waku-simulator`
2. checkout branch `hackers-playground`
3. `export NWAKU_IMAGE=harbor.status.im/wakuorg/nwaku:latest`
4. Start with `docker compose up -d`
5. wait till up and running
6. check the peers with `curl -X GET "http://127.0.0.1:8645/admin/v1/peers" -H "accept: application/json" | jq`
7. check the output of `curl -X GET "http://127.0.0.1:8645/admin/v1/peers/connected" -H "accept: application/json" | jq`

Step 6 will print all the peers known by servicenode, but all will have state connectedness = NotConnected
Step 7 will has empty result.

### Expected behavior
Nodes having connected via relay for ex must have status connectedness=Connected

### Screenshots/logs
```
 > curl -X GET "http://127.0.0.1:8645/admin/v1/peers" -H "accept: application/json" | jq
% Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
Dload  Upload   Total   Spent    Left  Speed
100  2225  100  2225    0     0   714k      0 --:--:-- --:--:-- --:--:--  724k
[
{
"multiaddr": "/ip4/10.2.0.15/tcp/60000/p2p/16Uiu2HAkvaLiDzCwSsFZtUXhghMg6wZnrLKh5dvqRmJTXztc5e2h",
"protocols": [
"/ipfs/id/1.0.0",
"/libp2p/autonat/1.0.0",
"/libp2p/circuit/relay/0.2.0/hop",
"/vac/waku/metadata/1.0.0",
"/vac/waku/relay/2.0.0",
"/rendezvous/1.0.0",
"/ipfs/ping/1.0.0",
"/vac/waku/lightpush/3.0.0",
"/vac/waku/lightpush/2.0.0-beta1",
"/vac/waku/filter-push/2.0.0-beta1"
],
"shards": [
0
],
"connected": "NotConnected",
"agent": "",
"origin": "Discv5"
},
{
"multiaddr": "/ip4/10.2.0.17/tcp/60000/p2p/16Uiu2HAm6TuokNudk6MddmHHNihR1fWsJGmLEwbJ3RgDZsvsEqm9",
"protocols": [
[...]

PR #3403, opened 2025-05-08
fix: addPeer could unintentionally override metadata of previously stored peer with defaults and empty

NagyZoltanPeter:
# Description

As found in #3400 connectedness info of the peer is not follows its real status.
During debugging it turned out, that by calling addPeer with a default initialized RemotePeerInfo can clean up metadata previously stored for the peer. 
Usually peer event connected and joined precedes finishing discv5 peer management.
So peer with all info already stored with joined (new peer) event will be wiped out by discv5 adding peer with RemotePeerInfo initialized with only from ENR - yet having all metadata to default.

This explains why connectedness shown always NotConnected but also other issue discovered by LiteProtocolTester, that many times agent information is empty.

# Changes

- [X] PeerStore.addPeer shall only update metadata if not already stored.

## How to test

Follow the reproduction steps in https://github.com/waku-org/nwaku/issues/3400

## Issue
https://github.com/waku-org/nwaku/issues/3400

### DIFF
--- a/waku/node/peer_manager/waku_peer_store.nim
+++ b/waku/node/peer_manager/waku_peer_store.nim
@@ -100,6 +100,7 @@ proc addPeer*(peerStore: PeerStore, peer: RemotePeerInfo, origin = UnknownOrigin
       protos.add($new_proto)
   peerStore[ProtoBook][peer.peerId] = protos
 
+  ## We don't care whether the item was already present in the table or not. Hence, we always discard the hasKeyOrPut's bool returned value
   discard peerStore[AgentBook].book.hasKeyOrPut(peer.peerId, peer.agent)
   discard peerStore[ProtoVersionBook].book.hasKeyOrPut(peer.peerId, peer.protoVersion)
   discard peerStore[KeyBook].book.hasKeyOrPut(peer.peerId, peer.publicKey)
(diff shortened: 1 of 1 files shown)


waku/node/peer_manager/waku_peer_store.nim
```diff
@@ -100,16 +100,22 @@ proc addPeer*(peerStore: PeerStore, peer: RemotePeerInfo, origin = UnknownOrigin
...
-  peerStore[ProtoVersionBook][peer.peerId] = peer.protoVersion
-  peerStore[KeyBook][peer.peerId] = peer.publicKey
-  peerStore[ConnectionBook][peer.peerId] = peer.connectedness
-  peerStore[DisconnectBook][peer.peerId] = peer.disconnectTime
-  peerStore[SourceBook][peer.peerId] =
-    if origin != UnknownOrigin: origin else: peer.origin
-  peerStore[DirectionBook][peer.peerId] = peer.direction
-  peerStore[LastFailedConnBook][peer.peerId] = peer.lastFailedConn
-  peerStore[NumberFailedConnBook][peer.peerId] = peer.numberFailedConn
+  discard peerStore[AgentBook].book.hasKeyOrPut(peer.peerId, peer.agent)
```
Ivansete-status, line 103:
```suggestion
  ## We don't care whether the item was already present in the table or not. Hence, we always discard the hasKeyOrPut's bool returned value
  discard peerStore[AgentBook].book.hasKeyOrPut(peer.peerId, peer.agent)
```

Ivansete-status (approved):
LGTM! Thanks for it! :100: 
I just added a tiny nitpick comment so that the reader doesn't get scared when seeing 
so many `discard` ;P

darshankabariya (approved):
LGTM

PR #3406, opened 2025-05-13
fix: timestamp based validation

### LINKED ISSUE #2972: chore(rln): move from epoch based gap to timestamp based.
### Background
We currently validate that an RLN message epoch is +- a given gap (with respect to the current one)  determined in epochs, [see](https://github.com/waku-org/nwaku/blob/master/waku/waku_rln_relay/rln_relay.nim#L209). This is good since it leaves some margins for clock jitter among nodes, propagation/processing times, etc.

However, when having long epoch sizes (eg several minutes or hours), an old message could be replayed into the future to spam the network. And since the gossipsub window just covers 2 minutes, the network will see this message as new.

To solve this issue, its proposed to:
* Make timestamp mandatory for rln messages
* Include the timestamp in the RLN signal, [here](https://github.com/waku-org/nwaku/blob/master/waku/waku_rln_relay/rln_relay.nim#L292)
* Use this timestamp as the allowed gap. Eg, +-10 seconds.

darshankabariya:
This PR updates the message validation logic from using epoch to timestamp.
The reason for this change is that the current epoch duration is too long, which allows spammers to resend the same message multiple time to bypass gossipsub.

Note:
Message validation is only upgraded to use the timestamp; the nullifier still relies on rlnMaxEpochGap. This means that currently,  the node stores logs of every message to monitor each user’s message rate. But once the epoch gap exceeds 1, those message longer requires storing messages. To prevent excessive storage usage, we’re continuously deleting those messages from the logs.

closes #2972

### DIFF
--- a/tests/waku_rln_relay/test_waku_rln_relay.nim
+++ b/tests/waku_rln_relay/test_waku_rln_relay.nim
@@ -758,7 +758,7 @@ suite "Waku rln relay":
       raiseAssert $error
 
     # usually it's 20 seconds but we set it to 2 for testing purposes which make the test faster
-    wakuRlnRelay.rlnMaxTimestampGap = 2
+    wakuRlnRelay.rlnMaxTimestampGap = 1
 
     var time = epochTime()
 
@@ -775,8 +775,8 @@ suite "Waku rln relay":
     # validate the first message because it's timestamp is the same as the generated timestamp
     let msgValidate1 = wakuRlnRelay.validateMessageAndUpdateLog(wm1)
 
-    # wait for 3 seconds to make the timestamp different from generated timestamp
-    await sleepAsync(3000)
+    # wait for 2 seconds to make the timestamp different from generated timestamp
+    await sleepAsync(2.seconds)
 
     # invalidate the second message because it's timestamp is different from the generated timestamp
     let msgValidate2 = wakuRlnRelay.validateMessageAndUpdateLog(wm2)
--- a/tests/waku_rln_relay/test_waku_rln_relay.nim
+++ b/tests/waku_rln_relay/test_waku_rln_relay.nim
@@ -689,7 +689,7 @@ suite "Waku rln relay":
     # it is a duplicate
     assert isDuplicate3, "duplicate should be found"
 
-  asyncTest "validateMessageAndUpdateLog: again epoch":
+  asyncTest "validateMessageAndUpdateLog: against epoch gap":
     let index = MembershipIndex(5)
 
     let wakuRlnConfig = WakuRlnConfig(
@@ -743,7 +743,7 @@ suite "Waku rln relay":
       msgValidate3 == MessageValidationResult.Valid
       msgValidate4 == MessageValidationResult.Invalid
 
-  asyncTest "validateMessageAndUpdateLog: against timestamp":
+  asyncTest "validateMessageAndUpdateLog: against timestamp gap":
     let index = MembershipIndex(5)
 
     let wakuRlnConfig = WakuRlnConfig(
--- a/tests/waku_rln_relay/test_waku_rln_relay.nim
+++ b/tests/waku_rln_relay/test_waku_rln_relay.nim
@@ -711,7 +711,7 @@ suite "Waku rln relay":
       # another message in the same epoch as wm1, it will break the messaging rate limit
       wm2 = WakuMessage(payload: "Spam message".toBytes(), timestamp: now())
 
-    await sleepAsync(1000)
+    await sleepAsync(1.seconds)
     let time_2 = epochTime()
 
     var
--- a/tests/waku_rln_relay/test_waku_rln_relay.nim
+++ b/tests/waku_rln_relay/test_waku_rln_relay.nim
@@ -1,7 +1,7 @@
 {.used.}
 
 import
-  std/[options, os, sequtils, times, tempfiles],
+  std/[options, os, sequtils, tempfiles],
   stew/byteutils,
   stew/shims/net as stewNet,
   testutils/unittests,
@@ -17,7 +17,10 @@ import
     waku_rln_relay/protocol_metrics,
     waku_keystore,
   ],
-  ./rln/waku_rln_relay_utils
+  ./rln/waku_rln_relay_utils,
+  ../testlib/[wakucore, futures, wakunode, testutils]
+
+from std/times import epochTime
 
 suite "Waku rln relay":
   test "key_gen Nim Wrappers":
@@ -686,7 +689,7 @@ suite "Waku rln relay":
     # it is a duplicate
     assert isDuplicate3, "duplicate should be found"
 
-  asyncTest "validateMessageAndUpdateLog test":
+  asyncTest "validateMessageAndUpdateLog: again epoch":
     let index = MembershipIndex(5)
 
     let wakuRlnConfig = WakuRlnConfig(
@@ -700,27 +703,31 @@ suite "Waku rln relay":
     let wakuRlnRelay = (await WakuRlnRelay.new(wakuRlnConfig)).valueOr:
       raiseAssert $error
 
-    # get the current epoch time
-    let time = epochTime()
+    let time_1 = epochTime()
 
-    #  create some messages from the same peer and append rln proof to them, except wm4
     var
-      wm1 = WakuMessage(payload: "Valid message".toBytes())
+      # create some messages from the same peer and append rln proof to them, except wm4
+      wm1 = WakuMessage(payload: "Valid message".toBytes(), timestamp: now())
       # another message in the same epoch as wm1, it will break the messaging rate limit
-      wm2 = WakuMessage(payload: "Spam".toBytes())
-      #  wm3 points to the next epoch
-      wm3 = WakuMessage(payload: "Valid message".toBytes())
-      wm4 = WakuMessage(payload: "Invalid message".toBytes())
+      wm2 = WakuMessage(payload: "Spam message".toBytes(), timestamp: now())
 
-    wakuRlnRelay.unsafeAppendRLNProof(wm1, time).isOkOr:
+    await sleepAsync(1000)
+    let time_2 = epochTime()
+
+    var
+      #  wm3 points to the next epoch bcz of the sleep 
+      wm3 = WakuMessage(payload: "Valid message".toBytes(), timestamp: now())
+      wm4 = WakuMessage(payload: "Invalid message".toBytes(), timestamp: now())
+
+    wakuRlnRelay.unsafeAppendRLNProof(wm1, time_1).isOkOr:
       raiseAssert $error
-    wakuRlnRelay.unsafeAppendRLNProof(wm2, time).isOkOr:
+    wakuRlnRelay.unsafeAppendRLNProof(wm2, time_1).isOkOr:
       raiseAssert $error
-    wakuRlnRelay.unsafeAppendRLNProof(wm3, time + float64(wakuRlnRelay.rlnEpochSizeSec)).isOkOr:
+
+    wakuRlnRelay.unsafeAppendRLNProof(wm3, time_2).isOkOr:
       raiseAssert $error
 
     # validate messages
-    # validateMessage proc checks the validity of the message fields and adds it to the log (if valid)
     let
       msgValidate1 = wakuRlnRelay.validateMessageAndUpdateLog(wm1)
       # wm2 is published within the same Epoch as wm1 and should be found as spam
@@ -735,6 +742,48 @@ suite "Waku rln relay":
       msgValidate2 == MessageValidationResult.Spam
       msgValidate3 == MessageValidationResult.Valid
       msgValidate4 == MessageValidationResult.Invalid
+
+  asyncTest "validateMessageAndUpdateLog: against timestamp":
+    let index = MembershipIndex(5)
+
+    let wakuRlnConfig = WakuRlnConfig(
+      dynamic: false,
+      credIndex: some(index),
+      userMessageLimit: 10,
+      epochSizeSec: 10,
+      treePath: genTempPath("rln_tree", "waku_rln_relay_2"),
+    )
+
+    let wakuRlnRelay = (await WakuRlnRelay.new(wakuRlnConfig)).valueOr:
+      raiseAssert $error
+
+    # usually it's 20 seconds but we set it to 2 for testing purposes which make the test faster
+    wakuRlnRelay.rlnMaxTimestampGap = 2
+
+    var time = epochTime()
+
+    var
+      wm1 = WakuMessage(payload: "timestamp message".toBytes(), timestamp: now())
+      wm2 = WakuMessage(payload: "timestamp message".toBytes(), timestamp: now())
+
+    wakuRlnRelay.unsafeAppendRLNProof(wm1, time).isOkOr:
+      raiseAssert $error
+
+    wakuRlnRelay.unsafeAppendRLNProof(wm2, time).isOkOr:
+      raiseAssert $error
+
+    # validate the first message because it's timestamp is the same as the generated timestamp
+    let msgValidate1 = wakuRlnRelay.validateMessageAndUpdateLog(wm1)
+
+    # wait for 3 seconds to make the timestamp different from generated timestamp
+    await sleepAsync(3000)
+
+    # invalidate the second message because it's timestamp is different from the generated timestamp
+    let msgValidate2 = wakuRlnRelay.validateMessageAndUpdateLog(wm2)
+
+    check:
+      msgValidate1 == MessageValidationResult.Valid
+      msgValidate2 == MessageValidationResult.Invalid
 
   asyncTest "validateMessageAndUpdateLog: multiple senders with same external nullifier":
     let index1 = MembershipIndex(5)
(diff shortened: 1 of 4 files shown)


jm-clius (changes requested):
Thanks!

At this stage, the verification can be easily fooled by just always adding a fake current `timestamp` to messages but then actually computing RLN proof against any number of very old epochs. In order to use `timestamp` base validation, we need to verify that the `proof` was indeed generated against the `timestamp` included in the message.
In other words:
1. on prover side: `timestamp` should be included in the RLN signal when proofs are generated here: https://github.com/waku-org/nwaku/blob/3464d81a6f83268654242162b348777e028347eb/waku/waku_rln_relay/rln_relay.nim#L311
2. on verifier side: we should verify that the received message `timestamp` was indeed the one used to generate the received proof by including it in the signal for verification here: https://github.com/waku-org/nwaku/blob/3464d81a6f83268654242162b348777e028347eb/waku/waku_rln_relay/rln_relay.nim#L246

At this stage we can trust that the timestamp is the real one used when generating the proof and not just a fake one that makes the message appear "current". However, we still don't know if the nullifier was generated with any old, expired epoch. In other words, we also have to verify that the received `timestamp` belongs to the received `epoch`. This is a simple check which we can do for example here: https://github.com/waku-org/nwaku/blob/3464d81a6f83268654242162b348777e028347eb/waku/waku_rln_relay/rln_relay.nim#L225 before continuing to verify that the `timestamp` is close to current time.

I hope this makes sense? Let me know if I can clarify anything.

darshankabariya:
> At this stage, the verification can be easily fooled by just always adding a fake current `timestamp` to messages while actually computing the RLN proof against any number of very old epochs. In order to use `timestamp` base validation, we need to verify that the `proof` was indeed generated against the `timestamp` included in the message. In other words:

>we still don't know if the nullifier was generated with any old, expired epoch. In other words, we also have to verify that the received timestamp belongs to the received epoch.
  
Thanks for the clarification @jm-clius 
I’ve updated the code to include a timestamp in the signal, as well as on the verifier side. The verification logic now handles epoch validation against the provided timestamp.

FYI: You can ignore the newly added test case for now — it’s still a work in progress and needs to be adjusted to properly test timestamp gaps.

( cc @Ivansete-status @stubbsta )

waku/waku_rln_relay/rln_relay.nim
```diff
@@ -211,25 +213,31 @@ proc validateMessage*(
...
-    # accept messages whose epoch is within +-MaxEpochGap from the current epoch
-    warn "invalid message: epoch gap exceeds a threshold",
-      gap = gap, payloadLen = msg.payload.len, msgEpoch = fromEpoch(proof.epoch)
-    waku_rln_invalid_messages_total.inc(labelValues = ["invalid_epoch"])
+  var timeDiff = uint64(currentTime - msg.timestamp)
+
+  debug "time info", currentTime = currentTime, messageTime = msg.timestamp
+
+  if msg.timestamp > currentTime:
+    warn "invalid message: timestamp is in the future", payloadLen = msg.payload.len
```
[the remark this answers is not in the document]
darshankabariya: Thanks !

waku/waku_rln_relay/rln_relay.nim
```diff
@@ -306,9 +316,11 @@ proc validateMessageAndUpdateLog*(
 proc toRLNSignal*(wakumessage: WakuMessage): seq[byte] =
   ## it is a utility proc that prepares the `data` parameter of the proof generation procedure i.e., `proofGen`  that resides in the current module
   ## it extracts the `contentTopic` and the `payload` of the supplied `wakumessage` and serializes them into a byte sequence
```
jm-clius, line 318: A good idea to update the description to make it clear we now include the `timestamp` in the signal. We may also need to update a spec or two to indicate this, specifically this one: https://rfc.vac.dev/waku/standards/core/17/rln-relay#publishing and perhaps the Waku Network spec too (not requirement for this PR, but in a follow-up)

waku/waku_rln_relay/rln_relay.nim
```diff
@@ -319,6 +331,9 @@ proc appendRLNProof*(
   ## `senderEpochTime` indicates the number of seconds passed since Unix epoch. The fractional part holds sub-seconds.
   ## The `epoch` field of `RateLimitProof` is derived from the provided `senderEpochTime` (using `calcEpoch()`)
 
+  # need to update the timestamp to the sender's timestamp bcz now we are using the sender's timestamp to generate the proof
+  msg.timestamp = senderEpochTime.int64
```
jm-clius, line 335: I am not sure I understand why this is necessary. Shouldn't the `msg.timestamp` here be the sender time in any case? Under what circumstance would `senderEpochTime` be computed on a different timestamp? What does change is that `timestamp` is now a compulsory field and we can return an Error if the message does not include a timestamp.

waku/waku_rln_relay/rln_relay.nim
```diff
@@ -211,25 +213,30 @@ proc validateMessage*(
...
+
+  if timeDiff > rlnPeer.rlnMaxTimestampGap:
+    warn "invalid message: timestamp difference exceeds threshold",
+      timeDiff = timeDiff, maxTimestampGap = rlnPeer.rlnMaxTimestampGap
+    waku_rln_invalid_messages_total.inc(labelValues = ["invalid_timestamp"])
+    return MessageValidationResult.Invalid
+
+  let computedEpoch = rlnPeer.calcEpoch(msg.timestamp.float64)
+  if proof.epoch != computedEpoch:
+    warn "invalid message: epoch mismatch",
```
jm-clius, line 237: I think `epoch mismatch` would be ambiguous - the message is not merely mismatching our epoch, its `timestamp` is mismatching it's own epoch. So I would rather say something like: 
`invalid message: timestamp mismatches epoch`

waku/waku_rln_relay/rln_relay.nim
```diff
@@ -211,25 +213,30 @@ proc validateMessage*(
...
-  trace "epoch info", currentEpoch = fromEpoch(epoch), msgEpoch = fromEpoch(msgEpoch)
-
-  # validate the epoch
-  if gap > rlnPeer.rlnMaxEpochGap:
-    # message's epoch is too old or too ahead
-    # accept messages whose epoch is within +-MaxEpochGap from the current epoch
-    warn "invalid message: epoch gap exceeds a threshold",
-      gap = gap, payloadLen = msg.payload.len, msgEpoch = fromEpoch(proof.epoch)
-    waku_rln_invalid_messages_total.inc(labelValues = ["invalid_epoch"])
+  var timeDiff = uint64(currentTime - msg.timestamp)
```
jm-clius, line 219: This should be an `abs` difference. Timestamps can be `maxTimestampGap` in the future or in the past. For example, our own clock can be slightly behind, making it appear as if new messages are in the future.

waku/waku_rln_relay/rln_relay.nim
```diff
@@ -211,25 +213,30 @@ proc validateMessage*(
...
+
+  debug "time info",
+    currentTime = currentTime, messageTime = msg.timestamp, msgHash = msg.hash
+
+  if msg.timestamp > currentTime:
```
jm-clius, line 224: The only validity requirement is that the `msg.timestamp` should not differ from ours by more than the max allowed gap. This can be both into the future or into the past, to cover normal clock jitter.

waku/waku_rln_relay/constants.nim
```diff
@@ -60058,7 +60058,10 @@ else:
 const StaticGroupMerkleRoot* =
   "2c149e48886b5ba3da2edf8db8d7a364ae7a25618489c04cf0c0380f7cdd4d6f"
 
-const MaxClockGapSeconds* = 20.0 # the maximum clock difference between peers in seconds
+# the maximum clock difference between peers in seconds
+const MaxClockGapSeconds* = 20.0
+# the maximum timestamp difference between the message and the current time
+const MaxTimestampGapSeconds* = 20.0
```
jm-clius, line 60064: I don't think we need both of these? We only need `MaxTimestampGapSeconds` afaics?

jm-clius (changes requested):
Almost there! Couple of comments below. LMK if anything is unclear.

waku/waku_rln_relay/rln_relay.nim
darshankabariya: Addressed !!

waku/waku_rln_relay/rln_relay.nim
darshankabariya: Thanks ! There are some test cases that initialize WakuMessage and generate a proof at different times. Initially, I handled that logic here, but I now realize that approach was overfitting. Instead of modifying it here, I’ve updated the relevant test cases instead.

waku/waku_rln_relay/rln_relay.nim
darshankabariya: Got it, Addressed !!

stubbsta (changes requested):
Thanks Darshan, this looks good. Is there a test to check that a message with an invalid timestamp correctly gets identified and rejected? If not, would it be possible to add one?

waku/waku_rln_relay/rln_relay.nim
```diff
@@ -93,6 +93,7 @@ type WakuRLNRelay* = ref object of RootObj
   nullifierLog*: OrderedTable[Epoch, Table[Nullifier, ProofMetadata]]
   lastEpoch*: Epoch # the epoch of the last published rln message
   rlnEpochSizeSec*: uint64
+  rlnMaxTimestampGap*: uint64
   rlnMaxEpochGap*: uint64
```
jm-clius, line 97: Can we remove maxEpochGap? It can be in a next PR or announced as a deprecation, since it affects configuration.

waku/waku_rln_relay/rln_relay.nim
```diff
@@ -211,25 +213,25 @@ proc validateMessage*(
...
-  trace "epoch info", currentEpoch = fromEpoch(epoch), msgEpoch = fromEpoch(msgEpoch)
-
-  # validate the epoch
-  if gap > rlnPeer.rlnMaxEpochGap:
-    # message's epoch is too old or too ahead
-    # accept messages whose epoch is within +-MaxEpochGap from the current epoch
-    warn "invalid message: epoch gap exceeds a threshold",
-      gap = gap, payloadLen = msg.payload.len, msgEpoch = fromEpoch(proof.epoch)
-    waku_rln_invalid_messages_total.inc(labelValues = ["invalid_epoch"])
+  var timeDiff = uint64(abs(currentTime - msg.timestamp))
```
jm-clius, line 219: Why is this a `var`and not a `let`?

jm-clius (approved):
Thanks for this! The minor comments below are non-blocking (i.e. you don't need to request a review again before merging :) ).
Can we create two spin-off issues from this and link to this one?
1. Issue to change the necessary RLN-RELAY and NETWORK specs (in waku-org/specs repo).
2. Issue to deprecate rlnMaxEpochGap as a config item and remove it from the code

waku/waku_rln_relay/rln_relay.nim
darshankabariya: Hi @jm-clius, I believe it’s not the right time to deprecate rlnMaxEpochGap yet, as clearNullifierLog still depends on it.

darshankabariya:
> Thanks Darshan, this looks good. Is there a test to check that a message with an invalid timestamp correctly gets identified and rejected? If not, would it be possible to add one?

Hi @stubbsta , I have added a test case that covers both valid and invalid scenarios.

tests/testlib/wakucore.nim
```diff
@@ -48,7 +48,7 @@ proc fakeWakuMessage*(
     payload: string | seq[byte] = "TEST-PAYLOAD",
     contentTopic = DefaultContentTopic,
     meta: string | seq[byte] = newSeq[byte](),
-    ts = now(),
+    ts = Timestamp(epochTime()),
```
Ivansete-status, line 51: This change is not needed. The `WakuMessageTimestamp` may contain the unix epoch time in nanoseconds. This change gives a second resolution. Although not strictly needed, I think is interesting to keep consistency

tests/waku_rln_relay/test_waku_rln_relay.nim
```diff
@@ -615,7 +616,7 @@ suite "Waku rln relay":
       absDiff(epoch1, epoch2) == uint64(1)
       absDiff(epoch2, epoch1) == uint64(1)
 
-  test "updateLog and hasDuplicate tests":
+  xasyncTest "updateLog and hasDuplicate xasyncTests":
```
Ivansete-status, line 619: Kindly recover this test back :)

tests/waku_rln_relay/test_waku_rln_relay.nim
```diff
@@ -700,27 +701,34 @@ suite "Waku rln relay":
...
+        WakuMessage(payload: "Valid message".toBytes(), timestamp: Timestamp(time_1))
       # another message in the same epoch as wm1, it will break the messaging rate limit
-      wm2 = WakuMessage(payload: "Spam".toBytes())
-      #  wm3 points to the next epoch
-      wm3 = WakuMessage(payload: "Valid message".toBytes())
-      wm4 = WakuMessage(payload: "Invalid message".toBytes())
+      wm2 = WakuMessage(payload: "Spam".toBytes(), timestamp: Timestamp(time_1))
 
-    wakuRlnRelay.unsafeAppendRLNProof(wm1, time).isOkOr:
+    await sleepAsync(1000)
```
Ivansete-status, line 713: I wonder if we could make the test faster by artificially creating the `time_1` in the past, such as: `let time_1 = epochTime() - 1000`
```suggestion
```

tests/waku_rln_relay/test_waku_rln_relay.nim
```diff
@@ -736,6 +744,48 @@ suite "Waku rln relay":
...
+      raiseAssert $error
+
+    wakuRlnRelay.unsafeAppendRLNProof(wm2, time).isOkOr:
+      raiseAssert $error
+
+    # validate the first message because it's timestamp is the same as the generated timestamp
+    let msgValidate1 = wakuRlnRelay.validateMessageAndUpdateLog(wm1)
+
+    # wait for 3 seconds to make the timestamp different from generated timestamp
+    await sleepAsync(3000)
```
Ivansete-status, line 780: Similarly to the previous comment , let's make this test faster , if possible ;P
```suggestion
```

tests/wakunode_rest/test_rest_relay.nim
```diff
@@ -260,7 +261,7 @@ suite "Waku v2 Rest API - Relay":
       RelayWakuMessage(
         payload: base64.encode("TEST-PAYLOAD"),
         contentTopic: some(DefaultContentTopic),
-        timestamp: some(int64(2022)),
+        timestamp: some(Timestamp(getTime().toUnixFloat())),
```
Ivansete-status, line 264: This is our idiomatic way to do that:
```suggestion
        timestamp: some(getNowInNanosecondTime()),
```
That suggestion applies elsewhere in this file

waku/waku_rln_relay/rln_relay.nim
```diff
@@ -211,25 +213,25 @@ proc validateMessage*(
   # track message count for metrics
   waku_rln_messages_total.inc()
 
-  # checks if the `msg`'s epoch is far from the current epoch
-  # it corresponds to the validation of rln external nullifier
-  # get current rln epoch
-  let epoch: Epoch = rlnPeer.getCurrentEpoch()
+  # checks if the message's timestamp is within acceptable range
+  let currentTime = Timestamp(getTime().toUnixFloat())
```
Ivansete-status, line 217:
```suggestion
  let currentTime = getNowInNanosecondTime()
```

waku/waku_rln_relay/rln_relay.nim
```diff
@@ -211,25 +213,25 @@ proc validateMessage*(
...
-  if gap > rlnPeer.rlnMaxEpochGap:
-    # message's epoch is too old or too ahead
-    # accept messages whose epoch is within +-MaxEpochGap from the current epoch
-    warn "invalid message: epoch gap exceeds a threshold",
-      gap = gap, payloadLen = msg.payload.len, msgEpoch = fromEpoch(proof.epoch)
-    waku_rln_invalid_messages_total.inc(labelValues = ["invalid_epoch"])
+  let timeDiff = uint64(abs(currentTime - msg.timestamp))
+
+  debug "time info",
+    currentTime = currentTime, messageTime = msg.timestamp, msgHash = msg.hash
```
Ivansete-status, line 222: For analogy with other code:
```suggestion
    currentTime = currentTime, messageTime = msg.timestamp, msg_hash = msg.hash
```

Ivansete-status (commented):
Thanks for it :raised_hands: ! I'm adding some nitpick comments that I hope you find interesting

waku/waku_rln_relay/rln_relay.nim
darshankabariya: Thanks @Ivansete-status, I’ve addressed all the suggestions. I’m now using nanosecond format in the message.

tests/waku_rln_relay/test_waku_rln_relay.nim
darshankabariya: Hi @Ivansete-status,
This isn’t possible with the current code because the validateMessage function uses the current clock time. The inject time doesn’t help in creating an artificial case. So, the machine needs to wait at least one second to test this properly. I think that should be okay, right? 
Screenshot 2025-05-26 at 02 57 54

tests/waku_rln_relay/test_waku_rln_relay.nim
darshankabariya: I’ve make bit faster from 3 seconds to 2 seconds. To reduce it further, we’d need to make rlnMaxTimestamp a float. I think it’s fine for now, but I can optimize it more if you suggest.

Ivansete-status (approved):
LGTM! Thanks for it! :100:

stubbsta (approved):
LGTM. Just some nitpicking about test naming, thanks Darshan!

PR #3419, opened 2025-05-21
feat: Extend ndoe /health REST endpoint with all protocol's state

NagyZoltanPeter:
# Description

Long time ago it was planned to have a health report over REST API that contains consistent state of a node overall and per protocol. So far it only considered RlnRelay (checking if it is syncing or not).

This PR intend to extend /health report to all relevant protocol Waku supports.

# Changes

Added report of 
- Relay
    - READY if mounted and node has Relay peers 
- Lightpush
   - READY if mounted and node has Relay peers 
- Lightpush legacy
    - READY if mounted and node has Relay peers 
- Store
- Store legacy
- Filter
- PeerExchange
- Rendezvous
    - READY if mounted and there are available peer supports Rendezvous protocol 

## How to test

1. run in nwaku-compose or standalone
2. run `curl -X GET http://127.0.0.1:8645/health | jq`
3. check result
4. repeat for some time and check if READY state changes of the protocols (Relay/RlnRelay/Lightpush)

#### Expect

```
{
  "nodeHealth": "Ready",
  "protocolsHealth": [
    {
      "Relay": "Not Ready"
    },
    {
      "Rln Relay": "Ready"
    },
    {
      "Lightpush v3": "Not Ready"
    },
    {
      "Lightpush Legacy": "Not Ready"
    },
    {
      "Filter": "Ready"
    },
    {
      "Store": "Not Mounted"
    },
    {
      "Legacy Store": "Not Mounted"
    },
    {
      "Peer Exchange": "Not Mounted"
    },
    {
      "Rendezvous": "Ready"
    }
  ]
}
```

## Issue
https://github.com/waku-org/nwaku/issues/3409

## More?

WDYT: can we add more sophisticated info about Store protocol? Can we add a way to check the underlying DB state?
Shall we consider Relay state for Filter also?

### DIFF
--- a/waku/node/health_monitor.nim
+++ b/waku/node/health_monitor.nim
@@ -1,8 +1,8 @@
 {.push raises: [].}
 
-import std/[options], chronos
+import std/[options, sets], chronos
 
-import waku_node, ../waku_rln_relay
+import waku_node, ../waku_rln_relay, ../waku_relay
 
 type
   HealthStatus* = enum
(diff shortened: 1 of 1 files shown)


Ivansete-status:
Thanks so much for taking care of that :partying_face: !

> WDYT: can we add more sophisticated info about Store protocol? Can we add a way to check the underlying DB state?

Maybe we can perform a simple query to the database, i.e. `messages exists`, and, if the query works, the consider `Ready`

> Shall we consider Relay state for Filter also?
Yes, I think that makes sense.

------

Besides, something interesting to consider is to retrieve a field, named `reason`, that explains the detail why a certain protocol is not considered `Ready`. And, if any protocol is `Ready` then not return that `reason` field.

NagyZoltanPeter:
> Thanks so much for taking care of that 🥳 !
> 
> > WDYT: can we add more sophisticated info about Store protocol? Can we add a way to check the underlying DB state?
> 
> Maybe we can perform a simple query to the database, i.e. `messages exists`, and, if the query works, the consider `Ready`

For now I found it heavy and considering enough to suppose that if store is set up for a node it would not start if DB is not working behind. Issueing query need to be designed more carefully in consideration what we mean by READY or NOT_READY for store service.
 
> > Shall we consider Relay state for Filter also?
> 
> Yes, I think that makes sense.

Added! Thank you. 

> Besides, something interesting to consider is to retrieve a field, named `reason`, that explains the detail why a certain protocol is not considered `Ready`. And, if any protocol is `Ready` then not return that `reason` field.

Hm... For now I feel it more part of the documentation than part of response. But definitely considerable!

waku/node/health_monitor.nim
```diff
@@ -56,29 +56,220 @@ proc init*(
 
 const FutIsReadyTimout = 5.seconds
 
-proc getNodeHealthReport*(hm: WakuNodeHealthMonitor): Future[HealthReport] {.async.} =
-  result.nodeHealth = hm.nodeHealth
+proc getRelayHealth(hm: WakuNodeHealthMonitor): Future[HealthStatus] {.async.} =
```
Ivansete-status, line 59: I think it is very important to give the precise feedback why we consider a particular protocol as `NOT_READY` :)
```suggestion
proc getRelayHealth(hm: WakuNodeHealthMonitor): Future[Result[HealthStatus, string]] {.async.} =
  ## Returns an error, with its description, when the relay protocol is not ready
```

waku/node/health_monitor.nim
```diff
@@ -56,29 +56,220 @@ proc init*(
...
+  if hm.node.get().wakuRelay == nil:
+    return HealthStatus.NOT_MOUNTED
```
Ivansete-status, line 61:
```suggestion
    return ok(HealthStatus.NOT_MOUNTED)
```

waku/node/health_monitor.nim
```diff
@@ -56,29 +56,220 @@ proc init*(
...
+
+  let relayPeers = hm.node
+    .get().wakuRelay
+    .getConnectedPubSubPeers(pubsubTopic = "").valueOr:
+      return HealthStatus.NOT_READY
```
Ivansete-status, line 66:
```suggestion
      return err("relay is mounted but error happened when retrieving peers: " & $error)
```

waku/node/health_monitor.nim
```diff
@@ -56,29 +56,220 @@ proc init*(
...
+
+  if relayPeers.len() == 0:
+    return HealthStatus.NOT_READY
```
Ivansete-status, line 69:
```suggestion
    return err("relay mounted but the number connected relay peers is zero")
```

Ivansete-status (commented):
Thanks for the recent commits!
I'm adding some comments which I think are interesting for future issue debug.
The suggested idea is to return a `Result` , which will be `ok` when either `IS_READY` or `NOT_MOUNTED`, and `err` when is mounted but a certain condition prevents the protocol from being ready, in which case, we return the detailed description why we consider it is not ready.

waku/node/health_monitor.nim
NagyZoltanPeter: I accept this, but given that NOT_READY is not an error state, I would rather extend HealthStatus with an optional description, instead of translating it as an error.
NOT_READY is just a particular state in time. Like at startup node is SYNCING or yet discovering peers to connect with - which is likely to change the next time you check /health

s-tikhomirov:
> Lightpush
> READY if mounted and node has Relay peers

Does it not check whether any of the Relay peers actually support Lightpush? Am I correct that this is not a given, generally?

The same question may apply to other light protocols; I'm asking about Lightpush just because I'm more familiar with it.

NagyZoltanPeter:
> > Lightpush
> > READY if mounted and node has Relay peers
> 
> Does it not check whether any of the Relay peers actually support Lightpush? Am I correct that this is not a given, generally?
> 
> The same question may apply to other light protocols; I'm asking about Lightpush just because I'm more familiar with it.

What exactly do you mean `relay peers actually support Lightush`?
For /health report we check available relay peers (READYness of relay protocol) due lightpush and filter depends on them.
Means if there are no available relay peers you lightpush will not be able to publish any message to the network.
Same for filter, it will not be able to filter and push messages from the relay network.

Of course this check a bit simplistic as for a particular lightpush request there shall be available relay peers supports the certain shard (pubsub-topic). In our current networks normally all relay peers support all shards.

s-tikhomirov:
> What exactly do you mean relay peers actually support Lightush?

Sorry, I was confused by the fact that Lightpush-client functionality is always enabled even on nodes started without `--lightpush`. Nevermind then!

NagyZoltanPeter:
@Ivansete-status : Added description in result for NOT_READY state, explaining the result per protocol.

waku/node/health_monitor.nim
```diff
@@ -16,6 +16,7 @@ type
   ProtocolHealth* = object
     protocol*: string
     health*: HealthStatus
+    desc*: Option[string]
```
Ivansete-status: Maybe?
```suggestion
    desc*: Option[string] ## describes why a certain protocol is considered `NOT_READY`
```

waku/node/health_monitor.nim
```diff
@@ -54,31 +55,234 @@ proc init*(
   else:
     raise newException(ValueError, "Invalid HealthStatus string representation")
 
+proc init*(p: typedesc[ProtocolHealth], protocol: string): ProtocolHealth =
+  result.protocol = protocol
+  result.health = HealthStatus.NOT_MOUNTED
+  result.desc = none[string]()
+  return result
```
Ivansete-status: We should avoid the use of implicit `result` variable. See https://status-im.github.io/nim-style-guide/language.result.html

tests/wakunode_rest/test_rest_health.nim
```diff
@@ -85,9 +85,14 @@ suite "Waku v2 REST API - health":
...
-      response.data.protocolsHealth[0].protocol == "Rln Relay"
-      response.data.protocolsHealth[0].health == HealthStatus.READY
+      response.data.protocolsHealth.len() == 14
+      response.data.protocolsHealth[0].protocol == "Relay"
+      response.data.protocolsHealth[0].health == HealthStatus.NOT_READY
+      response.data.protocolsHealth[0].desc == some("No connected peers")
+      response.data.protocolsHealth[1].protocol == "Rln Relay"
+      response.data.protocolsHealth[1].health == HealthStatus.READY
+      response.data.protocolsHealth[2].protocol == "Lightpush"
+      response.data.protocolsHealth[2].health == HealthStatus.NOT_MOUNTED
```
Ivansete-status, line 99: Even though being tedious , I wonder if it worth adding all the checks

waku/node/health_monitor.nim
```diff
@@ -54,31 +55,235 @@ proc init*(
...
+      return p.notMounted()
+
+  if relayPeers.len() == 0:
+    return p.notReady("No connected peers")
+
+  return p.ready()
+
+proc getRlnRelayHealth(hm: WakuNodeHealthMonitor): Future[ProtocolHealth] {.async.} =
+  var p = ProtocolHealth.init("Rln Relay")
+  if hm.node.get().wakuRlnRelay == nil:
```
Ivansete-status: Super nitpick comment. I think we tend to use `isNil` more often:
Applies elsewhere.
```suggestion
  if hm.node.get().wakuRlnRelay.isNil():
```

waku/node/health_monitor.nim
```diff
@@ -54,31 +55,235 @@ proc init*(
...
+  try:
+    if not isReadyStateFut.completed():
+      return p.notReady("Ready state check timed out")
+    elif isReadyStateFut.read():
+      return p.ready()
+
+    return p.synchronizing()
+  except:
+    error "exception reading state: " & getCurrentExceptionMsg()
+    return p.notReady("State cannot be determined")
```
Ivansete-status, line 129:
```suggestion
    return p.notReady("State cannot be determined: " & getCurrentExceptionMsg())
```

waku/node/health_monitor.nim
```diff
@@ -54,31 +55,235 @@ proc init*(
...
+    result.protocolsHealth.add(hm.getLegacyStoreHealth())
+    result.protocolsHealth.add(hm.getPeerExchangeHealth())
+    result.protocolsHealth.add(hm.getRendezvousHealth())
+
+    result.protocolsHealth.add(hm.getLightpushClientHealth(relayHealth.health))
+    result.protocolsHealth.add(hm.getLegacyLightpushClientHealth(relayHealth.health))
+    result.protocolsHealth.add(hm.getStoreClientHealth())
+    result.protocolsHealth.add(hm.getLegacyStoreClientHealth())
+    result.protocolsHealth.add(hm.getFilterClientHealth(relayHealth.health))
+  return result
```
Ivansete-status: As mentioned in other comment. I think is better to avoid implicit `result` ( also , for consistency with other nwaku code .)

waku/waku_api/rest/health/types.nim
```diff
@@ -10,26 +10,34 @@ proc writeValue*(
...
-    try:
-      health = HealthStatus.init(fieldValue)
-    except ValueError:
-      reader.raiseUnexpectedValue("Invalid `health` value")
+      let fieldValue = reader.readValue(string)
+      try:
+        health = some(HealthStatus.init(fieldValue))
+        protocol = some(fieldName)
+      except ValueError:
+        reader.raiseUnexpectedValue("Invalid `health` value")
```
Ivansete-status:
```suggestion
        reader.raiseUnexpectedValue("Invalid `health` value: " & getCurrentExceptionMsg())
```

Ivansete-status (approved):
LGTM! Thanks for the master piece PR! :partying_face: 
I just added some nitpick comments that I hope you find useful

gabrielmer (approved):
Looks amazing! Thanks so much!

waku/node/health_monitor.nim
NagyZoltanPeter: Hm, I would argue with how useful it is. As the endpoint is not part of /admin endpoint group it has just the role to give status about protocols at any time. The status can change and will change in time as repetitive queries done. If there is something realy wrong with rln-relay, it's not the way to find out with /health endpoint, imo.

Reading direct and full logs is the way... as probably the most info coming there upon error.

PR #3420, opened 2025-05-22
feat: nix libwaku output package in .aar archive

markoburcul:
# Description

This change is only modifying the way libwaku is packaged for Android in the Nix flake. For us is beneficial so later on when we reference it in status-go and status-mobile it is already an `.aar` archive.
# Changes
- output format of the libwaku-android-arm64 nix package

### DIFF
--- a/nix/default.nix
+++ b/nix/default.nix
@@ -10,6 +10,7 @@
     "x86_64-linux" "aarch64-linux"
   ],
   androidArch,
+  abidir,
   zerokitPkg,
 }:
 
@@ -32,6 +33,7 @@ in stdenv.mkDerivation rec {
   buildInputs = with pkgs; [
     openssl
     gmp
+    zip
   ];
 
   # Dependencies that should only exist in the build environment.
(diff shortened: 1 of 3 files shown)


nix/default.nix
```diff
@@ -98,8 +100,10 @@ in stdenv.mkDerivation rec {
   '';
 
   installPhase = ''
-    mkdir -p $out/build/android
-    cp -r ./build/android/* $out/build/android/
+    mkdir -p $out/jni
+    cp -r ./build/android/${abidir}/* $out/jni/
+    echo '<manifest xmlns:android="http://schemas.android.com/apk/res/android" package="com.example.mylibrary" />' > $out/jni/AndroidManifest.xml
```
jakubgs, line 105: Looks like a place for a variable.
```suggestion
    echo '${androidManifest}' > $out/jni/AndroidManifest.xml
```

jakubgs (approved):

jakubgs (approved):

PR #3422, opened 2025-05-25
fix: relay unsubscribe

Ivansete-status:
## Description
There's been detected issues in `waku-interop-tests` due to ineffective `unsubscribe` operations.
This PR aims to enhance the `WakuRelay`'s `unsubscribe` proc.

## Changes
- `mountRelay` no longer internally subscribes to a set of shards. The subscription should be made explicitly by calling `node.subscribe`.
- `mountRelay` no longer receives an array of shards as a parameter. The reason for that is that when subscribing, we always need to set a `WakuRelayHandler` proc.
- `waku_node::subscribe` proc no longer returns a proc. This is not needed.
- Keeping track of the `TopicHandler` linked to any subscription. This is needed to make effective the `unsubscribe` within `PubSub` (inside nim-libp2p.)
- In `unsubscribe` operation, properly remove the topic validator instance which was added in `subscribe`.
- `WakuRelay` only uses one single message handler. See `registerRelayHandler` in `waku/node/waku_node.nim`:
  ```nim
    let uniqueTopicHandler = proc(
      topic: PubsubTopic, msg: WakuMessage
  ): Future[void] {.async, gcsafe.} =
    await traceHandler(topic, msg)
    await filterHandler(topic, msg)
    await archiveHandler(topic, msg)
    await syncHandler(topic, msg)
    await appHandler(topic, msg)

  node.wakuRelay.subscribe(topic, uniqueTopicHandler)
  ```

## Issue

### DIFF
--- a/tests/wakunode2/test_validators.nim
+++ b/tests/wakunode2/test_validators.nim
@@ -73,7 +73,9 @@ suite "WakuNode2 - Validators":
 
     # Subscribe all nodes to the same topic/handler
     for node in nodes:
-      discard node.wakuRelay.subscribe($spamProtectedShard, handler)
+      node.subscribe((kind: PubsubSub, topic: $spamProtectedShard), handler).isOkOr:
+        assert false, "Failed to subscribe to topic: " & $error
+
     await sleepAsync(500.millis)
 
     # Each node publishes 10 signed messages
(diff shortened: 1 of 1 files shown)


gabrielmer (approved):
Wooow beautiful 😍 much simpler this way!

Thanks so much!

NagyZoltanPeter (approved):
Great work! LGTM!

PR #3430, opened 2025-05-28
feat: add shard info metrics

### LINKED ISSUE #3306: feat: Expand Waku Promotheus metric
### Problem
Need for a "shard" label in Waku Prometheus metrics for visualization purposes.

### Suggested solution

I will need a “shard” label for the following Prometheus metrics:

waku_discv5_discovered
waku_connected_peer
waku_rln_valid_messages_total_count
waku_histogram_message_size_sum
waku_histogram_message_size_count
waku_archive_messages

The waku_relay_network_bytes_total metric has a "topic" label that provides the shard (e.g., topic="/waku/2/rs/1/0"), which is what I would need for the above metrics.

### Additional context
[Bi issue related](https://github.com/status-im/data-docs/issues/3)

darshankabariya:
This PR will add shard into few metrics.

close #3306

### DIFF
--- a/waku/discovery/waku_discv5.nim
+++ b/waku/discovery/waku_discv5.nim
@@ -14,7 +14,8 @@ import ../node/peer_manager/peer_manager, ../waku_core, ../waku_enr
 
 export protocol, waku_enr
 
-declarePublicGauge waku_discv5_discovered, "number of nodes discovered"
+declarePublicGauge waku_discv5_discovered,
+  "number of nodes discovered", labels = ["shard"]
 declarePublicGauge waku_discv5_errors, "number of waku discv5 errors", ["type"]
 
 logScope:
@@ -232,7 +233,18 @@ proc findRandomPeers*(
   elif wd.predicate.isSome():
     discoveredRecords = discoveredRecords.filter(wd.predicate.get())
 
-  waku_discv5_discovered.inc(discoveredRecords.len)
+  # Increment metric for each discovered record's shards
+  for record in discoveredRecords:
+    let typedRecord = record.toTyped().valueOr:
+      # If we can't parse the record, skip it
+      continue
+
+    let relayShards = typedRecord.relaySharding().valueOr:
+      # If no relay sharding info, skip it
+      continue
+
+    for shardId in relayShards.shardIds:
+      waku_discv5_discovered.inc(labelValues = [$shardId])
 
   return discoveredRecords
 


waku/discovery/waku_discv5.nim
```diff
@@ -232,7 +233,18 @@ proc findRandomPeers*(
   elif wd.predicate.isSome():
     discoveredRecords = discoveredRecords.filter(wd.predicate.get())
 
-  waku_discv5_discovered.inc(discoveredRecords.len)
+  # Increment metric for each discovered record's shards
+  for record in discoveredRecords:
```
fryorcraken, line 237: As you said it's computer heavy because we are going through the list of records once returned.

I wonder the metrics could be done in `wd.protocol.queryRandom` call.

I do understand the the `.filter` logic influence that. Which does make me question why we do use `wd.predicate` here, and not within `wd`.

darshankabariya: It’s an increment, so there’s no harm. My comment was about the heavy computation on connected nodes — which I’ve already optimized
.

waku/node/peer_manager/peer_manager.nim
```diff
@@ -809,6 +813,16 @@ proc logAndMetrics(pm: PeerManager) {.async.} =
         protoStreamsOut.float64, labelValues = [$Direction.Out, proto]
       )
 
+    for shard in 0 .. 63: # Assuming shards are 0-63 as per Waku spec
```
NagyZoltanPeter: This is not true I think.
Like status.prod is running with 
```
--shard=1\
--shard=32\
--shard=64\
--shard=128\
--shard=256\
--cluster-id=16\
```

And shards can be sparse... so I would avoid this pre-initialization of the metric counter.

waku/waku_relay/protocol.nim
```diff
@@ -26,6 +26,30 @@ export WakuRelayCodec
...
+  "Average length of messages seen per shard",
+  labels = ["shard"],
+)
+
+declarePublicGauge(
+  waku_relay_sum_msg_bytes_per_shard,
+  "Sum of message lengths seen per shard",
+  labels = ["shard"],
+)
+
```
NagyZoltanPeter: I wonder if we can think something more generic yet providing the same information. These counters seems overlapping the `waku_relay_network_bytes` counter.

My first thought is to combine these four. Counting messages and register sizes is enough imo.
Min/max/avg can be done in grafana I believe.

`waku_relay_network_bytes` counts with direction and `direction=in` is matching the msg size here.
WDYT is it possible to simplify?

NagyZoltanPeter (changes requested):
Let's think of some optimization around relay throughput metrics.

waku/node/peer_manager/peer_manager.nim
darshankabariya: Thanks for the comments, @NagyZoltanPeter  I’ve optimized it by iterating over shards.

waku/waku_relay/protocol.nim
darshankabariya: Thanks for the suggestion! I removed the metrics. Using **waku_relay_network_bytes** felt a bit complex due to its parameters, so I used a global variable instead.

waku/discovery/waku_discv5.nim
```diff
@@ -231,7 +232,18 @@ proc findRandomPeers*(
...
+    let typedRecord = record.toTyped().valueOr:
+      # If we can't parse the record, skip it
```
Ivansete-status, line 238:
```suggestion
      # If we can't parse the record, skip it
      trace "could not parse discv5 record"
```

waku/discovery/waku_discv5.nim
```diff
@@ -231,7 +232,18 @@ proc findRandomPeers*(
...
+      continue
+
+    let relayShards = typedRecord.relaySharding().valueOr:
+      # If no relay sharding info, skip it
```
Ivansete-status, line 242:
```suggestion
      # If no relay sharding info, skip it
      trace "could not parse discv5 record"
```

waku/node/peer_manager/peer_manager.nim
```diff
@@ -809,6 +811,16 @@ proc logAndMetrics(pm: PeerManager) {.async.} =
         protoStreamsOut.float64, labelValues = [$Direction.Out, proto]
       )
 
+    for shard in pm.wakuMetadata.shards.items:
+      waku_connected_peers_per_shard.set(0.0, labelValues = [$shard])
```
Ivansete-status, line 784: Is this loop needed?
```suggestion
```

waku/waku_relay/protocol.nim
```diff
@@ -20,12 +20,31 @@ import
...
 from ../waku_core/codecs import WakuRelayCodec
 export WakuRelayCodec
 
 logScope:
   topics = "waku relay"
 
+declareCounter waku_relay_network_bytes,
+  "total traffic per topic, distinct gross/net and direction",
+  labels = ["topic", "type", "direction"]
```
Ivansete-status, line 34: Maybe?
```suggestion
  labels = ["shard", "type", "direction"]
```

waku/waku_archive/archive_metrics.nim
```diff
@@ -2,7 +2,8 @@
 
 import metrics
 
-declarePublicGauge waku_archive_messages, "number of historical messages", ["type"]
+declarePublicGauge waku_archive_messages,
+  "number of historical messages", ["type", "shard"]
```
Ivansete-status, line 6: The possible type values should be defined in `waku/waku_archive/archive_metrics.nim` as a `const`.

Ivansete-status (commented):
Some comments so far. Thanks :raised_hands: !

PR #3433, opened 2025-05-30
chore: exposing online state in libwaku

gabrielmer:
# Description
Exposing node's online state in libwaku
# Changes

- [x] adding `waku_is_online` function in libwaku

## Issue
#3076

### DIFF
--- a/library/libwaku.h
+++ b/library/libwaku.h
@@ -233,7 +233,6 @@ int waku_ping_peer(void* ctx,
                         void* userData);
 
 int waku_is_online(void* ctx,
-                        int timeoutMs,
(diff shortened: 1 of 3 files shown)


fryorcraken:
Is the logic employed to fdefine whether it's online similar to go-waku? cause it's an **advance** logic.

gabrielmer:
> Is the logic employed to fdefine whether it's online similar to go-waku? cause it's an **advance** logic.

Yes! The logic to check whether a node is online was made based on go-waku's implementation + an additional edge case being covered by a DNS query to make sure we're connected to the internet even if we don't have reachable peers

PR #3434, opened 2025-06-02
fix: refact `rln-relay` and post sync test

darshankabariya:
# Description
This refactor was necessary because we’ve already transitioned to a non-synchronous approach, and all sync-related code is being deprecated.

Over time, GroupManager became bloated with tightly coupled and complex logic. This PR initiates a refactor to improve the clarity, maintainability, and future extensibility of the RLN relay protocol.

Changes

- [x] Restructured the GroupManager hierarchy and introduced offchainGroupManager to replace staticGroupManager.
- [x] Deprecated the static manager class in favor of the offchainManager interface. (As suggested by @jm-clius, both on-chain and off-chain trees may exist in the future; currently, OffchainManager is just an interface.)
- [x] Removed all zerokit ffi for the merkle tree along with its associated proof generation and verification wrappers.
- [x] Deleted all unit tests related to the old merkle tree implementation.
- [x]  Update all effected with onchain anvail

## Issue


darshankabariya:
Hi @Ivansete-status, @jm-clius, @stubbsta —
The PR is now ready for review.
A few test cases are taking longer than expected — I’m still looking into that.
However, all the core logic changes for improving the RLN relay protocol are complete.

waku/waku_rln_relay/group_manager/off_chain/group_manager.nim
jm-clius, line 1: I might just be missing it, but why do we need an `OffchainGroupManager` at all? We now have the base interface, which we can implement again when we re-add a real offchain implementation?

waku/waku_rln_relay/group_manager/off_chain.nim
```diff
@@ -0,0 +1,3 @@
+import off_chain/group_manager
```
jm-clius: Why have a barrel import for off_chain, but not for on_chain?

(Barrel imports are somewhat controversial, but for now that's beside the point :D )

jm-clius (commented):
Thanks! I've just eyeballed the changes for now - I think this is a good step towards simplification. So far one-ish question below.

waku/waku_rln_relay/group_manager/off_chain/group_manager.nim
darshankabariya: Yes, you got it right. Actually, I couldn’t find any use for the StaticGroupManager class, so I deprecated it along with its unit tests. As you noticed, I introduced the OffchainGroupManager interface—this idea came from your suggestion that we might support both offchain and onchain features in the future. I’ve left it unimplemented for now.

waku/waku_rln_relay/group_manager/off_chain.nim
darshankabariya: There’s another file in the same directory, waku/waku_rln_relay/group_manager/on_chain.nim, similar to the offchain one. It looks a bit oversimplified — I’ll take a closer look. 

cc @Ivansete-status @jm-clius

tests/node/test_wakunode_legacy_lightpush.nim
```diff
@@ -158,7 +158,7 @@ suite "RLN Proofs as a Lightpush Service":
     await server.stop()
 
   suite "Lightpush attaching RLN proofs":
-    asyncTest "Message is published when RLN enabled":
+    xasyncTest "Message is published when RLN enabled":
```
Ivansete-status: Why that is commented if may I ask?

tests/waku_rln_relay/rln/test_wrappers.nim
```diff
@@ -135,7 +136,7 @@ suite "RlnConfig":
         backup
 
   suite "proofGen":
-    test "Valid zk proof":
+    xasyncTest "Valid zk proof":
```
Ivansete-status: Same for that one. Why skipped?

tests/waku_rln_relay/test_rln_group_manager_offchain.nim
```diff
@@ -0,0 +1,44 @@
...
+  let credRes = membershipKeyGen(rlnInstance)
+  return credRes.get()
+
+proc generateCredentials(rlnInstance: ptr RLN, n: int): seq[IdentityCredential] =
+  var credentials: seq[IdentityCredential]
+  for i in 0 ..< n:
+    credentials.add(generateCredentials(rlnInstance))
+  return credentials
+
+suite "Offchain group manager":
```
Ivansete-status: Is this test suite running any test ?
It seems this whole module can be removed.

waku/waku_rln_relay/group_manager/group_manager_base.nim
```diff
@@ -119,58 +119,23 @@ method stop*(g: GroupManager): Future[void] {.base, async.} =
...
     g: GroupManager, input: openArray[byte], proof: RateLimitProof
 ): GroupManagerResult[bool] {.base, gcsafe, raises: [].} =
-  ## verifies the proof against the input and the current merkle root
-  let proofVerifyRes =
-    g.rlnInstance.proofVerify(input, proof, g.validRoots.items().toSeq())
-  if proofVerifyRes.isErr():
-    return err("proof verification failed: " & $proofVerifyRes.error())
-  return ok(proofVerifyRes.value())
+  ## Dummy implementation for verifyProof
+  return ok(true)
```
Ivansete-status: Why not the following?
```suggestion
  return err("veryfyProof is not implemented")
```

waku/waku_rln_relay/group_manager/off_chain/group_manager.nim
```diff
@@ -0,0 +1,54 @@
...
+
+export group_manager_base
+
+# Placeholder for future OffchainGroupManager re-implementation
+type OffchainGroupManager* = ref object of GroupManager
+  groupKeys*: seq[IdentityCredential]
+  groupSize*: uint
+
+template initializedGuard*(g: OffchainGroupManager): untyped =
+  discard
```
Ivansete-status: Why is everything `discard`ed?

tests/waku_rln_relay/utils_onchain.nim
```diff
@@ -570,17 +570,16 @@ proc setupOnchainGroupManager*(
   )
 
   let testTokenAddress = (await deployTestToken(privateKey, acc, web3)).valueOr:
-    assert false, "Failed to deploy test token contract: " & $error
-    return
+    raise
+      newException(CatchableError, "Failed to deploy test token contract: " & $error)
```
Ivansete-status: Is it strictly needed the change? I would avoid explicitly raising exceptions, if possible

tests/waku_rln_relay/utils_onchain.nim
```diff
@@ -570,17 +570,16 @@ proc setupOnchainGroupManager*(
...
 
   # mint the token from the generated account
   discard await sendMintCall(
     web3, web3.defaultAccount, testTokenAddress, acc, ethToWei(1000.u256), some(0.u256)
   )
 
   let contractAddress = (await executeForgeContractDeployScripts(privateKey, acc, web3)).valueOr:
-    assert false, "Failed to deploy RLN contract: " & $error
-    return
+    raise newException(CatchableError, "Failed to deploy RLN contract: " & $error)
```
Ivansete-status: similar to above

tests/node/test_wakunode_legacy_lightpush.nim
```diff
@@ -131,13 +134,16 @@ suite "RLN Proofs as a Lightpush Service":
     server = newTestWakuNode(serverKey, parseIpAddress("0.0.0.0"), Port(0))
     client = newTestWakuNode(clientKey, parseIpAddress("0.0.0.0"), Port(0))
 
+    anvilProc = runAnvil()
+    tempManager =
+      cast[ptr OnchainGroupManager](allocShared0(sizeof(OnchainGroupManager)))
```
Ivansete-status: Why the manual allocation is needed?

tests/waku_rln_relay/rln/waku_rln_relay_utils.nim
```diff
@@ -1,23 +1,54 @@
...
-  let epoch = rlnPeer.calcEpoch(senderEpochTime)
+  let manager = cast[OnchainGroupManager](rlnPeer.groupManager)
+  let rootUpdated = waitFor manager.updateRoots()
 
-  # we do not fetch a nonce from the nonce manager,
-  # instead we use 0 as the nonce
-  let proof = rlnPeer.groupManager.generateProof(input, epoch, 0).valueOr:
+  # Fetch Merkle proof either when a new root was detected *or* when the cache is empty.
+  if rootUpdated or manager.merkleProofCache.len == 0:
+    let proofResult = waitFor manager.fetchMerkleProofElements()
```
Ivansete-status, line 27: Even though this is for testing purposes only, the `waitFor` should be used with care and only in the top logic layers. Better `await`

tests/waku_rln_relay/rln/waku_rln_relay_utils.nim
```diff
@@ -1,23 +1,54 @@
...
   return ok()
+
+proc getWakuRlnConfig*(
+    manager: OnchainGroupManager,
+    userMessageLimit: uint64 = 1,
+    epochSizeSec: uint64 = 1,
+    treePath: string = genTempPath("rln_tree", "waku_rln_relay"),
+    index: MembershipIndex = MembershipIndex(0),
+): WakuRlnConfig =
+  let wakuRlnConfig = WakuRlnConfig(
```
Ivansete-status, line 45:
```suggestion
): WakuRlnConfig =
  ## For testing purposes only

  let wakuRlnConfig = WakuRlnConfig(
```

tests/waku_rln_relay/test_rln_serde.nim
```diff
@@ -22,7 +23,7 @@ func defaultRateCommitment*(): RateCommitment =
   return RateCommitment(idCommitment: idCredential.idCommitment, userMessageLimit: 100)
 
 suite "RLN Relay v2: serde":
-  test "toLeaf: converts a rateCommitment to a valid leaf":
+  xasyncTest "toLeaf: converts a rateCommitment to a valid leaf":
```
Ivansete-status:
```suggestion
  asyncTest "toLeaf: converts a rateCommitment to a valid leaf":
```

tests/waku_rln_relay/test_rln_serde.nim
```diff
@@ -33,7 +34,7 @@ suite "RLN Relay v2: serde":
       "09beac7784abfadc9958b3176b352389d0b969ccc7f8bccf3e968ed632e26eca"
     check expectedLeaf == leafRes.value.inHex()
 
-  test "proofGen: generates a valid zk proof":
+  xasyncTest "proofGen: generates a valid zk proof":
```
Ivansete-status:
```suggestion
  asyncTest "proofGen: generates a valid zk proof":
```

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
```diff
@@ -202,20 +197,21 @@ proc trackRootChanges*(g: OnchainGroupManager) {.async: (raises: [CatchableError
...
     const rpcDelay = 5.seconds
 
     while true:
+      await sleepAsync(rpcDelay)
       let rootUpdated = await g.updateRoots()
 
       if rootUpdated:
         if g.membershipIndex.isNone():
-          error "membershipIndex is not set; skipping proof update"
+          debug "membershipIndex is not set; skipping proof update"
```
Ivansete-status: Should not this be considered as an error?

waku/waku_rln_relay/rln_relay.nim
```diff
@@ -175,9 +141,15 @@ proc updateLog*(
...
-  return rlnPeer.calcEpoch(epochTime())
+proc getCurrentEpoch*(rlnPeer: WakuRLNRelay, t: float64 = -1.0): Epoch =
+  ## Returns the RLN epoch corresponding to `t` (Unix seconds).
+  ## If `t` is not supplied (or set to a negative value), the current time is used.
+  let timeToUse =
+    if t < 0.0:
+      epochTime()
+    else:
+      t
+  return rlnPeer.calcEpoch(timeToUse)
```
Ivansete-status: This can be simplified, considering that `t` is never being passed:
```suggestion
proc getCurrentEpoch*(rlnPeer: WakuRLNRelay): Epoch =
  ## Returns the RLN epoch corresponding to current epochTimeme is used.
  return rlnPeer.calcEpoch(epochTime())
```

darshankabariya: Thanks for pointing that out. I had made the change for a specific test, but later realized that passing just one test isn’t a good reason to change the structure of this function. I forgot to revert it. finally remove that

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
darshankabariya: It’s no longer required because earlier registering RLN membership was mandatory, but now users can run the code without membership. so removed it. thanks for pointing out, it;s help me think new direction.

tests/waku_rln_relay/test_rln_serde.nim
darshankabariya: [tests/waku_rln_relay/test_rln_serde.nim] is no longer needed. It was testing proof generation in isolation, but after moving to the on-chain tree, proof generation requires running a node and registering membership. Since it’s no longer possible to generate proofs fully off-chain, this test case is obsolete. deleting whole file.

tests/waku_rln_relay/test_rln_serde.nim
darshankabariya: same !! as below

tests/waku_rln_relay/rln/waku_rln_relay_utils.nim
darshankabariya: It’s already in the test file, so why do we need to mention it explicitly ?

tests/node/test_wakunode_legacy_lightpush.nim
darshankabariya: Unable testcase !!

tests/waku_rln_relay/test_rln_group_manager_offchain.nim
darshankabariya: Yes, I agree — we no longer need the off-chain testcase since there’s nothing left to test. but also there no harm at all, that why not delete.

tests/waku_rln_relay/rln/waku_rln_relay_utils.nim
fryorcraken: a test is a top logic layer as it uses the library as a binary.

waku/waku_rln_relay/group_manager/group_manager_base.nim
```diff
@@ -119,58 +119,23 @@ method stop*(g: GroupManager): Future[void] {.base, async.} =
...
-    raise newException(ValueError, "failed to get merkle root")
-  let rootAfterUpdate = rootRes.get()
-
-  var rootBuffer: Deque[MerkleNode]
-  let overflowedRoots = slideRootQueue(g.validRoots, rootAfterUpdate)
-  if overflowedRoots.len > 0:
-    for root in overflowedRoots:
-      discard rootBuffer.slideRootQueue(root)
-  rootBuffer
+  return g.indexOfRoot(root) >= 0
```
fryorcraken: Such a one liner method sounds like unnecessary indirection. You are not "validating" anything, so it actually create confusion.

waku/waku_rln_relay/constants.nim
```diff
@@ -37,18 +37,18 @@ const
 
 # temporary variables to test waku-rln-relay performance in the static group mode
 const
-  StaticGroupSize* = 10000
-  #  StaticGroupKeys is a static list of 10000 membership keys in the form of (identity trapdoor, identity nullifier, identity secret hash, identity commitment)
+  OffchainGroupSize* = 10000
+  #  OffchainGroupKeys is a static list of 10000 membership keys in the form of (identity trapdoor, identity nullifier, identity secret hash, identity commitment)
   #  keys are created locally, using createMembershipList proc from waku_rln_relay_utils module, and the results are hardcoded in here
   #  this list is temporary and is created to test the performance of waku-rln-relay for the static groups
   #  in the later versions, this static hardcoded group will be replaced with a dynamic one
```
fryorcraken: Can this be moved under the `tests` folder?

Or, if it is to be used by a binary, just move it near said binary.

waku/waku_rln_relay/rln_relay.nim
```diff
@@ -49,45 +49,11 @@ type RlnRelayConf* = object of RootObj
   treePath*: string
   epochSizeSec*: uint64
   userMessageLimit*: uint64
+  ethPrivateKey*: Option[string]
```
fryorcraken: Is there a type we can use to prevent accidental printing of the `string`?

waku/waku_rln_relay/constants.nim
```diff
@@ -37,18 +37,18 @@ const
 
 # temporary variables to test waku-rln-relay performance in the static group mode
 const
-  StaticGroupSize* = 10000
-  #  StaticGroupKeys is a static list of 10000 membership keys in the form of (identity trapdoor, identity nullifier, identity secret hash, identity commitment)
+  OffchainGroupSize* = 10000
```
fryorcraken: `Static` was a better name. But this is test data, so I would name it like that.

waku/waku_rln_relay/group_manager/group_manager_base.nim
darshankabariya: In this snippet, I removed slideRootQueue and simplified validateRoot. I’m not sure what specific confusion you’re referring to.

waku/waku_rln_relay/constants.nim
darshankabariya: it's related to offchain group, that's why offchainGroupSize*

waku/waku_rln_relay/rln_relay.nim
darshankabariya: Thanks for pointing that out. I think we should add some logic to the log interface, specifically for ethPrivateKey.

waku/waku_rln_relay/constants.nim
darshankabariya: No, it’s not used in tests — it’s only part of the binary.

tests/node/test_wakunode_legacy_lightpush.nim
darshankabariya: Actually, I need this onchainGroupManager to fetch real-time RLN metadata such as contractAddress and useMessageLimit. If we hardcode these values, then every time the default RLN metadata changes, we’ll also need to update the tests. but yes definatly it's finest way to fetch we also usg RlnConf default value but i feel over code.

waku/waku_rln_relay/group_manager/group_manager_base.nim
darshankabariya: good idea !!

waku/waku_rln_relay/group_manager/off_chain/group_manager.nim
darshankabariya: Thanks for pointing out, I’ve updated all functions to return an unimplemented error.

tests/waku_rln_relay/utils_onchain.nim
darshankabariya: I updated it with the idea of unification. I noticed we mostly use exceptions in our code, so I replaced asserts with exceptions. But now i realize that is for is for binary not test. so assertion is better idea right ??

tests/waku_rln_relay/utils_onchain.nim
darshankabariya: same !!

tests/waku_rln_relay/rln/waku_rln_relay_utils.nim
darshankabariya: Got it !!

darshankabariya:
> LGTM! I'd make sure we remove all the tests that are not needed and keep the needed ones. F.e., it seems we should not skip the following in `tests/waku_rln_relay/rln/test_wrappers.nim
> And also, CI green :D

Hi @Ivansete-status , thank you for the approval. I’ve addressed all your suggestions and updated the changes accordingly. All CI checks are green now and tests have been updated as well. If you see this before merging the PR, kindly have a look.

waku/waku_rln_relay/group_manager/group_manager_base.nim
```diff
@@ -119,58 +119,23 @@ method stop*(g: GroupManager): Future[void] {.base, async.} =
...
+    g: GroupManager, input: seq[byte], proof: RateLimitProof
 ): GroupManagerResult[bool] {.base, gcsafe, raises: [].} =
-  ## verifies the proof against the input and the current merkle root
-  let proofVerifyRes =
-    g.rlnInstance.proofVerify(input, proof, g.validRoots.items().toSeq())
-  if proofVerifyRes.isErr():
-    return err("proof verification failed: " & $proofVerifyRes.error())
-  return ok(proofVerifyRes.value())
+  ## Dummy implementation for verifyProof
+  return err("verifyProof is not implemented")
```
stubbsta: I'm probably missing something obvious here, but what is the need for these "dummy" implementations?

darshankabariya: It’s based on GroupManager, which is an abstract class. The functionality is implemented in the derived class onchainGroupManager class.

waku/waku_rln_relay/group_manager/off_chain/group_manager.nim
```diff
@@ -0,0 +1,64 @@
+import ../group_manager_base
+
+export group_manager_base
+
+# Placeholder for future OffchainGroupManager re-implementation
+type OffchainGroupManager* = ref object of GroupManager
```
jm-clius: Again, I don't understand why we have this non-implementation as placeholder. There is no real implementation and this is only referenced in tests (and in `rln_relay.nim`, where it probably shouldn't be included anymore, see my comment there). In other words, wherever you still use `OffchainGroupManager` in the tests it could just be `GroupManager` and we delete the files and folders for `off_chain` for now. Am I missing some important context?

waku/waku_rln_relay/rln_relay.nim
```diff
@@ -447,11 +413,11 @@ proc mount(
 
   if not conf.dynamic:
```
jm-clius: We don't support non-dynamic group managers anymore, from what I can see. There's no "OffchainGroupManager" - just a stub with that name that will raise exceptions for every function call. In other words, we can remove this completely and mark the config item as deprecated.

jm-clius (changes requested):
Thanks! I'm ready to approve once I understand why the `OffchainGroupManager` exists at all. :) Will be happy to have a call if I'm missing some context or my comments are unclear.

waku/waku_rln_relay/group_manager/off_chain/group_manager.nim
darshankabariya: thanks for review. main objective for offchainGM is to serve as a future perspective, so that when we revisit the offchain tree, we can leverage the currently unimplemented offchainGM functionality. But I understand your point — we can remove offchainGM for now as well, should i remove or not ??

waku/waku_rln_relay/rln_relay.nim
darshankabariya: Yes!! we also need to remove that along with offchainGM.

waku/waku_rln_relay/rln_relay.nim
```diff
@@ -445,38 +411,25 @@ proc mount(
   let rlnInstance = createRLNInstance(tree_path = conf.treePath).valueOr:
     return err("could not create RLN instance: " & $error)
 
-  if not conf.dynamic:
```
jm-clius: I think this config item must be marked in the external config as deprecated (and then removed within two releases). Presumably it will not break any existing deployments and the default is `true` in any case?

apps/benchmarks/benchmarks.nim
```diff
@@ -17,7 +17,7 @@ proc main(): Future[string] {.async, gcsafe.} =
   let rlnIns = createRLNInstance(20).get()
   let credentials = toSeq(0 .. 1000).mapIt(membershipKeyGen(rlnIns).get())
 
-  let manager = StaticGroupManager(
+  let manager = OffchainGroupManager(
```
jm-clius: I think this reference does not exist anymore. Perhaps instantiate the base `GroupManager`?

jm-clius (approved):
Thanks! Two comments below, but no need to re-request review once addressed.

apps/benchmarks/benchmarks.nim
darshankabariya: Will update the benchmark accordingly.

waku/waku_rln_relay/rln_relay.nim
darshankabariya: it’s a really good idea. I found that the default value true for dynamic is set only inside NetworkMonitor. we will handle in seperate PR.

PR #3438, opened 2025-06-05
feat: add waku_disconnect_all_peers to libwaku

gabrielmer:
# Description

Adding a `waku_disconnect_all_peers` function to libwaku, which disconnects from all currently connected peers.

# Changes

- [x] adding `waku_disconnect_all_peers` to libwaku

## Issue
#3076

### DIFF
--- a/library/waku_thread/inter_thread_communication/requests/peer_manager_request.nim
+++ b/library/waku_thread/inter_thread_communication/requests/peer_manager_request.nim
@@ -12,6 +12,7 @@ type PeerManagementMsgType* {.pure.} = enum
   GET_CONNECTED_PEERS_INFO
   GET_PEER_IDS_BY_PROTOCOL
   DISCONNECT_PEER_BY_ID
+  DISCONNECT_ALL_PEERS
   DIAL_PEER
   DIAL_PEER_BY_ID
   GET_CONNECTED_PEERS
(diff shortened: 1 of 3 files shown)


library/waku_thread/inter_thread_communication/requests/peer_manager_request.nim
```diff
@@ -121,6 +122,16 @@ proc process*(
       return err($error)
     await waku.node.peerManager.disconnectNode(peerId)
     return ok("")
+  of DISCONNECT_ALL_PEERS:
+    let connectedPeers = waku.node.peerManager.switch.peerStore.peers().filterIt(
+        it.connectedness == Connected
+      )
```
Ivansete-status, line 128: I wonder if we could start making `switch` private within `PeerManager`
Something to consider in the future

Ivansete-status (approved):
LGTM! Just adding a possible future enhancement, completely unrelated to this PR

library/waku_thread/inter_thread_communication/requests/peer_manager_request.nim
gabrielmer: Ooh great point! It would imply changes in many places so it's definitely a task of its own. I'll open an issue so we don't forget about it :))

Thanks so much!

Update: opened https://github.com/waku-org/nwaku/issues/3439

library/waku_thread/inter_thread_communication/requests/peer_manager_request.nim
```diff
@@ -121,6 +122,16 @@ proc process*(
...
+
+    var futs: seq[Future[void]]
+    for peer in connectedPeers:
+      futs.add(waku.node.peerManager.disconnectNode(peer))
+    await allFutures(futs)
+    return ok("")
```
NagyZoltanPeter, line 134: Is it necessary to wait for all to disconnect if we just return ok at the end?

gabrielmer: Yes! So I initially considered to do it that way, but then thought that it's better for the libwaku function to return once we disconnected from all of them.

Otherwise, a user may attempt to perform an operation or a check assuming that we already disconnected when we aren't.

It's a bit of an edge case, but wanted to keep that "synchronous" logic for the libwaku user

PR #3442, opened 2025-06-06
fix: typo from `DIRVER` to `DRIVER`

danisharora099:
# Description
This PR changes the word `DIRVER` to `DRIVER`


Ivansete-status (approved):
LGTM! Thanks for it :partying_face:

weboko:
@danisharora099 tests fail :)

danisharora099:
> @danisharora099 tests fail :)

They fail probably because they are run against the LTS nwaku and nwaku master -- both of which have `DRIVER` as `DIRVER` and the mismatch thus causes the tests to fail -- will start with merging nwaku (this PR) and then js-waku should see nwaku_master tests pass ideally

PR #3449, opened 2025-06-11
fix: add rest api to update loglevel

darshankabariya:


NagyZoltanPeter:
Hey @darshankabariya!

I think I messed this up. Just picked this up and shot on it.
Although your solution seems more complete. The one diff I used path param for the POST call and not content body... for simplicity.
See: https://github.com/waku-org/nwaku/pull/3451
Lets adjust the two into one! WDYT?

darshankabariya:
Closing this PR due to duplication with existing work.

PR #3451, opened 2025-06-11
feat: dynamic logging via REST API

### LINKED ISSUE #3437: feat: enable nwaku change log level dynamically
### Problem
Right now if we want to change log level from `info` to `trace` we need to restart `nwaku` node.
This becomes tricky when we want to investigate issues on the fleet, because some of the issues can disappear after restart. 

### Suggested solution
Enable changing some environment variable or something similar to change logging level of `nwaku` without restarting.

NagyZoltanPeter:
# Description
Added /admin/v1/log-level/{logLevel} endpoint that is used for dynamic log level setting

## Limitation

You cannot set log level below the level the code was compiled with. 
wakunode2 is currently compiled with TRACE, btw.

# Changes
- [x] set log level runtime via new endpoint: POST /admin/v1/log-level/{logLevel}

## Issue

closes #3437 

### notice: we don't need to add `-d:chronicles_runtime_filtering=on` as it is already set at https://github.com/waku-org/nwaku/blob/66fee5c7d8a3d35304e26711e682531f6201f5f8/apps/wakunode2/nim.cfg#L3

### DIFF
--- a/waku/waku_api/rest/admin/handlers.nim
+++ b/waku/waku_api/rest/admin/handlers.nim
@@ -3,6 +3,7 @@
 import
   std/[sets, strformat, sequtils, tables],
   chronicles,
+  chronicles/topics_registry,
   json_serialization,
   presto/route,
   libp2p/[peerinfo, switch, peerid, protocols/pubsub/pubsubpeer]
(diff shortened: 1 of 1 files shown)


NagyZoltanPeter:
@Ivansete-status @darshankabariya 
I have adapted the compile time switch check from @darshankabariya 's solution.
Intentionally left out client part of the API, as I think it is not beneficial to have another app should set different log level this way (although it is still possible). This interface is intended to be used for human testing and debugging.

PR #3453, opened 2025-06-12
chore: refactor rm discv5-only

Ivansete-status:
## Description
Simple remove `discv5-only` parameter.
In the future, if needed, we may be able to achieve the same behavior by the `preset` config parameter.

## Issue

### DIFF
--- a/waku/factory/waku.nim
+++ b/waku/factory/waku.nim
@@ -364,7 +364,7 @@ proc startWaku*(waku: ptr Waku): Future[Result[void, string]] {.async.} =
   (await startNode(waku.node, waku.conf, waku.dynamicBootstrapNodes)).isOkOr:
     return err("error while calling startNode: " & $error)
 
-  # Update waku data that is set dynamically on node startdiscv5Only
+  ## Update waku data that is set dynamically on node start
   updateWaku(waku).isOkOr:
     return err("Error in updateApp: " & $error)
 


waku/factory/waku.nim
```diff
@@ -361,13 +361,12 @@ proc startWaku*(waku: ptr Waku): Future[Result[void, string]] {.async.} =
...
-  if conf.discv5Conf.isNone or not conf.discv5Conf.get().discv5Only:
-    (await startNode(waku.node, waku.conf, waku.dynamicBootstrapNodes)).isOkOr:
-      return err("error while calling startNode: " & $error)
+  (await startNode(waku.node, waku.conf, waku.dynamicBootstrapNodes)).isOkOr:
+    return err("error while calling startNode: " & $error)
 
-    # Update waku data that is set dynamically on node start
-    updateWaku(waku).isOkOr:
-      return err("Error in updateApp: " & $error)
+  # Update waku data that is set dynamically on node startdiscv5Only
```
gabrielmer: is the `startdiscv5Only` intentional? not sure I understand

gabrielmer (approved):
LGTM, thanks so much!

waku/factory/waku.nim
Ivansete-status: Ooh good catch , thanks ! Addressed in https://github.com/waku-org/nwaku/pull/3453/commits/3a8974b5f2546835d41ee81c3225676c31384187

PR #3454, opened 2025-06-12
chore: rest-relay-cache-capacity 

### LINKED ISSUE #3444: bug: rest-relay-cache-capacity no longer has an effect on the relay cached messages
Passing rest-relay-cache-capacity no longer seems to influence the number of messages cached and retrieve via relay/v1/messages GET call. No matter the value passed via that flag, we get 50 messages back.
This worked in the past

Steps:
- start 2 relay nodes with --rest-relay-cache-capacity=100
- node1 published 100 messages
- node 2 fetches the messages
Expected:
- 100 messages fetched
Actual:
- 50 messages fetched

[node2.log](https://github.com/user-attachments/files/20653859/node2.log)
[node1.log](https://github.com/user-attachments/files/20653860/node1.log)

darshankabariya:
This is a small PR to eliminate hard-coded values for rest-relay-cache-capacity.

close #3444

### DIFF
--- a/waku/factory/external_config.nim
+++ b/waku/factory/external_config.nim
@@ -503,7 +503,7 @@ with the drawback of consuming some more bandwidth.""",
 
     restRelayCacheCapacity* {.
       desc: "Capacity of the Relay REST API message cache.",
-      defaultValue: 30,
+      defaultValue: 50,
       name: "rest-relay-cache-capacity"
     .}: uint32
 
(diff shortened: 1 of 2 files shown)


NagyZoltanPeter (approved):
It's great! Thank you!
Maybe one though if we can adjust the cli arg default to the previously used hard coded default (50)?
https://github.com/waku-org/nwaku/blob/49751b777545405b2763c00eaff8e2dc943a7c52/waku/factory/external_config.nim#L506

Ivansete-status (approved):
LGTM! Thanks for it! :100:

PR #3455, opened 2025-06-13
feat: lighptush v3 for lite-protocol-tester

NagyZoltanPeter:
# Description
With the release and adoption of lightpuhs v3 it is essential to be able to test it with lpt.

# Changes

- [x] implemented configurable switch between legacy and v3 lightpush
- [x] support this configurability in all possible environment, running standalone, against any fleet from lpt-runner docker env and also on infra environment. 
- [x] leave default to legacy lightpush for now, to let current instance running at infra on status.prod not disturbed by the change as status.prod still not supports v3 lightpush.
## Issue

closes #
https://github.com/waku-org/nwaku/issues/3359


apps/liteprotocoltester/v3_publisher.nim
```diff
@@ -0,0 +1,28 @@
...
+import waku/[waku_node, waku_core, waku_lightpush]
+import publisher_base
+
+type V3Publisher* = ref object of PublisherBase
+
+proc new*(T: typedesc[V3Publisher], wakuNode: WakuNode): V3Publisher =
+  if isNil(wakuNode.wakuLightpushClient):
+    wakuNode.mountLegacyLightPushClient()
+
+  result = V3Publisher(wakuNode: wakuNode)
```
gabrielmer: I think we try to use `result`? probably would change it to `return`, but it's just a detail

apps/liteprotocoltester/v3_publisher.nim
```diff
@@ -0,0 +1,28 @@
...
+): Future[Result[void, string]] {.async.} =
+  discard (
+    await self.wakuNode.lightpushPublish(some(topic), message, some(servicePeer))
+  ).valueOr:
+    if error.code == NO_PEERS_TO_RELAY and
+        error.desc != some("No peers for topic, skipping publish"):
+      # TODO: We need better separation of errors happening on the client side or the server side.-
+      return err("dial_failure")
+    else:
+      return err($error.code)
```
gabrielmer, line 28: if we discard the output of `valueOr`, shouldn't we use `isOkOr` instead?

apps/liteprotocoltester/liteprotocoltester.nim
```diff
@@ -126,12 +126,20 @@ when isMainModule:
   nodeHealthMonitor = WakuNodeHealthMonitor()
   nodeHealthMonitor.setOverallHealth(HealthStatus.INITIALIZING)
 
-  let restServer = rest_server_builder.startRestServerEssentials(
-    nodeHealthMonitor, wakuConf
-  ).valueOr:
-    error "Starting esential REST server failed.", error = $error
+  let wakuConf = wConf.toWakuConf().valueOr:
```
gabrielmer, line 129: not sure I understand what's the difference between `wConf` and `wakuConf` 😶

apps/liteprotocoltester/v3_publisher.nim
NagyZoltanPeter: Oh, maybe a comment would fit here. As the two type of lightpush has no common result return I must accomodate at a common point. From the test pov it's enough to know if send succeeded, but failures are measured by metrics.
In a separate PR the TODO comment. It is not good from usage pov that NO_PEERS_TO_RELAY coming even if the problem occurs on the client side and also if its from the server side. So no easy chance to act properly.
Well its clearly how eating my own code looks like!

apps/liteprotocoltester/liteprotocoltester.nim
NagyZoltanPeter: This is a follow up after waku cli arg rework. So I needed to rename somehow. wConf represents the waku node external-configs here.

apps/liteprotocoltester/legacy_publisher.nim
```diff
@@ -0,0 +1,23 @@
...
+method send*(
+    self: LegacyPublisher,
+    topic: PubsubTopic,
+    message: WakuMessage,
+    servicePeer: RemotePeerInfo,
+): Future[Result[void, string]] {.async.} =
+  discard (
+    await self.wakuNode.legacyLightpushPublish(some(topic), message, servicePeer)
+  ).valueOr:
+    return err(error)
```
Ivansete-status, line 23: Even though it may be redundant, I prefer give extra context
```suggestion
    return err("failed send in legacy_publisher: " & error)
```

apps/liteprotocoltester/publisher_base.nim
```diff
@@ -0,0 +1,13 @@
...
+type PublisherBase* = ref object of RootObj
+  wakuNode*: WakuNode
+
+method send*(
+    self: PublisherBase,
+    topic: PubsubTopic,
+    message: WakuMessage,
+    servicePeer: RemotePeerInfo,
+): Future[Result[void, string]] {.base, async.} =
+  raiseAssert "Not implemented!"
```
Ivansete-status: Let's avoid raising exceptions ourselves, as much as possible :)
```suggestion
  return err("publisher_base - send method not implemented")
```

apps/liteprotocoltester/tester_config.nim
```diff
@@ -80,6 +84,12 @@ type LiteProtocolTesterConf* = object
     name: "test-func"
   .}: TesterFunctionality
 
+  lightpushVersion* {.
+    desc: "Version of the sender to use. Supported values: legacy, v3.",
+    defaultValue: LightpushVersion.LEGACY,
```
Ivansete-status, line 89: Why not ;P ?
```suggestion
    defaultValue: LightpushVersion.V3,
```

apps/liteprotocoltester/v3_publisher.nim
```diff
@@ -0,0 +1,28 @@
...
+import waku/[waku_node, waku_core, waku_lightpush]
+import publisher_base
+
+type V3Publisher* = ref object of PublisherBase
+
+proc new*(T: typedesc[V3Publisher], wakuNode: WakuNode): V3Publisher =
+  if isNil(wakuNode.wakuLightpushClient):
+    wakuNode.mountLegacyLightPushClient()
+
+  return V3Publisher(wakuNode: wakuNode)
```
Ivansete-status: Another possible option. Dunno which one is better
```suggestion
proc new*(T: type V3Publisher, wakuNode: WakuNode): T =
  if isNil(wakuNode.wakuLightpushClient):
    wakuNode.mountLegacyLightPushClient()

  return T(wakuNode: wakuNode)
```

Ivansete-status (approved):
LGTM! Thanks for it! :100: 
Just adding some nitpicks that I hope you find useful
Specially interesting to me the avoidance of raising exceptions. In my dreamed nwaku, we don't raise any exception and just catch any one raised by vendors.

PR #3456, opened 2025-06-13
chore: refactor to unify online and health monitors

Ivansete-status:
## Description
Simple PR to unify the current online and node health monitors.

If you agree with this PR , we will need to implement a small adaptation in [waku-interop-tests](https://github.com/waku-org/waku-interop-tests) and make it to wait for `READY` instead of `Ready`.

## Issue


waku/node/health_monitor/online_monitor.nim
```diff
@@ -0,0 +1,77 @@
...
+
+proc startOnlineMonitor*(self: OnlineMonitor) =
+  self.networkConnLoopHandle = self.networkConnectivityLoop()
+
+proc stopOnlineMonitor*(self: OnlineMonitor) {.async.} =
+  if not self.networkConnLoopHandle.isNil():
+    await self.networkConnLoopHandle.cancelAndWait()
+
+proc setPeerStoreToOnlineMonitor*(self: OnlineMonitor, peerStore: PeerStore) =
+  self.peerStore = peerStore
```
gabrielmer, line 66: Do we need this? shouldn't we just pass the peer store in `init()`?

waku/factory/waku.nim
```diff
@@ -151,6 +155,16 @@ proc setupAppCallbacks(
...
 
+proc startRestServerEssentials(
+    healthMonitor: NodeHealthMonitor, wakuConf: WakuConf
+): Result[WakuRestServerRef, string] =
+  let restServer = rest_server_builder.startRestServerEssentials(
+    healthMonitor, wakuConf.restServerConf.get(), wakuConf.portsShift
+  ).valueOr:
+    return err("Starting essential REST server failed: " & $error)
+
+  return ok(restServer)
```
gabrielmer: what's the purpose of this function? isn't it the same as directly calling `rest_server_builder.startRestServerEssentials`?

waku/node/health_monitor/online_monitor.nim
Ivansete-status: Yes good point! Unfortunately we can't do that without deeper refactor (doesn't seem super feasible though.)
The thing is that, first of all, the REST server is started, and later the node instance is created.

1. REST server starts and the following is invoked:
```nim
proc new*(
    T: type NodeHealthMonitor,
    dnsNameServers = @[parseIpAddress("1.1.1.1"), parseIpAddress("1.0.0.1")],
): T =
  T(
    nodeHealth: INITIALIZING,
    node: nil,
    onlineMonitor: OnlineMonitor.init(dnsNameServers),
  )
  ```
  Notice that we don't pass a `node` instance yet.
  
2. The node is created, started and passed to the `OnlineMonitor` instance, where the `setPeerStoreToOnlineMonitor` is invoked.

waku/factory/waku.nim
Ivansete-status: Good catch! Addressed in https://github.com/waku-org/nwaku/pull/3456/commits/54bf54c62f2001874c521b05fd3c7767116419dc

waku/node/health_monitor/online_monitor.nim
NagyZoltanPeter: Yes, unfortunately this is the price for having /health endpoint the earliest possible.
We will see that the new RLN sync concept might enables this to be changed though.

PR #3457, opened 2025-06-17
fix: dnsName servers is not properly set to waku node

Ivansete-status:
## Description

The `wakunode2` could not properly start because the dns servers list was not passed properly
We avoid the issue discovered by @siddarthkay and mentioned in:

- https://github.com/status-im/infra-role-nim-waku/pull/40#issuecomment-2969029505

## Issue


gabrielmer (approved):
Thanks so much!

PR #3458, opened 2025-06-18
chore(debug): disconnection handling

gabrielmer:
# Description

# Changes

- [ ] ...
- [ ] ...


waku/node/peer_manager/peer_manager.nim
```diff
@@ -387,6 +387,15 @@ proc disconnectNode*(pm: PeerManager, peer: RemotePeerInfo) {.async.} =
   let peerId = peer.peerId
   await pm.disconnectNode(peerId)
 
+proc disconnectAllPeers*(pm: PeerManager) {.async.} =
+  let connectedPeers =
+    pm.switch.peerStore.peers().filterIt(it.connectedness == Connected)
+
+  var futs: seq[Future[void]]
+  for peer in connectedPeers:
+    futs.add(pm.disconnectNode(peer))
```
Ivansete-status: Maybe :)?
```suggestion
  let futs = connectedPeers.mapIt(pm.disconnectNode(it))
```

waku/node/waku_node.nim
```diff
@@ -69,6 +69,8 @@ declarePublicGauge waku_px_peers,
 logScope:
   topics = "waku node"
 
+randomize()
```
Ivansete-status, line 74: Why this is needed if may I ask?
Interesting adding a comment

waku/node/waku_node.nim
```diff
@@ -1325,35 +1327,190 @@ proc mountLibp2pPing*(node: WakuNode) {.async: (raises: []).} =
...
+  try:
+    # Establish connection
+    let connResult = await node.peerManager.dialPeer(peerId, PingCodec)
+    if connResult.isNone():
+      error "pingPeer: failed dialing peer", peerId = peerId
+      return err("pingPeer failed dialing peer peerId: " & $peerId)
+
+    let conn = connResult.get()
+    defer:
+      # Always close the connection
```
Ivansete-status: I think the `Connection` term will be eventually adapted to `Stream` in nim-libp2p. I think we should treat it as stream
```suggestion
      # Always close the stream
```

waku/node/waku_node.nim
```diff
@@ -1325,35 +1327,190 @@ proc mountLibp2pPing*(node: WakuNode) {.async: (raises: []).} =
...
+  ## Ping a single peer and return the result
+
+  try:
+    # Establish connection
+    let connResult = await node.peerManager.dialPeer(peerId, PingCodec)
+    if connResult.isNone():
+      error "pingPeer: failed dialing peer", peerId = peerId
+      return err("pingPeer failed dialing peer peerId: " & $peerId)
+
+    let conn = connResult.get()
```
Ivansete-status: Dunno whether this is feasible, just in case ;P
```suggestion
    let stream = (await node.peerManager.dialPeer(peerId, PingCodec)).valueOr:
      error "pingPeer: failed dialing peer", peerId = peerId
      return err("pingPeer failed dialing peer peerId: " & $peerId)
```

waku/node/waku_node.nim
```diff
@@ -1325,35 +1327,190 @@ proc mountLibp2pPing*(node: WakuNode) {.async: (raises: []).} =
...
+  ## Select peers for random keepalive, prioritizing mesh peers
+
+  let meshPeers = node.wakuRelay.getPeersInMesh().valueOr:
+    error "Failed getting peers in mesh for ping", error = error
+    # Fallback to random selection from all outgoing peers
+    var randomPeers = outPeers
+    shuffle(randomPeers)
+    return randomPeers[0 ..< min(len(randomPeers), numRandomPeers)]
+
+  debug "Mesh peers for keepalive", meshPeers = meshPeers
```
Ivansete-status: That seems to be too verbose
```suggestion
  trace "Mesh peers for keepalive", meshPeers = meshPeers
```

waku/node/waku_node.nim
```diff
@@ -1325,35 +1327,190 @@ proc mountLibp2pPing*(node: WakuNode) {.async: (raises: []).} =
...
+
+  # Get non-mesh peers and shuffle them
+  var nonMeshPeers = outPeers.filterIt(it notin meshPeers)
+  shuffle(nonMeshPeers)
+
+  # Combine mesh peers + random non-mesh peers up to numRandomPeers total
+  let numNonMeshPeers = max(0, numRandomPeers - len(meshPeers))
```
Ivansete-status: This is so beautiful :heart_eyes:

waku/node/waku_node.nim
```diff
@@ -1325,35 +1327,190 @@ proc mountLibp2pPing*(node: WakuNode) {.async: (raises: []).} =
...
+  let selectedNonMeshPeers = nonMeshPeers[0 ..< min(len(nonMeshPeers), numNonMeshPeers)]
+
+  result = meshPeers & selectedNonMeshPeers
+  debug "Selected peers for keepalive", selected = result
```
Ivansete-status: For some reason I forgot we have to avoid the implicit `result` type
```suggestion
  let selectedPeers = meshPeers & selectedNonMeshPeers
  trace "Selected peers for keepalive", selected = selectedPeers
  return selectedPeers
```

waku/node/waku_node.nim
```diff
@@ -1325,35 +1327,190 @@ proc mountLibp2pPing*(node: WakuNode) {.async: (raises: []).} =
...
+    numRandomPeers = 10,
+) {.async.} =
+  # Validate input parameters
+  if randomPeersKeepalive.isZero() or allPeersKeepAlive.isZero():
+    error "keepaliveLoop: allPeersKeepAlive and randomPeersKeepalive must be greater than 0",
+      randomPeersKeepalive = $randomPeersKeepalive,
+      allPeersKeepAlive = $allPeersKeepAlive
+    raise newException(
+      CatchableError,
+      "keepaliveLoop: allPeersKeepAlive and randomPeersKeepalive must be greater than 0",
```
Ivansete-status: Let's avoid raising exceptions.
Given that this seems to be invoked in `startKeepalive` I think we can move that validation there and make `startKeepalive` return a `Result`

waku/node/waku_node.nim
```diff
@@ -1325,35 +1327,190 @@ proc mountLibp2pPing*(node: WakuNode) {.async: (raises: []).} =
...
-          warn "Failed dialing peer for keep alive", peerId = peerId
-          continue
-        let pingDelay = await node.libp2pPing.ping(conn)
-        await conn.close()
-      except CatchableError as exc:
-        waku_node_errors.inc(labelValues = ["keep_alive_failure"])
+    if countdownToPingAll > 0:
+      debug "Pinging random peers",
+        count = len(peersToPing), countdownToPingAll = countdownToPingAll
+      countdownToPingAll -= 1
```
Ivansete-status: Maybe more idiomatic? Similarly, with `inc()`
```suggestion
      countdownToPingAll.dec()
```

waku/node/waku_node.nim
```diff
@@ -1325,35 +1327,190 @@ proc mountLibp2pPing*(node: WakuNode) {.async: (raises: []).} =
...
+    else:
+      debug "Pinging all peers", count = len(peersToPing)
+      countdownToPingAll = max(0, randomToAllRatio - 1)
+
+    # Execute keepalive pings
+    let successfulPings = await keepalivePings(node, peersToPing)
+
+    debug "Keepalive results",
```
Ivansete-status: Seems to be quite spammy ;P
```suggestion
    trace "Keepalive results",
```

waku/node/waku_node.nim
```diff
@@ -1325,35 +1327,190 @@ proc mountLibp2pPing*(node: WakuNode) {.async: (raises: []).} =
...
-        let pingDelay = await node.libp2pPing.ping(conn)
-        await conn.close()
-      except CatchableError as exc:
-        waku_node_errors.inc(labelValues = ["keep_alive_failure"])
+    if countdownToPingAll > 0:
+      debug "Pinging random peers",
+        count = len(peersToPing), countdownToPingAll = countdownToPingAll
+      countdownToPingAll -= 1
+    else:
+      debug "Pinging all peers", count = len(peersToPing)
```
```suggestion
      trace "Pinging all peers", count = len(peersToPing)
```

waku/node/waku_node.nim
```diff
@@ -1325,35 +1327,190 @@ proc mountLibp2pPing*(node: WakuNode) {.async: (raises: []).} =
...
+    else:
+      consecutiveIterationFailures = 0
+
+    lastTimeExecuted = currentTime
 
 # 2 minutes default - 20% of the default chronosstream timeout duration
-proc startKeepalive*(node: WakuNode, keepalive = 2.minutes) =
-  info "starting keepalive", keepalive = keepalive
+proc startKeepalive*(
+    node: WakuNode, randomPeersKeepalive = 10.seconds, allPeersKeepalive = 2.minutes
```
Ivansete-status: I think this is a great opportunity to move this proc somewhere in `node_health_monitor.nim`, if possible

waku/node/waku_node.nim
```diff
@@ -1325,35 +1327,190 @@ proc mountLibp2pPing*(node: WakuNode) {.async: (raises: []).} =
...
+) =
+  info "starting keepalive",
+    randomPeersKeepalive = randomPeersKeepalive, allPeersKeepalive = allPeersKeepalive
 
-  asyncSpawn node.keepaliveLoop(keepalive)
+  asyncSpawn node.keepaliveLoop(randomPeersKeepalive, allPeersKeepalive)
```
Ivansete-status: I think this is a good opportunity as well to retrieve the loop future handle, which will help us in the future to perform a graceful shutdown, instead of the current spawn and forget :D
```suggestion
  monitor.keepAliveFut = node.keepaliveLoop(randomPeersKeepalive, allPeersKeepalive)
```

Ivansete-status (commented):
Thanks for it! :partying_face: 
I just add some nitpick comments that I hope you find useful

waku/node/waku_node.nim
```diff
@@ -1325,35 +1327,190 @@ proc mountLibp2pPing*(node: WakuNode) {.async: (raises: []).} =
...
     error "failed to mount libp2pPing", error = getCurrentExceptionMsg()
 
-# TODO: Move this logic to PeerManager
-proc keepaliveLoop(node: WakuNode, keepalive: chronos.Duration) {.async.} =
+proc pingPeer(node: WakuNode, peerId: PeerId): Future[Result[void, string]] {.async.} =
+  ## Ping a single peer and return the result
+
+  try:
+    # Establish connection
+    let connResult = await node.peerManager.dialPeer(peerId, PingCodec)
```
fryorcraken: isn't the point to ping peers you are already connected to?

waku/node/waku_node.nim
```diff
@@ -1325,35 +1327,190 @@ proc mountLibp2pPing*(node: WakuNode) {.async: (raises: []).} =
...
+    # Perform ping
+    let pingDuration = await node.libp2pPing.ping(conn)
+
+    trace "Ping successful", peerId = peerId, duration = pingDuration
+    return ok()
+  except CatchableError as e:
+    error "pingPeer: exception raised pinging peer", peerId = peerId, error = e.msg
+    return err("pingPeer: exception raised pinging peer: " & e.msg)
+
+proc selectRandomPeersForKeepalive(
```
fryorcraken: You are selecting random peers **in the mesh**.

What if we don't use relay?

I would suggest to split this function in two part:

1. proceed with random peer selection among a set, and do pings
2. select mesh peers as the set

single-purpose functions + ready for edge mode.

waku/node/waku_node.nim
gabrielmer: Ooh good point, just added the comment! It is to initialize `std/random`'s random number generator, if we don't call it, the randomization operations will give the same output in different runs. I think that's not a big deal in our case, but it's the correct way of using randomization procedures.

In our case, we need it for `shuffle()` when we try to select random peers to ping

waku/node/waku_node.nim
gabrielmer: Ooh great catch! Yes that log was spam for sure hahaha, my bad! Some logs I used for debugging I forgot to move to trace 😶

waku/node/waku_node.nim
gabrielmer: Niiiiice, thanks for noticing! That was my friend Don Claudio when it helped me refactor the code - when I checked the output I didn't notice it was using the implicit `result` :))

Amazing catch!

waku/node/waku_node.nim
gabrielmer: You think so? it just says how many pings were attempted and how many succeeded, I saw it as decently useful information and it's a short line.

If that's the case, maybe we should turn all the keep alive logs into trace except the error ones

update: that's what I did, I think it's better that way so we don't get keepalive logs super often :)

waku/node/waku_node.nim
gabrielmer: Done! When moving it to the node_health_monitor I did have to get rid of the `keep-alive` config option. I planned to do it in a separate PR but with that move I had to include it in this one

waku/node/waku_node.nim
gabrielmer: Yes! so `nim-libp2p` kind of mixes the concepts of connections and streams, they are all under the `Connection` object. In this case, what we actually do is opening a new stream for the ping and not really a new connection. Changed the naming of the variables to use stream instead of connection, which is misleading :)

waku/node/waku_node.nim
```diff
@@ -1325,35 +1329,160 @@ proc mountLibp2pPing*(node: WakuNode) {.async: (raises: []).} =
...
+
+    let res = fut.read()
+    if res.isOk():
+      successCount.inc()
+    else:
+      waku_node_errors.inc(labelValues = ["keep_alive_failure"])
+
+  return successCount
+
+proc keepaliveLoop*(
```
Ivansete-status: We can move this proc implementation to `node_health_monitor.nim`, can't we :) ?
```suggestion
proc keepAliveLoop*(
```

Ivansete-status (commented):
Thanks so much for the recent updates!
I'm just adding one nitpick

waku/node/waku_node.nim
gabrielmer: Ooh great point! Addressed in: https://github.com/waku-org/nwaku/pull/3458/commits/8c35b5780dbdd556111de95a3420817d955bd736

waku/node/waku_node.nim
gabrielmer: Done!

waku/node/health_monitor/node_health_monitor.nim
```diff
@@ -224,6 +233,144 @@ proc getRendezvousHealth(hm: NodeHealthMonitor): ProtocolHealth =
...
+    if currentTime - lastTimeExecuted > sleepDetectionInterval:
+      warn "Keep alive hasn't been executed recently. Killing all connections"
+      await node.peerManager.disconnectAllPeers()
+      lastTimeExecuted = currentTime
+      consecutiveIterationFailures = 0
+      continue
+
+    # Check for consecutive failures
+    if consecutiveIterationFailures > maxAllowedConsecutiveFailures:
+      warn "Too many consecutive ping failures, node likely disconnected. Killing all connections"
```
Ivansete-status: For the same prize :D
```suggestion
      warn "Too many consecutive ping failures, node likely disconnected. Killing all connections", consecutiveIterationFailures, maxAllowedConsecutiveFailures
```

waku/node/waku_node.nim
```diff
@@ -1325,35 +1329,60 @@ proc mountLibp2pPing*(node: WakuNode) {.async: (raises: []).} =
...
+  ## Ping a single peer and return the result
 
-    # Keep connected peers alive while running
-    # Each node is responsible of keeping its outgoing connections alive
-    trace "Running keepalive"
+  try:
+    # Establish a stream
+    let stream = (await node.peerManager.dialPeer(peerId, PingCodec)).valueOr:
+      error "pingPeer: failed dialing peer", peerId = peerId
+      return err("pingPeer failed dialing peer peerId: " & $peerId)
```
NagyZoltanPeter, line 1339: I just wonder and curoius, if dialPeer returns success isn't that mean that the other peer is alive?
Per se, dialer will make a handshake on negotiateStream via MultistreamSelect...
Is it something special ping data exchange will do?

NagyZoltanPeter (approved):
Great work!!! Thank you!
Just had a small question regarding keep alive internal mechanism...

waku/node/waku_node.nim
fryorcraken: is there a way to esnure we are **not** establishing a new connection?

JS and Rust libp2p implementations allow you to grab streams out of a connection, for example:

https://github.com/waku-org/js-waku/blob/cd1d909de38bfecb0543d0642feb219804119e19/packages/core/src/lib/stream_manager/stream_manager.ts#L69

Can we feedback to nim-libp2p that their API could be improved? cc @richard-ramos

gabrielmer: > is there a way to esnure we are **not** establishing a new connection?
> 
> JS and Rust libp2p implementations allow you to grab streams out of a connection, for example:
> 
> https://github.com/waku-org/js-waku/blob/cd1d909de38bfecb0543d0642feb219804119e19/packages/core/src/lib/stream_manager/stream_manager.ts#L69
> 
> Can we feedback to nim-libp2p that their API could be improved? cc @richard-ramos

Yes! So we actually checked that last week, when dialing libp2p eventualy calls

https://github.com/vacp2p/nim-libp2p/blob/2f9c3fb3e213cf26888204f8efd6f0e88f7d5ff1/libp2p/dialer.nim#L172-L179

which has the parameter `reuseConnection` set to true by default

waku/node/waku_node.nim
gabrielmer: Ooh great question! True, if we are able to negotiate a stream it does mean the peer is reachable and we shouldn't even need to ping. @richard-ramos are we missing anything?

NagyZoltanPeter: Yes, I would like to join @fryorcraken 's observation about re-using connection. Which I think exists in nim-libp2p but why does dial need to re-handshake and adjust codec each time?
Or I maybe missed sthg but from reading the code it comes this conclusion.

waku/node/waku_node.nim
fryorcraken: Yes but what if you were not connected to this peer in the first place?

gabrielmer: > Yes but what if you were not connected to this peer in the first place?

Mmm not sure I understand the issue. If we're connected to the peer, dial uses an existing connection and creates a stream. If we're not connected to the peer, then a new connection is established.

But we run the keep alive only for peers we're already connected to, so in this case it would be the first scenario for all peers we ping.

Maybe I'm missing something?

richard-ramos: > @fryorcraken  Can we feedback to nim-libp2p that their API could be improved? 

yeah, we had a look last week and saw how the naming for the connection/streams can be improved because as they're right now they are confusing both for users of nim-libp2p and for even devs from nim-libp2p. We have planned in our roadmap a connection abstraction refactoring that should fix this, but in the meantime something i could quickly try to offer for next version is to add an alias for the handlers,  something like `type Stream = Connection`, and that way the handlers would look like this
```nim
  proc handle(stream: Stream, proto: string) {.async: (raises: [CancelledError]).} =
```
which is IMO semantically correct!

waku/node/waku_node.nim
richard-ramos: @NagyZoltanPeter, good point wrt dialingPeer returning success meaning that the peer is alive. Indeed, just doing the dial should be enough due to the handshake that happens. It's important however to make sure there's a timeout in the dialPeer function of some seconds.

Regarding reusing connection, by connection here you mean the actual connection to the peer, or a stream created with that connection? 

if you mean the connection to the remote, then dial should reuse a connection if it has been already established, and obtain a new stream out of it. 

But if you mean reuse the stream, then we'd need to write something to the stream, and make sure that the remote node ACKs by giving you a response, because just writing ain't enough as the data you write goes into the OS send buffer, and this does not block as long as there's room in it, and it doesn't even care that the remote host is even alive.

gabrielmer: @richard-ramos thank you! So just to confirm, do you think it would be ok for the keep alive to not ping and just dial the peer with some timeout?

NagyZoltanPeter: > @NagyZoltanPeter, good point wrt dialingPeer returning success meaning that the peer is alive. Indeed, just doing the dial should be enough due to the handshake that happens. It's important however to make sure there's a timeout in the dialPeer function of some seconds.
> 
> Regarding reusing connection, by connection here you mean the actual connection to the peer, or a stream created with that connection?
> 
> if you mean the connection to the remote, then dial should reuse a connection if it has been already established, and obtain a new stream out of it.
> 
> But if you mean reuse the stream, then we'd need to write something to the stream, and make sure that the remote node ACKs by giving you a response, because just writing ain't enough as the data you write goes into the OS send buffer, and this does not block as long as there's room in it, and it doesn't even care that the remote host is even alive.

Thank you @richard-ramos for confirming.
Rolling fwd, from waku pov we would need such option to get a last known open stream/conn for a codec to a peer without handsake. This was the issue with filter-push we faced - we made a dial for every peer and every message we pushed and did not worked well with js-libp2p cc: @Ivansete-status

richard-ramos: > @gabrielmer  So just to confirm, do you think it would be ok for the keep alive to not ping and just dial the peer with some timeout?

I think so, yes, but still we should confirm how does ping behave on the receiving side tho! does the handler for ping gets called even if you don't send any data? what happens once it closes?

waku/node/waku_node.nim
fryorcraken: > If we're not connected to the peer, then a new connection is established.

> But we run the keep alive only for peers we're already connected to, so in this case it would be the first scenario for all peers we ping.

That's fair enough but again, the issue is the nim-libp2p API because the fact that you "only run keep alive for alraedy connected peers" context is lost once you are in `pingPeer`. Which means it's prone to bugs if you make a mistake upstream, it will establish a connection.

`pingToConnectedPeer` should be the actual behaviour you want to implement (but cannot? because of libp2p API) so that even if upstream, a mistake is made and a call of `pingToConnectedPeer` to a non-connected peer is done, then an error is log but no extra connection is made.

tl;dr: you want to implement a function that if called on a non-connected peer, will just fail, instead of a function that if  called on a non-connected peer, establishes a connection.

waku/node/waku_node.nim
gabrielmer: Makes sense! I think that to not enter this rabbit hole and stick to the known for now, I'll merge it using the pings. Will create the issue to revisit this point as an optimization :)

update: opened https://github.com/waku-org/nwaku/issues/3476

fryorcraken:
> In this PR, the keep-alive configuration option is removed.

Yes!

PR #3459, opened 2025-06-18
fix: remove echo from sync tests

SionoiS:
Removing left over `echo` from sync tests.


Ivansete-status (approved):
LGTM! Thanks so much! :100:

PR #3460, opened 2025-06-19
feat: libwaku dll for status go 

darshankabariya:
Added support for building dynamic libraries (*.dll) on Windows systems.

Now, both libwaku.dll and wakunode2.exe can be generated using ./script/build_windows.sh.

Also added CI checks to verify the successful build of libwaku.dll.
@gabrielmer

### DIFF
--- a/library/events/json_waku_not_responding_event.nim
+++ b/library/events/json_waku_not_responding_event.nim
@@ -1,0 +1,9 @@
+import system, std/json, ./json_base_event
+
+type JsonWakuNotRespondingEvent* = ref object of JsonEvent
+
+proc new*(T: type JsonWakuNotRespondingEvent): T =
+  return JsonWakuNotRespondingEvent(eventType: "waku_not_responding")
+
+method `$`*(event: JsonWakuNotRespondingEvent): string =
+  $(%*event)
--- a/library/libwaku.h
+++ b/library/libwaku.h
@@ -45,6 +45,8 @@ int waku_version(void* ctx,
                  WakuCallBack callback,
                  void* userData);
 
+// Sets a callback that will be invoked whenever an event occurs.
+// It is crucial that the passed callback is fast, non-blocking and potentially thread-safe.
 void waku_set_event_callback(void* ctx,
                              WakuCallBack callback,
                              void* userData);
--- a/library/libwaku.nim
+++ b/library/libwaku.nim
@@ -15,8 +15,7 @@ import
   waku/waku_core/topics/pubsub_topic,
   waku/waku_core/subscription/push_handler,
   waku/waku_relay,
-  ./events/
-    [json_message_event, json_topic_health_change_event, json_connection_change_event],
+  ./events/json_message_event,
   ./waku_thread/waku_thread,
   ./waku_thread/inter_thread_communication/requests/node_lifecycle_request,
   ./waku_thread/inter_thread_communication/requests/peer_manager_request,
@@ -48,25 +47,6 @@ template checkLibwakuParams*(
   if isNil(callback):
     return RET_MISSING_CALLBACK
 
-template callEventCallback(ctx: ptr WakuContext, eventName: string, body: untyped) =
-  if isNil(ctx[].eventCallback):
-    error eventName & " - eventCallback is nil"
-    return
-
-  foreignThreadGc:
-    try:
-      let event = body
-      cast[WakuCallBack](ctx[].eventCallback)(
-        RET_OK, unsafeAddr event[0], cast[csize_t](len(event)), ctx[].eventUserData
-      )
-    except Exception, CatchableError:
-      let msg =
-        "Exception " & eventName & " when calling 'eventCallBack': " &
-        getCurrentExceptionMsg()
-      cast[WakuCallBack](ctx[].eventCallback)(
-        RET_ERR, unsafeAddr msg[0], cast[csize_t](len(msg)), ctx[].eventUserData
-      )
-
 proc handleRequest(
     ctx: ptr WakuContext,
     requestType: RequestType,
@@ -80,21 +60,6 @@ proc handleRequest(
     return RET_ERR
 
   return RET_OK
-
-proc onConnectionChange(ctx: ptr WakuContext): ConnectionChangeHandler =
-  return proc(peerId: PeerId, peerEvent: PeerEventKind) {.async.} =
-    callEventCallback(ctx, "onConnectionChange"):
-      $JsonConnectionChangeEvent.new($peerId, peerEvent)
-
-proc onReceivedMessage(ctx: ptr WakuContext): WakuRelayHandler =
-  return proc(pubsubTopic: PubsubTopic, msg: WakuMessage) {.async.} =
-    callEventCallback(ctx, "onReceivedMessage"):
-      $JsonMessageEvent.new(pubsubTopic, msg)
-
-proc onTopicHealthChange(ctx: ptr WakuContext): TopicHealthChangeHandler =
-  return proc(pubsubTopic: PubsubTopic, topicHealth: TopicHealth) {.async.} =
-    callEventCallback(ctx, "onTopicHealthChange"):
-      $JsonTopicHealthChangeEvent.new(pubsubTopic, topicHealth)
 
 ### End of not-exported components
 ################################################################################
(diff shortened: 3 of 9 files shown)


scripts/build_wakunode_windows.sh
```diff
@@ -49,11 +49,14 @@ cd ../../../../..
...
 cd ./vendor/nim-nat-traversal/vendor/libnatpmp-upstream
-make CC="gcc -fPIC -D_WIN32_WINNT=0x0600 -DNATPMP_STATICLIB" libnatpmp.a V=1
+make CC="gcc -D_WIN32_WINNT=0x0600 -DNATPMP_STATICLIB" libnatpmp.a V=1
 cd ../../../../
 
 echo "9. -.-.-.- Building wakunode2 -.-.-.- "
-execute_command "make wakunode2 LOG_LEVEL=DEBUG V=1 -j8"
+execute_command "make wakunode2 LOG_LEVEL=DEBUG V=1 -j"
+
+echo "9. -.-.-.- Building libwaku -.-.-.- "
```
Ivansete-status:
```suggestion
echo "10. -.-.-.- Building libwaku -.-.-.- "
```

waku.nimble
```diff
@@ -69,9 +70,13 @@ proc buildLibrary(name: string, srcDir = "./", params = "", `type` = "static") =
       ".a --threads:on --app:staticlib --opt:size --noMain --mm:refc --header -d:metrics --nimMainPrefix:libwaku --skipParentCfg:on -d:discv5_protocol_id=d5waku " &
       extra_params & " " & srcDir & name & ".nim"
   else:
-    exec "nim c" & " --out:build/" & name &
-      ".so --threads:on --app:lib --opt:size --noMain --mm:refc --header -d:metrics --nimMainPrefix:libwaku --skipParentCfg:on -d:discv5_protocol_id=d5waku " &
-      extra_params & " " & srcDir & name & ".nim"
+    when defined(windows):
```
Ivansete-status, line 74: It seems the unix-based OSs are forgotten somehow

waku.nimble
```diff
@@ -164,28 +169,14 @@ task libwakuStatic, "Build the cbindings waku node library":
...
     "library/",
-    """-d:chronicles_line_numbers \
-       -d:chronicles_runtime_filtering=on \
-       -d:chronicles_sinks="textlines,json" \
-       -d:chronicles_default_output_device=Dynamic \
-       -d:chronicles_disabled_topics="eth,dnsdisc.client" \
-       --warning:Deprecated:off \
-       --warning:UnusedImport:on \
-       -d:chronicles_log_level=TRACE """,
+    """-d:chronicles_line_numbers -d:chronicles_runtime_filtering=on -d:chronicles_sinks="textlines,json" -d:chronicles_default_output_device=Dynamic -d:chronicles_disabled_topics="eth,dnsdisc.client" --warning:Deprecated:off --warning:UnusedImport:on -d:chronicles_log_level=TRACE """,
```
Ivansete-status: Let's keep it vertical :)
Easier to read and easier to see diffs on each line in future PRs
Also applicable to the next change

darshankabariya: Thanks for comment @Ivansete-status, I’ve addressed all your comments.

Makefile
```diff
@@ -401,14 +401,16 @@ docker-liteprotocoltester-push:
...
 libwaku: | build deps librln
-		rm -f build/libwaku*
+	rm -f build/libwaku*
+
 ifeq ($(STATIC), 1)
-		echo -e $(BUILD_MSG) "build/$@.a" && \
-		$(ENV_SCRIPT) nim libwakuStatic $(NIM_PARAMS) waku.nims
+	echo -e $(BUILD_MSG) "build/$@.a" && $(ENV_SCRIPT) nim libwakuStatic $(NIM_PARAMS) waku.nims
+else ifeq ($(detected_OS),Windows)
+	echo -e $(BUILD_MSG) "build/$@.dll" && $(ENV_SCRIPT) nim libwakuDLL $(NIM_PARAMS) waku.nims
```
NagyZoltanPeter: I would rather make separation by target platfom in one place of possible.
For me it sounds that shouldn't be different command in the makefile to build dyn lib in different platforms.
I would move it to nimble better.
But maybe it is just my taste :-)

waku.nimble
```diff
@@ -68,10 +69,17 @@ proc buildLibrary(name: string, srcDir = "./", params = "", `type` = "static") =
...
       ".a --threads:on --app:staticlib --opt:size --noMain --mm:refc --header -d:metrics --nimMainPrefix:libwaku --skipParentCfg:on -d:discv5_protocol_id=d5waku " &
       extra_params & " " & srcDir & name & ".nim"
-  else:
+  elif `type` == "dynamic":
     exec "nim c" & " --out:build/" & name &
       ".so --threads:on --app:lib --opt:size --noMain --mm:refc --header -d:metrics --nimMainPrefix:libwaku --skipParentCfg:on -d:discv5_protocol_id=d5waku " &
       extra_params & " " & srcDir & name & ".nim"
+  elif `type` == "dll":
+    exec "nim c" & " --out:build/" & name &
+      ".dll --nimcache:build/nimcache --threads:on --app:lib --opt:size --noMain --mm:refc --header -d:metrics --nimMainPrefix:libwaku -d:discv5_protocol_id=d5waku " &
```
NagyZoltanPeter: nimscript has support for properly add extension to target filename: 
`toDll` and `toExe`
https://github.com/nim-lang/Nim/blob/f7145dd26efeeeb6eeae6fff649db244d81b212d/lib/system/nimscript.nim#L147

Using these can simplify nimble code imho.
But beware `toDll` will result `lib`+name+ext... so maybe we need to replicate it for our purpose.

waku.nimble
darshankabariya: I’m getting a unrealistic error with vertical param, I can’t debug yet, but I’m working on it.

NagyZoltanPeter: - @darshankabariya Isn't that because of the multiline string literal by `"""` - triple quote?

darshankabariya: Yeah, it was a spacing issue. I’ve fixed it using string concatenation.

waku.nimble
darshankabariya: Thanks, @NagyZoltanPeter — I’ve addressed this suggestion.

Makefile
darshankabariya: Got it. I wasn’t sure it would work without OS checks since we need to echo .dll or .so. But your idea helped — we can drop libwakuDLL, libwakuDynamic can handle both libs.

waku.nimble
```diff
@@ -69,9 +70,15 @@ proc buildLibrary(name: string, srcDir = "./", params = "", `type` = "static") =
...
-      extra_params & " " & srcDir & name & ".nim"
+    var lib_name = toDll("libwaku")
+    when defined(windows):
+      exec "nim c" & " --out:build/" & lib_name &
+        ".so --threads:on --app:lib --opt:size --noMain --mm:refc --header -d:metrics --nimMainPrefix:libwaku --skipParentCfg:off -d:discv5_protocol_id=d5waku " &
+        extra_params & " " & srcDir & name & ".nim"
+    else:
+      exec "nim c" & " --out:build/" & lib_name &
+        ".so --threads:on --app:lib --opt:size --noMain --mm:refc --header -d:metrics --nimMainPrefix:libwaku --skipParentCfg:on -d:discv5_protocol_id=d5waku " &
+        extra_params & " " & srcDir & name & ".nim"
```
NagyZoltanPeter: warn: Isn't `toDll("libwaku")` will return `liblibwaku.so` or `liblibwaku.dll`?
Well also if we still need to branch by platform due to `--skipParentCfg`... I'm questioning my original reasoning here. But it's hard to recognise at first sight. Btw, why is this distinction?

darshankabariya: Thanks for pointing out the distinction with --skipParentCfg. On Windows, building the binary without external requirements is a bit tricky. Once we fully integrate windows support into the make command, we can revisit and properly address this. For now, I’ve removed --skipParentCfg specifically for the Windows build. cc (@NagyZoltanPeter)

darshankabariya: about warn. is will create libwaku.dll or libwaku.so accordingly

PR #3466, opened 2025-06-23
chore: Libwaku watchdog that can potentially raise a WakuNotResponding event if Waku is blocked

Ivansete-status:
## Description

This PR is inspired by a conversation with @NagyZoltanPeter :raised_hands: 

Libwaku implements an event-driven approach, based on callbacks that are passed.

The Waku SDK integrator will implement his/her own callback and hence, given that each callback will be actually invoked by _The Waku Thread_, we incur in the risk of blocking the _Waku Thread_ if the callback's body is not well implemented.

Therefore, this PR suggests the creation of a new possible event: _waku_not_responding_, which will be raised if the _Waku Thread_ is blocked somewhere, either at the integrator's codebase, or in the nwaku's codebase.

To achieve that, a new thread (watchdog) is created, which is aimed to be super simple.

(The idea is also inspired by a typical UI app, such as Discord, or any Android app. In Android, if the main UI thread is blocked for too long, then an ANR event is dispatched.)

## Issue

### DIFF
--- a/library/libwaku.h
+++ b/library/libwaku.h
@@ -46,7 +46,7 @@ int waku_version(void* ctx,
                  void* userData);
 
 // Sets a callback that will be invoked whenever an event occurs.
-// It is crucial that the passed callback is fast and non-blocking.
+// It is crucial that the passed callback is fast, non-blocking and potentially thread-safe.
 void waku_set_event_callback(void* ctx,
                              WakuCallBack callback,
                              void* userData);


library/events/json_topic_health_change_event.nim
```diff
@@ -7,6 +7,10 @@ type JsonTopicHealthChangeEvent* = ref object of JsonEvent
   pubsubTopic*: string
   topicHealth*: TopicHealth
 
+type JsonWakuHealthChangeEvent* = ref object of JsonEvent
+  pubsubTopic*: string
+  topicHealth*: TopicHealth
```
gabrielmer: what is this for?

Ivansete-status: Not needed, thanks! Addressed in https://github.com/waku-org/nwaku/pull/3466/commits/68b86616bfad45c73f4c718b2e8bfd081ce9ebe9

NagyZoltanPeter (commented):
Thanks for addressing this. 
I would roll fwd with a question:
- is the watchdog thread will call presumably the same evenCallback that we think might blocking our nim-waku-thread?
It will not interfare with the blocking code for sure as happening on a different (watchdog) thread, but assumes that eventcallback function is thread safe.
Can we assume that?
So to say, my assumption, in normal use-case event-callback need not to be thread safe as we will call it from one single thread - the nim-waku-thread (assuming, the user did not create more waku instances :-o).
But with watchdog... IDK how to warn the users of library to take care of thread safety within eventcallback.
WDYT can this be a problem for now or we can address it later when real issue arise? I'm keen on approving this PR as it is a great idea.

Ivansete-status:
Very wise comments @NagyZoltanPeter ! Thanks for them !
I think you are right about this risk. Nevertheless, this PR still remains pretty valid thanks to the following points:
1. If the event_callback is super slow, then the SDK integrator will get eventually notified, after a while
2. There is still a risk that the event_callback hangs forever the watchdog thread. However, we will mitigate that on higher `WakuSDK` layers and in the `Waku SDK bindings`

NagyZoltanPeter:
> Very wise comments @NagyZoltanPeter ! Thanks for them ! I think you are right about this risk. Nevertheless, this PR still remains pretty valid thanks to the following points:
> 
> 1. If the event_callback is super slow, then the SDK integrator will get eventually notified, after a while
> 2. There is still a risk that the event_callback hangs forever the watchdog thread. However, we will mitigate that on higher `WakuSDK` layers and in the `Waku SDK bindings`

Yeah, I think it is useful!
Maybe some idea to tackle properly... maybe in another PR.
- Allow only one waku instance at a time.
    - reason is to be fool proof for the future. It is likely to misuse the API, creating several nodes while giving the same eventCallback for all... thus requires to be thread safe... hard one.
 - require to provide different eventCallback fn for watchdog notification than waku-node eventCallback!
     - this prevents badly written event handler to crash due to race condition. Anyway we will be blamed if crash happens regardless of badly written eventCallback... so this is to save us from hard debugging :-)

NagyZoltanPeter (approved):
Thanks for the great initiative!!!

library/libwaku.h
```diff
@@ -45,6 +45,8 @@ int waku_version(void* ctx,
                  WakuCallBack callback,
                  void* userData);
 
+// Sets a callback that will be invoked whenever an event occurs.
+// It is crucial that the passed callback is fast and non-blocking.
```
NagyZoltanPeter:
```suggestion
// It is crucial that the passed callback is fast, non-blocking and potentially thread-safe.
```

NagyZoltanPeter (approved):
LGTM.

PR #3468, opened 2025-06-24
chore!: make sharding configuration explicit

fryorcraken:
# Description

Make it explicit that the node can either be configured with _auto sharding_ or _static sharding_. Treat static sharding as an happy path (prev. was implicit error or do-nothing-happy).
Make it clear that without auto-sharding, pubsub topics can't be derived from content topics.

Do so by improving the internal node configuration and make it clear that there are two options, and that the _number of shards in the network_ is only needed for auto sharding.

Also proceed with a number of naming improvement, as per https://github.com/waku-org/specs/pull/65/

- Cluster Conf is actually a (Waku) network conf - as previously mentioned by @ivansete-status in other pr
- tcp/ip "network" conf is renamed to "endpoint conf" to avoid name clashing

Some refactoring done along the way.

# Changes

- [x] Auto sharding is enabled by default, with `1` shard in network
- [x] new `shards` (or `subscribeShards`) defaults:
  - auto sharding: all shards
  - static sharding: no shards (because we cannot guess the shard numbers)
- [x] cluster conf renamed to network conf
- [x] sharding conf is made explicit static vs auto, number of shards in network is only useful for autosharding
- [x] conf `shards` is actually the shards we want relay to subscribe too, so rename accordingly
- [x] use `uint16` for cluster id and shard id more (but not totally) consistently across the code base.
- [x] move `WebSocketConf` to the right place 

# Follow-up changes

- probably remove `activeRelayShards` in config, in favour of having `wakunode2` just automatically subscribe to the shards passed on the CLI
- Number of TODOs left in code
- define a proper type for cluster id and shard ids

## Issue

- https://github.com/waku-org/nwaku/issues/3349
- https://github.com/waku-org/nwaku/issues/3493

### DIFF
--- a/tests/factory/test_external_config.nim
+++ b/tests/factory/test_external_config.nim
@@ -48,7 +48,9 @@ suite "Waku config - apply preset":
       check rlnRelayConf.chainId == expectedConf.rlnRelayChainId
       check rlnRelayConf.epochSizeSec == expectedConf.rlnEpochSizeSec
       check rlnRelayConf.userMessageLimit == expectedConf.rlnRelayUserMessageLimit
-    check conf.numShardsInNetwork == expectedConf.numShardsInNetwork
+      check conf.shardingConf.kind == expectedConf.shardingConf.kind
+      check conf.shardingConf.numShardsInCluster ==
+        expectedConf.shardingConf.numShardsInCluster
     check conf.discv5Conf.isSome() == expectedConf.discv5Discovery
     if conf.discv5Conf.isSome():
       let discv5Conf = conf.discv5Conf.get()
@@ -68,7 +70,7 @@ suite "Waku config - apply preset":
 
     ## Then
     let conf = res.get()
-    check conf.shards.len == expectedConf.numShardsInNetwork.int
+    check conf.activeRelayShards.len == expectedConf.shardingConf.numShardsInCluster.int
 
   test "Subscribes to some valid shards in twn":
     ## Setup
@@ -84,9 +86,9 @@ suite "Waku config - apply preset":
 
     ## Then
     let conf = resConf.get()
-    assert conf.shards.len() == shards.len()
+    assert conf.activeRelayShards.len() == shards.len()
     for index, shard in shards:
-      assert shard in conf.shards
+      assert shard in conf.activeRelayShards
 
   test "Subscribes to invalid shards in twn":
     ## Setup
@@ -131,7 +133,9 @@ suite "Waku config - apply preset":
       check rlnRelayConf.chainId == expectedConf.rlnRelayChainId
       check rlnRelayConf.epochSizeSec == expectedConf.rlnEpochSizeSec
       check rlnRelayConf.userMessageLimit == expectedConf.rlnRelayUserMessageLimit
-    check conf.numShardsInNetwork == expectedConf.numShardsInNetwork
+      check conf.shardingConf.kind == expectedConf.shardingConf.kind
+      check conf.shardingConf.numShardsInCluster ==
+        expectedConf.shardingConf.numShardsInCluster
     check conf.discv5Conf.isSome() == expectedConf.discv5Discovery
     if conf.discv5Conf.isSome():
       let discv5Conf = conf.discv5Conf.get()
@@ -164,7 +168,7 @@ suite "Waku config - Shards":
 
     ## Given
     let shards: seq[uint16] = @[0, 2, 4]
-    let numShardsInNetwork = 5.uint32
+    let numShardsInNetwork = 5.uint16
     let wakuNodeConf = WakuNodeConf(
       cmd: noCommand, shards: shards, numShardsInNetwork: numShardsInNetwork
     )
@@ -183,7 +187,7 @@ suite "Waku config - Shards":
 
     ## Given
     let shards: seq[uint16] = @[0, 2, 5]
-    let numShardsInNetwork = 5.uint32
+    let numShardsInNetwork = 5.uint16
     let wakuNodeConf = WakuNodeConf(
       cmd: noCommand, shards: shards, numShardsInNetwork: numShardsInNetwork
     )
--- a/tests/factory/test_external_config.nim
+++ b/tests/factory/test_external_config.nim
@@ -20,7 +20,7 @@ import
 suite "Waku config - apply preset":
   test "Default preset is TWN":
     ## Setup
-    let expectedConf = ClusterConf.TheWakuNetworkConf()
+    let expectedConf = NetworkConf.TheWakuNetworkConf()
 
     ## Given
     let preConfig = WakuNodeConf(
@@ -56,7 +56,7 @@ suite "Waku config - apply preset":
 
   test "Subscribes to all valid shards in twn":
     ## Setup
-    let expectedConf = ClusterConf.TheWakuNetworkConf()
+    let expectedConf = NetworkConf.TheWakuNetworkConf()
 
     ## Given
     let shards: seq[uint16] = @[0, 1, 2, 3, 4, 5, 6, 7]
@@ -72,7 +72,7 @@ suite "Waku config - apply preset":
 
   test "Subscribes to some valid shards in twn":
     ## Setup
-    let expectedConf = ClusterConf.TheWakuNetworkConf()
+    let expectedConf = NetworkConf.TheWakuNetworkConf()
 
     ## Given
     let shards: seq[uint16] = @[0, 4, 7]
@@ -103,7 +103,7 @@ suite "Waku config - apply preset":
 
   test "Apply TWN preset when cluster id = 1":
     ## Setup
-    let expectedConf = ClusterConf.TheWakuNetworkConf()
+    let expectedConf = NetworkConf.TheWakuNetworkConf()
 
     ## Given
     let preConfig = WakuNodeConf(
--- a/tests/factory/test_external_config.nim
+++ b/tests/factory/test_external_config.nim
@@ -70,7 +70,7 @@ suite "Waku config - apply preset":
 
     ## Then
     let conf = res.get()
-    check conf.activeRelayShards.len == expectedConf.shardingConf.numShardsInCluster.int
+    check conf.subscribeShards.len == expectedConf.shardingConf.numShardsInCluster.int
 
   test "Subscribes to some valid shards in twn":
     ## Setup
@@ -86,9 +86,9 @@ suite "Waku config - apply preset":
 
     ## Then
     let conf = resConf.get()
-    assert conf.activeRelayShards.len() == shards.len()
+    assert conf.subscribeShards.len() == shards.len()
     for index, shard in shards:
-      assert shard in conf.activeRelayShards
+      assert shard in conf.subscribeShards
 
   test "Subscribes to invalid shards in twn":
     ## Setup
--- a/tests/factory/test_external_config.nim
+++ b/tests/factory/test_external_config.nim
@@ -17,8 +17,44 @@ import
   ../../waku/common/logging,
   ../../waku/common/utils/parse_size_units
 
-suite "Waku config - apply preset":
-  test "Default preset is TWN":
+suite "Waku external config - default values":
+  test "Default sharding value":
+    ## Setup
+    let defaultShardingMode = AutoSharding
+    let defaultNumShardsInCluster = 1.uint16
+    let defaultSubscribeShards = @[0.uint16]
+
+    ## Given
+    let preConfig = defaultWakuNodeConf().get()
+
+    ## When
+    let res = preConfig.toWakuConf()
+    assert res.isOk(), $res.error
+
+    ## Then
+    let conf = res.get()
+    check conf.shardingConf.kind == defaultShardingMode
+    check conf.shardingConf.numShardsInCluster == defaultNumShardsInCluster
+    check conf.subscribeShards == defaultSubscribeShards
+
+  test "Default shards value in static sharding":
+    ## Setup
+    let defaultSubscribeShards: seq[uint16] = @[]
+
+    ## Given
+    var preConfig = defaultWakuNodeConf().get()
+    preConfig.numShardsInNetwork = 0.uint16
+
+    ## When
+    let res = preConfig.toWakuConf()
+    assert res.isOk(), $res.error
+
+    ## Then
+    let conf = res.get()
+    check conf.subscribeShards == defaultSubscribeShards
+
+suite "Waku external config - apply preset":
+  test "Preset is TWN":
     ## Setup
     let expectedConf = NetworkConf.TheWakuNetworkConf()
 
@@ -141,7 +177,7 @@ suite "Waku config - apply preset":
       let discv5Conf = conf.discv5Conf.get()
       check discv5Conf.bootstrapNodes == expectedConf.discv5BootstrapNodes
 
-suite "Waku config - node key":
+suite "Waku external config - node key":
   test "Passed node key is used":
     ## Setup
     let nodeKeyStr =
@@ -162,7 +198,7 @@ suite "Waku config - node key":
     assert utils.toHex(resKey.getRawBytes().get()) ==
       utils.toHex(nodekey.getRawBytes().get())
 
-suite "Waku config - Shards":
+suite "Waku external config - Shards":
   test "Shards are valid":
     ## Setup
 
@@ -202,7 +238,7 @@ suite "Waku config - Shards":
     ## Setup
 
     ## Given
-    let wakuNodeConf = WakuNodeConf.load(version = "", cmdLine = @["--shard=32"])
+    let wakuNodeConf = WakuNodeConf.load(version = "", cmdLine = @["--shard=0"])
 
     ## When
     let res = wakuNodeConf.toWakuConf()
@@ -211,3 +247,15 @@ suite "Waku config - Shards":
     let wakuConf = res.get()
     let vRes = wakuConf.validate()
     assert vRes.isOk(), $vRes.error
+
+  test "Imvalid shard is passed without num shards":
+    ## Setup
+
+    ## Given
+    let wakuNodeConf = WakuNodeConf.load(version = "", cmdLine = @["--shard=32"])
+
+    ## When
+    let res = wakuNodeConf.toWakuConf()
+
+    ## Then
+    assert res.isErr(), "Invalid shard was accepted"
--- a/waku/factory/conf_builder/waku_conf_builder.nim
+++ b/waku/factory/conf_builder/waku_conf_builder.nim
@@ -61,7 +61,7 @@ type WakuConfBuilder* = object
   clusterId: Option[uint16]
   shardingConf: Option[ShardingConfKind]
   numShardsInCluster: Option[uint16]
-  activeRelayShards: Option[seq[uint16]]
+  subscribeShards: Option[seq[uint16]]
   protectedShards: Option[seq[ProtectedShard]]
   contentTopics: Option[seq[string]]
 
@@ -151,8 +151,8 @@ proc withShardingConf*(b: var WakuConfBuilder, shardingConf: ShardingConfKind) =
 proc withNumShardsInCluster*(b: var WakuConfBuilder, numShardsInCluster: uint16) =
   b.numShardsInCluster = some(numShardsInCluster)
 
-proc withActiveRelayShards*(b: var WakuConfBuilder, shards: seq[uint16]) =
-  b.activeRelayShards = some(shards)
+proc withSubscribeShards*(b: var WakuConfBuilder, shards: seq[uint16]) =
+  b.subscribeShards = some(shards)
 
 proc withProtectedShards*(
     b: var WakuConfBuilder, protectedShards: seq[ProtectedShard]
@@ -445,9 +445,9 @@ proc build*(
       warn("No sharding conf specified, defaulting to static")
       ShardingConf(kind: StaticSharding)
 
-  let activeRelayShards =
-    if builder.activeRelayShards.isSome():
-      builder.activeRelayShards.get()
+  let subscribeShards =
+    if builder.subscribeShards.isSome():
+      builder.subscribeShards.get()
     else:
       case shardingConf.kind
       of AutoSharding:
@@ -617,7 +617,7 @@ proc build*(
     clusterId: clusterId,
     shardingConf: shardingConf,
     contentTopics: contentTopics,
-    activeRelayShards: activeRelayShards,
+    subscribeShards: subscribeShards,
     protectedShards: protectedShards,
     relay: relay,
     lightPush: lightPush,
--- a/waku/factory/conf_builder/waku_conf_builder.nim
+++ b/waku/factory/conf_builder/waku_conf_builder.nim
@@ -352,10 +352,10 @@ proc applyNetworkConf(builder: var WakuConfBuilder) =
       discarded = builder.numShardsInCluster
 
   case networkConf.shardingConf.kind
-  of Static:
-    builder.shardingConf = some(Static)
-  of Auto:
-    builder.shardingConf = some(Auto)
+  of StaticSharding:
+    builder.shardingConf = some(StaticSharding)
+  of AutoSharding:
+    builder.shardingConf = some(AutoSharding)
     builder.numShardsInCluster = some(networkConf.shardingConf.numShardsInCluster)
 
   if networkConf.discv5Discovery:
@@ -430,25 +430,27 @@ proc build*(
   let shardingConf =
     if builder.shardingConf.isSome():
       case builder.shardingConf.get()
-      of Static:
-        ShardingConf(kind: Static)
-      of Auto:
+      of StaticSharding:
+        ShardingConf(kind: StaticSharding)
+      of AutoSharding:
         if builder.numShardsInCluster.isSome():
-          ShardingConf(kind: Auto, numShardsInCluster: builder.numShardsInCluster.get())
+          ShardingConf(
+            kind: AutoSharding, numShardsInCluster: builder.numShardsInCluster.get()
+          )
         else:
           return err(
             "Number of shards in cluster must be specified with autosharding enabled"
           )
     else:
       warn("No sharding conf specified, defaulting to static")
-      ShardingConf(kind: Static)
+      ShardingConf(kind: StaticSharding)
 
   let activeRelayShards =
     if builder.activeRelayShards.isSome():
       builder.activeRelayShards.get()
     else:
       case shardingConf.kind
-      of Auto:
+      of AutoSharding:
         warn "Active relay shards not specified and autosharding is enabled, defaulting to all shards in network"
         let upperShard: uint16 = uint16(shardingConf.numShardsInCluster - 1)
         toSeq(0.uint16 .. upperShard)
--- a/waku/factory/conf_builder/waku_conf_builder.nim
+++ b/waku/factory/conf_builder/waku_conf_builder.nim
@@ -273,6 +273,8 @@ proc withMaxMessageSize*(builder: var WakuConfBuilder, maxMessageSize: string) =
 proc withStaticNodes*(builder: var WakuConfBuilder, staticNodes: seq[string]) =
   builder.staticNodes = concat(builder.staticNodes, staticNodes)
 
+## Building
+
 proc nodeKey(
     builder: WakuConfBuilder, rng: ref HmacDrbgContext
 ): Result[crypto.PrivateKey, string] =
@@ -284,6 +286,22 @@ proc nodeKey(
       error "Failed to generate key", error = error
       return err("Failed to generate key: " & $error)
     return ok(nodeKey)
+
+proc buildShardingConf(
+    bShardingConfKind: Option[ShardingConfKind],
+    bNumShardsInCluster: Option[uint16],
+    bSubscribeShards: Option[seq[uint16]],
+): (ShardingConf, seq[uint16]) =
+  echo "bSubscribeShards: ", bSubscribeShards
+  case bShardingConfKind.get(AutoSharding)
+  of StaticSharding:
+    (ShardingConf(kind: StaticSharding), bSubscribeShards.get(@[]))
+  of AutoSharding:
+    let numShardsInCluster = bNumShardsInCluster.get(1)
+    let shardingConf =
+      ShardingConf(kind: AutoSharding, numShardsInCluster: numShardsInCluster)
+    let upperShard = uint16(numShardsInCluster - 1)
+    (shardingConf, bSubscribeShards.get(toSeq(0.uint16 .. upperShard)))
 
 proc applyNetworkConf(builder: var WakuConfBuilder) =
   # Apply network conf, overrides most values passed individually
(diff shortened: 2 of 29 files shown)


waku/waku_lightpush/protocol.nim
```diff
@@ -26,92 +26,90 @@ type WakuLightPush* = ref object of LPProtocol
...
+      statusDesc: some(desc),
     )
-  else:
-    let pushRequest = reqDecodeRes.get()
-
-    let pubsubTopic = pushRequest.pubSubTopic.valueOr:
-      let parsedTopic = NsContentTopic.parse(pushRequest.message.contentTopic).valueOr:
-        let msg = "Invalid content-topic:" & $error
-        error "lightpush request handling error", error = msg
-        return LightpushResponse(
```
fryorcraken: several `return` in the error path skipped the metrics report!

waku/waku_lightpush/protocol.nim
```diff
@@ -26,92 +26,90 @@ type WakuLightPush* = ref object of LPProtocol
...
+    return ok(handleRes.get())
+
+  return err((code: handleRes.error.code, desc: handleRes.error.desc))
 
 proc handleRequest*(
     wl: WakuLightPush, peerId: PeerId, buffer: seq[byte]
 ): Future[LightPushResponse] {.async.} =
-  let reqDecodeRes = LightpushRequest.decode(buffer)
-  var isSuccess = false
   var pushResponse: LightpushResponse
```
fryorcraken: Storing results in a `var` is usually a bad smell because bigger nesting is needed (instead of just exiting on an error).
Bigger nesting = less readable code.
Proof-in-pudding: some `return` were including in nesting leading to metrics being missed: https://github.com/waku-org/nwaku/pull/3468/files#r2168086052

waku/waku_lightpush/protocol.nim
```diff
@@ -26,92 +26,90 @@ type WakuLightPush* = ref object of LPProtocol
   peerManager*: PeerManager
   pushHandler*: PushMessageHandler
   requestRateLimiter*: RequestRateLimiter
-  sharding: Sharding
+  autoSharding: Option[Sharding]
+
+proc handleRequest(
```
fryorcraken, line 31: Reviewers are free to request for this refactoring to be done in a separate PR.

waku/factory/internal_config.nim
```diff
@@ -143,11 +137,3 @@ proc networkConfiguration*(
   )
 
   return netConfigRes
-
-# TODO: numShardsInNetwork should be mandatory with autosharding, and unneeded otherwise
```
fryorcraken, line 147: TODO Done!

waku/factory/waku_conf.nim
```diff
@@ -166,26 +173,9 @@ proc validateNodeKey(wakuConf: WakuConf): Result[void, string] =
     return err("nodekey param is invalid")
   return ok()
 
-proc validateShards(wakuConf: WakuConf): Result[void, string] =
-  let numShardsInNetwork = wakuConf.numShardsInNetwork
-
-  # TODO: fix up this behaviour
```
fryorcraken, line 171: Done!

waku/waku_core/topics/sharding.nim
```diff
@@ -50,48 +51,37 @@ proc getShard*(s: Sharding, topic: ContentTopic): Result[RelayShard, string] =
 
   ok(shard)
 
-proc parseSharding*(
-    s: Sharding,
-    pubsubTopic: Option[PubsubTopic],
-    contentTopics: ContentTopic | seq[ContentTopic],
+proc getShardsFromContentTopics*(
```
fryorcraken, line 54: this function was always called when `pubsubTopic` was `none`...

waku/waku_lightpush/protocol.nim
```diff
@@ -26,92 +26,90 @@ type WakuLightPush* = ref object of LPProtocol
...
-    if pubsubTopic.isEmptyOrWhitespace():
-      let msg = "topic must not be empty"
-      error "lightpush request handling error", error = msg
-      return LightPushResponse(
-        requestId: pushRequest.requestId,
-        statusCode: LightpushStatusCode.BAD_REQUEST.uint32,
-        statusDesc: some(msg),
-      )
+  # TODO: is that fine?
+  let res = await handleRequest(wl, peerId, pushRequest)
```
fryorcraken: @NagyZoltanPeter 

Is that ok to call an await here? I am not yet familiar with Nim async and whether there are any gotcha

tests/factory/test_external_config.nim
```diff
@@ -68,11 +70,11 @@ suite "Waku config - apply preset":
 
     ## Then
     let conf = res.get()
-    check conf.shards.len == expectedConf.numShardsInNetwork.int
+    check conf.activeRelayShards.len == expectedConf.shardingConf.numShardsInCluster.int
```
SionoiS: `activeRelayShards` Does that mean we can have inactive ones???

waku/factory/node_factory.nim
```diff
@@ -137,10 +138,12 @@ proc initNode(
...
     node: WakuNode, contentTopics: seq[string]
 ): Result[seq[RelayShard], string] =
+  if node.wakuAutoSharding.isNone():
+    return err("Static sharding used, cannot get shards from content topics")
   var autoShards: seq[RelayShard]
   for contentTopic in contentTopics:
-    let shard = node.wakuSharding.getShard(contentTopic).valueOr:
-      return err("Could not parse content topic: " & error)
+    let shard = node.wakuAutoSharding.get().getShard(contentTopic).valueOr:
+        return err("Could not parse content topic: " & error)
```
SionoiS, line 146: extra tab?

SionoiS (approved):
LGTM

I just hope we didn't mess yet again some edge cases...

I swear this is the 5e time we refactor shard config stuff!

tests/factory/test_external_config.nim
fryorcraken: This is about expressing the fact that those "shards" are really shards we want to subscribe to.

Next step will be to remove that from the config entirely as it's not a node setup matter, but a run time matter.

fryorcraken:
Started to work on fixing interops tests: https://github.com/waku-org/waku-interop-tests/pull/126

fryorcraken:
interop-tests PR https://github.com/waku-org/waku-interop-tests/pull/126 is ready for review.

Once merged, I'll rerun tests on this PR and it should be green.

waku/factory/waku_conf.nim
```diff
@@ -68,11 +76,10 @@ type WakuConf* {.requiresInit.} = ref object
   nodeKey*: crypto.PrivateKey
 
   clusterId*: uint16
-  shards*: seq[uint16]
+  activeRelayShards*: seq[uint16]
```
gabrielmer: There's something I still don't get about `activeRelayShards`. Shouldn't the user configure which shards they want and that's it? And isn't that what it should be in the configuration, the desired shards by the user?

waku/waku_core/topics/sharding.nim
```diff
@@ -50,48 +51,37 @@ proc getShard*(s: Sharding, topic: ContentTopic): Result[RelayShard, string] =
...
+  for content in nsContentTopics:
+    let shardsRes = s.getShard(content)
 
-        if parseRes.isErr():
-          return err("Cannot parse pubsub topic: " & $parseRes.error)
-        else:
-          parseRes.get()
+    let shard =
+      if shardsRes.isErr():
+        return err("Cannot deduce shard from content topic: " & $shardsRes.error)
```
gabrielmer: Maybe this was it's simpler?

```suggestion
    let shard = s.getShard(content).valueOr:
        return err("Cannot deduce shard from content topic: " & $error)
```

(and then removing the `let shardsRes = s.getShard(content)` and the ` else:
        shardsRes.get()`, which GH doesn't let me add in the suggestion)

waku/node/waku_node.nim
```diff
@@ -322,11 +323,15 @@ proc subscribe*(
...
-
-      ($shard, some(subscription.topic))
+      if node.wakuAutoSharding.isSome():
+        let shard = node.wakuAutoSharding.get().getShard((subscription.topic)).valueOr:
+            error "Autosharding error", error = error
+            return err("Autosharding error: " & error)
+        ($shard, some(subscription.topic))
+      else:
+        let msg =
+          "Static sharding is used, relay subscriptions must specify a pubsub topic"
```
gabrielmer: I think this msg variable is not used anywhere, same for the `ContentUnsub`case

waku/waku_lightpush/protocol.nim
gabrielmer: I think it's better yes, the PR is already large enough, better to have this in a separate PR in case there's a bug in the future it will be easier to pinpoint to this

waku/factory/node_factory.nim
```diff
@@ -290,14 +288,22 @@ proc setupProtocols(
 
     peerExchangeHandler = some(handlePeerExchange)
 
-  let autoShards = node.getAutoshards(conf.contentTopics).valueOr:
-    return err("Could not get autoshards: " & error)
+  # TODO: when using autosharding, the user should not be expected to pass any shards, but only content topics
+  # Hence, this joint logic should be removed in favour of an either logic:
+  # use passed shards (static) or deduce shards from content topics (auto)
```
gabrielmer, line 293: Not sure I understand or agree with this, one might subscribe to static shards (for example, default TWN shards) but want to use autosharding to publish messages without having to think about shards in their application's logic. Isn't that legitimate usage?

gabrielmer (commented):
Woow amazing work! Thanks so much!

Added first some questions to further understand some points

waku/factory/waku_conf.nim
fryorcraken: spoiler: I will remove this parameter, as I am not including it in the Waku API.

these are the shards we want to subscribe to via relay when the node starts.

Define "desired shards by the user"

waku/factory/node_factory.nim
fryorcraken: It's the same as https://github.com/waku-org/nwaku/pull/3468#discussion_r2176864130

I think we first need to acknowledge that these are 2 separate use cases:

(1) A developer using nwaku (as a lib or otherwise) for their app. In this case, if they use autosharding then they only want to deal with content topic to subscribe to shards or send messages

(2) A node operator running a node on TWN. Assuming said operator don't want to run a node for all shards. Why would they do that? is it to support their app? in this case, which shard should they subscribe too? 1, 2, 3? How do they know. Well, this is back to step (1). They should just deal with content topics.

waku/factory/waku_conf.nim
gabrielmer: > Define "desired shards by the user"

Exactly this: "the shards we want to subscribe to via relay when the node starts"

When a node is configured and the operator selects the shard they want to operate in. Why call them `activeRelayShards` instead of `shards` if it's the config desire of the user/operator and not a run time issue?

or that is the combination of both the shards specified by the user explicitly + the resolved shards by autosharding?

gabrielmer: ok now I'm understanding better, that if we don't allow mixing autosharding and static sharding then makes sense!

gabrielmer (approved):
Thanks so much! I would wait until @Ivansete-status's review before merging just in case :)

waku/node/waku_node.nim
```diff
@@ -112,7 +112,7 @@ type
     wakuLightpushClient*: WakuLightPushClient
     wakuPeerExchange*: WakuPeerExchange
     wakuMetadata*: WakuMetadata
-    wakuSharding*: Sharding
+    wakuAutoSharding*: Option[Sharding]
```
Ivansete-status, line 115: The `wakuAutoSharding` shouldn't belong to `WakuNode`.
IMHO, this is linked to `wakuRelay` and should be there instead.

waku/factory/conf_builder/waku_conf_builder.nim
```diff
@@ -59,8 +59,9 @@ type WakuConfBuilder* = object
   nodeKey: Option[crypto.PrivateKey]
 
   clusterId: Option[uint16]
-  numShardsInNetwork: Option[uint32]
-  shards: Option[seq[uint16]]
+  shardingConf: Option[ShardingConfKind]
+  numShardsInCluster: Option[uint16]
```
Ivansete-status, line 63: This is a great opportunity to have a good default.
I would not give freedom to select the number of shards in a cluster. No one is going to give a value there with certain decision.
Instead, we need to use the maximum allowed shards within a cluster, which is 1024, according to the spec: https://github.com/waku-org/specs/blob/master/standards/core/relay-sharding.md

waku/factory/conf_builder/waku_conf_builder.nim
```diff
@@ -59,8 +59,9 @@ type WakuConfBuilder* = object
...
+  activeRelayShards: Option[seq[uint16]]
```
Ivansete-status: What are the inactive relay shards :)? I think that's been commented elsewhere: what's wrong with `shards` name?

waku/factory/conf_builder/waku_conf_builder.nim
```diff
@@ -136,20 +137,23 @@ proc init*(T: type WakuConfBuilder): WakuConfBuilder =
...
-  b.numShardsInNetwork = some(numShardsInNetwork)
+proc withShardingConf*(b: var WakuConfBuilder, shardingConf: ShardingConfKind) =
+  b.shardingConf = some(shardingConf)
 
-proc withShards*(b: var WakuConfBuilder, shards: seq[uint16]) =
-  b.shards = some(shards)
+proc withNumShardsInCluster*(b: var WakuConfBuilder, numShardsInCluster: uint16) =
+  b.numShardsInCluster = some(numShardsInCluster)
+
+proc withActiveRelayShards*(b: var WakuConfBuilder, shards: seq[uint16]) =
```
Ivansete-status: I prefer this. More concise ;P
```suggestion
proc withShards*(b: var WakuConfBuilder, shards: seq[uint16]) =
```

waku/factory/conf_builder/waku_conf_builder.nim
```diff
@@ -59,8 +59,9 @@ type WakuConfBuilder* = object
   nodeKey: Option[crypto.PrivateKey]
 
   clusterId: Option[uint16]
-  numShardsInNetwork: Option[uint32]
-  shards: Option[seq[uint16]]
+  shardingConf: Option[ShardingConfKind]
```
Ivansete-status, line 62: According to the spec, Automatic Sharding should be set by default. 
Therefore, this should not be an `Option` and be set to `AutoSharding` by default.

https://github.com/waku-org/specs/blob/master/standards/core/relay-sharding.md

waku/factory/conf_builder/waku_conf_builder.nim
```diff
@@ -282,77 +286,89 @@ proc nodeKey(
...
+    warn "Num Shards In Cluster was provided alongside a network conf",
+      used = networkConf.shardingConf.numShardsInCluster,
+      discarded = builder.numShardsInCluster
+
+  case networkConf.shardingConf.kind
+  of StaticSharding:
+    builder.shardingConf = some(StaticSharding)
+  of AutoSharding:
+    builder.shardingConf = some(AutoSharding)
+    builder.numShardsInCluster = some(networkConf.shardingConf.numShardsInCluster)
```
Ivansete-status, line 377: This shouldn't be needed
```suggestion
```

waku/factory/conf_builder/waku_conf_builder.nim
```diff
@@ -412,21 +428,36 @@ proc build*(
...
+        ShardingConf(kind: StaticSharding)
+      of AutoSharding:
+        if builder.numShardsInCluster.isSome():
+          ShardingConf(
+            kind: AutoSharding, numShardsInCluster: builder.numShardsInCluster.get()
+          )
+        else:
+          return err(
+            "Number of shards in cluster must be specified with autosharding enabled"
+          )
```
Ivansete-status: I think this can be simplified if we apply good default for `numShardsInCluster` ( `1024` ) and avoid making `shardingConf` an `Option`

waku/factory/external_config.nim
```diff
@@ -317,10 +317,11 @@ hence would have reachability issues.""",
     # TODO: This is trying to do too much, this should only be used for autosharding, which itself should be configurable
     # If numShardsInNetwork is not set, we use the number of shards configured as numShardsInNetwork
     numShardsInNetwork* {.
-      desc: "Number of shards in the network",
+      desc:
+        "Enables autosharding and set number of shards in the cluster, set to `0` to use static sharding",
```
Ivansete-status, line 319: As I mentioned before, this is a great opportunity to remove this parameter :D

waku/factory/networks_config.nim
```diff
@@ -1,18 +1,23 @@
...
+    AutoSharding
+    StaticSharding
 
-# TODO: Rename this type to match file name
+  ShardingConf* = object
+    case kind*: ShardingConfKind
+    of AutoSharding:
+      numShardsInCluster*: uint16
+    of StaticSharding:
+      discard
```
Ivansete-status, line 18: This is beautiful indeed :heart: 
Nevertheless, I think we can simplify it by having a good default for `numShardsInCluster` (1024) in which case, the `ShardingConf` wouldn't be needed.
Sorry for repeating this same simplification idea multiple times :D

waku/factory/node_factory.nim
Ivansete-status: I do agree with @gabrielmer in this point.

Autosharding seems as a technique for both distributing the traffic randomly across the available shards and to simplify the publish operation.

It seems that we will always have a combination of both static and automatic sharding. For example, _TWN_, and any other network, would not be able to operate with only automatic sharding, because the content-topic is only available when publishing a message and not when subscribing.

Ivansete-status (commented):

Not approving yet because I didn't review it 100%

I don't want to be a blocker so feel free to merge it if you consider appropriate.
Of course, happy to revisit again in the future if you want to apply any change or split it into smaller PRs :)

Thanks again! :heart:

waku/node/waku_node.nim
fryorcraken: I would disagree here, light push, store and filter also uses the autosharding logic, as pubsub topic are part of the request body and needs to be inferred from the content topics.

waku/factory/conf_builder/waku_conf_builder.nim
fryorcraken: when you start a new network, you want to start with a small number of shards. So that you can have several users on the same shard. If you have no user on a shard, you cannot form a gossipsub network.

1024 shards by default means that you would not form any gossipsub networks, and users would not connect to each other because their application is likely to only connect to one shard.

There is a reason why we selected 8 shards for TWN.

Most new networks would be fine with 1 or 8 shards.

Also, are you thinking a default for service nodes? or for applications?

A default value of `1` may be better, for applications that don't want to use TWN.

waku/factory/conf_builder/waku_conf_builder.nim
fryorcraken: because `shards` means nothing. I guess I shall rename this variable to `relayShardsToSubscribeTo`
These are the shards we want relay to subscribe to.

Do note that this is a pure `wakunode2` service node matter. In an application, you would want the application to run `subscribe` at the start, this option is purely for CLI and I will remove it from `WakuConf`.

waku/factory/conf_builder/waku_conf_builder.nim
fryorcraken: More concise because it does not convey any meaning. Please start to see nwaku as an application library. what `shards` should a developer select? The answer is `none` when using autosharding. And for static sharding, this can be done with `subscribe` API, meaning they are the shards to subscribe to.

waku/factory/conf_builder/waku_conf_builder.nim
fryorcraken: I am sorry but an `Option` means that it can be set, or it can be unset.

a `default` means that if the developer does not set it, then we apply a `default value`.

Meaning there is default value, does not mean it's the only possible value.

waku/factory/conf_builder/waku_conf_builder.nim
fryorcraken: What do you mean it should not be needed? The number of shards in the network impact the autosharding protocol. If one node sets `8` and another sets `1024` then a given content topic `foo` will end up on different shards, meaning you won't be able to transmit messages.

waku/factory/external_config.nim
fryorcraken: Please do not make 3 comments about the same thing, it's a waste of time for everyone. You open the debate about removing this parameter. We can discuss it in one comment thread.

waku/factory/node_factory.nim
fryorcraken: A given network should either have static sharding, or auto sharding, not both. Do we agree with that?

I am a developer that is building my new `foobar` app on TWN, hence auto sharding is enabled.

Please walk me through how I need to select the "right shard" to subscribe to.

waku/factory/waku_conf.nim
fryorcraken: > When a node is configured and the operator selects the shard they want to operate in. Why call them `activeRelayShards` instead of `shards` if it's the config desire of the user/operator and not a run time issue?

Because the action is unclear, why does one need to specify `shards` ? Well, because that the shards we want relay to subscribe to. But that only applies in one specific case: when running a service node with relay.

When running an application, the users or developer does **not** subscribe to shards, it subscribe to specific content topics they want to receive messages on.

And when not using relay, they don't need shards either.

And even in this case, there is still the problem of:

- I am using TWN
- I want to run a service node to support my node
- What "shard" do I select?

Well, you shouldn't need to "select" a shard, but a "content topic" instead (or a partial one), because autosharding abstract the concept of shards.

waku/waku_lightpush/protocol.nim
fryorcraken: https://github.com/waku-org/nwaku/pull/3486

waku/factory/waku_conf.nim
fryorcraken: > not a run time issue?

So we are clear, it is a run time issue. Getting relay to subscribe to shards is not an initialization matter, it is a run time matter.

waku/factory/conf_builder/waku_conf_builder.nim
fryorcraken: Also well spotted that there is a gap in the spec IMO. It says that auto sharding is the default choice, but does not recommend autosharding parameters such as the number of shards.

Cc @jm-clius: what should be the default number of shards, in the same vein as

> Autosharding selects shards automatically and is the default behavior for shard choice

waku/factory/conf_builder/waku_conf_builder.nim
Ivansete-status: Yes indeed, you are absolutely right. When I wrote my comment I was thinking about having 1024 by default to all nodes but, subscribing to 1024 shards doesn't seem like a great idea because I suspect it won't perform very well.
The static and auto shards generates a lot of confusion and is a source of possible discrepancies among nodes. For example, what happens if someone configures the following?

- shards [1, 2, 3]
  plus
- numShardsInCluster=4

In that case, the `foo` content-topic can potentially fall in wrong shard (considering all nodes are in 1, 2, 3.)

Therefore, I think we need to infer the value of `numShardsInCluster`, or have a good default.

Another possible option would be to implement a hybrid solution between static and auto sharding: always have defined the `shards` list and, in case of using auto-sharding, infer the available shards from that available `shards` list. wdyt?

fryorcraken: Let's take it from a user PoV. I am only covering when `auto-sharding` is enabled.

1. Library user/application developer: in this case, I don't think `--shards` should be used at all. They should use `subscribe (contentTopic)` to subscribe to a topic.
When sending a message,  then we will have two choices when they are not subscribed to the right shard:
  1. non-blocking logic: fallback to light push
  2. blocking logic: error and require `subscribe` to be done first
2. Node operator for a given application: In this case, the operator want to support their specific application and run relay/store/etc for their shard. which means they would use `--content-topics` to subscribe to the right shards
3. Node operators - for Waku/incentivization/etc: In this instance, a node operator may want to support all shards, or specific shards where there is most traffic. Meaning `--shards` is most appropriate.

## In conclusion

To setup relay shard subscription:

A. On the `wakunode2` CLI, it is fine for a user to either both `--shards` or `--content-topic` to specify what shards to subscribe to, but it should not be both *at the same time*. Note that a _partial content topic_ should be accepted (only first fields are needed for autosharding).
B. On the `Waku API`, then I would suggest that neither `--shards` nor `--content-topic` are available at node initialization, but instead, `subscribe(contentTopic)` is used.

When sending a message, in relay mode, on a shard not subscribed to. I suggest to `error out` for now. We can decide at a later stage to fallback to light push.

fryorcraken: > Therefore, I think we need to infer the value of numShardsInCluster, or have a good default.

The default nework conf value should be "TWN", so that this number is 8.
If someone want to roll out their own network, and own cluster id, then I agree we can propose a default value. Probably 1 or 4. 
@jm-clius WDYT?

waku/factory/conf_builder/waku_conf_builder.nim
fryorcraken: `ShardingConf` MUST remain an enum. Because `numShardsInCluster` does not make sense for a static sharding conf.

waku/waku_lightpush/protocol.nim
fryorcraken: rebased

waku/factory/conf_builder/waku_conf_builder.nim
fryorcraken: @Ivansete-status just so we are on the same page, you seem to say that `numShardsInCluster` is not needed, as well as wanting a good default, in lieu of the value.

Ultimately, it **is** needed because different network may have different shards. For example, we may grown the number of shards on Status network to 50 with the chat sdk. But someone bootstrapping a brand new network may prefer to stick to `1` until scaling is needed. So this parameter **is** needed.
Then the question of default value remain up to discussion, right?

Ivansete-status: A. and B. makes sense to me in this comment: https://github.com/waku-org/nwaku/pull/3468#discussion_r2181398151
The following looks great to me as well:
> When sending a message, in relay mode, on a shard not subscribed to. I suggest to error out for now.

------

> If someone want to roll out their own network, and own cluster id, then I agree we can propose a default value. Probably 1 or 4.

Why not 8 too?

I think we need a good first default, 8, and then, define an scaling algorithm that makes the network breath, i.e., subscribe to more or less shards depending on bandwidth needed. That will need a consensus in the network.

For that, I believe we need to also establish a fixed _consensus_ shard on every network, on every cluster. For example, we can enforce that every cluster will have the shard `0` to be reserved for only containing Waku consensus and coordination messages. (Another option is having `waku/2/rs/consensus` pubsubtopic and leave shard `0` for data.)

That consensus topic will help nodes to reach an agreement in:
- Define the total number of shards in the network. Each node will subscribe/unsubscribe accordingly.
- Determine which nodes will act as bridge or aggregator between shards.
- When to start scaling up/down, i.e., when to start changing the current shard set.
- Define the shards a certain node will subscribe to (we can do math considering the peerId.)

------

> Ultimately, it is needed because different network may have different shards.

I think you are absolutely right with that but, on the other hand, I believe we need to implement a network algorithm that makes the network scale up/down accordingly. I strongly suggest following the `consensus` topic idea that I commented above.

> Then the question of default value remain up to discussion, right?

Yes :) I'd suggest having a good default plus an algorithm that makes it adaptive as per network conditions.

waku/factory/conf_builder/waku_conf_builder.nim
jm-clius: This is because Autosharding is only defined as a concept in this spec and presented as the preferred default option when considering sharding for Waku networks, followed by static (and even named sharding, which we don't even implement anymore). This spec does not define any specific application of autosharding for specific Waku Network(s), as this more properly belongs to the "consensus" specification for a specific network (e.g. the Status Network(s), TWN, etc.). The _only_ network for which we have defined autosharding is TWN, as it's the only specified network with a consensus on cluster ID and number of shards: https://rfc.vac.dev/waku/standards/core/64/network#autosharding.
This out-of-band consensus on a network definition is a **must** for autosharding to work as expected. The spec should indeed be clarified in this regard.

My initial preference was that autosharding, as an app-level API simplification, should only be available when TWN is configured. However, it became clear for testing purposes and more modular use of Waku, that people preferred having the option of ad-hoc creating their own "network definitions" with autosharding by configuring a `numShardsInCluster` option. Note that this is a higher-level **network definition** and has nothing to do at all with which relay shards are actually being subscribed to. You could perfectly well set this parameter and not subscribe to any traffic at all. Also, it implies that these experimenters "know" that others participating in the experiment have configured their nodes in the same way (or are using the default config - see below). The way to formalise this is in the form of published network specifications for each new network, where the number of shards, services, etc. are agreed upon.

Autosharding also only makes sense in the API when applications build on top of a Waku node. In other words, it should only be used when the node is run within specific content topic (i.e. "App") context - it has no use for general service nodes that do not care about specific content topics.

My view of defaults:
Default configuration should always be for TWN.
If someone sets a different cluster ID, the default should be that that cluster is assumed to have only `1` shard. There's no need to pre-emptively "scale" these clusters - we had a specific reason for making TWN experimentally work with 8 out of the box, but that reasoning doesn't apply to the defaults.

waku/factory/conf_builder/waku_conf_builder.nim
jm-clius: I don't think at this stage we need a consensus algorithm to set a network definition. Consensus for the networks we define is reached by publishing an RFC. We have only one defined network, which is TWN for which number of shards is 8. If we need to scale, we'll publish a new version of the spec - TWN Gen 1 if you will - and autosharding is already defined in such a way that it will work with backwards compatibility.

waku/factory/conf_builder/waku_conf_builder.nim
fryorcraken: Thanks @jm-clius. I actually came to the same conclusion overnight. Which hopefully is in line with @Ivansete-status 's original comment.

With the introduction of the `preset` concept, we are moving towards TWN being the default. It is not straightforward and as RLN plays a part in it. 

I suggest to have nwaku default to autosharding with `1` shard when TWN is not set. For a few reason:

1. Autosharding means abstracting shards concept, which is better for dev ex
2. Autosharding is yet to be properly dogfood on large applications, TWN with `8` is a starter but no serious app is running on it. Hence, a value of `1` is low risk.

waku/factory/conf_builder/waku_conf_builder.nim
fryorcraken: As per https://github.com/waku-org/nwaku/pull/3468#discussion_r2182728357 I will proceed with the following changes:

1. autosharding by default
2. num of shards in cluster: `1`

waku/factory/conf_builder/waku_conf_builder.nim
Ivansete-status: Thanks for the feedback! I with the suggested approach :ok_hand:

PR #3469, opened 2025-06-24
chore: rename modules

Ivansete-status:
## Description
Rename some libwaku modules


library/libwaku.nim
```diff
@@ -16,17 +16,17 @@ import
...
-  ./waku_thread/inter_thread_communication/requests/protocols/relay_request,
-  ./waku_thread/inter_thread_communication/requests/protocols/store_request,
-  ./waku_thread/inter_thread_communication/requests/protocols/lightpush_request,
-  ./waku_thread/inter_thread_communication/requests/protocols/filter_request,
-  ./waku_thread/inter_thread_communication/requests/debug_node_request,
-  ./waku_thread/inter_thread_communication/requests/discovery_request,
-  ./waku_thread/inter_thread_communication/requests/ping_request,
-  ./waku_thread/inter_thread_communication/waku_thread_request,
+  ./waku_context/waku_context,
+  ./waku_context/waku_thread_requests/requests/node_lifecycle_request,
```
fryorcraken: repetition of the `waku` word seems unnecessary.

library/libwaku.nim
```diff
@@ -54,7 +54,7 @@ proc handleRequest(
     callback: WakuCallBack,
     userData: pointer,
 ): cint =
-  waku_thread.sendRequestToWakuThread(ctx, requestType, content, callback, userData).isOkOr:
+  waku_context.sendRequestToWakuThread(ctx, requestType, content, callback, userData).isOkOr:
```
fryorcraken, line 57: can't it just be `context`? we are in the waku lib, a `waku context` seems redundant.

library/libwaku.nim
Ivansete-status: Yes I completely agree that sounds too repetitive.
Addressed in https://github.com/waku-org/nwaku/pull/3469/commits/608dd012cbce2b3d12382d7beeeb08d52bdc0b80

library/libwaku.nim
Ivansete-status: The `waku_context` in that point refers to the `waku_context.nim` module, which in turn, defines the following type:

```nim
type WakuContext* = object
```
And this type should be named that way, IMO.
We need to keep aligned the module's name to the underlying defined type.

Ivansete-status:
Merging as Franck mentioned is fine from his PoV

fryorcraken:
Yes LGTM

PR #3470, opened 2025-06-25
wip: remove more pubsubtopic usage

fryorcraken:
# Description

Remove more pubsub topic usage mid code.

I actually realize I should start with relay instead.

The main reason is that pubsub topic are not strongly typed, so there is always doubt on whether it can match a shard/cluster.

The intent would be to only have pubsub topic near the wire, and use shards every where.

Ultimately, we could even remove pubsub topic from the wire.

Let me know how you feel on the concept before I spend more time on it.

# Changes

- [ ] ...
- [ ] ...

### DIFF
--- a/waku/node/waku_node.nim
+++ b/waku/node/waku_node.nim
@@ -521,7 +521,7 @@ proc mountFilterClient*(node: WakuNode) {.async: (raises: []).} =
 
 proc filterSubscribe*(
     node: WakuNode,
-    pubsubTopic: Option[PubsubTopic],
+    shard: Option[RelayShard],
     contentTopics: ContentTopic | seq[ContentTopic],
     peer: RemotePeerInfo | string,
 ): Future[FilterSubscribeResult] {.async: (raises: []).} =
@@ -603,6 +603,15 @@ proc filterSubscribe*(
 
     # return the last error or ok
     return subRes
+
+# proc filterSubscribe* {.deprecated: "Pass a shard instead of a pubsubtopic".} (
+#     node: WakuNode,
+#     pubsubTopic: Option[PubsubTopic],
+#     contentTopics: ContentTopic | seq[ContentTopic],
+#     peer: RemotePeerInfo | string,
+# ): Future[FilterSubscribeResult] {.async: (raises: []).} =
+
+#   echo "TODO"
 
 proc filterUnsubscribe*(
     node: WakuNode,
--- a/waku/waku_core/message/digest.nim
+++ b/waku/waku_core/message/digest.nim
@@ -59,6 +59,9 @@ proc computeMessageHash*(pubsubTopic: PubsubTopic, msg: WakuMessage): WakuMessag
 
   return ctx.finish() # Computes the hash
 
+proc computeMessageHash*(shard: RelayShard, msg: WakuMessage): WakuMessageHash =
+  return computeMessageHash(shard.toPubsubTopic(), msg)
+
 proc cmp*(x, y: WakuMessageHash): int =
   if x < y:
     return -1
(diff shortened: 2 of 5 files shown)


jm-clius (approved):
Adding my comment here with an approval, to show that I'm willing to go with the majority opinion here. 😄 The main issue here is inconsistency in the data model and this fixes things.

At the risk of (re)opening a can of worms here, I've always leaned towards cluster/shard (and related things like autosharding) _only_ being a function of configuration and consequently only existing at the configuration layer. This is because Waku core strongly derives its properties from its pubsub basis where you have (only) two simple concepts:  `un/subscribe` and `topic`. Cluster, shard, autosharding provide user-friendly, defensive programming means to construct and assign the "topic" here, but keeping the abstraction limited to config and the pubsub topic a simple string have some benefits IMO. For example, string topics in the wire protobufs communicate the "pubsubness" much more simply than split cluster-shard fields. It also prevents devs from divorcing shard and cluster, neither of which has a fully qualified meaning on its own from the core protocols' perspective. Using simple string pubsub topics internally allows for simple indexing against strings (or hashes of strings) and doesn't introduce an unclear border when moving from Waku to the encapsulated libp2p domain. I'd also argue that `PubsubTopic` does have strongly typed properties - within the pubsub domain two topics are equal if the string representations are exactly equal (not because some substrings match). Using the cluster/shard format simply allows constructing these topic strings to be less prone to bugs or typing mistakes. It prevents questions arising in the protocol core around things like "what if cluster matches but not the shard?" or vice versa, which doesn't make sense when understood as a single concept that defines the pubsub topic.

With my thinking explained, I'm not attached to this model and can see several counterarguments (e.g. how we encode cluster/shard and not string topic in discovery methods). Happy to help us move to something consistent here and review the Wire Protocols if we decide on a direction.

fryorcraken:
thanks for expressing your opinion here @jm-clius.

I was thrown off by another piece of code which made me go down this path.

I am going to leave this for now. I think the Waku API will start fixing this problem because we will remove any mention of pubsub at the higher layer.

We can then review case by case whether pubsub or cluster/id is better suited.

In terms of your comparison part - I think this mainly about ensure we have a good data structure representation that allows easy comparison.

The problem with a string is that you can only be sure it is well formed if you "checked" in the past, or check often, which lead to pitfalls and potential optimization issues (string check are cheap.. until you do it 50 times per message).

It is an interesting point re-pubsubness, and I would actually go in the other direction: we are leaking implementations details here from gossipsub into Waku. For example, from a (store/filter/etc) sync pov, woudn't it be better to just deal with content topics and shards numbers?

jm-clius:
> I would actually go in the other direction: we are leaking implementations details here from gossipsub into Waku. For example, from a (store/filter/etc) sync pov, woudn't it be better to just deal with content topics and shards numbers?

A valid point. Again, I'll be OK with any model that is just consistent. The reason I lean towards keeping the pubsubness is because the heart of our Waku Relay is simply a pubsub protocol and all other protocols interact with this core. For example a good external API for publishing using Waku Relay would be `publish(cluster, shard, contentTopic, content...)` with the API translating the `cluster` and `shard` to a `pubsubTopic` within the API before forwarding to the internal Relay interface that deals with `pubsubTopic`. Lightpush is a protocol for interacting with Relay. It could expose the same `publish(cluster, shard, contentTopic, content...)` external API. It then serialises this publication event to trigger the same internal Relay interface as before, this time at a distant Relay service node. Conceptually though it's the same publication event as when we directly published using the Relay API, even if the external API and the responding internal Relay events are in physically separate nodes. To me it feels strange therefore that we wouldn't expect this Lighpush version of the publish API to perform the same translation to `pubsubTopic` as we do for the equivalent Relay method.

PR #3472, opened 2025-06-25
feat: remove broken dos mechanism from transfer protocol

SionoiS:
# Description

The Dos mechanism designed to protect transfer protocol is flawed.

I decided to disable it until we can design something better.

### DIFF
--- a/waku/waku_store_sync/transfer.nim
+++ b/waku/waku_store_sync/transfer.nim
@@ -143,9 +143,10 @@ proc needsReceiverLoop(self: SyncTransfer) {.async.} =
 proc initProtocolHandler(self: SyncTransfer) =
   proc handler(conn: Connection, proto: string) {.async: (raises: [CancelledError]).} =
     while true:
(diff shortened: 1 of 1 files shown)


NagyZoltanPeter (approved):
How exactly is it flowed?
For filter wee use `PerPeerRateLimiter` for ex. Is such or similar can be better for sync use case?

SionoiS:
> How exactly is it flowed? For filter wee use `PerPeerRateLimiter` for ex. Is such or similar can be better for sync use case?

Concurrent logic in reconciliation vs transfer makes the mechanism brittle. Once we have RLN proof verification we won't need it anyway.

jm-clius (approved):
Oh, I just realised that this is the same as this PR that I never merged: https://github.com/waku-org/nwaku/pull/3412

Should have done so. Apologies.

Repeating point made in that PR: I think there should be explicit coordination between transfer and sync protocols in a single data structure to avoid implicit coordination/race conditions, but we can probably follow this up in future.

PR #3474, opened 2025-06-26
fix: libwaku.so compilation

gabrielmer:
# Description
Fixing library name for non-windows builds. Without this fix, we get `liblibwaku.so` as output

## Issue

#3236


darshankabariya (approved):
Cool, LGTM !!

Ivansete-status (approved):
LGTM! Thanks for it! :100: 
We might need to do something similar for `static` library, and name `.Lib` on Windows and `.so` on Unix-based

PR #3480, opened 2025-06-30
fix: make test configuration

darshankabariya:
Added support to run tests using the format: make test [<FILE_NAME>] [<TEST_NAME>]


fryorcraken:
Don't forget to fix the CI

Makefile
```diff
@@ -53,7 +53,16 @@ endif
...
+TEST_FILE := $(word 2,$(MAKECMDGOALS))
+TEST_NAME := $(wordlist 3,$(words $(MAKECMDGOALS)),$(MAKECMDGOALS))
+
+test:
+ifeq ($(strip $(TEST_FILE)),)
+	$(MAKE) testcommon
+	$(MAKE) testwaku
+else
+	$(MAKE) compile-test $(TEST_FILE) $(TEST_NAME)
+endif
```
Ivansete-status, line 65: The following helps to avoid the annoying error when passing a test name:

```suggestion
endif

# this prevents make from erroring on unknown targets like "Index"
%:
	@true
```

---

And this is an example of that error:

```nim
make[1]: *** No rule to make target 'Index'.  Stop.
make[1]: Leaving directory '/home/ivansete/workspace/status/nwaku'
make: *** [Makefile:65: test] Error 2
```

... which happened after running this:

```nim
make test tests/waku_archive/test_driver_queue_index.nim "Index comparison"
```

darshankabariya: Thanks for the suggestion, @Ivansete-status, Updated accordingly.

fryorcraken:
Checking now, but you did not update teh README. replace all instances of `make test/` with `make test `

PR #3485, opened 2025-07-02
fix: deprecate --dns-discovery

fryorcraken:
Properly deprecates `--dns-discovery` CLI arg.
DNS Discovery is enabled if a non-empty DNS Discovery URL is passed.


waku/factory/conf_builder/dns_discovery_conf_builder.nim
```diff
@@ -8,30 +8,26 @@ logScope:
...
-
 proc withEnrTreeUrl*(b: var DnsDiscoveryConfBuilder, enrTreeUrl: string) =
   b.enrTreeUrl = some(enrTreeUrl)
 
 proc withNameServers*(b: var DnsDiscoveryConfBuilder, nameServers: seq[IpAddress]) =
   b.nameServers = nameServers
 
 proc build*(b: DnsDiscoveryConfBuilder): Result[Option[DnsDiscoveryConf], string] =
-  if not b.enabled.get(false):
+  if b.enrTreeUrl.isNone():
```
chaitanyaprem, line 24: Does deprecated mean we want to support for backward compatibility? If so shouldn't we check for either of them?

If not, then we should remove the dnsDiscovery flag altogether.

fryorcraken: Yes, it is still here for backward compatibility so that we can deploy a new version of the node, and then remove it from the infra config after.

Otherwise, we have to deploy a new version AND change infra config at the same time, which is not ideal.

chaitanyaprem (approved):
LGTM

tests/all_tests_waku.nim
```diff
@@ -92,20 +92,9 @@ import
...
-  ./wakunode_rest/test_rest_debug_serdes,
-  ./wakunode_rest/test_rest_relay,
-  ./wakunode_rest/test_rest_relay_serdes,
-  ./wakunode_rest/test_rest_serdes,
-  ./wakunode_rest/test_rest_filter,
-  ./wakunode_rest/test_rest_lightpush,
-  ./wakunode_rest/test_rest_lightpush_legacy,
-  ./wakunode_rest/test_rest_admin,
-  ./wakunode_rest/test_rest_cors,
-  ./wakunode_rest/test_rest_health
```
gabrielmer, line 106: don't we need these tests?

gabrielmer (approved):
Thanks so much!

tests/all_tests_waku.nim
Ivansete-status: I've double-checked and just need to add the `./wakunode_rest/test_rest_lightpush,` into the `tests/wakunode_rest/test_all.nim` module. The others are there in different order
( cc @fryorcraken )

tests/factory/test_all.nim
```diff
@@ -0,0 +1 @@
+import ./test_external_config.nim, ./test_node_factory.nim, ./test_waku_conf.nim
```
Ivansete-status:
```suggestion
import ./test_external_config, ./test_node_factory, ./test_waku_conf
```

waku/factory/external_config.nim
```diff
@@ -996,9 +997,9 @@ proc toWakuConf*(n: WakuNodeConf): ConfResult[WakuConf] =
   b.metricsServerConf.withHttpPort(n.metricsServerPort)
   b.metricsServerConf.withLogging(n.metricsLogging)
 
-  b.dnsDiscoveryConf.withEnabled(n.dnsDiscovery)
-  b.dnsDiscoveryConf.withEnrTreeUrl(n.dnsDiscoveryUrl)
-  b.dnsDiscoveryConf.withNameServers(n.dnsAddrsNameServers)
+  if n.dnsDiscoveryUrl != "":
+    b.dnsDiscoveryConf.withEnrTreeUrl(n.dnsDiscoveryUrl)
+    b.dnsDiscoveryConf.withNameServers(n.dnsAddrsNameServers)
```
Ivansete-status: Maybe?
```suggestion
  if n.dnsDiscoveryUrl != "":
    b.dnsDiscoveryConf.withEnrTreeUrl(n.dnsDiscoveryUrl)

  b.dnsDiscoveryConf.withNameServers(n.dnsAddrsNameServers)
```

tests/all_tests_waku.nim
fryorcraken: > don't we need these tests?

We do, and we run them via `./wakunode_rest/test_all`

fryorcraken: oh, the confusion is that `test_rest_store` was not being tested before! well spotted @Ivansete-status

PR #3487, opened 2025-07-02
feat: use json serde library for config over ffi

fryorcraken:
# Description

Use nim-json-serialization instead of going through a default config to know what keys to look for.
# Changes

- [ ] ...
- [ ] ...


gabrielmer:
The point of using the default config is so we keep the default values for the fields not specified in the incoming json-string, I think with this solution we would have this gap. All the fields not specified won't have the default values set in `external_config.nim`

fryorcraken:
> The point of using the default config is so we keep the default values for the fields not specified in the incoming json-string, I think with this solution we would have this gap. All the fields not specified won't have the default values set in `external_config.nim`

Thanks for clarifying.

I will close this for now, and will use this strategy when rolling out the Waku API (PR incoming).

PR #3488, opened 2025-07-03
chore: use distinct type for Light push status codes

fryorcraken:
# Description

Use a distinct type for light push status codes, so that we don't need to cast to uint32 in several places.

# Changes

- [x] Create a `distinct` type for Light Push status codes

<!--

### DIFF
--- a/waku/node/waku_node.nim
+++ b/waku/node/waku_node.nim
@@ -82,6 +82,8 @@ const clientId* = "Nimbus Waku v2 node"
 
 const WakuNodeVersionString* = "version / git commit hash: " & git_version
 
+const LightPushError = lightpush_protocol.ErrorCode
+
 # key and crypto modules different
 type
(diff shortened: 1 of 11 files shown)


NagyZoltanPeter (approved):
Excelent one! Thanks for it!

waku/node/waku_node.nim
```diff
@@ -82,6 +82,8 @@ const clientId* = "Nimbus Waku v2 node"
 
 const WakuNodeVersionString* = "version / git commit hash: " & git_version
 
+const LightPushError = lightpush_protocol.ErrorCode
```
s-tikhomirov: Should this be `const LightPushErrorCode` instead of `LightPushError`? This would avoid confusion as to whether an "error" and an "error code" are the same thing or not.

s-tikhomirov (approved):
LGTM, added a minor comment.

PR #3489, opened 2025-07-03
chore: use type for rate limit config

fryorcraken:
# Description

Use type instead of `seq[string]` for rate limit config in waku config. Enables to fail faster (at config time) if the string is malformed.

Also enables using object in some scenarios (libwaku).

# Changes

- [x] Use `ProtocolRateLimitSettings` type in `WakuConf`
- [x] Move parse failure as part of `WakuConf`, instead of in `WakuNodeBuilder`.

## Issue

- https://github.com/waku-org/nwaku/issues/3493

### DIFF
--- a/waku/factory/conf_builder/rate_limit_conf_builder.nim
+++ b/waku/factory/conf_builder/rate_limit_conf_builder.nim
@@ -11,11 +11,8 @@ type RateLimitConfBuilder* = object
 proc init*(T: type RateLimitConfBuilder): RateLimitConfBuilder =
   RateLimitConfBuilder()
 
-proc with*(b: var RateLimitConfBuilder, rateLimits: seq[string]) =
+proc withRateLimits*(b: var RateLimitConfBuilder, rateLimits: seq[string]) =
   b.strValue = some(rateLimits)
-
-proc with*(b: var RateLimitConfBuilder, rateLimits: ProtocolRateLimitSettings) =
-  b.objValue = some(rateLimits)
 
 proc build*(b: RateLimitConfBuilder): Result[ProtocolRateLimitSettings, string] =
   if b.strValue.isSome() and b.objValue.isSome():
--- a/waku/factory/conf_builder/rate_limit_conf_builder.nim
+++ b/waku/factory/conf_builder/rate_limit_conf_builder.nim
@@ -1,0 +1,32 @@
+import chronicles, std/[net, options], results
+import waku/common/rate_limit/setting
+
+logScope:
+  topics = "waku conf builder rate limit"
+
+type RateLimitConfBuilder* = object
+  strValue: Option[seq[string]]
+  objValue: Option[ProtocolRateLimitSettings]
+
+proc init*(T: type RateLimitConfBuilder): RateLimitConfBuilder =
+  RateLimitConfBuilder()
+
+proc with*(b: var RateLimitConfBuilder, rateLimits: seq[string]) =
+  b.strValue = some(rateLimits)
+
+proc with*(b: var RateLimitConfBuilder, rateLimits: ProtocolRateLimitSettings) =
+  b.objValue = some(rateLimits)
+
+proc build*(b: RateLimitConfBuilder): Result[ProtocolRateLimitSettings, string] =
+  if b.strValue.isSome() and b.objValue.isSome():
+    return err("Rate limits conf must only be set once on the builder")
+
+  if b.objValue.isSome():
+    return ok(b.objValue.get())
+
+  if b.strValue.isSome():
+    let rateLimits = ProtocolRateLimitSettings.parse(b.strValue.get()).valueOr:
+      return err("Invalid rate limits settings:" & $error)
+    return ok(rateLimits)
+
+  return ok(DefaultProtocolRateLimit)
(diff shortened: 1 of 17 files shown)


waku/factory/conf_builder/rate_limit_conf_builder.nim
```diff
@@ -0,0 +1,32 @@
...
+  objValue: Option[ProtocolRateLimitSettings]
+
+proc init*(T: type RateLimitConfBuilder): RateLimitConfBuilder =
+  RateLimitConfBuilder()
+
+proc with*(b: var RateLimitConfBuilder, rateLimits: seq[string]) =
+  b.strValue = some(rateLimits)
+
+proc with*(b: var RateLimitConfBuilder, rateLimits: ProtocolRateLimitSettings) =
+  b.objValue = some(rateLimits)
```
NagyZoltanPeter: Is this used anywhere? I may not find it.

waku/factory/conf_builder/rate_limit_conf_builder.nim
```diff
@@ -0,0 +1,32 @@
...
+
+  if b.objValue.isSome():
+    return ok(b.objValue.get())
+
+  if b.strValue.isSome():
+    let rateLimits = ProtocolRateLimitSettings.parse(b.strValue.get()).valueOr:
+      return err("Invalid rate limits settings:" & $error)
+    return ok(rateLimits)
+
+  return ok(DefaultProtocolRateLimit)
```
NagyZoltanPeter, line 29: imo, this is a one way process, so getting string representation from conf and translate it to typed struct.
If we need objValue representation for the builder, I would rather parse the strValue into it rather having a kind of objValue precedence over the strValue.
I may overlook sthg but I think its enough to build from string representation.

NagyZoltanPeter (approved):
All ok, only to consider if we really need builder support for already structured rate_limit_setting

gabrielmer (approved):
LGTM, thank you!

waku/factory/conf_builder/rate_limit_conf_builder.nim
```diff
@@ -0,0 +1,32 @@
...
+  topics = "waku conf builder rate limit"
+
+type RateLimitConfBuilder* = object
+  strValue: Option[seq[string]]
+  objValue: Option[ProtocolRateLimitSettings]
+
+proc init*(T: type RateLimitConfBuilder): RateLimitConfBuilder =
+  RateLimitConfBuilder()
+
+proc with*(b: var RateLimitConfBuilder, rateLimits: seq[string]) =
```
Ivansete-status: I'd suggest using a more explicit proc name to make future lookups easier :)
```suggestion
proc withRateLimits*(b: var RateLimitConfBuilder, rateLimits: seq[string]) =
```

Ivansete-status (approved):
LGTM! Thanks for it! :100:

waku/factory/conf_builder/rate_limit_conf_builder.nim
fryorcraken: Excellent point. I intend to initialize rate limit appropriate to Waku as an app lib here: https://github.com/waku-org/nwaku/pull/3490/files (copying what I saw in status-go, may need adjustment now that @gabrielmer pointed to me where we actually config nwaku for status).

For this initialization, I prefer to use object initialization form over string to parse.

Note that would be a phase 1, a phase 2 would be to make the rate limit parameters appropriate for an application the default ones for `WakuConf`, and have a different default for `WakuNodeConf` when its run as a service node.

Nonetheless, I think it is best that I add some tests to demonstrate that indeed the parameters can be instantiated via object.

> If we need objValue representation for the builder, I would rather parse the strValue into it rather having a kind of objValue precedence over the strValue.

There is no precedence, you can see line 22 that it fails if both are set.

> I would rather parse the strValue 

Yes indeed, I hesitated on whether to parse in `with*` or when doing `build`.

I opted to do it in `build` because across all conf builders, the `with*` functions are dumb. I did not wanted to started making some of them more clever, and opted to keep a pattern here. I am happy to be convinced otherwise.

waku/factory/conf_builder/rate_limit_conf_builder.nim
fryorcraken: I hesitated, fair point, will correct.

waku/factory/conf_builder/rate_limit_conf_builder.nim
NagyZoltanPeter: I think it's ok to do the parsing in the build phase. Maybe to check if both str and obj is set to log a warning or even fail early rather than confuse the operator might not understand why the config is overriden. I accept this is a rare case might not happen ever.

fryorcraken: > Maybe to check if both str and obj is set to log a warning or even fail early rather than confuse the operator might not understand why the config is overriden. I accept this is a rare case might not happen ever.

This is already the case in the PR. Or Am I missing something?

```
  if b.strValue.isSome() and b.objValue.isSome():
    return err("Rate limits conf must only be set once on the builder")
```

fryorcraken: @NagyZoltanPeter Looking in more details, you are right and best to just use the string form

waku/factory/conf_builder/rate_limit_conf_builder.nim
fryorcraken: @NagyZoltanPeter Looking in more details, you are right and best to just use the string form

PR #3490, opened 2025-07-03
feat!: Use Waku API for libwaku

fryorcraken:
# Description

Demonstrate the Waku API defined in https://github.com/waku-org/specs/pull/65 in libwaku.

# Dependencies

Here are a list of changes to make this PR happen:

- https://github.com/waku-org/nwaku/pull/3489: enable instantiating the rate limit directly
- #3468 

# Changes

- [ ] ...
- [ ] ...

### DIFF
--- a/library/libwaku_api.nim
+++ b/library/libwaku_api.nim
@@ -4,7 +4,7 @@ import waku/factory/waku
 
 import ./libwaku_conf
 
-proc createNode*(config: LibWakuConf): Future[Result[Waku, string]] {.async.} =
+proc createNode*(config: LibWakuConfig): Future[Result[Waku, string]] {.async.} =
   let wakuConf = toWakuConf(config).valueOr:
     return err("Failed to handle the configuration: " & error)
 
--- a/library/libwaku_conf.nim
+++ b/library/libwaku_conf.nim
@@ -9,7 +9,7 @@ type ShardingMode* = enum
   AutoSharding = "auto"
   StaticSharding = "static"
 
-type AutoShardingConf* = object
+type AutoShardingConfig* = object
   numShardsInCluster*: uint16
 
 type RlnConfig* = object
@@ -21,32 +21,32 @@ type MessageValidation* = object
   maxMessageSizeBytes*: uint64
   rlnConfig*: Option[RlnConfig]
 
-type NetworkConf* = object
+type NetworkConfig* = object
   bootstrapNodes*: seq[string]
   staticStoreNodes*: seq[string]
   clusterId*: uint16
   shardingMode*: Option[ShardingMode]
-  autoShardingConf*: Option[AutoShardingConf]
+  autoShardingConfig*: Option[AutoShardingConfig]
   messageValidation*: Option[MessageValidation]
 
 type WakuMode* = enum
   Edge = "edge"
   Relay = "relay"
 
-type LibWakuConf* = object
+type LibWakuConfig* = object
   mode*: WakuMode
-  networkConf*: Option[NetworkConf]
+  networkConfig*: Option[NetworkConfig]
   storeConfirmation*: bool
   ethRpcEndpoints*: seq[string]
 
 proc DefaultShardingMode(): ShardingMode =
   return ShardingMode.Autosharding
 
-proc DefaultAutoShardingConf(): AutoShardingConf =
-  return AutoShardingConf(numShardsInCluster: 1)
+proc DefaultAutoShardingConfig(): AutoShardingConfig =
+  return AutoShardingConfig(numShardsInCluster: 1)
 
-proc DefaultNetworkConf(): NetworkConf =
-  return NetworkConf(
+proc DefaultNetworkConfig(): NetworkConfig =
+  return NetworkConfig(
     bootstrapNodes:
       @[
         "enrtree://AIRVQ5DDA4FFWLRBCHJWUWOO6X6S4ZTZ5B667LQ6AJU6PEYDLRD5O@sandbox.waku.nodes.status.im"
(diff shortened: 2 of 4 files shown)


library/libwaku_conf.nim
```diff
@@ -0,0 +1,66 @@
...
+    b.withPeerExchange(true)
+
+    # Values applied when used as a library - should probably become default values
+    b.filterServiceConf.withMaxPeersToServe(20)
+    b.withRateLimits(
+      @[
+        
+      ]
+    )
+nwakuCfg.RateLimits.Filter = &bindingscommon.RateLimit{Volume: 100, Period: 1, TimeUnit: bindingscommon.Second}
```
fryorcraken: doing #3489 so that I can just instantiate the rate limits here.

library/libwaku_conf.nim
```diff
@@ -0,0 +1,66 @@
+import std/options, waku/factory/waku_conf, waku/factory/conf_builder
+
+type ShardingMode* = enum
+  AutoSharding
+  StaticSharding
+
+type AutoShadingConf* = object
```
fryorcraken:
```suggestion
type AutoShardingConf* = object
```

library/libwaku_conf.nim
```diff
@@ -0,0 +1,66 @@
+import std/options, waku/factory/waku_conf, waku/factory/conf_builder
```
Ivansete-status: I think we need to rename this module to `libwaku_sdk_conf.nim` instead, to make it clearer it is aimed for the opinionated version of the library and not for the core libwaku

fryorcraken: I would say this would only fix the tip of the iceberg.

I think `waku` folder should be named `lib` and `library` can be renamed `sdk`

One could argue that my new config should be in `api` folder.

Is this addressed by the "architecture and api" virtual offsite recommendation? I havent see a forum post.

Ivansete-status: Thanks for the suggestion! Let me share my 2cs in case of willing to rename folders.

> I think waku folder should be named lib and library can be renamed sdk

- `waku` folder is properly named IMHO. We cannot name it `lib` because it does not expose a library outside. Also, the current `waku` approach seems to be the idiomatic way of naming it. See for example: [chronos](https://github.com/status-im/nim-chronos/tree/master/chronos) [chronicles](https://github.com/status-im/nim-chronicles)
- I would suggest keeping `library` folder and potentially create these folders as the need for that arises:
  - Create a new `library/core/` folder and move the current `library` content there. libwaku-core will be compiled from there.
  - Create a new `library/sdk` folder that will host the opinionated libwaku-sdk. In that case, my original comment above would make much more sense.

summarizing: disregard my previous comment as it seems to be a premature optimization
  
----

> Is this addressed by the "architecture and api" virtual offsite recommendation? I havent see a forum post.

We are working on it but I think this won't be covered there because that's a very specific for `nwaku`

fryorcraken: > I think we need to rename this module to `libwaku_sdk_conf.nim` instead, to make it clearer it is aimed for the opinionated version of the library and not for the core libwaku

"core libwaku" should not be exposed, and we should deprecate it. the Waku API should be the sole API exposed on FFI, do we agree?

Ivansete-status: > "core libwaku" should not be exposed, and we should deprecate it. the Waku API should be the sole API exposed on FFI, do we agree?

I think it makes sense to only maintain `Waku API`, for the sake of optimizing resources. Nevertheless, `core libwaku` might be still relevant for future integrators. Then, IMHO we cannot strongly state that `core libwaku` should be deprecated. I think it will depend on its future demand.

library/libwaku.nim
```diff
@@ -6,7 +6,7 @@ when defined(linux):
   {.passl: "-Wl,-soname,libwaku.so".}
 
 import std/[json, atomics, strformat, options, atomics]
-import chronicles, chronos, chronos/threadsync
+import chronicles, chronos, chronos/threadsync, results
```
fryorcraken: Need to revert

library/libwaku.nim
fryorcraken, line 1: @Ivansete-status I wonder if we should rename this file to `libwaku_ffi.nim` as it is more focused on FFI than actual library.

Ivansete-status: Yes, I think we can rename that

library/libwaku_api.nim
```diff
@@ -0,0 +1,16 @@
+import chronicles, chronos, results
+
+import waku/factory/waku
+
+import ./libwaku_conf
+
+proc createNode*(config: LibWakuConf): Future[Result[Waku, string]] {.async.} =
```
Ivansete-status: It seems this `createNode` duplicates the following: https://github.com/waku-org/nwaku/blob/d74d29e2f1463289291769e2a2107c64868f9a04/library/waku_thread_requests/requests/node_lifecycle_request.nim#L96

library/libwaku_conf.nim
```diff
@@ -0,0 +1,116 @@
+import
```
Ivansete-status, line 1: If possible, I would not add this file. It seems it adds an additional layer of possible bugs, when sending attributes between layers

Ivansete-status (commented):
Some comments so far. Thanks! :raised_hands:

library/libwaku_api.nim
fryorcraken: The whole point here @Ivansete-status is to start a fresh Waku API, aligned with the spec https://github.com/waku-org/specs/pull/65

There are 2 ways forward:
1. Modify existing functions
2. Create new functions

I went with (2) because until we have create, send, subscribe, it's going to be hard to dogfood. Also, makes it easier to not force upgrade to the new API in status-go integration, etc.

library/libwaku_conf.nim
fryorcraken: We are defining a new API. These is the config for this new API. The existing API uses `WakuNodeConf` which is the CLI args one, and hence not adapted.

The intent is to move away from `WakuNodeConf` and have a config object that works for application developers.

PR #3491, opened 2025-07-03
feat: expose available needed by chatsdk

### LINKED ISSUE #3497: feat: to use `token_bucket` as rate limiter we need to expose the currenct budget and cap
### Problem
We want to use `token_bucket` in `nim_chatsdk` to do so we need to expose the current budget and cap

plopezlpz:
# Description

Exposing available capacity

# Changes

- [x] exposing available capacity for both modes: strict and compensating

## How to test

Unit tests included, no further testing

closes #3497

### DIFF
--- a/tests/common/test_tokenbucket.nim
+++ b/tests/common/test_tokenbucket.nim
@@ -152,3 +152,100 @@ suite "Token Bucket":
     reqTime += 1.minutes
     let (maxCompensationBudget, _, _) = bucket.getAvailableCapacity(reqTime)
     check maxCompensationBudget == 1250 # 1000 + 250 max compensation
+
+  test "TokenBucket custom budget parameter":
+    let now = Moment.now()
+
+    # Test with default budget (-1, should use budgetCap)
+    var bucket1 = TokenBucket.new(1000, 1.seconds, budget = -1)
+    let (budget1, budgetCap1, _) = bucket1.getAvailableCapacity(now)
+    check budget1 == 1000
+    check budgetCap1 == 1000
+
+    # Test with custom budget less than capacity
+    var bucket2 = TokenBucket.new(1000, 1.seconds, budget = 500)
+    let (budget2, budgetCap2, _) = bucket2.getAvailableCapacity(now)
+    check budget2 == 500
+    check budgetCap2 == 1000
+
+    # Test with budget equal to capacity
+    var bucket3 = TokenBucket.new(1000, 1.seconds, budget = 1000)
+    let (budget3, budgetCap3, _) = bucket3.getAvailableCapacity(now)
+    check budget3 == 1000
+    check budgetCap3 == 1000
+
+    # Test with zero budget
+    var bucket4 = TokenBucket.new(1000, 1.seconds, budget = 0)
+    let (budget4, budgetCap4, _) = bucket4.getAvailableCapacity(now)
+    check budget4 == 0
+    check budgetCap4 == 1000
+
+    # Test consumption with custom budget
+    check bucket2.tryConsume(300, now) == true
+    let (budget2After, budgetCap2After, _) = bucket2.getAvailableCapacity(now)
+    check budget2After == 200
+    check budgetCap2After == 1000
+    check bucket2.tryConsume(300, now) == false # Should fail, only 200 remaining
+
+  test "TokenBucket custom lastTimeFull parameter":
+    let now = Moment.now()
+    let pastTime = now - 5.seconds
+
+    # Test with past lastTimeFull (bucket should be ready for refill)
+    var bucket1 = TokenBucket.new(1000, 1.seconds, budget = 0, lastTimeFull = pastTime)
+    # Since 5 seconds have passed and period is 1 second, bucket should refill
+    check bucket1.tryConsume(1000, now) == true
+
+    # Test with current time as lastTimeFull
+    var bucket2 = TokenBucket.new(1000, 1.seconds, budget = 500, lastTimeFull = now)
+    check bucket2.getAvailableCapacity(now) == (500, 1000, now)
+
+    # Test strict mode with past lastTimeFull
+    var bucket3 = TokenBucket.new(
+      1000, 1.seconds, ReplenishMode.Strict, budget = 0, lastTimeFull = pastTime
+    )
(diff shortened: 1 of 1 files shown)


Ivansete-status (approved):
LGTM! Thanks for it! :100:

waku/common/rate_limit/token_bucket.nim
```diff
@@ -109,6 +109,12 @@ proc update(bucket: TokenBucket, currentTime: Moment) =
   else:
     updateStrict(bucket, currentTime)
 
+## Returns the available capacity ratio of the bucket: 0.0 (empty); 1.0 (full)
+proc getAvailableCapacityRatio*(bucket: TokenBucket, currentTime: Moment): float =
```
NagyZoltanPeter:
```suggestion
proc getAvailableCapacityRatio*(bucket: TokenBucket, currentTime = Moment.now()): float =
```
Might ease the use with defaulting to now() as most of the time that's one in question. WDYT?

waku/common/rate_limit/token_bucket.nim
```diff
@@ -109,6 +109,12 @@ proc update(bucket: TokenBucket, currentTime: Moment) =
...
+  if periodElapsed(bucket, currentTime):
+    return 1.0
+  return bucket.budget.float / bucket.budgetCap.float
```
NagyZoltanPeter: Mind that this version of TokenBucket introduces two modes. 
The simple one called `strict` which behaves as expected, calculates buget use absolute to time period.
The other one can take consideration to refil the bugdet  over the capacity based on previous usage.
My guess this calculation will work in case of `strict` mode, otherwise you can get `> 1.0` results.
It's up to you if you would like to manage this case here or not.

NagyZoltanPeter (approved):
LGTM, just added some comment to consider.

waku/common/rate_limit/token_bucket.nim
plopezlpz: true

waku/common/rate_limit/token_bucket.nim
plopezlpz: yes

fryorcraken:
@plopezlpz what is the intent here?

Is the idea to use this as part of the rate limit manager solution?

In this case, I would suggest that we fully extract the code you need in its own library. cc @NagyZoltanPeter

waku/common/rate_limit/token_bucket.nim
```diff
@@ -109,6 +109,22 @@ proc update(bucket: TokenBucket, currentTime: Moment) =
   else:
     updateStrict(bucket, currentTime)
 
+## Returns the available capacity of the bucket: (budget, budgetCap)
```
Ivansete-status: nitpick that I missed: the comments should go under the proc's signature :)

Ivansete-status (approved):
This PR is correct and I don't see any harm in exposing the available capacity.
Just adding an additional nitpick comment but I think we can merge that :)

waku/common/rate_limit/token_bucket.nim
plopezlpz: like in python

NagyZoltanPeter:
> @plopezlpz what is the intent here?
> 
> Is the idea to use this as part of the rate limit manager solution?
> 
> In this case, I would suggest that we fully extract the code you need in its own library. cc @NagyZoltanPeter (doesn't have to be straight away, but end goal should be clear).

I agree, when was making DOS protection and found TockenBucket living in nim-libp2p, that would have been the good choice is to extract it into a separate high level utility library. @fryorcraken: Is that what you mean?
Also the reason I added finally into nim-waku code base was the special needs we had back than (and a bug in the original implementation).

fryorcraken:
> I agree, when was making DOS protection and found TockenBucket living in nim-libp2p, that would have been the good choice is to extract it into a separate high level utility library. @fryorcraken: Is that what you mean?

Yes, If we are re-using the same piece of code in now 3 places (nim-libp2p, nwaku and chat sdk) then it would be best to have one separate library that can be imported in all 3 projects.

Ivansete-status:
Wouldn't it be better to apply the necessary fixes to [nim-chronos](https://github.com/status-im/nim-chronos/blob/0646c444fce7c7ed08ef6f2c9a7abfd172ffe655/chronos/ratelimit.nim) first, and then only use it?

fryorcraken:
> Wouldn't it be better to apply the necessary fixes to [nim-chronos](https://github.com/status-im/nim-chronos/blob/0646c444fce7c7ed08ef6f2c9a7abfd172ffe655/chronos/ratelimit.nim) first, and then only use it?

What does "necessary fixes" mean? I assume that `ratelimit.nim` in chronos has a very specific utility and functionality.
I am not sure it would make sense to change it within chronos and then use it in other application simply because I am not really convinced chronos maintainer would want to see the generalization of their ratelimit.

But this is FOSS, I see 2 valid paths ahead:

1. Do a PR on chronos to generalize the `ratelimit` and make it fit both nwaku and Chat's rate limit manager use cases. Let's then see if chronos' maintainer welcome the change. I would still find it weird to import this functionality via chronos.

2. Create a generic library that is used by both nwaku and chat's rate limit manager. Youi can then offer to chronos maintainer/open a PR to replace their rate limit manager with ours.
3.

fryorcraken:
> I would go with 2., but 2. would be a really small library, I'll create the repo and see, just a copy/paste

Yes, I came back here for that. What are we talking about here? it's 140 lines of code. I don't see a point in blocking @plopezlpz to try doing some code communism here across 3 different layers of the stack.

Up to you @plopezlpz, my recommendation:

if you can use chronos' out of the box, go for it.
else if you can use nwaku's with minimal changes, do it, and once it works with it within nwaku and some potential dirty exposing, we can extract it
else just rewrite your own.

arnetheduck:
> Do a PR on chronos

a well-argued improvement to chronos would certainly be considered - at the end of the day, a token bucket is a well-described algorithm in literature and if chronos' implementation deviates from that, it's certainly welcome to improve it.

_On the other hand_, if the implementation _here_ deviates from literature, it's certainly a good question to ask whether it's due to a well-understood reason or if it's a local decision that later might come back to haunt us.

Issue #3483, opened 2025-07-01
Maintenance Y2025H2


Ivansete-status:
## Description
Gather all maintenance tasks during second half of 2025.
All maintenance PRs should be linked to this so that we don't have orphan PRs.

## PRs
- https://github.com/waku-org/nwaku/pull/3494
- https://github.com/waku-org/nwaku/pull/3486
- https://github.com/waku-org/nwaku/pull/3485
- https://github.com/waku-org/nwaku/pull/3513
- https://github.com/waku-org/nwaku/pull/3514
- https://github.com/waku-org/nwaku/pull/3516
- https://github.com/waku-org/nwaku/pull/3522
- https://github.com/waku-org/nwaku/pull/3525
- https://github.com/waku-org/nwaku/pull/3528
- https://github.com/waku-org/nwaku/pull/3530
- https://github.com/waku-org/nwaku/pull/3355
- https://github.com/waku-org/nwaku/pull/3552
- https://github.com/waku-org/nwaku/pull/3538
- https://github.com/waku-org/nwaku/pull/3559
- https://github.com/waku-org/nwaku/pull/3434
- https://github.com/waku-org/nwaku/pull/3562
- https://github.com/waku-org/nwaku/pull/3583
- https://github.com/waku-org/nwaku/pull/3608
- https://github.com/waku-org/nwaku/pull/3614
- https://github.com/waku-org/nwaku/pull/3622
- https://github.com/waku-org/nwaku/pull/3641
- https://github.com/waku-org/nwaku/pull/3599
- https://github.com/waku-org/nwaku/pull/3596
- https://github.com/waku-org/nwaku/pull/3650
- https://github.com/logos-messaging/nwaku/pull/3659
- https://github.com/logos-messaging/logos-messaging-nim/pull/3664
- https://github.com/logos-messaging/logos-messaging-nim/pull/3612

Ivansete-status:
Closing as we just jumped to another semester :)

PR #3494, opened 2025-07-04
fix: static sharding setup

gabrielmer:
# Description
When setting up the Relay libwaku callback when a node is using static sharding, node setup fails with

```
Could not get autoshards: Static sharding used, cannot get shards from content topics
```

Adding a fix so we don't attempt to get shards from content topics if autosharding isn't enabled

Thanks @fryorcraken for the fix!

# Changes

- [x] don't resolve autoshards when setting app callbacks if autosharding isn't enabled

## Issue

#3483


NagyZoltanPeter (approved):
LGTM

PR #3495, opened 2025-07-07
chore: cleaning waitFor instances

gabrielmer:
# Description

# Changes

- [ ] ...
- [ ] ...

### DIFF
--- a/apps/wakunode2/wakunode2.nim
+++ b/apps/wakunode2/wakunode2.nim
@@ -56,7 +56,7 @@ when isMainModule:
       error "Waku configuration failed", error = error
       quit(QuitFailure)
 
-    var waku = Waku.new(conf).valueOr:
+    var waku = (waitfor Waku.new(conf)).valueOr:
       error "Waku initialization failed", error = error
       quit(QuitFailure)
 
--- a/examples/filter_subscriber.nim
+++ b/examples/filter_subscriber.nim
@@ -77,7 +77,7 @@ proc setupAndSubscribe(rng: ref HmacDrbgContext) {.async.} =
   let node = builder.build().tryGet()
 
   node.mountMetadata(clusterId).expect("failed to mount waku metadata protocol")
-  waitFor node.mountFilterClient()
+  await node.mountFilterClient()
 
   await node.start()
 
--- a/examples/wakustealthcommitments/node_spec.nim
+++ b/examples/wakustealthcommitments/node_spec.nim
@@ -46,7 +46,7 @@ proc setup*(): Waku =
     conf.rlnRelay = twnNetworkConf.rlnRelay
 
   debug "Starting node"
-  var waku = Waku.new(conf).valueOr:
+  var waku = (waitfor Waku.new(conf)).valueOr:
     error "Waku initialization failed", error = error
     quit(QuitFailure)
 
--- a/library/waku_thread_requests/requests/node_lifecycle_request.nim
+++ b/library/waku_thread_requests/requests/node_lifecycle_request.nim
@@ -79,7 +79,7 @@ proc createWaku(
 
   wakuConf.restServerConf = none(RestServerConf) ## don't want REST in libwaku
 
-  let wakuRes = Waku.new(wakuConf, appCallbacks).valueOr:
+  let wakuRes = (await Waku.new(wakuConf, appCallbacks)).valueOr:
     error "waku initialization failed", error = error
     return err("Failed setting up Waku: " & $error)
 
@@ -91,11 +91,16 @@ proc process*(
   defer:
     destroyShared(self)
 
+  # return execution to the async loop before proceeding
+  await sleepAsync(0.milliseconds)
+
   case self.operation
   of CREATE_NODE:
+    echo "------------ calling createWaku"
     waku[] = (await createWaku(self.configJson, self.appCallbacks)).valueOr:
       error "CREATE_NODE failed", error = error
       return err($error)
+    echo "------------ after createWaku"
   of START_NODE:
     (await waku.startWaku()).isOkOr:
       error "START_NODE failed", error = error
--- a/waku/factory/internal_config.nim
+++ b/waku/factory/internal_config.nim
@@ -65,7 +65,7 @@ proc networkConfiguration*(
     dnsAddrsNameServers: seq[IpAddress],
     portsShift: uint16,
     clientId: string,
-): NetConfigResult =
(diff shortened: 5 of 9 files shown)


NagyZoltanPeter:
@gabrielmer What is a deal here? What is the real gain by using await instead of waitFor?
IMO, during initialization, we don't want to proceed and let some part finish async, while waitFor will let the async dispatch loop go.

gabrielmer:
> @gabrielmer What is a deal here? What is the real gain by using await instead of waitFor? IMO, during initialization, we don't want to proceed and let some part finish async, while waitFor will let the async dispatch loop go.

The idea of this came when we saw that when creating a node, the waku thread in libwaku had its execution taken over for a very long time without handing it back to the async loop and wasn't able to receive watchdog requests. At the end, that didn't completely fix the issue but I think that the pattern of using `waitFor` only once, on `wakunode2` is the safest and most flexible pattern.

In other words, `waitFor` should be used in `wakukunode2` but all the underlying functions called should be async. I feel it makes everything more consistent and easiest to reason, and more flexible in case one wants to create a node without that taking over the thread for a long time nonstop.

Although in practice the gain isn't big, so if we think that it's better using many waitFors instead of only one we can ditch this PR :))

NagyZoltanPeter:
> > @gabrielmer What is a deal here? What is the real gain by using await instead of waitFor? IMO, during initialization, we don't want to proceed and let some part finish async, while waitFor will let the async dispatch loop go.
> 
> The idea of this came when we saw that when creating a node, the waku thread in libwaku had its execution taken over for a very long time without handing it back to the async loop and wasn't able to receive watchdog requests. At the end, that didn't completely fix the issue but I think that the pattern of using `waitFor` only once, on `wakunode2` is the safest and most flexible pattern.
> 
> In other words, `waitFor` should be used in `wakukunode2` but all the underlying functions called should be async. I feel it makes everything more consistent and easiest to reason, and more flexible in case one wants to create a node without that taking over the thread for a long time nonstop.
> 
> Although in practice the gain isn't big, so if we think that it's better using many waitFors instead of only one we can ditch this PR :))

Thanks for explaining. It all makes sense than.
I wonder why can waitFor stuck the thread for other async operations like watchdog. :thinking: 
As I know waitFor blocks processing on that point but inside the meaning is to poll the state of the async operation while async thread dispatcher keeps running all the coroutines. Otherwise the whole would not work regardless of where the waitFor is happening.

NagyZoltanPeter (approved):
Thanks for explaining it. LGTM!

gabrielmer:
> Thanks for explaining. It all makes sense than. I wonder why can waitFor stuck the thread for other async operations like watchdog. 🤔 As I know waitFor blocks processing on that point but inside the meaning is to poll the state of the async operation while async thread dispatcher keeps running all the coroutines. Otherwise the whole would not work regardless of where the waitFor is happening.

Great point! Yes, I think there's two things going on - one that actually blocks the execution that might not be related to `waitFor` (although that was my initial hypothesis), and at the same time, if creating a node uses `waitFor` then the following line becomes blocking
https://github.com/waku-org/nwaku/blob/932cbe36909e220cc9d3d33335084470eb77f6ee/library/waku_context.nim#L174

and we wouldn't be able to go back to the top of the loop listening for new requests before the create node request is complete, also causing not being able to do anything until that request is complete.

Or in other words, maybe the `waitFor` isn't what is blocking the async loop, but in any case it makes the call to create a node synchronous and it blocks us from listening to new requests, although that might be solved by re-designing the mechanism in which we listen for requests, which most definitely has a race condition

NagyZoltanPeter:
> > Thanks for explaining. It all makes sense than. I wonder why can waitFor stuck the thread for other async operations like watchdog. 🤔 As I know waitFor blocks processing on that point but inside the meaning is to poll the state of the async operation while async thread dispatcher keeps running all the coroutines. Otherwise the whole would not work regardless of where the waitFor is happening.
> 
> Great point! Yes, I think there's two things going on - one that actually blocks the execution that might not be related to `waitFor` (although that was my initial hypothesis), and at the same time, if creating a node uses `waitFor` then the following line becomes blocking
> 
> https://github.com/waku-org/nwaku/blob/932cbe36909e220cc9d3d33335084470eb77f6ee/library/waku_context.nim#L174
> 
> and we wouldn't be able to go back to the top of the loop listening for new requests before the create node request is complete, also causing not being able to do anything until that request is complete.
> 
> Or in other words, maybe the `waitFor` isn't what is blocking the async loop, but in any case it makes the call to create a node synchronous and it blocks us from listening to new requests, although that might be solved by re-designing the mechanism in which we listen for requests, which most definitely has a race condition

Ok, now I see your problem.
And it is pretty much true. 
But IMO not the asyncSpawn operation is blocking, but as the async operations are truely a cooperative-multitasking (if anyone remembers back Win3.1 :-D). So until one is blocking (does not call await or sleepAsync somewhere) everything will be stopped. This way my assumption is that the waitFor loop of the thread cannot get called. Maybe there is something hidden. 
In any case, it is a good approach. 

What kind of race condition do you think happens here?

arnetheduck:
note sure what this PR is about, but I will just in case note that calling `waitFor` recursively is not allowed - ie if `waitFor` is called (directly or indirectly) from within an `async` function, this will cause difficult issues

NagyZoltanPeter:
> > What kind of race condition do you think happens here?
> 
> The race condition I think we have here is that if signals are received between
> 
> https://github.com/waku-org/nwaku/blob/932cbe36909e220cc9d3d33335084470eb77f6ee/library/waku_context.nim#L169
> 
> and
> 
> https://github.com/waku-org/nwaku/blob/932cbe36909e220cc9d3d33335084470eb77f6ee/library/waku_context.nim#L157
> 
> or in other words, between the moment we send the signal that we received a request and the moment we're back again in `await ctx.reqSignal.wait()` waiting for new signals, we might lose the signals of requests
> 
> other option is to have a thread-safe queue and poll it constantly, although using signaling is more efficient than constantly polling a queue. Maybe we can have a mix of both and poll not so often just in case we lost a signal.
> 
> Not a pressing issue as it's not something it's been problematic in practice, but it's something we have to keep an eye on and improve

Ah, I see. So it's not a classic race condition in general term.
Thanks for explaining.
I'm addressing this in my nim binding demo, exactly similar way. Using a lockless queue implementation. This can be combined with similar signal to avoid polling, but once it is signalled it will process the whole queue and gets back to wait after.

Ivansete-status (approved):
LGTM! Thanks for it! 💯

PR #3505, opened 2025-07-14
chore: update CI 

darshankabariya:
While working on the documentation, I noticed that the Windows CI gets triggered even when it’s not necessary—like during README or docs-only updates. It would be better if we could skip it for such changes.


gabrielmer (approved):
Niiiice, thank you!

Ivansete-status (approved):

PR #3506, opened 2025-07-15
chore: chat2 with mix

chaitanyaprem:
# Description
Building a chat2 example to dogfood lightpush with mix.
Still WIP. It will be ready once replies are implemented.

# Changes

- [x] add a mixify flag to lightpush publish method
- [x] create a chat2 example that publishes with mix
~- [ ] use rendezvous instead of discv5+peer-exchange to see if mix nodes get detected faster~
- [ ] remove unused code and config to make the app lean

**Note: This PR only will be merged into mix feature branch and not master.**

### DIFF
--- a/apps/chat2mix/config_chat2mix.nim
+++ b/apps/chat2mix/config_chat2mix.nim
@@ -1,15 +1,14 @@
+import chronicles, chronos, std/strutils, regex
+
 import
-  chronicles,
-  chronos,
-  confutils,
-  confutils/defs,
-  confutils/std/net,
   eth/keys,
   libp2p/crypto/crypto,
   libp2p/crypto/secp,
   nimcrypto/utils,
-  std/strutils,
-  regex
+  confutils,
+  confutils/defs,
+  confutils/std/net
+
 import waku/waku_core
 
 type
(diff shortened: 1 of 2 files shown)


apps/chat2mix/config_chat2mix.nim
```diff
@@ -0,0 +1,359 @@
...
+      desc: "Peer multiaddr to query for storage.", defaultValue: "", name: "storenode"
+    .}: string
+
+    ## Filter config
+    filter* {.
+      desc: "Enable filter protocol: true|false", defaultValue: false, name: "filter"
+    .}: bool
+
+    ## Lightpush config
+    lightpush* {.
```
fryorcraken, line 129: I would avoid copy-pasting the whole `WakuNodeConf` object.

Especially I assume that you would want to use lightpush over mix so this option does not make sense.

chaitanyaprem: makes sense, i just copied the chat2 app and started working on top of that. 
I will include a cleanup step so that things which are not needed can be removed before merging.

Makefile
```diff
@@ -228,6 +228,10 @@ chat2: | build deps librln
 	echo -e $(BUILD_MSG) "build/$@" && \
 		$(ENV_SCRIPT) nim chat2 $(NIM_PARAMS) waku.nims
 
+chat2mix: | build deps librln
```
jm-clius, line 231: Just to understand, why a separate chat2 target and not simply adding a `mixEnabled` or similar config to the existing `chat2`? I know `chat2` is already a fairly ugly monster, but it might make the changes required for mix a bit clearer.

apps/chat2mix/run_chat_mix.sh
jm-clius, line 1: Wouldn't these scripts make more sense as part of the `simulations/mixnet`? In other words, presumably `chat2mix` will work when connected to any mixnet constellation, but this is a helper script that asssumes a local simulated setup. I would also add a readme to the simulation to explain which scripts to run and what they do.

waku/node/waku_node.nim
```diff
@@ -1182,17 +1183,36 @@ proc lightpushPublishHandler(
...
+      let conn = MixEntryConnection.newConn(
+        $peer.addrs[0], #TODO: How to handle multiple addresses?
+        peer.peerId,
+        ProtocolType.fromString(WakuLightPushCodec),
+        node.mix,
+      )
+      return await node.wakuLightpushClient.publishWithConn(
+        pubsubTopic, message, conn, peer.peerId
+      )
+    else:
```
jm-clius, line 1207: To me creating the mix connection belongs to `wakuLightpushClient`. Could you add `mixify` as argument to `publish()` that then proceeds to attempt to create a mix entry connection?

jm-clius (approved):
Thanks! I've added some comments, but I don't think they're blocking. Perhaps make sure that other reviewers notice that this PR only merges to a feature branch, not `master`. :)

Makefile
chaitanyaprem: Well, it was using relay to begin with and the code was pretty cluttered. I just wanted freedom to remove and comment code to make it work :)
 
I agree that it might make changes required for mix more clearer but i had to change the app first to work as edgeNode whcih itself had nothing to do with mix and then add mix functionality. 

Let me see if i can work on existing chat2 itself to bring it to a sane state.  Will create a separate PR so not to mess this up and if that is ready, maybe we can discard this PR.

apps/chat2mix/run_chat_mix.sh
chaitanyaprem: agreed, i had a doubt while putting them in chat2, will move them to simulations folder as they are indeed tied to the simulation.

Will add a readme as well. Thanks!

waku/node/waku_node.nim
chaitanyaprem: Well, i kept it out of lightpush as of now so that it isn't dependent on mix specifically. And since MixProtocol instance is part of WakuNode it made sense to keep it at wakunode level. Otherwise we will have to pass this instance to all protos that want to use mix. 
It is a minor thing though and can be changed if required.

apps/chat2mix/config_chat2mix.nim
```diff
@@ -0,0 +1,363 @@
...
+  chronos,
+  confutils,
+  confutils/defs,
+  confutils/std/net,
+  eth/keys,
+  libp2p/crypto/crypto,
+  libp2p/crypto/secp,
+  nimcrypto/utils,
+  std/strutils,
+  regex
```
Ivansete-status: We usually split the imports in three big categories:
import cat1
import cat2
import cat3

cat1: contains the std and global modules
cat2: contains `vendor` modules ( libp2p, chronos, chronicles ... )
cat3: contains local nwaku modules

apps/chat2mix/config_chat2mix.nim
```diff
@@ -0,0 +1,363 @@
...
+      desc: "Enable spam protection through rln-relay: true|false",
+      defaultValue: false,
+      name: "rln-relay"
+    .}: bool
+
+    rlnRelayChainId* {.
+      desc:
+        "Chain ID of the provided contract (optional, will fetch from RPC provider if not used)",
+      defaultValue: 0,
+      name: "rln-relay-chain-id"
```
Ivansete-status: Shall remove this whole commented snippet?

Ivansete-status (approved):
LGTM! Thanks! :100: 
Just a couple of nits

apps/chat2mix/config_chat2mix.nim
fryorcraken: something to add to `CONTRIBUTE.md` ?

chaitanyaprem: > something to add to `CONTRIBUTE.md` ?

Not sure if we have something like this, but it is a good idea to indicate nim guidelines that we follow in nwaku. 
Something in the lines of [Status-guide](https://status-im.github.io/nim-style-guide) with other things that we follow?

PR #3507, opened 2025-07-15
fix: libwaku received signal

gabrielmer:
# Description
Sending received signal in libwaku only after the request is processed

## Issue
#3076


darshankabariya (commented):
Thanks ! LGTM

Ivansete-status (approved):

PR #3508, opened 2025-07-15
fix: sync dashboard panel

SionoiS:
# Description

The last Store sync panel had a wrong query and label. I fixed it, looks like that now.

Screenshot from 2025-07-15 11-03-54

### DIFF
--- a/metrics/waku-fleet-dashboard.json
+++ b/metrics/waku-fleet-dashboard.json
@@ -450,7 +450,7 @@
             "h": 9,
             "w": 12,
             "x": 0,
-            "y": 1074
+            "y": 197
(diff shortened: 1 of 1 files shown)


Ivansete-status (approved):
LGTM!
If possible, let's elaborate more about the information being shown
Also, notice that, in order to make the changes effective, there should be a manual push within Grafana

SionoiS:
> LGTM! If possible, let's elaborate more about the information being shown 

it's the number of messages cached by store sync nodes

> Also, notice that, in order to make the changes effective, there should be a manual push within Grafana

How do I do that?

PR #3509, opened 2025-07-15
chore: improve connection proc

gabrielmer:
# Description
Simplifying connection operation in libwaku and removing 5 second sleep in `connectToNodes`. If a sleep is needed, it's the responsibility of the caller to add it or to check if their node is connected.

Note: if tests fail because of removing this sleep, will add the sleep to the tests

# Changes

- [x] simplifying libwaku's `CONNECT_TO` operation
- [x] removing timeout in `connectToNodes`

## Issue

- #3483

### DIFF
--- a/library/waku_thread_requests/requests/peer_manager_request.nim
+++ b/library/waku_thread_requests/requests/peer_manager_request.nim
@@ -80,10 +80,9 @@ proc process*(
 
   case self.operation
   of CONNECT_TO:
-    let ret = waku.node.connectTo($self[].peerMultiAddr, self[].dialTimeout)
-    if ret.isErr():
-      error "CONNECT_TO failed", error = ret.error
-      return err(ret.error)
+    let peers = ($self[].peerMultiAddr).split(",").mapIt(strip(it))
+    await waku.node.connectToNodes(peers, source = "static")
+    return ok("")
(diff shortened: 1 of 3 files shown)


Ivansete-status (approved):
Great catch! Thanks so much for it! :raised_hands:

NagyZoltanPeter (approved):
LGTM!!!

library/waku_thread_requests/requests/peer_manager_request.nim
```diff
@@ -80,10 +64,9 @@ proc process*(
 
   case self.operation
   of CONNECT_TO:
```
fryorcraken, line 66: Why is it exposed? Peer management is Waku's domain, not status-go.

gabrielmer: I think it should be exposed even if `status-go` shouldn't use it, it makes sense to add the possibility to dynamically connect to peers and not only the ones that are included on the configuration parameters.

Regarding `status-go`, I believe that as long as there's reliability logic in Go we would probably still need to support that. `status-go` exposes API functions for dialing/connecting to nodes and makes use of them. We will need to remove that logic from there but I think it should most likely need to be done after not using `go-waku` anymore and making sure all the removed logic exists in nwaku.

PR #3510, opened 2025-07-16
chore: test config

darshankabariya:
Found a bug during testing: if the test name contains any special characters, it doesn’t match correctly.  
Screenshot 2025-07-16 at 15 07 10


gabrielmer (approved):
Thank you!

fryorcraken (approved):

PR #3511, opened 2025-07-16
chore: mark keep-alive as deprecated

Ivansete-status:
## Description
Simple PR to mark the `keep-alive` as deprecated and not break the current possible use of it


Ivansete-status:
> Would be nice to also print warnings on usage of deprecated params. But that's out of scope.

Yup great point! Thanks for the heads-up btw!

fryorcraken:
Maybe good to ensure it's not use in any fleet or nwaku-compose?

PR #3513, opened 2025-07-17
fix: tests using fetchMerkleRoot

stubbsta:
# Description
During RLN related updates some functionality was changed and it is now needed to add a `waitFor` when calling the `fetchMerkleRoot` function, because it returns a `Future`.

This fix is needed for 1 test passing without actually doing anything. (`register: should register successfully`) 
In this test the `waku/waku_rln_relay/group_manager/on_chain/group_manager.nim` `register` function being called was:
```
method register*(
    g: OnchainGroupManager, rateCommitment: RateCommitment
): Future[void] {.async: (raises: [Exception]).} =
```
but this if statement was never true:
```
 if g.registerCb.isSome():
```

Other tests also needed updates, but were running and passing properly. (`trackRootChanges: should sync to the state of the group` and `trackRootChanges: should fetch history correctly`)

# Changes

`tests/waku_rln_relay/test_rln_group_manager_onchain.nim`
Add wairFor and use normal register function that does not use thecallback.

## How to test

1. run test file `test_rln_group_manager_onchain.nim`

### DIFF
--- a/tests/waku_rln_relay/test_rln_group_manager_onchain.nim
+++ b/tests/waku_rln_relay/test_rln_group_manager_onchain.nim
@@ -135,7 +135,7 @@ suite "Onchain group manager":
     (waitFor manager.init()).isOkOr:
       raiseAssert $error
 
-    let merkleRootBefore = manager.fetchMerkleRoot()
+    let merkleRootBefore = waitFor manager.fetchMerkleRoot()
 
     try:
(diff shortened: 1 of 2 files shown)


Ivansete-status (approved):
LGTM! Thanks for it! :100:

stubbsta:
@Ivansete-status, are the failing CI checks known issues and can I go ahead with merging?

PR #3514, opened 2025-07-18
Update networks_config.nim

Ivansete-status:
## Description

When incorrect sharding config is passed, the following message can appear:

```nim
ERR 2025-07-18 10:20:44.808+02:00 validateShards failed                      topics="waku networks conf" tid=17834 file=networks_config.nim:70 error="validateShards invalid shard: 32 when numShardsInCluster: 1"
ERR 2025-07-18 10:20:44.808+02:00 Waku configuration failed                  topics="wakunode main" tid=17834 file=wakunode2.nim:56 error="validateShards invalid shard: 32 when numShardsInCluster: 1
```

This aims to reflect the exact paramater that needs to be properly configured

## Issue

### DIFF
--- a/waku/factory/networks_config.nim
+++ b/waku/factory/networks_config.nim
@@ -65,7 +65,7 @@ proc validateShards*(
     for shard in shards:
       if shard >= numShardsInCluster:
         let msg =
-          "validateShards invalid shard: " & $shard & " when numShardsInCluster: " &
+          "validateShards invalid shard: " & $shard & " when num-shards-in-network: " &
           $numShardsInCluster
         error "validateShards failed", error = msg
         return err(msg)


waku/factory/networks_config.nim
```diff
@@ -65,7 +65,7 @@ proc validateShards*(
     for shard in shards:
       if shard >= numShardsInCluster:
         let msg =
-          "validateShards invalid shard: " & $shard & " when numShardsInCluster: " &
+          "validateShards invalid shard: " & $shard & " when num-shards-in-network: " &
```
fryorcraken, line 68: You are assuming that the user is using `wakunode2` and passing `wakunode2` CLI arg name.

A developer will get confused because there would not be a `num-shards-in-network` option in the lib.

Ivansete-status: Yes, you are right that a dev might get consumed. This is pending to be defined from lib PoV.

But the change makes sense from the node operator's PoV, isn't it?

fryorcraken: Yes but nwaku must become a library to integrate Waku in applications, and `wakunode2`/service node is only ONE application that uses such library. 
We need to progress towards nwaku as an application library, and this needs to become the highest priority.
And it is fine for this to happen to the detriment of service nodes/nwaku, because we are the **only** developers building a service node for Waku.

PR #3519, opened 2025-07-21
Remove metadata protocol dependency on enr, relax check when nwaku is edge node

chaitanyaprem:
# Description
While working on chat2 for mix, noticed that nwaku acting as only light/edge node disconnects from relay node when metadata request is received indicating shards mismatch. The only way seems to be when shards are passed which is not necessary for an edge node. 

nwaku acting as relay has a check not to disconnect in case of no shards from a non-relay node, but vice-versa seems to be not present. 

There is also a need for a follow-up PR if possible to remove ENR initialization in case relay and discv5 are not enabled. 

# Changes

- [ ] Modified metadata to only take clusterID and shards as input for initialization rather than ENR
- [ ] Added additional check in peerMAnager to not disconnect when local node is acting as light-client/edge-node(with no shards defined) and receives metadata from a relay peer.


waku/node/peer_manager/peer_manager.nim
```diff
@@ -651,12 +651,16 @@ proc onPeerMetadata(pm: PeerManager, peerId: PeerId) {.async.} =
...
-      pm.switch.peerStore.hasPeer(peerId, WakuRelayCodec) and
-      not metadata.shards.anyIt(pm.wakuMetadata.shards.contains(it))
+      (
+        pm.switch.peerStore.hasPeer(peerId, WakuRelayCodec) and localShards.len != 0 and
+        not metadata.shards.anyIt(localShards.contains(it))
+      )
     ):
-      let myShardsString = "[ " & toSeq(pm.wakuMetadata.shards).join(", ") & " ]"
+      # scenario where nwaku is acting as a light node connecting to a relay and receives metadata request.
+      # as of now we disconnect the peer as local shards doesn't match remote shards which is wrong.
```
fryorcraken: You are doing a negative comment related to previous behaviour.

Focus your comment on the new behaviour.

waku/node/peer_manager/peer_manager.nim
```diff
@@ -651,12 +651,16 @@ proc onPeerMetadata(pm: PeerManager, peerId: PeerId) {.async.} =
         "different clusterId reported: " & $pm.wakuMetadata.clusterId & " vs " &
         $clusterId
       break guardClauses
-
+    let localShards = pm.wakuMetadata.getShards()
     if (
-      pm.switch.peerStore.hasPeer(peerId, WakuRelayCodec) and
-      not metadata.shards.anyIt(pm.wakuMetadata.shards.contains(it))
+      (
```
fryorcraken: I would say to **not** disconnect based on shards value. There is no scenario where it makes sense.

It's more of an *optimisation*, ie, maintain a good number of peers in shards of interest vs shards of not interest, but that should be a separate logic, not done at metadata req time.

waku/node/peer_manager/peer_manager.nim
```diff
@@ -651,12 +651,16 @@ proc onPeerMetadata(pm: PeerManager, peerId: PeerId) {.async.} =
         "different clusterId reported: " & $pm.wakuMetadata.clusterId & " vs " &
         $clusterId
       break guardClauses
```
fryorcraken, line 653: is that like a "go to" where it jumps to after trhe end of the block?

waku/node/peer_manager/peer_manager.nim
```diff
@@ -787,7 +791,7 @@ proc getOnlineStateObserver*(pm: PeerManager): OnOnlineStateChange =
 #~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~#
 
 proc manageRelayPeers*(pm: PeerManager) {.async.} =
-  if pm.wakuMetadata.shards.len == 0:
+  if pm.wakuMetadata.getShards().len == 0:
     return
```
fryorcraken: metadata shards should not be used to manage relay peers, libp2p-gossipsub already handles this. Remove these two lines.

waku/node/waku_node.nim
```diff
@@ -187,7 +187,17 @@ proc mountMetadata*(node: WakuNode, clusterId: uint32): Result[void, string] =
   if not node.wakuMetadata.isNil():
     return err("Waku metadata already mounted, skipping")
 
-  let metadata = WakuMetadata.new(clusterId, node.enr, node.topicSubscriptionQueue)
+  var shards = HashSet[uint32]()
+  #TODO: Ideally ENR would be nil for edge nodes, need to check that.
+  let enrRes = node.enr.toTyped()
```
fryorcraken: You are relying on ENR to decide what to do with metadata. I don't think it's the right approach. too much indirection, also, I think there should be different values:

ENR: if you run relay, then add RS to whatever shards you run on relay

Simply because the intent here is for any node (relay, edge) to find other nodes on appropriate shards

Metadata:
1. if relay is enabled
2. AND any shard-related service is mounted (store, light push, filter)
    then fill in shards
otherwise, do not fill shards

Simply because you are connected to a node (and running metadata), gossipsub can take over the relay part.

Only light protocol client nodes need to use metadata's shards to know which node to query for store,  light push or filter. Which means that only service node need to advertise their shards.

And again, there should not be any disconnection logic pure based on shard.

---

Only logic we could have to disconnect is a later optimisation to ensure we have a good numbers of peers in shards of interest.

waku/waku_metadata/protocol.nim
```diff
@@ -104,21 +109,11 @@ proc initProtocolHandler(m: WakuMetadata) =
 proc new*(
     T: type WakuMetadata,
     clusterId: uint32,
-    enr: Record,
-    queue: AsyncEventQueue[SubscriptionEvent],
+    shards: Option[HashSet[uint32]],
```
fryorcraken: I agree on stopping usage of `enr`.

But in this instance, I don't think an `Option` make sense because as per your other code change
`none === []`. Just ask the caller to pass an empty set.

waku/node/peer_manager/peer_manager.nim
chaitanyaprem: yea...seems that way. 
Because a block has been named guardClass and everytime we break with the block name it just breaks out of it. kind-of like breaking out of a loop.

waku/waku_metadata/protocol.nim
chaitanyaprem: Was little confused as to do with nullset or option, looks like finally caller is anyways passing null set. I can remove the option here.

waku/node/waku_node.nim
chaitanyaprem: Hmm, any challenge you see relying on ENR? because ENR is used during discovery hence it will anyways have relevant info (i.e shards to be supported etc for service nodes). 

The rest of the points, completely agree with you. Will change behaviour.

waku/node/peer_manager/peer_manager.nim
chaitanyaprem: well, this seems like a check kept to identify whether relay is mounted or not...i don't like it, but was thinking that should be handled in a different PR as i don't want to do too many things in this one. 
Will add a comment so that it needs to be changed. 

Also, ideally peerManager should manage either relay peers or service nodes. 
So the check should be based off of config mode with which node is initialized with.
I remember we implemented something similar in go-waku.

waku/node/peer_manager/peer_manager.nim
chaitanyaprem: ah...forgot to update the comment. I added it initially as a TODO, but did not re-review it.
Will change it. Thanks!

waku/node/waku_node.nim
fryorcraken: I would say it's better code, because better separation of concerns. There should not be a dependence of metadata on ENR.

fryorcraken:
I suggest we spec the desired behaviour, see https://github.com/vacp2p/rfc-index/pull/168

waku/node/waku_node.nim
chaitanyaprem: ok, then the only way is to directly read from config and then pass it to metadata, rather than linking to ENR. Will see if it is a straight-forward change, otherwise will do it in a separate PR.

jm-clius: Hopefully as simple as initialising with `shards` config item. https://github.com/waku-org/nwaku/blob/master/waku/factory/external_config.nim#L331
The `topicSubscriptionQueue` could take care of any subscriptions being added or removed afterwards.

jm-clius (approved):
Great! Looks like a very reasonable simplification to me. I also agree that not relying on ENR for shards information when mounting `metadata` would be preferred, but this already improves things.

waku/node/waku_node.nim
fryorcraken: > Hopefully as simple as initialising with `shards` config item. [`master`/waku/factory/external_config.nim#L331](https://github.com/waku-org/nwaku/blob/master/waku/factory/external_config.nim?rgh-link-date=2025-07-22T13%3A21%3A41Z#L331) The `topicSubscriptionQueue` could take care of any subscriptions being added or removed afterwards.

we use `WakuConf` now https://github.com/waku-org/nwaku/blob/d3f2bd010e9db6af9a2902c74cc83a71b7e30bd9/waku/factory/waku_conf.nim#L80C3-L80C18

waku/waku_metadata/protocol.nim
```diff
@@ -129,10 +119,10 @@ proc new*(
 proc subscriptionsListener(wm: WakuMetadata) {.async.} =
   ## Listen for pubsub topics subscriptions changes
 
-  let key = wm.topicSubscriptionQueue.register()
+  let key = wm.topicSubscriptionQueue.get().register()
```
Ivansete-status: This may eventually crash if `wm.topicSubscriptionQueue` is `None`.
Shall we avoid using `Option` in this case?

waku/node/peer_manager/peer_manager.nim
```diff
@@ -652,17 +652,6 @@ proc onPeerMetadata(pm: PeerManager, peerId: PeerId) {.async.} =
...
-      pm.switch.peerStore.hasPeer(peerId, WakuRelayCodec) and
-      not metadata.shards.anyIt(pm.wakuMetadata.shards.contains(it))
-    ):
-      let myShardsString = "[ " & toSeq(pm.wakuMetadata.shards).join(", ") & " ]"
-      let otherShardsString = "[ " & metadata.shards.join(", ") & " ]"
-      reason =
-        "no shards in common: my_shards = " & myShardsString & " others_shards = " &
-        otherShardsString
-      break guardClauses
-
```
Ivansete-status, line 665: EDITED: disregard the following comment. The suggested change is aligned with https://github.com/vacp2p/rfc-index/pull/168

Even though the [metadata spec](https://github.com/vacp2p/rfc-index/blob/main/waku/standards/core/66/metadata.md) doesn't mention that explicitly I think we should keep that validation because it doesn't make sense to keep connection if there are no shards in common ( cc @jm-clius .)

waku/node/waku_node.nim
```diff
@@ -183,11 +183,14 @@ proc disconnectNode*(node: WakuNode, remotePeer: RemotePeerInfo) {.async.} =
 
 ## Waku Metadata
 
-proc mountMetadata*(node: WakuNode, clusterId: uint32): Result[void, string] =
+proc mountMetadata*(
+    node: WakuNode, clusterId: uint32, shards: seq[uint16]
```
Ivansete-status, line 187: The [spec](https://github.com/vacp2p/rfc-index/blob/main/waku/standards/core/66/metadata.md) recommends uint32, hence we don't need to conversion with `mapIt`
```suggestion
    node: WakuNode, clusterId: uint32, shards: seq[uint32]
```

Ivansete-status (commented):
Thanks for it! :raised_hands: 
I'm just adding some nitpicks comments/questions

waku/node/waku_node.nim
chaitanyaprem: hmm, as per sharding spec max shard number will be `uint16` so it makes sense to leave the config at that. 
But metadata spec for some reason chose uint32, i wouldn't want to change protobuf right away as that woul involve changes in all clients. 
Maybe something to review when metadata spec gets updated next. 
cc @jm-clius

waku/waku_metadata/protocol.nim
chaitanyaprem: that shouldn't be an issue as there is a check in line 147 to start this only if queue is set and not start otherwise.

waku/node/peer_manager/peer_manager.nim
chaitanyaprem: well, if there are no shards in common the connection would anyway get disconnected from idle timeout at transport layer. having this check is causing more issues which is why imo its better to remove it unless we really need it.

waku/factory/node_factory.nim
```diff
@@ -154,7 +154,12 @@ proc setupProtocols(
   ## Optionally include persistent message storage.
   ## No protocols are started yet.
 
-  node.mountMetadata(conf.clusterId).isOkOr:
+  var allShards = conf.subscribeShards
+  for shardKey in conf.protectedShards:
```
fryorcraken: You don't need to do that, because a `protectedShards` not in `subscribeShards` will error out nwaku already.

waku/node/waku_node.nim
fryorcraken: We definitely need to sort clusterId and shard id typing across nwaku codebase.....

Also, uint16 should be enough for those data, spec suggesting u32 is ill-advised.

waku/waku_metadata/protocol.nim
fryorcraken: I would agree it would be best to keep the check near the usage of `.get()`.

the issue with having the check done outside this function body, is that it would be easy to introduce a crash by calling `subscriptionsListener` and forgetting to check that `topicSubscriptionQueue` is `some`.

What you could do is unwrap `topicSubscriptionQueue` and pass it to `subscriptionsListener` directly, instead of getting it from `wm`.

waku/waku_metadata/protocol.nim
```diff
@@ -143,18 +138,25 @@ proc subscriptionsListener(wm: WakuMetadata) {.async.} =
...
         continue
 
-  wm.topicSubscriptionQueue.unregister(key)
+  wm.topicSubscriptionQueue.get().unregister(key)
 
 proc start*(wm: WakuMetadata) =
   wm.started = true
-
-  asyncSpawn wm.subscriptionsListener()
+  #TODO: ideally in edge nodes, same queue should be linked for filter subscriptions??
```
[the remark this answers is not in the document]
fryorcraken: what do you mean?

filter subscription should not impact what you set in metadata payload. Only relay shard subscriptions matter.

waku/factory/node_factory.nim
chaitanyaprem: ah..did not know that. isn't it duplicate though asking user to include a shard in both protected and subscribe shards?

waku/waku_metadata/protocol.nim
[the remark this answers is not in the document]
chaitanyaprem: well for an edgenode since relay is not mounted adding shards might make sense. But now since we are having the understanding of not to include shards this is not needed. Will remove the TODO.

waku/factory/node_factory.nim
fryorcraken: Protected shards just need to be trashed so don't waste too many neurons on it.

Who is your user here?

Yes, it makes sense to say "shard 1, 3, 5 are protected"
and separately say:
- subscribe to shard 2
- now 4
- now 5 -> oh I know it's protected.

chaitanyaprem:
> I find it's an odd design pattern where you:
> 
>     1. set some start up information (`conf.shards`)
> 
>     2. Add an event queue "in case it changes".
> 
> 
> can't we have a pattern instead where we have access to an API (callback) where the metadata protocol can ask to relay "what shards are you subscribed to"?

Well this was the existing pattern and did not think of the need to change it.

waku/factory/node_factory.nim
chaitanyaprem: I don't see this validation, maybe i am missing something here.

chaitanyaprem: ok, i see that shard is skipped for adding validator ..will change this.

fryorcraken (approved):

PR #3521, opened 2025-07-22
chore: add conf for peer-exchange to not filter peers not found via discv5

chaitanyaprem:
# Description
When running simulations with mix, i had noticed that it takes a lot of time for peers to be discovered by a lightnode. This happens because in a simulation setup, we end up configuring static peers which connect to each other and are not discovered by discv5. Hence peer-exchange filters-out the peers as such and not returns them to a light/edge node. 
This causes delay in testing changes during simulations, because each time we have to wait for a long time for dicv5 to catchup and then peers are discovered. 

In case of mix, this becomes more important as we want a pool of nodes to be discovered before we can start publishing.

Added a config that disables the strict peer checking in peer-exchange service side and returns peers which are statically configured as well. 

TODO: Even after this change, time is taken for peer-exchange to return the peers indicating that peers don't have ENR. Will investigate further and see if there is an issue with logic of peers configured statically and fix that as well.
Else we don't need this flag itself.

# Changes

- [x] Add a new config `peer-exchange-strict-filter` which disables strict check of returning only peers discovered via discv5.
- [ ] Fix any issues wrt peers statically configured not having ENR updated after being identified via discv5

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -228,6 +228,10 @@ chat2: | build deps librln
 	echo -e $(BUILD_MSG) "build/$@" && \
 		$(ENV_SCRIPT) nim chat2 $(NIM_PARAMS) waku.nims
 
+chat2mix: | build deps librln
+	echo -e $(BUILD_MSG) "build/$@" && \
+		$(ENV_SCRIPT) nim chat2mix $(NIM_PARAMS) waku.nims
+
 rln-db-inspector: | build deps librln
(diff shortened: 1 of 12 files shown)


chaitanyaprem:
This PR is not required as discovery improvements are in progress.
As a workaround for the mix simulation statically configured nodes are being used.

PR #3522, opened 2025-07-22
chore: Adapt heaptrack support builds and description to latest nim 2.2.4 compiler

NagyZoltanPeter:
# Description

With the latest nimbus-build-system we started using along with Nim 2.2.4
The breaking change coming with nimbus-build-system is that we moved away Status's fork of Nim but now use the main repository, thus we can not push our patched branch.

# Changes

To circumvent the new situation with patching Nim:

- added patch file that can be easily used to make Nim core library heaptrack supporting.
- Updated heatrack.md description
- Updated Dockerfile to apply the patch on the fly when needed.
- Updated Makefile: removed un-necessary NIM_COMMIT and adapted arguments for docker builds.

# How to test

- checkout nwaku
- issue `make HEAPTRACKER=1 docker-image` or `make HEAPTRACKER=1 docker-quick-image`
- run as described in heaptracker.md document

# issue

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,6 +5,7 @@ ARG NIMFLAGS
 ARG MAKE_TARGET=wakunode2
 ARG NIM_COMMIT
 ARG LOG_LEVEL=TRACE
+ARG HEAPTRACK_BUILD=0
 
 # Get build tools and required header files
 RUN apk add --no-cache bash git build-base openssl-dev pcre-dev linux-headers curl jq
@@ -17,6 +18,10 @@ RUN apk update && apk upgrade
 
 # Ran separately from 'make' to avoid re-doing
 RUN git submodule update --init --recursive
+
+RUN if [ "$HEAPTRACK_BUILD" = "1" ]; then \
+      git apply --directory=vendor/nimbus-build-system/vendor/Nim docs/tutorial/nim.2.2.4_heaptracker_addon.patch; \
+    fi
 
 # Slowest build step for the sake of caching layers
 RUN make -j$(nproc) deps QUICK_AND_DIRTY_COMPILER=1 ${NIM_COMMIT}
--- a/Makefile
+++ b/Makefile
@@ -93,10 +93,8 @@ NIM_PARAMS := $(NIM_PARAMS) -d:git_version=\"$(GIT_VERSION)\"
 HEAPTRACKER ?= 0
 HEAPTRACKER_INJECT ?= 0
 ifeq ($(HEAPTRACKER), 1)
-# Needed to make nimbus-build-system use the Nim's 'heaptrack_support' branch
-DOCKER_NIM_COMMIT := NIM_COMMIT=heaptrack_support_v2.0.12
+# Assumes Nim's lib/system/alloc.nim is patched!
 TARGET := debug-with-heaptrack
-NIM_COMMIT := heaptrack_support_v2.0.12
 
 ifeq ($(HEAPTRACKER_INJECT), 1)
 # the Nim compiler will load 'libheaptrack_inject.so'
@@ -349,6 +347,7 @@ docker-image:
 		--build-arg="NIMFLAGS=$(DOCKER_IMAGE_NIMFLAGS)" \
 		--build-arg="NIM_COMMIT=$(DOCKER_NIM_COMMIT)" \
 		--build-arg="LOG_LEVEL=$(LOG_LEVEL)" \
+		--build-arg="HEAPTRACK_BUILD=$(HEAPTRACKER)" \
 		--label="commit=$(shell git rev-parse HEAD)" \
 		--label="version=$(GIT_VERSION)" \
 		--target $(TARGET) \
@@ -357,6 +356,7 @@ docker-image:
 docker-quick-image: MAKE_TARGET ?= wakunode2
 docker-quick-image: DOCKER_IMAGE_TAG ?= $(MAKE_TARGET)-$(GIT_VERSION)
 docker-quick-image: DOCKER_IMAGE_NAME ?= wakuorg/nwaku:$(DOCKER_IMAGE_TAG)
+docker-quick-image: NIM_PARAMS := $(NIM_PARAMS) -d:chronicles_colors:none -d:insecure -d:postgres --passL:$(LIBRLN_FILE) --passL:-lm
 docker-quick-image: | build deps librln wakunode2
 	docker build \
 		--build-arg="MAKE_TARGET=$(MAKE_TARGET)" \
--- a/docs/tutorial/heaptrack.md
+++ b/docs/tutorial/heaptrack.md
@@ -88,7 +88,7 @@ That will create a Docker image with both nwaku and heaptrack. The container's e
 
 #### Notice for using heaptrack supporting image with `docker compose`
 
-Take case of the wakunode2 should started as
+Take care that wakunode2 should be started as
 ```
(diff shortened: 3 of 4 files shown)


Makefile
```diff
@@ -357,6 +356,7 @@ docker-image:
 docker-quick-image: MAKE_TARGET ?= wakunode2
 docker-quick-image: DOCKER_IMAGE_TAG ?= $(MAKE_TARGET)-$(GIT_VERSION)
 docker-quick-image: DOCKER_IMAGE_NAME ?= wakuorg/nwaku:$(DOCKER_IMAGE_TAG)
+docker-quick-image: NIM_PARAMS := $(NIM_PARAMS) -d:chronicles_colors:none -d:insecure -d:postgres --passL:$(LIBRLN_FILE) --passL:-lm
```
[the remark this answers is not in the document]
NagyZoltanPeter: It was an intentional fix here.

docs/tutorial/heaptrack.md
```diff
@@ -40,13 +40,26 @@ It operates in two modes:
...
+git apply --directory=vendor/nimbus-build-system/vendor/Nim docs/tutorial/nim.2.2.4_heaptracker_addon.patch
+git add .
+git commit -m "Add heaptrack support to Nim compiler - temporary patch"
+```
+
+> We are never going to merge this patch into the official Nim repository, so it is important to keep it in the `nimbus-build-system` repository.
+> Commit ensures that `make update` will not override the patch unintentionally.
 
-The make command should have the 'NIM_COMMIT' setting as:
+> We are planning to make it avail through an official PR for Nim.
```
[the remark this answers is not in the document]
NagyZoltanPeter: This is just a notice for reader...

Ivansete-status:
Thanks for the comments @arnetheduck !
From our recent experiments, we didn't detect that double-alloc.

It seems `heaptrack` just jots down in a binary struct the alloc info of the child process it spawned:
https://github.com/KDE/heaptrack/blob/b54e92e88b0895f910a5b107f499d1155a086b80/src/track/libheaptrack.cpp#L880

https://github.com/KDE/heaptrack/blob/b54e92e88b0895f910a5b107f499d1155a086b80/src/track/libheaptrack.cpp#L522

Apparently, `heaptrack` just spawns the process and monitors it:
```nim
1478955 pts/4    S+     0:00  |   |       \_ /bin/sh /usr/bin/heaptrack ./build/wakunode2 --config-file=cfg.txt
1478968 pts/4    S+     0:09  |   |           \_ /usr/lib/heaptrack/libexec/heaptrack_interpret
1478969 pts/4    Sl+    0:00  |   |           \_ zstd -c
1478970 pts/4    Sl+    0:18  |   |           \_ ./build/wakunode2 --config-file=cfg.txt
```

The Zoltán's suggested approach is super super insightful :D

Considering these comments, do you think we can suggest these changes to `Nim`? ( we haven't explored `heaptrack` + `orc` yet )

NagyZoltanPeter:
> From what I can tell, the way nim is patched here means memory will be allocated _both_ from the nim gc and from heaptrack, effectively doubling the memory usage (at least doubling memory usage).
> 
> When using `orc`, the way to get heaptrack support is to pass `-d:useMalloc` to the allocator which natively calls `malloc` in the right spots - rebasing [nim-lang/Nim#16761](https://github.com/nim-lang/Nim/pull/16761) would be one way to add `-d:useMalloc` support to `--mm:refc` - as noted in the PR, support for 32-bit platforms isn't ready so the way to update that PR is to make it conditional for 64-bit platforms only.

I see, I think such double alloc is true in case using heaptrack_inject... we have not used it yet. Maybe it is worth not to support it for now. We can work with heaptrack instrumentation only.
cc: @Ivansete-status

Dockerfile
```diff
@@ -5,6 +5,7 @@ ARG NIMFLAGS
 ARG MAKE_TARGET=wakunode2
 ARG NIM_COMMIT
 ARG LOG_LEVEL=TRACE
+ARG HEAPTRACK_BUILD=0
```
fryorcraken, line 8:
```suggestion
ARG HEAPTRACK_BUILD=false
```

bit weird to use `0` IMO.

Dockerfile
```diff
@@ -18,6 +19,10 @@ RUN apk update && apk upgrade
 # Ran separately from 'make' to avoid re-doing
 RUN git submodule update --init --recursive
 
+RUN if [ "$HEAPTRACK_BUILD" = "1" ]; then \
```
fryorcraken, line 22:
```suggestion
RUN if [ "$HEAPTRACK_BUILD" ]; then \
```

See https://github.com/waku-org/nwaku/pull/3522/files#r2227311344

Dockerfile
```diff
@@ -18,6 +19,10 @@ RUN apk update && apk upgrade
...
+      git apply --directory=vendor/nimbus-build-system/vendor/Nim docs/tutorial/nim.2.2.4_heaptracker_addon.patch; \
```
fryorcraken, line 23:
```suggestion
      git apply --directory="vendor/nimbus-build-system/vendor/Nim docs/tutorial/nim.2.2.4_heaptracker_addon.patch; "\
```

Is it even working with the spaceS?

docs/tutorial/heaptrack.md
```diff
@@ -40,13 +40,26 @@ It operates in two modes:
 ## Heaptrack & Nwaku
 nwaku supports heaptrack but it needs a special compilation setting.
 
-### Build nwaku with heaptrack support
+### Patch Nim compiler to register allocations on Heaptrack
+
+Currently we are rely on official Nim repository. So we need to patch the Nim compiler to register allocations and deallocations on Heaptrack.
+For Nim 2.2.4 version we created a patch that can be applied as:
+```bash
+git apply --directory=vendor/nimbus-build-system/vendor/Nim docs/tutorial/nim.2.2.4_heaptracker_addon.patch
```
fryorcraken, line 48:
```suggestion
git apply --directory="vendor/nimbus-build-system/vendor/Nim docs/tutorial/nim.2.2.4_heaptracker_addon.patch"
```

docs/tutorial/heaptrack.md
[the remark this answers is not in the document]
fryorcraken: I have the same question as copilot

Dockerfile
Ivansete-status: We use `0` or `1` for compatibility with all possible Makefile settings. This should be kept `0` in that PR and open a separate PR to perform the transition from `0`/ `1` to `false`/`true`, which might have implication in CI files.

NagyZoltanPeter: I agree to stick to existing approach and use 0 and 1 to represent choice.

Dockerfile
NagyZoltanPeter: Let's stick to 0 and 1 values for now.

Dockerfile
NagyZoltanPeter: There is no space in the directories. The space devided two arguments here. First is `--directory=...` defines where to apply the last argument, the patch.

docs/tutorial/heaptrack.md
NagyZoltanPeter: It is good as its written originally, the space divides the two arguments for the apply command.

NagyZoltanPeter:
@fryorcraken catching up with this. Answered your findings and made some clarification in heaptrack.md.
Please re-iterate on reviewing this. Thank you.

fryorcraken (approved):
LGTM

PR #3523, opened 2025-07-23
chore: peer-exchange client separation

chaitanyaprem:
# Description
As of now peer-exchange implementation requires mounting of the peer-exchange protocol even if only the client side functionality is required. This PR address this and separates the implementation for client side and service side of the protocol.

Noticed while working on chat2 app for mix.
But opened the PR on master as this PR doesn't have dependency on any other PR.

### DIFF
--- a/waku/factory/node_factory.nim
+++ b/waku/factory/node_factory.nim
@@ -411,7 +411,7 @@ proc setupProtocols(
       return err("failed to set node waku filter peer: " & filterNode.error)
 
   # waku peer exchange setup
-  if conf.peerExchange:
+  if conf.peerExchangeService:
     try:
       await mountPeerExchange(
         node, some(conf.clusterId), node.rateLimitSettings.getSetting(PEEREXCHG)
@@ -428,6 +428,8 @@ proc setupProtocols(
       return
         err("failed to set node waku peer-exchange peer: " & peerExchangeNode.error)
 
+  if conf.peerExchangeDiscovery:
+    await node.mountPeerExchangeClient()
   return ok()
 
 ## Start node
@@ -472,7 +474,7 @@ proc startNode*(
   # 
   # Use px to periodically get peers if discv5 is disabled, as discv5 nodes have their own
   # periodic loop to find peers and px returned peers actually come from discv5
-  if conf.peerExchange and not conf.discv5Conf.isSome():
+  if conf.peerExchangeDiscovery and not conf.discv5Conf.isSome():
     node.startPeerExchangeLoop()
 
   # Maintain relay connections
--- a/waku/factory/waku_conf.nim
+++ b/waku/factory/waku_conf.nim
@@ -85,7 +85,8 @@ type WakuConf* {.requiresInit.} = ref object
 
   relay*: bool
   lightPush*: bool
-  peerExchange*: bool
+  peerExchangeService*: bool
+  peerExchangeDiscovery*: bool
 
   # TODO: remove relay peer exchange
   relayPeerExchange*: bool
@@ -145,7 +146,7 @@ proc logConf*(conf: WakuConf) =
     store = conf.storeServiceConf.isSome(),
     filter = conf.filterServiceConf.isSome(),
     lightPush = conf.lightPush,
-    peerExchange = conf.peerExchange
+    peerExchange = conf.peerExchangeService
 
   info "Configuration. Network", cluster = conf.clusterId
 
--- a/waku/node/waku_node.nim
+++ b/waku/node/waku_node.nim
@@ -111,6 +111,7 @@ type
     wakuLightPush*: WakuLightPush
     wakuLightpushClient*: WakuLightPushClient
     wakuPeerExchange*: WakuPeerExchange
+    wakuPeerExchangeClient*: WakuPeerExchangeClient
     wakuMetadata*: WakuMetadata
     wakuAutoSharding*: Option[Sharding]
     enr*: enr.Record
@@ -1278,21 +1279,26 @@ proc mountPeerExchange*(
   except LPError:
     error "failed to mount wakuPeerExchange", error = getCurrentExceptionMsg()
 
+proc mountPeerExchangeClient*(node: WakuNode) {.async: (raises: []).} =
+  info "mounting waku peer exchange client"
+  if node.wakuPeerExchangeClient.isNil():
+    node.wakuPeerExchangeClient = WakuPeerExchangeClient.new(node.peerManager)
+
 proc fetchPeerExchangePeers*(
     node: Wakunode, amount = DefaultPXNumPeersReq
 ): Future[Result[int, PeerExchangeResponseStatus]] {.async: (raises: []).} =
-  if node.wakuPeerExchange.isNil():
-    error "could not get peers from px, waku peer-exchange is nil"
+  if node.wakuPeerExchangeClient.isNil():
+    error "could not get peers from px, waku peer-exchange-client is nil"
     return err(
       (
         status_code: PeerExchangeResponseStatusCode.SERVICE_UNAVAILABLE,
-        status_desc: some("PeerExchange is not mounted"),
+        status_desc: some("PeerExchangeClient is not mounted"),
       )
     )
 
   info "Retrieving peer info via peer exchange protocol", amount
-  let pxPeersRes = await node.wakuPeerExchange.request(amount)
-  if pxPeersRes.isOk:
+  let pxPeersRes = await node.wakuPeerExchangeClient.request(amount)
+  if pxPeersRes.isOk():
     var validPeers = 0
     let peers = pxPeersRes.get().peerInfos
     for pi in peers:
@@ -1310,17 +1316,19 @@ proc fetchPeerExchangePeers*(
 
 proc peerExchangeLoop(node: WakuNode) {.async.} =
   while true:
-    await sleepAsync(1.minutes)
     if not node.started:
+      await sleepAsync(5.seconds)
       continue
     (await node.fetchPeerExchangePeers()).isOkOr:
       warn "Cannot fetch peers from peer exchange", cause = error
+    await sleepAsync(1.minutes)
 
 proc startPeerExchangeLoop*(node: WakuNode) =
-  if node.wakuPeerExchange.isNil():
+  if node.wakuPeerExchangeClient.isNil():
     error "startPeerExchangeLoop: Peer Exchange is not mounted"
     return
-  node.wakuPeerExchange.pxLoopHandle = node.peerExchangeLoop()
+  info "Starting peer exchange loop"
+  node.wakuPeerExchangeClient.pxLoopHandle = node.peerExchangeLoop()
 
 # TODO: Move to application module (e.g., wakunode2.nim)
 proc setPeerExchangePeer*(
@@ -1557,6 +1565,10 @@ proc stop*(node: WakuNode) {.async.} =
 
   if not node.wakuPeerExchange.isNil() and not node.wakuPeerExchange.pxLoopHandle.isNil():
     await node.wakuPeerExchange.pxLoopHandle.cancelAndWait()
+
+  if not node.wakuPeerExchangeClient.isNil() and
+      not node.wakuPeerExchangeClient.pxLoopHandle.isNil():
+    await node.wakuPeerExchangeClient.pxLoopHandle.cancelAndWait()
 
   if not node.wakuRendezvous.isNil():
     await node.wakuRendezvous.stopWait()
--- a/waku/waku_peer_exchange/client.nim
+++ b/waku/waku_peer_exchange/client.nim
@@ -1,0 +1,102 @@
+import std/options, results, chronicles, chronos, metrics
+
+import ./common, ./rpc, ./rpc_codec, ../node/peer_manager
+
+from ../waku_core/codecs import WakuPeerExchangeCodec
+
+declarePublicGauge waku_px_peers_received_total,
+  "number of ENRs received via peer exchange"
+declarePublicCounter waku_px_client_errors, "number of peer exchange errors", ["type"]
+
+logScope:
+  topics = "waku peer_exchange client"
+
+type WakuPeerExchangeClient* = ref object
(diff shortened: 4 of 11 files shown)


waku/node/waku_node.nim
```diff
@@ -1344,17 +1350,19 @@ proc fetchPeerExchangePeers*(
 
 proc peerExchangeLoop(node: WakuNode) {.async.} =
   while true:
-    await sleepAsync(1.minutes)
     if not node.started:
+      await sleepAsync(5.seconds)
```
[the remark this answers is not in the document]
chaitanyaprem: This is fine as node starting doesn't take this long and hence loop for sleeping can be shorter before it starts.

waku/node/waku_node.nim
```diff
@@ -1278,21 +1279,26 @@ proc mountPeerExchange*(
   except LPError:
     error "failed to mount wakuPeerExchange", error = getCurrentExceptionMsg()
 
+proc mountPeerExchangeClient*(node: WakuNode) {.async: (raises: []).} =
```
SionoiS, line 1282: Is this called from somewhere?

chaitanyaprem: well, it would be called once light/edge mode is implemented in nwaku. As of now it will be used by apps that want to act as light/edge nodes e.g chat2mix.

waku/node/waku_node.nim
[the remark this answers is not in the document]
fryorcraken: can we just start the peer exchange loop once peer exchange is mounted?

waku/node/waku_node.nim
fryorcraken: I disagree here as both relay and edge nodes should be using peer exchange client.

I would expect it to be mounted based on a configuration property on `WakuConf`.

chaitanyaprem: Why so? Relay nodes act as peer-exchange service-node as they already mount discv5 for discovery. 

I do agree that it would be a good idea to mount it based on a config so that it gives flexibility for users to not choose discv5, but is that something we want to give as an option ?

fryorcraken: Because using peer exchange allows you to bootstrap faster after start, by connecting to peers discovery via peer exchange. As done in Status Desktop.

Hence, in [`relay`](https://github.com/waku-org/specs/blob/4ca0e08038f7cc2df664ec9e06a3b32418d63ab7/standards/application/waku-api.md?plain=1#L103) mode, I would expect both peer exchange client, server and discv5 to be mounted.

No, I would not give the flexibility to users to not choose discv5, but I would give them the flexibility to use `edge` or `relay`, where `edge` does not use discv5.

I am not providing an opinion on whether this should be available via CLI (`WakuNodeConf`), but I am saying this should be part of the internal Waku config (`WakuConf`) where peer exchange client is mounted based on the `WakuConf`.

I would also say that probably,`mountPeerExchangeClient` should not be exposed, but solely mounted based on `WakuConf` values.

jm-clius (approved):
Approving, as I agree with this separation.

waku/factory/node_factory.nim
```diff
@@ -421,6 +421,13 @@ proc setupProtocols(
       return
         err("failed to set node waku peer-exchange peer: " & peerExchangeNode.error)
 
+  if conf.peerExchangeClient:
+    try:
+      await node.mountPeerExchangeClient()
+    except CatchableError:
+      # Not failing here as this only speeds up bootstrapping and doesnt affect any other functionality
+      error "failed to mount waku peer-exchange client protocol: ",
+        error = getCurrentExceptionMsg()
```
Ivansete-status: This exception will never happen as this proc is annotated with `(raises: [])`
```suggestion
    await node.mountPeerExchangeClient()
```

waku/waku_peer_exchange/client.nim
```diff
@@ -0,0 +1,105 @@
...
+
+  var buffer: seq[byte]
+  var callResult =
+    (status_code: PeerExchangeResponseStatusCode.SUCCESS, status_desc: none(string))
+  try:
+    await conn.writeLP(rpc.encode().buffer)
+    buffer = await conn.readLp(DefaultMaxRpcSize.int)
+  except CatchableError as exc:
+    error "exception when handling peer exchange request",
+      error = getCurrentExceptionMsg()
```
Ivansete-status:
```suggestion
      error = exc.msg
```

waku/waku_peer_exchange/client.nim
```diff
@@ -0,0 +1,105 @@
...
+    waku_px_client_errors.inc(labelValues = [exc.msg])
```
Ivansete-status: When dealing with `labelValues` it is very important to use a limited number of possible values, i.e., we cannot add arbitrary values because that would damage Prometheus performance, etc.
```suggestion
    waku_px_client_errors.inc(labelValues = ["error_sending_or_receiving_px_req"])
```

waku/waku_peer_exchange/client.nim
```diff
@@ -0,0 +1,105 @@
...
+): Future[WakuPeerExchangeResult[PeerExchangeResponse]] {.async: (raises: []).} =
+  let rpc = PeerExchangeRpc.makeRequest(numPeers)
+
+  var buffer: seq[byte]
+  var callResult =
+    (status_code: PeerExchangeResponseStatusCode.SUCCESS, status_desc: none(string))
+  try:
+    await conn.writeLP(rpc.encode().buffer)
+    buffer = await conn.readLp(DefaultMaxRpcSize.int)
+  except CatchableError as exc:
```
Ivansete-status, line 32: I think is better to split the try-catch in two so that we have more insight about whether the write or read failed

waku/waku_peer_exchange/client.nim
```diff
@@ -0,0 +1,105 @@
...
+    error "exception when handling peer exchange request",
+      error = getCurrentExceptionMsg()
+    waku_px_client_errors.inc(labelValues = [exc.msg])
+    callResult = (
+      status_code: PeerExchangeResponseStatusCode.SERVICE_UNAVAILABLE,
+      status_desc: some($exc.msg),
+    )
+  finally:
```
Ivansete-status, line 39: This is correct indeed, but if we split the exc catching in two, as suggested above, we can also use the following at the top of the proc implementation:
```nim
defer:
  await conn.closeWithEof()
```

waku/waku_peer_exchange/client.nim
```diff
@@ -0,0 +1,105 @@
...
+    error "peer exchange request error",
+      status_code = decodedBuff.get().response.status_code
+    return err(
+      (
+        status_code: decodedBuff.get().response.status_code,
+        status_desc: decodedBuff.get().response.status_desc,
+      )
+    )
+
+  return ok(decodedBuff.get().response)
```
Ivansete-status: Good opportunity for using the `valueOr` technique
```suggestion
  let decoded = PeerExchangeRpc.decode(buffer).valueOr:
    error "peer exchange request error decoding buffer", error = $error
    return err(
      (
        status_code: PeerExchangeResponseStatusCode.BAD_RESPONSE,
        status_desc: some($error),
      )
    )

  if decodedBuff.response.status_code != PeerExchangeResponseStatusCode.SUCCESS:
    error "peer exchange request error",
      status_code = decodedBuff.response.status_code
    return err(
      (
        status_code: decodedBuff.response.status_code,
        status_desc: decodedBuff.response.status_desc,
      )
    )

  return ok(decodedBuff.response)
```

waku/waku_peer_exchange/client.nim
```diff
@@ -0,0 +1,105 @@
...
+        status_desc: some("exception dialing peer: " & getCurrentExceptionMsg()),
+      )
+    )
+
+proc request*(
+    wpx: WakuPeerExchangeClient, numPeers = DefaultPXNumPeersReq
+): Future[WakuPeerExchangeResult[PeerExchangeResponse]] {.async: (raises: []).} =
+  let peerOpt = wpx.peerManager.selectPeer(WakuPeerExchangeCodec)
+  if peerOpt.isNone():
+    waku_px_client_errors.inc(labelValues = [peerNotFoundFailure])
```
Ivansete-status, line 94: Yes, this is the correct place to add an error label :partying_face:

chaitanyaprem:
@fryorcraken , do  re-review this

Ivansete-status (approved):
LGTM! Thanks for it! :100: 
Some side comments that I hope you find useful
Cheers

waku/waku_peer_exchange/client.nim
chaitanyaprem: actually this was suggested by AI and i did not review :P

waku/waku_peer_exchange/client.nim
chaitanyaprem: wouldn't the underlying error message indicate that anyways? 

splitting it just complicates the code to look at.

waku/factory/waku_conf.nim
```diff
@@ -86,6 +86,7 @@ type WakuConf* {.requiresInit.} = ref object
   relay*: bool
   lightPush*: bool
   peerExchange*: bool
+  peerExchangeClient*: bool
```
fryorcraken:
```suggestion
peerExchangeService*: bool
  peerExchange*: bool
```

Look few line below, we started to adopt the convention of marking the *service* one, I'd recommend to keep the same to make it easier to work on the codebase.

fryorcraken: Actually, I am not sure `peerExchangeClient` is a good name. Because it implies we are enabling peer exchange as client. But it's more than that: we are enabling peer exchange discovery and setting up automated routine to do it.

Hence, something like `peerExchangeDiscovery` may be better.

waku/node/waku_node.nim
```diff
@@ -1278,21 +1279,26 @@ proc mountPeerExchange*(
...
+  info "mounting waku peer exchange client"
+  if node.wakuPeerExchangeClient.isNil():
+    node.wakuPeerExchangeClient = WakuPeerExchangeClient.new(node.peerManager)
+
 proc fetchPeerExchangePeers*(
```
fryorcraken, line 1287: @Ivansete-status it's interesting how `waku_node.nim` is basically an SDK layer that brings usage of various protocols.

It may be wise to start thinking about splitting it, because the file is growing fast in size.

For example, the peer exchange logic could probably be in its own file (under `factory` is fine), and just brought in here if used.

waku/node/waku_node.nim
```diff
@@ -1278,21 +1279,26 @@ proc mountPeerExchange*(
...
     node: Wakunode, amount = DefaultPXNumPeersReq
```
fryorcraken, line 1288: Note comment for this PR, but more in general @waku-org/nwaku , there is too much of a tendency to pass big objects around in API, instead of passing only what is needed. Here is `wakuPeerExchangeClient` and `peerManager`.

This is problematic for several reasons:

1. We don't leverage type guarding as much as we could, we hence often need to start with some `isSome` or `!isNil` checks as you can see below.
2. we bloat those big objects such as `Wakunode`, and just flat add to it, instead of reviewing the data structure and possibility adopt a more nested approach (eg those 3 properties are always used together, let's clump then in one object)
3. We make refactoring and functional isolation harder: if someone want to extract the peer exchange loop logic from `waku_node.nim`, they would need to depend on `waku_node.nim` to import `WakuNode` object type. which is likely to lead to circular dependencies.

waku/waku_peer_exchange/client.nim
```diff
@@ -0,0 +1,102 @@
...
+    error "peer exchange request exception", error = getCurrentExceptionMsg()
+    return err(
+      (
+        status_code: PeerExchangeResponseStatusCode.BAD_RESPONSE,
+        status_desc: some("exception dialing peer: " & getCurrentExceptionMsg()),
+      )
+    )
+
+proc request*(
+    wpx: WakuPeerExchangeClient, numPeers = DefaultPXNumPeersReq
```
fryorcraken, line 90: I find the js-waku logic better here. Keen to hear your thoughts @jm-clius @weboko 

**Proposed nwaku logic**:
every 5 minute
  grab **one** random peer-exchange peer
  dial it
  do a request

**Current js-waku logic**:
when *discoverying* a new peer
   if it has peer-exchange advertised
	start querying it every 10s
	abort after 3 errors

I think the likelihood you'll get a successful p-x query with the proposed logic is too low, and I'd encourage to consider a  logic closer to js-waku's

waku/waku_peer_exchange/client.nim
Ivansete-status: Ideally the exceptions should have the smallest scope. Is just a matter of giving better feedback and avoid ambiguity. Not a huge deal though :D

waku/node/waku_node.nim
Ivansete-status: > It may be wise to start thinking about splitting it, because the file is growing fast in size.

Yes I agree 100% with that. Also was thinking about that the other day

waku/waku_peer_exchange/client.nim
chaitanyaprem: nwaku currently does px request every `1 minute` by selecting a peer at random from peerStore. This should be fine for relay nodes. 

iirc peerExchange is meant for 2 reasons
- used for quicker bootstrapping and peer discovery than discv5
- for light/edge nodes as discv5 is too much bandwidth intensive

for edge nodes as well maybe initially queries can be faster and later be left to 1 min. 
One more thought that comes to me is that nodes can always query on demand if they don't have any peers matching criteria for services they want to use. 

Not sure if it is a good idea to query so frequently(every 10 secs) all through lifetime of the node. 
This maybe useful for mix protocol as the node pool has to be built up fast but for others i don't see any benefit.

chaitanyaprem:
@fryorcraken if there are no further changes required, please indicate the same so that this can be merged.

CI seems to be failing for some or other reason even if i rerun. Issues are not related to this PR though. Seems like flaky CI

fryorcraken (approved):
good stuff

PR #3525, opened 2025-07-24
chore: guarding against double starting and stopping of nodes

gabrielmer:
# Description
Adding guards against starting/stopping nodes twice

# Changes

- [x] guarding against double starting/stopping of nodes

## Issue

#3483

### DIFF
--- a/waku/factory/waku.nim
+++ b/waku/factory/waku.nim
@@ -362,6 +362,10 @@ proc startDnsDiscoveryRetryLoop(waku: ptr Waku): Future[void] {.async.} =
     return
 
 proc startWaku*(waku: ptr Waku): Future[Result[void, string]] {.async.} =
+  if waku[].node.started:
+    warn "startWaku: waku node already started"
+    return ok()
+
   debug "Retrieve dynamic bootstrap nodes"
   let conf = waku[].conf
 
@@ -441,6 +445,9 @@ proc startWaku*(waku: ptr Waku): Future[Result[void, string]] {.async.} =
 
 proc stop*(waku: Waku): Future[void] {.async: (raises: [Exception]).} =
   ## Waku shutdown
+
+  if not waku.node.started:
+    warn "stop: attempting to stop node that isn't running"
 
   waku.healthMonitor.setOverallHealth(HealthStatus.SHUTTING_DOWN)
 
--- a/waku/factory/node_factory.nim
+++ b/waku/factory/node_factory.nim
@@ -432,10 +432,6 @@ proc startNode*(
   ## Connect to static nodes and start
   ## keep-alive, if configured.
 
-  if node.started:
-    warn "startNode: attempting to start node that is already running"
-    return ok()
-
   info "Running nwaku node", version = git_version
   try:
     await node.start()
--- a/waku/factory/node_factory.nim
+++ b/waku/factory/node_factory.nim
@@ -432,6 +432,10 @@ proc startNode*(
   ## Connect to static nodes and start
   ## keep-alive, if configured.
 
+  if node.started:
+    warn "startNode: attempting to start node that is already running"
+    return ok()
+
   info "Running nwaku node", version = git_version
   try:
     await node.start()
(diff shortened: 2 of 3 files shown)


waku/factory/waku.nim
```diff
@@ -442,6 +446,9 @@ proc startWaku*(waku: ptr Waku): Future[Result[void, string]] {.async.} =
 proc stop*(waku: Waku): Future[void] {.async: (raises: [Exception]).} =
   ## Waku shutdown
 
+  if not waku.node.started:
```
gabrielmer, line 449: here I use `waku.node` as a reference, assuming that if the node isn't running then anything else can be running. We could instead add a `waku.started` field if we don't want to use the node as a proxy, but didn't think it was worth maintaining another field

waku/factory/waku.nim
```diff
@@ -362,6 +362,10 @@ proc startDnsDiscoveryRetryLoop(waku: ptr Waku): Future[void] {.async.} =
     return
 
 proc startWaku*(waku: ptr Waku): Future[Result[void, string]] {.async.} =
+  if waku[].node.started:
+    warn "startWaku: waku node already started"
+    return ok()
```
Ivansete-status, line 367: I think this is the only place where it is needed

Ivansete-status (approved):
Thanks for it! :raised_hands: 
I think the PR can be simpler and not needed to add in so many places but I might be missing something

waku/factory/waku.nim
NagyZoltanPeter: Wonder if this is worth mentioning with a warning. We just need to ensure nothing harmful - aka crash - could happen. Thinking of libwaku, stopping a node might be intentional.

waku/factory/waku.nim
```diff
@@ -362,6 +362,10 @@ proc startDnsDiscoveryRetryLoop(waku: ptr Waku): Future[void] {.async.} =
     return
 
 proc startWaku*(waku: ptr Waku): Future[Result[void, string]] {.async.} =
+  if waku[].node.started:
```
NagyZoltanPeter, line 365: Is it possible to call `startWaku` - as its being async - twice while it is starting up?
How to ensure it should not happen? As `started` just record a successful start up... but not the period of starting.

NagyZoltanPeter (approved):
Good point to tackle it, but unsure if it is sufficient.

waku/factory/waku.nim
gabrielmer: mmm good question, I think that in a single-threaded application, calling `start` without awaiting and trying to do operations when the node hasn't started yet is a bad idea and if it fails is a good thing that exposes that poor pattern.

On the other hand, if it's a multi-threaded application with many threads interacting with one node, this guard can be useless although I wouldn't bother on that case unless we see the need.

We could change the function to mark the node as started as soon as `startWaku` is called, and same with `stop`. I have mixed feelings about it, I'm probably more of the philosophy of changing things whenever we see the issue happening.

@Ivansete-status what's your take?

waku/factory/waku.nim
gabrielmer: Not sure I understood 😶 

some other warning beyond the `warn "stop: attempting to stop node that isn't running"`?

the good thing is that libwaku calls this function to stop the node, so it can't happen that the node gets stopped from libwaku and have other services up unless it is in some invalid/error state

waku/factory/waku.nim
Ivansete-status: Good point!
I think it works well as it is currently being suggested because we do `await startWaku`
Another question is whether we should follow that idempotent approach or, instead, return an error if the node is already running. I lean a bit more towards the latter.

gabrielmer: Good point, I chose the idempotent approach to break as little as possible of the caller application if it's not necessary and nothing is failing underneath

NagyZoltanPeter: Yes, I think an interface must be usable with less friction and pre-assumption or pre-requisite knowledge.
So as soon as it can do its job correctly, it shall not return an error.

PR #3526, opened 2025-07-24
fix: only stop health monitor components if not nil

gabrielmer:
# Description
Only stopping health monitor components that aren't nil

## Issue

#3483


NagyZoltanPeter (approved):
LGTM

PR #3529, opened 2025-07-30
chore: Update RLN contract address to new one with restricted token minting

stubbsta:
# Description
The WakuRlnV2 contract has been redeployed and the address needs to be updated.
The redeployment is required because the current contract has been spammed and there are no more memberships left.

# Changes

All instances of the old contract address have been replaced with the new one.

## How to test

1. Test with this PR image and nwaku-compose branch: https://github.com/waku-org/nwaku-compose/tree/update-to-contract-with-restricted-mint


fryorcraken (changes requested):
I don't agree with this: why the new smart contract and token? because the previous is full? Is there any functional changes to the new instances?

- we can always increase the global rate limit on currently deployed contract
- we [should review](https://github.com/waku-org/pm/pull/331/commits/d257fbe8e5c687d328b9a2e844945ef58d869fed) our numbers to allow more registrations
- some people (me) have upgraded and deployed node with new smart contract, I don't see a point to make this effort wasted.

PR #3530, opened 2025-08-01
fix: apply modulus in sharding

Ivansete-status:
## Description
It seems that the suggested operation is not equivalent for all cases.

For example, the following don't give the same results:
```nim
  echo "25 mod 7 == " & $(25.uint64 mod 7.uint64)
  echo "25 and (7 - 1) == " & $(25.uint64 and uint64(7 - 1))
```

## Issue


darshankabariya (approved):
Amazing catch, Thanks @Ivansete-status. LGTM

NagyZoltanPeter (approved):
Amazing!
Actually the original only works for count that is power of two.
Btw, it is never a good idea to optimize a code this way. Trust the compiler it knows better.
If you check with compiler explorer the C code equivalent compiles exactly according to what compiler can predict. https://godbolt.org/z/f7MaKeKzG
So such a solution can only work if you are sure the count is always power of two.

PR #3536, opened 2025-08-08
chore: bump dependencies to v0.37.0

darshankabariya:


fryorcraken:
I wish I could help but I have little to offer to this PR :]

waku/node/waku_switch.nim
```diff
@@ -53,7 +53,8 @@ proc withWssTransport*(
         upgr,
         tlsPrivateKey = key,
         tlsCertificate = cert,
-        {TLSFlags.NoVerifyHost, TLSFlags.NoVerifyServerName},
+        autotls = nil, # required 5th param
```
Ivansete-status, line 56: @richard-ramos - should we better set a different value for this parameter?

Ivansete-status (approved):
LGTM! Thanks so much for it! :100: :partying_face:

waku/node/waku_switch.nim
richard-ramos: The whole proc can be replaced with this:
```nim
proc withWssTransport*(
    b: SwitchBuilder, secureKeyPath: string, secureCertPath: string
): SwitchBuilder {.raises: [Defect, IOError].} =
  let key: TLSPrivateKey = getSecureKey(secureKeyPath)
  let cert: TLSCertificate = getSecureCert(secureCertPath)
  b.withWsTransport(
    tlsPrivateKey = key,
    tlsCertificate = cert,
    {TLSFlags.NoVerifyHost, TLSFlags.NoVerifyServerName},
  )
```

However, curious to know the reason behind using those TLSFlags.

fryorcraken:
> Should be safe to use latest `master`, but dunno the policy of nwaku regarding pointing to non-release commits 

I would recommend to not point to latest `master` for any vendor, but instead to a tag so it matches `nimble` 's behaviour.

@richard-ramos if you have good fixes, why not doing a release?

@darshankabariya do you update to `master` or to `tag` for other dependencies?

NagyZoltanPeter (approved):
LGTM! Great work, thank you.

PR #3538, opened 2025-08-13
chore: Lightpush minor refactor

s-tikhomirov:
# Description

In the spirit of "make change easy", this PR is a minor refactor of Lightpush that would hopefully make the protocol easier to reason about.

# Changes

- [x] extract helper functions to make code more DRY
- [x] add a few comments

### DIFF
--- a/waku/waku_lightpush/client.nim
+++ b/waku/waku_lightpush/client.nim
@@ -17,8 +17,8 @@ logScope:
   topics = "waku lightpush client"
 
 type WakuLightPushClient* = ref object
-  peerManager*: PeerManager
   rng*: ref rand.HmacDrbgContext
+  peerManager*: PeerManager
   publishObservers: seq[PublishObserver]
 
 proc new*(
@@ -28,6 +28,17 @@ proc new*(
 
 proc addPublishObserver*(wl: WakuLightPushClient, obs: PublishObserver) =
   wl.publishObservers.add(obs)
+
+proc ensureTimestampSet(message: var WakuMessage) =
+  if message.timestamp == 0:
+    message.timestamp = getNowInNanosecondTime()
+
+## Short log string for peer identifiers (overloads for convenience)
+func shortPeerId(peer: PeerId): string =
+  shortLog(peer)
+
+func shortPeerId(peer: RemotePeerInfo): string =
+  shortLog(peer.peerId)
 
 proc sendPushRequest(
     wl: WakuLightPushClient, req: LightPushRequest, peer: PeerId | RemotePeerInfo
@@ -39,7 +50,13 @@ proc sendPushRequest(
       dialFailure & ": " & $peer & " is not accessible",
     )
 
-  await connection.writeLP(req.encode().buffer)
+  try:
+    await connection.writeLp(req.encode().buffer)
+  except LPStreamRemoteClosedError:
+    error "Failed to write request to peer", error = getCurrentExceptionMsg()
+    return lightpushResultInternalError(
+      "Failed to write request to peer: " & getCurrentExceptionMsg()
+    )
 
   var buffer: seq[byte]
   try:
@@ -55,8 +72,10 @@ proc sendPushRequest(
     waku_lightpush_v3_errors.inc(labelValues = [decodeRpcFailure])
     return lightpushResultInternalError(decodeRpcFailure)
 
-  if response.requestId != req.requestId and
-      response.statusCode != LightPushErrorCode.TOO_MANY_REQUESTS:
+  let isTooManyRequests = response.statusCode == LightPushErrorCode.TOO_MANY_REQUESTS
+  let requestIdMatches = response.requestId == req.requestId
+  if (not requestIdMatches) and (not isTooManyRequests):
+    # response with TOO_MANY_REQUESTS error code does not have requestId by design
     error "response failure, requestId mismatch",
       requestId = req.requestId, responseRequestId = response.requestId
     return lightpushResultInternalError("response failure, requestId mismatch")
@@ -65,42 +84,33 @@ proc sendPushRequest(
 
 proc publish*(
     wl: WakuLightPushClient,
-    pubSubTopic: Option[PubsubTopic] = none(PubsubTopic),
+    pubsubTopic: Option[PubsubTopic] = none(PubsubTopic),
     wakuMessage: WakuMessage,
     peer: PeerId | RemotePeerInfo,
 ): Future[WakuLightPushResult] {.async, gcsafe.} =
   var message = wakuMessage
-  if message.timestamp == 0:
-    message.timestamp = getNowInNanosecondTime()
+  ensureTimestampSet(message)
 
-  when peer is PeerId:
-    info "publish",
-      peerId = shortLog(peer),
-      msg_hash = computeMessageHash(pubsubTopic.get(""), message).to0xHex
-  else:
-    info "publish",
-      peerId = shortLog(peer.peerId),
-      msg_hash = computeMessageHash(pubsubTopic.get(""), message).to0xHex
+  let msgHash = computeMessageHash(pubsubTopic.get(""), message).to0xHex
+  info "publish", peerId = shortPeerId(peer), msg_hash = msgHash
 
   let pushRequest = LightpushRequest(
-    requestId: generateRequestId(wl.rng), pubSubTopic: pubSubTopic, message: message
+    requestId: generateRequestId(wl.rng), pubsubTopic: pubsubTopic, message: message
   )
-  let publishedCount = ?await wl.sendPushRequest(pushRequest, peer)
+  let relayPeerCount = ?await wl.sendPushRequest(pushRequest, peer)
 
   for obs in wl.publishObservers:
-    obs.onMessagePublished(pubSubTopic.get(""), message)
+    obs.onMessagePublished(pubsubTopic.get(""), message)
 
-  return lightpushSuccessResult(publishedCount)
+  return lightpushSuccessResult(relayPeerCount)
 
 proc publishToAny*(
-    wl: WakuLightPushClient, pubSubTopic: PubsubTopic, wakuMessage: WakuMessage
+    wl: WakuLightPushClient, pubsubTopic: PubsubTopic, wakuMessage: WakuMessage
 ): Future[WakuLightPushResult] {.async, gcsafe.} =
-  ## This proc is similar to the publish one but in this case
-  ## we don't specify a particular peer and instead we get it from peer manager
+  # Like publish, but selects a peer automatically from the peer manager
 
   var message = wakuMessage
-  if message.timestamp == 0:
-    message.timestamp = getNowInNanosecondTime()
+  ensureTimestampSet(message)
 
   info "publishToAny", msg_hash = computeMessageHash(pubsubTopic, message).to0xHex
 
@@ -110,14 +120,4 @@ proc publishToAny*(
       LightPushErrorCode.NO_PEERS_TO_RELAY, "no suitable remote peers"
     )
 
-  let pushRequest = LightpushRequest(
-    requestId: generateRequestId(wl.rng),
-    pubSubTopic: some(pubSubTopic),
-    message: message,
-  )
-  let publishedCount = ?await wl.sendPushRequest(pushRequest, peer)
-
-  for obs in wl.publishObservers:
-    obs.onMessagePublished(pubSubTopic, message)
-
-  return lightpushSuccessResult(publishedCount)
+  return await wl.publish(some(pubsubTopic), message, peer)
--- a/waku/waku_lightpush/common.nim
+++ b/waku/waku_lightpush/common.nim
@@ -22,7 +22,8 @@ const LightPushErrorCode* = (
 )
 
 type ErrorStatus* = tuple[code: LightpushStatusCode, desc: Option[string]]
-type WakuLightPushResult* = Result[uint32, ErrorStatus]
+type WakuLightPushResult* = Result[uint32, ErrorStatus] # Ok result is relayPeerCount
+type WakuLightPushTopicResult* = Result[PubsubTopic, ErrorStatus]
 
 type PushMessageHandler* = proc(
   peer: PeerId, pubsubTopic: PubsubTopic, message: WakuMessage
--- a/waku/waku_lightpush/common.nim
+++ b/waku/waku_lightpush/common.nim
@@ -23,7 +23,6 @@ const LightPushErrorCode* = (
 
 type ErrorStatus* = tuple[code: LightpushStatusCode, desc: Option[string]]
 type WakuLightPushResult* = Result[uint32, ErrorStatus] # Ok result is relayPeerCount
-type WakuLightPushTopicResult* = Result[PubsubTopic, ErrorStatus]
 
 type PushMessageHandler* = proc(
   peer: PeerId, pubsubTopic: PubsubTopic, message: WakuMessage
@@ -36,6 +35,7 @@ func isSuccess*(response: LightPushResponse): bool =
 
 func toPushResult*(response: LightPushResponse): WakuLightPushResult =
   if isSuccess(response):
+    # TODO: check relayPeerCount > 0? Publish to 0 peers considered a success?
     return ok(response.relayPeerCount.get(0))
   else:
     return err((response.statusCode, response.statusDesc))
@@ -52,6 +52,7 @@ func lightpushResultBadRequest*(msg: string): WakuLightPushResult =
 func lightpushResultServiceUnavailable*(msg: string): WakuLightPushResult =
   return err((LightPushErrorCode.SERVICE_NOT_AVAILABLE, some(msg)))
 
+# TODO: use a union type for desc?
 func lighpushErrorResult*(
     statusCode: LightpushStatusCode, desc: Option[string]
 ): WakuLightPushResult =
--- a/waku/waku_lightpush/protocol.nim
+++ b/waku/waku_lightpush/protocol.nim
@@ -28,41 +28,45 @@ type WakuLightPush* = ref object of LPProtocol
   requestRateLimiter*: RequestRateLimiter
   autoSharding: Option[Sharding]
 
+proc extractPubsubTopic(
+    wl: WakuLightPush, req: LightpushRequest
+): Result[PubsubTopic, ErrorStatus] =
+  proc mkErr(code: LightPushStatusCode, err: string): Result[PubsubTopic, ErrorStatus] =
+    error "Lightpush request handling error: ", error = err
+    err((code, some(err)))
+
+  let pubsubTopic = req.pubSubTopic.valueOr:
+    if wl.autoSharding.isNone():
+      # Static sharding is enabled, but no pubsub topic is provided
+      return mkErr(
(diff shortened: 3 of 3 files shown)


waku/waku_lightpush/common.nim
```diff
@@ -51,6 +52,7 @@ func lightpushResultBadRequest*(msg: string): WakuLightPushResult =
 func lightpushResultServiceUnavailable*(msg: string): WakuLightPushResult =
   return err((LightPushErrorCode.SERVICE_NOT_AVAILABLE, some(msg)))
 
+# TODO: use a union type for desc?
```
Ivansete-status: What do you mean?

waku/waku_lightpush/client.nim
```diff
@@ -17,8 +17,8 @@ logScope:
   topics = "waku lightpush client"
 
 type WakuLightPushClient* = ref object
-  peerManager*: PeerManager
   rng*: ref rand.HmacDrbgContext
+  peerManager*: PeerManager
```
Ivansete-status, line 21: May undo this movement to avoid possible confusion?

waku/waku_lightpush/protocol.nim
```diff
@@ -28,41 +28,45 @@ type WakuLightPush* = ref object of LPProtocol
...
   if pubsubTopic.isEmptyOrWhitespace():
-    let msg = "topic must not be empty"
-    error "lightpush request handling error", error = msg
-    return
-      WakuLightPushResult.err((code: LightPushErrorCode.BAD_REQUEST, desc: some(msg)))
+    return mkErr(
+      LightPushErrorCode.BAD_REQUEST, "Pubsub topic must not be empty or whitespace"
+    )
+
+  ok(pubsubTopic)
```
Ivansete-status: Nitpick one :)
```suggestion
  return ok(pubsubTopic)
```

waku/waku_lightpush/protocol.nim
```diff
@@ -104,23 +108,32 @@ proc handleRequest*(
...
+    buffer = await conn.readLp(DefaultMaxRpcSize)
+  except LPStreamError:
+    error "lightpush read stream failed", error = getCurrentExceptionMsg()
+    return none(seq[byte])
+
+  waku_service_network_bytes.inc(
+    amount = buffer.len().int64, labelValues = [WakuLightPushCodec, "in"]
+  )
+
+  return some(buffer)
```
Ivansete-status: I think I prefer the previous implementation in that particular snippet :)
I can't see the simplification

waku/waku_lightpush/client.nim
```diff
@@ -65,42 +84,33 @@ proc sendPushRequest(
...
-      msg_hash = computeMessageHash(pubsubTopic.get(""), message).to0xHex
+  let msgHash = computeMessageHash(pubsubTopic.get(""), message).to0xHex
+  info "publish", peerId = shortPeerId(peer), msg_hash = msgHash
 
   let pushRequest = LightpushRequest(
-    requestId: generateRequestId(wl.rng), pubSubTopic: pubSubTopic, message: message
+    requestId: generateRequestId(wl.rng), pubsubTopic: pubsubTopic, message: message
   )
-  let publishedCount = ?await wl.sendPushRequest(pushRequest, peer)
+  let relayPeerCount = ?await wl.sendPushRequest(pushRequest, peer)
```
NagyZoltanPeter: IDK, bit misleading for me, due we use the term relay peers in different meaning, while publish actually returns the number of peers it tries to relay the message, but it comes from different factors (like if using floodpublish).

waku/waku_lightpush/client.nim
```diff
@@ -65,42 +84,33 @@ proc sendPushRequest(
...
-  ## This proc is similar to the publish one but in this case
-  ## we don't specify a particular peer and instead we get it from peer manager
+  # Like publish, but selects a peer automatically from the peer manager
 
   var message = wakuMessage
-  if message.timestamp == 0:
-    message.timestamp = getNowInNanosecondTime()
+  ensureTimestampSet(message)
 
   info "publishToAny", msg_hash = computeMessageHash(pubsubTopic, message).to0xHex
```
NagyZoltanPeter:
```suggestion
  let msgHash = computeMessageHash(pubsubTopic, message).to0xHex
  info "publishToAny", msg_hash = msgHash
```
Its worth to separate, as log macros can lead to double evaluation of arguments, hence it is a function it will be called twice.

waku/waku_lightpush/common.nim
```diff
@@ -35,6 +35,7 @@ func isSuccess*(response: LightPushResponse): bool =
 
 func toPushResult*(response: LightPushResponse): WakuLightPushResult =
   if isSuccess(response):
+    # TODO: check relayPeerCount > 0? Publish to 0 peers considered a success?
```
NagyZoltanPeter: Interesting, but I think it shall not happen as service will answer with failed response and NO_PEERS_TO_RELAY.
Formerly, with legacy lightpush we had such scenario when even 0 peers meant success.

waku/waku_lightpush/protocol.nim
NagyZoltanPeter: My problem with this approach that due to async and option it will make a deep copy of buffer seq[byte] unecessarily.

NagyZoltanPeter (approved):
Looks good to me, thanks for the effort! Added some considerations though.

waku/waku_lightpush/common.nim
s-tikhomirov: We have two nearly identical `lighpushErrorResult` functions here that only differ in the type of the second argument `desc`: `Option[string]` and `string`. Would it make sense to merge them into one function like this? Or would it only complicate things?

```
func lighpushErrorResult*(
    statusCode: LightpushStatusCode, desc: string | Option[string]
): WakuLightPushResult =
  when desc is string:
    return err((statusCode, some(desc)))
  else:
    return err((statusCode, desc))
```

waku/waku_lightpush/client.nim
s-tikhomirov: My reasoning was to unify the field order between `WakuLightPush` and `WakuLightPushClient`. In `WakuLightPush` we have:

```
type WakuLightPush* = ref object of LPProtocol
  rng*: ref rand.HmacDrbgContext
  peerManager*: PeerManager
  autoSharding: Option[Sharding]
```

so I thought is would make sense for `WakuLightPushClient` to also have `rng` first, then `peerManager`. What are the drawbacks here?

waku/waku_lightpush/client.nim
s-tikhomirov: My reasoning was to unify terminology here with other parts of Lightpush codebase. From what I can see, the number of peers to which a message has been published following a Lightpush request is called `relayPeerCount` throughout the protocol, starting from the type definition:

https://github.com/waku-org/nwaku/blob/8c73de095b3ca33de1bfd7d1f75f47929e0b3a67/waku/waku_lightpush/rpc.nim#L20

I agree that naming may be confusing (it was to me at first). In this change, I aimed at unifying it first, and then we can think about renaming everywhere. WDYT?

As a sidenote: `publishedCount` may also be confusing as it's unclear whether it means "published to this many peers" or "published this many times" (can a message be published to the same peer more than once? IDK, probably not, but from `publishedCount` naming it's not obvious.)

waku/waku_lightpush/client.nim
s-tikhomirov: Well, TIL! I asked AI to explain exactly why this is happening:

> Chronicles generates code for all compile-time configured sinks (both text and JSON in Waku's case), and each sink gets its own copy of your `computeMessageHash(pubsubTopic, message).to0xHex` expression in the generated code. Even though only one sink is active at runtime (determined by which output writer is set), both sinks still execute their property-setting code, causing the expensive hash calculation to run twice. This happens because chronicles prioritizes runtime format flexibility over performance - it can't optimize away "inactive" sinks since their activity is only determined at runtime, not compile-time.

Is this true?

waku/waku_lightpush/common.nim
s-tikhomirov: If we assume `relayPeerCount > 0` here, would it make sense to return `err` if `relayPeerCount == 0`? Semantically "relayed to zero peers" sounds like an error case to me and not a success case.

waku/waku_lightpush/protocol.nim
s-tikhomirov: Thanks, I think I understand the issue now. My initial rationale was to avoid what seemed like too deep nesting of `initProtocolHandler`, but it's not worth extra byte array copying of course. Will revert this part.

waku/waku_lightpush/client.nim
Ivansete-status: ah is fine! I thought it was an accidental change :)

waku/waku_lightpush/common.nim
Ivansete-status: ah good catch! It seems we can remove that one:
```nim
func lighpushErrorResult*(
    statusCode: LightpushStatusCode, desc: Option[string]
): WakuLightPushResult =
  return err((statusCode, desc))
```

darshankabariya (approved):
LGTM

waku/waku_lightpush/common.nim
s-tikhomirov: Removed [349ab9e](https://github.com/waku-org/nwaku/pull/3538/commits/349ab9e0683df6a36a2ddf68d41e75cfd1d3fd59).

waku/waku_lightpush/common.nim
fryorcraken: Sounds good to have it as an error.

s-tikhomirov: Fixed 0c6d42ff124a0d0fc39d6a1fb5b6de91d84ea1d7.

waku/waku_lightpush/client.nim
s-tikhomirov: Renamed `relayPeerCount` to `publishedPeerCount` everywhere: maintains consistency and IMO better reflects the meaning: f3f4452

waku/waku_lightpush/client.nim
```diff
@@ -59,100 +65,64 @@ proc sendPushRequest(
...
+    pubsubTopic: Option[PubsubTopic] = none(PubsubTopic),
     wakuMessage: WakuMessage,
-    peer: PeerId | RemotePeerInfo,
+    dest: Connection | PeerId | RemotePeerInfo,
 ): Future[WakuLightPushResult] {.async, gcsafe.} =
+  let conn =
+    when dest is Connection:
+      dest
+    else:
+      (await wl.peerManager.dialPeer(dest, WakuLightPushCodec)).valueOr:
```
Ivansete-status, line 87: Comment aside from PR.
We would need to avoid `dial` on every publish.
This should only happen once at the beginning of the interaction with a certain peer ( @richard-ramos , correct me if I'm wrong .)

waku/waku_lightpush/client.nim
```diff
@@ -59,100 +65,64 @@ proc sendPushRequest(
...
+
   var message = wakuMessage
-  if message.timestamp == 0:
-    message.timestamp = getNowInNanosecondTime()
+  ensureTimestampSet(message)
+
+  let msgHash = computeMessageHash(pubsubTopic.get(""), message).to0xHex
+  info "publish",
+    myPeerId = wl.peerManager.switch.peerInfo.peerId,
+    peerId =
```
Ivansete-status:
```suggestion
    toPeerId =
```

waku/waku_lightpush/client.nim
```diff
@@ -59,100 +65,64 @@ proc sendPushRequest(
...
+      when dest is Connection:
+        "unknown (using connection directly)"
```
Ivansete-status: `Connection` has a public `peerId` attribute ( see `vendor/nim-libp2p/libp2p/stream/connection.nim` .)
```suggestion
      when dest is Connection:
        shortPeerId(dest.peerId)
```

waku/waku_lightpush/client.nim
richard-ramos: This would depend on Lightpush server implementation. If it is expected to reuse a single stream (a la gossipsub peer handler), with the client using that same peer for sending further messages, then we need to have something like this on the server side handler:
```nim
try:  
    while not conn.atEof:
        # TODO: read / write from stream
except LPStreamError as e:
    # TODO: log error
except CancelledError as e:
     raise e
finally:
    await conn.closeWithEOF()
```

However, if stream is meant to be short lived, then dialing for each message is what should happen like in this code. Having said that, i went over to see the server implementation and found the following: https://github.com/waku-org/nwaku/blob/master/waku/waku_lightpush/protocol.nim#L108-L148  I could not find a conn.closeWithEOF() or close() for the handler, so looks like we have a leak here, right?

arnetheduck:
```
except CancelledError as e:
     raise e
```

this bit is not needed unless you catch all exceptions / `CatchableError`

Ivansete-status: Thanks for the comments!
[This one](https://github.com/waku-org/nwaku/pull/3538#discussion_r2369169642) has been resolved by @NagyZoltanPeter in https://github.com/waku-org/nwaku/pull/3588

waku/waku_lightpush/client.nim
```diff
@@ -59,100 +65,64 @@ proc sendPushRequest(
...
-      msg_hash = computeMessageHash(pubsubTopic.get(""), message).to0xHex
-  else:
-    info "publish",
-      peerId = shortLog(peer.peerId),
-      msg_hash = computeMessageHash(pubsubTopic.get(""), message).to0xHex
-
-  let pushRequest = LightpushRequest(
-    requestId: generateRequestId(wl.rng), pubSubTopic: pubSubTopic, message: message
+  let request = LightpushRequest(
+    requestId: generateRequestId(wl.rng), pubsubTopic: pubsubTopic, message: message
```
fryorcraken: I think there is a correction needed here. Yes, it's fine for a pubsub topic to be an option, but if it's not passed, then we should fallback to autosharding and deduce it from there.

I am not convinced that the autosharding deduction should be done here. I think we should make pubsubTopic mandatory for `proc publish*(wl: WakuLightPushClient,...)` and have the upper layer handle it.

waku/waku_lightpush/protocol.nim
```diff
@@ -28,59 +28,67 @@ type WakuLightPush* = ref object of LPProtocol
   requestRateLimiter*: RequestRateLimiter
   autoSharding: Option[Sharding]
 
-proc handleRequest(
-    wl: WakuLightPush, peerId: PeerId, pushRequest: LightpushRequest
-): Future[WakuLightPushResult] {.async.} =
-  let pubsubTopic = pushRequest.pubSubTopic.valueOr:
+proc extractPubsubTopic(
```
fryorcraken: nit: I am wondering if this logic should be handled by the `pushHandler` instead.

waku/waku_lightpush/client.nim
s-tikhomirov: Fixed in [3a8f126](https://github.com/waku-org/nwaku/pull/3538/commits/3a8f12682dae49a4d9f3e59ba4869fcbe2eed8a3).

waku/waku_lightpush/protocol.nim
s-tikhomirov: Could you elaborate? I'm not sure I understand how this relates to your other [suggestion](https://github.com/waku-org/nwaku/pull/3538#discussion_r2425193584).

fryorcraken: Well, the question is where should the pubsubtopic/sharding logic be applied? Within `handleRequest` ? or by the caller of `handleRequest`?

I am not sure which one is a best choice. My comment was here to raise awareness so you can think of it. Maybe @NagyZoltanPeter has an opinion as he touched light push recently.

NagyZoltanPeter: > Well, the question is where should the pubsubtopic/sharding logic be applied? Within `handleRequest` ? or by the caller of `handleRequest`?
> 
> I am not sure which one is a best choice. My comment was here to raise awareness so you can think of it. Maybe @NagyZoltanPeter has an opinion as he touched light push recently.

Hm, good question. 
I'm not against this change, meanwhile I think `pubsub-topic` or better `shard` handling shall be refactored separately from this. Due we have this shard derivative spread in many places of the code (waku interface, rest api, filter and lightpush). It would be good to concentrate into fewer places. Ideally from send perspective it would be good to place it into behind relay's publish.

One naming comment from my side for @s-tikhomirov, is I would name `extractPubsubTopic` rather `extractShard`.
WDYT?

s-tikhomirov: I suggest addressing this separately, I'm closing this conversation for now as the master branch has moved on since.

waku/waku_lightpush/client.nim
```diff
@@ -78,88 +81,54 @@ proc publish*(
...
+      (await wl.peerManager.dialPeer(dest, WakuLightPushCodec)).valueOr:
+        waku_lightpush_v3_errors.inc(labelValues = [dialFailure])
+        return lighpushErrorResult(
+          LightPushErrorCode.NO_PEERS_TO_RELAY,
+          dialFailure & ": " & $dest & " is not accessible",
+        )
+
+  defer:
+    when not (dest is Connection):
+      await conn.closeWithEOF()
```
Ivansete-status: We can do the following if we also remove that `closeWithEOF()` from `sendPushRequestToConn`
```suggestion
    await conn.closeWithEOF()
```

waku/waku_lightpush/client.nim
```diff
@@ -78,88 +81,54 @@ proc publish*(
...
 ): Future[WakuLightPushResult] {.async, gcsafe.} =
+  let conn =
+    when dest is Connection:
+      dest
+    else:
+      (await wl.peerManager.dialPeer(dest, WakuLightPushCodec)).valueOr:
+        waku_lightpush_v3_errors.inc(labelValues = [dialFailure])
+        return lighpushErrorResult(
+          LightPushErrorCode.NO_PEERS_TO_RELAY,
+          dialFailure & ": " & $dest & " is not accessible",
```
Ivansete-status: In case of error, I think is better to follow the pattern
<ERROR> + colon + <inner-ERROR>

That way, if we hard-code "ERROR:", f.e., it would make it easier to lookup in the code because we can search for "ERROR:". I hope I'm explaining well :)

```suggestion
          "Peer is not accessible: " & dialFailure & " - " & $dest,
```

waku/waku_lightpush/client.nim
```diff
@@ -78,88 +81,54 @@ proc publish*(
...
+  defer:
+    when not (dest is Connection):
+      await conn.closeWithEOF()
+
   var message = wakuMessage
-  if message.timestamp == 0:
-    message.timestamp = getNowInNanosecondTime()
+  ensureTimestampSet(message)
+
+  let msgHash = computeMessageHash(pubSubTopic.get(""), message).to0xHex
```
Ivansete-status: Nitpick one :)
```suggestion
  let msgHash = computeMessageHash(pubSubTopic.get(""), message).to0xHex()
```

PR #3540, opened 2025-08-13
chore: do not mount lightpush without relay (fixes #2808)

### LINKED ISSUE #2808: chore: do not mount lightpush without relay
### Background
This is a follow-on for a [point raised](https://github.com/waku-org/nwaku/pull/2768#discussion_r1635279154) by @NagyZoltanPeter at regarding mounting lightpush without relay.

### Details
It was like this before in waku_node.nim and now in waku_lightpush/callbacks.nim . But given a mounted relay is needed for a lightpush service node, it is a question of why we would want to still mount it and not fail early.

### Acceptance criteria
- [ ] Gather consensus on whether we would want to fail early and not start a missconfigured node with lightpush and without relay (@jm-clius @SionoiS @gabrielmer )
- [ ] If we agree on doing it, then go ahead with implementation

s-tikhomirov:
# Description

Previously, when mounting Lightpush, if Relay was not mounted, a nil Lightpush protocol handler was used. This PR changes node setup logic: mounting Lightpush without Relay mounted results in an error. The rationale behind this change is that a node cannot serve a Lightpush request without Relay. Fixes #2808.

# Changes

- [x] change node setup logic to return error if Lightpush is mounted without Relay
- [x] add test for this behavior
- [x] adapt other usages of Lightpush mounting procedure (its signature was changed to return a `Result`-type value)

### DIFF
--- a/waku/factory/node_factory.nim
+++ b/waku/factory/node_factory.nim
@@ -369,7 +369,10 @@ proc setupProtocols(
   # NOTE Must be mounted after relay
   if conf.lightPush:
     try:
-      await mountLightPush(node, node.rateLimitSettings.getSetting(LIGHTPUSH))
+      let lpRes =
+        await mountLightPush(node, node.rateLimitSettings.getSetting(LIGHTPUSH))
+      if lpRes.isErr():
+        return err("failed to mount waku lightpush protocol: " & $lpRes.error)
       await mountLegacyLightPush(node, node.rateLimitSettings.getSetting(LIGHTPUSH))
     except CatchableError:
       return err("failed to mount waku lightpush protocol: " & getCurrentExceptionMsg())
--- a/tests/node/test_wakunode_lightpush.nim
+++ b/tests/node/test_wakunode_lightpush.nim
@@ -47,7 +47,7 @@ suite "Waku Lightpush - End To End":
 
     (await server.mountRelay()).isOkOr:
       assert false, "Failed to mount relay"
-    await server.mountLightpush() # without rln-relay
+    check (await server.mountLightpush()).isOk() # without rln-relay
     client.mountLightpushClient()
 
     serverRemotePeerInfo = server.peerInfo.toRemotePeerInfo()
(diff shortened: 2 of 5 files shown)


waku/factory/node_factory.nim
```diff
@@ -369,7 +369,10 @@ proc setupProtocols(
   # NOTE Must be mounted after relay
   if conf.lightPush:
     try:
-      await mountLightPush(node, node.rateLimitSettings.getSetting(LIGHTPUSH))
+      let lpRes =
+        await mountLightPush(node, node.rateLimitSettings.getSetting(LIGHTPUSH))
+      if lpRes.isErr():
+        return err("failed to mount waku lightpush protocol: " & $lpRes.error)
```
Ivansete-status: This snippet can be slightly simplified with `isOkOr` technique:
```suggestion
      (await mountLightPush(node, node.rateLimitSettings.getSetting(LIGHTPUSH))).isOkOr:
        return err("failed to mount waku lightpush protocol: " & $error)
```

Ivansete-status (approved):
LGTM! Thanks for it! :100:

s-tikhomirov:
Thanks @NagyZoltanPeter , I have a question about this:

> First of all legacy lightpush shall be treated the same as v3.

I assumed that calling it "legacy" implies we are not making new changes to it. Is this not the case? Are we aiming to develop two versions of Lightpush in parallel, what is the rationale there?

NagyZoltanPeter:
> Thanks @NagyZoltanPeter , I have a question about this:
> 
> > First of all legacy lightpush shall be treated the same as v3.
> 
> I assumed that calling it "legacy" implies we are not making new changes to it. Is this not the case? Are we aiming to develop two versions of Lightpush in parallel, what is the rationale there?

Sure, unlikely to expect changes in legacy lightpush protocol. But the rationale with this treatment that from clients pov it is mainly one service. Of course there are clients supports only old one... hopefully for not tooo long, the rest can support both for now. So it is not very good that a service node will unexpetedly support only the old protocol while v3 is not mounted due to the circumstances of not having relay. Meanwhile, I know this is a very, very rare case.
Also think of ENR reports lightpush capability without saying the version, so it is better either have them both (until retire legacy) or none.

Ivansete-status:
Thanks for [this comment](https://github.com/waku-org/nwaku/pull/3540#pullrequestreview-3166790694) @NagyZoltanPeter 
I think we agreed to always consider that lightpush, filter and relay are all either mounted or unmounted. In that scenario, the Capabilities can get simpler as we can assume that, if a node mounts relay, then it will also mount both lightpush and filter.
But, as you mentioned, I think we can think about it in a future PR

weboko (approved):

Issue #3539, opened 2025-08-13
chore: use getShards before sending a metadata request


Ivansete-status:
### Background
Issue inspired by https://github.com/waku-org/nwaku/pull/3519#pullrequestreview-3073815696

### Details
Instead of having an event-driven approach  where metadata protocol knows the current shards set, the proposal is to invoke `getShards` before performing a metadata request.

The aim is to simplify the code and make sure always using a fresh shards set.

PR #3545, opened 2025-08-21
chore: use a callback to get latest shards for metadata

chaitanyaprem:
# Description
An attempt to address #3539. Comments/suggestions are welcome :)

# Changes

- [x] A callback that provides current subscribed shards
- [x] Metadata always uses shards available with the Monitor via a Getter API

Rendezvous changes would be taken up in a separate PR.

### DIFF
--- a/tests/test_peer_manager.nim
+++ b/tests/test_peer_manager.nim
@@ -625,6 +625,15 @@ procSuite "Peer Manager":
     await allFutures(nodes.mapIt(it.mountRelay()))
     await allFutures(nodes.mapIt(it.start()))
 
+    proc simpleHandler(
+        topic: PubsubTopic, msg: WakuMessage
+    ): Future[void] {.async, gcsafe.} =
+      await sleepAsync(0.millis)
+
+    let topic = "/waku/2/rs/0/0"
+    for node in nodes:
+      node.wakuRelay.subscribe(topic, simpleHandler)
+
     # Get all peer infos
     let peerInfos = collect:
       for i in 0 .. nodes.high:
--- a/waku/common/callbacks.nim
+++ b/waku/common/callbacks.nim
@@ -1,0 +1,1 @@
+type GetShards* = proc(): seq[uint16] {.closure, gcsafe, raises: [].}
--- a/waku/node/peer_manager/peer_manager.nim
+++ b/waku/node/peer_manager/peer_manager.nim
@@ -13,6 +13,7 @@ import
 import
   ../../common/nimchronos,
   ../../common/enr,
+  ../../common/callbacks,
   ../../common/utils/parse_size_units,
   ../../waku_core,
   ../../waku_relay,
@@ -99,6 +100,7 @@ type PeerManager* = ref object of RootObj
   shardedPeerManagement: bool # temp feature flag
   onConnectionChange*: ConnectionChangeHandler
   online: bool ## state managed by online_monitor module
+  getShards: GetShards
 
 #~~~~~~~~~~~~~~~~~~~#
 # Helper Functions  #
@@ -769,12 +771,12 @@ proc logAndMetrics(pm: PeerManager) {.async.} =
         protoStreamsOut.float64, labelValues = [$Direction.Out, proto]
       )
 
-    for shard in pm.wakuMetadata.shards.items:
+    for shard in pm.getShards().items:
       waku_connected_peers_per_shard.set(0.0, labelValues = [$shard])
 
-    for shard in pm.wakuMetadata.shards.items:
+    for shard in pm.getShards().items:
       let connectedPeers =
-        peerStore.getPeersByShard(uint16(pm.wakuMetadata.clusterId), uint16(shard))
+        peerStore.getPeersByShard(uint16(pm.wakuMetadata.clusterId), shard)
       waku_connected_peers_per_shard.set(
         connectedPeers.len.float64, labelValues = [$shard]
       )
@@ -788,8 +790,9 @@ proc getOnlineStateObserver*(pm: PeerManager): OnOnlineStateChange =
 #~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~#
 
 proc manageRelayPeers*(pm: PeerManager) {.async.} =
+  let shardsCount = pm.getShards().len
   #TODO: this check should not be based on whether shards are present, but rather if relay is mounted
-  if pm.wakuMetadata.shards.len == 0:
+  if shardsCount == 0:
     return
 
   if not pm.online:
@@ -803,19 +806,18 @@ proc manageRelayPeers*(pm: PeerManager) {.async.} =
   var (inPeers, outPeers) = pm.connectedPeers(WakuRelayCodec)
 
   # Calculate in/out target number of peers for each shards
-  let inTarget = pm.inRelayPeersTarget div pm.wakuMetadata.shards.len
-  let outTarget = pm.outRelayPeersTarget div pm.wakuMetadata.shards.len
+  let inTarget = pm.inRelayPeersTarget div shardsCount
+  let outTarget = pm.outRelayPeersTarget div shardsCount
 
   var peerStore = pm.switch.peerStore
 
-  for shard in pm.wakuMetadata.shards.items:
+  for shard in pm.getShards().items:
     # Filter out peer not on this shard
-    let connectedInPeers = inPeers.filterIt(
-      peerStore.hasShard(it, uint16(pm.wakuMetadata.clusterId), uint16(shard))
-    )
+    let connectedInPeers =
+      inPeers.filterIt(peerStore.hasShard(it, uint16(pm.wakuMetadata.clusterId), shard))
 
     let connectedOutPeers = outPeers.filterIt(
-      peerStore.hasShard(it, uint16(pm.wakuMetadata.clusterId), uint16(shard))
+      peerStore.hasShard(it, uint16(pm.wakuMetadata.clusterId), shard)
     )
 
     # Calculate the difference between current values and targets
@@ -1000,6 +1002,9 @@ proc addExtPeerEventHandler*(
 #~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~#
 # Initialization and Constructor #
 #~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~#
+
+proc setShardGetter*(pm: PeerManager, c: GetShards) =
+  pm.getShards = c
 
 proc start*(pm: PeerManager) =
   pm.started = true
--- a/waku/node/waku_node.nim
+++ b/waku/node/waku_node.nim
@@ -48,7 +48,8 @@ import
   ../waku_rln_relay,
   ./net_config,
   ./peer_manager,
-  ../common/rate_limit/setting
+  ../common/rate_limit/setting,
+  ../common/callbacks
 
 declarePublicCounter waku_node_messages, "number of messages received", ["type"]
 declarePublicHistogram waku_histogram_message_size,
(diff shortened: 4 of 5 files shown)


waku/common/shard_subscription_monitor.nim
```diff
@@ -0,0 +1,36 @@
+{.push raises: [].}
+
+import std/sets, sequtils, results, chronicles
+import ../waku_core/topics/pubsub_topic
+
+type ShardsSubscriptionMonitor* = ref object of RootObj
```
Ivansete-status: I think not needed `of RootObj`. Only use that if want to make this the base class and have inheritance underneath
```suggestion
type ShardsSubscriptionMonitor* = ref object
```

waku/common/shard_subscription_monitor.nim
```diff
@@ -0,0 +1,36 @@
...
+
+proc onShardSubscribed*(self: ShardsSubscriptionMonitor, shard: uint16) =
+  self.subscribedShards.incl(shard)
+
+proc onShardUnSubscribed*(self: ShardsSubscriptionMonitor, shard: uint16) =
+  self.subscribedShards.excl(shard)
+
+proc onTopicSubscribed*(self: ShardsSubscriptionMonitor, topic: PubsubTopic) =
+  let shard = RelayShard.parse(topic).valueOr:
+    error "Invalid pubsub topic", error = error
```
Ivansete-status: To make it faster to debug:
```suggestion
    error "Invalid pubsub topic in onTopicSubscribed", error = error
```

waku/common/shard_subscription_monitor.nim
```diff
@@ -0,0 +1,36 @@
...
+    return
+  self.onShardSubscribed(shard.shardId)
+
+proc onTopicUnSubscribed*(self: ShardsSubscriptionMonitor, topic: PubsubTopic) =
+  let shard = RelayShard.parse(topic).valueOr:
+    error "Invalid pubsub topic", error = error
```
Ivansete-status:
```suggestion
    error "Invalid pubsub topic in onTopicUnSubscribed", error = error
```

waku/node/waku_node.nim
```diff
@@ -384,7 +386,7 @@ proc unsubscribe*(
   debug "unsubscribe", pubsubTopic, contentTopicOp
   node.wakuRelay.unsubscribe(pubsubTopic)
   node.topicSubscriptionQueue.emit((kind: PubsubUnsub, topic: pubsubTopic))
-
+  node.shardSubscriptionMonitor.onTopicUnSubscribed(pubsubTopic)
```
Ivansete-status: The previous line, `node.topicSubscriptionQueue.emit((kind: PubsubSub, topic: $pubsub))` seems not needed now.
This suggestion applies elsewhere `emit` is invoked ;P

waku/factory/waku.nim
```diff
@@ -64,7 +62,6 @@ type Waku* = ref object
   dynamicBootstrapNodes*: seq[RemotePeerInfo]
   dnsRetryLoopHandle: Future[void]
   networkConnLoopHandle: Future[void]
-  discoveryMngr: DiscoveryManager
```
Ivansete-status: These changes seem unrelated to the PR's purpose (not a big deal though.)

waku/node/peer_manager/peer_manager.nim
```diff
@@ -99,6 +100,7 @@ type PeerManager* = ref object of RootObj
   shardedPeerManagement: bool # temp feature flag
   onConnectionChange*: ConnectionChangeHandler
   online: bool ## state managed by online_monitor module
+  shardSubscriptionMonitor*: ShardsSubscriptionMonitor
```
Ivansete-status: It doesn't look 100% nice having `shardSubscriptionMonitor` owned by multiple items. Even though it is `ref object`, in the future it may change to `object` and that can cause bugs.
Therefore, I think `shardSubscriptionMonitor` should be owned by one single entity.

A better place would be `WakuMetadata` to own `shardSubscriptionMonitor`.

waku/waku_metadata/protocol.nim
```diff
@@ -22,14 +21,15 @@ const RpcResponseMaxBytes* = 1024
 
 type WakuMetadata* = ref object of LPProtocol
   clusterId*: uint32
-  shards*: HashSet[uint32]
-  topicSubscriptionQueue: Option[AsyncEventQueue[SubscriptionEvent]]
+  shardSubscriptionMonitor: ShardsSubscriptionMonitor
```
Ivansete-status: Looks great! I think this is the only place where we should put the `ShardsSubscriptionMonitor` :)

Ivansete-status (commented):
I'm adding some nitpick comments that I hope you find useful

waku/node/waku_node.nim
chaitanyaprem: I thought of removing it, but noticed that it is emitting some events probably being exposed by waku-lib for apps? 
Hence kept it as is, if we are not using this queue for anything else then i will remove completely its reference from the code.

waku/factory/waku.nim
chaitanyaprem: yea, got confused by this code and looked into nim-libp2p but later realized that we are not using this DiscoveryManager at all and hence removed it.

waku/node/peer_manager/peer_manager.nim
chaitanyaprem: I agree, ideally this should be owned only by WakuNode and everywhere else refs should be used?

I don't think metadata should own it, rather metadata is just a user using the API of this object. There would be other users such as Rendezvous etc. Hence i think this can be owned by WakuNode. It could also be owned by WakuRelay but i did not want to tie it to relay as Filter subscriptions also should count.

Ivansete-status: Okay perfect, then let's make it owned by `WakuNode`.
On the other hand, can it be owned by `Waku` ( a layer higher than `WakuNode` ) ?

chaitanyaprem: Well, that is another point i wanted to bring out separately. Why do we have Waku on top of WakuNode? 

Wouldn't just exposing WakuNode be enough for apps building on top of Waku - i.e any app can create a WakuNode and work with it. 

Unless i am missing something.

waku/node/waku_node.nim
chaitanyaprem: And i had noticed node ENR is updated based on events emitted in this queue.
One way to do it is to migrate event logic also to this monitor and then use those events from there.

waku/common/shard_monitor.nim
```diff
@@ -0,0 +1,44 @@
...
+  return sm
+
+proc getSubscribedShards*(self: ShardMonitor): seq[uint16] =
+  return toseq(self.subscribedShards)
+
+#This is implemented temporarily until all shard related refs are moved to uint16
+proc getSubscribedShardsUint32*(self: ShardMonitor): seq[uint32] =
+  return self.subscribedShards.mapIt(it.uint32)
+
+proc onShardSubscribed*(self: ShardMonitor, shard: uint16) =
```
fryorcraken: The `on` terminology is usually:

```
emitter.on("event-name", callback)
```

So it's a bit weird to have `shardMonitor.onShardSubscribed`.

some suggestions:

```suggestion
proc recordShardSubscribed*(self: ShardMonitor, shard: uint16) =
```

```suggestion
proc registerShardSubscribed*(self: ShardMonitor, shard: uint16) =
```

waku/node/waku_node.nim
```diff
@@ -384,7 +386,7 @@ proc unsubscribe*(
   debug "unsubscribe", pubsubTopic, contentTopicOp
   node.wakuRelay.unsubscribe(pubsubTopic)
   node.topicSubscriptionQueue.emit((kind: PubsubUnsub, topic: pubsubTopic))
```
fryorcraken: instead of adding a reference to the shard monitor in this method, why not having the shard monitor listen to `node.topicSubscriptionQueue`?

chaitanyaprem: Well, i actually wanted to move the queue itself into ShardMonitor as these are just multiple entities watching and reporting same thing i.e shard subscriptions. I want to keep a single place where subscriptions for shards are watched and notified from.

fryorcraken:
> Let me explore more of nim to see if this can be done without creating a dependency. one way i can think of is via eventQueues but maybe there is a simpler way to do.

IMO, make metadata accept a `proc` when initialized, and write a `proc` that you can pass for a given `Relay` Instance:

(this code was fully written by Claude and is working)

```nim
type
    GetShardsCallback = proc(): seq[int] {.closure.}

    Relay = object
      shards: seq[int]

    Metadata = object
      getShardsFunc: GetShardsCallback

  proc newRelay(numbers: seq[int]): Relay =
    result.shards = numbers

  proc getShards(relay: Relay): seq[int] =
    return relay.shards

  proc newMetadata(callback: GetShardsCallback): Metadata =
    result.getShardsFunc = callback

  proc process(metadata: Metadata) =
    let shards = metadata.getShardsFunc()
    echo "Shards: ", shards

  proc main() =
    let relay = newRelay(@[1, 2, 3, 4, 5])

    let getShards = proc(): seq[int] =
      relay.getShards()

    let metadata = newMetadata(getShards)

    metadata.process()

  when isMainModule:
    main()

```

fryorcraken:
> > Excellent insight. I am in favour of Relay only doing with Shards from an (relay's) API PoV, even if internally, relay has to deal with pubsub topics when dealing with libp2p-gossipsub.
> 
> Yep, i am completely in favor of doing this. Maybe in a separate PR.

I think for now, instead of creating a whole monitor, just have one function helper that can convert relay's data (pubsubtopics) in metadata's data (shards), that can be use as the `GetShardsCallback` type in my example above.

waku/common/callbacks.nim
```diff
@@ -0,0 +1,3 @@
+{.push raises: [].}
+
+type GetShards* = proc(): seq[uint16] {.closure, gcsafe, raises: [].}
```
Ivansete-status, line 1: I think this is not needed because you are passing it above to the whole module
```suggestion
type GetShards* = proc(): seq[uint16] {.closure, gcsafe.}
```

fryorcraken (approved):
Beautiful!

tests/test_peer_manager.nim
```diff
@@ -625,6 +625,16 @@ procSuite "Peer Manager":
...
 
+    proc simpleHandler(
+        topic: PubsubTopic, msg: WakuMessage
+    ): Future[void] {.async, gcsafe.} =
+      await sleepAsync(0.millis)
+
+    let topic = "/waku/2/rs/0/0"
+    for node in nodes:
+      node.wakuRelay.subscribe(topic, simpleHandler)
+    await sleepAsync(500.millis)
```
Ivansete-status: Is it possible to avoid that wait ?
```suggestion
```

waku/node/waku_node.nim
```diff
@@ -123,6 +123,19 @@ type
     topicSubscriptionQueue*: AsyncEventQueue[SubscriptionEvent]
     rateLimitSettings*: ProtocolRateLimitSettings
 
+proc getShards*(node: WakuNode): seq[uint16] {.gcsafe, raises: [].} =
```
Ivansete-status: This should be better be private:
```suggestion
proc getShards(node: WakuNode): seq[uint16] {.gcsafe, raises: [].} =
```

------

Besides, another possible option:
```nim
proc getShardsGetter(node: WakuNode): GetShards =
  return proc(): seq[uint16] {.closure, gcsafe, raises: [].} =
    # fetch pubsubTopics subscribed to relay and convert them to shards
   if node.wakuRelay.isNil():
      return @[]
   let subTopics = node.wakuRelay.subscribedTopics()
    let relayShards = topicsToRelayShards(subTopics).valueOr:
      error "could not convert relay topics to shards", error = $error
      return @[]
    if relayShards.isSome():
      let shards = relayShards.get().shardIds
      return shards
      
 ...
 
 peerManager.setShardGetter(node.getShardsGetter())
```

chaitanyaprem:
> I think we can also clean all the calls to `node.topicSubscriptionQueue.emit(...` given that we are cleaning the listener

I have noticed this is used by discv5 as well to update shards in ENR. 
This cannot be migrated to a callback approach and requires events to be generated on subscribe/unsubscribe. 
So we can't clean this.

tests/test_peer_manager.nim
chaitanyaprem: Should be fine i guess, will try it out and see.

waku/node/waku_node.nim
chaitanyaprem: > Besides, another possible option:

I like this approach as i don't have to keep creating local procs wherever i need to pass the Getshards.

waku/common/callbacks.nim
chaitanyaprem: I am getting a compilation error if i remove this , so removing push macro at the top of the file.

PR #3550, opened 2025-08-27
fix: make test <nim source file>

NagyZoltanPeter:
## Description
I noticed that I cannot run a single file test with `make test <nim source file>` command without specifying test case name or pattern

## Changes

This PR fixes the argument passing from Makefile to nimble `executeTest` task.
It will always pass at least empty quotes when no test case specified. Also `executeTest` task will take care of handling that situation properly.

This way it should be possible to run:
`make test`
`make test tests/<test file name>.nim`
and `make test tests/<test file name>.nim "<test case name>"`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -259,9 +259,9 @@ build/%: | build deps librln
 		$(ENV_SCRIPT) nim buildone $(NIM_PARAMS) waku.nims $*
 
 compile-test: | build deps librln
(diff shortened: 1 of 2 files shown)


Ivansete-status (approved):
LGTM! Thanks for it! :100:

darshankabariya (approved):
Thanks, I’ve also faced a similar issue when the test case name isn’t alphabetical or some sign like `-, +, {, }`

PR #3552, opened 2025-08-27
fix remove libpcre lib dependency used for docker builds 

NagyZoltanPeter:
## Description
Seems with debian:stable-slim now points to Debian Bookworm and that made libpcre3 deprecated we have a failing docker build

## Changes

As we already removed std/re dependencies and moved to using nim-regex lib that does not depend on pcre external lib.
As libpcre3 is reported to be vulnerable, it is betterto  completely remove it.

## Issue

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -43,9 +43,6 @@ EXPOSE 30303 60000 8545
 # Referenced in the binary
 RUN apk add --no-cache libgcc libpq-dev bind-tools
 
-# Fix for 'Error loading shared library libpcre.so.3: No such file or directory'
-# RUN ln -s /usr/lib/libpcre.so /usr/lib/libpcre.so.3
-
 # Copy to separate location to accomodate different MAKE_TARGET values
 COPY --from=nim-build /app/build/$MAKE_TARGET /usr/local/bin/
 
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -41,10 +41,10 @@ LABEL version="unknown"
 EXPOSE 30303 60000 8545
 
 # Referenced in the binary
-RUN apk add --no-cache libgcc pcre-dev libpq-dev bind-tools
+RUN apk add --no-cache libgcc libpq-dev bind-tools
 
 # Fix for 'Error loading shared library libpcre.so.3: No such file or directory'
-RUN ln -s /usr/lib/libpcre.so /usr/lib/libpcre.so.3
+# RUN ln -s /usr/lib/libpcre.so /usr/lib/libpcre.so.3
```
Ivansete-status: Shall we remove these two lines if not needed anymore?

Ivansete-status (approved):
LGTM as long as the docker image can run properly :)

darshankabariya (approved):
LGTM! I made the same change in the refactor branch but haven’t merged it yet. I’ll revert those changes.

PR #3555, opened 2025-08-29
fix: nimble libraries

kaichaosun:
Lock the libp2p version, add missing dnsdisc.

Related to https://github.com/waku-org/nim-chat-sdk/issues/7


waku.nimble
```diff
@@ -13,14 +13,15 @@ requires "nim >= 2.2.4",
...
+  "dnsdisc",
   "eth",
   "json_rpc",
   "libbacktrace",
   "nimcrypto",
   "stew",
   "stint",
   "metrics",
-  "libp2p", # Only for Waku v2
+  "libp2p == 1.10.1", # Only for Waku v2
```
Ivansete-status, line 24: Maybe better to use the latest one?
```suggestion
  "libp2p == 1.12.0", # Only for Waku v2
```

kaichaosun: Latest libp2p has compatible issues in my test, also the libp2p used in nwaku/vendors is still 1.10.1, we should make it consistent IMO. Let's do the upgrade in new tasks.

NagyZoltanPeter:
> @NagyZoltanPeter I didn't check all the deps in nimble file with vendors, the two libs updated in this PR is checked though, we'll see if more deps needs to be locked when I make further progress.

Sure, it is ok for now! I think we made a deliverable around to move to nimble packaging and build for nwaku. Coming sooner or later...

PR #3560, opened 2025-09-08
chore: an attempt to move node API's to separate files

chaitanyaprem:
## Description
An attempt to address [this](https://github.com/waku-org/nwaku/pull/3523#discussion_r2244318662) and to reduce size of `waku_node.nim` file as it is going out of control.

Please review and suggest any better way to do this.

## Changes
 
- [x] Move protocol specific API code out of waku_node file.


chaitanyaprem:
> Love this initiative @chaitanyaprem!!! One little remark, I would even separate the mountings from the api. As thinking of future Waku Core API/SDK, mounts are not part of it but the rest of the interface procs are highly are! WDYT?

Good Point, at first i did not move the mounts but later felt they belong in the same file. I am not too sure either, but if we are thinking of from an SDK point of view, maybe it makes sense. Let's wait for what others have to say.

PR #3561, opened 2025-09-08
chore: bump mix

richard-ramos:


waku/waku_mix/protocol.nim
```diff
@@ -142,13 +142,16 @@ proc new*(
     peermgr.switch.peerInfo.privateKey.skkey,
   )
 
+  let mixProto =
+    MixProtocol.new(localMixNodeInfo, initTable[PeerId, MixPubInfo](), peermgr.switch)
+
+  let wakuMix = WakuMix(mixProto)
```
chaitanyaprem: This fails during runtime as both are ref objects and throws an `invalid object conversion [ObjectConversionDefect]`

The previous method of using procCall doesn't seem to work with `new`. Any idea how to fix it?

richard-ramos:
> Getting a compilation error while building `wakunode2`. Error seems to be generated from trace logs present in `MixEntryConnection`,
> 
> ```
> /Users/prem/Code/nwaku/vendor/mix/mix/entry_connection.nim(76, 5) template/generic instantiation of `trace` from here
> /Users/prem/Code/nwaku/vendor/nim-chronicles/chronicles.nim(383, 10) template/generic instantiation of `log` from here
> /Users/prem/Code/nwaku/vendor/nim-chronicles/chronicles/scope_helpers.nim(10, 18) template/generic instantiation of `expandItIMPL` from here
> /Users/prem/Code/nwaku/vendor/nim-chronicles/chronicles/log_output.nim(835, 39) template/generic instantiation of `setProperty` from here
> /Users/prem/Code/nwaku/vendor/nim-chronicles/chronicles/log_output.nim(644, 17) Error: type mismatch
> Expression: $val
>   [1] val: MixEntryConnection
>  ```
> If i commented the trace log line, error goes away. Were you facing this error in the mix repo as well?
>

Fixed in https://github.com/waku-org/nwaku/pull/3561/commits/bae2d92d461a3d80a5b306883875af88615b6d79 . Interestngly enough i did not face such problem in mix repo

waku/waku_mix/protocol.nim
richard-ramos: Fixed in https://github.com/waku-org/nwaku/pull/3561/commits/8ae443dfafeff31e19398c0eb23f4305359267ae
Ended up adding back an init function :(

chaitanyaprem: > Ended up adding back an init function :(

thought so, i tried with github copilot to resolve this using many options and only one it has recommended was to add init function back in MixProtocol.

chaitanyaprem (approved):
LGTM

Thanks a lot for this!

waku/node/waku_switch.nim
```diff
@@ -47,14 +47,10 @@ proc withWssTransport*(
...
-      WsTransport.new(
-        upgr,
-        tlsPrivateKey = key,
-        tlsCertificate = cert,
-        {TLSFlags.NoVerifyHost, TLSFlags.NoVerifyServerName},
-      )
+  b.withWsTransport(
+    tlsPrivateKey = key,
+    tlsCertificate = cert,
+    {TLSFlags.NoVerifyHost, TLSFlags.NoVerifyServerName}, # THIS IS INSECURE, NO?
```
chaitanyaprem, line 53: @Ivansete-status maybe has an answer as to why these params are set this way.

Ivansete-status: These flags are there since [this commit](d1e06fa17ab7989e487b095548543df0936399fb) in 2021.

It seems we are loosing security constraints there.

- `NoVerifyHost`: that skips validating the host is contained in the provided certificate

- `NoVerifyServerName`: I believe the purpose of that is to allow connections against raw ip addresses and not only server names.

More details in https://github.com/status-im/nim-chronos/blob/master/chronos/streams/tlsstream.nim

Nevertheless, I don't foresee to establish wss connections from `nwaku` in short-term because this will be mostly tackled by js-waku. @weboko - correct me if this last statement is not accurate.

PR #3565, opened 2025-09-11
chore: add missing metrics

darshankabariya:
This PR aims to add the remaining support for the metrics required in #3306  

The status team needs shard information related to these metrics `waku_relay_network_bytes` `waku_connected_peers_per_shard` and `waku_rln_valid_messages`


waku/node/peer_manager/peer_manager.nim
```diff
@@ -72,7 +74,7 @@ const
   PrunePeerStoreInterval = chronos.minutes(10)
 
   # How often metrics and logs are shown/updated
-  LogAndMetricsInterval = chronos.minutes(3)
+  LogAndMetricsInterval = chronos.minutes(5)
```
Ivansete-status, line 77: Why not 3 :) ?

fryorcraken (approved):
LGTM

waku/node/peer_manager/peer_manager.nim
```diff
@@ -771,14 +774,33 @@ proc logAndMetrics(pm: PeerManager) {.async.} =
...
+          continue
+        let agent = agentStr.split("/")[0]
+        agentCounts[agent] = agentCounts.getOrDefault(agent, 0) + 1
+    for agent, count in agentCounts:
+      waku_connected_peers_per_agent.set(count.float64, labelValues = [$agent])
+
+    # connected-peers-per-shard
+    var connectedPeerIds: HashSet[PeerId]
+    for peerId, _ in connections:
+      connectedPeerIds.incl(peerId)
```
Ivansete-status: It seems we can avoid the double loop.

Besides, it is also interesting to keep stats about empty agents and agent's versions. If you agree with that, we can simplify it a little.

```suggestion
    var agentCounts = initTable[string, int]()
    var connectedPeerIds: HashSet[PeerId]
    for peerId, muxers in connections:
      connectedPeerIds.incl(peerId)
      if peerStore[AgentBook].contains(peerId):
        let agent = peerStore[AgentBook][peerId]
        agentCounts[agent] = agentCounts.getOrDefault(agent, 0) + 1
    for agent, count in agentCounts:
      waku_connected_peers_per_agent.set(count.float64, labelValues = [$agent])

```

waku/node/peer_manager/peer_manager.nim
darshankabariya: Actually, the complexity of `logAndMetrics` has increased, so I feel 5 is a better choice than 3.

waku/node/peer_manager/peer_manager.nim
darshankabariya: Cool thanks !!

PR #3567, opened 2025-09-11
chore: benchmark for proof generation and verification

darshankabariya:
## Issue
- https://github.com/waku-org/pm/issues/296

### DIFF
--- a/apps/liteprotocoltester/diagnose_connections.nim
+++ b/apps/liteprotocoltester/diagnose_connections.nim
@@ -27,19 +27,6 @@ import
 logScope:
   topics = "diagnose connections"
 
-proc `$`*(cap: Capabilities): string =
-  case cap
-  of Capabilities.Relay:
-    return "Relay"
-  of Capabilities.Store:
-    return "Store"
-  of Capabilities.Filter:
-    return "Filter"
-  of Capabilities.Lightpush:
-    return "Lightpush"
-  of Capabilities.Sync:
-    return "Sync"
-
 proc allPeers(pm: PeerManager): string =
   var allStr: string = ""
   for idx, peer in pm.switch.peerStore.peers():
--- a/apps/liteprotocoltester/v3_publisher.nim
+++ b/apps/liteprotocoltester/v3_publisher.nim
@@ -20,7 +20,7 @@ method send*(
   discard (
     await self.wakuNode.lightpushPublish(some(topic), message, some(servicePeer))
   ).valueOr:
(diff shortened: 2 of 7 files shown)


apps/benchmarks/benchmarks.nim
```diff
@@ -11,32 +11,29 @@ import
...
+    try:
+      await manager.register(idCredentials[i], UserMessageLimit(messageLimit + 1))
+    except Exception, CatchableError:
+      assert false, "exception raised: " & getCurrentExceptionMsg()
 
-  try:
-    await manager.register(idCredentials, UserMessageLimit(100))
-  except Exception, CatchableError:
-    assert false, "exception raised: " & getCurrentExceptionMsg()
+    if i >= registerCount - 50:
```
stubbsta: What is the purpose of this exactly?

darshankabariya: Generally, only need the last 50 roots for proof generation and verification because our validRoots size is 50. 

Logically here, we only use the most recent root for generating and verifying proofs bcz we are not generating or verifying proofs in between registration, so the latest root works fine. will update code accordingly. ( cc @stubbsta  )

stubbsta: Thanks, this totally makes sense, it just wasn't clear from the code, a comment to explain it would work.

apps/benchmarks/benchmarks.nim
```diff
@@ -11,32 +11,27 @@ import
...
-  if rootUpdated:
-    let proofResult = await manager.fetchMerkleProofElements()
-    if proofResult.isErr():
-      error "Failed to fetch Merkle proof", error = proofResult.error
-    manager.merkleProofCache = proofResult.get()
+  discard await manager.updateRoots()
+  let proofResult = await manager.fetchMerkleProofElements()
+  if proofResult.isErr():
+    error "Failed to fetch Merkle proof", error = proofResult.error
+  manager.merkleProofCache = proofResult.get()
```
Ivansete-status, line 34: Maybe?
```suggestion
  manager.merkleProofCache = await manager.fetchMerkleProofElements().valueOr:
    error "Failed to fetch Merkle proof", error = error
```

darshankabariya: Done

apps/benchmarks/benchmarks.nim
darshankabariya: Could you plz approve this @stubbsta ??

PR #3569, opened 2025-09-15
feat: adding rendezvous request interval

### LINKED ISSUE #3549: bug: Rendezvous peer discovery is not functioning
### Problem
Rendezvous peer discovery is not functioning as expected.  
When running the rendezvous tests, nodes fail to discover peers through the rendezvous mechanism, which contradicts the assumptions outlined in [Rendezvous FURPS spec](https://github.com/waku-org/pm/blob/master/FURPS/core/rendezvous.md).

### To reproduce
Use the test branch:  
https://github.com/waku-org/waku-interop-tests/pull/132

and check test **test_basic_rendezvous_register_and_discover**

### Expected behavior
Node3 should discover at least one peer via rendezvous.  

### Actual behavior
Node3 discovers **zero peers**, causing the assertion to fail.

### Logs / Screenshots
Example test failure:

<pre> ``` Node3 peers count is : [] AssertionError: No peers discovered via Rendezvous ``` </pre>

SionoiS:
## Description

I added an exponential back-off to the initial request done by the rendezvous code.

The goal was to balance the number of retries with the extra load on the network without adding too much complexity.

WDYT?

closes #3549

### DIFF
--- a/waku/waku_rendezvous/protocol.nim
+++ b/waku/waku_rendezvous/protocol.nim
@@ -1,7 +1,7 @@
 {.push raises: [].}
 
 import
-  std/[sugar, options],
+  std/sugar,
   results,
   chronos,
   chronicles,
@@ -10,14 +10,7 @@ import
   libp2p/switch,
   libp2p/utility
 
-import
-  ../node/peer_manager,
-  ../common/callbacks,
-  ../waku_enr/capabilities,
-  ../waku_core/peers,
-  ../waku_core/topics,
-  ../waku_core/topics/pubsub_topic,
-  ./common
+import ../node/peer_manager, ../common/callbacks, ./common
 
 logScope:
   topics = "waku rendezvous"
@@ -34,6 +27,9 @@ type WakuRendezVous* = ref object
 
   registrationInterval: timer.Duration
   periodicRegistrationFut: Future[void]
+
+  requestInterval: timer.Duration
+  periodicRequestFut: Future[void]
 
 proc batchAdvertise*(
     self: WakuRendezVous,
@@ -214,7 +210,7 @@ proc initialRequestAll*(
         rendezvousPeerFoundTotal.inc()
         self.peerManager.addPeer(record)
 
-  debug "waku rendezvous initial requests finished"
+  debug "waku rendezvous initial request finished"
 
   return ok()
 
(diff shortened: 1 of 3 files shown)


waku/waku_rendezvous/protocol.nim
```diff
@@ -276,6 +293,8 @@ proc start*(self: WakuRendezVous) {.async: (raises: []).} =
   # start registering forever
   self.periodicRegistrationFut = self.periodicRegistration()
 
+  self.periodicRequestFut = self.periodicRequests()
```
Ivansete-status, line 300: Maybe missing to cancel the future when `stopWait` is invoked?

waku/waku_rendezvous/protocol.nim
```diff
@@ -237,6 +240,19 @@ proc periodicRegistration(self: WakuRendezVous) {.async.} =
     # Back to normal interval if no errors
     self.registrationInterval = DefaultRegistrationInterval
 
+proc periodicRequests(self: WakuRendezVous) {.async.} =
+  debug "waku rendezvous periodic requests started", interval = self.requestInterval
+
+  # infinite loop
+  while true:
+    (await self.initialRequestAll()).isOkOr:
+      debug "waku rendezvous requests failed", error = error
```
Ivansete-status: Perhaps?
```suggestion
      error "waku rendezvous requests failed", error = error
```

waku/waku_rendezvous/protocol.nim
SionoiS: ah yes good catch!

waku/waku_rendezvous/protocol.nim
```diff
@@ -237,6 +240,19 @@ proc periodicRegistration(self: WakuRendezVous) {.async.} =
...
+
+  # infinite loop
+  while true:
+    (await self.initialRequestAll()).isOkOr:
+      error "waku rendezvous requests failed", error = error
+
+    await sleepAsync(self.requestInterval)
+
+    # Exponential backoff
+    self.requestInterval += self.requestInterval
```
jm-clius, line 255: Since `sleepAsync` works with a timer thread (IIRC), wouldn't it be best to set a cleaner stop condition here to remove state? For all practical purposes the interval will become "infinite" compared to node run times very soon, so we might as well cap this to run no more than, say, 8 times (after which the interval will already be several hours).

SionoiS: Yes indeed cleaning up is a good idea. I'll do that.

PR #3571, opened 2025-09-16
chore: update nimble for nwaku lib integration.

kaichaosun:
Fix the dependencies in nimble, so nwaku can be used as a library in other nim codebase.

Replace: https://github.com/waku-org/nwaku/pull/3556

### DIFF
--- a/waku.nimble
+++ b/waku.nimble
@@ -20,6 +20,7 @@ requires "nim >= 2.2.4",
   "json_rpc",
   "libbacktrace",
   "nimcrypto",
+  "serialization",
   "stew",
   "stint",
   "metrics",
@@ -30,7 +31,6 @@ requires "nim >= 2.2.4",
   "results",
   "db_connector",
   "minilru",
-  "quic",
   "https://github.com/vacp2p/mix#0.1.0"
 
 ### Helper functions


waku.nimble
```diff
@@ -18,17 +18,18 @@ requires "nim >= 2.2.4",
   "json_rpc",
   "libbacktrace",
   "nimcrypto",
+  "serialization",
   "stew",
   "stint",
   "metrics",
-  "libp2p == 1.10.1", # Only for Waku v2
+  "libp2p >= 1.12.0",
```
fryorcraken: the issue is that vendor/nim-libp2p does not contain 1.12.0 or a tagged version, so it means that nimble and vendor are still unsync after this PR

kaichaosun: Yeah, we should only use one build system, either nbs or nimble. It's just hard to make it sync with two different flavors.

fryorcraken: as per discord, I suggest

1. get waku to work **as** a nimble lib. `vendor/*` and `waku.nimble` are sync'd manually but GH action ensures it is sync'd
2. Get the `apps/*` (eg `wakunode2`) one by one to be built via nimble, as waku can now be imported via nimble.
3. finish off by getting rid of nimble-build-system and the `vendor/`

Cc @waku-org/nwaku

fryorcraken:
> Checked with nph format, it makes the waku.nimble in bad format.

I think it would better to have

```
requires libp2p
requires chronos
```

aka, repeat the `requires` keyword so it's one line per deps. So that git history is cleaner re tracking deps versions

Ivansete-status (approved):

PR #3573, opened 2025-09-17
chore: fix node break issue when RLN is unregistered

darshankabariya:
Merkle proof elements are not needed when a node is not registered for RLN. However, the current trackRootChanges function still requires the merkle proof, causing the node to break. This change resolves the issue.

### DIFF
--- a/waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
+++ b/waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
@@ -202,8 +202,10 @@ proc trackRootChanges*(g: OnchainGroupManager) {.async: (raises: [CatchableError
       let rootUpdated = await g.updateRoots()
 
       if rootUpdated:
+        ## The membership set on-chain has changed (some new members have joined or some members have left)
         if g.membershipIndex.isSome():
-          # merkleProofElements is only needed for rln registered nodes
+          ## A membership index exists only if the node has registered with RLN.
+          ## Non-registered nodes cannot have Merkle proof elements.
           let proofResult = await g.fetchMerkleProofElements()
           if proofResult.isErr():
             error "Failed to fetch Merkle proof", error = proofResult.error
--- a/waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
+++ b/waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
@@ -201,22 +201,23 @@ proc trackRootChanges*(g: OnchainGroupManager) {.async: (raises: [CatchableError
       await sleepAsync(rpcDelay)
       let rootUpdated = await g.updateRoots()
 
-      if rootUpdated:
+      # merkleProofElements is only needed for rln registered nodes
+      if rootUpdated and g.membershipIndex.isSome():
         let proofResult = await g.fetchMerkleProofElements()
         if proofResult.isErr():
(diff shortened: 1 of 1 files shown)


waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
```diff
@@ -201,22 +201,23 @@ proc trackRootChanges*(g: OnchainGroupManager) {.async: (raises: [CatchableError
...
-          error "Failed to fetch next free index", error = nextFreeIndex.error
-          raise newException(
-            CatchableError, "Failed to fetch next free index: " & nextFreeIndex.error
-          )
+      let nextFreeIndex = await g.fetchNextFreeIndex()
+      if nextFreeIndex.isErr():
+        error "Failed to fetch next free index", error = nextFreeIndex.error
+        raise newException(
+          CatchableError, "Failed to fetch next free index: " & nextFreeIndex.error
+        )
```
NagyZoltanPeter: It is hard to see in this diff, but why this section is in not below the condition of `if rootUpdated and g.membershipIndex.isSome():` as it was?

darshankabariya: Hi @NagyZoltanPeter ,
nextFreeIndex is simply a counter for the total number of RLN registrations inside the smart contract.
I bypassed the condition if `rootUpdated and g.membershipIndex.isSome()` bcz if node is not registered with RLN, it won’t be able to fetch the correct total registration count.
In that scenario, the condition becomes false, which is why we handle it outside the if block.

But your question sparked another thought — I don’t think we should keep this metric update here. We don’t need to check the total registration every 5 seconds, i find a better location these matrics update.

darshankabariya:
> LGTM! Are the failing CI tests expected/known issues?

It’s expected, I guess, but @Ivansete-status can confirm.

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
darshankabariya: Updated the code @NagyZoltanPeter — looks better now?

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
```diff
@@ -202,11 +202,13 @@ proc trackRootChanges*(g: OnchainGroupManager) {.async: (raises: [CatchableError
       let rootUpdated = await g.updateRoots()
 
       if rootUpdated:
-        let proofResult = await g.fetchMerkleProofElements()
-        if proofResult.isErr():
-          error "Failed to fetch Merkle proof", error = proofResult.error
-        else:
-          g.merkleProofCache = proofResult.get()
+        if g.membershipIndex.isSome():
+          # merkleProofElements is only needed for rln registered nodes
```
Ivansete-status: Maybe? I think we need to add some little explanation about each possible condition. Just adding some suggestions but feel free to bring a better description :D
```suggestion
        ## rootUpdated means that the membership set changed in the blockchain and therefore we need to update accordingly
        if g.membershipIndex.isSome():
          # Self node is considered to be part of the RLN membership set when there is a membership index
          # merkleProofElements is only needed for rln registered nodes
```

jm-clius (approved):
Thanks! Hopefully clears the error logs. :)

PR #3576, opened 2025-09-22
fixes noticed while dogfooding chatmix with TWN

chaitanyaprem:
## Description
Fixes for some issues noticed during dogfooding mix on TWN. 

## Changes

- [ ] Use only IPV4 address while passing mix nodes to node pool as currently mix only supports encoding ipv4 addresses in mix packet headers.
- [ ] update chat2 to work with TWN by default and some fixes in chat2
- [ ] fix simulation scripts after default config changes in chat2

## Issue
Reference issue https://github.com/waku-org/nwaku/issues/3280


waku/waku_mix/protocol.nim
```diff
@@ -61,18 +61,29 @@ proc appendPeerIdToMultiaddr*(multiaddr: MultiAddress, peerId: PeerId): MultiAdd
...
+    if multiaddr.contains(multiCodec("ip4")).get():
+      trace "found ipv4 multiaddr", addr = $multiaddr
+      return some(multiaddr)
+  trace "no ipv4 multiaddr found"
+  return none(MultiAddress)
+
 proc populateMixNodePool*(mix: WakuMix) =
   # populate only peers that i) are reachable ii) share cluster iii) support mix
-  let remotePeers = mix.peerManager.switch.peerStore.getReachablePeers().filterIt(
+  let remotePeers = mix.peerManager.switch.peerStore.peers().filterIt(
```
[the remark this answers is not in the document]
chaitanyaprem: This is fine as sender does not need to be reachable with all mix nodes in the generated mix path.

apps/chat2mix/chat2mix.nim
```diff
@@ -634,23 +618,40 @@ proc processInput(rfd: AsyncFD, rng: ref HmacDrbgContext) {.async.} =
...
-      node.peerManager.addServicePeer(peerInfo.value, WakuPeerExchangeCodec)
-      # Start maintaining subscription
-      asyncSpawn maintainSubscription(
-        node, pubsubTopic, conf.contentTopic, peerInfo.value, false
-      )
+  var servicePeerInfo: RemotePeerInfo
+  if conf.serviceNode != "": #TODO: use one of discovered nodes if not present.
+    let pInfo = parsePeerInfo(conf.serviceNode)
+    if pInfo.isOk():
+      servicePeerInfo = pInfo.value
```
Ivansete-status:
```suggestion
    servicePeerInfo = parsePeerInfo(conf.serviceNode).valueOr:
      error "LightPushClient not mounted. Couldn't parse conf.serviceNode",
        error = error
      quit(QuitFailure)
```

apps/chat2mix/chat2mix.nim
```diff
@@ -634,23 +618,40 @@ proc processInput(rfd: AsyncFD, rng: ref HmacDrbgContext) {.async.} =
...
+  if servicePeerInfo.isNil():
+    # Assuming that service node supports all services
+    let pInfo = selectRandomServicePeer(
+      node.peerManager, none(RemotePeerInfo), WakuLightpushCodec
+    )
+    if pInfo.isOk():
+      servicePeerInfo = pInfo.get()
+    else:
+      error "Couldn't find any service peer"
+      quit(QuitFailure)
```
Ivansete-status:
```suggestion
    servicePeerInfo = selectRandomServicePeer(
      node.peerManager, none(RemotePeerInfo), WakuLightpushCodec
    ).valueOr:
      error "Couldn't find any service peer"
      quit(QuitFailure)
```

apps/chat2mix/chat2mix.nim
```diff
@@ -634,23 +618,40 @@ proc processInput(rfd: AsyncFD, rng: ref HmacDrbgContext) {.async.} =
...
 
+  # Start maintaining subscription
+  asyncSpawn maintainSubscription(
+    node, pubsubTopic, conf.contentTopic, servicePeerInfo, false
+  )
+  while true:
+    if node.getMixNodePoolSize() >= 3:
+      break
+    discard await node.fetchPeerExchangePeers()
+    await sleepAsync(1000)
```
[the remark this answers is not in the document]
chaitanyaprem: Can't add retries since no point in proceeding without having peers discovered .Hence moving echo statement above the loop to indicate user that peers are being discovered.

apps/chat2mix/config_chat2mix.nim
```diff
@@ -94,20 +95,30 @@ type
...
         "Cluster id that the node is running in. Node in a different cluster id is disconnected.",
-      defaultValue: 2,
+      defaultValue: 1,
       name: "cluster-id"
     .}: uint16
 
     numShardsInNetwork* {.
       desc: "Number of shards in the network",
-      defaultValue: 1,
+      defaultValue: 8,
```
jm-clius, line 104: Here and elsewhere, would it make sense to allow `chat2` to use a `--preset` as in https://github.com/waku-org/nwaku/blob/c07537705ead30938344dae58713808c337af56e/waku/factory/external_config.nim#L158-L163?

chaitanyaprem: Oh yes, i did not know presets were already there. Will improve it to use presets where possible.

I am wondering if this `ExternalConfig` itself can be reused as is by various apps (maybe by providing it as an object) rather than each app copying all the config (which leads to maintenance nightmare). And each app can just extend this config to add their own specific configs.
wdyt @fryorcraken and @Ivansete-status

fryorcraken (commented):
use `results`, not `stew/results` (deprecated)

apps/chat2mix/config_chat2mix.nim
fryorcraken: check https://github.com/waku-org/nwaku/pull/3564 for my proposal re `external_config`.

This way, it can be re-used.

IMO, none of the approaches you listed are adequate, instead, I would suggest to have minimum CLI args and be opinionated on the options. 99% of external configs options are useless for you and your users when dogfooding mix

chaitanyaprem: Well, since chat2 is a edge/light node and i don't think we have presets defined yet for this. 
And once Franck's config changes are also merged, i will move this into the new form of CLI config.

Ivansete-status (approved):
LGTM! Thanks for it! :100:

PR #3580, opened 2025-09-25
feat: Waku API create node

fryorcraken:
## Description

Introduce `createNode`, part of the Waku API, as defined in https://github.com/waku-org/specs/pull/65.
## Changes

- new `createNode` function
- new `waku.nim` root file, developers are expected to use what is exported in there. Simplifies dev ex.
- moved `external_config` outside the `waku` (library), as it to be used by *binaries*.
- disabled websocket in default test config (port conflicts were happening in test and it's not useful for tests)
- new example in `examples/waku_api.nim` to show usage of API.

## Notes

Supersedes https://github.com/waku-org/nwaku/pull/3564 which also tried to make nwaku usable via nimble.

Follow-up PRs:
- use the new API in the FFI library
- check connectivity health
- send, subscribe, APIs, etc

## Issue

Related to https://github.com/waku-org/nwaku/issues/3493


examples/waku_api.nim
```diff
@@ -0,0 +1,38 @@
+import std/options
+import chronos, results, confutils, confutils/defs
+import waku
+
+type CliArgs* = object
```
Ivansete-status:
```suggestion
type CliArgs = object
```

examples/waku_api.nim
```diff
@@ -0,0 +1,38 @@
...
+  ethRpcEndpoint* {.
+    defaultValue: "", desc: "ETH RPC Endpoint, if passed, RLN is enabled"
+  .}: string
+
+when isMainModule:
+  let args = CliArgs.load()
+
+  echo("Starting Waku node...")
```
Ivansete-status: Oh, super nitpick, but I think we tend not using parenthesis
```suggestion
  echo "Starting Waku node"
```

waku/api/api_conf.nim
```diff
@@ -0,0 +1,211 @@
...
+type NetworkingConfig* {.requiresInit.} = object
+  listenIpv4*: string
+  p2pTcpPort*: uint16
+  discv5UdpPort*: uint16
+
+type MessageValidation* {.requiresInit.} = object
+  maxMessageSize*: string # Accepts formats like "150 KiB", "1500 B"
+  rlnConfig*: Option[RlnConfig]
+
+type WakuConfig* {.requiresInit.} = object
```
Ivansete-status, line 29: Notice that we have the [WakuConf](https://github.com/waku-org/nwaku/blob/dbecd721f9f86e8d05f1506126033fcfa206c084/waku/factory/waku_conf.nim#L66) type that might cause confusion. Therefore, maybe interesting rename the API config:
```suggestion
type WakuAPIConfig* {.requiresInit.} = object
```

waku/api/api_conf.nim
```diff
@@ -0,0 +1,211 @@
...
+    autoShardingConfig: AutoShardingConfig = DefaultAutoShardingConfig(),
+    messageValidation: MessageValidation = DefaultMessageValidation(),
+): WakuConfig =
+  return WakuConfig(
+    entryNodes: entryNodes,
+    staticStoreNodes: staticStoreNodes,
+    clusterId: clusterId,
+    autoShardingConfig: autoShardingConfig,
+    messageValidation: messageValidation,
+  )
```
Ivansete-status: I like the name that you suggest but it is not idiomatic.
For `object` types we need to expose their ctors in the following way

```suggestion
proc init*(T: typedesc[WakuConfig],
    entryNodes: seq[string],
    staticStoreNodes: seq[string] = @[],
    clusterId: uint16,
    autoShardingConfig: AutoShardingConfig = DefaultAutoShardingConfig(),
    messageValidation: MessageValidation = DefaultMessageValidation(),
): T =
  return T(
    entryNodes: entryNodes,
    staticStoreNodes: staticStoreNodes,
    clusterId: clusterId,
    autoShardingConfig: autoShardingConfig,
    messageValidation: messageValidation,
  )
```

Then:
```
let conf = WakuConfig.init(..)
```

Similarly, for `ref object` types we follow the same approach but name the proc as `new`.

waku/api/api_conf.nim
```diff
@@ -0,0 +1,211 @@
...
+  messageValidation: MessageValidation
+
+proc DefaultNetworkingConfig(): NetworkingConfig =
+  return NetworkingConfig(listenIpv4: "0.0.0.0", p2pTcpPort: 60000, discv5UdpPort: 9000)
+
+proc DefaultAutoShardingConfig(): AutoShardingConfig =
+  return AutoShardingConfig(numShardsInCluster: 1)
+
+proc DefaultMessageValidation(): MessageValidation =
+  return MessageValidation(maxMessageSize: "150 KiB", rlnConfig: none(RlnConfig))
```
Ivansete-status: I would try to use `const` attributes if possible
```suggestion
const
  DefaultNetworkingConfig = NetworkingConfig(listenIpv4: "0.0.0.0", p2pTcpPort: 60000, discv5UdpPort: 9000)
 DefaultAutoShardingConfig = AutoShardingConfig(numShardsInCluster: 1)
 DefaultMessageValidation = MessageValidation(maxMessageSize: "150 KiB", rlnConfig: none(RlnConfig))
```

waku/api/api_conf.nim
```diff
@@ -0,0 +1,211 @@
...
+          chainId: 59141,
+          epochSizeSec: 600, # 10 minutes
+        )
+      ),
+    ),
+  )
+
+type WakuMode* = enum
+  Edge = "edge"
+  Core = "core"
```
Ivansete-status: When using `pure` enums, the conversion to $ is automatic and it enforces the use of WakuNode.Edge, for example
```suggestion
type WakuMode* {.pure.} = enum
  Edge
  Core
```

waku/api/api_conf.nim
```diff
@@ -0,0 +1,211 @@
...
+
+type NodeConfig* {.requiresInit.} = object
+  mode: WakuMode
+  wakuConfig: WakuConfig
+  messageConfirmation: bool
+  networkingConfig: NetworkingConfig
+  ethRpcEndpoints: seq[string]
+
+proc newNodeConfig*(
```
Ivansete-status:
```suggestion
proc init*(T: typedesc[NodeConfig]
```

Besides, what if we avoid creating `NodeConfig` type and just use `WakuAPIConfig` for now?

waku/api/api_conf.nim
```diff
@@ -0,0 +1,211 @@
...
+  p2pTcpPort*: uint16
+  discv5UdpPort*: uint16
+
+type MessageValidation* {.requiresInit.} = object
+  maxMessageSize*: string # Accepts formats like "150 KiB", "1500 B"
+  rlnConfig*: Option[RlnConfig]
+
+type WakuConfig* {.requiresInit.} = object
+  entryNodes: seq[string]
+  staticStoreNodes: seq[string]
```
Ivansete-status, line 31: Maybe?
```suggestion
  entryNodes: seq[MultiAddress]
  staticStoreNodes: seq[MultiAddress]
```

waku/api/api_conf.nim
```diff
@@ -0,0 +1,211 @@
...
+  mode: WakuMode
+  wakuConfig: WakuConfig
+  messageConfirmation: bool
+  networkingConfig: NetworkingConfig
+  ethRpcEndpoints: seq[string]
+
+proc newNodeConfig*(
+    mode: WakuMode = WakuMode.Core,
+    wakuConfig: WakuConfig = TheWakuNetworkPreset(),
+    messageConfirmation: bool = true,
```
Ivansete-status: This comment might fit better in the spec PR but, why I should pick `false`? I think I'd always be interested in `true`

waku/api/entry_nodes.nim
```diff
@@ -0,0 +1,77 @@
+import std/strutils
+
+import results, eth/p2p/discoveryv5/enr
+
+import waku/waku_core/peers
+
+type EntryNodeType* = enum
+  EntryNodeEnrTree
+  EntryNodeEnr
+  EntryNodeMultiaddr
```
Ivansete-status: I think that is a bit more idiomatic:
```suggestion
type EntryNodeType*{.pure.} = enum
  EnrTree
  Enr
  Multiaddr
```

Then, to use it:
```
EntryNodeType.EnrTree
```

waku/api/api_conf.nim
fryorcraken: I agree with you but I want to keep things "nice" for our users.

Do you have a suggestion to rename `WakuConf` instead?

waku/api/api_conf.nim
fryorcraken: Good comment.

I was wondering if from an API point of view, we don't prefer to allow `string` type to be passed here. From a dev PoV, it may be slightly easier to use as they can use `"edge"` instead of having to use the type `WakuMode.Edge`. WDYT?

waku/api/api_conf.nim
fryorcraken: There may be a resource consumption question but I agree actually, we could just remove this parameter and optimize later.

waku/api/api_conf.nim
fryorcraken: they can accept ENRs and enrtree too.

waku/api/api_conf.nim
fryorcraken: Not sure I get it. I don't think we can avoid that due to the fact that people may want to disable RLN for local dev etc. see

```
  let config =
    if (args.ethRpcEndpoint == ""):
      # Create a basic configuration for the Waku node
      # No RLN as we don't have an ETH RPC Endpoint
      newNodeConfig(wakuConfig = WakuConfig.init(entryNodes = @[], clusterId = 42))
    else:
      # Connect to TWN, use ETH RPC Endpoint for RLN
      newNodeConfig(ethRpcEndpoints = @[args.ethRpcEndpoint])
```

waku/api/api_conf.nim
fryorcraken: This is what we do with js-waku, not sure it applies as well in Nim.

fryorcraken: actually I just tried and I am not sure I understand when it's **not** pure.

waku/api/api_conf.nim
NagyZoltanPeter: Hm, I would argue against such wording as `...APIConfig`, sounds controversial for me, as I would never config an API but config via API.
Imo, the inner WakuConf can be seen more as NodeConf, and might keep this one for the outer world of devs named WakuConfig...

examples/waku_example.nim
NagyZoltanPeter, line 1: Naming of the file of an example app as `waku_api.nim` can be misleading. Maybe `demo_api.nim` or `api_demo.nim` should describe better.  WDYT?

waku.nimble
```diff
@@ -140,6 +140,7 @@ task testwakunode2, "Build & run wakunode2 app tests":
   test "all_tests_wakunode2"
 
 task example2, "Build Waku examples":
+  buildBinary "waku_api", "examples/"
```
NagyZoltanPeter:
```suggestion
  buildBinary "api_demo", "examples/"
```
Maybe a better naming?

waku/api/api_conf.nim
```diff
@@ -0,0 +1,203 @@
...
+    # Metadata is always mounted
+
+    b.filterServiceConf.withEnabled(true)
+    b.filterServiceConf.withMaxPeersToServe(20)
+
+    b.withLightPush(true)
+
+    b.discv5Conf.withEnabled(true)
+    b.withPeerExchange(true)
+    b.withRendezvous(true)
```
NagyZoltanPeter, line 128: I bet these are going to be configurable in the future. Are they?

waku/api/api_conf.nim
NagyZoltanPeter: Ok, so we have NodeConfig too... lost in configs...

examples/waku_example.nim
fryorcraken: I agree. I'll keep it simple because ultimately, the file will contain all API calls and should be "the" reference to use Waku. I might just change to `waku_demo.nim` and while incomplete now, we can further expand it as we implement the other verbs of the Waku API.

waku/api/api_conf.nim
fryorcraken: Not really. The intent is to remove gunfoot, and choice, from the developer. There is no reason why you would remove a peer discovery strategy. We may fine or expose some parameters if there are issues around resource consumptions. But we should provide a "it just works" experience. Which does involve having several peer discovery mechanisms enabled.

waku/api/api_conf.nim
fryorcraken: I think the spec https://github.com/waku-org/specs/pull/65/files is the best place to discuss this.
Not an easy choice and I had enough back and forth with @Ivansete-status and @jm-clius to settle on this:

- `node`: this is your whole binary, so it includes all parameters, including things like your tcp port
- `waku`: theses are waku protocol related parameters, such as message validation, and cluster id. Hence it's a sub object of `node`.

An alternative name could be `ProtocolsConfig`

fryorcraken: I'll open a follow-up PR to change to protocol...

PR #3587, opened 2025-09-30
remove mixPubKey from ENR and add config  to pass static mix nodes 

chaitanyaprem:
## Description
On deploying waku-mix on fleet nodes ENR size limit was hit and nodes were not able to start because of it. 
This could be due to the fact that mixPubKey is also included in ENR which is already close to its size limit because of multiaddresses including domain name. 

In order to get around this issue, for now we plan to not include mixPubKey in ENR and provide an option to statically specify mix nodes in case anyone wants to publish using mix. 

## Changes

- [x] removed setting mixPubKey in ENR
- [x] removed related ENR code 
- [x] static config to specify mixnodes mulitaddr and mixPubKey
- [x] updated examples
- [x] updated simulation scripts
- [x] add check to allow only ipv4 addresses in static config
- [x] updated debug/info REST API to also include the mixPubKey if present
- [x] updated health REST API to include mix protocol health


apps/chat2mix/config_chat2mix.nim
```diff
@@ -82,8 +83,10 @@ type
     .}: seq[string]
 
     mixnodes* {.
-      desc: "Peer ENR to add as a mixnode. Argument may be repeated.", name: "mixnode"
-    .}: seq[string]
+      desc:
+        "Multiaddress and mix-key of mix node to be statically specified in format multiaddr:mixPubKey. Argument may be repeated.",
+      name: "mixnodes"
```
jm-clius:
```suggestion
      name: "mixnode"
```

The `name` property of a repeated config item should be singular so that the cmd line argument can be specified in the singular - a collection of these `--mixnode:` arguments then become the `seq` with the plural handle `mixnodes`. See https://github.com/waku-org/nwaku/blob/8aa87f19e84e32ef5452fbf8f74ae534222305aa/apps/chat2mix/config_chat2mix.nim#L80-L83 for another example.

examples/lightpush_mix/lightpush_publisher_mix_config.nim
```diff
@@ -26,3 +28,20 @@ type LightPushMixConf* = object
   mixDisabled* {.
     desc: "Do not use mix for publishing.", defaultValue: false, name: "without-mix"
   .}: bool
+
+  mixnodes* {.
+    desc:
+      "Multiaddress and mix-key of mix node to be statically specified in format multiaddr:mixPubKey. Argument may be repeated.",
+    name: "mixnodes"
```
jm-clius:
```suggestion
    name: "mixnode"
```

See reasoning elsewhere.

waku/factory/external_config.nim
```diff
@@ -616,6 +618,12 @@ with the drawback of consuming some more bandwidth.""",
       name: "mixkey"
     .}: Option[string]
 
+    mixnodes* {.
+      desc:
+        "Multiaddress and mix-key of mix node to be statically specified in format multiaddr:mixPubKey. Argument may be repeated.",
+      name: "mixnodes"
```
jm-clius:
```suggestion
      name: "mixnode"
```

See reasoning elsewhere.

waku/waku_mix/protocol.nim
```diff
@@ -27,9 +27,14 @@ type
     peerManager*: PeerManager
     clusterId: uint16
     nodePoolLoopHandle: Future[void]
+    bootstrapMixNodes*: Table[PeerId, MixPubInfo]
```
jm-clius: I'm not sure about this pattern of the mix protocol tracking these bootstrap nodes in a separate state. My idea of how it should work is simply that when the `WakuNode` gets built in the factory, it adds the static `mixnodes` to the peer manager. Similar to how statically configured filter nodes are added: https://github.com/waku-org/nwaku/blob/5118fc3f6dc587003240a31687a601534f9249d1/waku/factory/node_factory.nim#L406
The mix protocol continue using the `peerManager` to find suitable peers as before with no added state.

chaitanyaprem: > I'm not sure about this pattern of the mix protocol tracking these bootstrap nodes in a separate state

ah, this looks like a remnant of what i initially thought of and later changed my mind. no longer needed, will remove it. 

But we can use logic of peerManager because we also require mixPubKey which has to be again stored separately, instead as of now if you see we just pass list of nodes while [initializing](https://github.com/waku-org/nwaku/blob/8aa87f19e84e32ef5452fbf8f74ae534222305aa/waku/waku_mix/protocol.nim#L125) the mix protocol itself that are read from config. 

This way we are not adding any state, but just setting an initial config.

jm-clius: Ok, I do see your point! I was glossing over the `mixPubKey`.

This idea might indeed extend beyond the scope of a single PR, but I'm now wondering if a better solution wouldn't be to create a new `MixPubKeyBook` in the peer store similar to the `PeerBook`s we have here:

https://github.com/waku-org/nwaku/blob/5118fc3f6dc587003240a31687a601534f9249d1/waku/node/peer_manager/waku_peer_store.nim#L20-L40

This way we can track the mix pub key no matter the discovery method or having to introduce discovery-specific knowledge into the mix protocol. Perhaps it would also mean that the mix protocol wouldn't need to maintain a mix pool at all - this information is simply implicit in the peer store?

There may be flaws to this approach, but I'm trying to find a way to prevent mix from bloating with peerstore-like logic.

chaitanyaprem: That idea crossed my mind as well, but since mix-protocol will anyways have to be aware of the nodepool i.e multiaddr and publicKey of all mix nodes we discover as of now it is fine to pass it as initialization. Maybe this is something that can be changed in mix protocol functionality itself i.e to get mix node info from peerStore by adding this in PeerBook . 

Might be a good idea if this can be addressed at nim-libp2p level. 
wdyt @richard-ramos

examples/lightpush_mix/lightpush_publisher_mix_config.nim
```diff
@@ -26,3 +33,25 @@ type LightPushMixConf* = object
...
     desc: "Do not use mix for publishing.", defaultValue: false, name: "without-mix"
   .}: bool
+
+  mixnodes* {.
+    desc:
+      "Multiaddress and mix-key of mix node to be statically specified in format multiaddr:mixPubKey. Argument may be repeated.",
+    name: "mixnode"
+  .}: seq[MixNodePubInfo]
+
+proc parseCmdArg*(T: type MixNodePubInfo, p: string): T =
```
Ivansete-status: Apparently the following does the same but is a bit more idiomatic
```suggestion
proc parseCmdArg*(T: typedesc[MixNodePubInfo], p: string): T =
```

waku/factory/conf_builder/mix_conf_builder.nim
```diff
@@ -1,6 +1,6 @@
 import chronicles, std/options, results
 import libp2p/crypto/crypto, libp2p/crypto/curve25519, mix/curve25519
-import ../waku_conf
+import ../waku_conf, ../../waku_mix
```
Ivansete-status: nitpick:
```suggestion
import ../waku_conf, waku/waku_mix
```

tools/confutils/cli_args.nim
```diff
@@ -616,6 +619,12 @@ with the drawback of consuming some more bandwidth.""",
       name: "mixkey"
     .}: Option[string]
 
+    mixnodes* {.
```
Ivansete-status, line 621: I wonder if we need to set a default empty seq to make this param optional. Maybe not needed to do that explicitly.

waku/waku_mix/protocol.nim
```diff
@@ -109,43 +112,26 @@ proc startMixNodePoolMgr*(mix: WakuMix) {.async.} =
...
-  let bootNodesMixPubKeys = ["9d09ce624f76e8f606265edb9cca2b7de9b41772a6d784bddaf92ffa8fba7d2c",
-                             "9231e86da6432502900a84f867004ce78632ab52cd8e30b1ec322cd795710c2a",
-                             "275cd6889e1f29ca48e5b9edb800d1a94f49f13d393a0ecf1a07af753506de6c",
-                             "e0ed594a8d506681be075e8e23723478388fb182477f7a469309a25e7076fc18",
-                             "8fd7a1a7c19b403d231452a9b1ea40eb1cc76f455d918ef8980e7685f9eeeb1f"
-                             ]
-  for index, mixNodeMultiaddr in bootNodesMultiaddrs:
-    let peerIdRes = getPeerIdFromMultiAddr(mixNodeMultiaddr)
+  for node in bootnodes:
+    let peerIdRes = getPeerIdFromMultiAddr(node.multiAddr)
```
Ivansete-status: If possible, I think is cleaner to use `valueOr` approach :)

waku/waku_mix/protocol.nim
```diff
@@ -154,24 +140,23 @@ proc new*(
     nodeAddr, mixPubKey, mixPrivKey, peermgr.switch.peerInfo.publicKey.skkey,
     peermgr.switch.peerInfo.privateKey.skkey,
   )
-
-  # TODO : ideally mix should not be marked ready until certain min pool of mixNodes are discovered  
+  if bootnodes.len < 3:
+    warn "publishing with mix won't work as there are less than 3 mix nodes in node pool"
```
Ivansete-status: I think it is more appropriate to perform such validation, count # mix peers, at `publish` time, instead of at `new` time.
Besides, I suggest put the 3 in a `const` to give it some more meaning

waku/waku_mix/protocol.nim
richard-ramos: Sounds like a nice idea indeed...  the constructor of mixprotocol would then only need the switch and not  have the need for the list of `MixPubInfo`. I guess the same should apply to the `MixNodeInfo`, since the switch likely contains all the required information.

My only concern is wrt how to determine which multiaddress to use for mix? Currently mix only supports a single address, and the peerstore can contain multiple

cc: @AkshayaMani

tools/confutils/cli_args.nim
chaitanyaprem: I was assuming default would be empty seq automatically.

waku/waku_mix/protocol.nim
chaitanyaprem: > I think it is more appropriate to perform such validation, count # mix peers, at `publish` time, instead of at `new` time.

As we are not having discovery of mix nodes, for now it makes sense to keep it at protocol initialization. Once discovery is sorted, this check would be removed and is handled at publish time by underlying mix protocol.

waku/waku_mix/protocol.nim
chaitanyaprem: > I guess the same should apply to the `MixNodeInfo`, since the switch likely contains all the required information.

Is it possible? because the mixKeys for the localNode are not stored in the switch. Unless this is changed while merging into libp2p.

chaitanyaprem:
@jm-clius can you re-review and let me know if any other changes do you need?

waku/waku_mix/protocol.nim
Ivansete-status: Okay perfect!
Kindly also consider that nitpick comment ;P
`I suggest put the 3 in a const to give it some more meaning`

waku/waku_mix/protocol.nim
```diff
@@ -110,42 +115,24 @@ proc startMixNodePoolMgr*(mix: WakuMix) {.async.} =
...
-       error "Failed to get peer id from multiaddress: " , error = peerIdRes.error
-    let peerId = peerIdRes.get()
-    #if (not peerID == nil) and peerID == exceptPeerID:
-    #  continue
-    let mixNodePubInfo = createMixPubInfo(mixNodeMultiaddr, intoCurve25519Key(ncrutils.fromHex(bootNodesMixPubKeys[index])))
-
-    mixNodes[peerId] = mixNodePubInfo
+  for node in bootnodes:
+    let peerId = getPeerIdFromMultiAddr(node.multiAddr).valueOr:
+      error "Failed to get peer id from multiaddress: ", error = error
```
```suggestion
      error "Failed to get peer id from multiaddress", error = error, multiAddr = $node.multiAddr
```

waku/node/waku_node.nim
```diff
@@ -201,7 +202,11 @@ proc info*(node: WakuNode): WakuInfo =
     var fulladdr = $address & "/p2p/" & $peerInfo.peerId
     listenStr &= fulladdr
   let enrUri = node.enr.toUri()
-  let wakuInfo = WakuInfo(listenAddresses: listenStr, enrUri: enrUri)
+  var wakuInfo = WakuInfo(listenAddresses: listenStr, enrUri: enrUri)
+  if node.wakuMix != nil:
```
```suggestion
  if not node.wakuMix.isNil():
```

waku/node/health_monitor/node_health_monitor.nim
```diff
@@ -233,6 +233,15 @@ proc getRendezvousHealth(hm: NodeHealthMonitor): ProtocolHealth =
 
   return p.ready()
 
+proc getMixHealth(hm: NodeHealthMonitor): ProtocolHealth =
+  var p = ProtocolHealth.init("Mix")
+  checkWakuNodeNotNil(hm.node, p)
+
+  if hm.node.wakuMix == nil:
```
Ivansete-status:
```suggestion
  if hm.node.wakuMix.isNil():
```

jm-clius (approved):
Thanks!

PR #3603, opened 2025-10-01
chore: rename Waku API's "Waku Config" to "Protocols" Config

fryorcraken:
Make it clearer that with this config, we are configuring the Waku protocols, in contrast to other parameters which are more executable related.

Related to https://github.com/waku-org/nwaku/issues/3493


tests/api/test_node_conf.nim
```diff
@@ -23,7 +23,7 @@ suite "LibWaku Conf - toWakuConf":
 
   test "Core mode configuration":
     ## Given
-    let wakuConfig = WakuConfig.init(entryNodes = @[], clusterId = 1)
+    let wakuConfig = ProtocolsConfig.init(entryNodes = @[], clusterId = 1)
```
fryorcraken: var name must change too.

NagyZoltanPeter (approved):
Great, thank you for it.

Issue #1969, opened 2023-08-30
chore(refactoring): apply isOkOr || valueOr approach


Ivansete-status:
### Background
In order to simplify the `nwaku` code base, is interesting to reduce the number of lines as much as possible.

### Details
During a `libwaku` PR, @arnetheduck suggested an enhancement that can be extended to the whole `nwaku` repo.

https://github.com/waku-org/nwaku/pull/1957#discussion_r1310001945

We very often use the next approach:
```
let valueRes = item.doSth()
if valueRes.isErr():
  echo "Error: " & $valueRes.error

let v = valueRes.get()
```
which can be translated to:
```
let v = item.doSth.isOkOr:
  echo "Error: " & $error
```

### Acceptance criteria
Do the same with fewer lines.

SionoiS:
This is amazing! YES Please.

Ivansete-status:
For now, let's skip "vendor, tests" from the look up and only consider "isOk" and "isErr" changes in the rest of nwaku

PR #3610, opened 2025-10-02
chore(refactoring): replace some `isOk` usage with better alternatives

fcecin:
## Description

This PR refactors various patterns that use `isOk` across nwaku with less verbose and more idiomatic alternatives.

## Changes

This PR changes varied usage of `isOk` into better patterns with e.g. `valueOr` , `isOkOr`, `isErrOr`, etc. There are a few other non-isOk, miscellaneous refactors included.

There should not be any behavior (semantics) changes in this PR.

## Issue

This PR will address `isOk` refactors for #1969. A follow-up PR will address `isErr` refactors separately to merge less code at a time and minimize potential conflicts.

### DIFF
--- a/apps/benchmarks/benchmarks.nim
+++ b/apps/benchmarks/benchmarks.nim
@@ -28,10 +28,9 @@ proc benchmark(
       iter = i, elapsed_ms = (getTime() - start_time).inMilliseconds
 
   discard await manager.updateRoots()
-  let proofResult = await manager.fetchMerkleProofElements()
-  if proofResult.isErr():
-    error "Failed to fetch Merkle proof", error = proofResult.error
-  manager.merkleProofCache = proofResult.get()
+  manager.merkleProofCache = (await manager.fetchMerkleProofElements()).valueOr:
+    error "Failed to fetch Merkle proof", error = error
+    quit(QuitFailure)
 
   let epoch = default(Epoch)
   debug "epoch in bytes", epochHex = epoch.inHex()
--- a/apps/chat2/chat2.nim
+++ b/apps/chat2/chat2.nim
@@ -132,25 +132,14 @@ proc showChatPrompt(c: Chat) =
     except IOError:
       discard
 
-proc getChatLine(c: Chat, msg: WakuMessage): Result[string, string] =
+proc getChatLine(payload: seq[byte]): string =
   # No payload encoding/encryption from Waku
-  let
-    pb = Chat2Message.init(msg.payload)
-    chatLine =
-      if pb.isOk:
-        pb[].toString()
-      else:
-        string.fromBytes(msg.payload)
-  return ok(chatline)
+  let pb = Chat2Message.init(payload).valueOr:
+    return string.fromBytes(payload)
+  return $pb
 
 proc printReceivedMessage(c: Chat, msg: WakuMessage) =
-  let
-    pb = Chat2Message.init(msg.payload)
-    chatLine =
-      if pb.isOk:
-        pb[].toString()
-      else:
-        string.fromBytes(msg.payload)
+  let chatLine = getChatLine(msg.payload)
   try:
     echo &"{chatLine}"
   except ValueError:
@@ -173,18 +162,16 @@ proc startMetricsServer(
 ): Result[MetricsHttpServerRef, string] =
   info "Starting metrics HTTP server", serverIp = $serverIp, serverPort = $serverPort
 
-  let metricsServerRes = MetricsHttpServerRef.new($serverIp, serverPort)
-  if metricsServerRes.isErr():
-    return err("metrics HTTP server start failed: " & $metricsServerRes.error)
+  let server = MetricsHttpServerRef.new($serverIp, serverPort).valueOr:
+    return err("metrics HTTP server start failed: " & $error)
 
-  let server = metricsServerRes.value
   try:
     waitFor server.start()
   except CatchableError:
     return err("metrics HTTP server start failed: " & getCurrentExceptionMsg())
 
   info "Metrics HTTP server started", serverIp = $serverIp, serverPort = $serverPort
-  ok(metricsServerRes.value)
+  ok(server)
 
 proc publish(c: Chat, line: string) =
   # First create a Chat2Message protobuf with this line of text
@@ -202,19 +189,17 @@ proc publish(c: Chat, line: string) =
     version: 0,
     timestamp: getNanosecondTime(time),
   )
+
   if not isNil(c.node.wakuRlnRelay):
     # for future version when we support more than one rln protected content topic,
     # we should check the message content topic as well
-    let appendRes = c.node.wakuRlnRelay.appendRLNProof(message, float64(time))
-    if appendRes.isErr():
+    if c.node.wakuRlnRelay.appendRLNProof(message, float64(time)).isErr():
       debug "could not append rate limit proof to the message"
     else:
       debug "rate limit proof is appended to the message"
-      let decodeRes = RateLimitProof.init(message.proof)
-      if decodeRes.isErr():
+      let proof = RateLimitProof.init(message.proof).valueOr:
         error "could not decode the RLN proof"
-
-      let proof = decodeRes.get()
+        return
       # TODO move it to log after dogfooding
       let msgEpoch = fromEpoch(proof.epoch)
       if fromEpoch(c.node.wakuRlnRelay.lastEpoch) == msgEpoch:
@@ -438,7 +423,7 @@ proc processInput(rfd: AsyncFD, rng: ref HmacDrbgContext) {.async.} =
       let resolved = await dnsResolver.resolveTxt(domain)
       return resolved[0] # Use only first answer
 
-    var wakuDnsDiscovery = WakuDnsDiscovery.init(dnsDiscoveryUrl.get(), resolver)
+    let wakuDnsDiscovery = WakuDnsDiscovery.init(dnsDiscoveryUrl.get(), resolver)
     if wakuDnsDiscovery.isOk:
       let discoveredPeers = await wakuDnsDiscovery.get().findPeers()
       if discoveredPeers.isOk:
@@ -446,8 +431,10 @@ proc processInput(rfd: AsyncFD, rng: ref HmacDrbgContext) {.async.} =
         discoveredNodes = discoveredPeers.get()
         echo "Discovered and connecting to " & $discoveredNodes
         waitFor chat.node.connectToNodes(discoveredNodes)
+      else:
+        warn "Failed to find peers via DNS discovery", error = discoveredPeers.error
     else:
-      warn "Failed to init Waku DNS discovery"
+      warn "Failed to init Waku DNS discovery", error = wakuDnsDiscovery.error
 
   let peerInfo = node.switch.peerInfo
   let listenStr = $peerInfo.addrs[0] & "/p2p/" & $peerInfo.peerId
@@ -483,21 +470,19 @@ proc processInput(rfd: AsyncFD, rng: ref HmacDrbgContext) {.async.} =
             else:
               newSeq[byte](0)
 
-          let
-            pb = Chat2Message.init(payload)
-            chatLine =
-              if pb.isOk:
-                pb[].toString()
-              else:
-                string.fromBytes(payload)
+          let chatLine = getChatLine(payload)
           echo &"{chatLine}"
         info "Hit store handler"
 
-      let queryRes = await node.query(
-        StoreQueryRequest(contentTopics: @[chat.contentTopic]), storenode.get()
-      )
-      if queryRes.isOk():
-        storeHandler(queryRes.value)
+      block storeQueryBlock:
+        let queryRes = (
+          await node.query(
+            StoreQueryRequest(contentTopics: @[chat.contentTopic]), storenode.get()
+          )
+        ).valueOr:
+          error "Store query failed", error = error
+          break storeQueryBlock
+        storeHandler(queryRes)
 
   # NOTE Must be mounted after relay
   if conf.lightpushnode != "":
@@ -511,8 +496,9 @@ proc processInput(rfd: AsyncFD, rng: ref HmacDrbgContext) {.async.} =
         error = peerInfo.error
 
   if conf.filternode != "":
-    let peerInfo = parsePeerInfo(conf.filternode)
-    if peerInfo.isOk():
+    if (let peerInfo = parsePeerInfo(conf.filternode); peerInfo.isErr()):
+      error "Filter not mounted. Couldn't parse conf.filternode", error = peerInfo.error
+    else:
       await node.mountFilter()
       await node.mountFilterClient()
 
@@ -523,8 +509,6 @@ proc processInput(rfd: AsyncFD, rng: ref HmacDrbgContext) {.async.} =
         chat.printReceivedMessage(msg)
 
       # TODO: Here to support FilterV2 relevant subscription.
-    else:
-      error "Filter not mounted. Couldn't parse conf.filternode", error = peerInfo.error
 
   # Subscribe to a topic, if relay is mounted
   if conf.relay:
@@ -545,11 +529,8 @@ proc processInput(rfd: AsyncFD, rng: ref HmacDrbgContext) {.async.} =
 
       proc spamHandler(wakuMessage: WakuMessage) {.gcsafe, closure.} =
         debug "spam handler is called"
-        let chatLineResult = chat.getChatLine(wakuMessage)
-        if chatLineResult.isOk():
-          echo "A spam message is found and discarded : ", chatLineResult.value
-        else:
-          echo "A spam message is found and discarded"
+        let chatLineResult = getChatLine(wakuMessage.payload)
+        echo "spam message is found and discarded : " & chatLineResult
         chat.prompt = false
         showChatPrompt(chat)
 
--- a/apps/chat2bridge/chat2bridge.nim
+++ b/apps/chat2bridge/chat2bridge.nim
@@ -136,9 +136,7 @@ proc toMatterbridge(
 
 proc pollMatterbridge(cmb: Chat2MatterBridge, handler: MbMessageHandler) {.async.} =
   while cmb.running:
-    let getRes = cmb.mbClient.getMessages()
-
-    if getRes.isOk():
+    if (let getRes = cmb.mbClient.getMessages(); getRes.isOk()):
       for jsonNode in getRes[]:
         await handler(jsonNode)
     else:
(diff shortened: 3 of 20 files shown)


apps/chat2/chat2.nim
```diff
@@ -132,25 +132,14 @@ proc showChatPrompt(c: Chat) =
     except IOError:
       discard
 
-proc getChatLine(c: Chat, msg: WakuMessage): Result[string, string] =
+proc getChatLine(payload: seq[byte]): Result[string, string] =
```
Ivansete-status: Given that always returns `ok` maybe it can just return a `string`

apps/chat2/chat2.nim
```diff
@@ -132,25 +132,14 @@ proc showChatPrompt(c: Chat) =
...
-    pb = Chat2Message.init(msg.payload)
-    chatLine =
-      if pb.isOk:
-        pb[].toString()
-      else:
-        string.fromBytes(msg.payload)
-  return ok(chatline)
+  let pb = Chat2Message.init(payload).valueOr:
+    return ok(string.fromBytes(payload))
+  return ok(pb.toString())
```
Ivansete-status: Even though `toString` was already used. I think is more idiomatic to use `$`, if possible

apps/chat2/chat2.nim
```diff
@@ -205,16 +192,13 @@ proc publish(c: Chat, line: string) =
   if not isNil(c.node.wakuRlnRelay):
     # for future version when we support more than one rln protected content topic,
     # we should check the message content topic as well
-    let appendRes = c.node.wakuRlnRelay.appendRLNProof(message, float64(time))
-    if appendRes.isErr():
+    if c.node.wakuRlnRelay.appendRLNProof(message, float64(time)).isErr():
```
Ivansete-status, line 196: This particular snippet can be migrated to `isOkOr`

apps/chat2/chat2.nim
```diff
@@ -483,20 +468,16 @@ proc processInput(rfd: AsyncFD, rng: ref HmacDrbgContext) {.async.} =
...
-      let queryRes = await node.query(
-        StoreQueryRequest(contentTopics: @[chat.contentTopic]), storenode.get()
-      )
-      if queryRes.isOk():
+      if (
+        let queryRes = await node.query(
+          StoreQueryRequest(contentTopics: @[chat.contentTopic]), storenode.get()
+        )
+        queryRes.isOk()
+      ):
```
Ivansete-status: This can be converted to something like the following:
```suggestion
      let query = (await node.query(
        StoreQueryRequest(contentTopics: @[chat.contentTopic]), storenode.get()
      )).valuOr:
        error "Store query failed", error = $error
        return ## or break :)

      storeHandler(query)
```

apps/chat2/chat2.nim
```diff
@@ -545,11 +525,8 @@ proc processInput(rfd: AsyncFD, rng: ref HmacDrbgContext) {.async.} =
 
       proc spamHandler(wakuMessage: WakuMessage) {.gcsafe, closure.} =
         debug "spam handler is called"
-        let chatLineResult = chat.getChatLine(wakuMessage)
-        if chatLineResult.isOk():
-          echo "A spam message is found and discarded : ", chatLineResult.value
-        else:
-          echo "A spam message is found and discarded"
+        let chatLineResult = getChatLine(wakuMessage.payload).valueOr("<ERROR>")
```
Ivansete-status: Usually is more handy to have a more unique error message so that it is easier to look up in the code
```suggestion
        let chatLineResult = getChatLine(wakuMessage.payload).valueOr("<ERROR-could-not-get-chat-line>")
```

apps/chat2bridge/chat2bridge.nim
```diff
@@ -136,9 +136,7 @@ proc toMatterbridge(
 
 proc pollMatterbridge(cmb: Chat2MatterBridge, handler: MbMessageHandler) {.async.} =
   while cmb.running:
-    let getRes = cmb.mbClient.getMessages()
-
-    if getRes.isOk():
+    if (let getRes = cmb.mbClient.getMessages(); getRes.isOk()):
```
Ivansete-status, line 139: This whole block can be converted to the following:
```nim
    let msg = cmb.mbClient.getMessages().valueOr:
      error "Matterbridge host unreachable. Sleeping before retrying."
      await sleepAsync(chronos.seconds(10))

    for jsonNode in msg:
        await handler(jsonNode)
    
    await sleepAsync(cmb.pollPeriod)
 ```

File: apps/liteprotocoltester/publisher.nim [hunk 7]
```diff
@@ -193,22 +193,21 @@ proc publishMessages(
...
-            info "New service peer in use",
-              codec = lightpushPubsubTopic,
-              peer = constructMultiaddrStr(actualServicePeer)
-
-            noFailedPush = 0
-            noOfServicePeerSwitches += 1
-            lpt_change_service_peer_count.inc(labelValues = ["publisher"])
-            continue # try again with new peer without delay
-          else:
+          peerOpt.isOkOr:
```
Ivansete-status: Maybe? :)
```suggestion
          actualServicePeer = peerOpt.valueOr:
```
The same suggestion also applies to `apps/liteprotocoltester/receiver.nim`

apps/networkmonitor/networkmonitor.nim
```diff
@@ -374,14 +372,9 @@ proc retrieveDynamicBootstrapNodes(
...
-    var wakuDnsDiscovery = WakuDnsDiscovery.init(dnsDiscoveryUrl, resolver)
-    if wakuDnsDiscovery.isOk():
-      return (await wakuDnsDiscovery.get().findPeers()).mapErr(
-        proc(e: cstring): string =
-          $e
-      )
-    else:
-      warn "Failed to init Waku DNS discovery"
+    var wakuDnsDiscovery = WakuDnsDiscovery.init(dnsDiscoveryUrl, resolver).errorOr:
+      return (await value.findPeers()).mapErr(e => $e)
```
Ivansete-status, line 376: This is very awesome :heart:

apps/networkmonitor/networkmonitor.nim
```diff
@@ -391,11 +384,10 @@ proc getBootstrapFromDiscDns(
...
   try:
     let dnsNameServers = @[parseIpAddress("1.1.1.1"), parseIpAddress("1.0.0.1")]
-    let dynamicBootstrapNodesRes =
+    let dynamicBootstrapNodes = (
       await retrieveDynamicBootstrapNodes(conf.dnsDiscoveryUrl, dnsNameServers)
-    if not dynamicBootstrapNodesRes.isOk():
-      error("failed discovering peers from DNS")
-    let dynamicBootstrapNodes = dynamicBootstrapNodesRes.get()
+    ).valueOr:
+      raise newException(ValueError, $error)
```
Ivansete-status: Something important is that we try to avoid raising new exceptions as much as possible. There are few exceptions though, such as the RLN part, or REST, were exceptions are used (but I'd like them to disapear.) The preferable way of handling errors is by chainning errors to upper layers
```suggestion
      return err("Failed retrieving dynamic bootstrap nodes: " & $error)
```

waku/discovery/waku_dnsdisc.nim
```diff
@@ -67,14 +67,11 @@ proc findPeers*(
...
   for enr in discoveredEnr:
     # Convert discovered ENR to RemotePeerInfo and add to discovered nodes
-    let res = enr.toRemotePeerInfo()
-
-    if res.isOk():
-      discoveredNodes.add(res.get())
-    else:
-      error "Failed to convert ENR to peer info", enr = $enr, err = res.error()
+    let peerInfo = enr.toRemotePeerInfo().valueOr:
+      error "Failed to convert ENR to peer info", enr = $enr, err = error
```
Ivansete-status: Unrelated to the original purpose but this keep some consistency with others :)
```suggestion
      error "Failed to convert ENR to peer info", enr = $enr, error = error
```

waku/node/waku_node.nim
```diff
@@ -1386,22 +1386,19 @@ proc fetchPeerExchangePeers*(
...
-    info "Retrieved peer info via peer exchange protocol",
-      validPeers = validPeers, totalPeers = peers.len
-    return ok(validPeers)
-  else:
-    warn "failed to retrieve peer info via peer exchange protocol",
-      error = pxPeersRes.error
-    return err(pxPeersRes.error)
+  let pxPeers = (await node.wakuPeerExchangeClient.request(amount)).valueOr:
+    warn "failed to retrieve peer info via peer exchange protocol", error = error
+    return err(error)
```
Ivansete-status: This is a good opportunity to add a little bit of more context
```suggestion
    return err("fetchPeerExchangePeers failure: " & $error)
```

waku/waku_archive/driver/postgres_driver/postgres_driver.nim
```diff
@@ -1338,10 +1335,7 @@ proc removePartition(
   let partitionName = partition.getName()
   debug "beginning of removePartition", partitionName
 
-  var partSize = ""
-  let partSizeRes = await self.getTableSize(partitionName)
-  if partSizeRes.isOk():
-    partSize = partSizeRes.get()
+  let partSize = (await self.getTableSize(partitionName)).valueOr("")
```
Ivansete-status, line 1338: Gorgeous! :heart:

waku/waku_api/rest/server.nim
```diff
@@ -106,12 +106,9 @@ proc new*(
...
     middlewares = middlewares,
-  )
-  if sres.isOk():
-    server.httpServer = sres.get()
-    ok(server)
-  else:
-    err(sres.error)
+  ).valueOr:
+    return err(error)
+  ok(server)
```
Ivansete-status: Sorry for super nitpick comment. Even though Nim accepts that, I like the most explicit approach :)
```suggestion
  return ok(server)
```

waku/waku_archive/driver/postgres_driver/postgres_healthcheck.nim
```diff
@@ -24,8 +24,7 @@ proc checkConnectivity*(
 
         var numTrial = 0
         while numTrial < MaxNumTrials:
-          let res = await connPool.pgQuery(HealthCheckQuery)
-          if res.isOk():
+          (await connPool.pgQuery(HealthCheckQuery)).isOkOr:
```
Ivansete-status: In this particular case, I think this changes the logic, unless I'm missing something
```suggestion
          (await connPool.pgQuery(HealthCheckQuery)).isErrOr:
```

waku/waku_archive_legacy/driver/postgres_driver/postgres_healthcheck.nim
```diff
@@ -24,8 +24,7 @@ proc checkConnectivity*(
 
         var numTrial = 0
         while numTrial < MaxNumTrials:
-          let res = await connPool.pgQuery(HealthCheckQuery)
-          if res.isOk():
+          (await connPool.pgQuery(HealthCheckQuery)).isOkOr:
```
Ivansete-status: Similarly to the previous comment:
```suggestion
          (await connPool.pgQuery(HealthCheckQuery)).isErrOr:
```

fcecin:
Will address all the comments now, push and update, de-draft the PR and ping everyone.

apps/chat2bridge/chat2bridge.nim
fcecin: That looks like it should work, but it doesn't compile, unfortunately. I'm not 100% sure why. Maybe it's obvious, but I can't figure out how to make it work. But I can't find in nwaku any code block with `await` that is passed to the valueOr template as an argument. My general impression so far is that the Result templates can only handle simple cases and may not necessarily play well with other templates or macros.

waku/waku_api/rest/server.nim
fcecin: Was mimicking the original code but also itching to put a `return` there :)

waku/waku_archive/driver/postgres_driver/postgres_healthcheck.nim
fcecin: I was 100% sure I was going to screw up the logic somewhere. Code reviews to the rescue.

waku/node/waku_node.nim
fcecin: Turns out `error` there is not a string message but an error object of the `PeerExchangeResponseStatus` type, which is given as a result for calling `wakuPeerExchangeClient.request()`. That is the same type we return from `fetchPeerExchangePeers`, and looking into `request` it looks like the error object is sufficiently descriptive. We could append more information to the error by modifying `error.status_desc` at that point. Not sure if that is what we want here, since we have custom err() in `fetchPeerExchangePeers` that also isn't mentioning the function.

fcecin:
I will fix the "Commits must have verified signatures." mistake during the rebase/squash commit (my local git email was set to my gmail address instead of my waku.org email that I am using to sign the commits).

As for the CI errors, I can't think of a reason why these code changes would have broken anything. `test-ubuntu-22.04` shows a single test failure due to a port-already-in-use error that was already there (it's actually gone the last time I ran local tests).

Ivansete-status (approved):
LGTM! Thanks for it! :100:

apps/benchmarks/benchmarks.nim
```diff
@@ -28,10 +28,9 @@ proc benchmark(
       iter = i, elapsed_ms = (getTime() - start_time).inMilliseconds
 
   discard await manager.updateRoots()
-  let proofResult = await manager.fetchMerkleProofElements()
-  if proofResult.isErr():
-    error "Failed to fetch Merkle proof", error = proofResult.error
-  manager.merkleProofCache = proofResult.get()
+  manager.merkleProofCache = (await manager.fetchMerkleProofElements()).valueOr:
+    error "Failed to fetch Merkle proof", error = error
+    quit(QuitFailure)
```
NagyZoltanPeter, line 33: I guess it is intentional, but did not exist in the former code.

apps/chat2/chat2.nim
```diff
@@ -202,19 +189,19 @@ proc publish(c: Chat, line: string) =
...
-    if appendRes.isErr():
-      debug "could not append rate limit proof to the message"
-    else:
+    block appendRLNProofBlock:
+      # for future version when we support more than one rln protected content topic,
+      # we should check the message content topic as well
+      (c.node.wakuRlnRelay.appendRLNProof(message, float64(time))).isOkOr:
+        debug "could not append rate limit proof to the message"
+        break appendRLNProofBlock
+
```
NagyZoltanPeter: I tend to not agree here to use the isOkOr. I feel the older more appropriate in this particular case. For me it seems more easy to understand.

apps/chat2/chat2.nim
```diff
@@ -438,16 +425,17 @@ proc processInput(rfd: AsyncFD, rng: ref HmacDrbgContext) {.async.} =
...
-        waitFor chat.node.connectToNodes(discoveredNodes)
-    else:
-      warn "Failed to init Waku DNS discovery"
+    block discoveryBlock:
+      let wakuDnsDiscovery = WakuDnsDiscovery.init(dnsDiscoveryUrl.get(), resolver).valueOr:
+        warn "Failed to init Waku DNS discovery", error = error
+        break discoveryBlock
+      let discoveredPeers = (await wakuDnsDiscovery.findPeers()).valueOr:
+        warn "Failed to find peers via DNS discovery", error = error
+        break discoveryBlock
```
NagyZoltanPeter: Maybe just for me, but this block+break sounds like goto :-D
What about try use here isErrOr / errorOr?

NagyZoltanPeter (approved):
Added some nitpicks, if you would like to consider, but all in all congrats for your first nwaku PR! It is nice and has great solutions, improvements. Thanks for it.

apps/benchmarks/benchmarks.nim
fcecin: > I guess it is intentional, but did not exist in the former code.

Yes, my understanding is that in the previous pattern:

```
  let proofResult = await manager.fetchMerkleProofElements()
  if proofResult.isErr():
    error "Failed to fetch Merkle proof", error = proofResult.error
  manager.merkleProofCache = proofResult.get()
```

... the `if ... isErr()` path is just doing logging and not breaking away from the scope -- I believe the `error` macro/template doesn't expand into some return, throw, break, etc. that would change control flow (I tested that in a tiny standalone program also to be sure). That means that the original code is actually buggy: it prints an error message then proceeds to `proofResult.get()` which would crash (throw a runtime exception). Since there's no surrounding `try:` (that I can spot), it's the same as `quit(QuitFailure)` in this case which is cleaner.

This is an interesting case because it's an error that either never happens, or it happens when the benchmark tool is actually broken, in which case you would get the `error` log message explaining what's wrong plus a `get()` crash right after it in the logs. Strictly speaking, the original code had a bug, but one could argue that its a bug that never triggers and if it did it wouldn't cause any additional problems anyway.

apps/chat2/chat2.nim
fcecin: > I tend to not agree here to use the isOkOr. I feel the older more appropriate in this particular case. For me it seems more easy to understand.

Yeah, we tried to make isOkOr work here without the block: trick, but since isOkOr usually requires some control flow interruption, it's the only way I found to make it work here. I'll revert this for now, at least until I can learn why the results templates that require a return/break/continue/etc. sometimes work without these.

apps/chat2/chat2.nim
fcecin: As far as I can tell, the errorOr / valueOr templates allow us to provide a default/alternative value either directly or through a block of code that provides that alternative value, but in the original code if there's a failure then there's logic that is simply skipped -- it's not just a matter of providing an alternative value to some variable. The isOkOr / isErrOr macros allow us to provide a code block to execute on the "Or" path that needs (?) to break away from the current scope, which we can't do here because that would abort everything else that the proc is supposed to do -- unless we create a `block:` that we can break from (yep, it's goto-ish). 

So on long procs that have multiple steps (blocks of code, sequential parts) we will end up seeing the `if error do this else do that` pattern often (the pattern that was there before). I think the better refactor in these cases isn't obtained by simply switching the macro/templates from the `results` library that we are using there (at least I wasn't able to see how that would work) but by creating additional sub-procs to run each block that these long parent procs have.

I'm not sure if we want to be doing that deeper refactor. If we are we should probably create a separate refactor issue e.g. "break up long procs in smaller procs." Then inside each smaller proc we can just `return` inside the `Or:` path of the `Result` template we are using, which is nicer than `block: ... break` and allows us to naturally use `isOkOr` etc. But now one may have to jump around the source file more, between these smaller sub-procs and the parent proc that calls them e.g. `proc parentproc ... = subproc1() ; subproc2() ; subproc3()`

I wasn't too sure about this block: trick. It does look weird and the actual fix is probably a deeper kind of refactoring such as the one I sketched above, so here I'll just remove the `block:` trick and revert to the previous code.

apps/chat2bridge/chat2bridge.nim
fcecin: > This whole block can be converted to the following:
>  (...)

Was just missing a `continue` at the end of the `valueOr:` branch, now it compiles. `discard ... errorOr` also works here.

The original code also had a bug (I assume): sleeping for 10s plus cmd.pollPeriod on error.

This will go in #3615

PR #3611, opened 2025-10-03
chore: provide an option to specify log level per topic

chaitanyaprem:
## Description
Trying to provide a config to enable loglevel per topic. 

Noticed a weird issue where while trying to set loglevel using the config it fails if the topics have spaces in them e.g "waku mix" and works without any issues when using `wakumix` without spaces. 
Not sure if it is a bug in chronicles or something else. 

Also i was not able to set the logLevel for underlying mix protocol code eventhough i have specified a topic name. Need to debug this.


chaitanyaprem:
closing this

PR #3613, opened 2025-10-06
fix: json escaping backslash

NagyZoltanPeter:
## Description

nim-json-serialization missed to escape backslash (escape char) while converting strings to json format.
In https://github.com/status-im/nim-json-serialization/pull/132 it is fixed.

## Changes
- bump nim-json-serialization with the escaping fix.
- added test case to cover backslash escaping

## Issue

closes https://github.com/waku-org/nwaku/issues/3572


NagyZoltanPeter:
Due https://github.com/waku-org/nwaku/pull/3616 does bumps nim-json-serialization to the same level, this can be closed without merge.

PR #3615, opened 2025-10-06
chore(refactoring): replace some isErr usage with better alternatives

### LINKED ISSUE #1969: chore(refactoring): apply isOkOr || valueOr approach

fcecin:
## Description

This PR refactors various patterns that use `isErr` across nwaku with less verbose and more idiomatic alternatives.

## Changes

This PR replaces varied usage of `isErr` with better patterns. It also contains a handful of other opportunistic refactors around error handling code.

There should not be any behavior (semantics) changes in this PR.

## Issue

Closes #1969.

### DIFF
--- a/tests/wakunode_rest/test_rest_store.nim
+++ b/tests/wakunode_rest/test_rest_store.nim
@@ -485,7 +485,7 @@ procSuite "Waku Rest API - Store v3":
       $response.contentType == $MIMETYPE_TEXT
       response.data.messages.len == 0
       response.data.statusDesc ==
-        "Failed parsing remote peer info [MultiAddress.init [multiaddress: Invalid MultiAddress, must start with `/`]]"
+        "Failed parsing remote peer info: MultiAddress.init [multiaddress: Invalid MultiAddress, must start with `/`]"
 
     await restServer.stop()
     await restServer.closeWait()


apps/chat2bridge/chat2bridge.nim
```diff
@@ -252,25 +249,21 @@ when isMainModule:
...
-    (nodev2ExtIp, nodev2ExtPort, _) = natRes.get()
-    ## The following heuristic assumes that, in absence of manual
-    ## config, the external port is the same as the bind port.
-    extPort =
-      if nodev2ExtIp.isSome() and nodev2ExtPort.isNone():
-        some(Port(uint16(conf.libp2pTcpPort) + conf.portsShift))
-      else:
-        nodev2ExtPort
+  ).valueOr:
+    raise newException(ValueError, "setupNat error " & error)
```
Ivansete-status, line 258: We should not raise new exceptions :)

apps/chat2mix/chat2mix.nim
```diff
@@ -333,57 +331,60 @@ proc maintainSubscription(
   const maxFailedServiceNodeSwitches = 10
   var noFailedSubscribes = 0
   var noFailedServiceNodeSwitches = 0
+  const retryWaitMs = 2000 # Quick retry interval
+  const subscriptionMaintenanceMs = 30000 # Subscription maintenance interval
```
Ivansete-status: Super nitpick comment :)
```suggestion
  const RetryWaitMs = 2000 # Quick retry interval
  const SubscriptionMaintenanceMs = 30000 # Subscription maintenance interval
```

apps/chat2mix/chat2mix.nim
```diff
@@ -333,57 +331,60 @@ proc maintainSubscription(
...
+  var nextWaitMs = 0
   while true:
+    await sleepAsync(nextWaitMs)
```
Ivansete-status: Isn't that changing the logic a little? Maybe better to keep it at the end of the loop iteration

apps/chat2mix/chat2mix.nim
```diff
@@ -333,57 +331,60 @@ proc maintainSubscription(
...
+      nextWaitMs = retryWaitMs # Wait a bit before retrying
+    elif not preventPeerSwitch:
+      nextWaitMs = 0 # try again with new peer without delay
+      let peerOpt = selectRandomServicePeer(
+        wakuNode.peerManager, some(actualFilterPeer), WakuFilterSubscribeCodec
+      )
+      peerOpt.isOkOr:
+        error "Failed to find new service peer. Exiting."
+        noFailedServiceNodeSwitches += 1
+        break
```
Ivansete-status: Maybe better to use `valueOr` in this case?

apps/liteprotocoltester/diagnose_connections.nim
```diff
@@ -59,7 +59,4 @@ proc logSelfPeers*(pm: PeerManager) =
 {allPeers(pm)}
 *------------------------------------------------------------------------------------------*""".fmt()
 
-  if printable.isErr():
-    echo "Error while printing statistics: " & printable.error().msg
-  else:
-    echo printable.get()
+  echo printable.valueOr("Error while printing statistics: " & error.msg)
```
Ivansete-status, line 62: Awesome! Well done :partying_face:

waku/common/databases/db_sqlite.nim
```diff
@@ -497,9 +488,8 @@ proc performSqliteVacuum*(db: SqliteDatabase): DatabaseResult[void] =
 
   info "starting sqlite database vacuuming"
 
-  let resVacuum = db.vacuum()
-  if resVacuum.isErr():
-    return err("failed to execute vacuum: " & resVacuum.error)
+  db.vacuum().isOkOr:
+    return err("failed to execute vacuum")
```
Ivansete-status: This should work :)
```suggestion
    return err("failed to execute vacuum: " & error)
```

waku/waku_api/rest/admin/handlers.nim
```diff
@@ -426,14 +426,13 @@ proc installAdminV1GetFilterSubsHandler(router: var RestRouter, node: WakuNode)
         FilterSubscription(peerId: $peerId, filterCriteria: filterCriteria)
       )
 
-    let resp = RestApiResponse.jsonResponse(subscriptions, status = Http200)
-    if resp.isErr():
-      error "An error ocurred while building the json respose: ", error = resp.error
+    let resp = RestApiResponse.jsonResponse(subscriptions, status = Http200).valueOr:
+      error "An error ocurred while building the json respose: ", error = error
```
Ivansete-status: nitpick a little bit aside from original PR's purpose
```suggestion
      error "An error ocurred while building the json respose", error = error
```

waku/waku_api/rest/legacy_store/handlers.nim
```diff
@@ -34,20 +35,17 @@ proc performHistoryQuery(
...
 
-  let res = queryFut.read()
-  if res.isErr():
-    const msg = "Error occurred in queryFut.read()"
-    error msg, error = res.error
-    return RestApiResponse.internalServerError(fmt("{msg} [{res.error}]"))
+  let storeResp = queryFut.read().map(res => res.toStoreResponseRest()).valueOr:
+      const msg = "Error occurred in queryFut.read()"
+      error msg, error = error
+      return RestApiResponse.internalServerError(fmt("{msg} [{error}]"))
```
Ivansete-status, line 41: Another nitpick ;P the indentation doesn't seem two spaces
```suggestion
  let storeResp = queryFut.read().map(res => res.toStoreResponseRest()).valueOr:
    const msg = "Error occurred in queryFut.read()"
    error msg, error = error
    return RestApiResponse.internalServerError(fmt("{msg} [{error}]"))
```

waku/waku_api/rest/relay/handlers.nim
```diff
@@ -318,9 +314,7 @@ proc installRelayApiHandlers*(
     if not await publishFut.withTimeout(futTimeout):
       return RestApiResponse.internalServerError("Failed to publish: timedout")
 
-    var res = publishFut.read()
-
-    if res.isErr():
-      return RestApiResponse.badRequest("Failed to publish. " & res.error)
+    publishFut.read().isOkOr:
+      return RestApiResponse.badRequest("Failed to publish. " & error)
```
Ivansete-status: A good opportunity to enforce the "colon" style :D
```suggestion
      return RestApiResponse.badRequest("Failed to publish: " & error)
```

waku/waku_api/rest/rest_serdes.nim
```diff
@@ -45,15 +45,12 @@ func decodeRequestBody*[T](
...
   let reqBodyData = contentBody.get().data
 
-  let requestResult = decodeFromJsonBytes(T, reqBodyData)
-  if requestResult.isErr():
+  let requestResult = decodeFromJsonBytes(T, reqBodyData).valueOr:
     return err(
-      RestApiResponse.badRequest(
-        "Invalid content body, could not decode. " & $requestResult.error
-      )
+      RestApiResponse.badRequest("Invalid content body, could not decode. " & $error)
```
Ivansete-status:
```suggestion
      RestApiResponse.badRequest("Invalid content body, could not decode: " & $error)
```

waku/waku_api/rest/store/handlers.nim
```diff
@@ -35,14 +36,10 @@ proc performStoreQuery(
...
-  if futRes.isErr():
-    const msg = "Error occurred in queryFut.read()"
-    error msg, error = futRes.error
-    return RestApiResponse.internalServerError(fmt("{msg} [{futRes.error}]"))
-
-  let res = futRes.get().toHex()
+  let res = queryFut.read().map(val => val.toHex()).valueOr:
+      const msg = "Error occurred in queryFut.read()"
+      error msg, error = error
+      return RestApiResponse.internalServerError(fmt("{msg} [{error}]"))
```
Ivansete-status, line 42:
```suggestion
  let res = queryFut.read().map(val => val.toHex()).valueOr:
    const msg = "Error occurred in queryFut.read()"
    error msg, error = error
    return RestApiResponse.internalServerError(fmt("{msg} [{error}]"))
```

waku/waku_core/peers.nim
```diff
@@ -249,11 +249,10 @@ proc parseUrlPeerAddr*(
     return ok(none(RemotePeerInfo))
 
   let parsedAddr = decodeUrl(peerAddr.get())
-  let parsedPeerInfo = parsePeerInfo(parsedAddr)
-  if parsedPeerInfo.isErr():
-    return err("Failed parsing remote peer info [" & parsedPeerInfo.error & "]")
+  let parsedPeerInfo = parsePeerInfo(parsedAddr).valueOr:
+    return err("Failed parsing remote peer info [" & error & "]")
```
Ivansete-status: A good opportunity to keep the same "colon" format :D
```suggestion
    return err("Failed parsing remote peer info: " & error)
```

waku/waku_filter_v2/protocol.nim
```diff
@@ -309,15 +308,13 @@ proc initProtocolHandler(wf: WakuFilter) =
         amount = buf.len().int64, labelValues = [WakuFilterSubscribeCodec, "in"]
       )
 
-      let decodeRes = FilterSubscribeRequest.decode(buf)
-      if decodeRes.isErr():
+      #TODO: toAPI() split here
```
Ivansete-status: I think we can delete this comment
```suggestion
```

waku/waku_store_legacy/client.nim
```diff
@@ -112,11 +109,8 @@ when defined(waku_exp_store_resume):
     var messageList: seq[WakuMessage] = @[]
 
     while true:
-      let queryRes = await w.query(req, peer)
-      if queryRes.isErr():
-        return err($queryRes.error)
-
-      let response = queryRes.get()
+      let response = (await w.query(req, peer)).valueOr:
+        return err($error)
```
Ivansete-status, line 113: Maybe :) ?

```suggestion
      let response = ?(await w.query(req, peer))

```

waku/waku_store_sync/transfer.nim
```diff
@@ -58,9 +58,8 @@ proc sendMessage(
   let writeRes = catch:
     await conn.writeLP(rawPayload)
 
-  if writeRes.isErr():
-    return
-      err("remote " & $conn.peerId & " connection write error: " & writeRes.error.msg)
+  writeRes.isOkOr:
+    return err("remote " & $conn.peerId & " connection write error: " & error.msg)
```
Ivansete-status: Maybe useful in this case to indicate that anything between [ ... ] is not hard-coded and the is easier to lookup later on
```suggestion
    return err("remote [" & $conn.peerId & "] connection write error: " & error.msg)
```

apps/chat2bridge/chat2bridge.nim
fcecin: The previous code "crashes," that is, raises a `ResultDefect` from inside `vendor/nim-results/results.nim` library if `natRes.isErr()`:

```
  if natRes.isErr():
    error "Error in setupNat", error = natRes.error

  # Load address configuration
  let
    (nodev2ExtIp, nodev2ExtPort, _) = natRes.get()
```

Since `error` is just logging, program flow continues, then `Result.get()` calls `Result.value()` which calls `Result.withAssertOk()`:

```
template withAssertOk(self: Result, body: untyped): untyped =
  # Careful - `self` evaluated multiple times, which is fine in all current uses
  case self.oResultPrivate
  of false:
    when self.E isnot void:
      raiseResultDefect("Trying to access value with err Result", self.eResultPrivate)
    else:
      raiseResultDefect("Trying to access value with err Result")
  of true:
    body
```

This "setupNat()" code pattern appears several times in the code and in all of them it does throw that exception. This is the only place where it was still throwing the invariant violation defect. I think in this case we can write this off as an improvement (upgrade from defect to exception), especially because this exact code elsewhere does throw the same exception, so it remains consistent (until we refactor the code to purge all exception handling).

apps/chat2mix/chat2mix.nim
fcecin: You're right, this is worse, not better. I reverted to simply calling `sleepAsync` at each branch, which gets rid of that `nextWaitMs` var.

waku/common/databases/db_sqlite.nim
fcecin: Whoops!

waku/waku_api/rest/admin/handlers.nim
fcecin: Thanks!

waku/waku_api/rest/legacy_store/handlers.nim
fcecin: That is being enforced by the Nim formatter (nph?) at least in my VSCode extension installation. If you cut it to two spaces, your Nim formatter doesn't push it back? (it formats Nim code before save here, so when I hit CTRL+S to save, it puts the extra two spaces back for me).

This code that breaks up the chained map and the lambda will keep the formatter happy with two spaces:

```
  let storeResp = queryFut.read()
  storeResp = storeResp.map(res => res.toStoreResponseRest()).valueOr:
    const msg = "Error occurred in queryFut.read()"
    error msg, error = error
    return RestApiResponse.internalServerError(fmt("{msg} [{error}]"))
```

So the problem here I think is that the lambda is inside a function call, so it is adding up the indentation of the chained proc call and the lambda. Unless I screwed up somewhere in configuring the formatter on my side, I don't think we should be fighting the formatter here.

waku/waku_api/rest/store/handlers.nim
fcecin: Same as above: https://github.com/waku-org/nwaku/pull/3615#discussion_r2465298000

waku/waku_store_legacy/client.nim
fcecin: Doesn't compile unfortunately... :(

Need to compile with `make wakunode2 NIMFLAGS="-d:waku_exp_store_resume"` to see this one (there's also two missing conditional `import`s for this file to compile with that flag, so I'll include that fix as well in this PR).

The problem is that the `WakuStoreResult` that this proc returns (wrapped by a `Future`) is a `Result[T, string]`, which is why you need to `$` the `error` before returning. The `?` is a shorthand for `return err(error)` so it won't compile unless the error object has some way to know it can become a string automatically there:

```
/home/fcecin/waku/nwaku/waku/waku_store_legacy/client.nim(115, 22) template/generic instantiation of `?` from here
/home/fcecin/waku/nwaku/vendor/nim-results/results.nim(1560, 21) template/generic instantiation of `err` from here
/home/fcecin/waku/nwaku/vendor/nim-results/results.nim(474, 42) Error: type mismatch: got 'HistoryError' for 'v`gensym114.eResultPrivate' but expected 'string'
stack trace: (most recent call last)
```

... which it seems it doesn't, unfortunately. This is actually something we may want to look into when we get the time to do an overhaul of error handling in general, consider adding some type auto conversion between error objects and the string type (I understand implicit conversions in general are against the style guide by default, but maybe in this case is something we actually want).

For reference, `HistoryError` seems to be this: 

```
  HistoryError* = object
    case kind*: HistoryErrorKind
    of PEER_DIAL_FAILURE:
      address*: string
    of BAD_RESPONSE, BAD_REQUEST:
      cause*: string
    else:
      discard

(...)

proc `$`*(err: HistoryError): string =
  case err.kind
  of HistoryErrorKind.PEER_DIAL_FAILURE:
    "PEER_DIAL_FAILURE: " & err.address
  of HistoryErrorKind.BAD_RESPONSE:
    "BAD_RESPONSE: " & err.cause
  of HistoryErrorKind.BAD_REQUEST:
    "BAD_REQUEST: " & err.cause
  of HistoryErrorKind.TOO_MANY_REQUESTS:
    "TOO_MANY_REQUESTS"
  of HistoryErrorKind.SERVICE_UNAVAILABLE:
    "SERVICE_UNAVAILABLE"
  of HistoryErrorKind.UNKNOWN:
    "UNKNOWN"
```

fcecin: Actually, this `when defined(waku_exp_store_resume)` code seems to have quite a bit of code rot, so if we are going to address the compilation errors here we should probably create a separate issue. I'll leave this code alone for now.

darshankabariya (approved):
Thanks for taking care !! LGTM

PR #3617, opened 2025-10-08
feat: update rendezvous to broadcast and discover WakuPeerRecords

chaitanyaprem:
## Description

As per [this](https://github.com/waku-org/research/issues/118) , we can use rendezvous for discovering peers and their mix key. This is first PR using updated libp2p rendezvous code to publish WakuPeerRecord and fetch peers.

Reference deliverable https://github.com/waku-org/pm/issues/366
RFC https://github.com/waku-org/specs/pull/95

## Changes

- [x] make libp2p 1.14.3 as min version as that has [rendezvous related updates](https://github.com/vacp2p/nim-libp2p/pull/1719) 
- [x] Use separate codec to represent Waku-Rendezvous as we are changing the behaviour of default rendezvous
- [x] use rendezvous to advertise only mix info (namespace used is `/rs/<cluster-Id>/mix` and discover nodes as well using same namespace
- [x] add discovered peers mixKey to PeerBook so that they can be added to mix pool
- [x]  provide option to create rendezvous client mode so that lightclients can just use rendezvous client to query.
- [x] verify rendezvous discovery using mix simulation via chat2mix
- [x] takes mix related fixes done in libp2p
- [x] little larger simulation of mixnet(20-30 nodes) to see if there are any issues

next steps once PR is merged: 
- deploy and dogfooding in TWN (this can be done post merging into master so that fleet also can be updated)
- fine tune rendezvous peer discovery logic to suit mix (this can be done as part of subsequent PR as well)

### DIFF
--- a/waku/waku_rendezvous/protocol.nim
+++ b/waku/waku_rendezvous/protocol.nim
@@ -45,7 +45,6 @@ method discover*(
     self: WakuRendezVous, conn: Connection, d: Discover
 ) {.async: (raises: [CancelledError, LPStreamError]).} =
   # Override discover method to avoid collect macro generic instantiation issues
-  # TODO figure out if we can use parent generic discover
   trace "Received Discover", peerId = conn.peerId, ns = d.ns
   await procCall GenericRendezVous[WakuPeerRecord](self).discover(conn, d)
 
--- a/waku/waku_rendezvous/waku_peer_record.nim
+++ b/waku/waku_rendezvous/waku_peer_record.nim
@@ -47,7 +47,7 @@ proc decode*(
 
   ?pb.getRequiredField(4, record.mixKey)
 
-  ok(record)
+  return ok(record)
 
 proc encode*(record: WakuPeerRecord): seq[byte] =
   var pb = initProtoBuffer()
@@ -61,7 +61,7 @@ proc encode*(record: WakuPeerRecord): seq[byte] =
   pb.write(4, record.mixKey)
 
   pb.finish()
-  pb.buffer
+  return pb.buffer
 
 proc checkWakuPeerRecord*(
     _: WakuPeerRecord, spr: seq[byte], peerId: PeerId


waku/waku_rendezvous/waku_peer_record.nim
```diff
@@ -0,0 +1,74 @@
...
+  discard ?pb.getRepeatedField(3, record.addresses)
+
+  if record.addresses.len == 0:
+    return err(ProtoError.RequiredFieldMissing)
+
+  ?pb.getRequiredField(4, record.mixKey)
+
+  ok(record)
+
+proc encode*(record: WakuPeerRecord): seq[byte] =
```
jm-clius, line 52: Encoding differs from what seems to be proposed in https://github.com/waku-org/specs/pull/95/files, specifically introduction of `seqNo` and no `protocols` field. Perhaps intentional?

waku/waku_rendezvous/waku_peer_record.nim
```diff
@@ -0,0 +1,74 @@
...
+  # Considering only mix as of now, but we can keep extending this to include all capabilities part of Waku ENR
+  peerId*: PeerId
+  seqNo*: uint64
+  addresses*: seq[MultiAddress]
+  mixKey*: string
+
+proc payloadDomain*(T: typedesc[WakuPeerRecord]): string =
+  $multiCodec("libp2p-custom-peer-record")
+
+proc payloadType*(T: typedesc[WakuPeerRecord]): seq[byte] =
```
jm-clius, line 23: With quick search I can't find this defined anywhere or in the standard multicodec table - is this our own payload type for "custom peer record" or specifically `WakuPeerRecord`? If so, can we make this a `const` or add it to the spec?

waku/waku_rendezvous/protocol.nim
```diff
@@ -1,117 +1,157 @@
...
+  getPeerRecord: GetWakuPeerRecord
 
   registrationInterval: timer.Duration
   periodicRegistrationFut: Future[void]
 
   requestInterval: timer.Duration
   periodicRequestFut: Future[void]
-
-proc batchAdvertise*(
+  clientOnlyMode: bool
```
jm-clius: Not sure if I'm missing some subtlety, but I find it a bit weird that `client` is a state on `WakuRendezVous`. Why not split into a `...client` type (and module) and service/protocol module as we do for other protocols? I assume there might be too much shared state to make it worth the effort, but I cannot see any fundamental reason why not in the `discover` and `advertise` below.

waku/node/waku_node.nim
```diff
@@ -346,6 +358,24 @@ proc selectRandomPeers*(peers: seq[PeerId], numRandomPeers: int): seq[PeerId] =
   shuffle(randomPeers)
   return randomPeers[0 ..< min(len(randomPeers), numRandomPeers)]
 
+proc mountRendezvousClient*(node: WakuNode, clusterId: uint16) {.async: (raises: []).} =
+  info "mounting rendezvous discovery protocol"
+
+  node.wakuRendezvous = WakuRendezVous.new(
```
jm-clius: Mentioned below in another comment, but since we don't actually mount anything on the switch as a client, why not have a separate `WakuRendezvousClient` type?

waku/waku_rendezvous/waku_peer_record.nim
chaitanyaprem: seqNo is missed from the spec and protocols got added as that was not currently planned, will update the spec to match implementation.

waku/waku_rendezvous/waku_peer_record.nim
chaitanyaprem: Yes, i had added this in private section of multicodec's in [this](https://github.com/vacp2p/nim-libp2p/pull/1719) PR. Will make it a const and add to the spec. Thanks!

waku/waku_rendezvous/protocol.nim
chaitanyaprem: Yea, i did not want to duplicate code for discover which i had to write due to some generics limitation i was hitting. 
And if you notice there is way too much common between them and hence kept it as is. Will see if i can reuse it in any other way, but might take more effort considering generics.

apps/chat2mix/chat2mix.nim
```diff
@@ -150,7 +150,7 @@ proc getChatLine(payload: seq[byte]): string =
   # No payload encoding/encryption from Waku
   let pb = Chat2Message.init(payload).valueOr:
     return string.fromBytes(payload)
-  return $pb
+  return pb.toString()
```
Ivansete-status: It is clear what it does but maybe not super idiomatic :)

apps/chat2mix/chat2mix.nim
```diff
@@ -590,6 +593,7 @@ proc processInput(rfd: AsyncFD, rng: ref HmacDrbgContext) {.async.} =
   #await mountLegacyLightPush(node)
   node.peerManager.addServicePeer(servicePeerInfo, WakuLightpushCodec)
   node.peerManager.addServicePeer(servicePeerInfo, WakuPeerExchangeCodec)
+  #node.peerManager.addServicePeer(servicePeerInfo, WakuRendezVousCodec)
```
Ivansete-status: Shall we remove this?
```suggestion
```

waku/waku_mix/protocol.nim
```diff
@@ -74,34 +72,52 @@ func getIPv4Multiaddr*(maddrs: seq[MultiAddress]): Option[MultiAddress] =
...
-      peerId = remotePeers[i].peerId, enr = remotePeerENR, maddr = maddrWithPeerId
-
-    let peerMixPubKey = mixKey(remotePeerENR).get()
-    let mixNodePubInfo =
-      createMixPubInfo(maddrWithPeerId.value, intoCurve25519Key(peerMixPubKey))
+    let maddrWithPeerId = appendPeerIdToMultiaddr(ipv4addr, remotePeers[i].peerId)
+    trace "remote peer info", info = remotePeers[i]
+
+    if remotePeers[i].mixPubKey.isNone():
+      trace "peer has no mix Pub Key", peer = $remotePeers[i]
```
Ivansete-status: To make it a bit clearer what peer we are referring:
```suggestion
      trace "peer has no mix Pub Key", remotePeerId = $remotePeers[i]
```

waku/waku_mix/protocol.nim
```diff
@@ -74,34 +72,52 @@ func getIPv4Multiaddr*(maddrs: seq[MultiAddress]): Option[MultiAddress] =
...
+      continue
+
+    let peerMixPubKey = remotePeers[i].mixPubKey.get()
+    var peerPubKey: crypto.PublicKey
+    if not remotePeers[i].peerId.extractPublicKey(peerPubKey):
+      warn "Failed to extract public key from peerId, skipping node",
+        peerId = remotePeers[i].peerId
```
```suggestion
        remotePeerId = remotePeers[i].peerId
```

waku/waku_mix/protocol.nim
```diff
@@ -74,34 +72,52 @@ func getIPv4Multiaddr*(maddrs: seq[MultiAddress]): Option[MultiAddress] =
...
+      continue
+
+    if peerPubKey.scheme != PKScheme.Secp256k1:
+      warn "Peer public key is not Secp256k1, skipping node",
+        peerId = remotePeers[i].peerId, scheme = peerPubKey.scheme
```
```suggestion
        remotePeerId = remotePeers[i].peerId, scheme = peerPubKey.scheme
```

waku/waku_mix/protocol.nim
```diff
@@ -74,34 +72,52 @@ func getIPv4Multiaddr*(maddrs: seq[MultiAddress]): Option[MultiAddress] =
...
+      continue
+
+    let mixNodePubInfo = MixPubInfo.init(
+      remotePeers[i].peerId,
+      ipv4addr,
+      intoCurve25519Key(peerMixPubKey),
+      peerPubKey.skkey,
+    )
+    trace "adding mix node to pool",
+      peerId = remotePeers[i].peerId, multiAddr = $ipv4addr
```
```suggestion
      remotePeerId = remotePeers[i].peerId, multiAddr = $ipv4addr
```

waku/waku_mix/protocol.nim
```diff
@@ -37,19 +39,15 @@ type
 proc mixPoolFilter*(cluster: Option[uint16], peer: RemotePeerInfo): bool =
   # Note that origin based(discv5) filtering is not done intentionally
   # so that more mix nodes can be discovered.
-  if peer.enr.isNone():
-    trace "peer has no ENR", peer = $peer
+  if peer.mixPubKey.isNone():
+    trace "peer has no mix Pub Key", peer = $peer
```
```suggestion
    trace "remote peer has no mix Pub Key", peer = $peer
```

waku/waku_rendezvous/protocol.nim
chaitanyaprem: Was able to split client and service node implementations easily without any duplication!

waku/node/waku_node.nim
chaitanyaprem: Did the split in https://github.com/waku-org/nwaku/pull/3617/commits/f34633aee26d2e77a8df5738e4c5a5c4216656ff

waku/waku_rendezvous/client.nim
```diff
@@ -0,0 +1,138 @@
...
+    # Use the libp2p rendezvous request method
+    records = await self.rdv.request(
+      Opt.some(namespace), Opt.some(PeersRequestedCount), Opt.some(@[rpi.peerId])
+    )
+  except CatchableError as e:
+    return err("rendezvous request failed: " & e.msg)
+
+  trace "waku rendezvous client request got peers", count = records.len
+  for record in records:
+    rendezvousPeerFoundTotal.inc()
```
NagyZoltanPeter: Shall not we count unique peers within this counter? Or we expect somehow there are no dups?

waku/waku_rendezvous/protocol.nim
```diff
@@ -1,117 +1,146 @@
...
-  periodicRequestFut: Future[void]
-
-proc batchAdvertise*(
+# Override discover method to avoid collect macro generic instantiation issues
+# TODO figure out if we can use parent generic discover
+method discover*(
+    self: WakuRendezVous, conn: Connection, d: Discover
+) {.async: (raises: [CancelledError, LPStreamError]).} =
+  trace "Received Discover", peerId = conn.peerId, ns = d.ns
+  if d.ns.isSome() and d.ns.get().len > 255: #MaximumNamespaceLen
```
NagyZoltanPeter: Can't we use here a `const MaxNamespaceLen = 255` value?

waku/waku_rendezvous/protocol.nim
```diff
@@ -237,69 +198,95 @@ proc periodicRegistration(self: WakuRendezVous) {.async.} =
...
 
 proc start*(self: WakuRendezVous) {.async: (raises: []).} =
+  # Start the parent GenericRendezVous (starts the register deletion loop)
+  if self.started:
+    warn "waku rendezvous already started"
+    return
+  try:
+    await procCall GenericRendezVous[WakuPeerRecord](self).start()
+  except CancelledError as exc:
+    error "failed to start GenericRendezVous", cause = exc.msg
```
NagyZoltanPeter, line 223: I don't get, if it does not start why call issue periodicRegistration?

waku/waku_mix/protocol.nim
```diff
@@ -37,19 +39,15 @@ type
 proc mixPoolFilter*(cluster: Option[uint16], peer: RemotePeerInfo): bool =
```
NagyZoltanPeter: I think we need a better naming for this proc. From the name and even reading the condition flow I don't get the aim.

waku/waku_rendezvous/client.nim
chaitanyaprem: good point, as of now it is just counting peers returned which may already have been discovered already via rendezvous. let me see if there is a way we can make it unique. Not sure if we can do it at rendezvous level as addPeer doesn't return any info to indicate the same.

examples/lightpush_mix/lightpush_publisher_mix.nim
```diff
@@ -145,20 +135,26 @@ proc setupAndPublish(rng: ref HmacDrbgContext, conf: LightPushMixConf) {.async.}
...
   while i < conf.numMsgs:
+    var conn: Connection
     if conf.mixDisabled:
       let connOpt = await node.peerManager.dialPeer(dPeerId, WakuLightPushCodec)
       if connOpt.isNone():
         error "failed to dial peer with WakuLightPushCodec", target_peer_id = dPeerId
         return
       conn = connOpt.get()
+    else:
+      conn = node.wakuMix.toConnection(
```
Ivansete-status, line 146: The `toConnection` is implemented in both nwaku and in nim-libp2p. Yet, their signatures seem similar. I'd suggest make that clearer or avoid duplicate code.

examples/lightpush_mix/lightpush_publisher_mix.nim
```diff
@@ -167,22 +163,23 @@ proc setupAndPublish(rng: ref HmacDrbgContext, conf: LightPushMixConf) {.async.}
...
-      notice "published message",
-        text = text,
-        timestamp = message.timestamp,
-        psTopic = LightpushPubsubTopic,
-        contentTopic = LightpushContentTopic
-    else:
-      error "failed to publish message", error = $res.error
-      lp_mix_failed.inc(labelValues = ["publish_error"])
-
+    try:
```
Ivansete-status: This try-catch seems not to be needed? Or maybe you noticed an exception?

examples/lightpush_mix/lightpush_publisher_mix.nim
```diff
@@ -167,22 +163,23 @@ proc setupAndPublish(rng: ref HmacDrbgContext, conf: LightPushMixConf) {.async.}
...
+      if res.isOk():
+        lp_mix_success.inc()
+        notice "published message",
+          text = text,
+          timestamp = message.timestamp,
+          psTopic = LightpushPubsubTopic,
+          contentTopic = LightpushContentTopic
+      else:
+        error "failed to publish message", error = $res.error
+        lp_mix_failed.inc(labelValues = ["publish_error"])
```
Ivansete-status: We should always use `isOkOr` or `valueOr`
```suggestion
      (await node.wakuLightpushClient.publishWithConn(
        LightpushPubsubTopic, message, conn, dPeerId
      )).isOkOr:
        error "failed to publish message", error = $res.error
        lp_mix_failed.inc(labelValues = ["publish_error"])
        quit(1)
      
      lp_mix_success.inc()
      notice "published message",
        text = text,
        timestamp = message.timestamp,
        psTopic = LightpushPubsubTopic,
        contentTopic = LightpushContentTopic
      
```

tests/waku_discv5/test_waku_discv5.nim
```diff
@@ -468,6 +467,9 @@ suite "Waku Discovery v5":
       # leave some time for discv5 to act
       await sleepAsync(chronos.seconds(10))
 
+      # Connect peers via peer manager to ensure identify happens
+      discard await waku0.node.peerManager.connectPeer(waku1.node.switch.peerInfo)
```
Ivansete-status, line 471: Is the test bypassing something that should be done by the core logic?

waku/common/callbacks.nim
```diff
@@ -1,5 +1,7 @@
-import ../waku_enr/capabilities
+import ../waku_enr/capabilities, ../waku_rendezvous/waku_peer_record
```
Ivansete-status: A good opportunity to use absolute path :)

```suggestion
import waku/waku_enr/capabilities, waku/waku_rendezvous/waku_peer_record
```

waku/node/waku_node.nim
```diff
@@ -355,13 +381,19 @@ proc mountRendezvous*(node: WakuNode, clusterId: uint16) {.async: (raises: []).}
     clusterId,
     node.getShardsGetter(),
     node.getCapabilitiesGetter(),
+    node.getWakuPeerRecordGetter(),
```
Ivansete-status, line 385: Given that we pass a value, maybe we can adapt the following so that this param is mandatory:
```nim
proc new*(
    T: type WakuRendezVous,
    switch: Switch,
    peerManager: PeerManager,
    clusterId: uint16,
    getShards: GetShards,
    getCapabilities: GetCapabilities,
    getPeerRecord: GetWakuPeerRecord,
```

waku/node/waku_node.nim
```diff
@@ -346,6 +360,18 @@ proc selectRandomPeers*(peers: seq[PeerId], numRandomPeers: int): seq[PeerId] =
   shuffle(randomPeers)
   return randomPeers[0 ..< min(len(randomPeers), numRandomPeers)]
 
+proc mountRendezvousClient*(node: WakuNode, clusterId: uint16) {.async: (raises: []).} =
```
Ivansete-status, line 364: Considering the recent refactor, maybe this is a good opportunity to avoid `waku_node` keep growing and add the `rendezvous` logic within the upcoming `kernel_api` space

waku/waku_rendezvous/client.nim
```diff
@@ -0,0 +1,138 @@
...
+  libp2p/crypto/curve25519,
+  libp2p/switch,
+  libp2p/utils/semaphore
+
+import metrics except collect
+
+import
+  ../node/peer_manager,
+  ../waku_core/peers,
+  ../waku_core/codecs,
```
Ivansete-status: Nitpick: a bit clearer when using "absolute path"
```suggestion
  waku/node/peer_manager,
  waku/waku_core/peers,
  waku/waku_core/codecs,
```

waku/waku_rendezvous/client.nim
```diff
@@ -0,0 +1,138 @@
...
+proc requestAll*(
+    self: WakuRendezVousClient
+): Future[Result[void, string]] {.async: (raises: []).} =
+  trace "waku rendezvous client requests started"
+
+  let namespace = computeMixNamespace(self.clusterId)
+
+  # Get a random WakuRDV peer
+  let rpi = self.peerManager.selectPeer(WakuRendezVousCodec).valueOr:
+    return err("could not get a peer supporting WakuRendezVousCodec")
```
Ivansete-status, line 49: Maybe?
```suggestion
    return err("could not get a peer supporting WakuRendezVousCodec: " & $error)
```

waku/waku_rendezvous/client.nim
Ivansete-status: Another doubt: shouldn't we inc only if success?

waku/waku_rendezvous/client.nim
```diff
@@ -0,0 +1,138 @@
...
+
+proc periodicRequests(self: WakuRendezVousClient) {.async.} =
+  info "waku rendezvous periodic requests started", interval = self.requestInterval
+
+  # infinite loop
+  while true:
+    (await self.requestAll()).isOkOr:
+      error "waku rendezvous requests failed", error = error
+
+    await sleepAsync(self.requestInterval)
```
Ivansete-status: Let's make sure there's some delay even when failing?
```suggestion
    await sleepAsync(self.requestInterval)
    (await self.requestAll()).isOkOr:
      error "waku rendezvous requests failed", error = error
```

waku/waku_rendezvous/client.nim
```diff
@@ -0,0 +1,138 @@
...
+    peerManager: PeerManager,
+    clusterId: uint16,
+): Result[T, string] {.raises: [].} =
+  # Create a minimal GenericRendezVous instance for client-side requests
+  # We don't need the full server functionality, just the request method
+  let rng = newRng()
+  let rdv = GenericRendezVous[WakuPeerRecord](
+    switch: switch,
+    rng: rng,
+    sema: newAsyncSemaphore(5),
```
Ivansete-status: Shall we have a clear const and a comment stating why 5 (or other number) is needed?

waku/waku_rendezvous/client.nim
```diff
@@ -0,0 +1,138 @@
...
+  )
+
+  # Set codec separately as it's inherited from LPProtocol
+  rdv.codec = WakuRendezVousCodec
+
+  let client = T(
+    switch: switch,
+    peerManager: peerManager,
+    clusterId: clusterId,
+    requestInterval: 10.seconds,
```
Ivansete-status: Similarly, let's better use a clear `const` and its comment :)
```suggestion
    requestInterval: RdvLookupRateInterval???,
```

waku/waku_rendezvous/protocol.nim
```diff
@@ -1,117 +1,146 @@
...
 
-  requestInterval: timer.Duration
-  periodicRequestFut: Future[void]
-
-proc batchAdvertise*(
+# Override discover method to avoid collect macro generic instantiation issues
+# TODO figure out if we can use parent generic discover
+method discover*(
+    self: WakuRendezVous, conn: Connection, d: Discover
+) {.async: (raises: [CancelledError, LPStreamError]).} =
```
Ivansete-status: nitpick: in Nim, comments should always go below signatures

waku/waku_rendezvous/protocol.nim
```diff
@@ -1,117 +1,146 @@
...
-  ./common
+  ../waku_core/codecs,
+  ./common,
+  ./waku_peer_record
+
+# Explicitly make the items iterator visible in this module's scope
+# This is needed for generic instantiation of rendezvous methods
+iterator items*[T](o: OffsettedSeq[T]): T =
+  for item in offsettedseq.items(o):
+    yield item
```
Ivansete-status: I think this doesn't belong here. Instead, maybe need to adapt `vendor/nim-libp2p/libp2p/protocols/rendezvous/rendezvous.nim`?

waku/waku_rendezvous/protocol.nim
```diff
@@ -1,117 +1,146 @@
...
   registrationInterval: timer.Duration
   periodicRegistrationFut: Future[void]
 
-  requestInterval: timer.Duration
-  periodicRequestFut: Future[void]
-
-proc batchAdvertise*(
+# Override discover method to avoid collect macro generic instantiation issues
+# TODO figure out if we can use parent generic discover
+method discover*(
```
Ivansete-status, line 44: I believe we definitely need to revisit that :)
We cannot duplicate similar code that exists in `vendor/nim-libp2p/libp2p/protocols/rendezvous/rendezvous.nim`

waku/waku_rendezvous/protocol.nim
```diff
@@ -1,117 +1,146 @@
...
 ): Future[Result[void, string]] {.async: (raises: []).} =
-  ## Register with all rendezvous peers under a namespace
+  trace "advertising via waku rendezvous",
+    namespace = namespace, ttl = ttl, peers = $peers, peerRecord = $self.getPeerRecord()
+  let se = SignedPayload[WakuPeerRecord].init(
+    self.switch.peerInfo.privateKey, self.getPeerRecord()
+  ).valueOr:
+    return err("rendezvous advertisement failed: Failed to sign Waku Peer Record")
+  let sprBuff = se.encode().valueOr:
+    return err("rendezvous advertisement failed: Wrong Signed Peer Record")
```
Ivansete-status: Maybe?
```suggestion
  let se = SignedPayload[WakuPeerRecord].init(
    self.switch.peerInfo.privateKey, self.getPeerRecord()
  ).valueOr:
    return err("rendezvous advertisement failed: Failed to sign Waku Peer Record: " & $error)
  let sprBuff = se.encode().valueOr:
    return err("rendezvous advertisement failed: Wrong Signed Peer Record: " & $error)
```

examples/lightpush_mix/lightpush_publisher_mix.nim
chaitanyaprem: I don't think we have an implementation of toConnection in nwaku, we are just using the mix protocol's toConnection method.
can you point me to where it is implemented?

waku/waku_rendezvous/protocol.nim
chaitanyaprem: Well,  without this compilation keeps failing. This is the only way we could address the error which i don't remember exactly what it is.

I am happy if you can help debug the error and resolve it another way.

waku/waku_rendezvous/protocol.nim
chaitanyaprem: Well, I could not figure out a way to avoid compilation errors at this point in time without this. 

I don't want to duplicate code, but Generics in nim are causing crazy errors which are impossible to debug.
Would appreciate your help if there is a better way to resolve this.

apps/chat2mix/chat2mix.nim
chaitanyaprem: ok, can you recommend what is idiomatic then? 
Because i was facing compilation error with `$pb` which i was not facing before :) .

examples/lightpush_mix/lightpush_publisher_mix.nim
chaitanyaprem: Yea, initially i had a catch here but later i added one inside lightpushClient itself. Not needed i think anymore.

tests/waku_discv5/test_waku_discv5.nim
chaitanyaprem: well, without identify happening there is no way to know if peer supports PeerExchange as that is not included as part of peer's ENR. 
Only via identify PeerExchange protocol support is included in the peer's capability. 
Hence had to add this in order to ensure test doesn't fail. 

Odd thing is i am not sure how this test was passing before.

waku/node/waku_node.nim
chaitanyaprem: Oh, where is the `kernel_api` defined? 

or do you mean `node/api` ?

waku/waku_rendezvous/client.nim
chaitanyaprem: Since selectPeer returns an Option[RemotePeerInfo] , this wont work.

waku/waku_rendezvous/client.nim
NagyZoltanPeter: IDK if we have peer manager's measure on actual number of peers per protocol already?
It sounds something work for peer manager not for protocol client itself. WDYT?

waku/waku_mix/protocol.nim
NagyZoltanPeter: ...btw, looking at it again, does it necessary to be public proc?

waku/waku_rendezvous/client.nim
chaitanyaprem: Yea, for now just added a check if peer doesn't exist only then increment the counter.

apps/chat2mix/chat2mix.nim
Ivansete-status: Yes, the most idiomatic way is having an `$` operator defined for the given type so that it compiles well.
For example: https://github.com/waku-org/nwaku/blob/31e1a81552d72cd07133754e1953639da49ec954/waku/waku_filter_v2/rpc.nim#L94-L96

chaitanyaprem: ah, this was not a compilation issue but rather the chat ui is printing messages received not properly. 

this is the output when using $pb

`(timestamp: 1763707377, nick: "prem", payload: @[109, 115, 103, 49])
`

but when using pb.toString() , it prints like below. So, i think it is better to leave it like this for better UX.

`<Nov 21, 12:19> prem: test
`

waku/waku_rendezvous/protocol.nim
chaitanyaprem: Thanks a lot for the suggestion of using `procCall` , somehow it did not strike me that i can use it.
It worked and pushed the change.

waku/waku_rendezvous/protocol.nim
chaitanyaprem: this is the error i am getting if i remove this

```
/Users/prem/Code/nwaku/vendor/nim-libp2p/libp2p/protocols/rendezvous/rendezvous.nim(148, 18) Error: type mismatch
Expression: items(rdv.registered)
  [1] rdv.registered: OffsettedSeq[rendezvous.RegisteredData]

```

Note that this is already exported from `vendor/nim-libp2p/libp2p/protocols/rendezvous/rendezvous.nim
` 

Don't know how to resolve this.

Ivansete-status: This seems to be a type shadowing.
It can be solved by explicitly importing `libp2p/utils/offsettedseq` in `waku_node.nim`

apps/chat2mix/chat2mix.nim
Ivansete-status: Sorry for being picky but I think we can solve that with something like:
```nim
proc `$`(message: Chat2Message): string =
  return "Chat2Message(timestamp: " & $message.timestamp &
    ", nick: " & message.nick & ", payload: " & $message.payload & ")"
```

Or, maybe better, just change current

`proc toString*(message: Chat2Message): string =`
with
`proc `$`*(message: Chat2Message): string =`

examples/lightpush_mix/lightpush_publisher_mix.nim
Ivansete-status: You are right sorry. Disregard my previous comment :)

waku/waku_mix/protocol.nim
```diff
@@ -152,25 +174,26 @@ proc new*(
...
   )
   if bootnodes.len < mixMixPoolSize:
-    warn "publishing with mix won't work as there are less than 3 mix nodes in node pool"
-  let initTable = toMixNodeTable(bootnodes)
+    warn "publishing with mix won't work until atleast 3 mix nodes in node pool"
+  let initTable = processBootNodes(bootnodes, peermgr)
+
   if len(initTable) < mixMixPoolSize:
-    warn "publishing with mix won't work as there are less than 3 mix nodes in node pool"
+    warn "publishing with mix won't work until atleast  3 mix nodes in node pool"
```
```suggestion
    warn "publishing with mix won't work until at least  3 mix nodes in node pool"
```

waku/waku_mix/protocol.nim
```diff
@@ -152,25 +174,26 @@ proc new*(
...
   let nodeMultiAddr = MultiAddress.init(nodeAddr).valueOr:
     return err("failed to parse mix node address: " & $nodeAddr & ", error: " & error)
   let localMixNodeInfo = initMixNodeInfo(
     peermgr.switch.peerInfo.peerId, nodeMultiAddr, mixPubKey, mixPrivKey,
     peermgr.switch.peerInfo.publicKey.skkey, peermgr.switch.peerInfo.privateKey.skkey,
   )
   if bootnodes.len < mixMixPoolSize:
-    warn "publishing with mix won't work as there are less than 3 mix nodes in node pool"
-  let initTable = toMixNodeTable(bootnodes)
+    warn "publishing with mix won't work until atleast 3 mix nodes in node pool"
```
```suggestion
    warn "publishing with mix won't work until at least 3 mix nodes in node pool"
```

waku/waku_rendezvous/protocol.nim
chaitanyaprem: that worked. Thanks!

waku/node/waku_node.nim
Ivansete-status: Yes, kernel_api. It would be interesting having all the discovery mechanism gathered together under the same folder ( we have `discovery` folder .) That would help to move discovery modules into a separate repo from nwaku. Nevertheless, no need to rush

waku/waku_rendezvous/protocol.nim
```diff
@@ -1,117 +1,92 @@
...
 
-  requestInterval: timer.Duration
-  periodicRequestFut: Future[void]
+const MaximumNamespaceLen = 255
+
+method discover*(
+    self: WakuRendezVous, conn: Connection, d: Discover
+) {.async: (raises: [CancelledError, LPStreamError]).} =
+  # Override discover method to avoid collect macro generic instantiation issues
+  # TODO figure out if we can use parent generic discover
```
Ivansete-status: What do you mean?

waku/waku_rendezvous/waku_peer_record.nim
```diff
@@ -0,0 +1,74 @@
...
+  ?pb.getRequiredField(1, record.peerId)
+  ?pb.getRequiredField(2, record.seqNo)
+  discard ?pb.getRepeatedField(3, record.addresses)
+
+  if record.addresses.len == 0:
+    return err(ProtoError.RequiredFieldMissing)
+
+  ?pb.getRequiredField(4, record.mixKey)
+
+  ok(record)
```
Ivansete-status: nitpick to keep some consistency
```suggestion
  return ok(record)
```

waku/waku_rendezvous/waku_peer_record.nim
```diff
@@ -0,0 +1,74 @@
...
+  pb.write(1, record.peerId)
+  pb.write(2, record.seqNo)
+
+  for address in record.addresses:
+    pb.write(3, address)
+
+  pb.write(4, record.mixKey)
+
+  pb.finish()
+  pb.buffer
```
Ivansete-status:
```suggestion
  return pb.buffer
```

waku/waku_rendezvous/protocol.nim
chaitanyaprem: forgot to remove comment

PR #3619, opened 2025-10-08
tests(store-sync): fix flaky reconciliation test on macOS by anchoring timestamps to a fixed baseTime and aligning sync window via offset

NagyZoltanPeter:
Summary\n\nFix a flaky test in tests/waku_store_sync/test_protocol.nim that intermittently failed on macOS due to timing drift between message timestamps and the reconciliation sync window.\n\nProblem / Root cause\n\n- The test constructed test-message timestamps using getNowInNanosecondTime() while the reconciliation time window was computed separately using calculateTimeRange(). On macOS, slower scheduling made the two "now" calls diverge more, leading to mismatches.\n- As a consequence, some messages expected to be "in-window" were misclassified as "out-window", producing failures such as:\n  - remoteNeeds.len was 0 while diffInWin was 18\n  - Overlaps between in-window and out-window message sets in previous attempts\n\nSolution\n\n- Anchor the entire test to a single fixed baseTime captured once at test start.\n- Define an intended sync window of [baseTime - 1s, baseTime].\n- Compute the reconciliation offset so that the effective "now" used by calculateTimeRange equals baseTime:\n  - offset = actualNow - baseTime\n- Generate messages deterministically:\n  - Place 18 in-window messages strictly inside the window with 50 ms margins.\n  - Place 20 out-window messages well before the window start to avoid overlap.\n- Use consistent int64 timestamp arithmetic to avoid signed/unsigned pitfalls.\n\nFiles changed\n\n- tests/waku_store_sync/test_protocol.nim\n\nVerification\n\n- macOS: make test tests/waku_store_sync/test_protocol.nim passes locally; reconciliation properly detects the 18 in-window messages and ignores the 20 out-window messages.\n- Recommendation: run CI matrix (macOS + Ubuntu) to validate cross-platform stability.\n\nNotes\n\n- This change is test-only and does not alter the production reconciliation logic.\n- The approach eliminates the timing race by ensuring a deterministic relationship between message timestamps and the reconciliation time window.

### DIFF
--- a/tests/waku_store_sync/test_protocol.nim
+++ b/tests/waku_store_sync/test_protocol.nim
@@ -522,31 +522,54 @@ suite "Waku Sync: reconciliation":
 
     randomize()
 
-    let nowNs = getNowInNanosecondTime()
-    let sliceStart = Timestamp(uint64(nowNs) - 700_000_000'u64)
-    let sliceEnd = nowNs
-    let stepIn = (sliceEnd.int64 - sliceStart.int64) div diffInWin
-
-    let oldStart = Timestamp(uint64(sliceStart) - outOffsetNs)
+    # Use a fixed base time to eliminate race conditions between macOS and Ubuntu
+    let baseTime = getNowInNanosecondTime()
+    
+    # The reconciliation will use: calculateTimeRange(relayJitter, syncRange)
+    # which does: now = getNow() - jitter; return [now - syncRange, now]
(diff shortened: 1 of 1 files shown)


NagyZoltanPeter:
Closing without merge, due even though it seems solve the timing issue on Mac but started to fail on Ubuntu on the exact same test case.

PR #3620, opened 2025-10-08
Bump to macos-15 in GitHub ci workflow

Ivansete-status:
Simple PR to try avoid recent macos timeouts (1h tests running without completion.)

Note: this doesn't fix the slowness in macos tests but I think we need to bump it anyways


NagyZoltanPeter (approved):
With hope it helps

PR #3622, opened 2025-10-13
Move log level from debug to info

Ivansete-status:
## Description

Start having `info` log level as the default one.

Before, we were using `debug` as the default one and the next possible log level is `trace`, which is very intense, particularly, within nim-libp2p.

Therefore, with that, we will have some more freedom to actually add "debug" logs, that we don't want to always appear, but they are not the extremely spammy `trace`.

The original reason for that came after noticing that two waku relay protocol.nim logs were relatively intense, but I don't want to move them to `trace`. See the following log distribution from status.prod and v0.36.0:
image

btw - thanks @yakimant and @jakubgs to teach about the most consuming logs :raised_hands: 

## Changes

- Generic refactor to substitute `debug "...` with `info "...`
- Move `notice` to `debug` in https://github.com/waku-org/nwaku/blob/7b234ec78ab6b5dc06bcaf5f5ed6f6754c934873/waku/waku_relay/protocol.nim#L181-L190

## Issue

### DIFF
--- a/waku/waku_relay/protocol.nim
+++ b/waku/waku_relay/protocol.nim
@@ -204,7 +204,7 @@ proc logMessageInfo*(
   let payloadSize = float64(msg.payload.len)
 
   if onRecv:
-    notice "received relay message",
+    debug "received relay message",
       my_peer_id = w.switch.peerInfo.peerId,
       msg_hash = msg_hash,
       msg_id = msg_id_short,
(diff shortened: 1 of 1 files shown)


darshankabariya (approved):
Amazing, LGTM

Ivansete-status:
@darshankabariya - kindly include this change in v0.37.0 :partying_face: 
@jakubgs - as a reminder for the upcoming future :), after v0.37.x is deployed, we will need to change the log-level to `INFO` in infra

PR #3623, opened 2025-10-14
feat: ensure waku is installable via nimble

fryorcraken:
## Issue

closes #

### DIFF
--- a/waku.nimble
+++ b/waku.nimble
@@ -32,7 +32,7 @@ requires "nim >= 2.2.4",
   "results",
   "db_connector",
   "minilru",
-  "https://github.com/vacp2p/mix#0.1.0"
+  "https://github.com/vacp2p/mix#v0.1.0"
 
 ### Helper functions
 proc buildModule(filePath, params = "", lang = "c"): bool =


fryorcraken:
>     * the CI script installs one version of nwaku via `nimble install nwaku@#someversion` and then the package requests a _conflicting_ version - this is messy

Ah yes, I think this got messy as iterated and tried various strategies.

How would you recommend the cleanest way of doing this? remove `requires waku` in `example.nimble` file and just call `nimble add waku@#someversion` ?

fryorcraken:
we have a successful build: https://github.com/waku-org/nwaku/actions/runs/18641762303/job/53141969665?pr=3623
with commit: [`5ae6b01` (#3623)](https://github.com/waku-org/nwaku/pull/3623/commits/5ae6b01cbbd5b80ff96c910fc2e390f32d54dd9a)

However, suprisingly the `run` command fails afterwards. (nimble 0.20.1)

fryorcraken:
Same issue with nimble nightly. Some dependency conflict that is not picked up by nimble build. Will review if I can solve them.

arnetheduck:
> nimble add waku@#someversion

this command is for adding a new `requires` line.

> How would you recommend the cleanest way of doing this?

What's `this` exactly? 

If `this` is just "check that the waku nimble file looks good", the easiest thing is to run `nimble check -l` or `nimble deps -l` in the top-level folder with the _main_ waku nimble file (ie not on this example), possibly removing its `bin` setting (libraries that have `bin` set take longer to install)  - this ensures that the nimble file isn't internally contradictory, ie that version resolution works for someone that will consume the "next" version when it gets released.

For day-to-day building and running, I tend to use `nimble setup -l` followed by `nimble shell` which works like a python `venv`, ie you get an isolated shell where all the paths are set up correctly. In that shell, `nim c ...` works as usual and has all the paths set up to point to the local environment.

fryorcraken:
latest blocker locally: https://github.com/status-im/nim-dnsdisc/pull/25

not sure what is happening with the CI as it works better locally.

PR #3632, opened 2025-10-21
removing mix repo as dependency and using mix from libp2p repo

chaitanyaprem:
## Description
Remove separate mix dependency and use the mix version that is part of nim-libp2p.
Had to make min version `1.14.2` as there was a fix done in mix without which waku-mix won't compile.

Deliverable: https://github.com/waku-org/nwaku/pull/3617

## Issue

closes #

### DIFF
--- a/waku/waku_mix/protocol.nim
+++ b/waku/waku_mix/protocol.nim
@@ -117,19 +117,14 @@ proc startMixNodePoolMgr*(mix: WakuMix) {.async.} =
     mix.populateMixNodePool()
  ]#
 
-proc getPeerIdFromMultiAddr*(multiAddr: string): Result[PeerId, string] =
-  let parts = multiAddr.split("/")
-  if parts.len != 7:
-    return err("Invalid multiaddress format")
-  ok(PeerId.init(parts[6]).get())
-
 proc toMixNodeTable(bootnodes: seq[MixNodePubInfo]): Table[PeerId, MixPubInfo] =
   var mixNodes = initTable[PeerId, MixPubInfo]()
   for node in bootnodes:
-    let peerId = getPeerIdFromMultiAddr(node.multiAddr).valueOr:
+    let pInfo = parsePeerInfo(node.multiAddr).valueOr:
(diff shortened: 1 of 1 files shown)


jm-clius (approved):
Will leave decision on when to merge up to you, but LGTM

waku/waku_mix/protocol.nim
```diff
@@ -117,14 +116,35 @@ proc startMixNodePoolMgr*(mix: WakuMix) {.async.} =
   heartbeat "Updating mix node pool", 5.seconds:
     mix.populateMixNodePool()
  ]#
+
+proc getPeerIdFromMultiAddr*(multiAddr: string): Result[PeerId, string] =
+  let parts = multiAddr.split("/")
+  if parts.len != 7:
+    return err("Invalid multiaddress format")
+  ok(PeerId.init(parts[6]).get())
```
NagyZoltanPeter: Can't we use existing parse functionality `parsePeerInfo`?
https://github.com/waku-org/nwaku/blob/f097c1d19d6d2a30714b1176541ae0cc4ad19dea/waku/waku_core/peers.nim#L220

I agree this is just about to strip out peerId but I would rather avoid rely on some magic number `7`.
WDYT?

NagyZoltanPeter (approved):
LGTM!!! Only one small remark I had, if you may find it useful.
Do you think we need this into release under preparation?

waku/waku_mix/protocol.nim
chaitanyaprem: Thanks, was not aware something like this existed. 
Agree that we can instead use this rather than having another logic. 
Will update the same.

fryorcraken:
> > Do you think we need this into release under preparation?
> 
> Not sure as this only changes usage of mix in libp2p rather than separate dependency. Maybe @fryorcraken can comment if this is needed.

Well, I still cannot make a clean usage of nimble with nwaku (see https://discord.com/channels/864066763682218004/1430384784294547542/1430775598745387028 on Vac Discord). So the removal of any nimble related complexity is welcome.

PR #3637, opened 2025-11-06
chore: clarify api folders

NagyZoltanPeter:
## Description
In relation to https://github.com/waku-org/nwaku/issues/3432 we identified name clashes in nwaku implementation.
As now we are implementing waku-api as the only public interface for waku, we would like to "hide" the complex raw interface of the node.
Also we refered REST API interface as waku_api, that can cause misunderstanding.
In order to clean this up, with this PR I would like to suggest a more clarify namings throughout the code base.

## Changes

This pr is not changing any functionality, every change is the result of renaming of directories and imports (and some adjustments to make everything compile).

- `waku/waku_api/rest` renamed to `waku/waku_rest/endpoint`
- `waku/node/api` renamed to `waku/node/kernel_api` - I found `kernel` as describing best that the underlying interfaces are internal and likely to be changed without notice.
- `waku/api` remains the only top level public API for waku in the future.
  - with having `waku/api.nim` waku-api will be top level import for using waku as a nimble module.  

## Issue
https://github.com/waku-org/nwaku/issues/3432

### DIFF
--- a/tests/wakunode_rest/test_rest_relay_serdes.nim
+++ b/tests/wakunode_rest/test_rest_relay_serdes.nim
@@ -3,12 +3,7 @@
 import results, stew/byteutils, unittest2, json_serialization
 import
   waku/
-    [
-      common/base64,
-      rest_api/endpoint/serdes,
-      rest_api/endpoint/relay/types,
-      waku_core,
-    ]
+    [common/base64, rest_api/endpoint/serdes, rest_api/endpoint/relay/types, waku_core]
 
 suite "Waku v2 Rest API - Relay - serialization":
   suite "RelayWakuMessage - decode":


apps/liteprotocoltester/statistics.nim
```diff
@@ -8,6 +8,8 @@ import
   results,
   libp2p/peerid
 
+from std/sugar import `=>`
```
Ivansete-status, line 11: Interesting! What is it for?

apps/liteprotocoltester/service_peer_management.nim
```diff
@@ -11,7 +11,7 @@ import
   libp2p/wire
 
 import
-  ../wakunode2/cli_args,
+  ../../tools/confutils/cli_args,
```
Ivansete-status: Everything can be referenced from the repo's root folder. We can adopt that style as much as possible
```suggestion
  tools/confutils/cli_args,
```

apps/chat2bridge/chat2bridge.nim
```diff
@@ -240,7 +240,7 @@ proc stop*(cmb: Chat2MatterBridge) {.async: (raises: [Exception]).} =
 {.pop.}
   # @TODO confutils.nim(775, 17) Error: can raise an unlisted exception: ref IOError
 when isMainModule:
-  import waku/common/utils/nat, waku/waku_api/message_cache
+  import waku/common/utils/nat, waku/waku_rest/message_cache
```
Ivansete-status: Another possible option is naming it as "rest_api", to keep the consistency with othes like "kernel_api"
```suggestion
  import waku/common/utils/nat, waku/rest_api/message_cache
```

apps/liteprotocoltester/statistics.nim
NagyZoltanPeter: Yeah, LTP was not compiling as I found during working on this PR. Some former refactor moved things. 
This operator was needed here, and I did not want to import all from sugar, which is an extensive module anyway.

apps/chat2bridge/chat2bridge.nim
NagyZoltanPeter: I can buy in this idea! I just wanted to name as less things possible as API.
But yes, rest interface is an api anyway.
I make the change!

apps/liteprotocoltester/service_peer_management.nim
NagyZoltanPeter: Addressed in https://github.com/waku-org/nwaku/pull/3637/commits/ccf11414385be5a4496ea38111988b7a7ba7ecb9

apps/chat2bridge/chat2bridge.nim
NagyZoltanPeter: Addressed in https://github.com/waku-org/nwaku/pull/3637/commits/a07e5025e9b74c5eca681f9e973f2d5ac7f690a6

PR #3642, opened 2025-11-14
use exit==destination approach for mix for better anonymity

chaitanyaprem:
## Description

As per discussions [here](https://forum.vac.dev/t/mix-implementation-in-javascript-and-js-waku/563/13?u=prem) and in general for better anonymity, we should switch to configuration where the mix exit node is same as destination peer. This way the protocol handler i.e lightpush handler will be invoked from mix layer directly in same node.  

Ref Milestone https://github.com/waku-org/pm/issues/365, but since changes are available in mix we may as well take this change now instead of waiting to implement js-mix.


apps/chat2mix/chat2mix.nim
```diff
@@ -82,6 +82,8 @@ type
   PrivateKey* = crypto.PrivateKey
   Topic* = waku_core.PubsubTopic
 
+const MixMixNodePoolSize = 4
```
Ivansete-status: A little tongue twist ;P?
```suggestion
const MaxNixNodePoolSize = 4
```

waku/waku_mix/protocol.nim
```diff
@@ -21,7 +21,7 @@ import
 logScope:
   topics = "waku mix"
 
-const mixMixPoolSize = 3
+const mixMixPoolSize = 4
```
```suggestion
const minMixPoolSize = 4
```

Ivansete-status (approved):
LGTM! Thanks for it! :100: 
Just added two nitpicks :D

PR #3644, opened 2025-11-17
chore: Introduce EventBroker and RequestBroker

NagyZoltanPeter:
## Description
As needs during implementing WakuAPI and event driven design, this PR introduces decoupling helpers that represent reactive (event driven) and proactive (request/response) patterns without tight coupling between modules.

### Problem statement:
It has been a long discussion on how to achieve loose coupling between waku modules. The current approach that uses passing callbacks around modules makes it hard to follow the data flow while introducing heavy dependencies between originally independent modules.

### Solution proposal

Introduce two macro-driven facilities to define reactive / event-driven patterns and a proactive / request-response pattern into Waku. Both implement a way that are agnostic to data source and consumers, thus allowing loose coupling and more flexible solutions in designing and refactoring Waku modules and different level APIs.

Both solution is intended to be used across library boundaries.
It can also be seen as a flexible extension point without compromising library interfaces.
Both solutions aim to define data to be exchanged between various modules as contracts.

**Benefits**

Such a data and action-driven approach allows for easy revelation of dependencies between modules only by searching the code for the use of certain data types (event or request types defined at a high level). There will always be event emitters and data requesters, and listener/provider counterparts.

**Ideas for wide use**

EventBroker and RequestBroker alone are capable of defining the module interface that is plugable very easy. 
Both concepts represent a contract between a module and user. Also this allow us for very flexible design to glue together Nim libraries.
Imagine a plugable and interchangeable discovery implementation or RLN as a library. Such an approach allows easy mock implementations of segregated testing of concepts or modules that otherwise would come with strong dependencies. 

### EventBroker

Define the event as a contract at a high level
```nim
EventBroker:
   type GreetingEvent = object
     text*: string
```

Add a listener in one or more modules where you need to be notified
```nim
let handle = GreetingEvent.listen(
  proc(evt: GreetingEvent): Future[void] {.async.} =
    echo evt.text
   )
```
emit event from any place it is appropriate:
```nim
GreetingEvent.emit(text: "hi")
```

Event propagation is fully async. `emit` is not blocking.

#### Sample of generated code

```nim
type
  SampleEvent* = object
    value*: int
    label*: string
  SampleEventListener* = object
    id*: uint64
  SampleEventListenerProc* = proc (event`gensym0: SampleEvent): Future[void] {.
      gcsafe.}
  SampleEventBroker* = ref object
    listeners: Table[uint64, SampleEventListenerProc]
    nextId: uint64
var gSampleEventBroker {.threadvar.}: SampleEventBroker

proc accessSampleEventBroker(): SampleEventBroker ...
proc registerSampleEventListener(handler`gensym3: SampleEventListenerProc): SampleEventListener ... 
proc forgetSampleEventListener(handle`gensym4: SampleEventListener) ...
proc forgetAllSampleEventListeners() ...
proc listen*(_: typedesc[SampleEvent]; handler`gensym6: SampleEventListenerProc): SampleEventListener =
  registerSampleEventListener(handler`gensym6)

proc listen*(_: typedesc[SampleEventBroker];
             handler`gensym6: SampleEventListenerProc): SampleEventListener =
  registerSampleEventListener(handler`gensym6)

proc forget*(_: typedesc[SampleEvent]; handle`gensym7: SampleEventListener) =
  forgetSampleEventListener(handle`gensym7)

proc forget*(_: typedesc[SampleEventBroker]; handle`gensym7: SampleEventListener) =
  forgetSampleEventListener(handle`gensym7)

proc forgetAll*(_: typedesc[SampleEvent]) =
  forgetAllSampleEventListeners()

proc forgetAll*(_: typedesc[SampleEventBroker]) =
  forgetAllSampleEventListeners()

proc notifySampleEventListener(callback`gensym8: SampleEventListenerProc;
                               event`gensym8: SampleEvent) {.
    async: (raises: []), gcsafe.} =
 ...
  try:
    await callback`gensym8(event`gensym8)
  except Exception:
    error "Failed to execute event listener", error = getCurrentExceptionMsg()

proc emitSampleEventValue(event`gensym8: SampleEvent): Future[void] {.
    async: (raises: [Exception]), gcsafe.} =
  ...
  for cb`gensym8 in callbacks`gensym8:
    asyncSpawn notifySampleEventListener(cb`gensym8, event`gensym8)

proc emit*(event`gensym8: SampleEvent) =
  asyncSpawn emitSampleEventValue(event`gensym8)

proc emit*(_: typedesc[SampleEvent]; event`gensym8: SampleEvent) =
  asyncSpawn emitSampleEventValue(event`gensym8)

proc emit*(_: typedesc[SampleEventBroker]; event`gensym8: SampleEvent) =
  asyncSpawn emitSampleEventValue(event`gensym8)

template emit*(_: typedesc[SampleEvent]; args`gensym8: untyped): untyped =
  asyncSpawn emitSampleEventValue(SampleEvent(args`gensym8))

template emit*(_: typedesc[SampleEventBroker]; args`gensym8: untyped): untyped =
  asyncSpawn emitSampleEventValue(SampleEvent(args`gensym8))

```

### RequestBroker

Define request type at a high level
```nim
 RequestBroker:
  type Greeting = object
    text*: string
  proc signature*(): Future[Result[Greeting, string]]
  proc signatureWithInput*(forWhom: string): Future[Result[Greeting, string]]
```

In the module, you can provide an answer to the request, add a provider implementation
```nim
Greeting.setProvider(
  proc(): Future[Result[Greeting, string]] {.async.} =
    ok(Greeting(text: "Hello"))
  )

Greeting.setProvider(
  proc(in: string): Future[Result[Greeting, string]] {.async.} =
    ok(Greeting(text: "Hello " & in))
  )
```

Any module can ask for data from an arbitrary provider.
The concept is to give a tool in developers' hands to freely organise the code without introducing many dependencies.
Requests can go across library boundaries.
```nim
let res = await Greeting.request()
echo "Empty request returns " & res.get().text
 
let resHelloWorld = await Greeting.request("world")
echo "Request with arg returns " & resHelloWorld.get().text
```

#### Sample of generated code for a request broker

```nim
type
  SimpleResponse* = object
    value*: string
  SimpleResponseProviderNoArgs = proc (): Future[
      Result[SimpleResponse, string]] {.gcsafe.}
  SimpleResponseBroker* = object
    providerNoArgs: SimpleResponseProviderNoArgs
var gSimpleResponseBroker {.threadvar.}: SimpleResponseBroker
proc accessSimpleResponseBroker(): var SimpleResponseBroker =
  gSimpleResponseBroker

proc broker*(_: typedesc[SimpleResponse]): var SimpleResponseBroker =
  accessSimpleResponseBroker()

proc setProvider*(_: typedesc[SimpleResponse];
                  handler`gensym2: SimpleResponseProviderNoArgs) =
  accessSimpleResponseBroker().providerNoArgs = handler`gensym2

proc request*(_: typedesc[SimpleResponse]): Future[
    Result[SimpleResponse, string]] =
  let provider`gensym4 = accessSimpleResponseBroker().providerNoArgs
  if provider`gensym4.isNil:
    return errorFuture[SimpleResponse]("RequestBroker(" & "SimpleResponse" &
        "): no zero-arg provider registered")
  provider`gensym4()

proc clearProvider*(_: typedesc[SimpleResponse]) =
  accessSimpleResponseBroker().providerNoArgs = nil

```

## Changes

Added 
```
waku/core/broker/
               event_broker.nim
               request_broker.nim
```

Both modules come with explanations and example usage. Please read it there.

## Disclaimer

This is the initial version of such solution. I'm sure there is room for improvement - more specific compile-time checks and features of course.
Please consider it as such. I'm happy to discuss any observations or improvement ideas. It is to help us implement a more robust Waku (Logos messaging) library/libraries.

### Notice

_Let's battle test it in Waku first, then it might be worth looking for a proper library home for greater use._

This description is about to be extended with code examples and considerations for a more detailed view...

## Issue
https://github.com/waku-org/nwaku/issues/3432

## Late add-on: MultiRequestBroker
_for fun and covering more use cases_

MultiRequestBroker is based upon the same concept of RequestBroker, but insteadof  allowing one provider, it supports multiple providers dynamically. Responses from providers will be waited and collected into seq[ValueType].
If any provider runs into an error, the whole request will report a single error. The algorithm is that all succeed or fail completely.

Example of use from test case:
```nim
MultiRequestBroker:
  type ArgResponse = object
    id*: string

  proc signatureFetch*(
    suffix: string, numsuffix: int
  ): Future[Result[ArgResponse, string]] {.async.}

...

    discard ArgResponse.setProvider(
      proc(suffix: string, num: int): Future[Result[ArgResponse, string]] {.async.} =
        ok(ArgResponse(id: suffix & "-a-" & $num))
    )

    discard ArgResponse.setProvider(
      proc(suffix: string, num: int): Future[Result[ArgResponse, string]] {.async.} =
        ok(ArgResponse(id: suffix & "-b-" & $num))
    )

    let keyed = waitFor ArgResponse.request("topic", 1)
    check keyed.get().len == 2
    check keyed.get().anyIt(it.id == "topic-a-1")
    check keyed.get().anyIt(it.id == "topic-b-1")

    ArgResponse.clearProviders()
```


waku/waku_core/broker/helper/broker_utils.nim
```diff
@@ -0,0 +1,42 @@
+import std/macros
+
+proc sanitizeIdentName*(node: NimNode): string =
+  var raw = $node
+  result = newStringOfCap(raw.len)
```
Ivansete-status: If possible, I would suggest to avoid the use of the implicit `result` to keep consistency

waku/common/broker/event_broker.nim
```diff
@@ -0,0 +1,232 @@
...
+        `exportedHandlerProcIdent` = proc(event: `typeIdent`): Future[void] {.gcsafe.}
+        `exportedBrokerTypeIdent` = ref object
+          listeners: Table[uint64, `handlerProcIdent`]
+          nextId: uint64
+
+  )
+
+  result.add(
+    quote do:
+      var `globalVarIdent` {.threadvar.}: `brokerTypeIdent`
```
Ivansete-status, line 158: We should minimize the use of global vars as much as possible. That make it more difficult when exposing libs, and allowing several ctxs within the same process.

waku/common/broker/event_broker.nim
```diff
@@ -0,0 +1,232 @@
...
+        if broker.listeners.len == 0:
+          return
+        var callbacks: seq[`handlerProcIdent`] = @[]
+        for cb in broker.listeners.values:
+          callbacks.add(cb)
+        for cb in callbacks:
+          asyncSpawn `listenerTaskIdent`(cb, event)
+
+      proc emit*(event: `typeIdent`) =
+        asyncSpawn `emitImplIdent`(event)
```
Ivansete-status, line 261: This may end up having multiple scattered places where "start and forget" happens. I believe this is risky if the `emitImplIdent` raises an uncaught exception.

waku/common/broker/event_broker.nim
```diff
@@ -0,0 +1,232 @@
...
+          await callback(event)
+        except Exception:
+          error "Failed to execute event listener", error = getCurrentExceptionMsg()
+
+      proc `emitImplIdent`(
+          event: `typeIdent`
+      ): Future[void] {.async: (raises: [Exception]), gcsafe.} =
+        let broker = `accessProcIdent`()
+        if broker.listeners.len == 0:
+          return
```
Ivansete-status, line 253: Maybe this should return an error, stating that no listeners are added?

waku/waku_core/broker/event_broker.nim
```diff
@@ -0,0 +1,232 @@
...
+          asyncSpawn `listenerTaskIdent`(cb, event)
+
+      proc emit*(event: `typeIdent`) =
+        asyncSpawn `emitImplIdent`(event)
+
+      proc emit*(_: typedesc[`typeIdent`], event: `typeIdent`) =
+        asyncSpawn `emitImplIdent`(event)
+
+      proc emit*(_: typedesc[`brokerTypeIdent`], event: `typeIdent`) =
+        asyncSpawn `emitImplIdent`(event)
```
Ivansete-status: What's the diff between these two?

waku/common/broker/event_broker.nim
```diff
@@ -0,0 +1,232 @@
...
+    quote do:
+      proc `listenerTaskIdent`(
+          callback: `handlerProcIdent`, event: `typeIdent`
+      ) {.async: (raises: []), gcsafe.} =
+        if callback.isNil:
+          return
+        try:
+          await callback(event)
+        except Exception:
+          error "Failed to execute event listener", error = getCurrentExceptionMsg()
```
Ivansete-status, line 241: Considering that the expectation for "broker" change is to become a low-level lib, I think it shouldn't log anything at all. Instead, only deal with `Result`.
Maybe we can enforce that `callback` raises no exception, i.e., `raises []`

waku/waku_core/broker/event_broker.nim
```diff
@@ -0,0 +1,232 @@
...
+          new(`globalVarIdent`)
+          `globalVarIdent`.listeners = initTable[uint64, `handlerProcIdent`]()
+        `globalVarIdent`
+
+  )
+
+  result.add(
+    quote do:
+      proc `listenImplIdent`(handler: `handlerProcIdent`): `listenerHandleIdent` =
+        if handler.isNil:
```
Ivansete-status: Super nitpick sorry, the "verbs" should have parenthesis
```suggestion
        if handler.isNil():
```

tests/common/test_event_broker.nim
```diff
@@ -0,0 +1,99 @@
...
+    flag*: bool
+
+template waitForListeners() =
+  waitFor sleepAsync(1.milliseconds)
+
+suite "EventBroker":
+  test "delivers events to all listeners":
+    var seen: seq[(int, string)] = @[]
+
+    discard SampleEvent.listen(
```
Ivansete-status, line 27: How should be managed the returned `listenerHandleIdent` type?

Ivansete-status (commented):
Thanks for the PR! 🙌 

When we were talking about event-driven approach, I was thinking about something simpler, f.e., the technique with follow within waku_relay, `w.addObserver(administrativeObserver)`.

IMHO, instead of start the implementation from a macro, I would consider that simpler approach first and then, if needed, have the macro. That way my dumb neuron would be happier :)

Besides, the usage of the macro and its resulting procs look nice.

Would it be possible to use another term, different than "broker"? That term doesn't sound explanatory to me.

Nevertheless, what concerns me the most is the usage of global vars.

Thanks again! This is a very advanced PR 🚀 !

waku/common/broker/event_broker.nim
NagyZoltanPeter: > We should minimize the use of global vars as much as possible. That make it more difficult when exposing libs, and allowing several ctxs within the same process.

Well the idea comes from chronos async. That is exactly using this technique to implement async execution queue.
Also some type of global instance is needed for such a global decoupled dispatcher of such kind. Since nim has got no easy singleton pattern, this way we have a module owned global instance that is managed and accessed through public methods.

For waku it just the same, it must run on it's own single thread, hence the `{.threadvar.}`. _Of course in case a real multithreaded implementation would require a mutexed/atomic access approach to that singleton._ 
So for waku for now it just makes no problem using it in exposed lib use.

waku/common/broker/event_broker.nim
NagyZoltanPeter: I will review, there should be no exception coming out of this system by design, exactly for this mentioned reason.

waku/waku_core/broker/event_broker.nim
NagyZoltanPeter: Added for convenience:
```nim
   SentEvent.emit( sentEvtInstance)
   ... or ...
   SentEventBroker.emit(sentEvtInstance)
```
We can decide if needed both or just one form...

waku/common/broker/event_broker.nim
NagyZoltanPeter: Yes, observer must have `raises[]`
For waku logging is ok, for low level lib we may need to think of some mechanism to allow debugging.
As event propagation is fully async there is no way to let dev know if something not ok.

tests/common/test_event_broker.nim
NagyZoltanPeter: One can store return value of listen() and can use it to remove its own listener.
This test just not using it.

waku/common/broker/event_broker.nim
NagyZoltanPeter: No it is not an error having no subscriber for an event.

waku/common/broker/request_broker.nim
```diff
@@ -0,0 +1,363 @@
...
+        `accessProcIdent`()
+
+  )
+
+  var clearBody = newStmtList()
+  if not zeroArgSig.isNil:
+    result.add(
+      quote do:
+        proc setProvider*(_: typedesc[`typeIdent`], handler: `zeroArgProviderName`) =
+          `accessProcIdent`().`zeroArgFieldName` = handler
```
fcecin, line 311: Not sure if this makes sense, but if we call `setProvider` twice (zeroarg or with arg), say two components registering the same provider on accident, it just overrides? Maybe we want to throw an error here? Or maybe an alternative `setProviderSafe` that does the check first?

EDIT: another name idea is `setProviderOnce` (not a fan of "Safe" there)

waku/common/broker/event_broker.nim
Ivansete-status: > Well the idea comes from chronos async. That is exactly using this technique to implement async execution queue.

Interesting! I might have overlooked that detail. May share where this happens in nim-chronos? ( to learn something ;P) I thought it used iterators and yielded context.

If `threadvar` can ensure thread safety in global env vars, then I'm fine :)

waku/waku_core/broker/event_broker.nim
Ivansete-status: Awesome! I think for now we can keep the smallest set of procs.
In this particular case, I'd vote to `SentEvent.emit(...)`

waku/common/broker/event_broker.nim
NagyZoltanPeter: > > Well the idea comes from chronos async. That is exactly using this technique to implement async execution queue.
> 
> Interesting! I might have overlooked that detail. May share where this happens in nim-chronos? ( to learn something ;P) I thought it used iterators and yielded context.

Oh, the [async engine itself is a thread local global var.](https://github.com/status-im/nim-chronos/blob/0646c444fce7c7ed08ef6f2c9a7abfd172ffe655/chronos/internal/asyncengine.nim#L765)

> If `threadvar` can ensure thread safety in global env vars, then I'm fine :)
Yes, threadvar makes it thread safe for sure, yet also means that this implementation of broker can't get through thread boundaries as such cannot deliver events/requests through it.

waku/waku_core/broker/event_broker.nim
NagyZoltanPeter: Ok, I can accept this as agree with having a more clean and simpler interface.

waku/waku_core/broker/event_broker.nim
```diff
@@ -0,0 +1,236 @@
...
+
+  result.add(
+    quote do:
+      var `globalVarIdent` {.threadvar.}: `brokerTypeIdent`
+  )
+
+  result.add(
+    quote do:
+      proc `accessProcIdent`(): `brokerTypeIdent` =
+        if `globalVarIdent`.isNil:
```
Ivansete-status: Super nitpick comment. Applies elsewhere
```suggestion
        if `globalVarIdent`.isNil():
```

waku/waku_core/broker/event_broker.nim
```diff
@@ -0,0 +1,236 @@
...
+        var broker = `accessProcIdent`()
+        if broker.nextId == 0'u64:
+          broker.nextId = 1'u64
+        if broker.nextId == high(uint64):
+          error "Cannot add more listeners: ID space exhausted", nextId = $broker.nextId
+          return err("Cannot add more listeners, listener ID space exhausted")
+        let newId = broker.nextId
+        inc broker.nextId
+        broker.listeners[newId] = handler
+        ok(`listenerHandleIdent`(id: newId))
```
Ivansete-status: Nitpick comment for consistency purposes. Applies elsewhere
```suggestion
        return ok(`listenerHandleIdent`(id: newId))
```

waku/common/broker/event_broker.nim
```diff
@@ -0,0 +1,236 @@
...
+        if broker.listeners.len == 0:
+          # nothing to do as nobody is listening
+          return
+        var callbacks: seq[`handlerProcIdent`] = @[]
+        for cb in broker.listeners.values:
+          callbacks.add(cb)
+        for cb in callbacks:
+          asyncSpawn `listenerTaskIdent`(cb, event)
+
+      proc emit*(event: `typeIdent`) =
```
Ivansete-status, line 260: Maybe something to consider in the future is handling possible errors in `emit`, with `Result[T, E]`

waku/common/broker/multi_request_broker.nim
```diff
@@ -0,0 +1,517 @@
...
+    return false
+  if inner[0].kind != nnkIdent or not inner[0].eqIdent("Result"):
+    return false
+  if inner[1].kind != nnkIdent or not inner[1].eqIdent($typeIdent):
+    return false
+  inner[2].kind == nnkIdent and inner[2].eqIdent("string")
+
+proc cloneParams(params: seq[NimNode]): seq[NimNode] =
+  ## Deep copy parameter definitions so they can be reused in generated nodes.
+  result = @[]
```
Ivansete-status, line 100: Could we avoid the use of result in these non-macro procs? Applies elsewhere :)

tests/common/test_event_broker.nim
```diff
@@ -0,0 +1,99 @@
...
+      proc(evt: SampleEvent): Future[void] {.async: (raises: []).} =
+        inc counter
+    )
+
+    discard SampleEvent.listen(
+      proc(evt: SampleEvent): Future[void] {.async: (raises: []).} =
+        inc(counter, 2)
+    )
+
+    SampleEvent.forget(handleA.get())
```
Ivansete-status: For now, I would start without the `forget` option. Maybe not needed.
In any case, I would like to avoid having to use the `discard` term so often because it creates some confusion to the reader.

It would be great if we could have something like the following, so that we avoid `discard`. Similar suggestion applies to the `setProvider`:
```suggestion
    let handleA = proc(evt: SampleEvent): Future[void] {.async: (raises: []).} =
                                inc counter
    SampleEvent.listen( handleA )
    SampleEvent.forget ( handleA )
```

NagyZoltanPeter:
> The PR is awesome indeed! Nevertheless, not approving yet as aiming to see some more simplifications, if possible. For example, it would be great having comments around each operation, e.g. `setProvider`, `emit`, `request`, `listen`, `forget`.

It's ok. I will add some more description into the headers as all these functions are on the fly generated.

> Particularly, I can't quite get the need for `waku/waku_core/broker/multi_request_broker.nim`. Surely something I'm missing

This is to cover the last segment of the possible use cases for such a decoupled use or RPC.
Also it is very handy to apply for example to node health check. It is easy to collect health info from each protocol as all mounted protocols can answer to the same single request.

> I'm adding some more nitpick comments that I hope you find useful :)
Thanks! I will keep track of them!

tests/common/test_event_broker.nim
NagyZoltanPeter: Hm, I think `forget` is essential. Maybe I need to find a better name like `dropListener` or similar.
It is essential because if you have a component that has limited lifetime, you must take care of such externals as callbacks given here and there due it is easy to run into using deleted resources through lambda captures or the other way around not get freed by GC as such callback holds reference to it.
Think about that the broker is global and has no idea about the validity of those callbacks.

About the `discard`s they are just used for the test, representing I don't want to remove my listener/provider explicitly, as it will be removed all at once.

waku/common/broker/multi_request_broker.nim
NagyZoltanPeter: I don't get this. These are the generator macro helper procs, not exported at all.

waku/common/broker/event_broker.nim
NagyZoltanPeter: Hm, `emit` is a fire and forget type of call. No such case that can run into any result or error here.
The event emitter shall not care about that anybody is listening or not.

tests/common/test_event_broker.nim
Ivansete-status: > About the discards they are just used for the test, representing I don't want to remove my listener/provider explicitly, as it will be removed all at once.

Okay, I'd prefer avoid the `discard` term, if possible. When I see `discard` I use to think about a `Result` that is not properly handled

waku/common/broker/multi_request_broker.nim
Ivansete-status: Yes, we usually avoid that implicit `result` type. The comment was for consistency purposes

tests/common/test_event_broker.nim
NagyZoltanPeter: Hm, Nim, unlike other languages that don't require it explicitly to handle return value or do say explicitly `I don't care` aka `discard`, it sounds me that discard means it is intentionally not handled (also allows compiler to optimize one memory copy away).

waku/common/broker/request_broker.nim
NagyZoltanPeter: Hm... I'm ok with returning error if already having a provider and thus force explicit `clearProvider` to express intent.

fcecin: That seems safer. An idea would be to add an optional `force: bool = false` param to setProvider instead of forcing a previous clearProvider call (+1 line).

fcecin: I'm thinking about Result[T, E] being returned from setProvider:

          if not `accessProcIdent`().`zeroArgFieldName`.isNil():
            return err("Zero-arg provider already set")

I'm wondering if this isn't a code defect (i.e. a panic, a  ̶s̶t̶a̶t̶i̶c̶ runtime assert)? That is, if you don't check, or discard the Result, then you're potentially masking a static component wiring bug. And doing something like "setProvider.isOkOr:" everywhere looks like an anti-pattern.

NagyZoltanPeter: Hm, it's a good idea having it run to assertion instead of error result, but as I'm try reasoning vica-versa, I would stick to Result better. Reason is to keep it closer to `MultiRequestBroker` and also with assertion it might be harder to handle dynamic initializatoin orders. I'm thinking of this component will provider answer to request if no-body registered before.
As a user of the code you can still say `assert T.setProvider(...).isOk()` or similar.
Also I would not use such a `force` flag which makes code harder to read.

@fcecin  WDYT, keep the `setProvider():Result` signature and let see what comes out during use?

fcecin: > As a user of the code you can still say `assert T.setProvider(...).isOk()` or similar.

Ah, good point! One could even write a Nim template for this, downstream. And you could easily refactor all the code in a project (global find and replace or regex) to replace all setProvider with the template later at once as projects that use the broker decide their own style.

Similarly, you could write a force/overwrite template in some user project that does clear then set.

You're 100% right and Result is better.

NagyZoltanPeter: Thanks! Honestly, I don't think that this one the other is better, just a different choice of design.
I have in mind that while you stick to passing around callback pointers, you inherently enforce an initialization order by design. It's good to decouple such deps, but it comes with other types of issues to watch for. As you are not enforcing the existence of a certain object instance, it is harder to detect why you will not receive an event or why the request is not coming and such. There is no foolproof solution for that, I believe. 
But c'mon, we are building a truly complex p2p distributed system which has the same problem domain on a larger scale, so we shall be able to handle it in our code. :-)

fcecin: Just having some more fun thinking here (this comment is not a review action item).

Mostly I'm already thinking about this becoming a standalone Nim library later (because it's cool and everyone instantly loves it). Then people will argue about Defect, etc.

In setProvider, IMO returning Result can be considered to be objectively better than either nothing, or raise exception, or raise defect/assert, in the sense of keeping the current layer of broker functionality as a set of primitives upon which you can quickly layer some opinionated component wiring discipline (that has some understanding of what e.g. "safety" or "testability" or "monitoring" means, which is endless, fun discussion, which can be pushed to the outside of the scope of a `nim-broker` library of primitives IMO, or at least into a separate "broker_helpers.nim" file or somesuch with extra templates etc).

By "primitives" here I don't mean "minimal" set of features, just that all the features belong to the same logical level of being base functionality, e.g. MultiRequestBroker can be argued as whether it is part of a minimalistic design or not, but it's a primitive -- it can't be expressed from the other primitives (?).

And user code, which uses the primitives then exports something else upwards, can do that (e.g. add opinions on "safety" or whatever) very simply, using a handful of templates which can e.g. raise the Result into a panic/assert, for example, if that's what is wanted at some point. If setProvider for example would return nothing, as a primitive, if wouldn't allow that, and if it were to force an assert/defect opinion (my original suggestion), then it would just remove use cases from the primitives (not so primitive anymore; less good).

I'm fully onboard with keeping the base broker as crisp as possible: more like primitives, and giving room for different sets of utilities or interpretations to be put on top of it, in addition to expanding the primitives over time. If the primitives give the info (e.g. return Result) then layered expansion by downstream code is enabled.

NagyZoltanPeter: Yeah, completely agree. Actually possible further features based upon such small features are endless. I have some of them in mind - but they are not applicable on our current needs. Yet I'm in favour of single responsibility and open-close principles. So, agree with Ivan, to keep the interface as limited and easy to reason about as possible.

But for example, if logging were injectable (without this module directly depend on chronicles), I would add some debugging help later, that can reveal misconfigurations and could track who is emitting, who is listening (same for req/resp). Such things.

Still this code will not prevent anyone from having endless loops if emitting events in a round from the listeners.
I was thinking about such a compile-time check, but dropped it as it would make the code much more complicated.

But still let's start using it and reveal its weaknesses and applicability first.

NagyZoltanPeter:
> ```
> let handle = GreetingEvent.listen(
>   proc(evt: GreetingEvent): Future[void] {.async.} =
>     echo evt.text
>    )
> ```
> 
> So what it means when we want to implement `Waku API` events: https://github.com/waku-org/specs/blob/master/standards/application/waku-api.md#event-source
> 
> ```
>   MessageEvents:
>     type: event_emitter
>     description: "Event source for message-related events"
>     events:
>       "message:sent":
>         type: MessageSentEvent
>       "message:error":
>         type: MessageErrorEvent
>       "message:propagated":
>         type: MessagePropagatedEvent
> ```
> 
> then in this case, you could have two approaches:
> 

> 2. There is one event emitter per event types:
> 
> ```
> let handle = MessageSentEvent.listen(
>   proc(evt: MessageSentEvent): Future[void] {.async.} =
>     echo evt.requestId
>    )
> let handle = MessageErrorEvent.listen(
>   proc(evt: MessageErrorEvent): Future[void] {.async.} =
>     echo evt.requestId
>    )
> ```
> 
> I think (2) is better devex, because the developer does not need to worry about all the variants when there they only want to listen to one.

Yes, I've gone for the second approach already.
```nim
EventBroker:
  # Event emitted when a message is sent to the network
  type MessageSentEvent* = object
    requestId*: RequestId
    messageHash*: string

EventBroker:
  # Event emitted when a message send operation fails
  type MessageErrorEvent* = object
    requestId*: RequestId
    messageHash*: string
    error*: string

EventBroker:
  # Confirmation that a message has been correctly delivered to some neighbouring nodes.
  type MessagePropagatedEvent* = object
    requestId*: RequestId
    messageHash*: string
```
I do think this results a cleaner code on the receiver/listener side, so the case switch is already done out of the box, one just needs to write the logic for the specific case.

PR #3650, opened 2025-11-27
feat(tests): robustify waku_rln_relay test utils

fcecin:
## Description

If `ETH_FROM` or `ETH_PASSWORD` env vars are set before running unit tests (for example, you set these in `~/.bashrc` ...🤦‍♀️) the RLN tests loop forever with 100% CPU due to a failed forge script invocation. This leaves an orphaned anvil instance which hogs the localhost:8540 port, which makes the next anvil launch attempt to loop forever due to port already in use.

## Changes

* unset `ETH_FROM` and `ETH_PASSWORD` before running `forge script` so that it isn't misconfigured and fails
* fix deployTestToken forge error reporting
* robustify runAnvil proc w.r.t. any failed anvil launch

## Workaround

Don't set `ETH_FROM` and `ETH_PASSWORD` in the environment you're using to run unit tests.

## Issue

Maintenance Y2025H2 #3483

### DIFF
--- a/tests/waku_rln_relay/utils_onchain.nim
+++ b/tests/waku_rln_relay/utils_onchain.nim
@@ -82,6 +82,10 @@ proc getForgePath(): string =
   forgePath = joinPath(forgePath, ".foundry/bin/forge")
   return $forgePath
 
+template execForge(cmd: string): tuple[output: string, exitCode: int] =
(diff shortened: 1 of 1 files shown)


stubbsta (approved):
Thanks for this, LGTM!

PR #3652, opened 2025-11-29
fix: remove ENR cache from peer exchange  (#3578)

### LINKED ISSUE #3578: fetch peers from local peerStore instead of caching them as a temporary fix
### Background
@igor-sirotin did a great analysis (thanks for it 🥳) about what is happening:

I have a nwaku bootnode running.

Then I start a status-backend. 
The container gets some IP 172.20.0.6.
The go-waku instance selects a random port 22333
We use an ephemeral private key for waku, so it has a random PeerID
Let's call the above ENR-1

Here status-backend: 
✅ statically connects to the bootnode
✅ can be discovered through peer exchange

Now, keeping the bootnode running, I restart the status-backend
it gets the same IP 172.20.0.6
it selects a random port 33444
new ephemeral private key, so a new PeerID
Let's call the above ENR-2

The restarted status-backend:
✅ statically connects to the bootnode
❌ can not be discovered through peer exchange

This happens because the bootnode continues to share ENR-1, and does not share ENR-2.  So other backends get (through peer exchange with bootnode) an outdated ENR, try to connect to it and obviously fail.

After waiting for ~10 min, the bootnode forgets ENR-1 and starts to advertise ENR-2. Then the status-backend can be discovered and connected to

### Details

In order to enhance that, we need to apply @fryorcraken and @chaitanyaprem suggestions.
- Avoid [this poll](https://github.com/waku-org/nwaku/blob/48e9ece5446e02a6f66d34dc23e007770d49154e/waku/waku_peer_exchange/protocol.nim#L242) and instead ~, update the local peer-exchange cache with the recently new discovered/connected peers.~  whenever a request for peers is received fetch data from peerStore using getReachablePeers, filter it out and then respond.
~- Also, is interesting to keep a regular poll to check peers' liveness. If a counter-party peer doesn't reply properly in 10 minutes, then remove it from the peer-exchange cache.~
  - note: maybe this last point is not needed if we are able to reliably detect when a peer disconnects from us.

fcecin:
## Description

This PR removes the ENR cache and the worker loop that fed it in WakuPeerExchange, and replaces it with a direct query to the PeerStore. To avoid slowdowns, the new `getEnrsFromStore` doesn't allocate any memory other than the resulting sequence of ENR records to the caller.

## Changes

* remove WakuPeerExchange.enrCache
* add forEnrPeers to support fast PeerStore search
* add getEnrsFromStore
* fix peer exchange tests

## Issue

closes #3578


waku/waku_peer_exchange/protocol.nim
```diff
@@ -79,61 +76,46 @@ proc respondError(
...
-  return shuffledCache[0 ..< min(shuffledCache.len.int, numPeers.int)]
-
-proc poolFilter*(cluster: Option[uint16], peer: RemotePeerInfo): bool =
-  if peer.origin != Discv5:
-    trace "peer not from discv5", peer = $peer, origin = $peer.origin
+proc poolFilter*(
+    cluster: Option[uint16], origin: PeerOrigin, enr: enr.Record, log: bool = false
+): bool =
+  if origin != Discv5:
+    if log:
```
Ivansete-status: Maybe not needed?
I mean, we already have the log-level for that. We can use `trace` in all this proc's logs
```suggestion
```

waku/waku_peer_exchange/protocol.nim
```diff
@@ -79,61 +76,46 @@ proc respondError(
...
-    await sleepAsync(1.seconds)
-
-  heartbeat "Updating px enr cache", CacheRefreshInterval:
-    wpx.populateEnrCache()
+proc getEnrsFromStore(
+    wpx: WakuPeerExchange, numPeers: uint64
+): seq[enr.Record] {.gcsafe.} =
+  # Reservoir sampling (Algorithm R)
+  var i = 0
+  let k = if numPeers > int.high.uint64: int.high else: numPeers.int
```
Ivansete-status:
```suggestion
  let k = min(MaxPeersCacheSize, numPeers.int)
```
Notice that `MaxPeersCacheSize` still exist and we might need to rename it to another more appropriate name

waku/waku_peer_exchange/protocol.nim
```diff
@@ -79,61 +76,46 @@ proc respondError(
...
+  var enrs = newSeqOfCap[enr.Record](min(k, enrStoreLen))
+  wpx.peerManager.switch.peerStore.forEnrPeers:
+    if peerConnectedness == CannotConnect:
+      continue
+    if not poolFilter(wpx.cluster, peerOrigin, peerEnrRecord):
+      continue
+    if i < k:
+      enrs.add(peerEnrRecord)
+    else:
+      let j = rand(i)
```
Ivansete-status, line 118: Maybe :)?
```suggestion
    else:
      # Add some randomness
      let j = rand(i)
```

waku/waku_peer_exchange/protocol.nim
```diff
@@ -79,61 +76,46 @@ proc respondError(
...
+  # Reservoir sampling (Algorithm R)
+  var i = 0
+  let k = if numPeers > int.high.uint64: int.high else: numPeers.int
+  let enrStoreLen = wpx.peerManager.switch.peerStore[ENRBook].len
+  var enrs = newSeqOfCap[enr.Record](min(k, enrStoreLen))
+  wpx.peerManager.switch.peerStore.forEnrPeers:
+    if peerConnectedness == CannotConnect:
+      continue
+    if not poolFilter(wpx.cluster, peerOrigin, peerEnrRecord):
+      continue
```
Ivansete-status, line 113: Maybe we can add some debug logs
```suggestion
    if peerConnectedness == CannotConnect:
      debug "Could not retrieve ENR because cannot connect to peer", remotePeerId = ...
      continue
    poolFilter(wpx.cluster, peerOrigin, peerEnrRecord).isOkOr:
      debug "Could not get ENR because no peer matched pool", error = error
      continue
```

waku/waku_peer_exchange/protocol.nim
```diff
@@ -79,61 +76,46 @@ proc respondError(
...
-
-  # return numPeers or less if cache is smaller
-  return shuffledCache[0 ..< min(shuffledCache.len.int, numPeers.int)]
-
-proc poolFilter*(cluster: Option[uint16], peer: RemotePeerInfo): bool =
-  if peer.origin != Discv5:
-    trace "peer not from discv5", peer = $peer, origin = $peer.origin
+proc poolFilter*(
+    cluster: Option[uint16], origin: PeerOrigin, enr: enr.Record, log: bool = false
+): bool =
```
Ivansete-status: Instead of bool, it can return `Result[void, string]`, which can be a little bit more interesting from caller PoV

fcecin:
> > NOTE: the ci.yml workflow is broken, likely due to the project name change.
> > It was blowing up on line 135 of ci.yml. I changed it to:
> > ```
> > uses: logos-messaging/nwaku/.github/workflows/container-image.yml@master
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Now it blows up on line 140, which still references waku.org.
> 
> Can you do a dedicated PR to fixing the CI so we can get it merged asap (feel free to ping me for review)?

On it

waku/node/peer_manager/waku_peer_store.nim
```diff
@@ -227,3 +227,14 @@ proc getPeersByCapability*(
...
+
+template forEnrPeers*(peerStore: PeerStore, body: untyped) =
+  let enrBook = peerStore[ENRBook]
+  let connBook = peerStore[ConnectionBook]
+  let sourceBook = peerStore[SourceBook]
+  for pid, enrRecord in tables.pairs(enrBook.book):
+    let peerId {.inject.} = pid
+    let peerConnectedness {.inject.} = connBook.book.getOrDefault(pid, NotConnected)
+    let peerOrigin {.inject.} = sourceBook.book.getOrDefault(pid, UnknownOrigin)
+    let peerEnrRecord {.inject.} = enrRecord
```
NagyZoltanPeter: Congrats! I do like this solution. Only I would prefer that `{.inject.}` variables to be named in the template arguments rather than magically appear in the `body` code. This makes reading harder.


```suggestion
template forEnrPeers*(peerStore: PeerStore, peerId, peerConnectedness, peerOrigin, peerEnrRecord, body: untyped) =
  let enrBook = peerStore[ENRBook]
  let connBook = peerStore[ConnectionBook]
  let sourceBook = peerStore[SourceBook]
  for pid, enrRecord in tables.pairs(enrBook.book):
    let peerId {.inject.} = pid
    let peerConnectedness {.inject.} = connBook.book.getOrDefault(pid, NotConnected)
    let peerOrigin {.inject.} = sourceBook.book.getOrDefault(pid, UnknownOrigin)
    let peerEnrRecord {.inject.} = enrRecord
```

PR #3657, opened 2025-12-05
fix: store protocol issue in v0.37.0

darshankabariya:
resolving the issue reported by DST after v0.37.0 beta release. https://www.notion.so/Get-Store-Messages-Issue-2bb8f96fb65c80bc8e0fdb7801cd0c79

### DIFF
--- a/waku/factory/conf_builder/waku_conf_builder.nim
+++ b/waku/factory/conf_builder/waku_conf_builder.nim
@@ -30,6 +30,8 @@ import
 logScope:
   topics = "waku conf builder"
 
+const DefaultMaxConnections* = 150
+
 type MaxMessageSizeKind* = enum
   mmskNone
   mmskStr
@@ -585,10 +587,17 @@ proc build*(
       warn "Peer persistence not specified, defaulting to false"
       false
 
-  let maxConnections = builder.maxConnections.get()
-  if maxConnections < 150:
-    warn "max-connections less than 150; we suggest using 150 or more for better connectivity",
-      provided = maxConnections
+  let maxConnections =
+    if builder.maxConnections.isSome():
+      builder.maxConnections.get()
+    else:
+      warn "Max connections not specified, defaulting to DefaultMaxConnections",
+        default = DefaultMaxConnections
+      DefaultMaxConnections
+
+  if maxConnections < DefaultMaxConnections:
+    warn "max-connections less than DefaultMaxConnections; we suggest using DefaultMaxConnections or more for better connectivity",
+      provided = maxConnections, recommended = DefaultMaxConnections
 
   # TODO: Do the git version thing here
   let agentString = builder.agentString.get("nwaku")


waku/factory/conf_builder/waku_conf_builder.nim
```diff
@@ -588,12 +585,16 @@ proc build*(
...
-      builder.maxConnections.get()
-    else:
-      warn "Max Connections was not specified, defaulting to 300"
-      300
+  var maxConnections: int
+  if builder.maxConnections.isSome():
+    var mc = builder.maxConnections.get() # mutable to enforce minimum
+    if mc < 300:
+      warn "max-connections less than 300; using default 300", provided = mc
+      mc = 300
```
Ivansete-status: I think is better not to force that and give freedom to the user

waku/factory/waku_conf.nim
```diff
@@ -231,8 +231,19 @@ proc validateNoEmptyStrings(wakuConf: WakuConf): Result[void, string] =
 
   return ok()
 
+proc validateMaxConnectionsAndRatio(wakuConf: WakuConf): Result[void, string] =
+  if wakuConf.maxConnections < 300:
+    return
+      err("max-connections must be at least 300, provided " & $wakuConf.maxConnections)
+  if wakuConf.relayServiceRatio != "50:50":
+    return err(
+      "relay-service-ratio must be exactly 50:50, provided " & wakuConf.relayServiceRatio
```
Ivansete-status: Similarly, it is better to give freedom to dev to use another setup. We bring a good defaults from our PoV and that should suffice :D

tools/confutils/cli_args.nim
```diff
@@ -206,22 +206,17 @@ type WakuNodeConf* = object
     .}: bool
 
     maxConnections* {.
-      desc: "Maximum allowed number of libp2p connections.",
-      defaultValue: 50,
+      desc:
+        "Maximum allowed number of libp2p connections. (Default: 300) can't set it to less than 300",
+      defaultValue: 300,
```
Ivansete-status: Shall we set 150 considering the latest comments from DST/Farooq?

Ivansete-status (commented):
Some comments so far. Thanks!

waku/waku_store/client.nim
```diff
@@ -79,18 +87,34 @@ proc query*(
...
+    let connection = (await self.peerManager.dialPeer(peer, WakuStoreCodec)).valueOr:
+      waku_store_errors.inc(labelValues = [DialFailure])
+      warn "failed to dial store peer, trying next"
+      lastError = StoreError(kind: ErrorCode.PEER_DIAL_FAILURE, address: $peer)
+      continue
 
-  return await self.sendStoreRequest(request, connection)
+    let res = await self.sendStoreRequest(request, connection)
+    if res.isOk():
+      return res
```
Ivansete-status: Maybe?
```suggestion
    let res = (await self.sendStoreRequest(request, connection)).isErrOr:
      return value
```

waku/factory/conf_builder/waku_conf_builder.nim
darshankabariya: Update !!

waku/factory/waku_conf.nim
darshankabariya: Update it !!

tools/confutils/cli_args.nim
darshankabariya: Update it

waku/waku_store/client.nim
darshankabariya: Update it

waku/factory/conf_builder/waku_conf_builder.nim
```diff
@@ -588,12 +585,15 @@ proc build*(
...
-      300
+  var maxConnections: int
+  if builder.maxConnections.isSome():
+    maxConnections = builder.maxConnections.get()
+    if maxConnections < 150:
+      warn "max-connections less than 150; we suggest using 150 or more for better connectivity",
+        provided = maxConnections
+  else:
+    warn "Max Connections was not specified, defaulting to 150"
+    maxConnections = 150
```
Ivansete-status: I think we shouldn't add any logic in this point. Instead, just assign the value coming from `tools/confutils/cli_args.nim`

waku/waku_store/client.nim
```diff
@@ -79,18 +87,33 @@ proc query*(
...
 
-  let peer = self.peerManager.selectPeer(WakuStoreCodec).valueOr:
+  # Get all available store peers
+  var peers = self.peerManager.switch.peerStore.getPeersByProtocol(WakuStoreCodec)
+  if peers.len == 0:
     return err(StoreError(kind: BAD_RESPONSE, cause: "no service store peer connected"))
 
-  let connection = (await self.peerManager.dialPeer(peer, WakuStoreCodec)).valueOr:
-    waku_store_errors.inc(labelValues = [DialFailure])
+  # Shuffle to distribute load and limit retries
```
Ivansete-status, line 101: Where is the shuffle happening?

waku/factory/conf_builder/waku_conf_builder.nim
darshankabariya: Thanks for showing me. pls update it.

waku/waku_store/client.nim
darshankabariya: The logic behind shuffling is randomization; it distributes the load and limits retries.

getPeerByProtocol always returns a list of peers. If it is called without shuffling, the peers are returned in the same order every time. Here, we only try 5 peers. Without shuffling, every node would attempt to connect to the same first 5 peers, which is a problem because those peers have limited connections and eventually start rejecting new connections.

The shuffle function works well by distributing the load across all stored peers. 

@AlbertoSoutullo

AlbertoSoutullo: `Without shuffling, every node would attempt to connect to the same first 5 peers`
Is this true? Because this means that every node has the same `peerStore` and I doubt that is the case. Or am I missing something? Also, I don't see how we are calling `getPeerByProtocol` with shuffle.
Sorry to jump in here but I was tagged c:

Ivansete-status (approved):
LGTM! Thanks for it! 🙌

PR #3662, opened 2025-12-11
chore: Update anvil config

stubbsta:
## Description
The ci tests use Foundry Anvil as a testnet. The current config needs to be updated to be representative of Linea Sepolia as well as adding a config to bypass a anvil only scenario where the gas price doubles with each registration call. 

## Changes
Update params in utils_onchain.nim: runAnvil() to represent Linea Sepolia chain settings where the RLN contract is currently deployed:
- `base-fee=7`
- `gas-fee=7`
- `gas-limit=30,000,000`

## Issue closes #
https://github.com/logos-messaging/logos-messaging-nim/issues/3643

### DIFF
--- a/tests/waku_rln_relay/utils.nim
+++ b/tests/waku_rln_relay/utils.nim
@@ -24,7 +24,6 @@ proc deployContract*(
   tr.`from` = Opt.some(web3.defaultAccount)
   let sData = code & contractInput
   tr.data = Opt.some(hexToSeqByte(sData))
-  tr.gas = Opt.some(Quantity(3000000000000))
   if gasPrice != 0:
     tr.gasPrice = Opt.some(gasPrice.Quantity)
 
--- a/tests/waku_rln_relay/utils_onchain.nim
+++ b/tests/waku_rln_relay/utils_onchain.nim
@@ -549,7 +549,7 @@ proc runAnvil*(
         "10000000000",
         "--chain-id",
         $chainId,
-        "--disable-min-priority-fee"
+        "--disable-min-priority-fee",
       ]
 
     # Add state file argument if provided


tests/waku_rln_relay/utils.nim
```diff
@@ -24,7 +24,6 @@ proc deployContract*(
   tr.`from` = Opt.some(web3.defaultAccount)
   let sData = code & contractInput
   tr.data = Opt.some(hexToSeqByte(sData))
-  tr.gas = Opt.some(Quantity(3000000000000))
```
Ivansete-status, line 27: Why? Just to learn something :)

tests/waku_rln_relay/utils_onchain.nim
```diff
@@ -539,11 +540,16 @@ proc runAnvil*(
         "--port",
         $port,
         "--gas-limit",
-        "300000000000000",
+        "30000000",
+        "--gas-price",
+        "7",
+        "--base-fee",
+        "7",
```
Ivansete-status, line 547: May kindly elaborate in the PR's description why these param changes are relevant? i.e., what they do?

stubbsta: I added this to the description

tests/waku_rln_relay/utils.nim
stubbsta: This gas limit causes the RPC call to fail when using the new Anvil config because it exceeds the gas limit set for Anvil.
This optional parameter was removed as it is not explicitly required. The Anvil node is able to automatically estimate the gas fee required for the transaction.

fcecin (approved):

PR #3665, opened 2025-12-15
chore: extend RequestBroker with supporting native and external types  and added possibility to define non-async (aka sync) requests for simplicity and performance

NagyZoltanPeter:
## Description

During development of SendAPI https://github.com/logos-messaging/logos-messaging-nim/issues/3432
It's been made clear that RequestBrokers are useful beasts but with limited to async operations can make them inconvenient and adds unnecessary overhead for cases where only a simple answer with a string or bool is needed.

Due, such requests could simplify the code hugely and can remove the needs to spread callbacks or information through chain of ownerships or constructurs.

For example: Currently almost all component must know peerManager for the sake of getting a random peer for a certain protocol. Or, from whome I need to get actual connectivity info? I would need to register a callback to PeerManager's onConnectionChange handlers.

While with simple sync request we don't need to introduce such dependencies, just able to "throw in" our information request into the void and can truely hope the right answer without knowing the component it has the info.
This also avoids adding more and more interfaces to WakuNode that currently tightly coupled with everything because of being the center of all information. 

Also found that for many cases it would be easier just define request types as native types or user types without the additional overhead of bundle it into a Request object.
Eg.:
```nim
RequestBroker(sync):
  SelectPeerRequest = PeerId
  proc signature(proto: string): Result[SelectPeerRequest, string]
```
This is also needed as without async and withouth the object bundle more performant code can be generated that eliminates deep copies. It means `sync` mode RequestBrokers can be an a real alternatives of callbacks and direct class interfaces.

## Changes

- Added parameter for `RequestBroker` macro. Now it can be used with `sync` or `async` mode. Default kept for `async`.
- `sync` mode has only effect of the provider/request signature used. 
- Added support for native POD types and user types for definition of request's type for simplicity.
   - It comes with a limitation as it would be easy to define ambigous requests of strings for example and could lead to anoying compilation errors. To circumvent this code generates as `= distinct <type>` thus creating a distinguishable new type without object bundle.
- As more tests added it become very dense, it needed to be reorganized into a structured form. 

## Notice

For now I think MultiRequestBroker must stay with async approach only, due to the nature of undefined number of providers and data to collect. But in the near future it also can be extended with support for native and user types.

### DIFF
--- a/tests/common/test_request_broker.nim
+++ b/tests/common/test_request_broker.nim
@@ -59,7 +59,7 @@ suite "RequestBroker macro (async mode)":
 
   test "zero-argument request errors when unset":
     let res = waitFor SimpleResponse.request()
-    check res.isErr
+    check res.isErr()
     check res.error.contains("no zero-arg provider")
 
   test "serves input-based providers":
@@ -96,7 +96,7 @@ suite "RequestBroker macro (async mode)":
 
   test "input request errors when unset":
     let res = waitFor KeyedResponse.request("foo", 2)
-    check res.isErr
+    check res.isErr()
     check res.error.contains("input signature")
 
   test "supports both provider types simultaneously":
@@ -115,11 +115,11 @@ suite "RequestBroker macro (async mode)":
     .isOk()
 
     let noInput = waitFor DualResponse.request()
-    check noInput.isOk
+    check noInput.isOk()
     check noInput.value.note == "base"
 
     let withInput = waitFor DualResponse.request("-extra")
-    check withInput.isOk
+    check withInput.isOk()
     check withInput.value.note == "base-extra"
     check withInput.value.count == 6
 
@@ -135,7 +135,7 @@ suite "RequestBroker macro (async mode)":
     DualResponse.clearProvider()
 
     let res = waitFor DualResponse.request()
-    check res.isErr
+    check res.isErr()
 
   test "implicit zero-argument provider works by default":
     check ImplicitResponse
@@ -146,14 +146,14 @@ suite "RequestBroker macro (async mode)":
     .isOk()
 
     let res = waitFor ImplicitResponse.request()
-    check res.isOk
+    check res.isOk()
 
     ImplicitResponse.clearProvider()
     check res.value.note == "auto"
 
   test "implicit zero-argument request errors when unset":
     let res = waitFor ImplicitResponse.request()
-    check res.isErr
+    check res.isErr()
     check res.error.contains("no zero-arg provider")
 
   test "no provider override":
@@ -177,7 +177,7 @@ suite "RequestBroker macro (async mode)":
     check DualResponse.setProvider(overrideProc).isErr()
 
     let noInput = waitFor DualResponse.request()
-    check noInput.isOk
+    check noInput.isOk()
     check noInput.value.note == "base"
 
     let stillResponse = waitFor DualResponse.request(" still works")
@@ -197,7 +197,7 @@ suite "RequestBroker macro (async mode)":
     check DualResponse.setProvider(overrideProc).isOk()
 
     let nowSuccWithOverride = waitFor DualResponse.request()
-    check nowSuccWithOverride.isOk
+    check nowSuccWithOverride.isOk()
     check nowSuccWithOverride.value.note == "something else"
     check nowSuccWithOverride.value.count == 1
 
@@ -259,7 +259,7 @@ suite "RequestBroker macro (sync mode)":
 
   test "zero-argument request errors when unset (sync)":
     let res = SimpleResponseSync.request()
-    check res.isErr
+    check res.isErr()
     check res.error.contains("no zero-arg provider")
 
   test "serves input-based providers (sync)":
@@ -296,7 +296,7 @@ suite "RequestBroker macro (sync mode)":
 
   test "input request errors when unset (sync)":
     let res = KeyedResponseSync.request("foo", 2)
-    check res.isErr
+    check res.isErr()
     check res.error.contains("input signature")
 
   test "supports both provider types simultaneously (sync)":
@@ -315,11 +315,11 @@ suite "RequestBroker macro (sync mode)":
     .isOk()
 
     let noInput = DualResponseSync.request()
-    check noInput.isOk
+    check noInput.isOk()
     check noInput.value.note == "base"
 
     let withInput = DualResponseSync.request("-extra")
-    check withInput.isOk
+    check withInput.isOk()
     check withInput.value.note == "base-extra"
     check withInput.value.count == 6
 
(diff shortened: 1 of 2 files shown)


waku/common/broker/request_broker.nim
```diff
@@ -24,19 +24,69 @@
 ##   proc signature*(arg1: ArgType, arg2: AnotherArgType): Future[Result[TypeName, string]]
 ##
 ## ```
+##
+## Sync mode (no `async` / `Future`) can be generated with:
```
Ivansete-status, line 35: I might have overlooked it but maybe interesting to make it clear when to use `sync` mode. Reading the tests is clear but if we could have explained somewhere, much better :D Maybe in enum mode definition.

waku/common/broker/request_broker.nim
```diff
@@ -24,19 +24,69 @@
...
+## ```nim
+## RequestBroker(sync):
+##   type TypeName = object
+##     field1*: FieldType
+##
+##   proc signature*(): Result[TypeName, string]
+##   proc signature*(arg1: ArgType): Result[TypeName, string]
+## ```
+##
+## Note: When the request type is declared as a POD / alias / externally-defined
```
Ivansete-status: Shall we mention what does POD stand for?

tests/common/test_request_broker.nim
```diff
@@ -31,11 +35,14 @@ RequestBroker:
     suffix: string
   ): Future[Result[DualResponse, string]] {.async.}
 
-RequestBroker:
+RequestBroker(async):
```
Ivansete-status, line 38: I wonder if we should use upper case to emphasize this is not a variable.
```suggestion
RequestBroker(Async):
```

Or maybe better, we could define a "pure" enum with all possible modes.
Something like:

```nim
Or maybe better, we could define a "pure" enum with all possible modes.
Something like:

enum ReqBrokerMode* {.pure.} =
 Sync,  ## That does x y z
 RBsync, ## That does a b c d
  Async,
  RBasync":
```

tests/common/test_request_broker.nim
```diff
@@ -196,3 +202,301 @@ suite "RequestBroker macro":
...
+
+    let res = SimpleResponseSync.request()
+    check res.isOk()
+    check res.value.value == "hi"
+
+    SimpleResponseSync.clearProvider()
+
+  test "zero-argument request errors when unset (sync)":
+    let res = SimpleResponseSync.request()
+    check res.isErr
```
Ivansete-status: Sorry, super nitpick. "verbs" with parenthesis
Applies elsewhere ;P
```suggestion
    check res.isErr()
```

waku/common/broker/request_broker.nim
```diff
@@ -49,17 +99,34 @@
...
+##   type NeedThatInfo = string
+##
+##...
+##   NeedThatInfo.setProvider(
+##     proc(): Result[NeedThatInfo, string] =
+##       ok("this is the info you wanted")
+##   )
+## let res = NeedThatInfo.request().valueOr:
+##   echo "not ok due to: " & error
+##   NeedThatInfo(":-(")
```
Ivansete-status, line 134: Why sad? Always happy hahah. Errors are learnings :)
```suggestion
##   NeedThatInfo(":-)")
```

waku/common/broker/request_broker.nim
```diff
@@ -77,7 +144,11 @@ proc errorFuture[T](message: string): Future[Result[T, string]] {.inline.} =
   fut.complete(err(Result[T, string], message))
   fut
 
-proc isReturnTypeValid(returnType, typeIdent: NimNode): bool =
+type RequestBrokerMode = enum
```
Ivansete-status, line 154: Interesting! I was thinking about that. See my comment about "pure enum"

waku/common/broker/request_broker.nim
```diff
@@ -92,6 +163,23 @@ proc isReturnTypeValid(returnType, typeIdent: NimNode): bool =
     return false
   inner[2].kind == nnkIdent and inner[2].eqIdent("string")
 
+proc isSyncReturnTypeValid(returnType, typeIdent: NimNode): bool =
+  ## Accept Result[TypeIdent, string] as the contract.
```
Ivansete-status, line 174: I assume `TypeIdent` can be any arbitrary type, can't it?

waku/common/broker/request_broker.nim
```diff
@@ -378,8 +557,14 @@ macro RequestBroker*(body: untyped): untyped =
     for paramDef in requestParamDefs:
       formalParams.add(paramDef)
 
-    let requestPragmas = quote:
-      {.async: (raises: []), gcsafe.}
+    let requestPragmas =
+      case mode
+      of rbAsync:
+        quote:
+          {.async: (raises: []), gcsafe.}
```
Ivansete-status: IIRC `async` implies `gcsafe`
```suggestion
          {.async: (raises: []).}
```

NagyZoltanPeter: True!

waku/common/broker/request_broker.nim
NagyZoltanPeter: Yes it can be! This proc is about to ensure the signature is correctly set as `Result[<response_type>, string]`
Note that native types or externally defined types will have added `distinct` keyword to avoid call resolution ambiguity - for example when you might define to different purpose requests that have response type string for both.

waku/common/broker/request_broker.nim
NagyZoltanPeter: Hehe, of course, but here it means there was an error so we are said to have no proper value back.

waku/common/broker/request_broker.nim
NagyZoltanPeter: Yeah better word it correctly.

waku/common/broker/request_broker.nim
NagyZoltanPeter: As you see it is not exported, completely internal.
when you say `RequestBroker(sync):` it is not the enum used here, just a string for the parser... :-) So it does not matter to use pure or not.

tests/common/test_request_broker.nim
NagyZoltanPeter: Hm, I would vote for lower case, which matches `.async.` pragma usage with we all better familiar with.
Also the mode enum is internal. The macro argument is parsed as strings and translated to internal enum representation.

tests/common/test_request_broker.nim
NagyZoltanPeter: Right!

waku/common/broker/request_broker.nim
NagyZoltanPeter: I added an explanation here in the description of the facility.

PR #3669, opened 2025-12-15
feat: waku api send

NagyZoltanPeter:
# Notice

This PR is still under work, do expect some additions and fixes.
For completion:
- [ ] WIP: Add a dynamic service peer manager, that register last state (result of dial and libp2p call) of the service peer and can select next one accordingly - **_Will follow up in next PR_**
- [X] WIP: Fix Event- and RequestBroker implementation to support multiple usage context within a single thread.
   Currently the brokers are thread wise global, but this leads to test failures where we instantiate many Waku node instances.
- [X] Subscribe API added to support auto - subscribe upon send.
- [x] WIP: all-round unit testing, tests are not pushed yet as they fail to compile and run in many cases.

# Description

[Waku Send API](https://github.com/logos-messaging/specs/blob/master/standards/application/waku-api.md#send-messages) implementation.

Introduce new, simple `send` interface at ~~`waku`~~`logos-messaging api` level. That takes a new abstraction call `MessageEnvelop` which embodies content_topic and payload and ephemeral information in client side.

The interface returns with a RequestId - unique to each call and enables tracking of message events - on success, or error description.
The operation is non blocking, which means at the moment of call there are very limited what information can be reported to the caller.
Therefore user can expect three events to be emitted upon the operation result.
- `MessagePropagatedEvent` - Message is delivered to the network with less confidence (is relayed or lightpush initiated with success) 
- `MessageSentEvent` - Message is validated via a store node, ensures that message is on the gossip network.
- `MessageErrorEvent` - Message could not be sent, reason is described as string.

Former - unused - DeliverManager is reworked completely to fulfill the role of `DeliveryService`. Send is supported by `SendService` that manages `DeliveryTask`s in a state machine manner and takes care of Message event reporting.

# Breaking change:

Relay publish will return error `NoPeersToPublish` instead of returning OK with 0 peer count!
This derives to REST Api use of relay and lightpush also.

## Issue

- https://github.com/logos-messaging/logos-messaging-nim/issues/3432

### DIFF
--- a/.github/workflows/ci-daily.yml
+++ b/.github/workflows/ci-daily.yml
@@ -1,0 +1,79 @@
+name: Daily logos-messaging-nim CI
+
+on:
+  schedule:
+    - cron: '30 6 * * *'
+
+env:
+  NPROC: 2
+  MAKEFLAGS: "-j${NPROC}"
+  NIMFLAGS: "--parallelBuild:${NPROC} --colors:off -d:chronicles_colors:none"
+
+jobs:
+  build:
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-22.04, macos-15]
+    runs-on: ${{ matrix.os }}
+    timeout-minutes: 45
+
+    name: build-${{ matrix.os }}
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Get submodules hash
+        id: submodules
+        run: |
+          echo "hash=$(git submodule status | awk '{print $1}' | sort | shasum -a 256 | sed 's/[ -]*//g')" >> $GITHUB_OUTPUT
+
+      - name: Cache submodules
+        uses: actions/cache@v3
+        with:
+          path: |
+            vendor/
+            .git/modules
+          key: ${{ runner.os }}-vendor-modules-${{ steps.submodules.outputs.hash }}
+
+      - name: Make update
+        run: make update
+
+      - name: Build binaries
+        run: make V=1 QUICK_AND_DIRTY_COMPILER=1 examples tools
+
+      - name: Notify Discord
+        if: always()
+        env:
+          DISCORD_WEBHOOK_URL: ${{ secrets.DISCORD_WEBHOOK_URL }}
+        run: |
+          STATUS="${{ job.status }}"
+          OS="${{ matrix.os }}"
+          REPO="${{ github.repository }}"
+          RUN_URL="https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}"
+
+          if [ "$STATUS" = "success" ]; then
+            COLOR=3066993
+            TITLE="✅ CI Success"
+          else
+            COLOR=15158332
+            TITLE="❌ CI Failed"
+          fi
+
+          curl -H "Content-Type: application/json" \
+            -X POST \
+            -d "{
+              \"embeds\": [{
+                \"title\": \"$TITLE\",
+                \"color\": $COLOR,
+                \"fields\": [
+                  {\"name\": \"Repository\", \"value\": \"$REPO\", \"inline\": true},
+                  {\"name\": \"OS\", \"value\": \"$OS\", \"inline\": true},
+                  {\"name\": \"Status\", \"value\": \"$STATUS\", \"inline\": true}
+                ],
+                \"url\": \"$RUN_URL\",
+                \"footer\": {\"text\": \"Daily logos-messaging-nim CI\"}
+              }]
+            }" \
+            "$DISCORD_WEBHOOK_URL"
+
--- a/.github/workflows/ci-nix.yml
+++ b/.github/workflows/ci-nix.yml
@@ -1,0 +1,48 @@
+name: ci / nix
+permissions:
+  contents: read
+  pull-requests: read
+  checks: write
+on:
+  pull_request:
+    branches: [master]
+
+jobs:
+  build:
+    strategy:
+      fail-fast: false
+      matrix:
+        system:
+          - aarch64-darwin
+          - x86_64-linux
+        nixpkg:
+          - libwaku
+          - libwaku-android-arm64
+          - wakucanary
+
+        exclude:
+          # Android SDK limitation
+          - system: aarch64-darwin
+            nixpkg: libwaku-android-arm64
+
+        include:
+          - system: aarch64-darwin
+            runs_on: [self-hosted, macOS, ARM64]
+
+          - system: x86_64-linux
+            runs_on: [self-hosted, Linux, X64]
+
+    name: '${{ matrix.system }} / ${{ matrix.nixpkg }}'
+    runs-on: ${{ matrix.runs_on }}
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: recursive
+
+      - name: 'Run Nix build for {{ matrix.nixpkg }}'
+        shell: bash
+        run: nix build -L '.?submodules=1#${{ matrix.nixpkg }}'
+
+      - name: 'Show result contents'
+        shell: bash
+        run: find result -type f
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -80,7 +80,7 @@ jobs:
         run: make update
 
       - name: Build binaries
-        run: make V=1 QUICK_AND_DIRTY_COMPILER=1 all tools
+        run: make V=1 QUICK_AND_DIRTY_COMPILER=1 all
 
   build-windows:
     needs: changes
--- a/Makefile
+++ b/Makefile
@@ -51,10 +51,12 @@ endif
 ##########
 ## Main ##
 ##########
-.PHONY: all test update clean
+.PHONY: all test update clean examples
 
 # default target, because it's the first one that doesn't start with '.'
-all: | wakunode2 example2 chat2 chat2bridge libwaku
+all: | wakunode2 libwaku
+
+examples: | example2 chat2 chat2bridge
 
 test_file := $(word 2,$(MAKECMDGOALS))
 define test_name
@@ -190,9 +192,9 @@ LIBRLN_BUILDDIR := $(CURDIR)/vendor/zerokit
 LIBRLN_VERSION := v0.9.0
 
 ifeq ($(detected_OS),Windows)
-LIBRLN_FILE := rln.lib
+LIBRLN_FILE ?= rln.lib
 else
-LIBRLN_FILE := librln_$(LIBRLN_VERSION).a
+LIBRLN_FILE ?= librln_$(LIBRLN_VERSION).a
 endif
 
 $(LIBRLN_FILE):
@@ -483,8 +485,13 @@ ifndef ANDROID_NDK_HOME
 endif
 
 build-libwaku-for-android-arch:
-	$(MAKE) rebuild-nat-libs CC=$(ANDROID_TOOLCHAIN_DIR)/bin/$(ANDROID_COMPILER) && \
-	./scripts/build_rln_android.sh $(CURDIR)/build $(LIBRLN_BUILDDIR) $(LIBRLN_VERSION) $(CROSS_TARGET) $(ABIDIR) && \
+ifneq ($(findstring /nix/store,$(LIBRLN_FILE)),)
+	mkdir -p $(CURDIR)/build/android/$(ABIDIR)/
+	cp $(LIBRLN_FILE) $(CURDIR)/build/android/$(ABIDIR)/
+else
+	./scripts/build_rln_android.sh $(CURDIR)/build $(LIBRLN_BUILDDIR) $(LIBRLN_VERSION) $(CROSS_TARGET) $(ABIDIR)
+endif
+	$(MAKE) rebuild-nat-libs CC=$(ANDROID_TOOLCHAIN_DIR)/bin/$(ANDROID_COMPILER)
 	CPU=$(CPU) ABIDIR=$(ABIDIR) ANDROID_ARCH=$(ANDROID_ARCH) ANDROID_COMPILER=$(ANDROID_COMPILER) ANDROID_TOOLCHAIN_DIR=$(ANDROID_TOOLCHAIN_DIR) $(ENV_SCRIPT) nim libWakuAndroid $(NIM_PARAMS) waku.nims
 
 libwaku-android-arm64: ANDROID_ARCH=aarch64-linux-android
@@ -543,7 +550,7 @@ else
 	$(error iOS builds are only supported on macOS)
 endif
 
-# Build for iOS architecture 
+# Build for iOS architecture
 build-libwaku-for-ios-arch:
 	IOS_SDK=$(IOS_SDK) IOS_ARCH=$(IOS_ARCH) IOS_SDK_PATH=$(IOS_SDK_PATH) $(ENV_SCRIPT) nim libWakuIOS $(NIM_PARAMS) waku.nims
 
--- a/flake.lock
+++ b/flake.lock
@@ -2,17 +2,17 @@
   "nodes": {
     "nixpkgs": {
       "locked": {
-        "lastModified": 1740603184,
-        "narHash": "sha256-t+VaahjQAWyA+Ctn2idyo1yxRIYpaDxMgHkgCNiMJa4=",
+        "lastModified": 1757590060,
+        "narHash": "sha256-EWwwdKLMZALkgHFyKW7rmyhxECO74+N+ZO5xTDnY/5c=",
         "owner": "NixOS",
         "repo": "nixpkgs",
-        "rev": "f44bd8ca21e026135061a0a57dcf3d0775b67a49",
+        "rev": "0ef228213045d2cdb5a169a95d63ded38670b293",
         "type": "github"
       },
       "original": {
         "owner": "NixOS",
         "repo": "nixpkgs",
-        "rev": "f44bd8ca21e026135061a0a57dcf3d0775b67a49",
+        "rev": "0ef228213045d2cdb5a169a95d63ded38670b293",
         "type": "github"
       }
     },
@@ -51,18 +51,18 @@
         "rust-overlay": "rust-overlay"
       },
       "locked": {
-        "lastModified": 1749115386,
-        "narHash": "sha256-UexIE2D7zr6aRajwnKongXwCZCeRZDXOL0kfjhqUFSU=",
-        "owner": "vacp2p",
-        "repo": "zerokit",
-        "rev": "dc0b31752c91e7b4fefc441cfa6a8210ad7dba7b",
-        "type": "github"
+        "lastModified": 1762211504,
+        "narHash": "sha256-SbDoBElFYJ4cYebltxlO2lYnz6qOaDAVY6aNJ5bqHDE=",
+        "ref": "refs/heads/master",
+        "rev": "3160d9504d07791f2fc9b610948a6cf9a58ed488",
+        "revCount": 342,
+        "type": "git",
+        "url": "https://github.com/vacp2p/zerokit"
       },
       "original": {
-        "owner": "vacp2p",
-        "repo": "zerokit",
-        "rev": "dc0b31752c91e7b4fefc441cfa6a8210ad7dba7b",
-        "type": "github"
+        "rev": "3160d9504d07791f2fc9b610948a6cf9a58ed488",
+        "type": "git",
+        "url": "https://github.com/vacp2p/zerokit"
       }
     }
   },
--- a/flake.nix
+++ b/flake.nix
@@ -7,9 +7,12 @@
   };
 
   inputs = {
-    nixpkgs.url = "github:NixOS/nixpkgs?rev=f44bd8ca21e026135061a0a57dcf3d0775b67a49";
+    # We are pinning the commit because ultimately we want to use same commit across different projects.
+    # A commit from nixpkgs 24.11 release : https://github.com/NixOS/nixpkgs/tree/release-24.11
+    nixpkgs.url = "github:NixOS/nixpkgs/0ef228213045d2cdb5a169a95d63ded38670b293";
+    # WARNING: Remember to update commit and use 'nix flake update' to update flake.lock.
     zerokit = {
-      url = "github:vacp2p/zerokit?rev=dc0b31752c91e7b4fefc441cfa6a8210ad7dba7b";
+      url = "git+https://github.com/vacp2p/zerokit?rev=3160d9504d07791f2fc9b610948a6cf9a58ed488";
       inputs.nixpkgs.follows = "nixpkgs";
     };
   };
@@ -58,8 +61,6 @@
           inherit stableSystems;
           src = self;
           targets = ["libwaku"];
-          # We are not able to compile the code with nim-unwrapped-2_0
-          useSystemNim = false;
           zerokitRln = zerokit.packages.${system}.rln;
         };
 
@@ -67,12 +68,10 @@
           inherit stableSystems;
           src = self;
           targets = ["wakucanary"];
-          # We are not able to compile the code with nim-unwrapped-2_0
-          useSystemNim = false;
           zerokitRln = zerokit.packages.${system}.rln;
         };
 
-        default = libwaku-android-arm64;
+        default = libwaku;
       });
 
       devShells = forAllSystems (system: {
--- a/library/kernel_api/debug_node_api.nim
+++ b/library/kernel_api/debug_node_api.nim
@@ -8,7 +8,8 @@ import
   libp2p/peerid,
   metrics,
   ffi
-import waku/factory/waku, waku/node/waku_node, waku/node/health_monitor, library/declare_lib
+import
+  waku/factory/waku, waku/node/waku_node, waku/node/health_monitor, library/declare_lib
 
 proc getMultiaddresses(node: WakuNode): seq[string] =
   return node.info().listenAddresses
--- a/nix/default.nix
+++ b/nix/default.nix
@@ -1,9 +1,8 @@
 {
-  config ? {},
-  pkgs ? import <nixpkgs> { },
+  pkgs,
   src ? ../.,
   targets ? ["libwaku-android-arm64"],
-  verbosity ? 2,
+  verbosity ? 1,
   useSystemNim ? true,
   quickAndDirty ? true,
   stableSystems ? [
@@ -19,43 +18,31 @@ assert pkgs.lib.assertMsg ((src.submodules or true) == true)
 let
   inherit (pkgs) stdenv lib writeScriptBin callPackage;
 
-  revision = lib.substring 0 8 (src.rev or "dirty");
+  androidManifest = "<manifest xmlns:android=\"http://schemas.android.com/apk/res/android\" package=\"com.example.mylibrary\" />";
 
-in stdenv.mkDerivation rec {
+  tools = pkgs.callPackage ./tools.nix {};
+  version = tools.findKeyValue "^version = \"([a-f0-9.-]+)\"$" ../waku.nimble;
+  revision = lib.substring 0 8 (src.rev or src.dirtyRev or "00000000");
 
+in stdenv.mkDerivation {
   pname = "logos-messaging-nim";
-
-  version = "1.0.0-${revision}";
+  version = "${version}-${revision}";
 
   inherit src;
 
+  # Runtime dependencies
   buildInputs = with pkgs; [
-    openssl
-    gmp
-    zip
+    openssl gmp zip
   ];
 
   # Dependencies that should only exist in the build environment.
   nativeBuildInputs = let
     # Fix for Nim compiler calling 'git rev-parse' and 'lsb_release'.
     fakeGit = writeScriptBin "git" "echo ${version}";
-    # Fix for the zerokit package that is built with cargo/rustup/cross.
-    fakeCargo = writeScriptBin "cargo" "echo ${version}";
-    # Fix for the zerokit package that is built with cargo/rustup/cross.
-    fakeRustup = writeScriptBin "rustup" "echo ${version}";
-    # Fix for the zerokit package that is built with cargo/rustup/cross.
-    fakeCross = writeScriptBin "cross" "echo ${version}";
-  in
-    with pkgs; [
-      cmake
-      which
-      lsb-release
-      zerokitRln
-      nim-unwrapped-2_0
-      fakeGit
-      fakeCargo
-      fakeRustup
-      fakeCross
+  in with pkgs; [
+    cmake which zerokitRln nim-unwrapped-2_2 fakeGit
+  ] ++ lib.optionals stdenv.isDarwin [
+    pkgs.darwin.cctools gcc # Necessary for libbacktrace
   ];
 
   # Environment variables required for Android builds
@@ -63,14 +50,13 @@ in stdenv.mkDerivation rec {
   ANDROID_NDK_HOME="${pkgs.androidPkgs.ndk}";
   NIMFLAGS = "-d:disableMarchNative -d:git_revision_override=${revision}";
   XDG_CACHE_HOME = "/tmp";
-  androidManifest = "<manifest xmlns:android=\"http://schemas.android.com/apk/res/android\" package=\"com.example.mylibrary\" />";
 
   makeFlags = targets ++ [
     "V=${toString verbosity}"
     "QUICK_AND_DIRTY_COMPILER=${if quickAndDirty then "1" else "0"}"
     "QUICK_AND_DIRTY_NIMBLE=${if quickAndDirty then "1" else "0"}"
     "USE_SYSTEM_NIM=${if useSystemNim then "1" else "0"}"
-    "LIBRLN_FILE=${zerokitRln}/target/release/librln.a"
+    "LIBRLN_FILE=${zerokitRln}/lib/librln.${if abidir != null then "so" else "a"}"
   ];
 
   configurePhase = ''
@@ -80,12 +66,8 @@ in stdenv.mkDerivation rec {
   '';
 
   # For the Nim v2.2.4 built with NBS we added sat and zippy
-  preBuild = ''
-    ln -s waku.nimble waku.nims
-    
-    ${lib.optionalString (!useSystemNim) ''
+  preBuild = lib.optionalString (!useSystemNim) ''
     pushd vendor/nimbus-build-system/vendor/Nim
-
     mkdir dist
     mkdir -p dist/nimble/vendor/sat
     mkdir -p dist/nimble/vendor/checksums
@@ -98,9 +80,7 @@ in stdenv.mkDerivation rec {
     cp -r ${callPackage ./checksums.nix {}}/. dist/nimble/vendor/checksums
     cp -r ${callPackage ./zippy.nix {}}/.     dist/nimble/vendor/zippy
     chmod 777 -R dist/nimble csources_v2
-
     popd
-    ''}
   '';
 
   installPhase = if abidir != null then ''
@@ -110,10 +90,10 @@ in stdenv.mkDerivation rec {
     cd $out && zip -r libwaku.aar *
   '' else ''
     mkdir -p $out/bin $out/include
-    
+
     # Copy library files
     cp build/* $out/bin/ 2>/dev/null || true
-    
+
     # Copy the header file
     cp library/libwaku.h $out/include/
   '';
--- a/nix/pkgs/android-sdk/compose.nix
+++ b/nix/pkgs/android-sdk/compose.nix
@@ -5,19 +5,16 @@
 
 { androidenv, lib, stdenv }:
 
-assert lib.assertMsg (stdenv.system != "aarch64-darwin")
-  "aarch64-darwin not supported for Android SDK. Use: NIXPKGS_SYSTEM_OVERRIDE=x86_64-darwin";
-
 # The "android-sdk-license" license is accepted
 # by setting android_sdk.accept_license = true.
 androidenv.composeAndroidPackages {
   cmdLineToolsVersion = "9.0";
   toolsVersion = "26.1.1";
-  platformToolsVersion = "33.0.3";
+  platformToolsVersion = "34.0.5";
   buildToolsVersions = [ "34.0.0" ];
   platformVersions = [ "34" ];
   cmakeVersions = [ "3.22.1" ];
-  ndkVersion = "25.2.9519653";
+  ndkVersion = "27.2.12479018";
   includeNDK = true;
   includeExtras = [
     "extras;android;m2repository"
--- a/nix/shell.nix
+++ b/nix/shell.nix
@@ -1,16 +1,12 @@
-{
-  pkgs ? import <nixpkgs> { },
-}:
-let
-  optionalDarwinDeps = pkgs.lib.optionals pkgs.stdenv.isDarwin [
-    pkgs.libiconv
-    pkgs.darwin.apple_sdk.frameworks.Security
-  ];
-in
+{ pkgs  }:
+
 pkgs.mkShell {
   inputsFrom = [
     pkgs.androidShell
-  ] ++ optionalDarwinDeps;
+  ] ++ pkgs.lib.optionals pkgs.stdenv.isDarwin [
+    pkgs.libiconv
+    pkgs.darwin.apple_sdk.frameworks.Security
+  ];
 
   buildInputs = with pkgs; [
     git
@@ -18,7 +14,6 @@ pkgs.mkShell {
     rustup
     rustc
     cmake
-    nim-unwrapped-2_0
+    nim-unwrapped-2_2
   ];
-
 }
--- a/scripts/build_rln_android.sh
+++ b/scripts/build_rln_android.sh
@@ -25,4 +25,3 @@ cargo clean
 cross rustc --release --lib --target=${android_arch} --crate-type=cdylib
 cp ../target/${android_arch}/release/librln.so ${output_dir}/.
 popd
-
--- a/tests/waku_filter_v2/test_waku_filter_dos_protection.nim
+++ b/tests/waku_filter_v2/test_waku_filter_dos_protection.nim
@@ -217,7 +217,8 @@ suite "Waku Filter - DOS protection":
     for fut in finished:
       check not fut.failed()
       let pingRes = fut.read()
-      if pingRes.isErr() and pingRes.error().kind == FilterSubscribeErrorKind.TOO_MANY_REQUESTS:
+      if pingRes.isErr() and
+          pingRes.error().kind == FilterSubscribeErrorKind.TOO_MANY_REQUESTS:
         gotTooMany = true
         break
 


waku/api/api.nim
```diff
@@ -15,3 +19,34 @@ proc createNode*(config: NodeConfig): Future[Result[Waku, string]] {.async.} =
...
+  # check if health is satisfactory
+  # If Node is not healthy, return err("Waku node is not healthy")
+  let healthStatus = RequestNodeHealth.request(w.brokerCtx)
+
+  if healthStatus.isErr():
+    warn "Failed to get Waku node health status: ", error = healthStatus.error
+    # Let's suppose the node is hesalthy enough, go ahead
+  else:
+    if healthStatus.get().healthStatus != NodeHealth.Unhealthy:
+      return err("Waku node is not healthy, has got no connections.")
```
Ivansete-status, line 37: I think is better to make sure the health _API_ is working properly. If not, we are blind.
```suggestion
  let healthStatus = RequestNodeHealth.request(w.brokerCtx).valueOr:
    return err("could not retrieve node's health: " & $error)

  if healthStatus.healthStatus != NodeHealth.Unhealthy:
      return err("Waku node is not healthy, has got no connections.")
```

btw, this `healthStatus.healthStatus` seems not super correct :)

waku/api/request_id.nim
```diff
@@ -0,0 +1,13 @@
+{.push raises: [].}
+
+import bearssl/rand
+
+import waku/utils/requests as request_utils
+
+import ./types
+
+proc newRequestId*(rng: ref HmacDrbgContext): RequestId =
```
Ivansete-status: Even though I prefer your proposal, I think this suggestion is more idiomatic.
```suggestion
proc new*(T: typedesc[RequestId], rng: ref HmacDrbgContext): RequestId =
```

waku/api/types.nim
```diff
@@ -0,0 +1,65 @@
+{.push raises: [].}
+
+import bearssl/rand, std/times, chronos, chronicles
+import stew/byteutils
+import waku/utils/requests as request_utils
+import waku/waku_core/[topics/content_topic, message/message, time]
+import waku/requests/requests
+
+logScope:
+  topics = "message envelope"
```
Ivansete-status: It seems not needed. Not needed `chronicles` for now.

waku/api/types.nim
```diff
@@ -0,0 +1,65 @@
...
+    contentTopic*: ContentTopic
+    payload*: seq[byte]
+    ephemeral*: bool
+
+  RequestId* = distinct string
+
+  NodeHealth* {.pure.} = enum
+    Healthy
+    MinimallyHealthy
+    Unhealthy
```
Ivansete-status, line 20: I think we concluded to only have `Healthy` and `Unhealthy` but @fcecin might have much better insight :)

waku/api/types.nim
```diff
@@ -0,0 +1,65 @@
...
+
+proc newRequestId*(rng: ref HmacDrbgContext): RequestId =
+  ## Generate a new RequestId using the provided RNG.
+  RequestId(request_utils.generateRequestId(rng))
```
Ivansete-status: This seems to be already defined in `waku/api/request_id.nim`

waku/api/types.nim
```diff
@@ -0,0 +1,65 @@
...
+      contentTopic: contentTopic, payload: payload.toBytes(), ephemeral: ephemeral
+    )
+
+proc toWakuMessage*(envelope: MessageEnvelope): WakuMessage =
+  ## Convert a MessageEnvelope to a WakuMessage.
+  var wm = WakuMessage(
+    contentTopic: envelope.contentTopic,
+    payload: envelope.payload,
+    ephemeral: envelope.ephemeral,
+    timestamp: getNanosecondTime(getTime().toUnixFloat()),
```
Ivansete-status: I think let's better use `getNowInNanosecondTime()` from `waku/waku_core/time.nim`.

waku/common/broker/broker_context.nim
```diff
@@ -0,0 +1,68 @@
...
+    let previousBrokerCtx = globalBrokerContextValue
+    globalBrokerContextValue = brokerCtx
+    try:
+      body
+    finally:
+      globalBrokerContextValue = previousBrokerCtx
+      try:
+        globalBrokerContextLock.release()
+      except AsyncLockError:
+        doAssert false, "globalBrokerContextLock.release(): lock not held"
```
Ivansete-status, line 58: I wonder if we could return `Result[...]` instead of raising exception?

waku/events/message_events.nim
```diff
@@ -0,0 +1,23 @@
...
+  # Event emitted when a message is sent to the network
+  type MessageSentEvent* = object
+    requestId*: RequestId
+    messageHash*: string
+
+EventBroker:
+  # Event emitted when a message send operation fails
+  type MessageErrorEvent* = object
+    requestId*: RequestId
+    messageHash*: string
```
Ivansete-status, line 17: Maybe we can use, and adapt if needed, the `WakuMessageHash` type defined in `waku/waku_core/message/digest.nim`.

waku/factory/waku.nim
```diff
@@ -366,9 +363,15 @@ proc startWaku*(waku: ptr Waku): Future[Result[void, string]] {.async.} =
...
-      dnsDiscoveryConf.enrTreeUrl, dnsDiscoveryConf.nameServers
-    )
+    let dynamicBootstrapNodesRes =
+      try:
+        await waku_dnsdisc.retrieveDynamicBootstrapNodes(
+          dnsDiscoveryConf.enrTreeUrl, dnsDiscoveryConf.nameServers
+        )
+      except CatchableError:
+        Result[seq[RemotePeerInfo], string].err(
+          "Retrieving dynamic bootstrap nodes failed: " & getCurrentExceptionMsg()
```
Ivansete-status: Something that I learned recently is that this suggestion seems to be the most idiomatic one. Not a big deal, though.
```suggestion
      except CatchableError as exc:
        Result[seq[RemotePeerInfo], string].err(
          "Retrieving dynamic bootstrap nodes failed: " & exc.msg
```

waku/factory/waku.nim
```diff
@@ -382,8 +385,11 @@ proc startWaku*(waku: ptr Waku): Future[Result[void, string]] {.async.} =
     return err("error while calling startNode: " & $error)
 
   ## Update waku data that is set dynamically on node start
-  (await updateWaku(waku)).isOkOr:
-    return err("Error in updateApp: " & $error)
+  try:
+    (await updateWaku(waku)).isOkOr:
+      return err("Error in updateApp: " & $error)
+  except CatchableError:
+    return err("Error in updateApp: " & getCurrentExceptionMsg())
```
Ivansete-status: Is interesting having a separate message to hint a bit more about the issue,
```suggestion
    return err("Caught exception in updateApp: " & getCurrentExceptionMsg())
```

waku/factory/waku.nim
```diff
@@ -425,41 +431,64 @@ proc startWaku*(waku: ptr Waku): Future[Result[void, string]] {.async.} =
...
+        await (
+          waku_metrics.startMetricsServerAndLogging(
+            conf.metricsServerConf.get(), conf.portsShift
+          )
         )
+      ).valueOr:
+        return err("Starting monitoring and external interfaces failed: " & error)
+    except CatchableError:
+      return err(
+        "Starting monitoring and external interfaces failed: " & getCurrentExceptionMsg()
```
Ivansete-status:
```suggestion
    except CatchableError as exc:
      return err("Caught exception starting monitoring and external interfaces failed: " & exc.msg)
```

waku/factory/waku.nim
```diff
@@ -425,41 +431,64 @@ proc startWaku*(waku: ptr Waku): Future[Result[void, string]] {.async.} =
...
 
-proc stop*(waku: Waku): Future[void] {.async: (raises: [Exception]).} =
+proc stop*(waku: Waku): Future[Result[void, string]] {.async: (raises: []).} =
   ## Waku shutdown
 
   if not waku.node.started:
     warn "stop: attempting to stop node that isn't running"
 
-  waku.healthMonitor.setOverallHealth(HealthStatus.SHUTTING_DOWN)
+  try:
```
Ivansete-status, line 512: Shall we narrow down the `try..except` scope? Maybe we can avoid that exception being raised.

waku/node/delivery_service/recv_service/recv_service.nim
```diff
@@ -28,14 +32,15 @@ type RecvMessage = object
...
+type RecvService* = ref object of RootObj
+  brokerCtx: BrokerContext
   topicsInterest: Table[PubsubTopic, seq[ContentTopic]]
     ## Tracks message verification requests and when was the last time a
     ## pubsub topic was verified for missing messages
     ## The key contains pubsub-topics
-
-  storeClient: WakuStoreClient
-  deliveryCb: DeliveryFeedbackCallback
+  node: WakuNode
```
Ivansete-status, line 42: I think we are giving too much scope to the `RecvService`. I would try to give only what it needs which seems `storeClient`, or maybe something else, but not the whole node.

waku/node/delivery_service/recv_service/recv_service.nim
```diff
@@ -150,47 +156,57 @@ method onUnsubscribe(
   do:
     error "onUnsubscribe unsubscribing from wrong topic", pubsubTopic, contentTopics
 
-proc new*(
-    T: type RecvMonitor,
-    storeClient: WakuStoreClient,
-    wakuFilterClient: WakuFilterClient,
-): T =
+proc new*(T: type RecvService, node: WakuNode): T =
```
Ivansete-status: Sth super nitpick that I learned recently.
```suggestion
proc new*(T: typedesc[RecvService], node: WakuNode): T =
```

waku/factory/waku.nim
```diff
@@ -425,41 +431,64 @@ proc startWaku*(waku: ptr Waku): Future[Result[void, string]] {.async.} =
...
-    await waku.dnsRetryLoopHandle.cancelAndWait()
+proc isModeEdgeAvailable*(waku: Waku): bool =
+  return
+    waku.node.wakuRelay.isNil() and not waku.node.wakuStoreClient.isNil() and
+    not waku.node.wakuFilterClient.isNil() and not waku.node.wakuLightPushClient.isNil()
 
-  if not waku.healthMonitor.isNil():
-    await waku.healthMonitor.stopHealthMonitor()
+proc isP2PReliabilityEnabled*(waku: Waku): bool =
+  return not waku.deliveryService.isNil()
```
Ivansete-status: These seem not to be used anywhere.
```suggestion
```

waku/node/delivery_service/send_service/delivery_task.nim
```diff
@@ -0,0 +1,72 @@
+import std/[options, times], chronos
+import waku/waku_core, waku/api/types, waku/requests/node_requests
+import waku/common/broker/broker_context
+
+type DeliveryState* {.pure.} = enum
+  Entry
+  SuccessfullyPropagated
+  SuccessfullyValidated
```
Ivansete-status, line 9: Is interesting adding a tiny comment on each variant but, specially interesting in these two.

waku/node/delivery_service/send_service/delivery_task.nim
```diff
@@ -0,0 +1,72 @@
...
+proc create*(
+    T: type DeliveryTask,
+    requestId: RequestId,
+    envelop: MessageEnvelope,
+    brokerCtx: BrokerContext,
+): Result[T, string] =
+  let msg = envelop.toWakuMessage()
+  # TODO: use sync request for such as soon as available
+  let relayShardRes = (
+    waitFor RequestRelayShard.request(
```
Ivansete-status: We can not use `waitFor` at this level. This can only be used at the top and with care because that might cause deadlock.
We can have this `new` proc as async.

waku/node/delivery_service/send_service/delivery_task.nim
```diff
@@ -0,0 +1,72 @@
...
+  pubsubTopic*: PubsubTopic
+  msg*: WakuMessage
+  msgHash*: WakuMessageHash
+  tryCount*: int
+  state*: DeliveryState
+  deliveryTime*: Moment
+  errorDesc*: string
+
+proc create*(
+    T: type DeliveryTask,
```
Ivansete-status: More idiomatic :)
```suggestion
proc new*(
    T: typedesc[DeliveryTask],
```

waku/node/delivery_service/send_service/lightpush_processor.nim
```diff
@@ -0,0 +1,77 @@
...
+
+logScope:
+  topics = "send service lightpush processor"
+
+type LightpushSendProcessor* = ref object of BaseSendProcessor
+  peerManager: PeerManager
+  lightpushClient: WakuLightPushClient
+
+proc new*(
+    T: type LightpushSendProcessor,
```
Ivansete-status: Just for consistency with my other nitpick comment ;P
```suggestion
    T: typedesc[LightpushSendProcessor],
```

waku/node/delivery_service/send_service/lightpush_processor.nim
```diff
@@ -0,0 +1,77 @@
...
+
+method sendImpl*(
+    self: LightpushSendProcessor, task: DeliveryTask
+): Future[void] {.async.} =
+  task.tryCount.inc()
+  info "Trying message delivery via Lightpush",
+    requestId = task.requestId, msgHash = task.msgHash, tryCount = task.tryCount
+
+  let peer = self.peerManager.selectPeer(WakuLightPushCodec, some(task.pubsubTopic)).valueOr:
+    task.state = DeliveryState.NextRoundRetry
```
Ivansete-status, line 50: Maybe add a `debug` log line?

waku/node/delivery_service/send_service/relay_processor.nim
```diff
@@ -0,0 +1,72 @@
...
+  )
+
+proc isTopicHealthy(self: RelaySendProcessor, topic: PubsubTopic): bool {.gcsafe.} =
+  let healthReport = RequestRelayTopicsHealth.request(self.brokerCtx, @[topic]).valueOr:
+    return false
+
+  if healthReport.topicHealth.len() < 1:
+    return false
+  let health = healthReport.topicHealth[0].health
+  return health == MINIMALLY_HEALTHY or health == SUFFICIENTLY_HEALTHY
```
Ivansete-status, line 44: That seems to indicate that we don't really need the `SUFFICIENTLY_HEALTHY` state for now. It seems as a premature optimization.

waku/node/delivery_service/send_service/relay_processor.nim
```diff
@@ -0,0 +1,72 @@
...
+      DeliveryState.FailedToDeliver
+
+  return RelaySendProcessor(
+    publishProc: publishProc,
+    fallbackStateToSet: fallbackStateToSet,
+    brokerCtx: brokerCtx,
+  )
+
+proc isTopicHealthy(self: RelaySendProcessor, topic: PubsubTopic): bool {.gcsafe.} =
+  let healthReport = RequestRelayTopicsHealth.request(self.brokerCtx, @[topic]).valueOr:
```
Ivansete-status, line 35:
```suggestion
  let healthReport = RequestRelayTopicsHealth.request(self.brokerCtx, @[topic]).valueOr:
  debug "The topic is not healthy", reason = error
```

waku/node/delivery_service/send_service/relay_processor.nim
```diff
@@ -0,0 +1,72 @@
...
+    error "Failed to publish message with relay",
+      request = task.requestId, msgHash = task.msgHash, error = errorMessage
+    if pushResult.error.code != LightPushErrorCode.NO_PEERS_TO_RELAY:
+      task.state = DeliveryState.FailedToDeliver
+      task.errorDesc = errorMessage
+    else:
+      task.state = self.fallbackStateToSet
+    return
+
+  if pushResult.isOk and pushResult.get() > 0:
```
Ivansete-status: Seems not needed as we are checking `isErr` before.
```suggestion
  if pushResult.get() > 0:
```

waku/node/delivery_service/send_service/relay_processor.nim
```diff
@@ -0,0 +1,72 @@
...
+  if pushResult.isErr():
+    let errorMessage = pushResult.error.desc.get($pushResult.error.code)
+    error "Failed to publish message with relay",
+      request = task.requestId, msgHash = task.msgHash, error = errorMessage
+    if pushResult.error.code != LightPushErrorCode.NO_PEERS_TO_RELAY:
+      task.state = DeliveryState.FailedToDeliver
+      task.errorDesc = errorMessage
+    else:
+      task.state = self.fallbackStateToSet
+    return
```
Ivansete-status: Let's better have `valueOr`.
```suggestion
  let pushResult = (await self.publishProc(task.pubsubTopic, task.msg)).valueOr:
    let errorMessage = pushResult.error.desc.get($error.code)
    error "Failed to publish message with relay",
      request = task.requestId, msgHash = task.msgHash, error = errorMessage
    if pushResult.error.code != LightPushErrorCode.NO_PEERS_TO_RELAY:
      task.state = DeliveryState.FailedToDeliver
      task.errorDesc = errorMessage
    else:
      task.state = self.fallbackStateToSet
    return
```

waku/node/delivery_service/send_service/send_service.nim
```diff
@@ -0,0 +1,247 @@
...
+      else:
+        some(rlnRelay)
+    let publishProc = getRelayPushHandler(relay, rln)
+
+    processors.add(RelaySendProcessor.new(isLightPushAvail, publishProc, brokerCtx))
+  if isLightPushAvail:
+    processors.add(LightpushSendProcessor.new(peerManager, lightpushClient, brokerCtx))
+
+  var currentProcessor: BaseSendProcessor = processors[0]
+  for i in 1 ..< processors.len():
```
Ivansete-status: verb: parenthesis; noun: not :)
```suggestion
  for i in 1 ..< processors.len:
```

waku/node/delivery_service/send_service/send_service.nim
```diff
@@ -0,0 +1,247 @@
...
+
+proc checkMsgsInStore(self: SendService, tasksToValidate: seq[DeliveryTask]) {.async.} =
+  if tasksToValidate.len() == 0:
+    return
+
+  if not isStorePeerAvailable(self):
+    warn "Skipping store validation for ",
+      messageCount = tasksToValidate.len(), error = "no store peer available"
+    return
+
```
Ivansete-status, line 141: This is not needed because the `queryToAny` already performs such validation.

```suggestion

proc checkMsgsInStore(self: SendService, tasksToValidate: seq[DeliveryTask]) {.async.} =
  if tasksToValidate.len() == 0:
    return

```

waku/node/delivery_service/send_service/send_service.nim
```diff
@@ -0,0 +1,247 @@
...
+  var hashesToValidate = tasksToValidate.mapIt(it.msgHash)
+
+  let storeResp: StoreQueryResponse = (
+    await self.node.wakuStoreClient.queryToAny(
+      StoreQueryRequest(includeData: false, messageHashes: hashesToValidate)
```
Ivansete-status, line 147: Something we need to confirm is whether this needs the hashes in the `0x..` format or not.

waku/node/delivery_service/send_service/send_service.nim
```diff
@@ -0,0 +1,247 @@
...
+    w.peerManager, w.wakuLightPushClient, w.wakuRelay, w.wakuRlnRelay, w.brokerCtx
+  ).valueOr:
+    return err(error)
+
+  let sendService = SendService(
+    brokerCtx: w.brokerCtx,
+    taskCache: newSeq[DeliveryTask](),
+    serviceLoopHandle: nil,
+    sendProcessor: sendProcessorChain,
+    node: w,
```
Ivansete-status, line 120: I believe we only need to pass the `wakuStoreClient`, nothing more.

waku/node/delivery_service/send_service/send_service.nim
```diff
@@ -0,0 +1,247 @@
...
+
+  await self.checkMsgsInStore(tasksToValidate)
+
+proc reportTaskResult(self: SendService, task: DeliveryTask) =
+  case task.state
+  of DeliveryState.SuccessfullyPropagated:
+    # TODO: in case of of unable to strore check messages shall we report success instead?
+    info "Message successfully propagated",
+      requestId = task.requestId, msgHash = task.msgHash
+    MessagePropagatedEvent.emit(self.brokerCtx, task.requestId, task.msgHash.toString())
```
Ivansete-status: More idiomatic :)
Applies elsewhere.
```suggestion
    MessagePropagatedEvent.emit(self.brokerCtx, task.requestId, $task.msgHash)
```

waku/node/delivery_service/send_service/send_service.nim
```diff
@@ -0,0 +1,247 @@
...
+    return
+  of DeliveryState.FailedToDeliver:
+    error "Failed to send message",
+      requestId = task.requestId, msgHash = task.msgHash, error = task.errorDesc
+    MessageErrorEvent.emit(
+      self.brokerCtx, task.requestId, task.msgHash.toString(), task.errorDesc
+    )
+    return
+  else:
+    # rest of the states are intermediate and does not translate to event
```
Ivansete-status, line 203: Maybe we can add a `debug` log line.

waku/node/delivery_service/send_service/send_service.nim
```diff
@@ -0,0 +1,247 @@
...
+    discard
+
+  if task.messageAge() > MaxTimeInCache:
+    error "Failed to send message",
+      requestId = task.requestId, msgHash = task.msgHash, error = "Message too old"
```
Ivansete-status: Interesting to print the messageAge.

waku/node/delivery_service/send_service/send_service.nim
```diff
@@ -0,0 +1,247 @@
...
+    await sleepAsync(ServiceLoopInterval)
+
+proc startSendService*(self: SendService) =
+  self.serviceLoopHandle = self.serviceLoop()
+
+proc stopSendService*(self: SendService) =
+  if not self.serviceLoopHandle.isNil():
+    discard self.serviceLoopHandle.cancelAndWait()
+
+proc send*(self: SendService, task: DeliveryTask): Future[void] {.async.} =
```
Ivansete-status: Implicit for `async`.
```suggestion
proc send*(self: SendService, task: DeliveryTask) {.async.} =
```

waku/node/delivery_service/send_service/send_service.nim
```diff
@@ -0,0 +1,247 @@
...
+  assert(not task.isNil(), "task for send must not be nil")
```
Ivansete-status, line 257: Given that this `send` is started by `asyncSpawn` I think it is not healthy raising an exception here.
```suggestion
```

waku/node/health_monitor/topic_health.nim
```diff
@@ -1,18 +1,20 @@
...
   SUFFICIENTLY_HEALTHY
+  NOT_SUBSCRIBED
 
 proc `$`*(t: TopicHealth): string =
   result =
     case t
     of UNHEALTHY: "UnHealthy"
     of MINIMALLY_HEALTHY: "MinimallyHealthy"
     of SUFFICIENTLY_HEALTHY: "SufficientlyHealthy"
+    of NOT_SUBSCRIBED: "NotSubscribed"
```
Ivansete-status, line 17: This whole `$` proc is not needed if we declare `TopicHealth* {.pure.}`.

waku/node/kernel_api/lightpush.nim
```diff
@@ -193,29 +193,30 @@ proc lightpushPublishHandler(
...
-        error "could not create mix connection"
-        return lighpushErrorResult(
-          LightPushErrorCode.SERVICE_NOT_AVAILABLE,
-          "Waku lightpush with mix not available",
-        )
-
-      return await node.wakuLightpushClient.publish(some(pubsubTopic), message, conn)
+    if defined(libp2p_mix_experimental_exit_is_dest) and mixify:
+      #indicates we want to use mix to send the message
+      when defined(libp2p_mix_experimental_exit_is_dest):
```
Ivansete-status, line 205: This block is a little bit weird :)

waku/node/kernel_api/relay.nim
```diff
@@ -184,16 +184,20 @@ proc publish*(
       let msg = "Autosharding error: " & error
       return err(msg)
 
-  #TODO instead of discard return error when 0 peers received the message
-  discard await node.wakuRelay.publish(pubsubTopic, message)
+  let numPeers = (await node.wakuRelay.publish(pubsubTopic, message)).valueOr:
+    warn "waku.relay did not publish", error = error
+    # Todo: If NoPeersToPublish, we might want to return ok(0) instead!!!
+    return err($error)
```
Ivansete-status: I like adding some more context so that we can track the "error stack" later on.
```suggestion
    return err("publish failed in relay: " & $error)
```

waku/node/kernel_api/relay.nim
```diff
@@ -184,16 +184,20 @@ proc publish*(
...
   notice "waku.relay published",
     peerId = node.peerId,
     pubsubTopic = pubsubTopic,
     msg_hash = pubsubTopic.computeMessageHash(message).to0xHex(),
-    publishTime = getNowInNanosecondTime()
+    publishTime = getNowInNanosecondTime(),
+    numPeers = numPeers
 
-  return ok()
+  # TODO: investigate if we can return error in case numPeers is 0
```
Ivansete-status, line 206: Yes, I think we should return an error if numPeers is 0 regardless possible upcoming new peers.

waku/node/waku_node.nim
```diff
@@ -443,6 +464,20 @@ proc updateAnnouncedAddrWithPrimaryIpAddr*(node: WakuNode): Result[void, string]
 
   return ok()
 
+proc startProvidersAndListeners*(node: WakuNode) =
```
```suggestion
proc startProvidersAndListeners(node: WakuNode) =
```

waku/node/waku_node.nim
```diff
@@ -443,6 +464,20 @@ proc updateAnnouncedAddrWithPrimaryIpAddr*(node: WakuNode): Result[void, string]
...
+    proc(
+        pubsubTopic: Option[PubsubTopic], contentTopic: ContentTopic
+    ): Future[Result[RequestRelayShard, string]] {.async.} =
+      let shard = node.deduceRelayShard(contentTopic, pubsubTopic).valueOr:
+        return err($error)
+      return ok(RequestRelayShard(relayShard: shard))
+  ).isOkOr:
+    error "Can't set proveder for RequestRelayShard", error = error
+
+proc stopProvidersAndListeners*(node: WakuNode) =
```
```suggestion
proc stopProvidersAndListeners(node: WakuNode) =
```

waku/waku_core/event/event_emitter.nim
```diff
@@ -0,0 +1,20 @@
+
+
+type
+  EventEmitter* = object
+    # Placeholder for future event emitter implementation
+    observers*: seq[proc (data: EventData): void]
+
```
Ivansete-status: Is this parsed by `nph`? I see weird empty lines.

waku/waku_core/message/digest.nim
```diff
@@ -19,6 +19,11 @@ func shortLog*(hash: WakuMessageHash): string =
 func `$`*(hash: WakuMessageHash): string =
   shortLog(hash)
 
+func toString*(hash: WakuMessageHash): string =
```
Ivansete-status: Ah, now I see :)
Maybe this name avoids possible confusions.
```suggestion
func to0xHex*(hash: WakuMessageHash): string =
```

waku/waku_lightpush/callbacks.nim
```diff
@@ -31,15 +31,15 @@ proc checkAndGenerateRLNProof*(
 
 proc getNilPushHandler*(): PushMessageHandler =
   return proc(
-      peer: PeerId, pubsubTopic: string, message: WakuMessage
+      pubsubTopic: string, message: WakuMessage
```
Ivansete-status, line 34: Why? That seems that we are losing valuable information.

waku/waku_lightpush/client.nim
```diff
@@ -80,37 +82,34 @@ proc publish*(
...
   )
-  let relayPeerCount = ?await wl.sendPushRequestToConn(request, conn)
 
-  for obs in wl.publishObservers:
-    obs.onMessagePublished(pubSubTopic.get(""), message)
+  let relayPeerCount =
+    when dest is Connection:
+      ?await wl.sendPushRequest(request, dest.peerId, some(dest))
+    elif dest is RemotePeerInfo:
+      ?await wl.sendPushRequest(request, dest)
```
Ivansete-status:
```suggestion
```

waku/waku_rln_relay/rln_relay.nim
```diff
@@ -438,6 +449,18 @@ proc mount(
...
+  RequestGenerateRlnProof.setProvider(
+    proc(
+        msg: WakuMessage, senderEpochTime: float64
+    ): Future[Result[RequestGenerateRlnProof, string]] {.async.} =
+      let proof = createRlnProof(wakuRlnRelay, msg, senderEpochTime).valueOr:
+        return err("Could not create RLN proof: " & $error)
+
+      return ok(RequestGenerateRlnProof(proof: proof))
+  ).isOkOr:
+    return err("Proof generator provider cannot be set")
```
Ivansete-status:
```suggestion
    return err("Proof generator provider cannot be set: " & $error)
```

Ivansete-status (approved):
Awesome PR! Thanks so much for it! 🙌 
Just adding some nitpicks that I hope you find useful.

waku/node/delivery_service/send_service/delivery_task.nim
NagyZoltanPeter: It is addressed by using new RequestBroker feature which enable sync requests.

waku/node/delivery_service/send_service/send_service.nim
NagyZoltanPeter: This implies a programming error as no such thing must happen. So it must be caught on testing.

waku/node/delivery_service/send_service/relay_processor.nim
NagyZoltanPeter: Well, idk for now. It is worth to add to the health api disco. There is a difference between node predicted healthiness and weather we can send a message because of lack of supporting node for that specific topic.

waku/node/delivery_service/send_service/send_service.nim
NagyZoltanPeter: It is here purposfully, as a placeholder as in the upcoming PRs I would like to add peer management where availability, former errors got from a certain service peer will be tracked so as such queryToAny will not be used.

waku/node/delivery_service/send_service/relay_processor.nim
NagyZoltanPeter: Certainly if we cannot get topic health is a programming issue and an error here. We shall be able to get health state with whatever outcome.

waku/node/delivery_service/recv_service/recv_service.nim
NagyZoltanPeter: Somewhat agree, but let's cover this in a separate PR.

waku/node/delivery_service/send_service/send_service.nim
NagyZoltanPeter: Hm, I think as WakuNode represents the kernel-api we need this to be used by lmn-api. that's the layering we use. WDYT?

Ivansete-status: Well, nowadays I'm pretty convinced about providing the minimum necessary. We can wait for another opinion in that matter if you like :)

I agree that `WakuNode` can represent the kernel-api though.
Not blocking at all.

waku/events/message_events.nim
NagyZoltanPeter: I was considering to use cross language type here without additional interpretation. These events are meant to be consumed out of the library and I thought it is better without expressing inner types. WDYT?

waku/node/delivery_service/send_service/send_service.nim
NagyZoltanPeter: Maybe trace if it is that necessary. I expect one task can get here several times and as such can pollute the log.

waku/node/delivery_service/send_service/send_service.nim
```diff
@@ -0,0 +1,261 @@
...
+    waku_lightpush/callbacks,
+    events/message_events,
+    common/broker/broker_context,
+  ]
+
+logScope:
+  topics = "send service"
+
+# This useful util is missing from sequtils, this extends applyIt with predicate...
+template applyItIf*(varSeq, pred, op: untyped) =
```
fcecin, line 27: Just a note, I miss having a simple "waku utils" module that I can import, and where we can put all globally useful procs, templates and macros. I see `waku/common/utils` is deprecated, but maybe we could try doing something else here.

waku/api/types.nim
NagyZoltanPeter: Yeah, let's move it to her PR and spec to discuss.

waku/api/types.nim
NagyZoltanPeter: Sure

waku/node/kernel_api/lightpush.nim
NagyZoltanPeter: I know, but it will not compile without that compile time define. So either we default to `libp2p_mix_experimental_exit_is_dest` or make such distinction here.

waku/api/api.nim
```diff
@@ -15,3 +20,52 @@ proc createNode*(config: NodeConfig): Future[Result[Waku, string]] {.async.} =
...
+  if w.isNil():
+    return err("Waku node is not initialized")
+
+  # check if health is satisfactory
+  # If Node is not healthy, return err("Waku node is not healthy")
+  let healthStatus = RequestNodeHealth.request(w.brokerCtx)
+
+  if healthStatus.isErr():
+    warn "Failed to get Waku node health status: ", error = healthStatus.error
+    # Let's suppose the node is hesalthy enough, go ahead
```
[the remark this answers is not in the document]
Ivansete-status: Aside from typo, I believe we shouldn't assume that and instead, return an error. _Health API_ is a must.

waku/api/api.nim
```diff
@@ -15,3 +20,52 @@ proc createNode*(config: NodeConfig): Future[Result[Waku, string]] {.async.} =
...
 
   return ok(wakuRes)
+
+proc checkApiAvailability(w: Waku): Result[void, string] =
+  if w.isNil():
+    return err("Waku node is not initialized")
+
+  # check if health is satisfactory
+  # If Node is not healthy, return err("Waku node is not healthy")
+  let healthStatus = RequestNodeHealth.request(w.brokerCtx)
```
Ivansete-status, line 30: May we use `valueOr` instead?
( this `request` event-broker is extremely awesome btw .)

waku/api/api.nim
```diff
@@ -15,3 +20,52 @@ proc createNode*(config: NodeConfig): Future[Result[Waku, string]] {.async.} =
...
+    w: Waku, envelope: MessageEnvelope
+): Future[Result[RequestId, string]] {.async.} =
+  ?checkApiAvailability(w)
+
+  let requestId = RequestId.new(w.rng)
+
+  let deliveryTask = DeliveryTask.create(requestId, envelope, w.brokerCtx).valueOr:
+    return err("API send: Failed to create delivery task: " & error)
+
+  info "API send: scheduling delivery task",
```
Ivansete-status, line 63: I think is also interesting to log `my_peer_id`.

waku/factory/waku.nim
```diff
@@ -400,13 +408,68 @@ proc startWaku*(waku: ptr Waku): Future[Result[void, string]] {.async.} =
...
+        let healthReport = healthReportFut.read()
+
+        # Check if Relay or Lightpush Client is ready (MinimallyHealthy condition)
+        var relayReady = false
+        var lightpushClientReady = false
+        var storeClientReady = false
+        var filterClientReady = false
+
+        for protocolHealth in healthReport.protocolsHealth:
+          if protocolHealth.protocol == "Relay" and
```
Ivansete-status, line 436: Instead of hard-coded values I think is better use `const` defined somewhere.

waku/factory/waku.nim
```diff
@@ -400,13 +408,68 @@ proc startWaku*(waku: ptr Waku): Future[Result[void, string]] {.async.} =
...
+
+        debug "Providing health report",
+          nodeHealth = $nodeHealth,
+          relayReady = relayReady,
+          lightpushClientReady = lightpushClientReady,
+          storeClientReady = storeClientReady,
+          filterClientReady = filterClientReady,
+          details = $(healthReport)
+
+        ok(RequestNodeHealth(healthStatus: nodeHealth))
```
Ivansete-status: Simple consistency nitpick :)
```suggestion
        return ok(RequestNodeHealth(healthStatus: nodeHealth))
```

waku/node/delivery_service/recv_service/recv_service.nim
```diff
@@ -150,47 +157,63 @@ method onUnsubscribe(
...
 
-proc startRecvMonitor*(self: RecvMonitor) =
+proc startRecvService*(self: RecvService) =
   self.msgCheckerHandler = self.msgChecker()
   self.msgPrunerHandler = self.loopPruneOldMessages()
 
-proc stopRecvMonitor*(self: RecvMonitor) {.async.} =
+  self.onSubscribeListener = OnFilterSubscribeEvent.listen(
+    self.brokerCtx,
+    proc(subsEv: OnFilterSubscribeEvent): Future[void] {.async: (raises: []).} =
```
Ivansete-status:
```suggestion
    proc(subsEv: OnFilterSubscribeEvent) {.async: (raises: []).} =
```

waku/node/delivery_service/recv_service/recv_service.nim
```diff
@@ -150,47 +157,63 @@ method onUnsubscribe(
...
+      self.onSubscribe(subsEv.pubsubTopic, subsEv.contentTopics),
+  ).valueOr:
+    error "Failed to set OnFilterSubscribeEvent listener", error = error
+    quit(QuitFailure)
+
+  self.onUnsubscribeListener = OnFilterUnsubscribeEvent.listen(
+    self.brokerCtx,
+    proc(subsEv: OnFilterUnsubscribeEvent): Future[void] {.async: (raises: []).} =
```
Ivansete-status:
```suggestion
    proc(subsEv: OnFilterUnsubscribeEvent) {.async: (raises: []).} =
```

waku/node/delivery_service/send_service/delivery_task.nim
```diff
@@ -0,0 +1,72 @@
...
+  pubsubTopic*: PubsubTopic
+  msg*: WakuMessage
+  msgHash*: WakuMessageHash
+  tryCount*: int
+  state*: DeliveryState
+  deliveryTime*: Moment
+  propagateEventEmitted*: bool
+  errorDesc*: string
+
+proc create*(
```
Ivansete-status:
```suggestion
proc new*(
```

waku/node/delivery_service/send_service/delivery_task.nim
```diff
@@ -0,0 +1,72 @@
...
+    envelop: MessageEnvelope,
+    brokerCtx: BrokerContext,
+): Result[T, string] =
+  let msg = envelop.toWakuMessage()
+  # TODO: use sync request for such as soon as available
+  let relayShardRes = (
+    RequestRelayShard.request(brokerCtx, none[PubsubTopic](), envelop.contentTopic)
+  ).valueOr:
+    echo "RequestRelayShard.request error", $error
+    return err($error)
```
Ivansete-status: Interesting to add more context on each error layer.
```suggestion
    return err("Failed create DeliveryTask: " & $error)
```

waku/node/delivery_service/send_service/lightpush_processor.nim
```diff
@@ -0,0 +1,78 @@
...
+    requestId = task.requestId,
+    msgHash = task.msgHash.to0xHex(),
+    tryCount = task.tryCount
+
+  let peer = self.peerManager.selectPeer(WakuLightPushCodec, some(task.pubsubTopic)).valueOr:
+    task.state = DeliveryState.NextRoundRetry
+    return
+
+  let pushResult =
+    await self.lightpushClient.publish(some(task.pubsubTopic), task.msg, peer)
```
Ivansete-status:
```suggestion
  let numLightpushServers =
    await self.lightpushClient.publish(some(task.pubsubTopic), task.msg, peer).valueOr:
```

waku/node/delivery_service/send_service/relay_processor.nim
```diff
@@ -0,0 +1,78 @@
+import chronos, chronicles
+import std/options
```
Ivansete-status: The order we usually have:
1. std modules
2. non-std external deps
3. internal modules

```suggestion
import std/options
import chronos, chronicles

```

waku/node/delivery_service/send_service/relay_processor.nim
```diff
@@ -0,0 +1,78 @@
...
+
+logScope:
+  topics = "send service relay processor"
+
+type RelaySendProcessor* = ref object of BaseSendProcessor
+  publishProc: PushMessageHandler
+  fallbackStateToSet: DeliveryState
+
+proc new*(
+    T: type RelaySendProcessor,
```
Ivansete-status: To keep consistency with the rest of the PR. I overlooked some other cases.
```suggestion
    T: typedesc[RelaySendProcessor],
```

waku/node/delivery_service/send_service/relay_processor.nim
```diff
@@ -0,0 +1,78 @@
...
+  return health == MINIMALLY_HEALTHY or health == SUFFICIENTLY_HEALTHY
+
+method isValidProcessor*(
+    self: RelaySendProcessor, task: DeliveryTask
+): bool {.gcsafe.} =
+  # Topic health query is not reliable enough after a fresh subscribe...
+  # return self.isTopicHealthy(task.pubsubTopic)
+  return true
+
+method sendImpl*(self: RelaySendProcessor, task: DeliveryTask): Future[void] {.async.} =
```
Ivansete-status:
```suggestion
method sendImpl*(self: RelaySendProcessor, task: DeliveryTask) {.async.} =
```

waku/node/delivery_service/send_service/relay_processor.nim
```diff
@@ -0,0 +1,78 @@
...
+  task.tryCount.inc()
+  info "Trying message delivery via Relay",
```
Ivansete-status, line 55: Interesting to pint `my_peer_id` too.

waku/node/delivery_service/send_service/send_service.nim
Ivansete-status: Well, ideally that kind of utility should be placed in a separate API. In this particular case, I think this should go into: https://github.com/nim-lang/Nim/blob/devel/lib/pure/collections/sequtils.nim

@narimiran - do you think we can submit a PR somewhere in std so that `applyItIf` and `forEach` are added there instead?

waku/node/delivery_service/send_service/send_service.nim
```diff
@@ -0,0 +1,265 @@
...
+
+  var currentProcessor: BaseSendProcessor = processors[0]
+  for i in 1 ..< processors.len:
+    currentProcessor.chain(processors[i])
+    currentProcessor = processors[i]
+
+  return ok(processors[0])
+
+proc new*(
+    T: type SendService, preferP2PReliability: bool, w: WakuNode, s: SubscriptionService
```
Ivansete-status:
```suggestion
    T: typedesc[SendService], preferP2PReliability: bool, w: WakuNode, s: SubscriptionService
```

waku/node/delivery_service/send_service/send_service.nim
```diff
@@ -0,0 +1,265 @@
...
+): Result[T, string] =
+  if w.wakuRelay.isNil() and w.wakuLightpushClient.isNil():
+    return err(
+      "Could not create SendService. wakuRelay or wakuLightpushClient should be set"
+    )
```
Ivansete-status, line 106: We can split the underlying info a little.
```suggestion
  if w.wakuRelay.isNil():
    return err(
      "Could not create SendService. wakuRelay"
    )
  if w.wakuLightpushClient.isNil():
    return err(
      "Could not create SendService. wakuLightpushClient should be set"
    )
```

waku/node/delivery_service/send_service/send_service.nim
```diff
@@ -0,0 +1,265 @@
...
+
+  let checkStoreForMessages = preferP2PReliability and not w.wakuStoreClient.isNil()
+
+  let sendProcessorChain = setupSendProcessorChain(
+    w.peerManager, w.wakuLightPushClient, w.wakuRelay, w.wakuRlnRelay, w.brokerCtx
+  ).valueOr:
+    return err(error)
```
Ivansete-status:
```suggestion
    return err("failed call to setupSendProcessorChain: " & $error)
```

waku/node/delivery_service/send_service/send_service.nim
```diff
@@ -0,0 +1,265 @@
...
+    it.state == DeliveryState.SuccessfullyPropagated and it.deliveryAge() > ArchiveTime and
+      not it.isEphemeral()
+  )
+
+  await self.checkMsgsInStore(tasksToValidate)
+
+proc reportTaskResult(self: SendService, task: DeliveryTask) =
+  case task.state
+  of DeliveryState.SuccessfullyPropagated:
+    # TODO: in case of of unable to strore check messages shall we report success instead?
```
Ivansete-status:
```suggestion
    # TODO: in case of unable to strore check messages shall we report success instead?
```

I think is fine to keep `SuccessfullyPropagated` state.

waku/api/api.nim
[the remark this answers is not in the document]
NagyZoltanPeter: I have the impression that it might be better to try (with retry rounds within the timeframe before send error emitted) it, as due to fallback mechanism it can still succeed within time. Until further data is not gathered about how health status stable can be this approach may give a better performance.

waku/api/api.nim
NagyZoltanPeter: Ok! Will add.

waku/factory/waku.nim
NagyZoltanPeter: Yes its a leftover, I remove it.

waku/factory/waku.nim
NagyZoltanPeter: Right. Let's address this in a separate PR when merging with Health API.

waku/factory/waku.nim
NagyZoltanPeter: I think more of these calls can raise. But I will look later.

waku/api/types.nim
NagyZoltanPeter: Yes it was needed for the proof attach that is removed for now from here.

waku/node/delivery_service/send_service/relay_processor.nim
NagyZoltanPeter: This info is not available in send processor... but at higher level I will add, I think it is enough when task added.

waku/node/delivery_service/send_service/send_service.nim
NagyZoltanPeter: Yes, sure, I will do!

waku/node/delivery_service/send_service/send_service.nim
NagyZoltanPeter: Not exactly, it should not report error in edge mode where there is no relay mounted. So to say it is valid to have relay nil.

waku/node/health_monitor/topic_health.nim
NagyZoltanPeter: I would just keep it for having nicer PascalCase printouts. WDYT?

waku/node/kernel_api/relay.nim
NagyZoltanPeter: Actually it is now handled in wakuRelay.publish within this PR. So from now 0 peer return is an error `NoPeersToPublish`

waku/waku_core/event/event_emitter.nim
NagyZoltanPeter: Oh, I completely overlooked this file... unused... it was before EventBroker's added. I remove it completely

waku/waku_lightpush/callbacks.nim
NagyZoltanPeter: I don't think so. First of all it was not in use at all. That peerId is the source of the push request and already logged before the handler is called.

waku/api/api.nim
fcecin: I'll address the issue here in #3689

Ivansete-status:
@NagyZoltanPeter - something I forgot to mention is that we need to expose the _Send API_ in `libwaku.nim` ( maybe other functions too .)
Also, CI should be green but the js jobs ;P

NagyZoltanPeter:
> Why was this merged with failing Linting?
> 
> image

Thanks for pointing it out. It seems I overlooked it as formerly there were some other reds in the ci but not the lint.
I will tackle in the next PR.

PR #3670, opened 2025-12-16
chore: use chronos' TokenBucket

NagyZoltanPeter:
…nim-chronos -> nim-libp2p/nim-lsquic/nim-jwt -> adapt to latest libp2p changes

## Description

As finally fixed and extended `TokenBucket` is available from nim-chronos.
This PR, get in use it instead of logos-messaging implementation.

## Note

As this PR involves nim-chronos and nim-libp2p bump consider to include for the preparation of new release!
That effort can validate these bumps.

## Changes

- bump nim-chronos to latest master
- removed token_bucket.nim and related test (now all covered in nim-chronos' ratelimit.nim)
- Made necessary code changes to reflect the minimal differences with new implemenation.
- new nim-choros required to bump nim-libp2p (due to new AsyncSemaphore implementation)
- new libp2p requeries new dependencies: nim-lsquic and nim-jwt
- libp2p changes needed some code changes in logos-messaging
   - Most important change that new AsyncSemaphore has no exported  `size` (max value) anymore - this is used in our PeerManager to get MaxConnection setting. This is removed and maxConnection of `ConnManager`  of `Switch` is also added to PeerManager at config time.  

## Update

Due the TokenBucket's `Continuous` mode slightly changed, mostly in timing of how new tokens are minted, I needed to refactor DOS protection tests.

## Issue

closes https://github.com/logos-messaging/logos-messaging-nim/issues/3543

### DIFF
--- a/tests/waku_lightpush/test_ratelimit.nim
+++ b/tests/waku_lightpush/test_ratelimit.nim
@@ -83,11 +83,9 @@ suite "Rate limited push service":
     # Don't rely on per-request timing assumptions or a single shared Future.
     # CI can be slow enough that sequential requests accidentally refill tokens.
     # Instead we issue a small burst and assert we observe at least one rejection.
-    var handledCount = 0
     let handler = proc(
         peer: PeerId, pubsubTopic: PubsubTopic, message: WakuMessage
     ): Future[WakuLightPushResult] {.async.} =
-      inc handledCount
       return lightpushSuccessResult(1)
 
     let
--- a/tests/waku_lightpush_legacy/test_ratelimit.nim
+++ b/tests/waku_lightpush_legacy/test_ratelimit.nim
@@ -86,58 +86,52 @@ suite "Rate limited push service":
     await allFutures(serverSwitch.start(), clientSwitch.start())
 
     ## Given
-    var handlerFuture = newFuture[(string, WakuMessage)]()
     let handler = proc(
         peer: PeerId, pubsubTopic: PubsubTopic, message: WakuMessage
     ): Future[WakuLightPushResult[void]] {.async.} =
-      handlerFuture.complete((pubsubTopic, message))
       return ok()
 
     let
+      tokenPeriod = 500.millis
       server = await newTestWakuLegacyLightpushNode(
-        serverSwitch, handler, some((3, 500.millis))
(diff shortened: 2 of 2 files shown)


waku/common/rate_limit/single_token_limiter.nim
```diff
@@ -6,20 +6,28 @@ import std/[options], chronos/timer, libp2p/stream/connection, libp2p/utility
...
-import ./[token_bucket, setting, service_metrics]
+import chronos/ratelimit as token_bucket
+
+import ./[setting, service_metrics]
 export token_bucket, setting, service_metrics
 
 proc newTokenBucket*(
     setting: Option[RateLimitSetting],
-    replenishMode: ReplenishMode = ReplenishMode.Compensating,
+    replenishMode: ReplenishMode = ReplenishMode.Continuous,
```
Ivansete-status: Can we consider that `Continuous` is more strict?

.gitmodules
```diff
@@ -184,3 +184,9 @@
 	url = https://github.com/logos-messaging/waku-rlnv2-contract.git
 	ignore = untracked
 	branch = master
+[submodule "vendor/nim-lsquic"]
+	path = vendor/nim-lsquic
+	url = https://github.com/vacp2p/nim-lsquic
+[submodule "vendor/nim-jwt"]
+	path = vendor/nim-jwt
+	url = https://github.com/vacp2p/nim-jwt.git
```
Ivansete-status, line 192: Is this really needed? I don't see new imports but I might have overlooked something. I'd like to understand why

waku/common/rate_limit/single_token_limiter.nim
NagyZoltanPeter: No, it is just renamed.. due to the nature of minting new tokens over elapsed time. So in other words the bucket compensates time elapsed without using tokens. But it is continuous in terms of minting of tokens are not limited to period boundaries.

.gitmodules
NagyZoltanPeter: New libp2p requires these if you look at libp2p.nimble file. Although libp2p depends on lsquic which is likely needs jwt - I think for the authorization mechanism behind.

PR #3673, opened 2025-12-16
fix(tests): prevent race condition in postgres driver tests

fcecin:
## Description

Messaging tests (such as tests/waku_archive/test_driver_postgres_query.nim) can temporarily fail if a new partition in the DB is not created in time. This is unlikely to happen when running the test suite, but it can happen (the insert can jump ahead of the partition creation by a few milliseconds).

## Changes

* Add `PostgresDriver.waitForPartition()` which waits for a partition to be available for a given timeout period 
* Call `waitForPartition` in the archive test suite and wait for a partition to be created before starting each messaging test

## Issue

Maintenance Y2026H1 https://github.com/logos-messaging/logos-messaging-nim/issues/3686

### DIFF
--- a/tests/waku_archive/test_driver_postgres_query.nim
+++ b/tests/waku_archive/test_driver_postgres_query.nim
@@ -38,6 +38,8 @@ suite "Postgres driver - queries":
 
     driver = PostgresDriver(driverRes.get())
 
+    (await driver.waitForPartition()).expect("Test has no DB partition")
+
   asyncTeardown:
     let resetRes = await driver.reset()
 
--- a/waku/waku_archive/driver/postgres_driver/postgres_driver.nim
+++ b/waku/waku_archive/driver/postgres_driver/postgres_driver.nim
@@ -1425,7 +1425,6 @@ proc containsAnyPartition*(self: PostgresDriver): bool =
 proc waitForPartition*(
     self: PostgresDriver, timeout = chronos.seconds(5)
 ): Future[ArchiveDriverResult[void]] {.async.} =
-
   let pollInterval = chronos.milliseconds(100)
   var elapsed = chronos.milliseconds(0)
 
--- a/waku/waku_archive/driver/postgres_driver/postgres_driver.nim
+++ b/waku/waku_archive/driver/postgres_driver/postgres_driver.nim
@@ -1422,6 +1422,22 @@ proc removeOldestPartition(
 proc containsAnyPartition*(self: PostgresDriver): bool =
   return not self.partitionMngr.isEmpty()
 
+proc waitForPartition*(
+    self: PostgresDriver, timeout = chronos.seconds(5)
+): Future[ArchiveDriverResult[void]] {.async.} =
+
+  let pollInterval = chronos.milliseconds(100)
+  var elapsed = chronos.milliseconds(0)
+
+  while elapsed < timeout:
+    if self.containsAnyPartition():
+      return ok()
+
+    await sleepAsync(pollInterval)
+    elapsed += pollInterval
+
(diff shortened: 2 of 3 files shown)


tests/waku_archive/test_driver_postgres_query.nim
```diff
@@ -38,6 +38,8 @@ suite "Postgres driver - queries":
 
     driver = PostgresDriver(driverRes.get())
 
+    (await driver.waitForPartition()).expect("Test has no DB partition")
```
Ivansete-status, line 41: I don't know how the `expect` exception is handled. Also, seems interesting to track the returned error message even though now is only one
```suggestion
    (await driver.waitForPartition()).isOkOr:
    assert false, "Partition not available while testing: " & $error
```

waku/waku_archive/driver/postgres_driver/postgres_driver.nim
```diff
@@ -1422,6 +1422,21 @@ proc removeOldestPartition(
 proc containsAnyPartition*(self: PostgresDriver): bool =
   return not self.partitionMngr.isEmpty()
 
+proc waitForPartition*(
+    self: PostgresDriver, timeout = chronos.seconds(5)
+): Future[ArchiveDriverResult[void]] {.async.} =
```
Ivansete-status, line 1427: Shall we mention that this is meant to avoid flaky testing only?

waku/waku_archive/driver/postgres_driver/postgres_driver.nim
```diff
@@ -1422,6 +1422,21 @@ proc removeOldestPartition(
...
+  let pollInterval = chronos.milliseconds(100)
+  var elapsed = chronos.milliseconds(0)
+
+  while elapsed < timeout:
+    if self.containsAnyPartition():
```
Ivansete-status, line 1432: I wonder if we should confirm that there is a valid partition for "now".
Each partition contains data for o'clock hours unix time.

Ivansete-status (approved):
LGTM! Thanks so much for it! 🙌 
Just added some nitpicks.
Cheers.

tests/waku_archive/test_driver_postgres_query.nim
fcecin: The expect() macro on the Result does that for us! I tested it: it appends the error automatically to the message. So in the expect string we say what we expected, and the err() that makes it fail (from the operation) explains why the expectation was not fulfilled.

waku/waku_archive/driver/postgres_driver/postgres_driver.nim
fcecin: It's just a generic timed wrapper for containsAnyPartition so I think we would have to put the same warning there as well. I'm not sure we should be afraid of people using timed-out queries to "a partition exists" outside of tests or as part of helping an actual test to test something useful.

waku/waku_archive/driver/postgres_driver/postgres_driver.nim
fcecin: Yes, if we step out of "there is a partition" as a basic check we may step into second-guessing what the partition manager is doing w.r.t. time calc.

Maybe that's the actual solution to this. I should probably get more into what the partition manager actually does. Now I'm not sure I want to merge this fix as it is :-)

I think this PR will linger here a bit since this testing race condition is so difficult to reproduce. It's the opposite of urgent. Certainly not triggering in the CI machines, which is slower than our own machines (where this is *already* very difficult to trigger). So this is worth waiting for doing actually right.

fcecin:
The code has changed a lot since I was looking into this and the flake no longer manifests, so this PR is no longer relevant.

PR #3674, opened 2025-12-17
chore: context aware brokers

NagyZoltanPeter:
## Description

Implementation of Send API https://github.com/logos-messaging/logos-messaging-nim/pull/3669 revealed that logos-messaging-nim use of Events and Requests needs to be used in contextual manner. It is because all requests and events are valid within the context of a Waku instance.
Without this Multiple instances of Waku will mix up each others events and requests. 
This comes vital during running unit tests where we need to instantiate several instances of WakuNode to represent a local node network.

## Changes

- RequestBroker:
  -   Added a BrokerContext argument to all public interfaces: `request`, `setProvider`, `clearProvider`. Call to `request` with same BrokerContext will be dispatched only to the counterpart providers.
  -   For simplicity of use, in case no context has meaning, original interfaces kept which means a DefaultBrokerContext - that exists all the time.
- EventBroker
  -    Similar context aware support for `listen`,`emit` and `dropListener` interfaces.
  -    Same support for use native type typed events, or use externally defined types in event definition is added. 
- MultiRequestBroker
  - Add context aware use like the other brokers
  - Add same support for native typed or external typed requests. 
- BrokerContext is added as a distinct type with generate function and helpers.
- Code refactoring - move all common/duplicated code into broker_utils

# Notice
This feature is necessary to complete SendAPI feature.

### DIFF
--- a/waku/common/broker/broker_context.nim
+++ b/waku/common/broker/broker_context.nim
@@ -1,18 +1,68 @@
-import std/[strutils, concurrency/atomics]
+{.push raises: [].}
+
+import std/[strutils, concurrency/atomics], chronos
 
 type BrokerContext* = distinct uint32
 
-func `==`*(a, b: BrokerContext): bool {.borrow.}
+func `==`*(a, b: BrokerContext): bool =
+  uint32(a) == uint32(b)
+
+func `!=`*(a, b: BrokerContext): bool =
+  uint32(a) != uint32(b)
 
 func `$`*(bc: BrokerContext): string =
   toHex(uint32(bc), 8)
 
 const DefaultBrokerContext* = BrokerContext(0xCAFFE14E'u32)
 
+# Global broker context accessor.
+#
+# NOTE: This intentionally creates a *single* active BrokerContext per process
+# (per event loop thread). Use only if you accept serialization of all broker
+# context usage through the lock.
+var globalBrokerContextLock {.threadvar.}: AsyncLock
+globalBrokerContextLock = newAsyncLock()
+var globalBrokerContextValue {.threadvar.}: BrokerContext
+globalBrokerContextValue = DefaultBrokerContext
+proc globalBrokerContext*(): BrokerContext =
+  ## Returns the currently active global broker context.
+  ##
+  ## This is intentionally lock-free; callers should use it inside
+  ## `withNewGlobalBrokerContext` / `withGlobalBrokerContext`.
+  globalBrokerContextValue
+
 var gContextCounter: Atomic[uint32]
 
-proc newBrokerContext*(): BrokerContext =
+proc NewBrokerContext*(): BrokerContext =
   var nextId = gContextCounter.fetchAdd(1, moRelaxed)
   if nextId == uint32(DefaultBrokerContext):
     nextId = gContextCounter.fetchAdd(1, moRelaxed)
   return BrokerContext(nextId)
+
+template lockGlobalBrokerContext*(brokerCtx: BrokerContext, body: untyped): untyped =
+  ## Runs `body` while holding the global broker context lock with the provided
+  ## `brokerCtx` installed as the globally accessible context.
+  ##
+  ## This template is intended for use from within `chronos` async procs.
+  block:
+    await noCancel(globalBrokerContextLock.acquire())
+    let previousBrokerCtx = globalBrokerContextValue
+    globalBrokerContextValue = brokerCtx
+    try:
+      body
+    finally:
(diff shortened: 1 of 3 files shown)


waku/common/broker/broker_context.nim
```diff
@@ -0,0 +1,26 @@
...
+func `$`*(bc: BrokerContext): string =
+  toHex(uint32(bc), 8)
+
+const DefaultBrokerContext* = BrokerContext(0xCAFFE14E'u32)
+
+proc NewBrokerContext*(): BrokerContext =
+  ## Generates a random non-default broker context (as a raw uint32).
+  ##
+  ## The default broker context is reserved for the provider at index 0.
+  ## This helper never returns that value.
```
fcecin: The 32-bit int space is prone to collision (Birthday Paradox, 50% odds at around 77K tries) so it's not really fit for random ID gen (64-bit is fine with collisions at 50% in 4 billion). Instead, we could use a process-wide, global atomic uint32 counter here?

A global counter still helps with debugging since even brokers created across threads won't have repeated IDs in debugging output, just like with the random IDs but without the collision risk. Small ID values from a counter that starts at say 1 or 1,000 (latter is better for the auto-padding in output) also have the additional benefit of being easier to read, provided they are globally unique (i.e. come from a global/atomic ID gen). 

Could also try to get fancier and use something derived from the local clock or a fragment of the thread ID, or a combination, like in 128-bit UUIDs, but 32 bits isn't much space to play with.

NagyZoltanPeter: Well, you're absolutely right and thinking further, there is no need for randomness at all. 
Counter is just enough.
Also to mention, I do not expect large number of contexts at all. Our tests may require maybe 4-5 at max. Hence, I was using a seq as underlying store that is more effective than associative containers for such small number of items.
So yeah I change it to some atomic counter.

fcecin: Yes I saw the seq for context and I'm absolutely on board with it, table will probably be slower for small count.

waku/common/broker/helper/broker_utils.nim
```diff
@@ -41,3 +57,150 @@ proc baseTypeIdent*(defName: NimNode): NimNode =
...
+    error(macroName & " body must declare exactly one type", body)
+
+  result = ParsedBrokerType(
+    typeIdent: typeIdent,
+    objectDef: objectDef,
+    isRefObject: isRefObject,
+    hasInlineFields: hasInlineFields,
+    fieldNames: fieldNames,
+    fieldTypes: fieldTypes,
+  )
```
Ivansete-status, line 206: I suggest adding the following to all broker macros
```nim
when defined(dumpBrokerMacros):
  echo result.repr
```

waku/common/broker/broker_context.nim
```diff
@@ -0,0 +1,18 @@
+import std/[strutils, concurrency/atomics]
+
+type BrokerContext* = distinct uint32
+
+func `==`*(a, b: BrokerContext): bool {.borrow.}
+
+func `$`*(bc: BrokerContext): string =
+  toHex(uint32(bc), 8)
+
+const DefaultBrokerContext* = BrokerContext(0xCAFFE14E'u32)
```
Ivansete-status, line 16: haha good default :D Why 14E?

waku/common/broker/broker_context.nim
```diff
@@ -0,0 +1,18 @@
...
+
+var gContextCounter: Atomic[uint32]
+
+proc NewBrokerContext*(): BrokerContext =
```
Ivansete-status, line 36: Super nitpick. Maybe better use `init` term as we usually do for `object` types
```suggestion
proc newBrokerContext*(): BrokerContext =
```
Besides, very beautiful technique 👏

waku/common/broker/helper/broker_utils.nim
NagyZoltanPeter: It's there. Called eventBrokerDebug/requestBrokerDebug! ;-) I need them a lot, due that is the first I check.

waku/common/broker/broker_context.nim
NagyZoltanPeter: That's how you would write CAFFEINE in hex :-D.
Old habbits...

waku/common/broker/broker_context.nim
NagyZoltanPeter: Maybe `nextBrokerContext`?
Due to it's about to "generate" new ctx... which is an ID. I felt `init` not really fit.

PR #3678, opened 2025-12-19
Chore: Simplify on chain group manager error handling

stubbsta:
## Description
The on chain group manager raises multiple exceptions which does not align with the error handling format of the repo.
Instead we want errors to be propagated up the stack.

## Changes
- retry-wrapper.nim: 
This is used by several procs in the group-manager and is the reason for many of the raise exceptions.
The template was converted to a proc to simplify it and make the error handling more explicit
- group-manager.nim:
Updates to use new retry-wrapper and return Result type
- all tests using group-manager

## Issue closes #
https://github.com/logos-messaging/logos-messaging-nim/issues/3645

### DIFF
--- a/tests/waku_rln_relay/test_wakunode_rln_relay.nim
+++ b/tests/waku_rln_relay/test_wakunode_rln_relay.nim
@@ -58,11 +58,8 @@ procSuite "WakuNode - RLN relay":
       let manager1 = cast[OnchainGroupManager](node1.wakuRlnRelay.groupManager)
       let idCredentials1 = generateCredentials()
 
-      try:
-        waitFor manager1.register(idCredentials1, UserMessageLimit(20))
-      except Exception, CatchableError:
-        assert false,
-          "exception raised when calling register: " & getCurrentExceptionMsg()
+      (waitFor manager1.register(idCredentials1, UserMessageLimit(20))).isOkOr:
+        assert false, "error returned when calling register: " & error
 
       let rootUpdated1 = waitFor manager1.updateRoots()
       info "Updated root for node1", rootUpdated1
@@ -172,11 +169,8 @@ procSuite "WakuNode - RLN relay":
       let manager1 = cast[OnchainGroupManager](node1.wakuRlnRelay.groupManager)
       let idCredentials1 = generateCredentials()
 
-      try:
-        waitFor manager1.register(idCredentials1, UserMessageLimit(20))
-      except Exception, CatchableError:
-        assert false,
-          "exception raised when calling register: " & getCurrentExceptionMsg()
+      (waitFor manager1.register(idCredentials1, UserMessageLimit(20))).isOkOr:
+        assert false, "error returned when calling register: " & error
 
       let rootUpdated1 = waitFor manager1.updateRoots()
       info "Updated root for node", node = 1, rootUpdated = rootUpdated1
@@ -192,11 +186,8 @@ procSuite "WakuNode - RLN relay":
       let manager2 = cast[OnchainGroupManager](node2.wakuRlnRelay.groupManager)
       let idCredentials2 = generateCredentials()
 
-      try:
-        waitFor manager2.register(idCredentials2, UserMessageLimit(20))
-      except Exception, CatchableError:
-        assert false,
-          "exception raised when calling register: " & getCurrentExceptionMsg()
+      (waitFor manager2.register(idCredentials2, UserMessageLimit(20))).isOkOr:
+        assert false, "error returned when calling register: " & error
 
       let rootUpdated2 = waitFor manager2.updateRoots()
       info "Updated root for node", node = 2, rootUpdated = rootUpdated2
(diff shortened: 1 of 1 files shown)


tests/wakunode_rest/test_rest_relay.nim
```diff
@@ -41,8 +41,8 @@ suite "Waku v2 Rest API - Relay":
   var manager {.threadVar.}: OnchainGroupManager
 
   setup:
-    anvilProc = runAnvil()
-    manager = waitFor setupOnchainGroupManager()
+    anvilProc = runAnvil(stateFile = some(DEFAULT_ANVIL_STATE_PATH))
```
Ivansete-status, line 45: It seems the tests are taking too long and time out. Maybe this is the culprit :) ?

tools/rln_keystore_generator/rln_keystore_generator.nim
```diff
@@ -73,7 +73,11 @@ proc doRlnKeystoreGenerator*(conf: RlnKeystoreGeneratorConf) =
 
   # 4. register on-chain
   try:
-    waitFor groupManager.register(credential, conf.userMessageLimit)
+    let registerResult =
+      waitFor groupManager.register(credential, conf.userMessageLimit)
+    if registerResult.isErr():
+      error "Failed to register on-chain", error = registerResult.error
+      quit(QuitFailure)
```
Ivansete-status: Possible simplification with `isOkOr` :)

Also, maybe we can also get rid of the wrapping `try .. catch `.

```suggestion
    (waitFor groupManager.register(credential, conf.userMessageLimit)).isOkOr:
      error "Failed to register on-chain", error = error
      quit(QuitFailure)
```

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
```diff
@@ -172,40 +148,39 @@ proc updateRoots*(g: OnchainGroupManager): Future[bool] {.async.} =
...
-          raise
-            newException(CatchableError, "Failed to fetch next free index: " & error)
-
-        let memberCount = cast[int64](nextFreeIndex)
-        waku_rln_number_registered_memberships.set(float64(memberCount))
-  except CatchableError:
-    error "Fatal error in trackRootChanges", error = getCurrentExceptionMsg()
+proc trackRootChanges*(g: OnchainGroupManager): Future[Result[void, string]] {.async.} =
+  checkInitialized(g).isOkOr:
+    return err(error)
```
Ivansete-status: In this case, given that underlying `Result[...]` type matches, we can use the `?` operand, which handles it. In case of `ok`, the flow continues, and in case of `err`, the flow breaks and the error is returned.

This possible suggestion may apply elsewhere.

```suggestion
  ? checkInitialized(g)
```

tests/wakunode_rest/test_rest_relay.nim
stubbsta: This is the new config which makes the anvil setup faster, so the total test runtime should be shorter

tools/rln_keystore_generator/rln_keystore_generator.nim
stubbsta: updated in https://github.com/logos-messaging/logos-messaging-nim/pull/3678/commits/c725e3640fac21800532df51327d5c700cf06edb

waku/waku_rln_relay/group_manager/on_chain/group_manager.nim
stubbsta: updated all relevant instances in https://github.com/logos-messaging/logos-messaging-nim/pull/3678/commits/cd77e30fb7776db79430e03ab625812f55b58dcd

darshankabariya:
> @Ivansete-status as we discussed yesterday, I tried to remove the [group_manager_base.nim](https://github.com/logos-messaging/logos-messaging-nim/blob/c725e3640fac21800532df51327d5c700cf06edb/waku/waku_rln_relay/group_manager/group_manager_base.nim). This was to simplify the code, but then I saw that the methods in [group_manager.nim](https://github.com/logos-messaging/logos-messaging-nim/blob/c725e3640fac21800532df51327d5c700cf06edb/waku/waku_rln_relay/group_manager/on_chain/group_manager.nim) would all need to be changed to proc as methods without base methods are deprecated.
> 
> Do we still think this is the best way to go? The alternative is to keep everything as is, but then we have a `group_manager_base.nim` that defines methods that raise exceptions. The other alternative - to remove the raise exceptions from `group_manager_base.nim` is giving issues where other files use the methods and expect exceptions to be raised and as I try to modify one proc a new one has an error and it seems like and endless task.
> 
> I would prefer to keep changes minimal and not have to alter too many files and for this I think that the best option is to change the methods in group_manager.nim to proc, but I would like your input incase I am missing something?

darshankabariya:
LGTM. It looks like the issue is due to a broken CI. i rebase branch let see what happen.

darshankabariya (approved):
Looking much better now.

PR #3680, opened 2025-12-29
nix: add libwaku as pkg in flake

markoburcul:
## Description

Add libwaku to the flake.nix packages. 
Fix the build of Nim using nimbus build system within nix derivations.

## Changes

Only in nix code.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,5 +1,5 @@
 {
-  description = "NWaku build flake";
+  description = "Logos Messaging Nim build flake";
 
   nixConfig = {
     extra-substituters = [ "https://nix-cache.status.im/" ];
@@ -54,10 +54,21 @@
           zerokitRln = zerokit.packages.${system}.rln-android-arm64;
         };
 
+        libwaku = pkgs.callPackage ./nix/default.nix {
+          inherit stableSystems;
+          src = self;
+          targets = ["libwaku"];
+          # We are not able to compile the code with nim-unwrapped-2_0
+          useSystemNim = false;
+          zerokitRln = zerokit.packages.${system}.rln;
+        };
+
         wakucanary = pkgs.callPackage ./nix/default.nix {
           inherit stableSystems;
           src = self;
           targets = ["wakucanary"];
+          # We are not able to compile the code with nim-unwrapped-2_0
+          useSystemNim = false;
           zerokitRln = zerokit.packages.${system}.rln;
         };
 
--- a/nix/default.nix
+++ b/nix/default.nix
@@ -23,7 +23,7 @@ let
 
 in stdenv.mkDerivation rec {
 
-  pname = "nwaku";
+  pname = "logos-messaging-nim";
 
   version = "1.0.0-${revision}";
 
@@ -70,6 +70,7 @@ in stdenv.mkDerivation rec {
     "QUICK_AND_DIRTY_COMPILER=${if quickAndDirty then "1" else "0"}"
     "QUICK_AND_DIRTY_NIMBLE=${if quickAndDirty then "1" else "0"}"
     "USE_SYSTEM_NIM=${if useSystemNim then "1" else "0"}"
+    "LIBRLN_FILE=${zerokitRln}/target/release/librln.a"
   ];
 
   configurePhase = ''
@@ -78,19 +79,28 @@ in stdenv.mkDerivation rec {
     make nimbus-build-system-nimble-dir
   '';
 
+  # For the Nim v2.2.4 built with NBS we added sat and zippy
   preBuild = ''
     ln -s waku.nimble waku.nims
+    
+    ${lib.optionalString (!useSystemNim) ''
     pushd vendor/nimbus-build-system/vendor/Nim
+
     mkdir dist
-    cp -r ${callPackage ./nimble.nix {}}    dist/nimble
-    cp -r ${callPackage ./checksums.nix {}} dist/checksums
-    cp -r ${callPackage ./csources.nix {}}  csources_v2
+    mkdir -p dist/nimble/vendor/sat
+    mkdir -p dist/nimble/vendor/checksums
+    mkdir -p dist/nimble/vendor/zippy
+
+    cp -r ${callPackage ./nimble.nix {}}/.    dist/nimble
+    cp -r ${callPackage ./checksums.nix {}}/. dist/checksums
+    cp -r ${callPackage ./csources.nix {}}/.  csources_v2
+    cp -r ${callPackage ./sat.nix {}}/.       dist/nimble/vendor/sat
+    cp -r ${callPackage ./checksums.nix {}}/. dist/nimble/vendor/checksums
+    cp -r ${callPackage ./zippy.nix {}}/.     dist/nimble/vendor/zippy
     chmod 777 -R dist/nimble csources_v2
+
     popd
-    cp -r ${zerokitRln}/target vendor/zerokit/
-    find vendor/zerokit/target
-    # FIXME
-    cp vendor/zerokit/target/*/release/librln.a librln_v${zerokitRln.version}.a
+    ''}
   '';
 
   installPhase = if abidir != null then ''
@@ -99,8 +109,13 @@ in stdenv.mkDerivation rec {
     echo '${androidManifest}' > $out/jni/AndroidManifest.xml
     cd $out && zip -r libwaku.aar *
   '' else ''
-    mkdir -p $out/bin
-    cp -r build/* $out/bin
+    mkdir -p $out/bin $out/include
+    
+    # Copy library files
+    cp build/* $out/bin/ 2>/dev/null || true
+    
+    # Copy the header file
+    cp library/libwaku.h $out/include/
   '';
 
   meta = with pkgs.lib; {
--- a/nix/zippy.nix
+++ b/nix/zippy.nix
@@ -1,0 +1,9 @@
+{ pkgs }:
+
+pkgs.fetchFromGitHub {
+  owner = "guzba";
+  repo = "zippy";
+  rev = "a99f6a7d8a8e3e0213b3cad0daf0ea974bf58e3f";
+  # WARNING: Requires manual updates when Nim compiler version changes.
+  hash = "sha256-e2ma2Oyp0dlNx8pJsdZl5o5KnaoAX87tqfY0RLG3DZs=";
+}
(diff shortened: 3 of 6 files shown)


nix/default.nix
```diff
@@ -72,6 +72,20 @@ in stdenv.mkDerivation rec {
...
+  postPatch = ''
+    cat > scripts/build_rln.sh << 'EOF'
+    #!/usr/bin/env bash
+    set -e
+    build_dir=$1
+    rln_version=$2
+    output_filename=$3
+    # Just copy the library we already have
+    cp ${zerokitRln}/lib/librln.a "$output_filename" 2>/dev/null || \
+    cp ${zerokitRln}/target/release/librln.a "$output_filename"
```
Ivansete-status: I can't quite get why two instances of same lib?

nix/default.nix
```diff
@@ -99,8 +121,18 @@ in stdenv.mkDerivation rec {
     echo '${androidManifest}' > $out/jni/AndroidManifest.xml
     cd $out && zip -r libwaku.aar *
   '' else ''
-    mkdir -p $out/bin
-    cp -r build/* $out/bin
+    mkdir -p $out/bin $out/include
```
Ivansete-status, line 112: I think it is safer to use ${out}

nix/default.nix
```diff
@@ -99,8 +121,18 @@ in stdenv.mkDerivation rec {
...
+    
+    # Copy library files
+    cp build/*.so $out/bin/ 2>/dev/null || true
+    cp build/*.a $out/bin/ 2>/dev/null || true
+    
+    # Copy the compiler-generated header
+    cp nimcache/debug/libwaku/libwaku.h $out/include/ 2>/dev/null || \
```
Ivansete-status: libwaku.h header should be the same regardless build type

nix/default.nix
```diff
@@ -99,8 +121,18 @@ in stdenv.mkDerivation rec {
...
+    cp nimcache/release/libwaku/libwaku.h $out/include/
+
+    # Copy Nim's nimbase.h (required by libwaku.h)
+    cp vendor/nimbus-build-system/vendor/Nim/lib/nimbase.h $out/include/
```
Ivansete-status: I believe nimbase.h shouldn't be needed now unless I'm missing something :)
```suggestion
```

flake.nix
```diff
@@ -1,13 +1,13 @@
...
+  description = "Logos Messaging Nim build flake";
 
   nixConfig = {
     extra-substituters = [ "https://nix-cache.status.im/" ];
     extra-trusted-public-keys = [ "nix-cache.status.im-1:x/93lOfLU+duPplwMSBR+OlY4+mo+dCN7n0mr4oPwgY=" ];
   };
 
   inputs = {
-    nixpkgs.url = "github:NixOS/nixpkgs?rev=f44bd8ca21e026135061a0a57dcf3d0775b67a49";
+    nixpkgs.url = "github:NixOS/nixpkgs?rev=f44bd8ca21e026135061a0a57dcf3d0775b67a49";
```
Ivansete-status: Seems adding space as a change

nix/default.nix
```diff
@@ -80,17 +94,25 @@ in stdenv.mkDerivation rec {
...
+    mkdir -p dist/nimble/vendor/sat
+    mkdir -p dist/nimble/vendor/checksums
+    mkdir -p dist/nimble/vendor/zippy
+
+    cp -r ${callPackage ./nimble.nix {}}/.    dist/nimble
+    cp -r ${callPackage ./checksums.nix {}}/. dist/checksums
+    cp -r ${callPackage ./csources.nix {}}/.  csources_v2
+    cp -r ${callPackage ./sat.nix {}}/.       dist/nimble/vendor/sat
+    cp -r ${callPackage ./checksums.nix {}}/. dist/nimble/vendor/checksums
+    cp -r ${callPackage ./zippy.nix {}}/.     dist/nimble/vendor/zippy
```
Ivansete-status, line 99: Why is that needed? Not saying is incorrect. Just to learn something :)

nix/default.nix
```diff
@@ -99,8 +121,18 @@ in stdenv.mkDerivation rec {
     echo '${androidManifest}' > $out/jni/AndroidManifest.xml
     cd $out && zip -r libwaku.aar *
   '' else ''
-    mkdir -p $out/bin
-    cp -r build/* $out/bin
+    mkdir -p $out/bin $out/include
+    
+    # Copy library files
+    cp build/*.so $out/bin/ 2>/dev/null || true
```
Ivansete-status: I think we should also need to copy the dynamic macos libs, if any.
```suggestion
    cp build/*.so $out/bin/ 2>/dev/null || true
    cp build/*.dylib $out/bin/ 2>/dev/null || true
```

Ivansete-status (commented):
Thanks! I'm just adding some nitpick comments so far :D

nix/default.nix
markoburcul: This was meant if the zerokit wasn't built with 'release' flag but I believe we do that always so I'll remove this

nix/default.nix
markoburcul: When you try to make nimbus-build-system working to compile Nim then you end up adding these, depending on the nimble version being used. I've done the same think when doing derivation for `nim-codex` and `nwaku` before, but basically NBS expects these subdirectories with content in these places and they are not fetched when you add `?submodules=1` when building the packages from this flake.

nix/default.nix
markoburcul: When building libwaku there is no .dylib file, is that something you plan to add?

nix/default.nix
markoburcul: So I can take the one from `./library/libwaku.h`?

nix/default.nix
markoburcul: Actually no because it gets interpolated by Nix before the script runs

nix/default.nix
markoburcul: Because I was adding libwaku.h from the nimcache folder which is product of the build, it had include of nimbase.h so it was complaining where is it

nix/default.nix
markoburcul: I changed to that one but still in status-go when I try to run the tests within nix shell, it gives me inconsistencies between the bindings:
```
../go/pkg/mod/github.com/waku-org/waku-go-bindings@v0.0.0-20251202095947-63b01b7d0ca9/waku/nwaku.go:114:58: error: 'WakuCallBack' undeclared (first use in this function); did you mean 'WakuGoCallback'?

  114 |                                                         (WakuCallBack) WakuGoCallback,

      |                                                          ^~~~~~~~~~~~

      |                                                          WakuGoCallback

../go/pkg/mod/github.com/waku-org/waku-go-bindings@v0.0.0-20251202095947-63b01b7d0ca9/waku/nwaku.go:114:71: error: expected ')' before 'WakuGoCallback'

  114 |                                                         (WakuCallBack) WakuGoCallback,

      |                                                                       ^~~~~~~~~~~~~~~

      |                                                                       )

../go/pkg/mod/github.com/waku-org/waku-go-bindings@v0.0.0-20251202095947-63b01b7d0ca9/waku/nwaku.go:109:35: note: to match this '('

  109 |                 waku_content_topic(wakuCtx,

      |                                   ^

../go/pkg/mod/github.com/waku-org/waku-go-bindings@v0.0.0-20251202095947-63b01b7d0ca9/waku/nwaku.go:110:57: error: passing argument 2 of 'waku_content_topic' from incompatible pointer type [-Wincompatible-pointer-types]

  110 |                                                         appName,

      |                                                         ^~~~~~~

      |                                                         |

      |                                                         char *

/nix/store/w73hyyfkvqnpwjggsck8mr6lkjqhiiym-logos-m
```

nix/default.nix
Ivansete-status: Okay, may add a comment there stating that this is for n-b-s only? That way, once we use nimble we will be able to remove that snippet.

nix/default.nix
Ivansete-status: That's weird. It should be created indeed. If not now, in the future for sure.

nix/zippy.nix
```diff
@@ -0,0 +1,9 @@
+{ pkgs }:
+
```
Ivansete-status, line 2: Shall add a comment explaining why zippy is needed for?

Ivansete-status (approved):
LGTM! Thanks for it! 🙌 
Just adding some minor nitpicks

nix/zippy.nix
markoburcul: It's not only zippy, but all of the files: chechsums, csources, nimble, sat and zippy. They are all required if you want to build Nim using NBS within the Nix derivation for the current version of NBS and Nim used by LMN. I can add it, sure.

flake.nix
```diff
@@ -54,10 +54,21 @@
           zerokitRln = zerokit.packages.${system}.rln-android-arm64;
         };
 
+        libwaku= pkgs.callPackage ./nix/default.nix {
```
jakubgs:
```suggestion
        libwaku = pkgs.callPackage ./nix/default.nix {
```

nix/default.nix
jakubgs: So we need more of those here because this repo uses different Nim than `nimbus-eth2`?

Because there we only need those:
```
    cp -r ${callPackage ./nimble.nix {}}    dist/nimble
    cp -r ${callPackage ./checksums.nix {}} dist/checksums
    cp -r ${callPackage ./csources.nix {}}  csources_v2
```
https://github.com/status-im/nimbus-eth2/blob/a4e93d13f50293ac2fa6fb24cd17e1680cacce71/nix/default.nix#L65-L67

That is indeed worth a comment.

nix/default.nix
```diff
@@ -72,6 +72,19 @@ in stdenv.mkDerivation rec {
...
+    #!/usr/bin/env bash
+    set -e
+    build_dir=$1
+    rln_version=$2
+    output_filename=$3
+    # Just copy the library we already have
+    cp ${zerokitRln}/target/release/librln.a "$output_filename"
+    EOF
+    chmod +x scripts/build_rln.sh
+  '';
```
jakubgs: Why do we need any of that? Why can't we just set `LIBRLN_FILE` since that's the target being built by the `Makefile` in this repo? Since it's not a `.phony` target if the file exists Make will just skip building anything:
```suggestion
  LIBRLN_FILE = "${zerokitRln}/target/release/librln.a";
```

nix/default.nix
jakubgs: Can't we just copy everything from `build/` to `$out/bin/`? Or is there stuff we don't need?

nix/default.nix
jakubgs: Why do we need any of that? Why can't we just set `LIBRLN_FILE` since that's the target being built by the `Makefile` in this repo? Since it's not a `.phony` target if the file exists Make will just skip building anything:
```suggestion
    "USE_SYSTEM_NIM=${if useSystemNim then "1" else "0"}"
    "LIBRLN_FILE=${zerokitRln}/target/release/librln.a"
  ];
```

PR #3682, opened 2026-01-05
chore: depr vendor and leverage nimble

### LINKED ISSUE #3672: chore: completely stop using vendor folder and only use Nimble
### Background
Willing to move to Nimble-only, we need to rm the `vendor` folder and all the dependencies should be automatically consumed through nimble.

### Details
- [ ] Remove `vendor` folder
- [ ] Compile all apps with nimble
- [ ] Compile libwaku and install it with nimble

### Acceptance criteria
Everything works the same even though not having vendor folder

darshankabariya:
- [x] Removed all submodules except the native ones still required (nim-ffi, zerokit, waku-rlnv2-contract, nph).
- [x] Day-to-day commands (make wakunode2, make test, etc.) remain unchanged.
- [x] we no longer need nimbus-build-system.
- [ ] Update CI according. 
- [x] latest version of all submodule 
- [x] Plan to replace the remaining four vendor libraries with nimble packages and delete ./vendor entirely. 

closes https://github.com/logos-messaging/logos-messaging-nim/issues/3672


.github/workflows/ci.yml
```diff
@@ -104,6 +110,12 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v4
 
+      - name: Install Nim
+        uses: jiro4989/setup-nim-action@v2
```
arnetheduck: instead of installing nim, it makes more sense to install nimble and have nimble manage nim.

Nim itself also needs to be locked down for a reproducible build.

waku.nimble
```diff
@@ -11,27 +8,47 @@ license = "MIT or Apache License 2.0"
...
-  "presto",
-  "regex",
-  "results",
-  "db_connector",
-  "minilru",
-  "ffi"
+requires "nim >= 2.2.4"
+
+# Pure Nim packages - pinned to exact commits via Nimble
+requires "https://github.com/status-im/nim-chronicles#54f5b726025e8c7385e3a6529d3aa27454c6e6ff"
```
arnetheduck: using `requires` this way will effectively prevent third-parties from depending on waku, unless they use the exact same commit of everything.

`nimble lock` can be used to produce a lock file that allows an application to use a locked set of deps while at the same time allowing others that depend on waku to be more flexible in the versions they use. For example, an application might want to use a newer version of `chronicles` for logging, and it's unreasonable for waku to limit this.

.github/workflows/ci.yml
arnetheduck: See also github.com/nim-lang/setup-nimble-action

arnetheduck: also, for a reproducible build, you need to lock the nimble version, which setup-nimble-action allows you to do.

waku.nimble
```diff
@@ -11,29 +11,45 @@ license = "MIT or Apache License 2.0"
...
-  "db_connector",
-  "minilru",
-  "lsquic",
-  "jwt",
-  "ffi"
+requires "nim >= 2.2.4"
+
+# Pure Nim packages - use flexible version requirements
+# Exact versions are locked via nimble.lock for reproducible builds
+requires "chronicles"
```
Ivansete-status: @arnetheduck, ~~Sould we also add a new `.pinned` file so that we enforce a fixed version on each dep?
EDITED: I see we are adding `nimble.lock` but I don't know which option is better or more idiomatic.~~

EDITED: Disregard my previous comments. Juan helped me to confirm the approach suggested in this PR is the correct one.

waku.nimble
darshankabariya: Thanks for the idea. I’m now using nimble.lock to get an exact replica of the packages for the waku build.

darshankabariya:
> @darshankabariya FYI in this #3670 I bumped versions on nim-chronos and nim-libp2p which involves new deps: nim-lsquic and nim-jwt. These changes comes with some code adjustments. I suggest later incorporate these changes into this PR.

Sure, will do. I tried earlier, but it failed. I need to resolve the conflict carefully now.

Issue #3646, opened 2025-11-17
feat: Implement Health API


Ivansete-status:
Implement and expose the [health API spec](https://github.com/waku-org/specs/pull/92).

PR #3689, opened 2026-01-17
feat: implement Waku API Health spec (WIP)

fcecin:
## Description

This PR:
* Implements the node health monitor portion of the [Waku API spec](https://github.com/logos-messaging/specs/pull/92).
  * It allows the node to provide a `[Connected, PartiallyConnected, Disconnected]` overall node status based on the current state of the internal mounted protocols;
  * It allows the application to monitor Content Topic health;
  * Introduces some internal improvements.

## Changes

* Add `NodeHealthChangeHandler` callback and `JsonNodeHealthChangeEvent` callback event
* Add `nodeState` field (alongside the old `nodeHealth` field) to the current Rest API /health response (`HealthReport`), which reports a `NodeHealthStatus` (`[Connected, PartiallyConnected, Disconnected]`).
* Add `waku/common/waku_protocol.nim` to enumerate all Vac protocols mountable by logos-messaging-nim
* Add NodeHealthMonitor and Health API tests
* Add `strength[]` table to `NodeHealthMonitor` to internally measure connectivity strength of each mounted protocol (when applicable)
* Add `calculateConnectionState` proc to centralize `NodeHealthStatus` computing logic.
* Add event-driven `healthLoop` that triggers node connectivity health updates on relevant peer-connectivity (general) and relay (specific) events
* Connect the `RequestConnectionStatus` nim LM API to the new node health monitor logic
* Add `ConnectionStatusChangeEvent` nim LM API
* Add other Request and Event brokers for health monitoring
* Misc improvements, fixes, refactors

## Issue

Addresses #3646

## Notes

The Logos Messaging Health monitoring API spec is itself a work-in-progress, so merging this PR probably shouldn't close the issue yet.

To complete all health monitoring features, some prerequisites need to be implemented, such as:
* Merge with [AutoSharding PR](https://github.com/logos-messaging/logos-messaging-nim/pull/3719)
* [Content Topic Subscribe/Unsubscribe and Receive APIs](https://github.com/logos-messaging/logos-messaging-nim/issues/3710)
* Stability of connection for testable edge nodes connecting to core (relay) service nodes


tests/node/test_wakunode_health_monitor.nim
```diff
@@ -0,0 +1,324 @@
...
+    node/kernel_api/relay,
+    node/kernel_api/store,
+    node/kernel_api/lightpush,
+    node/kernel_api/filter,
+    waku_archive,
+  ]
+
+import ../testlib/[wakunode, wakucore], ../waku_archive/archive_utils
+
+proc p(kind: WakuProtocol, health: HealthStatus): ProtocolHealth =
```
Ivansete-status: Better use a longer name so it's easier to lookup.
```suggestion
proc pLongerName(kind: WakuProtocol, health: HealthStatus): ProtocolHealth =
```

waku/node/health_monitor/node_health_monitor.nim
```diff
@@ -82,173 +125,445 @@ proc getRlnRelayHealth(hm: NodeHealthMonitor): Future[ProtocolHealth] {.async.}
...
+
+  let rlnHealth = await hm.getRlnRelayHealth()
+  protocols.add(rlnHealth)
+
+  return protocols
+
+proc calculateConnectionState*(
+    protocols: seq[ProtocolHealth],
+    strength: Table[WakuProtocol, int],
+    relayFailoverThreshold: int,
```
Ivansete-status: Kindly add comments in this proc so that is clear what information `strength` and `relayFailoverThreshold` contains/represent.

waku/node/health_monitor/node_health_monitor.nim
```diff
@@ -14,52 +16,93 @@ import
...
-  NodeHealthMonitor* = ref object
-    nodeHealth: HealthStatus
-    node: WakuNode
-    onlineMonitor*: OnlineMonitor
-    keepAliveFut: Future[void]
+random.randomize()
+
+const
+  DefaultRelayFailoverThreshold* = 4
+  FailoverThreshold* = 2
```
Ivansete-status: We need comments for these

tests/node/test_wakunode_health_monitor.nim
```diff
@@ -0,0 +1,324 @@
...
+    let protocols =
+      @[
+        p(RelayProtocol, HealthStatus.NOT_READY),
+        p(StoreClientProtocol, HealthStatus.NOT_READY),
+        p(FilterClientProtocol, HealthStatus.NOT_READY),
+        p(LightpushClientProtocol, HealthStatus.NOT_READY),
+      ]
+    let strength = initTable[WakuProtocol, int]()
+    let state =
+      calculateConnectionState(protocols, strength, DefaultRelayFailoverThreshold)
```
Ivansete-status: Shouldn't be `DefaultRelayFailoverThreshold - 1` in this case ?

I'd assume state == Connected when DefaultRelayFailoverThreshold

waku/node/health_monitor/node_health_monitor.nim
```diff
@@ -14,52 +16,93 @@ import
...
-  if node.isNil():
+  if isNil(node):
     warn "WakuNode is not set, cannot check health", protocol_health_instance = $p
     return p.notMounted()
 
+proc getRelayFailoverThreshold(hm: NodeHealthMonitor): int =
+  if isNil(hm.node.wakuRelay):
+    # Could return an Optional[int] instead, but for simplicity just use a default.
+    # This also helps in writing mocks for the health monitor tests.
+    return DefaultRelayFailoverThreshold
```
Ivansete-status: I can't quite see that _failover_ is a correct term in this case. Maybe better without :)
```suggestion
proc getRelayThreshold(hm: NodeHealthMonitor): int =
  if isNil(hm.node.wakuRelay):
    # Could return an Optional[int] instead, but for simplicity just use a default.
    # This also helps in writing mocks for the health monitor tests.
    return DefaultRelayThreshold
```

On the other hand, I think we don't need `DefaultRelayFailoverThreshold` at all. We need to only use the values defined in gossipsub/relay.

If node.isNil() then the `getRelayThreshold` should return an error.
Another option is to perform such `node.isNil()` validation in `NodeHealthMonitor` ctor. We might need to refactor the `proc new*(T: type Waku ...` to allow that but I think the `node_health_monitor.nim` will get simpler then.

tests/node/test_wakunode_health_monitor.nim
```diff
@@ -0,0 +1,324 @@
...
+    let protocols =
+      @[
+        p(RelayProtocol, HealthStatus.READY), p(StoreClientProtocol, HealthStatus.READY)
+      ]
+    var strength = initTable[WakuProtocol, int]()
+    strength[RelayProtocol] = weakCount
+    strength[StoreClientProtocol] = 1
+    let state =
+      calculateConnectionState(protocols, strength, DefaultRelayFailoverThreshold)
+    check state == ConnectionStatus.PartiallyConnected
```
Ivansete-status, line 56: Maybe a comment explaining why "PartiallyConnected" would help.

tests/node/test_wakunode_health_monitor.nim
```diff
@@ -0,0 +1,324 @@
...
+
+  test "Connected, robust relay":
+    let protocols =
+      @[
+        p(RelayProtocol, HealthStatus.READY), p(StoreClientProtocol, HealthStatus.READY)
+      ]
+    var strength = initTable[WakuProtocol, int]()
+    strength[RelayProtocol] = DefaultRelayFailoverThreshold
+    strength[StoreClientProtocol] = FailoverThreshold
```
Ivansete-status: Maybe :)?
```suggestion
    strength[StoreClientProtocol] = HealthyThreshold
```

tests/node/test_wakunode_health_monitor.nim
```diff
@@ -0,0 +1,324 @@
...
+
+suite "Health Monitor - events":
+  asyncTest "Core (relay) health update":
+    let
+      nodeAKey = generateSecp256k1Key()
+      nodeA = newTestWakuNode(nodeAKey, parseIpAddress("127.0.0.1"), Port(0))
+
+    (await nodeA.mountRelay()).expect("Node A failed to mount Relay")
+
+    nodeA.mountStoreClient()
```
Ivansete-status: This is not needed. That happens in `node_factory.nim`

tests/node/test_wakunode_health_monitor.nim
```diff
@@ -0,0 +1,324 @@
...
+
+    proc dummyHandler(topic: PubsubTopic, msg: WakuMessage): Future[void] {.async.} =
+      discard
+
+    nodeA.subscribe((kind: PubsubSub, topic: DefaultPubsubTopic), dummyHandler).expect(
+      "Node A failed to subscribe"
+    )
+    nodeB.subscribe((kind: PubsubSub, topic: DefaultPubsubTopic), dummyHandler).expect(
+      "Node B failed to subscribe"
+    )
```
Ivansete-status, line 182: These might not be needed as `subscribeShards = @[DefaultShardId],` happens when creating node in the test.

tests/node/test_wakunode_health_monitor.nim
```diff
@@ -0,0 +1,324 @@
...
+    let connectTimeLimit = Moment.now() + 10.seconds
+    var gotConnected = false
+
+    while Moment.now() < connectTimeLimit:
+      if lastStatus != ConnectionStatus.Disconnected:
+        gotConnected = true
+        break
+
+      if healthChangeSignal.finished:
+        healthChangeSignal = newFuture[void]()
```
Ivansete-status: Maybe not needed as it happens a few lines below

tests/node/test_wakunode_health_monitor.nim
```diff
@@ -0,0 +1,324 @@
...
+      discard
+
+    nodeA.subscribe((kind: PubsubSub, topic: DefaultPubsubTopic), dummyHandler).expect(
+      "Node A failed to subscribe"
+    )
+    nodeB.subscribe((kind: PubsubSub, topic: DefaultPubsubTopic), dummyHandler).expect(
+      "Node B failed to subscribe"
+    )
+
+    let connectTimeLimit = Moment.now() + 10.seconds
```
Ivansete-status: We cannot wait so long :)
I'd also move that wait value into a `const` at the top of the test file.
```suggestion
    let connectTimeLimit = Moment.now() + 3.seconds
```

tests/node/test_wakunode_health_monitor.nim
```diff
@@ -0,0 +1,324 @@
...
+
+    if healthChangeSignal.finished:
+      healthChangeSignal = newFuture[void]()
+
+    await nodeB.stop()
+    await nodeA.disconnectNode(nodeB.switch.peerInfo.toRemotePeerInfo())
+
+    let disconnectTimeLimit = Moment.now() + 10.seconds
+    var gotDisconnected = false
+
```
Ivansete-status, line 287: I think is interesting to confirm the opposite state happens before entering the loop.
```suggestion
assert lastStatus == ConnectionStatus.Connected, "Expected last status to be connected"
```

waku/common/waku_protocol.nim
```diff
@@ -0,0 +1,24 @@
+{.push raises: [].}
+
+type WakuProtocol* {.pure.} = enum
+  RelayProtocol = "Relay"
+  RlnRelayProtocol = "Rln Relay"
+  StoreProtocol = "Store"
+  LegacyStoreProtocol = "Legacy Store"
```
Ivansete-status, line 7: Maybe simpler. That way we don't add yet another conversion/lookup step.
```suggestion
  RelayProtocol
  RlnRelayProtocol
  StoreProtocol
  LegacyStoreProtocol
  ...
```

waku/events/peer_events.nim
```diff
@@ -0,0 +1,14 @@
+import waku/common/broker/event_broker
+import libp2p/switch
+
+type
+  WakuPeerEventKind* = enum
+    Connected
+    Disconnected
+    Identified
+    MetadataUpdated
```
Ivansete-status: I suggest adding a prefix to avoid possible confusion with _Status_.
```suggestion
    EventConnected
    EventDisconnected
    EventIdentified
    EventMetadataUpdated
```

waku/factory/waku.nim
```diff
@@ -198,6 +198,9 @@ proc new*(
     node.peerManager.getOnlineStateObserver()
   )
 
+  if not isNil(appCallbacks) and not isNil(appCallbacks.connectionStatusChangeHandler):
+    healthMonitor.onConnectionStatusChange = appCallbacks.connectionStatusChangeHandler
```
Ivansete-status: I think we should have a better error control. It seems that it may silently fail if `connectionStatusChangeHandler` is not properly being set.

On the other hand, it seems we already have `appCallbacks.connectionChangeHandler`, which may clash with the new "..ChangeHandler".

waku/factory/waku.nim
```diff
@@ -415,60 +418,31 @@ proc startWaku*(waku: ptr Waku): Future[Result[void, string]] {.async: (raises:
...
     return err("failed to start health monitor: " & $error)
 
-  ## Setup RequestNodeHealth provider
+  ## Setup RequestConnectionStatus provider
+  RequestConnectionStatus.setProvider(
+    globalBrokerContext(),
+    proc(): Result[RequestConnectionStatus, string] =
+      try:
+        let healthReport = waku[].healthMonitor.getSyncNodeHealthReport()
+        ok(RequestConnectionStatus(connectionStatus: healthReport.connectionStatus))
```
Ivansete-status: I think we tend to be a bit more explicit

```suggestion
        return ok(RequestConnectionStatus(connectionStatus: healthReport.connectionStatus))
```

waku/node/health_monitor/connection_status.nim
```diff
@@ -0,0 +1,13 @@
+import chronos, results, std/strutils, ../../api/types
+
+export ConnectionStatus
```
Ivansete-status, line 3: I think I prefer having the `ConnectionStatus` definition in this file too.

Also, I'd instead:
1. Split the whole `types` module.
2. Create a folder named `types`.
3. Inside that folder, create all the types files.

Another option that we also do is having all that tiny type defs in a single `types.nim`, but I personally prefer separate type per file.

waku/node/health_monitor/health_status.nim
```diff
@@ -1,4 +1,4 @@
-import results, std/strutils
+import chronos, results, std/strutils
```
Ivansete-status: This file has some leftover changes :)

waku/node/health_monitor/node_health_monitor.nim
```diff
@@ -14,52 +16,93 @@ import
   ../peer_manager,
   ./online_monitor,
   ./health_status,
-  ./protocol_health
+  ./health_report,
+  ./connection_status,
+  ./protocol_health,
+  ../../api/types,
+  ../../events/health_events,
+  ../../events/peer_events
```
Ivansete-status: We should use "absolute" path as much as possible :)
May apply elsewhere.
```suggestion
  waku/api/types,
  waku/events/health_events,
  waku/events/peer_events
```

waku/node/health_monitor/node_health_monitor.nim
```diff
@@ -14,52 +16,93 @@ import
...
+  node: WakuNode
+  onlineMonitor*: OnlineMonitor
+  keepAliveFut: Future[void]
+  healthLoopFut: Future[void]
+  healthUpdateEvent: AsyncEvent
+  connectionStatus: ConnectionStatus
+  onConnectionStatusChange*: ConnectionStatusChangeHandler
+  cachedProtocols: seq[ProtocolHealth]
+  strength: Table[WakuProtocol, int]
+    ## latest connectivity strength (e.g. peer count) for a protocol
```
Ivansete-status: These should have detailed comments so that is crystal clear what is the `cache` for and what does `strenght` mean.

waku/node/health_monitor/node_health_monitor.nim
```diff
@@ -14,52 +16,93 @@ import
...
+func getHealth*(report: HealthReport, kind: WakuProtocol): ProtocolHealth =
+  for h in report.protocolsHealth:
+    if h.protocol == $kind:
+      return h
+  # Shouldn't happen, but if it does, then assume protocol is not mounted
+  return ProtocolHealth.init(kind)
+
+proc countConnectedPeers(hm: NodeHealthMonitor, codec: string): int =
+  if isNil(hm.node) or isNil(hm.node.peerManager):
+    return 0
```
Ivansete-status: I think is better to explicitly return err
```suggestion
  if isNil(hm.node):
    return err("node is nil")
  
  if isNil(hm.node.peerManager):
    return err("peerManager is nil")
```

waku/node/health_monitor/node_health_monitor.nim
```diff
@@ -14,52 +16,93 @@ import
...
+
+  var peerCount = 0
+  for remotePeerInfo in hm.node.peerManager.switch.peerStore.peers(codec):
+    if hm.node.peerManager.switch.isConnected(remotePeerInfo.peerId):
+      peerCount.inc()
```
Ivansete-status: This is not needed. This information should already be present in `PeerMananger`.
Maybe we need to create a tiny func in `peer_manager.nim`, which would use `connectedPeers` proc.

waku/node/health_monitor/node_health_monitor.nim
```diff
@@ -14,52 +16,93 @@ import
...
+    return DefaultRelayFailoverThreshold
+  return hm.node.wakuRelay.parameters.dLow
+
 proc getRelayHealth(hm: NodeHealthMonitor): ProtocolHealth =
-  var p = ProtocolHealth.init("Relay")
+  var p = ProtocolHealth.init(WakuProtocol.RelayProtocol)
   checkWakuNodeNotNil(hm.node, p)
 
   if hm.node.wakuRelay == nil:
+    hm.strength[WakuProtocol.RelayProtocol] = 0
```
Ivansete-status, line 70: I just noticed now that we should use `isNil()` in the above line :)

waku/node/health_monitor/node_health_monitor.nim
```diff
@@ -14,52 +16,93 @@ import
...
     return p.notMounted()
 
   let relayPeers = hm.node.wakuRelay.getConnectedPubSubPeers(pubsubTopic = "").valueOr:
+    hm.strength[WakuProtocol.RelayProtocol] = 0
     return p.notMounted()
 
-  if relayPeers.len() == 0:
+  let count = relayPeers.len()
```
Ivansete-status: Super nitpick sorry. Names goes without parenthesis, verbs with.
```suggestion
  let count = relayPeers.len
```

waku/node/health_monitor/node_health_monitor.nim
```diff
@@ -82,173 +125,445 @@ proc getRlnRelayHealth(hm: NodeHealthMonitor): Future[ProtocolHealth] {.async.}
...
-proc getLegacyLightpushHealth(
-    hm: NodeHealthMonitor, relayHealth: HealthStatus
-): ProtocolHealth =
-  var p = ProtocolHealth.init("Legacy Lightpush")
+proc getFilterHealth(hm: NodeHealthMonitor, relayHealth: HealthStatus): ProtocolHealth =
+  var p = ProtocolHealth.init(WakuProtocol.FilterProtocol)
   checkWakuNodeNotNil(hm.node, p)
 
-  if hm.node.wakuLegacyLightPush == nil:
+  if hm.node.wakuFilter == nil:
```
Ivansete-status:
```suggestion
  if isNil(hm.node.wakuFilter):
```

waku/node/health_monitor/node_health_monitor.nim
```diff
@@ -82,173 +125,445 @@ proc getRlnRelayHealth(hm: NodeHealthMonitor): Future[ProtocolHealth] {.async.}
...
+  return p.ready()
 
-  return p.notReady("No Lightpush service peer available yet")
+proc getLegacyStoreHealth(hm: NodeHealthMonitor): ProtocolHealth =
+  var p = ProtocolHealth.init(WakuProtocol.LegacyStoreProtocol)
+  checkWakuNodeNotNil(hm.node, p)
 
-proc getFilterHealth(hm: NodeHealthMonitor, relayHealth: HealthStatus): ProtocolHealth =
-  var p = ProtocolHealth.init("Filter")
+  if hm.node.wakuLegacyStore == nil:
```
Ivansete-status:
```suggestion
  if isNil(hm.node.wakuLegacyStore):
```

waku/node/health_monitor/node_health_monitor.nim
```diff
@@ -82,173 +125,445 @@ proc getRlnRelayHealth(hm: NodeHealthMonitor): Future[ProtocolHealth] {.async.}
...
+  if peerCount > 0:
+    return p.ready()
+  return p.notReady("No Lightpush service peer available yet")
 
-proc getFilterClientHealth(
-    hm: NodeHealthMonitor, relayHealth: HealthStatus
-): ProtocolHealth =
-  var p = ProtocolHealth.init("Filter Client")
+proc getLegacyLightpushClientHealth(hm: NodeHealthMonitor): ProtocolHealth =
+  var p = ProtocolHealth.init(WakuProtocol.LegacyLightpushClientProtocol)
```
Ivansete-status, line 192: Maybe not needed `var`?
```suggestion
  let p = ProtocolHealth.init(WakuProtocol.LegacyLightpushClientProtocol)
```

waku/node/health_monitor/node_health_monitor.nim
```diff
@@ -82,173 +125,445 @@ proc getRlnRelayHealth(hm: NodeHealthMonitor): Future[ProtocolHealth] {.async.}
...
+
+    if kind in RelayProtocols:
+      relayCount = max(relayCount, strength)
+    elif kind in StoreClientProtocols:
+      storeClientCount = max(storeClientCount, strength)
+    elif kind in LightpushClientProtocols:
+      lightpushCount = max(lightpushCount, strength)
+    elif kind in FilterClientProtocols:
+      filterCount = max(filterCount, strength)
+
```
Ivansete-status, line 407: Is very interesting to add a debug line somewhere:

```suggestion
debug "Conn state info", relayCount, storeClientCount, lightpushCount, strenght, ...
```

waku/node/health_monitor/node_health_monitor.nim
```diff
@@ -82,173 +125,445 @@ proc getRlnRelayHealth(hm: NodeHealthMonitor): Future[ProtocolHealth] {.async.}
...
+proc onRelayMsg(
+    hm: NodeHealthMonitor, peer: PubSubPeer, msg: var RPCMsg
+) {.gcsafe, raises: [].} =
+  if msg.subscriptions.len == 0:
+    if msg.control.isNone():
+      return
+    let ctrl = msg.control.get()
+    if ctrl.graft.len == 0 and ctrl.prune.len == 0:
+      return
```
Ivansete-status, line 514: This seems to belong to `nim-libp2p` instead.

waku/node/health_monitor/node_health_monitor.nim
```diff
@@ -82,173 +125,445 @@ proc getRlnRelayHealth(hm: NodeHealthMonitor): Future[ProtocolHealth] {.async.}
...
+        hm.connectionStatus = newConnectionStatus
+
+        EventConnectionStatusChange.emit(hm.node.brokerCtx, newConnectionStatus)
+
+        if not isNil(hm.onConnectionStatusChange):
+          await hm.onConnectionStatusChange(newConnectionStatus)
+    except CancelledError:
+      break
+    except Exception as e:
+      error "HealthMonitor: error in update loop", error = e.msg
```
Ivansete-status, line 540:
```suggestion
      error "HealthMonitor: error in update loop", error = e.msg
    info "Health monitor loop end"
```

waku/node/health_monitor/protocol_health.nim
```diff
@@ -1,5 +1,7 @@
 import std/[options, strformat]
 import ./health_status
+import ../../common/waku_protocol
```
Ivansete-status:
```suggestion
import waku/common/waku_protocol
```

waku/node/peer_manager/peer_manager.nim
```diff
@@ -1038,6 +1072,7 @@ proc stop*(pm: PeerManager) =
 proc new*(
     T: type PeerManager,
     switch: Switch,
+    brokerCtx: BrokerContext = globalBrokerContext(),
```
Ivansete-status: Shall we add the new parameter `brokerCtx` at the end? Maybe that way we need less changes somewhere else.

waku/node/waku_node.nim
```diff
@@ -57,12 +58,22 @@ import
...
   ],
   ./net_config,
-  ./peer_manager
+  ./peer_manager,
+  ./health_monitor/health_status,
+  ./health_monitor/topic_health,
+  ../common/rate_limit/setting,
+  ../common/callbacks,
+  ../common/nimchronos,
+  ../waku_mix
```
```suggestion
  waku/common/rate_limit/setting,
  waku/common/callbacks,
  waku/common/nimchronos,
  waku/waku_mix
```

waku/node/waku_node.nim
```diff
@@ -469,7 +487,51 @@ proc updateAnnouncedAddrWithPrimaryIpAddr*(node: WakuNode): Result[void, string]
 
   return ok()
 
-proc startProvidersAndListeners(node: WakuNode) =
+proc calculateEdgeTopicHealth(node: WakuNode, shard: PubsubTopic): TopicHealth =
```
Ivansete-status, line 486: Shouldn't these `calculateEdgeTopicHealth`, `loopEdgeHealth`, and `startProvidersAndListeners` belong to HealthMonitor instead?

waku/rest_api/endpoint/health/types.nim
```diff
@@ -3,6 +3,7 @@
 import results
 import chronicles, json_serialization, json_serialization/std/options
 import ../../../waku_node, ../serdes
+import ../../../api/types
```
Ivansete-status:
```suggestion
import waku/api/types
```

waku/waku_core/event/event_emitter.nim
```diff
@@ -0,0 +1,20 @@
+
+
+type
+  EventEmitter* = object
+    # Placeholder for future event emitter implementation
+    observers*: seq[proc (data: EventData): void]
+
```
Ivansete-status: This file wasn't parsed by `nph` for sure :)

```nim
find . ! -path "./vendor/*" -name "*.nim" -exec nph {} \;
```

Ivansete-status (commented):
Thanks for the PR!
I'm just adding some nitpicks that I hope you find useful.
I'll revisit the PR again once it gets ready :)

tests/node/test_wakunode_health_monitor.nim
fcecin: This test is mocking basically everything and is just testing the math behind the ConnectionStatus calculation. Here `DefaultRelayFailoverThreshold` is just a stand-in for the `DLow` config parameter for GossipSub. Since GossipSub doesn't exist here, we are just saying it is the default value for when an actual `DLow` config doesn't exist for some reason. The logic in `calculateConnectionState`, which is otherwise just a static calculator, depends on what `DLow` is, which is externally-provided from the node and the network.

waku/node/health_monitor/node_health_monitor.nim
fcecin: Renamed DefaultRelayFailoverThreshold as DLow and made it a proper Option[int] that is `none` in Edge.

tests/node/test_wakunode_health_monitor.nim
fcecin: Taking these out breaks the test unfortunately. The newTestWakuNode helper just passes this to the "enrBuilder"; it doesn't seem to execute relay subscriptions with it, like `setupAppCallbacks` eventually does for real nodes.

tests/node/test_wakunode_health_monitor.nim
fcecin: Changed from void Future to AsyncEvent to make it cleaner and clearer

waku/common/waku_protocol.nim
fcecin: That breaks the public API contract for `/health` endpoint, it will output e.g. "RelayProtocol" instead of the old "Relay". I wouldn't touch that stuff if we can avoid it.

waku/node/health_monitor/connection_status.nim
fcecin: Do we want a `waku/types/` folder? That would rhyme with `waku/events/` and `waku/requests/` (which erase the distinction of "API" requests/revents vs. "non-API" ones, which is probably what we want). Or is it `waku/api/types/` or maybe `waku/common/types/`?

waku/node/health_monitor/node_health_monitor.nim
fcecin: Now hm.node can no longer be nil so that check is gone. Then, if you don't have a peerManager (which never happens or should never happen) then you have zero peers -- you can't have peers, so the metric "0" is correct. Blowing up loudly on a failure to mount a peerManager into a node should be done in a more centralized place. The alternative is to not even check.

waku/node/health_monitor/node_health_monitor.nim
fcecin: Unfortunately you need it to be var for the clever e.g. notMounted() etc. mutators to work. I actually tried to do something more clever with how HealthStatus works (with `sink` so we can do both chaining and that "let" in the end) but nothing I tried worked (e.g. `sink` felt kind of dangerous, somehow I made it crash), so we are stuck with var or we change how this HealthStatus idiom works (with something like I tried to do with `sink`, but good).

waku/node/health_monitor/node_health_monitor.nim
fcecin: This is hooked to the Observer interface of WakuRelay. It looks like it's directly consuming libp2p, but it isn't. It's just the types are the same as those provided by libp2p to its direct apps, which is always confusing (why I always alias/replace all typenames on layer change if it's up to me...).

waku/node/peer_manager/peer_manager.nim
fcecin: Sorry, this brokerCtx coloring / DI / extra ctor arg stuff is garbage. That's not how Zoltan built context-aware brokers to be used. I'm fixing this.

waku/node/waku_node.nim
fcecin: This is in WakuNode because we don't have a "WakuEdge" virtual protocol or component to reflect all the stuff we centralize already in WakuRelay (which acts as a "WakuCore" component). In any case we can't move anything that depends heavily on data that is in the WakuNode (again, it could be in a "WakuEdge" to offload WakuNode a bit, just like WakuRelay offloads "WakuCore" logic from WakuNode), otherwise it will just add a level of "." to the entire business logic of request fulfillment and event firing (it needs to know about peer management, etc.). We can refactor all this in the future.

Also, `startProvidersAndListeners` is going to be very common everywhere when we're done refactoring everything to use brokering. All components will have something to say and to listen.

waku/waku_core/event/event_emitter.nim
fcecin: Fixed it. I'll do the mass-nph in a separate PR after we get the new nph release that is upcoming (or maybe we should do it twice, right now with the current nph and later we redo it).

waku/node/health_monitor/connection_status.nim
NagyZoltanPeter: I agree to have a more structured type folder.
And yes everything need to be known by API user shall be under `waku/api/types/...`
We might think of also `waku/api/events/...` too.
This would deliver the intent behind those definitions.

waku/waku_core/event/event_emitter.nim
NagyZoltanPeter, line 1: Is it used anywhere? Leftover?
Isn't EventBroker has this functionality?

waku/waku_relay/protocol.nim
```diff
@@ -437,38 +462,58 @@ proc calculateTopicHealth(wakuRelay: WakuRelay, topic: string): TopicHealth =
...
 
-    if oldHealth == currentHealth:
-      continue
+proc subscribedTopics*(w: WakuRelay): seq[PubsubTopic] =
+  return toSeq(GossipSub(w).topics.keys())
+
+proc topicsHealthLoop(w: WakuRelay) {.async.} =
+  while true:
+    await w.topicHealthUpdateEvent.wait()
+    w.topicHealthUpdateEvent.clear()
```
NagyZoltanPeter, line 474: It smells me we might need this loop at all, if we replace topicHealthUpdateEvent with EventBroker and listen to it to make action and emit the output event + call the observers.
(at some point it would be nice to replace those app observers with EventBroker as a common communication channel - not this PR definetely).

waku/node/health_monitor/node_health_monitor.nim
```diff
@@ -72,160 +106,437 @@ proc getRlnRelayHealth(hm: NodeHealthMonitor): Future[ProtocolHealth] {.async.}
...
+) {.gcsafe, raises: [].} =
+  if msg.subscriptions.len == 0:
+    if msg.control.isNone():
+      return
+    let ctrl = msg.control.get()
+    if ctrl.graft.len == 0 and ctrl.prune.len == 0:
+      return
+
+  # recomputing node health when peer relay events of interest trigger
+  hm.healthUpdateEvent.fire()
```
NagyZoltanPeter, line 516: I would need a bit more detailed comment here explaining why it is actually needed and what is the condition it looks for. 
This observer will be triggered a lot IMO.

waku/node/health_monitor/node_health_monitor.nim
NagyZoltanPeter, line 1: Just wonder if this can be simplified if using MultiRequestBroker with the intent each protocol should answer to the request if mounted (those NotMounted state can be derived from the list of expected protocol answers vs. actual answers). Such way we could delegate the health logic to protocols decoupling health monitor from the knowledge of them.
I would not block this aswome PR with such rework, just an idea.

waku/waku_relay/protocol.nim
```diff
@@ -287,6 +296,21 @@ proc initRelayObservers(w: WakuRelay) =
       )
 
   proc onRecv(peer: PubSubPeer, msgs: var RPCMsg) =
+    if msgs.control.isSome:
```
Ivansete-status: Sorry, super nitpick. Parenthesis with verbs, and no-parenthesis with nouns.
```suggestion
    if msgs.control.isSome():
```

tests/api/test_api_health.nim
```diff
@@ -0,0 +1,296 @@
...
+const TestContentTopic = ContentTopic("/waku/2/default-content/proto")
+
+proc dummyHandler(
+    topic: PubsubTopic, msg: WakuMessage
+): Future[void] {.async, gcsafe.} =
+  discard
+
+proc waitForConnectionStatus(
+    brokerCtx: BrokerContext, expected: ConnectionStatus
+): Future[void] {.async.} =
```
Ivansete-status:
```suggestion
) {.async.} =
```

waku/api/api.nim
```diff
@@ -25,16 +25,8 @@ proc checkApiAvailability(w: Waku): Result[void, string] =
...
-  # check if health is satisfactory
-  # If Node is not healthy, return err("Waku node is not healthy")
-  let healthStatus = RequestNodeHealth.request(w.brokerCtx)
-
-  if healthStatus.isErr():
-    warn "Failed to get Waku node health status: ", error = healthStatus.error
-    # Let's suppose the node is hesalthy enough, go ahead
-  else:
-    if healthStatus.get().healthStatus == NodeHealth.Unhealthy:
-      return err("Waku node is not healthy, has got no connections.")
```
Ivansete-status, line 37: Why not needed if may I ask?

waku/events/peer_events.nim
```diff
@@ -0,0 +1,13 @@
+import waku/common/broker/event_broker
+import libp2p/switch
+
+type WakuPeerEventKind* = enum
```
Ivansete-status:
```suggestion
type WakuPeerEventKind* {.pure.} = enum
```

waku/factory/waku.nim
```diff
@@ -409,60 +420,48 @@ proc startWaku*(waku: ptr Waku): Future[Result[void, string]] {.async: (raises:
...
+    globalBrokerContext(),
+    proc(
+        protocol: WakuProtocol
+    ): Future[Result[RequestProtocolHealth, string]] {.async.} =
+      try:
+        let protocolHealthStatus =
+          await waku[].healthMonitor.getProtocolHealthInfo(protocol)
+        ok(RequestProtocolHealth(healthStatus: protocolHealthStatus))
+      except CatchableError:
+        err("Failed to get protocol health: " & getCurrentExceptionMsg()),
```
Ivansete-status:
```suggestion
        return ok(RequestProtocolHealth(healthStatus: protocolHealthStatus))
      except CatchableError:
        return err("Failed to get protocol health: " & getCurrentExceptionMsg()),
```

waku/waku_relay/protocol.nim
fcecin: The goal of this AsyncEvent + loop pattern is to make the implementation of the event handler just an `AsyncEvent.fire()`, so the `asyncSpawn`ed task inside the broker does almost nothing and so the task can be collected faster.

We can discuss this pattern in an upcoming meeting and if we don't like it, we can kill this pretty fast with a maintenance PR.

What it has going for itself is that it gives us safety and certainty on how the event system will behave e.g. we get a barrage of events from libp2p and we don't just ripple this out unnecessarily to the app or spend too much in the handlers doing complicated and expensive things as a reaction to every single redundant state transition.

I was even thinking this loop infra could be hidden inside a macro/template that "requestifies and eventifies" middleware components and helps to wire them to each other, so it just becomes the way we use the broker library. Which would help us make the `asyncSpawn` inside the broker impl optional; it could work like e.g. Boost::Asio, where you can post a task to an IOContext or execute inline "in this task" and it will be fine -- in this case it would be fine because the inline would be just "AsyncEvent.fire()".

waku/waku_core/event/event_emitter.nim
fcecin: I deleted that EventEmitter (dead code).

(I edited this comment later, let me put this back in so the reply makes sense:) Let's follow up with some meeting time and maintenance PRs to move the types to better places. I have some questions, like, I'm not sure if any Request* will be pat of the API? Maybe all of them get wrapped in procs? Or even EventBroker, do we "hide" them in the LMD API behind some callback registration mechanic that's more familiar? The reverse is also possible: we could make the Nim API broker only, which would be very cool as I think you mentioned at some point and I think that's very possible as well.

waku/node/health_monitor/node_health_monitor.nim
fcecin: Yes, I saw the API hint you put there to use MultiRequestBroker. I tried doing it that way but hit a few roadblocks, like sync vs. async health metrics. We need first to rework how the protocols provide their health -- each one of the 15 protocols would be changed to provide in their own modules their own health metric instead of centralizing everything in the NodeHealthMonitor. Then the NodeHealthMonitor is deleted or becomes a shim. This is a pure refactor, so I'd rather not roll that into a PR that introduces large behavioral changes.

waku/node/health_monitor/node_health_monitor.nim
fcecin: Added comments.

Yes we have to assume this will be called "infinity times," so that's why it does almost nothing. It just checks if there's any mesh or subscription change whatsoever, and instead of trying to look into these and make complicated guesses on whether this "changes anything," it just schedules a recompute of health/state which we can do at our leisure and throttle at will as well.

waku/waku_relay/protocol.nim
NagyZoltanPeter: I see it as kind of equal. EventBroker listeners are fired with asyncSpawn by design, so the emitter is not effected by longer running event processors.
There are not much magic behind the wait/fire asyncEvent, due that is also built on future semaphore in the same chronos worker loop.
We can keep this one, it's really just a flavor. For me the EventBroker solution sounds more simple as it avoids `while true:` loop and other dependencies.

waku/node/health_monitor/node_health_monitor.nim
NagyZoltanPeter: Yeah, IDK how frequent such a update event will be fired. It would be good to have the `topicsHealthLoop ` processing fall over if within a certain amount of time, I don't think we need to recalculate health status more frequent than a second. WDYT?

waku/node/health_monitor/node_health_monitor.nim
NagyZoltanPeter: Agree!!!

waku/waku_core/event/event_emitter.nim
NagyZoltanPeter: Oh, yeah, current Event/Request -Brokers are only internal to a single execution unit/thread. So they cannott cross thread boundaries. As such we need to put interface at FFI level to translate Events to FFI callback events as of current form of FFI. So with requests, they can be issued upon incoming FFI call site.

waku/api/api.nim
fcecin: It is not "needed" since as we discussed in the last meeting, it can work either way. You can have it be nice and smart and reject a send because it knows the node is super dead right now, or you can have it be dumb and retry to send for 30 or 60 seconds because it doesn't know the difference between a permanent config failure and a temporary connectivity health condition that can be overcome.

However, it is definitely wanted: if we can know the connection is fully shot, then we should tell the app by either rejecting the send, or at least warning it that the node connection appears to be dead (which the app can also inspect via a health API request and via health events, if it can piece together the puzzle itself).

The reason we can't have it now is entirely pragmatic: writing a test that isn't tripped by this currently is super hard. We need a new test library for LMD API level testing that will get this right. There's some things we need to sort out before we can get another crack at this, and in any case this is a back-end improvement -- the API is still the same.

waku/node/health_monitor/connection_status.nim
fcecin: Let's follow up with some meeting time and maintenance PRs to move the types to better places. I have some questions, like, I'm not sure if any Request* will be pat of the API? Maybe all of them get wrapped in procs? Or even EventBroker, do we "hide" them in the LMD API behind some callback registration mechanic that's more familiar? The reverse is also possible: we could make the Nim API broker only, which would be very cool as I think you mentioned at some point and I think that's very possible as well.

waku/node/health_monitor/node_health_monitor.nim
fcecin: You're right! This is missing a sleep for sure, especially because most health calculations are sync. However 1s is too brutal IMO; we can initially gate this at 50ms I think, which in CPU terms is an eternity but still OK for human user feedback. We can increase this later if needed.

EDIT: Wait, no, this loop has a 100ms sleep already. So I think that's fine?

EDIT 2: The other two health loops are missing this; I'll add this sleep to them as well.

PR #3692, opened 2026-01-22
start using orc as memory management mode

Ivansete-status:
Follows a @NagyZoltanPeter's suggestion made in a meeting some time ago, to start using ORC.

Ref: https://nim-lang.org/blog/2020/10/15/introduction-to-arc-orc-in-nim.html

## Issue

### DIFF
--- a/waku.nimble
+++ b/waku.nimble
@@ -48,7 +48,7 @@ proc buildModule(filePath, params = "", lang = "c"): bool =
     echo "File to build not found: " & filePath
     return false
 
-  exec "nim " & lang & " --out:build/" & filepath & ".bin --mm:refc " & extra_params &
+  exec "nim " & lang & " --out:build/" & filepath & ".bin --mm:orc " & extra_params &
     " " & filePath
 
   # exec will raise exception if anything goes wrong
@@ -61,7 +61,7 @@ proc buildBinary(name: string, srcDir = "./", params = "", lang = "c") =
   var extra_params = params
   for i in 2 ..< paramCount():
     extra_params &= " " & paramStr(i)
-  exec "nim " & lang & " --out:build/" & name & " --mm:refc " & extra_params & " " &
+  exec "nim " & lang & " --out:build/" & name & " --mm:orc " & extra_params & " " &
     srcDir & name & ".nim"
 
 proc buildLibrary(lib_name: string, srcDir = "./", params = "", `type` = "static", srcFile = "libwaku.nim", mainPrefix = "libwaku") =
@@ -73,11 +73,11 @@ proc buildLibrary(lib_name: string, srcDir = "./", params = "", `type` = "static
     extra_params &= " " & paramStr(i)
   if `type` == "static":
     exec "nim c" & " --out:build/" & lib_name &
-      " --threads:on --app:staticlib --opt:size --noMain --mm:refc --header -d:metrics --nimMainPrefix:" & mainPrefix & " --skipParentCfg:on -d:discv5_protocol_id=d5waku " &
+      " --threads:on --app:staticlib --opt:size --noMain --mm:orc --header -d:metrics --nimMainPrefix:" & mainPrefix & " --skipParentCfg:on -d:discv5_protocol_id=d5waku " &
       extra_params & " " & srcDir & srcFile
   else:
     exec "nim c" & " --out:build/" & lib_name &
-      " --threads:on --app:lib --opt:size --noMain --mm:refc --header -d:metrics --nimMainPrefix:" & mainPrefix & " --skipParentCfg:off -d:discv5_protocol_id=d5waku " &
+      " --threads:on --app:lib --opt:size --noMain --mm:orc --header -d:metrics --nimMainPrefix:" & mainPrefix & " --skipParentCfg:off -d:discv5_protocol_id=d5waku " &
       extra_params & " " & srcDir & srcFile
 
 proc buildMobileAndroid(srcDir = ".", params = "") =
@@ -265,7 +265,7 @@ proc buildMobileIOS(srcDir = ".", params = "") =
       " --nimcache:" & nimcacheDir &
       " --os:ios --cpu:" & cpu &
       " --compileOnly:on" &
-      " --noMain --mm:refc" &
+      " --noMain --mm:orc" &
       " --threads:on --opt:size --header" &
       " -d:metrics -d:discv5_protocol_id=d5waku" &
       " --nimMainPrefix:libwaku --skipParentCfg:on" &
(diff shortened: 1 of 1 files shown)


fcecin (approved):
LGTM!

NagyZoltanPeter (approved):
I completely agree with this. Can we make our tests green somehow? I think that can give us some safety for this move.

darshankabariya (approved):
LGTM !!

Ivansete-status:
The tests are crashing due to that: https://github.com/nim-lang/Nim/issues/25693

Is pretty easy to replicate the crash. Just run
`make test tests/test_peer_store_extended.nim`

Ivansete-status:
We will invest time to fix that Nim crash around Y2026Q3

Ivansete-status:
The issue reported by @tersec (thanks for it! 🙌 ) https://github.com/nim-lang/Nim/issues/25693 , is covered by https://github.com/nim-lang/Nim/pull/25876 (thanks so much for the fix too!)

In turn, I've validated that this also prevents the crash from happening in our tests ( `tests/test_peer_store_extended.nim` .)

The fix will likely be introduced in 2.2.12.

------

  ## Some details about fix commit, in Nim repo (#25876)
  - Merged to `devel`: commit `7a5e35c8` (2026-06-08).
  - Backported to branch `version-2-2`: commit `8c02426` (+
  predecessor `ba047d0`).
  - Branch `version-2-2` tip at time of pinning:
  `2c6191aa4da97acdc6dc40e9a959e1ec75fae4d4`
    → `NimVersion == 2.2.11`.
  - **No released tag contains it** (v2.2.10 = unfixed).
  version-22-staging also lacked it.

Ivansete-status:
Still blocked on the toolchain: the fix for nim-lang/Nim#25693 (`8c02426`) is on
  `version-2-2` but unreleased — latest tag is `v2.2.10`, and the fix sits 35 commits
  ahead of it. We're pinned at 2.2.4, so this also implies a sizeable Nim bump.

fcecin:
FWIW @Ivansete-status the "build-health" job criteria isn't written in stone. It's just there to lock-in whatever we have decided. We can just "un-decide" whatever this is asserting and delete the assertion.

```
    FAIL  DEBUG=0 keeps link-time optimisation
            expected to contain: -d:lto_incremental
            actual: NIM_PARAMS :=  --verbosity:0 --hints:off -d:git_version=\"b66ade\" -d:release -d:strip -d:disable_libbacktrace 
```

EDIT: I'm not commenting on the lto_incremental situation specifically; this is a general comment that applies to all criteria in this check job.

tersec:
I can reproduce this, but it only occurs, in this context, with the signal handlers at https://github.com/logos-messaging/logos-delivery/blob/d037d9fb215bf7c32bf4a34e5ca7acab9157a80f/apps/logos_delivery_node/logosdeliverynode.nim#L59-L90 in place. These (likely) invoke undefined behavior per POSIX per https://pubs.opengroup.org/onlinepubs/9699919799/functions/V2_chap02.html#tag_15_04_03_03
> the behavior is undefined if the signal handler refers to any object other than errno with static storage duration other than by assigning a value to an object declared as **volatile sig_atomic_t**, or if the signal handler calls any function defined in this standard other than one of the functions listed in the following table.

Linux, per https://man7.org/linux/man-pages/man7/signal-safety.7.html has the same constraint:
```
       An async-signal-safe function is one that can be safely called
       from within a signal handler.  Many functions are not async-
       signal-safe.  In particular, nonreentrant functions are generally
       unsafe to call from a signal handler.

       The kinds of issues that render a function unsafe can be quickly
       understood when one considers the implementation of the stdio
       library, all of whose functions are not async-signal-safe.
```

There's a simpler non-UB repro, but regardless, this code is (likely; if not, it needs to be affirmatively shown, because it's quite suspicious and sloppy for a signal handler) buggy and without that, the ORC crash doesn't occur for me on 2.2.12.

fcecin:
> There's a simpler non-UB repro, but regardless, this code is (likely; if not, it needs to be affirmatively shown, because it's quite suspicious and sloppy for a signal handler) buggy and without that, the ORC crash doesn't occur for me on 2.2.12.

Yep, doing anything in a signal handler that isn't flipping a volatile boolean/int variable is (basically) evil. It's always sufficient to flip one variable there and then check/react elsewhere.

cc:  @Ivansete-status

tersec:
As far as I can tell, the underlying specific trigger is that nim-chronicles is non-reentrant on a thread, and this appears to be a crash in nim-chronicles, proximately, but the signal handlers doing what they do is... likely UB enough that the details of what nim-chronicles is or isn't doing aren't the key point here. This is not, directly, a bug in nim-chronicles, but likely improper API usage.

For example, this doesn't crash, but a similar mechanism of nested/re-entrant chronicles usage:
```nim
import chronicles
proc d(): int = info "y"
info "n", p = d()
```

outputs
```
INF 2026-09-19 13:13:50.807+00:00 n                                          tid=3197960INF 2026-09-19 13:13:50.807+00:00 y                                          tid=3197960
 p=0
```

This version just corrupts its own output, and is ... debatably a nim-chronicles issue (but comparatively harmless either way). But similar constructions combine with signals if they hit a certain race window to SIGSEGV:
```nim
import std/posix, chronicles
proc w(_: cint) {.noconv.} = info "v"
discard signal(SIGALRM, w)
let q = getpid()
if fork() == 0:
  while kill(q, SIGALRM) == 0: discard
  quit(0)
let v = newString(4096)
while true:
  info "d", r = v
```
which is, to be clear, UB, and with either plain `nim c -r` or `nim c --mm:refc -r` segfaults deterministically. With ORC it looks like
```
... loops ...

INF 18.809 v                                          tid=3199158
INF 18.809 v                                          tid=3199158
INF 18.809 v                                          tid=3199158
INF 18.809 v                                          tid=3199158
INF 18.809 v                                          tid=3199158
Segmentation fault
```
how often it has to loop varies, but it reliably crashes with a segfault within a few seconds for me (Nim 2.2.12, gcc 16, Debian Linux).

That the signal handler didn't happen to tickle whatever previous configuration(s) it's been tried on was luck.

Actually, _not_ segfaulting is worse: it means it corrupted memory within the process's own address space. The segfault is the good outcome.

tersec:
@Ivansete-status I don't see evidence LTO is the real bug here, re:
- https://github.com/logos-messaging/logos-delivery/pull/3692/commits/bbff5849997634f85efc1e4d5f491c9a2a8e21a0
- https://github.com/logos-messaging/logos-delivery/pull/3692/commits/9b8093e7acb47abc65d3e4d3f157da7cfbf4d827

and there is evidence for the signal handlers being responsible. See my comments.

Anecdotally, in my own testing, in the original logos-delivery context:
- LTO + signal handlers => segfaults
- yes, no-LTO + signal handlers => no segfaults, true, but also
- LTO + non-UB signal handlers => no segfaults

And I presented a mechanistic repro for the signal handler point, which does not depend on LTO, ORC, or even release mode, at all.

PR #3701, opened 2026-01-28
simple refactor to reduce PRs CI load

Ivansete-status:
- Adds daily run on `master` to build examples and tools. It is now sending notifications to discord, _always_, daily. Given that this is going to become spammy. We will need to change that to only notify on failure.

## Issue


NagyZoltanPeter (approved):
Great!

PR #3705, opened 2026-01-29
force epoll is used in chronos for Android

Ivansete-status:
Simple PR that follows this recommendation: https://github.com/status-im/nim-chronos/pull/616#issuecomment-3806271269

## Issue


NagyZoltanPeter (approved):
LGTM

arnetheduck:
if it turns out this works well, we should change the chronos default for android

PR #3714, opened 2026-02-06
feat: liblogosdelivery FFI library of new API

NagyZoltanPeter:
## Description
Create C-binding FFI for the new Waku/Logos-messaging API.

## Changes

- Added liblmapi
  - Featuring FFI interface and C header.
  - Example app that utilized existing API endpoints. `make liblmapi_example`
- new make target to build static and dynamic lmapi library. `make liblmapi` / `make liblmapi STATIC=1`
- new nix target `nix build '.?submodules=1#liblmapi'` 

## Issue

https://github.com/logos-messaging/logos-messaging-nim/issues/3709

### Notice
Expect some further refinement as I'm going forward to create `logos-message-module` for `logos-core`.

### DIFF
--- a/waku/waku_archive/driver/postgres_driver/postgres_driver.nim
+++ b/waku/waku_archive/driver/postgres_driver/postgres_driver.nim
@@ -1347,8 +1347,10 @@ proc removePartition(
   (await self.performWriteQuery(detachPartitionQuery)).isOkOr:
     info "detected error when trying to detach partition", error
 
-    if ($error).contains("FINALIZE") or
-        ($error).contains("already pending detach in part"):
+    if ($error).contains("FINALIZE") or ($error).contains("already pending"):
+      ## We assume "already pending detach in partitioned table ..." as possible error
+      debug "enforce detach with FINALIZE because of detected error", error
+
       ## We assume the database is suggesting to use FINALIZE when detaching a partition
       let detachPartitionFinalizeQuery =
         "ALTER TABLE messages DETACH PARTITION " & partitionName & " FINALIZE;"


liblogosdelivery/liblogosdelivery.h
igor-sirotin, line 1: I'm a bit confused though, why does it have to be a separate library?

I thought we want to just add the API into existing `libwaku` library. Then it would be up to the users which API to use — low- or high-level.

fcecin: Kind of a side point, but I was thinking that maybe we should gradually move towards hiding the "waku" string as much as possible, keeping that for internals, low-level, or "unsupported" (if you import and use something named "waku" directly then by default that's unsupported unless we say otherwise), and "lm" to name things that we definitely want people to see and use from now on.

So here this lib split would help to hide the "waku" string a bit more, if that's what we'd want.

Ivansete-status: The aim is to avoid any further _messaging_ consumers to use `libwaku` ( kernel API .) so that they are not tempted to used it.

NagyZoltanPeter: > I'm a bit confused though, why does it have to be a separate library?
> 
> I thought we want to just add the API into existing `libwaku` library. Then it would be up to the users which API to use — low- or high-level.

Well, we need to adjust our clocks than.
The original idea of this new API is to hide every complicated low level protocol and operations.
We identified through years and gathering inputs from hackathons that the whole waku concept needs huge background knowledge to use efficiently - even for us.
There are simple tooo much option and ways to use, which is confusing for anyone and makes a barrier to engage projects and developers.

So there had been a decision about to create a high level WakuAPI (Now LogosMessaging API), which eliminates the hurdles and provides a simplified interface (basically a simple init, subscribe, send, receive, retrieve).

To suggest by naming we renamed old nim api interface to KerenelAPI (that's exposed by libwaku).

Behind the surface, the new API intended to use the complex KernelAPI (as a row access to protocols) and make everything smooth to the library consumer.

This point to that direction, we do not want to promise anything about KernelAPI outward, that is open for change (however it is quite stable as is now). So we would like to lead out libwaku slowly and let Status/Chat/anybody only depend on this new API layer.
So it means a slim new library to deliver.

@fryorcraken idea was to first keep is as simple as possible, if in time any meaningful request come in, we can extend with care and consistent way, but do not expose Kernel layer.

cc: @Ivansete-status

NagyZoltanPeter: > Kind of a side point, but I was thinking that maybe we should gradually move towards hiding the "waku" string as much as possible, keeping that for internals, low-level, or "unsupported" (if you import and use something named "waku" directly then by default that's unsupported unless we say otherwise), and "lm" to name things that we definitely want people to see and use from now on.
> 
> So here this lib split would help to hide the "waku" string a bit more, if that's what we'd want.

Very good point, I don't have answer for it now. We renamed the repos and changed official naming but still have a lot of waku heritage in the code. 
`lm` seems a bit less informative what I'd expect from the public library though.

liblogosdelivery/examples/logosdelivery_example.c
```diff
@@ -0,0 +1,176 @@
...
+    if (ret != RET_OK || msg == NULL || len == 0) {
+        return;
+    }
+
+    // Create null-terminated string for easier parsing
+    char *eventJson = malloc(len + 1);
+    if (!eventJson) {
+        return;
+    }
+    memcpy(eventJson, msg, len);
```
Ivansete-status, line 48: Maybe `strncpy` is safer?

liblogosdelivery/examples/logosdelivery_example.c
```diff
@@ -0,0 +1,176 @@
...
+    buffer[len] = '\0';
+
+    return buffer;
+}
+
+// Event callback that handles message events
+void event_callback(int ret, const char *msg, size_t len, void *userData) {
+    if (ret != RET_OK || msg == NULL || len == 0) {
+        return;
+    }
```
Ivansete-status, line 41: I think is interesting to split that in three if statements and add a simple print for each.

liblogosdelivery/logos_delivery_api/messaging_api.nim
```diff
@@ -0,0 +1,86 @@
+import std/[json, base64]
```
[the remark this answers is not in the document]
Ivansete-status: Isn't that copilot a little bit spammy in general?

liblogosdelivery/logos_delivery_api/messaging_api.nim
```diff
@@ -0,0 +1,86 @@
...
+import chronos, results, ffi
+import stew/byteutils
+import
+  waku/factory/waku,
+  waku/waku_core/topics/content_topic,
+  waku/api/[api, types],
+  ../declare_lib
```
Ivansete-status, line 9: For new imports, let's use "absolute" path as much as possible
```suggestion
  lmapi/declare_lib
```

liblogosdelivery/logos_delivery_api/messaging_api.nim
```diff
@@ -0,0 +1,86 @@
...
+    let errMsg = $error
+    return err("Unsubscribe failed: " & errMsg)
+
+  return ok("")
+
+proc lmapi_send(
+    ctx: ptr FFIContext[Waku],
+    callback: FFICallBack,
+    userData: pointer,
+    messageJson: cstring,
```
Ivansete-status, line 51: Shall we split the possible params instead of using JSON message?
That can help to guide the users a little bit more.

liblogosdelivery/logos_delivery_api/node_api.nim
```diff
@@ -0,0 +1,155 @@
+import std/[json, options]
+import chronos, results, ffi
+import
+  waku/factory/waku,
+  waku/node/waku_node,
+  waku/api/[api, api_conf, types],
+  waku/events/message_events,
+  ../declare_lib,
+  ../json_event
```
Ivansete-status, line 9:
```suggestion
  lmapi/declare_lib,
  lm/json_event
```

liblmapi/lmapi/node_api.nim
```diff
@@ -0,0 +1,155 @@
...
+    ## Parse the JSON configuration and create a node
+    var jsonNode: JsonNode
+    try:
+      jsonNode = parseJson($configJson)
+    except Exception as e:
+      return err("Failed to parse config JSON: " & e.msg)
+
+    # Extract basic configuration
+    let mode =
+      if jsonNode.hasKey("mode") and jsonNode["mode"].getStr() == "Edge":
```
Ivansete-status: Maybe we can add some case protection.
```suggestion
      if jsonNode.hasKey("mode") and jsonNode["mode"].getStr().toLower() == "edge":
```

liblmapi/lmapi/node_api.nim
```diff
@@ -0,0 +1,155 @@
...
+
+    return ok("")
+
+proc lmapi_create_node(
+    configJson: cstring, callback: FFICallback, userData: pointer
+): pointer {.dynlib, exportc, cdecl.} =
+  initializeLibrary()
+
+  if isNil(callback):
+    echo "error: missing callback in lmapi_create_node"
```
Ivansete-status: Shall we use the `chronicles.error` as well?

liblogosdelivery/logos_delivery_api/node_api.nim
```diff
@@ -0,0 +1,155 @@
...
+    chronicles.error "START_NODE failed", err = errMsg
+    return err("failed to start: " & errMsg)
+  return ok("")
+
+proc lmapi_stop_node(
+    ctx: ptr FFIContext[Waku], callback: FFICallBack, userData: pointer
+) {.ffi.} =
+  MessageErrorEvent.dropAllListeners(ctx.myLib[].brokerCtx)
+  MessageSentEvent.dropAllListeners(ctx.myLib[].brokerCtx)
+  MessagePropagatedEvent.dropAllListeners(ctx.myLib[].brokerCtx)
```
Ivansete-status, line 105: Shall we move this snippet into the waku_node::stop() proc?

liblogosdelivery/BUILD.md
```diff
@@ -0,0 +1,164 @@
...
+    -llmapi \
+    -Wl,-rpath,./build
+
+# Or manually on Linux:
+gcc -o build/liblmapi_example \
+    liblmapi/examples/liblmapi_example.c \
+    -I./liblmapi \
+    -L./build \
+    -llmapi \
+    -Wl,-rpath='$ORIGIN'
```
Ivansete-status: Oh interesting. What is this `ORIGIN` variable?

liblogosdelivery/json_event.nim
```diff
@@ -0,0 +1,27 @@
+import std/[json, macros]
+
+type JsonEvent*[T] = ref object
+  eventType*: string
+  payload*: T
+
+macro toFlatJson*(event: JsonEvent): JsonNode =
```
Ivansete-status, line 7:
```suggestion
macro toFlatJson(event: JsonEvent): JsonNode =
```

liblmapi/liblmapi.nim
```diff
@@ -0,0 +1,29 @@
...
+
+  ffi.destroyFFIContext(ctx).isOkOr:
+    let msg = "liblmapi error: " & $error
+    callback(RET_ERR, unsafeAddr msg[0], cast[csize_t](len(msg)), userData)
+    return RET_ERR
+
+  ## always need to invoke the callback although we don't retrieve value to the caller
+  callback(RET_OK, nil, 0, userData)
+
+  return RET_OK
```
Ivansete-status: We can move this one, and `lmapi_create` into a new `lmapi/factory_api.nim`.

liblogosdelivery/README.md
```diff
@@ -0,0 +1,262 @@
+# Logos Messaging API (LMAPI) Library
```
Ivansete-status, line 1: I wouldn't add this README.md file so that we have less to maintain.
We can have this information in `liblmapi.h` and `liblmapi_example.c`.

nix/default.nix
```diff
@@ -94,8 +94,9 @@ in stdenv.mkDerivation {
     # Copy library files
     cp build/* $out/bin/ 2>/dev/null || true
 
-    # Copy the header file
-    cp library/libwaku.h $out/include/
+    # Copy header files
+    cp library/libwaku.h $out/include/ 2>/dev/null || true
+    cp liblmapi/liblmapi.h $out/include/ 2>/dev/null || true
```
Ivansete-status: I think we should best enforce these are exposed

```suggestion
    cp library/libwaku.h $out/include/
    cp liblmapi/liblmapi.h $out/include/
```

waku/api/api.nim
```diff
@@ -4,6 +4,7 @@ import waku/factory/waku
 import waku/[requests/health_request, waku_core, waku_node]
 import waku/node/delivery_service/send_service
 import waku/node/delivery_service/subscription_service
+import libp2p/peerid
```
Ivansete-status, line 7: Maybe a leftover?

Makefile
```diff
@@ -452,11 +453,40 @@ endif
 LIB_EXT := $(LIB_EXT_DYNAMIC)
 ifeq ($(STATIC), 1)
 	LIB_EXT = $(LIB_EXT_STATIC)
-	BUILD_COMMAND = libwakuStatic
+	LIBWAKU_BUILD_COMMAND = libwakuStatic
+	LMAPI_BUILD_COMMAND = liblmapiStatic
```
Ivansete-status: To keep consistency even though is a twister tongue
```suggestion
	LIBLMAPI_BUILD_COMMAND = liblmapiStatic
```

Ivansete-status (approved):
Wonderful PR! Thanks for it! 🥳 
Just adding some nitpicks that I hope you find useful.
Cheers

liblogosdelivery/logos_delivery_api/messaging_api.nim
NagyZoltanPeter: Well, it is contradict to what Send API spec says.
We have a send(MessageEnvelop) where the MessagEnvelop is composed of content_topic and payload...
I can be ok to provide overloads but than on both sides (in LMN Api as well).
Now you see why I don't like json ipc.... :-)

liblogosdelivery/logos_delivery_api/node_api.nim
NagyZoltanPeter: Hm... I think it's logically better here, otherwise you need to set up listeners on this (client side) of the api but will be dropped inside.
These listeners are for tunneling API events to FFI callback.

liblogosdelivery/BUILD.md
NagyZoltanPeter: Ah, these are leftovers as all these moved into makefile already.

waku/api/api.nim
NagyZoltanPeter: Actually libp2p/peerid defines chronicle formatting for peerId, without it just prints array of ints....

liblogosdelivery/logos_delivery_api/messaging_api.nim
Ivansete-status: haha yes I understand 100% :)
In that case, I think is fine with the internal control that the given json format is correct. As it is suggested in the PR.

liblogosdelivery/liblogosdelivery.h
igor-sirotin: @NagyZoltanPeter we are on the same page about hiding protocol complexities behind Messaging API, I am not appealing to this.

I am only questioning if we need it as a separate library. I was seeing it this way:
1. We introduce a new API to existing library in version X
2. Clients can upgrade to version X, they applications don't break. Yet now they can also use the new Messaging API.

If we believe that Kernel API should not be used by anyone, then we should not even deliver such a library.
If we assume some developers might want it to use Kernel API, then I think it is better to keep it in the same library.

I am on the side of giving options to developers. And worst case, we can always just deprecate and remove Kernel API from the "joined library".

And having/supporting/maintaining a single library should be simpler that 2 of them.

---

> we do not want to promise anything about KernelAPI outward, that is open for change

We can explicitly state that KernelAPI can be used on own risk, and that it can change at any time without notification. Though I agree with this argument, if you plan to make big changes to Kernel API, then it's reasonable to completely separate from it.

> So it means a slim new library to deliver.

Also, do you think that _library size_ can be significantly smaller, if we exclude KernelAPI from it?

---

I don't have a strong opinion. 
I prefer the simplest path, but that it will not shoot us in the leg later 🙂

fcecin: IMO both options are technically valid and this looks more like an aesthetic or marketing decision. I think the two-lib approach is easier to justify and to edge out the simpler monolib approach if we want to highlight the separation between Messaging API and Waku API (or "kernel API," same thing -- "Kernel" was coined before the project-wide rename and the subsequent idea of just using Waku to mean the core):

```
Old:           New:
Waku API   ->  Messaging API   (easy API for apps and app SDKs)
Kernel API ->  Waku API        (Waku "is-a" Kernel, 1:1)
```

So now "Waku" would mean the old level of service that allows you to mount and configure individual Vac protocols. It is the API layer you use to debug Vac implementations and to internally test middleware ideas, not to build apps.

And "Logos Messaging" or "Messaging" or etc. is the product used by apps. It has an opinion on how Waku (which would now be the the name of the core) should operate. The new layer does what we expected the layer of SDKs built on top of nwaku to do.

That's why I want to steal the Waku string for internal usage. I think we should retcon it to mean anything that's an internal object, instead of saying it is some kind of random legacy name that surfaces once in a while, making it easier to write consistent documentation and tutorials. Waku would be alive as a meaningful, formally-defined term: it means the Vac protocol implementation; any component built with direct (white-box) awareness of the lower-level Vac reality.

This will also make it so much easier to talk about what's going on: logos-messaging-nim is a tool with two products in it: one supported (Logos Messaging), and one internal and unsupported for app use (Waku) which is just the inner product used to build the outer product. And it'd be two libs to remind us of that. It was born as one product, and got another product layered on top of it and rolled into the same project, so now the new product is the sole consumer of the old, stable product (which is unsupported for external use and has no interface contracts to maintain -- use at your own risk).

NagyZoltanPeter: > @NagyZoltanPeter we are on the same page about hiding protocol complexities behind Messaging API, I am not appealing to this.
> 
> I am only questioning if we need it as a separate library. I was seeing it this way:
> 
> 1. We introduce a new API to existing library in version X
> 2. Clients can upgrade to version X, they applications don't break. Yet now they can also use the new Messaging API.
> 
> If we believe that Kernel API should not be used by anyone, then we should not even deliver such a library. If we assume some developers might want it to use Kernel API, then I think it is better to keep it in the same library.
> 
> I am on the side of giving options to developers. And worst case, we can always just deprecate and remove Kernel API from the "joined library".
> 
> And having/supporting/maintaining a single library should be simpler that 2 of them.
> 
> > we do not want to promise anything about KernelAPI outward, that is open for change
> 
> We can explicitly state that KernelAPI can be used on own risk, and that it can change at any time without notification. Though I agree with this argument, if you plan to make big changes to Kernel API, then it's reasonable to completely separate from it.
> 
> > So it means a slim new library to deliver.
> 
> Also, do you think that _library size_ can be significantly smaller, if we exclude KernelAPI from it?
> 
> I don't have a strong opinion. I prefer the simplest path, but that it will not shoot us in the leg later 🙂

My 2cs is, same as @fcecin wrote, I'm more on the strict approach side, that shall not mix the to kind of API's.
It is because the real meaning of new simplistic API is to engage devs to use our module. I believe the new API should fit for 98% of the application that needs reliable P2P message transport. Also we simplified configuration to remove doubts on how to make it properly (hence we added `mode` and `preset`).
Also behind the scenes we have our view of how to use properly our low level protocols to extend each other in various conditions.

I'm a bit against mixing the new API with the old because as I see API is always a contract (have I said it before :-D ). So when you read it, it must be obvious how to use it. I think if devs needs to understand the different approaches through long doc pages they may turn to other libs that promise easier results.

Still there are two approaches if anything won't fit with new API.
- extend the LM Api with desired functionality.
- as we are open source any one can fork it...

Do you think having libwaku and liblmapi for a while parallel has disadvantage?

igor-sirotin: As I said, I don't have a strong opinion on this. So if you guys prefer 2 libraries, then let's do 2 libraries 👍 
Shared my opinion in case it resonates with you.

fcecin (approved):

nix/submodules.json
```diff
@@ -0,0 +1,247 @@
+[
+  {
+    "path": "vendor/db_connector",
```
jakubgs, line 3: Why was this file added? I don't see it being used.

nix/default.nix
```diff
@@ -91,11 +95,39 @@ in stdenv.mkDerivation {
...
+    # Copy artifacts from build directory (created by Make during buildPhase)
+    # Note: build/ is in the source tree, not result/ (which is a post-build symlink)
+    if [ -d build ]; then
+      ${lib.optionalString copyLibwaku ''
+      cp build/libwaku.{so,dylib,dll,a,lib} $out/bin/ 2>/dev/null || true
+      ''}
+
+      ${lib.optionalString copyLiblogosdelivery ''
+      cp build/liblogosdelivery.{so,dylib,dll,a,lib} $out/bin/ 2>/dev/null || true
+      ''}
```
jakubgs, line 107: Why are we putting libraries into `bin` instead of `lib`?

nix/default.nix
```diff
@@ -91,11 +95,39 @@ in stdenv.mkDerivation {
...
+    cp library/libwaku.h $out/include/ 2>/dev/null || true
+    ''}
+
+    ${lib.optionalString copyLiblogosdelivery ''
+    cp liblogosdelivery/liblogosdelivery.h $out/include/ 2>/dev/null || true
+    ''}
+
+    ${lib.optionalString (!hasKnownInstallTarget) ''
+    cp library/libwaku.h $out/include/ 2>/dev/null || true
+    cp liblogosdelivery/liblogosdelivery.h $out/include/ 2>/dev/null || true
```
jakubgs, line 129: All of this logic you've added an be replaced with something as simple as:
```sh
    mkdir -p $out/bin $out/lib $out/include
    cp build/waku* $out/bin/ || true
    cp build/lib* $out/lib/ || true
    cp lib*/lib*.h $out/include/
```

Issue #3686, opened 2026-01-08
Maintenance Y2026H1


Ivansete-status:
## Description
Gather all maintenance tasks during first half of 2026.
All maintenance PRs should be linked to this so that we don't have orphan PRs.

## PRs
- https://github.com/logos-messaging/logos-messaging-nim/pull/3673
- https://github.com/logos-messaging/logos-messaging-nim-compose/pull/184
- https://github.com/logos-messaging/logos-messaging-nim-compose/pull/185
- https://github.com/logos-messaging/nim-sds/pull/42
- https://github.com/logos-messaging/logos-messaging-simulator/pull/116
- https://github.com/logos-messaging/logos-messaging-rlnv2-contract/pull/50
- https://github.com/logos-messaging/logos-messaging-nim/pull/3692
- https://github.com/logos-messaging/logos-messaging-nim/pull/3701
- https://github.com/logos-messaging/logos-messaging-nim/pull/3704
- https://github.com/logos-messaging/logos-messaging-nim/pull/3705
- https://github.com/logos-messaging/logos-messaging-nim/pull/3707
- https://github.com/logos-messaging/logos-messaging-nim/pull/3711
- https://github.com/logos-messaging/logos-messaging-nim/pull/3716
- https://github.com/logos-messaging/logos-messaging-nim/issues/3679
- https://github.com/logos-messaging/logos-messaging-nim/pull/3722
- https://github.com/logos-messaging/logos-messaging-nim/pull/3728
- https://github.com/logos-messaging/logos-delivery/pull/3729
- https://github.com/logos-messaging/logos-delivery/pull/3743
- https://github.com/logos-messaging/logos-delivery/pull/3759
- https://github.com/logos-messaging/logos-delivery/pull/3771
- https://github.com/logos-messaging/logos-delivery/pull/3792
- https://github.com/logos-messaging/logos-delivery/pull/3806
- https://github.com/logos-messaging/logos-delivery/pull/3809
- https://github.com/logos-messaging/logos-delivery/pull/3823
- https://github.com/logos-messaging/logos-delivery/pull/3839
- https://github.com/logos-messaging/logos-delivery/pull/3849

Ivansete-status:
H1 2026 is over. Remaining open sub-issues have been moved to the successor: https://github.com/logos-messaging/logos-delivery/issues/4043

Closed sub-issues and the PR list above are kept here as the H1 record.

PR #3716, opened 2026-02-09
Simplify NodeHealthMonitor creation

fcecin:
## Description

This PR simplifies the NodeHealthMonitor by removing the possibility of having a partially set up health monitor without a WakuNode ref. This allows us to remove all checks for NodeHealthMonitor.node being nil inside the health monitor.

(NOTE: I'm experiencing Anvil-related crashes in local testing in the test_rest_health.nim test. These happen in current master for me as well, and it's mostly because by chance that test seems to be the first one to try and use Anvil. If I remove Anvil / RLN from that test, it just fails in the next one that uses it. Can be a problem in my environment but I'm not sure. UPDATE: I have confirmed [i.e. I'm satisfied, after doing some testing] that this anvil failure is unrelated to any changes in our code.)

## Changes

* Force NodeHealthMonitor.new() to set up a WakuNode
* Remove all checks for isNil(node) in NodeHealthMonitor
* Fix tests to use the new NodeHealthMonitor.new()

## Issue

Maintenance Y2026H1 #3686

### DIFF
--- a/tests/wakunode_rest/test_rest_health.nim
+++ b/tests/wakunode_rest/test_rest_health.nim
@@ -73,6 +73,9 @@ suite "Waku v2 REST API - health":
     node.mountLightPushClient()
     await node.mountFilterClient()
 
+    # We don't have a Waku, so we need to set the overall health to READY here in its behalf
+    healthMonitor.setOverallHealth(HealthStatus.READY)
+
     # When
     var response = await client.healthCheck()
 


waku/factory/waku.nim
```diff
@@ -172,7 +172,13 @@ proc new*(
   ?wakuConf.validate()
   wakuConf.logConf()
 
-  let healthMonitor = NodeHealthMonitor.new(wakuConf.dnsAddrsNameServers)
+  var relay = newCircuitRelay(wakuConf.circuitRelayClient)
```
Ivansete-status: Maybe :)?
```suggestion
  let relay = newCircuitRelay(wakuConf.circuitRelayClient)
```

Ivansete-status (approved):
Superb! Thanks so much for it! 🙌

PR #3728, opened 2026-02-16
force FINALIZE partition detach after detecting shorter error

Ivansete-status:
Enforce partition detach with `FINALIZE` in postgres.

The issue happened the error message returned by the DB was one char shorter (dunno why.)

In this PR we are reducing the error message to consider that we need to force "detach FINALIZE".

See an screnshot of the new error message which we didn't detect properly and hence the node didn't try to force "detach FINALIZE".

image

## Related issue


igor-sirotin (approved):
Thanks for the finding!

Ivansete-status:
@darshankabariya - kindly cherry-pick this change into the release v0.38.0

Issue #3710, opened 2026-02-02
feat: LogosMessaging API: implement subscribe and recevie API


NagyZoltanPeter:
### Problem

Continuation of #3432 
Implement subscribe interface as of https://github.com/logos-messaging/specs/pull/86

Implement receive interface - event emitting for incoming messages.

### Additional context
subscribe interface hides the differences in core and edge mode subscribe handling (including Filter service peer management and subscription maintenance).
From App dev POV it shall look like the same.
- is manual (call to `subscribe`) interface
- is automatically done when sending a message under a content_topic is not yet managed/subscribed

receive message interface is based upon event driven concept. User of the lib shall be able to listen to `OnMessageReceived` events.`
Such and event must be emitted in case a message from relay network or from filter push is arrived and matched subscribe criteria.
Reliable receive shall involve inital and periodic check to available store service on the network to ensure no missed messages.

### Epic
TBD...

igor-sirotin:
@Ivansete-status I'd keep it named `Messaging API`. This name still makes sense. 
While `Delivery API` name kinda intends more guarantees, which Reliable Channels API will take care of.

Also, renaming this everywhere... let's not rename anything more please 😄

PR #3732, opened 2026-02-20
Implement stateful SubscriptionService for Core mode

fcecin:
## Description

This PR implements Messaging API (MAPI) content topic subscribe and unsubscribe and high-level message receipt for Core mode (Relay).

NOTE: This is just the (Relay/Core, for now) Nim back-end. This is not wiring up to any MAPI C FFI / bindings (the FFI back-end just has to listen to the `MessageReceivedEvent` and forward MAPI C calls to the Nim `subscribe()` and `unsubscribe()` APIs).

## Changes

* SubscriptionService tracks shard and content topic interest
* RecvService emits `MessageReceivedEvent` on subscribed content topics
* Route MAPI through old Kernel API relay unique-handler infra to avoid code duplication
* Encode current gen-zero network policy: on Core node boot, subscribe to all pubsub topics (all shards)
* Add test_api_subscriptions.nim (basic relay/core testing only)
* Removed any MAPI Edge sub/unsub/receive support code that was there (will add in next PR)

## Issue

Iterates #3710

To properly close the issue, MAPI Edge mode (filter/lightpush) support for sub/unsub/receive must be implemented.

### DIFF
--- a/waku/api/api.nim
+++ b/waku/api/api.nim
@@ -3,7 +3,7 @@ import chronicles, chronos, results, std/strutils
 import waku/factory/waku
 import waku/[requests/health_requests, waku_core, waku_node]
 import waku/node/delivery_service/send_service
-import waku/node/delivery_service/subscription_service
+import waku/node/delivery_service/subscription_manager
 import libp2p/peerid
 import ./[api_conf, types]
 
@@ -36,24 +36,24 @@ proc subscribe*(
 ): Future[Result[void, string]] {.async.} =
   ?checkApiAvailability(w)
 
-  return w.deliveryService.subscriptionService.subscribe(contentTopic)
+  return w.deliveryService.subscriptionManager.subscribe(contentTopic)
 
 proc unsubscribe*(w: Waku, contentTopic: ContentTopic): Result[void, string] =
   ?checkApiAvailability(w)
 
-  return w.deliveryService.subscriptionService.unsubscribe(contentTopic)
+  return w.deliveryService.subscriptionManager.unsubscribe(contentTopic)
 
 proc send*(
     w: Waku, envelope: MessageEnvelope
 ): Future[Result[RequestId, string]] {.async.} =
   ?checkApiAvailability(w)
 
-  let isSubbed = w.deliveryService.subscriptionService
+  let isSubbed = w.deliveryService.subscriptionManager
     .isSubscribed(envelope.contentTopic)
     .valueOr(false)
   if not isSubbed:
     info "Auto-subscribing to topic on send", contentTopic = envelope.contentTopic
-    w.deliveryService.subscriptionService.subscribe(envelope.contentTopic).isOkOr:
+    w.deliveryService.subscriptionManager.subscribe(envelope.contentTopic).isOkOr:
       warn "Failed to auto-subscribe", error = error
       return err("Failed to auto-subscribe before sending: " & error)
 
--- a/waku/factory/waku.nim
+++ b/waku/factory/waku.nim
@@ -35,7 +35,7 @@ import
     node/health_monitor,
     node/waku_metrics,
     node/delivery_service/delivery_service,
-    node/delivery_service/subscription_service,
+    node/delivery_service/subscription_manager,
     rest_api/message_cache,
     rest_api/endpoint/server,
     rest_api/endpoint/builder as rest_server_builder,
--- a/waku/node/delivery_service/recv_service/recv_service.nim
+++ b/waku/node/delivery_service/recv_service/recv_service.nim
@@ -4,7 +4,7 @@
 
 import std/[tables, sequtils, options, sets]
 import chronos, chronicles, libp2p/utility
-import ../[subscription_service]
+import ../[subscription_manager]
 import
   waku/[
     waku_core,
@@ -38,7 +38,7 @@ type RecvService* = ref object of RootObj
   brokerCtx: BrokerContext
   node: WakuNode
   seenMsgListener: MessageSeenEventListener
-  subscriptionService: SubscriptionService
+  subscriptionManager: SubscriptionManager
 
   recentReceivedMsgs: seq[RecvMessage]
 
@@ -91,7 +91,7 @@ proc msgChecker(self: RecvService) {.async.} =
     self.endTimeToCheck = getNowInNanosecondTime()
 
     var msgHashesInStore = newSeq[WakuMessageHash](0)
-    for sub in self.subscriptionService.getActiveSubscriptions():
+    for sub in self.subscriptionManager.getActiveSubscriptions():
       let storeResp: StoreQueryResponse = (
         await self.node.wakuStoreClient.queryToAny(
           StoreQueryRequest(
@@ -142,7 +142,7 @@ proc processIncomingMessageOfInterest(
     self.recentReceivedMsgs.add(rxMsg)
     MessageReceivedEvent.emit(self.brokerCtx, msgHash.to0xHex(), message)
 
-proc new*(T: typedesc[RecvService], node: WakuNode, s: SubscriptionService): T =
+proc new*(T: typedesc[RecvService], node: WakuNode, s: SubscriptionManager): T =
   ## The storeClient will help to acquire any possible missed messages
 
   let now = getNowInNanosecondTime()
@@ -150,7 +150,7 @@ proc new*(T: typedesc[RecvService], node: WakuNode, s: SubscriptionService): T =
     node: node,
     startTimeToCheck: now,
     brokerCtx: node.brokerCtx,
-    subscriptionService: s,
+    subscriptionManager: s,
     recentReceivedMsgs: @[],
   )
 
@@ -173,9 +173,11 @@ proc startRecvService*(self: RecvService) =
   self.seenMsgListener = MessageSeenEvent.listen(
     self.brokerCtx,
     proc(event: MessageSeenEvent) {.async: (raises: []).} =
-      if not self.subscriptionService.isSubscribed(
+      if not self.subscriptionManager.isSubscribed(
         event.topic, event.message.contentTopic
       ):
+        trace "skipping message as I am not subscribed",
+          shard = event.topic, contenttopic = event.message.contentTopic
         return
 
       self.processIncomingMessageOfInterest(event.topic, event.message),
--- a/waku/node/delivery_service/subscription_manager.nim
+++ b/waku/node/delivery_service/subscription_manager.nim
@@ -10,20 +10,20 @@ import
     events/delivery_events,
   ]
 
-type SubscriptionService* = ref object of RootObj
+type SubscriptionManager* = ref object of RootObj
   node: WakuNode
   contentTopicSubs: Table[PubsubTopic, HashSet[ContentTopic]]
     ## Map of Shard to ContentTopic needed because e.g. WakuRelay is PubsubTopic only.
     ## A present key with an empty HashSet value means pubsubtopic already subscribed
     ## (via subscribePubsubTopics()) but there's no specific content topic interest yet.
 
-proc new*(T: typedesc[SubscriptionService], node: WakuNode): T =
-  SubscriptionService(
+proc new*(T: typedesc[SubscriptionManager], node: WakuNode): T =
+  SubscriptionManager(
     node: node, contentTopicSubs: initTable[PubsubTopic, HashSet[ContentTopic]]()
   )
 
 proc addContentTopicInterest(
-    self: SubscriptionService, shard: PubsubTopic, topic: ContentTopic
+    self: SubscriptionManager, shard: PubsubTopic, topic: ContentTopic
 ): Result[void, string] =
   if not self.contentTopicSubs.hasKey(shard):
     self.contentTopicSubs[shard] = initHashSet[ContentTopic]()
@@ -38,7 +38,7 @@ proc addContentTopicInterest(
   return ok()
 
 proc removeContentTopicInterest(
-    self: SubscriptionService, shard: PubsubTopic, topic: ContentTopic
+    self: SubscriptionManager, shard: PubsubTopic, topic: ContentTopic
 ): Result[void, string] =
   self.contentTopicSubs.withValue(shard, cTopics):
     if cTopics[].contains(topic):
@@ -52,11 +52,11 @@ proc removeContentTopicInterest(
 
   return ok()
 
-proc subscribePubsubTopics*(
-    self: SubscriptionService, shards: seq[PubsubTopic]
+proc subscribePubsubTopics(
+    self: SubscriptionManager, shards: seq[PubsubTopic]
 ): Result[void, string] =
   if isNil(self.node.wakuRelay):
-    return err("subscribeShard requires a Relay")
+    return err("subscribePubsubTopics requires a Relay")
 
   var errors: seq[string] = @[]
 
@@ -73,41 +73,33 @@ proc subscribePubsubTopics*(
 
   return ok()
 
-proc startSubscriptionService*(self: SubscriptionService) =
-  if not isNil(self.node.wakuRelay):
-    if self.node.wakuAutoSharding.isSome():
-      # Subscribe relay to all shards in autosharding.
-      let autoSharding = self.node.wakuAutoSharding.get()
-      let clusterId = autoSharding.clusterId
-      let numShards = autoSharding.shardCountGenZero
+proc startSubscriptionManager*(self: SubscriptionManager) =
+  if isNil(self.node.wakuRelay):
+    return
 
-      if numShards > 0:
-        var clusterPubsubTopics = newSeqOfCap[PubsubTopic](numShards)
+  if self.node.wakuAutoSharding.isSome():
+    # Subscribe relay to all shards in autosharding.
+    let autoSharding = self.node.wakuAutoSharding.get()
+    let clusterId = autoSharding.clusterId
+    let numShards = autoSharding.shardCountGenZero
 
-        for i in 0 ..< numShards:
-          let shardObj = RelayShard(clusterId: clusterId, shardId: uint16(i))
-          clusterPubsubTopics.add(PubsubTopic($shardObj))
+    if numShards > 0:
+      var clusterPubsubTopics = newSeqOfCap[PubsubTopic](numShards)
 
-        self.subscribePubsubTopics(clusterPubsubTopics).isOkOr:
-          error "Failed to auto-subscribe Relay to cluster shards: ", error = error
-    else:
-      # NOTE: We can't fallback to configured shards when no autosharding here since
-      #       we don't currently have access to Waku.conf here. However, we don't support
-      #       manual/static sharding at the MAPI level anyway so wiring that up now is not needed.
-      #       When we no longer auto-subscribe to all shards in Core boot, we will probably
-      #       scan the shard config due to fleet nodes; then shard conf will have to be reachable here.
-      #       For non-fleet, interactive Core nodes (e.g. Desktop apps) this can't matter
-      #       as much since shard subscriptions originate from subscription to content topics, but
-      #       I guess even in that case subbing to some conf shards may make sense for some apps.
-      info "SubscriptionService has no AutoSharding for Relay, won't subscribe to shards by default."
+      for i in 0 ..< numShards:
+        let shardObj = RelayShard(clusterId: clusterId, shardId: uint16(i))
+        clusterPubsubTopics.add(PubsubTopic($shardObj))
 
-  discard
+      self.subscribePubsubTopics(clusterPubsubTopics).isOkOr:
+        error "Failed to auto-subscribe Relay to cluster shards: ", error = error
+  else:
+    info "SubscriptionManager has no AutoSharding configured; skipping auto-subscribe."
 
-proc stopSubscriptionService*(self: SubscriptionService) {.async.} =
+proc stopSubscriptionManager*(self: SubscriptionManager) {.async.} =
   discard
 
 proc getActiveSubscriptions*(
-    self: SubscriptionService
+    self: SubscriptionManager
 ): seq[tuple[pubsubTopic: string, contentTopics: seq[ContentTopic]]] =
   var activeSubs: seq[tuple[pubsubTopic: string, contentTopics: seq[ContentTopic]]] =
     @[]
@@ -122,16 +114,16 @@ proc getActiveSubscriptions*(
   return activeSubs
 
 proc getShardForContentTopic(
-    self: SubscriptionService, topic: ContentTopic
+    self: SubscriptionManager, topic: ContentTopic
 ): Result[PubsubTopic, string] =
   if self.node.wakuAutoSharding.isSome():
     let shardObj = ?self.node.wakuAutoSharding.get().getShard(topic)
     return ok($shardObj)
 
-  return err("SubscriptionService requires AutoSharding")
+  return err("SubscriptionManager requires AutoSharding")
 
 proc isSubscribed*(
-    self: SubscriptionService, topic: ContentTopic
+    self: SubscriptionManager, topic: ContentTopic
 ): Result[bool, string] =
   let shard = ?self.getShardForContentTopic(topic)
   return ok(
@@ -139,18 +131,15 @@ proc isSubscribed*(
   )
 
 proc isSubscribed*(
-    self: SubscriptionService, shard: PubsubTopic, contentTopic: ContentTopic
+    self: SubscriptionManager, shard: PubsubTopic, contentTopic: ContentTopic
 ): bool {.raises: [].} =
-  try:
-    return
-      self.contentTopicSubs.hasKey(shard) and
-      self.contentTopicSubs[shard].contains(contentTopic)
-  except KeyError:
-    discard
+  self.contentTopicSubs.withValue(shard, cTopics):
+    return cTopics[].contains(contentTopic)
+  return false
 
-proc subscribe*(self: SubscriptionService, topic: ContentTopic): Result[void, string] =
+proc subscribe*(self: SubscriptionManager, topic: ContentTopic): Result[void, string] =
   if isNil(self.node.wakuRelay) and isNil(self.node.wakuFilterClient):
-    return err("SubscriptionService requires either Relay or Filter Client.")
+    return err("SubscriptionManager requires either Relay or Filter Client.")
 
   let shard = ?self.getShardForContentTopic(topic)
 
@@ -162,10 +151,10 @@ proc subscribe*(self: SubscriptionService, topic: ContentTopic): Result[void, st
   return ok()
 
 proc unsubscribe*(
-    self: SubscriptionService, topic: ContentTopic
+    self: SubscriptionManager, topic: ContentTopic
 ): Result[void, string] =
   if isNil(self.node.wakuRelay) and isNil(self.node.wakuFilterClient):
-    return err("SubscriptionService requires either Relay or Filter Client.")
+    return err("SubscriptionManager requires either Relay or Filter Client.")
 
   let shard = ?self.getShardForContentTopic(topic)
 
--- a/tests/api/test_api_subscription.nim
+++ b/tests/api/test_api_subscription.nim
@@ -152,7 +152,7 @@ proc publishToMesh(
   )
   return await net.publisher.publish(some(shard), msg)
 
-suite "Messaging API, SubscriptionService":
+suite "Messaging API, SubscriptionManager":
   asyncTest "Subscription API, relay node auto subscribe and receive message":
     let net = await setupNetwork(1)
     defer:
--- a/waku/node/delivery_service/delivery_service.nim
+++ b/waku/node/delivery_service/delivery_service.nim
@@ -5,7 +5,7 @@ import chronos
 import
   ./recv_service,
   ./send_service,
-  ./subscription_service,
+  ./subscription_manager,
   waku/[
     waku_core,
     waku_node,
@@ -18,31 +18,31 @@ import
 type DeliveryService* = ref object
   sendService*: SendService
   recvService: RecvService
-  subscriptionService*: SubscriptionService
+  subscriptionManager*: SubscriptionManager
 
 proc new*(
     T: type DeliveryService, useP2PReliability: bool, w: WakuNode
 ): Result[T, string] =
   ## storeClient is needed to give store visitility to DeliveryService
   ## wakuRelay and wakuLightpushClient are needed to give a mechanism to SendService to re-publish
-  let subscriptionService = SubscriptionService.new(w)
-  let sendService = ?SendService.new(useP2PReliability, w, subscriptionService)
-  let recvService = RecvService.new(w, subscriptionService)
+  let subscriptionManager = SubscriptionManager.new(w)
+  let sendService = ?SendService.new(useP2PReliability, w, subscriptionManager)
+  let recvService = RecvService.new(w, subscriptionManager)
 
   return ok(
     DeliveryService(
       sendService: sendService,
       recvService: recvService,
-      subscriptionService: subscriptionService,
+      subscriptionManager: subscriptionManager,
     )
   )
 
 proc startDeliveryService*(self: DeliveryService) =
-  self.subscriptionService.startSubscriptionService()
+  self.subscriptionManager.startSubscriptionManager()
   self.recvService.startRecvService()
   self.sendService.startSendService()
 
 proc stopDeliveryService*(self: DeliveryService) {.async.} =
   await self.sendService.stopSendService()
   await self.recvService.stopRecvService()
-  await self.subscriptionService.stopSubscriptionService()
+  await self.subscriptionManager.stopSubscriptionManager()
--- a/waku/node/delivery_service/send_service/send_service.nim
+++ b/waku/node/delivery_service/send_service/send_service.nim
@@ -5,7 +5,7 @@ import std/[sequtils, tables, options]
 import chronos, chronicles, libp2p/utility
 import
   ./[send_processor, relay_processor, lightpush_processor, delivery_task],
-  ../[subscription_service],
+  ../[subscription_manager],
   waku/[
     waku_core,
     node/waku_node,
@@ -58,7 +58,7 @@ type SendService* = ref object of RootObj
 
   node: WakuNode
   checkStoreForMessages: bool
-  subscriptionService: SubscriptionService
+  subscriptionManager: SubscriptionManager
 
 proc setupSendProcessorChain(
     peerManager: PeerManager,
@@ -99,7 +99,7 @@ proc new*(
     T: typedesc[SendService],
     preferP2PReliability: bool,
     w: WakuNode,
-    s: SubscriptionService,
+    s: SubscriptionManager,
 ): Result[T, string] =
   if w.wakuRelay.isNil() and w.wakuLightpushClient.isNil():
     return err(
@@ -120,7 +120,7 @@ proc new*(
     sendProcessor: sendProcessorChain,
     node: w,
     checkStoreForMessages: checkStoreForMessages,
-    subscriptionService: s,
+    subscriptionManager: s,
   )
 
   return ok(sendService)
@@ -260,7 +260,7 @@ proc send*(self: SendService, task: DeliveryTask) {.async.} =
   info "SendService.send: processing delivery task",
     requestId = task.requestId, msgHash = task.msgHash.to0xHex()
 
-  self.subscriptionService.subscribe(task.msg.contentTopic).isOkOr:
+  self.subscriptionManager.subscribe(task.msg.contentTopic).isOkOr:
     error "SendService.send: failed to subscribe to content topic",
       contentTopic = task.msg.contentTopic, error = error
 
--- a/waku/waku_core/subscription/subscription_manager.nim
+++ b/waku/waku_core/subscription/subscription_manager.nim
@@ -5,19 +5,19 @@ import std/tables, results, chronicles, chronos
 import ./push_handler, ../topics, ../message
 
 ## Subscription manager
-type SubscriptionManager* = object
+type LegacySubscriptionManager* = object
   subscriptions: TableRef[(string, ContentTopic), FilterPushHandler]
 
-proc init*(T: type SubscriptionManager): T =
-  SubscriptionManager(
+proc init*(T: type LegacySubscriptionManager): T =
+  LegacySubscriptionManager(
     subscriptions: newTable[(string, ContentTopic), FilterPushHandler]()
   )
 
-proc clear*(m: var SubscriptionManager) =
+proc clear*(m: var LegacySubscriptionManager) =
   m.subscriptions.clear()
 
 proc registerSubscription*(
-    m: SubscriptionManager,
+    m: LegacySubscriptionManager,
     pubsubTopic: PubsubTopic,
     contentTopic: ContentTopic,
     handler: FilterPushHandler,
@@ -29,12 +29,12 @@ proc registerSubscription*(
     error "failed to register filter subscription", error = getCurrentExceptionMsg()
 
 proc removeSubscription*(
-    m: SubscriptionManager, pubsubTopic: PubsubTopic, contentTopic: ContentTopic
+    m: LegacySubscriptionManager, pubsubTopic: PubsubTopic, contentTopic: ContentTopic
 ) =
   m.subscriptions.del((pubsubTopic, contentTopic))
 
 proc notifySubscriptionHandler*(
-    m: SubscriptionManager,
+    m: LegacySubscriptionManager,
     pubsubTopic: PubsubTopic,
     contentTopic: ContentTopic,
     message: WakuMessage,
@@ -48,5 +48,5 @@ proc notifySubscriptionHandler*(
   except CatchableError:
     discard
 
-proc getSubscriptionsCount*(m: SubscriptionManager): int =
+proc getSubscriptionsCount*(m: LegacySubscriptionManager): int =
   m.subscriptions.len()


waku/node/delivery_service/subscription_service.nim
```diff
@@ -1,64 +1,167 @@
...
+    return err("Failed to subscribe: " & error)
+  return ok()
+
+proc doUnsubscribe(
+    self: SubscriptionService, shard: PubsubTopic
+): Result[void, string] =
+  self.node.unsubscribe((kind: PubsubUnsub, topic: shard)).isOkOr:
+    error "Failed to unsubscribe from Relay shard", shard = shard, error = error
+    return err("Failed to unsubscribe: " & error)
+  return ok()
```
NagyZoltanPeter: I'm not convinced we ever wanna do that while Core mode selected.
Our current way to go is to support the network while in Core mode. That means if we break it by unsubscribe from certain shards it would mean the participation is broken too.
Also to consider, Core mode nodes are going to provide lightpush/filter service - as per current design - to other edge mode nodes in the network. Such an unsub would lead we cannot full fill such a service.

cc: @Ivansete-status

fcecin: Yes! And if that is the case, the code driving the `SubscriptionService` can just use `subscribeShard()` and physical pubsub unsubscriptions will never trigger for that shard.

The sub service is generic; it avoids being opinionated on what's the correct way to stabilize the underlying mesh impl or deployment. It allows any API semantics to be expressed, whatever they turn out to be.

If an "edge core" node (a home desktop core node that is e.g. not headless and not a daemon, not supporting a fleet statically as, say, cloud infra) can know ahead of time which shards it should subscribe to (say, "all shards") then it just does so on new/boot (call `subscribeShard()` for e.g. all network shards in autosharding).

If you want I can just add a "Subscribe All Shards" on Waku/WakuNode creation on Core mode to this PR, meaning all Core nodes using this API will always participate on the whole network (we'd leave the ability to support just part of the network to a new "Fleet" mode that we would have to create in a new PR).

NagyZoltanPeter: Yes, IMHO current go to is to auto subscribe to all supported shards [0...numNetworkShards-1], by default and never unsubscribe.
Edge mode case will be in either case different by handling filter subs/unsubs.

I can imagine future feature to let lib user decide on network participation (based on traffic/bandwidth assumptions).

But surely there will be cases because of mix use which only works with lightpush, but config still can say `mode=Core`. This would mean, all messages (or selected messages?) will be sent only via mix+lightpush, but the node can still utilize relay network for receiving messages (which is very reliable per se).

But all for these complex logic we are up to this new API. :-)

fcecin: I can think of a few other ways it would be desirable for an interactive desktop node to behave:

* "Sticky" subs: shard interest grows over time as a side-effect of content topic interest, but never shrinks. Works well in any kind of sharding config.
* "Cooldown" subs: after X time (e.g. 15 minutes) drop a shard for which we haven't had a content topic interest.

Both of these avoid having to auto subscribe to all shards on boot, which might be tricky anyway for e.g. static/manual sharding.

waku/api/api.nim
```diff
@@ -48,6 +48,15 @@ proc send*(
...
   ?checkApiAvailability(w)
 
+  let isSubbed = w.deliveryService.subscriptionService
+    .isSubscribed(envelope.contentTopic)
+    .valueOr(false)
+  if not isSubbed:
+    info "Auto-subscribing to topic on send", contentTopic = envelope.contentTopic
+    let subRes = w.deliveryService.subscriptionService.subscribe(envelope.contentTopic)
+    if subRes.isErr():
+      warn "Failed to auto-subscribe", error = subRes.error
```
Ivansete-status: I think it makes sense to return the error to the API user

waku/node/peer_manager/peer_manager.nim
```diff
@@ -217,54 +225,60 @@ proc loadFromStorage(pm: PeerManager) {.gcsafe.} =
 
   trace "recovered peers from storage", amount = amount
 
-proc selectPeer*(
-    pm: PeerManager, proto: string, shard: Option[PubsubTopic] = none(PubsubTopic)
-): Option[RemotePeerInfo] =
-  # Selects the best peer for a given protocol
+proc selectPeers*(
```
Ivansete-status: I can't quite see the need for that. It seems is never used.
I'd suggest to only have `selectPeer` for now.

waku/node/delivery_service/subscription_service.nim
```diff
@@ -1,64 +1,167 @@
...
     waku_node,
+    waku_relay,
     common/broker/broker_context,
   ]
 
 type SubscriptionService* = ref object of RootObj
-  brokerCtx: BrokerContext
   node: WakuNode
+  shardSubs: HashSet[PubsubTopic]
+  contentTopicSubs: Table[PubsubTopic, HashSet[ContentTopic]]
```
Ivansete-status: Maybe we can simplify a little and keep all the info in one single `Table`.
```suggestion
  subscribedTopics: Table[PubsubTopic, HashSet[ContentTopic]]
```

waku/node/delivery_service/subscription_service.nim
```diff
@@ -1,64 +1,167 @@
...
+  let service = SubscriptionService(
+    node: node,
+    shardSubs: initHashSet[PubsubTopic](),
+    contentTopicSubs: initTable[PubsubTopic, HashSet[ContentTopic]](),
+  )
 
-  return SubscriptionService(brokerCtx: node.brokerCtx, node: node)
+  service.relayHandler = proc(
+      topic: PubsubTopic, msg: WakuMessage
+  ): Future[void] {.async, gcsafe.} =
```
Ivansete-status: Just remove the implicit data that comes from `async` pragma.
```suggestion
  ) {.async.} =
```

waku/node/delivery_service/subscription_service.nim
```diff
@@ -1,64 +1,167 @@
...
+    if not service.contentTopicSubs.hasKey(topic) or
+        not service.contentTopicSubs[topic].contains(msg.contentTopic):
+      return
+
+    let msgHash = computeMessageHash(topic, msg).to0xHex()
+    info "MessageReceivedEvent",
+      pubsubTopic = topic, contentTopic = msg.contentTopic, msgHash = msgHash
+
+    MessageReceivedEvent.emit(service.node.brokerCtx, msgHash, msg)
```
Ivansete-status: This is not needed. In fact, we don't need the `MessageReceiveEvent` at all because we already have this information in the following, considering we are only talking about `CORE` mode:

https://github.com/logos-messaging/logos-delivery/blob/a7872d59d1991c7f726c7845d8ac536c6bdf7fb4/waku/node/kernel_api/relay.nim#L85

waku/node/delivery_service/subscription_service.nim
Ivansete-status: I 100% agree with @NagyZoltanPeter that _Core_ nodes should always participate in the network. This logic that you designed is makes sense only if the node is configured in _Edge_ mode. So, maybe adding an `if mode == Edge` would suffice in that case.

I also agree that, in _Core_ mode, the node should subscribe to all shards of interest by default.

And re the `Fleet` mode, the only addition I see is the following: `Fleet` == `Core`+store-protocol. But let's leave that discussion for the future :)

waku/node/peer_manager/peer_manager.nim
fcecin: Yeah, this was a leftover from me doing Edge logic. Tried to smuggle that refactor in there; will leave it to the actual Edge PR (if it is actually needed).

waku/node/delivery_service/subscription_service.nim
fcecin: We can have only one table here for sure, but with a separate `shardSubs` we can more easily express that relay subs have to stay the same when content topic interest changes (not the only way to do that, of course -- we can just scan the relevant node state directly where we detect contentTopic interest goes from 0 to 1 or 1 to 0 on a shard).

waku/node/delivery_service/subscription_service.nim
fcecin: Whoops, I just realized node.subscribe() is wrong on my part here, you're right. This will *overwrite* the appHandler in the uniqueTopicHandler, right? I just assumed (very wrongly) it was adding to a list (like "Observer" infra elsewhere). EDIT: actually no, if there's another Kernel API user, it will just not do anything.

But anyway, we will need `MessageReceivedEvent` as a counterpart to `MessageSentEvent`, `MessagePropagatedEvent`. The question is how do we make the higher-level API implementor (SubscriptionService, our new "Interest Management" engine that hides everything, modes, protocols, etc.) be an additional _user_ of the Kernel API or infra. I see there's lots of stuff there already, Store, telemetry, topicSubscriptionQueue (ENR updates...), so I see the problem.

The easiest thing would be to make the new higher-level API we are writing the sole Kernel API user that is allowed to exist. That will effectively turn off the low-level API -- nobody can use it because the high-level API implementation is using it -- we are consuming it internally. Or we expand the kernel impl to be consumable by multiple parties.

The future we want I think is one where we can emit `MessageReceivedEvent` and we do not actually care who picks it up -- the code backing the new C bindings, some Nim app, tests, ...

waku/node/delivery_service/subscription_service.nim
fcecin: I added the code that does the subscription to _all_ shards if both Relay and AutoSharding are present. But with this Fleet comment, I think I may have fundamentally misunderstood what it means to "subscribe to all shards in Core/Relay mode" so let's probably check that in the next meeting.

waku/node/delivery_service/subscription_service.nim
NagyZoltanPeter: Hm, I think the problem is not that big here.

First of all, we do not expect someone should use Kernel API along with delivery/messaging API (TheNewAPI -> MAPI).
It would only be possible if from a nim app/lib using our nim library and would mix both worlds.
Or using MAPI and Old rest interface.

I think we can state clearly that's an abuse of the library.

What I would do:
Add another defaul handler for relay topics (at the source @Ivansete-status linked) where I would emit an internal event that could be catched anyplace in the library.
I would listent to that internal "message-received" at the subscription service and filter out by content_topic of interst which one need to be emitted with MessageReceivedEvent (which is the lib interface level event).
This will work in edge - filter case - where we need to listen on filter push receives.
Also it allows us to de-duplicate message receives before emitting to user.

WDYT?

waku/node/delivery_service/subscription_service.nim
NagyZoltanPeter: As thinking this through we can spare a lot of administration by do not care about shard/pubsub topic at all.
In `core` mode case, we must auto subscribe to all possible shard anyway - as @Ivansete-status proposed. So we unlikely need to manage shards vs content-topics.
In 'edge' mode our subscribe will be towards a filter service node, where this content_topic subscription already exists and just works.

shard handling only needed in case we would manually/dynamically manage topic participation in gossip network.

For now I would vote for this simpler case.

waku/node/peer_manager/peer_manager.nim
NagyZoltanPeter: Well, how I see, we will need an additional layer on top of peer_manager for the new API.
This might be worth to put in a separate PR. For now simple use original selectPeer.

The reason is the gathered experience with using service peers on a network where a lot of service nodes are actually desktop applications (applications running a node in `core` mode). Which is exactly what status desktop app does and matching the new API design of `core` mode.
In such cases those service nodes (lightpush and specially the filter services) can be unreliable for various reason in mid/long time frame (running on laptop and closing the lid, machine goes to sleep, wifi disconnect, etc.). So we cannot rely simple on a simple random selection of service node. I think we will need to manage a grey/black list of peers we know gets unreliable in a time-window manner.
So it can surely happen that we need to dynamically switch peers and re-manage our subscriptions on a new filter peer... but maybe we need redundant filter peer use simultanously.

All this is better to place into the api layer that hides the simple selectPeer interface of PeerManager.

fcecin: That's a good idea. The best next thing up from pure random selection would be a biased selection with a simple grief cooldown counter (kept in PeerStore, so RAM-based / dynamic) that starts at 0 for all peers in the Peer Store. We can also persist that to Peer Storage, but I'm not sure we need (or want) that immediately. Peers with the same lowest grief count (e.g. if many have a count of 0) are randomly selected as the "next" peer.

waku/node/delivery_service/subscription_service.nim
fcecin: This branch does that now (it calls `SubscriptionService.subscribeShard` for all autosharding shards on `startWakuNode`).

But we can just use the classic "is node.relay nil?" check inside SubscriptionService plus baked-in knowledge of what relays should be doing when content topic subs change. It's the same. Since we like that more, I'll do that instead.

NagyZoltanPeter: I think we need to check topic health from time to time and subscribe if shard missing, but we need not to manage the associations between shards and content topics.

waku/node/peer_manager/peer_manager.nim
NagyZoltanPeter: Until we don't encounter malfunction from a peer, like failed dial, lost filter subscription, failed send from lightpush peer, they are equally selectable, but when such events come and we need to forcefully switch to new peer, we can have a list kept in a TimedCache.(But possible not banned them forever as we can run out of peers, or the whole network will focus use only few stable fleet nodes that van compromise privacy factor).
But there are other corner cases to watch out, like service peer dos protection limit, which can enforce us to move to new peer or use them dinamically (more for lightpush and store).

waku/node/delivery_service/subscription_service.nim
fcecin: > I think we need to check topic health from time to time and subscribe if shard missing, but we need not to manage the associations between shards and content topics.

Ah, got it! That clears it up.

The architectural philosophy in waku is to prioritize state normalization; now it makes sense. We actively don't want any duplicate state tracking if said state can be inferred from existing lower-level components (e.g. the "protocols"). If we don't need it, we should avoid storing it. That's why the draft SubscriptionService was stateless. I'll see if we can get away with it there.

waku/node/peer_manager/peer_manager.nim
fcecin: I think we can make a solid step-up from blind random select with not much effort and then iterate in the same PR or further PRs.

NagyZoltanPeter: Exactly. 
A little bit more highlight and background of these design ideas. 

https://grafana.infra.status.im/d/fdw6pgh9odszkd/lite-protocol-test-on-status-prod?orgId=1&from=now-12h&to=now&timezone=browser&var-publisher=waku-lite-protocol-tester-publisher-metrics&var-receiver=waku-lite-protocol-tester-receiver-metrics&refresh=5s

This is LiteProtocolTester running periodically on status.prod fleet. 
There are 6 stable fleet node, the rest are status desktop application that offers lightpush and filter services. 
For test, two edge mode app is running, one is lightpushing messages the other one gets them via filter... And measures also also reliability of services nodes. 

So new API is about to implement an optimal use of protocols that ensures close to 100 percent reliability for clients.

waku/node/delivery_service/subscription_service.nim
fcecin: I'll try to delete *all* of the state tracking here. That leaves Edge shard subs (which drives connections/health) driveable only by content topic subs. Core subscribes to "all shards" that we can know exist, forever, which is completely fine for "gen zero" networks at the very least.

waku/node/delivery_service/subscription_service.nim
fcecin: > Ah, got it! That clears it up. (...)

Well, sort of. I found out about RecvService's own shard -> contentTopic tracking. Now that is properly fixed in the latest commit in this PR [fd670f](https://github.com/logos-messaging/logos-delivery/pull/3732/commits/fd6370f3fa7e15a62648c8ea7484397dd0b56a95)).

The idea there is to move what was in the RecvService (sub mapping) to the SubscriptionManager, and have them coordinate to achieve both responsibilities. The RecvService now "receives," and is responsible for emitting the MessageReceivedEvent. To do that, it defers to the SubscriptionManager for this content topic sub intent mapping for filtering by interest.

waku/node/delivery_service/subscription_service.nim
fcecin: Done in the [latest commit in this PR](https://github.com/logos-messaging/logos-delivery/pull/3732/commits/fd6370f3fa7e15a62648c8ea7484397dd0b56a95).  

I'll open the PR tomorrow after some more checking that everything is tip-top.

waku/factory/waku.nim
```diff
@@ -419,6 +420,37 @@ proc startWaku*(waku: ptr Waku): Future[Result[void, string]] {.async: (raises:
...
+      let numShards = autoSharding.shardCountGenZero
+
+      if numShards > 0:
+        var clusterPubsubTopics = newSeqOfCap[PubsubTopic](numShards)
+
+        for i in 0 ..< numShards:
+          let shardObj = RelayShard(clusterId: clusterId, shardId: uint16(i))
+          clusterPubsubTopics.add(PubsubTopic($shardObj))
+
+        subService.subscribeShard(clusterPubsubTopics).isOkOr:
```
Ivansete-status: I'd rename the proc name
```suggestion
        subService.subscribePubsubTopic(clusterPubsubTopics).isOkOr:
```

waku/node/delivery_service/subscription_service.nim
```diff
@@ -1,64 +1,207 @@
...
 proc new*(T: typedesc[SubscriptionService], node: WakuNode): T =
-  ## The storeClient will help to acquire any possible missed messages
+  SubscriptionService(
+    node: node, contentTopicSubs: initTable[PubsubTopic, HashSet[ContentTopic]]()
+  )
 
-  return SubscriptionService(brokerCtx: node.brokerCtx, node: node)
+proc addContentTopicInterest(
+    self: SubscriptionService, shard: PubsubTopic, topic: ContentTopic
+) =
```
Ivansete-status: This should return a `Result`

waku/node/delivery_service/subscription_service.nim
```diff
@@ -1,64 +1,207 @@
...
+      self.contentTopicSubs[shard] = initHashSet[ContentTopic]()
+
+    if not self.contentTopicSubs[shard].contains(topic):
+      self.contentTopicSubs[shard].incl(topic)
+
+      # Always notify EdgeDriver if filter is mounted
+      if not isNil(self.node.wakuFilterClient):
+        self.node.edgeDriver.subscribe(shard, topic)
+  except KeyError:
+    discard
```
Ivansete-status: Let's properly track this possible error :)

waku/node/delivery_service/subscription_manager.nim
```diff
@@ -1,64 +1,207 @@
...
+    waku_relay,
     common/broker/broker_context,
+    events/delivery_events,
+    node/edge_driver,
   ]
 
 type SubscriptionService* = ref object of RootObj
-  brokerCtx: BrokerContext
   node: WakuNode
+  contentTopicSubs: Table[PubsubTopic, HashSet[ContentTopic]]
```
Ivansete-status, line 15: This better belongs to `waku/waku_filter_v2/client.nim`.

waku/node/edge_driver/edge_driver.nim
```diff
@@ -0,0 +1,28 @@
+{.push raises: [].}
+
+import chronicles, waku/waku_core/topics
+
+# Plan:
+# - drive the continuous fulfillment and healing of edge peering and topic subscriptions
+# - offload the edgeXXX stuff from WakuNode into this and finish it
+
+type EdgeDriver* = ref object of RootObj # TODO: bg worker, ...
```
Ivansete-status: I think we don't need that for now.
We already have both lightpush and filter clients. The _Edge_ logic should happen there.

waku/node/delivery_service/recv_service/recv_service.nim
```diff
@@ -133,29 +129,18 @@ proc msgChecker(self: RecvService) {.async.} =
...
-
-    if remainingCTopics.len == 0:
-      self.topicsInterest.del(pubsubTopic)
-  do:
-    error "onUnsubscribe unsubscribing from wrong topic", pubsubTopic, contentTopics
+  let msgHash = computeMessageHash(pubsubTopic, message)
+  if not self.recentReceivedMsgs.anyIt(it.msgHash == msgHash):
+    let rxMsg = RecvMessage(msgHash: msgHash, rxTime: message.timestamp)
+    self.recentReceivedMsgs.add(rxMsg)
+    MessageReceivedEvent.emit(self.brokerCtx, msgHash.to0xHex(), message)
```
Ivansete-status, line 143: Isn't it a bit redundant processing a "message rx" event and then, emit another "message rx" event? We can make that simpler.

In other words, we just need to invoke the following in two places, one in `waku/node/kernel_api/relay.nim` , and the other in `waku/waku_filter_v2/client.nim`:
```nim
MessageReceivedEvent.emit(self.brokerCtx, msgHash.to0xHex(), message)
```

waku/node/delivery_service/subscription_service.nim
```diff
@@ -1,64 +1,207 @@
...
 
 type SubscriptionService* = ref object of RootObj
-  brokerCtx: BrokerContext
   node: WakuNode
+  contentTopicSubs: Table[PubsubTopic, HashSet[ContentTopic]]
+    ## Map of Shard to ContentTopic needed because e.g. WakuRelay is PubsubTopic only.
+    ## A present key with an empty HashSet value means pubsubtopic already subscribed
+    ## (via subscribeShard()) but there's no specific content topic interest yet.
+  filterSubListener: OnFilterSubscribeEventListener
+  filterUnsubListener: OnFilterUnsubscribeEventListener
```
Ivansete-status: These two are not needed because the api is directly calling the `subscribe` and `unsiscribe` procs in this module.

waku/node/delivery_service/subscription_service.nim
```diff
@@ -1,64 +1,207 @@
...
+  self.filterUnsubListener = OnFilterUnsubscribeEvent.listen(
+    self.node.brokerCtx,
+    proc(event: OnFilterUnsubscribeEvent) {.async: (raises: []), gcsafe.} =
+      for cTopic in event.contentTopics:
+        self.removeContentTopicInterest(event.pubsubTopic, cTopic),
+  ).valueOr:
+    return
+      err("SubscriptionService failed to listen to OnFilterUnsubscribeEvent: " & error)
+
+  return ok()
```
Ivansete-status: Not needed, as mentioned above. The content-topics of interest can be manager by filter client.
```suggestion
```

waku/node/delivery_service/subscription_manager.nim
```diff
@@ -1,64 +1,207 @@
...
+): seq[tuple[pubsubTopic: string, contentTopics: seq[ContentTopic]]] =
+  var activeSubs: seq[tuple[pubsubTopic: string, contentTopics: seq[ContentTopic]]] =
+    @[]
+
+  for pubsub, cTopicSet in self.contentTopicSubs.pairs:
+    if cTopicSet.len > 0:
+      var cTopicSeq = newSeqOfCap[ContentTopic](cTopicSet.len)
+      for t in cTopicSet:
+        cTopicSeq.add(t)
+      activeSubs.add((pubsub, cTopicSeq))
```
Ivansete-status, line 112: If mode is Core, then we can get this info from `node.wakuRelay`.
If mode is Edge, we can get that from `node.wakuFilterClient`.

waku/node/delivery_service/subscription_manager.nim
```diff
@@ -1,64 +1,207 @@
...
 
 proc isSubscribed*(
     self: SubscriptionService, topic: ContentTopic
 ): Result[bool, string] =
-  var isSubscribed = false
-  if self.node.wakuRelay.isNil() == false:
-    return self.node.isSubscribed((kind: ContentSub, topic: topic))
+  let shard = ?self.getShardForContentTopic(topic)
+  return ok(
+    self.contentTopicSubs.hasKey(shard) and self.contentTopicSubs[shard].contains(topic)
```
Ivansete-status, line 130: The same as in: https://github.com/logos-messaging/logos-delivery/pull/3732/changes#r2859723566

waku/node/delivery_service/subscription_service.nim
```diff
@@ -1,64 +1,207 @@
...
+      self.node.unsubscribe((kind: PubsubUnsub, topic: shard)).isOkOr:
+        errors.add("shard " & shard & ": " & error)
 
-    MessageReceivedEvent.emit(self.brokerCtx, msgHash, msg)
+      self.contentTopicSubs.del(shard)
+
+  if errors.len > 0:
+    return err("unsubscribeShard errors: " & errors.join("; "))
+
+  return ok()
```
Ivansete-status: This is not used now to let's not add it.
```suggestion
```

waku/factory/waku.nim
```diff
@@ -419,6 +420,37 @@ proc startWaku*(waku: ptr Waku): Future[Result[void, string]] {.async: (raises:
   if not waku[].deliveryService.isNil():
     waku[].deliveryService.startDeliveryService()
 
+  ## Subscription Service
+  if not waku.node.wakuRelay.isNil() and not waku.deliveryService.isNil():
```
Ivansete-status: Let's move this whole new snippet within `subscriptionService`.

waku/node/waku_node.nim
```diff
@@ -114,6 +115,7 @@ type
   WakuNode* = ref object
     peerManager*: PeerManager
     switch*: Switch
+    edgeDriver*: EdgeDriver
```
```suggestion
```

waku/node/kernel_api/relay.nim
```diff
@@ -82,14 +97,25 @@ proc registerRelayHandler(
...
+
   let uniqueTopicHandler = proc(
       topic: PubsubTopic, msg: WakuMessage
   ): Future[void] {.async, gcsafe.} =
     await traceHandler(topic, msg)
     await filterHandler(topic, msg)
     await archiveHandler(topic, msg)
     await syncHandler(topic, msg)
-    await appHandler(topic, msg)
+    await internalHandler(topic, msg)
```
Ivansete-status, line 110: What problems are we trying to solve in this?

waku/node/waku_node.nim
```diff
@@ -146,6 +148,8 @@ type
     started*: bool # Indicates that node has started listening
     topicSubscriptionQueue*: AsyncEventQueue[SubscriptionEvent]
     rateLimitSettings*: ProtocolRateLimitSettings
+    legacyAppHandlers*: Table[PubsubTopic, WakuRelayHandler]
+      ## Kernel API Relay appHandlers (if any)
```
Ivansete-status, line 150: It seems this is not needed either :)
```suggestion
```

waku/node/delivery_service/subscription_manager.nim
fcecin: This has to be here. The `DeliveryService` abstracts away both Core (Relay) and Edge (Filter Client) interest in content topics for message receipt. That's the same data structure that was already in `DeliveryService.RecvService`, I just moved it here since it belongs here.

waku/node/edge_driver/edge_driver.nim
fcecin: If we do that, the Filter Client will be running a worker that is searching for peers (possibly interacting with discovery) and maintaining edge client health. The filter protocol implementation is done; we shouldn't be growing it with logic that uses it IMO.

waku/node/delivery_service/recv_service/recv_service.nim
fcecin: The `MessageReceivedEvent` has to be emitted at the component that does higher-level processing (deduplication, store bookkeeping, content topic interest, ...). Since it is the byproduct of higher-level receipt logic, it should not be put into a lower-level component, even if we had just one such component (or just one of them at a time, which is the current case -- and may be different in the future).

fcecin: The internal messaging event is needed since we need to pull it out of the `waku/node/kernel_api/relay.nim` plumbing, after all the bookeeping and processing that happens there. Either that or the high-level API is going to e.g. hook into WakuRelay and Filter itself, and the Kernel API relay logic is bypassed entirely. That was Zoltan's suggestion in his previous review; I agreed with it and this is what it looks like.

waku/node/delivery_service/subscription_service.nim
fcecin: This is ported from the previous code that was there in the `RecvService`. I'll just delete this entirely for now since it does nothing.

waku/node/delivery_service/subscription_service.nim
fcecin: Yep, I'll just delete it for now.

waku/node/delivery_service/subscription_manager.nim
fcecin: This is called very infrequently for Store queries for missed messages (?) and it feeds on the content topic interest, which is tracked by SubscriptionService now. Also I don't think WakuRelay should mix with the concept of content topics in general?

waku/node/delivery_service/subscription_service.nim
fcecin: At some point, the network grows and the policy won't be "subscribe to all shards that exist." At that point, it becomes possible for a core/relay node to lose interest in shards. We should leave this in even if it is not called now.

waku/node/kernel_api/relay.nim
fcecin: This is part of Zoltan's suggestion to not entirely bypass the `waku/node/kernel_api/relay.nim` and the other handlers. We have to do something like that or alternatively we do a bigger refactor here.

waku/node/waku_node.nim
fcecin: Same as above. We need a way to not completely break the Kernel API, since we can't "conditionally" boot the MAPI.

waku/node/waku_node.nim
fcecin: Same as all other edge infra, this can reappear later.

waku/node/edge_driver/edge_driver.nim
fcecin: Will delete.

waku/node/delivery_service/recv_service/recv_service.nim
fcecin: Renamed to MessageSeenEvent vs MessageReceivedEvent so it's not so confusing.

waku/node/delivery_service/subscription_service.nim
fcecin: Will delete.

waku/node/delivery_service/subscription_manager.nim
fcecin: We decided to take the hit and hold the shard to content topic struct in the SubscriptionService.

waku/node/delivery_service/subscription_service.nim
fcecin: Will delete for now because it is misleading to have dead code.

waku/node/delivery_service/subscription_manager.nim
fcecin: Solved by paying for the shard to content topic mapping at the MAPI level.

waku/node/kernel_api/relay.nim
fcecin: This internalHandler will be removed when all use cases for direct use of the Kernel API are retired, so we agreed to keep this now because it supports current direct uses of the Kernel API together with the automatic mounting of the MAPI services.

waku/node/waku_node.nim
fcecin: Same as above (support Kernel API for now).

waku/node/delivery_service/recv_service/recv_service.nim
```diff
@@ -194,26 +170,24 @@ proc startRecvService*(self: RecvService) =
...
-    proc(subsEv: OnFilterSubscribeEvent) {.async: (raises: []).} =
-      self.onSubscribe(subsEv.pubsubTopic, subsEv.contentTopics),
-  ).valueOr:
-    error "Failed to set OnFilterSubscribeEvent listener", error = error
-    quit(QuitFailure)
+    proc(event: MessageSeenEvent) {.async: (raises: []).} =
+      if not self.subscriptionService.isSubscribed(
+        event.topic, event.message.contentTopic
+      ):
+        return
```
Ivansete-status, line 181: Maybe interesting to add a minimal possible "debug" log, in case we need it.
```suggestion
      ):
      trace "skipping message as I am not subscribed", topic = event.topic, contenttopic = event.message.contentTopic
        return
```

waku/node/delivery_service/subscription_service.nim
```diff
@@ -1,64 +1,175 @@
...
+
+      if cTopics[].len == 0 and isNil(self.node.wakuRelay):
+        self.contentTopicSubs.del(shard) # We're done with cTopics here
+
+      # TODO: Call a "unsubscribe(shard, topic)" on filter client here,
+      #       so the filter client can know that subscriptions changed.
+
+  return ok()
+
+proc subscribePubsubTopics*(
```
Ivansete-status: This can be private
```suggestion
proc subscribePubsubTopics(
```

waku/node/delivery_service/subscription_service.nim
```diff
@@ -1,64 +1,175 @@
...
+
+      self.contentTopicSubs[shard] = initHashSet[ContentTopic]()
+
+  if errors.len > 0:
+    return err("subscribeShard errors: " & errors.join("; "))
+
+  return ok()
+
+proc startSubscriptionService*(self: SubscriptionService) =
+  if not isNil(self.node.wakuRelay):
```
Ivansete-status: Nitpick comment sorry.
If we return earlier, we can keep less indentation levels


```suggestion
  if isNil(self.node.wakuRelay):
     debug "subscription service cannot start because wakuRelay is nil"
     return  
```

waku/node/delivery_service/subscription_service.nim
```diff
@@ -1,64 +1,175 @@
...
+      #       we don't currently have access to Waku.conf here. However, we don't support
+      #       manual/static sharding at the MAPI level anyway so wiring that up now is not needed.
+      #       When we no longer auto-subscribe to all shards in Core boot, we will probably
+      #       scan the shard config due to fleet nodes; then shard conf will have to be reachable here.
+      #       For non-fleet, interactive Core nodes (e.g. Desktop apps) this can't matter
+      #       as much since shard subscriptions originate from subscription to content topics, but
+      #       I guess even in that case subbing to some conf shards may make sense for some apps.
+      info "SubscriptionService has no AutoSharding for Relay, won't subscribe to shards by default."
+
+  discard
```
Ivansete-status: Maybe we can remove that `discard`.
```suggestion
```

waku/node/delivery_service/subscription_service.nim
```diff
@@ -1,64 +1,175 @@
...
+        continue
+
+      self.contentTopicSubs[shard] = initHashSet[ContentTopic]()
+
+  if errors.len > 0:
+    return err("subscribeShard errors: " & errors.join("; "))
+
+  return ok()
+
+proc startSubscriptionService*(self: SubscriptionService) =
```
Ivansete-status: Is a little bit misleading the name because we are not starting an infinite loop. It seems more appropriate something like `subscInAutosharding`, or something like that.

Also, it seems `subscription_service.nim` module should be better renamed to `subscription_manager.nim`

To me a "service" is when we have an infinite loop that is continuously polling a certain state.

waku/node/delivery_service/subscription_service.nim
```diff
@@ -1,64 +1,175 @@
...
-    MessageReceivedEvent.emit(self.brokerCtx, msgHash, msg)
+proc isSubscribed*(
+    self: SubscriptionService, shard: PubsubTopic, contentTopic: ContentTopic
+): bool {.raises: [].} =
+  try:
+    return
+      self.contentTopicSubs.hasKey(shard) and
+      self.contentTopicSubs[shard].contains(contentTopic)
+  except KeyError:
+    discard
```
Ivansete-status:
```suggestion
  except KeyError as e:
    error "could not properly check if is subscribed", shard, contentTopic, error  = e.msg
    return false
```

waku/node/waku_node.nim
Ivansete-status: Double-checking that, it seems we are not breaking the Kernel API because the following is still being used.
https://github.com/logos-messaging/logos-delivery/blob/cdc6d3a2d72797419229cdd9fd15f3bdc8bed488/waku/factory/waku.nim#L78

I would suggest to remove it for now. Another option is to keep it and we can validate now if is needed or not.

Ivansete-status (approved):
LGTM! Thanks for it! 🙌 
Just added minor nitpicks
Cheers

waku/node/delivery_service/recv_service/recv_service.nim
```diff
@@ -166,21 +151,12 @@ proc new*(T: typedesc[RecvService], node: WakuNode, s: SubscriptionService): T =
     startTimeToCheck: now,
     brokerCtx: node.brokerCtx,
     subscriptionService: s,
-    topicsInterest: initTable[PubsubTopic, seq[ContentTopic]](),
     recentReceivedMsgs: @[],
```
NagyZoltanPeter, line 154: How does this seq shrinks in time? 
Due I think we can follow gossip-sub 2min (maybe bigger) message window concept that track messages by hash within X min.
Check out `libp2p/protocols/pubsub/timedcache` for such purpose if you think viable.

waku/node/delivery_service/recv_service/recv_service.nim
```diff
@@ -35,13 +36,8 @@ type RecvMessage = object
...
-    ## Tracks message verification requests and when was the last time a
-    ## pubsub topic was verified for missing messages
-    ## The key contains pubsub-topics
   node: WakuNode
-  onSubscribeListener: OnFilterSubscribeEventListener
-  onUnsubscribeListener: OnFilterUnsubscribeEventListener
+  seenMsgListener: MessageSeenEventListener
   subscriptionService: SubscriptionService
 
   recentReceivedMsgs: seq[RecvMessage]
```
NagyZoltanPeter, line 43: I meant this `recentReceivedMsgs` to be a timely cache. forgive me my prev comment seems went on wrong line.

waku/node/delivery_service/subscription_service.nim
fcecin: For point 1, a better name would be perhaps `onStart` or `onDiscoveryServiceStart`, meaning the SubscriptionService would be reacting to when "it (Waku, WakuNode, ...) has started." My feeling is that `start` and `stop` in Wa... in Logos Discovery everywhere has been deployed to mean "on start" and "on stop" of the parent object, which means please "start" (properly) anything you want to start, but also do anything you need to do or are tasked to do "on start." Every component is entitled to denote a lifecycle for its parent object to respect.

So for example:

```
proc startDeliveryService*(self: DeliveryService) =
  self.subscriptionService.startSubscriptionService()
  self.recvService.startRecvService()
  self.sendService.startSendService()

proc stopDeliveryService*(self: DeliveryService) {.async.} =
  await self.sendService.stopSendService()
  await self.recvService.stopRecvService()
  await self.subscriptionService.stopSubscriptionService()
```

Do all these modules actually have loops? We don't know at the delivery service level here, but we have to treat them like they might: we promise to sync our own warmup and teardown with them.

We could push for a general refactor to change these names if we want start and stop to strictly denote e.g. task spawning. We could have e.g. `begin()` and `end()`, or `init()` and `finish()`, and then inside of these we'd call various internal start and stop named methods to do these things. But for now I'm kind of OK with this "generalization" of start and stop on an object to mean its lifetime management more generally? (in any case I don't think we ever want to depend on things like object ctors and dtors / finalizers and should manage this explicitly).

What I'm not so sure about is whether we should be naming these things e.g. "startDeliveryService" instead of just "start." Maybe just plain "start" and "stop" would be better. And we would standardize those to mean general lifetime management, irrespective of what the "start" or the "stop" of an object or service entails.

Also we don't want the parent object (DeliveryService) to call a lifetime method that is "do this specific thing" (e.g. subscribe all shards) because that's an internal detail of the subscription manager's lifetime that we don't want to leak. Every time we need to do something else that's also specific, it would land on a new lifetime management proc. We need to call something generically named like start/stop or begin/end, which we should standardize for all our objects / components actually.

For point 2, yep SubscriptionManager is much better, let's rename.

waku/node/delivery_service/recv_service/recv_service.nim
fcecin: Currently we have the `loopPruneOldMessages` task that wakes up every `PruneOldMsgsPeriod` (1 minute) and gets rid of anything older than `MaxMessageLife` (7 minutes) by checking every message with a seq keepItIf (linear ...). Yep, we can be faster. I'll try to speed this up in a follow-up PR.

waku/node/delivery_service/subscription_service.nim
fcecin: Ah turns out that we already have a SubscriptionManager (`waku/core/subscription/subscription_manager.nim`) which is causing import conflicts. I'm trying to figure out what that thing does; maybe we can delete it.

waku/node/delivery_service/recv_service/recv_service.nim
NagyZoltanPeter: Seems I overlooked it, than it's all good for now, no need for premature optimization 😉

waku/node/delivery_service/subscription_service.nim
fcecin: I renamed the old one to `LegacySubscriptionManager` to avoid confusion with the MAPI SubscriptionManager (which is the only one that should be exposed, really).

PR #3734, opened 2026-02-20
bump nim-metrics to proper tagged commit v0.2.1

Ivansete-status:
This is just the same as the following PR but bumping into a proper git tag nim-metrics v0.2.1

- https://github.com/logos-messaging/logos-delivery/pull/3730


Ivansete-status:
image

PR #3742, opened 2026-03-02
Add debug API in liblogosdelivery

Ivansete-status:
## Description

This introduces changes to implement the specs that have been added in [this](https://github.com/logos-messaging/specs/pull/102).

## Changes

- Add new functions in liblogosdelivery:
  - `int logosdelivery_get_available_node_info_ids`
  - `int logosdelivery_get_node_info`
  - `int logosdelivery_get_available_configs`
- Update `liblogosdelivery/examples/logosdelivery_example.c` to use the new functions.

### Related Issue

- https://github.com/logos-messaging/pm/issues/305


liblogosdelivery/logos_delivery_api/debug_api.nim
```diff
@@ -0,0 +1,49 @@
...
+    except ValueError:
+      return err("Invalid node info id: " & $nodeInfoId)
+
+  return ok(ctx.myLib[].stateInfo.getNodeInfoItem(infoItemIdEnum))
+
+proc logosdelivery_get_available_configs(
+    ctx: ptr FFIContext[Waku],
+    callback: FFICallBack,
+    userData: pointer,
+    nodeInfoId: cstring,
```
fcecin: Is this 4th argument right? Looks like `logosdelivery_get_available_configs` is declared with 3 args?

```
int logosdelivery_get_available_configs(void *ctx,
                                    FFICallBack callback,
                                    void *userData);
```

Ivansete-status: Wow great catch! Thanks ! :raised_hands: 
Addressed in https://github.com/logos-messaging/logos-delivery/pull/3742/changes/a0b6029544e2213cad4516fd6f39730c3ee2a9ce

PR #3743, opened 2026-03-03
Fix NodeHealthMonitor logspam

fcecin:
## Description

Improve NodeHealthMonitor logging.

## Changes

* Move logging out of calculateConnectionState inner proto loop
* Add debug logging when connection state actually changes

## Issue

[Maintenance Y2026H1](https://github.com/logos-messaging/logos-delivery/issues/3686)

### DIFF
--- a/waku/node/health_monitor/node_health_monitor.nim
+++ b/waku/node/health_monitor/node_health_monitor.nim
@@ -405,13 +405,8 @@ proc calculateConnectionState*(
     elif kind in FilterClientProtocols:
       filterCount = max(filterCount, strength)
 
-    debug "calculateConnectionState",
-      protocol = kind,
-      strength = strength,
-      relayCount = relayCount,
-      storeClientCount = storeClientCount,
-      lightpushCount = lightpushCount,
-      filterCount = filterCount
+  debug "calculateConnectionState",
+    relayCount, storeClientCount, lightpushCount, filterCount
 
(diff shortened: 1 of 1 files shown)


waku/node/health_monitor/node_health_monitor.nim
```diff
@@ -405,13 +405,8 @@ proc calculateConnectionState*(
...
       filterCount = max(filterCount, strength)
 
-    debug "calculateConnectionState",
-      protocol = kind,
-      strength = strength,
-      relayCount = relayCount,
-      storeClientCount = storeClientCount,
-      lightpushCount = lightpushCount,
-      filterCount = filterCount
+  debug "calculateConnectionState",
```
NagyZoltanPeter, line 408: I'm not sure how important is to get this at every loop? Maybe trace is more appropriate. WDYT?
 
```suggestion
  trace "calculateConnectionState",
```

fcecin: Yeah, we need to talk about logging levels at some point. E.g. the CI will only print up to `debug` (?), so `trace` won't help there, and so we ended up being biased to `debug`?

fcecin: I mean, there's probably a lot of logging that needs levels reviewed, and this one would be included in that.

PR #3744, opened 2026-03-03
Chore: adapt cli args for delivery api

NagyZoltanPeter:
## Description

We found that NodeConfig (api_conf) is incomplete for all of our purposes and dogfooding shown its structure is a bit complex for easy and straightforward use.

Decision is made that new logos-delivery API will return to utilize older cli arguments (WaluNodeConf) with some handful shortcuts.

Benefit is that we are able to use existing configs/setups as for libwaku or wakunode2.

> This is a breaking change compared what we have in the initial liblogosdelivery and logos-devivery-module!

## Changes

- Added support for `mode` switch which has `noMode`/`Core`/`Edge`.
   - If using `Core` or `Edge` right set of protocols and some defaults are set automatically.
- Added `preset` for logos.dev fleet, for easy connect.
- creatNode API now accepts the WakuNodeConf format as json.
- Tests are adjusted
- agentString default is changed from `nwaku` to `logos-delivery`

## Issue

- https://github.com/logos-messaging/pm/issues/305
- https://github.com/logos-messaging/logos-delivery/issues/3740

### DIFF
--- a/liblogosdelivery/README.md
+++ b/liblogosdelivery/README.md
@@ -32,17 +32,16 @@ void *logosdelivery_create_node(
 ```json
 {
   "mode": "Core",
-  "clusterId": 1,
-  "entryNodes": [
-    "enrtree://AIRVQ5DDA4FFWLRBCHJWUWOO6X6S4ZTZ5B667LQ6AJU6PEYDLRD5O@sandbox.waku.nodes.status.im"
-  ],
-  "networkingConfig": {
-    "listenIpv4": "0.0.0.0",
-    "p2pTcpPort": 60000,
-    "discv5UdpPort": 9000
-  }
+  "preset": "logos.dev",
+  "listenAddress": "0.0.0.0",
+  "tcpPort": 60000,
+  "discv5UdpPort": 9000
 }
 ```
+
+Configuration uses flat field names matching `WakuNodeConf` in `tools/confutils/cli_args.nim`.
+Use `"preset"` to select a network preset (e.g., `"twn"`, `"logos.dev"`) which auto-configures
+entry nodes, cluster ID, sharding, and other network-specific settings.
 
 #### `logosdelivery_start_node`
 Starts the node.
@@ -207,8 +206,9 @@ void callback(int ret, const char *msg, size_t len, void *userData) {
 
 int main() {
     const char *config = "{"
+        "\"logLevel\": \"INFO\","
         "\"mode\": \"Core\","
-        "\"clusterId\": 1"
+        "\"preset\": \"logos.dev\""
         "}";
 
     // Create node
--- a/tests/api/test_node_conf.nim
+++ b/tests/api/test_node_conf.nim
@@ -308,7 +308,7 @@ suite "WakuNodeConf JSON -> WakuConf integration":
   test "JSON with max message size":
     ## Given
     let confRes =
-      parseWakuNodeConfFromJson("""{"clusterId": 1, "maxMessageSize": "100KiB"}""")
+      parseWakuNodeConfFromJson("""{"clusterId": 42, "maxMessageSize": "100KiB"}""")
     require confRes.isOk()
     let conf = confRes.get()
 


tools/confutils/cli_args.nim
```diff
@@ -52,6 +52,11 @@ type StartUpCommand* = enum
   noCommand # default, runs waku
   generateRlnKeystore # generates a new RLN keystore
 
+type WakuMode* {.pure.} = enum
+  noMode # default - use explicit CLI flags as-is
```
Ivansete-status, line 57: Great idea

tools/confutils/cli_args.nim
```diff
@@ -907,12 +919,19 @@ proc toNetworkConf(
...
+      "Logos.dev - Logos.dev configuration will not be applied when `--cluster-id=2` is passed in future releases. Use `--preset=logos.dev` instead."
+    )
+    lcPreset = "logos.dev"
 
   case lcPreset
   of "":
     ok(none(NetworkConf))
   of "twn":
     ok(some(NetworkConf.TheWakuNetworkConf()))
+  of "logos.dev", "logosdev":
```
Ivansete-status, line 942: Shall we do `case lcPreset.toLowrAscii()` to make sure we don't face case issues?

tools/confutils/cli_args.nim
```diff
@@ -1074,4 +1093,26 @@ proc toWakuConf*(n: WakuNodeConf): ConfResult[WakuConf] =
   b.kademliaDiscoveryConf.withEnabled(n.enableKadDiscovery)
   b.kademliaDiscoveryConf.withBootstrapNodes(n.kadBootstrapNodes)
 
+  # Mode-driven configuration overrides
+  case n.mode
+  of WakuMode.Core:
```
Ivansete-status, line 1128: Should we explicitly enable store in Core for now and disable in Edge?

waku/api/api.nim
```diff
@@ -22,6 +24,19 @@ proc createNode*(config: NodeConfig): Future[Result[Waku, string]] {.async.} =
 
   return ok(wakuRes)
 
+# Deprecated: use WakuNodeConf overload instead
+proc createNode*(
```
Ivansete-status: If not used now, maybe we can delete?
If needed in the future, we can reimplement it back

waku/factory/conf_builder/waku_conf_builder.nim
```diff
@@ -393,6 +393,18 @@ proc applyNetworkConf(builder: var WakuConfBuilder) =
...
+  if networkConf.enableKadDiscovery:
+    if not builder.kademliaDiscoveryConf.enabled:
+      builder.kademliaDiscoveryConf.withEnabled(networkConf.enableKadDiscovery)
+
+    if builder.kademliaDiscoveryConf.bootstrapNodes.len == 0 and
+        networkConf.kadBootstrapNodes.len > 0:
+      builder.kademliaDiscoveryConf.withBootstrapNodes(networkConf.kadBootstrapNodes)
+
+  if networkConf.mix:
+    if builder.mix.isNone:
```
Ivansete-status, line 406: nitpick sorry
```suggestion
    if builder.mix.isNone():
```

waku/rest_api/endpoint/builder.nim
```diff
@@ -42,6 +41,9 @@ type RestServerConf* = object
 proc startRestServerEssentials*(
     nodeHealthMonitor: NodeHealthMonitor, conf: RestServerConf, portsShift: uint16
 ): Result[WakuRestServerRef, string] =
+  if restServerNotInstalledTab.isNil:
```
Ivansete-status, line 44:
```suggestion
  if restServerNotInstalledTab.isNil():
```

waku/rest_api/endpoint/builder.nim
```diff
@@ -42,6 +41,9 @@ type RestServerConf* = object
...
+    restServerNotInstalledTab = newTable[string, string]()
```
Ivansete-status, line 45: Another nitpick, I've seen somewhere that maybe is not needed to explicitly do that

tools/confutils/cli_args.nim
NagyZoltanPeter: It is already lowercase at this point.

tools/confutils/cli_args.nim
NagyZoltanPeter: Store service, I don't think so. That must be also driven by DBurl as prerequisite.

PR #3748, opened 2026-03-05
setting num-shards-in-network to 0 by default

Ivansete-status:
## Description

Setting num shards in network to zero by default, which implies that static sharding is considered by default.

This is needed to avoid having to set weird configuration settings in a node that is currently configuring static shards.

For example, this current config:
```nim
cluster-id = 16
shard = [
  1,
  32,
  64,
  128,
  256,
]
```
Doesn't allow a node to run properly with v0.37.0-beta, showing the following error:
```nim
nim-waku-store  | ERR 2026-03-05 15:17:13.794+00:00 Waku configuration failed                  topics="wakunode main" tid=1 file=wakunode2.nim:52 error="validateShards invalid shard: 1 when numShardsInCluster: 1"
```

Then, with the current implementation, we need to provide this weird configuration to make the node start, which doesn't look nice:
```nim
cluster-id = 16
num-shards-in-network = 257
shard = [
  1,
  32,
  64,
  128,
  256,
]
```

### DIFF
--- a/tests/wakunode2/test_cli_args.nim
+++ b/tests/wakunode2/test_cli_args.nim
@@ -23,9 +23,8 @@ import
 suite "Waku external config - default values":
   test "Default sharding value":
     ## Setup
-    let defaultShardingMode = AutoSharding
-    let defaultNumShardsInCluster = 1.uint16
-    let defaultSubscribeShards = @[0.uint16]
+    let defaultShardingMode = StaticSharding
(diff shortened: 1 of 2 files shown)


fcecin (approved):
LGTM!

NagyZoltanPeter (approved):
It's a good catch! Thanks!

PR #3754, opened 2026-03-12
chore: correct dynamic library extension on mac and update OS detection

darshankabariya:


Makefile
```diff
@@ -28,9 +28,10 @@ GIT_SUBMODULE_UPDATE := git submodule update --init --recursive
 else # "variables.mk" was included. Business as usual until the end of this file.
 
 # Determine the OS
-detected_OS := $(shell uname -s)
-ifneq (,$(findstring MINGW,$(detected_OS)))
+ifeq ($(OS),Windows_NT)
```
Ivansete-status: I'm not totally convinced this change is needed

darshankabariya: Thanks! Yes, it’s not required. I actually pushed that code while working on the Windows build. I’m removing it now haha 😄

PR #3772, opened 2026-03-20
feat: service discovery integration

SionoiS:
Service discovery integration for testing purposes. Will not be merged.


tools/confutils/cli_args.nim
```diff
@@ -637,12 +638,6 @@ with the drawback of consuming some more bandwidth.""",
       name: "mixkey"
     .}: Option[string]
 
-    mixnodes* {.
```
chaitanyaprem, line 639: I think we might still want to have this for qucker bootstrapping when the node comes up. otherwise until 5 mixnodes are discovered the mix protocol would become unsuable.

SionoiS: This is just a hacky test for service discovery, so I removed it. I agree that the final code should have a way to bootstrap faster.

SionoiS:
Can't make it compile anymore with latest libp2p. Closing in favor of something else that works...

PR #3798, opened 2026-04-03
Start using nimble and deprecate vendor dependencies

Ivansete-status:
### Reason why this PR
This continues https://github.com/logos-messaging/logos-delivery/pull/3682 because it became too convoluted.

### Detail
- Start using nimble
- Stop using nimbus-build-system, a.k.a NBS, and remove dependencies from vendor folder
- nim version: 2.2.4 (version used so far from NBS)
- nimble version: 0.18.2 (the one that gets automatically installed with `choosenim 2.2.4`)
- build liblogosdelivery on nix
- zerokit, and smart-contrat repos still live in vendor/ folder

### From now on
The source of truth for versions is `waku.nimble`.

`waku.nimble` adds rigid constraints to certain dependencies, such as `libp2p` or `nimcrypto` and leave more freedom to others. Then, the `nimble lock` command generated a `nimble.lock` file with known `libp2p` and `nimcrypto`, f.e., and get the latest version from other dependencies that don't have such fixed constraint.

In turn, `nimble.lock` is the source of truth for `nix/deps.nix`, which is generated with `./tools/generate-nix-deps.sh nimble.lock nix/deps.nix` ( this technique is copied from `nim-libp2p` so kudos to you @richard-ramos and your team 🙌 ! Thanks for it!, this PR doesn't suggest the use of `.pinned` though )

The `nimble.lock` is git-versioned and, when bumping dependencies, we have to make sure it doesn't refer to any nim version there, and only `logos-delivery` dependencies.

### Issue
- https://github.com/logos-messaging/logos-delivery/issues/3672


fcecin (approved):
Epic PR!
LGTM!

.github/workflows/ci.yml
```diff
@@ -63,24 +65,36 @@ jobs:
...
+            nimbledeps/
+            nimble.paths
+          key: ${{ runner.os }}-nimbledeps-${{ hashFiles('nimble.lock') }}
 
-      - name: Make update
-        run: make update
+      - name: Install nimble deps
+        if: steps.cache-nimbledeps.outputs.cache-hit != 'true'
+        run: |
+          nimble setup --localdeps -y
```
NagyZoltanPeter, line 91: Why do we need localdeps in ci?
Wonder if there is a way to use some kind of nimbla pkg cache in CI... that might speed up builds.

.github/workflows/ci-nix.yml
```diff
@@ -16,14 +16,7 @@ jobs:
...
         nixpkg:
-          - libwaku
-          - libwaku-android-arm64
-          - wakucanary
-
-        exclude:
-          # Android SDK limitation
-          - system: aarch64-darwin
-            nixpkg: libwaku-android-arm64
+          - liblogosdelivery
```
NagyZoltanPeter, line 19: Why?
... I see, I mixed up the changes and thought we exlude liblogosdelivey.
But still why we don't build libwaku on ci with nix?

nix/default.nix
```diff
@@ -1,149 +1,107 @@
...
+      --define:disable_libbacktrace \
+      --out:build/liblogosdelivery.a \
+      --app:staticlib \
+      --threads:on \
+      --opt:size \
+      --noMain \
+      --mm:refc \
+      --nimMainPrefix:liblogosdelivery \
+      --nimcache:$NIMCACHE \
+      liblogosdelivery/liblogosdelivery.nim
```
NagyZoltanPeter, line 92: Is it a good idea to maintain separately the proper parameterization of nim compiler for liblogosdelivery instead of rely on make or nimble?
Is it because of nimble path is unusable in nix env?

Makefile
```diff
@@ -4,57 +4,44 @@
...
 
 ##########
 ## Main ##
 ##########
-.PHONY: all test update clean examples
+.PHONY: all test update clean examples deps nimble
 
-# default target, because it's the first one that doesn't start with '.'
+# default target
 all: | wakunode2 libwaku
```
NagyZoltanPeter:
```suggestion
all: | wakunode2 libwaku liblogosdelivery
```

I think it's time to make liblogosdelivery a first class citizen of the repo :-)

waku.nimble
```diff
@@ -4,81 +4,161 @@ import os
...
+
+proc buildLibDynamicWindows(libName: string, folderName: string) =
+  buildLibrary libName & ".dll", folderName,
+    """-d:chronicles_line_numbers --warning:Deprecated:off --warning:UnusedImport:on -d:chronicles_log_level=TRACE """,
+    "dynamic"
+
+proc buildLibDynamicLinux(libName: string, folderName: string) =
+  buildLibrary libName & ".so", folderName,
+    """-d:chronicles_line_numbers --warning:Deprecated:off --warning:UnusedImport:on -d:chronicles_log_level=TRACE """,
+    "dynamic"
```
NagyZoltanPeter:
```suggestion
    "dynamic", libName & ".nim", libname
```
missing this result build failed for liblogosdelivery artifacts!

waku.nimble
```diff
@@ -4,81 +4,161 @@ import os
...
+  let sdkPath = staticExec("xcrun --show-sdk-path").strip()
+  when defined(arm64):
+    let archFlags = "--cpu:arm64 --passC:\"-arch arm64\" --passL:\"-arch arm64\" --passC:\"-isysroot " & sdkPath & "\" --passL:\"-isysroot " & sdkPath & "\""
+  elif defined(amd64):
+    let archFlags = "--cpu:amd64 --passC:\"-arch x86_64\" --passL:\"-arch x86_64\" --passC:\"-isysroot " & sdkPath & "\" --passL:\"-isysroot " & sdkPath & "\""
+  else:
+    {.error: "Unsupported macOS architecture".}
+  buildLibrary libName & ".dylib", folderName,
+    archFlags & " -d:chronicles_line_numbers --warning:Deprecated:off --warning:UnusedImport:on -d:chronicles_log_level=TRACE",
+    "dynamic"
```
NagyZoltanPeter:
```suggestion
    "dynamic", libName & ".nim", libname
```

waku.nimble
```diff
@@ -4,81 +4,161 @@ import os
...
   else:
     exec "nim c" & " --out:build/" & lib_name &
       " --threads:on --app:lib --opt:speed --noMain --mm:refc --header -d:metrics --nimMainPrefix:" & mainPrefix & " --skipParentCfg:off -d:discv5_protocol_id=d5waku " &
-      extra_params & " " & srcDir & srcFile
+      getMyCPU() & getNimParams() & " " & srcDir & "/" & srcFile
+
+proc buildLibDynamicWindows(libName: string, folderName: string) =
+  buildLibrary libName & ".dll", folderName,
+    """-d:chronicles_line_numbers --warning:Deprecated:off --warning:UnusedImport:on -d:chronicles_log_level=TRACE """,
+    "dynamic"
```
NagyZoltanPeter:
```suggestion
    "dynamic", libName & ".nim", libname
```

waku.nimble
```diff
@@ -4,81 +4,161 @@ import os
...
+  else:
+    {.error: "Unsupported macOS architecture".}
+  buildLibrary libName & ".dylib", folderName,
+    archFlags & " -d:chronicles_line_numbers --warning:Deprecated:off --warning:UnusedImport:on -d:chronicles_log_level=TRACE",
+    "dynamic"
+
+proc buildLibStaticWindows(libName: string, folderName: string) =
+  buildLibrary libName & ".lib", folderName,
+    """-d:chronicles_line_numbers --warning:Deprecated:off --warning:UnusedImport:on -d:chronicles_log_level=TRACE """,
+    "static"
```
NagyZoltanPeter:
```suggestion
    "static", libName & ".nim", libname
```

waku.nimble
```diff
@@ -4,81 +4,161 @@ import os
...
+
+proc buildLibStaticLinux(libName: string, folderName: string) =
+  buildLibrary libName & ".a", folderName,
+    """-d:chronicles_line_numbers --warning:Deprecated:off --warning:UnusedImport:on -d:chronicles_log_level=TRACE """,
+    "static"
```
NagyZoltanPeter:
```suggestion
    "static", libName & ".nim", libname
```

waku.nimble
```diff
@@ -4,81 +4,161 @@ import os
...
+  let sdkPath = staticExec("xcrun --show-sdk-path").strip()
+  when defined(arm64):
+    let archFlags = "--cpu:arm64 --passC:\"-arch arm64\" --passL:\"-arch arm64\" --passC:\"-isysroot " & sdkPath & "\" --passL:\"-isysroot " & sdkPath & "\""
+  elif defined(amd64):
+    let archFlags = "--cpu:amd64 --passC:\"-arch x86_64\" --passL:\"-arch x86_64\" --passC:\"-isysroot " & sdkPath & "\" --passL:\"-isysroot " & sdkPath & "\""
+  else:
+    {.error: "Unsupported macOS architecture".}
+  buildLibrary libName & ".a", folderName,
+    archFlags & " -d:chronicles_line_numbers --warning:Deprecated:off --warning:UnusedImport:on -d:chronicles_log_level=TRACE",
+    "static"
```
NagyZoltanPeter:
```suggestion
    "static", libName & ".nim", libname
```

waku.nimble
```diff
@@ -88,141 +168,17 @@ proc buildMobileAndroid(srcDir = ".", params = "") =
...
-    outdir & " --passL:-lrln --passL:-llog --cpu:" & cpu & " --os:android -d:androidNDK " &
-    extra_params & " " & srcDir & "/libwaku.nim"
-
-proc test(name: string, params = "-d:chronicles_log_level=DEBUG", lang = "c") =
-  # XXX: When running `> NIM_PARAMS="-d:chronicles_log_level=INFO" make test2`
-  # I expect compiler flag to be overridden, however it stays with whatever is
-  # specified here.
-  buildBinary name, "tests/", params
-  exec "build/" & name
+    "/liblogosdelivery.so --threads:on --app:lib --opt:speed --noMain --mm:refc -d:chronicles_sinks=textlines[dynamic] --header -d:chronosEventEngine=epoll --passL:-L" &
```
NagyZoltanPeter, line 172: Aren't we miss here `--nimMainPrefix:`?

waku.nimble
```diff
@@ -4,81 +4,161 @@ import os
...
+    let archFlags = "--cpu:amd64 --passC:\"-arch x86_64\" --passL:\"-arch x86_64\" --passC:\"-isysroot " & sdkPath & "\" --passL:\"-isysroot " & sdkPath & "\""
+  else:
+    {.error: "Unsupported macOS architecture".}
+  buildLibrary libName & ".a", folderName,
+    archFlags & " -d:chronicles_line_numbers --warning:Deprecated:off --warning:UnusedImport:on -d:chronicles_log_level=TRACE",
+    "static"
+
+### Mobile Android
 
 proc buildMobileAndroid(srcDir = ".", params = "") =
```
NagyZoltanPeter, line 163: It's a bit strange that `buildMobileAndroid` presents a liblogosdelivery artifact for Android while
`buildMobileIOS` still produce a libwaku!

NagyZoltanPeter (approved):
I can't wait to get it and start using it.
Added some questions and some needed code changes too.
> Note: without those fix, liblogosdelivery cannot be built.

flake.nix
```diff
@@ -1,92 +1,71 @@
...
-          overlays =  [
-            (final: prev: {
-              androidEnvCustom = prev.callPackage ./nix/pkgs/android-sdk { };
-              androidPkgs = final.androidEnvCustom.pkgs;
-              androidShell = final.androidEnvCustom.shell;
-            })
-          ];
+        in rec {
+          liblogosdelivery = mkPkg zerokit.packages.${system}.rln;
+          libwaku           = mkPkg zerokit.packages.${system}.rln;
```
NagyZoltanPeter: This actually will build liblogosdelivery artifact! Is this the goal here?
It is due that default.nix which is imported with mkPkg can build only liblogosdelivery but not libwaku.

NagyZoltanPeter (commented):
Added more questions.

nix/nimble.nix
```diff
@@ -1,12 +1,17 @@
...
-  hash = "sha256-8iutVgNzDtttZ7V+7S11KfLEuwhKA9TsgS51mlUI08k=";
+  lines       = pkgs.lib.splitString "\n" (builtins.readFile ../waku.nimble);
+  versionLine = builtins.head (builtins.filter
+    (l: builtins.match "^const NimbleVersion.*" l != null) lines);
+  version     = builtins.head (builtins.match ".*\"([0-9]+\\.[0-9]+\\.[0-9]+)\".*" versionLine);
+in
+pkgs.fetchgit {
+  url = "https://github.com/nim-lang/nimble";
+  rev = "v${version}";
+  sha256 = "18cwsjwcgjmnm42kr310hfbw06lym3vaj641i4pla6r8w22xqpqd";
```
NagyZoltanPeter:
```suggestion
  sha256 = "1jv4am285vw0isf2s666hisbprljsadifz6by80lq5hxra2ca362";
```

nix build in my local failed, just the same as ci-nix for this PR.
My research shown that the sha256 hash is stale. 
Imo this solution can be a problem later as we cannot control this just by using a fixed nimble version.
Still the tag maybe moved in the upstream repo of nimble or its submodules are changing will always result faild nix dependency for us which will cause headache in CI.

NagyZoltanPeter (changes requested):
nimble locking is stale...

.github/workflows/ci.yml
Ivansete-status: Yes, the idea is to use a cache to speed up builds :)

.github/workflows/ci-nix.yml
Ivansete-status: We only need liblogosdelivery in nix, afaik

nix/default.nix
Ivansete-status: Yes good point!
However, I'm afraid we need to do it in this way considering how zerokit and the dependencies are fetched, and considering the nix sandboxed approach. The Makefile and waku.nimble relies on `nimble setup --localdeps`, which generates the `nimbledeps` dependency folder, which is not available in nix.
Summarizing, we need the dependencies to be fetched at buildTime, in nix, throughout a _fixed-output derivation_ set ( `nix/deps.nix` ) and that doesn't allow us using Makefile or nimble directly, snif snif

flake.nix
Ivansete-status: Good point, thanks! This is addressed in https://github.com/logos-messaging/logos-delivery/pull/3798/changes/bc6fba84725a1c885b024674874f4499f2b629f9

waku.nimble
Ivansete-status: Great catch thanks! Addressed in https://github.com/logos-messaging/logos-delivery/pull/3798/changes/6aa53b64a3fbe6e69557756c8c721b5167c17a76

waku.nimble
Ivansete-status: Yes good point, I think we should only build liblogosdelivery in mobile. Addressed in https://github.com/logos-messaging/logos-delivery/pull/3798/changes/eba06eea9774659730fb1ee5d53d5b027863bd15

NagyZoltanPeter (approved):
Lgtm now! Great work!

waku.nimble
```diff
@@ -4,81 +4,161 @@ import os
...
+version = "0.37.4"
 author = "Status Research & Development GmbH"
 description = "Waku, Private P2P Messaging for Resource-Restricted Devices"
 license = "MIT or Apache License 2.0"
 #bin           = @["build/waku"]
 
+## This indicates the nim compiler version we are currently working on. It may compile with others
+## but we haven't tested.
+const NimVersion    = "2.2.4"
+## This is the underlying nimble version that gets installed after doing `choosenim 2.2.4`.
```
arnetheduck, line 16: in general, choosenim is deprecated and we _should_ be using a fixed nimble version that is recent, similar to how n-b-s installs a particular version of nimble. 

We no longer want to rely on the nimble version shipped with Nim in any shape or form and should instead always use a specific nimble version - this is important for reproducibility.

arnetheduck:
>  The nimble.lock is git-versioned and, when bumping dependencies, we have to make sure it doesn't refer to any nim version there, and only logos-delivery dependencies.

This is a significant regression and significant problem, compared to n-b-s - the exact nim version *must* be part of nimble.lock, else it no longer serves as a source of truth.

It is also a regression to introduce a hard dependency on nix given that it is not supported on windows and undesireable on android and similar platforms - ie while it's fine _as an option_ to provide a dev environment (gcc etc), it's also limiting compared to the status quo.

arnetheduck:
To be clear, what a reproducible flow looks like:

* something provides a dev environment - this can be docker / nix / android sdk / whatever
* a locked-down version of nimble is installed
* nimble reads nimble.lock and installs nim and all other dependencies using the right versions based on the lock file

flake.nix
```diff
@@ -1,92 +1,70 @@
...
-          targets = ["libwaku"];
-          zerokitRln = zerokit.packages.${system}.rln;
-        };
-
-        wakucanary = pkgs.callPackage ./nix/default.nix {
-          inherit stableSystems;
-          src = self;
-          targets = ["wakucanary"];
-          zerokitRln = zerokit.packages.${system}.rln;
-        };
```
jakubgs, line 76: Can you explain to me why you removed the `wakucanary` target?

Ivansete-status: Oh my bad, it was by accident sorry. We'll submit a separate PR to get it back.

EDITED: submitted this PR https://github.com/logos-messaging/logos-delivery/pull/3892

PR #3801, opened 2026-04-07
chore: Add pre-check of options used in config Json for liblogosdelivery pre-createNode, treat unrecognized options as error

NagyZoltanPeter:
## Description

liblogosdelivey silently ignores uncrecognized options given through json string config.

## Changes

- For liblogosdelivery pre calling createNode we parse the given json string configuration. Now we cross check json keys with known options and treat unrecognized options as error.

## Issue

https://github.com/logos-messaging/logos-delivery/issues/3800
closes #
https://github.com/logos-messaging/logos-delivery/issues/3800

### DIFF
--- a/liblogosdelivery/logos_delivery_api/node_api.nim
+++ b/liblogosdelivery/logos_delivery_api/node_api.nim
@@ -33,10 +33,13 @@ registerReqFFI(CreateNodeRequest, ctx: ptr FFIContext[Waku]):
     for confField, _ in fieldPairs(conf):
       knownFields.incl(confField)
     # Check for unknown keys
+    var unknownKeys = newSeq[string]()
     for key in jsonNode.keys:
       if key notin knownFields:
-        error "Invalid configuration option found.", option = key
-        return err("Invalid configuration option found: " & key)
+        unknownKeys.add(key)
+    if unknownKeys.len > 0:
+      error "Unrecognized configuration option(s) found.", option = unknownKeys
+      return err("Unrecognized configuration option(s) found: " & $unknownKeys)
 
     for confField, confValue in fieldPairs(conf):
       if jsonNode.contains(confField):
--- a/liblogosdelivery/logos_delivery_api/node_api.nim
+++ b/liblogosdelivery/logos_delivery_api/node_api.nim
@@ -23,17 +23,21 @@ registerReqFFI(CreateNodeRequest, ctx: ptr FFIContext[Waku]):
     try:
       jsonNode = parseJson($configJson)
     except Exception:
+      let exceptionMsg = getCurrentExceptionMsg()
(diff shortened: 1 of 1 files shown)


AYAHASSAN287 (approved):
THanks Zoltan this will let testers know if the paramter is valid or not

liblogosdelivery/logos_delivery_api/node_api.nim
```diff
@@ -28,6 +28,19 @@ registerReqFFI(CreateNodeRequest, ctx: ptr FFIContext[Waku]):
...
       )
 
+    # Collect known fields
+    var knownFields = initHashSet[string]()
+    for confField, _ in fieldPairs(conf):
+      knownFields.incl(confField)
+    # Check for unknown keys
+    var unknownKeys = newSeq[string]()
+    for key in jsonNode.keys:
+      if key notin knownFields:
```
Ivansete-status: It sounds interesting to strip and maybe perform `toLowAscii` in both comparing parts to avoid case sensitive comparison issues.
```suggestion
    for key in jsonNode.keys:
      let formattedString = ($jsonNode[confField]).strip(chars = {'\"'})
      if formattedString notin knownFields:
```

Ivansete-status: Maybe we can avoid the second loop (`    for confField, confValue in fieldPairs(conf):`) by adding an `else` statement.

liblogosdelivery/logos_delivery_api/node_api.nim
```diff
@@ -28,6 +28,19 @@ registerReqFFI(CreateNodeRequest, ctx: ptr FFIContext[Waku]):
...
+        unknownKeys.add(key)
+    if unknownKeys.len > 0:
+      error "Unrecognized configuration option(s) found.", option = unknownKeys
```
Ivansete-status: Minor nitpick :)
```suggestion
      error "Unrecognized configuration option(s) found", option = unknownKeys
```

Ivansete-status (approved):
LGTM! Thanks for it! 🙌 
The only suggestion is to apply strip and make sure we don't incur into case-sensitive comparison issues. Also, maybe we can only keep the loop that you are suggesting, which sounds better to me.
Thanks!

fcecin (approved):

liblogosdelivery/logos_delivery_api/node_api.nim
NagyZoltanPeter: Addressed in latest commit! Refactored the whole cycle of config key handling.

NagyZoltanPeter: Yepp, its refactored now.

PR #3803, opened 2026-04-08
add main loop lag monitor

### LINKED ISSUE #3804: monitor main chronos loop performance
The logos delivery node runs under nim-chronos, in a single thread.
We need a lightweight mechanism to measure the pressure the main loop is supporting each time, and change the node's health state accordingly if the node is accumulating too many async tasks.

Ivansete-status:
Monitor and control how much pressure the main chronos loop is supporting each time.

- New gauge metric `event_loop_lag_seconds`.
- New possible node `HealthStatus.EVENT_LOOP_LAGGING` state if the main loop lagging for more than five seconds.

closes https://github.com/logos-messaging/logos-delivery/issues/3804

### DIFF
--- a/waku/node/health_monitor/event_loop_monitor.nim
+++ b/waku/node/health_monitor/event_loop_monitor.nim
@@ -1,0 +1,48 @@
+{.push raises: [].}
+
+import chronos, chronicles, metrics
+
+logScope:
+  topics = "waku event_loop_monitor"
+
+const CheckInterval = 5.seconds
+
+declarePublicGauge event_loop_lag_seconds,
+  "chronos event loop lag in seconds: difference between actual and expected wake-up interval"
+
+proc eventLoopMonitorLoop*() {.async.} =
+  ## Monitors chronos event loop responsiveness.
+  ##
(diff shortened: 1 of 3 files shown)


fcecin (approved):
Nice!

darshankabariya (commented):
LGTM !!

darshankabariya (approved):

waku/node/health_monitor/event_loop_monitor.nim
```diff
@@ -0,0 +1,58 @@
...
+  var lagWasHigh = false
+  while true:
+    await sleepAsync(CheckInterval)
+
+    let now = Moment.now()
+    let actualElapsed = now - lastWakeup
+    let lag = actualElapsed - CheckInterval
+    let lagSecs = lag.nanoseconds.float64 / 1_000_000_000.0
+
+    event_loop_lag_seconds.set(lagSecs)
```
arnetheduck, line 39: the better way to do this is a counter that accumulates the lag - this way, the value doesn't get overwritten on each loop iteration which easily can be missed (ie a large lag spike overwritten by the next loop iter)

PR #3805, opened 2026-04-08
fix: recv_service delivers store-recovered messages

fcecin:
## Description

The recv_service was querying the store for missed messages, fetching them, logging "UNSUCCESSFUL / Missed message" line, and not calling MessageReceivedEvent, so store-recovered messages were never delivered via the LMAPI to the app (for messages not received live via relay).

This PR now wires the RecvService correctly to the MessageReceivedEvent and adds a regression test for this. Also only logs new store messages that are actually new and removed "missed" from the log message ("missed" means straggler message, we don't check for that at the recv_service level; that's e.g. SDS / Reliability API).

NOTE: The RecvService has a Store polling interval of 5 minutes, which would catch anything that e.g. SDS can't detect first. This looks like a good default; we can make this a configuration option in the future.

## Changes

* recv_service now delivers store-recovered messages via MessageReceivedEvent
* add regression test_api_receive to prove store recovery actually delivers messages
* confusing "UNSUCCESSFUL / Missed message" log replaced with "recv service store-recovered message" and is only logged for actually new messages
* removed some dead/duplicated code

## Issue

closes #3796

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -83,7 +83,7 @@ jobs:
           path: |
             nimbledeps/
             nimble.paths
-          key: ${{ runner.os }}-nimbledeps-${{ hashFiles('nimble.lock') }}
+          key: ${{ runner.os }}-nimbledeps-${{ hashFiles('nimble.lock', 'BearSSL.mk', 'Nat.mk') }}
 
       - name: Install nimble deps
         if: steps.cache-nimbledeps.outputs.cache-hit != 'true'
@@ -136,7 +136,7 @@ jobs:
           path: |
             nimbledeps/
             nimble.paths
-          key: ${{ runner.os }}-nimbledeps-${{ hashFiles('nimble.lock') }}
+          key: ${{ runner.os }}-nimbledeps-${{ hashFiles('nimble.lock', 'BearSSL.mk', 'Nat.mk') }}
 
       - name: Install nimble deps
         if: steps.cache-nimbledeps.outputs.cache-hit != 'true'
@@ -215,7 +215,7 @@ jobs:
           path: |
             nimbledeps/
             nimble.paths
-          key: ${{ runner.os }}-nimbledeps-${{ hashFiles('nimble.lock') }}
+          key: ${{ runner.os }}-nimbledeps-${{ hashFiles('nimble.lock', 'BearSSL.mk', 'Nat.mk') }}
 
       - name: Install nimble deps
         if: steps.cache-nimbledeps.outputs.cache-hit != 'true'
--- a/BearSSL.mk
+++ b/BearSSL.mk
@@ -22,6 +22,13 @@
 BEARSSL_NIMBLEDEPS_DIR := $(shell ls -dt $(CURDIR)/nimbledeps/pkgs2/bearssl-* 2>/dev/null | head -1)
 BEARSSL_CSOURCES_DIR   := $(BEARSSL_NIMBLEDEPS_DIR)/bearssl/csources
 
+BEARSSL_UNAME_M := $(shell uname -m)
+ifeq ($(BEARSSL_UNAME_M),x86_64)
+  PORTABLE_BEARSSL_CFLAGS := -W -Wall -Os -fPIC -mssse3
+else
+  PORTABLE_BEARSSL_CFLAGS := -W -Wall -Os -fPIC
+endif
+
 .PHONY: clean-bearssl-nimbledeps rebuild-bearssl-nimbledeps
 
 clean-bearssl-nimbledeps:
@@ -36,4 +43,4 @@ ifeq ($(BEARSSL_NIMBLEDEPS_DIR),)
 	$(error No bearssl package found under nimbledeps/pkgs2/ — run 'make update' first)
 endif
 	@echo "Rebuilding bearssl from $(BEARSSL_CSOURCES_DIR)"
-	+ "$(MAKE)" -C "$(BEARSSL_CSOURCES_DIR)" lib
+	+ "$(MAKE)" -C "$(BEARSSL_CSOURCES_DIR)" CFLAGS="$(PORTABLE_BEARSSL_CFLAGS)" lib
--- a/waku/events/delivery_events.nim
+++ b/waku/events/delivery_events.nim
@@ -1,21 +1,5 @@
 import waku/waku_core/[message/message, message/digest], waku/common/broker/event_broker
 
-type DeliveryDirection* {.pure.} = enum
-  PUBLISHING
-  RECEIVING
-
-type DeliverySuccess* {.pure.} = enum
-  SUCCESSFUL
-  UNSUCCESSFUL
-
-EventBroker:
-  type DeliveryFeedbackEvent* = ref object
-    success*: DeliverySuccess
-    dir*: DeliveryDirection
-    comment*: string
-    msgHash*: WakuMessageHash
-    msg*: WakuMessage
-
 EventBroker:
   type OnFilterSubscribeEvent* = object
     pubsubTopic*: string
--- a/waku/node/delivery_service/recv_service/recv_service.nim
+++ b/waku/node/delivery_service/recv_service/recv_service.nim
@@ -12,7 +12,6 @@ import
     waku_store/common,
     waku_filter_v2/client,
     waku_core/topics,
-    events/delivery_events,
     events/message_events,
     waku_node,
     common/broker/broker_context,
@@ -27,7 +26,8 @@ const PruneOldMsgsPeriod = chronos.minutes(1)
 const DelayExtra* = chronos.seconds(5)
   ## Additional security time to overlap the missing messages queries
 
-type TupleHashAndMsg = tuple[hash: WakuMessageHash, msg: WakuMessage]
+type TupleHashAndMsg =
+  tuple[hash: WakuMessageHash, msg: WakuMessage, pubsubTopic: PubsubTopic]
 
 type RecvMessage = object
   msgHash: WakuMessageHash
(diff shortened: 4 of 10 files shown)


fcecin:
For sure the interop-tests failure is unrelated to this PR. Maybe it's from the nimble migration?

fcecin:
Lost the CI machine lottery twice in a row: the CI runner that built the docker image injected instructions that the CI runner for the interop-tests considered SIGILL. Fixed by forcing the BearSSL and NAT deps into the `-mssse3` baseline.

NagyZoltanPeter (approved):
My 2cs:
- I think the BearSSL and Nat compile issue is unrelated to this PR. Next time it worth a separate PR.
- We need to rework this store use in these terms IMHO - btw already working on it - because we the delivery service started using store node(s) in an optimistic - heavy - way. I'm adding offline period recovery to this, that also needs this periodic check.
So we need to tunnel all of the different store checks and actual message retrieval into one channel.

But all in all this looks good as is! Thank you.

waku/node/delivery_service/recv_service/recv_service.nim
```diff
@@ -59,88 +59,79 @@ proc getMissingMsgsFromStore(
...
+    return true
+  return false
+
+proc checkStore*(self: RecvService) {.async.} =
+  ## Checks the store for messages that were not received directly and
+  ## delivers them via MessageReceivedEvent.
+  self.endTimeToCheck = getNowInNanosecondTime()
+
+  var msgHashesInStore = newSeq[WakuMessageHash](0)
+  for pubsubTopic, contentTopics in self.subscriptionManager.subscribedTopics:
```
NagyZoltanPeter, line 94: Just an idea, we won't recover messages of subscribed topics that we unsubscribed in between the period of checks. I mean, if we missed messages for `content_topic_a` but just before the store check period finishes we unsubscribe from it -> we will not retrieve those messages that were in interest for some fraction of the period of concern of the query....
I see this is more like an undefined behavior...

.github/workflows/ci.yml
```diff
@@ -83,7 +83,7 @@ jobs:
           path: |
             nimbledeps/
             nimble.paths
-          key: ${{ runner.os }}-nimbledeps-${{ hashFiles('nimble.lock') }}
+          key: ${{ runner.os }}-nimbledeps-${{ hashFiles('nimble.lock', 'BearSSL.mk', 'Nat.mk') }}
```
Ivansete-status, line 86: Good point! I wonder if we should also add zerokit somehow

waku/events/delivery_events.nim
```diff
@@ -1,21 +1,5 @@
 import waku/waku_core/[message/message, message/digest], waku/common/broker/event_broker
 
-type DeliveryDirection* {.pure.} = enum
-  PUBLISHING
-  RECEIVING
-
-type DeliverySuccess* {.pure.} = enum
-  SUCCESSFUL
-  UNSUCCESSFUL
```
Ivansete-status, line 9: Why not needed or where it was duplicated?

waku/node/delivery_service/recv_service/recv_service.nim
```diff
@@ -59,88 +59,79 @@ proc getMissingMsgsFromStore(
...
-    self.startTimeToCheck = self.endTimeToCheck
-
 proc processIncomingMessageOfInterest(
     self: RecvService, pubsubTopic: string, message: WakuMessage
-) =
-  ## Resolve an incoming network message that was already filtered by topic.
+): bool =
   ## Deduplicate (by hash), store (saves in recently-seen messages) and emit
   ## the MAPI MessageReceivedEvent for every unique incoming message.
+  ## Returns true if the message was new and delivered.
```
Ivansete-status: The delivered term in this case may confuse a little

```suggestion
  ## Returns true if the message was new and the MessageReceivedEvent was properly emitted.
```

waku/node/delivery_service/recv_service/recv_service.nim
```diff
@@ -59,88 +59,79 @@ proc getMissingMsgsFromStore(
...
+  var msgHashesInStore = newSeq[WakuMessageHash](0)
+  for pubsubTopic, contentTopics in self.subscriptionManager.subscribedTopics:
+    let storeResp: StoreQueryResponse = (
+      await self.node.wakuStoreClient.queryToAny(
+        StoreQueryRequest(
+          includeData: false,
+          pubsubTopic: some(pubsubTopic),
+          contentTopics: toSeq(contentTopics),
+          startTime: some(self.startTimeToCheck - DelayExtra.nanos),
+          endTime: some(self.endTimeToCheck + DelayExtra.nanos),
```
Ivansete-status, line 102: This is great for now but we might need to have a _cursor_-like query, i.e., keep querying from a known position. The time window and add this arbitrary `DelayExtra` seems as it is not going to handle scenarios with high jitter.

waku/node/delivery_service/recv_service/recv_service.nim
Ivansete-status, line 138: Sorry, this is completely aside from original PR's purpose.
We need to change that and instead, retrieve the last received timestamp, or maybe better, the latest received msgHash

Besides, I think we need to initialize the `self.endTimeToCheck` to `now` and then perform the store retrieval backwards, instead of forward. I'm more interested in the messages that I missed recently.

waku/node/delivery_service/recv_service/recv_service.nim
```diff
@@ -59,88 +59,79 @@ proc getMissingMsgsFromStore(
...
+        )
+      )
+    ).valueOr:
+      error "msgChecker failed to get remote msgHashes",
+        pubsubTopic = pubsubTopic, cTopics = toSeq(contentTopics), error = $error
+      continue
+
+    msgHashesInStore.add(storeResp.messages.mapIt(it.messageHash))
```
Ivansete-status: This is interesting but I believe is better to perform the store retrieval in a more lazy manner so that we can  keep giving info to LMAPI faster. Instead of doing that, I suggest performing the `processIncomingMessageOfInterest` logic on every loop and have `msgHashesInStore` as a local let attribute on each iteration.
```suggestion
    let msgHashesInStore = storeResp.messages.mapIt(it.messageHash)
```

waku/node/delivery_service/recv_service/recv_service.nim
```diff
@@ -176,7 +167,7 @@ proc startRecvService*(self: RecvService) =
           shard = event.topic, contenttopic = event.message.contentTopic
         return
 
-      self.processIncomingMessageOfInterest(event.topic, event.message),
+      discard self.processIncomingMessageOfInterest(event.topic, event.message),
```
Ivansete-status, line 169: Should we better track this error?

BearSSL.mk
```diff
@@ -22,6 +22,13 @@
 BEARSSL_NIMBLEDEPS_DIR := $(shell ls -dt $(CURDIR)/nimbledeps/pkgs2/bearssl-* 2>/dev/null | head -1)
 BEARSSL_CSOURCES_DIR   := $(BEARSSL_NIMBLEDEPS_DIR)/bearssl/csources
 
+BEARSSL_UNAME_M := $(shell uname -m)
+ifeq ($(BEARSSL_UNAME_M),x86_64)
+  PORTABLE_BEARSSL_CFLAGS := -W -Wall -Os -fPIC -mssse3
+else
+  PORTABLE_BEARSSL_CFLAGS := -W -Wall -Os -fPIC
+endif
+
```
Ivansete-status, line 31: This is interesting but seems unrelated. Shall we have it in a separate PR? Thanks!

Ivansete-status (approved):
LGTM! Thanks for it! 🙌 
I just added some nitpicks that I hope you find useful
Cheers

waku/node/delivery_service/recv_service/recv_service.nim
fcecin: My thinking is that although Logos Delivery is being developed for human-messaging-heavy use cases, but it could also be used for machine-to-machine. At this level, at least to me, we are talking about data pipes. "Unsubscribe" or "Subscribe" here is a metaphor for a relatively low-level connection. So the upper layer (e.g. Reliability API) would have to deal with rather simple or blunt semantics for the services offered at the current LMAPI level that implements uncomplicated Send, Receive, Sub/Unsub (on top of the kernel Vac proto impls).

But it's still a (relatively) dumb pipe for bytes. So if the upper layer is worried that the current LMAPI level just cuts off delivery when you say "unsubscribe," then don't. Wait until you know you are safe to do that. This is similar to border conditions in socket protocols. If the app using the socket layer wants fancy transport guarantees, it has to teach those to the lower-level mechanism. Or at least that's how I'm understanding this currently.

So the simpler and clearer the current LMAPI level is defined as, the better. If we can nail down the semantics here, ones we can pretty much stick with until the end of time, then we win. But they can't be a thing that keeps shifting as we discover more about QoS in messaging products.

(Not sure if that makes any actual sense)

waku/events/delivery_events.nim
fcecin: This was the code I was *supposed* to hook into when I finished upgrading the RecvService. This was Zoltan's original design which was supposed to be the actual higher-level interface for the *whole* DeliveryService (Send *and* Receive). I didn't understand it when I picked up the RecvService, so I ended up writing MessageReceivedEvent (I think I added this) to match MessageSentEvent. Since I didn't understand it I just left it here. When I glazed over the Store integration with the RecvService, I missed that these were two completely different things. That screwup was entirely on me.

So it turns out that since we have the more specific e.g. MessageSentEvent and MessageReceivedEvent emitted by the individual components of the DeliveryService, we don't need the unified one at the DeliveryService, so we can just remove. They are not wired to anything.

waku/node/delivery_service/recv_service/recv_service.nim
fcecin: This is not an error. The return bool is whether the message was new or not.

BearSSL.mk
fcecin: Yes, will move the entire CI fix to another PR.

.github/workflows/ci.yml
fcecin: I'll take a look at it; will move the CI fix to another PR.

waku/node/delivery_service/recv_service/recv_service.nim
fcecin: Good point! I think it's what Zoltan was raising: there's something different that we want to do here. I think this cursor idea might be better than guessing network conditions. Simple query; let the upper layer add any padding.

waku/node/delivery_service/recv_service/recv_service.nim
fcecin: I think we have a new Issue (improvement/fix) incoming to track this. We should polish the Store use at the simple send/rcv/sub/unsub LMAPI level that we have now. IMO it's great that the network has a memory at this level and we should use it in some way that is the correct one. Then the Reliability API can build on top of that.

waku/node/delivery_service/recv_service/recv_service.nim
fcecin: Fixed!

PR #3806, opened 2026-04-09
Fix BearSSL and NAT lib build reproducibility

fcecin:
## Description

BearSSL and the NAT lib dependencies are using native arch by default, so the resulting binary is not portable. This manifests as e.g. interop-tests failures in CI as one CI machine builds the docker image with a larger instruction set and another CI machine runs the interop tests with a reduced instruction set.

Forces -mssse3 for these dependencies.

Also added them to the cache key in nibledeps. (Zerokit uses Cargo so it's not covered by the nimbledeps cache, and it should be fine).

## Changes

* pass -mssse3 on x86_64 to BearSSL and NAT C lib builds
* add BearSSL.mk and Nat.mk to nimbledeps cache key

## Issue

Maintenance Y2026H1 #3686

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -83,7 +83,7 @@ jobs:
           path: |
             nimbledeps/
             nimble.paths
-          key: ${{ runner.os }}-nimbledeps-${{ hashFiles('nimble.lock') }}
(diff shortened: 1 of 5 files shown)


Ivansete-status (approved):
Wonderful thanks ! 💯

darshankabariya (approved):
Very Cool !!

PR #3807, opened 2026-04-09
feat(mix): cover traffic with constant rate and spam protection toggle

chaitanyaprem:
## Summary

Adds constant-rate cover traffic and a spam-protection toggle to waku's mix integration.

**Stacked on top of [#3931](https://github.com/logos-messaging/logos-delivery/pull/3931)** (DoS protection + libp2p v2.0.0). Merge order: #3931 → this PR.

## What lands here

| Commit | What |
|---|---|
| `fc6af2fd` | `feat(mix): cover traffic with constant rate` — `ConstantRateCoverTraffic` wired into `WakuMix.new`; `--mix-user-message-limit` / `--mix-disable-spam-protection` CLI flags; `check_cover_traffic.sh` metrics helper; per-node sim configs. |
| `4365df4b` | `fixup(mix): address PR #3807 review + sim alignment` — drops the magic-2 / hardcoded-10s cover-traffic defaults; sources `totalSlots` + `epochDuration` from spam-protection plugin's config when RLN is on, and from named waku constants (`WakuCoverTrafficTotalSlots=40`, `WakuCoverTrafficEpochDuration=60s` ≈ 10 cover packets/min/node) when RLN is off; chat clients pass `--rln-user-message-limit=4`. |

## Scope

Diff is **16 files / +177 / −23** vs the base PR's tip, all cover-traffic-specific:

```
simulations/mixnet/README.md, check_cover_traffic.sh, config[1-4].toml
simulations/mixnet/run_chat_mix{,1}.sh, setup_credentials.nim
tools/confutils/cli_args.nim
waku/factory/conf_builder/mix_conf_builder.nim
waku/factory/node_factory.nim
waku/factory/waku_conf.nim
waku/node/waku_node.nim                     (mountMix gains disableSpamProtection)
waku/waku_mix/protocol.nim                  (the integration)
```

No libp2p-version-compat changes here — those live in the base PR.

## How to Test

See [`simulations/mixnet/README.md`](simulations/mixnet/README.md) for full setup. The fast path:

```bash
make wakunode2 chat2mix
cd simulations/mixnet
bash roundtrip_check.sh                 # ~5-8 min: 5 mix nodes + alice/bob roundtrip
bash check_cover_traffic.sh             # poll mix_cover_* / mix_slot_* metrics
```

**Hardware note:** RLN proof generation is computationally expensive. Running 5 nodes locally with proofs enabled can overwhelm a modest laptop. Use a beefier machine or pass `--mix-disable-spam-protection` to skip RLN for quick checks.

## Validation

End-to-end mixnet sim (5 mix nodes + alice + bob via SURB) on this branch.

**Result: `[PASS] bob received the message`** — `<Jun 05, 12:48> alice: sim-roundtrip-1780643218`.

### Cover traffic active across the pool

Live metrics during the run (`check_cover_traffic.sh`):

- `mix_cover_emitted_total` increasing on every reachable node
- `mix_cover_received_total` matching emissions — cover packets making full 3-hop round trips
- `mix_cover_error_total{error="BUILD_FAILED"}` appears briefly during pool warm-up (expected — sphinx packet build fails until ≥3 mix peers are discovered), then stops

### RLN proof gen + verify (cover traffic exercises the proof path)

Per-node post-sim log counts:

| node | `MixRlnSpamProtection started` | `generateProof` called | `verified successfully` | `SPAM_PROOF_*` errors |
|---|---|---|---|---|
| mix_node | 1 | 73 | 73 | 0 |
| mix_node1 | 1 | 73 | 73 | 0 |
| mix_node2 | 1 | 70 | 70 | 0 |
| mix_node3 | 1 | 53 | 53 | 0 |
| mix_node4 | 1 | 27 | 27 | 0 |
| **total** | **5** | **296** | **296** | **0** |

- **gen == verified** per node — every proof emitted by an in-path mix node was verified by the next hop.
- **0 `SPAM_PROOF_*` errors** — no `GEN_FAILED`, `EMBED_FAILED`, `EXTRACTION_FAILED`, `VERIFY_ERROR`, or `INVALID` across the run.
- ~296 proofs vs the ~5 from a single roundtrip without cover traffic — cover-traffic emission drives the RLN throughput.

### Build

- `make wakunode2`: clean, 32.6s, 416,464 lines.
- `make chat2mix`: clean, 32.4s, 394,057 lines.

## Pre-merge checklist (depends on base PR landing)

1. Wait for [#3931](https://github.com/logos-messaging/logos-delivery/pull/3931) to land — this PR's base disappears otherwise.
2. After #3931 merges to master, this PR's base auto-retargets to `master`. Verify the diff still cleanly applies.
3. If [#3772 (service discovery)](https://github.com/logos-messaging/logos-delivery/pull/3772) lands first, no action needed here — overlap is limited to `waku_discv5`/`waku_kademlia` already in #3931, not this PR.


chaitanyaprem:
## Cover Traffic + RLN Spam Protection Simulation Report

**Date:** 2026-04-10 | **Duration:** ~10 minutes

### System Configuration

| Component | Value |
|-----------|-------|
| CPU | Apple M2 Pro |
| Cores | 10 |
| Memory | 32 GB |
| Architecture | arm64 (macOS Darwin 25.3.0) |

### Simulation Configuration

| Parameter | Value |
|-----------|-------|
| Nodes | 5 mix nodes (1 bootstrap + 4) |
| RLN Spam Protection | **Enabled** (`mix-disable-spam-protection=false`) |
| User Message Limit (R) | 2 slots per epoch |
| Epoch Duration (P) | 10 seconds |
| Mix Path Length (L) | 3 hops |
| Cover Traffic Precomputation | Disabled (default) |
| Tree Members | 7 (5 mix nodes + 2 chat clients) |
| Log Level | TRACE |

### System Resource Usage

| Metric | Value |
|--------|-------|
| Memory per node | ~55-67 MB RSS |
| Steady-state CPU (all 5 nodes) | **1-5% total** |
| Peak CPU (burst during epoch proof gen) | **~142% total** |
| Load average | 1.65 - 2.84 |

CPU usage is **bursty** — each node generates 1-2 RLN proofs per 10s epoch (~200ms of work), then idles. Spikes are brief and non-disruptive on this hardware.

### RLN Proof Generation Times

206 proofs collected across all 5 nodes:

| Stat | Value |
|------|-------|
| Min | 25ms |
| Max | 605ms |
| **Average** | **112.9ms** |
| Median | 71.5ms |
| P95 | 340ms |

**Distribution:**
```
  0-49ms  |  75 | ###########################################################################
 50-99ms  |  49 | #################################################
100-149ms |  35 | ###################################
150-199ms |  14 | ##############
200-299ms |  17 | #################
300-499ms |  13 | #############
500ms+    |   3 | ###
```

60% of proofs complete under 100ms. Tail latency (300ms+) occurs when multiple nodes generate proofs simultaneously at epoch boundaries (single-threaded chronos async per process).

### Cover Traffic Metrics (Final Snapshot)

| Node | Port | Emitted | Received | Slot Exhausted (fwd) | Forwarded (Cover) | Forwarded (Intermediate) |
|------|------|---------|----------|----------------------|-------------------|--------------------------|
| 0 (bootstrap) | 8009 | 23 | 11 | 5 | 23 | 56 |
| 1 | 8010 | 31 | 24 | 10 | 31 | 52 |
| 2 | 8011 | 31 | 27 | 2 | 31 | 44 |
| 3 | 8012 | 31 | 29 | 5 | 31 | 50 |
| 4 | 8013 | 31 | 26 | 4 | 31 | 49 |

- All 5 nodes actively emitting and receiving cover traffic
- Receive rate ~80-93% of emitted (losses due to slot exhaustion at intermediate hops)
- Node 0 has fewer emissions (started earlier, different epoch alignment)

### Errors Observed

| Error | Total Count | Affected Nodes | Notes |
|-------|-------------|----------------|-------|
| SPAM_PROOF_GEN_FAILED (Cover) | 8 | Node 0 | Early epoch proofs before tree sync |
| SPAM_PROOF_GEN_FAILED (Intermediate) | 4 | Nodes 0, 2, 3 | Rate limit exhausted during forwarding — "Message id (2) is not within user_message_limit (2)" |
| SLOT_EXHAUSTED (forward) | 26 | All nodes | Expected — R=2 is too low for both cover + forwarding |

The `SPAM_PROOF_GEN_FAILED` errors during forwarding occur because the R-budget (2 slots) is shared across cover origination AND forwarding. When a node uses both slots for its own cover traffic, it has none left to generate proofs for forwarded packets. This is expected with R=2; higher limits (e.g., R=10+) would provide headroom for both.

### Conclusion

Cover traffic with RLN spam protection works correctly on consumer hardware (M2 Pro). Resource usage is manageable with R=2 (~5% steady CPU for 5 nodes). The main limitation is the low rate limit causing slot exhaustion during forwarding — this is a configuration constraint, not a bug.

chaitanyaprem:
### Addendum: Rate Limit Analysis & Recommended Configuration

The `SPAM_PROOF_GEN_FAILED` and `SLOT_EXHAUSTED` errors observed in the test above are caused by using a non-clean rate limit value.

**Root Cause:**

With R=2 (user message limit) and L=3 (path length):
- Emission interval = `(1+L) × P / R` = `4 × 10 / 2` = **20s**
- Target cover packets per epoch = `R / (1+L)` = `2 / 4` = **0.5** (not an integer)

Because 0.5 is fractional, epoch boundary jitter causes some epochs to see 2 cover emissions and others 0. When 2 land in the same epoch, all slots are consumed by cover traffic, leaving none for forwarding — triggering proof generation failures at intermediate hops.

**Fix:**

R must be a multiple of `(1+L) = 4` for clean integer cover emissions per epoch:

| R | Cover/epoch | Interval | Slots for forwarding | Proofs/min/node |
|---|------------|----------|----------------------|-----------------|
| **4** | 1 | 10s | 3 | ~6 |
| 8 | 2 | 5s | 6 | ~12 |
| 12 | 3 | 3.3s | 9 | ~18 |

**Recommended: R=4** — exactly 1 cover packet per epoch, 3 remaining slots for forwarding, no fractional math or boundary jitter issues. This is also less CPU load than R=2 in practice (fewer spurious double-emissions).

The config files (`config.toml` through `config4.toml`) and `setup_credentials.nim` should be updated from `mix-user-message-limit=2` to `mix-user-message-limit=4` before final merge.

chaitanyaprem:
## Cover Traffic Precomputation + RLN Spam Protection Simulation Report

**Date:** 2026-04-10 | **Duration:** ~4 minutes | **Config change:** `enablePrecomputation=true` in `ConstantRateCoverTraffic.new()`

### System Configuration

| Component | Value |
|-----------|-------|
| CPU | Apple M2 Pro (10 cores) |
| Memory | 32 GB |
| Architecture | arm64 (macOS) |

### Simulation Configuration

| Parameter | Value |
|-----------|-------|
| Nodes | 5 mix nodes |
| RLN Spam Protection | **Enabled** |
| User Message Limit (R) | 2 slots per epoch |
| Epoch Duration (P) | 10 seconds |
| Mix Path Length (L) | 3 hops |
| **Cover Traffic Precomputation** | **Enabled** |

### Key Finding: Precomputation Produces 0 Packets with R=2

With `enablePrecomputation=true`, the precompute loop runs but builds **zero** packets because:

```
targetCount = totalSlots / (1 + pathLength) = 2 / 4 = 0 (integer division)
```

All cover traffic falls through to on-demand building. The `mix_cover_precomputed_total` metric remained at `0.0` on all nodes and `mix_cover_emitted{type="prebuilt"}` was never incremented. All emissions were `{type="on_demand"}`.

**For precomputation to be effective, R must be ≥ 4** (so `R / (1+L) ≥ 1`).

### System Resource Usage

| Metric | Value |
|--------|-------|
| Memory per node | ~70-75 MB RSS |
| Steady-state CPU (all 5 nodes) | **0.4-6.7% total** |
| Load average | 1.54 - 2.37 |

No measurable difference from the non-precomputation run — expected since precomputation built 0 packets.

### RLN Proof Generation Times (124 proofs)

| Stat | Value |
|------|-------|
| Min | 24ms |
| Max | 590ms |
| Average | 116.0ms |
| Median | 79ms |
| P95 | 322ms |

**Distribution:**
```
  0-49ms  |  42 | ##########################################
 50-99ms  |  32 | ################################
100-149ms |  16 | ################
150-199ms |  14 | ##############
200-299ms |  12 | ############
300-499ms |   6 | ######
500ms+    |   2 | ##
```

Consistent with the previous run (avg 112.9ms vs 116.0ms).

### Cover Traffic Metrics (Final — 240s)

| Node | Prebuilt | On-Demand | Received | Precomputed | Slot Exhausted | Proof Fail (Intermediate) |
|------|----------|-----------|----------|-------------|----------------|--------------------------|
| 0 | 0 | 16 | 16 | 0 | 0 | 1 |
| 1 | 0 | 16 | 14 | 0 | 1 | 1 |
| 2 | 0 | 15 | 13 | 0 | 3 | 0 |
| 3 | 0 | 16 | 12 | 0 | 3 | 0 |
| 4 | 0 | 15 | 8 | 0 | 5 | 1 |

### Conclusion

Precomputation with R=2 is a no-op due to integer division yielding 0 target packets. The precompute loop activates but produces nothing, falling back entirely to on-demand packet building. To test precomputation meaningfully, R should be set to ≥4 (yielding ≥1 precomputed packet per epoch). All other behavior (emission, forwarding, slot exhaustion, proof generation) is consistent with the non-precomputation run.

chaitanyaprem:
## Proof Token Reuse Validation — R=4 with Precomputation + RLN

**Date:** 2026-04-10 | **Duration:** ~3 minutes | **Feature:** Opaque proof token reuse for discarded precomputed cover packets

### What Changed

When a precomputed cover packet is discarded (because a forwarded/originated packet claims its slot), the RLN messageId used for that proof is now **reclaimed and reused** instead of being wasted. This prevents premature rate limit exhaustion.

Implementation uses opaque proof tokens — `SpamProtection.generateProof()` returns a `ProofResult(proof, token)`. The token is stored in the `CoverPacket` and returned via `reclaimProofToken()` when the packet is discarded. The RLN plugin decodes the token internally, recovering the messageId for reuse.

### Results: Zero Proof Failures ✅

| Node | Prebuilt | On-Demand | Received | Precomputed | Exhausted | **Proof Fail (Cover)** | **Proof Fail (Fwd)** | Reclaimed Tokens |
|------|----------|-----------|----------|-------------|-----------|----------------------|---------------------|------------------|
| 0 | 11 | 8 | 17 | 20 | 2 | **0** | **0** | 9 |
| 1 | 13 | 7 | 17 | 20 | 2 | **0** | **0** | 7 |
| 2 | 9 | 10 | 17 | 20 | 3 | **0** | **0** | 11 |
| 3 | 12 | 7 | 17 | 20 | 1 | **0** | **0** | 8 |
| 4 | 12 | 8 | 20 | 20 | 1 | **0** | **0** | 8 |

**43 total proof tokens reclaimed and reused** across all nodes. Zero `SPAM_PROOF_GEN_FAILED` errors. Zero `"Message id not within user_message_limit"` errors.

### Comparison: Before vs After Proof Token Reuse

| Metric | R=4 Without Reuse | R=4 With Reuse |
|--------|-------------------|----------------|
| Proof failures (forwarding) | 4 | **0** |
| Slot exhaustions | 5 | **9** (more forwarding now succeeds) |
| CPU (steady state) | 18.5% | **6.1%** |
| Proof gen avg | 117ms | **107ms** |
| Reclaimed tokens | N/A | **43** |

### MessageId Reuse Trace (Node 1)

```
22:08:20 generateProof messageId=0 reused=true   ← reusing freed ID
22:08:20 generateProof messageId=1 reused=true
22:08:21 Reclaimed proof token messageId=1        ← cover packet discarded, ID returned
22:08:21 generateProof messageId=1 reused=true    ← immediately reused for forwarded packet
22:08:21 generateProof messageId=2 reused=true
22:08:30 generateProof messageId=0 reused=true    ← new epoch, IDs reset
```

### System Resources

| Metric | Value |
|--------|-------|
| Memory per node | ~56-59 MB RSS |
| Steady-state CPU | **6.1% total** |
| Load average | 2.60 - 3.24 |

### Configuration

| Parameter | Value |
|-----------|-------|
| Nodes | 5 |
| RLN Spam Protection | Enabled |
| R (user message limit) | 4 |
| Epoch Duration | 10s |
| Path Length | 3 |
| Precomputation | Enabled (same-epoch) |
| Proof Token Reuse | **Enabled** |

chaitanyaprem:
## Cover Rate Fraction — Simulation Comparison (f=0.7 vs f=1.0)

**Date:** 2026-04-23 | **Feature:** `cover_rate_fraction` config (default `f = 0.7`, per spec §7)

Tracks upstream PR [vacp2p/nim-libp2p#2322](https://github.com/vacp2p/nim-libp2p/pull/2322). Vendor bump in this branch: `a9d5f3ff`.

### Setup (identical for both runs)

| Parameter | Value |
|---|---|
| Nodes | 5 (1 bootstrap + 4) |
| R (mix-user-message-limit) | 4 |
| P (epoch) | 10s |
| L (path length) | 3 |
| RLN spam protection | Enabled |
| Precomputation | Disabled |
| Duration | ~26 epochs (~260s) |

### Per-node totals over the run

| Metric | Node 0 | Node 1 | Node 2 | Node 3 | Node 4 |
|---|---|---|---|---|---|
| **f = 0.7 (new default)** | | | | | |
| Cover emitted | 13 | 13 | 13 | 13 | 13 |
| Intermediate forwarded | 26 | 28 | 32 | 20 | 24 |
| Slot-rejected (forward) | 0 | 0 | 3 | 0 | 0 |
| SLOT_EXHAUSTED | 0 | 0 | 0 | 0 | 0 |
| SPAM_PROOF_GEN_FAILED | 0 | 0 | 0 | 0 | 0 |
| **f = 1.0 (prior default, for comparison)** | | | | | |
| Cover emitted | 27 | 27 | 27 | 27 | 27 |
| Intermediate forwarded | 52 | 55 | 54 | 50 | 46 |
| Slot-rejected (forward) | 0 | 3 | 3 | 0 | 0 |
| SLOT_EXHAUSTED | 0 | 3 | 3 | 0 | 0 |
| SPAM_PROOF_GEN_FAILED | 1 | 2 | 2 | 0 | 0 |

### Per-node per-epoch averages

| Metric (avg/epoch/node) | f = 0.7 | f = 1.0 | Δ |
|---|---|---|---|
| Cover emitted | 0.5 | 1.04 | ≈ ½ |
| Intermediate forwarded | ~1.0 | ~2.0 | ≈ ½ |
| **Total outgoing** | **~1.5** | **~3.0** | **≈ ½** |
| % of R=4 budget used | ~37% | ~75% | — |
| Emission interval (logged) | 20s | 10s | 2× |

### Observations

1. **~50% less traffic per node per epoch under f=0.7.** `floor(0.7 × 4) = 2` halves the scaled-slot count at R=4, doubling emission interval from 10s → 20s. Intermediate forwarding also halves because neighbors emit half as much cover.

2. **Budget utilization drops 75% → 37%.** Meaningful headroom gain — in f=1.0 at R=4, nodes ran near capacity; under f=0.7 they're under half-used.

3. **All error classes vanish under f=0.7.** The `SPAM_PROOF_GEN_FAILED` (5 total, 3 nodes) and `SLOT_EXHAUSTED` (6 total, 2 nodes) errors from the f=1.0 run disappear entirely — exactly the effect spec §10.5 predicts for the `cover_rate_fraction` headroom.

4. **Some slot-rejected-forwards still occur at f=0.7 (3 on one node).** Natural traffic burstiness briefly exceeding the remaining 2 slots, not cover starving forwards.

5. **Caveat — R=4 floor rounding amplifies the effect.** At R=4, `floor(0.7 × 4) = 2`, so f=0.7 effectively behaves as f=0.5. At larger R (e.g., R=100) the reduction would be a cleaner ~30%. Prod deployments with bigger R won't see the 50% drop — they'll see the nominal 30%.

### Vs. prior proof-token-reuse report on this PR

Prior run (R=4, f=1.0 implicit, precomputation enabled, 3 min, [comment](https://github.com/logos-messaging/logos-delivery/pull/3807#issuecomment-4225312188)): ~1 emission/epoch/node, zero proof failures (thanks to token reuse).

Current f=0.7 (precomputation disabled, 4.3 min): 0.5 emissions/epoch/node, zero proof failures. Qualitative behavior (loop-back healthy, no failures) preserved; quantitative drop of ~50% per-node traffic as expected.

waku/common/option_shims.nim
jm-clius, line 1: This is interesting. Any reason why these functions were removed? Shouldn't we adapt to whatever is considered the most idiomatic, or use a different library?

simulations/mixnet/setup_credentials.nim
```diff
@@ -20,8 +20,8 @@ import
 
 const
   KeystorePassword = "mix-rln-password" # Must match protocol.nim
-  DefaultUserMessageLimit = 100'u64 # Network-wide default rate limit
-  SpammerUserMessageLimit = 3'u64 # Lower limit for spammer testing
+  DefaultUserMessageLimit = 4'u64 # R=4 slots per 10s epoch
+  SpammerUserMessageLimit = 3'u64 # Higher limit for spammer testing
```
jm-clius: Truly a nitpick, but I'm trying to parse the configuration here and elsewhere, and I don't understand the change of "Lower" to "Higher" here. 😅

waku/waku_mix/protocol.nim
```diff
@@ -97,33 +99,50 @@ proc new*(
...
-  # Use peerID in keystore path so multiple peers can run from same directory
-  # Tree path is shared across all nodes to maintain the full membership set
-  let peerId = peermgr.switch.peerInfo.peerId
-  var spamProtectionConfig = defaultConfig()
-  spamProtectionConfig.keystorePath = "rln_keystore_" & $peerId & ".json"
-  spamProtectionConfig.keystorePassword = "mix-rln-password"
-  if userMessageLimit.isSome():
-    spamProtectionConfig.userMessageLimit = userMessageLimit.get()
-  # rlnResourcesPath left empty to use bundled resources (via "tree_height_/" placeholder)
+  let totalSlots = userMessageLimit.get(2)
```
jm-clius: Why use the magic number `2` here and not the default configuration (besides, IIUC your comments previously - this will result in fractional cover-messages-per-epoch)

jm-clius (commented):
Good progress. Feel free to raise a review once this is ready for merging. One concern is complexity in configuration - if we have restrictions on certain configuration items (e.g. user message rate limit should be a multiple of 4), we should refactor our configuration to prevent misconfiguration.

simulations/mixnet/setup_credentials.nim
chaitanyaprem: Ah, this was something i added initially to simulate a spammer but later realized that there is no way to simulate it as zerokit itself seems to have a check once a user crosses his own message limit, it doesnt generate proofs. So, right now this is more of useless code, will try to remove it.

waku/common/option_shims.nim
chaitanyaprem: Forgot to mention that this was. temporary hack that was done to avoid changes to be made in rest of logos delivery code. 
I think the original issue came from libp2p migrating to a newer version which was causing issues in code no related to changes in this PR. 

This file should be removed before merging as i assume this would get fixed by some other PR in master.

waku/waku_mix/protocol.nim
chaitanyaprem: Oh, this is not right. 
You are right and we should use default config. Will fix it.

chaitanyaprem:
> Good progress. Feel free to raise a review once this is ready for merging. One concern is complexity in configuration - if we have restrictions on certain configuration items (e.g. user message rate limit should be a multiple of 4), we should refactor our configuration to prevent misconfiguration.

ah, right...but that is just for simulation env. Once we migrate this to onchain registry, this config would not be required or rather will be fetched from there itself. 
Another reason i had to set the config such as `mix-user-message-limit` is when we run simulation locally, we cant have full rate limit of 100msgs/sec as system can't handle soo much RLN proof generation especially in a 5 node simulation. 
Not sure if there is a way around this rather than labelling it advanced config only useful for simulation. Similarly to test by disabling spam protection, i added the config. 
Ideally these would be deployment level configs and only defaults would be used i guess.

waku/waku_mix/protocol.nim
chaitanyaprem: Modified to pass config properly.

tools/confutils/cli_args.nim
```diff
@@ -642,6 +642,18 @@ with the drawback of consuming some more bandwidth.""",
...
     .}: seq[MixNodePubInfo]
 
+    mixUserMessageLimit* {.
+      desc:
+        "Maximum messages per RLN epoch for mix cover traffic. If not set, uses plugin default.",
+      name: "mix-user-message-limit"
+    .}: Option[int]
+
+    mixDisableSpamProtection* {.
+      desc: "Disable RLN spam protection for mix protocol (for testing only).",
```
Ivansete-status, line 679: Isn't it enough by just using the "rln enabled"?
We are having two parallel RLN development paths. Ideally, RLN should be a protocol-agnostic rate limit component. Maybe we can accept that now but in the upcoming future we will need to refactor that to unify the RLN config into one.

logos_delivery/waku/waku_mix/protocol.nim
```diff
@@ -101,36 +120,58 @@ proc new*(
...
+  # can't outpace proof minting when RLN is on.
+  var ctTotalSlots = WakuCoverTrafficTotalSlots
+  var ctEpochDuration = WakuCoverTrafficEpochDuration
+
+  var spamProtectionOpt = default(Opt[SpamProtection])
+  if not disableSpamProtection:
+    # Initialize spam protection with persistent credentials
+    let peerId = peermgr.switch.peerInfo.peerId
+    var spamProtectionConfig = defaultConfig()
+    spamProtectionConfig.keystorePath = "rln_keystore_" & $peerId & ".json"
```
Ivansete-status, line 136: btw - why having different json keystore file? Doesn't it make it even more difficult to handle? I'd suggest just having one keystore file.

logos_delivery/waku/waku_mix/protocol.nim
```diff
@@ -101,36 +120,58 @@ proc new*(
...
+
+  let ct = ConstantRateCoverTraffic.new(
+    totalSlots = ctTotalSlots,
+    epochDuration = ctEpochDuration,
+    useInternalEpochTimer = disableSpamProtection,
+  )
 
-  let spamProtection = newMixRlnSpamProtection(spamProtectionConfig).valueOr:
-    return err("failed to create spam protection: " & error)
+  var mixRlnSpam: MixRlnSpamProtection
```
Ivansete-status, line 161: Why not have the `MixRlnSpamProtection` definition in this `logos-delivery` repo? That would make things simpler and more direct.

logos_delivery/waku/waku_mix/protocol.nim
```diff
@@ -101,36 +120,58 @@ proc new*(
...
+  if spamProtectionOpt.isSome():
+    mixRlnSpam = MixRlnSpamProtection(spamProtectionOpt.get())
```
Ivansete-status, line 163: I would expect seeing `MixRlnSpamProtection.new(...)`.

Ivansete-status (approved):
Approving to not block but we may need to unify efforts and have one single RLN solution that can be used by any protocol.
Thanks!

tools/confutils/cli_args.nim
chaitanyaprem: yes, but we cant have a single config to disable or enable RLN for both mix and waku relay. Also note that this is meant for testing purposes. 

RLN is a protocol agnostic component, but each protocol will define its own membership and rate limits hence not sure if we can combine everything. 

What if we decide to disable RLN and use something else for spam protection in mix? Then unifying it creates problems right.

logos_delivery/waku/waku_mix/protocol.nim
chaitanyaprem: That is because mix is designed to support different Spam protection mechanisms implemented as plugins, current implemenation is via RLN and hence that plugin lives in a separate repo. If we bring that code into this repo, then we are tying it concretely which i am not sure is a good idea as in future we may decide not to use RLN itself.

logos_delivery/waku/waku_mix/protocol.nim
chaitanyaprem: This is a good idea indeed, will have an issue to track this in the mix-rln-plugin repo so that it uses same format as relay. Looks like the file format is little different as of now and i took the easy path of having a separate file.

chaitanyaprem: created an issue https://github.com/logos-co/mix-rln-spam-protection-plugin/issues/10

chaitanyaprem:
> but we may need to unify efforts and have one single RLN solution that can be used by any protocol.

If you mean for memberships, that is what @adklempner has been working on to have a RLN membership module that can be used by any protocol to register memberships.

logos_delivery/waku/waku_mix/protocol.nim
chaitanyaprem: Done — the call site now uses \`MixRlnSpamProtection.new(...)\`.

This needed a small change in the plugin too: the constructor was renamed from the free proc \`newMixRlnSpamProtection(config)\` to the typedesc form \`MixRlnSpamProtection.new(config)\` (logos-co/mix-rln-spam-protection-plugin@61ee3e5), to match the nim-libp2p `T.new` convention. The plugin pin is bumped accordingly here and on the base branch #3931.

fcecin (approved):
LGTM

tools/confutils/cli_args.nim
```diff
@@ -681,6 +681,12 @@ hence would have reachability issues.""",
       name: "mix-disable-spam-protection"
     .}: bool
 
+    mixDisableCoverTraffic* {.
```
AkshayaMani, line 684: Optional:
Add anonymity warning to match `mixDisableSpamProtection`'s style:
_e.g.,_ (weakens anonymity; for testing only)

AkshayaMani (approved):
Thanks, LGTM! Added an optional suggestion.

PR #3808, opened 2026-04-10
use EWMA to show main loop lag information

Ivansete-status:
This PR is motivated by this @arnetheduck 's comment:
- https://github.com/logos-messaging/logos-delivery/pull/3803#discussion_r3059141572

The main idea is to show how much async load the main loop is supporting each time.

The load is shown similarly to how `top` app shows the load in a machine, i.e., load for `1minutes`, `5m` and `15m`.

In this case, `1` means the main async loop is fully loaded, which means that it is currently supporting the maximum accepted lag.

The `MaxAcceptedLag` is set to `50ms`, hard-coded. Then, a load of `1` indicates that the main loop is having that `MaxAcceptedLag` (50ms). Similarly, a load of `N` indicates the main loop is having `N * MaxAcceptedLag` lag.

### DIFF
--- a/waku/node/health_monitor/event_loop_monitor.nim
+++ b/waku/node/health_monitor/event_loop_monitor.nim
@@ -1,58 +1,78 @@
 {.push raises: [].}
 
+import std/math
 import chronos, chronicles, metrics
 
 logScope:
   topics = "waku event_loop_monitor"
 
-const CheckInterval = 5.seconds
-
-declarePublicGauge event_loop_lag_seconds,
-  "chronos event loop lag in seconds: difference between actual and expected wake-up interval"
+declarePublicGauge event_loop_load,
+  "chronos event loop load EWMA by window (1.0 = sustained lag at MaxAcceptedLag)",
+  labels = ["window"]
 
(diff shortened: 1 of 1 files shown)


darshankabariya (approved):
LGTM

fcecin (approved):

Ivansete-status:
@AlbertoSoutullo - whenever is possible, it would be very interesting to see how these new metrics behave in the lab.

arnetheduck:
this PR doesn't really capture what I meant .. ie metrics are about providing raw data and in the case of lag, the right "kind" of granular data is the accumulated lag, ie if requests cause 2, 3, and 5 s of lag, we have a counter that increases to 10 in those steps. This is similar to counting the total amount of bytes sent etc.

this way, no matter when you sample the metric, you'll see the total amount of lag that happened since startup - tools like prometheus then take this information and apply maths like averages and rates on it to transform it to a pretty graph. Critically, this kind of data massage does not need to happen in the application (in fact, it's worse because the performing computations on metrics in the application leads to data loss typically)

PR #3815, opened 2026-04-16
Fix peer stats endpoint

### LINKED ISSUE #3541: bug:
### Summary
When more than one relay peer is connected, `GET /admin/v1/peers/stats` reports an incorrect total under **Relay peers**. Specifically, `"Total relay peers"` is `1` while other sections (and `GET /admin/v1/peers`) indicate **2+** relay peers are connected.

### Reproduction
Run test **`test_admin_peers_stats_counts`** and compare these values:
- `Relay peers["Total relay peers"]`
- `By Protocols["/vac/waku/relay/2.0.0"]`
- The number of connected peers from `GET /admin/v1/peers`

### Expected behavior
`Relay peers["Total relay peers"]` should equal the actual number of connected relay peers (e.g., **2** when two are connected). All counts within `GET /admin/v1/peers/stats` should be internally consistent and also match `GET /admin/v1/peers`.

### Actual behavior
With two relay peers connected, the stats endpoint returns something like:

```json
{
  "Relay peers": {
    "Total relay peers": 1
  },
  "By Protocols": {
    "/vac/waku/relay/2.0.0": 2
  }
}
```

Meanwhile, `GET /admin/v1/peers` (and the actual connections) show **2** relay peers.

fcecin:
## Description

handlers.nim used .len on intermediary data structures instead of using the actual counts that were already there. 

## Changes

* Fix total relay peers stat (return totalRelayPeers instead of relayPeers.len)
* Fix total mesh peers stat (return totalMeshPeers instead of meshPeers.len)

## Issue

closes #3541

### DIFF
--- a/waku/rest_api/endpoint/admin/handlers.nim
+++ b/waku/rest_api/endpoint/admin/handlers.nim
@@ -344,7 +344,7 @@ proc installAdminV1GetPeersHandler(router: var RestRouter, node: WakuNode) =
       for ps in relayPeers:
         totalRelayPeers += ps.peers.len
         stat[$ps.shard] = ps.peers.len
-      stat["Total relay peers"] = relayPeers.len
+      stat["Total relay peers"] = totalRelayPeers
       stat
 
     # stats of mesh peers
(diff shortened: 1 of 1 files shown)


NagyZoltanPeter (approved):
LGTM! Can you please check why interop test fails for this PR while it was not with your previous?

fcecin:
> LGTM! Can you please check why interop test fails for this PR while it was not with your previous?

Good idea, will do.

EDIT: Just re-ran the job and now it passes. Last time this happened to me was legit (the symptom is the same when the waku binary is shot and just insta-crashes) but this has to be a flake here since this PR changes "nothing" (almost).

Ivansete-status (approved):

PR #3816, opened 2026-04-16
fix: prefer --num-shards-in-network over preset (#3795)

fcecin:
## Description

In applyNetworkConf, the AutoSharding branch unconditionally applied:

```
builder.numShardsInCluster = some(networkConf.shardingConf.numShardsInCluster)
```

... which ignore any value the user set via withNumShardsInCluster, so--num-shards-in-network=N (N > 0) was overwritten by the network preset.

Now if in AutoSharding, check builder.numShardsInCluster.isSome() first, and if that is set, keep it and warn that the preset value is overriden. Otherwise, fill it from networkConf.shardingConf.numShardsInCluster.

Added two regression tests to cover --num-shards-in-network being set to zero and to non-zero. If it is zero, the override is not applied: cli_args.nim leaves builder.numShardsInCluster as none in both the explicit flag = 0 case and flag omitted case (zero is the default value), so applyNetworkConf can't tell them apart. We don't fix this because it's not a bug; overriding an AutoSharding preset with StaticSharding has no meaningful use case.

## Changes

* fill numShardsInCluster from preset when builder slot is none
* add regression tests

## Issue

closes #3795

### DIFF
--- a/tests/factory/test_waku_conf.nim
+++ b/tests/factory/test_waku_conf.nim
@@ -213,6 +213,54 @@ suite "Waku Conf - build with cluster conf":
       check rlnRelayConf.epochSizeSec == networkConf.rlnEpochSizeSec
       check rlnRelayConf.userMessageLimit == userMessageLimit.uint
 
+  test "num-shards-in-network > 0 overrides preset":
+    ## Setup
+    let networkConf = NetworkConf.LogosDevConf()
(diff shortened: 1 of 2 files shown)


NagyZoltanPeter (approved):
LGTM!

Ivansete-status (approved):

PR #3817, opened 2026-04-17
Fix redundant start/stop calls (#3785)

fcecin:
## Description

Fixes life-cycle (start/stop) issues with node, wakurelay, gossipsub, and other protos.

## Changes

* remove redundant proto start/stop calls from node start/stop
* fix WakuRelay start/stop not overriding GossipSub start/stop (remove {.base.})
* fix most duplicate "relay already started" warnings
* delete startRelay
* add reconnectRelayPeers
* wire custom handling to kernel API relay start

## Issue

closes #3785

### DIFF
--- a/waku/node/waku_node.nim
+++ b/waku/node/waku_node.nim
@@ -369,37 +369,16 @@ proc mountStoreSync*(
 
   return ok()
 
-proc startRelay*(node: WakuNode) {.async.} =
-  ## Setup and start relay protocol
-  ## This proc has two roles
-  ## 1. Reconnect relay peer connections at the switch.
-  ## 2. In case the switch is started and *then* the relay protocol is mounted, this proc
-  ##    catches that and starts it late (the switch won't do that because it has already started).
-  ##    Relevant for apps or tests that have that workflow, for whatever reason.
-
-  info "starting relay protocol"
-
+proc reconnectRelayPeers*(node: WakuNode) {.async.} =
+  ## Reconnect to previously-seen WakuRelay peers.
   if node.wakuRelay.isNil():
-    error "Failed to start relay. Not mounted."
     return
-
-  ## Setup relay protocol
-
-  # Resume previous relay connections
-  if node.peerManager.switch.peerStore.hasPeers(protocolMatcher(WakuRelayCodec)):
-    info "Found previous WakuRelay peers. Reconnecting."
-
-    # Reconnect to previous relay peers. This will respect a backoff period, if necessary
-    let backoffPeriod =
-      node.wakuRelay.parameters.pruneBackoff + chronos.seconds(BackoffSlackTime)
-
-    await node.peerManager.reconnectPeers(WakuRelayCodec, backoffPeriod)
-
-  if node.started:
-    # if switch.start() already finished before this wakuRelay was mounted, then start the relay now.
-    await node.wakuRelay.start()
-
-  info "relay started successfully"
+  if not node.peerManager.switch.peerStore.hasPeers(protocolMatcher(WakuRelayCodec)):
+    return
+  info "Found previous WakuRelay peers. Reconnecting."
+  let backoffPeriod =
+    node.wakuRelay.parameters.pruneBackoff + chronos.seconds(BackoffSlackTime)
+  await node.peerManager.reconnectPeers(WakuRelayCodec, backoffPeriod)
 
 proc selectRandomPeers*(peers: seq[PeerId], numRandomPeers: int): seq[PeerId] =
   var randomPeers = peers
@@ -607,8 +586,7 @@ proc start*(node: WakuNode) {.async.} =
   await node.switch.start()
 
   # After switch.start, run custom Logos Delivery relay start logic
-  if not node.wakuRelay.isNil():
-    await node.startRelay()
+  await node.reconnectRelayPeers()
 
   node.started = true
 
--- a/waku/node/waku_node.nim
+++ b/waku/node/waku_node.nim
@@ -371,6 +371,12 @@ proc mountStoreSync*(
 
 proc startRelay*(node: WakuNode) {.async.} =
   ## Setup and start relay protocol
+  ## This proc has two roles
+  ## 1. Reconnect relay peer connections at the switch.
+  ## 2. In case the switch is started and *then* the relay protocol is mounted, this proc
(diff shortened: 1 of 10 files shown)


waku/node/waku_node.nim
```diff
@@ -389,8 +395,9 @@ proc startRelay*(node: WakuNode) {.async.} =
 
     await node.peerManager.reconnectPeers(WakuRelayCodec, backoffPeriod)
 
-  # Start the WakuRelay protocol
-  await node.wakuRelay.start()
+  if node.started:
+    # if switch.start() already finished before this wakuRelay was mounted, then start the relay now.
+    await node.wakuRelay.start()
```
Ivansete-status: Sorry I don't fully get this logic.
Also, it seems a good opportunity to make the `startRelay` proc to return a `Result`.
```suggestion
  if not node.started:
    return err("the node should be started when mounting relay")
```

waku/node/waku_node.nim
```diff
@@ -430,7 +437,10 @@ proc mountRendezvous*(
     return
 
   if node.started:
-    await node.wakuRendezvous.start()
+    try:
+      await node.wakuRendezvous.start()
+    except CancelledError as exc:
+      error "failed to start wakuRendezvous", error = exc.msg
```
Ivansete-status, line 422: Why we need to catch now?

waku/node/waku_node.nim
```diff
@@ -612,8 +603,13 @@ proc start*(node: WakuNode) {.async.} =
   node.switch.peerInfo.addressMappers.add(addressMapper)
 
   ## The switch will update addresses after start using the addressMapper
+  ## NOTE: This will dispatch gossipsub start to the WakuRelay.start method override
   await node.switch.start()
 
+  # After switch.start, run custom Logos Delivery relay start logic
+  if not node.wakuRelay.isNil():
+    await node.startRelay()
```
Ivansete-status: For some reason I expect that the `waku_relay`'s `start` proc will get eventually called by the Switch.
Maybe this snippet is not needed and neither needed the `startRelay` proc. I mean, we can have the `startRelay` logic within `waku_relay`'s `start` proc.

waku/waku_relay/protocol.nim
```diff
@@ -515,12 +515,12 @@ proc topicsHealthLoop(w: WakuRelay) {.async.} =
     # safety cooldown to protect from edge cases
     await sleepAsync(100.milliseconds)
 
-method start*(w: WakuRelay) {.async, base.} =
+method start*(w: WakuRelay) {.async: (raises: [CancelledError]).} =
```
Ivansete-status, line 520: I believe is better to prevent it from raising anything, if possible.
And, in case of capturing exception, then update the node's health monitor appropriately.
Applies elsewhere :)

Ivansete-status (commented):
Thanks for the PR! 🙌 
Some nitpick comments so far that I hope you find useful.

waku/node/waku_node.nim
fcecin: Just deleted startRelay and wrote a reconnectRelayPeers instead, which doesn't need to have this "if."

waku/node/waku_node.nim
fcecin: ̶R̶e̶v̶e̶r̶t̶e̶d̶;̶ ̶n̶o̶w̶ ̶W̶a̶k̶u̶R̶e̶n̶d̶e̶z̶V̶o̶u̶s̶.̶s̶t̶a̶r̶t̶ ̶i̶s̶ ̶r̶a̶i̶s̶e̶s̶:̶ ̶[̶]̶ ̶i̶n̶s̶t̶e̶a̶d̶.̶ Actually can't do this: under {.async.}, a method override that narrows raises silently drops out of the parent's dispatch path, that is, the same failure mode as putting {.base.} in the derived method (got test failures for that). So WakuRendezVous.start now matches LPProtocol.start's raises: [CancelledError], which means mountRendezvous, which is declared {.async: (raises: []).} and calls WakuRendezVous.start, has to catch and log exceptions from the new start that matches the async exception signature from LPProtocol.start (which is what we want). Not ideal, but the alternative is reverting to not making WakuRendezVous.start a proper method override at all, which is the same failure mode the original code had on WakuRelay.start with {.base.}.

waku/node/waku_node.nim
fcecin: Same as above, I just deleted startRelay. The logic that was there for reconnection is moved to reconnectRelayPeers and is called explicitly where required (which is better than trying to wire the PeerManager into the WakuRelay, which would be weird and/or impossible).

waku/waku_relay/protocol.nim
fcecin: ̶D̶e̶l̶e̶t̶e̶d̶ ̶t̶h̶e̶ ̶e̶x̶c̶e̶p̶t̶i̶o̶n̶ ̶r̶a̶i̶s̶e̶,̶ ̶c̶a̶u̶g̶h̶t̶ ̶i̶n̶t̶e̶r̶n̶a̶l̶l̶y̶ ̶n̶o̶w̶.̶ Can't do this; tried narrowing raises to [] on all four methods (WakuRelay.start, WakuRendezVous.start, SyncReconciliation.start, SyncTransfer.start) and it broke method dispatch under {.async.}, same as above.

For now logging ERR. Monitoring should be improved to programmatically track e.g. all ERR emitted everywhere. This is part of what we were discussing a while back about improving monitoring in general. All these "never happens" ERR log sites could be measured and we could show a tally of the most frequent per call site (line that prints it), per module, per time period, etc. Could be worth filing as a new feat upcoming issue to do.

waku/waku_store_sync/reconciliation.nim
```diff
@@ -484,13 +484,15 @@ proc start*(self: SyncReconciliation) =
 
   info "Store Sync Reconciliation protocol started"
 
-proc stop*(self: SyncReconciliation) =
+method stop*(self: SyncReconciliation) {.async: (raises: []).} =
+  self.started = false
```
Ivansete-status: Maybe better to move this to the end?
```suggestion
  defer:
    self.started = false
```

Ivansete-status (approved):
LGTM! Thanks for it! 🙌

NagyZoltanPeter (approved):
I agree, thanks for this!

PR #3818, opened 2026-04-17
Improve logging of content topic on server

### LINKED ISSUE #3811: add log with content_topic value on each protocol interaction
We need better logs to debug content_topic distribution across protocols.

fcecin:
## Description

Improve logging of content topics.

This should be sufficient on the server-side, but if there's a need to log more, we can add new logging in this PR or in follow-ups.

## Changes

Add content topic field logging to existing server-side debug/info log statements.

## Issue

closes #3811


Ivansete-status (approved):
LGTM! Thanks for it! 🙌

PR #3823, opened 2026-04-21
fix: relay validator registration and sync filter

fcecin:
## Description

Fixes two issues raised by Joo on the `delivery-studio` Discord channel (thanks!)

* The instance handed to gossipsub is not the one stored in topicValidator -- in waku_relay/protocol.nim
* Should be "if rangeType == RangeType.Skip:" -- was "!=" in waku_store_sync/reconciliation.nim

## Changes

* reuse stored validator in relay
* fix skip check in store sync
* increase sync tolerance in test (matches similar test)

## Issue

Maintenance Y2026H1 #3686

### DIFF
--- a/tests/waku_store_sync/test_protocol.nim
+++ b/tests/waku_store_sync/test_protocol.nim
@@ -372,7 +372,7 @@ suite "Waku Sync: reconciliation":
     const
       msgCount = 400_000
       diffCount = 100_000
-      tol = 1000
+      tol = 10_000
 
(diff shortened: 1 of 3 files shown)


Ivansete-status (approved):
LGTM! Thanks!
I believe the PR is correct but, just to give a quick double-check @SionoiS , kindly tell if the "skip" change doesn't sound right to you.

SionoiS:
I can confirm it was my mistake. Good catch! Thanks for the fix.

PR #3826, opened 2026-04-22
Chore/make nix build phase configurable

osmaczko:
## Description

nix improvements needed by chat cli client: https://github.com/logos-messaging/libchat/pull/91

## Changes

[nix: parameterize build flags with named args](https://github.com/logos-messaging/logos-delivery/commit/f8e7b2d92df70b0dbba0371c3753e245d5a413ee) 

Expose `enablePostgres`, `enableNimDebugDlOpen`, and `chroniclesLogLevel`
as arguments on `nix/default.nix`. Defaults preserve today's hardcoded
behavior, so `nix build .#liblogosdelivery` with no overrides is a
no-op change.

Consume the package via `callPackage` in `flake.nix` so consumers can
use `.override { ... }` without extra wrapping.

[nix: link libstdc++ on Linux so consumers don't need patchelf](https://github.com/logos-messaging/logos-delivery/commit/6d3e15fb94e97063552760bbf60c8549d3716d2c) 

Append `stdenv.cc.cc.lib` to `buildInputs` on Linux and add `-lstdc++`
to the Nim `--passL` flags. Nix stdenv's fixupPhase will auto-inject
`${stdenv.cc.cc.lib}/lib` into the output's RUNPATH, so downstream
consumers can drop their patchelf step.

macOS resolves the C++ stdlib via dyld/libc++ and is unaffected.

[nix: bundle librln into the output for a self-contained package](https://github.com/logos-messaging/logos-delivery/commit/7b81122b0d94a4821465fa548a1defd471c2cc2a) 

Copy the librln shared library (`librln.so` / `librln.dylib`) from the
zerokit input into `$out/lib` and rewrite the internal reference in
`liblogosdelivery`:

- Darwin: set librln's install name to `@rpath/librln.dylib`, change the
  consumer's reference to match, and add `@loader_path` as an rpath.
- Linux: add `$ORIGIN` to the rpath so `librln.so` resolves from the
  sibling directory, preserving the gcc-lib entry injected by the stdenv
  fixupPhase for libstdc++.

The installed `liblogosdelivery` no longer carries a `/nix/store/...`
absolute path to zerokit, so downstream consumers can ship the bundle
as-is.

## Issue

closes #

### DIFF
--- a/nix/default.nix
+++ b/nix/default.nix
@@ -46,7 +46,8 @@ pkgs.stdenv.mkDerivation {
     which
   ] ++ pkgs.lib.optionals pkgs.stdenv.isDarwin [ pkgs.darwin.cctools ];
 
-  buildInputs = [ zerokitRln ];
+  buildInputs = [ zerokitRln ]
+    ++ pkgs.lib.optionals pkgs.stdenv.isLinux [ pkgs.stdenv.cc.cc.lib ];
 
   buildPhase = ''
     export HOME=$TMPDIR
@@ -74,7 +75,7 @@ pkgs.stdenv.mkDerivation {
       ${pathArgs} \
       --path:$NAT_TRAV \
       --path:$NAT_TRAV/src \
-      --passL:"-L${zerokitRln}/lib -lrln" \
+      --passL:"-L${zerokitRln}/lib -lrln${pkgs.lib.optionalString pkgs.stdenv.isLinux " -lstdc++"}" \
       ${nimDefineArgs} \
       --out:build/liblogosdelivery.${libExt} \
       --app:lib \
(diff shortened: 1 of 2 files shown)


nix/default.nix
```diff
@@ -1,8 +1,22 @@
-{ pkgs, src, zerokitRln }:
+{ pkgs
+, src
+, zerokitRln
+, enablePostgres       ? true
+, enableNimDebugDlOpen ? true
+, chroniclesLogLevel   ? null
```
Ivansete-status, line 6: I'd set a good default
```suggestion
, chroniclesLogLevel   ? "INFO"
```

Ivansete-status (approved):
LGTM! Thanks for it! 🙌

NagyZoltanPeter (approved):
Great one!

PR #3822, opened 2026-04-20
feat: override ports with values supplied via env vars

fcecin:
## Description

Allows overriding any port value in the config with an (optional) environment-variable-supplied value.

NOTE: This just changes the base configured value. Transforms like --ports-shift still apply on top of it, just like it does with default values or explicit port values given via switches.

## Changes

* Add checks for environment variable overrides for each configurable port
* Add tests
* Update README.md with instructions

## Issue

closes #3819

Should also close: https://github.com/logos-co/logos-delivery-module/issues/18

Issue #3819, opened 2026-04-20
Allow port selection automatic with config to 0 


Ivansete-status:
### Background
This is to link the original issue:
- https://github.com/logos-co/logos-delivery-module/issues/18

NagyZoltanPeter:
TLDR discussion is:

- Allow client specify port as 0 -> auto select free port
- preset default to all ports to 0 (allow override from config)
- Currently `tcpPort = 0` is invalid option. We must enable this for all port configs.

Ivansete-status:
This is blocked by https://github.com/logos-messaging/logos-delivery/issues/3845

PR #3828, opened 2026-04-23
feat: allow a port value of zero for service ports (auto-assign port)

fcecin:
## Description

This PR adds the capability of specifying a port value of zero (0) for any service port, which results in the choosing of a free port, automatically, for that service.

NOTE: conf builder defaults remain concrete (60000 for p2pTcp, 9000 for discv5 UDP, 8000 for WebSocket, 8645 for REST, 8008 for metrics). Auto-port is an opt-in capability: callers pass Port(0) explicitly to get auto-bind.

For TCP, WebSocket, and REST the bind logic was prepared to work with a port of 0. For discv5 and metrics, port 0 can't be passed through directly because eth discv5 and nim-metrics don't echo the actually bound port back, so we add a retry loop on our side.

`setupDiscoveryV5` is now stricter: it returns Result and errs on udpPort==0. If config port is 0, must go through the new `setupAndStartDiscv5`.

Once any service has bound, its real port is written back into the running `WakuConf` ref and into a new `BoundPorts` struct on `WakuNode`. The port value in `BoundPorts` for a service is reported as 0 if the service is disabled.

The ENR rebuild (`updateWaku`) now runs after discv5 startup so the advertised multiaddresses reflect the actually-bound UDP port, not whatever was in the conf before (potentially "port=0").

The Messaging Debug API also gains a `MyBoundPorts` info item exposing the same bound port values as a JSON string for library consumers.

## Changes

* any port set to 0 on conf results in a random port bound
* Debug API MyBoundPorts reports actually bound ports for all services, reports 0 if disabled
* write back bound values to both WakuConf and WakuNode.ports
* setupDiscoveryV5 returns Result and errors out on port 0
* rename setupAndStartDiscv5WithAutoPort to setupAndStartDiscv5
* updateWaku ENR rebuild now runs after discv5 startup
* Add DefaultP2pTcpPort, DefaultDiscv5UdpPort, DefaultWebSocketPort, DefaultRestPort, DefaultMetricsHttpPort
* add tests

## Superseded PRs

closes #3822

## Issue

iterates https://github.com/logos-co/logos-delivery-module/issues/18

iterates #3819


tools/confutils/cli_args.nim
NagyZoltanPeter, line 1: I would leave the good defaults here. Due these are a kind of commonly known ports of a delivery node.
But yes we should allow the user to set them to 0 or even add a new switch option like `useRandomPorts`.

WDYT?
cc: @igor-sirotin , @Ivansete-status

fcecin: I have no particular preference on any of the possible semantics here. Whatever is decided here I'll just push to the PR.

NagyZoltanPeter: I mean we have good default as for popular games back in time, everyone know which port to open on firewall :-). So it is predictable in the 98% of the cases... we try tackle those 2% now.

Also I would include into the PR that logos.dev/twn/ or any preset should default to port = 0s with use override possible. Hm?

tests/factory/test_node_factory.nim
```diff
@@ -68,5 +73,143 @@ asynctest "Start a node based on default test configuration":
...
+    proc buildMetricsConf(port: Port): MetricsServerConf =
+      var b = MetricsServerConfBuilder.init()
+      b.withEnabled(true)
+      b.withHttpPort(port)
+      b.build().value.get()
+
+    try:
+      let failRes =
+        await startMetricsServerAndLogging(buildMetricsConf(takenPort), 0'u16)
+      check failRes.isErr
```
Ivansete-status: Super nitpick sorry. Applies elsewhere.
```suggestion
      check failRes.isErr()
```

tests/factory/test_node_factory.nim
```diff
@@ -68,5 +73,143 @@ asynctest "Start a node based on default test configuration":
...
+
+      let okRes = await startMetricsServerAndLogging(buildMetricsConf(freePort), 0'u16)
+      check okRes.isOk
+      if okRes.isOk:
+        await okRes.get().server.close()
+    finally:
+      taken.stop()
+      await taken.closeWait()
```
Ivansete-status: Maybe we can have these snippets in `defer:` section. Another option could be having a `tearDown` (and `setup`) section for all tests.

tests/factory/test_node_factory.nim
```diff
@@ -68,5 +73,143 @@ asynctest "Start a node based on default test configuration":
...
+    let nodeKey = generateSecp256k1Key()
+    let node = newTestWakuNode(nodeKey, parseIpAddress("0.0.0.0"), Port(0))
+    await node.start()
+
+    try:
+      var mb = MetricsServerConfBuilder.init()
+      mb.withEnabled(true)
+      mb.withHttpPort(0'u16)
+      let metricsRes = await startMetricsServerAndLogging(mb.build().value.get(), 0'u16)
+      check metricsRes.isErr
```
Ivansete-status: What is being validated in this test?

tests/waku_discv5/test_waku_discv5.nim
```diff
@@ -506,7 +506,8 @@ suite "Waku Discovery v5":
         waku.conf.nodeKey,
         waku.conf.endpointConf.p2pListenAddress,
         waku.conf.portsShift,
-      )
+      ).valueOr:
+        raiseAssert error
```
Ivansete-status: I think is interesting to give some more context.
```suggestion
        assert false, "failed setup discv5 in test: " & $error
```

waku/node/waku_node.nim
```diff
@@ -96,11 +96,19 @@ const WakuNodeVersionString* = "version / git commit hash: " & git_version
 
 # key and crypto modules different
 type
+  BoundPorts* = object ## Set by the factory once each service has bound to a port.
```
Ivansete-status: Let's better have this in a dedicated module so that we keep waku_node "simple" :)
Also, important to have `init` proc, considering that it is an `object` type. 

```suggestion
  BoundPorts* {.requiresInit.} = object ## Set by the factory once each service has bound to a port.
```

waku/node/waku_node.nim
```diff
@@ -244,11 +253,13 @@ proc info*(node: WakuNode): WakuInfo =
   let peerInfo = node.switch.peerInfo
 
   var listenStr: seq[string]
-  for address in node.announcedAddresses:
+  # Post-bind: when a transport was given port=0, this reflects the real
+  # OS-assigned port rather than the pre-bind configured 0.
+  for address in peerInfo.listenAddrs:
```
Ivansete-status: It seems it is better to use `node.announcedAddresses` so that we make sure we use routable addresses, and avoid things like `0.0.0.0`.

tools/confutils/cli_args.nim
Ivansete-status: Even though that represents a breaking change I think this is the simplest option. Then, if a node operator is aiming to configure a firewall, etc, then s/he would explicitly set the ports of interest (as we do in our fleets.)

I suggest to make the description clear that `0 means auto-port lookup`, or sth like that.

waku/common/auto_port.nim
```diff
@@ -0,0 +1,13 @@
+{.push raises: [].}
```
Ivansete-status: That's nice but we need to avoid the use of `common` (or utils) as much as possible.
It seems we can create a new folder `waku/net`, f.e., and add this module there.
(Also, `net_config.nim` can go into that folder afterwards, and maybe other modules in next PRs.)

waku/common/auto_port.nim
```diff
@@ -0,0 +1,13 @@
...
+
+import std/random
+
+const AutoPortRetryCount* = 20
+
+var
+  autoPortMin* = 50000'u16
+  autoPortMax* = 59000'u16
```
Ivansete-status: Shouldn't these be private and `const`?
```suggestion
 const AutoPortMin = 50000'u16
 const AutoPortMax = 59000'u16
```
I mentioned somewhere else but I can't quite get why this test is needed: `"exhausted retries err for metrics and discv5"`

waku/discovery/waku_discv5.nim
```diff
@@ -441,10 +447,52 @@ proc setupDiscoveryV5*(
     autoupdateRecord: conf.enrAutoUpdate,
   )
 
-  WakuDiscoveryV5.new(
-    rng, discv5Conf, some(myENR), some(nodePeerManager), nodeTopicSubscriptionQueue
+  ok(
```
Ivansete-status: nitpick sorry
```suggestion
  return ok(
```

waku/discovery/waku_discv5.nim
```diff
@@ -409,7 +409,13 @@ proc setupDiscoveryV5*(
     key: crypto.PrivateKey,
     p2pListenAddress: IpAddress,
     portsShift: uint16,
-): WakuDiscoveryV5 =
+): Result[WakuDiscoveryV5, string] =
```
Ivansete-status, line 412: Awesome movement 🥳

waku/discovery/waku_discv5.nim
```diff
@@ -441,10 +447,52 @@ proc setupDiscoveryV5*(
...
+
+  for attempt in 1 .. attempts:
+    if autoMode:
+      c.udpPort = Port(getAutoPort())
+
+    let wd = setupDiscoveryV5(
+      myENR, nodePeerManager, nodeTopicSubscriptionQueue, c, dynamicBootstrapNodes, rng,
+      key, p2pListenAddress, portsShift,
+    ).valueOr:
+      return err(error)
```
Ivansete-status, line 480: To give some more context.
```suggestion
      return err("error in setupAndStartDiscv5: " & error)
```

waku/discovery/waku_discv5.nim
```diff
@@ -441,10 +447,52 @@ proc setupDiscoveryV5*(
...
+
+    let startRes = await wd.start()
+    if startRes.isOk():
+      return ok(wd)
+    lastErr = startRes.error
+
+  if autoMode:
+    return err("discv5: auto-port bind exhausted; last error: " & lastErr)
+  return err(lastErr)
```
Ivansete-status:
```suggestion
  return err("final err line in setupAndStartDiscv5: " & lastErr)
```

waku/discovery/waku_discv5.nim
```diff
@@ -441,10 +447,52 @@ proc setupDiscoveryV5*(
...
+    conf: Discv5Conf,
+    dynamicBootstrapNodes: seq[RemotePeerInfo],
+    rng: ref HmacDrbgContext,
+    key: crypto.PrivateKey,
+    p2pListenAddress: IpAddress,
+    portsShift: uint16,
+): Future[Result[WakuDiscoveryV5, string]] {.async: (raises: []).} =
+  ## Construct and start a `WakuDiscoveryV5` instance, handling auto-port
+  ## retry when the caller asks for `udpPort == 0`.
+  var c = conf
```
Ivansete-status: I'm curious why this is needed, just to learn something :)

waku/factory/conf_builder/discv5_conf_builder.nim
```diff
@@ -57,7 +60,7 @@ proc build*(b: Discv5ConfBuilder): Result[Option[Discv5Conf], string] =
         bucketIpLimit: b.bucketIpLimit.get(2),
         enrAutoUpdate: b.enrAutoUpdate.get(true),
         tableIpLimit: b.tableIpLimit.get(10),
-        udpPort: b.udpPort.get(9000.Port),
+        udpPort: b.udpPort.get(),
```
Ivansete-status: Maybe to make it a bit more explicit?
```suggestion
        udpPort: b.udpPort.get(0),
```

waku/factory/conf_builder/metrics_server_conf_builder.nim
```diff
@@ -36,11 +36,14 @@ proc build*(b: MetricsServerConfBuilder): Result[Option[MetricsServerConf], stri
...
 
+  if b.httpPort.isNone():
+    return err("metricsServer.httpPort is not specified")
+
   return ok(
     some(
       MetricsServerConf(
         httpAddress: b.httpAddress.get(static parseIpAddress("127.0.0.1")),
-        httpPort: b.httpPort.get(8008.Port),
+        httpPort: b.httpPort.get(),
```
Ivansete-status:
```suggestion
        httpPort: b.httpPort.get(0),
```

waku/factory/conf_builder/waku_conf_builder.nim
```diff
@@ -574,12 +576,9 @@ proc build*(
...
 
-  let p2pTcpPort =
-    if builder.p2pTcpPort.isSome():
-      builder.p2pTcpPort.get()
-    else:
-      warn "P2P Listening TCP Port is not specified, listening on 60000"
-      60000.Port
+  if builder.p2pTcpPort.isNone():
+    return err("p2pTcpPort is not specified")
+  let p2pTcpPort = builder.p2pTcpPort.get()
```
Ivansete-status: Perhaps?
```suggestion
  let p2pTcpPort = builder.p2pTcpPort.get(0)
```

waku/node/waku_metrics.nim
```diff
@@ -57,27 +57,44 @@ proc startMetricsLog*() =
...
-  try:
-    await server.start()
-  except CatchableError:
-    return err("metrics HTTP server start failed: " & getCurrentExceptionMsg())
-
-  info "Metrics HTTP server started", serverIp = $serverIp, serverPort = $serverPort
-  return ok(server)
+): Future[Result[StartedMetricsServer, string]] {.async.} =
+  let autoMode = serverPort == Port(0)
+  let attempts = if autoMode: AutoPortRetryCount else: 1
```
Ivansete-status: It seems the algorithm in this proc for trial, while being great, it seems is a bit duplicated. Maybe we can create a template in `auto_port.nim`.

waku/rest_api/endpoint/debug/types.nim
```diff
@@ -18,10 +19,45 @@ proc toDebugWakuInfo*(nodeInfo: WakuInfo): DebugWakuInfo =
     listenAddresses: nodeInfo.listenAddresses,
     enrUri: some(nodeInfo.enrUri),
     mixPubKey: nodeInfo.mixPubKey,
+    ports: nodeInfo.ports,
   )
 
 #### Serialization and deserialization
 
+proc writeValue*(
```
Ivansete-status: That's great. It would be interesting if we could have the chance to expose this information into the `Messaging Debug API`, if not done already ;P

Ivansete-status (commented):
Some comments so far, thanks for the PR!
The only big concern is the change in the loop, in `waku/node/waku_node.nim‎`.
Cheers

tools/confutils/cli_args.nim
fcecin: I have pushed a commit now that changes the port=0 feature to only apply when an explicit `--preset` is given via the CLI. If we decide that that's not what we want, I'll revert it.

fcecin: > Even though that represents a breaking change I think this is the simplest option. Then, if a node operator is aiming to configure a firewall, etc, then s/he would explicitly set the ports of interest (as we do in our fleets.)
> 
> I suggest to make the description clear that `0 means auto-port lookup`, or sth like that.

Whoops, I just pushed a "applies only to --preset" commit. Should I revert it?

Ivansete-status: haha no worries @fcecin 
I suggest to revert it, yes.
That kind of internal conditions look nice now but we will forget in the future so the more explicit we are with the defaults, the better. I mean, I think config params should be independent, and the `preset` one is a weak helper that should be overridden by any other config param.
EDITED: `preset` is convenient because it sets good defaults but any of those "good defaults" should get overridden by any explicit config set by the user.

tests/wakunode2/test_cli_args.nim
```diff
@@ -139,6 +139,64 @@ suite "Waku external config - apply preset":
...
+
+    check wakuConf.endpointConf.p2pTcpPort == Port(0)
+    check wakuConf.discv5Conf.isSome()
+    check wakuConf.discv5Conf.get().udpPort == Port(0)
+    check wakuConf.metricsServerConf.isSome()
+    check wakuConf.metricsServerConf.get().httpPort == Port(0)
+    check wakuConf.restServerConf.isSome()
+    check wakuConf.restServerConf.get().port == Port(0)
+    check wakuConf.webSocketConf.isSome()
+    check wakuConf.webSocketConf.get().port == Port(0)
```
NagyZoltanPeter: I'm not pretty sure if this is ok.
Given that explicit config says `--tcp-port=23456`, I suspect the user wants it to be there.
How I see, we had good defaults so far, now the default should be 0 = auto select, but when I configure it explicitly the preset overrides?
How should the user than predict the port?

I think preset should only default to 0 autoselect if not other that good default (if we leave it).
We can think of it as an optional config (if not set we use autoselect when preset is also used).

Also this is even worse in case of rest-port... in order to use rest api I should gather the bound port from logs or how?
Same for metrics port, I see it as a node operator feature, so unlikely normal devs/users will handle it. Having it random makes Prometheus config a nightmare, imo.

Ivansete-status: > How I see, we had good defaults so far, now the default should be 0 = auto select, but when I configure it explicitly the preset overrides?

Yes, `0` performs auto and any explicit port config should override it.

----
> How should the user than predict the port?

A desktop or mobile user doesn's need to specify the port.
Only server node operators will explicitly take care of the ports that need to be accessible and hence they will explicitly set values to the ports of interest.

----

> Also this is even worse in case of rest-port... in order to use rest api I should gather the bound port from logs or how?

Yes, also by checking debug API results.

----
> Same for metrics port, I see it as a node operator feature, so unlikely normal devs/users will handle it. Having it random makes Prometheus config a nightmare, imo.

For the case of metrics, I think node operators should lookup which port to configure anyways.

fcecin: > I'm not pretty sure if this is ok. Given that explicit config says `--tcp-port=23456`, I suspect the user wants it to be there. How I see, we had good defaults so far, now the default should be 0 = auto select, but when I configure it explicitly the preset overrides? How should the user than predict the port?

Sorry, this "--preset" feature was completely broken. I just reverted it back to "port 0 is the default for everything" which is what we ended up deciding to do.

> Also this is even worse in case of rest-port... in order to use rest api I should gather the bound port from logs or how? Same for metrics port, I see it as a node operator feature, so unlikely normal devs/users will handle it. Having it random makes Prometheus config a nightmare, imo.

It's not usual for servers to auto-bind ports, but if we're going to go down that road, we'll have to deal with the fallout.

I'm thinking that if logos-delivery is linked as a library in the app, the user should call LMAPI procs to ask for the effective port configs (e.g. support arbitrary REST queries via LMAPI). It's the same as passing a config that has specific ports, but if the app doesn't care about the ports then it can just ask after the fact.

tests/factory/test_node_factory.nim
fcecin: Deleted

waku/node/waku_node.nim
fcecin: Oooff! Fixed

waku/discovery/waku_discv5.nim
fcecin: I didn't want to make the `conf` proc param a `var`. It's easier (no need to write it back later) but I'd rather make changes to config objects explicit.

fcecin:
I'm thinking here that maybe (app users == library users) and (node operators == CLI users). If that holds, then we could restore the previous non-zero port defaults to wakunode2 (or any other standalone "node" / CLI binary that front-ends the library for operators) and leave the zero/auto-port anti-port-collision feature for library use only. If so, we can also remove the REST changes in this PR.

NagyZoltanPeter:
> I'm thinking here that maybe (app users == library users) and (node operators == CLI users). If that holds, then we could restore the previous non-zero port defaults to wakunode2 (or any other standalone "node" / CLI binary that front-ends the library for operators) and leave the zero/auto-port anti-port-collision feature for library use only. If so, we can also remove the REST changes in this PR.

Formerly I was suggesting a third `mode` such as `Fleet` or similar in compared to `Core` and `Edge`.
Such a distinction would allow us to offer easy config for everyone with good defaults.
It would be very rare when a single operator would run two node on the same host, but application modes could easily default to auto-port selection.

WDYT?
cc: @igor-sirotin, @Ivansete-status

fcecin:
> > I'm thinking here that maybe (app users == library users) and (node operators == CLI users). If that holds, then we could restore the previous non-zero port defaults to wakunode2 (or any other standalone "node" / CLI binary that front-ends the library for operators) and leave the zero/auto-port anti-port-collision feature for library use only. If so, we can also remove the REST changes in this PR.
> 
> Formerly I was suggesting a third `mode` such as `Fleet` or similar in compared to `Core` and `Edge`. Such a distinction would allow us to offer easy config for everyone with good defaults. It would be very rare when a single operator would run two node on the same host, but application modes could easily default to auto-port selection.
> 
> WDYT? cc: @igor-sirotin, @Ivansete-status

So maybe the idea would be:

* noMode, Core, Edge: default to port 0 for service ports (auto port selection), both for the library interface and the node binary (wakunode2) as it points to application-user.
* Fleet:  apply non-zero port defaults (the current behavior), both for the library interface and the node binary (wakunode2).

So the mode selector drives the application profile (is this an application user, or is this a node operator), instead of the layer (wakunode2 vs. library). Is that it?

EDIT: I think that #3845 basically settles this discussion: `mode` for the upcoming "easy config" library API path (needed to address that new issue) is a mandatory field and it must be either "Core" or "Edge," with no noMode allowed. The equivalent for a fleet user would be to use the full-config route via the same library API or the CLI, which is what is already done. I'll think about this more but I think that's where this is going.

waku/factory/waku.nim
```diff
@@ -390,29 +396,39 @@ proc startWaku*(waku: ptr Waku): Future[Result[void, string]] {.async: (raises:
...
-    (await updateWaku(waku)).isOkOr:
-      return err("Error in updateApp: " & $error)
-  except CatchableError:
-    return err("Caught exception in updateApp: " & getCurrentExceptionMsg())
+  let bound = getPorts(waku.node.switch.peerInfo.listenAddrs).valueOr:
+    return err("failed to read bound ports from switch: " & $error)
+  if bound.tcpPort.isSome():
+    waku[].node.ports.tcp = some(bound.tcpPort.get().uint16)
+  if bound.websocketPort.isSome():
+    waku[].node.ports.webSocket = some(bound.websocketPort.get().uint16)
```
Ivansete-status: Maybe better to explicitly set port zero as default? Or perhaps not needed.
```suggestion
  waku[].node.ports.tcp = some(bound.tcpPort.get(0).uint16)
  waku[].node.ports.webSocket = some(bound.websocketPort.get(0).uint16)
```

waku/net/auto_port.nim
```diff
@@ -0,0 +1,38 @@
...
+  let autoMode = startingPort == Port(0)
+  let attempts = if autoMode: AutoPortRetryCount else: 1
+  var lastErr = ""
+  for i in 1 .. attempts:
+    let port =
+      if autoMode:
+        Port(getAutoPort())
+      else:
+        startingPort
+    let res = await attempt(port)
```
Ivansete-status: Maybe set a .withTimeout? Not a big deal though

waku/net/auto_port.nim
```diff
@@ -0,0 +1,38 @@
...
+    if res.isOk():
+      return ok(res.get())
+    lastErr = res.error
+  if autoMode:
+    return err("auto-port exhausted; last error: " & lastErr)
+  return err(lastErr)
```
Ivansete-status:
```suggestion
  return err("reached last err in tryWithAuthoPort: " & lastErr)
```

waku/net/auto_port.nim
```diff
@@ -0,0 +1,38 @@
...
+
+const
+  AutoPortRetryCount* = 20
+  AutoPortMin = 50000'u16
+  AutoPortMax = 59000'u16
+
+var rng = initRand()
+
+proc getAutoPort*(): uint16 =
+  uint16(rng.rand(AutoPortMin.int .. AutoPortMax.int))
```
Ivansete-status: Maybe?
```suggestion
proc getAutoPort*(): uint16 =
  let rng = initRand()
  uint16(rng.rand(AutoPortMin.int .. AutoPortMax.int))
```

waku/net/bound_ports.nim
```diff
@@ -0,0 +1,27 @@
...
+type BoundPorts* {.requiresInit.} = object
+  ## Set by the factory once each service has bound to a port.
+  tcp*: Option[uint16]
+  webSocket*: Option[uint16]
+  rest*: Option[uint16]
+  discv5Udp*: Option[uint16]
+  metrics*: Option[uint16]
+
+proc init*(T: type BoundPorts): BoundPorts =
+  BoundPorts(
```
Ivansete-status: Super nitpick one, sorry
```suggestion
  return BoundPorts(
```

waku/net/bound_ports.nim
```diff
@@ -0,0 +1,27 @@
...
+    discv5Udp: none(uint16),
+    metrics: none(uint16),
+  )
+
+proc toJsonString*(p: BoundPorts): string =
+  var obj = newJObject()
+  for name, value in fieldPairs(p):
+    if value.isSome():
+      obj[name] = %value.get()
+  return $obj
```
Ivansete-status:
```suggestion
  return $(%*p)
```

waku/node/waku_metrics.nim
```diff
@@ -57,27 +56,35 @@ proc startMetricsLog*() =
...
-): Future[Result[MetricsHttpServerRef, string]] {.async.} =
-  info "Starting metrics HTTP server", serverIp = $serverIp, serverPort = $serverPort
+): Future[Result[StartedMetricsServer, string]] {.async.} =
+  proc attempt(
+      port: Port
+  ): Future[Result[StartedMetricsServer, string]] {.async: (raises: []).} =
+    info "Starting metrics HTTP server", serverIp = $serverIp, serverPort = $port
+
+    let server = MetricsHttpServerRef.new($serverIp, port).valueOr:
+      return err($error)
```
Ivansete-status:
```suggestion
      return err("fail to start service metrics server, attempt:" & $error)
```

waku/node/waku_metrics.nim
```diff
@@ -57,27 +56,35 @@ proc startMetricsLog*() =
...
 
-  let server = MetricsHttpServerRef.new($serverIp, serverPort).valueOr:
-    return err("metrics HTTP server start failed: " & $error)
+    try:
+      await server.start()
+    except CatchableError:
+      return err(getCurrentExceptionMsg())
```
Ivansete-status:
```suggestion
      return err("exception while startMetricsServer, attempt: " & getCurrentExceptionMsg())
```

waku/discovery/waku_discv5.nim
```diff
@@ -441,10 +447,47 @@ proc setupDiscoveryV5*(
...
+    var c = conf
+    c.udpPort = port
+    let wd = setupDiscoveryV5(
+      myENR, nodePeerManager, nodeTopicSubscriptionQueue, c, dynamicBootstrapNodes, rng,
+      key, p2pListenAddress, portsShift,
+    ).valueOr:
+      return err(error)
+    let startRes = await wd.start()
+    if startRes.isErr():
+      return err(startRes.error)
```
Ivansete-status:
```suggestion
      return err("failed to start discovery, attempt: " & startRes.error)
```

Ivansete-status (approved):
LGTM! Thanks for it! 🙌 
Just adding some minor nitpicks that I hope you find useful.
Cheers.

waku/net/auto_port.nim
fcecin: Ha! This proc is called literally just twice, and we can drop the memory for the RNG. Nice! Must be `var` to compile, tho.

waku/factory/waku.nim
fcecin: The root issue here is that `BoundPorts` should arguably have `uint16` fields, not `Option[uint16]` as it is now. Arguably, there's no benefit from differentiating "the bound port was reported as actually zero" (to catch extremely weird/rare errors) vs. "the port was not specified or the service is disabled, so the port is reported as zero here." I'll just remove the Option wrapper and use raw uint16.

waku/net/bound_ports.nim
fcecin: By making `BoundPorts` drop the Option wrapper on the uint16 port values (see above) this change will fit well.

waku/node/waku_metrics.nim
```diff
@@ -57,27 +56,36 @@ proc startMetricsLog*() =
...
-  try:
-    await server.start()
-  except CatchableError:
-    return err("metrics HTTP server start failed: " & getCurrentExceptionMsg())
+    info "Metrics HTTP server started", serverIp = $serverIp, serverPort = $port
+    return ok((server: server, port: port))
 
-  info "Metrics HTTP server started", serverIp = $serverIp, serverPort = $serverPort
-  return ok(server)
+  let started = (await tryWithAutoPort[StartedMetricsServer](serverPort, attempt)).valueOr:
```
NagyZoltanPeter, line 81: I think metrics server needs an explicit port set. It has no mean to assign an automatic port. It will avoid collision but will make prometheus data collection impossible or inconvenient, imo.

waku/factory/conf_builder/rest_server_conf_builder.nim
```diff
@@ -56,7 +54,7 @@ proc build*(b: RestServerConfBuilder): Result[Option[RestServerConf], string] =
       RestServerConf(
         allowOrigin: b.allowOrigin,
         listenAddress: b.listenAddress.get(),
-        port: b.port.get(),
+        port: b.port.get(Port(0)),
```
NagyZoltanPeter: Same as metrics server. How inconvenient to use a rest endpoint if you need to dig out the port of it from the logs?

NagyZoltanPeter (approved):
Approve not to block. Still have the impression that for metrics and rest server - if enabled - user must assign exact port or just use the good old default, we should not allow config port 0 for them and try assign random.

fcecin:
> Approve not to block. Still have the impression that for metrics and rest server - if enabled - user must assign exact port or just use the good old default, we should not allow config port 0 for them and try assign random.

I think we can do a further refinement of this PR.

We could change this PR to support a config of "port=0" ("please, logos-delivery instance, auto-allocate a port for me") *but  we would not make the value of zero (auto-alloc) the default value of any server port parameter*. Instead, if no value is given by the app, for any logos-delivery server port parameter, then logos-delivery just keeps the existing non-zero defaults (that would apply for all cases: both for the CLI *and* the library use case).

I think that would accomplish two key things:

* *Entirely support the requirement (use case) raised by the original issue:* N components launching their own local-machine copy of `logos-delivery` don't have to know about each other's port allocations, but they absolutely _can_ know that *that's their use case*, that is, there's going to be N copies launched and they want all to avoid having to coordinate port allocation, and if so, these applications (or whatever underlying helper, harness, component-library, etc. that is built to serve them specifically) would *explicitly override, with a value of zero* the otherwise non-zero default port configuration that would continue to apply for all service ports.
*  *Does not affect any existing client:* The breakage surface of this change becomes literally zero now, since the keep the default, concrete, nonzero port value for all service ports if the application does not request auto-ports via an explicit value of zero.

As for metrics vs. p2pTcp etc., we can't just do something different for every service port. If we tried to write auto-allocation of ports just for services that are enabled by default, like p2pTcp, and do something different for metrics, etc., the problem would persist: N components launch their own local-machine copies of `logos-delivery`, and they _do_ know they want e.g. the optional metrics, etc. endpoints but they still _don't know or care_ where these ports will end up since they know there's going to be 10 attempts to allocate the same port of *any* server regardless.

If you are using `logos-delivery` as a library, you can ask for the actually-allocated ports via the API ("Debug API" surface). Now, it can be argued that that would not help if the problem is not with the app linking the library, but with _other_ processes that now can't guess these ports and would have to e.g. grep logs (or otherwise using file system scanning as rendez-vous). But the counter-argument there is that if we are living in a world where N copies of the library are being launched in the same machine by applications, then it is reasonable to expect that this orchestration would have to be ultimately carried out by the applications that are linked with the library and that are launching these N copies. Now, some unrelated monitoring tooling running in the same machine wouldn't be any better by listening to "the port" in that case since there are N copies to monitor anyway. So the app stack cannot just wish "automatic" wiring of e.g. metrics/monitoring infra by "someone else" in that case -- app would still have to own the wiring of all that up. If someone is being assigned the task to "grep logs" in that scenario to service the app, that means the app (or the app stack) is trying to shirk it's newly-minted responsibility now and that has to be pushed back.

(It's worth noting that "ports" being a scarce system-wide resource will always give heartburn to any P2P component system. P2P component systems can't live in the world of "server" metaphors where the port config is some known and pre-configured value by a centralized entity. We could instead have native <pid:port> routing or the like, achievable with a machine-wide socket/message "Connection/Message Broker" routing service -- which would actually own all the physical sockets -- and at least 10% of the reason why people use Docker for emulating P2P networks on a single machine would be gone... bottom line, there's no "100% clean" solution to this problem since the "communication port values are a system-wide resource known ahead of time per-instance" is fundamentally broken for P2P.)

I'll wait for a decision here before merging; CC: @Ivansete-status @igor-sirotin ; I think the PR in its current form is mergeable, but requiring explicit port=0 arguments from the app stack for the auto-alloc feature is a clear improvement, IMO.

waku/discovery/waku_discv5.nim
Ivansete-status, line 412: Kindly add a comment in this proc explaining that this is public only for testing purposes and the exposed proc is actually: `setupAndStartDiscv5`.

waku/factory/waku.nim
```diff
@@ -390,29 +396,37 @@ proc startWaku*(waku: ptr Waku): Future[Result[void, string]] {.async: (raises:
...
 
-    (await waku.wakuDiscV5.start()).isOkOr:
-      return err("failed to start waku discovery v5: " & $error)
+    waku[].node.ports.discv5Udp = waku[].wakuDiscV5.udpPort.uint16
+    waku[].conf.discv5Conf.get().udpPort = waku[].wakuDiscV5.udpPort
+
+  ## Update waku data that is set dynamically on node start
+  try:
+    (await updateWaku(waku)).isOkOr:
+      return err("Error in updateWaku: " & $error)
```
Ivansete-status: Sorry, nitpick one
```suggestion
      return err("Error in startWaku: " & $error)
```

waku/net/bound_ports.nim
```diff
@@ -0,0 +1,20 @@
...
+  rest*: uint16
+  discv5Udp*: uint16
+  metrics*: uint16
+
+proc init*(T: type BoundPorts): BoundPorts =
+  return BoundPorts(
+    tcp: 0'u16, webSocket: 0'u16, rest: 0'u16, discv5Udp: 0'u16, metrics: 0'u16
+  )
+
+proc toJsonString*(p: BoundPorts): string =
```
Ivansete-status:
```suggestion
proc `$`*(p: BoundPorts): string =
```

waku/node/waku_metrics.nim
Ivansete-status: Yes I agree with Zoltán.
The metrics ports will come from `metricsServerConf` , and this config is only set when using the cli wakunode2 app, i.e., not used from the library.

Ivansete-status (commented):
Adding some more comments.
We can merge but is important to leave auto-port only for library, and keep fixed ports for app-config items such as metrics service or REST service.

fcecin:
> Adding some more comments. We can merge but is important to leave auto-port only for library

Yes. So, to get that done, I will first go ahead and execute here, on this PR, my last proposal above: at the builder level, the defaults are _concrete_ nonzero port values, for _all_ service ports, which is the existing behavior for both CLI and library. That fixes my earlier interpretation that we wanted automatic port selection by default for all ports all the time, which was a broken assumption.

So this PR now mostly just adds the *capability* of automatically selecting a port for a service if a value of zero is given to it. No Port == 0 defaults here, anywhere.

In the follow-up API Consistency PR #3866, we will take advantage of what @igor-sirotin specified: that the `createNode(preset, mode, overrides ...)` entry point signals `app developer`:

| Entry point | Audience |
|-|-|
| `createNode(preset, mode, overrides?)` | Developer (Messaging API) |
| `createNode(conf: WakuNodeConf)` | Tester, advanced tooling |

... and it is inside the "createNode() for app developers" library API that we will implicitly pre-set the ports to a default value of "0", that is, automatic port selection.

That way, we won't have to try and guess the user profile (advanced vs dev) from the interface being used (CLI vs library), which doesn't work -- one of the central points of the API Consistency issue is to fix that.

Then, if the app developer still wants concrete ports, they can inform concrete ports in the `overrides` parameter. In any case, they can confirm the ports that they eventually got allocated via the library's Debug API.

This effectively "keeps auto-port only for library," in the sense that you must use one of the library APIs (maybe the only one, if we choose to eliminate `createNode(conf: WakuNodeConf)`) to get the auto-port default behavior. The CLI won't default to `0` because it creates the Waku object directly with a full conf arg, which is the same that `createNode(conf: WakuNodeConf)` API does, so it works not because it is the CLI interface or the createNode full conf API, but because all paths into creating a Waku object do that unless that specific path actively *doesn't* do that.

> , and keep fixed ports for app-config items such as metrics service or REST service.

I will make it so PR #3866 will _not_ apply a 0-port (auto-port) default for Metrics and REST via the createNode app-dev API. It will keep the default-port-0 (auto-port) for Discv5, p2p-tcp, and websocket port only. If we still get port collision problems after that, we can relax that and set 0-port for all services again, under the app-developer library API for createNode.

If I still screwed up somewhere in my interpretation here, let me know. But I *think* that's right this time and that should cover everyone's reviews.

PR #3829, opened 2026-04-24
Fix websock nimble dependency version restriction to match lock file

NagyZoltanPeter:
This updates the declared `websock` dependency in `waku.nimble` from `>= 0.2.1` to `>= 0.3.0`.

Why:
The lock file already resolves against `websock` 0.3.x, so the previous lower bound was out of sync with the actual pinned dependency set. Aligning the constraint avoids version mismatch during dependency resolution.

### DIFF
--- a/waku.nimble
+++ b/waku.nimble
@@ -33,7 +33,7 @@ requires "nim >= 2.2.4",
   "dnsdisc",
   "dnsclient",
   "httputils >= 0.4.1",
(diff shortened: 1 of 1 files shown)


fcecin (approved):

Ivansete-status (approved):

NagyZoltanPeter:
under: https://github.com/logos-messaging/logos-delivery/issues/3686

PR #3830, opened 2026-04-24
fix: restore -d:postgres in nimble task and propagate NIMFLAGS

darshankabariya:
Regression from #3798 ( nimble migration ):
- make wakunode2 target now delegates to 'nimble wakunode2'
- nimble task hard-codes only -d:chronicles_log_level=TRACE; -d:postgres was lost
- getNimParams() only read NIM_PARAMS env var, so NIMFLAGS passed via Dockerfile/Jenkinsfile.release was silently dropped

### DIFF
--- a/.github/workflows/container-image.yml
+++ b/.github/workflows/container-image.yml
@@ -83,7 +83,7 @@ jobs:
         id: build
         if: ${{ steps.secrets.outcome == 'success' }}
         run: |
-          make -j${NPROC} V=1 NIMFLAGS="-d:disableMarchNative -d:postgres -d:chronicles_colors:none" wakunode2
+          make -j${NPROC} V=1 POSTGRES=1 NIMFLAGS="-d:disableMarchNative -d:chronicles_colors:none" wakunode2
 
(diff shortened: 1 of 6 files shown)


fcecin (approved):

Ivansete-status (approved):
Good catch! Thanks for it! 🙌 
I wonder though if we should only use `NIM_PARAMS`, and adapt the Dockerfile/Makefile etc?

NagyZoltanPeter (commented):
I would rather not hardcode postgres as mandatory dependency here.
Can't we restore the functionality as optional? Due it comes with hard runtime dependency with libpq.
cc: @Ivansete-status

darshankabariya:
> I would rather not hardcode postgres as mandatory dependency here. Can't we restore the functionality as optional? Due it comes with hard runtime dependency with libpq. cc: @Ivansete-status

Thanks, Pushed an update —  dropped the nimble hardcoding and postgres is now fully opt-in ( Dockerfile defaults to off ) 

cc @Ivansete-status @NagyZoltanPeter

NagyZoltanPeter (approved):
Cool thanks!

PR #3831, opened 2026-04-24
chore: Removed duplicates of multiaddresses

NagyZoltanPeter:
During fleet node deploy with logos-delivery-module it turned out that more rigid config handling revealed formerly silently ignored option.
But using `dns4DomainName` not just force replace Ip addresses in multiaddresses but used together with `extMultiAddrs` cause duplicate announced and ext-multi addresses in registry. 
This had the consequence of failed ENR record generation - that because of the duplicate exceeded the size limit -> preventing the node to start.

This pr intends to prevent both failure case.

- removes unintended duplicates from multi-address lists.
- try create an ENR record with the most multiadrresses that fits in the size limits.

### DIFF
--- a/tests/factory/test_node_factory.nim
+++ b/tests/factory/test_node_factory.nim
@@ -1,11 +1,19 @@
 {.used.}
 
-import testutils/unittests, chronos, libp2p/protocols/connectivity/relay/relay
+import
+  std/[options, sequtils],
+  testutils/unittests,
+  chronos,
+  libp2p/multiaddress,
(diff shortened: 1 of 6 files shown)


fcecin (approved):
LGTM!

Ivansete-status (approved):
LGTM! Thanks for it! 🙌

NagyZoltanPeter:
under: https://github.com/logos-messaging/logos-delivery/issues/3686

PR #3836, opened 2026-04-28
chore: windows ci failure 

darshankabariya:
Switch to prebuilt nimble binary from nim-lang/nimble releases.
Avoids koch self-bootstrap that hits Windows file-lock errors.


darshankabariya:
Windows build CI is fixed in master, so this PR is not needed anymore.

PR #3839, opened 2026-04-29
adapt event types and values to messaging-api spec

Ivansete-status:
While revisiting the [messaging-api.md](https://github.com/logos-messaging/specs/blob/c8de3434d2cd5bacda1680399e4381c8801e8f73/standards/application/messaging-api.md) I noticed that there are some minor discrepancies regarding some event types. This PR aims to perform such small adjustments.

## Issue

### DIFF
--- a/liblogosdelivery/logos_delivery_api/node_api.nim
+++ b/liblogosdelivery/logos_delivery_api/node_api.nim
@@ -131,34 +131,34 @@ proc logosdelivery_start_node(
     ctx.myLib[].brokerCtx,
     proc(event: MessageSentEvent) {.async: (raises: []).} =
       callEventCallback(ctx, "onMessageSent"):
-        $newJsonEvent("message_sent", event),
+        $newJsonEvent("message:sent", event),
   ).valueOr:
     chronicles.error "MessageSentEvent.listen failed", err = $error
     return err("MessageSentEvent.listen failed: " & $error)
 
-  let errorListener = MessageErrorEvent.listen(
+  let errorListener = MessageSendErrorEvent.listen(
     ctx.myLib[].brokerCtx,
-    proc(event: MessageErrorEvent) {.async: (raises: []).} =
-      callEventCallback(ctx, "onMessageError"):
-        $newJsonEvent("message_error", event),
+    proc(event: MessageSendErrorEvent) {.async: (raises: []).} =
+      callEventCallback(ctx, "onMessageSendError"):
+        $newJsonEvent("message:send-error", event),
   ).valueOr:
-    chronicles.error "MessageErrorEvent.listen failed", err = $error
-    return err("MessageErrorEvent.listen failed: " & $error)
+    chronicles.error "MessageSendErrorEvent.listen failed", err = $error
+    return err("MessageSendErrorEvent.listen failed: " & $error)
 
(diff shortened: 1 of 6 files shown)


liblogosdelivery/logos_delivery_api/node_api.nim
NagyZoltanPeter, line 1: These are breaking changes for logos-delivery-module! Needs follow up and new release afterward.

liblogosdelivery/logos_delivery_api/node_api.nim
```diff
@@ -131,34 +131,34 @@ proc logosdelivery_start_node(
     ctx.myLib[].brokerCtx,
     proc(event: MessageSentEvent) {.async: (raises: []).} =
       callEventCallback(ctx, "onMessageSent"):
-        $newJsonEvent("message_sent", event),
+        $newJsonEvent("message:sent", event),
```
NagyZoltanPeter, line 134: Why the `:` is better than `_`? just curious!

waku/node/delivery_service/send_service/send_service.nim
```diff
@@ -181,7 +181,7 @@ proc reportTaskResult(self: SendService, task: DeliveryTask) =
     if not task.propagateEventEmitted:
       info "Message successfully propagated",
         requestId = task.requestId, msgHash = task.msgHash.to0xHex()
-      MessagePropagatedEvent.emit(
+      MessageSendPropagatedEvent.emit(
```
NagyZoltanPeter, line 184: I think here the spec is wrong and the simpler form `MessagePropagatedEvent` is the better.
The wording of Send here misleading... compared to 'MessageSentEvent'.
WDYT?

NagyZoltanPeter (approved):
Approve with reservations. We should not forget the follow up on delivery_module

waku/node/delivery_service/send_service/send_service.nim
fcecin: This is my fault; I should have renamed the events in the code when I cleaned up the spec _before_ the LMAPI release.

I agree with Zoltan here. We should fix the spec, not the code.

MessageSendErrorEvent is _technically_ correct, but MessageErrorEvent stuck and at the "simple send" LMAPI level, the only MessageErrorEvent _is_ the send event, so the discrimination that I added there is doing zero actual work.

Also MessageSendPropagatedEvent -> MessagePropagatedEvent with even less loss of meaning or risk of symbol collision or rot.

What do you guys think? Should we submit a fix to the spec instead?

Ivansete-status: Okay fair enough. Thanks for the comments both!
Let's keep this opened and we can adapt the spec once we fully move to lips. https://github.com/logos-co/logos-lips/pull/315

liblogosdelivery/logos_delivery_api/node_api.nim
igor-sirotin: Also looks like a breaking change 🤔

Ivansete-status:
Let's carry on the discussion in the following PR: https://github.com/logos-co/logos-lips/pull/333
( not closing this one yet as this is more visible for us .)

Ivansete-status:
Closing because in the end we adapted the spec.

PR #3843, opened 2026-04-30
avoid keeping delivery tasks in propagated state when check store is disabled

Ivansete-status:
## Issue

### DIFF
--- a/waku/node/delivery_service/send_service/send_service.nim
+++ b/waku/node/delivery_service/send_service/send_service.nim
@@ -225,9 +225,10 @@ proc evaluateAndCleanUp(self: SendService) =
       it.state != DeliveryState.FailedToDeliver
   )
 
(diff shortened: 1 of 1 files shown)


fcecin (approved):
LGTM!

Side note, worth revisiting at some point:

```
  of DeliveryState.SuccessfullyPropagated:
    # TODO: in case of unable to strore check messages shall we report success instead?
```

I _prefer_ the current behavior (i.e. `MessageSentEvent` being possibly never emitted -- that's one of the reasons why having two signals is useful here), but the wording choice of `MessageSentEvent` now doesn't look very fortunate. Would have been better to call it `MessageSendConfirmedEvent` or somesuch (an objectively worse-looking name, but more descriptive).

darshankabariya (approved):
LGTM !

PR #3849, opened 2026-05-06
rx_service: ensure fetch msgs query is performed when missing msg

Ivansete-status:
This is to fix an issue detected by @AlbertoSoutullo 🙌  where the nodes where continuously sending store requests to one store node, in the DST lab.

The problem was that, regardless having an empty list of message hashes, the `getMissingMsgsFromStore` generates a simple store query that just brings the latest messages from the database.

## Issue


fcecin (approved):
Thanks!!

NagyZoltanPeter (approved):
LGTM! good catch.

PR #3857, opened 2026-05-07
Evict peer instead of abrupt disconnect and avoid sending unnecessary store requests

Ivansete-status:
Update release/v0.38 branch

- Max relay connections can be overpassed in case on going store requests. This avoids abruptly closing a connection.
- Receive service doesn't perform store queries in case there are no missed messages.

closes
- https://github.com/logos-messaging/logos-delivery/issues/3841


NagyZoltanPeter (approved):
Looks good!

PR #3860, opened 2026-05-07
ensure peers are retrieved in random order from peer store

Ivansete-status:
This is to enhance an issue detected by @AlbertoSoutullo 🙌 where `recv_monitor` always performed store queries to the same store node (considering two store node candidates are available.)

With that:
- Randomize `proc peers*(peerStore: PeerStore): seq[RemotePeerInfo]` in `waku/node/peer_manager/waku_peer_store.nim`.
- Add simple unit test to validate that.
- Add nitpick parenthesis to other `peers()` calls within `waku/node/peer_manager/waku_peer_store.nim`.

### DIFF
--- a/tests/test_peer_store_extended.nim
+++ b/tests/test_peer_store_extended.nim
@@ -1,7 +1,7 @@
 {.used.}
 
 import
-  std/[sequtils, times, sets],
+  std/[sequtils, times],
   chronos,
   libp2p/crypto/crypto,
   libp2p/peerid,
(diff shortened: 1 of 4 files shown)


fcecin (approved):
LGTM

darshankabariya (approved):
Hi ivan
This changes works, but I think the randomization done at the wrong layer. getPeer is used so many other location which like ( getConnectedPeers, getPeersByProtocol, getPeersByDirection ) which i don't think randomness required there.  

I would suggest that use randomize at query level ( recv_monitor.nim ) is better.

PR #3862, opened 2026-05-07
Evict peer instead of abrupt disconnect and avoid sending store requests

Ivansete-status:
This is to update master with recent changes/fixes in `release/v0.38`.

- Peer manager not disconnect abruptly ongoing service store requests.
- If there are no missing messages, then receive monitor won't send store queries.


Ivansete-status:
We need to update master from another branch, that is not `release/v0.38`, so that we can rebase it.

Issue #3845, opened 2026-05-02
API design and consistency


igor-sirotin:
# API design and consistency

We have a few interfaces to configure a Delivery node: CLI, library (Nim / C-bindings), REST. We should agree which is for which role, and what each looks like.

## Discussions

- https://github.com/logos-messaging/logos-delivery/issues/3795
- https://github.com/logos-messaging/logos-delivery/pull/3828
  - https://github.com/logos-messaging/logos-delivery/pull/3828#issuecomment-4317261531
  - https://github.com/logos-messaging/logos-delivery/pull/3828#issuecomment-4337495581
  - https://github.com/logos-messaging/logos-delivery/pull/3828#issuecomment-4353802313
- https://github.com/logos-co/logos-delivery-module/issues/18

## Roles

- **App developer** — sends/receives messages on a known network.
- **Node operator** — runs a fleet node 24/7.
- **Tester (DST / QA)** — drives custom configs, often many nodes per host.

## Requirements

- **No breaking changes.** Soft deprecations are fine.
- **Good defaults.** App developers should get a working node from a one-liner.
- **Flexibility for node operators.** Every preset/mode-controlled value must be overridable by an explicit field (#3795).

## Principles

1. Every node starts from a `WakuNodeConf`. Every interface must produce one deterministically:

   ```mermaid
   flowchart LR
       CLI["CLI flags<br/>--preset, --tcp-port, --relay, ..."] -->|build| Conf
       Msg["Library: Messaging API<br/>preset, mode, overrides"] -->|build| Conf
       Full["Library: <br/>full WakuNodeConf"] -->|pass| Conf
       Conf[WakuNodeConf] --> Waku["Waku.new(conf)"]
   ```
2. Different interfaces don't need identical argument lists. Each is shaped for its role.
3. Developer gets a shortcut: `(preset, mode, overrides?)`. They never assemble a full `WakuNodeConf`.
4. Operator gets the full `WakuNodeConf` surface.

## Problems

1. **No `(preset, mode)` entry point.** The library only has `createNode(conf: WakuNodeConf)`:
    https://github.com/logos-messaging/logos-delivery/blob/75864a705ea0b913d517a5f3640747f8709e9e53/waku/api/api.nim#L16
    and the C FFI mirror at:
    https://github.com/logos-messaging/logos-delivery/blob/75864a705ea0b913d517a5f3640747f8709e9e53/liblogosdelivery/logos_delivery_api/node_api.nim#L94-L96
[...]

PR #3866, opened 2026-05-08
feat: Improve config

### LINKED ISSUE #3819: Allow port selection automatic with config to 0

fcecin:
## Description

NOTE: This PR addresses most points raised in #3845, with the exception of removing `--mode` and `createNode(conf: WakuNodeConf)`.

This PR changes how network presets work by making the various scalar field values in a selected preset "lose" to any scalar values provided explicitly by user configs instead. And in the case of list parameters (seq), user-provided values can either override the preset or concatenate to the preset. This is dealt with by a new "additions?" optional field in the `configJson` parameter provided to the C FFI API `createNode`.

A new type `WakuNodeConfOverlay` is auto-generated from the `WakuNodeConf` authoritative type using a macro that makes all fields relevant to overrides and additions `Option[T]` where the `T` type of the field in the original `WakuNodeConf` type wasn't already an `Option`.

Some int/bool fields in `WakuNodeConf` itself had to be promoted to `Option[int/bool]` so that the new `applyNetworkConf` proc could apply the new logic of user values overriding preset values without relying on chosen sentinel int or bool values for each field.

The JSON parser inside the FFI entry point for createNode was moved to a dedicated parser that was also rewritten.

Literal value defaults across the various config builder files were replaced with `DefaultXXX` const values instead. These const defaults owned by each conf builder are referenced in the CLI help text whenever the field in question has been upgraded to `Option`, which results in the CLI option having a default of `none` (unset), which is not very helpful for the generated help text.

And finally, when using the new `createNode(preset, mode, overrides, additions)` Nim API, which signals a library app developer use case, this PR defaults the `p2pTcp`, `discv5Udp` and `websocket` ports to zero (0), which now mean a request to automatically choose and bind a port. That closes the issues related to port collisions for app developers running multiple instances of `logos-delivery` on the same machine. For reference, the complete discussion on this topic can be found in PR #3828's conversation tab.

## Changes

* Add createNode(preset, mode, overrides, additions) nim api
* Set p2pTcp/discv5Udp/websocket ports to 0 (auto-bind) in new createNode()
* Soft-deprecate --cluster-id=N triggering the associated preset selection
* Rewrite applyNetworkConf to apply user-set fields over preset fields
* Generate WakuNodeConfOverlay (all Option fields) from WakuNodeConf
* New parser for configJson handles new messaging shape and full conf shape
* Change all confbuilder defaults from literal values to DefaultXXX consts
* Change int/bool WakuNodeConf fields to Option to get user intent w/o sentinels
* Make Option CLI default-value help mention defaults now owned by confbuilder
* Misc refactors, fixes
* Add tests

## Issue

iterates #3845

closes https://github.com/logos-co/logos-delivery-module/issues/18

closes #3819


tools/confutils/cli_args.nim
```diff
@@ -170,12 +173,15 @@ type WakuNodeConf* = object
...
-        "Cluster id that the node is running in. Node in a different cluster id is disconnected.",
-      defaultValue: 0,
+      desc: static(
+        "Cluster id that the node is running in. Node in a different cluster id is disconnected. Default is " &
+          $DefaultClusterId & "."
+      ),
+      defaultValue: none(uint16),
       name: "cluster-id"
-    .}: uint16
+    .}: Option[uint16]
```
Ivansete-status, line 184: What could be an scenario where a node doesn't need to know about a `cluster-id`?

tools/confutils/cli_args.nim
```diff
@@ -466,13 +475,14 @@ hence would have reachability issues.""",
...
       desc:
-        """Adds an extra effort in the delivery/reception of messages by leveraging store-v3 requests.
-with the drawback of consuming some more bandwidth.""",
-      defaultValue: true,
+        """Adds an extra effort in the delivery/reception of messages by leveraging store-v3 requests, with the drawback of consuming some more bandwidth. Default is """ &
+        $DefaultP2pReliability & ".",
+      defaultValue: none(bool),
       name: "reliability"
-    .}: bool
+    .}: Option[bool]
```
Ivansete-status, line 485: How we should interpret a `none` value? As a `true` or `false` :) ?
Questions applies elsewhere.

tools/confutils/conf_from_json.nim
```diff
@@ -0,0 +1,257 @@
...
+  KeyPreset = "preset"
+  KeyOverrides = "overrides"
+  KeyAdditions = "additions"
+
+const CreateNodeWithOverridesExplicitKeys = [KeyMode, KeyPreset]
+  ## Keys that map to explicit parameters of `createNode(preset, mode, ...)`,
+  ## hence parsed at the messaging shape's top level and rejected inside
+  ## `overrides`/`additions` to avoid ambiguity.
+
+proc collectJsonFields*(
```
Ivansete-status, line 17: Just fyi, this week or after, `nim-ffi` will handle an arbitrary `Config` type, and the wire conversion (through CBOR) will happen within `nim-ffi` internally.
In the upcoming future we should not have to manually deal with that as `nim-ffi` will be more type safe.

tools/confutils/conf_from_json.nim
```diff
@@ -0,0 +1,257 @@
...
+  var conf = ?defaultWakuNodeConf()
+  var fields = jsonFields
+  ?applyJsonFieldsToConf(
+    conf, fields, "Failed to parse field", "Unrecognized configuration option(s) found"
+  )
+  return ok(conf)
+
+proc parseConfJson*(jsonStr: string): Result[WakuNodeConf, string] =
+  ## Parse a JSON config, route to messaging or full-config shape based on
+  ## whether `overrides` or `additions` fields are in the config object top-level.
```
igor-sirotin, line 241: I think `overrides` and `additions` gives too much flexibility, and introduces too much unwated complexity.

I was envisioning the overrides as a concrete list of possible advanced configuration of `MessagingClient`. Otherwise it allows to have e.g. `mode: Core` with `overrides {relay: false}`, which is an invalid config I believe. 

At the moment I would just not add overrides/additions at all. And later, introduce certain overrides on demand.

tools/confutils/messaging_conf.nim
```diff
@@ -0,0 +1,40 @@
...
+
+const WakuNodeConfOverlayExcludes* = ["cmd", "execute"]
+  ## Variant-safety: `cmd` is the CLI subcommand discriminator (not dispatched
+  ## on by the library) and `execute` lives in its inactive branch. Excluded
+  ## from the overlay AND used as the JSON parser's hard-reject list.
+
+# Generates the WakuNodeConfOverlay type from the WakuNodeConf type.
+# The generated type converts fields from type T to Option[T] if T != Option.
+# Skips fields that are in WakuNodeConfOverlayExcludes.
+optionalizeType(WakuNodeConfOverlay, WakuNodeConf, WakuNodeConfOverlayExcludes)
```
Ivansete-status, line 16: I wonder if is better to be more explicit and just add the regular type definition.
Maybe not needed to use a macro for this particular case.

tools/confutils/optionalize.nim
```diff
@@ -0,0 +1,74 @@
+{.push raises: [].}
+
+import std/[macros, options]
```
Ivansete-status, line 3: Maybe not needed this file if agree with previous comment.

waku/api/api.nim
```diff
@@ -24,6 +26,30 @@ proc createNode*(conf: WakuNodeConf): Future[Result[Waku, string]] {.async.} =
...
 
+proc seedDeveloperProfile(conf: var WakuNodeConf) =
+  # TODO: Remember to add QUIC port here as well when that is added.
+  var devPorts = WakuNodeConfOverlay.init()
+  devPorts.tcpPort = some(Port(0))
+  devPorts.discv5UdpPort = some(Port(0))
+  devPorts.websocketPort = some(Port(0))
+  applyAsOverride(conf, devPorts)
+
+proc createNode*(
```
Ivansete-status, line 37: Just a note for the future. We might need to have this proc within the upcoming refactoring, inside the `messaging/api` section.

Ivansete-status (approved):
Nice refactor, thanks!
Just curious how to properly handle a ternary `Option[bool]` attribute.
In the upcoming future I think we should only have three or four config object types, e.g., `ChannelApiConf`, `WakuAppConf`, `MessagingApiConf`, `KernelApiConf`.

tools/confutils/cli_args.nim
fcecin: The node would know, via a `--preset`. You just don't need to inform one via the CLI.

tools/confutils/cli_args.nim
fcecin: It is interpreted as literally none -- the CLI didn't specify one via this config knob. Then we have a dense pipeline of decision-making downstream that will either populate the configured concern somehow or blow up with "I reached this point and it's still set to none," same as the cluster-id case.

tools/confutils/messaging_conf.nim
fcecin: Either a macro or a hand-maintained mirror type, in any case I was just following the design from @igor-sirotin literally (or trying to). If we could instead make the original WakuNodeConf *be* forcefully made entirely of Option[T], then we wouldn't need a second type.

I think I've seen Igor arguing about CLI breakage being OK for the window of time we have now, in which case yes, we should just have one config object that is Option[T], which allows it to function as a base config *and* as a config overlay.

waku/api/api.nim
fcecin: Ah yes, the refactoring PR is going to move a lot of stuff around. My current draft is +45,000 lines.

fcecin:
> Nice refactor, thanks! Just curious how to properly handle a ternary `Option[bool]` attribute. In the upcoming future I think we should only have three or four config object types, e.g., `ChannelApiConf`, `WakuAppConf`, `MessagingApiConf`, `KernelApiConf`.

I think the case for nuking the CLI / config stuff and rewriting this is about right now. So maybe we should close this PR and submit a new one with a full conf pipeline redesign and rewrite. If we can afford the breakage budget.

tools/confutils/conf_from_json.nim
fcecin: In this PR, "overrides" and "additions" is just the way the user accesses the semantics and capability of achieving *any* configuration goal, at any level (kernel, messaging, and channels -- upcoming), just like mode and preset are (in principle) free to affect configuration at any layer.

As for "invalid configs," this brings up the more fundamental discussion (which I guess we will have a meeting about very soon) about what _is_ it that developers have accessible to them. If we choose (my preferred route) of giving the user full control, that means by "user" we also mean e.g. QA/DST, so the node is fully testable -- so "relay: false" is not something for us to judge whether that's "valid" or not; if the user did it, then it is valid.

An alternative is to do what Ivan has already anticipated his comments below: rewrite the config system and split it into layers also (somehow). So maybe the CLI retains the "I can configure any layer" semantics, but via library you are accessing a different configuration pipeline for each (kernel?, messaging, channels) layer.

igor-sirotin: @fcecin I like the suggestion of layered configuration.

As for the _flexibility_  point, it can be achieved this way (which I mentioned in https://github.com/logos-messaging/logos-delivery/issues/3865):

### Kernel API

```nim
# full flexibility, build what you need
proc createNode(WakuNodeConf)
```

### Messaging API
```nim
# no flexibility, only preset+mode (and maybe some messaging client specific config)
proc createMessagingClient(preset, mode)
    # Translate preset+mode -> WakuNodeConf
    # Create the node with createNode(wakuNodeConf)

# adds flexibility, you can create the node directly with Kernel API and then attach a Messaging Client. Without layering violations.
proc attachMessagingClient(node)
    # attaches a messaging client to a running node
```

fcecin: Got it. I'll figure out how to translate this to PR.

fcecin:
As for right now, I'm going to _slightly_ retarget my attetion to the main layer-refactoring task. This will move in parallel while I figure out how to precisely synthesize the result of our decisions in the last API design meeting, and the new constraints that we have articulated there, w.r.t. to the config types and config surface in general.

We can continue discussing the config surface design here -- the more information, the better. 

EDIT: Just so that this comment isn't entirely useless:

```
# no flexibility, only preset+mode (and maybe some messaging client specific config)
proc createMessagingClient(preset, mode)
    # Translate preset+mode -> WakuNodeConf
    # Create the node with createNode(wakuNodeConf)
```

... we realized in the meeting that:
1.  "(preset, mode)" is likely _not_ sufficient to fully express an "easy-start-up intent" with removed power and thus increased safety and UX for app developers (ie. removing footguns and "baby-proofing" the config / node creation entry point).
2. But at the _same time_, we realized that we _do_ want a different, specialized signature for higher-level creation of e.g. a MessagingClient at the Messaging layer of API access that is not just "here's the full conf object" (be it proxied via Json <--> single cstring FFI arg, or be it some native struct/type FFI bridge machinery).
 
And also at the same time, in the review of _this_ PR in its current form, we realized here that (preset, mode, overrides, additions) _as implemented here_ is not what we want either, because it gives too much config freedom with invalid combinations, etc.

Since that decision and/or its implementation are clearly _affected_ by what the relatively large re-layering refactor will do, it is probably a good idea to delay config improvements for _after_ the re-layering effort. But we can also decide something here and push this first; that also works if we want to make the calls now and get this done with.

(Also I understand the whole CLI vs Library issue is 100% settled; the CLI adjustments can be resolved after we decide the config-related APIs and config struct type(s) we want).

tools/confutils/conf_from_json.nim
NagyZoltanPeter: I can't really agree with this as is. I think we will always need to provide support for json and/or toml formatted string / file based config. This will be needed for node operators and infra.
But kind-a yes, structured typed api will help devs.

tests/api/test_messaging_conf.nim
```diff
@@ -0,0 +1,199 @@
...
+    let conf = res.get()
+    check conf.preset == "twn"
+
+  test "Messaging shape applies additions to list fields":
+    let res = parseConfJson(
+      """{"mode": "Core", "overrides": {}, "additions": {"staticnodes": ["/ip4/1.2.3.4/tcp/60000/p2p/16Uiu2HAmTUbnxLGT9JvV6mu9oPyDjqHK4Phs1VDJNUgESgNSkuby"]}}"""
+    )
+    require res.isOk()
+    let conf = res.get()
+    check conf.staticnodes.len == 1
```
NagyZoltanPeter, line 45: IMHO this test case must fail. without preset and cluster-id it is an invalid setting. Either one of the should exist, no?

tools/confutils/cli_args.nim
```diff
@@ -170,12 +173,15 @@ type WakuNodeConf* = object
...
+    # Option-typed; desc states the default since the CLI can't auto-show it for none().
     clusterId* {.
-      desc:
-        "Cluster id that the node is running in. Node in a different cluster id is disconnected.",
-      defaultValue: 0,
+      desc: static(
+        "Cluster id that the node is running in. Node in a different cluster id is disconnected. Default is " &
+          $DefaultClusterId & "."
+      ),
+      defaultValue: none(uint16),
```
NagyZoltanPeter, line 182: These are just exluding each other. What could be a defaultClusterId btw?

tools/confutils/cli_args.nim
NagyZoltanPeter: IMHO its either `--preset` or `--cluster-id`, one must exist from the two...
So optional a bit misleading - although I have no better option - but description must be explicit about it.
If --preset exist and --cluster-id is given there are two path - if matches --preset's one than silently go ok, but if not that's a config error.

NagyZoltanPeter (approved):
Approve it with the request on re-think role of cluster-id setting.

fcecin:
> Approve it with the request on re-think role of cluster-id setting.

This PR was killed (in the current format). It was decided against generating a mirror type that applies `Option` to everything so it can act as a full config overlay (while not disturbing the original type for compat).

Thanks for approving this one -- that helps, because it is the base branch for the [follow-up](https://github.com/logos-messaging/logos-delivery/pull/3925), which will be much simpler.

By default we're going with @Ivansete-status 's idea of creating an explicit `MessagingClientConf` object which contains everything that we want to strictly expose as the node configuration at that layer. That allows us to completely internalize the current configuration system -- it moves to the Kernel layer, meaning it is unsupported and thus we are free to change it as it is part of the new kernel API, which is also something we are free to change.

My current plan is to allow the user of the messaging_client layer to specify a `preset`, a `mode`, and then optionally pass in an `overrides: MessagingClientConf` that will be all `Option` natively (we can do that, since it is an original type). Every field that is `some` can potentially override a preset/mode field. preset can be unset (say, `""` sentinel) also, and mode has to be either Edge or Core. That should [address most of the original issue](https://github.com/logos-messaging/logos-delivery/issues/3845).

As for the cluster ID setting, I'm going with [this interpretation](https://github.com/logos-messaging/logos-delivery/issues/3845#issuecomment-4435148150) by default. So if you set  the cluster ID option in the `MessagingClientConf`, it means you know what you are doing and it gets overriden in any preset you're giving it.

Note that another possible interpretation of what a `MessagingClientConf` object would mean is "it's just the messaging layer configuration items!" I.e. it would mean that to actually configure the whole node with meaningful overridable defaults that are homed in the lower-level (kernel, WakuNodeConf), you'd need to actually know about and access the low-level config object as well (which is part of the low-level API not because of its "low-level" configuration contents, but because of which node construction procs take it). Although initially attractive, I'm against that idea, and I think that idea in particular is dominated by *this* PR -- then it is better to just have one global config object that contains everything and doesn't belong to any layer in particular. So instead, `MessagingClientConf` is an _editor_ that allows us to define precisely what are _all_ the options that should be available to a MessagingClient-layer (app-dev) user, meaning some low-level configs make it untouched, others modified, and others removed.

PR #3874, opened 2026-05-13
stop recv_service from delivering messages on unsubscribed topics for store-recovered messages

### LINKED ISSUE #3872: Store-recovery sweep delivers messages for unsubscribed content topics
## Symptom

Clients receive `MessageReceivedEvent` for content topics they never passed to `subscribe()`. Reproduced from `logos-delivery-demo` (Core mode, `logos.dev` preset, 8 shards): with no subscriptions — or a single subscription — events arrive within ~5 min on foreign topics such as `/chronicle/1/document-index/json` and `/test/2/waku-filter/default`.

## Root cause

The live-relay path is fine. `MessageSeenEvent` listener correctly checks `subscriptionManager.isSubscribed(shard, contentTopic)` and drops unmatched messages: [recv_service.nim#L163-L174](https://github.com/logos-messaging/logos-delivery/blob/27ae07adaaea7beeae02cea9f8647b18cd9fb482/waku/node/delivery_service/recv_service/recv_service.nim#L163-L174).

The **store-recovery sweep** (`RecvService.checkStore`, runs every `StoreCheckPeriod` = 5 min) does not:

1. In Core mode, `SubscriptionManager.startSubscriptionManager` auto-subscribes the relay to every shard in the cluster, inserting each shard into `contentTopicSubs` with an **empty** `HashSet[ContentTopic]` — [subscription_manager.nim#L562-L580](https://github.com/logos-messaging/logos-delivery/blob/27ae07adaaea7beeae02cea9f8647b18cd9fb482/waku/node/delivery_service/subscription_manager.nim#L562-L580).
2. `checkStore` iterates `subscribedTopics` and issues a `StoreQueryRequest` per shard with `contentTopics: toSeq(contentTopics)`. For shards with no user interest this is an empty filter, so the store returns every message on the shard — [recv_service.nim#L94-L108](https://github.com/logos-messaging/logos-delivery/blob/27ae07adaaea7beeae02cea9f8647b18cd9fb482/waku/node/delivery_service/recv_service/recv_service.nim#L94-L108).
3. Returned messages are handed to `processIncomingMessageOfInterest`, which **only dedups by hash** before emitting `MessageReceivedEvent` — no content-topic check — [recv_service.nim#L73-L86](https://github.com/logos-messaging/logos-delivery/blob/27ae07adaaea7beeae02cea9f8647b18cd9fb482/waku/node/delivery_service/recv_service/recv_service.nim#L73-L86) (called from [#L122-L126](https://github.com/logos-messaging/logos-delivery/blob/27ae07adaaea7beeae02cea9f8647b18cd9fb482/waku/node/delivery_service/recv_service/recv_service.nim#L122-L126)).

Net effect: foreign traffic sitting in the cluster store gets re-emitted to the client every 5 minutes.

## Reproduction

1. Build/run `logos-delivery-demo` with the default config (`preset: "logos.dev"`, `mode: "Core"`).
2. Subscribe to nothing, or to one arbitrary content topic.
3. Wait up to `StoreCheckPeriod` (5 min).
4. Observe `messageReceived` events on content topics that were never subscribed.

## Suggested fix

- In `checkStore`, skip shards whose `contentTopics` set is empty (or never issue a store query without an explicit content-topic filter).
- Enforce `isSubscribed(shard, contentTopic)` in (or right before) `processIncomingMessageOfInterest` so every path that emits `MessageReceivedEvent` shares the same invariant as the relay listener.

## Severity

Correctness — clients receive events they didn't ask for; wastes bandwidth/CPU and can mislead apps. No security impact beyond exposing what is already publicly readable from the store.

fcecin:
## Description

This PR fixes a bug where the Messaging API's RecvService would query the store with an empty content topic list, signaling interest in messages for all content topics in a given pubsubtopic (shard), instead of no interest in any content topic for the pubsubtopic.

## Changes

* RecvService.processIncomingMessage enforces filtering messages for content topic interest
* RecvService.checkStore skips store queries for shards with no content topics of interest 
* Fix the test_api_receive test

## Issue

closes #3872

### DIFF
--- a/waku/node/delivery_service/recv_service/recv_service.nim
+++ b/waku/node/delivery_service/recv_service/recv_service.nim
@@ -70,20 +70,24 @@ proc getMissingMsgsFromStore(
     )
   )
 
-proc processIncomingMessageOfInterest(
+proc processIncomingMessage(
     self: RecvService, pubsubTopic: string, message: WakuMessage
 ): bool =
-  ## Deduplicate (by hash), store (saves in recently-seen messages) and emit
-  ## the MAPI MessageReceivedEvent for every unique incoming message.
-  ## Returns true if the message was new and the MessageReceivedEvent was properly emitted.
+  ## Return false if the incoming message is from a non-subscribed topic,
+  ## or if the message is a duplicate (recently-seen). Otherwise, save it as
+  ## recently-seen, emit a MessageReceivedEvent, and return true.
+
+  if not self.subscriptionManager.isSubscribed(pubsubTopic, message.contentTopic):
+    return false
 
   let msgHash = computeMessageHash(pubsubTopic, message)
-  if not self.recentReceivedMsgs.anyIt(it.msgHash == msgHash):
-    let rxMsg = RecvMessage(msgHash: msgHash, rxTime: message.timestamp)
-    self.recentReceivedMsgs.add(rxMsg)
-    MessageReceivedEvent.emit(self.brokerCtx, msgHash.to0xHex(), message)
-    return true
-  return false
+  if self.recentReceivedMsgs.anyIt(it.msgHash == msgHash):
+    return false
+
+  let rxMsg = RecvMessage(msgHash: msgHash, rxTime: message.timestamp)
+  self.recentReceivedMsgs.add(rxMsg)
+  MessageReceivedEvent.emit(self.brokerCtx, msgHash.to0xHex(), message)
+  return true
 
 proc checkStore*(self: RecvService) {.async.} =
   ## Checks the store for messages that were not received directly and
@@ -92,6 +96,10 @@ proc checkStore*(self: RecvService) {.async.} =
 
   ## query store and deliver new recovered messages per subscribed topic
   for pubsubTopic, contentTopics in self.subscriptionManager.subscribedTopics:
+    # skip Store query if we are not subscribed to any content topics on the shard
+    if contentTopics.len == 0:
+      continue
+
     let storeResp: StoreQueryResponse = (
       await self.node.wakuStoreClient.queryToAny(
         StoreQueryRequest(
@@ -121,7 +129,7 @@ proc checkStore*(self: RecvService) {.async.} =
       let missingMsgsRet = await self.getMissingMsgsFromStore(missedHashes)
       if missingMsgsRet.isOk():
         for msgTuple in missingMsgsRet.get():
-          if self.processIncomingMessageOfInterest(msgTuple.pubsubTopic, msgTuple.msg):
+          if self.processIncomingMessage(msgTuple.pubsubTopic, msgTuple.msg):
             info "recv service store-recovered message",
               msg_hash = shortLog(msgTuple.hash), pubsubTopic = msgTuple.pubsubTopic
       else:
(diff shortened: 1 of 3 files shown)


darshankabariya (approved):
The PR title confused me first. after seeing changes look correct to me.

waku/node/delivery_service/recv_service/recv_service.nim
```diff
@@ -70,20 +70,24 @@ proc getMissingMsgsFromStore(
...
 ): bool =
-  ## Deduplicate (by hash), store (saves in recently-seen messages) and emit
-  ## the MAPI MessageReceivedEvent for every unique incoming message.
-  ## Returns true if the message was new and the MessageReceivedEvent was properly emitted.
+  ## Return false if the incoming message is from a non-subscribed topic,
+  ## or if the message is a duplicate (recently-seen). Otherwise, save it as
+  ## recently-seen, emit a MessageReceivedEvent, and return true.
+
+  if not self.subscriptionManager.isSubscribed(pubsubTopic, message.contentTopic):
+    return false
```
Ivansete-status, line 83: Shall we add `debug` logs for `false` cases given that now there are more?

waku/node/delivery_service/recv_service/recv_service.nim
```diff
@@ -163,14 +171,7 @@ proc startRecvService*(self: RecvService) =
...
     proc(event: MessageSeenEvent) {.async: (raises: []).} =
-      if not self.subscriptionManager.isSubscribed(
-        event.topic, event.message.contentTopic
-      ):
-        trace "skipping message as I am not subscribed",
-          shard = event.topic, contenttopic = event.message.contentTopic
-        return
-
-      discard self.processIncomingMessageOfInterest(event.topic, event.message),
+      discard self.processIncomingMessage(event.topic, event.message),
```
Ivansete-status, line 176: I realized now that maybe we can remove the `bool` return from `processIncomingMessage` and then always trace the `false` scenario inside.

waku/node/delivery_service/recv_service/recv_service.nim
```diff
@@ -92,6 +96,10 @@ proc checkStore*(self: RecvService) {.async.} =
 
   ## query store and deliver new recovered messages per subscribed topic
   for pubsubTopic, contentTopics in self.subscriptionManager.subscribedTopics:
```
Ivansete-status, line 104: I wonder if, instead, we should make the `subscribedTopics` iterator to not return the items that has no contentTopics.

Ivansete-status (approved):
LGTM! Thanks for it! 💯

waku/node/delivery_service/recv_service/recv_service.nim
fcecin: I'll re-add it at trace level, which is what we had there. We probably don't want "are the way we fundamentally filter message delivery working?" extraordinary condition checked while we're `debug`ing more sane kinds of failures.

waku/node/delivery_service/recv_service/recv_service.nim
fcecin: The compiler needs the bool there to preserve this bit of preexisting logging code in the filter acceptance path for Store-recovered messages:

```
          if self.processIncomingMessage(msgTuple.pubsubTopic, msgTuple.msg):
            info "recv service store-recovered message",
              msg_hash = shortLog(msgTuple.hash), pubsubTopic = msgTuple.pubsubTopic
```

I usually don't mind functions returning unused status bits -- at least that's pretty common in e.g. C++. Nim not liking that kind of thing and requiring `discard` for explicit handling is the price we pay for the extra compiler-enforced safety discipline.

But I agree, it's a good idea to embrace the host language's implicit culture and try to avoid these. But to do that here:

* Create a function to filter (which does return bool) and another one to deliver (returns nothing);
* Pass "bool store" for composing the log text inside the `processIncomingMessage` filter+deliver proc;
* Or pass "string reason" to it instead; 
* Abandon this "success case for Store" logging hint;
* Something else.

I'm not sure what is it that we actually want here. This pattern is general enough that if we establish it, then it probably collapses similar decisions elsewhere. Then we can put this in our upcoming style guide and possibly auto-styling tools.

waku/node/delivery_service/recv_service/recv_service.nim
fcecin: Brilliant catch! The semantics of `SubscriptionManager` are content-topic-level; shards are the core abstraction for subscriptions at the _kernel_ level. This is actually a bug, let me fix this.

fcecin:
Accidentally figured out that the test_api_receive regression test that was preexisting was broken. Fixed it while we're here.

PR #3878, opened 2026-05-13
chore: fixing daily ci

### LINKED ISSUE #3876: bug: Daily CI failing
Ubuntu and macOS jobs fail at make update with exit code 2. 
Ubuntu shows an explicit error: Package not found in solution: nim 2.2.10 — nimble's solver can't find nim 2.2.10 but our lock file only pins 2.2.4. Likely a transitive dependency bumped its minimum nim version. 
macOS fails at the same step with no explicit message.

darshankabariya:
close #3876

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,6 +24,7 @@ export PATH := $(HOME)/.nimble/bin:$(PATH)
 # NIM binary location
 NIM_BINARY := $(shell which nim 2>/dev/null)
 NPH := $(HOME)/.nimble/bin/nph
+NIMBLE := $(HOME)/.nimble/bin/nimble
 NIMBLEDEPS_STAMP := nimbledeps/.nimble-setup
 
 # Compilation parameters
@@ -71,7 +72,7 @@ waku.nims:
 	ln -s waku.nimble $@
 
 $(NIMBLEDEPS_STAMP): nimble.lock | install-nimble build-nph waku.nims
-	nimble setup --localdeps
+	$(NIMBLE) setup --localdeps
 	touch $@
 
 # Must be phony so the recipe always runs and the sub-make re-evaluates
--- a/Makefile
+++ b/Makefile
@@ -72,7 +72,7 @@ waku.nims:
 
 $(NIMBLEDEPS_STAMP): nimble.lock | waku.nims
 	$(MAKE) install-nimble
-	nimble setup --localdeps
+	$(NIMBLE) setup --localdeps
 	$(MAKE) build-nph
 	$(MAKE) rebuild-bearssl-nimbledeps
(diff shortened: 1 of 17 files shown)


waku.nimble
```diff
@@ -9,14 +9,13 @@ author = "Status Research & Development GmbH"
 description = "Waku, Private P2P Messaging for Resource-Restricted Devices"
 license = "MIT or Apache License 2.0"
 
-const RequiredNimVersion = "2.2.4"
+const RequiredNimVersion = "2.2.10"
```
NagyZoltanPeter: I think this is not needed yet.
First we will need to validate the build with new compiler on a know good version of master -> run complete tests on fleets and maybe DST. @Ivansete-status

Makefile
```diff
@@ -24,6 +24,7 @@ export PATH := $(HOME)/.nimble/bin:$(PATH)
 # NIM binary location
 NIM_BINARY := $(shell which nim 2>/dev/null)
 NPH := $(HOME)/.nimble/bin/nph
+NIMBLE := $(HOME)/.nimble/bin/nimble
```
NagyZoltanPeter, line 27: Is this applies to all ci run on all platform?

nimble.lock
```diff
@@ -2,13 +2,13 @@
   "version": 2,
   "packages": {
     "nim": {
-      "version": "2.2.4",
-      "vcsRevision": "911e0dbb1f76de61fa0215ab1bb85af5334cc9a8",
+      "version": "2.2.10",
+      "vcsRevision": "9fe2137fa2f3f66cf5a44f357d461829ac9e20c4",
```
NagyZoltanPeter: Same as above, we need to keep compiler version we know as producing good code.

NagyZoltanPeter (commented):
Added comment. 
Why compiler version bump is needed to make ci happy?

Makefile
Ivansete-status: I think it will work fine as long as the `NIMBLE_DIR` is not overridden, which is the case.

nimble.lock
Ivansete-status: Yes indeed, not needed to change nimble.lock

waku.nimble
Ivansete-status: Yes exactly, we will stick to 2.2.4 for now ( cc @darshankabariya .)

Ivansete-status (changes requested):
Thanks for the PR!, some comments so far

Ivansete-status (approved):
LGTM! Thanks for it! 💯

fcecin (approved):

NagyZoltanPeter (approved):
Look all ok for me! Thank you.

PR #3880, opened 2026-05-14
feat: persistency

NagyZoltanPeter:
Introduce data schema agnostic local storage of states.

## Summary

- New \`waku/persistency/\` library: per-job SQLite-backed CRUD primitives, ports-and-adapters architecture, accessed through a process-wide \`Persistency\` singleton and per-job \`Job\` handles.
- Writes are fire-and-forget over \`EventBroker(mt) PersistEvent\`; reads are async \`RequestBroker(mt)\` per response shape (KvGet/KvExists/KvScan/KvCount/KvDelete) returning \`Result[T, PersistencyError]\`. One storage thread per job, jobs isolated by \`BrokerContext\`.
- Key + payload encoders: \`encodePart\` family + variadic \`key(...)\` / \`payload(...)\` macros, generic over tuples/objects via \`fields()\` recursion; sort-stable for keys, stable across compiler upgrades (no \`sizeof\`/memcpy).
- Public surface: \`Persistency.instance(rootDir)\` / \`.instance()\` / \`.reset()\`; \`p.openJob/closeJob/dropJob/close\`; string-id lookup (\`p.job(id)\`, \`p[id]\`, \`p.hasJob(id)\`); per-op procs in both Job-ref and jobId-string forms. \`Teardown\` provider integrated with \`Waku.stop()\`.
- Temporarily pins \`nim-brokers\` v2.0.1 by URL+commit (\`6e7b7cc1\`) until the nim-lang/packages registry entry for \`brokers\` is refreshed past v0.1.0. Inline comment in \`waku.nimble\` documents the path back to the bare \`brokers >= 2.0.1\` form.

## Test plan

- [x] \`nim c -r tests/persistency/test_all.nim\` -- 69/69 pass:
  - composite-key sort-order invariants (test_keys.nim)
  - SQLite backend round-trip / replace / batched atomicity / scan asc-desc-half-open-open-ended / category isolation (test_backend.nim)
  - lifecycle: open-or-create rootDir, non-dir collision, reopen-across-sessions, two-tenant parallel isolation, closeJob joins worker, dropJob removes file (test_lifecycle.nim)
  - facade put-then-get, batch, scanPrefix asc/desc, deleteAcked, persistDelete, two-tenant facade isolation (test_facade.nim)
  - generic encoding: tuples / named tuples / objects / nested / enums / field-major composite sort / payload struct round-trip through SQLite (test_encoding.nim)
  - string-id lookup: peJobNotFound semantics, hasJob, subscript, persistPut+get via id, reads short-circuit, writes drop+warn, persistEncoded via id, scan parity Job-ref vs id (test_string_lookup.nim)
  - singleton: idempotent same-rootDir, conflicting-rootDir rejection, no-arg lifecycle, reset retargets, reset idempotence, Teardown.request end-to-end (test_singleton.nim)
- [ ] CI build + full nimble test suite on this branch.

## delayed - follow up later

- schema versioning solution to be added to avoid upgrade caused incompatible load of data
- SDS adapter (Phase 5 of the design) -- nim-sds is developed side-by-side; this layer is intentionally SDS-agnostic.
- Hardening (Phase 6): backpressure on PersistEvent, synchronous=FULL toggle for mobile, online SQLite backup, benchmarks.
- metrics to be added

## closes

https://github.com/logos-messaging/logos-delivery/issues/3825


fcecin (approved):

apps/chat2/config_chat2.nim
```diff
@@ -140,7 +140,8 @@ type
 
     metricsServerAddress* {.
       desc: "Listening address of the metrics server.",
-      defaultValue: parseIpAddress("127.0.0.1"),
+      defaultValue:
+        IpAddress(family: IpAddressFamily.IPv4, address_v4: [127'u8, 0, 0, 1]),
```
Ivansete-status, line 144: This seems unrelated.

nix/deps.nix
```diff
@@ -129,6 +129,13 @@
     fetchSubmodules = true;
   };
 
+  cbor_serialization = pkgs.fetchgit {
```
Ivansete-status, line 132: Why this is needed now?

tests/api/test_api_send.nim
```diff
@@ -77,10 +77,12 @@ proc newSendEventListenerManager(brokerCtx: BrokerContext): SendEventListenerMan
 
   return manager
 
-proc teardown(manager: SendEventListenerManager) =
-  MessageSentEvent.dropListener(manager.brokerCtx, manager.sentListener)
-  MessageErrorEvent.dropListener(manager.brokerCtx, manager.errorListener)
-  MessagePropagatedEvent.dropListener(manager.brokerCtx, manager.propagatedListener)
+proc teardown(manager: SendEventListenerManager) {.async.} =
+  await MessageSentEvent.dropListener(manager.brokerCtx, manager.sentListener)
```
Ivansete-status, line 81: I see the change everywhere and that's fine. Just to learn something, why `await` is needed now?
Shall we submit a separate PR to only bump nim-brokers?

waku/factory/conf_builder/waku_conf_builder.nim
```diff
@@ -719,6 +724,7 @@ proc build*(
     relayShardedPeerManagement: relayShardedPeerManagement,
     p2pReliability: builder.p2pReliability.get(false),
     wakuFlags: wakuFlags,
+    localStoragePath: builder.localStoragePath.get("./data"),
```
Ivansete-status: Maybe we should define a public `const` somewhere
```suggestion
    localStoragePath: builder.localStoragePath.get(DefaultStoragePath),
```

waku/factory/waku.nim
```diff
@@ -393,6 +395,12 @@ proc startWaku*(waku: ptr Waku): Future[Result[void, string]] {.async: (raises:
     else:
       waku[].dynamicBootstrapNodes = dynamicBootstrapNodesRes.get()
 
+  ## Initialize persistency singleton instance - we don't need the instance itself here,
+  ## but this ensures it's initialized before any store job starts.
+  discard Persistency.instance(conf.localStoragePath).valueOr:
```
Ivansete-status, line 399: Why is "the warm-up" needed? For some reason, I was expecting that each job would instantiate the first time.

waku/factory/waku.nim
```diff
@@ -523,6 +531,16 @@ proc stop*(waku: Waku): Future[Result[void, string]] {.async: (raises: []).} =
   try:
     waku.healthMonitor.setOverallHealth(HealthStatus.SHUTTING_DOWN)
 
+    let teardownRes = await Teardown.request()
+    if teardownRes.isErr():
+      error "Teardown request failed", error = teardownRes.error
+      return err("Teardown request failed: " & teardownRes.error)
+    else:
+      info "Teardown request completed successfully",
+        components = teardownRes.get().mapIt(it.component)
```
Ivansete-status: Maybe?
```suggestion
    (await Teardown.request()).isOkOr:
      error "Teardown request failed", error = error
      return err("Teardown request failed: " & error)
    
    info "Teardown request completed successfully",
      components = teardownRes.get().mapIt(it.component)
```

waku/persistency/backend_comm.nim
```diff
@@ -0,0 +1,161 @@
...
+
+# ── mt codec overloads for non-POD library types ────────────────────────
+#
+# brokers 2.0.0's mtMarshalValue / mtUnmarshalValue handle scalars, enums,
+# strings, seqs, arrays, and plain object/tuple recursion -- but they do
+# not see through `distinct seq[byte]`, nor do they know how to dispatch
+# a variant (case) object. We provide explicit overloads for the types
+# that appear in our broker payloads.
+
+proc mtMarshalValue*(
```
Ivansete-status, line 40: Shall we use a different proc names? Using the same as in nim-brokers might make it difficult to lookup later on which implementation is doing what. Applies elsewhere.

waku/persistency/backend_comm.nim
```diff
@@ -0,0 +1,161 @@
+## Cross-thread broker declarations for the persistency library.
+##
+## One EventBroker (writes, fire-and-forget) and five RequestBrokers (reads
+## + acked delete). All in (mt) mode: the listener / provider runs on the
+## job's storage thread; callers on any thread reach it via the shared
+## BrokerContext owned by the Job.
```
Ivansete-status, line 6: What does `Job` refers to? The "persistence job"?

waku/persistency/backend_comm.nim
```diff
@@ -0,0 +1,161 @@
+## Cross-thread broker declarations for the persistency library.
+##
+## One EventBroker (writes, fire-and-forget) and five RequestBrokers (reads
+## + acked delete). All in (mt) mode: the listener / provider runs on the
```
Ivansete-status:
```suggestion
## + acked delete). All in multi-thread (mt) mode: the listener / provider runs on the
```

waku/persistency/backend_thread.nim
```diff
@@ -0,0 +1,271 @@
...
+export broker_context, backend_comm
+
+logScope:
+  topics = "persistency thread"
+
+type
+  ReadyState = enum
+    rsPending = 0
+    rsReady = 1
+    rsError = 2
```
Ivansete-status: nit:
```suggestion
  ReadyState {.pure.} = enum
    Pending = 0
    Ready = 1
    Error = 2
```

waku/persistency/backend_thread.nim
```diff
@@ -0,0 +1,271 @@
...
+    shutdownFlag: Atomic[int]
+    readyFlag: Atomic[int] ## values from ReadyState
+    errBuf: array[256, char] ## last error message, NUL-terminated
+
+  StorageThread = Thread[ptr StorageThreadArg]
+
+# ── arg helpers ─────────────────────────────────────────────────────────
+
+proc allocArg(ctx: BrokerContext, dbPath: string): ptr StorageThreadArg =
+  result = cast[ptr StorageThreadArg](allocShared0(sizeof(StorageThreadArg)))
```
Ivansete-status: nit, let's not use result. Applies elsewhere.

waku/persistency/backend_thread.nim
```diff
@@ -0,0 +1,271 @@
...
+  result.ctx = ctx
+  result.dbPathLen = dbPath.len + 1
+  result.dbPath = cast[cstring](allocShared0(result.dbPathLen))
+  if dbPath.len > 0:
+    copyMem(result.dbPath, unsafeAddr dbPath[0], dbPath.len)
+
+proc freeArg(a: ptr StorageThreadArg) =
+  if a == nil:
```
Ivansete-status: nit, 
```suggestion
  if a.isNil():
```

waku/persistency/backend_thread.nim
```diff
@@ -0,0 +1,271 @@
...
+    ok(KvDelete(existed: r.get()))
+
+  # PersistEvent listener — fire-and-forget; we log on backend failure
+  # because the caller has no return channel.
+  proc onPersist(ev: PersistEvent): Future[void] {.async: (raises: []).} =
+    let r = backend.applyOps(ev.ops)
+    if r.isErr:
+      let pe: PersistencyError = r.error()
+      error "PersistEvent applyOps failed",
+        err = pe.msg, kind = $pe.kind, opCount = ev.ops.len
```
Ivansete-status: It is interesting to inform the main thread about that and also add a metric there.

----
Nit,
```suggestion
        error = pe.msg, kind = $pe.kind, opCount = ev.ops.len
```

waku/persistency/backend_thread.nim
```diff
@@ -0,0 +1,271 @@
...
+
+  let getRes = KvGet.setProvider(ctx, onGet)
+  if getRes.isErr:
+    return err("KvGet.setProvider: " & getRes.error())
```
Ivansete-status:
```suggestion
  KvGet.setProvider(ctx, onGet).isOkOr:
    return err("KvGet.setProvider: " & error)
```

waku/persistency/persistency.nim
```diff
@@ -0,0 +1,484 @@
...
+  ## Emit a batched persist event. The handler treats >1 ops as a single
+  ## BEGIN IMMEDIATE/COMMIT transaction (see backend_sqlite.applyOps).
+  ## ``waitFor`` is bounded by the mt emit's ``sendSync`` — a brief
+  ## channel push, not a wait on the listener.
+  var seqOps: seq[TxOp]
+  seqOps.setLen(ops.len)
+  for i in 0 ..< ops.len:
+    seqOps[i] = ops[i]
+  try:
+    waitFor PersistEvent.emit(t.context, PersistEvent(ops: seqOps))
```
Ivansete-status: We cannot use `waitFor` in so deep layer. That represents a block to the whole chronos loop. Is a fire-and-forget but in reality it blocks everything for a very short time, though. The worse is that this can lead to a future deadlocks.

waku/factory/waku.nim
```diff
@@ -523,6 +531,16 @@ proc stop*(waku: Waku): Future[Result[void, string]] {.async: (raises: []).} =
   try:
     waku.healthMonitor.setOverallHealth(HealthStatus.SHUTTING_DOWN)
 
+    let teardownRes = await Teardown.request()
```
Ivansete-status: It seems this magic is not needed in this case.
Instead, we could have something more direct, roughly:
```nim
let pRes = Persistency.instance()
if pRes.isOk():
  await pRes.get().shutdown()  # closes all jobs, joins storage threads
```

Ivansete-status (commented):
Thanks for the PR! 🙌 
Some comments so far.
The major suggestion is to perform the nim-brokers addition in a separate PR.
Cheers.

NagyZoltanPeter:
> Thanks for the PR! 🙌 Some comments so far. The major suggestion is to perform the nim-brokers addition in a separate PR. Cheers.

Thanks, will go through.
For the separate PR to introduce nim-brokers' dependency, IMO has no meaning, as this exact PR introduce the needs to have MT brokers. So I think its better to add here.

waku/persistency/persistency.nim
NagyZoltanPeter: True, and it should not be here... leftover.

apps/chat2/config_chat2.nim
NagyZoltanPeter: It was not compiling... hence fix it.

nix/deps.nix
NagyZoltanPeter: Transient dep of brokers.

tests/api/test_api_send.nim
NagyZoltanPeter: That's change coming from the diff of locally (logos-delivery) origin impl of EventBroker and the now used nim-brokers' implementation. Facade change, same functionality.

waku/factory/conf_builder/waku_conf_builder.nim
NagyZoltanPeter: Yess, good idea.

waku/factory/waku.nim
NagyZoltanPeter: This instantiate the Job orchestrator - Persistency instance, mainly needed to set up storage path.
It wont create any thread, nor create DB not even read. That's left for the specific user of the persistency.

waku/factory/waku.nim
NagyZoltanPeter: Yes, I may remove that, the idea of making Persistency a singleton came later and before I wanted a single shot decoupled tear down signal, that any component can action on.
It is made a MultiRequest to make it waitable.

waku/factory/waku.nim
NagyZoltanPeter: This suggestion is invalid code. But can be to with 
```nim
let teardownRes = (await ...).valueOr:
...
```

NagyZoltanPeter: Will address this

waku/persistency/backend_comm.nim
NagyZoltanPeter: Yes. Persistency instance orchestrates Jobs - each job has a broker context to make event/requests run on its own thread.

waku/persistency/backend_comm.nim
NagyZoltanPeter: Well it is explicit here. with `mixin` the broker marshaller can utilize user specific marshallers...

waku/persistency/backend_thread.nim
NagyZoltanPeter: That's a architectural question. Now persist events are fire and forget... will not block.
Let's say best effort store of last state. Otherwise, doing it on separate thread has no real meaning.

waku/factory/waku.nim
NagyZoltanPeter: Removed

waku/persistency/persistency.nim
NagyZoltanPeter: Reworked completely, all await now

waku/factory/conf_builder/waku_conf_builder.nim
NagyZoltanPeter: Done with best effort

apps/chat2/config_chat2.nim
Ivansete-status: ah okay, we didn't detect that issue because the daily CI failed to compile and then, the discord message wasn't delivered to us, I think @darshankabariya is reviewing that issue and make the daily-ci to always inform us.

Ivansete-status (approved):
LGTM! Thanks so much for it! 🙌

waku/factory/waku.nim
NagyZoltanPeter: Obsolate as Teardown request is removed.

PR #3885, opened 2026-05-15
ci: add daily rln simulator e2e workflow

darshankabariya:
We're moving logos delivery to the logos-dev fleet, where RLN is disabled — so no production traffic exercises RLN anymore. Proof generation, on-chain registration, rate limiting, and epoch recovery can silently regress with nothing to catch it until RLN is re-enabled there.

This PR automates that as a daily safety net: build wakunode2 on the runner, run logos-delivery-simulator, and assert the full RLN flow (keystore → on-chain register → gossipsub propagation → per-epoch rate limit → epoch recovery).


fcecin (approved):

.github/workflows/ci-rln-simulator.yml
```diff
@@ -0,0 +1,249 @@
...
+# Why this exists: logos-dev runs with RLN disabled, so there is no
+# production traffic exercising RLN. Until RLN is enabled there, this is
+# the only end-to-end coverage of the RLN + zerokit path.
+#
+# The image is built ON the runner and tested ON the same runner, so the
+# AVX-512 portability issue in container-image.yml does not apply here.
+
+on:
+  schedule:
+    - cron: '0 4 * * *'        # daily, 30 min before ci-daily.yml
```
Ivansete-status: Why not just have one single daily task that starts anything else?

.github/workflows/ci-rln-simulator.yml
```diff
@@ -0,0 +1,249 @@
...
+      epoch_sec:
+        description: 'RLN_RELAY_EPOCH_SEC (large enough that a burst cannot straddle an epoch)'
+        type: string
+        default: '120'
+
+env:
+  NPROC: 2
+  MAKEFLAGS: "-j2"
+  NIM_VERSION: '2.2.4'
+  NIMBLE_VERSION: '0.22.3'
```
Ivansete-status, line 56: Comment aside from the PR. Is interesting having a separate source of truth. Similar to this: https://github.com/logos-messaging/nim-ffi/blob/master/versions.env

Ivansete-status (approved):
LGTM, thanks for it!
The only suggestion is to only have one cron task that calls others. This simulator task can then be either called manually, or called by the parent daily task.

.github/workflows/ci-rln-simulator.yml
darshankabariya: thanks, fix in f1f37dad7ae78f9d93795f17ad69d3fb3a97cfb4

.github/workflows/ci-rln-simulator.yml
darshankabariya: thanks. fix in f1f37dad7ae78f9d93795f17ad69d3fb3a97cfb4

PR #3886, opened 2026-05-18
start basic reliable channel folder

### LINKED ISSUE #3853: create basic reliable channel components
- Create a clear layer/folder (if doesn't exist) where to add the channel-api. One possible candidate is `waku/api/channel_api/`.
- Inside `waku/api/channel_api`, have a module with all the procs and components that we need. For example, have the reliable-channel entity. - For now, the `channel_api` modules need to consume SDS and messaging-api implementation.

Notice that segmentation and rate limit will be implemented separately.

See the [Reliable Channel Spec](https://lip.logos.co/messaging/standards/application/reliable-channel-api.html) for further reference.

Ivansete-status:
This PR:
- Is inspired by pair session with @jazzz , 🙌 !

- Adds a very basic folder structure of reliable channel and its components.
The main suggested new folder is `channels` and inside, it contains `segmentation`, `rate-limit-manager`, `sds`, and `encryption`. This is just structural PR but no functionality is implemented yet.

In separate PRs:
- Properly accommodate all the broker events and properly refactor the MessagingClient/MessagingAPI implementation.
- Properly implement `channels/scalable_data_sync/scalable_data_sync.nim` and `channels/scalable_data_sync/sds_persistence.nim` modules.
- Implement unit tests for reliable channel.

closes https://github.com/logos-messaging/logos-delivery/issues/3853

### DIFF
--- a/channels/encryption/encryption.nim
+++ b/channels/encryption/encryption.nim
@@ -17,5 +17,11 @@ type
     encrypt*: EncryptFn
     decrypt*: DecryptFn
 
-proc isConfigured*(h: EncryptionHook): bool =
-  not h.encrypt.isNil() and not h.decrypt.isNil()
+proc isConfigured*(self: EncryptionHook): bool =
+  not self.encrypt.isNil() and not self.decrypt.isNil()
+
+proc send*(self: EncryptionHook, payload: seq[byte]): seq[byte] =
+  ## Stage 4 of the outgoing pipeline (segmentation -> sds -> rate_limit_manager -> encryption).
+  ## For now: passthrough — return the payload unencrypted.
+  ## TODO: invoke `self.encrypt` when configured and surface errors.
+  return payload
--- a/channels/encryption/encryption.nim
+++ b/channels/encryption/encryption.nim
@@ -1,0 +1,21 @@
+## Optional encryption hook for the Reliable Channel API.
+##
+## Applied per-segment after SDS processing on outgoing, and before
+## SDS processing on incoming. No specific scheme is mandated.
+##
+## See: https://lip.logos.co/messaging/raw/reliable-channel-api.html
+
+import results
+
+type
+  EncryptionError* = object of CatchableError
+
+  EncryptFn* = proc(payload: seq[byte]): Result[seq[byte], string] {.gcsafe, raises: [].}
+  DecryptFn* = proc(payload: seq[byte]): Result[seq[byte], string] {.gcsafe, raises: [].}
+
+  EncryptionHook* = object
+    encrypt*: EncryptFn
+    decrypt*: DecryptFn
+
+proc isConfigured*(h: EncryptionHook): bool =
+  not h.encrypt.isNil() and not h.decrypt.isNil()
--- a/channels/encryption/encryption.nim
+++ b/channels/encryption/encryption.nim
@@ -20,8 +20,14 @@ type
 proc isConfigured*(self: EncryptionHook): bool =
   not self.encrypt.isNil() and not self.decrypt.isNil()
 
-proc send*(self: EncryptionHook, payload: seq[byte]): seq[byte] =
+proc encrypt*(self: EncryptionHook, payload: seq[byte]): seq[byte] =
   ## Stage 4 of the outgoing pipeline (segmentation -> sds -> rate_limit_manager -> encryption).
   ## For now: passthrough — return the payload unencrypted.
-  ## TODO: invoke `self.encrypt` when configured and surface errors.
+  ## TODO: invoke the configured `EncryptFn` when present and surface errors.
+  return payload
+
+proc decrypt*(self: EncryptionHook, payload: seq[byte]): seq[byte] =
+  ## Stage 1 of the incoming pipeline (decryption -> sds -> reassemble -> emit).
+  ## For now: passthrough — return the payload as-is.
+  ## TODO: invoke the configured `DecryptFn` when present and surface errors.
   return payload
--- a/channels/encryption/encryption.nim
+++ b/channels/encryption/encryption.nim
@@ -1,33 +1,25 @@
-## Optional encryption hook for the Reliable Channel API.
+## Optional encryption hooks for the Reliable Channel API.
+##
+## Modelled as `RequestBroker`s: the broker pattern lets the channel
+## delegate work to a provider that may live in any module without
+## introducing a direct dependency. If no provider is registered the
+## broker returns an error, so installing the noop providers from
+## `noop_encryption` is required when the application does not want
+## actual encryption.
 ##
 ## Applied per-segment after SDS processing on outgoing, and before
 ## SDS processing on incoming. No specific scheme is mandated.
 ##
 ## See: https://lip.logos.co/messaging/raw/reliable-channel-api.html
 
-import results
+import waku/common/broker/request_broker
 
-type
-  EncryptionError* = object of CatchableError
+export request_broker
 
-  EncryptFn* = proc(payload: seq[byte]): Result[seq[byte], string] {.gcsafe, raises: [].}
-  DecryptFn* = proc(payload: seq[byte]): Result[seq[byte], string] {.gcsafe, raises: [].}
+RequestBroker:
+  type Encrypt* = seq[byte]
+  proc signature*(payload: seq[byte]): Future[Result[Encrypt, string]] {.async.}
 
-  EncryptionHook* = object
-    encrypt*: EncryptFn
-    decrypt*: DecryptFn
-
-proc isConfigured*(self: EncryptionHook): bool =
-  not self.encrypt.isNil() and not self.decrypt.isNil()
-
-proc encrypt*(self: EncryptionHook, payload: seq[byte]): seq[byte] =
-  ## Stage 4 of the outgoing pipeline (segmentation -> sds -> rate_limit_manager -> encryption).
-  ## For now: passthrough — return the payload unencrypted.
-  ## TODO: invoke the configured `EncryptFn` when present and surface errors.
-  return payload
-
-proc decrypt*(self: EncryptionHook, payload: seq[byte]): seq[byte] =
-  ## Stage 1 of the incoming pipeline (decryption -> sds -> reassemble -> emit).
-  ## For now: passthrough — return the payload as-is.
-  ## TODO: invoke the configured `DecryptFn` when present and surface errors.
-  return payload
+RequestBroker:
+  type Decrypt* = seq[byte]
+  proc signature*(payload: seq[byte]): Future[Result[Decrypt, string]] {.async.}
--- a/channels/encryption/encryption.nim
+++ b/channels/encryption/encryption.nim
@@ -12,7 +12,7 @@
 ##
 ## See: https://lip.logos.co/messaging/raw/reliable-channel-api.html
 
-import waku/common/broker/request_broker
+import brokers/request_broker
 
 export request_broker
 
--- a/channels/encryption/noop_encryption.nim
+++ b/channels/encryption/noop_encryption.nim
@@ -1,14 +1,18 @@
-## No-op encryption hook. Useful as the default when no encryption is
-## configured by the application.
+## No-op encryption providers. Install these when the application does
+## not want actual encryption so the `Encrypt` / `Decrypt` brokers have
+## something to dispatch to.
 
 import results
+import chronos
 import ./encryption
 
-proc noopEncrypt(payload: seq[byte]): Result[seq[byte], string] {.gcsafe, raises: [].} =
-  return ok(payload)
+proc setNoopEncryption*() =
+  discard Encrypt.setProvider(
+    proc(payload: seq[byte]): Future[Result[Encrypt, string]] {.async.} =
+      return ok(Encrypt(payload))
+  )
 
-proc noopDecrypt(payload: seq[byte]): Result[seq[byte], string] {.gcsafe, raises: [].} =
-  return ok(payload)
-
-proc init*(T: typedesc[EncryptionHook]): T =
-  return T(encrypt: noopEncrypt, decrypt: noopDecrypt)
+  discard Decrypt.setProvider(
+    proc(payload: seq[byte]): Future[Result[Decrypt, string]] {.async.} =
+      return ok(Decrypt(payload))
+  )
--- a/channels/encryption/noop_encryption.nim
+++ b/channels/encryption/noop_encryption.nim
@@ -1,0 +1,14 @@
+## No-op encryption hook. Useful as the default when no encryption is
+## configured by the application.
+
+import results
+import ./encryption
+
+proc noopEncrypt(payload: seq[byte]): Result[seq[byte], string] {.gcsafe, raises: [].} =
+  return ok(payload)
+
+proc noopDecrypt(payload: seq[byte]): Result[seq[byte], string] {.gcsafe, raises: [].} =
+  return ok(payload)
+
+proc init*(T: typedesc[EncryptionHook]): T =
+  return T(encrypt: noopEncrypt, decrypt: noopDecrypt)
--- a/channels/events.nim
+++ b/channels/events.nim
@@ -1,27 +1,23 @@
 ## Reliable Channel event types emitted to API consumers.
+##
+## Lifecycle events for individual segments (sent / propagated / errored)
+## are the same as the network-level ones the DeliveryService already
+## emits — `requestId` is shared across layers — so we just re-export
+## `waku/events/message_events` and avoid declaring duplicates.
+##
+## Only the channel-level `MessageReceivedEvent` carries data that has
+## no analogue in the lower layer (reassembled application payload,
+## senderId, channelId), so it lives here.
 
-import waku/api/types
+import waku/events/message_events as waku_message_events
+import waku/common/broker/event_broker
 
 import ./types as channel_types
 
-export types, channel_types
+export waku_message_events, channel_types, event_broker
 
-type
-  MessageReceivedEvent* = object
+EventBroker:
+  type ChannelMessageReceivedEvent* = object
     channelId*: ChannelId
     senderId*: SdsParticipantID
     payload*: seq[byte]
-
-  MessageSentEvent* = object
-    requestId*: RequestId
-
-  MessageDeliveredEvent* = object
-    requestId*: RequestId
-
-  MessageSendErrorEvent* = object
-    requestId*: RequestId
-    reason*: string
-
-  MessageDeliveryErrorEvent* = object
-    requestId*: RequestId
-    reason*: string
--- a/channels/events.nim
+++ b/channels/events.nim
@@ -10,7 +10,7 @@
 ## senderId, channelId), so it lives here.
 
 import waku/events/message_events as waku_message_events
-import waku/common/broker/event_broker
+import brokers/event_broker
 
 import ./types as channel_types
 
--- a/channels/events.nim
+++ b/channels/events.nim
@@ -13,15 +13,15 @@ type
     payload*: seq[byte]
 
   MessageSentEvent* = object
-    requestId*: ReliableRequestId
+    requestId*: RequestId
 
   MessageDeliveredEvent* = object
-    requestId*: ReliableRequestId
+    requestId*: RequestId
 
   MessageSendErrorEvent* = object
-    requestId*: ReliableRequestId
+    requestId*: RequestId
     reason*: string
 
   MessageDeliveryErrorEvent* = object
-    requestId*: ReliableRequestId
+    requestId*: RequestId
     reason*: string
--- a/channels/events.nim
+++ b/channels/events.nim
@@ -1,0 +1,26 @@
+## Reliable Channel event types emitted to API consumers.
+
+import ./scalable_data_sync/scalable_data_sync
+
+type
+  ChannelId* = SdsChannelID
+  RequestId* = string
+
+  MessageReceivedEvent* = object
+    channelId*: ChannelId
+    senderId*: SdsParticipantID
+    payload*: seq[byte]
+
+  MessageSentEvent* = object
+    requestId*: RequestId
+
+  MessageDeliveredEvent* = object
+    requestId*: RequestId
+
+  MessageSendErrorEvent* = object
+    requestId*: RequestId
+    reason*: string
+
+  MessageDeliveryErrorEvent* = object
+    requestId*: RequestId
+    reason*: string
--- a/channels/events.nim
+++ b/channels/events.nim
@@ -1,26 +1,27 @@
 ## Reliable Channel event types emitted to API consumers.
 
-import ./scalable_data_sync/scalable_data_sync
+import waku/api/types
+
+import ./types as channel_types
+
+export types, channel_types
 
 type
-  ChannelId* = SdsChannelID
-  RequestId* = string
-
   MessageReceivedEvent* = object
     channelId*: ChannelId
     senderId*: SdsParticipantID
     payload*: seq[byte]
 
   MessageSentEvent* = object
-    requestId*: RequestId
+    requestId*: ReliableRequestId
 
   MessageDeliveredEvent* = object
-    requestId*: RequestId
+    requestId*: ReliableRequestId
 
   MessageSendErrorEvent* = object
-    requestId*: RequestId
+    requestId*: ReliableRequestId
     reason*: string
 
   MessageDeliveryErrorEvent* = object
-    requestId*: RequestId
+    requestId*: ReliableRequestId
     reason*: string
--- a/channels/rate_limit_manager/rate_limit_manager.nim
+++ b/channels/rate_limit_manager/rate_limit_manager.nim
@@ -9,22 +9,25 @@ import std/times
 import sds/message
 import waku/common/broker/event_broker
 
-export message, event_broker
+export event_broker, broker_context
+export message.SdsChannelID
 
 const
   DefaultEpochPeriodSec* = 600
   DefaultMessagesPerEpoch* = 1
 
 EventBroker:
-  ## Emitted by `enqueueToSend` carrying the batch of SDS messages that
-  ## may now leave the rate limiter and continue down the outgoing
-  ## pipeline (encryption -> dispatch).
+  ## Emitted by `enqueueToSend` carrying the batch of opaque message
+  ## blobs that may now leave the rate limiter and continue down the
+  ## outgoing pipeline (encryption -> dispatch). Bytes only: the rate
+  ## limiter is intentionally agnostic of SDS, so anything serialisable
+  ## can flow through it.
   ##
   ## `channelId` lets listeners filter to their own channel, since all
   ## reliable channels share the underlying Waku node's broker context.
   type ReadyToSendEvent* = object
     channelId*: SdsChannelID
-    msgs*: seq[SdsMessage]
+    msgs*: seq[seq[byte]]
 
 type
   RateLimitConfig* = object
@@ -33,13 +36,17 @@ type
 
   RateLimitManager* = ref object
     config*: RateLimitConfig
-    queue*: seq[SdsMessage]
+    queue*: seq[seq[byte]]
     currentEpochStart*: Time
     sentInCurrentEpoch*: int
     channelId*: SdsChannelID ## tag for the emitted `ReadyToSendEvent`
+    brokerCtx: BrokerContext
 
 proc new*(
-    T: type RateLimitManager, config: RateLimitConfig, channelId: SdsChannelID
+    T: type RateLimitManager,
+    config: RateLimitConfig,
+    channelId: SdsChannelID,
+    brokerCtx: BrokerContext = globalBrokerContext(),
 ): T =
   return T(
     config: config,
@@ -47,9 +54,10 @@ proc new*(
     currentEpochStart: getTime(),
     sentInCurrentEpoch: 0,
     channelId: channelId,
+    brokerCtx: brokerCtx,
   )
 
-proc enqueueToSend*(self: RateLimitManager, msg: SdsMessage) =
+proc enqueueToSend*(self: RateLimitManager, msg: seq[byte]) =
   ## Stage 3 of the outgoing pipeline (segmentation -> sds -> rate_limit_manager -> encryption).
   ##
   ## For now: enqueue the message and immediately dequeue the full
@@ -62,10 +70,10 @@ proc enqueueToSend*(self: RateLimitManager, msg: SdsMessage) =
   self.queue = @[]
 
   ReadyToSendEvent.emit(
-    globalBrokerContext(), ReadyToSendEvent(channelId: self.channelId, msgs: ready)
+    self.brokerCtx, ReadyToSendEvent(channelId: self.channelId, msgs: ready)
   )
 
-proc dequeueReady*(self: RateLimitManager): seq[SdsMessage] =
+proc dequeueReady*(self: RateLimitManager): seq[seq[byte]] =
   ## Returns the set of queued messages that may be dispatched now
   ## without exceeding the configured rate limit. Advances epoch
   ## bookkeeping as needed.
--- a/channels/rate_limit_manager/rate_limit_manager.nim
+++ b/channels/rate_limit_manager/rate_limit_manager.nim
@@ -3,11 +3,15 @@
 ## Tracks messages sent per RLN epoch and delays dispatch when the
 ## limit is approached, ensuring RLN compliance on enforcing relays.
 ##
+## For the skeleton this is a pass-through: messages are immediately
+## released as ready-to-send. Real epoch budgeting will be added later.
+##
 ## See: https://lip.logos.co/messaging/raw/reliable-channel-api.html
 
 import std/times
-import sds/message
-import waku/common/broker/event_broker
+import message
+import brokers/event_broker
+import brokers/broker_context
 
 export event_broker, broker_context
 export message.SdsChannelID
@@ -31,6 +35,7 @@ EventBroker:
 
 type
   RateLimitConfig* = object
+    enabled*: bool ## spec: rate limiting opt-in; SHOULD be true when RLN active
     epochPeriodSec*: int
     messagesPerEpoch*: int
 
@@ -58,25 +63,16 @@ proc new*(
   )
 
 proc enqueueToSend*(self: RateLimitManager, msg: seq[byte]) =
-  ## Stage 3 of the outgoing pipeline (segmentation -> sds -> rate_limit_manager -> encryption).
-  ##
-  ## For now: enqueue the message and immediately dequeue the full
-  ## queue, emitting `ReadyToSendEvent` with the batch ready to be sent.
-  ## TODO: park `msg` on `self.queue` and only emit when the RLN-epoch
-  ## budget allows; advance epoch bookkeeping on `dequeueReady`.
-  self.queue.add(msg)
-
-  let ready = self.queue
-  self.queue = @[]
-
+  ## Skeleton behaviour: enqueue and immediately release as a single
+  ## ready batch. Real per-epoch budgeting will park messages on
+  ## `self.queue` and emit only when the budget allows.
   ReadyToSendEvent.emit(
-    self.brokerCtx, ReadyToSendEvent(channelId: self.channelId, msgs: ready)
+    self.brokerCtx, ReadyToSendEvent(channelId: self.channelId, msgs: @[msg])
   )
 
 proc dequeueReady*(self: RateLimitManager): seq[seq[byte]] =
   ## Returns the set of queued messages that may be dispatched now
-  ## without exceeding the configured rate limit. Advances epoch
-  ## bookkeeping as needed.
+  ## without exceeding the configured rate limit.
   discard
 
 proc resetEpoch*(self: RateLimitManager) =
--- a/channels/rate_limit_manager/rate_limit_manager.nim
+++ b/channels/rate_limit_manager/rate_limit_manager.nim
@@ -1,0 +1,43 @@
+## Rate Limit Manager for the Reliable Channel API.
+##
+## Tracks messages sent per RLN epoch and delays dispatch when the
+## limit is approached, ensuring RLN compliance on enforcing relays.
+##
+## See: https://lip.logos.co/messaging/raw/reliable-channel-api.html
+
+import std/times
+import sds/message
+
+export message
+
+const
+  DefaultEpochPeriodSec* = 600
+  DefaultMessagesPerEpoch* = 1
+
+type
+  RateLimitConfig* = object
+    epochPeriodSec*: int
+    messagesPerEpoch*: int
+
+  RateLimitManager* = ref object
+    config*: RateLimitConfig
+    queue*: seq[SdsMessage]
+    currentEpochStart*: Time
+    sentInCurrentEpoch*: int
+
+proc new*(T: type RateLimitManager, config: RateLimitConfig): T =
+  return T(config: config, queue: @[], currentEpochStart: getTime(), sentInCurrentEpoch: 0)
+
+proc enqueue*(mgr: RateLimitManager, msg: SdsMessage) =
+  ## Append an SDS message to the pending dispatch queue.
+  mgr.queue.add(msg)
+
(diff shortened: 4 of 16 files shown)


channels/reliable_channel.nim
```diff
@@ -0,0 +1,181 @@
...
+    pendingRequests: @[],
+  )
+
+proc onReadyToSend*(self: ReliableChannel, msgs: seq[SdsMessage]) =
+  ## Tail of the outgoing pipeline. Invoked from the `ReadyToSendEvent`
+  ## listener once `rate_limit_manager` releases a batch of SDS messages:
+  ##
+  ##   ... -> rate_limit_manager -> [encryption] -> dispatch
+  for m in msgs:
+    let wireBytes = self.encryption.encrypt(m.encode())
```
Ivansete-status: We will need to revisit which fields need to be encrypted so that we allow proper routing without having for force-brute try to decrypt everything.

fcecin (approved):

channels/encryption/encryption.nim
```diff
@@ -0,0 +1,33 @@
...
+## SDS processing on incoming. No specific scheme is mandated.
+##
+## See: https://lip.logos.co/messaging/raw/reliable-channel-api.html
+
+import results
+
+type
+  EncryptionError* = object of CatchableError
+
+  EncryptFn* = proc(payload: seq[byte]): Result[seq[byte], string] {.gcsafe, raises: [].}
```
NagyZoltanPeter: This can be:


```suggestion
  RequestBroker:
    type Encrypt = seq[byte]
    proc signature*(payload: seq[byte]): Future[Result[Encrypt, string]] {.async.}
```

channels/encryption/encryption.nim
```diff
@@ -0,0 +1,33 @@
...
+  DecryptFn* = proc(payload: seq[byte]): Result[seq[byte], string] {.gcsafe, raises: [].}
```
NagyZoltanPeter: ...and...

```suggestion
  RequestBroker:
    type Decrypt = seq[byte]
    proc signature*(payload: seq[byte]): Future[Result[Decrypt, string]] {.async.}
```

channels/encryption/noop_encryption.nim
```diff
@@ -0,0 +1,14 @@
...
+import ./encryption
+
+proc noopEncrypt(payload: seq[byte]): Result[seq[byte], string] {.gcsafe, raises: [].} =
+  return ok(payload)
+
+proc noopDecrypt(payload: seq[byte]): Result[seq[byte], string] {.gcsafe, raises: [].} =
+  return ok(payload)
+
+proc init*(T: typedesc[EncryptionHook]): T =
+  return T(encrypt: noopEncrypt, decrypt: noopDecrypt)
```
NagyZoltanPeter: ...and follows...

```suggestion
proc setNoopEncryption() = 
  Encrypt.setProvider(proc(payload: seq[byte]): Future[Result[Encrypt, string]] {.async.} = 
    return ok(payload)
  )

  Decrypt.setProvider(proc(payload: seq[byte]): Future[Result[Decrypt, string]] {.async.} = 
    return ok(payload)
  )
```

channels/rate_limit_manager/rate_limit_manager.nim
```diff
@@ -0,0 +1,76 @@
...
+  ## queue, emitting `ReadyToSendEvent` with the batch ready to be sent.
+  ## TODO: park `msg` on `self.queue` and only emit when the RLN-epoch
+  ## budget allows; advance epoch bookkeeping on `dequeueReady`.
+  self.queue.add(msg)
+
+  let ready = self.queue
+  self.queue = @[]
+
+  ReadyToSendEvent.emit(
+    globalBrokerContext(), ReadyToSendEvent(channelId: self.channelId, msgs: ready)
```
NagyZoltanPeter: instead using `globalBrokerContext` here, the value of it can be captured and stored in `RateLimitManager` => during tests and multiple ffi context case the source of broker context truth shall come from initialization phase.

channels/scalable_data_sync/sds_persistence.nim
```diff
@@ -0,0 +1,26 @@
...
+
+method storeOutgoing*(
+    self: SdsPersistence, msg: SdsMessage
+) {.base.} =
+  discard
+
+method markAcknowledged*(
+    self: SdsPersistence, messageId: SdsMessageID
+) {.base.} =
+  discard
```
NagyZoltanPeter: Is this a public facing part? I though such should come internally from SDS machinery.

NagyZoltanPeter: But maybe this surface in incomplete in the sense of current SDS module.
@darshankabariya

channels/events.nim
```diff
@@ -0,0 +1,27 @@
...
+  MessageDeliveredEvent* = object
+    requestId*: ReliableRequestId
+
+  MessageSendErrorEvent* = object
+    requestId*: ReliableRequestId
+    reason*: string
+
+  MessageDeliveryErrorEvent* = object
+    requestId*: ReliableRequestId
+    reason*: string
```
NagyZoltanPeter: We should brainstorm these namings though, might be misleading to overlap with Messaging API events.
Also it is a question if to emit those events when reliable channels in use..... This is a spec question though.

channels/reliable_channel.nim
```diff
@@ -0,0 +1,185 @@
...
+
+proc onMessageReceived*(self: ReliableChannel, wakuMsg: WakuMessage) =
+  ## Ingress pipeline made visible:
+  ##
+  ##   WakuMessage -> ReliablePayload -> decrypt -> sds -> reassemble -> emit
+  ##
+  ## Invoked from the waku `MessageReceivedEvent` listener after the
+  ## inbound `WakuMessage` has been filtered to this channel's
+  ## `contentTopic`. Each stage is a minimal stub for now.
+  let inWakuMsg: WakuMessage = wakuMsg
```
NagyZoltanPeter: If I could suggest, to implement it as a listener to the Waku's (or rather MessagingAPI's) MessageReceivedEvent.
To decouple the two layer and not to let lower layer MessagingAPI exactly know about upper ReliableChannel layer.
WDYT? 
cc: @fcecin

channels/reliable_channel_manager.nim
```diff
@@ -0,0 +1,141 @@
...
+    return err("unknown channel: " & channelId)
+  manager.channels.del(channelId)
+  return ok()
+
+proc send*(
+    manager: ReliableChannelManager,
+    channelId: ChannelId,
+    appPayload: seq[byte],
+    ephemeral: bool = false,
+): Result[ReliableRequestId, string] =
```
NagyZoltanPeter: This looks the public facing API of all.
I think we need to think of a mechanism to shadow/hide MessagingAPI level send when ReliableChannel is in use. WDYT?
Or... is it a valid scenario to send most of my messages through channels but some I may want to send as row low level payloads?

NagyZoltanPeter (commented):
I left some comments to think of ;-)

channels/events.nim
fcecin: Sorry to barge in here but we are bailed out by scoping the type by e.g. import "as", e.g. channels.MessageSentEvent vs. messaging.MessageSentEvent (I think); nim-brokers dispatches on the actual type, not type strings/names AFAICT. Also solves human disambiguation by using the module import prefix. (not touching here on the consumer-producer relationship between events are different levels).

channels/reliable_channel.nim
fcecin: Yes, so the natural way to think of it is that ReliableChannels is implemented entirely using the Messaging layer (which will contain a MessagingClient root type, etc. as specced by Igor, but that comes in my PR -- so you can use whatever is in the MAPI now instead to implement channels -- channels will be the first module merged with the correct layering, which is easier since it is new; and for now it uses whatever we have there, and I'll rewire it later).

Now in *practice* (if we want to take it up with Igor later) some "triangle-shape" interactions could make sense. That is, the ReliableChannels layer *could* have the Waku/Kernel/Core layer be a direct dependency -- that is, the channels layer (layer 3) implementation could call the API of the kernel layer (layer 1) since it IS an API, so that's fine, but the question then is whether we want that in general (not really) and whether it is the smart thing to do in some cases (possible). For example, you don't need to create a MessagingMessage type at the messaging layer (layer 2) to substitute for the WakuMessage API type which is a core type. And there may be other examples of actual functions that you may want to call directly from the core. I don't have a strong opinion either way, but as long as we're having implementations wired to APIs *only*, improving on this is easy -- you just expand the API at the layer directly beneath you to supply what you need so you can abandon making a layer 3 ---> layer 1 API direct dial.

channels/events.nim
NagyZoltanPeter: This is exact! But Nim cannot enforce it (wish for namespace kind... )

channels/reliable_channel_manager.nim
fcecin: The logos-delivery library will give the ability to callers to operate at any layer they want. They should be able to instantiate a ReliableChannel *and* a MessagingClient *and* even a Waku if we end up exposing that via a logos delivery kernel.h. They can operate at all levels simultaneously -- they *shouldn't*, but they can, because we will have public methods to do that.

So when a ReliableChannel is created, it creates a MessagingClient for *itself*, internally, to get its job done. The MessagingClient in turn creates a Waku *for itself*. The only way the library user can access the MessagingClient that the ReliableChannel created directly is if we give a channels layer API that returns the MessagingClient associated with that ReliableChannel. Either that or the application can create their own MessagingClient and manage that separately. That's the API-spec decision. 

That's what happens when we create a library that exposes multiple levels of service at once.

channels/events.nim
fcecin: > This is exact! But Nim cannot enforce it (wish for namespace kind... )

Right, I mean, in a module the compiler will complain if it cannot disambiguate, so you will have to do e.g. `import xxx as yyy` if you don't like the default dot-disambiguator it gives you (e.g. "import ...../channels" allows you to do channels.MyType in the code instead of MyType and it already works. Unless I'm missing something (possible, even likely).

channels/reliable_channel.nim
igor-sirotin: I'm also a bit confused, this doesn't seem to be in specs 🤔

channels/reliable_channel.nim
```diff
@@ -0,0 +1,185 @@
...
+
+const Lip173Meta* = "LIP173"
+  ## Wire-level marker for the Reliable Channel layer. A `WakuMessage`
+  ## whose `meta` field does not equal these bytes is not addressed to
+  ## this layer and is silently dropped on ingress.
+
+type
+  ReliablePayload* = object
+    channelId*: ChannelId
+    payload*: seq[byte]
```
igor-sirotin: What is this for? I think we got rid of it in the spec?

channels/reliable_channel.nim
```diff
@@ -0,0 +1,185 @@
...
+    segmentation*: SegmentationHandler
+    sdsHandler*: SdsHandler
+    rateLimit*: RateLimitManager
+    encryption*: EncryptionHook
+
+    requestIds*: Table[ReliableRequestId, seq[RequestId]]
+      ## Maps each reliable-channel-level (parent) `ReliableRequestId`
+      ## returned to the caller of `send` to the set of delivery-service
+      ## `RequestId`s it fanned out into (one per dispatched segment).
+    pendingRequests*: seq[tuple[parent: ReliableRequestId, ephemeral: bool]]
```
igor-sirotin: I think all of these fields should be private. With getters & setter where needed.
Otherwise the user can change any field and break the internals.

channels/reliable_channel.nim
```diff
@@ -0,0 +1,185 @@
...
+
+proc new*(
+    T: type ReliableChannel,
+    deliveryService: DeliveryService,
+    channelId: ChannelId,
+    contentTopic: ContentTopic,
+    senderId: SdsParticipantID,
+    segmentation: SegmentationHandler,
+    sdsHandler: SdsHandler,
+    rateLimit: RateLimitManager,
```
igor-sirotin: So a user would need to specify these 3 handlers, which are implementation details? 
Or is this `ReliableChannel` private, not exported to user?

igor-sirotin: Also, this doesn't seem to follow the spec

channels/scalable_data_sync/scalable_data_sync.nim
```diff
@@ -0,0 +1,55 @@
...
+    config*: SdsConfig
+
+proc new*(T: type SdsHandler, config: SdsConfig): T =
+  return T(config: config)
+
+proc wrapOutgoing*(
+    self: SdsHandler,
+    channelId: SdsChannelID,
+    senderId: SdsParticipantID,
+    segment: SegmentMessageProto,
```
igor-sirotin: SDS should be agnostic of segmentation. Payload is just bytes

channels/rate_limit_manager/rate_limit_manager.nim
```diff
@@ -0,0 +1,76 @@
...
+EventBroker:
+  ## Emitted by `enqueueToSend` carrying the batch of SDS messages that
+  ## may now leave the rate limiter and continue down the outgoing
+  ## pipeline (encryption -> dispatch).
+  ##
+  ## `channelId` lets listeners filter to their own channel, since all
+  ## reliable channels share the underlying Waku node's broker context.
+  type ReadyToSendEvent* = object
+    channelId*: SdsChannelID
+    msgs*: seq[SdsMessage]
```
igor-sirotin: Same here, `RateLimitManager` should be SDS-agnostic. Just messages/bytes.

channels/encryption/encryption.nim
```diff
@@ -0,0 +1,33 @@
...
+
+  EncryptionHook* = object
+    encrypt*: EncryptFn
+    decrypt*: DecryptFn
+
+proc isConfigured*(self: EncryptionHook): bool =
+  not self.encrypt.isNil() and not self.decrypt.isNil()
```
igor-sirotin: I think if one of those 2 is defined, and the other is not — it should be an invalid `EncryptionHook`.
It should then fail at configuration stage, if such encryption is provided.

igor-sirotin: And then you won't need `isConfigured` method. If `nil` -> replace with noop implementation.

channels/types.nim
```diff
@@ -0,0 +1,33 @@
...
+import waku/utils/requests as request_utils
+
+import ./scalable_data_sync/scalable_data_sync
+
+export scalable_data_sync
+
+type
+  ChannelId* = SdsChannelID
+
+  ReliableRequestId* = distinct string
```
igor-sirotin: I thought we agreed to use jus `RequestId`, or not? 😄

channels/reliable_channel_manager.nim
igor-sirotin: I was actually seeing it like this (as I described in https://github.com/logos-messaging/logos-delivery/issues/3865):
```nim
# First create a MessagingClient
let client = MessagingClient.new(preset = TheWakuNetwork, mode = Edge)
await client.start()

# Create many channels, use same MessagingClient
let chat1 = client.createReliableChannel("channel-id-1", contentTopic="/myapp/1/chat1/proto")
let chat2 = client.createReliableChannel("channel-id-2", contentTopic="/myapp/1/chat2/proto")
```

So:
- User is allowed to freely mix the usage of any layers. It shouldn't break.
- Reliable Channels require a`MessagingClient`
- Reliable Channels are **created from** a `MessagingClient`. And pointers to them are kept inside this `MessagingClient` instance. 

Then such `ReliableChannelManager` would be a private thing held by `MessagingClient`. And it should not be the user-facing class. Only `ReliableChannel` itself should be public.

channels/reliable_channel.nim
igor-sirotin: UPD: I see, you intended this as private object. But I think we should have a public `ReliableChannel`, see https://github.com/logos-messaging/logos-delivery/pull/3886/changes#r3262304270. And actually, this is what spec defines — a `ReliableChannel` public object

channels/reliable_channel_manager.nim
fcecin: I think that means the Reliable Channels feature goes into the Messaging layer. Now we have two layers (waku/kernel, and messaging) instead of three (waku/kernel, messaging, channels). My read of the diagrams was just wrong.

Channels is then an optional feature of the messaging client, so there's no "ReliableChannelsManager" root object (or something to that effect) (EDIT: with its own independent lifetime) in a separate layer that's analogous to MessagingClient and Waku.

That actually makes more sense.

cc: @darshankabariya

igor-sirotin: @fcecin yeah when writing the ticket I was thinking if it's fair to call it a separate layer. 
Ended up convinced that it is 😄

It's still an optional layer that can be ignored by the client. It introduces another object for the client to communicate with. And it contains quite some logic in it. But indeed, it depends on the `MessagingClient`.

But it doesn't matter much if we call it a layer really

igor-sirotin: > Reliable Channels are created from a `MessagingClient`. And pointers to them are kept inside this `MessagingClient` instance.

I think I was wrong here. `MessagingClient` doesn't need to remember all channels. It only needs to subscribe it to own events, then return it to the client.

channels/reliable_channel.nim
Ivansete-status: Yes, I think is better to properly decouple layers as Zoltán mentioned and that way I think it will be closer to specs. I'll revisit that asap.

channels/reliable_channel_manager.nim
Ivansete-status: Thanks for the comments.
We decided to merge kernel (libwaku) and messaging (liblogosdelivery) libraries into one, ok.

Now that we are starting to introduce ReliableChannels, I think it should generate its own separate library. That way, we will have one single `send` on each library, and resolve Zoltán's concern.

Status will consume `libwaku+liblogosdelivery`.

LogosChat will consume `libreliablechannels`. ( or the names that you prefer .)

@NagyZoltanPeter , @fcecin , @igor-sirotin - wdyt about that lib separation?

NagyZoltanPeter: I start getting loss here.
If messaging has the interface to initialize a new reliable-channel-id (chat) than it sounds ReliableChannel is not usable standalone... thus it's against this concept of separate it into it's own library.
Separate library would assume the opposite ownership model. MessagingAPI -> ReliableChannelMgr vs. ReliableChannelMgr -> MessagingAPI.

My real question was, whether to allow use MessagingApi->send on content_topic used for an already initialized channel?
> side note: As Igor stated, as being open source, we cannot prevent anybody to do so in a forked delivery lib.
> So might be better to be prepared for such scenario of sending exploited payloads on channel-content-topics.

For me having yet another library, sounds, not solving problems, but generates more.

My proposal than is rather divide by naming:

~ Kernel_RelayPublish / Kernel_LightpushPublish <- after merge
~ Messaging_send() <- this is the current MAPI simple send interface
~ Channel_send() <- reliable channel API

Also events must be distinguishable, thus user can decide which to subscribe.
MessageSentEvent / MessageReceivedEvent vs. ChannelMessageSent / ChannelMessageReceived

_We can debate on naming, just wanted to demonstrate the idea._

If we are talking about libraries, it also means ffi... for which we don't support hierarchical API (yet).

channels/reliable_channel.nim
jazzz: I suggested it in our Pairing session. Its a lot easier to remove, then to add in; If its still there and it doesn't contain any additional fields it can be collapsed out.

"Concrete layers for concrete protocols" ensures that protocol can handle growth during implementation.

channels/scalable_data_sync/scalable_data_sync.nim
jazzz: SDS is agnostic of payload, however a Reliable channel is an explicit pipeline. The code should reflect that. 

To that effect Type safety is desired for protocol clarity. Its not modular, it follows a deterministic parsing tree. The function can encode the segment into bytes inside, and be spec compliant

channels/reliable_channel.nim
igor-sirotin: I'm happy with distinct names. But `ReliablePayload` just doesn't make sense to me, because payload is not "reliable". Maybe `ReliableChannelPayload` then?

channels/reliable_channel_manager.nim
fcecin: Guys, in my ongoing refactor work (lift messaging, separate messaging and waku core) I am:

* Assuming we are shipping one actual library (liblogosdelivery) which packs all code. There is no reason to ship two distinct libraries. The only reason libwaku.so is separate is because we have marked it for its own clean, discrete, independent deprecation and deletion cycle (which has to be negotiated, I assume).
* This library packs a set of layers, which are high-level, formal components with boundaries, contracts, invariants, etc.
* Not assuming where channels go (we can put them into its own separate layer, or implement it inside the messaging layer -- both will work; it's just a question of whether we decide they are separately mountable and startable as a standalone internal meta-component, that is, a layer, or not).
* Not assuming what the C API is and not changing it unnecessarily. We control precisely what we expose in the C API at all times, irrespective of what Nim code we have behind it.
* Not deleting libwaku.so yet, and not changing its behavior (unless you tell me to delete it, then it is gone).
* Not changing config one bit -- that work will be resolvable once this lands (each layer can have its own config, or we can make it work with the "union", "global" config object we already have. Now that we have formally-modeled layers, we can actually make sense when discussing a layered config system.

About this new _formal_ concept being introduced, of layers:

* Aren't just a prefix/naming convention, although we do generally use name `prefixes_` to distinguish in e.g. the otherwise flat list of functions that a C API is.
* Can be mounted and unmounted individually in the library by controlling the existence of the zero-or-one-instance singleton Nim root object for the layer (e.g. `Waku` vs `MessagingClient`).
* Have their own `nim-brokers` facade which the layer's implementation types cooperate to populate upon mounting the layer and starting the root object, and that is unplugged when stopping the root object and unmounting the layer.

In other words, we can _guarantee_ that e.g. Waku, the most fundamental layer, does not depend on any other layer, and is individually mountable. So when you have e.g. a dedicated intrastructure node that is not doing any messaging, you can just mount that single layer and you will be guaranteed to not pull any messaging functionality. Each layer has its own independent lifetime and interface, and they cooperate to get things done, e.g. the Messaging layer calls the Waku (kernel, core) layer, not the other way around.

EDIT: We have decided to not go this route for now, and will instead do a [simpler reorg](https://github.com/logos-messaging/logos-delivery/pull/3886#issuecomment-4554233672).

channels/reliable_channel.nim
jazzz: Even better.

jazzz:
Adding context: from the pairing session that @Ivansete-status  and I had, I think it makes sense to do a quick lightning implementation to discover key gaps in the protocol and then make spec compliant after words. 

There were some missing gaps that surfaced once we tried to integrate it into an actually use case, persuing this further would help uncover them all. Specifically [reliable-channel-api spec](https://github.com/logos-co/logos-lips/blob/master/docs/messaging/raw/reliable-channel-api.md) defines the API but does not define key payload parsing/transformation requirements. 

My suggestion would be:
1) Lighting implementation: get something that works end-to-end. 
2) I'll gather what You all like and what needs to change into a ReliableChannel protocol specification. 
3) Update impl into final form. 
4) Integration into libchat:InboxV2.
5) Profit

channels/rate_limit_manager/rate_limit_manager.nim
Ivansete-status: Thanks, I think I will suggest, in upcoming commits, to make `ReliableChannelManager` the broker ctx owner for this layer.

channels/reliable_channel_manager.nim
```diff
@@ -0,0 +1,174 @@
...
+    conf: WakuNodeConf,
+    brokerCtx: BrokerContext = globalBrokerContext(),
+): Future[Result[T, string]] {.async.} =
+  ## TODO !! The proper ownership chain is:
+  ## ReliableChannelManager -> DeliveryService (MessagingClient) -> Waku (Kernel/Protocols) -> WakuNode,
+  ## and this will be implemented in the future. For now, `createNode`
+  ## is called here to get a DeliveryService instance, and the WakuNode is immediately discarded.
+  ## This is a temporary workaround to get the API
+
+  let waku = ?(await createNode(conf))
```
igor-sirotin, line 44: Can't we just pass the `DeliveryService` in as arg then? 🙂

channels/reliable_channel.nim
```diff
@@ -0,0 +1,221 @@
...
+import ./encryption/encryption
+
+export
+  delivery_service, send_service, events, segmentation, scalable_data_sync,
+  rate_limit_manager, encryption
+
+const Lip173Meta* = "LIP173"
+  ## Wire-level marker for the Reliable Channel layer. A `WakuMessage`
+  ## whose `meta` field does not equal these bytes is not addressed to
+  ## this layer and is silently dropped on ingress.
```
igor-sirotin: I would like to see this discussed in spec first, before adding to implementation.

1. What are the consequences of having this as unencrypted metadata?
2. The goal of this field is to simplify the incoming pipeline, but:
    - This mechanism doesn't allow selecting 1 designated reliable channel. All instances of `ReliableChannel` will try to decrypt a message. Find the match by trying all — it's ok IMO, at least this is what I expected for the first implementation. But then `Lip173Meta` doesn't really help much with it.
    - An attacker can spam us with random data tagged with `Lip173Meta`, forcing us to try to decrypt it. Though it's the same case without `Lip173Meta` 😄

channels/reliable_channel_manager.nim
```diff
@@ -0,0 +1,174 @@
...
+    manager.brokerCtx,
+    proc(
+        evt: waku_message_events.MessageReceivedEvent
+    ): Future[void] {.async: (raises: []).} =
+      if string.fromBytes(evt.message.meta) != Lip173Meta:
+        return
+
+      for chn in manager.channels.values:
+        if chn.getContentTopic() == evt.message.contentTopic:
+          await chn.onMessageReceived(evt.message.payload)
```
igor-sirotin: There's only one possible match in all `manager.channels`. So we should break the loop as soon as match found.

channels/reliable_channel.nim
```diff
@@ -0,0 +1,221 @@
...
+  channelId: ChannelId
+  contentTopic: ContentTopic
+  senderId: SdsParticipantID
+  rng: ref HmacDrbgContext
+  segmentation: SegmentationHandler
+  sdsHandler: SdsHandler
+  rateLimit: RateLimitManager
+
+  requestIds: Table[RequestId, seq[RequestId]]
+  pendingRequests: seq[tuple[parent: RequestId, ephemeral: bool]]
```
igor-sirotin, line 59: This can be simplified by storing extra information (like `ephemeral` flag) in table
```suggestion
  pendingRequests: Table[RequestId, bool]
```

channels/reliable_channel.nim
```diff
@@ -0,0 +1,221 @@
...
+    pendingRequests: @[],
+    brokerCtx: brokerCtx,
+  )
+
+proc onReadyToSend*(
+    self: ReliableChannel, msgs: seq[seq[byte]]
+) {.async: (raises: []).} =
+  ## Tail of the outgoing pipeline. Invoked from the `ReadyToSendEvent`
+  ## listener once `rate_limit_manager` releases a batch of opaque
+  ## blobs (already-encoded SDS messages):
```
igor-sirotin, line 81: This way `onReadyToSend` becomes a public method of `ReliableChannel`, which it shouldn't be.

I see 2 ways to avoid it:
- Listen to the signal right in `ReliableChannel` (I like this)
- Send from `ReliableChannelManager`

channels/reliable_channel.nim
```diff
@@ -0,0 +1,221 @@
...
+    self.pendingRequests.delete(0)
+
+    ## TODO: revisit which fields of the SDS message must be encrypted.
+    ## Encrypting the whole encoded blob forces every receiver to attempt
+    ## decryption before it can route, which breaks selective dispatch.
+    ## Leave routing metadata (channelId, causal-history references) in
+    ## clear and encrypt only the application payload.
+    let encRes = await Encrypt.request(m)
+    let wireBytes =
+      if encRes.isOk():
```
igor-sirotin: Why not fail the operation if encryption failed?

channels/reliable_channel.nim
```diff
@@ -0,0 +1,221 @@
...
+    let sdsBytes = self.sdsHandler.wrapOutgoing(
+      self.channelId, self.senderId, segment.encode()
+    ).valueOr:
+      return err("SDS wrap failed: " & error)
+    self.pendingRequests.add((parent: parentReqId, ephemeral: ephemeral))
+    self.rateLimit.enqueueToSend(sdsBytes)
+
+  return ok(parentReqId)
+
+proc onMessageReceived*(
```
igor-sirotin: Same here, this method should be private

channels/reliable_channel.nim
```diff
@@ -0,0 +1,221 @@
...
+  ##
+  ## The ReliableChannelManager already validated LIP173 on the WakuMessage and
+  ## stripped the wire framing, so the channel only sees the raw
+  ## payload bytes for itself.
+
+  ## Notice that the following "request" is implemented implicitly as a broker call to
+  ## the `Decrypt` request broker.
+  let decRes = await Decrypt.request(payload)
+  let plaintext: seq[byte] =
+    if decRes.isOk():
```
igor-sirotin: If decryiption failed — halt the operation?

channels/scalable_data_sync/scalable_data_sync.nim
igor-sirotin: Yes, but I think that Reliable Channel API shouldn't even know of `SegmentMessageProto`.
Anyway, it's resolved now, I'm happy 😄

channels/reliable_channel.nim
```diff
@@ -0,0 +1,221 @@
...
+      seq[byte](decRes.get())
+    else:
+      payload
+
+  let unwrapped = self.sdsHandler.handleIncoming(plaintext)
+  if unwrapped.isErr():
+    return
+
+  let segment = SegmentMessageProto.decode(unwrapped.get().content)
```
igor-sirotin: `SegmentMessageProto` is a segmentation internal detail, we shouldn't operate with it directly in `ReliableChannel`.

igor-sirotin (approved):
This PR is about initial objects definition and code structure — I'm happy with what I see, thank you! 👍 The comments I left inline are minor, and go beyond the goal of this PR.

channels/reliable_channel_manager.nim
Ivansete-status: Yes that's an option too. In fact, I was thinking about just passing a rel-channels conf. Something like:
```nim
proc new*(
    T: type ReliableChannelManager,
    conf: ReliableChannelManagerConf,
    brokerCtx: BrokerContext = globalBrokerContext(),
): Future[Result[T, string]] {.async.} =
```
Then, internally build the `DeliveryService`/`MessagingClient`.
wdyt?

igor-sirotin: @Ivansete-status my point was that that we never build `MessagingClient` internally at all. Always use the provided one. 

But your comment here states that it's a temporary solution, and I guess it's for compilation success, that's why I suggested to pass a `DeliveryService` right away — it would compile. If it's for testing, then fine to keep as is.

channels/reliable_channel.nim
Ivansete-status: Started the discussion in https://github.com/logos-co/logos-lips/pull/348
Also, the slug attribute seems not very good. Instead, I suggest using a wire version string that can be adjusted manually within the spec itself.
Adjusted this PR in the following: e274189e3422627b33bf954e23782d78d617a825

channels/reliable_channel_manager.nim
Ivansete-status: Thanks! Addressed in 4035cd036f5f4200bf8c3c40d910eca480f1aedd

NagyZoltanPeter (commented):
@Ivansete-status I think we lost some focus within here.
I don't really get why does ReliableChannelManager calls createNode?
It seems we promoted ReliableChannelManager to the lib root class.

Maybe I'm wrong but from those many discussion I distilled the following structure.

MessagingClient (library root) - creates node
                                                    - creates Kernel (Waku in old terms)
                                                    - creates DeliveryService (represents MessagingAPI)
                                                    - can create ReliableChannelManager (layer root of ReliableChannel - will orchestrate channels within)

So it is up to users decision which layer to use.
I think that should be one place where configuration is taken and instantiate the node itself (basically Kernel root object until we don't separate the impl from interface).
So the master orchestrator and first hand client facing object is MessagingClient.

It's up to us how we manifest it into FFI context.

appologize is my comment seems in-medias-res and I may have missed some part of the conversation. If so, please update me.
@fcecin @igor-sirotin @darshankabariya

fcecin:
> @Ivansete-status I think we lost some focus within here. I don't really get why does ReliableChannelManager calls createNode? It seems we promoted ReliableChannelManager to the lib root class.

Sorry guys, I'm late with this; we have rerouted in the architecture aspect.

In my current (much shorter) reorg PR:

* Root lib FFI context object: still always `Waku` (we can rename it later to e.g. LogosDelivery)
* Component objects inside `Waku`:
  * `WakuNode` (and all other kernel objects)
  * `MessagingClient` (can be nil; if not nil, messaging is enabled)
  * `ReliableChannelManager` (can be nil; if not nil, reliable channels are enabled)

The ReliableChannelManager can also be put inside the MessagingClient if it becomes an implementation detail of the messaging layer, instead of its own layer.

Whatever code sits on top of this (FFI code or Nim application code) just does `new Waku`, then optionally `mountMessagingClient` / `mountReliableChannelManager`, then `Waku.start()` just knows what to do.

Ivansete-status:
Thanks for the comments @NagyZoltanPeter ! And you are absolutely right that it doesn't look 100% nice now. This PR is just a very initial step towards introducing `ReliableChannelManager`.

I think the following diagram explains very well the final structure that we want to have (extracted from [lift API issue](https://github.com/logos-messaging/logos-delivery/issues/3865))

image

So, the "root library" concept fundamentally depends on the user, as you mentioned.

| # | User                               | Root library object    |
|---|------------------------------------|------------------------|
| 1 | Reliable App Dev, e.g., logoschat | ReliableChannelManager |
| 2 | Plain App Dev, e.g., Status | MessagingClient/DeliveryService |
| 3 | Node Operator, e.g., DST | Kernel/Waku |

With that, the layer `#1` will internally instantiate `#2` which in turn will instantiate `#3`.
The layer `#2` only instantiates `#3` and knows nothing about `#1`.

Each layer may have its own config object. The most internal the layer is, the less opinionated the config is, being the Kernel/Waku config layer (current WakuNodeConf) the one that brings full flexibility to the NodeOperator/DST.

Ivansete-status:
Thanks @fcecin for the comments!
I think the best way of ensuring the diagram I linked above is by having a composition from top to down. In your proposal the `Waku`/`Kernel` has notion about `ReliableChannelManager` but this is not what we want. We want top-down ownership, i.e., 

**`ReliableChannelManager` ➡︎ _uses_ ➡︎ `MessagingClient` ➡︎ _uses_ ➡︎ `Waku`/`Kernel`**

fcecin:
> Thanks @fcecin for the comments! I think the best way of ensuring the diagram I linked above is by having a composition from top to down. In your proposal the `Waku`/`Kernel` has notion about `ReliableChannelManager` but this is not what we want. We want top-down ownership, i.e.,
> 
> **`ReliableChannelManager` ➡︎ _uses_ ➡︎ `MessagingClient` ➡︎ _uses_ ➡︎ `Waku`/`Kernel`**

Yes, that's what I said. `Waku` is not the kernel. It's the entire system. It's just named "Waku" because in the past that's all that existed. We can rename it in my PR.

The kernel objects don't access the messaging objects, and the messaging objects don't access the reliable channel objects (if that's how we are going to structure it). So the arrows in the diagram (who knows who, who uses who) are respected.

So the requirement is literally fulfilled by that mechanic. Unless we want a different specific mechanic (EDIT: e.g. specific object composition hierarchy -- which type literally has which type as their field), which then I will need the explicit code structure that we want here spelled out. We can sort that out in reviews.

channels/reliable_channel.nim
Ivansete-status: The current proposal is indeed too poor. I will tackle that in a separate PR.

channels/reliable_channel.nim
Ivansete-status: Addressed in 0b8cd51866bbc6bdca111e8bd1370ecebce3ef20

channels/reliable_channel.nim
Ivansete-status: Addressed in 0b8cd51866bbc6bdca111e8bd1370ecebce3ef20

NagyZoltanPeter:
> Thanks @fcecin for the comments! I think the best way of ensuring the diagram I linked above is by having a composition from top to down. In your proposal the `Waku`/`Kernel` has notion about `ReliableChannelManager` but this is not what we want. We want top-down ownership, i.e.,
> 
> **`ReliableChannelManager` ➡︎ _uses_ ➡︎ `MessagingClient` ➡︎ _uses_ ➡︎ `Waku`/`Kernel`**

Ok, here is my problem.
That diagram is a dependency diagram and not ownership diagram - at least how I understood it.

And I suggest to separate internal dependency from DevEx.

In order a developer should be clearly understand the roles of each layers we shall not expose such an ownership matrix.
Hence I vote for an orchestrator - library face - class (we can call it LogosDelivery (as @fcecin described called `Waku`), or even MessagingClient). But from DevEx point of view as a user of the library shall be able to clearly distinguish layers and how can I access them. So if I have an first hand library face, that returns me the proper layer accessor I can choose easily and naturally.

```cpp
class LogosDelivery {
public:
   Result<void> init(config);
   Result<ReliableChannelManafeg> getReliableChannelManager(...);
   Result<DeliveryService> getDeliveryService(...);
   Result<DeliveryKernel> getDeliveryKernel(...);
}
```
This is just a showcasing example, namings are just arbitrary only for understanding the concept.

I think there are clean advantages to have a separate orchestrator class at the top that can not just give access to layers but can drive upon configuration (modes, roles, etc.) and can also manage lifetimes.
WDYT?

channels/reliable_channel.nim
Ivansete-status: Addressed in e4caa0be3d662dc5394114e785839f1c09e75c00

Ivansete-status:
Sorry @fcecin , I misunderstood you.

I agree that we need to have such main-obj or orchestrator type, and I think we can follow @NagyZoltanPeter 's suggestion in that regard, i.e., keep a `LogosDelivery` type as the main library object that handles everything else. Let's see and coordinate for upcoming PRs.

Thanks both! 🥳

channels/reliable_channel.nim
Ivansete-status: Addressed in 8d95fcb98a0ebc83e2da264f1935d4a29f24143c

fcecin:
> Sorry @fcecin , I misunderstood you.
> 
> I agree that we need to have such main-obj or orchestrator type, and I think we can follow @NagyZoltanPeter 's suggestion in that regard, i.e., keep a `LogosDelivery` type as the main library object that handles everything else. Let's see and coordinate for upcoming PRs.
> 
> Thanks both! 🥳

Right this somewhat my fault because "a reliable channel having a messaging client having a waku/kernel" was my idea also when I started the massive layering PR. But we won't do that massive refactor, so we should keep doing what we are doing: keeping a single `FFIContext[SomeLibraryObject]` at all times, and that object can "nil" certain objects -- controlled by whoever is using that object. Which is what we do internally in Waku today to decide what internal protocols are mounted (relay.isNil()? etc.); same idea, and the layering is implicit in the dependency chain between the objects, not necessarily in composition hierarchies. (We can change this later also).

channels/reliable_channel.nim
Ivansete-status: Addressed in 4648d1597461f101fa554fba862c885ec643b3c0

channels/reliable_channel.nim
Ivansete-status: Addressed in the following: https://github.com/logos-messaging/logos-delivery/pull/3914

PR #3889, opened 2026-05-19
fix: real getNodeInfo Version in Nix/lgpm builds

### LINKED ISSUE #3884: getNodeInfo Version returns "n/a"; also expose Commit
`logoscore call delivery_module getNodeInfo Version --json` currently returns `"n/a"` on builds produced by both `lgpm`-installed releases and `nix build '.#install-portable'`. Expected: real `liblogosdelivery` semver / tag.

Additionally, expose a `Commit` node info ID returning the build's git commit SHA — the dashboard needs both fields per-node (matches wakunode2's existing reporting).

Refs: status-im/infra-logos#4

igor-sirotin:
Closes #3884

# Description

`getNodeInfo Version` returned `"n/a"` on Nix/lgpm builds because the Nix derivation never passed `-d:git_version`.

- Nix: derive `Version` from `waku.nimble` + flake short commit; pass the `git_version` define
- CI `version-check`: PR asserts `waku.nimble >= latest tag`; tag push asserts tag `==` nimble

Refs: status-im/infra-logos#4

### DIFF
--- a/waku.nimble
+++ b/waku.nimble
@@ -4,7 +4,7 @@ import os
 mode = ScriptMode.Verbose
 
 ### Package
-version = "0.37.4"
+version = "0.38.1"
 author = "Status Research & Development GmbH"
 description = "Waku, Private P2P Messaging for Resource-Restricted Devices"
 license = "MIT or Apache License 2.0"
--- a/.github/ISSUE_TEMPLATE/prepare_release.md
+++ b/.github/ISSUE_TEMPLATE/prepare_release.md
@@ -18,7 +18,10 @@ For detailed info on the release process refer to https://github.com/logos-messa
 All items below are to be completed by the owner of the given release.
 
 - [ ] Create release branch with major and minor only ( e.g. release/v0.X ) if it doesn't exist.
-- [ ] Update the `version` field in `waku.nimble` to match the release version (e.g. `version = "0.X.0"`).
+- [ ] Update the `version` field in `waku.nimble` to match the release version (e.g. `version = "0.X.0"`) **and merge it before assigning any tag**. The version must be bumped first because:
+  - the `version check` CI job fails any PR whose `waku.nimble` is behind its nearest ancestor tag;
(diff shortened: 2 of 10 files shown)


Ivansete-status (approved):
LGTM! Thanks for it! 🙌

waku.nimble
```diff
@@ -4,7 +4,7 @@ import os
 mode = ScriptMode.Verbose
 
 ### Package
-version = "0.37.4"
+version = "0.38.1"
```
NagyZoltanPeter, line 7: Updating this would be good to have part of release process, if not already.
Maybe with some automation...

NagyZoltanPeter (approved):
LGTM.

fcecin (approved):
LGTM. Claude found a nit, not sure if relevant:

```
  🟢 Low — Nix version string format diverges from Makefile builds

  - Makefile (Makefile:115): git describe --abbrev=6 --always --tags → v0.38.1-4-g52e980 (leading v, commit-count, 6-char sha).
  - This PR (Nix): 0.38.1-g52e9807 (no v, no commit-count, 7-char sha).

  Functionally harmless — nothing parses git_version (it's used as a metrics label, a REST text response, and equality-compared in tests).
  But two binaries of the same project will self-report different version shapes, which can confuse dashboards/humans correlating Nix-built
  vs make-built nodes. Consider matching the v-prefixed describe-style format for consistency.
```

waku.nimble
igor-sirotin: Actually it was already there in the `prepare_release.md` 🤷‍♂️
But I've now added a job to CI release, [it's in this PR](https://github.com/logos-messaging/logos-delivery/pull/3889/changes#diff-7953647cb607addbd9ee1c22b4a7cf5234c7629bae8aaee279ea80775c7a6072)

igor-sirotin:
>  Nix version string format diverges from Makefile builds

@fcecin Indeed, good catch. 

I'll add the `v` prefix and fix sha length, but it seems it's nearly impossible (or at least not worth it) to get it fully consistent with what `git describe` returns — the commits count will always be missing. So I think we'll just have to live with it 🤷

PR #3894, opened 2026-05-20
Update and Improve READMEs

stubbsta:
## Description
The README needs to be updated to reflect recent changes to the Makefile for the nimble build system.
Some links and wording also need to be updated or improved.

## Changes
- README.md
- waku/README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,10 @@ For more details see the [source code](waku/README.md)
 
 These instructions are generic. For more detailed instructions, see the source code above.
 
+Recommended and tested toolchain versions (these are installed when you follow the build instructions below):
+- Nim 2.2.4
+- Nimble 0.22.3
+
 ### Prerequisites
 
 The standard developer tools, including a C compiler, GNU Make, Bash, and Git.
@@ -30,10 +34,6 @@ Rust:
 ```bash
 curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh
 ```
-
-Recommended and tested toolchain versions:
-- Nim 2.2.4
-- Nimble 0.22.3
 
 ### Wakunode
 
--- a/README.md
+++ b/README.md
@@ -5,9 +5,9 @@
 This repository implements a set of libp2p protocols aimed to bring
 private communications.
 
-- Nim implementation of [these specs](https://github.com/vacp2p/rfc-index/tree/main/waku).
+- Nim implementation of [these specs](https://github.com/logos-co/logos-lips/tree/master/docs/messaging).
 - C library that exposes the implemented protocols.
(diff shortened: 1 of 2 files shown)


stubbsta:
This PR is part of the Maintenance milestone https://github.com/logos-messaging/logos-delivery/issues/3686

README.md
```diff
@@ -19,22 +19,26 @@ These instructions are generic. For more detailed instructions, see the source c
...
 You'll also need an installation of Rust and its toolchain (specifically `rustc` and `cargo`).
 The easiest way to install these, is using `rustup`:
 
+Rust:
 ```bash
 curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh
 ```
 
+Recommended and tested toolchain versions:
+- Nim 2.2.4
```
igor-sirotin: Is installing `Nim` a requirement? I think it's fetched automatically by Nimble?

README.md
```diff
@@ -19,22 +19,26 @@ These instructions are generic. For more detailed instructions, see the source c
...
+- Nimble 0.22.3
+
 ### Wakunode
 
 ```bash
-# The first `make` invocation will update all Git submodules.
-# You'll run `make update` after each `git pull` in the future to keep those submodules updated.
```
igor-sirotin, line 37: If `make update` is not used anymore, should we also remove it from `Makefile`?

igor-sirotin (approved):

README.md
stubbsta: No, I don't think it's required to be installed, this section is to provide info on which tools and their versions are recommended. I guess incase the dev is using some other version of these, @Ivansete-status is this correct?

README.md
stubbsta: I plan to update the `Makefile` in a separate PR to remove `make update` as well as some other simplifications.

README.md
Ivansete-status: Yes exactly, is a recommendation but users may have another nim version.

Ivansete-status (approved):
LGTM! Thanks for it! 🙌

README.md
igor-sirotin: Yes, but this section is called "Prerequisites" 😄  which I read as "requirements". Or am I wrong?

stubbsta: indeed :) I'll move it out of the prerequisites section

stubbsta: updated in https://github.com/logos-messaging/logos-delivery/pull/3894/commits/d50e51250734b6fa7090eddc3af8becafdc28e0a

README.md
stubbsta: see https://github.com/logos-messaging/logos-delivery/pull/3897

PR #3897, opened 2026-05-21
Remove makefile target update

stubbsta:
## Description
The Makefile target `update` is no longer required because we moved from nimbus to nimble. The updating of submodules is no applicable.
This target needs to be removed.

## Changes
Makefile
BearSSL.mk
Nat.mk


stubbsta:
Updates to the README are done in https://github.com/logos-messaging/logos-delivery/pull/3894

stubbsta:
This PR is part of the Maintenance milestone https://github.com/logos-messaging/logos-delivery/issues/3686

scripts/install_nim.sh
```diff
@@ -17,26 +17,36 @@ if [ -z "${NIM_VERSION}" ]; then
...
+
+# 2. Already installed at our expected location from a previous run, but not on PATH.
+#    Re-link binaries into ~/.nimble/bin.
+if [ -f "${NIM_DEST}/lib/system.nim" ]; then
+  echo "Nim ${NIM_VERSION} already installed at ${NIM_DEST}, re-linking binaries."
+  mkdir -p "${HOME}/.nimble/bin"
+  for bin_path in "${NIM_DEST}/bin/"*; do
+    ln -sf "${bin_path}" "${HOME}/.nimble/bin/$(basename "${bin_path}")"
+  done
   exit 0
```
stubbsta, line 40: Hi @Ivansete-status, I originally used your changes to this file from this PR https://github.com/logos-messaging/logos-delivery/pull/3861/changes#diff-9aff0306d244936911b4d1ef813e322c6d9e0bbe1c4addfe7f96b47c26783e6e, 
but then the `build-docker-image` ci job breaks.

I think this is the reason:
The ci job already installs Nim via `setup-nim-action`, but only skips a second installation attempt if the specific path `~/.nim/nim-2.2.4/lib/system.nim` exists and this doesn't exist in the build-docker-image job.

Does this new Path check make sense and cover the scenario that you originally tried to fix?

darshankabariya (commented):
Thanks! I have tested locally on mac, and it works fine.
Just one comment: why is this duplicate line. 

Screenshot 2026-05-25 at 12 23 08

darshankabariya (approved):

scripts/install_nim.sh
Ivansete-status: That movement looks correct to me, thanks!

Makefile
```diff
@@ -93,15 +92,14 @@ REQUIRED_NIM_VERSION    := $(shell grep -E '^const RequiredNimVersion\s*=' waku.
 REQUIRED_NIMBLE_VERSION := $(shell grep -E '^const RequiredNimbleVersion\s*=' waku.nimble | grep -oE '"[0-9]+\.[0-9]+\.[0-9]+"' | tr -d '"')
 
 install-nim:
+ifneq ($(detected_OS),Windows)
```
Ivansete-status, line 95: How it Nim and Nimble installed on windows then?

Ivansete-status (commented):
Looks great so far, thanks!
Something left to ensure is how the Nim and Nimble installation occurs on Windows.

fcecin (approved):
LGTM

stubbsta:
> Thanks! I have tested locally on mac, and it works fine. Just one comment: why is this duplicate line.
> 

Hi @darshankabariya, could you pull latest and please try to build on your mac again to see if the repeated lines are gone and everything still works?

NagyZoltanPeter (approved):
Great work!

Makefile
stubbsta: hmmm, okay there are some things to point out, but TL;DR:
I'll remove this check and add a check for Windows in the install scripts so it downloads the nim `zip` file instead of `tar` file to check for pre-installed nim or nimble.

The long version:
I got a Windows build errored running `install-nim.sh` / `install-nimble.sh`, which I assumed was a PowerShell incompatibility issue, but then I saw that Windows actually builds under MSYS2 bash, so bash isn't the issue.

Checking the CI build job workflow, I see that nim/nimble are installed for each OS before make runs (via setup-nim-action + nimble install) and for local Windows via scripts/build_windows.sh (builds Nim, then runs make wakunode2). The README's Windows instructions point to the `build_windows` script, not bare `make wakunode2`.

stubbsta: Upon further inspection, local build on Windows is broken because the build_windows.sh file has not yet been updated for nimble. I'll tackle this in a separate PR: https://github.com/logos-messaging/logos-delivery/issues/3911

PR #3868, opened 2026-05-12
feat: migrate to zerokit v2.0.2

darshankabariya:
Bumps zerokit v0.9 → v2.0.1. Migrates the FFI from v0.9's raw byte buffers to v2.0's safer-ffi typed handles. Built against the stateless feature — the Merkle tree lives on-chain, so we no longer carry zerokit's pmtree-ft local tree.

What changed

- rln_interface.nim rewritten against the v2.0.1 typed-handle API (CFr, Vec_CFr, CResult*)
- Proof gen/verify wired through generateRlnProofWithWitness / verifyRlnProof
- verify_with_roots replaces verify_rln_proof — matches the contract's recentRoots ring buffer
- Wire format: 290-byte versioned encoding (layout in the rln_interface.nim header)
- Dead v0.9 buffer-API code removed

Checklist

- [x] make librln builds against vendor/zerokit at v2.0.1
- [x] Nim type-checks and links against the new librln
- [x] Unit tests pass (make test)
- [x] Roundtrip: membershipKeyGen → register → generateRlnProofWithWitness → verifyRlnProof 
- [x] Daily simulator E2E CI until RLN is enabled in logos.dev ( #3885 )

After merged 
- [ ] 48h soak on Logos fleet — no regressions in non-RLN paths
- [ ] Mix protocol functionality tested on Logos fleet (also uses zerokit 2.0.1)

Related issue
- https://github.com/logos-messaging/logos-delivery/pull/3885

PR #3901, opened 2026-05-21
test : add memory-leak guard for the v2.0 FFI procs

darshankabariya:
With the v2.0 FFI, our RLN wrappers free the memory they allocate by hand. That works today, but it's easy to break in a future change (raised by @Ivansete-status in #3868 pr [review](https://github.com/logos-messaging/logos-delivery/pull/3868#discussion_r3274602538) ), so this adds a guard test. It runs each new proc in a loop — poseidon ×10k and generate→verify ×50 — checks the result stays correct (catching double-free / use-after-free), and watches memory usage to catch leaks.


darshankabariya:
Closed, it's handled at the zeroKit side.

PR #3904, opened 2026-05-22
Separate core (Waku) and MessagingClient using nim-brokers

### LINKED ISSUE #3865: Lift Messaging API out of `Waku` into its own layer
## Summary

The Messaging API is described in the [spec](https://github.com/logos-messaging/specs/blob/master/standards/application/messaging-api.md) as an *opinionated layer above the kernel protocols* (RELAY, LIGHTPUSH, FILTER, STORE, discovery). It owns reliability strategies, automatic re-subscriptions, store-based catch-up, peer management, and the Messaging API event surface.

In code, those routines are not a layer above the node — they live **inside** `Waku`, are constructed in `Waku.new()`, started in `startWaku`, and stopped in `Waku.stop()`. The Messaging API surface itself (`waku/api/api.nim`) is just thin free procs over `Waku.deliveryService`, and `liblogosdelivery` is a thin C-FFI shim over those procs. There is no Nim-level Messaging-API object — it cannot be used as a layer from Nim, only as C bindings.

This is a layering inversion of what the spec describes.

### Messaging API routines live inside `Waku`

`DeliveryService` is what implements Messaging API's background work:
- [`send_service.nim`](https://github.com/logos-messaging/logos-delivery/blob/27ae07adaaea7beeae02cea9f8647b18cd9fb482/waku/node/delivery_service/send_service/send_service.nim) — store-node confirmations of published messages, retries.
- [`recv_service.nim`](https://github.com/logos-messaging/logos-delivery/blob/27ae07adaaea7beeae02cea9f8647b18cd9fb482/waku/node/delivery_service/recv_service/recv_service.nim) — store-based catch-up of missed messages, de-dup.
- [`subscription_manager.nim`](https://github.com/logos-messaging/logos-delivery/blob/27ae07adaaea7beeae02cea9f8647b18cd9fb482/waku/node/delivery_service/subscription_manager.nim) — per-shard subscription state, Messaging API auto-subscribe semantics.

`Waku` directly owns `deliveryService` and runs its lifecycle:

- `deliveryService*: DeliveryService` is a field of `Waku`:

  https://github.com/logos-messaging/logos-delivery/blob/27ae07adaaea7beeae02cea9f8647b18cd9fb482/waku/factory/waku.nim#L75

- `Waku.new()` constructs the delivery service unconditionally:

  https://github.com/logos-messaging/logos-delivery/blob/27ae07adaaea7beeae02cea9f8647b18cd9fb482/waku/factory/waku.nim#L213-L214

- `startWaku()` starts it unconditionally:

  https://github.com/logos-messaging/logos-delivery/blob/27ae07adaaea7beeae02cea9f8647b18cd9fb482/waku/factory/waku.nim#L418-L420

### Messaging API is not an object

[`waku/api/api.nim`](https://github.com/logos-messaging/logos-delivery/blob/27ae07adaaea7beeae02cea9f8647b18cd9fb482/waku/api/api.nim) is the only Nim-level entry point and it has no state of its own — every call dereferences `Waku.deliveryService`:

- `subscribe` → `w.deliveryService.subscriptionManager.subscribe(...)`
- `unsubscribe` → `w.deliveryService.subscriptionManager.unsubscribe(...)`
- `send` → `w.deliveryService.subscriptionManager.subscribe(...)` + `asyncSpawn w.deliveryService.sendService.send(...)`

> [!NOTE]
> There is no `MessagingClient` type that could contain Messaging API state and routines.

### Kernel API consumers transitively run Messaging API routines

Because the delivery service is created inside `Waku`, every Kernel API consumer silently runs the Messaging API send/recv/subscription loops. There is no opt-out and no boundary in code between "kernel" and "messaging".

The same is true for Node Operators: running `wakunode2` they call `Waku.new()` + `startWaku()` and therefore run the full `DeliveryService`.

This already caused a real incident. PR https://github.com/logos-messaging/logos-delivery/pull/3849 fixed a bug where DST relay nodes were saturating a store node with periodic store queries — the `recv-service`'s 5-minute `checkStore` was running on every node.

- The fix was correct, but the deeper issue is structural: a relay-only node has no business running Messaging API store-based catch-up at all. It only does so because `DeliveryService` lives inside `Waku`.

### Other Messaging API leaks beyond `deliveryService`

`DeliveryService` is the most obvious leak, but it isn't the only one. Several Messaging API concerns are wired into the kernel layer (`Waku`, `WakuNode`, `NodeHealthMonitor`, the broker, the kernel relay/filter handlers).

1. `health:connection-status` lives inside the kernel health monitor.

    <details>

    `NodeHealthMonitor` is mixed-concern. Most of it is legitimately kernel-side — per-protocol health (relay peer count, RLN/store ready), event-loop lag detection, the REST `/health` endpoint for liveness/readiness probes. A node operator wants all of that.

    But the same monitor also entangles some Messaging API-specific features:
    - `ConnectionStatus` derivation with `HealthyThreshold` thresholds
    - `EventConnectionStatusChange.emit(...)` from the health loop — emits the spec's `health:connection-status`
    - The `RequestConnectionStatus` broker provider, set up unconditionally in `startWaku`:

      https://github.com/logos-messaging/logos-delivery/blob/27ae07adaaea7beeae02cea9f8647b18cd9fb482/waku/factory/waku.nim#L428

    The monitor itself stays in the kernel. The Messaging API-shaped subset (ConnectionStatus derivation, the emit, the broker provider) lifts to `MessagingClient`, which consumes raw protocol health and peer counts and produces the Messaging API surface from them.
    </details>

2. One shared broker mixes kernel and Messaging API events.

    <details>

    Today there's a single global `BrokerContext`; both `Waku` and `WakuNode` hold a reference to it, and both kernel and Messaging API events/requests flow through the same bus:

    - Kernel events: `MessageSeenEvent`, peer events, topic-health changes.
    - Kernel requests: `RequestRelayShard`, `RequestShardTopicsHealth`, `RequestProtocolHealth`, `RequestHealthReport`.
    - Messaging API events: `MessageSentEvent`, `MessagePropagatedEvent`, `MessageReceivedEvent`, `MessageErrorEvent`, `EventConnectionStatusChange`.
    - Messaging API requests: `RequestConnectionStatus`.

    The boundary between "kernel" and "Messaging API" event surfaces exists only in naming and module ownership, not in code. Anyone with a broker handle can emit or subscribe across the boundary.

    </details>

3. Messaging API-only request types are queried by the kernel.

    <details>

    `RequestEdgeShardHealth` and `RequestEdgeFilterPeerCount` are Messaging API events, but they are *requested* from `WakuNode.startProvidersAndListeners`, so the kernel layer asks Messaging API for state:

    https://github.com/logos-messaging/logos-delivery/blob/27ae07adaaea7beeae02cea9f8647b18cd9fb482/waku/node/waku_node.nim#L514

    The dependency direction is wrong: the kernel should not know about edge-filter Messaging API state.

    </details>

## Why current state is wrong

1. **Layering violation.** The spec puts Messaging API *above* the kernel protocols. The code puts Messaging API logic *underneath* the Messaging API surface, inside the node lifecycle.
2. **No Nim identity for Messaging API.** Messaging API-only routines have nowhere natural to live and end up bolted onto `Waku` or `WakuNode`.
3. **No opt-out.** Every `Waku` runs the Messaging API engine — kernel-API consumers and node operators included.

### Real-world symptom (DST report)

This has already shown up in production. PR [#3849](https://github.com/logos-messaging/logos-delivery/pull/3849) fixed a bug where DST relay nodes were saturating a store node with periodic Messaging API store catch-up queries they didn't need. While the bug was real, the structural cause remains — relay-only nodes can't opt out of Messaging API routines.

## Roles and why layering matters

Two distinct roles consume the library (through Logos Core or directly):

- **Node Operator** — runs a fleet/relay node 24/7.
- **App Developer** — sends/receives messages on a known network. Wants Messaging API.

Both need to create and run a node. Only the App Developer needs `DeliveryService`. Today the Node Operator pays for it anyway, because it lives inside `Waku`.

### Layers — today

```mermaid
flowchart TB
    AD([App Developer])
    NO([Node Operator])
    AD --> API["waku/api/api.nim<br/>(thin free procs)"]
    API --> Waku
    NO --> Waku
    subgraph Waku["Waku"]
        DS["DeliveryService<br/>(Messaging API routines)"]
        WN["WakuNode + kernel protocols"]
        DS -.uses.-> WN
    end
```

The Node Operator path lands directly on `Waku` and silently includes `DeliveryService` because it's a field of `Waku`.

### Naming

- **Spec / surface name**: `Messaging API` (unchanged).
- **Nim implementation of the API**: `MessagingClient`

Follows the *Client* convention used in gRPC, Kafka, etc. — and avoids the awkward "create a Messaging API".

### Layers

```mermaid
flowchart TB
    AD([App Developer])
    NO([Node Operator])
    AD --> MC["MessagingClient<br/>(preset, mode, overrides)"]
    NO --> Waku["Waku<br/>(WakuNodeConf)"]
    MC --> DS["DeliveryService"]
    MC --> Waku
    DS -.uses.-> Waku
    Waku --> WN["WakuNode + kernel protocols"]
```

`Waku` is the kernel node and nothing more. `MessagingClient` is the layer that adds Messaging API routines and the Messaging API surface; the App Developer constructs it, the Node Operator never does.

```nim
# Tier: Kernel
# User: Node Operator, Tester
proc new*(T: type Waku, conf: WakuNodeConf):
  Future[Result[Waku, string]]
proc startWaku*(w: ptr Waku): Future[Result[void, string]]
proc stop*(w: Waku): Future[Result[void, string]]

# Tier: Messaging API
# User: App Developer
type MessagingClient* = ref object
  node: Waku
  deliveryService: DeliveryService
  brokerCtx: BrokerContext   # Messaging API broker (separate from Waku's kernel broker)
  # ...future Messaging API-only state

proc new*(T: type MessagingClient,
          preset: Preset, mode: WakuMode,
          overrides: WakuNodeConfOverrides = default): Result[T, string]
  ## Builds WakuNodeConf from (preset, mode, overrides), calls Waku.new(),
  ## constructs DeliveryService over it.

proc new*(T: type MessagingClient, node: Waku): Result[T, string]
  ## Escape hatch: attach Messaging API on top of an existing kernel node.
  ## Useful for tests and advanced consumers.

proc start*(self: MessagingClient): Future[Result[void, string]]
proc stop*(self: MessagingClient): Future[Result[void, string]]

proc subscribe*(self: MessagingClient, contentTopic: ContentTopic): ...
proc unsubscribe*(self: MessagingClient, contentTopic: ContentTopic): ...
proc send*(self: MessagingClient, envelope: MessageEnvelope):
  Future[Result[RequestId, string]]
```

Concretely:

- Move `deliveryService` ownership from `Waku` into `MessagingClient`. `Waku` / `startWaku` does not construct or start it.
- Split the broker into two: a **kernel broker** owned by `Waku` and a **Messaging API broker** owned by `MessagingClient`, each for the events of its tier. `MessagingClient` holds references to both — it subscribes to the kernel broker for inputs and emits Messaging API events on its own broker. A subscriber attached to `MessagingClient`'s broker cannot receive kernel events by construction.
- Drop the kernel-side queries for Messaging API state; the kernel does not know about Messaging API.
- `liblogosdelivery` constructs a `MessagingClient` and FFI calls become a thin shim over `MessagingClient` methods.

### Call order — App Developer

```mermaid
sequenceDiagram
    actor App as App Developer
    App->>MessagingClient: new(preset, mode, overrides)
    MessagingClient->>MessagingClient: build WakuNodeConf
    MessagingClient->>Waku: Waku.new(conf)
    MessagingClient->>DeliveryService: new(node)
    App->>MessagingClient: start()
    MessagingClient->>Waku: startWaku()
    MessagingClient->>DeliveryService: startDeliveryService()
    App->>MessagingClient: send(envelope)
    MessagingClient->>DeliveryService: sendService.send(task)
    DeliveryService-->>App: MessageSentEvent / MessagePropagatedEvent / MessageErrorEvent
```

### Call order — Node Operator

```mermaid
sequenceDiagram
    actor Op as Node Operator
    Op->>Waku: Waku.new(WakuNodeConf)
    Op->>Waku: startWaku()
    Note over Op,Waku: No MessagingClient.<br/>No DeliveryService loops.<br/>No periodic store catch-up.
```

### How Reliable Channel API fits

Reliable Channels sit above Messaging API and should use the same pattern as Messaging API, one level up.

The proposal extends naturally to a **third object.** With `MessagingClient` lifted out of `Waku`, `ReliableChannel.new(client, channelId, contentTopic, sdsConfig, ...)` is a clean compose. Each layer is constructable, startable, stoppable, testable.

```mermaid
flowchart TB
    RCD([Reliable-Channel App Developer])
    AD([Plain App Developer])
    NO([Node Operator])
    RCD --> RC["ReliableChannel<br/>(channelId, contentTopic,<br/>sdsConfig, rateLimitConfig)"]
    AD --> MC["MessagingClient<br/>(preset, mode, overrides)"]
    NO --> Waku["Waku<br/>(WakuNodeConf)"]
    RC --> MC
    MC --> Waku
    Waku --> WN["WakuNode + kernel protocols"]
```

Usage example:

```nim
# First create a MessagingClient
let client = MessagingClient.new(preset = TheWakuNetwork, mode = Edge)
await client.start()

# Create many channels, use same MessagingClient
let chat1 = client.createReliableChannel("channel-id-1", contentTopic="/myapp/1/chat1/proto")
let chat2 = client.createReliableChannel("channel-id-2", contentTopic="/myapp/1/chat2/proto")
```

## Related

- https://github.com/logos-messaging/logos-delivery/issues/3845
- https://github.com/logos-messaging/logos-delivery/issues/3851
- [MESSAGING-API](https://github.com/logos-messaging/specs/blob/master/standards/application/messaging-api.md)
- [RELIABLE-CHANNEL-API](https://github.com/logos-messaging/specs/blob/master/standards/application/reliable-channel-api.md)

fcecin:
## Description

NOTE: 
* This PR is WIP.
* This PR is not the _only_ way to solve #3865. I will prepare a small "Plan-B," alternative PR that can close the issue without this major refactor.
* Before reviewing code, we have to validate the _approach_. So I'm more interested on whether people like what this PR is attempting to do (the architecture).
* Having said that, code-based reviews are of course welcome, especially ones with higher-level / design concerns.

This PR refactors the code into two distinct layers: waku and messaging. Both layers publish a `nim-brokers` based interface for using the layer. The waku layer is used by the messaging layer, and both layers are used by the FFI interface to liblogosdelivery. libwaku (deprecated) is supported without any changes.

## Changes

Changed:
* Added logosdelivery_create_node_preset_mode (NOTE: we may change the signature in reviews)
* FFIContext[LogosDelivery] used for liblogosdelivery
* Create layering infra; layers have a nim-brokers-based `api/` and a dedicated `api/ffi/` folder
* Lift messaging out of the kernel into `messaging`
* Subscription engine now handles relay, edge, and direct-shard cases independently
* Core offers complete feature set (subscriptions, edge maintenance, health monitoring, ...)
* Implement EventContentTopicHealthChange (was TODO)
* Move metrics shared liblogosdelivery/libwaku to waku_telemetry
* several refactors, fixes, cleanups

_Not_ changed:
* Config system
* Any behavior (behavioral changes are bugs)
* Existing C APIs
* Any heavy renaming, e.g. Waku
* Moving `/messaging` and `/waku` into `/layers` (deferred; avoids making this PR's diff larger than it already is)
* libwaku / old kernel_api (backdoor; same structure, behavior, etc.)

TODO
* New tests (e.g. new nim-brokers layer APIs)
* Reviews, improvements, ...

## Issue

closes #3865


messaging/api/ffi/messaging_ffi.nim
```diff
@@ -0,0 +1,374 @@
...
+  if isNil(ctx):
+    let errMsg {.inject.} = opName & " failed: invalid context"
+    onError
+  elif isNil(ctx.myLib) or isNil(ctx.myLib[]):
+    let errMsg {.inject.} = opName & " failed: client is not initialized"
+    onError
+
+# ---- Construction requests (run on the FFI worker thread) ----
+
+registerReqFFI(CreateMessagingClientByPresetMode, ctx: ptr FFIContext[LogosDelivery]):
```
Ivansete-status, line 50: Continuing with my previous comment, let's perform the ffi changes in future PRs

messaging/api/api_send.nim
```diff
@@ -0,0 +1,52 @@
...
+import messaging/delivery_service/delivery_service
+import messaging/delivery_service/send_service
+import messaging/delivery_service/send_service/delivery_task
+import ./api_subscribe
+import ./types
+
+logScope:
+  topics = "messaging-api send"
+
+proc send*(
```
Ivansete-status, line 18: That file looks great to me.

---

Besides, is very interesting to move the `messaging` folder into `layers` folder so that we have the three main components (channels, messaging, protocols) in a single folder.

messaging/api/events.nim
```diff
@@ -0,0 +1,40 @@
+{.push raises: [].}
+
+## Messaging API event types. Re-exports the waku-tier event types too.
+
+import brokers/event_broker
```
Ivansete-status, line 5: That file looks correct IMO. We likely need to have an events.nim within each layer

messaging/api/api_subscribe.nim
```diff
@@ -0,0 +1,26 @@
...
+import waku/api/requests/subscription as kernel_subscription_api
+import messaging/messaging_client_type
+
+proc subscribe*(
+    client: MessagingClient, contentTopic: ContentTopic
+): Future[Result[void, string]] {.async: (raises: []).} =
+  if client.isNil():
+    return err("MessagingClient.subscribe: client is nil")
+  if client.relayMounted:
+    kernel_subscription_api.RequestRelaySubscribeContentTopic.request(
```
Ivansete-status, line 15: This is not needed because the MessagingClient already has access to `MessagingClient`/`DeliveryService`.

messaging/api/messaging.nim
```diff
@@ -0,0 +1,37 @@
...
+
+## Messaging API broker request types.
+
+import chronos
+import brokers/[broker_context, request_broker]
+import waku/waku_core/[topics/content_topic]
+import ./types
+
+# Subscribe to a content topic.
+RequestBroker:
```
Ivansete-status, line 11: This file is not needed. The dependency path is pretty clear: `channels` -> `messaging` -> `kernel`.

waku/api/types.nim
```diff
@@ -14,11 +13,6 @@ type
 
   RequestId* = distinct string
 
-  ConnectionStatus* {.pure.} = enum
-    Disconnected
-    PartiallyConnected
-    Connected
```
Ivansete-status, line 20: The movement to `waku/node/health_monitor/connection_status.nim` is correct IMO.

messaging/api/ffi/json_event.nim
```diff
@@ -5,13 +5,12 @@ type JsonEvent*[T] = ref object
   payload*: T
 
 macro toFlatJson*(event: JsonEvent): JsonNode =
-  ## Serializes JsonEvent[T] to flat JSON with eventType first, 
-  ## followed by all fields from T's payload
+  ## Serialize JsonEvent[T] to flat JSON: eventType first, then T's payload fields.
```
Ivansete-status, line 8: Just a side note: we will only use cbor in the upcoming nim-ffi bump.

messaging/delivery_service/recv_service/recv_service.nim
```diff
@@ -51,12 +49,16 @@ type RecvService* = ref object of RootObj
 proc getMissingMsgsFromStore(
     self: RecvService, msgHashes: seq[WakuMessageHash]
 ): Future[Result[seq[TupleHashAndMsg], string]] {.async.} =
-  let storeResp: StoreQueryResponse = (
-    await self.node.wakuStoreClient.queryToAny(
-      StoreQueryRequest(includeData: true, messageHashes: msgHashes)
+  let req = (
+    await kernel_store_api.RequestStoreQueryToAny.request(
```
Ivansete-status, line 53: Instead of doing that, I think is better to directly pass a reference to `wakuStoreClient`. With that, we would avoid creating the new module `waku/node/providers/store.nim`.

waku/api/ffi/kernel_ffi.nim
```diff
@@ -0,0 +1,175 @@
...
+proc waku_relay_subscribe_shard(
+    ctx: ptr FFIContext[LogosDelivery],
+    callback: FFICallBack,
+    userData: pointer,
+    pubsubTopic: cstring,
+) {.ffi.} =
+  requireInitializedKernel(ctx, "waku_relay_subscribe_shard"):
+    return err(errMsg)
+  RequestRelaySubscribeShard.request(
+    ctx.myLib[].brokerCtx, PubsubTopic($pubsubTopic)
```
Ivansete-status, line 95: The problem of making it in this way is that we don't get feedback whether the subscription actually worked or not. In this particular case, it is better to have the `WakuSubscriptionManager` subcription manager owned by `Waku/Kernel` object. Applies elsewhere to the "subcribe" broker requests.

layers/logos_delivery.nim
```diff
@@ -0,0 +1,117 @@
...
+
+logScope:
+  topics = "logos-delivery"
+
+type LogosDelivery* = ref object
+  brokerCtx*: BrokerContext
+  waku*: Waku
+    ## Kernel layer. Always present.
+  messaging*: MessagingClient
+    ## Messaging layer. `nil` in the kernel-only composition.
```
Ivansete-status, line 21: I'd rather seeing a composition where Waku belongs to MessagingClient/DeliveryService, instead of both being at the same logical level.

I may have commented somewhere else but it sounds more natural having this composition path:
ReliableChannelManager ➜ MessagingClient (DeliveryService) ➜ Waku

layers/mounts.nim
```diff
@@ -0,0 +1,38 @@
...
+
+type LayerKey = tuple[layerName: string, ctxId: uint32]
+
+var layerMounts {.threadvar.}: HashSet[LayerKey]
+
+proc isLayerMounted*(T: typedesc, ctx: BrokerContext): bool =
+  let key: LayerKey = ($T, ctx.uint32)
+  key in layerMounts
+
+proc mountLayer*(T: typedesc, ctx: BrokerContext): Result[void, string] =
```
Ivansete-status, line 22: Using that mount concept sounds interesting but maybe not needed. Shall we explore the possibility of not adding it for now? Besides, the `mount` term seems pretty related to libp2p.

liblogosdelivery/logos_delivery_api/debug_api.nim
```diff
@@ -1,56 +0,0 @@
-import std/[json, strutils]
-import waku/factory/waku_state_info
-import tools/confutils/[cli_args, config_option_meta]
-
-proc logosdelivery_get_available_node_info_ids(
-    ctx: ptr FFIContext[Waku], callback: FFICallBack, userData: pointer
-) {.ffi.} =
-  ## Returns the list of all available node info item ids that
-  ## can be queried with `get_node_info_item`.
-  requireInitializedNode(ctx, "GetNodeInfoIds"):
```
Ivansete-status, line 10: Shall we keep the ffi part unchanged and only focus on internal modules? Once we achieve that we will need to bump the nim-ffi to latest version and while refactoring the ffi part.

messaging/delivery_service/recv_service/recv_service.nim
```diff
@@ -36,9 +36,7 @@ type RecvMessage = object
 
 type RecvService* = ref object of RootObj
   brokerCtx: BrokerContext
-  node: WakuNode
   seenMsgListener: MessageSeenEventListener
-  subscriptionManager: SubscriptionManager
```
Ivansete-status, line 41: It seems the `SubscriptionManager` can be owned by `MessagingClient` because all the `subscribe` and `unsubscribe` operations will pass through it.

messaging/delivery_service/send_service/lightpush_processor.nim
```diff
@@ -40,34 +44,50 @@ method sendImpl*(
...
     debug "No peer available for Lightpush, request pushed back for next round",
       requestId = task.requestId
     task.state = DeliveryState.NextRoundRetry
     return
+  let peer = peerReq.peer.get()
 
-  let numLightpushServers = (
-    await self.lightpushClient.publish(some(task.pubsubTopic), task.msg, peer)
+  let pubReq = (
+    await kernel_lightpush_api.RequestLightpushPublish.request(
```
Ivansete-status, line 61: I like more having a reference to `lightpushClient` in this case.

messaging/delivery_service/send_service/relay_processor.nim
```diff
@@ -57,17 +56,48 @@ method sendImpl*(self: RelaySendProcessor, task: DeliveryTask) {.async.} =
     msgHash = task.msgHash.to0xHex(),
     tryCount = task.tryCount
 
-  let noOfPublishedPeers = (await self.publishProc(task.pubsubTopic, task.msg)).valueOr:
-    let errorMessage = error.desc.get($error.code)
+  let pubReq = (
+    await kernel_relay_api.RequestRelayPublish.request(
```
Ivansete-status, line 60: In this example, I think I like more having the `publishProc`, which seems more direct and explicit.

messaging/delivery_service/send_service/send_service.nim
```diff
@@ -263,9 +253,19 @@ proc send*(self: SendService, task: DeliveryTask) {.async.} =
   info "SendService.send: processing delivery task",
     requestId = task.requestId, msgHash = task.msgHash.to0xHex()
 
-  self.subscriptionManager.subscribe(task.msg.contentTopic).isOkOr:
-    error "SendService.send: failed to subscribe to content topic",
-      contentTopic = task.msg.contentTopic, error = error
+  # Auto-subscribe routes relay/edge: relay if mounted, else edge (filter).
+  if self.relayMounted:
```
Ivansete-status, line 257: If we move the `subscriptionManager` to `DeliveryService`, then, the `DeliveryService` can call the `self.subscriptionManager.subscribe` after the `self.sendService.send()` succeeded.

messaging/delivery_service/delivery_service.nim
```diff
@@ -0,0 +1,39 @@
+## This module helps to ensure the correct transmission and reception of messages
+
+import results
+import chronos, chronicles
+import brokers/broker_context
+import ./recv_service, ./send_service
+
+type DeliveryService* = ref object
```
Ivansete-status, line 8: I wonder if we could just convert it to `MessagingClient`, to simplify a little.

messaging/messaging_client.nim
```diff
@@ -0,0 +1,118 @@
...
+export api_subscribe, api_unsubscribe, api_send
+
+logScope:
+  topics = "messaging-client"
+
+proc registerMessagingApiProviders(
+    client: MessagingClient
+): Result[void, string] =
+  ## Bind the messaging broker providers to the client API procs.
+  messaging_brokers.RequestMessagingSubscribe.setProvider(
```
Ivansete-status, line 29: IMO, all the `setProviders` invoked in this proc are not needed because the `MessagingClient` will handle all the subscribe/unsubscribe/send operations.

waku/node/kernel_api/relay.nim
```diff
@@ -30,16 +30,16 @@ import
...
 export waku_relay.WakuRelayHandler
 
-declarePublicHistogram waku_histogram_message_size,
-  "message size histogram in kB",
-  buckets = [
-    0.0, 1.0, 3.0, 5.0, 15.0, 50.0, 75.0, 100.0, 125.0, 150.0, 500.0, 700.0, 1000.0, Inf
-  ]
+# NOTE: `waku_node_messages` + `waku_histogram_message_size` are declared in
+# `waku/node/waku_telemetry` (re-exported via `node/waku_node`) so both this
+# handler and WakuSubscriptionManager observe the same Prometheus collectors.
```
Ivansete-status, line 42: Why this change? This seems aside.

waku/node/peer_manager/peer_manager.nim
```diff
@@ -774,7 +774,7 @@ proc refreshPeerMetadata(pm: PeerManager, peerId: PeerId) {.async.} =
     # TODO: should only trigger an event if metadata actually changed
     #       should include the shard subscription delta in the event when
     #         it is a MetadataUpdated event
-    WakuPeerEvent.emit(pm.brokerCtx, peerId, WakuPeerEventKind.EventMetadataUpdated)
+    EventWakuPeer.emit(pm.brokerCtx, peerId, EventWakuPeerKind.EventMetadataUpdated)
```
Ivansete-status, line 777: I think we tend to use `..Event` format in both code and specs.

Ivansete-status (commented):
Thanks for the PR! Adding some comments that I hope you find useful.

fcecin:
Me and Ivan killed this PR because it is too large, and we don't think we need a formal bureaucracy separating the code in two. We will manage the relationship between our types in a case-by-case basis.

EDIT: If someone wants to resurrect the topic of formal layer modeling, we can resurrect this.

PR #3913, opened 2026-05-26
feat: introduce SDS persistency glue

NagyZoltanPeter:
# Description

This PR is updated after https://github.com/logos-messaging/nim-sds/pull/72 refactoring of SDS persistency is done.
This refactoring hugely simplified the work of the glue code.
It reduced the number of persistency points to 4 (5) and to two categories:
- channel-meta : actual complete snapshot of a channels working state
- channel-history: this is written as aggregates and evicted bundled into single write transaction with adding new messages.
- read channel history and meta is done inside complete load
- dropChannel - a sinlge transaction eviction of channles complete history and the single blob of meta state.

## Notice

setRetrievalHint is no-op by design here as long as there is a real need in nim-sds to really persist it because in the current design it is never read back from persisntecy but rebuilt on the fly.

# Outdated ------>
## Pre-read-notice

This PR introduces the glue between actaul logos-delivery persistency module and SDS library.
Until the final structure of reliable channel's (SDS within) integration happens the sds_persistency.nim hangs in the air as thus it lives under waku/persistency. Final place should be under reliable_channel/sds.

## What

Introduces the glue that persists nim-sds reliability state on top of the waku/persistency `Job` API, plus a small generic blob codec the adapter is built on.

`sds_persistency.nim` -> `newSdsPersistence` fulfills the SDS libraries persistence interface.

## Changes

- **`waku/persistency/payload_codec.nim`** (new) — generic, length-prefixed payload codec, the symmetric counterpart to `keys.nim`'s `encodePart`:
  - `writePart`/`readPart` over a `ReadCtx` cursor for primitives, `string`, `seq[byte]`, `enum`, `distinct`, `Time`, `seq[T]`, `HashSet[T]`, and tuples.
  - **4-byte BE length prefix** for strings/blobs (4 GiB ceiling) — payloads aren't sort-stable, so they don't share the 2-byte (64 KiB) cap of the key codec.
  - **`BlobCodec(T)` macro** — the single opt-in for a named value object: emits `writePart`/`readPart` from its public fields (declaration order) and reconstructs via `T.init(...)`. Because it always goes through `init`, `{.requiresInit.}` types work with no special-casing.
  - Entry points `toBlob(v)` / `fromBlob(bytes, T)` (`fromBlob` raises `ValueError` on truncated/corrupt input).

- **`waku/persistency/sds_persistency.nim`** (new) — `newSdsPersistence(job)` materialises the nim-sds `Persistence` contract onto one `Job`:
  - Every field is async (`proc(..): Future[void] {.async: (raises: []), gcsafe.}`), matching the updated nim-sds contract; `safePut`/`safeDelete` helpers trap backend errors so the closures never raise.
  - All six SDS payload types (`HistoryEntry`, `SdsMessage`, `UnacknowledgedMessage`, `IncomingMessage`, `OutgoingRepairEntry`, `IncomingRepairEntry`) are persisted via `BlobCodec`; repair rows store `(msgId, entry)` tuples.
  - Rows are namespaced per category with `channelId` as the leading key component for cheap per-channel prefix scans; `loadAllForChannel` rebuilds the snapshot, `dropChannel` wipes a channel in one transactional batch.

- **Dependency bump** (`nimble.lock`, `waku.nimble`) — nim-sds (now async `Persistence`) and brokers.

- **Tests** (`tests/persistency/test_blob_codec.nim`, wired into `test_all.nim`) — round-trips for every SDS type, a >64 KiB payload (exercising the 4-byte length), and truncation raising `ValueError`.

## Verification

- Full persistency suite: **78/78 passing** (`nimble buildTest tests/persistency/test_all.nim`), `--mm:refc`, Nim 2.2.4.
- All changed files pass `nph --check`.

# closes issue

https://github.com/logos-messaging/logos-delivery/issues/3891


tests/persistency/test_blob_codec.nim
```diff
@@ -0,0 +1,89 @@
...
+    let decoded = fromBlob(toBlob(m), SdsMessage)
+    check decoded.content.len == 70_000
+    check decoded == m
+
+  test "truncated input raises ValueError":
+    let bytes = toBlob(sampleMessage())
+    expect ValueError:
+      discard fromBlob(bytes[0 ..< bytes.len - 5], SdsMessage)
+    expect ValueError:
+      discard fromBlob(@[0xFF'u8, 0xFF, 0xFF, 0xFF], string) # claims 4 GiB
```
Ivansete-status: Why the comment of "claims 4GiB" ?

waku/persistency/payload_codec.nim
```diff
@@ -0,0 +1,310 @@
...
+type ReadCtx* = object
+  buf*: seq[byte]
+  pos*: int
+
+proc initReadCtx*(bytes: openArray[byte]): ReadCtx =
+  ReadCtx(buf: @bytes, pos: 0)
+
+proc need(r: ReadCtx, n: int) {.raises: [ValueError].} =
+  if n < 0 or r.pos + n > r.buf.len:
+    raise newException(ValueError, "truncated payload: need " & $n & " more bytes")
```
Ivansete-status: Let's use `Result` instead of raising exception

waku/persistency/payload_codec.nim
```diff
@@ -0,0 +1,310 @@
...
+
+proc writePart*(buf: var seq[byte], v: uint32) =
+  buf.add(byte((v shr 24) and 0xFF'u32))
+  buf.add(byte((v shr 16) and 0xFF'u32))
+  buf.add(byte((v shr 8) and 0xFF'u32))
+  buf.add(byte(v and 0xFF'u32))
+
+proc readPart*(r: var ReadCtx, _: typedesc[uint32]): uint32 {.raises: [ValueError].} =
+  r.need(4)
+  result =
```
Ivansete-status: Avoid implicit `result`. Applies elsewhere.

waku/persistency/sds_persistency.nim
```diff
@@ -0,0 +1,309 @@
...
+    let r = await job.get(CatLamport, chanKey)
+    if r.isOk:
+      let opt = r.get
+      if opt.isSome:
+        try:
+          snap.lamportTimestamp = fromBlob(opt.get, int64)
+        except ValueError as e:
+          warn "sds-persistency: invalid lamport bytes", channelId, err = e.msg
+    else:
+      warn "sds-persistency: get lamport failed", channelId, err = $r.error
```
Ivansete-status: It seem we can encapsulate these snippets into one proc.

waku/persistency/sds_persistency.nim
```diff
@@ -0,0 +1,309 @@
...
+BlobCodec(HistoryEntry)
+BlobCodec(SdsMessage)
+BlobCodec(UnacknowledgedMessage)
+BlobCodec(IncomingMessage)
+BlobCodec(OutgoingRepairEntry)
+BlobCodec(IncomingRepairEntry)
+
+# ── Async backing procs ─────────────────────────────────────────────────
+
+proc doLoadAll(job: Job, channelId: SdsChannelID): Future[ChannelSnapshot] {.async.} =
```
Ivansete-status: There are certain internal conditions that might be better controlled in a `Result` instead of just a `warn` log?

waku/persistency/sds_persistency.nim
```diff
@@ -0,0 +1,309 @@
...
+  ): Future[void] {.async: (raises: []), gcsafe.} =
+    await safeDelete(job, CatInRepair, key(channelId, msgId))
+
+  persistence.dropChannel = proc(
+      channelId: SdsChannelID
+  ): Future[void] {.async: (raises: []), gcsafe.} =
+    try:
+      await doDropChannel(job, channelId)
+    except CatchableError as e:
+      error "sds-persistency: dropChannel failed", channelId, err = e.msg
```
Ivansete-status: It sounds a bit safer to return `Result` for all the closers that may fail

Ivansete-status (approved):
LGTM! Thanks for it! 🙌 
Just added some minor nits that I hope you find useful

tests/persistency/test_blob_codec.nim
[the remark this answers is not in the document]
NagyZoltanPeter: Goodies! Thx.

NagyZoltanPeter: The original max claim was two bytes and thus derive only to 64K... which counting with our payload can exceed that easily. 4bytes can extend to 4GB...
Here larger claim to read is checked against valid blob length. So you cannot read more than we have.

waku/persistency/payload_codec.nim
NagyZoltanPeter: I will consider to move to use Result.

waku/persistency/payload_codec.nim
NagyZoltanPeter: Right!

waku/persistency/sds_persistency.nim
NagyZoltanPeter: Well, I will consider this, but this leads back to nim-sds where these interfaces are defined - currently they are void returns or just return the desired type.
I'm starting to think of rather we shall use brokers for interface definitions for such lib as sds.
Would be much easier.

waku/persistency/sds_persistency.nim
NagyZoltanPeter: Maybre into template, but I feel a bit overkill here. This way the code seems more readable. I would rather keep it this way.

waku/persistency/sds_persistency.nim
NagyZoltanPeter: Same as above, `dropChannel` is nim-sds defined interface. I need to go back for another round. Will track it.

darshankabariya (approved):
LGTM !

NagyZoltanPeter:
@darshankabariya and @Ivansete-status : please re-review it. This is the state after nim-sds persistency refactoring.

One open question: is it a real need to persist retrievalHints or not. Currently my assumption from nim-sds code and history of the code that it is never were intended to be re-read from disk. Thus this code leaves it intentionally no-op.

waku/persistency/persistency.nim
```diff
@@ -284,6 +284,11 @@ proc persistPut*(
 proc persistDelete*(t: Job, category: string, key: Key): Future[void] {.async.} =
   await persist(t, TxOp(category: category, key: key, kind: txDelete))
 
+proc persistDeletePrefix*(
```
Ivansete-status, line 287: I see none of the persistDeletePrefix are used. Maybe waiting for future PRs though.
```suggestion
proc persistDeletePrefix(
```

waku/persistency/sds_persistency.nim
```diff
@@ -0,0 +1,184 @@
...
+logScope:
+  topics = "sds-persistency"
+
+const
+  CatMeta* = "sds.meta"
+  CatLog* = "sds.log"
+
+# ── Public factory ──────────────────────────────────────────────────────
+
+proc newSdsPersistence*(job: Job): Persistence {.gcsafe, raises: [].} =
```
Ivansete-status, line 76: I like the proc's name but maybe we can make it more idiomatic with sth like the following:
```suggestion
proc init*(T: typedesc[Persistence], job: Job): Persistence {.gcsafe, raises: [].} =
```

waku/persistency/sds_persistency.nim
```diff
@@ -0,0 +1,184 @@
...
+
+  persistence.saveChannelMeta = proc(
+      channelId: SdsChannelID, meta: ChannelMeta
+  ): Future[Result[void, string]] {.async: (raises: []), gcsafe.} =
+    try:
+      await job.persistPut(CatMeta, toKey(channelId), encode(meta).buffer)
+      return ok()
+    except CatchableError as e:
+      warn "sds-persistency: saveChannelMeta failed", channelId, err = e.msg
+      return err(e.msg)
```
Ivansete-status, line 99: Whenever possible, let's add some more context. Applies elsewhere.
```suggestion
      return err("exception calling persistPut: " & e.msg)
```

Ivansete-status (approved):
LGTM! Thanks for it! 💯

waku/persistency/sds_persistency.nim
NagyZoltanPeter: Persist put may not mean anything to the outer world. But as we propagate error the other side of the API it can be logged correctly.

waku/persistency/sds_persistency.nim
NagyZoltanPeter: I like more the current one, the reason is the Persistence itself does not tells it is really about sds persistence. Also using as a ctor of such dispatch class maybe misleading.

waku/persistency/sds_persistency.nim
```diff
@@ -0,0 +1,177 @@
...
+##
+##  * **`saveChannelMeta`** — the complete fast-changing per-channel state
+##    (lamport clock, outgoing/incoming buffers, both SDS-R repair buffers)
+##    as ONE blob. Idempotent; a missed write self-heals on the next save.
+##  * **`updateHistory`** — append newly-delivered messages / evict the
+##    oldest past the cap, applied as one transactional batch.
+##  * **`loadChannel`** — bootstrap: returns the prior `ChannelData`
+##    (meta + ordered message history) or an empty one. Surfaces errors.
+##  * **`dropChannel`** — wipe all state for a channel. Surfaces errors.
+##  * **`setRetrievalHint`** — see below; deliberate no-op here.
```
darshankabariya: Update all the comments after we discard setRetrievalHint.

darshankabariya (approved):
LGTM. 
Made a few comments to address before merge.

waku/persistency/sds_persistency.nim
```diff
@@ -0,0 +1,177 @@
...
+            return err("loadChannel: corrupt or unsupported ChannelMeta blob")
+
+      block history:
+        let rows = (await job.scanPrefix(CatLog, chanKey)).valueOr:
+          return err("loadChannel: scan log: " & $error)
+        var msgs = newSeq[SdsMessage]()
+        for row in rows:
+          let m = SdsMessage.decode(row.payload).valueOr:
+            warn "sds-persistency: skipping undecodable log row", channelId
+            continue
```
darshankabariya, line 146: Is the silent skip intended here? Since the bloom is rebuilt from messageHistory, a skipped row is also absent from the bloom also.

waku/persistency/persistency.nim
NagyZoltanPeter: Actually it was used, but no applicable now for the actual SDS persistency needs.
But it is part of the core persistency feature set = delete by partial key (in case of composed key in use).

waku/persistency/sds_persistency.nim
NagyZoltanPeter: Yes, indeed it is a trade-off between canceling the complete load due to one row in history or having it incomplete.
Per we agreed, let's have it like this for now with a follow up issue in nim-sds side, as this must and can only be handled on that side.
> Proposal is here to separate load of channel-meta from history load. This enables better error handling and decision making.

PR #3914, opened 2026-05-28
better pending segments management

Ivansete-status:
Tackles https://github.com/logos-messaging/logos-delivery/pull/3886#discussion_r3307151055

## Issue
- https://github.com/logos-messaging/pm/issues/412

### DIFF
--- a/channels/reliable_channel.nim
+++ b/channels/reliable_channel.nim
@@ -137,11 +137,7 @@ proc pruneCompletedChannelReqs(self: ReliableChannel) =
   ## segment doesn't trigger a drop on its own — we wait until siblings
   ## are also accounted for, so the channel-level outcome is decided
   ## from a complete picture.
-  var ongoing: HashSet[RequestId]
-  for entry in self.pendingMessagingRequests:
-    if not entry.segmentSendState.isFinal:
-      ongoing.incl(entry.channelReqId)
-  self.pendingMessagingRequests.keepItIf(it.channelReqId in ongoing)
+  self.pendingMessagingRequests.keepItIf(not it.segmentSendState.isFinal())
 
 proc onMessageSent(self: ReliableChannel, messagingReqId: RequestId) =
   ## Invoked from this channel's `MessageSentEvent` listener. Flips
--- a/channels/reliable_channel.nim
+++ b/channels/reliable_channel.nim
@@ -220,7 +220,6 @@ proc onReadyToSend(
       ## parent `channelReqId` can still be pruned once its siblings
       ## are also final.
       self.pendingMessagingRequests[idx].segmentSendState = SegmentSendState.Failed
-      self.pruneCompletedChannelReqs()
       idx.inc()
       continue
     let wireBytes = seq[byte](encrypted)
@@ -254,7 +253,6 @@ proc onReadyToSend(
         ),
       )
       self.pendingMessagingRequests[idx].segmentSendState = SegmentSendState.Failed
-      self.pruneCompletedChannelReqs()
       idx.inc()
       continue
 
@@ -262,6 +260,10 @@ proc onReadyToSend(
     self.pendingMessagingRequests[idx].segmentSendState = SegmentSendState.InFlight
     self.requestIds.mgetOrPut(channelReqId, @[]).add(messagingReqId)
     idx.inc()
+
+  ## Prune once after the index walk completes. Calling it inside the
+  ## loop would shift entries and misalign `idx` for the next iteration.
+  self.pruneCompletedChannelReqs()
 
 proc send*(
     self: ReliableChannel, payload: seq[byte], ephemeral: bool = false
--- a/channels/reliable_channel.nim
+++ b/channels/reliable_channel.nim
@@ -13,7 +13,7 @@
 ##
 ## See: https://lip.logos.co/messaging/raw/reliable-channel-api.html
 
-import std/[options, tables]
+import std/[options, sets, tables]
 import results
 import chronos
 import bearssl/rand
@@ -42,27 +42,59 @@ const LipWireReliableChannelVersion* = "RELIABLE-CHANNEL-API/1"
   ## The trailing `/N` is the wire-format version and is bumped only
   ## on breaking on-the-wire changes; implementations pin one version.
 
-type ReliableChannel* = ref object
-  ## Spec-defined public type. Fields are private so callers cannot
-  ## mutate internals and break invariants. Getters are added below
-  ## for the few values consumers may need.
-  deliveryService: DeliveryService
-  channelId: ChannelId
-  contentTopic: ContentTopic
-  senderId: SdsParticipantID
-  rng: ref HmacDrbgContext
-  segmentation: SegmentationHandler
-  sdsHandler: SdsHandler
-  rateLimit: RateLimitManager
+type
+  MessagePersistence {.pure.} = enum
+    Persistent
+    Ephemeral
 
-  requestIds: Table[RequestId, seq[RequestId]]
-  pendingRequests: seq[tuple[parent: RequestId, ephemeral: bool]]
-  brokerCtx: BrokerContext
-    ## Captured here so the channel emits `ChannelMessageReceivedEvent`
-    ## on the same broker context the owning manager registered its
-    ## listeners on. Without this, an emit via `globalBrokerContext()`
-    ## would land on whatever context happens to be thread-local at
-    ## emit time, which is not necessarily the manager's.
+  SegmentSendState {.pure.} = enum
+    ## Lifecycle of a single segment as tracked by the channel. The
+    ## messaging layer has its own richer `DeliveryState` (retries,
+    ## propagated-vs-validated); here we only model what's needed to
+    ## decide when a `channelReqId` is fully accounted for.
+    AwaitingRateLimit
+      ## Pushed by `send`; not yet released by rate_limit_manager.
+    InFlight
+      ## Released by rate_limit_manager and handed to delivery_service;
+      ## `messagingReqId` is now set.
+    Confirmed
+      ## `MessageSentEvent` arrived for `messagingReqId`.
+    Failed
+      ## `MessageErrorEvent` arrived for `messagingReqId`, or the local
+      ## delivery-task construction failed before any id was reachable.
+
+  PendingMessagingRequest = object
+    ## One entry per segment (i.e. per messaging-layer request). The
+    ## relative order of `AwaitingRateLimit` entries must match the
+    ## order in which `rate_limit_manager` re-emits messages, which is
+    ## FIFO with `send()`.
+    channelReqId*: RequestId
+      ## The channel-layer parent id returned to the caller of `send()` in channel layer.
+      ## One channel request maps to N pending messaging requests.
+    messagingReqId*: Option[RequestId]
+      ## Per-segment messaging layer id. `none` until `onReadyToSend` assigns it.
+    persistenceReqType: MessagePersistence
+    segmentSendState*: SegmentSendState
+
+  ReliableChannel* = ref object
+    ## Spec-defined public type. Fields are private so callers cannot
+    ## mutate internals and break invariants. Getters are added below
+    ## for the few values consumers may need.
+    deliveryService: DeliveryService
+    channelId: ChannelId
+    contentTopic: ContentTopic
+    senderId: SdsParticipantID
+    rng: ref HmacDrbgContext
+    segmentation: SegmentationHandler
+    sdsHandler: SdsHandler
+    rateLimit: RateLimitManager
+
+    requestIds: Table[RequestId, seq[RequestId]]
+    pendingMessagingRequests: seq[PendingMessagingRequest]
+      ## Entries are kept until the matching segment reaches a final
+      ## state (`Confirmed` or `Failed`); a whole channel request is
+      ## then pruned in one pass once all its segments are final.
+    brokerCtx: BrokerContext
 
 func getChannelId*(self: ReliableChannel): ChannelId {.inline.} =
   self.channelId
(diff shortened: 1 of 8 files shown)


channels/reliable_channel.nim
```diff
@@ -73,6 +105,65 @@ func getContentTopic*(self: ReliableChannel): ContentTopic {.inline.} =
...
+  return self.pendingMessagingRequests.len
+
+proc forceInjectInFlightForTest*(
+    self: ReliableChannel, channelReqId: RequestId, messagingReqId: RequestId
+) =
+  ## Test-only: inject a pending entry already in `InFlight`. Bypasses
+  ## `send` / `onReadyToSend` so unit tests can exercise final-state
+  ## handling and the `pruneCompletedChannelReqs` rule (drop only when
+  ## *all* siblings of a `channelReqId` are final) without having
+  ## to drive — and race with — the real send pipeline.
```
NagyZoltanPeter: This sounds a wish for mockable send or delivery-service.

channels/reliable_channel.nim
```diff
@@ -73,6 +105,65 @@ func getContentTopic*(self: ReliableChannel): ContentTopic {.inline.} =
...
+  ## Drop every `pendingMessagingRequests` entry whose `channelReqId`
+  ## has all of its segments in a final state. A single failing
+  ## segment doesn't trigger a drop on its own — we wait until siblings
+  ## are also accounted for, so the channel-level outcome is decided
+  ## from a complete picture.
+  var ongoing: HashSet[RequestId]
+  for entry in self.pendingMessagingRequests:
+    if not entry.segmentSendState.isFinal:
+      ongoing.incl(entry.channelReqId)
+  self.pendingMessagingRequests.keepItIf(it.channelReqId in ongoing)
```
NagyZoltanPeter: Isn't it equals to
```suggestion
  self.pendingMessagingRequests.keepItIf(not it.segmentSendState.isFinal())
```

channels/reliable_channel.nim
```diff
@@ -73,6 +105,65 @@ func getContentTopic*(self: ReliableChannel): ContentTopic {.inline.} =
...
+  ## Invoked from this channel's `MessageSentEvent` listener. Flips
+  ## the matching `InFlight` segment to `Confirmed` and prunes. The
+  ## listener routes every event through here; entries that don't
+  ## belong to this channel simply don't match and are no-ops.
+  for i in 0 ..< self.pendingMessagingRequests.len:
+    if self.pendingMessagingRequests[i].segmentSendState == SegmentSendState.InFlight and
+        self.pendingMessagingRequests[i].messagingReqId == some(messagingReqId):
+      self.pendingMessagingRequests[i].segmentSendState = SegmentSendState.Confirmed
+      self.pruneCompletedChannelReqs()
+      return
```
NagyZoltanPeter: For exactly such scenario we have a utility already... but in a very bad place...
```nim
template applyItIf*(varSeq, pred, op: untyped) =
  for i in low(varSeq) .. high(varSeq):
    let it {.inject.} = varSeq[i]
    if pred:
      op
      varSeq[i] = it
```
https://github.com/logos-messaging/logos-delivery/blob/5b4a421c219f7b4f40a521d53da94b6f9dadf502/waku/node/delivery_service/send_service/send_service.nim#L27

We need to find place (before pushing it to std somehow) for such utilities.

channels/reliable_channel.nim
```diff
@@ -73,6 +105,65 @@ func getContentTopic*(self: ReliableChannel): ContentTopic {.inline.} =
...
+proc onMessageSent(self: ReliableChannel, messagingReqId: RequestId) =
+  ## Invoked from this channel's `MessageSentEvent` listener. Flips
+  ## the matching `InFlight` segment to `Confirmed` and prunes. The
+  ## listener routes every event through here; entries that don't
+  ## belong to this channel simply don't match and are no-ops.
+  for i in 0 ..< self.pendingMessagingRequests.len:
+    if self.pendingMessagingRequests[i].segmentSendState == SegmentSendState.InFlight and
+        self.pendingMessagingRequests[i].messagingReqId == some(messagingReqId):
+      self.pendingMessagingRequests[i].segmentSendState = SegmentSendState.Confirmed
+      self.pruneCompletedChannelReqs()
```
NagyZoltanPeter: I think it is a bad idea to do it inside the cycle as it will mutate the `pendingMessagingRequests` seq thus indexing can overrun if anything deleted meanwhile...

channels/reliable_channel.nim
```diff
@@ -73,6 +105,65 @@ func getContentTopic*(self: ReliableChannel): ContentTopic {.inline.} =
...
+      return
+
+proc onMessageError(self: ReliableChannel, messagingReqId: RequestId) =
+  ## Symmetric to `onMessageSent` but for `MessageErrorEvent`.
+  for i in 0 ..< self.pendingMessagingRequests.len:
+    if self.pendingMessagingRequests[i].segmentSendState == SegmentSendState.InFlight and
+        self.pendingMessagingRequests[i].messagingReqId == some(messagingReqId):
+      self.pendingMessagingRequests[i].segmentSendState = SegmentSendState.Failed
+      self.pruneCompletedChannelReqs()
+      return
```
NagyZoltanPeter: Same problems and suggestion as above.

channels/reliable_channel.nim
```diff
@@ -73,6 +105,65 @@ func getContentTopic*(self: ReliableChannel): ContentTopic {.inline.} =
...
+
 proc onReadyToSend(
     self: ReliableChannel, msgs: seq[seq[byte]]
 ) {.async: (raises: []).} =
```
NagyZoltanPeter, line 181: I think because of `async` it will deep copy `msgs`

channels/reliable_channel.nim
```diff
@@ -97,21 +205,45 @@ proc onReadyToSend(
...
+      contentTopic: self.contentTopic, payload: wireBytes, ephemeral: isEphemeral
     )
 
-    let deliveryReqId = RequestId.new(self.rng)
-    let deliveryTask = DeliveryTask.new(deliveryReqId, envelope, globalBrokerContext()).valueOr:
-      ## TODO: emit waku `MessageErrorEvent` for the parent request id.
+    let messagingReqId = RequestId.new(self.rng)
+    let deliveryTask = DeliveryTask.new(
+      messagingReqId, envelope, self.brokerCtx
+    ).valueOr:
```
NagyZoltanPeter: This is violation of layer boundaries, isn't it?
This should be sthg `waku.send(...)`
> I know we are in the middle of layer restructuring but it's good to keep in mind that such detail should be MessagingClient internal (current LMAPI).

channels/reliable_channel.nim
```diff
@@ -97,21 +205,45 @@ proc onReadyToSend(
...
+        self.brokerCtx,
+        MessageErrorEvent(
+          requestId: channelReqId,
+          messageHash: "",
+          error: "delivery task setup failed: " & error,
+        ),
+      )
+      self.pendingMessagingRequests[idx].messagingReqId = some(messagingReqId)
+      self.pendingMessagingRequests[idx].segmentSendState = SegmentSendState.Failed
+      self.pruneCompletedChannelReqs()
```
NagyZoltanPeter: Do not mutate the seq under index iteration.

channels/reliable_channel.nim
```diff
@@ -155,10 +294,17 @@ proc send*(
...
-    self.pendingRequests.add((parent: parentReqId, ephemeral: ephemeral))
+    self.pendingMessagingRequests.add(
+      PendingMessagingRequest(
+        channelReqId: channelReqId,
+        messagingReqId: none(RequestId),
+        persistenceReqType: persistenceReqType,
+        segmentSendState: SegmentSendState.AwaitingRateLimit,
+      )
+    )
     self.rateLimit.enqueueToSend(sdsBytes)
```
NagyZoltanPeter, line 318: I need some explanation here.
Where does the payload lives until it got sent?
Is it rateLimit manager or SDS?

I feel I miss some important puzzle from the flow. Or maybe it is just not yet completed.
I don't get if SDS sees only segments, how can it repair full messages and store channel casual history?
cc: @darshankabariya

NagyZoltanPeter (changes requested):
I think pending message seq mutation during processing must be addressed.

channels/reliable_channel.nim
Ivansete-status: Addressed in 41a43174be39b4dd4fff77847ba35add3258253c

channels/reliable_channel.nim
Ivansete-status: Addressed in 41a43174be39b4dd4fff77847ba35add3258253c

channels/reliable_channel.nim
Ivansete-status: Addressed in 41a43174be39b4dd4fff77847ba35add3258253c

channels/reliable_channel.nim
Ivansete-status: Addressed in e59b35b2e0a2e070483d48a1fb2c3b831477c3fd
The broker pipeline itself does 3 prior deep-copies (emit → emitImpl env → listenerTask env → user-listener env) which sink here can't touch, but that's something to consider for future PRs.

NagyZoltanPeter: I'm not sure about the count 3 but for such reason large event payload needs to be ref over event emit.

channels/reliable_channel.nim
Ivansete-status: You are absolutely right! Thanks! Addressed in 3f1a45c73ce7c024255bf15036cbd2bd2ff19cb8
( we'll need to properly refactor later on to use `MessagingClient` as you mentioned .)

channels/reliable_channel.nim
Ivansete-status: Addressed in 20c71c692968a6a2fd18a19c28f951a9f9583ad4

channels/reliable_channel.nim
Ivansete-status: > Where does the payload lives until it got sent?
Is it rateLimit manager or SDS?

These are very interesting questions. Now the responsibility is somehow shared between SDS and rate-limit-manager but it seems more natural to make SDS the only payload owner until delivered.

----

> I feel I miss some important puzzle from the flow. Or maybe it is just not yet completed.
I don't get if SDS sees only segments, how can it repair full messages and store channel casual history?

I think the channel's causal history will be fully managed by the SDS component.

NagyZoltanPeter: But how does SDS know about the greater picture - the full payload - if it only got segments?
But if SDS can build casual history (of full payloads) - than it secretly embeds knowledge from other layers - most probable Channel level knowledge - idk if it is the plan and whether it is ok or not.
 
I'm trying to see this as a pipeline processing but it breaks for me here and there - I need to study it more.

Ivansete-status: > But how does SDS know about the greater picture - the full payload - if it only got segments?

The whole payload picture will be handled by the segmentation component.

----

> But if SDS can build casual history (of full payloads) - than it secretly embeds knowledge from other layers - most probable Channel level knowledge - idk if it is the plan and whether it is ok or not.

Well, SDS just handles a stream of segments and it doesn't know anything about whether those segments are part of one message or are from different messages. SDS just makes sure the all the segments are properly delivered and the segmentation component just reassembles the segments into the message. The main recovery and delivery logic will be kept by SDS.

channels/reliable_channel.nim
Ivansete-status: Thanks so much indeed for the recommendation. Addressed in 9976fe3d07174c8f64572367b23c21663bc3548c
( the 3 deep copy is something claude mentioned. Something to maybe double-check but, if we use ref events then that's fine .)

channels/reliable_channel.nim
Ivansete-status: Oh right! Thanks for the suggestion. It should be better after this 2f9d9336cc82c2d9406b9922df85f1f86864d757

channels/reliable_channel.nim
```diff
@@ -42,27 +43,66 @@ const LipWireReliableChannelVersion* = "RELIABLE-CHANNEL-API/1"
...
-  brokerCtx: BrokerContext
-    ## Captured here so the channel emits `ChannelMessageReceivedEvent`
-    ## on the same broker context the owning manager registered its
-    ## listeners on. Without this, an emit via `globalBrokerContext()`
-    ## would land on whatever context happens to be thread-local at
-    ## emit time, which is not necessarily the manager's.
+type
+  SendHandler* = proc(envelope: MessageEnvelope): Future[Result[RequestId, string]] {.
+    async: (raises: [CatchableError]), gcsafe
+  .}
```
NagyZoltanPeter, line 49: For my taste this could be:
```nim
RequestBroker:
  type SendEnvelop = RequestId
  proc signature(envelop: MessageEnvelop): Future[Result[RequestId, string]] {.async.}
```

In new v3.1.0 syntax that allows this as well:
```nim
RequestBroker:
   proc SendEnvelop(envelop: MessageEnvelop): Future[Result[RequestId, string]] {.async.}
```

But in any case if the return is Result[T,E] it shall not raise any exception imho.

channels/reliable_channel.nim
```diff
@@ -73,19 +113,101 @@ func getContentTopic*(self: ReliableChannel): ContentTopic {.inline.} =
...
+  ## Invoked from this channel's `MessageSentEvent` listener. Flips
+  ## the matching `InFlight` segment to `Confirmed` and prunes. The
+  ## listener routes every event through here; entries that don't
+  ## belong to this channel simply don't match and are no-ops.
+  self.pendingMessagingRequests.applyItIf(
+    it.segmentSendState == SegmentSendState.InFlight and
+    it.messagingReqId == some(messagingReqId)
+  ):
+    it.segmentSendState = SegmentSendState.Confirmed
+  self.pruneCompletedChannelReqs()
```
NagyZoltanPeter, line 168: Still I feel it a bit dangerous, as this call hit async from an Event listener, yet you mutate the state of the pending requests list... so IDK if it can harm those cycles of processing.
The only way to know it if examining if they hit any async await or not during the loop.
I feel prune should happen better from a single controlled place.
Question: How much it harm if there are stale item in it a bit longer and not remove element that often.

channels/reliable_channel.nim
NagyZoltanPeter: Yes, that's the way to tackle it for sure, cheap copy to happen for sure.

channels/reliable_channel.nim
NagyZoltanPeter: I need to understand it better, due if SDS messageID will be stale from actual message, it's own kept history is not very useful for app but only for repair between peers... but still not guaranty that the whole message is reconstructable... only with Store queries...
But again, I need to understand the whole flow better.

channels/reliable_channel_manager.nim
```diff
@@ -55,17 +55,18 @@ proc start*(self: ReliableChannelManager): Result[void, string] =
...
   ## can register encryption providers / create channels before traffic
   ## starts flowing.
-  self.deliveryService.startDeliveryService()
+  self.waku.deliveryService.startDeliveryService()
 
 proc stop*(self: ReliableChannelManager) {.async.} =
-  if not self.deliveryService.isNil():
-    await self.deliveryService.stopDeliveryService()
+  if not self.waku.isNil():
+    await self.waku.deliveryService.stopDeliveryService()
```
NagyZoltanPeter, line 60: Why ReliableChannelManager stops Waku's DeliveryService?
I feel it still mixing responsibilities....

waku/node/delivery_service/send_service/send_service.nim
```diff
@@ -26,7 +26,7 @@ logScope:
 # This useful util is missing from sequtils, this extends applyIt with predicate...
 template applyItIf*(varSeq, pred, op: untyped) =
   for i in low(varSeq) .. high(varSeq):
-    let it {.inject.} = varSeq[i]
+    var it {.inject.} = varSeq[i]
     if pred:
       op
       varSeq[i] = it
```
NagyZoltanPeter, line 32: Yeah I though we can relocate it... maybe in another PR we can tackle these. It's my fault placing such in here.

NagyZoltanPeter (approved):
Put some more thoughts and question. Not blocking this PR as it has good advance and will have follow ups later. Thanks for it.

channels/reliable_channel.nim
Ivansete-status: Okay thanks for the recommendation, let's tackle that in a separate one.

channels/reliable_channel.nim
Ivansete-status: okay, let's tackle that in another PR too.

channels/reliable_channel_manager.nim
Ivansete-status: Yes, this is not correct, we will fix that in upcoming PRs.

waku/node/delivery_service/send_service/send_service.nim
Ivansete-status: Sure, that's fine. Thanks!

fcecin (approved):
LGTM

channels/reliable_channel.nim
Ivansete-status: This will get covered by https://github.com/logos-messaging/logos-delivery/pull/3919

PR #3915, opened 2026-05-28
feat: fetch prebuilt zerokit rln, fall back to source build

darshankabariya:
## What

The nix build now downloads the prebuilt `librln` from the [zerokit v2.0.2 release](https://github.com/vacp2p/zerokit/releases/tag/v2.0.2) instead of compiling zerokit from source.

## Why

Building zerokit from source made nix fetch ~100 Rust crates from crates.io at once. crates.io rate-limits that and returns 403, so the `liblogosdelivery` nix jobs failed randomly on both Linux and macOS.

Using the prebuilt binary means nix never touches crates.io, so the flake build is reliable (and faster).

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -17,19 +17,9 @@
       url = "github:oxalica/rust-overlay";
       inputs.nixpkgs.follows = "nixpkgs";
     };
-
-    # External flake input: Zerokit pinned to a specific commit.
-    # Update the rev here when a new zerokit version is needed.
-    zerokit = {
-      # Pinned to v2.0.2 (5e64cb8822bee65eed6cf459f95ae72b80c6ba63) to match
-      # the vendor/zerokit submodule. Keep these two in sync: the nix build
-      # links librln from this input, the Makefile build from the submodule.
-      url = "github:vacp2p/zerokit/5e64cb8822bee65eed6cf459f95ae72b80c6ba63";
-      inputs.nixpkgs.follows = "nixpkgs";
-    };
   };
 
-  outputs = { self, nixpkgs, rust-overlay, zerokit }:
+  outputs = { self, nixpkgs, rust-overlay }:
     let
       systems = [
         "x86_64-linux" "aarch64-linux"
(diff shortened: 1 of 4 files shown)


Ivansete-status (approved):

fcecin (approved):
LGTM

scripts/build_rln.sh
```diff
@@ -1,8 +1,16 @@
 #!/usr/bin/env bash
 
-# This script is used to build the rln library for the current platform.
-# Previously downloaded prebuilt binaries, but due to compatibility issues
-# we now always build from source.
+# Provides the rln static library for the current platform.
+#
+# Strategy: if zerokit publishes a prebuilt `stateless` asset for this host's
+# target triple, download it — that's faster than compiling and avoids pulling
```
Ivansete-status: What is a triple?

scripts/build_rln.sh
```diff
@@ -1,8 +1,16 @@
...
+# zerokit's ~100 crates from crates.io. Otherwise fall back to building from
+# the vendored zerokit submodule.
+#
+# (Prebuilt was dropped in #3712 and is restored here. The earlier "compatibility
```
Ivansete-status: Maybe not needed to add that. I don't know if that number refers to a PR (I think so) or an issue. Is better to add a neutral comment

Ivansete-status (approved):
Thanks!
LGTM but I'd like to get feedback from @jakubgs re the nix changes.
Cheers

scripts/build_rln.sh
darshankabariya: A "triple" is the Rust host *target triple* the platform identifier `rustc` reports (e.g. `x86_64-unknown-linux-gnu`, `aarch64-apple-darwin`)

scripts/build_rln.sh
darshankabariya: Done — dropped

igor-sirotin:
FYI, this is a common problem: 
- https://github.com/logos-messaging/libchat/pull/106/changes/038447f5e83944b0e5fe1f51691fe2b1e9f34677
- https://github.com/logos-co/logos-chat-module/issues/30
- https://github.com/status-im/status-go/pull/7485

@darshankabariya please check with infra that the solution in `logos-delivery` is ok

flake.nix
```diff
@@ -69,19 +59,78 @@
         inherit system;
         overlays = [ (import rust-overlay) nimbleOverlay ];
       };
+
+      # Prebuilt zerokit librln, fetched from the upstream GitHub release
+      # rather than compiled from source. Compiling zerokit makes Nix download
+      # its many crate dependencies from crates.io in one parallel burst, which
+      # crates.io intermittently rejects with HTTP 403 (rate limiting from the
```
jakubgs, line 66: Why are you adding 70 lines to `flake.nix` to simply do what an override of `nixpkgs` could have done?

Are you not aware this has been already fixed in upstream `nixpkgs`?

- https://github.com/NixOS/nixpkgs/pull/512735 
- https://github.com/NixOS/nixpkgs/commit/0fb82de3fbc378e21d67a8657f8afe3aae8dae16

Also, even if this fix was the correct one WHY would you pollute `flake.nix` with this garbage instead of putting the derivation in a separate file? Or fixing it in a branch in sourc repo and updating `input` to use a fixed branch?

jakubgs: This honestly looks like LLM slop.

darshankabariya: I wasn't aware of the upstream fix. I thought it was issue in crates.io's side, which is why I had the idea to use the prebuilt asset from zerokit. thinking was that it saves time by downloading instead of building locally and CI . Yes, I used Claude to implement it, but the idea was my own. 

Thanks for pointing out the nixpkgs fix, I've switched to that in the follow-up.

PR #3916, opened 2026-05-28
ci: pass -d:disableMarchNative to avoid secp256k1 build failures

Ivansete-status:
## Summary
- GitHub-hosted runners pick up host-specific CPU features via `-march=native`, which combined with `-Og -O3` (emitted when `--debugger:native` and `--opt:speed` are both active) exhausts the register constraints in secp256k1's `scalar_4x64_impl.h` inline assembly and breaks the build with `'asm' operand has impossible constraints`.
- This PR enables the existing `disableMarchNative` escape hatch in `config.nims` for CI, which falls back to `-mssse3` and keeps register allocation predictable across the heterogeneous runner pool.

Observed failure (e.g. job [78254955293](https://github.com/logos-messaging/logos-delivery/actions/runs/26564318538/job/78254955293)):
```
secp256k1/../vendor/secp256k1/src/scalar_4x64_impl.h:356:5: error: 'asm' operand has impossible constraints
secp256k1/../vendor/secp256k1/src/scalar_4x64_impl.h:470:5: error: 'asm' operand has impossible constraints
```

## Test plan
- [ ] `test-ubuntu-22.04` passes on this branch
- [ ] `x86_64-linux / liblogosdelivery` passes on this branch
- [ ] Other downstream build jobs that consume `NIMFLAGS` are unaffected

### DIFF
--- a/.github/workflows/ci-daily.yml
+++ b/.github/workflows/ci-daily.yml
@@ -8,7 +8,8 @@ on:
 env:
   NPROC: 2
   MAKEFLAGS: "-j${NPROC}"
-  NIMFLAGS: "--parallelBuild:${NPROC} --colors:off -d:chronicles_colors:none"
+  NIMFLAGS: "--parallelBuild:${NPROC} --colors:off -d:chronicles_colors:none -d:disableMarchNative"
+  NIM_PARAMS: "-d:disableMarchNative"
 
 jobs:
   build:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,7 @@ concurrency:
 env:
   NPROC: 2
   MAKEFLAGS: "-j${NPROC}"
-  NIMFLAGS: "--parallelBuild:${NPROC} --colors:off -d:chronicles_colors:none"
+  NIMFLAGS: "--parallelBuild:${NPROC} --colors:off -d:chronicles_colors:none -d:disableMarchNative"
   NIM_VERSION: '2.2.4'
   NIMBLE_VERSION: '0.22.3'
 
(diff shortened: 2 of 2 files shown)


fcecin:
My Claude is complaining that "there is more related stuff to fix." Maybe feed this back into it?

```
  1. The build job isn't covered

  ci.yml has two parallel jobs that compile secp256k1: build (lines 56-99) and test (lines 104-166). The second commit (04421708) added export 
  NIM_PARAMS=... only inside the test job's "Run tests" step. The build job runs make V=1 all with no equivalent change — so whatever the test job needed
  NIM_PARAMS for, the build job is still missing it.

  Either lift NIM_PARAMS: "-d:disableMarchNative" to the workflow-level env: block (covers both jobs in one line, and the test-job's per-step exports
  become redundant), or replicate the export in the build job's "Build binaries" step. The first is cleaner.

  2. ci-daily.yml is not updated

  .github/workflows/ci-daily.yml:11 still has plain NIMFLAGS: "--parallelBuild:${NPROC} --colors:off -d:chronicles_colors:none" — no
  -d:disableMarchNative. Same runner pool, same lottery. The daily build will keep failing intermittently for the same reason. If the goal is "stop being
  bitten by this," daily needs the same env line.
```

fcecin (approved):

Ivansete-status:
Thanks — both points addressed in 5d7dde94:

1. Lifted `NIM_PARAMS: "-d:disableMarchNative"` to the workflow-level `env:` block in `ci.yml` so both the `build` and `test` jobs inherit it. Dropped the now-redundant per-step export in the test job.
2. Mirrored the same change in `ci-daily.yml` (both `NIMFLAGS` and `NIM_PARAMS`) so the nightly hits the same flag set.

PR #3917, opened 2026-05-28
disable js-waku from ci

Ivansete-status:
Pause interaction with js-waku in CI.
Stop running js-waku interop tests in CI because they are failing.
The reason why fail is because the js-waku node need to operate with at least two relay peers because now, with one single relay peer in the js-waku interop tests, the lightpush service always returns "NO_PEERS_TO_RELAY".

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -177,20 +177,6 @@ jobs:
 
     secrets: inherit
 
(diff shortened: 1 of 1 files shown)


fcecin (approved):

NagyZoltanPeter (approved):
Agree

